PR #3585, opened 2020-10-28
Basic autogen support for --stripe-packages

aisamanra:
### Motivation
We currently do not run the `packager` when in `--autogen` mode at all. These changes allow us to run the packager but produce the same output as before by stripping off the package name when referring to constants, but also lays the groundwork for a separate "packaged autoloader" mode that generates autoloads suitable for packages.

### Test plan

See included automated tests. Added a new CLI test to verify the output with `--stripe-packages`; this can be diff'ed against the output from the other autoloader CLI test to show that the only differences are in the file paths.

### DIFF
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -80,7 +80,7 @@ NamedDefinition NamedDefinition::fromDef(const core::GlobalState &gs, ParsedFile
     QualifiedName parentName;
     if (def.data(parsedFile).parent_ref.exists()) {
         auto parentRef = def.data(parsedFile).parent_ref.data(parsedFile);
-        if (!parentRef.resolved.nameParts.empty()) {
+        if (!parentRef.resolved.empty()) {
             parentName = parentRef.resolved;
         } else {
             parentName = parentRef.name;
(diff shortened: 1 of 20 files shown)


main/realmain.cc
```diff
@@ -215,7 +215,7 @@ void runAutogen(const core::GlobalState &gs, options::Options &opts, const autog
             for (auto result = fileq->try_pop(idx); !result.done(); result = fileq->try_pop(idx)) {
                 ++n;
                 auto &tree = indexed[idx];
-                if (tree.file.data(gs).isRBI()) {
+                if (tree.file.data(gs).isRBI() || tree.file.data(gs).isPackage()) {
```
aisamanra, line 218: This just means we don't generate autoloader information for `__package.rbi` files.

main/realmain.cc
```diff
@@ -531,7 +531,9 @@ int realmain(int argc, char *argv[]) {
             gs->suppressErrorClass(core::errors::Namer::ModuleKindRedefinition.code);
             gs->suppressErrorClass(core::errors::Resolver::StubConstant.code);
 
+            indexed = pipeline::package(*gs, move(indexed), opts, *workers);
```
aisamanra, line 534: The `pipeline::package` function already handles "do not run if `--stripe-packages` is not supplied," so we don't need to do any conditional logic here.

packager/packager.cc
```diff
@@ -545,7 +545,7 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file, const Pa
     }
 
     // Sanity check: __package.rb files _must_ be typed: strict
-    if (file.file.data(ctx).strictLevel < core::StrictLevel::Strict) {
+    if (file.file.data(ctx).originalSigil < core::StrictLevel::Strict) {
```
aisamanra, line 548: We set the `strictLevel` lower artificially for autogen so we don't get errors even from invalid code. That means that in autogen every `__package.rb` file would raise this error. We can sidestep that by checking the `originalSigil`, which is retained even if `strictLevel` is set lower.

jvilk-stripe (changes requested):
My only hard request is to change the `isPackage()` helper, which could cause issues down the line / is more expensive than it needs to be.

Everything else looks good to me. Thanks for adding tests!

main/autogen/autoloader.cc
```diff
@@ -10,8 +10,13 @@
 using namespace std;
 namespace sorbet::autogen {
 
+// Like the rest of autogen, the `autoloader` pass works by walking an existing tree of data and converting it to a
+// different representation before using it. In this case, it takes the `autogen::ParsedFile` representation and
+// converts it to a `DefTree`, and then emits the autoloader files based on passed-in string fragments
+
 bool AutoloaderConfig::include(const NamedDefinition &nd) const {
-    return !nd.nameParts.empty() && topLevelNamespaceRefs.find(nd.nameParts[0]) != topLevelNamespaceRefs.end();
+    return !nd.qname.nameParts.empty() &&
```
jvilk-stripe, line 18:
```suggestion
    return !nd.qname.empty() &&
```

jvilk-stripe (approved):
LGTM

aisamanra:

→ https://go/builds/bui_IMlBHuSzWpVO0F

aisamanra:
Pay-server prework is landing so I'm merging this.

Issue #3482, opened 2020-10-06
Confusing error message referencing a variable before T.let


bhuga:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20bar%0A%20%20%20%20x%20%3D%205%0A%20%20%20%20x%20%3D%20T.let(%225%22%2C%20String)%0A%20%20end%0Aend%0A%0A)

```ruby
# typed: true
class A
  extend T::Sig

  sig {void}
  def bar
    x = 5
    x = T.let("5", String)
  end
end
```

#### Observed output

```
editor.rb:7: Incompatible assignment to variable declared via let: Integer(5) is not a subtype of NilClass https://srb.help/7001
     7 |    x = 5
                ^
Errors: 1
```

#### Expected behavior

This error is technically correct[1], but it's a little confusing. If you reference a variable 15 lines before the T.let and get the type wrong, there's no pointer to the `T.let` that helps you understand why it's wrong. I think the error message would be improved if it did any or all of these:
 * Considered it an error to use a variable before it is T.let if T.let is eventually called
 * Linked to the location of the T.let in the error message

1: The best kind of correct.
- - -

jez:
SEO pinning cfg

aprocter-stripe:
@jez I guess I'm not sure what the expected behavior actually is here. Should it be an error, or should it behave the same way as this (which the type checker seems happy with)?

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20bar%0A%20%20%20%20x%20%3D%20T.let(5%2C%20Integer)%0A%20%20%20%20T.reveal_type(x)%0A%20%20%20%20x%20%3D%20T.let(%225%22%2C%20String)%0A%20%20%20%20T.reveal_type(x)%0A%20%20end%0Aend)

```ruby
# typed: true
class A
  extend T::Sig

  sig {void}
  def bar
    x = T.let(5, Integer)
    T.reveal_type(x) # reveals `Integer`
    x = T.let("5", String)
    T.reveal_type(x) # reveals `String`
  end
end
```

jez:
From Slack:

[...]

PR #3590, opened 2020-10-29
Emit better error message when assignment precedes T.let declaration

aprocter-stripe:
Partially addresses an issue reported at #3482. If a `let`-less assignment to a variable came _before_ it was "declared" with `let`, info on the declared type would be missing, so we would default to treating the declared type as `NilClass` even if that was not the eventually declared type, which resulted in confusing error messages. With this change we check if the type info was missing rather than accepting the default, and produce a more reasonable error message.

It would be nice if we could add some signposts to the error message pointing out the `let` that conflicts with the assignment, but I think that's probably out of reach without some major work.

### Motivation

See #3482

### Test plan

See included automated tests.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1327,7 +1327,19 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                         }
                         break;
                     case cfg::CFG::MIN_LOOP_LET:
-                        if (!asGoodAs) {
+                        if (!hasType(ctx, bind.bind.variable)) {
+                            if (auto e =
+                                    ctx.beginError(bind.loc, core::errors::Infer::PinnedVariableAssignedBeforeLet)) {
+                                e.setHeader("Assignment to a variable declared via `{}` occurs before the `{}` "
+                                            "declaration has been reached",
+                                            "let", "let");
+                                if (tp.origins.size() == 1) {
+                                    e.replaceWith(fmt::format("Use `{}` at initial assignment", "let"), tp.origins[0],
+                                                  "T.let({}, {})", tp.origins[0].source(ctx),
+                                                  core::Types::dropLiteral(tp.type)->show(ctx));
+                                }
+                            }
+                        } else if (!asGoodAs) {
                             if (auto e = ctx.beginError(bind.loc, core::errors::Infer::PinnedVariableMismatch)) {
                                 e.setHeader("Incompatible assignment to variable declared via `{}`: `{}` is not a "
                                             "subtype of `{}`",
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1331,10 +1331,10 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                             if (auto e =
                                     ctx.beginError(bind.loc, core::errors::Infer::PinnedVariableAssignedBeforeLet)) {
                                 e.setHeader("Assignment to a variable declared via `{}` occurs before the `{}` "
-                                            "declaration has been reached",
-                                            "let", "let");
+                                            "declaration",
+                                            "T.let", "T.let");
                                 if (tp.origins.size() == 1) {
-                                    e.replaceWith(fmt::format("Use `{}` at initial assignment", "let"), tp.origins[0],
+                                    e.replaceWith(fmt::format("Use `{}` at initial assignment", "T.let"), tp.origins[0],
                                                   "T.let({}, {})", tp.origins[0].source(ctx),
                                                   core::Types::dropLiteral(tp.type)->show(ctx));
                                 }
--- a/test/testdata/infer/let.rb
+++ b/test/testdata/infer/let.rb
@@ -12,11 +12,11 @@ class TestLet
   end
 
   def test_assign_before_let
-    x = "foo" # error: Assignment to a variable declared via `let` occurs before the `let` declaration has been reached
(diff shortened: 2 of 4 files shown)


aprocter-stripe:
Example:

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Ax%20%3D%20%22foo%22%0Ax%20%3D%20T.let(%22bar%22%2C%20String)%0A)

```ruby
# typed: true
x = "foo"
x = T.let("bar", String)
```

Behavior before:

```
editor.rb:2: Incompatible assignment to variable declared via let: String("foo") is not a subtype of NilClass https://srb.help/7001
     2 |x = "foo"
            ^^^^^
Errors: 1
```

Behavior after:

```
editor.rb:3: Assignment to a variable declared via let occurs before the let declaration has been reached https://srb.help/7033
     3 |x = "foo"
            ^^^^^
    editor.rb:3: Replace with T.let("foo", String)
     3 |x = "foo"
            ^^^^^
Errors: 1
```
```

File: website/docs/error-reference.md [hunk 1]
```diff
@@ -986,3 +986,26 @@ See [Exhaustiveness Checking](exhaustiveness.md) for more information.
...
+
+To fix this issue, first make sure that the error is not caused by, say, copying
+and pasting a code snippet whose variable names conflict with existing code.
+Then wrap the first assignment's expression with `T.let`:
+
+```ruby
+# typed: true
+
+x = T.let("foo", String)   # <-- this will no longer produce an error
+x = T.let("bar", String)   # <-- optionally, can replace this with `x = "bar"'
```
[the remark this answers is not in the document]
jez: Sorry i don't like `<--` 😅 none of our other docs use it

infer/environment.cc
```diff
@@ -1327,7 +1327,19 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                         }
                         break;
                     case cfg::CFG::MIN_LOOP_LET:
```
jez, line 1329: I rarely get the chance to look at how loop pinning works. Would you mind summarizing for me what it means when we're in the `CFG::MIN_LOOP_LET` case?

infer/environment.cc
```diff
@@ -1327,7 +1327,19 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
...
-                        if (!asGoodAs) {
+                        if (!hasType(ctx, bind.bind.variable)) {
+                            if (auto e =
+                                    ctx.beginError(bind.loc, core::errors::Infer::PinnedVariableAssignedBeforeLet)) {
+                                e.setHeader("Assignment to a variable declared via `{}` occurs before the `{}` "
+                                            "declaration has been reached",
```
jez, line 1334:
```suggestion
                                            "declaration",
```

non-blocking suggestion

infer/environment.cc
```diff
@@ -1327,7 +1327,19 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
...
+                                            "let", "let");
```
jez, line 1335:
```suggestion
                                            "T.let", "T.let");
```

infer/environment.cc
```diff
@@ -1327,7 +1327,19 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
...
+                                if (tp.origins.size() == 1) {
+                                    e.replaceWith(fmt::format("Use `{}` at initial assignment", "let"), tp.origins[0],
```
jez, line 1337:
```suggestion
                                    e.replaceWith(fmt::format("Use `{}` at initial assignment", "T.let"), tp.origins[0],
```

jez: Also: this autocorrect is great!

test/testdata/infer/let.rb
```diff
@@ -10,4 +10,13 @@ def testit
...
   end
+
+  def test_assign_before_let
+    x = "foo" # error: Assignment to a variable declared via `let` occurs before the `let` declaration has been reached
+    x = T.let("bar", String)
+  end
+
+  def test_default_param_before_let(x = "foo") # error: Assignment to a variable declared via `let` occurs before the `let` declaration has been reached
+    x = T.let("bar", String)
+  end
```
jez, line 21: ooo nice test case!

jez (approved):
I have an undue amount of copy suggestions... I've marked the ones that I consider to be at your discretion. The unmarked ones i feel strongly about.

I also have one question so I can learn more about how pinning works.

infer/environment.cc
aprocter-stripe: As far as I understand it, I think that the `minLoops` field on each variable ordinarily tracks the minimum loop nesting depth at which that variable is ever referenced (I think blocks also count as "loops" here, but I'm not sure). This is tracked so we can enforce (in the `default` case of this switch) the restriction on changing a variable's type inside a loop, and is set for each variable at CFG build time.

If a variable is `let`-pinned, however, the restriction is different: we _are_ allowed to change the type of the variable, as long as the new type is a subtype of the pinned type. In that case we don't need to track `minLoops` at all, so we can save a little memory by encoding the fact that it is `let`-pinned as a special (negative) constant stored in `minLoops`.

aprocter-stripe:
@jez Copy suggestions always warmly welcomed :)

aprocter-stripe:
For paper trail purposes: we reverted this (#3595) because I forgot to run our standard tests against Stripe's codebase. :flushed: Will likely re-PR this soon.

PR #3593, opened 2020-10-29
Froydnj lsp code reduction

froydnj:
These changes make serializing enums more efficient by calling into conversion functions less and making the conversion functions themselves more efficient.  It might make the compiler do slightly less work in other cases.

### Motivation

Epsilon efficiency improvements.

### Test plan

Existing tests should be sufficient.


jez (approved):
cc @jvilk-stripe

PR #3587, opened 2020-10-28
fix crash in namer when redefining modules as static fields

froydnj:
In the example test cases, we would enter `A::B` into the symbol table as a static field.  But the name lookup during the symbol definition phase would find that static field when it was expecting to find a class/module, and problems would ensue.

Instead of indiscriminately looking for a matching symbol, we need to consider the context in which `squashNames` is being called, looking up classes or (dealiased) static fields as appropriate.

### Motivation

Fixes #3175.

### Test plan

See included automated tests.

PR #3594, opened 2020-10-29
print the symbol kind for constant literals

froydnj:
I think this is what's meant from https://github.com/sorbet/sorbet/pull/3587#discussion_r513747759  I think it might have been helpful for the issue in the debugger, but I don't think we can create `.exp` files for the testcases added in that issue -- or at least I couldn't make the updater produce `.exp` files. 

### Motivation

More information in the debugger is helpful.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -713,7 +713,8 @@ string ConstantLit::showRaw(const core::GlobalState &gs, int tabs) {
     printTabs(buf, tabs + 1);
     fmt::format_to(buf, "orig = {}\n", this->original ? this->original.showRaw(gs, tabs + 1) : "nullptr");
     printTabs(buf, tabs + 1);
-    fmt::format_to(buf, "symbol = {}\n", this->symbol.dataAllowingNone(gs)->showFullName(gs));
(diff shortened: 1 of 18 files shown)


ast/Trees.cc
```diff
@@ -745,7 +745,8 @@ string ConstantLit::showRaw(const core::GlobalState &gs, int tabs) {
     printTabs(buf, tabs + 1);
     fmt::format_to(buf, "orig = {}\n", this->original ? this->original->showRaw(gs, tabs + 1) : "nullptr");
     printTabs(buf, tabs + 1);
-    fmt::format_to(buf, "symbol = {}\n", this->symbol.dataAllowingNone(gs)->showFullName(gs));
+    fmt::format_to(buf, "symbol = {}, symbol-kind = {}\n", this->symbol.dataAllowingNone(gs)->showFullName(gs),
+                   this->symbol.dataAllowingNone(gs)->showKind(gs));
```
elliottt, line 749: What about `symbol = ({symbol-kind}) {full-name}\n` instead?

elliottt (changes requested):
I like this change! What about changing the format slightly to not include the comma? There was a suggestion in the linked comment about giving a prefix, and I suggested putting the symbol-kind in parens instead.

elliottt (approved):
🚢

PR #3597, opened 2020-10-29
Re-do "Emit better error message when assignment precedes T.let declaration"

aprocter-stripe:
_(Re-do of #3590, which was reverted at #3595 because I forgot to run some internal tests.)_

Partially addresses an issue reported at #3482. If a `let`-less assignment to a variable came _before_ it was "declared" with `let`, info on the declared type would be missing, so we would default to treating the declared type as `NilClass` even if that was not the eventually declared type, which resulted in confusing error messages. With this change we check if the type info was missing rather than accepting the default, and produce a more reasonable error message.

It would be nice if we could add some signposts to the error message pointing out the `let` that conflicts with the assignment, but I think that's probably out of reach without some major work.

### Motivation

See #3482

### Test plan

See included automated tests.


aprocter-stripe:

→ https://go/builds/bui_IIIm2mK2mI0ztU

aprocter-stripe:
Hello me.

PR #3569, opened 2020-10-24
[WIP] Use tagged pointer for TypePtr, reduce/remove virtual dispatch in types

jvilk-stripe:
This PR is super WIP and will be landed as a bunch of smaller changes. The commits in this PR build on each other and are each individually green (at least, in testing locally) which will make it easy to cherry-pick them into smaller PRs later.

Still TODO:
- [x] Inline ClassType (and potentially some other types) into the tagged pointer.
- [x] Remove ProxyType
- [ ] Determine if it is feasible to make BlamedUntyped/etc not use subclassing on ClassType.
  - Possible to set a bit in ClassType and store data in a object to the side?
  - Could steal the `counter` field on TypePtr to point to something...?

### Motivation

Speed. I noticed atomic counter contention on shared class types (e.g., nilClass) is expensive during infer due to cache thrashing. This is a super roundabout way of fixing that problem that also happens to have other benefits (smaller Type objects, faster casting, etc).

### Test plan

See included automated tests.

PR #3598, opened 2020-10-30
Use TypePtr tag in serialization logic

jvilk-stripe:
Use TypePtr tag in serialize. Now we can lean on compiler-enforced enum exhaustiveness checking to complain if we add a new type! :D 

Also, no more magic constants for type type!

### Motivation

Cherry-picked from https://github.com/sorbet/sorbet/pull/3569

### Test plan

Covered by existing tests.


elliottt (approved):
🎉

PR #3599, opened 2020-10-30
Stop using typecase for Types.

jvilk-stripe:
Instead, use a sequence of if/else casts that leverage tags _or_ use a switch statement on tags.

Why not use a switch statement everywhere? Unfortunately, we are using the class hierarchy for some types. For example, `ClassType` has separate subtypes. Same with `GroundType` and `ProxyType`. So a `switch` author would need to know about that, and I think that might make the code a little harder to grok. 

Note that `typecase` does this sort of if/else check, except with `fast_cast`, so this should still be strictly better.

This avoids using `fast_cast` for determining castability, which uses typeid/dynamic casts.

### Motivation

Broken off from https://github.com/sorbet/sorbet/pull/3569

### Test plan

Covered by existing tests.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -1,6 +1,5 @@
 #include "ast/Trees.h"
 #include "common/formatting.h"
-#include "common/typecase.h"
 #include "core/Symbols.h"
 #include <sstream>
 #include <utility>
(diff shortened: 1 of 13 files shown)


jez:
Can we measure this? I really like the ergonomics of typecase and I wouldn’t want to give it up for a wash on performance.

elliottt:
> Can we measure this? I really like the ergonomics of typecase and I wouldn’t want to give it up for a wash on performance.

It'll be a necessary step to fully remove the `Type` base class, allowing us to remove the vtable for types entirely. Maybe we can rework `typecase` to rely on tags instead of inheritance?

jvilk-stripe:
Removing review request; I'm going to spend some time on Godbolt to see if I can make a tag-based typecase degenerate into a switch statement. Some complexities:

* We have a few types that map to multiple enums due to subclasses (ClassType, GroundType, ...).
* `ClassType` and a few others will eventually be inlined into the pointer, so they will cast to a value type rather than a reference/pointer type. I need to make sure we can have typecases for `ClassType c`!

jvilk-stripe:

→ https://go/builds/bui_IIzfoMCY6psmgI

jvilk-stripe:
Closed in favor of https://github.com/sorbet/sorbet/pull/3605

@jez there is no huge performance win; it's minor. But as Trevor says, it's needed to remove vtables.

PR #3601, opened 2020-10-31
Remove typecase from Tree serialization

elliottt:
Switch on the tree tag instead of using typecase in the pickle and unpickle code for `TreePtr`. As an added bonus, stop serializing any location information for `EmptyTree`.

### Motivation

Fewer magic numbers, potential performance wins.

### Test plan

See included automated tests.

### DIFF
--- a/core/serialize/serialize.cc
+++ b/core/serialize/serialize.cc
@@ -917,151 +917,151 @@ void SerializerImpl::pickle(Pickler &p, const ast::TreePtr &what) {
 
     switch (tag) {
         case ast::Tag::Send: {
-            auto *s = ast::cast_tree<ast::Send>(what);
-            pickle(p, s->loc);
-            p.putU4(s->fun._id);
+            auto &s = ast::cast_tree_nonnull<ast::Send>(what);
+            pickle(p, s.loc);
+            p.putU4(s.fun._id);
             u1 flags;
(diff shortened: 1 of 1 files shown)


core/serialize/serialize.cc
```diff
@@ -911,25 +906,19 @@ CachedFile Serializer::loadFile(const core::GlobalState &gs, core::FileRef fref,
...
-    typecase(
-        what.get(),
-        [&](ast::Send *s) {
-            pickleAstHeader(p, 2, s->loc);
+    auto tag = what.tag();
+    p.putU4(static_cast<u4>(tag));
+
+    switch (tag) {
+        case ast::Tag::Send: {
+            auto *s = ast::cast_tree<ast::Send>(what);
```
jvilk-stripe, line 920: You're paying the cost of two tag checks here. A nonnull cast is 'free'.

```suggestion
            auto &s = ast::cast_tree_nonnull<ast::Send>(what);
```

jvilk-stripe (approved):
One nit: Use nonnull casts to avoid checking tag twice.

core/serialize/serialize.cc
elliottt: Yes, but this was the least invasive refactoring. I guess given how large the diff already is, it's probably not a big deal to change a bunch of `->` to `.`.

Issue #3435, opened 2020-09-18
Go to implementation (on abstract method, RBI method)


jez:
### Problem

I want to click a button and have Sorbet show me all the implementations of a given abstract method.

Find all references doesn't do this, because technically the abstract method definition and the implementation method definition are separate definitions on separate classes.

### Proposed solution

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_implementation

There is a `textDocument/implementation` request that we can probably use for this.

We might also want to use that request to implement "jump from RBI definition to source code definition".

jez:
Implemented in #4598

PR #3607, opened 2020-11-02
"Go to Implementation" LSP feature

ilyailya:
Change brings a "Go to implementation" feature for the Sorbet LSP. There are four main scenarios how the feature can be called:

- From the definition of the abstract function
- From the call of the abstract function
- From the reference of the abstract class
- From the definition of the abstract class

The algorithm behind the feature is trivial: if a request was called on the method node, I go up to the owner class node of the function. If the owner class is abstract, I go over all of the defined classes and pick ones, which derive from owner class. Then find a method, which overrides the originally requested one. Logic for the other cases is similar. 

gti1

gti2

### Motivation

https://github.com/sorbet/sorbet/issues/3435

### Test plan

I've update an existing tests to pass. Change requires a proper testing like it's done for the "Find Definition" andthe "Find All References" feature. It will be implemented in the further PRs for the sake of easy and small reviews. Postponing tests shouldn't be a problem, because all new functionality is hidden behind a flag.


main/options/options.h
```diff
@@ -211,6 +211,7 @@ struct Options {
     bool lspDocumentSymbolEnabled = false;
     bool lspDocumentFormatRubyfmtEnabled = false;
     bool lspSignatureHelpEnabled = false;
+    bool lspGoToImplementationEnabled = true;
```
jvilk-stripe, line 214:
```suggestion
    bool lspGoToImplementationEnabled = false;
```

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,58 @@
...
+
+    if (queryResult.error) {
+        // An error happened while setting up the query.
+        response->error = move(queryResult.error);
+        return response;
+    }
+
+    vector<unique_ptr<Location>> result;
+    for (auto const &queryResponse : queryResult.responses) {
+        auto def = queryResponse->isDefinition();
```
jvilk-stripe, line 29: I think you should also be able to support go to implementation from:
* Abstract class definition sites
* Abstract class/module references
* Abstract method call sites.

As-written, this implementation only supports go to implementation from abstract method definition sites.

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,58 @@
...
+        if (!isAbstract) {
+            response->error = make_unique<ResponseError>(
+                (int)LSPErrorCodes::InvalidParams, "Go to definition can be used only for methods of abstract classes");
+        }
+
+        auto used = gs.classAndModulesUsed();
+        for (u4 idx = 1; idx < used; idx++) {
+            core::SymbolRef ref(gs, core::SymbolRef::Kind::ClassOrModule, idx);
+            if (ref.data(gs)->derivesFrom(gs, owningClassSymbolRef)) {
+                auto members = ref.data(gs)->members();
```
jvilk-stripe, line 44: I think you can use `findMember` instead of this loop:

https://github.com/sorbet/sorbet/blob/922d74ca/core/Symbols.h#L479

jvilk-stripe (changes requested):
Initial review prior to our meeting. :)

main/options/options.h
ilyailya: I had it set to `true` for the ease of testing. Change it to `false` as it should be.

main/lsp/requests/go_to_implementation.cc
ilyailya: Added the support for this use cases in the last commit. Please take a look!

main/lsp/requests/go_to_implementation.cc
ilyailya: Thank you, this is really handy

ilyailya:
@jvilk-stripe I've added new changes with an input form your comments. Can you take a look please?

test/helpers/lsp.cc
```diff
@@ -208,7 +208,7 @@ void checkServerCapabilities(const ServerCapabilities &capabilities) {
     // We don't support all possible features. Make sure we don't make any false claims.
     CHECK(capabilities.definitionProvider.value_or(false));
     CHECK(capabilities.typeDefinitionProvider.value_or(false));
-    CHECK_FALSE(capabilities.implementationProvider.has_value());
+    CHECK(capabilities.implementationProvider.has_value());
```
jvilk-stripe, line 211: We should be asserting that this is `true` (we enable all experimental features in tests):

```suggestion
    CHECK(capabilities.implementationProvider.value_or(false));
```

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,113 @@
+#include "main/lsp/requests/go_to_implementation.h"
+#include "core/lsp/QueryResponse.h"
+#include "main/lsp/json_types.h"
+#include "main/lsp/lsp.h"
+
+using namespace std;
+
+namespace sorbet::realmain::lsp {
+
+vector<core::SymbolRef> findChildren(const core::GlobalState &gs, core::SymbolRef ancestorClass) {
```
jvilk-stripe, line 10: Nit: You should put these freestanding non-class functions into an anonymous namespace like so:

```
namespace {
  // functions defined here
}
```

The anonymous namespace tells the compiler that these functions will be compiled with "internal linkage", meaning that they cannot be referenced outside of this `.cc` file (by default, this is not the case!).

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,113 @@
...
+
+    if (queryResult.responses.empty()) {
+        return response;
+    }
+
+    vector<unique_ptr<Location>> result;
+    auto queryResponse = move(queryResult.responses[0]);
+    if (auto def = queryResponse->isDefinition()) {
+        // User called "Go to Implementation" from the abstract function definition
+        core::SymbolRef method = def->symbol;
```
jvilk-stripe, line 68: This may not be a method. It could also be the definition site to e.g. a class. If this codepath requires a method, you should check if `def->symbol.data(gs)->isMethod()`.

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,113 @@
...
+    for (u4 idx = 1; idx < used; idx++) {
+        core::SymbolRef ref(gs, core::SymbolRef::Kind::ClassOrModule, idx);
+        if (ref.data(gs)->derivesFrom(gs, ancestorClass)) {
+            results.push_back(ref);
+        }
+    }
+    return results;
+}
+
+variant<vector<core::Loc>, unique_ptr<ResponseError>> findMethodImplementations(const core::GlobalState &gs,
```
jvilk-stripe, line 22: Feel free to make this a simple `struct` and define it in the anonymous namespace. I think that'll be easier to code against than variants, which are a bit awkward, and will be faster to compile.

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,113 @@
...
+                                                                                core::SymbolRef method) {
+    vector<core::Loc> locations;
+    auto owningClassSymbolRef = method.data(gs)->owner;
+    auto isAbstract = owningClassSymbolRef.data(gs)->isClassOrModuleAbstract();
```
jvilk-stripe, line 26: `owner` may not be a class or module, so you should check that before calling `isClassOrModuleAbstract`:

```suggestion
    auto isAbstract = owningClassSymbolRef.data(gs)->isClassOrModule() && owningClassSymbolRef.data(gs)->isClassOrModuleAbstract();
```

main/lsp/requests/go_to_implementation.cc
jvilk-stripe: btw, there's a new function in `rename.cc` that may be useful called `getSubclasses`:

https://github.com/sorbet/sorbet/blob/HEAD/main/lsp/requests/rename.cc#L61

It's a bit of a misnomer; it finds all subclasses and, if a module, all modules that have the module mixed in. So, it exactly matches your use case here. It's much more efficient than your current loop, as it reimplements `derivesFrom` in a manner that memoizes intermediate results.

I'd be down to move it to a helper method on `LSPTask` so it can be reused in a place like this.

jvilk-stripe: You don't have to change your code immediately to use this method or anything; that can be a later change.

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,113 @@
...
+        } else if (auto locations = get_if<vector<core::Loc>>(&locationsOrError)) {
+            for (const auto &location : *locations) {
+                addLocIfExists(gs, result, location);
+            }
+        }
+    } else if (auto constant = queryResponse->isConstant()) {
+        // User called "Go to Implementation" from the abstract class reference
+        auto classSymbol = constant->symbol;
+
+        if (!classSymbol.data(gs)->isClassOrModuleAbstract()) {
```
jvilk-stripe, line 83: You need to check if this is a classOrMethod before calling this method. A constant symbol could be something else (e.g., a type argument or a static field).

```suggestion
        if (!classSymbol.data(gs)->isClassOrModule() || !classSymbol.data(gs)->isClassOrModuleAbstract()) {
```

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,113 @@
...
+            response->error = make_unique<ResponseError>(
+                (int)LSPErrorCodes::InvalidParams,
+                "Go to implementation can be used only for methods or references of abstract classes");
+            return response;
+        }
+
+        auto childClasses = findChildren(gs, classSymbol);
+        for (const auto &childClass : childClasses) {
+            addLocIfExists(gs, result, childClass.data(gs)->loc());
```
jvilk-stripe, line 92: Classes can be reopened in multiple files, so they may have multiple locs. You may want to have an inner loop here:
```suggestion
            for (auto loc : childClass.data(gs)->locs()) {
              addLocIfExists(gs, result, loc);
            }
```

main/lsp/requests/go_to_implementation.cc
jvilk-stripe: Do you want to handle "Go to Implementation" on the definition of an abstract class itself? Looks like you aren't doing that right now.

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,113 @@
...
+        if (auto error = get_if<unique_ptr<ResponseError>>(&locationsOrError)) {
+            response->error = move(*error);
+            return response;
+        } else if (auto locations = get_if<vector<core::Loc>>(&locationsOrError)) {
+            for (const auto &location : *locations) {
+                addLocIfExists(gs, result, location);
+            }
+        }
+    }
+
```
jvilk-stripe, line 109: I wonder if it would be cleaner if the above logic extracted a `SymbolRef` (be it a method or an abstract class) and then you had logic down here that handled those two cases.

jvilk-stripe (changes requested):
We're getting there! I left some comments that should help avoid crashes and other issues.

main/lsp/requests/go_to_implementation.cc
ilyailya: Thanks for an idea with a check! I've added that.

The case of with abstract class definition is handled in go_to_implementation.cc lines 68-81. 
I'll update the description with all use cases I have

ilyailya: What else can it be if not a method? I've tried calling go to implementations on abstract class definition and it went to `queryResponse->isConstant()` branch

main/lsp/requests/go_to_implementation.cc
ilyailya: I moved this code to `Symbols.cc`, because it didn't fit well into `LSPTask`. I also added a param with a default value to filter the self out of the result.

main/lsp/requests/go_to_implementation.cc
ilyailya: I've replaced `variant` with custom struct. Can you explain why it'll be faster? I don't really understand c++, but would be glad to learn

main/lsp/requests/go_to_implementation.cc
ilyailya: Surprisingly, this didn't work out. `Symbol::locs` store only two values. As I understood, it should be the primary location of the symbol and optional rbi file. I've added new property, which stores all the possible locations of the module/class.

However, I'm not sure how hard it will hit the memory consumptions. Would it be a large issue, how do you think?

main/lsp/requests/go_to_implementation.cc
ilyailya: Can you elaborate on this a bit? I didn't get, which exact logic I should extract into `SymbolRef`

ilyailya:
I want tests to be running automatically on push, so I'm going to close this PR and open a new one based on the branch in sorbet repo, not in my fork.

PR #3608, opened 2020-11-02
Add CLI option to error on kwargs hash without splat

aprocter-stripe:
In Ruby 2.7, passing a keyword arg hash without `**` results in a warning:

```
$ cat badsplat.rb 
def f(x, y:)
end

someArgs = {y: 3}
f(6, someArgs)
$ ruby -W badsplat.rb 
badsplat.rb:5: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
badsplat.rb:1: warning: The called method `f' is defined here
```

This will become an error in Ruby 3.0.

It would be nice to have Sorbet detect this. However, we don't want to make it an error by default since that would probably be too disruptive.

This PR adds an error for "legacy" kwarg hashes, hidden behind a `--error-kwarg-hash-without-splat` CLI option. It makes an autocorrect suggestion (e.g., changing `someArgs` to `**someArgs` in the above example) as well, but that autocorrect suggestion doesn't actually work yet, due to a bug in Sorbet.

### Motivation
Should be handy for folks upgrading to 2.7 and/or looking ahead to 3.0.

### Test plan

See included automated tests. Had to make them CLI tests instead of corpus tests, since I don't think there's any way to pass the flag into the corpus tests.


core/types/calls.cc
```diff
@@ -696,7 +696,19 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, DispatchArgs args, core
...
-        // NOTE: this would be a good place for an autocorrect to using `**kwhash`
+        auto splatLoc = core::Loc(args.locs.file, args.locs.args[args.args.size() - 1]);
+
+        // If --error-kwarg-hash-without-splat is set, we will treat "**-less" keyword hash argument as an error.
+        if (gs.errorKwargHashWithoutSplat) {
+            if (auto e = gs.beginError(splatLoc, errors::Infer::KeywordArgHashWithoutSplat)) {
+                e.setHeader("(`{}`) Supplying a keyword argument hash as the last "
+                            "positional argument without using `{}` is deprecated "
+                            "in Ruby 2.7, and will be an error in Ruby 3.0",
+                            "--error-kwarg-hash-without-splat", "**");
```
elliottt, line 707: This error looks a little bit long, how does it look on an 80-col terminal? You could also use `e.addErrorSection` to provide more information below the actual error.

elliottt (changes requested):
This looks good to me! Only one comment on the length of the error message.

core/types/calls.cc
aprocter-stripe: I split this up into a few lines. Still doesn't quite _fit_ on an 80-column terminal, but won't look as bad as this was. What do you think?

```
badsplat.rb:7: Keyword argument hash without ** is deprecated https://srb.help/7033
     7 |f(6, someArgs)
             ^^^^^^^^
    badsplat.rb:7: This produces a runtime warning in Ruby 2.7, and will be an error in Ruby 3.0
     7 |f(6, someArgs)
             ^^^^^^^^
    badsplat.rb:7: Note: disabling --error-kwarg-hash-without-splat will suppress this error
     7 |f(6, someArgs)
             ^^^^^^^^
    badsplat.rb:7: Replace with **someArgs
     7 |f(6, someArgs)
             ^^^^^^^^
Errors: 1
```

core/types/calls.cc
```diff
@@ -696,7 +696,20 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, DispatchArgs args, core
...
+        auto splatLoc = core::Loc(args.locs.file, args.locs.args[args.args.size() - 1]);
+
+        // If --error-kwarg-hash-without-splat is set, we will treat "**-less" keyword hash argument as an error.
+        if (gs.errorKwargHashWithoutSplat) {
+            if (auto e = gs.beginError(splatLoc, errors::Infer::KeywordArgHashWithoutSplat)) {
+                e.setHeader("Keyword argument hash without `{}` is deprecated", "**");
+                e.addErrorLine(splatLoc, "This produces a runtime warning in Ruby 2.7, "
+                                         "and will be an error in Ruby 3.0");
+                e.addErrorLine(splatLoc, "Note: disabling `{}` will suppress this error",
+                               "--error-kwarg-hash-without-splat");
```
elliottt, line 708: How about removing this line? Since it's off by default, It would be something that the user would know they had enabled already.

aprocter-stripe: Sure, that sounds good.

elliottt (approved):

aprocter-stripe:

→ https://go/builds/bui_IJph2igyGDSAnb

core/GlobalState.h
```diff
@@ -223,6 +223,10 @@ class GlobalState final {
     // Contains a string to be used as the base of the error URL.
     // The error code is appended to this string.
     std::string errorUrlBase;
+
+    // If 'true', issue an error for pre-Ruby 2.7-style keyword hash argument (without '**').
+    bool errorKwargHashWithoutSplat = false;
```
jez, line 228: I think we will likely want to add multiple Ruby 3 keywords-related errors, and I don't know that it makes sense to add options for each one of them, given that you can already `--error-black-list` individual error codes if you don't want every kind of keyword args error.

aprocter-stripe: @jez So something like `--error-obsolete-kwargs`, which would enable this new error but possibly _also_ enable future kwarg-related errors?

jez: Yes, I would have called the flag `--ruby3-keyword-args` and made it emit errors for code known to cause errors in Ruby 3.0 (i.e., deprecation warnings in Ruby 2.7)

PR #3610, opened 2020-11-02
`URI.regexp` returns `Regexp` not `String` array

paracycle:
As documented in the description of the `URI.regexp` method, the return value of the method is always a `Regexp` instance, regardless of the values passed into the `schemes` parameter.

### Motivation

With the current method definition valid code like the one below will cause errors with type-checking:

```ruby
body = "Some text that includes a URI: https://stripe.com"
URI.regexp.match?(body) #=> true
```

### Test plan

No automated tests.


jez (approved):
Thanks!

PR #3611, opened 2020-11-03
remove support for boolean types from LiteralType

froydnj:
We had two different ways of representing this information and we weren't always consistent in picking which one.  This version of the representation at least makes everything consistent.  This representation of `true` and `false` is also slightly smaller.

### Motivation

Consistency.  No more confusion when looking at `sorbet_llvm` code and wondering why `LiteralTypeKind::{True,False}` aren't being handled.

### Test plan

### DIFF
--- a/proto/Type.proto
+++ b/proto/Type.proto
@@ -23,7 +23,8 @@ message Type {
             INTEGER = 0;
             STRING = 1;
             SYMBOL = 2;
-            // skip a few
+            // TRUE = 3;
+            // FALSE = 4;
             FLOAT = 5;
         }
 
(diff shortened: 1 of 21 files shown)


proto/Type.proto
```diff
@@ -23,8 +23,7 @@ message Type {
             INTEGER = 0;
             STRING = 1;
             SYMBOL = 2;
-            TRUE = 3;
-            FALSE = 4;
+            // skip a few
```
jez, line 26:
```suggestion
            // TRUE = 3;
            // FALSE = 4;
```

jez (approved):

PR #3614, opened 2020-11-04
Fix Numeric#truncate signature to support optional ndigits arg

Colex:
Fix the type signature of `Numeric#truncate` to match Ruby's
documentation.
https://docs.ruby-lang.org/en/2.6.0/Numeric.html#method-i-truncate

The new signature adds support to optional `ndigits` argument and
expands the return types to `Integer` and `Float` when the argument
is given. When not `ndigits` is not present, the method returns
always `Integer`.

### Motivation

Currently, the `Numeric#truncate` signature does not support the optional argument `ndigits` as described in the documentation.

### Test plan

RBI change


elliottt:

→ https://go/builds/bui_IKHn0jr28Tj6FC

elliottt (approved):
Thanks for fixing this signature!

PR #3615, opened 2020-11-04
Move a bunch of functions from `Type` to `TypePtr`

jvilk-stripe:
Move a bunch of functions from `Type` to `TypePtr`.

In the next PR, I will move the rest and may introduce some preprocessor/templatey goodness to clean up those case statements.

### Motivation

Required to delete `Type` class and make all `Type` objects smaller.

### Test plan

Covered by existing tests.


core/TypePtr.cc
```diff
@@ -91,4 +91,325 @@ bool TypePtr::isBottom() const {
...
+            return 2;
+        case Tag::LiteralType:
+            return 3;
+        case Tag::ShapeType:
+            return 4;
+        case Tag::TupleType:
+            return 5;
+        case Tag::LambdaParam:
+        case Tag::SelfTypeParam:
+            return 6;
```
elliottt, line 111: It's unfortunate how close `kind()` is to `tag()`.

core/TypePtr.cc
```diff
@@ -91,4 +91,325 @@ bool TypePtr::isBottom() const {
...
+            return "OrType";
+        case Tag::AndType:
+            return "AndType";
+        case Tag::SelfType:
+            return "SelfType";
+    }
+}
+
+bool TypePtr::isFullyDefined() const {
+    switch (tag()) {
```
elliottt, line 165: This version of `isFullyDefined` is really easy to reason about, with most of the cases all in one place 👍

core/types/calls.cc
```diff
@@ -1103,25 +1099,28 @@ TypePtr getMethodArguments(const GlobalState &gs, SymbolRef klass, NameRef name,
     return TupleType::build(gs, move(args));
 }
 
-TypePtr ClassType::getCallArguments(const GlobalState &gs, NameRef name) {
+TypePtr ClassType::getCallArguments(const GlobalState &gs, NameRef name) const {
     if (hasUntyped()) {
-        return Types::untyped(gs, untypedBlame());
+        return Types::untyped(gs, Symbols::noSymbol());
```
elliottt, line 1104: Nice: it's OK to use `Symbols::noSymbol()` here because of the specialized implementation for `BlameUntyped` below.

elliottt (approved):
This change is great! I'm definitely biased, but I much prefer pattern-matching (or matching on tags in this case) to overloads 👍

core/TypePtr.cc
jvilk-stripe: If there weren't ties I'd just align the enum type to match `kind`. But there are ties, unfortunately.

PR #3616, opened 2020-11-04
Rename --error-kwarg-hash-without-splat to --ruby3-keyword-args

aprocter-stripe:
### Motivation

Per @jez comments on #3608, `--ruby3-keyword-args` is probably a better name for this flag, since down the road there may be other errors we'll want to add than just "you have to use `**` on a keyword arg hash passed positionally", and we won't want to add a separate enable flag for each of those.

### Test plan

See included automated tests.


jez (approved):

elliottt (approved):

aprocter-stripe:

→ https://go/builds/bui_IKWFiXt8CoXeq0

PR #3619, opened 2020-11-06
Inline Type::hasUntyped and isFullyDefined into TypePtr methods

jvilk-stripe:
Inline Type::hasUntyped and isFullyDefined into TypePtr methods.

These methods are trivial and recursive (e.g., composite types call the method on the types they contain). It's easier to understand them within the context of one big switch statement than distributed between different methods.

### Motivation

Simplify code.

### Test plan

Covered by existing tests.


elliottt (approved):
Thanks for refactoring this! It's much easier to think about these operations when they're all together in one function 👍

Issue #36, opened 2019-01-16
Better error message on unnecessary safe navigation


DarkDimius:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams(x%3A%20Integer).void%7D%0Adef%20foo(x)%0A%20%20x%26.to_s%0Aend)

```ruby
# typed: strict
extend T::Sig

sig {params(x: Integer).void}
def foo(x)
  x&.to_s
end
```

#### Observed output

```
editor.rb:6: This code is unreachable https://srb.help/7006
     6 |  x&.to_s
          ^
Errors: 1
```

#### Expected behavior

Better error message pointing to the fact that the unreachable code is fake code synthesized as a part of the way Sorbet models Ruby's conditional send mechanism ("safe navigation" / `&.`).

- - -

Sorbet treats this construct as if the user wrote:

```ruby
# typed: strict
extend T::Sig

sig {params(x: Integer).void}
def foo(x)
  magic_tmp_var = x
  if NilClass === magic_tmp_var
    nil
  else
    magic_tmp_var.to_s
  end
end
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams(x%3A%20Integer).void%7D%0Adef%20foo(x)%0A%20%20magic_tmp_var%20%3D%20x%0A%20%20if%20NilClass%20%3D%3D%3D%20magic_tmp_var%0A%20%20%20%20nil%0A%20%20else%0A%20%20%20%20magic_tmp_var.to_s%0A%20%20end%0Aend)

Which shows that in fact the error message is coming from the synthetic `nil` in the `if` condition:

```
editor.rb:8: This code is unreachable https://srb.help/7006
     8 |    nil
            ^^^
Errors: 1
```

One idea here might be: when we're creating the `if ... nil` branch, synthesize a call to something new like

```
if NilClass === magic_tmp_var
  <Magic>.<nilForCsend>(magic_tmp_var)
else
  # ...
end
```

[...]

Issue #621, opened 2019-05-24
Missing a dead code error that we should be printing


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aa%20%3D%201%0Aif%20a%0A%20%20'reachable'%0Aelse%0A%20%20'unreachable'%0Aend)

```ruby
# typed: true
a = 1
if a
  'reachable'
else
  'unreachable'
end
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

The `else` branch is actually unreachable, because only `nil` and `false` are falsy.

- - -

This happens because Sorbet's CFG simplifier simplifies the CFG before we do dead code analysis. If make slight modifications to the tree to prevent the simplification from happening, the error is correctly reported:

```ruby
# typed: true
a = 1
x = if a
  'reachable'
else
  'unreachable'
end
```
[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aa%20%3D%201%0Ax%20%3D%20if%20a%0A%20%20'reachable'%0Aelse%0A%20%20'unreachable'%0Aend)

```ruby
# typed: true
a = 1
if a
  'reachable'
else
  puts a
  'unreachable'
end
```
[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aa%20%3D%201%0Aif%20a%0A%20%20'reachable'%0Aelse%0A%20%20puts%20a%0A%20%20'unreachable'%0Aend)

PR #3620, opened 2020-11-06
Better error message for &. on non-nilable receiver

aprocter-stripe:
This adds a less confusing error message for cases where `&.` ("safe navigation") is used on a receiver that is not nilable. (Currently this shows up as a "code is unreachable" error inside the if-then-else generated by the desugarer.)

Implementation is as @DarkDimius outlines in #36, with the very slight difference that the "magic send" winds up being the second instruction in the block, not the first. (First instruction is an `alias` for the `<Magic>` class constant.)

Example:

```ruby
# typed: true

extend T::Sig

if T.unsafe(false)
  foo = "hi"
else
  foo = 33
end

bar = foo&.to_s
```

Before:

```
    11 |bar = foo&.to_s
              ^
Errors: 1
```

After:

```
editor.rb:11: Used &. operator on a receiver which can never be nil https://srb.help/7034
    11 |bar = foo&.to_s
              ^
  Type of receiver is String | Integer, from:
    editor.rb:6:
     6 |  foo = "hi"
                ^^^^
    editor.rb:8:
     8 |  foo = 33
                ^^
Errors: 1
```

### Motivation

The weird error message in this case is a pretty common source of confusion among users.

### Test plan

See included automated tests.

### DIFF
--- a/core/SymbolRef.h
+++ b/core/SymbolRef.h
@@ -560,7 +560,7 @@ public:
     }
 
     static constexpr int MAX_SYNTHETIC_CLASS_SYMBOLS = 200;
-    static constexpr int MAX_SYNTHETIC_METHOD_SYMBOLS = 30;
+    static constexpr int MAX_SYNTHETIC_METHOD_SYMBOLS = 31;
     static constexpr int MAX_SYNTHETIC_FIELD_SYMBOLS = 1;
     static constexpr int MAX_SYNTHETIC_TYPEARGUMENT_SYMBOLS = 2;
     static constexpr int MAX_SYNTHETIC_TYPEMEMBER_SYMBOLS = 99;
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2130,6 +2130,14 @@ public:
     }
 } Magic_selfNew;
 
+class Magic_nilForSafeNavigation : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, DispatchArgs args, DispatchResult &res) const override {
+        ENFORCE(args.args.size() == 1);
+        res.returnType = Types::nilClass();
+    }
+} Magic_nilForSafeNavigation;
+
 class DeclBuilderForProcs_void : public IntrinsicMethod {
 public:
     void apply(const GlobalState &gs, DispatchArgs args, DispatchResult &res) const override {
@@ -2583,6 +2591,7 @@ const vector<Intrinsic> intrinsicMethods{
     {Symbols::Magic(), Intrinsic::Kind::Singleton, Names::callWithSplatAndBlock(), &Magic_callWithSplatAndBlock},
     {Symbols::Magic(), Intrinsic::Kind::Singleton, Names::suggestType(), &Magic_suggestUntypedConstantType},
     {Symbols::Magic(), Intrinsic::Kind::Singleton, Names::selfNew(), &Magic_selfNew},
+    {Symbols::Magic(), Intrinsic::Kind::Singleton, Names::nilForSafeNavigation(), &Magic_nilForSafeNavigation},
 
     {Symbols::DeclBuilderForProcsSingleton(), Intrinsic::Kind::Instance, Names::void_(), &DeclBuilderForProcs_void},
     {Symbols::DeclBuilderForProcsSingleton(), Intrinsic::Kind::Instance, Names::returns(),
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2130,14 +2130,6 @@ public:
     }
 } Magic_selfNew;
 
-class Magic_nilForSafeNavigation : public IntrinsicMethod {
-public:
-    void apply(const GlobalState &gs, DispatchArgs args, DispatchResult &res) const override {
-        ENFORCE(args.args.size() == 1);
-        res.returnType = Types::nilClass();
-    }
-} Magic_nilForSafeNavigation;
-
 class DeclBuilderForProcs_void : public IntrinsicMethod {
 public:
     void apply(const GlobalState &gs, DispatchArgs args, DispatchResult &res) const override {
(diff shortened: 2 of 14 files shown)


core/SymbolRef.h
```diff
@@ -560,7 +560,7 @@ class Symbols {
     }
 
     static constexpr int MAX_SYNTHETIC_CLASS_SYMBOLS = 200;
-    static constexpr int MAX_SYNTHETIC_METHOD_SYMBOLS = 30;
+    static constexpr int MAX_SYNTHETIC_METHOD_SYMBOLS = 31;
```
jez, line 563: I don't remember seeing this need to be bumped frequently. What was the problem you ran into that caused this to be bumped?

core/types/calls.cc
```diff
@@ -2132,6 +2132,14 @@ class Magic_selfNew : public IntrinsicMethod {
     }
 } Magic_selfNew;
 
+class Magic_nilForSafeNavigation : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, DispatchArgs args, DispatchResult &res) const override {
+        ENFORCE(args.args.size() == 1);
+        res.returnType = Types::nilClass();
+    }
+} Magic_nilForSafeNavigation;
```
jez, line 2141: I don't think we need this, because you have already set the symbol's `resultType` in GlobalState.cc.

test/testdata/desugar/blockpass.rb
```diff
@@ -45,7 +45,6 @@ def foo(&blk)
     calls_with_object(&HasToProc.new)
     calls_with_object {|*args| HasToProc.new.to_proc.call(*args)}
     CallsWithObject.calls_with_object(&:meth)
-    CallsWithObject&.calls_with_object(&:meth)
```
jez, line 48: Why did we delete this line?

website/docs/error-reference.md
```diff
@@ -986,3 +986,23 @@ See [Exhaustiveness Checking](exhaustiveness.md) for more information.
 [report an issue]: https://github.com/sorbet/sorbet/issues
 
 <script src="/js/error-reference.js"></script>
+
+## 7034
+
+Sorbet detected that the safe navigation operator `&.` was being used on a
+receiver that can never be nil. Replace the offending occurrence of `&.` with
+`.`.
```
jez, line 994:
```suggestion
Sorbet detected that the safe navigation operator (`&.`) was being used on a
receiver that can never be nil. Replace the offending occurrence of `&.` with
a normal method call (`.`).
```

infer/inference.cc
```diff
@@ -128,17 +128,61 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
...
                         continue;
                     }
                     if (cfg::isa_instruction<cfg::TAbsurd>(expr.value.get())) {
                         continue;
                     }
-                    if (auto e = ctx.beginError(expr.loc, core::errors::Infer::DeadBranchInferencer)) {
+
+                    auto send = cfg::cast_instruction<cfg::Send>(expr.value.get());
+                    if (send != nullptr && send->fun == core::Names::nilForSafeNavigation()) {
+                        unreachableInstruction = expr.value.get();
```
jez, line 156: Why not emit the error right here and break? Why have a separate variable that we use to emit the error later?

infer/inference.cc
```diff
@@ -128,17 +128,61 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
...
+                                ctx.beginError(locForUnreachable, core::errors::Infer::UnnecessarySafeNavigation)) {
+                            e.setHeader("Used `{}` operator on a receiver which can never be nil", "&.");
+
+                            // Just a failsafe check; args.size() should always be 1.
+                            if (send->args.size() > 0) {
+                                auto ty = current.getAndFillTypeAndOrigin(ctx, send->args[0]);
+                                e.addErrorSection(
+                                    core::ErrorSection(core::ErrorColors::format("Type of receiver is `{}`, from:",
+                                                                                 ty.type->toString(ctx)),
+                                                       ty.origins2Explanations(ctx, current.ownerLoc)));
```
jez, line 180: I don't like that there's no obvious connection here that `ownerLoc` is being passed along because it happens to be the loc we used to represent uninitialized values.

Instead, can we make a public helper function on Environment that uses `Environment::uninitialized` to project out the uninitialized loc?

core/SymbolRef.h
aprocter-stripe: This check winds up failing, since we now register `<Magic>.<nil-for-safe-navigation>`:

https://github.com/sorbet/sorbet/blob/20fad5888ade5aa7a2605c3a1aba5ddf90b3bd90/core/GlobalState.cc#L734-L736

core/types/calls.cc
aprocter-stripe: Yup, everything seems to work without it! Will remove.

test/testdata/desugar/blockpass.rb
aprocter-stripe: Uh, good question. It was raising the new error on this line because `CallsWithObject` cannot be nil, but that raises the question of why it wasn't raising "code is unreachable" error in the first place (or maybe it was, but somehow we had marked it as expected). Will look into that.

core/SymbolRef.h
jez: gotcha

test/testdata/desugar/blockpass.rb
jez: In any case, we should almost never delete code to make tests pass when new errors are introduced, instead of adding an additional `# error:` comment

(i.e., this implicitly was the only place in Sorbet's test suite where we were testing that Sorbet doesn't crash when you use `&.` on a constant literal receiver! so it's definitely code we want to continue to test, and it's fine if the behavior has changed.)

infer/inference.cc
aprocter-stripe: If we find the send to `nilForSafeNavigation` we could go ahead and emit the error and break. The tricky thing is that for other instructions we can't emit the ordinary "code is unreachable" error right away, because the magic send unfortunately does not wind up occurring first. The guts of the dead block wind up looking like this, with the magic send in the second position:

```
<cfgAlias>$28: T.class_of(<Magic>) = alias <C <Magic>>
<statTemp>$21: NilClass = <cfgAlias>$28: T.class_of(<Magic>).<nil-for-safe-navigation>(y: NilClass)
<unconditional>
```

When you put it all together, doing it this way ended up feeling a little cleaner to me.

jez (approved):
I left my comments. Feel free to request another review if you have followup comments.

infer/inference.cc
jez: Gotcha. I guess i was just surprised to see how much code this required but it seems to get the job done nonetheless so 👍

aprocter-stripe: > surprised to see how much code this required

same tbh  ;)

test/testdata/desugar/blockpass.rb
aprocter-stripe: (Pulling in some Slack discussion) Looks like the new desugaring for `&.` no longer triggers the bug in https://github.com/sorbet/sorbet/issues/621, which is the only reason this line worked in the first place.

I'll restore this line with a suitable `# error:`, and will watch out for any new errors when testing against Stripe code.

aprocter-stripe:

→ https://go/builds/bui_ILJVnmh6v450Th

PR #3621, opened 2020-11-06
Avoid clobbering `this` in `recordSealedSubclass`

elliottt:
`Symbol::singletonClass` can cause a reallocation of the symbol table, which will invalidate the `this` pointer in `recordSealedSubclass`. This change avoids the problem by indirecting via a `SymbolRef` in `recordSealedSubclass` instead.

### Motivation

Fix an intermittent crash.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -960,13 +960,16 @@ void Symbol::recordSealedSubclass(MutableContext ctx, SymbolRef subclass) {
     ENFORCE(subclass.exists(), "Can't record sealed subclass for {} when subclass doesn't exist", this->show(ctx));
     ENFORCE(subclass.data(ctx)->isClassOrModule(), "Sealed subclass {} must be class", subclass.show(ctx));
 
+    auto selfRef = this->ref(ctx);
+
     // We record sealed subclasses on a magical method called core::Names::sealedSubclasses(). This is so we don't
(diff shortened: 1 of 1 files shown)


jez (approved):

core/Symbols.cc
```diff
@@ -960,13 +960,16 @@ void Symbol::recordSealedSubclass(MutableContext ctx, SymbolRef subclass) {
     ENFORCE(subclass.exists(), "Can't record sealed subclass for {} when subclass doesn't exist", this->show(ctx));
     ENFORCE(subclass.data(ctx)->isClassOrModule(), "Sealed subclass {} must be class", subclass.show(ctx));
 
+    auto selfRef = this->ref(ctx);
```
jez, line 963: Can you comment this? We have in the past looked at completely removing `ref`, and this is one that someone could easily come along not knowing the reasoning and inline this into the one usage below.

trevor-stripe: Should we comment all the uses of `ref` in this file? It seems like most methods end up using `this->ref(gs)` to avoid the same bug, perhaps this is why @DarkDimius was suggesting we always pass in the `SymbolRef` to these methods?

jez: Either way works for me.

PR #3625, opened 2020-11-06
Method rename support (initial implementation)

soam-stripe:
Implements a first cut of method rename support, building on the constant rename work. This implementation supports subclasses but doesn't handle union types, mixins, and probably more cases.

The goal of this change is to enable method renames in experimental mode and gather some user feedback about which missing bits are most important.

### Motivation

We have lots of user demand for refactoring features, and we think method renames would one of the useful ones.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/rename.cc
+++ b/main/lsp/requests/rename.cc
@@ -7,7 +7,6 @@
 #include "main/lsp/ShowOperation.h"
 #include "main/lsp/json_types.h"
 #include "main/lsp/lsp.h"
-#include <iostream>
 #include <stdio.h>
 using namespace std;
 
@@ -126,7 +125,6 @@ public:
         } else { // send
             newsrc = replaceMethodNameInSend(source);
         }
-        cout << "MethodRename: old/new: " << source << "/" << newsrc << "\n";
         edits[location->uri].push_back(make_unique<TextEdit>(move(location->range), newsrc));
     }
 
--- a/main/lsp/requests/rename.cc
+++ b/main/lsp/requests/rename.cc
@@ -38,37 +38,37 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
 
 // Replaces originalName with newName in the last dotted component if the string is of the form x.y.z; otherwise does a
 // simple ReplaceAll.
-std::string replaceLastDotted(std::string_view input, std::string_view originalName, std::string_view newName) {
-    if (absl::StrContains(input, ".")) {
-        std::vector<std::string> dotted = absl::StrSplit(input, ".");
-        // TODO: check for exactly one match
-        dotted[dotted.size() - 1] = absl::StrReplaceAll(dotted[dotted.size() - 1], {{originalName, newName}});
-        return absl::StrJoin(dotted, ".");
-    }
-    return absl::StrReplaceAll(input, {{originalName, newName}});
+string replaceLastDotted(string_view input, string_view originalName, string_view newName) {
+    vector<string> dotted = absl::StrSplit(input, ".");
+    // TODO: check for exactly one match
+    dotted[dotted.size() - 1] = absl::StrReplaceAll(dotted[dotted.size() - 1], {{originalName, newName}});
+    return absl::StrJoin(dotted, ".");
 }
 
 // Checks if s is a subclass of root, and updates isSubclass, visited, and subclasses vectors.
 bool updateSubclass(const core::GlobalState &gs, core::SymbolRef root, core::SymbolRef s, vector<bool> &isSubclass,
                     vector<bool> &visited, vector<core::SymbolRef> &subclasses) {
-    if (visited[s.rawId()] == true) {
-        return isSubclass[s.rawId()];
+    // don't visit the same class twice
+    if (visited[s.classOrModuleIndex()] == true) {
+        return isSubclass[s.classOrModuleIndex()];
     }
-    visited[s.rawId()] = true;
-    if (s.rawId() == root.rawId()) {
-        subclasses.push_back(s);
-        isSubclass[s.rawId()] = true;
+    visited[s.classOrModuleIndex()] = true;
+
+    // s is a subclass of root if it's equal to root, or if its parent is a subclass of root
+    if (s == root) {
+        subclasses.emplace_back(s);
+        isSubclass[s.classOrModuleIndex()] = true;
         return true;
     }
     auto super = s.data(gs)->superClass();
     if (super.exists()) {
         if (updateSubclass(gs, root, super, isSubclass, visited, subclasses)) {
-            subclasses.push_back(s);
-            isSubclass[super.rawId()] = true;
+            subclasses.emplace_back(s);
+            isSubclass[super.classOrModuleIndex()] = true;
             return true;
         }
     }
-    isSubclass[s.rawId()] = false;
+    isSubclass[s.classOrModuleIndex()] = false;
     return false;
 }
 
@@ -85,14 +85,15 @@ vector<core::SymbolRef> getSubclasses(LSPTypecheckerDelegate &typechecker, core:
     return subclasses;
 }
 
-// Follow superClass links to find the root of a class hierarchy.
-core::SymbolRef findRootSuperclass(const core::GlobalState &gs, core::SymbolRef klass) {
+// Follow superClass links until we find the highest class that contains the given method. In other words we find the
+// "root" of the tree of classes that define a method.
+core::SymbolRef findRootClassWithMethod(const core::GlobalState &gs, core::SymbolRef klass, core::NameRef methodName) {
     auto root = klass;
+    ENFORCE(klass.data(gs)->isClassOrModule());
     while (true) {
         auto tmp = root.data(gs)->superClass();
-        ENFORCE(tmp.exists()); // everything derives from Kernel::Object
-        // TODO is there a better way to do this check for the base object class?
-        if (!tmp.exists() || tmp.data(gs)->show(gs) == "Object") {
+        ENFORCE(tmp.exists()); // everything derives from Kernel::Object so we can't ever reach the actual top type
+        if (!tmp.exists() || !(tmp.data(gs)->findMember(gs, methodName).exists())) {
             break;
         }
         root = tmp;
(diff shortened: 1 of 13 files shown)


jvilk-stripe (changes requested):
This is looking good! I had one big correctness note and one big maintainability note. The rest are small nits.

I haven't reviewed the test coverage; I'll do that once the other comments are addressed.

main/lsp/requests/rename.cc
```diff
@@ -29,37 +30,135 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
...
-    UnorderedMap<string, vector<unique_ptr<TextEdit>>> edits;
-    for (auto &location : references) {
-        // Get text at location.
-        auto fref = config.uri2FileRef(gs, location->uri);
-        if (fref.data(gs).isPayload()) {
-            // We don't support renaming things in payload files.
-            return JSONNullObject();
+    vector<core::SymbolRef> symbolsToRename;
+    if (symbolData->isMethod()) {
+        // We have to check for methods as part of a class hierarchy: Follow superClass() links till we find the root;
```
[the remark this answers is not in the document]
soam-stripe: Excellent point! Will fix.

main/lsp/requests/rename.cc
```diff
@@ -29,37 +30,135 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
...
+    visited[s.rawId()] = true;
+    if (s.rawId() == root.rawId()) {
+        subclasses.push_back(s);
+        isSubclass[s.rawId()] = true;
+        return true;
+    }
+    auto super = s.data(gs)->superClass();
+    if (super.exists()) {
+        if (updateSubclass(gs, root, super, isSubclass, visited, subclasses)) {
+            subclasses.push_back(s);
```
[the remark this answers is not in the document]
soam-stripe: I don't see how it duplicates results? The function can't ever do line 59 and this line for the same class because it tracks `visited`.  If we move out the vector from this function we'll add a factor of `class hierarchy length` to the overall time complexity, I think? Because we'll have to walk up the path from class to root for every class we visit, e.g. if we have `A < B < C`, we'll have to walk the B->C link twice if we happen to visit A before B in the outer loop.

soam-stripe:
Okay I've addressed all the comments except the ones about updateSubclass. 

ptal @jvilk-stripe

main/lsp/requests/rename.cc
```diff
@@ -29,37 +30,135 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
...
         }
     }
     return true;
 }
+
+// Replaces originalName with newName in the last dotted component if the string is of the form x.y.z; otherwise does a
+// simple ReplaceAll.
+std::string replaceLastDotted(std::string_view input, std::string_view originalName, std::string_view newName) {
+    if (absl::StrContains(input, ".")) {
+        std::vector<std::string> dotted = absl::StrSplit(input, ".");
```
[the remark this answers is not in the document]
soam-stripe: (I think I can't use auto here because there are a few return-type overloads of absl::StrSplit)

soam-stripe:
Thanks for the review @jvilk-stripe, the `isSubclass` function is updated too.

main/lsp/requests/rename.cc
```diff
@@ -114,15 +241,24 @@ unique_ptr<ResponseMessage> RenameTask::runRequest(LSPTypecheckerDelegate &typec
             auto resp = move(queryResponses[0]);
             // Only supports rename requests from constants and class definitions.
             if (auto constResp = resp->isConstant()) {
+                // Sanity check the text.
+                if (islower(params->newName[0])) {
```
jvilk-stripe, line 245: This check is also going to be needed if `resp->isDefinition()` and the symbol `isClassOrModule()`.

This reminds me that _method_ renames have to be lower case. You can choose if you want to add that check now or later. :)

test/testdata/lsp/rename/method_attr.rb
```diff
@@ -0,0 +1,10 @@
+# typed: true
+# frozen_string_literal: true
+
+class Foo
+  attr_reader :bar
+end
+
+f = Foo.new
+f.bar
+# ^ apply-rename: [A] newName: baz
```
jvilk-stripe, line 10: Note that this won't completely work since `attr_reader :bar` also creates a `@bar` in the class, which the class assigns to to populate the reader. But that's an acceptable limitation for now; we'd need the ability 1) to recognize it's an `attr_reader` and 2) to rename the field.

test/testdata/lsp/rename/method_class_hierarchy.rb
```diff
@@ -0,0 +1,30 @@
...
+end
+
+class B < BaseWithMethod
+  def foo
+  end
+end
+
+class C < Base
+  def foo
+    # this one should not be renamed!
```
jvilk-stripe, line 26: Nice

jvilk-stripe (approved):
The code looks good; I had a few small suggestions.

You might want to add a test for renaming class methods, e.g.:

```ruby
class Foo
  def self.foo
    #       ^ renaming this
  end

  def foo
    # this is a separate method
  end
end

Foo.foo # this is a callsite of self.foo 

Foo.new.foo.# this is a callsite of foo
```

I think it should 'just work'.

I'd also recommend testing that we _reject_ renames on call sites of `new` (e.g. `Foo.new`) and on `initialize` (which `Foo.new` calls).

main/lsp/requests/rename.cc
```diff
@@ -29,48 +30,187 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
...
-    const core::GlobalState &gs = typechecker.state();
-    vector<unique_ptr<Location>> references = getReferencesToSymbol(typechecker, symbol);
-    auto originalName = symbol.data(gs)->name.toString(gs);
-    auto we = make_unique<WorkspaceEdit>();
+// Replaces originalName with newName in the last dotted component if the string is of the form x.y.z; otherwise does a
+// simple ReplaceAll.
+string replaceLastDotted(string_view input, string_view originalName, string_view newName) {
+    vector<string> dotted = absl::StrSplit(input, ".");
+    // TODO: check for exactly one match
+    dotted[dotted.size() - 1] = absl::StrReplaceAll(dotted[dotted.size() - 1], {{originalName, newName}});
```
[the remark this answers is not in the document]
soam-stripe: Your question made me realize this code is just wrong -- in case of a send the loc contains the full method args, so this "split by dots" thing is too simple. It needs to properly parse the function name out of an arbitrary method call. I'll update with a fix.

test/testdata/lsp/rename/method_attr.rb
soam-stripe: Yeah... good point. Do you think we should explicitly check for this and reject the rename? Or is it enough just to document this as a current limitation?

soam-stripe:
Thanks again! Addressed the comments. I'm still not a 100% sure the method parsing stuff is really correct, let me know if you see anything I missed?

PTAL @jvilk-stripe

test/testdata/lsp/rename/method_attr.rb
jvilk-stripe: I'm OK with it as a current limitation.

jvilk-stripe: Updating the field references is much easier than the method references, as the field can only be referenced within the class methods.

main/lsp/requests/rename.cc
```diff
@@ -29,47 +29,208 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
...
+class MethodRenamer : public Renamer {
+public:
+    MethodRenamer(const string oldName, const string newName) : Renamer(oldName, newName) {}
+    ~MethodRenamer() {}
+    void rename(const core::GlobalState &gs, unique_ptr<Location> &location, core::FileRef fref) override {
+        auto loc = location->range->toLoc(gs, fref);
+        auto source = loc.source(gs);
+        string newsrc;
+        if (absl::StartsWith(source, "def ") || absl::StartsWith(source, "attr_reader") ||
+            absl::StartsWith(source, "attr_accessor")) {
```
jvilk-stripe, line 123: `attr_writer` is another.

We can remove this hardcoded logic, btw (maybe in the next PR). `getReferencesToSymbol` actually knows the provenance of these locations from the `QueryResponse` objects it processes and if they are from sends or definitions:

https://github.com/sorbet/sorbet/blob/e932d6f8/main/lsp/LSPTask.cc#L226

Maybe we need to introduce a new method (maybe that `getReferencesToSymbol` calls?) that returns a vector of relevant `QueryResponse` objects.

main/lsp/requests/rename.cc
```diff
@@ -29,47 +29,208 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
...
+private:
+    // Parse a method name from a send (call), replace it with the new name, and return the full send source with the
+    // new name
+    string replaceMethodNameInSend(string send) {
+        // Split into method and args
+        string methodExpr, args;
+        for (string::size_type i = 0; i < send.size(); i++) {
+            if (send[i] == ' ' || send[i] == '(') {
+                methodExpr = send.substr(0, i);
+                args = send.substr(i, send.size() - i);
```
jvilk-stripe, line 140: Should this break out of the loop?

Also, does this work on callsites with a receiver? e.g.:

```ruby
  foo.bar(a, b, c)
#  ^^^ this is not the method
#.    ^^^ this is the method
```

is `foo` part of the send loc? It _looks like_ the tests pass, so maybe this isn't a problem.

jvilk-stripe: Nit: Actually, maybe we could replace this with something simpler, like `send.find(methodName)` that verifies that the next token in `send` (if present) isn't alphanumeric? Up to you if you want to do it that way.

main/lsp/requests/rename.cc
```diff
@@ -114,15 +268,37 @@ unique_ptr<ResponseMessage> RenameTask::runRequest(LSPTypecheckerDelegate &typec
...
             } else if (auto defResp = resp->isDefinition()) {
-                if (defResp->symbol.data(gs)->isClassOrModule()) {
+                if (defResp->symbol.data(gs)->isClassOrModule() && islower(params->newName[0])) {
+                    response->error =
+                        make_unique<ResponseError>((int)LSPErrorCodes::InvalidRequest,
+                                                   "Class and Module names must begin with an uppercase letter.");
+                    return response;
+                }
+
+                if (defResp->symbol.data(gs)->isMethod() && isupper(params->newName[0])) {
```
jvilk-stripe, line 288: This is missing in `isSend()`.

Maybe we should move new name sanity checks to the renamer or something.

jvilk-stripe: You can test this error in the test suite, btw.

jvilk-stripe (approved):
Looking good! I had a few minor nits.

main/lsp/requests/rename.cc
soam-stripe: Yes it should break from the loop, will fix that. I wanted to avoid searching for methodName because then we have to handle cases where it appears in the string but isn't actually the method we're looking for (like in the method_names test case)

jvilk-stripe: oh yeah! d'oh. makes sense :)

soam-stripe: And yes, `foo` is part of the send loc for `foo.bar(a,b,c)`. But we only break from the loop on `(` or space, so we get `foo.bar` in methodExpr

jvilk-stripe: After discussing in DM about other corner cases, we will fix this issue in a subsequent PR by surfacing receiver loc information in `SendQueryResponse`s. That will make it easy to exclude the part of the send loc that corresponds to the receiving object.

PR #3627, opened 2020-11-07
Support go-to-def / go-to-type-def / hover doc strings for classes/modules opened in multiple files

jvilk-stripe:
Support go-to-def / go-to-type-def / hover doc strings for classes/modules opened in multiple files.

### Motivation

These features previously only reported/used a single definition site.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/definitions_multiple__a.rb
+++ b/test/testdata/lsp/definitions_multiple__a.rb
@@ -1,12 +1,17 @@
 # typed: true
 
+# Documentation 2
 class MyCommand
   #   ^^^^^^^^^ def: MyCommand
   #   ^^^^^^^^^ type-def: MyCommand
+  #   ^^^^^^^^^ hover: Documentation 1
+  #   ^^^^^^^^^ hover: Documentation 2
 end
 
 def main
   MyCommand.new
 # ^^^^^^^^^ usage: MyCommand
 #           ^^^ type: MyCommand
+# ^^^^^^^^^ hover: Documentation 1
+# ^^^^^^^^^ hover: Documentation 2
 end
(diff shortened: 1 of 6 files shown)


test/testdata/lsp/definitions_multiple__a.rb
```diff
@@ -0,0 +1,17 @@
...
+  #   ^^^^^^^^^ hover: Documentation 1
+  #   ^^^^^^^^^ hover: Documentation 2
+end
+
+def main
+  MyCommand.new
+# ^^^^^^^^^ usage: MyCommand
+#           ^^^ type: MyCommand
+# ^^^^^^^^^ hover: Documentation 1
+# ^^^^^^^^^ hover: Documentation 2
```
jez, line 16: wow i hope no one is using this feature

test/testdata/lsp/hover.rb
```diff
@@ -5,8 +5,8 @@ class BigFoo; extend T::Sig
 
 # The docs for FOO_CONSTANT
   FOO_CONSTANT = 1
-# ^^^^^^^^^^^^ The docs for FOO_CONSTANT
-# ^^^^^^^^^^^^ Integer(1)
+# ^^^^^^^^^^^^ hover: The docs for FOO_CONSTANT
+# ^^^^^^^^^^^^ hover: Integer(1)
```
jez, line 9: 🙈

jez (approved):
i wonder if people will complain that we might jump to an RBI def or something but maybe this is expected

test/testdata/lsp/definitions_multiple__a.rb
jvilk-stripe: Primarily, this will fix the issue where the documentation is on an 'unlucky' loc.

jvilk-stripe:
>i wonder if people will complain that we might jump to an RBI def or something but maybe this is expected

tbh, I expect this behavior. We will see if our abuse of RBI files internally will lead to complaints. :)

PR #3631, opened 2020-11-10
Move ResolveMixesInClassMethods AST pass to ResolveConstants.

jvilk-stripe:
Move ResolveMixesInClassMethods AST pass to ResolveConstants.

Converts a serial AST walk into a parallel AST walk that collects work to perform serially, which fits the ResolveConstants pattern.

* Question: Does this have any impact on autogen? Autogen previously did not run `resolveMixesInClassMethods`. It looks like it's working fine, though.
* Answer: From chatting w/ team, autogen doesn't consume method information, so it should not impact it. The additional runtime to `resolveConstants` with this new change is negligible; the new timer for `resolver.mixes_in_class_methods` doesn't show up with --counters since it completes within a handful of milliseconds!

### Motivation

Speed!

### Test plan

Covered by existing tests.

### DIFF
--- a/docs/namer-resolver-pipeline.md
+++ b/docs/namer-resolver-pipeline.md
@@ -87,12 +87,6 @@ For a fuller picture of Sorbet's pipeline, see [pipeline.md](pipeline.md).
 │                                                   │ ast::Expression │                                                   │
 │                                                   └─────────────────┘                                                   │
 │                                                            │                                                            │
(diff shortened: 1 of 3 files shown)


jvilk-stripe:

→ https://go/builds/bui_IMVRRfIWOhYj64

elliottt (approved):
Nice! Looking forward to the next set of PRs 👍

PR #3632, opened 2020-11-11
Compute external types during incremental resolver.

jvilk-stripe:
Compute external types during incremental resolver.

May resolve an LSP crash in features that use externalType() to pretty print information to user.

I was unable to reproduce the crash in a test, unfortunately. :( In order to do so, I'd need to induce the following conditions:

* incrementalResolver creates a new class or module symbol
* incrementalResolver does not compute the external type via `unsafeComputeExternalType()`
* LSP references the class in a constant reference which the test hovers over

### Motivation

Users were crashing at this line of code, which suggests that `externalType()` returned nullptr:
https://github.com/sorbet/sorbet/blob/master/main/lsp/lsp_helpers.cc#L240

I know there are some cases where the fast path / incremental resolver may end up adding new symbols to the symbol table, so my guess is that the user found themselves in a state where LSP tried to view the external type of a symbol created on the incremental path.

### Test plan

I could not reproduce the crash in a test. Ideas appreciated!

### DIFF
--- a/resolver/resolver.h
+++ b/resolver/resolver.h
@@ -26,7 +26,6 @@ private:
     static void finalizeAncestors(core::GlobalState &gs);
     static void finalizeSymbols(core::GlobalState &gs);
     static void computeLinearization(core::GlobalState &gs);
-    static void computeExternalTypes(core::GlobalState &gs);
     static ast::ParsedFilesOrCancelled resolveSigs(core::GlobalState &gs, std::vector<ast::ParsedFile> trees);
     static void sanityCheck(core::GlobalState &gs, std::vector<ast::ParsedFile> &trees);
 };
--- a/resolver/resolver.h
+++ b/resolver/resolver.h
@@ -26,6 +26,7 @@ private:
     static void finalizeAncestors(core::GlobalState &gs);
     static void finalizeSymbols(core::GlobalState &gs);
     static void computeLinearization(core::GlobalState &gs);
+    static void computeExternalTypes(core::GlobalState &gs);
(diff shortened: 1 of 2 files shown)


resolver/resolver.h
```diff
@@ -26,6 +26,7 @@ class Resolver final {
     static void finalizeAncestors(core::GlobalState &gs);
     static void finalizeSymbols(core::GlobalState &gs);
     static void computeLinearization(core::GlobalState &gs);
+    static void computeExternalTypes(core::GlobalState &gs);
```
elliottt, line 29: Does this need to be exposed, or can it live in an anonymous namespace in resolver.cc?

elliottt (approved):
Looks good! My only question is about whether or not `computeExternalType` needs to be exposed in the header, but it's not blocking.

resolver/resolver.h
jvilk-stripe: Why would it need to be exposed? It's only called within `resolver.cc`. I added it here as that's where `finalizeAncestors` and friends are defined.

I actually don't know why some of the private passes are exposed in `resolver.h`.

elliottt: Well, maybe we can address that in a follow-up at some point :)

jvilk-stripe: Oh okay wait I realize what you're saying. Yeah I can make it anonymous.

jvilk-stripe: I think they are in `resolver.h` bc they are defined in separate files.

PR #3636, opened 2020-11-11
Implements rewriter for `Forwardable#def_delegator`

jeffcarbs:
Implements `Forwardable#def_delegator` by desugaring things of the form

```ruby
class MyClass
  def_delegator :thing, :foo
  def_delegator :thing, :bar, :aliased_bar
end
```

into

```ruby
class MyClass
  sig {params(arg0: T.untyped, blk: T.nilable(Proc)).returns(T.untyped)}
  def foo(*arg0, &blk); end

  sig {params(arg0: T.untyped, blk: T.nilable(Proc)).returns(T.untyped)}
  def aliased_bar(*arg0, &blk); end
end
```

Some notes:
- This is similar to the rewriter that already exists for ActiveSupport's `Module#delegate`.
- There's also a `Forwardable#def_delegators` (plural) that this PR does not handle since the usage is different enough that it likely warrants its own rewriter.

### Motivation

Enable usage of `Forwardable#def_delegator` (a ruby stdlib construct) with sorbet:

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20Forwardable%0A%0A%20%20def_delegator%20%3Athing%2C%20%3Afoo%0Aend%0A%0AA.new.foo%20%23%20Method%20foo%20does%20not%20exist%20on%20A)
```ruby
class A
  extend Forwardable

  def_delegator :thing, :foo
end

A.new.foo # error: Method foo does not exist on A (but the method does exist!)
```

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/DefDelegator.cc
+++ b/rewriter/DefDelegator.cc
@@ -9,47 +9,43 @@ using namespace std;
 namespace sorbet::rewriter {
 
 vector<ast::TreePtr> DefDelegator::run(core::MutableContext ctx, const ast::Send *send) {
-    vector<ast::TreePtr> empty;
+    vector<ast::TreePtr> methodStubs;
     auto loc = send->loc;
 
     if (send->fun != core::Names::defDelegator()) {
-        return empty;
+        return methodStubs;
     }
 
     // This method takes 2..3 positional arguments and no keyword args:
     // `def_delegator(accessor, method, ali = method)`
     if (!(send->numPosArgs == 2 || send->numPosArgs == 3)) {
-        return empty;
+        return methodStubs;
     }
 
     if (send->hasKwArgs()) {
-        return empty;
+        return methodStubs;
     }
 
     auto *accessor = ast::cast_tree<ast::Literal>(send->args[0]);
-    auto *method = ast::cast_tree<ast::Literal>(send->args[1]);
-
     if (!accessor || !(accessor->isSymbol(ctx) || accessor->isString(ctx))) {
-        return empty;
+        return methodStubs;
     }
 
+    auto *method = ast::cast_tree<ast::Literal>(send->args[1]);
     if (!method || !method->isSymbol(ctx)) {
-        return empty;
+        return methodStubs;
     }
 
     core::NameRef methodName = method->asSymbol(ctx);
 
     if (send->numPosArgs == 3) {
         auto *alias = ast::cast_tree<ast::Literal>(send->args[2]);
-
         if (!alias || !alias->isSymbol(ctx)) {
-            return empty;
+            return methodStubs;
         }
 
         methodName = alias->asSymbol(ctx);
     }
-
-    vector<ast::TreePtr> methodStubs;
 
     // sig {params(arg0: T.untyped, blk: Proc).returns(T.untyped)}
     ast::Send::ARGS_store sigArgs;
@@ -66,7 +62,12 @@ vector<ast::TreePtr> DefDelegator::run(core::MutableContext ctx, const ast::Send
     args.emplace_back(ast::MK::RestArg(loc, ast::MK::Local(loc, core::Names::arg0())));
     args.emplace_back(ast::make_tree<ast::BlockArg>(loc, ast::MK::Local(loc, core::Names::blkArg())));
 
-    methodStubs.push_back(ast::MK::SyntheticMethod(loc, loc, methodName, std::move(args), ast::MK::EmptyTree()));
+    methodStubs.push_back(
+        ast::MK::SyntheticMethod(loc, loc, methodName, std::move(args), ast::MK::RaiseUnimplemented(loc)));
+
+    // Include the original call to def_delegator so sorbet will still type-check it
+    // and throw errors if the class (or its parent) didn't `extend Forwardable`
+    methodStubs.push_back(send->deepCopy());
 
     return methodStubs;
 }
--- a/rewriter/DefDelegator.cc
+++ b/rewriter/DefDelegator.cc
@@ -1,0 +1,74 @@
+#include "rewriter/DefDelegator.h"
+#include "ast/Helpers.h"
+#include "core/GlobalState.h"
+#include "rewriter/Util.h"
+#include <optional>
+
+using namespace std;
+
+namespace sorbet::rewriter {
+
+vector<ast::TreePtr> DefDelegator::run(core::MutableContext ctx, const ast::Send *send) {
+    vector<ast::TreePtr> empty;
+    auto loc = send->loc;
+
+    if (send->fun != core::Names::defDelegator()) {
+        return empty;
+    }
+
+    // This method takes 2..3 positional arguments and no keyword args:
+    // `def_delegator(accessor, method, ali = method)`
+    if (!(send->numPosArgs == 2 || send->numPosArgs == 3)) {
+        return empty;
+    }
+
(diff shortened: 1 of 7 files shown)


rewriter/DefDelegator.h
```diff
@@ -0,0 +1,35 @@
...
+ *    end
+ *
+ * into
+ *
+ *    class MyClass
+ *      sig {params(arg0: T.untyped).returns(T.untyped)}
+ *      def foo(*arg0); end
+ *
+ *      sig {params(arg0: T.untyped).returns(T.untyped)}
+ *      def aliased_bar(*arg0); end
```
paracycle, line 23: I think these should be:
```suggestion
 *      sig {params(arg0: T.untyped, blk: T.nilable(Proc)).returns(T.untyped)}
 *      def foo(*arg0, &blk); end
 *
 *      sig {params(arg0: T.untyped, blk: T.nilable(Proc)).returns(T.untyped)}
 *      def aliased_bar(*arg0, &blk); end
```

jeffcarbs: You’re right, this was copy/pasted from the description of the “delegate” rewriter which also should have a block arg so I’ll update both.

rewriter/DefDelegator.cc
```diff
@@ -0,0 +1,74 @@
...
+#include "core/GlobalState.h"
+#include "rewriter/Util.h"
+#include <optional>
+
+using namespace std;
+
+namespace sorbet::rewriter {
+
+vector<ast::TreePtr> DefDelegator::run(core::MutableContext ctx, const ast::Send *send) {
+    vector<ast::TreePtr> empty;
```
elliottt, line 12: Could you rename this vector to `methodStubs`, and remove the declaration of `methodStubs` on line 52?

rewriter/DefDelegator.cc
```diff
@@ -0,0 +1,74 @@
...
+    if (!(send->numPosArgs == 2 || send->numPosArgs == 3)) {
+        return empty;
+    }
+
+    if (send->hasKwArgs()) {
+        return empty;
+    }
+
+    auto *accessor = ast::cast_tree<ast::Literal>(send->args[0]);
+    auto *method = ast::cast_tree<ast::Literal>(send->args[1]);
```
elliottt, line 30: Could you move the declaration of method to right above the conditional where it's tested?

rewriter/DefDelegator.cc
```diff
@@ -0,0 +1,74 @@
...
+
+    methodStubs.push_back(ast::MK::Sig(loc, std::move(sigArgs), ast::MK::Untyped(loc)));
+
+    // def $methodName(*arg0, &blk); end
+    ast::MethodDef::ARGS_store args;
+    args.emplace_back(ast::MK::RestArg(loc, ast::MK::Local(loc, core::Names::arg0())));
+    args.emplace_back(ast::make_tree<ast::BlockArg>(loc, ast::MK::Local(loc, core::Names::blkArg())));
+
+    methodStubs.push_back(ast::MK::SyntheticMethod(loc, loc, methodName, std::move(args), ast::MK::EmptyTree()));
+
```
elliottt, line 70: Could you include the original call to `def_delegator` at the end of the `methodStubs` vector? Doing this will mean that we will still type-check the original call, and will throw errors if the class (or its parent) didn't `extend Forwardable`.

```suggestion
    methodStubs.push_back(send->deepCopy());
```

test/testdata/rewriter/def_delegator.rb
```diff
@@ -0,0 +1,65 @@
...
+class EnumerableUsage
+  extend T::Generic
+  extend Forwardable
+
+  include Enumerable
+
+  Elem = type_member
+
+  def_delegator :thing, :each
+end
```
elliottt, line 65: Could you add a test with a class that doesn't `extend Forwardable`?

rewriter/DefDelegator.cc
```diff
@@ -0,0 +1,74 @@
...
+    sigArgs.emplace_back(ast::MK::Nilable(loc, ast::MK::Constant(loc, core::Symbols::Proc())));
+
+    methodStubs.push_back(ast::MK::Sig(loc, std::move(sigArgs), ast::MK::Untyped(loc)));
+
+    // def $methodName(*arg0, &blk); end
+    ast::MethodDef::ARGS_store args;
+    args.emplace_back(ast::MK::RestArg(loc, ast::MK::Local(loc, core::Names::arg0())));
+    args.emplace_back(ast::make_tree<ast::BlockArg>(loc, ast::MK::Local(loc, core::Names::blkArg())));
+
+    methodStubs.push_back(ast::MK::SyntheticMethod(loc, loc, methodName, std::move(args), ast::MK::EmptyTree()));
```
elliottt, line 69: Instead of `ast::MK::EmptyTree()` for the body, could you use `ast::MK::RaiseUnimplemented(loc)`?

elliottt (changes requested):
Thanks! I've left some comments about restructuring the output of the rewriter a bit, and also requested an additional test.

I think it would be great if we could figure out a way to trigger this rewriter only when we know that `def_delegator` is being used in a context that is known to `extend Forwardable`, but unfortunately we don't have that information until the resolver pass :(

elliottt (approved):
This looks great, thank you! I'll merge it once it passes on stripe's codebase 👍

elliottt:

→ https://go/builds/bui_IRgK9EfNHYxWd4

will:
I was happy to find this pull request, then later sad to discover that Sorbet makes everything coming through a def delegator `T.untyped`

Is it possible ever for sorbet to copy forward the sig from the delegated method?

[sorbet.run link
](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20%0A%20%20sig%20%7B%20returns%20String%20%7D%0A%20%20def%20hi%0A%20%20%20%20%22hi%22%0A%20%20end%0Aend%0A%0Aclass%20B%0A%20%20extend%20Forwardable%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns%20A%20%7D%0A%20%20def%20a%0A%20%20%20%20A.new%0A%20%20end%0A%0A%20%20def_delegator%20%3Aa%2C%20%3Ahi%0Aend%0A%0A%0AT.reveal_type%20A.new.hi%0AT.reveal_type%20B.new.hi)

``` ruby
# typed: strict

class A
  extend T::Sig
  
  sig { returns String }
  def hi
    "hi"
  end
end

class B
  extend Forwardable
  extend T::Sig

  sig { returns A }
  def a
    A.new
  end

  def_delegator :a, :hi
end


T.reveal_type A.new.hi
T.reveal_type B.new.hi
```

``` gdb
editor.rb:25: Revealed type: String https://srb.help/7014
    25 |T.reveal_type A.new.hi
        ^^^^^^^^^^^^^^^^^^^^^^
  Got String originating from:
    editor.rb:25:
    25 |T.reveal_type A.new.hi
                      ^^^^^^^^

editor.rb:26: Revealed type: T.untyped https://srb.help/7014
    26 |T.reveal_type B.new.hi
        ^^^^^^^^^^^^^^^^^^^^^^
    editor.rb:26:
    26 |T.reveal_type B.new.hi
                      ^^^^^^^^
Errors: 2
```

PR #3637, opened 2020-11-12
Delete Type, GroundType, and ProxyType classes

jvilk-stripe:
Delete Type, GroundType, and ProxyType classes.

I believe the ProxyType deletion may be the most controversial, as we were getting some amount of mileage from that abstraction. I'm happy to remove it from this PR if that is the case.

### Motivation

Reduce the size of Type objects and eliminate virtual dispatch where possible.

### Test plan

Covered by existing tests.

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -205,8 +205,7 @@ template <> inline bool isa_type<ClassType>(const TypePtr &what) {
     }
 }
 
-class GroundType;
-template <> inline bool isa_type<GroundType>(const TypePtr &what) {
+inline bool is_ground_type(const TypePtr &what) {
     if (what == nullptr) {
         return false;
     }
@@ -279,8 +278,6 @@ template <> inline TypePtr const &TypePtr::cast<TypePtr>(const TypePtr &what) {
     template <> struct TypePtr::TypeToTag<name> { static constexpr TypePtr::Tag value = TypePtr::Tag::name; }; \
     class __attribute__((aligned(8))) name
 
-class GroundType {};
-
 class ProxyType {
 public:
     // TODO: use shared pointers that use inline counter
@@ -294,7 +291,7 @@ public:
 };
 CheckSize(ProxyType, 8, 8);
 
-TYPE(ClassType) : public GroundType {
+TYPE(ClassType) {
 public:
     SymbolRef symbol;
     ClassType(SymbolRef symbol);
@@ -420,7 +417,7 @@ public:
 };
 CheckSize(TypeVar, 8, 8);
 
-TYPE(OrType) final : public GroundType {
+TYPE(OrType) final {
 public:
     TypePtr left;
     TypePtr right;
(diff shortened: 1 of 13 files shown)


core/Types.h
```diff
@@ -315,13 +290,13 @@ TYPE(ClassType) : public GroundType {
     bool derivesFrom(const GlobalState &gs, SymbolRef klass) const;
     void _sanityCheck(const GlobalState &gs) const;
 };
-CheckSize(ClassType, 16, 8);
+CheckSize(ClassType, 8, 8);
```
jvilk-stripe, line 293: Smaller types 🎉

core/types/subtyping.cc
```diff
@@ -38,8 +38,8 @@ TypePtr Types::any(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2)
 }
 
 const TypePtr underlying(const TypePtr &t1) {
-    if (auto *f = cast_type<ProxyType>(t1)) {
-        return f->underlying();
+    if (is_proxy_type(t1)) {
+        return t1.underlying();
```
jez, line 42: So am I correct in understanding that there's nothing checking you if you try to call `.underlying()` on a type which it does not make sense to call `underlying` on?

Can we at least make `TypePtr::underlying` have something like `ENFORCE(is_proxy_type(*this))`?

core/Types.h
```diff
@@ -231,8 +222,7 @@ template <> inline bool isa_type<GroundType>(const TypePtr &what) {
     }
 }
 
-class ProxyType;
-template <> inline bool isa_type<ProxyType>(const TypePtr &what) {
+inline bool is_proxy_type(const TypePtr &what) {
```
jez, line 225: Can we mention all the types in the `switch` below? I want people to be forced to consider whether they are a proxy type or a ground type if they add a new `TypePtr::Tag`.

jez: I believe it used to be that exactly one of `is_ground_type` or `is_proxy_type` would be true. Can we implement one of these two helpers in terms of the other so that it remains true?

core/types/subtyping.cc
jvilk-stripe: You're looking at the wrong function. This is an existing function `underlying()`, not `TypePtr::underlying()`. `TypePtr::underlying()` throws an exception if it's not a proxy type.

core/Types.h
jvilk-stripe: Oh, sure!

core/types/subtyping.cc
jvilk-stripe: https://github.com/sorbet/sorbet/pull/3637/files#diff-8a69c6e75a3c53920d6c4a7e8b59337d5c99e35ed53d3b83b9fc0bb821726d00R53

core/Types.h
jvilk-stripe: @jez The following are neither ground nor proxy:

* LambdaParam
* SelfTypeParam
* AliasType
* SelfType
* TypeVar
* AppliedType

jez (approved):

core/Types.h
jvilk-stripe: Done; I made them explicit lists w/ no default.

PR #3639, opened 2020-11-12
Add a rewrite pass for `include Singleton`

elliottt:
This introduces a rewrite pass that adds a signature and method definition for `self.instance` when classes `include Singleton`. The rewrite takes a definition like:

```ruby
class A
  include Singleton
end
```

and turns it into:

```ruby
class A
  def self.instance
    raise "Sorbet rewriter pass partially unimplemented"
  end

  include Singleton
end
```

When the class is marked `final!` the generated `self.instance` method will have a signature that is also marked final.

### Motivation

Allowing singleton classes to also be marked as `final!`.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Singleton.cc
+++ b/rewriter/Singleton.cc
@@ -1,0 +1,58 @@
+#include "rewriter/Singleton.h"
+#include "ast/Helpers.h"
+#include "core/GlobalState.h"
+
+using namespace std;
+
+namespace sorbet::rewriter {
+
+namespace {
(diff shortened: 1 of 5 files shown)


elliottt:

→ https://go/builds/bui_INJ2cRKsB21MGC

jez (approved):
idk if you want to try to write an `srb init` test of the missing methods behavior with `final!` or not

PR #3640, opened 2020-11-12
pass DispatchArgs by const reference everywhere

froydnj:
We copy `DispatchArgs` in a number of different places, even though we don't really need to.  We make a half-hearted attempt to `move` in one particular place, but as `DispatchArgs` doesn't benefit from moving, that attempt doesn't do much better than the copying we do elsewhere.

So, let's just...not copy.

### Motivation

Epsilon efficiency gains, clearer code.

### Test plan

### DIFF
--- a/core/Symbols.h
+++ b/core/Symbols.h
@@ -27,7 +27,7 @@ class SerializerImpl;
 }
 class IntrinsicMethod {
 public:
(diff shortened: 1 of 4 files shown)


jvilk-stripe:
I was just thinking about this, as we use it like an immutable data structure anyway.

elliottt (approved):
Nice find!

PR #3641, opened 2020-11-12
Do not produce unreported errors for AndType::dispatchCall.

jvilk-stripe:
Do not produce unreported errors for AndType::dispatchCall.

Failure is the common case, and these errors are expensive to produce.

If all components in an AndType fail, re-dispatch the call with errors enabled.

Question:
* Are there any other places where we throw out errors? Could we change DispatchCall so that it doesn't force the caller to report errors?

### Motivation

Levenstein distance was showing up as super hot on profiles, and I tracked it to the (unreported) errors produced in AndType::dispatchCall.

### Test plan

Covered by existing tests.

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -677,9 +677,9 @@ struct DispatchArgs {
     const TypePtr &thisType;
     const std::shared_ptr<const SendAndBlockLink> &block;
     Loc originForUninitialized;
-    // Special option used for AndType::dispatchCall: Do not produce dispatch-related errors while evaluating the call.
-    // AndTypes only require one of their component types to handle the call, so failure is the common case -- and
-    // producing good error messages is expensive!
+    // Do not produce dispatch-related errors while evaluating the call. This is a performance optimization, as there
+    // are cases where we call dispatchCall with no intention of showing the errors to the user. Producing those
+    // unreported errors is expensive!
     bool suppressErrors = false;
 
     DispatchArgs withSelfRef(const TypePtr &newSelfRef) const;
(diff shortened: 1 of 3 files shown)


jez:
Can you share an example of how the error message changes at the command line? Or add a CLI test or something that shows this? (this is a user-visible change, right?)

jez:
If the problem is specifically Levenstein distance, can we do some sort of ad hoc short circuiting like we do for suggesting "did you mean" on constants in resolver? i.e., still report all the errors, but bail out of adding "did you mean" entries to the errors.

jvilk-stripe:
> Can you share an example of how the error message changes at the command line? Or add a CLI test or something that shows this? (this is a user-visible change, right?)

@jez This is _not_ a user visible change at all. We are not throwing away errors. We are delaying producing errors until we know they will become user visible.

jez:
> @jez This is _not_ a user visible change at all. We are not throwing away errors. We are delaying producing errors until we know they will become user visible.

perfect, thanks for the clarification!

jez (approved):

core/Types.h
```diff
@@ -677,9 +677,14 @@ struct DispatchArgs {
     const TypePtr &thisType;
     const std::shared_ptr<const SendAndBlockLink> &block;
     Loc originForUninitialized;
+    // Special option used for AndType::dispatchCall: Do not produce dispatch-related errors while evaluating the call.
+    // AndTypes only require one of their component types to handle the call, so failure is the common case -- and
+    // producing good error messages is expensive!
```
jvilk-stripe, line 682: Sounds like this comment might be confusing. I can clarify this field's semantics and the AndType details I can isolate to `AndType::dispatchCall`.

jvilk-stripe: Updated.

PR #3642, opened 2020-11-12
handle subtyping for AndType and TypeVars more carefully

### LINKED ISSUE #884: Using to_h with T.all forgets that the value type was a T.all
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20Parent%3B%20end%0Amodule%20Interface%3B%20end%0Aclass%20Child%20%3C%20Parent%0A%20%20include%20Interface%0Aend%0A%0Adef%20bar%0A%20%20x%20%3D%20T.let(%5B%5B%3Achild%2C%20Child.new%5D%5D%2C%20T%3A%3AArray%5B%5BSymbol%2C%20T.all(Parent%2C%20Interface)%5D%5D)%0A%20%20T.reveal_type(x.to_h)%0Aend)

```ruby
# typed: true
extend T::Sig

class Parent; end
module Interface; end
class Child < Parent
  include Interface
end

def bar
  x = T.let([[:child, Child.new]], T::Array[[Symbol, T.all(Parent, Interface)]])
  T.reveal_type(x.to_h)
end
```

#### Observed output

```
editor.rb:12: Revealed type: T::Hash[Symbol, Parent] https://srb.help/7014
    12 |  T.reveal_type(x.to_h)
          ^^^^^^^^^^^^^^^^^^^^^
  From:
    editor.rb:12:
    12 |  T.reveal_type(x.to_h)
                        ^^^^^^
Errors: 1
```

#### Expected behavior

```
editor.rb:12: Revealed type: T::Hash[Symbol, T.all(Parent, Interface)] https://srb.help/7014
    12 |  T.reveal_type(x.to_h)
          ^^^^^^^^^^^^^^^^^^^^^
  From:
    editor.rb:12:
    12 |  T.reveal_type(x.to_h)
                        ^^^^^^
Errors: 1
```

It looks like for some reason our `to_h` intrinsic method drops the fact that the value type was a `T.all`. I haven't confirmed whether this is expected behavior or a bug in the intrinsic method (or something else).

Reported by @akshay-stripe

froydnj:
When we were determining subtyping relations for (`AndType`, `TypeVar`), we were only considering half of the `AndType` in `isSubTypeUnderConstraint` -> `isSubTypeUnderConstraintSingle` -> `TypeConstraint::rememberIsSubtype`.

This is almost certainly not the most elegant way to solve this, but it makes a certain amount of intuitive sense to me, and all the other ways of addressing this didn't seem to work -- suggestions welcome!  It's possible that this needs to be some kind of interface directly on `TypeConstraint`, but I'm not sure what that would look like, nor do I understand why only checking for one-half of the `AndType` even worked before.  (The standard library RBIs have cases where one runs into (`Symbol & U`, `U`) being passed into `isSubTypeUnderConstraint`, at least, so maybe that's part of it.)

### Motivation

Fixes #884.

### Test plan

See included automated tests, which include all of the provided testcases from the above issue.

### DIFF
--- a/core/types/subtyping.cc
+++ b/core/types/subtyping.cc
@@ -1298,12 +1298,28 @@ bool Types::isSubTypeUnderConstraint(const GlobalState &gs, TypeConstraint &cons
         }
     }
 
+    // For these two cases, we have special cases if we are determining subtyping
+    // with respect to a TypeVar.  The rationale is that the non-TypeVar type has
+    // some structure that we wish to record: it's not correct to record each member
+    // individually in a recursive call to isSubTypeUnderConstraint since a) the
+    // recursive calls to isSubTypeUnderConstraint on each member would short-circuit
+    // and therefore not examine some members and b) we lose the structure of the
+    // original type.  But we only want to special-case TypeVars when recording type
+    // constraints prior to solving; once we have solved the type constraints, we
+    // want to look at each member individually and short-circuit as appropriate.
+
     // This order matters
     if (o2 != nullptr) { // 3
+        if (isa_type<TypeVar>(t1) && !constr.isSolved()) {
+            return constr.rememberIsSubtype(gs, t1, t2);
+        }
         return Types::isSubTypeUnderConstraint(gs, constr, t1, o2->left, mode) ||
                Types::isSubTypeUnderConstraint(gs, constr, t1, o2->right, mode);
     }
     if (a1 != nullptr) { // 4
+        if (isa_type<TypeVar>(t2) && !constr.isSolved()) {
+            return constr.rememberIsSubtype(gs, t1, t2);
+        }
         return Types::isSubTypeUnderConstraint(gs, constr, a1->left, t2, mode) ||
                Types::isSubTypeUnderConstraint(gs, constr, a1->right, t2, mode);
     }
--- a/test/testdata/infer/bug_generics_pr3642.rb
+++ b/test/testdata/infer/bug_generics_pr3642.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+# typed: strict
+extend T::Sig
+
+sig do
+  type_parameters(:U)
+    .params(
+      f: T.proc.params(x: T.type_parameter(:U)).void,
+      x: T.type_parameter(:U)
+    )
+    .void
+end
+def foo(f, x)
+end
+
+f = T.let(
+  proc {|x| nil},
+  T.proc.params(x: T.any(String, Integer)).void
+)
+
+foo(f, 0)
--- a/test/testdata/infer/generic_all.rb
+++ b/test/testdata/infer/generic_all.rb
@@ -1,0 +1,66 @@
+# typed: true
+
+module One
+  extend T::Sig
+
+  module A; end
+
+  module B; end
+
+  sig {params(x: Integer).returns(T.all(A, B))}
+  def return_ab(x)
+    return_ab(x)
+  end
+
+  def main
+    ab_list = [1,2,3].map {|i| return_ab(i)}
+    T.reveal_type(ab_list) # error: Revealed type: `T::Array[T.all(One::A, One::B)]`
+  end
+end
+
+module Two
+  extend T::Sig
+
+  class Parent; end
+  module Interface; end
+  class Child < Parent
+    include Interface
+  end
+
+  def bar
+    x = T.let([[:child, Child.new]], T::Array[[Symbol, T.all(Interface, Parent)]])
+    T.reveal_type(x.to_h) # error: Revealed type: `T::Hash[Symbol, T.all(Two::Interface, Two::Parent)]`
+  end
+end
+
+module Three
+  module WhateverInterface; end
+  module WhyInterface; end
+  module HelpInterface; end
+
+  class Test
+    include WhateverInterface
+    include WhyInterface
+    include HelpInterface
+  end
+
+  class A
+    extend T::Sig
+
+    Everything = T.type_alias do 
+      T.all(
+        WhateverInterface,
+        WhyInterface,
+        HelpInterface,
+      )
+    end
+    sig {returns(T::Array[Everything])}
+    def something
+      arr = T.let([Test.new], T::Array[Everything])
+      results = arr.map do |elem|
+        elem
+      end
+      results
+    end
+  end
+end


jez:
I happened to look at some code last night and noticed this TODO:

https://github.com/sorbet/sorbet/blob/b97904fb19df3c4b015d4e0e0073522251d9f42a/infer/environment.cc#L1080-L1081

I think that this TODO is causing the problem in #884. At that line, we're only looking at `.main` and not at the linked list chain of `.secondary` DispatchResults. (That's my hunch as to what the problem is)

froydnj:
I was seeing things go through https://github.com/sorbet/sorbet/blob/b97904fb19df3c4b015d4e0e0073522251d9f42a/infer/environment.cc#L957 and wind up in https://github.com/sorbet/sorbet/blob/b97904fb19df3c4b015d4e0e0073522251d9f42a/core/types/calls.cc#L2215-L2236 and getting back `Hash[Symbol, Parent]` (for the initial bug in #884).  I didn't see the `SolveConstraint` stuff on the stack, but it's certain possible that things went sideways there first.

core/types/subtyping.cc
```diff
@@ -1303,6 +1303,13 @@ bool Types::isSubTypeUnderConstraint(const GlobalState &gs, TypeConstraint &cons
                Types::isSubTypeUnderConstraint(gs, constr, t1, o2->right, mode);
     }
     if (a1 != nullptr) { // 4
+        auto *t2v = cast_type<TypeVar>(t2);
+        if (t2v && !constr.isSolved() && constr.hasLowerBound(t2v->sym) &&
+            ((const TypeConstraint &)constr).findLowerBound(t2v->sym) == Types::bottom()) {
```
jez, line 1308: Why are we casting here?

core/types/subtyping.cc
```diff
@@ -1303,6 +1303,13 @@ bool Types::isSubTypeUnderConstraint(const GlobalState &gs, TypeConstraint &cons
...
+            constr.rememberIsSubtype(gs, t1, t2);
+            return true;
+        }
+
```
jez, line 1312: This looks out of place. There is not similar code anywhere else in this function, so it surprises me to think that this is the solution.

i.e., most of the rest of this code is symmetric. that probably means we either need code like this everywhere, or the solution lies elsewhere.

jez: Is there a test we could write that gets a bug to manifest in the other cases? like

- t1 is a type var, and t2 is an or type
- t2 is a type var, and t1 is an or type
- ... etc

core/types/subtyping.cc
froydnj: The non-`const` overload is selected here, but it is `private` and therefore not applicable.  I don't understand why the `const` overload can't be considered for resolution here, but it's probably because I have not sufficiently internalized the C++ overload rules.

core/types/subtyping.cc
froydnj: I think "t2 is a type var and t1 is an or type" [already works](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20Parent%3B%20end%0Amodule%20Interface%3B%20end%0Aclass%20Child%20%3C%20Parent%0A%20%20include%20Interface%0Aend%0A%0Adef%20bar%0A%20%20x%20%3D%20T.let(%5B%5B%3Achild%2C%20Child.new%5D%5D%2C%20T%3A%3AArray%5B%5BSymbol%2C%20T.any(Parent%2C%20Interface)%5D%5D)%0A%20%20T.reveal_type(x.to_h)%0Aend).

If I understand the code correctly, "t1 is a type variable and t2 is an or type" should also fail to reveal the correct type, because we'd hit:

https://github.com/sorbet/sorbet/blob/c3344f71641f737916e1d38c8d64c83d225b3d5f/core/types/subtyping.cc#L1300-L1304

but we'd a) never check the right side of the or for subtyping and b) even if we did, we'd run into:

https://github.com/sorbet/sorbet/blob/c3344f71641f737916e1d38c8d64c83d225b3d5f/core/types/subtyping.cc#L990-L996
https://github.com/sorbet/sorbet/blob/c3344f71641f737916e1d38c8d64c83d225b3d5f/core/TypeConstraint.cc#L98-L122

I *think* the other combinations of type var and and/or type get the correct relations out of that code in `TypeConstraint`.

core/types/subtyping.cc
trevor-stripe: Another option for this case would be the following:
```
     if (a1 != nullptr) { // 4
         auto l = Types::isSubTypeUnderConstraint(gs, constr, a1->left, t2, mode);
         auto r = Types::isSubTypeUnderConstraint(gs, constr, a1->right, t2, mode);
         return l || r;
     }
```
The problem is that `Types::isSubTypeUnderConstraint` will learn new constraints as it processes types, so short-circuiting while processing the AndType in the LHS of the subtyping relation causes some constraints to be forgotten.

froydnj: We could do short-circuiting if `!isa_type<TypeVar>(t2)` to only take a small perf hit for the generic case.

elliottt: That seems reasonable to me. Would you mind running this on stripe's codebase, and seeing what sort of performance hit we would have without the special-case for TypeVar?

jez: I'm kind of thinking maybe we should just yoink these lines out of isSubtypeUnderConstraintSingle into  the top of isSubtypeUnderconstraint?

https://github.com/sorbet/sorbet/blob/a7b4be1fbe928c129b7cc02cdae161bab2f33f4f/core/types/subtyping.cc#L990-L996

jez: (we should test that this is correct and fast though, i don't claim for sure that it's either)

core/types/subtyping.cc
jez: Here is a test case that currently fails but I don't think it should:

```ruby
# typed: strict
extend T::Sig

sig do
  type_parameters(:U)
    .params(
      f: T.proc.params(x: T.type_parameter(:U)).void,
      x: T.type_parameter(:U)
    )
    .void
end
def foo(f, x)
end

f = T.let(
  proc {|x| nil},
  T.proc.params(x: T.any(String, Integer)).void
)

# These should be the constraints that we find:
#
#   T.type_parameter(:U) <: T.any(String, Integer)
#   Integer <: T.type_parameter(:U)
#
# which has a solution (really many solutions,
# of which TypeConstraint::solve should return any.)
#
# But in fact, these are the constraints that we find:
#
#   T.type_parameter(:U) <: String
#   Integer <: T.type_parameter(:U)
foo(f, 0)
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Asig%20do%0A%20%20type_parameters(%3AU)%0A%20%20%20%20.params(%0A%20%20%20%20%20%20f%3A%20T.proc.params(x%3A%20T.type_parameter(%3AU)).void%2C%0A%20%20%20%20%20%20x%3A%20T.type_parameter(%3AU)%0A%20%20%20%20)%0A%20%20%20%20.void%0Aend%0Adef%20foo(f%2C%20x)%0Aend%0A%0Af%20%3D%20T.let(%0A%20%20proc%20%7B%7Cx%7C%20nil%7D%2C%0A%20%20T.proc.params(x%3A%20T.any(String%2C%20Integer)).void%0A)%0A%0A%23%20These%20should%20be%20the%20constraints%20that%20we%20find%3A%0A%23%0A%23%20%20%20T.type_parameter(%3AU)%20%3C%3A%20T.any(String%2C%20Integer)%0A%23%20%20%20Integer%20%3C%3A%20T.type_parameter(%3AU)%0A%23%0A%23%20which%20has%20a%20solution%20(really%20many%20solutions%2C%0A%23%20of%20which%20TypeConstraint%3A%3Asolve%20should%20return%20any.)%0A%23%0A%23%20But%20in%20fact%2C%20these%20are%20the%20constraints%20that%20we%20find%3A%0A%23%0A%23%20%20%20T.type_parameter(%3AU)%20%3C%3A%20String%0A%23%20%20%20Integer%20%3C%3A%20T.type_parameter(%3AU)%0Afoo(f%2C%200))

More generally, this is what I was getting at when I was saying this solution felt ad-hoc:

Almost every type system feature that deals with union types has a dual feature for intersection types.
It is a huge red flag when the code to handle one case does not share symmetry with the other case.
With a little extra effort, we can find test cases that go through the other branches that share the same bugs.

froydnj: That's fair.  The good news is that this test case is easily fixable by changing:

https://github.com/sorbet/sorbet/blob/c3344f71641f737916e1d38c8d64c83d225b3d5f/core/types/subtyping.cc#L1300-L1304

to not short-circuit.

core/types/subtyping.cc
froydnj: I don't think changing case 4 in `isSubTypeUnderConstraint` to:

```c++
    if (a1 != nullptr) { // 4
        bool l = Types::isSubTypeUnderConstraint(gs, constr, a1->left, t2, mode);
        bool r = Types::isSubTypeUnderConstraint(gs, constr, a1->right, t2, mode);
        return l || r;
    }
```

works.  Running that change over the added test cases from the linked bug (which are added in this PR) gives:

```
test/testdata/infer/generic_all.rb:17: ERROR: Expected error of form:
17:     T.reveal_type(ab_list) # error: Revealed type: `T::Array[T.all(One::A, One::B)]`
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Revealed type: `T::Array[T.all(One::A, One::B)]`
Found error:
17:     T.reveal_type(ab_list) # error: Revealed type: `T::Array[T.all(One::A, One::B)]`
        ^^^^^^^^^^^^^^^^^^^^^^ error: Revealed type: `T::Array[T.any(One::A, One::B)]`

test/testdata/infer/generic_all.rb:32: ERROR: Expected error of form:
32:     T.reveal_type(x.to_h) # error: Revealed type: `T::Hash[Symbol, T.all(Two::Interface, Two::Parent)]`
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Revealed type: `T::Hash[Symbol, T.all(Two::Interface, Two::Parent)]`
Found error:
32:     T.reveal_type(x.to_h) # error: Revealed type: `T::Hash[Symbol, T.all(Two::Interface, Two::Parent)]`
        ^^^^^^^^^^^^^^^^^^^^^ error: Revealed type: `T::Hash[Symbol, T.any(Two::Interface, Two::Parent)]`

test/testdata/infer/generic_all.rb:63: ERROR: Found unexpected error:
63:       results
          ^^^^^^^ error: Returning value that does not conform to method result type
Note: If there is already an assertion for this error, then this is a duplicate error. Change the assertion to `# error-with-dupes: <error message>` if the duplicate is expected.
```

So we're not throwing away information, at least, but we're deriving `T.any` instead of `T.all`.

We are running into the asymmetry in `TypeConstraint::rememberIsSubtype`, and so we have to account for that somehow.

froydnj:
All right, updated the patch after going over things with @elliottt .  Notable changes:

* Added test case from @jez 
* and/or cases now have a pleasing symmetry, which addresses the aforementioned testcase
* Comment describes why we're doing what we're doing and (ideally) why we're not putting this check someplace else (e.g. dealing with `TypeVar` directly at the top of `isSubTypeUnderConstraint`.

elliottt (approved):
This looks good to me, glad we can finally land this!

froydnj:

→ https://go/builds/bui_JGGPPtzww8Pkge
→ https://go/builds/bui_JGGPtJlFa6t0SO

jez:
Thanks for the diligence!

PR #3644, opened 2020-11-12
don't needlessly allocate in _instantiate and _approximate

froydnj:
There's a TODO (effectively nine TODOs) in typemaps.cc that we shouldn't allocate new vectors in `_instantiate` and `_approximate` unless we actually needed to.  This PR makes that change, and commons up all the logic so we only have to implement it once, possibly making it a little more readable in the process.

### Motivation

Epsilon efficiency improvements, more readable code.

### Test plan

### DIFF
--- a/core/types/typemaps.cc
+++ b/core/types/typemaps.cc
@@ -56,168 +56,88 @@ TypePtr TypeVar::_approximate(const GlobalState &gs, const TypeConstraint &tc) c
     return Types::top();
 }
 
+namespace {
+template <typename... TransformArgs>
+optional<vector<TypePtr>> mungeArgs(const vector<TypePtr> &elems,
+                                    TypePtr (TypePtr::*method)(const TransformArgs &...) const,
+                                    const TransformArgs &... transformArgs) {
+    optional<vector<TypePtr>> newArgs;
+    int i = -1;
+    for (auto &e : elems) {
+        ++i;
+        auto t = (e.*method)(transformArgs...);
+        if (!newArgs.has_value() && !t) {
+            continue;
+        }
+
+        if (!newArgs.has_value()) {
+            // Oops, need to fixup all the elements that should be there.
+            newArgs.emplace();
+            newArgs->reserve(elems.size());
+            for (int j = 0; j < i; ++j) {
+                newArgs->emplace_back(elems[j]);
+            }
+        }
+
+        if (!t) {
+            t = e;
+        }
+
+        ENFORCE(newArgs->size() == i);
+        newArgs->emplace_back(t);
+    }
+    return newArgs;
+}
+} // anonymous namespace
+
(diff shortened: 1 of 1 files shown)


jez:
Can you test this on Stripe's codebase to make sure it doesn't crash?

core/types/typemaps.cc
```diff
@@ -56,168 +56,88 @@ TypePtr TypeVar::_approximate(const GlobalState &gs, const TypeConstraint &tc) c
...
-    for (auto &a : this->elems) {
-        auto t = a._instantiate(gs, params, targs);
-        if (changed || t) {
-            changed = true;
-            if (!t) {
-                t = a;
+namespace {
+template <typename... MethodArgs>
+optional<vector<TypePtr>> mungeElems(const vector<TypePtr> &elems,
+                                     TypePtr (TypePtr::*method)(const MethodArgs &...) const,
```
jez, line 62: i think this is the first function pointer to a method that i've seen in sorbet's codebase. not sure how i feel about it. do you think we could just implement this twice, once for _instantiate and once for _approximate?

jez (approved):

core/types/typemaps.cc
froydnj: I am always unreasonably happy when I find a legitimate place to use pointers to methods (which happens maybe once a year).  Do you have particular uneasynesses (sp?) that you could share?  I feel like in a codebase that already takes advantage of `if constexpr`, variable-arg templates, and sfinae metaprogramming (among other things), pointers to methods are a relatively minor thing.  But I could but convinced otherwise.

froydnj:

→ https://go/builds/bui_INs3bWup3GkThE

core/types/typemaps.cc
jez: I mostly just don't like the indirection. It breaks jump to definition, where now i have to see all the call sites to understand what this method could be doing (and in fact that's what I did when reviewing: i looked at the call sites to see that it's only ever one of `_instantiate` or `_approximate`).

I think that luckily our usage of variable arg templates is currently limited to forwarding args to `fmt` (for things like `Exception::raise`)

also idk what `sp?` means, if you could expand that for me

trevor-stripe: What about accepting a `std::function<TypePtr()>` instead, and putting the method call in the body of a lambda? Given that it wouldn't be a template anymore, it might be worth checking on godbolt to see how it inlines.

trevor-stripe: Oh, I suppose it would need to be `std::function<TypePtr(TypePtr &)>`, but still :)

froydnj: Indirecting through `std::function` looks pretty bad: https://gcc.godbolt.org/z/E31c8T  It inlines, at least, but produces a bunch of other garbage that the compiler can't seem to get rid of.

Making jump-to-definition not work seems reasonable.  I can make that change.

> also idk what `sp?` means, if you could expand that for me

Oh, sorry, "spelling?".

PR #3646, opened 2020-11-13
Add remote support for the cfg-view.sh script

jez:
### Motivation

Testing on linux machines.

### Test plan

Tested manually


elliottt (approved):
Oooh excellent, I've been wanting this!

PR #3650, opened 2020-11-13
TypeConstraint::toString, but different

jez:
This way it shows the upper and lower bounds co-located, instead of the
raw representation which would show them how they appear in the raw
storage.

### Test plan

This function is not called anywhere in the codebase. It appears to exist only for interactive debugging purposes.


elliottt (approved):
Nice!

PR #3651, opened 2020-11-13
Inline ClassType, SelfTypeParam, AliasType, SelfType, LiteralType, BlamedUntyped into TypePtr.

jvilk-stripe:
Inline ClassType, SelfTypeParam, AliasType, SelfType, LiteralType, BlamedUntyped into TypePtr.

Inlined types have the following restrictions:
* You _cannot_ use `cast_type` on them, since `cast_type` returns a pointer. There is a static_assert that tells you to use `isa_type` to check the type within the TypePtr and `cast_type_nonnull` to cast a TypePtr to a type.
  * 99% of this diff is converting `cast_type` to `isa_type`/`cast_type_nonnull` checks.

Making a new inlined type involves the following steps (compilation will fail if any are missed :)):
* The type must be defined with `TYPE_INLINED`.
* `make_type` must contain specializations for every unique combination of parameter types passed to `make_type<YourClass>`.
  * This is a limitation of C++, which does not support partial specialization of function templates.
* `cast_type_nonnull` must contain a specialization for your class that returns a value type rather than a reference type.

Typecase still supports lambdas that accept inlined types as const references, as C++ allows local const references to prolong the lifetime of values. Thus, I did not have to change the type signature of any typecase lambdas! 🎉  

### Motivation

Speed!

* Inlined types no longer have to increment/decrement an atomic reference counter, which is a scalability bottleneck for commonly accessed types (e.g., nilType).
* Inlined types do not have to allocate a Type object or an atomic counter object on the heap.
* Inlined types can be compared by value.

### Test plan

See included automated tests.

### DIFF
--- a/core/TypePtr.h
+++ b/core/TypePtr.h
@@ -252,10 +252,16 @@ public:
     }
 
     bool operator!=(const TypePtr &other) const {
-        return store != other.store || counter != other.counter;
+        // There's a lot going on in this line.
+        // * If store == other.store, both `this` and `other` have the same value of `containsPtr()`.
+        // * If store == other.store and both contain a pointer, there's no need to compare `counter`; they point to the
+        // same Type object.
+        // * If store == other.store and both do not contain a pointer, then we need to compare the inlined values.
+        return store != other.store || (!containsPtr() && value != other.value);
     }
     bool operator==(const TypePtr &other) const {
-        return store == other.store && counter == other.counter;
+        // Inverse of !=
+        return store == other.store && (containsPtr() || value == other.value);
     }
     bool operator!=(std::nullptr_t n) const {
         return store != 0;
--- a/core/TypePtr.h
+++ b/core/TypePtr.h
@@ -41,12 +41,17 @@ public:
     // A mapping from type to whether or not that type is inlined into TypePtr.
     template <typename T> struct TypeToIsInlined;
 
+    // A mapping from type to the type returned by `cast_type_nonnull`.
+    template <typename T, bool isInlined> struct TypeToCastType {};
+    template <typename T> struct TypeToCastType<T, true> { using type = T; };
+    template <typename T> struct TypeToCastType<T, false> { using type = const T &; };
+
     // Required for typecase.
     template <class To> static bool isa(const TypePtr &what);
 
-    template <class To> static To const &cast(const TypePtr &what);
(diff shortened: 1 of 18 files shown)


jvilk-stripe:

→ https://go/builds/bui_INx7zjNvIt51AS

core/TypePtr.h
```diff
@@ -131,7 +206,11 @@ class TypePtr final {
 
     TypePtr &operator=(TypePtr &&other) noexcept {
         handleDelete();
-        counter = other.releaseCounter();
+        if (other.containsPtr()) {
```
elliottt, line 209: What about when `*this == other`?

core/Types.h
```diff
@@ -312,6 +331,24 @@ TYPE(ClassType) {
 };
 CheckSize(ClassType, 8, 8);
 
+template <> inline TypePtr make_type<ClassType, core::SymbolRef &>(core::SymbolRef &ref) {
+    return TypePtr(TypePtr::Tag::ClassType, ref.rawId(), 0);
+}
+
+template <> inline TypePtr make_type<ClassType, core::SymbolRef>(core::SymbolRef &&ref) {
+    return TypePtr(TypePtr::Tag::ClassType, ref.rawId(), 0);
+}
```
elliottt, line 340: Just golfing on memory layout, but: what if you store the rawId in `value` instead? It's not used for anything at this point, and if the tags for inlined values always fit in the first three bits of the ptr, you could do a pretty quick check and then return `value`, avoiding the shift. You could also use the first three bits being non-null for the inlined check this way.

elliottt (approved):
This looks great! Super excited for the perf improvements 🎉 

I wondered about maybe always storing the values of inlined types in `value`, so that you could avoid the shift to retrieve it, but it's definitely a non-blocking change.

core/Types.h
```diff
@@ -64,6 +65,8 @@ class ArgInfo {
 CheckSize(ArgInfo, 48, 8);
 
 template <class T, class... Args> TypePtr make_type(Args &&... args) {
+    // Every inlined type must specialize `make_type`.
+    static_assert(!TypePtr::TypeToIsInlined<T>::value);
```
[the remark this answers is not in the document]
jvilk-stripe: Can do.

core/Types.h
jvilk-stripe: I can explore that as a follow-up change and see if there's any perf change. I suspect that, with everything being in cache, there's no difference.

PR #3652, opened 2020-11-14
.bazelrc: Enable -Wmissing-field-initializers

jez:
### Motivation

This helps tremendously when refactoring.

### Test plan

If it compiles it works.


elliottt (approved):
Yay! I really could have used this recently!

PR #3655, opened 2020-11-14
Stop using "any" when it doesn't mean `T.any`

jez:
### Motivation

This was confusing. In Sorbet, "any" always means `T.any`.

It should not mean "what we call the unwriteable top type" nor "some concrete
type variable."

### Test plan

Updated snapshots.


elliottt (approved):
Big fan of this change 👍

PR #3657, opened 2020-11-14
Call parameters "params" not "arguments"

jez:
### Motivation

When reading `dispatchCallSymbol`, it's really annoying to keep track whether a
variable refers to something at the method call site or the method definition
site.

I want to fix this by being more intentional with our variable names.
We should call the things at the definition site "parameters", because that's
what they are. Arguments are the things at the call site.

This PR starts by renaming `Symbol::arguments` to `Symbol::params`, and fixes
up some variable names.

It doesn't eradicate the problem, but it is a good first start

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/arity-messages/arity-messages.out
+++ b/test/cli/arity-messages/arity-messages.out
@@ -4,7 +4,7 @@ test/cli/arity-messages/arity-messages.rb:4: Too many positional arguments provi
     test/cli/arity-messages/arity-messages.rb:3: `foo` defined here
      3 |def foo(x: nil); end
         ^^^^^^^^^^^^^^^
-    test/cli/arity-messages/arity-messages.rb:4: `Object#foo` has optional keyword arguments. Did you mean to provide a value for `x`?
+    test/cli/arity-messages/arity-messages.rb:4: `Object#foo` has optional keyword parameters. Did you mean to provide a value for `x`?
      4 |foo 1
         ^^^^^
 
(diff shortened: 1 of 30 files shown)


test/cli/arity-messages/arity-messages.out
```diff
@@ -4,7 +4,7 @@ test/cli/arity-messages/arity-messages.rb:4: Too many positional arguments provi
     test/cli/arity-messages/arity-messages.rb:3: `foo` defined here
      3 |def foo(x: nil); end
         ^^^^^^^^^^^^^^^
-    test/cli/arity-messages/arity-messages.rb:4: `Object#foo` has optional keyword arguments. Did you mean to provide a value for `x`?
+    test/cli/arity-messages/arity-messages.rb:4: `Object#foo` has optional keyword parameters. Did you mean to provide a value for `x`?
```
elliottt, line 7: Does ruby take a stance on naming here? I'm fine with us keeping the parameters/arguments distinction internally, but it would be good to follow the same terminology that ruby uses when displaying errors.

A quick experiment shows that they call the method parameters arguments as well, how do you feel about using `arguments` to refer to the method parameters in error messages?

```
irb(main):006:0> def foo(x,x)
irb(main):007:1> end
        3: from /Users/trevor/.rbenv/versions/2.6/bin/irb:23:in `<main>'
        2: from /Users/trevor/.rbenv/versions/2.6/bin/irb:23:in `load'
        1: from /Users/trevor/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
SyntaxError ((irb):6: duplicated argument name)
... _; rescue _.class; def foo(x,x)
...                              ^
```

elliottt (approved):
I think this is a useful change for our internal representation, but am interested in your thoughts about trying to match ruby as closely as possible for terminology in error messages.

jez:
I think this is a huge yak and i don't really want to do it now.

PR #3658, opened 2020-11-15
Add test for method generics, and better error

jez:
### Motivation

There was not a single test for this error message before.
Also the error message was bad.

I made some additions to ErrorLine because I've wanted to be able to write
notes like this on an error message for a while.

### Test plan

**Before**

Screen Shot 2020-11-14 at 7 32 49 PM

**After**

Screen Shot 2020-11-14 at 7 35 18 PM

### DIFF
--- a/core/Error.cc
+++ b/core/Error.cc
@@ -53,7 +53,14 @@ string restoreColors(string_view formatted, rang::fg color) {
 string ErrorLine::toString(const GlobalState &gs, bool color) const {
     stringstream buf;
     string indent = "  ";
-    buf << indent << FILE_POS_STYLE << loc.filePosToString(gs) << RESET_STYLE << ":";
+    buf << indent;
+    switch (this->displayLoc) {
+        case LocDisplay::Shown:
+            buf << FILE_POS_STYLE << loc.filePosToString(gs) << RESET_STYLE << ":";
+            break;
+        case LocDisplay::Hidden:
+            break;
+    }
     if (!formattedMessage.empty()) {
         buf << " ";
         if (color) {
--- a/core/Error.cc
+++ b/core/Error.cc
@@ -62,7 +62,13 @@ string ErrorLine::toString(const GlobalState &gs, bool color) const {
             break;
     }
     if (!formattedMessage.empty()) {
-        buf << " ";
+        switch (this->displayLoc) {
+            case LocDisplay::Shown:
+                buf << " ";
+                break;
+            case LocDisplay::Hidden:
+                break;
+        }
         if (color) {
             buf << DETAIL_COLOR << restoreColors(formattedMessage, DETAIL_COLOR) << RESET_COLOR;
         } else {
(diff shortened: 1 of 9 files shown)


core/Error.cc
```diff
@@ -53,7 +53,14 @@ string restoreColors(string_view formatted, rang::fg color) {
...
     string indent = "  ";
-    buf << indent << FILE_POS_STYLE << loc.filePosToString(gs) << RESET_STYLE << ":";
+    buf << indent;
+    switch (this->displayLoc) {
+        case LocDisplay::Shown:
+            buf << FILE_POS_STYLE << loc.filePosToString(gs) << RESET_STYLE << ":";
+            break;
+        case LocDisplay::Hidden:
+            break;
+    }
```
jez, line 63: I've wanted something like this for a while. We have a handful of errors that show messages in `ErrorSection` titles (which turns them yellow!) because they don't have Loc's associated with them.

core/TypeConstraint.cc
```diff
@@ -276,4 +276,41 @@ std::string TypeConstraint::toString(const core::GlobalState &gs) const {
...
+    for (const auto &[sym, lowerBound] : this->lowerBounds) {
+        auto &[lowerRef, _upperRef] = boundsFor[sym];
+        ENFORCE(lowerRef == nullptr, "{} in lowerBounds twice?", sym.show(gs));
+        lowerRef = lowerBound;
+    }
+    for (const auto &[sym, upperBound] : this->upperBounds) {
+        auto &[_lowerRef, upperRef] = boundsFor[sym];
+        ENFORCE(upperRef == nullptr, "{} in upperBounds twice?", sym.show(gs));
+        upperRef = upperBound;
+    }
```
elliottt, line 291: You added this for pretty-printing the constraint set as well, what do you think about extracting it as a utility function?

test/cli/no-valid-instantiation/no-valid-instantiation.out
```diff
@@ -0,0 +1,9 @@
+test/cli/no-valid-instantiation/no-valid-instantiation.rb:26: Could not find valid instantiation of type parameters for `Object#foo` https://srb.help/7020
+    26 |foo(f, 0) # error: Could not find valid instantiation of type parameters
+        ^^^^^^^^^
+    test/cli/no-valid-instantiation/no-valid-instantiation.rb:12: `Object#foo` defined here
+    12 |def foo(f, x)
+        ^^^^^^^^^^^^^
+  Found no solution for these constraints:
+    `Integer(0)` must be a subtype of `U` which must be a subtype of `String`
```
elliottt, line 8: This is great!

elliottt (approved):
I love this new error message!

test/cli/no-valid-instantiation/no-valid-instantiation.out
jez: I actually think it's a bad error message, but the old one was worse.

We don't keep TypeAndOrigins data structures around through all the `rememberIsSubtype` calls, so we don't have any locations to point to as to why we gathered these constraints.

For this example it's trivial to see where they come from but that is not always the case.

core/TypeConstraint.cc
jez: https://github.com/sorbet/sorbet/blob/c07eeaa6989dc16ae18ccfa7b0cf9852574f26f5/core/TypeConstraint.cc#L247-L262

PR #3659, opened 2020-11-15
Add addErrorNote to give a name to a common pattern

jez:
### Motivation

We have "notes" on certain error messages, i.e., suggestions that aren't
associated with locations.

These look bad because most of the time we use ErrorSection titles for these,
which show up entirely in yellow. Doing it this way makes it more consistent
and look nicer.

### Test plan

**Before**

Screen Shot 2020-11-14 at 8 03 08 PM

**After**

Screen Shot 2020-11-14 at 8 17 31 PM


elliottt (approved):
Nice!

@aprocter-stripe, rebasing over this once it lands would mean you could leave a more wordy error message :)

jez:
> @aprocter-stripe, rebasing over this once it lands would mean you could leave a more wordy error message :)

Not encouraging wordy error messages 😅 we should have wordy error messages in srb.help links, and use `Note` for things that are short and nearly universally applicable, but which don't have locations.

PR #3660, opened 2020-11-15
Better error/debug printing for TypeVar's and LiteralType's

jez:
### Motivation

In error messages, we used to show `T.type_parameter(:U)` as `U`, which is
actually the same type syntax as a class named `U` 😱

We also used to put quotes around symbols somewhat needlessly when printing
them in trees (and types for that matter).

### Test plan

Lots of snapshots change


elliottt (approved):
Nice, I like this display for symbols much more 👍

PR #3663, opened 2020-11-16
Add test for private methods with T.any/T.all

jez:
### Motivation

This was already implemented correctly, but as far as I could tell it
was not exercised by our test suite, so I figured I'd add a test.

### Test plan

See included automated tests.


elliottt (approved):
👍  nice test!

PR #3565, opened 2020-10-23
Type check for private constants

bss:
This adds support for type checking private constants statically:
```ruby
class Foo
  BAR = 123
  private_constant :BAR
end
Foo::BAR # error: private constant Foo::BAR referenced
```

### Motivation
This is currently not checked statically by sorbet.

### Test plan
See included automated tests.

PR #3666, opened 2020-11-16
wip: ZSuper methods (private_class_method :new)

jez:
Based off of #3565 because it would conflict otherwise.

### Motivation

A huge gaping hole with private method support is being able to support
`private_class_method :new`.

We could special case that but I don't really like that, because I think
there's a way to do it that's not too invasive.

It's late and the weekend is over so I'm throwing this PR up.
It's not ready yet. There are still some changes to make to get it working, but
I've gotten far enough that I think the overall strategy is workable.

Apart from the TODO comments which are there to prevent me forgetting
something, one other thing I want to write down while it's top of mind: I
audited call sites to `findMemberTransitive` and `findMember`.

- The **vast** majority of `findMember` sites are looking at constants. Don't
  be scared to audit them again, because there's only like a handful that look
  at methods.

- When I looked at the `findMemberTransitive` calls, almost all of them looked
  like they actually would prefer to skip these ZSuper methods, but as I went
  on further, I was less convinced.

  Again: there's only like 3 places where we call `findMemberTransitive`, so
  it's totally fine to just look again (and a bunch of them are looking for
  constants, not methods).

### Test plan

I wrote a lot of tests. I think I actually wrote some pretty good tests.

I'm also pretty confident that running this change on Stripe's codebase will
provide more good test cases if it doesn't pass outright.


test/testdata/infer/private_result_type_as_seen_from.rb
```diff
@@ -0,0 +1,28 @@
...
+  private :get_self
+
+  def example
+    T.reveal_type(elem) # error: Revealed type: `Integer`
+
+    self.elem = '' # error: Assigning a value to `elem` that does not match expected type `Integer`
+
+    T.reveal_type(get_self) # error: Revealed type: `Child`
+  end
+end
```
jez, line 28: My first approach was to register an intrinsic that behaved somewhat like `<Magic>.<self-new>` and simulated the dispatch call on the super method.

I abandoned that approach halfway through when I realized that `dispatchCallSymbol` takes a `targs` vector but we don't have a targs vector in the `apply` callback of IntrinsicMethod, and even if we did, I didn't want to have to futz around with aligning those type args as if they were seen from the parent class.

jez:
Some notes from chatting with dmitry:

- Scala / Java calls these bridge methods, and we might want to use that name (I'd rather use what the Ruby VM calls it, but worth noting)
  - https://docs.oracle.com/javase/tutorial/java/generics/bridgeMethods.html#bridgeMethods
- this breaks liskov substitution principle even more than before
  - tests that we would not catch a static error caused by passing a `PrivateInChild` instance to a method that takes `PublicInParent` instances
  - add checks that you can't mark `abstract` methods private in a child class (maybe overridable too? what about abstract -> override -> private zsuper?)
- another way to do this would be to break namer + resolver into class namer/resolver and method namer/resolver
  - run namer for found defs that are classes / constants, run ResolveConstantsWalk, run class linearization
  - run namer for found defs—now we have the ancestor hierarchy finalized, so we can do fast method lookups during method namer
  - (this would be instead of changing findMemberTransitive)

jez:

→ https://go/builds/bui_IQ8KTHjjKJS8Kk

jez:

→ https://go/builds/bui_IQM1KxKFmj07NJ

jez:

PR #3667, opened 2020-11-16
reduce copying in QualifiedName::fromFullName

froydnj:
We can do a better job of re-using the passed-in vector in `QualifiedName::fromFullName`.  The compiler is unable to merge the two `erase` calls, so we should help it out.

I tried looking a little farther afield to see if we could avoid the O(N) `erase` call, replacing it with a decrease in the vector length, by storing names reversed in most cases, but I think this runs into too much duplication and complexity.

### Motivation

Epsilon efficiency improvements.

### Test plan


jez (approved):

froydnj:

→ https://go/builds/bui_IPTSihNBMhzMYB

PR #3672, opened 2020-11-16
Small definition validator cleanup/optimization

jvilk-stripe:
Small definition validator cleanup/optimization.

* Don't call `bestNonRBIFile` unless we encounter a sealed class.
* Put most of validator into anonymous namespace.
* Use `cast_type_nonnull` in tree walk.

### Motivation

Speedup. I noticed `bestNonRBIFile` shows up in profiles, and discovered we could avoid calling it most of the time.

### Test plan

Covered by existing tests.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -347,25 +347,25 @@ void validateOverriding(const core::Context ctx, core::SymbolRef method) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


jvilk-stripe:

→ https://go/builds/bui_IP1LTu73L6mZ7P

jez (approved):

PR #3675, opened 2020-11-17
Stop trying to detect a keyword args hash while processing positional args

### LINKED ISSUE #3647: Defaulted hash and keyword arguments are conflicting
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams(y%3A%20T%3A%3AHash%5BString%2CString%5D%2C%20z%3A%20String).void%7D%0A%20%20def%20self.bar(y%3D%7B%7D%2C%20z%3A)%0A%20%20end%0Aend%0A%0AA.bar(%7B%20%22a%22%20%3D%3E%20%22foo%22%20%7D))

```ruby
# typed: strict
class A
  extend T::Sig

  sig {params(y: T::Hash[String,String], z: String).void}
  def self.bar(y={}, z:)
  end
end

A.bar({ "a" => "foo" })
```

#### Observed output

```
editor.rb:10: Missing required keyword argument z for method A.bar https://srb.help/7004
    10 |A.bar({ "a" => "foo" })
        ^^^^^^^^^^^^^^^^^^^^^^^

editor.rb:10: Unrecognized keyword argument a passed for method A.bar https://srb.help/7004
    10 |A.bar({ "a" => "foo" })
        ^^^^^^^^^^^^^^^^^^^^^^^
Errors: 2
```

#### Expected behavior

There shouldn't be an error raised about the keyword argument `"a"`, as that's not a keyword argument the method accepts.
- - -

This is related to the refactoring to move keyword arguments from the last hash argument to inlined in the send itself. The part of `dispatchCallSymbol` that handles the case where the last positional argument may be used as the keyword args hash isn't correct currently:
https://github.com/sorbet/sorbet/blob/0b44045adf1050d1358505952485546e42670057/core/types/calls.cc#L822-L825

elliottt:
Before we inlined keyword arguments into the Send node, we would break out of positional argument processing early when a final hash argument was detected. That logic is at odds with the current implementation of keyword args, yielding the bug in #3647. Removing it fixes the issue, as the case it handles is now explicitly encoded in the AST.

### Motivation

Fixes #3647.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -695,10 +695,6 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
         if (spec.flags.isKeyword) {
             break;
         }
-        if (ait + 1 == aend && hasKwargs && (spec.flags.isDefault || spec.flags.isRepeated) &&
-            Types::approximate(gs, arg->type, *constr).derivesFrom(gs, Symbols::Hash())) {
-            break;
-        }
 
         auto offset = ait - args.args.begin();
         if (auto e = matchArgType(gs, *constr, core::Loc(args.locs.file, args.locs.call),
--- a/test/testdata/infer/kwargs.rb
+++ b/test/testdata/infer/kwargs.rb
@@ -14,3 +14,17 @@ end
 # Each repeated argument is checked as element type
 A.foo(x: 1, y: 2, z: '') # error: Expected `Integer` but found `String("")` for argument `kwargs`
 A.foo(x: 1, y: 2, z: 3)
+
+
+class B
+  extend T::Sig
+
+  sig {params(y: T::Hash[String,String], z: String).void}
+  def self.foo(y={}, z:)
+  end
+end
+
+# There is some tricky logic in dispatchCallSymbol to handle the case where a final positional hash argument is used for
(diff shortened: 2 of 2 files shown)


elliottt:

→ https://go/builds/bui_IP74aERbi2QqEW

jez (approved):

core/types/calls.cc
```diff
@@ -695,10 +695,6 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
         if (spec.flags.isKeyword) {
             break;
         }
-        if (ait + 1 == aend && hasKwargs && (spec.flags.isDefault || spec.flags.isRepeated) &&
-            Types::approximate(gs, arg->type, *constr).derivesFrom(gs, Symbols::Hash())) {
-            break;
-        }
```
jez, line 701: Does this logic need to be gated on the `--ruby3-keyword` flag or whatever we called it?

elliottt: No, It's just plain redundant now.

elliottt:
OK, this error existed before the change to inline keyword arguments in the send. Given how difficult it would be to model the runtime behavior of ruby in this case, I'm just going to close this PR and the issue associated.

PR #3676, opened 2020-11-17
`Pathname.glob` can take a `Pathname`.

reitermarkus:
### Motivation

```ruby
require "pathname"
Pathname.glob(Pathname.new("*.md"))
```

does not type-check currently, but runs as expected.

### Test plan

N/A


jez (approved):

jez:

→ https://go/builds/bui_IP7TweeXanhhT8

jez (approved):

PR #3678, opened 2020-11-17
Argument for `Pathname#children` is optional.

reitermarkus:
### Motivation

```ruby
require "pathname"
Pathname.pwd.children
```

does not type-check currently, but runs as expected.

### Test plan

N/A


jez:

→ https://go/builds/bui_IPO43Omc0ph5fX

jez (approved):

PR #3679, opened 2020-11-17
Delete dead code for final, abstract methods

jez:
### Motivation

I noticed this code was super redundant.

Check the commit history to see the before+after of the test.

### Test plan

See included automated tests.


elliottt (approved):
Good catch! I was just reading through this today, and wondering if we could consolidate the number of places we're raising errors about final methods.

PR #3680, opened 2020-11-17
Show nice type errors for all literals

jez:
### Motivation

I happened to see this code and wanted to change it.

Review by commit.

### Test plan

Check the commit history for a before+after.


test/cli/autocorrect-kwargs/autocorrect-kwargs.out
```diff
@@ -1,16 +1,102 @@
...
----
-> sig {returns(T::Array[{foo: Integer}])}
-18c18
-< foo = T::Array[foo: Integer, bar: String].new
----
-> foo = T::Array[{foo: Integer, bar: String}].new
-21c21
-< sig {params({a: Integer, b: String}).void}
----
-> sig {params(a: Integer, b: String).void}
```
jez, line 16: I converted this file to the new format in stages so that you can see what changed.

At the end of the day, the contents of the autocorrect is unchanged, but one of the old errors is gone.

Writing the test this way helped me debug it this test, because it failed at first.

Also, reviewing changes to autocorrects is easier this way: you only have to read a diff of the autocorrected file before and after the PR, whereas structured like a diff above, you have to read a diff of a diff to understand the change.

elliottt (approved):
Nice!

PR #3681, opened 2020-11-17
Show all method flags in symbol-table output

jez:
### Motivation

I chose a representation that wouldn't cause a ton of churn. Namely: if
you're a method symbol and you're public, we don't print that you're public
(this is what was already done). Also I chose to use the `|` separator
between flags because it reminds me of bitflags, which these flags actually
are under the hood.

### Test plan

I added some extra snapshot tests to showcase this a little better, you can see
the commit history for a before and after.


elliottt (approved):
Nice, this additional output is super helpful 👍

PR #3682, opened 2020-11-17
Dispatch <Magic>.<self-new> through new, not initialize

### LINKED ISSUE #3662: Intrinsic for self.new does not respect overrides of self.new
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Breturns(Integer)%7D%0A%20%20def%20self.new%3B%200%3B%20end%0A%0A%20%20T.reveal_type(new)%0Aend)

```ruby
# typed: true
class A
  extend T::Sig
  sig {returns(Integer)}
  def self.new; 0; end

  T.reveal_type(new)
end
```

#### Observed output

```
editor.rb:7: Revealed type: T.attached_class (of A) https://srb.help/7014
     7 |  T.reveal_type(new)
          ^^^^^^^^^^^^^^^^^^
  From:
    editor.rb:7:
     7 |  T.reveal_type(new)
                        ^^^
Errors: 1
```

#### Expected behavior

The revealed type should be `Integer`, as `new` has been overridden.

- - -

The `<Magic>.<self-new>` intrinsic (`Magic_selfNew`) dispatches directly to `initialize`, unconditionally.

Instead, it should dispatch to some method called `new_` on `self`, and just see where the resulting dispatch goes.

In that new world, the `<Magic>.<self-new>` intrinsic should only throw away the result type and replace it with `T.attached_class` if the simulated dispatch ended up dispatching to `Class#new`.

jez:
### Motivation

Fixes #3662.

### Test plan

Review by commit to see the before + after diff of the new test.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1355,6 +1355,15 @@ public:
         res.main.errors.clear();
         res.returnType = instanceTy;
         res.main = move(dispatched.main);
+        if (!res.main.method.exists()) {
+            // If we actually dispatched to some `initialize` method, use that method as the result,
+            // because it will be more interesting to people downstream who want to look at the
+            // result.
+            //
+            // But if this class hasn't defined a custom `initialize` method, still record that we
+            // dispatched to *something*, namely `Class#new`.
+            res.main.method = core::Symbols::Class_new();
+        }
         res.main.sendTp = instanceTy;
     }
 } Class_new;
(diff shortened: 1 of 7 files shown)


test/testdata/infer/attached_class_self_new.rb
```diff
@@ -94,3 +94,28 @@ def test
...
+class OverridesNewBad
+  extend T::Sig
+
+  sig {params(arg0: Integer).returns(Integer)}
+  def self.new(arg0); 0; end
+
+  def initialize; end
+
+  x = new
+  T.reveal_type(x) # error: Revealed type: `T.attached_class (of OverridesNewBad)`
```
jez: This is pretty 🙈

core/types/calls.cc
```diff
@@ -2129,33 +2138,20 @@ class Magic_selfNew : public IntrinsicMethod {
...
-                                   selfTy,        args.block, args.originForUninitialized};
-            dispatched = selfTy.dispatchCall(gs, innerArgs);
-            returnTy = dispatched.returnType;
-        } else {
-            // Otherwise, we know that this is the proper new intrinsic, and we
-            // should be returning something of type `T.attached_class`
+        DispatchArgs innerArgs{Names::new_(), sendLocs,   numPosArgs,
+                               sendArgStore,  selfTy,     selfTy,
+                               selfTy,        args.block, args.originForUninitialized};
+        auto dispatched = selfTy.dispatchCall(gs, innerArgs);
```
elliottt, line 2144: This is a really nice simplification -- going via the existing new intrinsic makes a lot of sense.

main/lsp/requests/hover.cc
```diff
@@ -89,8 +89,10 @@ unique_ptr<ResponseMessage> HoverTask::runRequest(LSPTypecheckerDelegate &typech
             if (constraint) {
                 retType = core::Types::instantiate(gs, retType, *constraint);
             }
-            if (sendResp->dispatchResult->main.method.exists() && sendResp->dispatchResult->main.method.isSynthetic()) {
-                // For synthetic methods, just show the return type
+            if (sendResp->dispatchResult->main.method.exists() &&
+                sendResp->dispatchResult->main.method.data(gs)->owner == core::Symbols::MagicSingleton()) {
+                // Most <Magic>.<foo> are not meant to be exposed to the user. Instead, just show
+                // the result type.
```
elliottt, line 95: What motivated removing the check for `isSynthetic`? Does that not overlap with the magic methods?

elliottt (approved):
I really like the simplification in calls.cc for handling `Magic.selfNew`!

main/lsp/requests/hover.cc
jez: See this commit:

https://github.com/sorbet/sorbet/pull/3682/commits/dc3d807ae030d5954e6e7a3f10f9fd5165afac06

and the attached commit message:

```
Fix failing LSP test

Now that we included Class#new in the dispatch result, a hover result
changed from `(nothing)` to something, but it looked bad, because
`Class#new` is entered as a "synthetic symbol" (i.e., any method entered
in GlobalState.cc).

That heuristic was too broad for what this code wanted to do, which is
show `{foo: Integer(0)}` when you hover on a hash instead of
`<Magic>.<build-hash>`, etc. (there are a lot `<Magic>.<build-*>`).

So instead, I check whether the owner is `T.class_of(<Magic>)` because
that's a better indication that the thing shouldn't be printed.
```

PR #3683, opened 2020-11-17
use bind_call on Ruby 2.7

froydnj:
This set started out as a "much elegant, such wow" idea, and then degraded as various real-world constraints were discovered.

The core idea is that wherever we use `.bind(x).call(...)` in `sorbet-runtime`, we should use `bind_call(x, ...)` in Ruby 2.7; such a substitution can be scripted.  Unfortunately, all the files in `sorbet-runtime` are run through Sorbet itself, and that creates problems with the 2.7-specific files (at least the ones that are typed), because:

1. `bind_call` doesn't exist on `UnboundMethod`; and
2. we can't use `T.absurd` for exhaustiveness-checking in non-typed files.

Both of these considerations lead to the hacks at the end of the patch series.

The good news is that if we wanted to make Ruby 2.7 the "official" supported version of Sorbet, the sense of the changes can be reversed (the 2.7 files become the canonical ones, and the 2.6 files are derivatives thereof) and the typing hacks can go away.  The alternative path is to just add `UnboundMethod#bind_call` in the appropriate rbi -- not sure of how much 2.7-specific stuff should be added to core rbi files, though.

### Motivation

`.bind_call` should theoretically be more efficient than `.bind(x).call(...)`, so we should use it where we can.

### Test plan

See included automated tests.


froydnj:
I pushed a `bind_call` rbi update so that this branch could reasonable be comparing apples to apples for a 2.6 vs. 2.7 comparison.

.buildkite/linters.sh
```diff
@@ -49,6 +49,13 @@ if [ "$err" -ne 0 ]; then
     globalErr=$err
 fi
 
+err=0
+./tools/scripts/add_27_bindcall.sh -c &> add_27_bindcall || err=$?
```
jez, line 53: do you think it makes sense to use bazel for this?

gems/sorbet-runtime/lib/types/compatibility_patches.rb_27.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
```
jez, line 1: Can we make sure to either add something in the file path and/or add a preamble to the file indicating which files are autogenerated and should not be edited manually?

froydnj: I think adding something in the file path is easier than trying to shoehorn it in after the sigil lines.  I will make them `$f_27_autogenerated.rb` instead.

jez: All of our other autogenerated files have preambles (e.g., every autogenerated RBI file).
We do this so we can show the command you need to run to regenerate it.

froydnj: OK, then I guess that wasn't an "or" ask. :)

.buildkite/linters.sh
froydnj: Maybe, but insofar as nothing else does...I'm not sure it's worth the trouble.

gems/sorbet-runtime/lib/types/compatibility_patches.rb_27.rb
jez: i guess in my mind the `or` was for the path name change (i don't mind if it mentions autogenerated in the path or not)😅

jez (approved):
This looks good to me in principle, but until we're sure that this will actually make code run faster I'd like to hold off merging, because it duplicates a fair amount of code that we don't test in CI.

gems/sorbet-runtime/lib/types/props/private/serde_transform_27.rb
```diff
@@ -0,0 +1,193 @@
...
+        end
+      end
+
+      # Guard against overrides of `name` or `to_s`
+      MODULE_NAME = T.let(Module.instance_method(:name), UnboundMethod)
+      private_constant :MODULE_NAME
+
+      sig {params(type: Module).returns(T.nilable(String)).checked(:never)}
+      private_class_method def self.module_name(type)
+        MODULE_NAME.bind_call(type)
```
djudd-stripe, line 189: This isn't hot - it's done once per serializable class - so we don't really need to duplicate the file for 2.7

jez:
Superceded by #4019

PR #3684, opened 2020-11-17
Remove TypeSyntax need for MutableContext.

jvilk-stripe:
Remove `TypeSyntax` need for `MutableContext`.

Previously, `TypeSyntax::getResultType` (and friends) needed `MutableContext` in order to call the following two methods:
* `singletonClass()`, which defines a singleton class if not present.
* `unsafeComputeExternalType()`, which computes an external type if not present.

The former can be replaced with `lookupSingletonClass` with no controversy, as `TypeSyntax::getResultType()` is never called with an unfrozen symbol table, so it should not cause new symbols / singleton classes to be defined.

The latter is tougher, and required some thought. Every symbol should have an `externalType()` defined after `Resolver::computeExternalTypes` runs, but `TypeSyntax::getResultType()` is called just before that in `ResolveTypeMembersWalk`.

During `ResolveTypeMembersWalk`, every class/module symbol that _does not_ have any type members has an `externalType` thanks to a pass over the symbol table in `finalizeSymbols()` that occurs earlier. That leaves one problem case: **When `ResolveTypeMembersWalk` needs the `externalType` of a type that has at least one type member**.

Looking through the code, `externalType()` is called in `TypeSyntax::getResultType()` under two scenarios:
* To compute `T.class_of(symbol)`
* To resolve constant literals.

It turns out that `ResolveTypeMembersWalk` already collects these symbols as `dependencies` of each `job`! All we have to do is make sure they have an external type defined before we perform the job, which is exactly what this change does.

Now, `TypeSyntax::getResultType()` is called in `ResolveTypeMembersWalk` to resolve the types of type members and type alias. I've added tests that cover all four possible combinations:

| | ConstantLit of a class w/ type param | T.class_of(class w/ type param on singleton) |
|--------------|---------------------------------|-----------------------------|
| `T.type_alias` | `T.type_alias{Parent[Integer]}` | `T.type_alias{T.class_of(A)}` |
| `T.type_member` | `type_member(:out, upper: Parent[Integer])` |`type_member(:in, upper: T.class_of(A))` |

I have confirmed that all four of these cases cause an exception when I remove the new `unsafeExternalType()` call in `resolver.cc`.

### Motivation

Unlocks parallelization opportunities in resolve signatures walk. Now we can call `TypeSyntax::getResultType` with a non-mutable context!

### Test plan

Covered by new tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1260,8 +1260,8 @@ class ResolveTypeMembersWalk {
         auto it = std::remove_if(job.dependencies.begin(), job.dependencies.end(), [&](core::SymbolRef dep) {
             if (isGenericResolved(ctx, dep)) {
                 if (dep.data(ctx)->isClassOrModule()) {
-                    // `dep`'s dependencies are resolved. Compute its externalType here so that TypeSyntax can access
-                    // it.
+                    // `dep`'s dependencies are resolved. Compute its externalType here so that we can resolve this
+                    // type member or type alias's type.
                     dep.data(ctx)->unsafeComputeExternalType(ctx);
                 }
                 return true;
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1257,13 +1257,21 @@ class ResolveTypeMembersWalk {
             return true;
         }
 
-        auto it = std::remove_if(job.dependencies.begin(), job.dependencies.end(),
-                                 [&](core::SymbolRef dep) { return isGenericResolved(ctx, dep); });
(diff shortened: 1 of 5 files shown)


jvilk-stripe:

→ https://go/builds/bui_IPPVE9maUq5xAd

jvilk-stripe:
As expected, there's a failure. Time to minimize, reproduce, and fix. :)

jvilk-stripe:

→ https://go/builds/bui_IPSuKlRkZYubpr

jvilk-stripe:
New stripe build is green. I also typechecked w/ a release debug build and it came out clean.

jez (approved):
This is a gorgeous change. Thanks for the explanation of the external type stuff.

PR #3685, opened 2020-11-17
Method renames: robustness in send expression handling

soam-stripe:
Add receiver location info to the SendResponse query response type; use that to find the method name in a send expression more reliably. This removes the adhoc search-and-replace based stuff that the previous change had for method renames. 

### Motivation

When renaming a method we have to update send expressions; unfortunately a send expression as returned in a SendResponse contains the entire `<receiver>.<method><args>` expression. So the previous code was trying to parse the receiver expression to skip over it to find the method; and it was doing so incorrectly.

### Test plan

See included automated tests.

### DIFF
--- a/core/lsp/QueryResponse.h
+++ b/core/lsp/QueryResponse.h
@@ -11,14 +11,15 @@ class TypeConstraint;
 class SendResponse final {
 public:
     SendResponse(core::Loc termLoc, std::shared_ptr<core::DispatchResult> dispatchResult, core::NameRef callerSideName,
-                 bool isPrivateOk, core::SymbolRef enclosingMethod)
+                 bool isPrivateOk, core::SymbolRef enclosingMethod, core::Loc receiverLoc)
         : dispatchResult(std::move(dispatchResult)), callerSideName(callerSideName), termLoc(termLoc),
-          isPrivateOk(isPrivateOk), enclosingMethod(enclosingMethod){};
+          isPrivateOk(isPrivateOk), enclosingMethod(enclosingMethod), receiverLoc(receiverLoc){};
     const std::shared_ptr<core::DispatchResult> dispatchResult;
     const core::NameRef callerSideName;
     const core::Loc termLoc;
     const bool isPrivateOk;
     const core::SymbolRef enclosingMethod;
+    const core::Loc receiverLoc;
 };
 
 class IdentResponse final {
--- a/core/lsp/QueryResponse.h
+++ b/core/lsp/QueryResponse.h
@@ -2,7 +2,6 @@
 #define SORBET_LSP_QUERYRESPONSE
 #include "core/Loc.h"
 #include "core/LocalVariable.h"
-#include "core/Symbols.h"
 #include "core/Types.h"
 #include <variant>
 
@@ -22,27 +21,7 @@ public:
     const core::SymbolRef enclosingMethod;
     const core::Loc receiverLoc;
 
-    const std::optional<core::Loc> getMethodNameLoc(const core::GlobalState &gs) const {
-        // TODO: handle dispatchResult->secondary
-        auto methodName = dispatchResult->main.method.data(gs)->name.show(gs);
-        auto expr = termLoc.source(gs);
-        // There are two possible forms of a send expression:
-        //   <receiver expr><whitespace?>.<whitespace?><method>
-        //   <method>
-        std::string::size_type methodNameOffset = receiverLoc.endPos() - termLoc.beginPos();
-        if (methodNameOffset != 0) {
-            methodNameOffset = expr.find_first_of(".", methodNameOffset) + 1;
-            methodNameOffset = expr.find_first_not_of(" \t", methodNameOffset);
-        }
-        auto offsets = termLoc.offsets();
-        offsets.beginLoc += methodNameOffset;
-        offsets.endLoc = offsets.beginLoc + methodName.length();
-        auto methodNameLoc = core::Loc(termLoc.file(), offsets);
-        if (offsets.endPos() > termLoc.endPos() || methodName != methodNameLoc.source(gs)) {
-            return std::nullopt;
-        }
-        return std::optional<core::Loc>(methodNameLoc);
-    }
+    const std::optional<core::Loc> getMethodNameLoc(const core::GlobalState &gs) const;
 };
 
 class IdentResponse final {
--- a/core/lsp/QueryResponse.h
+++ b/core/lsp/QueryResponse.h
@@ -2,6 +2,7 @@
 #define SORBET_LSP_QUERYRESPONSE
 #include "core/Loc.h"
 #include "core/LocalVariable.h"
+#include "core/Symbols.h"
 #include "core/Types.h"
 #include <variant>
 
@@ -20,6 +21,28 @@ public:
     const bool isPrivateOk;
     const core::SymbolRef enclosingMethod;
     const core::Loc receiverLoc;
+
+    const std::optional<core::Loc> getMethodNameLoc(const core::GlobalState &gs) const {
+        // TODO: handle dispatchResult->secondary
+        auto methodName = dispatchResult->main.method.data(gs)->name.show(gs);
+        auto expr = termLoc.source(gs);
+        // There are two possible forms of a send expression:
+        //   <receiver expr><whitespace?>.<whitespace?><method>
+        //   <method>
+        std::string::size_type methodNameOffset = receiverLoc.endPos() - termLoc.beginPos();
+        if (methodNameOffset != 0) {
+            methodNameOffset = expr.find_first_of(".", methodNameOffset) + 1;
+            methodNameOffset = expr.find_first_not_of(" \t", methodNameOffset);
+        }
+        auto offsets = termLoc.offsets();
+        offsets.beginLoc += methodNameOffset;
+        offsets.endLoc = offsets.beginLoc + methodName.length();
+        auto methodNameLoc = core::Loc(termLoc.file(), offsets);
+        if (offsets.endPos() > termLoc.endPos() || methodName != methodNameLoc.source(gs)) {
+            return std::nullopt;
+        }
+        return std::optional<core::Loc>(methodNameLoc);
+    }
 };
 
 class IdentResponse final {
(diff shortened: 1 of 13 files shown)


main/lsp/requests/prepare_rename.cc
```diff
@@ -20,6 +20,41 @@ variant<JSONNullObject, unique_ptr<PrepareRenameResult>> getPrepareRenameResult(
...
+    // The send expression is of the form <receiver>.<method><args>
+    // We want to return the range and placeholder for the method part of the expression, because this is what the
+    // editor will highlight in the rename UI.
+
+    // TODO(soam): handle dispatchResult->secondary
+    auto method = sendResp->dispatchResult->main.method;
+    if (!method.exists()) {
+        return JSONNullObject();
+    }
+    auto methodName = sendResp->dispatchResult->main.method.data(gs)->name.show(gs);
```
jvilk-stripe, line 35:
```suggestion
    auto methodName = method.data(gs)->name.show(gs);
```

main/lsp/requests/prepare_rename.cc
```diff
@@ -20,6 +20,41 @@ variant<JSONNullObject, unique_ptr<PrepareRenameResult>> getPrepareRenameResult(
...
+    auto expr = sendResp->termLoc.source(gs);
+    // find the end of the receiver, find the dot, and then the next non-whitespace char
+    string::size_type receiverOffset = sendResp->receiverLoc.endPos() - sendResp->termLoc.beginPos();
+    if (receiverOffset != 0) {
+        receiverOffset = expr.find_first_of(".", receiverOffset) + 1;
+        receiverOffset = expr.find_first_not_of(" \t", receiverOffset);
+    }
+    auto offsets = sendResp->termLoc.offsets();
+    offsets.beginLoc += receiverOffset;
+    offsets.endLoc = offsets.beginLoc + methodName.length();
```
jvilk-stripe, line 45: Is this always the case, btw? (Could the source text refer to a shorter name somehow...? Should we verify that the source text at `offsets` contains `methodName`?)

I don't know the answers to these questions, so maybe we'll find out in testing. :)

main/lsp/requests/rename.cc
```diff
@@ -111,67 +114,80 @@ unique_ptr<WorkspaceEdit> Renamer::buildEdit() {
...
+        if (auto sendResp = response->isSend()) {
+            auto receiverLoc = sendResp->receiverLoc;
+            ENFORCE(loc.contains(receiverLoc), "receiver expression not within send expression");
+            ENFORCE(loc.beginPos() == receiverLoc.beginPos(), "send expression doesn't start with receiver");
+            string::size_type methodNameOffset;
+            if (receiverLoc.endPos() == receiverLoc.beginPos()) {
+                // no receiver expression, method is start of loc
+                methodNameOffset = 0;
+            } else {
+                // find the method name: <receiver expression> <whitespace?> <dot> <whitespace?> <method name>
```
jvilk-stripe, line 138: This seems to be the same logic as in `prepare_rename.cc`. Should this be a helper method on `SendResponse`? (getMethodNameLoc()?) Then you could potentially even unit test this logic!

main/lsp/requests/rename.cc
```diff
@@ -212,22 +228,32 @@ variant<JSONNullObject, unique_ptr<WorkspaceEdit>> RenameTask::getRenameEdits(LS
...
 
-        for (auto &location : references) {
-            // Get text at location.
-            auto fref = config.uri2FileRef(gs, location->uri);
-            if (fref.data(gs).isPayload()) {
+        // We want location but also the type of the expression at that location; and for some expression types like
+        // sends, we need more than just a location, for parsing purposes (the send location is too broad and makes us
+        // parse too much).
+        // TODO(soam): do we need to dedup locations?
+        for (auto &response : queryResult.responses) {
```
jvilk-stripe, line 247: Note that `getReferencesToSymbol` prunes Send results from untyped files and dedupes locations. I suspect that you will want to dedupe locations. DSL passes like `attr_writer` are likely to trigger the dedupe logic as they insert synthetic AST nodes that blame to the same source line.

jvilk-stripe (approved):
Two requests:

* Move the logic for determining the method name loc into `SendResponse` so it can be shared between `prepare_rename.cc` and `rename.cc`.
* (Nit, can be added later) Add deduping for locations and filter out responses that we typically ignore (as per logic in `extractLocations.cc`). You could even change the `findReferencesToSymbol` and `extractLocations` methods so you don't have to copy+paste the logic.

main/lsp/requests/prepare_rename.cc
soam-stripe: Hmm, interesting. I'll add a check. I can think of `new` vs `initialize`, but we don't want to rename those anyway... which I also need to add a check for.

(But also today I learned about `alias_method`, I'm not sure how Sorbet handles it maybe that's relevant here.)

soam-stripe:
Addressed review comments, PTAL @jvilk-stripe

core/lsp/QueryResponse.h
```diff
@@ -11,14 +12,37 @@ class TypeConstraint;
...
-          isPrivateOk(isPrivateOk), enclosingMethod(enclosingMethod){};
+          isPrivateOk(isPrivateOk), enclosingMethod(enclosingMethod), receiverLoc(receiverLoc){};
     const std::shared_ptr<core::DispatchResult> dispatchResult;
     const core::NameRef callerSideName;
     const core::Loc termLoc;
     const bool isPrivateOk;
     const core::SymbolRef enclosingMethod;
+    const core::Loc receiverLoc;
+
+    const std::optional<core::Loc> getMethodNameLoc(const core::GlobalState &gs) const {
```
jvilk-stripe, line 25: You should define the method body in `QueryResponse.cc`. There's no need to define it in the header for inlining.

main/lsp/requests/prepare_rename.cc
```diff
@@ -20,6 +20,28 @@ variant<JSONNullObject, unique_ptr<PrepareRenameResult>> getPrepareRenameResult(
     result->placeholder = name;
     return result;
 }
+
+variant<JSONNullObject, unique_ptr<PrepareRenameResult>>
+getPrepareRenameResultForSend(const core::GlobalState &gs, const core::lsp::SendResponse *sendResp) {
```
jvilk-stripe, line 25: Nice.

main/lsp/requests/rename.cc
```diff
@@ -111,67 +114,73 @@ unique_ptr<WorkspaceEdit> Renamer::buildEdit() {
...
+            return def;
         }
+        return replaceAt(def, pos);
     }
-
-    string replaceMethodNameInDef(string def) {
-        return absl::StrReplaceAll(def, {{oldName, newName}});
+    string replaceAt(string input, string::size_type pos) {
+        auto suffixOffset = pos + oldName.length();
+        return input.substr(0, pos) + newName + input.substr(suffixOffset, input.length() - suffixOffset);
```
jvilk-stripe, line 171: Super micro optimization: This will reduce the number of intermediate string allocations (concat-ing string views).

```suggestion
    string replaceAt(string_view input, string::size_type pos) {
        auto suffixOffset = pos + oldName.length();
        return absl::StrCat(input.substr(0, pos), newName, input.substr(suffixOffset, input.length() - suffixOffset));
```

main/lsp/requests/rename.cc
```diff
@@ -111,67 +114,73 @@ unique_ptr<WorkspaceEdit> Renamer::buildEdit() {
...
+    MethodRenamer(const core::GlobalState &gs, const LSPConfiguration &config, const string oldName,
+                  const string newName)
+        : Renamer(gs, config, oldName, newName) {}
     ~MethodRenamer() {}
-    void rename(const core::GlobalState &gs, unique_ptr<Location> &location, core::FileRef fref) override {
-        auto loc = location->range->toLoc(gs, fref);
+    void rename(unique_ptr<core::lsp::QueryResponse> &response) override {
+        auto loc = response->getLoc();
         auto source = loc.source(gs);
+        auto location = config.loc2Location(gs, loc);
```
jvilk-stripe, line 127: Note: `loc2Location` can return `nullptr` if `loc` is not valid.

main/lsp/requests/rename.cc
```diff
@@ -111,67 +114,73 @@ unique_ptr<WorkspaceEdit> Renamer::buildEdit() {
...
+    ConstRenamer(const core::GlobalState &gs, const LSPConfiguration &config, const string oldName,
+                 const string newName)
+        : Renamer(gs, config, oldName, newName) {}
     ~ConstRenamer() {}
-    void rename(const core::GlobalState &gs, unique_ptr<Location> &location, core::FileRef fref) override {
-        auto loc = location->range->toLoc(gs, fref);
+    void rename(unique_ptr<core::lsp::QueryResponse> &response) override {
+        auto loc = response->getLoc();
         auto source = loc.source(gs);
+        auto location = config.loc2Location(gs, loc);
```
jvilk-stripe, line 183: Note: loc2Location can return nullptr if loc is not valid.

main/lsp/requests/rename.cc
```diff
@@ -212,22 +221,28 @@ variant<JSONNullObject, unique_ptr<WorkspaceEdit>> RenameTask::getRenameEdits(LS
...
         }
     } else {
-        renamer = make_unique<ConstRenamer>(originalName, newName);
+        renamer = make_unique<ConstRenamer>(gs, config, originalName, newName);
         symbolsToRename.push_back(symbol);
     }
 
     for (auto sym : symbolsToRename) {
-        vector<unique_ptr<Location>> references = getReferencesToSymbol(typechecker, sym);
+        // vector<unique_ptr<Location>> references = getReferencesToSymbol(typechecker, sym);
```
jvilk-stripe, line 229: I think we can remove this now :D

core/lsp/QueryResponse.h
```diff
@@ -2,6 +2,7 @@
 #define SORBET_LSP_QUERYRESPONSE
 #include "core/Loc.h"
 #include "core/LocalVariable.h"
+#include "core/Symbols.h"
```
jvilk-stripe, line 5: Move this include to the .cc too.

jvilk-stripe (approved):
I had some minor edge case notes. The main thing I'd like you to fix is not defining the `QueryResponse` method in the `.h` file. That file is included all over Sorbet, so minimizing its contents is good for compilation speeds.

PR #3687, opened 2020-11-18
Allow changing root object in autoloader

aisamanra:
Right now, the autoloader always installs root packages on `Object`; this allows you to pass a flag which can override this, although it defaults to `Object`.

### Motivation
Prep for packaged autoloader.

### Test plan
Included tests to show output with new flag.

See included automated tests.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -414,6 +414,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                     cxxopts::value<string>()->default_value("autoloader"));
     options.add_options("advanced")("autogen-registry-module", "Name of Ruby module used for autoloader registry",
                                     cxxopts::value<string>()->default_value("Opus::Require"));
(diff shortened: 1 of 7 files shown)


jez (approved):

main/options/options.cc
```diff
@@ -414,6 +414,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                     cxxopts::value<string>()->default_value("autoloader"));
     options.add_options("advanced")("autogen-registry-module", "Name of Ruby module used for autoloader registry",
                                     cxxopts::value<string>()->default_value("Opus::Require"));
+    options.add_options("advanced")("autogen-root-object", "Name of Ruby object on which root autoloads should be installed",
+                                    cxxopts::value<string>()->default_value("Object"));
```
ngroman, line 418: I'm pretty sure this is true, but double checking, would this would support a root obj with leading `::`?

aisamanra: Yup! All this does is stitch the string in without checking anything, so you could have qualified names or even a send here.

PR #3693, opened 2020-11-19
don't copy vectors when computing file hashes

froydnj:
When we called `pipeline::computeFileHashes` from the LSP indexer, we were copying the vector of files.  Let's not do that.

### Motivation

Epsilon performance improvements.

### Test plan

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -1289,7 +1289,7 @@ core::FileHash computeFileHash(shared_ptr<core::File> forWhat, spdlog::logger &l
 }
 }; // namespace
 
-void computeFileHashes(const vector<shared_ptr<core::File>> files, spdlog::logger &logger, WorkerPool &workers) {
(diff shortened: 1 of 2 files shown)


jez (approved):
You say epsilon, but this vector is tens of thousands of items long, and copying it requires incrementing atomic counters, so you might want to actually see if this makes LSP startup faster! cc @jvilk-stripe

jvilk-stripe:
Thanks for fixing!

froydnj:
> You say epsilon, but this vector is tens of thousands of items long, and copying it requires incrementing atomic counters, so you might want to actually see if this makes LSP startup faster!

That doesn't matter much for the VSCode case, right?  (At least not very much?)  It would only matter for a rapid start and stop of the LSP server?

jvilk-stripe:
fwiw I don't expect a measurable improvement in startup time. the overhead here is incrementing tens of thousands of counters on a single core which should be ~nothing. but it's a good improvement nonetheless.

PR #3698, opened 2020-11-20
Nest classes/modules inside eachother when generating hidden definitions

bss:
After #3565 got merged, sorbet would error out on auto-generated hidden definitions that contained private constants. This happened since hidden definitions would always write out the fully-qualified name of constants instead of opening up modules/classes and using the shortest possible name for the lookup.
E.g. given the following code:

```ruby
Object.const_set(
  :Foo,
  Module.new do |m_foo|
    m_foo.const_set(
      :Bar,
      Module.new do |m_bar|
        m_bar.const_set(:Baz, Module.new)
        m_bar.private_constant(:Baz)
        m_bar.include(m_bar.const_get(:Baz))
      end
    )
  end
)
```

hidden definitions would generate the following code:

```ruby
module Foo; end
module Foo::Bar
  include ::Foo::Bar::Baz
end
module Foo::Bar::Baz; end
```

which in certain cases would error out on the `include` above.

this PR changes the output of hidden definitions to:

```ruby
module Foo
  module Bar
    module Baz; end

    include Baz
  end
end
```

for the example above.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet/lib/hidden-definition-finder.rb
+++ b/gems/sorbet/lib/hidden-definition-finder.rb
@@ -265,7 +265,11 @@ class Sorbet::Private::HiddenMethodFinder
       fqn = real_name(my_klass)
       if fqn
         klass_str = String.new
-        klass_str << (Sorbet::Private::RealStdlib.real_is_a?(my_klass, Class) ? "class #{fqn}\n" : "module #{fqn}\n")
+        klass_str << (Sorbet::Private::RealStdlib.real_is_a?(my_klass, Class) ? "class #{name}\n" : "module #{name}\n")
+        children = serialize_constants(source_children, rbi_children, my_klass, my_klass_is_singleton, source_symbols, rbi_symbols)
+        if children != ""
+          klass_str << indent!(children)
+        end
         klass_str << includes.join("\n")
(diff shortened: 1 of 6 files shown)


jez:
I think i'm generally fine with this change but i'm gonna send it to our weekly team meeting and get back to you.

bss:
@jez all good, no rush. As mentioned in the other PR, there is a few workarounds for anybody running into the particular issue.

jez:
Some alternatives:

- don't check private constant literal references in RBIs
- don't generate a reference to a private constant if we can detect it would cause an error
- keep the `A::B::C` for class defs but shorten `A::B::C::D` to `D` if the nesting scope matches a prefix of the constant reference we are about to reference

bss:
I don't think the last alternative would work, e.g. if a class/module in the middle of the chain is private:
```
module A
  module B
    module C
      module D; end
      include D
    end
  end
  private_constant :B
end
```
cannot be written as:

```ruby
module A; end
module A::B; end # error on private constant ref
module A::B::C:D; end # error on private constant ref
module A::B::C # error on private constant ref
  include D
end
```

jez:
Superseded by #3767

PR #3699, opened 2020-11-20
Forbid making final method public in child

jez:
### Motivation

This would already raise:

```ruby
class Parent
  sig(:final) {void}
  def foo; end
end

class Child < Parent
  private :foo
end
```

(public in parent made private in child)

But if that were flipped around (private in parent, public in child),
for some reason that was allowed.

I don't think this should be asymmetrical. If something is final, it
should not be modifiable.

### Test plan

See included automated tests.


elliottt (approved):
Nice, this behavior makes sense to me 👍

PR #3701, opened 2020-11-20
store TypePtr's tag in the lower 8 bits and payload in the upper 56

froydnj:
This change is just like #3694, but for `TypePtr` instead.  We have to do things slightly differently because we have the "not inlined" bit to deal with.  So we store the tag in the lower 8 bits, which is sufficient space and still permits easy extraction with `movzx`.

### Motivation

The fastest code is the code you don't execute.

### Test plan

Existing automated tests.

### DIFF
--- a/core/TypePtr.h
+++ b/core/TypePtr.h
@@ -75,10 +75,10 @@ private:
 
     // We use a 0 to indicate not inlined so that nullptr (which has counter value 0) is naturally viewed as
     // 'inlined'.
-    static constexpr tagged_storage NOT_INLINED_MASK = 0x100;
+    static constexpr tagged_storage NOT_INLINED_MASK = 0x0100;
     // We use a 8-bit tag mask so that tags can be efficiently extracted via
     // movzx on x86.
-    static constexpr tagged_storage TAG_MASK = 0xFF;
+    static constexpr tagged_storage TAG_MASK = 0x00FF;
 
     static constexpr tagged_storage PTR_MASK = ~(NOT_INLINED_MASK | TAG_MASK);
 
--- a/core/TypePtr.h
+++ b/core/TypePtr.h
@@ -90,7 +90,9 @@ private:
     static tagged_storage tagValue(Tag tag, u4 inlinedValue) {
         auto val = tagToMask(tag);
 
-        // Store value into val.
+        // Store value into val.  It doesn't much matter where we put it in
+        // the upper 48 bits, but we put it in the uppermost 32 bits to
+        // ensure that retrieve it requires only a shift and no masking.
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_IQSwCiD3TbtI4p

core/TypePtr.h
```diff
@@ -73,30 +73,27 @@ class TypePtr final {
...
-    // Top bit indicates if value is inlined into pointer.
-    // We use a 0 in top bit to indicate not inlined so that nullptr (which has counter value 0) is naturally viewed as
+    // We use a 0 to indicate not inlined so that nullptr (which has counter value 0) is naturally viewed as
     // 'inlined'.
-    static constexpr tagged_storage NOT_INLINED_MASK = 0x8000000000000000;
-    static constexpr tagged_storage TAG_MASK = 0x7FFF000000000007;
+    static constexpr tagged_storage NOT_INLINED_MASK = 0x100;
+    // We use a 8-bit tag mask so that tags can be efficiently extracted via
+    // movzx on x86.
+    static constexpr tagged_storage TAG_MASK = 0xFF;
```
trevor-stripe, line 81: How do you feel about giving these leading `0`s, so that they look like 16-bit constants?
```suggestion
    static constexpr tagged_storage NOT_INLINED_MASK = 0x0100;
    // We use a 8-bit tag mask so that tags can be efficiently extracted via
    // movzx on x86.
    static constexpr tagged_storage TAG_MASK = 0x00FF;
```

core/TypePtr.h
```diff
@@ -73,30 +73,27 @@ class TypePtr final {
...
+        // Store the tag in the lower bits of the pointer, regardless of size.
+        return static_cast<tagged_storage>(tag);
     }
 
     static tagged_storage tagValue(Tag tag, u4 inlinedValue) {
         auto val = tagToMask(tag);
 
         // Store value into val.
-        val |= static_cast<tagged_storage>(inlinedValue) << 3;
+        val |= static_cast<tagged_storage>(inlinedValue) << 32;
```
trevor-stripe, line 94: Why store the inlined value in the upper 32 bits? Are there more efficient instructions if it's in the upper 32-bits vs shifted by 16? Could you add a comment here explaining the reasoning?

trevor-stripe (approved):
Looks good! Just one suggestion, and a request for a comment, both are non-blocking.

core/TypePtr.h
froydnj: No strong reason, but storing it in the upper 32 implies that retrieving it only requires a shift and no masking (since the right shift by 32 implicitly masks).  I'll add a comment.

PR #3703, opened 2020-11-20
remove Expression's vtable

froydnj:
`Expression` has a vtable for the sole purpose of having a virtual destructor.  But for every `Expression` subclass, we call the subclass's destructor directly through a pointer of the type of the subclass.  So the virtual destructor isn't buying us anything, except wasting space in every subclass with a vtable that's never going to be used.  Let's get rid of it.

This is perhaps not the safest change, because subclasses of `Expression` can still `new` and `delete` pointers to `Expression` with abandon, but...ideally no one will be doing that?  The `operator->` and `operator*` changes are aimed at trying to not have raw `Expression` pointers floating around, at least.  I guess we could `private`-ize `new`/`delete` on `Expression` and then redeclare it `public` on every subclass if we wanted to try and make things safe(r)?

### Motivation

Less space for Sorbet's AST representation is a good thing.

### Test plan

Existing automated tests should sufficient.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -6,7 +6,6 @@
 #include <utility>
 
 // makes lldb work. Don't remove please
-template class std::unique_ptr<sorbet::ast::Expression>;
 template class std::unique_ptr<sorbet::ast::ClassDef>;
 template class std::unique_ptr<sorbet::ast::MethodDef>;
 template class std::unique_ptr<sorbet::ast::If>;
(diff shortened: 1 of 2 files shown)


trevor-stripe (approved):
This is great! Let's just remove `Expression` completely in a follow-up PR 👍

froydnj:

→ https://go/builds/bui_IQUwwHRL2WHN3h

PR #3705, opened 2020-11-20
Upgrade docusaurus to fix dot-prop warning

jez:
```
cd website
yarn upgrade
```

### Motivation

~~Getting rid of that annoying nag that shows up on every push~~ security

### Test plan

I ran the server locally and clicked around a few pages.


elliottt (approved):
🥇

PR #3709, opened 2020-11-21
Add Sorbet::Private::Static::VERSION

jez:
### Motivation

This constant only exists statically, so it is in the
`Sorbet::Private::Static` namespace. Any attempt to read from this variable
will raise a `NameError` at runtime.

This is useful for testing things like sorbet.run, where you want to
check what version we're on (e.g., did the WASM redownload or am I still
testing a cached version).

### Test plan

See included automated tests.


elliottt (approved):
This is a great idea!

PR #3711, opened 2020-11-21
pack Instruction subclasses more tightly

froydnj:
The Itanium (Unix) C++ ABI has this interesting feature where if you have a class/struct `NotFull`:

```c++
struct NotFull {
  void *p;
  char c;
  // Seven bytes of unused space on 64-bit platforms.
};
```

and you inherit from `NotFull`, class members will, if possible, be allocated starting in the unused space immediately after `c`.

This situation is exactly the situation we have with `cfg::Instruction`: there are seven bytes after `isSynthetic` and, in many cases, we are not using those bytes to the fullest.  This patch rearranges class members to pack them into that empty space where possible.

It would be nice to pack `VariableUseSite` a little more tightly with other members in several places, but that is a much more involved change; this change merely shuffles members around and updates checks accordingly.

### Motivation

Less space is more better.

### Test plan

Existing automated tests.

### DIFF
--- a/cfg/Instructions.h
+++ b/cfg/Instructions.h
@@ -74,23 +74,23 @@ CheckSize(Alias, 24, 8);
 
 class SolveConstraint final : public Instruction {
 public:
-    std::shared_ptr<core::SendAndBlockLink> link;
     LocalRef send;
-    SolveConstraint(const std::shared_ptr<core::SendAndBlockLink> &link, LocalRef send) : link(link), send(send){};
+    std::shared_ptr<core::SendAndBlockLink> link;
+    SolveConstraint(const std::shared_ptr<core::SendAndBlockLink> &link, LocalRef send) : send(send), link(link){};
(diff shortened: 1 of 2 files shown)


cfg/Instructions.h
```diff
@@ -100,7 +100,7 @@ class Send final : public Instruction {
     virtual std::string toString(const core::GlobalState &gs, const CFG &cfg) const;
     virtual std::string showRaw(const core::GlobalState &gs, const CFG &cfg, int tabs = 0) const;
 };
-CheckSize(Send, 160, 8);
+CheckSize(Send, 144, 8);
```
elliottt, line 103: This will translate to a pretty huge savings on pay-server, nice find!

elliottt (approved):
Looking forward to to see the memory usage graph go down next week!

froydnj:

→ https://go/builds/bui_IQoY2KnqatoXgD

PR #3713, opened 2020-11-21
Introduce ConstSymbolData for const references to Symbols.

jvilk-stripe:
Introduce ConstSymbolData for const references to Symbols.

`const SymbolData` is trivially convertable to a `SymbolData`, which permits mutations to `Symbol`s even with a `const GlobalState&`!

This change locks that down so that this escape hatch is not viable. Instead of relying on `const SymbolData` to protect the symbol against mutation, introduce a new class called `ConstSymbolData`.

It includes an explicit `const_cast` to permit code that relied on this behavior until it can be fully fixed.

### Motivation

Prevent code from mutating symbol table with a `const GlobalState`.

### Test plan

Compiles and tests pass.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -180,28 +180,28 @@ SymbolData SymbolRef::dataAllowingNone(GlobalState &gs) const {
     }
 }
 
-const SymbolData SymbolRef::data(const GlobalState &gs) const {
+ConstSymbolData SymbolRef::data(const GlobalState &gs) const {
     ENFORCE_NO_TIMER(this->exists());
     return dataAllowingNone(gs);
 }
 
-const SymbolData SymbolRef::dataAllowingNone(const GlobalState &gs) const {
+ConstSymbolData SymbolRef::dataAllowingNone(const GlobalState &gs) const {
     switch (kind()) {
         case SymbolRef::Kind::ClassOrModule:
             ENFORCE_NO_TIMER(classOrModuleIndex() < gs.classAndModules.size());
-            return SymbolData(const_cast<Symbol &>(gs.classAndModules[classOrModuleIndex()]), gs);
+            return ConstSymbolData(gs.classAndModules[classOrModuleIndex()], gs);
         case SymbolRef::Kind::Method:
             ENFORCE_NO_TIMER(methodIndex() < gs.methods.size());
-            return SymbolData(const_cast<Symbol &>(gs.methods[methodIndex()]), gs);
+            return ConstSymbolData(gs.methods[methodIndex()], gs);
         case SymbolRef::Kind::Field:
             ENFORCE_NO_TIMER(fieldIndex() < gs.fields.size());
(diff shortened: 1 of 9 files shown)


core/Symbols.cc
```diff
@@ -1400,7 +1400,9 @@ vector<std::pair<NameRef, SymbolRef>> Symbol::membersStableOrderSlow(const Globa
     return result;
 }
 
-SymbolData::SymbolData(Symbol &ref, const GlobalState &gs) : DebugOnlyCheck(gs), symbol(ref) {}
+SymbolData::SymbolData(Symbol &ref, GlobalState &gs) : DebugOnlyCheck(gs), symbol(ref) {}
+
+ConstSymbolData::ConstSymbolData(const Symbol &ref, const GlobalState &gs) : DebugOnlyCheck(gs), symbol(ref) {}
```
jez, line 1405: So this is the magic, right? It's a type error to construct a `SymbolData` if you don't have a `const GlobalState&`?

resolver/resolver.cc
```diff
@@ -1418,7 +1419,8 @@ class ResolveTypeMembersWalk {
             // >   Boolean = T.let(nil, T.untyped)
             // > end
             if (data->isTypeAlias() && send->fun == core::Names::let()) {
-                data->resultType = core::Types::untypedUntracked();
+                // TODO(jvilk): Fix this.
+                const_cast<core::TypePtr &>(data->resultType) = core::Types::untypedUntracked();
```
jez, line 1423: 😱

resolver/resolver.cc
```diff
@@ -1418,7 +1419,8 @@ class ResolveTypeMembersWalk {
             // >   Boolean = T.let(nil, T.untyped)
             // > end
             if (data->isTypeAlias() && send->fun == core::Names::let()) {
-                data->resultType = core::Types::untypedUntracked();
+                // TODO(jvilk): Fix this.
```
jez, line 1422:
```suggestion
                // TODO(jvilk): This mutates the symbol table from an immutable context due to a const bug in Symbols.cc. DO NOT COPY THIS
```

jez (approved):

core/Symbols.cc
jvilk-stripe: The magic is that you must now have a non-const `GlobalState` to construct a `SymbolData`.

Previously, you could construct a `const SymbolData` with a `const GlobalState`. However, you can trivially copy a `const SymbolData` into a `SymbolData` and have mutable access to the symbol. It turns out that the compiler would do this transformation for you with the `auto` keyword, which is some bad stuff.

jez: So if we did `const auto symData` everywhere it would work?

jvilk-stripe: More-or-less. You could still accidentally copy it into a non-const `SymbolData` but we don't typically use `SymbolData` objects like that.

PR #3715, opened 2020-11-21
Fix some locs on T::Struct

jez:
### Motivation

There was a point during my work on #3666 where I was using the loc of `new` on a T::Struct to enter a positional arg into the Symbol table (🙈 it doesn't work that way anymore don't worry). But a side effect of that was that it made me realize that we had bad locs for this.

Pulling this out to land it anyways.

### Test plan

The test doesn't actually catch a regression, but I didn't feel like writing a
CLI test just for this. Here's a screenshot of the behavior on the provided
test case before and after:

**Before**

Screen Shot 2020-11-20 at 10 40 49 PM

**After**

Screen Shot 2020-11-20 at 10 41 02 PM


elliottt (approved):
🚢

PR #3717, opened 2020-11-21
Overloads should copy the visibility of the method they overload

jez:
### Motivation

I don't actually think this affects any real world, because I'm not sure that
we set private in any RBIs right now, and you can't write overloads in normal
files.

Pulling this out of #3666 to make for a smaller diff.

### Test plan

See included automated tests.


elliottt (approved):
👍

PR #3719, opened 2020-11-21
Introduce alternative naming for `--error-{white,black}-list` options

issyl0:
# Motivation

- This makes the naming of these options more inclusive, while not being a breaking change by removing the other ones entirely.

- I was going to go for `error-{allow,block}-list`, but I asked in the Sorbet Slack and `isolate-error-code` and `suppress-error-code` were suggested as better names that have easier to understand intent (they work for a single error code, but can be passed multiple times).

- This keeps the old names around so this doesn't have to be a breaking change, but switches docs and tests to use the new names.

### Test plan

See included automated tests. I also built Sorbet with these changes locally and tested the exclusivity of both sets of options with each other.

### Other

Relevant Sorbet Slack conversation (for however long the history lasts): https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1605983741147100. I'll mark this as not-draft when the tests pass - unfortunately, I couldn't figure out how to run them running locally.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -459,10 +459,22 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                "Errors not mentioned will be silenced. "
                                "This option can be passed multiple times.",
                                cxxopts::value<vector<int>>(), "errorCode");
+    options.add_options("dev")("isolate-error-code",
+                               "Error code to report. "
+                               "Errors not mentioned will be silenced. "
+                               "This option can be passed multiple times. "
+                               "An alternative to --error-white-list.",
+                               cxxopts::value<vector<int>>(), "errorCode");
     options.add_options("dev")("error-black-list",
                                "Error code to blacklist from reporting. "
                                "Errors mentioned will be silenced. "
                                "This option can be passed multiple times.",
+                               cxxopts::value<vector<int>>(), "errorCode");
+    options.add_options("dev")("suppress-error-code",
+                               "Error code to not report. "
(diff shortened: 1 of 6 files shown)


main/options/options.cc
```diff
@@ -459,11 +459,23 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                "Errors not mentioned will be silenced. "
                                "This option can be passed multiple times.",
                                cxxopts::value<vector<int>>(), "errorCode");
+    options.add_options("dev")("isolate-error-code",
+                               "Error code to report. "
+                               "Errors not mentioned will be silenced. "
+                               "This option can be passed multiple times. "
+                               "An alternative to --error-white-list.",
```
aprocter-stripe, line 466: I wonder if it's better to cut this line and say under the entry for `--error-white-list` something like `Deprecated, renamed to --isolate-error-code`?

issyl0: I flipped between both these ways of doing it. I will gladly flip back. 😄

aprocter-stripe: No strong preference, personally---it's an improvement either way. :)

issyl0: I pushed a new commit for this. I also don't know what was up with the failing help output test - I thought I had the indentation right as I copied it straight from the locally built sorbet `--help`. Maybe it'll work now I've changed it again! 🤷🏻

issyl0:
Sorry, but I have no time to work on this any more, so I'm going to close this. Thanks for the help here though. 🍨

PR #3723, opened 2020-11-23
redo tagging scheme for {Tree,Type}Ptr

froydnj:
This PR redoes the `TreePtr` changes and `TypePtr` changes from their respective PRs (the commits here are cherry-picked from master), but with fixes for Emscripten and one small typo fix.  The fix was obvious (5eb1900), figuring out how to debug it properly took waaaay too long.

### Motivation

The fastest code is the code you don't execute...but it helps if the replacement code is actually correct in all cases.

### Test plan

Existing automated tests.

### DIFF
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -180,13 +180,7 @@ public:
 
     Expression *get() const noexcept {
         auto val = ptr & PTR_MASK;
-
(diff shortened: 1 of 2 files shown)


elliottt (approved):
Thanks for debugging these!

PR #3725, opened 2020-11-24
Don't mutate Symbol mixins in incremental resolver.

jvilk-stripe:
Don't mutate Symbol mixins in incremental resolver.

### Motivation

[Watch an error flip on and off.](https://sorbet.run/#%23%20typed%3A%20true%0A%23%20frozen_string_literal%3A%20true%0A%0Amodule%20Base%0Aend%0A%0Amodule%20BaseWithMethod%0A%20%20include%20Base%0A%20%20def%20foo%0A%0A%20%20end%0Aend%0A%0Amodule%20OtherModuleWithMethod%0A%20%20def%20foo(i)%0A%0A%20%20end%0Aend%0A%0Aclass%20E%0A%20%20include%20BaseWithMethod%0A%20%20include%20OtherModuleWithMethod%0Aend%0A%0AE.new.foo(1)%20%23%20put%20a%20space%20in%20this%20comment%20once.%20then%20twice.%20then%20three%20times.%0A) This bug occurs because the incremental path is (incorrectly) recomputing mixins. However, we shouldn't be adding mixins at all on the incremental path.

We discovered this bug while trying to implement method rename on mixins.

### Test plan

See included automated tests.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -312,6 +312,8 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
             // Cancellation cannot occur during incremental namer.
             ENFORCE(result.hasResult());
             what = move(result.result());
+
+            // Required for autogen tests, which need to control which phase to stop after.
             if (opts.stopAfterPhase == options::Phase::NAMER) {
                 return what;
             }
@@ -320,6 +322,8 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
         {
             Timer timeit(gs.tracer(), "incremental_resolve");
             gs.tracer().trace("Resolving (incremental pass)...");
+            // TODO: We should eventually be able to freeze the symbol and name tables, but overloads currently pose
+            // a problem.
             core::UnfreezeSymbolTable symbolTable(gs);
             core::UnfreezeNameTable nameTable(gs);
 
@@ -327,6 +331,8 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
             // incrementalResolve is not cancelable.
             ENFORCE(result.hasResult());
             what = move(result.result());
+
+            // Required for autogen tests, which need to control which phase to stop after.
             if (opts.stopAfterPhase == options::Phase::RESOLVER) {
                 return what;
             }
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -312,6 +312,9 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
             // Cancellation cannot occur during incremental namer.
             ENFORCE(result.hasResult());
             what = move(result.result());
+            if (opts.stopAfterPhase == options::Phase::NAMER) {
+                return what;
+            }
         }
 
         {
@@ -324,6 +327,9 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
             // incrementalResolve is not cancelable.
             ENFORCE(result.hasResult());
             what = move(result.result());
+            if (opts.stopAfterPhase == options::Phase::RESOLVER) {
+                return what;
+            }
         }
     } catch (SorbetException &) {
         if (auto e = gs.beginError(sorbet::core::Loc::none(), sorbet::core::errors::Internal::InternalError)) {
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -465,7 +465,8 @@ private:
         }
     }
 
-    static bool resolveAncestorJob(core::MutableContext ctx, AncestorResolutionItem &job, bool lastRun) {
+    static bool resolveAncestorJob(core::MutableContext ctx, AncestorResolutionItem &job, bool lastRun,
+                                   bool isIncrementalResolver) {
         auto ancestorSym = job.ancestor->symbol;
         if (!ancestorSym.exists()) {
             return false;
@@ -527,7 +528,10 @@ private:
             }
         } else {
             ENFORCE(resolved.data(ctx)->isClassOrModule());
-            job.klass.data(ctx)->addMixin(resolved);
+            // Don't mutate mixins during incremental resolver.
+            if (!isIncrementalResolver) {
+                job.klass.data(ctx)->addMixin(resolved);
+            }
         }
 
         if (ancestorPresent) {
@@ -808,7 +812,7 @@ public:
     }
 
     static vector<ast::ParsedFile> resolveConstants(core::GlobalState &gs, vector<ast::ParsedFile> trees,
-                                                    WorkerPool &workers) {
+                                                    WorkerPool &workers, bool isIncrementalResolver) {
         Timer timeit(gs.tracer(), "resolver.resolve_constants");
         const core::GlobalState &igs = gs;
         auto resultq = make_shared<BlockingBoundedQueue<ResolveWalkResult>>(trees.size());
@@ -905,15 +909,15 @@ public:
                 // We try to resolve most ancestors second because this makes us much more likely to resolve everything
                 // else.
                 int origSize = todoAncestors.size();
-                auto it =
-                    remove_if(todoAncestors.begin(), todoAncestors.end(), [&gs](AncestorResolutionItem &job) -> bool {
-                        core::MutableContext ctx(gs, core::Symbols::root(), job.file);
-                        auto resolved = resolveAncestorJob(ctx, job, false);
-                        if (resolved) {
-                            tryRegisterSealedSubclass(ctx, job);
-                        }
-                        return resolved;
-                    });
+                auto it = remove_if(todoAncestors.begin(), todoAncestors.end(),
+                                    [&gs, isIncrementalResolver](AncestorResolutionItem &job) -> bool {
+                                        core::MutableContext ctx(gs, core::Symbols::root(), job.file);
+                                        auto resolved = resolveAncestorJob(ctx, job, false, isIncrementalResolver);
+                                        if (resolved) {
+                                            tryRegisterSealedSubclass(ctx, job);
+                                        }
+                                        return resolved;
+                                    });
                 todoAncestors.erase(it, todoAncestors.end());
                 progress = (origSize != todoAncestors.size());
                 categoryCounterAdd("resolve.constants.ancestor", "retry", origSize - todoAncestors.size());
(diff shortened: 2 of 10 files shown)


resolver/resolver.cc
```diff
@@ -2553,8 +2557,8 @@ void Resolver::sanityCheck(core::GlobalState &gs, vector<ast::ParsedFile> &trees
 
 ast::ParsedFilesOrCancelled Resolver::runIncremental(core::GlobalState &gs, vector<ast::ParsedFile> trees) {
     auto workers = WorkerPool::create(0, gs.tracer());
-    trees = ResolveConstantsWalk::resolveConstants(gs, std::move(trees), *workers);
-    computeLinearization(gs);
```
jvilk-stripe, line 2557: I removed linearization from the fast path. It should be a no-op as mixins should not change.

main/pipeline/pipeline.cc
```diff
@@ -312,6 +312,9 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
             // Cancellation cannot occur during incremental namer.
             ENFORCE(result.hasResult());
             what = move(result.result());
+            if (opts.stopAfterPhase == options::Phase::NAMER) {
```
jvilk-stripe, line 315: We set `stopAfterPhase` in autogen tests. Not respecting it in incremental mode causes new issues now that we aren't (accidentally) recomputing mixins.

DarkDimius: can we add a comment in source for this?

main/pipeline/pipeline.cc
```diff
@@ -324,6 +327,9 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
             // incrementalResolve is not cancelable.
             ENFORCE(result.hasResult());
             what = move(result.result());
+            if (opts.stopAfterPhase == options::Phase::RESOLVER) {
```
DarkDimius, line 330: ditto

resolver/resolver.cc
```diff
@@ -527,7 +528,10 @@ class ResolveConstantsWalk {
             }
         } else {
             ENFORCE(resolved.data(ctx)->isClassOrModule());
-            job.klass.data(ctx)->addMixin(resolved);
+            // Don't mutate mixins during incremental resolver.
+            if (!isIncrementalResolver) {
```
DarkDimius, line 532: can we make it idempotent by design rather than hardcoding it? e.g. test if it's already there.

resolver/resolver.cc
DarkDimius: can we enforce that?

resolver/resolver.cc
jvilk-stripe: Sure, but that'll require a vector crawl on every `addMixin`.

jvilk-stripe: I can try that out and measure the perf.

resolver/resolver.cc
jvilk-stripe: Sure. I can enforce that.

main/pipeline/pipeline.cc
jvilk-stripe: Sure. And as for testing autogen more specifically.... we hardcode it into the test harness right now so it stops after certain phases. I'm not in a mood to change that.

resolver/resolver.cc
DarkDimius: > Sure, but that'll require a vector crawl on every addMixin.
> I can try that out and measure the perf.

Alternatively, let's ENFORCE it so that we know it for sure.

test/testdata/lsp/fast_path/linearization.1.rbupdate
```diff
@@ -0,0 +1,26 @@
+# typed: true
```
jez, line 1: fwiw i looked back and it looks like this test has failed for at least the past 6 months (i was curious to know whether it was a recent regression, and it doesn't look like it). Nice find!

resolver/resolver.cc
jvilk-stripe: On second thought, I don't know if I can do this. Ancestor order matters. Testing if it is there isn't sufficient.

jvilk-stripe: e.g., the following case would be broken:

```
class Foo
  include Bar
  include Baz
  include Bar
end
```

In this scenario, `Bar`'s methods shadow `Baz` because it is included after `Baz`.

jvilk-stripe: Actually, I think we can do this. Symbols have a bit set when their mixins are linearized. If that bit is set, I can opt into the membership check. If the mixin isn't part of the vector, add it.

I wish we could just make addMixin a no-op when the symbol is linearized, but unfortunately we _do_ have to support mixing in classes to linearized symbols. This is how we support mixins on classes/modules defined in payload.

jvilk-stripe: I have this implemented and am currently debugging how/why we are still adding new mixins to classes during incremental resolver -- mixins which should have been added the first time through.

jvilk-stripe: Done.

main/pipeline/pipeline.cc
jvilk-stripe: Done.

main/pipeline/pipeline.cc
jvilk-stripe: Comment added.

resolver/resolver.cc
jvilk-stripe: Enforce added.

resolver/resolver.cc
```diff
@@ -2554,7 +2554,14 @@ void Resolver::sanityCheck(core::GlobalState &gs, vector<ast::ParsedFile> &trees
...
     trees = ResolveConstantsWalk::resolveConstants(gs, std::move(trees), *workers);
-    computeLinearization(gs);
+    // NOTE: Linearization does not need to be recomputed as we do not mutate mixins() during incremental resolve.
+    if (debug_mode) {
+        for (auto i = 1; i < gs.classAndModulesUsed(); i++) {
+            core::SymbolRef sym(gs, core::SymbolRef::Kind::ClassOrModule, i);
+            // If class is not marked as 'linearization computed', then we added a mixin to it since the last slow path.
+            ENFORCE_NO_TIMER(sym.data(gs)->isClassOrModuleLinearizationComputed(), sym.toString(gs));
+        }
+    }
```
jez, line 2564: slight preference to use the `DEBUG_ONLY(...)` macro here.

jvilk-stripe:

→ https://go/builds/bui_IS4gxZyVypBrnB

resolver/resolver.cc
```diff
@@ -2554,7 +2554,14 @@ void Resolver::sanityCheck(core::GlobalState &gs, vector<ast::ParsedFile> &trees
 ast::ParsedFilesOrCancelled Resolver::runIncremental(core::GlobalState &gs, vector<ast::ParsedFile> trees) {
     auto workers = WorkerPool::create(0, gs.tracer());
     trees = ResolveConstantsWalk::resolveConstants(gs, std::move(trees), *workers);
-    computeLinearization(gs);
+    // NOTE: Linearization does not need to be recomputed as we do not mutate mixins() during incremental resolve.
+    if (debug_mode) {
+        for (auto i = 1; i < gs.classAndModulesUsed(); i++) {
```
jez, line 2559: is this "all the classes and modules that had been added by this file" not "all symbols globally"?

jvilk-stripe: This is all symbols globally. We don't have a way to run symbol passes only over the subset of the symbol table that may have been impacted by the given files.

jez (approved):
This looks good to me but does this actually work?

Let's say we're running the fast path to check if we can, but the edit was to insert

```
include CompletelyNewModule
```

to a class in the file. Isn't this a completely normal edit that would fire an ENFORCE in debug mode?

jvilk-stripe:
> Let's say we're running the fast path to check if we can, but the edit was to insert
> `include CompletelyNewModule`
> to a class in the file. Isn't this a completely normal edit that would fire an ENFORCE in debug mode?

Discussed offline; this edit would perform a full end-to-end typecheck (it would not run incrementally).

jez:
jez:t-unsafe:  4 minutes ago
:thinky_face: i left a comment. i’m confused how this passes
jvilk:oregon-tree:  4 minutes ago
Let's say we're running the fast path to check if we can, but the edit was to insert
include CompletelyNewModule
that edit cannot and will not take the incremental path
jvilk:oregon-tree:  3 minutes ago
we will hash the new file, compare with the old hash, hash is different => slow path
jez:t-unsafe:  3 minutes ago
how will we compute the hash?
jvilk:oregon-tree:  3 minutes ago
1 sec
:acknowledge:
1

jez:t-unsafe:  3 minutes ago
i thought we hashed the contents of global state after running the fast path
jvilk:oregon-tree:  2 minutes ago
each file has a content hash
jvilk:oregon-tree:  2 minutes ago
https://github.com/sorbet/sorbet/blob/8e17b530/main/pipeline/pipeline.cc#L1288
given that edit, we will index and resolve the file in an empty global state (ignoring errors). then we hash the resulting global state.
GitHubGitHub
sorbet/sorbet
A fast, powerful type checker designed for Ruby. Contribute to sorbet/sorbet development by creating an account on GitHub.
jez:t-unsafe:  1 minute ago
i see
jvilk:oregon-tree:  1 minute ago
so the hash before the edit should be different because a symbol's mixins are different.
jvilk:oregon-tree:  1 minute ago
hence we take the slow path
jez:t-unsafe:  1 minute ago
so it’s not runIncremental that runs to compute the edit’s hash, it’s all of resolver, including computeLinearization ?
New
jvilk:oregon-tree:  < 1 minute ago
correct, on an empty fresh global state
jvilk:oregon-tree:  < 1 minute ago
so it's on a per-file basis.
jez:t-unsafe:  < 1 minute ago
:thumbsup: makes sense thanks!

resolver/resolver.cc
jez: do you think this debug mode check is going to make it like really long to debug on Stripe's codebase? should we make it respect skip_slow_enforce ?

jvilk-stripe: Nope. It's a linear walk of the symbol table checking a single bit per class. It should take a handful of ms I'd imagine.

jvilk-stripe:
No measurable impact on Stripe typechecking runtime, and it typechecks clean. Will make the DEBUG_ONLY adjustment now.

resolver/resolver.cc
jvilk-stripe: Done.

PR #3727, opened 2020-11-24
Method renames: Disallow attr methods, constructors

soam-stripe:
This change disallows attr_* method renames, as well as constructors. We can't rename attr_ methods because we don't yet rename fields. We only block prepare_rename when invoked at the definition, to keep it fast it succeeds for sends. The rename task fails when invoked from anywhere.

### Motivation

Correctness, it doesn't make sense to rename attr methods but not the field, and we don't do field renames yet.

### Test plan

See included automated tests.


jvilk-stripe (changes requested):
I think that we should give the user an intelligible error message when they actually try to rename `attr_reader`.

That is, `PrepareRename` can allow it (so "Rename Symbol" appears in menu), but then `Rename` constructs a descriptive error message.

Otherwise, the UX is "I tried to rename this and it didn't do anything" which leads users to think that Sorbet is broken.

main/lsp/requests/prepare_rename.cc
```diff
@@ -79,7 +89,12 @@ unique_ptr<ResponseMessage> PrepareRenameTask::runRequest(LSPTypecheckerDelegate
                 if (defResp->symbol.data(gs)->isClassOrModule()) {
                     response->result = getPrepareRenameResult(gs, defResp->symbol);
                 } else if (defResp->symbol.data(gs)->isMethod()) {
-                    response->result = getPrepareRenameResult(gs, defResp->symbol);
+                    if (PrepareRenameTask::validateDef(defResp->termLoc.source(gs))) {
```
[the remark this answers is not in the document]
soam-stripe: It goes through, asks the user for a new name, and then fails at rename.

main/lsp/requests/prepare_rename.cc
```diff
@@ -79,7 +89,12 @@ unique_ptr<ResponseMessage> PrepareRenameTask::runRequest(LSPTypecheckerDelegate
...
+                        response->result = getPrepareRenameResult(gs, defResp->symbol);
+                    } else {
+                        response->error = make_unique<ResponseError>((int)LSPErrorCodes::InvalidRequest,
+                                                                     "Unsupported method rename operation");
```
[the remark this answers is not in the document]
soam-stripe: "Rename Symbol" appears in the menu in any case. But when prepare_rename fails, we get a little "This element can't be renamed" pop up, not the error message :(  Screenshot below.  When rename fails, it's even worse, we get a "No result" pop up at the bottom right. But I think this is probably because we're returning JSONNullObject instead of a proper error message -- I'll change that to send a better error.

If we allow prepare_rename to go through the user will actually have to enter a new name and hit enter to get the error. OTOH we'll be able to give them a better explanation. 🤔 Thoughts?

Screen Shot 2020-11-24 at 2 45 04 PM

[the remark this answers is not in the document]
jvilk-stripe: >If we allow prepare_rename to go through the user will actually have to enter a new name and hit enter to get the error. OTOH we'll be able to give them a better explanation. 🤔 Thoughts?

The better explanation sounds like a better UX to me. It confirms to the user that Sorbet understood their request, and cannot support it at the moment. It will also save us debugging time when users encounter the limitation themselves; it's self-documenting.

soam-stripe:
Updated, ptal @jvilk-stripe? We now get the "Attribute renames are unsupported" error while renaming attr_ methods from anywhere.

main/lsp/requests/prepare_rename.cc
```diff
@@ -70,23 +69,30 @@ unique_ptr<ResponseMessage> PrepareRenameTask::runRequest(LSPTypecheckerDelegate
         // Note: Need to correctly type variant here so it goes into right 'slot' of result variant.
         response->result = variant<JSONNullObject, unique_ptr<PrepareRenameResult>>(JSONNullObject());
         auto &queryResponses = result.responses;
-        if (!queryResponses.empty()) {
-            auto resp = move(queryResponses[0]);
+
+        for (auto const &resp : queryResponses) {
```
jvilk-stripe, line 73: Why are you iterating through the responses? The first response should always be the one you want. Is there a case that this choice is broken for?

main/lsp/requests/prepare_rename.cc
```diff
@@ -70,23 +69,30 @@ unique_ptr<ResponseMessage> PrepareRenameTask::runRequest(LSPTypecheckerDelegate
...
+                break;
+            } else if (resp->isField()) {
+                // for attr methods we get multiple responses; ignore the field response
+                continue;
             }
         }
     }
     return response;
-}
+} // namespace sorbet::realmain::lsp
```
jvilk-stripe, line 96: This comment might've been accidentally added to the auto-formatter

main/lsp/requests/rename.cc
```diff
@@ -148,14 +168,24 @@ class MethodRenamer : public Renamer {
 
 private:
     string replaceMethodNameInDef(string def) {
-        // We can't do a simple string replace-all because of cases like:
+        const vector<string> unsupportedDefPrefixes{"attr_reader", "attr_accessor", "attr_writer"};
+        for (auto u : unsupportedDefPrefixes) {
+            if (absl::StartsWith(def, u)) {
+                invalid = true; // ensures the entire rename operation is blocked, not just this particular location
+                error = "Attribute method renames are unsupported.";
```
jvilk-stripe, line 175: WDYT about a message like:

```suggestion
                error = fmt::format("Sorbet does not support renaming `{}`s", u);
```

main/lsp/requests/rename.cc
```diff
@@ -279,51 +314,59 @@ unique_ptr<ResponseMessage> RenameTask::runRequest(LSPTypecheckerDelegate &typec
...
+    auto &queryResponses = result.responses;
+    if (queryResponses.empty()) {
+        return response;
+    }
 
-                if (defResp->symbol.data(gs)->isClassOrModule() || defResp->symbol.data(gs)->isMethod()) {
-                    if (isValidRenameLocation(defResp->symbol, gs, response)) {
-                        response->result = getRenameEdits(typechecker, defResp->symbol, params->newName);
-                    }
+    for (auto const &resp : queryResponses) {
```
jvilk-stripe, line 328: Ditto: Why are you iterating through the queryResponses?

main/lsp/requests/rename.cc
```diff
@@ -279,51 +314,59 @@ unique_ptr<ResponseMessage> RenameTask::runRequest(LSPTypecheckerDelegate &typec
...
+            return response;
+        } else if (resp->isField()) {
+            // we get duplicate responses, ignore field rename requests
+            continue;
         }
     }
 
     return response;
-}
+} // namespace sorbet::realmain::lsp
```
jvilk-stripe, line 370: Ditto about this comment

test/testdata/lsp/rename/method_attr.rb
```diff
@@ -2,9 +2,14 @@
 # frozen_string_literal: true
 
 class Foo
-  attr_reader :bar
+  attr_reader :foo
+#              ^ apply-rename: [A] newName: x invalid: true
```
jvilk-stripe, line 6: I think patrick added in the ability to test the error message, too!

jvilk-stripe (changes requested):
My only blocking comment is: Why are you iterating through query responses? That may cause some weird bugs.

soam-stripe:
👍  removed the iteration and improved the error message with your suggestion. 

ptal @jvilk-stripe

main/lsp/requests/prepare_rename.cc
```diff
@@ -64,35 +64,28 @@ unique_ptr<ResponseMessage> PrepareRenameTask::runRequest(LSPTypecheckerDelegate
...
-                break;
-            } else if (resp->isField()) {
-                // for attr methods we get multiple responses; ignore the field response
-                continue;
-            }
+    // An explicit null indicates that we don't support this request (or that nothing was at the location).
+    // Note: Need to correctly type variant here so it goes into right 'slot' of result variant.
+    response->result = variant<JSONNullObject, unique_ptr<PrepareRenameResult>>(JSONNullObject());
+    auto &queryResponses = result.responses;
+    auto resp = move(queryResponses[0]);
```
jvilk-stripe, line 74: You need to check that `!queryResponses.empty()` otherwise this will crash. It can be empty if you try to rename, say, a blank space.

main/lsp/requests/rename.cc
```diff
@@ -325,48 +325,41 @@ unique_ptr<ResponseMessage> RenameTask::runRequest(LSPTypecheckerDelegate &typec
...
-            // Sanity check the text.
-            if (islower(params->newName[0])) {
-                response->error = make_unique<ResponseError>((int)LSPErrorCodes::InvalidRequest,
-                                                             "Constant names must begin with an uppercase letter.");
-                return response;
-            }
-            if (isValidRenameLocation(constResp->symbol, gs, response)) {
-                getRenameEdits(typechecker, constResp->symbol, params->newName, response);
-            }
+    auto resp = move(queryResponses[0]);
```
jvilk-stripe, line 328: Ditto.

jvilk-stripe (changes requested):
LGTM, just fix the small issues I noted.

jvilk-stripe (approved):
Whoops, I meant to approve. Please do fix the two issues before merging.

main/lsp/requests/prepare_rename.cc
soam-stripe: 🤦

PR #3728, opened 2020-11-24
Fix signature for `Kernel.system`.

reitermarkus:
### Motivation

`Kernel.system` has some functionality which is not covered by the current signature at all.

Currently, out of these

```ruby
system('echo')
system('echo', err: File::NULL)
system('echo', out: :err)
system('echo', 'hello')
system('echo', 'hello', err: File::NULL)
system('echo', 'hello', out: :err)
system(['echo', 'echo'])
system(['echo', 'echo'], 'hello')
system(['echo', 'echo'], out: :err)
system(['echo', 'echo'], 'hello', out: :err)
system(({'VAR' => 'VAL'}), 'echo')
system(({'VAR' => 'VAL'}), 'echo', out: :err)
system(({'VAR' => 'VAL'}), ['echo', 'echo'])
system(({'VAR' => 'VAL'}), ['echo', 'echo'], out: :err)
system(({'VAR' => 'VAL'}), ['echo', 'echo'], 'hello')
system(({'VAR' => 'VAL'}), ['echo', 'echo'], 'hello', out: :err)
```

only

```ruby
system('echo')
system('echo', 'hello')
```

type-check. The fixed signature should cover all of these, however it also accepts some configurations which are wrong, i.e.

```ruby
system('echo', ['echo', 'echo'])
system(['echo', 'echo'], ['hello', 'hello'])
```

I'm not quite sure if there is a way to make the second argument depend on the first argument, i.e. only allow `[String, String]` as the second if the first is `T::Hash[String, T.nilable(String)]`.

### Test plan

n/a

### DIFF
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -2783,6 +2783,13 @@ module Kernel
   # ```
   #
   # See `Kernel.exec` for the standard shell.
-  sig { params(args: String).returns(T.any(NilClass, FalseClass, TrueClass)) }
-  def system(*args); end
+  sig do
+    params(
+      env: T.any(String, [String, String], T::Hash[String, T.nilable(String)]),
+      argv0: T.any(String, [String, String]),
+      args: String,
+      options: T.untyped,
+    ).returns(T.nilable(T::Boolean))
+  end
+  def system(env, argv0 = T.unsafe(nil), *args, **options); end
 end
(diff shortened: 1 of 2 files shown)


elliottt:
Could you add some tests for calling `Kernel.system` with this new signature in `test/testdata/rbi/kernel.rb`?

elliottt (changes requested):
The signature change makes sense, but adding some tests to make sure that expected uses are possible would be great.

elliottt:

→ https://go/builds/bui_IS5iftBlSiD2Rm

elliottt (approved):
Thanks!

We should hold off on merging this until the [stripe migration PR](https://go/ps/pull/285646) has also been merged.

PR #3729, opened 2020-11-24
`chown` arguments can be `nil`.

reitermarkus:
### Motivation

Documentation states that `owner` or `group` can be ignored by passing `nil` or `-1`.

### Test plan

n/a


elliottt:

→ https://go/builds/bui_IS3fINk8KiMgeR

elliottt (approved):
Thank you!

PR #3730, opened 2020-11-24
Support `include FileUtils`.

reitermarkus:
### Motivation

Previously e.g.

```ruby
class MyClass
  include FileUtils

  def update
    touch "#{__dir__}/#{__FILE__}"
  end
end
```

did not type-check.

See https://sorbet.run/#%23%20typed%3A%20true%0A%0Arequire%20'fileutils'%0A%0Aclass%20MyClass%0A%20%20include%20FileUtils%0A%0A%20%20def%20update%0A%20%20%20%20touch%20%22%23%7B__dir__%7D%2F%23%7B__FILE__%7D%22%0A%20%20end%0Aend%0A

### Test plan

n/a

### DIFF
--- a/rbi/stdlib/fileutils.rbi
+++ b/rbi/stdlib/fileutils.rbi
@@ -99,6 +99,8 @@ module FileUtils
   OPT_TABLE = T.let(T.unsafe(nil), T::Hash[T.untyped, T.untyped])
   VERSION = T.let(T.unsafe(nil), String)
 
+  module_function
+
   # Copies `src` to `dest`. If `src` is a directory, this method copies all its
   # contents recursively. If `dest` is a directory, copies `src` to `dest/src`.
   #
@@ -130,7 +132,7 @@ module FileUtils
       remove_destination: T.nilable(T::Boolean)
     ).returns(T::Array[String])
   end
-  def self.cp_r(src, dest, preserve: nil, noop: nil, verbose: nil, dereference_root: true, remove_destination: nil); end
+  def cp_r(src, dest, preserve: nil, noop: nil, verbose: nil, dereference_root: true, remove_destination: nil); end
 
   # remove files `list[0]` `list[1]`... If `list[n]` is a directory, removes its
   # all contents recursively. This method ignores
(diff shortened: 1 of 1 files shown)


elliottt:

→ https://go/builds/bui_IS3WsJqoFWBsWT

rbi/stdlib/fileutils.rbi
```diff
@@ -99,6 +99,8 @@ module FileUtils
   OPT_TABLE = T.let(T.unsafe(nil), T::Hash[T.untyped, T.untyped])
   VERSION = T.let(T.unsafe(nil), String)
 
+  module_function
```
elliottt, line 102: It looks like in the ruby source, `module_function` is applied to individual methods: are you sure that every method with a sig that we have in the RBI should be marked `module_function`?

elliottt (changes requested):
Looks great! Just one question about applying `module_function` to the whole module, vs individual methods.

rbi/stdlib/fileutils.rbi
vaporyhumo: `module_function` when called standalone does apply to all subsequent methods, just like `private` does in a class.
[apidoc documentation](https://apidock.com/ruby/Module/module_function)

trevor-stripe: Yes, my question was: should we apply `module_function` to all methods, given that in the fileutils.rb source in the standard library, they apply it on an individual basis.

vaporyhumo: I'm sorry I deleted last comment. Yes, I realized that was what you meant, and actually looking at fileutils.rb it seems that some methods are marked with `private_module_fuction`

reitermarkus: I had a quick glance and it doesn't seem like there are any methods in here which are marked with `private_module_fuction` in `fileutils.rb`.

SeekingMeaning:
I have checked, these are the methods without `module_function`:

- `collect_method`
- `commands`
- `have_option?`
- `options`
- `options_of`
- `private_module_function`

elliottt:
Closing this in favor of #3887.

PR #3731, opened 2020-11-24
Correct return type of Ruby Core Hash#any?

gmacdougall:
### Motivation

The core Hash rbi previously contained a method definition for Hash#any?
without any type definitions associated with it. This would lead to the
return value of Hash#any? to be T.untyped:

```
T.reveal_type({}.any?) # Revealed type: T.untyped https://srb.help/7014
```

This now provides types for all the potential invocations of Hash#any?

```
T.reveal_type({}.any?) # Revealed type: T::Boolean
```

### Test plan

This PR provides this update and a regression test in the Hash test data.

### DIFF
--- a/rbi/core/hash.rbi
+++ b/rbi/core/hash.rbi
@@ -308,7 +308,7 @@ class Hash < Object
   sig {returns(T::Boolean)}
   sig do
     params(
-        blk: T.proc.params(arg0: K, arg1: V).returns(BasicObject),
+        blk: T.proc.params(arg0: Elem).returns(BasicObject),
     )
     .returns(T::Boolean)
   end
--- a/rbi/core/hash.rbi
+++ b/rbi/core/hash.rbi
@@ -305,7 +305,15 @@ class Hash < Object
 
   # See also
   # [`Enumerable#any?`](https://docs.ruby-lang.org/en/2.6.0/Enumerable.html#method-i-any-3F)
-  def any?(*_); end
+  sig {returns(T::Boolean)}
+  sig do
+    params(
+        blk: T.proc.params(arg0: K, arg1: V).returns(BasicObject),
+    )
+    .returns(T::Boolean)
+  end
+  sig { params(pattern: T.untyped).returns(T::Boolean) }
+  def any?(pattern = nil, &blk); end
 
   # Removes all key-value pairs from *hsh*.
   #
--- a/test/testdata/rbi/hash.rb
+++ b/test/testdata/rbi/hash.rb
@@ -89,8 +89,11 @@ initial_hash.transform_values!.with_index do |v, i|
 end
 
 T.assert_type!({}.any?, T::Boolean)
-T.assert_type!({ a: 'ant', b: 'bear', c: 'cat' }.any?(/d/), T::Boolean)
-T.assert_type!(
-  { a: 'ant', b: 'bear', c: 'cat' }.any? { |key, value| key == value },
-  T::Boolean
-)
+T.reveal_type(T::Hash[Symbol, Integer].new.any? do |key, value| # error: Revealed type: `T::Boolean`
+  T.reveal_type(key) # error: Revealed type: `Symbol`
+  T.reveal_type(value) # error: Revealed type: `Integer`
+end)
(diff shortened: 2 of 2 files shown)


rbi/core/hash.rbi
```diff
@@ -303,10 +303,6 @@ class Hash < Object
   end
   def assoc(arg0); end
 
-  # See also
-  # [`Enumerable#any?`](https://docs.ruby-lang.org/en/2.6.0/Enumerable.html#method-i-any-3F)
-  def any?(*_); end
```
jez, line 308: There is in fact an override of `Enumerable#any?` in `Hash`:

https://ruby-doc.org/core-2.6.3/Hash.html#method-i-any-3F

https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/hash.c#L6422

Can you instead fix / add a signature for the override as it appears in `hash.rbi`?

Omitting this definition will cause things like the hidden-definitions RBI generation to find this override as hidden, and since that autogenerated RBI definition will not have a type, it will thwart this attempt to type it.

jez (changes requested):
^

jez (approved):

jez:

→ https://go/builds/bui_ISPqChwfJ2HHMF

gmacdougall:
No problem. Thanks for the review!

test/testdata/rbi/hash.rb
```diff
@@ -87,3 +87,10 @@
   T.assert_type!(v, Float)
   T.assert_type!(i, Integer)
 end
+
+T.assert_type!({}.any?, T::Boolean)
+T.assert_type!({ a: 'ant', b: 'bear', c: 'cat' }.any?(/d/), T::Boolean)
+T.assert_type!(
+  { a: 'ant', b: 'bear', c: 'cat' }.any? { |key, value| key == value },
+  T::Boolean
+)
```
jez, line 96:
```suggestion
T.reveal_type({ a: 'ant', b: 'bear', c: 'cat' }.any? do |key, value| # error: Revealed type: `T::Boolean`
  T.reveal_type(key) # error: Revealed type: `Symbol`
  T.reveal_type(key) # error: Revealed type: `String`
end)
T.reveal_type({ a: 'ant', b: 'bear', c: 'cat' }.any? do |(key, value)| # error: Revealed type: `T::Boolean`
  T.reveal_type(key) # error: Revealed type: `Symbol`
  T.reveal_type(key) # error: Revealed type: `String`
end)
```

rbi/core/hash.rbi
```diff
@@ -305,7 +305,15 @@ class Hash < Object
 
   # See also
   # [`Enumerable#any?`](https://docs.ruby-lang.org/en/2.6.0/Enumerable.html#method-i-any-3F)
-  def any?(*_); end
+  sig {returns(T::Boolean)}
+  sig do
+    params(
+        blk: T.proc.params(arg0: K, arg1: V).returns(BasicObject),
```
jez, line 311:
```suggestion
        blk: T.proc.params(arg0: Elem).returns(BasicObject),
```

test/testdata/rbi/hash.rb
jez: The second part of this test (using `|(key, value)|` fails until the above suggestion on the other file is applied.

jez (changes requested):

test/testdata/rbi/hash.rb
gmacdougall: The `reveal_type` assertions fail for me here as it's not able to correctly type key or value.

Example:

```
test/testdata/rbi/hash.rb:94: ERROR: Expected error of form:
94:   T.reveal_type(key) # error: Revealed type: `Symbol`
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Revealed type: `Symbol`
Found error:
94:   T.reveal_type(key) # error: Revealed type: `Symbol`
      ^^^^^^^^^^^^^^^^^^ error: Revealed type: `T.untyped`
```

gmacdougall: The new commit posted solves this issue for me. Thanks for that!

jez (approved):

PR #3732, opened 2020-11-24
Fix signature for `Module#define_method`.

reitermarkus:
### Motivation

When calling `define_method` inside of a class, it thinks the receiver the block binds to is `T.class_of(T.self_type)`.

See https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20def%20my_method%3B%20end%0A%0A%20%20define_method(%3Amy_method)%20%7B%0A%20%20%20%20my_method%0A%20%20%7D%0Aend%0A

### Test plan

n/a

### DIFF
--- a/rbi/core/module.rbi
+++ b/rbi/core/module.rbi
@@ -730,7 +730,7 @@ class Module < Object
   sig do
     params(
         arg0: T.any(Symbol, String),
-        blk: BasicObject,
+        blk: T.proc.bind(T.untyped).returns(T.untyped),
     )
     .returns(Symbol)
   end


elliottt:

→ https://go/builds/bui_IS3SVm11n2x3R7

elliottt (changes requested):
This signature introduces some problems, for example:

```ruby
# typed: true

class Foo
  def self.test(name)
    define_method "foo" do |val|
      T.reveal_type(val)
    end
  end
end
```

produces the following output:

```
test.rb:6: Revealed type: NilClass https://srb.help/7014
     6 |      T.reveal_type(val)
              ^^^^^^^^^^^^^^^^^^
  From:
    test.rb:5:
     5 |    define_method "foo" do |val|
                                    ^^^
Errors: 1
```

You can fix this somewhat by changing the type of `blk` to `Proc`, but then you get the following:

```
test.rb:6: Revealed type: BasicObject https://srb.help/7014
     6 |      T.reveal_type(val)
              ^^^^^^^^^^^^^^^^^^
  From:
    test.rb:5:
     5 |    define_method "foo" do |val|
                                    ^^^
Errors: 1
```

This is a little bit better, but still not quite right -- you'll still have to cast the arguments to the types you expect.

Ultimately, I think that what you'd like here is the ability to say that `define_method` takes a block that accepts any number of parameters, and we just don't have the ability to say that currently. There are some suggestions for how to address that on #1142, but we haven't settled on anything concrete yet.

aprocter-stripe:

PR #3733, opened 2020-11-25
Make `Only modules can be included` error appear in incremental resolver.

jvilk-stripe:
Make `Only modules can be included` error appear in incremental resolver. Also updates the error message to specify `include` or `extend`.

### Motivation

One fewer difference between resolver and incremental resolver.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -11,7 +11,6 @@
 #include "core/Names.h"
 #include "core/Types.h"
 #include "core/errors/internal.h"
-#include "core/errors/resolver.h"
 #include <string>
 
 template class std::vector<sorbet::core::TypeAndOrigins>;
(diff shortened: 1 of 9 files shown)


core/Symbols.cc
```diff
@@ -273,6 +274,18 @@ TypePtr ArgInfo::argumentTypeAsSeenByImplementation(Context ctx, core::TypeConst
 
 void Symbol::addMixin(const GlobalState &gs, SymbolRef sym) {
```
jez, line 275: Can you make this return either nothing or some error marked nodiscard, and handle reporting an error in the two places where `addMixin` is called? You will have a lot more information in scope to give a much better error message.

jez:
> * Is it OK that this error logic is located within `addMixin()`? There's no other clear place to put it, unless I made a resolver helper method called `addMixin`.

I left a suggestion for how to change this.

core/Symbols.cc
jvilk-stripe: What's `nodiscard`?

jvilk-stripe: Is there an existing place where we do this? `dispatchCall` defers reporting errors, but they aren't mutated w/ extra info once completed.

jvilk-stripe:
Done.

core/Symbols.cc
jvilk-stripe: We discussed online; `[[nodescript]]` is a function attribute in C++. I've made the requested change.

core/Symbols.cc
```diff
@@ -271,8 +272,22 @@ TypePtr ArgInfo::argumentTypeAsSeenByImplementation(Context ctx, core::TypeConst
...
 
-void Symbol::addMixin(const GlobalState &gs, SymbolRef sym) {
+unique_ptr<ErrorBuilder> Symbol::addMixin(const GlobalState &gs, SymbolRef sym) {
     ENFORCE(isClassOrModule());
+    // Note: Symbols without an explicit declaration may not have class or module set. They default to modules in
+    // GlobalPass.cc.
+    unique_ptr<ErrorBuilder> rv = nullptr;
+    if (sym.data(gs)->isClassModuleSet() && !sym.data(gs)->isClassOrModuleModule()) {
+        if (sym != core::Symbols::BasicObject()) {
+            if (auto e = gs.beginError(loc(), core::errors::Resolver::IncludesNonModule)) {
```
jez, line 282: This is the key Loc I wanted to be better. This shows the error in the file wherever the module is defined, not at each offending `include` line.

jez (approved):

core/Symbols.cc
jvilk-stripe: Do you want me to pass a loc into `addMixin`?

resolver/resolver.cc
```diff
@@ -527,7 +528,12 @@ class ResolveConstantsWalk {
             }
         } else {
             ENFORCE(resolved.data(ctx)->isClassOrModule());
-            job.klass.data(ctx)->addMixin(ctx, resolved);
+            if (!job.klass.data(ctx)->addMixin(ctx, resolved)) {
+                if (auto e = ctx.beginError(job.ancestor->loc, core::errors::Resolver::IncludesNonModule)) {
+                    e.setHeader("Only modules can be `{}`d, but `{}` is a class", job.isInclude ? "include" : "extend",
+                                resolved.data(ctx)->show(ctx));
```
jez, line 534:
```suggestion
                    e.setHeader("Only modules can be `{}`d, but `{}` is a class", job.isInclude ? "include" : "extend",
                                resolved.data(ctx)->show(ctx));
                    e.addErrorLine(resolved.data(ctx)->loc(), "`{}` defined as a class here",
                                   resolved.data(ctx)->show(ctx));
```

jez: image

jez (approved):

PR #3734, opened 2020-11-25
Don't force redefinition of fixed type members

Morriar:
### Motivation

With its current behaviour, Sorbet forces users to redefine every fixed type members used in one of the ancestors.

As shown in the example below, introducing a fixed type member "pollute" the class hierarchy and requires a lot of boilerplate code to be added in all the descendants.

Furthermore, since Sorbet implicitly copies the type member when it's not found locally, error messages point to the wrong place and one has to go up the hierarchy class by class to find the real introduction:

```ruby
# typed: true

class Base
  extend T::Generic

  MYPARAM = type_member(fixed: String)
end

class C1 < Base # error: Type MYPARAM declared by parent Base must be re-declared in C1
end

class C2 < C1 # error: Type MYPARAM declared by parent C1 must be re-declared in C2
end

class C3 < C2 # error: Type MYPARAM declared by parent C2 must be re-declared in C3
end

class C4 < C3 # error: Type MYPARAM declared by parent C4 must be re-declared in C3
end
```

This pull-request makes it optional to redefine a type member which is already fixed by an ancestor, then silencing the errors.

Now, the type member is implicitly copied with the fixed bound from the parent and marked as invariant:

```ruby
# typed: true

class Base
  extend T::Generic

  MYPARAM = type_member(fixed: String)
end

class C1 < Base
  MYPARAM = type_member(fixed: String) # Implicitly copied from Base
end

class C2 < C1
  MYPARAM = type_member(fixed: String) # Implicitly copied from C1
end

# ...
```

This PR also cleans all the unnecessary type members from the core/stdlib RBIs.

Note that this doesn't change anything to the behaviour of non-fixed type members.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/GlobalPass.cc
+++ b/resolver/GlobalPass.cc
@@ -45,21 +45,34 @@ bool resolveTypeMember(core::GlobalState &gs, core::SymbolRef parent, core::Symb
     core::NameRef name = parentTypeMember.data(gs)->name;
     core::SymbolRef my = sym.data(gs)->findMember(gs, name);
     if (!my.exists()) {
-        auto code =
-            parent == core::Symbols::Enumerable() || parent.data(gs)->derivesFrom(gs, core::Symbols::Enumerable())
-                ? core::errors::Resolver::EnumerableParentTypeNotDeclared
-                : core::errors::Resolver::ParentTypeNotDeclared;
+        // The type member for `my` is not manually defined in the module/class pointed by `sym`
+        if (parentTypeMember.data(gs)->isFixed()) {
+            // If the type member was fixed in the `parent`, we can safely copy it over in `sym`
+            // and make it invariant.
+            my = gs.enterTypeMember(sym.data(gs)->loc(), sym, name, core::Variance::Invariant);
+            my.data(gs)->setFixed();
+            auto parentType = parentTypeMember.data(gs)->resultType;
+            my.data(gs)->resultType = parentType;
+            typeAliases[sym.classOrModuleIndex()].emplace_back(parentTypeMember, my);
+            return true;
+        } else {
+            // If the type member was not fixed, this is an error, the type should be redefined manually.
+            // We insert a dummy invariant definition fixed on `T.untyped` so errors don't propagate.
+            auto code =
+                parent == core::Symbols::Enumerable() || parent.data(gs)->derivesFrom(gs, core::Symbols::Enumerable())
+                    ? core::errors::Resolver::EnumerableParentTypeNotDeclared
+                    : core::errors::Resolver::ParentTypeNotDeclared;
 
-        if (auto e = gs.beginError(sym.data(gs)->loc(), code)) {
-            e.setHeader("Type `{}` declared by parent `{}` must be re-declared in `{}`", name.show(gs),
-                        parent.data(gs)->show(gs), sym.data(gs)->show(gs));
-            e.addErrorLine(parentTypeMember.data(gs)->loc(), "`{}` declared in parent here", name.show(gs));
+            if (auto e = gs.beginError(sym.data(gs)->loc(), code)) {
+                e.setHeader("Type `{}` declared by parent `{}` must be re-declared in `{}`", name.show(gs),
+                            parent.data(gs)->show(gs), sym.data(gs)->show(gs));
+                e.addErrorLine(parentTypeMember.data(gs)->loc(), "`{}` declared in parent here", name.show(gs));
+            }
+            my = gs.enterTypeMember(sym.data(gs)->loc(), sym, name, core::Variance::Invariant);
+            auto untyped = core::Types::untyped(gs, sym);
+            my.data(gs)->resultType = core::make_type<core::LambdaParam>(my, untyped, untyped);
+            return false;
         }
-        my = gs.enterTypeMember(sym.data(gs)->loc(), sym, name, core::Variance::Invariant);
-        my.data(gs)->setFixed();
-        auto untyped = core::Types::untyped(gs, sym);
-        my.data(gs)->resultType = core::make_type<core::LambdaParam>(my, untyped, untyped);
-        return false;
     }
     const auto &data = my.data(gs);
     if (!data->isTypeMember() && !data->isTypeArgument()) {
--- a/test/testdata/resolver/fuzz_type_member_forget.rb
+++ b/test/testdata/resolver/fuzz_type_member_forget.rb
@@ -9,4 +9,4 @@ class DifferentArityChild < Parent # error: Type `TParent` declared by parent `P
   TChild = type_member(fixed: String)
 end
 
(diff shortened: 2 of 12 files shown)


resolver/GlobalPass.cc
```diff
@@ -45,21 +45,34 @@ bool resolveTypeMember(core::GlobalState &gs, core::SymbolRef parent, core::Symb
...
+                            parent.data(gs)->show(gs), sym.data(gs)->show(gs));
+                e.addErrorLine(parentTypeMember.data(gs)->loc(), "`{}` declared in parent here", name.show(gs));
+            }
+            my = gs.enterTypeMember(sym.data(gs)->loc(), sym, name, core::Variance::Invariant);
+            auto untyped = core::Types::untyped(gs, sym);
+            my.data(gs)->resultType = core::make_type<core::LambdaParam>(my, untyped, untyped);
+            return false;
         }
-        my = gs.enterTypeMember(sym.data(gs)->loc(), sym, name, core::Variance::Invariant);
-        my.data(gs)->setFixed();
```
Morriar, line 59: We don't fix the type member on `T.untyped` in case of error so we can differentiate between really fixed types and types added to fix missing type member redefinition.

The resulting behaviour is that Sorbet will also complain about missing type parameter when instantiating an incorrectly defined generic class.

test/testdata/resolver/fuzz_type_member_forget.rb
```diff
@@ -9,4 +9,4 @@ class DifferentArityChild < Parent # error: Type `TParent` declared by parent `P
   TChild = type_member(fixed: String)
 end
 
-T.cast(1, DifferentArityChild[Integer, String, Symbol]) # error-with-dupes: Wrong number of type parameters for `DifferentArityChild`. Expected: `0`, got: `3`
+T.cast(1, DifferentArityChild[Integer, String, Symbol]) # error-with-dupes: Wrong number of type parameters for `DifferentArityChild`. Expected: `1`, got: `3`
```
Morriar, line 12: The type was previously fixed to `T.untyped` because it was missing. Since we don't fix it anymore, the arity is now `1`.

test/testdata/resolver/redefinition_of_subclass_type_member.rb
```diff
@@ -28,4 +28,5 @@ class Bar < Foo # error: Type `V` declared by parent `Foo` must be re-declared i
   K = Bar[String,String].new.foo('a', 2)
 # ^ error: Type variable `K` needs to be declared as `= type_member(SOMETHING)`
     # ^^^^^^^^^^^^^^^^^^ error: Wrong number of type parameters
+                                    # ^ error: Expected `String` but found `Integer(2)` for argument `v`
```
Morriar, line 31: Since it's no more fixed on `T.untyped`.

test/testdata/resolver/type_member_missing.rb.symbol-table-raw.exp
```diff
@@ -8,8 +8,8 @@ class <C <U <root>>> < <C <U Object>> ()
     type-member(+) <S <C <U Base>> $1><C <U <AttachedClass>>> -> LambdaParam(<S <C <U Base>> $1><C <U <AttachedClass>>>, lower=T.noreturn, upper=AppliedType {   klass = <C <U Base>>   targs = [     <C <U Elem>> = T.untyped   ] }) @ Loc {file=test/testdata/resolver/type_member_missing.rb start=3:7 end=3:11}
     method <S <C <U Base>> $1><U <static-init>> (<blk>) @ Loc {file=test/testdata/resolver/type_member_missing.rb start=3:1 end=7:4}
       argument <blk><block> @ Loc {file=test/testdata/resolver/type_member_missing.rb start=??? end=???}
-  class <C <U Child>> < <C <U Base>> () @ Loc {file=test/testdata/resolver/type_member_missing.rb start=9:1 end=9:19}
-    type-member(=) <C <U Child>><C <U Elem>> -> LambdaParam(<C <U Child>><C <U Elem>>, fixed=T.untyped) @ Loc {file=test/testdata/resolver/type_member_missing.rb start=9:1 end=9:19}
+  class <C <U Child>>[<C <U Elem>>] < <C <U Base>> () @ Loc {file=test/testdata/resolver/type_member_missing.rb start=9:1 end=9:19}
+    type-member(=) <C <U Child>><C <U Elem>> -> LambdaParam(<C <U Child>><C <U Elem>>, lower=T.untyped, upper=T.untyped) @ Loc {file=test/testdata/resolver/type_member_missing.rb start=9:1 end=9:19}
```
Morriar, line 12: Ditto

DarkDimius:
The reason we have this behavior is the following: when you inherit generic classses you might have different(both larger and smaller, but it's the larger case that matters here) set of type members. Given that use site will need to know their order(e.g. to instantiate the generic), we need definition side to be explicit about order in which type members are.

Implicit copying makes it magic and thus will lead to surprises for users and implementation bugs.

Disclaimer: I'm no longer on the team and the current team opinion takes precedence over mine.

DarkDimius:
Upd: if this is specific to fixed one, this is actually fine

Morriar:
This is indeed specific to fixed type members and doesn't change how non-fixed ones are handled.

Non-fixed type members still need to be redeclared in each subclass as to specify their order.

elliottt:
Thanks for the suggestion! I agree that it's not great to have to rewrite the `type_member` declarations for fixed type members, but I don't see a way around it.

The `ResolveTypeParamsWalk` pass in resolver.cc walks the trees to resolve the bounds of each type member declared on a generic class. If a type member was added through a mechanism that doesn't trace directly back to the tree, then it won't get resolved and will remain a `<todo sym>`. You can see this in the following example:
```ruby
# typed: true
class A
  extend T::Sig
  extend T::Generic

  X = type_member(fixed: Integer)

  sig {returns(X)}
  def foo
    10
  end
end

class B < A
  extend T::Sig

  def self.bar
    new
  end
end

T.reveal_type(B.new.foo) # error: Revealed type: `<todo sym>`
T.reveal_type(B.bar) # error: Revealed type: `<todo sym>`
```
The `X` type member never being resolved will block the attached class for `B` from being resolved, as you can see in the second call to to `T.reveal_type`.

Separately, I think that there are potentially problems lurking when the `SymbolRef` stored in the resultType of the type_member doesn't point back to the class that the type_member belongs to, which will be the case when these type members are introduced in the GlobalPass. (The situation I can imagine is an error message that points the programmer back to the wrong file when referencing a usage of the implicitly defined type member on a subclass).

The sorbet team has talked about this PR, and we all generally agree that this is a syntax-motivated change which introduces a lot of edge cases that we don’t want to have to think about, in a part of the codebase that is already requires careful consideration when changes are made. This change makes that code harder to reason about.

We feel like the downsides outweigh the benefit here, especially given Sorbet’s design principles of being explicit rather than implicit.

aprocter-stripe:

PR #3735, opened 2020-11-25
don't check abstract methods on singleton classes

### LINKED ISSUE #3692: Duplicate error when using abstract, singleton, class << self methods
Sometimes I look at my issue titles and think "what in the world kind of nonsense is this" and this is definitely one of those.

#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.returns(Integer)%7D%0A%20%20def%20self.foo%3B%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20class%20%3C%3C%20self%0A%20%20%20%20extend%20T%3A%3ASig%0A%20%20end%0Aend)

```ruby
# typed: true
class Parent
  extend T::Sig
  extend T::Helpers
  abstract!

  sig {abstract.returns(Integer)}
  def self.foo; end
end

class Child < Parent
  class << self
    extend T::Sig
  end
end
```

#### Observed output

```
editor.rb:12: Missing definition for abstract method Parent.foo https://srb.help/5023
    12 |  class << self
          ^^^^^
    editor.rb:8: defined here
     8 |  def self.foo; end
          ^^^^^^^^^^^^

editor.rb:12: Missing definition for abstract method Parent.foo https://srb.help/5023
    12 |  class << self
          ^^^^^
    editor.rb:8: defined here
     8 |  def self.foo; end
          ^^^^^^^^^^^^
Errors: 2
```

#### Expected behavior

Ideally we only report one error. Not a huge deal though

- - -

This error happens because for each ClassDef we encounter, we validate whether that class implements all the required instance methods, and then repeat the logic for that class's singleton class.

But when using `class << self`, there is **another** ClassDef in the tree:

```
class ::Child<<C Child>> < (::Parent)
  def self.<static-init>(<blk>)
    begin
      ::<Magic>.<define-top-class-or-module>(::<Class:Child>)
      <emptyTree>
    end
  end
end
class <singleton class><<Class:Child>> < ()
  def self.<static-init>(<blk>)
    <self>.extend(::T::Sig)
  end
end
```

(this is post class flatten)

note that `class <singleton class>`. So what happens is we do

```ruby
# -- preTransformClassDef for first class def --
validateAbstract(Child)
validateAbstract(T.class_of(Child))

# -- preTransformClassDef for second class def --
validateAbstract(T.class_of(Child))
validateAbstract(T.class_of(T.class_of(Child)))
```

We can probably fix this by knowing that if the `ClassDef` name is `<singleton class>` then necessarily we will have already validated it once and we can just do `validateAbstract(T.class_of(T.class_of(Child)))`

froydnj:
When we are validating abstract methods, we shouldn't validate them if the classDef's symbol is a singleton class; we will have validated against the same class elsewhere, and we don't want to issue duplicate errors.

### Motivation

Fixes #3692 

### Test plan

See included automated tests.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -579,6 +579,8 @@ public:
         auto singleton = sym.data(ctx)->lookupSingletonClass(ctx);
         validateTStructNotGrandparent(ctx, sym);
         if (!sym.data(ctx)->isSingletonClass(ctx)) {
+            // Only validateAbstract for this class if we haven't already (we already have if this
+            // is a `class << self` ClassDef)
             validateAbstract(ctx, sym);
         }
         validateAbstract(ctx, singleton);
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -578,7 +578,9 @@ public:
         auto sym = classDef.symbol;
         auto singleton = sym.data(ctx)->lookupSingletonClass(ctx);
         validateTStructNotGrandparent(ctx, sym);
-        validateAbstract(ctx, sym);
+        if (!sym.data(ctx)->isSingletonClass(ctx)) {
+            validateAbstract(ctx, sym);
+        }
         validateAbstract(ctx, singleton);
         validateFinal(ctx, sym, classDef);
         validateSealed(ctx, sym, classDef);
--- a/test/testdata/duplicate_abstract_errors.rb
+++ b/test/testdata/duplicate_abstract_errors.rb
@@ -1,0 +1,15 @@
+# typed: true
+class Parent
+  extend T::Sig
+  extend T::Helpers
+  abstract!
+
+  sig {abstract.returns(Integer)}
+  def self.foo; end
+end
+
+class Child < Parent
+  class << self # error: Missing definition for abstract method `Parent.foo`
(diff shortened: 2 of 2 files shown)


definition_validator/validator.cc
```diff
@@ -578,7 +578,9 @@ class ValidateWalk {
         auto sym = classDef.symbol;
         auto singleton = sym.data(ctx)->lookupSingletonClass(ctx);
         validateTStructNotGrandparent(ctx, sym);
-        validateAbstract(ctx, sym);
+        if (!sym.data(ctx)->isSingletonClass(ctx)) {
+            validateAbstract(ctx, sym);
```
jez, line 582:
```suggestion
            // Only validateAbstract for this class if we haven't already (we already have if this
            // is a `class << self` ClassDef)
            validateAbstract(ctx, sym);
```

jez (approved):
Nice!

PR #3736, opened 2020-11-25
Typos in FAQ example, missing @ prefix

soam-stripe:
Fixes some typos in the attr_reader example from the FAQ.

### Test plan

none


trevor-stripe (approved):
Good find, thank you!

PR #3738, opened 2020-11-26
Fix double free heap corruption bug.

jvilk-stripe:
Fix double free heap corruption bug. We had code that passed a raw pointer from one `unique_ptr` to another `unique_ptr`, causing that pointer to be freed twice.

Changes `filterAndDedup` to filter and dedup in place, and to consume the input vector.

cc @soam-stripe 

### Motivation

I believe this will fix https://github.com/sorbet/sorbet/issues/3737

### Test plan

idk why CI didn't catch this before.

### DIFF
--- a/main/lsp/LSPTask.cc
+++ b/main/lsp/LSPTask.cc
@@ -199,24 +199,25 @@ bool LSPTask::canPreempt(const LSPIndexer &indexer) const {
     return !needsMultithreading(indexer);
 }
 
-// Filter for untyped locations, and dedup responses that are at the same location
+// Filter for untyped locations and dedup responses that are at the same location (inplace)
 vector<unique_ptr<core::lsp::QueryResponse>>
-LSPTask::filterAndDedup(const core::GlobalState &gs,
-                        const vector<unique_ptr<core::lsp::QueryResponse>> &queryResponses) const {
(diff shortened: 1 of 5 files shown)


main/lsp/LSPTask.cc
```diff
@@ -199,24 +199,25 @@ bool LSPTask::canPreempt(const LSPIndexer &indexer) const {
...
     // Filter for responses with a loc that exists and points to a typed file, unless it's a const, field or
     // definition in which case we're ok with untyped files (because we know where those things are even in untyped
     // files.)
-    for (auto &q : queryResponses) {
-        core::Loc loc = q->getLoc();
-        if (loc.exists() && loc.file().exists()) {
-            auto fileIsTyped = loc.file().data(gs).strictLevel >= core::StrictLevel::True;
-            // If file is untyped, only support responses involving constants and definitions.
-            if (fileIsTyped || q->isConstant() || q->isField() || q->isDefinition()) {
-                responses.push_back(make_unique<core::lsp::QueryResponse>(*q));
```
jvilk-stripe, line 216: This is what did it,

jvilk-stripe:
Actually, I just realized... that should invoke the copy constructor. 🤦

soam-stripe:
Maybe this PR is worth merging anyway for the in-place filter?

PR #3740, opened 2020-11-27
Add documentation for `T::Sig::WithoutRuntime.sig`

KaanOzkan:
Adds a brief documentation on `T::Sig::WithoutRuntime.sig` right after `.checked()`. It's placed here since at this point docs have mentioned why runtime checks are important and we already mentioned one way to disable runtime checks on a per sig basis. 

### Motivation
`T::Sig::WithoutRuntime.sig` is an important part of Sorbet and should be available for users to learn about in the docs similar to `checked(:never)`.

Screen Shot 2020-11-27 at 12 29 02 PM


elliottt (approved):
Thanks for documenting this!

PR #3742, opened 2020-11-30
Nudge people away from `# typed: strong`

jez:
### Motivation

We mention this in https://sorbet.org/docs/static but I think it's worth
reiterating in a bunch of places because people think they should be
using it still.

### Test plan

n/a


elliottt (approved):
👍

PR #3743, opened 2020-11-30
Don't try to dereference an empty optional object in hover.cc.

jvilk-stripe:
Don't try to dereference an empty optional object in hover.cc.

Also contains minor code cleanups.

### Motivation

Fixes https://github.com/sorbet/sorbet/issues/3737

Maybe it'll also fix some crashes we're seeing at Stripe.

### Test plan

I could not reproduce the bug locally. I could only reproduce in sorbet.run. :(

Manually verified that sorbet.run does not exhibit this problem with this fix.


main/lsp/requests/hover.cc
```diff
@@ -117,8 +117,8 @@ unique_ptr<ResponseMessage> HoverTask::runRequest(LSPTypecheckerDelegate &typech
         vector<string> documentation;
         for (auto loc : documentationLocations) {
             auto doc = findDocumentation(loc.file().data(gs).source(), loc.beginPos());
-            if (!doc->empty() && !(*doc).empty()) {
-                documentation.emplace_back(move(*doc));
+            if (doc.has_value() && !(*doc).empty()) {
```
jvilk-stripe, line 120: This is the fix. `doc->empty()` dereferences an empty optional and tries to call `string::empty()` 🤦

trevor-stripe (approved):

jez (approved):

PR #3744, opened 2020-11-30
Watchman: Ignore rsync tempfiles.

jvilk-stripe:
Watchman: Ignore rsync tempfiles.

### Motivation

Stripe uses rsync to keep development boxes in sync with user laptops. I noticed logs where watchman would overwhelm Sorbet with tmp files created during prolonged rsyncs.

### Test plan

We don't have tests that run Watchman in CI. I tested manually in a local folder, and verified that Sorbet could see updates to files outside of `.~tmp~` folders and could not see updates to files inside those folders.

### DIFF
--- a/main/lsp/watchman/WatchmanProcess.cc
+++ b/main/lsp/watchman/WatchmanProcess.cc
@@ -38,19 +38,22 @@ void WatchmanProcess::start() {
         // laptops have 4.9.0. Thus, we use [ "anyof", [ "suffix", "suffix1" ], [ "suffix", "suffix2" ], ... ].
         // Note 2: `empty_on_fresh_instance` prevents Watchman from sending entire contents of folder if this
         // subscription starts the daemon / causes the daemon to watch this folder for the first time.
-        string subscribeCommand = fmt::format("[\"subscribe\", \"{}\", \"{}\", {{\n"
-                                              "  \"expression\": [\"allof\", "
-                                              "    [\"type\", \"f\"],\n"
-                                              "    [\"anyof\", {}]"
-                                              "  ],\n"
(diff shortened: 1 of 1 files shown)


main/lsp/watchman/WatchmanProcess.cc
```diff
@@ -38,19 +38,22 @@ void WatchmanProcess::start() {
...
-                                                  return fmt::format("[\"suffix\", \"{}\"]", ext);
-                                              }));
+        string subscribeCommand = fmt::format(
+            "[\"subscribe\", \"{}\", \"{}\", {{\n"
+            "  \"expression\": [\"allof\", "
+            "    [\"type\", \"f\"],\n"
+            "    [\"anyof\", {}],\n"
+            "    [\"not\", [\"match\", \"**/.~tmp~/**\", \"wholename\", {{\"includedotfiles\": true}}]]\n" // Exclude
+                                                                                                           // rsync
+                                                                                                           // tmpfiles
```
jez, line 48: :shakes-fist: clang-format

Do you want to try seeing if clang-format is happier with this:

```suggestion
            // Exclude rsync tmpfiles
            "    [\"not\", [\"match\", \"**/.~tmp~/**\", \"wholename\", {{\"includedotfiles\": true}}]]\n"
```

jez (approved):

main/lsp/watchman/WatchmanProcess.cc
jvilk-stripe: Sure!

jvilk-stripe: That works much better, thanks.

PR #3745, opened 2020-12-01
WIP: Preallocate Symbol and Name tables during index.

jvilk-stripe:
Preallocate Symbol and Name tables during index.

In a multi-core environment, preallocating all of these tables to their maximal sizes imposes memory overhead that increases with the number of cores.

This change resizes the name and symbol tables only during the procedure that merges symbol and name tables together which should substantially reduce memory consumption on many core machines but may increase per-core runtime slightly due to resizing.

### Motivation

Reduce max RSS.

### Test plan

N/A

### DIFF
--- a/core/serialize/serialize.cc
+++ b/core/serialize/serialize.cc
@@ -760,35 +760,35 @@ void SerializerImpl::unpickleGS(UnPickler &p, GlobalState &result) {
 
         int classAndModuleSize = p.getU4();
         ENFORCE(classAndModuleSize > 0);
-        classAndModules.reserve(nearestPowerOf2(classAndModuleSize));
+        classAndModules.reserve(classAndModuleSize);
         for (int i = 0; i < classAndModuleSize; i++) {
             classAndModules.emplace_back(unpickleSymbol(p, &result));
         }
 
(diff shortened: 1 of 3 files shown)


DarkDimius:
nice!

jvilk-stripe:

→ https://go/builds/bui_IULEGaIyqIXnZx

jvilk-stripe:
This change didn't have the impact I expected.

First off, this change does not have any impact on warm cache typechecking memory consumption because we preallocate to the nearest power of 2 when deserializing GlobalState out of the cache, which happens prior to indexing. We have to do this for the name table for correctness (due to how namesByHash works), but we don't have to do this for the symbol table. In fact, resizing the symbol table to the nearest power of two is new behavior that I introduced when I split up the symbol table, and may be the root cause of the resulting memory increase! I'll test that hypothesis next. 

This change seems to have ~5% impact on the memory consumption of cold cache typechecking, which is less than I expected.

jvilk-stripe:

→ https://go/builds/bui_IUMXDLO633IMb4

jvilk-stripe:
Well, dang. None of this had any impact. I must be missing something. Closing for now.

PR #3746, opened 2020-12-01
Method renames: handle secondary dispatch components (union types)

### LINKED ISSUE #3665: Rename symbol fails to consider all components of a method call
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20def%20foo%3B%20end%0Aend%0Aclass%20B%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aextend%20T%3A%3ASig%0Asig%20%7Bparams(x%3A%20T.any(A%2C%20B)).void%7D%0Adef%20example(x)%0A%20%20x.foo%0A%23%20%20%20%5E%20apply-rename%3A%20%5BA%5D%20newName%3A%20bar%0Aend)

```ruby
# typed: true
class A
  def foo; end
end
class B
  def foo; end
end

extend T::Sig
def example(x)
  x.foo
#   ^ apply-rename: [A] newName: bar
end
```

#### Observed output

Making the indicated rename request in sorbet.run only applies the rename to the first component of the dispatch result.

Namely, it only renames `A#foo` → `A#bar`, and neglects the fact that the LSP query returned that the method call could have also been to `B#foo`.

#### Expected behavior

Either:

- We should rename each method in the dispatch result, or
- We should detect the case where there are multiple dispatch results, and abort (better than doing the wrong thing).

- - -

soam-stripe:
While searching for related symbols in method renames, handle multiple dispatch results in sends. When we find a secondary dispatch result, we follow the dispatch result linked list, adding any new methods we find. You can think of this code as the traversal of a graph where nodes are methods, and edges exist between methods that are related either by overriding or by being used in a union type.

### Motivation

Fixes #3665 -- the current code doesn't handle cases where there is more than one dispatch result for a send.

### Test plan

See included automated tests.


core/lsp/QueryResponse.cc
```diff
@@ -18,7 +18,8 @@ const SendResponse *QueryResponse::isSend() const {
 }
 
 const optional<core::Loc> SendResponse::getMethodNameLoc(const core::GlobalState &gs) const {
-    // TODO: handle dispatchResult->secondary
+    // We don't handle dispatchResult->secondary because we only need the method name; we assume here that the method
+    // name is the same for all dispatch components
     auto methodName = dispatchResult->main.method.data(gs)->name.show(gs);
     auto expr = termLoc.source(gs);
     // There are two possible forms of a send expression:
```
jez, line 25: There are actually a lot more... 😅

```
foo&.bar
foo[bar]
foo[bar]=qux
!foo
foo + bar
...
```

core/lsp/QueryResponse.cc
```diff
@@ -18,7 +18,8 @@ const SendResponse *QueryResponse::isSend() const {
 }
 
 const optional<core::Loc> SendResponse::getMethodNameLoc(const core::GlobalState &gs) const {
-    // TODO: handle dispatchResult->secondary
+    // We don't handle dispatchResult->secondary because we only need the method name; we assume here that the method
+    // name is the same for all dispatch components
     auto methodName = dispatchResult->main.method.data(gs)->name.show(gs);
```
jez, line 23:
```suggestion
    auto methodName = this->callerSideName.show(gs);
```

core/lsp/QueryResponse.cc
soam-stripe: Oof, thanks 😀 I'll update the comment but I think all these cases end up return nullopt due to the check a few lines below.

main/lsp/requests/rename.cc
```diff
@@ -224,6 +224,55 @@ class ConstRenamer : public Renamer {
...
+        auto included = false;
+        for (auto m : methods) {
+            if (m == method) {
+                included = true;
+                break;
+            }
+        }
+        if (!included) {
+            // this also adds `method` itself
+            addSubclassRelatedMethods(typechecker, method, methods);
```
jvilk-stripe, line 271: `addSubclassRelatedMethods` doesn't deduplicate. I'm worried about footguns here with respect to forgetting to dedupe in various places as we build out this functionality. I wonder if we should have a simple wrapper around the vector for this use case.

main/lsp/requests/rename.cc
```diff
@@ -236,32 +285,17 @@ void RenameTask::getRenameEdits(LSPTypecheckerDelegate &typechecker, core::Symbo
...
+        addSubclassRelatedMethods(typechecker, symbol, symbolsToRename);
     } else {
         renamer = make_unique<ConstRenamer>(gs, config, originalName, newName);
         symbolsToRename.push_back(symbol);
     }
 
-    for (auto sym : symbolsToRename) {
+    // We use symbolsToRename as a queue -- the loop body may add more symbols that need to be renamed, which is why we
+    // don't use an iterator here.
+    for (int i = 0; i < symbolsToRename.size(); i++) {
```
jvilk-stripe, line 296: Ideally, this would be a `std::deque` and we loop `while (!deque.empty())`. But we also need to make sure we don't process duplicate symbols.

Should we build a simple class wrapper that provides this functionality (e.g., with a `tryEnqueue` that returns a `bool` indicating if the thing was enqueued)? It could contain a simple `UnorderedSet<SymbolRef>` to track symbols that have already been enqueued.

main/lsp/requests/rename.cc
```diff
@@ -278,7 +312,17 @@ void RenameTask::getRenameEdits(LSPTypecheckerDelegate &typechecker, core::Symbo
                 responseMsg->result = JSONNullObject();
                 return;
             }
+
+            // The renamer de-duplicates edits at the same location
             renamer->rename(response);
+
+            // Handle secondary dispatch results (union types)
+            // TODO: move this into method renamer?
```
jvilk-stripe, line 320: Yeah, I think that would be a good idea.

jvilk-stripe (changes requested):
I had some suggestions for how to clean up the deduplication logic, and pointed out a place where we aren't deduping.

main/lsp/requests/rename.cc
soam-stripe: Yes this is a good idea, will do.

soam-stripe:
Addressed the review comments but I seem to have messed up the diff 😢 

I'll close this and open a new PR.

PR #3747, opened 2020-12-01
Rename methods in mixins

vilhena-stripe:
### Motivation
Allow renaming methods defined in mixins

### Test plan
See included automated tests.

### DIFF
--- a/main/lsp/requests/rename.cc
+++ b/main/lsp/requests/rename.cc
@@ -35,7 +35,7 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
     return true;
 }
 
-// Checks if s is a subclass or mixin of root, and updates visited and memoized vectors.
+// Checks if s is a subclass of root or contains root as a mixin, and updates visited and memoized vectors.
 bool isSubclassOrMixin(const core::GlobalState &gs, core::SymbolRef root, core::SymbolRef s, vector<bool> &memoized,
                        vector<bool> &visited) {
     // don't visit the same class twice
--- a/main/lsp/requests/rename.cc
+++ b/main/lsp/requests/rename.cc
@@ -35,21 +35,25 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
     return true;
 }
 
-// Checks if s is a subclass of root, and updates isSubclass, visited, and subclasses vectors.
-bool isSubclass(const core::GlobalState &gs, core::SymbolRef root, core::SymbolRef s, vector<bool> &memoized,
-                vector<bool> &visited) {
+// Checks if s is a subclass or mixin of root, and updates visited and memoized vectors.
+bool isSubclassOrMixin(const core::GlobalState &gs, core::SymbolRef root, core::SymbolRef s, vector<bool> &memoized,
+                       vector<bool> &visited) {
     // don't visit the same class twice
     if (visited[s.classOrModuleIndex()] == true) {
         return memoized[s.classOrModuleIndex()];
     }
     visited[s.classOrModuleIndex()] = true;
 
-    auto super = s.data(gs)->superClass();
-    if (super.exists()) {
-        memoized[s.classOrModuleIndex()] = isSubclass(gs, root, super, memoized, visited);
-    } else {
-        memoized[s.classOrModuleIndex()] = false;
+    for (core::SymbolRef a : s.data(gs)->mixins()) {
+        if (a == root) {
+            memoized[s.classOrModuleIndex()] = true;
+            return true;
+        }
     }
+    if (s.data(gs)->superClass().exists()) {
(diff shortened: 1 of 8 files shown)


main/lsp/requests/rename.cc
```diff
@@ -44,12 +44,24 @@ bool isSubclass(const core::GlobalState &gs, core::SymbolRef root, core::SymbolR
     }
     visited[s.classOrModuleIndex()] = true;
 
-    auto super = s.data(gs)->superClass();
-    if (super.exists()) {
-        memoized[s.classOrModuleIndex()] = isSubclass(gs, root, super, memoized, visited);
+    if (s.data(gs)->derivesFrom(gs, root)) {
```
jvilk-stripe, line 47: `derivesFrom` actually checks the superclass and mixins in the proper order.

I think we can actually consolidate this if/else into one loop so we can avoid doing the expensive `derivesFrom` check on every symbol and instead make use of our memoization.

main/lsp/requests/rename.cc
```diff
@@ -44,12 +44,24 @@ bool isSubclass(const core::GlobalState &gs, core::SymbolRef root, core::SymbolR
...
+        memoized[s.classOrModuleIndex()] = true;
     } else {
-        memoized[s.classOrModuleIndex()] = false;
+        auto super = s.data(gs)->superClass();
+        auto parents = s.data(gs)->mixins();
+        // Reverse parents so the last loaded mixin is first
```
jvilk-stripe, line 52: Mixins should already be in resolution order, although I'm not sure if order is important here.

main/lsp/requests/rename.cc
```diff
@@ -44,12 +44,24 @@ bool isSubclass(const core::GlobalState &gs, core::SymbolRef root, core::SymbolR
...
+        reverse(parents.begin(), parents.end());
+        parents.emplace_back(super);
+        for (auto p : parents) {
+            if (p.exists()) {
```
jvilk-stripe, line 56: You can assume mixins exist (but cannot assume that superclass exists).

main/lsp/requests/rename.cc
```diff
@@ -44,12 +44,24 @@ bool isSubclass(const core::GlobalState &gs, core::SymbolRef root, core::SymbolR
...
+                auto result = isSubclass(gs, root, p, memoized, visited);
+                memoized[s.classOrModuleIndex()] = result;
```
jvilk-stripe, line 58: I don't understand this logic. It seems like the last parent encountered determines if rename derives from a given class, but that's not true?

main/lsp/requests/rename.cc
jvilk-stripe: e.g., iterate thru all mixins until one `isSubclass` then check parent class

main/lsp/requests/rename.cc
```diff
@@ -77,13 +89,30 @@ core::SymbolRef findRootClassWithMethod(const core::GlobalState &gs, core::Symbo
...
+        auto super = root.data(gs)->superClass();
+        ENFORCE(super.exists()); // everything derives from Kernel::Object so we can't ever reach the actual top type
+
+        auto parents = root.data(gs)->mixins();
+        // Reverse parents so the last loaded mixin is first
+        reverse(parents.begin(), parents.end());
+        parents.emplace_back(super);
+
+        core::SymbolRef tmp;
+        for (auto p : parents) {
```
jvilk-stripe, line 101: You can just use `findMemberTransitive` which implements this logic for you.

main/lsp/requests/rename.cc
```diff
@@ -44,12 +44,24 @@ bool isSubclass(const core::GlobalState &gs, core::SymbolRef root, core::SymbolR
     }
     visited[s.classOrModuleIndex()] = true;
 
-    auto super = s.data(gs)->superClass();
```
jvilk-stripe, line 47: This whole function is named `isSubclass`, but that is no longer what you are checking for. Now you are checking if a class `derivesFrom` another class? We should rename it.

jvilk-stripe (changes requested):

main/lsp/requests/rename.cc
```diff
@@ -35,21 +35,25 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
     return true;
 }
 
-// Checks if s is a subclass of root, and updates isSubclass, visited, and subclasses vectors.
-bool isSubclass(const core::GlobalState &gs, core::SymbolRef root, core::SymbolRef s, vector<bool> &memoized,
-                vector<bool> &visited) {
+// Checks if s is a subclass or mixin of root, and updates visited and memoized vectors.
```
jvilk-stripe, line 38:
```suggestion
// Checks if s is a subclass of root or contains root as a mixin, and updates visited and memoized vectors.
```

jvilk-stripe (approved):

PR #3749, opened 2020-12-01
ActiveSupport::TestCase support through a rewriter

KaanOzkan:
Implemented a simple rewriter to rewrite calls to methods called `test`, with a single positional argument (String) and a block argument. These calls are converted into a method definition using the argument as a method name and block argument as the method body.

### Motivation

This rewriter is motivated by common usages of `ActiveSupport::TestCase` in Rails applications. As described [here](https://jonathanpike.net/2016/02/03/activesupport-vs-minitest), `ActiveSupport::TestCase` provides the means to declare a test case using a block format (via [`ActiveSupport::Testing::Declarative`](https://api.rubyonrails.org/classes/ActiveSupport/Testing/Declarative.html)). Since the block passed to the `test` method is called on the instance of the concrete test subclass, this was impossible to type externally.

With this rewriter we are able to expose the block to Sorbet in an instance method body and Sorbet can correctly typecheck the body. As a result there is a high chance that existing applications that were using `#typed: true` strictness in their `ActiveSupport::TestCase` test files will get new typechecking errors.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/TestCase.cc
+++ b/rewriter/TestCase.cc
@@ -12,11 +12,11 @@ std::vector<ast::TreePtr> TestCase::run(core::MutableContext ctx, ast::Send *sen
     }
 
     std::cout << "Checking method name" << std::endl;
-    if (send->fun != core::Names::test1()) {
+    if (send->fun != core::Names::test()) {
         return stats;
     }
 
-    std::cout << "Checking nunmber of args" << std::endl;
+    std::cout << "Checking number of args" << std::endl;
     if (send->args.size() != 1) {
         return stats;
     }
--- a/rewriter/TestCase.cc
+++ b/rewriter/TestCase.cc
@@ -1,0 +1,51 @@
+#include "rewriter/TestCase.h"
+#include "ast/Helpers.h"
+#include "core/GlobalState.h"
+
+namespace sorbet::rewriter {
+
+std::vector<ast::TreePtr> TestCase::run(core::MutableContext ctx, ast::Send *send) {
+    std::vector<ast::TreePtr> stats;
+    if (ctx.state.runningUnderAutogen) {
+        return stats;
+    }
+
+    std::cout << "Checking method name" << std::endl;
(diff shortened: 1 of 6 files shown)


rewriter/TestCase.cc
```diff
@@ -0,0 +1,47 @@
...
+
+    if (send->args.size() != 1) {
+        return stats;
+    }
+
+    if (send->numPosArgs != 1) {
+        return stats;
+    }
+
+    if (!send->block) {
```
jez, line 26:
```suggestion
    if (send->block == nullptr) {
```

rewriter/TestCase.cc
```diff
@@ -0,0 +1,47 @@
...
+        return stats;
+    }
+
+    auto *arg0 = ast::cast_tree<ast::Literal>(send->args[0]);
+    if (!arg0 || !arg0->isString(ctx)) {
```
jez, line 31:
```suggestion
    if (arg0 == nullptr || !arg0->isString(ctx)) {
```

jez:

→ https://go/builds/bui_IUfaW0TxX9UwMJ

jez (approved):
This looks good to me, pending a test run on Stripe's codebase.

I'm slightly concerned that we're adding more DSLs that fire on all sends for a method with a generic name (`test`) for something that isn't built into Ruby.

paracycle:
@jez I see that you have approved this, but also see that you have concerns about this being available by default. 

Is that the main reason why this PR has not been merged yet? If so, would it make sense to add this rewriter behind some kind of a flag so that it is not active by default?

jez:
Sorry, forgot about this. It was because of this:

> This looks good to me, pending a test run on Stripe's codebase.

I'll kick that off now.

jez:

→ https://go/builds/bui_ImldqFOnLTudff
→ https://go/builds/bui_ImldUQrPAqv5O8

PR #3750, opened 2020-12-02
Method renames: test for untyped files, operator overload renames

soam-stripe:
Block some invalid renames -- untyped files, and operator overloads.

### Motivation

We can't rename methods with sends in untyped files; it doesn't make much sense to rename operators.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/rename.cc
+++ b/main/lsp/requests/rename.cc
@@ -133,8 +133,20 @@ public:
     MethodRenamer(const core::GlobalState &gs, const LSPConfiguration &config, const string oldName,
                   const string newName)
         : Renamer(gs, config, oldName, newName) {
-        if (oldName == "initialize") {
-            invalid = true;
+        const vector<string> invalidNames = {"initialize", "call"};
+        for (auto name : invalidNames) {
+            if (oldName == name) {
+                invalid = true;
+                error = fmt::format("The `{}` method cannot be renamed.", oldName);
+                return;
+            }
+        }
+        // block any method not in /[a-zA-Z0-9_]+/. This blocks operator overloads.
+        for (auto c : oldName) {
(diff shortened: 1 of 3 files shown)


main/lsp/requests/rename.cc
```diff
@@ -133,8 +133,20 @@ class MethodRenamer : public Renamer {
...
-            invalid = true;
+        const vector<string> invalidNames = {"initialize", "call"};
+        for (auto name : invalidNames) {
+            if (oldName == name) {
+                invalid = true;
+                error = fmt::format("The `{}` method cannot be renamed.", oldName);
+                return;
+            }
+        }
+        // block any method not in /[a-zA-Z0-9_]+/. This blocks operator overloads.
```
jvilk-stripe, line 144: This will prevent renaming methods like:

```ruby
def has_something?
  true
end

# typically indicates that it mutates something in place
def do_something!

end
```

If this is just for operator overloads, can we just check the first character of `oldName`? I think the first character isn't alnum for those, and I don't think that would block most other renames.

jvilk-stripe (changes requested):

main/lsp/requests/rename.cc
jvilk-stripe: Alternatively, you could create a concrete list of operator `NameRefs`.

`Names_gen.h` contains things like `core::Names::unaryPlus()`, `core::Names::squareBrackets()`, etc.

jvilk-stripe: If you can pass a `NameRef` in as `oldName`, you can actually compare `NameRef`s (which are always identical if the string matches, and are simply integer comparisons).

jvilk-stripe (approved):

PR #3752, opened 2020-12-02
WIP: Break up name table into three vectors.

jvilk-stripe:
WIP: Break up name table into three vectors.

We have three types of names:

* UTF8Names (raw strings) (16 bytes)
  * Wraps a `string_view` (pointer + length)
* ConstantNames (4 bytes)
  * Wraps another Name that is used as a Constant.
* UniqueNames (12 bytes)
  * Wraps another Name with a counter. Used in a variety of circumstances.

Previously, we wrapped each of these in a 24 byte `Name` object. This object was 24 bytes because it stored a 16 byte union field (to accommodate a UTF8Name), a 1 byte Tag (to indicate the type of name), plus padding.

This change makes the name table (not including NamesByHash) _at least_ 33% smaller by making the largest Name type take up 16 bytes instead of 24 bytes. This memory savings is multiplied across the cores on your system, as we clone the Name table once per core during indexing.

I will need to break up this PR into manageable chunks to land it. I imagine I'll break it up like so:
* Introduce `asInteger()`/`asFloat()`/`asName()` methods to `LiteralType` and forbid direct access to its value.
* Change `NameRef` to store a `u4` rather than an `int`, and add `u4` to `LiteralType`'s union.
* Move methods from `Name` onto `NameRef`.
* Delete `Name` class and break up the symbol table.

### Motivation

Reduce memory consumption and maybe make some operations, like `GlobalState.deepClone()`, faster.

### Test plan

Covered by existing tests.


jvilk-stripe:
I should've opened this as a draft 🤦

jvilk-stripe:
Closing; I've landed most of this separately and am about to enter the final phase of breaking up that vector.

PR #3754, opened 2020-12-03
Add accessors to LiteralKind and make value/floatval private

jvilk-stripe:
Add accessors to LiteralKind. Makes it safer to extract the inner double/int/NameRef contained within.

### Motivation

Delurk lurky code that directly accesses LiteralType's value and floatval properties without the benefit of ENFORCEs checking that the literal is the expected kind of literal.

### Test plan

Covered by existing tests.

### DIFF
--- a/core/types/types.cc
+++ b/core/types/types.cc
@@ -334,6 +334,21 @@ LiteralType::LiteralType(SymbolRef klass, NameRef val)
     ENFORCE(klass == Symbols::String() || klass == Symbols::Symbol());
 }
 
+int64_t LiteralType::asInteger() const {
+    ENFORCE_NO_TIMER(literalKind == LiteralTypeKind::Integer);
+    return value;
+}
+
+double LiteralType::asFloat() const {
+    ENFORCE_NO_TIMER(literalKind == LiteralTypeKind::Float);
+    return floatval;
+}
+
+core::NameRef LiteralType::asName() const {
+    ENFORCE_NO_TIMER(literalKind == LiteralTypeKind::Symbol || literalKind == LiteralTypeKind::String);
+    return NameRef(NameRef::WellKnown{}, value);
+}
+
 TypePtr LiteralType::underlying() const {
     switch (literalKind) {
         case LiteralTypeKind::Integer:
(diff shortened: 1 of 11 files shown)


core/types/types.cc
```diff
@@ -334,6 +334,21 @@ LiteralType::LiteralType(SymbolRef klass, NameRef val)
...
+}
+
+double LiteralType::asFloat() const {
+    ENFORCE_NO_TIMER(literalKind == LiteralTypeKind::Float);
+    return floatval;
+}
+
+core::NameRef LiteralType::asName() const {
+    ENFORCE_NO_TIMER(literalKind == LiteralTypeKind::Symbol || literalKind == LiteralTypeKind::String);
+    return NameRef(NameRef::WellKnown{}, value);
```
jez, line 349: This seems wrong, doesn't it? shouldn't we pass a GlobalState from the call site?

jvilk-stripe: Sure, I can do that.

jez: (every call site this replaced seems to have a GlobalState in scope directly)

core/types/types.cc
```diff
@@ -367,15 +382,17 @@ void LiteralType::_sanityCheck(const GlobalState &gs) const {
 }
 
 bool LiteralType::equals(const LiteralType &rhs) const {
-    if (this->value != rhs.value) {
```
jez, line 370: oh no was this doing bitwise floating point comparison

jez (approved):
Looks good to me except the GlobalState comment.

core/types/types.cc
jvilk-stripe: SURE WAS

core/types/types.cc
trevor-stripe: The only place that would be complicated by that is the pickling code, as it used to just always encode the `u8` value of the union.

jvilk-stripe: Fixed. I did have to add an `unsafe` variant for the serializer callsite.

jez: Gotcha, so at that point, we could pass WellKnown, and write a comment about why it doesn't matter that we're using WellKnown for that call site.

PR #3755, opened 2020-12-03
Move most methods from Name to NameRef.

jvilk-stripe:
Move most methods from Name to NameRef.

Required to delete Name and break up the nametable into three vectors for each type of name to save memory.
Also reduces the amount of typing required to call these methods.

And, hey, it gets rid of the lurky `Name::ref()` method!

### Motivation

See above.

### Test plan

Covered by existing tests.


jez:
Did I miss a discussion of why we're making this change?

core/Names.h
```diff
@@ -91,27 +91,7 @@ class Name final {
 
     ~Name() noexcept;
 
-    bool operator==(const Name &rhs) const;
-
-    bool operator!=(const Name &rhs) const;
```
jvilk-stripe, line 96: These two operator overloads weren't actually defined anywhere.

jez (approved):

PR #3756, opened 2020-12-03
Make NameRef._id into an unsigned integer.

jvilk-stripe:
Make NameRef._id into an unsigned integer.

This is easy to do safely now that `LiteralType` hides its internal fields and has accessors.

### Motivation

This value is always positive.

### Test plan

Covered by existing tests.


core/NameRef.h
```diff
@@ -52,9 +52,9 @@ class NameRef final : private DebugOnlyCheck<NameRefDebugCheck> {
     // generated by tools/generate_names.cc
     struct WellKnown {};
 
-    constexpr NameRef(WellKnown, unsigned int id) : _id(id) {}
+    constexpr NameRef(WellKnown, u4 id) : _id(id) {}
```
elliottt, line 55:
```suggestion
    constexpr NameRef(WellKnown, u4 id) : _id{id} {}
```
Using braced initialization here means it would be harder for us to end up in a situation like we had before -- numeric conversions aren't allowed.

core/Names.cc
```diff
@@ -14,7 +14,7 @@ using namespace std;
 
 namespace sorbet::core {
 
-NameRef::NameRef(const GlobalState &gs, unsigned int id) : DebugOnlyCheck(gs, id), _id(id) {}
+NameRef::NameRef(const GlobalState &gs, u4 id) : DebugOnlyCheck(gs, id), _id(id) {}
```
elliottt, line 17:
```suggestion
NameRef::NameRef(const GlobalState &gs, u4 id) : DebugOnlyCheck(gs, id), _id{id} {}
```

elliottt (approved):
This is great, thank you for catching this!

PR #3757, opened 2020-12-03
Method renames: handle secondary dispatch components (union types)

### LINKED ISSUE #3665: Rename symbol fails to consider all components of a method call

soam-stripe:
While searching for related symbols in method renames, handle multiple dispatch results in sends. When we find a secondary dispatch result, we follow the dispatch result linked list, adding any new methods we find. You can think of this code as the traversal of a graph where nodes are methods, and edges exist between methods that are related either by overriding or by being used in a union type.

(previous PR: #3746)

### Motivation

Fixes #3665 -- the current code doesn't handle cases where there is more than one dispatch result for a send.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/rename.cc
+++ b/main/lsp/requests/rename.cc
@@ -58,8 +58,7 @@ bool isSubclassOrMixin(const core::GlobalState &gs, core::SymbolRef root, core::
 }
 
 // Returns all subclasses of root (including root)
-vector<core::SymbolRef> getSubclasses(LSPTypecheckerDelegate &typechecker, core::SymbolRef root) {
-    const core::GlobalState &gs = typechecker.state();
+vector<core::SymbolRef> getSubclasses(const core::GlobalState &gs, core::SymbolRef root) {
     vector<bool> memoized(gs.classAndModulesUsed());
     vector<bool> visited(gs.classAndModulesUsed());
     memoized[root.classOrModuleIndex()] = true;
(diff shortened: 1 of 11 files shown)


main/lsp/requests/rename.cc
```diff
@@ -91,6 +90,74 @@ core::SymbolRef findRootClassWithMethod(const core::GlobalState &gs, core::Symbo
     return root;
 }
 
+class UniqueSymbolQueue {
+public:
+    UniqueSymbolQueue() : iter(0) {}
+
+    bool tryEnqueue(core::SymbolRef s) {
+        auto insertResult = set.insert(s);
+        bool isNew = insertResult.second;
```
jvilk-stripe, line 99: TIL this returns pair that indicates if insertion added an element!

main/lsp/requests/rename.cc
```diff
@@ -91,6 +90,74 @@ core::SymbolRef findRootClassWithMethod(const core::GlobalState &gs, core::Symbo
...
+            auto result = symbols[iter];
+            iter++;
+            return result;
+        }
+        return core::SymbolRef();
+    }
+
+private:
+    int iter;
+    vector<core::SymbolRef> symbols;
```
jvilk-stripe, line 117: You can actually make this a `std::deque` so you can push things on the end and pop things off the front (or vice-versa). Then you don't need `iter`.

main/lsp/requests/rename.cc
```diff
@@ -91,6 +90,74 @@ core::SymbolRef findRootClassWithMethod(const core::GlobalState &gs, core::Symbo
...
+        auto insertResult = set.insert(s);
+        bool isNew = insertResult.second;
+        if (isNew) {
+            symbols.emplace_back(s);
+        }
+        return isNew;
+    }
+
+    core::SymbolRef pop() {
+        if (iter < symbols.size()) {
```
jvilk-stripe, line 107: With a deque, this becomes `if (!symbols.empty())`

test/testdata/lsp/rename/method_nilable.rb
```diff
@@ -0,0 +1,18 @@
+# typed: true
+# frozen_string_literal: true
+
+class C1
+  def m1
+#      ^ apply-rename: [A] newName: m2
```
jez, line 6: Nice! Super cool to see that `c&.m1` is renamed below.

jez (commented):
i only did a cursory look. excited to see this bug fixed, and i'll let john finish the review

jvilk-stripe (approved):
I had a minor suggestion that will clean up the UniqueSymbolQueue.

PR #3758, opened 2020-12-03
Respect orphaned visibility method invocations

AlexEvanczuk:
# Explanation of Change
Today, methods declared private/public/protected via  'orphaned' visibility method invocations (i.e. we invoke the visibility via the method invocation and then declare the method after) are not supported in sorbet.

```ruby
class A
  private
  def foo; end
end

A.new.foo # This does not throw an error
```

We first began investigating this by having some fancy work during the parsing step, where during the ast walk we kept track of the visibility in the current context, and then intended to mark these methods with the correct visibility later. This was pretty complicated, and also we found the issue (as @jez predicted) that when trees were flattened, as a seemingly unintentional implementation quirk, the orphaned visibility calls were moved to the bottom of the file. This made our AST walk implementation coupled to this implementation quirk and forced us to look into changing rewriting to not do this. However, we decided it wouldn't be worth it to remove this implementation quirk, because that would cause parsing behavior to remain coupled to new rewriter implementation details, which we wanted to avoid.

We realized this is simpler to implement entirely during a rewriter step, where we could simply rewrite to use the inline visibility expression:
```ruby
class A
  extend T::Sig
  private
  def foo; end
end
```
becomes
```ruby
class A
  private :foo
end
```

This allowed us to keep parsing simpler and use the existing way to track privacy. We also believe this is more in line with the idea of desugaring, which is to take complicated syntactical expressions which are more complicated to type-check and turn them into simpler expressions.

### Motivation

We made this change because Ruby supports this manner of declaring visibility. Although some organizations use inline visibility (e.g. `private def foo; end` or `private :foo`, our organization almost solely uses the orphaned style. 

### Test plan

We changed an old test that was passing to now include an error (and continue to pass). We also included a flatten tree test for private methods, since our change specifically concerns itself around implementation details with flattening.

### Kudos

Thank you to @bss for all of the pairing sessions on this, he is the true MVP here.


rewriter/Visibility.cc
```diff
@@ -0,0 +1,46 @@
...
+using namespace std;
+
+namespace sorbet::rewriter {
+
+void Visibility::run(core::MutableContext ctx, ast::ClassDef *classDef) {
+    core::NameRef currentVisibility = core::Names::public_();
+
+    for (auto &stat : classDef->rhs) {
+        typecase(
+            stat.get(), [&](ast::Assign *assign) {},
```
jez, line 19:
```suggestion
            stat.get(),
```

This is redundant with the fallback case below for `ast::Expression`, right?

rewriter/Visibility.cc
```diff
@@ -0,0 +1,46 @@
...
+                } else if (send->fun == core::Names::protected_()) {
+                    currentVisibility = core::Names::protected_();
+                } else if (send->fun == core::Names::public_()) {
+                    currentVisibility = core::Names::public_();
+                }
+            },
+            [&](ast::MethodDef *mdef) {
+                if (currentVisibility == core::Names::private_()) {
+                    ast::MethodDef::ARGS_store newArgs;
+                    newArgs.reserve(0);
```
jez, line 36:
```suggestion
```

AlexEvanczuk:
There are two failing tests:
```
~/workspace/sorbet - (ae-bs-respect-orphaned-visibility-method-invocations) $ bazel test //test/cli:test_autogen-errors --test_output=all
INFO: Analyzed target //test/cli:test_autogen-errors (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
FAIL: //test/cli:test_autogen-errors (see /private/var/tmp/_bazel_alex.evanczuk/645cee74aa4c9b8a8f1213d0701f5c81/execroot/com_stripe_ruby_typer/bazel-out/darwin-fastbuild/testlogs/test/cli/test_autogen-errors/test.log)
INFO: From Testing //test/cli:test_autogen-errors:
==================== Test output for //test/cli:test_autogen-errors:
test/cli/autogen-errors/autogen-errors.sh: line 2: 82860 Illegal instruction: 4  main/sorbet --silence-dev-message -p autogen-msgpack --autogen-version=2 --stop-after=namer --skip-rewriter-passes test/cli/autogen-errors/autogen-errors.rb 2>&1 > /dev/null
--- test/cli/autogen-errors/autogen-errors.out	2020-01-23 21:45:29.000000000 +0000
+++ /dev/fd/63	2020-12-03 17:10:03.000000000 +0000
@@ -1,10 +1,2 @@
-test/cli/autogen-errors/autogen-errors.rb:5: Circular dependency: `Bar` and `Foo` are declared as parents of each other https://srb.help/5011
-     5 |class Bar < Foo
-                    ^^^
-    test/cli/autogen-errors/autogen-errors.rb:5: One definition
-     5 |class Bar < Foo
-        ^^^^^^^^^^^^^^^
-    test/cli/autogen-errors/autogen-errors.rb:3: Other definition
-     3 |class Foo < Bar
-        ^^^^^^^^^^^^^^^
-Errors: 1
+mpack breakpoint hit at external/com_github_ludocode_mpack/src/mpack/mpack-common.c:423
+attempting to close a mpack_type_array but there are 9 elements left
================================================================================
Target //test/cli:test_autogen-errors up-to-date:
  bazel-bin/test/cli/test_autogen-errors
INFO: Elapsed time: 2.622s, Critical Path: 1.12s
INFO: 2 processes: 2 darwin-sandbox.
INFO: Build completed, 1 test FAILED, 2 total actions
//test/cli:test_autogen-errors                                           FAILED in 1.1s
  /private/var/tmp/_bazel_alex.evanczuk/645cee74aa4c9b8a8f1213d0701f5c81/execroot/com_stripe_ruby_typer/bazel-out/darwin-fastbuild/testlogs/test/cli/test_autogen-errors/test.log

INFO: Build completed, 1 test FAILED, 2 total actions
~/workspace/sorbet - (ae-bs-respect-orphaned-visibility-method-invocations) $ bazel test //test/cli:test_print_to_file --test_output=all
INFO: Analyzed target //test/cli:test_print_to_file (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
FAIL: //test/cli:test_print_to_file (see /private/var/tmp/_bazel_alex.evanczuk/645cee74aa4c9b8a8f1213d0701f5c81/execroot/com_stripe_ruby_typer/bazel-out/darwin-fastbuild/testlogs/test/cli/test_print_to_file/test.log)
INFO: From Testing //test/cli:test_print_to_file:
==================== Test output for //test/cli:test_print_to_file:
mpack breakpoint hit at external/com_github_ludocode_mpack/src/mpack/mpack-common.c:423
attempting to close a mpack_type_array but there are 9 elements left
test/cli/print_to_file/print_to_file.sh: line 6: 82969 Illegal instruction: 4  main/sorbet --silence-dev-message --stop-after namer --autogen-version=2 -p autogen:autogen.txt -p autogen-msgpack:autogen.msgpack test/cli/print_to_file/*.rb
--- test/cli/print_to_file/print_to_file.out	2020-05-27 21:00:44.000000000 +0000
+++ /dev/fd/63	2020-12-03 17:10:11.000000000 +0000
@@ -1,231 +0,0 @@
---- autogen.txt start ---
-# ParsedFile: test/cli/print_to_file/a.rb
-requires: []
-## defs:
-[def id=0]
- type=module
- defines_behavior=0
- is_empty=0
-[def id=1]
- type=module
- defines_behavior=0
- is_empty=0
- defining_ref=[A]
-[def id=2]
- type=class
- defines_behavior=0
- is_empty=1
- defining_ref=[Foo]
-## refs:
-[ref id=0]
- scope=[]
- name=[A]
- nesting=[]
- resolved=[A]
- loc=test/cli/print_to_file/a.rb:3
- is_defining_ref=1
-[ref id=1]
- scope=[A]
- name=[Foo]
- nesting=[[A]]
- resolved=[A Foo]
- loc=test/cli/print_to_file/a.rb:4
- is_defining_ref=1
-# ParsedFile: test/cli/print_to_file/b.rb
-requires: []
-## defs:
-[def id=0]
- type=module
- defines_behavior=0
- is_empty=0
-[def id=1]
- type=module
- defines_behavior=0
- is_empty=0
- defining_ref=[B]
-[def id=2]
- type=class
- defines_behavior=0
- is_empty=1
- defining_ref=[Bar]
-## refs:
-[ref id=0]
- scope=[]
- name=[B]
- nesting=[]
- resolved=[B]
- loc=test/cli/print_to_file/b.rb:3
- is_defining_ref=1
-[ref id=1]
- scope=[B]
- name=[Bar]
- nesting=[[B]]
- resolved=[B Bar]
- loc=test/cli/print_to_file/b.rb:4
- is_defining_ref=1
-# ParsedFile: test/cli/print_to_file/c.rb
-requires: []
-## defs:
-[def id=0]
- type=module
- defines_behavior=0
- is_empty=0
-[def id=1]
- type=module
- defines_behavior=0
- is_empty=0
- defining_ref=[C]
-[def id=2]
- type=class
- defines_behavior=0
- is_empty=1
- defining_ref=[Bar]
-[def id=3]
- type=class
- defines_behavior=1
- is_empty=0
- defining_ref=[BarChild]
- parent_ref=[Bar]
-[def id=4]
- type=module
- defines_behavior=1
- is_empty=0
- defining_ref=[TestExtend]
-## refs:
-[ref id=0]
- scope=[]
- name=[C]
- nesting=[]
- resolved=[C]
- loc=test/cli/print_to_file/c.rb:3
- is_defining_ref=1
-[ref id=1]
- scope=[C]
- name=[Bar]
- nesting=[[C]]
- resolved=[C Bar]
- loc=test/cli/print_to_file/c.rb:4
- is_defining_ref=1
-[ref id=2]
- scope=[C]
- name=[BarChild]
- nesting=[[C]]
- resolved=[C BarChild]
- loc=test/cli/print_to_file/c.rb:6
- is_defining_ref=1
-[ref id=3]
- scope=[C]
- name=[Bar]
- nesting=[[C]]
- resolved=[C Bar]
- loc=test/cli/print_to_file/c.rb:6
- is_defining_ref=0
- parent_of=[C BarChild]
-[ref id=4]
- scope=[]
- name=[X]
- nesting=[]
- resolved=[]
- loc=test/cli/print_to_file/c.rb:7
- is_defining_ref=0
- parent_of=[C BarChild]
-[ref id=5]
- scope=[]
- name=[InMethodX]
- nesting=[]
- resolved=[]
- loc=test/cli/print_to_file/c.rb:9
- is_defining_ref=0
-[ref id=6]
- scope=[C]
- name=[TestExtend]
- nesting=[[C]]
- resolved=[C TestExtend]
- loc=test/cli/print_to_file/c.rb:13
- is_defining_ref=1
-[ref id=7]
- scope=[]
- name=[Y]
- nesting=[]
- resolved=[]
- loc=test/cli/print_to_file/c.rb:14
- is_defining_ref=0
-[ref id=8]
- scope=[]
- name=[InMethodY]
- nesting=[]
- resolved=[]
- loc=test/cli/print_to_file/c.rb:16
- is_defining_ref=0
-[ref id=9]
- scope=[]
- name=[ViaMethod]
- nesting=[]
- resolved=[]
- loc=test/cli/print_to_file/c.rb:19
- is_defining_ref=0
-# ParsedFile: test/cli/print_to_file/d.rb
-requires: []
-## defs:
-[def id=0]
- type=module
- defines_behavior=0
- is_empty=0
-[def id=1]
- type=module
- defines_behavior=0
- is_empty=1
- defining_ref=[MyMixin]
-[def id=2]
- type=class
- defines_behavior=1
- is_empty=0
- defining_ref=[MyClass]
-[def id=3]
- type=alias
- defines_behavior=1
- is_empty=0
- defining_ref=[X]
- aliased_ref=[Y]
-## refs:
-[ref id=0]
- scope=[]
- name=[MyMixin]
- nesting=[]
- resolved=[MyMixin]
- loc=test/cli/print_to_file/d.rb:3
- is_defining_ref=1
-[ref id=1]
- scope=[]
- name=[MyClass]
- nesting=[]
- resolved=[MyClass]
- loc=test/cli/print_to_file/d.rb:4
- is_defining_ref=1
-[ref id=2]
- scope=[MyClass]
- name=[MyMixin]
- nesting=[[MyClass]]
- resolved=[MyMixin]
- loc=test/cli/print_to_file/d.rb:6
- is_defining_ref=0
- parent_of=[MyClass]
-[ref id=3]
- scope=[MyClass]
- name=[X]
- nesting=[[MyClass]]
- resolved=[MyClass X]
- loc=test/cli/print_to_file/d.rb:5
- is_defining_ref=1
-[ref id=4]
- scope=[MyClass]
- name=[Y]
- nesting=[[MyClass]]
- resolved=[]
- loc=test/cli/print_to_file/d.rb:5
- is_defining_ref=0
---- autogen.txt end ---
-
-304d23737beec8076d49ec970a5d6b85c940b9c6  autogen.msgpack
-
---print=cfg specified multiple times with inconsistent output options
================================================================================
Target //test/cli:test_print_to_file up-to-date:
  bazel-bin/test/cli/test_print_to_file
INFO: Elapsed time: 2.927s, Critical Path: 1.86s
INFO: 2 processes: 2 darwin-sandbox.
INFO: Build completed, 1 test FAILED, 2 total actions
//test/cli:test_print_to_file                                            FAILED in 1.8s
  /private/var/tmp/_bazel_alex.evanczuk/645cee74aa4c9b8a8f1213d0701f5c81/execroot/com_stripe_ruby_typer/bazel-out/darwin-fastbuild/testlogs/test/cli/test_print_to_file/test.log

INFO: Build completed, 1 test FAILED, 2 total actions
```

rewriter/Visibility.cc
AlexEvanczuk: Is `ast::Assign` a subtype of `ast::Expression`? That is, will it also be considered to satisfy that case in the typecase?

rewriter/Visibility.cc
AlexEvanczuk: Thanks, missed this one.

rewriter/Visibility.cc
```diff
@@ -0,0 +1,46 @@
...
+                    auto privateCall = ast::MK::Send1(mdef->loc, ast::MK::Self(mdef->loc), core::Names::private_(),
+                                                      ast::MK::Symbol(mdef->loc, mdef->name));
```
jez, line 38:
```suggestion
                    auto privateCall = ast::MK::Send1(mdef->declLoc, ast::MK::Self(mdef->declLoc), core::Names::private_(),
                                                      ast::MK::Symbol(mdef->declLoc, mdef->name));
```

`MethodDef::loc` is the whole method def, from the `def` keyword all the way to the `end` keyword at the end of the body. It usually isn't great to show in error messages. We probably want `declLoc` here, which is just from the `def` to the end of the line

rewriter/Visibility.cc
```diff
@@ -0,0 +1,46 @@
...
+                    classDef->rhs.insert(classDef->rhs.end(), std::move(privateCall));
```
jez, line 37: You are iterating over and modifying `classDef->rhs` in the same loop. Can we queue up the edits and process them after the loop?

rewriter/Visibility.cc
```diff
@@ -0,0 +1,44 @@
...
+void Visibility::run(core::MutableContext ctx, ast::ClassDef *classDef) {
+    core::NameRef currentVisibility = core::Names::public_();
+
+    for (auto &stat : classDef->rhs) {
+        typecase(
+            stat.get(),
+            [&](ast::Send *send) {
+                if (send->args.size() != 0 && send->recv != nullptr) {
+                    return;
+                }
```
jez, line 23: It looks like this is currently prohibited:

```
class A
  self.private
  def foo; end
end
```

rewriter/Visibility.cc
jez: Yep.

rewriter/Visibility.cc
AlexEvanczuk: Ah okay -- yeah, Bo and I were wondering about this. @bss I can try to do this, but I might need some help with the syntax on this. If you are able to push to my branch or suggest edits that would work too.

rewriter/Visibility.cc
AlexEvanczuk: By prohibited, do you mean that way of specifying orphaned calls will continue to not be respected by Sorbet?

rewriter/Visibility.cc
AlexEvanczuk: Pushed this change.

rewriter/Visibility.cc
jez: The syntax is just make yourself a `vector<TreePtr>` before your loop, collect all your newly minted send nodes into that vector with `.emplace_back(privateCall)` and then at the end of the loop, insert all the new private calls into the end of `rhs`.

rewriter/Visibility.h
```diff
@@ -0,0 +1,29 @@
...
+ * This class is meant to support expressions of visibility
+ *
+ * It desugars things of the form
+ *
+ *   private
+ *   def foo; end
+ *
+ * into
+ *
+ *   private :foo
```
jez, line 17:
```suggestion
 *   private
 *   def foo; end
 *   private :foo
```

rewriter/Visibility.h
```diff
@@ -0,0 +1,29 @@
...
+
+namespace sorbet::rewriter {
+
+/**
+ * This class is meant to support expressions of visibility
+ *
+ * It desugars things of the form
+ *
+ *   private
+ *   def foo; end
```
jez, line 13:
```suggestion
 *   def foo; end
 *   def bar; end
```

rewriter/rewriter.cc
```diff
@@ -47,6 +48,7 @@ class Rewriterer {
         Prop::run(ctx, classDef);
         TypeMembers::run(ctx, classDef);
         Singleton::run(ctx, classDef);
+        Visibility::run(ctx, classDef);
```
jez, line 51: idk if you care about this, but

```ruby
private
attr_reader :foo
```

is not going to make `foo` private.

jez: The attr_reader DSL fires after your Visibility rewriter pass.

Essentially, what you want is for your rewriter pass to fire "after all other rewriter passes," because any rewriter pass might introduce new methods.

With that in mind, have you considered doing this in namer instead? Namer has a view of "all the methods that are defined" that rewriter passes don't (because namer comes after rewriter)

test/testdata/infer/private_methods.rb.flatten-tree.exp
```diff
@@ -0,0 +1,111 @@
+begin
```
jez, line 1: flatten-tree is the pass that happens after namer, resolver, and class-flatten (in that order). rewriter comes before all three of those passes.

Would you mind replacing this with a `rewrite-tree` expectation file? That will make it easier to see what the changes from this PR are (because once namer reads private visibilities, it doesn't matter how passes after namer further transform and mangle the tree)

rewriter/Visibility.cc
jez: ah sorry i meant that my (now edited) test case will fail with this PR as written, but i think it should pass.

rewriter/rewriter.cc
AlexEvanczuk: Hmm... @bss What do you think about this? How do you all feel about shipping this as is and keeping that for a separate change?

jez (commented):
This looks mostly fine. It has some limitations which I think can only be overcome by moving this into namer, but honestly I'm fine merging this anyways if you think it's Good Enough for the time being (and we can make issues for the problems that moving this to namer would fix).

rewriter/rewriter.cc
jez: > How do you all feel about shipping this as is and keeping that for a separate change?

(I mention being fine with this in a comment below fwiw)

AlexEvanczuk:
> This looks mostly fine. It has some limitations which I think can only be overcome by moving this into namer, but honestly I'm fine merging this anyways if you think it's Good Enough for the time being (and we can make issues for the problems that moving this to namer would fix).

@jez By moving into namer, do you mean going back to our old implementation plan? Can you elaborate a little?

jez:
> > This looks mostly fine. It has some limitations which I think can only be overcome by moving this into namer, but honestly I'm fine merging this anyways if you think it's Good Enough for the time being (and we can make issues for the problems that moving this to namer would fix).
> 
> @jez By moving into namer, do you mean going back to our old implementation plan? Can you elaborate a little?

I kind of imagined a solution where when namer populating it's various FoundDefinition lists, it also keeps track of like the most recent found visibility. So then when namer is calling `addMethod` to create found methods, it looks up to see if there is a current default visibility, and if so it calls `addModifier` right after calling `addMethod`.

AlexEvanczuk:
Let me sync up with Bo to address your various comments and your comment about it being in the namer.

He's out tomorrow but we'll get back to you next week!

rewriter/rewriter.cc
AlexEvanczuk: We'll move the visibility rewriter pass as the last rewriter so it picks up any new methods that are added.

AlexEvanczuk:
@jez 
We've addressed all comments, with the exception of:
1) We didn't yet fix orphaned private calls invoked via `self` as an explicit receiver of the method invocation. We were having some issues there and would prefer to come back to it in a subsequent change set, if that works. We've left the test case explicitly unhandled for now.
2) We stuck with the rewriter approach for now. We think it'll work -- I'm curious what you perceive the remaining limitations are of the approach of implementing this in the rewriter.

Let me know if there are any other concerns you'd like addressed before we merge this!

jez:
Thanks for putting this together.

When I was talking about moving this to namer, this is what I had in mind:

https://github.com/sorbet/sorbet/pull/3817

The change is considerably shorter (doesn't need a whole new rewriter pass) and is co-located with all other private method handling logic (less fragmented).

If it's all good with you, I'd like to move forward with that implementation. Please do test it on your codebase and let me know if it's buggy.

jez:
> We didn't yet fix orphaned private calls invoked via `self` as an explicit receiver of the method invocation. We were having some issues there and would prefer to come back to it in a subsequent change set, if that works. We've left the test case explicitly unhandled for now.

I noticed that this is a problem with all private methods, not just with this PR.

I've opened an issue to track that separately (another benefit to having the changes colocated in namer: there will only be one place to fix that bug, not two).

#3816

AlexEvanczuk:
@jez 

I bumped sorbet up to the latest version in our codebase, fixed errors, then pulled your branch and ran sorbet using the branch's version, and there are no errors in our codebase. I also just ran a quick smoke test where I marked a set of methods as private using subsequent visibility syntax and I see the expected results.

I'm totally fine with whatever implementation you think is best.

I'm glad that seeing our work here has helped us realize there are better alternatives for implementation :) Given that all of the same tests are passing, let's land it!

Thank you for working with us here.

Issue #3761, opened 2020-12-04
Enum value parameter types fail to typecheck  depending on file order


jvilk-stripe:
#### Input

`file1.rb`:
```ruby
# typed: true
class Foo
  extend T::Sig

  def initialize
    @a = T.let(Suit::Spades, Suit::Spades)
  end

  sig{params(a: Suit::Spades).returns(T::Boolean)}
  def foo(a)
    true
  end

end
```

`file2.rb`:
```ruby
# typed: true

class Suit < T::Enum
  enums do
    Spades = new('SPADES')
    Hearts = new('HEARTS')
    Clubs = new('CLUBS')
    Diamonds = new('DIAMONDS')
  end
end
```

#### Observed output

```
lol1.rb:6: Constant Suit::Spades is not a class or type alias https://srb.help/5004
     6 |    @a = T.let(Suit::Spades, Suit::Spades)
                                     ^^^^^^^^^^^^
    lol2.rb:5: If you are trying to define a type alias, you should use T.type_alias here
     5 |    Spades = new('SPADES')
            ^^^^^^

lol1.rb:9: Constant Suit::Spades is not a class or type alias https://srb.help/5004
     9 |  sig{params(a: Suit::Spades).returns(T::Boolean)}
                        ^^^^^^^^^^^^
    lol2.rb:5: If you are trying to define a type alias, you should use T.type_alias here
     5 |    Spades = new('SPADES')
            ^^^^^^
Errors: 2
```

#### Expected behavior

No typechecking errors.

If you reverse the order of the files (rename file1 to file2 and file2 to file1) the files typecheck cleanly.

The bug is in ResolveSignaturesWalk, and I noticed it while working on https://github.com/sorbet/sorbet/pull/3686 which exacerbates the issue. This pass both defines the resultType of static fields and parses the type of sigs and T.let. However, the type of sigs and T.lets with enum values depends on the result type of static fields being set.

The solution is to resolve the resultType of all static fields _before_ ResolveSignaturesWalk.

jvilk-stripe:
Solving this issue could also allow us to type alias enum values.

Issue #3602, opened 2020-10-31
Create type aliases of T::Enum values


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20MyEnum%20%3C%20T%3A%3AEnum%0A%20%20enums%20do%0A%20%20%20%20X%20%3D%20new%0A%20%20%20%20Y%20%3D%20new%0A%20%20%20%20Z%20%3D%20new%0A%20%20end%0Aend%0A%0Aextend%20T%3A%3ASig%0Asig%20%7Bparams(xy%3A%20T.any(MyEnum%3A%3AX%2C%20MyEnum%3A%3AY)).void%7D%20%23%20ok%0Adef%20foo(xy)%0Aend%0A%0AXY%20%3D%20T.type_alias%20%7BT.any(MyEnum%3A%3AX%2C%20MyEnum%3A%3AY)%7D%20%23%20error)

```ruby
# typed: true
class MyEnum < T::Enum
  enums do
    X = new
    Y = new
    Z = new
  end
end

extend T::Sig
sig {params(xy: T.any(MyEnum::X, MyEnum::Y)).void} # ok
def foo(xy)
end

XY = T.type_alias {T.any(MyEnum::X, MyEnum::Y)} # error
```

#### Observed output

```
editor.rb:15: Constant MyEnum::X is not a class or type alias https://srb.help/5004
    15 |XY = T.type_alias {T.any(MyEnum::X, MyEnum::Y)} # error
                                 ^^^^^^^^^
    editor.rb:4: If you are trying to define a type alias, you should use T.type_alias here
     4 |    X = new
            ^

editor.rb:15: Constant MyEnum::Y is not a class or type alias https://srb.help/5004
    15 |XY = T.type_alias {T.any(MyEnum::X, MyEnum::Y)} # error
                                            ^^^^^^^^^
    editor.rb:5: If you are trying to define a type alias, you should use T.type_alias here
     5 |    Y = new
            ^
Errors: 2
```

#### Expected behavior

The only reason why this is an error is because it's hard to implement, and when we implemented `T::Enum` at first, we punted on supporting this.

It would be a good feature to add, but it's tricky to get the implementation to fit into resolver.

- - -

jvilk-stripe:
Fixed in https://github.com/sorbet/sorbet/pull/3762

PR #3762, opened 2020-12-05
Define fields in ResolveTypeParamsWalk

jvilk-stripe:
Define fields in ResolveTypeParamsWalk rather than ResolveSignaturesWalk. Makes it possible for `type_aliases` to be set to enum values (#3602), and fixes a bug where Sorbet would sometimes erroneously disallow enum values in signatures (#3761).

Internally, enum values are modeled as static fields. The following code:
```ruby
class MyEnum < T::Enum
     enums do
       X = new
     end
   end
```

...gets translated into the following form:
```ruby
class MyEnum < T::Enum
  extend T::Helpers
  sealed!
  abstract!

  class X$1 < MyEnum; include Singleton; final!; end
  X = T.let(X$1.instance, X$1)
  new
end
```

Thus, `MyEnum::X` is a static field. However, unlike a regular static field, Sorbet lets you use it in type signatures! Sorbet does _not_ let you do this with arbitrary static fields.

```ruby
sig {params(x: MyEnum::X).void}
def method(x)
end
```

Unfortunately, this design poses a slight problem for resolver. Prior to the changes in this PR, static fields were resolved _after_ type aliases and _in parallel with_ signatures, causing the aforementioned bugs. Here's a diagram of dependencies between the types of various symbols and casts in Sorbet, with the numbers (and shading) indicating the order in which items are defined:

Type dependencies (before PR)

Unlike static fields in general, enum values always have a type that is a simple class with no type members. Thus, it should be possible to resolve them _before_ type aliases. Moreover, since the type of fields is not influenced by method signatures, we can resolve them all before ResolveSignaturesWalk!

Those changes look like this:
Type dependencies (after PR)

Now, everything that a symbol needs to resolve its type is ready when we resolve it, fixing the bugs.

I made a few minor adjustments as part of this change as well:

* ResolveTypeMembersWalk now resolves the type of all classes (externalType). Previously, I had this as a freestanding method called from the main resolver run methods.
* ResolveTypeMembersWalk is no longer a shallow tree map; it's a full tree walk. This change was needed to visit the static fields.

As an added bonus, since ResolveTypeMembersWalk is parallelized and ResolveSignaturesWalk is not, this change should speed up resolver by a bit.

EDIT: The change did not speed up resolver. It made `ResolveSignaturesWalk` a little bit faster, but added slightly more time to type params. The end result is a minor negligible slowdown on Stripe's codebase (within noise). But since this change unlocks the ability to parallelize `ResolveSignaturesWalk`, I expect resolver to get much faster in the near future!

### Motivation

Fixes https://github.com/sorbet/sorbet/issues/3761 and https://github.com/sorbet/sorbet/issues/3602.

### Test plan

See included automated tests.


jvilk-stripe:
This change should unblock my work on parallelizing resolve signatures walk, too. :)

jvilk-stripe:

→ https://go/builds/bui_IVtNUvBK2G5vfS

jvilk-stripe:

→ https://go/builds/bui_IW6k0DkD0GFGcO

jvilk-stripe:

→ https://go/builds/bui_IW6sKYKBw16xne

ast/TreeSanityChecks.cc
```diff
@@ -76,7 +76,8 @@ void Break::_sanityCheck() {
 
 void Cast::_sanityCheck() {
     ENFORCE(arg);
-    ENFORCE(type);
+    // Note: We initially set `type` to `nullptr` in resolver before filling it in.
+    // ENFORCE(type);
```
jvilk-stripe, line 80: If we have a better way of doing this (TODO type?) I'm all ears.

resolver/resolver.cc
```diff
@@ -1120,6 +1155,185 @@ class ResolveTypeMembersWalk {
...
+                    }
+                }
+                result = cast.type;
+            },
+            [&](const ast::InsSeq &outer) { result = resolveConstantType(ctx, outer.expr); },
+            [&](const ast::TreePtr &expr) {});
+        return result;
+    }
+
+    [[nodiscard]] static core::TypePtr tryResolveStaticField(core::Context ctx, ResolveStaticFieldItem &job) {
```
elliottt, line 1294: Why did you only mark this one as `[[nodiscard]]`?

elliottt (approved):
Just a few questions, but otherwise this looks great!

* Can `ResolveSignaturesWalk` be a shallow map now? Most of what it was doing before has been moved into the type members pass, so perhaps it can stop traversing method bodies?
* We currently can't resolve type parameters from method sigs with casts in method bodies. Previously this seemed like it was just a bug to fix, but processing casts before method signatures greatly complicates resolving it. Is it absolutely required that we resolve casts in this pass?

jvilk-stripe:
> Can ResolveSignaturesWalk be a shallow map now? Most of what it was doing before has been moved into the type members pass, so perhaps it can stop traversing method bodies?

Yeah, it probably can be! I'll check.

> We currently can't resolve type parameters from method sigs with casts in method bodies. Previously this seemed like it was just a bug to fix, but processing casts before method signatures greatly complicates resolving it. Is it absolutely required that we resolve casts in this pass?

Good question. It actually seems like we could easily address this limitation by deferring the definition of 'complicated' casts and fields until after signature resolution. I'd argue this change makes it easier to fix this limitation.

resolver/resolver.cc
```diff
@@ -1521,6 +1882,15 @@ class ResolveTypeMembersWalk {
...
         }
 
+        // Resolve simple field declarations. Required so that `type_alias` can refer to an enum value type
+        // (which is a static field).
+        {
+            for (auto &job : combined.todoResolveSimpleStaticFieldItems) {
+                job.sym.data(gs)->resultType = job.resultType;
+            }
+            combined.todoResolveSimpleStaticFieldItems.clear();
+        }
```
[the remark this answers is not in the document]
jvilk-stripe: 🤷 can fix

core/Context.h
```diff
@@ -57,14 +57,14 @@ class MutableContext final {
         : state(state), owner(owner), file(file) {}
     MutableContext(const MutableContext &other) noexcept : state(other.state), owner(other.owner), file(other.file) {}
 
+    bool permitOverloadDefinitions(FileRef sigLoc) const;
+
```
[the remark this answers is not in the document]
jvilk-stripe: Accident.

ast/TreeSanityChecks.cc
jvilk-stripe: We don't have a TODO `TypePtr` though, so I don't know what to use here.

resolver/resolver.cc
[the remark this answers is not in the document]
jvilk-stripe: Fixed.

core/Context.h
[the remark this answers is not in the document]
jvilk-stripe: Fixed.

resolver/resolver.cc
jvilk-stripe: Good point. I marked the others as well.

jvilk-stripe:
>Can ResolveSignaturesWalk be a shallow map now? Most of what it was doing before has been moved into the type members pass, so perhaps it can stop traversing method bodies?

I can do this once I move its `Send` visitor to the type params pass. I'll do that in a follow-up PR.

elliottt:
> Good question. It actually seems like we could easily address this limitation by deferring the definition of 'complicated' casts and fields until after signature resolution. I'd argue this change makes it easier to fix this limitation.

@jez had a good suggestion -- we can implement this by adding an intrinsic for `T.type_parameter` so that it's resolved during inference instead. This change makes sense to me, so I say don't think too hard about my original suggestion :)

ast/TreeSanityChecks.cc
jvilk-stripe: After conversation: Will use a TypePtr to a ClassType of todo.

resolver/resolver.cc
```diff
@@ -1311,8 +1533,57 @@ class ResolveTypeMembersWalk {
...
+                e.setHeader("Use `{}` to specify the type of constants", "T.let");
+            }
+        }
+
+        ResolveFieldItem &job = todoResolveFieldItems_.emplace_back();
+        job.file = ctx.file;
+        job.owner = ctx.owner;
+        job.cast = cast;
+        job.ident = uid;
+        job.atTopLevel = nestedBlockCounts.back() == 0;
```
jez, line 1566: Why did we chose this method? If someone adds a new field, clang won't catch all the places where you need to now initialize that field.

Versus

```cpp
todoResolveFieldItems_.emplace_back(ResolveFieldItem{
  ctx.file,
  ctx.owner,
  cast,
  uid,
  nestedBlockCounts.back() == 0,
});
```

resolver/resolver.cc
```diff
@@ -1441,6 +1776,22 @@ class ResolveTypeMembersWalk {
                     send->fun == core::Names::typeTemplate());
 
             todoAssigns_.emplace_back(ResolveAssignItem{ctx.owner, sym, send, dependencies_, ctx.file});
+        } else if (data->isStaticField()) {
+            ResolveStaticFieldItem job;
+            job.file = ctx.file;
+            job.sym = sym;
+            job.asgn = &asgn;
```
jez, line 1783: (ditto field initialization choices above)

resolver/resolver.cc
```diff
@@ -1522,6 +1893,13 @@ class ResolveTypeMembersWalk {
             resolveAttachedClass(ctx, job.klass, resolvedAttachedClasses);
         }
 
+        // Resolve simple field declarations. Required so that `type_alias` can refer to an enum value type
+        // (which is a static field).
```
jez, line 1897: Do you mind saying that this is a stronger claim than we need? We only need that static fields used by `T::Enum` are resolved, right?

Also, I think that we technically could have implemented that logic precisely by asking whether the RHS of the static field assignment had an enum NameRef with `isTEnumName`.

test/testdata/resolver/enum_val_types_order1__file1.rb
```diff
@@ -0,0 +1,15 @@
+# typed: true
```
jez, line 1: Our test suite has no coverage of "generic T::Enum"

There was a point in the past when T::Enum was a generic (it included Enumerable) and it's not unreasonable to think that it might become a generic in the future (you'd be required to specify what type the enum serializes to, so that `MyEnum::X.serialize` could return `String` or `Integer` or `Symbol` instead of `T.untyped`)

"generics + T::Enum" could presumably have codepaths that are only triggered by paths introduced in this PR. would you mind adding some tests as such? [for example, here's a starter](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20MyEnum%20%3C%20T%3A%3AEnum%0A%20%20extend%20T%3A%3AGeneric%0A%20%20Value%20%3D%20type_template(fixed%3A%20String)%0A%0A%20%20enums%20do%0A%20%20%20%20X%20%3D%20new%0A%20%20%20%20Y%20%3D%20new%0A%20%20%20%20Z%20%3D%20new%0A%20%20end%0Aend%0A%0Aextend%20T%3A%3ASig%0Asig%20%7Bparams(xy%3A%20T.any(MyEnum%3A%3AX%5BInteger%5D%2C%20MyEnum%3A%3AY%5BInteger%5D)).void%7D%20%23%20ok%0Adef%20foo(xy)%0Aend%0A%0AXY%20%3D%20T.type_alias%20%7BT.any(MyEnum%3A%3AX%5BInteger%5D%2C%20MyEnum%3A%3AY%5BInteger%5D)%7D%20%23%20error)

such tests will likely have lots of errors, and it's fine to say that we expect them, but it should make sure we don't have for example: a nullptr that we were expecting to be non-null.

jez:
Separately, this lets me remove a couple TODOs in `tenum.md` that I had been putting off until this bug was fixed!

https://github.com/sorbet/sorbet/blame/master/website/docs/tenum.md#L173-L174

resolver/resolver.cc
```diff
@@ -1040,7 +1040,11 @@ class ResolveConstantsWalk {
     }
 };
 
-class ResolveTypeMembersWalk {
+class ResolveTypeMembersAndFieldsWalk {
+    static core::TypePtr todoType() {
+        return core::make_type<core::ClassType>(core::Symbols::todo());
+    }
```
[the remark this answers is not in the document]
jvilk-stripe: Good idea.

resolver/resolver.cc
```diff
@@ -1121,6 +1160,189 @@ class ResolveTypeMembersWalk {
...
+        ENFORCE(job.sym.data(ctx)->isStaticField());
+        auto &asgn = job.asgn;
+        auto data = job.sym.data(ctx);
+        if (data->resultType == nullptr) {
+            auto resultType = resolveConstantType(ctx, asgn->rhs);
+            if (!resultType) {
+                return todoType();
+            } else {
+                return resultType;
+            }
```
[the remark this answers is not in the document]
jvilk-stripe: I actually recently learned via an insidious bug that this might not be true.

https://godbolt.org/z/Gsbrss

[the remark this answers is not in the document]
jvilk-stripe: (but I agree w/ you)

resolver/resolver.cc
jvilk-stripe: It's self-documenting; the way you have it doesn't document the field names, and clang won't complain if I miss one (it'll default initialize it). But I have no strong preference here and can do it that way.

resolver/resolver.cc
```diff
@@ -1121,6 +1160,189 @@ class ResolveTypeMembersWalk {
...
+            scope = ctx.selfClass();
+        }
+
+        auto prior = scope.data(ctx)->findMember(ctx, uid->name);
+        if (prior.exists()) {
+            if (core::Types::equiv(ctx, prior.data(ctx)->resultType, cast->type)) {
+                // We already have a symbol for this field, and it matches what we already saw, so we can short
+                // circuit.
+                return;
```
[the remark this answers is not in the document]
jvilk-stripe: At this point in the code, we've discovered that there's a field with the same name (it has to be a field since `uid->name` is going to be `@something`). If the field is defined, it must have a `resultType` (we never define a field and _not_ add `resultType`). So now it's checking if the type matches.

resolver/resolver.cc
```diff
@@ -1395,6 +1691,46 @@ class ResolveTypeMembersWalk {
...
+            case core::Names::let()._id:
+            case core::Names::uncheckedLet()._id:
+            case core::Names::assertType()._id:
+            case core::Names::cast()._id: {
+                if (trackDependencies_) {
+                    classOfDepth_.pop_back();
+                }
+
+                auto *id = ast::cast_tree<ast::ConstantLit>(send.recv);
+                // Must be T::let/uncheckedLet/etc.
```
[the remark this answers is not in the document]
jvilk-stripe: Both? We're processing something names `let`/etc and we want to check it's `T::let`/etc.

resolver/resolver.cc
```diff
@@ -1325,6 +1596,31 @@ class ResolveTypeMembersWalk {
...
+
+    ast::TreePtr preTransformBlock(core::Context ctx, ast::TreePtr tree) {
+        nestedBlockCounts.back() += 1;
+        return tree;
+    }
+
+    ast::TreePtr postTransformBlock(core::Context ctx, ast::TreePtr tree) {
+        nestedBlockCounts.back() -= 1;
+        return tree;
+    }
```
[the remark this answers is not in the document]
jvilk-stripe: Sure, can do.

resolver/resolver.cc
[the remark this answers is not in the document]
jez: :ohno:

resolver/resolver.cc
jez: > It's self-documenting

definitely agree that this is nice



> clang won't complain if I miss one (it'll default initialize it)

this is the one that worries me

once we upgrade to C++20 we will be able to use designated initializers to make initializer lists like that self documenting:

https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers

since you wrote it the way you did i'm happy to leave it for now, but i might make another pass over in our codebase at some point.

jvilk-stripe: I fixed it in https://github.com/sorbet/sorbet/pull/3771

PR #3763, opened 2020-12-05
remove Instruction's vtable

froydnj:
These changes apply similar techniques from `TreePtr` and `TypePtr` to convert `Instruction` to a tag-based typing scheme, which means we can dispense with virtual dispatch for the three virtual methods that we have  We don't quite go all the way to full tagged pointers, but ideally this makes it a lot simpler to do so, though we'll only gain 8 bytes on `cfg::Send`.

### Motivation

Smaller instructions are better for everybody.  We shouldn't pay 8 bytes of vtable pointer just for some debugging methods.

### Test plan

Existing tests should be sufficient.

### DIFF
--- a/cfg/CFG.h
+++ b/cfg/CFG.h
@@ -40,9 +40,9 @@ public:
     VariableUseSite bind;
     core::LocOffsets loc;
 
-    std::unique_ptr<Instruction> value;
+    InsnPtr value;
 
-    Binding(LocalRef bind, core::LocOffsets loc, std::unique_ptr<Instruction> value);
+    Binding(LocalRef bind, core::LocOffsets loc, InsnPtr value);
     Binding(Binding &&other) = default;
     Binding() = default;
 
--- a/cfg/CFG.h
+++ b/cfg/CFG.h
@@ -40,9 +40,9 @@ public:
     VariableUseSite bind;
     core::LocOffsets loc;
 
-    InsnPtr value;
+    InstructionPtr value;
 
-    Binding(LocalRef bind, core::LocOffsets loc, InsnPtr value);
+    Binding(LocalRef bind, core::LocOffsets loc, InstructionPtr value);
     Binding(Binding &&other) = default;
     Binding() = default;
 
--- a/cfg/CFG.cc
+++ b/cfg/CFG.cc
@@ -105,8 +105,7 @@ CFG::ReadsAndWrites CFG::findAllReadsAndWrites(core::Context ctx) {
              * assignment. Treating every write as also reading from the
              * variable serves to represent this.
              */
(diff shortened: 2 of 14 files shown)


froydnj:
The emscripten build is failing because its libc++ `<memory>` seems to be buggy; it's calling `std::default_delete<Instruction>::operator()`, but we shouldn't be calling that because our instantiation of `std::unique_ptr` supplies a custom deleter.

I guess this just means we'll have to go the whole way and make `InsnPtr` do tagged storage of pointers like `TreePtr` or `TypePtr`.  The nice part about that is that `Send` will decrease by 8 bytes.

froydnj:
Turns out I didn't need to go full tagged pointers, I think; if I had done 69ed9d8 earlier, tagged pointers could have been left for later.  But doing tagged pointers means `cfg::Send` got smaller and that's a good thing.

(Also, does that mean than Emscripten isn't doing linker gc, or something like that?  Something to investigate later...)

froydnj:

→ https://go/builds/bui_IXAV8IX6LggTr4

froydnj:
It's very sad that this doesn't seem to do anything to max rss.

cfg/CFG.h
```diff
@@ -40,9 +40,9 @@ class Binding final {
     VariableUseSite bind;
     core::LocOffsets loc;
 
-    std::unique_ptr<Instruction> value;
+    InsnPtr value;
```
jez, line 43: Can we still call this `InstructionPtr`?

jez (approved):
I'd like to change the name (separately, I think we should also change TreePtr -> ExpressionPtr or `ast::Expression` -> `ast::Tree`. Eventually we will update internals.md to talk about why there are all these separate data structures, and I'd like to be able to use words like "expression" colloquially and not confuse people as to why one is called `Expression` and one is called `Tree`).

Thanks!

froydnj:
This PR is much easier to deal with now that the compiler is in the same repo.  Patch series rebased and the necessary compiler fixes + a few things from rebasing + review feedback to use `InstructionPtr` applied.

jez (approved):
this still looks good to me.

have you tested on sorbet.run just to be sure that the wasm build isn't going to a problem? (that's usually what breaks when we make a change with heavy bit-twiddling like this)

froydnj:
> this still looks good to me.

Thanks for the re-review!

> have you tested on sorbet.run just to be sure that the wasm build isn't going to a problem? (that's usually what breaks when we make a change with heavy bit-twiddling like this)

I built `//emscripten:sorbet-wasm.tar --config=webasm-linux --strip=always`, untarred the result inside the `docs` directory of a sorbet.run checkout, ran the server, and clicked through the examples locally.  Everything seemed good!

PR #3764, opened 2020-12-06
Introduce safe accessors for Name subtypes.

jvilk-stripe:
Introduce safe accessors for Name subtypes.

Delurks NameRef.data, introduces NameRef.{raw, cnst, unique} and NameRef.kind(). Also delurks the ability to get a mutable reference to a Name given a mutable GlobalState (this feature was never used anywhere, and never should be used).

This change ensures that ~all references to the union field on `Name` are checked in debug builds (e.g., if you try to call NameRef.raw() on a non-UTF8 Name, an ENFORCE will trigger).

### Motivation

Prerequisite for breaking up the name table.

### Test plan

Covered by existing tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1367,14 +1367,14 @@ TreePtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) {
             },
             [&](parser::Complex *complex) {
                 auto kernel = MK::Constant(loc, core::Symbols::Kernel());
-                core::NameRef complex_name = core::Names::Constants::Complex().cnst(dctx.ctx)->original;
+                core::NameRef complex_name = core::Names::Constants::Complex().dataCnst(dctx.ctx)->original;
                 core::NameRef value = dctx.ctx.state.enterNameUTF8(complex->value);
                 auto send = MK::Send2(loc, std::move(kernel), complex_name, MK::Int(loc, 0), MK::String(loc, value));
                 result = std::move(send);
             },
             [&](parser::Rational *complex) {
                 auto kernel = MK::Constant(loc, core::Symbols::Kernel());
-                core::NameRef complex_name = core::Names::Constants::Rational().cnst(dctx.ctx)->original;
+                core::NameRef complex_name = core::Names::Constants::Rational().dataCnst(dctx.ctx)->original;
                 core::NameRef value = dctx.ctx.state.enterNameUTF8(complex->val);
                 auto send = MK::Send1(loc, std::move(kernel), complex_name, MK::String(loc, value));
                 result = std::move(send);
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1367,14 +1367,14 @@ TreePtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) {
             },
             [&](parser::Complex *complex) {
                 auto kernel = MK::Constant(loc, core::Symbols::Kernel());
-                core::NameRef complex_name = core::Names::Constants::Complex().data(dctx.ctx)->cnst.original;
+                core::NameRef complex_name = core::Names::Constants::Complex().cnst(dctx.ctx)->original;
                 core::NameRef value = dctx.ctx.state.enterNameUTF8(complex->value);
                 auto send = MK::Send2(loc, std::move(kernel), complex_name, MK::Int(loc, 0), MK::String(loc, value));
                 result = std::move(send);
             },
             [&](parser::Rational *complex) {
                 auto kernel = MK::Constant(loc, core::Symbols::Kernel());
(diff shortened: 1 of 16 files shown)


core/NameRef.h
```diff
@@ -74,9 +104,11 @@ class NameRef final : private DebugOnlyCheck<NameRefDebugCheck> {
         return _id;
     }
 
-    NameData data(GlobalState &gs) const;
+    NameKind kind(const GlobalState &gs) const;
```
jvilk-stripe, line 107: In the next change (breaking up the name table), `kind` will be stored in the `NameRef` directly and will not require a GlobalState argument.

core/Names.h
```diff
@@ -11,13 +11,6 @@
 namespace sorbet::core {
 class GlobalState;
 class Name;
-enum class NameKind : u1 {
```
jvilk-stripe, line 14: Moved to `NameRef.h`.

ast/desugar/Desugar.cc
```diff
@@ -1367,14 +1367,14 @@ TreePtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) {
             },
             [&](parser::Complex *complex) {
                 auto kernel = MK::Constant(loc, core::Symbols::Kernel());
-                core::NameRef complex_name = core::Names::Constants::Complex().data(dctx.ctx)->cnst.original;
+                core::NameRef complex_name = core::Names::Constants::Complex().cnst(dctx.ctx)->original;
```
jez, line 1370: I anticipate wanting to update the docs to talk about why we have all these `NameRef` classes. (NameRef is a key data structure, so it's important to be documented in internals.md). If these are called `cnstData` or `dataCnst`, it will be easier to refer to them in the docs as "all these `.data*(...)` methods" and then they can figure out which are available. Also easier to grep for.

Can we update the name to make it more apparent that these helpers are more related to each other than they appear currently?

core/GlobalState.cc
```diff
@@ -1390,10 +1388,10 @@ NameRef GlobalState::lookupNameConstant(NameRef original) const {
...
         return core::NameRef::noName();
     }
-    ENFORCE(original.data(*this)->kind == NameKind::UTF8 ||
-                (original.data(*this)->kind == NameKind::UNIQUE &&
-                 (original.data(*this)->unique.uniqueNameKind == UniqueNameKind::ResolverMissingClass ||
-                  original.data(*this)->unique.uniqueNameKind == UniqueNameKind::TEnum)),
+    ENFORCE(original.kind(*this) == NameKind::UTF8 ||
+                (original.kind(*this) == NameKind::UNIQUE &&
+                 (original.unique(*this)->uniqueNameKind == UniqueNameKind::ResolverMissingClass ||
+                  original.unique(*this)->uniqueNameKind == UniqueNameKind::TEnum)),
```
jez, line 1394: These `kind ==` checks are redundant now, right, because the `.unique()` call will check that?

We might want to get rid of those for brevity.

core/NameRef.h
```diff
@@ -74,9 +104,11 @@ class NameRef final : private DebugOnlyCheck<NameRefDebugCheck> {
...
 
-    const NameData data(const GlobalState &gs) const;
+    const UniqueNameData unique(const GlobalState &gs) const;
+    const RawNameData raw(const GlobalState &gs) const;
```
jez, line 110: What is a RawName? Does this correspond to a `utf8` name kind? Can we pick one name and use that everywhere (in the enum class and the `*Data` class)?

jez: (I'm fine pushing this into a separate PR if you want)

jez (approved):
mostly lgtm. some comments

ast/desugar/Desugar.cc
jvilk-stripe: > Can we update the name to make it more apparent that these helpers are more related to each other than they appear currently?

Sure, I can do that.

> I anticipate wanting to update the docs to talk about why we have all these NameRef classes.

? These classes already existed before.

core/GlobalState.cc
jvilk-stripe: good point

core/NameRef.h
jvilk-stripe: Yeah, I debated calling it `UTF8Name`. I can change that easily. I called it `RawName` because the old union type property was `.raw`.

ast/desugar/Desugar.cc
jez: > These classes already existed before

Gotcha. I only just now noticed it, and found it confusing.

Feel free to not change in this PR, but I don't really know why we have `RawName` vs `UTF8` name kind and equate the two. I can add that to a list of changes to make in my free time if you don't want to do it.

jvilk-stripe: raw == utf8. I just pushed a change that unifies the terminology.

core/NameRef.h
jvilk-stripe: I went all-in on UTF8. Pushed.

PR #3766, opened 2020-12-07
Introduce methods for checking type of symbol on SymbolRef

jvilk-stripe:
Introduce methods for checking type of symbol on SymbolRef. Puts the `kind` bits in `SymbolRef` to work! Avoids a `.data` call in the common case.

### Motivation

Prerequisite for introducing different classes for different types of symbols, which will save memory and make code safer to write.

Here's the path I'm envisioning:
* Introducing `ClassOrModuleRef`/`MethodRef`, and change code paths that expect one specific type of symbol to accept them instead of the general purpose `SymbolRef`.
  * `SymbolRef` will have `asMethod()` / etc methods to convert to a symbol-specific ref.
* Remove `.data()` on `SymbolRef`, and have symbol type-specific methods (`.klass`, `.method`, etc.).
  * Methods common to all symbol types will be moved up to SymbolRef for convenience.
  * We cannot have `.data()` when/if different symbol types do not share a parent class.
  * Naturally, the symbol type-specific ref classes (e.g., `ClassOrModuleRef`) will continue to have a `.data()` method. I expect most call sites to be using these.
* Break up `Symbol` class into separate classes for `ClassOrModule`, `Method`, etc.
  * By this point, most of the changes will be in `GlobalState`, which has a some symbol-type-agnostic code dealing with flags.

### Test plan

See included automated tests.


core/SymbolRef.h
```diff
@@ -75,6 +75,29 @@ class SymbolRef final {
...
+        return kind() == Kind::Field;
+    }
+
+    inline bool isTypeArgument() const {
+        return kind() == Kind::TypeArgument;
+    }
+
+    inline bool isTypeMember() const {
+        return kind() == Kind::TypeMember;
+    }
```
jez, line 96: Knowing that we have an enum class for this kind now, I kind of want to start moving away from these `isFoo` helpers and towards forcing people to pattern match at the call sites (for exhaustiveness, because most people don't realize that they want to handle more symbol kinds than they start out handling).

Maybe i'll try that some time.

core/Symbols.cc
```diff
@@ -155,6 +155,13 @@ SymbolRef Symbol::ref(const GlobalState &gs) const {
     return SymbolRef(gs, type, distance);
 }
 
+bool SymbolRef::isField(const GlobalState &gs) const {
+    return isFieldOrStaticField() && data(gs)->isField();
+}
+bool SymbolRef::isStaticField(const GlobalState &gs) const {
+    return isFieldOrStaticField() && data(gs)->isStaticField();
+}
```
jez, line 163: Why are these two special?

core/types/calls.cc
```diff
@@ -607,7 +607,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
                                 addedAutocorrect = true;
                             }
                         } else {
-                            const auto replacement = possibleSymbol->name.toString(gs);
+                            const auto replacement = possibleSymbol.data(gs)->name.toString(gs);
```
jez, line 610: Why does the `.data(gs)` get added here but removed in most other places?

core/SymbolRef.h
```diff
@@ -75,6 +75,29 @@ class SymbolRef final {
...
+        return kind() == Kind::ClassOrModule;
+    }
+
+    inline bool isMethod() const {
+        return kind() == Kind::Method;
+    }
+
+    inline bool isFieldOrStaticField() const {
+        return kind() == Kind::Field;
+    }
```
jez, line 88: Why is `Kind::Field` not `isField()`? Should we rename `Kind::Field` to `Kind::FieldOrStaticField`?

Why is the distinction between field and static field not reflected in the enum class?

jez (approved):
This looks mostly fine to me but i'm confused about a couple things.

jez:
With respect to long term plan, I am slightly worried that this will make it easier for features to be implemented on a really specific thing when they mean to work for many kinds of symbols. For example, Usually if you're doing something for classes and modules, you also want to handle static fields (non-module constants) and type members ("generic" static fields). For example, this is how constant completion works.

A scenario I'd like to avoid is:

- someone thinks they only need to handle a `ClassOrModuleRef` case, and threads that through a ton of methods.
- later, someone realizes that they want to handle `StaticFieldRef` and `TypeMemberRef` too.
- the code is already passing these specific references along, and now it's a huge rewrite to thread the potential for "it could be one of three things" through everywhere it needs need to be threaded, so they don't do the refactor.

core/Symbols.cc
jvilk-stripe: Good question. I think it would work in the header too. I can change that.

core/types/calls.cc
jvilk-stripe: `possibleSymbol` is a `SymbolRef` now and not a `SymbolData`.

core/SymbolRef.h
jvilk-stripe: Good point. I can make that change. While working on this PR I regretted lumping fields and static fields together in my original break-up-symbol-table PR.

jvilk-stripe:
> it's a huge rewrite to thread the potential for "it could be one of three things" through everywhere it needs need to be threaded, so they don't do the refactor.

This is the current situation though; we have tons of enforces that check that something is the proper symbol type. I don't know if going from `ClassOrModuleRef` to `SymbolRef` would be a huge rewrite; just change it to `SymbolRef`, fight the compiler, and insert ENFORCEs as needed.

I'm hoping that this change makes people more conscious of the types of symbols being passed around. `SymbolRef` is still going to be useful for a wide variety of use cases (especially in-editor). Any methods that are useful for ~all symbol types can be lifted to `SymbolRef` for convenience (e.g., `show`, `toString`).

fwiw, when I first started working on sorbet I tripped on what a symbolref can and cannot be in various parts of the codebase. I still trip on it (e.g., can owner be a non-class? yes it can). I hope that the more-specific refs can permeate the codebase to self-document places where the symbol type is truly constrained.

core/SymbolRef.h
jvilk-stripe: I'm 👍 to that.

core/SymbolRef.h
jvilk-stripe: Fixed.

core/Symbols.cc
jvilk-stripe: Ah, it's in the cc file because we don't pull in `Symbols.h` so I can't access the incomplete symbol type.

PR #3505, opened 2020-10-13
Error out on non-private calls to private methods

bss:
This will type-check private methods and ensure they can actually be called. Based off of https://github.com/sorbet/sorbet/compare/jez-private-methods

The following can still not be type-checked:
- splat/splat&block/block methods since they do not carry visibility information when being switched to the magic replacements.
- "subsequent visibility": using the private keyword, thus marking the following methods private.

I intend to work on those separately though.

### Motivation
Currently sorbet does not warn users about calling a private method on an object, instead a runtime error will occur. This will make sorbet error out  instead.

### Test plan
See included automated tests.

PR #3767, opened 2020-12-07
Ignore private constants inside RBI files

jeffcarbs:
This came out of a [discussion in slack](https://sorbet-ruby.slack.com/archives/CFT8Y4909/p1606946244031900?thread_ts=1605834003.026000&cid=CFT8Y4909) about referencing private constants in RBI. 

### Motivation

_Mostly extracted from that thread_

I feel like it might have to be all-or-nothing (i.e. either you can reference private constants with any usage inside RBI or you can't). If it's allowed then you can always generate "compact" RBI:

```ruby
module Public::Something
  include Public::PrivateModule # Currently error: Non-private reference

  sig { returns(Public::PrivateClass) } # Currently error: Non-private reference
  def foo; end
end
```

if it's not allowed in even just one circumstance then you'd have to always generate the "nested" RBI (or somehow intelligently know what's private or not) to be safe:

```ruby
module Public
  class Something
    include PrivateModule

    sig { returns(PrivateClass) }
    def foo; end
  end
end
```

The argument for why it ought to be allowed is that RBIs simply tell sorbet what already exists. It happens to be valid ruby which is nice for readers/writers but aside from being valid syntax it doesn't have to run without error. For instance:
- you don't need `extend T::Sig` when you're defining sigs in RBI but in source code that would raise an exception
- you can define classes out of order (e.g. `class A < B; end` can be defined before `class B; end` even though that would raise at runtime)

The argument against is that it might be a slippery slope of behavior differences between parsing source code vs RBI.

It seemed agreeable to simply allow it in RBI so that's what this PR does.

### Test plan

I'm not sure how to test RBI but I'd be happy to add some if someone wants to point me in the right direction.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -242,10 +242,11 @@ private:
             core::SymbolRef result = resolved.data(ctx)->findMember(ctx, c.cnst);
 
             // Private constants are allowed to be resolved, when there is no scope set (the scope is checked above),
-            // otherwise we should error out:
+            // otherwise we should error out. Private constant references _are not_ enforced inside RBI files.
             if (result.exists() &&
                 ((result.data(ctx)->isClassOrModule() && result.data(ctx)->isClassOrModulePrivate()) ||
-                 (result.data(ctx)->isStaticField() && result.data(ctx)->isStaticFieldPrivate()))) {
+                 (result.data(ctx)->isStaticField() && result.data(ctx)->isStaticFieldPrivate())) &&
+                !ctx.file.data(ctx).isRBI()) {
                 if (auto e = ctx.beginError(c.loc, core::errors::Resolver::PrivateConstantReferenced)) {
                     e.setHeader("Non-private reference to private constant `{}` referenced",
                                 result.data(ctx)->show(ctx));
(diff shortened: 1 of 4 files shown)


jez:
> I'm not sure how to test RBI but I'd be happy to add some if someone wants to point me in the right direction.

RBI tests are the same as all the other testdata tests like you wrote for private methods and private constants, but just give the test files `*.rbi` extensions.

jez:
#3505 #3565 have some tests that you should be able to model your tests off of.

jeffcarbs:
@jez - Thanks for the info! Just added tests so this should be good to go 👍 

I realized that an RBI-only test wouldn't be picked up so I needed to add an accompanying `.rb` file. In other words, with
- only `testdata/infer/private_constant_in_rbi.rbi` I got a `no such target` when trying to run the test.
- both `testdata/infer/private_constant_in_rbi__1.rbi` and `testdata/infer/private_constant_in_rbi__2.rb` it ran as expected

This was probably for the best since I was able to add tests ensuring that when you have private constants defined in RBI, referencing them in a regular ruby file would be a type error (as expected).

paracycle:
@jeffcarbs Thanks for changing this behaviour since we also ran into some edge cases like this when using RBI files generated for the DSLs in our codebase.

@jez If this change is good, it would be great to have it merged sooner than later, since it will unblock us from doing our weekly Sorbet version bump. Thanks.

jez (approved):

jez:
Thanks!

PR #3770, opened 2020-12-08
Encode NameKind into NameRef, and remove special 0th index name

jvilk-stripe:
Encode NameKind into NameRef, and remove special 0th index name.

This PR prepares Sorbet for a subsequent change which will break up the name table into three vectors (one per type of name).

This PR contains the following changes:
* Encodes `NameKind` into `NameRef`, so one can call `nameref.kind()` to determine a name's kind.
* Changes `namesByHash` to store tuples of name hash and a name ref's raw ID, which contains the name's kind and index in the name table.
* Removes the special Name at the 0th index of `gs.names`.
  * Since `kind` is stored in NameRef, we can use the special value `0` (kind of 0 and index of 0) in NameRef to represent noName.
  * The actual name at index 0 has a non-zero NameRef ID since the ID contains the kind and the index.

### Motivation

Prerequisite for splitting up the name table for memory savings.

### Test plan

See included automated tests.


core/NameRef.h
```diff
@@ -82,14 +96,29 @@ class NameRef final : private DebugOnlyCheck<NameRefDebugCheck> {
     // generated by tools/generate_names.cc
     struct WellKnown {};
 
-    constexpr NameRef(WellKnown, u4 id) : _id{id} {}
+    constexpr NameRef(WellKnown, NameKind kind, u4 id) : _id{id | (static_cast<u4>(kind) << ID_BITS)} {
+        // If this fails, the symbol table is too big :(
+        ENFORCE_NO_TIMER(id <= ID_MASK);
```
elliottt, line 101: Is the decision to not mask off id intentional? That case would be caught by the enforce, but only if it shows up in a test somewhere. (I'm not really sure what a great way to test this would be, perhaps a one off test in c++ for name refs?)

core/NameRef.h
```diff
@@ -114,7 +162,7 @@ class NameRef final : private DebugOnlyCheck<NameRefDebugCheck> {
 
     // Returns the `0` NameRef, used to indicate non-existence of a name
     static NameRef noName() {
-        return NameRef(WellKnown{}, 0);
+        return NameRef();
     }
```
elliottt, line 166: Any reason to not mark `NameRef()` and this function as `constexpr`?

parser/Builder.cc
```diff
@@ -146,7 +146,7 @@ class Builder::Impl {
 
     unique_ptr<Node> accessible(unique_ptr<Node> node) {
         if (auto *id = parser::cast_node<Ident>(node.get())) {
-            ENFORCE(id->name.kind(gs_) == core::NameKind::UTF8);
+            ENFORCE(id->name.kind() == core::NameKind::UTF8);
```
elliottt, line 149: `:chef-kiss:`

core/Context.cc
```diff
@@ -111,34 +111,28 @@ GlobalSubstitution::GlobalSubstitution(const GlobalState &from, GlobalState &to,
     }
 
     if (!fastPath || debug_mode) {
-        bool seenEmpty = false;
```
elliottt, line 114: I feel like we should have been able to catch this weird dead code sooner.

core/NameRef.h
```diff
@@ -70,9 +71,22 @@ class NameRef final : private DebugOnlyCheck<NameRefDebugCheck> {
...
+        return _id & ID_MASK;
+    }
+
 public:
     friend GlobalState;
     friend Name;
+    // TODO(jvilk): Delurk this friend once name table is broken up.
+    friend GlobalSubstitution;
 
     NameRef() : _id(0){};
```
elliottt, line 91: It might be worth putting a comment on this constructor to document the fact that the value `0` now also implies that there's no kind tag present, which is why it's OK to use value as the default for all the NameRef kinds.

core/Names.cc
```diff
@@ -14,7 +14,16 @@ using namespace std;
 
 namespace sorbet::core {
 
-NameRef::NameRef(const GlobalState &gs, u4 id) : DebugOnlyCheck(gs, id), _id{id} {}
+NameRef::NameRef(const GlobalState &gs, NameKind kind, u4 id)
+    : DebugOnlyCheck(gs, id), _id{id | (static_cast<u4>(kind) << ID_BITS)} {
+    // If this fails, the symbol table is too big :(
+    ENFORCE_NO_TIMER(id <= ID_MASK);
```
elliottt, line 20: Same questions as in `NameRef.h` about not masking the id.

core/tools/generate_names.cc
```diff
@@ -486,15 +487,15 @@ void emit_register(ostream &out) {
...
+        out << "    ENFORCE(" << name.srcName << "_id." << (name.isConstant ? "constantIndex" : "utf8Index")
+            << "() == " << name.id << "); /* " << name.srcName << "() */" << '\n';
     }
     out << '\n';
     out << "}" << '\n';
 }
 
 int main(int argc, char **argv) {
-    int i = 1;
+    int i = 0;
```
elliottt, line 498: This is OK now because the tag bits differentiate the valid name from the `noName` case, right?

core/GlobalState.cc
```diff
@@ -1520,21 +1521,21 @@ NameRef GlobalState::freshNameUnique(UniqueNameKind uniqueNameKind, NameRef orig
...
-    names[idx].kind = NameKind::UNIQUE;
-    names[idx].unique.num = num;
-    names[idx].unique.uniqueNameKind = uniqueNameKind;
-    names[idx].unique.original = original;
-    ENFORCE(NameRef(*this, idx).hash(*this) == hs);
+    nameData.kind = NameKind::UNIQUE;
+    nameData.unique.num = num;
+    nameData.unique.uniqueNameKind = uniqueNameKind;
+    nameData.unique.original = original;
+    ENFORCE(name.hash(*this) == hs);
```
elliottt, line 1535: Would it be worth adding a constructor to `Name` for this case, so that you can just `emplace_back` the arguments to it?

elliottt (approved):
Looks great, just a few questions :)

core/NameRef.h
jvilk-stripe: Not intentional. I can mask it off.

core/NameRef.h
jvilk-stripe: No good reason; I can do that.

core/Context.cc
jvilk-stripe: Ah this isn't dead code. This handles the special 0-indexed case which I have removed!

core/tools/generate_names.cc
jvilk-stripe: Yup.

core/Context.cc
elliottt: Ah yep -- I misread. Thanks!

core/GlobalState.cc
jvilk-stripe: `Name` is going away real soon, so I'm not terribly invested in improving its ergonomics.

core/NameRef.h
jvilk-stripe: Done!

core/NameRef.h
jvilk-stripe: Done!

core/NameRef.h
jvilk-stripe: Done

core/Names.cc
jvilk-stripe: Done

PR #3771, opened 2020-12-09
Make ResolveSignaturesWalk a ShallowMap and small code tweaks.

jvilk-stripe:
Make ResolveSignaturesWalk a ShallowMap and small code tweaks. The `send` nodes that `ResolveSignaturesWalk` was processing are readily available earlier in the type params pass.

Addresses lingering feedback in https://github.com/sorbet/sorbet/pull/3762 from @jez and @elliottt.

Should result in a speedup as `ResolveSignaturesWalk` is a serial pass over all of the ASTs, so it doing less work is significant on the critical path.

### Motivation

Speed.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1041,10 +1041,6 @@ public:
 };
 
 class ResolveTypeMembersAndFieldsWalk {
-    static core::TypePtr todoType() {
-        return core::make_type<core::ClassType>(core::Symbols::todo());
-    }
-
     // A type_member, type_template, or T.type_alias that needs to have types
     // resolved.
     struct ResolveAssignItem {
@@ -1282,7 +1278,7 @@ class ResolveTypeMembersAndFieldsWalk {
         typecase(
             expr, [&](const ast::Literal &a) { result = a.value; },
             [&](const ast::Cast &cast) {
-                if (cast.type != todoType() && cast.cast != core::Names::let() &&
+                if (cast.type != core::Types::todo() && cast.cast != core::Names::let() &&
                     cast.cast != core::Names::uncheckedLet()) {
                     if (auto e = ctx.beginError(cast.loc, core::errors::Resolver::ConstantAssertType)) {
                         e.setHeader("Use `{}` to specify the type of constants", "T.let");
@@ -1295,7 +1291,7 @@ class ResolveTypeMembersAndFieldsWalk {
         return result;
     }
 
-    // Tries to resolve the given static field. Returns todoType() if it is unable to resolve the field.
+    // Tries to resolve the given static field. Returns Types::todo() if it is unable to resolve the field.
     [[nodiscard]] static core::TypePtr tryResolveStaticField(core::Context ctx, ResolveStaticFieldItem &job) {
         ENFORCE(job.sym.data(ctx)->isStaticField());
         auto &asgn = job.asgn;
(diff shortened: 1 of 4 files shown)


jvilk-stripe:

→ https://go/builds/bui_IXaxybc2eGsN5Y

resolver/resolver.cc
```diff
@@ -1282,8 +1289,11 @@ class ResolveTypeMembersAndFieldsWalk {
         typecase(
             expr, [&](const ast::Literal &a) { result = a.value; },
             [&](const ast::Cast &cast) {
-                if (cast.type != todoType() && cast.cast != core::Names::let() &&
-                    cast.cast != core::Names::uncheckedLet()) {
+                if (cast.type == core::Types::todo()) {
+                    return;
```
jez, line 1293: I believe this is not quite the same as before unless you set `result = cast.type` before the early return

jvilk-stripe: You're correct. In this case it returns `nullptr` which is an expected return value for "we failed to extract the type".

For example, see this callsite:

```c++
            if (auto resultType = resolveConstantType(ctx, asgn->rhs)) {
                return resultType;
            } else {
                return core::Types::todo();
            }
```

If `resolveConstantType` returns `todo()`, the caller returns `todo()`. If it's `nullptr`... the caller returns `todo()`.

resolver/resolver.cc
```diff
@@ -1576,6 +1624,172 @@ class ResolveTypeMembersAndFieldsWalk {
...
+            // Infer will report a type error
+            return;
+        }
+
+        core::TypePtr packageType = nullptr;
+        optional<core::LocOffsets> packageLoc;
+        if (send.hasKwArgs()) {
+            // this means we got the third package arg
+            auto *key = ast::cast_tree<ast::Literal>(send.args[posEnd]);
+            if (!key || !key->isSymbol(ctx) || key->asSymbol(ctx) != ctx.state.lookupNameUTF8("package")) {
```
jez, line 1668: Dang this should never have been written this way—we should have `"package"` in `generate_names.cc`.

jvilk-stripe: (Note: I didn't write this; I'm just moving it.)

jvilk-stripe: I can spruce this up in a subsequent PR.

jez: yep, i should have commented on the red part of the diff to make that clear

resolver/resolver.cc
```diff
@@ -1041,10 +1041,6 @@ class ResolveConstantsWalk {
 };
 
 class ResolveTypeMembersAndFieldsWalk {
```
elliottt, line 1043: I'm beginning to think that we should change the name of this pass :D

elliottt (approved):
🚢

resolver/resolver.cc
```diff
@@ -1295,22 +1305,21 @@ class ResolveTypeMembersAndFieldsWalk {
...
         auto data = job.sym.data(ctx);
         if (data->resultType == nullptr) {
-            auto resultType = resolveConstantType(ctx, asgn->rhs);
-            if (!resultType) {
-                return todoType();
-            } else {
+            if (auto resultType = resolveConstantType(ctx, asgn->rhs)) {
                 return resultType;
+            } else {
+                return core::Types::todo();
```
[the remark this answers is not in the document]
jvilk-stripe: Thanks, done.

PR #3774, opened 2020-12-09
Add missing static error for abstract.override

jez:
### Motivation

This already raises at runtime. We should have a static error. This caused a
problem in production at Stripe.

### Test plan

See included automated tests.


elliottt (approved):
🚢

PR #3778, opened 2020-12-10
Temporarily ignore packages that aren't found in non_forcing_is_a?

aisamanra:
### Motivation
Right now, it's not possible for code to be simultaneously valid packaged or valid unpackaged if it uses `non_forcing_is_a?`. This temporary change allows us to "fall back" to the unpackaged workflow. We probably want to re-enable it in the future, since "simultaneously unpackaged and packaged" is not a state that will make sense once we finish implementing the runtime: it's only transitional.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/packager/non_forcing_constants/foo/foo_methods.rb
+++ b/test/testdata/packager/non_forcing_constants/foo/foo_methods.rb
@@ -51,7 +51,9 @@ module FooMethods
 
   sig {params(arg: T.untyped).returns(T::Boolean)}
   def bad_non_existent_package(arg)
-    if T::NonForcingConstants.non_forcing_is_a?(arg, "Bar", package: "Project::Quux") # error: Unable to find package
+    # this is /temporarily/ okay because of #3778, but once we back it
+    # out this will fail with `Unable to find package`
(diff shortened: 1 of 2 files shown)


elliottt (approved):

elliottt:
Could you add a test that will fail when this is switched back over?

aisamanra:
The test is changed! This test will have to be fixed when we walk this change back out.

test/testdata/packager/non_forcing_constants/foo/foo_methods.rb
```diff
@@ -51,7 +51,9 @@ def bad_check_for_global_bar(arg)
 
   sig {params(arg: T.untyped).returns(T::Boolean)}
   def bad_non_existent_package(arg)
-    if T::NonForcingConstants.non_forcing_is_a?(arg, "Bar", package: "Project::Quux") # error: Unable to find package
+    # this is /temporarily/ okay because of #3778, but once we back it
+    # out this will fail with `Unable to find package`
+    if T::NonForcingConstants.non_forcing_is_a?(arg, "Bar", package: "Project::Quux")
```
elliottt, line 56: Thanks!

elliottt (approved):
🚢

PR #3780, opened 2020-12-11
Parallelize ResolveSignaturesWalk.

jvilk-stripe:
Parallelize ResolveSignaturesWalk.

It gathers parsed sigs during the parallel walk, and resolves them serially.

Also changes the `argsToKeep` argument to `enterNewMethodOverload` into a `vector<bool>` to make it smaller and cheaper to look up arguments in.

### Motivation

Speed++

### Test plan

Covered by existing tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2282,7 +2282,7 @@ private:
         }
     }
 
-    static void handleAbstractOrInterfaceMethods(core::Context ctx, ast::MethodDef &mdef) {
+    static void handleAbstractMethod(core::Context ctx, ast::MethodDef &mdef) {
         if (mdef.symbol.data(ctx)->isAbstract()) {
             if (!ast::isa_tree<ast::EmptyTree>(mdef.rhs)) {
                 if (auto e = ctx.beginError(mdef.rhs.loc(), core::errors::Resolver::AbstractMethodWithBody)) {
@@ -2639,7 +2639,7 @@ private:
                     // OVERLOAD
                     lastSigs.clear();
                 } else {
-                    handleAbstractOrInterfaceMethods(ctx, mdef);
+                    handleAbstractMethod(ctx, mdef);
                 }
             },
             [&](const ast::ClassDef &cdef) {
@@ -2681,7 +2681,7 @@ public:
             }
             fillInInfoFromSig(ctx, overloadSym, sig.loc, move(sig.sig), isOverloaded, mdef);
         }
-        handleAbstractOrInterfaceMethods(ctx, mdef);
+        handleAbstractMethod(ctx, mdef);
     }
 
     ast::TreePtr postTransformClassDef(core::Context ctx, ast::TreePtr tree) {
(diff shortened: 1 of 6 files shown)


jvilk-stripe:

→ https://go/builds/bui_IYKLEfJ7vwOcnw

resolver/resolver.cc
```diff
@@ -2245,9 +2245,29 @@ class ResolveTypeMembersAndFieldsWalk {
...
+    struct MethodSignature {
+        core::LocOffsets loc;
+        ParsedSig sig;
+        vector<bool> argsToKeep;
+    };
+    struct ResolveMethodSignaturesJob {
+        core::FileRef file;
+        core::SymbolRef owner;
+        ast::MethodDef *mdef;
+        InlinedVector<MethodSignature, 1> sigs;
```
jvilk-stripe, line 2258: My next move will be to try seeing if separating overloaded sigs from non-overloaded sigs results in a speed improvement. I suspect it will, as non-overloaded sigs will not require a vector here (or a `vector<bool>` within it).

resolver/resolver.cc
```diff
@@ -2262,8 +2282,64 @@ class ResolveSignaturesWalk {
         }
     }
 
-    void fillInInfoFromSig(core::MutableContext ctx, core::SymbolRef method, core::LocOffsets exprLoc, ParsedSig sig,
-                           bool isOverloaded, const ast::MethodDef &mdef) {
+    static void handleAbstractOrInterfaceMethods(core::Context ctx, ast::MethodDef &mdef) {
```
jez, line 2285: What is an interface method? afaik there are only abstract methods.

resolver/resolver.cc
```diff
@@ -2510,127 +2586,60 @@ class ResolveSignaturesWalk {
...
-                    }
-                    int i = 0;
-
                     // process signatures in the context of either the current
                     // class, or the current singleton class, depending on if
                     // the current method is a self method.
                     core::SymbolRef sigOwner;
                     if (mdef.flags.isSelfMethod) {
-                        sigOwner = ctx.owner.data(ctx)->singletonClass(ctx);
+                        sigOwner = ctx.owner.data(ctx)->lookupSingletonClass(ctx);
```
jez, line 2604: When do we create the singletonClass? (e.g., why do we expect it to already exist?)

jez (approved):

resolver/resolver.cc
jvilk-stripe: Poorly named method; I can rename it 😓  `handleAbstractMethods` captures it since methods defined on an interface are de-facto abstract.

resolver/resolver.cc
jvilk-stripe: Great question. Let me verify that we explicitly create the expected singletons.

jvilk-stripe: `namer` creates it. It creates a singleton class for every class def, and ENFORCEs that all defined methods have an owner (and it handles the `self` case).

jvilk-stripe: I added an ENFORCE and a comment just in case

PR #3785, opened 2020-12-11
Avoid allocating a string and string comparison for every parameter in every sig.

jvilk-stripe:
Avoid allocating a string and performing string comparison for every parameter in every sig.

Compare NameRefs instead.

### Motivation

Speed up resolver.

### Test plan

Covered by existing tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2487,8 +2487,7 @@ private:
                 auto param = defParams[j];
                 auto sname = spec.name;
                 auto dname = param->localVariable._name;
-                // Common case: NameRefs match. Uncommon case: NameRefs don't match but their show strings are
-                // identical.
+                // TODO(jvilk): Do we need to check .show? Typically NameRef equality is equal to string equality.
                 if (sname != dname && sname.show(ctx) != dname.show(ctx)) {
                     if (auto e = ctx.beginError(param->loc, core::errors::Resolver::BadParameterOrdering)) {
                         e.setHeader("Bad parameter ordering for `{}`, expected `{}` instead", dname.show(ctx),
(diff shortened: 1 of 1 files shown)


jvilk-stripe:

→ https://go/builds/bui_IYRzhvaBwmZKPp

resolver/resolver.cc
```diff
@@ -2485,11 +2485,14 @@ class ResolveSignaturesWalk {
             int j = 0;
             for (auto spec : sigParams) {
                 auto param = defParams[j];
-                auto sname = spec.name.show(ctx);
-                auto dname = param->localVariable._name.show(ctx);
-                if (sname != dname) {
+                auto sname = spec.name;
+                auto dname = param->localVariable._name;
+                // Common case: NameRefs match. Uncommon case: NameRefs don't match but their show strings are
+                // identical.
```
jvilk-stripe, line 2491: I actually don't even know if namerefs can be != and show be ==, but I don't want to break anything.

elliottt: How hard would it be to generate a coverage report? It would be interesting to see if any of our test suite triggers this case.

elliottt (approved):
Nice find!

resolver/resolver.cc
jvilk-stripe: I can just insert an ENFORCE and see what happens.

jvilk-stripe: it, uh, it never triggers that case.

jvilk-stripe: I'll insert a comment about it.

elliottt: Nice!

PR #3787, opened 2020-12-12
Break up name table by name type.

jvilk-stripe:
Break up name table by name type.

The name table currently contains a vector of 24-byte Name objects. A full 8 bytes of this object is devoted to a one-byte tag, which indicates if the name is a UTF8Name, ConstantName, or a UniqueName. Furthermore, ConstantName and UniqueNames take fewer than 16 bytes and have even more padding.

This PR splits the name table into separate vectors for UTF8Names, ConstantNames, and UniqueNames. UTF8Names are 16 bytes (33% savings), ConstantNames are 4 bytes (83% savings), and UniqueNames are 12 bytes (50% savings).

On large codebases typechecked on machines with many cores and with a cached name table, this PR reduces max RSS induced during the parallel indexing phase, which duplicates the name table once per core. This change also causes a small typechecking speedup.

### Motivation

Reduce max RSS.

### Test plan

Covered by existing tests. The GlobalSubstitution test has been updated to cover an additional corner case.

### DIFF
--- a/core/Context.cc
+++ b/core/Context.cc
@@ -150,10 +150,9 @@ GlobalSubstitution::GlobalSubstitution(const GlobalState &from, GlobalState &to,
                     to.freshNameUnique(nm.uniqueNameKind, substitute(nm.original), nm.num));
                 ENFORCE(!fastPath || uniqueNameSubstitution.back().uniqueIndex() == i);
             }
-            i = constantNameSubstitution.size() - 1;
-            for (const ConstantName &nm : from.constantNames) {
-                i++;
+            for (i = constantNameSubstitution.size(); i < from.constantNames.size(); i++) {
                 ENFORCE_NO_TIMER(constantNameSubstitution.size() == i, "Constant name substitution has wrong size");
+                auto &nm = from.constantNames[i];
                 constantNameSubstitution.emplace_back(to.enterNameConstant(substitute(nm.original)));
                 ENFORCE(!fastPath || constantNameSubstitution.back().constantIndex() == i);
             }
--- a/core/Context.cc
+++ b/core/Context.cc
@@ -136,6 +136,7 @@ GlobalSubstitution::GlobalSubstitution(const GlobalState &from, GlobalState &to,
                 ENFORCE(uniqueNameSubstitution.size() == i, "Unique name substitution has wrong size");
                 if (nm.original.kind() == NameKind::CONSTANT &&
                     nm.original.constantIndex() >= constantNameSubstitution.size()) {
+                    // Note: Duplicate of loop body below. If you change one, change the other!
                     for (u4 i = constantNameSubstitution.size(); i <= nm.original.constantIndex(); i++) {
                         auto &cnst = from.constantNames[i];
                         ENFORCE_NO_TIMER(constantNameSubstitution.size() == i,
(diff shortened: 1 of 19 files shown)


jvilk-stripe:

→ https://go/builds/bui_IYlVmhxZtXgGkE

jvilk-stripe:

→ https://go/builds/bui_IYoSsf8pCVlvNY

jvilk-stripe:
https://go/build/bui_IZYNicyaPwQNrx shows a passing Stripe build.

core/Context.cc
```diff
@@ -117,26 +117,44 @@ GlobalSubstitution::GlobalSubstitution(const GlobalState &from, GlobalState &to,
...
+                ENFORCE(uniqueNameSubstitution.size() == i, "Unique name substitution has wrong size");
+                if (nm.original.kind() == NameKind::CONSTANT &&
+                    nm.original.constantIndex() >= constantNameSubstitution.size()) {
+                    for (u4 i = constantNameSubstitution.size(); i <= nm.original.constantIndex(); i++) {
+                        auto &cnst = from.constantNames[i];
+                        ENFORCE_NO_TIMER(constantNameSubstitution.size() == i,
+                                         "Constant name substitution has wrong size");
+                        // N.B.: cnst may reference a UniqueName, but since names are linearizeable we should have
+                        // already substituted it by now.
+                        constantNameSubstitution.emplace_back(to.enterNameConstant(substitute(cnst.original)));
```
elliottt, line 145: It's a little unfortunate that this loop body gets duplicated below. It's probably worth adding a warning that changing one will require changing the other.

core/GlobalState.cc
```diff
@@ -47,6 +47,20 @@ SymbolRef GlobalState::synthesizeClass(NameRef nameId, u4 superclass, bool isMod
...
+u4 nextPowerOfTwo(u4 v) {
+    v--;
+    v |= v >> 1;
+    v |= v >> 2;
+    v |= v >> 4;
+    v |= v >> 8;
+    v |= v >> 16;
+    v++;
+    return v;
+}
```
elliottt, line 61: What do you think about exposing this through a header, and re-using it in the pickler?

core/GlobalState.cc
```diff
@@ -1755,13 +1773,26 @@ unique_ptr<GlobalState> GlobalState::deepCopy(bool keepId) const {
...
     result->dslRubyExtraArgs = this->dslRubyExtraArgs;
-    result->names.reserve(this->names.capacity());
+    result->utf8Names.reserve(this->utf8Names.capacity());
+    result->constantNames.reserve(this->constantNames.capacity());
+    result->uniqueNames.reserve(this->uniqueNames.capacity());
     if (keepId) {
-        result->names.resize(this->names.size());
-        ::memcpy(result->names.data(), this->names.data(), this->names.size() * sizeof(Name));
+        result->utf8Names.resize(this->utf8Names.size());
+        ::memcpy(result->utf8Names.data(), this->utf8Names.data(), this->utf8Names.size() * sizeof(UTF8Name));
```
elliottt, line 1781: Does this use `memcpy` for performance reasons? I wonder how different it would be with `std::copy_n` instead.

elliottt (approved):
Looks great! Just a few non-blocking questions :)

core/GlobalState.cc
jvilk-stripe: As far as I can tell, we only use this in the fuzzer for performance.

core/GlobalState.cc
jvilk-stripe: I can do that as a follow-on (this isn't new code; this duplication is present in master already. I only moved the function up in the file.)

core/Context.cc
jvilk-stripe: done

PR #3788, opened 2020-12-12
Skip syntactically invalid class names when serializing

haltwise:
When running `srb rbi hidden-definitions` in my project, I was encountering a persistent error:
```
Generating /tmp/d20201212-15780-1vce1q8/reflection.rbi with 6155 modules and 132 aliases
Printing your code's symbol table into /tmp/d20201212-15780-1vce1q8/from-source.json
Printing /tmp/d20201212-15780-1vce1q8/reflection.rbi's symbol table into /tmp/d20201212-15780-1vce1q8/reflection.json
	5: from /home/chris/.rvm/gems/ruby-2.6.6@mixdr/gems/sorbet-0.5.6134/bin/srb-rbi:237:in `<main>'
	4: from /home/chris/.rvm/gems/ruby-2.6.6@mixdr/gems/sorbet-0.5.6134/bin/srb-rbi:224:in `main'
	3: from /home/chris/.rvm/gems/ruby-2.6.6@mixdr/gems/sorbet-0.5.6134/bin/srb-rbi:232:in `block in make_step'
	2: from /home/chris/.rvm/gems/ruby-2.6.6@mixdr/gems/sorbet-0.5.6134/lib/hidden-definition-finder.rb:38:in `main'
	1: from /home/chris/.rvm/gems/ruby-2.6.6@mixdr/gems/sorbet-0.5.6134/lib/hidden-definition-finder.rb:47:in `main'
/home/chris/.rvm/gems/ruby-2.6.6@mixdr/gems/sorbet-0.5.6134/lib/hidden-definition-finder.rb:151:in `write_constants': /tmp/d20201212-15780-1vce1q8/reflection.rbi had unexpected errors. Check this file for a clue: /tmp/d20201212-15780-1vce1q8/reflection.json.err (RuntimeError)
```

The error details show it's a syntax problem in the generated file:
```
/tmp/d20201212-15780-1vce1q8/reflection.rbi:13: unexpected token tNL https://srb.help/2001
    13 |  end
    14 |  def proper_table_name(*args); end

/tmp/d20201212-15780-1vce1q8/reflection.rbi:15: unexpected token "end" https://srb.help/2001
    15 |end
...
```

And in the generated file, we can see we've written out a class name that is interpreted as a comment:
```ruby
module #<Class:0x000056403a83ee08>::ProperTableNameWithHashAwarenessSupport
  # Skipping `extend Sorbet::Private::GemGeneratorTracepoint::Tracer::ModuleOverride` because it is an invalid name
  sig do
    params(
      args: ::T.untyped,
    )
    .returns(::T.untyped)
  end
  def proper_table_name(*args); end
end
```

This appears to be related to [this file](https://github.com/jenseng/hair_trigger/blob/89eefa2afdfaf6bec634bb9c411ce4500b8fcc97/lib/hair_trigger/migrator.rb) in the `hairtrigger` gem.

I couldn't figure out the right way to reproduce this in tests; but my proposed change does solve the problem in my actual project. I'm happy to add a test case if someone can point me to the right example to emulate.

### DIFF
--- a/gems/sorbet/lib/serialize.rb
+++ b/gems/sorbet/lib/serialize.rb
@@ -253,7 +253,7 @@ class Sorbet::Private::Serialize
 
   def valid_class_name(name)
     name.split("::").each do |piece|
-      return false if piece[0].upcase != piece[0]
+      return false unless piece =~ /^[[:upper:]][[:word:]]*$/
     end
     return false if [
       'Sorbet::Private::GemGeneratorTracepoint::Tracer::ClassOverride',


gems/sorbet/lib/serialize.rb
```diff
@@ -253,7 +253,7 @@ def serialize_method(method, static=false, with_sig: true)
 
   def valid_class_name(name)
     name.split("::").each do |piece|
-      return false if piece[0].upcase != piece[0]
+      return false unless piece =~ /^[[:upper:]][[:word:]]*$/
```
haltwise, line 256: I couldn't find any authoritative reference on the question of what non-ASCII characters are legal for class names in Ruby. It seems that maybe anything goes when it comes to Unicode; but since these character classes seem to do the right thing in ASCII I'm hoping the spirit of them is consistent in other character sets, and reasonable to use here.

We could also just add a special case to exclude `#`, but there are probably other things that would cause syntax problems (although it might be harder to wind up with them).

jez: Do you think you could change this so that we filter out the things we know aren't ok, rather than attempt to do a complete validation, and potentially reject things that are fine?

For example, I think

- return false if spaces
- return false if punctuation
- return false if doesn't start with uppercase

is similar in spirit to what you've written here, but has a lower chance of breaking valid usage.

jez (changes requested):
^

gems/sorbet/lib/serialize.rb
haltwise: So, I tested this, and it works on my codebase:
```ruby
      return false unless piece =~ /^[[:upper:]]/
      return false if piece =~ /[[:space:][:cntrl:]]/
      return false if piece =~ /[[:punct:]&&[^_]]/
```

But I also noticed that `:word:` is already used in [valid_method_name](https://github.com/sorbet/sorbet/blob/a630ef3d431c2b1c4a1ff979abcc0f78a6e8af46/gems/sorbet/lib/serialize.rb#L202), so it seems like it would make sense to apply the same rule(s) in both places. Thoughts?

jez: That regex is already too restrictive. `def !(); end` and `def ==(); end` are both valid method definitions that regex rejects.

More generally: all this runtime reflection code is incredibly brittle and poorly tested. I'd rather not change things that aren't actively causing problems. Let's just leave it at what you've implemented and call it done.

haltwise: :+1: updated the PR

haltwise:
Was there anything else blocking this?

jez:
I can't reproduce this and I would prefer not to make change to `srb init` without tests because it is very brittle already.

If you can reproduce the problem in isolation, please open an issue and we can go from there.

PR #3793, opened 2020-12-14
Add a packager test for nested inner package namespaces.

jvilk-stripe:
Add a packager test for nested inner package namespaces.

### Motivation

@nroman-stripe and @jez were confused as to how the packager pass dealt with inner package namespaces that get exported. I added the following test and confirmed that exports are flattened as expected.

### Test plan

This is just a test; no code has changed. :)


jez (approved):
Thanks. Worth adding a package tree test?

jvilk-stripe:
Good point! I'll add it.

elliottt (approved):

PR #3804, opened 2020-12-16
Move nextPowerOfTwo to common and use it in GlobalState and serialize

jvilk-stripe:
Move nextPowerOfTwo to common and use it in GlobalState and serialize

### Motivation

Address feedback from a previous PR.

### Test plan

Covered by existing tests.


elliottt (approved):
Thanks for following up on this one!

PR #3807, opened 2020-12-17
Add T::Configuration#module_name_mangler

ngroman:
Add a configuration hook that allows a customized type -> string transform to be provided that's used during prop code-generation.

### Motivation
The runtime implementation of `--stripe-packages` mode loads classes in anonymous modules (one per package) instead of at the root-level in order to constant resolution within each package. The names returned by `Module#name` are not resolvable from the scope from which the generated code is `class_eval`d. 

The hook here allows us to manipulate the name into one that is resolvable. Default behavior of using `Module#name` implementation is unchanged.

### Test plan
Added automated test

### DIFF
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -363,7 +363,16 @@ module T::Configuration
     @scalar_types || @default_scalar_types
   end
 
+  # Guard against overrides of `name` or `to_s`
+  MODULE_NAME = Module.instance_method(:name)
+  private_constant :MODULE_NAME
+
+  @default_module_name_mangler = ->(type) {MODULE_NAME.bind(type).call}
   @module_name_mangler = nil
+
+  def self.module_name_mangler
+    @module_name_mangler || @default_module_name_mangler
+  end
 
   # Set to override the default behavior for converting types
   #   to names in generated code. Used by the runtime implementation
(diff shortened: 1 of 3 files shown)


ngroman:
cc @djudd-stripe

ngroman:
Open to 🚲 -shedding the name of the method.

elliottt (changes requested):
This all seems reasonable to me, and I'm fine with that name :)

Could you add some tests in `//gems/sorbet-runtime/test/types/configuration.rb`?

gems/sorbet-runtime/lib/types/configuration.rb
```diff
@@ -363,6 +363,22 @@ def self.scalar_types
     @scalar_types || @default_scalar_types
   end
 
+  @module_name_mangler = nil
```
paracycle, line 366: Why not make the default something like:
```suggestion
  @module_name_mangler = ->(type) { MODULE_NAME.bind(type).call }
```
so that you don't need the `if mangler` check in `serde_transform.rb` and can unconditionally do `T::Configuration.module_name_mangler.call(type)`?

ngroman: I considered that, but wanted to preserve the ability to go back to the default implementation with `module_name_mangler = nil`

paracycle: You can still do something like what `scalar_types` does:
```ruby
@default_module_name_mangler = ->(type) { MODULE_NAME.bind(type).call }
@module_name_mangler = nil

def self.module_name_mangler
  @module_name_mangler || @default_module_name_mangler
end

def self.module_name_mangler=(handler)
  @module_name_mangler = handler
end
```
which will allow you to get back to the default implementation with `module_name_mangler = nil`.

elliottt (approved):

PR #3808, opened 2020-12-17
Consider methods defined with `override` to also be `overridable`

### LINKED ISSUE #2746: Missing `overridable` only blows up during sorbet-runtime
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Boverridable.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aclass%20B%20%3C%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Boverride.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aclass%20C%20%3C%20B%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20%23%20NOTE%3A%20Here%20we're%20missing%20%60override%60%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20foo%0A%20%20end%0Aend%0A%0AC.new.foo%20%23%20this%20will%20blow%20up%20in%20prod)

```ruby
# typed: strict

class A
  extend T::Sig

  sig {overridable.void}
  def foo; end
end

class B < A
  extend T::Sig

  sig {override.void}
  def foo; end
end

class C < B
  extend T::Sig

  # NOTE: Here we're missing `override`
  sig {void}
  def foo
  end
end

C.new.foo # this will blow up in prod
```

#### Observed output

from `srb tc`
```
No errors! Great job.
```

from `sorbet-runtime`:

```
RuntimeError: You must use `.override` when overriding the existing method `foo`.
  Parent definition: B at /var/folders/gq/vz0kmn3d215bm2fztcwdk9040000gn/T/foo.rb:14
  Child definition:  C at /var/folders/gq/vz0kmn3d215bm2fztcwdk9040000gn/T/foo.rb:22
from /Users/me/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/sorbet-runtime-0.5.5410/lib/types/private/methods/signature_validation.rb:70:in `validate_override_mode'
```

#### Expected behavior

The `srb tc` should also show the same error msg as the runtime.

elliottt:
When overriding a method whose signature is marked as `override`, require that signature to also have `override`. The reasoning here is that if the parent signature is valid then it must either be overriding something that is `abstract` or `overridable`, or be overriding something that is marked `override` in which case we proceed recursively.

### Motivation

Incrementally work towards stricter override checking.

Partially addresses #2746.

### Test plan

See included automated tests.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -329,8 +329,9 @@ void validateOverriding(const core::Context ctx, core::SymbolRef method) {
         }
         auto isRBI = absl::c_any_of(method.data(ctx)->locs(), [&](auto &loc) { return loc.file().data(ctx).isRBI(); });
         if (!method.data(ctx)->isOverride() && method.data(ctx)->hasSig() &&
-            overridenMethod.data(ctx)->isOverridable() && !anyIsInterface && overridenMethod.data(ctx)->hasSig() &&
-            !method.data(ctx)->isRewriterSynthesized() && !isRBI) {
+            (overridenMethod.data(ctx)->isOverridable() || overridenMethod.data(ctx)->isOverride()) &&
+            !anyIsInterface && overridenMethod.data(ctx)->hasSig() && !method.data(ctx)->isRewriterSynthesized() &&
+            !isRBI) {
             if (auto e = ctx.state.beginError(method.data(ctx)->loc(), core::errors::Resolver::UndeclaredOverride)) {
                 e.setHeader("Method `{}` overrides an overridable method `{}` but is not declared with `{}`",
                             method.data(ctx)->show(ctx), overridenMethod.data(ctx)->show(ctx), "override.");
--- a/test/testdata/resolver/overrides.rb
+++ b/test/testdata/resolver/overrides.rb
@@ -56,6 +56,14 @@ class B5 < A5
   def foo; end
 end
 
+# however, not using override when overriding B5#foo should be an error
+class C5 < B5
+  extend T::Sig
+  sig {void}
+  def foo; end
+# ^^^^^^^ error: Method `C5#foo` overrides an overridable method
+end
+
 class A6
   extend T::Sig
   sig {void}


definition_validator/validator.cc
```diff
@@ -329,8 +329,9 @@ void validateOverriding(const core::Context ctx, core::SymbolRef method) {
         }
         auto isRBI = absl::c_any_of(method.data(ctx)->locs(), [&](auto &loc) { return loc.file().data(ctx).isRBI(); });
         if (!method.data(ctx)->isOverride() && method.data(ctx)->hasSig() &&
-            overridenMethod.data(ctx)->isOverridable() && !anyIsInterface && overridenMethod.data(ctx)->hasSig() &&
-            !method.data(ctx)->isRewriterSynthesized() && !isRBI) {
+            (overridenMethod.data(ctx)->isOverridable() || overridenMethod.data(ctx)->isOverride()) &&
```
elliottt, line 332: This conditional is getting pretty huge, but the meat of the change is that we raise this error if the parent is `overridable` or `override`, based on the assumption that the parent is correct.

test/testdata/resolver/overrides.rb
```diff
@@ -56,6 +56,14 @@ class B5 < A5
   def foo; end
 end
 
+# however, not using override when overriding B5#foo should be an error
+class C5 < B5
+  extend T::Sig
+  sig {void}
+  def foo; end
+# ^^^^^^^ error: Method `C5#foo` overrides an overridable method
+end
```
elliottt, line 65: This test case builds on previous classes present in this file: the parent class B5 has a valid method defined with `override`.

jez (approved):
lgtm, be sure to test on Stripe's codebase

elliottt:

→ https://go/builds/bui_IacTu5P7wI1NWO
→ https://go/builds/bui_IacTPNhVuVdnDm

elliottt:
Stripe migration pr: https://go/ps/pull/289810

PR #3810, opened 2020-12-17
Remove `Expression`

elliottt:
Remove the `Expression` type from `core/trees.h`. Given that there are no more virtual methods on `Expression`, this change is just a cleanup step.

### Motivation

This type is not longer used, and the use of `TreePtr::get` function is pretty minimal at this point.

### Test plan

See included automated tests.

### DIFF
--- a/ast/TreeCopying.cc
+++ b/ast/TreeCopying.cc
@@ -7,9 +7,9 @@ namespace sorbet::ast {
 
 namespace {
 
-TreePtr deepCopy(const Expression *avoid, const TreePtr &tree, bool root = false);
+TreePtr deepCopy(const void *avoid, const TreePtr &tree, bool root = false);
 
-template <class T> T deepCopyVec(const Expression *avoid, const T &origin) {
+template <class T> T deepCopyVec(const void *avoid, const T &origin) {
     T copy;
     copy.reserve(origin.size());
     for (const auto &memb : origin) {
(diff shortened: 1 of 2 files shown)


jez (approved):
end of an era

can we rename TreePtr to ExpressionPtr next?

ast/TreeCopying.cc
```diff
@@ -7,9 +7,9 @@ namespace sorbet::ast {
 
 namespace {
 
-TreePtr deepCopy(const Expression *avoid, const TreePtr &tree, bool root = false);
+TreePtr deepCopy(const void *avoid, const TreePtr &tree, bool root = false);
```
jvilk-stripe, line 10: Could this be a `TreePtr&`?

ast/Trees.h
```diff
@@ -178,9 +176,9 @@ class TreePtr {
         return static_cast<Tag>(value);
     }
 
-    Expression *get() const noexcept {
+    void *get() const noexcept {
```
jvilk-stripe, line 179: It'd be nice if this were private or otherwise restricted from use.

jvilk-stripe (approved):
Small non-blocking nits and questions.

ast/Trees.h
elliottt: Yes, it would be. There are parts of the codebase that still rely on being able to use pointer equality for things like map keys, so it'll be some work to make this private.

ast/TreeCopying.cc
elliottt: I'll investigate a bit, this is definitely one of the blockers to making `TreePtr::get` private.

elliottt: OK, I investigated this a bit and found that the thing preventing this is that we often call the `deepCopy` member function from the individual tree nodes, where we won't have a `TreePtr` available to pass through. I think this will also be a problem for making the `TreePtr::get` function private as well :(

PR #3812, opened 2020-12-18
Move the rename feature out of experimental status

soam-stripe:
Remove Renames from the set of experimental features; remove the config flag which enables it.

### Motivation

The rename feature has been tested in beta; we're ready to promote it to being always enabled.

### Test plan

Manually tested with VS code.


jvilk-stripe (approved):
There are some small tests to fixup, but otherwise this is good to go.

jez (approved):

PR #3815, opened 2020-12-18
Only look at top-level private/public calls

jez:
### Motivation

Only consider sends at (or near) the top level of a class def, but
**not** inside any other definition (like a method).

This makes it easier to implement `private` with no arguments.

### Test plan

Check the commit history for a before/after version of the test.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -467,6 +467,12 @@ public:
     }
 
     ast::TreePtr postTransformSend(core::Context ctx, ast::TreePtr tree) {
+        // Only consider sends at (or near) the top level of a class def, but **not** inside any
+        // other definition (like a method).
+        if (ownerStack.back().kind() != DefinitionKind::Class) {
+            return tree;
+        }
+
         auto &original = ast::cast_tree_nonnull<ast::Send>(tree);
 
         switch (original.fun.rawId()) {
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -467,12 +467,6 @@ public:
     }
 
     ast::TreePtr postTransformSend(core::Context ctx, ast::TreePtr tree) {
-        // Only consider sends at (or near) the top level of a class def, but **not** inside any
-        // other definition (like a method).
-        if (ownerStack.back().kind() != DefinitionKind::Class) {
-            return tree;
-        }
-
         auto &original = ast::cast_tree_nonnull<ast::Send>(tree);
 
         switch (original.fun.rawId()) {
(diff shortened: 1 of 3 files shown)


test/testdata/rewriter/flatten_module_private_class_method.rb.symbol-table-raw.exp
```diff
@@ -22,6 +20,4 @@ class <C <U <root>>> < <C <U Object>> ()
       argument <blk><block> @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=??? end=???}
     method <S <C <U B>> $1><U thing> (<blk>) @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=11:3 end=11:17}
       argument <blk><block> @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=??? end=???}
-  class <S <S <C <U B>> $1> $1>[<C <U <AttachedClass>>>] < <S <S <C <U Object>> $1> $1> () @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=10:7 end=10:8}
-    type-member(+) <S <S <C <U B>> $1> $1><C <U <AttachedClass>>> -> LambdaParam(<S <S <C <U B>> $1> $1><C <U <AttachedClass>>>, lower=T.noreturn, upper=AppliedType {   klass = <S <C <U B>> $1>   targs = [     <C <U <AttachedClass>>> = B   ] }) @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=10:7 end=10:8}
```
jez, line 26: Check the commit messages for a lengthy description of this file's diff.

namer/namer.cc
```diff
@@ -467,6 +467,12 @@ class SymbolFinder {
     }
 
     ast::TreePtr postTransformSend(core::Context ctx, ast::TreePtr tree) {
+        // Only consider sends at (or near) the top level of a class def, but **not** inside any
+        // other definition (like a method).
+        if (ownerStack.back().kind() != DefinitionKind::Class) {
```
jvilk-stripe, line 472: There's actually a way to do this that makes namer faster: Make this pass a `ShallowMap`!

jvilk-stripe: (We might be able to do the same with `TreeSymbolizer`? Not sure.)

jvilk-stripe: (Unless we otherwise care about definitions within method definitions? I don't know their semantics in Ruby.)

elliottt: +1 to this, I like the idea of skipping method bodies unless we actually have to traverse them.

jez:

→ https://go/builds/bui_Ib2h2OVmMUrDlD
→ https://go/builds/bui_Ib2hgsemCisRHE

jez:
ptal @elliottt 

Stripe codebase passes, test suite passes, and I can't imagine this makes sorbet slower (but i'm too lazy to measure)

elliottt:
> ptal @elliottt
> 
> Stripe codebase passes, test suite passes, and I can't imagine this makes sorbet slower (but i'm too lazy to measure)

If anything, I think this would improve performance as we'd be avoiding method body traversals.

elliottt (approved):

PR #3818, opened 2020-12-18
Add ability to exclude values in runtime TypeError

bss:
This adds a new configuration option `T::Configuration.exclude_value_in_type_errors` that when true will exclude values from messages in `TypeError`s.

### Motivation
It is important for Gusto that we don't leak any user data to our error tracker. We are currently relying on filtering out the data of the error message before we send it to our error tracker. Any problems with that filtering and we might be sending user data by mistake. This instead will make sure the user data isn't included in the exception in the first place.

### Test plan
See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -42,6 +42,34 @@ module T::Configuration
     T::Private::Methods.set_final_checks_on_hooks(false)
   end
 
+  @include_value_in_type_errors = true
+  # Whether to include values in TypeError messages.
+  #
+  # Including values is useful for debugging, but can potentially leak
+  # sensitive information to logs.
+  #
+  # @return [T::Boolean]
+  def self.include_value_in_type_errors?
+    @include_value_in_type_errors
+  end
+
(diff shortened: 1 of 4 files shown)


jez:
@bss I reworked your PR to make it impossible for someone to misuse the API.

Imagine a future where instead of blanket including/excluding values, there's something more nuanced (maybe: only in some sort of test mode, or only some sort of values, etc). In that case, we'd want to change the API to reflect the user's choice.

But if people were somehow misusing this API, we might accidentally break them (maybe they were using `nil` and `false` interchangeably, when we wanted to differentiate them. Or maybe maybe instead of just `true`/`false`, they're using `:true` or `'true'` or `1` accidentally (because they pulled it from an environment variable and forgot to coerce to a boolean).

In this new API, the underlying data is guaranteed to be a boolean. This mimics what we do for other `T::Configuration` APIs.

jez:
Also: you had forgotten to update `t.rbi`, so you would have had type errors trying to use that new API in your application. I went ahead and fixed that.

jez (approved):
Thanks!

bss:
Sweet, thanks for the changes and the review.

PR #3819, opened 2020-12-18
Support nullary private method (toggle visibility)

jez:
### Motivation

This is another way to mark methods private that we should support (some
codebases use this style heavily).

Supersedes #3758.

### Test plan

See included automated tests.


jez:
Reposting #3817 because I accidentally merged it into a feature branch (stacked PR).

elliottt (approved):

PR #3822, opened 2020-12-21
wip: More accurate shape and tuple literals

### LINKED ISSUE #11: Static checker does not verify hash types in many cases
```
> git rev-parse HEAD
f96dc929d9cd4e10048ec4e907e659c50b7a33b7
> cat repro1.rb
# typed: true
a = T.let({symbol_key: 'integer_value'.size}, T::Hash[Integer, Symbol])
> bazel-bin/main/sorbet repro1.rb
No errors! Great job.
```
It is also not checked for method parameters:

```rb
# typed: true
extend T::Sig
  
sig {params(x: T::Hash[Integer, String]).returns(String)}
def foo(x)
  x.to_s
end

foo(symbol_key: 'integer value'.size) # An error here would be nice
```

Playground: https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%20%20%0Asig%20%7Bparams(x%3A%20T%3A%3AHash%5BInteger%2C%20String%5D).returns(String)%7D%0Adef%20foo(x)%0A%20%20x.to_s%0Aend%0A%0Afoo(symbol_key%3A%20'integer%20value'.size)%20%23%20An%20error%20here%20would%20be%20nice%0A

### LINKED ISSUE #713: Shapes are compatible incorrectly with Hashes
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aa%20%3D%20%7B%201%20%3D%3E%20'b'%20%7D%0Ab%20%3D%20T.let(a%2C%20T%3A%3AHash%5BString%2C%20Integer%5D)%0AT.reveal_type(a)%0AT.reveal_type(b))

```ruby
# typed: strict

a = { 1 => 'b' }
b = T.let(a, T::Hash[String, Integer])
T.reveal_type(a)
T.reveal_type(b)
```

#### Observed output

```
editor.rb:5: Revealed type: {Integer(1) => String("b")} https://sorbet.org/docs/error-reference#7014
     5 |T.reveal_type(a)
        ^^^^^^^^^^^^^^^^
  From:
    editor.rb:3:
     3 |a = { 1 => 'b' }
            ^^^^^^^^^^^^

editor.rb:6: Revealed type: T::Hash[String, Integer] https://sorbet.org/docs/error-reference#7014
     6 |T.reveal_type(b)
        ^^^^^^^^^^^^^^^^
  From:
    editor.rb:4:
     4 |b = T.let(a, T::Hash[String, Integer])
Errors: 2
```

#### Expected behavior

An error about not being allowed to `T.let` that

- - -

jez:
### Motivation

Closes #11 
Closes #713

### Commit summary

The core change is basically: instead of computing `underlying` when the type
is built, compute it when it's needed. By that point, we're guaranteed to have
a `GlobalState` on hand, which means we have no problem using `Types::any` to
lub all the elements together.

Before, there were some places where we didn't have a `GlobalState` populated,
so we could not reliably lub all the elements together at shape/tuple
allocation time.

- **Delete underlying_, update all constructor calls** (2fdb44298)

- **Thread GlobalState through call sites** (8fed7668a)

- **Thread GlobalState through GENERATE_CALL_MEMBER** (f3c4e097c)

- **Types::lubAll -> lubAllDropLiteral** (3aeb4ebf9)

  `Types::lubAll` was only used in one place: `TupleType::build`, which I
  removed.

  I have repurposed it to help implement the assorted `underlying` methods
  below, and made it private. Also, `TupleType::build` used to call
  `dropLiteral` on its own. I have also moved that call into the helper.

- **Fix some use-after-free errors** (edde0ec9d)

  Since `underlying` is computed on the fly now, if it's not stored by the
  caller, it's dropped immediately. So if you try to get the underlying
  and immediately cast it to something, it will destruct before you have a
  chance to make use of the downcast value.

  This is super annoying, and I'm glad it came up in our test case, but I
  would love to figure out how to make it a compile error.

- **ShapeType::showWithMoreInfo** (edb29fcc3)

  I noticed that we had a `showWithMoreInfo` for `TupleType`, but we
  didn't have one for `ShapeType`. That felt inconsistent, and also helped
  me make sense of some of the tests.

- **Empty arrays and hashes are containers of untyped** (2896cc75a)

  Largely, this is to preserve the existing behavior.

  - Magic_buildArray already made it so that empty array literals behaved
   like `T::Array[T.untyped]` (by short circuiting in the intrinsic).

  - Magic_buildHash also did, but by passing `hashOfUntyped` when
   constructing the underlying, instead of short circuiting like for
   arrays.

  Now, this untypedness is handled consistently for both in `underlying`.
  This strikes a balance--for places that actually try to look at whether
  you have a shape type or not, it's still visibly an empty shape type
  (e.g., some keyword arg stuff was depending on this). But if you ever
  try to call a method on it that isn't implemented as a
  ShapeType/TupleType intrinsic (and thus falls back up to Hash or Array),
  then it will decay as an untyped container.

- **Add and fix tests** (ec828d227)

  I believe that every change is expected, and is not going to cause undue
  breakage (i.e., in every case, to get back the original behavior, people
  should be able to `T.unsafe(...)` their hash literal, or more accurately
  `T.let(..., T::Hash[T.untyped, T.untyped])`).

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_IbpW48VvJ1Hrph
→ https://go/builds/bui_IbpWn3pNNHIlvP

jez:

→ https://go/builds/bui_IbppXUCtvo0K6u
→ https://go/builds/bui_IbppkPV0PYI7A4

jez:
This causes ~5,000 errors on Stripe's code base.

A skim of some of the major classes of errors:

#### 1

```ruby
x = {foo: ''}
x[:foo].size
```

We should be able to write an intrinsic that says `x[:foo]` is non-nil.

#### 2

```ruby
x = {foo: 0}
x[:bar] = ''
```

maybe we should write an intrinsic for `[]=` that silently allows this? it's not much worse than it is currently but would allow for other things to be typed better

#### 3

```ruby
sig {params(x: {foo: T.nilable(Integer)}).void}
def foo(x)
  if x[:foo]
    x[:foo].pred
  end
end
```

... this is tricky. maybe we should detect type errors on hashes and have an autocorrect to pin the origin of the shape type to `T::Hash[T.untyped, T.untyped]`?

#### 4

... even then there are a lot that don't seem to fall into a particular category.

What we might want to do is:

- give people on Slack a heads up
- ask them to give feedback about whether they want this or not
- merge support for this in behind a flag
  - tbd: should this behavior be opt in or opt out

jez:

→ https://go/builds/bui_IbrdPMzRMkn81V
→ https://go/builds/bui_IbrdgL391r4lqe

jez:
It's down to only ~2,000 errors on Stripe's codebase with the changes in #3823.

jez:
Probably the best way to land this:

- remove `underlying_`, but continue to compute `underlying(gs)` as if it were `T::Hash[T.untyped, T.untyped]`
  - we can land that with no behavior changes, so the large diff doesn't have to bitrot
  - we'll then have a convenient single place to tweak how aggressively to tighten up shapes and tuple types
- land #3823 which adds some `[]` and `[]=` intrinsics for hashes
  - this will cause some errors, but hopefully a manageable amount

then we'd be back to landing the core of this PR's changes

ideas:

- if there's ever a shape type in an error message, make an autocorrect to `T.let(..., T::Hash[T.untyped, T.untyped]` around the definition
- land it behind a command line flag, so that people can turn it on and chip away at the errors it in their free time
  - depending on community reception, might want to flip it to be opt-out at some point
- make some way to change the behavior of `buildHash` per file? this seems bad because it means sorbet behaves way differently in different places, but it would make migrating portions of a large codebase easier

PR #3823, opened 2020-12-21
wip: Implement intrinsics for [] and []= on shape types

jez:
SEO: square brackets index square_brackets squareBrackets squareBracketsEq eq

### Motivation

This should help ease the pain of #3822.

Right now I've implemented it stacked on top, but we might want to land them in
the opposite order, or at the same time.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2304,6 +2304,71 @@ public:
     }
 } Tuple_concat;
 
+namespace {
+
+optional<size_t> indexForKey(const ShapeType &shape, const LiteralType &argLit) {
+    auto fnd = absl::c_find_if(
+        shape.keys, [&argLit](auto &elemLit) { return argLit.equals(cast_type_nonnull<LiteralType>(elemLit)); });
+
+    if (fnd == shape.keys.end()) {
+        return nullopt;
+    } else {
+        return fnd - shape.keys.begin();
+    }
+}
+
+} // namespace
+
+// TODO(jez) Add tests for this
+class Shape_squareBrackets : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        auto &shape = cast_type_nonnull<ShapeType>(args.thisType);
+
+        if (args.args.size() != 1) {
+            // Skip over cases for which arg matching should report errors
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_Ibs0EnWGUF9tJX
→ https://go/builds/bui_Ibs0MPuuS6D5JC

jez:

→ https://go/builds/bui_Ibsp7eRrmQwBtQ
→ https://go/builds/bui_IbsqSkaRpZbdKt

jez:

→ https://go/builds/bui_IbsuzN5y7tIB67
→ https://go/builds/bui_IbsulFkmdiWaFd

jez:

→ https://go/builds/bui_Ibt0FddeLcnAn2
→ https://go/builds/bui_Ibt0qMn99xieXX

karmakaze:
This would be amazing to land on its own, as I've begun working on a specific subset in #4497 that could be very useful in combination.

jez:

karmakaze:
@jez Thanks for taking the time to explain the situation. I completely understand and agree with the outcome.

I became interested in contributing because I was bumping into some issues, so scratching the itch.

Unrelated to this issue, but since it was suggested I look for another area... One sore spot is re: https://srb.help/7019 . I ran into this as using `T.unsafe` is very unsafe for common Hash that use splats. Could there be anything in that area where I could contribute, perhaps some .rbi's for methods or anything?

_For sure, I'll search the issues and read any existing information on this topic as well._ Thanks again!

jez:
The problem with splats is that if `opts` is typed loosely as `T::Hash[Symbol, T.untyped]`, we don't know whether all the required keyword args are included in `opts` when doing something like

```ruby
foo(**opts)
```

We have two options:

- Be explicit about this and require people to opt-out of type checking by untyping the code
- Being silent, and letting the code pass (possibly erroneously)

Ultimately, the only thing that would truly solve this in any ergonomic sense would be to improve sorbet's support for shape types.

karmakaze:
Oh sorry I wasn't clear. I was bumping more into the single splats rather than double splats:
```
foo(*opts)
```

jez:
ah got it. I agree that is an example that should be able to type check but I don't have any good ideas for how to accomplish that off the top of my head

PR #3825, opened 2020-12-21
disallow cast_type and friends on rvalue references

froydnj:
Don't permit `cast_type` and various similar functions on rvalue references, because such operations may not be safe.  If `cast_type` and `cast_type_nonnull` were written as member functions, we'd do this with reference qualifiers on the relevant member functions, which might be slightly clearer?

### Motivation

@jez had to fix a couple problems related to this in #3822, but enforcing this at compile time a) catches more cases and b) ensures we don't have to do debugging like this again.

I left `isa` as still working on rvalue references (overload resolution on rvalue references selects the `const &` overload) since there's no safety problems, but I can be convinced to apply similar changes there for symmetry's sake.

I didn't do the same thing for `TreePtr`, but can do the same thing as a followup.

### Test plan

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -277,11 +277,17 @@ template <class To> To *cast_type(TypePtr &what) {
     return const_cast<To *>(cast_type<To>(static_cast<const TypePtr &>(what)));
 }
 
+template <class To> To *cast_type(TypePtr &&what) = delete;
+
 template <class To>
 typename TypePtr::TypeToCastType<To, TypePtr::TypeToIsInlined<To>::value>::type cast_type_nonnull(const TypePtr &what) {
     ENFORCE_NO_TIMER(isa_type<To>(what));
     return *reinterpret_cast<const To *>(what.get());
 }
+
+template <class To>
+typename TypePtr::TypeToCastType<To, TypePtr::TypeToIsInlined<To>::value>::type
+cast_type_nonnull(TypePtr &&what) = delete;
 
 // Simple forwarders defined on TypePtr which make `typecase` work.
 template <class To> inline bool TypePtr::isa(const TypePtr &what) {
--- a/core/types/types.cc
+++ b/core/types/types.cc
@@ -412,7 +412,8 @@ OrType::OrType(const TypePtr &left, const TypePtr &right) : left(move(left)), ri
 
 void TupleType::_sanityCheck(const GlobalState &gs) const {
     sanityCheckProxyType(gs, underlying());
-    auto *applied = cast_type<AppliedType>(this->underlying());
+    auto underlying = this->underlying();
+    auto *applied = cast_type<AppliedType>(underlying);
     ENFORCE(applied);
     ENFORCE(applied->klass == Symbols::Array());
 }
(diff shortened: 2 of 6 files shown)


core/Types.h
```diff
@@ -277,12 +277,18 @@ template <class To> To *cast_type(TypePtr &what) {
...
 template <class To>
 typename TypePtr::TypeToCastType<To, TypePtr::TypeToIsInlined<To>::value>::type cast_type_nonnull(const TypePtr &what) {
     ENFORCE_NO_TIMER(isa_type<To>(what));
     return *reinterpret_cast<const To *>(what.get());
 }
 
+template <class To>
+typename TypePtr::TypeToCastType<To, TypePtr::TypeToIsInlined<To>::value>::type
+cast_type_nonnull(TypePtr &&what) = delete;
+
```
jez, line 291: Can we add a comment here saying this is deleted because it's too easy to accidentally have the TypePtr destruct before you have a chance to do something interesting with the casted result?

core/types/types.cc
```diff
@@ -412,7 +412,8 @@ OrType::OrType(const TypePtr &left, const TypePtr &right) : left(move(left)), ri
 
 void TupleType::_sanityCheck(const GlobalState &gs) const {
     sanityCheckProxyType(gs, underlying());
-    auto *applied = cast_type<AppliedType>(this->underlying());
+    auto underlying = this->underlying();
+    auto *applied = cast_type<AppliedType>(underlying);
```
jez, line 416: These two usages were exactly the problem (in TupleType). To confirm: the change you made caused these to become compile errors as written? You didn't hunt this call site down on your own?

jez (approved):

jez:
> I left `isa` as still working on rvalue references (overload resolution on rvalue references selects the `const &` overload) since there's no safety problems, but I can be convinced to apply similar changes there for symmetry's sake.

Leaving `isa` as written wfm.

jez:
> I didn't do the same thing for `TreePtr`, but can do the same thing as a followup.

Works for me!

core/types/types.cc
froydnj: Yes, I made the relevant `= delete` changes and then all the other changes were required to fix the build.

froydnj:
> > with reference qualifiers
> 
> What are these?

You can restrict whether class methods can be called on rvalue references, lvalue references, or both (default):

```c++
struct Foo {
  void only_on_rvalues() &&;
  void only_on_lvalues() &;
};
```

The canonical (in my mind) use-case is methods returning raw pointers to class-owned objects; you want to lvalue-qualify them so those methods can't be called on temporary values, for exactly the same reasons as this PR:

```c++
struct Foo {
  Foo *some_owned_ptr();
  Foo *some_owned_ptr_safe() &;
};

Foo compute();
// Unsafe because the temporary from compute is destroyed at
// the end of this statement.
Foo *unsafe = compute().some_owned_ptr();
Foo obj = compute();
// Possibly unsafe due to raw pointers, but better, because `obj` is guaranteed to still
// be live here.
Foo *safe = obj.some_owned_ptr_safe();
```

elliottt:
We should definitely apply this to the tagged pointer cast functions that we have in the codebase 👍

elliottt (approved):

PR #3826, opened 2020-12-21
add ./ to bazel invocations to ensure we use our bazel

froydnj:
I noticed these invocations broke on a local machine that didn't have bazel installed.

### Motivation

More robust utility scripts.

### Test plan


jez (approved):
Thanks!

PR #3827, opened 2020-12-21
wip: Proof of Concept super strict shape types

jez:
### Motivation

This implementation of shape types is overly pedantic, and causes all sorts of new problems.

We should not land it as is (and in fact, it's based on top of two other PRs that we also can't land as they are).

But anyways, it was a fun weekend project to see what's possible and how much breaks.

### Test plan

See included automated tests.


jez:

Issue #3792, opened 2020-12-14
T::Struct causes false errors during `srb init`


AaronC81:
This was spotted by @manhhung741 and @jsmestad trying to use Parlour with recent versions of Sorbet: [related Parlour issue](https://github.com/AaronC81/parlour/issues/110)

To reproduce:

1. Create a Gemfile containing Parlour 4.0.1 (the latest non-prerelease version) and the latest Sorbet
2. `bundle install`
3. `bundle exec srb init` and `y`

I get:

```
Generating: sorbet/config
Generating: sorbet/rbi/sorbet-typed/
Generating: sorbet/rbi/gems/
Generating: sorbet/rbi/hidden-definitions/
Requiring all of your code
Naming all Modules
Naming Bundler::ThorGot LoadError when trying to get nested name Bundler::Thor::RakeCompat
Naming Struct/home/aaron/src/sorbet/gems/sorbet/lib/real_stdlib.rb:77: warning: constant Struct::Tms is deprecated
Generating /tmp/d20201214-74918-bjoxsq/reflection.rbi with 1077 modules and 43 aliases
Printing your code's symbol table into /tmp/d20201214-74918-bjoxsq/from-source.json
Printing /tmp/d20201214-74918-bjoxsq/reflection.rbi's symbol table into /tmp/d20201214-74918-bjoxsq/reflection.json
Traceback (most recent call last):
        6: from /home/aaron/src/sorbet/gems/sorbet/bin/srb-rbi:237:in `<main>'
        5: from /home/aaron/src/sorbet/gems/sorbet/bin/srb-rbi:196:in `main'
        4: from /home/aaron/src/sorbet/gems/sorbet/bin/srb-rbi:121:in `init'
        3: from /home/aaron/src/sorbet/gems/sorbet/bin/srb-rbi:232:in `block in make_step'
        2: from /home/aaron/src/sorbet/gems/sorbet/lib/hidden-definition-finder.rb:38:in `main'
        1: from /home/aaron/src/sorbet/gems/sorbet/lib/hidden-definition-finder.rb:47:in `main'
/home/aaron/src/sorbet/gems/sorbet/lib/hidden-definition-finder.rb:151:in `write_constants': /tmp/d20201214-74918-bjoxsq/reflection.rbi had unexpected errors. Check this file for a clue: /tmp/d20201214-74918-bjoxsq/reflection.json.err (RuntimeError)
```

...where that .err file contains:

```
[...]

PR #3828, opened 2020-12-21
srb init: Ignore gem RBIs when generating the symbol table of reflection.rbi

jvilk-stripe:
srb init: Ignore gem RBIs when generating the symbol table of reflection.rbi

Unfortunately, I am unsure how to test this. We would need the ability to have gem tests that pull in gemfile dependencies. Is that possible?

### Motivation

Fixes https://github.com/sorbet/sorbet/issues/3792

### Test plan

Not covered by any tests. Tested manually using repro case from https://github.com/sorbet/sorbet/issues/3792

### DIFF
--- a/gems/sorbet/lib/hidden-definition-finder.rb
+++ b/gems/sorbet/lib/hidden-definition-finder.rb
@@ -121,6 +121,7 @@ class Sorbet::Private::HiddenMethodFinder
     puts "Printing #{TMP_RBI}'s symbol table into #{RBI_CONSTANTS}"
     io = IO.popen(
       [
+        {'SRB_SKIP_GEM_RBIS' => 'true'},
         File.realpath("#{__dir__}/../bin/srb"),
         'tc',
         # Make sure we don't load a sorbet/config in your cwd


jvilk-stripe:
As for why the behavior changed: I'm investigating that currently.

jez:
> Unfortunately, I am unsure how to test this. We would need the ability to have gem tests that pull in gemfile dependencies. Is that possible?

It's probably testable, but I don't really want to think about how.

srb init tests are slow and brittle.

jez (approved):

jez:
if you feel so compelled to try writing a test, there are some instructions in

https://github.com/sorbet/sorbet/blob/master/gems/sorbet/README.md#writing-tests

jvilk-stripe:
I'm going to wait to merge this until I understand why those errors only just started appearing.

jvilk-stripe:
In #3792 I've figured out why the errors are appearing, but not why they are appearing now. I confirmed that, with this patch, `hidden.rbi` is basically identical to what is generated in the failing example before the error started occurring. Thus, I believe it is safe to merge.

PR #3829, opened 2020-12-22
Also forbid casting of TreePtr temporaries

elliottt:
Follow up with the same refactoring to #3825.

### Motivation

Avoiding bugs!

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -268,6 +268,11 @@ template <class To> const To *cast_tree(const TreePtr &what) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


jez (approved):
Nice!

i kind of figured we’d see some call sites to update but maybe not?

elliottt:
> Nice!
> 
> i kind of figured we’d see some call sites to update but maybe not?

There weren't any to update when I built locally -- I think we're a bit more careful when making expressions because they started as `unique_ptr<Expression>`. It'll be nice to get a compiler error if we slip up though :)

PR #3805, opened 2020-12-17
WIP Resolve constants: Create per-file vectors to make serial sorting faster

jvilk-stripe:
WIP Resolve constants: Create per-file vectors to make serial sorting faster.

Attempts to reduce a serial bottleneck on large codebases (appending and sorting very large arrays).

Since the previous vectors were sorted by file first followed by loc, this does not change behavior/sort order 🎉  As a bonus, we can sort the per-file vectors in parallel, and stash fref on the struct holding the vectors.

### Motivation

I noticed that vector insertion/sorting takes up a noticeable amount of time on Stripe's codebase. This type of solution worked well for the other resolver passes which previously had the same problem.

### Test plan

Covered by existing tests.

PR #3830, opened 2020-12-22
WIP: Use InlinedVector rather than shared_ptr for Nesting in ResolveConstantsWalk

jvilk-stripe:
https://github.com/sorbet/sorbet/pull/3805 speeds up ResolveConstantsWalk, but I noticed that it spends a lot of time destructing. I can't tell what (since it's all inlined in production builds), but my theory is that shared_ptr destruction is expensive.

I think an InlinedVector will be a lot cheaper. SymbolRefs are 4 bytes each, so an InlinedVector of 4 has 16 bytes of storage (the same size as a shared_ptr!) plus normal vector overheads. It inflates the size of the `ResolutionItem` struct by 8 bytes.

WIP while I benchmark and determine if a size of 4 is sufficient (I think it is).

### Motivation

Speed!

### Test plan

Covered by existing tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -96,7 +96,7 @@ class ResolveConstantsWalk {
 
 private:
     // TODO: Figure out ideal size.
-    typedef InlinedVector<core::ClassOrModuleRef, 4> Nesting;
+    typedef InlinedVector<core::ClassOrModuleRef, 6> Nesting;
     Nesting nesting_;
 
     struct ResolutionItem {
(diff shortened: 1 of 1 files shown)


jvilk-stripe:

→ https://go/builds/bui_IcDfbDpgNvWmnE
→ https://go/builds/bui_IcDfmIGCkl2FhY

jvilk-stripe:

→ https://go/builds/bui_IcEMGn5ReP4pnO
→ https://go/builds/bui_IcEMMeu5xmCPKB

jvilk-stripe:
This did not have the expected impact. No measurable slowdown or memory savings/increase.

PR #3831, opened 2020-12-22
De-duplicate rename edits

soam-stripe:
This change adds checks to deduplicate the edits created by method renames. Duplicate edits are created when we reach the same method send through multiple definitions. VS Code rejects duplicate edits with an unhelpful "failed to apply edits" error (and they also cause sorbet.run to silently ignore the edits).

### Motivation

Method renames for union types were broken. Earlier tests failed to catch this because the length of the old and new names were the same. 🤦  

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/rename.cc
+++ b/main/lsp/requests/rename.cc
@@ -228,6 +228,19 @@ public:
         if (location == nullptr) {
             return;
         }
+
+        // Have we already edited this exact location?
+        // If we're renaming the exact same place twice, silently ignore it. We reach this condition when we find the
+        // same method send through multiple definitions (e.g. in the case of union types)
+        auto it = edits.find(location->uri);
+        if (it != edits.end()) {
+            for (auto &textedit : it->second) {
+                if (location->range->cmp(*textedit->range) == 0) {
+                    return;
+                }
+            }
+        }
+
         string newsrc;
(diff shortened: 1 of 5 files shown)


main/lsp/requests/rename.cc
```diff
@@ -228,6 +228,19 @@ class MethodRenamer : public Renamer {
         if (location == nullptr) {
             return;
         }
+
+        // Have we already edited this exact location?
```
jvilk-stripe, line 232: Rather than eagerly dedupe, should we dedupe at the end? I think that'd be cheaper. Regardless, I'm OK with approving this as-is so we can get this in.

jvilk-stripe (approved):

main/lsp/requests/rename.cc
soam-stripe: Hmm you're right, n^2 vs lg n in the number of edits in one file. Or maybe we should load it into a set and turn it into a vector at the end. I think I'll do this is in a separate change so we can at least unblock this use case for now.

jvilk-stripe: Sorbet's `loc` data structure is much smaller to store and cheaper to compare than the JSON data structures (3 numbers rather than string + numbers), so we could consider lazily converting edits into the desired JSON-style data structure.

PR #3833, opened 2020-12-22
Remove ShapeType::underlying_ field

jez:
### Motivation

This breaks another change out of #3822.

It should have zero behavior change, so it will let us more easily experiment
with small tweaks to how shape types are checked, so that we can land
incremental updates.

If anything, this should just be a performance win, as it drops the size of
ShapeType's by 25% (64 bytes -> 48 bytes).

### Test plan

No behavior change.

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -656,7 +656,6 @@ public:
     std::vector<TypePtr> keys; // TODO: store sorted by whatever
     std::vector<TypePtr> values;
     const TypePtr underlying_;
-    ShapeType();
     ShapeType(TypePtr underlying, std::vector<TypePtr> keys, std::vector<TypePtr> values);
 
     std::string toStringWithTabs(const GlobalState &gs, int tabs = 0) const;
--- a/core/Types.h
+++ b/core/Types.h
@@ -655,8 +655,7 @@ TYPE(ShapeType) final {
 public:
     std::vector<TypePtr> keys; // TODO: store sorted by whatever
     std::vector<TypePtr> values;
-    const TypePtr underlying_;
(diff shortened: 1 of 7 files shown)


core/types/types.cc
```diff
@@ -418,18 +418,13 @@ void TupleType::_sanityCheck(const GlobalState &gs) const {
...
-ShapeType::ShapeType(TypePtr underlying, vector<TypePtr> keys, vector<TypePtr> values)
-    : keys(move(keys)), values(move(values)), underlying_(std::move(underlying)) {
+ShapeType::ShapeType(vector<TypePtr> keys, vector<TypePtr> values) : keys(move(keys)), values(move(values)) {
     DEBUG_ONLY(for (auto &k : this->keys) { ENFORCE(isa_type<LiteralType>(k)); };);
     categoryCounterInc("types.allocated", "shapetype");
 }
 
 TypePtr ShapeType::underlying(const GlobalState &gs) const {
-    return this->underlying_;
+    return Types::hashOfUntyped();
```
jez, line 427: This is the key part of the change.

core/Types.h
```diff
@@ -670,7 +668,7 @@ TYPE(ShapeType) final {
     TypePtr underlying(const GlobalState &gs) const;
     bool derivesFrom(const GlobalState &gs, core::ClassOrModuleRef klass) const;
 };
-CheckSize(ShapeType, 64, 8);
+CheckSize(ShapeType, 48, 8);
```
elliottt, line 671: 🎉

elliottt (approved):
Nice!

jez:

→ https://go/builds/bui_IcFVWa8lacBXd2
→ https://go/builds/bui_IcFV1FD08hDiuD

PR #3834, opened 2020-12-22
Remove TupleType::underlying_ field

jez:
### Motivation

Basically the same change as #3833, but for TupleType instead of ShapeType.

Same motivation:

- prework broken out of #3822
- no behavior changes
- drops TupleType size

Has one unique downside:

- we're doing lubAll inside underlying, so we could potentially be doing more
  lubbing work over and over again. if this ends up being too slow, we can
  always memoize the result of this function.

### Test plan

See included automated tests.

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -676,10 +676,8 @@ private:
 
 public:
     std::vector<TypePtr> elems;
-    const TypePtr underlying_;
 
-    TupleType(TypePtr underlying, std::vector<TypePtr> elements);
-    static TypePtr build(const GlobalState &gs, std::vector<TypePtr> elements);
+    TupleType(std::vector<TypePtr> elements);
 
     std::string toStringWithTabs(const GlobalState &gs, int tabs = 0) const;
     std::string show(const GlobalState &gs) const;
@@ -696,7 +694,7 @@ public:
     TypePtr underlying(const GlobalState &gs) const;
     bool derivesFrom(const GlobalState &gs, core::ClassOrModuleRef klass) const;
 };
-CheckSize(TupleType, 40, 8);
+CheckSize(TupleType, 24, 8);
 
 TYPE(AppliedType) final {
 public:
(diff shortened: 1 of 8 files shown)


jez:

→ https://go/builds/bui_IcGsM9m348CPtd
→ https://go/builds/bui_IcGsWfGRZO95RB

core/Types.h
```diff
@@ -696,7 +694,7 @@ TYPE(TupleType) final {
     TypePtr underlying(const GlobalState &gs) const;
     bool derivesFrom(const GlobalState &gs, core::ClassOrModuleRef klass) const;
 };
-CheckSize(TupleType, 40, 8);
+CheckSize(TupleType, 24, 8);
```
elliottt, line 697: 👍

resolver/resolver.cc
```diff
@@ -450,7 +450,7 @@ class ResolveConstantsWalk {
         auto uaSym = ctx.state.enterMethodSymbol(core::Loc::none(), item.klass, core::Names::unresolvedAncestors());
         core::TypePtr resultType = uaSym.data(ctx)->resultType;
         if (!resultType) {
-            uaSym.data(ctx)->resultType = core::TupleType::build(ctx, {ancestorType});
+            uaSym.data(ctx)->resultType = core::make_type<core::TupleType>(vector<core::TypePtr>{ancestorType});
```
elliottt, line 453: Do you need to name the `vector<core::TypePtr>` part of the initializer, or does it work without it?

elliottt (approved):
Nice!

resolver/resolver.cc
jez: yeah, i think it has to do with explicit constructors. i got a compile error when i omitted this. i was going to play around with it.

jez: no, adding `explicit` to the TupleType constructor doesn't do it. it must be because `TupleType::build` is a normal method, but `make_type` is templated and forwards args?

PR #3836, opened 2020-12-22
Empty array literal is zero-tuple

### LINKED ISSUE #3751: Sorbet thinks empty array literal (`[]`) is T::Array[T.untyped] instead of a zero-tuple
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20%0A%20%20sig%20%7B%20params(x%3A%20T%3A%3AArray%5BString%5D).void%20%7D%0A%20%20def%20foo1(x)%0A%20%20%20%20T.reveal_type(x)%20%23%20T%3A%3AArray%5BString%5D%20(as%20expected)%0A%20%20end%0A%0A%20%20sig%20%7B%20params(x%3A%20T%3A%3AArray%5BString%5D).void%20%7D%0A%20%20def%20foo2(x%20%3D%20%5B%5D)%0A%20%20%20%20T.reveal_type(x)%20%23%20T%3A%3AArray%5BT.untyped%5D%20(expected%20T%3A%3AArray%5BString%5D)%0A%20%20end%0Aend)

```ruby
# typed: true
class A
  extend T::Sig
  
  sig { params(x: T::Array[String]).void }
  def foo1(x)
    T.reveal_type(x) # T::Array[String] (as expected)
  end

  sig { params(x: T::Array[String]).void }
  def foo2(x = [])
    T.reveal_type(x) # T::Array[T.untyped] (expected T::Array[String])
  end
end
```

#### Observed output

```
editor.rb:7: Revealed type: T::Array[String] https://srb.help/7014
     7 |    T.reveal_type(x) # T::Array[String] (as expected)
            ^^^^^^^^^^^^^^^^
  From:
    editor.rb:6:
     6 |  def foo1(x)
                   ^

editor.rb:12: Revealed type: T::Array[T.untyped] https://srb.help/7014
    12 |    T.reveal_type(x) # T::Array[T.untyped] (expected T::Array[String])
            ^^^^^^^^^^^^^^^^
  From:
    editor.rb:11:
    11 |  def foo2(x = [])
                   ^
    editor.rb:11:
    11 |  def foo2(x = [])
                       ^^
Errors: 2
```

#### Expected behavior

Expected `x` to be `T::Array[String]` in both methods (i.e. regardless of the default value).

jez:
- - - - -

👋 **Hey there!** This causing a lot of new type errors for your codebase?

You can either

- trace the type error back to an empty array literal in the file (`[]`), and replace it with `T::Array[T.untyped].new`
  - this explicitly creates an untyped array, rather than the old default behavior (`[]` is implicit untyped array)
- wrap the problems in `T.unsafe(...)` and fix the problems later
- downgrade the file to `# typed: false`, and fix the problems later

The benefit to you is:

```ruby
xs = compute_array_or_nil() || []
```

no longer types `xs` as `T::Array[T.untyped]` but instead as whatever the result type of `compute_array_or_nil()` says.

- - - - -

### Motivation

Instead of treating `[]` as type `T::Array[T.untyped]`, we treat it as
`[] (0-tuple)` and special case `TupleType::underlying` to continue to
decay to `T::Array[T.untyped]`.

Should fix #3751

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_IcGtcjILz1GaiY
→ https://go/builds/bui_IcGtNX4OLAnlXi

elliottt:
What about the following:

```ruby
# typed: true

x = []
x << 10

T.reveal_type(x)
```

Currently this will print out `T::Array[T.untyped]`, but with the change you're proposing it will print `[] (0-tuple)` despite there being one element in it now.

jez:
> Currently this will print out `T::Array[T.untyped]`, but with the change you're proposing it will print `[] (0-tuple)` despite there being one element in it now.

Yeah, and it causes 500 errors on pay-server as a result.

jez:
But in principle, the same problem exists if you had done this:

```ruby
# typed: true

x = [17]
x << 42

T.reveal_type(x)
```

https://sorbet.run/#%23%20typed%3A%20true%0Ax%20%3D%20%5B17%5D%0Ax%20%3C%3C%2042%0A%0AT.reveal_type(x)%0A

so arguably we shouldn't special case `[]` as much as we have and if we can easily migrate to this PR, then things would be more symmetric (and also it'd be more obvious when people need to add type annotations to prevent T.untyped from creeping in)

jez:
@elliottt Stripe's codebase is fixed so this should be good to go whenever you have a chance.

aprocter-stripe (approved):

PR #3839, opened 2020-12-22
Improve slow duplicate location check in renamer

soam-stripe:
Removes the expensive duplicate location check in the method renamer and replaces it with a hashmap.

### Motivation

Deferred review comments from previous PR #3831 

### Test plan

See included automated tests; also tested manually through vs code.

### DIFF
--- a/main/lsp/requests/rename.cc
+++ b/main/lsp/requests/rename.cc
@@ -198,7 +198,7 @@ variant<JSONNullObject, unique_ptr<WorkspaceEdit>> Renamer::buildEdit() {
         if (location == nullptr) {
             continue;
         }
-        tmpEdits[location->uri].push_back(make_unique<TextEdit>(move(location->range), newsrc));
+        tmpEdits[location->uri].push_back(make_unique<TextEdit>(move(location->range), move(newsrc)));
     }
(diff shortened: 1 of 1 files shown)


main/lsp/requests/rename.cc
```diff
@@ -187,13 +187,25 @@ variant<JSONNullObject, unique_ptr<WorkspaceEdit>> Renamer::buildEdit() {
...
+    // collect changes per file
     for (auto &item : edits) {
+        core::Loc loc = item.first;
+        string newsrc = item.second;
+        auto location = config.loc2Location(gs, loc);
+        ENFORCE(location != nullptr); // loc should always exist
+        if (location == nullptr) {
+            continue;
+        }
+        tmpEdits[location->uri].push_back(make_unique<TextEdit>(move(location->range), newsrc));
```
jvilk-stripe, line 201:
```suggestion
        tmpEdits[location->uri].push_back(make_unique<TextEdit>(move(location->range), move(newsrc)));
```

No need to copy the string twice.

jvilk-stripe (approved):
Excellent! I only had one minor optional request.

PR #3840, opened 2020-12-22
wip: Add Shape_squareBrackets intrinsic

jez:
Very wip: just trying to see how many errors this will cause.

### Motivation

Better typing for shape types.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2304,6 +2304,46 @@ public:
     }
 } Tuple_concat;
 
+namespace {
+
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_IcYGQQWDF3WOfq
→ https://go/builds/bui_IcYGoFNrAEOW5b

jez:
I think it might make sense to land this in the opposite order:

prevent `[]=` if it would widen the type of a known key (pinning error)

PR #3841, opened 2020-12-22
[wip] model foreign key props more accurately

froydnj:
Sorbet's modeling for props with foreign keys has some differences with how `sorbet-runtime` models them at runtime.  It would be better if the static and dynamic views matched up.

### Motivation

More accuracy is more better.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Prop.cc
+++ b/rewriter/Prop.cc
@@ -395,32 +395,37 @@ vector<ast::TreePtr> processProp(core::MutableContext ctx, PropInfo &ret, PropCo
             nonNilType = ASTUtil::dupType(ret.foreign);
         }
 
-        // sig {params(opts: T.untyped).returns(T.nilable($foreign))}
-        nodes.emplace_back(
-            ast::MK::Sig1(loc, ast::MK::Symbol(nameLoc, core::Names::opts()), ast::MK::Untyped(loc), std::move(type)));
+        ast::TreePtr boolean = ast::MK::UnresolvedConstant(loc, ast::MK::T(loc), core::Names::Constants::Boolean());
+        ast::TreePtr nilableBoolean = ast::MK::Nilable(loc, std::move(boolean));
 
-        // def $fk_method(**opts)
+        // sig {params(allow_direct_mutation: T.nilable(Boolean)).returns(T.nilable($foreign))}
+        nodes.emplace_back(ast::MK::Sig1(loc, ast::MK::Symbol(nameLoc, core::Names::allowDirectMutation()),
+                                         nilableBoolean.deepCopy(), std::move(type)));
+
+        // def $fk_method(allow_direct_mutation: nil)
         //  T.unsafe(nil)
         // end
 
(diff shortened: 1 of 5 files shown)


rewriter/Prop.cc
```diff
@@ -395,32 +395,37 @@ vector<ast::TreePtr> processProp(core::MutableContext ctx, PropInfo &ret, PropCo
...
-            ast::MK::Sig1(loc, ast::MK::Symbol(nameLoc, core::Names::opts()), ast::MK::Untyped(loc), std::move(type)));
+        ast::TreePtr boolean = ast::MK::UnresolvedConstant(loc, ast::MK::T(loc), core::Names::Constants::Boolean());
+        ast::TreePtr nilableBoolean = ast::MK::Nilable(loc, std::move(boolean));
+
+        // sig {params(allow_direct_mutation: T.nilable(Boolean)).returns(T.nilable($foreign))}
+        nodes.emplace_back(ast::MK::Sig1(loc, ast::MK::Symbol(nameLoc, core::Names::allowDirectMutation()),
+                                         nilableBoolean.deepCopy(), std::move(type)));
 
-        // def $fk_method(**opts)
+        // def $fk_method(allow_direct_mutation: nil)
```
jez, line 405: https://github.com/sorbet/sorbet/blob/3786e5a0e92ac3edd710f7cdfc1e0afe0d06e57f/gems/sorbet-runtime/lib/types/props/decorator.rb#L530

rewriter/Prop.cc
```diff
@@ -395,32 +395,37 @@ vector<ast::TreePtr> processProp(core::MutableContext ctx, PropInfo &ret, PropCo
...
 
-        // sig {params(opts: T.untyped).returns($foreign)}
-        nodes.emplace_back(ast::MK::Sig1(loc, ast::MK::Symbol(nameLoc, core::Names::opts()), ast::MK::Untyped(loc),
-                                         std::move(nonNilType)));
+        // sig {params(allow_direct_mutation: T.nilable(Boolean)).returns($foreign)}
+        nodes.emplace_back(ast::MK::Sig1(loc, ast::MK::Symbol(nameLoc, core::Names::allowDirectMutation()),
+                                         std::move(nilableBoolean), std::move(nonNilType)));
 
-        // def $fk_method_!(**opts)
+        // def $fk_method_!(allow_direct_mutation: nil)
```
jez, line 421: https://github.com/sorbet/sorbet/blob/75181be8e510664f7a0a7468e07132171eebef06/gems/sorbet-runtime/lib/types/props/decorator.rb#L554

jez (approved):

jez:
Be sure to test that this doesn't introduce errors just in case.

froydnj:

→ https://go/builds/bui_IcsJW3xXOHrof5
→ https://go/builds/bui_IcsJpDj4ae661u

froydnj:
This change doesn't actually work, because too many places define their own overrides with an `**opts` arglist, leading Sorbet to complain about the mismatched definitions.  And the modeling is not accurate enough in any event.

PR #3468, opened 2020-09-29
Partial ActiveSupport::Concern support through a rewriter

KaanOzkan:
Blocked by: https://github.com/sorbet/sorbet/pull/3959

Introduced ActiveSupport::Concern support. The rewriter supports using `class_methods` and `module ClassMethods` to include both instance methods and class methods into a class.

Rewriter works in 2 stages. First stage is triggering. We need to trigger the rewrite whenever `extend ActiveSupport::Concern` is found on the tree. Solution utilizes the existing tree traversal in `rewriter.cc` instead of doing an additional traversal inside the rewriter for performance reasons.

Second stage of the rewriter replaces nodes similar to other rewriters. It operates on a class definition and preserves the part of the body that is out of scope for the rewriter.

### Motivation

This is another step in enabling Sorbet to provide support for Rails constructs. In Rails projects ActiveSupport::Concern is heavily used since it simplifies the code: https://api.rubyonrails.org/classes/ActiveSupport/Concern.html#method-i-class_methods

Sorbet currently does not understand methods defined inside `class_methods` and those methods are not exposed to the class that includes the module.

### Test plan

See included automated tests. I tried to cover as many cases as I can think of. They are all in the same file for easier navigation.

To be merged after https://github.com/sorbet/sorbet/pull/3424

PR #3843, opened 2020-12-23
Put expected/actual types in error message

jez:
### Motivation

Better error messages

### Test plan

See included automated tests.

**Before**
Screen Shot 2020-12-22 at 11 56 59 PM

**After**
Screen Shot 2020-12-22 at 11 58 07 PM

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1164,7 +1164,8 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                                                            core::UntypedMode::AlwaysCompatible)) {
                     if (auto e = ctx.beginError(bind.loc, core::errors::Infer::ReturnTypeMismatch)) {
                         auto ownerData = ctx.owner.data(ctx);
-                        e.setHeader("Returning value that does not conform to method result type");
+                        e.setHeader("Expected `{}` but found `{}` for method result type", methodReturnType.show(ctx),
+                                    typeAndOrigin.type.show(ctx));
                         e.addErrorSection(core::ErrorSection(
                             "Expected " + methodReturnType.show(ctx),
                             {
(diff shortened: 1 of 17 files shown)


elliottt (approved):
This is a much nicer error message!

infer/environment.cc
```diff
@@ -1164,7 +1164,8 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                                                            core::UntypedMode::AlwaysCompatible)) {
                     if (auto e = ctx.beginError(bind.loc, core::errors::Infer::ReturnTypeMismatch)) {
                         auto ownerData = ctx.owner.data(ctx);
-                        e.setHeader("Returning value that does not conform to method result type");
+                        e.setHeader("Expected `{}` but found `{}` for method result type", methodReturnType.show(ctx),
```
Morriar, line 1167:
```suggestion
                        e.setHeader("Expected return type to be `{}` but found `{}`", methodReturnType.show(ctx),
```

Morriar (commented):
Nice!

infer/environment.cc
jez: I was planning to factor this all out into a helper in a future PR, where the form is always “Expected X but found Y for Z” (this is already the form that argument type errors and others fall into).

@Morriar how strongly do you feel here?

Morriar: Not strongly at all, but when reading the error, I'd prefer to have the "*return* type" information first so I know what I need to think of before reading the expected type.

jez: Sounds good. I’ll leave it as is for now, and in the PR to make all the errors more consistent I’ll revisit it.

przemyslaw2000:
#3468

PR #3846, opened 2020-12-24
WIP: Add timers to resolveConstants to track how much time we spend merging and sorting vectors

jvilk-stripe:
Add timers to resolveConstants to track how much time we spend merging and sorting vectors.

### Motivation

In Stripe's codebase, there is a gap between when the parallel threads finish processing files and the serial phase of resolve_constants begins. I do not know if this is attributable to sorting or insertion overhead.

### Test plan

Does not change behavior.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -863,6 +863,7 @@ public:
         vector<ClassMethodsResolutionItem> todoClassMethods;
 
         {
+            Timer timeit1(gs.tracer(), "resolver.resolve_constants.parallel_phase");
             ResolveWalkResult threadResult;
             for (auto result = resultq->wait_pop_timed(threadResult, WorkerPool::BLOCK_INTERVAL(), gs.tracer());
                  !result.done();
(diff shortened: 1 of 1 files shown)


jvilk-stripe:

→ https://go/builds/bui_IdIDIh0gDWFs3g
→ https://go/builds/bui_IdIDaCRJiosdKw

jvilk-stripe:
Survey says: It's sorting. We're spending time sorting. I'm going to try to break this down with finer-grained timers.

jvilk-stripe:

→ https://go/builds/bui_IdIUNjtscPD16u
→ https://go/builds/bui_IdIU5asMGpMOYx

jvilk-stripe:
Turns out that sort order is important. If we keep items in tree visit order, we end up putting ancestors in the wrong order! :( 

If we can make the tree visit order match what we expect, we can sort cheaply by moving contiguous vector items for a given file. I have code for this, but it requires us to enqueue ancestor resolution items in the correct order.

PR #3847, opened 2020-12-24
Fix sysread sig

Gasparila:
Documentation says:
> If the optional *outbuf* argument is present, it must reference a

so make the 2nd outbuf param optional

### Motivation

Existing code that does `some_file.sysread(8)` is failing typecheck

### Test plan

Only change is to rbi

### DIFF
--- a/rbi/core/io.rbi
+++ b/rbi/core/io.rbi
@@ -1621,7 +1621,7 @@ class IO < Object
     )
     .returns(String)
   end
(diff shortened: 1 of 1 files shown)


elliottt:

→ https://go/builds/bui_IhSmr9XgjEukLD
→ https://go/builds/bui_IhSmbm1ALBLK27

elliottt (approved):
👍

PR #3848, opened 2020-12-25
Fix Sorbet LSP tree copying crash in large codebases.

jvilk-stripe:
Fix Sorbet LSP tree copying crash in large codebases.

core::FileRef ids are u4s now.

### Motivation

Fixes a crash.

### Test plan

Not possible to easily test w/o a super large input to Sorbet.

### DIFF
--- a/main/lsp/LSPIndexer.cc
+++ b/main/lsp/LSPIndexer.cc
@@ -236,14 +236,14 @@ LSPFileUpdates LSPIndexer::commitEdit(SorbetWorkspaceEditParams &edit) {
 
     // Index changes in initialGS. pipeline::index sorts output by file id, but we need to reorder to match the order of
     // other fields.
-    UnorderedMap<u2, int> fileToPos;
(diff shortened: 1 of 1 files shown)


main/lsp/LSPIndexer.cc
```diff
@@ -236,14 +236,14 @@ LSPFileUpdates LSPIndexer::commitEdit(SorbetWorkspaceEditParams &edit) {
 
     // Index changes in initialGS. pipeline::index sorts output by file id, but we need to reorder to match the order of
     // other fields.
-    UnorderedMap<u2, int> fileToPos;
```
jvilk-stripe, line 239: On the slow path, this map will contain every file in the project. Files with IDs > 2^16 will overwrite entries of files with IDs < 2^16 in this map. One of those files will be left with a `nullptr` tree, causing a crash later on.

elliottt (approved):

PR #3853, opened 2020-12-28
Added "random" parameter to shuffle's sig

zanella:
### Motivation

The Array::shuffle method has a `random` parameter, it was missing from the `sig`

### Test plan

I'm not familiar w/ the codebase, nothing in the test directory jumped out as being a correct place to write a test for this use case, if there's one please point me to it and I'll try to write it (test)


elliottt:

→ https://go/builds/bui_Ihqa6Wzlz1UKFl
→ https://go/builds/bui_IhqaqdxwXUulo3

elliottt (approved):
Thanks!

PR #3857, opened 2021-01-03
Remove optional `base` parameter for `Float#to_s` in `float.rbi`

DanielGilchrist:
Currently `to_s` in the RBI for the `Float` class specifies that an optional parameter `base` can be passed. This is accurate for `Integer#to_s` but not `Float#to_s`.
Docs: https://ruby-doc.org/core-2.7.0/Float.html#method-i-to_s
image

### Motivation
Sorbet is unable to pick up on `Float#to_s` calls that could potentially cause errors during runtime.

### Test plan
See included automated tests.


DanielGilchrist:
Oh boy I accidentally cooked it 😬 Think I might close this and make a fresh PR my bad

Issue #3411, opened 2020-09-09
Return type for Array#zip is wrong when more than 1 arguments are passed


DarkDimius:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0As%20%3D%20%5B1%5D.zip(%5B2%5D%2C%20%5B3%5D)%0AT.reveal_type(s))

```ruby
# typed: true
s = [1].zip([2], [3])
T.reveal_type(s)
```

#### Observed output

```
editor.rb:3: Revealed type: T::Array[[Integer, T.nilable(Integer)]] https://srb.help/7014
```

#### Expected behavior

`T::Array[[Integer, T.nilable(Integer), T.nilable(Integer)]] `
- - -

DarkDimius:
Removing RBI label as this will have to be intrinsified

PR #3858, opened 2021-01-03
Fix return type for Array#zip when more than 1 arguments are passed

izquierdo:
Used the "intrinsic method" approach instead of modifying RBIs based on this comment from the linked issue:

> Removing RBI label as this will have to be intrinsified

This is based on the [existing Array#product](https://github.com/sorbet/sorbet/blob/8eb243fd44f5260a985aacad38b12b4ea6a1bc40/core/types/calls.cc#L2461) intrinsic method code. I'm still getting familiar with the codebase so there are a few details that aren't fully clear to me yet, such as why `{AppliedType, TupleType}` is the full set of types to check for here. Any pointers around this area are appreciated.

### Motivation

Fixes https://github.com/sorbet/sorbet/issues/3411

### Test plan

See included automated tests.

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -355,10 +355,11 @@ NameDef names[] = {
     // Enumerable#flat_map has special-case logic in Infer
     {"flatMap", "flat_map"},
 
-    // Array#flatten, #product and #compact are also custom-implemented
+    // Array#flatten, #product, #compact and #zip are also custom-implemented
     {"flatten"},
     {"product"},
     {"compact"},
+    {"zip"},
 
     // Pattern matching
     {"patternMatch", "<pattern-match>"},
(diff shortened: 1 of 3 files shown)


elliottt:

→ https://go/builds/bui_IhphdC8tCe5gOj
→ https://go/builds/bui_Ihph9h31yxDePY

elliottt (approved):
This looks good, thank you! It raised a couple of issues on Stripe's codebase, so I'll take it from here and land the PR once our codebase has been migrated.

To answer your question about AppliedType and TupleType:
* singleton classes are represented as AppliedType internally in sorbet, so the singleton class for `Array` will be an AppliedType
* TupleType is used to type array literals, so using syntax like `[1]` will end up making a value whose internal type is TupleType
Any other type of value receiving the zip method wouldn't be correct in this case, so it's fine to raise an error and return when the receiver is not one of those two.

elliottt:
Stripe migration PR: https://go/ps/pull/292145

Issue #498, opened 2019-05-13
T.proc.bind does not work if the block param is also T.nilable(...)


XrXr:
### Problem

It is possible to rebind the type of `self` in a block: 

```ruby
# typed: true
class SelfSwitch
  extend T::Sig
  sig { params(block: T.proc.bind(Integer).void).void }
  def foo(&block)
  end
end


SelfSwitch.new.foo do
  T.reveal_type(self)   # Revealed type: Integer
end
```

So it seems natural to allow this capability for optional blocks too:

```ruby
# typed: true
class SelfSwitch
  extend T::Sig

  # Using `bind` is not permitted here https://sorbet.org/docs/error-reference#5004
  sig { params(block: T.nilable(T.proc.bind(Integer).void)).void }
  def optional_block(&block)
  end
end

```

manhhung741:
@AlexEvanczuk @trevor-stripe  The issue with `T.nilable` is reported here. @XrXr did a good job filing the bug already. I also find it doesn't work with `T.any` (T.nilable is just T.any under the hood).

Here is a [sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aextend%20T%3A%3ASig%0Aclass%20A%0A%20%20def%20m%3B%20end%0Aend%0A%0Asig%20%7B%20params(block%3A%20T.nilable(T.proc.bind(A).void)).void%20%7D%0Adef%20run(%26block)%0A%20%20self.m%0Aend%0A%0Arun%20%7B%20'Hello'%20%7D)

deecewan:
It also can't be used inside of `T.type_alias`

connorshea:
This is making it hard to create a type definition for graphql-ruby :(

deecewan:
@connorshea that's where I first ran into this issue as well.

L-as:
Are there any major blockers for implementing this, or has it just gone without attention? Wondering if it's possible for me to fix it.

deecewan:
@L-as I think it's just going to be difficult (based on the `hard` label).

djrodgerspryor:
I'm having a crack at writing types for [Sequel](https://github.com/jeremyevans/sequel) and this has basically stopped me dead — Sequel makes heavy use of optional blocks with custom bindings.

Is there any workaround for this?

I'd be interested in making a PR to try and support this properly in Sorbet, but the 'hard' label makes me suspect that there are known problems — it would be super helpful if they were documented here @jez

[...]

PR #3859, opened 2021-01-04
Support T.proc.bind when block is nilable

dpkp:
Attempting to specify a binding for a nilable block currently results in an error in the resolver stage. The reason for this appears to be that a bare `T.proc` is resolved w/ rebinding enabled, while all other `T.*` uses are resolved without. There is a method, `isTProc`, that is used to identify the former case; all others are passed to `interpretTCombinator`, which attempts to resolve using recursive calls back into the outer resolver. These recursive calls do not support rebinding and I believe this is why it is not currently possible to specify a binding when using a nilable T.proc.

This patch adds a companion helper function, `isNilableTProc`, and uses it to branch interpretTCombinator logic so that nilable T.procs allow rebinding. I don't believe that we can simply add nilable T procs to the current `isTProc` block because the resolver needs to bind the outer type -- the T.nilable -- not the inner type -- the T.proc. But I'm happy for feedback on this.

There are 2 commits included, each solving this problem in a different way. The first commit is a less intrusive change that identifies nilable procs prior to calling interpretTCombinator. It does not modify interpretTCombinator but instead adds a binding after receiving the type from the current implementation. The downside of this approach is that it needs to resolve the T.proc 2x and this causes some errors to be reported multiple times. To avoid 2x resolver calls here requires a bigger change, and that is implemented in the second commit. This commit changes the return type of `interpretTCombinator` from its current signature: returning a type only, to a new signature: return a full Result (a struct of type + binding). This change requires updating every return case and is a bit more intrusive. It's also uglier, and forgive me as I'm not a C++ expert!

Anywho, let me know if this approach makes sense, whether you think there's a better alternative here, and/or whether you'd prefer the commit 1 or commit 2 approach.

Happy trails, and thanks for all your work on this project!

### Motivation
Issue  #498 . Like other commenters on the issue, flexport uses graphql-ruby and would like to improve sorbet coverage. The use of nilable blocks for optional configuration is a common pattern in that and other libraries.

### Test plan
See included automated tests.

### DIFF
--- a/resolver/type_syntax.cc
+++ b/resolver/type_syntax.cc
@@ -464,21 +464,23 @@ ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend
     return sig;
 }
 
-core::TypePtr interpretTCombinator(core::Context ctx, const ast::Send &send, const ParsedSig &sig,
-                                   TypeSyntaxArgs args) {
+TypeSyntax::ResultType interpretTCombinator(core::Context ctx, const ast::Send &send, const ParsedSig &sig,
+                                            TypeSyntaxArgs args) {
     switch (send.fun.rawId()) {
         case core::Names::nilable().rawId():
             if (send.numPosArgs != 1 || send.hasKwArgs()) {
-                return core::Types::untypedUntracked(); // error will be reported in infer.
+                return {.type = core::Types::untypedUntracked()}; // error will be reported in infer.
             }
-            return core::Types::any(ctx,
-                                    getResultTypeAndBindWithSelfTypeParams(ctx, send.args[0], sig,
-                                        isNilableTProc(ctx, &send) ? args.withRebind() : args.withoutRebind()).type,
-                                    core::Types::nilClass());
+            if (isNilableTProc(ctx, &send)) {
+                auto tproc = getResultTypeAndBindWithSelfTypeParams(ctx, send.args[0], sig, args.withRebind());
+                return {.type = core::Types::any(ctx, tproc.type, core::Types::nilClass()), .rebind = tproc.rebind};
+            }
+            return {.type = core::Types::any(ctx, getResultTypeWithSelfTypeParams(ctx, send.args[0], sig, args),
+                                             core::Types::nilClass())};
         case core::Names::all().rawId(): {
             if (send.args.empty()) {
                 // Error will be reported in infer
-                return core::Types::untypedUntracked();
+                return {.type = core::Types::untypedUntracked()};
             }
             auto result = getResultTypeWithSelfTypeParams(ctx, send.args[0], sig, args);
             int i = 1;
@@ -486,12 +488,12 @@ core::TypePtr interpretTCombinator(core::Context ctx, const ast::Send &send, con
                 result = core::Types::all(ctx, result, getResultTypeWithSelfTypeParams(ctx, send.args[i], sig, args));
                 i++;
             }
-            return result;
+            return {.type = result};
         }
         case core::Names::any().rawId(): {
             if (send.args.empty()) {
                 // Error will be reported in infer
-                return core::Types::untypedUntracked();
+                return {.type = core::Types::untypedUntracked()};
             }
             auto result = getResultTypeWithSelfTypeParams(ctx, send.args[0], sig, args);
             int i = 1;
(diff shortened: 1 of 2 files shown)


resolver/type_syntax.cc
```diff
@@ -450,143 +464,147 @@ ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend
...
             }
-            return core::Types::any(ctx, getResultTypeWithSelfTypeParams(ctx, send.args[0], sig, args),
-                                    core::Types::nilClass());
+            return {.type = core::Types::any(ctx, getResultTypeWithSelfTypeParams(ctx, send.args[0], sig, args),
+                                             core::Types::nilClass())};
         case core::Names::all().rawId(): {
             if (send.args.empty()) {
                 // Error will be reported in infer
-                return core::Types::untypedUntracked();
+                return {.type = core::Types::untypedUntracked()};
```
elliottt, line 483: Could you change these to `TypesSyntax::ResultType{core::Types::untypedUntracked(), nullptr};`? This should be an error as we have `-Werror -Wmissing-field-initializers` in .bazelrc, but it only seems to raise with gcc.

elliottt (changes requested):
I had a couple of requests for the implementation, but otherwise this seems pretty reasonable. Have you experimented with cases like `T.nilable(T.nilable(... T.proc.bind(String).void...))`? I think that they should work exactly the same way, as `T.nilable(T.nilable(String))` is equivalent to `T.nilable(String)`, but it would be nice to have a test to back this up.

dpkp:
> I had a couple of requests for the implementation, but otherwise this seems pretty reasonable. Have you experimented with cases like `T.nilable(T.nilable(... T.proc.bind(String).void...))`? I think that they should work exactly the same way, as `T.nilable(T.nilable(String))` is equivalent to `T.nilable(String)`, but it would be nice to have a test to back this up.

The current implementation does not support this because `isTProc` only recurses through the sender's `recv` , not its args. So only the inner T.nilable would get recognized by isTProc, and all outer resolves would use the existing branch that does not support rebinding.

The solution to this is a little more intrusive: to support rebinding when resolving all `T.nilable` . I will push up this change for review, as I think it is ok right now: the only place that sets a result rebind is the current TProc block. But I'm not sure if this is inconsistent with future plans or design? But I will add a test and push the change for you to review.

Thanks for the feedback!

elliottt:
> The current implementation does not support this because `isTProc` only recurses through the sender's `recv` , not its args. So only the inner T.nilable would get recognized by isTProc, and all outer resolves would use the existing branch that does not support rebinding.

Yes, my suggestion was to apply `isTProc` to the first argument of the send if you pass the check for the outer send being `T.nilable`.

resolver/type_syntax.cc
dpkp: yes, though `nullptr` does not work here because the TypeSyntax::ResultType struct includes a full core::SymbolRef object, not a pointer. I believe the default value is `core::SymbolRef()`, so I'll use that to be explicit.

dpkp:
> > The current implementation does not support this because `isTProc` only recurses through the sender's `recv` , not its args. So only the inner T.nilable would get recognized by isTProc, and all outer resolves would use the existing branch that does not support rebinding.
> 
> Yes, my suggestion was to apply `isTProc` to the first argument of the send if you pass the check for the outer send being `T.nilable`.

Yes, but this approach only detects a `T.nilable(T.proc(...` . It definitely simplifies the block and eliminates the need for the isNilableTProc helper and so I'll removed that. But it still doesn't handle the nested `T.nilable(T.nilable(T.proc(...` case. The problem is this case only passes the `isTProc` test on the inner resolve (`T.nilable(T.proc(...`). And because the outer resolve (`T.nilable(T.nilable(...`) uses the current logic, it ignores the inner's rebind.

elliottt:
> Yes, but this approach only detects a `T.nilable(T.proc(...` . It definitely simplifies the block and eliminates the need for the isNilableTProc helper and so I'll remove that. But it still doesn't handle the nested `T.nilable(T.nilable(T.proc(...` case. The problem is this case only passes the `isTProc` test on the inner resolve (`T.nilable(T.proc(...`). And because the outer resolve (`T.nilable(T.nilable(...`) uses the current logic, it ignores the inner's rebind.
> 
> I'll push new commits once local test suite passes. That just takes a little while ;)

Ah good point, thanks!

dpkp:
I've pushed 3 new commits: the first adds (failing) test cases for nested T.nilable. The second fixes up the struct initialization. And the third simplifies the interpretTCombinator logic, removing isNilableTProc and supporting nested T.nilable by always returning the recursed inner rebind value.

Let me know what you think. If you'd like me to squash commits or otherwise rework anything to make it more likely to land, let me know!

elliottt (approved):
Thanks for the refactoring, I like this approach much more!

elliottt:

→ https://go/builds/bui_IidmHSL1EGpg3j
→ https://go/builds/bui_IidmoBejItsuQ2

connorshea:
@dpkp thank you for this fix, it's made Sorbet much more useful! :D

PR #3861, opened 2021-01-04
IO's `encoding` also accepts an Encoding

Morriar:
### Motivation

As shown in the [documentation](https://ruby-doc.org/core-2.7.0/IO.html#method-c-read-label-Options), `IO.read(encoding:)` can be either a `String` or an `Encoding`.

### Test plan

See included automated tests.


elliottt:

→ https://go/builds/bui_IhjbC4yGkwTYQG
→ https://go/builds/bui_Ihjb3XtkZ6JwTx

elliottt (approved):
Thanks for the fix and tests!

PR #3862, opened 2021-01-04
tweak prop rewriter to not use else after a return

froydnj:
As per `$SUBJECT`

### Motivation

Clearer code.

### Test plan

Existing automated tests.


elliottt (approved):
Hiding whitespace changes makes this PR really readable :D

PR #3864, opened 2021-01-04
add a variadic template helper for building Send::ARGS_store

froydnj:
Building `Send::ARGS_store` by hand is a little tedious, so let's add a variadic template helper to automate some of the work for us.  (Thank you, C++ 17, for fold expressions!)

### Motivation

For more accurate modeling of foreign props, I've had to build a number of send arglists manually, and it's a little bit annoying.  This helper (and accompanying changes in the rewriter to use it) would make things somewhat easier.  I think it mostly makes the code clearer, even if `clang-format` makes some things a bit harder to match up to the comments.

I didn't add the same for `MethodDef::ARGS_store` yet because the existing uses don't seem to warrant it, and I haven't needed it in the foreign props.

### Test plan

Existing automated tests.

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -28,6 +28,12 @@ public:
         return Block(loc, std::move(body), std::move(args));
     }
 
+    template <typename... Args> static Send::ARGS_store SendArgs(Args &&... args) {
+        Send::ARGS_store store;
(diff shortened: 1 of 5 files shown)


elliottt (approved):
This is a nice clean-up!

PR #3866, opened 2021-01-05
fix types for foreign prop stub methods

froydnj:
We were using the type of the  loader for the return type of the foreign prop stub methods, which is not correct; they should be returning the type of the prop instead.

### Motivation

This probably doesn't matter much at the moment, but it does matter when/if foreign prop methods expand into something larger than stubs.

### Test plan

See included automated tests.  I snuck in an extra test as well.


froydnj:

→ https://go/builds/bui_Ihg9LVbX6D8rFj
→ https://go/builds/bui_Ihg965OVqRKFmU

froydnj:
Despite this passing tests on Stripe's codebase, it turns out this patch is badly misguided.

PR #3867, opened 2021-01-05
Fix the signature in Find::find

lugray:
As per the docs included in the file it "Returns an enumerator if no block is given." This allows it to be used in that way. This is still not ideal as it then requires use as, e.g.

```
T.must(Find.find(some_path)).filter { ... }
```

even though we should allow

```
Find.find(some_path).filter
```

since the use of it without a block ensures that an enumerator will be returned. It also gives an explicit return of `nil` when used with a block though the original `void` is more accurate.

Ideally, we'd have a mechanism for alternate signatures:

```
sig do
  T::Sig.any(
    params(paths: String, ignore_error: T::Boolean).returns(Enumerator),
    params(paths: String, ignore_error: T::Boolean, blk: T.proc.params(path: String).void).void,
  )
end
```

Does something like that exist?

cc @Morriar

### DIFF
--- a/rbi/stdlib/find.rbi
+++ b/rbi/stdlib/find.rbi
@@ -33,8 +33,8 @@ module Find
     params(
       paths: String,
       ignore_error: T::Boolean,
-      blk: T.proc.params(path: String).void
-    ).void
+      blk: T.nilable(T.proc.params(path: String).void)
+    ).returns(T.nilable(T::Enumerator[String]))
   end
   def self.find(*paths, ignore_error: true, &blk); end
 


elliottt:
> Does something like that exist?

We do allow for signature overloads in standard library rbis, but currently it's only supported for methods that don't use keyword arguments. This function is a great example of where that functionality would be useful :)

elliottt:

→ https://go/builds/bui_IhjXje9wDolFzU
→ https://go/builds/bui_IhjXwcLZZKqULj

rbi/stdlib/find.rbi
```diff
@@ -33,8 +33,8 @@ module Find
     params(
       paths: String,
       ignore_error: T::Boolean,
-      blk: T.proc.params(path: String).void
-    ).void
+      blk: T.nilable(T.proc.params(path: String).void)
+    ).returns(T.nilable(Enumerator))
```
elliottt, line 37:
```suggestion
    ).returns(T.nilable(T::Enumerator[String]))
```

elliottt (changes requested):
Tests failed because of the use of `Enumerator` without supplying the type of elements yielded. I made a suggestion for how to fix it, but am not entirely confident that the elements yielded are always `String`.

rbi/stdlib/find.rbi
lugray: It will indeed always be a`String` (file or dir name). I've applied your suggestion.

elliottt:

→ https://go/builds/bui_IhkEPqffYitteb
→ https://go/builds/bui_IhkEo6HXNr3LEr

elliottt (approved):
Thank you!

PR #3874, opened 2021-01-08
core::SymbolRef() -> core::Symbols::noSymbol()

jez:
### Motivation

We should be more explicit. Nearly every new usage of this was introduced in
#3859, merged a few hours ago.

### Test plan

Existing tests.


elliottt (approved):

dpkp:
thanks my droogs

PR #3876, opened 2021-01-10
Improve error message for "method does not exist"

jez:
Threads the TypeAndOrigin information of the receiver all the way from
`processBinding` to `dispatchCallSymbol`, so that we can show an
`origin2Explanation` error message for "method does not exist" errors.

### Commit summary

Review by commit.

- **Thread origin of receiver through DispatchArgs** (b2a013cb9)

  Purely mechanical (follow the compiler).

- **guessOverload doesn't actually use fullType arg** (efbe071db)

  I happened to notice that an arg wasn't being used, so I deleted it from the
  definition and call site.

- **Use origins2Explanations in more places** (cd016ea8c)

  The actual change (one line).

### Motivation

In particular I think that this will greatly help with future changes to make
shape types better, because most of the newly introduced errors take this
shape--in the future, we could consider autocorrects that use this
information:

- Wrap the definition in `T.unsafe` (maybe behind a flag)
  - i.e., this could be a Big Hammer for future codemods
- Wrap the definition in `T.let(..., T::Hash[T.untyped, T.untyped])` to get
  back the previous behavior
- etc.

I haven't added any such autocorrects in this PR.

### Test plan

This change only affects non-header error information:

**Before**
Screen Shot 2021-01-09 at 5 28 05 PM

**After**
Screen Shot 2021-01-09 at 5 27 09 PM


test/cli/autocorrect-helpers/autocorrect-helpers.out
```diff
@@ -6,6 +6,10 @@ autocorrect-helpers.rb:4: Method `interface!` does not exist on `T.class_of(S1)`
 `
      4 |  interface!
         ^
+  Got T.class_of(S1) originating from:
+    autocorrect-helpers.rb:3:
+     3 |module S1
```
jez, line 11: Looks like the loc that we use for `<self>` when it's implicit is the `declLoc` of the containing definition.

Not great, but also not the thing I'm trying to change in this PR.

test/cli/incremental-resolver/incremental-resolver.out
```diff
@@ -48,6 +48,14 @@ test/cli/incremental-resolver/expect-failures/constant_override.rb:111: Redefini
...
      110 |B = e
               ^
+  Got T.class_of(<root>) originating from:
+    test/cli/incremental-resolver/expect-failures/constant_override.rb:110:
+     110 |B = e
+     111 |module B
+     112 |  extend T::Sig
+     113 |  sig { returns(T.all(B,T)) }
+     114 |  def foo; T.unsafe(nil); end
+     115 |end
```
jez, line 58: Similarly, it looks like the loc we use for `<self>` when it's the file top level is a loc meaning "the entire file" (because there is implicitly a class def (with a declLoc) around the entire file contents' AST)

test/cli/no-stdlib/no-stdlib.out
```diff
@@ -3,4 +3,8 @@ No errors! Great job.
 -e:1: Method `puts` does not exist on `T.class_of(<root>)` https://srb.help/7003
      1 |puts 'Hello > World'
         ^^^^^^^^^^^^^^^^^^^^
+  Got T.class_of(<root>) originating from:
+    -e:1: Possibly uninitialized (`NilClass`) in:
```
jez, line 7: Our error printing make this show up as a weird loc, because it just so happens that the loc we use for `<self>` at the file top level is also the same loc we use for pointing to uninitialized variables.

test/cli/suggest-kernel/suggest-kernel.out
```diff
@@ -1,6 +1,11 @@
 test/cli/suggest-kernel/suggest-kernel.rb:4: Method `raise` does not exist on `Foo` https://srb.help/7003
      .. |    raise "hi"
             ^^^^^^^^^^
+  Got Foo originating from:
+    test/cli/suggest-kernel/suggest-kernel.rb:3:
+     .. |  def foo
+     .. |    raise "hi"
+     .. |  end
```
jez, line 8: Looks like for method defs we use the method loc, instead of the declLoc? :psyduck:

core/types/calls.cc
```diff
@@ -569,6 +569,8 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
                     }
                 }
             }
+            e.addErrorSection(ErrorSection("Got " + args.fullType.type.show(gs) + " originating from:",
+                                           args.fullType.origins2Explanations(gs, args.originForUninitialized)));
```
jez: This is the core of the change.

test/cli/autocorrect-helpers/autocorrect-helpers.out
jez: Update: I decided to change this in this PR.

elliottt (approved):

PR #3877, opened 2021-01-10
Add more information to error message

dogweather:
Currently:

```
Sorbet's language server requires a single input directory.
```

With this PR:

```
Sorbet's language server requires a single input directory. However, 2 are configured: /x/y/z, /w/x/y/z
```

### Motivation

I'm personally getting the above message, and I can only guess why it's happening. However, the code has all the context info to help debug the problem.

### Test plan

I'm completely guessing how to implement this. I'm not set up for C++ development, and I took a stab at the proper way to output this info. I appreciate any touch-ups needed!


jez:
Thanks. I fixed the compilation error in your change and added a test.

jez (approved):

Issue #3674, opened 2020-11-17
Hidden definitions cannot be generated for methods with more than one anonymous variable


ajoneil:
Methods with more than one anonymous/unnamed variable generate invalid method definitions in hidden definitions. For example, if you include the `statsd-instrument` gem in your project, there are issues:

Statsd-instrument generates a number of methods with unnamed parameters, see https://github.com/Shopify/statsd-instrument/blob/2772e56f10ca614238817167f10be86ff4a4bbb6/lib/statsd/instrument/datagram_builder.rb#L22.

This causes the hidden definitions generation to put this in `reflection.rbi`:
```ruby
class StatsD::Instrument::StatsDDatagramBuilder < StatsD::Instrument::DatagramBuilder
  sig do
    params(
      _: ::T.untyped,
      _: ::T.untyped,
      _: ::T.untyped,
      _: ::T.untyped,
    )
    .returns(::T.untyped)
  end
  def d(_, _, _, _); end

  sig do
    params(
      _: ::T.untyped,
      _: ::T.untyped,
      _: ::T.untyped,
      _: ::T.untyped,
    )
    .returns(::T.untyped)
  end
  def h(_, _, _, _); end

  sig do
    params(
      _: ::T.untyped,
      _: ::T.untyped,
      _: ::T.untyped,
      _: ::T.untyped,
    )
    .returns(::T.untyped)
  end
  def kv(_, _, _, _); end

end
```

This then causes a bunch of the following errors in `reflection.json.err`:
```
/tmp/d20201117-13-f33zr2/reflection.rbi:753996: Hash key `_` is duplicated https://srb.help/3011
      753996 |      _: ::T.untyped,
                    ^
    /tmp/d20201117-13-f33zr2/reflection.rbi:753994: First occurrence of `_` hash key
      753994 |      _: ::T.untyped,
                    ^

/tmp/d20201117-13-f33zr2/reflection.rbi:753997: Hash key `_` is duplicated https://srb.help/3011
      753997 |      _: ::T.untyped,
                    ^
    /tmp/d20201117-13-f33zr2/reflection.rbi:753994: First occurrence of `_` hash key
      753994 |      _: ::T.untyped,
```

This issue with `statsd-instrument` as well as related issue #3603 with `aws-sdk-core` has meant we've had to lock our project to sorbet 0.5.5944, which is the last known release without these issues. It's broken in 0.5.6001, so the issues must've been introduced between those two versions.

jez:
[...]

PR #3879, opened 2021-01-11
Fix more than one unnamed parameter in method definition

tvallois:
Hi, first thanks for Sorbet, an amazing tool for ruby developers. :+1: 

In this change, i'm looking at the number of parameters equal to underscore because it is the only parameter that can occur multiple times in a same function in ruby.
Looking at this comment https://github.com/sorbet/sorbet/issues/3674#issuecomment-728674364, a suggestion was to remove the function from hidden-definitions, that's what i've done in this PR.

### Motivation

Since we're upgraded sorbet-runtime from 0.5.5944 to the latest version, we're unable to regenerate rbis anymore. We have the error describe in the issue above.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet/test/snapshot/partial/unnamed_arguments/src/src.rb
+++ b/gems/sorbet/test/snapshot/partial/unnamed_arguments/src/src.rb
@@ -1,0 +1,5 @@
+class UnnamedArguments
+  def first_method(_, _)
+    nil
+  end
+end
--- a/gems/sorbet/lib/serialize.rb
+++ b/gems/sorbet/lib/serialize.rb
@@ -197,7 +197,7 @@ class Sorbet::Private::Serialize
 
   def blacklisted_method(method)
     return true if /__validator__[0-9]{8}/ || method.name =~ /.*:.*/
-    return true if method.parameters.map { |(kind, name)| name }.filter { |name| name.to_s == "_" }.length >= 2
+    return true if method.parameters.map { |(kind, name)| name }.select { |name| name.to_s == "_" }.length >= 2
     false
   end
 
--- a/gems/sorbet/lib/serialize.rb
+++ b/gems/sorbet/lib/serialize.rb
@@ -328,7 +328,7 @@ class Sorbet::Private::Serialize
   def from_method(method)
     uniq = 0
     method.parameters.map.with_index do |(kind, name), index|
-      if !name
+      if !name || name.to_s == "_"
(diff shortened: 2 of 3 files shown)


elliottt (changes requested):
Thanks!

Could you add a test in `/gems/sorbet/test/snapshot/partial` that shows that methods with multiple unnamed arguments don't make it into reflection.rbi?

tvallois:
Hi,

Should be good now ! Thanks for review @elliottt

gems/sorbet/test/snapshot/partial/unnamed_arguments/src/src.rb
```diff
@@ -0,0 +1,5 @@
+class UnnamedArguments
+  def first_method(_, _)
+    nil
+  end
```
jez, line 4: Do you want to add another method to this test case in a way that Sorbet can't see the definition statically?

```suggestion
  def first_method(_, _)
    nil
  end
  
  define_method(:second_method) do |_, _|
    nil
  end
```

jez:
Thanks for the PR. It seems to me that a better solution would be to rename the positional arguments from `_, _` to `_1, _2` if we detect that they're duplicated, instead of just not serializing the method.

Sorbet doesn't actually care about the name of positional arguments to a method, just how many there are (and if there's a sig, that they match what's mentioned in the sig).

While I see that the current behavior is obviously bad (`_, _` will immediately generate an error), I wonder if while you're here that you want to take a shot at fixing the root of the problem, rather than just treating the symptoms?

tvallois:
> Thanks for the PR. It seems to me that a better solution would be to rename the positional arguments from `_, _` to `_1, _2` if we detect that they're duplicated, instead of just not serializing the method.
> 
> Sorbet doesn't actually care about the name of positional arguments to a method, just how many there are (and if there's a sig, that they match what's mentioned in the sig).
> 
> While I see that the current behavior is obviously bad (`_, _` will immediately generate an error), I wonder if while you're here that you want to take a shot at fixing the root of the problem, rather than just treating the symptoms?

hey @jez , definitely, it was my first idea to rename the positional arguments from `_, _` to `_1, _2`. Do you want me to create a new merge request to apply those changes or should i do everything on this one?

jez:
Heads up I fixed this in a different way in #4055

jez:
Closed via #4055

PR #3880, opened 2021-01-11
Fix `T::Struct#with` behavior for nested structs and enums

aisamanra:
### Motivation
Right now, if you use `T::Struct#with` to update a field whose type is an enum, it fails: this is because it uses the deserialization code to implement `#with` but enums do not robustly handle being asked to deserialize a version of themselves (rather than the actual serialized format). This adds a special case to enum deserialization that handles this, which makes `#with` work again.

### Test plan

Added a runtime test: [here is what this test looks like when it fails](https://buildkite.com/sorbet/sorbet/builds/14306#01b704c8-c126-482b-a44c-b3e6d2571901).

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/enum.rb
+++ b/gems/sorbet-runtime/lib/types/enum.rb
@@ -90,6 +90,7 @@ class T::Enum
   # @raise [KeyError] if serialized value does not match any instance.
   sig {overridable.params(serialized_val: SerializedVal).returns(T.attached_class).checked(:never)}
   def self.from_serialized(serialized_val)
+    return serialized_val if serialized_val.is_a?(self)
     res = try_deserialize(serialized_val)
     if res.nil?
       raise KeyError.new("Enum #{self} key not found: #{serialized_val.inspect}")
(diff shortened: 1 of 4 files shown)


gems/sorbet-runtime/lib/types/enum.rb
```diff
@@ -90,6 +90,7 @@ def self.try_deserialize(serialized_val)
   # @raise [KeyError] if serialized value does not match any instance.
   sig {overridable.params(serialized_val: SerializedVal).returns(T.attached_class).checked(:never)}
   def self.from_serialized(serialized_val)
+    return serialized_val if serialized_val.is_a?(self)
```
jez, line 93: This seems wrong. Why is it ok for `from_serialized` to get something that is not the output of `serialize`?

At the very least, we should change the sig (here, and in the RBI) to make it clear that `from_serialized` is sometimes called with `T.attached_class`.

aisamanra: On reflection, you're right, this is the wrong place. The problem here is in `#with`: I've changed the approach to modify that.

jez (approved):
is there a similar bug if the prop is another `T::Struct` instead of a `T::Enum`?

aisamanra:
It does look like this bug is present for `T::Struct`, too. I've gone ahead and fixed that and also included tests for both.

PR #3882, opened 2021-01-12
Differentiate optional from nilable optional

zanella:
Added one more param to the documentation example to clarify the distinction between a nilable param and an optional && nilable one.


elliottt (changes requested):
Thanks for this doc change!

What do you think about making the `z` parameter in both examples have a default value of `nil` instead of `'bar'`? That would keep it close to the original doc while also communicating the distinction between nilable and defaulted arguments.

elliottt (approved):

PR #3885, opened 2021-01-13
only allow symbol-typed names for props and consts

ezra-stripe:
Narrow the allowable types for names of the `prop` and `const` DSLs. In practice, name-types that do not repond to `to_sym` already fail, since they are symbolized internally.

### Motivation

There are no non-symbol prop or const names internally at Stripe, and we'd like to keep it that way.

### Test plan

Not yet tested integrated with internal Stripe code due to issues with Github. While resolving that, I wanted to open the PR and ensure this is a change we want. It is technically a breaking change, and I don't know if users outside of Stripe are relying on String typed `prop` and `const` names. This might not be the right layer of the stack to enforce our preferences.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -75,40 +75,6 @@ module T::Private::Methods
     T::Types::Proc.new(decl.params, decl.returns)
   end
 
-  # See docs at T::Utils.register_forwarder.
-  def self.register_forwarder(from_method, to_method, mode: Modes.overridable, remove_first_param: false)
-    # Normalize the method (see comment in signature_for_key).
-    from_method = from_method.owner.instance_method(from_method.name)
-
-    from_key = method_to_key(from_method)
-    maybe_run_sig_block_for_key(from_key)
-    if @signatures_by_method.key?(from_key)
-      raise "#{from_method} already has a method signature"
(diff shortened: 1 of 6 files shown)


jez:
(removing my review—feel free to re-request once non-wip)

ezra-stripe:
cc @trevor-stripe

elliottt:

→ https://go/builds/bui_IkkqfZWf0w49Vk
→ https://go/builds/bui_IkkrGiVssXJ0jT

ezra-stripe:
> I've kicked off a test run for the current PR. When the build
> finishes, I'll share with you whether or how it failed. Thanks!
> 
> 
> → https://go/builds/bui_IkkqfZWf0w49Vk
> → https://go/builds/bui_IkkrGiVssXJ0jT

Thanks for that. I'm still interested in discussing whether or not we want the change though. WDYT?

elliottt (changes requested):
@ezra-stripe could you add a test for each in `gems/sorbet-runtime/test/types/struct.rb` showing what the behavior of these methods is with `String` arguments? It would be good to make sure that we're raising runtime errors as well.

elliottt:

→ https://go/builds/bui_IpGtBGTUnCkART
→ https://go/builds/bui_IpGtNeChbo1Rd1

elliottt (approved):
Thanks for the test!

PR #3891, opened 2021-01-14
Add `module_function` to various modules

SeekingMeaning:
This PR marks the following as `module_function`:
- `Find::find` and `Find::prune` - https://github.com/ruby/ruby/blob/v2_6_0/lib/find.rb#L87
- `Kconv` methods - https://github.com/ruby/ruby/blob/v2_6_0/ext/nkf/lib/kconv.rb
- `Open3` methods
- Some `Shellwords` methods
- Some `Drb` methods

### Motivation

 Parity with Ruby standard library

### Test plan

None


elliottt:

→ https://go/builds/bui_Il6Z5lmh2um2NY
→ https://go/builds/bui_Il6ZUBX2KB93VI

elliottt (approved):
Thank you!

PR #3892, opened 2021-01-14
Support `include Kconv`

SeekingMeaning:
Marked `Kconv` methods as `module_function`

See https://github.com/ruby/ruby/blob/master/ext/nkf/lib/kconv.rb for original Ruby methods
(Ctrl + F &rarr; `module_function`)

### Motivation

Support `include Kconv`

### Test plan

None


SeekingMeaning:
Merged with #3891

PR #3895, opened 2021-01-15
Define sigs in a stable order.

jvilk-stripe:
WIP: Define sigs in a stable order.

### Motivation

Fixes nondeterminism when two RBI files define conflicting sigs for the same method (the last sig wins). Shopify reported this bug.

### Test plan

I tried to conjure up a flaky test but couldn't.

Tested in Stripe's codebase; no measurable speed regression.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2289,7 +2289,6 @@ public:
 class ResolveSignaturesWalk {
 public:
     struct ResolveSignatureJob {
-        core::FileRef file;
         core::ClassOrModuleRef owner;
         ast::MethodDef *mdef;
         core::LocOffsets loc;
@@ -2304,16 +2303,20 @@ public:
     };
 
     struct ResolveMultiSignatureJob {
-        core::FileRef file;
         core::ClassOrModuleRef owner;
         ast::MethodDef *mdef;
         bool isOverloaded;
         InlinedVector<OverloadedMethodSignature, 2> sigs;
     };
 
+    struct ResolveFileSignatures {
+        core::FileRef file;
+        vector<ResolveSignatureJob> sigs;
+        vector<ResolveMultiSignatureJob> multiSigs;
+    };
+
     struct ResolveSignaturesWalkResult {
-        vector<ResolveSignatureJob> signatureJobs;
-        vector<ResolveMultiSignatureJob> multiSignatureJobs;
+        vector<ResolveFileSignatures> fileSigs;
         vector<ast::ParsedFile> trees;
     };
 
@@ -2695,7 +2698,7 @@ private:
 
                     if (lastSigs.size() == 1) {
                         auto &lastSig = lastSigs.front();
-                        signatureJobs.emplace_back(ResolveSignatureJob{ctx.file, ctx.owner.asClassOrModuleRef(), &mdef,
+                        signatureJobs.emplace_back(ResolveSignatureJob{ctx.owner.asClassOrModuleRef(), &mdef,
                                                                        lastSig->loc,
                                                                        parseSig(ctx, sigOwner, *lastSig, mdef)});
                     } else {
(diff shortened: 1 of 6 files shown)


jvilk-stripe:

→ https://go/builds/bui_IlE4hLc9X6NOSw
→ https://go/builds/bui_IlE4nEILbXmb9T

KaanOzkan:
Thank you for this @jvilk-stripe!

I tested it on my repro repository and the flakiness is gone.

I also tried my hand on a POC flaky test here https://github.com/sorbet/sorbet/commit/993c3e27b8cf9cc2efcc4a0c5c70aa8d764b4b5f . Wanted to share in case it may be of help. It seems to fail consistently on `master` and pass with your changes locally. There's an empty `.rbi` file to trigger non determinism which isn't ideal.

jvilk-stripe:
@KaanOzkan Very nice! thank you. :)

I just tested at Stripe and this change introduces completely negligible overhead. I'll see about adding your test and getting this under review.

jvilk-stripe:
Confirmed that the test fails in master (ran with `--runs_per_test 100` -- no need for the bash loop!). Thanks again!

resolver/resolver.cc
```diff
@@ -2894,23 +2896,28 @@ ast::ParsedFilesOrCancelled Resolver::resolveSigs(core::GlobalState &gs, vector<
...
-                combinedSignatures.emplace_back(move(threadResult.signatureJobs));
-                combinedMultiSignatures.emplace_back(move(threadResult.multiSignatureJobs));
+                combinedFileJobs.insert(combinedFileJobs.end(), make_move_iterator(threadResult.fileSigs.begin()),
+                                        make_move_iterator(threadResult.fileSigs.end()));
             }
         }
     }
 
+    // We need to define sigs in a stable order since, when there are conflicting sigs in multiple RBI files, the last
+    // sig 'wins'.
```
jez, line 2906: One thing I've always wondered was whether we should do some heuristics to determine the "best" location, i.e., maybe we should pick the sig with the fewest counts of `T.untyped` in the sig.

It seems like having this comment here is a good start in case we want to revisit that—we'd have a concrete place to look.

jez (approved):
Thanks!

resolver/resolver.cc
jvilk-stripe: iirc sorting over all sigs in a big project can still get pretty expensive; we sort them in batches by file here to make it cheaper.

I initially looked to see if we could make this process idempotent, but we don't track that we previously registered a sig on a particular method.

Hmm... now that I think of it, we could just maintain a vector that acts as a map from method symbol to sig info to determine if a sig we encounter should "override" what was seen before. So maybe there's a cheap path forward for this. This solution may also let us remove the sorting step entirely.

jez-stripe: even in a world where we'd have some other heuristic, we'd want a consistent way to break ties:

```ruby
sig {params(x: Integer).returns(T.untyped)}
def foo(x); end

sig {params(x: String).returns(T.untyped)}
def foo(x); end
```

paracycle: Some of the flakyness in our codebase was originating from exactly this. People seem to have the assumption that they can "refine" the signature of a method by adding an(other) RBI file. So some of the manual RBI files written like that were racing with the autogenerated DSL RBI files wrt the sigs of some of the methods.

All that to say that it would be great to have a way of pulling this off in a structured way. Breaking ties could still be "last one seen" or something like that. As long as that ordering is documented, we could redirect people to put their "override" signatures at certain locations in the file system.

jvilk-stripe: It's not documented (maybe we should do so), but right now files are ordered by their paths (the sort occurs in `options.cc`) and the "last sig wins".

connorshea: FWIW even with this fix I'm not getting signature documentation if there's more than one signature.

e.g. I have Faker sigs at `sorbet/rbi/gems/faker@2.15.1.rbi` which is generated by Tapioca and has no docs or types. I also have `sorbet/rbi/sorbet-typed/lib/faker/~>2.15/faker.rbi`, and that's generated by Sord so it does have docs and type signatures.

If I hover over `Faker::Date.between(from: 1.month.ago.to_date, to: 1.day.ago.to_date)` I don't get the documentation for it, only the type signatures. I've tried restarting VS Code and some other stuff. I assume if I moved the sorbet-typed directory around that might fix it? For now I've just chosen to delete the Tapioca-generated RBI for now since all the methods are covered by the one in sorbet-typed anyway.

jez: @connorshea that sounds like an unrelated issue.

This pull request was designed to ensure that sorbet returned the same typecheck results no matter in which order arbitrary typechecking threads were scheduled by the operating system. It picked **an** order, and made that order apply all the time. It made no claims to say it was the best order.

Feel free to open another feature request for the documentation on hover.

Issue #3884, opened 2021-01-13
`srb rbi gems` fails with vendored gems


dduugg:
To repro, first clone https://github.com/dduugg/yard-sorbet

Next, confirm normal behavior:

```
bundle install
bundle exec srb rbi gems
```

Confirm that `srb/rbi/gems` exists as expected.

Now, switch to vendored gems:
```
bundle config set --local deployment 'true'
```

And repeat the sequence from above:
```
bundle install
bundle exec srb rbi gems
```

The command now exits unsuccessfully, the last part of the output being:
```
[567/2329] require_relative './vendor/bundle/ruby/2.6.0/gems/rake-13.0.3/lib/rake/rake_test_loader.rb'
File does not exist: /path/to/yard-sorbet/gems
```

(I thought I could outsmart this, as one does with `srb tc`, by appending `--ignore=vendor/`, but that does not seem to be supported here.)

jez:
> I thought I could outsmart this, as one does with srb tc, by appending --ignore=vendor/, but that does not seem to be supported here.

Does that mean you tried appending that to `srb rbi gems`? or did you add that to your `sorbet/config` file?

dduugg:
Ah, thanks for the suggestion suggestion! It seems one can indeed work around this by including the following lines in `sorbet/config`
```
--ignore
vendor/
```

I think there's still a bug in the inconsistency in CLI support of `--ignore`, but feel free to close this one out. (I'd also suggest including the above lines in the default `sorbet/config`.)

jez:
Going to close this in favor of Tapioca

https://sorbet.org/blog/2022/07/27/srb-tapioca

Issue #2985, opened 2020-05-07
srb tc segfault 


will:
#### Input

https://github.com/will/sorbetsegfault

```bash
srb tc
```

#### Observed output

```
/Users/will/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/sorbet-0.5.5585/bin/srb: 
  line 46: 58729 Segmentation fault: 11  "${sorbet}" "${args[@]}"
```

#### Expected behavior

not segfault

- - -

```
/t/sorbetseqf (master)> bundle info sorbet
  * sorbet (0.5.5585)
	Summary: A Typechecker for Ruby
	Homepage: https://sorbet.run
	Path: /Users/will/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/sorbet-0.5.5585
/t/sorbetseqf (master)> ruby -v
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin19]
```

will:
backtrace 
``` gdb
/t/sorbetseqf (master) [139]> lldb /Users/will/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/sorbet-static-0.5.5585-universal-darwin-19/libexec/sorbet
(lldb) target create "/Users/will/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/sorbet-static-0.5.5585-universal-darwin-19/libexec/sorbet"
ruCurrent executable set to '/Users/will/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/sorbet-static-0.5.5585-universal-darwin-19/libexec/sorbet' (x86_64).
(lldb) run
Process 78919 launched: '/Users/will/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/sorbet-static-0.5.5585-universal-darwin-19/libexec/sorbet' (x86_64)
Process 78919 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x8)
    frame #0: 0x000000010054f2d4 sorbet`sorbet::core::File::isPayload() const + 4
sorbet`sorbet::core::File::isPayload:
->  0x10054f2d4 <+4>:  movl   0x8(%rdi), %eax
    0x10054f2d7 <+7>:  decl   %eax
    0x10054f2d9 <+9>:  cmpl   $0x2, %eax
[...]

Issue #2898, opened 2020-04-17
Vendored gems cause typecheck errors, even if not included on command line


francois:
#### Input

```sh
bundle gem power
cd power
# add sorbet / sorbet-runtime to gemspec:
patch <<EOF
diff --git power.gemspec b/power.gemspec
index d1ec85f..10515c5 100644
--- power.gemspec
+++ power.gemspec
@@ -6,17 +6,17 @@ Gem::Specification.new do |spec|
   spec.authors       = ["François Beausoleil"]
   spec.email         = ["francois@teksol.info"]

-  spec.summary       = %q{TODO: Write a short summary, because RubyGems requires one.}
-  spec.description   = %q{TODO: Write a longer description or delete this line.}
-  spec.homepage      = "TODO: Put your gem's website or public repo URL here."
+  spec.summary       = %q{Some summary}
+  spec.description   = %q{Some desc}
+  spec.homepage      = "https://example.com"
   spec.license       = "MIT"
   spec.required_ruby_version = Gem::Requirement.new(">= 2.3.0")

-  spec.metadata["allowed_push_host"] = "TODO: Set to 'http://mygemserver.com'"
+  spec.metadata["allowed_push_host"] = "https://example.com"

   spec.metadata["homepage_uri"] = spec.homepage
-  spec.metadata["source_code_uri"] = "TODO: Put your gem's public repo URL here."
-  spec.metadata["changelog_uri"] = "TODO: Put your gem's CHANGELOG.md URL here."
+  spec.metadata["source_code_uri"] = "https://example.com"
+  spec.metadata["changelog_uri"] = "https://example.com"

   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
@@ -26,4 +26,7 @@ Gem::Specification.new do |spec|
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
+
+  spec.add_runtime_dependency "sorbet-runtime"
+  spec.add_development_dependency "sorbet"
 end
EOF
bundle install --path vendor/gems
bundle exec srb init
bundle exec srb typecheck
bundle exec srb typecheck lib
bundle exec srb typecheck lib/power/version.rb
```

#### Observed output

```
$ bundle exec srb typecheck
[...]

Issue #2110, opened 2019-11-07
srb init hangs


andrewbaxter:
#### Input
On Linux, Ruby 2.6.3p62.

Our Gemfile looks like:
```
source 'https://rubygems.org'
git_source(:github) { |repo| "https://github.com/#{repo}.git" }

ruby '>= 2.5.1'

gem 'sidekiq'

# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
gem 'rails', '~> 5.2.0'
# Use postgresql as the database for Active Record
gem 'pg', '>= 0.18', '< 2.0'
# Use Puma as the app server
gem 'puma', '~> 3.11'
# Use SCSS for stylesheets
gem 'sass-rails', '~> 5.0'
# Use Uglifier as compressor for JavaScript assets
gem 'uglifier', '>= 1.3.0'
# See https://github.com/rails/execjs#readme for more supported runtimes
# gem 'mini_racer', platforms: :ruby

# Use CoffeeScript for .coffee assets and views
gem 'coffee-rails', '~> 4.2'
# Turbolinks makes navigating your web application faster. Read more: https://github.com/turbolinks/turbolinks
gem 'turbolinks', '~> 5'
# Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
gem 'jbuilder', '~> 2.5'
# Use Redis adapter to run Action Cable in production
# gem 'redis', '~> 4.0'
# Use ActiveModel has_secure_password
# gem 'bcrypt', '~> 3.1.7'

# Use ActiveStorage variant
# gem 'mini_magick', '~> 4.8'

# Use Capistrano for deployment
# gem 'capistrano-rails', group: :development

# Reduces boot times through caching; required in config/boot.rb
gem 'bootsnap', '>= 1.1.0', require: false

group :development, :test do
  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
  gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
end

group :development do
  # Access an interactive console on exception pages or by calling 'console' anywhere in the code.
  gem 'web-console', '>= 3.3.0'
  gem 'listen', '>= 3.0.5', '< 3.2'
  # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
  gem 'spring'
  gem 'spring-watcher-listen', '~> 2.0.0'
  # style checking
  gem 'rubocop', '~> 0.57.2', require: false
end

group :test do
[...]

PR #3897, opened 2021-01-15
Ignore vendor/bundle in default config

dduugg:
### Motivation

The `vendor/bundle` directory is used by projects that bundle gems for production, and it's generally assumed the code there won't/shouldn't be touched. Thus it's a surprising cause of sorbet failures, when `srb` fails by attempting to typecheck the `vendor` code:

https://github.com/sorbet/sorbet/issues/3884
https://github.com/sorbet/sorbet/issues/2985#issuecomment-663904639 (note the reacji)
https://github.com/sorbet/sorbet/issues/2898
https://github.com/sorbet/sorbet/issues/2110#issuecomment-553822953
https://github.com/sorbet/sorbet/issues/1744
https://github.com/sorbet/sorbet/issues/975#issuecomment-638324635 (might not specifically be a `vendor` issue)
https://github.com/sorbet/sorbet/issues/945#issuecomment-504374837

I propose that sorbet ignores the `vendor/bundle` directory by default, which I've attempted to implement in this PR.

(Note that I've used `vendor/bundle`, to be consistent with what github [puts](https://github.com/github/gitignore/blob/master/Ruby.gitignore#L43) in the `.gitignore` of Ruby projects by default, but `vendor` is probably fine.)

### Test plan

Updated existing tests. (It might be a good idea to have an e2e test that bundles some gems and does a srb init, but i'd need some hand-holding to figure out how to write it.)

I've also confirmed it works in an [existing](https://github.com/dduugg/yard-sorbet/blob/f219c08/sorbet/config#L7) project.

### DIFF
--- a/gems/sorbet/lib/create-config.rb
+++ b/gems/sorbet/lib/create-config.rb
@@ -24,8 +24,7 @@ class Sorbet::Private::CreateConfig
     File.open(SORBET_CONFIG_FILE, 'w') do |f|
       f.puts('--dir')
       f.puts('.')
-      f.puts('--ignore')
-      f.puts('vendor/bundle')
+      f.puts('--ignore=/vendor/bundle')
     end
   end
 
--- a/gems/sorbet/lib/create-config.rb
+++ b/gems/sorbet/lib/create-config.rb
@@ -24,6 +24,8 @@ class Sorbet::Private::CreateConfig
     File.open(SORBET_CONFIG_FILE, 'w') do |f|
       f.puts('--dir')
       f.puts('.')
+      f.puts('--ignore')
+      f.puts('vendor/bundle')
     end
   end
 
--- a/gems/sorbet/test/snapshot/partial/bad-hash/expected/sorbet/config
+++ b/gems/sorbet/test/snapshot/partial/bad-hash/expected/sorbet/config
@@ -1,4 +1,3 @@
 --dir
 .
---ignore
-vendor/bundle
+--ignore=/vendor/bundle
--- a/gems/sorbet/test/snapshot/partial/bad-hash/expected/sorbet/config
+++ b/gems/sorbet/test/snapshot/partial/bad-hash/expected/sorbet/config
@@ -1,2 +1,4 @@
 --dir
 .
+--ignore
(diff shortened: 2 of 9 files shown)


gems/sorbet/lib/create-config.rb
```diff
@@ -24,6 +24,8 @@ def self.main
     File.open(SORBET_CONFIG_FILE, 'w') do |f|
       f.puts('--dir')
       f.puts('.')
+      f.puts('--ignore')
+      f.puts('vendor/bundle')
```
jez, line 28: This ignores `vendor/bundle` anywhere in the path:

```
--ignore string           Ignores input files that contain the given
                          string in their paths (relative to the input
                          path passed to Sorbet). Strings beginning with /
                          match against the prefix of these relative
                          paths; others are substring matchs. Matches must
                          be against whole folder and file names, so
                          `foo` matches `/foo/bar.rb` and `/bar/foo/baz.rb`
                          but not `/foo.rb` or `/foo2/bar.rb`.
```

I think you want

```suggestion
      f.puts('--ignore=/vendor/bundle')
```

dduugg: 👍  (though it's slightly inconsistent to have `--dir` on separate lines and `--ignore` as a single line)

jez: ok i'm fine any way you prefer

dduugg: :shipit:

jez:
Thanks dug, I'll leave this open for a bit longer to give people a chance to chime in but otherwise I'll merge sometime next week.

wakeless:
We .gitignore basically all RBIs in our codebase and generate them before we run sorbet in CI. This works very well with the tapioca workflow.

I’d recommend a sorbet/ignore file so you aren’t overloading something that is used for something completely different. This perhaps by default it’s vendor/bundle but giving users control would be preferrable.

jez:
> I’d recommend a sorbet/ignore file

This is what `sorbet/config` already is. It specifies which files and directories to include or exclude.

jez (approved):

Issue #3872, opened 2021-01-06
Nullary private detection is too greedy


paracycle:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Foo%0A%20%20def%20self.foo(%26blk)%3B%20end%0A%0A%20%20foo%20do%0A%20%20%20%20private%0A%0A%20%20%20%20def%20bar%3B%20end%0A%20%20end%0A%0A%20%20def%20baz%3B%20end%0Aend%0A%0AFoo.new.baz)

```ruby
# typed: true

class Foo
  def self.foo(&blk); end

  foo do
    private

    def bar; end
  end

  def baz; end
end

Foo.new.baz # Non-private call to private method baz on Foo https://srb.help/7031
```

#### Observed output

```
editor.rb:15: Non-private call to private method baz on Foo https://srb.help/7031
    15 |Foo.new.baz
        ^^^^^^^^^^^
    editor.rb:12: Defined in Foo here
    12 |  def baz; end
          ^^^^^^^
Errors: 1
```

#### Expected behavior

No errors since `baz` is a public method of `Foo`

- - -

The nullary `private` call in the block passed to `Foo.foo` method only modifies the visibility of the methods defined in the block itself. However, Sorbet leaks the visibility change to the scope below it as well. I suspect a similar situation happens with `protected` too.

elliottt:
Thanks for finding this! I think that we need to be pushing/popping entries from the visibility stack for method definitions as well, in the namer

paracycle:
Fixed by #3898

PR #3898, opened 2021-01-15
Push entries to the visbility stack for blocks as well

SeekingMeaning:
This PR adds `preTransformBlock` to the `SymbolFinder` class in `namer.cc` so that entries can be pushed to the visibility stack for blocks.

I feel like the fix shouldn't be this simple, but it seemingly works

### Motivation

https://github.com/sorbet/sorbet/issues/3872

### Test plan

```rb
# typed: true

class Foo
  def self.foo(&blk); end

  foo do
    private

    def bar; end
  end

  def baz; end
end

Foo.new.baz # Non-private call to private method baz on Foo https://srb.help/7031
```

Before:

```
foo.rb:15: Non-private call to private method baz on Foo https://srb.help/7031
    15 |Foo.new.baz # Non-private call to private method baz on Foo https://srb.help/7031
        ^^^^^^^^^^^
    foo.rb:12: Defined in Foo here
    12 |  def baz; end
          ^^^^^^^
Errors: 1
```

After:

```
No errors! Great job.
```

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -452,6 +452,16 @@ public:
         return tree;
     }
 
+    ast::TreePtr preTransformBlock(core::Context ctx, ast::TreePtr block) {
+        methodVisiStack.emplace_back(nullopt);
+        return block;
+    }
+
+    ast::TreePtr postTransformBlock(core::Context ctx, ast::TreePtr block) {
+        methodVisiStack.pop_back();
+        return block;
+    }
+
     ast::TreePtr preTransformMethodDef(core::Context ctx, ast::TreePtr tree) {
         auto &method = ast::cast_tree_nonnull<ast::MethodDef>(tree);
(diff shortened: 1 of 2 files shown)


jez:
> I feel like the fix shouldn't be this simple, but it seemingly works

sometimes when you’ve got the right abstractions, things are surprisingly simple 🙂

jez:
Can you add the test case in your PR description as an automated test? See the README for how to add new tests to the test/testdata folder, and see also tests that have the word `private` in their name. Ideally you’d add this as a new test file instead of modifying an existing test.

Thanks!

elliottt (approved):
Thanks for fixing this one!

elliottt:

→ https://go/builds/bui_IlZXWkMFY6qiou
→ https://go/builds/bui_IlZXJYcXJJ9eQ2

PR #3900, opened 2021-01-15
Don't use a full completion snippet when completing a method with existing args or block

garuma:
Previously when completing a method name that already had an argument list or a block attached, our snippet would do double duty and re-add a whole new set of them.

With this change we first check that the send site we are completing doesn't already have those specified and in that case simply focus on completing the name.

### Motivation

When trying to complete the name of an otherwise fully-formed method call, having another set of parameter or block template added ruins the typing flow and causes frustration.

### Test plan

See included automated tests.

### DIFF
--- a/core/lsp/QueryResponse.h
+++ b/core/lsp/QueryResponse.h
@@ -11,15 +11,16 @@ class TypeConstraint;
 class SendResponse final {
 public:
     SendResponse(core::Loc termLoc, std::shared_ptr<core::DispatchResult> dispatchResult, core::NameRef callerSideName,
-                 bool isPrivateOk, core::MethodRef enclosingMethod, core::Loc receiverLoc)
+                 bool isPrivateOk, core::MethodRef enclosingMethod, core::Loc receiverLoc, u2 totalArgs)
         : dispatchResult(std::move(dispatchResult)), callerSideName(callerSideName), termLoc(termLoc),
-          isPrivateOk(isPrivateOk), enclosingMethod(enclosingMethod), receiverLoc(receiverLoc){};
+          isPrivateOk(isPrivateOk), enclosingMethod(enclosingMethod), receiverLoc(receiverLoc), totalArgs(totalArgs){};
     const std::shared_ptr<core::DispatchResult> dispatchResult;
     const core::NameRef callerSideName;
     const core::Loc termLoc;
     const bool isPrivateOk;
     const core::MethodRef enclosingMethod;
     const core::Loc receiverLoc;
+    const u2 totalArgs;
 
     const std::optional<core::Loc> getMethodNameLoc(const core::GlobalState &gs) const;
 };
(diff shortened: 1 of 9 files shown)


core/Types.h
```diff
@@ -821,6 +821,7 @@ struct DispatchArgs {
 struct DispatchComponent {
     TypePtr receiver;
     MethodRef method;
+    u2 totalArgs;
```
jez, line 824: It's very strange to me to see this data structure changing given the PR description.

Could we instead do this change by changing the type of `lsp::SendResponse`?

A large part of my skepticism is seeing how much of calls.cc (which is not related to LSP or completion) has to change in service of this PR.

garuma: Yeah a lot of the changes are simply there to thread through the `totalArgs` value but I can totally do this with `lsp::SendResponse` instead.

infer/environment.cc
```diff
@@ -1011,7 +1011,8 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                     core::lsp::QueryResponse::pushQueryResponse(
                         ctx, core::lsp::SendResponse(core::Loc(ctx.file, bind.loc), retainedResult, send->fun,
                                                      send->isPrivateOk, ctx.owner.asMethodRef(),
-                                                     core::Loc(ctx.file, send->receiverLoc)));
+                                                     core::Loc(ctx.file, send->receiverLoc),
+                                                     static_cast<u2>(dispatchArgs.args.size())));
```
jez, line 1015: Thanks for refactoring! One of the reasons why I suggested this refactoring is because the number of args actually has nothing to do with the dispatch itself, but is a property of the ast / cfg itself:

```suggestion
                                                     static_cast<u2>(send->args.size())));
```

this refactoring should make that more clear.

jez (approved):
Thanks for the refactor, I think the code is in a much better place.

core/lsp/QueryResponse.h
```diff
@@ -11,15 +11,16 @@ class TypeConstraint;
...
         : dispatchResult(std::move(dispatchResult)), callerSideName(callerSideName), termLoc(termLoc),
-          isPrivateOk(isPrivateOk), enclosingMethod(enclosingMethod), receiverLoc(receiverLoc){};
+          isPrivateOk(isPrivateOk), enclosingMethod(enclosingMethod), receiverLoc(receiverLoc), totalArgs(totalArgs){};
     const std::shared_ptr<core::DispatchResult> dispatchResult;
     const core::NameRef callerSideName;
     const core::Loc termLoc;
     const bool isPrivateOk;
     const core::MethodRef enclosingMethod;
     const core::Loc receiverLoc;
+    const u2 totalArgs;
```
jez, line 23: Why use a `u2` (which requires a static_cast)?

I don't think that this particular structure is performance sensitive w.r.t. allocation size.

garuma: No special reason, I think I was following what some other underlying storage was somewhere else. I will do `size_t` instead to avoid casting.

PR #3902, opened 2021-01-17
Unify TypeAndOrigins-based error messages

jez:
### Motivation

Sorbet tracks a wealth of location information, which is one of the main reasons
why people consider Sorbet to have good error messages.

But individual error messages make inconsistent usage of this information. In
particular, pretty much every error message that shows a type which is
influenced by control-flow sensitive typing should show the TypeAndOrigins data
structure on hand explaining how Sorbet arrived at that type.

This PR adds a couple helpers to make that process **super** easy (one short
line of code) so that people are highly encouraged to print such information
when adding new, similar kinds of errors.

Some of these errors look like trivial changes ("it's just duplicating information
present in the error") but in fact all of the `Got ...` sections will remember the
control flow path through a method resulting in Sorbet knowing a specific type.

### Commit summary

There are a lot of individual changes in this PR, but I've (almost excessively)
broken it up into small commits.

Each commit does only one thing.

### Test plan

**Before**

Screen Shot 2021-01-16 at 10 34 26 PM

**After**

Screen Shot 2021-01-16 at 10 35 09 PM

Note:

- consistent error messages
- more colors
- more errors showing TypeAndOrigins information

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -767,6 +767,7 @@ public:
 
     std::vector<ErrorLine> origins2Explanations(const GlobalState &gs, Loc originForUninitialized) const;
 
+    static ErrorSection explainExpected(const GlobalState &gs, TypePtr type, Loc origin, const std::string &for_);
     ErrorSection explainGot(const GlobalState &gs, Loc originForUninitialized) const;
 
     ~TypeAndOrigins() noexcept;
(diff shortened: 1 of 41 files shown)


jez:

→ https://go/builds/bui_ImNHWsnXSeNtBC
→ https://go/builds/bui_ImNHzRgUUxQ5fp

elliottt (approved):
Nice! Those helpers are great, and the consistency in the error messages is really helpful 👍

Issue #3769, opened 2020-12-08
Calling `nil?` on non-nilable value in a conditional doesn't report dead code


DanielGilchrist:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Adef%20main%0A%20%20if%205.nil%3F%0A%20%20%20%20puts%20%22This%20should%20error%20as%20unreachable%22%0A%20%20end%0A%0A%20%20unless%205%0A%20%20%20%20puts%20%22Errors%20as%20unreachable%20as%20expected%22%0A%20%20end%0Aend)

```ruby
# typed: true

def main
  if 5.nil?
    puts "This should error as unreachable"
  end

  unless 5
    puts "Errors as unreachable as expected"
  end
end
```

#### Observed output

```
editor.rb:9: This code is unreachable https://srb.help/7006
     9 |    puts "Errors as unreachable as expected"
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior
I would expect Sorbet to error stating that `puts "This should error as unreachable"` is unreachable (error 7006)

jez:
This is a dupe of #621

Issue #2867, opened 2020-04-09
Missing unreachable code error with type comparisons that are always false


amurray-stripe:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Afoo%20%3D%201%0Aif%20!foo.is_a%3F(Numeric)%0A%20%20%23%20This%20code%20is%20unreachable%20but%20there's%20no%20error%0A%20%20puts%20'Must%20be%20Numeric'%0Aend%0A%0A%23%20uncommenting%20this%20line%20causes%20the%20expected%20error%20to%20appear%0A%23%20foo)

```ruby
# typed: true
extend T::Sig

foo = 1
if !foo.is_a?(Numeric)
  # This code is unreachable but there's no error
  puts 'Must be Numeric'
end

# uncommenting this line causes the expected error to appear
# foo
```

It seems a reference the variable (`foo`) must occur after the unreachable code in order for Sorbet to detect the error. The bug occurs inside methods too.

#### Automated test
I created a failing test for you: https://github.com/amurray-stripe/sorbet/commit/d1701d9bb5f0fa5efd41f13461809e7cc3d11201

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

```
editor.rb:7: This code is unreachable https://srb.help/7006
     7 |  puts 'Must be Numeric'
               ^^^^^^^^^^^^^^^^^
Errors: 1
```

jez:
SEO dead code

jez:
See also #3769

Issue #1747, opened 2019-09-10
Refactor updateKnowledge to be based on intrinsics, instead of name-based


jez:
### Problem

Sorbet updates it's flow-sensitive typing knowledge based strictly on the name of the method (!). This is alarming, but has worked well enough in practice because there are only a handful of them. They're defined here:

https://github.com/sorbet/sorbet/blob/334ea1e185168914acfb8b8f7632bdc3f1ad0571/infer/environment.cc#L340

(Things like `==` and `is_a?`.)

A better alternative to this would be to tap into the way that method intrinsics work (where there's a registry of intrinsics, stored on Symbols) and make updating knowledge be done by those methods.

This is mostly a question of code hygiene, but it does cause Sorbet to be wrong:

#### Input

```ruby
# typed: true
class A
  extend T::Sig

  sig {params(other: T.untyped).returns(TrueClass)}
  def is_a?(other)
    true
  end
end
class B; end

a = A.new
if a.is_a?(B)
  puts a # This code IS reachable!
end
```
[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams(other%3A%20T.untyped).returns(TrueClass)%7D%0A%20%20def%20is_a%3F(other)%0A%20%20%20%20true%0A%20%20end%0Aend%0Aclass%20B%3B%20end%0A%0Aa%20%3D%20A.new%0Aif%20a.is_a%3F(B)%0A%20%20puts%20a%20%23%20This%20code%20IS%20reachable!%0Aend)

#### Output:

```
editor.rb:14: This code is unreachable https://srb.help/7006
    14 |  puts a # This code IS reachable!
          ^^^^^^
Errors: 1
```

izquierdo:
Hi, I'm considering working on this issue and wanted to check in case someone else already is.

I also would like some guidance regarding the approach you were thinking of.

One idea is to add an `updateKnowledge` function to `IntrinsicMethod` and call it during dispatch when an intrinsic is seen (maybe after `apply` calls like [this one](https://github.com/sorbet/sorbet/blob/5753872590c7d982f5e12f7f967d4d2702c1b6fb/core/types/calls.cc#L1047)) but this requires `Symbols.h` to know about CFGs and inference-related stuff like `TestedKnowledge`, which seems to be outside its domain.

[...]

PR #3903, opened 2021-01-17
[wip] Fix some instances of undetected unreachable code

izquierdo:
(will fill in explanation shortly)

### Motivation

Fixes https://github.com/sorbet/sorbet/issues/2867
Fixes https://github.com/sorbet/sorbet/issues/3769

### Test plan

See included automated tests.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -505,6 +505,8 @@ void Environment::updateKnowledge(core::Context ctx, cfg::LocalRef local, core::
             auto mayBeOverridden = s.symbol.data(ctx.state)->findMemberTransitive(ctx.state, send->fun);
             auto objectBang = core::Symbols::Object().data(ctx.state)->findMemberTransitive(ctx.state, send->fun);
 
+            ENFORCE(mayBeOverridden.exists());
+            ENFORCE(objectBang.exists());
             if (mayBeOverridden != objectBang) {
                 return;
             }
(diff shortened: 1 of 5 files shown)


jez:

→ https://go/builds/bui_ImNE765RalkSVg
→ https://go/builds/bui_ImNErFyQ7aVeyu

jez:
Hey thanks for the change!

I haven't convinced myself how the change works yet. I will take another look this week.

Just a heads up: this change causes almost a thousand new type errors on Stripe's codebase, so it will be hard for us to accept it because we'll have to fix those new errors before we can merge this change.

izquierdo:
Thanks for taking the time to look at this!

I understand re: errors on Stripe's codebase. No worries.

jez:
> * test/testdata/cfg/override_bang.rb

Would you mind actually not changing updateKnowledge in this PR, and instead just recording the new behavior on the override_bang.rb test?

I think we'd prefer a solution like #1747 to handle that whole class of problems from first principles, rather than solving the case for the `!` intrinsic in an ad hoc way (also: findMemberTransitive is slow, and the approach in #1747 is fast)

izquierdo:
Makes sense. Updated the PR, but had to make more changes to `override_bang.rb` to get it to pass. It might make sense to wait until another solution (like #1747) lands before taking this up again.

PR #3905, opened 2021-01-18
Fix and improve error for failed SolveConstraint

jez:
### Motivation

- This error wasn't exercised by our codebase at all before.
- Unsurprisingly, it was buggy—it caused Sorbet to crash if it was ever
  reported.
- Even after just fixing the crash, it was a useless error message, so I added
  more information to it.

### Test plan

Review by commit to see how the test cases change as a result of each commit.

**Before**

Screen Shot 2021-01-17 at 4 09 28 PM

**After**

Screen Shot 2021-01-17 at 4 10 08 PM


elliottt (approved):
Thanks for the explanation in the test case!

PR #3912, opened 2021-01-19
TreePtr -> ExpressionPtr

jez:
### Motivation

We have multiple kinds of "tree"'s in Sorbet (parser::Node, core::Type, etc.).

`TreePtr` used to be called `unique_ptr<ast::Expression>` and I think we should
recover that naming.

### Test plan

**Review by commit** for a good summary of the changes.

### DIFF
--- a/ast/ArgParsing.cc
+++ b/ast/ArgParsing.cc
@@ -8,7 +8,7 @@ using namespace std;
 namespace sorbet::ast {
 
 namespace {
-ParsedArg parseArg(const ast::TreePtr &arg) {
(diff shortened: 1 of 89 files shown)


elliottt (approved):
Seems reasonable to me. Not blocking, but what do you think about `ExprPtr` instead?

jez:
I’d like the class name, the macro for defining the class, and the Ptr class to all be the same.

I generally prefer abbreviations in variable names not class names, but I prefer ExprPtr over TreePtr so if you have a strong preference for ExprPtr I’d take that instead.

PR #3896, opened 2021-01-15
Detect attr_reader-style methods

jez:
### Motivation

This records more information for later parts of the pipeline.

It is currently unused by Sorbet.

### Commit summary

I recommend reviewing by commit.
The diff looks huge, but that's just because there are a ton of super mechanical exp file changes.

- **Add failing test** (8aa12eb88)

- **Detect attr_reader-style methods** (9b46f1023)

- **Fix failing test** (14f250d41)

- **tools/scripts/update_testdata_exp.sh** (61fc7c745)

- **fix a stragler** (1e2e34eaf)

- **Add a test using self method** (a070e3bf5)

- **Remove spurious T::Struct** (a155a25fb)

### Test plan

See included automated tests.

PR #3913, opened 2021-01-20
Only record attr_reader methods that have no sig checking

jez:
### Motivation

The approach in #3896 was too eager.

### Test plan

Review by commit to see how the tests change.


rewriter/AttrReader.cc
```diff
@@ -65,7 +65,7 @@ bool isTNilableOrUntyped(const ast::TreePtr &expr) {
            isT(send->recv);
 }
 
-ast::Send *findSendReturns(core::MutableContext ctx, ast::Send *sharedSig) {
+ast::Send *findSendReturns(ast::Send *sharedSig) {
```
elliottt, line 68: Thanks for removing the unused `ctx` parameter!

rewriter/AttrReader.cc
```diff
@@ -119,18 +119,61 @@ void ensureSafeSig(core::MutableContext ctx, const core::NameRef attrFun, ast::S
...
+    }
+
+    auto lit = ast::cast_tree<ast::Literal>(checked->args[0]);
+    if (lit == nullptr || !lit->isSymbol(ctx)) {
+        // Unknown: default to false
+        return false;
+    }
+
+    // Treats `.checked(:tests)` as unknown, therefore not unchecked.
+    return lit->asSymbol(ctx) == core::Names::never();
```
elliottt, line 163: 👍

elliottt (approved):

rewriter/AttrReader.cc
```diff
@@ -119,18 +119,61 @@ void ensureSafeSig(core::MutableContext ctx, const core::NameRef attrFun, ast::S
...
+// Heuristic to check if user-provided sig adds no runtime checking on top of an attr_reader.
+//
+// A user-provided sig causes an attr_reader method to behave differently from a normal attr_reader
+// method at runtime.
+//
+// If the answer would be "maybe adds runtime checking, but hard to tell," we answer that it does add
+// checking to be safe. Thus, the naming of this method is important: we can't rename this method to
+// `sigIsChecked` and negate all true/false. (Put another way: double negation elimination doesn't
+// apply here).
+bool sigIsUnchecked(core::MutableContext ctx, ast::Send *sig) {
```
froydnj, line 144: I like this comment, but perhaps we could have called it `sigCouldBeChecked`?  Then concerns about double negation go away.

PR #3916, opened 2021-01-21
Run --store-state with multiple threads

jez:
### Motivation

There are about 150 RBI files in Sorbet's payload.
Right now it processes them entirely serially.
This PR allows processing them with max hardware parallelism.

This leads to a slight (but not mindblowing speedup) on my machine for the
`generate-state-payload-raw` step.

This only affects builds that run with `--store-state`, which is what we pass to
`sorbet-orig` to generate the binary payload data in `payload/binary/BUILD`.

### Test plan

Ran a bunch of rebuilds where I changed nothing but adding whitespace to
`rbi/core/kernel.rbi` and ran `bazel build --config=dbg //main:sorbet`.

Three such rebuilds on an already-cached master: ~17s start to finish

Three such rebuilds after building once to cache this branch: ~12s start to finish.


elliottt (approved):
Nice!

PR #3917, opened 2021-01-21
Add --suggest-unsafe option

jez:
### Motivation

This is an option to make mass codemods easier.

Frequently, upgrading to a new Sorbet version is hard because either Sorbet
fixed a bug which was hiding untypedness, or Sorbet added type annotations to
it's standard library RBI definitions.

In almost all of these cases, it's better to burn in a few calls to `T.unsafe`
and upgrade, so that at least new code gains the advantages of the Sorbet
upgrade. In the meanwhile, people can go back through the newly-added calls to
`T.unsafe` manually to find and fix the sources of the errors. Frequently
people do this out of curiosity when they see a `T.unsafe` in the file they
have open.

This functionality is all behind an option that must be manually passed on the
command line. We might want to eventually write a Sorbet doc page like "tips
for doing codemods" that suggests how to use this option. I **do not** think
the default should be to show these `T.unsafe` autocorrects during normal
Sorbet usage, because I don't want to encourage people to sprinkle `T.unsafe`
to fix all their Sorbet problems.

I expect this to be the first of many such PRs. I have purposely kept this PR
small and only added two new autocorrects. Partly, this was because it was easy
to cargo cult the existing `T.must` autocorrects. But in general, I think that
there are **many** errors in Sorbet that can be fixed with a particular usage
of `T.unsafe`, and over time we should introduce these into as many places as
possible (where it's not possible to add another, better autocorrect).

### Commit summary

I did some minor test cleanups while I was looking at this code.

- **Add --suggest-unsafe option, thread into GlobalState** (bd131b0f0)

- **Hijack some T.must autocorrects** (230617894)

- **Reword the helpers test because I keep copying it** (9660c5294)

- **Make the T.must cli test better** (e85a602cf)

- **Add new test** (03d99d3c0)

### Test plan

I have included a CLI test showing two examples.

### DIFF
--- a/test/cli/suggest_t_unsafe/suggest_t_unsafe.sh
+++ b/test/cli/suggest_t_unsafe/suggest_t_unsafe.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+tmp="$(mktemp -d)"
+infile="test/cli/suggest_t_unsafe/suggest_t_unsafe.rb"
+cp "$infile" "$tmp"
+
+cwd="$(pwd)"
+cd "$tmp" || exit 1
+
+if "$cwd/main/sorbet" --silence-dev-message --suggest-unsafe -a suggest_t_unsafe.rb 2>&1; then
+  echo "Expected to fail!"
+  exit 1
+fi
+
+echo
+echo --------------------------------------------------------------------------
+echo
+
+# Also cat the file, to make that the autocorrect applied
+cat suggest_t_unsafe.rb
+
+rm suggest_t_unsafe.rb
+rm "$tmp"
--- a/test/cli/suggest_t_unsafe/suggest_t_unsafe.sh
+++ b/test/cli/suggest_t_unsafe/suggest_t_unsafe.sh
@@ -1,23 +1,50 @@
 #!/usr/bin/env bash
 
-tmp="$(mktemp -d)"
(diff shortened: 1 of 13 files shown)


test/cli/suggest_t_unsafe/suggest_t_unsafe.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+tmp="$(mktemp -d)"
+infile="test/cli/suggest_t_unsafe/suggest_t_unsafe.rb"
+cp "$infile" "$tmp"
+
+cwd="$(pwd)"
+cd "$tmp" || exit 1
+
+if "$cwd/main/sorbet" --silence-dev-message --suggest-unsafe -a suggest_t_unsafe.rb 2>&1; then
```
elliottt, line 10: Could you add a test for a non-default value of `--suggest-unsafe` as well?

elliottt (approved):
Looks good, only one suggestion about adding a test case for `--suggest-unsafe` with a non-default value.

test/cli/suggest_t_unsafe/suggest_t_unsafe.sh
jez: Ah i knew you were gonna ask for this i was trying to sneak it past you 😅

jez: 1fb89b2

PR #3921, opened 2021-01-22
Fix autocorrects for packager names

jez:
### Motivation

LSP-based quickfixes and command-line-based autocorrects are broken when Sorbet
would have otherwise been able to suggest a "Did you mean?" autocorrect for a
bad constant name.

See the test, but tl;dr is that we put the `PkgRoot` in the name when we
shouldn't.

### Test plan

See included automated tests.

### Commit summary

There are unfortunately a bunch of random changes here, because I kept
getting bitten by random bugs as I was fixing the bug I wanted to fix.

- **Rename a test for conformity** (1b4aedf75)

  If the file isn't literally called `pass.package-tree.exp`, it doesn't
  get picked up by `update_testdata_exp.sh`

  I'm too lazy to figure out how to make the test itself error if you
  misname such a file.

- **Use UniqueNameKind for packager names** (5402ee5a0)

- **Make some conditionals exhaustively checked** (b1b1f3d7e)

  I had to find all of these ENFORCEs at runtime when adding a new
  constant kind. I would have liked to have the compiler tell me all the
  places I needed to change upfront.

- **tools/scripts/update_testdata_exp.sh** (62ee19399)

- **Add simple-package CLI test** (aa2012a7e)

- **Show how it fails to do autocorrects** (70cb8e6be)

- **Fix failing autocorrect test** (5a6730385)

- **Fix other tests** (54cff0970)

  In particular, I think that all error messages for packager names are a
  garbage fire right now. We should be suggesting that you might want to
  import the other package, or you might have meant to use this similar
  thing in another imported package, etc.

  So while this error message changes, I don't even think it's worth
  considering whether it's "better" or "worse" because they are all bad
  errors and they will all have to be fixed before a mass rollout of
  packages.


jez:
cc @aisamanra @ngroman @jvilk-stripe

core/Symbols.cc
```diff
@@ -952,14 +952,18 @@ string Symbol::show(const GlobalState &gs) const {
 
     if (this->owner == core::Symbols::PackageRegistry()) {
         // Pretty print package name (only happens when `--stripe-packages` is enabled)
-        auto nameStr = this->name.show(gs);
-        constexpr string_view packageNameSuffix = "_Package"sv;
-        if (absl::EndsWith(nameStr, packageNameSuffix)) {
```
jez, line 957: Sorbet is and never has been in the business of doing name-based decisions based on string comparisons, and I don't see why that should stop being the case for packager names.

As such I've redone the support for packager names in this PR to use a new unique name kind, so that asking whether a name is a packager name is a matter of checking a few bits.

main/autogen/autoloader.cc
```diff
@@ -421,7 +421,7 @@ void AutoloadWriter::write(const core::GlobalState &gs, const AutoloaderConfig &
                 ENFORCE(child->qname.package);
                 auto namespaceName = child->qname.package->show(gs);
                 // this package name will include the suffix _Package on the end, and we want to remove that
-                constexpr int suffixLen = "_Package"sv.length();
+                constexpr int suffixLen = char_traits<char>::length("_Package");
```
jez, line 424: If you check the source code, this is exactly what the string_view constructor will call to populate it's internal `length` field.

core/Symbols.cc
jvilk-stripe: Good idea. I didn't feel emboldened to do this sort of thing when I first added packager support to Sorbet.

test/cli/simple-package/pass.package-tree.exp
```diff
@@ -0,0 +1,130 @@
+class <emptyTree><<C <root>>> < (::<todo sym>)
```
jvilk-stripe, line 1: What is this file for? `*.exp` files are only tested on testdata tests.

jvilk-stripe (approved):
Thanks for fixing this bug! Our users will be very happy (be sure to let our VS Code channel know about this once you bump Sorbet at Stripe).

test/cli/simple-package/pass.package-tree.exp
jez: Whoops. I `cp -r`'d this test from a testdata test and left all the files it came with. I'll delete this file.

jez:

→ https://go/builds/bui_Ipg2NhWzXkoir9
→ https://go/builds/bui_Ipg2WlPtNNi0Ug

PR #3923, opened 2021-01-23
Remove references to Opus in favor of configuration callbacks

aisamanra:
### Motivation
We use `defined?(Opus)` in a few places in sorbet-runtime, which (especially with forthcoming package work) is brittle and also just bad style. Replace these with callbacks provided in `T::Configuration` so that we won't have `Opus` names hard-coded in the runtime any more.

Note that we're going to have to define these callbacks when we roll this out, but it should be pretty simple to do so.

### Test plan

See included automated tests. Added tests for each of the added configuration options.

### DIFF
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -412,6 +412,19 @@ module T::Configuration
     @module_name_mangler = handler
   end
 
+  # Set to a PII handler function. This will be called with the `sensitivity:`
+  # annotations on things that use `T::Props` and can modify them ahead-of-time.
+  #
+  # @param [Lambda, Proc, nil] value Proc that takes a hash mapping symbols to the
+  # prop values. Pass nil to avoid changing `sensitivity:` annotations.
+  def self.normalize_sensitivity_and_pii_handler=(handler)
+    @sensitivity_and_pii_handler = handler
+  end
+
+  def self.normalize_sensitivity_and_pii_handler
+    @sensitivity_and_pii_handler
+  end
+
   # Temporarily disable ruby warnings while executing the given block. This is
   # useful when doing something that would normally cause a warning to be
   # emitted in Ruby verbose mode ($VERBOSE = true).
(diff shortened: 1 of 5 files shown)


paracycle (commented):
I think this PR should also change the check for required sensitivity option based on prop name, to also be based on if the new normalize handler is set.

gems/sorbet-runtime/lib/types/configuration.rb
```diff
@@ -412,6 +412,32 @@ def self.module_name_mangler=(handler)
...
+  # @param [Lambda, Proc, nil] value Proc that takes a class and
+  # produces its owner, or `nil` if it does not have one.
+  def self.class_owner_finder=(handler)
+    @class_owner_finder = handler
+  end
+
+  def self.class_owner_finder
+    @class_owner_finder
+  end
+
```
jez, line 440: You'll need to add these in the RBI to be able to use them in Stripe's codebase.

gems/sorbet-runtime/test/types/props/decorator.rb
```diff
@@ -362,4 +384,26 @@ class MatrixStruct
...
   end
+
+  it 'applies the supplied sensitivity and PII handler' do
+    begin
+      T::Configuration.normalize_sensitivity_and_pii_handler = ->(meta) do
+        meta[:pii] = :set
+        meta[:sensitivity] += 1
+        meta
+      end
+      e = Class.new(T::Struct) do
```
jez, line 395: Can we define this class at the top of the file, instead of in a `Class.new` hook?

jez (approved):

aisamanra:

→ https://go/builds/bui_IpFHYtFSqWc4DJ
→ https://go/builds/bui_IpFHaMsc3HUKpD

aisamanra:
I do think at some point I want to come back and rethink how we handle Stripe-specific props in `T::Props` generally, 'cause we've got a lot of weird special-purpose stuff and I think we can go back and reconsider how we implement that. But for now I'm going to punt on some of that, since this is a step up from the old version.

Issue #3922, opened 2021-01-22
Typechecking crashes upon reassignment of `self` inside a loop


KaanOzkan:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aa%20%3D%20self%0A%5B1%2C%202%5D.each%20do%20%7Ci%7C%0A%20%20a%20%3D%2010%0Aend)

```ruby
# typed: true
a = self
[1, 2].each do |i|
  a = 10
end
```

#### Observed output

```
Backtrace:
__cxa_throw (in sorbet) + 113
std::__1::__throw_out_of_range(char const*) (in sorbet) + 56
sorbet::core::Loc::source(sorbet::core::GlobalState const&) const (in sorbet) (Loc.cc:271)
sorbet::infer::Environment::processBinding(sorbet::core::Context, sorbet::cfg::CFG const&, sorbet::cfg::Binding&, int, int, sorbet::infer::KnowledgeFilter&, sorbet::core::TypeConstraint&, sorbet::core::TypePtr&) (in sorbet) (environment.cc:1389)
sorbet::infer::Inference::run(sorbet::core::Context, std::__1::unique_ptr<sorbet::cfg::CFG, std::__1::default_delete<sorbet::cfg::CFG> >) (in sorbet) (inference.cc:197)
sorbet::realmain::pipeline::CFGCollectorAndTyper::preTransformMethodDef(sorbet::core::Context, sorbet::ast::TreePtr) (in sorbet) (pipeline.cc:70)
sorbet::ast::TreeMapper<sorbet::realmain::pipeline::CFGCollectorAndTyper, sorbet::core::Context>::mapMethodDef(sorbet::ast::TreePtr, sorbet::core::Context) (in sorbet) (treemap.h:180)
sorbet::ast::TreeMapper<sorbet::realmain::pipeline::CFGCollectorAndTyper, sorbet::core::Context>::mapIt(sorbet::ast::TreePtr, sorbet::core::Context) (in sorbet) (treemap.h:491)
sorbet::ast::TreeMapper<sorbet::realmain::pipeline::CFGCollectorAndTyper, sorbet::core::Context>::mapClassDef(sorbet::ast::TreePtr, sorbet::core::Context) (in sorbet) (treemap.h:169)
[...]

PR #3924, opened 2021-01-23
Fix crash in pinning autocorrect

jez:
Fix the error reported in #3922, which was caused by the changes in #3876.

### Motivation

In defense of #3876, the old behavior was this:

Screen Shot 2021-01-23 at 1 16 31 PM

because we were using the Loc of the whole method definition as the origin of `self`.

I'd argue that it's better to say "I don't have a Loc" than it is to say "the Loc is the entire method body" because then we can at least detect when we shouldn't be suggesting an autocorrect, rather than suggesting a terrible autocorrect.

### Test plan

See included automated tests.

### DIFF
--- a/core/Loc.cc
+++ b/core/Loc.cc
@@ -267,6 +267,7 @@ string Loc::filePosToString(const GlobalState &gs, bool showFull) const {
 }
 
 string Loc::source(const GlobalState &gs) const {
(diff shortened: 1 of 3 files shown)


paracycle (approved):
Thanks for the fix. There was nothing wrong with the changes of the previous PR, except for the missing existence check added in this PR.

elliottt (approved):

PR #3926, opened 2021-01-24
Add some more debugging information

jez:
### Motivation

These functions are handy to have in a debugger.

### Test plan

See included automated tests.

### DIFF
--- a/core/Files.h
+++ b/core/Files.h
@@ -19,6 +19,7 @@ public:
     FileRef() : _id(0){};
     FileRef(unsigned int id);
 
+    FileRef(FileRef &f) = default;
     FileRef(const FileRef &f) = default;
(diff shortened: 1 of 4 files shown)


elliottt (approved):

core/Files.h
```diff
@@ -19,6 +19,7 @@ class FileRef final {
     FileRef() : _id(0){};
     FileRef(unsigned int id);
 
+    FileRef(FileRef &f) = default;
```
froydnj, line 22: Why is this needed?  Does something in the debugger not do proper resolution to select the `const &` overload?

jez: Yeah, I kept getting warnings in lldb until I added this.

PR #3929, opened 2021-01-24
Vastly improve dead code error message

jez:
### Motivation

We have #3903 which is a great PR fixing a long standing bug, but it introduces
almost a thousand new errors on Stripe's codebase.

On a more modest scale, my experiments to add better Shape typing are
introducing a bunch of dead code errors as well.

And last, the error for "This error is unreachable" is about as barren as an
error can get.

I fix all of these problems in this PR.

### Test plan

**Before**

Screen Shot 2021-01-23 at 11 45 17 PM

**After**

Screen Shot 2021-01-24 at 12 10 53 AM

### Commit summary

Review by commit to see how the new test evolves.

- **pre-work: Fix bug when combining two basic blocks** (3f2f5ed22)

  We used to say that the `bexit.loc` of the two combined `BasicBlock`'s
  was the loc of the `BasicBlock` that unconditionally jumped into the
  second block, instead of loc of the second `BasicBlock` jumping out
  somewhere else.

  We don't actually have anywhere in the codebase that shows this Loc in
  an error message until this change.

- **Add failing test** (79bc84bc6)

- **Add ErrorLine on unconditional condition** (b7b4d1276)

- **Explain where type came from** (12f2d72d2)

  We have this fancy helper for explaining things, so I figured we may as
  well use it.

  It's also super handy that we keep all the environments around even
  after we're done typechecking a certain block, because technically all
  the conditions' variables that I want to be checking are only in scope
  in the previous environment.

  Also I had to change `getTypeAndOrigin` from private to public because I
  didn't want to use `getAndFillTypeAndOrigin` (public), which mutates the
  Environment and the VariableUseSite you give it.

- **Add autocorrect in --suggest-unsafe mode** (d98144a4c)

  This only applies in --suggest-unsafe mode because I figure that's the
  common case for dead code errors. It's usually "some codemod made
  something untyped into something typed" which causes a bunch of deadcode
  errors, because the revealed typing information wasn't actually good
  enough.

- **Skip dead backedges** (98c246cf8)

  This change makes sense intuitively, and I was definitely seeing cases
  that this code fixed. I could have sworn that I had a test case for it,
  but I can't find one anymore.

  Anyways, the intuition is in the comment. I figured it was harmless
  enough to include even though the test case it was needed to fix has
  been lost to time.

### DIFF
--- a/infer/environment.h
+++ b/infer/environment.h
@@ -171,10 +171,6 @@ class Environment {
 
     bool getKnownTruthy(cfg::LocalRef var) const;
 
-    // NB: you can't call this function on vars in the first basic block since
-    // their type will be nullptr
-    const core::TypeAndOrigins &getTypeAndOrigin(core::Context ctx, cfg::LocalRef symbol) const;
-
     /* propagate knowledge on `to = from` */
     void propagateKnowledge(core::Context ctx, cfg::LocalRef to, cfg::LocalRef from, KnowledgeFilter &knowledgeFilter);
 
@@ -213,6 +209,10 @@ public:
     void setUninitializedVarsToNil(const core::Context &ctx, core::Loc origin);
 
     std::string toString(const core::GlobalState &gs, const cfg::CFG &cfg) const;
+
+    // NB: you can't call this function on vars in the first basic block since
+    // their type will be nullptr
+    const core::TypeAndOrigins &getTypeAndOrigin(core::Context ctx, cfg::LocalRef symbol) const;
 
     const core::TypeAndOrigins &getAndFillTypeAndOrigin(core::Context ctx, cfg::VariableUseSite &symbol) const;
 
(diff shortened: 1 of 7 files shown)


infer/environment.h
```diff
@@ -214,6 +210,10 @@ class Environment {
 
     std::string toString(const core::GlobalState &gs, const cfg::CFG &cfg) const;
 
+    // NB: you can't call this function on vars in the first basic block since
+    // their type will be nullptr
+    const core::TypeAndOrigins &getTypeAndOrigin(core::Context ctx, cfg::LocalRef symbol) const;
```
elliottt, line 215: Was this private before because of the comment?

infer/inference.cc
```diff
@@ -181,6 +181,38 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
...
+                                // the error message.
+                                continue;
+                            }
+
+                            const auto &cond = prevBasicBlock->bexit.cond;
+                            if (cond.type == nullptr) {
+                                // This previous block is actually a future block we haven't processed yet.
+                                // (Remember: our inference pass is an approximate forwards toposort
+                                // of a graph that can have cycles). It can't have been a block that
+                                // caused the current error.
```
elliottt, line 199: This is a really helpful comment!

elliottt (approved):
The additional information on error messages is so helpful!

infer/environment.h
jez: Honestly, not sure.

PR #3931, opened 2021-01-25
Convert some `# error:` tests to `# ^^^ error:` tests

jez:
### Motivation

I started today thinking that I would make the location information better for
keyword args errors.

I ended up realizing that I misunderstood how calls.cc handles keyword args,
and got discouraged after realizing that the problem wasn't quite as easy as I
thought it was.

In any case, I took the time to refactor these tests so that they capture the
inline location information, so that we'll know of any future changes to those
locations (intentional or unintentional).

### Test plan

no-op test-only change


elliottt (approved):
Thanks for improving these!

PR #3933, opened 2021-01-25
Delete configatron pass.

jvilk-stripe:
Dmitry previously introduced the configatron pass to type codepaths at Stripe that use configatron. However, we no longer use the configatron gem, so these code paths were doing meaningless work. Furthermore, I have traced a crash to the types generated by this pass, so it is actively causing harm at Stripe.

Should we ever desire to have similar functionality in the future, it would probably be better to take a code generation approach (e.g., generate rbis).

### Motivation

See above.

### Test plan

CI passes


jez:
> generate rbis

or even generate the code that runs itself, like we do for feature flags.

jez (approved):
:fire:

PR #3934, opened 2021-01-25
externalType: Never return nullptr.

jvilk-stripe:
externalType: Never return nullptr.

Proactively prevent crashes in prod builds due to bugs.

At the same time, emit a metric when this occurs so we can track when this is happening.

### Motivation

We've been observing many Sorbet crashes in IDE at Stripe. I isolated one (after great effort!) to externalType returning `nullptr`, and fixed it by disabling the configatron pass. However, I can't be certain that I've found all of these cases.

### Test plan

If we had a test that triggered this code path, then we would have a bug to fix.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -72,6 +72,12 @@ TypePtr Symbol::selfType(const GlobalState &gs) const {
 
 TypePtr Symbol::externalType() const {
     ENFORCE_NO_TIMER(resultType);
+    if (!resultType) {
+        // Don't return nullptr in prod builds, which would cause a disruptive crash
+        // Emit a metric and return untyped instead.
+        prodCounterInc("symbol.externalType.nullptr");
+        return Types::untypedUntracked();
+    }
     return resultType;
 }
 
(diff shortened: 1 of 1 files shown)


jez (approved):

jez:
> Proactively prevent crashes in prod builds due to bugs.

Where do you stand on using `Exception::raise` for things like this?

There was a bug that KaanOzkan reported ( #3922 ) which boils down to calling `Loc::source` on a Loc that doesn't exist.

I was debating whether to `ENFORCE(loc.exists())` or `Exception::raise`.

jvilk-stripe:
> Where do you stand on using Exception::raise for things like this?

Do you mean in production builds or debug builds? Is the idea to get prod builds to produce useful crash info?

The end result would be similar to the current situation I think -- Sorbet will crash, restart, and be unresponsive while it boots up. I do have crash info from prod builds, but it's of limited utility since so many stack frames are omitted. Perhaps it would be slightly better crash info w/ exception::raise, but even so I'm concerned about the user experience right now with the crashes in place.

So I'm -1 to the idea for this particular bug.

PR #3936, opened 2021-01-25
Fix the signature for BigDecimal

che-burashco:
Change `truncate` to return `BigDecimal` instead of `Rationale`

### Motivation
According to the docs `truncate` should return `BigDecimal`

### Test plan
It's a part of Ruby core, didn't add the tests for the signatures here, since I didn't see this pattern being wildly used, but I can add if people feel like that is necessary

### DIFF
--- a/rbi/stdlib/bigdecimal.rbi
+++ b/rbi/stdlib/bigdecimal.rbi
@@ -1538,7 +1538,7 @@ class BigDecimal < Numeric
     params(
         arg0: Integer,
     )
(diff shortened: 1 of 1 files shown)


aprocter-stripe:

→ https://go/builds/bui_IpGxVD5NXTqUqE
→ https://go/builds/bui_IpGxFyxAOPFoss

aprocter-stripe (approved):
Thanks for this fix!

PR #3940, opened 2021-01-26
Move hashing logic out of pipeline and into its own package (//hashing)

jvilk-stripe:
Move hashing logic out of pipeline and into its own package called //hashing. (I am open to other package names/locations if folks object.)

Prelude to more invasive changes to the hashing interface.

### Motivation

I have performance optimizations planned for Sorbet startup in the editor which mostly center around the hashing and indexing interfaces. I will likely expand the hashing interface to support faster hashing operations e.g. when we already have an AST for the file.

### Test plan

No functional change. If CI passes everything should work.


hashing/hashing.h
```diff
@@ -0,0 +1,28 @@
...
+namespace sorbet::hashing {
+
+/**
+ * Contains logic pertaining to producing NameHashes for files, which are used in LSP mode to:
+ * - Determine if a file change can be typechecked incrementally or not.
+ * - Determine if a file contains a reference to a particular identifier (fast filter for _find references_).
+ *
+ * However, this file is not just used in the LSP package! `realmain` invokes hashing when the `--store-state` argument
+ * is passed so that it can serialize file contents and their hashes to disk. This is really important for the payload
+ * files embedded in Sorbet itself. As a result, this package is standalone from //main/lsp.
```
elliottt, line 17: Thanks for this!

elliottt (approved):

PR #3944, opened 2021-01-28
Mark props with 'foreign:' annotations as unimplemented

jez:
### Motivation

And by 'unimplemented', I mean that we don't emit the call to `keep_def` that
the prop's MethodDef would otherwise get.

### Test plan

Review by commit for nice diffs.

- **Add failing test** (7418cce5a)

- **Skip emitting keep_def for unimplemented methods** (213840890)

- **Fix failing test** (c01d1f59f)

- **tools/scripts/update_testdata_exp.sh** (c43812463)


test/testdata/rewriter/prop_foreign.rb.ast.exp
```diff
@@ -0,0 +1,163 @@
+begin
+  <emptyTree>
+  class <emptyTree><<C <root>>> < (::<todo sym>)
+    def self.<static-init><<static-init>$CENSORED>(<blk>)
```
elliottt, line 4: Should this file be `.rewrite-tree.exp`?

jez: (done)

elliottt (approved):
👍

PR #3946, opened 2021-01-28
Faster cold cache LSP startup by re-using ASTs for hashing

jvilk-stripe:
Faster cold cache LSP startup by re-using ASTs for hashing. Hashing is now ~44% faster. That does not mean cold cache startup is 44% faster, as hashing only forms one part of startup. I'm seeing ~15% improvements end-to-end. I have additional plans to make startup time faster.

The speed improvements come from:
* Files are only parsed _once_ at startup. The AST from pipeline::index is rewritten into an AST for an empty GlobalState which can be used for hashing purposes.
* The pass that rewrites the AST does double duty and collects sends and constant references (core::UsageHash), which removes the need for an extra full AST pass.

The main change involved adding a second type of `ast::substitute` pass that supports lazily adding new names to a GlobalState. (The existing pass is designed to re-use a lookup table constructed for all names in a GlobalState for many passes, whereas the lazy pass only cares about the subset of names in one particular AST.)

Measurements show that:
* Hashing the entire project is ~44% faster than before because the cost of the new logic is a small fraction of the cost of reparsing each file. On larger projects, that translates into substantial cold cache startup improvements.
  * With that said, cold cache startup is still appreciably slow; we haven't won yet!
* The new substitute pass is 70% slower than just copying an AST. That is, it's relatively cheap.

The new hashing procedure is not yet used once the LSP server is online. That is, edits by the user still incur two parses. That change is slightly riskier; I dropped a comment explaining why. I plan to tackle that change in the next PR.

In addition to the optimizations, I have added a new CLI parameter called `--force-hashing` that causes sorbet to run hashing over the codebase when run from the CLI. There is no reason why a user would ever need to use this option, but it will make it much easier for me to profile the hashing code.

### Motivation

Speed!

### Test plan

Covered by existing tests.

### DIFF
--- a/ast/substitute/Substitute.cc
+++ b/ast/substitute/Substitute.cc
@@ -135,7 +135,7 @@ ExpressionPtr Substitute::run(core::MutableContext ctx, const core::GlobalSubsti
     return what;
 }
 
-ExpressionPtr Substitute::run(core::MutableContext ctx, core::LazyGlobalSubstitution &subst, TreePtr what) {
+ExpressionPtr Substitute::run(core::MutableContext ctx, core::LazyGlobalSubstitution &subst, ExpressionPtr what) {
     SubstWalk walk(subst);
     what = TreeMap::apply(ctx, walk, std::move(what));
     return what;
--- a/ast/substitute/Substitute.cc
+++ b/ast/substitute/Substitute.cc
@@ -6,9 +6,9 @@
 namespace sorbet::ast {
 
 namespace {
-class SubstWalk {
+template <typename T> class SubstWalk {
 private:
-    const core::GlobalSubstitution &subst;
+    T &subst;
 
     ExpressionPtr substClassName(core::MutableContext ctx, ExpressionPtr node) {
         auto *constLit = cast_tree<UnresolvedConstantLit>(node);
@@ -20,7 +20,7 @@ private:
         }
 
         auto scope = substClassName(ctx, std::move(constLit->scope));
-        auto cnst = subst.substitute(constLit->cnst);
+        auto cnst = subst.substituteConstant(constLit->cnst);
 
         return make_expression<UnresolvedConstantLit>(constLit->loc, std::move(scope), cnst);
     }
@@ -42,36 +42,41 @@ private:
     }
 
 public:
-    SubstWalk(const core::GlobalSubstitution &subst) : subst(subst) {}
+    SubstWalk(T &subst) : subst(subst) {}
 
     ExpressionPtr preTransformClassDef(core::MutableContext ctx, ExpressionPtr tree) {
-        auto *original = cast_tree<ClassDef>(tree);
-        original->name = substClassName(ctx, std::move(original->name));
-        for (auto &anc : original->ancestors) {
+        auto &original = cast_tree_nonnull<ClassDef>(tree);
+        original.name = substClassName(ctx, std::move(original.name));
+        for (auto &anc : original.ancestors) {
             anc = substClassName(ctx, std::move(anc));
         }
         return tree;
     }
 
     ExpressionPtr preTransformMethodDef(core::MutableContext ctx, ExpressionPtr tree) {
-        auto *original = cast_tree<MethodDef>(tree);
-        original->name = subst.substitute(original->name);
-        for (auto &arg : original->args) {
+        auto &original = cast_tree_nonnull<MethodDef>(tree);
+        original.name = subst.substitute(original.name);
+        for (auto &arg : original.args) {
             arg = substArg(ctx, std::move(arg));
         }
         return tree;
     }
 
     ExpressionPtr preTransformBlock(core::MutableContext ctx, ExpressionPtr tree) {
-        auto *original = cast_tree<Block>(tree);
-        for (auto &arg : original->args) {
+        auto &original = cast_tree_nonnull<Block>(tree);
+        for (auto &arg : original.args) {
             arg = substArg(ctx, std::move(arg));
         }
         return tree;
     }
 
     ExpressionPtr postTransformUnresolvedIdent(core::MutableContext ctx, ExpressionPtr original) {
-        cast_tree<UnresolvedIdent>(original)->name = subst.substitute(cast_tree<UnresolvedIdent>(original)->name);
+        auto &id = cast_tree_nonnull<UnresolvedIdent>(original);
+        if (id.kind != ast::UnresolvedIdent::Kind::Local) {
+            id.name = subst.substituteConstant(cast_tree<UnresolvedIdent>(original)->name);
+        } else {
+            id.name = subst.substitute(cast_tree<UnresolvedIdent>(original)->name);
+        }
         return original;
     }
 
@@ -81,7 +86,7 @@ public:
     }
 
     ExpressionPtr preTransformSend(core::MutableContext ctx, ExpressionPtr original) {
-        cast_tree<Send>(original)->fun = subst.substitute(cast_tree<Send>(original)->fun);
+        cast_tree_nonnull<Send>(original).fun = subst.substituteSend(cast_tree_nonnull<Send>(original).fun);
         return original;
     }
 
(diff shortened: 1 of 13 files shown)


jvilk-stripe:

→ https://go/builds/bui_IqPVAG1icGVhkT
→ https://go/builds/bui_IqPVl1jNr9HTpN

jvilk-stripe:

→ https://go/builds/bui_IqPgLtYdP3GxEi
→ https://go/builds/bui_IqPgQRRsxLBZnq

jvilk-stripe:

→ https://go/builds/bui_IqSDpGR1lP5hmV
→ https://go/builds/bui_IqSDT5xfk6TyQS

jvilk-stripe:

→ https://go/builds/bui_IqiNpCNvRTjDVP
→ https://go/builds/bui_IqiNRqqSoa1uUs

ast/substitute/Substitute.cc
```diff
@@ -6,9 +6,9 @@
 namespace sorbet::ast {
 
 namespace {
-class SubstWalk {
+template <typename T> class SubstWalk {
```
jvilk-stripe, line 9: I figured that this is an OK use of a template since it's confined to this file. If not, I can introduce a level of indirection via an interface.

ast/substitute/Substitute.cc
```diff
@@ -42,36 +42,41 @@ class SubstWalk {
...
 
 public:
-    SubstWalk(const core::GlobalSubstitution &subst) : subst(subst) {}
+    SubstWalk(T &subst) : subst(subst) {}
 
     ExpressionPtr preTransformClassDef(core::MutableContext ctx, ExpressionPtr tree) {
-        auto *original = cast_tree<ClassDef>(tree);
-        original->name = substClassName(ctx, std::move(original->name));
-        for (auto &anc : original->ancestors) {
+        auto &original = cast_tree_nonnull<ClassDef>(tree);
```
jvilk-stripe, line 48: `cast_tree_nonnull` avoids a tag check at runtime (it's still checked in debug builds). We _know_ that this tree is a ClassDef.

core/GlobalSubstitution.h
```diff
@@ -50,6 +59,49 @@ class GlobalSubstitution {
...
+        }
+
+        auto it = nameSubstitution.find(from);
+        if (it == nameSubstitution.end()) {
+            return defineName(from, allowSameFromTo);
+        }
+        return it->second;
+    }
+
+    NameRef substituteConstant(NameRef from, bool allowSameFromTo = false) {
```
jvilk-stripe, line 92: These new semantically-enhanced:TM: substitute functions allow us to build up the LSP UsageHash (used for faster Find All References) in the same pass that rewrites the tree. Win/win!

Note that `substituteConstant` is a bit of a misnomer. It's not actually a Constant name. It could be a method name. It's called "substituteConstant" because the UsageHash has a field called `constants`.

Perhaps I should use `substituteSymbolName` instead?

hashing/hashing.cc
```diff
@@ -10,87 +12,60 @@
 using namespace std;
 namespace sorbet::hashing {
 namespace {
-class AllNamesCollector {
```
jvilk-stripe, line 13: Folded into `LazyGlobalSubstitution` 🎉

hashing/hashing.cc
```diff
@@ -143,4 +118,74 @@ void Hashing::computeFileHashes(const vector<shared_ptr<core::File>> &files, spd
...
+    workers.multiplexJob("lspStateHash", [fileq, resultq, &asts, &sharedGs, &logger]() {
+        unique_ptr<Timer> timeit;
+        vector<pair<size_t, unique_ptr<const core::FileHash>>> threadResult;
+        int processedByThread = 0;
+        size_t job;
+        {
+            for (auto result = fileq->try_pop(job); !result.done(); result = fileq->try_pop(job)) {
+                if (result.gotItem()) {
+                    if (timeit == nullptr) {
+                        timeit = make_unique<Timer>(logger, "computeFileHashesWorker");
```
jvilk-stripe, line 151: Avoids a race condition when outer function exits before a thread gets work.

Ugh, we really should introduce a better abstraction for parallelization.

main/lsp/LSPIndexer.cc
```diff
@@ -171,25 +171,23 @@ void LSPIndexer::initialize(LSPFileUpdates &updates, WorkerPool &workers) {
...
+            ENFORCE_NO_TIMER(id < indexed.size());
+            indexed[id] = move(ast);
         }
     }
 
-    hashing::Hashing::computeFileHashes(initialGS->getFiles(), *config->logger, workers);
     cache::maybeCacheGlobalStateAndFiles(OwnedKeyValueStore::abort(move(ownedKvstore)), config->opts, *initialGS,
                                          workers, indexed);
 
-    // When inputFileNames is 0 (as in tests), indexed ends up being size 0 because we don't index payload files.
```
jvilk-stripe, line 187: idk what past me was thinking. :)

main/lsp/LSPIndexer.cc
```diff
@@ -208,6 +206,12 @@ LSPFileUpdates LSPIndexer::commitEdit(SorbetWorkspaceEditParams &edit) {
     // Ensure all files have hashes.
     computeFileHashes(edit.updates);
 
+    // TODO(jvilk): Use the new hashing package's new index and hash functionality here to avoid reparsing files twice.
+    // The challenge here is to update `canTakeFastPath` to take old hashes as an argument, since:
+    // - We need the _old_ and _new_ hashes to run `canTakeFastPath`
+    // - `canTakeFastPath` grabs the old hashes out of GlobalState
+    // - But the index and hash functionality updates the file hashes inside GlobalState.
+    // We can probably generalize the 'containsPendingTypecheckUpdate' parameter for this purpose.
```
jvilk-stripe, line 214: I figure this work is best left to a follow-up PR

main/realmain.cc
```diff
@@ -618,8 +628,6 @@ int realmain(int argc, char *argv[]) {
 
         if (!opts.storeState.empty()) {
             gs->markAsPayload();
-            // Store file hashes for LSP.
-            hashing::Hashing::computeFileHashes(gs->getFiles(), *logger, *workers);
```
jvilk-stripe, line 622: Replaced with logic above, which indexes and hashes at the same time. This should speed up payload generation. :)

elliottt (approved):
Had some suggestions, but otherwise this looks great!

hashing/hashing.cc
```diff
@@ -10,87 +12,60 @@
...
+    unique_ptr<core::GlobalState> lgs;
+    core::FileRef fref = makeEmptyGlobalStateForFile(logger, move(forWhat), /* out param */ lgs);
+    auto ast = realmain::pipeline::indexOne(opts(), *lgs, fref);
+
+    // Calculate UsageHash. We use LazyGlobalSubstitution for this purpose, but it will not do any actual substitution
+    // when fromGS == toGS (hence we intentionally do not unfreeze name table).
+    core::LazyGlobalSubstitution subst(*lgs, *lgs);
+    core::MutableContext ctx(*lgs, core::Symbols::root(), fref);
+    ast.tree = ast::Substitute::run(ctx, subst, move(ast.tree));
+    return computeFileHash(lgs, subst.getAllNames(), move(ast));
```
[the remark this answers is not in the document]
jvilk-stripe: Works for me

hashing/hashing.cc
jvilk-stripe: Yeah. I was thinking of defining an interface that contained the functionality used in most of these loops:

* Enqueues work
* Encodes the name of the task (used for metrics/timers)
* Filter function to ignore items that aren't important
* Function that performs the transformation and returns one object for that transformation
* Logic for interruption in LSP mode
* (Maybe) Special cases loops over small number of items to not use WorkerPool
  * We have this optimization in index; dunno how much it saves us.

I started poking at this before but got caught in the weeds trying to support _all_ of the things we do in these passes.

jvilk-stripe: There's an added complication of being thread safe, too. Some of that functionality is accessed from coordinator thread, others is accessed from worker thread.

PR #3951, opened 2021-01-30
Rename UsageHash::constants to symbolNames.

jvilk-stripe:
Rename UsageHash::constants to symbolNames.

We put _all_ symbol names here, not just constants.

### Motivation

Follow-up from https://github.com/sorbet/sorbet/pull/3946

### Test plan

No behavior change.

### DIFF
--- a/core/NameHash.h
+++ b/core/NameHash.h
@@ -47,7 +47,7 @@ struct GlobalStateHash {
 
 struct UsageHash {
     std::vector<core::NameHash> sends;
-    std::vector<core::NameHash> constants;
+    std::vector<core::NameHash> symbolNames;
 };
 
 struct FileHash {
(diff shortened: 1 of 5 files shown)


core/NameHash.h
```diff
@@ -47,7 +47,7 @@ struct GlobalStateHash {
 
 struct UsageHash {
     std::vector<core::NameHash> sends;
-    std::vector<core::NameHash> constants;
+    std::vector<core::NameHash> symbolNames;
```
froydnj, line 50: Nit: I think since this isn't storing names, but hashes, that `symbolsNames` is not the best.  Maybe `symbols`...which is still not quite right but is at least parallel to `sends`, above?

elliottt (approved):
Thanks for the follow up! What do you think about @froydnj's suggestion for `symbols` instead of `symbolNames`?

core/NameHash.h
jvilk-stripe: Fine with me!

PR #3953, opened 2021-02-01
Add LSP support for returning a useful definition for require_relative arguments

garuma:
When detecting that a definition request is made against the pseudo-file path given to a `require_relative` the code will now simply construct a response linking to the relevant file.

### Motivation

Previously trying to navigate on the argument of `require_relative` would simply redirect to itself. This is not consistent with other languages behavior (like C/C++) which allows following relative includes for instance.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/definition.cc
+++ b/main/lsp/requests/definition.cc
@@ -1,6 +1,8 @@
 #include "main/lsp/requests/definition.h"
+#include "common/FileOps.h"
 #include "core/lsp/QueryResponse.h"
 #include "main/lsp/json_types.h"
+#include <filesystem>
 
 using namespace std;
 
@@ -8,6 +10,26 @@ namespace sorbet::realmain::lsp {
 DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
                                unique_ptr<TextDocumentPositionParams> params)
     : LSPRequestTask(config, move(id), LSPMethod::TextDocumentDefinition), params(move(params)) {}
+
+core::Loc
+DefinitionTask::findRequireRelativeLoc(const core::GlobalState &gs,
+                                       const std::vector<std::unique_ptr<core::lsp::QueryResponse>> &responses) {
+    if (responses.size() < 2)
+        return core::Loc::none();
+
+    auto parentSendResp = move(responses[1])->isSend();
+    const bool isRequireRelative = parentSendResp && parentSendResp->callerSideName == core::Names::require_relative();
+    auto literal = move(responses[0])->isLiteral();
+    if (isRequireRelative && literal) {
+        auto literalValue = core::cast_type_nonnull<core::LiteralType>(literal->retType.type).asName(gs).shortName(gs);
+        auto relativeFileName = fmt::format("{}.rb", literalValue);
+        auto baseFilePath = std::filesystem::path(config.uri2FileRef(gs, params->textDocument->uri).data(gs).path());
+        auto targetFilePath = baseFilePath.replace_filename(relativeFileName);
+        auto loc = core::Loc(gs.findFileByPath(targetFilePath.string()), 0, 0);
+        return loc;
+    }
+    return core::Loc::none();
+}
 
 unique_ptr<ResponseMessage> DefinitionTask::runRequest(LSPTypecheckerDelegate &typechecker) {
     auto response = make_unique<ResponseMessage>("2.0", id, LSPMethod::TextDocumentDefinition);
@@ -20,7 +42,13 @@ unique_ptr<ResponseMessage> DefinitionTask::runRequest(LSPTypecheckerDelegate &t
     } else {
         auto &queryResponses = result.responses;
         vector<unique_ptr<Location>> result;
-        if (!queryResponses.empty()) {
+
+        // First, special case detecting require_relative and in that case go to the referenced file.
+        auto requireLoc = findRequireRelativeLoc(gs, queryResponses);
+
+        if (requireLoc.exists()) {
+            addLocIfExists(gs, result, requireLoc);
+        } else if (!queryResponses.empty()) {
             const bool fileIsTyped =
                 config.uri2FileRef(gs, params->textDocument->uri).data(gs).strictLevel >= core::StrictLevel::True;
             auto resp = move(queryResponses[0]);
--- a/test/helpers/position_assertions.cc
+++ b/test/helpers/position_assertions.cc
@@ -10,6 +10,7 @@
 #include "main/lsp/LSPConfiguration.h"
 #include "test/helpers/lsp.h"
 #include "test/helpers/position_assertions.h"
+#include <filesystem>
 #include <iterator>
 #include <regex>
 
@@ -35,6 +36,7 @@ const UnorderedMap<
         {"def", DefAssertion::make},
         {"type", TypeAssertion::make},
         {"type-def", TypeDefAssertion::make},
+        {"file-def", FileDefAssertion::make},
         {"disable-fast-path", BooleanPropertyAssertion::make},
         {"disable-stress-incremental", BooleanPropertyAssertion::make},
         {"enable-packager", BooleanPropertyAssertion::make},
@@ -431,7 +433,7 @@ tuple<string_view, vector<int>, string_view> getSymbolVersionAndOption(string_vi
 
 DefAssertion::DefAssertion(string_view filename, unique_ptr<Range> &range, int assertionLine, string_view symbol,
                            int version, bool isDefOfSelf, bool isDefaultArgValue)
-    : RangeAssertion(filename, range, assertionLine), symbol(symbol), version(version), isDefOfSelf(isDefOfSelf),
+    : BaseDefAssertion(filename, range, assertionLine, symbol), version(version), isDefOfSelf(isDefOfSelf),
       isDefaultArgValue(isDefaultArgValue) {}
 
 shared_ptr<DefAssertion> DefAssertion::make(string_view filename, unique_ptr<Range> &range, int assertionLine,
@@ -472,9 +474,13 @@ vector<unique_ptr<DocumentHighlight>> &extractDocumentHighlights(ResponseMessage
     return get<vector<unique_ptr<DocumentHighlight>>>(highlightsOrNull);
 }
 
-void DefAssertion::check(const UnorderedMap<string, shared_ptr<core::File>> &sourceFileContents, LSPWrapper &lspWrapper,
-                         int &nextId, const Location &queryLoc,
-                         const std::vector<std::shared_ptr<DefAssertion>> &definitions) {
+BaseDefAssertion::BaseDefAssertion(string_view filename, unique_ptr<Range> &range, int assertionLine,
+                                   string_view symbol)
+    : RangeAssertion(filename, range, assertionLine), symbol(symbol) {}
+
+void BaseDefAssertion::check(const UnorderedMap<string, shared_ptr<core::File>> &sourceFileContents,
+                             LSPWrapper &lspWrapper, int &nextId, const Location &queryLoc,
+                             const std::vector<std::shared_ptr<BaseDefAssertion>> &definitions) {
     REQUIRE_FALSE(definitions.empty());
     const int line = queryLoc.range->start->line;
     // Can only query with one character, so just use the first one.
@@ -512,7 +518,7 @@ void DefAssertion::check(const UnorderedMap<string, shared_ptr<core::File>> &sou
 
     vector<unique_ptr<Location>> expectedLocations;
     for (auto &def : definitions) {
-        expectedLocations.emplace_back(def->getLocation(config));
+        expectedLocations.emplace_back(def->getDefinitionLocation(config));
     }
     auto diff = diffLocations(expectedLocations, locations, [](const auto &a, const auto &b) -> int {
         // Note: Sorbet will point to the *statement* that defines the symbol, not just the symbol.
@@ -761,6 +767,39 @@ shared_ptr<TypeAssertion> TypeAssertion::make(string_view filename, unique_ptr<R
 
 string TypeAssertion::toString() const {
     return fmt::format("type: {}", symbol);
+}
+
+FileDefAssertion::FileDefAssertion(std::string_view filename, std::unique_ptr<Range> &range, int assertionLine,
+                                   std::string targetFilename, int targetLine, int targetColumn)
+    : BaseDefAssertion(filename, range, assertionLine, ""sv), targetFilename(targetFilename), targetLine(targetLine),
+      targetColumn(targetColumn) {}
+
+std::shared_ptr<FileDefAssertion> FileDefAssertion::make(std::string_view filename, std::unique_ptr<Range> &range,
+                                                         int assertionLine, std::string_view assertionContents,
+                                                         std::string_view assertionType) {
+    vector<string_view> split = absl::StrSplit(assertionContents, ' ');
+    CHECK_EQ(split.size(), 3);
+    {
+        INFO(fmt::format("Invalid file-def assertion; multiple words found:\n{}\nFile definition assertions should be "
+                         "of the form:\n# [^*] file-def: file-path line column",
+                         assertionContents));
+    }
+    auto file_path = split[0];
+    auto line = stoi(string(split[1]));
+    auto col = stoi(string(split[2]));
+
+    return make_shared<FileDefAssertion>(filename, range, assertionLine, string(file_path), line, col);
+}
+
+std::unique_ptr<Location> FileDefAssertion::getDefinitionLocation(const LSPConfiguration &config) const {
+    auto targetPath = std::filesystem::path(filename).replace_filename(targetFilename);
+    auto targetUri = filePathToUri(config, targetPath.string());
+    auto targetRange = RangeAssertion::makeRange(targetLine, targetColumn, targetColumn + 1);
+    return make_unique<Location>(targetUri, targetRange->copy());
+}
+
+string FileDefAssertion::toString() const {
+    return fmt::format("file: {}:{}+{}", targetFilename, targetLine, targetColumn);
 }
 
 void reportMissingError(const string &filename, const ErrorAssertion &assertion, string_view sourceLine,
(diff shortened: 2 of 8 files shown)


garuma:
@jvilk-stripe I'm having some trouble with the test assertion addition, it seems like everything should work as expected but I'm being nagged by the way the runner handles (or rather doesn't handle) URIs.

Basically the problem seems to be that to construct "URIs" the runner will take the file path and prepend a base URI. To go the other way around it will simply substring URIs to remove the length of that prefix to convert them back to path.

Now in my case I have seen URIs get formed that basically append absolute paths with that base URI (which is wacky) to the effect of something like `file://my/sorbet/prefix//Users/jeremie/bla/bla/file`. Problem is that then some LSP response like mine will then return `file://my/sorbet/prefix/Users/jeremie/bla/bla/otherfile` (notice how the double slash gets removed because of [the conversion logic in LSPConfiguration](https://github.com/sorbet/sorbet/blob/d570e3bac6ccbc5e7678eeef00e8b7763b4a78f8/main/lsp/LSPConfiguration.cc#L122-L170)) thus causing the file path to be understood as a relative path starting with `Users` which then can't be matched back.

Any idea what needs to be changed here for this to work?

main/lsp/requests/definition.cc
```diff
@@ -9,6 +11,26 @@ DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
...
     : LSPRequestTask(config, move(id), LSPMethod::TextDocumentDefinition), params(move(params)) {}
 
+core::Loc
+DefinitionTask::findRequireRelativeLoc(const core::GlobalState &gs,
+                                       const std::vector<std::unique_ptr<core::lsp::QueryResponse>> &responses) {
+    if (responses.size() < 2)
+        return core::Loc::none();
+
+    auto parentSendResp = move(responses[1])->isSend();
+    const bool isRequireRelative = parentSendResp && parentSendResp->callerSideName == core::Names::require_relative();
```
jvilk-stripe, line 21: Note: This is going to match on _any_ function named `require_relative`. It looks like we already do this with `require`, so maybe it's fine, but you might want to check `isPrivateOk` to avoid matching `foo.require_relative`:

https://github.com/sorbet/sorbet/blob/a9063fd1/main/lsp/requests/completion.cc#L803

main/lsp/requests/definition.cc
```diff
@@ -9,6 +11,26 @@ DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
...
+    auto literal = move(responses[0])->isLiteral();
```
jvilk-stripe, line 22: Why are you `move`ing here? This is a weird side effect.

main/lsp/requests/definition.cc
```diff
@@ -9,6 +11,26 @@ DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
...
+    if (isRequireRelative && literal) {
+        auto literalValue = core::cast_type_nonnull<core::LiteralType>(literal->retType.type).asName(gs).shortName(gs);
+        auto relativeFileName = fmt::format("{}.rb", literalValue);
```
jvilk-stripe, line 25: This is going to be `../../some/relative/path.rb`, right?

main/lsp/requests/definition.cc
```diff
@@ -9,6 +11,26 @@ DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
...
+        auto baseFilePath = std::filesystem::path(config.uri2FileRef(gs, params->textDocument->uri).data(gs).path());
```
jvilk-stripe, line 26: Note: There should be some `loc` on the QueryResponse that you can use to get the file reference. No need to do `uri2FileRef`!

main/lsp/requests/definition.cc
```diff
@@ -9,6 +11,26 @@ DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
...
+        auto targetFilePath = baseFilePath.replace_filename(relativeFileName);
```
jvilk-stripe, line 27: Does this simplify the path? e.g. `foo/../foo` => `foo`?

main/lsp/requests/definition.cc
```diff
@@ -9,6 +11,26 @@ DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
...
+        auto loc = core::Loc(gs.findFileByPath(targetFilePath.string()), 0, 0);
```
jvilk-stripe, line 28: You're going to have to canonicalize the path so it contains the exact workspace root. This may be `~/stripe/foo`, or `.`, or `../foo/...`, or whatever. We don't canonicalize it AFAIK.

LSPConfiguration has a rootPath you can use for this:
https://github.com/sorbet/sorbet/blob/a9063fd1/main/lsp/LSPConfiguration.h#L83

I'd suggest:
* Resolving the `require_relative` path to an absolute path
* Replacing the absolute path of the workspace root with the specific `rootPath` form.
* Then finding the file with that canonicalized path.

test/helpers/position_assertions.cc
```diff
@@ -10,6 +10,7 @@
 #include "main/lsp/LSPConfiguration.h"
 #include "test/helpers/lsp.h"
 #include "test/helpers/position_assertions.h"
+#include <filesystem>
```
jez, line 13: We don't use `<filesystem>` elsewhere in Sorbet. Is this ready yet? I don't think this works on all the platforms we want yet.

We have wrappers for filesystem-y things that we want in `common/FileSystem.{h,cc}`

cc @jvilk-stripe would know more here.

jvilk-stripe (changes requested):

main/lsp/requests/definition.cc
jvilk-stripe: Also, I _think_ `require_relative` works if you explicitly specify the extension too...

main/lsp/requests/definition.cc
```diff
@@ -9,6 +11,26 @@ DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
                                unique_ptr<TextDocumentPositionParams> params)
     : LSPRequestTask(config, move(id), LSPMethod::TextDocumentDefinition), params(move(params)) {}
 
+core::Loc
+DefinitionTask::findRequireRelativeLoc(const core::GlobalState &gs,
+                                       const std::vector<std::unique_ptr<core::lsp::QueryResponse>> &responses) {
+    if (responses.size() < 2)
+        return core::Loc::none();
```
jez, line 18:
```suggestion
    if (responses.size() < 2) {
        return core::Loc::none();
    }
```

main/lsp/requests/definition.cc
jvilk-stripe: See `localName2Remote` for more information here; this shows you how to take a `core::File`'s path and turn it into a URI.

https://github.com/sorbet/sorbet/blob/a9063fd1/main/lsp/LSPConfiguration.cc#L122

test/helpers/position_assertions.cc
jvilk-stripe: Yup, +1; put any new needed FS functionality into the file system files.

garuma: Yep, I only use path manipulation functionality from the library (and no actual system case wrapper) which are lacking currently in Sorbet. I was going to implement them back in that layer once the core approach was validated.

main/lsp/requests/definition.cc
```diff
@@ -1,6 +1,8 @@
 #include "main/lsp/requests/definition.h"
+#include "common/FileOps.h"
 #include "core/lsp/QueryResponse.h"
 #include "main/lsp/json_types.h"
+#include <filesystem>
```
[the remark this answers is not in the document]
garuma: See https://github.com/sorbet/sorbet/pull/3953#discussion_r568742203

main/lsp/requests/definition.cc
garuma: @jez any chance to encode that in `format_cxx.sh`?

main/lsp/requests/definition.cc
```diff
@@ -9,6 +11,26 @@ DefinitionTask::DefinitionTask(const LSPConfiguration &config, MessageId id,
                                unique_ptr<TextDocumentPositionParams> params)
     : LSPRequestTask(config, move(id), LSPMethod::TextDocumentDefinition), params(move(params)) {}
 
+core::Loc
+DefinitionTask::findRequireRelativeLoc(const core::GlobalState &gs,
+                                       const std::vector<std::unique_ptr<core::lsp::QueryResponse>> &responses) {
+    if (responses.size() < 2)
```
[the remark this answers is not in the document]
garuma: Correct. Only working on the basis of the literal string doesn't give you enough context to know it's being passed into `require_relative`

main/lsp/requests/definition.cc
garuma: I'll check, maybe that's a left over of something else I did previously

main/lsp/requests/definition.cc
garuma: Correct for 1/, for 2/ I'll check but that seems plausible and this code should be made to support it

main/lsp/requests/definition.cc
garuma: I have a feeling I had started doing that (getting the path from the `loc`) but that made the test unhappy (or maybe it was something else and I didn't revert back) I'll take a look

main/lsp/requests/definition.cc
garuma: It doesn't canonicalize

main/lsp/requests/definition.cc
jez: Happy to if you can find the setting. I don’t think that clang-format has such an option and I don’t feel like fussing with clang-tidy in bazel (in three years of reviewing Sorbet PRs this is the first time I’ve seen someone use no braces, so ROI is probably low)

garuma: A cursory glance at the Googling literature seems to indicate `clang-format` doesn't support this (or that you have to combine it with another tool/clang-tidy to add the braces in the first place). I'll keep it in mind in the future (maybe there is a VSCode side change I can do).

jez:
i'm fine with jvilk's review. thanks for continuing to re-assign me but i don't think it's super necessary

common/Path.cc
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * A simple naive implementation of Path for platform that do not have <filesystem> yet
+ */
+
+template <> BasePath<string>::BasePath(const string_view path) : internal_path(std::string(path)) {}
+
+template <> BasePath<string> &BasePath<string>::replace_filename(const string_view filename) {
+    auto last_sep = internal_path.rfind(file_separator);
+    if (last_sep == std::string::npos)
```
jvilk-stripe, line 20: Braces please.

common/Path.cc
```diff
@@ -0,0 +1,107 @@
...
+        internal_path = filename;
+    else
+        internal_path.replace(last_sep + 1, internal_path.size() - last_sep - 1, filename);
+
+    return *this;
+}
+
+template <> BasePath<string> &BasePath<string>::replace_extension(const string_view extension) {
+    auto last_sep = internal_path.rfind(file_separator);
+    last_sep = last_sep == string::npos ? static_cast<size_t>(-1) : last_sep;
```
jvilk-stripe, line 30: `size_t` is unsigned, right? So this might overflow.

common/Path.cc
```diff
@@ -0,0 +1,107 @@
...
+    if (last_dot == string::npos)
+        internal_path += extension;
+    else
+        internal_path.replace(last_dot, internal_path.size() - last_dot, extension);
+
+    return *this;
+}
+
+template <> BasePath<string> &BasePath<string>::lexically_normal() {
+    // TODO
```
jvilk-stripe, line 41: ????

common/Path.cc
```diff
@@ -0,0 +1,107 @@
...
+    internal_path += path;
+
+    return *this;
+}
+
+template <> const string BasePath<string>::string() {
+    return internal_path;
+}
+
+#ifdef USE_STL_FILESYSTEM
```
jvilk-stripe, line 63: Rather than using templates, can you do what we do for e.g. KeyValueStore and have a `-emscripten.cc` variant? That removes the need to use a template.

common/Path.cc
```diff
@@ -0,0 +1,107 @@
+#include "common/Path.h"
+
+namespace sorbet {
+using namespace std;
+
+#if defined(_WIN32) || defined(_WIN64)
```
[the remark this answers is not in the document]
jvilk-stripe: Looks like you only use this on the emscripten path. Emscripten pretends to be Linux, so you can remove the Windows path.

common/test/common_test.cc
```diff
@@ -238,4 +242,37 @@ TEST_SUITE("UIntSet") {
     }
 }
 
+typedef sorbet::BasePath<std::string> SimplePath;
+
+TEST_SUITE("SimplePath") {
```
jvilk-stripe, line 247: Good good. I was worried this would be untested code. 👍

common/Path.cc
```diff
@@ -0,0 +1,107 @@
...
+constexpr char file_separator = '/';
+#endif
+
+/**
+ * A simple naive implementation of Path for platform that do not have <filesystem> yet
+ */
+
+template <> BasePath<string>::BasePath(const string_view path) : internal_path(std::string(path)) {}
+
+template <> BasePath<string> &BasePath<string>::replace_filename(const string_view filename) {
```
jvilk-stripe, line 18: We use camelCase, btw, ~everywhere. We used to have a mix of snake and camel.

test/helpers/position_assertions.cc
```diff
@@ -763,6 +769,39 @@ string TypeAssertion::toString() const {
...
+    : BaseDefAssertion(filename, range, assertionLine, ""sv), targetFilename(targetFilename), targetLine(targetLine),
+      targetColumn(targetColumn) {}
+
+std::shared_ptr<FileDefAssertion> FileDefAssertion::make(std::string_view filename, std::unique_ptr<Range> &range,
+                                                         int assertionLine, std::string_view assertionContents,
+                                                         std::string_view assertionType) {
+    vector<string_view> split = absl::StrSplit(assertionContents, ' ');
+    CHECK_EQ(split.size(), 3);
+    {
+        INFO(fmt::format("Invalid file-def assertion; multiple words found:\n{}\nFile definition assertions should be "
```
jvilk-stripe, line 783: `INFO` has to go before `CHECK_EQ` and both have to be in the same scope. `INFO` is printed if any checks within the scope fail.

test/helpers/position_assertions.cc
```diff
@@ -763,6 +769,39 @@ string TypeAssertion::toString() const {
...
+
+std::unique_ptr<Location> FileDefAssertion::getDefinitionLocation(const LSPConfiguration &config) const {
+    auto targetPath = make_path(filename).replace_filename(targetFilename);
+    auto targetUri = filePathToUri(config, targetPath.string());
+    auto targetRange = RangeAssertion::makeRange(targetLine, targetColumn, targetColumn + 1);
+    return make_unique<Location>(targetUri, targetRange->copy());
+}
+
+string FileDefAssertion::toString() const {
+    return fmt::format("file: {}:{}+{}", targetFilename, targetLine, targetColumn);
```
jvilk-stripe, line 802: aren't they separated by spaces?

```suggestion
    return fmt::format("file-def: {} {} {}", targetFilename, targetLine, targetColumn);
```

test/helpers/position_assertions.cc
```diff
@@ -763,6 +769,39 @@ string TypeAssertion::toString() const {
...
+    vector<string_view> split = absl::StrSplit(assertionContents, ' ');
+    CHECK_EQ(split.size(), 3);
+    {
+        INFO(fmt::format("Invalid file-def assertion; multiple words found:\n{}\nFile definition assertions should be "
+                         "of the form:\n# [^*] file-def: file-path line column",
+                         assertionContents));
+    }
+    auto file_path = split[0];
+    auto line = stoi(string(split[1]));
+    auto col = stoi(string(split[2]));
```
jvilk-stripe, line 789: Will these ever be anything but `0`? This assertion seems overly general.

jvilk-stripe (changes requested):
Let's not use templates here and make an `-emscripten.cc` version. As a minor note, use camelcase and braces.

I also believe I found an overflow bug.

common/Path.cc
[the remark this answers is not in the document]
garuma: There was at least [this function](https://github.com/sorbet/sorbet/blob/4587ceddc526e48ad574c940a991a409c786e43e/common/common.cc#L132-L135) from FileOps that did also support `\` as a separator so I added that but can remove

common/Path.cc
garuma: Ah crap I forgot about leaving this. I will need to implement that somehow. Flipping back the WIP bit.

common/Path.cc
garuma: I also hate the complexity but the end goal here was to still be able to write tests against the simple implementation that could be run no matter the target. Any idea how I can retain that ability while making this simpler?

common/Path.cc
garuma: Here I model the API to be the same as <filesystem> itself but I'll change to camelCase

test/helpers/position_assertions.cc
garuma: Ultimately I was thinking of improving this functionality to jump to the first definition in the file (e.g. the first module/class definition) hence why I would need a specific line/column

common/Path.cc
jvilk-stripe: You can make a new test target in bazel that includes the test runner and the emscripten variant. lmk if you have qs about it!

test/helpers/position_assertions.cc
jvilk-stripe: In that case, a landing label like we do for other jump-to assertions seems nice. that avoids the brittle annotations problem, where adding whitespace would inadvertently break an assertion

PR #3955, opened 2021-02-02
Fix hashing bug in indexAndComputeFileHashes + small code cleanups

jvilk-stripe:
Fix hashing bug + small code cleanups.

While working on using the new `indexAndComputeFileHashes` routine on the fast path, I uncovered a bug where files were getting the wrong hash. I realized that the cost assumed that the input files and ASTs were in the same order (FileRef order). This _happens to be true_ for all uses of the function now, but is _not true_ on the fast path. That is, this bug is not currently happening in Sorbet.

The fix is simple and cleaner: Use core::FileRef rather than indices to FileRefs for each unit of work.

### Test plan

This change will be covered by LSP tests once I modify the fast path to use this method.

### DIFF
--- a/hashing/hashing.cc
+++ b/hashing/hashing.cc
@@ -127,7 +127,7 @@ vector<ast::ParsedFile> Hashing::indexAndComputeFileHashes(unique_ptr<core::Glob
     auto asts = realmain::pipeline::index(gs, files, opts, workers, kvstore);
     ENFORCE_NO_TIMER(asts.size() == files.size());
 
-    // In parallel, rewrite ASTs to an empty GlobalState and use them for hashing.
+    // Below, we rewrite ASTs to an empty GlobalState and use them for hashing.
     auto fileq = make_shared<ConcurrentBoundedQueue<size_t>>(asts.size());
     for (size_t i = 0; i < asts.size(); i++) {
         auto copy = i;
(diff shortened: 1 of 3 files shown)


hashing/hashing.cc
```diff
@@ -73,20 +73,20 @@ core::FileHash computeFileHashForFile(shared_ptr<core::File> forWhat, spdlog::lo
...
+    for (size_t i = 0; i < files.size(); i++) {
         auto copy = i;
         fileq->push(move(copy), 1);
     }
 
     logger.debug("Computing state hashes for {} files", files.size());
 
-    shared_ptr<BlockingBoundedQueue<vector<pair<int, unique_ptr<const core::FileHash>>>>> resultq =
-        make_shared<BlockingBoundedQueue<vector<pair<int, unique_ptr<const core::FileHash>>>>>(files.size());
+    auto resultq =
```
elliottt, line 84: 🎉

hashing/hashing.cc
```diff
@@ -73,20 +73,20 @@ core::FileHash computeFileHashForFile(shared_ptr<core::File> forWhat, spdlog::lo
 void Hashing::computeFileHashes(const vector<shared_ptr<core::File>> &files, spdlog::logger &logger,
                                 WorkerPool &workers) {
     Timer timeit(logger, "computeFileHashes");
-    shared_ptr<ConcurrentBoundedQueue<int>> fileq = make_shared<ConcurrentBoundedQueue<int>>(files.size());
-    for (int i = 0; i < files.size(); i++) {
+    auto fileq = make_shared<ConcurrentBoundedQueue<size_t>>(files.size());
+    for (size_t i = 0; i < files.size(); i++) {
         auto copy = i;
         fileq->push(move(copy), 1);
```
elliottt, line 79: Why move a size_t?

hashing/hashing.cc
```diff
@@ -28,16 +28,16 @@ pair<ast::ParsedFile, core::UsageHash> rewriteAST(const core::GlobalState &origi
...
+                                                 ast::ParsedFile file) {
     vector<ast::ParsedFile> single;
     single.emplace_back(move(file));
 
     core::Context ctx(*lgs, core::Symbols::root(), single[0].file);
     auto workers = WorkerPool::create(0, lgs->tracer());
     realmain::pipeline::resolve(lgs, move(single), opts(), *workers);
 
-    return {move(*lgs->hash()), move(usageHash)};
+    return make_unique<core::FileHash>(core::FileHash{move(*lgs->hash()), move(usageHash)});
```
elliottt, line 40: What do you think about adding a constructor to `core::FileHash` so that we can forward the args here?

elliottt (approved):
👍

hashing/hashing.cc
jvilk-stripe: iirc `push` takes a && so you have to

hashing/hashing.cc
jvilk-stripe: Works for me.

PR #3956, opened 2021-02-02
Docs: Clarify that sorbet-runtime only has partial support for generics

jaredbeck:
`sorbet-runtime` only has partial support for generics.

```ruby
require 'sorbet-runtime'
T.let('1', Integer) # TypeError
T.let(['1', '2'], T::Array[Integer]) # No error
```

Should we make this more prominent [in the docs]?

### DIFF
--- a/website/docs/stdlib-generics.md
+++ b/website/docs/stdlib-generics.md
@@ -6,7 +6,9 @@ sidebar_label: Arrays & Hashes
 
 > TODO: This page is still a fragment. Contributions welcome!
 
-Sorbet supports generic types. The syntax looks likes `MyClass[Elem]`. For user
+`sorbet-static` supports generic types. (`sorbet-runtime` has partial support, see below) 
+
+The syntax looks likes `MyClass[Elem]`. For user
 defined generic classes, it's possible to make this valid Ruby syntax.
 
 However, it's not possible to change the syntax for classes in the Ruby standard


jez:
Not a huge fan of this wording. `sorbet-static` vs `sorbet-runtime` is a mouthful in the first sentence. Nearly every feature in Sorbet has differing levels of runtime/static support.

In the same file, further down, there is this large warning:

```markdown
The type parameters for the generic types provided by the standard library are
only used as hints for the static type system, and are ignored entirely by the
runtime type system.

> **Note**: Sorbet used to take these type parameters into account during
> runtime type-checking, but this turned out to be an common and
> difficult-to-debug source of performance problems: In order to verify that an
> array contained the values it claimed it did, the Sorbet runtime used to
> recursively check the type of every member of a collection, which would take a
> long time for arrays or hashes of a sufficiently large size. Consequently,
> this behavior was eventually removed.
```

It sounds like this is not enough? Would you like this more clear by adding a code snippet that shows what is and is not checked statically and at runtime?

jaredbeck:
Hi Jake :wave:

> Not a huge fan of this wording. `sorbet-static` vs `sorbet-runtime` is a mouthful in the first sentence.

Agreed.



> Nearly every feature in Sorbet has differing levels of runtime/static support.

I .. didn't know that. 😟 

> The type parameters for the generic types provided by the standard library are
only used as hints for the static type system, and are ignored entirely by the
runtime type system.

This is nicely written.

I guess where I'm coming from is .. I was using `sorbet-runtime` today, and I expected it to catch the things that static catches, and I feel like that's going to trip a lot of people up. So, I'm just trying to find a way to counter that expectation earlier in the docs. Totally open to any ideas on how to achieve that.

I can talk a bit more about which development tasks depend on runtime checking, eg. places in a rails app that sorbet-static can't "see", if that's helpful.

trevor-stripe (changes requested):
@jaredbeck is there a change that we can make to call this difference out higher in the stdlib-generics document, or is there somewhere else that you'd like to see this difference surfaced?

jaredbeck:
> @jaredbeck is there a change that we can make to call this difference out higher in the stdlib-generics document, or is there somewhere else that you'd like to see this difference surfaced?

Higher than line 9, as I suggested in f7fe918ac1780d352fd577d0b58cd3d8cb07a6e8? Sorry, I don't really understand the question. f7fe918ac1780d352fd577d0b58cd3d8cb07a6e8 (this PR) is what I'm proposing. It's my only idea so far. The fact is, runtime support for generics is _partial_. People shouldn't rely on it. So that's what I wrote. That's what I'm trying to emphasize. I think the high placement of line 9 gives this fact the correct gravitas. I hope this is a coherent explanation of my motivation here :)

elliottt:
> Higher than line 9, as I suggested in [f7fe918](https://github.com/sorbet/sorbet/commit/f7fe918ac1780d352fd577d0b58cd3d8cb07a6e8)? Sorry, I don't really understand the question. [f7fe918](https://github.com/sorbet/sorbet/commit/f7fe918ac1780d352fd577d0b58cd3d8cb07a6e8) (this PR) is what I'm proposing. It's my only idea so far. The fact is, runtime support for generics is _partial_. People shouldn't rely on it. So that's what I wrote. That's what I'm trying to emphasize. I think the high placement of line 9 gives this fact the correct gravitas. I hope this is a coherent explanation of my motivation here :)

Sorry, it wasn't clear from the earlier discussion on this PR if you were going to reword your original commit to address @jez's comments, or if you were waiting on us to provide more feedback.

Would it be reasonable to put the comments that @jez pointed out in a section called something like "runtime type erasure", and put a warning with a link to that section at the top? Would adding a section header be enough? Would adding an example of where things can go wrong but still pass the static phase be good?

jaredbeck:
> Would it be reasonable to put the comments that @jez pointed out in a section called something like "runtime type erasure", and put a warning with a link to that section at the top? Would adding a section header be enough? Would adding an example of where things can go wrong but still pass the static phase be good?

While I like these ideas less than f7fe918ac1780d352fd577d0b58cd3d8cb07a6e8, they may still be helpful to someone, and I'd encourage you to pursue them. Thanks for reviewing my suggestion.

PR #3957, opened 2021-02-02
Hash and index at the same time on the fast path.

jvilk-stripe:
Hash and index at the same time on the fast path.

Also stop using fref.id() in maps.

Also convert a few remaining cast_tree => cast_tree_nonnull.

### Motivation

Speed improvements. We now only have to parse file changes once when they occur in the common case.

There's one scenario where we may still need to parse files twice: If the slow path is running and we need to check if an edit can preempt. In order to do this, we need to hash the file changes _before_ scheduling them to be commited to LSPIndexer. However, this is an edge case, and the "cost" involved is much much lower than restarting the slow path.

### Test plan

Covered by existing tests (which surfaced the bug I fixed in #3955).

### DIFF
--- a/main/lsp/LSPIndexer.cc
+++ b/main/lsp/LSPIndexer.cc
@@ -18,8 +18,8 @@ using namespace std;
 namespace sorbet::realmain::lsp {
 namespace {
 const core::File &getOldFile(core::FileRef fref, const core::GlobalState &gs,
-                             const UnorderedMap<int, shared_ptr<core::File>> &evictedFiles) {
-    const auto &it = evictedFiles.find(fref.id());
+                             const UnorderedMap<core::FileRef, shared_ptr<core::File>> &evictedFiles) {
+    const auto &it = evictedFiles.find(fref);
     if (it != evictedFiles.end()) {
         return *it->second;
     }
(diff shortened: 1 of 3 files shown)


main/lsp/LSPIndexer.cc
```diff
@@ -82,8 +82,8 @@ void LSPIndexer::computeFileHashes(const vector<shared_ptr<core::File>> &files)
     computeFileHashes(files, *emptyWorkers);
 }
 
-bool LSPIndexer::canTakeFastPath(const std::vector<std::shared_ptr<core::File>> &changedFiles,
-                                 bool containsPendingTypecheckUpdates) const {
+bool LSPIndexer::canTakeFastPath(const vector<shared_ptr<core::File>> &changedFiles,
+                                 const UnorderedMap<core::FileRef, shared_ptr<core::File>> &evictedFiles) const {
```
jvilk-stripe, line 86: I generalized an existing mechanism which supported checking if an edit can take a fast path in the immediate past. The boolean becomes a map containing files that have been replaced in GlobalState so that we can fetch their contents.

main/lsp/LSPIndexer.cc
```diff
@@ -259,14 +249,18 @@ LSPFileUpdates LSPIndexer::commitEdit(SorbetWorkspaceEditParams &edit) {
...
+                                                                 *emptyWorkers, kvstore);
         update.updatedFileIndexes.resize(trees.size());
         for (auto &ast : trees) {
             const int i = fileToPos[ast.file];
             update.updatedFileIndexes[i] = move(ast);
         }
     }
 
+    // _Now_ that we've computed file hashes, we can make a fast path determination.
+    update.canTakeFastPath = canTakeFastPath(update, newlyEvictedFiles);
```
jvilk-stripe, line 262: We used to run `canTakeFastPath` and _then_ update `GlobalState` with the new files.

Now, we update `GlobalState` and _then_ run `canTakeFastPath`. Thus, we need to pass in the old files, pre-update, which contain their old pre-update hashes.

main/lsp/LSPIndexer.h
```diff
@@ -36,18 +36,20 @@ class LSPIndexer final {
      * that we correctly track the latency of canceled & rescheduled typechecking operations. */
     std::vector<std::unique_ptr<Timer>> pendingTypecheckDiagnosticLatencyTimers;
     /** Contains files evicted by `pendingTypecheckUpdates`. Used to make fast path decisions in the immediate past. */
-    UnorderedMap<int, std::shared_ptr<core::File>> evictedFiles;
+    UnorderedMap<core::FileRef, std::shared_ptr<core::File>> evictedFiles;
```
jvilk-stripe, line 39: It's unlikely we'll see 2^31 files anytime soon, but I don't want to take any chances.

elliottt (approved):
🎉

PR #3958, opened 2021-02-02
Implement a hash function on every Type

jvilk-stripe:
Implement a hash function on every Type

This change will substantially speed up Sorbet start-up in editor on a cold cache (up to ~50% on the hashing part of startup). Measured on Stripe's codebase, the effective hashing speedup is 40%.

When Sorbet starts up on a cold cache, it needs to hash every file in the project. This can take awhile on a large project. Hashing involves running resolver on the file's AST in an empty GlobalState and taking the hash of that GlobalState object.

Metrics show that ~50% of the hashing time spent in `computeFileHash` is in `TypePtr::hash`, which expensively hashes `TypePtr::toString()` (which allocates and formats strings -- with tabs! -- and often calls `Symbol::toString()`).

The new hashing routines use Symbol IDs and should overall be much cheaper to run.

### Motivation

Speed++

### Test plan

Covered by existing tests.

### DIFF
--- a/core/types/hashing.cc
+++ b/core/types/hashing.cc
@@ -34,23 +34,23 @@ u4 LiteralType::hash(const GlobalState &gs) const {
     ClassOrModuleRef undSymbol = cast_type_nonnull<ClassType>(underlying).symbol;
     result = mix(result, undSymbol.id());
 
-    // Strings or numbers encode an extra value. True/False are completely represented by the class symbol.
-    // TODO: Casts are bad???
-    if (undSymbol == Symbols::String() || undSymbol == Symbols::Symbol()) {
-        result = mix(result, asName(gs).hash(gs));
-    } else if (undSymbol == Symbols::Integer()) {
-        u8 val = absl::bit_cast<u8>(asInteger());
-        u4 topBits = static_cast<u4>(val >> 32);
-        u4 bottomBits = static_cast<u4>(val & 0xFFFF);
-        result = mix(result, topBits);
-        result = mix(result, bottomBits);
-    } else if (undSymbol == Symbols::Float()) {
-        u8 val = absl::bit_cast<u8>(asFloat());
-        u4 topBits = static_cast<u4>(val >> 32);
-        u4 bottomBits = static_cast<u4>(val & 0xFFFF);
-        result = mix(result, topBits);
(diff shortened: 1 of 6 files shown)


jvilk-stripe:

→ https://go/builds/bui_IrxrGfO0kfl1jX
→ https://go/builds/bui_IrxrBN1tCZzaTu

core/types/hashing.cc
```diff
@@ -0,0 +1,128 @@
...
+    result = mix(result, undSymbol.id());
+
+    // Strings or numbers encode an extra value. True/False are completely represented by the class symbol.
+    // TODO: Casts are bad???
+    if (undSymbol == Symbols::String() || undSymbol == Symbols::Symbol()) {
+        result = mix(result, asName(gs).hash(gs));
+    } else if (undSymbol == Symbols::Integer()) {
+        u8 val = absl::bit_cast<u8>(asInteger());
+        u4 topBits = static_cast<u4>(val >> 32);
+        u4 bottomBits = static_cast<u4>(val & 0xFFFF);
```
froydnj, line 44: You probably meant 0xFFFF_FFFF here?

core/types/hashing.cc
```diff
@@ -0,0 +1,128 @@
...
+        result = mix(result, topBits);
+        result = mix(result, bottomBits);
+    } else if (undSymbol == Symbols::Float()) {
+        u8 val = absl::bit_cast<u8>(asFloat());
+        u4 topBits = static_cast<u4>(val >> 32);
+        u4 bottomBits = static_cast<u4>(val & 0xFFFF);
```
froydnj, line 50: 0xFFFF_FFFF here as well?  Or maybe early return in the string/symbol case and merge the integer/float cases (at least once you've extracted the u8 from them)?

froydnj (commented):
Hashing the result of `toString`... :man_facepalming:

core/types/hashing.cc
jvilk-stripe: oof yes

core/types/hashing.cc
jvilk-stripe: Good idea. Fixed.

froydnj (approved):

PR #3959, opened 2021-02-02
Create a Magic.mixes_in_class_methods method

KaanOzkan:
Created a `Magic.mixes_in_class_methods` method to be used with ActiveSupport::Concern rewriter https://github.com/sorbet/sorbet/pull/3468#issuecomment-771137943

### Motivation
We don't want to insert `extend T::Helpers` in the rewriter stage in order to use `mixes_in_class_methods` as it would cause runtime errors that wouldn't be caught in the static check. In the rewriter we can use this method instead to achieve the same result without `extend T::Helpers`

### Test plan

These diffs were tested as part of the rewriter logic in https://github.com/sorbet/sorbet/pull/3468. Did not add unit tests specific to this method but can add if necessary.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -567,6 +567,15 @@ private:
 
         auto encounteredError = false;
         for (auto &arg : send->args) {
+            auto self = ast::cast_tree<ast::Local>(arg);
+            if (self != nullptr && self->localVariable._name == core::Names::selfLocal()) {
+                auto recv = ast::cast_tree<ast::ConstantLit>(send->recv);
+                if (recv != nullptr && recv->symbol == core::Symbols::Magic()) {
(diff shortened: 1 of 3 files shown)


resolver/resolver.cc
```diff
@@ -567,6 +567,15 @@ class ResolveConstantsWalk {
 
         auto encounteredError = false;
         for (auto &arg : send->args) {
+            auto self = ast::cast_tree<ast::Local>(arg);
+            if (self != nullptr && self->localVariable._name == core::Names::selfLocal()) {
```
jez, line 571:
```suggestion
            if (self != nullptr && self->isSelfReference()) {
```

jez, line 571:
```suggestion
            if (arg.isSelfReference()) {
```

jez (approved):
Thanks!

PR #3961, opened 2021-02-02
Type hashing: Avoid NameRef::hash.

jvilk-stripe:
Type hashing: Avoid NameRef::hash.

NameRef::hash can use name IDs in hashing, which are not stable during hashing across file changes; it's impacted by the order in which names are encountered in an AST pass.

NameRef::hash still has use; it is safely used to quickly find a slot in GlobalState's name table. However, we cannot use it in file hashes.

I have replaced the only use of this in file hash routines with a hash of the name's shortname.

Note that it's possible that these namerefs are always of type "UTF8", in which case this is a no-op change; the ::hash of a UTF8 is its shortname. Still, I don't want code around that encourages the use of NameRef::hash in file hashes.

### Motivation

Prevent bugs.

### Test plan

Covered by existing tests.

### DIFF
--- a/core/types/hashing.cc
+++ b/core/types/hashing.cc
@@ -38,7 +38,7 @@ u4 LiteralType::hash(const GlobalState &gs) const {
     switch (literalKind) {
         case LiteralType::LiteralTypeKind::String:
         case LiteralType::LiteralTypeKind::Symbol:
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Thanks for the explanation!

PR #3962, opened 2021-02-02
[WIP] Faster Symbol::membersStableOrderSlow

jvilk-stripe:
[WIP] Faster Symbol::membersStableOrderSlow

Rather than repeatedly compare strings, calculate a hash for each name eagerly and use that for comparison. Use symbol ID as a tiebreaker.

Results in a ~12% hashing speedup.

TODO:
* String sort the output for completion and maybe document symbol as it's semantically important
* Is there a way to get rid of the tuple=>pair transformation?
* Regenerate exp files.

### Motivation

Hashing spends ~37% of its time in this method.

### Test plan

Covered by existing tests.


jvilk-stripe:

→ https://go/builds/bui_IsFfB9pjAyNFSf
→ https://go/builds/bui_IsFfyNm8e83g3z

jvilk-stripe:
Closing in favor of https://github.com/sorbet/sorbet/pull/3965; may re-open with more targeted changes to membersStableOrderSlow.

PR #3963, opened 2021-02-02
make string variables for versioning arrays, not pointers

froydnj:
This makes them take up slightly less space, since we're now only allocating space for the character array itself, not the character array + a pointer to the character array.  A nice side-effect is the removal of some read/write data, which can't be shared across processes.

### Motivation

Epsilon space savings.

### Test plan

See included automated tests.


elliottt (approved):
Nice!

PR #3964, opened 2021-02-02
make the error message for final method violations slightly clearer

froydnj:
We're very close to turning on final method errors in Stripe's codebase, and the current error message is not super-helpful.  This change at least points you to the definition of the final method and whatever caused the redefinition error to occur.  The latter is not necessarily the actual definition of the method, but is often an `include` or similar pulling in the actual class/module that redefines the method.  So there's still a bit of sleuthing required, but at least this message is more helpful than before.

### Motivation

More helpful error messages.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -129,6 +129,7 @@ module T::Private::Methods
           begin
             raise pretty_message
           rescue => e
+            T::Private::DeclState.current.reset!
             T::Configuration.sig_validation_error_handler(e, {})
           end
         end
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -114,7 +114,7 @@ module T::Private::Methods
            final_method?(method_owner_and_name_to_key(ancestor, method_name))
           definition_file, definition_line = T::Private::Methods.signature_for_method(ancestor.instance_method(method_name)).method.source_location
           is_redefined = target == ancestor
-          caller_loc = caller_locations&.find {|l| !l.to_s.match?(%r{sorbet-runtime[^/]*/lib/types/private/methods/}) }
+          caller_loc = caller_locations&.find {|l| !l.to_s.match?(%r{sorbet-runtime[^/]*/lib/}) }
           extra_info = "\n"
           if caller_loc
             extra_info = (is_redefined ? "Redefined" : "Overriden") + " here: #{caller_loc.path}:#{caller_loc.lineno}\n"
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -117,7 +117,7 @@ module T::Private::Methods
           caller_loc = caller_locations&.find {|l| !l.to_s.match?(%r{sorbet-runtime[^/]*/lib/types/private/methods/}) }
           extra_info = "\n"
           if caller_loc
-            extra_info = "Overriden here: #{caller_loc.path}:#{caller_loc.lineno}\n"
+            extra_info = (is_redefined ? "Redefined" : "Overriden") + " here: #{caller_loc.path}:#{caller_loc.lineno}\n"
           end
 
           error_message = "The method `#{method_name}` on #{ancestor} was declared as final and cannot be " +
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -474,6 +474,12 @@ module T::Private::Methods
       return
     end
 
+    # See https://github.com/sorbet/sorbet/pull/3964 for an explanation of why this
+    # check (which theoretically should not be needed) is actually needed.
+    if !mod.is_a?(Module)
+      return
+    end
+
     if mod.singleton_class?
       mod.include(SingletonMethodHooks)
     else
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -129,6 +129,12 @@ module T::Private::Methods
           begin
             raise pretty_message
           rescue => e
+            # sig_validation_error_handler raises by default; on the off chance that
+            # it doesn't raise, we need to ensure that the rest of signature building
+            # sees a consistent state.  This sig failed to validate, so we should get
+            # rid of it.  If we don't do this, errors of the form "You called sig
+            # twice without declaring a method in between" will non-deterministically
+            # crop up in tests.
             T::Private::DeclState.current.reset!
             T::Configuration.sig_validation_error_handler(e, {})
           end
@@ -168,6 +174,9 @@ module T::Private::Methods
     end
     _check_final_ancestors(mod, mod.ancestors, [method_name])
 
+    # We need to fetch the active declaration again, as _check_final_ancestors
+    # may have reset it (see the comment in that method for details).
+    current_declaration = T::Private::DeclState.current.active_declaration
     if current_declaration.nil?
       return
     end
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -126,7 +126,11 @@ module T::Private::Methods
                            "Made final here: #{definition_file}:#{definition_line}\n" \
                            "#{extra_info}"
 
-          T::Configuration.sig_validation_error_handler(pretty_message, {})
+          begin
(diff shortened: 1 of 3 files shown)


gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -449,6 +455,10 @@ def self.install_hooks(mod)
       return
     end
 
+    if !mod.is_a?(Module)
+      return
+    end
```
jez, line 460: We should not land this! This was a hack I did to make the logs less chunderous.

jez: If this did not cause a test to fail, please add a test so that this for sure fails.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -112,10 +112,16 @@ def self._check_final_ancestors(target, target_ancestors, source_method_names)
             ancestor.private_method_defined?(method_name) ||
             ancestor.protected_method_defined?(method_name)) &&
            final_method?(method_owner_and_name_to_key(ancestor, method_name))
-          raise(
-            "The method `#{method_name}` on #{ancestor} was declared as final and cannot be " +
-            (target == ancestor ? "redefined" : "overridden in #{target}")
-          )
+          caller_loc = T.must(caller_locations&.find {|l| !l.to_s.match?(%r{sorbet-runtime[^/]*/lib/types/private/methods/}) })
```
jez, line 115: We should not raise here if we can't find a caller_location

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -112,10 +112,16 @@ def self._check_final_ancestors(target, target_ancestors, source_method_names)
...
+          definition_file, definition_line = T::Private::Methods.signature_for_method(ancestor.instance_method(method_name)).method.source_location
+
+          error_message = "The method `#{method_name}` on #{ancestor} was declared as final and cannot be " +
+                          (target == ancestor ? "redefined" : "overriden in #{target}")
+          pretty_message = "#{error_message}\n" \
+                           "Made final here: #{definition_file}:#{definition_line}\n" \
+                           "Overriden here: #{caller_loc.path}:#{caller_loc.lineno}\n"
+
+          T::Configuration.sig_validation_error_handler(pretty_message, {})
```
jez, line 124: Can we add a test that we're calling `sig_validation_error_handler` in the same way we do for other `T::Configuration` handlers? For example we should assert that if you set the sig validation handler to something that doesn't raise, that the code overall that has a final violation doesn't raise either.

jez (commented):
^

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
froydnj: Making this `mod = mod.singleton_class if !mod.is_a?(Module)` leads to equivalent results (and should perhaps be moved up a bit if we decided to go that route).  Removing this bit leads to problems with custom decorators and props.

jez: Yes I'm saying that those problems are real final violations, and when I was looking into those I knew that we'd have to fix all of those, and wanted to see the rest of the errors, so I skipped them with this line, knowing that we'd have to go back and fix those violations too.

froydnj: So, `mod = mod.singleton_class if !mod.is_a?(Module)` is probably wrong, but the early return here is not.  Actually, the early return should probably be quite a bit earlier so that we don't shove non-`Module` things into `@installed_hooks`.  The sequence of events is:

1. A class including `T::Props` (the particular instance has a custom decorator), `Model`, includes a module that defines `ClassMethods`.  Call this included module G.
2. We call `T::Props::ClassMethods#included`, passing `Model` as the child.

https://github.com/sorbet/sorbet/blob/88e72f6da215d2952b761b0f4398d4296bfc16e3/gems/sorbet-runtime/lib/types/props/_props.rb#L141-L144

3. We, via `super`, call `T::Props::Plugin::ClassMethods#included`.

https://github.com/sorbet/sorbet/blob/88e72f6da215d2952b761b0f4398d4296bfc16e3/gems/sorbet-runtime/lib/types/props/plugin.rb#L8-L13

4. The call to `plugin` resolves to the mixed-in `T::Props::ClassMethods#plugin`; `mod` is G.

https://github.com/sorbet/sorbet/blob/88e72f6da215d2952b761b0f4398d4296bfc16e3/gems/sorbet-runtime/lib/types/props/_props.rb#L122-L125

5. The call to `plugin` resolves to `T::Props::Decorator::Plugin#plugin`.

https://github.com/sorbet/sorbet/blob/88e72f6da215d2952b761b0f4398d4296bfc16e3/gems/sorbet-runtime/lib/types/props/decorator.rb#L670-L675

6.  We call `T::Props::Plugin::Private.apply_decorator_methods(mod, self)`.  Note that `self` here is an *instance* of the decorator.

https://github.com/sorbet/sorbet/blob/88e72f6da215d2952b761b0f4398d4296bfc16e3/gems/sorbet-runtime/lib/types/props/plugin.rb#L29-L35

7. `target` is an instance of our decorator and `plugin` is G, above, so we call `decorator.extend(G::DecoratorMethods)`
8. We wind up in our replaced method hook for `extended`:

https://github.com/sorbet/sorbet/blob/88e72f6da215d2952b761b0f4398d4296bfc16e3/gems/sorbet-runtime/lib/types/private/methods/_methods.rb#L408-L411

9. `arg` is an instance of the decorator and `self` is `G::DecoratorMethods`.  We think `arg` is a module, but that is not necessarily the case.
10. We call `_hook_impl(arg, arg.singleton_class.ancestors, self)`.

https://github.com/sorbet/sorbet/blob/88e72f6da215d2952b761b0f4398d4296bfc16e3/gems/sorbet-runtime/lib/types/private/methods/_methods.rb#L384-L393

11. We call `install_hooks(target)`, and `install_hooks` is receiving an instance, not a `Module`.

This is all a bit deep in the metaprogramming weeds for me, but I am loath to tug on any piece of this.  If I had to pull on something, I'd guess that the `apply_decorator_methods` call/code needs to be modified, but I am completely unsure of how.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -112,10 +112,26 @@ def self._check_final_ancestors(target, target_ancestors, source_method_names)
...
+          error_message = "The method `#{method_name}` on #{ancestor} was declared as final and cannot be " +
+                          (is_redefined ? "redefined" : "overridden in #{target}")
+          pretty_message = "#{error_message}\n" \
+                           "Made final here: #{definition_file}:#{definition_line}\n" \
+                           "#{extra_info}"
+
+          begin
+            raise pretty_message
+          rescue => e
+            T::Private::DeclState.current.reset!
```
jez, line 132: What's this `DeclState.current.reset!` for?

froydnj: Otherwise it looks like we're still building a sig and we get spurious test failures.

jez: from offline: when we were unconditionally raising, it didn't matter whether whether we clean up to a known good state because the `raise` would crash the program anyways. but when the validation handler might not raise, we have to go out of our way to make sure we're setting the state good again.

@froydnj one question: does calling `reset!` here make this read value stale?

https://github.com/sorbet/sorbet/blob/d570e3bac6ccbc5e7678eeef00e8b7763b4a78f8/gems/sorbet-runtime/lib/types/private/methods/_methods.rb#L147

i.e., maybe we should re-get current_declaration after calling `_check_final_ancestors`? not sure what the current behavior is if we're not doing that.

jez (approved):
one question, one request for a comment

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
froydnj: To expand on that explanation: we currently always `raise` for `final` errors.  This PR changes things so that we go through `sig_validation_error_handler` for `final` errors and also adds a test that:

1. We do the "correct" thing when `sig_validation_error_handler` is using its default behavior.
2. Things continue to work even if `sig_validation_error_handler` doesn't `raise`.

But this second test causes a problem: after we don't `raise` and return normally, `DeclState` still thinks it's building a sig.  So, depending on which test executes next, we might get errors about attempting to build a sig while we're building a sig.  Therefore, we need this reset to ensure that `DeclState` is consistent no matter what the handler does.

jez: (were you planning to leave that comment here, or check it into the code?)

froydnj: I checked in a cut-down version of the comment.

PR #3965, opened 2021-02-03
Symbol::hash: Speed up hashing by using SymbolRef order.

jvilk-stripe:
SymbolRef IDs are stable during hashing.

Delurks use of membersStableOrderSlow, which uses alphabetical order which is _much_ more expensive to compute!

Speeds up hashing by ~14%.

### Motivation

Hashing spends ~37% of its time in membersStableOrderSlow.

### Test plan

Covered by existing tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -1341,9 +1341,19 @@ u4 Symbol::hash(const GlobalState &gs) const {
     result = mix(result, this->owner._id);
     result = mix(result, this->superClassOrRebind._id);
     // argumentsOrMixins, typeParams, typeAliases
-    for (auto e : membersStableOrderSlow(gs)) {
-        if (e.second.exists() && !e.second.data(gs)->ignoreInHashing(gs)) {
-            result = mix(result, _hash(e.second.data(gs)->name.shortName(gs)));
+    if (!members().empty()) {
+        // Rather than use membersStableOrderSlow, which is... slow..., use an order dictated by symbol ref ID.
+        // It's faster to sort, and SymbolRef IDs are stable during hashing.
+        vector<core::SymbolRef> membersToHash;
+        membersToHash.reserve(this->members().size());
+        for (auto e : this->members()) {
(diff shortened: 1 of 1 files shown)


jvilk-stripe:

→ https://go/builds/bui_IsbLqDfWWCINKf
→ https://go/builds/bui_IsbLbpAKwhw0Tm

core/Symbols.cc
```diff
@@ -1341,9 +1341,19 @@ u4 Symbol::hash(const GlobalState &gs) const {
...
+        vector<core::SymbolRef> membersToHash;
+        membersToHash.reserve(this->members().size());
+        for (auto e : this->members()) {
+            if (e.second.exists() && !e.second.data(gs)->ignoreInHashing(gs)) {
+                membersToHash.emplace_back(e.second);
+            }
+        }
+        fast_sort(membersToHash, [](const auto &a, const auto &b) -> bool { return a.rawId() < b.rawId(); });
+        for (auto member : membersToHash) {
+            result = mix(result, _hash(member.data(gs)->name.shortName(gs)));
```
jez, line 1356: I see that the old code did this, but why are we using short name here?

That would mean that if a name changes from being a constant name to a utf8 name (which would no longer be the same name for many purposes), the hash would remain unchanged.

jvilk-stripe: Gooooooooooooooood question.

jvilk-stripe: I can re-introduce the NameRef::contentHash from my other PR if that helps?

jvilk-stripe: Is there a case where this would change and Sorbet doesn't remove or delete a symbol?

jez: @jvilk-stripe Not quite sure, I can think about it later if you need me to.

jvilk-stripe: I think the answer is 'no', but directionally speaking it'd be good to include a comprehensive content-based hash of the name.

jez (approved):
thanks!

PR #3966, opened 2021-02-03
Bump dependencies in advance of clang-10

elliottt:
### Motivation

Bumping dependencies for  clang-10 compatibility, and adding "-Wno-implicit-fallthrough" for the ones that fail with it enabled.

### Test plan

See included automated tests.


jez (approved):
Thanks!

PR #3968, opened 2021-02-03
Remove NameRef::hash and inline it completely into GlobalState.cc.

jvilk-stripe:
Remove NameRef::hash and inline it completely into GlobalState.cc.

Some notes on this function:

* It _cannot_ be used like Symbol::hash because it encodes properties that are unstable across file hashes (name ID)
* The hash itself is intended for one very specific purpose: Determining the position of the name in `namesByHash`
* The hash itself is only ever used in GlobalState.cc.
* The implementation of the hash is distributed between Hashing.h, GlobalState.cc, and NameRef.cc.

I've also removed a function that converts the NameKind enum into an integer since we can just cast them (they have the same value as returned by that conversion function).

In a subsequent PR, I'll introduce a _new_ NameRef::hash that _can_ be used like Symbol::hash (but will not replace this very specific hash).

### Motivation

Clean up the code and remove a footgun. If we were to depend on NameRef::hash in FileHash, we would have subtle fast path decision bugs where we take the slow path and don't have to.

### Test plan

Covered by existing tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -31,23 +31,23 @@ namespace sorbet::core {
 namespace {
 // Hash functions used to determine position in namesByHash.
 
-inline unsigned int _hash_mix_unique(UniqueNameKind unk, unsigned int num, unsigned int rawId) {
+inline unsigned int hashMixUnique(UniqueNameKind unk, unsigned int num, unsigned int rawId) {
     return mix(mix(num, static_cast<u4>(unk)), rawId) * HASH_MULT2 + static_cast<u4>(NameKind::UNIQUE);
 }
 
-inline unsigned int _hash_mix_constant(unsigned int id) {
+inline unsigned int hashMixConstant(unsigned int id) {
     return id * HASH_MULT2 + static_cast<u4>(NameKind::CONSTANT);
 }
 
-inline unsigned int _hash_nameref(const GlobalState &gs, NameRef nref) {
+inline unsigned int hashNameRef(const GlobalState &gs, NameRef nref) {
     switch (nref.kind()) {
         case NameKind::UTF8:
             return _hash(nref.shortName(gs));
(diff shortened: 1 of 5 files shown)


core/Hashing.h
```diff
@@ -32,7 +24,7 @@ inline unsigned int _hash(std::string_view utf8) {
         res = mix(res, *it - '!'); // "!" is the first printable letter in ASCII.
         // This will help Latin1 but may harm utf8 multibyte
     }
-    return res * HASH_MULT2 + _NameKind2Id_UTF8(NameKind::UTF8);
+    return res * HASH_MULT2 + static_cast<u4>(NameKind::UTF8);
```
jvilk-stripe, line 27: I don't know why we mix this in here, but this hash function is our hashing function for all strings in Sorbet.

core/GlobalState.cc
```diff
@@ -28,6 +28,32 @@ using namespace std;
...
+
+inline unsigned int _hash_mix_unique(UniqueNameKind unk, unsigned int num, unsigned int rawId) {
+    return mix(mix(num, static_cast<u4>(unk)), rawId) * HASH_MULT2 + static_cast<u4>(NameKind::UNIQUE);
+}
+
+inline unsigned int _hash_mix_constant(unsigned int id) {
+    return id * HASH_MULT2 + static_cast<u4>(NameKind::CONSTANT);
+}
+
+inline unsigned int _hash_nameref(const GlobalState &gs, NameRef nref) {
```
elliottt, line 42: What do you think about renaming these to not be prefixed with an underscore, now that they don't live in a header?

elliottt (approved):

core/GlobalState.cc
jvilk-stripe: works for me!

PR #3969, opened 2021-02-03
Use xxHash for string hashing.

jvilk-stripe:
Use xxHash for string hashing. Use the 32-bit version which appears to be faster than the 64-bit variant for our use case.

There's no noticeable end-to-end speed improvement, but GlobalState::hash does seem slightly faster in profiles.

Pros:
* Delurks our existing handrolled string hashing algorithm.
* No regression. May even be slightly faster?
* Decreases hash collisions by >10X on UTF8 names in name table 🎉 
* Is ~5% faster than our old hash.

Cons:
* New dependency (although it's a simple 1 header file).

I investigated using xxHash's "streaming" API for GlobalState::hash so we can delurk `mix` (see commits reverted below), but found the following issues:

* Mixing in a uint to the hasher seemed to be more expensive for some reason. It's possible that I didn't get inlining right or could have combined multiple invocations easily, though. I noted a ~7% slowdown.
* The "state" struct for the hasher is not easily stack allocated. It has specific undocumented alignment requirements. The documentation wants you to heap allocate it with a createState function, which makes it less cheap for small transient hashing operations.
  * Note: This overhead is not present in the simpler non-streaming function used in this PR, so we could likely avoid it in all places where we quickly hash a few things together (e.g., name hashing).

I want to give it the streaming change a second try, but maybe in a follow-up PR.

### Motivation

Faster hash with lower hash collisions. And we're no longer hand-rolling a hash.

### Test plan

Covered by existing tests.


jvilk-stripe:

→ https://go/builds/bui_Iseq4gzjk2bI3L
→ https://go/builds/bui_IseqbaZZ6NQ8KT

jvilk-stripe:

→ https://go/builds/bui_IsgtoGm4k0XLos
→ https://go/builds/bui_Isgt64dxzwoDXA

jvilk-stripe:

→ https://go/builds/bui_IslNOdQKLNUiVg
→ https://go/builds/bui_IslNU50Vo5rUoj

jvilk-stripe:

→ https://go/builds/bui_IsxFlnmuflAJ6J
→ https://go/builds/bui_IsxFsdbGgjt53J

core/Hashing.h
```diff
@@ -12,19 +13,8 @@ inline unsigned int mix(unsigned int acc, unsigned int nw) {
...
-    auto it = utf8.begin();
-#ifndef HAS_SANITIZER
-#pragma clang loop unroll(enable)
-#endif
-    for (; it != end; it++) {
-        res = mix(res, *it - '!'); // "!" is the first printable letter in ASCII.
-        // This will help Latin1 but may harm utf8 multibyte
-    }
-    return res * HASH_MULT2 + static_cast<u4>(NameKind::UTF8);
+    // Take the lower 32-bits of the 64-bit hash. The 64-bit hash is faster than the 32-bit hash on 64-bit platforms.
```
jvilk-stripe, line 16: The author recommended this approach:
https://github.com/Cyan4973/xxHash/issues/453#issuecomment-696838445

third_party/xxhash.BUILD
```diff
@@ -0,0 +1,10 @@
+cc_library(
+    name = "xxhash",
+    srcs = glob(["xxhash.c"]),
+    hdrs = glob(["xxhash.h"]),
+    linkstatic = select({
+        "@com_stripe_ruby_typer//tools/config:linkshared": 0,
+        "//conditions:default": 1,
+    }),
+    visibility = ["//visibility:public"],
+)
```
jez, line 10: Can you make sure that jump to definition works with these functions?

jez:
cc @DarkDimius for hash algorithm changes

this looks fine to me but maybe poke dmitry

jez (approved):

third_party/xxhash.BUILD
jvilk-stripe: We have many automated tests for jump-to-def. What's your concern?

jez: Sorry, I meant clangd jump to def. I think you'll need to update some checked in files when you add new bazel targets.

jvilk-stripe:
Dmitry requests:

>can you check perf & key collision rate on name table?

I can do that before deciding to merge.

jvilk-stripe:
More context on the existing hash function:

>the prior function has been result of tunning pascal -> java -> scala -> dotty
> it’s super fast and good for english language
> it’s good at hashing lowercase letters numbers and uppercase letters
> it’s bad at other stuff

jvilk-stripe:

→ https://go/builds/bui_IuUEV3ON3f3rg2
→ https://go/builds/bui_IuUEGVqdZzGpxX

jvilk-stripe:
I updated the PR description. xxHash reduces utf8 name collisions in the name table by >10X. I'm looking into perf now; I'll put together a microbenchmark.

jvilk-stripe:
I've confirmed that xxhash reduces name table collisions for utf8 names by >10X. 🎉 

I'm going to look into throughput next.

jvilk-stripe:

→ https://go/builds/bui_Iua0iK5r1YB5C7
→ https://go/builds/bui_Iua01zqDgKnkLZ

third_party/xxhash.BUILD
jvilk-stripe: It works 👍

jvilk-stripe:
xxhash 32 w/ inline flag is ~5% faster than our old hash.

Faster and fewer collusions? Win/win.

PR #3971, opened 2021-02-03
Bump to clang-10

elliottt:
Upgrade to clang-10.0.0. By renaming our llvm_toolchain package we avoid the problems that showed up in the clang-9 upgrade, but it does mean that any tooling that depended on `llvm_toolchain` paths in the sandbox will need to be updated to point to `llvm_toolchain_10_0_0` instead.

### Motivation

Upgrading clang for better c++20 support, as well as access to a better version of lldb that will segfault less frequently.

### Test plan

See included automated tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,7 +12,7 @@ common:test --test_env=GTEST_INSTALL_FAILURE_SIGNAL_HANDLER=1
 ##
 ##  Custom toolchain.
 ##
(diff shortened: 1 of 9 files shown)


aprocter-stripe (approved):
LGTM. Will we have to update the build container for clang-10?

elliottt:
> LGTM. Will we have to update the build container for clang-10?

Nope, it's managed by bazel 👍

elliottt:
Closing this for now, as there isn't good coverage of prebuilt clang+llvm binaries for our build targets.

PR #3973, opened 2021-02-04
Typed Setters for raise_on_nil_write Props

ezra-stripe:
Make it so setters for raise_on_nil_write props no longer allow nil.

### Motivation

We currently enforce this via runtime checks, but it would be more helpful if the signature correctly implied the use case.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/test/types/struct.rb
+++ b/gems/sorbet-runtime/test/types/struct.rb
@@ -44,4 +44,85 @@ class Opus::Types::Test::StructValidationTest < Critic::Unit::UnitTest
       end
     end
   end
+
+  it "raises a TypeError setting nil on raise_on_nil_write prop" do
+    klass = Class.new(T::Struct) do
+      prop :foo, T.nilable(String), raise_on_nil_write: true
+    end
+    m = klass.new(foo: "baz")
+    assert_raises(TypeError) do
+      m.foo = nil
+    end
+  end
+
+  it "raises an error omitting raise_on_nil_write prop on initialization" do
+    klass = Class.new(T::Struct) do
+      prop :foo, T.nilable(String), raise_on_nil_write: true
+    end
+    assert_raises(ArgumentError) do
+      klass.new
+    end
+  end
+
+  it "raises an error explicitly setting nil on raise_on_nil_write prop on initialization" do
+    klass = Class.new(T::Struct) do
+      prop :foo, T.nilable(String), raise_on_nil_write: true
+    end
+    assert_raises(TypeError) do
+      klass.new(foo: nil)
+    end
+  end
+
+  # pre-existing behavior
+  it "allows explicit nil default for raise_on_nil_write prop" do
+    klass = Class.new(T::Struct) do
+      prop :foo, T.nilable(String), raise_on_nil_write: true, default: nil
+    end
+    m = klass.new
+    assert(m.foo.nil?)
+  end
+
+  it "does not allow setting nil with T.any(NilClass, U) with raise_on_nil_write props" do
+    klass = Class.new(T::Struct) do
+      prop :foo, T.any(NilClass, String), raise_on_nil_write: true
+    end
+    m = klass.new(foo: 'foo')
+    assert_raises(TypeError) do
+      m.foo = nil
+    end
+  end
+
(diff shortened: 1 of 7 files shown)


gems/sorbet-runtime/test/types/struct.rb
```diff
@@ -44,4 +44,38 @@ class Opus::Types::Test::StructValidationTest < Critic::Unit::UnitTest
...
+      m.foo = nil
+    end
+  end
+
+  # Failing - missing required prop
+  it "raises a TypeError omitting raise_on_nil_write prop on initialization" do
+    klass = Class.new(T::Struct) do
+      prop :foo, T.nilable(String), raise_on_nil_write: true
+    end
+    klass::new
```
jez, line 63: Can you write `klass.new` instead?

ezra-stripe:
I'm pretty happy with this change now, but please take a look and let me know if you want any specific changes.
It also still needs to go through internal testing against pay-server, but I'm not able to run those tests due to permissioning issues on the llvm repo.

jez:
I'm on leave this week; assigning someone else

test/testdata/rewriter/prop_raise_on_nil_write.rb
```diff
@@ -0,0 +1,5 @@
+# typed: strict
+class Thing
+  include T::Props
+  prop :foo, T.nilable(String), raise_on_nil_write: true
+end
```
elliottt, line 5: Could you add some cases that show the behavior of writes with nil values?

rewriter/Prop.cc
```diff
@@ -337,7 +351,13 @@ vector<ast::ExpressionPtr> processProp(core::MutableContext ctx, PropInfo &ret,
 
     // Compute the setter
     if (!ret.isImmutable) {
-        auto setType = ASTUtil::dupType(ret.type);
+        ast::ExpressionPtr setType;
+        // When we have raise_on_nil_write, the setType should be the non-nilable underlying type
+        if (ret.raiseOnNilWrite && isTNilable(ret.type)) {
+            setType = ASTUtil::dupType(getUnderlyingType(ret.type));
```
elliottt, line 357: Could you add a test case that shows that this won't detect `T.any(NilClass, <other-type>)`? That's the interpretation of `T.nilable(<other-type>)`, but this implementation is sensitive to the use of `T.nilable`.

elliottt (changes requested):
A few requests for additional test cases, but otherwise this makes sense to me!

test/testdata/rewriter/prop_raise_on_nil_write.rb
ezra-stripe: Added a few tests, both in the rewriter and also in the runtime for struct.

gems/sorbet-runtime/test/types/struct.rb
```diff
@@ -44,4 +44,65 @@ class Opus::Types::Test::StructValidationTest < Critic::Unit::UnitTest
...
+
+  it "allows setting non-nil values with T.any(NilClass, U) with raise_on_nil_write props" do
+    klass = Class.new(T::Struct) do
+      prop :foo, T.any(NilClass, String), raise_on_nil_write: true
+    end
+    m = klass.new
+    m.foo = 'foo'
+  end
+
+  it "allows setting nil values with T.nilable(NilClass) with raise_on_nil_write props, which is fine maybe?" do
```
ezra-stripe, line 101: This is a bit odd, but I'm not sure what the desired behavior would be in any case, so I suppose it is fine as is.

ezra-stripe:

→ https://go/builds/bui_J5PtpMh6D7L5vd
→ https://go/builds/bui_J5Pt9z30461rSz

ezra-stripe:

→ https://go/builds/bui_J6523xAqRbtRD9
→ https://go/builds/bui_J652faVsV38ela

ezra-stripe:
The new approach appears to generally work, and I'm now fixing pieces in pay-server so we can get a green build. More things might come up, we'll see.

ezra-stripe:
Now that we got it mostly working, we've decided to pursue removing `raise_on_nil_write` rather than make these changes. ;)

PR #3975, opened 2021-02-05
Use DispatchResult::merge in AndType::dispatchCall

elliottt:
Like #3974, `AndType::dispatchCall` was losing the `secondary` value from the left side of the tree.

### Motivation

Avoid potentially losing errors with `T.all` dispatch.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -46,9 +46,7 @@ DispatchResult OrType::dispatchCall(const GlobalState &gs, const DispatchArgs &a
     categoryCounterInc("dispatch_call", "ortype");
     auto leftRet = left.dispatchCall(gs, args.withSelfRef(left));
     auto rightRet = right.dispatchCall(gs, args.withSelfRef(right));
-    auto resultType = Types::any(gs, leftRet.returnType, rightRet.returnType);
-    return DispatchResult::merge(std::move(resultType), DispatchResult::Combinator::OR, std::move(leftRet),
-                                 std::move(rightRet));
+    return DispatchResult::merge(gs, DispatchResult::Combinator::OR, std::move(leftRet), std::move(rightRet));
 }
 
 TypePtr OrType::getCallArguments(const GlobalState &gs, NameRef name) const {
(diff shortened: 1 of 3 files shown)


elliottt:

→ https://go/builds/bui_ItSNi4ipyPdYiZ
→ https://go/builds/bui_ItSNe8wippyqj7

core/types/calls.cc
```diff
@@ -46,8 +46,9 @@ DispatchResult OrType::dispatchCall(const GlobalState &gs, const DispatchArgs &a
     categoryCounterInc("dispatch_call", "ortype");
     auto leftRet = left.dispatchCall(gs, args.withSelfRef(left));
     auto rightRet = right.dispatchCall(gs, args.withSelfRef(right));
-    return DispatchResult::merge(Types::any(gs, leftRet.returnType, rightRet.returnType),
-                                 DispatchResult::Combinator::OR, std::move(leftRet), std::move(rightRet));
+    auto resultType = Types::any(gs, leftRet.returnType, rightRet.returnType);
+    return DispatchResult::merge(std::move(resultType), DispatchResult::Combinator::OR, std::move(leftRet),
+                                 std::move(rightRet));
```
froydnj, line 51: I assume this was done for clarity and for symmetry with the `AndType` case?  +1 to pulling out some temporary variables to make the code clearer.

core/types/calls.cc
```diff
@@ -92,12 +93,10 @@ DispatchResult AndType::dispatchCall(const GlobalState &gs, const DispatchArgs &
...
     }
-    DispatchResult ret{Types::all(gs, leftRet.returnType, rightRet.returnType), move(leftRet.main),
-                       make_unique<DispatchResult>(move(rightRet)),
 
-                       DispatchResult::Combinator::AND};
-
-    return ret;
+    auto resultType = Types::all(gs, leftRet.returnType, rightRet.returnType);
+    return DispatchResult::merge(std::move(resultType), DispatchResult::Combinator::AND, std::move(leftRet),
+                                 std::move(rightRet));
```
froydnj, line 99: This is super-minor, and probably not worth it given that `DispatchResult::merge` is only liable to get used in two places, but WDYT about making `merge` instead just take the combinator type and the left/right results?  Then you can't get the result type wrong.

froydnj (approved):

core/types/calls.cc
elliottt: The result type is different for each case: it's `Types::any` for the OrType case and `Types::all` for the `AndType` case.

elliottt: Or were you suggesting casing on the combinator for figuring out which type to make?

core/types/calls.cc
elliottt: It was done to avoid potentially relying on a moved value, but also the clarity is good :)

PR #3976, opened 2021-02-05
mark Command self.call as unimplemented

froydnj:
Mark the synthesized `self.call` of the `Opus::Command` rewriter as unimplemented.

### Motivation

We don't need to keep this method around.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Command.cc
+++ b/rewriter/Command.cc
@@ -83,6 +83,8 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
         ast::MK::SyntheticMethod(call->loc, call->loc, call->name, std::move(newArgs), ast::MK::UntypedNil(call->loc));
     auto &mdef = ast::cast_tree_nonnull<ast::MethodDef>(selfCall);
     mdef.flags.isSelfMethod = true;
+    // This method is only for type checking. It doesn't actually exist at runtime, and instead all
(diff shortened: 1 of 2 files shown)


rewriter/Command.cc
```diff
@@ -81,7 +81,9 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
 
     auto selfCall =
         ast::MK::SyntheticMethod(call->loc, call->loc, call->name, std::move(newArgs), ast::MK::UntypedNil(call->loc));
-    ast::cast_tree<ast::MethodDef>(selfCall)->flags.isSelfMethod = true;
+    auto &mdef = ast::cast_tree_nonnull<ast::MethodDef>(selfCall);
+    mdef.flags.isSelfMethod = true;
+    mdef.flags.isUnimplemented = true;
```
jez, line 86:
```suggestion
    # This method is only for type checking. It doesn't actually exist at runtime, and instead all
    # Opus::Command subclasses inherit the `self.call` method from `Opus::Command` directly.
    mdef.flags.isUnimplemented = true;
```

jez (approved):
I wonder if we should call that flag "discardDef" or `keepDef = false` or something like that.

Unimplemented makes it seem like it still needs to be implemented.

PR #3977, opened 2021-02-05
rename Flags::isUnimplemented to Flags::discardDef, and make Flags more functional

froydnj:
This is a change suggested by jez on #3976 (first commit) and a change that I thought I might as well implement while I was in the neighborhood: making `Flags` more functional so we don't have to keep casting to `MethodDef` to set things.

If folks don't like the latter change, I can just drop it.

### Motivation

Better naming is more better.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -363,10 +363,10 @@ public:
         bool isSelfMethod : 1;
         bool isRewriterSynthesized : 1;
         bool isAttrReader : 1;
-        bool isUnimplemented : 1;
+        bool discardDef : 1;
 
         // In C++20 we can replace this with bit field initialzers
-        Flags() : isSelfMethod(false), isRewriterSynthesized(false), isAttrReader(false), isUnimplemented(false) {}
+        Flags() : isSelfMethod(false), isRewriterSynthesized(false), isAttrReader(false), discardDef(false) {}
     };
     CheckSize(Flags, 1, 1);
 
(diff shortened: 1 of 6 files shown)


ast/Trees.h
```diff
@@ -367,6 +367,27 @@ EXPRESSION(MethodDef) {
...
+        Flags markAttrReader() const {
+            Flags copy(*this);
+            copy.isAttrReader = true;
+            return copy;
+        }
+        Flags markDiscardable() const {
+            Flags copy(*this);
+            copy.discardDef = true;
+            return copy;
+        }
```
jez, line 390: I don't really get the point of this?

Why not just

- default argument to `MK::Method` like `flags = Flags()`
- initialize flags the same way as normal:

```
Flags flags;
flags.discardDef = true
MK::Method(..., flags)
```
?

jez (approved):
Second commit looks like a lot of boilerplate for little gain, i'd prefer to drop it.

ast/Trees.h
froydnj: Ah, that makes more sense.  I was thinking of builder pattern-type things, but this is probably better for C++ and Sorbet's general attitude of "struct fields are public".

jez: Yeah sorry I wasn't clear with what I wrote before; if you prefer the way it's done in this PR i don't oppose it strongly enough to overturn. My preference is only a slight preference.

PR #3978, opened 2021-02-08
WIP: Introduce a ConcurrentTask abstraction to replace most uses of multiplexjob

jvilk-stripe:
WIP: Introduce a ConcurrentTask abstraction to replace most uses of multiplexjob. Net negative LOC PR since it's much easier to specify parallel tasks now.

Features:
* **Threadsafe handling of queues and stack frames.** Unlike our previous handrolled loops, `ConcurrentTask` uses a barrier to wait for all worker threads to finish before letting the coordinator thread proceed, which avoids nasty race conditions. This change fixes some possible-but-unlikely race conditions in the code.
  * Bonus: Fixing this bug should resolve some "queue depth" bugs that occur with WorkerPool, since the queue should never get deeper than 1 + non-concurrent-task-uses-of-workerpool.
* **Timers for coordinator + worker threads.** It "just happens".
* **Need to collect counters from worker threads? That's a template argument.** We don't collect counters all of the time because it's not completely free, and batching is good. (Perhaps this should just be its own ConcurrentTask sometime.)

### Motivation

I'm tired of writing these by hand and hitting the same bugs over and over.

### Test plan

CI + perf testing on Stripe's codebase.

### DIFF
--- a/common/concurrency/ConcurrentTask.h
+++ b/common/concurrency/ConcurrentTask.h
@@ -3,11 +3,12 @@
 
 #include "absl/synchronization/barrier.h"
 #include "common/ConstExprStr.h"
+#include "common/Counters.h"
 #include "common/concurrency/ConcurrentQueue.h"
 #include "common/concurrency/WorkerPool.h"
 
 namespace sorbet {
-template <typename I, typename WorkerState> class ConcurrentTask {
+template <typename I, typename WorkerState, bool CollectCounters = false> class ConcurrentTask {
     std::shared_ptr<ConcurrentBoundedQueue<I>> inputq;
     ConstExprStr metricName;
     ConstExprStr workerMetricName;
@@ -30,6 +31,11 @@ public:
         Timer timeit(tracer, metricName);
 
         auto outputq = std::make_shared<BlockingBoundedQueue<WorkerState>>(size);
+        std::shared_ptr<BlockingBoundedQueue<CounterState>> counters;
+        if (CollectCounters) {
+            // If workers is empty, the main thread will run the job.
+            counters = std::make_shared<BlockingBoundedQueue<CounterState>>(std::max(workers.size(), 1));
+        }
 
         // Note: Cannot be stack allocated; see docs for absl::Barrier. The +1 is for the control thread.
         auto workerBarrier = new absl::Barrier(workers.size() + 1);
@@ -51,7 +57,12 @@ public:
                 if (count > 0) {
                     outputq->push(std::move(state), count);
                 }
+
+                if (CollectCounters) {
+                    counters->push(getAndClearThreadCounters(), 1);
+                }
             }
+
             // Prevent deadlock with an empty workerpool, where all logic runs on one thread.
             // After this LOC, it is no longer safe to use `this` or refer to any variables in the `run` stack frame.
             if (workers.size() > 0 && barrier->Block()) {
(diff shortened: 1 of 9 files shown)


jvilk-stripe:

→ https://go/builds/bui_IuToGSLvDQXgXy
→ https://go/builds/bui_IuTom5P02whOf5

resolver/resolver.cc
```diff
@@ -867,225 +858,206 @@ class ResolveConstantsWalk {
     }
 
     static vector<ast::ParsedFile> resolveConstants(core::GlobalState &gs, vector<ast::ParsedFile> trees,
```
jvilk-stripe, line 860: Note: I had to declare before defining since this code references `ResolveConstantsTaskState` which contains a `ResolveConstantsWalk` so it had to be declared _after_ `ResolveConstantsWalk`.

resolver/resolver.cc
```diff
@@ -2126,181 +2086,164 @@ class ResolveTypeMembersAndFieldsWalk {
         return tree;
     }
 
-    static vector<ast::ParsedFile> run(core::GlobalState &gs, vector<ast::ParsedFile> trees, WorkerPool &workers) {
-        Timer timeit(gs.tracer(), "resolver.type_params");
+    static vector<ast::ParsedFile> run(core::GlobalState &gs, vector<ast::ParsedFile> trees, WorkerPool &workers);
```
jvilk-stripe, line 2089: Same as `ResolveConstantsWalk`.

common/concurrency/ConcurrentTask.h
```diff
@@ -0,0 +1,108 @@
...
+
+#include "absl/synchronization/barrier.h"
+#include "common/ConstExprStr.h"
+#include "common/Counters.h"
+#include "common/concurrency/ConcurrentQueue.h"
+#include "common/concurrency/WorkerPool.h"
+
+namespace sorbet {
+template <typename I, typename WorkerState, bool CollectCounters = false> class ConcurrentTask {
+    std::shared_ptr<ConcurrentBoundedQueue<I>> inputq;
```
dmitry-stripe, line 12: these are pretty heavy templates, you may want to pimpl to save compilation time. your call

DarkDimius, line 12: these are pretty heavy templates, you may want to consider pimpl to save compilation time. your call

common/concurrency/ConcurrentTask.h
```diff
@@ -0,0 +1,108 @@
...
+                if (CollectCounters) {
+                    counters->push(getAndClearThreadCounters(), 1);
+                }
+            }
+
+            // Prevent deadlock with an empty workerpool, where all logic runs on one thread.
+            // After this LOC, it is no longer safe to use `this` or refer to any variables in the `run` stack
+            // frame.
+            if (workers.size() > 0 && barrier->Block()) {
+                delete barrier;
```
DarkDimius, line 72: can we do without raw pointer delete ops?

DarkDimius: can we make unique_ptr instead?

common/concurrency/ConcurrentTask.h
```diff
@@ -0,0 +1,108 @@
...
+                I input;
+                size_t count = 0;
+                WorkerState state;
+                for (auto result = inputq->try_pop(input); !result.done(); result = inputq->try_pop(input)) {
+                    if (result.gotItem()) {
+                        count++;
+                        processTask(std::move(input), state);
+                    }
+                }
+                if (count > 0) {
```
DarkDimius, line 59: this doesn't support streaming results, seems like it's intentional, right?

jvilk-stripe: Yes. I figured if we want to fold in streaming results we may need a slightly different abstraction anyway. Or another template argument.

common/concurrency/ConcurrentTask.h
jvilk-stripe: I'm not sure if I can. This is the pattern the absl docs want you to use. I fear that unique_ptr delete may race with worker threads waking up from block.

common/concurrency/ConcurrentTask.h
jvilk-stripe: I can look into this.

common/concurrency/ConcurrentTask.h
jvilk-stripe: I might be able to hide this in an RAII class.

PR #3983, opened 2021-02-09
Remove --dsl-plugins option, code, and documentation.

jvilk-stripe:
As discussed previously, remove --dsl-plugins option, code, and documentation. This change will make it easier for me to experiment with compressing ASTs in LSP during index. This was always marked as an experimental option, so we are going to remove it without a minor version bump.

### Motivation

Take a look at pipeline.cc. Our indexing (parsing) code is _significantly_ simpler now, with far fewer redundancies to handle the plugin / non-plugin case. This change has also found one bug caused directly by that redundancy -- indexOne() had invalid printing logic and an ENFORCE that didn't make sense.

### Test plan

CI.


main/pipeline/pipeline.cc
```diff
@@ -222,78 +225,6 @@ ast::ParsedFile indexOne(const options::Options &opts, core::GlobalState &lgs, c
     }
 }
 
-pair<ast::ParsedFile, vector<shared_ptr<core::File>>> emptyPluginFile(core::FileRef file) {
-    return {emptyParsedFile(file), vector<shared_ptr<core::File>>()};
-}
-
-pair<ast::ParsedFile, vector<shared_ptr<core::File>>>
-indexOneWithPlugins(const options::Options &opts, core::GlobalState &gs, core::FileRef file, ast::ExpressionPtr tree) {
```
jez, line 230: 🙏

jez (commented):
This is great. Thanks!

jez (approved):

alekstorm:
Will plugins be returning in the future in some form? Flexport found this feature quite useful for some of our internal metaprogramming, and came to more-or-less rely on it, even though it was marked experimental.

paracycle:
> Will plugins be returning in the future in some form? Flexport found this feature quite useful for some of our internal metaprogramming, and came to more-or-less rely on it, even though it was marked experimental.

@alekstorm I can't speak for the future plans of the Sorbet team, but despite having contributed most of the implementation of this plugin system, we had completely stopped using it at Shopify for over a year now. The drawbacks were very real for us: it was slowing down typechecking times by a lot and writing plugins was brittle since you didn't have access to the full context of the source match.

Instead, we built the [DSL RBI generation](https://github.com/Shopify/tapioca#generate-dsl-rbi-files) in Tapioca. They don't work during typechecking, you need to keep them in sync with your source code, and that is the only drawback. On the other hand, those generators work using runtime reflection and can generate very good RBI files for any metaprogramming you can think of.

Tapioca comes with [implementations of a few of them out of the box](https://github.com/Shopify/tapioca/blob/master/manual/generators.md), but it is easy to add custom ones in your app, as well.

alekstorm:
@paracycle Thanks, that does seem like a better solution. However, maybe I'm missing something obvious, but is there a way to target Tapioca at (Rails) app code, rather than a gem? Or would I be required to extract any DSL-using code into a gem? That would be a significant change to our project structure; we use DSLs pervasively.

Issue #3986, opened 2021-02-09
Investigate libmdbx as an alternative to lmdb


jvilk-stripe:
Sorbet uses lmdb for caching ASTs and hashes to disk. However, lmdb is incredibly persnickety. It has restrictions on which threads can do what, fails somewhat silently and catastrophically, and cannot resize itself -- so a full database causes crashes.

libmdbx is an alternative that claims better performance _and_ without the rough edges:

https://github.com/erthink/libmdbx

We should check it out.

PR #3988, opened 2021-02-10
[WIP] Replace lmdb with libmdbx

jvilk-stripe:
[WIP] Replace lmdb with libmdbx.

Since these libraries share a heritage, the replacement process was relatively straightforward.

Is there any perf change? Let's find out.

TODO:
* _Gracefully_ handle the case where the destination folder already has an _lmdb_ database.
* See if we can write DB files to separate files so there's no collision w/ lmdb so if folks switch between lmdb and libmdbx versions there's no disruption/crashes.

### Motivation

https://github.com/sorbet/sorbet/issues/3986

### Test plan

Covered by existing tests.


jvilk-stripe:

→ https://go/builds/bui_IvDcxNCvPCQ0zb
→ https://go/builds/bui_IvDdCW3z3V6Bm9

jvilk-stripe:
Looks like I need to either port dmitry's patches to lmdb (to avoid routines that invoke glibc malloc) or find a way to configure libmdbx to not use them.

PR #3989, opened 2021-02-10
Return length of data with kvstore reads, and avoid copying in writeString.

jvilk-stripe:
Return length of data with kvstore reads, and avoid copying in writeString.

We were stashing the length of strings but we didn't have to.

### Motivation

* Expose length of data. This is useful for my work on using compressed ASTs in LSP mode.
* Cleaner and faster code.

### Test plan

Covered by existing tests (see kvstore_test).

### DIFF
--- a/common/kvstore/KeyValueStore.cc
+++ b/common/kvstore/KeyValueStore.cc
@@ -127,7 +127,7 @@ OwnedKeyValueStore::OwnedKeyValueStore(unique_ptr<KeyValueStore> kvstore)
     // Writer thread may have changed; reset the primary transaction.
     refreshMainTransaction();
     {
-        if (read(OLD_VERSION_KEY)) { // remove databases that use old(non-string) versioning scheme.
+        if (read(OLD_VERSION_KEY).data != nullptr) { // remove databases that use old(non-string) versioning scheme.
             clear();
         }
         auto dbVersion = readString(VERSION_KEY);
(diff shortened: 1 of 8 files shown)


common/kvstore/KeyValueStore.cc
```diff
@@ -258,21 +262,15 @@ void OwnedKeyValueStore::clear() {
...
 string_view OwnedKeyValueStore::readString(string_view key) const {
     auto rawData = read(key);
-    if (!rawData) {
+    if (rawData.data == nullptr) {
         return string_view();
     }
-    size_t sz;
-    memcpy(&sz, rawData, sizeof(sz));
-    string_view result(((const char *)rawData) + sizeof(sz), sz);
+    string_view result((const char *)rawData.data, rawData.len);
```
jez, line 268: iiuc, the point of this PR is to delete these three `memcpy`, and the only reason why we were using `memcpy` before was to populate `sz`?

Can you remind me why we expect the `rawData.data` to live longer than `result`?

jez (approved):

common/kvstore/KeyValueStore.cc
jvilk-stripe: >iiuc, the point of this PR is to delete these three memcpy, and the only reason why we were using memcpy before was to populate sz?

The primary point of this PR is that I need to know the length of binary AST data stored in kvstore for my work on compressed ASTs. After changing the api of `read`, I realized I could simplify existing code and did so!

> Can you remind me why we expect the rawData.data to live longer than result?

The lifetime of `rawData.data` is tied to the database. When the database is closed, it's unmapped from memory and the memory is reclaimed.

Note that I'm not changing the lifetime of this `string_view` at all in this PR, I'm just removing size from the stored data.

PR #3990, opened 2021-02-10
[WIP] Checked annotation for structs

alexander-stripe:
This introduces the ability to write `checked(:tests)` on `T::Struct` definitions to avoid potentially expensive recursive type checking on struct fields at runtime:

```
class Foo < T::Struct
   checked(:tests)
   prop :foo, String
end

f = Foo.new(foo: 12) # only throws in tests
f.foo = 43 # only throws in tests
```

### Motivation
We profiled some of our code and found that adding `checked(:tests)` to `sig`s still resulted in lot's of runtime type checks. The code in question makes heavy use of structs, hence this PR.

### Test plan
I included some automated tests, let me know if I need to add more elsewhere.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
+++ b/gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
@@ -13,12 +13,13 @@ module T::Props
         params(
           klass: T.all(Module, T::Props::ClassMethods),
           prop: Symbol,
-          rules: T::Hash[Symbol, T.untyped]
+          rules: T::Hash[Symbol, T.untyped],
+          check_level: Symbol,
         )
         .returns(SetterProc)
         .checked(:never)
       end
-      def self.build_setter_proc(klass, prop, rules)
+      def self.build_setter_proc(klass, prop, rules, check_level)
         # Our nil check works differently than a simple T.nilable for various
         # reasons (including the `raise_on_nil_write` setting and the existence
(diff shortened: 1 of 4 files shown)


froydnj:
@jez has opinions on this and it's my EOD at the moment, so bouncing to him.

gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
```diff
@@ -78,33 +82,48 @@ def self.build_setter_proc(klass, prop, rules)
...
-            validate&.call(prop, val)
-            instance_variable_set(accessor_key, val)
           else
-            T::Props::Private::SetterFactory.raise_pretty_error(
-              klass,
-              prop,
-              non_nil_type,
-              val,
-            )
+            if T::Props::Private::SetterFactory.should_check?(check_level)
```
djudd-stripe, line 94: Have this if-statement and method call in the setter is likely to eat up a lot of potential gains (as well as having a cost even for folks not using `checked(:tests)`). 

Ideally we'd push the check to load time, ie outside this proc. Then you could just use `attr_writer` instead of using a proc at all. You'll have to worry about load order if you do that, though.

Whatever you do, I'd recommend adding/modifying some benchmarks (there's already a folder with a few).

alexander-stripe:
Closing this for now as it didn't have the desired impact.

PR #3997, opened 2021-02-12
WIP: What happens if we remove RLE from serialize?

jvilk-stripe:
WIP: What happens if we remove RLE from serialize?

### Motivation

Speed.

### Test plan

Covered by existing tests.

### DIFF
--- a/core/serialize/pickler.h
+++ b/core/serialize/pickler.h
@@ -6,7 +6,6 @@
 namespace sorbet::core::serialize {
 class Pickler {
     std::vector<u1> data;
-    u1 zeroCounter = 0;
 
 public:
     void putU4(u4 u);
@@ -19,7 +18,6 @@ public:
 
 class UnPickler {
     int pos;
-    u1 zeroCounter = 0;
     std::vector<u1> data;
 
 public:
--- a/core/serialize/serialize.cc
+++ b/core/serialize/serialize.cc
@@ -69,10 +69,6 @@ void Pickler::putStr(string_view s) {
 constexpr size_t SIZE_BYTES = sizeof(int) / sizeof(u1);
 
 vector<u1> Pickler::result(int compressionDegree) {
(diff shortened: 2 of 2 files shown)


jvilk-stripe:

→ https://go/builds/bui_IvwtvnCqIQ25Ep
→ https://go/builds/bui_IvwtRUs68O10ZO

froydnj:
Random connection: maybe the way to go is removing RLE *and* being smarter about how we pack integers so we do SIMD decoding.  I'm not sure if this would require us to change the serialization interface so we present more information in one pack/unpack call, or if we buffer things up.

Jeff Dean gave a talk with some ideas:

http://research.google.com/people/jeff/WSDM09-keynote.pdf

And people have expanded on those ideas:

https://lemire.me/blog/2012/09/12/fast-integer-compression-decoding-billions-of-integers-per-second/ + links

jvilk-stripe:
Nice, thanks for the links! My plan was to:

1. See how this (1) impacts the cache size on pay-server and (2) impacts compression performance.
2. If 1 and 2 are acceptable, go to town on making these methods faster.

This will be useful for phase 2.

jvilk-stripe:
Without RLE:

* The size of the compressed global state for Stripe's code grows by ~16%, which is a noticeable bump.
* Serialization time for global state is _~67% slower_. There is a marked change in the amount of time it takes to compress and a slight increase for how long it takes to write large amounts of data (vector resizing overhead?).

Some thoughts:

* We don't preallocate the unpickler vector... but I also don't think we can preallocate it large enough for Stripe's GlobalState object.
* I should run lzbench on payload w/ this change.
* Sorbet should maybe emit a metric for time spent in `Pickler::result` (compression).

jvilk-stripe:
My findings indicate that RLE is totally useful. As Nathan says, it may be good to look into other types of integer packing in the future.

PR #3998, opened 2021-02-12
model `when *<expr>` more accurately

froydnj:
### Motivation

The desugaring for this was not really accurate:

* The `to_a` call may fail, which was not correctly modeled;
* The check is looping over the splatted array, not comparing the splatted array to the pattern.

This version is (ideally) better.

### Test plan

See included automated tests.


ast/Helpers.h
```diff
@@ -175,8 +175,7 @@ class MK {
     }
 
     static ExpressionPtr Splat(core::LocOffsets loc, ExpressionPtr arg) {
-        auto to_a = Send0(loc, std::move(arg), core::Names::toA());
-        return Send1(loc, Constant(loc, core::Symbols::Magic()), core::Names::splat(), std::move(to_a));
+        return Send1(loc, Constant(loc, core::Symbols::Magic()), core::Names::splat(), std::move(arg));
```
elliottt, line 178: With this change, `x = *1` would go from `x = Magic::splat(1.to_a)` to `x = Magic::splat(1)`. Is that expected?

elliottt (changes requested):
I think this change makes sense, but we seem to be lacking changes for other uses of the array coercion operator. Could you add a few additional tests for things like:
```
x = *1
```

ast/Helpers.h
froydnj: Yes; `1.to_a` throws `NoMethodError`, while `*1` returns `[1]`.  We want the latter, not the former.

core/types/calls.cc
```diff
@@ -2175,6 +2175,30 @@ class Magic_selfNew : public IntrinsicMethod {
...
+        }
+
+        if (auto *ap = cast_type<AppliedType>(arg->type)) {
+            if (ap->klass == Symbols::Array() || ap->klass.data(gs)->derivesFrom(gs, Symbols::Array())) {
+                res.returnType = arg->type;
+                return;
+            }
+        }
+
+        res.returnType = Types::untypedUntracked();
```
elliottt, line 2198: I'm not sure that this intrinsic is returning the correct type. Here's an example from irb:
```
irb(main):001:0> xs = [1,2,3]
=> [1, 2, 3]
irb(main):002:0> ys = *xs
=> [1, 2, 3]
irb(main):003:0> ys.class
=> Array
irb(main):004:0> ys = *1
=> [1]
irb(main):005:0> ys.class
=> Array
```

froydnj: Hm, I guess the right thing to do here would be:

* Attempt to dispatch `arg.to_a`
* If that returns something, use that
* If that returns an error (or something non-arrayish?), then try to return a one-element `Array` (maybe untyped, depending?)

?  I tried a quick and dumb version of this and returned `Array[T.untyped]`, but that ran afoul of code in `callWithSplat` that was expecting `untyped` or `TupleType`, and I was not clever enough to see what introducing arrays there might do.  I guess if `callWithSplat` gets an array type, it should just treat that as untyped?

froydnj:

→ https://go/builds/bui_Izl0uB0Oh0x8HD
→ https://go/builds/bui_Izl0AvRDPhFwBz

froydnj:

→ https://go/builds/bui_IzlPUJec5CLIIn
→ https://go/builds/bui_IzlPp81Oi53Ilw

froydnj:

→ https://go/builds/bui_IzmVI5xflzPHkt
→ https://go/builds/bui_IzmV1YyoZKO8hs

froydnj:
Overview of changes:

* dispatching to `to_a` when determining the type of the splat
* adjusting the default type of `Magic.<splat>`, since it's not receiving an array anymore
* adding at least one bit of test (`test/testdata/infer/splat.rb` changes) to ensure that we test non-array things

froydnj:

→ https://go/builds/bui_J00pGEcpHVJcuy
→ https://go/builds/bui_J00pS1ffxrs42T

core/types/calls.cc
```diff
@@ -2175,6 +2175,47 @@ class Magic_selfNew : public IntrinsicMethod {
...
+            res.returnType = arg->type;
+            return;
+        }
+
+        if (auto *ap = cast_type<AppliedType>(arg->type)) {
+            if (ap->klass == Symbols::Array() || ap->klass.data(gs)->derivesFrom(gs, Symbols::Array())) {
+                res.returnType = arg->type;
+                return;
+            }
+        }
```
elliottt, line 2196: Since both `TupleType` and `Array` have `to_a` sigs, I think neither of these cases are necessary.

ast/desugar/Desugar.cc
```diff
@@ -1886,14 +1886,28 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
...
+                            // TODO(froydnj): use the splat's var directly so we can elide the
+                            // coercion to an array where possible.
+                            auto splat = node2TreeImpl(dctx, std::move(cnode));
+                            auto patternloc = splat.loc();
+                            test = MK::Send2(patternloc, std::move(recv), core::Names::checkMatchArray(),
+                                             std::move(local), std::move(splat));
                         } else {
-                            test = std::move(ctree);
+                            ExpressionPtr ctree;
+                            ctree = node2TreeImpl(dctx, std::move(cnode));
```
elliottt, line 1902:
```suggestion
                            auto ctree = node2TreeImpl(dctx, std::move(cnode));
```

elliottt (approved):
I had a few suggestions, but otherwise this looks great!

PR #4000, opened 2021-02-13
Implement a monopackage-compatible non_forcing_is_a?

aisamanra:
### Motivation
This implements barebones logic to allow `non_forcing_is_a?` to work based on the current monopackage implementation of runtime packages.

### Test plan

See included automated tests. Added a few tests for this case.

### DIFF
--- a/gems/sorbet-runtime/lib/types/non_forcing_constants.rb
+++ b/gems/sorbet-runtime/lib/types/non_forcing_constants.rb
@@ -2,16 +2,24 @@
 # typed: strict
 
 module T::NonForcingConstants
+  T::Sig::WithoutRuntime.sig {returns(T::Boolean)}
+  private_class_method def self.stripe_packages_enabled?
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/non_forcing_constants.rb
```diff
@@ -2,18 +2,26 @@
 # typed: strict
 
 module T::NonForcingConstants
+  T::Sig::WithoutRuntime.sig {returns(T::Boolean)}
+  private_class_method def self.stripe_packages_enabled?
```
aisamanra, line 6: This is mostly defined separately so we can mock it in the tests below.

aprocter-stripe (approved):
:four: :zero: :zero: :zero: 🎊 🎉 

Looks like a safe change to me, so stamping. Don't have much context on `StripePackages`, so please feel free to seek a second review from someone who does, if you think that's necessary.

PR #4002, opened 2021-02-13
Hashing spends ~15% of its time in GlobalState's constructor incrementing an atomic integer.

jvilk-stripe:
Hashing spends ~15% of its time in GlobalState's constructor incrementing an atomic integer.

This integer corresponds to GlobalStateId, which is not semantically important for hashing... so let's not do that? :)

Introduces a new static method on GlobalState for creating empty GlobalStates for hashing.

### Motivation

Speed++.

### Test plan

Covered by existing tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -6,6 +6,7 @@
 #include "core/NameHash.h"
 #include "core/Names.h"
 #include "core/Names_gen.h"
+#include "core/NullFlusher.h"
 #include "core/Types.h"
 #include "core/Unfreeze.h"
 #include "core/errors/errors.h"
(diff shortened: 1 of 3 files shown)


jvilk-stripe:

→ https://go/builds/bui_Iw99mgzHK7WkK4
→ https://go/builds/bui_Iw99C2Ldvqctay

core/GlobalState.cc
```diff
@@ -98,6 +103,16 @@ GlobalState::GlobalState(shared_ptr<ErrorQueue> errorQueue, shared_ptr<lsp::Type
     ENFORCE((namesByHashSize & (namesByHashSize - 1)) == 0, "namesByHashSize is not a power of 2");
 }
 
+unique_ptr<GlobalState> GlobalState::makeEmptyGlobalStateForHashing(spdlog::logger &logger) {
+    // Note: Private constructor.
+    unique_ptr<GlobalState> rv(
+        new GlobalState(make_shared<core::ErrorQueue>(logger, logger, make_shared<core::NullFlusher>()),
+                        make_shared<lsp::TypecheckEpochManager>(), 999));
```
froydnj, line 110: WDYT about making this `INT_MIN` or something that's not likely to be a "real" id?  (-1 is used elsewhere for other purposes.)  IIUC, the id isn't even mixed in with the hash, so the value shouldn't matter much...but an unusual negative number might stand out a bit more in a debugging session.

froydnj (approved):

core/GlobalState.cc
jvilk-stripe: Works for me 👍

Issue #1487, opened 2019-08-04
sorbet-static has "Illegal instruction" on AMD Phenom II


kolAflash:
CPU: AMD Phenom(tm) II X4 955 Processor (model 4, stepping 2)
OS: openSUSE-LEAP-15.1

```
$ gem install --user sorbet-static
$ ~/.gem/ruby/2.5.0/gems/sorbet-static-0.4.4573-x86_64-linux/libexec/sorbet
Illegal instruction
```

I guess this is somehow caused, by how the binaries for https://rubygems.org/gems/sorbet are being build.

I tried to compile sorbet-static myself, but that failed due to some bazel errors. See: #1486

DarkDimius:
Sorbet by default gets compiled with support for recent CPU instructions.
The following line needs to be modified to make it stop:
https://github.com/sorbet/sorbet/blob/703e91a3ea002ffd7c9a700396449658f8a04209/.bazelrc#L29

DarkDimius:
As this is the only time we've seen it mentioned, we're not going to invest effort into this now.

if you `dmesg` it should tell you what instruction wasn't supported, I'd be curious to know which one you ran into.

kolAflash:
> [...]
> if you `dmesg` it should tell you what instruction wasn't supported, I'd be curious to know which one you ran into.

dmesg output:
`trap invalid opcode ip:459978 sp:7ffd19f5be80 error:0 in sorbet[3a0000+8ff000]`

If there's no other option, could you consider providing builds without `--copt=-march=sandybridge`?

fabioperrella:
Hi! I'm having a similar issue in Ubuntu 16.04:

```
$ bundle exec srb typecheck -e 'puts "Hello, world!"'
/home/fabio/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/sorbet-0.4.4761/bin/srb: line 46: 24128 Illegal instruction     (core dumped) "${sorbet[0]}" "${args[@]}"
```

```
$ dmesg | tail -n 1
[ 2729.592603] traps: sorbet[24128] trap invalid opcode ip:504d68 sp:7fff6de8ff80 error:0 in sorbet[44a000+920000]
```

I tried with ruby 2.6.3 and I got the same error.

incazteca:
Hi,

Just wanted to document that I've also been affected by this issue:
```
[...]

PR #4003, opened 2021-02-16
[WIP] serialize integers by grouping, rather than by individual varints

froydnj:
This was a weekend experiment to see how much better "group varint encoding" (https://github.com/facebook/folly/blob/master/folly/docs/GroupVarint.md) might work for Sorbet's serialization.  The above links to the original presentation of this idea which suggests decoding is ~2x faster; depending on the data, it might also be smaller as well.

Good news first: Sorbet's (uncompressed) payload gets ~2% smaller with these changes, with minimal changes to the compressed size.

There are a lot of known problems, and probably some unknown ones as well:

* No implementation for non-SSSE3 platforms (i.e. wasm)
* The SIMD bits were cut-and-pasted from Facebook's implementation (above) and need to be cleaned up
* The current implementation of decoding is almost certainly going to get flagged by sanitizers
* The groupings chosen are probably not optimal for some of the AST nodes (maybe even some of the payload-related things?)
* I would really like to know why 8d5e526 is such a loss
* No testing on whether this is actually a speed win or not
* Some of the names probably need to be cleaned up (e.g. `s/serialize/pickle/gi`)
* No idea how this does for the AST nodes

cc @jvilk-stripe

### Motivation

--space, ++speed.

### Test plan

See included automated tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -33,6 +33,7 @@ build --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
 # Developer laptops run skylake, devboxes run skylake-avx512
 # however some AWS instances in our fleet still run Sandy Bridge (Skylake predecessor)
 build --copt=-march=sandybridge
+build --host_copt=-march=sandybridge
 
 build --copt=-fno-omit-frame-pointer
 
--- a/core/serialize/pickler.h
+++ b/core/serialize/pickler.h
@@ -13,6 +13,9 @@ public:
     void putU1(const u1 u);
     void putS8(const int64_t i);
     void putStr(std::string_view s);
+
+    void putU4Group(u4, u4, u4, u4);
+
     std::vector<u1> result(int compressionDegree);
     Pickler() = default;
 };
@@ -27,6 +30,9 @@ public:
     u1 getU1();
     int64_t getS8();
     std::string_view getStr();
+
+    void getU4Group(u4*, u4*, u4*, u4*);
+
     explicit UnPickler(const u1 *const compressed, spdlog::logger &tracer);
 };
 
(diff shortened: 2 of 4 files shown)


jvilk-stripe:
>The SIMD bits were cut-and-pasted from Facebook's implementation (above) and need to be cleaned up

Could you use streamvbyte as a library? https://github.com/lemire/streamvbyte

I poked around at applying it this weekend. It requires haswell or above. I couldn't find any evidence that AWS uses anything older than haswell in EC2 though; do we really need SB support?

jvilk-stripe:
> No implementation for non-SSSE3 platforms (i.e. wasm)

I think we just need a simple abstraction in front of it with an `-emscripten` backend that emulates support or provides ~no optimization.

froydnj:
> > The SIMD bits were cut-and-pasted from Facebook's implementation (above) and need to be cleaned up
> 
> Could you use streamvbyte as a library? https://github.com/lemire/streamvbyte

Ooh, that looks useful, and better than some copy-pasted code.

> I poked around at applying it this weekend. It requires haswell or above. I couldn't find any evidence that AWS uses anything older than haswell in EC2 though; do we really need SB support?

There are comments in `.bazelrc` suggesting that sandybridge is required:

https://github.com/sorbet/sorbet/blob/482f30e3bc0fa29ccb294f85611bc67b2f29c2fd/.bazelrc#L33-L35

Perhaps those are outdated?  There have been a (very) few complaints in at least #1487 (maybe other places) about the ISA support requirements.  OTOH, Haswell is seven years old at this point, so maybe we just bite the bullet...

jvilk-stripe:
> Perhaps those are outdated?

Yeah, I'd want to know which machines were implicated there. https://aws.amazon.com/intel/ documents the intel families used in EC2 and nothing is older than Haswell. I don't know if that page is a comprehensive document though!

froydnj:
> > No implementation for non-SSSE3 platforms (i.e. wasm)
> 
> I think we just need a simple abstraction in front of it with an `-emscripten` backend that emulates support or provides ~no optimization.

I should have been more complete here: it's wasm + the compilation host + non-x86 (and I don't think we want to shut out arm64) -- unless we want to declare that you can only compile/run Sorbet on Haswell+-era machines.

PR #4005, opened 2021-02-17
WIP: Speed up pickling/unpickling numbers with prefix varint packing.

jvilk-stripe:
WIP: Speed up pickling and unpickling numbers with prefix varint packing.

Removes the zero special-case, although we can add it back in if it seems useful.

Intent is to _speed up_ encoding and decoding.

cc @froydnj who is looking at more complex SIMD-based approaches approaches

### Motivation

Serialization and deserialization is still slower than I'd like. Maybe this will help?

### Test plan

Covered by existing tests.

### DIFF
--- a/core/serialize/pickler.h
+++ b/core/serialize/pickler.h
@@ -6,7 +6,6 @@
 namespace sorbet::core::serialize {
 class Pickler {
     std::vector<u1> data;
-    u1 zeroCounter = 0;
 
 public:
     void putU4(u4 u);
@@ -19,7 +18,6 @@ public:
 
 class UnPickler {
     int pos;
-    u1 zeroCounter = 0;
     std::vector<u1> data;
 
 public:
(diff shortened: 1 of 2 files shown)


jvilk-stripe:

→ https://go/builds/bui_IxdoOeYWLFZoxk
→ https://go/builds/bui_Ixdo6o2r61gAZs

jvilk-stripe:
Hah, I think this might be _slower_. I need to put together a microbenchmark so I can iterate on this faster.

jvilk-stripe:
After careful microbenchmarking I've concluded that this will never be faster than our current approach unless we do tricks to pack multiple numbers at once. This approach actually has one additional step of overhead, counting bits, that is on the critical path to the common 1 byte case. So, I'm closing this out.

Interesting side-note: On Stripe's codebase, this change, which removes zero counting, barely increases the size of the cache (by 0.1%).

PR #4006, opened 2021-02-17
fix ENFORCE when dealing with applied types in Array#zip

froydnj:
Fix crash when `Array#zip` is used with non-`Array` types.

### Motivation

Sorbet should not crash.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2673,8 +2673,7 @@ public:
         for (auto arg : args.args) {
             auto argTyp = arg->type;
             if (auto *ap = cast_type<AppliedType>(argTyp)) {
-                ENFORCE(ap->klass == Symbols::Array() || ap->klass.data(gs)->derivesFrom(gs, Symbols::Array()) ||
-                        ap->klass == Symbols::Enumerable() ||
+                ENFORCE(ap->klass == Symbols::Enumerable() ||
                         ap->klass.data(gs)->derivesFrom(gs, Symbols::Enumerable()));
                 ENFORCE(!ap->targs.empty());
                 unwrappedElems.emplace_back(Types::any(gs, ap->targs.front(), Types::nilClass()));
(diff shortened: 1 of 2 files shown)


elliottt (approved):
Is it necessary to still check if the symbol derives from `Array`? I think that would be covered by the check for whether or not it derives from `Enumerable` now.

froydnj:
> Is it necessary to still check if the symbol derives from `Array`? I think that would be covered by the check for whether or not it derives from `Enumerable` now.

Good call, change made, tests pass locally.  :tada:

jez:
This change doesn't go far enough. Here is a test case that still fails, even after merging this PR:

```ruby
# typed: true

xs = T::Array[Integer].new

class Box
  extend T::Generic
  Elem = type_member
end

xs.zip(Box[Integer].new)
```

elliottt:
> @froydnj This change doesn't go far enough. Here is a test case that still fails, even after merging this PR:
> 
> ```ruby
> # typed: true
> 
> xs = T::Array[Integer].new
> 
> class Box
>   extend T::Generic
>   Elem = type_member
> end
> 
> xs.zip(Box[Integer].new)
> ```

This one fails at runtime as well, what's the expected behavior?

```
% ruby test.rb
        1: from test.rb:12:in `<main>'
test.rb:12:in `zip': wrong argument type Box (must respond to :each) (TypeError)
```

froydnj:
> This one fails at runtime as well, what's the expected behavior?

I presume the right thing is to not `ENFORCE` this, but gracefully return from the `apply` call and let Sorbet flag the error for the user.

elliottt:
> > This one fails at runtime as well, what's the expected behavior?
> 
> I presume the right thing is to not `ENFORCE` this, but gracefully return from the `apply` call and let Sorbet flag the error for the user.

`:facepalm:` yep, that makes sense. I ran the example on sorbet.run and the error was what I expected, so I wasn't sure what needed to change :)

PR #4009, opened 2021-02-17
more tightly bounded checks when searching for methods in ancestors

froydnj:
Use the optional second arg to `method_defined?` and friends to bound ancestor chain searches.

### Motivation

In attempting to turn on `final` method checks for Stripe's codebase, we're seeing various timeouts when these checks are turned on.  I'm not 100% certain that this change solves things, but it does seem like a reasonable place to start.  By default, we're going to check the entire ancestor chain for each invocation of `method_defined?` for each ancestor we're examining.  This scales badly; we can do better by limiting the check to the specific ancestor we're looking at.  (It strikes me that we should also not be checking `protected_method_defined?`, since `method_defined?` is defined to examine public and protected methods, but that is a change for a separate PR, I think.)

This argument is a Ruby 2.6+ addition; I don't believe we support <= Ruby 2.5, but happy to hear otherwise.

### Test plan

I believe we already have tests that exercise this code, e.g.

https://github.com/sorbet/sorbet/blob/03f5d54c917f2371ba1ff304a3812a12257b4035/gems/sorbet-runtime/test/types/final_method.rb#L365-L384

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -100,17 +100,15 @@ module T::Private::Methods
     if !module_with_final?(target)
       return
     end
-    # use reverse_each to check farther-up ancestors first, for better error messages. we could avoid this if we were on
-    # the version of ruby that adds the optional argument to method_defined? that allows you to exclude ancestors.
-    target_ancestors.reverse_each do |ancestor|
+    target_ancestors.each do |ancestor|
       source_method_names.each do |method_name|
         # the usage of method_owner_and_name_to_key(ancestor, method_name) instead of
         # method_to_key(ancestor.instance_method(method_name)) is not (just) an optimization, but also required for
         # correctness, since ancestor.method_defined?(method_name) may return true even if method_name is not defined
         # directly on ancestor but instead an ancestor of ancestor.
-        if (ancestor.method_defined?(method_name) ||
-            ancestor.private_method_defined?(method_name) ||
-            ancestor.protected_method_defined?(method_name)) &&
+        if (ancestor.method_defined?(method_name, false) ||
+            ancestor.private_method_defined?(method_name, false) ||
+            ancestor.protected_method_defined?(method_name, false)) &&
            final_method?(method_owner_and_name_to_key(ancestor, method_name))
           definition_file, definition_line = T::Private::Methods.signature_for_method(ancestor.instance_method(method_name)).method.source_location
           is_redefined = target == ancestor
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -103,9 +103,9 @@ module T::Private::Methods
     target_ancestors.each do |ancestor|
       source_method_names.each do |method_name|
         # the usage of method_owner_and_name_to_key(ancestor, method_name) instead of
-        # method_to_key(ancestor.instance_method(method_name)) is not (just) an optimization, but also required for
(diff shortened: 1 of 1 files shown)


jez:
I think this would be the first thing that explicitly breaks support for 2.5

froydnj:
Ruby 2.5 is EOL at the end of next month: https://www.ruby-lang.org/en/downloads/branches/

froydnj:
Unfortunately, something seems to be broken with the contract of `method_defined?` (or I am misunderstanding what its contract is).  Testcase:

```ruby
require 'benchmark'

def chain(n)
  n.times.reduce(Object) do |superclass, i|
    sym = "foo#{i}".to_sym
    Class.new(superclass) do
      def sym; end
    end
  end
end

def search(klass, sym, inherit=true)
  klass.ancestors.each do |a|
    if a.method_defined?(sym, inherit)
      return true
    end
  end

  return false
end

POWER = 14
width = "#{2**POWER} ".length

Benchmark.bm(width) do |x|
  [true, false].each do |inherit|
    puts "Searching with inherit=#{inherit}"
    POWER.times do |i|
      chain_length = 2**i
      x.report("#{chain_length}") { search(chain(chain_length), :frob, inherit) }
    end
  end
end
```

Output:

```
froydnj@pyro:~$ ./ruby/bin/ruby --version
ruby 2.7.3p146 (2020-12-31 revision d069339110) [x86_64-linux]
froydnj@pyro:~$ ./ruby/bin/ruby ancestors.rb
             user     system      total        real
Searching with inherit=true
1        0.000000   0.000000   0.000000 (  0.000014)
2        0.000000   0.000000   0.000000 (  0.000009)
4        0.000000   0.000000   0.000000 (  0.000014)
8        0.000000   0.000000   0.000000 (  0.000034)
16       0.000000   0.000000   0.000000 (  0.000034)
32       0.000000   0.000000   0.000000 (  0.000063)
64       0.000000   0.000000   0.000000 (  0.000139)
128      0.000000   0.004000   0.004000 (  0.000797)
256      0.000000   0.000000   0.000000 (  0.001831)
512      0.000000   0.004000   0.004000 (  0.002942)
1024     0.008000   0.000000   0.008000 (  0.010423)
2048     0.048000   0.004000   0.052000 (  0.048818)
4096     0.276000   0.000000   0.276000 (  0.276298)
8192     1.780000   0.000000   1.780000 (  1.783581)
Searching with inherit=false
1        0.000000   0.000000   0.000000 (  0.000015)
2        0.000000   0.000000   0.000000 (  0.000010)
4        0.000000   0.000000   0.000000 (  0.000016)
8        0.000000   0.000000   0.000000 (  0.000016)
16       0.000000   0.000000   0.000000 (  0.000032)
32       0.000000   0.000000   0.000000 (  0.000057)
64       0.000000   0.004000   0.004000 (  0.000135)
128      0.000000   0.000000   0.000000 (  0.000315)
256      0.000000   0.000000   0.000000 (  0.000941)
512      0.004000   0.000000   0.004000 (  0.002911)
1024     0.008000   0.000000   0.008000 (  0.009907)
2048     0.044000   0.000000   0.044000 (  0.042939)
4096     0.276000   0.000000   0.276000 (  0.277501)
8192     2.020000   0.000000   2.020000 (  2.019376)
```

These numbers show us two things:

1. The `inherit=true` case is quadratic, as expected.
2. The `inherit=false` case is *not* linear and is even slightly **worse** than the `inherit=true` case.  :exploding_head: 

Both of these things seem like real problems.

froydnj:
(I mean, I really hope we're not encountering ancestor chains that are 2000+ entries deep in the wild, but you never know?)

froydnj:
After discussion with @elliottt , I don't think this is going anywhere, so closing.

PR #4012, opened 2021-02-18
Move NameRef and SymbolRef kind bits to end for better varint packing.

jvilk-stripe:
Move NameRef and SymbolRef kind bits to end for better varint packing.

### Motivation

We serialize `u4`s using variable-length integers (varints) to reduce the size of the payload baked into Sorbet and stored on disk in the cache. Varints are really nice for small numbers (numbers up to 2^7-1 fit in 1 byte!) but have overhead for large numbers (28+ bits require 5 bytes since it stores 7 bits of data per byte).

We store SymbolRef/NameRefs as u4s in the cache using their rawId, which encodes their kind and their index in that kind's table. Storing the kind bits in the _upper_ bits of rawId ensures that ~all of these references are stored inefficiently in the cache (5 bytes instead of 1 for small IDs!). I hadn't thought of this downside when initially engineering them this way. 

A secondary consequence of this inefficiency is that it's _slower_ to deserialize and serialize an integer from 5 bytes rather than 1 or 2. I am very interested in optimizing deserialization for my compressed AST work, which is the primary motivation of this work.

Fortunately, the solution is simple: Move the kind bits to the _end_ of the rawId. This solution shrinks payload by ~4% (0.1MB) and Stripe's cache directory (which contains ASTs) by ~2%. I do not see any noticeable changes in end-to-end runtime.

### Test plan

See included automated tests.


jvilk-stripe:

→ https://go/builds/bui_IyHuV66Tgiv87k
→ https://go/builds/bui_IyHuTbGRmte7Gd

core/SymbolRef.h
```diff
@@ -111,7 +111,7 @@ class SymbolRef final {
     // Stores the symbol's Kind and Index. Kind occupies the upper bits.
```
froydnj, line 111: Please change this comment to align with the code.

test/cli/simple-package/simple-package.out
```diff
@@ -2,15 +2,15 @@ foo/foo.rb:11: Unable to resolve constant `Bard` https://srb.help/5002
...
   Autocorrect: Use `-a` to autocorrect
-    foo/foo.rb:11: Replace with `Project::Bar`
+    foo/foo.rb:11: Replace with `Project::Bar::Bar`
     11 |  Project::Bar::Bard
           ^^^^^^^^^^^^^^^^^^
-    foo/__package.rb: Did you mean: `Project::Bar::Bar`?
     foo/__package.rb:5: Did you mean: `Project::Bar`?
      5 |  import Project::Bar
                  ^^^^^^^^^^^^
+    foo/__package.rb: Did you mean: `Project::Bar::Bar`?
```
froydnj, line 15: These changes worry me.  Are we sorting on the raw id somewhere, and reordering the bits has changed things?  Is something reaching into the innards of `NameRef`/`SymbolRef` and didn't get updated for this change?  Or something else?

core/NameRef.h
```diff
@@ -71,12 +71,13 @@ class NameRef final : private DebugOnlyCheck<NameRefDebugCheck> {
     // NameKind takes up this many bits in _id.
     static constexpr u4 KIND_BITS = 2;
     static constexpr u4 ID_BITS = 32 - KIND_BITS;
-    static constexpr u4 ID_MASK = (1 << ID_BITS) - 1;
+    static constexpr u4 KIND_MASK = (1 << KIND_BITS) - 1;
+    static constexpr u4 MAX_ID = (1 << ID_BITS) - 1;
```
froydnj, line 75: Thank you for fixing up the "mask is max" nomenclature here.

froydnj (approved):
LGTM, but I would really like it if the test expectation changes were some combination of investigated/understood/fixed before this landed.

core/SymbolRef.h
jvilk-stripe: Can do.

test/cli/simple-package/simple-package.out
jvilk-stripe: Yes, `findMemberFuzzyMatchConstant` sorts on rawId:

https://github.com/sorbet/sorbet/blob/03f5d54c/core/Symbols.cc#L569

That's what is used for `Did you mean:` suggestions.

jvilk-stripe: (The sorting is done for determinism reasons, not for semantic reasons.)

froydnj: Works for me, thanks for clearing this up!

PR #4013, opened 2021-02-18
Fix inline kwargs parsing to not consider all inline hashes kwargs

elliottt:
Hashes that are implicitly constructed in method sends like `foo(10 => 20)` were being treated as keyword arguments, which could lead to crashes in `dispatchCallSymbol` and further down the pipeline.

### Motivation

Fixing problems with keyword arguments that can lead to crashes.

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -597,6 +597,15 @@ public:
     }
 };
 
+bool isInlininableKwargPair(parser::Node *node) {
+    auto *pair = parser::cast_node<parser::Pair>(node);
+    if (pair == nullptr) {
+        return false;
+    }
+
+    return parser::isa_node<parser::Symbol>(pair->key.get());
+}
+
 ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) {
     try {
         if (what.get() == nullptr) {
@@ -631,17 +640,16 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                     if (!send->args.empty()) {
                         // Deconstruct the kwargs hash in the last argument if it's present.
                         if (auto *hash = parser::cast_node<parser::Hash>(send->args.back().get())) {
-                            if (hash->kwargs) {
-                                // hold a reference to the node, and remove it from the back fo the send list
+                            if (hash->inline_) {
+                                // hold a reference to the node, and remove it from the back of the send list
                                 auto node = std::move(send->args.back());
                                 send->args.pop_back();
 
                                 parser::NodeVec elts;
 
                                 // skip inlining the kwargs if there are any non-key/value pairs present
-                                if (absl::c_any_of(hash->pairs, [](auto &arg) {
-                                        return !parser::isa_node<parser::Pair>(arg.get());
-                                    })) {
+                                if (!absl::c_all_of(hash->pairs,
+                                                    [](auto &node) { return isInlininableKwargPair(node.get()); })) {
                                     elts.emplace_back(std::move(node));
                                 } else {
                                     // inline the hash into the send args
@@ -758,13 +766,12 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                     if (numPosArgs > 0) {
                         // Deconstruct the kwargs hash in the last argument if it's present.
                         if (auto *hash = parser::cast_node<parser::Hash>(send->args.back().get())) {
-                            if (hash->kwargs) {
+                            if (hash->inline_) {
                                 numPosArgs--;
 
                                 // skip inlining the kwargs if there are any non-key/value pairs present
-                                if (absl::c_all_of(hash->pairs, [](auto &arg) {
-                                        return parser::isa_node<parser::Pair>(arg.get());
-                                    })) {
+                                if (absl::c_all_of(hash->pairs,
+                                                   [](auto &arg) { return isInlininableKwargPair(arg.get()); })) {
                                     // hold a reference to the node, and remove it from the back fo the send list
                                     auto node = std::move(send->args.back());
                                     send->args.pop_back();
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -597,15 +597,6 @@ public:
     }
 };
 
-bool isInlininableKwargPair(parser::Node *node) {
-    auto *pair = parser::cast_node<parser::Pair>(node);
-    if (pair == nullptr) {
-        return false;
-    }
-
-    return parser::isa_node<parser::Symbol>(pair->key.get());
-}
-
 ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) {
     try {
         if (what.get() == nullptr) {
@@ -640,16 +631,21 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                     if (!send->args.empty()) {
                         // Deconstruct the kwargs hash in the last argument if it's present.
                         if (auto *hash = parser::cast_node<parser::Hash>(send->args.back().get())) {
-                            if (hash->inline_) {
+                            if (hash->kwargs) {
                                 // hold a reference to the node, and remove it from the back of the send list
                                 auto node = std::move(send->args.back());
                                 send->args.pop_back();
 
                                 parser::NodeVec elts;
 
-                                // skip inlining the kwargs if there are any non-key/value pairs present
-                                if (!absl::c_all_of(hash->pairs,
-                                                    [](auto &node) { return isInlininableKwargPair(node.get()); })) {
+                                // skip inlining the kwargs if there are any kwsplat nodes present
+                                if (absl::c_any_of(hash->pairs, [](auto &node) {
+                                        // the parser guarantees that if we see a kwargs hash it only contains pair or
+                                        // kwsplat nodes
+                                        ENFORCE(parser::isa_node<parser::Kwsplat>(node.get()) ||
+                                                parser::isa_node<parser::Pair>(node.get()));
+                                        return parser::isa_node<parser::Kwsplat>(node.get());
+                                    })) {
                                     elts.emplace_back(std::move(node));
                                 } else {
                                     // inline the hash into the send args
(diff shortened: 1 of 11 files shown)


test/testdata/rewriter/rails/delegate.rb.rewrite-tree.exp
```diff
@@ -169,7 +169,7 @@ class <emptyTree><<C <root>>> < (::<todo sym>)
 
     <self>.delegate(:foo, :to, :thing, :prefix, local)
 
-    <self>.delegate(:foo, local, :thing)
+    <self>.delegate(:foo, {local => :thing})
```
elliottt, line 172: This is a good example of what's fixed by this PR: the `local => :thing` should never have been treated as keyword arguments.

parser/tools/generate_ast.cc
```diff
@@ -305,7 +305,7 @@ NodeDef nodes[] = {
     {
         "Hash",
         "hash",
-        vector<FieldDef>({{"kwargs", FieldType::Bool}, {"pairs", FieldType::NodeVec}}),
+        vector<FieldDef>({{"inline_", FieldType::Bool}, {"pairs", FieldType::NodeVec}}),
```
elliottt, line 308: As this boolean would be true for cases like `foo(10 => 20)` as well, it seemed easier to rename it rather than special case symbol keys in the parser.

ast/desugar/Desugar.cc
```diff
@@ -758,13 +766,12 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
...
-                            if (hash->kwargs) {
+                            if (hash->inline_) {
                                 numPosArgs--;
 
                                 // skip inlining the kwargs if there are any non-key/value pairs present
-                                if (absl::c_all_of(hash->pairs, [](auto &arg) {
-                                        return parser::isa_node<parser::Pair>(arg.get());
-                                    })) {
+                                if (absl::c_all_of(hash->pairs,
+                                                   [](auto &arg) { return isInlininableKwargPair(arg.get()); })) {
```
elliottt, line 774: This original condition was too lax: it was only looking for pairs present in the hash, not also that those pairs had keys that were symbols.

elliottt:

→ https://go/builds/bui_IyioeAL9gb5gfW
→ https://go/builds/bui_IyioHVnL1dNYxU

test/testdata/desugar/keyword_args.rb
```diff
@@ -0,0 +1,25 @@
...
+foo(other1 => 10) # error: Passing a hash where the specific keys are unknown
+foo(**{other1 => 10}) # error: Passing a hash where the specific keys are unknown
+
+other2 = "foo"
+foo(other2 => 10) # error: Passing a hash where the specific keys are unknown
+foo(**{other2 => 10}) # error: Passing a hash where the specific keys are unknown
+
+other3 = T.let(:foo, Symbol)
+foo(other3 => 10) # error: Passing a hash where the specific keys are unknown
+foo(**{other3 => 10}) # error: Passing a hash where the specific keys are unknown
```
aprocter-stripe, line 25: As is already the case, these will _not_ error if we simply say `other3 = :foo` instead of using `T.let(...,Symbol)`, right?

core/types/calls.cc
```diff
@@ -493,6 +493,31 @@ optional<core::AutocorrectSuggestion> maybeSuggestExtendTHelpers(const GlobalSta
         {core::AutocorrectSuggestion::Edit{nextLineLoc.value(), fmt::format("{}extend T::Helpers\n", prefix)}}};
 }
 
+// Ensure that a ShapeType used as a keyword args splat in a send has only symbol keys present.
+const ShapeType *isKwargsHash(const GlobalState &gs, const TypePtr &ty) {
```
aprocter-stripe, line 497:
```suggestion
const ShapeType *fromKwargsHash(const GlobalState &gs, const TypePtr &ty) {
```

aprocter-stripe: Or `tryConvertFrom` or something, I dunno :)

core/types/calls.cc
```diff
@@ -800,7 +825,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
             auto kwSplatType = Types::approximate(gs, kwSplatArg->type, *constr);
 
             if (hasKwargs) {
-                if (auto *hash = cast_type<ShapeType>(kwSplatType)) {
+                if (auto *hash = isKwargsHash(gs, kwSplatType)) {
```
aprocter-stripe, line 828: Fallback behavior when this is null is that we pass the kwargsplatted hash as-is as a posarg, right?

How would the following example be handled?

https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams(h%3A%20T%3A%3AHash%5BT.untyped%2C%20Integer%5D).void%7D%0Adef%20foo(h)%0Aend%0A%0Ah%20%3D%20%7B42%20%3D%3E%2043%7D%0Afoo(**h%2C%20x%3A%20222)%0A

aprocter-stripe (changes requested):
Maybe no actual changes needed, but I have a question about a weird corner case (see comments).

elliottt:

→ https://go/builds/bui_IzjSZao9qGBEQE
→ https://go/builds/bui_IzjSI4M9RFvWn3

elliottt:
Successful run on stripe-ci here: https://go/builds/bui_Izkz1tm4bgAlnS

core/types/calls.cc
elliottt: In this case we'll behave like ruby, and treat this as a single hash passed as positional argument.

test/testdata/desugar/keyword_args.rb
elliottt: Yes, that's correct. The main difference is that `:foo` will have the type `Symbol(:foo)`, which can't currently write, while `T.let(:foo, Symbol)` will be typed as `Symbol` and not work for keyword args checking purposes.

aprocter-stripe (approved):

parser/tools/generate_ast.cc
jez: worth noting that the boolean is in fact called `kwargs` in the Ruby VM iirc

jez:
```
❯ ruby --dump=parsetree -e 'foo({x: 0})'
###########################################################
## Do NOT use this node dump for any purpose other than  ##
## debug and research.  Compatibility is not guaranteed. ##
###########################################################

# @ NODE_SCOPE (line: 1, location: (1,0)-(1,11))
# +- nd_tbl: (empty)
# +- nd_args:
# |   (null node)
# +- nd_body:
#     @ NODE_FCALL (line: 1, location: (1,0)-(1,11))*
#     +- nd_mid: :foo
#     +- nd_args:
#         @ NODE_ARRAY (line: 1, location: (1,4)-(1,10))
#         +- nd_alen: 1
#         +- nd_head:
#         |   @ NODE_HASH (line: 1, location: (1,4)-(1,10))
#         |   +- nd_alen: 1 (hash literal)
#         |   +- nd_head:
#         |       @ NODE_ARRAY (line: 1, location: (1,5)-(1,9))
#         |       +- nd_alen: 2
#         |       +- nd_head:
#         |       |   @ NODE_LIT (line: 1, location: (1,5)-(1,7))
#         |       |   +- nd_lit: :x
#         |       +- nd_head:
#         |       |   @ NODE_LIT (line: 1, location: (1,8)-(1,9))
#         |       |   +- nd_lit: 0
#         |       +- nd_next:
#         |           (null node)
#         +- nd_next:
#             (null node)

~/sorbet (master)
❯ ruby --dump=parsetree -e 'foo(x: 0)'
###########################################################
## Do NOT use this node dump for any purpose other than  ##
## debug and research.  Compatibility is not guaranteed. ##
###########################################################

# @ NODE_SCOPE (line: 1, location: (1,0)-(1,9))
# +- nd_tbl: (empty)
# +- nd_args:
# |   (null node)
# +- nd_body:
#     @ NODE_FCALL (line: 1, location: (1,0)-(1,9))*
#     +- nd_mid: :foo
#     +- nd_args:
#         @ NODE_ARRAY (line: 1, location: (1,4)-(1,8))
#         +- nd_alen: 1
#         +- nd_head:
#         |   @ NODE_HASH (line: 1, location: (1,4)-(1,8))
#         |   +- nd_alen: 0 (keyword argument)
#         |   +- nd_head:
#         |       @ NODE_ARRAY (line: 1, location: (1,4)-(1,8))
#         |       +- nd_alen: 2
#         |       +- nd_head:
#         |       |   @ NODE_LIT (line: 1, location: (1,4)-(1,6))
#         |       |   +- nd_lit: :x
#         |       +- nd_head:
#         |       |   @ NODE_LIT (line: 1, location: (1,7)-(1,8))
#         |       |   +- nd_lit: 0
#         |       +- nd_next:
#         |           (null node)
#         +- nd_next:
#             (null node)

~/sorbet (master)
❯ ruby --dump=parsetree -e 'foo(10 => 20)'
###########################################################
## Do NOT use this node dump for any purpose other than  ##
## debug and research.  Compatibility is not guaranteed. ##
###########################################################

# @ NODE_SCOPE (line: 1, location: (1,0)-(1,13))
# +- nd_tbl: (empty)
# +- nd_args:
# |   (null node)
# +- nd_body:
#     @ NODE_FCALL (line: 1, location: (1,0)-(1,13))*
#     +- nd_mid: :foo
#     +- nd_args:
#         @ NODE_ARRAY (line: 1, location: (1,4)-(1,12))
#         +- nd_alen: 1
#         +- nd_head:
#         |   @ NODE_HASH (line: 1, location: (1,4)-(1,12))
#         |   +- nd_alen: 0 (keyword argument)
#         |   +- nd_head:
#         |       @ NODE_ARRAY (line: 1, location: (1,4)-(1,12))
#         |       +- nd_alen: 2
#         |       +- nd_head:
#         |       |   @ NODE_LIT (line: 1, location: (1,4)-(1,6))
#         |       |   +- nd_lit: 10
#         |       +- nd_head:
#         |       |   @ NODE_LIT (line: 1, location: (1,10)-(1,12))
#         |       |   +- nd_lit: 20
#         |       +- nd_next:
#         |           (null node)
#         +- nd_next:
#             (null node)
```

and in fact, it looks like `keyword argument` is `0` from the parser on the `10 => 20` case, so there might be an argument in favor of fixing this in the `parser/Builder.cc` code

PR #4016, opened 2021-02-20
Don't store file contents in the cache.

jvilk-stripe:
Don't store file contents in the cache.

We don't actually need or use them, and they take a lot of space. On Stripe's codebase, it shrinks the cache by ~24% and speeds up typechecking by a little bit when the cache is cold.

### Background

The reason Sorbet stores file contents in the cache is to gracefully handle the situation when there is a hash collision on the contents of a single file. Sorbet stores ASTs for the file at path `P1` with contents `C1` at key `P1:SHA(C1)`. If Sorbet is later run on file `P1` with different contents `C2` such that `SHA(C1) == SHA(C2)` (hash collision), then it will reuse the AST for `C1` in the cache.

This situation, the same file having different contents across runs that hash to the same value, is vanishingly unlikely. The most immediate consequence of this action is that the user will observe typechecking results for the old version of the file, which is acceptable behavior for something so unlikely.

Now, why store the old version of the file? Why not store just the AST and use the new version of the file? The answer lies in locs, Sorbet's struct for location information. If a loc refers to a position in the source text that is beyond the size of the source text, pretty printing will try to print invalid memory, which is bad. (Unfortunately, there _are_ situations in the IDE case where Sorbet can end up with invalid locs like this, but that's another story...)

### This PR

We're already giving the user invalid typechecking results in this (very unlikely) situation. If we only care about crashes, we can just ensure that the file's new length is as long as expected (or is at least as long as was last encountered). The only change in behavior is that any errors pointing to locations in the new file may look garbled. That is exactly what this PR accomplishes.

The upside to this change is a smaller cache (so it reduces virtual memory usage a bit), faster compression/serialization into the cache, and faster cache flushes to disk -- which results in slightly faster typechecks/IDE startup when the cache is cold.

### Motivation

Reduce cache size and slightly speed up IDE startup w/ a cache.

### Test plan

See included automated tests.

### DIFF
--- a/core/serialize/serialize.cc
+++ b/core/serialize/serialize.cc
@@ -884,19 +884,26 @@ u4 Serializer::loadGlobalStateUUID(const GlobalState &gs, const u1 *const data) 
     return SerializerImpl::unpickleGSUUID(p);
 }
 
-vector<u1> Serializer::storeFile(const core::File &file, ast::ParsedFile &tree) {
+vector<u1> Serializer::storeTree(const core::File &file, const ast::ParsedFile &tree) {
     Pickler p;
-    SerializerImpl::pickle(p, file);
+    // See comment in `serialize.h` above `loadTree`.
+    p.putU4(file.source().size());
+    SerializerImpl::pickle(p, file.getFileHash());
     SerializerImpl::pickle(p, tree.tree);
     return p.result();
 }
 
-CachedFile Serializer::loadFile(const core::GlobalState &gs, const u1 *const data) {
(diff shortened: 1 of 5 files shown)


jvilk-stripe:

→ https://go/builds/bui_Iyk6BR6YEtxTjt
→ https://go/builds/bui_Iyk6LaYt8sYuwC

jvilk-stripe:
Now ready for review.

elliottt (approved):
🚀

PR #4018, opened 2021-02-22
Amend the message for 5013

elliottt:
Add an example of how to avoid the error when using `T.cast` to the docs for error 5013.

### Motivation

The error message for 5013 can be confusing if you're using `T.cast` as a runtime assertion and not intending to declare an instance/class variable.

### Test plan

See included automated tests.

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -339,9 +339,9 @@ end
 
 For how to fix, see [Type Annotations](type-annotations.md).
 
-If your intent was not to declare the type of the instance variable but was
-instead just to assign to it, you can assign to an intermediate variable to
-avoid this error:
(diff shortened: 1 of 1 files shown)


website/docs/error-reference.md
```diff
@@ -339,6 +339,17 @@ end
 
 For how to fix, see [Type Annotations](type-annotations.md).
 
+If your intent was not to declare the type of the instance variable but to check
+the type of the value being assigned to it, you can assign to an intermediate
+variable to avoid this error:
```
jez, line 344:
```suggestion
To instead use `T.cast` as a runtime-only type check (that is, neither as a
statically-checked assertion nor as an instance variable declaration), assign
the cast result to an intermediate variable:
```

jez (approved):

PR #4019, opened 2021-02-23
Scaffolding to generate call_validation.rb via Bazel

jez:
### Motivation

We have a bunch of pseudo generated code in call_validation.rb (all the fast
paths).

We also want to make a change to use `UnboundMethod#bind_call` if available,
with zero overhead (i.e., if you're using Ruby 2.7 or greater, it's as if the
whole project used it exclusively, and if you're on 2.6 or before,
sorbet-runtime still works, with no required monkey patches).

@djudd-stripe has done some work to measure the overhead of various `bind_call`
alternatives:

```
master:                                               476.321 ns
2.6/2.7 compatibility wrappers:                       384.027 ns
use bind_call exclusively, breaking compatibility:    343.558 ns
```

The plan to get to the best-case scenario below, while maintaining
compatibility is to generate two `call_validation.rb` files, and require one or
the other by reflecting on the current Ruby version.

**Thus**: this PR sets up the Bazel scaffolding to generate
`call_validation.rb`.

### Next steps:

tbd whether these should be in this change, a stacked change, or an unstacked future change.

- [ ] Get to parity (make `generate_call_validation.cc` emit the same thing that's in call_validation.rb now)
- [ ] Update the generator to take advantage of `bind_call`.
- [ ] (optional) refactor it so that the various fast paths are generated too

### Test plan

I committed a failing test to make sure that Buildkite fails:

<https://buildkite.com/sorbet/sorbet/builds/14915#ed3756ba-1838-4cb7-a88f-1c4d99500c39/254-257>

I also verified that the new `update` commands work locally, as written in the
README.


froydnj (approved):
This is basically #3683 with less sed, more bazel, and more tightly scoped?  I do like having everything in bazel.

Do we think that `bind_call` taking up slightly more stack space than `bind().call` (even given the fewer allocations) is likely to cause problems?

jez:
> Do we think that `bind_call` taking up slightly more stack space than `bind().call` (even given the fewer allocations) is likely to cause problems?

What do you mean slightly more stack space? I was not aware of anything taking more stack space.

The thing that took too much stack space last time was when we had compatibility wrappers that added another frame to the call stack, but I think that the approach of having two files will be mean that the stack space is unchanged.

> This is basically #3683 with less sed, more bazel, and more tightly scoped? I do like having everything in bazel.

Yes, and also I think that it will be easier to use this scaffolding to reduce all the code duplication that already exists in call_validation (e.g., we could generate the 10 fast-path validators).

jez:
> with less sed

Also to be clear: I think it was great that we used `sed` at first just to quickly check the approach overall. Thanks for putting that together!

froydnj:
> > Do we think that `bind_call` taking up slightly more stack space than `bind().call` (even given the fewer allocations) is likely to cause problems?
> 
> What do you mean slightly more stack space? I was not aware of anything taking more stack space.
> 
> The thing that took too much stack space last time was when we had compatibility wrappers that added another frame to the call stack, but I think that the approach of having two files will be mean that the stack space is unchanged.

Mmmm, compat wrappers, yes.  I was thinking actual C stack usage:

https://github.com/ruby/ruby/blob/931815bfd86df603337194f3fcefb46bfe3e7940/proc.c#L2472-L2488

since we're putting `struct METHOD` on the stack now...which might matter?  (I mean, we should try to make it not matter, but maybe it does right now?)

> > This is basically #3683 with less sed, more bazel, and more tightly scoped? I do like having everything in bazel.
> 
> Yes, and also I think that it will be easier to use this scaffolding to reduce all the code duplication that already exists in call_validation (e.g., we could generate the 10 fast-path validators).

:+1:

> > with less sed
> 
> Also to be clear: I think it was great that we used `sed` at first just to quickly check the approach overall. Thanks for putting that together!

:+1:

jez:
After implementing the code generation, here's a benchmark run:

Tested on Ryzen 3700X, ruby 2.7.2p137

    master:                                               sig {params(x: Integer).void}: 422.979 ns
    2.6/2.7 compatibility wrappers:                       sig {params(x: Integer).void}: 367.138 ns
    use bind_call exclusively, breaking compatibility:    sig {params(x: Integer).void}: 322.575 ns
    this PR:                                              sig {params(x: Integer).void}: 304.047 ns

This PR is technically 20ns faster than just throwing away compatibility and using `bind_call` exclusively, so there's probably a reasonable amount of noise here.

But approximately 100ns reduction (39% faster), and as fast as breaking compatibility.

gems/sorbet-runtime/lib/types/private/methods/call_validation_2_6.rb
```diff
@@ -73,7 +81,7 @@ def self.create_validator_method_fast0(mod, original_method, method_sig, return_
             message,
             'Return value',
             nil,
-            return_type,
+            method_sig.return_type,
```
jez, line 84: This method appeared to be the only one using `return_type` here instead of `method_sig.return_type`, which appears to be insignificant.

Yay for code generation! The code is now more consistent.

gems/sorbet-runtime/lib/types/private/methods/call_validation_2_6.rb
```diff
@@ -335,13 +340,11 @@ def self.create_validator_method_fast3(mod, original_method, method_sig, return_
     end
   end
 
-  def self.create_validator_method_fast4(mod, original_method, method_sig, return_type,
-      arg0_type, arg1_type, arg2_type, arg3_type)
+  def self.create_validator_method_fast4(mod, original_method, method_sig, return_type, arg0_type, arg1_type, arg2_type, arg3_type)
```
jez, line 343: I didn't feel like special casing this line wrapping behavior.

gems/sorbet-runtime/lib/types/private/methods/call_validation_2_6.rb
```diff
@@ -91,7 +99,6 @@ def self.create_validator_method_fast1(mod, original_method, method_sig, return_
       # This block is called for every `sig`. It's critical to keep it fast and
       # reduce number of allocations that happen here.
       # This method is a manually sped-up version of more general code in `validate_call`
-
```
jez: There are a handful of whitespace-only changes, feel free to view ignoring whitespace.

gems/sorbet-runtime/lib/types/private/methods/call_validation_2_6.rb
```diff
@@ -24,7 +32,7 @@ def self.create_validator_method_fast(mod, original_method, method_sig)
                                     method_sig.arg_types[2][1].raw_type,
                                     method_sig.arg_types[3][1].raw_type)
     else
-      raise "should not happen"
+      raise 'should not happen'
```
jez, line 35: Single quotes makes it easier to put this code in a C++ string literal.

gems/sorbet-runtime/lib/types/private/methods/call_validation_2_6.rb
```diff
@@ -782,3 +778,4 @@ def self.create_validator_procedure_fast4(mod, original_method, method_sig,
     end
   end
 
+end
```
jez, line 781: This matches with the new `module` line at the top of the file.

gems/sorbet-runtime/lib/types/private/methods/call_validation_2_6.rb
```diff
@@ -1,6 +1,14 @@
+# frozen_string_literal: true
+# typed: false
+
+# DO NOT EDIT. This file is autogenerated. To regenerate, run:
+#
+#     bazel test //gems/sorbet-runtime:update_call_validation
+
+module T::Private::Methods::CallValidation
```
jez, line 8: This is the same namespace as `call_validation.rb` sets up (i.e., it reopens this module).

If you prefer, I can instead generate the header and footer, and make `call_validation.rb` literally nothing more than a conditional require statement.

gems/sorbet-runtime/tools/generate_call_validation.cc
```diff
@@ -171,7 +171,12 @@ void generateCreateValidatorFast(ValidatorKind kind, size_t arity) {
             returnValueVar = "";
             break;
     }
-    fmt::print("      {}original_method.bind(self).call(", returnValueVar);
+    fmt::print("      {}original_method", returnValueVar);
+    if (options.bindCall) {
+        fmt::print(".bind_call(self, ");
+    } else {
+        fmt::print(".bind(self).call(");
+    }
```
jez: There it is! All for these few lines.

gems/sorbet-runtime/test/types/method_validation.rb
```diff
@@ -181,7 +181,9 @@ def @mod.foo(x, y)
         before = GC.stat(:total_allocated_objects)
         @mod.foo("foo", 1)
         allocated = GC.stat(:total_allocated_objects) - before
-        assert_equal(2, allocated) # dmitry: for some reason, when run locally this numeber is 0, in CI it's 2. IDK why.
+
+        expected_allocations = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7') ? 1 : 2
+        assert_equal(expected_allocations, allocated) # dmitry: for some reason, when run locally this numeber is 0, in CI it's 2. IDK why.
```
jez, line 186: 🎉

gems/sorbet-runtime/tools/generate_call_validation.cc
```diff
@@ -0,0 +1,261 @@
...
+            fmt::print(",\n");
+            fmt::print("                                    method_sig.arg_types[{}][1].raw_type", i);
+        }
+        fmt::print(")\n");
+    }
+    fmt::print("    else\n");
+    fmt::print("      raise 'should not happen'\n");
+    fmt::print("    end\n");
+    fmt::print("  end\n");
+    fmt::print("\n");
```
froydnj, line 98: Slight preference for using concatenated string literals:

```c++
fmt::print("  else\n"
           "    raise 'should not happen'\n"
           "  end\n"
           ...);
```

or use the new-fangled raw-string literals: https://en.cppreference.com/w/cpp/language/string_literal

you can even do a (typically C++ ugly) here-doc sort of thing to avoid having to escape anything:

```c++
fmt::print(R"STRLIT(    else
        raise "should not happen"
      end
 ....)STRLIT")
```

I have a small preference for the former, but either one is easier to read than a long string of `fmt::print` calls.

gems/sorbet-runtime/lib/types/private/methods/call_validation_2_6.rb
froydnj: This is fine; I have no particular preference.

froydnj (approved):
lgtm

gems/sorbet-runtime/tools/generate_call_validation.cc
jez: 😅

I can make this change, but I think it's a worse structuring,

- variables end up way farther down, at the bottom of many lines, instead of on the line they're being interpolated into
- if you happen to not end a line with `\n` clang-format will have its way with formatting the code
- for the same reason why we prefer trailing commas, i prefer this format (e.g., the diff from inserting one line at the start or at the end is small, you can copy / paste either the start or the end line to add one more line and start changing it)

in particular i think the first point (location of variable interpolation) is a deal breaker for me, so i'll only coallesce `fmt::print` lines where there isn't an intervening interpolation.

froydnj: > in particular i think the first point (location of variable interpolation) is a deal breaker for me, so i'll only coallesce `fmt::print` lines where there isn't an intervening interpolation.

Your explanation makes sense to me and is what I had in mind, though I failed to communicate it adequately. :+1:

gems/sorbet-runtime/bench/typecheck.rb
```diff
@@ -129,6 +129,26 @@ def self.run
...
+
+      if T::Configuration::AT_LEAST_RUBY_2_7
+        time_block(".bind_call(example)") do
+          Object.instance_method(:class).bind_call(example)
+        end
+
+        time_block("if AT_LEAST_RUBY_2_7; .bind_call(example); end") do
+          if T::Configuration::AT_LEAST_RUBY_2_7
+            Object.instance_method(:class).bind_call(example)
+          else
```
froydnj, line 145: Why are we doing this test twice?

PR #4021, opened 2021-02-23
Re-order some parts of call_validation.rb

jez:
### Motivation

This will make it easier to have a clean diff in #4019, because the parts at
the top and bottom of the file will be not auto-generated, while the other
parts will be.

### Test plan

n/a (existing tests)


aprocter-stripe (approved):
Rubber-stamping on the assumption that this is only re-ordering code, but not otherwise changing anything. (pff to bat it back to me if this assumption is wrong.)

jez:
> Rubber-stamping on the assumption that this is only re-ordering code, but not otherwise changing anything. (pff to bat it back to me if this assumption is wrong.)

Yeah it's just cut + paste. Unfortunately GitHub doesn't show purely moved code in the diffs.

PR #4025, opened 2021-02-24
Infer Array#flatten type taking into account #to_ary

kddeisz:
When you call Array#flatten, Ruby will check if the object that's being flattened responds to #to_ary. If it does, then it will implicitly call it. So in order to correct infer the type, we need to mimic that behavior.

### Motivation

I was working on adding sorbet to a project that was flattening a list of ActiveRecord::Relation objects, and it was incorrectly inferring the result.

### Test plan

See included automated tests.

I haven't written C++ in a long time and this is my first time contributing, so apologies if I got literally all of this wrong. But I believe it works maybe?

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2379,8 +2379,30 @@ class Shape_to_hash : public IntrinsicMethod {
 } Shape_to_hash;
 
 class Array_flatten : public IntrinsicMethod {
+    // If the element type supports the #to_ary method, then Ruby will implicitly call it when flattening. So here we
+    // dispatch #to_ary and recurse further down the result if it succeeds, otherwise we just return the type.
+    static TypePtr typeToAry(const GlobalState &gs, const DispatchArgs &args, const TypePtr &type, const int newDepth) {
+        NameRef toAry = core::Names::toAry();
+
+        InlinedVector<const TypeAndOrigins *, 2> sendArgs;
+        InlinedVector<LocOffsets, 2> sendArgLocs;
+        CallLocs sendLocs{args.locs.file, args.locs.call, args.locs.receiver, sendArgLocs};
+
+        DispatchArgs innerArgs{toAry,    sendLocs, 0,
+                               sendArgs, type,     {type, args.fullType.origins},
+                               type,     nullptr,  args.originForUninitialized};
+
+        auto dispatched = type.dispatchCall(gs, innerArgs);
+        if (dispatched.main.errors.empty()) {
+            return recursivelyFlattenArrays(gs, args, move(dispatched.returnType), newDepth);
+        }
+
+        return type;
+    }
+
     // Flattens a (nested) array all way down to its (inner) element type, stopping if we hit the depth limit first.
-    static TypePtr recursivelyFlattenArrays(const GlobalState &gs, const TypePtr &type, const int64_t depth) {
+    static TypePtr recursivelyFlattenArrays(const GlobalState &gs, const DispatchArgs &args, const TypePtr &type,
+                                            const int64_t depth) {
         ENFORCE(type != nullptr);
 
         if (depth == 0) {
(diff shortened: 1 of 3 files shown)


core/types/calls.cc
```diff
@@ -2395,20 +2395,43 @@ class Array_flatten : public IntrinsicMethod {
...
-                                    recursivelyFlattenArrays(gs, o.right, newDepth));
+                result = Types::any(gs, recursivelyFlattenArrays(gs, args, o.left, newDepth),
+                                    recursivelyFlattenArrays(gs, args, o.right, newDepth));
+            },
+
+            [&](const ClassType &c) {
+                // If the element type supports the #to_ary method, then Ruby will implicitly call it when
+                // flattening. So here we need to get the return type of calling #to_ary on an element.
+                auto toAry = c.symbol.data(gs)->findMember(gs, Names::toAry());
+                if (toAry.exists()) {
```
froydnj, line 2406: We can unconditionally dispatch here, and then discard the result if there are errors.  See the `Magic_splat` implementation in this same file for how to do that.

test/testdata/infer/flatten.rb
```diff
@@ -22,6 +22,26 @@
...
+  sig { returns(T::Array[Integer]) }
+  def to_ary
+    [@left, @right]
+  end
+end
+
+flat_to_ary = T.let(
+  [ToAry.new(1, 2), ToAry.new(3, 4)],
+  T::Array[ToAry]
+)
```
froydnj, line 43: Thank you for adding a test.  Can you also add a test where `to_ary` is not defined on the class, but on a superclass?  I believe your current implementation will fail in such a case because the `findMember` check you used above will only check in the current class for such a symbol.

core/types/calls.cc
```diff
@@ -2395,20 +2395,43 @@ class Array_flatten : public IntrinsicMethod {
...
             [&](const AppliedType &a) {
                 if (a.klass != Symbols::Array()) {
                     result = type;
-                    return;
+                } else {
+                    ENFORCE(a.targs.size() == 1);
+                    result = recursivelyFlattenArrays(gs, args, a.targs.front(), newDepth);
                 }
-                ENFORCE(a.targs.size() == 1);
-                result = recursivelyFlattenArrays(gs, a.targs.front(), newDepth);
```
froydnj, line 2408: Could you please retain the early-return logic here, and just change the `recursivelyFlattenArrays` call?

froydnj (commented):
Thanks for the contribution!  This looks pretty good: a few minor comments and the check for correctly-formatted C++ files failed (I forget to do this all the time, too!).  You can fix up the formatting problems by running `./tools/scripts/format_cxx.sh` and committing the changes.  (Remember to do that after you make changes from review comments, too.)

froydnj (changes requested):
Whoops, sorry, forgot to set the "request changes" bit.

kddeisz:
@froydnj thanks for getting back to me so fast! I've rerun the formatting after I made the changes so that should pass now. As for the other changes - I'm unconditionally dispatching `#to_ary` now to `ClassType` types.

I also realized it was not working correctly for generics that weren't arrays, so I've added that logic as well.

core/types/calls.cc
```diff
@@ -2395,20 +2396,52 @@ class Array_flatten : public IntrinsicMethod {
...
+                result = Types::any(gs, recursivelyFlattenArrays(gs, args, o.left, newDepth),
+                                    recursivelyFlattenArrays(gs, args, o.right, newDepth));
+            },
+
+            [&](const ClassType &c) {
+                // If the element type supports the #to_ary method, then Ruby will implicitly call it when
+                // flattening. So here we need to get the return type of calling #to_ary on an element.
+                DispatchArgs innerArgs{Names::toAry(), args.locs, 0,
+                                       args.args,      type,      {type, args.fullType.origins},
+                                       type,           nullptr,   args.originForUninitialized};
```
froydnj, line 2408: This is the right idea, but the call to `to_ary` that you're dispatching doesn't have the arguments of the original call to `flatten`, it has zero arguments.  So you need to set up things for a zero-argument call.  `Magic_splat.apply` shows how this is done.

core/types/calls.cc
```diff
@@ -2395,20 +2396,52 @@ class Array_flatten : public IntrinsicMethod {
...
                     return;
                 }
-                ENFORCE(a.targs.size() == 1);
-                result = recursivelyFlattenArrays(gs, a.targs.front(), newDepth);
+
+                // If we're dealing with some other generic and it supports #to_ary then we can bypass it and just go
+                // straight to the return type of that method.
+                DispatchArgs innerArgs{Names::toAry(), args.locs, 0,
+                                       args.args,      type,      {type, args.fullType.origins},
+                                       type,           nullptr,   args.originForUninitialized};
```
froydnj, line 2431: Likewise here; the code as written dispatches `to_ary` with the original arguments of the `flatten` call.

core/types/calls.cc
```diff
@@ -2395,20 +2396,52 @@ class Array_flatten : public IntrinsicMethod {
...
+
+                auto dispatched = a.dispatchCall(gs, innerArgs);
+                if (dispatched.main.errors.empty()) {
+                    result = recursivelyFlattenArrays(gs, args, dispatched.returnType, newDepth);
+                    return;
+                }
+
+                // Otherwise just bail.
+                result = type;
+                return;
```
froydnj, line 2441: Tiny nit: you don't need this `return`.

froydnj (changes requested):
This looks pretty good, but there's one small thing to iron out in two different places.  I'm not really sure how you would add a test for it, though.

kddeisz:
@froydnj I think I got what you were saying in my last commit here.

core/types/calls.cc
```diff
@@ -2379,8 +2379,30 @@ class Shape_to_hash : public IntrinsicMethod {
 } Shape_to_hash;
 
 class Array_flatten : public IntrinsicMethod {
+    // If the element type supports the #to_ary method, then Ruby will implicitly call it when flattening. So here we
+    // dispatch #to_ary and recurse further down the result if it succeeds, otherwise we just return the type.
+    static TypePtr typeToAry(const GlobalState &gs, const DispatchArgs &args, const TypePtr &type, const int newDepth) {
```
froydnj, line 2384: Nice refactoring.

froydnj (approved):
Thank you!

froydnj:
Sorry about the delay in merging this!  Github's display of the CI status looked like something was just off on the CI side...and then I remembered that I needed to manually unblock the tests. :man_facepalming:

kddeisz:
No worries thanks that was fun

PR #4026, opened 2021-02-24
Update FAQ for supported versions

jez:
### Motivation

https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1614184992012900

### Test plan

See included automated tests.

### DIFF
--- a/website/docs/faq.md
+++ b/website/docs/faq.md
@@ -265,7 +265,7 @@ used to run a given project. You will have to rely on (runtime) test suites to
 verify that a project does not use new standard library APIs on an old Ruby
 version.
 
-Specifically, `sorbet-static` is known to support Ruby 2.4, Ruby 2.6, Ruby 2.6,
+Specifically, `sorbet-static` is known to support Ruby 2.4, Ruby 2.5, Ruby 2.6,
 and Ruby 2.7 to a minimum level (i.e., it can at least parse syntax introduced
 in those versions). Some language features are typed more strictly than others
 (generally, language features in newer Ruby versions have looser type support).
--- a/website/docs/faq.md
+++ b/website/docs/faq.md
@@ -252,20 +252,45 @@ bundle exec srb rbi suggest-typed
 
 ## What platforms does Sorbet support?
 
-The `sorbet` and `sorbet-runtime` gems are currently only tested on Ruby 2.5 and
-Ruby 2.6.
+The `sorbet-runtime` gem is currently only tested on Ruby 2.5 and Ruby 2.6, and
+we believe it to work on Ruby 2.7. It is known to not support Ruby 2.4. Feel
+free to report issues for any current or future Ruby version.
 
-Ruby 2.7 has
-[known issues](https://github.com/sorbet/sorbet/issues?q=is%3Aissue+2.7+) and is
-[not being worked on yet](https://github.com/sorbet/sorbet/issues/2771#issuecomment-599761098)
-(as of May 2020) but PRs are welcome!
+Feel free to report runtime issues for any current or future Ruby version.
 
-The static checker is only tested on macOS 10.14 (Mojave) and Ubuntu 18 (Bionic
-Beaver). We expect it to work on macOS 10.10 (Yosemite) and most Linux
-distributions where `glibc`, `git` and `bash` are present.
+The `sorbet-static` gem aims to support a superset of all Ruby versions. In
+particular, the standard library [RBI files](rbi.md) might reflect classes,
+methods, or APIs that are only availble in a version of Ruby newer than the one
+used to run a given project. You will have to rely on (runtime) test suites to
+verify that a project does not use new standard library APIs on an old Ruby
+version.
 
-If you are using one of the official minimal Ruby Docker images you will need to
-install the extra dependencies yourself:
+Specifically, `sorbet-static` is known to support Ruby 2.4, Ruby 2.6, Ruby 2.6,
+and Ruby 2.7 to a minimum level (i.e., it can at least parse syntax introduced
+in those versions). Some language features are typed more strictly than others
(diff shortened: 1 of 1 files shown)


website/docs/faq.md
```diff
@@ -252,20 +252,45 @@ bundle exec srb rbi suggest-typed
...
-If you are using one of the official minimal Ruby Docker images you will need to
-install the extra dependencies yourself:
+The `sorbet-runtime` gem is currently only tested on Ruby 2.5 and Ruby 2.6, and
+we believe it to work on Ruby 2.7. It is known to not support Ruby 2.4. Feel
+free to report issues for any current or future Ruby version.
+
+Feel free to report runtime issues for any current or future Ruby version.
+
+The `sorbet-static` gem aims to support a superset of all Ruby versions. In
+particular, the standard library [RBI files](rbi.md) might reflect classes,
```
aprocter-stripe, line 262:
```suggestion
We publish a single `sorbet-static` gem that supports all Ruby versions, rather than
supplying different `sorbet-static` gems for each Ruby version. This means that the standard
library [RBI files](rbi.md) might reflect classes,
```

website/docs/faq.md
```diff
@@ -252,20 +252,45 @@ bundle exec srb rbi suggest-typed
...
+methods, or APIs that are only availble in a version of Ruby newer than the one
+used to run a given project. You will have to rely on (runtime) test suites to
+verify that a project does not use new standard library APIs on an old Ruby
+version.
+
+Specifically, `sorbet-static` is known to support Ruby 2.4, Ruby 2.6, Ruby 2.6,
```
aprocter-stripe, line 268:
```suggestion
Specifically, `sorbet-static` is known to support Ruby 2.4, Ruby 2.5, Ruby 2.6,
```

jaredbeck:
Nice improvement, sharing the goals as well as current status.

website/docs/faq.md
```diff
@@ -252,20 +252,45 @@ bundle exec srb rbi suggest-typed
...
+This is not by design, just by convenience. Feel free to open feature requests
+that various (new or old) language features be typed more strictly.
+
+The `sorbet-static` gem is only tested on macOS 10.14 (Mojave) and Ubuntu 18
+(Bionic Beaver). We expect it to work as far back as macOS 10.10 (Yosemite), as
+far forward as macOS 11.0 Big Sur, and on most Linux distributions where
+`glibc`, `git` and `bash` are present. We do not test nor publish prebuilt
+binaries for macOS on Apple Silicon. We have reports that it doesn't work, but
+no one on the Sorbet team has access to Apple Silicon-based macOS machines, and
+thus have been unable to diagnose and any problems. If you are interested in
```
aprocter-stripe, line 281:
```suggestion
no one on the Sorbet team has access to Apple Silicon-based macOS machines, so
we have been unable to diagnose any problems. If you are interested in
```

website/docs/faq.md
```diff
@@ -252,20 +252,45 @@ bundle exec srb rbi suggest-typed
...
+working on this, feel free to reach out in the #internals channel on our
+[Sorbet Slack](/slack). There is currently no Windows support.
```
aprocter-stripe, line 283: Suggest moving `There is currently no Windows support.` up, to between the first and second sentences, or leaving it here but with a paragraph break.

website/docs/faq.md
```diff
@@ -252,20 +252,45 @@ bundle exec srb rbi suggest-typed
...
+
+The `sorbet-static` gem includes platform-specific, precompiled binaries. On
+Linux, these binaries link against `glibc`. If you are running on a platform
+like `alpine` which uses the `musl` libc implementation, you will need to
+install some support libraries:
```
aprocter-stripe, line 288: Should there be a list after the colon?

aprocter-stripe (approved):
LGTM! Just a couple small typos and a few minor style suggestions.

website/docs/faq.md
jez: > We publish a single `sorbet-static` gem that supports all Ruby versions, rather than

We actually publish like 10 different sorbet-static gems, technically. Do you feel strongly about your wording? I worry that it's incorrect, technically speaking.

aprocter-stripe: Eh, yeah, don't say it if it's incorrect. :)

website/docs/faq.md
jez: the code snippet that follows shows what support libraries to install.

website/docs/faq.md
aprocter-stripe: I think the only reason I was poking on this is that I found the phrasing `superset of all Ruby versions` a bit confusing. Maybe something like `the union of all Ruby versions` or... I dunno.

website/docs/faq.md
aprocter-stripe: Ah, okay. There's some text between this and the snippet now, though.

website/docs/faq.md
```diff
@@ -252,20 +252,45 @@ bundle exec srb rbi suggest-typed
...
-install the extra dependencies yourself:
+The `sorbet-runtime` gem is currently only tested on Ruby 2.5 and Ruby 2.6, and
+we believe it to work on Ruby 2.7. It is known to not support Ruby 2.4. Feel
+free to report issues for any current or future Ruby version.
+
+Feel free to report runtime issues for any current or future Ruby version.
+
+The `sorbet-static` gem aims to support a superset of all Ruby versions. In
+particular, the standard library [RBI files](rbi.md) might reflect classes,
+methods, or APIs that are only availble in a version of Ruby newer than the one
```
connorshea, line 263:
```suggestion
methods, or APIs that are only available in a version of Ruby newer than the one
```

PR #4029, opened 2021-02-26
don't double-check protected methods when doing final methods checks

froydnj:
### Motivation

We already (implicitly) check these because method_defined? checks both public and protected methods.  So in the common case of there being no final method violations, we are doing 50% more work than we need to.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -108,9 +108,7 @@ module T::Private::Methods
         # method_to_key(ancestor.instance_method(method_name)) is not (just) an optimization, but also required for
         # correctness, since ancestor.method_defined?(method_name) may return true even if method_name is not defined
         # directly on ancestor but instead an ancestor of ancestor.
-        if (ancestor.method_defined?(method_name) ||
-            ancestor.private_method_defined?(method_name)) &&
-           final_method?(method_owner_and_name_to_key(ancestor, method_name))
+        if final_method?(method_owner_and_name_to_key(ancestor, method_name))
           definition_file, definition_line = T::Private::Methods.signature_for_method(ancestor.instance_method(method_name)).method.source_location
           is_redefined = target == ancestor
           caller_loc = caller_locations&.find {|l| !l.to_s.match?(%r{sorbet-runtime[^/]*/lib/}) }
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -109,8 +109,7 @@ module T::Private::Methods
         # correctness, since ancestor.method_defined?(method_name) may return true even if method_name is not defined
         # directly on ancestor but instead an ancestor of ancestor.
         if (ancestor.method_defined?(method_name) ||
-            ancestor.private_method_defined?(method_name) ||
-            ancestor.protected_method_defined?(method_name)) &&
+            ancestor.private_method_defined?(method_name)) &&
            final_method?(method_owner_and_name_to_key(ancestor, method_name))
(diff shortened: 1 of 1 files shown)


gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -109,8 +109,7 @@ def self._check_final_ancestors(target, target_ancestors, source_method_names)
         # correctness, since ancestor.method_defined?(method_name) may return true even if method_name is not defined
         # directly on ancestor but instead an ancestor of ancestor.
         if (ancestor.method_defined?(method_name) ||
-            ancestor.private_method_defined?(method_name) ||
-            ancestor.protected_method_defined?(method_name)) &&
+            ancestor.private_method_defined?(method_name)) &&
            final_method?(method_owner_and_name_to_key(ancestor, method_name))
```
froydnj, line 113: OK, so doing this is (probably) a win, but maybe we don't have to do the `method_defined?` dance at all, we can just say:

`if final_method?(method_owner_and_name_to_key(ancestor, method_name))`

or at least check that *first*, because that's a much faster check than `method_defined?` -- I'm not really sure of the circumstances where the situation would be:

* `final_method?` returns true; but
* `method_defined?`/`private_method_defined?` returns false

I guess that if we *were* in such a situation, we wouldn't be diagnosing it currently, and that would be a problem.

froydnj (commented):
Maybe we can make this faster still.

froydnj:
@jez, ptal at the comment above.

froydnj:

→ https://go/builds/bui_J1B94nKlc122uU
→ https://go/builds/bui_J1B9vDXCuMTebA

froydnj:
A little more concretely, do we want to do:

* Patch as currently written.  Obviously correct.
* Additionally swap the ordering of the `final_method?` and `method_defined?` checks here, which ought to be significantly faster (`method_defined?` is slow).  This should also be correct, but I'm not sure if there are corner cases here that our test suite is not covering.
* Drop the `method_defined?` check entirely, because `final_method?` should only be true iff the method was defined in the first place.  Much more nervous about this one.

froydnj:
@azdavis-stripe Do you remember why we have both the `method_defined?` check and the `final_method?` check in this code?  We think that only the `final_method?` check should be necessary, but do not know the history here.  Looking through the commit history of `gems/sorbet-runtime/lib/types/private/methods/_methods.rb` and comments on the relevant PRs doesn't shed any light on that either.

jez:
@froydnj as per standup i'm pretty much fine with any of those refactorings, pending whether azdavis has anything to add.

azdavis:
Going by the comment on this method:

> …check if there is already a method defined on one of target_ancestors with the same name that is final.

I'm guessing I wrote the code to check literally "it is defined and it is final" without really stopping to think that if it is final, then surely it is defined.

I would be happy to see at least doing the final check first since that's faster, but I suspect _only_ doing that check would be enough.

I'd also be especially ok with moving to only the check for final, because by my reckoning, doing that could only give us _more_ errors, not fewer. Since it makes it 'easier' for an error to be raised. (As discussed, it might not actually be making it easier - they might be equivalent. But it's certainly not making it harder).

So if it turns out that this is the wrong decision, we'll know because we'll start getting spurious errors, and then we can come back and revert. We won't have a situation where we start _not_ reporting errors that we should, which would be worse.

froydnj:
> I'd also be especially ok with moving to only the check for final, because by my reckoning, doing that could only give us _more_ errors, not fewer. Since it makes it 'easier' for an error to be raised. (As discussed, it might not actually be making it easier - they might be equivalent. But it's certainly not making it harder).
> 
> So if it turns out that this is the wrong decision, we'll know because we'll start getting spurious errors, and then we can come back and revert. We won't have a situation where we start _not_ reporting errors that we should, which would be worse.

Thanks for the second pair of eyes.  Patch updated to only do the check for `final_method?`.

jez (approved):

PR #4039, opened 2021-03-01
rbi/tools/sync-rdoc.rb rbi/core/unbound_method.rbi

jez:
### Motivation

Splitting out work from #4019 to make the diff smaller.

### Test plan

n/a

### DIFF
--- a/rbi/core/unbound_method.rbi
+++ b/rbi/core/unbound_method.rbi
@@ -1,15 +1,19 @@
 # typed: __STDLIB_INTERNAL
 
 # Ruby supports two forms of objectified methods.
-# [`Class`](https://docs.ruby-lang.org/en/2.6.0/Class.html) `Method` is used to
+# [`Class`](https://docs.ruby-lang.org/en/2.6.0/Class.html)
+# [`Method`](https://docs.ruby-lang.org/en/2.6.0/Method.html) is used to
 # represent methods that are associated with a particular object: these method
 # objects are bound to that object. Bound method objects for an object can be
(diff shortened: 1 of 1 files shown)


rbi/core/unbound_method.rbi
```diff
@@ -1,15 +1,19 @@
 # typed: __STDLIB_INTERNAL
 
 # Ruby supports two forms of objectified methods.
-# [`Class`](https://docs.ruby-lang.org/en/2.6.0/Class.html) `Method` is used to
+# [`Class`](https://docs.ruby-lang.org/en/2.6.0/Class.html)
+# [`Method`](https://docs.ruby-lang.org/en/2.6.0/Method.html) is used to
```
froydnj, line 5: Should `Class` here even be linked?  Seems like a bit of noise.  (Or should there even be "Class" starting the sentence?  Compare the `UnboundMethod` sentence below, which has no such introduction.)

froydnj (approved):

rbi/core/unbound_method.rbi
jez: This is autogenerated from the official documentation.

jez: https://docs.ruby-lang.org/en/2.7.0/UnboundMethod.html

We have a script that scrapes this documentation, converts it to markdown style, and inserts the comments in the right places:

```
rbi/tools/sync-rdoc.rb rbi/core/unbound_method.rbi
```

PR #4040, opened 2021-03-01
Re-run rbi/tools/sync-rdoc.rb for Ruby 2.7

jez:
### Motivation

I noticed that a bunch of changes came in for Ruby 2.7-specific changes in
#4039. I'd rather get ahead of those now.

### Commit summary

- **Prefer Ruby 2.7.0 for RBI doc strings** (3063bca06)

- **Skip SCLASS nodes in sync-rdoc.rb** (4e5c686a7)

  There are a handful of places in RBIs where we have to use `class <<
  self` because both a class and it's singleton class need to register a
  type_member/type_template with the name Elem

  We should probably learn how to handle these in the sync-rdoc.rb script,
  but it happens rarely enough (two usages in one file as far as I can
  tell) that I'd rather not sink time into it right now.

- **Re-run rbi/tools/sync-rdoc.rb** (e86efdfd1)

  I ran this:

      rbi/tools/sync-rdoc.rb rbi/{core,stdlib}/**/*.rbi

  but there were a few files that misbehaved, and I ran out of my timebox
  trying to figure out why:

  - rbi/stdlib/shellwords.rbi
  - rbi/stdlib/open3.rbi
  - rbi/stdlib/kconv.rbi
  - rbi/stdlib/find.rbi
  - rbi/stdlib/fileutils.rbi
  - rbi/stdlib/etc.rbi
  - rbi/stdlib/drb.rbi

  In each case, the error was something like this:

      rbi/tools/sync-rdoc.rb:101:in `error!': rbi/stdlib/drb.rbi:316:2:
  unconsumed sig (RuntimeError)

  I'm guessing maybe a few methods were removed? I'm not sure if that's
  actually the problem and if so what the best way to work around these
  would be.

  In any case, I think that this is the first time this script has been
  run for a Ruby version different from the version it was built against.
  It's probably never been tested for this use case of upgrading a Ruby
  version.

### Test plan

RBI changes only. Specifically, I checked that these are only comment changes.


rbi/tools/sync-rdoc.rb
```diff
@@ -147,6 +147,11 @@ def initialize(file)
         walk_scope(scope, &blk)
         assert_clean!
       end
+    when :SCLASS
+      # There are a handful of places in RBIs where we have to use `class <<
+      # self` because both a class and it's singleton class need to register a
+      # type_member/type_template with the name Elem
+      puts 'skipping :SCLASS node, documentation will not be generated'
```
jez, line 154: cc @univerio if you have an idea for how hard it would be to handle `class << self` nodes in this script, I'd love your input. For example, this is the file the script choked on:

https://github.com/sorbet/sorbet/blob/e86efdfd1e9b1f3f55a6c917a27385c7d03de21e/rbi/stdlib/etc.rbi#L564-L571

rbi/tools/sync-rdoc.rb
```diff
@@ -583,6 +583,12 @@ def run!(argv)
       return -1
     end
 
+    if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.7')
+      puts "Warning! Detected RUBY_VERSION=#{RUBY_VERSION}."
+      puts "This script uses the currenet Ruby version to parse documentation from."
```
aprocter-stripe, line 593: Typo (`currenet`).

(GitHub won't let me make a "suggestion" here for some reason)

aprocter-stripe (approved):

rbi/tools/sync-rdoc.rb
univerio: Oof, it's been a while since I've looked at this. Let me dig into this.

univerio: Turns out it was really simple:

https://github.com/univerio/sorbet/commits/sync-rdoc-2-7

It's spewing a bunch of warnings due to some constants missing from ri though. 😕 Not sure what's going on there.

jez: Thanks!

I've merged this PR. Feel free to open a new one on latest master.

PR #4046, opened 2021-03-02
don't check non-final modules for final method violations

froydnj:
`sorbet-runtime` maintains a set of all modules that contain at least one final method.  If we are checking the ancestor chain, we only need to check modules that are in this set for potential final method violations.  This removes an entire inner loop's worth of checks.

### Motivation

Speed.

### Test plan

See included automated tests.  Something is not quite right, though: the tests pass, but this changes the number of final method violations reported in Stripe's codebase, so we must be doing something not quite right here.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -103,6 +103,7 @@ module T::Private::Methods
     # use reverse_each to check farther-up ancestors first, for better error messages. we could avoid this if we were on
     # the version of ruby that adds the optional argument to method_defined? that allows you to exclude ancestors.
     target_ancestors.reverse_each do |ancestor|
+      next if !module_with_final?(ancestor)
       source_method_names.each do |method_name|
         # the usage of method_owner_and_name_to_key(ancestor, method_name) instead of
         # method_to_key(ancestor.instance_method(method_name)) is not (just) an optimization, but also required for


froydnj:
>  Something is not quite right, though: the tests pass, but this changes the number of final method violations reported in Stripe's codebase, so we must be doing something not quite right here.

Turns out I lied about this: I mistakenly thought that there were actually violations (that this branch wasn't catching) when there were not actually any violations.  And I've been unable to construct new test cases that are different than our current ones and represent the violations I've seen in burning down `final` method violations.  So I think this is ready for review.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -103,6 +103,7 @@ def self._check_final_ancestors(target, target_ancestors, source_method_names)
     # use reverse_each to check farther-up ancestors first, for better error messages. we could avoid this if we were on
     # the version of ruby that adds the optional argument to method_defined? that allows you to exclude ancestors.
     target_ancestors.reverse_each do |ancestor|
+      next if !module_with_final?(ancestor)
```
jez, line 106: Does this make the `!module_with_final?` check from ~6 lines above redundant?

jez (approved):

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
froydnj: I don't think so.  The one above is a fast reject for when the module has never ever seen `final` anywhere, either in itself, or in its ancestor chain.  The one being added here is when something has added some bit of a`final` thing to `target` -- we don't know what -- and we're rejecting things that obviously can't be responsible for an addition of `final`.

Come to think of it, that suggests another optimization we could be doing: right now, we're eagerly computing `target_ancestors` and `source_method_names`, but both of those are relatively expensive, especially in the case of an included-esque module: we're set differencing ancestor chains and computing all instance methods on the included-ish module.  Maybe if we saved both of those until we knew we actually needed them, we could save cycles?

Issue #3497, opened 2020-10-09
Autocorrect inserts T.let into method definition


bhuga:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Bparams(path%3A%20String).returns(Integer)%7D%0A%20%20def%20for_file(path)%0A%20%20%20%20while%20!path.empty%3F%0A%20%20%20%20%20%20return%205%20if%20rand%20%3E%200.5%0A%20%20%20%20%20%20path%20%3D%20path%5B0...(path.rindex('%2F')%20%7C%7C%200)%5D%0A%20%20%20%20end%0A%20%20%20%201%0A%20%20end%0Aend)

```ruby
# typed: true
class A
  extend T::Sig
  sig {params(path: String).returns(Integer)}
  def for_file(path)
    while !path.empty?
      return 5 if rand > 0.5
      path = path[0...(path.rindex('/') || 0)]
    end
    1
  end
end
```

#### Observed output

Sorbet would like to auto-correct to `def for_file(T.let(path, T.nilable(String)))`, which is invalid syntax.

```

# typed: true
class A
  extend T::Sig
  sig {params(path: String).returns(Integer)}
  def for_file(path)
    while !path.empty?
      return 5 if rand > 0.5
      path = path[0...(path.rindex('/') || 0)]
    end
    1
  end
end
editor.rb:8: Changing the type of a variable in a loop is not permitted https://srb.help/7001
     8 |      path = path[0...(path.rindex('/') || 0)]
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Existing variable has type: String
  Attempting to change type to: T.nilable(String)

  Autocorrect: Use `-a` to autocorrect
    editor.rb:5: Replace with T.let(path, T.nilable(String))
     5 |  def for_file(path)
                       ^^^^
Errors: 1
```

#### Expected behavior

Sorbet should not suggest a `T.let` annotation when the first definition of the variable comes from a sig.
Or it should suggest widening the type of the argument in the sig, but getting the location information for that is harder.
- - -

jez:
Relates to #1474

jez:
I fixed this recently, but I don’t have a link to the pull request that fixed it handy.

Issue #3190, opened 2020-06-12
Autocorrects for E7001 result in T.let inside method parameter list


sushain97:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams(klass%3A%20Class).void%7D%0Adef%20foo(klass)%0A%20%20loop%20do%0A%20%20%20%20klass%20%3D%20klass.superclass%0A%20%20end%0Aend%0A)

```ruby
# typed: true
extend T::Sig

sig {params(klass: Class).void}
def foo(klass)
  loop do
    klass = klass.superclass
  end
end
```

#### Observed output

if you quick fix, you get:

```ruby
# typed: true
extend T::Sig

sig {params(klass: Class).void}
def foo(T.let(klass, T.nilable(Class)))
  loop do
    klass = klass.superclass
  end
end
```

clearly garbage

#### Expected behavior

this probably shouldn't be quick fixable at all

Issue #1848, opened 2019-09-25
Quick fixes suggest T.let on a type parameter


sushain97:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aextend%20T%3A%3ASig%0A%0Asig%20do%0A%20%20type_parameters(%3AU)%0A%20%20.params(%0A%20%20%20%20bar%3A%20T.type_parameter(%3AU)%2C%0A%20%20)%0A%20%20.returns(T.nilable(T.type_parameter(%3AU)))%0Aend%0Adef%20foo(bar)%0A%20%20baz%20%3D%20nil%0A%20%203.times%20do%0A%20%20%20%20baz%20%3D%20bar%0A%20%20end%0A%20%20baz%0Aend%0A)

```ruby
# typed: strict

extend T::Sig

sig do
  type_parameters(:U)
  .params(
    bar: T.type_parameter(:U),
  )
  .returns(T.nilable(T.type_parameter(:U)))
end
def foo(bar)
  baz = nil
  3.times do
    baz = bar
  end
  baz
end
```

#### Observed output

```
editor.rb:15: Changing the type of a variable in a loop is not permitted https://srb.help/7001
    15 |    baz = bar
                  ^^^
  Existing variable has type: NilClass
  Attempting to change type to: Object#foo#U

  Autocorrect: Use `-a` to autocorrect
    editor.rb:13: Replace with T.let(nil, T.nilable(Object#foo#U))
    13 |  baz = nil
                ^^^
Errors: 1
```

#### Expected behavior

There should be no suggestion since it isn't possible to `T.let` this currently inside a function body. It just results in nonsense code.

- - -

cc @jvilk-stripe

jez:
Maybe... I'm of the mind here that the code action is actually correct, but the problem is that `T.type_parameter(:U)` **should** be allowed in a method body, and it's not currently.

We have an issue for that (#1715). If you really want to work around that bug in the mean time, sure, but like, it seems kind of beside the point to me.

PR #4047, opened 2021-03-03
Avoid crash when autocorrect has origin pointing to payload

Morriar:
### Motivation

Origins seem messed up for constant references when applying autocorrect for 7001 errors:

```rb
FOO = 10

a = FOO
[].each { |i| a = nil } # Error: changing the type of a variable in a loop
```

Is autocorrected as 

```rb
T.let(FOO, T.nilable(Integer)) = 10

a = FOO
[].each { |i| a = nil }
```

There are a few issues about this already: https://github.com/sorbet/sorbet/issues/3497,  https://github.com/sorbet/sorbet/issues/3190, https://github.com/sorbet/sorbet/issues/1848.

One side effect of this problem is when the constant used comes from the payload:

```rb
b = Float::INFINITY
[].each { |i| b = nil } # Error: changing the type of a variable in a loop
```

The autocorrect will try to write the payload file and sorbet will crash.

While I'm not sure how to fix the original problem, I can at least make it so it doesn't try to write the payload and doesn't crash. When you try to fix hundreds of errors thanks to the autocorrect, a few bad ones are better than a crash that produce no correction at all.

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/autocorrect-payload/autocorrect-payload.out
+++ b/test/cli/autocorrect-payload/autocorrect-payload.out
@@ -1,0 +1,12 @@
+# typed: true
+
+T.let(FOO, T.nilable(Integer)) = 10 # TODO Also fix this autocorrect
+
+a = FOO
+[].each { |i| a = nil }
+
+b = Float::INFINITY
+[].each { |i| b = nil }
+
+c = T.let(10, T.nilable(Integer))
+[].each { |i| c = nil }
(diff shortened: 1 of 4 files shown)


test/cli/autocorrect-payload/autocorrect-payload.out
```diff
@@ -0,0 +1,12 @@
+# typed: true
+
+T.let(FOO, T.nilable(Integer)) = 10 # TODO Also fix this autocorrect
```
Morriar, line 3: I captured the root cause in the test but this PR does not fix it.

froydnj:
302 @jez.  This looks plausible, but I am unsure whether the fix should actually land someplace else.

jez (approved):
Thanks! I wonder if we should bake this into `e.replaceWith` that it drop the suggestion if the edit would be in a payload file, just so that we don't have to play whack-a-mole with this into the future.

PR #4049, opened 2021-03-04
Use ClassOrModuleRef / MethodRef in more places.

jvilk-stripe:
Use ClassOrModuleRef / MethodRef in more places.

### Motivation

Reduce usage of the general-purpose SymbolRef class so that `SymbolRef::data` is used in fewer places.

Next steps will be to define FieldRef, TypeArgumentRef, etc. Then we can make `.data()` on `SymbolRef` private and add general-purpose top-level methods on `SymbolRef` for common utilities (e.g., `toString()`). Then we can begin to define specialized `Symbol` classes!

### Test plan

Covered by existing tests.


aprocter-stripe (approved):
Don't have a lot of context here, but I bounced it off someone who does and the words "slam dunk" were used in response, so I guess that's good enough for a stamp. :)

Issue #3852, opened 2020-12-27
"Illegal instruction" crash on Apple Silicon


nathunsmitty:
#### Input

The sorbet-static binary crashes on new M1 arm-based macs with the error `Illegal instruction`.

```
srb typecheck -e 'puts "Hello, world!"'
```

Attempting to run Sorbet under Rosetta 2 also fails.

```
arch -x86_64 srb typecheck -e 'puts "Hello, world!"'
```

#### Observed output

Both with and without Rosetta 2:

```
/usr/local/lib/ruby/gems/2.6.0/gems/sorbet-0.5.6189/bin/srb: line 46: 37685 Illegal instruction: 4  "${sorbet}" "${args[@]}"
```

#### Expected behavior

Sorbet correctly typechecks the given statement by printing `No errors! Great job.`

#### My Research

I dug into this issue a bit and came across the very similar #1487. (Apologies if this is a dupe.) @DarkDimius' [suggestion](https://github.com/sorbet/sorbet/issues/1487#issuecomment-521506401) of removing the `-march=sandybridge` line produced a binary that appears to be working on my machine for `--config=dbg`. `--config=release-mac` caused compile issues with some libraries.

Since I'm assuming you don't want to give up fancy Intel optimizations, it seems like the best solution would be to [create a universal binary](https://developer.apple.com/documentation/xcode/building_a_universal_macos_binary) through merging an ARM build and an x86 build via `lipo` or another tool. (This is alluded to in #3369.) 

Cross-compiling C++ binaries is a little outside my area of expertise, but I'm happy to help in any way I can!

dnsco:
Hey, I'm trying to figure out how to work with this situation,

looks like I've gotta download a preview build of open jdk and patch bazel to build myslef, I was wondering if anyone has come up with an easier solution?

nathunsmitty:
Unfortunately not. 🙁 

Here's my set up:

1. A custom fork of Sorbet. (You can make your own, or use [mine](https://github.com/nathunsmitty/sorbet)!) There are two lines in `.bazelrc` that I changed:
    1. [Remove the `-march=sandybridge` argument.](https://github.com/nathunsmitty/sorbet/commit/c71789e31ac9a8b147bdbc87e683adb099c47cfe)
[...]

PR #4051, opened 2021-03-04
Only pass `-march` flag in linux release builds

jez:
### Motivation

This should hopefully allow Sorbet to work under Rosetta 2 on Apple
Silicon Macs?

It will come as a slight performance degradation (potentially, I haven't
measured) on Intel-based Macs.

### Test plan

Existing tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -29,11 +29,6 @@ build --incompatible_strict_action_env
 # Don't depend on system compiler
 build --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
 
-
-# Developer laptops run skylake, devboxes run skylake-avx512
-# however some AWS instances in our fleet still run Sandy Bridge (Skylake predecessor)
-build --copt=-march=sandybridge
-
 build --copt=-fno-omit-frame-pointer
 
 build --copt=-fstack-protector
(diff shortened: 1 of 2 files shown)


froydnj (approved):
I wonder if this has negative effects on the local dev cycle, given that payload compression/decompression is no longer using fancy vector instructions (assuming they use anything from SSE3/SSSE3/SSE4/AVX+).  Probably lost in the noise.

jez:
> I wonder if this has negative effects on the local dev cycle, given that payload compression/decompression is no longer using fancy vector instructions (assuming they use anything from SSE3/SSSE3/SSE4/AVX+). Probably lost in the noise.

Yep, I'm down to measure. But IMO local dev on macOS is already so much slower than Linux I'm not too worried about it.

connorshea (approved):
These binaries work for me on both my Intel and ARM Mac (through Rosetta 2). No performance penalty on Intel that I can see.

nathunsmitty:
For error tracking purposes: this change fixes #3852

jez:
I've heard three people try this out and say that it works on their Apple Silicon Mac, and also that it doesn't regress performance noticeably.

PR #4054, opened 2021-03-05
Introduce TypeArgumentRef and TypeMemberRef.

jvilk-stripe:
Introduce TypeArgumentRef and TypeMemberRef.

### Motivation

Reduce usage of SymbolRef::data() so we can eventually specialize Symbol classes.

### Test plan

Covered by existing tests.

### DIFF
--- a/definition_validator/variance.cc
+++ b/definition_validator/variance.cc
@@ -138,9 +138,6 @@ private:
             // This is where the actual variance checks are done.
             [&](const core::LambdaParam &param) {
                 auto paramData = param.definition.data(ctx);
-
-                ENFORCE(paramData->isTypeMember());
-
                 auto paramVariance = paramData->variance();
 
                 if (!hasCompatibleVariance(polarity, paramVariance)) {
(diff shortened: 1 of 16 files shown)


resolver/GlobalPass.cc
```diff
@@ -61,7 +61,7 @@ bool resolveTypeMember(core::GlobalState &gs, core::SymbolRef parent, core::Symb
         typeMember.data(gs)->resultType = core::make_type<core::LambdaParam>(typeMember, untyped, untyped);
         return false;
     }
-    if (!my.isTypeMember() && !my.isTypeArgument()) {
+    if (!my.isTypeMember()) {
```
jvilk-stripe, line 64: I don't know why both of these were checked before. The implementation of these two are conflated elsewhere, so my assumption was that it was incorrect cargo culting.

definition_validator/variance.cc
```diff
@@ -138,9 +138,6 @@ class VarianceValidator {
             // This is where the actual variance checks are done.
             [&](const core::LambdaParam &param) {
                 auto paramData = param.definition.data(ctx);
-
-                ENFORCE(paramData->isTypeMember());
```
elliottt, line 142: 🎉

elliottt (approved):
Awesome work! 🚀

PR #4055, opened 2021-03-05
Stop emitting useless sig annotations in reflection.rbi

### LINKED ISSUE #3674: Hidden definitions cannot be generated for methods with more than one anonymous variable

jez:
### Motivation

Fixes #3674

For context see https://github.com/sorbet/sorbet/issues/3674#issuecomment-728674364

### Test plan

Added failing test, verified that it failed before the fix.


gems/sorbet/test/snapshot/partial/underscores/src/foo.rb
```diff
@@ -0,0 +1,7 @@
+
+class A
+  # This one tests hidden-definitions working
+  define_method(:foo) do |_, _|
+    puts 'hello, world!'
+  end
```
jez, line 6: hidden definitions uses a completely different codepath, and so it was working already. but I figured I'd add a test anyway

gems/sorbet/lib/serialize.rb
```diff
@@ -174,7 +174,7 @@ def class_or_module(class_name)
       end
       next if denylisted_method(method)
       next if ancestor_has_method(method, Sorbet::Private::RealStdlib.real_singleton_class(klass))
-      serialize_method(method, true)
+      serialize_method(method, true, with_sig: false)
```
froydnj, line 177: If grepping is correct, it looks like these changes to pass `with_sig: false` make all the callsites pass `false`...is there any point in keeping the parameter and attendant code around?

froydnj (approved):
:stamp:

gems/sorbet/lib/serialize.rb
jez: Nope, there's no point. We could do that as a follow up. In fact I actually made that change 9 months ago in Stripe's internal copy of this file:

http://go/ps/commit/85eb538af75

Issue #4065, opened 2021-03-08
Special-case autocorrect for &:foo style blocks with T.must errors


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AT%3A%3AArray%5BT.nilable(Integer)%5D.new.map(%26%3Aeven%3F)%0A%0A%23%20Autocorrect%20this%20to%3A%0A%0AT%3A%3AArray%5BT.nilable(Integer)%5D.new.map%20%7B%7Cx%7C%20T.must(x).even%3F%7D)

```ruby
# typed: true

T::Array[T.nilable(Integer)].new.map(&:even?)

# Autocorrect this to:

T::Array[T.nilable(Integer)].new.map {|x| T.must(x).even?}
```

#### Observed output

```
editor.rb:3: Method even? does not exist on NilClass component of T.nilable(Integer) https://srb.help/7003
     3 |T::Array[T.nilable(Integer)].new.map(&:even?)
                                              ^^^^^^
  Got T.nilable(Integer) originating from:
    editor.rb:3:
     3 |T::Array[T.nilable(Integer)].new.map(&:even?)
                                              ^
  Autocorrect: Use `-a` to autocorrect
    editor.rb:3: Insert T.must()
     3 |T::Array[T.nilable(Integer)].new.map(&:even?)
                                              ^
Errors: 1
```

#### Expected behavior

The autocorrect should be aware of the `&:even?` block calling syntax, and instead edit the code to expand the block, and `T.must` the receiver directly.

- - -

We should do this by literally looking for the characters `(&:` preceding the name of the method in the source code (we don't have a better way to do it that's as little work) and followed by `)`.

jez:
Closed in #4067

PR #4067, opened 2021-03-08
Add autocorrect to expand block and add T.must

jez:
### Motivation

This makes a lot of codemods easier.

### Test plan

Review by commit for a before/after to the tests.


aprocter-stripe (approved):

jez:
Fixes #4065

PR #4068, opened 2021-03-08
add early filtering of final method names for final method checking

froydnj:
This PR attempts to make the inner loop of final method checking less costly by filtering out names that we've never seen declared as `final`.  Ideally, this should make `_on_method_added` checking much faster by never having to enter the doubly-nested loop in `_check_final_ancestors`, and the module-level checking we need to do much faster by reducing the number of method names we need to check for every ancestor.

### Motivation

Speed.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -104,7 +104,7 @@ module T::Private::Methods
       return
     end
     source_method_names.filter! do |method_name|
-      self.was_ever_final?(method_name)
+      was_ever_final?(method_name)
     end
     if source_method_names.empty?
       return
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -11,6 +11,9 @@ module T::Private::Methods
   # - they are done possibly before any sig block has run.
   # - they are done even if the method being defined doesn't have a sig.
   @final_methods = Set.new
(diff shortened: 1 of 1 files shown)


froydnj:
This change appears to be very effective in large codebases, so I think it's time to remove the `wip` tag.

elliottt (approved):
Nice! Would you mind adding a go link to a successful run on stripe ci?

froydnj:
> Nice! Would you mind adding a go link to a successful run on stripe ci?

https://go/builds/bui_J50R7ujxsNa25N

paracycle:
This PR seems to have made the runtime library only compatible with Ruby 2.6 and above, due to the use of `source_method_names.filter!`. Here is a test failure from `Tapioca` on 2.4: https://github.com/Shopify/tapioca/pull/253/checks?check_run_id=2080578832#step:7:1912

However, I see that `filter!` is actually an alias for `select!` and the latter exists all the way back to Ruby 2.4. Would be like me to change the method call in a PR to keep the backwards compatibility?

froydnj:
> This PR seems to have made the runtime library only compatible with Ruby 2.6 and above, due to the use of `source_method_names.filter!`. Here is a test failure from `Tapioca` on 2.4: https://github.com/Shopify/tapioca/pull/253/checks?check_run_id=2080578832#step:7:1912
> 
> However, I see that `filter!` is actually an alias for `select!` and the latter exists all the way back to Ruby 2.4. Would be like me to change the method call in a PR to keep the backwards compatibility?

I would be OK with that.

paracycle:
Already done: https://github.com/sorbet/sorbet/pull/4074

PR #3477, opened 2020-10-03
Added "User Defined Generics" documentation

AaronC81:
This adds the beginnings of documentation for user defined generic types. (Progress towards #1803.)

I don't know if "User Defined Generics" is the best title for this, since the principles also apply to any of the built-in generics, but I went with it to try to differentiate it from the `stdlib-generics.md` page. 

I'm also happy to add an unstable/subject to change/etc warning, if that is the case for generics.

### Motivation
As mentioned in the issue referenced above, generics are an important feature of Sorbet which come up a lot when dealing with RBIs or more advanced use cases of the type checker. Despite this, they don't appear to be documented anywhere currently, except with incidental uses of `type_member` in the documentation for `T.self_type` and in the error reference.

### Test plan
image

PR #4071, opened 2021-03-10
Added "User Defined Generics" documentation (reopened)

AaronC81:
**Reopen of #3477.**


aprocter-stripe:
I chatted with Stripe folks about this yesterday, and the sense here is that generics are still enough in flux that we probably do not want to document them in their current state. We think it would be best to wait on taking up this PR until generics are more stable.

smackesey:
@aprocter-stripe You guys should edit the current docs then, [a page in the current docs](https://sorbet.org/docs/stdlib-generics) claims that this functionality is supported. Led me on a bit of a wild goose chase. Hat tip to @AaronC81 for directing me to this PR from the Slack channel.

PR #4072, opened 2021-03-10
Move enclosingClass to SymbolRef and make Symbol.rebind a ClassOrModuleRef

jvilk-stripe:
Here are a few small improvements that involve the *Ref classes. This PR contains the following changes:

* Moves `enclosingClass` to `SymbolRef` and reimplements it without a loop.
  * Prepares us for having different underlying Symbol objects per-symbol-type.
* Changes `superClassOrRebind` to `ClassOrModuleRef`, since both uses of that field are class or modules.

### Motivation

Slowly moving us toward a world where we can have separate symbol classes, which reduces memory usage and allocation overhead.

### Test plan

Covered by existing tests

### DIFF
--- a/cfg/builder/builder_entry.cc
+++ b/cfg/builder/builder_entry.cc
@@ -29,7 +29,7 @@ unique_ptr<CFG> CFGBuilder::buildFor(core::Context ctx, ast::MethodDef &md) {
 
         auto selfClaz = md.symbol.data(ctx)->rebind();
         if (!selfClaz.exists()) {
-            selfClaz = md.symbol.data(ctx)->enclosingClass(ctx);
+            selfClaz = md.symbol.enclosingClass(ctx);
         }
         synthesizeExpr(
             entry, LocalRef::selfVariable(), core::LocOffsets::none(),
--- a/cfg/builder/builder_entry.cc
+++ b/cfg/builder/builder_entry.cc
@@ -29,12 +29,11 @@ unique_ptr<CFG> CFGBuilder::buildFor(core::Context ctx, ast::MethodDef &md) {
 
         auto selfClaz = md.symbol.data(ctx)->rebind();
         if (!selfClaz.exists()) {
-            selfClaz = md.symbol;
+            selfClaz = md.symbol.data(ctx)->enclosingClass(ctx);
         }
-        synthesizeExpr(entry, LocalRef::selfVariable(), core::LocOffsets::none(),
-                       make_unique<Cast>(LocalRef::selfVariable(),
-                                         selfClaz.data(ctx)->enclosingClass(ctx).data(ctx)->selfType(ctx),
-                                         core::Names::cast()));
+        synthesizeExpr(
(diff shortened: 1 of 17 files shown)


cfg/builder/builder_entry.cc
```diff
@@ -29,12 +29,11 @@ unique_ptr<CFG> CFGBuilder::buildFor(core::Context ctx, ast::MethodDef &md) {
 
         auto selfClaz = md.symbol.data(ctx)->rebind();
```
jvilk-stripe, line 30: `rebind` is already a `ClassOrModuleRef` with my change, so we don't need to call enclosingClass on it.

cfg/builder/builder_entry.cc
```diff
@@ -29,12 +29,11 @@ unique_ptr<CFG> CFGBuilder::buildFor(core::Context ctx, ast::MethodDef &md) {
...
         if (!selfClaz.exists()) {
-            selfClaz = md.symbol;
+            selfClaz = md.symbol.enclosingClass(ctx);
         }
-        synthesizeExpr(entry, LocalRef::selfVariable(), core::LocOffsets::none(),
-                       make_unique<Cast>(LocalRef::selfVariable(),
-                                         selfClaz.data(ctx)->enclosingClass(ctx).data(ctx)->selfType(ctx),
```
jvilk-stripe, line 36: I moved enclosingClass call to the if condition above.

core/Symbols.cc
```diff
@@ -1536,13 +1536,28 @@ void Symbol::sanityCheck(const GlobalState &gs) const {
...
-        owner = owner.data(gs)->owner;
+ClassOrModuleRef MethodRef::enclosingClass(const GlobalState &gs) const {
+    // Methods can only be owned by classes or modules.
+    return data(gs)->owner.asClassOrModuleRef();
+}
+
+ClassOrModuleRef SymbolRef::enclosingClass(const GlobalState &gs) const {
+    switch (kind()) {
+        case SymbolRef::Kind::ClassOrModule:
+            return asClassOrModuleRef();
```
jez, line 1547: Why is there no `owner` here anymore?

jez: I wonder if it makes sense to call `enclosingClass` on a `ClassOrModule` symbol. out of curiosity, can you try to ENFORCE that we never call it on one?

jez (approved):
One question, but otherwise lgtm

core/Symbols.cc
jvilk-stripe: Look at the old loop carefully:

```
SymbolRef owner = ref(gs);
while (!owner.isClassOrModule()) {
  //
}
return owner;
```

_It's not returning the owner.__ It's returning itself if it's already a class.

>I wonder if it makes sense to call enclosingClass on a ClassOrModule symbol.

It does! We do this all of the time. The common case is calling `enclosingClass` on `ctx.owner`, which is either a method or a class.

jvilk-stripe: Oh, wait, I see your question. Good question.

PR #4076, opened 2021-03-11
delay work in final method checks until we absolutely need it

froydnj:
We do a lot of computations during final method checks eagerly, potentially throwing away the results if we determine we don't need to do the full amount of checking.  We shouldn't do that, because in many cases the work can be delayed until we need it.

In `_check_final_ancestors`, we can probably avoid doing either the ancestors computation or determining instance methods, should one or the other say we don't need it.  I want to try and get numbers tomorrow; my hunch is that it's faster to compute instance methods, reject everything that's never been final, and then early exit if we don't have any methods to check.

### Motivation

Speed.

### Test plan

See included automated tests.


froydnj:
This makes things go faster and doesn't break anything, so let's put it up for review.

aprocter-stripe:
Probably best if we re-roll this one. :( I don't have much context on this.

jez (approved):

Issue #3855, opened 2020-12-31
File passes typecheck as strict, but suggest-typed marks it false


connorshea:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns(T%3A%3ABoolean)%20%7D%0A%20%20def%20bar%0A%20%20%20%20baz.qux%20%3D%3D%20'true'%0A%20%20end%0A%0A%20%20private%0A%0A%20%20sig%20%7B%20returns(Baz)%20%7D%0A%20%20def%20baz%0A%20%20%20%20T.unsafe(nil)%0A%20%20end%0Aend%0A%0Aclass%20Baz%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20params(other%3A%20Baz).returns(T%3A%3ABoolean)%20%7D%0A%20%20def%20%3D%3D(other)%0A%20%20%20%20T.unsafe(nil)%0A%20%20end%0A%0A%20%20sig%20%7B%20returns(T.nilable(T.any(String%2C%20Baz)))%20%7D%0A%20%20def%20qux%0A%20%20%20%20T.unsafe(nil)%0A%20%20end%0Aend%0A)

```ruby
# typed: strict
class Foo
  extend T::Sig

  sig { returns(T::Boolean) }
  def bar
    baz.qux == 'true'
  end

  private

  sig { returns(Baz) }
  def baz
    T.unsafe(nil)
  end
end

class Baz
  extend T::Sig

  sig { params(other: Baz).returns(T::Boolean) }
  def ==(other)
    T.unsafe(nil)
  end

  sig { returns(T.nilable(T.any(String, Baz))) }
  def qux
    T.unsafe(nil)
  end
end
```

#### Observed output

It passes the typecheck, but if you run `bundle exec srb rbi suggest-typed`, it'll mark the file as `# typed: false`.

#### Expected behavior

I'd expect it to stay as `# typed: strict` (or if it's set to `# typed: true`, I'd expect it to be upgraded to `# typed: strict`), because it's passing the typecheck.

- - -

My best guess for why this code is problematic is because of the `T.nilable(T.any())` on the `qux` method.

[...]

Issue #2496, opened 2020-01-16
[LSP]


SolaWing:
#### Input

`srb tc --lsp`
root directory contains a rbi folder

#### Observed output

error msg:  Sorbet's language server requires a single input directory.

use `srb tc --lsp --no-config --verbose`, seems rbi folder are implicit add to input directory

#### Expected behavior

can launch lsp

connorshea:
Assuming you mean that gems with an `rbi/` directory cause the LSP to break, I think this is fixed by #2730?

dogweather:
I'm seeing this too.

feliperaul:
Here as well

RB-Lab:
I'm trying to use it with this extension for VSCode https://github.com/danhuynhdev/sorbet-lsp, and still set this error. Sobet version in bundle is `sorbet (0.5.6307)`.

kbacha:
I'm seeing this issue as well, both in the gem that has the `rbi/` directory and a gem that pulls it in as a dependency.

Prajjwal:
Bump. Seeing the same error while trying to get it working with Vim + `ale`.

Fire-Dragon-DoL:
I'm seeing this problem too. I have a gem with `rbi` directory (needed to make typings work), but with --lsp, it fails.

I must point out that being able to add multiple directories to the configuration is very important, I think the `--lsp` option should keep supporting multiple directories. Is this possible?
Does `--ignore` support wildcards?

arubis:
With 7 confirmations (and add mine as 8th), I think it'd be appropriate to remove the `Unconfirmed` tag so this can get some love.

mcraeklaura:
You can add mine as a 9th as well :)

crokobit:
10th here :)

arpitsr:
11th here :). Please can you help here.

Sorbet's language server requires a single input directory. However, 0 are configured: []

styrmis:
Experiencing the same issue—this was working last week but reverting to an earlier version of Sorbet does not resolve the issue.

```
[T1266592][2022-06-21T09:59:09.801717] Debug logging enabled
[T1266592][2022-06-21T09:59:09.801778] Running sorbet version 0.5.10037 git 5395cb663a1fc8bf4ba7a129f4138e3596530d54 debug_symbols=true clean=0 with arguments: sorbet --lsp --no-config --verbose
[...]

Issue #2426, opened 2020-01-05
suggest-typed marks files using T.all as false


connorshea:
#### Input

This is related to #1951.

We start with this file, which comes from [the Intersection Types doc](https://sorbet.org/docs/intersection-types):

```ruby
# typed: true
extend T::Sig

module I1
  def i1; end
end

module I2
  def i2; end
end

class C
  include I1
  include I2
end

class D
  include I1
  include I2
end

sig { params(x: T.all(I1, I2)).void }
def foo(x)
  x.i1
  x.i2
end

foo(C.new)
foo(D.new)
```

It passes typechecking just fine, and I have no reason to believe it should be invalid.

The reason I believe the issue is coming from `T.all` specifically is because suggest-typed marks the file as `true` if you replace that line with `sig { params(x: T.untyped).void }`.

#### Observed output

After running suggest-typed, it becomes `typed: false`.

#### Expected behavior

I'd expect it to keep the file marked as `typed: true`.

I've created a repository for demonstrating this issue: https://github.com/connorshea/sorbet-t-all-suggest-typed-issue

In [the initial commit](https://github.com/connorshea/sorbet-t-all-suggest-typed-issue/commit/4033c1e7eecb5467eadc27c0884da52c8e2d5b8d), I init Sorbet and add the `foo.rb` file. Then in [the next commit](https://github.com/connorshea/sorbet-t-all-suggest-typed-issue/commit/247c7e47f326038c5a45c3afc0a9be73e0320468) I run `bundle exec srb rbi suggest-typed`, which causes `foo.rb` to become `typed: false`.

You can clone it and see the problem by running these commands:

```sh
git clone https://github.com/connorshea/sorbet-t-all-suggest-typed-issue
git checkout 4033c1e7eecb5467eadc27c0884da52c8e2d5b8d
bundle install
cat foo.rb #=> you'll see that it's `typed: true`
bundle exec srb tc #=> No errors! Great job.
bundle exec srb rbi suggest-typed
cat foo.rb #=> you'll see that it's now `typed: false`
```

### Other notes

I wanted to note here that this issue has been impacting the controllers in my Rails app since September, and I only just now went through the trouble of trying to determine the source of the issue by bisecting, which took a solid 45 minutes (and that's after I tried to do a bisect of the sorbet versions to see when the regression was introduced a few months ago).

It'd be great if suggest-typed had test coverage for issues like this to prevent them from happening in the future, because they're _incredibly_ annoying to track down.

Issue #1945, opened 2019-10-12
Huge increase (8 => 29) in false/ignore typed's on Oct. 8


jasnow:
Something happened on October 8 regarding sorbet-related upgrades, I went from
**8 false/ignore typed's to 29 false/ignore typed's** 
in one commit after upgrading the 3 sorbet gems to 0.4.4842 and sorbet-rails to 0.5.7 at 12:01 and then running "srb init"/etc and committing at 12:53. Here is the 12:53 commit:
https://github.com/jasnow/sample_app_4_2_rt_255/commit/4d7cb2305c65620afe395118f4ab10d9c23b10be

jasnow:
This is related to sorbet-rails' command "rake rails_rbi:all" and running "srb tc -a --suggest-typed --typed=strict --error-white-list=7022" (4 times) until "srb tc" is zero errors. I can still reproduce it.

Before
```
   2 # typed: autogenerated
   6 # typed: false
   1 # typed: ignore
  14 # typed: strict
  62 # typed: strong
  28 # typed: true
 113
```

After
```
   2 # typed: autogenerated
  18 # typed: false
  11 # typed: ignore
   1 # typed: strict
  57 # typed: strong
  24 # typed: true
 113
```

jasnow:
Conclusion: Only use "rake rails:custom" and "rake rails:all" if you need them.

PR #4078, opened 2021-03-11
Add option for not printing error sections to CLI

vinistock:
This PR adds the option `--no-error-sections` to the CLI to not print error sections.

It includes the new CLI option in the global state and then reads it in the `toString` method to decide whether to add the sections to the buffer or not.

Considering the following example:
```rb
# typed: true

class Foo
  def foo
  end
end

Foo.new.foos # typo
```
Without the new CLI option, the output is:
```bash
./example.rb:8: Method foos does not exist on Foo https://srb.help/7003
     8 |Foo.new.foos
        ^^^^^^^^^^^^
  Got Foo originating from:
    ./example.rb:8:
     8 |Foo.new.foos
        ^^^^^^^
    ./example.rb:8: Replace with foo
     8 |Foo.new.foos
                ^^^^
Errors: 1
```
With the new option, the output is:
```bash
./example.rb:8: Method foos does not exist on Foo https://srb.help/7003
     8 |Foo.new.foos
        ^^^^^^^^^^^^
Errors: 1
```

### Motivation

We'd like to have an option with a smaller amount of output for usage in conjunction with [Spoom](https://github.com/Shopify/spoom)

### Test plan

See included automated tests.

I have also verified that running with both `--no-error-sections` and auto correct works as expected (auto corrects, but doesn't print sections).

### DIFF
--- a/core/GlobalState.h
+++ b/core/GlobalState.h
@@ -251,6 +251,9 @@ public:
     // The error code is appended to this string.
     std::string errorUrlBase;
 
+    // If 'true', print error sections
+    bool includeErrorSections = true;
+
     // If 'true', enforce use of Ruby 3.0-style keyword args.
     bool ruby3KeywordArgs = false;
 
--- a/core/Error.cc
+++ b/core/Error.cc
@@ -112,9 +112,12 @@ string Error::toString(const GlobalState &gs) const {
         buf << '\n' << loc.toStringWithTabs(gs, 2);
     }
 
-    for (auto &section : this->sections) {
-        buf << '\n' << section.toString(gs);
+    if (gs.includeErrorSections) {
+        for (auto &section : this->sections) {
+            buf << '\n' << section.toString(gs);
+        }
     }
+
     return buf.str();
 }
 
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1842,6 +1842,7 @@ unique_ptr<GlobalState> GlobalState::deepCopy(bool keepId) const {
     result->kvstoreUuid = this->kvstoreUuid;
     result->lspTypecheckCount = this->lspTypecheckCount;
     result->errorUrlBase = this->errorUrlBase;
+    result->includeErrorSections = this->includeErrorSections;
     result->ignoredForSuggestTypedErrorClasses = this->ignoredForSuggestTypedErrorClasses;
     result->suppressedErrorClasses = this->suppressedErrorClasses;
     result->onlyErrorClasses = this->onlyErrorClasses;
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -409,6 +409,7 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
     options.add_options("dev")("error-black-list",
                                "(DEPRECATED) Alias for --suppress-error-code. Will be removed in a later release.",
                                cxxopts::value<vector<int>>(), "errorCode");
+    options.add_options("dev")("no-error-sections", "Do not print error sections.");
     options.add_options("dev")("typed", "Force all code to specified strictness level",
                                cxxopts::value<string>()->default_value("auto"), "{false,true,strict,strong,[auto]}");
     options.add_options("dev")("typed-override", "Yaml config that overrides strictness levels on files",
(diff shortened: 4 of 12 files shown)


paracycle:
As @vinistock stated in the PR description, when we do strictness bumps using Spoom, what Spoom does is to bump all the target files to the desired strictness level and rolls back the ones that are causing new errors at that level. Thus, in that mode, it would be great and much faster if Spoom could just parse the actual error output without all the error sections.

jez:
> what Spoom does is to bump all the target files to the desired strictness level and rolls back the ones that are causing new errors at that level

Does it make sense to build this into Sorbet? This is what the `--suggest-typed` mode is supposed to be doing.


> We'd like to have an option with a smaller amount of output for usage in conjunction with [Spoom](https://github.com/Shopify/spoom)

Does passing the Sorbet output through `... | sed -e '/^ /d'` achieve the same effect? (in general I'm suspicious of features that have alternative solutions with one line of bash)

jez:
If I recall correctly, Spoom already has ways to interact with Sorbet via `--lsp`. Would it make sense to use the structured, typed LSP diagnostic information reported by Sorbet in LSP mode, instead of parsing the plain text output?

Morriar:
> This is what the --suggest-typed mode is supposed to be doing.

Let’s put this back into context, this feature was created when `suggest-typed` was not reliable enough to provide a good experience when enabled on CI:

* Randomly bumping files to `typed: ignore`: https://github.com/sorbet/sorbet/issues/1945
* Reverting files to `typed: false` depending on the constructs used in the file: https://github.com/sorbet/sorbet/issues/2426 
* Inconsistent behavior with the result of typechecking: https://github.com/sorbet/sorbet/issues/3855 
* And more: https://github.com/sorbet/sorbet/issues/1519 

Another limitation we’ve observed with `suggest-typed` is aggressiveness. You generally start by committing a file with some content, Sorbet will mark it `typed: strong`, then the first thing you need to do is to down-type it to `true` so you can actually work in the file without producing hundreds of errors.

Considering all of this, it just proved more reliable to parse the type-checking output and bump it to exactly the desired strictness level.

> Does passing the Sorbet output through ... | sed -e '/^ /d' achieve the same effect?

Yes, you can also do wonderful things by re-parsing the output of Sorbet to make it easier to read when there are a lot of errors. However, this is still much slower than not generating the strings to just parse them out afterwards. Besides, we believe this flag is useful for more use-cases beyond Spoom, since not every user needs to see error sections for every use-case. Sometimes you are making some broad changes in the codebase (like bump a component’s files to `typed: true`) and want to be able to quickly see what errors are coming from where in the codebase and we’ve found that a more concise output really helps in those cases.

Here’s the output of vanilla Sorbet:

image

Here’s the output with `--no-error-sections`:

image

Finally, here’s the output thanks to `spoom tc`:

image


> Does it make sense to build this into Sorbet?

Regardless of Spoom being used, the option is still useful to make the output more readable. So let’s build _this_ inside Sorbet. Making `spoom bump` and `spoom tc` faster is just a happy accident.

For the general question, yes, it would be great to have something like `spoom bump` inside of Sorbet, with file path filters and tight control over “from” and “to” strictness levels or the filtering, sorting, formatting options from `spoom tc`. We would be interested in contributing these features. However, so far, we’ve had much faster iterations by building these experimental solutions on top of Sorbet as opposed to inside of Sorbet.

> If I recall correctly, Spoom already has ways to interact with Sorbet via --lsp. Would it make sense to use the structured, typed LSP diagnostic information reported by Sorbet in LSP mode, instead of parsing the plain text output?

It would, but it would also make `spoom` not work on projects with multiple directories since it’s [not supported by Sorbet](https://github.com/sorbet/sorbet/issues/2496). Still, however, this does not cover the use-case improvements that this PR is trying to introduce: a more readable output, regardless of the tooling used on top of it.

core/GlobalState.h
```diff
@@ -251,6 +251,9 @@ class GlobalState final {
     // The error code is appended to this string.
     std::string errorUrlBase;
 
+    // If 'true', do not print error sections
+    bool noErrorSections = false;
```
jez, line 255: Can you make this be

```suggestion
    bool includeErrorSections = true;
```

I generally don't like double negations: `!noErrorSections`.

jez (approved):
Thanks for the additional context, that helps a lot. My takeaway:

- `--suggest-typed` is bad, people don't like it, and we should fix it
- but prototyping things in spoom is faster than (potentially breaking) changes in sorbet
- when sorbet reports hundreds or thousands of errors, the output is needlessly verbose
- the speed up from skipping the error sections output with hundreds or thousands of errors is meaningful, and `sed` doesn't prevent the strings from being allocated so it doesn't have the same latency wins

This all makes sense. One small suggestion about naming, otherwise lgtm

elliottt (approved):
This looks good to me, and I like @jez's suggestion about removing the double negative.

core/GlobalState.h
vinistock: Sure! Do you mean changing it only in the global state or everywhere, including the CLI argument?

jez: I'm fine either way. You can either leave it as is, or change it to `--include-error-sections` with a default value of `true` using cxxopts.

vinistock: Switched the variable name for the global state, but kept the CLI option as `-no-error-sections`.

PR #4079, opened 2021-03-12
wip: Add sorbet/showSymbol method

### LINKED ISSUE #2937: [IDE] Copy full symbol path to clipboard feature
### Problem

When working in a large codebase, you often need to refer to classes/modules by their full path, e.g.:

```ruby
module Foo
  module Bar
    class Baz
    end
  end
end

module Loo
  def make_baz
    # This is tedious to type out. Autocomplete is of limited help if there are a lot of possible symbols.
    Foo::Bar::Baz.new
  end
end
```

### Proposed solution

In the IDE, include the ability to copy a symbol's full path to a clipboard. Users within Stripe have been requesting this ability.

A workaround right now is to manually copy some of the text reported by the `hover` feature.

Within Sorbet, this could be implemented as a custom LSP request, `sorbet/getFullSymbolPath`, that shares 99% of its implementation with hover.

The IDE extension would be responsible for hooking this up to the IDE UI and clipboard.

jez:
Review by commit, because I landed some pre-work first.

### Motivation

Fixes #2937

This custom LSP method takes a TextDocumentPositionParams and returns a
SymbolInformation (the symbol at that position). This is sufficient for
editors to build on top of this to implement functionality like "copy
fully-qualified name to clipboard" and I have a prototype of that working
in Vim with LanguageClient-neovim.

### TODO

- [ ] Write tests. Be sure to test edge cases, like what happens when
  there's no result. Can probably model this like the `hover` tests.
- [ ] Implement proper VS Code support (in Stripe's VS Code extension)


jez:
cc @jvilk-stripe

main/lsp/LSPPreprocessor.cc
```diff
@@ -166,52 +166,55 @@ unique_ptr<LSPTask> LSPPreprocessor::getTaskForMessage(LSPMessage &msg) {
...
-                return make_unique<ShutdownTask>(*config, id);
-            case LSPMethod::SorbetError:
-                return make_unique<SorbetErrorTask>(*config, move(get<unique_ptr<SorbetErrorParams>>(rawParams)), id);
-            case LSPMethod::GETCOUNTERS:
-                return make_unique<GetCountersTask>(*config, id);
-            case LSPMethod::TextDocumentPrepareRename:
-                return make_unique<PrepareRenameTask>(*config, id,
-                                                      move(get<unique_ptr<TextDocumentPositionParams>>(rawParams)));
-            case LSPMethod::TextDocumentRename:
-                return make_unique<RenameTask>(*config, id, move(get<unique_ptr<RenameParams>>(rawParams)));
```
jez, line 214: I sorted this list. See the commits for a better view.

jez:
@jvilk-stripe I think this is more or less good enough to land. I added some tests showcasing the known limitations. I can turn those into tickets if you want, fix them in this PR, or just check in the TODOs.

But I figure that this at least solves for the common case of wanting to copy the fully-qualified path of a constant (mostly), and that should be good enough to get the next nerd snipe rolling, namely finding someone who wants to implement the VS Code side of things (👉👃)

main/lsp/tools/make_lsp_types.cc
```diff
@@ -1373,6 +1374,7 @@ void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_pt
                                                 {"workspace/symbol", WorkspaceSymbolParams},
```
jvilk-stripe, line 1374: Can you also add `sorbetShowSymbolProvider` as an optional boolean to `ServerCapabilities`, and then update the initialize request to set it?

https://github.com/sorbet/sorbet/blob/40e3746f/main/lsp/requests/initialize.cc#L18

That will enable the client to conditionally surface the option to copy reference only if the server supports it.

test/testdata/lsp/show_symbol.rb
```diff
@@ -0,0 +1,58 @@
...
+      # TODO(jez) We should probably do method calls too, but I was lazy
+      foo
+      # ^ show-symbol: null
+
+      puts(:foo)
+      #    ^ show-symbol: null
+
+      sig {params(x: Integer).void}
+      def bar(x)
+        # TODO(jez) Does it make sense to just return the variable name here?
```
jvilk-stripe, line 52: I'd say no; it's not a symbol and doesn't have a FQN. Unless we want to make this request "show-name".

test/testdata/lsp/show_symbol.rb
```diff
@@ -0,0 +1,58 @@
+# typed: strict
+
+module First
```
jvilk-stripe, line 3: Can you do a packaging test too? Stripe has packaging mode on which has a material impact on this feature.

jvilk-stripe (changes requested):
I've identified one small blocking issue and had one test request.

jez:

→ https://go/builds/bui_KRhC8CjsBin59v
→ https://go/builds/bui_KRhC93enzLTTVu

PR #4084, opened 2021-03-15
Add missing block for `Pathname.glob` method

paracycle:
### Motivation
`Pathname.glob` takes a block, just like `Dir.glob`, but that was missing from the RBI definition. This has started causing problems since https://github.com/sorbet/sorbet/pull/4082 landed.

### Test plan
No extra tests.


froydnj:

→ https://go/builds/bui_J7Uz9W6WBZzOsL
→ https://go/builds/bui_J7UzkmHpBxJNn7

froydnj (approved):

PR #4086, opened 2021-03-15
Add blocks to some sigs that are missing them

RyanBrushett:
Following up on https://github.com/sorbet/sorbet/pull/4084, these are a few other methods that take blocks, but the block is missing from the RBI definition.

This has started causing problems since #4082 landed.

### Test plan

See included automated tests. They'll fail if the blocks are removed from the signatures.

### DIFF
--- a/rbi/stdlib/set.rbi
+++ b/rbi/stdlib/set.rbi
@@ -992,8 +992,8 @@ end
 #
 # for pack.c
 class Array
-  sig {returns(T::Set[T.untyped])}
-  def to_set; end
+  sig { params(blk: T.nilable(T.proc.params(arg0: T.untyped).void)).returns(T::Set[T.untyped]) }
+  def to_set(&blk); end
 end
 
 # The [`Enumerable`](https://docs.ruby-lang.org/en/2.7.0/Enumerable.html) mixin
--- a/rbi/stdlib/resolv.rbi
+++ b/rbi/stdlib/resolv.rbi
@@ -357,10 +357,11 @@ class Resolv::DNS
         String,
         { nameserver: T.any(String, T::Array[String]), search: T::Array[String], ndots: Integer },
         { nameserver_port: T::Array[[String, Integer]], search: T::Array[String], ndots: Integer }
-      )
+      ),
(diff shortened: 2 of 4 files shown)


jez (approved):
Thanks!

RyanBrushett:
Oops, test failure. On it.

RyanBrushett:
@jez Should be fixed. 👀 again if you don't mind?

aprocter-stripe (approved):

rbi/stdlib/set.rbi
```diff
@@ -992,8 +992,8 @@ end
 #
 # for pack.c
 class Array
-  sig {returns(T::Set[T.untyped])}
-  def to_set; end
+  sig { params(blk: T.nilable(T.proc.params(arg0: T.untyped).void)).returns(T::Set[T.untyped]) }
+  def to_set(&blk); end
```
paracycle, line 996: Btw, I am not even sure if `Array#to_set` is actually a thing. Looking at the Ruby `set` code, I [see](https://github.com/ruby/ruby/blob/master/lib/set.rb#L19):
> The method +to_set+ is added to Enumerable for convenience.

and only [a definition for `Enumerable#to_set`](https://github.com/ruby/ruby/blob/master/lib/set.rb#L684-L690) in the file.

Is this something that can be removed? @jez was this needed for some internal operation?

jez: I don't know about the historical context, it's possible it's incorrect.

paracycle: Yeah, I think it is best to remove the `Array` specialization, since the `Enumerable` one does the right thing. I believe it even passes the tests @RyanBrushett has added in this PR.

@RyanBrushett can you open a new PR with this specialization removed please? 🙏

jez: #4087

paracycle: You are too quick 😂

RyanBrushett: The tests I added will need to be removed/changed but otherwise that should work yep 👍

PR #4087, opened 2021-03-15
wip: Delete Array#to_set, leave Enumerable#to_set

jez:
### Motivation

https://github.com/sorbet/sorbet/pull/4086#discussion_r594708789

### Test plan

n/a


jez:

→ https://go/builds/bui_J7dkWh72t9N5xj
→ https://go/builds/bui_J7dk004VBmT2tX

rbi/stdlib/set.rbi
```diff
@@ -646,356 +646,6 @@ class SortedSet < Set
...
-# arr.keep_if {|a| a < 4}   #=> [1, 2, 3]
-# arr                       #=> [1, 2, 3]
-# ```
-#
-# for pack.c
-class Array
-  sig { params(blk: T.nilable(T.proc.params(arg0: T.untyped).void)).returns(T::Set[T.untyped]) }
-  def to_set(&blk); end
-end
-
```
paracycle, line 998: I think this will work fine, but we should also add a `sig` on the `Enumerable#to_set` to match the one that is being removed from `Array#to_set`. I guess the tests would be expecting the specific `Set` return.

paracycle: At the same time, `Enumerable#to_set` is a really weird method, if you use the `klass` argument and such:
```ruby
[5] pry(main)> class Foo; def initialize(x); @x = x end; end
[6] pry(main)> [1,2,3].to_set(Foo)
=> #<Foo:0x00007fd88d8f1598 @x=[1, 2, 3]>
```

Can I suggest that we forbid the usage of `klass` and `args` in `Enumerable#to_set` by intentionally making the sig `def to_set(&blk); end`?

paracycle: Oh, I see what the intent of the `klass` parameter is now. It is there to support `to_set(SortedSet)` and similar `Set` subclass usages. But, since Ruby does not have a way to codify that type requirement 😉, it is not explicit in the code. So, I suggest the following signature for `Enumerable#to_set`:
```ruby
module Enumerable
  # Makes a set from the enumerable object with given arguments. Needs to
  # +require "set"+ to use this method.
  sig { 
    type_parameters(:U, :V)
    .params(
      klass: T::Set[T.type_parameter(:U)],
      args: T.untyped,
      block: T.nilable(T.proc.params(arg0: T.type_parameter(:U)).returns(T.type_parameter(:V)))
    )
    .returns(T::Set[T.type_parameter(:V)]) 
  }
  def to_set(klass = _, *args, &block); end
end

Comment (jez) [reply rbi/stdlib/set.rbi:998]: With your edit I get

```
Malformed type declaration. Generic class without type arguments `Set` (fix available)
```

Do you have a suggestion here?

jez: (my default action right now is to leave it without a sig)

paracycle: Yeah, I think I really should have said `T.class_of(Set)`. I am fine either way with or without sigs for the moment. I can test and add a sig later if you want.

jez:

→ https://go/builds/bui_J7eseUgC2EEf9c
→ https://go/builds/bui_J7esmBh0L4Z7m1

jez:
closing due to staleness.

PR #4088, opened 2021-03-16
nano-optimize final method checks

froydnj:
Final method checks are (probably) still too slow.  We've taken care of -- I think -- all the "don't do unnecessary work" bits, so the only path left is to make the actual checking faster.

* b7dfd83 was the first stab at this: we're creating the lookup key for `@final_methods` twice: once in the hook, and once when we're running the method for the first time, plus we're uselessly grabbing the instance method just for the purposes of constructing that key.  Let's not do that by creating the key once for everything to use.
* 821197e removes some confusion: `T::Private::Final` has its own bits for determining whether a particular module was itself declared as `final`, and indeed the hooks already consult that.  So we should be safe in removing the bit that inserts the module into the table `@modules_with_final` maintained by `T::Private::Methods`.  Said table is now easier to describe: "the set of modules that have at least one method as `final`".
* 90eb529 is motivated by `@final_methods` not really being in the form we want for fast checks: we have to construct a string key out of a module and method_name, requiring string formatting and consing during final method checks.  (And we're repeatedly formatting the `owner_id` of the relevant ancestor for each possibly-final name, which is just wasted work.)  Let's not do that: instead, now that `@modules_with_final` has been clarified, we can turn it into a mapping from `Module` to the set of final method names defined on that module.  That turns our "is this method final on this module?" check into set membership, with no consing.  We are maintaining more information, which is not great for memory, but...
* 96dbe63 notices that, given the above change, we're not using `@final_methods` anymore, so we can delete all the code associated with that, which should make the overall memory changes...roughly neutral?
* cda482e recognizes that we can do slightly less work if we didn't have to do final method checks; since not doing final method checks is the common case, this is a very small speedup.

All of this ~should make things faster, but the only way to be sure is testing.

### Motivation

Speed.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -168,10 +168,6 @@ module T::Private::Methods
     @modules_with_final[mod.singleton_class]
   end
 
-  private_class_method def self.module_with_final?(mod)
-    @modules_with_final.include?(mod)
-  end
-
   # Only public because it needs to get called below inside the replace_method blocks below.
   def self._on_method_added(hook_mod, method_name, is_singleton_method: false)
     if T::Private::DeclState.current.skip_on_method_added
@@ -185,7 +181,7 @@ module T::Private::Methods
       raise "#{mod} was declared as final but its method `#{method_name}` was not declared as final"
     end
     # Don't compute mod.ancestors if we don't need to bother checking final-ness.
-    if was_ever_final?(method_name) && module_with_final?(mod)
+    if was_ever_final?(method_name) && @modules_with_final.include?(mod)
       _check_final_ancestors(mod, mod.ancestors, [method_name])
       # We need to fetch the active declaration again, as _check_final_ancestors
       # may have reset it (see the comment in that method for details).
@@ -428,8 +424,8 @@ module T::Private::Methods
   def self._hook_impl(target, singleton_class, source)
     # we do not need to call add_was_ever_final here, because we have already marked
     # any such methods when source was originally defined.
-    if !module_with_final?(target)
-      if !module_with_final?(source)
+    if !@modules_with_final.include?(target)
+      if !@modules_with_final.include?(source)
         return
       end
       note_module_deals_with_final(target)
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -144,14 +144,6 @@ module T::Private::Methods
     end
   end
 
-  private_class_method def self.add_was_ever_final(method_name)
-    @was_ever_final_names.add(method_name)
-  end
-
-  private_class_method def self.was_ever_final?(method_name)
-    @was_ever_final_names.include?(method_name)
-  end
-
   def self.add_module_with_final_method(mod, method_name, is_singleton_method)
     m = is_singleton_method ? mod.singleton_class : mod
     methods = @modules_with_final[m]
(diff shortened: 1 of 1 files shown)


jez:
@froydnj unassigning because title has `wip`

froydnj:
This branch has improved and is somewhat easier to explain since #4089 landed separately.  The main motivation we stated above still applies:

>  ... @final_methods not really being in the form we want for fast checks: we have to construct a string key out of a module and method_name, requiring string formatting and consing during final method checks. (And we're repeatedly formatting the owner_id of the relevant ancestor for each possibly-final name, which is just wasted work.) Let's not do that

but we've reworked things slightly because -- at least for now -- we really do need to track "modules with `final` methods" and "modules with `final` methods somewhere in their inheritance chain" simultaneously.  But we can at least track both of those pieces in a way such that we can do efficient checks for final method violations.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -160,13 +148,24 @@ def self._check_final_ancestors(target, target_ancestors, source_method_names)
...
+
+  def self.note_module_deals_with_final(mod)
+    # Side-effectfully initialize the value if it's not already there
+    @modules_with_final[mod]
+    @modules_with_final[mod.singleton_class]
   end
 
   private_class_method def self.module_with_final?(mod)
-    @modules_with_final.include?(mod)
+    @modules_with_final.has_key?(mod)
```
elliottt, line 168: Why change `include?` to `has_key?` here?

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -106,13 +105,10 @@ def self._check_final_ancestors(target, target_ancestors, source_method_names)
     # use reverse_each to check farther-up ancestors first, for better error messages. we could avoid this if we were on
     # the version of ruby that adds the optional argument to method_defined? that allows you to exclude ancestors.
     target_ancestors.reverse_each do |ancestor|
-      next if !module_with_final?(ancestor)
+      final_methods = @modules_with_final.fetch(ancestor, nil)
+      next if !final_methods
```
elliottt, line 109: This includes the exception that was outlined above where a `nil` entry means that the entry has final methods somewhere along its ancestor chain. Is it OK to continue in that case? If so, could you include a comment that explains why?

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -423,16 +422,14 @@ def self.run_all_sig_blocks
...
-      return
-    end
     # we do not need to call add_was_ever_final here, because we have already marked
     # any such methods when source was originally defined.
-    add_module_with_final(target)
-    install_hooks(target)
-
-    if !target_was_final
+    if !module_with_final?(target)
+      if !module_with_final?(source)
```
elliottt, line 428: Is it worth inlining the use of the hash here to avoid pushing function frames?
```suggestion
    if !@modules_with_final.has_key?(target)
      if !@modules_with_final.has_key?(source)
```

elliottt (approved):
Just a few questions, but otherwise this looks great!

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
froydnj: I thought I undid this, but apparently not.  I'll change it back.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
froydnj: It's OK to continue because either the ancestor has `final` methods somewhere along its ancestor chain, but no methods in itself, or the ancestor doesn't have `final` methods along its ancestor chain.  Either way, there are no methods to check for `final` violations in ancestor itself.  I'll add a comment.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
froydnj: Did this for all uses of `module_with_final?` and then removed the function, for consistency.

froydnj:
I went ahead and inlined references to `@was_ever_final_names` too for good measure.

PR #4089, opened 2021-03-16
minor tweaks to final method checks and final method tests

froydnj:
Some small changes:

* eb038f7: removes some confusion: `T::Private::Final` has its own bits for determining whether a particular module was itself declared as `sealed!`, and indeed the hooks already consult that. So we should be safe in removing the bit that inserts the module into the table `@modules_with_final` maintained by `T::Private::Methods`. Said table is now easier to describe: "the set of modules that have at least one method as final".
* b3393d7: small observation that we don't need to reset this unless we actually did final method checks.
* d892f15: some tests motivated by some failures that other changes of mine induced on Stripe's codebase, but not on our current tests.
* c739177: we're creating the lookup key for `@final_methods` twice: once in the hook, and once when we're running the method for the first time, plus we're uselessly grabbing the instance method just for the purposes of constructing that key. Let's not do that; instead, create the key once for everything to use.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/test/types/final_method.rb
+++ b/gems/sorbet-runtime/test/types/final_method.rb
@@ -220,6 +220,24 @@ class Opus::Types::Test::FinalMethodTest < Critic::Unit::UnitTest
     assert_equal(:class, c.foo)
   end
 
+  it "allows declaring a final instance method and a non-final class method with the same name" do
+    c = Class.new do
+      extend T::Sig
+      sig(:final) {returns(Symbol)}
+      def foo
+        :instance
+      end
+
+      sig {returns(Symbol)}
+      def self.foo
+        :class
+      end
+    end
+
+    assert_equal(:instance, c.new.foo)
+    assert_equal(:class, c.foo)
+  end
+
   it "forbids toggling a final method's visibility in a child class" do
     c = Class.new do
       extend T::Sig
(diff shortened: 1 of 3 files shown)


froydnj:

→ https://go/builds/bui_J8FtmwUxPRsujt
→ https://go/builds/bui_J8FtfTp5e6oQVx

froydnj:
`sorbet-runtime` upgrade build: https://go/builds/bui_J8G7sxhRCJtMVg

froydnj:
> Said table is now easier to describe: "the set of modules that have at least one method as final".

This should really be "the set of modules that have at least one final method somewhere in their ancestors chain" (which includes the module itself, of course).

gems/sorbet-runtime/test/types/final_method.rb
```diff
@@ -184,6 +184,42 @@ def bar; end
...
+
+      sig {returns(Symbol)}
+      def foo
+        :instance
+      end
+    end
+
+    assert_equal(:instance, c.new.foo)
+    assert_equal(:class, c.foo)
+  end
```
elliottt, line 221: Does it make sense to add a test for a final instance method, and non-final class method as well, just to complete the enumeration of cases?

elliottt (approved):
Looks good! One non-blocking comment about test cases.

gems/sorbet-runtime/test/types/final_method.rb
froydnj: Yup, done!

PR #4090, opened 2021-03-16
Add a redaction handler config hook

elliottt:
Make the redaction behavior for props with a redaction spec determined by the behavior set in `T::Configuration.redaction_handler`.

### Motivation

https://go/jira/RUBYLANG-209

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -433,6 +433,22 @@ module T::Configuration
     @sensitivity_and_pii_handler
   end
 
+  @redaction_handler = nil
+  # Set to a redaction handling function. This will be called when the
+  # `_redacted` version of a prop reader is used. By default this is set to
+  # `nil` and will raise an exception when the redacted version of a prop is
+  # accessed.
+  #
+  # @param [Lambda, Proc, nil] value Proc that converts a value into its
+  # redacted version according to the spec passed as the second argument.
+  def self.redaction_handler=(handler)
+    @redaction_handler = handler
+  end
+
+  def self.redaction_handler
+    @redaction_handler
+  end
+
   # Set to a function which can get the 'owner' of a class. This is
   # used in reporting deserialization errors
   #
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -451,7 +451,7 @@ class T::Props::Decorator
       value = self.public_send(prop_name)
       handler = T::Configuration.redaction_handler
       if !handler
-        raise RuntimeError, "No redaction handler configured"
+        raise "Using `redaction:` on a prop requires specifying `T::Configuration.redaction_handler`"
       end
       handler.call(value, redaction)
     end
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -449,8 +449,11 @@ class T::Props::Decorator
 
     @class.send(:define_method, redacted_method) do
       value = self.public_send(prop_name)
-      Chalk::Tools::RedactionUtils.redact_with_directive(
-        value, redaction)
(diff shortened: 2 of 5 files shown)


gems/sorbet-runtime/test/types/props/_props.rb
```diff
@@ -172,7 +172,29 @@ class TestRedactedProps
     prop :secret, String, redaction: [:truncate, 4], sensitivity: []
   end
 
+  describe 'redacted props with no redaction handler' do
+    it 'raises when fetching redacted values' do
+      assert_raises do
```
aprocter-stripe, line 177: Totally optional: might want to make sure the raised exception has expected class. (If one deleted the `if !handler` in the implementation, I think one would get a `NoMethodError` which is maybe not ideal.)

gems/sorbet-runtime/lib/types/configuration.rb
```diff
@@ -433,6 +433,22 @@ def self.normalize_sensitivity_and_pii_handler
...
+  # @param [Lambda, Proc, nil] value Proc that converts a value into its
+  # redacted version according to the spec passed as the second argument.
+  def self.redaction_handler=(handler)
+    @redaction_handler = handler
+  end
+
+  def self.redaction_handler
+    @redaction_handler
+  end
+
```
jez, line 451: You won't be able to call this method in pay-server until you edit `t.rbi`.

gems/sorbet-runtime/test/types/props/_props.rb
```diff
@@ -172,7 +172,29 @@ class TestRedactedProps
...
+        # this will raise an error without a redaction handler
+        d.str_redacted
+      end
+    end
+  end
+
   describe 'redacted props' do
+    before do
+      T::Configuration.redaction_handler = lambda do |value, redaction|
+        Chalk::Tools::RedactionUtils.redact_with_directive(value, redaction)
```
aprocter-stripe, line 190: Optional: Might be good to decouple this test from `Chalk::Tools::RedactionUtils` by defining a redaction function locally.

aprocter-stripe (approved):
LGTM! Feel free to ignore or punt on either/both of the nits.

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -449,8 +449,11 @@ def prop_defined(name, cls, rules={})
 
     @class.send(:define_method, redacted_method) do
       value = self.public_send(prop_name)
-      Chalk::Tools::RedactionUtils.redact_with_directive(
-        value, redaction)
+      handler = T::Configuration.redaction_handler
+      if !handler
+        raise RuntimeError, "No redaction handler configured"
```
jez, line 454:
```suggestion
        raise "Using `redaction:` on a prop requires specifying `T::Configuration.redaction_handler`"
```

`RuntimeError` is implied

gems/sorbet-runtime/test/types/props/_props.rb
jez, line 190: Can we take this opportunity to delete some stuff from `test_helper.rb`?

https://github.com/sorbet/sorbet/blob/6d88fd4451001b27bd0b0daa7d95e1d756330893/gems/sorbet-runtime/test/test_helper.rb#L31-L60

I'd love to just put this either in this test or into a helper function at the top of this file, rather than have it in a global `Chalk` constant.

gems/sorbet-runtime/test/types/props/_props.rb
elliottt: That's a good catch, thanks!

gems/sorbet-runtime/test/types/props/_props.rb
elliottt: This version is defined locally in `test_helper.rb`

elliottt: Good point, I'll inline it in the test.

jez (approved):

elliottt:
Stripe migration pr: https://go/ps/pull/312430

PR #4091, opened 2021-03-17
wip: type T::InterfaceWrapper#dynamic_cast

froydnj:
`T::InterfaceWrapper#dynamic_cast` appears to be `T.untyped`, which isn't helpful; we can do somewhat better by rewriting it into a type assertion (dynamically checked, unfortunately).  This at least lets us detect places where `dynamic_cast` could return `nil` but people are obviously not expecting it to.  The followup is to make such places use `non_nil_dynamic_cast`, which raises if `nil` would be returned.

### Motivation

We want to get rid of `wrap_instance` and `dynamic_cast`.  A good first step to doing that is making sure the latter is typed correctly for all existing uses, so they're easier to convert away.

A further ratchet down would be turning the cast into `T.let`, which ideally would be safe, but would probably turn up its own set of bugs.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -356,6 +356,10 @@ public:
         return Send2(loc, T(loc), core::Names::let(), std::move(value), std::move(type));
     }
 
+    static ExpressionPtr Cast(core::LocOffsets loc, ExpressionPtr value, ExpressionPtr type) {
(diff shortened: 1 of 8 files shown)


jez:
We deleted interface wrapper :tada:

PR #4092, opened 2021-03-17
FIx bundler with_*_env missing block in signatures

vinistock:
The bundler methods `with_*_env` all take a block that is missing from the signature. I added the block in and also added a signature for `with_unbundled_env`, since it's the new recommended method instead of `with_clean_env`.

### Motivation

Type checking is incorrectly failing saying that the method does not accept a block.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet/test/hidden-method-finder/simple/ruby_2_5_hidden.rbi.exp
+++ b/gems/sorbet/test/hidden-method-finder/simple/ruby_2_5_hidden.rbi.exp
@@ -1894,8 +1894,6 @@ module Bundler
   def self.unbundled_exec(*args); end
 
   def self.unbundled_system(*args); end
-
-  def self.with_unbundled_env(); end
 end
 
 class Class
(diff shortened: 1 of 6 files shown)


froydnj:

→ https://go/builds/bui_J8I1GJ5wqJMevu
→ https://go/builds/bui_J8I1lnoOsCvoMi

froydnj (changes requested):
It looks like there are some hidden-method-finder test failures, would you please take a look at those?  I guess the problem is that the expectations declare that there "should" be a `self.with_unbundled_env` available, but with the new declaration in the RBI, we can determine that there's not?

jez (approved):

vinistock:
@jez Thanks for pushing the test fix, I was trying to figure out how to properly generate the snapshots 😄.

froydnj (approved):

jez:
> Thanks for pushing the test fix, I was trying to figure out how to properly generate the snapshots 😄.

https://github.com/vinistock/sorbet/tree/520b3fd4a7bbe960ae93a42ec9dc79c128dcae99/gems/sorbet#updating-tests

PR #4101, opened 2021-03-19
Fix syntax in requires-ancestor docs

ryanwilsonperkin:
Remove trailing colon from method definition in the docs for `requires_ancestor`

### Motivation
Fix the code examples so they can be copy-pasted

### Test plan
N/A


Morriar (approved):
Thank you! 🙏

jez (approved):

Issue #4102, opened 2021-03-20
many warning during bundle install: ...gems/sorbet/bin/bootsnap does not exist, maybe gem pristine sorbet will fix it?...


dorianmariefr:
#### Input

Gemfile https://gist.github.com/ed61bb1592db29b09d188e0a5a5189b4

#### Observed output

```
Using sorbet-static 0.0.0 (x86_64-darwin-20) from https://github.com/sorbet/sorbet.git (at master@b2dc56d)
Using sorbet 0.0.0 from https://github.com/sorbet/sorbet.git (at master@b2dc56d)
`/Users/dorianmariefr/.rvm/gems/ruby-3.0.0/bundler/gems/sorbet-b2dc56d0760e/gems/sorbet/bin/bootsnap` does not exist, maybe `gem pristine sorbet` will fix it?
`/Users/dorianmariefr/.rvm/gems/ruby-3.0.0/bundler/gems/sorbet-b2dc56d0760e/gems/sorbet/bin/bundle` does not exist, maybe `gem pristine sorbet` will fix it?
`/Users/dorianmariefr/.rvm/gems/ruby-3.0.0/bundler/gems/sorbet-b2dc56d0760e/gems/sorbet/bin/byebug` does not exist, maybe `gem pristine sorbet` will fix it?
`/Users/dorianmariefr/.rvm/gems/ruby-3.0.0/bundler/gems/sorbet-b2dc56d0760e/gems/sorbet/bin/cap` does not exist, maybe `gem pristine sorbet` will fix it?
`/Users/dorianmariefr/.rvm/gems/ruby-3.0.0/bundler/gems/sorbet-b2dc56d0760e/gems/sorbet/bin/capify` does not exist, maybe `gem pristine sorbet` will fix it?
`/Users/dorianmariefr/.rvm/gems/ruby-3.0.0/bundler/gems/sorbet-b2dc56d0760e/gems/sorbet/bin/coderay` does not exist, maybe `gem pristine sorbet` will fix it?
`/Users/dorianmariefr/.rvm/gems/ruby-3.0.0/bundler/gems/sorbet-b2dc56d0760e/gems/sorbet/bin/commonmarker` does not exist, maybe `gem pristine sorbet` will fix it?
`/Users/dorianmariefr/.rvm/gems/ruby-3.0.0/bundler/gems/sorbet-b2dc56d0760e/gems/sorbet/bin/delayed_job` does not exist, maybe `gem pristine sorbet` will fix it?
[...]

PR #4103, opened 2021-03-20
Fix bundle install bin warnings

dorianmariefr:
Inline bin names to avoid searching to missing binaries

https://github.com/sorbet/sorbet/issues/4102

### Motivation

I was getting warning when installing the `sorbet` gem from GitHub.

### Test plan

No test, because I don't know how to test it, except that I got the warnings, and then didn't after the change.


dorianmariefr:
Also fixes `bundle exec srb init` not finding the `srb` executable

jez:
Sorbet is not designed to install in the way you mention in #4102, which is the reason why the `glob` line is messing up. We only support installing from pre-built `*.gem` archives, not from `github: 'sorbet/sorbet'` stanzas.

PR #4105, opened 2021-03-20
fix JSON loading error when encountering empty strings

dorianmariefr:
happens when requiring everything

I had an error when trying to load JSON with an empty string, probably because loading of dependent gems failed, not sure.

### Motivation

This is more to make you aware of the problem.

### Test plan

It should probably have tests, this isn't meant to be merged just to show you there is a place where an error can happen.

### DIFF
--- a/gems/sorbet/lib/require_everything.rb
+++ b/gems/sorbet/lib/require_everything.rb
@@ -131,7 +131,7 @@ class Sorbet::Private::RequireEverything
     #     ...
     #   ]
     # }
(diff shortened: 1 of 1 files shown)


jez:
The only way that command outputs an empty JSON file is if if the `srb` command crashed. If the `srb` command crashed, it likely printed something else in the error message that indicates what the problem was. You haven't shared the `srb` output so I can't tell what the problem is. In any case, this PR is not the solution. Feel free to open an issue.

PR #4107, opened 2021-03-23
add a cfg-text output format

froydnj:
Add `-p cfg-text`.

We talked about rewriting `cfg-raw` to be this format, but I chose to add another format.  Alternatives:

1. Make `cfg-raw` this format.
2. Make `cfg-text` and `cfg-text-raw` and delete `cfg-raw`, on the theory that nobody wants to look at the raw output in `dot` anyway.

It's also possible that folks might think `toString` and `toTextualString` should be `toDotString` and `toString`, respectively.  I have no strong opinions here.

### Motivation

Sometimes you don't want to fire up a `dot` viewer (e.g. you're working on a remote machine) and sometimes it's easier to just have a textual output format; `dot` graphs can get very unwieldy for even small-to-moderate size CFGs.

### Test plan

See included automated tests.


elliottt (approved):
I really like this new output format!

Would it be a big bother to add the back edge ids as a comment above each basic block?

PR #4109, opened 2021-03-24
Make isPackagerName work on singleton packager class names.

jvilk-stripe:
Make isPackagerName work on singleton packager class names.

### Motivation

@elliottt and @aprocter-stripe discovered this bug while showing me a different bug.

### Test plan

We don't have tests for this. I can add an explicit test into core/test/core_test.cc if it seems useful.


elliottt (approved):
Thanks for the quick fix!

PR #4120, opened 2021-03-29
Support comments in sorbet/config

### LINKED ISSUE #4106: Comments in `sorbet/config`
Hi @trevor-stripe 👋 

### Problem

```
Jared Beck Today at 3:37 PM
Is it possible to write comments in sorbet/config? I'd like to document why I'm `--ignore`ing certain paths.
The docs say
> The config file is actually just a newline-separated list of arguments to pass to srb tc, the same as if they’d been passed at the command line.
so I was not terribly surprised when Bash comments were treated as arguments.
Trevor Elliott  6 minutes ago
this is a pretty reasonable feature to add, would you mind making an issue?

https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1616441854001100
```

### Proposed solution

Bash/Ruby-style comments.

elliottt:
Support comments in `sorbet/config`, as lines that start with `#`. Note that this doesn't support trailing comments, as that would require a bit more thorough parsing of the config file.

### Motivation

Fix #4106

### Test plan

See included automated tests.

### DIFF
--- a/main/options/ConfigParser.cc
+++ b/main/options/ConfigParser.cc
@@ -12,6 +12,11 @@ namespace sorbet::realmain::options {
 namespace {
 
 // Returns true when the line starts with a '#', ignoring leading space.
+//
+// NOTE: we don't handle trailing comments because it would be difficult to handle the following case without a more
+// complicated parser:
+//
+// > -e 'puts "hello" # this is a comment' -p parse-tree
 bool isComment(string_view line) {
     auto hashPos = line.find('#');
     if (hashPos == string_view::npos) {
--- a/main/options/ConfigParser.cc
+++ b/main/options/ConfigParser.cc
@@ -1,11 +1,28 @@
 #include "ConfigParser.h"
+#include "absl/algorithm/container.h"
 #include "absl/strings/str_split.h"
 #include "common/FileOps.h"
 #include "common/common.h"
 #include "options.h"
+#include <cctype> // for isspace
 
 using namespace std;
 namespace sorbet::realmain::options {
+
+namespace {
+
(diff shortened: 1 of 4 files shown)


main/options/ConfigParser.cc
```diff
@@ -16,8 +33,8 @@ void ConfigParser::readArgsFromFile(std::shared_ptr<spdlog::logger> logger, stri
             argsPView = argsPView.substr(0, argsPView.size() - 1);
         }
         for (string_view arg : absl::StrSplit(argsPView, '\n')) {
-            if (arg.size() == 0) {
-                // skip empty line
+            if (arg.size() == 0 || isComment(arg)) {
```
Morriar, line 36: Can't you just strip the line from the position of the `#` to the end of the line?

That way you could support trailing comments too.

jez: No, because then if you tried to do

```
-e
"puts 'hello' # comment"
```

In the config file, it wouldn't work.

I'm 👍 to the approach trevor chose (I would have made it even more restrictive and required that the `#` be anchored at column `0`, because nowhere else in the file do we have anything that strips leading whitespace, but I don't care about that as strongly).

jez (approved):

main/options/ConfigParser.cc
elliottt: I'll put an additional comment in on `isComment` explaining why we don't allow trailing comments.

Morriar: Good catch! 👍

Morriar (approved):

PR #4132, opened 2021-04-06
Fix Kernel#catch signature

risentveber:
Was: catch returns `BaseObject`
Now: catch returns `T.untyped`

See https://apidock.com/ruby/Kernel/catch

### DIFF
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -206,9 +206,9 @@ module Kernel
   sig do
     params(
         tag: Object,
-        blk: T.proc.params(arg0: Object).returns(BasicObject),
+        blk: T.proc.params(arg0: Object).returns(T.untyped),
     )
-    .returns(BasicObject)
+    .returns(T.untyped)
   end
   def catch(tag=Object.new, &blk); end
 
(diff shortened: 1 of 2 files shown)


trevor-stripe (changes requested):
What's motivating this change? Could you include a bit of information about what use case this addresses for you?

Slike9:
> What's motivating this change? Could you include a bit of information about what use case this addresses for you?

We were adding types to a method containing `catch`, let's name it `foo`. It was surprising that `catch(tag, &block)` returned `BasicObject`, not `T.untyped`. So, we needed to apply `T.cast`, we can't apply `T.let` (as I know `T.let` is considered safer than `T.cast`):
```ruby
sig { returns(Integer) }
def foo
  bar = catch(:baz) { 10 }
  T.cast(bar, Integer) + 32
end
```

After the fix we could write:
```ruby
sig { returns(Integer) }
def foo
  bar = catch(:baz) { 10 }
  bar + 32
end
```

So, the fix allows us to avoid `T.cast` there. Further, `T.untyped` for `catch`'s result seems more natural.

elliottt:

→ https://go/builds/bui_JFrxWpVLjAs37s
→ https://go/builds/bui_JFrxJcOgqc5oXP

elliottt (approved):
Thank you!

trevor-stripe (approved):

Issue #4129, opened 2021-04-02
Implicitly returning `T.untyped` in one branch poisons all returns with untyped


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Breturns(Integer)%7D%0Adef%20silently_bad%0A%20%20if%20Random.rand(2).even%3F%0A%20%20%20%20'not%20an%20int'%0A%20%20else%0A%20%20%20%20T.unsafe(nil)%0A%20%20end%0Aend%0A%0Asig%20%7Breturns(Integer)%7D%0Adef%20loudly_bad%0A%20%20if%20Random.rand(2).even%3F%0A%20%20%20%20return%20'not%20an%20int'%0A%20%20else%0A%20%20%20%20return%20T.unsafe(nil)%0A%20%20end%0Aend)

```ruby
# typed: true
extend T::Sig

sig {returns(Integer)}
def silently_bad
  if Random.rand(2).even?
    'not an int'
  else
    T.unsafe(nil)
  end
end

sig {returns(Integer)}
def loudly_bad
  if Random.rand(2).even?
    return 'not an int'
  else
    return T.unsafe(nil)
  end
end
```

#### Observed output

```
editor.rb:16: Expected Integer but found String("not an int") for method result type https://srb.help/7005
    16 |    return 'not an int'
            ^^^^^^^^^^^^^^^^^^^
  Expected Integer for result type of method loudly_bad:
    editor.rb:14:
    14 |def loudly_bad
        ^^^^^^^^^^^^^^
  Got String("not an int") originating from:
    editor.rb:16:
    16 |    return 'not an int'
                   ^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

In general, unifying the result of an `if` expression where one of the branches is `T.untyped` will mean that the resulting expression's type overall is `T.untyped`.

But for the specific case where the conditional branch occurs as the last expression of a method body, we should be able to effectively make the code behave as if each branch of the conditional had been written with an explicit return the `loudly_bad` case above.

This way, we could prevent a single `T.untyped` branch from clobbering the type information across all branches.

- - -

qaisjp:
I'm seeing some interesting behaviour that I think is related to this issue.

In **both** of the below snippets, the revealed type of `@data` (via `T.reveal_type`) is `T.untyped`.

However, when you hover over the `@data`/`data` variable, you get different results.

[...]

PR #4134, opened 2021-04-07
wip: Don't let T.untyped poison method results

jez:
Marking WIP while I fix Stripe's codebase, but feel free to review in the mean time.

### Motivation

Fixes <https://github.com/sorbet/sorbet/issues/4129>

Note to self: this can likely fix at least part of #3482 (the part of that bug that affects reutrn method temps).

I recommend reviewing by commit, which will make the diff slightly cleaner and provide more context for what's changing.

### Test plan

See included automated tests.


jez:
- [ ] @jez Double check exp files.

test/testdata/cfg/dealias_with_return.rb.cfg.exp
```diff
@@ -55,7 +55,7 @@ subgraph "cluster_::Object#a" {
     "bb::Object#a_8" -> "bb::Object#a_6" [style="bold"];
     "bb::Object#a_9" [
         color = black;
-        label = "block[id=9, rubyBlockId=0](a: Integer(2))\l<statTemp>$14: Integer(3) = 3\l<returnMethodTemp>$2: Integer = a: Integer(2).+(<statTemp>$14: Integer(3))\l<finalReturn>: T.noreturn = return <returnMethodTemp>$2: Integer\l<unconditional>\l"
+        label = "block[id=9, rubyBlockId=0](a: Integer(2))\l<statTemp>$14: Integer(3) = 3\l<returnMethodTemp>$2: Integer = a: Integer(2).+(<statTemp>$14: Integer(3))\l<returnMethodTemp>$2: T.noreturn = return <returnMethodTemp>$2: Integer\l<finalReturn> = return <returnMethodTemp>$2\l<unconditional>\l"
```
jez, line 58: You'll notice two `return` instructions, one after another here. This happens because `builder_entry` unconditionally adds a final return, after synthesizing all instructions. Even on master, there will sometimes be cases where the CFG has a "double return" like this, if the user had written

```ruby
def foo
  return 0
end
```

directly. The second return is a stop-gap in case there was no explicit return from the method (e.g., an `if` that only had a `then` branch and an empty `else` branch).

Because `builder_entry` uses `synthesizeExpr` to call the make `return`, it is ignored by dead code checking, and thus ignored by inferrence.

jez: ... actually: this is a bug 😅 it manifested here but also in the `suggest-sig` test. I've since fixed the bug, which makes this comment obsolete.

jez:

→ https://go/builds/bui_JGQJfVK6RHCkeG
→ https://go/builds/bui_JGQJcEaRNKWxwI

jez:
This causes ~300 errors on Stripe's codebase

Questions:

- [ ] Does a sampling of them correspond with real errors? Or are there any bugs?
- [ ] How well does `--suggest-unsafe` silence them, to get back to the behavior before this PR?
  - [ ] Can we make `--suggest-unsafe` not wrap the thing if the `Loc` is equal to the method def's loc?
- [ ] Is the new "implicit `NilClass`" error confusing? Can we make it better?

aprocter-stripe (commented):
Just a couple comments on a first read-through, since it's still wip; I can give it another once-over when you think it's ready to go in. I'd probably want to drill down just a bit more on the specific cases of `eagerReturn` handling.

As I look through this I am finding myself a little confused that Sorbet defines `lub(?,X)` as `?` in the first place, but I'll unconfuse myself about that on my own time. :) Given that lub _is_ defined this way, I think this approach makes sense!

jez:

→ https://go/builds/bui_JW5QPoDCkTomC9
→ https://go/builds/bui_JW5Qoz6aXUOvi5

jez:

→ https://go/builds/bui_JbHHk0Cts1FPqn
→ https://go/builds/bui_JbHHUvUtKfJQel

jez:

PR #4136, opened 2021-04-09
--stripe-packages changes

ngroman:
### Motivation
Support the new design for `--stripe-packages` mode. This implements:
1. Package prefix enforcement: definitions in a package must be prefixed by the package's name
2. Export conflict checking: for a package, no export may be a prefix of another export. This is a special case of rule [3] but with much friendlier errors.
3. Import conflict checking: After merging together the exported constants from all imported packages, none may be prefixes of each other.

### Test plan

See included automated tests.


packager/packager.cc
```diff
@@ -599,13 +689,16 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file, const Pa
...
+ast::ParsedFile rewritePackagedFile(core::MutableContext ctx, ast::ParsedFile file, core::NameRef packageMangledName,
+                                    const PackageInfo *pkg) {
     if (ast::isa_tree<ast::EmptyTree>(file.tree)) {
         // Nothing to wrap. This occurs when a file is marked typed: Ignore.
         return file;
     }
 
     auto &rootKlass = ast::cast_tree_nonnull<ast::ClassDef>(file.tree);
+    EnforcePackagePrefix enforcePrefix(pkg);
+    file.tree = ast::TreeMap::apply(ctx, enforcePrefix, move(file.tree));
```
jvilk-stripe, line 701: Sorbet has two AST traversals:

* `ast::TreeMap` traverses the entire AST, including method bodies.
* `ast::ShallowMap` traverses just the outside definitions and doesn't traverse method bodies.

The latter is a bit faster. I think you don't need to traverse method bodies here.

```suggestion
    file.tree = ast::ShallowMap::apply(ctx, enforcePrefix, move(file.tree));
```

packager/packager.cc
```diff
@@ -231,6 +232,95 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
     return target;
 }
 
+struct EnforcePackagePrefix {
```
jvilk-stripe, line 235: You can make this a class; the traversal methods and constructor just need to be in a `public:` section.

packager/packager.cc
```diff
@@ -231,6 +232,95 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+
+    ast::ExpressionPtr preTransformClassDef(core::MutableContext ctx, ast::ExpressionPtr tree) {
+        auto &classDef = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+        if (classDef.symbol == core::Symbols::root()) {
+            // Ignore top-level <root>
+            return tree;
+        }
+        auto &pkgName = pkg->name.fullName.parts;
+        bool skipCheck = nameParts.size() > pkgName.size();
+        auto *constantLit = &ast::cast_tree_nonnull<ast::UnresolvedConstantLit>(classDef.name);
```
jvilk-stripe, line 280: You probably just want to take a reference here:

```suggestion
        auto &constantLit = ast::cast_tree_nonnull<ast::UnresolvedConstantLit>(classDef.name);
```

packager/packager.cc
```diff
@@ -231,6 +232,95 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+        pushConstantLit(constantLit);
```
jvilk-stripe, line 281: I see that this function accepts a pointer for convenience; you can take the pointer here:

```suggestion
        pushConstantLit(&constantLit);
```

test/cli/package-prefix-enforcement/nested/nested.rb
```diff
@@ -0,0 +1,54 @@
+# typed: strict
+
+extend T::Sig
+
+module Wrong
+  class Inside; end
+end
+
+module Root::Nested
```
jvilk-stripe, line 9: Does it work OK with:

```ruby
module Root
  module Nested
    class SomeClass; end
  end
end
```

packager/packager.cc
```diff
@@ -231,6 +232,95 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+struct EnforcePackagePrefix {
+    const PackageInfo *pkg;
+    vector<core::NameRef> nameParts;
+    int rootConsts = 0;
+
+    EnforcePackagePrefix(const PackageInfo *pkg) : pkg(pkg) {
+        ENFORCE(pkg != nullptr);
+    }
+
+    void pushConstantLit(ast::UnresolvedConstantLit *lit) {
```
jvilk-stripe, line 244: I have ideas for making this function faster once this pass stabilizes. :)

packager/packager.cc
```diff
@@ -231,6 +232,95 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+    }
+
+    ast::ExpressionPtr preTransformClassDef(core::MutableContext ctx, ast::ExpressionPtr tree) {
+        auto &classDef = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+        if (classDef.symbol == core::Symbols::root()) {
+            // Ignore top-level <root>
+            return tree;
+        }
+        auto &pkgName = pkg->name.fullName.parts;
+        bool skipCheck = nameParts.size() > pkgName.size();
```
jvilk-stripe, line 279: Can you skip pushing names, too?

jvilk-stripe: I think you just need an associated counter of skipped pushes so you can decrement those w/o popping names in postTransformClassDef.

test/cli/package-prefix-enforcement/nested/nested.rb
ngroman: added a test case

ngroman:
TODO:
- [ ] Clean up a few more failing tests
- [ ] Test out the LSP features to ensure hover on imports/exports works correctly.

test/cli/autogen-multipkg-autoloader/foo/__package.rb
```diff
@@ -3,5 +3,5 @@
 class Project::Foo < PackageSpec
   import Project::Bar
 
-  export Foo
+  export Project::Foo::FooClass
```
elliottt, line 6: Why are the full prefixes required in the `export` declaration now?

elliottt (changes requested):
I had a few questions, but am looking forward to seeing these changes land!

packager/packager.cc
```diff
@@ -231,6 +267,112 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+    void popConstantLit(ast::UnresolvedConstantLit *lit) {
+        while (lit != nullptr) {
+            nameParts.pop_back();
+            auto scope = ast::cast_tree<ast::ConstantLit>(lit->scope);
+            lit = ast::cast_tree<ast::UnresolvedConstantLit>(lit->scope);
+            if (scope != nullptr) {
+                ENFORCE(lit == nullptr);
+                ENFORCE(scope->symbol == core::Symbols::root());
+                rootConsts--;
+            }
```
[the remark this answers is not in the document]
ngroman: This doesn't work because I need to grab `scope` (`scope = lit->scope`) before updating lit.

packager/packager.cc
```diff
@@ -467,6 +593,117 @@ unique_ptr<PackageInfo> getPackageInfo(core::MutableContext ctx, ast::ParsedFile
...
+};
+
+void ImportTreeBuilder::mergeImports(core::Context ctx, const PackageInfo &importedPackage, core::LocOffsets loc) {
+    for (const auto &exportedFqn : importedPackage.exports) {
+        addImport(ctx, importedPackage, loc, exportedFqn);
+    }
+}
+
+void ImportTreeBuilder::addImport(core::Context ctx, const PackageInfo &importedPackage, core::LocOffsets loc,
+                                  const FullyQualifiedName &exportFqn) {
```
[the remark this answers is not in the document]
ngroman: Good catch, in an earlier iteration I was doing the conflict checking during `addImport` but ended up moving all that to the merge.

packager/packager.cc
```diff
@@ -231,6 +267,112 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+    }
+
+    ast::ExpressionPtr preTransformAssign(core::MutableContext ctx, ast::ExpressionPtr original) {
+        auto &asgn = ast::cast_tree_nonnull<ast::Assign>(original);
+        auto lhs = ast::cast_tree<ast::UnresolvedConstantLit>(asgn.lhs);
+        if (lhs != nullptr) {
+            auto &pkgName = pkg->name.fullName.parts;
+            size_t minSize = std::min(pkgName.size(), nameParts.size());
+            if (rootConsts == 0 &&
+                !std::equal(pkgName.begin(), pkgName.end(), nameParts.begin(), nameParts.begin() + minSize)) {
```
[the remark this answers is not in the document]
ngroman: You don't want minSize in that spot because that would allow for the definition:

```
class Foo
  X = 'something'
end
```
Inside code that was meant to be packaged inside `Foo::Bar`. (min size = 1, so we'd only compare that `[Foo] == [Foo]`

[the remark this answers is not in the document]
ngroman: I'll factor into funcs with more descriptive name to highlight the subtle difference.

packager/packager.cc
```diff
@@ -471,11 +608,141 @@ unique_ptr<PackageInfo> getPackageInfo(core::MutableContext ctx, ast::ParsedFile
...
+            return importLoc.exists();
+        }
+    };
+
+    // To avoid conflicts, a node should either be a leaf (source exists, no children) OR have children
+    // and an non-existent source. This is validated in `makeModule`.
+    UnorderedMap<core::NameRef, std::unique_ptr<ImportTree>> children;
+    Source source;
+
+public:
```
elliottt, line 627: Why make this private? It's already in an anonymous namespace, so it won't leak out beyond this file.

packager/packager.cc
```diff
@@ -631,9 +890,9 @@ vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers
                 if (result.gotItem()) {
                     filesProcessed++;
                     if (job.file.data(gs).sourceType == core::File::Type::Normal) {
-                        core::Context ctx(gs, core::Symbols::root(), job.file);
+                        core::MutableContext ctx(gs, core::Symbols::root(), job.file);
```
elliottt, line 893: What changed that a `MutableContext` is required now?

test/cli/autogen-multipkg-autoloader/autogen-multipkg-autoloader.sh
```diff
@@ -7,11 +7,12 @@ preamble="# frozen_string_literal: true
 
 rm -rf output
 mkdir -p output
+echo "TODO: fix this test once autogen-autoloader-packaged is supported"
 
 main/sorbet --silence-dev-message --stop-after=namer \
   --stripe-packages \
   -p autogen-autoloader:output \
-  --autogen-autoloader-modules={Foo,FooMethods,Bar,BarMethods} \
+  --autogen-autoloader-modules=Project \
```
elliottt, line 15: Does this change the intent of the original test?

test/cli/autogen-multipkg-autoloader/foo/__package.rb
elliottt: Do these still end up needing to be fully-qualified? Would it be easy to make them scoped as though they were defined under `Project::Foo`?

nroman-stripe:

→ https://go/builds/bui_JYd7bWbIzvwtK3
→ https://go/builds/bui_JYd7XU0NPeV41w

elliottt (changes requested):
Do we have any tests that document the expected structure of the import tree that gets generated? I was looking for something to check my understanding of what that tree would look like to aid LSP, since we had discussed that last week.

test/cli/autogen-multipkg-autoloader/foo/__package.rb
ngroman: It would be easy to go either way. My thought here was to force the FQ name for grep-ability

test/cli/autogen-multipkg-autoloader/autogen-multipkg-autoloader.sh
ngroman: Not really, now all the packages have the same root. That said this test is disabled for now.

ngroman: See https://github.com/sorbet/sorbet/pull/4136/files#diff-bb011a52ccb3f4f207975974228d74b84e99c231b375ea892430861c76de4fffR591

packager/packager.cc
elliottt: The location of this comment is a bit misleading, I meant to ask why the map and source were private :)

packager/packager.cc
ngroman: Ahh, I think I thought you needed Mutable for errors in a tree walk. Fixing.

ngroman:
> Do we have any tests that document the expected structure of the import tree that gets generated? I was looking for something to check my understanding of what that tree would look like to aid LSP, since we had discussed that last week.

You can see this structure in the `pass.package-tree.exp` tests. For example https://github.com/sorbet/sorbet/pull/4136/files/e945e43c1c03c1c0db64449b8f9a56d42aced67c#diff-ef2b42c002ea43885eccc896ab5d1686d210bb4b8d902a5b013e410eb814497dR73

packager/packager.cc
ngroman: While true, I think this highlights the expected use better.

test/cli/autogen-multipkg-autoloader/foo/__package.rb
elliottt: Is this easy to change if people complain in the future?

elliottt (approved):
Thanks for switching away from `MutableContext`, this looks great!

PR #4139, opened 2021-04-10
Add a test for final methods with `.checked(:never)`

elliottt:
Ensure that `.checked(:never)` and `sig(:final)` work together as expected.

### Motivation

More test coverage with `sig(:final)`.

### Test plan

See included automated tests.


jez (approved):
:phew: Glad to know that azdavis did his due-diligence with this feature!

PR #4140, opened 2021-04-10
refactor method definition in GlobalState

froydnj:
This change refactors the method definition in `GlobalState::initEmpty` to be somewhat less repetitive.  Whether this is better or not is a matter of taste, I suppose, but I think it's an improvement over mini blocks and `Loc::none()` everywhere.

### Motivation

I found `initEmpty` to be somewhat hard to follow; maybe this is an improvement.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -113,6 +113,79 @@ unique_ptr<GlobalState> GlobalState::makeEmptyGlobalStateForHashing(spdlog::logg
     return rv;
 }
 
+struct MethodBuilder {
+    GlobalState &gs;
+    MethodRef method;
+
+    MethodBuilder(GlobalState &gs, MethodRef m) : gs(gs), method(m) {}
+
+    MethodBuilder &defaultArg(NameRef name) {
+        auto &arg = gs.enterMethodArgumentSymbol(Loc::none(), method, name);
+        arg.flags.isDefault = true;
+        return *this;
+    }
+
+    MethodBuilder &blockArg() {
+        auto &arg = gs.enterMethodArgumentSymbol(Loc::none(), method, Names::blkArg());
+        arg.flags.isBlock = true;
+        return *this;
+    }
+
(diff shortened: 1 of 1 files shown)


core/GlobalState.cc
```diff
@@ -113,6 +113,79 @@ unique_ptr<GlobalState> GlobalState::makeEmptyGlobalStateForHashing(spdlog::logg
...
+
+    // Auto-conversion when we're done.
+    operator MethodRef() {
+        return method;
+    }
+};
+
+MethodBuilder def(GlobalState &gs, ClassOrModuleRef klass, NameRef name) {
+    return MethodBuilder{gs, gs.enterMethodSymbol(Loc::none(), klass, name)};
+}
```
jez, line 187: What do you think about giving this method a slightly longer / more descriptive name?

core/GlobalState.cc
```diff
@@ -113,6 +113,79 @@ unique_ptr<GlobalState> GlobalState::makeEmptyGlobalStateForHashing(spdlog::logg
...
+        auto &arg = gs.enterMethodArgumentSymbol(Loc::none(), method, name);
+        arg.flags.isDefault = true;
+        return *this;
+    }
+
+    MethodBuilder &blockArg() {
+        auto &arg = gs.enterMethodArgumentSymbol(Loc::none(), method, Names::blkArg());
+        arg.flags.isBlock = true;
+        return *this;
+    }
```
jez, line 132: Every method needs a block argument. What do you think about making the `.build()` be explicit, instead of happing on implicit conversion to a `MethodRef`, and moving this block arg stuff to that method?

jez (approved):
Generally love it. Two small suggestions.

core/GlobalState.cc
froydnj: WFM.  I was going for Ruby-ish, but I changed this to `enterMethod`.

core/GlobalState.cc
froydnj: Done.

PR #4145, opened 2021-04-14
LSP - Assume large changes cannot take the incremental (fast) path.

jvilk-stripe:
LSP - Assume large changes cannot take the incremental (fast) path.

To determine if a change (`SorbetWorkspaceEdit`) can be typechecked incrementally, the LSP server does the following:

* Parses the files.
* Calculates each file's file hash.
* Compares each file's hash against the previous version of the file's file hash.

This operation is serial and expensive if many files change. It is also unlikely that a large change can take the fast path. The user impact is that **Sorbet pauses for many seconds and does not respond to user input**.

To fix this problem, this PR does the following: If many files change, it assumes that the edit cannot take the fast path and _defers_ calculating file hashes (which are required for making the fast path determination for subsequent file edits) until just before the change typechecks, when a `WorkerPool` is available to parallelize the work.

### Motivation

After fixing a crash that is correlated with large file edits, I noticed an uptick in tail latency and memory consumption and user complaints about Sorbet being unresponsive.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPIndexer.h
+++ b/main/lsp/LSPIndexer.h
@@ -48,8 +48,11 @@ class LSPIndexer final {
      */
     bool canTakeFastPath(const LSPFileUpdates &edit,
                          const UnorderedMap<core::FileRef, std::shared_ptr<core::File>> &evictedFiles) const;
-    bool canTakeFastPath(const std::vector<std::shared_ptr<core::File>> &changedFiles,
-                         const UnorderedMap<core::FileRef, std::shared_ptr<core::File>> &evictedFiles) const;
+    /**
+     * INVARIANT: `changedFiles` must have hashes computed.
+     */
+    bool canTakeFastPathInternal(const std::vector<std::shared_ptr<core::File>> &changedFiles,
+                                 const UnorderedMap<core::FileRef, std::shared_ptr<core::File>> &evictedFiles) const;
 
 public:
     LSPIndexer(std::shared_ptr<const LSPConfiguration> config, std::unique_ptr<core::GlobalState> initialGS,
@@ -74,6 +77,7 @@ public:
      * Commits the given edit to `initialGS`, and returns a canonical LSPFileUpdates object containing indexed trees
      * and file hashes. Also handles canceling the running slow path.
      */
+    LSPFileUpdates commitEdit(SorbetWorkspaceEditParams &edit, WorkerPool &workers);
     LSPFileUpdates commitEdit(SorbetWorkspaceEditParams &edit);
 
     /**
--- a/main/lsp/notifications/initialized.cc
+++ b/main/lsp/notifications/initialized.cc
@@ -13,8 +13,8 @@ void InitializedTask::preprocess(LSPPreprocessor &preprocessor) {
 }
 
 void InitializedTask::index(LSPIndexer &indexer) {
-    // Hacky: We need to use the indexer, but with the WorkerPool from runSpecial. This is the only task to have this
-    // special requirement.
+    // Hacky: We need to use the indexer, but with the WorkerPool from runSpecial. This + SorbetWorkspaceEdit are the
+    // only tasks to have this special requirement.
     this->indexer = &indexer;
 }
 
--- a/main/lsp/notifications/sorbet_workspace_edit.cc
+++ b/main/lsp/notifications/sorbet_workspace_edit.cc
@@ -56,7 +56,7 @@ void SorbetWorkspaceEditTask::preprocess(LSPPreprocessor &preprocessor) {
 }
 
 void SorbetWorkspaceEditTask::index(LSPIndexer &indexer) {
-    if (params->updates.size() <= config.maxFilesOnFastPath) {
+    if (params->updates.size() <= config.opts.lspMaxFilesOnFastPath) {
         updates = make_unique<LSPFileUpdates>(indexer.commitEdit(*params));
     } else {
         // HACK: Too many files to `commitEdit` serially. Index in `runSpecial`.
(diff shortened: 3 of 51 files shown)


main/lsp/notifications/sorbet_workspace_edit.cc
```diff
@@ -119,7 +137,6 @@ bool SorbetWorkspaceEditTask::canTakeFastPath(const LSPIndexer &index) const {
         return updates->canTakeFastPath;
     }
     if (!cachedFastPathDecisionValid) {
-        index.computeFileHashes(params->updates);
```
jvilk-stripe, line 122: This responsibility has been moved into `canTakeFastPath` where it belongs.

main/lsp/notifications/sorbet_workspace_edit.cc
```diff
@@ -87,6 +94,17 @@ void SorbetWorkspaceEditTask::runSpecial(LSPTypechecker &typechecker, WorkerPool
     if (latencyTimer != nullptr) {
         latencyTimer->setTag("path", "slow");
     }
+    if (indexer) {
+        ENFORCE(updates == nullptr);
+        // Using the `indexer` here is safe; the indexing thread is blocked until `startedNotification.Notify` is called
+        // later in this function.
+        // This is really gnarly; there's got to be a cleaner way to do threading here. We can't move this out because
+        // we need `workers`, which is a resource that is explicitly managed by typechecking.
+        updates = make_unique<LSPFileUpdates>(indexer->commitEdit(*params, workers));
```
jvilk-stripe, line 103: Generally speaking, LSP has an indexer thread that runs `commitEdit` (to maintain a GlobalState with a name table that we use as the 'seed' for the slow path) and a typechecking thread that actually typechecks files. The typechecking thread owns and manages `WorkerPool`, hence the need to run this operation here.

This function runs on the typechecker thread, _but_ the indexer thread is blocked on `startedNotification` so we will not race on the indexer resource here.

We use this pattern in one other place -- initialization. The commonality is a need to parallelize indexer operations.

WorkerPool _is_ threadsafe, so we could consider sharing it with the dispatch logic. I resist doing that because I do not want it lurked into places it does not belong, causing pipeline stalls because new edits are waiting for the slow path to finish using the WorkerPool. It's possible that a simple constrained API would be fine for this purpose.

jvilk-stripe:
No longer WIP.

main/lsp/notifications/sorbet_workspace_edit.h
```diff
@@ -15,6 +15,8 @@ class SorbetWorkspaceEditTask final : public LSPDangerousTypecheckerTask {
     // Caches the fast path decision for the provided update. Becomes invalidated when the update changes.
     mutable bool cachedFastPathDecisionValid = false;
     mutable bool cachedFastPathDecision = false;
+    // HACK: In the event that this edit is too large to index serially, stash the indexer here for use in `runSpecial`.
+    LSPIndexer *indexer = nullptr;
```
jez, line 19: Are there lifetime problems here? Can the pointed to thing have been freed by the time the edit task wakes up and is processed?

main/lsp/notifications/initialized.cc
```diff
@@ -13,8 +13,8 @@ void InitializedTask::preprocess(LSPPreprocessor &preprocessor) {
 }
 
 void InitializedTask::index(LSPIndexer &indexer) {
-    // Hacky: We need to use the indexer, but with the WorkerPool from runSpecial. This is the only task to have this
-    // special requirement.
+    // Hacky: We need to use the indexer, but with the WorkerPool from runSpecial. This + SorbetWorkspaceEdit are the
+    // only tasks to have this special requirement.
```
jez, line 17: Not necessarily in this change, but given that this hack has grown, is there a restructuring that would make it easier to share the indexer rather than taking addresses of arguments?

main/lsp/notifications/sorbet_workspace_edit.h
jvilk-stripe: Nope. The thread that could do so is parked until the notification is hit.

main/lsp/notifications/initialized.cc
jvilk-stripe: Probably. I think I need to expose a thin API for specific indexer operations on the typechecker thread which avoids these hacky shenanigans, and then call into that from `request_dispatch` as a special case.

main/lsp/LSPIndexer.h
```diff
@@ -48,8 +48,11 @@ class LSPIndexer final {
      */
     bool canTakeFastPath(const LSPFileUpdates &edit,
                          const UnorderedMap<core::FileRef, std::shared_ptr<core::File>> &evictedFiles) const;
-    bool canTakeFastPath(const std::vector<std::shared_ptr<core::File>> &changedFiles,
-                         const UnorderedMap<core::FileRef, std::shared_ptr<core::File>> &evictedFiles) const;
+    /**
+     * INVARIANT: `changedFiles` must have hashes computed.
+     */
```
aprocter-stripe, line 53: Just OOC: This precondition has always been assumed, it just wasn't previously documented, right?

aprocter-stripe: Looks like yes, based on https://github.com/sorbet/sorbet/pull/4145/files#r613591907 .

aprocter-stripe (approved):
All seems reasonable to me!

main/lsp/LSPIndexer.h
jvilk-stripe: Correct.

PR #4158, opened 2021-04-18
Two sorbet-runtime call-validation optimizations

djudd-stripe:
Two changes:
1. Deprecate T::Profile, and rip out the sampled timing code from call validation.
2. Add a new semi-fast path for call validation with up to four positional arguments and no keyword arguments or other special cases, very similar to the existing fast path except without the restriction that types must be T::Types::Simple. (~~Also method-vs-procedure is left for a runtime check, for now, to avoid blowing up code size and because one extra branch didn't seem like a big deal relative to the rest.~~ edit: when merging the generator code, it was simpler to just leave this distinction in for both paths.)

These are really related only in intent; I actually implemented T::Profile support in the new fast-ish path before ripping it out. So I'm happy to split into separate PRs if folks prefer. 

### Motivation
1. I don't believe this timer measures anything useful. It doesn't account for a significant proportion of runtime typechecking overhead, probably the majority of overhead, because (a) it doesn't capture the impact of adding the `define_method` stack frame itself, and (b) it doesn't capture any impact of `T.let`, `T.must`, `T.unsafe` etc. A tool like StackProf will give a much better picture, and (if configured well) with much lower measurement overhead. Microbenchmarks show >20% speedup on single-argument methods.

2. I don't have hard statistics but my impression is that a large number of relatively-simple methods fail to qualify for the existing fast path because they use something like T::Boolean, T.nilable, or T.untyped, even as just one argument among otherwise simple ones. Microbenchmarks show a >40% speedup on single-nilable-argument methods. 

`bundle exec rake bench:typecheck` results for 2.7, slightly abridged, on:

master:
```
T::Types::Simple#valid?: 41.252 ns
T::Types::Union#valid?: 139.641 ns
T.let(..., Integer): 428.153 ns
sig {params(x: Integer).void}: 337.683 ns
T.let(..., T.nilable(Integer)): 1.036 μs
sig {params(x: T.nilable(Integer)).void}: 862.766 ns
T.let(..., Example): 409.354 ns
sig {params(x: Example).void}: 307.338 ns
T.let(..., T.nilable(Example)): 1.043 μs
sig {params(x: T.nilable(Example)).void}: 875.392 ns
sig {params(s: Symbol, x: Integer, y: Integer).void} (with kwargs): 1.504 μs
```

just the new fast path:
```
T::Types::Simple#valid?: 42.637 ns
T::Types::Union#valid?: 142.595 ns
T.let(..., Integer): 389.78 ns
sig {params(x: Integer).void}: 330.284 ns
T.let(..., T.nilable(Integer)): 1.004 μs
sig {params(x: T.nilable(Integer)).void}: 492.045 ns
T.let(..., Example): 396.391 ns
sig {params(x: Example).void}: 313.137 ns
T.let(..., T.nilable(Example)): 1.01 μs
sig {params(x: T.nilable(Example)).void}: 500.454 ns
sig {params(s: Symbol, x: Integer, y: Integer).void} (with kwargs): 1.507 μs
```

with both changes:
```
T::Types::Simple#valid?: 40.547 ns
T::Types::Union#valid?: 161.091 ns
T.let(..., Integer): 405.242 ns
sig {params(x: Integer).void}: 215.672 ns
T.let(..., T.nilable(Integer)): 1.038 μs
sig {params(x: T.nilable(Integer)).void}: 377.882 ns
T.let(..., Example): 393.038 ns
sig {params(x: Example).void}: 207.139 ns
T.let(..., T.nilable(Example)): 1.078 μs
sig {params(x: T.nilable(Example)).void}: 394.03 ns
sig {params(s: Symbol, x: Integer, y: Integer).void} (with kwargs): 1.424 μs
```

(Only the changes in the `sig` results could be meaningful, but I've left the others in as points of comparison & indicators of the noise level.)

### Test plan
Relying on existing tests


jez:
Do the api method tests use T::Profile right now in Stripe codebase? Is the plan to delete or change those tests?

djudd-stripe:
@jez they don't. If you're talking about the no-hot-runtime-checking tests, they use a custom mechanism so they can report more information about which methods are called. T::Profile is only used in a couple of production log-lines and I don't believe it provides real value there: https://livegrep.corp.stripe.com/search/stripe?q=T%3A%3AProfile%20repo%3Apay-server&fold_case=auto&regex=false&context=true

(It was maybe a bit more valuable before the introduction of the no-hot-runtime-checking tests, and when recursive checking was the default, so it was easier to introduce a regression and having a number right in the canonical log line could be valuable. I don't think this really applies anymore.)

gems/sorbet-runtime/lib/types/profile.rb
```diff
@@ -1,30 +1,25 @@
 # typed: true
 # frozen_string_literal: true
 
+# Deprecated, kept only for partial backwards compatibility
```
jez, line 4: When do you plan to remove this completely?

jez:
> T::Profile is only used in a couple of production log-lines and I don't believe it provides real value there

Gotcha, so it seems like it's basically just populating some log lines, and since you've made a case that those log lines are not even really accurate, that it's not worth keeping these things around.

gems/sorbet-runtime/tools/generate_call_validation.cc
```diff
@@ -121,17 +121,7 @@ void generateCreateValidatorFast(const Options &options, ValidatorKind kind, siz
     }
     fmt::print("&blk|\n");
 
-    fmt::print("      # This block is called for every `sig`. It's critical to keep it fast and\n"
-               "      # reduce number of allocations that happen here.\n"
```
jez, line 125: ooc why have these comments been dropped?

gems/sorbet-runtime/tools/generate_call_validation.cc
```diff
@@ -220,6 +197,108 @@ void generateCreateValidatorFast(const Options &options, ValidatorKind kind, siz
...
+               "          CallValidation.report_error(\n"
+               "            method_sig,\n"
+               "            message,\n"
+               "            'Return value',\n"
+               "            nil,\n"
+               "            return_type,\n"
+               "            return_value,\n"
+               "            caller_offset: -1\n"
+               "          )\n"
+               "        end\n"
```
jez, line 290: The other two (methods, procedures) take almost the same codepath but then branch on `validatorKind` in a few places.

It seems like the "medium" path is pretty similar to the fast path, differing only by

- the names of the individual methods
- the condition used to check the types

Do you think it's worth trying to share more code between the different kinds of validators?

Or at least factor out some of the shared code?

jez (commented):
Change looks mostly good to me. I have one question about rollout / deprecation strategy, and one question about how we can share a little more code.

Happy to approve after having a chat about those things.

gems/sorbet-runtime/tools/generate_call_validation.cc
djudd-stripe: It's not literally true - none of these individual methods is called for every `sig` anymore - and it didn't seem like it was adding enough value to be worth trying to reword.

gems/sorbet-runtime/tools/generate_call_validation.cc
djudd-stripe: The handling of `return_type` is different too, but that doesn't affect your point much.

I agree that more code could be shared; I started off with a separate method because I wasn't sure how much would end up being shared, but it ended up being most of it.

I think merging might make things less readable - I already find the switches on ValidatorKind a bit hard to read -but it would make it less likely that some difference could be introduced accidentally, so maybe it's worth it. WDYT?

gems/sorbet-runtime/tools/generate_call_validation.cc
jez: wfm

gems/sorbet-runtime/tools/generate_call_validation.cc
jez: i tend to think that in terms of "hard to read," if you're trying to change this code you're going to be looking at the generated code, imagining how you'd like the generated code to change, and back-solving for a diff to the the generator (i.e., the C++ code is not read in a vacuum, it is read in tandem with the generated code).

Given that, I place more value on the second part (introducing an accidental divergence).

So yeah, if you don't feel strongly one way or another i'd love to share more of the repetitive code between these two.

> but it ended up being most of it

for what it's worth, this is basically the same process and realization that I had when I wrote it initially (it started out as two, and then I realized that so little was different)

gems/sorbet-runtime/lib/types/profile.rb
jez: yeah i don't think we every documented or publicized this much, so i think it's safe to just remove the ~5 usages of it in Stripe's codebase pre-emptively and then land this PR.

djudd-stripe:
@jez updated to remove T::Profile fully and to share more call-validation-generator code. The diff does look much cleaner with the latter change. The Stripe codebase is also now prepped for this (PR just merged).

jez (approved):

PR #4160, opened 2021-04-20
Remove packages export_method support

ngroman:
### Motivation
This is no longer part of the design. This helps pave the way for https://github.com/sorbet/sorbet/pull/4136

### Test plan

See included automated tests.


ngroman:
cc @aisamanra

packager/packager.cc
```diff
@@ -292,29 +289,6 @@ struct PackageInfoFinder {
             }
         }
 
-        if (send.fun == core::Names::exportMethods()) {
```
jvilk-stripe, line 295: You can remove `exportMethods` from `generate_names.cc`:

https://github.com/sorbet/sorbet/blob/4f146522/core/tools/generate_names.cc#L399

And you can remove the symbol from GlobalState:
https://github.com/sorbet/sorbet/blob/4f146522/core/GlobalState.cc#L454

You'll have to delete this method which returns the export_method symbolref:
https://github.com/sorbet/sorbet/blob/4f146522/core/SymbolRef.h#L848

and shift the MethodRefs in that file with later IDs up one (e.g., `Class_new` becomes ID 8).

jvilk-stripe (approved):
Thanks! You should finish removing exportMethods from the name table and symbol table too before landing.

PR #4167, opened 2021-04-21
move deserialization errors to a separate method

froydnj:
### Motivation

This PR essentially reimplements #4148, but puts the error handler not on `T::Configuration`.  I will quote the motivation from that PR below:

> It seems better to have a separate path for deserialization errors, rather than forcing everything to go through `soft_assert_handler`. I also have some changes to how deserialization works where it would be nice if the error reporting were wrapped up in a single method to call, rather than having to manually reproduce the `T::Configuration.soft_assert_handler` call.

> I don't know whether/how much it actually matters, but one nice side-effect of this change is that the generated deserialization code becomes relatively smaller in cases where we would be throwing errors.

It would be nice if it were not quite so complicated to get to `raise_deserialization_error`, but that is life.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/private/deserializer_generator.rb
+++ b/gems/sorbet-runtime/lib/types/props/private/deserializer_generator.rb
@@ -57,15 +57,11 @@ module T::Props
               begin
                 #{transformation}
               rescue NoMethodError => e
-                T::Configuration.soft_assert_handler(
-                  'Deserialization error (probably unexpected stored type)',
-                  storytime: {
-                    klass: self.class,
-                    prop: #{prop.inspect},
-                    value: val,
-                    error: e.message,
-                    notify: 'djudd'
-                  }
+                self.class.decorator.raise_deserialization_error(
+                  self.class,
+                  #{prop.inspect},
+                  val,
+                  e,
                 )
                 val
               end
--- a/gems/sorbet-runtime/lib/types/props/private/deserializer_generator.rb
+++ b/gems/sorbet-runtime/lib/types/props/private/deserializer_generator.rb
@@ -57,8 +57,7 @@ module T::Props
               begin
                 #{transformation}
               rescue NoMethodError => e
-                self.class.decorator.raise_deserialization_error(
-                  self.class,
+                raise_deserialization_error(
                   #{prop.inspect},
                   val,
                   e,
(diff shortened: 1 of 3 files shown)


elliottt (approved):
Looks good!

gems/sorbet-runtime/lib/types/props/private/deserializer_generator.rb
```diff
@@ -57,15 +57,11 @@ def self.generate(props, defaults)
...
-                T::Configuration.soft_assert_handler(
-                  'Deserialization error (probably unexpected stored type)',
-                  storytime: {
-                    klass: self.class,
-                    prop: #{prop.inspect},
-                    value: val,
-                    error: e.message,
-                    notify: 'djudd'
-                  }
+                self.class.decorator.raise_deserialization_error(
```
jez, line 60: Should this really live on the `decorator`? Wouldn't that mean that we're expecting to allow people to customize this?

I think that if we put the `raise_deserialization_error` on `DeserializerGenerator` we can call it like

```
T::Props::Private::DeserializerGenerator.raise_deserialization_error
```

and not have to worry about indirections. (tbh, i have no real idea what `decorator` is, and even though it's probably not possible anymore, i kind of still wish for us to remove it at some point, so i try to find ways to avoid mentioning it in new code).

jez (approved):
One comment about whether we want `self.class.decorator` or whether we want to drop the indirections, but otherwise looks good to me.

froydnj:
d3ff527 moves `raise_deserialize_error` to `T::Props::Serializable`, which also enables us to make the error case somewhat smaller.  That at least limits what decorators are responsible for so we don't make more work for ourselves later.

It does mean that subclasses can override `raise_deserialize_error`; this is also a possibility with the existing `required_prop_missing_from_{de,}serialize`.  I wasn't thinking about this possibility when I put `raise_deserialize_error` on `T::Props::Serializable`.  I can understand if we want to limit the extensibility surface (even via overriding private methods), though.  wdyt @jez ?

(Arguably, `self.class.decorator.raise_nil_deserialize_error` should probably live on `T::Props::Serializable`, but that is a discussion for another day.)

jez (approved):

jez:
> wdyt @jez ?

lgtm thanks for tweaking

PR #4169, opened 2021-04-22
Add `resultType` and `resultTypeUntyped` methods to MethodBuilder

elliottt:
Add methods to `MethodBuilder` to support setting the `resultType` of the symbol being defined. `resultTypeUntyped` is a special case, as it requires the `MethodRef` of the method being defined.

### Motivation

More consistent api for `MethodBuilder`.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -110,6 +110,16 @@ struct MethodBuilder {
         return *this;
     }
 
+    MethodBuilder &resultTypeUntyped() {
+        method.data(gs)->resultType = Types::untyped(gs, method);
+        return *this;
+    }
+
+    MethodBuilder &resultType(TypePtr &&type) {
+        method.data(gs)->resultType = type;
+        return *this;
+    }
+
     MethodRef build() {
         auto &arg = gs.enterMethodArgumentSymbol(Loc::none(), method, Names::blkArg());
         arg.flags.isBlock = true;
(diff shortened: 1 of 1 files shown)


core/GlobalState.cc
```diff
@@ -110,6 +110,16 @@ struct MethodBuilder {
...
 
+    MethodBuilder &resultTypeUntyped() {
+        method.data(gs)->resultType = Types::untyped(gs, method);
+        return *this;
+    }
+
+    MethodBuilder &resultType(TypePtr &&type) {
+        method.data(gs)->resultType = type;
+        return *this;
+    }
```
froydnj, line 121: Could these be done as `buildWithUntypedResult` and `buildWithResult(TypePtr &&type)` and thus replace `build()`?

elliottt: Seems reasonable to me!

elliottt: I couldn't remove `build` completely, but I did make these two functions return the `MethodRef` instead.

aprocter-stripe (approved):
I think I actually like `resultType` a little better than `buildType`, unless we need to force setting of the result type to be the last step for some safety reason I don't know about. Don't have strong feelings, though, and things are definitely cleaner after this change than they were before it.

elliottt:
> I think I actually like `resultType` a little better than `buildType`, unless we need to force setting of the result type to be the last step for some safety reason I don't know about. Don't have strong feelings, though, and things are definitely cleaner after this change than they were before it.

I like it too, but this does make it harder to miss setting it, and also makes it so that you can't overwrite it without having to jump through a lot of hoops.

Issue #230, opened 2019-03-18
T.proc.bind only works on block parameter


XrXr:
```ruby
# typed: true

extend(T::Sig)

class A
  def super_secret_method
    puts "Very secret"
  end
end

sig do
  params(
    scope: T.proc.bind(A).void,
    block: T.proc.bind(A).void
  ).void
end
def has_many(scope, &block)
  a = A.new
  a.instance_exec(&scope)
  a.instance_exec(&block)
end

has_many(->{ super_secret_method }) do
  super_secret_method
end
```

```
hello.rb:23: Method super_secret_method does not exist on T.class_of(<root>) http://go/e/7003
    23 |has_many(->{ super_secret_method }) do
                     ^^^^^^^^^^^^^^^^^^^
    https://git.corp.stripe.com/stripe-internal/sorbet/tree/master/rbi/core/kernel.rbi#L275: Did you mean: Kernel#protected_methods?
     275 |  def protected_methods(all=T.unsafe(nil)); end
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This has_many is very similar to the real [has_many in Active Record](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_and_belongs_to_many-label-Scopes) which takes both a Proc and a block.

jez:
"Dupe" of #4095

PR #4170, opened 2021-04-22
Add T.bind for changing bindings inside procs and lambdas

### LINKED ISSUE #4095: Error if `.bind` used with non-block argument
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20%23%20it%20doesn't%20make%20sense%20to%20use%20%60.bind%60%20for%0A%20%20%23%20anything%20other%20than%20a%20%60%26blk%60%20argument%0A%20%20sig%20%7Bparams(f%3A%20T.proc.bind(A).void).void%7D%0A%20%20def%20self.takes_fn(f)%0A%20%20end%0Aend%0A%0Aclass%20B%0A%20%20def%20main%0A%20%20%20%20%23%20self%20is%20not%20%60A%60%20(does%20not%20match%20%60.bind(A)%60%20in%20sig)%0A%20%20%20%20f%20%3D%20-%3E%20%7BT.reveal_type(self)%7D%0A%0A%20%20%20%20%23%20Sorbet%20does%20not%20(cannot)%20report%20an%20error%20here%0A%20%20%20%20A.takes_fn(f)%0A%0A%20%20%20%20%23%20Sorbet%20inference%20algorithm%20does%20not%20use%20argument%0A%20%20%20%20%23%20type%20of%20%60takes_fn%60%20when%20typechecking%20body%20of%0A%20%20%20%20%23%20function%20value%20passed%20as%20positional%20argument.%0A%20%20%20%20A.takes_fn(-%3E%20%7BT.reveal_type(self)%7D)%0A%20%20end%0Aend)

```ruby
# typed: true

class A
  extend T::Sig

  # it doesn't make sense to use `.bind` for
  # anything other than a `&blk` argument
  sig {params(f: T.proc.bind(A).void).void}
  def self.takes_fn(f)
  end
end

class B
  def main
    # self is not `A` (does not match `.bind(A)` in sig)
    f = -> {T.reveal_type(self)}

    # Sorbet does not (cannot) report an error here
    A.takes_fn(f)

    # Sorbet inference algorithm does not use argument
    # type of `takes_fn` when typechecking body of
    # function value passed as positional argument.
    A.takes_fn(-> {T.reveal_type(self)})
  end
end
```

#### Observed output

```
editor.rb:16: Revealed type: B https://srb.help/7014
    16 |    f = -> {T.reveal_type(self)}
                    ^^^^^^^^^^^^^^^^^^^
  Got B originating from:

editor.rb:24: Revealed type: B https://srb.help/7014
    24 |    A.takes_fn(-> {T.reveal_type(self)})
                           ^^^^^^^^^^^^^^^^^^^
  Got B originating from:
Errors: 2
```

#### Expected behavior

It doesn't make sense to use `.bind` on a non-block argument, because Sorbet does not use it in the inference algorithm.

We should ban usages of `.bind` on non-block arguments, because it probably means the user is doing something they didn't mean to.

That is, the `self.takes_fn` definition should have an error.

- - -

Separately, it might make sense to have some way to get a `Proc` object (value) where the block *was* typechecked under a `.bind` constraint. You could imagine a feature akin to `xs = T::Array[Integer].new`, like:

```ruby
class B
  def main
    f = proc do
      T.reveal_type(self) # => `B`
    end

    g = T.proc.bind(A).void do
      T.reveal_type(self) # => `A`
    end
  end
end
```

vinistock:
This PR introduces `T.bind` outside of the context of method signatures to allow users to bind procs and lambdas to different types. This is especially useful in the Rails world, where a lot of procs are executed with `instance_exec`.

For example,
```rb
class Post < ApplicationRecord
  # In callbacks from ActiveRecord, ActiveModel and Actionpack, `if` and `unless` options
  # can be passed as lambdas to determine whether it should run or not. These keys are part
  # of keyword arguments and the block is captured and then executed in the context of the instance.
  #
  # Sorbet doesn't know that it will be executed in an instance context and errors with
  # "method `should_do_it?` is not defined for T.class_of(Post)
  before_create :do_something, if: -> { should_do_it? }

  private

  def should_do_it?
    true
  end
end
```

With the changes in this PR, users can indicate the correct binding inside each proc, guaranteeing that type checking is verifying against the right types. Below is an example usage and more examples can be found in the tests included.
```rb
class Post < ApplicationRecord
  # Proc is now bound to `Post` and not `T.class_of(Post)`. No errors!
  before_create :do_something, if: -> { T.bind(self, Post); should_do_it? }

  private

  def should_do_it?
    true
  end
end
```

### Motivation

Sorbet currently does not allow indicating the context in which a captured block will run, which limits typing in common Rails use cases, such as the previously mentioned callbacks, but also active support concerns.

### Test plan

The included tests display examples for concerns and callbacks.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -760,6 +760,16 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                 current = walk(cctx.withTarget(tmp), c.arg, current);
                 if (c.cast == core::Names::uncheckedLet()) {
                     current->exprs.emplace_back(cctx.target, c.loc, make_unique<Ident>(tmp));
+                } else if (c.cast == core::Names::bind()) {
+                    if (c.arg.isSelfReference()) {
+                        auto self = cctx.inWhat.enterLocal(core::LocalVariable::selfVariable());
+                        current->exprs.emplace_back(self, c.loc, make_unique<Cast>(tmp, c.type, core::Names::cast()));
+                        current->exprs.emplace_back(cctx.target, c.loc, make_unique<Ident>(self));
+                    } else {
+                        if (auto e = cctx.ctx.beginError(what.loc(), core::errors::CFG::MalformedTBind)) {
+                            e.setHeader("T.bind can only be used with `self`");
+                        }
+                    }
                 } else {
                     current->exprs.emplace_back(cctx.target, c.loc, make_unique<Cast>(tmp, c.type, c.cast));
                 }
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -767,7 +767,7 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                         current->exprs.emplace_back(cctx.target, c.loc, make_unique<Ident>(self));
                     } else {
                         if (auto e = cctx.ctx.beginError(what.loc(), core::errors::CFG::MalformedTBind)) {
-                            e.setHeader("T.bind can only be used with `self`");
+                            e.setHeader("`{}` can only be used with `{}`", "T.bind", "self");
                         }
                     }
                 } else {
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -147,7 +147,7 @@ module T
     Private::Casts.cast(value, type, cast_method: "T.let")
   end
 
-  # Tells the typechecker to bind the current proc to the type `type`.
+  # Tells the type checker to treat `self` in the current block as `type`.
   # Useful for blocks that are captured and executed later with instance_exec.
   # Use like:
   #
@@ -156,10 +156,10 @@ module T
   #    ...
   #  end
   #
-  # Compared to `T.cast`, `T.let` is _checked_ by static system.
+  # `T.bind` behaves like `T.cast` in that it is assumed to be true statically.
   #
   # If `checked` is true, raises an exception at runtime if the value
-  # doesn't match the type.
+  # doesn't match the type (this is the default).
   def self.bind(value, type, checked: true)
     return value unless checked
 
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -147,6 +147,25 @@ module T
     Private::Casts.cast(value, type, cast_method: "T.let")
   end
 
+  # Tells the typechecker to bind the current proc to the type `type`.
+  # Useful for blocks that are captured and executed later with instance_exec.
+  # Use like:
+  #
+  #  seconds = lambda do
+  #    T.bind(self, NewBinding)
+  #    ...
+  #  end
+  #
+  # Compared to `T.cast`, `T.let` is _checked_ by static system.
+  #
+  # If `checked` is true, raises an exception at runtime if the value
+  # doesn't match the type.
+  def self.bind(value, type, checked: true)
+    return value unless checked
+
+    Private::Casts.cast(value, type, cast_method: "T.bind")
+  end
+
   # Tells the typechecker to ensure that `value` is of type `type` (if not, the typechecker will
   # fail). Use this for debugging typechecking errors, or to ensure that type information is
   # statically known and being checked appropriately. If `checked` is true, raises an exception at
--- a/gems/sorbet-runtime/test/types/configuration.rb
+++ b/gems/sorbet-runtime/test/types/configuration.rb
@@ -15,6 +15,23 @@ module Opus::Types::Test
       def self.receive(*); end
     end
 
+    module Readable; end
+    module Writable
+      def write
+        T.bind(self, T.all(Readable, Writable))
+        true
+      end
+    end
+
+    class BadArticle
+      include Writable
+    end
+
+    class GoodArticle
+      include Writable
+      include Readable
+    end
+
     describe 'inline_type_error_handler' do
       describe 'when in default state' do
         it 'T.must raises an error' do
@@ -27,6 +44,33 @@ module Opus::Types::Test
           assert_raises(TypeError) do
             T.let(1, String)
           end
+        end
+
+        it 'T.bind raises an error if block is executing on the wrong type' do
+          block = -> {T.bind(self, String); upcase}
+
+          assert_raises(TypeError) do
+            123.instance_exec(&block)
+          end
+        end
+
+        it 'T.bind raises an error if block is executing on the wrong type' do
+          block = -> {T.bind(self, String); upcase}
+
+          assert_raises(TypeError) do
+            123.instance_exec(&block)
+          end
+        end
+
+        it 'T.bind raises error if type constraints are not all satisfied' do
+          bad_article = BadArticle.new
+
+          assert_raises(TypeError) do
+            bad_article.write
+          end
+
+          good_article = GoodArticle.new
+          assert good_article.write
         end
       end
 
--- a/gems/sorbet-runtime/test/types/configuration.rb
+++ b/gems/sorbet-runtime/test/types/configuration.rb
@@ -54,14 +54,6 @@ module Opus::Types::Test
           end
         end
 
-        it 'T.bind raises an error if block is executing on the wrong type' do
-          block = -> {T.bind(self, String); upcase}
-
-          assert_raises(TypeError) do
-            123.instance_exec(&block)
-          end
-        end
-
         it 'T.bind raises error if type constraints are not all satisfied' do
           bad_article = BadArticle.new
 
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -928,7 +928,7 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
         core::TypeAndOrigins tp;
         bool noLoopChecking = cfg::isa_instruction<cfg::Alias>(bind.value.get()) ||
                               cfg::isa_instruction<cfg::LoadArg>(bind.value.get()) ||
-                              cfg::isa_instruction<cfg::LoadSelf>(bind.value.get());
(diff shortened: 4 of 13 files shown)


infer/environment.cc
```diff
@@ -928,6 +928,7 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
         core::TypeAndOrigins tp;
         bool noLoopChecking = cfg::isa_instruction<cfg::Alias>(bind.value.get()) ||
                               cfg::isa_instruction<cfg::LoadArg>(bind.value.get()) ||
+                              bind.bind.variable == cfg::LocalRef::selfVariable() ||
                               cfg::isa_instruction<cfg::LoadSelf>(bind.value.get());
```
jez, line 932: I am pretty sure this line is redundant with the `variable == selfVariable` check on the line above. Did you add it back because you found that it was not in fact redundant?

test/testdata/bound_proc.rb
```diff
@@ -0,0 +1,63 @@
+# typed: true
```
jez, line 1: Can you please copy the entire test case that I wrote in my branch into this PR?

gems/sorbet-runtime/test/types/configuration.rb
```diff
@@ -28,6 +28,12 @@ def self.receive(*); end
             T.let(1, String)
           end
         end
+
+        it 'T.bind raises an error' do
+          assert_raises(TypeError) do
+            T.bind(self, String)
+          end
+        end
```
jez, line 36: Two things:

Can you add a test that mimics more normal usage? i.e., something that actually uses instance_exec or instance_eval or something like that.

Can you also add a test using this inside a method at the top level? i.e., not inside a block:

```ruby
module Mixin1; end
module Mixin2
  def foo
    T.bind(self, T.all(Mixin1, Mixin2))
  end
end

class BadExample
  include Mixin2
end
BadExample.new.foo

class GoodExample
  include Mixin1
  include Mixin2
end
GoodExample.new.foo
```

gems/sorbet-runtime/lib/types/_types.rb
```diff
@@ -147,6 +147,12 @@ def self.let(value, type, checked: true)
     Private::Casts.cast(value, type, cast_method: "T.let")
   end
 
+  def self.bind(value, type, checked: true)
```
jez, line 150: Can you add a comment here, like for all the other things in this file?

infer/environment.cc
```diff
@@ -1281,7 +1282,7 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
 
                 const core::TypeAndOrigins &ty = getAndFillTypeAndOrigin(ctx, c->value);
                 ENFORCE(c->cast != core::Names::uncheckedLet());
```
jez, line 1284: Do you notice how here we assert that `uncheckedLet` doesn't make it this far? That's because we remove references to `uncheckedLet` in the `builder_walk` phase so that instead of 4 kinds of casts to consider in `infer`, there are only 3.

Can you change builder walk to effectively convert `T.bind`-style cast `Instruction`s to `T.cast`-style cast `Instruction`s? This way, we can be sure that future changes to `infer` never have to handle `T.bind` specially, and thus the behavior of `T.cast` and `T.bind` never diverge in ways we don't want them to.

test/testdata/bound_proc.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+end
+
+module Readable
+  extend Concern
+
+  included do
+    before_create :do_this
+  # ^^^^^^^^^^^^^^^^^^^^^^ error: Method `before_create` does not exist on `T.class_of(Readable)`
+  end
```
jez, line 15: These tests don't look like the tests that you had written in your PR description.

We very much prefer test cases that mimic real-world usage.

Doesn't Sorbet already have a `rewriter` pass to support `Concern`'s `included do` pattern? I get that it's possible to use `T.bind` for this example, but I don't think that in practice people will be using `T.bind` for this particular use case much, over the rewriter pass (correct me if I'm mistaken).

With that in mind, the example in your PR description with the `before_create` hook looked more representative, so I'd love if you could update your test with those examples.

test/testdata/bound_proc.rb
```diff
@@ -0,0 +1,63 @@
...
+end
+
+class Article
+  def self.before_save(something); end
+
+  include Writable
+  include Shareable
+
+  PROC = lambda do
+    T.bind(self, Article)
```
Morriar, line 58: Can we also add a test showing the expected behaviour when passing something else than `self` to `T.bind` please 🙏

Morriar:
I think this PR is also providing a solution for https://github.com/sorbet/sorbet/issues/230?

gems/sorbet-runtime/test/types/configuration.rb
vinistock: Added, let me know if it's the right style.

test/testdata/bound_proc.rb
vinistock: I updated the tests a little bit, however, I must argue that they do reflect real world usage. The `Concern` rewriter is for handling class methods - not the included block which Sorbet still cannot understand the correct binding.

Additionally, defining callbacks inside included blocks, such as `before_create` is a common use case of Rails. I added your tests and also changed the examples a bit, but again they do reflect day to day usage of Rails.

test/testdata/bound_proc.rb
vinistock: Added that case.

vinistock:
Updated the PR description to also close #230, which indeed will be fixed by this.

@jez should I add documentation about `T.bind` to the website code in this PR or in a subsequent one?

test/testdata/bound_proc.rb
jez: Cool thanks for the context!

jez:
> @jez should I add documentation about `T.bind` to the website code in this PR or in a subsequent one?

I prefer in this PR, but if you wanted to do it in a follow up PR I wouldn't block this PR on it.

cfg/builder/builder_walk.cc
```diff
@@ -760,6 +760,16 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
...
                 if (c.cast == core::Names::uncheckedLet()) {
                     current->exprs.emplace_back(cctx.target, c.loc, make_unique<Ident>(tmp));
+                } else if (c.cast == core::Names::bind()) {
+                    if (c.arg.isSelfReference()) {
+                        auto self = cctx.inWhat.enterLocal(core::LocalVariable::selfVariable());
+                        current->exprs.emplace_back(self, c.loc, make_unique<Cast>(tmp, c.type, core::Names::cast()));
+                        current->exprs.emplace_back(cctx.target, c.loc, make_unique<Ident>(self));
+                    } else {
+                        if (auto e = cctx.ctx.beginError(what.loc(), core::errors::CFG::MalformedTBind)) {
+                            e.setHeader("T.bind can only be used with `self`");
```
jez, line 770:
```suggestion
                            e.setHeader("`{}` can only be used with `{}`", "T.bind", "self");
```

The

```
`{}`
```

bits will mean that those regions show up with terminal colors.

gems/sorbet-runtime/lib/types/_types.rb
```diff
@@ -147,6 +147,25 @@ def self.let(value, type, checked: true)
     Private::Casts.cast(value, type, cast_method: "T.let")
   end
 
+  # Tells the typechecker to bind the current proc to the type `type`.
+  # Useful for blocks that are captured and executed later with instance_exec.
+  # Use like:
```
jez, line 152:
```suggestion
  # Tells the type checker to treat `self` in the current block as `type`.
  # Useful for blocks that are captured and executed later with instance_exec.
  # Use like:
```

gems/sorbet-runtime/lib/types/_types.rb
```diff
@@ -147,6 +147,25 @@ def self.let(value, type, checked: true)
...
+  #
+  #  seconds = lambda do
+  #    T.bind(self, NewBinding)
+  #    ...
+  #  end
+  #
+  # Compared to `T.cast`, `T.let` is _checked_ by static system.
+  #
+  # If `checked` is true, raises an exception at runtime if the value
+  # doesn't match the type.
```
jez, line 162:
```suggestion
  # `T.bind` behaves like `T.cast` in that it is assumed to be true statically.
  #
  # If `checked` is true, raises an exception at runtime if the value
  # doesn't match the type (this is the default).
```

gems/sorbet-runtime/test/types/configuration.rb
```diff
@@ -28,6 +45,33 @@ def self.receive(*); end
...
+
+        it 'T.bind raises an error if block is executing on the wrong type' do
+          block = -> {T.bind(self, String); upcase}
+
+          assert_raises(TypeError) do
+            123.instance_exec(&block)
+          end
+        end
```
jez, line 63: This test is identical to the previous test.

jez (approved):
This looks great, thanks for those changes!

Some small suggested changes, and then waiting to hear back w.r.t. whether you want to do docs in this PR or another. I can merge as is or wait for the docs, just let me know.

vinistock:
Cool! I'll add the documentation to this PR and address the remaining comments, so that we can ship it all in one go.

vinistock:
Alright, added website documentation. Let me know if the copy looks good!

jez:
@vinistock Does this actually close #4095?

That issue requests a static error when using `T.proc.bind` on a non-block argument.

This PR implements an alternative for people who think that they want the behavior they imagine would be possible with `T.proc.bind` on a non-block argument.

I don't think we should expand the scope of this PR, but I am going to mark this PR as not closing that issue.

jez (approved):

jez:

→ https://go/builds/bui_JOUwqSiKKgp6Fd
→ https://go/builds/bui_JOUwYlCyr4Mq8r

jez:
Thanks!

PR #4171, opened 2021-04-22
add a T::Configuration option for VM prop serde and appropriate modifications

froydnj:
### Motivation

We have some changes to prop serialization/deserialization that are faster than the `sorbet-runtime` codegen and avoid the use of `class_eval`.  To be able to reasonably test those changes, we'd like to add a `T::Configuration` option that selects between the status quo and the accelerated routines.

### Test plan

There's not much to test here.  I did verify that the `test` and `test_vm_serde` Rakefile targets pass (well, subject to error handling problems with the VM-defined routines) on an appropriately configured Ruby.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/serializable.rb
+++ b/gems/sorbet-runtime/lib/types/props/serializable.rb
@@ -214,8 +214,13 @@ module T::Props::Serializable::DecoratorMethods
     rules[:serialized_form] = rules.fetch(:name, prop.to_s)
     res = super
     prop_by_serialized_forms[rules[:serialized_form]] = prop
-    enqueue_lazy_method_definition!(:__t_props_generated_serialize) {generate_serialize_source}
-    enqueue_lazy_method_definition!(:__t_props_generated_deserialize) {generate_deserialize_source}
+    if T::Configuration.use_vm_prop_serde?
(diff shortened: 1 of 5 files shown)


gems/sorbet-runtime/lib/types/props/serializable.rb
```diff
@@ -230,6 +235,18 @@ def add_prop_definition(prop, rules)
     )
   end
 
+  private def generate_serialize2
+    T::Props::Private::SerializerGenerator.generate2(decorated_class, props)
+  end
```
elliottt, line 240: Where is `generate2` defined?

elliottt (changes requested):
This makes sense to me. Could you point me to the definition of `generate2`? I was curious to see how it differed from the existing generate/eval implementation.

Would you mind adding a test to the runtime test suite that exercises the errors we'd expect to see if the vm doesn't support deserialization?

froydnj:
> This makes sense to me. Could you point me to the definition of `generate2`? I was curious to see how it differed from the existing generate/eval implementation.

Pointed this out offline, since it's currently in Stripe-internal systems.

> Would you mind adding a test to the runtime test suite that exercises the errors we'd expect to see if the vm doesn't support deserialization?

Pushed bits with a test and a reworking of the Rakefile to placate Sorbet.

elliottt (approved):

PR #3313, opened 2020-07-22
ruby27: Add begin-less ranges support

Morriar:
### Motivation

New try on the begin-less ranges (https://github.com/sorbet/sorbet/pull/3260).

The problem raised by @elliottt with:

```ruby
low = T.unsafe(0)
high = T.unsafe(16)
T.reveal_type(low..high) # T::Range[T.noreturn], but T::Range[T.untyped] on master
```

Was happening because with the proposed signatures for `Range.new` the parameter type could not resolve when one of the end was `NilClass` and Sorbet used `bottom`.

This pull-request adds signatures resolving with `NilClass` so we can get the `T::Range[T.untyped]` as expected.

Regarding the case `Range(nil, nil)`, the [original idea](https://github.com/sorbet/sorbet/pull/3260#discussion_r451028410) was to use `T.noreturn` as to raise errors on calls to `first`, `last` and `each` because since Ruby 2.7:

```
irb(main):001:0> r = Range.new(nil, nil)
=> nil..nil
irb(main):002:0> r.first
=> RangeError (cannot get the first element of beginless range)
irb(main):003:0> r.last
=> RangeError (cannot get the last element of endless range)
irb(main):004:0> r.each
=> #<Enumerator: nil..nil:each>
```

But it turns out this might no be compatible with previous versions of Ruby:

```
irb(main):001:0> r = Range.new(nil, nil)
=> nil..
irb(main):002:0> r.first
=> nil
irb(main):003:0> r.last
=> RangeError (cannot get the last element of endless range)
irb(main):004:0> r.each
=> #<Enumerator: nil..:each>
```

Only `last` raises an error with 2.6 and `each` return an enumerator over `nil` in both version so I don't think it makes sense to type it as `T::Range[T.noreturn]`. The changes in this pull-request makes it a `T::Range[T.noreturn]`.

### Test plan

I added a bunch of examples in the `range.rb` test.

PR #4178, opened 2021-04-26
Fix parser shift/reduce conflicts

Morriar:
### Motivation

I introduced shift/reduce conflicts in the grammar while adding support for begin-less ranges with `..` and `...` (https://github.com/sorbet/sorbet/pull/3313). This PR fixes them.

Example when compiling the parser before this PR:

image

The approach is the same as the one used by Whitequark (https://github.com/whitequark/parser/blob/de619a2ff36a39b4d43efa9b86cdec98621ed435/lib/parser/lexer.rl#L2026-L2035): duplicating the tokens `tDOT2` and `tDOT3` and lexing them differently depending on the Ruby version used.

One quirk with this approach is the way we display the parsing errors. Without any additional change, the error message would be:

```
^^^ error: unexpected token tBDOT3
```

In order to get a friendlier output, I tweaker the error generation method (see second commit) so we get back to the expected following error:

```
^^^ error: unexpected token "..."
```

Compiling the parser after this PR:

image

No more shift/reduce conflict warning. 🎉 

### Test plan

See included automated tests.

### DIFF
--- a/third_party/parser/cc/grammars/typedruby.ypp
+++ b/third_party/parser/cc/grammars/typedruby.ypp
@@ -179,6 +179,8 @@ using namespace std::string_literals;
   tANDDOT             1142    "&."
   tRATIONAL_IMAGINARY 1143
   tFLOAT_IMAGINARY    1144
+  tBDOT2              1145    // Used to avoid shift/reduce conflicts with productions using `tDOT2`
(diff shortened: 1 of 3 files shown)


jez (approved):
Thanks!

PR #4179, opened 2021-04-27
add some sorbet-runtime serialization tests

froydnj:
### Motivation

We weren't testing error cases very well, and we weren't testing set props at all.

(Some of the error cases for the sorbet-runtime generated serialization/deserialization code can theoretically be lumped together, but those cases have distinct handling in some changes we have.)

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/test/types/props/serializable.rb
+++ b/gems/sorbet-runtime/test/types/props/serializable.rb
@@ -688,12 +688,10 @@ class Opus::Types::Test::Props::SerializableTest < Critic::Unit::UnitTest
       struct.set = set
 
       h = struct.serialize
-      assert_equal(1, h['set'].length)
-      assert(h['set'].include?('foo'))
+      assert_equal(set, h['set'])
 
       roundtripped = SetPropStruct.from_hash(h)
-      assert_equal(1, roundtripped.set.length)
-      assert(roundtripped.set.include?('foo'))
+      assert_equal(roundtripped.set, set)
     end
 
     it 'does not share structure on serialize' do
--- a/gems/sorbet-runtime/test/types/props/serializable.rb
+++ b/gems/sorbet-runtime/test/types/props/serializable.rb
@@ -494,6 +494,11 @@ class Opus::Types::Test::Props::SerializableTest < Critic::Unit::UnitTest
     prop :hash_both, T::Hash[CustomType, CustomType], default: {}
   end
 
+  class CustomTypeWrapper
+    include T::Props::Serializable
+    prop :struct, CustomTypeStruct
+  end
+
   describe 'custom type' do
     it 'round trips as plain value' do
       assert_equal('foo', CustomTypeStruct.from_hash({'single' => 'foo'}).serialize['single'])
@@ -510,6 +515,25 @@ class Opus::Types::Test::Props::SerializableTest < Critic::Unit::UnitTest
     end
 
     # It's hard to write tests for a CustomType with a custom deserialize method, so skip that.
+
+    it 'raises serialize errors for members with a serializable subtype' do
+      struct = CustomTypeWrapper.new
+      struct.instance_variable_set(:@struct, 'not a serializable thing')
+      e = assert_raises(NoMethodError) do
+        struct.serialize
+      end
+
+      assert_includes(e.message, "undefined method `serialize'")
+    end
+
+    it 'raises deserialize errors for members with a serializable subtype' do
+      obj = 'not a serializable thing'
+      e = assert_raises(TypeError) do
+        CustomTypeWrapper.from_hash({'struct' => obj})
+      end
+
+      assert_includes(e.message, 'provided to from_hash')
+    end
 
     it 'round trips as array value' do
       assert_equal(['foo'], CustomTypeStruct.from_hash({'array' => ['foo']}).serialize['array'])
--- a/gems/sorbet-runtime/test/types/props/serializable.rb
+++ b/gems/sorbet-runtime/test/types/props/serializable.rb
@@ -640,6 +640,95 @@ class Opus::Types::Test::Props::SerializableTest < Critic::Unit::UnitTest
     end
   end
 
+  class SetPropStruct
+    include T::Props::Serializable
+    prop :set, T::Set[String]
+  end
+
+  describe 'set props' do
+    it 'round trips' do
+      set = Set['foo']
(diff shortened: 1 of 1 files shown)


gems/sorbet-runtime/test/types/props/serializable.rb
```diff
@@ -494,26 +494,275 @@ class CustomTypeStruct < T::Struct
...
+    include T::Props::Serializable
+    prop :struct, CustomTypeStruct
+  end
+
   describe 'custom type' do
     it 'round trips as plain value' do
       assert_equal('foo', CustomTypeStruct.from_hash({'single' => 'foo'}).serialize['single'])
     end
 
+    it 'raises serialize errors for members with a custom subtype' do
```
aprocter-stripe, line 507: The body of this `it` makes sense, but I'm not sure I understand the description. What does `members with a custom subtype` mean here? (Same general 

Oh, hmm, maybe this means something like `raises serialize errors when the types of members of a custom subtype are wrong`?

gems/sorbet-runtime/test/types/props/serializable.rb
```diff
@@ -494,26 +494,275 @@ class CustomTypeStruct < T::Struct
...
+      e = assert_raises(TypeError) do
+        struct.serialize
+      end
+
+      assert_includes(e.message, "Expected type T::Props::CustomType")
+    end
+
+    # It's hard to write tests for a CustomType with a custom deserialize method, so skip that.
+
+    it 'raises serialize errors for members with a serializable subtype' do
```
aprocter-stripe, line 519: Similar comment here (the code for the test makes sense, but I don't understand the description)

gems/sorbet-runtime/test/types/props/serializable.rb
```diff
@@ -494,26 +494,275 @@ class CustomTypeStruct < T::Struct
...
+      struct = CustomTypeWrapper.new
+      struct.instance_variable_set(:@struct, 'not a serializable thing')
+      e = assert_raises(NoMethodError) do
+        struct.serialize
+      end
+
+      assert_includes(e.message, "undefined method `serialize'")
+    end
+
+    it 'raises deserialize errors for members with a serializable subtype' do
```
aprocter-stripe, line 529: Similar comment here (the code for the test makes sense, but I don't understand the description)

gems/sorbet-runtime/test/types/props/serializable.rb
```diff
@@ -494,26 +494,275 @@ class CustomTypeStruct < T::Struct
...
+      end
+
+      assert_includes(e.message, 'provided to from_hash')
+    end
+
     it 'round trips as array value' do
       assert_equal(['foo'], CustomTypeStruct.from_hash({'array' => ['foo']}).serialize['array'])
     end
 
+    it 'raises serialize errors for members needing map' do
```
aprocter-stripe, line 542: Maybe something like `members that require map but don't have it`, idk. (Gonna stop harping on the test descriptions at this point, this is all hard to write in English. :))

gems/sorbet-runtime/test/types/props/serializable.rb
```diff
@@ -494,26 +494,275 @@ class CustomTypeStruct < T::Struct
...
+      T::Configuration.soft_assert_handler = proc do |msg, extra|
+        msg_string = msg
+        extra_hash = extra
+      end
+
+      obj = CustomType.new
+      result = CustomTypeStruct.from_hash({'array' => obj})
+      assert_equal(obj, result.array)
+
+      refute_nil(msg_string)
```
aprocter-stripe, line 565: Can we say anything more about `msg_string` here? (I'm assuming it would say something about the fact that `map` is missing, though I may not understand what it is.)

gems/sorbet-runtime/test/types/props/serializable.rb
```diff
@@ -494,26 +494,275 @@ class CustomTypeStruct < T::Struct
...
+      T::Configuration.soft_assert_handler = proc do |msg, extra|
+        msg_string = msg
+        extra_hash = extra
+      end
+
+      obj = 'not a hash'
+      result = CustomTypeStruct.from_hash({'hash_key' => obj})
+      assert_equal('not a hash', result.hash_key)
+
+      refute_nil(msg_string)
```
aprocter-stripe, line 600: Similar question (anything interesting about the contents of `msg_string`?) ...and in a few more places below where this pattern pops up.

gems/sorbet-runtime/test/types/props/serializable.rb
```diff
@@ -494,26 +494,275 @@ class CustomTypeStruct < T::Struct
...
+      set = Set['foo']
+      struct = SetPropStruct.new
+      struct.set = set
+
+      h = struct.serialize
+      assert_equal(1, h['set'].length)
+      assert(h['set'].include?('foo'))
+
+      roundtripped = SetPropStruct.from_hash(h)
+      assert_equal(1, roundtripped.set.length)
```
aprocter-stripe, line 695: Could we just `assert_equal(roundtripped.set, set)`?

aprocter-stripe (approved):
Looks good to me! Just a few clarifications/minor style things you may want to take a look at, but please feel free to merge with or without actioning them.

gems/sorbet-runtime/test/types/props/serializable.rb
froydnj: Maybe "raises serialization errors when dprops with a custom subtype store the wrong datatype"?

And so on for some of the (largely cut-and-pasted) errors below.

gems/sorbet-runtime/test/types/props/serializable.rb
froydnj: Ah, yeah, that'd be easier, wouldn't it...

gems/sorbet-runtime/test/types/props/serializable.rb
aprocter-stripe: Yeah, that sounds good!

gems/sorbet-runtime/test/types/props/serializable.rb
froydnj: We can't say anything super-useful about `msg_string` (it's always "Deserialization error (probably unexpected stored type)").  I suppose we could assert something about `storytime[:error]` a little further on.  I'll do that instead.

PR #4181, opened 2021-04-29
Really unchecked version of `T.let`

jez:
## Motivation

The `checked: false` annotation on a `T.let` is somewhat of a lie, because
the program still pays the cost of constructing the runtime type, even
though it's going to be discarded.

Inside hot code paths, constructing these runtime types is costly, both in
terms of allocations and time.

## Summary

T.let and friends now can take the type as a block:

```ruby
T.let(expr, checked: false) {Type}
```

In this form, the block will never run, so the program will not spend time
building the runtime type.

## Performance

Collected using

```
cd gems/sorbet-runtime
bundle exec rake bench:typecheck
```

| benchmark                                         | time        |
| ---------                                         | ---:        |
| `T.unsafe(...)`                                   | 23.941 ns   |
| `T.let(..., Integer)`                             | 463.187 ns  |
| `T.let(..., Integer, checked: false)`             | 91.872 ns   |
| `T.let(..., checked: false) {Integer}`            | 92.954 ns   |
| &nbsp;                                            | &nbsp;      |
| `T.let(..., T.nilable(Integer))`                  | 1126.000 ns |
| `T.let(..., T.nilable(Integer), checked: false)`  | 773.141 ns  |
| `T.let(..., checked: false) {T.nilable(Integer)}` | 92.947 ns   |
| &nbsp;                                            | &nbsp;      |
| `T.let(..., Example)`                             | 459.601 ns  |
| `T.let(..., Example, checked: false)`             | 90.671 ns   |
| `T.let(..., checked: false) {Example}`            | 93.025 ns   |
| &nbsp;                                            | &nbsp;      |
| `T.let(..., T.nilable(Example))`                  | 1125.000 ns |
| `T.let(..., T.nilable(Example), checked: false)`  | 768.239 ns  |
| `T.let(..., checked: false) {T.nilable(Example)}` | 94.174 ns   |

All the block forms:

- are about ~90 ns (regardless of type complexity)
- are never more than ~1 ns slower than non-block form (likely noise)
- take about 4x the time spent to run T.unsafe

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1746,6 +1746,28 @@ class ResolveTypeMembersAndFieldsWalk {
         }
     }
 
+    ast::ExpressionPtr sendToCast(core::Context ctx, core::LocOffsets castLoc, core::NameRef castName,
+                                  ast::ExpressionPtr expr, ast::ExpressionPtr type) {
+        ResolveCastItem item;
+        item.file = ctx.file;
+
+        // Compute the containing class when translating the type, as there's a very good chance
+        // this has been called from a method context.
+        item.owner = ctx.owner.enclosingClass(ctx);
+
+        auto typeExpr = ast::MK::KeepForTypechecking(std::move(type));
+        auto cast = ast::make_expression<ast::Cast>(castLoc, core::Types::todo(), std::move(expr), castName);
+        item.cast = ast::cast_tree<ast::Cast>(cast);
+        item.typeArg = &ast::cast_tree_nonnull<ast::Send>(typeExpr).args[0];
+
+        // We should be able to resolve simple casts immediately.
+        if (!tryResolveSimpleClassCastItem(ctx.withOwner(item.owner), item)) {
+            this->todoResolveCastItems_.emplace_back(move(item));
+        }
+
+        return ast::MK::InsSeq1(castLoc, move(typeExpr), move(cast));
+    }
+
     void validateNonForcingIsA(core::Context ctx, const ast::Send &send) {
         constexpr string_view method = "T::NonForcingConstants.non_forcing_is_a?";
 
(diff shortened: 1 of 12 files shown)


elliottt (changes requested):
This is great! My only question is about the choice to only allow the block form with `checked: false`. I think that the ideal situation would be to raise the error in the runtime as well, or handle the type being provided through a block.

jez:
> This is great! My only question is about the choice to only allow the block form with checked: false. I think that the ideal situation would be to raise the error in the runtime as well, or handle the type being provided through a block.

Good idea. Only looking at `checked` makes the fast path slightly faster, makes the slow path slower, but it's the slow path anyways, and it's dominated by type checking time anyways.

| benchmark                                         | time        |
| ---------                                         | ---:        |
| `T.let(..., Integer)`                             | 505.243 ns  |
| `T.let(..., Integer, checked: false)`             | 76.393 ns   |
| `T.let(..., checked: false) {Integer}`            | 81.275 ns   |
| &nbsp;                                            | &nbsp;      |
| `T.let(..., T.nilable(Integer))`                  | 1178.000 μs |
| `T.let(..., T.nilable(Integer), checked: false)`  | 752.29 ns   |
| `T.let(..., checked: false) {T.nilable(Integer)}` | 81.111 ns   |
| &nbsp;                                            | &nbsp;      |
| `T.let(..., Example)`                             | 501.744 ns  |
| `T.let(..., Example, checked: false)`             | 74.293 ns   |
| `T.let(..., checked: false) {Example}`            | 80.212 ns   |
| &nbsp;                                            | &nbsp;      |
| `T.let(..., T.nilable(Example))`                  | 1171.000 μs |
| `T.let(..., T.nilable(Example), checked: false)`  | 745.589 ns  |
| `T.let(..., checked: false) {T.nilable(Example)}` | 81.034 ns   |

test/testdata/resolver/t_let_block.rb
```diff
@@ -0,0 +1,58 @@
...
+
+T.let(expr, checked: false, garbage: true) {MyType} # error: Unrecognized keyword argument `garbage`
+
+T.let(expr) {MyType} # error: Using `T.let` in block form requires `checked: false`
+
+T.let(expr, garbage: true) {MyType} # error: Unrecognized keyword argument `garbage`
+T.let(expr, checked: nil) {MyType} # error: Expected `T::Boolean` but found `NilClass`
+T.let(expr, checked: 1) {MyType} # error: Expected `T::Boolean` but found `Integer(1)`
+
+T.let(expr, checked: true) {MyType} # error: Using `T.let` in block form requires `checked: false`
```
froydnj, line 32: Since we require `checked: false` if we use blocks, should there be a test for non-constant `checked:` + block?  Actually, should we require a constant `checked:` argument always, block or no block?

jez: What do you mean “constant checked argument”

jez: Oh nvm re-reading I see what you mean.

we didn’t do that before, and I didn’t want to make a breaking change if we didn’t need to. Good call about writing tests for non-constant checked with a block

resolver/resolver.cc
```diff
@@ -2052,31 +2074,78 @@ class ResolveTypeMembersAndFieldsWalk {
...
-                    item.cast = ast::cast_tree<ast::Cast>(cast);
-                    item.typeArg = &ast::cast_tree_nonnull<ast::Send>(typeExpr).args[0];
+                        if (numKwargs < 2) {
+                            if (auto e = ctx.beginError(send.loc, core::errors::Resolver::InvalidTypeDeclaration)) {
+                                e.setHeader("Using `{}` in block form requires `{}`",
+                                            fmt::format("T.{}", send.fun.show(ctx)), "checked: false");
+                                auto endPos = send.loc.endPos();
+                                if (send.loc.exists() && endPos > 0) {
+                                    auto insertLoc = core::Loc(ctx.file, endPos - 1, endPos - 1);
+                                    e.replaceWith("Insert `checked: false`", insertLoc, ", checked: false");
```
jez, line 2102: write tests for these autocorrects

test/testdata/resolver/t_let_block.rb
jez: I have added a test for non-literal `checked: ...` and required that the block for must have `checked: false` syntactically  (not `checked: expr` where `expr = false`)

elliottt (approved):

jeffcarbs:
If this PR does get revived, thoughts on making the block form always usable? If checked is truthy then you’d just blk.call and perform the runtime check.

This way you could make “checked:” conditional and run it in development and tests but not prod eg:
```ruby
T.let(nil, checked: ENV[]) { T.nilable(Integer) }
```

More likely in our case we’d apply a patch in our runtime to make T.let/cast/etc behave like T.unsafe in prod and not explicitly set checked. In that case I’d want to be able use the block format.

Not a huge deal either way, would love to see this land either way.

PR #4185, opened 2021-04-29
Add an intrinsic for hash construction

elliottt:
Our previous desugaring for hash literals that interspersed keyword arguments and double-splats produced chained calls to `.merge`, which produces a lot of intermediates. This PR replaces that desugaring with a single intrinsic, `<Magic>.<merge-hash>` which accepts 2 or more arguments, and will merge them all together preserving Shape types when possible. As this desugaring is applied when a double-splat is present in a send, this means that we'll better mirror the behavior of the vm for this case.

Two intrinsic that mirror the behavior of `to_hash` were also added: `<Magic>.<to-hash-dup>` and `<Magic>.<to-hash-nodup>`. The difference in behavior with `.to_hash` is that the `<to-hash-dup>` variant is used to signal that the argument will need to be duplicated after calling `.to_hash`, while `<to-hash-nodup>` indicates that only calling `to_hash` is necessary. (`.to_hash` will return its receiver when given a hash, so it is not safe to assume that its result can be mutated.)

Before:
```
foo(a: 10, **args, b: 20)
bar(**args, x: 10)
===
foo({a: 10}.merge(args.to_hash).merge({b: 20}))
bar(args.to_hash.merge({x: 10}))
```

After
```
foo(a: 10, **args, b: 20)
bar(**args, x: 10)
===
  foo(begin
      <hashTemp>$2 = {:a => 10}
      <hashTemp>$2 = ::<Magic>.<merge-hash>(<hashTemp>$2, ::<Magic>.<to-hash-nodup>(<self>.args()))
      <hashTemp>$2 = ::<Magic>.<merge-hash-values>(<hashTemp>$2, :b, 20)
      <hashTemp>$2
    end)

  bar(begin
      <hashTemp>$3 = ::<Magic>.<to-hash-dup>(<self>.args())
      <hashTemp>$3 = ::<Magic>.<merge-hash-values>(<hashTemp>$3, :x, 10)
      <hashTemp>$3
    end)
```

### Motivation

Improving keyword argument handling.

### Test plan

See included automated tests.


core/GlobalState.cc
```diff
@@ -580,6 +580,21 @@ void GlobalState::initEmpty() {
...
 
+    // Synthesize <Magic>.<to-hash-dup>(arg: T.untyped) => T.untyped
+    method = enterMethod(*this, Symbols::MagicSingleton(), Names::toHashDup())
+                 .untypedArg(Names::arg0())
+                 .buildWithResultUntyped();
+
+    // Synthesize <Magic>.<to-hash-nodup>(arg: T.untyped) => T.untyped
+    method = enterMethod(*this, Symbols::MagicSingleton(), Names::toHashNoDup())
+                 .untypedArg(Names::arg0())
+                 .buildWithResultUntyped();
```
elliottt, line 591: I'm having second thoughts about this being necessary: we could just fall back on `x.to_hash` instead.

ast/desugar/Desugar.cc
```diff
@@ -1534,6 +1544,13 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
                 values.reserve(hash->pairs.size()); // overapproximation in case there are KwSpats
                 ExpressionPtr lastMerge;
 
+                Send::ARGS_store mergeEntries;
+                mergeEntries.reserve(hash->pairs.size());
+
+                // Desguar
+                //   {**x, a: 'a', **y, remaining}
+                // into
+                //   <Magic>.<merge-hash>(<Magic>.<to-hash-dup>(y), {a: 'a'}, <Magic>.<to-hash-nodup>(x), remaining)
```
froydnj, line 1553: It looks like the VM will turn this into something more like:

```
t = <Magic>.merge-into({}, x)
t = <Magic>.add-entries(t, :a, 'a')
t = <Magic>.merge-into(t, y)
# not entirely sure how `remaining` plays in here, so ignoring it
```

That seems slightly better, because we're not explicitly constructing a hash for `{a: 'a'}` and we don't need to worry about no-dup vs. dup because merge-into + the initial fresh hash takes care of that for us?

froydnj (changes requested):
I'm not sure that we want primitives that require building hashes to insert single entries.

ast/desugar/Desugar.cc
trevor-stripe: Yep that's a good point. As long as we're threading new variables through as well, this should behave just like our existing merge intrinsics, so I think that I'll rework it to look like this instead.

elliottt: I've implemented this new strategy, and updated the description to indicate what the new desugaring looks like.

core/types/calls.cc
```diff
@@ -2615,6 +2615,110 @@ class Shape_to_hash : public IntrinsicMethod {
...
+        DispatchArgs dispatch{core::Names::toHash(),
+                              locs,
+                              0,
+                              innerArgs,
+                              arg->type,
+                              {arg->type, args.fullType.origins},
+                              arg->type,
+                              nullptr,
+                              args.originForUninitialized};
+        res = arg->type.dispatchCall(gs, dispatch);
```
froydnj, line 2638: Do you know if we would have tests that ensure something like `foo({**x, a: 1})` (i.e. the types derived by this function and others) actually do what we think they do?  I guess that example might run afoul of Ruby's hash-args-are-kwargs behavior, but maybe there are other ways we could test this?  (Straight `T.reveal_type` tests?)

froydnj (approved):
Excited to see this land!  This all looks good, I just have a couple of comments.

core/types/calls.cc
```diff
@@ -2615,6 +2615,110 @@ class Shape_to_hash : public IntrinsicMethod {
...
+    }
+
+} Magic_toHash;
+
+// Interpret this `<Magic>.<merge-hash>(a, b)` as `a.merge(b)`
```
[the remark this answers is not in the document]
elliottt: That's a good call, it has the semantics of `merge!` in that it updates the receiver. I'll update the comment.

core/types/calls.cc
```diff
@@ -2615,6 +2615,110 @@ class Shape_to_hash : public IntrinsicMethod {
...
+                               {accType, args.args[0]->origins},
+                               accType,
+                               nullptr,
+                               args.originForUninitialized};
+
+        res = accType.dispatchCall(gs, mergeArgs);
+    }
+} Magic_mergeHash;
+
+// Interpret this `<Magic>.<merge-hash-values>(a, k, v, ...)` as `a.merge({k => v, ...})`
```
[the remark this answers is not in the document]
elliottt: Ditto

ast/desugar/Desugar.cc
```diff
@@ -1528,72 +1533,123 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
...
+                        hashKeyDupes.check(key);
+                        mergeValues.emplace_back(std::move(key));
+
                         auto value = node2TreeImpl(dctx, std::move(pair->value));
-                        keys.emplace_back(std::move(key));
-                        values.emplace_back(std::move(value));
+                        mergeValues.emplace_back(std::move(value));
+
+                        hashLitPresent = true;
                     } else {
```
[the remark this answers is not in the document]
elliottt: Yep, we could definitely do that.

ast/desugar/Desugar.cc
```diff
@@ -1528,72 +1533,123 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
...
+                updateStmts.reserve(hash->pairs.size());
+
+                auto acc = dctx.ctx.state.freshNameUnique(core::UniqueNameKind::Desugar, core::Names::hashTemp(),
+                                                          ++dctx.uniqueCounter);
+
+                DuplicateHashKeyCheck hashKeyDupes(dctx);
+                Send::ARGS_store mergeValues;
+                mergeValues.reserve(hash->pairs.size() * 2 + 1);
+                mergeValues.emplace_back(MK::Local(loc, acc));
+                bool hashLitPresent = false;
```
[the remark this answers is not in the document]
elliottt: Yep, that name makes sense too 👍

core/types/calls.cc
elliottt: The `test/testdata/desugar/hash.rb` test covers a lot of cases like this. I can look at adding any cases when reworking it to better reflect the new desugaring.

core/types/calls.cc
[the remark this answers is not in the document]
elliottt: I think that in terms of changing it to dispatch on `merge!` instead, I'm inclined to leave it the way it is. We only generate calls to this intrinsic that look like `x = <Magic>.<merge-hash>(x, y)`, so the buggy details of `merge!` not updating the receiver type of a ShapeType won't show up here.

core/types/calls.cc
elliottt: Actually this case looks like it's already present in https://github.com/sorbet/sorbet/blob/35c488b6/test/testdata/infer/kwargs_splat.rb#L25

froydnj: That makes sense, it seemed like there ought to be some cases for kwsplat typing somewhere. :)

PR #4188, opened 2021-04-30
Upgrade abseil in advance of clang-12

elliottt:
Upgrade abseil to enable building with clang-12.

### Motivation

Clang-12 unblocks builds on big sur.

### Test plan

See included automated tests.


elliottt:
It looks like this will need to be done at the same time as the clang-12 upgrade.

PR #4189, opened 2021-04-30
Upgrade to clang 12.0.0

elliottt:
Upgrade us to clang-12.0.0 🎉 

### Motivation

Building sorbet on Big Sur is currently broken, and upgrading to clang-12 fixes the problems.

### Test plan

* [x] Manually run a mac build in CI
* [x] Run on Stripe CI
* [x] Disabled mac branch builds
* [x] Upload the llvm binaries to our s3 bucket
* [x] Make sure that sorbet.run still works

### DIFF
--- a/third_party/externals.bzl
+++ b/third_party/externals.bzl
@@ -166,9 +166,9 @@ package(default_visibility = ["//visibility:public"])
     # NOTE: we use the sorbet branch for development to keep our changes rebasable on grailio/bazel-toolchain
     http_archive(
         name = "com_grail_bazel_toolchain",
-        urls = _github_public_urls("sorbet/bazel-toolchain/archive/0c3130d83ed918d7a51490ea4dd814cb2cc68a85.zip"),
-        sha256 = "4d23f1e8b6ec0e8f612d47ae6189eb00e01461c8ad37cc6c0a41c0e9d5727084",
-        strip_prefix = "bazel-toolchain-0c3130d83ed918d7a51490ea4dd814cb2cc68a85",
+        urls = _github_public_urls("sorbet/bazel-toolchain/archive/a685e1e6bd1e7cc9a5b84f832539585bb68d8ab4.zip"),
+        sha256 = "90c59f14cada755706a38bdd0f5ad8f0402cbf766387929cfbee9c3f1b4c82d7",
+        strip_prefix = "bazel-toolchain-a685e1e6bd1e7cc9a5b84f832539585bb68d8ab4",
     )
 
     http_archive(
--- a/third_party/externals.bzl
+++ b/third_party/externals.bzl
@@ -146,9 +146,9 @@ def register_sorbet_dependencies():
 
     http_archive(
         name = "com_google_absl",
-        urls = _github_public_urls("abseil/abseil-cpp/archive/62f05b1f57ad660e9c09e02ce7d591dcc4d0ca08.zip"),
-        sha256 = "afcab9f226ac4ca6b6b7c9ec704a995fe32a6b555d6935b0de247ae6ac6940e0",
-        strip_prefix = "abseil-cpp-62f05b1f57ad660e9c09e02ce7d591dcc4d0ca08",
+        urls = _github_public_urls("abseil/abseil-cpp/archive/e9b9e38f67a008d66133535a72ada843bd66013f.zip"),
+        sha256 = "49c93740b3b09f73cd2f10da778ea4129d59733085393f458a4acd17774503fb",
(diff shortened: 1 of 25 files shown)


froydnj (approved):
r=me assuming that all the non-CI-related tests pass.

Can confirm that building on Ubuntu 16.04 and 9 still works.

elliottt:
@froydnj there have been some changes since you last looked at this PR. I rebased the branch so that the only commits that really need to land together are the last two, f7aa8c3d58cc74dbb02968f39f67c0e47354ba29 and e07596009ff6a5e630584e3d9e5d8fa39853a021, the others could be landed separately if we'd prefer to go that direction. I think that it will probably be easiest to review this pr by commit.

third_party/ruby/ruby.BUILD
```diff
@@ -4,6 +4,12 @@ load("@com_stripe_ruby_typer//third_party/ruby:build-ruby.bzl", "ruby")
 
 ruby(
     bundler = "@bundler_2_1_4//file",
+    linkopts = select({
+        "@com_stripe_ruby_typer//tools/config:darwin": [
+            "-mlinker-version=400",
+        ],
```
froydnj, line 10: I do not have a problem with this being committed as-is, but I am curious whether you tried to make the build work with the linker (if there is one) in LLVM 12.

third_party/externals.bzl
```diff
@@ -146,9 +146,9 @@ def register_sorbet_dependencies():
 
     http_archive(
         name = "com_google_absl",
-        urls = _github_public_urls("abseil/abseil-cpp/archive/62f05b1f57ad660e9c09e02ce7d591dcc4d0ca08.zip"),
-        sha256 = "afcab9f226ac4ca6b6b7c9ec704a995fe32a6b555d6935b0de247ae6ac6940e0",
-        strip_prefix = "abseil-cpp-62f05b1f57ad660e9c09e02ce7d591dcc4d0ca08",
+        urls = _github_public_urls("abseil/abseil-cpp/archive/e9b9e38f67a008d66133535a72ada843bd66013f.zip"),
+        sha256 = "49c93740b3b09f73cd2f10da778ea4129d59733085393f458a4acd17774503fb",
+        strip_prefix = "abseil-cpp-e9b9e38f67a008d66133535a72ada843bd66013f",
```
froydnj, line 151: Do changes in the compiler make it necessary to upgrade abseil?  I'm not opposed to this, I would just like to have a rationale written down because I don't see one in the PR or the commit message for the particular commit that changed this.

third_party/jemalloc.BUILD
```diff
@@ -31,7 +31,23 @@ JEMALLOC_BUILD_COMMAND = """
...
+      ;;
+    Darwin)
+      LDFLAGS="$${LDFLAGS} -mlinker-version=400"
+      ;;
+  esac
+
+  export LDFLAGS
+
+  # we need to include CFLAGS in CPPFLAGS so that the --sysroot flag makes it in
+  export CPPFLAGS="$${CFLAGS}"
```
froydnj, line 49: So great that the jemalloc build system has `EXTRA_CXXFLAGS` above but `CPPFLAGS` here.

Actually, this script has `export CFLAGS=$(CC_FLAGS)`/`export CXXFLAGS=$(CC_FLAGS)` just a few lines above this, and we don't seem to be making additional changes to `CFLAGS` that would affect this...is this just necessary because the script should be using `export CXXFLAGS=$$(CC_FLAGS)`?  Or is it because the script should have been setting `CPPFLAGS` instead of `CXXFLAGS` in the first place?

froydnj (approved):
This is fine.  I would kind of like the jemalloc situation straightened out, but I can understand that testing that is a pain due to having to run it through Mac CI, and this PR is blocking people, so I would be OK deferring that until later.

third_party/ruby/ruby.BUILD
elliottt: The build works on big sur without this, but fails in mac ci with it. This was the easiest compromise for getting a build that worked for both.

third_party/externals.bzl
elliottt: Yes, the build will fail on clang-12 without upgrading abseil.

elliottt: Additionally, the build will fail on abseil head, so this commit was the compromise that got us a working build.

third_party/ruby/ruby.BUILD
elliottt: Oh! I haven't tried using the `lld` that's present in the clang download. I'd be up for trying that as a follow up task, though.

PR #4191, opened 2021-04-30
treat sig(:final) .checked(:never) sigs as not unchecked

froydnj:
### Motivation

The state of affairs decided on by the patch is not ideal, but it is the most expedient thing to do.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/AttrReader.cc
+++ b/rewriter/AttrReader.cc
@@ -149,6 +149,15 @@ bool sigIsUnchecked(core::MutableContext ctx, ast::Send *sig) {
         return true;
     }
 
+    // We want sig(:final) to overrule .checked(:never)
+    if (sig->args.size() > 0) {
+        int final_arg = sig->args.size() == 1 ? 0 : 1;
+        auto lit = ast::cast_tree<ast::Literal>(sig->args[final_arg]);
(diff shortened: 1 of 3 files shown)


jez:
Dumping some notes we brain stormed together here for posterity:

```ruby
keep_def() # says could be IVAR_METHOD_TYPE
ast::MethodDef / cfg::CFG # says could be IVAR_METHOD_TYPE

sig {returns(Integer)} # normal method like anything else
attr_reader :foo

sig {returns(Integer).checked(:never)} # IVAR_METHOD_TYPE
attr_reader :foo

sig(:final) {returns(Integer)} # normal method like anything else
attr_reader :foo

sig(:final) {returns(Integer).checked(:never)}
attr_reader :foo
# SORBET_FUNC + special implementation of IREmitter
# - ONLY the ivar get, and no other setup, teardown, or type checking
# - but, this is kind of tricky
# - for now, let's just make it be a normal method
# - Whenever possible, DON'T use `sig(:final)` + `.checked(:never)`
#   (unless, e.g., the class is `final!`
#   - But, we can write a benchmark to confirm our suspicion here
# - tl;dr: ignore the `.checked(:never)` entirely when final
```

froydnj:
We discussed this offline, and are going to approach this differently.

PR #4193, opened 2021-05-03
use abseil's contains method rather than std's find()/end() idiom

froydnj:
abseil's sets and maps helpfully provide a `contains()` method, so you don't have to `find()` and then compare to `end()`, as you would with `std::` associative containers.

### Motivation

More readable code.  I don't claim that this is complete, this is just what grep told me was easily doable.

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -566,9 +566,9 @@ public:
             return;
         }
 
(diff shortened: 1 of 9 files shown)


jez (approved):
🙌

froydnj:

→ https://go/builds/bui_JPviFqPxa9iZH3
→ https://go/builds/bui_JPviVByWtSGskU

PR #4194, opened 2021-05-03
Validate overrides for attr accessors and T::Struct props

### LINKED ISSUE #1694: Interface type checking does not work with T::Struct
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Amodule%20MyInterface%0A%20%20include%20Kernel%0A%20%20extend%20T%3A%3AHelpers%0A%20%20extend%20T%3A%3ASig%0A%20%20interface!%20%20%0A%0A%20%20sig%20%7Babstract.returns(String)%7D%0A%20%20def%20fun%3B%20end%0Aend%0A%0Aclass%20A%20<%20T%3A%3AStruct%0A%20%20include%20MyInterface%0A%0A%20%20prop%20%3Afun%2C%20T%3A%3ABoolean%20%23%20should%20not%20type%20check!%0Aend%0A)

```ruby
# typed: strict

module MyInterface
  include Kernel
  extend T::Helpers
  extend T::Sig
  interface!  

  sig {abstract.returns(String)}
  def fun; end
end

class A < T::Struct
  include MyInterface

  prop :fun, T::Boolean # should not type check!
end

```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

It should print

```
Return type T::Boolean does not match return type of abstract method MyInterface#fun
```

- - -

Right now, interface and override checking will ignore DSL-generated methods because otherwise they would produce a number of errors that would make them unusable in practice. (For example, if you can't explicitly write a `sig` on a DSL-generated method, then you also can't explicitly write an `implementation` in that sig, but we demand that `implementation` appear explicitly in the types of things which implement abstract methods!) That however also means we miss out on useful error messages, like the one desired above.

### LINKED ISSUE #4094: "attr_reader" which implements an abstract method but lacks "override" does not cause error
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Amodule%20Test%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20interface!%0A%0A%20%20sig%20%7B%20abstract.returns(String)%20%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aclass%20Implementation%0A%20%20extend%20T%3A%3ASig%0A%20%20include%20Test%0A%0A%20%20def%20initialize%0A%20%20%20%20%40foo%20%3D%20T.let('Foo'%2C%20String)%0A%20%20end%0A%0A%20%20sig%20%7B%20returns(String)%20%7D%0A%20%20attr_reader%20%3Afoo%0Aend%0A%0Asig%20%7Bvoid%7D%0Adef%20foo%0A%20%20Implementation.new.foo%0Aend%0A%0Afoo())

```ruby
# typed: true
extend T::Sig

module Test
  extend T::Sig
  extend T::Helpers

  interface!

  sig { abstract.returns(String) }
  def foo; end
end

class Implementation
  extend T::Sig
  include Test

  def initialize
    @foo = T.let('Foo', String)
  end

  sig { returns(String) }
  attr_reader :foo
end
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

I expected a error indicating that the reader sig was not declared with override. Similar to this:
```
editor.rb:23: Method Implementation#foo implements an abstract method Test#foo but is not declared with override. https://srb.help/5051
    23 |  def foo
          ^^^^^^^
    editor.rb:11: defined here
    11 |  def foo; end
          ^^^^^^^
Errors: 1
```

### LINKED ISSUE #4100: Interfaces implemented by attr accessors or T::Structs do no identify signature mismatches
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20A%3B%20end%0Aclass%20B%3B%20end%0A%0Amodule%20Interface%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20interface!%0A%0A%20%20sig%20%7B%20abstract.returns(A)%20%7D%0A%20%20def%20do_it%3B%20end%0Aend%0A%0Aclass%20ClassImpl%0A%20%20extend%20T%3A%3ASig%0A%20%20include%20Interface%0A%0A%20%20%23%20Error%3A%20Return%20type%20T.nilable(String)%20does%20not%20match%20return%20type%20of%20abstract%20method%20%0A%20%20%23%20Correct!%0A%20%20sig%20%7B%20override.returns(B)%20%7D%0A%20%20def%20do_it%0A%20%20%20%20B.new%0A%20%20end%0Aend%0A%0Aclass%20StructImpl%20%3C%20T%3A%3AStruct%0A%20%20extend%20T%3A%3ASig%0A%20%20include%20Interface%0A%0A%20%20%23%20Does%20not%20identify%20the%20error%0A%20%20%23%20Trying%20to%20add%20sig%20here%20says%20the%20sig%20is%20unused%0A%20%20prop%20%3Ado_it%2C%20B%0Aend%0A%0Aclass%20AttrImpl%0A%20%20extend%20T%3A%3ASig%0A%20%20include%20Interface%0A%0A%20%20sig%20%7B%20void%20%7D%0A%20%20def%20initialize%0A%20%20%20%20%40do_it%20%3D%20T.let(B.new%2C%20B)%0A%20%20end%0A%0A%20%20%23%20Does%20not%20identify%20the%20error%0A%20%20sig%20%7B%20override.returns(B)%20%7D%0A%20%20attr_reader%20%3Ado_it%0Aend%0A)

```ruby
# typed: strict

class A; end
class B; end

module Interface
  extend T::Sig
  extend T::Helpers

  interface!

  sig { abstract.returns(A) }
  def do_it; end
end

class ClassImpl
  extend T::Sig
  include Interface

  # Error: Return type T.nilable(String) does not match return type of abstract method 
  # Correct!
  sig { override.returns(B) }
  def do_it
    B.new
  end
end

class StructImpl < T::Struct
  extend T::Sig
  include Interface

  # Does not identify the error
  # Trying to add sig here says the sig is unused
  prop :do_it, B
end

class AttrImpl
  extend T::Sig
  include Interface

  sig { void }
  def initialize
    @do_it = T.let(B.new, B)
  end

  # Does not identify the error
  sig { override.returns(B) }
  attr_reader :do_it
end
```

#### Observed output

```
editor.rb:23: Return type B does not match return type of abstract method Interface#do_it https://srb.help/5035
    23 |  def do_it
          ^^^^^^^^^
    editor.rb:13: Super method defined here with return type A
    13 |  def do_it; end
          ^^^^^^^^^
Errors: 1
```

#### Expected behavior

The `attr_reader` and the `T::Struct` versions of the implementation should also throw the error `Return type B does not match return type of abstract method` since they are too implementing the interface.

vinistock:
Closes #4094
Closes #4100
Closes #1694

This PR proposes a few changes to type check `attr_accessors` and `props` that implement abstract methods. The implementation consists of doing the following:
1. Sorbet was not trying to check overrides for methods with `isRewriterSynthesized` true. I've removed those checks to let it type check overrides for any method. Please, let me know if this is okay or if there are other consequences I'm unaware about
2. The type signature for props are generated internally by Sorbet, using the arguments passed to `prop`. Therefore, I don't see a way of indicating that a `prop` is an override other than adding another option. This one a bit odd, since the option does nothing in runtime, but allows us to mark the prop as an override and correctly type check

Please, let me know if this is the right direction to fix this issue and if I am missing something. The example below displays the new `override` option (which again does nothing in the runtime).

```ruby
module Interface
  extend T::Sig
  extend T::Helpers

  sig { abstract.void }
  def foo; end
end

class A
  include Interface
  
  sig { returns(Integer) }
  attr_reader :foo # error Integer doesn't match void
end

class B < T::Struct
  include Interface
 
  prop :foo, String, override: true # error String doesn't match void
end
```

### Motivation

Sorbet should be able to type check overrides implemented in attributes or props.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -312,8 +312,15 @@ private:
     }
 
 public:
-    static ExpressionPtr Sig(core::LocOffsets loc, Send::ARGS_store args, ExpressionPtr ret) {
-        auto params = Params(loc, Self(loc), std::move(args));
+    static ExpressionPtr Sig(core::LocOffsets loc, Send::ARGS_store args, ExpressionPtr ret, bool isOverride = false) {
+        sorbet::ast::ExpressionPtr receiver;
+        if (isOverride) {
+            receiver = Send0(loc, Self(loc), core::Names::override_());
+        } else {
+            receiver = Self(loc);
+        }
+
+        auto params = Params(loc, std::move(receiver), std::move(args));
         auto returns = Send1(loc, std::move(params), core::Names::returns(), std::move(ret));
         auto sig = Send1(loc, Constant(loc, core::Symbols::Sorbet_Private_Static()), core::Names::sig(),
                          Constant(loc, core::Symbols::T_Sig_WithoutRuntime()));
@@ -334,8 +341,15 @@ public:
         return sig;
     }
 
-    static ExpressionPtr Sig0(core::LocOffsets loc, ExpressionPtr ret) {
-        auto returns = Send1(loc, Self(loc), core::Names::returns(), std::move(ret));
+    static ExpressionPtr Sig0(core::LocOffsets loc, ExpressionPtr ret, bool isOverride = false) {
+        sorbet::ast::ExpressionPtr receiver;
+        if (isOverride) {
+            receiver = Send0(loc, Self(loc), core::Names::override_());
+        } else {
+            receiver = Self(loc);
+        }
+
+        auto returns = Send1(loc, std::move(receiver), core::Names::returns(), std::move(ret));
         auto sig = Send1(loc, Constant(loc, core::Symbols::Sorbet_Private_Static()), core::Names::sig(),
                          Constant(loc, core::Symbols::T_Sig_WithoutRuntime()));
         auto sigSend = ast::cast_tree<ast::Send>(sig);
@@ -344,8 +358,9 @@ public:
         return sig;
     }
 
-    static ExpressionPtr Sig1(core::LocOffsets loc, ExpressionPtr key, ExpressionPtr value, ExpressionPtr ret) {
-        return Sig(loc, SendArgs(std::move(key), std::move(value)), std::move(ret));
+    static ExpressionPtr Sig1(core::LocOffsets loc, ExpressionPtr key, ExpressionPtr value, ExpressionPtr ret,
+                              bool isOverride = false) {
+        return Sig(loc, SendArgs(std::move(key), std::move(value)), std::move(ret), isOverride);
     }
 
     static ExpressionPtr T(core::LocOffsets loc) {
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -312,10 +312,15 @@ private:
     }
 
 public:
-    static ExpressionPtr Sig(core::LocOffsets loc, Send::ARGS_store args, ExpressionPtr ret, bool isOverride = false) {
+    static ExpressionPtr Sig(core::LocOffsets loc, Send::ARGS_store args, ExpressionPtr ret, bool isOverride = false,
+                             bool allowIncompatible = false) {
         sorbet::ast::ExpressionPtr receiver;
+
         if (isOverride) {
-            receiver = Send0(loc, Self(loc), core::Names::override_());
+            auto incompatible = allowIncompatible ? True(loc) : False(loc);
+            auto args = SendArgs(Symbol(loc, core::Names::allowIncompatible()), std::move(incompatible));
+
+            receiver = Send(loc, Self(loc), core::Names::override_(), 0, std::move(args));
         } else {
             receiver = Self(loc);
         }
@@ -341,10 +346,14 @@ public:
         return sig;
     }
 
-    static ExpressionPtr Sig0(core::LocOffsets loc, ExpressionPtr ret, bool isOverride = false) {
+    static ExpressionPtr Sig0(core::LocOffsets loc, ExpressionPtr ret, bool isOverride = false,
+                              bool allowIncompatible = false) {
         sorbet::ast::ExpressionPtr receiver;
         if (isOverride) {
-            receiver = Send0(loc, Self(loc), core::Names::override_());
+            auto incompatible = allowIncompatible ? True(loc) : False(loc);
+            auto args = SendArgs(Symbol(loc, core::Names::allowIncompatible()), std::move(incompatible));
+
+            receiver = Send(loc, Self(loc), core::Names::override_(), 0, std::move(args));
         } else {
             receiver = Self(loc);
         }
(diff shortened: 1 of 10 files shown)


jez:
> Please, let me know if this is the right direction to fix this issue and if I am missing something.

Heads up that this is likely to cause a ton of errors on Stripe's codebase. Please let us know as early as possible once you have something worth running over our codebase, because it could easily be the case that this introduces thousands of errors for us to fix.

vinistock:
@jez yes, every attribute or prop that implements abstracts will - at least - alert that the implementation is not marked as override (there could be more errors if there are type mismatches of course).

Do you believe the general idea of the implementation is correct and worth pursuing? If that's the case, I'll check if all tests are passing and test it on our code base as well, before asking for formal reviews.

jez:

→ https://go/builds/bui_JPwkOGGfPEn2wT
→ https://go/builds/bui_JPwk6qLe6doCXf

jez:
Thanks. It looks like this introduces ~200 errors on Stripe's codebase. That's not impossible to fix, but it will require some dedicated time to fix.

Another question raised by this PR: `sorbet-runtime` currently does not do the same variance checks for `override` that would be done for normal methods:

https://github.com/vinistock/sorbet/blob/26f2472b8eaf43a4ae325198006dd9156385d370/gems/sorbet-runtime/lib/types/props/decorator.rb#L48-L52

I wonder if we should:

- add the runtime variance checks here, like would be done for `sig`s.
- allow the prop DSL to accept an `allow_incompatible` flag, like is possible for `sig {override(allow_incompatible: true)...}` to skip the variance checks in the runtime / statically.

vinistock:
I agree that we should add the runtime checks and add an `allow_incompatible` option. But before I even take a stab at it, I'd like your thoughts on another issue.

This implementation breaks certain tests. For example, `def_delegator` will now trigger errors when overriding methods that exist somewhere in the hierarchy.
```ruby
class EnumerableUsage
  include Enumerable

  def_delegator :thing, :each # implements an abstract method but is not marked as override
end
```
This is happening because I removed the check for synthesized methods. I think that in order to fix this in a deterministic way, we'd need the rewriters to include a special flag indicating whether we need to check for overrides or not (e.g.: `skipOverrideChecks`), which can then be used in `validator.cc` to not throw errors.

What do you think?

vinistock:
The two latest commits implement:
1. a way of skipping override checks for synthesized methods, so that we don't validate overrides in places where it does not make sense, such as in delegators
2. the implementation of `allow_incompatible` for props in the static analyzer

I am only missing the runtime validations for `prop` overrides.

vinistock:
I've finished adding runtime checks for prop overrides. It was messier than I anticipated. Let me know if you have concerns and how we can address them. I did leave a few comments in the code to try to explain my rationale.

aprocter-stripe:
I'm going to go ahead and reassign this to @jez since it looks like you guys have already had some discussion and I don't think I have enough context to review this in a timely fashion.

jez:
Thanks for the update @vinistock 

I'm out today, and monday is a US holiday. I'll get to this some time next week.

PR #4198, opened 2021-05-04
Use git clone --single-branch for sorbet.run

aprocter-stripe:
### Motivation
Work around git dying by `SIGKILL` (perhaps OOM?) during the `git clone sorbet.run` step in CI by passing `--single-branch`, so that we only clone `master` (whose history has recently been cleaned up).

### Test plan
Existing CI, and the `SIGKILL` did not seem to pop up during branch builds.

### DIFF
--- a/.buildkite/publish.sh
+++ b/.buildkite/publish.sh
@@ -31,7 +31,7 @@ buildkite-agent artifact download "_out_/**/*" .
 echo "--- releasing sorbet.run"
 
 rm -rf sorbet.run
-git clone --single-branch git@github.com:sorbet/sorbet.run.git
+git clone git@github.com:sorbet/sorbet.run.git --single-branch --branch master
 tar -xvf ./_out_/webasm/sorbet-wasm.tar ./sorbet-wasm.wasm ./sorbet-wasm.js
 mv sorbet-wasm.wasm sorbet.run/docs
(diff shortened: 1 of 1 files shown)


.buildkite/publish.sh
```diff
@@ -31,7 +31,7 @@ buildkite-agent artifact download "_out_/**/*" .
 echo "--- releasing sorbet.run"
 
 rm -rf sorbet.run
-git clone git@github.com:sorbet/sorbet.run.git
+git clone --single-branch git@github.com:sorbet/sorbet.run.git
```
froydnj, line 34: I thought that we should use shallow clones for this, and

https://www.perforce.com/blog/vcs/git-beyond-basics-using-shallow-clones

was the first thing that popped up, which says that `--single-branch` is basically the `--depth` thing I was thinking of.  It suggests using `git clone $URL --branch $BRANCH --single-branch` instead.  Is the `master` implicit here and/or is it worth making explicit?

froydnj (approved):
r=me, your call on what to do with the suggestion below.

.buildkite/publish.sh
aprocter-stripe: The `master` is implicit here. I'll go ahead and explicify it.

jez (approved):

PR #4200, opened 2021-05-06
Remove super_method call from hidden-definition-finder.rb - Fixes #4199

### LINKED ISSUE #4199: srb rbi hidden-definitions leads to segmentation fault with aliased methods
#### Input

```ruby
srb rbi hidden-definitions
```

#### Observed output

```
.../sorbet-0.5.600/lib/hidden-definition-finder.rb:375: [BUG] Segmentation fault at 0x0000000000000010
ruby 2.7.3p183 [2021-04-05 revision 6847ee089d) [x86_64-darwin19]

-- Control frame information -----------------------------------------
c:0023 p:---- s:0211 e:000210 CFUNC   :super_method
c:0022 p:0155 s:0207 e:000206 BLOCK   ..../sorbet-0.5.6400/lib/hidden-definition-finder.rb:375  [FINISH]
.....
```

#### Expected behavior

not segfault, successfully generate hidden definitions
- - -

There is a (recently fixed) bug in Ruby (https://github.com/ruby/ruby/commit/d8a13e504992a45d52063f7c925408d7aad3595a and https://github.com/ruby/ruby/commit/c60aaed1856b2b6f90de0992c34771830019e021) where calling `.super_method` on an aliased class method will lead to a segmentation fault. (the first fix is for instance methods, the second is for singleton methods I believe).

i.e. the following code will crash:
```ruby
m = Module.new do
  def foo
  end
  alias :f :foo 
end
method = m.instance_method(:f)
method.super_method
```

In https://github.com/sorbet/sorbet/blob/master/gems/sorbet/lib/hidden-definition-finder.rb#L375, there's the following block where `.super_method` is called

```ruby
     ...
      begin
        if is_singleton
          method = klass.singleton_method(name)
        else
          method = klass.instance_method(name)
        end
      rescue => e
        ret << "# #{e.message.gsub("\n", "\n# ")}"
        next
      end

      super_method = method.super_method     # <--------- 
      # next if super_method && T::AbstractUtils.abstract_method?(method) == T::AbstractUtils.abstract_method?(super_method)
     ...
```

On Ruby 2.7.3, this is causing a segmentation fault when trying to generate hidden definitions for any classes with aliased methods. 

I dropped a `puts method` in `hidden-definition-finder.rb:374` to see what method was being loaded before crashing:

```
#<UnboundMethod: ActionText::Serialization#_dump(*) ..../actiontext-6.0.3.5/lib/action_text/serialization.rb:30>
#<UnboundMethod: ActionText::Serialization#_load(load)(content) ..../actiontext-6.0.3.5/lib/action_text/serialization.rb:8>
.../sorbet-0.5.600/lib/hidden-definition-finder.rb:375: [BUG] Segmentation fault at 0x0000000000000010
```

In my case it was crashing with `actiontext` due to the `_load` method, which is an aliased method defined here: https://github.com/rails/rails/blob/main/actiontext/lib/action_text/serialization.rb#L27

Since it looks like `super_method` is being defined but not actually being used anywhere, I commented it out on a local fork and I'm no longer getting any seg faults.

Line `376` is the only place where `super_method` is used and it's commented it out, does it make sense to just delete lines `375-376`? That way this won't be an issue for older versions of Ruby.

alihussain5:
Deleting the following lines in https://github.com/sorbet/sorbet/blob/master/gems/sorbet/lib/hidden-definition-finder.rb#L375:

```ruby
super_method = method.super_method
# next if super_method && T::AbstractUtils.abstract_method?(method) == T::AbstractUtils.abstract_method?(super_method)
```

Because they're not being used. Also wrote some tests to ensure aliased methods work correctly in `hidden-definition-finder`.

### Motivation

See https://github.com/sorbet/sorbet/issues/4199 - existing bug in Ruby causes a crash if calling `.super_method` on an aliased method.

Since `.super_method` is defined but not being used, I figured its safe to just get rid of it.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet/lib/hidden-definition-finder.rb
+++ b/gems/sorbet/lib/hidden-definition-finder.rb
@@ -372,9 +372,6 @@ class Sorbet::Private::HiddenMethodFinder
         next
       end
 
-      super_method = method.super_method
-      # next if super_method && T::AbstractUtils.abstract_method?(method) == T::AbstractUtils.abstract_method?(super_method)
-
       errors = capture_stderr do
         ret << maker.serialize_method(method, is_singleton, with_sig: false)
       end
--- a/gems/sorbet/test/hidden-method-finder/hidden-method-tests.rb
+++ b/gems/sorbet/test/hidden-method-finder/hidden-method-tests.rb
@@ -20,8 +20,8 @@ TEST_CASES = [
 
 class Sorbet::Private::HiddenMethodFinder::Test::Simple < MiniTest::Spec
   TEST_CASES.each do |path|
-
     it "works on the #{path} example" do
+
       Dir.mktmpdir do |dir|
         FileUtils.cp_r(File.join(__dir__, path), dir)
         olddir = __dir__
--- a/gems/sorbet/test/hidden-method-finder/hidden-method-tests.rb
+++ b/gems/sorbet/test/hidden-method-finder/hidden-method-tests.rb
@@ -20,8 +20,8 @@ TEST_CASES = [
 
 class Sorbet::Private::HiddenMethodFinder::Test::Simple < MiniTest::Spec
   TEST_CASES.each do |path|
-    it "works on the #{path} example" do
 
+    it "works on the #{path} example" do
       Dir.mktmpdir do |dir|
         FileUtils.cp_r(File.join(__dir__, path), dir)
         olddir = __dir__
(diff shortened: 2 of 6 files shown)


jez (approved):
It looks like this has been commented out since the day it was added.

Thanks

aprocter-stripe:

→ https://go/build_groups/bg_JR4zOBqndPqvYS
https://go/builds/bui_JR4zmP88KOpKql
→ https://go/build_groups/bg_JR4zxVhxHAI58c
https://go/builds/bui_JR4zaMmq7wOF1Q

alihussain5:
Oops, looks like I messed up the thorough/expectations.json

aprocter-stripe (approved):
My only question was going to be whether @jez had any idea what was up with the commented-out check here in the first place, so LGTM :)

PR #4202, opened 2021-05-10
T::Utils.coerce: Push special-case module->T::Types::Base logic behind cache

djudd-stripe:
First commit: When coercing a module to a type, we do a bunch of checks for specific special classes like Array before falling back to the generic `T::Types::Simple.new`. But these are edge cases that don't occur often, plus the same caching mechanism we use for Simple can work here, so we can speed up `T.let` and friends in the common case by pushing these checks into `type_for_module` and caching the result.

Second commit: Lower impact and only loosely related, but since `T::Utils.coerce` already resolves aliases there's no need to invoke `resolve_alias` separately when initializing a `T::Types::Union`.

### Motivation
Before (relevant excerpt):
```
 sorbet-runtime git:(master) ✗ bundle exec rake bench:typecheck
...
T.let(..., Integer): 431.782 ns
T.let(..., T.nilable(Integer)): 1.129 μs
T.let(..., Example): 425.731 ns
T.let(..., T.nilable(Example)): 1.119 μs
```

After:
```
sorbet-runtime git:(djudd/coerce-opts) ✗ bundle exec rake bench:typecheck
...
T.let(..., Integer): 304.652 ns
T.let(..., T.nilable(Integer)): 1.001 μs
T.let(..., Example): 310.781 ns
T.let(..., T.nilable(Example)): 1.028 μs
```

### Test plan
Existing tests

### DIFF
--- a/gems/sorbet-runtime/lib/types/types/simple.rb
+++ b/gems/sorbet-runtime/lib/types/types/simple.rb
@@ -41,11 +41,26 @@ module T::Types
     module Private
       module Pool
         def self.type_for_module(mod)
-          cached = mod.instance_variable_get(:@__as_sorbet_simple_type)
+          cached = mod.instance_variable_get(:@__as_sorbet_type)
           return cached if cached
 
-          type = Simple.new(mod)
-          mod.instance_variable_set(:@__as_sorbet_simple_type, type) unless mod.frozen?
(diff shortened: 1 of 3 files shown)


elliottt (approved):
Looks great, thanks for the writeup!

PR #4205, opened 2021-05-11
Add support for def_delegators (plural)

richardxia:
This PR builds on top of the constructs added in https://github.com/sorbet/sorbet/pull/3636 for handling `#def_delegator` by modifying the rewriter to also support [`#def_delegators`](https://ruby-doc.org/stdlib-2.5.1/libdoc/forwardable/rdoc/Forwardable.html#method-i-def_delegators), which is a variadic version of `#def_delegator`. My solution was to emit a synthetic delegator method for each of the methods passed to `#def_delegators`.

I split up the `DefDelegator::run()` into a few separate functions so that I could reuse the pieces that should be the same between `#def_delegator` and `#def_delegators`, but without trying to be overly DRY, since there are a few differences between the two. The biggest difference is that `#def_delegators` does not support the optional third argument of `#def_delegator`, which specifies an alias name for the delegator method. Another difference it is legal to call `#def_delegators` without specifying _any_ method names (e.g. `def_delegators :@foo`), and it just results in a no-op.

Because of these differences, I just had separate functions for the "argument parsing" of the two different methods, and I only shared the code for generating method stubs.

I'm not that experienced with C++, so apologies in advance if my code is not idiomatic, but I definitely welcome feedback on how to write this better. I especially don't really have a good feel for how ownership, references, and copies should should be done here.

Also, I apologize for not exactly following the instructions in CONTRIBUTING.md. If you want me to file a GitHub issue to discuss this more first, I can do that retroactively.

### Motivation

Sorbet already had an existing AST rewriter for `#def_delegator`, but not [`#def_delegators`](https://ruby-doc.org/stdlib-2.5.1/libdoc/forwardable/rdoc/Forwardable.html#method-i-def_delegators) (plural), which is a similar method in Ruby's `Forwardable` module.  `#def_delegators` is the same as `#def_delegator` except it takes a variable number of positional parameters, each specifying a different method symbol to delegate to another object. Since Sorbet already has basic support for `#def_delegator`, it would be convenient for it to support `#def_delegators` in a similar way.

### Test plan

I updated the tests that were first added in https://github.com/sorbet/sorbet/pull/3636 to include new test cases for `def_delegators`. I more or less applied the same test scenarios to `#def_delegators`.

### DIFF
--- a/rewriter/DefDelegator.cc
+++ b/rewriter/DefDelegator.cc
@@ -8,13 +8,29 @@ using namespace std;
 
 namespace sorbet::rewriter {
 
-vector<ast::ExpressionPtr> DefDelegator::run(core::MutableContext ctx, const ast::Send *send) {
+/// Generate method stub and sig for the delegator method
+void generateStub(vector<ast::ExpressionPtr> &methodStubs, const core::LocOffsets &loc,
+                  const core::NameRef &methodName) {
+    // sig {params(arg0: T.untyped, blk: Proc).returns(T.untyped)}
+    auto sigArgs = ast::MK::SendArgs(ast::MK::Symbol(loc, core::Names::arg0()), ast::MK::Untyped(loc),
+                                     ast::MK::Symbol(loc, core::Names::blkArg()),
+                                     ast::MK::Nilable(loc, ast::MK::Constant(loc, core::Symbols::Proc())));
+
+    methodStubs.push_back(ast::MK::Sig(loc, std::move(sigArgs), ast::MK::Untyped(loc)));
+
+    // def $methodName(*arg0, &blk); end
+    ast::MethodDef::ARGS_store args;
+    args.emplace_back(ast::MK::RestArg(loc, ast::MK::Local(loc, core::Names::arg0())));
+    args.emplace_back(ast::make_expression<ast::BlockArg>(loc, ast::MK::Local(loc, core::Names::blkArg())));
+
+    methodStubs.push_back(
+        ast::MK::SyntheticMethod(loc, loc, methodName, std::move(args), ast::MK::RaiseUnimplemented(loc)));
+}
+
+/// Handle #def_delegator for a single delegate method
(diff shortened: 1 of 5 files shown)


jez:
@jeffcarbs you added DefDelegators.cc. Can you look at this too?

jeffcarbs:
I think that PR was my first time writing C++ 🙂 so I can't speak much to the code/refactoring. Based on the fact that none of the original tests were touched and they're all still passing I think this looks good to me 👍

froydnj:

→ https://go/builds/bui_JTaxR3koYs0zia
→ https://go/builds/bui_JTaxVo4HEVQdp2

rewriter/DefDelegator.cc
```diff
@@ -47,26 +63,61 @@ vector<ast::ExpressionPtr> DefDelegator::run(core::MutableContext ctx, const ast
...
-        ast::MK::SyntheticMethod(loc, loc, methodName, std::move(args), ast::MK::RaiseUnimplemented(loc)));
+/// Handle #def_delegators for zero or more delegate methods
+vector<ast::ExpressionPtr> runDefDelegators(core::MutableContext ctx, const ast::Send *send) {
+    vector<ast::ExpressionPtr> methodStubs;
+    auto loc = send->loc;
 
-    // Include the original call to def_delegator so sorbet will still type-check it
+    // This method takes 1.. positional arguments and no keyword args:
+    // `def_delegators(accessor, methods*)`
+    if (!(send->numPosArgs >= 1)) {
```
froydnj, line 82: Nit: I think this is more idiomatic as `send->numPosArgs == 0`.

froydnj (approved):
This looks good!  Can you push a fix for the tiny nit below?  I'll merge once that's done.

rewriter/DefDelegator.cc
richardxia: Oh yes, just amended my commit to make that change and force-pushed it back up.

richardxia:
Thanks for reviewing!

PR #4207, opened 2021-05-12
Allow abstract modules to require ONE unrelated ancestor class

Morriar:
### Motivation

The validator for `requires_ancestor` was considering abstract modules like abstract class thus this was rejected while valid:

```rb
class R; end

module M # error: M` requires unrelated class `R` making it impossible to include
  extend T::Helpers
  abstract!
  requires_ancestor R
end
```

This PR makes this example pass correctly.

### Test plan

See included automated tests.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -544,7 +544,7 @@ void validateUnsatisfiableRequiredAncestors(core::Context ctx, const core::Class
         }
     }
 
-    if (data->isClassOrModuleAbstract()) {
+    if (data->isClassOrModuleClass() && data->isClassOrModuleAbstract()) {
         for (auto ancst : requiredClasses) {
             if (!sym.data(ctx)->derivesFrom(ctx, ancst.symbol) && !ancst.symbol.data(ctx)->derivesFrom(ctx, sym)) {
                 if (auto e = ctx.state.beginError(data->loc(), core::errors::Resolver::UnsatisfiableRequiredAncestor)) {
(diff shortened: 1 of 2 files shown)


elliottt (approved):
I think this change makes sense, thanks for the test cases!

elliottt (changes requested):
Could you add some test cases for `interface!` as well? Should `interface!` be treated the same way as `abstract!`?

Morriar:
> Should interface! be treated the same way as abstract!

[This check](https://github.com/sorbet/sorbet/blob/master/definition_validator/validator.cc#L547-L558) was ensuring that a class does not require an ancestor that is unrelated to itself. We just need to relax it for modules that are abstract because they might be included in a class making the requirement possible:

```rb
class RA; end

module M;
  requires_ancestor RA # The class including `M` must inherit `RA`
  abstract! # This was making the check behave as if `M` was a class
end

class C < RA
  include M # The requirement is satisfied since `C` can inherit `RA`
end
```

For all modules (abstract, interface or normal module), the check will be made [here](https://github.com/sorbet/sorbet/blob/master/definition_validator/validator.cc#L564-L581) so we ensure we don't have this case:

```rb
class RA1; end
class RA2; end

module I
  interface!
  requires_ancestor RA1, RA2 # You can't do this since you can't inheriting from both `RA1` and `RA2`
end
```

I added a few test cases for interfaces 👍

elliottt (approved):
Thanks for the additional tests!

PR #4208, opened 2021-05-12
add T::Configuration.can_enable_vm_prop_serde?

froydnj:
### Motivation

It turns out there was one wrinkle in #4171: to know whether you could safely call `T::Configuration.enable_vm_prop_serde`, you had to dive into internal implementation details of `sorbet-runtime`, which is not a good thing.  Instead of that, we should provide a method on `T::Configuration` that hides the implementation details.

The test for ensuring that VM-based serialization/deserialization has been removed, given the hard assertion in `enable_vm_prop_serde`; happy to have a discussion about whether we should remove the assert and/or find some way to work around the assert.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/test/types/configuration.rb
+++ b/gems/sorbet-runtime/test/types/configuration.rb
@@ -327,6 +327,28 @@ module Opus::Types::Test
           assert_equal('T.let: Expected type Integer, got type String', e.message.split("\n").first)
         end
       end
+
+      describe 'enable_vm_prop_serde' do
+        it "fails if the VM doesn't support it" do
+          return if T::Configuration.can_enable_vm_prop_serde?
+
+          assert_raises(RuntimeError) do
+            T::Configuration.enable_vm_prop_serde
+          end
+        end
+
+        it "succeeds if the VM does support it" do
+          return unless T::Configuration.can_enable_vm_prop_serde?
+
+          was_enabled = T::Configuration.use_vm_prop_serde?
+
+          begin
+            T::Configuration.enable_vm_prop_serde
+          ensure
+            T::Configuration.disable_vm_prop_serde unless was_enabled
+          end
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


gems/sorbet-runtime/test/types/configuration.rb
```diff
@@ -328,31 +328,5 @@ def @mod.foo(a)
...
-          i = c.new(p: 5)
-          e = assert_raises(NoMethodError) do
-            s = i.serialize
-          end
-          assert_equal(:generate2, e.name)
-        ensure
-          T::Configuration.disable_vm_prop_serde unless was_enabled
-        end
-      end
-    end
```
jez, line 356: Shouldn't this test still pass as written? Why delete it? At the least, why have we not replaced it with another test?

froydnj: Because `T::Configuration.enable_vm_prop_serde` now hard-asserts if the VM doesn't support the necessary bits, so the test falls over.  I don't really think it's worth dancing around that to placate the test.

jez: Ah sorry I missed that line. But we can still add a new test, right?

froydnj: We could add:

```ruby
describe 'enable_vm_prop_serde' do
  it "fails if the VM doesn't support it" do
    return if T::Configuration.can_enable_vm_prop_serde?

    assert_raises(RuntimeError) do
      T::Configuration.enable_vm_prop_serde
    end
  end
end
```

Or somesuch?

jez: Sure. Even better would be to run `enable_vm_prop_serde` in both cases though, right? In one ensure that it raises, in the other ensure that it doesn't

jez (approved):

PR #4214, opened 2021-05-14
Fix Pathname#find signature

Bo98:
### Motivation
`ignore_error` is an optional keyword argument, as documented at https://ruby-doc.org/stdlib-2.6.3/libdoc/pathname/rdoc/Pathname.html#method-i-find.

### DIFF
--- a/rbi/stdlib/pathname.rbi
+++ b/rbi/stdlib/pathname.rbi
@@ -809,17 +809,11 @@ class Pathname < Object
   sig do
     params(
         ignore_error: T::Boolean,
-        blk: T.proc.params(arg0: Pathname).returns(BasicObject),
+        blk: T.nilable(T.proc.params(arg0: Pathname).void),
     )
-    .returns(T.untyped)
+    .returns(T.nilable(T::Enumerator[Pathname]))
   end
-  sig do
-    params(
-        ignore_error: T::Boolean,
-    )
-    .returns(T::Enumerator[Pathname])
-  end
-  def find(ignore_error, &blk); end
+  def find(ignore_error: true, &blk); end
 
   # Return `true` if the receiver matches the given pattern.
   #


aprocter-stripe:

→ https://go/builds/bui_JU5Ds5xOseHCyB
→ https://go/builds/bui_JU5DJsrW7QTTlX

aprocter-stripe (changes requested):
It looks like the tests are failing because `find` has more than one overloaded signature, which isn't compatible with keyword args. If the return type weren't dependent on the choice of overloading, I suppose we could just make the `blk` nilable instead of relying on overloading, but as things stand I'm not sure how to make it work without falling back to `returns(T.untyped)` in all cases.

@jez thoughts on how to make this work? (I might very well be missing something!)

jez:
> thoughts on how to make this work

you will have to either remove the overloads and make it have a single signature, or you will have to make it overloaded without keyword arguments by e.g. making an overly wide signature that accepts any keyword args or any args.

Bo98:
I completely missed the double signature here - my bad.

I'm not sure I understand why it's `T.untyped`. When does `Pathname.find` with a block return anything?

I'm leaning torwards a nilable block with a nilable return value. In fact, that's exactly how it's defined for `Find.find`: https://github.com/sorbet/sorbet/blob/4f146522e25047f6527a98797672f9b1c462eced/rbi/stdlib/find.rbi#L32-L39

aprocter-stripe:

→ https://go/builds/bui_JVFyWvPYAFjm6Z
→ https://go/builds/bui_JVFywmq6z2E2HW

aprocter-stripe:
> I'm leaning torwards a nilable block with a nilable return value. In fact, that's exactly how it's defined for `Find.find`:

Ah, good find (no pun intended)!

The `nilable(Enumerator)` formulation looks reasonable to me. Thanks for the fix!

aprocter-stripe (approved):

PR #4219, opened 2021-05-18
Add the missing Enumerable#chain method and signature

lugray:
[Enumerable#chain](https://ruby-doc.org/core-2.6/Enumerable.html#method-i-chain) introduced in ruby 2.6 wasn't defined.

### DIFF
--- a/gems/sorbet/test/hidden-method-finder/simple/ruby_2_6_hidden.rbi.exp
+++ b/gems/sorbet/test/hidden-method-finder/simple/ruby_2_6_hidden.rbi.exp
@@ -2027,8 +2027,6 @@ class Encoding
 end
 
 module Enumerable
-  def chain(*_); end
-
   def sum(*_); end
 end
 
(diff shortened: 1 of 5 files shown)


aprocter-stripe:

→ https://go/builds/bui_JVdeRDEvcFRp9s
→ https://go/builds/bui_JVdeWqgtKF93DP

aprocter-stripe (changes requested):
Thanks for the rbi update! I think this signature seems reasonable.

Looks like tests are failing because output expectations for the hidden method finder need to be updated. Would you mind running `gems/sorbet/test/hidden-method-finder/update_hidden_methods_exp.sh` and committing the changes it makes? I should be able to stamp and merge this once that's fixed.

lugray:
I had to run `PATH="./:$PATH" gems/sorbet/test/hidden-method-finder/update_hidden_methods_exp.sh` as I don't generally have `bazel` on my PATH, but otherwise all seemed to go well and the diff makes sense. Thanks for the help.

aprocter-stripe (approved):
Thanks again @lugray!

PR #4223, opened 2021-05-19
`FixedHash` serialization should not assume symbol keys

paracycle:
Runtime output should make the distinction that static checker does: https://sorbet.run/#%23%20typed%3A%20true%0AT.reveal_type(%7Ba%3A%20String%2C%20b%3A%20Integer%7D)%0AT.reveal_type(%7B%22a%22%20%3D%3E%20String%2C%20%22b%22%20%3D%3E%20Integer%7D)

### Motivation

`FixedHash` current serializes itself assuming that the keys are symbols but it can also be the case that keys are strings. Thus, it is more correct to use the `{ a => b }` syntax when serializing fixed hashes entries with non-symbol keys, along with calling `inspect` on those keys instead of an implicit `to_s`.

Entries with symbol keys are serialized using the `{a: b}` syntax stil.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/types/fixed_hash.rb
+++ b/gems/sorbet-runtime/lib/types/types/fixed_hash.rb
@@ -13,7 +13,15 @@ module T::Types
 
     # @override Base
     def name
-      "{#{@types.map {|(k, v)| "#{k}: #{v}"}.join(', ')}}"
+      entries = @types.map do |(k, v)|
+        if Symbol === k && ":#{k}" == k.inspect
+          "#{k}: #{v}"
+        else
+          "#{k.inspect} => #{v}"
+        end
+      end
+
+      "{#{entries.join(', ')}}"
     end
 
     # @override Base
--- a/gems/sorbet-runtime/test/types/types_to_ruby.rb
+++ b/gems/sorbet-runtime/test/types/types_to_ruby.rb
@@ -29,6 +29,9 @@ class Opus::Types::Test::TypesToRubyTest < Critic::Unit::UnitTest
 
     # FixedHash:
     [{a: T.nilable(String)}, "{a: T.nilable(String)}"],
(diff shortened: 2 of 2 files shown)


jez (approved):
Nice find, thanks!

jez:
> If desired, I could update the implementation so that symbol keys are rendered as `a: b` and all other keys as `a => b`, like how the static checker does.

I think this would be convenient? I'm not sure whether we depend on this output internally (given the nature of the bug I'm guessing we don't?) but it might be nice if only for the sake of good error messages.

paracycle:
@jez Agreed, I made the change to special case `Symbol` keys to use the `{a: b}` syntax and updated the PR description accordingly.

gems/sorbet-runtime/lib/types/types/fixed_hash.rb
```diff
@@ -13,7 +13,15 @@ def initialize(types)
 
     # @override Base
     def name
-      "{#{@types.map {|(k, v)| "#{k}: #{v}"}.join(', ')}}"
+      entries = @types.map do |(k, v)|
+        if Symbol === k
+          "#{k}: #{v}"
```
jez, line 18: Not sure how far down this rabit hole you care to go:

```
irb(main):001:0> x = :"foo bar"
=> :"foo bar"
irb(main):002:0> "#{x} is it"
=> "foo bar is it"
```

PR is still strictly an improvement, so I'm still fine, but figured I'd call it out.

paracycle: Ohhh, good call. I will go all the way down the rabbit hole :)

paracycle: Ok, I think `":#{k}" == k.inspect" should be guarding those cases too. Pushed the change and added tests for those cases.

jez:
Stripe build passed

http://go/builds/bui_JW0sXBcWozauKD

PR #4226, opened 2021-05-20
Factor out typeDrivenAutocorrect helper

jez:
### Motivation

We have a bunch of duplicated code.

There is more code that I want to add, and don't want to have to add it
in both places.

This change makes those future changes easier, but does not change any
behavior.

### Test plan

Nothing changes. Covered by existing tests.

### DIFF
--- a/core/TypeDrivenAutocorrect.h
+++ b/core/TypeDrivenAutocorrect.h
@@ -1,0 +1,16 @@
+#ifndef SORBET_TYPE_DRIVEN_AUTOCORRECT_H
+#define SORBET_TYPE_DRIVEN_AUTOCORRECT_H
+
+#include "core/Error.h"
+#include "core/GlobalState.h"
+#include "core/TypeConstraint.h"
+#include "core/Types.h"
+
+namespace sorbet::core {
+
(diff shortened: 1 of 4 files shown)


core/TypeDrivenAutocorrect.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include "core/Error.h"
+#include "core/GlobalState.h"
+#include "core/TypeConstraint.h"
+#include "core/Types.h"
+
+namespace sorbet::core {
+
+void typeDrivenAutocorrect(const core::GlobalState &gs, core::ErrorBuilder &e, core::Loc loc,
+                           core::TypeConstraint &constr, core::TypePtr expectedType, core::TypePtr actualType);
```
elliottt, line 12: Any reason to not wrap this in a class? We tend to avoid top-level functions elsewhere in the codebase.

elliottt (approved):
Looks good! Just one non-blocking comment.

core/TypeDrivenAutocorrect.h
jez: Pushed a follow up commit.

```
typeDrivenAutocorrect -> TypeDrivenAutocorrect::maybeAutocorrect
```

Issue #2390, opened 2019-12-23
Add type definitions for new methods in Ruby 2.7


connorshea:
### Problem

Ruby 2.7 comes out on Wednesday and adds some new methods to core classes. Currently, Sorbet doesn't have type definitions for any of these.

### Proposed solution

Add type signatures for all the new methods in Ruby 2.7 :)

**New Methods:**

- [x] `Array#intersection` [Feature #16155](https://bugs.ruby-lang.org/issues/16155)
- [x] `Array#minmax` (faster implementation than `Enumerable#minmax`) [Bug #15929](https://bugs.ruby-lang.org/issues/15929)
- [ ] `Complex#<=>` [Bug #15857](https://bugs.ruby-lang.org/issues/15857)
- [x] `Enumerable#filter_map`. [Feature #15323](https://bugs.ruby-lang.org/issues/15323)
- [x] `Enumerable#tally` [Feature #11076](https://bugs.ruby-lang.org/issues/11076)
- [ ] `Enumerator.produce` [Feature #14781](https://bugs.ruby-lang.org/issues/14781)
- [ ] `Enumerator::Lazy#eager` [Feature #15901](https://bugs.ruby-lang.org/issues/15901)
- [ ] `Enumerator::Yielder#to_proc` [Feature #15618](https://bugs.ruby-lang.org/issues/15618)
- [ ] `Fiber#raise` [Feature #10344](https://bugs.ruby-lang.org/issues/10344)
- [ ] `FrozenError#receiver` [Feature #15751](https://bugs.ruby-lang.org/issues/15751)
- [ ] `GC.compact` [Feature #15626](https://bugs.ruby-lang.org/issues/15626)
- [ ] `IO#set_encoding_by_bom` [Bug #15210](https://bugs.ruby-lang.org/issues/15210)
- [x] `Module#const_source_location` [Feature #10771](https://bugs.ruby-lang.org/issues/10771)
- [x] `Module#ruby2_keywords` [Bug #16154](https://bugs.ruby-lang.org/issues/16154)
- [x] `Proc#ruby2_keywords` [Feature #16404](https://bugs.ruby-lang.org/issues/16404)
- [x] `Range#minmax` (faster implementation than `Enumerable#minmax`) [Bug #15807](https://bugs.ruby-lang.org/issues/15807)
- [x] `Symbol#start_with?` [Feature #16348](https://bugs.ruby-lang.org/issues/16348)
[...]

PR #4227, opened 2021-05-20
Adds signature for `Complex#<=>`

mutecipher:
Adds the signature for `Complex#<=>` to `rbi/core/complex.rbi`.

### Motivation

This was a requested update from https://github.com/sorbet/sorbet/issues/2390.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/complex.rbi
+++ b/rbi/core/complex.rbi
@@ -273,6 +273,23 @@ class Complex < Numeric
   end
   def ==(arg0); end
 
+  # Returns 0, 1, -1 based on the imaginary number's real part compared to the object. Otherwise, returns nil.
+  #
+  # ```ruby
+  # Complex(2, 3)  <=> Complex(2, 3)   #=> nil
+  # Complex(2, 3)  <=> 1               #=> nil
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_JWLh2mIAvJLR0r
→ https://go/builds/bui_JWLh28SsHj7sDD

rbi/core/complex.rbi
```diff
@@ -273,6 +273,23 @@ class Complex < Numeric
   end
   def ==(arg0); end
 
+  # Returns 0, 1, -1 based on the imaginary number's real part compared to the object. Otherwise, returns nil.
```
jez, line 276:
```suggestion
  # If `cmp`'s imaginary part is zero, and `object` is also a real number (or a
  # [`Complex`](https://docs.ruby-lang.org/en/2.7.0/Complex.html) number where
  # the imaginary part is zero), compare the real part of `cmp` to object.
  # Otherwise, return nil.
```

This is the official documentation.

jez (approved):

jez:
Thanks!

PR #4229, opened 2021-05-21
Add Boolean -> T::Boolean autocorrect

jez:
### Motivation

Stripe's codebase has something that looks like this.

```ruby
# typed: true

extend T::Sig

module ::Boolean
end

class TrueClass
  include Boolean
end

class FalseClass
  include Boolean
end

sig {params(x: Boolean).returns(T::Boolean)}
def boolean_to_t_boolean(x)
  x
end
```

It's a dumb legacy thing that we'd love to get rid of, but it causes no end of
questions when people see it for the first time. It also shows up from time to
time when fixing previously-silent errors (like those in #4134).

The fix is just to prepend `!!`, which converts `Boolean` to `T::Boolean`, which
we can write a quick autocorrect for.

### Test plan

Need to write tests.

### DIFF
--- a/core/TypeDrivenAutocorrect.cc
+++ b/core/TypeDrivenAutocorrect.cc
@@ -18,6 +18,15 @@ void TypeDrivenAutocorrect::maybeAutocorrect(const GlobalState &gs, ErrorBuilder
         if (!withoutNil.isBottom() &&
             Types::isSubTypeUnderConstraint(gs, constr, withoutNil, expectedType, UntypedMode::AlwaysCompatible)) {
             e.replaceWith("Wrap in `T.must`", loc, "T.must({})", loc.source(gs).value());
+        } else if (Types::isSubTypeUnderConstraint(gs, constr, expectedType, Types::Boolean(),
+                                                   UntypedMode::AlwaysCompatible)) {
+            if (core::isa_type<ClassType>(actualType)) {
+                auto classSymbol = core::cast_type_nonnull<ClassType>(actualType).symbol;
+                if (classSymbol.exists() && classSymbol.data(gs)->owner == core::Symbols::root() &&
+                    classSymbol.data(gs)->name == core::Names::Constants::Boolean()) {
+                    e.replaceWith("Prepend `!!`", loc, "!!({})", loc.source(gs).value());
+                }
+            }
         }
     }
 }
--- a/test/testdata/infer/boolean_t_boolean.rb
+++ b/test/testdata/infer/boolean_t_boolean.rb
@@ -1,0 +1,34 @@
+# typed: true
+
+extend T::Sig
+
+module ::Boolean
+end
+
+class TrueClass
+  include Boolean
+end
+
+class FalseClass
+  include Boolean
+end
+
+sig {params(arg0: T.untyped).returns(Boolean)}
+def returns_boolean(arg0)
+  false
+end
+
+sig {params(x: Boolean).returns(T::Boolean)}
+def boolean_to_t_boolean(x)
+  x # error: Expected `T::Boolean` but found `Boolean`
+end
+
+sig {params(x: Boolean).returns(T::Boolean)}
+def explicit_return(x)
+  return x # error: Expected `T::Boolean` but found `Boolean`
+end
+
+sig {returns(T::Boolean)}
+def can_have_spaces
+  returns_boolean T.unsafe(nil) # error: Expected `T::Boolean` but found `Boolean`
+end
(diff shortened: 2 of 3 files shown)


core/TypeDrivenAutocorrect.cc
```diff
@@ -18,6 +18,15 @@ void TypeDrivenAutocorrect::maybeAutocorrect(const GlobalState &gs, ErrorBuilder
         if (!withoutNil.isBottom() &&
             Types::isSubTypeUnderConstraint(gs, constr, withoutNil, expectedType, UntypedMode::AlwaysCompatible)) {
             e.replaceWith("Wrap in `T.must`", loc, "T.must({})", loc.source(gs).value());
+        } else if (Types::isSubTypeUnderConstraint(gs, constr, expectedType, Types::Boolean(),
+                                                   UntypedMode::AlwaysCompatible)) {
+            if (core::isa_type<ClassType>(actualType)) {
+                auto classSymbol = core::cast_type_nonnull<ClassType>(actualType).symbol;
+                if (classSymbol.exists() && classSymbol.data(gs)->owner == core::Symbols::root() &&
+                    classSymbol.data(gs)->name == core::Names::Constants::Boolean()) {
```
Morriar, line 26: Why only `Boolean`? It seems like a lot of other cases could benefit from such an autocorrect?

Morriar: Anything that is a BasicObject could potentially be autocorrected here since `!` is defined: https://ruby-doc.org/core-2.5.1/BasicObject.html#method-i-21

jez: I thought so too, but then I couldn’t come up with a compelling test case, so I erred on the cautious side and didn’t make it general. Do you have a good test case that it should apply to?

jez: Specifically, given how few things are falsy in Ruby, all the examples I came up with seemed contrived. Like Integer to T::Boolean autocorrect seems bad—that almost certainly indicates a bug in your code, because Integer is always truthy.

Morriar: It depends how you see the autocorrect:

* If it's a tool used to mass-port a project to correct typing then indeed Sorbet should not suggest this fix.
* If it's a tool that suggests solutions to local problems that can be hand-picked by the developer, then it makes sense to suggest it.

I personally see it as the later rather than the former so for me it would make sense to always suggest it and let the developer decide if they want to apply it or not.

jez: I think that for the time being I'll ship the minimal thing. I'm open to expanding this in the future.

test/testdata/infer/boolean_t_boolean.rb
```diff
@@ -0,0 +1,48 @@
...
+end
+
+sig {returns(T::Boolean)}
+def can_have_spaces
+  returns_boolean T.unsafe(nil) # error: Expected `T::Boolean` but found `Boolean`
+end
+
+sig {params(x: BadBang).returns(T::Boolean)}
+def should_not_suggest_bang_bang(x)
+  x # error: Expected `T::Boolean` but found `Boolean`
```
aprocter-stripe, line 47:
```suggestion
  x # error: Expected `T::Boolean` but found `BadBang`
```

(Looks like CI caught it, at any rate)

test/testdata/infer/boolean_t_boolean.rb.autocorrects.exp
```diff
@@ -0,0 +1,50 @@
...
+end
+
+sig {returns(T::Boolean)}
+def can_have_spaces
+  !!(returns_boolean T.unsafe(nil)) # error: Expected `T::Boolean` but found `Boolean`
+end
+
+sig {params(x: BadBang).returns(T::Boolean)}
+def should_not_suggest_bang_bang(x)
+  x # error: Expected `T::Boolean` but found `Boolean`
```
aprocter-stripe, line 48:
```suggestion
  x # error: Expected `T::Boolean` but found `BadBang`
```

core/TypeDrivenAutocorrect.cc
aprocter-stripe, line 26: Do we want to suggest in the case where `actualType` is `::TrueClass` or `::FalseClass`? (Or is that not possible for some reason?)

aprocter-stripe (approved):
Seems reasonable to me! I like @Morriar's idea of expanding this to other bangable types, but landing this for now seems reasonable.

core/TypeDrivenAutocorrect.cc
jez: `TrueClass` is already a subtype of `T::Boolean`, so it wouldn't report an error and we wouldn't get here.

aprocter-stripe: Oh yeah, good point :)

PR #4231, opened 2021-05-22
Add `*.autocorrects.exp` expectation tests

jez:
### Motivation

We have a lot of `test/cli/` tests that all have more or less the same 50 lines
of shell copy/pasted and slightly modified for the purpose of testing
`--autocorrect` results.

This makes it so that you can test autocorrects like anything else in Sorbet.
The idea is to treat the autocorrected file contents as (conceptually) as `-p`
output of Sorbet.

If you have a test file `test/testdata/infer/foo.rb`, you can now create a file
`test/testdata/infer/foo.rb.autocorrects.exp` which will contain a dump of the
autocorrected file content, had the file been run with `--autocorrect` to edit
it in place.

I think this approach is elegant and easy to use. I would have no problem
requesting someone add autocorrect tests on a PR now, because it is next to zero
friction.

### Alternatives considered

We already have `^^^ apply-code-action` assertions. The way those work is that
each test must have one addition file per assertion, containing the file
contents after the individually requested code action is applied.

That's useful for testing that LSP works (that the apply code action handler
actually carries out the right edit), but it's annoying when you you want to
test individual code actions, not the LSP harness.

I considered adding `^^^ quickfix: <edit contents>`, but:

- it would have been hard distinguishing between whether `^` meant a zero-width
  loc starting at that point, or a length 1 loc (this distinction comes up
  frequently with autocorrects that insert something at a point).

- it would not have been easy to allow for autocorrects that span multiple lines
  (`^^^` ranges are not allowed to span multiple lines either)

- it would not have been able to easily support autocorrects that supply
  multiple edits (e.g., the sig suggestion autocorrect will insert a sig, but
  also `extend T::Sig` if the class is missing `extend T::Sig`).

- related to the above, the location/region where an error is reported is not
  the same as the region of text that will be replaced

So there was a mountain of evidence against using LSP-style position assertions
for this.

### Commit summary

The core change is actually quite small, and I've broken it up to showcase that.

- **Refactor a private Loc helper into a public File helper** (793684124)

  Same reasoning as previous commit: I want to be able to use this in the
  tests.

- **Move some logic from ErrorFlusherStdout to AutocorrectSuggestion** (bc9e9e0a8)

  This makes it easier to use just the parts I want in the test.

  Conceptually, it splits the steps into "read files and apply
  autocorrects" and "write autocorrected files out". In the tests, I just
  want the reading--the writing will happen in a special way so as to not
  clobber the input files.

  A note on the deleted test:

  I don't like Sorbet's unit tests.

  This one in particular is completely redundant with Sorbet's existing
  autocorrect tests.

  To keep this test, I would have had to rewrite the test to operate with
  `FileSystem &fs`. But that's exactly the code that the new test suite
  I'm writing does, so why bother?

- **Check for `*.autocorrects.exp` expectation files** (9f3fc8a39)

- **Incorporate autocorrects.exp into update_testdata_exp.sh** (b8eb64b0a)

- **Add example for autocorrect_method_return.rb** (5b5e49313)

- **Add an example for multi-file support too** (7ea42712f)

- **Update README.md to talk about multi file exp tests** (aa65339db)

- **Allow enabling --suggest-unsafe in testdata tests** (12f3069b6)

- **Add test showcasing --suggest-unsafe** (f8d5bb537)

### Test plan

Made sure that my new tests fail in CI:

✅ https://buildkite.com/sorbet/sorbet/builds/15721#d7c9bf7c-317b-4801-bc25-2d1c9f147425

and that my `update_testdata_exp.sh` changes are able to successfully fix them when run locally: ✅


froydnj (approved):
Thanks for splitting things up so neatly.

It's sort of weird that the autocorrect `.exp` has both the autocorrected bit and the error mixed up together, but given that they are autogenerated, I guess that should not be much of a problem in practice?

jez:
> It's sort of weird that the autocorrect `.exp` has both the autocorrected bit and the error mixed up together, but given that they are autogenerated, I guess that should not be much of a problem in practice?

Yeah, I think it shouldn't be much of a problem.

My biggest problem right now is that it's a little tricky to record a before/after commit that captures just what a new autocorrect does, because if there are no autocorrects at all, the `*.autocorrect.exp` file will be empty, instead of containing all the test files' sources. E.g., I want to show a diff that's like

```diff
-  x
+  T.unsafe(x)
```

but right now the only diff i will be able to show is

```diff
+def foo(x)
+  T.unsafe(x)
+end
```

(I might change that in the future.)

PR #4234, opened 2021-05-26
Document error code: 7035

builtbyproxy:
### Motivation

I keep getting the error code 7035 when I run a select over an array of objects rather than a single object. I've documented the fix.

### Test plan

No tests because this is a documentation change.

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1033,4 +1033,28 @@ def foo(x, y)
 end
 ```
 
+## 7035
+
+Sorbet detected that the method `select` does not take a block. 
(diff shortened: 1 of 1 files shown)


jez (changes requested):
It looks like you're using `sorbet-rails`, and running into a limitation of that:

https://github.com/chanzuckerberg/sorbet-rails#select-with-a-block

The 7035 error code is more general than just this specific rails-ism.

Do you think you can take a pass over your doc to hedge a little more? Start by describing the error simply and generally (e.g., using something available in the standard library), and then following up with "if you are using sorbet-rails, see this note in their readme"?

PR #2078, opened 2019-11-02
Handle more situations in check for instantiating abstract classes

XrXr:
### Motivation
The two added tests did not pass before this commit.

### Test plan

See included automated tests.

PR #4236, opened 2021-05-27
Add configuration option for short-circuiting type assertions during runtime

RyanBrushett:
### Motivation

We patch `let`, `must`, `cast`, `bind`, and `absurd` to be a noop - basically anything that can ultimately call `inline_type_error_handler`. Having this be configurable might be handy for others.

I'm very keen for feedback on the idea, approach, and scope here for sure. If there are other calls I should capture or if the net is cast a bit too wide, let me know. I followed the pattern of `enable_legacy_t_enum_migration_mode` a few lines above in `configuration.rb`.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -126,7 +126,7 @@ module T
   # exception at runtime if the value doesn't match the type.
   #
   # Compared to `T.let`, `T.cast` is _trusted_ by static system.
-  def self.cast(value, type, checked: true)
+  def self.cast(value, type, checked: T::Configuration.runtime_type_assertions_enabled?)
     return value unless checked
 
     Private::Casts.cast(value, type, cast_method: "T.cast")
@@ -141,7 +141,7 @@ module T
   #
   # If `checked` is true, raises an exception at runtime if the value
   # doesn't match the type.
-  def self.let(value, type, checked: true)
+  def self.let(value, type, checked: T::Configuration.runtime_type_assertions_enabled?)
     return value unless checked
 
     Private::Casts.cast(value, type, cast_method: "T.let")
@@ -160,7 +160,7 @@ module T
   #
   # If `checked` is true, raises an exception at runtime if the value
   # doesn't match the type (this is the default).
-  def self.bind(value, type, checked: true)
+  def self.bind(value, type, checked: T::Configuration.runtime_type_assertions_enabled?)
     return value unless checked
 
     Private::Casts.cast(value, type, cast_method: "T.bind")
@@ -210,6 +210,7 @@ module T
   #
   # sig {params(arg: T.nilable(A)).returns(A)}
   def self.must(arg)
+    return arg unless T::Configuration.runtime_type_assertions_enabled?
     return arg if arg
     return arg if arg == false
 
@@ -231,6 +232,7 @@ module T
   # happens. Commonly used to assert that a case or if statement exhausts all
   # possible cases.
   def self.absurd(value)
+    return value unless T::Configuration.runtime_type_assertions_enabled?
     msg = "Control flow reached T.absurd."
 
     case value
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -523,6 +523,24 @@ module T::Configuration
     @legacy_t_enum_migration_mode || false
   end
 
+  # Disable runtime type assertions. This is useful for a bit of a performance
+  # boost at the expense of type safety during runtime.
+  # If this option is not explicity set, its default state will mean that all
+  # type assertions behave normally during runtime.
+  #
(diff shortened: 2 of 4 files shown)


froydnj (commented):
I think @jez has more opinions than I do about this.

You'll also want to update `rbi/sorbet/t.rbi` to add the new methods on `T::Configuration`.

gems/sorbet-runtime/lib/types/_types.rb
```diff
@@ -126,7 +126,7 @@ def self.type_parameter(name)
   # exception at runtime if the value doesn't match the type.
   #
   # Compared to `T.let`, `T.cast` is _trusted_ by static system.
-  def self.cast(value, type, checked: true)
+  def self.cast(value, type, checked: T::Configuration.runtime_type_assertions_enabled?)
```
jez, line 129: Noting before I forget: this will incur the runtime penalty of calling `T::Configuration.runtime_type_assertions_enabled?` every time `T.let` / `T.cast` / etc. is called.

I wonder if there are alternatives that don't suffer from this.

jez (commented):
I'm out today, and monday is a US holiday. I will get to this some time next week. Thanks!

gems/sorbet-runtime/lib/types/_types.rb
RyanBrushett: Good shout. I'll give it some thought!

RyanBrushett: I forgot to get to this the weekend but I had a thought:

Similar to how setting the `default_checked_level` to `:never`, `:tests`, or `:always` gives you one chance to call it, could we do a similar thing with this kinda like:

- User calls something like `runtime_type_assertions(:disable)`, presumably really early in the application booting.
- We do like `T.singleton_class.prepend` new versions of the methods that are functionally noops
- That's it it's done and no longer configurable.

It should avoid any runtime penalty of checking another method call and it follows an existing pattern for setting something scary. 
Is that sensible? Again keen for feedback.

jez:
@RyanBrushett can you remind me what the motivation is here?

(Because if the motivation is primarily performance, the dominating factor in performance is actually allocating the `type` argument to all these cast instructions, which will happen regardless of whether the `checked` argument is true or false.)

gems/sorbet-runtime/lib/types/_types.rb
jez: > * We do like `T.singleton_class.prepend` new versions of the methods that are functionally noops

Heads up that we do not (cannot) use `prepend` in sorbet-runtime. Stripe's codebase does not support it. You can see previous discussions around not using `prepend` here: https://github.com/sorbet/sorbet/pull/2078

I think it would be fine to remove the original method and replace it with another method (we already have a helper for this, it's how sigs work. Grep for `replace_method`.

If you want to try implementing a solution that uses `replace_method` plus the "one-shot" logic you mention above, I'd be open to looking at the code.

(and again: please take a look at my question of motivation below, because if this is for primarily performance concerns, this will not solve them)

jez (commented):
^

paracycle:
> @RyanBrushett can you remind me what the motivation is here?
> 
> (Because if the motivation is primarily performance, the dominating factor in performance is actually allocating the `type` argument to all these cast instructions, which will happen regardless of whether the `checked` argument is true or false.)

@jez Indeed, allocating the `type` argument does take a lot of time, but the rest of the checks also take a non-trivial amount of time as well.

Here are relevant snippets to running `bundle exec rake bench:typecheck`:

### Master

```
T.let(..., Integer): 360.507 ns
T.let(..., T.nilable(Integer)): 1.287 μs
T.let(..., Example): 344.638 ns
T.let(..., T.nilable(Example)): 1.196 μs
```

### An ivar on module `T`, set to `true` (inline checks enabled)

```
T.let(..., Integer): 384.378 ns
T.let(..., T.nilable(Integer)): 1.277 μs
T.let(..., Example): 362.241 ns
T.let(..., T.nilable(Example)): 1.28 μs
```

### An ivar on module `T`, set to `false` (inline checks disabled)

```
T.let(..., Integer): 114.234 ns
T.let(..., T.nilable(Integer)): 714.101 ns
T.let(..., Example): 88.866 ns
T.let(..., T.nilable(Example)): 683.182 ns
```

As you can see, the time difference is non-trivial. For most calls, they take about 1/3 of the time and in some complex cases (like `T.nilable` construction) that falls down to ~1/2 of the time.

While this does not remove the overhead completely, it removes a lot of it.

RyanBrushett:
> can you remind me what the motivation is here?

Performance for sure. And yeah, it's not perfect to still have e.g. `T.let` allocate the type argument but we've found that having the method call return `value` still has quite a bit of an impact that we've been happy with.

Bad news: I'll close this for now, mostly because I won't get to it this week or early next, but I'll do my best to get back to this toward the end of next week.

Thanks for the chat and the feedback everyone! 👍 I appreciate your time! I'll pick this up again soon.

Issue #3604, opened 2020-10-31
IDE: Place limit on number of diagnostics to report


jvilk-stripe:
### Problem

If the user makes a big change to a key symbol in a big project, Sorbet can report thousands of errors that cause VS Code to drag to a crawl (and network bottlenecks if Sorbet is running remotely). This has caused issues internally for teams doing cross-cutting changes.

### Proposed solution

Sorbet's LSP server should place a (configurable) limit on the maximum number of diagnostics to report. We should experiment to understand what the default upper bound should be (500?).

A simple implementation could enforce this restriction within Sorbet's language server. If errors are sorted prior to filtering, then Sorbet will show a consistent set of errors to the user after a slow path.

A more complex implementation could enforce this restriction within Sorbet's typechecking passes and prevent them from constructing error messages the user doesn't see. This would be a meaningful performance improvement when dealing with thousands of errors (iirc, Sorbet's fuzzy lookup strategy for close matches can be expensive), but will require dealing with error reporting in a multithreaded context. One UX downside of this approach is that the user will not see a consistent set of errors, which honestly might outweigh the performance benefits.

akshay1992kalbhor:
Hey! Can I try this one? Can you guys help me with it?

jvilk-stripe:
Hey @akshay1992kalbhor! Sorry, I didn't see your comment until now. I think this is easily implemented in `ErrorCollector`'s `drainErrors` method:
https://github.com/sorbet/sorbet/blob/4ae08bf2/core/ErrorCollector.cc#L20

We could change `ErrorCollector` to:
1. Accept a limit in the constructor.
2. Perform a `stable_sort` on the errors in `drainErrors` and only report the first `limit`.

Then, we can add a new command line option to `options.cc` to set the limit and pipe it through to `ErrorCollector` construction.

PR #4239, opened 2021-06-02
LSP: Limit maximum number of diagnostics to report to the editor.

jvilk-stripe:
Limit the maximum number of diagnostics to report to the editor when in LSP mode. I have set the value to `1000`, but I may ratchet it down with testing on Stripe's codebase. This limit does not apply in CLI mode.

Before explaining how it works, recall that Sorbet's LSP server uses the concept of an epoch. An epoch is simply a typecheck run, which can be full (i.e., retypecheck _every_ file) or incremental (i.e., only retypecheck a subset of the files in the project). Incremental typecheck runs can _preempt_ full typecheck runs to respond to code edits during a full typecheck, so multiple epochs can be in play at once.

To manage the complexity of error reporting in the face of preemption, Sorbet's LSP server uses an `ErrorReporter` class which is responsible for sending diagnostics (i.e., typechecking errors) to the client. It does some sanity checking to avoid sending stale typechecking errors, since it is possible that a full typecheck may report errors for an older file revision _after_ it is preempted by a incremental typecheck. 

One other important detail of the LSP protocol is that servers update the set of errors in a file by sending the full list of errors in the file. That is, if a file has 5 errors, it is not possible to send 2 and then send 3 -- the server must send a complete listing. A corollary of this design decision is if a file has 0 errors but previously had >0, an empty list of diagnostics must be sent to the editor to invalidate the old errors.

With this PR, `ErrorReporter` is also responsible for limiting the number of diagnostics sent to the editor in the following manner:
* `ErrorReporter` contains a counter of the number of errors reported to the client (`clientErrorCount`).
* `clientErrorCount` is kept up-to-date as new errors are reported for files -- including when an incremental typechecking run shrinks the error count.
* `clientErrorCount` is reset to `0` whenever a full typecheck runs since that typecheck will re-report all errors in the project.

#### Determinism

When the limit is enforced, the set of errors reported is not guaranteed to be consistent between runs, since errors are reported in the order received from worker threads. Thus, the user may see a slightly different set of errors with each full retypecheck. Fixing this problem is somewhat at odds with error streaming.

One possible solution that could be bolted on to this effort is to stream out errors in a specific file order rather than in the order typechecked. Files enter the worker queue in a deterministic order, and we could re-sort the files on the way out into the same order. This fix would slightly reduce the responsiveness of error streaming and may not be worth the complication.

#### UX

The user does not receive an indication that Sorbet is limiting errors. `clangd` limits errors and also does not provide any user indication that it is doing so. I contemplated adding a faux error to a file that stated that errors are limited, but that is easily missed. Suggestions appreciated.

### Motivation

Sorbet effectively crashes VS Code when it reports thousands of errors.

Fixes https://github.com/sorbet/sorbet/issues/3604.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/ErrorReporter.cc
+++ b/main/lsp/ErrorReporter.cc
@@ -173,12 +173,17 @@ u4 ErrorReporter::lastDiagnosticEpochForFile(core::FileRef file) {
 }
 
 void ErrorReporter::sanityCheck() const {
-    DEBUG_ONLY(u4 errorCount = 0; for (auto &status
-                                       : this->fileErrorStatuses) {
+    if (!debug_mode) {
+        return;
+    }
+
+    u4 errorCount = 0;
+    for (auto &status : this->fileErrorStatuses) {
         if (status.lastReportedEpoch >= this->lastFullTypecheckEpoch) {
             errorCount += status.errorCount;
         }
-    } ENFORCE(errorCount == this->clientErrorCount));
+    }
+    ENFORCE(errorCount == this->clientErrorCount);
 }
 
 ErrorEpoch::ErrorEpoch(ErrorReporter &errorReporter, u4 epoch, bool isIncremental,
--- a/main/lsp/ErrorReporter.cc
+++ b/main/lsp/ErrorReporter.cc
@@ -19,7 +19,7 @@ vector<core::FileRef> ErrorReporter::filesWithErrorsSince(u4 epoch) {
     return filesUpdatedSince;
 }
 
-void ErrorReporter::beginEpoch(u4 epoch, vector<unique_ptr<Timer>> diagnosticLatencyTimers) {
+void ErrorReporter::beginEpoch(u4 epoch, bool isIncremental, vector<unique_ptr<Timer>> diagnosticLatencyTimers) {
     ENFORCE(epochTimers.find(epoch) == epochTimers.end());
     vector<Timer> firstDiagnosticLatencyTimers;
     if (config->getClientConfig().enableTypecheckInfo) {
@@ -139,10 +139,10 @@ u4 ErrorReporter::lastDiagnosticEpochForFile(core::FileRef file) {
     return getFileErrorStatus(file).lastReportedEpoch;
 }
 
-ErrorEpoch::ErrorEpoch(ErrorReporter &errorReporter, u4 epoch,
+ErrorEpoch::ErrorEpoch(ErrorReporter &errorReporter, u4 epoch, bool isIncremental,
                        std::vector<std::unique_ptr<Timer>> diagnosticLatencyTimers)
     : errorReporter(errorReporter), epoch(epoch) {
-    errorReporter.beginEpoch(epoch, move(diagnosticLatencyTimers));
+    errorReporter.beginEpoch(epoch, isIncremental, move(diagnosticLatencyTimers));
 };
 
 ErrorEpoch::~ErrorEpoch() {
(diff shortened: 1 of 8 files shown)


main/options/options.cc
```diff
@@ -320,6 +320,11 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
         "End users should prefer to use `--enable-all-beta-lsp-features`, instead.)");
     options.add_options("advanced")("enable-all-beta-lsp-features",
                                     "Enable (expected-to-be-non-crashy) early-access LSP features.");
+    options.add_options("advanced")(
+        "lsp-max-errors-reported",
```
jez, line 324: suggestion: `--lsp-error-cap`

main/options/options.cc
```diff
@@ -320,6 +320,11 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
...
+        "Limits the maximum number of simultaneous errors that LSP reports to the editor. Can prevent "
```
jez, line 325: what do you mean by "simultaneous errors"?

jvilk-stripe:

→ https://go/builds/bui_JatfIPy3QSnnyZ
→ https://go/builds/bui_JatfKXqCx6MM72

main/options/options.cc
```diff
@@ -320,6 +320,11 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
...
+        "editor slowdowns with large error lists.",
```
jez, line 326: You don't mention this here, but you do in the comment: "0 means no limit". is that the behavior? (it looks like it is from the code, but I don't see any tests)

main/lsp/ErrorReporter.cc
```diff
@@ -139,10 +172,19 @@ u4 ErrorReporter::lastDiagnosticEpochForFile(core::FileRef file) {
...
 }
 
-ErrorEpoch::ErrorEpoch(ErrorReporter &errorReporter, u4 epoch,
+void ErrorReporter::sanityCheck() const {
+    DEBUG_ONLY(u4 errorCount = 0; for (auto &status
+                                       : this->fileErrorStatuses) {
+        if (status.lastReportedEpoch >= this->lastFullTypecheckEpoch) {
+            errorCount += status.errorCount;
+        }
+    } ENFORCE(errorCount == this->clientErrorCount));
```
jez, line 181: clang-format really garbled this. also, it looks like we usually just do

```cpp
    if (!debug_mode) {
        return;
    }
```

at the top of sanityCheck methods. Do you want to reformat use that instead of `DEBUG_ONLY`?

jez (approved):
lgtm

some comments. (formatting suggestions are not blocking)

did you test the edge cases? (`=0` etc.)

main/options/options.cc
jvilk-stripe: The number of errors that the client reports at one point.

I guess removing 'simultaneous' here is still clear.

main/options/options.cc
jvilk-stripe: Yes; i should mention this here -- thanks!

(How do you test the lack of a limit? I guess I can see about generating a large number of errors.)

main/lsp/ErrorReporter.cc
jvilk-stripe: In the past I was encouraged to use `DEBUG_ONLY` for readability. I'm happy to use `if (!debug_mode)`.

main/options/options.cc
jez: The other option for zero would have been "reports no errors" so if you say 0 and it reports any errors, that's fine by me.

jvilk-stripe: ah, smart!

jvilk-stripe:
Reposting this from Stripe slack...

I manually tested this PR. It works; Sorbet reports 1000 errors at most and VS Code's UI remains responsive.
Unfortunately, Sorbet gets stuck in the typecheck pass -- likely because it's really expensive to compute nice messages for so many errors (90% of runtime is in `findMemberFuzzyMatch`).

However, this behavior is vastly preferable to VS Code hard locking up. I'll land this as-is and will look into if there's a way to either avoid generating those excess errors or make error generation faster.

PR #4137, opened 2021-04-09
T::Props::Serializable#with return T.self_type

jez:
### Test plan

See included automated tests.

PR #4242, opened 2021-06-02
Add type signature for T::Struct#serialize

andrejewski-stripe:
This will catch this and other similar bugs: `struct.serialize.transform_keys(:to_sym)`

### Motivation

Similar to https://github.com/sorbet/sorbet/pull/4137, I and others have hit a few bugs expecting T::Struct to be better typed; `struct.serialize` always returns a `T::Hash[String, T.untyped]`. 

---

> `struct.serialize` always returns a `T::Hash[String, T.untyped]`

Is this true? It should always return a hash but must the keys be strings? I follow the code: https://github.com/sorbet/sorbet/blob/ef1c143ee2c02886cdced928bda61ce88e9dc6d0/gems/sorbet-runtime/lib/types/props/serializable.rb#L42 and see the generator does String keys: https://github.com/sorbet/sorbet/blob/4f146522e25047f6527a98797672f9b1c462eced/gems/sorbet-runtime/lib/types/props/private/serializer_generator.rb#L61

I don't know if anything else is fussing this these internals though (I hope not).

### Test plan

See included automated tests.

### DIFF
--- a/rbi/sorbet/tprops.rbi
+++ b/rbi/sorbet/tprops.rbi
@@ -148,6 +148,7 @@ end
 module T::Props::Serializable
   def deserialize(hash, strict = nil); end
   def recursive_stringify_keys(obj); end
+  sig {params(strict: T.nilable(T::Boolean)).returns(T::Hash[String, T.untyped])}
   def serialize(strict = nil); end
   sig {params(changed_props: T.untyped).returns(T.self_type)}
   def with(changed_props); end


aprocter-stripe:

→ https://go/builds/bui_JbCOVHrYOul0m0
→ https://go/builds/bui_JbCOik4QO62zvN

rbi/sorbet/tprops.rbi
```diff
@@ -148,6 +148,7 @@ end
 module T::Props::Serializable
   def deserialize(hash, strict = nil); end
   def recursive_stringify_keys(obj); end
+  sig {params(strict: T.nilable(T::Boolean)).returns(T::Hash[String, T.untyped])}
```
jez, line 151: Why is this nilable? The source code defines it like this:

https://github.com/sorbet/sorbet/blob/bfae3ff4d9173587b27bb596c785f1ac08baf89b/gems/sorbet-runtime/lib/types/props/serializable.rb#L14-L18

andrejewski-stripe: I was following that code + the existing RBI which had `strict = nil`; can I just make that `strict = true` to match?

jez: > can I just make that `strict = true` to match?

yes please!

elliottt (changes requested):
Thanks for adding this signature!

This PR raised new errors on stripe's codebase https://go/builds/bui_JbCOik4QO62zvN, could you fix those before we land this PR?

marianosimone:

→ https://go/builds/bui_LbcPr9v51aJbhv
→ https://go/builds/bui_LbcPFgwXTHi48X

marianosimone:
I got interested in stale PRs, and started looking at this one. One problem with adding a signature like this is that sub-classes that override `serialize` might try to return a different type.

In Stripe's codebase, in particular, there are at least 6 cases, returning `String`, or `T::Hash[Symbol, <SomethingElse>`.

PR #4243, opened 2021-06-02
Faster and more correct Symbol::findMemberFuzzyMatchConstant

jvilk-stripe:
Speed up fuzzy constant match and fix a significant bug.
    
Fuzzy constant match called `Symbol::membersStableOrderSlow` to deterministically loop through a symbol's members. As the name implies, this method is expensive; it sorts members by their string names. When Sorbet generates many errors, this function call dominates runtime (by >80%).
    
However, this function call is not needed to ensure deterministic fuzzy matches! We can cheaply sort and tiebreak after finding the best matches in a given scope.
    
Unfortunately, I noticed that this method has some peculiar behavior:
* Rather than pick one best candidate per scope (as a comment at the top of the method says), it picks `n` semi-arbitrary best candidates within the scope that happen to beat "betterThan". This behavior creates a dependency between iteration order and the contents of `result`, and ensures that less-than-ideal matches are presented to the user.
* The function sorts the candidates to place the suggestions with the smallest edit distance first, __and then reverses the vector__ to place the largest edit distances first. As a result, Sorbet would recommend the _worst_ replacements first!
  * Look at `constant-fuzzy.out` to see how we used to recommend `Date` over `Integer` for `Intege` (!!!)
    
I have changed the behavior to:
* Pick the single best candidates in the current scope that beats out candidates from inner scopes, and sort them by symbol ID to tiebreak.
* Place the items with the smallest edit distance _first_ in the result vector without extraneous sorting.
    
From the changes in test output, I believe this will significantly improve Sorbet's autocorrect suggestions _and_ is a massive perf win when Sorbet generates many errors.

### Motivation

See https://github.com/sorbet/sorbet/pull/4239#issuecomment-853243842

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -723,8 +723,10 @@ vector<Symbol::FuzzySearchResult> Symbol::findMemberFuzzyMatchConstant(const Glo
                     // order the recommendations from this scope.
                     // We order in decreasing symbol ID order because `result` is later reversed and we want earlier
                     // ID'd symbols to be recommended first.
-                    fast_sort(scopeBest, [&](auto lhs, auto rhs) -> bool { return lhs.symbol._id > rhs.symbol._id; });
-                    for (auto item : scopeBest) {
+                    fast_sort(scopeBest, [&](const auto &lhs, const auto &rhs) -> bool {
+                        return lhs.symbol._id > rhs.symbol._id;
+                    });
+                    for (auto &item : scopeBest) {
                         result.emplace_back(item);
                     }
                 }
@@ -775,7 +777,8 @@ vector<Symbol::FuzzySearchResult> Symbol::findMemberFuzzyMatchConstant(const Glo
         // deterministic order.
         // We order in decreasing symbol ID order because `result` is later reversed and we want earlier
         // ID'd symbols to be recommended first.
-        fast_sort(globalBest, [&](auto lhs, auto rhs) -> bool { return lhs.symbol._id > rhs.symbol._id; });
+        fast_sort(globalBest,
+                  [&](const auto &lhs, const auto &rhs) -> bool { return lhs.symbol._id > rhs.symbol._id; });
         for (auto &e : globalBest) {
             result.emplace_back(e);
         }
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -700,18 +700,26 @@ vector<Symbol::FuzzySearchResult> Symbol::findMemberFuzzyMatchConstant(const Glo
                 i++;
             }
             for (const auto scope : candidateScopes) {
-                for (auto member : scope.data(gs)->membersStableOrderSlow(gs)) {
+                bool newBestChosen = false;
+                for (auto member : scope.data(gs)->members()) {
                     if (member.first.kind() == NameKind::CONSTANT &&
                         member.first.dataCnst(gs)->original.kind() == NameKind::UTF8 && member.second.exists()) {
                         auto thisDistance = Levenstein::distance(
                             currentName, member.first.dataCnst(gs)->original.dataUtf8(gs)->utf8, best.distance);
-                        if (thisDistance <= best.distance) {
+                        // NOTE: Iteration order over members is nondeterministic, so we use SymbolId as a consistent
+                        // tiebreaker. With this tiebreaker, the 'best' symbol from this scope is deterministic.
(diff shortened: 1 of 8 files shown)


core/Symbols.cc
```diff
@@ -700,29 +701,40 @@ vector<Symbol::FuzzySearchResult> Symbol::findMemberFuzzyMatchConstant(const Glo
...
+                            scopeBest.emplace_back(best);
                         }
                     }
                 }
+                if (!scopeBest.empty()) {
+                    // NOTE: Iteration order over members is nondeterministic, so we use SymbolId to deterministically
+                    // order the recommendations from this scope.
+                    // We order in decreasing symbol ID order because `result` is later reversed and we want earlier
+                    // ID'd symbols to be recommended first.
+                    fast_sort(scopeBest, [&](auto lhs, auto rhs) -> bool { return lhs.symbol._id > rhs.symbol._id; });
```
froydnj, line 726: I don't have the rules for `auto` deduction in lambda lists memorized, but I'm fairly sure that here it resolves to `Symbol::FuzzySearchResult` and we're therefore copy-constructing two items for every comparison.  Even though that copy construction is cheap, I don't think we want to do that.  Can we please use `const auto &` for `lhs` and `rhs` here?

(I realize this is just following the existing practice for the previously-existing sort in this function, but I think it was wrong there, too.)

core/Symbols.cc
```diff
@@ -758,10 +770,19 @@ vector<Symbol::FuzzySearchResult> Symbol::findMemberFuzzyMatchConstant(const Glo
                 }
             }
         }
+        // globalBest is nondeterministically ordered since iteration over `members()` is non deterministic.
+        // Everything in globalBest has the same edit distance, so we just have to sort by symbol ID to get a
+        // deterministic order.
+        // We order in decreasing symbol ID order because `result` is later reversed and we want earlier
+        // ID'd symbols to be recommended first.
+        fast_sort(globalBest, [&](auto lhs, auto rhs) -> bool { return lhs.symbol._id > rhs.symbol._id; });
```
froydnj, line 778: Same ask for `const auto &` on `lhs`/`rhs` here.

jvilk-stripe:

→ https://go/builds/bui_JbBiG8WZg46uQR
→ https://go/builds/bui_JbBibe9ICny0KB

test/testdata/lsp/missing_typed_sigil.A.rbedited
```diff
@@ -12,7 +12,7 @@ class Hello
 end
 
 def main
-  puts Gem.new
+  puts Hello.new
   #    ^^^^ error: Unable to resolve constant `Helo`
-  #    ^^^^ apply-code-action: [A] Replace with `Gem`
+  #    ^^^^ apply-code-action: [A] Replace with `Hello`
```
jvilk-stripe, line 17: :chefskiss:

core/Symbols.cc
```diff
@@ -700,29 +701,40 @@ vector<Symbol::FuzzySearchResult> Symbol::findMemberFuzzyMatchConstant(const Glo
...
                 }
+                if (!scopeBest.empty()) {
+                    // NOTE: Iteration order over members is nondeterministic, so we use SymbolId to deterministically
+                    // order the recommendations from this scope.
+                    // We order in decreasing symbol ID order because `result` is later reversed and we want earlier
+                    // ID'd symbols to be recommended first.
+                    fast_sort(scopeBest, [&](auto lhs, auto rhs) -> bool { return lhs.symbol._id > rhs.symbol._id; });
+                    for (auto item : scopeBest) {
+                        result.emplace_back(item);
+                    }
```
froydnj, line 729: I don't know whether the compiler would generate the same code here as with `auto &item`, but I think the latter is less confusing to reason about.  (Note that `auto &` is used when iterating over `globalBest`, below, so it doesn't seem completely unreasonable to use `auto &` here.)  Would you please use `auto &item` instead?

froydnj (approved):
Thank you for the comments in `findMemberFuzzyMatchConstant`!  They made things a lot easier to follow.

jvilk-stripe:
I tested this change on Stripe's codebase. While tens of thousands of errors still slows down typechecking, it does complete in a finite amount of time and interrupting it to fix the issue seems to be reasonably responsive. The errors print out to the CLI significantly faster too.

PR #4244, opened 2021-06-02
.ruby-version: 2.6.3 -> 2.7.2

jez:
### Motivation

We've been using Ruby 2.7.2 at Stripe for a while now.
No reason to stay on 2.6.3.

### Test plan

CI


aprocter-stripe (approved):
Onwards and upwards!

jez:
CI is broken (need to either switch to running ruby via bazel or install 2.7.2 into the docker container used by CI)

^ is the full extent of the context. there's no further use keeping this change open.

PR #4245, opened 2021-06-02
place block statements after all synthesized statements in struct rewriting

### LINKED ISSUE #4166: Private access modifier inside of Struct definition breaks instantiation
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AFoo%20%3D%20Struct.new(%3Afoo)%20do%0A%20%20private%0Aend%0A%0AFoo.new%0A)

```ruby
# typed: true

Foo = Struct.new(:foo) do
  private
end

Foo.new
```

#### Observed output

```
editor.rb:7: Non-private call to private method initialize on Foo https://srb.help/7031
     7 |Foo.new
        ^^^^^^^
    editor.rb:3: Defined in Foo here
     3 |Foo = Struct.new(:foo) do
     4 |  private
     5 |end
Errors: 1
```

#### Expected behavior

No error, because MRI doesn't complain about a private method call:

```
irb(main):001:0> Foo = Struct.new(:foo) do
irb(main):002:1*   private
irb(main):003:1> end
=> Foo
irb(main):004:0> 
irb(main):005:0> Foo.new
=> #<struct Foo foo=nil>
```

froydnj:
Fixes #4166.

### Motivation

Correctness is a good thing.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Struct.cc
+++ b/rewriter/Struct.cc
@@ -122,6 +122,10 @@ vector<ast::ExpressionPtr> Struct::run(core::MutableContext ctx, ast::Assign *as
                             std::move(typeMember)));
     }
 
+    body.emplace_back(ast::MK::SigVoid(loc, std::move(sigArgs)));
+    body.emplace_back(ast::MK::SyntheticMethod(loc, loc, core::Names::initialize(), std::move(newArgs),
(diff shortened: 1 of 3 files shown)


jez (approved):
Thanks!

PR #4246, opened 2021-06-03
use <Magic>.<rewriter-raise-unimplemented> for raising in unimplemented methods

froydnj:
### Motivation

Having a dedicated function to raise in unimplemented methods makes it easier to identify such raises in plugins analyzing the CFG.  You could pattern-match on the current `T.unsafe(Kernel).raise(...)` and verify that it appears in a `rewriterSynthesized` method, or somesuch, but doing things this way is much more convenient.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -601,7 +601,7 @@ void GlobalState::initEmpty() {
                  .repeatedUntypedArg(Names::arg())
                  .buildWithResultUntyped();
 
-    // Synthesize <Magic>.<raise-unimplemented>() => T.untyped
+    // Synthesize <Magic>.<rewriter-raise-unimplemented>() => T.untyped
     method =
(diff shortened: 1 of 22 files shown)


elliottt (approved):
I think this makes sense!

core/GlobalState.cc
```diff
@@ -601,6 +601,10 @@ void GlobalState::initEmpty() {
                  .repeatedUntypedArg(Names::arg())
                  .buildWithResultUntyped();
 
+    // Synthesize <Magic>.<raise-unimplemented>() => T.untyped
```
jez, line 604:
```suggestion
    // Synthesize <Magic>.<rewriter-raise-unimplemented>() => T.untyped
```

froydnj:
Trevor pointed out that this probably isn't going to work as expected: we would be emitting static errors, but it's entirely possible that methods that would `raise` actually get redefined (e.g. the prop machinery would dynamically `define_method`s).

PR #4248, opened 2021-06-04
add a rewriter error for dynamic constant assignment inside blocks

### LINKED ISSUE #1334: Dynamic constant assignment allowed in blocks
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20%5B%5D.each%20do%0A%20%20%20%20X%20%3D%2010%0A%20%20end%0Aend)

```ruby
# typed: true
class A
  [].each do
    X = 10
  end
end
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

This should raise an error, as the dynamic constant assignment is not valid.

- - -

mojanjz:
### Motivation
Resolves #1334

Adding a boolean flag within the SymbolFinder class to check wether a constant assignment is called within a block or not. 
If the assignment is within a block, it throws a dynamic constant error.
### Test plan

See included automated tests.


namer/namer.cc
```diff
@@ -347,6 +347,7 @@ class SymbolFinder {
     // `private` with no arguments toggles the visibility of all methods below in the class def.
     // This tracks those as they appear.
     vector<optional<Modifier>> methodVisiStack = {nullopt};
+    bool inBlock = false;
```
jez, line 350: Look at the comment on line 339 in this file:

> Does not report any errors, which lets us cache its output.

You cannot add errors in this phase of Sorbet.

I would encourage you to look into doing this in rewriter/Flatten.cc or a new, dedicated `rewriter/` pass.

jez (changes requested):

test/testdata/namer/dynamic_constant.rb
```diff
@@ -5,4 +5,12 @@ def self.foo
     X = 10
 #   ^ error: dynamic constant assignment
```
mojanjz, line 6: We noticed that some errors are capitalized and some are lower case as seen in [this](https://github.com/sorbet/sorbet/blob/master/third_party/parser/codegen/generate_diagnostics.cc#L49) file.  If there is a preferred way of doing this let me know.

namer/namer.cc
mojanjz: good catch thank you!

test/testdata/namer/dynamic_constant.rb
jez: For some reason, all parser errors are lower case. I think this is an artifact of porting the whitequark parser to C++.

I don't know whether or not we can change those easily, because our whitequark parser test harness is set up to be able to automatically import and test existing whitequark tests.

All other errors should be uppercase.

It would be nice to add an ENFORCE here to maintain that invariant:

https://github.com/sorbet/sorbet/blob/df5101e2e955679a9e28b6f9bee6625d38fc81f9/core/Error.h#L102-L103

namer/namer.cc
```diff
@@ -694,6 +697,13 @@ class SymbolFinder {
             return tree;
         }
 
+        if (inBlock) {
+            if (auto e = ctx.state.beginError(core::Loc(ctx.file, lhs->loc),
+                                              core::errors::Namer::DynamicConstantAssignment)) {
+                e.setHeader("Dynamic constant assignment");
```
jez, line 703: Let's make this error say something like

```
Assigning constant `{}` inside a block
```

(This is more clear what the cause of the problem is and includes the name of the offending constant, which helps people locate it more quickly without having to read line numbers.)

and potentially use `addErrorNote` to suggest people move the constant assignment outside of the block.

mojanjz:
see comment in original issue.

PR #4250, opened 2021-06-08
Fix else block conditional in builder walk

vinistock:
This conditional branch is never reached because the condition is identical to the previous one. This PR updates the conditional to check against the right block.

### Test plan

Existing tests will cover this.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -257,7 +257,7 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                 if (thenEnd != cctx.inWhat.deadBlock() || elseEnd != cctx.inWhat.deadBlock()) {
                     if (thenEnd == cctx.inWhat.deadBlock()) {
                         ret = elseEnd;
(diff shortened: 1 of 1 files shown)


cfg/builder/builder_walk.cc
```diff
@@ -257,8 +257,6 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                 if (thenEnd != cctx.inWhat.deadBlock() || elseEnd != cctx.inWhat.deadBlock()) {
                     if (thenEnd == cctx.inWhat.deadBlock()) {
                         ret = elseEnd;
-                    } else if (thenEnd == cctx.inWhat.deadBlock()) {
-                        ret = thenEnd;
```
froydnj, line 261: I think it is more correct to change this to test `elseEnd` -- both for symmetry and for not introducing superfluous blocks in the case that `elseEnd` is dead.  That change passes locally for me as well, which is kind of unfortunate.

froydnj (changes requested):
Thanks for spotting this.

froydnj (approved):
Thanks!

PR #4251, opened 2021-06-08
Introduce chained `sig` syntax

vinistock:
Make `abstract` a parameter of the sig method and disallow explicit usage of `DeclBuilder#abstract` inside sig blocks.

I tried to reuse as much as possible to avoid having to rewrite the entire abstract logic.

### Motivation

Having the `abstract` mode move into the sig parameters makes the information available sooner, since sig blocks are evaluated lazily.

With this information at an earlier stage, it's possible to completely turn off method wrapping and unwrapping for non-abstract methods, when using `default_checked_level = :never`.

### Test plan

The existing tests should mostly cover all scenarios.


resolver/type_syntax.cc
```diff
@@ -312,6 +316,21 @@ ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend
...
+
+                        auto loc = core::Loc(ctx.file, sigSend.loc);
+                        auto original = loc.source(ctx).value();
+                        auto corrected = original.replace(0, 3, "sig(:abstract)");
+                        auto abstractBegin = original.find("abstract.");
+                        corrected = corrected.erase(abstractBegin, 9);
+                        abstractBegin = original.find(".abstract");
+                        corrected = corrected.erase(abstractBegin, 9);
+
+                        e.replaceWith(fmt::format("Change `{}` to `{}`", original, corrected), loc, "{}", corrected);
```
jez, line 331: we should not make this an error in the same PR

we should first have a version that allows both syntaxes, and then have a version that has the autocorrect (two separate PRs + published rubygems versions)

resolver/type_syntax.cc
```diff
@@ -426,11 +426,7 @@ ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend
                 case core::Names::onFailure().rawId():
                     break;
                 case core::Names::final_().rawId():
-                    if (auto e = ctx.beginError(send->loc, core::errors::Resolver::InvalidMethodSignature)) {
-                        reportedInvalidMethod = true;
-                        e.setHeader("The syntax for declaring a method final is `sig(:final) {{...}}`, not `sig "
-                                    "{{final. ...}}`");
-                    }
```
jez, line 433: Can we do something that prevents people from doing

```ruby
sig {final.void}
```

because we definitely want to prevent that.

gems/sorbet-runtime/lib/sorbet-runtime.rb
```diff
@@ -24,6 +24,8 @@ module T::Private::Types; end
 require_relative 'types/private/class_utils'
 require_relative 'types/private/runtime_levels'
 require_relative 'types/private/methods/_methods'
+require_relative 'types/private/methods/modes'
+require_relative 'types/private/methods/decl_builder'
```
jez, line 28: ooc, why do these have to move earlier?

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
```diff
@@ -127,15 +128,27 @@ def abstract
...
-      raise BuilderError.new("The syntax for declaring a method final is `sig(:final) {...}`, not `sig {final. ...}`")
+
+      active_declaration = T::Private::DeclState.current.active_declaration
+
+      if blk
+        active_declaration.blk = blk
+        active_declaration.final = true
+      end
+
+      self
```
jez, line 148: similar to the static error below, i think we should still have a runtime error if someone attempts to do

```
sig {final.void}
```

and only allow `sig.final {void}` in sorbet-runtime.

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
```diff
@@ -6,6 +6,7 @@ module T::Private::Methods
 
   class DeclBuilder
     attr_reader :decl
+    attr_writer :raw
```
jez, line 9: ooc, what is this `raw` parameter? i don't quite see where it's used.

jez: also: is it intentional that this is only `attr_writer`?

rewriter/rewriter.cc
```diff
@@ -185,6 +186,11 @@ class Rewriterer {
             return expr;
         }
 
+        if (auto expr = ChainedSig::run(ctx, send)) {
+            SigRewriter::run(ctx, ast::cast_tree<ast::Send>(expr));
+            return expr;
+        }
+
```
jez, line 193: I think this transformation happens to work by accident right now. I think that what we really want is to translate the new syntax to the old syntax as early as possible, namely: before any other rewriter passes fire.

The treemap where you've written this pass interleaves all the rewriter passes at the same time in a single traversal over the tree. That means that depending on a somewhat arbitrary ordering, portions of other rewriter passes could fire before this `ChainedSig` pass runs, which could then observe the new syntax instead of the old syntax.

What do you think about structuring the `ChainedSig` pass similar to the `Cleanup` pass, which manages it's own treemap, and then running `ChainedSig::run` before the `TreeMap::apply` on these lines?

https://github.com/sorbet/sorbet/blob/a9fb984be106118b7f5b757c29f09b70993e8792/rewriter/rewriter.cc#L202-L203
There are some rewriter passes that attempt to parse sigs. If this `ChainedSig` pass ran

jez: (That way, all the rewriter passes see a single, unified view of sigs when then attempt to parse them, and don't need to see multiple syntaxes. It also gives us a convenient leverage point to implement an autocorrect in the future.)

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "ast/Helpers.h"
+#include "ast/treemap/treemap.h"
+
+using namespace std;
+namespace sorbet::rewriter {
+
+// Rewrite the new chained syntax into the old syntax. E.g.:
+// sig.abstract { void } is rewritten as sig { abstract.void }
+// Only tries to rewrite if the current send function is one of the accepted methods
+// and the receiver is sig.
```
jez, line 11: Can you move this comment to the header file, and structure it similar to the header comments on other rewriter passes?

For example

https://github.com/sorbet/sorbet/blob/a9fb984be106118b7f5b757c29f09b70993e8792/rewriter/Prop.h#L7-L25

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,40 @@
+#include "rewriter/ChainedSig.h"
+#include "ast/Helpers.h"
+#include "ast/treemap/treemap.h"
+
+using namespace std;
+namespace sorbet::rewriter {
```
jez, line 6: This file doesn't introduce any tests. When you add tests, can you be sure to add rewrite-tree exp file tests? Ideally mentioning all of the builder methods that could show up.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,40 @@
...
+    ast::ExpressionPtr body =
+        ast::MK::Send(send->loc, std::move(blockBody->recv), send->fun, send->numPosArgs, std::move(send->args));
+
+    auto newBody =
+        ast::MK::Send(send->loc, std::move(body), blockBody->fun, blockBody->numPosArgs, std::move(blockBody->args));
+    ast::Send::ARGS_store args;
+
+    return ast::MK::Send(send->loc, std::move(sigSend->recv), core::Names::sig(), 0, std::move(args), {},
+                         ast::MK::Block0(send->block.loc(), std::move(newBody)));
+}
```
jez, line 39: We don't have to (and indeed shouldn't) write an autocorrect that lands synchronously with this PR (because there is value in having a version of Sorbet that people can use to migrate to the new syntax), but would you mind prototyping such an autocorrect?

I strongly suspect it will be possible, but I'd love to confirm empirically that it's easy to change this `ChainedSig` rewriter pass to emit an error + autocorrect enforcing usage of the new syntax, so that we don't have to perpetually maintain two syntaxes.

jez:
> With this information at an earlier stage, it's possible to completely turn off method wrapping and unwrapping for non-abstract methods, when using `default_checked_level = :never`.

You are not actually taking advantage of this fact yet, right?

gems/sorbet-runtime/lib/sorbet-runtime.rb
vinistock: We need the modes and `DeclBuilder` to exist when we invoke `sig` now, since it returns a builder to permit chaining.

rewriter/ChainedSig.cc
vinistock: Done! I was waiting for initial feedback before putting some tests together.

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
vinistock: I think this was a bad rebase or something. We didn't need that line!

rewriter/rewriter.cc
vinistock: Moved it to be an initial pass on the ast, like you mentioned.

vinistock:
@jez indeed this PR doesn't turn off wrapping. I guess we will prototype it on our implementation and then we can discuss upstreaming it.

I applied the changes based on the feedback, but haven't prototyped auto-correct just yet, although I'm sure we'd be able to do it.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,88 @@
...
+    ast::ExpressionPtr postTransformSend(core::MutableContext ctx, ast::ExpressionPtr tree) {
+        auto *send = ast::cast_tree<ast::Send>(tree);
+
+        // When we receive a `sig` send, we don't know ahead of time of whether it's a chained sig
+        // or an incomplete `sig` without a block. Save it for later, so that we can add errors
+        // to sigs missing block parameters
+        if (send->fun == core::Names::sig() && send->block == nullptr) {
+            incompleteSigs.push_back(send);
+            return tree;
+        }
```
vinistock, line 21: Let me know what you folks think about this approach. I wanted to emit the error of missing block in cases where people do this
```ruby
sig # no block!
def foo; end
```
But I need a way of avoiding emitting this error for chained sigs. For example, in this snippet
```ruby
sig.abstract { void }
```
the walker is invoked both with `sig` alone and with `sig.abstract` as send nodes. In these cases, it's not incorrect to not pass a block to the `sig` method, so we don't want to add errors.

vinistock:
I'm having a bit of trouble prototyping an auto-correct. Not because we can't change one syntax into the other, but because there are so many usages of the `sig` that simply adding an empty auto-correct already exceeds the limit number of bytes for stderr.
```cpp
if (auto e = ctx.beginError(send->loc, core::errors::Infer::BlockNotPassed)) {
  e.setHeader("This signature uses the old syntax. Please use the new syntax, wink wink");

  string replacement = "sig blah";
  e.addAutocorrect(core::AutocorrectSuggestion{
    fmt::format("Replace `{}` with `{}`", send->toStringWithTabs(ctx), replacement),
    {core::AutocorrectSuggestion::Edit{core::Loc(ctx.file, send->loc), replacement}}});
}
```
Adding this small snippet for old sigs stops me from being able to compile with the following error
```
INFO: Analyzed target //main:sorbet (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /Users/viniciusstock/src/github.com/vinistock/sorbet/payload/binary/BUILD:71:8: Executing genrule //payload/binary:generate-state-payload-raw failed: (Exit 1): bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped)
stderr (/private/var/tmp/_bazel_viniciusstock/716dd1681d03852ffa76d62c35d3668f/execroot/com_stripe_ruby_typer/bazel-out/_tmp/actions/stderr-47) exceeds maximum size of --experimental_ui_max_stdouterr_bytes=1048576 bytes; skipping
Target //main:sorbet failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 43.656s, Critical Path: 30.06s
INFO: 58 processes: 13 internal, 45 local.
FAILED: Build did NOT complete successfully
```
Suggestions on how to proceed?

froydnj:
> I'm having a bit of trouble prototyping an auto-correct. Not because we can't change one syntax into the other, but because there are so many usages of the `sig` that simply adding an empty auto-correct already exceeds the limit number of bytes for stderr.
>
> ```
> INFO: Analyzed target //main:sorbet (0 packages loaded, 0 targets configured).
> INFO: Found 1 target...
> ERROR: /Users/viniciusstock/src/github.com/vinistock/sorbet/payload/binary/BUILD:71:8: Executing genrule //payload/binary:generate-state-payload-raw failed: (Exit 1): bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped)
> stderr (/private/var/tmp/_bazel_viniciusstock/716dd1681d03852ffa76d62c35d3668f/execroot/com_stripe_ruby_typer/bazel-out/_tmp/actions/stderr-47) exceeds maximum size of --experimental_ui_max_stdouterr_bytes=1048576 bytes; skipping
> Target //main:sorbet failed to build
> Use --verbose_failures to see the command lines of failed build steps.
> INFO: Elapsed time: 43.656s, Critical Path: 30.06s
> INFO: 58 processes: 13 internal, 45 local.
> FAILED: Build did NOT complete successfully
> ```
> 
> Suggestions on how to proceed?

Two thoughts:

1. Bump up the relevant Bazel option mentioned in the stderr output in `.bazelrc`.  (You might have to make this eye-wateringly high, because you're probably generating autocorrects for all the sigs in all the rbis sorbet comes with.)
2. Disable the autocorrect when generating the payload, e.g. when the `--store-state` option is given.

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
```diff
@@ -167,6 +198,10 @@ def overridable
         raise BuilderError.new(".overridable cannot be repeated in a single signature")
       end
 
+      if blk
+        T::Private::DeclState.current.active_declaration.blk = blk
+      end
+
       self
     end
```
jez, line 207: You are missing cases in ~all the other DeclBuilder methods that prevent those builders from used outside the block. For example, `returns`, `void`, `params`, `checked`, `on_failure`, etc. should all be banned outside the block.

There should be tests for all of these case as well.

jez: It would also be nice to have better static errors for misusing these builders outside the block as well. For example:

#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Arequire_relative%20'.%2Flib%2Fsorbet-runtime'%0A%0Aextend%20T%3A%3ASig%0A%0Asig.void%0Adef%20foo%0A%20%20puts%20'in%20foo'%0Aend%0A%0Afoo)

```ruby
# typed: strict

require_relative './lib/sorbet-runtime'

extend T::Sig

sig.void
def foo
  puts 'in foo'
end

foo
```

#### Observed output

```
❯ sorbet qux.rb
qux.rb:7: sig requires a block parameter, but no block was passed https://srb.help/7021
     7 |sig.void
        ^^^

qux.rb:8: This function does not have a sig https://srb.help/7017
     8 |def foo
        ^^^^^^^
    qux.rb:8: Insert sig {returns(NilClass)}

     8 |def foo
        ^

qux.rb:7: Method void does not exist on NilClass https://srb.help/7003
     7 |sig.void
        ^^^^^^^^
    qux.rb:3: Possibly uninitialized (NilClass) in:
     3 |require_relative './lib/sorbet-runtime'
     4 |
     5 |extend T::Sig
     6 |
     7 |sig.void
     8 |def foo
     9 |  puts 'in foo'
    10 |end
    11 |
    12 |foo
  Did you mean:
    https://github.com/sorbet/sorbet/tree/master/rbi/core/kernel.rbi#L1560: Kernel#load
    1560 |  def load(filename, arg0=T.unsafe(nil)); end
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 3
```

#### Expected behavior

None of the errors reported above are particularly helpful in telling the user what they did wrong (putting the builder method chained on the sig, rather than inside the block).

I think we should go out of our way to make good error messages here, because `sig` syntax is one of the first things that trips up new users of Sorbet.

(Sorbet already does not report errors here, but arguably not reporting errors was ~fine before because it was unlikely for people to attempt to chain methods like this. Now that some of the builders are chained outside and some inside, it's more likely that people will trip up.)

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
```diff
@@ -115,7 +124,7 @@ def on_failure(*args)
...
 
-    def abstract
+    def abstract(&blk)
       check_live!
 
       case decl.mode
```
jez, line 130: It looks like sorbet-runtime already implements the correct behavior as a result of the chosen implementation, but can we add tests that ensure that the "cannot be repeated in a single signature" errors are correctly reported when repeating `abstract` (and all other builders) once outside and once inside the block?

For example

```ruby
# typed: strict

require_relative './lib/sorbet-runtime'

class Parent
  extend T::Sig
  extend T::Helpers

  abstract!

  sig.abstract {abstract.void}
  def foo; end
end

class Child < Parent
  sig.override {void}
  def foo
    puts 'Child#foo'
  end
end

Child.new.foo
```

(this correctly raises right now, but there are no tests for this)

jez (changes requested):
Ask not just, "What do I have to do to make this work?" but rather "What do I have to do to make sure that this can't be misused?"

Think very creatively. Search for as many possible ways to misuse the feature you're building, and make it so that it can't be misused. I found a bunch of things by thinking this way, but I would encourage you to also do the same, and maybe in the process you'll find things that I didn't catch on my own.

I've done a first pass only looking at the sorbet-runtime implementation. Given what I've said above, I'd love if you take another pass over the Sorbet implementation to see if there's anything you'd like to change about it before I complete the review of the PR by reading the sorbet-static implementation.

If you have any questions, please lmk.

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
```diff
@@ -30,6 +31,14 @@ def initialize(mod, raw)
         ARG_NOT_PROVIDED, # type_parameters
         raw
       )
+      @inside_sig_block = false
+    end
+
+    def compile!(&block)
+      @inside_sig_block = true
```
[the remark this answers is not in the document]
vinistock: I do think it makes sense, but the implementation of that check conflicts with some edge case tests where running a sig block raises once and then passes the second time. For those cases, we would immediately raise when beginning to run the block the second time.

I'm not sure of the context of this functionality for sigs that fail once and then pass, so not sure if we can add these new checks.

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
```diff
@@ -167,6 +198,10 @@ def overridable
         raise BuilderError.new(".overridable cannot be repeated in a single signature")
       end
 
+      if blk
+        T::Private::DeclState.current.active_declaration.blk = blk
```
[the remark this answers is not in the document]
vinistock: I think the naming here is a bit unfortunate. `active_declaration` is actually a `DeclarationBlock` whereas `decl` is a `Declaration`.

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
vinistock: It's a bit hard to implement these checks in the runtime, because all of these methods are allowed in type aliases, which would be considered outside of the sig block. Sorbet itself uses this often.

```ruby
MyType = T.type_alias { params().... }
```

Would it be okay to catch these only using static checks?

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,213 @@
...
+        for (ast::Send *sig : incompleteSigs) {
+            if (auto e = ctx.beginError(sig->loc, core::errors::Rewriter::InvalidChainedSig)) {
+                e.setHeader("Signature declarations expect a block");
+                e.addErrorNote("Complete the signature by adding a block declaration: sig `{}`", "{ ... }");
+            }
+        }
+    }
+
+    void checkDuplicates(core::MutableContext ctx, ast::Send *sigBlock, core::NameRef fun) {
+        auto expr = sigBlock->deepCopy();
```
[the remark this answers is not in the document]
vinistock: This one wasn't necessary, I removed it.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,213 @@
...
+
+            return true;
+        }
+
+        return false;
+    }
+
+    ast::ExpressionPtr buildReplacement(core::MutableContext ctx, ast::Send *send, ast::Send *blockBody) {
+        ast::Send::ARGS_store args;
+        auto treeCopy = send->deepCopy();
```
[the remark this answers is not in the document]
vinistock: I think this one is necessary because we're updating the pointer in the loops and we need to keep the original reference for later.

test/testdata/rewriter/chained_sig.rb
```diff
@@ -0,0 +1,186 @@
...
+               # ^^^^^^^^ error: Duplicate invocation of `abstract` in signature declaration
+  def baz; end
+end
+
+class DuplicateButMissingBlock
+  extend T::Sig
+  include Interface
+
+  sig.final { void }.override
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Cannot add more signature statements after the declaration block
```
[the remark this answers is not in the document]
vinistock: It wasn't, but I just added an error for it.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,213 @@
+#include "rewriter/ChainedSig.h"
+#include "ast/Helpers.h"
+#include "ast/treemap/treemap.h"
+#include "core/errors/rewriter.h"
+
+using namespace std;
+namespace sorbet::rewriter {
+
+struct ChainedSigWalk {
+    vector<ast::Send *> incompleteSigs;
```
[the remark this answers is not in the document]
vinistock: I explored this and indeed the implementation is much easier using `preTransformSend`. I still needed an auxiliary variable, to prevent adding incorrect errors to `sig` without a block, but I don't think it's a big deal.

Let me know what you think!

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,184 @@
...
+
+using namespace std;
+namespace sorbet::rewriter {
+
+struct ChainedSigWalk {
+    // The previousSigSend is used to allow us to add errors to `sig` with missing block wihtout accidentally adding it
+    // to a chained sig. E.g.:
+    // `sig` -> error no block
+    // `sig.final {}` -> no error. Need to prevent adding the error when we receive `sig` by itself
+    ast::Send *previousSigSend;
```
jez, line 14: Do you mind resetting this to `nullptr` in `postTransformSend`?

That way, an invariant is that if we're currently transforming a sig send chain, it's set, but if we're transforming a completely unrelated (chain) of sigs, it's `nullptr`.

jez: hm, actually on second thought, it's not quite clear to me that this is straightforward to do, so maybe it's not worth doing. i'll think about it some more.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,184 @@
...
+    bool invalidChainedStatement(core::MutableContext ctx, ast::Send *send) {
+        if (send->fun == core::Names::params() || send->fun == core::Names::returns() ||
+            send->fun == core::Names::void_() || send->fun == core::Names::bind() ||
+            send->fun == core::Names::checked() || send->fun == core::Names::onFailure() ||
+            send->fun == core::Names::typeParameters()) {
+            auto receiver = ast::cast_tree<ast::Send>(send->recv);
+
+            while (receiver && receiver->fun != core::Names::sig()) {
+                receiver = ast::cast_tree<ast::Send>(receiver->recv);
+            }
```
jez, line 112: we're going to run this loop for literally every method call in the whole codebase

what do you think about first factoring out a check in `preTransformSend` that asks whether the current send has `sig` as one of its receivers, and early exiting from `preTransformSend` if not?

jez: it looks like you already have such a helper: `firstReceiverIsSig`. Can we just call that before calling invalidChainedStatement, and delete the loop from this method?

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,184 @@
...
+            send->fun != core::Names::override_() && send->fun != core::Names::overridable()) {
+            return tree;
+        }
+
+        ast::Send *sigSend = ast::cast_tree<ast::Send>(send->recv);
+
+        // Make sure the first receiver is sig. E.g.: `sig.abstract {}`, `sig.override.final {}`
+        if (sigSend == nullptr || !firstReceiverIsSig(sigSend)) {
+            return tree;
+        }
```
jez, line 44: (from comment below):

concretely, can we move this stanza up to be before invalidChainedStatement, and delete the while loop from invalidChainedStatement?

jez:

→ https://go/builds/bui_KbQdFEtJcvZaAJ
→ https://go/builds/bui_KbQdmn8dzQozZV

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,184 @@
...
+
+    ast::ExpressionPtr buildReplacement(core::MutableContext ctx, ast::Send *send, ast::Send *blockBody) {
+        ast::Send::ARGS_store args;
+        auto treeCopy = send->deepCopy();
+        auto sendCopy = ast::cast_tree<ast::Send>(treeCopy);
+
+        // Go through each part of the chained sig and make sure it's not duplicated inside the block
+        // E.g.: `sig.abstract { abstract.void }`
+        while (sendCopy && sendCopy->fun != core::Names::sig()) {
+            checkDuplicates(ctx, blockBody, sendCopy->fun);
```
jez, line 136: ooc, do you think we can clearly separate the "checking improper use" from the "mutating the tree" phase?

mixing those two modes of operation has been one of the most common sources of bugs in sorbet over the years.

jez: ah, it looks like checking and mutating is just really intertwined below. maybe not worth it.

jez (commented):
thanks! this looks good. i have one small suggestion for how to refactor some things in the main preTransformSend method

i've tested this, and it passes on Stripe's codebase, so at least it doesn't regress anything.

I also tested performance on Stripe's codebase, and this change doesn't seem to substantially impact the cold-cache performance of sorbet (the changes in this PR affect the cold cache path most of the time, because its result will be cached and reused if the file doesn't change)

so i think that this change is basically good to go. is there anything else you want to do before merging this?

rewriter/ChainedSig.cc
vinistock: Done, got rid of the loop.

rewriter/ChainedSig.cc
vinistock: My initial approach was exactly that, first checking all invalid usage and then trying to build a replacement, but we end up having a lot more loops, so I changed it.

Out of curiosity, what kinds of bugs have been caused by this pattern?

vinistock:
I think this is good to go as well. After this goes in, we will test shutting down all signature blocks other than the ones for abstract methods. Then we can iterate on this and improve it if necessary.

Thank you for all the effort on helping us getting this ready.

rewriter/ChainedSig.cc
jez: Usually it works like this:

- someone mutates the tree (usually to `std::move(...)` something out of it) early in the method than needed
- for some reason, we decide to return early
- the logic to compute the replacement for the thing that was `move`'d didn't fire

jez (approved):

jez:
There's a failing test case in the the revert PR:

https://github.com/sorbet/sorbet/pull/4973#issuecomment-1028485497

PR #4030, opened 2021-02-28
Test sorbet-runtime on 2.7.2

jez:
### Motivation

Pre-work for #4019

### Test plan

CI-only change.

PR #4252, opened 2021-06-09
Change unwrap_nilable to allow for larger unions

jez:
### Motivation

Reported as a bug internally at Stripe: http://go/Ugw2yoeo

cc @jnovak-stripe 

### Test plan

Verified that this works on Stripe's codebase.

http://go/builds/bui_JdplkLGHwx1j31

### Rollout plan

At least one person was attempting to workaround this bug in a way that caused failures after this bug was fixed.

Also, some generated, checked-in files change in response to this change. This means when we upgrade in Stripe's codebase, we will have to bump the version and make changes in lock step (instead of fixing the errors ahead of time, and bumping after.

No one else is making changes to sorbet-runtime right now, so this should not be disruptive.

### DIFF
--- a/gems/sorbet-runtime/lib/types/utils.rb
+++ b/gems/sorbet-runtime/lib/types/utils.rb
@@ -79,10 +79,11 @@ module T::Utils
     case type
     when T::Types::Union
       non_nil_types = type.types.reject {|t| t == Nilable::NIL_TYPE}
-      if non_nil_types.length == 1
-        non_nil_types.first
+      case non_nil_types.length
+      when 0 then nil
+      when 1 then non_nil_types.first
       else
-        nil
+        T::Types::Union::Private::Pool.union_of_types(non_nil_types[0], non_nil_types[1], non_nil_types[2..])
       end
     else
       nil
--- a/gems/sorbet-runtime/lib/types/utils.rb
+++ b/gems/sorbet-runtime/lib/types/utils.rb
@@ -79,7 +79,7 @@ module T::Utils
     case type
     when T::Types::Union
       non_nil_types = type.types.reject {|t| t == Nilable::NIL_TYPE}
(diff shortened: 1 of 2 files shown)


aprocter-stripe (approved):
LGTM!

gems/sorbet-runtime/lib/types/utils.rb
```diff
@@ -79,10 +79,12 @@ def self.unwrap_nilable(type)
...
       non_nil_types = type.types.reject {|t| t == Nilable::NIL_TYPE}
-      if non_nil_types.length == 1
-        non_nil_types.first
+      return nil if type.types.length == non_nil_types.length
+      case non_nil_types.length
+      when 0 then nil
+      when 1 then non_nil_types.first
       else
-        nil
+        T::Types::Union::Private::Pool.union_of_types(non_nil_types[0], non_nil_types[1], non_nil_types[2..])
```
Morriar, line 87: @jez this will break for Ruby <2.5

jez: Thanks. We only support 2.6 and above

https://github.com/sorbet/sorbet/pull/4030/files#diff-9cf20d6cd4e2fe34e9089339cf947f3c6c455d58b2a75e5d84f843b83fcbd239L255-R257

Morriar: I read this as untested, not explicitly unsupported.

We can change this and still keep compatibility with Ruby 2.5: https://github.com/sorbet/sorbet/pull/4261

dduugg: @jez consider updating the gemspecs (filed as https://github.com/sorbet/sorbet/issues/4265 before i thought to look here)

PR #4254, opened 2021-06-09
Update pattern matching syntax

Morriar:
### Motivation

Port changes to pattern matching syntax from https://github.com/whitequark/parser/pull/777.

### Test plan

See included automated tests.


ast/desugar/Desugar.cc
```diff
@@ -454,6 +454,21 @@ void desugarPatternMatchingVars(InsSeq::STATS_store &vars, DesugarContext dctx,
     }
 }
 
+// Desugar `in` and `=>` online pattern matching
```
elliottt, line 457:
```suggestion
// Desugar `in` and `=>` oneline pattern matching
```

ast/desugar/Desugar.cc
```diff
@@ -454,6 +454,21 @@ void desugarPatternMatchingVars(InsSeq::STATS_store &vars, DesugarContext dctx,
...
+    auto bodyExpr = MK::RaiseUnimplemented(loc);
+    auto elseExpr = MK::EmptyTree();
+
+    InsSeq::STATS_store vars;
+    desugarPatternMatchingVars(vars, dctx, match);
+    if (!vars.empty()) {
+        bodyExpr = MK::InsSeq(match->loc, std::move(vars), std::move(bodyExpr));
+    }
+
+    return MK::If(loc, std::move(matchExpr), std::move(bodyExpr), std::move(elseExpr));
```
elliottt, line 469: Is this emitting an `if` where the condition is `MK::RaiseUnimplemented` because we don't have a way of saying whether the pattern matched yet?

elliottt (changes requested):
I think this looks good!

I had one question about the intent behind using `MK::RaiseUnimplemented` as the condition of an `if`, but am happy to approve after that.

ast/desugar/Desugar.cc
Morriar: The idea was to allow these constructs to parse while doing as little as possible until the feature stabilizes upstream.

Once we have a better idea of what we want to do with pattern matching regarding typing, we can implement all these calls.

elliottt (approved):

PR #4260, opened 2021-06-10
add the ability to declare final methods directly for C extensions

froydnj:
### Motivation

C extensions can co-exist gracefully with `sorbet-runtime`, but they may occasionally want to declare `final` methods and there's no way to do that short of setting up `sig`s and such.  (We assume that the C extension will be performing its own typechecking where necessary.)

This PR provides a way for them to do this: they can call private methods to indicate that any methods being defined in the associated block should be considered `final` methods.  `sorbet-runtime` will record them as such, and normal definitions continue once the block is done executing.

Of course, as this lives under `T::Private`, details are subject to change, etc. etc.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -26,6 +26,8 @@ module T::Private::Methods
   # in installed_hooks.
   @old_hooks = nil
 
+  @declaring_final_method = false
+
   ARG_NOT_PROVIDED = Object.new
   PROC_TYPE = Object.new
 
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -180,6 +182,16 @@ def self.note_module_deals_with_final(mod)
     @modules_with_final[mod.singleton_class]
   end
 
+  def self._with_declaring_final_method(&blk)
```
jez, line 185: Do we want the interface to be a block? or to be forced to repeat something for the next method?

like `_make_next_method_final`

jez, line 185: Can we pick a name that makes it a little more clear that this is a private API? like `_INTERNAL` or something? Can we also add a comment describing expected usage?

jez (approved):

PR #4261, opened 2021-06-10
Do not use endless range in sorbet-runtime

Morriar:
### Motivation

Remove usage of end-less range so sorbet-runtime still work with 2.5.

The end-less range was introduced [here](https://github.com/sorbet/sorbet/pull/4252/files#diff-3701c6bc454263c98078696834aa766aa480276e0498b92b5eefb374047ec413R87) and can be rewritten with an explicit end.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/utils.rb
+++ b/gems/sorbet-runtime/lib/types/utils.rb
@@ -84,7 +84,8 @@ module T::Utils
       when 0 then nil
       when 1 then non_nil_types.first
       else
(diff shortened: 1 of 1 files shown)


Morriar:
See discussion here: https://github.com/sorbet/sorbet/pull/4252#discussion_r649491081

jez (approved):

Morriar:
Actually, support for 2.5 ended 2 months ago. Let's drop this and explicitly break so people are forced to update 👍

PR #4264, opened 2021-06-11
do bulk erasing in extractClassInit

froydnj:
### Motivation

We shouldn't erase single elements at a time from `klass->rhs` in this function; it's more efficient to let `std::remove_if` shift everything around in linear time and erase everything in one go.

### Test plan

This should be equivalent to the existing code, so existing tests should suffice.

### DIFF
--- a/class_flatten/class_flatten.cc
+++ b/class_flatten/class_flatten.cc
@@ -35,14 +35,14 @@ bool shouldExtract(core::Context ctx, const ast::ExpressionPtr &what) {
 ast::ExpressionPtr extractClassInit(core::Context ctx, ast::ClassDef *klass) {
     ast::InsSeq::STATS_store inits;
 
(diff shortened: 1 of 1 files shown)


class_flatten/class_flatten.cc
```diff
@@ -35,14 +35,14 @@ bool shouldExtract(core::Context ctx, const ast::ExpressionPtr &what) {
...
+        bool b = shouldExtract(ctx, expr);
+        if (b) {
+            inits.emplace_back(std::move(expr));
         }
-        inits.emplace_back(std::move(*it));
-        it = klass->rhs.erase(it);
-    }
+        return b;
+    });
+    klass->rhs.erase(it, klass->rhs.end());
```
aprocter-stripe, line 45: Took me a sec to convince myself this is logically equivalent, but I think I buy it. :)

aprocter-stripe (approved):

PR #4267, opened 2021-06-15
Add a sig for Array#collect!

elliottt:
There was no signature for `Array#collect!`. Luckily `Array#map!` is an alias to it, so their signatures will be the same.

### Motivation

More signatures for core methods.

### Test plan

See included automated tests.


elliottt:

→ https://go/builds/bui_JfoU9SiBEydd5e
→ https://go/builds/bui_JfoUGViPyBs2wf

jez (approved):

PR #4269, opened 2021-06-15
Update Builder to not emit truncated nodes of heredocs

vinistock:
Implement the grammar change https://github.com/whitequark/parser/commit/b87613d2b99733e1400f076675daf501c19ede93. Do not emit truncated nodes from heredocs.

We were able to move `test_parse_dedenting_heredoc_13.rb` from disabled to enabled after the fixes.

### Motivation

Another step towards Ruby 2.7 full compatibility.

### Test plan

See included automated tests.

cc @Morriar


froydnj (approved):
Thanks!

PR #4271, opened 2021-06-16
add and sort autogen subclasses in-place

froydnj:
### Motivation

Less memory allocation and copying data around is a good thing.

Bikeshedding welcome on `fast_sort_range` -- I know ranges are a separate thing than iterators, so maybe we want to call this `fast_sort_iterators`?  But that just sounds weird, so I went with `fast_sort_range`.

### Test plan

See included automated tests.  We have `test/cli/autogen-subclasses` for exercising this code.

### DIFF
--- a/main/autogen/subclasses.cc
+++ b/main/autogen/subclasses.cc
@@ -112,19 +112,17 @@ vector<string> Subclasses::serializeSubclassMap(const Subclasses::Map &descendan
         auto type = children.classKind == ClassKind::Class ? "class" : "module";
         descendantsMapSerialized.emplace_back(fmt::format("{} {}", type, parentName));
 
-        vector<string> serializedChildren;
+        auto subclassesStart = descendantsMapSerialized.size();
         for (const auto &[name, type] : children.entries) {
             // Ignore Modules
             if (type == autogen::Definition::Type::Class) {
-                serializedChildren.emplace_back(fmt::format(" class {}", name));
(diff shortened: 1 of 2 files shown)


main/autogen/subclasses.cc
```diff
@@ -112,19 +112,17 @@ vector<string> Subclasses::serializeSubclassMap(const Subclasses::Map &descendan
         auto type = children.classKind == ClassKind::Class ? "class" : "module";
         descendantsMapSerialized.emplace_back(fmt::format("{} {}", type, parentName));
 
-        vector<string> serializedChildren;
+        auto subclassesStart = descendantsMapSerialized.size();
```
elliottt, line 115: Is it worth adding `descendantsMapSerialized.resize(descendantsMapSerialized.size() + children.entries.size())` here as well?

elliottt (approved):
Nice find!

main/autogen/subclasses.cc
froydnj: My sense is no: the `reserve`-up-front strategy makes sure that you don't do a bunch of copying over the course of the entire loop.  We can't pre-reserve all the space here, so one way or the other, we have to eat the cost of copying in the loop body itself.  The current way accepts that cost and is shorter.  (Presumably we're not going to resize multiple times during this inner loop...although I guess we might in the early stages of finding subclasses.)

PR #4273, opened 2021-06-16
avoid accessing uninitialized memory in genDescendantsMap

froydnj:
(Why, yes, browser, I did mean hitting enter in the title field to submit the entire form...)

### Motivation

This fixes something relatively harmless, but also reduces the "what?!" factor when reading the code.

In the current code, we check `descendants`, an `optional<SubclassInfo>` for `Some`.  If it's `None`, we insert an empty entry in `descendantsMap`...but then we go on to access the contained `SubclassInfo` in the `optional`, which is not there.

I think this is not a problem currently because `emplace` won't modify existing entries, and we already created an empty one, so providing a reference to uninitialized memory is harmless.  Nonetheless, reading the current code is confusing, and we should fix that.

(Fun note: the tests passed with continuing regardless of whether we inserted the empty entry in `descendantsMap` or not...I left it there for now.)

### Test plan

See included automated tests.

### DIFF
--- a/main/autogen/subclasses.cc
+++ b/main/autogen/subclasses.cc
@@ -156,11 +156,11 @@ vector<string> Subclasses::genDescendantsMap(Subclasses::Map &childMap, vector<s
 
         auto descendants = Subclasses::descendantsOf(childMap, parentName);
         if (!descendants) {
+            // Initialize an empty entry associated with parentName.
             descendantsMap[parentName];
-            continue;
+        } else {
+            descendantsMap.emplace(parentName, std::move(*descendants));
         }
-
-        descendantsMap.emplace(parentName, std::move(*descendants));
     }
 
     return Subclasses::serializeSubclassMap(descendantsMap, parentNames);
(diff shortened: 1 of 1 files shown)


main/autogen/subclasses.cc
```diff
@@ -158,6 +158,7 @@ vector<string> Subclasses::genDescendantsMap(Subclasses::Map &childMap, vector<s
         auto descendants = Subclasses::descendantsOf(childMap, parentName);
         if (!descendants) {
             descendantsMap[parentName];
+            continue;
         }
 
         descendantsMap.emplace(parentName, *descendants);
```
aprocter-stripe, line 164: Two suggestions, both optional:

1. Add a comment explaining that the `descendantsMap[parentName];` statement is (I think???) there for its side effect of creating a value for the key `parentName` if it doesn't already exist. (I always find this behavior of `operator[]` confusing.)
2. Instead of `continue`, move the `emplace` into an `else`.

aprocter-stripe (approved):
LGTM! Just some optional suggestions.

PR #4274, opened 2021-06-16
permit ||= and &&= on pinned variables in loops

### LINKED ISSUE #4262: Desguaring of &&= leaks pinning error
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Breturns(T%3A%3ABoolean)%7D%0Adef%20returns_t_boolean%0A%20%20false%0Aend%0A%0Asig%20%7Bparams(x%3A%20T%3A%3ABoolean).void%7D%0Adef%20using_and_and(x)%0A%20%20result%20%3D%20x%0A%20%201.times%20do%0A%20%20%20%20result%20%26%26%3D%20returns_t_boolean%0A%20%20end%0Aend%0A%0Asig%20%7Bparams(x%3A%20T%3A%3ABoolean).void%7D%0Adef%20desugared(x)%0A%20%20result%20%3D%20x%0A%20%201.times%20do%0A%20%20%20%20if%20result%0A%20%20%20%20%20%20result%20%3D%20returns_t_boolean%0A%20%20%20%20else%0A%20%20%20%20%20%20result%0A%20%20%20%20end%0A%20%20end%0Aend)

```ruby
# typed: strict
extend T::Sig

sig {returns(T::Boolean)}
def returns_t_boolean
  false
end

sig {params(x: T::Boolean).void}
def using_and_and(x)
  result = x
  1.times do
    result &&= returns_t_boolean
  end
end

sig {params(x: T::Boolean).void}
def desugared(x)
  result = x
  1.times do
    if result
      result = returns_t_boolean
    else
      result
    end
  end
end
```

#### Observed output

```
editor.rb:13: Changing the type of a variable in a loop is not permitted https://srb.help/7001
    13 |    result &&= returns_t_boolean
                       ^^^^^^^^^^^^^^^^^
  Existing variable has type: TrueClass
  Attempting to change type to: T::Boolean

  Original type from:
    editor.rb:10:
    10 |def using_and_and(x)
                          ^
    editor.rb:13:
    13 |    result &&= returns_t_boolean
            ^^^^^^

editor.rb:22: Changing the type of a variable in a loop is not permitted https://srb.help/7001
    22 |      result = returns_t_boolean
                       ^^^^^^^^^^^^^^^^^
  Existing variable has type: TrueClass
  Attempting to change type to: T::Boolean

  Original type from:
    editor.rb:18:
    18 |def desugared(x)
                      ^
    editor.rb:21:
    21 |    if result
               ^^^^^^
Errors: 2
```

#### Expected behavior

Both errors are confusing at least and undesirable at most.

We should either make the error message better, or make the error message go away.

- - -

The `desugared` method approximates the `&&=` desugaring, which reveals why there is an error in the first place.

froydnj:
Fixes #4262 

### Motivation

Typecheck reasonable code.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_JgNzI02UddCeeK
→ https://go/builds/bui_JgNznNHJpbOXM1

ast/desugar/Desugar.cc
```diff
@@ -617,6 +617,17 @@ class DuplicateHashKeyCheck {
...
 
+core::NameRef freshNameFromRef(DesugarContext dctx, const ExpressionPtr &ref) {
+    if (auto *nm = cast_tree<UnresolvedIdent>(ref)) {
+        return dctx.ctx.state.freshNameUnique(core::UniqueNameKind::Desugar, nm->name, ++dctx.uniqueCounter);
+    }
+    if (auto *nm = cast_tree<ast::Local>(ref)) {
+        return dctx.ctx.state.freshNameUnique(core::UniqueNameKind::Desugar, nm->localVariable._name,
+                                              ++dctx.uniqueCounter);
+    }
+    Exception::notImplemented();
```
jez, line 628: I generally don’t like this because half the time when it gets tripped there’s no backtraces or line numbers and it’s annoying to get one CS if there had just been a mention of the function name in the error message. So I always raise my own error with a unique looking message

test/testdata/desugar/andand_pinned.rb
```diff
@@ -0,0 +1,16 @@
+# typed: strict
+extend T::Sig
+
+sig {returns(T::Boolean)}
```
jez, line 4: Can you add a test to show that we still emit a pinning error if this method’s return type did widen the type?

And because I’m curious, would you mind another test that demonstrates what happens when `x` starts out untyped? Not sure if the behavior on that example changes in this PR but `¯\_(ツ)_/¯` maybe it’ll be a good test to have

jez (approved):

elliottt (changes requested):
This desugaring doesn't seem quite right in the presence of an operator that performs an assignment:

```ruby
class Foo
  def bar
    puts "accessing bar (#{@bar})"
    @bar
  end

  def bar=(bar)
    puts "writing bar (#{@bar} => #{bar})"
    @bar = bar
  end
end

puts "Current implementation of ||="
x = Foo.new
x.bar ||= 10
x.bar ||= 20


puts "Proposed implementation of ||="
x = Foo.new
t1 = x.bar
t2 = if t1
       t1
     else
       10
     end
x.bar = t2

t3 = x.bar
t4 = if t3
       t3
     else
       10
     end
x.bar = t4
```
Running this program produces the following:
```
% ruby test.rb
Current implementation of ||=
accessing bar ()
writing bar ( => 10)
accessing bar (10)
Proposed implementation of ||=
accessing bar ()
writing bar ( => 10)
accessing bar (10)
writing bar (10 => 10)
```
The difference for the proposed desugaring is that we see an additional call to the assignment method `Foo#bar=` at the end.

elliottt:
> This desugaring doesn't seem quite right in the presence of an operator that performs an assignment:

Actually scratch that, this would only be a problem if we were treating sends on the LHS this way, which is a different case outlined above.

elliottt (approved):

froydnj:

→ https://go/builds/bui_JgkDHVxcCAYz3a
→ https://go/builds/bui_JgkDq49qXoqrjz

froydnj:

→ https://go/builds/bui_JglAwHeBXHkCC7
→ https://go/builds/bui_JglAqZkkj4Afkw

PR #4275, opened 2021-06-17
factor out fresh name generation into DesugarContext

froydnj:
### Motivation

It's silly to have to repeat `dctx.ctx.state.freshNameUnique(...)` everywhere; we can make things shorter and more consistent.

### Test plan

See included automated tests.  The exp files change in one place because one call was using post-increment on `dctx.uniqueCounter` instead of pre-increment.


ast/desugar/Desugar.cc
```diff
@@ -2066,8 +2053,7 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
             [&](parser::CaseMatch *caseMatch) {
                 // Create a local var to store the expression used in each match clause
                 auto exprLoc = caseMatch->expr->loc;
-                auto exprName = dctx.ctx.state.freshNameUnique(core::UniqueNameKind::Desugar, core::Names::assignTemp(),
-                                                               dctx.uniqueCounter++);
```
froydnj, line 2070: This is the lone call that was different, and whose change is responsible for the exp changes.

jez (approved):
Nice 👍

Issue #1856, opened 2019-09-25
 Unsupported node type Backref


bubaflub:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20def%20foo%0A%20%20%20%20'asdf'.gsub(%2F(a)%2F)%20%7B%20%24%26%20%7D%0A%20%20end%0Aend)

```ruby
# typed: true

class A
  def foo
    'asdf'.gsub(/(a)/) { $& }
  end
end
```

#### Observed output

```
editor.rb:5: Unsupported node type Backref https://srb.help/3003
     5 |    'asdf'.gsub(/(a)/) { $& }
                                 ^^
Errors: 1
```

#### Expected behavior

I expected Sorbet to handle `$&` and typecheck it as a String.

- - -

Found this while trying to apply this an existing Ruby on Rails codebase.  I'd be happy to help implement a fix but I'll need some guidance on how to help.

jez:
https://github.com/sorbet/sorbet/blob/4bef9ab4572a5e4eef714cca9c7586de94eb051f/ast/desugar/Desugar.cc#L1669-L1672

These would be the lines to change. You'll want to transform the parse tree that we get from Ruby into an AST suitable for Sorbet. The easiest way to do this is to convert it into another existing node type. For example, if it's guaranteed that `Backref` is always a term of type `String`, you could just cheat and say something like `Backref` -> `String.new`.

If that's not always the case, then you could do something like `Backref` -> `T.unsafe(nil)`

It might be the case that other members of the team care about modeling this more carefully, which would require more work so you should come talk to us on Slack (https://sorbet.org/slack) in the #internals channel before trying to implement this.

bubaflub:
Everything I'm reading from the docs for [`Regexp`](https://ruby-doc.org/core-2.4.5/Regexp.html) and [`MatchData`](https://ruby-doc.org/core-2.4.5/MatchData.html) indicates that `$&` is a `String` and is equivalent to `Regexp.last_match[0]`.  It seems like sorbet happily type checks `Regexp.last_match[0]` so that's an okay work-around.

I'd like to try and tackle this ticket next week -- mind if I assign it to myself?

bubaflub:
Looking at https://github.com/sorbet/sorbet/blob/586a106954f17f27655455dcf6580424583d6b1e/parser/tools/generate_ast.cc#L66-L71

[...]

PR #4276, opened 2021-06-17
Add support for desugaring regex backreferences

RyanBrushett:
Desugar backref expressions as `T.let(nil, T.nilable(String))`
Resolve https://github.com/sorbet/sorbet/issues/1856

### Motivation

These come up from time to time. When they do, we're required to set a file as `typed: ignore` or refactor in order to have sorbet run happily.

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -2085,7 +2085,8 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
             },
             [&](parser::Backref *backref) {
                 auto recv = MK::Constant(loc, core::Symbols::Magic());
-                result = MK::Send0(loc, std::move(recv), core::Names::regexBackref());
+                auto arg = MK::Symbol(backref->loc, backref->name);
+                result = MK::Send1(loc, std::move(recv), core::Names::regexBackref(), std::move(arg));
             },
             [&](parser::EFlipflop *eflipflop) {
                 auto res = unsupportedNode(dctx, eflipflop);
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -2084,7 +2084,10 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                 result = std::move(res);
             },
             [&](parser::Backref *backref) {
-                auto res = unsupportedNode(dctx, backref);
+                // Desugar backref expression as T.let(nil, T.nilable(String))
+                auto value = MK::Nil(loc);
+                auto type = MK::Nilable(loc, MK::Constant(loc, core::Symbols::String()));
+                ExpressionPtr res = MK::Let(loc, std::move(value), std::move(type));
                 result = std::move(res);
             },
             [&](parser::EFlipflop *eflipflop) {
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -2084,11 +2084,8 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                 result = std::move(res);
             },
             [&](parser::Backref *backref) {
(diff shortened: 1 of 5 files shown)


ast/desugar/Desugar.cc
```diff
@@ -2084,7 +2084,10 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
                 result = std::move(res);
             },
             [&](parser::Backref *backref) {
-                auto res = unsupportedNode(dctx, backref);
+                // Desugar backref expression as T.let(nil, T.nilable(String))
+                auto value = MK::Nil(loc);
+                auto type = MK::Nilable(loc, MK::Constant(loc, core::Symbols::String()));
+                ExpressionPtr res = MK::Let(loc, std::move(value), std::move(type));
```
jez, line 2090: Using `T.let` is going to cause the variable to be pinned even when the user did not explicitly pin the variable.

Can we instead desugar this to a custom method like `<Magic>.<regex-backref>(...)`, and make that method's signature say `T.nilable(String)`?

Here's a test case for you

```ruby
# typed: true
extend T::Sig

sig {returns(T.nilable(String))}
def magic_backref; end

/foo/.match("foo") do |m|
  x = $&
  x = false

  y = magic_backref
  y = false
end
```

These two things should behave the same, but this change treats them differently. (This PR causes an error on `x = false` but no error on `y = false`).

This will also be better long term. At runtime, the value is not actually `nil`. Modeling the runtime behavior is important for us too.

jez: Look in GlobalState.cc for how to define new <Magic> methods.

froydnj (changes requested):
jez, as usual, saw things that I would not have caught; please address those concerns and resubmit.  Thanks!

ast/desugar/Desugar.cc
RyanBrushett: Thank you! Will dig back into this.

RyanBrushett:
Thanks for the feedback and the help!

Pushed up some changes based on Jez's comment. Getting warmer?

core/GlobalState.cc
```diff
@@ -504,6 +504,10 @@ void GlobalState::initEmpty() {
                  .untypedArg(Names::arg2())
                  .buildWithResult(Types::rangeOfUntyped());
 
+    // Synthesize <Magic>,<regex-backref>() => T.nilable(String)
```
jez, line 507: typo: `,` → `.`

and actually: what do you think about making this take the backref kind as an argument?

```
<Magic>.<regex-backref>(:"$&")
```

etc.

You should be able to do this pretty easily by just calling something like

```cpp
MK::Symbol(backref->loc, backref->name)
```

test/testdata/desugar/backref.rb
```diff
@@ -0,0 +1,31 @@
...
+T.reveal_type(x7) # error: Revealed type: `T.untyped`
+
+class TestNotPinningBackrefs
+  /foo/.match("foo") do |m|
+    x = $&
+    T.reveal_type(x) # error: Revealed type: `T.nilable(String)`
+    x = false
+    T.reveal_type(x) # error: Revealed type: `FalseClass`
+  end
+end
```
jez, line 31: thanks for committing this test case.

jez (approved):
lgtm with one suggested change.

froydnj (changes requested):
Just requesting changes to echo jez's changes and to move this PR off my plate in github (multiple reviewers on a PR doesn't have great UX, apparently).

RyanBrushett:
Thanks for the feedback! How's this looking?
Also: Should I rebase this into one commit or is this fine?

jez:
> Also: Should I rebase this into one commit or is this fine?

Multiple commits is always preferable, bonus points if each commit means something when viewed individually (vs just work-in-progress commits that don't mean anything in isolation).

Your change will always be squashed and merged into master as a final step (you don't need to do this yourself).

jez (approved):
thanks!

PR #4277, opened 2021-06-17
add some desugar tests for &&= and ||=

froydnj:
### Motivation

These all came up as problematic cases when debugging #4274, and it seemed reasonable to add them to the tree now so that we can be aware we would be changing behavior later.

### Test plan

See included automated tests.


elliottt (approved):
Thanks for the extra tests!

PR #4278, opened 2021-06-17
Enable requires_ancestor in the playground

Morriar:
### Motivation

Just checking how you guys feel about this 😄 

### Test plan

See included automated tests.

### DIFF
--- a/emscripten/main.cc
+++ b/emscripten/main.cc
@@ -12,7 +12,8 @@ using namespace std;
 
 extern "C" {
 void EMSCRIPTEN_KEEPALIVE typecheck(const char *rubySrc) {
-    const char *argv[] = {"sorbet", "--color=always", "--silence-dev-message", "-e", rubySrc};
+    const char *argv[] = {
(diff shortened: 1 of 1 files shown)


emscripten/main.cc
```diff
@@ -12,7 +12,8 @@ using namespace std;
 
 extern "C" {
 void EMSCRIPTEN_KEEPALIVE typecheck(const char *rubySrc) {
-    const char *argv[] = {"sorbet", "--color=always", "--silence-dev-message", "-e", rubySrc};
+    const char *argv[] = {
+        "sorbet", "--enable-experimental-requires-ancestor", "--color=always", "--silence-dev-message", "-e", rubySrc};
```
jez, line 16: Can we make it opt-in? like maybe put something in a URL param.

jez: like `https://sorbet.run?--enable-experimental-requires-ancestor`

elliottt (changes requested):
I agree with @jez here, it would be great if this could be opt-in somehow.

jez:
Closing in favor of #4732

PR #4279, opened 2021-06-18
wip: better error messages for unsupported tuple/shape types in prop types

froydnj:
### Motivation

The current behavior is that we silently drop the prop from any kind of codegen, and people get weird errors about calls to `initialize` or somesuch.  We should at least issue an error message to let them know bad things are happening at the source of the problem.

The PR thus far illustrates the basic idea, but I think the Right Thing to do is have `dupType` take a `std::optional<FailureContext> &`.  We can fill in the optional with information about the actual expression that failed and its send context (if any).  If there is a send context, we can use that to generate error messages about unsupported shape types if we have the right kind of expression, etc.  If there's not a send context, but the failing expression is the top-level expression, we can generate error messages about unsupported shape types if we have the right kind of expression.  Otherwise, we'll just issue the generic message this PR currently has.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Util.cc
+++ b/rewriter/Util.cc
@@ -8,12 +8,20 @@ using namespace std;
 
 namespace sorbet::rewriter {
 
-ast::ExpressionPtr ASTUtil::dupType(const ast::ExpressionPtr &orig) {
+ast::ExpressionPtr ASTUtil::dupType(const ast::ExpressionPtr &orig, const ast::ExpressionPtr **failing) {
+#define RECORD_FAILING(tree)                  \
+    do {                                      \
+        if (failing && *failing == nullptr) { \
+            *failing = &tree;                 \
+        }                                     \
+    } while (0)
+
     auto send = ast::cast_tree<ast::Send>(orig);
     if (send) {
         ast::Send::ARGS_store args;
-        auto dupRecv = dupType(send->recv);
+        auto dupRecv = dupType(send->recv, failing);
         if (!dupRecv) {
+            RECORD_FAILING(send->recv);
             return nullptr;
         }
         if (send->fun == core::Names::enum_()) {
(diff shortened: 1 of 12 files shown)


rewriter/Util.cc
```diff
@@ -8,12 +8,20 @@ using namespace std;
 
 namespace sorbet::rewriter {
 
-ast::ExpressionPtr ASTUtil::dupType(const ast::ExpressionPtr &orig) {
+ast::ExpressionPtr ASTUtil::dupType(const ast::ExpressionPtr &orig, const ast::ExpressionPtr **failing) {
+#define RECORD_FAILING(tree)                  \
```
jez, line 12: 😅 do we have to use a macro?

jez:
> Otherwise

Another option would be to actually parse the tuple/shape type into a tuple/shape type, and allow the behavior.

Another option (depending on how many problems there are from the current approach or ^ this approach) would be to parse the type with no error and make it be `T.untyped` until we migrate to fix the problems.

froydnj:
> > Otherwise
> 
> Another option would be to actually parse the tuple/shape type into a tuple/shape type, and allow the behavior.

Just for clarity's sake, are you saying that this is a separate option from making errors better (and not just a substitute for the "Otherwise" clause in the above)?

I don't really get the history here, especially when type aliases work just fine, but maybe the desire was to avoid spreading shape and tuple types around too much...?

> Another option (depending on how many problems there are from the current approach or ^ this approach) would be to parse the type with no error and make it be `T.untyped` until we migrate to fix the problems.

That is also a reasonable path.

froydnj:
> > Another option (depending on how many problems there are from the current approach or ^ this approach) would be to parse the type with no error and make it be `T.untyped` until we migrate to fix the problems.
> 
> That is also a reasonable path.

I'm doubting this statement after thinking for a while -- the user is expecting static type-checking of the array/hash/whatever element and Sorbet is going to silently drop type information and let everything through.  That does not seem like the best approach.

rewriter/Util.cc
froydnj: It's not a requirement, no -- certainly it seems like Sorbet style is to shy away from macros.  But I think in this case it makes it clearer what's going on and easier to visually verify that the expression that you're recording as failing is the same expression that you attempted to call `dupType` on slightly earlier.

jez: > makes it clearer what's going on and easier to visually verify

would this not have been the case had it been implemented with a function?

PR #4280, opened 2021-06-18
Fix DependencyDB display for type template owners

aisamanra:
### Motivation
Previously, in the DependencyDB produced by autogen, we would display the owners of type templates using the `<Class:...>` syntax instead of just the name of the class, so references to a type template `Foo` in a class `Opus::Bar` might have a resolved name that looked like `[:Opus, :'<Class:Bar>', :Foo]`, which is never what we want. We in fact worked around this in several places.

This uses `shortName` for this display, which will always display the original name without any extra stuff. That allows us to sidestep the problems.

This also identified and fixed a problem where we were exiting the test runner too early and not running pipeline tests.

### Test plan

See included automated tests. Added a test case for this, although the exercised code is slightly different (i.e. we don't have expectation tests around the msgpack itself, but the `toString` of the autogen info is a good proxy.)


froydnj (approved):
Thanks!

aisamanra:

→ https://go/builds/bui_JimB8nf3OKsnG5
→ https://go/builds/bui_JimBnQuos0xCd8

Issue #4265, opened 2021-06-11
Required ruby version is incorrect


dduugg:
The [required ruby version](https://github.com/sorbet/sorbet/blob/master/gems/sorbet-runtime/sorbet-runtime.gemspec#L14) of `sorbet-runtime` is listed as `>= 2.3.0`.

However, the gem uses [endless ranges](https://github.com/sorbet/sorbet/blob/master/gems/sorbet-runtime/lib/types/utils.rb#L87), which is a Ruby 2.6 [feature](https://bugs.ruby-lang.org/issues/12912), breaking on earlier ruby releases.

Could I also humbly request that the test suite be updated to run on the latest patch of each supported minor ruby release (e.g. 2.6, 2.7)? Thanks!

jez:
We document what versions of Ruby Sorbet supports here:

https://sorbet.org/docs/faq#what-platforms-does-sorbet-support

If you feel particularly strongly that we should capture that in the gemspec I can, but I tend to think that version ranges are brittle, because the true test is whether it works, not whether the versions match.

> Could I also humbly request that the test suite be updated to run on the latest patch of each supported minor ruby release (e.g. 2.6, 2.7)? Thanks!

The sorbet-runtime test suite already runs against both ruby 2.6 and ruby 2.7.

dduugg:
Thanks for the quick reply! The version range is critical, so external developers know if they're on a supported path (and internal ones know which ruby features they can use).

It doesn't have to be `required_ruby_version` though, but please remove the entry if it won't be maintained (it's an [optional](https://guides.rubygems.org/specification-reference/) part of a gemspec).

sambostock:
> The version range is critical

I agree with this. Honestly, I **personally** can't think of a reason why not to have the `required_ruby_version` reflect the versions these gems are tested against. To me, `required_ruby_version` is a commitment saying

> this gem works with the stated Ruby versions, and if it doesn't then a bug exists

As I mentioned in https://github.com/sorbet/sorbet/pull/4281#issuecomment-870939262, I think it is entirely Stripe's prerogative which Ruby versions are supported, but I think setting `required_ruby_version` to a version less than those that the gem is tested against is not an ideal choice as it can lead to the scenario discussed here: a breaking change was made for Ruby < 2.6 (the use of an endless change), but it was not caught by automated tests.

[...]

PR #4281, opened 2021-06-18
Replace an endless range with an ordinal one in sorbet-runtime

Slike9:
The fix is needed for compatibility with ruby <= 2.5, as endless ranges are introduced in ruby 2.6.

### DIFF
--- a/gems/sorbet-runtime/lib/types/utils.rb
+++ b/gems/sorbet-runtime/lib/types/utils.rb
@@ -84,7 +84,7 @@ module T::Utils
       when 0 then nil
       when 1 then non_nil_types.first
       else
-        T::Types::Union::Private::Pool.union_of_types(non_nil_types[0], non_nil_types[1], non_nil_types[2..])
+        T::Types::Union::Private::Pool.union_of_types(non_nil_types[0], non_nil_types[1], non_nil_types[2..-1])
       end
     else
       nil


Morriar:
Related: https://github.com/sorbet/sorbet/pull/4261.

Ruby 2.5 reached EOL [three months ago](https://www.ruby-lang.org/en/downloads/branches/):

> Ruby 2.5
> status: eol
> release date: 2017-12-25
> EOL date: 2021-03-31

Slike9:
> Related: #4261.
> 
> Ruby 2.5 reached EOL [three months ago](https://www.ruby-lang.org/en/downloads/branches/):
> 
> > Ruby 2.5
> > status: eol
> > release date: 2017-12-25
> > EOL date: 2021-03-31

What about reflecting this in https://github.com/sorbet/sorbet/blob/master/gems/sorbet-runtime/sorbet-runtime.gemspec#L14?

aprocter-stripe (approved):
I think that even though 2.5 is not officially supported, there's no reason not to merge this. No implied guarantee that we won't break 2.5 compatibility in new/additional ways in the future, of course. :)

(BTW, sorry for the delayed review here. Was on vacation last week.)

aprocter-stripe:

→ https://go/builds/bui_Jl0hRLfY62TUop
→ https://go/builds/bui_Jl0hmEccy9f7P8

sambostock:
👋 Coming here after opening #4324 and closing it after discovering it duplicated this.

Would it make sense for Sorbet's tests to be run against all major supported `required_ruby_version`s so it's clear when compatibility has been broken? I think it's fine for Sorbet to support whatever versions it wants, but it would be good if there was something ensuring `required_ruby_version` isn't wrong.

aprocter-stripe:

→ https://go/builds/bui_JlhCaeGIRcp5vL
→ https://go/builds/bui_JlhC3j2BnXuNhW

aprocter-stripe:
HI @sambostock. We do run tests against the supported versions (2.6 and 2.7), but it's true that the `required_ruby_version` is out of sync with this. There's a bit more context at #4265.

PR #4282, opened 2021-06-18
Add an error if a `void` type is used in a conditional

### LINKED ISSUE #2913: Emit error if void used in conditional branch
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20void%20%7D%0A%20%20def%20self.foo%3B%20end%0Aend%0A%0Aif%20A.foo%0A%20%20puts%20%22Dead%20code%22%0Aend)

```ruby
# typed: true

class A
  extend T::Sig

  sig { void }
  def self.foo; end
end

if A.foo
  puts "Dead code"
end
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

There should be an error for attempting to use `A.foo` in a conditional.
At runtime, void values are always truthy.
When people branch on void, it means they think that the type could be falsy, which it never can be.

- - -

Using `unless` will show a [dead code error](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20void%20%7D%0A%20%20def%20self.foo%3B%20end%0Aend%0A%0Aunless%20A.foo%0A%20%20puts%20%22Dead%20code%22%0Aend):

```ruby
# typed: true

class A
  extend T::Sig

  sig { void }
  def self.foo; end
end

unless A.foo
  puts "Dead code"
end
```

```
    11 |  puts "Dead code"
               ^^^^^^^^^^^
Errors: 1
```

but we would like to reliably show an error in both cases by making the error show up at the conditional itself.

vinistock:
Closes #2913

Add an error if a `void` type is used in a conditional.

Notes:
- I explored marking all condition branches as dead, but decided not to go forward with that idea because: it's better to report a single error in the conditional when `void` is used, instead of reporting the `void` error + the dead code errors
- Additionally, marking both branches of a conditional as dead marks the entire rest of the code as dead as well, which is also not a great experience
- I saved previous places where the error has been reported to avoid duplicates. When we are processing an if else statement, both the if and the else branches report the same error about the same parent conditional. Please, let me know if there's a better way of preventing these duplicates

This implementation will avoid marking the conditional branches as dead, but will report the error for using `void` in the conditional. After fixing that error, then any other dead code violations will still show up.

```ruby
sig { void }
def foo
  # something for which the return is not important
end

if foo # Can't use void types in conditional
end
```

### Motivation

Void types shouldn't be used in conditionals, since `void` indicates that they do not return anything useful.

### Test plan

See included automated tests.

### DIFF
--- a/infer/inference.cc
+++ b/infer/inference.cc
@@ -270,11 +270,13 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
             pair.second.knowledge.emitKnowledgeSizeMetric();
         }
 
-        core::TypePtr conditionalType = current.getTypeAndOrigin(ctx, bb->bexit.cond.variable).type;
+        auto conditionalType = current.getTypeAndOrigin(ctx, bb->bexit.cond.variable);
 
-        if (!conditionalType.isUntyped() && core::Types::isSubType(ctx, core::Types::void_(), conditionalType)) {
+        if (!conditionalType.type.isUntyped() && conditionalType.type != core::Types::Object() &&
+            core::Types::isSubType(ctx, core::Types::void_(), conditionalType.type)) {
             if (auto e = ctx.beginError(bb->bexit.loc, core::errors::Infer::InvalidVoidUsage)) {
                 e.setHeader("Can't use void types in conditional");
+                e.addErrorSection(conditionalType.explainGot(ctx, core::Loc(ctx.file, bb->bexit.loc)));
             }
         }
     }
--- a/infer/inference.cc
+++ b/infer/inference.cc
@@ -269,6 +269,14 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
         for (auto &pair : current.vars()) {
             pair.second.knowledge.emitKnowledgeSizeMetric();
         }
+
+        core::TypePtr conditionalType = current.getTypeAndOrigin(ctx, bb->bexit.cond.variable).type;
+
+        if (!conditionalType.isUntyped() && core::Types::isSubType(ctx, core::Types::void_(), conditionalType)) {
+            if (auto e = ctx.beginError(bb->bexit.loc, core::errors::Infer::InvalidVoidUsage)) {
+                e.setHeader("Can't use void types in conditional");
+            }
+        }
     }
     if (startErrorCount == ctx.state.totalErrors()) {
         counterInc("infer.methods_typechecked.no_errors");
--- a/infer/inference.cc
+++ b/infer/inference.cc
@@ -271,8 +271,10 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
         }
 
         auto conditionalType = current.getTypeAndOrigin(ctx, bb->bexit.cond.variable);
+        core::TypePtr basicObject = core::make_type<core::ClassType>(core::Symbols::BasicObject());
 
         if (!conditionalType.type.isUntyped() && conditionalType.type != core::Types::Object() &&
+            conditionalType.type != basicObject &&
             core::Types::isSubType(ctx, core::Types::void_(), conditionalType.type)) {
             if (auto e = ctx.beginError(bb->bexit.loc, core::errors::Infer::InvalidVoidUsage)) {
                 e.setHeader("Can't use `{}` types in conditional", "void");
--- a/infer/inference.cc
+++ b/infer/inference.cc
@@ -275,7 +275,7 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
         if (!conditionalType.type.isUntyped() && conditionalType.type != core::Types::Object() &&
             core::Types::isSubType(ctx, core::Types::void_(), conditionalType.type)) {
             if (auto e = ctx.beginError(bb->bexit.loc, core::errors::Infer::InvalidVoidUsage)) {
-                e.setHeader("Can't use void types in conditional");
+                e.setHeader("Can't use `{}` types in conditional", "void");
                 e.addErrorSection(conditionalType.explainGot(ctx, core::Loc(ctx.file, bb->bexit.loc)));
             }
         }
--- a/test/testdata/infer/void_in_conditional.rb
+++ b/test/testdata/infer/void_in_conditional.rb
@@ -1,0 +1,39 @@
+# typed: true
+
+class A
+  extend T::Sig
+
+  sig { void }
+  def self.foo; end
+end
+
+class B
+  extend T::Sig
+
+  sig { void }
+  def self.foo; end
+end
+
+class C
+  extend T::Sig
+
+  sig { void }
+  def self.foo; end
+end
+
+if A.foo
+ # ^^^^^ error: Can't use void types in conditional
+  puts "Dead code"
+end
+
+if B.foo
+ # ^^^^^ error: Can't use void types in conditional
+  puts "Dead code"
+else
+  puts "More dead code" # error: This code is unreachable
+end
+
+unless C.foo
+     # ^^^^^ error: Can't use void types in conditional
+  puts "Dead code" # error: This code is unreachable
+end
--- a/test/testdata/infer/void_in_conditional.rb
+++ b/test/testdata/infer/void_in_conditional.rb
@@ -21,6 +21,20 @@ class C
   def self.foo; end
 end
 
+class NoErrorBasicObject
+  extend T::Sig
+
+  sig { returns(BasicObject) }
+  def self.foo; end
+end
+
+class NoErrorObject
+  extend T::Sig
+
+  sig { returns(Object) }
+  def self.foo; end
+end
+
 if A.foo
  # ^^^^^ error: Can't use `void` types in conditional
   puts "Dead code"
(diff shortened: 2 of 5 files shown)


jez:
Can you share an example of what it looked like without the logic to prevent duplicate errors? Adding a Boolean to the environment seems like something to avoid if possible.

vinistock:
This is what the error looks like avoiding duplication.

Screen Shot 2021-06-18 at 1 05 17 PM

This is what it looks like without avoiding duplication. It's a literal copy of the previous error.

Screen Shot 2021-06-18 at 1 06 29 PM

I don't love the implementation either, but I'm not sure of an alternative that would prevent these.

elliottt (changes requested):
I think there's an easier version of this change: instead of adding a boolean to Knowledge and tracking it in `assumeKnowledge`, why not check the `bb->bexit.cond.variable` at the end of the main loop in `inference.cc`? I think you'll need to make sure that it's not `T.untyped`, and check that void is a subtype of it to capture what you're looking for here.

vinistock:
@elliottt That would look nicer if it works. What's the correct way to get the type based on a `LocalRef` (i.e.: `bb->bexit.cond.variable`)?

elliottt:
> @elliottt That would look nicer if it works. What's the correct way to get the type based on a `LocalRef` (i.e.: `bb->bexit.cond.variable`)?

You can use `current.getTypeAndOrigin(ctx, bb->bexit.cond.variable)`, and use the `type` field of the result.

vinistock:
Indeed, it works if we do it in the end of the loop. The only caveat is, I couldn't find a good way of preventing the dead code reports without tinkering with the knowledge class.

Let me know if you believe it's okay to keep it like this or if there's another we can remove the dead code errors when a void in conditional error occurs.

Screen Shot 2021-06-21 at 4 26 26 PM

infer/inference.cc
```diff
@@ -269,6 +269,14 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
         for (auto &pair : current.vars()) {
             pair.second.knowledge.emitKnowledgeSizeMetric();
         }
+
+        core::TypePtr conditionalType = current.getTypeAndOrigin(ctx, bb->bexit.cond.variable).type;
+
+        if (!conditionalType.isUntyped() && core::Types::isSubType(ctx, conditionalType, core::Types::void_())) {
```
jez, line 275: i think you've got this backwards (you want void < conditionalType)

This will only catch the case where the branch variable's type is void (or some subtype of void, which there are none)

you want to raise an error if any component of the type is void. here's your test case:

```ruby
extend T::Sig

sig {void}
def returns_void; end

if T.unsafe(nil)
  x = returns_void
else
  x = false
end

if x
  T.reveal_type(x)
else
  T.reveal_type(x)
end
```

vinistock: Indeed it was, although this particular example worked with the backwards version too.

jez:
> The only caveat is, I couldn't find a good way of preventing the dead code reports without tinkering with the knowledge class.

I don't see these changes. Did you not push them?

infer/inference.cc
jez: Wait really? In that case, can you look for a test that would have failed had the order not been swapped?

vinistock:
The current code only adds the error in the conditional about an invalid void usage, but it does not prevent the branches from being marked as dead.

I did not find a good way of preventing it without adding extra things to knowledge. I attempted to reach back in the previous blocks around [here](https://github.com/sorbet/sorbet/blob/master/infer/inference.cc#L146), but it did not work as I expected.

jez:
> but it does not prevent the branches from being marked as dead.

Ah got it.

For what it's worth we **should not** change that. They are dead code, and there's no reason to silence that for this change (the condition is always truthy in those cases, so the error is correct).

infer/inference.cc
```diff
@@ -269,6 +269,14 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
         for (auto &pair : current.vars()) {
             pair.second.knowledge.emitKnowledgeSizeMetric();
         }
+
+        core::TypePtr conditionalType = current.getTypeAndOrigin(ctx, bb->bexit.cond.variable).type;
+
+        if (!conditionalType.isUntyped() && core::Types::isSubType(ctx, core::Types::void_(), conditionalType)) {
+            if (auto e = ctx.beginError(bb->bexit.loc, core::errors::Infer::InvalidVoidUsage)) {
+                e.setHeader("Can't use void types in conditional");
```
jez, line 277: Can we make this error message a little better by using:

```cpp
e.addErrorSection(conditionalType.explainGot(gs, originForUninitialized));
```

?

infer/inference.cc
jez: > Wait really? In that case, can you look for a test that would have failed had the order not been swapped?

Here is a better test case:

```ruby
# typed: true

class D
  extend T::Sig
  sig {void}
  def returns_void; end

  def test
    if T.unsafe(nil)
      x = returns_void
    else
      x = false
    end

    puts 'defeat simiplification'

    if x
      T.reveal_type(x)
    else
      T.reveal_type(x)
    end
  end
end
```

as i wrote it before it happened to work because Sorbet simplified the CFG into an equivalent program that had slightly different behavior

jez: (this test does not report an error but should have had the order of `conditionalType` and `void` been the other way around)

infer/inference.cc
vinistock: That looks better, I added the section.

Some tests were giving false positives when the type of an argument was `Object`, which seemed to match the condition of having `void` as a subtype. I added a check to prevent that if the type is object.

Here's what the output looks like now.

Screen Shot 2021-06-22 at 9 14 57 AM

jez: > Some tests were giving false positives when the type of an argument was `Object`, which seemed to match the condition of having `void` as a subtype. I added a check to prevent that if the type is object.

Interesting. I wonder if this means that we should make void be a subtype of `<top>` rather than `Object`.

> Here's what the output looks like now.

🤔 i'm surprised to see it say `NilClass` instead of `Void`.

infer/inference.cc
```diff
@@ -269,6 +269,16 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
         for (auto &pair : current.vars()) {
             pair.second.knowledge.emitKnowledgeSizeMetric();
         }
+
+        auto conditionalType = current.getTypeAndOrigin(ctx, bb->bexit.cond.variable);
+
+        if (!conditionalType.type.isUntyped() && conditionalType.type != core::Types::Object() &&
+            core::Types::isSubType(ctx, core::Types::void_(), conditionalType.type)) {
+            if (auto e = ctx.beginError(bb->bexit.loc, core::errors::Infer::InvalidVoidUsage)) {
+                e.setHeader("Can't use void types in conditional");
```
jez, line 278:
```suggestion
                e.setHeader("Can't use `{}` types in conditional", "void");
```

test/testdata/infer/void_in_conditional.rb
```diff
@@ -0,0 +1,39 @@
...
+
+class C
+  extend T::Sig
+
+  sig { void }
+  def self.foo; end
+end
+
+if A.foo
+ # ^^^^^ error: Can't use void types in conditional
```
jez, line 25:
```suggestion
 # ^^^^^ error: Can't use `void` types in conditional
```

test/testdata/infer/void_in_conditional.rb
```diff
@@ -0,0 +1,39 @@
...
+  puts "Dead code"
+end
+
+if B.foo
+ # ^^^^^ error: Can't use void types in conditional
```
jez, line 30:
```suggestion
 # ^^^^^ error: Can't use `void` types in conditional
```

test/testdata/infer/void_in_conditional.rb
```diff
@@ -0,0 +1,39 @@
...
+  puts "Dead code"
+else
+  puts "More dead code" # error: This code is unreachable
+end
+
+unless C.foo
+     # ^^^^^ error: Can't use void types in conditional
```
jez, line 37:
```suggestion
     # ^^^^^ error: Can't use `void` types in conditional
```

jez (approved):
I think we should change how void works in a follow up PR to fix the special case with `Object`, which is a pretty gnarly hack, in a different way.

For now, thanks for your work on this! I think we're good to go.

froydnj:

→ https://go/builds/bui_JifOvNj04ETcDP
→ https://go/builds/bui_JifOOTEacOUwzA

infer/inference.cc
vinistock: I was surprised by that too. Could it be because the method body is empty (thus returning `nil`)?

vinistock: Actually, I just tested this and it also gives us false positives for `BasicObject`. I can add another check for that for the time being.

vinistock: Added a new check for `BasicObject`, which now passes in our code base.

vinistock:
Closing this as we'll probably need more discussions to achieve a more correct solution.

PR #4284, opened 2021-06-18
Fix escaping new lines inside heredocs

vinistock:
Fix escaping new lines inside heredocs. Our behavior was slightly different than [whitequark's implementation](https://github.com/whitequark/parser/blob/7ecddb67b2f82df92e424d36b6642cb887148cbc/lib/parser/lexer/dedenter.rb#L36) for squiggly heredocs and other literals.

### Test plan

See included automated tests.

### DIFF
--- a/test/whitequark/test_slash_slash_n_escaping_in_literals_0.rb.parse-tree-whitequark.exp
+++ b/test/whitequark/test_slash_slash_n_escaping_in_literals_0.rb.parse-tree-whitequark.exp
@@ -1,4 +1,3 @@
 s(:dstr,
-  s(:str, "a\
-"),
+  s(:str, "a"),
   s(:str, "b"))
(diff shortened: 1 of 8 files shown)


test/whitequark/test_slash_slash_n_escaping_in_literals_0.rb.parse-tree-whitequark.exp
```diff
@@ -1,4 +1,3 @@
 s(:dstr,
-  s(:str, "a\
-"),
+  s(:str, "a"),
```
vinistock, line 2: This is now closer to whitequark.

```ruby
a = "a\
b"

Parser::CurrentRuby.parse(a)
=> s(:str, "ab") # closer to what we have now
```

jez:

→ https://go/builds/bui_JiKeKXW3WiNpOv
→ https://go/builds/bui_JiKeADzMNQxHrY

jez (approved):

PR #4285, opened 2021-06-18
Fix an integer overflow bug with large strings

elliottt:
Inserting large strings (size greater than 65535) into the string table was triggering an overflow bug with the variable we use to track the offset into the last page (`stringsLastPageUsed`). The solution is to not update the `stringsLastPageUsed` variable by the size of the string inserted when the string is larger than a page of the strings table, and instead:
* when the table was not empty before the large string was inserted, swap the new entry and previous last entry, leaving the intent of `stringsLastPageUsed` intact
* when the table was empty before the large string was inserted, add a new empty page at the end of the table and reset the value of `stringsLastPgaeUsed` to zero

### Motivation

Fixing string table corruption when large strings are present.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1124,6 +1124,14 @@ string_view GlobalState::enterString(string_view nm) {
         if (strings.size() > 1) {
             // last page wasn't full, keep it in the end
             swap(*(strings.end() - 1), *(strings.end() - 2));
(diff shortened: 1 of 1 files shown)


froydnj (approved):
WFM.  Thanks for debugging this.

PR #4288, opened 2021-06-21
Remove some unnecessary type cases in calls.cc

elliottt:
There were some cases in `Array` intrinsics for `TupleType`, which don't seem to be possible to hit. Reviewing with whitespace changes turned off makes the changes much easier to see.

### Motivation

Removing dead code and simplifying calls.cc a bit in the process.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2796,15 +2796,10 @@ public:
     void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
         // Unwrap the array one time to get the element type (we'll rewrap it down at the bottom)
         TypePtr element;
-        if (auto *ap = cast_type<AppliedType>(args.thisType)) {
-            ENFORCE(ap->klass == Symbols::Array() || ap->klass.data(gs)->derivesFrom(gs, Symbols::Array()));
-            ENFORCE(!ap->targs.empty());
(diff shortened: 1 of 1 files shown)


elliottt:

→ https://go/builds/bui_JiGzFngDKuuamd
→ https://go/builds/bui_JiGzYbDtWhnSHV

elliottt:

→ https://go/builds/bui_JiHGynIDY3mPsN
→ https://go/builds/bui_JiHGQZqBHHudvY

froydnj (approved):
I am willing to believe this -- we have testcases that call at least `zip`, `flatten`, and `compact` on tuples as the receiver -- but what is the mechanism by which those tuples turn into arrays to pass through these particular intrinsics?

elliottt:
> I am willing to believe this -- we have testcases that call at least `zip`, `flatten`, and `compact` on tuples as the receiver -- but what is the mechanism by which those tuples turn into arrays to pass through these particular intrinsics?

https://github.com/sorbet/sorbet/blob/86e89e68/core/types/calls.cc#L126-L138

When dispatching a call on a `TupleType`, it falls back on dispatching on the `underlying()` type if the method doesn't exist on `TupleType` directly. A result of this is that the Array intrinsics mentioned in this PR will only ever see Array receivers.

PR #4291, opened 2021-06-22
preallocate and avoid copying in reserveFiles

froydnj:
### Motivation

++speed.  This change is good for ~10% of "reserveFiles" time on Stripe's codebase -- which is only ~12ms, but hey, a win is a win.

### Test plan

See included automated tests.


jez (approved):
&👍

PR #4292, opened 2021-06-22
do less copying of shared_ptrs during resolve

froydnj:
### Motivation

Taking value types is all well and good, but doing needless reference counting while we're resolving a bunch of stuff is not so good.  Taking these pointers by `const &` ensures that we're not doing any more reference counting than we need.

### Test plan

See included automated tests.


elliottt (approved):
🚀

PR #4293, opened 2021-06-22
pack ResolutionItem more tightly

froydnj:
### Motivation

Taking up less space with one of resolver's data structures seems like a good thing.  This change combined with #4292 takes ~2% off the resolving time (~50ms) on Stripe's codebase.

### Test plan

See included automated tests.


elliottt (approved):
Nice find!

PR #4294, opened 2021-06-22
do less copying around parsed sigs in resolver

froydnj:
### Motivation

Make things go faster.

Each commit in this series is self-contained and can be reviewed independently.  Everything except the last one should be fairly uncontroversial, but the last one is possible a matter of taste (but is also the largest speedup).  The whole series speeds up "resolver.resolve_sigs" by ~10% (~40ms).

`fillInInfoFromSig` currently takes `ParsedSig` by value, because `fillInInfoFromSig` is going to modify the `argTypes` field:

https://github.com/sorbet/sorbet/blob/3502cd361dea95e33cf10a6127815cab8e3a0997/resolver/resolver.cc#L2596-L2603

for keeping track of signature/definition mismatches.  You can obviously modify fields of a taken-by-value struct.

You would think that we could just take `ParsedSig` by reference -- after all, the parsed sig only exists during sig resolution and is intended to be a temporary data structure.  But `resolve{Multi,}SignatureJob` take their `job` arguments by constant reference and therefore the `job.sig` field can only be taken by constant reference, barring `const_cast` tricks.  So to take `ParsedSig&`, we have to take non-const references to the jobs themselves.

I think having to modify `resolve{Multi,}SignatureJob` to take non-constant references is mostly OK, but reasonable people can differ on this.

One alternative would be for `resolve{Multi,}SignatureJob` to take their `job` arguments by _rvalue_ reference -- this would at least make it clearer to the reader and clearer to any future modifier of this code that, at the callsite to those functions, `job` is essentially going away.  And then `fillInInfoFromSig` could take `ParsedSig&&`, also indicating at the callsite that `sig` is going away.  (Note that `std::move(...)`'ing something for an rvalue reference arg _doesn't_ mean that you're actually copying anything.  You're just indicating to the compiler, "tell my callee that they can treat this as a temporary value."  Nothing would change at the actual machine code level versus an lvalue reference.)

This alternative is maybe overly clever, which is why I didn't write the code this way.  Thoughts?

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2859,7 +2859,7 @@ public:
             } else {
                 overloadSym = mdef.symbol;
             }
-            fillInInfoFromSig(ctx, overloadSym, sig.loc, move(sig.sig), isOverloaded, mdef);
+            fillInInfoFromSig(ctx, overloadSym, sig.loc, sig.sig, isOverloaded, mdef);
         }
         handleAbstractMethod(ctx, mdef);
     }
@@ -2867,7 +2867,7 @@ public:
         prodCounterInc("types.sig.count");
         auto &mdef = *job.mdef;
         bool isOverloaded = false;
-        fillInInfoFromSig(ctx, mdef.symbol, job.loc, move(job.sig), isOverloaded, mdef);
+        fillInInfoFromSig(ctx, mdef.symbol, job.loc, job.sig, isOverloaded, mdef);
         handleAbstractMethod(ctx, mdef);
     }
 
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2562,10 +2562,7 @@ private:
         }
 
         // Get the parameters order from the signature
-        vector<ParsedSig::ArgSpec> sigParams;
-        for (auto &spec : sig.argTypes) {
-            sigParams.push_back(spec);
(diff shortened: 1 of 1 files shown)


jez (approved):
Thanks for the description. I'm not a huge fan of making the ResolveSignatureJob be mutable (this phase of resolver is parallel, and it's generally easier to not rely on mutation in those cases).

I haven't looked but: are you sure that it's safe to do these const -> mutable changes given how resolver chooses to parallelize this phase?

froydnj:
> Thanks for the description. I'm not a huge fan of making the ResolveSignatureJob be mutable (this phase of resolver is parallel, and it's generally easier to not rely on mutation in those cases).
> 
> I haven't looked but: are you sure that it's safe to do these const -> mutable changes given how resolver chooses to parallelize this phase?

`resolve_sigs` happens entirely serially:

https://github.com/sorbet/sorbet/blob/3502cd361dea95e33cf10a6127815cab8e3a0997/resolver/resolver.cc#L3001-L3020

I suppose it's possible that could change in the future...?  Even then, I think the worry wouldn't be the modification of `job`s, since that data wouldn't be shared between jobs, but whatever else resolving sigs does (modifying symbols and such)...?

PR #2439, opened 2020-01-09
Signatures for DidYouMean::SpellChecker

ob-stripe:
Add signatures for `DidYouMean::SpellChecker`.

Definitions:
- `initialize`: https://github.com/ruby/did_you_mean/blob/0da01bcdcc94d5e975d030a6ce60e4cd0a3d7293/lib/did_you_mean/spell_checker.rb#L8
- `correct`: https://github.com/ruby/did_you_mean/blob/0da01bcdcc94d5e975d030a6ce60e4cd0a3d7293/lib/did_you_mean/spell_checker.rb#L12

Example usage: https://github.com/ruby/did_you_mean#using-the-didyoumeanspellchecker

### Motivation
Can't use `DidYouMean::SpellChecker` at the moment: https://sorbet.run/#%23%20typed%3A%20true%0A%0Adef%20main%0A%20%20spell_checker%20%3D%20DidYouMean%3A%3ASpellChecker.new(dictionary%3A%20%5B'email'%2C%20'fail'%2C%20'eval'%5D)%0A%0A%20%20spell_checker.correct('meail')%0A%20%20spell_checker.correct('afil')%0Aend%0A

### Test plan
Not sure how to write tests for this.

PR #4295, opened 2021-06-22
DidYouMean::SpellChecker.initialize: correct parameter type

bayandin:
### Motivation

Currently `DidYouMean::SpellChecker.initialize` parameter type described as `T::Array` which doesn't allow to pass `Set` to it. From a code perspective, there's should be no problems with sets (the parameter uses only with `#select` call https://github.com/ruby/did_you_mean/blob/f1573d602f9e47510f5b00eeccbad906032ebea7/lib/did_you_mean/spell_checker.rb#L16).

See https://sorbet.run/#%23%20typed%3A%20true%0A%0Adef%20main%0A%20%20items%20%3D%20Set.new%20%25w%5Bemail%20fail%20eval%5D%0A%20%20spell_checker%20%3D%20DidYouMean%3A%3ASpellChecker.new(dictionary%3A%20items)%0A%0A%20%20spell_checker.correct('meail')%0A%20%20spell_checker.correct('afil')%0Aend%0A

Initially, types for `DidYouMean` modules were added in https://github.com/sorbet/sorbet/pull/2439

### Test plan

I'm not sure what we could additionally test here

### DIFF
--- a/rbi/gems/didyoumean.rbi
+++ b/rbi/gems/didyoumean.rbi
@@ -44,7 +44,7 @@ class DidYouMean::NullChecker < Object
 end
 
 class DidYouMean::SpellChecker < Object
-  sig {params(dictionary: T::Array[T.any(String, Symbol)]).returns(DidYouMean::SpellChecker)}
+  sig {params(dictionary: T::Enumerable[T.any(String, Symbol)]).returns(DidYouMean::SpellChecker)}
   def initialize(dictionary:)
   end
 


froydnj:

→ https://go/builds/bui_JieTBDRT6LbhEk
→ https://go/builds/bui_JieTSyhScrCltp

froydnj (approved):
Thanks!

bayandin:
Thanks @froydnj  🎉

PR #4296, opened 2021-06-22
provide internal methods to separate signature building from method definition

froydnj:
### Motivation

This PR wants to do basically the same thing as #4260.  Except that #4260's approach was a) not tested well enough :dug-shame: and b) not complete enough even in the limited scope for final methods.  Specifically for final methods, just saying "hey, this thing is a final method" is not enough: final method checking also ties deeply into having signatures associated with methods.  #4260 didn't address that side of things at all.

Instead, this PR proposes that we have secret internal hooks to build declaration blocks for signatures and to activate those declaration blocks for whatever method the user wants to define.  This approach is more general than #4260, more useful (there are plenty of reasons to want signatures attached to methods, not just yes/no for final), and...well, actually works.

### Test plan

See included automated tests.  We have somehow gotten away with mostly not having tests for `T::Utils.signature_for_method`; I can add a test or two demonstrating that this approach actually sticks a signature on methods defined via this private backdoor.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -77,7 +77,7 @@ module T::Private::Methods
   end
 
   def self.start_proc
-    DeclBuilder.new(PROC_TYPE)
+    DeclBuilder.new(PROC_TYPE, false)
   end
 
   def self.finalize_proc(decl)
@@ -348,7 +348,7 @@ module T::Private::Methods
   end
 
   def self.run_builder(declaration_block)
-    builder = DeclBuilder.new(declaration_block.mod)
+    builder = DeclBuilder.new(declaration_block.mod, declaration_block.raw)
     builder
       .instance_exec(&declaration_block.blk)
       .finalize!
@@ -379,6 +379,7 @@ module T::Private::Methods
         check_level: current_declaration.checked,
         on_failure: current_declaration.on_failure,
         override_allow_incompatible: current_declaration.override_allow_incompatible,
+        defined_raw: current_declaration.raw,
       )
 
       SignatureValidation.validate(signature)
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -26,8 +26,6 @@ module T::Private::Methods
   # in installed_hooks.
   @old_hooks = nil
 
-  @declaring_final_method = false
-
   ARG_NOT_PROVIDED = Object.new
   PROC_TYPE = Object.new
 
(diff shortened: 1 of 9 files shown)


froydnj:
OK, patch series updated:

* we start with a straight-up revert of #4260 so some of the logic changes are easier to see.
* we add bits so `final` methods can work, with tests
* for the C extension methods that we're wrapping, the implementations perform their own argument validation.  So the sigs only exist to provide a consistent view across "native" Ruby methods and "foreign" C methods.  Therefore we need to ensure that the sigs are *not* checked, and the last two commits patch up the relevant bits of code.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -45,15 +56,28 @@ def self.declare_sig(mod, arg, &blk)
...
 
-    loc = caller_locations(2, 1).first
+    loc = caller_locations(caller_depth + 1, 1).first
 
-    T::Private::DeclState.current.active_declaration = DeclarationBlock.new(mod, loc, blk, arg == :final)
+    DeclarationBlock.new(mod, loc, blk, arg == :final, raw)
+  end
 
-    nil
+  def self.__with_declared_signature(mod, declblock, &blk)
```
jez, line 64: We use `_` for private-but-needs-to-be-public methods elsewhere in this file. Is it worth sticking to that convention in the handful of changes this PR makes?

jez (approved):

jez:
Thanks for all the tests!

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -26,14 +26,25 @@ module T::Private::Methods
...
-
   ARG_NOT_PROVIDED = Object.new
   PROC_TYPE = Object.new
 
-  DeclarationBlock = Struct.new(:mod, :loc, :blk, :final)
+  DeclarationBlock = Struct.new(:mod, :loc, :blk, :final, :raw)
 
   def self.declare_sig(mod, arg, &blk)
+    # caller_depth is 2: 1 to get to our caller and 1 to get to sig()'s caller.
+    T::Private::DeclState.current.active_declaration = __declare_sig_internal(mod, arg, caller_depth: 2, &blk)
```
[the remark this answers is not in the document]
froydnj: I like the idea, but existing practice in `sorbet-runtime` appears to be a `caller_depth`-style approach -- possibly because the `caller_depth`-style enables you to delay the actual location gathering until you actually need to emit an error.

I suppose that logic doesn't apply here, since errors are unlikely and getting the location "early" isn't liable to cause performance issues.  (Compare with the implementation of casts, where we expect to not emit errors -- and therefore typically will not need location information -- so we're OK with `caller_depth`-style code.)

I'll fix this up.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
froydnj: Sure, I can do that.

PR #4297, opened 2021-06-23
make UnresolvedClassType's constructor more efficient

froydnj:
### Motivation

Don't allocate when you don't need to.

### Test plan

See included automated tests.


elliottt (approved):
Nice find!

PR #4298, opened 2021-06-23
add some tests for T::Utils.signature_with_method

froydnj:
### Motivation

We had a single test for `signature_with_method`, now we have a few more.  Very basic, but better than nothing.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/test/types/utils.rb
+++ b/gems/sorbet-runtime/test/types/utils.rb
@@ -13,5 +13,29 @@ module Opus::Types::Test
         assert(unwrapped.subtype_of?(T.any(String, Float)))
       end
     end
+
+    describe 'T::Utils.signature_for_method' do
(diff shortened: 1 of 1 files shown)


jez (approved):
thanks!

gems/sorbet-runtime/test/types/utils.rb
```diff
@@ -13,5 +13,29 @@ class UtilsTest < Critic::Unit::UnitTest
...
+          extend T::Sig
+          sig {returns(Integer)}
+          def sigfun
+            85
+          end
+        end
+        sfm = T::Utils.signature_for_method(c.instance_method(:sigfun))
+        refute_nil(sfm)
+        assert_equal(:sigfun, sfm.method.name)
+        assert_equal(:sigfun, sfm.method_name)
```
jez, line 36: This is a duplicate line?

PR #4300, opened 2021-06-24
Add join definition to Set

dee-see:
### Motivation

Join is missing from the definition of `Set`. Reference: https://ruby-doc.org/stdlib-3.0.1/libdoc/set/rdoc/Set.html#method-i-join

### Test plan

Not familiar enough with the project to add the tests, but let me know where to do it if this changes requires it!

### DIFF
--- a/rbi/stdlib/set.rbi
+++ b/rbi/stdlib/set.rbi
@@ -284,6 +284,29 @@ class Set < Object
   end
   def intersection(enum); end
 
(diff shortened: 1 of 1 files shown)


elliottt (approved):
This change looks good to me.

Just as a warning: It looks like this method was added in ruby-3.0 and we haven't updated sorbet to work fully with ruby-3.0 yet.

PR #4301, opened 2021-06-24
WIP: Autogen speedups

jvilk-stripe:
I'm opening this WIP PR to test some autogen speedups. There's nothing to see here right now; once I iterate on testing these a bit I'll write up my findings.

### Motivation

Speed up autogen.

### Test plan

Covered by existing tests (hopefully?).

PR #4305, opened 2021-06-28
Autogen: Avoid sorting autoloader output, add missing timers, poll more frequently for completed work

jvilk-stripe:
Autogen: Avoid sorting autoloader output, add missing timers, and poll more frequently for results.

@nroman-stripe Feel free to reassign or ask someone else to review if appropriate!

### Motivation

Speed up autogen and improve our ability to trace where time is going. The increased polling frequency ensures that autogen remains snappy even if we make the parallel section faster than 1 second.

This is the first PR containing optimizations from:
https://github.com/sorbet/sorbet/pull/4301

### Test plan

Covered by existing tests.

### DIFF
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -271,8 +271,8 @@ void runAutogen(const core::GlobalState &gs, options::Options &opts, const autog
     autogen::DefTree root;
     AutogenResult out;
     vector<pair<int, AutogenResult::Serialized>> merged;
-    for (auto res = resultq->wait_pop_timed(out, chrono::seconds{1}, *logger); !res.done();
-         res = resultq->wait_pop_timed(out, chrono::seconds{1}, *logger)) {
+    for (auto res = resultq->wait_pop_timed(out, WorkerPool::BLOCK_INTERVAL(), *logger); !res.done();
+         res = resultq->wait_pop_timed(out, WorkerPool::BLOCK_INTERVAL(), *logger)) {
         if (!res.gotItem()) {
             continue;
         }
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -286,12 +286,15 @@ void runAutogen(const core::GlobalState &gs, options::Options &opts, const autog
         }
     }
 
-    for (auto &elem : merged) {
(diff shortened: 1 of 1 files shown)


main/realmain.cc
```diff
@@ -270,27 +271,30 @@ void runAutogen(const core::GlobalState &gs, options::Options &opts, const autog
...
-        merged.insert(merged.end(), make_move_iterator(out.prints.begin()), make_move_iterator(out.prints.end()));
+        for (auto &print : out.prints) {
+            merged[print.first] = move(print.second);
+        }
         if (opts.print.AutogenAutoloader.enabled) {
             Timer timeit(logger, "autogenAutoloaderDefTreeMerge");
             root = autogen::DefTreeBuilder::merge(gs, move(root), move(*out.defTree));
         }
     }
-    fast_sort(merged, [](const auto &lhs, const auto &rhs) -> bool { return lhs.first < rhs.first; });
```
jez, line 286: Are you sure we can drop this sort? Do we really not need it for the purposes of making our exp files consistent?

jvilk-stripe: @jez Read the PR. I am not changing sort order. :) The output comes bundled with its final index!

jvilk-stripe: Output is a pair of <index, output data>. Rather than make a vector of pairs of indices and output data and sorting by indices, I create a presized vector of output data and slot in the output into the appropriate index.

jez: Makes sense, thanks!

nroman-stripe: OOC: Did you consider handing the thread a pointer into `merged` to write to instead of writing out the idx's? Or is that worse for thread-pinning issues?

nroman-stripe (approved):
One minor question inline, but otherwise 👍

main/realmain.cc
jvilk-stripe: I think that would be worse as it would cause cache contention on the `merged` vector. Core 1 would have to load part of the vector exclusively into memory, write the output into it, and flush it to main memory before core 2 could load it up exclusively to do the same.

PR #4306, opened 2021-06-28
Autogen: Faster CRC calculations with a table.

jvilk-stripe:
Autogen: Calculate CRCs with a table. `perf` seemed convinced that CRC calculations take up ~50% of `Autogen::generate`'s time, and with this optimization it only takes up ~30%. Oddly, end-to-end time for worker threads is only mildly impacted -- so there could be some measurement error here.

### Motivation

Speed!

We can improve CRC calculations by a further 10X by using hardware instructions (see https://github.com/komrad36/CRC#option-11-1-byte-hardware-accelerated, thanks to @froydnj!), but I'll leave that to a future PR since it involves landing cargo-culted CRC code from that markdown file. It also slots nicely into this CRCBuilder abstraction.

### Test plan

Covered by existing tests.

### DIFF
--- a/main/autogen/BUILD
+++ b/main/autogen/BUILD
@@ -3,12 +3,14 @@ cc_library(
     srcs = [
         "autogen.cc",
         "autoloader.cc",
+        "crc.cc",
         "packages.cc",
         "subclasses.cc",
     ],
     hdrs = [
         "autogen.h",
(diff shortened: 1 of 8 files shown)


aisamanra (approved):
This all looks good to me!

PR #4307, opened 2021-06-28
Autogen: Shrink `Reference` by 32 bits, use cast_tree_nonnull, avoid vector copying

jvilk-stripe:
Three minor autogen optimizations. They do not have a significant impact on runtime:

* Use `LocOffsets` rather than `Loc` in `Reference`. All `Loc` within a `Reference` are from its `FileRef`.
* Use `cast_tree_nonnull` when we know a cast will not fail. It avoids a runtime check.
* Avoid vector copying in `QualifiedName` abstraction. Instead, have it consume the vector so it gets `move`d into place.

### Motivation

Reducing autogen runtime.

### Test plan

Covered by existing tests.

### DIFF
--- a/main/autogen/autogen.cc
+++ b/main/autogen/autogen.cc
@@ -37,8 +37,9 @@ class AutogenWalk {
     }
 
     // Convert a constant literal into a fully qualified name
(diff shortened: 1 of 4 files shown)


aisamanra (approved):
All this looks good to me, too!

PR #4311, opened 2021-06-28
[Ruby 3.0] Import new tests for args assocs

vinistock:
Starting the work to update the parser for Ruby 3.0. These new `args_assocs` tests are all passing without any code modifications.

Note: only edit I made was changing the argument nodes from `kwargs` to `hash`, since that's the node type we are currently emitting.

### Test plan

See included automated tests.


elliottt (approved):
Thanks for adding these!

PR #4313, opened 2021-06-28
WIP: Defer symbol table resizing until after indexing has run.

jvilk-stripe:
WIP: Defer symbol table resizing until after indexing has run.

Indexing performs a deep copy of the GlobalState once per core, but doesn't define any symbols. It only defines names. As a result, we are wasting (number of cores - 1) * (preallocated size) * (size of Symbol) bytes of memory -- which can be large in big codebases.

### Motivation

Reduce max RSS.

### Test plan

Going to test this on Stripe's codebase and compare w/ the currently deployed version.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -723,30 +723,38 @@ void GlobalState::installIntrinsics() {
     }
 }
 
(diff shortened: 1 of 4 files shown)


jvilk-stripe:

→ https://go/builds/bui_Jkyka9eDQNeiDT
→ https://go/builds/bui_JkykgDHlxYReiz

jvilk-stripe:
This seems to have had no effect.

PR #4323, opened 2021-06-29
[Ruby 3.0] Reserved for numparams

vinistock:
This PR implements the changes for Ruby 3.0 to numbered parameters based on https://github.com/whitequark/parser/pull/725.

The change is to check that the reserved names for numbered parameters are not used anywhere except for numbered parameters themselves.

I had to update some unrelated RBIs and tests because they used these reserved names and started throwing the new error. It might be easier to review per commit.

### Motivation

In Ruby 2.7, assigning a numbered parameter would throw an error only inside blocks. That is
```ruby
lambda {_1 = nil} # error
```

Since Ruby 3.0, the numbered parameter names are actually reserved everywhere. Trying to use `_1` as a variable, method definition or any type of argument will throw an error and the message has changed.

```ruby
def _1; end # error
_1 = nil # error
def foo(_1); end # error
def bar(*_1); end # error
def baz(&_1); end # error
```

### Test plan

See included automated tests.


froydnj (changes requested):
It looks like there are some test failures with some test code that need to be resolved first.

vinistock:
@froydnj fixed the hidden definition generation in the latest commit.

third_party/parser/codegen/generate_diagnostics.cc
```diff
@@ -57,7 +57,7 @@ tuple<string, string> MESSAGES[] = {
     {"InvalidRegexp", "{}"},
     {"InvalidReturn", "invalid return in class/module body"},
     {"CSendInLHSOfMAsgn", "&. inside multiple assignment destination"},
-    {"CantAssignToNumparam", "cannot assign to numbered parameter {}"},
+    {"ReservedForNumparam", "{} is reserved for numbered parameter"},
```
froydnj, line 60: I don't know if this is the best error message, but I think we can fix this in a followup.

froydnj (approved):
Thanks for fixing up the CI failures!

third_party/parser/codegen/generate_diagnostics.cc
vinistock: Agreed. I used the same message Ruby uses.

PR #4324, opened 2021-06-29
Remove usage of endless range

sambostock:
This replaces the use of an endless range to index into an array, with the start index and number of elements to extract.

### Motivation

Endless ranges were introduced in Ruby `2.6`, but `sorbet-runtime` only depends on Ruby `>= 2.3.0`. This is causing errors in CI pipelines testing compatibility with old versions of Ruby.

The alternative would be to bump the minimum version of Ruby instead, but it seemed like an easy fix and allows us to maintain backwards compatibility.

### Test plan

This is a refactor, so (I assume) should be covered by existing tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/utils.rb
+++ b/gems/sorbet-runtime/lib/types/utils.rb
@@ -84,7 +84,7 @@ module T::Utils
       when 0 then nil
       when 1 then non_nil_types.first
       else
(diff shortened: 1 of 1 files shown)


sambostock:
Looks like this is a duplicate of https://github.com/sorbet/sorbet/pull/4281

PR #4330, opened 2021-07-06
Use cast_tree_nonnull in treemap.h.

jvilk-stripe:
Use cast_tree_nonnull in treemap.h when we assume the cast will succeed.

### Motivation

cast_tree_nonnull avoids an extra isa_type check in production builds, and should be slightly faster.

### Test plan

Covered by existing tests.


jvilk-stripe:

→ https://go/builds/bui_Jnt4t2CGY8mvyD
→ https://go/builds/bui_Jnt4n8hMHE03uV

aprocter-stripe (approved):

PR #4333, opened 2021-07-06
WIP Autogen: Don't have all worker threads write to the same vector.

jvilk-stripe:
WIP Autogen: Don't have all worker threads write to the same vector.

This may be unsafe on some platforms, and will undoubtedly result in false sharing as multiple threads attempt to write to adjacent memory locations.

### Motivation

Speed.

### Test plan

Covered by existing tests.

### DIFF
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -194,7 +194,7 @@ struct AutogenResult {
         optional<autogen::Subclasses::Map> subclasses;
     };
     CounterState counters;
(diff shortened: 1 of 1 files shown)


jvilk-stripe:

→ https://go/builds/bui_JnyEP2be50Z871
→ https://go/builds/bui_JnyEg9EgselfkK

jvilk-stripe:
Heh. This didn't really do anything.

PR #4339, opened 2021-07-12
Slightly faster path for simple setters

djudd:
When there's a simple type (and no extra `validate` proc) just use `is_a?` directly instead of going through the T::Types::Base type, inspired by what we do for call validation.

### Motivation
`rake bench:setters` before:
```
T::Struct setter, simple type (μs/iter):
  0.165682   0.000272   0.165954 (  0.166053)
T::Struct setter, nilable type (μs/iter):
  0.174619   0.000403   0.175022 (  0.175314)
```

after:
```
T::Struct setter, simple type (μs/iter):
  0.118319   0.000189   0.118508 (  0.118731)
T::Struct setter, nilable type (μs/iter):
  0.124673   0.000253   0.124926 (  0.125048)
```

### Test plan
Existing tests

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
+++ b/gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
@@ -34,9 +34,40 @@ module T::Props
         # Use separate methods in order to ensure that we only close over necessary
         # variables
         if !T::Props::Utils.need_nil_write_check?(rules) || has_explicit_nil_default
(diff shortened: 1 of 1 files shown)


jez (approved):

jez:
Screen Shot 2021-07-12 at 4 28 00 PM

😅

djudd:
Hah I forgot I switched to being logged into Github.com as my personal account! Will probably switch back for future Sorbet contributions just to avoid confusing myself...

PR #4343, opened 2021-07-12
Better output for the formatting build job

elliottt:
Change the output from all of the linting scripts to instead output markdown when given the `-t` switch. This renders nicer in buildkite, and makes the action expected for each failure more obvious. For an example of the new linter job output: https://buildkite.com/sorbet/sorbet/builds/16223.

### Motivation

Closes #4343 

### Test plan

See included automated tests.

### DIFF
--- a/tools/scripts/format_build_files.sh
+++ b/tools/scripts/format_build_files.sh
@@ -9,7 +9,6 @@ if [ "$1" == "-t" ]; then
 
   # quieter bazel output
   bazel_args=(
-    "--curses=no"
     "--ui_event_filters=-info,-stdout,-stderr"
     "--noshow_progress"
   )
--- a/tools/scripts/format_build_files.sh
+++ b/tools/scripts/format_build_files.sh
@@ -4,14 +4,17 @@ set -e
 
 cd "$(dirname "$0")/../.."
 
+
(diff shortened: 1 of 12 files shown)


tools/scripts/format_build_files.sh
```diff
@@ -4,14 +4,21 @@ set -e
...
-    echo -ne "\\e[0m"
-    echo -e "✨✨ Run \\e[97;1;42m ./tools/scripts/format_build_files.sh\\e[0m to fix them up.  ✨✨"
-    bazel run //test/lint/buildifier:diff
+
+  # quieter bazel output
+  bazel_args=(
+    "--curses=no"
+    "--ui_event_filters=-info,-stdout,-stderr"
+    "--noshow_progress"
+  )
```
jez, line 15: Don't know if you care but some of these are duplicated with `.buildkite/tools/setup-bazel.sh`, and also some aren't there.

(e.g., maybe we want those caching ones?)

just fyi, not blocking.

jez (approved):

tools/scripts/format_build_files.sh
elliottt: Oh nice, I'll remove `--curses=no` 👍

PR #4337, opened 2021-07-08
Revert "[Ruby 3.0] Reserved for numparams"

elliottt:
Reverts sorbet/sorbet#4323

PR #4344, opened 2021-07-12
[Ruby 3.0] reserved for numparams

vinistock:
This PR brings back the parser changes for reserved numbered parameters in Ruby 3.0.

Revert: (#4337)
Original: (#4323)

Commits:
1. Revert of the revert
2. Add auto-correct to change `_[0-9]` into `arg[0-9]` (i.e.: `_1` into `arg1`)
3. Add errors-with-dup to the related tests since we are now emitting two of them 

### Test plan

See included automated tests.


elliottt:
Thanks for adding the autocorrect! I know this will raise a lot of errors on Stripe's codebase, so I'll start trying to land incremental fixes.

jez:

→ https://go/builds/bui_Jyo7UCZoMr43Pq
→ https://go/builds/bui_Jyo79n9WGps3CN

jez:

→ https://go/builds/bui_JzWPo3PodyIEyS
→ https://go/builds/bui_JzWPKZAsf6hCJE

jez:
@vinistock @elliottt I made the changes to Stripe's codebase that allow this change to work on Stripe's codebase.

I don't know what strategy we want to take to roll out this change w.r.t. this being a change that might break other people (in particular, this is a Ruby 3–specific breaking change).

If we could find a way to make this error opt-in for at least one release, and then follow up with a subsequent release that makes it always enforced, that would allow people to benefit from the autocorrect to migrate their codebase incrementally, and then upgrade to the always-enforced version after that.

vinistock:
Does Sorbet have the concept of a warning or a deprecation message? Would we be able to add a CLI option to prevent adding errors, but still display some message to users that their code is using reserved numbered parameters?

elliottt:
> Does Sorbet have the concept of a warning or a deprecation message? Would we be able to add a CLI option to prevent adding errors, but still display some message to users that their code is using reserved numbered parameters?

Sorbet does not have warnings, so adding a CLI option to enable this error would be a great path forward. Maybe something like `--ruby3-numparams`?

vinistock:
But would a CLI option fully solve the rollout issue?

I assume most users don't check periodically for new CLI options. So, if they don't realize there's a new option, when we make this error enabled by default it will still seem - from their perspective - like the errors began all of a sudden.

Would it be possible to print a message that is not associated to an error that simply says something along these lines: `You are using names that are reserved for numbered parameters starting with Ruby 3.0. This will fail in a future version of Sorbet`?

Let me know what you think.

elliottt:
What about having the error on by default, and using the CLI flag to disable it? That way we could use the error message as the place to point out that the CLI flag can be used to disable it during the migration.

vinistock:
That's a great idea. I'll put it together.

vinistock:
Added the new CLI option and adapted the error message. Let me know what you think.

Screen Shot 2021-08-12 at 2 57 02 PM

main/realmain.cc
```diff
@@ -483,6 +483,9 @@ int realmain(int argc, char *argv[]) {
     if (opts.noErrorSections) {
         gs->includeErrorSections = false;
     }
+    if (opts.disableRuby3Numparams) {
+        gs->checkReservedNumberedParameters = false;
+    }
```
elliottt, line 488: @froydnj brought up a good point today, which is that we could use the existing `--suppress-error-code` option to disable this error if it had a unique identifier. What do you think about adding a custom error to `core/errors/parser.h` for the numparams error, and then recommending using `--suppress-error-code` with that number in the error message instead?

elliottt (changes requested):
This looks good, thank you for adding the additional flag! I mentioned this PR at our stand-up today, and @froydnj gave a good suggestion for how to implement the flag without needing to plumb through additional state. What do you think about the approach of using `--suppress-error-code` instead?

Sorry again for how much back and forth there's been on this PR, it's definitely in the home stretch!

main/realmain.cc
vinistock: I like that idea better indeed. Changed the code to use a new error class in the latest commit.

elliottt:

→ https://go/builds/bui_K3cWx4jiBfgPFc
→ https://go/builds/bui_K3cWAdY3ri5fwy

elliottt (approved):
This looks great! I'm going to merge it as-is, as it passed Sorbet's CI and the autocorrect and error message give a clear path forward for anyone that runs into it on the next release.

elliottt:
Looks like I spoke too soon, @vinistock could you take a look at those two test failures?

vinistock:
Weird that I didn't spot those two 🤔. Anyway, they should be fixed now! 😅.

elliottt:

→ https://go/builds/bui_K40BxzvbGaBrmL
→ https://go/builds/bui_K40CPMveWdzJcH

PR #4346, opened 2021-07-12
Fix a typo in the README section about compile_commands.json

aprocter-stripe:
I think the reference to `dbg` instead of `opt` here is stale (the script to generate `compile_commands.json` used to use `dbg` but now uses `opt`).

### Motivation
Fixin' a typo.

### Test plan
Change is docs only.


jez (approved):

README.md
```diff
@@ -903,7 +903,7 @@ tools/scripts/build_compilation_db.sh
 This builds a `./compile_commands.json` file (that is gitignored). This file
 hard-codes some paths into the Bazel sandbox. These files can get stale,
 especially when they're generated by Bazel `genrule`'s. In particular, the
-`./compile_commands.json` references files in Bazel's `dbg` configuration (e.g.,
+`./compile_commands.json` references files in Bazel's `opt` configuration (e.g.,
 whatever was last built with `-c opt` / `--compilation_mode=opt`). If you're
```
jez, line 907:
```suggestion
whatever was last built with `-c dbg` / `--compilation_mode=dbg`). If you're
```

README.md
```diff
@@ -903,7 +903,7 @@ tools/scripts/build_compilation_db.sh
...
 seeing stale errors, consider running a command like `./bazel build
 //main:sorbet -c opt`.
```
jez, line 909:
```suggestion
//main:sorbet --config=dbg`.
```

aprocter-stripe:
@jez Everything needs to be consistent with https://github.com/sorbet/sorbet/blob/master/tools/scripts/build_compilation_db.sh , right? That currently uses `opt`, so I just changed the one `dbg` in the README to be consistent with the script. Do you mean we should change the script to use `dbg` instead? (I'm happy to do whatever.)

jez:
> @jez Everything needs to be consistent with https://github.com/sorbet/sorbet/blob/master/tools/scripts/build_compilation_db.sh , right? That currently uses `opt`, so I just changed the one `dbg` in the README to be consistent with the script. Do you mean we should change the script to use `dbg` instead? (I'm happy to do whatever.)

Ahh sorry, i was misremembering. Thanks for double checking.

PR #4350, opened 2021-07-12
Combine some unnecessarily overloaded sigs

aisamanra:
### Motivation
These sigs use overloads but don't actually need the full power of overloads (since they always return the same type) so this turns them to a `T.any` of their arguments instead.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/float.rbi
+++ b/rbi/core/float.rbi
@@ -329,25 +329,7 @@ class Float < Numeric
   # is returned.
   sig do
     params(
-        arg0: Integer,
-    )
-    .returns(T::Boolean)
-  end
-  sig do
-    params(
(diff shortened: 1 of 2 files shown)


jez:
this _might_ have problems? i seem to remember something weird here in the past

feel free to try running on pay-server first

aisamanra:

→ https://go/builds/bui_JqEKoaHl2JBHxU
→ https://go/builds/bui_JqEKzKHwe4IesF

jez (approved):
lgtm if it passes pay-server

aisamanra:
It's passed pay-server (c.f. the links above), so merging it!

PR #4351, opened 2021-07-12
Implement T::Enum#=== intrinsic

### LINKED ISSUE #4115: A case over T::Enum should not require an explicit T.absurd
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Aclass%20MyEnum%20%3C%20T%3A%3AEnum%0A%20%20enums%20do%0A%20%20%20%20X%20%3D%20new%0A%20%20%20%20Y%20%3D%20new%0A%20%20end%0Aend%0A%0Asig%20%7Bparams(x%3A%20MyEnum).returns(Integer)%7D%0Adef%20example(x)%0A%20%20res%20%3D%20case%20x%0A%20%20when%20MyEnum%3A%3AX%20then%201%0A%20%20when%20MyEnum%3A%3AY%20then%202%0A%20%20end%0A%0A%20%20res%0Aend)

```ruby
# typed: strict
extend T::Sig

class MyEnum < T::Enum
  enums do
    X = new
    Y = new
  end
end

sig {params(x: MyEnum).returns(Integer)}
def example(x)
  res = case x
  when MyEnum::X then 1
  when MyEnum::Y then 2
  end

  res
end
```

#### Observed output

```
editor.rb:18: Expected Integer but found T.nilable(Integer) for method result type https://srb.help/7005
    18 |  res
          ^^^
  Expected Integer for result type of method example:
    editor.rb:12:
    12 |def example(x)
        ^^^^^^^^^^^^^^
  Got T.nilable(Integer) originating from:
    editor.rb:14:
    14 |  when MyEnum::X then 1
                              ^
    editor.rb:15:
    15 |  when MyEnum::Y then 2
                              ^
    editor.rb:12: Possibly uninitialized (NilClass) in:
    12 |def example(x)
        ^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

There should be no error, because the `case` is already exhaustive, so Sorbet should know that the (implicit) `else` branch of the `case` is dead and thus `res` is always initialized with an `Integer`.

- - -

By way of comparison, if we make an analogous example using explicit sealed classes instead of `T::Enum`, the `T.absurd` is not required:

```ruby
# typed: strict
extend T::Sig

class MyEnum
  extend T::Helpers
  abstract!
  sealed!

  class X < MyEnum; include Singleton; final!; end
  class Y < MyEnum; include Singleton; final!; end
end

sig {params(x: MyEnum).returns(Integer)}
def example_with_class_objects(x)
  res = case x
  when MyEnum::X then 1
  when MyEnum::Y then 2
  end

  res # no error
end
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Aclass%20MyEnum%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%20%20sealed!%0A%0A%20%20class%20X%20%3C%20MyEnum%3B%20include%20Singleton%3B%20final!%3B%20end%0A%20%20class%20Y%20%3C%20MyEnum%3B%20include%20Singleton%3B%20final!%3B%20end%0Aend%0A%0Asig%20%7Bparams(x%3A%20MyEnum).returns(Integer)%7D%0Adef%20example_with_class_objects(x)%0A%20%20res%20%3D%20case%20x%0A%20%20when%20MyEnum%3A%3AX%20then%201%0A%20%20when%20MyEnum%3A%3AY%20then%202%0A%20%20end%0A%0A%20%20res%0Aend)

And another example: if instead of the class objects (`MyEnum::X`) we use the singleton instance objects (`MyEnum::X.instance`), we see the same behavior as the example using `T::Enum`:

```ruby
# typed: strict
extend T::Sig

class MyEnum
  extend T::Helpers
  abstract!
  sealed!

  class X < MyEnum; include Singleton; final!; end
  class Y < MyEnum; include Singleton; final!; end
end

sig {params(x: MyEnum).returns(Integer)}
def example_with_singleton_instances(x)
  res = case x
  when MyEnum::X.instance then 1
  when MyEnum::Y.instance then 2
  end

  res # error: Expected `Integer` but found `T.nilable(Integer)` for method result type
end
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Aclass%20MyEnum%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%20%20sealed!%0A%0A%20%20class%20X%20%3C%20MyEnum%3B%20include%20Singleton%3B%20final!%3B%20end%0A%20%20class%20Y%20%3C%20MyEnum%3B%20include%20Singleton%3B%20final!%3B%20end%0Aend%0A%0Asig%20%7Bparams(x%3A%20MyEnum).returns(Integer)%7D%0Adef%20example_with_singleton_instances(x)%0A%20%20res%20%3D%20case%20x%0A%20%20when%20MyEnum%3A%3AX.instance%20then%201%0A%20%20when%20MyEnum%3A%3AY.instance%20then%202%0A%20%20end%0A%0A%20%20res%0Aend)

aprocter-stripe:
The root of #4115 seems to be that the `case` desugars to `===` on the singletons for the enums, not the classes. Implementing an intrinsic for `T::Enum#===` seems to me like a reasonable way to fix it, ~~but I'm tossing a wip PR up first to get feedback on whether it's the right approach~~.

### Motivation
Fixes #4115

### Test plan
See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3106,7 +3106,7 @@ public:
         }
         auto lhs = args.thisType;
         ENFORCE(!lhs.isUntyped(), "lhs of T::Enum.=== must be typed");
-        if (Types::glb(gs, rhs, lhs).isBottom()) {
+        if (Types::glb(gs, rhs, lhs).isBottom() && Types::glb(gs, rhs, Types::String()).isBottom()) {
             res.returnType = Types::falseClass();
             return;
         }
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3072,25 +3072,27 @@ public:
     }
 } Module_tripleEq;
 
-// Returns true if the referenced class is a child of T::Enum.
-static bool isEnumClass(const GlobalState &gs, ClassOrModuleRef klass) {
-    if (!klass.exists()) {
+// Returns true if the type is exactly the class of a specific enum value. For example, given:
+//
+//   class C < T::Enum
+//     enums do
+//       X = new
+//       Y = new
+//     end
+//   end
+//
+//  class TotallyUnrelatedThing ; end
+//
+// returns true for "X" and "Y", but false for "C", "T::Enum", and "TotallyUnrelatedThing".
+static bool isEnumValueClass(const GlobalState &gs, const TypePtr &type) {
+    bool must_exist = false;
+    auto unwrapped = unwrapSymbol(gs, type, must_exist);
+
+    if (!unwrapped.exists()) {
         return false;
     }
 
-    auto superClass = klass.data(gs)->superClass();
-    return superClass.exists() && superClass == Symbols::T_Enum();
-}
-
-// Returns true if the referenced class is a grandchild of T::Enum, i.e., one of the singleton classes for the enum
-// values.
-static bool isEnumValue(const GlobalState &gs, ClassOrModuleRef klass) {
-    if (!klass.exists()) {
-        return false;
-    }
-
-    auto superClass = klass.data(gs)->superClass();
-    return isEnumClass(gs, superClass);
+    return unwrapped.data(gs)->name.isTEnumName(gs);
 }
 
 class T_Enum_tripleEq : public IntrinsicMethod {
@@ -3106,17 +3108,14 @@ public:
         }
         auto lhs = args.thisType;
         ENFORCE(!lhs.isUntyped(), "lhs of T::Enum.=== must be typed");
+
+        // We have to allow String on the rhs, in order to support legacy_t_enum_migration_mode.
         if (Types::glb(gs, rhs, lhs).isBottom() && Types::glb(gs, rhs, Types::String()).isBottom()) {
             res.returnType = Types::falseClass();
             return;
         }
 
-        // If lhs and rhs are both grandchildren of T::Enum, we can safely assume they are singleton classes created by
-        // the T::Enum rewriter, since inheriting from children of T::Enum is otherwise prohibited.
-        bool must_exist = false;
-        auto lhs_unwrapped = unwrapSymbol(gs, lhs, must_exist);
-        auto rhs_unwrapped = unwrapSymbol(gs, rhs, must_exist);
-        if (Types::isSubType(gs, rhs, lhs) && isEnumValue(gs, lhs_unwrapped) && isEnumValue(gs, rhs_unwrapped)) {
+        if (isEnumValueClass(gs, lhs) && Types::isSubType(gs, rhs, lhs)) {
             res.returnType = Types::trueClass();
             return;
         }
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3072,6 +3072,7 @@ public:
     }
 } Module_tripleEq;
 
+// Returns true if the referenced class is a child of T::Enum.
 static bool isEnumClass(const GlobalState &gs, ClassOrModuleRef klass) {
     if (!klass.exists()) {
         return false;
@@ -3081,6 +3082,8 @@ static bool isEnumClass(const GlobalState &gs, ClassOrModuleRef klass) {
     return superClass.exists() && superClass == Symbols::T_Enum();
 }
 
+// Returns true if the referenced class is a grandchild of T::Enum, i.e., one of the singleton classes for the enum
+// values.
 static bool isEnumValue(const GlobalState &gs, ClassOrModuleRef klass) {
     if (!klass.exists()) {
         return false;
@@ -3108,6 +3111,8 @@ public:
             return;
         }
 
+        // If lhs and rhs are both grandchildren of T::Enum, we can safely assume they are singleton classes created by
+        // the T::Enum rewriter, since inheriting from children of T::Enum is otherwise prohibited.
         bool must_exist = false;
         auto lhs_unwrapped = unwrapSymbol(gs, lhs, must_exist);
         auto rhs_unwrapped = unwrapSymbol(gs, rhs, must_exist);
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3072,6 +3072,31 @@ public:
     }
 } Module_tripleEq;
 
+class T_Enum_tripleEq : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        if (args.args.size() != 1) {
+            return;
+        }
+        auto rhs = args.args[0]->type;
+        if (rhs.isUntyped()) {
+            res.returnType = rhs;
+            return;
+        }
+        auto lhs = args.thisType;
+        ENFORCE(!lhs.isUntyped(), "lhs of T::Enum.=== must be typed");
+        if (Types::isSubType(gs, rhs, lhs)) {
+            res.returnType = Types::trueClass();
+            return;
+        }
+        if (Types::glb(gs, rhs, lhs).isBottom()) {
+            res.returnType = Types::falseClass();
+            return;
+        }
+        res.returnType = Types::Boolean();
+    }
+} T_Enum_tripleEq;
+
 } // namespace
 
 const vector<Intrinsic> intrinsicMethods{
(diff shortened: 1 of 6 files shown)


aprocter-stripe:
(Just for some additional context: this does seem to fix the errors in #4115, and I think it matches the behavior of the runtime [see https://github.com/sorbet/sorbet/blob/b12cbff755c13f2c659ccfa4688e78729d8faa46/gems/sorbet-runtime/lib/types/enum.rb#L220-L233] except for the `legacy_t_enum_migration_mode`.)

core/types/calls.cc
```diff
@@ -3010,6 +3010,31 @@ class Module_tripleEq : public IntrinsicMethod {
     }
 } Module_tripleEq;
 
+class T_Enum_tripleEq : public IntrinsicMethod {
```
aprocter-stripe, line 3013: Note that the implementation is based on that of `Module_tripleEq` above.

elliottt: Oh, `Module_tripleEq` wasn't firing because the receiver was a value of one of the generated enum classes?

aprocter-stripe: Yup!

elliottt (changes requested):
I think that this approach makes sense, given that the receiver of `===` before was one of the singleton enum values instead of a class/module. What do you think @jez?

core/types/calls.cc
```diff
@@ -3010,6 +3010,31 @@ class Module_tripleEq : public IntrinsicMethod {
...
+            return;
+        }
+        auto lhs = args.thisType;
+        ENFORCE(!lhs.isUntyped(), "lhs of T::Enum.=== must be typed");
+        if (Types::isSubType(gs, rhs, lhs)) {
+            res.returnType = Types::trueClass();
+            return;
+        }
+        if (Types::glb(gs, rhs, lhs).isBottom()) {
+            res.returnType = Types::falseClass();
```
jez, line 3031: Can you make sure to write a test for this?

I think that this should work:

```ruby
T.reveal_type(MyEnum::Y.===(MyEnum::X))
```

also it would be great to have both the examples from #4115 committed as test cases.

jez:
> I think that this approach makes sense, given that the receiver of `===` before was one of the singleton enum values instead of a class/module. What do you think @jez?

This is the best approach. Thanks!

elliottt (approved):
Nice, great tests!

aprocter-stripe:

→ https://go/builds/bui_JqWM7IQSVLNGVZ
→ https://go/builds/bui_JqWNbeFJ6YHUR4

aprocter-stripe:
Looks like I goofed something up here.

```ruby
# typed: true
extend T::Sig

class MyEnum < T::Enum ; end

sig {params(x: T::Enum, y: MyEnum).void}
def f(x, y)
  T.reveal_type(x===y)
end

sig {params(x: MyEnum, y: MyEnum).void}
def g(x, y)
  T.reveal_type(x===y)
end
```

On sorbet.run, both of the above reveal as `T::Boolean`. On this branch they reveal as `T::TrueType`, which is uh wrong. :) Will look into what I did wrong in cargo-culting from `Module#===`.

aprocter-stripe:

→ https://go/builds/bui_JqcojtMIA1LKLN
→ https://go/builds/bui_JqcoNstf71fgLx

aprocter-stripe:

→ https://go/builds/bui_JqdJt09NHtwwq7
→ https://go/builds/bui_JqdJ8CzkfQ6QI5

aprocter-stripe:

→ https://go/builds/bui_JqfS5Hym0W3K8i
→ https://go/builds/bui_JqfSnnD4hB2MoH

aprocter-stripe:
So I fixed a couple things:

1. Unbroke behavior when `===`-comparing abstract enum things (i.e. `x===y` where x and y have `T::Enum` or `SomeEnum`; initially the issues here didn't reveal themselves since I was just looking at desugared `case` statements).
2. Added support for `String` on the rhs, to support `legacy_t_enum_migration_mode` (just returns `T::Boolean`).

I'm feeling reasonably good about the correctness here, and I have a green build on Stripe's codebase that doesn't require anything surprising, but since the logic has changed quite a bit a second review is probably in order.

core/types/calls.cc
```diff
@@ -3072,6 +3072,59 @@ class Module_tripleEq : public IntrinsicMethod {
...
+        return false;
+    }
+
+    auto superClass = klass.data(gs)->superClass();
+    return superClass.exists() && superClass == Symbols::T_Enum();
+}
+
+// Returns true if the referenced class is a grandchild of T::Enum, i.e., one of the singleton classes for the enum
+// values.
+static bool isEnumValue(const GlobalState &gs, ClassOrModuleRef klass) {
```
elliottt, line 3087:
```suggestion
// Returns true if the referenced class is a grandchild of T::Enum, i.e., one of the classes for the singleton enum
// values.
static bool isEnumValueClass(const GlobalState &gs, ClassOrModuleRef klass) {
```

core/types/calls.cc
```diff
@@ -3072,6 +3072,59 @@ class Module_tripleEq : public IntrinsicMethod {
...
+            res.returnType = Types::falseClass();
+            return;
+        }
+
+        // If lhs and rhs are both grandchildren of T::Enum, we can safely assume they are singleton classes created by
+        // the T::Enum rewriter, since inheriting from children of T::Enum is otherwise prohibited.
+        bool must_exist = false;
+        auto lhs_unwrapped = unwrapSymbol(gs, lhs, must_exist);
+        auto rhs_unwrapped = unwrapSymbol(gs, rhs, must_exist);
+        if (Types::isSubType(gs, rhs, lhs) && isEnumValue(gs, lhs_unwrapped) && isEnumValue(gs, rhs_unwrapped)) {
```
elliottt, line 3119:
```suggestion
        if (Types::isSubType(gs, rhs, lhs) && isEnumValueClass(gs, lhs_unwrapped) && isEnumValueClass(gs, rhs_unwrapped)) {
```

elliottt: What do you think about reordering this conditional to check if rhs is a subtype of lhs last?

elliottt: What about rewriting this to check that RHS is a subtype of LHS, and LHS is a ClassType with a TEnum name? (You can use the `NameRef::isTEnumName` helper to test that.)

elliottt (changes requested):

aprocter-stripe:
@elliottt Think I got everything, and also added a test for "subset" enums as you pointed out in Slack. (Those actually did seem to work with the old approach (checking for grandchildren of T::Enum), and they also work with the new (checking the name kind).)

elliottt (approved):
I think this looks great, thanks for all the tests!

core/types/calls.cc
```diff
@@ -3072,6 +3072,58 @@ class Module_tripleEq : public IntrinsicMethod {
...
+static bool isEnumValueClass(const GlobalState &gs, const TypePtr &type) {
+    bool must_exist = false;
+    auto unwrapped = unwrapSymbol(gs, type, must_exist);
+
+    if (!unwrapped.exists()) {
+        return false;
+    }
+
+    return unwrapped.data(gs)->name.isTEnumName(gs);
+}
```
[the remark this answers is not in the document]
aprocter-stripe: Good call --- I think inlining is the way to go. (Started writing it this way 'cause I thought I'd need it in more than one place.) Will do that before merge.

aprocter-stripe:

→ https://go/builds/bui_JqxlF7P4JZZNMR
→ https://go/builds/bui_JqxmI6BrAFpFVW

aprocter-stripe:

→ https://go/builds/bui_JrJaYE9CYrk26d
→ https://go/builds/bui_JrJaP2T0kFNkSY

PR #4354, opened 2021-07-13
Don't recursively dispatch on MetaType

jez:
### Motivation

This just leads to nonsensical errors, like the two that are now gone from the linked file.

I tested this on Stripe's codebase to make sure that no one was actually relying on dispatching to the underlying type (which is good, because it wouldn't really make sense). We already emit an error for this codepath otherwise.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_JqGuCUt2O14P2t
→ https://go/builds/bui_JqGugrangPYNrl

elliottt (approved):

PR #4355, opened 2021-07-13
Fix sig suggestions when keyword arguments are present

### LINKED ISSUE #3656: Recent keyword args changes broke sig suggestion
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams(x%3A%20Integer%2C%20y%3A%20String).void%7D%0Adef%20foo(x%3A%2C%20y%3A)%0Aend%0A%0Adef%20bar(x%2C%20y)%0A%20%20foo(x%3A%20x%2C%20y%3A%20y)%0Aend)

```ruby
# typed: strict
extend T::Sig

sig {params(x: Integer, y: String).void}
def foo(x:, y:)
end

def bar(x, y)
  foo(x: x, y: y)
end
```

#### Observed output

```
editor.rb:8: This function does not have a sig https://srb.help/7017
     8 |def bar(x, y)
        ^^^^^^^^^^^^^
    editor.rb:8: Insert sig {params(x: String, y: T.untyped).void}

     8 |def bar(x, y)
        ^
Errors: 1
```

#### Expected behavior

The suggested sig should be

```ruby
sig {params(x: Integer, y: String).void}
```

- - -

SigSuggestion.cc is doing something dumb—indexing directly into the `params` vector of the method symbol with an index that comes from a send's arguments array:

https://github.com/sorbet/sorbet/blob/7bd53b68bfb8165a5bcd76057522671cbeecfc46/infer/SigSuggestion.cc#L59-L70

This ignores the fact that a send's keyword args effectively are stored like this:

```
foo(:x, x, :y, y)
```

with `Send::numPosArgs` controlling where the positional arguments end and the keyword key/value pairs start.

elliottt:
Fix signature suggestions for methods that have kwargs sends in their body.

### Motivation

Fixes #3656 

### Test plan

See included automated tests.

### DIFF
--- a/infer/SigSuggestion.cc
+++ b/infer/SigSuggestion.cc
@@ -115,7 +115,7 @@ void extractSendArgumentKnowledge(core::Context ctx, core::LocOffsets bindLoc, c
 
     auto numPosArgs = snd->numPosArgs;
     auto numKwArgs = snd->args.size() - numPosArgs;
-    bool hasKwSplat = numKwArgs & 0x1;
+    bool hasKwSplat = numKwArgs % 2 == 1;
 
     // Since this is a "fake" dispatch and we are not going to display the errors anyway,
     // core::Loc::none() should be okay here.
@@ -141,7 +141,7 @@ void extractSendArgumentKnowledge(core::Context ctx, core::LocOffsets bindLoc, c
                 }
             }
 
-            // increment over the keyword argument
+            // increment over the keyword argument symbol
             i++;
         }
 
--- a/infer/SigSuggestion.cc
+++ b/infer/SigSuggestion.cc
@@ -56,16 +56,34 @@ optional<core::AutocorrectSuggestion::Edit> maybeSuggestExtendTSig(core::Context
     return core::AutocorrectSuggestion::Edit{nextLineLoc.value(), fmt::format("{}extend T::Sig\n", prefix)};
 }
 
-core::TypePtr extractArgType(core::Context ctx, cfg::Send &send, core::DispatchComponent &component, int argId) {
+core::TypePtr extractArgType(core::Context ctx, cfg::Send &send, core::DispatchComponent &component,
+                             optional<core::NameRef> keyword, int argId) {
     ENFORCE(component.method.exists());
     const auto &args = component.method.data(ctx)->arguments();
     if (argId >= args.size()) {
         return nullptr;
     }
-    const auto &to = args[argId].type;
+
+    core::TypePtr to = nullptr;
+    if (keyword.has_value()) {
+        auto name = *keyword;
+        for (auto &argInfo : args) {
+            if (argInfo.flags.isKeyword && argInfo.name == name) {
+                to = argInfo.type;
+                break;
+            }
+        }
+        if (to == nullptr) {
+            return nullptr;
+        }
+    } else {
+        to = args[argId].type;
+    }
+
     if (!to || !to.isFullyDefined()) {
         return nullptr;
     }
+
     return to;
 }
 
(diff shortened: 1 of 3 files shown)


infer/SigSuggestion.cc
```diff
@@ -56,16 +56,34 @@ optional<core::AutocorrectSuggestion::Edit> maybeSuggestExtendTSig(core::Context
...
+                to = argInfo.type;
+                break;
+            }
+        }
+        if (to == nullptr) {
+            return nullptr;
+        }
+    } else {
+        to = args[argId].type;
+    }
```
elliottt, line 81: We were only considering positional arguments before as sends previously took all the keyword arguments as a single positional hash. With inlined keyword arguments we can make use of the symbols we parse out of the send.

infer/SigSuggestion.cc
```diff
@@ -89,30 +107,52 @@ void extractSendArgumentKnowledge(core::Context ctx, core::LocOffsets bindLoc, c
...
     auto dispatchInfo = snd->recv.type.dispatchCall(ctx, dispatchArgs);
 
-    int i = -1;
-
     // See if we can learn what types should they have
-    for (auto &arg : snd->args) {
-        i++;
+    bool inKwArgs = false;
+    for (int i = 0, argIdx = 0; i < snd->args.size(); i++, argIdx++) {
+        inKwArgs = inKwArgs || (i >= numPosArgs && !hasKwSplat);
```
elliottt, line 127: The previous for loop was incrementing `i` for each element of `snd->args`, which would treat the inlined keyword argument pairs the same way as normal positional arguments.

test/cli/suggest-sig/suggest-sig.out
```diff
@@ -306,7 +306,7 @@ suggest-sig.rb:112: This function does not have a `sig` https://srb.help/7017
      112 |  def self.load_account_business_profile(merchant)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   Autocorrect: Done
-    suggest-sig.rb:112: Inserted `sig {params(merchant: T.untyped).returns(Integer)}
+    suggest-sig.rb:112: Inserted `sig {params(merchant: TestCarash::Merchant).returns(Integer)}
```
elliottt, line 309: This is a case where we weren't able to get information about the `merchant` parameter before, as it was used locally with a keyword args send.

infer/SigSuggestion.cc
```diff
@@ -89,30 +112,52 @@ void extractSendArgumentKnowledge(core::Context ctx, core::LocOffsets bindLoc, c
...
+        optional<core::NameRef> keyword;
+        if (inKwArgs) {
+            if (core::isa_type<core::LiteralType>(snd->args[i].type)) {
+                auto lit = core::cast_type_nonnull<core::LiteralType>(snd->args[i].type);
+                if (lit.literalKind == core::LiteralType::LiteralTypeKind::Symbol) {
+                    keyword = lit.asName(ctx);
+                }
+            }
+
+            // increment over the keyword argument
```
jez, line 144:
```suggestion
            // increment over the keyword argument symbol
```

infer/SigSuggestion.cc
```diff
@@ -89,30 +112,52 @@ void extractSendArgumentKnowledge(core::Context ctx, core::LocOffsets bindLoc, c
         snd->receiverLoc,
         snd->argLocs,
     };
+
+    auto numPosArgs = snd->numPosArgs;
+    auto numKwArgs = snd->args.size() - numPosArgs;
+    bool hasKwSplat = numKwArgs & 0x1;
```
jez, line 118:
```suggestion
    bool hasKwSplat = numKwArgs % 2 == 1;
```

infer/SigSuggestion.cc
```diff
@@ -56,16 +56,39 @@ optional<core::AutocorrectSuggestion::Edit> maybeSuggestExtendTSig(core::Context
...
+            return nullptr;
+        }
+    } else {
+        auto &argInfo = args[argId];
+        if (argInfo.flags.isKeyword) {
+            return nullptr;
+        }
+
+        to = argInfo.type;
+    }
```
jez, line 86: Is it worth leaving a comment here somewhere about what might need to be changed to do something smarter for implicit keyword args? (`i.e., to get `T::Hash[Symbol, Integer]`)

jez (approved):
Thanks!

infer/SigSuggestion.cc
elliottt: Yep, definitely. I'll add one.

PR #4360, opened 2021-07-13
Reject modules as superclasses

### LINKED ISSUE #2252: Superclasses are not checked to be classes
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Amodule%20A%3B%20end%0Aclass%20B%20%3C%20A%3B%20end%0A)

```ruby
# typed: true

module A; end
class B < A; end
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

I would expect this to raise some kind of type error when `srb tc` is run, as subclassing a non-Class is always illegal (to my knowledge) and will throw an exception in the Ruby interpreter. I found this by forgetting to change `module` to `class` after refactoring an interface into an abstract class.

elliottt:
Reject classes declared with a superclass that is a module

### Motivation

Fixes #2252

### Test plan

See included automated tests.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -502,6 +502,17 @@ void validateSuperClass(core::Context ctx, const core::ClassOrModuleRef sym, con
         return;
     }
 
+    // If the ancestors are empty or the first element is a todo, this means that there was syntactically no superclass
+    if (classDef.ancestors.empty()) {
+        return;
+    }
+
+    if (auto *cnst = ast::cast_tree<ast::ConstantLit>(classDef.ancestors.front())) {
+        if (cnst->symbol == core::Symbols::todo()) {
+            return;
+        }
+    }
+
     const auto superClass = sym.data(ctx)->superClass();
     if (!superClass.exists()) {
         // Happens for certain special classes at the top of the inheritance hierarchy.
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -517,8 +517,6 @@ void validateSuperClass(core::Context ctx, const core::ClassOrModuleRef sym, con
         return;
     }
 
-    // These base cases are to replicate the behavior of the VM:
-    // https://github.com/ruby/ruby/blob/b91b3bc7717a97f4f1cdf6131b1688e1958dcfed/class.c#L660-L677
     if (superSingleton.data(ctx)->derivesFrom(ctx, core::Symbols::Class())) {
         return;
     }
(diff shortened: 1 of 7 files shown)


elliottt:

→ https://go/builds/bui_Jqu8akXrojX7H7
→ https://go/builds/bui_Jqu91vHFEE1Gjd

definition_validator/validator.cc
```diff
@@ -495,6 +495,39 @@ void validateSealed(core::Context ctx, const core::ClassOrModuleRef klass, const
     validateSealedAncestorHelper(ctx, singleton, classDef, klass, "extended");
 }
 
+void validateSuperClass(core::Context ctx, const core::ClassOrModuleRef sym, const ast::ClassDef &classDef) {
+    if (!sym.data(ctx)->isClassOrModuleClass()) {
+        return;
```
jez, line 500:
```suggestion
        ENFORCE(classDef.kind != ast::ClassDef::Kind::Class);
        return;
```

definition_validator/validator.cc
```diff
@@ -495,6 +495,39 @@ void validateSealed(core::Context ctx, const core::ClassOrModuleRef klass, const
...
+    }
+
+    const auto superClass = sym.data(ctx)->superClass();
+    if (!superClass.exists()) {
+        return;
```
jez, line 505:
```suggestion
        // Happens for certain special classes at the top of the inheritance hierarchy.
        return;
```

test/testdata/lsp/symbol_query_filter__main.rb
```diff
@@ -16,7 +16,7 @@ def isFoo
   end
 end
 
-class SomeClass
+class SomeClass # error: The super class `AnotherModule::NestedModule` of `SomeClass` does not derive from `Class`
```
aprocter-stripe, line 19: Extremely slight bummer that this error locates here when the actual offense is down below when we reopen the class. (But I guess one sees it down there as well, and I dunno if there's any way to make this not happen.)

test/testdata/resolver/module_superclass.rb
```diff
@@ -0,0 +1,20 @@
...
+class SubclassModule2 < Module; end
+
+# It's okay to subclass Class
+class SubclassClass < Class; end
+
+module M; end
+
+# It's not okay to subclass a module
+  class SubclassModule1 < M; end
+# ^^^^^^^^^^^^^^^^^^^^^^^^^ error: The super class `M` of `SubclassModule1` does not
```
aprocter-stripe, line 19:
```suggestion
# ^^^^^^^^^^^^^^^^^^^^^^^^^ error: The super class `M` of `SubclassModule1` does not derive from `Class`
```

aprocter-stripe (approved):
Good stuff!

test/testdata/lsp/symbol_query_filter__main.rb
elliottt: Good point. I've added an additional check to exit early if the `classDef` is missing a superclass 👍

PR #4362, opened 2021-07-14
Don't run typecheck when --stop-after namer is present

elliottt:
Parts of the resolver pass are still run when `--stop-after namer` is given, but should be skipped.

### Motivation

Fixes a test failure in #4360.

### Test plan

See included automated tests.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -535,6 +535,10 @@ ast::ParsedFile typecheckOne(core::Context ctx, ast::ParsedFile resolved, const 
     ast::ParsedFile result{ast::MK::EmptyTree(), resolved.file};
     core::FileRef f = resolved.file;
 
+    if (opts.stopAfterPhase == options::Phase::NAMER) {
+        return result;
+    }
+
     resolved = definition_validator::runOne(ctx, std::move(resolved));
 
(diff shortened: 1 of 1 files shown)


main/realmain.cc
```diff
@@ -589,9 +589,14 @@ int realmain(int argc, char *argv[]) {
...
-            if (gs->hadCriticalError()) {
-                gs->errorQueue->flushAllErrors(*gs);
+
+            // pipeline::resolve runs the namer, and will exit early if `--stop-after namer` is present.
+            // pipeline::typecheck doesn't need to be run if that flag was present.
+            if (opts.stopAfterPhase != options::Phase::NAMER) {
+                indexed = move(pipeline::typecheck(gs, move(indexed), opts, *workers).result());
+                if (gs->hadCriticalError()) {
+                    gs->errorQueue->flushAllErrors(*gs);
+                }
```
jez, line 599: https://github.com/sorbet/sorbet/blob/a73ce091fe2ee639a5b970fd994a47b5f8089879/main/pipeline/pipeline.cc#L549

Should we delete this?

jez (approved):

main/realmain.cc
elliottt: I think so. Thanks for this comment, I think moving the namer case above the trailing end of the resolver here is better 👍

Issue #4064, opened 2021-03-08
Withhold autocorrects suggesting private methods at public-only callsites


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20private%20def%20my_private_method%3B%20end%0Aend%0A%0AA.new.my_private_metho)

```ruby
# typed: true

class A
  private def my_private_method; end
end

A.new.my_private_metho
```

#### Observed output

```
editor.rb:7: Method my_private_metho does not exist on A https://srb.help/7003
     7 |A.new.my_private_metho
        ^^^^^^^^^^^^^^^^^^^^^^
  Got A originating from:
    editor.rb:7:
     7 |A.new.my_private_metho
        ^^^^^
  Autocorrect: Use `-a` to autocorrect
    editor.rb:7: Replace with my_private_method
     7 |A.new.my_private_metho
              ^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

If you apply the suggestion with `-a`, or take the code action in your editor, the resulting program will have more errors:

```ruby
# typed: true

class A
  private def my_private_method; end
end

A.new.my_private_method
```

```
editor.rb:7: Non-private call to private method my_private_method on A https://srb.help/7031
     7 |A.new.my_private_method
        ^^^^^^^^^^^^^^^^^^^^^^^
    editor.rb:4: Defined in A here
     4 |  private def my_private_method; end
                  ^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20private%20def%20my_private_method%3B%20end%0Aend%0A%0AA.new.my_private_method)

We should not suggest private methods in the "did you mean?" suggestions. (Maybe entirely, which would definitely be easier, or maybe only when the method call site has the `isPrivateOk` flag, though I don't think we currently thread that information down through the places where we would need it to, currently, so it would  be a more invasive change.) I'm fine with either the approximate former solution or the exact latter solution.

- - -

froydnj:
I believe #4363 fixed this issue.

PR #4363, opened 2021-07-14
Autocorrect should not suggest private methods

jmcclellan-stripe:
If a call to a nonexistent method had a similar name to a private method, sorbet would suggest replacing the call with one to the private method, resulting in a new error. This makes it so private methods are not suggested.

Intern hackathon project with @melaniecooray-stripe @haoyang-stripe @ameles-stripe @nehadhingra-stripe!

### Motivation

https://github.com/sorbet/sorbet/issues/4064

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -632,7 +632,8 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
                     for (auto alternative : alternatives) {
                         auto possibleSymbol = alternative.symbol;
                         if (!possibleSymbol.isClassOrModule() &&
-                            (!possibleSymbol.isMethod() || possibleSymbol.data(gs)->isMethodPrivate())) {
+                            (!possibleSymbol.isMethod() ||
+                             (possibleSymbol.data(gs)->isMethodPrivate() && !args.isPrivateOk))) {
                             continue;
                         }
 
@@ -1237,7 +1238,8 @@ DispatchResult MetaType::dispatchCall(const GlobalState &gs, const DispatchArgs 
                                           {wrapped, args.fullType.origins},
                                           wrapped,
                                           args.block,
-                                          args.originForUninitialized};
+                                          args.originForUninitialized,
+                                          args.isPrivateOk};
             auto original = wrapped.dispatchCall(gs, innerArgs);
             original.returnType = wrapped;
             original.main.sendTp = wrapped;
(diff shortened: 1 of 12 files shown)


core/types/calls.cc
```diff
@@ -631,7 +631,8 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
                     lines.reserve(alternatives.size());
                     for (auto alternative : alternatives) {
                         auto possibleSymbol = alternative.symbol;
-                        if (!possibleSymbol.isClassOrModule() && !possibleSymbol.isMethod()) {
+                        if (!possibleSymbol.isClassOrModule() &&
+                            (!possibleSymbol.isMethod() || possibleSymbol.data(gs)->isMethodPrivate())) {
```
jez, line 635: Thanks for the PR! This approach is not quite right. The problem is that this will mean not suggesting private methods when they're actually in scope. Here's a test case:

```ruby
class A
  private def my_private_method; end

  def example
    my_private_metho
  end
end
```

The ticket mentions `isPrivateOk`, which is a property of a method call site (`ast::Send`), not a symbol. You will want to combine the information from the symbol and the send and the symbol.

jez (changes requested):
^

test/cli/autocorrect-private-not-suggested/autocorrect-private-not-suggested.sh
```diff
@@ -0,0 +1,7 @@
+# verifies that the file is unchanged when suggestions are applied
+# i.e. it does not replace the unknown method with a private one
+tmp=$(mktemp -d)
+cp test/cli/autocorrect-private-not-suggested/*.rb "$tmp"
+main/sorbet --silence-dev-message -a "$tmp"/*.rb 2>/dev/null
+cat "$tmp"/*.rb
```
jez, line 6: For simplicity, can you combine both of these tests into a single CLI test, and make that one CLI test output both the errors and the file post-autocorrect?

If you're looking for a test to cargo cult, this is a good one

https://github.com/sorbet/sorbet/blob/master/test/cli/suggest_t_must/suggest_t_must.sh

jez (approved):
This looks great, thanks! I had one small request for how to change the test slightly, but it looks great on the whole.

Thanks again!

jmcclellan-stripe:
Thanks for reviewing and for your help on this!

PR #4364, opened 2021-07-14
[Ruby 3.0] Add support for find pattern

vinistock:
Changes from whitequark
- https://github.com/whitequark/parser/pull/764/
- https://github.com/whitequark/parser/pull/714

This PR adds support for the find pattern, which is another variation of the pattern matching syntax.

The syntax works as follows: it allows finding specific matches inside structures and (optionally) assigning it to a variable. For example,
```ruby
foo = [1, 2, 3, 4, 5]

case foo; in [*x, 3 => a, *y] then true; end # foo has the element 3 inside, therefore this will return true and set a = 3

case foo; in [*x, 7 => a, *y] then true; end # foo does not have the element 7, therefore this will raise NoMatchingPatternError
```

### Motivation

Part of the effort to support Ruby 3.0.

### Test plan

See included automated tests.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -779,6 +779,20 @@ public:
         return make_unique<False>(tokLoc(tok));
     }
 
+    unique_ptr<Node> find_pattern(const token *lbrack_t, sorbet::parser::NodeVec elements, const token *rbrack_t) {
+        auto loc = collectionLoc(elements);
+
+        if (lbrack_t != nullptr) {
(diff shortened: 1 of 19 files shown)


jez:

→ https://go/builds/bui_JwEGm4dUTxuhPS
→ https://go/builds/bui_JwEGsy8spQriOj

jez (approved):
Thanks!

PR #4368, opened 2021-07-15
Allow updating true/false in a loop

### LINKED ISSUE #4174: Incorrect "unreachable code" error (sorbet 7006) for loop conditional on boolean switch
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0As%20%3D%20false%0A2.times%20do%0A%20%20puts%20'ok'%20if%20s%0A%20%20s%20%3D%20true%0Aend%0A)

```ruby
s = false
2.times do
  puts 'ok' if s
  s = true
end
```

#### Observed output

```

# typed: true
extend T::Sig

s = false
2.times do
  puts 'ok' if s
  s = true
end
editor.rb:6: This code is unreachable https://srb.help/7006
     6 |  puts 'ok' if s
               ^^^^
    editor.rb:6: This condition was always falsy (FalseClass)
     6 |  puts 'ok' if s
                       ^
  Got FalseClass originating from:
    editor.rb:4:
     4 |s = false
            ^^^^^

editor.rb:7: Changing the type of a variable in a loop is not permitted https://srb.help/7001
     7 |  s = true
              ^^^^
  Existing variable has type: FalseClass
  Attempting to change type to: TrueClass

  Original type from:
    editor.rb:4:
     4 |s = false
            ^^^^^
    editor.rb:6:
     6 |  puts 'ok' if s
                       ^
Errors: 2 
```

#### Expected behavior

There should not be a `7006` error. It's just incorrect, the code is not unreachable. I understand why it says that, since it infers that s is always `false`. But still, claiming this is unreachable based on an aggressive inference seems inconsistent with Sorbet's philosophy of gradualism.

Might it make more sense to by default infer that locals assigned to `true`/`false` are `T::Boolean` rather than `TrueClass`/`FalseClass`?

- - -

The error goes away when you declare `s` with `s = T.let(false, T::Boolean)`.

jez:
### Motivation

Fixes #4174

### Test plan

See included automated tests.

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -198,6 +198,12 @@ public:
         return make_expression<ast::Literal>(loc, core::Types::falseClass());
     }
 
+    static ExpressionPtr Boolean(core::LocOffsets loc, bool value) {
+        auto name = value ? core::Names::booleanTrue() : core::Names::booleanFalse();
+        auto magic = Constant(loc, core::Symbols::Magic());
+        return Send0(loc, std::move(magic), name);
+    }
+
     static ExpressionPtr UnresolvedConstant(core::LocOffsets loc, ExpressionPtr scope, core::NameRef name) {
         return make_expression<UnresolvedConstantLit>(loc, std::move(scope), name);
     }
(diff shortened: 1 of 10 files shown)


jez:

→ https://go/builds/bui_JrIKHxVvD2mend
→ https://go/builds/bui_JrIKwl3ZQqtMIm

PR #4372, opened 2021-07-15
Better error message for `T.class_of(T.any(...))`

### LINKED ISSUE #4342: Better error message for `T.class_of(T.any(...))`
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20A%3B%20end%0Aclass%20B%3B%20end%0A%0Asig%20%7Breturns(T.class_of(T.any(A%2C%20B)))%7D%0Adef%20foo%3B%20end)

```ruby
# typed: true
extend T::Sig

class A; end
class B; end

sig {returns(T.class_of(T.any(A, B)))}
def foo; end
```

#### Observed output

```
editor.rb:7: T.class_of needs a Class as its argument https://srb.help/5004
     7 |sig {returns(T.class_of(T.any(A, B)))}
                     ^^^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

- The `T.class_of` and `Class` things should be cyan.
- The `Class` should say `Module` or `class or module`
- The error message should detect when the special case of `T.any` / `T.all` of only class types, and suggest distributing the `T.class_of` inside the `T.any` / `T.all`

- - -

neilparikh:
### Motivation

Fixes #4342 

### Test plan

See included automated tests.

### DIFF
--- a/resolver/type_syntax.cc
+++ b/resolver/type_syntax.cc
@@ -463,19 +463,16 @@ ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend
     return sig;
 }
 
-std::optional<std::string> autoCorrectOrOfOnlyClasses(core::Context ctx, core::TypePtr type) {
-    if (auto *o = core::cast_type<core::OrType>(type) ) {
-        auto left = autoCorrectOrOfOnlyClasses(ctx, o->left);
-        auto right = autoCorrectOrOfOnlyClasses(ctx, o->right);
-        if (left && right) {
-            return fmt::format("T.any({}, {})", left.value(), right.value());
-        } else {
-            return {};
-        }
+// This function recurses through an OrType, and accumlates all the class names,
+// wrapped in T.class_of, and checks if the type is only made up of Classes and OrTypes
+bool recurseOrType(core::Context ctx, core::TypePtr type, std::vector<std::string> &v) {
+    if (auto *o = core::cast_type<core::OrType>(type)) {
+        return recurseOrType(ctx, o->left, v) && recurseOrType(ctx, o->right, v);
     } else if (core::isa_type<core::ClassType>(type)) {
-        return fmt::format("T.class_of({})", type.show(ctx.state));
+        v.push_back(fmt::format("T.class_of({})", type.show(ctx)));
+        return true;
     } else {
-        return {};
+        return false;
     }
 }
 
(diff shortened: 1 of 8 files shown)


test/cli/class_of-any/class_of-any.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+main/sorbet --silence-dev-message test/cli/class_of-any/class_of-any.rb 2>&1
```
jez, line 2: Would you mind cargo-culting this test:

https://github.com/sorbet/sorbet/blob/master/test/cli/suggest_t_must/suggest_t_must.sh

So that we can test the autocorrect codepaths as well?

resolver/type_syntax.cc
```diff
@@ -558,7 +574,15 @@ TypeSyntax::ResultType interpretTCombinator(core::Context ctx, const ast::Send &
             auto *obj = ast::cast_tree<ast::ConstantLit>(send.args[0]);
             if (!obj) {
                 if (auto e = ctx.beginError(send.loc, core::errors::Resolver::InvalidTypeDeclaration)) {
-                    e.setHeader("T.class_of needs a Class as its argument");
+                    auto type = getResultTypeWithSelfTypeParams(ctx, send.args.front(), sig, args);
+                    auto autocorrect = autoCorrectOrOfOnlyClasses(ctx, type);
+                    if (core::isa_type<core::OrType>(type) && autocorrect) {
+                        e.setHeader("`{}` needs a class or module as its argument. Apply `{}` to each element of the `{}` instead",
+                                    "T.class_of", "T.class_of", "T.any");
```
jez, line 581: I'm not a fan of double sentences in the header for the same reason as the "no period at the end" ENFORCE.

What do you think about just mentioning the more relevant message?

```suggestion
                        e.setHeader("`{}` must wrap each individual class type, not the outer `{}`", "T.class_of", "T.any");
```

resolver/type_syntax.cc
```diff
@@ -463,6 +463,22 @@ ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend
...
+    if (auto *o = core::cast_type<core::OrType>(type) ) {
+        auto left = autoCorrectOrOfOnlyClasses(ctx, o->left);
+        auto right = autoCorrectOrOfOnlyClasses(ctx, o->right);
+        if (left && right) {
+            return fmt::format("T.any({}, {})", left.value(), right.value());
+        } else {
+            return {};
+        }
+    } else if (core::isa_type<core::ClassType>(type)) {
+        return fmt::format("T.class_of({})", type.show(ctx.state));
```
jez, line 476:
```suggestion
        return fmt::format("T.class_of({})", type.show(ctx));
```

(ctx will be implicitly converted into a GlobalState. no need to type `ctx.state` manually)

resolver/type_syntax.cc
```diff
@@ -463,6 +463,22 @@ ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend
     return sig;
 }
 
+std::optional<std::string> autoCorrectOrOfOnlyClasses(core::Context ctx, core::TypePtr type) {
+    if (auto *o = core::cast_type<core::OrType>(type) ) {
+        auto left = autoCorrectOrOfOnlyClasses(ctx, o->left);
+        auto right = autoCorrectOrOfOnlyClasses(ctx, o->right);
```
jez, line 469: I think that this has the unfortunate side effect of converting this:

```ruby
T.class_of(T.any(A, B, C))
```

into this:

```ruby
T.any(T.class_of(A), T.any(T.class_of(B), T.class_of(C)))
```

when it would be nice if we could turn this directly into:

```ruby
T.any(T.class_of(A), T.class_of(B), T.class_of(C))
```

What do you think about writing it that way?

One way to do this would be to return a vector of either symbols or strings, and then using something like

```cpp
fmt::format("T.any({})", fmt::map_join(symbols, ", ", [&](auto sym) {
  return fmt::format("T.class_of({})", sym.show(gs);
}));
```

jez (commented):
Looks mostly great, one suggestion about whether you want to make the autocorrect even better.

resolver/type_syntax.cc
neilparikh: That approach sounds reasonable to me, not much harder to implement.

jez (approved):

PR #4374, opened 2021-07-16
Break ties to favor then branch, not else branch

jez:
### Motivation

I noticed that we tend to report errors from the `else` branch before the `then`
branch. We can break ties the other way to make it less likely for that to
happen.

The CFG is processed in an approximate forward toposort of the CFG, so in
general it's not always possible or desirable to show errors from earlier lines
before later lines, but breaking ties this way seems to work slightly better.

### Test plan

See commits for before/after diff.


elliottt (approved):
Nice!

jez:

→ https://go/builds/bui_Jrc5cFcdfwwW1m
→ https://go/builds/bui_Jrc5TdCKwhHcQt

PR #4375, opened 2021-07-16
Disallow calling 'new' on certain metatypes

### LINKED ISSUE #4172: Prohibit calling `new` on MetaTypes of unions and type aliases
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0AIntegerOrString%20%3D%20T.type_alias%20%7BT.any(Integer%2C%20String)%7D%0A%0AT.any(Integer%2C%20String).new%0AIntegerOrString.new%0A%0ATypeAliasToInteger%20%3D%20T.type_alias%20%7BInteger%7D%0ATypeAliasToInteger.new)

```ruby
# typed: strict
IntegerOrString = T.type_alias {T.any(Integer, String)}

T.any(Integer, String).new
IntegerOrString.new

TypeAliasToInteger = T.type_alias {Integer}
TypeAliasToInteger.new
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

All of these calls to `.new` should be errors, because it doesn't make sense to call `.new` on a union nor a type alias at runtime.

- - -

This is likely a good first bug, in particular because there is so much prior art to cargo cult here:

#3425, #3588, #3985, #3426, #3428.

Here would be the place to start:

https://github.com/sorbet/sorbet/blob/84d982b45857ff2bfe0a5483ec33f88defa5eb4d/core/types/calls.cc#L1240-L1254

aprocter-stripe:
This makes it an error to call `new` on certain metatypes, like `T.any`, `T.or`, and a few others.

Not quite sure if this approach is as comprehensive as it should be!

### Motivation
Closes #4172

### Test plan
See included automated tests.

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -755,6 +755,9 @@ public:
 
     TypePtr _approximate(const GlobalState &gs, const TypeConstraint &tc) const;
     TypePtr underlying(const GlobalState &gs) const;
+
+private:
+    bool canCallNew(const GlobalState &gs) const;
 };
 CheckSize(MetaType, 16, 8);
 
--- a/core/Types.h
+++ b/core/Types.h
@@ -755,9 +755,6 @@ public:
 
     TypePtr _approximate(const GlobalState &gs, const TypeConstraint &tc) const;
     TypePtr underlying(const GlobalState &gs) const;
-
-private:
-    bool canCallNew(const GlobalState &gs) const;
 };
 CheckSize(MetaType, 16, 8);
 
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -528,8 +528,9 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
         return DispatchResult(Types::untypedUntracked(), std::move(args.selfType), Symbols::noMethod());
     } else if (symbol == Symbols::DeclBuilderForProcsSingleton() && args.name.rawId() == Names::new_().rawId()) {
         if (!args.suppressErrors) {
-            if (auto e = gs.beginError(core::Loc(args.locs.file, args.locs.call), errors::Infer::UnknownMethod)) {
-                e.setHeader("Cannot call `{}` on declaration builder type `{}`", Names::new_().show(gs),
+            if (auto e =
+                    gs.beginError(core::Loc(args.locs.file, args.locs.call), errors::Infer::MetaTypeDispatchCall)) {
+                e.setHeader("Call to method `{}` on `{}` mistakes a type for a value", Names::new_().show(gs),
                             symbol.show(gs));
             }
         }
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1258,7 +1258,11 @@ TypePtr AppliedType::getCallArguments(const GlobalState &gs, NameRef name) const
     return getMethodArguments(gs, klass, name, targs);
 }
 
-bool MetaType::canCallNew(const GlobalState &gs) const {
+namespace {
+
+// Determines whether we will allow `new` on a type wrapped by a `MetaType`. Note that this function is conservative,
+// in that there are some cases we want to reject but cannot detect here.
+bool canCallNew(const GlobalState &gs, const TypePtr &wrapped) {
     if (isa_type<OrType>(wrapped) || isa_type<AndType>(wrapped)) {
         return false;
     }
@@ -1279,17 +1283,24 @@ bool MetaType::canCallNew(const GlobalState &gs) const {
     return true;
 }
 
+} // namespace
+
 DispatchResult MetaType::dispatchCall(const GlobalState &gs, const DispatchArgs &args) const {
     switch (args.name.rawId()) {
         case Names::new_().rawId(): {
-            if (!canCallNew(gs)) {
+            if (!canCallNew(gs, wrapped)) {
                 auto callLoc = core::Loc(args.locs.file, args.locs.call);
                 if (auto e = gs.beginError(callLoc, errors::Infer::MetaTypeDispatchCall)) {
-                    e.setHeader("Cannot call `{}` on type `{}`", Names::new_().show(gs), wrapped.show(gs));
+                    e.setHeader("Call to method `{}` on `{}` mistakes a type for a value", Names::new_().show(gs),
+                                wrapped.show(gs));
+
+                    // For T.class_of(Foo), we'll suggest replacing it with the attached class (Foo).
                     if (auto *appliedType = cast_type<AppliedType>(wrapped)) {
-                        auto receiverLoc = core::Loc(args.locs.file, args.locs.receiver);
-                        e.replaceWith("Replace with class name", receiverLoc, "{}",
-                                      appliedType->klass.data(gs)->attachedClass(gs).show(gs));
+                        if (appliedType->klass.data(gs)->isSingletonClass(gs)) {
+                            auto receiverLoc = core::Loc(args.locs.file, args.locs.receiver);
+                            e.replaceWith("Replace with class name", receiverLoc, "{}",
+                                          appliedType->klass.data(gs)->attachedClass(gs).show(gs));
+                        }
                     }
                 }
                 return DispatchResult(Types::untypedUntracked(), std::move(args.selfType), Symbols::noMethod());
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -526,6 +526,14 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
             }
         }
         return DispatchResult(Types::untypedUntracked(), std::move(args.selfType), Symbols::noMethod());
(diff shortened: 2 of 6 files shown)


aprocter-stripe:

→ https://go/builds/bui_JrQdX6U4URkLqx
→ https://go/builds/bui_JrQeBTSsDFsshs

core/Types.h
```diff
@@ -755,6 +755,9 @@ TYPE(MetaType) final {
 
     TypePtr _approximate(const GlobalState &gs, const TypeConstraint &tc) const;
     TypePtr underlying(const GlobalState &gs) const;
+
+private:
+    bool canCallNew(const GlobalState &gs) const;
```
jez, line 760: This is the only private helper method in all of `Types.h`. Can you move this to being a function in an anonymous namespace in `calls.cc` like all other helper functions?

core/types/calls.cc
```diff
@@ -1227,9 +1235,43 @@ TypePtr AppliedType::getCallArguments(const GlobalState &gs, NameRef name) const
     return getMethodArguments(gs, klass, name, targs);
 }
 
+bool MetaType::canCallNew(const GlobalState &gs) const {
```
jez, line 1238: Do you want to add a comment here about how this is conservative, because there are some cases where we can't tell whether to allow it or not?

core/types/calls.cc
```diff
@@ -1227,9 +1235,43 @@ TypePtr AppliedType::getCallArguments(const GlobalState &gs, NameRef name) const
...
+
+    if (isa_type<ClassType>(wrapped)) {
+        auto sym = cast_type_nonnull<ClassType>(wrapped).symbol;
+        if (sym == Symbols::untyped() || sym == Symbols::bottom()) {
+            return false;
+        }
+    }
+
+    if (auto *appliedType = cast_type<AppliedType>(wrapped)) {
+        if (appliedType->klass.data(gs)->isSingletonClass(gs)) {
```
jez, line 1251: what case is this for?

core/types/calls.cc
```diff
@@ -1227,9 +1235,43 @@ TypePtr AppliedType::getCallArguments(const GlobalState &gs, NameRef name) const
...
+    return true;
+}
+
 DispatchResult MetaType::dispatchCall(const GlobalState &gs, const DispatchArgs &args) const {
     switch (args.name.rawId()) {
         case Names::new_().rawId(): {
+            if (!canCallNew(gs)) {
+                auto callLoc = core::Loc(args.locs.file, args.locs.call);
+                if (auto e = gs.beginError(callLoc, errors::Infer::MetaTypeDispatchCall)) {
+                    e.setHeader("Cannot call `{}` on type `{}`", Names::new_().show(gs), wrapped.show(gs));
```
jez, line 1265: ooc, why did you chose a different error message from below?

```suggestion
                    e.setHeader("Call to method `{}` on `{}` mistakes a type for a value", Names::new_().show(gs),
                                this->wrapped.show(gs));
```

test/testdata/infer/metatype_new.rb
```diff
@@ -0,0 +1,37 @@
...
+T.any(Integer, String).new         # error: Cannot call `new` on type `T.any(Integer, String)`
+T.all(Kernel, Comparable).new      # error: Cannot call `new` on type `T.all(Kernel, Comparable)`
+T.nilable(String).new              # error: Cannot call `new` on type `T.nilable(String)`
+
+T.untyped.new                      # error: Cannot call `new` on type `T.untyped`
+T.self_type.new                    # error: Cannot call `new` on type `T.untyped`
+
+T.class_of(String).new             # error: Cannot call `new` on type `T.class_of(String)`
+T.class_of(MyGeneric).new          # error: Cannot call `new` on type `T.class_of(MyGeneric)`
+T.class_of(MyGeneric[Integer]).new # error: Cannot call `new` on type `T.class_of(MyGeneric)`
```
jez, line 33: I sure hope we don't even parse `T. class_of(MyGeneric[Integer]).new`. Does this test pass? I would have hoped there was another error reported on this line in addition to the `Cannot call new` error.

test/testdata/infer/metatype_new.rb
```diff
@@ -0,0 +1,37 @@
...
+
+T.proc.void.new                    # error: Cannot call `new` on declaration builder type
```
jez, line 35: I didn't see code to handle this case. How does this work?

test/testdata/infer/metatype_new.rb
aprocter-stripe: The test does pass, yeah. :)

Seems that we do parse it:

https://sorbet.run/#class%20MyGeneric%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20X%20%3D%20type_member%0Aend%0A%0AT.class_of(MyGeneric%5BInteger%5D).new

test/testdata/infer/metatype_new.rb
aprocter-stripe: This gets handled in `dispatchCallSymbol`. It doesn't seem to go through `MetaType::dispatchCall`.

test/testdata/infer/metatype_new.rb
jez: looks like we only report the error when it's for sure in type syntax position

https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams(x%3A%20T.class_of(T%3A%3AArray%5BInteger%5D)).void%7D%0Adef%20foo(x)%0A%20%20T.reveal_type(x)%0Aend%0A%0AT%3A%3AArray%5BT.class_of(T%3A%3AArray%5BInteger%5D)%5D.new

We should probably have a separate ticket to make these two cases consistent.

core/types/calls.cc
aprocter-stripe: No particular reason! I think I had it in my head that "`new` isn't a method" but that's, well, wrong. Will change as suggested.

core/types/calls.cc
aprocter-stripe: `T.class_of(String).new`, `T.class_of(MyGeneric).new`, and `T.class_of(MyGeneric[Integer]).new` all are handled here. (Had to comment this out and re-run the test to remind myself. ;))

test/testdata/infer/metatype_new.rb
jez: Sounds good, thanks!

You've found some more bugs:

```ruby
# typed: true
extend T::Sig

fs = T::Array[T.proc.params(x: Integer).void].new
T.reveal_type(fs)

fs.each do |f|
  f.call(0)
  f.call('')
end

# ---

g = T.let(->(x) {}, T.proc.params(x: Integer).void)
T.reveal_type(g)

g.call(0)
g.call('')
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Afs%20%3D%20T%3A%3AArray%5BT.proc.params(x%3A%20Integer).void%5D.new%0AT.reveal_type(fs)%0A%0Afs.each%20do%20%7Cf%7C%0A%20%20f.call(0)%0A%20%20f.call('')%0Aend%0A%0A%23%20---%0A%0Ag%20%3D%20T.let(-%3E(x)%20%7B%7D%2C%20T.proc.params(x%3A%20Integer).void)%0AT.reveal_type(g)%0A%0Ag.call(0)%0Ag.call(''))

The code above and below the line should behave the same.

core/types/calls.cc
jez: Can you add this as a test case?

https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20MyGeneric%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20X%20%3D%20type_template%0Aend%0A%0AMyGeneric.new

(I think that we should allow this code, and it is currently allowed on master.)

aprocter-stripe: Added (but not yet pushed). Seems to work as expected!

test/testdata/infer/metatype_new.rb
aprocter-stripe: You mean `g` (where the `call` is allowed if the arg types match) is the correct behavior and `f` is the one that's bugged, right?

test/testdata/infer/metatype_new.rb
aprocter-stripe: #4377

test/testdata/infer/metatype_new.rb
jez: Yes, the `g` code is correct, and the `f` code should behave like the `g` code.

aprocter-stripe: Does #4378 capture it correctly?

jez: yep!

aprocter-stripe:
@jez I think I've addressed all the comments, so re-requesting review. No rush!

jez (approved):
Thanks!

PR #4379, opened 2021-07-16
website: Promote `T.class_of` doc from fragment

jez:
### Motivation

docs 🎉

### Rendered

localhost_3000_docs_class-of

### DIFF
--- a/website/docs/class-of.md
+++ b/website/docs/class-of.md
@@ -1,28 +1,48 @@
 ---
 id: class-of
-title: T.class_of
+title: Types for Class Objects via T.class_of
+sidebar_label: T.class_of
 ---
 
-> TODO: This page is still a fragment. Contributions welcome!
+Classes are also values in Ruby. Sorbet uses `T.class_of(...)` to describe the
+type of those class objects.
 
 ```ruby
 T.class_of(Integer)
 ```
 
-It can be confusing whether you want `MyClass` or `T.class_of(MyClass)`. For
-reference, these assertions are true:
+The difference between `MyClass` and `T.class_of(MyClass)` can be confusing.
+Here are some examples to make it less confusing:
+
+| These expressions... | ...have these types   |
+| -------------------- | --------------------- |
+| `0`, `1`, `2 + 2`    | `Integer`             |
+| `Integer`            | `T.class_of(Integer)` |
+| `42.class`           | `T.class_of(Integer)` |
+
+Here's a playground link to confirm these types:
 
 ```ruby
-T.let(5, Integer)
+# typed: true
+T.let(0, Integer)
+T.let(1, Integer)
+T.let(2 + 2, Integer)
 
-T.let(5.class, T.class_of(Integer))
 T.let(Integer, T.class_of(Integer))
+T.let(42.class, T.class_of(Integer))
 ```
 
-You can also use modules. It respects inheritance in the same way that
-[Class Types](class-types.md) do:
+<a href="https://sorbet.run/#%23%20typed%3A%20true%0AT.let(0%2C%20Integer)%0AT.let(1%2C%20Integer)%0AT.let(2%20%2B%202%2C%20Integer)%0A%0AT.let(Integer%2C%20T.class_of(Integer))%0AT.let(42.class%2C%20T.class_of(Integer))">
+  → View on sorbet.run
+</a>
+
+## `T.class_of` and inheritance
+
+As with [Class Types](class-types.md#inheritance), `T.class_of` types work with
+inheritance:
 
 ```ruby
+# typed: true
 extend T::Sig
 
 class Grandparent; end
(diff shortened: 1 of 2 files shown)


website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
 
-Gotcha: module ancestor chains (`include`, `extend`) work in a way many people
-don't expect:
+Consider we have some code like this:
 
 ```ruby
-# typed: true
+class MyClass
+  def some_instance_method; end
+  def self.some_class_methodd; end
```
nroman-stripe, line 87: Typo
```suggestion
  def self.some_class_method; end
```

website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
+
+Next, let's explain the other two errors:
+
+```
+❯ MyInterface.singleton_class.ancestors
+=> [#<Class:MyInterface>, T::Private::MixesInClassMethods, T::Helpers, Module, T::Sig, Object, Kernel, BasicObject]
+```
+
+For the `MyInterface` class object, we see that its only ancestor is itself
+(ignoring those common ancestors). Notably, **none** of the classes in this list
```
aprocter-stripe, line 161:
```suggestion
(ignoring common ancestors like `Object`). Notably, **none** of the classes in this list
```

website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
+define either a method called `new` (because `Class` is not there) nor
```
aprocter-stripe, line 162:
```suggestion
defines either a method called `new` (because `Class` is not there) nor
```

website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
+`some_class_method` (because `MyInterface::ClassMethods` is not there).
+
+While these errors are correct, **we want to be able to type this code**. There
+are two options:
+
+1.  Use an abstract class instead of an interface.
+
+    Sometimes this is not possible, because the class in question already has a
+    superclass that can't be changed. However, if this is an option to you, it's
```
aprocter-stripe, line 171:
```suggestion
    superclass that can't be changed. However, if this option is available, it's
```

website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
+❯ MyInterface.singleton_class.ancestors
+=> [#<Class:MyInterface>, T::Private::MixesInClassMethods, T::Helpers, Module, T::Sig, Object, Kernel, BasicObject]
+```
+
+For the `MyInterface` class object, we see that its only ancestor is itself
+(ignoring those common ancestors). Notably, **none** of the classes in this list
+define either a method called `new` (because `Class` is not there) nor
+`some_class_method` (because `MyInterface::ClassMethods` is not there).
+
+While these errors are correct, **we want to be able to type this code**. There
```
aprocter-stripe, line 165:
```suggestion
While these errors are technically correct, **we want to be able to type this code**. There
```

website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
+1.  Use an abstract class instead of an interface.
+
+    Sometimes this is not possible, because the class in question already has a
+    superclass that can't be changed. However, if this is an option to you, it's
+    likely the most straightforward. If we change `MyInterface` to
+    `MyAbstractClass`, all our problems vanish.
+
+2.  Use `T.all(Class, MyInterface::ClassMethods)`.
+
+    This is only a partial solution, but might be good enough.
```
aprocter-stripe, line 177:
```suggestion
    For our example this is only a partial solution, but in many cases it is good enough.
```

website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
+example3(MyClass)             # ok
+```
+
+We’re down to only one error now. The error is still technically correct: since
+we’re using `Class` instead of `T.class_of(...)`, Sorbet has no way to know what
+the instance type created by `x.new` will be (it could be anything), so it
+treats the type as `Object`, causing `some_instance_method` to not be found.
+
+But if this tradeoff is okay, both the top-level call site and the call to
+`x.some_class_method` now typecheck successfully.
```
aprocter-stripe, line 197:
```suggestion
We’re down to only one error now. The error is still technically correct: since
we’re using `Class` instead of `T.class_of(...)`, Sorbet has no way to know what
the instance type created by `x.new` will be (it could be anything), so it
treats the type as `Object`, causing `some_instance_method` to not be found.
However, both the top-level call site to `example3` and the call to `x.some_class_method`
now typecheck successfully. In cases where we don't actually need to use instance methods
from `MyInterface`, this may be an acceptable workaround.
```

website/docs/class-types.md
```diff
@@ -83,6 +83,8 @@ end
 
 ## Inheritance
 
+<!-- id="inheritance" is important -->
```
aprocter-stripe, line 86:
```suggestion
<!-- id="inheritance" is important because other pages link here, so please don't change the section title -->
```

aprocter-stripe (approved):
Looks great! Just a few suggestions for your consideration.

website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
+example1(MyClass)             # ok
+```
+
+This is a class which has an instance method and a class method.
+`T.class_of(MyClass)` works like we expect.
+
+Now imagine that we have a lot of these classes and we want to factor out an
+interface. The straightforward way to do this uses
+[`mixes_in_class_methods`](abstract#interfaces-and-the-included-hook), like
+this:
```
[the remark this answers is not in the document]
aprocter-stripe: Actually I'll retract this suggestion. (The narrative unfolded differently from what I was imagining when I wrote it.)

website/docs/class-of.md
jez: I think this should be `define` (plural) from both my intuition of "does this sound right" and my reading of a handful of grammar articles online.

website/docs/class-of.md
```diff
@@ -1,75 +1,200 @@
...
+example(Child)         # ok
 ```
 
-<a href="https://sorbet.run/#extend%20T%3A%3ASig%0A%0Aclass%20Grandparent%3B%20end%0Aclass%20Parent%20%3C%20Grandparent%3B%20end%0Aclass%20Child%20%3C%20Parent%3B%20end%0A%0Asig%20%7Bparams(x%3A%20T.class_of(Parent)).void%7D%0Adef%20foo(x)%3B%20end%0A%0A%0Afoo(Parent)%20%23%20ok%0Afoo(Child)%20%23%20ok%0A%0Afoo(Grandparent)%20%23%20error">
+<a href="https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20Grandparent%3B%20end%0Aclass%20Parent%20%3C%20Grandparent%3B%20end%0Aclass%20Child%20%3C%20Parent%3B%20end%0A%0Asig%20%7Bparams(x%3A%20T.class_of(Parent)).void%7D%0Adef%20example(x)%3B%20end%0A%0Aexample(Grandparent)%20%20%20%23%20error%0Aexample(Parent)%20%20%20%20%20%20%20%20%23%20ok%0Aexample(Child)%20%20%20%20%20%20%20%20%20%23%20ok">
+  → View on sorbet.run
+</a>
+
+The most surprising feature of `T.class_of` comes from not understanding
+inheritance in Ruby, especially with `include` or `extend` plus modules.
```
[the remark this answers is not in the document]
jez: - `include` and `extend` are not keywords, they're methods.
- I'd prefer not using `/` in written documents like this, because it's ambiguous. `or` is more clear
- We don't use `you` in Sorbet docs, because it leads to a tone like, "you who make mistakes, vs me who is always right" throughout the documentation.

Going to pass on this. Feel free to suggest something else.

website/docs/class-of.md
aprocter-stripe: Works for me!

PR #4382, opened 2021-07-19
Docs: Document 5037

ivoanjo:
Hey, and big thanks for sorbet! 👋 

I'm experimenting with it in [@Datadog's `ddtrace` gem](https://github.com/DataDog/dd-trace-rb) and got a bit confused by error 5037 as I really couldn't see what was wrong.

As I was searching for older issues if anyone had reported this, it turns out that they did, multiple times (sorbet#3888, sorbet#2378, sorbet#1443). So I decided to sent a PR your way to add this to the main error documentation, so that others in the future don't lose time on this.

Hope this is reasonable, and let me know your feedback :)

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -465,6 +465,10 @@ See [5014](#5014). 5036 is the same error as [5014](#5014) but slightly modified
 to allow more common Ruby idioms to pass by in `# typed: true` (5036 is only
 reported in `# typed: strict`).
 
+## 5037
+
+Sorbet does not support aliasing a method from a parent class. See [this issue](https://github.com/sorbet/sorbet/issues/2378) for more details.
+
 ## 5041
 
 Sorbet does not allow inheriting from a class which inherits from `T::Struct`.


jez:
Would you mind adding docs for any/all of these cases too:

- also mentioning the case where the error is straightforward because there was a typo in the method name
- mentioning the case where it can’t find the method because one method is a Singleton class method and one is an instance method
- providing code snippets, so that people can see at a glance whether the error is actually relevant to the code they’re working with

thanks!

froydnj (changes requested):
As @jez said, this needs some changes.

ivoanjo:
Thanks for the feedback. I hadn't considered that simple typos could of course trigger the same issue. I copied over a couple of items from error `#7003` which I think generally talk about the case of "the method doesn't actually exist in the current class, or it exists and Sorbet can't see it".

I then added a code snippet for the aliasing from a parent class.

I did not explicitly call out the singleton example you suggested @jez , as I think the generic text I copied from `#7003` already would cover all such cases, but let me know if you'd still prefer that I explicitly spell out that one case.

jez (approved):
A great improvement. Thanks!

ivoanjo:
Hey! Anything missing on this one?

PR #4384, opened 2021-07-19
Ignore `\r` when parsing the `typed: ...` line

### LINKED ISSUE #1372: Sorbet doesn't correctly handle windows line endings
I have a file called order.rb and order.rbi:

order.rbi:
```ruby
# typed: true

class Order < ApplicationRecord
  extend T::Sig

  sig { returns(T::Array[T::Hash[String, String]]) }
  def changelog(); end

  sig { params(value: T::Array[T::Hash[String, String]]).void }
  def changelog=(value); end
end
```

First line of order.rb:
```ruby
# typed: true
```

When i run type check i get these errors:

```
$ bundle exec srb tc
app/models/order.rbi:6: To use `sig`, this file must declare an explicit `# typed:` sigil (found: none). If you're not sure which one to use, start with `# typed: false` https://srb.help/5038
     6 |  sig { returns(T::Array[T::Hash[String, String]]) }
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

app/models/order.rbi:9: To use `sig`, this file must declare an explicit `# typed:` sigil (found: none). If you're not sure which one to use, start with `# typed: false` https://srb.help/5038
     9 |  sig { params(value: T::Array[T::Hash[String, String]]).void }
```

I am using sorbet-rails and i attempted to do this: https://github.com/chanzuckerberg/sorbet-rails#overriding-generated-signatures

neilparikh:
### Motivation

Fixes #1372 

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/windows-line-endings/windows-line-endings.sh
+++ b/test/cli/windows-line-endings/windows-line-endings.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+file --brief "$(realpath test/cli/windows-line-endings/windows-line-endings.rb)"
+
+main/sorbet --silence-dev-message test/cli/windows-line-endings/windows-line-endings.rb 2>&1
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 *.rbi diff=ruby
 *.md diff=markdown
 test/**/*.out diff
+test/cli/windows-line-endings/windows-line-endings.rb eol=crlf
--- a/core/Files.cc
+++ b/core/Files.cc
@@ -77,6 +77,9 @@ StrictLevel File::fileSigil(string_view source) {
         while (end < source.size() && source[end] != ' ' && source[end] != '\n') {
             ++end;
         }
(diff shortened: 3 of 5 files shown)


aprocter-stripe (approved):
Looks good to me!

test/cli/windows-line-endings/windows-line-endings.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+file --brief "$(realpath test/cli/windows-line-endings/windows-line-endings.rb)"
```
jez, line 3: @neilparikh the realpath is redundant here, right? Can we drop it if so?

neilparikh: `test/cli/windows-line-endings/windows-line-endings.rb` is a symlink, so without `realpath`, `file` prints something like `symbolic link to...` on Linux, rather than the line endings.

PR #4386, opened 2021-07-20
refactor DispatchArgs's Loc management

froydnj:
### Motivation

I got tired of `core::Loc(args.locs.file, args.locs....)` all over the place.  I think this is slightly clearer.

I think it would be even clearer if `DispatchArgs` was renamed to `DispatchContext`, so that the code would read `ctx.callLoc()` or similar, but that renaming would be a follow-on patch.

### Test plan

See included automated tests.


elliottt (approved):
Nice, this is much more readable!

PR #4390, opened 2021-07-22
Check type_member bounds for mixins

### LINKED ISSUE #4389: Type restrictions for `type_member` in Modules not enforced on including class
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Foo%0Aend%0A%0Aclass%20NonFoo%0Aend%0A%0A%0Amodule%20M%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20interface!%0A%0A%20%20X%20%3D%20type_member(upper%3A%20Foo)%0Aend%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20include%20M%0A%0A%20%20X%20%3D%20type_member(fixed%3A%20NonFoo)%0Aend)

```ruby
# typed: true

class Foo
end

class NonFoo
end


module M
  extend T::Sig
  extend T::Generic
  extend T::Helpers

  interface!

  X = type_member(upper: Foo)
end

class A
  extend T::Sig
  extend T::Generic

  include M

  X = type_member(fixed: NonFoo)
end
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

I would expect `NonFoo` to not be allowed as a type_member `X` in `A` as `M` has specified all types `X` must be descendants of `Foo`, which `NonFoo` is not.
- - -

elliottt:
Validate type member bounds when a mixin is the source of the original definition.

### Motivation

Fixes #4389 

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1483,7 +1483,19 @@ class ResolveTypeMembersAndFieldsWalk {
         auto owner = data->owner;
 
         core::LambdaParam *parentType = nullptr;
-        auto parentMember = owner.data(ctx)->superClass().data(ctx)->findMember(ctx, data->name);
+        core::SymbolRef parentMember = core::Symbols::noSymbol();
+        parentMember = owner.data(ctx)->superClass().data(ctx)->findMember(ctx, data->name);
+
+        // check mixins if the type member doesn't exist in the parent
+        if (!parentMember.exists()) {
+            for (auto mixin : owner.data(ctx)->mixins()) {
+                parentMember = mixin.data(ctx)->findMember(ctx, data->name);
+                if (parentMember.exists()) {
+                    break;
+                }
+            }
+        }
+
         if (parentMember.exists()) {
             if (parentMember.isTypeMember()) {
                 parentType = core::cast_type<core::LambdaParam>(parentMember.data(ctx)->resultType);
@@ -1653,11 +1665,6 @@ class ResolveTypeMembersAndFieldsWalk {
             return false;
         }
         if (job.lhs.isTypeMember()) {
-            auto superclass = job.lhs.data(ctx)->owner.data(ctx)->superClass();
-            if (!isGenericResolved(ctx, superclass)) {
-                return false;
-            }
-
             resolveTypeMember(ctx.withOwner(job.owner), job.lhs.asTypeMemberRef(), job.rhs, resolvedAttachedClasses);
         } else {
             resolveTypeAlias(ctx.withOwner(job.owner), job.lhs, job.rhs);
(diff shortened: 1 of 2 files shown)


elliottt:

→ https://go/builds/bui_JthBGzT8kQ8xHp
→ https://go/builds/bui_JthBK9FOmnesS8

elliottt:

→ https://go/builds/bui_JthpT4Bs0rJ6Vz
→ https://go/builds/bui_JthpB6Bg0JsysD

resolver/resolver.cc
```diff
@@ -1653,11 +1665,6 @@ class ResolveTypeMembersAndFieldsWalk {
             return false;
         }
         if (job.lhs.isTypeMember()) {
-            auto superclass = job.lhs.data(ctx)->owner.data(ctx)->superClass();
-            if (!isGenericResolved(ctx, superclass)) {
-                return false;
-            }
-
```
froydnj, line 1660: Why are we OK to delete this?  Is the generic resolution handled by declaring the superclass as a dependency, below, and thus this check is redundant?  Relatedly, I think we can run parts of resolver multiple times, so is it possible that we can come through here without the generic bits being resolved, despite declaring the superclass as a dependency?

And if the answer to the latter question is "yes", wouldn't we have to do the same for mixins now?

resolver/resolver.cc
```diff
@@ -1483,7 +1483,19 @@ class ResolveTypeMembersAndFieldsWalk {
         auto owner = data->owner;
 
         core::LambdaParam *parentType = nullptr;
-        auto parentMember = owner.data(ctx)->superClass().data(ctx)->findMember(ctx, data->name);
+        core::SymbolRef parentMember = core::Symbols::noSymbol();
+        parentMember = owner.data(ctx)->superClass().data(ctx)->findMember(ctx, data->name);
```
froydnj, line 1487: OOC, why do we not traverse the entire inheritance hierarchy here?  Is it not available at this point, or have we just decided that's How Generics Work?

resolver/resolver.cc
```diff
@@ -2005,6 +2012,23 @@ class ResolveTypeMembersAndFieldsWalk {
             }
 
             if (symbol.isClassOrModule()) {
+                // This is the same as the implementation of T::Generic.[] in calls.cc
```
froydnj, line 2015: Non-blocking, but it would be nice if this list was in one and only one place.

froydnj (changes requested):
I am willing to believe that this should work, but I know very little about how generics Should work, and so I have questions before stamping this.

resolver/resolver.cc
elliottt: It's ok to delete this because the check was moved into the dependencies vector instead. Checking the superclass was sort of redundant work, given that putting it in the dependencies vector would allow us to potentially check it fewer times.

resolver/resolver.cc
elliottt: It works inductively: the parent being marked as resolved depends on all of its parents being marked resolved, so we only need to depend on direct ancestors.

resolver/resolver.cc
elliottt: Yep, I agree. There are a few different places that this sort of check is done.

froydnj (approved):
Thanks for the explanations!

PR #4392, opened 2021-07-22
add some ENFORCEs around args and corresponding locs

froydnj:
### Motivation

I added these `ENFORCE`s to debug things and they found an actual problem (4089c9e).  I figure we might as well keep them around permanently.  I've also wanted an actual constructor for `DispatchArgs` for a while and I'm glad I found an excuse to add it!

### Test plan

See included automated tests.


aprocter-stripe (approved):
lgtm!

Issue #3786, opened 2020-12-12
Suggest type_member / type_template if used in the wrong scope


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20TypeMember%20%3D%20type_member%0A%20%20TypeTemplate%20%3D%20type_template%0A%0A%20%20sig%20%7Breturns(TypeTemplate)%7D%0A%20%20def%20foo%3B%20end%0A%0A%20%20sig%20%7Breturns(TypeMember)%7D%0A%20%20def%20self.bar%3B%20end%0Aend)

```ruby
# typed: true
class A
  extend T::Sig
  extend T::Generic

  TypeMember = type_member
  TypeTemplate = type_template

  sig {returns(TypeTemplate)}
  def foo; end

  sig {returns(TypeMember)}
  def self.bar; end
end
```

#### Observed output

```
editor.rb:9: type_template type T.class_of(A)::TypeTemplate used in an instance method definition https://srb.help/5027
     9 |  sig {returns(TypeTemplate)}
                       ^^^^^^^^^^^^

editor.rb:12: type_member type A::TypeMember used in a singleton method definition https://srb.help/5027
    12 |  sig {returns(TypeMember)}
                       ^^^^^^^^^^
Errors: 2
```

#### Expected behavior

The error should suggest changing type_member to type_template or vice versa.

Alternatively, we could just hide this in the docs for 5027, where we could give more nuance, like "it won't work if you want to access the same type from both instance and singleton methods"

- - -

jez:
This was mostly fixed in #4394 (it didn't go so far as to actually implement an autocorrect, but it at least gave a good explanation of why we might not want an autocorrect for a case like this, so absent a concrete user request for one, I'm inclined to agree. Most of the time when people are asking for this, what they're really asking for is to be able to have something like a type_member or a type_template but which can be used from both instance and singleton methods).

PR #4394, opened 2021-07-22
add "Did you mean" prompts for incorrect usage of type_{member,template}

froydnj:
### Motivation

#3786 

Even though there's no autocorrects, I think we could claim #3786 as being fixed.  I wrote out the rationale for avoiding autocorrects in a comment, which might be too much, but it at least serves as a warning to the next person who comes along.

There was a recent conversation about this inside Stripe's Slack, and the suggestion there was to add a pointer to using `type_member` or `type_template` as appropriate.  I think this message qualifies.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/type_syntax.cc
+++ b/resolver/type_syntax.cc
@@ -818,12 +818,30 @@ TypeSyntax::ResultType getResultTypeAndBindWithSelfTypeParams(core::Context ctx,
                             string typeStr = sym.show(ctx);
 
                             if (usedOnSourceClass) {
+                                // Autocorrects here are awkward, because we want to
+                                // offer the autocorrect at the definition of the
+                                // type_member or type_template and we only have
+                                // access to the use of the type_member or
+                                // type_template here.  We could examine the source
+                                // and attempt to identify the location for the
+                                // autocorrect, but that gets messy.
+                                //
+                                // Plus, it's not absolutely clear that the
+                                // definition is really at fault: it might be that
+                                // the user is using the wrong
+                                // type_member/type_template constant for the given
+                                // context, or they need to change the definition of
+                                // the method this use is associated with.  Try to give
+                                // them enough of a hint to decide what to do on their
+                                // own.
                                 if (ctxIsSingleton) {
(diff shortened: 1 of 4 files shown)


resolver/type_syntax.cc
```diff
@@ -818,12 +818,30 @@ TypeSyntax::ResultType getResultTypeAndBindWithSelfTypeParams(core::Context ctx,
...
+                                // the user is using the wrong
+                                // type_member/type_template constant for the given
+                                // context, or they need to change the definition of
+                                // the method this use is associated with.  Try to give
+                                // them enough of a hint to decide what to do on their
+                                // own.
                                 if (ctxIsSingleton) {
                                     e.setHeader("`{}` type `{}` used in a singleton method definition", typeSource,
                                                 typeStr);
+                                    e.addErrorNote("Did you mean to use a similarly-named `type_template`?");
```
aprocter-stripe, line 840: I wonder if it would be better just to say something like:

`Only type_templates can be used in singleton method definitions, not type_members.`

aprocter-stripe (approved):
I think this is an improvement over what we have, so it seems reasonable to me to merge it. I would that an autocorrect is not the right thing, since it's not clear which side of the inconsistency is "wrong". Have a suggestion about the wording of the error message, but I don't feel too strongly about it.

resolver/type_syntax.cc
froydnj: I was just thinking about trying to stuff ", or to use `type_template` in the definition of `{}`", but I think your message is clearer!

aprocter-stripe: Yeah, the first suggestion I started typing was to squeeze both possibilities into the error message, but I couldn't make it brief enough. :)

froydnj: I pushed a slightly different note in 5df84dc that didn't require thinking hard about whether it should be "type_templates" or "`type_template`s".

froydnj:

→ https://go/builds/bui_JuE7ssyEXj7TwB
→ https://go/builds/bui_JuE7Z2wCdyM8ON

PR #4395, opened 2021-07-22
add a note to `@var = T.cast(...)` errors to suggest a way forward

### LINKED ISSUE #4017: Confusing error using `T.cast` on ivar
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20def%20initialize%0A%20%20%20%20%40val%20%3D%20T.cast(0%2C%20Integer)%0A%20%20end%0Aend)

```ruby
# typed: true

class A
  extend T::Sig
  def initialize
    @val = T.cast(0, Integer)
  end
end
```

#### Observed output

```
editor.rb:6: Use T.let to specify the type of constants https://srb.help/5013
     6 |    @val = T.cast(0, Integer)
                   ^^^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

https://srb.help/5013 should mention the workaround for when `T.cast` is intended:

```ruby
val = T.cast(x, X)
@val = val
```

we might also want to use `addErrorNote` on the error message to suggest something like "if you really meant to use `T.cast`, do this...".

froydnj:
### Motivation

Fixes #4017.

We updated the docs for this (https://srb.help/5013), presumably after the original bug was filed.  But it's nice to save somebody a click.  (I don't know how to do an autocorrect that involves temporary variables, so I opted to leave that up to the user -- and they may want to use `T.let` in any event.)

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1731,6 +1731,12 @@ class ResolveTypeMembersAndFieldsWalk {
         } else if (cast->cast != core::Names::let()) {
             if (auto e = ctx.beginError(cast->loc, core::errors::Resolver::ConstantAssertType)) {
                 e.setHeader("Use `{}` to specify the type of constants", "T.let");
+                if (cast->cast == core::Names::cast()) {
+                    e.addErrorNote("If you really want to use `T.cast`, "
+                                   "assign to an intermediate variable first and then "
+                                   "assign that variable to `{}`",
+                                   asgn.lhs.toString(ctx));
+                }
             }
         }
 
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1731,6 +1731,9 @@ class ResolveTypeMembersAndFieldsWalk {
         } else if (cast->cast != core::Names::let()) {
             if (auto e = ctx.beginError(cast->loc, core::errors::Resolver::ConstantAssertType)) {
                 e.setHeader("Use `{}` to specify the type of constants", "T.let");
+                auto rhsLoc = core::Loc(ctx.file, asgn.rhs.loc());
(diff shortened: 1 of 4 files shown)


resolver/resolver.cc
```diff
@@ -1731,6 +1731,12 @@ class ResolveTypeMembersAndFieldsWalk {
         } else if (cast->cast != core::Names::let()) {
             if (auto e = ctx.beginError(cast->loc, core::errors::Resolver::ConstantAssertType)) {
                 e.setHeader("Use `{}` to specify the type of constants", "T.let");
+                if (cast->cast == core::Names::cast()) {
+                    e.addErrorNote("If you really want to use `T.cast`, "
+                                   "assign to an intermediate variable first and then "
+                                   "assign that variable to `{}`",
+                                   asgn.lhs.toString(ctx));
+                }
```
elliottt, line 1739: This seems reasonable to me, but what do you think about adding an autocorrect as well?

elliottt (approved):
Thanks for adding this! Only one non-blocking comment about perhaps making it an autocorrect as well.

resolver/resolver.cc
froydnj: I'm happy to add one.

Do we have a policy on adding autocorrects for things?  My assumption was that we only have autocorrects for things that are "obviously" the right thing to do, but adding an autocorrect here is more like "probably" the right thing to do.

froydnj:

→ https://go/builds/bui_JvLZ4Wj6h8w77c
→ https://go/builds/bui_JvLZwLBQWQTRhp

PR #4396, opened 2021-07-22
Fix type signature for Kernel#select to match IO.select.

richardxia:
This PR copies the Sorbet type signature for `IO.select` intto `Kernel#select`, since they are the same method in Ruby but Sorbet's types for the latter are incorrect.

### Motivation

[`Kernel#select`](https://ruby-doc.org/core-2.7.2/Kernel.html#method-i-select)'s type signature was incorrect, as it should return `Array[Array[IO]]`, not `Array[String]`. There is no way that `Array[String]` could ever have been correct, since this method always returns a 3-element array, where the first element is the array of IO objects that are ready for reading, the second element is the ones that are ready for writing, and the third is the ones that have encountered errors.

The same method exists on [`IO`](https://ruby-doc.org/core-2.7.2/IO.html#method-c-select), and the documentation is identical, so I believe they are the same method in Ruby. The Sorbet type signature for `IO.select` [has the correct return type](https://github.com/sorbet/sorbet/blob/e45dc6a2b594b9b8fee93d2582f92a4a4d71d6ba/rbi/core/io.rbi#L2342-L2351), and it is also more precise about the nilability of its arguments as well as the `timeout` parameter's specific types.

### Test plan

I just ran all the tests described in the README with `bazel test //... --config=dbg`, since I wasn't sure if it was even possible to write tests for the accuracy of the rbi type signatures.

In addition, in my own, non-open source project where I am using Sorbet, I created an `.rbi` file that overrides the definition of `Kernel#select` and I confirmed that it works for me.


aprocter-stripe:

→ https://go/builds/bui_JtzVK6vlEZSlfZ
→ https://go/builds/bui_JtzVKWVC012DjJ

aprocter-stripe (changes requested):
Thanks for the fix! Could you please run `tools/scripts/update_exp_files.sh` and commit/push the changes it makes? That should fix the test errors. Everything else looks good.

richardxia:
Yep, thanks! Just committed and pushed up the changes.

richardxia:
I see that I'm still failing some of the tests. I'm wondering if some of the files that `tools/scripts/update_exp_files.sh` changed shouldn't have been committed in the first place? Let me revert everything except for the changes to `test/cli/autocorrect-helpers/autocorrect-helpers.out`.

aprocter-stripe:
@richardxia Yup, that should be the fix (just committing the `autocorrect-helpers.out`). Sorry for the hassle.

richardxia:
No problem! Just pushed up a new commit that reverts everything but `autocorrect-helpers.out`.

aprocter-stripe (approved):

PR #3309, opened 2020-07-21
runtime: Best-effort one T::Types::Simple instance per module & one T.nilable(<mod>) per module

djudd-stripe:
The basic idea here is:
1. For `T::Types::Simple`, we cache an instance on the module that we wrap, as an instance variable. We prefer to use this instance rather than construct a new one when using standard helpers.
2. For `T::Types::Union` uses that are a simple union of NilClass and a T::Types::Simple (ie, a simple T.nilable use), we cache an instance on the corresponding (cached) T::Types::Simple instance. We prefer to use this instance rather than construct a new one when using standard helpers.

I also added simple microbenchmarks for runtime typechecking.

Notably this does not _guarantee_ that only one equivalent instance of a T::Types::Simple or simple-nilable T::Union can exist. There are a couple of reasons that's difficult:
1. For T::Types::Simple, modules can be frozen. So we'd have to do the caching in something like a global Hash keyed by module instead. That'd be a little more complex and also I worry that would perform worse due to worse memory locality, although I admit I haven't tested.
2. For T::Types::Union, catching all possible ways that a simple-nilable union could define would require running after the `flat_map` and `uniq` logic currently in `initialize`, and I couldn't find a way of doing that which didn't reverse any gains in the benchmarks.

Since `new` on `T::Types::Simple` and `T::Types::Union` is public, for backwards compatibility we'd also have to override `new` instead of just adding a wrapper. That's doable but seems smellier.

I could be persuaded to reconsider the above reasoning and try to actually offer a uniqueness guarantee. It'd have the advantage of giving us cheaper equality and hash implementations. But those aren't hot paths, are they?

Another alternative would be to just hardcode constants for a handful of the most commonly-wrapped classes, like Integer, String, Float, Symbol etc. That seems like it would end up with less benefit for just as much complexity, though.

### Motivation
This is aiming to help with two things:
1. Total memory usage from `sig` definitions in a large codebase;
2. Runtime allocations from `T.let` uses (where type definitions aren't necessarily restricted to load-time like they are for `sig`).

I don't have memory usage data currently, but benchmarks before:
```
T.let(..., Integer) twice, μs/iter
  0.930970   0.000963   0.931933 (  0.932905)
sig {params(x: Integer).void} twice, μs/iter
  1.084934   0.001938   1.086872 (  1.087867)
T.let(..., T.nilable(Integer)) twice, μs/iter
  5.485798   0.003305   5.489103 (  5.491234)
[...]

PR #4397, opened 2021-07-22
Change type caching to stop polluting modules

### LINKED ISSUE #4391: Stack level too deep (SystemStackError) when calling String.to_json after T::Utils.run_all_sig_blocks
#### Input

Can be reproduced with the following:
On a rails app w/ Sorbet Runtime, open a rails console, call `T::Utils.run_all_sig_blocks`, followed by `String.to_json`

```ruby
irb(main):001:0> T::Utils.run_all_sig_blocks
=> nil
irb(main):002:0> String.to_json
/Users/ryanbrushett/.gem/ruby/3.0.1/gems/activesupport-6.1.4/lib/active_support/core_ext/object/instance_variables.rb:15:in `instance_values': stack level too deep (SystemStackError)
```

#### Observed output

```
/Users/ryanbrushett/.gem/ruby/3.0.1/gems/activesupport-6.1.4/lib/active_support/core_ext/object/instance_variables.rb:15:in `instance_values': stack level too deep (SystemStackError)
```

#### Expected behavior

Something like this:

```ruby
irb(main):001:0> String.to_json
=> "{}"
```

## Troubleshooting

The exception shows that the problem gets thrown from `activesupport-6.1.4/lib/active_support/core_ext/object/instance_variables.rb:15` so, if we add a breakpoint there:

```ruby
  def instance_values
    require 'pry'
    binding.pry
    Hash[instance_variables.map { |name| [name[1..-1], instance_variable_get(name)] }]
  end
```

We can see that `instance_variables` will be `[:@__as_sorbet_type]`

```ruby
[1] pry(String)> instance_variables
=> [:@__as_sorbet_type]
```

Letting execution continue, we immediately hit the breakpoint again. `instance_variables` is now `[:@raw_type, :@name, :@nilable]`

```ruby
[1] pry(#<T::Types::Simple>)> instance_variables
=> [:@raw_type, :@name, :@nilable]
```

If you do `instance_variable_get(:@raw_type)`, the result is `String` and there's our recursion.

It looks like it's an issue with storing things on the type itself. Interesting.

cc @paracycle 

```ruby
[3] pry(#<T::Types::Simple>)> instance_variable_get(:@raw_type)
=> String
```

<details>
  <summary><b>Click to see my full shell Output</b></summary>

```ruby
Running via Spring preloader in process 57947
Loading development environment (Rails 6.1.4)
irb(main):001:0> T::Utils.run_all_sig_blocks
=> nil
irb(main):002:0> String.to_json

From: /Users/ryanbrushett/.gem/ruby/3.0.1/gems/activesupport-6.1.4/lib/active_support/core_ext/object/instance_variables.rb:16 Object#instance_values:

    14: def instance_values
    15:   require 'pry'
 => 16:   binding.pry
    17:   Hash[instance_variables.map { |name| [name[1..-1], instance_variable_get(name)] }]
    18: end

[1] pry(String)> instance_variables
=> [:@__as_sorbet_type]
[2] pry(String)> exit

From: /Users/ryanbrushett/.gem/ruby/3.0.1/gems/activesupport-6.1.4/lib/active_support/core_ext/object/instance_variables.rb:16 Object#instance_values:

    14: def instance_values
    15:   require 'pry'
 => 16:   binding.pry
    17:   Hash[instance_variables.map { |name| [name[1..-1], instance_variable_get(name)] }]
    18: end

[1] pry(#<T::Types::Simple>)> instance_variables
=> [:@raw_type, :@name, :@nilable]
[2] pry(#<T::Types::Simple>)> exit

From: /Users/ryanbrushett/.gem/ruby/3.0.1/gems/activesupport-6.1.4/lib/active_support/core_ext/object/instance_variables.rb:16 Object#instance_values:

    14: def instance_values
    15:   require 'pry'
 => 16:   binding.pry
    17:   Hash[instance_variables.map { |name| [name[1..-1], instance_variable_get(name)] }]
    18: end

[1] pry(String)> instance_variables
=> [:@__as_sorbet_type]
[2] pry(String)> exit

From: /Users/ryanbrushett/.gem/ruby/3.0.1/gems/activesupport-6.1.4/lib/active_support/core_ext/object/instance_variables.rb:16 Object#instance_values:

    14: def instance_values
    15:   require 'pry'
 => 16:   binding.pry
    17:   Hash[instance_variables.map { |name| [name[1..-1], instance_variable_get(name)] }]
    18: end

[1] pry(#<T::Types::Simple>)> instance_variables
=> [:@raw_type, :@name, :@nilable]
[2] pry(#<T::Types::Simple>)> instance_variable_get(:@name)
=> "String"
[3] pry(#<T::Types::Simple>)> instance_variable_get(:@raw_type)
=> String
[4] pry(#<T::Types::Simple>)>

```

</details>

paracycle:
Fixes #4391

### Motivation

Since #3309 was merged, whenever a module is coerced to a Sorbet type an instance variable is created on the original module to cache the created `T::Types::Simple` instance.

However, that action ends up adding instance variables to types not under Sorbet runtime's control. Even worse, the value that is added has a reference to the module, so any operation applied to the parent module that walks over instance variables recursively would end up in an infinite recursion. This is already happening in #4391 for JSON serialization via ActiveSupport.

The fix is to store the cached values on a `WeakMap` on the `Pool` class. This ensures that the cached values can be garbage collected, stops polluting modules and does not seem to be negatively impacting the
performance either.

##### Before:
```
Vanilla Ruby method call: 20.955 ns
Vanilla Ruby is_a?: 27.88 ns
T::Types::Simple#valid?: 54.84 ns
T::Types::Union#valid?: 195.672 ns
T.let(..., Integer): 411.705 ns
sig {params(x: Integer).void}: 439.973 ns
T.let(..., T.nilable(Integer)): 1.25 μs
sig {params(x: T.nilable(Integer)).void}: 590.034 ns
T.let(..., Example): 372.786 ns
sig {params(x: Example).void}: 401.277 ns
T.let(..., T.nilable(Example)): 1.325 μs
sig {params(x: T.nilable(Example)).void}: 582.362 ns
sig {params(s: Symbol, x: Integer, y: Integer).void} (with kwargs): 2.248 μs
.bind(example).call: 202.877 ns
```

##### After:
```
Vanilla Ruby method call: 19.429 ns
Vanilla Ruby is_a?: 25.425 ns
T::Types::Simple#valid?: 49.577 ns
T::Types::Union#valid?: 155.493 ns
T.let(..., Integer): 399.602 ns
sig {params(x: Integer).void}: 569.334 ns
T.let(..., T.nilable(Integer)): 1.313 μs
sig {params(x: T.nilable(Integer)).void}: 550.468 ns
T.let(..., Example): 432.728 ns
sig {params(x: Example).void}: 351.026 ns
T.let(..., T.nilable(Example)): 1.298 μs
sig {params(x: T.nilable(Example)).void}: 559.341 ns
sig {params(s: Symbol, x: Integer, y: Integer).void} (with kwargs): 2.196 μs
.bind(example).call: 209.101 ns
```
### Test plan

See included automated tests.

/cc @djudd-stripe


gems/sorbet-runtime/lib/types/types/simple.rb
```diff
@@ -60,7 +62,7 @@ def self.type_for_module(mod)
             Simple.new(mod)
           end
 
-          mod.instance_variable_set(:@__as_sorbet_type, type) unless mod.frozen?
+          @cache[mod] = type unless mod.frozen?
```
djudd, line 65: We don't need the `frozen?` check here anymore

djudd (commented):
I'm a little surprised at the benchmark results since when I benchmarked this the performance of using a hash was worse. Maybe machine/architecture dependent? But the concerns with the current implementation are compelling enough to me that I'd be fine sacrificing the marginal performance anyway FWIW.

gems/sorbet-runtime/lib/types/types/simple.rb
paracycle: I thought so initially as well, but then the `frozen module` test failure reminded me why we still need it. Internally, `WeakMap` adds a finalizer to the stored object, so that it can clear that map entry when the object is garbage collected. For a frozen object, though, adding a finalizer is not a valid operation, so this still raises.

djudd: Huh, weird. That could use a comment probably?

froydnj (changes requested):
Please add the comment that djudd requested.  Otherwise, this seems totally reasonable!

gems/sorbet-runtime/lib/types/types/simple.rb
paracycle: Sorry for the delay here, just added the requested comment.

froydnj (approved):
Thank you!

PR #4398, opened 2021-07-22
Upgrade spdlog to version 1.9.0

elliottt:
Upgrade spdlog to 1.9.0, which includes fmt-8.0.

The update to fmt-8.0 made this a little bit of a noisy upgrade, as they deprecated an overload for `fmt::format_to` that we were relying on. The solution is to apply `std::back_inserter` to the first argument, but that made the diff pretty big.

### Motivation

We were on version 1.3.1, which was released in January 2019.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -400,7 +400,7 @@ namespace {
 void printTabs(fmt::memory_buffer &to, int count) {
     int i = 0;
     while (i < count) {
-        fmt::format_to(to, "  ");
+        fmt::format_to(std::back_inserter(to), "  ");
         i++;
     }
 }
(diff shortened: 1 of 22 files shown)


elliottt:

→ https://go/builds/bui_JtzYIFMfT8DLlf
→ https://go/builds/bui_JtzYP5C9cfAvtX

aprocter-stripe (approved):
`std::back_inserter(lgtm)`

(Some context from Slack: `fmt` removed an overload for `format_to` that automatically adds these `back_inserter` calls, so here we are.) (EDIT: oh, you even said that in the PR comment :sweat_smile:)

PR #4400, opened 2021-07-23
provide more precise location for (most) kwargs

### LINKED ISSUE #4387: location information for mistyped keyword arguments is too broad
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams(package%3A%20Integer%2C%20name%3A%20Integer).void%7D%0A%20%20def%20initialize(package%3A%2C%20name%3A)%0A%20%20end%0A%0A%20%20def%20self.foo%0A%20%20%20%20pkg%20%3D%20T.let(nil%2C%20T.nilable(Integer))%0A%0A%20%20%20%20A.new(package%3A%20pkg%2C%20name%3A%20pkg)%0A%20%20end%0Aend%0A)

```ruby
# typed: true
class A
    extend T::Sig

  sig {params(package: Integer, name: Integer).void}
  def initialize(package:, name:)
  end

  def self.foo
    pkg = T.let(nil, T.nilable(Integer))

    A.new(package: pkg, name: pkg)
  end
end
```

#### Observed output

```
editor.rb:12: Expected Integer but found T.nilable(Integer) for argument package https://srb.help/7002
    12 |    A.new(package: pkg, name: pkg)
                  ^^^^^^^^^^^^^^^^^^^^^^^
  Expected Integer for argument package of method A#initialize:
    editor.rb:5:
     5 |  sig {params(package: Integer, name: Integer).void}
                      ^^^^^^^
  Got T.nilable(Integer) originating from:
    editor.rb:12:
    12 |    A.new(package: pkg, name: pkg)
                  ^^^^^^^^^^^^^^^^^^^^^^^

editor.rb:12: Expected Integer but found T.nilable(Integer) for argument name https://srb.help/7002
    12 |    A.new(package: pkg, name: pkg)
                  ^^^^^^^^^^^^^^^^^^^^^^^
  Expected Integer for argument name of method A#initialize:
    editor.rb:5:
     5 |  sig {params(package: Integer, name: Integer).void}
                                        ^^^^
  Got T.nilable(Integer) originating from:
    editor.rb:12:
    12 |    A.new(package: pkg, name: pkg)
                  ^^^^^^^^^^^^^^^^^^^^^^^
Errors: 2
```

#### Expected behavior

The location shown for the error and for the actual types (`Got $TYPE originating from`) of each keyword argument should be narrower, e.g.:

```
editor.rb:12: Expected Integer but found T.nilable(Integer) for argument package https://srb.help/7002
    12 |    A.new(package: pkg, name: pkg)
                  ^^^^^^^^^^^^
  Expected Integer for argument package of method A#initialize:
    editor.rb:5:
     5 |  sig {params(package: Integer, name: Integer).void}
                      ^^^^^^^
  Got T.nilable(Integer) originating from:
    editor.rb:12:
    12 |    A.new(package: pkg, name: pkg)
                           ^^^

editor.rb:12: Expected Integer but found T.nilable(Integer) for argument name https://srb.help/7002
    12 |    A.new(package: pkg, name: pkg)
                                ^^^^^^^^^
  Expected Integer for argument name of method A#initialize:
    editor.rb:5:
     5 |  sig {params(package: Integer, name: Integer).void}
                                        ^^^^
  Got T.nilable(Integer) originating from:
    editor.rb:12:
    12 |    A.new(package: pkg, name: pkg)
                                      ^^^
Errors: 2
```
- - -

This is an offshoot of #3498.

froydnj:
### Motivation

Fixes #4387, which makes wrong kwargs in the IDE autocorrectable, etc.

### Test plan

See included automated tests.  I did not include the testcase from #4387 because I felt that the existing testcases covered everything already.  For reference, the new output is:

```
froydnj@pyro:~/src/sorbet$ ./bazel-bin/main/sorbet 4387.rb
4387.rb:12: Expected Integer but found T.nilable(Integer) for argument package https://srb.help/7002
    12 |    A.new(package: pkg, name: pkg)
                           ^^^
  Expected Integer for argument package of method A#initialize:
    4387.rb:5:
     5 |  sig {params(package: Integer, name: Integer).void}
                      ^^^^^^^
  Got T.nilable(Integer) originating from:
    4387.rb:12:
    12 |    A.new(package: pkg, name: pkg)
                           ^^^
    4387.rb:12: Replace with T.must(pkg)
    12 |    A.new(package: pkg, name: pkg)
                           ^^^

4387.rb:12: Expected Integer but found T.nilable(Integer) for argument name https://srb.help/7002
    12 |    A.new(package: pkg, name: pkg)
                                      ^^^
  Expected Integer for argument name of method A#initialize:
    4387.rb:5:
     5 |  sig {params(package: Integer, name: Integer).void}
                                        ^^^^
  Got T.nilable(Integer) originating from:
    4387.rb:12:
    12 |    A.new(package: pkg, name: pkg)
                                      ^^^
    4387.rb:12: Replace with T.must(pkg)
    12 |    A.new(package: pkg, name: pkg)
                                      ^^^
Errors: 2
```

The major difference in this output from the "expected result" output in the report in #4387 is that the initial error ("Expected Integer but found...") in the "expected result" would highlight both the keyword and the value in the call ("package: pkg") rather than just the value ("pkg", as above)

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -826,7 +826,10 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
             }
         }
 
-        // merge in the keyword splat argument if it's present
+        // Merge in the keyword splat argument if it's present.
+        // Note that we rely on this merging being done after the explicitly-named
+        // arguments in the call when determining whether keyword args are properly
+        // typed, below.
         bool kwSplatIsHash = false;
         TypePtr kwSplatType;
         if (hasKwsplat) {
@@ -933,6 +936,8 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
     // keep this around so we know which keyword arguments have been supplied
     UnorderedSet<NameRef> consumed;
     if (hasKwargs) {
+        // Remember where the kwargs started
+        auto kwait = aPosEnd;
         // Mark the keyword args as consumed
         ait += numKwargs;
 
@@ -945,6 +950,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
             }
             pend = kwit;
 
+            bool sawKwSplat = false;
             while (kwit != data->arguments().end()) {
                 const ArgInfo &spec = *kwit;
                 if (spec.flags.isBlock) {
@@ -976,6 +982,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
                             result.main.errors.emplace_back(std::move(e));
                         }
                     }
+                    sawKwSplat = true;
                     break;
                 }
                 ++kwit;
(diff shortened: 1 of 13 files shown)


froydnj:

→ https://go/builds/bui_JuENWXuRV6EZkh
→ https://go/builds/bui_JuEN8nFRMw92g8

core/types/calls.cc
```diff
@@ -995,12 +1002,42 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                // We want the precise location of the kwarg that we are processing.
+                // To do that, we rely on hash->{keys,values} being added in the order
+                // that the kwargs appear in the the call.  We also rely on args from
+                // a shapely-typed kwsplat hash (implicit or explicit) being added to
+                // hash->{keys,values} after the explicitly-named kwargs in the call.
                 auto offset = arg - hash->keys.begin();
+                // offset * 2 skips kwargs we have already seen.
+                // + 1 moves us to the value being passed.
+                ENFORCE(!sawKwSplat);
+                auto kwaValue = kwait + (offset * 2) + 1;
```
elliottt, line 1014:
```suggestion
                auto kwaValue = kwait + nonPosArgs;
```
The number of non-positional arguments is named right before positional arguments are processed.

core/types/calls.cc
```diff
@@ -995,12 +1002,42 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                TypeAndOrigins tpe;
+                ENFORCE(args.args.size() == args.locs.args.size(), "what?!");
```
elliottt, line 1016:
```suggestion
                ENFORCE(args.args.size() == args.locs.args.size(), "Send arg and loc vectors did not match in length");
```

core/types/calls.cc
```diff
@@ -995,12 +1002,42 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                // to check whether the thing we're pointing at lives inside the range
+                // of passed kwargs in the call.
+                //
+                // If we encounter this case, we've already setup kwargsLoc to reflect
+                // as closely as possible the location of the entire shapely-typed
+                // hash, so use that for the "origin" of the argument and say that the
+                // argument has no location.  The latter also inhibits type-driven
+                // autocorrects from kicking in.
+                bool argInBounds = kwaValue < ait;
+                // Note: it's OK to compute with kwaValue even if it's out-of-bounds.
```
elliottt, line 1033: Because it's a random-access iterator?

elliottt (approved):
This is great! The changes to the suggest-unsafe CLI test output are super exciting!

core/types/calls.cc
froydnj: Sorry, I was being unclear here.  What this comment really intends to point out is that it's OK to compute a pointer (or an iterator, more generally) that's off the end of `args.args` and even to do pointer arithmetic on it to get an integer offset from the beginning of `args.args` -- so long as you don't actually dereference the memory at the pointer, or the memory that would live at `args.locs.args.begin() + kwargOffset` (which is effectively what `args.argLoc(kwargOffset)` is going to do in a line or two).  But expressing all of that is kind of a mouthful, so maybe we're better off dropping this note?

core/types/calls.cc
froydnj: `kwait` points to the beginning of the keyword arguments (it's not modified at all during this loop), so `kwait + nonPosArgs` would always point at the same thing, namely the argument just after all the keyword arguments, which I don't think is what we want?

core/types/calls.cc
froydnj: Whoops, sorry, I should have made this a better message.

core/types/calls.cc
elliottt: Sure, that works for me. I think the reasoning about why it's safe makes sense 👍

core/types/calls.cc
elliottt: Ah, for some reason I totally read past the fact that it's using `offset` in the calculation of `kwaValue`! You can ignore this suggestion :)

PR #4406, opened 2021-07-27
Add documentation for 3001 errors

Morriar:
image

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -53,6 +53,47 @@ and should not be reported.
 
 [#1993]: https://github.com/sorbet/sorbet/pull/1993
 
+## 3001
+
+Sorbet doesn’t support singleton definitions outside of the class itself:
+
+```rb
+class << MyClass # error: `class << EXPRESSION` is only supported for `class << self`
+  def foo
+    # ...
+  end
+end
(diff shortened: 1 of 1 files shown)


website/docs/error-reference.md
```diff
@@ -53,6 +53,47 @@ and should not be reported.
...
+Sometimes, `EXPRESSION` is not a constant literal like in what follows:
+
+```rb
+class << some_variable
+  include Foo
+end
+```
+
+In this case, it is possible to directly call `include` on the the singleton
+class of `EXPRESSION`, but it should be done with **utmost caution**, as Sorbet
```
Morriar, line 89: The `utmost caution` part is copied from https://sorbet.org/docs/error-reference#4002

froydnj: Thanks for flagging this.

Nitpicking, I think we should revisit the language here and in the 4002 documentation, "far less accurate predictions" sounds vaguely like Sorbet is consulting a crystal ball -- maybe something like "make Sorbet's analyses far less accurate" would be better?  Regardless, this is followup material.

froydnj (approved):
Thank you!

jez:
Thanks!

PR #4409, opened 2021-07-27
Interpolate colors in more errors

jez:
### Motivation

Consistency

### Test plan

**Before**
Screen Shot 2021-07-27 at 11 07 20 AM
**After**
Screen Shot 2021-07-27 at 11 05 59 AM


froydnj (approved):
I now understand why you would format-interpolate things rather than having them directly in the format string!

jez:
> I now understand why you would format-interpolate things rather than having them directly in the format string!

Yeah, it's a gross hack that looks for the exact string 
```
`{}`
```
and replaces it with the start and end color codes if possible 😅

PR #4412, opened 2021-07-28
Add the Sorbet Compiler

elliottt:
Add the sorbet compiler.

### Test plan

See included automated tests.


elliottt:

→ https://go/builds/bui_JwZ94WPn4PdgFS
→ https://go/builds/bui_JwZ9vCeAR9mSgl

elliottt:

→ https://go/builds/bui_JwagymeSKCTxbG
→ https://go/builds/bui_Jwagy7cpkJypbd

third_party/zlib.BUILD
```diff
@@ -1,16 +1,65 @@
-# From https://github.com/bazelbuild/bazel/blob/master/third_party/zlib/BUILD
+# Originally from https://github.com/protocolbuffers/protobuf/blob/2514f0bd7da7e2af1bed4c5d1b84f031c4d12c10/third_party/zlib.BUILD
```
jez: out of curiosity, why did this zlib.BUILD file change?

third_party/sorbet_test_corpus_runner.patch
```diff
@@ -1,13 +0,0 @@
-diff --git a/test/BUILD b/test/BUILD
```
jez: Q: This patch file isn't needed anymore because we can just run the test files directly?

third_party/ruby/ruby.BUILD
```diff
@@ -1,23 +0,0 @@
-# vim: ft=bzl sw=4 ts=4 et
```
jez, line 1: And this file becomes unnecessary because of the `ruby_unpatched.BUILD` file?

third_party/ruby/ruby-versions.txt
```diff
@@ -1,2 +1,2 @@
-sorbet_ruby
+sorbet_ruby_2_6
```
jez, line 1: This will technically affect what artifacts we build and publish in Stripe's CI, right? I think this means we have to be sure that we don't mention `@sorbet_ruby` anywhere in Stripe's codebase anymore (which I think is correct, but I figured I'd call it out).

test/run_ruby.sh
```diff
@@ -21,7 +21,7 @@ if [ -z "$rb_file" ]; then
 fi
 
 ruby="./bazel-bin/external/sorbet_ruby_2_7/toolchain/bin/ruby"
-sorbet_runtime="./bazel-sorbet_llvm/external/com_stripe_ruby_typer/gems/sorbet-runtime/lib/sorbet-runtime.rb"
+sorbet_runtime="./gems/sorbet-runtime/lib/sorbet-runtime.rb"
```
jez, line 24: 👍

test/patch_require.rb
```diff
@@ -19,7 +19,7 @@ def require(name)
       case RUBY_PLATFORM
       when "x86_64-linux"
         suffix = '.so'
-      when "x86_64-darwin18", "x86_64-darwin19"
+      when "x86_64-darwin18", "x86_64-darwin19", "x86_64-darwin20"
```
jez, line 22: Where does this extra `darwin20` come from?

elliottt: Running tests locally on Big Sur :D

plugin_injector/BUILD
```diff
@@ -3,6 +3,7 @@ cc_library(
     srcs = [
         "plugin_injector.cc",
     ],
+    defines = ["SORBET_COMPILER"],
```
jez, line 6: Do we need this variable still?

third_party/ruby/ruby-versions.txt
elliottt: When we upgraded to ruby-2.7 we stripped out all references to the original sorbet_ruby target. I'll grep around and make sure though.

third_party/ruby/ruby.BUILD
elliottt: Yes, exactly. I favored our internal version of the ruby build scripts, as they were more up to date.

third_party/sorbet_test_corpus_runner.patch
elliottt: Yep, no need to patch them to use the plugin injector now.

gems/sorbet/test/snapshot/BUILD
```diff
@@ -23,10 +23,10 @@ sh_binary(
...
         "//main:sorbet",
         "@gems//gems",
-        "@ruby_2_6//:bundle",
-        "@ruby_2_6//:ruby",
-        "@ruby_2_7//:bundle",
-        "@ruby_2_7//:ruby",
+        "@sorbet_ruby_2_6//:bundle",
+        "@sorbet_ruby_2_6//:ruby",
+        "@sorbet_ruby_2_7//:bundle",
+        "@sorbet_ruby_2_7//:ruby",
```
jez, line 29: Is it worth trying to run these tests with an unpatched version of ruby, instead of sorbet_ruby?

gems/sorbet/test/hidden-method-finder/update_hidden_methods_exp.sh
```diff
@@ -7,9 +7,9 @@ base_dir="$(pwd)"
 cd ../../../..
 # we are now at the repo root
 
-versions=("ruby_2_6" "ruby_2_7")
+versions=("sorbet_ruby_2_6" "sorbet_ruby_2_7")
 
-if ! bazel test //gems/sorbet/test/hidden-method-finder -c opt "$@"; then
+if ! bazel test //gems/sorbet/test/hidden-method-finder --define compiler=false -c opt "$@"; then
```
jez, line 12: I don't think we need `compiler=false` here, right?

third_party/zlib.BUILD
elliottt: We had two versions committed, one in sorbet_llvm and one in sorbet. I standardized on the one that built successfully with the sanitizer on, though that might not be necessary now that we're running the sorbet_llvm tests completely separately.

plugin_injector/BUILD
elliottt: We do not, good call.

gems/sorbet/lib/serialize.rb
```diff
@@ -223,9 +223,9 @@ def constant(const, value)
     if KEYWORDS.include?(const.to_sym)
       return "# Illegal constant name: #{const}"
     end
-    if defined?(T::Types) && Sorbet::Private::RealStdlib.real_is_a?(value, T::Types::TypeMember)
+    if defined?(T::Types::TypeMember) && Sorbet::Private::RealStdlib.real_is_a?(value, T::Types::TypeMember)
       value.variance == :invariant ? "  #{const} = type_member" : "  #{const} = type_member(#{value.variance.inspect})"
-    elsif defined?(T::Types) && Sorbet::Private::RealStdlib.real_is_a?(value, T::Types::TypeTemplate)
+    elsif defined?(T::Types::TypeTemplate) && Sorbet::Private::RealStdlib.real_is_a?(value, T::Types::TypeTemplate)
```
jez, line 228: Why does this change show up?

gems/sorbet/test/snapshot/BUILD
elliottt: Possibly? I don't think that the patching changes the build time dramatically, and our goal has been that there aren't any changes to non-compiled code-paths with our patches.

gems/sorbet/test/hidden-method-finder/update_hidden_methods_exp.sh
elliottt: Yep, good catch.

.buildkite/test-compiler.sh
```diff
@@ -23,33 +23,27 @@ source .buildkite/tools/setup-bazel.sh
...
+  --test_summary=terse \
+  --test_output=errors || err=$?
 
 echo "--- uploading test results"
 
 rm -rf _tmp_
 mkdir -p _tmp_/log/junit/
 
-./bazel query '(tests(//...) + tests(@com_stripe_ruby_typer//test)) except attr("tags", "manual", //...)' | \
+./bazel query '(tests(//test) + tests(//test:compiler) + tests(//test/cli/compiler)) except attr("tags", "manual", //...)' | \
```
jez, line 43: Can you remind me which tests we're testing here? `//test` is making sure that the normal testdata tests still typecheck when typechecked by the compiler?

gems/sorbet/lib/serialize.rb
elliottt: I was getting test failures without it, as the TypeMember and TypeTemplate constants were used before being defined. I have no idea how this was happening, so it would be good to try the tests again and see if it's still necessary. This was arising from the snapshot tests, so they're easy to run locally.

.buildkite/test-compiler.sh
elliottt: Now that you bring this up, I don't think that running `//test` is going to do anything when we don't patch the test runner to include the compiler. Maybe we don't need to include this set of tests anymore?

aprocter-stripe (approved):
🚀

PR #4418, opened 2021-07-28
Make `check_type_recursive!` public

djudd:
Move `check_type_recursive!` from `T::Private` to `T::Utils`.

### Motivation
Per discussion in Stripe Slack, we have no strong reason to keep this private and a few legitimate use cases for calling it outside Sorbet internals.

This is a breaking change, but that's explicitly the contract of `T::Private`, and it seems pretty simple to `sed` on upgrade.

### Test plan
Updated existing tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/casts.rb
+++ b/gems/sorbet-runtime/lib/types/private/casts.rb
@@ -2,14 +2,6 @@
 # typed: false
 
 module T::Private
-  # Dynamically confirm that `value` is recursively a valid value of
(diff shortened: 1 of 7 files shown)


jez:
> This is a breaking change, but that's explicitly the contract of `T::Private`, and it seems pretty simple to `sed` on upgrade.

👍

jez (approved):

jez:
Do you mind prepping the Stripe PR to remove those too?

PR #4419, opened 2021-07-28
Allow various T::Types::Base operations on MetaType

jez:
### Motivation

cc @djudd-stripe

There are a handful of operations on MetaTypes that are actually supported,
because they're implemented on `T::Types::Base`. Those get used from time to
time when people are doing metaprogramming.

We might want to ascribe better types to their results, but for now I've left
them as `T.untyped`.

As a suggestion for future implementation, we could do something similar to how
we define intrinsics for `ShapeType` or `TupleType`: invent a special magic
symbol that holds the intrinsics, and attempt to forward the dispatch to one of
those intrinsics in `MetaType::dispatchCall`.

### Test plan

See included automated tests.

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -141,6 +141,13 @@ NameDef names[] = {
     {"revealType", "reveal_type"},
     {"absurd"},
     {"nonForcingIsA_p", "non_forcing_is_a?"},
+    {"valid_p", "valid?"},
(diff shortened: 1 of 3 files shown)


jez:

→ https://go/builds/bui_JwEekfdYkNHYkP
→ https://go/builds/bui_JwEf5T1rBS6r4Q

aprocter-stripe (approved):
This makes sense to me!

PR #2248, opened 2019-11-28
Add support for optional `random` argument to `Array#sample`

olivierbellone:
See https://ruby-doc.org/core-2.6.5/Array.html#method-i-sample

### Motivation
RBI fix.

### Test plan
See included automated tests.

PR #4420, opened 2021-07-29
Add mode parameter to CSV's `foreach` rbi

williamhakim10:
This PR corrects the RBI for CSV's `foreach` method. This method accepts a second argument `mode` (see https://docs.ruby-lang.org/en/2.7.0/CSV.html#method-c-foreach), but the RBI does not currently include that argument.

### Motivation
The RBI for CSV's `foreach` method is incorrect and does not match the underlying method.

### DIFF
--- a/rbi/stdlib/csv.rbi
+++ b/rbi/stdlib/csv.rbi
@@ -269,6 +269,7 @@ class CSV < Object
   sig do
     params(
         path: T.any(String, ::Sorbet::Private::Static::IOLike),
+        mode: String,
         options: T::Hash[Symbol, BasicObject],
         blk: T.proc.params(arg0: T.any(T::Array[T.nilable(T.untyped)], CSV::Row)).void,
     )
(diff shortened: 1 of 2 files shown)


jez (changes requested):
@williamhakim10 The tests failed on this branch. It looks like the signature doesn't work the way you expect it to.

williamhakim10:
@jez whoops, let me fix :)

rbi/stdlib/csv.rbi
```diff
@@ -277,11 +278,12 @@ class CSV < Object
   sig do
     params(
         path: T.any(String, ::Sorbet::Private::Static::IOLike),
+        mode: String,
         options: T::Hash[Symbol, BasicObject],
```
jez, line 282: if options is declared as a kwsplat now, both this type and the type on 273 should say `options: BasicObject` instead of `T::Hash[Symbol, BasicObject]`.

jez (changes requested):
^

rbi/stdlib/csv.rbi
williamhakim10: Fixed, and thanks for helping out. First time opening a PR so the extra help is totally appreciated.

williamhakim10:
Hmm, it appears the issue now is with overloaded functions with keyword arguments. I saw this https://github.com/sorbet/sorbet/pull/2248 - looks like the accepted suggestion as of that discussion is to return `T.any(type1, type2)`. In this case though, one sig returns `void` and the other returns a value?

jez:
> In this case though, one sig returns `void` and the other returns a value?

I don't think at runtime it actually returns `void`, but instead `nil`. You could use `T.nilable(...)`.

You may also want to consider `T.untyped`.

williamhakim10:
Cool. I went with `T.nilable` since it seems better to type the output more strictly if possible?

jez:

→ https://go/builds/bui_Jy1cLcBhsAlsrK
→ https://go/builds/bui_Jy1cJjxVwvhFYM

jez (approved):
Thanks!

PR #4421, opened 2021-07-29
Allow passing in a keyword splat if one of the declared args is a keyword splat

### LINKED ISSUE #4404: Error raised when passing a hash to a double splat
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Adef%20foo(**args)%0A%20%20foo(**args)%0Aend)

```ruby
# typed: true

def foo(**args)
  foo(**args)
end
```

#### Observed output

```
editor.rb:4: Passing a hash where the specific keys are unknown to a method taking keyword arguments https://srb.help/7019
     4 |  foo(**args)
          ^^^^^^^^^^^
  Got T::Hash[Symbol, T.untyped] originating from:
    editor.rb:4:
     4 |  foo(**args)
              ^^^^^^
Errors: 1
```

#### Expected behavior

No error should be raised here.

- - -

neilparikh:
### Motivation

Fixes #4404 

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -899,13 +899,36 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
                 ait += numKwargs;
             }
         } else if (kwSplatIsHash) {
-            if (auto e = gs.beginError(args.callLoc(), errors::Infer::UntypedSplat)) {
-                e.setHeader("Passing a hash where the specific keys are unknown to a method taking keyword "
-                            "arguments");
-                auto &kwSplatArg = *aend;
-                auto kwSplatTPO = TypeAndOrigins{kwSplatType, kwSplatArg->origins};
-                e.addErrorSection(kwSplatTPO.explainGot(gs, args.originForUninitialized));
-                result.main.errors.emplace_back(e.build());
+            bool hasKwSplatParam = false;
+            bool hasKwParam = false;
+            const ArgInfo *kwSplatParam;
+            for (auto &param : data->arguments()) {
+                if (param.flags.isKeyword && param.flags.isRepeated) {
+                    kwSplatParam = &param;
+                    hasKwSplatParam = true;
+                } else if (param.flags.isKeyword && !param.flags.isRepeated) {
+                    hasKwParam = true;
+                }
+            }
+
+            auto &kwSplatArg = *aend;
+            auto kwSplatTPO = TypeAndOrigins{kwSplatType, kwSplatArg->origins};
+
+            if (hasKwSplatParam && !hasKwParam) {
+                if (auto e = matchArgType(gs, *constr, args.callLoc(), args.receiverLoc(), symbol, method,
+                                          kwSplatTPO, *kwSplatParam, args.selfType, targs, kwargsLoc,
+                                          args.originForUninitialized, args.args.size() == 1)) {
+                    result.main.errors.emplace_back(std::move(e));
+                }
+            } else {
+                // A keyword splat was passed in, but none of the declared arguments are keyword splats,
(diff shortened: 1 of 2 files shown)


neilparikh:

→ https://go/builds/bui_JwQ5F3Irh5am4b
→ https://go/builds/bui_JwQ5Mkg48aiMKY

froydnj (approved):
Works for me!

jez (changes requested):
Can you add a test where the types mismatch for the kwsplat?

neilparikh:
@jez Do you mean something like this?

```ruby
sig {params(kw_splat: T::Hash[Symbol, Integer]).void}
def i(**kw_splat)
  puts kw_splat
end

string_values_hash = T.let({}, T::Hash[Symbol, String])
i(**string_values_hash)
```

This doesn't raise any errors, with and without this PR (other than the untyped splat error).

jez:
> doesn't raise any errors ... without this PR (other than the untyped splat error)

Great, that means that there can't be any problems due to it in Stripe's codebase right now.

Can you do the proper type checking now, instead of deferring it to the future where it would require a codemod?

neilparikh:

→ https://go/builds/bui_Jyqr5Ea0GPTDoX
→ https://go/builds/bui_JyqrlBi1TpaEoS

core/types/calls.cc
```diff
@@ -899,13 +899,36 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                                          *kwSplatParam, args.selfType, targs, kwargsLoc, args.originForUninitialized,
+                                          args.args.size() == 1)) {
+                    result.main.errors.emplace_back(std::move(e));
+                }
+            } else {
+                // A keyword splat was passed in, but none of the declared arguments are keyword splats,
+                // or there is a keyword splat argument, but there is one ore more keyword non-splat argument
+                if (auto e = gs.beginError(args.callLoc(), errors::Infer::UntypedSplat)) {
+                    e.setHeader("Passing a hash where the specific keys are unknown to a method taking keyword "
+                                "arguments");
```
jez, line 928: Do you think it's worth making this error message more specific? e.g. this comment has an "either/or" which is very helpful, but neither sides of the either/or end up in the message that we show to the user.

test/testdata/infer/kwargs_splat.rb
```diff
@@ -40,3 +40,26 @@ def g(x, y:, z:, w:)
...
+def i(**kw_splat)
+  puts kw_splat
+end
+
+string_values_hash = T.let({}, T::Hash[Symbol, String])
+i(**string_values_hash)
+
+integer_values_hash = T.let({}, T::Hash[Symbol, Integer])
+i(**integer_values_hash)
+# ^^^^^^^^^^^^^^^^^^^^^ error: Expected `T::Hash[Symbol, String]` but found `T::Hash[Symbol, Integer]` for argument `kw_splat`
```
jez, line 65: awesome, thanks!

jez (approved):
Glad to see that there are no deletions in the tests now, only additions.

One suggestion about improving an error message, but otherwise lgtm

neilparikh:

→ https://go/builds/bui_Jz2POblqoJUmaQ
→ https://go/builds/bui_Jz2P3Pefc6a1KI

core/types/calls.cc
neilparikh: I can't think of any obvious ways to make it more helpful. In both cases, the method will have declared keyword arguments. In one of the cases, there's also a declared keyword splat argument, but the reason for the error in both cases are the keyword arguments, not the keyword splat.

PR #4415, opened 2021-07-28
Add documentation for 5037 errors

Morriar:
image
image

Linked example: https://sorbet.run/#%23%20typed%3A%20true%0A%0Amodule%20Bar%0A%20%20def%20bar%3B%20end%0Aend%0A%0Amodule%20Foo%0A%20%20include%20Bar%0A%0A%20%20alias_method%20%3Afoo%2C%20%3Abar%20%23%20aliases%20are%20resolved%20before%20includes%2C%20so%20%60bar%60%20is%20not%20found%20yet%0A%0A%20%20def%20baz%0A%20%20%20%20bar%20%23%20includes%20are%20resolved%20when%20we%20analyze%20this%20code%0A%20%20end%0Aend

PR #4422, opened 2021-07-29
Fix bad indent in documentation example

Morriar:
### Motivation

This line from #4415 I pushed yesterday was using mixed tab+spaces indent.

This doesn't show in the output but is visible in the Markdown source.

Sorry about that 🙏 

### Test plan

See included automated tests.


elliottt (approved):
Thanks for the follow-up!

PR #4423, opened 2021-07-29
Add documentation for 7005 errors

Morriar:
image

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1050,7 +1050,14 @@ bar(x: 1) # ok
 ## 7005
 
 Sorbet detected a mismatch between the declared return type for the method and
-the type of the returned value.
+the type of the returned value:
+
+```rb
+sig { returns(Integer) }
+def answer
+  "42" # error: Expected `Integer` but found `String("42")` for method result type
+end
+```
 
 Here we specified in the signature that `find` returns an instance of
 `Configuration`, yet the returned value might be `nil`:
@@ -1072,7 +1079,7 @@ def find(name)
 end
 ```
 
-In some case, we're already being cautious and perform some checks before
+In some cases, we're already being cautious and perform some checks before
 returning yet Sorbet still complains about the return type:
 
 ```rb
@@ -1083,7 +1090,7 @@ def find(name)
 end
 ```
 
-While this code is correct, Sorbet cannot assert the state of `@lookup` didn’t
+While this code is correct, Sorbet cannot assume the state of `@lookup` didn’t
 change between the `key?` check and the `[]` read. To fix this, we can take
 advantage of flow-typing to make the whole method work without inline type
 annotations:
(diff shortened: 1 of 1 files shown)


website/docs/error-reference.md
```diff
@@ -1047,6 +1047,59 @@ bar(1) # error
...
+sig { params(name: String).returns(Configuration) }
+def find(name)
+  config = @lookup[name]
+  raise ArgumentError, "Configuration #{name} not found" unless config
+  config
+end
+```
+
+By using a local variable, we allow Sorbet to assert that `config` is never
+nilable passed the `raise` instruction.
```
aprocter-stripe, line 1101:
```suggestion
nilable past the `raise` instruction.
```

website/docs/error-reference.md
```diff
@@ -1047,6 +1047,59 @@ bar(1) # error
...
+is to use `T.must` when returning the value:
+
+```rb
+sig { params(name: String).returns(Configuration) }
+def find(name)
+  T.must(@lookup[name])
+end
+```
+
+In some case, we're already being cautious and perform some checks before
```
aprocter-stripe, line 1075:
```suggestion
In some cases, we're already being cautious and perform some checks before
```

website/docs/error-reference.md
```diff
@@ -1047,6 +1047,59 @@ bar(1) # error
 bar(x: 1) # ok
 ```
 
+## 7005
+
+Sorbet detected a mismatch between the declared return type for the method and
+the type of the returned value.
+
```
aprocter-stripe, line 1054: (non-blocking suggestion) Might be good to throw in an ultra-obvious example to start with, something along the lines of:

```rb
sig { returns(Integer) }
def answer
  "42" # error: Expected `Integer` but found `String("42")` for method result type 
end
```

website/docs/error-reference.md
```diff
@@ -1047,6 +1047,59 @@ bar(1) # error
...
+```rb
+sig { params(name: String).returns(Configuration) }
+def find(name)
+  raise ArgumentError, "Configuration #{name} not found" unless @lookup.key?(name)
+  @lookup[name] # error: Expected `Configuration` but found `T.nilable(Configuration)` for method result type
+end
+```
+
+While this code is correct, Sorbet cannot assert the state of `@lookup` didn’t
```
aprocter-stripe, line 1086: Maybe `assume` or `infer` instead of `assert`?

aprocter-stripe (changes requested):
Thanks much for this! Just a couple of small typos/wording suggestions, and one non-blocking suggestion for an additional example.

website/docs/error-reference.md
Morriar: Good idea 👍

aprocter-stripe (approved):
Thanks!

aprocter-stripe:
Sorry for the delayed merge here! :)

PR #4425, opened 2021-07-29
Add documentation for 7030 errors

Morriar:
It's pretty much my own interpretation of the error based on [this example](https://github.com/sorbet/sorbet/blob/master/test/cli/autocorrect-case-meta-type/autocorrect-case-meta-type.rb). Not sure how it relates to [this one](https://github.com/sorbet/sorbet/blob/master/test/cli/suggest-class-new-not-singleton/suggest-class-new-not-singleton.rb) tho? Any help would be appreciated to make the documentation better 🙏 

image

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1263,8 +1263,11 @@ See [Exhaustiveness Checking](exhaustiveness.md) for more information.
 
 ## 7030
 
-This kind of error generally occurs when we try to use generic types in pattern
-matching:
+This error is consistently used when the user is trying (implicitly or
+explicitly) to call some method on a Sorbet type (e.g. `T::Array[Integer]`)
+which would actually return a Sorbet-runtime representation of a type.
+
+This error generally occurs when generic types are used in pattern matching:
 
 ```rb
 def get_value(input)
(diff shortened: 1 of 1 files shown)


website/docs/error-reference.md
```diff
@@ -1261,6 +1261,37 @@ See [Exhaustiveness Checking](exhaustiveness.md) for more information.
...
+    input
+  when T::Array[Integer] # error: Call to method `===` on `T::Array[Integer]` mistakes a type for a value
+    input.first
+  end
+end
+```
+
+Since generic types are erased at runtime, this construct would never work if we
+execute the program. Sorbet protects us from this and suggests to use the erased
+type `Array` instead:
```
froydnj, line 1282: Similarly here with the "we" language.  I think despite Sorbet offering an autocorrect for this, it would be better to phrase this the user altering their code to use the Ruby type instead.

froydnj (changes requested):
This error is consistently used when the user is trying (implicitly or explicitly) to call some method on a Sorbet type (e.g. `T::Array[Integer]` at runtime would return a sorbet-runtime representation of a type; so would `T.class_of(Foo)`) instead of a value of that type, which is generally not allowed.  So I think we should restructure the documentation to reflect that, and then provide pattern-matching as a common place that this would come up.

Morriar:
Thanks for your review Nathan, I copied your explanation as an introduction for this class of error. Let me know if it works 👍

jez (dismissed):

Morriar:
@froydnj are you happy with the changes?

froydnj (approved):
Works for me, thank you!

PR #4428, opened 2021-07-30
Call rb_hash_update directly instead of through rb_funcallv

elliottt:
Move the definition of sorbet_updateHash into a hash.c patch, so that we can access the c implementation directly and avoid emitting a call to rb_funcallv.

### Motivation

Code cleanliness.

### Test plan

See included automated tests.


froydnj (approved):
WFM!  :jurassic-park:

PR #4429, opened 2021-07-30
Split .so and LLVM IR output dirs to separate command line flags

aprocter-stripe:
(Moving over a compiler-related PR that was originally on the private sorbet_llvm repo)

This adds a `--so-folder=` flag, which tells the compiler where to put the output `.so` files, and changes the semantics of `--llvm-ir-folder` to only dump `.ll`, `.lll`, and `.llo` files in the target directory. (If `--llvm-ir-folder` is not given, no `.ll{,l,o}` files will be produced.)

Thoughts on rollout: there are only two places in Stripe's codebase where `--llvm-ir-folder` is specified, so we could probably fix those up simultaneously with the bump. Alternatively, I could tweak this so that `--llvm-ir-folder` on its own also works backwards compatibly until we're able to make changes on the Stripe side.

### Motivation
We'd like to control how much stuff we write out to disk when compiling in production.

### Test plan
See included automated tests.


compiler/ObjectFileEmitter/ObjectFileEmitter.cc
```diff
@@ -404,15 +406,26 @@ bool ObjectFileEmitter::run(spdlog::logger &logger, llvm::LLVMContext &lctx, uni
     addModulePasses(pm);
     pm.add(Passes::createRemoveUnnecessaryHashDupsPass());
     // print lowered IR
-    auto nameOptl = ((string)dir) + "/" + (string)objectName + ".lll";
-    llvm::raw_fd_ostream lllFile(nameOptl, ec1, llvm::sys::fs::F_Text);
-    auto *printLowered = llvm::createPrintModulePass(lllFile, "");
+    llvm::ModulePass *printLowered = nullptr;
+    // We put the ostream in an std::optional declared outside the "if", because unfortunately `createPrintModulePass`
+    // doesn't take ownership of it.
```
elliottt, line 411: Thanks for the comment here! I can totally see myself refactoring this to move the file handle into the conditional :)

elliottt (approved):
This looks great! Thanks for adding the deprecated flag that sets both of the new ones, that'll make the migration so much easier :D

PR #4433, opened 2021-07-30
--stripe-packages Test module + test_import support

ngroman:
### Motivation
In --stripe-packages add a separate package for test code:
* Test code must be prefixed with `Test::<Package Namespace>`
* Test code is wrapped into a separate registry `<PackageTests>::<Mangled_Package_Name>`
* Test package gets an alias to see all code in it's companion normal package.
* `test_import` works like `import` but also these packages are only accessible by the test code.
* Add ability to `export Test::...` names (these only are imported into tests)

TODO

- [ ] Add more tests
- [ ] Fix bug where if there if the package is empty the alias added does not resolve (see failing tests)
- [ ] cleanup TODOs
- [ ] `export` support for `Test...` contants.

### Test plan

See included automated tests.

### DIFF
--- a/core/Files.cc
+++ b/core/Files.cc
@@ -191,7 +191,7 @@ bool File::isPackage() const {
 }
 
 bool File::isTest() const {
-    return absl::StrContains(path(), "/test/"); // TODO do we need to pre-compute?
+    return absl::EndsWith(path(), ".test.rb") || absl::StrContains(path(), "/test/"); // TODO do we need to pre-compute?
 }
 
 vector<int> &File::lineBreaks() const {
--- a/core/Files.cc
+++ b/core/Files.cc
@@ -190,6 +190,10 @@ bool File::isPackage() const {
     return sourceType == File::Type::Package;
 }
 
+bool File::isTest() const {
+    return absl::StrContains(path(), "/test/"); // TODO do we need to pre-compute?
+}
+
 vector<int> &File::lineBreaks() const {
     ENFORCE(this->sourceType != File::Type::TombStone);
     ENFORCE(this->sourceType != File::Type::NotYetRead);
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -401,6 +401,7 @@ NameDef names[] = {
 
     // Packager
     {"import"},
+    {"test_import"},
     {"export_", "export"},
     {"PackageSpec", "PackageSpec", true},
     {"PackageRegistry", "<PackageRegistry>", true},
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -404,7 +404,7 @@ NameDef names[] = {
     {"export_", "export"},
     {"PackageSpec", "PackageSpec", true},
     {"PackageRegistry", "<PackageRegistry>", true},
-    {"PackageMethods", "<PackageMethods>", true},
+    {"PackageTests", "<PackageTests>", true},
 
     // GlobalState initEmpty()
     {"Top", "<top>", true},
@@ -488,6 +488,7 @@ NameDef names[] = {
     {"NonForcingConstants", "NonForcingConstants", true},
     {"VERSION", "VERSION", true},
     {"Thread", "Thread", true},
+    {"Test", "Test", true}, // TODO will this break people?
 };
 
 void emit_name_header(ostream &out, NameDef &name) {
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -698,7 +698,8 @@ public:
         }
     }
 
-    ast::ClassDef::RHS_store makeModule(core::Context ctx, ImportType importType) { // TODO TODO better name for importType
+    ast::ClassDef::RHS_store makeModule(core::Context ctx,
+                                        ImportType importType) { // TODO TODO better name for importType
         vector<core::NameRef> parts;
         ast::ClassDef::RHS_store modRhs;
         makeModule(ctx, &root, parts, modRhs, importType, ImportTree::Source());
@@ -755,8 +756,8 @@ private:
                 //   D = <Mangled A::B>::A::B::C::D
                 // end
                 auto importLoc = node->source.importLoc;
-                auto assignRhs =
-                    prependPackageScope(parts2literal(parts, core::LocOffsets::none()), node->source.packageMangledName);
+                auto assignRhs = prependPackageScope(parts2literal(parts, core::LocOffsets::none()),
+                                                     node->source.packageMangledName);
                 auto assign = ast::MK::Assign(core::LocOffsets::none(), name2Expr(parts.back(), ast::MK::EmptyTree()),
                                               std::move(assignRhs));
 
@@ -847,11 +848,12 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file, const Pa
         //   <Imports>
         //   <Mangled_Pkg_A>::Pkg::A = <PackageRegistry>::<Mangled_Pkg_A>::Pkg::A
         // end
-        auto assignRhs = prependPackageScope(package->name.fullName.toLiteral(core::LocOffsets::none()), package->name.mangledName);
-        auto assign = ast::MK::Assign(
-                core::LocOffsets::none(),
-                prependScope(package->name.fullName.toLiteral(core::LocOffsets::none()), name2Expr(package->name.mangledName)),
-                std::move(assignRhs));
+        auto assignRhs =
+            prependPackageScope(package->name.fullName.toLiteral(core::LocOffsets::none()), package->name.mangledName);
+        auto assign = ast::MK::Assign(core::LocOffsets::none(),
+                                      prependScope(package->name.fullName.toLiteral(core::LocOffsets::none()),
+                                                   name2Expr(package->name.mangledName)),
+                                      std::move(assignRhs));
         testImportedPackages.emplace_back(std::move(assign));
     }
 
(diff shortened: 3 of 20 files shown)


ngroman:
cc @jvilk-stripe
While I still have some TODOs, the basics are all here. Definitely enough to examine the approach.

nroman-stripe:
Note: can use `<Magic>` for class name to ensure no problem with empty

core/Files.cc
```diff
@@ -190,6 +190,10 @@ bool File::isPackage() const {
     return sourceType == File::Type::Package;
 }
 
+bool File::isTest() const {
```
nroman-stripe, line 193: TODO pull this into packager, not on File

core/tools/generate_names.cc
```diff
@@ -488,6 +489,7 @@ NameDef names[] = {
     {"NonForcingConstants", "NonForcingConstants", true},
     {"VERSION", "VERSION", true},
     {"Thread", "Thread", true},
+    {"Test", "Test", true}, // TODO will this break people?
```
jvilk-stripe, line 492: Nope. Names don't have semantic meaning, so they can be shared between things.
```suggestion
    {"Test", "Test", true},
```

packager/packager.cc
```diff
@@ -700,15 +762,15 @@ class ImportTreeBuilder final {
                                 fmt::map_join(parts, "::", [&](const auto &nr) { return nr.show(ctx); }));
                     e.addErrorLine(core::Loc(ctx.file, parentSrc.importLoc), "Conflict from");
                 }
-            } else {
+            } else if (importType == ImportType::Test || node->source.importType == ImportType::Normal) {
```
jvilk-stripe, line 765: This codepath is now handling:

* Normal imports in normal files.
* Normal imports in test files.
* Test imports in test files.

~...but it's handling all three of these the same way. I think this is correct for the first two scenarios, but not for test files. Don't test exports live in a different `<PackageTests>` namespace?~

I got my wires crossed; this is fine. What's missing is importing `Test::` constants, which you described in the PR.

jvilk-stripe: I think this method is also the place that needs to set up importing the main package into the test package, e.g. making all of `Opus::Foo` available in `Tests::Opus::Foo`.

packager/packager.cc
```diff
@@ -782,35 +846,58 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file, const Pa
...
             } else {
                 importedNames[imported.mangledName] = imported.loc;
-                treeBuilder.mergeImports(*importedPackage, imported.loc);
+                treeBuilder.mergeImports(*importedPackage, import);
             }
         }
-        importedPackages = treeBuilder.makeModule(ctx);
+        importedPackages = treeBuilder.makeModule(ctx, ImportType::Normal);
+        testImportedPackages =
+            treeBuilder.makeModule(ctx, ImportType::Test); // TODO Also need to add alias for original pkg
```
jvilk-stripe, line 854: Don't we do this just below?

packager/packager.cc
```diff
@@ -782,35 +846,58 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file, const Pa
...
+    }
+
+    {
+        // In the test namespace for this package add an alias to give tests full access to the
```
jvilk-stripe, line 858: Ah, I see, we do this here, not in makeModules.

jvilk-stripe (commented):
Left some comments.

With this system, we have, like, four different import scenarios (I almost want to create a matrix):

* `import`ed packages, which are available in the package and test package.
* `test_import`ed packages, which are available only in the test package.
* A package's full internal namespace, which is implicitly imported into the test package.
* `Tests::` exports, which are imported only into test packages via `import` and (confusingly) `test_import`.

One concern I have is that, when someone says "test import" in conversation, it is a little ambiguous (is it a `test_import` or a `Tests::` import?). I almost wonder if there should be a `test_export`, but then there may be confusion over whether or not that export is only available via `test_import`.

I wonder if any other language has solved this problem. I looked at Cargo, and it doesn't seem to have a notion of test exports.

ngroman:
@jvilk-stripe fixed the outstanding issues, added tests. One lsp test is failing that I'm not exactly sure how to fix though.

packager/packager.cc
```diff
@@ -248,15 +274,31 @@ ast::ExpressionPtr prependName(ast::ExpressionPtr scope, core::NameRef name) {
...
         return scope;
     }
 }
 
+// Prefix a constant reference with a name: `Foo::Bar` -> `<REGISTRY>::<name>::Foo::Bar`
+// Registry is either <PackageRegistry> or <PackageTests>. The latter if following the convention
+// that if scope starts with `Test::`.
+ast::ExpressionPtr prependPackageScope(ast::ExpressionPtr scope, core::NameRef mangledName) {
+    auto *lastConstLit = ast::cast_tree<ast::UnresolvedConstantLit>(scope);
+    ENFORCE(lastConstLit != nullptr);
```
jvilk-stripe, line 290: Nit: `cast_tree_nonnull` makes this `ENFORCE` for you in debug builds, and is slightly faster than `cast_tree` in release builds because it doesn't check if the cast failed to return `null`.

If does make the LHS a reference, though, so some lines below need to be changed.

```suggestion
    auto lastConstLit = ast::cast_tree_nonnull<ast::UnresolvedConstantLit>(scope);
```

packager/packager.cc
```diff
@@ -686,29 +763,37 @@ class ImportTreeBuilder final {
...
-                    // TODO Fix flaky ordering of errors. This is strange...not being done in parallel,
-                    // and the file processing order is consistent.
-                    e.setHeader("Conflicting import sources for `{}`",
-                                fmt::map_join(parts, "::", [&](const auto &nr) { return nr.show(ctx); }));
-                    e.addErrorLine(core::Loc(ctx.file, parentSrc.importLoc), "Conflict from");
+                // A conflicting import exist. Only report errors while constructing the test output
+                // to avoid duplicate errors because test imports are a superset of normal imports.
+                if (moduleType == ImportType::Test) {
+                    if (auto e = ctx.beginError(node->source.importLoc, core::errors::Packager::ImportConflict)) {
+                        // TODO Fix flaky ordering of errors. This is strange...not being done in parallel,
```
jvilk-stripe, line 781: Where is this happening / showing up? Is this in a CLI test?

jvilk-stripe: Can you confirm via prints that file processing/error emission is actually consistent?

test/cli/package-test-simple/package-test-simple.out
```diff
@@ -0,0 +1,11 @@
+main_lib/lib.rb:8: Unable to resolve constant `TestOnly` https://srb.help/5002
+     8 |  Project::TestOnly::SomeHelper.new
+          ^^^^^^^^^^^^^^^^^
+  Autocorrect: Use `-a` to autocorrect
+    main_lib/lib.rb:8: Replace with `Project::TestOnly`
```
jvilk-stripe, line 5: 😬 this looks like a bug?

test/cli/package-test-simple/main_lib/__package.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
jvilk-stripe, line 1: Why are all of these defined as `cli` tests rather than `testdata` tests?

`testdata` tests are tested in LSP and CLI modes, whereas these are only tested in whatever command you run.

test/cli/package-test-simple/util/__package.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
+class Project::Util < PackageSpec
+  export Project::Util::MyUtil
+  export Test::Project::Util::UtilHelper
```
jvilk-stripe, line 7: Is this used anywhere? Can you test that this is accessible in tests and not accessible outside of tests?

jvilk-stripe (changes requested):
Small comments. I think we should test `export Test::` and you should consider making more testdata tests and fewer cli tests (unless there's a good reason for those!).

packager/packager.cc
```diff
@@ -18,11 +18,22 @@ namespace sorbet::packager {
 namespace {
 
 constexpr string_view PACKAGE_FILE_NAME = "__package.rb"sv;
+constexpr core::NameRef TEST_NAME = core::Names::Constants::Test();
+
+bool isTestFile(core::File &file) {
+    return absl::EndsWith(file.path(), ".test.rb") || absl::StrContains(file.path(), "/test/");
```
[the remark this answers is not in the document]
ngroman: I think that's the kind of thing we should enforce with inside our own codebase's lint rules, and leave this to be more flexible. My guess is that between `.test.rb` and `test/` other users can probably work it into their setup.

packager/packager.cc
```diff
@@ -18,11 +18,22 @@ namespace sorbet::packager {
...
 struct FullyQualifiedName {
     vector<core::NameRef> parts;
     core::Loc loc;
     ast::ExpressionPtr toLiteral(core::LocOffsets loc) const;
+
+    FullyQualifiedName() = default;
+    explicit FullyQualifiedName(const FullyQualifiedName &) = default;
+    FullyQualifiedName(FullyQualifiedName &&) = default;
+    FullyQualifiedName &operator=(const FullyQualifiedName &) = delete;
+    FullyQualifiedName &operator=(FullyQualifiedName &&) = default;
```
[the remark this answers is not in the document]
ngroman: yep, found a place where I was accidentally copying

packager/packager.cc
ngroman: Maybe I'm missing something but wasn't sure how to get that working properly with the reassignment in the loop below. 

I'll switch it to ` auto *lastConstLit = &ast::cast_tree_nonnull<ast::UnresolvedConstantLit>(scope);`

packager/packager.cc
ngroman: it's consistent in a single file and multiple files w/o multithreading hence `--max-threads 0` in https://github.com/sorbet/sorbet/blob/096051e98f8e7128b5506c95aa5bbad645ae930a/test/cli/package-import-conflicts/package-import-conflicts.sh#L3

I believe you said that was ok a while back.

test/cli/package-test-simple/package-test-simple.out
ngroman: The autocorrects need a lot more work (they should be suggesting imports/exports), but that seems like too much for this PR. I'd prefer to tackle in a follow-up.

test/cli/package-test-simple/util/__package.rb
ngroman: fixed

test/cli/package-test-simple/main_lib/__package.rb
ngroman: The cli test is only here for the basics, the more extensive tests are in testdata. That includes missing test imports, test imports in non-test code etc.

ngroman:
@jvilk-stripe
I'm still figuring out one weird issue with the lsp test, but rest of the issues are taken care of

ngroman:
All tests passing now

PR #4435, opened 2021-07-31
lub: Better reduce AndTypes created during type inference.

jvilk-stripe:
lub: Better reduce AndTypes created during type inference.

Adds the following support to `lub`:

* When lubbing an `AndType` and an `OrType`, call `lubDistributeOr` if our `AndType` lub attempt fails rather than making a new `OrType`.
* When lubbing an `AndType` and an `AndType`, check if their parts are (deeply) referentially equivalent.

There are tests that demonstrate the change.

Not handled in this PR is the following scenario:

* lubbing an `OrType` and an `OrType`

Attempts to handle this use-case ended up being too expensive to land, and changed the types that Sorbet inferred causing type errors on Stripe's codebase. I may pursue this change in a subsequent PR, as type inference still produces some `T.any` types with redundant members.

### Motivation

Type inference creates AndTypes which get nested into OrTypes. This process causes `lub` to get called in a few scenarios that, prior to this change, it was ill-equipped to deal with:

* An `AndType` and an `OrType`.
* An `AndType` and an `AndType` that are equivalent.

One particular file at Stripe spent a lot of time lubbing these types together, which resulted in monstrous types such as the following (note: `Boolean` != `T::Boolean` in this case; Stripe has a legacy module named `Boolean`):

```
T.any(T.all(Boolean, T.nilable(FalseClass)), T.all(Boolean, NilClass), T.all(Boolean, NilClass), T.all(Boolean, NilClass), T.all(Boolean, T.nilable(FalseClass)), FalseClass, T.all(Boolean, NilClass), TrueClass)
``` 

All of this time spent in lub added up to a long typechecking time. The changes in this PR halves the time it spends in typechecking.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/subtyping.cc
+++ b/core/types/subtyping.cc
@@ -108,10 +108,9 @@ TypePtr filterOrComponents(const TypePtr &originalType, const InlinedVector<Type
 TypePtr lubDistributeOr(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2) {
     InlinedVector<TypePtr, 4> originalOrComponents;
     InlinedVector<TypePtr, 4> typesConsumed;
-    auto *o1 = cast_type<OrType>(t1);
-    ENFORCE(o1 != nullptr);
-    fillInOrComponents(originalOrComponents, o1->left);
-    fillInOrComponents(originalOrComponents, o1->right);
+    const auto o1 = cast_type_nonnull<OrType>(t1);
+    fillInOrComponents(originalOrComponents, o1.left);
+    fillInOrComponents(originalOrComponents, o1.right);
 
     for (auto &component : originalOrComponents) {
         auto lubbed = Types::any(gs, component, t2);
--- a/core/types/subtyping.cc
+++ b/core/types/subtyping.cc
@@ -105,7 +105,12 @@ TypePtr filterOrComponents(const TypePtr &originalType, const InlinedVector<Type
     }
 }
 
+TypePtr lubDistributeOrSquared(const GlobalState &gs, const TypePtr &o1, const TypePtr &t2);
 TypePtr lubDistributeOr(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2) {
+    if (isa_type<OrType>(t2)) {
+        return lubDistributeOrSquared(gs, t1, t2);
+    }
+
     InlinedVector<TypePtr, 4> originalOrComponents;
     InlinedVector<TypePtr, 4> typesConsumed;
     auto *o1 = cast_type<OrType>(t1);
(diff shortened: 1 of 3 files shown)


jvilk-stripe:

→ https://go/builds/bui_Jy0MXf1Qg6i0w7
→ https://go/builds/bui_Jy0N7Xz6bCqyjx

jvilk-stripe:

→ https://go/builds/bui_K1XqpZuDKbPthq
→ https://go/builds/bui_K1XrCjUgKKBnq9

jvilk-stripe:

→ https://go/builds/bui_K1cj40IBKnKe0W
→ https://go/builds/bui_K1cjcfWLajxCaq

jvilk-stripe:

→ https://go/builds/bui_K1d18BgVAqvfQ6
→ https://go/builds/bui_K1d1pH0cE1OK3D

jvilk-stripe:

→ https://go/builds/bui_K2COTM1FA9u1mw
→ https://go/builds/bui_K2COgS4J6B0A3e

core/types/subtyping.cc
```diff
@@ -10,6 +10,37 @@ namespace sorbet::core {
...
+    }
+    if (isa_type<AndType>(t1)) {
+        return compositeTypeDeepRefEqual(cast_type_nonnull<AndType>(t1), cast_type_nonnull<AndType>(t2));
+    }
+    return false;
+}
+
+// Returns 'true' if the tree of types stemming from this AndType are referentially equal.
+bool compositeTypeDeepRefEqual(const AndType &a1, const AndType &a2) {
+    return compositeTypeDeepRefEqualHelper(a1.left, a2.left) && compositeTypeDeepRefEqualHelper(a1.right, a2.right);
```
karmakaze, line 35: [Hi, I'm a new (2 days in) contributor, and curious about this, Tuples, and Shapes.]

How do we know that we don't need to also check `compositeTypeDeepRefEqualHelper(a1.left, a2.right) && compositeTypeDeepRefEqualHelper(a1.right, a2.left);`? Or in general, they could be structured differently:
```
t1: (and (and a b) (and c d))
t2:(and a (and b (and c d)))
```

jvilk-stripe: 👋 Hey there! That's a great question.

For performance reasons, `lub` is best-effort -- so it doesn't always return a least upper bound. This is _especially_ true for shape types, which are not well supported in Sorbet (they are [expeirmental](https://sorbet.org/docs/shapes) for a reason!) and which are expensive to lub precisely.

As you expect, this code will not work if left and right are flipped relative to one another in the second type. We could handle this robustly by also checking left against right and right against left, but then you have an exponential blowout as that gets applied recursively!

In the cases I was looking at, this particular compromise worked well enough to simplify the types I was seeing and improve a real performance problem on a complex piece of code.

jvilk-stripe: (To be completely fair, I haven't benchmarked the exponential case. It might be neutral. I may try it after this lands.)

core/types/subtyping.cc
```diff
@@ -10,6 +10,37 @@ namespace sorbet::core {
 
 using namespace std;
 
+namespace {
+bool compositeTypeDeepRefEqual(const OrType &o1, const OrType &o2);
+bool compositeTypeDeepRefEqual(const AndType &a1, const AndType &a2);
+bool compositeTypeDeepRefEqualHelper(const TypePtr &t1, const TypePtr &t2) {
```
elliottt, line 16: Just checking an assumption here: this won't handle cases where the arguments to the And/Or types are equivalent but in a different order, right? This is probably another good argument for a more stable ordering for the arguments to thesetypes.

elliottt (approved):
I think this change makes sense. I can imagine us wanting to push harder on a normal form for the and/or types in the future, which would definitely make the equality check you've added go further.

core/types/subtyping.cc
jvilk-stripe: Correct!

PR #4437, opened 2021-07-31
Avoid building llvm in test-static-sanitized

elliottt:
The `tags` argument to `test_suite` is for finding tests, not tagging the test suite itself (with the exception of the `"manual"` tag, which is used to keep the test suite from expanding in a wildcard query `//...`). We had a few different places in build system where we were supplying tags other than `"manual"` to the `test_suite` rule, with the effect of it including more tests than we had intended.

Additionally, `bazel test //...` will cause more targets to be built than just tests, so I've also added in `--build_tests_only` in `test-static-sanized.sh` to avoid accidentally building llvm there.

### Motivation

Better control over when llvm is built in CI.

### Test plan

* [x] Run `./bazel test //test/cli:update -c opt` to ensure that cli test updates still work after updating the test suite tags

See included automated tests.

### DIFF
--- a/gems/sorbet/test/snapshot/BUILD
+++ b/gems/sorbet/test/snapshot/BUILD
@@ -65,9 +65,6 @@ snapshot_tests(
 
 test_suite(
     name = "snapshot",
-
-    # NOTE: this is manual to avoid being caught with `//...`
-    tags = ["manual"],
     tests = [
         ":snapshot-sorbet_ruby_2_6",
         ":snapshot-sorbet_ruby_2_7",
@@ -76,9 +73,6 @@ test_suite(
 
 test_suite(
     name = "update",
-
-    # NOTE: this is manual to avoid being caught with `//...`
-    tags = ["manual"],
     tests = [
         ":update-sorbet_ruby_2_6",
         ":update-sorbet_ruby_2_7",
--- a/gems/sorbet/test/snapshot/BUILD
+++ b/gems/sorbet/test/snapshot/BUILD
@@ -65,6 +65,9 @@ snapshot_tests(
 
 test_suite(
     name = "snapshot",
+
+    # NOTE: this is manual to avoid being caught with `//...`
+    tags = ["manual"],
     tests = [
         ":snapshot-sorbet_ruby_2_6",
         ":snapshot-sorbet_ruby_2_7",
@@ -73,6 +76,9 @@ test_suite(
 
 test_suite(
     name = "update",
+
+    # NOTE: this is manual to avoid being caught with `//...`
+    tags = ["manual"],
     tests = [
         ":update-sorbet_ruby_2_6",
         ":update-sorbet_ruby_2_7",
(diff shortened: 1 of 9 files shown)


gems/sorbet/test/snapshot/BUILD
```diff
@@ -65,9 +65,6 @@ snapshot_tests(
 
 test_suite(
     name = "snapshot",
-
-    # NOTE: this is manual to avoid being caught with `//...`
-    tags = ["manual"],
```
froydnj, line 70: Removing `manual` here seems unusual: https://docs.bazel.build/versions/main/test-encyclopedia.html#tag-conventions

Some of the other tag removals ("external") are also defined at the above link; why are we removing all the tags?

elliottt: I think the difference is that these tags aren't applied to the `test_suite` target, but are used to find tests to include in it. This is pretty confusing given that in every other use `tags` applies the tags to the target: https://docs.bazel.build/versions/main/be/general.html#:~:text=Only%20test%20rules,by%20visibility%20constraints).

elliottt: Though now that I'm reading beyond what I linked, I'm seeing this:

> The manual tag keyword is treated differently than the above by the "test_suite expansion" performed by the bazel test command on invocations involving wildcard target patterns. There, test_suite targets tagged "manual" are filtered out (and thus not expanded). This behavior is consistent with how bazel build and bazel test handle wildcard target patterns in general.

Which indicates that the `"manual"` tag is treated not as a filter for finding tests, but instead is used for having the `test_suite` skipped during wildcard expansion. That's pretty unfortunate, but does explain why all of a sudden I started seeing compiler targets showing up in this build :)

I'll put the manual uses back in, but remove the other cases where the tags might accidentally widen the scope of the test suite. Thanks for pointing this out!

jez (approved):
🎉

test/cli/cli_test.bzl
```diff
@@ -65,10 +65,6 @@ def update_test():
     ]
     native.test_suite(
         name = "update",
-        tags = [
-            "manual",
-            "external",
-            "local",
-        ],
```
froydnj, line 72: Given https://docs.bazel.build/versions/main/test-encyclopedia.html#tag-conventions I don't think we want to remove "external" (unless we *do* want caching for this test).  https://docs.bazel.build/versions/main/be/common-definitions.html#common.tags also suggests "local" has a useful meaning too...

froydnj (approved):

test/cli/cli_test.bzl
elliottt: They do have a special meaning for a normal rule, but as far as I can tell the only tag that has a special meaning for a `test_suite` is `"manual"`. I verified this by re-running the target locally, and am pretty sure that the only thing that keeping these tags here would do is to add additional targets to this test suite that were tagged either `"external"` or `"local"`.

PR #4441, opened 2021-08-01
sorbet_llvm -> the Sorbet Compiler

jez:
### Motivation

There were a handful of references to sorbet_llvm that don't make sense anymore.

### Test plan

See included automated tests.


compiler/IREmitter/IREmitter.cc
```diff
@@ -866,7 +866,7 @@ void IREmitter::buildInitFor(CompilerState &cs, const core::SymbolRef &sym, stri
     builder.SetInsertPoint(bb);
 
     if (IREmitterHelpers::isFileOrClassStaticInit(cs, sym)) {
-        // We include sorbet_version.c when compiling sorbet_llvm itself to get the expected version.
+        // We include sorbet_version.c when compiling the Sorbet Compiler itself to get the expected version.
```
froydnj, line 869: Just out of curiosity, did we verify that this still works?

compiler/IREmitter/Intrinsics/wrap-intrinsics.rb
```diff
@@ -5,7 +5,7 @@
 require 'set'
 require 'optparse'
 require 'stringio'
-require_relative '../../../bazel-sorbet_llvm/external/com_stripe_ruby_typer/gems/sorbet-runtime/lib/sorbet-runtime'
+require_relative '../../../gems/sorbet-runtime/lib/sorbet-runtime'
```
froydnj, line 8: This sort of thing is one nice perk of having a merged codebase.

compiler/README.md
```diff
@@ -255,7 +247,7 @@ option that will drop into `lldb` before running.
 
 ### `.lldbinit`
 
-We have a custom `.lldbinit` file in `sorbet_llvm` with some helper commands.
+We have a custom `.lldbinit` file with some helper commands.
```
froydnj, line 250: Does that still live in `compiler/` or did it get merged into the toplevel?

compiler/README.md
```diff
@@ -353,26 +345,20 @@ ownership of `y` is transfered to `z`. This means that the memory associated
...
 
 ## Editor and environment
 
-Nearly all of the [Editor and
-environment](https://github.com/sorbet/sorbet#editor-and-environment) tips from
-the Sorbet repo apply here too.
+Nearly all of the [Editor and environment] tips from the Sorbet repo apply here
+too.
+
+[Editor and environment]: https://github.com/sorbet/sorbet#editor-and-environment
```
froydnj, line 356: Do we need this?  If we do, it seems weird to talk about the tips from the same repo that we're in?

froydnj (approved):
This looks reasonable; all the comments below are non-blocking.

compiler/README.md
jez: It’s in the top level (it Changes in this pr)

compiler/IREmitter/IREmitter.cc
jez: I did not, but I see no reason why it would have stopped working. The way we compute the version information is using git information from the repo where Bazel is, which in the case of this line is the Git SHA of Stripe's codebase at the time that the Sorbet version is bumped.

PR #4444, opened 2021-08-02
make compiled methods have sigs from sorbet-runtime's perspective

froydnj:
(This is pulled over from the original Sorbet compiler repo; @elliottt OK'd this PR there, but resubmitting since we've merged everything.)

### Motivation

This brings compiled methods closer to parity with interpreted methods (e.g. `signature_for_method` works). It also means that `final` compiled methods are subject to the runtime checking that sorbet-runtime does, which means that this change is also a significant correctness change.

I am not completely happy with the compiler runtime support setup; the hashtables are size `O(number of modules defined by user code)`, but perhaps that can be adjusted (per-compilation unit hashtables?) at some future date.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_Jy0uw3JGqICxXG
→ https://go/builds/bui_Jy0uqTTLwKBOPW

aprocter-stripe (approved):
Quoth @elliottt on the old repo:

> This all looks good to me! I think if you move the PR over to sorbet we should just approve and merge it 👍

So, LGTM!

PR #4445, opened 2021-08-02
fix fixnum/NaN comparisons for <= and >=

froydnj:
(This is pulled over from the original Sorbet compiler repo; @aprocter-stripe  OK'd this PR there, but resubmitting since we've merged everything.)

### Motivation

Our slowpath code was not handling NaNs correctly in the fixnum comparison case for some comparison functions. `int_ge.rb` and `int_le.rb` failed with the added tests in baf3a33

I added more tests everywhere to a) make sure we didn't have problems elsewhere and b) completeness.

I assume we copied code from the static functions in numeric.c because the slowpath code was written for the codegen payload at a time when we didn't want to export static functions from `libruby.so`. But now that we directly patch files in the Ruby source, we can reuse the code in numeric.c without copying and avoid subtle bugs like this one.

I suppose we could use `rb_int_gt` etc. directly in our slowpath functions; I have no strong feelings on the matter, though using them is slightly more correct if somebody decided to subclass `Integer` (?!).  I don't think we should handle this case -- we decided in previous discussion not to.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/Payload/patches/numeric.c
+++ b/compiler/IREmitter/Payload/patches/numeric.c
@@ -36,16 +36,7 @@ VALUE sorbet_rb_int_minus_slowpath(VALUE recv, VALUE y) {
 
 VALUE sorbet_rb_int_gt_slowpath(VALUE recv, VALUE y) {
     if (LIKELY(FIXNUM_P(recv))) {
-        if (LIKELY(FIXNUM_P(y))) {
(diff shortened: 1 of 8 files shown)


froydnj:

→ https://go/builds/bui_Jy1C4g5qKAYvFN
→ https://go/builds/bui_Jy1C6mEr5LtosP

aprocter-stripe (approved):

PR #4446, opened 2021-08-02
Add documentation for 7009 errors

Morriar:
image

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1186,6 +1186,27 @@ def publish_item(name, category = nil)
 end
 ```
 
+## 7009
+
+This error occurs when a value is used in place of a type. There are many
+different situations where this can happen; one example is given below:
+
+```
+class Box
+  extend T::Generic
+
+  E = type_member
+end
+
+Box[true].new # error: Unsupported usage of bare type
+```
+
+To fix this error, replace the `true` value with the type `TrueClass`:
+
+```rb
+Box[TrueClass].new
+```
+
 ## 7013
 
 Sorbet detected that an instance variable was reassigned with different types:


website/docs/error-reference.md
```diff
@@ -1126,6 +1126,26 @@ def publish_item(name, category = nil)
...
+class Box
+  extend T::Generic
+
+  E = type_member
+end
+
+Box[true].new # error: Unsupported usage of bare type
+```
+
+To fix this error, replace the `true` value but the type `TrueClass`:
```
froydnj, line 1143: Did you mean to write `s/but/with/` here?

froydnj (changes requested):
Apart from the nit below, I am not sure this captures the details of the situation.  See e.g.

`test/testdata/infer/pinned_control_flow.rb`

which does not fit into the description given here.  It is, admittedly, a secondary error in that testcase, so maybe we should ignore that for the purposes of writing documentation here?

jez:
For what it's worth, this error shows up in _all sorts_ of completely unrelated situations.

There's probably only so much targeted help we can give people for this error. Some suggestions (any/all of):

- make the doc more clear that there are lots of ways this error can happen
- figure work out more cases where this error can happen, add more examples
- split the error up in to more than one code, so it's possible to give better help / make the message better

website/docs/error-reference.md
```diff
@@ -1126,6 +1126,26 @@ def publish_item(name, category = nil)
 end
 ```
 
+## 7009
+
+This error occurs when a value is used in place of a type:
```
froydnj, line 1131: Given jez's comments, I think it would be good to add a note here that the example below is not exhaustive:

```suggestion
This error occurs when a value is used in place of a type.  There are many different situations where this can happen; one example is given below:
```

froydnj (approved):
I am not sure that it's worth chasing down all the situations that this can happen, so let's at least provide a hint to the user that the below example is not exhaustive.

froydnj:
Looks like we have some rebasing required @Morriar ?

Morriar:
@froydnj rebased, thanks! 👍

PR #4447, opened 2021-08-02
Add documentation for 7010 errors

Morriar:
image

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1166,7 +1166,7 @@ Sorbet found a reference to a generic type with the wrong number of type
 arrguments.
 
 Here we defined `MyMap` as a generic class expecting two type parameters
-`KeyType` and `ValueType` but we try to instanciate it with only one type
+`KeyType` and `ValueType` but we try to instantiate it with only one type
 argument:
 
 ```rb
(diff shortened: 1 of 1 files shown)


website/docs/error-reference.md
```diff
@@ -1126,6 +1126,38 @@ def publish_item(name, category = nil)
...
+
+  # ...
+end
+
+MyMap[String].new # error: Wrong number of type parameters for `MyMap`. Expected: `2`, got: `1`
+```
+
+Unless a type member was `fixed`, it is always required to pass the correct
+amount of type arguments. `T.untyped` can also be used is the type is not
+relevant at this point:
```
vaporyhumo, line 1153:
```suggestion
amount of type arguments. `T.untyped` can also be used as the type is not
relevant at this point:
```

Morriar: I meant `if`, thanks! 👍

website/docs/error-reference.md
```diff
@@ -1126,6 +1126,38 @@ def publish_item(name, category = nil)
 end
 ```
 
+## 7010
+
+Sorbet found a reference to a generic type with the wrong number of type
+arrguments.
+
+Here we defined `MyMap` as a generic class expecting two type parameters
+`KeyType` and `ValueType` but we try to instanciate it with only one type
```
jez, line 1135:
```suggestion
`KeyType` and `ValueType` but we try to instantiate it with only one type
```

jez (approved):
You have conflicts with another PR (probably #4448)

Morriar:
Rebased on master 👍

Morriar:
@jez I think this one can be safely merged 👍

PR #4448, opened 2021-08-02
Add documentation for 7013 errors

Morriar:
image

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1149,7 +1149,7 @@ end
 If the instance variable can hold both an `Integer` and a `String`, maybe the
 type specified with `T.let` should be enlarged:
 
-```
+```ruby
 @x = T.let(0, Object)
 ```
 
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1126,6 +1126,40 @@ def publish_item(name, category = nil)
 end
 ```
 
+## 7013
+
+Sorbet detected that an instance variable was reassigned with different types:
+
+```rb
+class A
+  extend T::Sig
+
+  sig { void }
+  def initialize
+    @x = T.let(0, Integer)
+  end
+
+  sig { void }
+  def foo
+    @x = 'not an integer' # error: Reassigning field with a value of wrong type: `String("not an integer")` is not a subtype of `Integer`
+  end
+end
+```
+
+If the instance variable can hold both an `Integer` and a `String`, maybe the
+type specified with `T.let` should be enlarged:
+
+```
+@x = T.let(0, Object)
+```
+
+Similarly, Sorbet will reject constants reassigned with different types:
+
+```rb
+FOO = 42 # error: Reassigning field with a value of wrong type: `Integer(42)` is not a subtype of `String("Hello, world!")`
+FOO = "Hello, world!"
+```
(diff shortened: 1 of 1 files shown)


website/docs/error-reference.md
```diff
@@ -1126,6 +1126,23 @@ def publish_item(name, category = nil)
 end
 ```
 
+## 7013
+
+Sorbet does not support constant reassignment:
```
jez, line 1131: It actually does, it just requires that they be the same type. Do you think that's worth mentioning?

jez: #### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0AFoo%20%3D%201%0AFoo%20%3D%202)

```ruby
# typed: strict
Foo = 1
Foo = 2
```

#### Observed output

```
No errors! Great job.
```

website/docs/error-reference.md
```diff
@@ -1126,6 +1126,23 @@ def publish_item(name, category = nil)
...
+FOO = "Hello, world!"
+```
+
+If the value of `FOO` can be changed, it should be a proper variable or an
+attribute instead of a constant:
+
+```rb
+foo = 42
+foo = "Hello, world!"
+```
```
jez, line 1144: This error is not unique to constants, it can affect instance variables as well:

#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20initialize%0A%20%20%20%20%40x%20%3D%20T.let(0%2C%20Integer)%0A%20%20end%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20foo%0A%20%20%20%20%40x%20%3D%20'not%20an%20integer'%0A%20%20end%0Aend)

```ruby
# typed: strict

class A
  extend T::Sig
  sig {void}
  def initialize
    @x = T.let(0, Integer)
  end
  sig {void}
  def foo
    @x = 'not an integer'
  end
end
```

#### Observed output

```
editor.rb:11: Reassigning field with a value of wrong type: String("not an integer") is not a subtype of Integer https://srb.help/7013
    11 |    @x = 'not an integer'
                 ^^^^^^^^^^^^^^^^
Errors: 1
```

This is likely the more common way to discover this error. Can you update the doc to mention that?

jez (changes requested):
^

Morriar:
Good catch, thanks!

I rewrote the documentation taking this into account 👍

Morriar:
New version:

image

website/docs/error-reference.md
```diff
@@ -1126,6 +1126,40 @@ def publish_item(name, category = nil)
...
+  def foo
+    @x = 'not an integer' # error: Reassigning field with a value of wrong type: `String("not an integer")` is not a subtype of `Integer`
+  end
+end
+```
+
+If the instance variable can hold both an `Integer` and a `String`, maybe the
+type specified with `T.let` should be enlarged:
+
+```
```
jez, line 1152:
````suggestion
```ruby
````

jez (approved):

PR #4449, opened 2021-08-03
fail harder when lowering sorbet_i_getRuby{Class,Constant}

froydnj:
### Motivation

The current setup for lowering `sorbet_i_getRuby{Class,Constant}` emits LLVM undef instructions when encountering a shape of IR that we don't understand how to lower.  This decision results in hard-to-debug failures, as @neilparikh has experienced, and is just a bad idea from a compiler reliability perspective: we shouldn't be emitting LLVM undef (which LLVM can do whatever it likes with) if we don't understand the IR we're receiving.

Instead, we should crash and crash loudly, because getting badly-shaped IR is a bug of some kind.  This PR implements that behavior.

It is a little bit messy, because we can't use the usual `failCompilation` bits that the rest of the compiler uses.  Ideally the comments in the source code will help explain why we can't and shed light on why this approach was chosen.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/Core/OptimizerException.h
+++ b/compiler/Core/OptimizerException.h
@@ -1,0 +1,14 @@
+#include "common/Exception.h"
+
+namespace sorbet::compiler {
+// Occasionally we need to throw errors from within LLVM optimization passes,
+// where we don't have access to GlobalState and therefore cannot use failCompilation.
+// We have this separate class so that the compiler plugin can differentiate this
+// exception from AbortCompilation and know that a Sorbet-level error needs to be
+// emitted.  The user can therefore have some idea of why Sorbet failed.
+class OptimizerException : public sorbet::SorbetException {
+public:
+    OptimizerException(const std::string &message) : SorbetException(message){};
+    OptimizerException(const char *message) : SorbetException(message){};
+};
+} // namespace sorbet::compiler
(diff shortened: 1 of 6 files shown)


froydnj:

→ https://go/builds/bui_JyL1MdsUNkAEGs
→ https://go/builds/bui_JyL2AXpoU28bVX

compiler/Core/OptimizerException.h
```diff
@@ -0,0 +1,14 @@
+#include "common/Exception.h"
+
+namespace sorbet::compiler {
+// Occasionally we need to throw errors from within LLVM optimization passes,
+// where we don't have access to GlobalState and therefore cannot use failCompilation.
+// We have this separate class so that the compiler plugin can differentiate this
+// exception from AbortCompilation and know that a Sorbet-level error needs to be
+// emitted.  The user can therefore have some idea of why Sorbet failed.
+class OptimizerException : public sorbet::SorbetException {
```
jez, line 9: Should this use inheritance? Or should it define its own hierarchy?

jez: ah looks like we use inheritance when we define AbortCompilation. Makes sense to keep it the same then.

jez (approved):

PR #4453, opened 2021-08-04
Fail on Kernel private methods

### LINKED ISSUE #4434: Instance methods on Kernel are not marked private
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%22%22.sleep(1)%0A%22%22.puts)

```ruby
# typed: true
"".sleep(1)
"".puts
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

Both of these calls will raise `NoMethodError` at runtime because they're both private methods.

- - -

Note that the kernel methods are also all accessible from the `Kernel` singleton class object, which we should continue to allow:

#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0AKernel.sleep(1)%0AKernel.puts)

```ruby
# typed: true
Kernel.sleep(1)
Kernel.puts
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

currently working as intended, so:

```
No errors! Great job.
```

aledalgrande:
### Motivation

Fixes #4434 by declaring all methods in Object coming from Kernel as private. As I understand, we want to keep these methods working with no caller, but failing when called on any object.

The change:
- Kernel methods like `puts 1` still work
- private instance methods added to Object by Kernel like `''.puts 1` used to work, now fail

### Test plan

**holding back for now, see comment**

### DIFF
--- a/rbi/core/object.rbi
+++ b/rbi/core/object.rbi
@@ -129,4 +129,2567 @@ class Object < BasicObject
       .returns(T.type_parameter(:X))
   end
   def then(&blk); end
+
+    ### A note on global functions:
+  ###
+  ### Ruby tends to define global (e.g. "require", "puts") as
+  ### module_function's on Kernel. In practice, this is mostly
+  ### superfluous; since Kernel is itself an ancestor of
+  ### Kernel.singleton_class, all of Kernel's instance methods will
+  ### automatically appear on Kernel directly. For this reason, we omit
+  ### module_function and just define these methods as instance methods,
+  ### for clarity and simplicity.
+
+  # Returns the current execution stack---an array containing strings in the
+  # form `file:line` or `file:line: in `method'`.
+  #
+  # The optional *start* parameter determines the number of initial stack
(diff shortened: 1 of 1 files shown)


aledalgrande:
First contribution.

Planning to write tests, but first want to confirm if this PR is already too big and needs to be divided in multiple smaller PRs, I read >100 lines is already a big change.

Another thing I want to confirm is if signatures are used on private methods or can be removed.

jez (changes requested):
The thing I'm worried about is that these new method definitions weren't previously required for things to type check. E.g.:

```ruby
Kernel.puts
puts
```

both of these type checked before. Why do we need new method defs now?

Also, if I understand this change correctly, this will not solve the problem for when private Kernel methods are used inside modules that include or extend `Kernel`.

aledalgrande:
Ohhh gotcha. You are correct. I will have to rethink the change.

PR #4457, opened 2021-08-05
Remove unnecessary calls to rb_hash_dup when keyword args are directly passed through to another method call

neilparikh:
### Motivation

Follow up PR to stripe/sorbet_llvm#520

This PR handles the following case:
```ruby
def foo(**args)
  bar(**args)
end
```
Here, the keyword args are being directly passed through to `bar`, so there's no need to duplicate them again.

### Test plan

See included automated tests.

Benchmark (on `master`):
```
ruby vm startup time: 0.129
stripe/while_1_000_000.rb       0.136   0.144
stripe/kwargs_splat_pass_through.rb     0.545   0.351
stripe/kwargs_splat_pass_through.rb - baseline  0.409   0.207
```
Benchmark (on this branch):

```
ruby vm startup time: 0.122
stripe/while_1_000_000.rb       0.135   0.133
stripe/kwargs_splat_pass_through.rb     0.538   0.292
stripe/kwargs_splat_pass_through.rb - baseline  0.403   0.159
```

### DIFF
--- a/compiler/Passes/Lowerings.cc
+++ b/compiler/Passes/Lowerings.cc
@@ -658,8 +658,7 @@ public:
         auto *sorbetGlobalConstDupHashFn = mod.getFunction("sorbet_globalConstDupHash");
         auto *sorbetSendFn = mod.getFunction("sorbet_i_send");
 
-        if (rbHashDupFn == nullptr || rbHashNewFn == nullptr || rbToHashTypeFn == nullptr ||
-            sorbetGlobalConstDupHashFn == nullptr || sorbetSendFn == nullptr) {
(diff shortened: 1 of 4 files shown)


elliottt (changes requested):
I think this looks good, but it would be great to only disable one or the other of the two detection methods if one of the functions you compare against is missing from the module.

elliottt (approved):
Thanks for moving the check lower!

neilparikh:

→ https://go/builds/bui_K2BxsMzZxUg9fx
→ https://go/builds/bui_K2ByqYvLJCxQoh

PR #4458, opened 2021-08-05
add testcases for block return across ensure

froydnj:
### Motivation

The first step towards fixing things is usually small easily understandable test cases.

### Test plan

See included automated tests.


jez (approved):
Nice!

does it require Thread? Could the same thing happen with just instance variables? (Trying to think about how to minimize, not blocking)

froydnj:
> Nice!
> 
> does it require Thread? Could the same thing happen with just instance variables? (Trying to think about how to minimize, not blocking)

I do not believe `Thread` is required; the same thing should occur with instance variables, locals, etc.

froydnj:
Added some testcases that `return` and `break` across `else` which runs through similar code paths, but doesn't exhibit the same issues.

PR #4460, opened 2021-08-05
Create sorbet-static-and-runtime gem

Morriar:
### Motivation

Create a single gem that depends on both `sorbet`  and `sorbet-runtime`. Most people use both gems at the same version so this will make it easier to install/update Sorbet in many projects (especially when using Dependabot).

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-static-and-runtime/sorbet-static-and-runtime.gemspec
+++ b/gems/sorbet-static-and-runtime/sorbet-static-and-runtime.gemspec
@@ -1,0 +1,18 @@
+Gem::Specification.new do |s|
+  s.name        = 'sorbet-static-and-runtime'
+  s.version     = '0.0.0'
+  s.summary     = 'A Typechecker for Ruby'
+  s.description = 'Sorbet static and runtime in one gem'
+  s.authors     = ['Stripe']
+  s.email       = 'sorbet@stripe.com'
+  s.homepage    = 'https://sorbet.org'
+  s.license     = 'Apache-2.0'
+  s.metadata = {
+    'source_code_uri' => 'https://github.com/sorbet/sorbet'
+  }
+
+  s.add_dependency 'sorbet', '0.0.0'
+  s.add_dependency 'sorbet-runtime', '0.0.0'
+
+  s.required_ruby_version = ['>= 2.3.0']
+end
--- a/.buildkite/build-sorbet-static-and-runtime.sh
+++ b/.buildkite/build-sorbet-static-and-runtime.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -eo pipefail
+
+pushd gems/sorbet-static-and-runtime
+
+echo "--- setup :ruby:"
+eval "$(rbenv init -)"
+
+runtime_versions=(2.6.3 2.7.2)
+
+for runtime_version in "${runtime_versions[@]}"; do
+  rbenv install --skip-existing "$runtime_version"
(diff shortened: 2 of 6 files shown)


Morriar:
I didn't unlock the tests as they are also running publish related things and I prefer to have an adult look at the code first.

jez:
> I didn't unlock the tests as they are also running publish related things and I prefer to have an adult look at the code first.

https://github.com/sorbet/sorbet/pull/4460/files#diff-54e25afd8f5d3edc378b93ad0eac18f6a751454cce518692c708258969b66e7aR5-R8

The whole job you edited is skipped if it is not the nightly publish job

https://github.com/sorbet/sorbet/pull/4460/files#diff-54e25afd8f5d3edc378b93ad0eac18f6a751454cce518692c708258969b66e7aR5-R8

gems/sorbet-static-and-runtime/sorbet-static-and-runtime.gemspec
```diff
@@ -0,0 +1,18 @@
...
+  s.license     = 'Apache-2.0'
+  s.metadata = {
+    'source_code_uri' => 'https://github.com/sorbet/sorbet'
+  }
+
+  s.add_dependency 'sorbet', '0.0.0'
+  s.add_dependency 'sorbet-runtime', '0.0.0'
+
+  s.required_ruby_version = ['>= 2.3.0']
+end
```
jez, line 18: This gem is not going to be automatically built. You will need to find somewhere to run `gem build`, and also do the logic to `sed` in the version.

gems/sorbet-static-and-runtime/sorbet-static-and-runtime.gemspec
```diff
@@ -0,0 +1,18 @@
+Gem::Specification.new do |s|
+  s.name        = 'sorbet-static-and-runtime'
+  s.version     = '0.0.0'
+  s.summary     = 'A Typechecker for Ruby'
+  s.description = 'Sorbet static and runtime in one gem'
+  s.authors     = ['Stripe']
+  s.email       = 'sorbet@stripe.com'
+  s.homepage    = 'https://sorbet.run'
```
jez, line 8:
```suggestion
  s.homepage    = 'https://sorbet.org'
```

It's possible that the other gems say `sorbet.run` but that's probably an artifact of a time when we didn't have `sorbet.org` yet.

jez (changes requested):
^

(thanks!)

gems/sorbet-static-and-runtime/sorbet-static-and-runtime.gemspec
Morriar: I added a `.buildkite/build-sorbet-static-and-runtime.sh` script to call `gem build` and apply the `sed`.

I also added an entry for it in the `.buildkite/pipeline.yaml`.

Morriar: The result can be seen here: https://buildkite.com/sorbet/sorbet/builds/19757#4c991446-3857-4711-9f79-488725fd6ca1

jez (approved):
Looks great, thanks!

jez:
As a follow up, do you want to mention this in the sorbet.org docs?

Morriar:
> As a follow up, do you want to mention this in the sorbet.org docs?

Good point, see https://github.com/sorbet/sorbet/pull/5106.

PR #4462, opened 2021-08-06
ensure that the ensure block of a rescue is called from a single place

froydnj:
### Motivation

`ensure` is broken in a couple different ways; see the recent testcases in #4458, and probably some other, yet to be written, testcases.  This PR is a step towards fixing those by moving the actual execution of the `ensure` block to a single place in the generated LLVM IR, rather than executing it twice along different codepaths.

Once that's done, it's much easier to wrap the bulk of the exception logic with a unwind handler (`EC_EXEC_TAG` and friends), execute the ensure handler regardless of how `EC_EXEC_TAG` returned, and `EC_JUMP_TAG` or return the actual computed value, as appropriate.  (I haven't decided whether the latter wrapping changes will be in this PR or the next.)

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/Exceptions.cc
+++ b/compiler/IREmitter/Exceptions.cc
@@ -26,8 +26,14 @@ class ExceptionState {
     llvm::Function *currentFunc;
 
     llvm::Value *previousException = nullptr;
-    llvm::BasicBlock *exceptionEntry = nullptr;
     llvm::Value *exceptionResultPtr = nullptr;
+
+    llvm::BasicBlock *exceptionEntry = nullptr;
+    llvm::BasicBlock *bodyReturn = nullptr;
+    llvm::BasicBlock *handlersBlock = nullptr;
+    llvm::BasicBlock *ensureBlock = nullptr;
+    llvm::BasicBlock *exceptionContinue = nullptr;
+    llvm::BasicBlock *exceptionReturn = nullptr;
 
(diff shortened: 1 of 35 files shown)


jez:
(removing my review request due to wip, re-request if i should look anyways)

froydnj:
Compiler tests are going to fail because I haven't been able to regenerate the exp files, which is the subject of #4466.  I will fix those up in the obvious way.

compiler/IREmitter/Exceptions.cc
```diff
@@ -119,17 +149,24 @@ class ExceptionState {
         auto [bodyResult, exceptionResultPtr] = sorbetTry(bodyFunction, previousException);
 
         // Check for an early return from the body.
-        auto *earlyReturnBlock = llvm::BasicBlock::Create(cs, "exception-body-return", currentFunc);
-        auto *continueBlock = llvm::BasicBlock::Create(cs, "exception-body-continue", currentFunc);
+        auto *earlyReturnBlock = this->bodyReturn;
+        auto *continueBlock = this->handlersBlock;
```
elliottt, line 153: Is it worth deleting these declarations and inlining the fields from the rhs?

elliottt (approved):
This looks great, thanks!

compiler/IREmitter/Exceptions.cc
froydnj: I wondered about this when I was writing the patch, but ultimately decided to keep them for the purposes of this patch to try and minimize the diff.  I am totally willing to delete "useless" declarations as a followup.

PR #4464, opened 2021-08-06
Remove obsolete tools/scripts/bump_sorbet.sh

aprocter-stripe:
### Motivation
This came over in the compiler merge, but we don't need it anymore because of, well, the compiler merge. :)

### Test plan
n/a, just deletes code. (I know, famous last words.) FWIW I also grepped Stripe's codebase to make sure there wasn't some weird lingering reference to this script.


jez (approved):
🔥 🎉

PR #4466, opened 2021-08-09
unbreak compiler exp re-generation by creating a separate script

froydnj:
### Motivation

When we merged in the compiler repo, we copied the compiler's `update_testdata_exp.sh` script into the existing `update_testdata_exp.sh` script.  Unfortunately, this copying didn't take into account that the binary for the compiler lives in a slightly different place in the build tree -- `bazel-bin/compiler/sorbet` vs. `bazel-bin/main/sorbet`.  Which means that regenerating the compiler's `.exp` files is busted.

This PR seeks to remedy that by creating a separate `update_compiler_exp.sh` script that is more-or-less an exact copy of the original compiler's `update_testdata_exp.sh` flag, but with the path to sorbet modified appropriately.

I briefly considered just changing the path in the existing `update_testdata_exp.sh` and calling it a day, but this presents a problem: doing that means that people working on Sorbet only now have to build the compiler (and all of LLVM) to update testdata files, which doesn't seem very nice.  We could make `update_testdata_exp.sh` perform some gymnastics to only build `//compiler:sorbet` under certain circumstances, but that seems prone to errors.

The most reasonable path forward seemed to be having a separate script for the compiler's `.exp` files...and those files along with said script, will ideally go away altogether someday when we have completely transitioned to filecheck-style tests.

### Test plan

See included automated tests.


elliottt (approved):
Looks good to me! I was wondering about including a call to it in `tools/scripts/update_exp_files.sh`, but your reasoning about it building the compiler when someone isn't working on it makes sense. I think that we could probably detect when it's being run with `tools/scripts/remote-script` and run it then, but that's definitely not necessary right now.

PR #4468, opened 2021-08-09
rework tag initialization slightly

froydnj:
### Motivation

The way `sorbet_initializeTag` and `sorbet_processThrowReturnSetJmp` are written make at least `sorbet_initializeTag` tricky to reuse.  A new caller must remember to handle some of the tag reset bits and tag push bits performed by `sorbet_processThrowReturnSetJmp`.  I think it would be easier to understand if `sorbet_initializeTag` worked a little more like `EC_PUSH_TAG` + `EC_EXEC_TAG`.

Thus this PR, which moves more of the tag handling into `sorbet_initializeTag`.  I'm not too worried about performing extra branches to test `state`: jump threading is a thing and LLVM is pretty smart.  A new user of `sorbet_initializeTag` now can handle the return value directly as a clue to what kind of processing they need to do, and they don't have to remember to push or pop tags as necessary.

### Test plan

See included automated tests.  I need to fixup some of the exp tests, which will rely on #4466 landing.

### DIFF
--- a/compiler/IREmitter/Payload/codegen-payload.c
+++ b/compiler/IREmitter/Payload/codegen-payload.c
@@ -147,9 +147,9 @@ SORBET_ALIVE(VALUE, sorbet_magic_mergeHashHelper, (VALUE, VALUE));
 SORBET_ALIVE(VALUE, sorbet_vm_getivar, (VALUE obj, ID id, struct iseq_inline_iv_cache_entry *cache));
 SORBET_ALIVE(void, sorbet_vm_setivar, (VALUE obj, ID id, VALUE val, struct iseq_inline_iv_cache_entry *cache));
 
-SORBET_ALIVE(int, sorbet_initializeTag, (struct rb_vm_tag * tag));
+SORBET_ALIVE(enum ruby_tag_type, sorbet_initializeTag, (struct rb_vm_tag * tag));
 SORBET_ALIVE(VALUE, sorbet_processThrowReturnSetJmp,
-             (int setjmp_retval, rb_control_frame_t *cfp, struct rb_vm_tag *tag));
+             (enum ruby_tag_type state, rb_control_frame_t *cfp, struct rb_vm_tag *tag));
 SORBET_ALIVE(void, sorbet_teardownTagForThrowReturn, (struct rb_vm_tag * tag));
 
 SORBET_ALIVE(void, sorbet_vm_register_sig,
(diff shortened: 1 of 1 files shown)


compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -2315,49 +2315,60 @@ void sorbet_throwReturn(VALUE retval) {
     sorbet_ec_jump_tag(ec, ec->tag->state);
 }
 
-// This is invoked at the beginning of a method's body, and is analogous to EC_PUSH_TAG followed by the setjmp in
-// EC_EXEC_TAG:
+// This is invoked at the beginning of a method's body, and is analogous to EC_PUSH_TAG + EC_EXEC_TAG
```
elliottt, line 2318: Would you mind extending this comment to say that it's safe to return after the call to setjmp because this function will always be inlined?

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -147,9 +147,9 @@ SORBET_ALIVE(VALUE, sorbet_magic_mergeHashHelper, (VALUE, VALUE));
 SORBET_ALIVE(VALUE, sorbet_vm_getivar, (VALUE obj, ID id, struct iseq_inline_iv_cache_entry *cache));
 SORBET_ALIVE(void, sorbet_vm_setivar, (VALUE obj, ID id, VALUE val, struct iseq_inline_iv_cache_entry *cache));
 
-SORBET_ALIVE(int, sorbet_initializeTag, (struct rb_vm_tag * tag));
+SORBET_ALIVE(enum ruby_tag_type, sorbet_initializeTag, (struct rb_vm_tag * tag));
```
elliottt, line 150: Could we replace this with a use of `KEEP_ALIVE` after the definition of `sorbet_initializeTag`?

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -147,9 +147,9 @@ SORBET_ALIVE(VALUE, sorbet_magic_mergeHashHelper, (VALUE, VALUE));
...
 SORBET_ALIVE(VALUE, sorbet_processThrowReturnSetJmp,
-             (int setjmp_retval, rb_control_frame_t *cfp, struct rb_vm_tag *tag));
+             (enum ruby_tag_type state, rb_control_frame_t *cfp, struct rb_vm_tag *tag));
```
elliottt, line 152: Same for this function. If it's defined in the codegen payload, I don't think we need to use `sorbet_alive`.

elliottt (approved):
Thanks for this refactoring! I like that the re-push case is handled right at the beginning now 👍 

I had a few requests for changes, but none of them are blocking.

compiler/IREmitter/Payload/codegen-payload.c
froydnj: I can, but I confess I don't understand the rationale for doing so.  Why wouldn't it be safe to return after a `setjmp`?  (And if it is unsafe to return after a `setjmp`, why wouldn't that same rationale apply to any function that calls `setjmp` anywhere else in all of `libruby.so`?)

compiler/IREmitter/Payload/codegen-payload.c
froydnj: I've used `KEEP_ALIVE` for the two functions I modified here and for `sorbet_teardownTagForThrowReturn`.

compiler/IREmitter/Payload/codegen-payload.c
froydnj: Oh, I see, the `setjmp_buf` is really only valid for the lifetime of the function in which you call `setjmp`.  OK, that makes sense, I can add a comment for that.

PR #4469, opened 2021-08-09
Fix run_compiled.sh and run-benchmarks.rb

neilparikh:
### Motivation

Scripts were failing due to LLVM IR emitting change (#4429) as well as move to this repo.

### Test plan

N/A


aprocter-stripe (approved):
Thanks for the fix!

PR #4470, opened 2021-08-09
Remove Chalk symbols and rbi

ngroman:
Remove `chalk.rbi` and any entering of symbols for `Chalk::*` from Sorbet. This will help support `--stripe-packages` mode code because `Chalk::ODM` is itself a package.

Replaces `soft_freeze_logic` with a call to `T::Configuration.prop_freeze_handler().call(<self>, prop_name)`. (Default is a noop lambda).

Had to update a few tests to include stub definitions of `Chalk::ODM` related classes. This would break anyone trying to use `encrypted_prop` outside Stripe, but this should not be an issue. Since that requires a lot of specialized code.

Note: this will require a lock-step change to upgrade the version of sorbet in Stripe code.

### Motivation
Support `--stripe-packages` mode. (Also cleans up the code)

### Test plan

- [ ] Tested integration with stripe code

See included automated tests.


nroman-stripe:

→ https://go/builds/bui_K0l3LKnRkQEBO8
→ https://go/builds/bui_K0l3Isny3OogqC

nroman-stripe:

→ https://go/builds/bui_K1634um9zbywUY
→ https://go/builds/bui_K163aPtGJJikLV

ngroman:
Here's a stripe code build that integrates this + runtime changes: https://go/builds/bui_K16s8f8rgVMDN7

PR is number 358574

ngroman:
Updated build with compiler enabled https://go/builds/bui_K17U909kviNnGx

froydnj (approved):
This looks good, thank you for testing it with the compiler too!

I wonder whether `prop_freeze_handler` would be better handled as some sort of decorator method in the props machinery rather than a global configuration hook, but I suppose we can change that out at some later date.

ngroman:
> I wonder whether `prop_freeze_handler` would be better handled as some sort of decorator method in the props machinery rather than a global configuration hook, but I suppose we can change that out at some later date.

My understanding is that rewriters don't get to look at the global config (for cacheability reasons), precluding that strategy.

PR #4473, opened 2021-08-10
ensure that ensure blocks get run even when being unwound through

froydnj:
### Motivation

We were doing the wrong thing for `ensure` blocks in the presence of non-exception-related non-local exits.  This PR fixes that mistake by arranging for the bulk of exception handling logic to be guarded with an `rb_vm_tag` that will catch `longjmp`s through that region.

### Test plan

See included automated tests.


elliottt (approved):
Looks great!

PR #4474, opened 2021-08-10
update bits around update_compiler_exp.sh

froydnj:
### Motivation

Our documentation didn't mention this script, nor did anything enforce that the script needed to be run on Linux.  Let's fix that.

I also added an informative message to `test/validate_exp.sh` about what you need to do if you see failures.  You can see that message in action here:

https://buildkite.com/sorbet/sorbet/builds/16751

It would be nice to get it to show up pretty and styled in the buildkite output, but I am not sure how to do that.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/README.md
+++ b/compiler/README.md
@@ -206,8 +206,8 @@ idea is that you should skim the diffs, make sure they look appropriate given
 what was meant to change, and then record the new exp files with this command:
 
 ```bash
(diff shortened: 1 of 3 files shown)


aprocter-stripe (approved):
LGTM! Agreed that pretty buildkite output would be nice, but we can always do that in a separate PR.

PR #4475, opened 2021-08-10
run app_raise.rb 10_000_000 times to line up with a benchmark baseline

froydnj:
### Motivation

@elliottt and I were talking about this benchmark the other day and the weirdness of the benchmark with respect to our current benchmarking baselines came up.  This PR fixes that.

Just for kicks, the numbers:

```
froydnj@pyro:~/src/sorbet$ ./tools/scripts/run_benchmarks.sh -b test/testdata/ruby_benchmark/stripe/while_10_000_000.rb -f test/testdata/ruby_benchmark/app_raise.rb
[...]
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]
ruby vm startup time: .094
stripe/while_10_000_000.rb      .223    .091
app_raise.rb    4.493   .092
app_raise.rb - baseline 4.270   .001
```

I feel pretty good about the compiler's exception performance for now; the IR looks like we're not taking any shortcuts.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/ruby_benchmark/app_raise.rb
+++ b/test/testdata/ruby_benchmark/app_raise.rb
@@ -2,10 +2,12 @@
 # typed: strict
 # compiled: true
 i = 0
-while i<300000
-  i += 1
+while i < 10_000_000
   begin
     raise
   rescue
   end
+  i += 1
 end
+
+p i


elliottt (approved):
Nice, thanks for changing this!

PR #4476, opened 2021-08-10
Upgrade to bazel-4.1

elliottt:
Upgrade to bazel 4.1, fixing some escaping issues in build-ruby.bzl.

### Motivation

Keeping up with bazel releases.

### Test plan

See included automated tests.

### DIFF
--- a/third_party/externals.bzl
+++ b/third_party/externals.bzl
@@ -4,6 +4,13 @@ load("//third_party/cargo:crates.bzl", "raze_fetch_remote_crates")
 # We define our externals here instead of directly in WORKSPACE
 def register_sorbet_dependencies():
     http_archive(
+        name = "platforms",
+        urls = _github_public_urls("bazelbuild/platforms/archive/d4c9d7f51a7c403814b60f66d20eeb425fbaaacb.zip"),
+        sha256 = "a5058ac93023092c406432ec650f30ec5a8c75d8b9d13c73150f60a9050a5663",
+        strip_prefix = "platforms-d4c9d7f51a7c403814b60f66d20eeb425fbaaacb",
+    )
+
+    http_archive(
         name = "doctest",
         urls = _github_public_urls("onqtam/doctest/archive/7d42bd0fab6c44010c8aed9338bd02bea5feba41.zip"),
         sha256 = "b33c8e954d15a146bb744ca29f4ca204b955530f52b2f8a895746a99cee4f2df",
--- a/third_party/externals.bzl
+++ b/third_party/externals.bzl
@@ -3,6 +3,10 @@ load("//third_party/cargo:crates.bzl", "raze_fetch_remote_crates")
 
 # We define our externals here instead of directly in WORKSPACE
 def register_sorbet_dependencies():
+
+    # At some point the builtin @platforms package willbe removed, and we'll no longer be able to refer to
+    # @platforms//os:macos etc. The long-term workaround for this is to depend directly on bazelbuild/platforms as
+    # @platforms. See https://github.com/bazelbuild/bazel/issues/8622 for more information.
     http_archive(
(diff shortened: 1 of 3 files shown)


third_party/externals.bzl
```diff
@@ -3,6 +3,13 @@ load("//third_party/cargo:crates.bzl", "raze_fetch_remote_crates")
 
 # We define our externals here instead of directly in WORKSPACE
 def register_sorbet_dependencies():
+    http_archive(
+        name = "platforms",
+        urls = _github_public_urls("bazelbuild/platforms/archive/d4c9d7f51a7c403814b60f66d20eeb425fbaaacb.zip"),
```
froydnj, line 8: bazel 4.1 has decided to split definitions of platforms out into a separate tarball?  o.O

(I mean, I think I can understand why you would want to do that, but still seems a little weird.)

third_party/ruby/build-ruby.bzl
```diff
@@ -76,7 +76,7 @@ run_cmd() {{
     fi
 }}
 
-run_cmd sed -i.bak -e 's@^COMMONOBJS\( *\)=@COMMONOBJS\\1= {extra_srcs_object_files}@' common.mk
+run_cmd sed -i.bak -e 's@^COMMONOBJS\\( *\\)=@COMMONOBJS\\1= {extra_srcs_object_files}@' common.mk
```
froydnj, line 79: Do you know why we're so particular about keeping the exact number of spaces here?  It seems like we could drop all but one with no loss of functionality...unless there's something in Ruby's build system that depends on potential multiple spaces here?

froydnj (approved):
Comments below are non-blocking, but it would be good to make sure that this builds on Stripe's CI before landing just in case there are weird bazel version interactions.  r=me assuming a Stripe CI build works.

third_party/externals.bzl
elliottt: This is definitely weird, and is caused by something in the sorbet repository (making a fresh empty bazel project that uses bazel-4.1 doesn't have the problem with macos builds that pinning the platforms package fixes). I think it would be great if we didn't need this, but one of our dependencies is messing with it :(

third_party/ruby/build-ruby.bzl
elliottt: I believe we're keeping the spaces so that we match the correct part of the makefile -- matching the space was the easiest way to ensure that we rewrite the right declaration. Maybe a better option would be to rewrite based on the line number, as it's patching something that's not generated by autotools?

third_party/externals.bzl
elliottt: OK, reading deeper I found a link to this issue: https://github.com/bazelbuild/bazel/issues/8622. Apparently the future is to depend on the platforms repository directly if you need it, and that if you rely on `@platforms` without it you'll get whatever snapshot happens to be included with that release of bazel. I think it might be worth putting `--incompatible_use_platforms_repo_for_constraints` in our .bazelrc to make this explicit.

elliottt: Turns out we can't turn on `--incompatible_use_platforms_repo_for_constraints` yet as it breaks the version of abseil that we're using. At some point it would be great to iterate on updating dependencies so to the point that we can turn that flag on in .bazelrc.

third_party/ruby/build-ruby.bzl
elliottt: Maybe `^COMMONOBJS[^=]*=` would be better here? Experimenting on `common.mk` directly, that does seem to match exactly one line.

elliottt:

→ https://go/builds/bui_K1PhvNCAmUdMxe
→ https://go/builds/bui_K1PhmMHj10uioj

PR #4477, opened 2021-08-11
Fail for llvm parsing errors in diff-diff.rb

elliottt:
When llvm-diff fails because of parsing errors, diff-diff.rb should fail as well. This PR fixes this bug in diff-diff.rb and adds a disabled failing test for exp validation to verify that it's working.

### Motivation

Fixing a bug in validate_exp.sh that was not reporting llvm parse errors as test failures.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/compiler/disabled/validate_exp_tests/test.llo.exp
+++ b/test/testdata/compiler/disabled/validate_exp_tests/test.llo.exp
@@ -1,0 +1,234 @@
+; ModuleID = 'payload'
+source_filename = "llvm-link"
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-unknown-linux_gnu"
+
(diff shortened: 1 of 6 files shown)


test/testdata/compiler/disabled/validate_exp_tests/test.llo.exp
```diff
@@ -0,0 +1,234 @@
...
+  unreachable
+}
+
+; Function Attrs: sspreq
+define void @Init_test() local_unnamed_addr #3 {
+entry:
+  %locals.i.i = alloca i64, i32 0, align 8
+
+  ; Call a function that doesn't exist to force a error that won't show up as a diff
+  %realpath = tail call i64 @sorbet_readRealpath_fake()
```
aprocter-stripe, line 140: So it looks like we have manually injected an error here. Will `update_compiler_exp.sh` clobber this if run?

test/validate_exp.sh
```diff
@@ -81,14 +101,23 @@ for ext in "${exts[@]}"; do
         if grep "exists only in" "$diff_out" > /dev/null ; then
           cat "$diff_out"
           info "If this was an expected difference, you need to run tools/scripts/update_compiler_exp.sh"
-          fatal "* $(basename "$exp")"
+          something_failed
```
aprocter-stripe, line 104: Might be nice to pass `"$exp"` into `something_failed` and inject it into the error/success message there.

aprocter-stripe (approved):
Looks good to me, a couple questions you might want to consider before merge.

test/validate_exp.sh
elliottt: That seems totally reasonable 👍

test/testdata/compiler/disabled/validate_exp_tests/test.llo.exp
elliottt: That's a really good point, I wonder if the best way to do this is to just exempt the `disabled/validate_exp_tests` path from exp updates?

aprocter-stripe: I think that probably makes sense! Racking my brain a bit, I can't imagine the use of updating `exp` files for a disabled test.

elliottt: I took that approach for now. If there's a better way to maintain the hand-edits, we can make that change in the future :)

PR #4483, opened 2021-08-11
ide: report unused code to editor with special diagnostictag

### LINKED ISSUE #4482: IDE: Report unused code to editor with special DiagnosticTag
### Problem

Dead code is rendered with red squiggles.

### Proposed solution

We should report dead code with the new `Unnecessary` tag:

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnostic

> Clients are allowed to render diagnostics with this tag faded out instead of having an error squiggle.

We can still make it an error, but it will look better.

apratti-stripe:
#### Problem
There are two issues:
  1. Dead code is rendered with red squiggles, which is too much visual signal.
  2. Only the inner most expression in marked as dead code.
#### Solution
  1. Add Unnecessary tag, which enables editors to fade out the content. Note: the red underline will still be there, but more subdued.
  2. Join all locations for dead code instead using first found location.
dead_code

### Motivation
Fix #4482.

### Test plan
Relying on existing tests.


jvilk-stripe (commented):
The changes look good to me, but there are some failing tests because we've expanded the "unreachable" error's LOC (see BuildKite output).

You should update/fix those, and then we can request a second review from a Sorbet team member.

infer/inference.cc
```diff
@@ -162,6 +162,8 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
                         } else if (unreachableInstruction == nullptr) {
                             unreachableInstruction = expr.value.get();
                             locForUnreachable = expr.loc;
+                        } else {
+                            locForUnreachable = locForUnreachable.join(expr.loc);
```
froydnj, line 166: This change looks unrelated to reporting unused code via a tag, can you walk me through why we're doing this in this PR?

froydnj (changes requested):
jvilk's comments are better than anything else I could say.

infer/inference.cc
jvilk-stripe: The PR description briefly touches upon this, but prior to this change we were only using the `loc` for the first `Binding` within the dead basic block.

What this looks like right now is the following:

```ruby
if false
  method_call(10)
  #           ^^ error: dead code
  second_method_call(20)
end
```

`10` is the first `Binding` in the CFG, since we assign it to a temporary variable. We're now using this loc to dim the dead code, but dimming just that argument looks bad / buggy!

We've changed the logic to make the loc cover the _entire dead basic block_, which looks much better.

Note that further in this method, Sorbet explicitly does not emit an error on other dead blocks that follow from this block, since that would be a redundant error. We can't expand the loc to cover those, since they aren't guaranteed to be contiguous. So we still don't dim all dead code. We can address this limitation in a subsequent PR that adds a notion of "hint" errors that Sorbet doesn't emit on the command line but that LSP uses to dim the code.

jvilk-stripe: We should probably add a comment above this explaining what we are doing.
```suggestion
                        } else {
                            // Expand the loc to cover the entire dead basic block
                            locForUnreachable = locForUnreachable.join(expr.loc);
```

jvilk-stripe (approved):

froydnj (approved):

PR #4484, opened 2021-08-12
Add script to refresh sorbet_ruby patches from the patch branches

aprocter-stripe:
### Motivation
Now that we have set up https://github.com/sorbet/ruby as a central place to hold our Ruby patches, it'll be handy to have a script to regenerate the patches in the sorbet repo.

### Test plan
Ran some offline testing to verify that after running this script, the new patches look sane and the build still works. Also manually tested various scenarios wrt an already-existing `$HOME/stripe/sorbet_ruby` directory.


third_party/externals.bzl
```diff
@@ -360,14 +361,7 @@ package(default_visibility = ["//visibility:public"])
...
                 build_file = ruby_patched_build,
-                # If you're trying to use `git diff` to generate this patch, pass the `--no-prefix` flag
-                # (Removes the `a/` and `b/` prefixes that `patch` doesn't understand.)
-                patches = [
-                    "@com_stripe_ruby_typer//third_party/ruby:gc-remove-write-barrier.patch",
-                    "@com_stripe_ruby_typer//third_party/ruby:vm-method-type-sorbet.patch",
-                    "@com_stripe_ruby_typer//third_party/ruby:is-lambda-ifunc.patch",
-                    "@com_stripe_ruby_typer//third_party/ruby:init-sorbet-t-modules.patch",
-                ],
+                patches = ruby_patches(),
```
aprocter-stripe, line 364: It's kind of a bummer that `glob` isn't available here. :(

froydnj (commented):
Why are we not setting things up so that our external ruby source is `sorbet/ruby`?  Is that because we still have to have unpatched ruby for various things?

Would it break horribly if we had two different ruby upstreams, since we use them for different things?

aprocter-stripe:
> Why are we not setting things up so that our external ruby source is `sorbet/ruby`?  Is that because we still have to have unpatched ruby for various things?

Personally I wouldn't mind doing things that way! Just wasn't sure we knew exactly what we wanted yet, so this is an easy way to move the source of truth to the new fork while otherwise leaving the current build process ~undisturbed until we get the particulars nailed down.

jez:
> Why are we not setting things up so that our external ruby source is sorbet/ruby?

(fwiw this was my first thought too, happy to have this halfway step or just jump straight to the endgame)

elliottt:
> > Why are we not setting things up so that our external ruby source is sorbet/ruby?
> 
> (fwiw this was my first thought too, happy to have this halfway step or just jump straight to the endgame)

I think I was the only hold-out for the Old Way, as it was useful to be able to quickly revert one of the patches recently. Given that it wouldn't be much harder to revert a commit in sorbet/ruby and bump the dependency, I'm fine moving over to depending directly on our ruby fork 👍

aprocter-stripe:
Okay, sounds like we're all on board with depending directly on the fork! I'd be glad to take that on, either as a follow-on to this PR or in lieu of it entirely.

aprocter-stripe:
Closing in favor of the straight-from-GitHub approach (will be working on that today)

PR #4485, opened 2021-08-12
Stop running //test in test-compiler.sh

elliottt:
### Motivation

We don't need to run these tests during test-compiler.sh.

### Test plan

See included automated tests.


aprocter-stripe (approved):
Sure does what it says on the tin :)

PR #4486, opened 2021-08-12
Enforce that we use @platforms for constraints

elliottt:
Enable `--incompatible_use_platforms_repo_for_constraints` in our .bazelrc so that we get an error if we start depending on `@bazel_tools//platforms` again. Additionally, patch abseil to use the `@platforms` repository when we build it. I've opened a pull request with abseil to upstream this change at abseil/abseil-cpp#1001.

### Motivation

Ensuring that we don't fall back on using `@bazel_tools//platforms` anywhere.

### Test plan

See included automated tests.

### DIFF
--- a/third_party/externals.bzl
+++ b/third_party/externals.bzl
@@ -155,9 +155,10 @@ def register_sorbet_dependencies():
 
     http_archive(
         name = "com_google_absl",
-        urls = _github_public_urls("abseil/abseil-cpp/archive/e9b9e38f67a008d66133535a72ada843bd66013f.zip"),
-        sha256 = "49c93740b3b09f73cd2f10da778ea4129d59733085393f458a4acd17774503fb",
-        strip_prefix = "abseil-cpp-e9b9e38f67a008d66133535a72ada843bd66013f",
+        urls = _github_public_urls("abseil/abseil-cpp/archive/b699707f0bfeae034e36cdfd909b66b0fcab696c.zip"),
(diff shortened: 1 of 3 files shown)


third_party/externals.bzl
```diff
@@ -155,9 +155,10 @@ def register_sorbet_dependencies():
 
     http_archive(
         name = "com_google_absl",
-        urls = _github_public_urls("abseil/abseil-cpp/archive/e9b9e38f67a008d66133535a72ada843bd66013f.zip"),
-        sha256 = "49c93740b3b09f73cd2f10da778ea4129d59733085393f458a4acd17774503fb",
-        strip_prefix = "abseil-cpp-e9b9e38f67a008d66133535a72ada843bd66013f",
+        urls = _github_public_urls("abseil/abseil-cpp/archive/b699707f0bfeae034e36cdfd909b66b0fcab696c.zip"),
+        sha256 = "29a165fcbf802f3cdb6d7a17327ede0af2c799a11f88611c005867e994b984e2",
```
froydnj, line 159: For my own edification -- I don't think we need to block on this -- if abseil deletes your branch from their repo, does this ref eventually get GC'd?

froydnj (approved):
WFM!

third_party/externals.bzl
elliottt: This ref points to the current master -- I updated abseil at the same time as I thought they might have already addressed the issue with `@bazel_tools//platforms` dependence. The build worked, so I figured that it would be reasonable to keep the abseil update.

PR #4487, opened 2021-08-13
Switch to a single patch file, sourced from the sorbet/ruby repo

aprocter-stripe:
Moves the sorbet_ruby patch to a single patch file, and adds a script to re-generate it from a branch on the `sorbet/ruby` repo.

### Motivation
Per discussion on #4484, we'd prefer to build by pulling directly from git. Unfortunately this isn't quite as straightforward to implement as we'd hoped, since building straight from a git checkout requires some stuff (e.g., an existing Ruby install, autoconf, bison(?)...) that building from the tarball doesn't. (The release tarball has a lot of generated stuff pre-generated.) So this is at least a good stopgap until we can figure out all the details of the preferred approach.

### Test plan
See test plan comments on #4484. (This is just a simplified version of that script.)


tools/scripts/update_sorbet_ruby_patch.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
aprocter-stripe, line 1: TIL that there exists this magic `.patch` URL thing on GitHub that would *almost* work as a stand-in for this script, e.g.:

https://github.com/sorbet/ruby/compare/v2_7_2...sorbet_ruby_2_7.patch

But unfortunately this includes the `a/` and `b/` paths that we need to avoid by passing `--no-prefix`, so I don't think we can use it directly here.

elliottt: You could use the `patch_args` argument to `http_archive` to pass in `-p 1`.

aprocter-stripe: Ooh, didn't realize `patch_args` was a thing. I'll give it a shot.

aprocter-stripe: Seems to work! I went ahead and pushed a change to the script.

We could probably actually make use of `remote_patch` to avoid the need to store the patch in the repo entirely. (Might still want a bump script, but it would just need to update a git SHA and a checksum.) What do you think of that option?

aprocter-stripe: Ah, never mind --- apparently that's not available in 4.1.0.

elliottt (approved):
Nice, I like the patch url feature!

aprocter-stripe:

→ https://go/builds/bui_K2BiKbc3mm6pgi
→ https://go/builds/bui_K2BiI86rDoJh8d

PR #4488, opened 2021-08-13
Force exception variables to escape

elliottt:
Force exception values to live on the ruby stack by adding an artificial reference in a different ruby block.

### Motivation

We need to ensure that exception variables are stored in the ruby stack for fixing issues with exception handling.

### Test plan

See included automated tests.


froydnj (approved):
This is more elegant than I was imagining!

elliottt:

→ https://go/builds/bui_K2BeILYRHJGSkT
→ https://go/builds/bui_K2Be5gVT3Jp6yk

PR #4492, opened 2021-08-13
Issue 4489

karmakaze:
### Test plan

See included automated tests.


karmakaze:
*I'm very new to the codebase, and had a question about `lub` in general.* It's how `Any/Or` gets implemented, and I'm guessing the name (used?) to mean 'least-upper-bound' which if we think of up as super, then would apply for computing underlying types. The computation for the union is more involved than that for the underlying as it reduces less (maintains more granularity). So my question is how can we accurately compute both the union and the underlying type with a recursive, single-return value function? It seems that the historical implementation including the present often favours making the union result match the underlying, rather than preserving distinct union cases.

froydnj:
This PR looks jez-shaped.  I'm not entirely sure of the history behind not constructing `OrType` for or-ing shapes together.  My guess is that the idea was to prevent types from growing unreasonably large in the presence of many small but different shape types, but that's only a guess.

core/types/subtyping.cc
```diff
@@ -404,28 +404,35 @@ TypePtr Types::lub(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2)
...
+                    return t1;
+                } else if (t2LubT1Right == t2) {
+                    return OrType::make_shared(orType1->left, t2);
+                }
+            } else {
+                return OrType::make_shared(t1, t2);
+            }
         } else {
-            return lub(gs, t1, und);
+            return OrType::make_shared(t1, t2);
```
karmakaze, line 435: I have to rethink this entire `// only 2nd is proxy` case to see if it applies specifically to this `T.any(Shape, Shape)` PR. I had a number of changes in the area of Tuples, Shapes, Hashes and tried best to split them up into small changes associated with specific issues.

This block would change the result of 'proxy / non-proxy' interactions making `T.any(...)` e.g. for 'Tuple / non-Tuple/Shape' as well as 'Shape / non-Shape/Tuple'.

It could be a separate PR for the case where a Shape (or Tuple) is being Or-combined (lub'ed) with a Shape (or Tuple) and one is a supertype of the other.

core/types/subtyping.cc
```diff
@@ -404,28 +404,35 @@ TypePtr Types::lub(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2)
...
             // only 1st is proxy
-            TypePtr und = t1.underlying(gs);
-            if (isSubType(gs, und, t2)) {
-                return t2;
-            } else if (allowProxyInLub) {
-                return OrType::make_shared(t1, t2);
-            } else {
-                return lub(gs, t2, und);
-            }
+            return lub(gs, t2, t1);
```
karmakaze, line 408: This won't work, because we already reverse/recurse on `t1.kind() > t2.kind()`. https://github.com/sorbet/sorbet/pull/4495#discussion_r689782209

core/types/subtyping.cc
```diff
@@ -470,6 +454,31 @@ TypePtr Types::lub(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2)
     return lubGround(gs, t1, t2);
 }
 
+TypePtr lubProxyNonProxy(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2) {
```
karmakaze, line 457: Is there a way to either statically type or (in debug build) assert t1 is proxy and t2 is non-proxy?

core/types/subtyping.cc
```diff
@@ -470,6 +454,32 @@ TypePtr Types::lub(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2)
...
+    bool allowProxyInLub = isa_type<TupleType>(t1) || isa_type<ShapeType>(t1);
+    TypePtr und = t1.underlying(gs);
+    if (Types::isSubType(gs, und, t2)) {
+        return t2;
+    } else if (allowProxyInLub) {
+        auto *orType = cast_type<OrType>(t2);
+        if (orType != nullptr) {
```
karmakaze, line 464: What we really want to do in this block is produce an OrType tree replacing a node with t1, if t1 is a supertype of that node. This is where a refactor like #4496 would be very handy.

karmakaze: Using `lubDistributeOr` to do the thing.

core/types/subtyping.cc
karmakaze: Extracted into friend `lubProxyNonProxy()`

PR #4493, opened 2021-08-13
ensure that ensure blocks get run even when being unwound through, take 2

froydnj:
### Motivation

#4473 contains the original motivation, which is to make `ensure` blocks robust in the presence of non-exception non-local control flow (e.g. `return` or `break`).  We reverted that patch in #4480 because of crashes in Stripe's codebase related to the compiler.

This PR turns the support back on, is more invasive, and fixes potential problems in the compiler's return-from-block support to boot.

The debugging was initially fruitless: we had (Ruby and C!) stacktraces from the crashes in Stripe's codebase, but setting up a reduced testcase that replicated the pattern of interpreted/compiled code + blocks + control flow didn't reproduce crashes.  The fun really started when we looked at `sorbet_initializeTag`:

https://github.com/sorbet/sorbet/blob/556591293648c7b44706075a8a74314f6ab24a6c/compiler/IREmitter/Payload/codegen-payload.c#L2322-L2349

which refers to `rb_ec_tag_state`:

https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/eval_intern.h#L160-L167

and `rb_ec_tag_state`'s use in `EC_EXEC_TAG`:

https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/eval_intern.h#L181-L182

Notice that `EC_EXEC_TAG` has some weird `VAR_FROM_MEMORY` macro on `rb_ec_tag_state`'s argument, which `sorbet_initializeTag` does not replicate.  Huh, that's weird.  Why don't we try to make our code more like Ruby's code?

Lo and behold, once we did that, our original testcases started crashing, somewhat unexpectedly, in ways that looked almost exactly identical to the crashes we observed in Stripe's codebase.

The original thought, which you can see in 3b3a9b5, was that we weren't adding proper lifetime annotations for the `rb_vm_tag` that our `ensure` handling required.  The tag lifetime is not obvious from the read/write behavior of the code: the last write is very early on in the variable's lifetime, but the actual place the variable "dies" is much later, not until almost the end of the exception handling region.  But that didn't fix the issue.  (Note also that once I implemented the fix below, I commented out the lifetime annotations and verified that they did not seem to be having an effect.  It makes me sleep a little better at night having them there though, even if they're commented out.)

Debugging turned up the following problem: we would generate code for a given exception-handling region.  In the `longjmp` path in `sorbet_initializeTag`, LLVM was accessing `ec` from the same location on the stack that `ec` was being accessed in the non-`longjmp` path.  Except that stack location had already gotten shared and overwritten by the `struct ExceptionClosure` that was used by `sorbet_try` to execute the body of the exception-handling region.  So when the body eventually wound its way to a `return`, we'd `longjmp` back, eventually find ourselves on the `longjmp` path mentioned earlier, access what we thought was `ec`, and boom, some kind of segfault.

(Interesting semi-relevant digression: `RUBY_SETJMP` is typically not implemented via `setjmp(3)`.  Instead, the default choice is to implement it with `__builtin_setjmp`:

https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/tool/m4/ruby_setjmp_type.m4

Which you would think is roughly equivalent.  But no!  `__builtin_setjmp` does the barest minimum so that you can implement `setjmp`/`longjmp`-based exception handling for C++ -- the original exception handling implementation style.  Your glibc implementation of `setjmp` for x86-64 Linux will save a bunch of registers:

https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/setjmp.S;h=93a7b12454eb7cf1c65cf925575f480540e4104c;hb=HEAD

whereas the `__builtin_setjmp` implementation for x86-64 Linux will only save `%rsp`, `%rbp`, and `%rip`:

https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/X86/X86ISelLowering.cpp

Sorry, github can't show you 2MB files; you want to search for `emitEHSjLjSetJmp`.

And, in particular, Clang knows that `setjmp(3)` can "return twice" (and propagates that information down to LLVM), whereas Clang/LLVM has no such knowledge for `__builtin_setjmp`.  And other parts of LLVM know about functions that can "return twice" and take steps to limit the amount of stuff they do in the face of such peculiar behavior.  So if you are using `__builtin_setjmp`, none of these steps actually kick in, which is what was biting us above.  It's *possible* that our issues would have gone away if we had forced Ruby to use `setjmp(3)`, but that change would come with performance penalties and I didn't want to walk down that road right now.

This difference between `setjmp` and `__builtin_setjmp` is extremely weird, but I (speculatively) think that this is justified by some reasoning like "people who use `__builtin_setjmp` Know What They Are Doing and will take appropriate steps" -- i.e. `volatile` hacks everywhere or somesuch.  And to be fair, `__builtin_setjmp` was meant to be a compiler-internal thing anyway for implementing exception handling in C++, not a faster `setjmp(3)` replacement.  Anyway, back to the main explanation.)

Presumably this was happening in Stripe's codebase and not in our testcases because of the vagaries of LLVM's optimization pipeline, complexity of the original Ruby CFG, etc. etc.

The plan to fix this seemed relatively simple: we'd force the execution context to be cached in memory ourselves across the entire exception handling region.  We'd then load it "volatile-y" every time we needed to access it and pass it into the requisite C functions in the codegen payload.  This is not dissimilar to what some Ruby code already does; compare the definition of `rb_vrescue2` for instance:

https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/eval.c#L977-L1030

where `ec` (indeed, a number of variables) are declared `volatile`, presumably for very similar reasons to what's happening here.

This idea (implemented in 2e0ea66 and 80a307f) worked great.  Except that, for whatever reason, other tests related to return-from-block started failing.  OK, fine, all of the code for return-from-block has exactly the same set of issues, so we need to fix that bit of codegen too (9e090c4 and 9cb8026 as well as 1890067 and 143c344, the latter of which clean up some temporarily-inserted functions so I could test ideas on exception handling without also having to touch return-from-block right away).

That's all well and good, except that we still have the gross hacks that led us to having a reproducible test case in the first place.  Perhaps we could resolve those hacks now?  We tried that, but then discovered that `sorbet_initializeTag` still suffered from the same sort of issues. The fix + gross hacks removal, implemented in ffe97bb, was to declare that `sorbet_initializeTag` didn't take an execution context directly, but rather a place where an execution context was stored, which we can declare has the proper `volatile` semantics that will make the C compiler do what we want and ensures that the `ec` used across the entire exception region is only ever stored in one place.

I added the original testcase I tried writing in 4c6f78f; I don't claim that it's a perfect model of testcase writing and I don't think it is radically different than the existing return-across-ensure testcase, but it did at least induce crashes at one point.

### Test plan

See included automated tests.  I will of course be testing this on Stripe's codebase.


test/run_sorbet.sh
```diff
@@ -101,7 +101,7 @@ fi
 if [ -z "$debug" ]; then
   command=( "bazel-bin/compiler/sorbet" )
 else
-  command=( "${LLDB:-lldb}" "--" "./bazel-bin/compiler/sorbet" )
```
aprocter-stripe, line 104: This change seems unrelated. Can we revert it?

froydnj:

→ https://go/builds/bui_K2BLGjfkN4v9vt
→ https://go/builds/bui_K2BLTMkq6LT6x4

aprocter-stripe (commented):
Oh man!
```c
# define VAR_FROM_MEMORY(var) __extension__(*(__typeof__(var) volatile *)&(var))
```

aprocter-stripe:
:facepalm: Apparently "shift-enter" while you're editing a comment just summarily summit that and all your other pending comments :)

test/run_sorbet.sh
froydnj: :shame:

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -2319,33 +2324,45 @@ void sorbet_throwReturn(VALUE retval) {
 // because this function will be inlined, meaning that `tag->buf` will reflect the
 // state of this function's caller and `tag->buf` will be valid for the lifetime of
 // this function's caller.
+//
+// The `volatile` here is a bit gross.  If you look at `rb_ec_tag_state` and its
+// accompanying logic, referenced below, you'll notice that it contains some
+// gross hacks to force clang to stick `ec` in memory before accessing it and that
```
aprocter-stripe, line 2330: Is `VAR_FROM_MEMORY` the hack you're referring to here? (Might give a pointer to it if so.)

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -2319,33 +2324,45 @@ void sorbet_throwReturn(VALUE retval) {
...
-    if (setjmp_retval) {
+    if (RUBY_SETJMP(tag->buf) != 0) {
         // See rb_ec_tag_state:
         // https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/eval_intern.h#L160-L167
-        state = ec->tag->state;
-        ec->tag->state = TAG_NONE;
+        enum ruby_tag_type *statePtr = &(*ec)->tag->state;
+        enum ruby_tag_type state = *statePtr;
+        *statePtr = TAG_NONE;
+        return state;
```
aprocter-stripe, line 2359: Is `statePtr` here just for efficiency's sake to avoid repeating the indirection through `ec` and `ec->tag`, or is `volatile`-ity somehow part of the motivation?

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -2319,33 +2324,45 @@ void sorbet_throwReturn(VALUE retval) {
...
         // See EC_REPUSH_TAG:
         // https://github.com/ruby/ruby/blob/5445e0435260b449decf2ac16f9d09bae3cafe72/eval_intern.h#L144
-        ec->tag = tag;
+        (*ec)->tag = tag;
+        return TAG_NONE;
     }
-
-    // This is subtle, but tags are organized in such a way that TAG_NONE is 0
-    // (i.e. a "normal" return from setjmp) and every other tag is non-zero
-    // (i.e. a "abnormal" return from setjmp).
```
aprocter-stripe, line 2347: I like the incidental cleanup here which gets rid of this weirdness :)

test/testdata/compiler/exceptions/ensure_crash__1.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
```
aprocter-stripe, line 1: This test is a reproducer (we think) for the segfaults we got from #4473, right? Could we add a comment to that effect here, and maybe move it to `testdata/compiler/regression` (or, idk, maybe `testdata/compiler/exceptions/regression`? :))

aprocter-stripe (approved):
Excellent writeup, and I think this all makes sense to me! Would like give @elliottt the opportunity to give it a once-over as well, so will request further review from him.

At some point it might be nice to reduce some of the duplication between the return-from-block state/codegen and the exception state/codegen, but the duplication is pretty minor and per standup discussion this morning we may be considering another option anyway, so let's not worry about it for now.

froydnj:
OK, we have moved the wip from #4502 to this PR.  As detailed in #4520, there are significant issues with having calls to `__builtin_setjmp` and `__builtin_longjmp` in the codegen payload, due to compiler incompatibilities that manifest only with certain compilation flags.  So the new approach moves the entirety of `sorbet_run_exception_handling` into the VM.  This is not ideal, as we would like to inline some of the exception handling code into the relevant areas and even remove calls to exception handling code that wasn't present in the original Ruby source (e.g. no point in executing synthesized `rescue` blocks if they're not actually going to catch exceptions).

But we do gain a little bit from moving it out of line: we don't have to worry about compiler issues surrounding the need for certain variables for the exception handling process to live in memory, or worry about the variables for _inlined_ blocks perhaps needing to live in memory.  And the rewrite in C is, I think, more easily understandable, better commented, and easier to modify in the future should we decide to change anything.

froydnj:

→ https://go/builds/bui_K5qPK3YjA0fxbN
→ https://go/builds/bui_K5qPnVxLpYvbe5

compiler/IREmitter/Payload/patches/vm_insnhelper.c
```diff
@@ -24,6 +24,9 @@ void sorbet_setExceptionStackFrame(rb_execution_context_t *ec, rb_control_frame_
     // NOTE: there's no explicit check for stack overflow, because `vm_push_frame` will do that check
     cfp = vm_push_frame(ec, iseq, VM_FRAME_MAGIC_RESCUE, self, blockHandler, me, iseq->body->iseq_encoded, sp,
                         num_locals, iseq->body->stack_max);
+
+    ec->tag->state = TAG_NONE;
+    ec->errinfo = Qnil;
```
elliottt, line 29: A comment for why this is necessary would be great.

compiler/IREmitter/Payload/patches/vm_insnhelper.c
```diff
@@ -481,3 +484,232 @@ void sorbet_throwReturn(rb_execution_context_t *ec, VALUE retval) {
...
+SORBET_INLINE
+static VALUE sorbet_readLocal(rb_control_frame_t *cfp, long index, long level) {
+    int offset = computeLocalIndex(index);
+    return *(vm_get_ep(cfp->ep, level) - offset);
+}
+SORBET_INLINE
+static void sorbet_writeLocal(rb_control_frame_t *cfp, long index, long level, VALUE value) {
+    int offset = computeLocalIndex(index);
+    vm_env_write(vm_get_ep(cfp->ep, level), -offset, value);
+}
```
elliottt, line 507: Thanks for marking all these functions `static`, it would be really bad if generated code started using them instead of the versions in `codegen-payload.c`. Maybe it would be worth adding a comment here to indicate that these are intentionally duplicated? Maybe different names would be better?

compiler/IREmitter/Payload/patches/vm_insnhelper.c
```diff
@@ -481,3 +484,232 @@ void sorbet_throwReturn(rb_execution_context_t *ec, VALUE retval) {
...
+                // rb_rescue2/rb_vrescue2 would check ec->errinfo here to determine if it
+                // was the "right" kind of error.  Sorbet has already generated code to check
+                // the exception value for us, we can dispatch directly to the handlers here,
+                // which avoids a little bit of overhead.  But we do need to tell the
+                // handlers what the exception value *is*.
+                bodyException = ec->errinfo;
+            } else {
+                // Any other kind of non-local exit will skip the rescue/else handlers.
+                goto execute_ensure;
+            }
```
elliottt, line 611:
```suggestion
            if (nleType != TAG_RAISE) {
                // Any other kind of non-local exit will skip the rescue/else handlers.
                goto execute_ensure;
            }
            
            // rb_rescue2/rb_vrescue2 would check ec->errinfo here to determine if it
            // was the "right" kind of error.  Sorbet has already generated code to check
            // the exception value for us, we can dispatch directly to the handlers here,
            // which avoids a little bit of overhead.  But we do need to tell the
            // handlers what the exception value *is*.
            bodyException = ec->errinfo;
```
Slight preference for guarding the goto rather than hiding it in the `else`.

compiler/IREmitter/Payload/patches/vm_insnhelper.c
```diff
@@ -481,3 +484,232 @@ void sorbet_throwReturn(rb_execution_context_t *ec, VALUE retval) {
...
+    if ((nleType = EC_EXEC_TAG()) == TAG_NONE) {
+    execute_body:
+        state = RunningBody;
+        // tag.state will have been reset appropriately if we got here via `retry`,
+        // but not this variable.
+        nleType = TAG_NONE;
+
+        // Clear out the local variable shared across exception handling regions
+        // where we store the current exception value.
+        sorbet_writeLocal(cfp, exceptionValueIndex, exceptionValueLevel, Qnil);
```
elliottt, line 563: Because it's the sort of optimization I can imagine us wanting us to write in the future: could you write a comment here saying that it's necessary to use `sorbet_writeLocal` here, as the locals could move to the heap between runs of this statement? I can imagine us wanting to refactor this to write to a pointer instead, and that's not going to be correct.

compiler/IREmitter/Payload/patches/vm_insnhelper.c
```diff
@@ -481,3 +484,232 @@ void sorbet_throwReturn(rb_execution_context_t *ec, VALUE retval) {
...
+            // TODO(froydnj): this is where we would handle TAG_RETRY if we were implementing
+            // `retry` in terms of Ruby's non-local exit handling rather than our current
+            // retry singleton mechanism.
+
+            if (nleType == TAG_RAISE) {
+                handlerException = ec->errinfo;
+                sorbet_writeLocal(cfp, exceptionValueIndex, exceptionValueLevel, handlerException);
+            } else {
+                goto execute_ensure;
+            }
```
elliottt, line 644: Again, I'd rather that we invert this so that only the `goto` is guarded.
```suggestion
            if (nleType != TAG_RAISE) {
                goto execute_ensure;
            }
            
            handlerException = ec->errinfo;
            sorbet_writeLocal(cfp, exceptionValueIndex, exceptionValueLevel, handlerException);
```

elliottt (changes requested):
This looks great! It's very exciting to see so many disabled tests get enabled!

froydnj:
Pushed four commits to address review points; once the PR is approved, I will rebase to fix up the merge conflicts vs. #4520 .

elliottt (approved):
Thanks for the changes, this looks great!

froydnj:

→ https://go/builds/bui_K6EzI7Ajdvrwz2
→ https://go/builds/bui_K6EzLbfy7lZf90

Issue #4490, opened 2021-08-13
T.any(Shape, Tuple) degenerates to T.any(T::Hash[T.untyped, T.untyped], T::Array[String])


karmakaze:
#### Input

[→ View on sorbet.run](TODO_replace_this_with_your_sorbet_run_link)

```ruby
# typed: true
extend(T::Sig)

sig { params(x: T.any({a: Integer}, [String])).void }
def foo(x)
  T.reveal_type(x)
  puts x
end

foo({a: 1})
foo(["ok"])
# these should all be errors
foo({a: "bad"})
foo({b: 1})
foo(["bad", "too"])
```

#### Observed output

```
[symbolize_emscripten.inc : 43] RAW: Symbolization unavailable. Rebuild with -sWASM=1 and -sUSE_OFFSET_CONVERTER=1.
editor.rb:6: Revealed type: T.any(T::Hash[T.untyped, T.untyped], T::Array[String]) https://srb.help/7014
     6 |  T.reveal_type(x)
          ^^^^^^^^^^^^^^^^
  Got T.any(T::Hash[T.untyped, T.untyped], T::Array[String]) originating from:
    editor.rb:5:
     5 |def foo(x)
                ^
Errors: 1
```

#### Expected behavior

- revealed type to be `T.any({a: Integer}, [String])`
- errors for:
  - foo({a: "bad"})
  - foo({b: 1})
  - foo(["bad", "too"])

- - -

PR #4495, opened 2021-08-14
subtyping lub: use 'only 2nd is proxy' to handle 'only 1st is proxy' case

karmakaze:
Call `lub(gs, t2, t1)` itself when only `t1` is a proxy type (leveraging the handling when only `t2` is a proxy type) as the result should be symmetric. This is to avoid duplication and potential deviation in implementation.

### Motivation

While working on #4490, I noticed an incorrect comment "// only 1st is proxy" (following a correct comment "// only 2nd is proxy") indicating a copy/paste from another case. This was originally fixed up in the PR #4491 for #4490 but realizing that more work will likely need to be done on that, split this change out on its own.

### Test plan

This is a refactoring only and is covered by passing existing automated tests.

### DIFF
--- a/core/types/subtyping.cc
+++ b/core/types/subtyping.cc
@@ -404,21 +404,12 @@ TypePtr Types::lub(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2) 
             ENFORCE(result != nullptr);
             return result;
         } else {
-            bool allowProxyInLub = isa_type<TupleType>(t1) || isa_type<ShapeType>(t1);
             // only 1st is proxy
-            TypePtr und = t1.underlying(gs);
-            if (isSubType(gs, und, t2)) {
-                return t2;
-            } else if (allowProxyInLub) {
-                return OrType::make_shared(t1, t2);
-            } else {
-                return lub(gs, t2, und);
-            }
+            return lub(gs, t2, t1);
         }
     } else if (is_proxy_type(t2)) {
         // only 2nd is proxy
         bool allowProxyInLub = isa_type<TupleType>(t2) || isa_type<ShapeType>(t2);
-        // only 1st is proxy
         TypePtr und = t2.underlying(gs);
         if (isSubType(gs, und, t1)) {
             return t1;


aprocter-stripe:

→ https://go/builds/bui_K3Gdypa0EqdOlz
→ https://go/builds/bui_K3GdboP4FZt4mS

core/types/subtyping.cc
```diff
@@ -404,21 +404,12 @@ TypePtr Types::lub(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2)
...
             // only 1st is proxy
-            TypePtr und = t1.underlying(gs);
-            if (isSubType(gs, und, t2)) {
-                return t2;
-            } else if (allowProxyInLub) {
-                return OrType::make_shared(t1, t2);
-            } else {
-                return lub(gs, t2, und);
-            }
+            return lub(gs, t2, t1);
```
aprocter-stripe, line 408: This results in infinite recursion when we hit the logic up on line 182:

https://github.com/sorbet/sorbet/pull/4495/files#diff-eb2da3e78eea17626fb563370b444cc444ebc51aa7b8cd35da2fad4e43cbdd36R182-R184

Note also that even if this issue weren't present, all the logic in the runup to the `is_proxy_type(t1)` check would be repeated, which could have perf implications. So I think we might have to live with the repetition here, or come up with a different strategy to DRY it.

aprocter-stripe (changes requested):
Thanks for the PR! Unfortunately I don't think it will work as is---please see comment.

core/types/subtyping.cc
karmakaze: Ah thanks :+1:, I suspected such a thing was in place as I was only seeing certain sequences and not others. I'll think about other ways to DRY this up.

karmakaze:
This PR was an attempt to simplify/dedup in preparation to make #4492 smaller. https://github.com/sorbet/sorbet/pull/4492/files#diff-eb2da3e78eea17626fb563370b444cc444ebc51aa7b8cd35da2fad4e43cbdd36L407-R430

Closing this PR, and dealing with any simplification in #4492 on its own.

PR #4497, opened 2021-08-14
Naive attempt at starting #11

karmakaze:
*A newbie, diving into issue #11 to learn what it is I don't know.* 

### Motivation

Issue #11 is the most important thing to fix in my usage of Sorbet that I'm aware of--leaving it open is the greatest `false sense of security`.

The best background on the issue is in PR #3822 (closed), and this PR is beginning to retrace some of those steps to learn more.

My understanding of `lub` is that it implements `T.any` using `OrType` among other things. I'm guessing the name (used?) to mean `lowest upper bound`, if we consider 'up' to be supertypes. That makes sense for computing the `underlying` types but the function also computes the union type which should be more precise.

This PR follows the [pattern for `underlying` used by TupleType](https://github.com/sorbet/sorbet/blob/f791e18ed67d9d03d653836858b6886ab50bd82d/core/types/types.cc#L424-L430), e.g.:
```
```

### Test plan

See included automated tests.


jez:
Heads up that #11 is marked `hard` which means that it requires extensive involvement and expertise from members of the Sorbet team at Stripe to implement the fix that we have in mind and to refactor Stripe's codebase to be able to support the change. Feel free to poke around for the sake of learning but we will likely decline to merge PRs that propose solutions for #11.

karmakaze:
> Heads up that #11 is marked `hard` which means that it requires extensive involvement and expertise from members of the Sorbet team at Stripe to implement the fix that we have in mind and to refactor Stripe's codebase to be able to support the change. Feel free to poke around for the sake of learning but we will likely decline to merge PRs that propose solutions for #11.

Thanks for the note :+1:. This was also my expectation, which is why it was framed as "learn what it is I don't know". I looked at the proposed solution in #3822 (closed) and the part that's in #3823 (wip). This PR doesn't try to solve #11, only a specific subset which as great value for the codebase that I work in. Together with #3823 it would give a lot of confidence working with Hashes vs lack of checking that is largely unknown to most developers.

What I would greatly appreciate is to get some details on why certain changes are or are not feasible to guide further work. I'm hoping to at least get a report of statistics when run on a corpus with extensive type usage.

karmakaze:
@jez So I rebased this on top of #3823 (itself rebased on master) in an experimental branch to see how they would combine. Seems I rediscovered, piece by piece what was done in that original PR and this PR ends up being reducing the scope/blast-radius to only the cases where the Hash's value type is single or nilable single. Essentially:
```
TypePtr ShapeType::underlying(const GlobalState &gs) const {
    if (this->keys.empty()) {
        // We could imagine one day making this an error in `typed: strict`, and requiring type
        // annotations on empty literal hashes.
        return Types::hashOfUntyped();
    } else {
         auto keysLub = lubAllDropLiteral(gs, this->keys);
         auto valuesLub = lubAllDropLiteral(gs, this->values);
+        auto vo = cast_type<OrType>(valuesLub);
+        if (vo != nullptr && !(vo->left.isNilClass() && isa_type<ClassType>(vo->right))
+                          && !(vo->right.isNilClass() && isa_type<ClassType>(vo->left))) {
+            valuesLub = Types::untypedUntracked();
+        }
+        return Types::hashOf(gs, keysLub, valuesLub);
    }
}
```
And the `hashOf`/`hashOfSymbolKey` is a convenience factoring meaning the same as:
```
-        vector<TypePtr> tupleArgs{keysLub, valuesLub};
-        vector<TypePtr> targs{keysLub, valuesLub, make_type<TupleType>(move(tupleArgs))};
-        return make_type<AppliedType>(Symbols::Hash(), targs);
```

froydnj (dismissed):
It looks like there are a number of test failures; fixing those would be a prerequisite for testing this PR on Stripe's codebase.  You can edit the relevant `.rb` files to indicate the proper error messages and run `./tools/scripts/format_cxx.sh` to make the necessary changes for fixing the linting errors.

karmakaze:
While fixing up examples used in tests to match how Hash literals are now interpreted, I realized that it may be better to infer `T::Boolean` (for the underlying) even if the only value encountered was only `true` (or only `false`) as this is more commonly the intention. I got this much working, but is failing to produce `T.nilable(T::Boolean)` when the encountered values include all of `true`, `false`, and `nil`. WIP to continue.

rex-remind101:
hello, what's left to get this pr over the line? is this solution still relevant? thanks!

karmakaze:
> hello, what's left to get this pr over the line? is this solution still relevant? thanks!

This was based on PR https://github.com/sorbet/sorbet/pull/3823 which was closed due to large number of errors that couldn't be easily silenced in the Stripe codebase and only they could fix. The issue is work/valid and could be re-opened later if the errors can be managed. See the [comment on it]().

This PR even if done independently of #3823 would likely also generate many errors to a large existing codebase so can't really ship either until some way of managing new errors is done, perhaps that should be opened as a separate issue (of managing/silencing new errors), since it's blocking these two PRs.

PR #4499, opened 2021-08-15
ensure rb_execution_context_t is volatile-y loaded across return-from-block regions

froydnj:
### Motivation

This PR splits out the return-from-block changes from #4493, since we still need these changes for return-from-block even if we are going to change the strategy for handling exceptions.

Accordingly, the motivation is the same: we want to make sure that any variables we access that might cross a `RUBY_SETJMP` are guaranteed to live in memory so that we don't have to depend on the contents of particular registers.

The patches are generally lifted from #4493:

* everything up to 21845b0 is actually cherry-picked directly from that PR
* 7d68f06, 5e1b6ae, 0d911b1 fixup some issues with cherry-picking and also import the `setjmp` undefined behavior fix
* 916ac5c fixes up `sorbet_throwReturn` to have an explicit `ec` passed in
* f033639 addresses [some review feedback](https://github.com/sorbet/sorbet/pull/4493#discussion_r688780377)
* 691273f changes a comment to be more general than just "exception handling region", since `sorbet_initializeTag` is used for other things.

### Test plan

See included automated tests.


aprocter-stripe (approved):
LGTM! Nothing new to say relative to #4493. :)

PR #4501, opened 2021-08-16
Update abseil to avoid patching

elliottt:
8910297baf87e1777c4fd30fb0693eecf9f2c134 is the first abseil release that starts including an explicit dependency on the `@platforms` repo. Rolling back before that commit will require patching abseil before building. We did this previously in #4486.

### Motivation

Abseil now includes an explicit dependency on bazelbuild/platforms, so we no longer need to patch it to build with `--incompatible_use_platforms_repo_for_constraints`.

### Test plan

See included automated tests.


third_party/externals.bzl
```diff
@@ -153,12 +153,14 @@ def register_sorbet_dependencies():
         strip_prefix = "xxHash-0.8.0",
     )
 
+    # NOTE: 8910297baf87e1777c4fd30fb0693eecf9f2c134 is the first revision that starts including an explicit dependency
+    # on the `@platforms` repo. Rolling back before that commit will require patching abseil before building. We did
+    # this previously in https://github.com/sorbet/sorbet/pull/4486.
```
jez, line 158: Do you think it's better to just leave this in the git log? Seems like the kind of comment that's out of date immediately after writing.

Up to you.

jez (approved):

third_party/externals.bzl
elliottt: My worry was if we had to roll-back an abseil change for some reason, and picked something before the commit mentioned. Given that it was a bit painful to figure out why it failed when using that flag in the .bazelrc, I thought it might be kind to document it. I'd be happy to move this comment to the PR or the git commit though, perhaps that's a better place for it to live.

jez: > had to roll-back an abseil change for some reason

makes sense. it's hard for me to see us hand-rolling back a change here instead of reverting the commit to get a known good abseil version (but again: up to you)

PR #4504, opened 2021-08-17
Refactor packager to use prefix map for looking up packages

aadi-stripe:
Replaces the primary data structure of `packager` to be a map of path prefix -> package rather than a sorted vector of packages. Lookups now involve finding package in a map rather than iterating over the sorted vector.

### Motivation

We want to allow packages to have multiple file paths (in addition to the default one where the `__package.rb` file lives). This change will make the multiple file paths implementation easier, as you could then map from multiple path prefixes to the same package.

### Test plan

Tested with existing packager tests + added a new packager test which creates a file in a subfolder and a depth-2 nested package.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -139,8 +139,7 @@ public:
         auto curPrefixPos = path.find_last_of('/');
 
         while (curPrefixPos != std::string::npos) {
-            path = path.substr(0, curPrefixPos + 1);
-            const auto &it = packageInfoByPathPrefix.find(path);
+            const auto &it = packageInfoByPathPrefix.find(path.substr(0, curPrefixPos + 1));
 
             if (it != packageInfoByPathPrefix.end()) {
                 return it->second.get();
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -107,7 +107,7 @@ public:
      * Given a file of type PACKAGE, return its PackageInfo or nullptr if one does not exist.
      */
     const PackageInfo *getPackageByFile(core::Context ctx, core::FileRef packageFile) const {
-        const auto path = packageFile.data(ctx).path();
+        const std::string_view path = packageFile.data(ctx).path();
         const auto &it = packageInfoByPathPrefix.find(path.substr(0, path.find_last_of('/') + 1));
         if (it == packageInfoByPathPrefix.end()) {
             return nullptr;
@@ -135,8 +135,8 @@ public:
             Exception::raise("Cannot map files to packages until all packages are added and PackageDB is finalized");
         }
 
-        auto path = ctx.file.data(ctx).path();
-        auto curPrefixPos = path.find_last_of('/');
+        std::string_view path = ctx.file.data(ctx).path();
+        int curPrefixPos = path.find_last_of('/');
 
         while (curPrefixPos != std::string::npos) {
             const auto &it = packageInfoByPathPrefix.find(path.substr(0, curPrefixPos + 1));
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -67,7 +67,7 @@ class PackageDB final {
 private:
     // The only thread that is allowed write access to this class.
     const std::thread::id owner;
-    vector<shared_ptr<const PackageInfo>> packages;
+    UnorderedMap<std::string, shared_ptr<const PackageInfo>> packageInfoByPathPrefix;
     bool finalized = false;
     UnorderedMap<core::FileRef, shared_ptr<const PackageInfo>> packageInfoByFile;
     UnorderedMap<core::NameRef, shared_ptr<const PackageInfo>> packageInfoByMangledName;
@@ -97,16 +97,11 @@ public:
         }
 
         packageInfoByFile[ctx.file] = pkg;
-        packages.emplace_back(pkg);
+        packageInfoByPathPrefix[pkg->packagePathPrefix] = pkg;
     }
 
     void finalizePackages() {
         ENFORCE(owner == this_thread::get_id());
-        // Sort packages so that packages with the longest/most specific paths are first.
-        // That way, the first package to match a file's path is the most specific package match.
-        fast_sort(packages, [](const auto &a, const auto &b) -> bool {
-            return a->packagePathPrefix.size() > b->packagePathPrefix.size();
-        });
         finalized = true;
     }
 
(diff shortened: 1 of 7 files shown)


jez:
Can you include some measurements, maybe showing how the runtime and max memory usage changes before and after this?

packager/packager.cc
```diff
@@ -67,7 +67,7 @@ class PackageDB final {
 private:
     // The only thread that is allowed write access to this class.
     const std::thread::id owner;
-    vector<shared_ptr<const PackageInfo>> packages;
+    UnorderedMap<std::string, shared_ptr<const PackageInfo>> packageInfoByPathPrefix;
```
jvilk-stripe, line 70: `GlobalState` contains a map from string to file. Can we use that instead?

https://github.com/sorbet/sorbet/blob/e08172cf/core/GlobalState.h#L151

* Make this a map from `FileRef` to `shared_ptr<const PackageInfo>`.
* Run `gs.findFileByPath` on `$prefix/__packager.rb`
* If the return value `.exists()`, look up the corresponding `PackageInfo` in the `packageInfoByFile` map.

That will avoid redundantly storing these strings another time and will avoid needing this map.

Also, since the string is continually getting shorter during the lookup loop _and_ since `std::string` is mutable, you can perform lookup with only one string allocation:

 * Allocate the string for the file's folder + `__packager`.
 * Perform lookup with `findFileByPath`.
 * If nothing found, then call [resize](https://www.cplusplus.com/reference/string/string/resize/) up to the next `/` and then [append](https://www.cplusplus.com/reference/string/string/append/) with the `__packager.rb` constant. That should avoid re-allocating the string and will write to the same memory buffer.
   * Or, better yet, [replace](https://www.cplusplus.com/reference/string/string/replace/) at the next `/` and _then_ [resize](https://www.cplusplus.com/reference/string/string/resize/) to shrink the string to the last character of `__packager.rb`. In many cases I bet resize won't even be needed.

froydnj (changes requested):
What @jez said.

jvilk-stripe (changes requested):
I think we can reuse the file map in GlobalState.

jvilk-stripe:
Also, what @jez said!

packager/packager.cc
aadi-stripe: @jvilk-stripe I can look into the `resize` suggestion. With regards to using `findFileByPath` -- the purpose of making this change is to allow us to add more custom paths to a package, some of which won't have a `__package.rb` in them. So I'm not sure if that will work for us.

aadi-stripe:
@jez @jvilk @froydnj -- UPDATE
Afaik we have not tested the packager on all of the Stripe codebase. But, we have a small sandbox repo containing 22 source files and 14 packages (a total of 36 files).

Runtime:
Both before/after, Sorbet runs in ~5.9s on the sandbox repo.

Memory:
I measured memory by looking at the increase in max RSS from before the packager runs to just after it runs. Before this change, I see an increase of ~335KB for the sandbox repo, whereas after the change, I see an increase of ~360KB. This leaves us with a before/after delta of ~700B/file, wherein I am assuming that the memory usage is roughly proportional to the total number of files (in this case, 36).

If we were to extrapolate this data to the entire Stripe codebase, we would say that the total packager memory usage increase from this change could total around ~84MB.

However, one interesting fact is that if I look at the delta RSS right at the end of Step 1 in the packager, i.e. after the PackageDB is built, there is no change from before and after at that point. Both times, the increase in max RSS is around 127KB. The divergence happens after the other 2 steps, which add around 206KB and 233KB respectively. So the total change in the full Stripe codebase could be lower than 84MB; not sure. It's definitely upper-bounded at 84MB, though.

jez:
> Both before/after, Sorbet runs in ~5.9s on the sandbox repo.

Wow really? Isn't this **incredibly** slow? Stripe's codebase is like 10,000x bigger but takes 6x longer to type check.

aadi-stripe:
@jez Without the packager, we run at 4s on the sandbox repo. So without the packager, we increase runtime by 9x when we essentially 3000x the codebase size.

I know there's a lot of inherent parallelism in Sorbet -- and I don't think the packager is defeating that? Thinking about it in terms of Amdahl's law I don't know if we've increased the serial overhead by that much...

But I haven't measured this on all of Stripe's codebase.

nroman-stripe:
Was it an optimized build?

aadi-stripe:
> Was it an optimized build?

It wasn't, actually. I ran again with a release build. Sure enough, it's a lot faster overall. Also I do see a slight slowdown in the after case.

Before: 0.3s
After: 0.4s

packager/packager.cc
aadi-stripe: As per out-of-band convos, going to stick with `substr` as `path` is a `string_view` object.

aadi-stripe:
@nroman-stripe @jvilk-stripe @jez @froydnj 

Got RSS down to same as before (330KB) on the sandbox repo by removing unnecessary `packageInfoByFile`. As `packageInfoByPathPrefix` can be used for the file lookup.

aadi-stripe:
> @nroman-stripe @jvilk-stripe @jez @froydnj
> 
> Got delta-RSS down to same as/slightly less than before (330KB) on the sandbox repo by removing unnecessary `packageInfoByFile`. As `packageInfoByPathPrefix` can be used for the file lookup.

In terms of further performance optimizations -- since we are not running in packager mode on all of Stripe's codebase yet, and this change should not impact Sorbet globally, we feel that it's safe to merge at the moment. We definitely will instrument the packager on Stripe's codebase soon, at which point we will determine where exactly the performance improvements (if any) are needed.

jvilk-stripe (approved):
I'm OK with these changes since packager is not enabled on pay-server at the moment. Before turning it on, we should examine its runtime and change to max RSS.

jvilk-stripe:
>  if any, are needed (like for example converting package keys to NameRefs to prevent repeated string allocation, etc)

@aadi-stripe if you did this, then a prefix tree / trie may be a good choice for a data structure here.

jvilk-stripe:
> we will determine where exactly the performance improvements, if any, are needed 

I wonder if a trie could help at all. Explode a path into its components, make each a `NameRef`, and then have a trie using those `NameRef` IDs.

C++ doesn't provide the data structure, but it wouldn't be hard to build. Each Node could contain a sorted vector of children.

aadi-stripe:
> > we will determine where exactly the performance improvements, if any, are needed
> 
> I wonder if a trie could help at all. Explode a path into its components, make each a `NameRef`, and then have a trie using those `NameRef` IDs.
> 
> C++ doesn't provide the data structure, but it wouldn't be hard to build. Each Node could contain a sorted vector of children.

Agreed, it's worth a try (trie?) down the road 🥁 

@jez Let me know if there's any further feedback beyond the performance stuff, or if you are ok to approve.

froydnj (approved):
If jvilk is happy, then I am happy.

aadi-stripe:
mergebot_submit

PR #4505, opened 2021-08-17
Add .checked(:compiled)

neilparikh:
Adds a new `checked` option, which specifies that it will only be checked when the code is compiled.

### Motivation

The compiler always checks sigs at runtime, so this provides a way to specify that sigs should not be checked in both tests and production, while accurately reflecting that the sigs will be checked when compiled (which might in fact include at test time, if the Sorbet Compiler is being used to run the tests).

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -112,7 +112,7 @@ module T::Configuration
   # statically, so that methods don't have to guard themselves from being
   # called incorrectly by untyped code.
   #
-  # @param [:never, :tests, :always] default_checked_level
+  # @param [:compiled, :never, :tests, :always] default_checked_level
   def self.default_checked_level=(default_checked_level)
     T::Private::RuntimeLevels.default_checked_level = default_checked_level
   end
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -112,7 +112,7 @@ module T::Configuration
   # statically, so that methods don't have to guard themselves from being
   # called incorrectly by untyped code.
   #
-  # @param [:compiled, :never, :tests, :always] default_checked_level
+  # @param [:never, :compiled, :tests, :always] default_checked_level
   def self.default_checked_level=(default_checked_level)
     T::Private::RuntimeLevels.default_checked_level = default_checked_level
   end
--- a/gems/sorbet-runtime/lib/types/private/runtime_levels.rb
+++ b/gems/sorbet-runtime/lib/types/private/runtime_levels.rb
@@ -12,6 +12,8 @@ module T::Private::RuntimeLevels
     # Don't even validate in tests, b/c too expensive,
     # or b/c we fully trust the static typing
     :never,
+    # Only validate when compiled. Equivalent to :never in sorbet-runtime
+    :compiled,
   ].freeze
 
   @check_tests = false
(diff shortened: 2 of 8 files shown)


gems/sorbet-runtime/lib/types/configuration.rb
```diff
@@ -112,7 +112,7 @@ def self.disable_vm_prop_serde
   # statically, so that methods don't have to guard themselves from being
   # called incorrectly by untyped code.
   #
-  # @param [:never, :tests, :always] default_checked_level
+  # @param [:compiled, :never, :tests, :always] default_checked_level
```
jez, line 115:
```suggestion
  # @param [:never, :compiled, :tests, :always] default_checked_level
```

gems/sorbet-runtime/lib/types/private/runtime_levels.rb
```diff
@@ -12,6 +12,8 @@ module T::Private::RuntimeLevels
     # Don't even validate in tests, b/c too expensive,
     # or b/c we fully trust the static typing
     :never,
+    # Only validate when compiled. Equivalent to :never in sorbet-runtime
```
jez, line 15:
```suggestion
    # Validate the sig when the file is using the Sorbet Compiler.
    # Behaves like :never when interpreted.
```

jez:
https://github.com/sorbet/sorbet/blob/4ce2239697a5c6ce9985fa4af33f4fe9cdf4672f/rewriter/AttrReader.cc#L164

^ do you mind adding another comment here like

```cpp
 // Also treats `.checked(:compiled)` as checked, therefore not unchecked.
```

?

Also, would you mind writing a compiler test for this case?

There's some further notes about this in the `Complications` section of the Stripe-internal design doc.

jez:
https://github.com/sorbet/sorbet/blob/4ce2239697a5c6ce9985fa4af33f4fe9cdf4672f/gems/sorbet-runtime/lib/types/private/methods/signature_validation.rb#L173-L179

^ one thing i was curious to see here was whether we could write for whether overrides are validated with `.checked(:compiled)`. I **think** that we should be able to have enough of the `Signature` data structures populated that we should actually be able to do the signature override checking here **even when** the code is running in the interpreter (despite the `.checked(:compiled)` behaving like `.checked(:never)`) **as well as when** the code is running in the compiler. 

I think you should be able to set up a compiler test that calls like

```ruby
# compiled: true
class Parent
  # ...
end
class Child < Parent
  # ...
end

T::Utils.run_all_sig_blocks
T::Private::Abstract::Validate.validate_subclass(Parent)
T::Private::Abstract::Validate.validate_subclass(Child)
```

or something and makes sure that if there's an override error on a child or parent sig that was marked `.checked(:compiled)` that even the compiler raises the error. (You could write a sorbet-runtime test that does the same thing if you wanted, but it would be redundant with the compiler test.)

jez (approved):
Changes look good to me, and existing changes look great.

A couple suggestions for a test we could write to verify a non-obvious interaction with `.checked(:compiled)` and runtime signature validation.

jez:
If you'd like (up to you) feel free to add a note to the Sorbet docs about this:

https://sorbet.org/docs/runtime#checked-whether-to-check-in-the-first-place

or alternatively, document it somewhere in the compiler's README or `docs/` folder.

test/testdata/compiler/checked_compiled_override.rb
```diff
@@ -0,0 +1,26 @@
...
+  sig { returns(String).checked(:compiled) }
+  def foo
+    "a"
+  end
+end
+
+begin
+  T::Utils.run_all_sig_blocks
+rescue => e
+  puts e.message.include?("Incompatible return type in signature for override of method `foo`")
```
jez, line 25: so glad this works!

cc @froydnj looks like everything in the C API for registering signatures with `sorbet-runtime` is working as expected 🎉

jez (approved):

PR #4502, opened 2021-08-16
[wip] rewrite exception handling in C

froydnj:
### Motivation

Mostly, I need a PR for automatic testing on Stripe's codebase.  If that works out, this will probably get force-pushed on #4493's branch.

### Test plan

See included automated tests.

PR #4508, opened 2021-08-18
make ordering of break instructions more closely reflect runtime behavior

froydnj:
### Motivation

For a Ruby snippet like:

```ruby
  v = wrap do |value|
    if value == :in_raise
      p "inside block #{$!}"
      break :early_return
    end
    :late_return
  end
```

we'll generate CFG IR for the `if` block that looks roughly like:

```
    <cfgAlias>$27: T.class_of(<Magic>) = alias <C <Magic>>
    <statTemp>$28: String("inside block ") = "inside block "
    <statTemp>$25: String = <cfgAlias>$27: T.class_of(<Magic>).<string-interpolate>(<statTemp>$28: String("inside block "), $!$30: T.untyped)
    <statTemp>$23: NilClass = <self>: T.class_of(<root>).p(<statTemp>$25: String)
    <returnTemp>$31: Symbol(:early_return) = :early_return
    <block-break-assign>$32: Symbol(:early_return) = <returnTemp>$31
    v: Symbol(:early_return) = <block-break-assign>$32
    <magic>$33: T.class_of(<Magic>) = alias <C <Magic>>
    <block-break>$34: T.untyped = <magic>$33: T.class_of(<Magic>).<block-break>(<returnTemp>$31: Symbol(:early_return))
 ```

This IR is a bit gross, as `builder_walk.cc` documents:

https://github.com/sorbet/sorbet/blob/86b4a602488825444682836e4afaddc095963362/cfg/builder/builder_walk.cc#L563-L574

That is, what the above IR says is that we assign to the outermost variable `v` before calling `<Magic>.<block-break>`.  This is fine for static analysis purposes (mostly), but is not OK in the compiler.  If we widen the above example to:

```ruby
def wrap(&blk)
  begin
    p "starting begin"
    raise 'exception'
  rescue
    p "running raise #{$!}"
    yield :in_raise
  ensure
    p "running ensure #{$!}"
    raise 'from ensure'
  end
end

v = :unmodified

begin
  v = wrap do |value|
    if value == :in_raise
      p "inside block #{$!}"
      break :early_return
    end
    :late_return
  end
  p "got after v"
rescue Exception => e
...
```

`v` is not changed as the result of executing `wrap`, because `wrap` never actually returns a value.  But the above IR says that we should assign to `v` *before* breaking...except that the results of the break are not guaranteed to actually be returned, because something could preempt them -- a `raise` from an `ensure`, for instance, or a `return` from the same.

This PR continues the spirit of hacks around `break` by reordering the statements generated for a break: we generate the same statements, but `<Magic>.<block-break>` now comes before the assignment to the outermost variable.  We will still generate some dead code for the assignment that never happens (and, sadly, the variable still escapes because it's referenced in multiple blocks), but at least the runtime non-local exit `break` happens before the assignment.

### Test plan

I do not have compiler tests for this because such tests would depend on features -- `ensure` working in the presence of non-`Exception` non-local control flow -- that are currently broken in the compiler (and I am working on fixing, cf #4493, #4502 ).  I will add those tests when one of those PRs lands.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -574,7 +574,6 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                 // This is a temporary hack until we change how pining works to handle this case.
                 auto blockBreakAssign = cctx.newTemporary(core::Names::blockBreakAssign());
                 afterBreak->exprs.emplace_back(blockBreakAssign, a.loc, make_unique<Ident>(exprSym));
-                afterBreak->exprs.emplace_back(cctx.blockBreakTarget, a.loc, make_unique<Ident>(blockBreakAssign));
 
                 // call intrinsic for break
                 auto magic = cctx.newTemporary(core::Names::magic());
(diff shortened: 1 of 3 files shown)


jez:
> I do not have compiler tests for this because such tests would depend on features -- `ensure` working in the presence of non-`Exception` non-local control flow -- that are currently broken in the compiler (and I am working on fixing, cf #4493, #4502 ). I will add those tests when one of those PRs lands.

Are the changes in this PR to the cfg exp files sufficient to demonstrate the desired behavior? Would you not like to include one or two new CFG exp tests that include the snippets you mentioned in the PR description?

jez (approved):
lgtm, with one question about whether we want to add another test or two

froydnj:
> > I do not have compiler tests for this because such tests would depend on features -- `ensure` working in the presence of non-`Exception` non-local control flow -- that are currently broken in the compiler (and I am working on fixing, cf #4493, #4502 ). I will add those tests when one of those PRs lands.
> 
> Are the changes in this PR to the cfg exp files sufficient to demonstrate the desired behavior? Would you not like to include one or two new CFG exp tests that include the snippets you mentioned in the PR description?

The changes to the exp files are sufficient to show the effect of the change, and all of our other tests are sufficient to show that reordering doesn't change behavior here.

Extra CFG exp tests wouldn't demonstrate anything beyond the cfg exp tests here, since they would just show the same local transformation as a result of this patch.

froydnj:

→ https://go/builds/bui_K43XBKZa7fIOvU
→ https://go/builds/bui_K43XeOm7fHtI77

PR #4513, opened 2021-08-20
Add support for multiple directory paths for a package

aadi-stripe:
Adds a command line flag called `--extra-package-files-directory-prefix` which can have multiple instances and does the following.

Say we have packages `Project::Foo` and `Project::Bar`. For every `--extra-package-files-directory-prefix=extra_dir/` value, the packager will look for package files associated with `Project::Foo` in `extra_dir/Project_Foo/` and those associated with `Project::Bar` in `extra_dir/Project_Bar/`.

### Motivation

In Stripe's codebase, package source files aren't just located in the default source tree, but also in other directories (where they get name-spaced by the package name munged together with underscores).

### Test plan

Tested on a sandbox repository similar to Stripe's codebase. Also see included automated tests.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -848,7 +848,8 @@ void readOptions(Options &opts,
             }
             for (string dirName : raw["extra-package-files-directory-prefix"].as<vector<string>>()) {
                 if (dirName.back() != '/') {
-                    logger->error("--extra-package-files-directory-prefix directory path must have slash (/) at the end");
+                    logger->error(
+                        "--extra-package-files-directory-prefix directory path must have slash (/) at the end");
                     throw EarlyReturnWithCode(1);
                 }
                 opts.extraPackageFilesDirectoryPrefixes.emplace_back(dirName);
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -846,7 +846,7 @@ void readOptions(Options &opts,
                 logger->error("--extra-package-files-directory-prefix can only be specified in --stripe-packages mode");
                 throw EarlyReturnWithCode(1);
             }
-            for (string dirName : raw["extra-package-files-directory-prefix"].as<vector<string>>()) {
+            for (const string &dirName : raw["extra-package-files-directory-prefix"].as<vector<string>>()) {
                 if (dirName.back() != '/') {
                     logger->error(
                         "--extra-package-files-directory-prefix directory path must have slash (/) at the end");
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -342,7 +342,7 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
     options.add_options("advanced")("stripe-packages", "Enable support for Stripe's internal Ruby package system",
                                     cxxopts::value<bool>());
     options.add_options("dev")("extra-package-files-directory-prefix",
-                               "Extra parent directories which contain package files"
+                               "Extra parent directories which contain package files. "
                                "This option must be used in conjunction with --stripe-packages",
                                cxxopts::value<vector<string>>(), "string");
 
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -341,6 +341,10 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
     options.add_options("advanced")("stripe-mode", "Enable Stripe specific error enforcement", cxxopts::value<bool>());
     options.add_options("advanced")("stripe-packages", "Enable support for Stripe's internal Ruby package system",
                                     cxxopts::value<bool>());
+    options.add_options("dev")("extra-package-files-directory-prefix",
+                               "Extra parent directories which contain package files"
+                               "This option must be used in conjunction with --stripe-packages",
+                               cxxopts::value<vector<string>>(), "string");
 
     options.add_options("advanced")(
         "autogen-autoloader-exclude-require",
@@ -837,6 +841,20 @@ void readOptions(Options &opts,
         }
         opts.stripeMode = raw["stripe-mode"].as<bool>();
         opts.stripePackages = raw["stripe-packages"].as<bool>();
+        if (raw.count("extra-package-files-directory-prefix")) {
+            if (!opts.stripePackages) {
+                logger->error("--extra-package-files-directory-prefix can only be specified in --stripe-packages mode");
+                throw EarlyReturnWithCode(1);
+            }
+            for (string dirName : raw["extra-package-files-directory-prefix"].as<vector<string>>()) {
+                if (dirName.back() != '/') {
+                    logger->error("--extra-package-files-directory-prefix directory path must have slash (/) at the end");
+                    throw EarlyReturnWithCode(1);
+                }
+                opts.extraPackageFilesDirectoryPrefixes.emplace_back(dirName);
+            }
+        }
+
         extractAutoloaderConfig(raw, opts, logger);
         opts.errorUrlBase = raw["error-url-base"].as<string>();
         opts.noErrorSections = raw["no-error-sections"].as<bool>();
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -592,7 +592,7 @@ struct PackageInfoFinder {
 };
 
 // TODO (aadi-stripe) we can avoid syscalls if we invent an efficient way of looking up
-// directories in the source tree. Might be tied to https://github.com/sorbet/sorbet/issues/4509
+// directories in the source tree via GlobalState. Might be tied to https://github.com/sorbet/sorbet/issues/4509
 bool pathExists(const std::string &path) {
     struct stat buffer;
     return (stat(path.c_str(), &buffer) == 0);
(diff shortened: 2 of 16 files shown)


test/testdata/packager/extra_package_paths/bar/__package.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# typed: strict
+# enable-packager: true
+# extra-package-files-directory-prefix: test/testdata/packager/extra_package_paths/extra/
```
aadi-stripe, line 5: A special string property to test for this feature.

packager/packager.cc
```diff
@@ -837,7 +844,8 @@ bool checkContainsAllPackages(const core::GlobalState &gs, const vector<ast::Par
 
 } // namespace
 
-vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers, vector<ast::ParsedFile> files) {
+vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers, vector<ast::ParsedFile> files,
+                                      vector<std::string> extraPackageFilesDirectoryPrefixes) {
```
froydnj, line 848: Instead of copying `extraPackageFilesDirectoryPrefixes` on a call to this function, can we pass it as `const &`?

packager/packager.cc
```diff
@@ -591,7 +592,8 @@ struct PackageInfoFinder {
 
 // Sanity checks package files, mutates arguments to export / export_methods to point to item in namespace,
 // builds up the expression injected into packages that import the package, and codegens the <PackagedMethods>  module.
-unique_ptr<PackageInfo> getPackageInfo(core::MutableContext ctx, ast::ParsedFile &package) {
+unique_ptr<PackageInfo> getPackageInfo(core::MutableContext ctx, ast::ParsedFile &package,
+                                       vector<std::string> extraPackageFilesDirectoryPrefixes) {
```
froydnj, line 596: Instead of copying `extraPackageFilesDirectoryPrefixes` on a call to this function, can we pass it as `const &`?

packager/packager.cc
```diff
@@ -603,7 +605,12 @@ unique_ptr<PackageInfo> getPackageInfo(core::MutableContext ctx, ast::ParsedFile
     package.tree = ast::TreeMap::apply(ctx, finder, move(package.tree));
     finder.finalize(ctx);
     if (finder.info) {
-        finder.info->packagePathPrefix = packageFilePath.substr(0, packageFilePath.find_last_of('/') + 1);
+        finder.info->packagePathPrefixes.emplace_back(packageFilePath.substr(0, packageFilePath.find_last_of('/') + 1));
+        for (string prefix : extraPackageFilesDirectoryPrefixes) {
```
froydnj, line 609: This can be `for (const string &prefix : extraPackageFilesDirectoryPrefixes) {` to avoid copies.

packager/packager.cc
```diff
@@ -95,7 +95,9 @@ class PackageDB final {
             packageInfoByMangledName[pkg->name.mangledName] = pkg;
         }
 
-        packageInfoByPathPrefix[pkg->packagePathPrefix] = pkg;
+        for (std::string packagePathPrefix : pkg->packagePathPrefixes) {
```
froydnj, line 98: Please make this `for (const std::string &packagePathPrefix : ...` so at least we're only copying the string once to make the hash key rather than twice.

main/options/options.cc
```diff
@@ -837,6 +841,21 @@ void readOptions(Options &opts,
         }
         opts.stripeMode = raw["stripe-mode"].as<bool>();
         opts.stripePackages = raw["stripe-packages"].as<bool>();
+        if (raw.count("extra-package-files-directory-prefix")) {
+            if (!opts.stripePackages) {
+                logger->error("--extra-package-files-directory-prefix can only be specified in --stripe-packages mode");
+                throw EarlyReturnWithCode(1);
+            }
+            for (string dirName : raw["extra-package-files-directory-prefix"].as<vector<string>>()) {
```
froydnj, line 849: Let's make this `for (const string &dirName : ...` instead.

packager/packager.cc
```diff
@@ -603,7 +605,12 @@ unique_ptr<PackageInfo> getPackageInfo(core::MutableContext ctx, ast::ParsedFile
     package.tree = ast::TreeMap::apply(ctx, finder, move(package.tree));
     finder.finalize(ctx);
     if (finder.info) {
-        finder.info->packagePathPrefix = packageFilePath.substr(0, packageFilePath.find_last_of('/') + 1);
+        finder.info->packagePathPrefixes.emplace_back(packageFilePath.substr(0, packageFilePath.find_last_of('/') + 1));
+        for (string prefix : extraPackageFilesDirectoryPrefixes) {
+            string_view shortName = finder.info->name.mangledName.shortName(ctx.state);
+            string additionalDirPath = absl::StrCat(prefix, shortName.substr(0, shortName.find("_Package")), "/");
```
froydnj, line 611: Is it worth `ENFORCE`'ing that this `find` call doesn't return `shortName.end()`?

packager/packager.cc
```diff
@@ -603,7 +605,12 @@ unique_ptr<PackageInfo> getPackageInfo(core::MutableContext ctx, ast::ParsedFile
...
+            finder.info->packagePathPrefixes.emplace_back(additionalDirPath);
```
froydnj, line 612: Let's `std::move(additionalDirPath)` here to avoid some unnecessary copying.

packager/packager.cc
```diff
@@ -930,14 +938,15 @@ vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers
     return files;
 }
 
-vector<ast::ParsedFile> Packager::runIncremental(core::GlobalState &gs, vector<ast::ParsedFile> files) {
+vector<ast::ParsedFile> Packager::runIncremental(core::GlobalState &gs, vector<ast::ParsedFile> files,
+                                                 vector<std::string> extraPackageFilesDirectoryPrefixes) {
```
froydnj, line 942: Instead of copying `extraPackageFilesDirectoryPrefixes` on a call to this function, can we pass it as `const &`?

packager/packager.cc
froydnj: Can we pull the computation of `shortName` and the `substr`'ing on it out of the loop manually?  That way we're not doing a string `find` every time through the loop.

froydnj (changes requested):
This all seems pretty reasonable to me, but I have a couple changes that would reduce copying of temporary objects.  Maybe this is not a huge concern because there are not going to be that many directory paths, but it looks like we process `O(# of files x # prefixes)` in some places, so faster would be better for scaling on large codebases.

packager/packager.cc
aadi-stripe: Aha, I'm still building my intuition transitioning from call-by-reference to call-by-value semantics :). Will make this a pass by reference.

packager/packager.cc
aadi-stripe: Ah yes, this can be hoisted, will do.

aadi-stripe: Also on the `shortName.end()` `ENFORCE`, I'm not sure if it's worth enforcing here, as we ourselves in the packager are constructing this `NameRef` earlier in the piece with a `_Package` at the end.

froydnj: > Also on the `shortName.end()` `ENFORCE`, I'm not sure if it's worth enforcing here

That seems totally fair. :+1:

aadi-stripe:
ptal @froydnj

main/options/options.cc
```diff
@@ -837,6 +841,21 @@ void readOptions(Options &opts,
         }
         opts.stripeMode = raw["stripe-mode"].as<bool>();
         opts.stripePackages = raw["stripe-packages"].as<bool>();
+        if (raw.count("extra-package-files-directory-prefix")) {
+            if (!opts.stripePackages) {
+                logger->error("--extra-package-files-directory-prefix can only be specified in --stripe-packages mode");
+                throw EarlyReturnWithCode(1);
+            }
+            for (const string &dirName : raw["extra-package-files-directory-prefix"].as<vector<string>>()) {
```
froydnj, line 849: WDYT about checking that these exist on disk as directories here?

packager/packager.cc
```diff
@@ -589,9 +591,17 @@ struct PackageInfoFinder {
     }
 };
 
+// TODO (aadi-stripe) we can avoid syscalls if we invent an efficient way of looking up
+// directories in the source tree via GlobalState. Might be tied to https://github.com/sorbet/sorbet/issues/4509
+bool pathExists(const std::string &path) {
+    struct stat buffer;
+    return (stat(path.c_str(), &buffer) == 0);
```
froydnj, line 598: Do you want to additionally check `S_ISDIR(buffer.st_mode)`?

froydnj (approved):
Thanks!  Two non-blocking comments below.

packager/packager.cc
aadi-stripe: I think it's ok here because the only usage of this method passes paths ending in `/`, so it's unlikely that those are files and not directories.

main/options/options.cc
aadi-stripe: That's a good idea -- we can add that at a later time :)

PR #4514, opened 2021-08-20
Propagate whether type tests are done to final method calls

neilparikh:
### Test plan

See included automated tests.


jez:
Compiler is gone 😢

PR #4511, opened 2021-08-19
Disable INFO logging in abseil for emscripten builds

elliottt:
### Motivation

Remove the abseil output from sorbet.run.

### Test plan

See included automated tests.

PR #4515, opened 2021-08-20
Try fixing absl INFO loggin in a different way

jez:
### Motivation

Tries fixing https://github.com/sorbet/sorbet/pull/4511 in a way that doesn't involve turning off all INFO logging.

### Test plan

I copied the `tar` file out of Buildkite and tested it in sorbet.run locally.
Steps looked like:

- Verify that sorbet.run master shows no error
- Verify that sorbet.run master^ shows the error that #4511 fixed
- Verify that untar'ing this PR's Buildkite artifact over master^ fixes the
  error again

I also tested that this builds on macOS by temporarily commenting out the `master` filter in the pipeline:

https://buildkite.com/sorbet/sorbet/builds/16994#85ca57cf-b08b-4a6e-b9e0-bce63f6e7235

### DIFF
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -4,7 +4,6 @@
 #define FULL_BUILD_ONLY(X) X;
 #include "core/proto/proto.h" // has to be included first as it violates our poisons
 // intentional comment to stop from reformatting
-#include "absl/debugging/symbolize.h"
 #include "common/statsd/statsd.h"
 #include "common/web_tracer_framework/tracing.h"
 #include "main/autogen/autogen.h"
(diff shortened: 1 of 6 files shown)


jez:
cc @elliottt

main/realmain.cc
```diff
@@ -358,7 +357,7 @@ void runAutogen(const core::GlobalState &gs, options::Options &opts, const autog
 
 int realmain(int argc, char *argv[]) {
 #ifndef SORBET_REALMAIN_MIN
-    absl::InitializeSymbolizer(argv[0]);
+    initializeSymbolizer(argv[0]);
```
aprocter-stripe, line 360: Is there some way we could namespace this (e.g. `sorbet::initializeSymbolizer`) or otherwise disambiguate it from `absl::InitializeSymbolizer`? Just thinking it might be nice to make it clear to folks who are reading this code and happen to be familiar with `absl::InitializeSymbolizer` that we're not directly calling the `absl` function here.

elliottt (approved):

aprocter-stripe (approved):
LGTM, and this seems like a more precise workaround! One extremely minor suggestion.

main/realmain.cc
jez: Yeah i thought about that, but i noticed that none of our os.cc functions do any namespacing. I'm not opposed to it if you want to take a stab at it, but it seems to work as is, so i'm inclined to not put further work into it.

PR #4517, opened 2021-08-20
move the implementation of sorbet_throwReturn into the VM

froydnj:
### Motivation

The current implementation of `sorbet_throwReturn` will transition into `libruby.so` twice.  Instead, why don't we just transition once and make the code smaller in the process?

### Test plan

See included automated tests.


jez (approved):
👍

PR #4518, opened 2021-08-20
Stop depending on system node for emscripten build

jez:
### Motivation

I don't have `node` installed system-wide, so I wasn't able to build the WASM
build of Sorbet with emscripten.

(I also don't have `python` system-wide, I only have `python3`, but that's a
battle for another day, because it seems harder to get a hermetic python
executable than to get a hermetic node executable.)

### Test plan

Ran this command on my personal desktop, which is what we also invoke in
Buildkite:

```
bazel build //emscripten:sorbet-wasm.tar --config=webasm-linux --strip=always
```

### DIFF
--- a/.buildkite/build-emscripten.sh
+++ b/.buildkite/build-emscripten.sh
@@ -4,7 +4,6 @@ set -euo pipefail
 
 echo "--- Pre-setup"
 
-command -v node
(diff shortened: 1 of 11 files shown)


froydnj (approved):
WFM!

PR #4519, opened 2021-08-21
Remove the single argument restriction from test_each

elliottt:
### Motivation

Extend the behavior of `test_each` allowing destructuring in the block argument list.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Minitest.cc
+++ b/rewriter/Minitest.cc
@@ -244,11 +244,11 @@ ast::ExpressionPtr runSingle(core::MutableContext ctx, bool isClass, ast::Send *
     }
 
     if ((send->fun == core::Names::testEach() || send->fun == core::Names::testEachHash()) && send->args.size() == 1) {
-        if ((send->fun == core::Names::testEach() && block->args.size() != 1) ||
+        if ((send->fun == core::Names::testEach() && block->args.size() < 1) ||
             (send->fun == core::Names::testEachHash() && block->args.size() != 2)) {
             if (auto e = ctx.beginError(send->block.loc(), core::errors::Rewriter::BadTestEach)) {
                 e.setHeader("Wrong number of parameters for `{}` block: expected `{}`, got `{}`", send->fun.show(ctx),
-                            1, block->args.size());
+                            send->fun == core::Names::testEach() ? "at least 1" : "2", block->args.size());
             }
             return nullptr;
         }
--- a/rewriter/Minitest.cc
+++ b/rewriter/Minitest.cc
@@ -178,7 +178,8 @@ ast::ExpressionPtr getIteratee(ast::ExpressionPtr &exp) {
 
 // this applies to each statement contained within a `test_each`: if it's an `it`-block, then convert it appropriately,
 // otherwise flag an error about it
-ast::ExpressionPtr runUnderEach(core::MutableContext ctx, core::NameRef eachName, ast::ExpressionPtr stmt,
+ast::ExpressionPtr runUnderEach(core::MutableContext ctx, core::NameRef eachName,
+                                const ast::InsSeq::STATS_store &destructuringStmts, ast::ExpressionPtr stmt,
                                 ast::MethodDef::ARGS_store &args, ast::ExpressionPtr &iteratee) {
     // this statement must be a send
     if (auto *send = ast::cast_tree<ast::Send>(stmt)) {
@@ -199,6 +200,16 @@ ast::ExpressionPtr runUnderEach(core::MutableContext ctx, core::NameRef eachName
             for (auto &arg : args) {
                 new_args.emplace_back(arg.deepCopy());
             }
+
+            // add the destructuring statements to the block if they're present
+            if (!destructuringStmts.empty()) {
+                ast::InsSeq::STATS_store stmts;
+                for (auto &stmt : destructuringStmts) {
+                    stmts.emplace_back(stmt.deepCopy());
+                }
+                body = ast::MK::InsSeq(body.loc(), std::move(stmts), std::move(body));
+            }
+
             auto blk = ast::MK::Block(send->loc, move(body), std::move(new_args));
             auto each = ast::MK::Send0Block(send->loc, iteratee.deepCopy(), core::Names::each(), move(blk));
             // put that into a method def named the appropriate thing
@@ -207,6 +218,7 @@ ast::ExpressionPtr runUnderEach(core::MutableContext ctx, core::NameRef eachName
             return constantMover.addConstantsToExpression(send->loc, move(method));
         }
     }
+
     // if any of the above tests were not satisfied, then mark this statement as being invalid here
     if (auto e = ctx.beginError(stmt.loc(), core::errors::Rewriter::BadTestEach)) {
         e.setHeader("Only valid `{}`-blocks can appear within `{}`", "it", eachName.show(ctx));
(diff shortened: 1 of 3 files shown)


test/testdata/rewriter/minitest_tables.rb
```diff
@@ -86,13 +86,48 @@ def self.it(name, &blk); end
   end
 
 
-  test_each [1, 2, 3] do |k, v| # error: Wrong number of parameters for `test_each` block
-    it "does not handle more than one argument" do
+  test_each [1, 2, 3] do |k, v|
```
jez, line 89: Can you add (or are there already) tests for when the first argument to `test_each` isn't a list syntactically?

```ruby
test_each(value) do |x, y|
end
```

elliottt: Are you looking for the combination of a value argument and block with multiple arguments? If not, there is this case:
https://github.com/sorbet/sorbet/blob/077ec005/test/testdata/rewriter/minitest_tables.rb#L51-L57

jez: > Are you looking for the combination of a value argument and block with multiple arguments?

yep, specifically multiple block arguments too

elliottt: I added multiple arg/destructuring variants of the existing tests in b54e51fd532a2b1eebb8a06984239da3d115d1d4

jez: awesome, thanks!

rewriter/Minitest.cc
```diff
@@ -215,18 +227,58 @@ ast::ExpressionPtr runUnderEach(core::MutableContext ctx, core::NameRef eachName
...
 // this just walks the body of a `test_each` and tries to transform every statement
 ast::ExpressionPtr prepareTestEachBody(core::MutableContext ctx, core::NameRef eachName, ast::ExpressionPtr body,
-                                       ast::MethodDef::ARGS_store &args, ast::ExpressionPtr &iteratee) {
-    auto *bodySeq = ast::cast_tree<ast::InsSeq>(body);
-    if (bodySeq) {
+                                       ast::MethodDef::ARGS_store &args, ast::InsSeq::STATS_store destructuringStmts,
+                                       ast::ExpressionPtr &iteratee) {
+    if (auto *bodySeq = ast::cast_tree<ast::InsSeq>(body)) {
+        if (isDestructuringInsSeq(ctx, args, bodySeq)) {
+            ENFORCE(destructuringStmts.empty(), "Nested destructuring statements");
```
jez, line 268: i haven't looked carefully but judging from the text of the enforce, is it actually possible for a user to encounter this enforce? should this be a `beginError` instead?

elliottt: The implication of this enforce firing would be that the destructuring desugaring managed to fire twice, but without anything that introduced new names in the middle. I couldn't figure out a case that would trigger this, but it felt like a reasonable defensive measure against changes to the desugar pass in the future.

rewriter/Minitest.cc
```diff
@@ -215,18 +227,58 @@ ast::ExpressionPtr runUnderEach(core::MutableContext ctx, core::NameRef eachName
...
+
+    return absl::c_find_if(args, [name](auto &argExpr) {
+               auto *arg = ast::cast_tree<ast::UnresolvedIdent>(argExpr);
+               return arg && arg->name == name;
+           }) != args.end();
+}
+
+// Destructuring blocks are added to method/block entries as an InsSeq where the method body is the distinguished final
+// statement, and the statements of the InsSeq are all individual InsSeq expressions that make up the destructuring for
+// a specific argument.
```
jez, line 249:
```suggestion
// When given code that looks like
//
//     test_each(pairs) do |(x, y)|
//       # ...
//     end
//
// Sorbet desugars it to essentially
//
//     test_each(pairs) do |<destructureArg$1|
//       x = <destructureArg$1>[0]
//       y = <destructureArg$1>[1]
//
//       # ...
//     end
//
// which would otherwise defeat the "Only valid it-blocks can appear within test_each" error message.
//
// Because this case is so common, we have special handling to detect "contains only valid it-blocks
// plus desugared destruturing assignments."
```

it took me longer than I would have liked to figure out why this change works. If this had been the comment here, I think it would have made sense much faster. (Also, maybe I'm just rusty)

test/testdata/rewriter/minitest_tables.rb
```diff
@@ -86,13 +86,93 @@ def self.it(name, &blk); end
...
+    end
+  end
+
+  local_tuple = [[1,'a'], [2,'b']]
+  test_each local_tuple do |i,s|
+    it "succeed with local variables but cannot type them" do
+      T.reveal_type(i) # error: type: `T.untyped`
+      T.reveal_type(s) # error: type: `T.untyped`
+    end
+  end
```
jez, line 165: Can you add a test for when the programmer does the destructuring manually?

```ruby

test_each([[1, 2], [3, 4]]) do |pair|
  x, y = pair
  it 'manual destructuring' do
    # ...
  end
end
```

(no preference whether this should work or not, just want to capture the behavior)

test/testdata/rewriter/minitest_tables.rb
```diff
@@ -86,13 +86,93 @@ def self.it(name, &blk); end
...
+    end
+
+    it "multiple it blocks are ok with multiple destructured arguments" do
+      T.reveal_type(k) # error: Revealed type: `Integer`
+      T.reveal_type(v) # error: Revealed type: `NilClass`
+    end
+  end
+
+  test_each [[1, ['hi', false]], [2, ['bye', true]]] do |i, (s,b), (x, y)|
+    it "handles mixed destructuring and positional arguments" do
```
johnp-stripe, line 114: Should we have at least one test case here with an interpolated `it` string? That is a key user flow in the wild (at least, it's how we found this). Something like:

```ruby
test_each([['red', 2], ['blue', 3]]) do |(color, rank)|
  it "color #{color} which is not purple is not number 1" do
    refute_equal(1, rank)
  end
end
```

jez (approved):

johnp-stripe (commented):
Thanks for taking this on! My team appreciates it!! 🚀 

I do have one comment about adding a representative test case. I'll let you decide if it is a valid case or not. Feel free to ignore if it won't add value.

rewriter/Minitest.cc
jez: when i left this comment i was thinking it was talking about situations like

```ruby
test_each(...) do |(x, y)|
  test_each(...) do |(a, b)|
  end
end
```

but after reading the code more carefully, i don't think that's what you meant by nested here.

elliottt: That's correct. The example that you wrote would have a send with a block between the destructuring statements added to the beginning of the outer block.

test/testdata/rewriter/minitest_tables.rb
elliottt: Happy to add the test, it's going to fail currently as we only allow `it` blocks within `test_each`.

test/testdata/rewriter/minitest_tables.rb
elliottt: This exposed a problem with the current implementation: when we make the test name we generate a string based on the AST of the string given to `it`, so your `it` block would have a name like:
```
def <it '::<Magic>.<string-interpolate>("color ", color, " which is not purple is not number 1)'><<todo method>>(&<blk>)
```
The problem with this is that any variables used for string interpolation here that don't support `.to_s` will not have an error raised. I added a test for this and will file a ticket for a follow-on fix.

elliottt: https://github.com/sorbet/sorbet/issues/4524

gdritter-stripe (approved):
This all looks reasonable to me!

johnp-stripe (approved):

elliottt:

→ https://go/builds/bui_K5z1CRRQUVxW8b
→ https://go/builds/bui_K5z1EqutmS7ZNg

PR #4520, opened 2021-08-21
move return-from-block machinery into the VM itself

froydnj:
### Motivation

Investigation into #4502 revealed that [GCC 9.3.0 on Ubuntu Focal sets `-fcf-protection` by default](https://github.com/ipxe/ipxe/commit/e8393c372).  This setting changes the values that get stored in the `jmpbuf` data structure (there's no actual typedef for this, it's just a array of 5 `void*` pointers) used by `__builtin_{set,long}jmp`.  Without that flag, GCC would store `{%rbp, %rip, %rsp}` into the `jmpbuf`.  With the flag, GCC stores `{%rbp, %rip, %ssp, %rsp}`, where `%ssp` is the shadow stack pointer used by [Intel's Control-flow Enforcement Technology](https://software.intel.com/content/www/us/en/develop/articles/technical-look-control-flow-enforcement-technology.html).

You may notice that the two different `jmpbuf` layouts are different depending on the value of `-fcf-protection`.  You might think that this would be a problem, and [you would be right](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98997).  What is less obvious, but still equally a problem, is that clang, while compatible with GCC's default layout of `jmpbuf` when `-fcf-protection` is turned off (for both compilers), is _incompatible_ with GCC's layout when `-fcf-protection` is turned on for GCC, _regardless_ of how you compile things with clang.  That is:

* GCC (no flag) + clang (no flag): interoperates
* GCC (no flag) + clang (flag): does not interoperate (listed only for completeness here)
* GCC (flag) + clang (no flag): does not interoperate
* GCC (flag) + clang (flag): does not interoperate

(clang will store `{%rbp, %rip, %rsp, %ssp}` with `-fcf-protection`; note that the positions of `%rsp` and `%ssp` differ between GCC and clang.  You can check out https://gcc.godbolt.org/z/89cvGYYM7 for a live demo of this.)

Which means that if you have a Ruby compiled with GCC using `-fcf-protection`, a clang-compiled caller that uses (clang's) `__builtin_setjmp` (note that it doesn't matter what compiler settings clang uses), and code that eventually wants to execute, inside Ruby, (gcc's) `__builtin_longjmp` on a buffer filled by clang's `__builtin_setjmp`, you're gonna have a bad time.

All of this means that, for the Sorbet compiler, all uses of `__builtin_setjmp` and `__builtin_longjmp` should be confined inside the Ruby VM itself so that they are consistently built with the same compiler and interoperability concerns go away.  Which implies that the return-from-block implementation needs to live inside the VM.  (There are alternatives involving gnarly uses of inline asm and bazel build flags, but we're opting not to go down that pathway right now.  Maybe later.)

That was the motivation behind e98b8c9: instead of recognizing that we have a `return X` statement inside a block and inserting some machinery to catch a thrown `TAG_RETURN` in the function itself, what we can do instead is codegen as normal (including the thrown `TAG_RETURN`).  But once the function is complete, we need to check whether there was a `return X` statement inside a block somewhere inside the function.  If so, we clone our codegenned function, clear out the original function, and make our original function call into a wrapper function in the VM.  This wrapper function receives the cloned function (which actually implements all the logic of the original Ruby method), and handles setting up the appropriate tag machinery to catch a `TAG_RETURN` destined for our frame before calling the cloned function.

That approach worked much better than expected, except for one wrinkle: we had a test that checked that returned values from blocks were typechecked with the return type of the method.  This checking had gone away with the approach above, because we had expected values to be typechecked at the point of catching the `TAG_RETURN` while still inside the function, instead of at the actual `return X` in the block.  Now there was no catch of the `TAG_RETURN` inside the function: the catch of the `TAG_RETURN` had been moved into the runtime, where there couldn't be a typecheck.

Discussion indicated that we could plausibly move the typecheck to the point of the `return X` (i.e. the point at which we called `sorbet_throwReturn`).  So, to fix one test, there are another ten commits or so on top of a single patch.  Classic.

To move that typecheck, we decided that we need to refactor some things.

de08864 through 548a5ba could plausibly have been split out into their own PR, but they make a lot more sense in the context of this PR, I think.  Part of the strategy for implementing return-from-block was:

* generate a boolean flag for every block, initialized to false
* for `return X` inside a block, we'd set that flag to true for that particular block before codegen'ing the actual "return this value"
* the machinery handling codegen'ing the actual "return this value" at the LLVM level would consult that (runtime!) variable to determine whether to call `sorbet_throwReturn` or emit an LLVM `ret` instruction.

This setup is more complicated than it needs to be, because we will only ever set the boolean flag to true when we're returning from a block, which we know we're doing statically.  So we might as well, for an explicit `return X` from a block, statically generate `sorbet_throwReturn` (d690194 and f6a9437), which means that we can delete all the boolean flag bits (548a5ba).

Once we have that, making sure that we do the typechecking at the point of return rather than at the point of catching `TAG_RETURN` is relatively simple: we need to split out a function to do the actual typechecking, if necessary, of the return value (8f26c75), move it to a place that it's visible from the function we're calling to generate return-from-block (b0dc064), and then actually call that function prior to throw (c14f65a).

I also belatedly realized that we no longer need a bunch of state in `IREmitterContext` for return-from-block: we can just have a simple boolean flag indicating that said construct is present (27eea8b).

### Test plan

See included automated tests.


froydnj:
Possible wrinkle:

```ruby
# frozen_string_literal: true
# typed: true
# compiled: true

extend T::Sig

class A
  extend T::Sig

  attr_reader :proc

  def initialize(p)
    @proc = p
  end
end

sig {returns A}
def a
  A.new(lambda {return T.unsafe(5)})
end

sig {params(a: A).returns(Integer)}
def b(a)
  a.proc.call
end

p b(a())
```

works before this change and fails after, because the return value of the lambda is type-checked against `A`, rather than waiting until its point of call to be type-checked as an `Integer`.  This doesn't seem like an unreasonable use-case. :(

Perhaps we could pass a small validation function into `sorbet_vm_return_from_block` wrapper for checking the type of any thrown values, and then just throw unchecked values.

froydnj:
> Perhaps we could pass a small validation function into `sorbet_vm_return_from_block` wrapper for checking the type of any thrown values, and then just throw unchecked values.

A better solution was implemented in 57d9d23: we can *know* on the way out of the function whether our return value is coming from return-from-block and typecheck the value there.  This is a minor amount of code bloat, but I don't expect that return from block is _super_ common.  I also added the testcase from the comment above and an explicit testcase for return-from-block with a function returning void.

froydnj:

→ https://go/builds/bui_K5qfxaOLAbywne
→ https://go/builds/bui_K5qfjgx0Gb8OGd

compiler/IREmitter/IREmitterContext.cc
```diff
@@ -805,8 +784,7 @@ IREmitterContext IREmitterContext::getSorbetBlocks2LLVMBlockMapping(CompilerStat
         move(blockExits),
         move(blockScopes),
         move(blockUsesBreak),
-        move(throwReturnFlagByBlock),
-        move(returnFromBlockState),
+        move(hasReturnFromBlock),
```
aprocter-stripe, line 787:
```suggestion
        hasReturnFromBlock,
```

compiler/IREmitter/IREmitterHelpers.cc
```diff
@@ -236,11 +221,26 @@ void IREmitterHelpers::emitReturn(CompilerState &cs, llvm::IRBuilderBase &build,
     }
 }
 
-void IREmitterHelpers::setThrowReturnFlag(CompilerState &cs, llvm::IRBuilderBase &build, const IREmitterContext &irctx,
-                                          int rubyBlockId) {
-    auto &builder = static_cast<llvm::IRBuilder<> &>(build);
+llvm::Value *IREmitterHelpers::maybeCheckReturnValue(CompilerState &cs, cfg::CFG &cfg, llvm::IRBuilderBase &build,
```
aprocter-stripe, line 224: The fact that this can return `voidSingleton` seems elegant to me, but it wasn't clear at the call site. Kinda wonder if that suggest that this function could use a better name, but can't think of anything so :shrug:

aprocter-stripe (approved):
This seems a lot cleaner!

At a high level the only concern I had as I was reading through it is how return from lambda winds up skipping its type check. Then I remembered it's this: https://github.com/sorbet/sorbet/blob/22b927d8797680106d2cdf9778a8bc2fd4c0a2b1/compiler/IREmitter/Payload/patches/vm_insnhelper.c#L262-L268 . So I think we're fine!

compiler/IREmitter/IREmitter.cc
```diff
@@ -824,6 +791,81 @@ void IREmitter::run(CompilerState &cs, cfg::CFG &cfg, const ast::MethodDef &md)
...
+    // Completely delete the function body.
+    func->dropAllReferences();
+
+    // Turn the original function into a trampoline for this new function.
+    auto *entryBlock = llvm::BasicBlock::Create(cs, "entry", func);
+    builder.SetInsertPoint(entryBlock);
+
+    auto *wrapper = cs.getFunction("sorbet_vm_return_from_block_wrapper");
+    // Adding the function argument at the end means that we don't have to perform
+    // any register shuffling.
```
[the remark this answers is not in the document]
froydnj: I meant that if you had the four initial arguments passed in registers A, B, C, D, and you, say, passed the function pointer argument first, you'd need to execute, pseudo-code-ly in intel asm syntax:

```
mov E, D
mov D, C
mov C, B
mov B, A,
mov A, function-pointer
```

which is a lot more code than:

```
mov E, function-pointer
```

if you pass the function last.  (Possibly not any slower, since the CPU just does magic register renaming internally, or something?)

compiler/IREmitter/IREmitter.cc
```diff
@@ -824,6 +791,81 @@ void IREmitter::run(CompilerState &cs, cfg::CFG &cfg, const ast::MethodDef &md)
...
+                                      {func->arg_begin(), func->arg_begin() + 1, func->arg_begin() + 2,
+                                       func->arg_begin() + 3, implementationFunction},
+                                      "returnedFromBlock");
+
+    // TODO(froydnj): LLVM IR is somewhat machine-specific when it comes to calling
+    // functions returning a structure, like sorbet_vm_return_from_block_wrapper.
+    // The conventions we have here are correct for x86-64 (Linux and Mac), but the
+    // LLVM IR generated by clang for returning the same structure on arm64 Linux
+    // returns a 2 x i64 vector.  If the structure was slightly bigger, the function
+    // would actually take a pointer to the structure as a separate argument.
```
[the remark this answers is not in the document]
froydnj: I believe you are responsible for setting up the appropriate arguments to `CreateCall`, so you'd have to take the responsibility for setting up the alloca, passing in the alloca pointer as the first argument to `CreateCall`, etc.  Which means that if you are generating LLVM IR yourself, then you need to be intimately aware of the argument-passing conventions of your particular target.  (Clang is intimately aware of said conventions and how each LLVM backend would like its arguments passed, which is one of the gnarliest parts of Clang.)  This is...gross, and LLVM upstream would like to fix it, but it's a big job.

compiler/IREmitter/IREmitterContext.cc
```diff
@@ -637,16 +618,14 @@ IREmitterContext IREmitterContext::getSorbetBlocks2LLVMBlockMapping(CompilerStat
...
-        throwReturnFlagByBlock.emplace_back(
-            builder.CreateAlloca(llvm::Type::getInt1Ty(cs), nullptr, "throwReturnFlag"));
-        builder.CreateStore(builder.getFalse(), throwReturnFlagByBlock[i]);
         if (i == 0) {
-            // If no return statements are actually present inside blocks, we will not need to push an EC tag.
+            // If return statements are present inside blocks, we need to codegen
+            // differently when finalizing the function.
             //
-            // TODO(aprocter): I think this is a little bit more conservative than it needs to be, because it will push
+            // TODO(aprocter): I think this is a little bit more aggressive than it needs to be, because it will push
```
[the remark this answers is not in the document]
froydnj: You know, I changed this thinking that pushing the typecheck to throwing the return value somehow made this go away, and then forgot to change it back when I realized I still had to typecheck things on the actual way out!  I will put this back the way it was, nice catch.

compiler/IREmitter/IREmitterHelpers.cc
froydnj: Yeah, I don't know what a better name for this function is.  `maybeTypecheckOrAdjustReturnValue`?  That's a mouthful.

froydnj (commented):
Thanks for the review!  I'll address comments and resolve any conflicts tomorrow morning.

compiler/IREmitter/IREmitterHelpers.cc
aprocter-stripe: Indeed. As it is I suppose most readers will probably notice at the call site that it's returning an `llvm::Value *`, wonder why, and chase this down anyway. (So maybe a comment would help here, but it's up to you.)

froydnj: I added a comment at the declaration.

PR #4526, opened 2021-08-24
enable control-flow protection unconditionally for compiled modules

froydnj:
### Motivation

This mechanism is a defense-in-depth switch for compiled code; on x86-ish machines, these annotations on the module instructs the code generator to emit instructions for [Intel's Control-flow Enforcement Technology](https://software.intel.com/content/www/us/en/develop/articles/technical-look-control-flow-enforcement-technology.html).  Adding these annotations is, as far as I can tell, a no-op on targets that don't have similar mechanisms (not that the Sorbet compiler supports anything but x86-64 at the moment anyway).

Code compiled with this option interoperates just fine with code compiled without this option.  My system libraries are not compiled with `-fcf-protection` (the corresponding option for C/C++ compilers), nor does my computer have a processor that is CET-capable.  But I can run tests with this PR just fine, and I have verified that the appropriate instructions (which are NOPs on machines that don't support CET) exist in the compiled artifacts.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/compiler/all_arguments.llo.exp
+++ b/test/testdata/compiler/all_arguments.llo.exp
@@ -198,206 +198,206 @@ declare void @llvm.memcpy.p0i8.p0i8.i64(i8* noalias nocapture writeonly, i8* noa
 
 ; Function Attrs: nounwind ssp uwtable
 define weak i32 @sorbet_getIsReleaseBuild() local_unnamed_addr #4 {
-  %1 = load i64, i64* @rb_eRuntimeError, align 8, !tbaa !4
+  %1 = load i64, i64* @rb_eRuntimeError, align 8, !tbaa !6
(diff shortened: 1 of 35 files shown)


froydnj:

→ https://go/builds/bui_K6HJcMuiYgGQRc
→ https://go/builds/bui_K6HJyl74nIk6sM

test/testdata/compiler/all_arguments.llo.exp
```diff
@@ -1303,132 +1303,134 @@ attributes #10 = { nounwind }
 attributes #11 = { willreturn }
 attributes #12 = { nounwind allocsize(0,1) }
 
-!llvm.module.flags = !{!0}
-!llvm.dbg.cu = !{!1}
+!llvm.module.flags = !{!0, !1, !2}
```
elliottt, line 1306: Amazing that this throws off the entire diff :)

elliottt (approved):

jez (approved):

PR #4527, opened 2021-08-24
Improve typing of Enumerable#to_h; fixes #3368

elliottt:
When no arguments are present and a block is passed to `Enumerable#to_h`, skip dispatching to the internal `Sorbet::Private::Static.enumerable_to_h` method. Also improve the signature for `Enumerable#to_h` when a block is passed to return a typed hash.

### Motivation

Fixes #3368

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3134,6 +3134,11 @@ class Enumerable_toH : public IntrinsicMethod {
 public:
     // Forward Enumerable.to_h to RubyType.enumerable_to_h[self]
     void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        // Exit early when this is the case that's handled by the enumerable.rbi sig
+        if (args.args.empty() && args.block != nullptr) {
+            return;
+        }
+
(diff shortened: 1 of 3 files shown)


elliottt:

→ https://go/builds/bui_K6LcXxwy1watbW
→ https://go/builds/bui_K6LcdKEwBO4cCh

core/types/calls.cc
```diff
@@ -3134,6 +3134,11 @@ class Enumerable_toH : public IntrinsicMethod {
 public:
     // Forward Enumerable.to_h to RubyType.enumerable_to_h[self]
     void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        // Exit early when this is the case that's handled by the enumerable.rbi sig
```
jez, line 3137:
```suggestion
        // Exit early when this is the case that's handled by the enumerable.rbi sig
        // The result type will already have been populated by dispatchCall.
```

jez (approved):

core/types/calls.cc
elliottt: Thanks, this is definitely more clear.

elliottt:
Stripe migration PR: https://go/ps/pull/363275

PR #4529, opened 2021-08-25
[Ruby 3.0] Add Ractor RBI

vinistock:
Add RBI signatures for Ractors.

### Motivation

For completion sake, since we finished adding syntax support for Ruby 3.0, we should also provide all the necessary RBIs to go with it.

We didn't have any Ractor information, so I thought we'd add an RBI for it.


vinistock:
It seems the issue was outdated compiler expectation files after adding the Ractor RBI. I ran the script to update them and it seems okay now.

aprocter-stripe (approved):

PR #4532, opened 2021-08-25
delete slow already-disabled ruby benchmarks

froydnj:
### Motivation

I find myself running `./bazel test //test:compiler_oracle` a lot and I find myself noticing that these benchmarks take a long time to compile, despite them being disabled (15s+ apiece or so).  (They take *significantly* longer to compile with `-c dbg`, but that is a different story).  

We're not really using these benchmarks for anything, so I figured we could just delete them and save some time during tests.  It would be even better if we just deleted all the disabled benchmarks, IMHO, but baby steps.  (All the other deleted benchmarks take <4s to run with the aforementioned command invocation, so I left them alone.)

### Test plan

See included automated tests.


froydnj:
Please take another look; I've moved everything into the already-existing `too_slow` directory, which I think is sufficient to address the feedback here.

PR #4533, opened 2021-08-25
Auto correct initialize sigs not returning void

### LINKED ISSUE #2929: Autocorrect return type of `initialize` to void
### Problem

When making a constructor, Sorbet wants you to explicitly mark the return type as `void`:

```ruby
class Foo
  sig do
    params(x: Integer)
    .void # errors without this line
  end
  def initialize(x)
    @x = x
  end
end
```

All constructors should be `void` until the end of time, unless there's a dark corner of Ruby I've missed (which is totally possible!).

### Proposed solution

We should just automatically set the `void` return type on all constructors!

vinistock:
Closes #2929

Add an error and auto-correct for `initialize` signatures that don't return `void`.

The first commit contains the entire implementation. The second one is just fixing internal RBIs that weren't returning `void` for initializers.

### Motivation

The recommendation is to have all initializers return `void`, so we should help enforce the correct behaviour.

### Test plan

See included automated tests.

### DIFF
--- a/core/errors/rewriter.h
+++ b/core/errors/rewriter.h
@@ -13,5 +13,6 @@ constexpr ErrorClass TEnumConstNotEnumValue{3506, StrictLevel::False};
 constexpr ErrorClass BadTestEach{3507, StrictLevel::True};
 constexpr ErrorClass PropForeignStrict{3508, StrictLevel::False};
 constexpr ErrorClass ComputedBySymbol{3509, StrictLevel::False};
+constexpr ErrorClass InitializeReturnType{3510, StrictLevel::False};
 } // namespace sorbet::core::errors::Rewriter
 #endif
--- a/rbi/core/file.rbi
+++ b/rbi/core/file.rbi
@@ -1721,7 +1721,7 @@ class File::Stat < Object
     params(
         file: String,
     )
-    .returns(Object)
+    .void
   end
   def initialize(file); end
 
--- a/rbi/core/thread.rbi
+++ b/rbi/core/thread.rbi
@@ -414,7 +414,7 @@ class Thread < Object
   sig {returns(T.nilable(ThreadGroup))}
   def group; end
 
-  sig {params(args: T.untyped, blk: T.untyped).returns(Thread)}
+  sig {params(args: T.untyped, blk: T.untyped).void}
   def initialize(*args, &blk); end
 
   # The calling thread will suspend execution and run this `thr`.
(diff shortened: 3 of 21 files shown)


froydnj:

→ https://go/builds/bui_KBWoFXxlKtDNVT
→ https://go/builds/bui_KBWpkTQx9Z1ySs

froydnj (changes requested):
Thanks for the changes; sorry about the delay in getting feedback to you.  One more wrinkle:

```ruby
# typed: true

class Params < T::Struct
  extend T::Sig
  prop :raw, T.nilable(String)

  sig {params(hash: T::Hash[Symbol, T.untyped]).returns(Params)}
  def self.initialize(hash)
    Params.new(raw: hash[:raw])
  end
end
```

`self.initialize` should not be auto-corrected to return void.  (I think that means that the rewriter should not fire entirely on `self.initialize`?)

froydnj:
Another one:

```ruby
sig do
  params(
    path: String,
    key: String
  )
  .returns(T.self_type)
  .checked(:tests)
end
def initialize(path, key)
  ...
end
```

should not be auto-corrected to `.void.checked(:tests)`, but to:

```
  .void
  .checked(:tests)
```

This is a Stripe-specific rubocop, but it seems somewhat reasonable to implement here: the autocorrect shouldn't change the formatting of items before or after the `.returns/.void`.

vinistock:
The latest commit contains the fix for line breaks after the returns statement. Thanks for all the attention to the detail, this is looking a lot more robust than it was in the beginning 🚀.

froydnj:

→ https://go/builds/bui_KCFKcvo1eBGWOO
→ https://go/builds/bui_KCFKSvmff2tVVt

froydnj:

→ https://go/builds/bui_KCHITQ2h4yfhWX
→ https://go/builds/bui_KCHI2fKw6Fz1P1

froydnj (approved):
Thank you!

dduugg:
This is causing errors in `sorbet-typed`, e.g.:

```
./sorbet/rbi/sorbet-typed/lib/rainbow/all/rainbow.rbi:33: The initialize method should always return void https://srb.help/3510
    33 |      sig { params(ground: Symbol, num: Integer).returns(Indexed) }
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ./sorbet/rbi/sorbet-typed/lib/rainbow/all/rainbow.rbi:33: Replace with params(ground: Symbol, num: Integer).void
    33 |      sig { params(ground: Symbol, num: Integer).returns(Indexed) }
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

(Would it be possible to add definitely-typed to the test suite, in the manner of testing against Stripe's codebase?)

PR #4540, opened 2021-08-26
Add chained sig syntax auto correct

vinistock:
Auto-correct prototype for #4251 as requested here https://github.com/sorbet/sorbet/pull/4251#discussion_r683015746

**Note**: this branch is based on the other one. Only the last commit contains the auto-correct.

Auto-correct old style method signatures into the new chained style. E.g.:

```ruby
# These signatures
sig(:final) { void }
sig { abstract.void }
sig { overridable.void }
sig { override.void }

# are auto-corrected to
sig.final { void }
sig.abstract { void }
sig.overridable { void }
sig.override { void }
```

Screen Shot 2021-08-26 at 11 16 26 AM

### Motivation

Providing an auto-correct is important to facilitate moving over to the new syntax and avoiding having to support multiple ways of writing sigs in the future.

### Test plan

For now, this is a prototype. I will add tests once we decide to move forward with it.

/cc @jez


froydnj (changes requested):
Sorry for taking so long to get around to this...but it looks like some/all of this was landed in #4251?  Can you rebase this so that it's clearer what changes are being made in this particular PR, or close it if it's not longer applicable?

vinistock:
Closing this one for now. We'll add an auto-correct at a later PR.

PR #4551, opened 2021-08-27
Add disabled test for Regexp.new being evaluated too eagerly(?)

aprocter-stripe:
### Motivation
It looks like maybe we are somehow evaluating `Regexp.new` too eagerly (maybe as a result of some rewrite?) which causes problems when `Regexp.new` throws an exception.

Interpreter output of this test case is just `hi`, compiled output is a `RegexpError` stack trace before `hi` even gets printed.

### Test plan
See included automated tests.

### DIFF
--- a/test/testdata/compiler/disabled/regexp_error_too_eager.rb
+++ b/test/testdata/compiler/disabled/regexp_error_too_eager.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+# typed: true
+# compiled: true
+
+puts "hi"
+if T.unsafe(false)
+  Regexp.new("?")
+end


froydnj:
This is kind of weird, because it would imply that the regexp is invalid even when control hits the `Regexp.new` call, which should mean the code is busted on its own?

aprocter-stripe:
The `Regexp.new` is inside an `if T.unsafe(false)`, so control won't reach it.

froydnj:
> The `Regexp.new` is inside an `if T.unsafe(false)`, so control won't reach it.

In the testcase, sure.  But also in the relevant code that failed compilation??  Wild!

aprocter-stripe:
> In the testcase, sure. But also in the relevant code that failed compilation?? Wild!

Oh, this didn't exactly surface "in the wild" if that's what you mean. I noticed it while working on https://github.com/sorbet/sorbet/pull/4552 (evaluation was so eager that the `RegexpError` couldn't be caught as intended, so I just changed it to some other exception class since the choice seemed to be arbitrary).

PR #4553, opened 2021-08-30
don't eagerly evaluate regexps at module load time

froydnj:
### Motivation

People might "comment out" code with `Regexp.new` in it that have invalid regexps by alternate means than actual comments, e.g. `if T.unsafe(false)`.  Which implies that the `Regexp.new` call wouldn't have executed at all.  Which means that we shouldn't be eagerly creating regexps at module load time.

This PR moves to just calling the C API function as a translation for `Regexp.new` in the cases where we would have allocated a new regexp at module load time.  This is less efficient than the current scheme, but obviously correct, and I suspect people are caching their regexps anyway.  We can change things if that turns out to not be the case.

### Test plan

See included automated tests.


aprocter-stripe (commented):
So I did some blame spelunking and I guess this optimization was actually considered desirable at one point, which leaves me a little hesitant to touch it. (See issue 50 on the old repo.)

What's kinda wild is that it actually is a valid optimization in certain cases (and maybe actually a necessary one, if you take MRI's behavior as determining the semantics). Check this out:

```ruby
> begin Regexp.new("?").match?('hi') ; rescue RegexpError ; "ok" ; end
=> "ok"
> begin /?/.match?('hi') ; rescue RegexpError ; "ok" ; end
        3: from /Users/aprocter/.rbenv/versions/2.7/bin/irb:23:in `<main>'
        2: from /Users/aprocter/.rbenv/versions/2.7/bin/irb:23:in `load'
        1: from /Users/aprocter/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
SyntaxError ((irb):14: target of repeat operator is not specified: /?/)
```

It seems that when `/regex literal/` syntax is in use, Ruby lifts the regex out of the block and evaluates it eagerly (or something like that). So I think to "really" fix it we would have to disambiguate between `Regexp.new` calls and regex literals. Assuming (I haven't looked into it!) that we don't have the ability to disambiguate that by the time the Sorbet CFG reaches us, I kinda think we might want to stick with what we have, even though it is technically incorrect. What do you think?

froydnj:
> Assuming (I haven't looked into it!) that we don't have the ability to disambiguate that by the time the Sorbet CFG reaches us, I kinda think we might want to stick with what we have for the moment, even though it is technically incorrect. What do you think?

This makes sense to me!  I was kind of surprised by the failing test case and assumed it was derived from "real-world" code, but if we're not seeing this sort of thing, then I totally agree that we can just do the optimized thing always.

PR #4554, opened 2021-08-30
Fix package prefix enforcement bug with top-level casgns

ngroman:
### Motivation
A bug in package mode's prefix enforcement only considered the scope a constant was defined in when determining if nesting was correct. This simple fix also considers the LHS of the constant assignment as part of the name.

### Test plan

See included automated tests.


ngroman:
cc @aadi-stripe @jvilk-stripe

elliottt (approved):

PR #4555, opened 2021-08-30
--stripe-packages: do not consider top-level test dir a test

nroman-stripe:
### Motivation
Fix a stripe-specific edge case where the top-level `test/` dir also contains the test-harness and runtime support for tests that isn't actually a test itself. The original intent of this function was to _not_ classify these files as tests for the purposes of packages, however the path's actually contain a `./` prefix which defeated this logic.

### Test plan

See included automated tests.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -21,8 +21,8 @@ namespace {
 constexpr string_view PACKAGE_FILE_NAME = "__package.rb"sv;
 constexpr core::NameRef TEST_NAME = core::Names::Constants::Test();
 
-bool isTestFile(core::File &file) {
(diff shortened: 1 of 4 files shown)


jez:
> Fix a stripe-specific edge case where the top-level `test/` dir also contains the test-harness and runtime support for tests that isn't actually a test itself. The original intent of this function was to _not_ classify these files as tests for the purposes of packages, however the path's actually contain a `./` prefix which defeated this logic.

Is it possible to write a CLI test for this behavior? If so I'd love for us to try adding one.

Otherwise, lgtm.

jez (approved):

nroman-stripe:
> Is it possible to write a CLI test for this behavior? If so I'd love for us to try adding one.

Extended an existing CLI test to include this behavior.

PR #4557, opened 2021-08-31
Get Rubocop passing again

djudd:
It seems that the build hasn't been running Rubocop validations, so this gets them passing again. Via combination of safe autocorrects, manual fixes for Layout/SpaceInsideBlockBraces where autocorrect seems to be buggy, exclusions for generated code, and regenerating rubocop_todo.yml

Note this _doesn't_ actually enable Rubocop for the main build; I'm not sure how to do that and would need a pointer from someone on the Sorbet team

### Motivation
Linting is good

### Test plan
`bundle exec rake` is now happy locally


jez:
(heads up it looks like you created the pull request from a fork so your branch builds aren't running automatically)

jez (approved):

PR #4559, opened 2021-08-31
Run RuboCop in Buildkite

jez:
cc @djudd

### Motivation

I decided to put this here instead of linters.sh to avoid `bundle install`'ing
twice.

### Test plan

I tested what a build with a RuboCop failure looked like here:

<https://buildkite.com/sorbet/sorbet/builds/17188#92a202e3-c738-4db5-91d0-4002e53b95a2>

### DIFF
--- a/.buildkite/build-sorbet-runtime.sh
+++ b/.buildkite/build-sorbet-runtime.sh
@@ -22,7 +22,21 @@ for runtime_version in "${runtime_versions[@]}"; do
 
   rbenv exec ruby --version
 
-  rbenv exec bundle exec rake test
+  failed=
+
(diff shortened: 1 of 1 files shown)


elliottt (approved):
Seems pretty reasonable. It would be great to move this into test-static-sanitzed or test-compiler at some point in the future, as it would fail earlier and we have ruby available in those jobs already.

jez:
> It would be great to move this into test-static-sanitzed or test-compiler at some point in the future, as it would fail earlier and we have ruby available in those jobs already.

Yep, using bazel would be nice for caching and selective test execution, but I don't feel like figuring out how to install gems and run ruby in bazel right now.

PR #4561, opened 2021-08-31
Some easy Rubocop cleanup

djudd:
Autocorrect various cops without too high of an offense count; move some configuration we likely want to keep indefinitely out of rubocop_todo and into either rubocop.yml or inline; fix a couple of easy things manually.

### Motivation
Cleanup, boredom. Probably done with Rubocop stuff for now though.

### Test plan
`bundle exec rake` passes

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/modes.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/modes.rb
@@ -2,12 +2,24 @@
 # typed: true
 
 module T::Private::Methods::Modes
-  def self.standard; 'standard'; end
-  def self.abstract; 'abstract'; end
-  def self.overridable; 'overridable'; end
-  def self.override; 'override'; end
-  def self.overridable_override; 'overridable_override'; end
-  def self.untyped; 'untyped'; end
+  def self.standard
+    'standard'
+  end
+  def self.abstract
+    'abstract'
+  end
+  def self.overridable
+    'overridable'
+  end
+  def self.override
+    'override'
+  end
+  def self.overridable_override
+    'overridable_override'
+  end
+  def self.untyped
+    'untyped'
+  end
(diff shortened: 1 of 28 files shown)


djudd:
@jez btw I can't push directly to sorbet/sorbet from this account, and I'm on my personal laptop

gems/sorbet-runtime/lib/types/private/methods/modes.rb
```diff
@@ -2,12 +2,24 @@
...
+  end
+  def self.override
+    'override'
+  end
+  def self.overridable_override
+    'overridable_override'
+  end
+  def self.untyped
+    'untyped'
+  end
```
jez, line 22: 😢

jez (approved):
lgtm. would love to be able to keep around certain single-line method defs but ¯\_(ツ)_/¯ i don't feel like implementing it nor requiring it.

jez:
> btw I can't push directly to sorbet/sorbet from this account

👍 lmk if you'd like me to change that

gems/sorbet-runtime/lib/types/private/methods/modes.rb
djudd: @jez what single-line method defs do you want to keep? Where the body is just a literal? Since I was kind of just looking for something to do anyway, maybe I'll submit a PR to open-source Rubocop to add that as an option...

jez: yeah not sure, was kind of just hoping for "whatever Stripe's codebase allows," which after a quick grep looks like it's doing this:

```yaml
# -- .rubocop.yml --
# Super simple single line methods are totally fine
Style/SingleLineMethods:
  Enabled: false
```

PR #4562, opened 2021-09-01
emit constant strings as utf8 in the compiler

froydnj:
### Motivation

`rb_fstring_new` was treating the characters passed to it as ASCII.  So for the newly-enabled testcase in this PR, we were emitting the UTF-8 bytes for the constant string and reinterpreting those bytes as ASCII at runtime, which did not end well.

I am fairly confident that UTF-8 by default is Stripe's policy -- maybe even the default for Ruby? -- and I think this is the sort of change we should make anyway.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_K9F7K4JnaoELIJ
→ https://go/builds/bui_K9F73lEnjUQDOI

aprocter-stripe (approved):
Could you the following test case (which currently fails, and I think should pass with this change)?

```ruby
# frozen_string_literal: true
# typed: true
# compiled: true

puts "".encoding
```

froydnj:
> Could you add the following test case (which currently fails, and I think should pass with this change)?

Done, and verified that it currently doesn't pass!

PR #4563, opened 2021-09-01
Avoid adding synthetic `initialize` on Structs if already defined

### LINKED ISSUE #4289: Structs with custom initializers have parameters typed as `NilClass` despite initialize signature
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AConst%20%3D%20Struct.new(%3Asomething%2C%20keyword_init%3A%20true)%20do%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20params(something%3A%20Integer).void%20%7D%0A%20%20def%20initialize(something%3A%200)%0A%20%20end%0Aend%0A)

```ruby
# typed: true

Const = Struct.new(:something, keyword_init: true) do
  extend T::Sig

  sig { params(something: Integer).void }
  def initialize(something: 0)
  end
end
```

#### Observed output

```
editor.rb:3: Argument does not have asserted type Integer https://srb.help/7007
     3 |Const = Struct.new(:something, keyword_init: true) do
                            ^^^^^^^^^
    editor.rb:3:
     3 |Const = Struct.new(:something, keyword_init: true) do
                            ^^^^^^^^^
Errors: 1
```

#### Expected behavior

The Struct should be type checked considering the types defined in the overridden initializer, but it's considering that the type of the argument is `nil`. Can potentially be related to #4245.

vinistock:
Avoid adding a synthetic `initialize` on the struct rewriter if we can find an occurrence of `initialize` inside the given block.

### Motivation

Closes #4289

Avoid adding a synthetic `initialize` method with corresponding signature if a struct is already defining one on its own. If one is being manually defined, we might end up causing a conflict of expected types, as described in the linked issue.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Struct.cc
+++ b/rewriter/Struct.cc
@@ -21,7 +21,7 @@ bool isKeywordInitKey(const core::GlobalState &gs, const ast::ExpressionPtr &nod
     return false;
 }
 
-bool missingInitialize(const core::GlobalState &gs, const ast::Send *send) {
+bool isMissingInitialize(const core::GlobalState &gs, const ast::Send *send) {
     if (send->block == nullptr) {
         return true;
     }
@@ -33,6 +33,14 @@ bool missingInitialize(const core::GlobalState &gs, const ast::Send *send) {
 
         if (methodDef && methodDef->name == core::Names::initialize()) {
             return false;
+        }
+
+        for (auto &&stat : insSeq->stats) {
+            methodDef = ast::cast_tree<ast::MethodDef>(stat);
+
+            if (methodDef && methodDef->name == core::Names::initialize()) {
+                return false;
+            }
         }
     }
 
@@ -140,7 +148,7 @@ vector<ast::ExpressionPtr> Struct::run(core::MutableContext ctx, ast::Assign *as
                             std::move(typeMember)));
     }
 
-    if (missingInitialize(ctx, send)) {
+    if (isMissingInitialize(ctx, send)) {
         body.emplace_back(ast::MK::SigVoid(loc, std::move(sigArgs)));
         body.emplace_back(ast::MK::SyntheticMethod(loc, loc, core::Names::initialize(), std::move(newArgs),
                                                    ast::MK::RaiseUnimplemented(loc)));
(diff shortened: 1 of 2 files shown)


jez (approved):
Thanks!

rewriter/Struct.cc
```diff
@@ -21,6 +21,24 @@ bool isKeywordInitKey(const core::GlobalState &gs, const ast::ExpressionPtr &nod
...
+
+    auto &block = ast::cast_tree_nonnull<ast::Block>(send->block);
+
+    if (auto *insSeq = ast::cast_tree<ast::InsSeq>(block.body)) {
+        auto methodDef = ast::cast_tree<ast::MethodDef>(insSeq->expr);
+
+        if (methodDef && methodDef->name == core::Names::initialize()) {
+            return false;
+        }
+    }
```
froydnj, line 37: It seems like you'd want to check the entire `stats` vector for an appropriate `MethodDef` too?  So something like:

```ruby
Const = Struct.new(:something, keyword_init: true) do
  extend T::Sig

  sig { params(something: Integer).void }
  def initialize(something: 0)
    T.reveal_type(something) # error: Integer
  end

  def other_method
    p "just here for example purposes"
  end
end
```

should also return `false`.  IIUC, with the current code it would return `true`.

jez:

→ https://go/builds/bui_K9G0PHb9KIOy8s
→ https://go/builds/bui_K9G0CXrCU77fIU

rewriter/Struct.cc
```diff
@@ -21,6 +21,24 @@ bool isKeywordInitKey(const core::GlobalState &gs, const ast::ExpressionPtr &nod
     return false;
 }
 
+bool missingInitialize(const core::GlobalState &gs, const ast::Send *send) {
```
froydnj, line 24: Nit: `isMissingInitialize` is a better name for a predicate method.

froydnj (changes requested):
I am willing to believe I am wrong on the comment about scanning `InsSeq.stats` below, but I don't think I am, and it would be a good testcase to add either way.

rewriter/Struct.cc
vinistock: Yes, you're right we do indeed need to go through `stats` to make sure we capture the `initialize` definition regardless of its position. The latest commit contains the changes.

froydnj (approved):
Thank you!

PR #4568, opened 2021-09-01
(runtime) Add fast-path Union subclass for unions of two simple types

djudd:
This covers, e.g., `T::Boolean` and `T.nilable(Example)`.

Probably best reviewed commit-by-commit: first two are prep, final commit makes the change.

### Motivation
Duration of a simple `valid?` check for a `T.nilable(Integer)` is cut by almost 2/3rds; the effect is smaller but still noticeable with actual `sig` and `T.let` uses.

Before (updated 2021-11-06):
```
$ bundle exec rake bench:typecheck
...
T.nilable(Integer).valid?: 100.384 ns
T.any(Integer, Float, T::Boolean).valid?: 140.021 ns
T.let(..., T.nilable(Integer)): 633.98 ns
sig {params(x: T.nilable(Integer)).void}: 232.429 ns
T.let(..., T.nilable(Example)): 633.894 ns
sig {params(x: T.nilable(Example)).void}: 233.363 ns
```

After:
```
$ bundle exec rake bench:typecheck
...
T.nilable(Integer).valid?: 38.858 ns
T.any(Integer, Float, T::Boolean).valid?: 140.908 ns
T.let(..., T.nilable(Integer)): 601.688 ns
sig {params(x: T.nilable(Integer)).void}: 164.325 ns
T.let(..., T.nilable(Example)): 606.617 ns
sig {params(x: T.nilable(Example)).void}: 164.904 ns
```

### Test plan
Made sure existing Union tests covered both fast-path and normal cases, where relevant. Added tests for handling duplicate types passed to a `T.any`, which broke in a previous version of this PR.

### DIFF
--- a/gems/sorbet-runtime/bench/typecheck.rb
+++ b/gems/sorbet-runtime/bench/typecheck.rb
@@ -72,17 +72,21 @@ module SorbetBenchmarks
       end
 
       type = T::Utils.coerce(T.nilable(Integer))
-      time_block("T::Types::Union#valid?", iterations_in_block: 10) do
+      time_block("T.nilable(Integer).valid?", iterations_in_block: 5) do
         type.valid?(0)
         type.valid?(1)
         type.valid?(2)
         type.valid?(nil)
         type.valid?(false)
+      end
+
+      type = T::Utils.coerce(T.any(Integer, Float, T::Boolean))
+      time_block("T.any(Integer, Float, T::Boolean).valid?", iterations_in_block: 5) do
         type.valid?(0)
         type.valid?(1)
         type.valid?(2)
         type.valid?(nil)
-        type.valid?(false)
+        type.valid?('hi')
       end
 
       time_block("T.let(..., Integer)") do
@@ -128,26 +132,6 @@ module SorbetBenchmarks
       time_block("sig {params(s: Symbol, x: Integer, y: Integer).void} (with kwargs)") do
         arg_plus_kwargs(:foo, x: 1, y: 2)
         arg_plus_kwargs(:bar, x: 1)
-      end
-
-      time_block(".bind(example).call") do
-        Object.instance_method(:class).bind(example).call
-      end
-
-      if T::Configuration::AT_LEAST_RUBY_2_7
-        time_block(".bind_call(example)") do
-          Object.instance_method(:class).bind_call(example)
-        end
-
-        time_block("if AT_LEAST_RUBY_2_7; .bind_call(example); end") do
-          if T::Configuration::AT_LEAST_RUBY_2_7
-            Object.instance_method(:class).bind_call(example)
-          else
-            raise "must be run on 2.7"
-          end
-        end
-      else
-        puts 'skipping UnboundMethod#bind_call tests (re-run on Ruby 2.7+)'
       end
     end
 
--- a/gems/sorbet-runtime/lib/types/types/base.rb
+++ b/gems/sorbet-runtime/lib/types/types/base.rb
@@ -161,8 +161,12 @@ module T::Types
     end
 
     def ==(other)
-      (T::Utils.resolve_alias(other).class == T::Utils.resolve_alias(self).class) &&
+      case other
+      when T::Types::Base
         other.name == self.name
+      else
+        false
+      end
     end
 
     alias_method :eql?, :==
--- a/gems/sorbet-runtime/.rubocop_todo.yml
+++ b/gems/sorbet-runtime/.rubocop_todo.yml
@@ -167,6 +167,8 @@ Metrics/BlockLength:
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
   Max: 1098
+  Exclude:
+    - 'test/types/types.rb'
 
 # Offense count: 38
 # Configuration parameters: IgnoredMethods.
--- a/gems/sorbet-runtime/lib/sorbet-runtime.rb
+++ b/gems/sorbet-runtime/lib/sorbet-runtime.rb
@@ -54,6 +54,7 @@ require_relative 'types/private/types/not_typed'
 require_relative 'types/private/types/void'
 require_relative 'types/private/types/string_holder'
 require_relative 'types/private/types/type_alias'
+require_relative 'types/private/types/simple_pair_union'
 
 require_relative 'types/types/type_variable'
 require_relative 'types/types/type_member'
--- a/gems/sorbet-runtime/lib/types/private/types/simple_pair_union.rb
+++ b/gems/sorbet-runtime/lib/types/private/types/simple_pair_union.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+# typed: true
+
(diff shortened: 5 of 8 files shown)


gems/sorbet-runtime/bench/typecheck.rb
```diff
@@ -129,26 +133,6 @@ def self.run
         arg_plus_kwargs(:foo, x: 1, y: 2)
         arg_plus_kwargs(:bar, x: 1)
       end
-
-      time_block(".bind(example).call") do
```
djudd, line 133: Some of the Metrics Rubocops were complaining about this file before I removed these, and while it would probably be fine to just allowlist it, these benchmarks also didn't seem like they were adding much value now that we're done working on `bind_call`.

jez:
Nice. This is probably a good change but we're going to have to test it on Stripe's codebase to make sure nothing breaks. Technically speaking, this subclass of `T::Types::Union` doesn't pass the liskov substitution sniff test because it doesn't call the super class constructor, but maybe that's fine.

djudd:
Yeah, of course we need to test on Stripe's code too.

We could call the parent constructor in the new class, and the validation speedup would be the same, but then we might have a net slowdown in constructing new instances (which can be hot with T.let) instead of a net speedup. I thought it was better to just override the `types` accessor so that external-facing behavior would be the same; I hope we can treat the instance variables as an implementation detail.

jez:
Crazy idea would be do not have a new class but just define something like this:

```ruby
class T::Types::Union
  def initialize(type1, type2, *types)
    if types
      @types = types
      def self.types; @types; end
    else
      @type1 = type1
      @type2 = type2
      def self.types; @types ||= [type1, type2]; end
    end
  end
 end
 ```
 
 😅  maybe this is a terrible idea (maybe it doesn't even work)

**Comment by djudd** [comment]
I'd have general FUD that dynamically adding a method to an instance (the `def self.types` thing) is going to have horrible runtime performance effects (busting some method cache or whatever) in some way that might be hard to detect in a microbenchmark. But you know the Ruby VM a lot better than me and if you're not worried about that, I could be persuaded not to be either.

The standard subclassing seems fine for this use case, but there are actually other places I could imagine using a strategy like what you suggest if we were generally comfortable with it (e.g. the horrible `limit_accessible_fields` stuff in pay-server).

Not really related except speaking of wacky metaprogramming - I ran across https://github.com/benchmark-driver/benchmark-driver for the first time the other day and I think we should probably convert the benchmarks to use it. If that makes sense to you, I can file a ticket for it (ought to be an easy-first-change type of thing).

**Comment by froydnj** [comment]
@jez sounds excited about this PR and has more context anyway.  Regarding:

>  Technically speaking, this subclass of T::Types::Union doesn't pass the liskov substitution sniff test because it doesn't call the super class constructor, but maybe that's fine.

Is it reasonable to refactor part of `T::Types::Union` into `T::Types::UnionBase`, which has all the logic that doesn't care about the actual members, and then have `class Union < UnionBase` and `class PairUnion < UnionBase`?  Or is that what djudd suggested in:

> The standard subclassing seems fine for this use case

**Comment by djudd** [comment]
@froydnj since `T::Types::Union` is public I wanted to keep everything a `T::Types::Union`; I know Stripe does metaprogramming that's like:
```
case type
when T::Types::Union
  ...
```

But I could make `T::Types::Union` the base class and have two separate private subclasses if that seems better. @jez would you prefer that?

**Comment by jez** [comment]
> Not really related except speaking of wacky metaprogramming - I ran across https://github.com/benchmark-driver/benchmark-driver for the first time the other day and I think we should probably convert the benchmarks to use it. If that makes sense to you, I can file a ticket for it (ought to be an easy-first-change type of thing).

I have no preference! I think you wrote or have most heavily changed the sorbet-runtime benchmarks at this point versus anyone else on the team. Feel free to change away.

**Comment by jez** [comment]
> > This is probably a good change but we're going to have to test it on Stripe's codebase to make sure nothing breaks.
>
> Yeah, of course we need to test on Stripe's code too.

I tested this out, and it looks like it causes a bunch of changes to generated files where a generated type that used to look like `String` now looks like `T.any(String, String)`. Not sure why that is, and unfortunately I don't have a ton of time to look into it right now.

The build is here for whoever's interested:

http://go/builds/bui_KHDTNM6EiXKZoH

**Comment by froydnj** [comment]
> > > This is probably a good change but we're going to have to test it on Stripe's codebase to make sure nothing breaks.
> > 
> > 
> > Yeah, of course we need to test on Stripe's code too.
> 
> I tested this out, and it looks like it causes a bunch of changes to generated files where a generated type that used to look like `String` now looks like `T.any(String, String)`. Not sure why that is, and unfortunately I don't have a ton of time to look into it right now.

I'm pretty sure this is because `SimplePairUnion#types` needs to `.uniq` the list it stores in `@types`, just like `Union`'s constructor does (line 18):

https://github.com/sorbet/sorbet/blob/6eabfe7ab642612a32a39e0f1c401d65cce4d245/gems/sorbet-runtime/lib/types/types/union.rb#L9-L19

then `#types` will return a single-element list, and `#name` and `#type_shortcuts` on `Union` will act as expected.

Alternately, equivalent types can be checked for in `Union::Private::Pool.union_of_types` and a plain `Union` can be returned in that case.  (Though it'd be better if we could just return the single type itself.  I don't know if that would break too many things...)

**Comment by djudd** [comment]
I think `.uniq` is more of a performance cost than we want here, since it allocates. I think we can probably put a `|| type_a == type_b` in the first clause of the if statement that constructs this object instead? But I'm avoiding my work computer while on leave so I won't be able to verify anything on Stripe code.

I also agree that returning a non-union type in the case of a single-element union would probably make more sense, but seems like potentially not backwards compatible. Maybe worth trying to see if it passes on pay-server?

**Comment by djudd** [comment]
Updated this to cover the case that broke before @froydnj @jez

**Comment by djudd-stripe** [comment]

→ https://go/builds/bui_Kh0bh4t2ELPIJj
→ https://go/builds/bui_Kh0bNOnC2syvsd

File: gems/sorbet-runtime/lib/types/types/base.rb [hunk 2]
```diff
@@ -161,8 +161,12 @@ def hash
     end
 
     def ==(other)
-      (T::Utils.resolve_alias(other).class == T::Utils.resolve_alias(self).class) &&
+      case other
```
djudd, line 164: This change preserves equality between `T::Types::Union` and `T::Private::Types::SimplePairUnion` when they're equivalent (which can basically only happen if the former is explicitly constructed, circumventing `T.any`).

It's slightly wider-radius than needed - we could just do this on `T::Types::Union` - but this is a more standard way of implementing equality and will only produce different results if someone makes a class whose name makes it pretend to be a non-simple `T::Types::Base`, which doesn't seem worth worrying about (seems in the same category as overriding `class` for the current implementation).

djudd-stripe:

→ https://go/builds/bui_Kh1sCBTtgjzEBU
→ https://go/builds/bui_Kh1sarLhhCISCp

PR #4569, opened 2021-09-01
Handle merchant_token_prop in Prop.cc

jez:
### Motivation

We refactored the way that some of Stripe's database layer works (in the service
of modularity).

We need to update Sorbet to match the new logic.

### Commit summary

Review by commit; there are a lot of spurious exp file changes.

- **Move merchant_prop stuff out of prop.rb** (5ed8b9ad1)

  prop.rb is a huge test and I hate it; much rather have stuff in smaller
  files.

  No new behavior, no new tests, just a different file.

- **Add failing test case** (0c65f6b08)

- **Fix failing test** (138ec1f03)

### Test plan

See included automated tests.


jez:
cc @jnovak-stripe

jez:

→ https://go/builds/bui_K9M1lY83rac3L5
→ https://go/builds/bui_K9M1V8fwUtbSc6

rewriter/Prop.cc
```diff
@@ -152,11 +152,32 @@ optional<PropInfo> parseProp(core::MutableContext ctx, const ast::Send *send) {
...
+            // This is a lie; technically the API allows an optional `merchant:` keyword arg to
+            // customize the method name. It defaults to `:merchant`
             ret.name = core::Names::merchant();
             // 5 is the length of the _prop suffix
             ret.nameLoc = core::LocOffsets{send->loc.beginPos(), send->loc.endPos() - 5};
             ret.type = ast::MK::Constant(send->loc, core::Symbols::String());
             break;
+        case core::Names::merchantTokenProp().rawId():
+            ret.isImmutable = true;
+            // This is a lie; technically the API allows an optional `merchant:` keyword arg to
```
elliottt, line 164:
```suggestion
            // This is a lie; technically the API allows an optional `name:` keyword arg to
```

elliottt: Will defaulting the name to `:merchant` be problematic for go-to-definition or the compiler?

rewriter/Prop.cc
```diff
@@ -152,11 +152,32 @@ optional<PropInfo> parseProp(core::MutableContext ctx, const ast::Send *send) {
         }
         case core::Names::merchantProp().rawId():
             ret.isImmutable = true;
+            // This is a lie; technically the API allows an optional `merchant:` keyword arg to
```
elliottt, line 155:
```suggestion
            // This is a lie; technically the API allows an optional `name:` keyword arg to
```

elliottt (approved):
Looks good! My only question was about whether or not defaulting the name to `:merchant` is a problem.

rewriter/Prop.cc
jez: it's definitely not the right thing to do, but it's the same problem that used to exist for `merchant_prop`, and that case hasn't come up yet, so I'm not worried about fixing it in this PR

PR #4570, opened 2021-09-02
Add `T::Props::ValueObject` for `T::Struct` comparisons

### LINKED ISSUE #1540: Make `T::Struct` use same equality semantics as Ruby core `Struct`
### Problem

`T::Struct` does not have the same equality semantics as core Ruby `Struct`s.  Ruby core `Struct`s define `==` as:

```
Returns true if other has the same struct subclass and has equal member values (according to Object#==).
```

Whereas `T::Struct`s are using `Object`'s `==` (I think).

I would expect that `T::Struct`s use the same equality semantics as Ruby core `Struct`s.

```ruby
#!/usr/bin/env ruby
require 'sorbet-runtime'

class MyTypedStruct < T::Struct
  prop :my_prop, Integer
end

ts1 = MyTypedStruct.new(my_prop: 1)
ts2 = MyTypedStruct.new(my_prop: 1)

if ts1 == ts2
  puts "T::Struct's are equal"
else
  puts "T::Struct's are NOT equal"
end

MyStruct = Struct.new(:my_prop)

s1 = MyStruct.new(1)
s2 = MyStruct.new(2)

if s1 == s2
  puts "Struct's are equal"
else
  puts "Struct's are NOT equal"
end
```
```
T::Struct's are equal
Struct's are equal
```

### Proposed solution

If others agree that `T::Struct` equality should be like Ruby core `Struct` equality, then I'm not sure offhand what the best way to implement it would be.  Ruby's `Struct#==` is implemented in C.  I'm not sure what the most efficient way would be to implement it for `T::Struct`.

vinistock:
Implement equality and `hash` for prop objects in an optional module.

### Motivation

Closes #1540

Ruby structs implement equality by comparing the classes and attributes of the two instances. The fact that `T::Struct` does not do the same is often surprising or confusing to developers. However, it's likely that at this point a lot of code depends on the old behavior, which does not match regular structs.

So, we are suggesting adding an optional module that can be included in any class that includes `T::Props` to then implement equality like Ruby structs. This approach allows devs to have more control as they migrate to use the new equality.

### Test plan

See included automated tests.

cc/ @elliottt

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/value_object.rb
+++ b/gems/sorbet-runtime/lib/types/props/value_object.rb
@@ -1,0 +1,32 @@
+# typed: strict
+# frozen_string_literal: true
+
+module T::Props::ValueObject
+  extend T::Sig
+
+  # A struct is equal to another if the classes and all prop definitions (attributes) and values are the same
+  sig {params(other: Object).returns(T::Boolean)}
+  def ==(other)
+    return false unless self.class == other.class
+
+    other_prop = T.cast(other, T::Props::ValueObject)
+    decorator = self.class.decorator
+    other_decorator = other_prop.class.decorator
+    return false unless decorator.all_props == other_decorator.all_props
+
+    decorator.all_props.all? do |prop|
+      decorator.get(self, prop) == other_decorator.get(other_prop, prop)
+    end
+  end
+
+  alias eql? ==
+
+  # Match Struct#hash by using both the class and all prop definitions (attributes) and values to calculate the hash
+  sig {returns(Integer)}
+  def hash
+    decorator = self.class.decorator
+    prop_values = decorator.all_props.map {|prop| decorator.get(self, prop)}
+
+    [self.class, decorator.all_props, prop_values].hash
+  end
+end
--- a/gems/sorbet-runtime/lib/sorbet-runtime.rb
+++ b/gems/sorbet-runtime/lib/sorbet-runtime.rb
@@ -105,6 +105,7 @@ require_relative 'types/props/serializable'
 require_relative 'types/props/type_validation'
 require_relative 'types/props/private/parser'
 require_relative 'types/props/generated_code_validation'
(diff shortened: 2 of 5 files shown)


jez (commented):
Why do this in sorbet-runtime? Structural / value equality is a can of worms that i don't really want to open right now. it looks like all the code you've written in this PR could just as easily live in a library outside of sorbet-runtime. if we put it in sorbet-runtime, it can immediately be used by all stripe code, including the parts of stripe's custom ORM that are entangled with `T::Props`, which takes care to get right.

vinistock:
Having it live inside `sorbet-runtime` makes it easier to maintain, as the related code for props is all inside of it as well. So, we'd catch changes in the runtime that break this functionality faster. I also think it makes sense to for it to be package together with all the prop related things.

Would it be an option to have a custom cop prohibit the usage of this module inside Stripe's code base (or the relevant parts of it)?

gems/sorbet-runtime/lib/types/props/value_object.rb
```diff
@@ -0,0 +1,28 @@
...
+  def ==(other)
+    return false unless self.class == other.class
+
+    decorator = self.class.decorator
+    return false unless decorator.all_props == other.class.decorator.all_props
+
+    decorator.all_props.all? do |prop|
+      decorator.get(self, prop) == decorator.get(other, prop)
+    end
+  end
```
openbl, line 18: Curious if you thought about adding this so any class mixing this in can also be used as a Hash key?<sup>[1](https://ruby-doc.org/core-3.0.2/Hash.html#class-Hash-label-User-Defined+Hash+Keys)</sup>
```suggestion
  end
  alias eql? ==
```

vinistock: I think it makes sense, it's the same behavior as regular Ruby structs. I added the alias and an additional test for using value objects as hash keys.

PR #4572, opened 2021-09-02
Stop mutating the kwargs hash, always check for required kwargs

elliottt:
We're currently removing entries from the keyword args hash as we parse it, which is causing test failures for some internal tests that use meta-programming to intercept arguments before they're given to the compiled function. The VM does this as well, but only in the event that there is a double-splat in the arguments list; when there is no double splat in the args list the vm will only read elements out of the hash that contains the keyword args. We now match that behavior, and only mutate the keyword args hash when pruning out required and optional args from the hash that will be used as the double splat.

While I was modifying keyword argument parsing, I improved handling of missing keyword arguments by following the lead from the VM and threading through an array that contains the missing keyword args. The array is initialized the first time we encounter a missing arg, so we only pay the cost for it when we would raise an error anyway. Improving the handling of missing required keyword args fixes a disabled test, and brings us more in line with the behavior of the VM.

### Motivation

Fixing bugs

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/IREmitter.cc
+++ b/compiler/IREmitter/IREmitter.cc
@@ -417,6 +417,15 @@ void setupArguments(CompilerState &base, cfg::CFG &cfg, const ast::MethodDef &md
                                     builder, irctx, rubyBlockId);
                 }
                 if (hasKWArgs) {
+                    // required arguments remaining to be parsed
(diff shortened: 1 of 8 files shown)


elliottt:

→ https://go/builds/bui_KA5UF9OtYLsu0c
→ https://go/builds/bui_KA5UH7uyQDqMYr

jez (approved):
Looks great, thanks!

PR #4575, opened 2021-09-04
[wip] Apple Silicon support

jez:
### Motivation

This is nowhere near ready. I just want a place where I can push code that will
eventually be required, so that I don't forget it.

### Test plan

🙃


jez:
broke this out into #4576 and #4578

still need to emit an arm64 binary, but that can be another PR

PR #4576, opened 2021-09-04
Detect the CPU, install the same Bazel version

jez:
### Motivation

Using x86_64 Bazel on Apple Silicon Mac's happens to work right now, but
when we want to start building arm64 binaries directly, that'll need to
change. This should preserve the current state of things while also
allowing for experimentation with only small changes.

### Test plan

- [x] Linux Buildkite CI
- [x] Linux personal PC
- [x] Linux Stripe devbox
- [x] Linux Stripe CI
- [x] macOS Buildkite CI
- [x] macOS Stripe laptop
- [x] macOS personal laptop
- [x] ~~macOS Stripe arm64 laptop~~
  - I can't even get master to build on arm64, so I don't think I'm going to block this PR on it.


jez:
(this is ready for review now)

elliottt (approved):

PR #4577, opened 2021-09-04
cfg-view: Open SVGs with default browser

jez:
### Motivation

Open SVGs with user's default SVG program.

I prefer using Firefox on my personal laptop. I think this script should respect
whichever program people prefer to open SVGs with.

To get the old behavior, change the default program for opening SVGs:

<https://www.imore.com/how-set-mac-app-default-when-opening-file>

(alternatively: we could add a command-line flag for this if people feel
strongly)

### Test plan

n/a


froydnj (approved):
I am a fan of anything which moves away from using Chrome. ;)

Issue #4566, opened 2021-09-01
Runtime error when using `class << self` syntax


RyanBrushett:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20frozen_string_literal%3A%20true%0A%23%20typed%3A%20true%0A%23%20compiled%3A%20true%0A%0Aclass%20Foo%0A%20%20class%20%3C%3C%20self%0A%20%20%20%20def%20bar%0A%20%20%20%20%20%20%22hello%22%0A%20%20%20%20end%0A%20%20end%0Aend%0A%0Aputs%20Foo.bar%0A)

```ruby
# frozen_string_literal: true
# typed: true
# compiled: true

class Foo
  class << self
    def bar
      "hello"
    end
  end
end

puts Foo.bar
```

#### Observed output

An error message during compilation:

```
│test/testdata/compiler/classfields.rb:6:in `<class:Foo>': wrong constant name Object<Class:Object> (RuntimeError)
│        from test/testdata/compiler/classfields.rb:5:in `<top (required)>'
│        from /private/var/tmp/bazel/.../com_stripe_ruby_typer/bazel-out/darwin-fastbuild/bin/external/sorbet_ruby_2_7/toolchain/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
│        from /private/var/tmp/bazel/.../com_stripe_ruby_typer/bazel-out/darwin-fastbuild/bin/external/sorbet_ruby_2_7/toolchain/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
│        from /private/var/tmp/bazel/.../com_stripe_ruby_typer/test/patch_require.rb:33:in `require'
│        from /var/folders/vf/wxc53gg16lb269pj4nh29bcr0000gn/T/tmp.eCLDI4mo:1:in `<main>'
```

#### Expected behavior

Compile and execute!

My example is a modified version of the test at [test/testdata/compiler/classfields.rb](https://github.com/sorbet/sorbet/blob/bd1ca6ba27eae7fee3128556220571a4e9ef30ba/test/testdata/compiler/classfields.rb). In that file, the `class << self` example is commented out, uncommenting it demonstrates the error.

Rewriting the example like `def self.bar # ...` allows it to compile happily!

froydnj:
Thanks!  We have a few testcases for this at the moment:

[...]

PR #4580, opened 2021-09-06
implement support for `class << self` in the compiler

froydnj:
### Motivation

People like to use this syntax, Sorbet supports it, so the compiler should too.

The core idea is pretty straightforward: whenever we have a `class << self` class (a singleton of a singleton) that we're going to access at runtime, we can't ask `Payload::getRubyConstant` to fetch it for us (since it has an invalid constant name, which also implies that we can't define it as a class "directly"): we need to manually access the singleton class at runtime via `rb_singleton_class`.  Once we do that, most everything Just Works.

It's debatable whether this support should be added to `Payload::getRubyConstant` directly.

I think there is an oddity exposed by this: in most cases for defining methods, we should be trusting the `self` that was passed in to the `<static-init>` method, rather than engaging in runtime reflection to find the type of `self` as a Ruby constant.  If we did that, I think some of the hackery in `DefineMethodIntrinsic` goes away.

### Test plan

See included automated tests.  Credit to @RyanBrushett for the testcase from #4566.

It's very possible that we don't handle "nested" `class << self` (which is kind of a weird construct) or `class << self` at the toplevel.  I'm not terribly worried about these, but it's possible we need to support them in Stripe's codebase (I haven't looked, but I hope not).


compiler/IREmitter/NameBasedIntrinsics.cc
```diff
@@ -82,29 +82,34 @@ class DefineClassIntrinsic : public NameBasedIntrinsicMethod {
...
-                auto rawCall = Payload::getRubyConstant(cs, sym.data(cs)->superClass(), builder);
-                module =
-                    builder.CreateCall(cs.getFunction("sorbet_defineNestedClass"), {getOwner, classNameCStr, rawCall});
+                if (isModule) {
+                    module = builder.CreateCall(cs.getFunction("sorbet_defineTopLevelModule"), {classNameCStr});
+                } else {
+                    auto rawCall = Payload::getRubyConstant(cs, sym.data(cs)->superClass(), builder);
+                    module =
+                        builder.CreateCall(cs.getFunction("sorbet_defineTopClassOrModule"), {classNameCStr, rawCall});
+                }
```
aprocter-stripe, line 108: Just voicing my understanding out loud in case it's wrong: all the green code in this block is pre-existing, it was just moved inside the `if(!attachedClass.data(cs)->isSingletonClass(cs))`.

aprocter-stripe (approved):

compiler/IREmitter/NameBasedIntrinsics.cc
froydnj: This is a correct summary of the situation!

PR #4583, opened 2021-09-07
Add a failing test for captured kwsplats

elliottt:
Add a disabled test case showing that the behavior of compiled code with a keyword splat differs from the vm when the arguments are captured.

### Motivation

More tests.

### Test plan

See included automated tests.


froydnj (approved):
I take it that you will fix up the comments when this is re-enabled?

elliottt:
> I take it that you will fix up the comments when this is re-enabled?

Yep, if we fix this case we could adjust the description. I'm not sure it's worth fixing right now, as #4572 handles the immediate issues with stubbing in tests.

PR #4584, opened 2021-09-08
use `uname -m` for platform detection in bazel wrapper script

froydnj:
### Motivation

I pulled and compiled this morning:

```
froydnj@pyro:~/src/sorbet$ ./bazel build //main:sorbet //compiler:sorbet
No cached Bazel v4.2.1 found, installing...
Building on linux-unknown is not implemented
froydnj@pyro:~/src/sorbet$ uname -p
unknown
```

which seems bad.  The help output for uname says `-p` is "non-portable"; I'm not exactly sure what that means in this context, but there you go.

`uname -m` seems to do the right thing (returns `x86_64`) on that machine, on two different other Linux machines I have at home, on a Stripe devbox, and on my Stripe (x86-64) mac.  As a bonus on the (Big Sur) mac, `uname -m` returns `x86_64` whereas `uname -p` returns `i386`, so I think we can get rid of the gross case in the `bazel` wrapper.

### Test plan

See included automated tests.

### DIFF
--- a/bazel
+++ b/bazel
@@ -50,7 +50,7 @@ fi
 echo >&2 "No cached Bazel v$BAZEL_VERSION found, installing..."
 
 kernel_name="$(uname -s | tr 'A-Z' 'a-z')"
-processor_name="$(uname -p)"
+processor_name="$(uname -m)"
 
 bazel_installer_platform="${kernel_name}-${processor_name}"
 case "$bazel_installer_platform" in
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_KBnqT9gdhfdHeH
→ https://go/builds/bui_KBnqChBYKgo9ns

froydnj:
`uname -m` produces reasonable results (`x86_64` instead of `i386`) on a High Sierra (10.12) and Catalina (10.15) Mac.  It also builds OK on Stripe CI.

jez (approved):
Thanks!

PR #4586, opened 2021-09-08
Start emitting defaults for block args

elliottt:
Begin emitting logic for setting defaults in block arguments. Currently the defaults are ignored and block args will always be initialized as `nil`. This shows up in the disabled test [test/testdata/compiler/disabled/block_with_defaults.rb](https://github.com/sorbet/sorbet/blob/cb55f38db4906c593445ff66d182fd0fc82ff920/test/testdata/compiler/disabled/block_with_defaults.rb), which will produce the wrong value when compiled.

The change is to emit similar logic to the method entry for handling default arguments, guarding them with a new instruction `YieldParamPresent` (similar behavior to `ArgPresent`, but specific to yield parameters in blocks). That instruction returns a boolean result, indicating whether or not the argument was present in the array of arguments given to the block.
Screen Shot 2021-09-08 at 2 00 45 PM

This PR adds the new instruction, and hard-codes its compiled behavior to return `true` which preserves the current behavior. The next step will be to unify the handling of defaults for block and method parameters in `setupArguments`.

### Motivation

Prework for handling block argument defaults.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/IREmitter.cc
+++ b/compiler/IREmitter/IREmitter.cc
@@ -663,8 +663,8 @@ void emitUserBody(CompilerState &base, cfg::CFG &cfg, const IREmitterContext &ir
                     /* intentionally omitted, it's part of method preambula */
                 },
                 [&](cfg::YieldParamPresent *i) {
-                    failCompilation(cs, core::Loc(cfg.file, bind.loc),
-                                    "Unable to compile blocks with default arguments!");
+                    auto *val = Payload::rubyTrue(cs, builder);
+                    Payload::varSet(cs, bind.bind.variable, val, builder, irctx, bb->rubyBlockId);
                 },
                 [&](cfg::Cast *i) {
                     auto val = Payload::varGet(cs, i->value.variable, builder, irctx, bb->rubyBlockId);
--- a/compiler/IREmitter/IREmitter.cc
+++ b/compiler/IREmitter/IREmitter.cc
@@ -662,6 +662,10 @@ void emitUserBody(CompilerState &base, cfg::CFG &cfg, const IREmitterContext &ir
                     loadYieldParamsResults.insert(bind.bind.variable);
                     /* intentionally omitted, it's part of method preambula */
                 },
+                [&](cfg::YieldParamPresent *i) {
+                    failCompilation(cs, core::Loc(cfg.file, bind.loc),
+                                    "Unable to compile blocks with default arguments!");
+                },
                 [&](cfg::Cast *i) {
                     auto val = Payload::varGet(cs, i->value.variable, builder, irctx, bb->rubyBlockId);
 
--- a/compiler/IREmitter/IREmitterContext.cc
+++ b/compiler/IREmitter/IREmitterContext.cc
@@ -662,6 +662,44 @@ vector<optional<string>> getBlockLocationNames(CompilerState &cs, cfg::CFG &cfg,
     return blockLocationNames;
 }
 
+void collectRubyBlockArgs(const cfg::CFG &cfg, const cfg::BasicBlock *b, vector<cfg::LocalRef> &blockArgs) {
+    bool insideThenBlock = false;
+
+    while (true) {
+        for (auto &maybeCallOnLoadYieldArg : b->exprs) {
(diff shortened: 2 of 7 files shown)


compiler/IREmitter/IREmitterContext.cc
```diff
@@ -843,6 +872,11 @@ IREmitterContext IREmitterContext::getSorbetBlocks2LLVMBlockMapping(CompilerStat
                 userEntryBlockByFunction[ensureBlockId] = llvmBlocks[ensureBlock->id];
             }
         } else if (b->bexit.cond.variable.data(cfg)._name == core::Names::argPresent()) {
+            // TODO(trevor) we don't currently handle argPresent decisions in ruby-Argblocks, only for the method entry.
```
elliottt, line 875: I'm happy to remove this TODO, but my plan was to address it immediately in a follow-up PR.

compiler/IREmitter/IREmitter.cc
```diff
@@ -662,6 +662,10 @@ void emitUserBody(CompilerState &base, cfg::CFG &cfg, const IREmitterContext &ir
                     loadYieldParamsResults.insert(bind.bind.variable);
                     /* intentionally omitted, it's part of method preambula */
                 },
+                [&](cfg::YieldParamPresent *i) {
+                    auto *val = Payload::rubyTrue(cs, builder);
+                    Payload::varSet(cs, bind.bind.variable, val, builder, irctx, bb->rubyBlockId);
```
elliottt, line 667: This behavior is temporary, and after unifying the behavior for method and block default arguments it will be a no-op.

elliottt:

→ https://go/builds/bui_KBwVtM0iAhaOq1
→ https://go/builds/bui_KBwVCtOr0PP4gr

jez:
summary of DM:

> **jez**
> just checking, because my first instinct whenever i see new cfg instructions: could this be a magic send instead?
>
> **trevor**
> when i implemented `ArgPresent` i thought the same thing, but dmitry pushed me to make it an instruction. i was just reusing that original decision this time around 🙂
>
> **jez**
> sounds good
>
> have you thought through how it will work with the cfg_finalizer things, variable pinning, etc.?
>
> (basically, do a quick check of all the places where we're down casting on the instruction type and see if it makes sense to add an extra case in those places)
>
> **trevor**
> we initialize all block args to `T.untyped` in the cfg, so it should be a no-op
>
> the problems that i remember from when we added `ArgPresent` were related to assignments in the presence of pinning. do we pin block args?
>
> **jez**
>
> > do we pin block args?
>
> not sure. but i would probably just look at all these grep results and see whether we should change any of them:
>
> ```
> ~/sorbet/cfg/builder
> ❯ rg 'cast_instruction|isa_instruction'
> builder_entry.cc
> 92:    if (cont->exprs.empty() || isa_instruction<LoadArg>(cont->exprs.back().value.get())) {
>
> builder_finalize.cc
> 192:            if (auto *i = cast_instruction<Ident>(bind.value.get())) {
> 210:                if (auto *v = cast_instruction<Ident>(bind.value.get())) {
> 212:                } else if (auto *v = cast_instruction<Send>(bind.value.get())) {
> 217:                } else if (auto *v = cast_instruction<TAbsurd>(bind.value.get())) {
> 219:                } else if (auto *v = cast_instruction<Return>(bind.value.get())) {
> 225:            if (auto *i = cast_instruction<Ident>(bind.value.get())) {
> 272:                                          if (isa_instruction<Ident>(bind.value.get()) ||
> 273:                                              isa_instruction<Literal>(bind.value.get()) ||
> 274:                                              isa_instruction<LoadSelf>(bind.value.get()) ||
> 275:                                              isa_instruction<LoadArg>(bind.value.get()) ||
> 276:                                              isa_instruction<LoadYieldParams>(bind.value.get())) {
> ```
>
> **trevor**
>
> i grepped around for `ArgPresent` and used that to guide where i'd need to add cases for `YieldParamPresent`
> i'm also going to run it on pay-server before merging 👍
>
> **jez**
> 👍 sounds good. this might just be one extra thing to do. i'll look at the rest of the change

jez (approved):
Looks great, thanks!

elliottt:
Stripe migration PR: https://go/ps/pull/367874

PR #4587, opened 2021-09-09
handle return not directly appearing inside blocks

froydnj:
### Motivation

Our return-from-block support was too narrow: we only considered the case of `return` appearing directly inside a block.  But as the testcase from #4585 shows, we really need to consider any case where a block appears as the ancestor of the location of the return.

This means that we need to change the conditions for a) computing `IREmitterContext.hasReturnFromBlock` and b) deciding when a `cfg::Return` statement merits throwing the value rather than just returning the value normally.

Since the new conditions are more general, I've chosen to rename everything to use "AcrossBlock" terminology rather than "FromBlock" terminology.  This change in decision-making also means that a bunch of `ENFORCE`s in `IREmitterHelpers.cc` had to go.

### Test plan

See included automated tests.


compiler/IREmitter/IREmitter.cc
```diff
@@ -609,9 +609,17 @@ void emitUserBody(CompilerState &base, cfg::CFG &cfg, const IREmitterContext &ir
                     isTerminated = true;
 
                     auto *var = Payload::varGet(cs, i->what.variable, builder, irctx, bb->rubyBlockId);
+                    bool hasBlockAncestor = false;
+                    int rubyBlockId = bb->rubyBlockId;
 
-                    if (irctx.rubyBlockType[bb->rubyBlockId] == FunctionType::Block) {
-                        IREmitterHelpers::emitReturnFromBlock(cs, cfg, builder, irctx, bb->rubyBlockId, var);
+                    while (rubyBlockId != 0) {
+                        hasBlockAncestor = hasBlockAncestor || irctx.rubyBlockType[rubyBlockId] == FunctionType::Block;
```
aprocter-stripe, line 616: Might be nice to move the deleted `ENFORCE`s up here? (To be honest, though, I'm not sure how critical they really are.)

```suggestion
                        if (irctx.rubyBlockType[rubyBlockId] == FunctionType::Block) {
                            ENFORCE(!functionTypeNeedsPostprocessing(irctx.rubyBlockType[rubyBlockId]));
                            ENFORCE(!functionTypePushesFrame(irctx.rubyBlockType[rubyBlockId]));
                            hasBlockAncestor = true;
                            break;
                        }
```

aprocter-stripe (approved):
Looks great! Small suggestion wrt the deleted `ENFORCE`s.

compiler/IREmitter/IREmitter.cc
froydnj: At least the `functionTypeNeedsPostprocessing` one is reasonable to keep around, since that would imply there's some typechecking we need to do.  I think technically we'd want to assert that for the entire ancestor chain, because otherwise we might inadvertently miss some postprocessing by doing the throw.  Does that change sound good to you?

I think we can drop the `functionTypePushesFrame` one.

aprocter-stripe: > I think technically we'd want to assert that for the entire ancestor chain, because otherwise we might inadvertently miss some postprocessing by doing the throw. Does that change sound good to you?

Yep, that's a good point! I'd be confused and bewildered if the assert ended up failing anywhere between the return statement and the enclosing `FunctionType::Block`.

> I think we can drop the functionTypePushesFrame one.

Agreed.

PR #4591, opened 2021-09-10
reject <Magic>.$method at compile time, rather than runtime

froydnj:
### Motivation

`<Magic>` methods should always be implemented by the compiler, so if we ever try to have the Ruby VM execute one, we should fail at compile time, rather than getting peculiar errors at runtime.

I tried and failed to make this check via `core::Types::isSubType`, but that wasn't working for reasons I don't completely understand.  So it's done manually.

### Test plan

There are no tests because we shouldn't be generating such calls generally.  I did verify that some of the changes in #4558 that introduce `<Magic>` methods where I hadn't implemented the corresponding calls in the compiler did fail, as expected.

### DIFF
--- a/compiler/IREmitter/sends.cc
+++ b/compiler/IREmitter/sends.cc
@@ -457,6 +457,18 @@ llvm::Value *IREmitterHelpers::emitMethodCallViaRubyVM(MethodCallContext &mcctx)
     auto &irctx = mcctx.irctx;
     auto rubyBlockId = mcctx.rubyBlockId;
 
+    // If we get here with <Magic>, then something has gone wrong.
+    // TODO(froydnj): We want to do the same thing with Sorbet::Private::Static,
+    // but we'd need to do some surgery on either a) making those methods not
+    // be emitted via symbol-based intrinsics or b) making it possible for
+    // (some) symbol-based intrinsics to bypass typechecks completely.
+    if (auto *at = core::cast_type<core::AppliedType>(send->recv.type)) {
+        if (at->klass == core::Symbols::MagicSingleton()) {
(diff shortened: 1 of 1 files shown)


compiler/IREmitter/sends.cc
```diff
@@ -457,6 +457,18 @@ llvm::Value *IREmitterHelpers::emitMethodCallViaRubyVM(MethodCallContext &mcctx)
...
 
+    // If we get here with <Magic>, then something has gone wrong.
+    // TODO(froydnj): We want to do the same thing with Sorbet::Private::Static,
+    // but we'd need to do some surgery on either a) making those methods not
+    // be emitted via symbol-based intrinsics or b) making it possible for
+    // (some) symbol-based intrinsics to bypass typechecks completely.
+    if (auto *at = core::cast_type<core::AppliedType>(send->recv.type)) {
+        if (at->klass == core::Symbols::MagicSingleton()) {
+            failCompilation(cs, core::Loc(irctx.cfg.file, send->receiverLoc),
+                            "No runtime implemention for <Magic> method exists");
```
jez, line 468: Can we interpolate the magic method's name into the error message here?

jez (approved):
one suggestion

compiler/IREmitter/sends.cc
froydnj: Unfortunately not at the moment, because of how `failCompilation` is defined:

https://github.com/sorbet/sorbet/blob/b4d6348ae968f44c2de13a3dff0c2318589e7956/compiler/Core/CompilerState.h#L71-L74

We can change that in a followup, though?

PR #4592, opened 2021-09-10
be sure to test both sides of @@var ||= ... in the compiler

froydnj:
### Motivation

This test is disabled, but I discovered that it doesn't provide good coverage of whatever AST we provide because we're not testing both cases of `||=`. :man_facepalming:  This PR fixes that.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/compiler/disabled/uninitialized_class_variable.rb
+++ b/test/testdata/compiler/disabled/uninitialized_class_variable.rb
@@ -14,4 +14,6 @@ class A
 end
 
 A.init_change('constant')
+A.init_change('other')
 p A.get_change('constant')
+p A.get_change('other')


aprocter-stripe (approved):

aprocter-stripe:
(ooc, why is this test currently failing?)

froydnj:
> (ooc, why is this test currently failing?)

Because `@@var ||= ...` gets [desugared](https://github.com/sorbet/sorbet/blob/b4d6348ae968f44c2de13a3dff0c2318589e7956/ast/desugar/Desugar.cc#L1016-L1021) into:

```ruby
if @@var
  @@var
else
  @@var = ...
end
```

which is not correct from the perspective of the VM: class variables are not permitted to be accessed before initializing them.  (This is maddeningly different from instance variables, but there you go.)  So you have to do something more like:

```ruby
if defined?(@@var) && @@var
  @@var
else
  @@var = ...
end
```

but then the compiler doesn't properly implement `defined?` on class variables, which is being addressed in #4558 (along with a better desugaring).

aprocter-stripe:
Oh! Gotcha---I had it in my head that #4558 had already landed :sweat_smile:

PR #4593, opened 2021-09-10
strengthen the tests in ivar_cvar_oror_eq.rb to cover more cases

froydnj:
### Motivation

I thought I was done with #4558 until jez mentioned another test I should try, which led to more tests being added here.  The `@@falsy_var` tests are to explicitly test that inference drops `FalseClass` in the true branches of the desugaring of `||=`.  This property is trivially true now, but #4558 brings some changes that make it not trivially true.

### Test plan

See included automated tests.


elliottt (approved):
:shipit:

PR #4596, opened 2021-09-10
Adds TaskRabbit as a user of Sorbet

kbacha:
Adds TaskRabbit as a user of sorbet and includes an SVG version of our logo.

### Motivation

We have been using Sorbet incrementally over the past couple years and are beginning to broaden our adoption of it. As such, we thought it would be nice to be listed along side the other users.

### Test plan

No tests written because this is a static asset.


aprocter-stripe (approved):
Looks great! Thanks very much!

PR #4598, opened 2021-09-11
[LSP] Go to implementation

ilyailya:
This is reopen of https://github.com/sorbet/sorbet/pull/3607

The change brings a "Go to implementation" feature for the Sorbet LSP. 
The algorithm behind the feature is trivial: if a request was called on the method node, I go up to the owner class node of the function. If the owner class is abstract, I go over all of the defined classes and pick ones, which derive from owner class. Then find a method, which overrides the originally requested one. Logic for the other cases is similar. The solution also supports reopened modules.

The use cases I was aiming to cover:
1. Call from the definition of the interface class or abstract method

```ruby
module Runnable
     # ^ call from here works
  extend T::Sig
  extend T::Helpers
  interface!

  sig { abstract.params(args: T::Array[String]).void }
  def run(args); end
     # ^ call from here works too
end
```

2. Call from the definition of the class which implements the interface

```ruby
class HelloWorld
  extend T::Sig
  include Runnable
         # ^ call from here works
  sig { override.params(args: T::Array[String]).void }
  def run(args)
    # ^ call from here works too
    puts 'Hello, World!'
  end
end
```

3. Calls from references of the interface or abstract method

```ruby
sig {params(runnable: Runnable).void}
                    # ^ call from here works
def self.run(runnable)
  runnable.run([])
           # ^ call from here works
end

hello = T.let(HelloWorld.new, Runnable)
                            # ^ call from here works 
hello.run([])
    # ^ call from here works 
HelloWorld.new.main([])
             # ^ call from here works 
```

### Motivation

https://github.com/sorbet/sorbet/issues/3435

### Test plan

I've tested the change in a sandbox project. I've also tested it in pay-server on one of the popular interfaces. Seems to work as expected!

Automated tests included


ilyailya:

→ https://go/builds/bui_KDQKgbqIUC57UX
→ https://go/builds/bui_KDQKOG1T0OdkCs

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,152 @@
...
+
+    vector<unique_ptr<Location>> result;
+    auto queryResponse = move(queryResult.responses[0]);
+    if (auto def = queryResponse->isDefinition()) {
+        // User called "Go to Implementation" from the abstract function definition
+        core::SymbolRef method = def->symbol;
+        if (!method.data(gs)->isMethod())
+            response->error = make_unique<ResponseError>(
+                (int)LSPErrorCodes::InvalidParams,
+                "Go to implementation can be used only for methods or references of abstract classes");
```
elliottt, line 86: Could you put braces around this conditional? Also, does it make sense to continue processing results at this point? The body of the conditionals on line 106 and 127 return after setting the error code, which makes me think it might be worth doing the same here.

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,152 @@
...
+const MethodImplementationResults findMethodImplementations(const core::GlobalState &gs, core::SymbolRef method) {
+    if (!method.data(gs)->isMethod() || !method.data(gs)->isAbstract()) {
+        return {.error = {makeInvalidParamsError(
+                    "Go to implementation can be used only for methods or references of abstract classes")}};
+    }
+
+    vector<core::Loc> locations;
+    auto owner = method.data(gs)->owner;
+    if (!owner.isClassOrModule())
+        return {.error = {makeInvalidParamsError("Abstract method can only be inside a class or module")}};
```
elliottt, line 32: Could you add braces to this conditional?

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,152 @@
...
+
+unique_ptr<ResponseError> makeInvalidParamsError(std::string error) {
+    return make_unique<ResponseError>((int)LSPErrorCodes::InvalidParams, error);
+}
+
+const MethodImplementationResults findMethodImplementations(const core::GlobalState &gs, core::SymbolRef method) {
+    if (!method.data(gs)->isMethod() || !method.data(gs)->isAbstract()) {
+        return {.error = {makeInvalidParamsError(
+                    "Go to implementation can be used only for methods or references of abstract classes")}};
+    }
```
elliottt, line 27: What about making a default-initialized `MethodImplementationResults` at the top of the method, and then mutating its fields before returning in each case?
```suggestion
    MethodImplementationResults res;
    if (!method.data(gs)->isMethod() || !method.data(gs)->isAbstract()) {
        res.error = makeInvalidParamsError("Go to implementation can be used only for methods or references of abstract classes");
        return res;
    }
```

This way you can avoid the initializer labels, and be more explicit about what's being returned here.

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,152 @@
...
+
+    auto owningClassSymbolRef = owner.asClassOrModuleRef();
+    auto childClasses = owningClassSymbolRef.getSubclasses(gs, false);
+    auto methodName = method.data(gs)->name;
+    for (const auto &childClass : childClasses) {
+        auto methodImplementation = childClass.data(gs)->findMember(gs, methodName);
+        locations.push_back(methodImplementation.data(gs)->loc());
+    }
+
+    return {.locations = {locations}};
```
elliottt, line 42: If you go the route of adding the default-initialized `res` at the top of the method, you could make the following change as well:

```suggestion
    res.locations = locations;
    return res;
```

main/lsp/requests/go_to_implementation.cc
```diff
@@ -0,0 +1,152 @@
...
+
+        if (mainResponse.errors.size() != 0) {
+            response->error = makeInvalidParamsError("Failed to fetch implementations");
+            return response;
+        }
+
+        core::SymbolRef overridedMethod = core::SymbolRef(calledMethod);
+        if (calledMethod.data(gs)->isOverride()) {
+            overridedMethod = findOverridedMethod(gs, overridedMethod);
+        }
```
elliottt, line 135:
```suggestion
        // User called "Go to Implementation" from the abstract function call
        if (mainResponse.errors.size() != 0) {
            response->error = makeInvalidParamsError("Failed to fetch implementations");
            return response;
        }

        auto calledMethod = mainResponse.method;
        core::SymbolRef overridedMethod = calledMethod;
        if (calledMethod.data(gs)->isOverride()) {
            overridedMethod = findOverridedMethod(gs, overridedMethod);
        }
```
The constructor for `SymbolRef` that takes a `MethodRef` argument is not marked explicit, so you don't need be explicit when assigning it to a `core::SymbolRef`. However, if you do want to be explicit, I would recommend something like `core::SymbolRef overridedMethod{calledMethod};` instead.

elliottt (changes requested):
This is looking great, I'm really looking forward to using this feature!

.gitignore
```diff
@@ -9,6 +9,7 @@
 
 .idea/
 .vscode/
+.vim/
```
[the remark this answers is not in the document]
ilyailya: Yep, I store local `coc-settings.json` here. It sets up the clangd support for me. I'm being lazy to migrate from coc to native lsp client :p

If it spoils our `.gitingore`, I can remove it no problem.

main/lsp/requests/go_to_implementation.cc
ilyailya: Sounds great, looks much better this way

elliottt (approved):
Looks good to me! Feel free to merge after addressing the namespace comment -- no need to re-r :)

PR #4599, opened 2021-09-11
mangle Ruby method names to avoid dynamic linker interactions

froydnj:
### Motivation

It has always puzzled me why the functions generated for Ruby methods have external linkage, rather than internal linkage.  There's no particular reason for them to be external; the best argument I could think of was that if you defined the same method in different files, you'd want the dynamic linker to fix one as The Implementation.  But that seems like a weak argument (shouldn't every method be defined once?) and an awful problem to debug ("this method is doing something completely different than it says in this file!").  It doesn't even work if you wanted to guard against something like autoloading loading the different files in different orders, depending on the vagaries of how code gets run.

Actually, C backtraces from Ruby errors might be more informative if you're dealing with stripped binaries, because external symbols (and therefore dynamic symbols) are guaranteed to have symbol names associated with them, whereas internal functions wouldn't.  Note that Ruby's error handler is smart enough to find names for internal symbols, so assuming you don't strip binaries (and you probably shouldn't for server apps), there should be no reason to have method implementation symbols be externally visible.

Overall, it seems more consistent with how code is actually written to give things internal linkage and not let the dynamic linker mess with things.  Making the implementations internal also makes things ever so slightly faster.

So that is a simple change.  What's (somewhat) surprising about this change is that it also fixes a testcase (!).  Why does that happen?  Well, that's a fascinating story that involves the dynamic linker too.

The test in question, `test/testdata/compiler/disabled/negate.rb`, would fail with a most peculiar error (I enabled it for the purposes of this exposition):

```
[ .. ] ├─ Full compiled output (except bazel paths shortened):
       │SorbetLLVM using compiled: /home/froydnj/.cache/bazel/bazel/.../com_stripe_ruby_typer/test/testdata/compiler/negate.sorbet.build/test/testdata/compiler/negate.rb.so for /home/froydnj/src/sorbet/test/testdata/compiler/negate.rb
       │test/testdata/compiler/negate.rb:20:in `-@': wrong number of arguments (given 0, expected 1) (ArgumentError)
       │        from test/testdata/compiler/negate.rb:21:in `-'
       │        from test/testdata/compiler/negate.rb:33:in `<top (required)>'
       │        from /home/froydnj/.cache/bazel/bazel/.../com_stripe_ruby_typer/bazel-out/k8-fastbuild/bin/external/sorbet_ruby_2_7/toolchain/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
       │        from /home/froydnj/.cache/bazel/bazel/.../com_stripe_ruby_typer/bazel-out/k8-fastbuild/bin/external/sorbet_ruby_2_7/toolchain/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
       │        from /home/froydnj/.cache/bazel/bazel/.../com_stripe_ruby_typer/test/patch_require.rb:33:in `require'
       │        from /tmp/tmp.7DRrRbepWA:1:in `<main>'
```

You can see the code here:

https://github.com/sorbet/sorbet/blob/e99b27f6dc6d716568c3fe7191bfcdbcbfa2d90a/test/testdata/compiler/disabled/negate.rb#L14-L34

The sequence is:

1. We construct two `Amount`s
2. We subtract them, so we call `Amount#-`
3. We negate the argument, so we call `Amount#-@` (`-@` is the name for unary minus)
4. We get an `ArgumentError` in `Amount#-@`...but it's saying we expected an argument, which makes no sense for unary minus.

This makes no sense, especially when you look at the LLVM IR and see that everything is being generated as you would expect.  You look at the assembly generated and see that everything is being generated as you expect.  But at runtime, when `-@` is defined, it gets defined with the implementation for `-`...what?

To understand why this is happening, we need a short digression into implementation details.

ELF, the object file format on Linux (and almost all non-OS X Unices), has the concept of symbol versions.  glibc, among other libraries, uses symbol versioning to maintain backwards compatibility: external symbols (i.e. the public API of the library) are given versions corresponding in some way to the release in which they first appeared.  If, for whatever reason, something changes in the implementation of an external symbol that requires breaking backwards compatibility, the new implementation can be given a different symbol version.  Old code -- which may have depended on the previously broken behavior -- can remain blissfully unaware, while code compiled against the new library can start using the new improved behavior.

Versioned symbols are named differently, either `name@ver` or `name@@ver`.  The difference doesn't concern us here; the pertinent thing to note is that the `@` character is significant in symbol names.

If you look at the symbol table for the generated `negate.rb.so` + some filtering, you see:

```
froydnj@pyro:~/src/sorbet$ readelf -s -W test/testdata/compiler/negate.rb.so |egrep 'func_Amount#-|Size|table' |grep -v stackFrame
Symbol table '.dynsym' contains 53 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
    12: 0000000000002870   262 FUNC    GLOBAL DEFAULT    7 func_Amount#-
    52: 0000000000002750   288 FUNC    GLOBAL DEFAULT    7 func_Amount#-
Symbol table '.symtab' contains 156 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
   115: 0000000000002870   262 FUNC    GLOBAL DEFAULT    7 func_Amount#-@
   155: 0000000000002750   288 FUNC    GLOBAL DEFAULT    7 func_Amount#-
```

The compiler, for `Amount#-`, generates a function called `func_Amount#-`, and similarly for `Amount#-@`.  We can see in the above output that `.symtab` contains these two functions, as expected.

But the dynamic symbols (i.e. what the dynamic linker will use for symbol resolution) are different: we have two different addresses with the *same* symbol name.  AIUI, this is because `func_Amount#-@` is really "`func_Amount#-` with a null (?) version" as far as the dynamic linker is concerned.

Therefore at runtime, when the code runs what we thought was:

```
sorbet_vm_defined_method(klass, "-@", func_Amount#-@, ...);
```

for that `func_Amount#-@` argument, what actually happens is that we ask the dynamic linker "hey, please give me a pointer to the function named `func_Amount#-@`".  Which, thanks to symbol versioning, gets translated to "hey, please give me a pointer to the function named `func_Amount#-`".  And we have the problem leading to the peculiar error message from Ruby above, where we are getting an error for argument arity from a completely different method.

Making method implementations internal therefore cuts the dynamic linker out of the picture, fixing the testcase, and enables us to keep the same naming scheme that we're currently using.  I noted in a comment that if we did want to keep methods external for some reason, we'd need to come up with some sort of name mangling scheme to avoid having `@` characters in function names.

### Test plan

See included automated tests.


dmitry-stripe:
IMHO this doesen't fix the underlying bug, but instead hides it. Could you add the following testcase?

```
def foo@1(a)
  1
end

def foo@2(a, b)
  2
end

  foo@1(1) + foo@2(2,3)
```

froydnj:
> IMHO this doesen't fix the underlying bug, but instead hides it. Could you add the following testcase?
> 
> ```
> def foo@1(a)
>   1
> end
> 
> def foo@2(a, b)
>   2
> end
> 
> p foo@1(1) + foo@2(2,3)
> ```

This isn't valid Ruby syntax; is there a syntax to let you define methods with names containing `@` that aren't unary methods (`-@`, e.g.)?

You can, of course, use `define_method` to get `@` symbols in names, but then you wouldn't have C-level functions with `@` in their names in the first place.

froydnj:
Dmitry also pointed out that we may want method implementations to have external visibility for cross-file `final` method optimizations...which if that is enabled, I don't think we have any tests for.

froydnj:
Turns out we had to implement name mangling after all, since cross-file `final` method calls need at least the relevant method available.  I still think this is suboptimal, but maybe there are ways of addressing this so that we don't slow down all method calls.

compiler/IREmitter/IREmitterHelpers.cc
```diff
@@ -58,7 +59,22 @@ string IREmitterHelpers::getFunctionName(CompilerState &cs, core::SymbolRef sym)
...
+    bool isFirst = true;
+    for (auto part : absl::StrSplit(suffix, '@')) {
+        if (!isFirst) {
+            mangled += "at";
+        } else {
+            isFirst = false;
+        }
+        mangled += to_string(part.size());
+        mangled += part;
+    }
```
elliottt, line 75: Is prefixing each part with the length sufficient to avoid collisions?

elliottt (changes requested):
I think this makes sense, but it would be great to add a test that exercises the `@` case for the escaping.

jez:
> I think this makes sense, but it would be great to add a test that exercises the `@` case for the escaping.

If you want help thinking up an interesting test case:

#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%20%3C%20T%3A%3AStruct%0A%20%20prop%20%3A%22foo%401%22%2C%20T.nilable%28Integer%29%0A%20%20prop%20%3A%22foo%402%22%2C%20T.nilable%28Integer%29%0Aend%0A%0Aas%20%3D%20T%3A%3AArray%5BA%5D.new%28%5BA.new%28%22foo%401%22%3A%2017%2C%20%22foo%402%22%3A%2042%29%5D%29%0AT.reveal_type%28as.map%28%26%3A%22foo%401%22%29%29)

```ruby
# typed: true

class A < T::Struct
  prop :"foo@1", T.nilable(Integer)
  prop :"foo@2", T.nilable(Integer)
end

as = T::Array[A].new([A.new("foo@1": 17, "foo@2": 42)])
T.reveal_type(as.map(&:"foo@1"))
```

#### Observed output

```
editor.rb:9: Revealed type: T::Array[T.nilable(Integer)] https://srb.help/7014
     9 |T.reveal_type(as.map(&:"foo@1"))
  Got T::Array[T.nilable(Integer)] originating from:
    editor.rb:9:
     9 |T.reveal_type(as.map(&:"foo@1"))
                      ^^^^^^^^^^^^^^^^^
Errors: 1
```

froydnj:
> I think this makes sense, but it would be great to add a test that exercises the `@` case for the escaping.

negate.rb, enabled in the very first commit, exercises this precise thing!

elliottt (approved):

elliottt:
> negate.rb, enabled in the very first commit, exercises this precise thing!

Thanks for pointing this out, I missed that it was moved out of `disabled` :D

PR #4601, opened 2021-09-13
generate correct code for `@@var ||= ...` in the compiler

froydnj:
### Motivation

This PR builds on #4558 to provide a correct translation for `@@var ||= ...` in the compiler.  The current expansion:

```ruby
if @@var
  @@var
else
  @@var = ...
end
```

is not permitted, because the VM will throw errors if you attempt to access a class variable prior to its definition.  This is different behavior from instance variables, which spring to life fully-formed on their first use.

The expansion of `@@var ||= ...` by the VM bytecode, in pseudocode, is:

```ruby
if defined?(@@var) && @@var
  @@var
else
  @@var = ...
end
```

The above is fairly easy to implement in desugar.  We implement the conditional with a single send for efficiency:

```ruby
if <Magic>.<defined-or-truthy-class-var>(:@@var)
  @@var
else
  @@var = ...
end
```

But we need to be a little more clever.  

Sorbet takes advantage of being able to know that if we take the true branch, we can narrow the type of `@@var`.  Our send in the proposed expansion above hides that information.  So we need to introduce type derivation for `<Magic>.<defined-or-truthy-class-var>`.  I initially implemented this in terms of the existing support for `<Magic>.<defined-class-var>` before I realized that we need slightly different semantics.  The commits therefore have type derivation added for `<Magic>.<defined-class-var>` first; `<defined-or-truthy-class-var>` comes later.

We also need to know the type of the send so we can issue unreachable code errors; we have tests for this.

But this comes with a second issue: we now have type information for the send, but there's no way of propagating that knowledge to the type associated with `@@var`.  (This would be done automatically by Sorbet if we were just testing `@@var` in the conditional.)  We therefore have to explicitly tell Sorbet that the variable is definitely truthy.  I tried using `T.must` here, but there are two problems:

1. `T.must` doesn't have the right semantics (the variable is not just non-`nil`, but non-`false`, too); and
2. `T.must` will issue errors when you apply it to things that can be non-`nil`.

Therefore, we now have `<Magic>.<must-be-truthy>` that implies truthiness of its result and avoids issuing errors.

Finally, as alluded to above, just using the `defined?` support in #4558 is not enough: we need `<Magic>.<defined-or-truthy-class-var>` and we take some steps to make this and `<Magic>.<must-be-truthy>` somewhat more efficient than naive expansions would be.

### Test plan

See included automated tests.


plugin_injector/plugin_injector.cc
```diff
@@ -147,6 +147,9 @@ class LLVMSemanticExtension : public SemanticExtension {
                     switch (send->fun.rawId()) {
                         case core::Names::keepForIde().rawId():
                         case core::Names::keepForTypechecking().rawId():
+                        // This method exists only to make inference work in certain
+                        // cases and doesn't actually compute anything.
+                        case core::Names::mustBeTruthy().rawId():
```
jez, line 152: ooc, why do this here instead of with a no-op intrinsic?

froydnj: Because a no-op intrinsic isn't really a no-op: it still assigns into the alias associated with the the class variable.  And if we want to get the right value to do that, we still have to get the value associated with the class variable.  Which is a whole bunch of unnecessary overhead.

I will admit this is not super clean, and it's also potentially bad: if we use the intrinsic in other contexts, it might not be safe to elide the intrinsic entirely.  There is probably a better way to do this, but I'm not sure what it is.  Maybe this is another reasons to have a real compiler-specific IR?

froydnj: > it still assigns into the alias associated with the the class variable

Note that this is a class variable set (`rb_cvar_set`, which is not super-cheap, since class variables don't have caches associated with them the way instance variables do).

core/types/calls.cc
```diff
@@ -1461,36 +1461,54 @@ class T_attached_class : public IntrinsicMethod {
...
-class T_must : public IntrinsicMethod {
-public:
-    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
-        if (args.args.empty()) {
-            return;
+namespace {
+
+enum class MustBehavior {
+    T,
+    Magic,
```
jez, line 1468: Slight preference for naming this according to what functionality it gates (pedantic redundancy check vs not) instead of tracking which Ruby method uses which (not saying this will necessarily happen soon, but imagine adding one more intrinsic that uses `implementMust`: it'd probably want to specify which behavior to get, like "give me pedantic checks," rather than have to say "give me the `T.must` behavior."

core/types/calls.cc
```diff
@@ -1461,36 +1461,54 @@ class T_attached_class : public IntrinsicMethod {
...
+};
+
+void implementMust(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res, MustBehavior behavior) {
+    if (args.args.empty()) {
+        return;
+    }
+    const auto loc = args.callLoc();
+    if (!args.args[0]->type.isFullyDefined()) {
+        if (behavior == MustBehavior::Magic) {
+            /* Nothing */
```
jez, line 1478: If you're already committed to exhausting the elements of the `enum class`, you may as well switch this from an `if` to a `switch` so that Clang will warn for future additions to this enum.

core/types/calls.cc
```diff
@@ -1461,36 +1461,54 @@ class T_attached_class : public IntrinsicMethod {
...
+            const auto locWithoutTMust = loc.adjust(gs, 7, -1);
+            if (loc.exists() && locWithoutTMust.exists()) {
+                e.replaceWith("Remove `T.must`", loc, "{}", locWithoutTMust.source(gs).value());
             }
         }
-        res.returnType = move(ret);
+    }
+    if (behavior == MustBehavior::Magic) {
+        ret = Types::approximateSubtract(gs, ret, Types::falseClass());
+    }
```
jez, line 1503: Slight preference for moving this out of `implementMust` and up into `Magic_mustBeTruthy` (i.e., after calling `implementMust`)

core/types/calls.cc
```diff
@@ -2919,6 +2937,81 @@ class Magic_mergeHashValues : public IntrinsicMethod {
...
+    if (args.args.size() != 2) {
+        return;
+    }
+
+    auto &receiver = args.args[0];
+    if (receiver->type.isUntyped()) {
+        res.returnType = receiver->type;
+        return;
+    }
+    if (!receiver->type.isFullyDefined()) {
```
jez, line 2953: Do you have a test for this?

core/types/calls.cc
```diff
@@ -2919,6 +2937,81 @@ class Magic_mergeHashValues : public IntrinsicMethod {
...
+    }
+
+    auto &classVar = args.args[1];
+    auto classLit = cast_type_nonnull<LiteralType>(classVar->type);
+    auto classVarName = classLit.asName(gs);
+
+    SymbolRef klass;
+    if (isa_type<AppliedType>(receiver->type)) {
+        klass = cast_type_nonnull<AppliedType>(receiver->type).klass;
+        klass = klass.data(gs)->attachedClass(gs);
```
jez, line 2964: Why are these cases not symmetric w.r.t. their treatment of `attachedClass`?

I would have expected to either see them both take attached class on the symbol or to see neither of them.

Do you have test cases for this?

core/types/calls.cc
```diff
@@ -2919,6 +2937,81 @@ class Magic_mergeHashValues : public IntrinsicMethod {
...
+        ENFORCE(klass.exists());
+    } else if (isa_type<ClassType>(receiver->type)) {
+        klass = cast_type_nonnull<ClassType>(receiver->type).symbol;
+
+    } else {
```
jez, line 2969:
```suggestion
        klass = cast_type_nonnull<ClassType>(receiver->type).symbol;
    } else {
```

core/types/calls.cc
```diff
@@ -2919,6 +2937,81 @@ class Magic_mergeHashValues : public IntrinsicMethod {
...
+} // namespace
+
+class Magic_definedInstanceVar : public IntrinsicMethod {
+    // We want `<Magic>.<defined-instance-var>(self, :varname)` to appear to inference
+    // as `@varname` so things like `if <Magic>.<defined-instance-var>(...)` will potentially
+    // warn about unreachable code.
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        convertDefinedToType(gs, args, res);
+    }
+} Magic_definedInstanceVar;
```
jez, line 2995: Does this PR silence the error that Sorbet currently reports for this snippet?

```ruby
# typed: strict

class A
  @@x ||= @@x
end
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20A%0A%20%20%40%40x%20%7C%7C%3D%20%40%40x%0Aend)

```
editor.rb:4: Use of undeclared variable @@x https://srb.help/6002
     4 |  @@x ||= @@x
          ^^^
Errors: 1
```

jez: if not, do we have a test case like this already?

test/testdata/infer/unreachable_code_class_variable.rb
```diff
@@ -0,0 +1,15 @@
+# typed: true
+
+extend T::Sig
+
+class C
+  @@x = T.let(42, Integer)
+
+  def f
+    @@x ||= 43 # error: This code is unreachable
```
jez, line 9: I haven't looked at the other tests that exist for this stuff, but do you think it's worth adding a test like

```ruby
@@y = T.let(0, T.any(FalseClass, Integer)

def self.h
  @@y ||= 1
end
```

jez: i.e., something that doesn't conflate "undefined" and "defined but `nil`" but still wants to use `||=`?

(because the unreachable code error would prevent the compiler from activating)

jez (commented):
Some small suggestions but mostly looks good.

core/types/calls.cc
froydnj: I do not.  This was cargo-culted from the `<Magic>.<call-with-*>` intrinsics.  I'm not entirely sure how one would write a test for this, and it seems reasonable to give up on such types in any event?

test/testdata/infer/unreachable_code_class_variable.rb
froydnj: Sorbet [doesn't give an error](https://sorbet.run/#%23%20typed%3A%20true%0A%0A%23%20Test%20that%20Sorbet%20correctly%20infers%20the%20types%20of%20class%20variables%20and%20instance%0A%23%20variables%20when%20they%20are%20used%20as%20the%20LHS%20of%20%7C%7C%3D.%0A%0Aclass%20C%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20%40%40never_falsy%20%3D%20T.let%280%2C%20T.any%28FalseClass%2C%20Integer%29%29%0A%0A%20%20sig%20%7Bparams%28x%3A%20Integer%29.returns%28Integer%29%7D%0A%20%20def%20self.f6%28x%29%0A%20%20%20%20%40%40never_falsy%20%7C%7C%3D%20x%20%23%20error%3A%20This%20code%20is%20unreachable%0A%20%20end%0A%0A%20%20sig%20%7Bparams%28x%3A%20Integer%29.returns%28Integer%29%7D%0A%20%20def%20h6%28x%29%0A%20%20%20%20%40%40never_falsy%20%7C%7C%3D%20x%20%23%20error%3A%20This%20code%20is%20unreachable%0A%20%20end%0Aend%0A) on this today.  I don't know if it's worth adding a test.

froydnj (commented):
Some extra testcases and whatnot pushed.  I'm inclined to keep `MustBehavior` the way it is, because I think it's easier to understand than `MustBehavior::Errors` or `MustBehavior::PedanticErrors` or `MustBehavior::Silent`.

core/types/calls.cc
```diff
@@ -1461,36 +1461,54 @@ class T_attached_class : public IntrinsicMethod {
...
+            if (loc.exists() && locWithoutTMust.exists()) {
+                e.replaceWith("Remove `T.must`", loc, "{}", locWithoutTMust.source(gs).value());
             }
         }
-        res.returnType = move(ret);
+    }
+    if (behavior == MustBehavior::Magic) {
+        ret = Types::approximateSubtract(gs, ret, Types::falseClass());
+    }
+    res.returnType = move(ret);
```
jez, line 1513: > It's not clear to me that res.returnType will always be appropriately defined after we return from this function, hence doing it here.

I'm not sure how it wouldn't be? I'm suggesting this here:

```suggestion
    res.returnType = move(ret);
```

and this in `Magic_mustBeTruthy` below:

```cpp
        implementMust(gs, args, res, MustBehavior::Magic);
        auto resultType = res.returnType;
        res.returnType = Types::approximateSubtract(gs, resultType, Types::falseClass());
```

core/types/calls.cc
jez:
```ruby
# typed: true

class A
  extend T::Generic

  Elem = type_member

  @@x = T.let(nil, T.nilable(Elem))

  sig {params(x: Elem).void}
  def foo(x)
    @@x ||= x
  end
end
```

Your PR crashes with a backtrace on this test case.

core/types/calls.cc
jez: Do you mind linking me to those?

core/types/calls.cc
jez: (continued below)

core/types/calls.cc
jez: btw this (well, and the crash posted above) is the thing holding me back from approving this PR. it looks suspicious, so i'm eager to learn more why it's needed.

jez (changes requested):
comments inline

core/types/calls.cc
froydnj: Sure, but you can return from that function without setting `res.returnType`, no?  Or is it true that `res.returnType` is always set to *something*, and intrinsic functions are just refining the type in some way?

jez: > Sure, but you can return from that function without setting `res.returnType`, no? Or is it true that `res.returnType` is always set to _something_, and intrinsic functions are just refining the type in some way?

Yes, if you don't set the `DispatchResult::returnType`, it will just stay as what it was default initialized to (`nullptr`) and fall through to being set using the `resultType` if a signature is present:

https://github.com/sorbet/sorbet/blob/db1373aa424b5fe1a02e1e71d4032cf0ebd6eeac/core/types/calls.cc#L1176-L1186

or defaulted to `T.untyped` if no signature was present:

https://github.com/sorbet/sorbet/blob/db1373aa424b5fe1a02e1e71d4032cf0ebd6eeac/core/types/calls.cc#L1210-L1211

**BUT** in neither case here are we not setting it. In `implementMust` we'd be setting to `ret` (unconditionally) and only in `Magic_mustBeTruthy` would we be mutating it to drop subtypes of `FalseClass`.

froydnj: Thanks for the explanation.  I still think it makes more sense to have the subtraction of `FalseClass` in `implementMust` itself, but :man_shrugging: .

froydnj:
I have been frustrated with how complicated this PR was getting, and last night I had an epiphany: we can handle this all in the compiler when we are branching on class variables.  The PR is now much simpler and much easier to reason about.

jez (approved):
🎉  amazing

jez:
> I have been frustrated with how complicated this PR was getting, and last night I had an epiphany: we can handle this all in the compiler when we are branching on class variables. The PR is now much simpler and much easier to reason about.

Great! You should test this on Stripe's codebase once just to be sure.

froydnj:

→ https://go/builds/bui_KEWGR12jyPwPyq
→ https://go/builds/bui_KEWG2dr9tqFQa8

PR #4606, opened 2021-09-14
delete random type member generated by the Struct.new rewriter

froydnj:
### Motivation

This type member doesn't make any sense: I guess maybe we might have had it in for generating `Struct` declarations with extra type information?  But then I would think we'd want per-field type information, not type information on the structure level.

Deleting it doesn't seem to cause any harm; we'll see what happens with respect to Stripe's codebase.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_KE64Sp5Re0jrHC
→ https://go/builds/bui_KE64tiiGmtq2Yc

froydnj:
Oh man, `Struct` inherits from `Enumerable`, which explains that.  Never mind!

PR #4607, opened 2021-09-14
fix cross-file final method calling

froydnj:
### Motivation

This came up in #4599, but it also came up when compiling more code in Stripe's codebase: when we do a send from file A to a final method F defined in file B, we need for two things to be visible from the compiled artifact from file A:

1. The function associated with the method F
2. The iseq associated with the method F

We need the former to actually execute the call, and the latter to be able to push the appropriate Ruby frame.  (Normally the iseq would be stored as part of the method definition inside the VM and dealt with during method sends, but since we are calling things directly, we need to do more of the work ourselves.)  The iseq was not external to the compiled artifact, resulting in LLVM assertions in debug mode and linker failures in opt mode.

The fix is easy: just give iseqs external linkage.  Some testcases are added to verify that compiled->compiled final method calls work, compiled->interpreted final method calls are OK with the undefined external linkage symbol (I don't think this would be the case on Mac, but we're not super-worried about Mac at the moment), and finally, we clean up one final method-related testcase to use filecheck testing (because we were in the area).

### Test plan

See included automated tests.


test/testdata/compiler/cross_file_compiled_final_method__2.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+# typed: true
+# compiled: true
+
+def f
+  Utils.final_method
```
elliottt, line 6: Is it worth adding some filecheck annotations here?

elliottt (approved):
Nice find!

test/testdata/compiler/cross_file_compiled_final_method__2.rb
froydnj: We have filechecks in `call_final.rb` and other places; I'm not sure what additional ones we would have here.

PR #4609, opened 2021-09-15
Add a failing test for `proc do |*args|`

elliottt:
Add a test case for an example that the compiler can't handle.

### Motivation

More tests!

### Test plan

See included automated tests.


jez (approved):
Do you think it’s worth assigning it to a variable and calling it as well? This test doesn’t print anything

elliottt:
> Do you think it’s worth assigning it to a variable and calling it as well? This test doesn’t print anything

Yes, that's a good idea. Given that the compiler was crashing when compiling it this felt like a good enough start, but I'm happy to add some prints as well.

PR #4610, opened 2021-09-15
add some more testcases for operations on @{,@}vars

froydnj:
### Motivation

Splitting these out from #4601, since they're not really relevant in that PR anymore.  We didn't have test coverage of some of these cases before.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/infer/unreachable_code_instance_variable.rb
+++ b/test/testdata/infer/unreachable_code_instance_variable.rb
@@ -1,0 +1,13 @@
+# typed: true
+
+extend T::Sig
+
+class C
+  def initialize
+    @x = T.let(42, Integer)
+  end
+
+  def f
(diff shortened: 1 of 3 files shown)


test/testdata/infer/unreachable_code_instance_variable.rb
```diff
@@ -0,0 +1,13 @@
...
+extend T::Sig
+
+class C
+  def initialize
+    @x = T.let(42, Integer)
+  end
+
+  def f
+    @x ||= 43 # error: This code is unreachable
+  end
```
aprocter-stripe, line 12: I think this case is _not_ covered in `ivar_cvar_oror_eq.rb` (`g1` and `g2` only deal with nilable case) but it might make sense to fold it into that file.

aprocter-stripe (approved):
No blocking comments!

PR #4614, opened 2021-09-15
Make DispatchArgs::suppressErrors required

jez:
### Motivation

Errors were not always being suppressed properly, because certain
locations forgot to propagate the old value when constructing a new
`DispatchArgs` struct.

This meant that we were actually doing substantially more work than we
meant to.

This might also fix a source of non-deterministic behavior in Sorbet on
Stripe's codebase (not yet tested) but the performance benefits (not yet
measured) should be justification enough to land this.

### Test plan

Existing tests.

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -830,7 +830,7 @@ struct DispatchArgs {
     // Do not produce dispatch-related errors while evaluating the call. This is a performance optimization, as there
     // are cases where we call dispatchCall with no intention of showing the errors to the user. Producing those
     // unreported errors is expensive!
-    bool suppressErrors = false;
+    bool suppressErrors;
 
     Loc callLoc() const {
         return core::Loc(locs.file, locs.call);
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1331,7 +1331,8 @@ DispatchResult MetaType::dispatchCall(const GlobalState &gs, const DispatchArgs 
                                           wrapped,
                                           args.block,
                                           args.originForUninitialized,
-                                          args.isPrivateOk};
+                                          args.isPrivateOk,
+                                          args.suppressErrors};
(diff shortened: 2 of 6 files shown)


jez:

→ https://go/builds/bui_KEYCPANlNQ2UqV
→ https://go/builds/bui_KEYDHpbH4D3GEj

froydnj (approved):
Whoops!

I have tried and failed at least twice to give `DispatchArgs` a proper constructor, but I keep getting very very strange sanitizer errors every time I do and I haven't tracked them down yet.  (I can't build sanitizer builds on my local machine, for some reason...)

jez:
|     | old times | new times |
| --- | --------- | --------- |
|     | 20.838s   | 19.605s   |
|     | 20.910s   | 19.836s   |
|     | 21.012s   | 20.112s   |
|     | 22.605s   | 20.262s   |
| **avg** | **21.341s**   | **19.954s**   |

| stat  | delta   |
| ---   | ---     |
| Δ min | −1.233s |
| Δ max | −2.343s |
| Δ avg | −1.388s |

jez:
This appears to have fixed the bug (80 type check runs on Stripe's codebase and no errors). Prior experience showed that the bug would have manifested 2 times in that many runs.

I'm not sure of a way to get this to reproduce outside of Stripe's codebase. Unfortunately, I'm going to have to go ahead and land this without a test, so we won't have a good way to prevent this from recurring.

It's honestly rather surprising to me that this behavior could be racy. Maybe we're like... flushing the error queue eagerly when we shouldn't be? @jvilk-stripe do you know why that might be? When the bug manifested, it manifested in Sorbet's batch mode, so I can't imagine that LSP eager error-queue flushing shenanigans would have been at play, but figured I'd ask anyways.

PR #4616, opened 2021-09-16
handle nil blocks in yield

froydnj:
### Motivation

I am so sad to write this patch, but one can have nil blocks passed to `yield`, and `yield` has to make a very specific error to match the Ruby interpreter.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_KEqKJNrLVbPlzm
→ https://go/builds/bui_KEqK40sYH3sgM3

elliottt (approved):

PR #4619, opened 2021-09-16
only allow same-file optimized final method calls

froydnj:
### Motivation

This is a temporary measure; the more complete fix to allow cross-file final method calls is more involved, and we'd like to unblock other work.

The newly-enabled tests show that this commit resolves the problems with loading compiled objects with external references to function stack frame data structures.

### Test plan

See included automated tests.


aprocter-stripe (approved):
LGTM! Do we have a Jira for (eventually) fixing this?

froydnj:
> LGTM! Do we have a Jira for (eventually) fixing this?

We do not, but I can certainly file one!

PR #4621, opened 2021-09-17
remove builderCast from the compiler

froydnj:
### Motivation

The LLVM API weirdness that motivated this is gone; `llvm::IRBuilderBase` contains basically all the functionality we want to access in any IR-building function in LLVM 12.  So let's get rid of this wart.

### Test plan

See included automated tests.


elliottt (approved):
Oh man, this is amazing!

PR #4623, opened 2021-09-17
error if Struct.new is given fields that end in `=`

froydnj:
### Motivation

`Struct.new` will error at runtime if you give it a field name that ends in `=`:

https://github.com/ruby/ruby/blob/ruby_2_7/struct.c#L555-L559

```
froydnj@pyro:~/src/sorbet$ ~/ruby/bin/irb
# Unfortunately irb overwrites the erroneous line, but trust me that it was `Foo = Struct.new(:baz=)`
        5: from /home/froydnj/ruby/bin/irb:23:in `<main>'
        4: from /home/froydnj/ruby/bin/irb:23:in `load'
        3: from /home/froydnj/ruby/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        2: from (irb):1
        1: from (irb):1:in `new'
ArgumentError (invalid struct member: baz=)
```

We might as well detect this statically.

Suggestions on the error message welcome.

### Test plan

See included automated tests.

### DIFF
--- a/core/errors/rewriter.h
+++ b/core/errors/rewriter.h
@@ -14,5 +14,6 @@ constexpr ErrorClass BadTestEach{3507, StrictLevel::True};
 constexpr ErrorClass PropForeignStrict{3508, StrictLevel::False};
 constexpr ErrorClass ComputedBySymbol{3509, StrictLevel::False};
 constexpr ErrorClass InitializeReturnType{3510, StrictLevel::False};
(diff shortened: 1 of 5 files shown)


froydnj:

→ https://go/builds/bui_KFDtKzc4PBgCMK
→ https://go/builds/bui_KFDtEpCMFl4V5z

jez (approved):
lgtm, i think that `shortName` returns a string_view so all this should involve no new string allocations.

PR #4624, opened 2021-09-17
Track the compiled status in File

elliottt:
Track the compiled level for a file in the same way that we do the typed level. This avoids re-parsing the file in multiple places in the compiler, and opens the door to tracking the compiled status of files more generally.

### Motivation

Avoiding duplicating work, pre-work for tracking the compiled status of files.

### Test plan

See included automated tests.

### DIFF
--- a/core/CompiledLevel.h
+++ b/core/CompiledLevel.h
@@ -1,0 +1,19 @@
+#ifndef SORBET_CORE_COMPILED_LEVEL_H
+#define SORBET_CORE_COMPILED_LEVEL_H
+
+namespace sorbet::core {
+enum class CompiledLevel {
+
+    // This file has no compiled sigil present. The behavior here is the same as for `# compiled: false`, but it's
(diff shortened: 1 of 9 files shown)


core/CompiledLevel.h
```diff
@@ -0,0 +1,19 @@
...
+
+    // This file has no compiled sigil present. The behavior here is the same as for `# compiled: false`, but it's
+    // useful to distinguish files that have no sigil present for stats.
+    Unknown = 0,
+
+    // This file is explicitly opted-out of compilation.
+    False = 1,
+
+    // This file should be compiled.
+    True = 1,
```
froydnj, line 15: You probably want this to be `2`?

elliottt: I sure do. That's what I get for adding `None` after the others :)

froydnj (approved):
WFM!

PR #4627, opened 2021-09-20
Add -p compilied-files-json

elliottt:
Add the `compiled-files-json` print option, which prints out a json array that contains objects describing the compiled status of each file processed. Example output looks like:

```[{"file":"compiled-false.rb","compiled":"false"},{"file":"compiled-none.rb","compiled":"none"},{"file":"compiled-true.rb","compiled":"true"}]```

And pretty-printed:

```
[
  {
    "file": "compiled-false.rb",
    "compiled": "false"
  },
  {
    "file": "compiled-none.rb",
    "compiled": "none"
  },
  {
    "file": "compiled-true.rb",
    "compiled": "true"
  }
]
```

### Motivation

Collecting stats on the number of files that are compiled.

### Test plan

See included automated tests.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -70,6 +70,7 @@ const vector<PrintOptions> print_options({
     {"autogen-autoloader", &Printers::AutogenAutoloader, true, false},
     {"autogen-subclasses", &Printers::AutogenSubclasses, true},
     {"package-tree", &Printers::Packager},
+    {"compiled-files-json", &Printers::CompiledFilesJson},
 });
 
 PrinterConfig::PrinterConfig() : state(make_shared<GuardedState>()){};
@@ -135,6 +136,7 @@ vector<reference_wrapper<PrinterConfig>> Printers::printers() {
         AutogenAutoloader,
         AutogenSubclasses,
         Packager,
+        CompiledFilesJson,
     });
 }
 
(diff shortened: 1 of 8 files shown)


froydnj:
Can you explain the rationale behind doing it as a `-p` option, rather than specifying a filename to print the json to?

elliottt:
> Can you explain the rationale behind doing it as a `-p` option, rather than specifying a filename to print the json to?

It followed well with existing printing options like symbol-table-json. Since PrintOptions also support writing to a file, I'd be happy to add that support as well.

elliottt:
> > Can you explain the rationale behind doing it as a `-p` option, rather than specifying a filename to print the json to?
> 
> It followed well with existing printing options like symbol-table-json. Since PrintOptions also support writing to a file, I'd be happy to add that support as well.

@froydnj I'm not sure how I missed this, but the printing options support this already: `--print compiled-files-json:out.json` will write the output to `out.json`.

froydnj:
> > > Can you explain the rationale behind doing it as a `-p` option, rather than specifying a filename to print the json to?
> > 
> > 
> > It followed well with existing printing options like symbol-table-json. Since PrintOptions also support writing to a file, I'd be happy to add that support as well.
> 
> @froydnj I'm not sure how I missed this, but the printing options support this already: `--print compiled-files-json:out.json` will write the output to `out.json`.

TIL!

froydnj (approved):

PR #4629, opened 2021-09-21
redo format_cxx.sh to accept a list of files to format

froydnj:
### Motivation

I wanted to be able to run `./tools/scripts/remote-script ./tools/script/format_cxx.sh`, but there's no `.git` on a Stripe devbox. The next best thing is providing a list of files to format. (It might be even better to setup editor auto-format on save, but I am a curmudgeon.)

(A very similar change landed when we had a separate compiler repo, but the change did not get carried over from the compiler repo.)

### Test plan

See included automated tests.


elliottt (approved):
Looks good!

PR #4631, opened 2021-09-21
Package state in GlobalState

ngroman:
### Test plan

See included automated tests.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -92,6 +92,10 @@ public:
         return packagePathPrefixes;
     }
 
+    bool exists() const {
+        return name.mangledName.exists();
+    }
+
     // The possible path prefixes associated with files in the package, including path separator at end.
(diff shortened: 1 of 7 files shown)


packager/packager.cc
```diff
@@ -81,7 +82,24 @@ struct Import {
     Import(PackageName &&name, ImportType type) : name(std::move(name)), type(type) {}
 };
 
-struct PackageInfo {
+class PackageInfoImpl final : public sorbet::core::packages::PackageInfo {
```
jez, line 85: ooc, why is the implementation a subclass of an abstract base class if (from what i can see) there's only one implementation?

ngroman: @jvilk's suggestion wanted to keep the details of the package representation (FQN, PackageName, ImportType, etc.) out of core. Similar to WorkerPoolImpl.

PR #4632, opened 2021-09-22
Return richer data to `inline_type_error_handler`

jeffcarbs:
Enable `inline_type_error_handler` to take a second argument which has more data about the type error.

### Motivation

We ran into an issue trying to programmatically customize the handling of inline type errors. The argument passed to `inline_type_error_handler` is a simple `TypeError` whose message is [potentially truncated](https://github.com/sorbet/sorbet/blob/d9591b81594d4d096cfbe16cc469492ebc16db2c/gems/sorbet-runtime/lib/types/types/base.rb#L123) which makes it difficult or impossible to programmatically operate on the given value or expected type.

By contrast, the second argument passed to `call_validation_error_handler` is an object with a lot of detail: https://github.com/sorbet/sorbet/blob/061809c551ff62c546a8f9181a27b30906c8b89d/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb#L191-L200

This PR updates `inline_type_error_handler` to take a second argument that's a hash with the failed value and (optionally) the expected type and caller location. We check the arity of the configured proc to keep backwards compatibility for existing handlers.

### Test plan

Will add some tests if this is something you're open to merging.

### DIFF
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -214,7 +214,7 @@ module T
     begin
       raise TypeError.new("Passed `nil` into T.must")
     rescue TypeError => e # raise into rescue to ensure e.backtrace is populated
-      T::Configuration.inline_type_error_handler(e)
+      T::Configuration.inline_type_error_handler(e, kind: 'T.must', value: arg, type: nil)
     end
   end
 
@@ -239,7 +239,7 @@ module T
     begin
       raise TypeError.new(msg)
     rescue TypeError => e # raise into rescue to ensure e.backtrace is populated
-      T::Configuration.inline_type_error_handler(e)
+      T::Configuration.inline_type_error_handler(e, kind: 'T.absurd', value: value, type: nil)
     end
   end
 
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -214,7 +214,7 @@ module T
     begin
       raise TypeError.new("Passed `nil` into T.must")
     rescue TypeError => e # raise into rescue to ensure e.backtrace is populated
-      T::Configuration.inline_type_error_handler(e, kind: 'T.must', value: arg, type: nil)
+      T::Configuration.inline_type_error_handler(e, {kind: 'T.must', value: arg, type: nil})
     end
   end
 
@@ -239,7 +239,7 @@ module T
     begin
       raise TypeError.new(msg)
     rescue TypeError => e # raise into rescue to ensure e.backtrace is populated
-      T::Configuration.inline_type_error_handler(e, kind: 'T.absurd', value: value, type: nil)
+      T::Configuration.inline_type_error_handler(e, {kind: 'T.absurd', value: value, type: nil})
     end
   end
 
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -152,7 +152,12 @@ module T::Configuration
 
   def self.inline_type_error_handler(error, opts={})
     if @inline_type_error_handler
-      @inline_type_error_handler.call(error, opts)
+      # Backwards compatibility before `inline_type_error_handler` took a second arg
+      if @inline_type_error_handler.arity == 1
+        @inline_type_error_handler.call(error)
+      else
+        @inline_type_error_handler.call(error, opts)
+      end
     else
       inline_type_error_handler_default(error, opts)
     end
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -138,7 +138,7 @@ module T::Configuration
   # @option opts [Object] :value Actual param/return value
   #
   # @example
-  #   T::Configuration.inline_type_error_handler = lambda do |error|
+  #   T::Configuration.inline_type_error_handler = lambda do |error, opts|
   #     puts error.message
   #   end
   def self.inline_type_error_handler=(value)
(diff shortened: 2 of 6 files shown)


gems/sorbet-runtime/lib/types/private/methods/inline_type_validation.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+# typed: false
+
+module T::Private::Methods::InlineTypeValidation
+  def self.report_error(error, value, type: nil, location: nil)
+    # Backwards compatibility before `inline_type_error_handler` took a second arg
+    if T::Configuration.inline_type_error_handler.arity == 1
+      return T::Configuration.inline_type_error_handler(e)
+    end
```
froydnj, line 9: This doesn't seem quite right: this expression would be testing the arity of `T::Configuration.inline_type_error_handler` itself or maybe the value returned by a zero-argument send to `T::Configuration.inline_type_error_handler` (?).  But what you want to be checking is the arity of the thing that would be called by `T::Configuration.inline_type_error_handler` (`@inline_type_error_handler` or `inline_type_error_handler_default`)...right?

I guess you could just dispense with this helper module and move the arity logic into `T::Configuration.inline_type_error_handler`?  That would change the public interface of `T::Configuration`, which you might have been trying to avoid.  I'm not sure what the right thing to do here is.

froydnj (changes requested):
I'm in favor of richer error handling interfaces.  As you mentioned, this would need tests, and it would need documentation updates in `gems/sorbet-runtime/lib/types/configuration.rb`.  Depending on how you handle the comment below, `rbi/sorbet/t.rbi` would also need updates.

gems/sorbet-runtime/lib/types/private/methods/inline_type_validation.rb
jeffcarbs: Ah I saw the `@inline_type_error_handler` instance variable and assumed it was an `attr_accessor`. Moving this handling into `T::Configuration.inline_type_error_handler` definitely seems like the right move. I'll update this and add some tests.

jeffcarbs:
@froydnj - This should be ready for another review. I added tests and comments and updated the RBI and documentation.

One thing I was wondering is if we should also make the arity check for a splat (arity == -1). I could imagine a consumer doing something like:
```ruby
T::Configuration.inline_type_error_handler = lambda { |*args| some_method(*args) }
```
where `some_method` is assuming an arity of 1 and would break. I think this is probably fairly unlikely given the docs suggest 
`T::Configuration.inline_type_error_handler = lambda do |error| ... end` so I could probably go either way on that.

rbi/sorbet/t.rbi
```diff
@@ -151,7 +151,7 @@ module T::Configuration
   def self.disable_vm_prop_serde; end
   def self.hard_assert_handler(str, extra); end
   def self.hard_assert_handler=(value); end
-  def self.inline_type_error_handler(error); end
+  def self.inline_type_error_handler(error, opts); end
```
jez, line 154: if there is a default for `opts` in the source, there should be a default in the RBI as well

```suggestion
  def self.inline_type_error_handler(error, opts={}); end
```

gems/sorbet-runtime/lib/types/configuration.rb
```diff
@@ -131,6 +131,11 @@ def self.default_checked_level=(default_checked_level)
...
   #
   # @param [TypeError] error TypeError that was raised
+  # @param [Hash] opts A hash containing contextual information on the error:
+  # @option opts [String] :kind One of:
+  #   ['T.cast', 'T.let', 'T.bind', 'T.assert_type!', 'T.must', 'T.absurd']
+  # @option opts [Object, nil] :type Expected param/return value type
+  # @option opts [Object] :value Actual param/return value
   #
   # @example
   #   T::Configuration.inline_type_error_handler = lambda do |error|
```
jez, line 141: did you want to update this example to mention / use `opts`?

gems/sorbet-runtime/lib/types/_types.rb
```diff
@@ -239,7 +239,7 @@ def self.absurd(value)
     begin
       raise TypeError.new(msg)
     rescue TypeError => e # raise into rescue to ensure e.backtrace is populated
-      T::Configuration.inline_type_error_handler(e)
+      T::Configuration.inline_type_error_handler(e, kind: 'T.absurd', value: value, type: nil)
```
jez, line 242: these kinds of things will cause warnings in future ruby versions. can you wrap the keyword args in `{ ... }`? (or alternatively, declare `inline_type_error_handler` as taking keyword `**opts`, but I prefer making there be a hash literal at the call site)

jeffcarbs: Update to use a hash 👍

gems/sorbet-runtime/lib/types/configuration.rb
jeffcarbs: I did, thanks for catching!

rbi/sorbet/t.rbi
jeffcarbs: I did this to keep this consistent with the RBI for `call_validation_error_handler` and `sig_validation_error_handler`, which have an optional `opts` param in the code but not in the RBI since I wasn't sure if that was intentional. I updated the RBI for all three to be `opts={}` to match their runtime args.

jeffcarbs:
@jez @froydnj - Made those updates, ready for another review (but no rush), thanks!

rbi/sorbet/t.rbi
jez: yep not intentional, thanks for fixing!

jez:
lgtm, thanks

✅ http://go/builds/bui_KHXC8DQRjEyhR7

jez (approved):

PR #4633, opened 2021-09-22
rename .cfg.exp tests to .cfg-text.exp for better diffs

froydnj:
### Motivation

I was working on some CFG changes and when I had to regenerate the diffs for `*.cfg.exp` tests, they were basically unreadable.  This wasn't helpful for me and it's probably not helpful for the eventual reviewer.

We added a `cfg-text` format a while back that's more readable (if you're not using `dot`), and this PR proposes moving all of the existing `.cfg.exp` tests to use `.cfg-text.exp` instead for better diffing and readability.

The commits on their own are probably not particularly readable, but the commit messages basically describe what was done.

### Test plan

See included automated tests.


jez (changes requested):
From a reading of cfg-text, the information contained in cfg-text is not as good as cfg:

- cfg-text does not show the the `elseb` branch for a basic block, only the `thenb`
- cfg records whether `firstDeadInstruction` is 0 or not, cfg-text does not

Can you scour the implementation to be absolutely, 100% sure that we're not missing any other information in the two representations? It's fine to add ad-hoc textual representations that we use for one-off debugging, but any new output must contain at least as much information as the old information before we can replace existing tests with it.

PR #4634, opened 2021-09-22
add explicit load argument instructions for blocks

froydnj:
### Motivation

Block arguments are currently represented using a combination of the `LoadYieldParams` cfg instruction and sends to the result of said instruction to retrieve the arguments.  This makes a certain amount of sense: the block arguments are essentially represented as an array under the hood, and there are special rules about arrays being passed as the first block arg in some cases that destructure the array into the other arguments.  This arrangement also cleverly enables reuse of some of the typing machinery: assuming block arguments are represented as tuple types (note that [keyword arguments are documented as unsupported](https://sorbet.org/docs/procs)), we can lean on the typing for `Tuple#[]` to match up block arguments to their types.

But this representation is not really suitable for the compiler: the compiler has to specially recognize the sends that create block arguments.  Splat arguments don't use the send mechanism: more cleverness in the CFG construction represents splat arguments as the type of the result of `LoadYieldParams` (if the splat arg is the first arg to the block) or an alias for untyped otherwise.  (Again, keyword arguments are not supported in blocks.)  The compiler therefore doesn't recognize splat args in blocks, which is a problem for compiling a lot of real-world code.  (Finding splat args if they are the first arg seems doable, since presumably the only assignment *from* the result of `LoadYieldParams` is the splat arg; finding splat args otherwise seems doomed to heuristics that could fail.)  And the compiler also has to know during code generation that some sends are skippable.

It seems better to me to have an explicit representation for block arguments in the CFG rather than piggy-backing on sends, at least for the purposes of the compiler.  This PR does just that, adding a `YieldLoadArg` instruction for representing the loading of block arguments.

I will admit this scheme is slightly less nice for the inference part of Sorbet: we have to duplicate some of the logic to handle typing of `Send` nodes when we're determining types of `YieldLoadArg` nodes...and it's not totally clear to me whether some of the LSP logic should be duplicated as well.  (Testing in CI suggests that we do need to duplicate some of the LSP logic, or at least pay more careful attention to our location information.)

But I think the balance is that things are more understandable; the compiler changes are particularly nice, and it was very pleasing to have the adding of explicit representations for splat args in `YieldLoadArg` ("put real instructions in the CFG for splat args in blocks") magically make compiler testcases start working without any compiler-side modifications.  The new representation is slightly smaller, though I don't think space consumption of the CFG itself is a major concern for Sorbet's inference phase.

### Test plan

See included automated tests.


infer/environment.cc
```diff
@@ -1172,6 +1172,65 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
...
+                                                numPosArgs,
+                                                args,
+                                                recvType.type,
+                                                recvType,
+                                                recvType.type,
+                                                block,
+                                                core::Loc(ctx.file, bind.loc),
+                                                isPrivateOk,
+                                                suppressErrors};
+                auto dispatched = recvType.type.dispatchCall(ctx, dispatchArgs);
```
elliottt, line 1225: I think that simulating a call to `[]` makes sense here, but what would you think about pulling this out into a separate function in this file so that we could give the setup and dispatch a more informative name?

elliottt (approved):
I think this makes sense! Only non-blocking comments.

froydnj:

→ https://go/builds/bui_KHQhYFRm9rrIMc
→ https://go/builds/bui_KHQhxS2T2nWLNC

infer/environment.cc
froydnj: I'm not a fan of all the code here (clang-format also not doing the code any favors by choosing the tallest layout for everything), but it at least has some precedent in `calls.cc`.  I'm inclined to leave this as-is.

PR #4635, opened 2021-09-23
Allow fundamental types to be passed by value to ConcurrentQueue::push

elliottt:
There are a number of places in the codebase where we use `ConcurrentQueue<int>` and other similar instantiations. The current behavior of `ConcurrentQueue::push` is that the element argument must be an rvalue reference, which yields loops that look like: https://github.com/sorbet/sorbet/blob/4a5cbc15d277cbc1cf7f5c3c6fa55ccb86c04cf2/hashing/hashing.cc#L76-L79

While requiring that non-fundamental types are moved into the queue is good, it's not as nice to have to move `int` elements in. This PR adds a little bit of template hackery to the argument to `push`, requiring that it's passed by value when it's a fundamental type and an rvalue reference otherwise.

### Motivation

Less surprising use of `ConcurrentQueue::push`.

### Test plan

See included automated tests.

### DIFF
--- a/common/concurrency/ConcurrentQueue.h
+++ b/common/concurrency/ConcurrentQueue.h
@@ -38,7 +38,7 @@ public:
     AbstractConcurrentBoundedQueue(const AbstractConcurrentBoundedQueue &other) = delete;
     AbstractConcurrentBoundedQueue(AbstractConcurrentBoundedQueue &&other) = delete;
 
-    // When `Elem` is a fundamental type (int, bool, etc) push takes a value, but if it's anythign else it expects an
+    // When `Elem` is a fundamental type (int, bool, etc) push takes a value, but if it's anything else it expects an
     // rvalue reference so that we don't forget to move the argument.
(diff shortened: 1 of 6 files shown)


common/concurrency/ConcurrentQueue.h
```diff
@@ -37,7 +38,11 @@ template <class Elem, class Queue> class AbstractConcurrentBoundedQueue {
     AbstractConcurrentBoundedQueue(const AbstractConcurrentBoundedQueue &other) = delete;
     AbstractConcurrentBoundedQueue(AbstractConcurrentBoundedQueue &&other) = delete;
 
-    inline void push(Elem &&elem, int count) noexcept {
+    // When `Elem` is a fundamental type (int, bool, etc) push takes a value, but if it's anythign else it expects an
```
froydnj, line 41:
```suggestion
    // When `Elem` is a fundamental type (int, bool, etc) push takes a value, but if it's anything else it expects an
```

froydnj (approved):

PR #4636, opened 2021-09-23
Fix desugaring to call-with-splat-and-block

aprocter-stripe:
This fixes a bug in desugaring sends to `call-with-splat-and-block`. The trouble seemed to be that, when looking for the kwarg hash to deconstruct it, we assumed it would always be in the last position, which isn't true when a final `&blk` is present. Since that check fails, the kwarg hash winds up getting treated like an ordinary positional argument.

Example:

```ruby
def f(&blk)
  g(1, 2, *[3,4,5], 6, **{j: 99}, &blk)
end
```

Output of `-p desugar-tree` before change (note that the `begin...end` is inside the args to `concat` and the penultimate arg to call-with-splat-and-block is `nil`):

```
  def f<<todo method>>(&blk)
    ::<Magic>.<call-with-splat-and-block>(<self>, :g, [1, 2].concat(::<Magic>.<splat>([3, 4, 5])).concat([6, begin
            <hashTemp>$2 = ::<Magic>.<to-hash-dup>({:j => 99})
            <hashTemp>$2
          end]), nil, blk)
  end
end
```

Output of `-p desugar-tree` after change (the `begin...end` is now the penultimate arg to call-with-splat-and-block):

```
  def f<<todo method>>(&blk)
    ::<Magic>.<call-with-splat-and-block>(<self>, :g, [1, 2].concat(::<Magic>.<splat>([3, 4, 5])).concat([6]), [begin
          <hashTemp>$2 = ::<Magic>.<to-hash-dup>({:j => 99})
          <hashTemp>$2
        end], blk)
  end
end
```

### Motivation
I'm working on a separate change to support forwarding of splat and block args to "zsuper" in the compiler, which works by synthesizing `call-with-splat-and-block` calls, and as I was throwing things at the desugarer to double check my understanding of the semantics I came across this. Wanted to make sure it worked when synthesized as intended by the desugarer, before moving on to synthesizing it in novel contexts.

### Test plan
I think the change to `forward_args.rb.desugar-tree.exp` implies that this change is covered well by existing tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -659,7 +659,7 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                         send->args.pop_back();
                     }
 
-                    // Deconstruct the kwargs hash in if it's present.
+                    // Deconstruct the kwargs hash if it's present.
                     if (!send->args.empty()) {
                         if (auto *hash = parser::cast_node<parser::Hash>(send->args.back().get())) {
                             if (hash->kwargs) {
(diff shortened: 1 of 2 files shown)


elliottt (approved):
Great find and fix!

froydnj-stripe (approved):

PR #4640, opened 2021-09-25
Correctly set proc arity

elliottt:
The sorbet compiler currently builds procs with `rb_vm_ifunc_proc_new`, which will always indicate that the proc expects unlimited arguments. This breaks a few internal cases where tests fail if a proc takes arguments at all.

This PR starts tracking proc arity for `FunctionType::Block` blocks, and passes that information to `sorbet_i_send` to use in the event that a proc will be allocated. It also adds `sorbet_rb_iterate` which is a version of `rb_iterate` that expects a non-null block and its arity information.

### Motivation

More well-behaving `Proc#arity`.

### Test plan

See included automated tests.


elliottt:

→ https://go/builds/bui_KJS3EDGfqUHbxX
→ https://go/builds/bui_KJS3VBh74Oo0Rr

compiler/IREmitter/IREmitterContext.h
```diff
@@ -82,6 +82,11 @@ constexpr bool functionTypeNeedsPostprocessing(FunctionType ty) {
 
 struct Alias;
 
+struct BlockArity {
+    int min = 0;
+    int max = 0;
+};
```
froydnj, line 88: :heart: for using a real structure and not `std::pair<int, int>`

test/testdata/compiler/proc_arity.rb
```diff
@@ -24,6 +24,15 @@ def test_case(name, p)
...
 test_case("2 required, 2 defaulted", lambda {|a, b, c=nil, d=nil|})
+test_case("2 required, 2 defaulted, break", lambda {|a, b, c=nil, d=nil| break})
+
+# keywords
+test_case("1 required", lambda {|a:|})
+test_case("2 required", lambda {|a:, b:|})
+test_case("1 optional", lambda {|a: 10|})
+test_case("2 optional", lambda {|a: 10, b: 20|})
+test_case("1 required, 1 optional", lambda {|a:, b: 20|})
+test_case("2 required, 1 optional", lambda {|a:, b:, c: 30|})
```
froydnj, line 35: Have testcases with positional and/or optional args + required and/or optional kwargs as well?

test/testdata/compiler/proc_arity.rb
```diff
@@ -32,3 +41,4 @@ def test_case(name, p)
 test_case("1 required, 1 defaulted, repeated", lambda {|a, b=10, *c|})
 test_case("2 required, 1 defaulted, repeated", lambda {|a, b, c=10, *d|})
 test_case("2 required, 2 defaulted, repeated", lambda {|a, b, c=10, d=nil, *e|})
+test_case("2 required, 2 defaulted, repeated, break", lambda {|a, b, c=10, d=nil, *e| break})
```
froydnj, line 44: Comment in the same spirit as above: we should try to have some required/defaulted kwargs in here too?

froydnj (approved):
This all makes sense to me, though I can't help feeling that our argument passing in some places is getting unwieldy.

test/testdata/compiler/proc_arity.rb
elliottt: Good suggestion, this exposed some problems!

PR #4642, opened 2021-09-25
Populate the isBlock flag in CFG construction

elliottt:
We weren't setting the `isBlock` flag for the argument info during CFG construction. This turns out to be required for correctly supporting `Proc#arity`.

### Motivation

Suporting the implementation of `Proc#arity`.

### Test plan

See included automated tests.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -426,6 +426,7 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                         target.isRepeated = e.flags.isRepeated;
                         target.isDefault = e.flags.isDefault;
                         target.isShadow = e.flags.isShadow;
+                        target.isBlock = e.flags.isBlock;
                     }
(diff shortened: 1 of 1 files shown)


cfg/builder/builder_walk.cc
```diff
@@ -426,6 +426,7 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                         target.isRepeated = e.flags.isRepeated;
                         target.isDefault = e.flags.isDefault;
                         target.isShadow = e.flags.isShadow;
+                        target.isBlock = e.flags.isBlock;
```
froydnj, line 429: Should we just make this whole thing `argFlags.emplace_back(e.flags)`?

froydnj (approved):
LGTM.  I think the below is a good cleanup, but your call.

cfg/builder/builder_walk.cc
elliottt: I like that, thanks!

PR #4643, opened 2021-09-25
delete outdated explicit instantiations of unique_ptr on trees

froydnj:
### Motivation

The comment here says "don't delete", but this dates back to when the AST was managed via `unique_ptr` rather than our custom `ExpressionPtr`.  Seeing as how we no longer use `unique_ptr` in our AST, these explicit instantiations can be deleted.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -5,37 +5,6 @@
 #include <sstream>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


elliottt (approved):
🔥

PR #4645, opened 2021-09-27
enforce that we don't make VM calls to Sorbet::Private::Static methods

froydnj:
### Motivation

This is a followon PR to #4591, where we rejected VM calls to `<Magic>` methods.  We wanted to do the same for calls on `Sorbet::Private::Static`, but we couldn't, because our current compilation strategy turns a call like:

```ruby
  Sorbet::Private::Static.keep_def(self, :foo, :normal)
```

into (pseudocode)

```
c = get_constant("Sorbet::Private::Static")
recv = get_constant("Sorbet::Private::Static")
if is_class_of(recv, c)
  ...do the fast thing...
else
  ...do a call into the Ruby VM...
end
```

The slow path gets optimized out by a combination of LLVM and clever declarations on things in the codegen payload.  But this strategy is not great for ensuring that we never make calls into the VM for `Sorbet::Private::Static` methods.  It also generates a bunch of LLVM IR that is just dead code, creating more work for the optimizers, etc.

This PR provides a separate function on `SymbolBasedIntrinsicMethod` for making *compile-time* decisions about whether we need a fast path test, and modifies our symbol-based intrinsic implementation strategy to use it.  As the comments note, this function should not normally be overridden, but there are several cases where it comes in very handy, like this PR.  (I have a WIP patch to provide type assumptions for the result of `Thread.current`, which makes the generated code for `Thread.current[:foo]` much nicer.)

Once we have such a function, the only things to do are return `true` from it where appropriate and to modify `sends.cc` to take advantage of the function.  I looked at integrating the logic for the single-intrinsic, skippable fast path with the current logic and decided it was too gnarly, so we have a little bit of duplication.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/SymbolBasedIntrinsicMethod.h
+++ b/compiler/IREmitter/SymbolBasedIntrinsicMethod.h
@@ -30,6 +30,16 @@ public:
     // Ruby VM.
     virtual llvm::Value *receiverFastPathTest(MethodCallContext &mcctx, core::ClassOrModuleRef potentialClass) const;
 
+    // The above is a runtime test, since it returns an llvm::Value (which might be `true`).
+    // This method is a compile test for instances where we statically know the compile-time
+    // test would succeed.  This might be done for intrinsics which we know don't have
+    // a slow path through the VM (e.g. calls on Sorbet::Private::Static) or where having
+    // a slow path through the VM would hinder optimizations like having an assumption
+    // about the return type of the method.
+    //
+    // You should not typically be returning true from this.
+    virtual bool skipFastPathTest(MethodCallContext &mcctx, core::ClassOrModuleRef potentialClass) const;
+
     SymbolBasedIntrinsicMethod(Intrinsics::HandleBlock blockHandled) : blockHandled(blockHandled){};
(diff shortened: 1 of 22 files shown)


compiler/IREmitter/SymbolBasedIntrinsicMethod.h
```diff
@@ -30,6 +30,16 @@ class SymbolBasedIntrinsicMethod {
...
 
+    // The above is a runtime test, since it returns an llvm::Value (which might be `true`).
+    // This method is a compile test for instances where we statically know the compile-time
+    // test would succeed.  This might be done for intrinsics which we know don't have
+    // a slow path through the VM (e.g. calls on Sorbet::Private::Static) or where having
+    // a slow path through the VM would hinder optimizations like having an assumption
+    // about the return type of the method.
+    //
+    // You should not typically be returning true from this.
+    virtual bool skipFastPathTest(MethodCallContext &mcctx, core::ClassOrModuleRef potentialClass) const;
```
jez, line 41: nice!

jez (approved):

PR #4646, opened 2021-09-27
--compiled-files-json -> --file-sigils-json

jez:
### Motivation

We might want to put more stuff in this eventually; I figure we may as
well rename it earlier rather than later.

(For example, `typed:`, `frozen_string_literal`, autoloader preamble, etc.)

### Test plan

n/a


froydnj (approved):
What's the rationale for keeping the old option?

froydnj:
> What's the rationale for keeping the old option?

Oh, you had a comment for this, which needed scrolling on mobile. 🤣

PR #4649, opened 2021-09-28
optimize Thread.current and Thread.current[:constant_sym]

froydnj:
### Motivation

This PR builds on #4645 to change things so that `Thread.current` doesn't implement a slow path and always directly calls `rb_thread_current`.  The current code would eventually do the same, but at the cost of having a slow path that would be optimized out by LLVM.  This change means that we can meaningfully assert that `Thread.current` returns a `Thread`, which in turns means that `Thread.current[:foo]` doesn't do type checks anymore.

But wait, there's more!  Our implementation of `Thread#[]` checks at runtime for ID-ness of its argument.  But if we have `Thread.current[:foo]`, then we know at compile time that our argument is guaranteed to be convertible to an ID.  So we can check for a constant symbol argument to `Thread#[]` (resp. `Thread#[]=`) and eliminate some runtime checking.  We also need to make `rb_id2sym` `const` so that the symbol for `:foo`, which is now unused, never actually gets materialized at runtime.

The benefits can be see in `test/testdata/ruby_benchmarks/stripe/lspace.rb`.  Before:

```
froydnj@pyro:~/src/sorbet$ ./tools/scripts/run_benchmarks.sh -b test/testdata/ruby_benchmark/stripe/while_10_000_000.rb -f test/testdata/ruby_benchmark/stripe/lspace.rb
...
stripe/while_10_000_000.rb      .230    .101
stripe/lspace.rb        .897    .602
stripe/lspace.rb - baseline     .667    .501
```

After:

```
froydnj@pyro:~/src/sorbet$ ./tools/scripts/run_benchmarks.sh -b test/testdata/ruby_benchmark/stripe/while_10_000_000.rb -f test/testdata/ruby_benchmark/stripe/lspace.rb
...
stripe/while_10_000_000.rb      .224    .099
stripe/lspace.rb        .880    .549
stripe/lspace.rb - baseline     .656    .450
```

Roughly 10% faster than before.

This could probably be improved further by combining interrupt checks in a single basic block; for `Thread.current[:foo]` we'll wind up with a basic block that looks like:

```
current = sorbet_Thread_current(...)
sorbet_afterIntrinsic()
val = sorbet_Thread_square_br_symarg(current, id_for_foo)
sorbet_afterIntrinsic()
```

and I think we could get away with combining the first `sorbet_afterIntrinsic` with the second, so long as we make sure to check for interrupts somewhere.  But that is a separate patch.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/SymbolBasedIntrinsics.cc
+++ b/compiler/IREmitter/SymbolBasedIntrinsics.cc
@@ -772,6 +772,25 @@ public:
                          string cMethodWithBlock)
         : CallCMethod(rubyClass, rubyMethod, cMethod, cMethodWithBlock){};
 
+    // It is safe to skip the test if the receiver is a constant of the given class.
+    virtual bool skipFastPathTest(MethodCallContext &mcctx, core::ClassOrModuleRef potentialClass) const override {
+        auto &cs = mcctx.cs;
+        auto &irctx = mcctx.irctx;
+        auto &recv = mcctx.send->recv;
+
+        auto aliasit = irctx.aliases.find(recv.variable);
+        if (aliasit == irctx.aliases.end()) {
+            return false;
+        }
+
+        if (aliasit->second.kind != Alias::AliasKind::Constant) {
+            return false;
+        }
+
+        ENFORCE(potentialClass.data(cs)->isSingletonClass(cs));
+        auto attachedClass = potentialClass.data(cs)->attachedClass(cs);
+        return aliasit->second.constantSym == attachedClass;
+    }
     virtual InlinedVector<core::ClassOrModuleRef, 2> applicableClasses(const core::GlobalState &gs) const override {
         return {rubyClass.data(gs)->lookupSingletonClass(gs)};
     };
--- a/compiler/IREmitter/SymbolBasedIntrinsics.cc
+++ b/compiler/IREmitter/SymbolBasedIntrinsics.cc
@@ -790,6 +790,34 @@ public:
     }
 } Thread_squareBrackets;
 
+class Thread_squareBracketsEq : public CallCMethod {
+public:
+    Thread_squareBracketsEq() : CallCMethod(core::Symbols::Thread(), "[]="sv, CMethod{"sorbet_Thread_square_br_eq"}) {}
+
+    virtual llvm::Value *makeCall(MethodCallContext &mcctx) const override {
+        auto &cs = mcctx.cs;
+        auto &builder = mcctx.builder;
+        auto &irctx = mcctx.irctx;
+        auto *send = mcctx.send;
+
+        if (send->args.size() != 2 || send->numPosArgs != 2) {
+            return CallCMethod::makeCall(mcctx);
+        }
+
+        auto &arg = send->args[0];
+        auto symit = irctx.symbols.find(arg.variable);
+        if (symit == irctx.symbols.end()) {
+            return CallCMethod::makeCall(mcctx);
+        }
+
+        auto *recv = mcctx.varGetRecv();
+        auto *id = Payload::idIntern(cs, builder, symit->second);
+        return builder.CreateCall(cs.getFunction("sorbet_Thread_square_br_eq_symarg"),
+                                  {recv, id, Payload::varGet(cs, send->args[1].variable, builder, irctx,
+                                                             mcctx.rubyBlockId)});
+    }
+} Thread_squareBracketsEq;
+
 class CallCMethodSingleton : public CallCMethod {
 public:
     CallCMethodSingleton(core::ClassOrModuleRef rubyClass, string_view rubyMethod, CMethod cMethod)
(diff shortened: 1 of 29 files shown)


froydnj:

→ https://go/builds/bui_KJKT2BTffmjASL
→ https://go/builds/bui_KJKTgvGlORrG8B

compiler/Passes/Lowerings.cc
```diff
@@ -278,13 +278,21 @@ class TypeTest : public IRIntrinsic {
...
+    {"sorbet_i_isa_Symbol", "sorbet_isa_Symbol"},
+    {"sorbet_i_isa_Float", "sorbet_isa_Float"},
+    {"sorbet_i_isa_Untyped", "sorbet_isa_Untyped"},
+    {"sorbet_i_isa_Hash", "sorbet_isa_Hash"},
+    {"sorbet_i_isa_Array", "sorbet_isa_Array"},
+    {"sorbet_i_isa_Regexp", "sorbet_isa_Regexp"},
+    {"sorbet_i_isa_String", "sorbet_isa_String"},
+    {"sorbet_i_isa_Proc", "sorbet_isa_Proc"},
+    {"sorbet_i_isa_Thread", "sorbet_isa_Thread"},
+    {"sorbet_i_isa_RootSingleton", "sorbet_isa_RootSingleton"},
```
jez, line 295: ty clang-format 🙃

compiler/IREmitter/SymbolBasedIntrinsics.cc
```diff
@@ -756,18 +756,99 @@ class TPrivateCompiler_compilerVersion : public SymbolBasedIntrinsicMethod {
     };
 } TPrivateCompiler_compilerVersion;
 
+class Thread_squareBrackets : public CallCMethod {
```
jez, line 759: This is the first time we're subclassing `CallCMethod`. I guess that's probably fine?

(I'm generally not a fan of subclassing for the purpose of code reuse, in favor of helper functions, but this seems ~fine I guess.)

compiler/IREmitter/SymbolBasedIntrinsics.cc
```diff
@@ -756,18 +756,99 @@ class TPrivateCompiler_compilerVersion : public SymbolBasedIntrinsicMethod {
...
+public:
+    Thread_squareBrackets() : CallCMethod(core::Symbols::Thread(), "[]"sv, CMethod{"sorbet_Thread_square_br"}) {}
```
jez, line 761:
```suggestion
    // This sorbet_Thread_square_br is a slower version that will do arity checking and convert the
    // argument to a symbol. If our `makeCall` fast path doesn't apply, we'll fall back to calling
    // the slow version (via super class `makeCall`).
    Thread_squareBrackets() : CallCMethod(core::Symbols::Thread(), "[]"sv, CMethod{"sorbet_Thread_square_br"}) {}
```

compiler/IREmitter/SymbolBasedIntrinsics.cc
```diff
@@ -756,18 +756,99 @@ class TPrivateCompiler_compilerVersion : public SymbolBasedIntrinsicMethod {
...
+
+        auto *recv = mcctx.varGetRecv();
+        auto *id = Payload::idIntern(cs, builder, symit->second);
+        return builder.CreateCall(cs.getFunction("sorbet_Thread_square_br_symarg"), {recv, id});
+    }
+} Thread_squareBrackets;
+
+class Thread_squareBracketsEq : public CallCMethod {
+public:
+    Thread_squareBracketsEq() : CallCMethod(core::Symbols::Thread(), "[]="sv, CMethod{"sorbet_Thread_square_br_eq"}) {}
```
jez, line 787:
```suggestion
    // This sorbet_Thread_square_br_eq is a slower version that will do arity checking and convert the
    // argument to a symbol. If our `makeCall` fast path doesn't apply, we'll fall back to calling
    // the slow version (via super class).
    Thread_squareBracketsEq() : CallCMethod(core::Symbols::Thread(), "[]="sv, CMethod{"sorbet_Thread_square_br_eq"}) {}
```

compiler/IREmitter/SymbolBasedIntrinsics.cc
```diff
@@ -756,18 +756,99 @@ class TPrivateCompiler_compilerVersion : public SymbolBasedIntrinsicMethod {
...
-    CallCMethodSingleton(core::ClassOrModuleRef rubyClass, string_view rubyMethod, string cMethod)
+    CallCMethodSingleton(core::ClassOrModuleRef rubyClass, string_view rubyMethod, CMethod cMethod)
         : CallCMethod(rubyClass, rubyMethod, cMethod){};
 
-    CallCMethodSingleton(core::ClassOrModuleRef rubyClass, string_view rubyMethod, string cMethod,
+    CallCMethodSingleton(core::ClassOrModuleRef rubyClass, string_view rubyMethod, CMethod cMethod,
                          string cMethodWithBlock)
         : CallCMethod(rubyClass, rubyMethod, cMethod, cMethodWithBlock){};
 
+    // It is safe to skip the test if the receiver is a constant of the given class.
```
jez, line 822: This presumes the method is not stubbed, but `¯\_(ツ)_/¯` that's probably fine?

jez (approved):
Nice!

compiler/IREmitter/SymbolBasedIntrinsics.cc
froydnj: It's worth saying explicitly that this is not a behavior change from the previous code; we'd optimize `Thread.current` down to `rb_thread_current()`, we'd just take a more roundabout way of doing it and we wouldn't get the same quality of information in the optimizer.

PR #4650, opened 2021-09-28
Clean up YARD cruft

dduugg:
### Motivation

This cleans up the warnings when invoking `yard` on the `sorbet-runtime` gem. I don't think the gem uses `yard` in any official capacity, but it retains a lot of cruft. Some examples:

- `@override` is not a valid tag
- `@!method` is used needlessly/incorrectly
- A number of links (`{}`) aren't valid (and date to when `Struct` was contained in `Chalk::ODM`)
- `@param`/`@option` tags are used to document non-params, leading to warnings and incorrect documentation
- Some `param` names no longer match
- etc.

(I still find it useful to generate and browse the docs. For instance, a lot of class names are tricky to map back to file names.)

### Test plan

I browsed the resulting documentation. There are still formatting issues but it's significantly better.

This PR only touches comments, it has no impact on live code.

### DIFF
--- a/gems/sorbet-runtime/.yardopts
+++ b/gems/sorbet-runtime/.yardopts
@@ -1,0 +1,1 @@
+--markup=markdown
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -82,8 +82,9 @@ module T
 
   # Constructs a type alias. Used to create a short name for a larger type. In Ruby this returns a
   # wrapper that contains a proc that is evaluated to get the underlying type. This syntax however
-  # is needed for support by the static checker. Example usage:
+  # is needed for support by the static checker.
   #
+  # @example
   #  NilableString = T.type_alias {T.nilable(String)}
   #
   #  sig {params(arg: NilableString, default: String).returns(String)}
@@ -106,8 +107,9 @@ module T
   # References a type parameter which was previously defined with
   # `type_parameters`.
   #
-  # This is used for generic methods. Example usage:
+  # This is used for generic methods.
   #
+  # @example
   #  sig
   #  .type_parameters(:U)
   #  .params(
@@ -185,7 +187,7 @@ module T
   # cycle. However, we also don't actually need to do so; An untyped
   # identity method works just as well here.
   #
-  # sig {params(value: T.untyped).returns(T.untyped)}
+  # `sig {params(value: T.untyped).returns(T.untyped)}`
   def self.unsafe(value)
     value
   end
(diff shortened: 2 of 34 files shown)


gems/sorbet-runtime/.yardopts
```diff
@@ -0,0 +1 @@
+--markup=markdown
```
dduugg, line 1: This mostly just means that backticks are used instead of `+` to delimit code

gems/sorbet-runtime/lib/types/_types.rb
```diff
@@ -185,7 +187,7 @@ def self.assert_type!(value, type, checked: true)
   # cycle. However, we also don't actually need to do so; An untyped
   # identity method works just as well here.
   #
-  # sig {params(value: T.untyped).returns(T.untyped)}
+  # `sig {params(value: T.untyped).returns(T.untyped)}`
```
dduugg, line 190: This renders better, and signals that the `{}` aren't links

gems/sorbet-runtime/lib/types/configuration.rb
```diff
@@ -130,7 +130,7 @@ def self.default_checked_level=(default_checked_level)
   #
   # Parameters passed to value.call:
   #
-  # @param [TypeError] error TypeError that was raised
+  #  @param [TypeError] error TypeError that was raised
```
dduugg, line 133: The extra space seems to be enough to prevent `yard` from trying to attach the tag to a method param (but the resulting formatting is terrible)

gems/sorbet-runtime/lib/types/enum.rb
```diff
@@ -47,7 +47,7 @@ class T::Enum
   SerializedVal = T.type_alias {T.untyped}
   private_constant :SerializedVal
 
-  ## Enum class methods ##
+  ### Enum class methods ###
```
dduugg, line 50: The double `#` seemed to have some [special meaning](https://github.com/lsegal/yard/blob/main/lib/yard/parser/ruby/legacy/statement_list.rb#L256) to `yard`, leading to a warning 🤷

jez (approved):
You have some conflicts.

> (I still find it useful to generate and browse the docs. For instance, a lot of class names are tricky to map back to file names.)

What command do you use to do this? Is there some enforcing yard mode we could add to CI?

dduugg:
> You have some conflicts.
> 
> > (I still find it useful to generate and browse the docs. For instance, a lot of class names are tricky to map back to file names.)
> 
> What command do you use to do this? Is there some enforcing yard mode we could add to CI?

Conflicts resolved. I use `yard doc --plugin sorbet` in the `sorbet-runtime` directory. There's a `fail-on-warning` option if you wanted to enforce something in CI.

Note that's it's pretty easy to set up a GH [action](https://github.com/dduugg/yard-sorbet/blob/main/.github/workflows/publish-to-github-pages.yml) to publish the resulting documentation to GH pages. You can also set it up in a separate repo, as [Homebrew](https://github.com/Homebrew/rubydoc.brew.sh) does.

PR #4652, opened 2021-09-28
Enforce that errors from pipeline tests must have locations

aadi-stripe:
Currently the pipeline tests do nothing if they encounter errors without valid locations, resulting in tests missing issues sometimes.

### Motivation

This limitation/bug unfortunately missed a fundamental issue with Rigorous Packaging design, in the packager.cc pass.

Adding an allow-list/exemption list for tests that currently fail due to location-less errors. These will be fixed incrementally

### Test plan

See included automated tests.

cc @jez @ngroman @elliottt

### DIFF
--- a/test/pipeline_test_runner.cc
+++ b/test/pipeline_test_runner.cc
@@ -86,6 +86,19 @@ UnorderedSet<string> knownExpectations = {"parse-tree",       "parse-tree-json",
                                           "document-symbols", "package-tree",     "document-formatting-rubyfmt",
                                           "autocorrects"};
 
+// TODO: This list will be progressively made empty, and we will not have any tests that
+// can swallow errors that don't have association locations. See lines 481-484 below.
+UnorderedSet<string> locationCheckExemptTests = {
+    "test/testdata/packager/nested_inner_namespaces",
+    "test/testdata/packager/import_subpackage",
+    "test/testdata/packager/nested_packages",
+    "test/testdata/packager/deeply_nested_packages",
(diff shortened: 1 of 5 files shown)


test/pipeline_test_runner.cc
```diff
@@ -86,6 +86,16 @@ UnorderedSet<string> knownExpectations = {"parse-tree",       "parse-tree-json",
                                           "document-symbols", "package-tree",     "document-formatting-rubyfmt",
                                           "autocorrects"};
 
+// TODO: This list will be progressively made empty, and we will not have any tests that
+// can swallow errors that don't have associated locations. See lines 477-481 below.
```
jez, line 90:
```suggestion
// This list will be progressively made empty, and we will not have any tests that can
// swallow errors that don't have associated locations. See lines 477-481 below.
// DO NOT add new tests here.
// TODO(jez) Fix these tests
```

test/testdata/packager/import_subpackage/a/__package.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
aadi-stripe, line 1: Another packager test -- which has this issue. Figure I'd leave it here for extra coverage; eventually we can isolate this error.

test/pipeline_test_runner.cc
aadi-stripe: Fixed

jez (approved):

Issue #1068, opened 2019-06-26
`super` is untyped


mvidner:
### Problem

I have a library class that the users are supposed to specialize. It has tricky initialization arguments, so I want to make sure the subclass is calling the base class's `initialize` with the correct types.

I wrote a `sig` for Base#initialize. Yet, type errors are not found when Derived#initialize calls `super` with wrong types.

Example: ([srb.run link](https://sorbet.run/#%23%20typed%3A%20strict%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams(x%3A%20String).void%7D%0A%20%20def%20bar(x)%0A%20%20%20%20puts%20x%0A%20%20end%0Aend%0A%0Aclass%20B%20%3C%20A%0A%20%20sig%20%7Bparams(x%3A%20Symbol).returns(String)%7D%0A%20%20def%20bar(x)%0A%20%20%20%20super%20%23%20Type%20mismatch%20for%20params%20AND%20return%20(AST%3A%20zsuper)%0A%20%20%20%20super(1%2C%202)%20%23%20Type%20mismatch%20for%20params%20AND%20return%20(AST%3A%20super)%0A%20%20end%0Aend%0A))

```ruby
# typed: strict
class A
  extend T::Sig

  sig {params(x: String).void}
  def bar(x)
    puts x
  end
end

class B < A
  sig {params(x: Symbol).returns(String)}
  def bar(x)
    super # Type mismatch for params AND return (AST: zsuper)
    super(1, 2) # Type mismatch for params AND return (AST: super)
  end
end
```

In the example I am using a regular method instead of initialize, to show that both the parameters and the return value are untyped.

(For my reference, the actual base class is [CFA::BaseModel](https://github.com/config-files-api/config_files_api/blob/af0127113f767b0f26ddcd43eb6551d1db393004/lib/cfa/base_model.rb#L16-L25) and one derived class is [CFA::Grub2::DeviceMap](https://github.com/config-files-api/config_files_api_grub2/blob/702fa0037f1c809e7e7debedace91eec5703736f/lib/cfa/grub2/device_map.rb#L24-L25))

### Proposed solution

`super` is untyped. I can't see any workarounds to make it typed. Please make it typed.

I have grepped the sorbet repo for `super` and found too many matches, so I have no idea how complex this is to fix.

mvidner:
[...]

PR #4653, opened 2021-09-28
Support zero-argument 'super' with splats and blocks

aprocter-stripe:
The existing lowering of `ZSuperArgs` in `local_vars.cc` silently drops splat arguments (e.g. `*args`) and block arguments (e.g. `&blk`) when forwarding the functions arguments. This updates `local_vars.cc` to synthesize calls to `<call-with-splat>`, `<call-with-block>`, or `<call-with-splat-and-block>`, as needed, and updates those intrinsics to treat `super` appropriately.

Commit-by-commit review is recommended.

### Motivation
Compile more programs!

### Test plan
See included automated tests.


local_vars/local_vars.cc
```diff
@@ -226,35 +232,246 @@ class LocalNameInserter {
...
+        }
+
+        auto method = ast::MK::Literal(
+            original.loc, core::make_type<core::LiteralType>(core::Symbols::Symbol(), core::Names::super()));
+
+        if (posArgsArray) {
+            ast::Send::ARGS_store sendargs;
+            // We wrap self with T.unsafe in order to get around the requirement for <call-with-splat> and
+            // <call-with-splat-and-block> that the shapes of the splatted hashes be known statically. This is a bit of
+            // a hack, but 'super' is currently treated as untyped anyway.
```
aprocter-stripe, line 373: Wanted to call this out in a comment to make sure it doesn't get buried. I think this is a reasonable thing to do, but open to other suggestions.

aprocter-stripe:

→ https://go/builds/bui_KJS0WNMgZXI1vf
→ https://go/builds/bui_KJS0trOc1G1ASV

local_vars/local_vars.cc
```diff
@@ -226,35 +232,246 @@ class LocalNameInserter {
...
+        }
+
+        // If there were any keyword args after a keyword splat, fold them into the splatted hash.
+        if (kwArgsHash && !kwArgKeyEntries.empty()) {
+            kwArgsHash =
+                ast::MK::Send1(original.loc, std::move(kwArgsHash), core::Names::merge(),
+                               ast::MK::Hash(original.loc, std::move(kwArgKeyEntries), std::move(kwArgValueEntries)));
+            kwArgKeyEntries.clear();
+            kwArgValueEntries.clear();
+        }
```
aprocter-stripe, line 364: Thinking about this a bit more, this whole "if" is actually a can't-happen, and we even `ENFORCE` as much as we're walking the args. Will delete this.

local_vars/local_vars.cc
```diff
@@ -226,35 +232,246 @@ class LocalNameInserter {
...
+            } else if (arg.flags.isKeywordSplat()) {
+                ENFORCE(!kwArgsHash, "Saw multiple keyword splats");
+
+                // TODO(aprocter): is it necessary to duplicate the hash here?
+                kwArgsHash =
+                    ast::MK::Send1(original.loc, ast::MK::Constant(original.loc, core::Symbols::Magic()),
+                                   core::Names::toHashDup(), ast::make_expression<ast::Local>(original.loc, arg.arg));
+                if (!kwArgKeyEntries.empty()) {
+                    // TODO(aprocter): it might make more sense to replace this with an InsSeq that calls
+                    // <Magic>::<merge-hash>, which is what's done in the desugarer.
```
jez, line 329: What's the plan for these two `TODO(aprocter)`'s? did you want to address those in this PR?

aprocter-stripe: I think they're okay for now, but they might not be the best choice perf-wise. Happy to address before merge, open a ticket, what-have-ya.

local_vars/local_vars.cc
jez: Can we detect if we'd need to do this? For example, some cases where we don't need to do this:

- The method only has positional args
- (i think) The method doesn't have any `*args` or `**kwargs`

Basically, write some code to detect, based on the for loop above, whether something that is not a shape or tuple will have been constructed.

If we ever get around to fixing #1068 for classes (not modules), it will be nice to not have to remember to change this code.

local_vars/local_vars.cc
```diff
@@ -226,35 +232,246 @@ class LocalNameInserter {
...
+                ast::MK::Send1(original.loc, std::move(kwArgsHash), core::Names::merge(),
+                               ast::MK::Hash(original.loc, std::move(kwArgKeyEntries), std::move(kwArgValueEntries)));
+            kwArgKeyEntries.clear();
+            kwArgValueEntries.clear();
+        }
+
+        auto method = ast::MK::Literal(
+            original.loc, core::make_type<core::LiteralType>(core::Symbols::Symbol(), core::Names::super()));
+
+        if (posArgsArray) {
```
jez, line 369: I have a slight stylistic preference for always writing `!= nullptr` (helps readability IMO) but this is not blocking.

local_vars/local_vars.cc
```diff
@@ -226,35 +232,246 @@ class LocalNameInserter {
     }
 
     ast::ExpressionPtr postTransformSend(core::MutableContext ctx, ast::ExpressionPtr tree) {
+        // Replace ZSuperArgs with an explicit forwarding of the enclosing method's arguments.
+
         auto &original = ast::cast_tree_nonnull<ast::Send>(tree);
-        if (original.args.size() == 1 && ast::isa_tree<ast::ZSuperArgs>(original.args[0])) {
```
jez, line 230: You _might_ want to pre-emptively move this `super` logic into a helper function. I don't think that we do anything other than super stuff in `postTransformSend` right now, but I imagine the first person who needs to touch this code after your change lands for something not super-related will appreciate having already had things teased apart.

jez (approved):

local_vars/local_vars.cc
aprocter-stripe: I _think_ that as it stands this will only `unsafe` exactly when we need to in order to placate `call-with-splat{,-and-block}`. (We'll only generate these if `*args` is present.) Will add a comment to that effect.

An alternative (probably best in a separate PR?) would be what's mentioned in this comment (basically, make `call-with-splat` and `call-with-splat-and-block` aware that `super` is special, which we may need to do for non-`zsuper`-synthesized things anyway): https://github.com/sorbet/sorbet/pull/4654#pullrequestreview-766089106 . We could then delurk the unsafe here. That still leaves us with work to find when fixing #1068, but maybe not in such a weird place.

PR #4654, opened 2021-09-29
Add a super-splat test

elliottt:
`super` and the `call-with-splat` intrinsic aren't interacting well, and the
result is that compiler-generated code will end up trying to call a method with
the name `super` instead of invoking super when splat args are given.

### Motivation

More tests.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/compiler/disabled/super_splat.rb
+++ b/test/testdata/compiler/disabled/super_splat.rb
@@ -1,0 +1,24 @@
+# compiled: true
+# typed: true
+# frozen_string_literal: true
+
+class Parent
+  extend T::Sig
+
+  def foo(*args)
+    puts args
+  end
+
+end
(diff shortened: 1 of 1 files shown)


aprocter-stripe (approved):
Test lgtm!

It might make sense to fold a fix for this into #4653. Currently that PR is working around a similar issue when lowering zero-arg `super` to `call-with-splat` et al by `T.unsafe`ing the receiver. A better fix, which I think would also fix this test case, might be to special-case `super` inside `call-with-splat`. That would mirror what we do in the "normal" case where splats aren't involved. (See https://github.com/sorbet/sorbet/blob/f6f4ec1a7dcf01e12f270e82c81260dd9fc3c415/core/types/calls.cc#L567-L569)

PR #4657, opened 2021-09-29
Make `requires_ancestor` take a block

Morriar:
### Motivation

Change `requires_ancestor` syntax from:

```rb
module M
  extend T::Helpers

  requires_ancestor RA
end
```

to:

```rb
module M
  extend T::Helpers

  requires_ancestor { RA }
end
```

As to avoid eager loading constants.

Also add an autocorrect:

```
autocorrect-requires-ancestor-block.rb:12: `requires_ancestor` only accepts a block https://srb.help/5062
    12 |  requires_ancestor C
          ^^^^^^^^^^^^^^^^^^^
  Autocorrect: Done
    autocorrect-requires-ancestor-block.rb:12: Replaced with `requires_ancestor { C }`
    12 |  requires_ancestor C
          ^^^^^^^^^^^^^^^^^^^
```

### Test plan

See included automated tests.

cc @vinistock.

### DIFF
--- a/rbi/sorbet/t.rbi
+++ b/rbi/sorbet/t.rbi
@@ -91,23 +91,25 @@ module T::Helpers
   sig {void}
   def sealed!; end
 
-  # We do not use signatures on the following methods as to not duplicate errors
+  # We do not use signatures on `mixes_in_class_methods` as to not duplicate errors
   # between the `namer` phase and typechecking for calls.
   #
   # With the following example:
   #
   # ```
   # class A
-  #   requires_ancestor "A"
+  #   mixes_in_class_methods A
   # end
   # ```
   #
   # Using a signature would mean we would generate two errors:
-  # 1. error: `requires_ancestor` must only contain constant literals (from namer)
+  # 1. error: `mixes_in_class_methods` must only contain constant literals (from namer)
   # 2. error: Expected `Module` but found `NilClass` for argument `mod` (from calls)
 
   def mixes_in_class_methods(mod, *mods); end
-  def requires_ancestor(mod, *mods); end
+
+  sig { params(block: T.proc.returns(Module)).void }
+  def requires_ancestor(&block); end
 end
 
 module T::Array
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -661,31 +661,53 @@ private:
             return;
         }
 
-        if (send->args.empty()) {
-            return; // The arity mismatch error will be emitted later by infer.
+        auto *block = ast::cast_tree<ast::Block>(send->block);
+
+        if (!send->args.empty()) {
+            if (auto e = gs.beginError(loc, core::errors::Resolver::InvalidRequiredAncestor)) {
+                e.setHeader("`{}` only accepts a block", send->fun.show(gs));
+
+                string replacement = "";
+                int indent = core::Loc::offset2Pos(todo.file.data(gs), send->loc.beginPos()).column - 1;
+                int index = 1;
+                for (auto &arg : send->args) {
+                    auto argLoc = core::Loc(todo.file, arg.loc());
+                    replacement += fmt::format("{:{}}{} {{ {} }}{}", "", index == 1 ? 0 : indent, send->fun.show(gs),
+                                               argLoc.source(gs).value(), index < send->args.size() ? "\n" : "");
+                    index += 1;
+                }
+                e.addAutocorrect(
+                    core::AutocorrectSuggestion{fmt::format("Replace `{}` with `{}`", send->fun.show(gs), replacement),
+                                                {core::AutocorrectSuggestion::Edit{loc, replacement}}});
+            }
+            return;
         }
 
-        for (auto &arg : send->args) {
-            auto argLoc = core::Loc(todo.file, arg.loc());
-            auto *id = ast::cast_tree<ast::ConstantLit>(arg);
-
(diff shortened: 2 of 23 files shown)


rbi/sorbet/t.rbi
```diff
@@ -91,23 +91,25 @@ module T::Helpers
   sig {void}
   def sealed!; end
 
-  # We do not use signatures on the following methods as to not duplicate errors
+  # We do not use signatures on `mixes_in_class_methods` as to not duplicate errors
```
Morriar, line 94: This comment only applies to `mixes_in_class_methods` now.

resolver/resolver.cc
```diff
@@ -661,31 +661,63 @@ class ResolveConstantsWalk {
...
+                for (auto &arg : send->args) {
+                    auto argLoc = core::Loc(todo.file, arg.loc());
+                    string argString = argLoc.source(gs).value();
+                    if (index > 0) {
+                        auto locDetail = core::Loc::offset2Pos(todo.file.data(gs), send->loc.beginPos());
+                        for (int i = 1; i < locDetail.column; i++) {
+                            replacement += " ";
+                        }
+                    }
+                    replacement += send->fun.show(gs) + " { " + argString + " }";
```
Morriar, line 681: Is there a way to use `fmt` with a string that contains literal curly braces `{ ... }`?

resolver/resolver.cc
```diff
@@ -661,31 +661,63 @@ class ResolveConstantsWalk {
...
+
+                string replacement = "";
+                int index = 0;
+                for (auto &arg : send->args) {
+                    auto argLoc = core::Loc(todo.file, arg.loc());
+                    string argString = argLoc.source(gs).value();
+                    if (index > 0) {
+                        auto locDetail = core::Loc::offset2Pos(todo.file.data(gs), send->loc.beginPos());
+                        for (int i = 1; i < locDetail.column; i++) {
+                            replacement += " ";
```
Morriar, line 678: If the call had many arguments such as `requires_ancestor A, B`, we need to introduce a new line and adjust the indentation.

resolver/resolver.cc
paracycle: From https://fmt.dev/latest/syntax.html :
> If you need to include a brace character in the literal text, it can be escaped by doubling: {{ and }}

resolver/resolver.cc
```diff
@@ -661,31 +661,63 @@ class ResolveConstantsWalk {
...
 
-            if (id == nullptr || !id->symbol.exists() || !id->symbol.data(gs)->isClassOrModule()) {
-                if (auto e = gs.beginError(argLoc, core::errors::Resolver::InvalidRequiredAncestor)) {
-                    e.setHeader("Argument to `{}` must be statically resolvable to a class or a module",
-                                send->fun.show(gs));
+                    auto argLoc = core::Loc(todo.file, arg.loc());
+                    string argString = argLoc.source(gs).value();
+                    string ln = index < send->args.size() - 1 ? "\n" : "";
+
+                    replacement += fmt::format("{}{} {{ {} }}{}", indent, send->fun.show(gs), argString, ln);
```
Morriar, line 685: If the call had many arguments such as `requires_ancestor A, B`, we need to introduce a new line and adjust the indentation.

test/cli/autocorrect-requires-ancestor-block/autocorrect-requires-ancestor-block.rb
```diff
@@ -0,0 +1,14 @@
...
+class C; end
+
+class A
+  extend T::Helpers
+
+  abstract!
+
+  requires_ancestor C
+  requires_ancestor RA1, RA2
+end
```
elliottt, line 14: This program will raise errors even after the autocorrects have been applied, would it be worth making `A` a module instead?

elliottt (changes requested):
This looks good, thank you for the autocorrect!

Just one question about the new autocorrect test, and a recommendation for how to simplify the indenting logic.

resolver/resolver.cc
```diff
@@ -661,31 +661,63 @@ class ResolveConstantsWalk {
...
+                string replacement = "";
+                int index = 0;
+                for (auto &arg : send->args) {
+                    string indent = "";
+                    if (index > 0) {
+                        auto locDetail = core::Loc::offset2Pos(todo.file.data(gs), send->loc.beginPos());
+                        for (int i = 1; i < locDetail.column; i++) {
+                            indent += " ";
+                        }
+                    }
```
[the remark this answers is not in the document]
Morriar: TIL, thanks 👍

test/cli/autocorrect-requires-ancestor-block/autocorrect-requires-ancestor-block.rb
Morriar: Switched to a module 👍

elliottt (changes requested):
This looks great, thank you for the changes! One last request before we merge it: could you add a test case for `requires_ancestor(Foo) { Bar }`?

Morriar:
> could you add a test case for `requires_ancestor(Foo) { Bar }`?

Good catch, I disabled the autocorrect if a block is already present and added a few tests to show this new behaviour 👍

elliottt (approved):
This looks good, thank you for adding that extra case!

PR #4660, opened 2021-09-30
Fix Pathname#cleanpath signature

FnControlOption:
### Motivation

`cleanpath` always returns a `Pathname`

https://github.com/ruby/ruby/blob/master/ext/pathname/lib/pathname.rb

### DIFF
--- a/rbi/stdlib/pathname.rbi
+++ b/rbi/stdlib/pathname.rbi
@@ -526,7 +526,7 @@ class Pathname < Object
     params(
         consider_symlink: T::Boolean,
     )
-    .returns(T.untyped)
+    .returns(Pathname)
   end
   def cleanpath(consider_symlink=T.unsafe(nil)); end
 


jez:
Nice, thanks! From reading the source, it looks like it actually always returns `self.class.new`, which will respect whether the class has been subclassed or not. In particular:

```
irb(main):003:0> class MyPathname < Pathname; end
=> nil
irb(main):004:0> pn = MyPathname.new("/usr/bin/ruby")
=> #<MyPathname:/usr/bin/ruby>
irb(main):005:0> pn.cleanpath
=> #<MyPathname:/usr/bin/ruby>
```

if you wanted, you could change the return type from `Pathname` to `T.self_type` to be super precise, but leaving it as is works as well.

jez:

→ https://go/builds/bui_KJrusUKTyaZICZ
→ https://go/builds/bui_KJru869GzhRwrt

jez:
This change looks good. It introduces one new error on Stripe's codebase (thanks!) and I've opened a PR internally to fix the type error. Once that fix has landed, I'll merge this change.

jez (approved):

PR #4661, opened 2021-09-30
pass the call data into sorbet method functions 

froydnj:
### Motivation

We are going to need the call data to write improved parsing for kwargs.  The appropriate patch landed in sorbet/ruby#3; this PR updates sorbet's patches for Ruby and makes the appropriate compiler changes to make everything else line up.

Note that we're passing the call data as a `void *`; whatever we do on the parsing side, we'll need to cast it to `struct rb_kwarg_call_data *` before using it.

### Test plan

See included automated tests.


third_party/ruby/sorbet_ruby_2_7.patch
```diff
@@ -2667,3 +2667,333 @@ index 79a6cf014ec85..8596e416d7b7a 100644
...
+ typedef struct rb_method_sorbet_struct {
+     /* cf. rb_method_cfunc_struct, but we only support one argument style */
+diff --git a/vm_eval.c b/vm_eval.c
+index 52805d317f02..26d3d8906f3f 100644
+--- a/vm_eval.c
++++ b/vm_eval.c
+@@ -144,7 +144,7 @@ vm_call0_sorbet_with_frame(rb_execution_context_t* ec, struct rb_calling_info *c
+                                                     0, reg_cfp->sp, sorbet->iseqptr->body->local_table_size, sorbet->iseqptr->body->stack_max);
+ 
+         /* TODO: eventually we want to pass cd in here to assist with kwargs parsing */
```
elliottt, line 2981: Does this TODO still make sense?

third_party/ruby/sorbet_ruby_2_7.patch
```diff
@@ -2667,3 +2667,333 @@ index 79a6cf014ec85..8596e416d7b7a 100644
...
+ 	CHECK_CFP_CONSISTENCY("vm_call0_sorbet_with_frame");
+ 	rb_vm_pop_frame(ec);
+diff --git a/vm_insnhelper.c b/vm_insnhelper.c
+index a891794d0a08..05171ad04ca7 100644
+--- a/vm_insnhelper.c
++++ b/vm_insnhelper.c
+@@ -2622,7 +2622,7 @@ vm_call_sorbet_with_frame_normal(rb_execution_context_t *ec, rb_control_frame_t
+ 
+     reg_cfp->sp -= argc + 1;
+     /* TODO: eventually we want to pass cd in here to assist with kwargs parsing */
```
elliottt, line 2994: Same, should we remove these TODOs?

elliottt (approved):
Looks good to me, just a few non-blocking questions.

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -1986,15 +1986,16 @@ VALUE sorbet_callFuncDirect(struct FunctionInlineCache *cache, rb_sorbet_func_t
...
     sorbet_popFrame();
     return res;
 }
 
 SORBET_INLINE
 VALUE sorbet_callStaticInitDirect(rb_sorbet_func_t methodPtr, int argc, VALUE *argv, VALUE recv) {
     rb_control_frame_t *cfp = sorbet_pushStaticInitFrame(recv);
-    VALUE res = methodPtr(argc, argv, recv, cfp);
+    // XXX we don't really have a call data here, but we shouldn't be parsing any kwargs either.
+    VALUE res = methodPtr(argc, argv, recv, cfp, NULL);
```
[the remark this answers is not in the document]
froydnj: There is not, but not that this is calling static init methods, not final methods.  We'd just have to synthesize a call data that said we had no arguments and no keyword arguments.

third_party/ruby/sorbet_ruby_2_7.patch
froydnj: Yeah, we should eventually remove these.  I will take a note to do that.

PR #4663, opened 2021-09-30
add a very early global DCE pass to send less stuff through LLVM

froydnj:
### Motivation

We actually have quite a bit of stuff in the codegen payload that might not actually be referenced by most of the files that we compile.  Deleting it *ought* to provide some kind of speedup.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_KK9sHyzfip9hmg
→ https://go/builds/bui_KK9sdXFVTqZSR2

PR #4664, opened 2021-09-30
Change the suffix for the llvm emitted at different stages

elliottt:
The sorbet compiler will output llvm at different stages of the pipeline, currently as `.ll`, `.lll`, and `.llo` files. These suffixes are a little clunky, so this PR replaces `.lll` with `.lowered.ll` and `.llo` with `.opt.ll`. These suffixes give a more clear description of what phase of the pipeline the output comes from, and doesn't break filetype detection based on extensions in text editors.

### Motivation

Cleaning up the compiler outputs.

### Test plan

Workflows tested:
* [x] updating exp files with `tools/scripts/update_compiler_exp.sh`
* [x] syncing compiled files back from my devbox with `test/run_compiled.sh`


froydnj (approved):
:shipit:

PR #4668, opened 2021-10-01
Packager: export_for_test

ngroman:
For `--stripe-packages` mode only:
Changes how `Test` code accesses its corresponding normal package. It no longer automatically gets an alias set for all of the normal package's code because this caused name redefinition errors (this bug was tested for but masked by https://github.com/sorbet/sorbet/pull/4652).

We instead introduce an explicit `export_for_test` DSL method that adds an assignment into the `<PackageTest>::<Mangled Pkg Name>` module much like any other import.

* `export_for_test` is prohibited from exporting anything in `Test::*` itself
* Test code also automatically gets access to anything that is `exported` publicly by its package.
* Removes special `<PackageRegistry>::<Managled Pkg Name>::<Magic> placeholder since this is no longer need. (Exports are explicit instead of implicit)

### Motivation
Fix redifinition bug in `--stripe-packages`

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/package-test-simple/main_lib/__package.rb
+++ b/test/cli/package-test-simple/main_lib/__package.rb
@@ -5,4 +5,5 @@
 class Project::MainLib < PackageSpec
   import Project::Util
   test_import Project::TestOnly
+  export_for_test Project::MainLib::Lib
 end
--- a/test/pipeline_test_runner.cc
+++ b/test/pipeline_test_runner.cc
@@ -87,17 +87,6 @@ UnorderedSet<string> knownExpectations = {"parse-tree",       "parse-tree-json",
                                           "document-symbols", "package-tree",     "document-formatting-rubyfmt",
                                           "autocorrects"};
 
-// This list will be progressively made empty, and we will not have any tests that
-// can swallow errors that don't have associated locations. See lines 479-483 below.
-// DO NOT add new tests here.
-// TODO(aadi-stripe) Fix these tests
-UnorderedSet<string> locationCheckExemptTests = {
-    "test/testdata/packager/nested_inner_namespaces",
-    "test/testdata/packager/import_subpackage",
-    "test/testdata/packager/nested_packages",
-    "test/testdata/packager/deeply_nested_packages",
-};
-
 ast::ParsedFile testSerialize(core::GlobalState &gs, ast::ParsedFile expr) {
     auto &savedFile = expr.file.data(gs);
     auto saved = core::serialize::Serializer::storeTree(savedFile, expr);
@@ -477,9 +466,6 @@ TEST_CASE("PerPhaseTest") { // NOLINT
                 continue;
             }
             auto diag = errorToDiagnostic(*gs, *error);
-            if (diag == nullptr && locationCheckExemptTests.contains(singleTest)) {
-                continue;
-            }
             ENFORCE(diag != nullptr, "Error was given no valid location - '{}'", error->toString(*gs));
 
             auto path = error->loc.file().data(*gs).path();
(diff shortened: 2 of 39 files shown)


ngroman:
cc @aadi-stripe

ngroman:
Happy to bikeshed on the `export_for_test` it's not my favorite, but I ran out of better ideas.

aadi-stripe:
> Happy to bikeshed on the `export_for_test` it's not my favorite, but I ran out of better ideas.

Sure! We can settle on that later tbh, this unblocks forward progress for now.

test/cli/package-test-simple/main_lib/__package.rb
```diff
@@ -5,4 +5,5 @@
 class Project::MainLib < PackageSpec
   import Project::Util
   test_import Project::TestOnly
+  export_for_test Project::MainLib::Lib
```
elliottt, line 8: If we're still open to bikeshedding the name, what about `test_export` for symmetry with `test_import`?

test/pipeline_test_runner.cc
```diff
@@ -87,17 +87,6 @@ UnorderedSet<string> knownExpectations = {"parse-tree",       "parse-tree-json",
...
-// can swallow errors that don't have associated locations. See lines 479-483 below.
-// DO NOT add new tests here.
-// TODO(aadi-stripe) Fix these tests
-UnorderedSet<string> locationCheckExemptTests = {
-    "test/testdata/packager/nested_inner_namespaces",
-    "test/testdata/packager/import_subpackage",
-    "test/testdata/packager/nested_packages",
-    "test/testdata/packager/deeply_nested_packages",
-};
-
```
elliottt, line 100: 🎉

test/pipeline_test_runner.cc
```diff
@@ -477,9 +466,6 @@ TEST_CASE("PerPhaseTest") { // NOLINT
                 continue;
             }
             auto diag = errorToDiagnostic(*gs, *error);
-            if (diag == nullptr && locationCheckExemptTests.contains(singleTest)) {
-                continue;
-            }
```
elliottt, line 482: 🥇

test/testdata/packager/export_imported/pass.package-tree.exp
```diff
@@ -9,17 +9,12 @@ class <emptyTree><<C <root>>> < (::<todo sym>)
...
     end
-
-    class <emptyTree>::<C A_Package$1>::<C A>::<C <Magic>><<C <todo sym>>> < ()
-    end
   end
 
   module <emptyTree>::<C <PackageTests>><<C <todo sym>>> < ()
     module <emptyTree>::<C A_Package$1>::<C B><<C <todo sym>>> < ()
-      <emptyTree>::<C BClass> = <emptyTree>::<C <PackageRegistry>>::<C B_Package$1>::<C B>::<C BClass>
+      <emptyTree>::<C BClass> = <emptyTree>::<C <PackageRegistry>>::<C A_Package$1>::<C B>::<C BClass>
```
elliottt, line 16: This is a little bit odd looking as the name comes from `A_Package` now, but I think the explanation is that it's using the re-exported version of the name instead of the original from `B_Package`?

elliottt (approved):
I think that this all makes sense.

I had two comments: changing the name to `export_for_test` is definitely not blocking, but if you could confirm that I'm right about the exp file changes I'd appreciate it 👍

test/testdata/packager/export_imported/pass.package-tree.exp
ngroman: Yeah it's a weird case that I think we may want to ban in the future, but the names do end up all resolving.

test/cli/package-test-simple/main_lib/__package.rb
aadi-stripe: The problem is that `test_export` isn't technically visible to tests in other packages. So it's not the opposite of `test_import`.

PR #4669, opened 2021-10-01
return string_views instead of strings from a couple methods

froydnj:
### Motivation

Noticed this while working on other things.  Not allocating unless you really need to seems like a good thing, even if this probably only happens in error paths.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Initializer.cc
+++ b/rewriter/Initializer.cc
@@ -79,7 +79,7 @@ void checkSigReturnType(core::MutableContext ctx, const ast::Send *send) {
             e.setHeader("The {} method should always return {}", "initialize", "void");
 
             auto loc = core::Loc(ctx.file, originalSend.loc());
-            string original = loc.source(ctx).value();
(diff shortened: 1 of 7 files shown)


rewriter/Initializer.cc
```diff
@@ -79,7 +79,7 @@ void checkSigReturnType(core::MutableContext ctx, const ast::Send *send) {
             e.setHeader("The {} method should always return {}", "initialize", "void");
 
             auto loc = core::Loc(ctx.file, originalSend.loc());
-            string original = loc.source(ctx).value();
+            auto original = (string)loc.source(ctx).value();
```
trevor-stripe, line 82:
```suggestion
            auto original = string(loc.source(ctx).value());
```
Why the c-style cast here?

trevor-stripe (approved):
Looks good. Just one suggestion about removing the c-style cast.

PR #4673, opened 2021-10-01
Don't prepend ./ to file paths' when using --dir=.

jez:
### Motivation

This makes some things easier in the Sorbet Compiler. It also looks better in
error messages.

### Test plan

See included automated tests.


common/common.cc
```diff
@@ -248,6 +248,8 @@ bool sorbet::FileOps::isFileIgnored(string_view basePath, string_view filePath,
     return false;
 }
 
+namespace {
+
```
elliottt, line 252: 🎉

jez:

→ https://go/builds/bui_KKZRmzef7ODrna
→ https://go/builds/bui_KKZSHZ7G4EVt10

common/common.cc
```diff
@@ -279,22 +282,30 @@ void appendFilesInDir(string_view basePath, const string &path, const sorbet::Un
             if (dotLocation != string::npos) {
                 string_view ext(fullPath.c_str() + dotLocation, fullPath.size() - dotLocation);
                 if (extensions.contains(ext)) {
-                    result.push_back(move(fullPath));
+                    // Remove the ./ prefix in the final path (but it's convenient to have it
+                    // above for the sake of matching ignore patterns). Also don't allocate unless we have to.
+                    if (basePath == ".") {
+                        result.push_back(fullPath.substr(2));
```
elliottt, line 288: What about using `fullPath.erase(0, 2)` instead of copying to make the substring?

packager/packager.cc
```diff
@@ -182,6 +182,11 @@ class PackageDB final {
             curPrefixPos = path.find_last_of('/', curPrefixPos - 1);
         }
 
+        const auto &root = packageInfoByPathPrefix.find("");
```
elliottt, line 185: This returns an iterator, I'm not sure it needs the `&`.
```suggestion
        const auto root = packageInfoByPathPrefix.find("");
```

elliottt (approved):
Looks good, just a few comments 👍

PR #4674, opened 2021-10-01
Replace file-sigils-json with file-table-json

jez:
### Motivation

We already more or less had all this functionality. No need to make a second
option.

This has the added benefit of using protobuf to generate the JSON according to a
schema instead of ad hoc with a JSON library.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_KKZoI67Ftlepl8
→ https://go/builds/bui_KKZo98789mmmYs

core/proto/proto.cc
```diff
@@ -342,8 +342,17 @@ com::stripe::rubytyper::File::StrictLevel strictToProto(core::StrictLevel strict
             return com::stripe::rubytyper::File::StrictLevel::File_StrictLevel_Autogenerated;
         case core::StrictLevel::Stdlib:
             return com::stripe::rubytyper::File::StrictLevel::File_StrictLevel_Stdlib;
-        default:
-            ENFORCE(false, "bad strict level: {}", (int)strict);
```
jez, line 346: not having this here will enable exhaustiveness checking

test/cli/file-sigils-json/file-sigils-json.out
```diff
@@ -1,9 +1,2784 @@
 ----- ./*.rb -----------------------------------------------------------
 No errors! Great job.
-[{"file":"./compiled-false.rb","compiled":"false"},{"file":"./compiled-none.rb","compiled":"none"},{"file":"./compiled-true.rb","compiled":"true"}]
+{
+ "files": [
+  {
+   "path": "https://github.com/sorbet/sorbet/tree/master/bazel-out/host/bin/rbi/procs.rbi",
+   "min_error_level": "Max"
```
jez, line 7: it's not clear to me why sigil and strict are missing here.

also, compiled is always missing.

elliottt (approved):
Looks good to me!

test/cli/file-sigils-json/file-sigils-json.out
jez: actually, it looks like zero values might be elided?

elliottt: Was including the rbis intentional? They look like new additions in this PR.

aprocter-stripe: > actually, it looks like zero values might be elided?

I think that must be it? `compiled-false.rb` and `compiled-true.rb` have it.

common/common.cc
```diff
@@ -285,7 +285,7 @@ void appendFilesInDir(string_view basePath, const string &path, const sorbet::Un
                     // Remove the ./ prefix in the final path (but it's convenient to have it
                     // above for the sake of matching ignore patterns). Also don't allocate unless we have to.
                     if (basePath == ".") {
-                        result.push_back(fullPath.substr(2));
+                        result.push_back(fullPath.erase(0, 2));
```
aprocter-stripe, line 288: ooc, what's motivating this change?

packager/packager.cc
```diff
@@ -174,15 +174,15 @@ class PackageDB final {
...
+            const auto it = packageInfoByPathPrefix.find(path.substr(0, curPrefixPos + 1));
             if (it != packageInfoByPathPrefix.end()) {
                 return it->second.get();
             }
 
             curPrefixPos = path.find_last_of('/', curPrefixPos - 1);
         }
 
-        const auto &root = packageInfoByPathPrefix.find("");
+        const auto root = packageInfoByPathPrefix.find("");
```
aprocter-stripe, line 185: again just ooc: why are we dropping the `&`s here?

aprocter-stripe (approved):

common/common.cc
elliottt: Avoiding allocations, `substr` will copy the string.

packager/packager.cc
elliottt: `find` returns an iterator value, so it's not clear to me that it's good to declare these as references to intermediates.

test/cli/file-sigils-json/file-sigils-json.out
jez: > Was including the rbis intentional? They look like new additions in this PR.

They are included because they are in the symbol table.

elliottt: Fair enough, do you know why they were missing before?

elliottt: Resolved in a DM: the `file-sigils-json` printer was omitting them explicitly.

PR #4681, opened 2021-10-05
fix C-style string casts around the codebase

froydnj:
### Motivation

This PR started life as a followup PR for review comments in #4669.  But then I figured, why not go the extra mile and delete `(string)` casts elsewhere in the codebase too?

I tried adding a `const std::string &pathAsString() const` method to `core::File`, but it resulted in some overload issues, so it was easier to keep the casting of `string(...path())` everywhere. :(

### Test plan

See included automated tests.


aprocter-stripe (approved):
🎉

PR #4682, opened 2021-10-05
pass calling info into sorbet method functions

froydnj:
### Motivation

This is basically #4661, only with `struct rb_calling_info *` instead of `struct rb_call_data *`.  The only wrinkle here is that we're not just appending it to the parameters of sorbet method functions, we're inserting it into the middle of the arguments, and we also have to synthesize a `rb_calling_info` when calling method functions directly (either via final methods or static-init methods).

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_KMIpsYryj4YhPO
→ https://go/builds/bui_KMIptAmFGaRdPp

PR #4684, opened 2021-10-05
Add bundler installer RBI and change Psych.load_file

vinistock:
Add missing definitions for `Bundler::Installer` in `bundler.rbi` and change the parameter for `Psych.load_file` to be `**kwargs`.

### Motivation

[Psych.load_file](https://github.com/ruby/psych/blob/a8922edb85abce2412588a0b2044cd343e1caa70/lib/psych.rb#L669) indeed takes `**kwargs` and accepts other options beyond `fallback`.

The [Bundler::Installer](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/installer.rb) class was missing from `bundler.rbi`. I put all signatures as `T.untyped` for now, but let me know if you know the right types for it.

### Test plan


froydnj:

→ https://go/builds/bui_KLzSGOhNs4YJsP
→ https://go/builds/bui_KLzS8IJUaadvtK

froydnj (approved):
Thanks!

PR #4687, opened 2021-10-05
use an existing variable in resolveJumpTarget

froydnj:
### Motivation

The code is much clearer when this variable is re-used.

### Test plan

See included automated tests.


elliottt (approved):
Much clearer, thank you!

PR #4689, opened 2021-10-05
Update LoadArg message on type test failure to match sorbet-runtime more closely

aprocter-stripe:
This updates the type test failure message from `LoadArg` to match `sorbet-runtime` a bit more closely, giving strings like `Parameter 'foo': Expected bar, got baz` instead of `sig: Expected bar, got baz`.

It's worth noting that this doesn't _quite_ get us to a perfect match for two reasons:

1. The error produced by `sorbet-runtime` actually has two extra lines, indicating the caller and definition sites. This PR still doesn't implement that in the compiler.
2. The string we use to show the value goes through some extra-careful processing in `sorbet-runtime`, as opposed to just running it through `inspect` or similar --- and this PR still doesn't implement all of that behavior in the compiler: https://github.com/sorbet/sorbet/blob/4f146522e25047f6527a98797672f9b1c462eced/gems/sorbet-runtime/lib/types/types/base.rb#L110-L130

This gets us close enough to make the tests mentioned under 'Motivation' happy, though.

### Motivation
There are a couple of tests in Stripe's codebase that are a little bit picky about the exact form of the error message here.

### Test plan
See included automated tests.

### DIFF
--- a/compiler/IREmitter/IREmitter.cc
+++ b/compiler/IREmitter/IREmitter.cc
@@ -667,10 +667,11 @@ void emitUserBody(CompilerState &base, cfg::CFG &cfg, const IREmitterContext &ir
                     auto local = irctx.rubyBlockArgs[0][i.argId];
                     auto var = Payload::varGet(cs, local, builder, irctx, 0);
                     if (auto &expectedType = argInfo.type) {
+                        auto description = fmt::format("Parameter '{}'", bind.bind.variable.toString(cs, cfg));
                         if (argInfo.flags.isBlock) {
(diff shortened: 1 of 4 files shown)


froydnj:
(This is total spitballing, as I haven't even looked at the patch, and is in no way a blocking request.)

Is it feasible to refactor sorbet-runtime so the compiler can call into the same function the interpreter would have used?

aprocter-stripe:

→ https://go/builds/bui_KM5QLJcuUKDAIT
→ https://go/builds/bui_KM5QREGP8qI4ru

PR #4691, opened 2021-10-06
Allow `requires_ancestor` to still accept positional arguments for backwards compatibility

vinistock:
Allow `requires_ancestor` to still support the old syntax in the runtime. Also, provide an error note about the easiest way to use the auto-correct for the new syntax.

### Motivation

The changes made in #4657 are easy to fix for the codebase being worked on. But unfortunately many gems are now using `sorbet-runtime`, which ends up requiring that every gem using `requires_ancestor` be released using the newest syntax before we can successfully upgrade Sorbet on codebases.

### Test plan

See included automated tests.


paracycle (approved):

rbi/sorbet/t.rbi
```diff
@@ -108,8 +108,8 @@ module T::Helpers
 
   def mixes_in_class_methods(mod, *mods); end
 
-  sig { params(block: T.proc.returns(Module)).void }
-  def requires_ancestor(&block); end
+  sig { params(mod: T.nilable(Module), mods: T.nilable(Module), block: T.nilable(T.proc.returns(Module))).void }
+  def requires_ancestor(mod = nil, *mods, &block); end
```
paracycle, line 112:
```suggestion
  sig { params(mods: T.nilable(Module), block: T.nilable(T.proc.returns(Module))).void }
  def requires_ancestor(*mods, &block); end
```

elliottt (changes requested):
Would it be worth splitting this into two PRs so that when it comes to deprecate the old functionality we can `git revert` the changes to the runtime and rbis? What is the schedule for deprecation, is a week too little or a month too much?

vinistock:
I split the backwards compatibility into a different commit, so that we can still revert it, but kept it in the same PR.

Concerning the time before reverting it, I'm not sure what the most appropriate one would be. My gut feeling would say a month. Do you have opinions on how long?

Morriar:
> I think a month is reasonable, my concern was that it would end up sticking around for longer than we intended and having an explicit date would help avoid that.

The shorter the better. The feature is already protected by `--enable-EXPERIMENTAL-requires-ancestor` and we do warn about this in the documentation:

image

vinistock:
@elliottt I cherry-picked the first commit into a different PR #4692. Didn't realize that all PRs were squashed. Once that is in, I can rebase this one.

PR #4698, opened 2021-10-07
generate code to parse the keyword args using interpreter-style keyword argument passing

froydnj:
### Motivation

Stop allocating a hash table for some method calls with keyword args.

#### Background 
The background here has been mostly written up in sorbet/ruby#5.  To summarize briefly, the calling convention for methods implemented in C (and, by extension, Sorbet compiler methods) has a very simple argument passing convention: all the positional arguments are passed on the Ruby stack, followed by a keyword splat hash (if applicable).  This convention implies that if you pass keyword arguments, you are always allocating a keyword splat hash, which is suboptimal.

Especially compared to the interpreter, which can avoid this hash allocation in some cases: the interpreter's calling convention for methods that take keyword args is to pass all the positional arguments, followed by all the values of the keyword arguments, and store the actual keys somewhere separate.  The compiler knows how to generate calls in this manner, and there is code inside the VM that knows how to convert from this convention to the convention for methods implemented in C.  There is also code inside the VM that knows how to roll up interpreter-style keyword arguments into a keyword splat hash if needed (e.g. a positional arg needs to be satisfied, or the particular method being called is too complex to deal with).

#### Changes in this PR

What we want to do in _this_ PR is to teach Sorbet compiler methods to understand when they are receiving interpreter-style arguments and use a relatively faster (and non-allocating) path to do this.  The basic idea is:

1. identify (at runtime) whether the method is being passed interpreter-style or C-style keyword arguments.  This boils down to (see `sorbet_can_efficiently_parse_kwargs` in the codegen payload):
  a. were we passed what looks like a keyword splat hash that we need to deal with?
  b. does the particular call have keyword arguments passed (this is testing for `VM_CALL_KWARG`)?
2. take the appropriate parsing path.

Note that compiled methods still have to support the keyword splat hash method of passing keyword arguments, because we don't have any mechanism for multiple entry points for a given compiled method.  (Multiple entry points is something worth exploring in the future, but for now this is the simplest thing.)  This also means that keyword argument parsing can generate quite a lot of code.

For now, I think the added code size is OK, but if we find there are problems with this, there are ways we can move some or all of the keyword argument parsing into VM support routines (at the cost of possibly forcing a few more things into memory, which may or may not be a problem).  I found it useful to move the actual "did we pass an arg for this keyword (`sorbet_kwarg_passed_value`)?" into the VM for debugging purposes so I could set breakpoints on it, and everything worked fine.

It's also worth noting that this is all-or-nothing: if the VM determines that it can pass keyword arguments efficiently, it's going to, and if we were for some reason to get things wrong on the compiler side, things will break.  I'm reasonably confident that there will be no problems due to mismatches: there have been several testcases added for cases where we were getting it wrong, and we have better test coverage of keyword arguments as a result.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/IREmitter.cc
+++ b/compiler/IREmitter/IREmitter.cc
@@ -119,21 +119,16 @@ void setupStackFrames(CompilerState &base, const ast::MethodDef &md, const IREmi
     }
 }
 
-void parseKeywordArgsFromCallData(CompilerState &cs, llvm::IRBuilderBase &builder,
-                                  const IREmitterContext &irctx,
-                                  cfg::LocalRef kwRestArgName,
-                                  llvm::Value *argCountRaw,
-                                  llvm::Value *argArrayRaw,
-                                  llvm::Value *hashArgs,
-                                  int maxPositionalArgCount,
+void parseKeywordArgsFromCallData(CompilerState &cs, llvm::IRBuilderBase &builder, const IREmitterContext &irctx,
+                                  cfg::LocalRef kwRestArgName, llvm::Value *argCountRaw, llvm::Value *argArrayRaw,
+                                  llvm::Value *hashArgs, int maxPositionalArgCount,
                                   const vector<core::ArgInfo::ArgFlags> &argsFlags, int rubyBlockId) {
     ENFORCE(rubyBlockId == 0);
     auto *func = irctx.rubyBlocks2Functions[rubyBlockId];
     auto &argPresentVariables = irctx.argPresentVariables[rubyBlockId];
     // required arguments remaining to be parsed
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_KNAzS2kwIX0OM3
→ https://go/builds/bui_KNAzxhsfoY3TQ7

froydnj:

→ https://go/builds/bui_KNB8DN7VvulWEh
→ https://go/builds/bui_KNB8vgvtBwZdbs

froydnj:

→ https://go/builds/bui_KOX7dacFNORuUn
→ https://go/builds/bui_KOX7o40QH6bhNK

froydnj:

→ https://go/builds/bui_KPJwcBpm62C92r
→ https://go/builds/bui_KPJwrxAfY1na7t

PR #4701, opened 2021-10-08
Fix array uniq! return type

MathyouMB:
The uniq! array function can return nil as described by the method documentation on lines 2695-2702 on `rbi/core/array.rbi`, but the current signature for the uniq! function does not allow the return to be nil.  This pr adds `T.nilable()` to the signature of uniq! to fix this issue.


froydnj:

→ https://go/builds/bui_KN5do5rvbPtwTT
→ https://go/builds/bui_KN5dR6LPQwsSYw

PR #4722, opened 2021-10-13
Reimplement `flatMapHack` as `map.flatten(1)`

paracycle:
Finally `flat_map` respects `to_ary` return type as well.

### Motivation

The current implementation of `flat_map` was using code which was similar to the `flatten` code but it was incomplete since it was not respecting the return type of `to_ary` methods implemented by the types inside arrays.

A better implementation is to implement it as `map.flatten(1)`, that is dispatch `flatten(1)` on the return type of the `map`ping operation, which allows us to reuse the full `Array_flatten` implementation that uses `to_ary` return types for better typing. Moreover, any improvements to `Array_flatten` will be reflected automatically in `flat_map` implementation as well.

### Test plan

See included automated tests.

I added a new test case for testing `to_ary` support in `flat_map` and also took the liberty of updating test to use `T.reveal_type` rather than `T.assert_type!` since the error output from the latter is not helpful at all.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -886,33 +886,8 @@ core::TypePtr Environment::getReturnType(core::Context ctx, const core::TypePtr 
     return applied->targs.front();
 }
 
-core::TypePtr flattenArrays(core::Context ctx, const core::TypePtr &type) {
-    core::TypePtr result;
-
-    typecase(
-        type,
-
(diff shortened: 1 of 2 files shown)


elliottt:

→ https://go/builds/bui_KP3buTiOetWJBi
→ https://go/builds/bui_KP3bqw80HxTQXI

PR #4727, opened 2021-10-13
Suppress errors::Resolver::RecursiveTypeAlias in autogen

ngroman:
Allow autogen to complete even if some `T.type_alias` RHS's don't resolve. Any true error here would be found during real typechecking.

### Motivation
Related to changes in https://github.com/sorbet/sorbet/pull/4721

### Test plan
Tested on Stripe's codebase.

See included automated tests.


nroman-stripe:

→ https://go/builds/bui_KP4xemJ4P1RDiF
→ https://go/builds/bui_KP4xE1YC2ipbc5

PR #4734, opened 2021-10-15
Implement rename local variable for lsp

vinistock:
Implement the local variable rename code action for the LSP. I have split the solution into separate commits to make it easier to review.

The main challenge to get this working is figuring out whether an `Ident` is a local variable or something else. This is how the solution is currently structured:
1. Prepare rename needs to allow trying to rename a local variable, but it's not easy to figure out whether something is a local variable or not without a complete tree walker, which feels excessive for the prepare rename action. Therefore, the solution instead defines that local variables must be renamed where they are assigned and looks for a `=` character in the source. Please, let me know your thoughts about this approach
2. Create a `LocalVarUsages` tree walker to find the locations where a given local variable is used inside a method
3. Change the rename action to process `Ident` responses, find all of the usages of the selected local variable and apply the edits

### Motivation

Similarly to renaming methods and constants, being able to rename other things such as local variables is a super nice quality of life improvement.

### Test plan

The included automated tests cover scenarios like having the same name used for two locals in different scope levels. Let me know if you have more ideas of how this could break.


vinistock:
@jez in the latest commit, I tried to get rid of the extra `getIdentLoc` method. It turns out that the reason trying to rename a local variable in assignment was returning the entire line was because that's the location we're using in `builder_walk`.

I changed it to use the location of the LHS. This simplifies the code, but breaks a few other tests. Let me know your thoughts about the approach.

froydnj:
I spent some time this morning looking at test failures and figuring out why `getIdentLoc` was needed.

The results are in #5706.  We don't need `getIdentLoc` for assignments, but we do need some filtering for locs corresponding to `attr_reader` and friends.  The location information around `attr_reader` and company is a giant pile of hacks, so I don't feel too bad about piling a couple more special cases on.

froydnj:
I talked with @vinistock about the issues around `getIdentLoc` and assignments, and he helped me understand what the problem is.  For something like:

```ruby
  type = out || type
  # ^ cursor on this `type`
```

if you are running #5706 and you hit F2 in VSCode with your cursor positioned as above, the text box that pops up says:

```
type = out || type
Enter to rename, Shift-Enter to preview
```

which seems weird, like it's going to nuke the entire line.  Typing `foo` into the text box results in:

```
foo = out || foo
```

which is the desired result, but is very weird for the user.  Running the version here has nicer behavior -- the text box will just have `type` in it, but there are obviously test failures.

There are two different issues here.  One is that in `builder_walk.cc`, we give the assignment to `type` the location of the entire line:

https://github.com/sorbet/sorbet/blob/1542140a549ba5b1dc21fe29831d59902a571687/cfg/builder/builder_walk.cc#L336-L337

and that is the location that gets matched by the location query we perform in `prepare_rename.cc`:

https://github.com/sorbet/sorbet/blob/1542140a549ba5b1dc21fe29831d59902a571687/main/lsp/requests/prepare_rename.cc#L57-L64

As @vinistock has pointed out, it doesn't seem very feasible to change the location information in `builder_walk.cc`; those testcase failures seem kind of weird.  (Though, really, I think it'd probably be a good idea if we gave the actual assignment in the CFG an non-existent location...)

The above issue leads to the second issue: when we get an `IdentResponse` from the above `byLoc` query, it comes back with a location that covers the entire assignment, not just the LHS as we would prefer.  That location then feeds into this line (in #5706; the code in this PR is currently similar, but uses the heuristically-adjusted location):

https://github.com/sorbet/sorbet/blob/3f8b52e4690490e177524a05bd3dba7f83d9b066/main/lsp/requests/prepare_rename.cc#L59

and we return the wrong placeholder text to VSCode.

Find all references suffers from this same sort of problem, since it uses a location-based query to get a response.  But instead of taking information from the `IdentResponse` of the location-based query, it then does a second query:

https://github.com/sorbet/sorbet/blob/da8484c594ea2a20e14fa77092afdbc9850a8be1/main/lsp/requests/references.cc#L58-L65

and all of the locations returned by that are based on the actual identifier, so the locations on those are much tighter.

So we could do this second sort of query...but we already have a `LocalVariable` for the exact identifier that we want to rename, so why not just use its name?  Changing the placeholder text to be:

```c++
    result->placeholder = identResp->variable._name.show(gs);
```

seems to do the right thing for me in VSCode.  I'm not sure if this would give weird results (local variables don't seem like they are meant to be shown to the user, since they don't have a `show()` method on them?), but it certainly seems like it should work...

test/testdata/lsp/rename/locals.rb
```diff
@@ -0,0 +1,103 @@
+# typed: true
```
jez, line 1: This is just a note for future reference, not a request to action in this PR, but

For `.rbedited` tests, I usually prefer making many small tests rather than one large test, because in the case where someone wants to e.g. make a small change to a single portion of this test, it now requires updating 10 extra files.

main/lsp/requests/prepare_rename.cc
```diff
@@ -22,6 +34,32 @@ variant<JSONNullObject, unique_ptr<PrepareRenameResult>> getPrepareRenameResult(
     return result;
 }
 
+variant<JSONNullObject, unique_ptr<PrepareRenameResult>>
+getPrepareRenameResultForIdent(const core::GlobalState &gs, const core::lsp::IdentResponse *identResp) {
+    ENFORCE(identResp->termLoc.exists());
+
+    // The loc for the instance variable local in `attr_reader :foo`
```
jez, line 41: Do we want tests for `prop` and/or `const` as well?

main/lsp/requests/rename.cc
```diff
@@ -38,6 +38,43 @@ bool isValidRenameLocation(const core::SymbolRef &symbol, const core::GlobalStat
...
+class LocalRenamer : public AbstractRenamer {
+public:
+    LocalRenamer(const core::GlobalState &gs, const LSPConfiguration &config, const string oldName,
+                 const string newName, std::vector<core::Loc> localUsages)
+        : AbstractRenamer(gs, config, oldName, newName), localUsages(localUsages) {
+        // If the name is the same as before or empty, return an error. VS Code already prevents this on its own, but
+        // other IDEs might not
+        if (newName.empty()) {
+            invalid = true;
+            error = "The new name must not be empty.";
```
jez, line 50: ooc, do we have a test for this?

test/testdata/lsp/rename/locals.rb
```diff
@@ -0,0 +1,103 @@
...
+    5.times do
+      higher = 321
+    end
+
+    higher
+  end
+end
+
+class MethodParameter
+  def add(a, b)
```
jez, line 75: This will not currently handle signatures, either:

```ruby
sig {params(x: Integer).void}
def foo(x)
  x
end
```

I think that's fine for now (I don't think that we have this information readily available, so not worth blocking this change on it), but could we add a test to capture the current behavior?

jez (approved):
Just a few suggestions for extra tests to write. This change looks great. Thanks especially to @froydnj for typing up that comment about why the hacks around assignment source parsing are no longer required.

test/testdata/lsp/rename/locals.rb
```diff
@@ -0,0 +1,103 @@
...
+  private
+
+  def process(a)
+  end
+end
+
+class Bar
+  def foo
+    a = "hello"
+#   ^ apply-rename: [E] newName: a placeholderText: a invalid: true expectedErrorMessage: The new name cannot be the same as the oldname.
```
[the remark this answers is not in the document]
vinistock: VS Code actually does not send the request if the name matches, but I don't if that's what happens for other editors. I'm okay with removing the error. What do you think?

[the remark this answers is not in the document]
jez: wfm

main/lsp/requests/rename.cc
vinistock: I can't seem to get a test working with an empty `newName` attribute. Is there some trick? Alternatively, we can also just remove this since VS Code won't allow empty names.

jez: In that case I’d rather just leave this behavior in and not test it.

test/testdata/lsp/rename/locals.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+
+    higher
+  end
+end
+
+class MethodParameter
+  def add(a, b)
+        # ^ apply-rename: [H] newName: c placeholderText: a
+    a + b
```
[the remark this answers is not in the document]
vinistock: Added new tests and included the signatures in them.

main/lsp/requests/prepare_rename.cc
vinistock: Can props be matched somehow? The issue with attributes was that they are connected to variables, which my trigger incorrect renames.

jez:

→ https://go/builds/bui_LgU9x3Bkt69ZbI
→ https://go/builds/bui_LgU9Om2BuFaiXB

PR #4737, opened 2021-10-15
URL decode remote names when mapping them to local names

### LINKED ISSUE #4736: LSP server does not handle URL encoded file names properly
#### Input

Here is the LSP trace when opening a file named `ansi@1.5.0.rbi` (contents of the file text has been shortened):

<details>
<summary>LSP Trace</summary>

```
[Trace - 10:55:34 PM] Sending notification 'textDocument/didOpen'.
Params: {
    "textDocument": {
        "uri": "file:///Users/ufuk/src/github.com/Shopify/tapioca/sorbet/rbi/gems/ansi%401.5.0.rbi",
        "languageId": "ruby",
        "version": 1,
        "text": "# DO NOT EDIT MANUALLY\n# This is an autogenerated file for types exported from the `ansi` gem.\n# Please instead update this file by running `bin/tapioca gem ansi`.\n\n# typed: true\n\n# ANSI namespace module contains all the ANSI related classes.\nmodule ANSI\n  extend ::ANSI::Constants\n  extend ::ANSI::Code\nend\n"
    }
}


[Trace - 10:55:34 PM] Sending request 'textDocument/codeAction - (42)'.
Params: {
    "textDocument": {
        "uri": "file:///Users/ufuk/src/github.com/Shopify/tapioca/sorbet/rbi/gems/ansi%401.5.0.rbi"
    },
    "range": {
        "start": {
            "line": 0,
            "character": 0
        },
        "end": {
            "line": 0,
            "character": 0
        }
    },
    "context": {
        "diagnostics": []
    }
}


[Trace - 10:55:34 PM] Received notification 'sorbet/showOperation'.
Params: {
    "operationName": "SlowPathBlocking",
    "description": "Typechecking...",
    "status": "start"
}


[Trace - 10:55:34 PM] Received notification 'sorbet/showOperation'.
Params: {
    "operationName": "SlowPathBlocking",
    "description": "Typechecking...",
    "status": "end"
}


[Trace - 10:55:34 PM] Received notification 'sorbet/showOperation'.
Params: {
    "operationName": "SlowPathNonBlocking",
    "description": "Typechecking in background",
    "status": "start"
}


[Trace - 10:55:34 PM] Received response 'textDocument/codeAction - (42)' in 196ms.
Result: []


[Trace - 10:55:34 PM] Received notification 'sorbet/showOperation'.
Params: {
    "operationName": "SlowPathNonBlocking",
    "description": "Typechecking in background",
    "status": "end"
}
```

</details>

#### Observed output

The LSP server is hitting the slow path.

#### Expected behavior

No slow path should be run, the opened file is a part of the Sorbet folder.

- - -

We've narrowed the problem down to the fact that the file `ansi@1.5.0.rbi` is being opened as `ansi%401.5.0.rbi`. Since Sorbet LSP server does no URL decoding on the `file://` the file is treated as a new addition to the workspace and slow path is triggered.

paracycle:
`remoteName2Local` should make sure the URI it works on is URL-decoded before doing anything else to ensure that file name and `rootPath` matches are correct.

Coincidentally, this also allows us to remove the special handling for `%3A` to handle `sorbet:https://` URLs.

I rolled my own implementation of `urlDecode` which _should_ ™️ work, but please let me know if I should be using something more tested and/or an alternative implementation.

### Motivation

Fixes #4736 

### Test plan

~I tried to write a test for this in `protocol_test_corpus` and got much of the way there but I am having problems replicating the slow path behaviour with a URL encoded open request. With some help I should be able to put a test case together, though.~

See included automated tests.

### DIFF
--- a/main/lsp/LSPConfiguration.cc
+++ b/main/lsp/LSPConfiguration.cc
@@ -145,9 +145,13 @@ string urlDecode(string_view uri) {
     vector<pair<const absl::string_view, string>> replacements;
 
     for (size_t pos = uri.find('%'); pos != string::npos; pos = uri.find('%', pos + 1)) {
+        // "%3a"
         auto from = uri.substr(pos, 3);
-        auto to = absl::HexStringToBytes(from.substr(1));
-        replacements.push_back({from, to});
+        // add replacement only if % is actually followed by exactly 2 hex digits
+        if (from.size() == 3 && isxdigit(from[1]) && isxdigit(from[2])) {
+            auto to = absl::HexStringToBytes(from.substr(1));
+            replacements.push_back({from, to});
+        }
     }
 
     return absl::StrReplaceAll(uri, replacements);


jez:
> > I tried to write a test for this in `protocol_test_corpus` and got much of the way there but I am having problems replicating the slow path behaviour with a URL encoded open request. With some help I should be able to put a test case together, though.
> 
> I can think about this for a second. The rest of the change looks good.

As a first step, I think you could cargo cult the existing test that asserts things about `https%3A` URLs and make sure that if you send a `textDocument/didOpen` and send some requests on it, that things work. It might be that such a test already passes on master if the only thing that doesn't work on master is that sorbet fails to equate two files, but it would be a good first step, and from there we could figure out if there's some way to assert something about the fast path.

paracycle:
> As a first step, I think you could cargo cult the existing test that asserts things about `https%3A` URLs and make sure that if you send a `textDocument/didOpen` and send some requests on it, that things work. It might be that such a test already passes on master if the only thing that doesn't work on master is that sorbet fails to equate two files, but it would be a good first step, and from there we could figure out if there's some way to assert something about the fast path.

Yeah, that's exactly what I tried to do in the test above, but I guess I cargo-culted from the `DoesNotTypecheckSorbetURIs` test instead of the `SorbetURIsWork` one and got hung up on asserting for the failure of the slow path. Let me see if I can do a `readFile` to assert that the file is the one that is expected.

paracycle:
Ok, I have a test that fails without this PR and passes with it. It also caught a bug in my `urlDecode` implementation 😅 that I've fixed.

PR #4740, opened 2021-10-15
Re-enable cross-file direct final method calls

elliottt:
### Lazy Stack Frame Values
In #4619 we disabled direct calls to compiled final methods that would exist in different shared objects. The reason for disabling this optimization was that the calling convention for final methods required loading their stack frame value from a global stored in a different shared object, and this dependency was being resolved by the runtime linker at load time instead of lazily. The solution to the eager loading is to wrap up the use of that stack frame value in a function that returns the value of the global, as functions will be resolved lazily by the linker at runtime. Additionally llvm will optimize away the use of the function internally in the module, so there should be no performance penalty from this change.

### Accidentally Recursive Forwarding Functions
Fixing the eager loading of the stack values led to a different problem: the forwarding functions that we were generating for the case where we emitted a final method call to an interpreted function were going into an infinite loop when the final method existed in a different file. The reason for this was that if the final method's function symbol hadn't been resolved by the time the call happened in the file that had the forwarder, the forwarder would win and become the implementation of that method. At this point calling the method would result in an infinite loop, as the forwarder would dispatch to the vm which would dispatch back to the forwarder again.

The solution here ended up being pretty simple: we already require that final methods are turned into direct calls only when the callee exists in another compiled file. Since the forwarding function existed to handle the case where the callee was interpreted, we don't need it anymore. Now we generate a forward-declaration when the function exists in a different file, and rely on the runtime linker for the real symbol at runtime.

### Motivation

Re-enabling cross-file direct calls to final methods.

### Test plan

See included automated tests.


elliottt:

→ https://go/builds/bui_KPqEYEfqoUW1Ji
→ https://go/builds/bui_KPqEWdX9MAYJmy

elliottt:

→ https://go/builds/bui_KQvboj119aqVxM
→ https://go/builds/bui_KQvbmq9abI2U3v

PR #4743, opened 2021-10-18
delete initializeStaticInitNamesEntry field/basic block

froydnj:
### Motivation

This field/basic block was write-only.  I'm not sure what was intended to go here...maybe things like `Payload::cPtrToRubyRegexp` were intended to write their calls to this basic block rather than the global constructors basic block?

### Test plan

See included automated tests.


elliottt:
I think this was left over from when static-init functions all shared the same frame, and we needed a place to guarantee initialization had taken place before any of them ran. Thanks for cleaning it up!

PR #4757, opened 2021-10-19
Refine error message when constant names are the same

paracycle:
I pulled the error message construction into a private method in `T::Types::Base` so that `T::Types::Simple` can override it and refine the error message with an extra notice if the constant names are the same.

### Motivation

If we have the name of the expected type and the name of the object class the same but referring to different constants, the default error message ends up being confusing. In those circumstances, it would be helpful to show that the constants refer to different objects even though they have the same name and that this is probably a code reloading issue.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/types/base.rb
+++ b/gems/sorbet-runtime/lib/types/types/base.rb
@@ -133,7 +133,7 @@ module T::Types
       if valid?(obj)
         nil
       else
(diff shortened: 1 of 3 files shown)


paracycle:
Oh right, I totally forgot to include an example. Thanks for including it @aprocter-stripe 🙏

PR #4765, opened 2021-10-20
Apply patch to bundler to remove dependency on set

aprocter-stripe:
When building `sorbet_ruby`, we bundle a few things into the tarball, including, well, `bundler`. The version of `bundler` we're using depends on `set` from the Ruby standard library, which is a problem because we want to be able to compile that file, but `bundler` loads it before we get a chance to do the required shenanigans to substitute in the compiled version.

Note that the patch is sourced from https://github.com/sorbet/bundler (branch `sorbet_ruby_bundler`). As with our Ruby 2.7 patch, it can be regenerated with a script, added as part of this PR.

### Motivation
Want to compile `set.rb`.

### Test plan
Existing CI, plus a test against Stripe's codebase should cover this well.

### DIFF
--- a/third_party/ruby/build-ruby.bzl
+++ b/third_party/ruby/build-ruby.bzl
@@ -158,6 +158,10 @@ run_cmd gem uninstall rubygems-update
 cp "$out_dir/bin/bundle" "$out_dir/bin/bundler"
 
 {install_gems}
+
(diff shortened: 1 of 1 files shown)


aprocter-stripe:

→ https://go/builds/bui_KRhkvBgocBifJ7
→ https://go/builds/bui_KRhknH9H7gX4N2

PR #4766, opened 2021-10-21
Add shim for Class.new

Morriar:
As per https://ruby-doc.org/core-2.5.3/Class.html#method-c-new.

Mostly wondering what it would break 🤔

### DIFF
--- a/rbi/core/class.rbi
+++ b/rbi/core/class.rbi
@@ -171,14 +171,14 @@ class Class < Module
   end
   sig do
     params(
-        blk: T.proc.params(arg0: Class).returns(BasicObject),
+        blk: T.proc.bind(Class).params(arg0: Class).returns(BasicObject),
     )
     .void
   end
   sig do
     params(
         superclass: Class,
-        blk: T.proc.params(arg0: Class).returns(BasicObject),
+        blk: T.proc.bind(Class).params(arg0: Class).returns(BasicObject),
     )
     .void
   end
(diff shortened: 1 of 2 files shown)


Morriar:
Without the binding, inside `self` is bound to the outer scope where the block is defined.

So code like this will work by accident:

```rb
class Foo 
  Class.new do
    T.reveal_type(self) # T.class_of(Foo)

    include Mixin
  end
end

class Bar
  def self.bar
    Class.new do
      T.reveal_type(self) # T.class_of(Bar)

      include Mixin
    end
  end
end
```

This too:

```rb
module Mixin; end

Class.new do
  T.reveal_type(self) # T.class_of(<root>) 

  include Mixin
end
```

But this not:

```rb
module Mixin; end

def foo
  Class.new do
    T.reveal_type(self) # Object

    include Mixin # error: Method include does not exist on Object
  end
end
```

Where the last case might be the most used one.
By applying this shim we get closer from the truth than by using the outer scope.

Note that the user can still use a local `T.bind(self, X)` as an escape hatch when the type is not precise enough:

```rb
class Foo
  def self.foo; end
end

Class.new(Foo) do
  T.bind(self, T.class_of(Foo))

  foo
end
```

Or even locally bind `self` to `T.untyped` to remove all checks:

```rb
Class.new do
  T.bind(self, T.untyped)

  anything
end
```

I added a few more tests and merged this definition with the ones already existing for `Class#initialize`.

PR #4773, opened 2021-10-22
add sigs to T::Props::Serializable#{de,}serialize

froydnj:
### Motivation

Somebody on Slack mentioned this; might as well try it to see how hard it is.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/sorbet/tprops.rbi
+++ b/rbi/sorbet/tprops.rbi
@@ -146,8 +146,10 @@ module T::Props::PrettyPrintable::DecoratorMethods
 end
 
 module T::Props::Serializable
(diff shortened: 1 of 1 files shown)


rbi/sorbet/tprops.rbi
```diff
@@ -146,8 +146,10 @@ module T::Props::PrettyPrintable::DecoratorMethods
 end
 
 module T::Props::Serializable
+  sig {params(hash: T::Hash[T.untyped, T.untyped], strict: T.nilable(T::Boolean)).returns(T.self_type)}
```
djudd, line 149: This should probably be `T::Hash[String, T.untyped]`. That might need more fixes to existing code but it seems more likely to be helpful too - the most common kind of error I've seen here is passing Symbol-keyed hashes.

PR #4783, opened 2021-10-26
rework block arg handling to avoid allocations in (some) common cases

froydnj:
### Motivation

The motivating example here is something like:

```ruby
def foo(args, &blk)
  ...
  begin
    yield
  ensure
    ...
end
```

With the current compiler, it will look like `blk` is referenced by the top-level method (ruby block 0) and the begin "block" (ruby block 1 or so).  Therefore, `blk` will be eagerly converted to a proc (setting aside issues of type-checking) so that it can be stored on the Ruby stack, since the "variable" for the block is referenced in multiple Ruby blocks.

This is clearly suboptimal: the extra allocation is totally unnecessary.

This approach this PR takes is to more carefully track how the block arg is used throughout the method: we can avoid turning the block into a proc if all its uses are either in the top-level method or in a "block" that would share the same stack frame as the method (since the original block could still be fetched always as though we were in the top-level method).  `TrackCaptures`, then, is augmented to do this extra tracking of how the block argument is used.  I think this provides a much firmer foundation for our current optimizations of when we can pass through block and yield directly to blocks...but it is also a moderately large amount of code.

This PR does not (yet?) address the equally common case of:

```ruby
def foo(args, &blk)
  bar do
    ...
    yield
    ...
  end
  ...
end
```

The VM avoids allocating a complete proc for `blk` here by passing a proxy object that knows how to fetch the original block.  We should do this too, and I think this PR might be able to be extended to handle that case.

### Test plan

See included automated tests.


jez (approved):

froydnj:

→ https://go/builds/bui_KbPIWw4ON3dgR3
→ https://go/builds/bui_KbPIyMw6ZvpXCG

PR #4793, opened 2021-10-27
Update casting methods (`T.let`, `T.cast`, etc) to take the type as a block

jeffcarbs:
NOTE: I know this would take some updates to the static type checking also but wanted to get a temperature check on this approach first.

### Motivation

When debugging a performance issue we found that sometimes simply initializing types is slow. This is especially true to `T.any` (and thus `T.nilable` since it's an alias for `T.any(NilClass, ...)`). This means if you have a `T.let` on a hot path, there's no way to not pay a performance penalty even if you make an attempt with `checked: false`.

Here are some benchmarks

```ruby
Benchmark.bm do |x|
  x.report('any                 ') { 10000.times { T.let(1, T.any(Integer, String), checked: false) } }
  x.report('nilable any         ') { 10000.times { T.let(1, T.nilable(T.any(Integer, String)), checked: false) } }
  x.report('complex type        ') { 10000.times { T.let(1..1, T::Range[Integer], checked: false) } }
  x.report('nilable complex type') { 10000.times { T.let(1..1, T.nilable(T::Range[Integer]), checked: false) } }
  x.report('simple type         ') { 10000.times { T.let(1, Integer, checked: false) } }
  x.report('nilable simple type ') { 10000.times { T.let(1, T.nilable(Integer), checked: false) } }
end

       user     system      total        real
any                   0.044301   0.000395   0.044696 (  0.050076)
nilable any           0.064349   0.000205   0.064554 (  0.064723)
complex type          0.005004   0.000049   0.005053 (  0.005101)
nilable complex type  0.035156   0.000288   0.035444 (  0.035752)
simple type           0.001076   0.000019   0.001095 (  0.001120)
nilable simple type   0.007957   0.000043   0.008000 (  0.008037)
```

I've explicitly disabled type checking in these benchmarks sure that simply initializing the types is the bottleneck

This PR updates `T.let` to take a block (similar to `T.type_alias`) and only evaluates the block if we actually want to type check. Here are the benchmarks after that PR when passing the types via block:
```
Benchmark.bm do |x|
  x.report('any                 ') { 10000.times { T.let(1, checked: false) { T.any(Integer, String) } } }
  x.report('nilable any         ') { 10000.times { T.let(1, checked: false) { T.nilable(T.any(Integer, String)) } } }
  x.report('complex type        ') { 10000.times { T.let(1..1, checked: false) { T::Range[Integer] } } }
  x.report('nilable complex type') { 10000.times { T.let(1..1, checked: false) { T.nilable(T::Range[Integer]) } } }
  x.report('simple type         ') { 10000.times { T.let(1, checked: false) { Integer } } }
  x.report('nilable simple type ') { 10000.times { T.let(1, checked: false) { T.nilable(Integer) } } }
end

       user     system      total        real
any                   0.001261   0.000065   0.001326 (  0.001369)
nilable any           0.001186   0.000002   0.001188 (  0.001190)
complex type          0.001253   0.000041   0.001294 (  0.001339)
nilable complex type  0.001200   0.000002   0.001202 (  0.001202)
simple type           0.001176   0.000002   0.001178 (  0.001179)
nilable simple type   0.001234   0.000007   0.001241 (  0.001272)
```

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -127,10 +127,10 @@ module T
   # exception at runtime if the value doesn't match the type.
   #
   # Compared to `T.let`, `T.cast` is _trusted_ by static system.
-  def self.cast(value, type, checked: true)
+  def self.cast(value, type=nil, checked: true, &blk)
     return value unless checked
 
-    Private::Casts.cast(value, type, cast_method: "T.cast")
+    Private::Casts.cast(value, type || blk.call, cast_method: "T.cast", &blk)
   end
 
   # Tells the typechecker to declare a variable of type `type`. Use
@@ -142,10 +142,10 @@ module T
   #
   # If `checked` is true, raises an exception at runtime if the value
   # doesn't match the type.
-  def self.let(value, type, checked: true)
+  def self.let(value, type=nil, checked: true, &blk)
     return value unless checked
 
-    Private::Casts.cast(value, type, cast_method: "T.let")
+    Private::Casts.cast(value, type || blk.call, cast_method: "T.let", &blk)
   end
 
   # Tells the type checker to treat `self` in the current block as `type`.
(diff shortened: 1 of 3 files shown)


jeffcarbs:
Just wanted to clarify - is this (or #4181) something you would definitely not consider currently? This isn't something that can be monkey-patched since sorbet-static would need to be updated to handle parsing the type from the block (as you did in that other PR) so just wanted to bump this.

There technically is a workaround by using a type alias but it's still an order of magnitude slower than the block syntax. You can also pre-save the alias which makes it just as fast as the block syntax but makes for worse readability and a gotcha that you'd have to know to do.

```
CachedAlias = T.type_alias { T.nilable(T.any(Integer, String)) }

Benchmark.bm do |x|
  x.report('nilable any') { 10000.times { T.let(1, T.nilable(T.any(Integer, String)), checked: false) } }
  x.report('type alias ') { 10000.times { T.let(1, T.type_alias { T.nilable(T.any(Integer, String)) }, checked: false) } }
  x.report('saved alias') { 10000.times { T.let(1, CachedAlias, checked: false) } }
  x.report('block      ') { 10000.times { T.let(1, nil, checked: false) { T.nilable(T.any(Integer, String)) } } }
  x.report('simple     ') { 10000.times { T.let(1, Integer, checked: false) } }
  x.report('no let     ') { 10000.times { 1 } }
end; nil

       user     system      total        real
nilable any  0.105534   0.010640   0.116174 (  0.117069)
type alias   0.009197   0.001514   0.010711 (  0.010732)
saved alias  0.001278   0.000001   0.001279 (  0.001279)
block        0.001299   0.000004   0.001303 (  0.001311)
simple       0.001289   0.000020   0.001309 (  0.001333)
no let       0.000464   0.000001   0.000465 (  0.000465)
```

elliottt:
We're definitely open to making this change to the runtime. I think that because we've already vetted #4181 that we'll revive that PR. Knowing that others would make use of this functionality makes it easier to add, given that it didn't quite fit our original use-case.

jeffcarbs:
> We're definitely open to making this change to the runtime. I think that because we've already vetted #4181 that we'll revive that PR.

Great to hear! I’ll close this out in that case.

PR #4800, opened 2021-10-28
Use `default_checked_level` in inline type checking (`T.let`, `T.cast`, etc)

jeffcarbs:
Alternative approach to #4236, using the `default_checked_level` to determine whether to do inline type checking, not just for sig type checking.

I read through that PR and tried to address some of the performance concerns by using `replace_method` to swap to a faster implementation. This poses a challenge that I'll comment on inline.

### Motivation

We want to be able to easily disable sorbet-runtime in production while keeping it during development and tests. Currently there's `T::Configuration.default_checked_level` which enables globally opting out of type checking but this only works for sigs currently, not inline type checking like `T.let` and friends. This PR enables using the same mechanism to turn off more of the runtime type checking.

### Test plan

No tests yet, wanted to validate the approach.

### DIFF
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -123,14 +123,22 @@ module T
   end
 
   # Tells the typechecker that `value` is of type `type`. Use this to get additional checking after
-  # an expression that the typechecker is unable to analyze. If `checked` is true, raises an
-  # exception at runtime if the value doesn't match the type.
+  # an expression that the typechecker is unable to analyze. If `checked` is :always, raises an
+  # exception at runtime if the value doesn't match the type (defaults to `T::Configuration.default_checked_level`).
   #
   # Compared to `T.let`, `T.cast` is _trusted_ by static system.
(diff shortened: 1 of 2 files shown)


aprocter-stripe:
So we chatted about this a bit at our team meeting today and I think the main concern is that the functions being changed are pretty perf-sensitive.

@jeffcarbs Could you try measuring the impact to the benchmarks in https://github.com/sorbet/sorbet/blob/master/gems/sorbet-runtime/bench/typecheck.rb ? (Note that you may need to make some tweaks to run them with `default_checked_level` adjusted.)

jeffcarbs:
Thanks for chatting about this and I definitely appreciate the concern given how performance sensitive this is. I'm gonna close this out for now since I don't really have time to pursue further and this technically can be fixed via monkey-patch for now (ugly but technically possible).

Also while this definitely makes things more symmetrical I'm not entirely sure it's the right approach. Might be better to do something simpler and more explicit like in the PR linked in the description where sorbet exposes a method to turn this methods into no-ops.

PR #4806, opened 2021-10-29
make "generic" prop getters take the attr_reader fast path at runtime when possible

froydnj:
### Motivation

Props in the interpreter have a fast path defined for them, assuming their `prop_get` method was defined by `T::Props::Decorator`.  (We're not considering `prop_set` for the purposes of this PR.)

https://github.com/sorbet/sorbet/blob/b4bd63aa4f3646900dca12e9eb93e7f04238dbe2/gems/sorbet-runtime/lib/types/props/decorator.rb#L367-L389

The prop rewriter in Sorbet attempts to emulate this, but it only chooses the `attr_reader` fastpath when the syntactically enclosing class is known to be a non-model:

https://github.com/sorbet/sorbet/blob/b4bd63aa4f3646900dca12e9eb93e7f04238dbe2/rewriter/Prop.cc#L336-L364

Everything else that *could* be a subclass of `Chalk::ODM::Document` takes the much slower path, which goes through `instance_variable_get` and `prop_get_logic`.

#4801 sped up the `instance_variable_get` path, but it's still significantly slower than the interpreter.

This PR proposes that we propagate information from the rewriter to the compiler about prop getters that we think need to go through the fully generic path.  The compiler then generates special code to define the prop getter via a new payload function.  This function uses the same logic as sorbet-runtime to determine whether we can take the same `attr_reader` fast path.  If we can't, the compiled version of the slowpath is likely faster due to the `instance_variable_get` optimizations, so we use that.

This change does hurt type safety, as such props will no longer have their types checked, since accesses go through `attr_reader`.  I think, however, that we already have the same sorts of issues with props that would take the rewriter fastpath, so I don't think this technically makes anything worse.

Benchmarks on a toy benchmark:

```ruby
# frozen_string_literal: true
# typed: true
# compiled: true

class MySuperclass < T::InexactStruct
end

# Hide that we're actually a T::InexactStruct subclass to force the slowpath
class MyStruct < MySuperclass
  const :foo, Integer
end

my_struct = MyStruct.new(foo: 430)

i = 0
while i < 10_000_000

  my_struct.foo

  i += 1
end

puts i
puts my_struct.foo
```

Before this change:

```
stripe/while_10_000_000.rb      .225    .102
stripe/inexact_struct_getter.rb .292    1.171
stripe/inexact_struct_getter.rb - baseline      .067    1.069
```

After this change:

```
stripe/while_10_000_000.rb      .223    .097
stripe/inexact_struct_getter.rb .296    .189
stripe/inexact_struct_getter.rb - baseline      .073    .092
```

Not quite as fast as the interpreter (though some of that might be noise), but at least we're not an order of magnitude+ slower any more.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -364,9 +364,10 @@ public:
         bool isRewriterSynthesized : 1;
         bool isAttrReader : 1;
         bool discardDef : 1;
+        bool genericPropGetter : 1;
 
         // In C++20 we can replace this with bit field initialzers
-        Flags() : isSelfMethod(false), isRewriterSynthesized(false), isAttrReader(false), discardDef(false) {}
+        Flags() : isSelfMethod(false), isRewriterSynthesized(false), isAttrReader(false), discardDef(false), genericPropGetter(false) {}
     };
     CheckSize(Flags, 1, 1);
 
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -367,7 +367,9 @@ public:
         bool genericPropGetter : 1;
 
         // In C++20 we can replace this with bit field initialzers
-        Flags() : isSelfMethod(false), isRewriterSynthesized(false), isAttrReader(false), discardDef(false), genericPropGetter(false) {}
+        Flags()
+            : isSelfMethod(false), isRewriterSynthesized(false), isAttrReader(false), discardDef(false),
+              genericPropGetter(false) {}
     };
     CheckSize(Flags, 1, 1);
 
--- a/compiler/IREmitter/Payload/codegen-payload.c
+++ b/compiler/IREmitter/Payload/codegen-payload.c
@@ -158,6 +158,8 @@ SORBET_ALIVE(void, sorbet_vm_register_sig,
              (VALUE isSelf, VALUE method, VALUE self, VALUE arg, rb_block_call_func_t block));
 SORBET_ALIVE(void, sorbet_vm_define_method,
              (VALUE klass, const char *name, rb_sorbet_func_t methodPtr, void *paramp, rb_iseq_t *iseq, bool isSelf));
+SORBET_ALIVE(void, sorbet_vm_define_prop_getter,
+             (VALUE klass, const char *name, rb_sorbet_func_t methodPtr, void *paramp, rb_iseq_t *iseq));
 
 SORBET_ALIVE(VALUE, sorbet_maybeAllocateObjectFastPath, (VALUE recv, struct FunctionInlineCache *newCache));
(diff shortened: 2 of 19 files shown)


aprocter-stripe:
So I think the change makes sense as far as how the information is plumbed down to `DefineMethodIntrinsic`, and how we act on that information. A question about this comment:

> This change does hurt type safety, as such props will no longer have their types checked, since accesses go through attr_reader. I think, however, that we already have the same sorts of issues with props that would take the rewriter fastpath, so I don't think this technically makes anything worse.

The implementation in this PR should be exactly as type-unsafe as what we get with `define_getter_and_setter`, right?

froydnj:
> So I think the change makes sense as far as how the information is plumbed down to `DefineMethodIntrinsic`, and how we act on that information. A question about this comment:
> 
> > This change does hurt type safety, as such props will no longer have their types checked, since accesses go through attr_reader. I think, however, that we already have the same sorts of issues with props that would take the rewriter fastpath, so I don't think this technically makes anything worse.
> 
> The implementation in this PR should be exactly as type-unsafe as what we get in interpreted code via `define_getter_and_setter`, right?

Right.  Well, mostly.  The situation in interpreted code today:

* `T::Struct` gets type-unsafe accessors (`attr_reader`)
* `Chalk::ODM::Document` gets type-unsafe accessors, but it depends on the particular class whether it gets `attr_reader` accessors or `prop_get_logic`-style accessors

The situation in compiled code today:

* `T::Struct` gets type-unsafe accessors (`attr_reader`)
* `Chalk::ODM::Document` gets *type-safe* accessors always, since Sorbet rewrites props on such classes to always go through the `prop_get_logic`-style accessors

The situation this patch proposes:

* `T::Struct` gets type-unsafe accessors (`attr_reader`) -- nothing changes here
* `Chalk::ODM::Document` gets type-unsafe `attr_reader`-style accessors or type-safe `prop_get_logic` style accessors, depending on the particular class, but we don't know which until runtime.

Which is not ideal from a type-safety perspective, but like you say, at least not worse than what interpreted code faces today.

Actually, come to think of it, I think we have the checks for not-`final` `attr_reader` methods when defining methods precisely because we would normally assume `final` compiled methods have their return values type-checked.  Except that `attr_reader`-style methods wouldn't be doing that checking, and therefore we might do incorrect optimizations if we did assume that checking happened.  Perhaps we should add similar `ENFORCE`s or even `failCompilation` for final `genericPropGetter` methods because of a similar reasoning?

aprocter-stripe (approved):
Thanks for the clarification! :shipit:

PR #4809, opened 2021-10-29
Propagate `T.bind` binding to rescue blocks

vinistock:
Propagate the type assigned to `self` by a `T.bind` to the `rescue` and `ensure` blocks inside the same method.

**Before this change**
```ruby
def foo
  T.bind(self, String)
  T.reveal_type(self) # reveals String
rescue
  T.reveal_type(self) # reveals Object
end
```
**After this change**
```ruby
def foo
  T.bind(self, String)
  T.reveal_type(self) # reveals String
rescue
  T.reveal_type(self) # reveals String
end
```

### Motivation

Although, we could circumvent the issue by adding other `T.bind` statements inside `rescue` and `ensure` blocks, the developer experience is a bit counterintuitive.

It's more natural to expect that binding at the beginning will apply to the entire method/block. If by any reason the binding needs to be different in the `rescue` block, we can change it again by adding another `T.bind`, but I think it makes sense to propagate from the higher level `T.bind`.

### Concerns

Anything unexpected that could come from propagating the `T.bind` into the rescue scope?

### Test plan

See included automated tests.


jez (approved):
Thanks, sorry for the long review.

Do you mind rebasing this on top of the latest master?

I just noticed yesterday that `cfg-text` exp files were not getting checked in CI, and landed #4857 to fix it.

PR #4811, opened 2021-10-29
Fix `non_forcing_is_a?` in package mode

aisamanra:
### Motivation
Fix the behavior of `non_forcing_is_a?` in package mode.

The issue is specifically that we want `non_forcing_is_a?("C::D", package: "A::B")` to act like we're looking up `::A::B::C::D` in non-package mode. In packaged mode, this means we need to be looking up the constant `A::B::C::D` as defined by the `A::B` package. We were previously assuming an older semantics in which we'd look up `C::D` in the `A::B` package.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1929,7 +1929,6 @@ class ResolveTypeMembersAndFieldsWalk {
                     current = core::Symbols::root();
                     continue;
                 } else {
-
                     auto package = core::cast_type_nonnull<core::LiteralType>(packageType);
                     auto packageName = package.asName(ctx);
                     auto mangledName = packageName.lookupMangledPackageName(ctx.state);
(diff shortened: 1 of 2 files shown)


resolver/resolver.cc
```diff
@@ -1909,15 +1929,6 @@ class ResolveTypeMembersAndFieldsWalk {
...
                     continue;
                 } else {
-                    if (part == "") {
-                        if (auto e = ctx.beginError(stringLoc, core::errors::Resolver::LazyResolve)) {
-                            e.setHeader("The string given to `{}` should not be an absolute constant reference if a "
-                                        "package name is also provided",
-                                        method);
-                        }
-                        return;
-                    }
```
aisamanra, line 1919: This check was moved upwards, not removed.

aprocter-stripe (approved):

aisamanra:

→ https://go/builds/bui_KV4L25OfMlt02T
→ https://go/builds/bui_KV4L4HkRhp3w18

PR #4812, opened 2021-11-01
add a fastpath for instance_variable_set with a constant symbol argument

froydnj:
### Motivation

#4801 was pretty effective; let's do the same change for `instance_variable_set` so we can reap speedups there too.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/NameBasedIntrinsics.cc
+++ b/compiler/IREmitter/NameBasedIntrinsics.cc
@@ -732,6 +732,45 @@ public:
     }
 } InstanceVariableGet;
 
+class InstanceVariableSet : public NameBasedIntrinsicMethod {
+public:
+    InstanceVariableSet() : NameBasedIntrinsicMethod{Intrinsics::HandleBlock::Unhandled} {};
+
+    virtual llvm::Value *makeCall(MethodCallContext &mcctx) const override {
+        if (mcctx.send->args.size() != 2) {
+            return IREmitterHelpers::emitMethodCallViaRubyVM(mcctx);
+        }
+
+        auto &cs = mcctx.cs;
+        auto &builder = mcctx.builder;
+        auto &var = mcctx.send->args[0].type;
+        if (!core::isa_type<core::LiteralType>(var)) {
+            return IREmitterHelpers::emitMethodCallViaRubyVM(mcctx);
+        }
+        auto lit = core::cast_type_nonnull<core::LiteralType>(var);
+        if (lit.literalKind != core::LiteralType::LiteralTypeKind::Symbol) {
+            return IREmitterHelpers::emitMethodCallViaRubyVM(mcctx);
+        }
+
+        auto varName = lit.asName(cs);
+        auto varNameStr = varName.shortName(cs);
+
+        auto *callCache = mcctx.getInlineCache();
+        auto *ivarCache = Payload::buildInstanceVariableCache(cs, varNameStr);
+        auto *cfp = Payload::getCFPForBlock(cs, builder, mcctx.irctx, mcctx.rubyBlockId);
+        auto *recv = mcctx.varGetRecv();
+        auto *ivarID = Payload::idIntern(cs, builder, varNameStr);
+        auto *value = Payload::varGet(cs, mcctx.send->args[1].variable, builder, mcctx.irctx, mcctx.rubyBlockId);
+
+        return builder.CreateCall(cs.getFunction("sorbet_vm_instance_variable_set"),
+                                  {callCache, ivarCache, cfp, recv, ivarID, value});
+    }
+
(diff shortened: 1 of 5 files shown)


froydnj:

→ https://go/builds/bui_KW5cZA1Oae7VL3
→ https://go/builds/bui_KW5csbuOLc7EZj

compiler/IREmitter/Payload/patches/vm_insnhelper.c
```diff
@@ -978,6 +979,24 @@ VALUE sorbet_vm_instance_variable_get(struct FunctionInlineCache *getCache, IVC
     return sorbet_callFuncWithCache(getCache, VM_BLOCK_HANDLER_NONE);
 }
 
+VALUE sorbet_vm_instance_variable_set(struct FunctionInlineCache *setCache, IVC varCache, rb_control_frame_t *reg_cfp,
+                                      VALUE recv, ID var, VALUE value) {
+    sorbet_vmMethodSearch(setCache, recv);
+    rb_method_definition_t *setDef = setCache->cd.cc.me->def;
+    if (setDef->type == VM_METHOD_TYPE_CFUNC &&
+        setDef->body.cfunc.func == sorbet_vm_Kernel_instance_variable_set_func()) {
```
elliottt, line 987: Is it worth wrapping this in `LIKELY`?
```suggestion
    if (LIKELY(setDef->type == VM_METHOD_TYPE_CFUNC &&
        setDef->body.cfunc.func == sorbet_vm_Kernel_instance_variable_set_func())) {
```

compiler/IREmitter/Payload/patches/vm_insnhelper.c
```diff
@@ -978,6 +979,24 @@ VALUE sorbet_vm_instance_variable_get(struct FunctionInlineCache *getCache, IVC
...
+        sorbet_vm_setivar(recv, var, value, varCache);
+        return value;
+    }
+
+    VALUE *sp = reg_cfp->sp;
+    *(sp + 0) = recv;
+    *(sp + 1) = rb_id2sym(var);
+    *(sp + 2) = value;
```
elliottt, line 995: Might be better to use `vm_env_write` here, as it'll be available in this context.

compiler/IREmitter/NameBasedIntrinsics.cc
```diff
@@ -732,6 +732,45 @@ class InstanceVariableGet : public NameBasedIntrinsicMethod {
...
+        auto &cs = mcctx.cs;
+        auto &builder = mcctx.builder;
+        auto &var = mcctx.send->args[0].type;
+        if (!core::isa_type<core::LiteralType>(var)) {
+            return IREmitterHelpers::emitMethodCallViaRubyVM(mcctx);
+        }
+        auto lit = core::cast_type_nonnull<core::LiteralType>(var);
+        if (lit.literalKind != core::LiteralType::LiteralTypeKind::Symbol) {
+            return IREmitterHelpers::emitMethodCallViaRubyVM(mcctx);
+        }
```
elliottt, line 753: What about supporting string literals as well? It looks like it's converted to a symbol automatically: https://ruby-doc.org/core-2.7.2/Object.html#method-i-instance_variable_set

Looking into `rb_obj_ivar_set` it seems like this might not be super straight forward, but it would be nice to handle in the future.

elliottt (approved):
Looks great, just a few comments!

compiler/IREmitter/NameBasedIntrinsics.cc
froydnj: I think it would be straightforward to handle in the future, but as strings don't come up for cases this patch is intended to accelerate, I'm comfortable leaving it this way for now.

compiler/IREmitter/Payload/patches/vm_insnhelper.c
froydnj: Isn't `vm_env_write` intened to be used for writing local variables for the current stack frame?  These are intended to be arguments for the call, not local variables.

compiler/IREmitter/Payload/patches/vm_insnhelper.c
froydnj: Ehhh...maybe?  My sense is that all the pointer chasing involved in method lookup and these checks are likely to swamp any benefit gained from branch layout here.  I'll defer this until later; I think this and some related things really want a `sorbet_method_cfunc_is` primitive along the lines of `vm_method_cfunc_is`.

froydnj (commented):
Thanks for the review!

compiler/IREmitter/Payload/patches/vm_insnhelper.c
elliottt: Good point. I remembered there being a helper for pushing values to the ruby stack, but looking at `vm_call0_body` it looks like such a thing doesn't exist.

PR #4813, opened 2021-11-01
Describe error code 5046

supermario:
Adds an error description for 5046.

Example: https://sorbet.run/#%23%20typed%3A%20strict%0A%23%20%20%20%20%20%20%20%20%5E%5E%5E%5E%20try%20changing%20this%20to%20strict%0A%0AT.let%28%7B%7D%2C%20Hash%29

### Motivation

The error "Malformed type declaration. Generic class without type arguments Hash" is confusing if you don't realise that `Hash`/`Array` aliases aren't available in `type: strict`. "Malformed" primes you to think you've typed something wrong.

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -686,9 +686,26 @@ class ChildTwo < T::Struct
 end
 ```
 
-## 5046 
+## 5046
 
-In `typed: strict` the `Array` and `Hash` aliases needs to be the expanded `T::Array[T.untyped]` or `T::Hash[T.untyped, T.untyped]` types respectively.
+Generic classes must be passed all their generic type arguments when being used
+as types. For example:
+
+```ruby
+T.let([], Array)              # error
+T.let([], T::Array[Integer])  # ok
+```
+
+Many classes in the standard library are generic classes ([see
+here](stdlib.md)), and must be passed type arguments, including `Array` and
+`Hash`. Any user-defined generic classes must similarly be provided type
+arguments when used.
+
+For legacy reasons relating to the intial rollout of Sorbet, this error is only
+reported at `# typed: strict` for standard library generic classes and `# typed:
+true` for all user-defined generic classes. (In an ideal world, it would have
+always been reported at `# typed: true`, and we might change this in the
+future.)
 
 ## 5047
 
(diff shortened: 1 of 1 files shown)


website/docs/error-reference.md
```diff
@@ -686,6 +686,10 @@ class ChildTwo < T::Struct
 end
 ```
 
+## 5046 
+
+In `typed: strict` the `Array` and `Hash` aliases needs to be the expanded `T::Array[T.untyped]` or `T::Hash[T.untyped, T.untyped]` types respectively.
+
```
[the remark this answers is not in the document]
supermario: That's a much better description. 😅

website/docs/error-reference.md
```diff
@@ -686,6 +686,27 @@ class ChildTwo < T::Struct
...
+here](stdlib.md)), and must be passed type arguments, including `Array` and
+`Hash`. Any user-defined generic classes must similarly be provided type
+arguments when used.
+
+For legacy reasons relating to the intial rollout of Sorbet, this error is only
+reported at `# typed: strict` for standard library generic classes and `# typed:
+true` for all user-defined generic classes. (In an ideal world, it would have
+always been reported at `# typed: true`, and we might change this in the
+future.)
+
```
jez, line 709:
````suggestion
## 5046

Generic classes must be passed all their generic type arguments when being used
as types. For example:

```ruby
T.let([], Array)              # error
T.let([], T::Array[Integer])  # ok
```

Many classes in the standard library are generic classes
([see here](stdlib.md)), and must be passed type arguments, including `Array`
and `Hash`. Any user-defined generic classes must similarly be provided type
arguments when used.

For legacy reasons relating to the intial rollout of Sorbet, this error is only
reported at `# typed: strict` for standard library generic classes and
`# typed: true` for all user-defined generic classes. (In an ideal world, it
would have always been reported at `# typed: true`, and we might change this in
the future.)

````

jez: linter requires it be formatted like this

supermario: Sorry @jez missed this, sorted now.

jez (approved):

Issue #4814, opened 2021-11-01
Gems marked as `require: false` are being loaded


jflinter:
#### Input

- Start a new project with the following Gemfile:
```rb
source 'https://rubygems.org'

gem 'sorbet'
gem 'ulid', '1.3.0', 'require' => false # A random gem I picked from my project that doesn't have any dependencies
```
- Run `bundle exec srb init`

#### Observed output

`sorbet/rbi/gems` contains `ulid.rbi`

#### Expected behavior

`sorbet/rbi/gems` should not contain `ulid.rbi`

- - -

I believe https://github.com/sorbet/sorbet/pull/2637 caused this regression - `require` is not passed into `Gemfile.lock`, so parsing the lockfile instead of the Gemfile makes `srb` think every gem is requred.

jez:
Yeah that looks like the right PR. I have no real preference, because Stripe doesn't use this code. @connorshea maybe you want to chime in?

I'm defaulting to reverting that PR, because it seems worse to require things that explicitly should not be required than it does to not require enough things (because you can ~easily hand write RBIs if something wasn't required aggressively enough).

Also @jflinter you may want to consider using https://github.com/Shopify/tapoica. People who use Sorbet outside of Stripe have reported that it works better in a lot of cases.

connorshea:
I use tapioca so I don't think this change would effect me anyway, but the intent was to load all gems because you need type signatures for all gems. Most of the time when you have a gem set to require: false, you still _use_ that gem it's just that it gets required manually when the gem is needed. And IIRC Sorbet got/gets confused by that.

So I'm not going to block such a change, but I do think it might be a bad idea to revert the change.

jflinter:
I'm using tapioca already to generate .rbi for my gems, but still depend on `srb rbi hidden-definitions` to pick up other hidden methods in our codebase. That command is going to call `require_everything` anyway though, so in practice I just get a ton of unneeded gem sigs in my `hidden.rbi` (that's what got me poking around here).

I agree that sigs should be generated for required gems and their dependencies, but per the original discussion in https://github.com/sorbet/sorbet/issues/1104 there also does need to be a way to indicate to sorbet that a gem should be ignored, and `require: false` seems like the right way to do that.

https://github.com/sorbet/sorbet/pull/4815 is my stab at fixing this, although I'm having trouble writing a test case for it because I can't seem to get the test suite running locally...

univerio:
[...]

PR #4815, opened 2021-11-01
Do not load gems marked as require: false

jflinter:
Fixes https://github.com/sorbet/sorbet/issues/4814 . Note that sigs will still be generated for gems' dependencies, just not for ones marked `require: false`.

### Test plan
There should probably be an automated test for this, I just haven't been able to get the test suite running locally on my M1 mac :(


jez (approved):
lgtm if tests pass

PR #4817, opened 2021-11-02
move is_a?/kind_of? to be a name-based intrinsic

froydnj:
### Motivation

`is_a?` gets used on untyped values fairly often, for whatever reason, and our symbol-based intrinsic implementation of `is_a?` was missing that.  This change lets us take advantage of it more often.

Future work would be identifying arguments where the type test can be done cheaply (e.g. `x.is_a?(String)`), but that would require some consideration of whether we want to bake in that behavior and ignore potential `is_a?` overrides.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_KWT1m0lkBpwnk2
→ https://go/builds/bui_KWT13KMt3WYVTL

elliottt (approved):

PR #4825, opened 2021-11-03
enable symbol-based intrinsics to avoid after-intrinsic processing

froydnj:
### Motivation

We have a hook for processing interrupts after an intrinsic because normal Ruby method sends will already check for interrupts, and we found that intrinsics that did C calls directly did not check for interrupts, leading to (IIRC) GC issues.

This is a good plan, but sometimes we don't want to bloat the code...or we know that we don't need to check for interrupts, or can delay emitting the check for interrupts.  This PR introduces a way of enabling `SymbolBasedIntrinsicMethod`s to signal that they don't need this extra processing, and I've added such signaling to what I think are relatively uncontroversial intrinsics: method definition, and various things associated with sigs.

(I noticed this bloating the code for static-init methods; there's a lot of code there to trim down, and this is but one way of addressing that.)

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/SymbolBasedIntrinsicMethod.h
+++ b/compiler/IREmitter/SymbolBasedIntrinsicMethod.h
@@ -40,6 +40,13 @@ public:
     // You should not typically be returning true from this.
     virtual bool skipFastPathTest(MethodCallContext &mcctx, core::ClassOrModuleRef potentialClass) const;
 
+    // Returns whether we should be calling Payload::afterIntrinsic to handle any
+    // post-intrinsic processing (e.g. checking whether VM-level interrupts have arrived).
+    // The default is to call.
+    //
+    // You should not typically be returning false from this.
+    virtual bool needsAfterIntrinsicProcessing() const;
+
     SymbolBasedIntrinsicMethod(Intrinsics::HandleBlock blockHandled) : blockHandled(blockHandled){};
     virtual ~SymbolBasedIntrinsicMethod() = default;
     static std::vector<const SymbolBasedIntrinsicMethod *> &definedIntrinsics(const core::GlobalState &gs);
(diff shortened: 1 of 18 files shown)


jez:
> (I noticed this bloating the code for static-init methods; there's a lot of code there to trim down, and this is but one way of addressing that.)

ooc, what is the downside to having bloat in static-init methods, other than just having less LLVM IR to read?

froydnj:
> > (I noticed this bloating the code for static-init methods; there's a lot of code there to trim down, and this is but one way of addressing that.)
> 
> ooc, what is the downside to having bloat in static-init methods, other than just having less LLVM IR to read?

More time in the LLVM optimizers running over a bunch of repetitive code that's only going to execute once.  I have Plans(tm) to try and mark some of the static-init-related stuff `__attribute__((cold))` or so, but having less code generally in them just seems like a good idea.

froydnj:
(The static-init methods are easily the biggest methods we compile in Stripe's codebase and therefore contribute the most to some of the slow compile times we're seeing.)

compiler/IREmitter/SymbolBasedIntrinsicMethod.h
```diff
@@ -40,6 +40,13 @@ class SymbolBasedIntrinsicMethod {
     // You should not typically be returning true from this.
     virtual bool skipFastPathTest(MethodCallContext &mcctx, core::ClassOrModuleRef potentialClass) const;
 
+    // Returns whether we should be calling Payload::afterIntrinsic to handle any
+    // post-intrinsic processing (e.g. checking whether VM-level interrupts have arrived).
+    // The default is to call.
+    //
+    // You should not typically be returning false from this.
```
elliottt, line 47: This is exactly what I was hoping to see, thanks for adding this comment :)

elliottt (approved):
I think this makes sense, good catch!

froydnj:

→ https://go/builds/bui_KXeiGfkGWUPOdL
→ https://go/builds/bui_KXeicvIL68IOW2

froydnj:

→ https://go/builds/bui_KYnZnv9B6miEC2
→ https://go/builds/bui_KYnZXe3uYOrBoL

PR #4830, opened 2021-11-04
Switch to tracking final method modules by object_id

elliottt:
Switch from keying internal final metadata on `Module` values to object ids.

### Motivation

We encountered an internal issue where using `.hash` during the `extended` hook caused issues, and keying the final map on `object_id` avoids the problem.

### Test plan

See included automated tests.


elliottt:
Successful build on Stripe CI: https://go/builds/bui_KXFSx8D0GvcS5p

jez (approved):
thanks!

PR #4833, opened 2021-11-05
Add `Pathname#glob` to stdlib. Fix second argument.

k3rni:
Previously, Sorbet knew only about `Pathname.glob` the class-method. This adds support for the instance-method version. It is nearly identical, but globs from PWD instead. 

The second argument to both (and to Dir.glob which is used underneath) is a logical-OR of flags
defined in File, e.g. File::FNM_DOTMATCH which makes `*` match filenames beginning with a dot.

### Motivation

Annoyed by sorbet pointing out that `Pathname#glob` is not a method.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/pathname.rbi
+++ b/rbi/stdlib/pathname.rbi
@@ -245,19 +245,19 @@ class Pathname < Object
   sig do
     params(
         p1: T.any(String, Pathname),
-        p2: String,
+        p2: Integer,
     )
     .returns(T::Array[Pathname])
   end
   sig do
     params(
         p1: T.any(String, Pathname),
-        p2: String,
+        p2: Integer,
         blk: T.proc.params(arg0: Pathname).void,
     )
     .returns(NilClass)
   end
-  def self.glob(p1, p2=T.unsafe(nil), &blk); end
+  def self.glob(p1, p2=0, &blk); end
 
   def initialize(p); end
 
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_KYgxeuoHXJmgbr
→ https://go/builds/bui_KYgyDFYODuYMu5

froydnj (changes requested):
Thanks for the PR!  It looks like there are some test failures that are easily fixed by deleting some code; can you take a look at those and fix the errors, please?

froydnj:
> We have a policy of testing changes to Sorbet against Stripe's codebase before merging them. I've kicked off a test run for the current PR. When the build finishes, I'll share with you whether or how it failed. Thanks!
> 
> 
> → https://go/builds/bui_KYgxeuoHXJmgbr → https://go/builds/bui_KYgyDFYODuYMu5

These failed for very similar reasons to the initial test failures seen in this PR (Sorbet now knows about `Pathname#glob` whereas it didn't before), but a fixup build was green.

k3rni:
@froydnj I am not sure what to change - running `bazel test test` passes for me locally both as-is and with the last commit which removes the lines. The links you've posted don't work for me (I guess they're internal to Stripe).

jez:

→ https://go/builds/bui_KYl9WEjuznqEib
→ https://go/builds/bui_KYlATcMdiwTYzT

froydnj (approved):
> @froydnj I am not sure what to change - running `bazel test test` passes for me locally both as-is and with the last commit which removes the lines. The links you've posted don't work for me (I guess they're internal to Stripe).

Sorry about that!  I meant that you could see test failures at https://buildkite.com/sorbet/sorbet/builds/18389#b582f93c-38cd-4e19-b039-6de452714a5b, which ideally would have pointed out the lines to modify.  (The output also points out the Bazel tests to run, but that's not at all obvious; I'll remember to point that out to people in the future.)  I didn't expect you to view Stripe-internal links. :)

It looks like you found the correct lines to modify!

PR #3351, opened 2020-08-07
Add overload for Array#sample for no-argument case

will:
Before
``` ruby
T.reveal_type [3,5].sample
# Revealed type: T.nilable(T.any(Integer, T::Array[Integer]))
T.reveal_type [3,5].sample(2)
# Revealed type: T.nilable(T.any(Integer, T::Array[Integer]))
```

After
``` ruby
T.reveal_type [3,5].sample
# Revealed type: T.nilable(Integer)
T.reveal_type [3,5].sample(2)
# Revealed type: T.nilable(T.any(Integer, T::Array[Integer]))
```

[playground for fix](https://sorbet.run/#%23%20typed%3A%20strict%0A%0A%0Aclass%20Array%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7B%20returns%20T.nilable(Elem)%20%7D%0A%20%20sig%20do%0A%20%20%20%20params(%0A%20%20%20%20%20%20%20%20arg0%3A%20Integer%2C%0A%20%23%20%20%20%20%20%20%20random%3A%20Random%3A%3AFormatter%2C%0A%20%20%20%20)%0A%20%20%20%20.returns(T.any(T.nilable(Elem)%2C%20T%3A%3AArray%5BElem%5D))%0A%20%20end%0A%20%20def%20sample(arg0%3DT.unsafe(nil)%2C%20random%3A%20T.unsafe(nil))%3B%20end%0Aend%0A%0A%0AT.reveal_type%20%5B3%2C5%5D.sample%0AT.reveal_type%20%5B3%2C5%5D.sample(2)%0A%0A%0A%0A%0A%0A%0A%0A%0A%0A%0A%0A%0A%0A)

### Motivation
I couldn't do 
``` ruby
maby_empty_array.sample || default
```
when sorbet incorrectly thought `#sample` with no arguments could maybe return an array.

I saw that `first` worked, so then found out that you can overlaod sigs (which seems to be undocumented? but I didn't search the docs that hard, maybe you all use a different term)

[...]

PR #2530, opened 2020-01-23
Allow Array#sample to have more sensitive types when not passing in any parameters

AlexEvanczuk:
`Array#sample` could return an array of things if an integer is passed in. It returns a single element if no parameter is passed in:
https://ruby-doc.org/core-2.4.1/Array.html#method-i-sample

### Motivation
This allows us to use `.sample` and have more constrained types.

This sorbet.run illustrates the issue: https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns(Integer)%20%7D%0A%20%20def%20bar%0A%20%20%20%5B1%2C%202%5D.sample%0A%20%20end%0Aend%0A%0Adef%20main%0A%20%20A.new.bar%0Aend

### Test plan
See included automated tests.

PR #4836, opened 2021-11-06
Add docs for how `Array#sample` is surprising

ghiculescu:
This has come up a few times (https://github.com/sorbet/sorbet/pull/3351 https://github.com/sorbet/sorbet/pull/2530 https://github.com/sorbet/sorbet/pull/2248), and anecdotally I get confused about it fairly often before remembering those PRs. Adding some docs to explain why `Array#sample` has a weird sig, and what you can do to decrease the annoyance if you are displined about it.

### DIFF
--- a/website/docs/faq.md
+++ b/website/docs/faq.md
@@ -160,7 +160,7 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
 [0, 1, 2].fetch(3) # IndexError: index 3 outside of array bounds
 ```
 
-## Sigs are vague for std lib methods that accept keyword arguments & have multiple return types
+## Sigs are vague for stdlib methods that accept keyword arguments & have multiple return types
 
 You might notice this when calling `Array#sample`, `Pathname#find`, or other std
 lib methods that accept a keyword argument and can have different return types
--- a/website/docs/faq.md
+++ b/website/docs/faq.md
@@ -170,7 +170,7 @@ based on arguments:
 T.reveal_type([1, 2, 3].sample) # Revealed type: T.nilable(T.any(Integer, T::Array[Integer]))
 ```
 
-The sig in Sorbet's std lib is quite wide, since it has to cover every possible
+The sig in Sorbet's stdlib is quite wide, since it has to cover every possible
 return type. Sorbet does not have good support for this for methods that accept
 keyword arguments. [#37](https://github.com/sorbet/sorbet/issues/37) is the
 original report of this.
--- a/website/docs/faq.md
+++ b/website/docs/faq.md
@@ -160,23 +160,39 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
 [0, 1, 2].fetch(3) # IndexError: index 3 outside of array bounds
 ```
 
-## The types for `Array#sample` are weird
+## Sigs are vague for std lib methods that accept keyword arguments & have multiple return types
+
+You might notice this when calling `Array#sample`, `Pathname#find`, or other std
+lib methods that accept a keyword argument and can have different return types
+based on arguments:
 
 ```ruby
 T.reveal_type([1, 2, 3].sample) # Revealed type: T.nilable(T.any(Integer, T::Array[Integer]))
 ```
 
-`Array#sample` has a complex method signature, in that its return type differs
-based on the caller array's size and a positional argument, _and_ it accepts a
-keyword argument. This is unlikely to be improved, see
-[#2530](https://github.com/sorbet/sorbet/pull/2530) and
-[#2248](https://github.com/sorbet/sorbet/pull/2248) for past discussion.
+The sig in Sorbet's std lib is quite wide, since it has to cover every possible
+return type. Sorbet does not have good support for this for methods that accept
+keyword arguments. [#37](https://github.com/sorbet/sorbet/issues/37) is the
+original report of this.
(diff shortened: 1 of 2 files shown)


website/docs/faq.md
```diff
@@ -160,6 +160,53 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
 [0, 1, 2].fetch(3) # IndexError: index 3 outside of array bounds
 ```
 
+## The types for `Array#sample` are weird
```
jez, line 163: Can you generalize this heading to say something about how overloads + keyword arguments are unsupported, moving Array#sample to a specific example of the general behavior, and avoid using the term “weird” instead of something more precise?

jez (changes requested):

website/docs/faq.md
ghiculescu: Can do. Apart from sample, are there other common examples of this worth calling out?

jez: https://github.com/sorbet/sorbet/blob/b0537331e412423cbc494d74bd9d39144d39b837/rbi/core/kernel.rbi#L761

Also #37 mentions Dir.glob (#37 is the more general issue to link to)

jez: https://github.com/sorbet/sorbet/pull/4214

Another one

ghiculescu: Thanks!! I generalised it based on this feedback, and added some more examples.

website/docs/faq.md
```diff
@@ -160,6 +160,90 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
 [0, 1, 2].fetch(3) # IndexError: index 3 outside of array bounds
 ```
 
+## Sigs are vague for std lib methods that accept keyword arguments & have multiple return types
```
Morriar, line 163: I think we generally use `stdlib` or `standard library` see [here](https://sorbet.org/docs/faq#it-looks-like-sorbets-types-for-the-stdlib-are-wrong) for example.

```suggestion
## Sigs are vague for stdlib methods that accept keyword arguments & have multiple return types
```

website/docs/faq.md
```diff
@@ -160,6 +160,90 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
...
+
+You might notice this when calling `Array#sample`, `Pathname#find`, or other std
+lib methods that accept a keyword argument and can have different return types
```
Morriar, line 166:
```suggestion
You might notice this when calling `Array#sample`, `Pathname#find`, or other
stdlib methods that accept a keyword argument and can have different return types
```

website/docs/faq.md
```diff
@@ -160,6 +160,90 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
...
+based on arguments:
+
+```ruby
+T.reveal_type([1, 2, 3].sample) # Revealed type: T.nilable(T.any(Integer, T::Array[Integer]))
+```
+
+The sig in Sorbet's std lib is quite wide, since it has to cover every possible
```
Morriar, line 173:
```suggestion
The sig in Sorbet's stdlib is quite wide, since it has to cover every possible
```

website/docs/faq.md
```diff
@@ -160,6 +160,90 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
...
+
+  sig do
+    params(arg0: Integer, random: Random::Formatter)
+    .returns(Elem) # or T.nilable(Elem), to also support empty arrays
+  end
+  def sample(arg0=T.unsafe(nil), random: T.unsafe(nil)); end
+end
+```
+
+> Overriding std lib RBIs can make type checking less safe, since Sorbet will
```
Morriar, line 225:
```suggestion
> Overriding stdlib RBIs can make type checking less safe, since Sorbet will
```

website/docs/faq.md
```diff
@@ -160,6 +160,90 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
...
+> now have an incorrect understanding of how the std lib behaves.
```
Morriar, line 226:
```suggestion
> now have an incorrect understanding of how the stdlib behaves.
```

website/docs/faq.md
```diff
@@ -160,6 +160,90 @@ out-of-bounds. If you would rather raise an exception than handle `nil`, use the
...
+
+Another alternative is to define new methods that are stricter about arguments,
+and use these in place of std lib methods:
```
Morriar, line 229:
```suggestion
and use these in place of stdlib methods:
```

website/docs/tenum.md
```diff
@@ -35,10 +35,13 @@ can be used:
 ```ruby
 sig {returns(Suit)}
 def random_suit
-  Suit.values.sample
+  T.cast(Suit.values.sample, Suit)
 end
 ```
 
+> The `T.cast` is necessary because of how
+> [Sorbet behaves with `Array#sample`](faq.md#sigs-are-vague-for-std-lib-methods-that-accept-keyword-arguments--have-multiple-return-types).
```
Morriar, line 43:
```suggestion
> [Sorbet behaves with `Array#sample`](faq.md#sigs-are-vague-for-stdlib-methods-that-accept-keyword-arguments--have-multiple-return-types).
```

jez (approved):

ghiculescu:
sorry @jez or @Morriar just had one more lint error. should be green now.

PR #4837, opened 2021-11-07
Infer types of instance variables from literals in initialize

### LINKED ISSUE #3513: Infer type annotations for instance variables
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20X%20%3D%200%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20initialize%0A%20%20%20%20%40y%20%3D%200%0A%20%20end%0Aend)

```ruby
# typed: strict

class A
  extend T::Sig

  X = 0

  sig {void}
  def initialize
    @y = 0
  end
end
```

#### Observed output

```
editor.rb:10: Use of undeclared variable @y https://srb.help/6002
    10 |    @y = 0
            ^^
Errors: 1
```

#### Expected behavior

We should infer the annotation for `@x` because it is a literal type. We should use the same type defaulting rules that we use for constants (as depicted) to also handle instance variables.

- - -

- This should probably only work for instance variable assignments that happen inside `initialize` or the top level of a class.
- Ideally we would consolidate the logic that handles constants to handle instance variables.
  To this end, we can probably move the logic to be done in a rewriter pass.
  The rewriter pass would insert a `T.let` inside assignments to either constants or instance variables when the RHS is  a literal.
  (As opposed to now, where it is done in the internals of namer/resolver.)

djudd:
This PR extends the existing Initializer rewriter pass to infer the types of instance variables based on literal assignments, not just assignments from params.

It's worth calling out that surprisingly-to-me this isn't 100% backwards compatible: cases where the type inferred from the literal is inconsistent with a type explicitly assigned to the instance variable somewhere outside of the initializer can break. See the change to the `attr_checked.rb` test. That seems ok to me since it's an edge case, where plausibly something unintended is going on, and it's easy to work around it by just adding an explicit `T.let` in the initializer if it really is what you want, but it might require some updates to Stripe code that I'm not going to tackle while on leave.

### Motivation
Fixes #3513

I didn't end up taking the suggested approach of consolidating with the logic that handles constants. I took a stab at it, starting by cargo-culting the existing Regexp rewriter, but I think it's actually a bit complex. If you try to use plain `lit->value` as the type inside the T.let, you run into "Unsupported literal in type syntax" errors, because we don't allow `T.let(..., :foo)`, and the rewriter pass seems to happen before that validation. But if you try to lift the type up to the class of the literal, you're actually throwing away type information that's currently still available later in the pipeline (and a bunch of `.exp` tests fail).

I also didn't handle class-instance-level or class-level variables, which do seem worth doing as a followup.

I don't really know how to write C++ so please keep an eye out for dumb (or just unidiomatic) language things.

### Test plan

See included automated tests.


test/testdata/rewriter/attr_checked.rb
```diff
@@ -27,7 +27,10 @@ class TestAttr
   attr_accessor :str8 # error: This function does not have a `sig`
 
   sig {params(str9: T.nilable(String)).returns(T.nilable(String)).checked(:never)}
-  attr_writer :str9
+  attr_writer :str9 # error: Redeclaring variable `@str9` with mismatching type
```
djudd, line 30: This is the compatibility break

test/testdata/rewriter/initializer.rb
```diff
@@ -46,6 +46,35 @@ def foo
...
+  end
+
+  sig {void}
+  def foo
+    T.reveal_type(@int) # error: Revealed type: `Integer`
+    T.reveal_type(@str) # error: Revealed type: `String`
+    T.reveal_type(@sym) # error: Revealed type: `Symbol`
+    T.reveal_type(@float) # error: Revealed type: `Float`
+    T.reveal_type(@nil) # error: Revealed type: `NilClass`
+    T.reveal_type(@true) # error: Revealed type: `TrueClass`
```
djudd, line 72: Way out of scope, but IMHO we should change type inference globally so that we always infer `T::Boolean` from bare literal `true` and `false` values instead of inferring the too-specific `TrueClass` or `FalseClass`; how often do you actually want a boolean that can only ever have one value?

jez:

→ https://go/builds/bui_KYlGRnynIXxIiM
→ https://go/builds/bui_KYlHbc1C4j83tf

test/testdata/rewriter/initializer.rb
jez: https://github.com/sorbet/sorbet/blob/master/test/testdata/infer/true_false_knowledge.rb#L3-L15

If you can get it to work in a way that doesn't break this test case, then be my guest. I think that this is unfortunately going to be a limitation of Sorbet's inference algorithm for a while.

djudd: Oh ugh. I don't know anything about compilers really but that does seem like something where it'd be hard to get the best of both worlds without doing something way fancier than the AST rewriter for type inference.

djudd: (Appreciate getting the chance to learn a bit about this stuff thanks to the good-first-issue tag!)

jez:
> If you try to use plain `lit->value` as the type inside the T.let, you run into "Unsupported literal in type syntax" errors, because we don't allow `T.let(..., :foo)`

Right, the logic would have to unwrap the literal type to a class type. That's easy, you can see that change here: #4840.

In other news, this change introduces about a hundred type errors on Stripe's codebase, which is workable, not quick to fix.

djudd:
Is losing that information not a problem for compiler optimizations or anything?

It occurred to me when thinking about the TrueClass-vs-T::Boolean thing that the constant and instance-variable situations are maybe a bit different in this respect. With a constant it kind of makes sense to infer the more specific type because it's, y'know, constant. But the fact that something is an instance variable means you can probably assume that it can take different values at least under some circumstances. So it wouldn't necessarily feel awful to me to just infer T::Boolean here but leave the constant case inferring TrueClass or FalseClass. Similarly, we should arguably not be inferring anything beyond T.untyped when initializing to nil here. Neither really addresses the https://github.com/sorbet/sorbet/pull/4368 issue more broadly but should that stop us from a possible quality-of-life improvement here? Also might reduce some of the type errors introduced. WDYT?

jez:

→ https://go/builds/bui_Kh0qlAJaFmQ1p2
→ https://go/builds/bui_Kh0qa8wd4olQKt

mrcasals:
@jez any update to this?

djudd:
I don't work at Stripe anymore so I'm unfortunately unable to check whether there are still a bunch of type errors from this, sorry. Seems likely though if nobody has had time to prioritize it.

PR #4838, opened 2021-11-08
Fix link typo

openbl:
### Motivation

Found a broken link, which appears to be the only of its kind. I'm not sure if the site builder will remove ".md" extensions though.

### Test plan

n/a

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -697,7 +697,7 @@ T.let([], T::Array[Integer])  # ok
 ```
 
 Many classes in the standard library are generic classes
-([see here](stdlib.md)), and must be passed type arguments, including `Array`
+([see here](stdlib-generics.md)), and must be passed type arguments, including `Array`
 and `Hash`. Any user-defined generic classes must similarly be provided type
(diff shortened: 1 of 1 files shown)


jez (approved):
Thanks!

website/docs/error-reference.md
```diff
@@ -697,7 +697,7 @@ T.let([], T::Array[Integer])  # ok
 ```
 
 Many classes in the standard library are generic classes
-([see here](stdlib.md)), and must be passed type arguments, including `Array`
+([see here](stdlib-generics.md)), and must be passed type arguments, including `Array`
 and `Hash`. Any user-defined generic classes must similarly be provided type
 arguments when used.
```
jez, line 702:
```suggestion
([see here](stdlib-generics.md)), and must be passed type arguments, including
`Array` and `Hash`. Any user-defined generic classes must similarly be provided
type arguments when used.
```

jez: you'll need to format things like this to pass our markdown linter

jez (approved):

PR #4839, opened 2021-11-08
reuse input trees vector in Resolver::resolveSigs

froydnj:
### Motivation

Per the PR title.  This saves some memory churn, since we can reuse the allocation of `trees`.  (We also weren't reserving space in `combinedTrees`, either.)  Notice that we already do the same thing in ResolveConstantsWalk.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -3086,6 +3086,8 @@ ast::ParsedFilesOrCancelled Resolver::resolveSigs(core::GlobalState &gs, vector<
         inputq->push(move(tree), 1);
     }
 
+    trees.clear();
+
     workers.multiplexJob("resolveSignaturesWalk", [&gs, inputq, outputq]() -> void {
         ResolveSignaturesWalk walk;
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_KYmAoqsoRB4IPw
→ https://go/builds/bui_KYmAYiKShYOlpM

resolver/resolver.cc
```diff
@@ -3144,7 +3145,7 @@ ast::ParsedFilesOrCancelled Resolver::resolveSigs(core::GlobalState &gs, vector<
         }
     }
 
-    return move(combinedTrees);
+    return move(trees);
```
elliottt, line 3148: Is the move necessary here?

elliottt (approved):
Looks great, good call on avoiding the reallocation 👍

resolver/resolver.cc
froydnj: Nope, fixed!

PR #4841, opened 2021-11-09
Run props pass during autogen

aisamanra:
This allows the props DSL to run under autogen, something which was previously disallowed.

### Motivation
The props rewriter [introduces constant references](https://github.com/sorbet/sorbet/blob/master/rewriter/Prop.cc#L170-L179) which we want to include in the `DependencyDB`. We can see that testing this internally produces the constant references we want (for Stripe employees, c.f. the build failure [go/builds/bui_KZ710mMhNsXMgM](https://go/builds/bui_KZ710mMhNsXMgM): this is what we want to have happen, because that packaging failure indicates we're correctly seeing the result of this rewrite applying).

### Test plan

See included automated tests. Also tested over Stripe and ensured that including this rewriter pass does not regress autogen times.

### DIFF
--- a/rewriter/Prop.cc
+++ b/rewriter/Prop.cc
@@ -547,9 +547,6 @@ vector<ast::ExpressionPtr> mkTypedInitialize(core::MutableContext ctx, core::Loc
 } // namespace
 
 void Prop::run(core::MutableContext ctx, ast::ClassDef *klass) {
-    if (ctx.state.runningUnderAutogen) {
-        return;
-    }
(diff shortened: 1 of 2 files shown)


aisamanra:

→ https://go/builds/bui_KZ6qeUKwOnltBQ
→ https://go/builds/bui_KZ6qggyYZ3wO4l

froydnj (approved):
> Also tested over Stripe and ensured that including this rewriter pass does not regress autogen times.

Thank you for calling this out!  This was the only question I was going to ask.  :shipit:

PR #4843, opened 2021-11-09
always keep the original dsl_{required,optional} send

froydnj:
### Motivation

#4639 kept around the original send when rewriting `DSLBuilder` methods, but it didn't go far enough: for the compiler's purposes, we always need to keep the original send around, even if the getter and setter are both skipped (!) because there might be other arguments to the send that need to take effect at runtime.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_KZ7oPRYDHM3Eqr
→ https://go/builds/bui_KZ7ol6lvP3jUwJ

elliottt (approved):
Makes sense to me!

PR #4844, opened 2021-11-09
more correctly dedent strings

froydnj:
### Motivation

For the included testcase in this PR, Sorbet was parsing things as:

```
s(:begin,
  s(:class,
    s(:const, nil, :Session), nil,
    s(:begin,
      s(:assign,
        s(:casgn, nil, :MAX_OBJS),
        s(:int, "849201")),
      s(:assign,
        s(:casgn, nil, :MAX_OBJS_MSG),
        s(:dstr,
          s(:str, "There are more than "),
          s(:begin,
            s(:const, nil, :MAX_OBJS)),
          # BUG: we dropped a space here
          s(:str, "objects!  Fire the missiles!
"))))),
  s(:send, nil, :p,
    s(:const,
      s(:const, nil, :Session), :MAX_OBJS_MSG)))
```

This not-quite-correct parsing of heredocs is fine for the static checker, but not so much for the compiler.

This PR cargo-cults the dedent logic from whitequark's parser:

https://github.com/whitequark/parser/blob/b87613d2b99733e1400f076675daf501c19ede93/lib/parser/builders/default.rb#L358-L380
https://github.com/whitequark/parser/blob/b87613d2b99733e1400f076675daf501c19ede93/lib/parser/lexer/dedenter.rb#L36-L79

into our parser.

### Test plan

There's a compiler testcase, which is fine, but every other testcase we have passes, which was surprising to me -- I would have expected some whitequark tests to transition from disabled to enabled.  Or do I have to reimport the upstream parser to add any necessary testcases?  Not really sure what to do here.

cc @vinistock who has worked on this stuff in the past.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -33,7 +33,14 @@ namespace sorbet::parser {
 string Dedenter::dedent(string_view str) {
     string out;
 
-    auto lines = absl::StrSplit(str, "\\\n");
+    vector<string_view> lines = absl::StrSplit(str, "\\\n");
+    if (!at_line_begin && !lines.empty()) {
+        out.append(lines[0]);
+        if (lines.size() > 1) {
+            out.push_back('\n');
+        }
+        lines.erase(lines.begin());
+    }
 
     for (auto line : lines) {
         spacesToRemove = dedentLevel;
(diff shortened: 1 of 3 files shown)


froydnj:

→ https://go/builds/bui_KZ7dZUc37p1dHu
→ https://go/builds/bui_KZ7dw8SQ01XHjM

froydnj:

→ https://go/builds/bui_KZ9qYCQ2HTxY05
→ https://go/builds/bui_KZ9qyCRtaTy1I4

froydnj:
The initial test build linked above were green, but an actual push that tested the compiler failed

I added another compiler testcase and https://go/builds/bui_KZA1UNurKW75Cy should be good.

aprocter-stripe (approved):
I think I believe that the cargo-culting from the whitequark parser is good, and the test looks good too.

PR #4845, opened 2021-11-09
Only emit <blockBreak> when inside a block

elliottt:
`break` inside of a while loop doesn't behave like break in a block: it only affects control flow. This PR implements this distinction by only emitting the `<blockBreak>` intrinsic when the use is directly inside of a block.

### Motivation

Correctly compiling uses of `break` in while loops.

### Test plan

See included automated tests.


jez (approved):

elliottt:

→ https://go/builds/bui_KZCmKQ1MWowxzZ
→ https://go/builds/bui_KZCmAJ4nCj9CCH

PR #4848, opened 2021-11-10
check types on rest/splat args properly

froydnj:
### Motivation

When we have something like:

```ruby
sig {params(x: T::Hash[T.untyped, T.untyped]).void}
def f(*x)
  ...
end
```

sorbet-runtime checks that each element of the rest arg is a hash.  The compiler was checking that the rest arg itself was a hash, which obviously didn't work all that well.  This PR type checks rest args properly.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/IREmitter.cc
+++ b/compiler/IREmitter/IREmitter.cc
@@ -754,8 +754,13 @@ void emitUserBody(CompilerState &base, cfg::CFG &cfg, const IREmitterContext &ir
                     auto var = Payload::varGet(cs, local, builder, irctx, 0);
                     if (auto &expectedType = argInfo.type) {
                         auto description = fmt::format("Parameter '{}'", bind.bind.variable.toString(cs, cfg));
-                        ENFORCE(!argInfo.flags.isBlock);
-                        IREmitterHelpers::emitTypeTest(cs, builder, var, expectedType, description);
+                        if (argInfo.flags.isRepeated && !argInfo.flags.isKeyword) {
+                            // Signature types for rest arguments apply to each individual element of
+                            // the rest arg, not the actual argument itself.
+                            IREmitterHelpers::emitTypeTestForRestArg(cs, builder, var, expectedType, description);
+                        } else {
+                            IREmitterHelpers::emitTypeTest(cs, builder, var, expectedType, description);
+                        }
                     }
                 },
                 [&](cfg::LoadYieldParams &i) {
--- a/compiler/IREmitter/Payload/codegen-payload.c
+++ b/compiler/IREmitter/Payload/codegen-payload.c
@@ -623,6 +623,11 @@ int sorbet_rubyArrayLen(VALUE array) {
 }
 
 SORBET_INLINE
+VALUE sorbet_rubyArrayAref(VALUE array, int i) {
+    return RARRAY_AREF(array, i);
+}
+
+SORBET_INLINE
 const VALUE *sorbet_rubyArrayInnerPtr(VALUE array) {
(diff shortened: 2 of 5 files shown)


froydnj:
For completeness, we ascribe the type to each element of the rest/splat arg here:

https://github.com/sorbet/sorbet/blob/16aac1a5dc93d0a051a4d4678bff520be07ab4de/gems/sorbet-runtime/lib/types/private/methods/signature.rb#L178-L179

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -622,6 +622,11 @@ int sorbet_rubyArrayLen(VALUE array) {
     return RARRAY_LEN(array);
 }
 
+SORBET_INLINE
+VALUE sorbet_rubyArrayAref(VALUE array, int i) {
+    return RARRAY_AREF(array, i);
+}
+
```
jez, line 629: Do you want to add a note here that this differs from the C function implementing the `Array#[]` intrinsic because this basically treats `array` as a C array and directly indexes it without bounds checking?

(Or maybe: change the name of the `sorbet_` wrapper to make that more clear, like `sorbet_rubyArrayArefUnchecked`?)

compiler/IREmitter/IREmitter.cc
```diff
@@ -751,6 +751,10 @@ void emitUserBody(CompilerState &base, cfg::CFG &cfg, const IREmitterContext &ir
                         auto description = fmt::format("Parameter '{}'", bind.bind.variable.toString(cs, cfg));
                         if (argInfo.flags.isBlock) {
                             IREmitterHelpers::emitTypeTestForBlock(cs, builder, var, expectedType, description);
+                        } else if (argInfo.flags.isRepeated && !argInfo.flags.isKeyword) {
+                            // Signature types for rest arguments apply to each individual element of
+                            // the rest arg, not the actual argument itself.
+                            IREmitterHelpers::emitTypeTestForRestArg(cs, builder, var, expectedType, description);
```
jez, line 757: Do we have the same problem with `**kwargs`?

For those, sorbet-runtime will check that for each keyword arg, the value has the mentioned type

```ruby
sig {params(kwargs: Integer).void}
def foo(**kwargs)
end

foo(a: 1, b: 2, c: 'nope')
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28kwargs%3A%20Integer%29.void%7D%0Adef%20foo%28**kwargs%29%0Aend%0A%0Afoo%28a%3A%201%2C%20b%3A%202%2C%20c%3A%20'nope'%29)

jez (approved):

compiler/IREmitter/IREmitter.cc
froydnj: I haven't tested it, but I very much imagine that we do.  I can write a disabled testcase for this tomorrow.  I think fixing the kwsplat case is a little more complicated because iterating over a hash table would require a separate function that the typecheck needs to live in.

(Also not sure how many people write types on their kwsplat arguments.)

compiler/IREmitter/Payload/codegen-payload.c
froydnj: Changing the name sounds like a good idea, I'll do that.

froydnj (commented):
Thanks for the review!

compiler/IREmitter/IREmitter.cc
froydnj: Testcase in #4860.

PR #4849, opened 2021-11-11
fix to_s method in kernel.rbi to accept an Integer as an optional arg

oneandonlymike:
### Motivation

The `to_s` method in the `kernel.rbi` was not aware of the fact that passing an Integer as an optional argument is valid https://apidock.com/ruby/v2_6_3/Integer/to_s. This results in code like [this example in the Sorbet Playground](https://sorbet.run/#%23%20typed%3A%20true%0A%0Arand%2810**100%29.to_s%2836%29) failing a type check. 

You can see this code is valid here:
```
irb(main):003:0> rand(10**100).to_s
=> "8235066421852743943633840297423609396022289043887726302077545134757967651325206712046158158536019151"
irb(main):004:0> rand(10**100).to_s(36)
=> "gaawupbkuv0xfpf33ztiiv7vsgg3186f83mewg3r274irs2dm00mnahsq2uz8mtf"
irb(main):005:0> 
```

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -645,8 +645,8 @@ module Kernel
   end
   def to_enum(method=T.unsafe(nil), *args, &blk); end
 
-  sig {returns(String)}
-  def to_s(); end
+  sig {params(arg: Integer).returns(String)}
+  def to_s(*arg); end
 
   sig {returns(T.self_type)}
   def trust(); end
--- a/test/testdata/rbi/kernel.rb
+++ b/test/testdata/rbi/kernel.rb
@@ -19,6 +19,10 @@ T.assert_type!(Kernel.rand(1.0..1.0), Float)
 T.assert_type!(Kernel.rand(36**8), Numeric)
 T.assert_type!(Kernel.rand((36**8..36**9)), Numeric)
 
+# to_s
+T.assert_type!(10.to_s, String)
+T.assert_type!(10.to_s(36), String)
+
 # make sure we don't regress and mark `loop` as returning `nil`
 x = loop {break 1}
 T.assert_type!(x, Integer)


oneandonlymike:
Moved this back to draft as it looks like this is more of an issue with the rbi expecting `rand` to return a `Numeric` when it can actually return an `Integer`.

rbi/core/kernel.rbi
```diff
@@ -645,8 +645,8 @@ module Kernel
   end
   def to_enum(method=T.unsafe(nil), *args, &blk); end
 
-  sig {returns(String)}
-  def to_s(); end
+  sig {params(arg: Integer).returns(String)}
+  def to_s(*arg); end
```
froydnj, line 649: I think you want to be more precise here, since `x.to_s(5, 6)` would be a runtime error that would not be caught by Sorbet.  I know the docs say `*arg`, but I think that's an implementation detail of how the docs are generated.

This probably wants to be

```ruby
sig {params(base: Integer).returns(String)}
def to_s(base=10); end
```

froydnj (commented):
> Moved this back to draft as it looks like this is more of an issue with the rbi expecting `rand` to return a `Numeric` when it can actually return an `Integer`.

`Integer` is a subclass of `Numeric`, so this seems fine?

oneandonlymike:
> Integer is a subclass of Numeric, so this seems fine?

Mainly the issue that was pointed out to me is that `Numeric` inherits from `Object` and does not implement `to_s`. Which means that indeed to_s does not accept an argument. `Integer#to_s` does accept an argument, but `Numeric#to_s` does not.

So this error _may_ actually just be fixed by adjusting the return types from `rand` to also include `Integer`.

froydnj:
> > Integer is a subclass of Numeric, so this seems fine?
> 
> Mainly the issue that was pointed out to me is that `Numeric` inherits from `Object` and does not implement `to_s`. Which means that indeed `to_s` does not accept an argument. `Integer#to_s` does accept an argument, but `Numeric#to_s` does not.
> 
> So this error _may_ actually just be fixed by adjusting the return types from `rand` to also include `Integer`.

Sorbet already knows that [`rand` with an integer argument returns an integer](https://sorbet.run/#%23%20typed%3A%20true%0AT.reveal_type%28rand%28100%29%29).

oneandonlymike:
It looks like Sorbet actually thinks it is a [Numeric if you make it more complex](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aa%20%3D%20rand%2810**100%29%0AT.reveal_type%28a%29) like the example we had where this came up.

You can even see [in this example](https://sorbet.run/#%23%20typed%3A%20true%0A%0Arand%28100%29.to_s%2836%29%0A%0Arand%2810**100%29.to_s%2836%29) that in yours where you displayed that Sorbet does in fact know it is an Integer, it does not throw an error about too many args, but in ours where it believes it is a `Numeric` it does throw the error.

froydnj:
> It looks like Sorbet actually thinks it is a [Numeric if you make it more complex](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aa%20%3D%20rand%2810**100%29%0AT.reveal_type%28a%29) like the example we had where this came up.
> 
> You can even see [in this example](https://sorbet.run/#%23%20typed%3A%20true%0A%0Arand%28100%29.to_s%2836%29%0A%0Arand%2810**100%29.to_s%2836%29) that in yours where you displayed that Sorbet does in fact know it is an Integer, it does not throw an error about too many args, but in ours where it believes it is a `Numeric` it does throw the error.

Indeed, thank you for pointing these out.  My mistake for not thinking things through thoroughly.

oneandonlymike:
Closing this as I have not had a chance to circle back to it

PR #4850, opened 2021-11-12
Add a disabled test for def inside instance_eval from self.included

aprocter-stripe:
Adds a disabled compiler test case for defining a method inside `instance_eval` within `self.included`.

### Motivation
Encountered this pattern in the wild, and we don't seem to compile it correctly.

### Test plan
See included automated tests.

### DIFF
--- a/test/testdata/compiler/disabled/instance_eval_def.rb
+++ b/test/testdata/compiler/disabled/instance_eval_def.rb
@@ -6,7 +6,7 @@ module M
   include Kernel
 
   def self.included(klass)
-    puts "hello from C::M.included"
+    puts "hello from M.included"
 
     klass.instance_eval do
       puts "hello from the instance_eval block"
--- a/test/testdata/compiler/disabled/instance_eval_def.rb
+++ b/test/testdata/compiler/disabled/instance_eval_def.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+# typed: true
+# compiled: true
+
+module M
+  include Kernel
+
+  def self.included(klass)
+    puts "hello from C::M.included"
+
+    klass.instance_eval do
+      puts "hello from the instance_eval block"
+      puts "self is: #{self}"
+      puts "let's def f"
+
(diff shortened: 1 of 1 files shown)


aprocter-stripe:
Not sure if I'm missing something but I think the rewrite tree would explain what we're seeing here. Looks like `f` has been lifted out of the `instance_eval` block to become `M#f`:

```
  module <emptyTree>::<C M><<C <todo sym>>> < ()
    def self.included<<todo method>>(klass, &<blk>)
      begin
        <self>.puts("hello from M.included")
        klass.instance_eval() do ||
          begin
            <self>.puts("hello from the instance_eval block")
            <self>.puts(::<Magic>.<string-interpolate>("self is: ", <self>))
            <self>.puts("let\'s def f")
            ::Sorbet::Private::Static.keep_def(<self>, :f, :normal)
            <self>.puts("done deffing f")
          end
        end
      end
    end

    def f<<todo method>>(&<blk>)
      <self>.puts("hello from f")
    end

    <self>.include(<emptyTree>::<C Kernel>)

    ::Sorbet::Private::Static.keep_self_def(<self>, :included, :normal)
  end

  module <emptyTree>::<C Z><<C <todo sym>>> < ()
    <self>.include(<emptyTree>::<C M>)
  end

  <emptyTree>::<C T>.unsafe(<emptyTree>::<C Z>).f()
end
```

elliottt (approved):

froydnj:
> Not sure if I'm missing something but I think the rewrite tree would explain what we're seeing here. Looks like `f` has been lifted out of the `instance_eval` block to become `M#f`:

This is OK: for the purposes of the compiler, `def` defines the C function behind the method.  The actual definition of the Ruby method is done via `keep_def`.

I think the problem here is that `keep_def` doesn't properly access `self`, instead pretending that Sorbet's idea of the type of `self` is correct...which is not always the same thing (especially inside `instance_eval`).  I have ideas about how to solve this.

PR #4851, opened 2021-11-15
don't type check blocks in the compiler

froydnj:
### Motivation

I had a FIXME in #4783 where I skipped typechecking blocks to make my life easier, but I knew I needed to put it back.

...then I read that we don't type check procs at runtime: https://sorbet.org/docs/procs

(see also the code driving type-checking for signatures:

https://github.com/sorbet/sorbet/blob/4725eb7442dd348470d11d164913ed51982b1831/gems/sorbet-runtime/lib/types/private/methods/signature.rb#L159-L204

which doesn't yield types or names for block parameters)

So we should stop doing this in the compiler because a) it's extra work compared to the interpreter runtime and b) converting blocks into procs just for type-checking purposes is unnecessary work.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/IREmitter.cc
+++ b/compiler/IREmitter/IREmitter.cc
@@ -745,15 +745,17 @@ void emitUserBody(CompilerState &base, cfg::CFG &cfg, const IREmitterContext &ir
 
                     // Argument values are loaded by `setupArguments`, we just need to check their type here
                     auto &argInfo = arguments[i.argId];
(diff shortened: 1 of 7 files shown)


froydnj:

→ https://go/builds/bui_KbMQYvba2sOiCa
→ https://go/builds/bui_KbMQsOe5FaF0no

aprocter-stripe (approved):
Matching sorbet-runtime behavior here makes sense to me!

PR #4856, opened 2021-11-16
Add missing separator in SelfTypeParam::toString

jez:
### Motivation

Noticed this while working on something else.

### Test plan

Look how exp files change.


test/testdata/cfg/break.rb.cfg-text.exp
```diff
@@ -277,8 +277,6 @@ bb22[rubyBlockId=0, firstDead=-1]():
     # outerLoops: 1
     <returnTemp>$99: Symbol(:abc) = :abc
     <block-break-assign>$100: Symbol(:abc) = <returnTemp>$99
-    <magic>$101: T.class_of(<Magic>) = alias <C <Magic>>
-    <block-break>$102: T.untyped = <magic>$101: T.class_of(<Magic>).<block-break>(<returnTemp>$99: Symbol(:abc))
```
jez, line 281: It looks like these change because the `cfg-text` exp files are not getting checked on master, and these changes were introduced in a PR that already landed: #4845

elliottt (approved):
🚢

PR #4857, opened 2021-11-16
Actually check cfg-text exp files

jez:
### Motivation

These exp files were not actually getting checked :horror:

We switched to using this test format relatively recently: #4764

That PR seemingly did not update the pipeline test runner in enough places.

### Test plan

You'll notice that the reported status on the first commit in this stack fails
in Buildkite, to prove that those tests are getting checked now.

The second commit regenerates the exp files for the failing tests (changes are
expected, as a result of #4845)


froydnj (approved):
Doh, good catch!

PR #4861, opened 2021-11-16
propagate type assumptions for never-modified method arguments into blocks

froydnj:
[This is an updated, rebased version of stripe/sorbet_llvm#490]

### Motivation

DeepFreeze has code like:

```ruby
    sig(:final) {params(todo: T::Array[T.untyped], obj: T.untyped).void.checked(:tests)}
    def self.freeze_one(todo, obj)
      case obj
      ...
      when Hash
        obj.freeze
        obj.each do |key, value|
          todo << key
          todo << value
        end
    ...
```

The block provided to each ought to be able to treat todo as an Array, just like todo would be at the "toplevel" of the method, outside of any blocks. Unfortunately, since blocks are compiled as separate functions, the information that todo has been type-checked to be an Array is lost and we need to continually re-check on each iteration.

This PR attempts to change that by recording very basic read/write information for locals that escape (and therefore get stuffed into the locals array) as well as type information for locals that correspond to method arguments. Once we have all that information, if an escaped variable is:

* a method argument
* never written to anywhere
* has a reasonable type that we can add llvm.assume for

we can "propagate" that information into the LLVM IR for the block. The included test shows the sort of thing that should happen.

Obviously there are a lot of caveats. I could see making the framework a little more flexible; it would be nice to catch something like:

```ruby
sig {params(other_hash: T::Hash[...]).returns(...)
def frob(other_hash)
  ...
  hash = {}
  other_hash.each do |k, v|
    modified_k = ...
    modified_v = ...
    hash[modified_k] = modified_v
  end
  ...
```

I think we can do this during `findCaptures` by having some analysis on `Send`s that ask whether we can "trust" the return type or not.  We can trust the return type from, e.g.:

* `<Magic>.<build-hash>`
* `<Magic>.<build-array>`
* `"constant-string".dup` (this one comes up in logging code)

and maybe a couple other cases.  Then we'd just say, on the first write to a variable (?) that the send type could be trusted and make it "read-only", trusting any later writes to make it actually `WrittenTo`.  But that is a followup PR in any event.  (The actual code for `frob` has `hash = other_hash.class.new` which is a different kind of wrinkle.)

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/Payload.cc
+++ b/compiler/IREmitter/Payload.cc
@@ -986,7 +986,29 @@ llvm::Value *Payload::varGet(CompilerState &cs, cfg::LocalRef local, llvm::IRBui
     if (irctx.escapedVariableIndices.contains(local)) {
         auto *cfp = getCFPForBlock(cs, builder, irctx, rubyBlockId);
         auto info = escapedVariableInfo(cs, local, irctx, rubyBlockId);
-        return builder.CreateCall(cs.getFunction("sorbet_readLocal"), {cfp, info.index, info.level});
+        auto *value = builder.CreateCall(cs.getFunction("sorbet_readLocal"), {cfp, info.index, info.level});
+        // If we ever wrote to this local, we cannot guarantee that the type has
+        // been checked prior to the access from the locals array.
+        if (info.use.used == LocalUsedHow::WrittenTo) {
+            return value;
+        }
+
+        core::ClassOrModuleRef klass;
+        typecase(info.use.type,
+                 [&klass](const core::ClassType &ct) { klass = ct.symbol; },
+                 [&klass](const core::AppliedType &at) { klass = at.klass; },
+                 [](const core::TypePtr &def) {});
+        if (!klass.exists()) {
+            return value;
+        }
+
+        if (!hasOptimizedTest(klass)) {
+            return value;
+        }
+
+        Payload::assumeType(cs, builder, value, klass);
+
+        return value;
     }
 
     // normal local variable
--- a/compiler/IREmitter/Payload.cc
+++ b/compiler/IREmitter/Payload.cc
@@ -550,7 +550,9 @@ void fillLocals(CompilerState &cs, llvm::IRBuilderBase &builder, const IREmitter
         escapedVariables.emplace_back(entry);
     }
 
-    fast_sort(escapedVariables, [](const auto &left, const auto &right) -> bool { return left.second.localIndex < right.second.localIndex; });
+    fast_sort(escapedVariables, [](const auto &left, const auto &right) -> bool {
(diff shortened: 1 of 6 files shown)


froydnj:

→ https://go/builds/bui_KbiLumFgT8aKg9
→ https://go/builds/bui_KbiLYDzkuNlWGw

compiler/IREmitter/Payload.cc
```diff
@@ -985,8 +987,29 @@ llvm::Value *Payload::varGet(CompilerState &cs, cfg::LocalRef local, llvm::IRBui
...
+            [&klass](const core::AppliedType &at) { klass = at.klass; }, [](const core::TypePtr &def) {});
+        if (!klass.exists()) {
+            return value;
+        }
+
+        if (!hasOptimizedTest(klass)) {
+            return value;
+        }
+
+        Payload::assumeType(cs, builder, value, klass);
```
elliottt, line 1010: It's unfortunate that you end up having to replicate the `core::TypePtr` overload of `hasOptimizedTest` here. What do you think about these two options:

* Modify `hasOptimizedTest` to return an `optional<core::ClassOrModuleRef>` to indicate that the optimized test exists
* Add an overload to `assumeType` that accepts a `core::TypePtr` argument and only emits the annotation if the type given is a `ClassType` or an `AppliedType`

My preference would be the first option, where we could also rely on `.exists()` to indicate success instead of using an `optional`.

compiler/IREmitter/Payload.cc
```diff
@@ -985,8 +987,29 @@ llvm::Value *Payload::varGet(CompilerState &cs, cfg::LocalRef local, llvm::IRBui
...
         auto *cfp = getCFPForBlock(cs, builder, irctx, rubyBlockId);
-        auto [index, level] = escapedVariableIndexAndLevel(cs, local, irctx, rubyBlockId);
-        return builder.CreateCall(cs.getFunction("sorbet_readLocal"), {cfp, index, level});
+        auto info = escapedVariableInfo(cs, local, irctx, rubyBlockId);
+        auto *value = builder.CreateCall(cs.getFunction("sorbet_readLocal"), {cfp, info.index, info.level});
+        // If we ever wrote to this local, we cannot guarantee that the type has
+        // been checked prior to the access from the locals array.
+        if (info.use.used == LocalUsedHow::WrittenTo) {
+            return value;
+        }
```
elliottt, line 996: Do you have a sense for whether or not we have enough static information to relax this in some cases? Since this PR tracks whether or not a variable is ever written to, the optimization would be disabled for cases where a variable is written to once at the end of a function, but is only read inside of a block.

elliottt (approved):
Looks great!

One question, and a suggestion about how we might change the type assertion functions to inline a bit less in `varGet`.

PR #4862, opened 2021-11-16
Add some intrinsics for identity functions

elliottt:
`to_ary`, `to_hash`, `to_i`, `to_int`, and `to_sym` all return their receiver when it's the target type. This PR adds intrinsics for them that return the receiver and adds the appropriate type annotation.

### Motivation

Performance.

### Test plan

See included automated tests.


froydnj (approved):
Cool!  Can we add one for `String#to_s` too, or does that have strange behavior?

elliottt:
> Cool! Can we add one for `String#to_s` too, or does that have strange behavior?

Also strange behavior for subtypes of `String`. I'll do that one in a follow-up as well :)

PR #4868, opened 2021-11-17
Support testing static vm functions in intrinsic tests

elliottt:
14cd06c53ddb87887173d7214e76ea515ba3a767 - Extend `IREmitterHelpers::receiverFastPathTestWithCache` to support calling another function to get the address of a static vm function when emitting a comparison. This result is stored in a global variable and cached at module load time to avoid jumping back and forth between the loaded module and the vm.

a10204e5f5ab4b7b77d965230a9e610fb6bc67af - As an example, this PR adds an implementation of `String#to_each`, which is defined by testing the results of method search against the static symbol `rb_str_to_s`.

c35613563c6bc1341e807298255253aef4df4542 - Add benchmarks for different uses of `String#to_s`.

```
# this branch
ruby vm startup time: .090
source                                          interpreted     compiled
stripe/while_10_000_000.rb                      .239            .116
stripe/string_subclass_to_s.rb                  .806            .556
stripe/string_subclass_to_s.rb - baseline       .567            .440
stripe/string_to_s.rb                           .454            .217
stripe/string_to_s.rb - baseline                .215            .101


# master
source                                          interpreted     compiled
stripe/while_10_000_000.rb                      .238            .115
stripe/string_subclass_to_s.rb                  .811            .675
stripe/string_subclass_to_s.rb - baseline       .573            .560
stripe/string_to_s.rb                           .447            .313
stripe/string_to_s.rb - baseline                .209            .198
```

### Motivation

More flexible symbol based intrinsics, and a  faster implementation of `String#to_s`.

### Test plan

See included automated tests.


compiler/IREmitter/Payload/patches/string.c
```diff
@@ -8,3 +8,11 @@ VALUE sorbet_vm_fstring_new(const char *ptr, long len) {
...
     return str_uplus(str);
 }
+
+VALUE (*sorbet_rb_str_to_s_func(void))(VALUE) {
+    return rb_str_to_s;
+}
+
+VALUE sorbet_rb_str_to_s(VALUE str) {
+    return rb_str_to_s(str);
+}
```
froydnj, line 18: Why do we have both of these?  Can't we just say in the intrinsic that we expect the cached function to be `rb_str_to_s`?

(And then if we have that, I suppose we don't need the `KnownFunction` bits?  At least not for this particular method...)

test/testdata/compiler/intrinsics/identities.rb
```diff
@@ -45,6 +45,12 @@ def self.integer_to_int(x)
   def self.symbol_to_sym(x)
     x.to_sym
   end
+
+  sig {params(x: String).returns(String)}
+  def self.string_to_s(x)
+    x.to_s
+  end
```
froydnj, line 52: Can we have some filecheck tests for this function?

test/testdata/compiler/intrinsics/identities.rb
```diff
@@ -56,3 +62,9 @@ def self.symbol_to_sym(x)
 p M.integer_to_int(20)
 
 p M.symbol_to_sym(:foo)
+
+class StringSubclass < String
+end
```
froydnj, line 67: Shouldn't this class be defining its own `to_s` method to ensure that our intrinsic implementation isn't just calling `rb_str_to_s` directly?

compiler/IREmitter/IREmitterHelpers.h
```diff
@@ -58,6 +58,31 @@ class Intrinsics {
...
+        CachedSymbol,
+    };
+
+    Type type;
+    std::string name;
+
+    // Implicit constructor from a string for backwards compability
+    KnownFunction(std::string name) : KnownFunction(Type::Symbol, std::move(name)) {}
+
+    KnownFunction(Type type, std::string name) : type{type}, name{std::move(name)} {}
```
froydnj, line 76: WDYT about making this function private?

froydnj (approved):
I think this makes sense, though I am not totally sure why we need the `KnownFunction` bits for this particular function (I think we need it for other methods, if they're implemented with Ruby-internal functions?).

compiler/IREmitter/Payload/patches/string.c
elliottt: `rb_str_to_s` is `static` in string.c, which is what led to all of this. The entry in the inline cache will be `rb_str_to_s`, not `sorbet_rb_str_to_s`, so there's some hoops to jump through.

test/testdata/compiler/intrinsics/identities.rb
elliottt: We should test both. The default implementation of `String#to_s` will do something different for subclasses of `String`, which is what this case is testing.

compiler/IREmitter/IREmitterHelpers.h
elliottt: Works for me!

compiler/IREmitter/Payload/patches/string.c
froydnj: Oh, of course.  Sorry, I thought these were in the codegen payload; this makes more sense now.

elliottt:

→ https://go/builds/bui_KcB2SAV5VdjxGU
→ https://go/builds/bui_KcB2yxBgyrgsJQ

elliottt:

→ https://go/builds/bui_KcBOCpZjy0WhiI
→ https://go/builds/bui_KcBO09kyjXI72i

PR #4872, opened 2021-11-17
Remove showRaw from membersStableOrderSlow

jez:
### Motivation

This revives #3536, fixing the bug that I was too lazy to debug last time.

It should substantially speed up operations that need to use
`membersStableOrderSlow`, which includes LSP requests like document symbol and
completion, as well as tasks that dump Sorbet's symbol table (in any plain text
or binary format).

On the step in Stripe's codebase that uses Sorbet to generate missing method RBIs, this drops the time spent computing `symbol-table-full-messagepack` from 127.9s to 73.2.s (savings of 54.7s)

### Test plan

Existing tests.

I confirmed that missing_methods.rb still works on Stripe's codebase.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -2027,6 +2027,24 @@ vector<std::pair<NameRef, SymbolRef>> Symbol::membersStableOrderSlow(const Globa
         if (compareRaw != 0) {
             return compareRaw < 0;
         }
+        int i = -1;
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_KcAzquhNTT0i5i
→ https://go/builds/bui_KcAzq6V2tijida

elliottt (approved):

PR #4873, opened 2021-11-17
Add intrinsic for Array#uniq

aprocter-stripe:
This adds a compiler intrinsic for `Array#uniq`.

### Motivation
We're pushing on adding compiler intrinsics in general, and this one's on the list. :)

### Test plan
See included automated tests.

### DIFF
--- a/compiler/IREmitter/Payload/codegen-payload.c
+++ b/compiler/IREmitter/Payload/codegen-payload.c
@@ -1307,8 +1307,7 @@ VALUE sorbet_rb_array_uniq_withBlock(VALUE recv, ID fun, int argc, const VALUE *
     if (RARRAY_LEN(ary) <= 1) {
         hash = 0;
         uniq = rb_ary_dup(ary);
-    }
-    else {
+    } else {
         // inline of ary_tmp_hash_new
         long size = RARRAY_LEN(ary);
         hash = rb_hash_new_with_size(size);
--- a/compiler/IREmitter/Payload/codegen-payload.c
+++ b/compiler/IREmitter/Payload/codegen-payload.c
@@ -113,6 +113,8 @@ SORBET_ALIVE(VALUE, sorbet_stringInterpolate,
 SORBET_ALIVE(VALUE, sorbet_rb_array_square_br_slowpath,
              (VALUE recv, ID fun, int argc, const VALUE *const restrict argv, BlockFFIType blk, VALUE closure));
 SORBET_ALIVE(VALUE, rb_ary_compact_bang_forwarder, (VALUE recv));
+SORBET_ALIVE(VALUE, sorbet_ary_make_hash, (VALUE ary));
+SORBET_ALIVE(void, sorbet_ary_recycle_hash, (VALUE hash));
 
 SORBET_ALIVE(void, sorbet_hashUpdate, (VALUE hash, VALUE other));
 
@@ -160,6 +162,8 @@ SORBET_ALIVE(void, sorbet_vm_define_method,
              (VALUE klass, const char *name, rb_sorbet_func_t methodPtr, void *paramp, rb_iseq_t *iseq, bool isSelf));
 SORBET_ALIVE(void, sorbet_vm_define_prop_getter,
              (VALUE klass, const char *name, rb_sorbet_func_t methodPtr, void *paramp, rb_iseq_t *iseq));
+
+SORBET_ALIVE(void, sorbet_rbasic_set_class, (VALUE obj, VALUE klass));
 
 SORBET_ALIVE(VALUE, sorbet_maybeAllocateObjectFastPath, (VALUE recv, struct FunctionInlineCache *newCache));
 SORBET_ALIVE(VALUE, sorbet_vm_instance_variable_get,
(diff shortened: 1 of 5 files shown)


compiler/IREmitter/Payload/patches/object.c
```diff
@@ -126,3 +126,7 @@ VALUE (*sorbet_vm_Kernel_instance_variable_get_func(void))(VALUE obj, VALUE iv)
 VALUE (*sorbet_vm_Kernel_instance_variable_set_func(void))(VALUE obj, VALUE iv, VALUE value) {
     return rb_obj_ivar_set;
 }
+
+void sorbet_rbasic_set_class(VALUE obj, VALUE klass) {
+    RBASIC_SET_CLASS(obj, klass);
+}
```
elliottt, line 132: Do we need this? It looks like it's defined in `internal.h`, which we do include in the codegen-payload. If we need it because it's defined in terms of something that we don't have access to, could you add a comment noting that?

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -1277,6 +1281,80 @@ VALUE sorbet_rb_array_empty(VALUE recv, ID fun, int argc, const VALUE *const res
...
+    }
+
+    return uniq;
+}
+
+// This is the block version of rb_ary_uniq: https://github.com/ruby/ruby/blob/ruby_2_7/array.c#L5018-L5041
+SORBET_INLINE
+VALUE sorbet_rb_array_uniq_withBlock(VALUE recv, ID fun, int argc, const VALUE *const restrict argv, BlockFFIType blk,
+                                     const struct rb_captured_block *captured, VALUE closure, int numPositionalArgs) {
+    rb_check_arity(argc, 0, 0);
```
elliottt, line 1313: This version is defined in the codegen payload, and will only call functions in the vm if it's going to raise an error.
```suggestion
    sorbet_ensure_arity(argc, 0);
```

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -1277,6 +1281,80 @@ VALUE sorbet_rb_array_empty(VALUE recv, ID fun, int argc, const VALUE *const res
...
+        hash = sorbet_ary_make_hash(ary);
+        uniq = rb_hash_values(hash);
+    }
+    sorbet_rbasic_set_class(uniq, rb_obj_class(ary));
+    if (hash) {
+        sorbet_ary_recycle_hash(hash);
+    }
+
+    return uniq;
+}
```
elliottt, line 1307: What do you think about moving this whole function into `array.c`? It doesn't look like it would benefit from being inlined, and as most of the work it does is in the vm already this would mean we'd cross the boundary only once per call.

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -1277,6 +1281,80 @@ VALUE sorbet_rb_array_empty(VALUE recv, ID fun, int argc, const VALUE *const res
...
+        // inline of ary_make_hash_by
+        for (int i = 0; i < RARRAY_LEN(ary); ++i) {
+            // inline of rb_ary_elt
+            VALUE v;
+            long len = RARRAY_LEN(ary);
+            if (len == 0) v = Qnil;
+            else if (i < 0 || len <= i) {
+                v = Qnil;
+            }
+            else v = RARRAY_AREF(ary, i);
```
elliottt, line 1341: Would you mind putting curly braces around these conditionals? I know the ruby vm has a lot of uncurlied blocks, but I prefer to avoid them when possible :)

I also wonder if the check for `RARRAY_LEN` on line 1336 is necessary. I know it's inherited from inlining `rb_ary_elt`, but wouldn't the loop body not execute if it was 0?

elliottt (approved):
I think this all looks good, just had a few questions :)

compiler/IREmitter/Payload/patches/object.c
aprocter-stripe: (:portal-from: Slack DMs...)

Originally I did this because `RBASIC_SET_CLASS` was expanding to something that includes `typeof`, which makes clang angry because that's a gcc-ism. We worked out that setting `-std=gnu99` (instead of `c99`) fixes this. Going to PR that separately.

compiler/IREmitter/Payload/codegen-payload.c
aprocter-stripe: Makes sense! We don't want to do the same for `sorbet_rb_array_uniq_withBlock`, since we want opportunities to inline the block, right?

compiler/IREmitter/Payload/codegen-payload.c
aprocter-stripe: Oh you're right about the `RARRAY_LEN` thing too. For a moment I was scratching my head because I wasn't sure if mutating the array inside the block could cause `i` to be out of bounds, but we're re-evaluating `RARRAY_LEN` in the conditional right before we jump in here, so that doesn't appear to be possible. Will simplify this.

compiler/IREmitter/Payload/codegen-payload.c
elliottt: Yep, exactly.

PR #4877, opened 2021-11-18
Switch copts for payload build from -std=c99 to -std=gnu99

aprocter-stripe:
PR title says it all.

### Motivation
https://github.com/sorbet/sorbet/pull/4873#discussion_r751799241 (tl;dr when compiling the payload we need some GNU-flavored extensions, like the `typeof` keyword, that are depended upon by Ruby)

### Test plan
See included automated tests.

### DIFF
--- a/compiler/IREmitter/Payload/BUILD
+++ b/compiler/IREmitter/Payload/BUILD
@@ -68,6 +68,9 @@ payload_files = [
     copts = [
         "-Wno-gcc-compat",
         "-fembed-bitcode",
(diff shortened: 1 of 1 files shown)


compiler/IREmitter/Payload/BUILD
```diff
@@ -68,7 +68,7 @@ payload_files = [
     copts = [
         "-Wno-gcc-compat",
         "-fembed-bitcode",
-        "-std=c99",
+        "-std=gnu99",
```
jez, line 71: Can you add a comment here?

jez (approved):

PR #4884, opened 2021-11-18
Migrate uses of --spawn_strategy=standalone to local

elliottt:
`standalone` is no longer a valid argument to `--spawn_strategy` or `--genrule_strategy`, but we missed this because bazel doesn't seem to emit an error when the flag is given an incorrect value.

### Motivation

Build performance -- we run builds with `--spawn_strategy=standalone` when using the `buildfarm-sanitized-*` configurations, wihch has been using the wrong argument to `--spawn_strategy` for a while now.

### Test plan

See included automated tests.


jez (approved):
lmao

PR #4886, opened 2021-11-19
Packages: nice errors for missing imports

ngroman:
When constant resolution fails and in packaged mode, search the possible resolution scopes for the prefix+name and find packages that start with that name. Suggest importing them. If no matches are found fall back to fuzzy name matching.

This is the first of other error handling changes to make improve UX of `--stripe-packages` mode. We also plan to add auto-correctors for these errors in future PRs.

Few other things this does:
* Expose `vector<NameRef> &fullName()` on `PackageInfo`
* Keep and expose a list of all mangled package names on `PackageDB` to make it easy to iterate over all packages

### Motivation
UX

### Test plan

See included automated tests.

cc @jvilk-stripe @aisamanra

### DIFF
--- a/core/packages/PackageDB.cc
+++ b/core/packages/PackageDB.cc
@@ -12,6 +12,11 @@ public:
         return NameRef::noName();
     }
 
+    const vector<core::NameRef> &fullName() const {
+        ENFORCE(false);
+        return emptyName;
+    }
+
     const vector<string> &pathPrefixes() const {
         ENFORCE(false);
         return prefixes;
@@ -31,6 +36,7 @@ public:
 
 private:
     const vector<string> prefixes;
+    const vector<core::NameRef> emptyName;
 };
 static const NonePackage NONE_PKG;
 } // namespace
@@ -53,11 +59,12 @@ NameRef PackageDB::enterPackage(unique_ptr<PackageInfo> pkg) {
     ENFORCE(!frozen);
     ENFORCE(writerThread == this_thread::get_id(), "PackageDB writes are not thread safe");
     auto nr = pkg->mangledName();
-    auto prev = packages.find(nr);
-    if (prev == packages.end()) {
+    auto prev = packages_.find(nr);
+    if (prev == packages_.end()) {
         for (const auto &prefix : pkg->pathPrefixes()) {
             packagesByPathPrefix[prefix] = nr;
         }
+        mangledNames.emplace_back(nr);
     } else {
         // Package files do not have full featured content hashing. If the contents of one changes
         // we always run slow-path and fully rebuild the set of packages. In some cases, the LSP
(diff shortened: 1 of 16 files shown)


core/packages/PackageDB.cc
```diff
@@ -102,15 +112,19 @@ const PackageInfo &PackageDB::getPackageForFile(const core::GlobalState &gs, cor
...
     }
     return *it->second;
 }
 
 bool PackageDB::empty() const {
-    return packages.empty();
+    return packages_.empty();
+}
+
+const vector<core::NameRef> &PackageDB::packages() const {
```
jvilk-stripe, line 126: Question about the name of this function: Are these package namespaces?

resolver/SuggestPackage.cc
```diff
@@ -0,0 +1,135 @@
...
+    //
+    // Consider the package `Foo::Bar::Baz` that exports the name `Foo::Bar::Baz::Thing`.
+    // The following reference to that name from code in the package `Foo::Quuz`:
+    //   Foo::Bar::Baz::Thing
+    //        ^^^  The resolution error will be for `Bar` not resolving with a resolution scope
+    //             `Foo`
+    // In this case all we search with the prefix `Foo::Bar` and find that the `Foo::Bar::Baz`
+    // package matches.
+    void findPackagesWithPrefix(const vector<core::NameRef> &prefix, vector<PackageMatch> &matches) const {
+        ENFORCE(prefix.size() > 0);
```
jvilk-stripe, line 86: Nit:

```suggestion
        ENFORCE(!prefix.empty());
```

resolver/SuggestPackage.cc
```diff
@@ -0,0 +1,135 @@
...
+        return ctx.state.packageDB();
+    }
+
+    vector<PackageMatch> findPossibleMissingImports(const ast::ConstantLit::ResolutionScopes &scopes,
+                                                    core::NameRef name) const {
+        vector<PackageMatch> matches;
+        vector<core::NameRef> prefix;
+        // Resolution scopes are always longest to shortest. Reserve additional space for the name we're
+        // looking for.
+        prefix.reserve(nameLen(ctx, scopes.front()) + 1);
```
jvilk-stripe, line 51: This seems like a premature optimization. I'd remove this line and `nameLen`, and just let `prefix` resize naturally with the first run of `symbol2NameParts`. That reduces the duplicate code between `nameLen` and `symbol2NameParts`.

core/packages/PackageDB.cc
```diff
@@ -121,16 +135,23 @@ const std::vector<std::string> &PackageDB::extraPackageFilesDirectoryPrefixes()
     return extraPackageFilesDirectoryPrefixes_;
 }
 
+bool PackageDB::isTestFile(const core::GlobalState &gs, const core::File &file) {
```
jvilk-stripe, line 138: Hopefully in the future this will change to just check if the file is within a test package or a normal package.

jvilk-stripe (approved):
These changes look fine to me. I had a few minor suggestions.

core/packages/PackageDB.cc
ngroman: These are the mangled names, which I'm effectively using as refs to pass around for packages.

PR #4887, opened 2021-11-19
Fix typo in index

jsegura:
### Motivation

Fix a small typo on index page.


jez:
This typo is intentional, to show how sorbet can catch typos.

PR #4888, opened 2021-11-19
Add intrinsic for Array#reject!

aprocter-stripe:
Adds a compiler intrinsic for `Array#reject!`

### Motivation

### Test plan
See included automated tests.


aprocter-stripe:

→ https://go/builds/bui_KcuPUjSwO1GxAv
→ https://go/builds/bui_KcuPB6DP6majvP

froydnj (approved):
This seems reasonable, but I kind of wonder how much win we get from this, given that the compiler can't inline the block because it's passed through the `rb_ensure`.  Do you have a benchmark, out of curiosity?

aprocter-stripe:
@froydnj I don't have a benchmark but I think you're probably right that it's not a very big win. At some point I'll construct a benchmark and see what falls out. If it doesn't turn out to do much for us, maybe we can look for a way to make it inline better.

PR #4890, opened 2021-11-19
Add intrinsic for Integer#to_f

aprocter-stripe:
Adds a compiler intrinsic for `Integer#to_f`.

So I went with the easy route and did this through `wrap-intrinsics.rb` which I think should probably be an improvement, but I think it might actually make more sense to implement this in `SymbolBasedIntrinsics` so it inlines better. I think it probably makes sense to land this as-is and follow up later, so throwin' it out there for review.

### Motivation

### Test plan
See included automated tests.


aprocter-stripe:

→ https://go/builds/bui_KcuQkW9TvkjXaY
→ https://go/builds/bui_KcuQbJI4UGj3es

elliottt (approved):

PR #4892, opened 2021-11-19
Add an intrinsic for `String#split`

elliottt:
Though `String#split` can take a block argument, it looks like we'll get a pretty big win just by having an intrinsic for the non-block variant.

### Motivation

More intrinsics for common methods

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/compiler/intrinsics/string_split.rb
+++ b/test/testdata/compiler/intrinsics/string_split.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+# compiled: true
+# typed: true
+# run_filecheck: OPT
+
+module M
+  extend T::Sig
+
+  # OPT-LABEL: define internal i64 @func_M.10test_split(i32
+  sig {params(str: String).void}
+  def self.test_split(str)
+    # OPT: sorbet_rb_str_split_m
+    p str.split(' ')
+
+    # OPT: sorbet_rb_str_split_m
+    p str.split(/\s/)
+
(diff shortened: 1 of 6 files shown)


elliottt:

→ https://go/builds/bui_KcwphkWz8huS1v
→ https://go/builds/bui_KcwqW3jpEIIQXc

test/testdata/compiler/intrinsics/string_split.rb
```diff
@@ -0,0 +1,27 @@
...
+    # OPT: sorbet_rb_str_split_m
+    p str.split(/\s/)
+
+    # The current `String#split` intrinsic doesn't support blocks, so this case
+    # should fall back on vm dispatch
+    #
+    # OPT-NOT: sorbet_rb_str_split_m
+    p(str.split(/l/) {|x| puts "sub: #{x}"})
+  end
+  # OPT{LITERAL}: }
```
aprocter-stripe, line 24: Probably doesn't matter _too_ much for purposes of this test, but note that `split` can also take 0 arguments (the default pattern seems to be effectively `/\s+/`) or 2 (the second argument is an int limiting the number of splits, e.g. `"a:b:c:d".split(":",3)` is `["a","b","c:d"])`. Might consider throwing tests in for those.

aprocter-stripe (approved):

elliottt:
Closing, as this intrinsic will accidentally inherit blocks from the calling context in its current implementation.

PR #4900, opened 2021-11-22
Make SymbolRef::data private

jvilk-stripe:
Make SymbolRef::data private. The next step is to completely remove this method. This step prevents most new occurrences from SymbolRef::data from appearing in the codebase.

I have intentionally structured this PR so that we can break it up into multiple PRs or have it be reviewed commit-by-commit. I am opening this PR to open the discussion concerning how to break this up and land it. After discussing with @trevor-stripe , we believe that this PR is best to land all at once and would be harder to review piecemeal.

For convenience, I have introduced several methods onto `SymbolRef` that dispatch to the proper symbol-type `*Ref` class. I intend to remove as many of these methods as feasible. Fixing some of the callsites will involve reworking some code paths to switch on symbol type earlier. I did not want to introduce those changes into this PR, as it is already a bit big.

### Motivation

Currently, all symbol types (class or module, method, field, ...) share a common class: `Symbol`. This class contains a superset of the fields needed by any given symbol, which means that every method allocates a `members_` map that it doesn't need[0].

By removing `SymbolRef::data`, we can implement separate data structures per symbol type, which will have the following downstream effects:

* **More precise types for cleaner code.** This is the biggest motivator for this change in my book. For example, we can specify that a `Method` can only have a `ClassOrModuleRef` `owner`. This removes a lot of "casting" of `SymbolRef` with `isMethod()`/`asMethodRef()` and should make it more obvious what symbol type to expect[1]. We can also define symbol-specific methods, like `setIsModule()`, on the individual symbol types that it applies to. Currently, we ENFORCE that these methods are not called on the wrong symbol type at runtime in debug builds only.
* **Smaller Symbol types.** We can remove `UnorderedMap` from every non-class or module symbol, which saves a little bit of memory. The savings will not be super large since ASTs dominate Sorbet's memory footprint. However, shrinking symbol size has a more promising second impact... (see next bullet)
* **Faster namer.** Methods make up the majority of entries in the symbol table at Stripe. Creating method symbols will become faster because they will be smaller and will have fewer fields to allocate. It looks like the namer pass could become up to 8% faster[2]. 

[0] OK, it's currently used for type arguments -- but it _really_ doesn't have to be. Methods don't have enough to justify an UnorderedMap over a vector (and it already has a vector of type arguments we could use!).

[1] `typeArguments()` / `typeMembers()` can also return vectors of `TypeArgumentRef`/`TypeMemberRef`. There are a lot of small API improvements here that add up to big improvements at callsites that currently have to "cast" the SymbolRef to the proper type.

[2] I estimated this value by introducing a _second_ `UnorderedMap` into the `Symbol` class and measuring Sorbet performance on the Stripe codebase. It introduced an 8% slowdown to namer, which estimates the impact of allocating/constructing this data type for every symbol in the symbol table. Since methods make up over half of the entries in the symbol table, we could conservatively estimate a 5% speedup. There are other fields in Symbol that will be removed from the `Method` class, but `UnorderedMap` is the biggest 'win'.

### Test plan

Covered by existing tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1419,15 +1419,16 @@ class ResolveTypeMembersAndFieldsWalk {
         }
 
         auto prior = scope.data(ctx)->findMember(ctx, uid->name);
-        if (prior.exists()) {
-            if (core::Types::equiv(ctx, prior.data(ctx)->resultType, cast->type)) {
+        if (prior.exists() && prior.isFieldOrStaticField()) {
+            auto priorField = prior.asFieldRef();
+            if (core::Types::equiv(ctx, priorField.data(ctx)->resultType, cast->type)) {
                 // We already have a symbol for this field, and it matches what we already saw, so we can short
                 // circuit.
                 return;
             } else {
                 // We do some normalization here to ensure that the file / line we report the error on doesn't
                 // depend on the order that we traverse files nor the order we traverse within a file.
-                auto priorLoc = prior.loc(ctx);
+                auto priorLoc = priorField.data(ctx)->loc();
                 core::Loc reportOn;
                 core::Loc errorLine;
                 core::Loc thisLoc = core::Loc(job.file, uid->loc);
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1402,7 +1402,7 @@ class ResolveTypeMembersAndFieldsWalk {
             // class or body, rather then nested in some block
             if (job.atTopLevel && ctx.owner.data(ctx)->isClassOrModule()) {
                 // Declaring a class instance variable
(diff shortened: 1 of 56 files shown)


resolver/resolver.cc
```diff
@@ -1724,11 +1726,9 @@ class ResolveTypeMembersAndFieldsWalk {
     }
 
     static void resolveMethodAlias(core::MutableContext ctx, const ResolveMethodAliasItem &job) {
-        core::SymbolRef member = ctx.owner.data(ctx)->findMemberNoDealias(ctx, job.toName);
-        // TODO(jvilk): Would be nice to have findMember that returned MethodRef.
```
elliottt, line 1728: 🎉

elliottt (approved):
It's so exciting to see `ENFORCE`s disappear because they're no longer necessary!

PR #4903, opened 2021-11-22
Add a pass to redirect the fast path over unnecessary type tests

elliottt:
Our intrinsics add both a fast and a slow path, and when their result is chained with another method send we get diamond-shaped control flow. This PR attempts to fix this by finding fast paths that assert the result of the type test being performed where the two diamonds join together, and redirecting it to the fast path on the exit. This makes a slow path that can re-join the fast path, and one long fast path that avoids re-checking types.

### Motivation

Better performance on the fast path by avoiding unnecessary type tests.

### Test plan

See included automated tests.


elliottt:
This had somewhat underwhelming results: the difference to master for a program with `x.to_a.length` in a tight loop was negligible. This might be worth reviving if we can make it work for user-defined types, but all the types that we have type-testing intrinsics for currently are pretty fast to test.

PR #4904, opened 2021-11-23
Fix `Tempfile#read` in tempfile.rbi

Slike9:
### Motivation

Due to https://github.com/ruby/ruby/blob/master/io.c#L3320, the result of `Temfile#read` can be `nil`.

```
irb(main):008:0> require 'tempfile'; Tempfile.new.read(1)
=> nil
```

### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/tempfile.rbi
+++ b/rbi/stdlib/tempfile.rbi
@@ -286,7 +286,7 @@ class Tempfile
         length: Integer,
         outbuf: String,
     )
-    .returns(String)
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_KeKgSP6huNVjKA
→ https://go/builds/bui_KeKgNs7TuiV3FS

Slike9:
This is going to be fixed in https://github.com/sorbet/sorbet/pull/4907.

PR #4908, opened 2021-11-23
.ruby-version: 2.6.3 -> 2.7.2

jez:
### Motivation

We already run the sorbet-runtime tests in 2.6.3 and 2.7.2. This commit
doesn't change that. This is more of a convenience, for e.g., when I run
`irb` from the `~/sorbet` folder, or I have ALE use `ruby` to check a
program for syntax errors, and the file is using 2.7-only syntax.

### Test plan

See included automated tests.


aprocter-stripe:
Looks like we'll need to tweak the `Dockerfile` in `sorbet-build-image` to make this work:

https://github.com/sorbet/sorbet-build-image/blob/cec2defc199fc8774cd056ad7df2f793daad9f0f/Dockerfile#L33-L36

aprocter-stripe (approved):
Oh nice, just running `rbenv install` makes even more sense :)

PR #4909, opened 2021-11-24
Add helper for package autocorrect imports

aisamanra:
### Motivation
This adds (but does not yet expose) an interface to `PackageInfo` that can be used to add imports via autocorrect. This right now adds the imports at the end—we can later try to position them "correctly", but this gives us a baseline to work from.

This also adds an extra method to `AutocorrectSuggestion` that's useful for testing, since the existing interface explicitly relies on re-reading the file in question, and the test suite included here uses inline string literals instead of on-disk files.

### Test plan

See included automated tests. Adds a separate unit test suite that tests only these autocorrects.

### DIFF
--- a/core/AutocorrectSuggestion.cc
+++ b/core/AutocorrectSuggestion.cc
@@ -26,9 +26,16 @@ bool hasSeen(const UnorderedSet<Loc> &seen, Loc loc) {
     return false;
 }
 
-string AutocorrectSuggestion::apply(const std::string_view source) {
+string AutocorrectSuggestion::unsafeApplyToString(const std::string_view source) {
     UnorderedSet<Loc> seen;
     vector<AutocorrectSuggestion::Edit> edits = move(this->edits);
+
+    if (edits.size() > 1) {
+        auto file = edits.front().loc.file();
+        for (int i = 1; i < edits.size(); i++) {
+            ENFORCE(edits[i].loc.file() == file, "All edits in `unsafeApplyToString` need to apply to the same file");
+        }
+    }
 
     auto compare = [](const AutocorrectSuggestion::Edit &left, const AutocorrectSuggestion::Edit &right) {
         return left.loc.beginPos() > right.loc.beginPos();
--- a/core/AutocorrectSuggestion.cc
+++ b/core/AutocorrectSuggestion.cc
@@ -26,6 +26,27 @@ bool hasSeen(const UnorderedSet<Loc> &seen, Loc loc) {
     return false;
 }
 
+string AutocorrectSuggestion::apply(const std::string_view source) {
+    UnorderedSet<Loc> seen;
+    vector<AutocorrectSuggestion::Edit> edits = move(this->edits);
+
+    auto compare = [](const AutocorrectSuggestion::Edit &left, const AutocorrectSuggestion::Edit &right) {
+        return left.loc.beginPos() > right.loc.beginPos();
+    };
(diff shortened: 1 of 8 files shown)


core/AutocorrectSuggestion.cc
```diff
@@ -26,6 +26,32 @@ bool hasSeen(const UnorderedSet<Loc> &seen, Loc loc) {
     return false;
 }
 
+string AutocorrectSuggestion::unsafeApplyToString(const std::string_view source) {
+    UnorderedSet<Loc> seen;
+    vector<AutocorrectSuggestion::Edit> edits = move(this->edits);
```
froydnj, line 31: Can we just duplicate `this->edits` here, rather than mutating `this`?  (Bonus points for making this method `const`, I guess.)  It's not immediately obvious to me that `unsafeX` would mutate the object.

core/AutocorrectSuggestion.cc
```diff
@@ -26,6 +26,32 @@ bool hasSeen(const UnorderedSet<Loc> &seen, Loc loc) {
...
+
+    if (edits.size() > 1) {
```
froydnj, line 33: Nit: this could just be `if (!edits.empty())`?

core/AutocorrectSuggestion.cc
```diff
@@ -26,6 +26,32 @@ bool hasSeen(const UnorderedSet<Loc> &seen, Loc loc) {
...
+    fast_sort(edits, compare);
+
+    string replaced{source};
+    for (auto &edit : edits) {
+        auto start = edit.loc.beginPos();
+        auto end = edit.loc.endPos();
+        replaced = absl::StrCat(replaced.substr(0, start), edit.replacement, replaced.substr(end, -1));
+    }
+
+    return replaced;
```
froydnj, line 52: The documentation for this function in the header says that it applies a single edit, but the function as written here applies multiple edits.  The loop as written here relies on subsequent edits having locs in the mutated string rather than the original string, which seems like a subtle thing to get right.

I think this function should apply only a single edit and throw in the `0` or `> 1` edit cases, especially as this is really only for testing.  (And, to that end, should this be called `applySingleEditForTesting` or somesuch?)

froydnj (changes requested):
This all looks reasonable, but I think `unsafeApplyToString` needs some changes before this lands.

core/AutocorrectSuggestion.cc
aisamanra: I think that makes sense. I can make that change.

aisamanra: (FWIW, the documentation says it applies a single `AutocorrectSuggestion`, which is true: a single `AutocorrectSuggestion`, however, might have multiple edits.) This is in contrast to `AutocorrectSuggestion::apply`, which applies a vector of `AutocorrectSuggestion`s.)

froydnj: > (FWIW, the documentation says it applies a single AutocorrectSuggestion, which is true: a single AutocorrectSuggestion, however, might have multiple edits. This is in contrast to AutocorrectSuggestion::apply, which applies a vector of AutocorrectSuggestions.)

Doh, reading comprehension fail on my part.

aisamanra:
Okay, I've renamed the method and made it significantly more single-purpose.

froydnj (approved):
:ship:

PR #4911, opened 2021-11-24
Preserve mixin ordering in resolution

ngroman:
The resolver attempts to resolve ancestors (superclass and mixins) early as possible as an optimization. If a mixin is referenced via an alias this is deferred until the alias itself has been resolved, and at that time it is pushed class's vector of mixins. In this case it results in re-ordering. (See test case on master)

Preserve the ordering by adding a placeholder value and updating the job to hold that index.

Note: that this is still best effort w.r.t. ordering. This change allows for Sorbet to more closely model what would occur at runtime if a class is defined in a single file. However, if a class's `include`s happen in multiple files it is impossible for Sorbet to know what order they'd be executed in.

### Motivation
`--stripe-packages` mode introduces encapsulation via constant aliases making this existing issue more prounced.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -544,7 +544,7 @@ bool Symbol::addMixin(const GlobalState &gs, ClassOrModuleRef sym, std::optional
     return isValidMixin;
 }
 
-u2 Symbol::addStubMixin(const GlobalState &gs) {
+u2 Symbol::addMixinPlaceholder(const GlobalState &gs) {
     ENFORCE(isClassOrModule());
     ENFORCE(ref(gs) != core::Symbols::StubMixin(), "Created a cycle through StubMixin");
     mixins_.emplace_back(core::Symbols::StubMixin());
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -502,7 +502,7 @@ TypePtr ArgInfo::argumentTypeAsSeenByImplementation(Context ctx, core::TypeConst
     return Types::arrayOf(ctx, instantiated);
 }
 
-bool Symbol::addMixin(const GlobalState &gs, ClassOrModuleRef sym, std::optional<size_t> index) {
+bool Symbol::addMixin(const GlobalState &gs, ClassOrModuleRef sym, std::optional<u2> index) {
     ENFORCE(isClassOrModule());
     // Note: Symbols without an explicit declaration may not have class or module set. They default to modules in
     // GlobalPass.cc. We also do not complain if the mixin is BasicObject.
@@ -514,7 +514,7 @@ bool Symbol::addMixin(const GlobalState &gs, ClassOrModuleRef sym, std::optional
         // semantically important!)
         // This is the 99% common case.
         if (index.has_value()) {
-            size_t i = index.value();
+            auto i = index.value();
             ENFORCE(mixins_.size() > i);
             ENFORCE(!mixins_[i].exists());
             mixins_[i] = sym;
@@ -544,9 +544,10 @@ bool Symbol::addMixin(const GlobalState &gs, ClassOrModuleRef sym, std::optional
     return isValidMixin;
 }
 
-size_t Symbol::addStubMixin(const GlobalState &gs) {
+u2 Symbol::addStubMixin(const GlobalState &gs) {
     ENFORCE(isClassOrModule());
     mixins_.emplace_back(ClassOrModuleRef());
+    ENFORCE(mixins_.size() < numeric_limits<u2>::max());
     return mixins_.size() - 1;
 }
 
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -547,7 +547,6 @@ bool Symbol::addMixin(const GlobalState &gs, ClassOrModuleRef sym, std::optional
 size_t Symbol::addStubMixin(const GlobalState &gs) {
     ENFORCE(isClassOrModule());
     mixins_.emplace_back(ClassOrModuleRef());
-    fmt::print("STUB {}\n", mixins_.size() - 1);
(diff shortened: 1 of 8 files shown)


core/Symbols.h
```diff
@@ -142,7 +142,10 @@ class Symbol final {
 
     // Attempts to add the given mixin to the symbol. If the mixin is invalid because it is not a module, it returns
     // `false` (but still adds the mixin for processing during linearization) and the caller should report an error.
-    [[nodiscard]] bool addMixin(const GlobalState &gs, ClassOrModuleRef sym);
+    [[nodiscard]] bool addMixin(const GlobalState &gs, ClassOrModuleRef sym, std::optional<u2> index = std::nullopt);
+
+    // Add a placeholder for a mixin and return index in mixins()
+    u2 addStubMixin(const GlobalState &gs);
```
jez, line 148: minor thing: can we potentially use a different name? sorbet currently uses `stub` exclusively to describe the state when it had to synthesize a constant because the source referred to something that didn't exist in the symbol table (i.e., "Unable to resolve constant" errors)

Maybe placeholder?

ngroman:
Test case fails on master with:
```
test/pipeline_test_runner.cc:160:
TEST CASE:  PerPhaseTest

test/testdata/resolver/aliased_mixin_ordering.rb:44: ERROR: Expected error of form:
44: T.reveal_type(MixesAlias.new.f) # error: Revealed type: `Integer`
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Revealed type: `Integer`
Found error:
44: T.reveal_type(MixesAlias.new.f) # error: Revealed type: `Integer`
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Revealed type: `String`

```

resolver/resolver.cc
```diff
@@ -126,6 +126,7 @@ class ResolveConstantsWalk {
 
         bool isSuperclass; // true if superclass, false for mixin
         bool isInclude;    // true if include, false if extend
+        std::optional<u2> mixinIndex;
```
ngroman, line 129: "65,535 ancestors should be enough for anyone"

core/Symbols.cc
```diff
@@ -536,6 +544,14 @@ bool Symbol::addMixin(const GlobalState &gs, ClassOrModuleRef sym) {
     return isValidMixin;
 }
 
+u2 Symbol::addMixinPlaceholder(const GlobalState &gs) {
+    ENFORCE(isClassOrModule());
+    ENFORCE(ref(gs) != core::Symbols::StubMixin(), "Created a cycle through StubMixin");
+    mixins_.emplace_back(core::Symbols::StubMixin());
```
ngroman, line 550: I'm re-using this existing sentinel value but if it's important we could create a new one. I tried using an empty value but that causes problems in checks that expect all `mixins_` to exist like `derivesFrom`.

jez: Yeah i'd kind of like another symbol here. Right now, if you print the symbol table at some point and see `StubMixin` in the output, you know where sorbet went wrong. Re-using `StubMixin` here makes it trickier to debug in the future.

Also, if we changed this to something like `MixinPlaceholder` here, we could add an ENFORCE later down the line that says that there are no `MixinPlaceholder`'s in the mixins list, which forces Sorbet to have to fill them back in, or replace them with `StubMixin` if need be (This could potentially be in the same place that you mentioned was enforce'ing from the first attempt when using `Symbols::noSymbol()`).

test/testdata/resolver/aliased_mixin_ordering.rb
```diff
@@ -0,0 +1,46 @@
...
+end
+
+class MixesAlias
+  extend T::Sig
+
+  StringyAlias = Stringy
+
+  include StringyAlias
+  include Inty
+end
```
jez, line 28: is it worth having a `MixesAliasSwapped` case?

jez (approved):
resolver change looks good to me, but i would love if we could pick a different name for the placeholder symbol

jez:
@ngroman do you have a sense for how this affects typecheck duration on Stripe's codebase? hoping it's negligible but you never know.

ngroman:
> @ngroman do you have a sense for how this affects typecheck duration on Stripe's codebase? hoping it's negligible but you never know.

@jez I'll pull some numbers. To clarify are you more concerned about the performance penalty of this change alone or this change _plus_ `--stripe-packages` (since all the aliases cause more iterations in the resolver loop)?

jez:
> @jez I'll pull some numbers. To clarify are you more concerned about the performance penalty of this change alone or this change _plus_ `--stripe-packages` (since all the aliases cause more iterations in the resolver loop)?

For now just this change, not stripe packages, but it would be nice to know both.

nroman-stripe:

→ https://go/builds/bui_KeknuH9wLiXjh9
→ https://go/builds/bui_KeknunicaKh9j8

core/Symbols.cc
ngroman: I'll look into it. I'm pretty sure we could do that enforce when globalpass linearizes the classes.

ngroman:
I profiled and did not find a meaningful difference in time.

test/testdata/resolver/aliased_mixin_ordering.rb
ngroman: I think this covers it well enough

ngroman:
Re-review for symbol and enforcement changes

jez (approved):
great thanks for the ENFORCE and rename!

PR #4913, opened 2021-11-24
Add sig for blockless Hash#select

aprocter-stripe:
Adds a sig for the blockless variant of `Hash#select` (the one that returns an `Enumerator`).

### Motivation
Was working on an intrinsic for `Hash#select` and noticed we don't have a sig for this.

### Test plan
See included automated tests.

### DIFF
--- a/rbi/core/hash.rbi
+++ b/rbi/core/hash.rbi
@@ -1096,6 +1096,7 @@ class Hash < Object
     )
     .returns(T::Hash[K, V])
   end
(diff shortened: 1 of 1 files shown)


aprocter-stripe:

→ https://go/builds/bui_KelqciZUcUOlSS
→ https://go/builds/bui_KelqthqSSmKZFq

jez (approved):

PR #4914, opened 2021-11-24
Add Hash#select intrinsic

aprocter-stripe:
Adds a compiler intrinsic for `Hash#select`.

**Benchmarks**

`Hash#select` with block:

| source | interpreted | compiled |
|-|-|-|
| stripe/hash_select_baseline.rb | .140 | .118 |
| stripe/hash_select.rb | .495 | .405 |
| stripe/hash_select.rb - baseline | .355 | .287 |

`Hash#select` without block (returns an enumerator):

| source | interpreted | compiled |
|-|-|-|
| stripe/hash_select_baseline.rb | .139 | .120 |
| stripe/hash_select_enum.rb | .799 | .625 |
| stripe/hash_select_enum.rb - baseline | .660 | .505 |

### Motivation

### Test plan
See included automated tests.

### DIFF
--- a/compiler/IREmitter/Payload/codegen-payload.c
+++ b/compiler/IREmitter/Payload/codegen-payload.c
@@ -1975,6 +1975,50 @@ VALUE sorbet_rb_hash_merge_withBlock(VALUE recv, ID fun, int argc, const VALUE *
                                            numPositionalArgs);
 }
 
+// This is the no-block version of rb_hash_select: https://github.com/ruby/ruby/blob/ruby_2_7/hash.c#L2694-L2705
+// In that version, the `RETURN_SIZED_ENUMERATOR` macro is what causes the early return when a block is not passed. In
+// this case, we know that the block wasn't passed, so we always return an enumerator
+SORBET_INLINE
+VALUE sorbet_rb_hash_select(VALUE recv, ID fun, int argc, const VALUE *const restrict argv, BlockFFIType blk,
+                            VALUE closure) {
+    sorbet_ensure_arity(argc, 0);
+    return rb_enumeratorize_with_size(recv, ID2SYM(fun), argc, argv, sorbet_hash_enum_size);
+}
+
+struct sorbet_select_i_args {
+    VALUE result;
+    BlockFFIType blk;
+    VALUE closure;
+};
+
+static int sorbet_select_i(VALUE key, VALUE value, VALUE argsv) {
+    struct sorbet_select_i_args *args = (struct sorbet_select_i_args *)argsv;
+    VALUE block_argv[2] = {key, value};
+    if (RTEST(args->blk(key, args->closure, 2, &block_argv[0], Qnil))) {
+        rb_hash_aset(args->result, key, value);
+    }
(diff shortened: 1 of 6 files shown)


aprocter-stripe:

→ https://go/builds/bui_KemYNkPlKVeQLM
→ https://go/builds/bui_KemYTWf0sal5UM

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -1975,6 +1975,50 @@ VALUE sorbet_rb_hash_merge_withBlock(VALUE recv, ID fun, int argc, const VALUE *
...
+// This is the block version of rb_hash_select: https://github.com/ruby/ruby/blob/ruby_2_7/hash.c#L2694-L2705
+SORBET_INLINE
+VALUE sorbet_rb_hash_select_withBlock(VALUE recv, ID fun, int argc, const VALUE *const restrict argv, BlockFFIType blk,
+                                      const struct rb_captured_block *captured, VALUE closure, int numPositionalArgs) {
+    sorbet_ensure_arity(argc, 0);
+
+    // must push a frame for the captured block
+    sorbet_pushBlockFrame(captured);
+
+    struct sorbet_select_i_args args = {rb_hash_new(), blk, closure};
```
elliottt, line 2012: Is it worth pre-allocating the hash? That might end up wasting some memory, but it could also avoid reallocations during the loop.

elliottt (approved):
This looks good to me. Do you happen to have results from running your benchmarks on master?

aprocter-stripe:
> This looks good to me. Do you happen to have results from running your benchmarks on master?

Good point!

`master`:

| source | interpreted | compiled |
|-|-|-|
| stripe/hash_select_baseline.rb | .142  |  .123 |
| stripe/hash_select.rb |  .516  |  .486 |
| stripe/hash_select.rb - baseline    |    .374  |  .363 |
| stripe/hash_select_enum.rb   |   .699 |   .668 |
| stripe/hash_select_enum.rb - baseline  | .557   | .545 |

PR branch (rebased on same `master` rev):

| source | interpreted |    compiled |
|-|-|-|
| stripe/hash_select_baseline.rb | .140  |  .123 |
| stripe/hash_select.rb |  .514   | .413 |
| stripe/hash_select.rb - baseline  |      .374 |   .290 |
| stripe/hash_select_enum.rb    |  .698   | .644 |
| stripe/hash_select_enum.rb - baseline |  .558  |  .521 |

So actually I guess about a third of the observed speedup on `hash_select_enum.rb`

compiler/IREmitter/Payload/codegen-payload.c
aprocter-stripe: I guess my thinking is that since the existing intrinsic doesn't do any preallocation, and the memory wasted could be pretty large under some ~realistic circumstances if the filter is "sparse" enough, we probably don't want to introduce preallocation. (I could imagine pathological cases like using k `select` calls to partition a very large hash into k categories. Probably not the right tool for the job, but I'm sure folks are doing it.)

PR #4915, opened 2021-11-26
Remove "IDE ready" section

quinn:
Since Sorbet does not have IDE support, this section is a bit misleading.

### Motivation
Sorbet does not have IDE support, this is closed source and (I assume) only available internally at Stripe. 

### Test plan
I'm guessing the static homepage isn't under test

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -124,11 +124,6 @@ class Index extends React.Component {
                     'Sorbet is multithreaded, scaling linearly across cores on your CPU. It checks your types in seconds, giving you feedback as you code.\n\n[Get started with Sorbet →](docs/adopting)',
                 },
                 {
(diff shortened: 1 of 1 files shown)


jez:
Sorbet has numerous LSP features that are all open source and built into Sorbet behind the `--lsp` flag, available to any editor with LSP support out of the box. This is what we mean by IDE ready.

Thanks!

quinn:
Oh, interesting. There is a gif on the homepage showing usage in vs code, but googling has so far indicated that there is no vs code support. Is there documentation for setting up sorbet with vs code? This seems like an important piece of info to support the claim of “IDE ready”

I’m not aware of any technique  to integrate vs code with a language server directly, without the use of an extension. Is this what engineers at Stripe are doing? Somehow manually connecting their editors to the language server?

PR #4918, opened 2021-11-29
[LSP] Deduplicate document symbols

ilyailya:
### Motivation

Some files had duplicates in the document outline. Usually they looked like this:
```
Module A:
    def foo()
    def bar()

def foo()
def bar()
```

The bug happens, because in the task we will iterate over modules and classes recursively first, and then we will iterate by all the other kinds of symbols
https://github.com/sorbet/sorbet/blob/f349fb5ca882f687ec9c84fcd153a545b591f9ec/main/lsp/requests/document_symbol.cc#L105-L111

For example above, firstly we will recursively get all available symbols for `Module A` (which includes `foo` and `bar`), and after that we will get `def foo` and `def bar` one more time from iterating over `core::SymbolRef::Kind::Method`.

I've fixed that by adding a set of symbols we will include into the response.

### Test plan

See included automated tests + I've built this change on my devbox and run it over pay-server

### DIFF
--- a/main/lsp/requests/document_symbol.cc
+++ b/main/lsp/requests/document_symbol.cc
@@ -6,30 +6,41 @@
 using namespace std;
 
 namespace sorbet::realmain::lsp {
+
 std::unique_ptr<DocumentSymbol> symbolRef2DocumentSymbol(const core::GlobalState &gs, core::SymbolRef symRef,
-                                                         core::FileRef filter);
+                                                         core::FileRef filter,
+                                                         UnorderedSet<core::SymbolRef> &addedRefs);
 
 void symbolRef2DocumentSymbolWalkMembers(const core::GlobalState &gs, core::SymbolRef sym, core::FileRef filter,
-                                         vector<unique_ptr<DocumentSymbol>> &out) {
+                                         vector<unique_ptr<DocumentSymbol>> &out,
+                                         UnorderedSet<core::SymbolRef> &addedRefs) {
     for (auto mem : sym.data(gs)->membersStableOrderSlow(gs)) {
+        auto ref = mem.second;
+
+        if (addedRefs.contains(ref)) {
+            continue;
+        }
+
         if (mem.first != core::Names::attached() && mem.first != core::Names::singleton()) {
             bool foundThisFile = false;
-            for (auto loc : mem.second.data(gs)->locs()) {
+            for (auto loc : ref.data(gs)->locs()) {
                 foundThisFile = foundThisFile || loc.file() == filter;
             }
             if (!foundThisFile) {
                 continue;
             }
-            auto inner = symbolRef2DocumentSymbol(gs, mem.second, filter);
+            auto inner = symbolRef2DocumentSymbol(gs, ref, filter, addedRefs, config);
             if (inner) {
                 out.push_back(move(inner));
+                addedRefs.insert(ref);
             }
         }
     }
 }
 
 std::unique_ptr<DocumentSymbol> symbolRef2DocumentSymbol(const core::GlobalState &gs, core::SymbolRef symRef,
-                                                         core::FileRef filter) {
+                                                         core::FileRef filter,
+                                                         UnorderedSet<core::SymbolRef> &addedRefs) {
     if (!symRef.exists()) {
         return nullptr;
     }
@@ -50,6 +61,7 @@ std::unique_ptr<DocumentSymbol> symbolRef2DocumentSymbol(const core::GlobalState
         prefix = "self.";
     }
     auto result = make_unique<DocumentSymbol>(prefix + sym->name.show(gs), kind, move(range), move(selectionRange));
+    addedRefs.insert(symRef);
 
     // Previous versions of VSCode have a bug that requires this non-optional field to be present.
     // This previously tried to include the method signature but due to issues where large signatures were not readable
@@ -58,11 +70,11 @@ std::unique_ptr<DocumentSymbol> symbolRef2DocumentSymbol(const core::GlobalState
     result->detail = "";
 
     vector<unique_ptr<DocumentSymbol>> children;
-    symbolRef2DocumentSymbolWalkMembers(gs, symRef, filter, children);
+    symbolRef2DocumentSymbolWalkMembers(gs, symRef, filter, children, addedRefs);
     if (sym->isClassOrModule()) {
         auto singleton = sym->lookupSingletonClass(gs);
         if (singleton.exists()) {
-            symbolRef2DocumentSymbolWalkMembers(gs, singleton, filter, children);
+            symbolRef2DocumentSymbolWalkMembers(gs, singleton, filter, children, addedRefs);
         }
     }
     result->children = move(children);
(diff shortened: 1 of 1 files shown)


jez:
Quick question: instead of allocating a new structure to detect this, is it possible to change the algorithm to make it de-duplicate items by construction?

ilyailya:
@jez I've been thinking about that too. We probably can change the way how we iterate over symbols. 
But we can check if symbol has an owner, and skip it if so. For example, if method is inside some module, which is not root, we can skip it. I'll try to sketch this tonight

froydnj (changes requested):
It sounds like there's still some work to be done here, please let me know if that's not the case.

ilyailya:
@jez thanks for pushing me that way! I've come up with a more elegant fix

@froydnj  @jvilk PR is ready for review

Thanks!

jez:
> See included automated tests

I don't see an included test. Is it possible to write a test that would have shown this behavior/ensure we don't regress?

jez (approved):
Change looks much better, thanks for reworkig it!

ilyailya:
>I don't see an included test. Is it possible to write a test that would have shown this behavior/ensure we don't regress?

@jez I meant the tests which are already in the repo. I'm not sure if it would be possible to create such a test, 'cause I wasn't able to recreate the bug in a sandbox repo. It occurred only in pay-server. I'm not sure why exactly. Maybe because we have a large number of files in p-s? In this case module like `Opus::Something` could have a bunch of `locs` in different files, which could affect the bug.

jez:
> Maybe because we have a large number of files in p-s? In this case module like `Opus::Something` could have a bunch of `locs` in different files, which could affect the bug.

Did you ever try creating a multi-file LSP test?

If you name your test files like `test/testdata/lsp/my_test__1.rb` and `test/testdata/lsp/my_test__2.rb` (must be two underscores) then both of those files will be processed as a single test. So if you defined the same class in both those files:


```ruby
class A
  def foo; end
end
```

```ruby
class A
  def bar; end
end
```

something like this might showcase the problem? you can play around with it.

froydnj (approved):
Stamping this for completeness, but jez's review is sufficient.  Adding the multi-file LSP test would be interesting, though.

main/lsp/requests/document_symbol.cc
```diff
@@ -77,6 +79,27 @@ bool DocumentSymbolTask::isDelayable() const {
     return true;
 }
 
+bool isOwnerInTheSameFile(const core::GlobalState &gs, core::SymbolRef ref, core::FileRef fref) {
+    auto locs = ref.data(gs)->owner.data(gs)->locs();
```
jvilk-stripe, line 83: I think this might be copying the vector? If you take it by reference it will definitely not copy the vector:

```suggestion
    auto &locs = ref.data(gs)->owner.data(gs)->locs();
```

main/lsp/requests/document_symbol.cc
```diff
@@ -77,6 +79,27 @@ bool DocumentSymbolTask::isDelayable() const {
...
+
+    for (auto loc : locs) {
```
jvilk-stripe, line 85: if you wanna get fancy you can probably do `absl::c_any_of(locs, [fref](auto &loc) { return loc.file() == fref; });` in a few places in this file

ilyailya:
@jvilk-stripe thanks, that looks much better!

I was able to add a multi-file test, but it was comparably as hard as fixing the bug :feelsgood: It was hard to find out a test case, which would cause a duplication in a sandbox. Turns out something similar to `dev/scripts` in p-s works good.

Also our current setup for `document-symbols` expectations does work with multi-file tests. 
The first issue was `test/print_document_symbols.cc` is not expecting more than one file to as an input. I've made it possible to pass multiple files to the program, but with a constraint the symbols would be generated only for the first file. I hope we wouldn't need a multi-file multi-expectation any time soon (: 

The second issue was a filename for the `.exp` file. It didn't include the index of a test file. So we had no way to determine, which test file should be asserted against the `.exp` file. I've fixed that by giving an index to the expectation file.

main/lsp/requests/document_symbol.cc
```diff
@@ -79,6 +81,16 @@ bool DocumentSymbolTask::isDelayable() const {
     return true;
 }
 
+bool isOwnerInTheSameFile(const core::GlobalState &gs, core::SymbolRef ref, core::FileRef fref) {
+    auto &locs = ref.owner(gs).locs(gs);
+
+    return absl::c_any_of(locs, [fref](auto &loc) { return loc.file() == fref; });
```
jvilk-stripe, line 87: Nit: Isn't this equivalent?

```suggestion
    return isRefInFile(gs, ref.owner(gs), fref);
```

(You'll need to move `isRefInFile` above `isOwnerInTheSameFile`.)

test/print_document_symbols.cc
```diff
@@ -104,23 +108,28 @@ int printDocumentSymbols(string_view filePath) {
...
+            auto params = make_unique<DidChangeTextDocumentParams>(
+                make_unique<VersionedTextDocumentIdentifier>(fileUri, static_cast<double>(fileId++)), move(edits));
+            auto notif = make_unique<NotificationMessage>("2.0", LSPMethod::TextDocumentDidChange, move(params));
+            lspWrapper->getLSPResponsesFor(make_unique<LSPMessage>(move(notif)));
+        }
     }
 
-    auto docSymbolParams = make_unique<DocumentSymbolParams>(make_unique<TextDocumentIdentifier>(fileUri));
+    auto docSymbolParams =
+        make_unique<DocumentSymbolParams>(make_unique<TextDocumentIdentifier>(absl::StrCat(uriPrefix, paths[1])));
```
jvilk-stripe, line 132: 👍  OK, `paths[1]` is guarded by the `argc < 2` check in `main`.

test/print_document_symbols.cc
```diff
@@ -145,10 +154,11 @@ int printDocumentSymbols(string_view filePath) {
 } // namespace sorbet::realmain::lsp
 
 int main(int argc, char *argv[]) {
-    if (argc != 2) {
+    if (argc < 2) {
         std::cout << "Usage: print_document_symbols path/to/file.rb\n";
+        std::cout << "If you pass multiple files, symbols would be generate for the first file only";
```
jvilk-stripe, line 159:
```suggestion
        std::cout << "If you pass multiple files, symbols will be generated for the first file only";
```

test/print_document_symbols.cc
```diff
@@ -48,12 +48,16 @@ pair<string, vector<string>> findEditsToApply(string_view filePath) {
     return make_pair(uri, move(fileContents));
 }
 
-int printDocumentSymbols(string_view filePath) {
+int printDocumentSymbols(char *paths[], int fileAmount) {
```
jvilk-stripe, line 51: I think I would slightly prefer a `std::vector<string_view>` -- shield the inner routine from the nastiness of argc/argv.

test/print_document_symbols.cc
```diff
@@ -145,10 +154,11 @@ int printDocumentSymbols(string_view filePath) {
...
 int main(int argc, char *argv[]) {
-    if (argc != 2) {
+    if (argc < 2) {
         std::cout << "Usage: print_document_symbols path/to/file.rb\n";
+        std::cout << "If you pass multiple files, symbols would be generate for the first file only";
         return 1;
     }
 
-    return sorbet::realmain::lsp::printDocumentSymbols(argv[1]);
+    return sorbet::realmain::lsp::printDocumentSymbols(argv, argc);
```
jvilk-stripe, line 163:
```suggestion
    vector<string_view> args;
    for (auto i = 0; i < argc; ++i) {
      args.emplace_back(argv[i]);
    }
    return sorbet::realmain::lsp::printDocumentSymbols(args);
```

jvilk-stripe (approved):
Thanks for fixing this limitation and bug! I had a few minor nits, but otherwise LGTM.

test/print_document_symbols.cc
ilyailya: That was my initial thought, but I couldn't find a decent way to create a `vector` from the `*char[]` without rolling a loop, so I decided to go with what I have

PR #4920, opened 2021-11-29
make basic `super` calls use the Ruby stack

froydnj:
### Motivation

This makes `super` calls more similar to our normal Ruby calls, eliminates unnecessary allocations (e.g. keyword arguments can be passed efficiently to `super` now), and paves the way for moving `call-with-block` and `call-with-splat-and-block` bits to use `sorbet_i_send` as well.

### Test plan

See included automated tests.


jez (approved):
nice!

Canpolat17:
super

Canpolat17:
master

Canpolat17:
super

PR #4921, opened 2021-11-29
don't call `super` in `initialize` for `T::Struct` subclasses

froydnj:
### Motivation

The title says it all.  The compiler's view of `initialize` is that we're going to:

1. typecheck the arguments
2. set all the instance variables
3. call `super` (currently requiring allocating a kwargs hash, which wouldn't be required after #4920)
4. run all the sorbet-runtime code associated with initializing `T::Struct` subclasses...but theoretically our `initialize` method *already did that*

So dropping step 3 (and consequently 4) should be safe, since we're typechecking the arguments (which sorbet-runtime would do) and setting the instance variables (which sorbet-runtime would also do).

It's worth noting that this shouldn't affect the actual static typechecking Sorbet does because a) `super` is currently considered to be essentially untyped and b) even if we more correctly computed what `super` was calling, Sorbet doesn't see the actual implementation anyway.

A benchmark is included.  Before this change:

```
stripe/while_10_000_000.rb      .224    .101
stripe/t_struct_creation.rb     12.043  12.271
stripe/t_struct_creation.rb - baseline  11.819  12.170
```

These timings are somewhat noisy, given the allocations involved, but we can see that the compiler is no faster than the interpreter -- which makes sense, since it's basically doing all the work the interpreter would be doing, plus whatever work the rewriter's `initialize` method does.

After this change:

```
stripe/while_10_000_000.rb      .222    .100
stripe/t_struct_creation.rb     11.731  .887
stripe/t_struct_creation.rb - baseline  11.509  .787
```

I will take order-of-magnitude speedups.

### Test plan

I attempted to add tests to verify the compiler's behavior is ~compatible with sorbet-runtime's; I didn't go so far as to check the actual text of the error messages, because I don't think that's necessary in this case.  I don't claim that I captured all the behavior: extra review for those tests welcome.

### DIFF
--- a/rewriter/Prop.cc
+++ b/rewriter/Prop.cc
@@ -535,7 +535,19 @@ vector<ast::ExpressionPtr> mkTypedInitialize(core::MutableContext ctx, core::Loc
         stats.emplace_back(ast::MK::Assign(prop.loc, ast::MK::Instance(prop.nameLoc, ivarName),
                                            ast::MK::Local(prop.nameLoc, prop.name)));
     }
-    auto body = ast::MK::InsSeq(klassLoc, std::move(stats), ast::MK::Nil(klassDeclLoc));
+    // Normally we wouldn't need to call super here: the compiler will use the types
+    // in the sig to typecheck everything, just like sorbet-runtime, and we've
+    // generated a body to set all the appropriate instance variables, just like
+    // sorbet-runtime.  (deprecated) enum props, however, are not typechecked
+    // properly by the compiler, so we need to use super to call into sorbet-runtime
+    // to get the correct handling.
+    ast::ExpressionPtr maybeSuper;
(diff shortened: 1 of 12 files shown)


froydnj:

→ https://go/builds/bui_KgaNxArBJc2M5V
→ https://go/builds/bui_KgaNJ6hb0MbqMf

elliottt (approved):
This all looks good to me! Pretty nice speedup for a one line change in the struct rewrite :D

froydnj:

→ https://go/builds/bui_KgboCGTqAIQrGw
→ https://go/builds/bui_KgbozFqpARwsZT

froydnj:
ptal @elliottt 

Turns out we (currently) need to call `super` when we have enum props.

elliottt (approved):

PR #4923, opened 2021-11-29
Add better error messaging for packages importing prefixes of themselves

aadi-stripe:
### Motivation

In `--stripe-packages` mode, there are 3 classes of import conflict errors:
1. package `A::B` imports a prefix package `A`, which happens to export its whole namespace.
2. package `A` imports a suffix package `A::B`, but also defines constants inside the `A::B` namespace.
3. package `C` imports 2 packages `A` and `A::B`, where-in A exports its whole namespace.

Currently, we report a packager error for class 3, but not for 1 & 2, instead, relying on namer errors to inform the user. These namer errors can be difficult to understand. This change catches case 1 and reports a packager error. Case 2 might be a bit more complex to catch here (and is probably rarer), but it's something we can address later.

### Test plan

Added a CLI test.

See included automated tests.

### DIFF
--- a/test/cli/package-import-parent-package-conflict/parent/__package.rb
+++ b/test/cli/package-import-parent-package-conflict/parent/__package.rb
@@ -1,0 +1,5 @@
+# typed: strict
+
+class Parent < PackageSpec
+  export Parent
+end
--- a/test/cli/package-import-parent-package-conflict/parent/child/__package.rb
+++ b/test/cli/package-import-parent-package-conflict/parent/child/__package.rb
@@ -1,0 +1,5 @@
+# typed: strict
+
+class Parent::Child < PackageSpec
+  import Parent
+end
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -797,13 +797,15 @@ public:
 
 class ImportTreeBuilder final {
     // PackageInfoImpl package; // The package we are building an import tree for.
+    const FullyQualifiedName *fullPkgName;
     core::NameRef pkgMangledName;
     core::NameRef privatePkgMangledName;
     ImportTree root;
 
 public:
     ImportTreeBuilder(const PackageInfoImpl &package)
-        : pkgMangledName(package.name.mangledName), privatePkgMangledName(package.privateMangledName) {}
+        : fullPkgName(&(package.name.fullName)), pkgMangledName(package.name.mangledName),
+          privatePkgMangledName(package.privateMangledName) {}
     ImportTreeBuilder(const ImportTreeBuilder &) = delete;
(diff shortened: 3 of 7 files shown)


aadi-stripe:
tagging @nroman-stripe , @jvilk-stripe

aadi-stripe:

→ https://go/builds/bui_Kgg7MShNIwgwOO
→ https://go/builds/bui_Kgg7WCexGDmiVc

test/cli/package-import-parent-package-conflict/parent/__package.rb
```diff
@@ -0,0 +1,5 @@
+# typed: strict
+
+class Parent < PackageSpec
+  export Parent
```
jvilk-stripe, line 4: Ah, this is legal? So this would appear in an importing package as `Parent` and not `Parent::Parent`?

test/cli/package-import-parent-package-conflict/parent/child/__package.rb
```diff
@@ -0,0 +1,5 @@
+# typed: strict
+
+class Parent::Child < PackageSpec
+  import Parent
```
jvilk-stripe, line 4: The issue, as I understand it, is that this `import` statement defines a constant `Parent` that is equal to the module `Parent`. But then you can't define `Parent::Child`... right?

test/cli/package-import-parent-package-conflict/parent/__package.rb
jvilk-stripe: If `Parent` exported `Parent::Inheritance`, would that not be a problem? Because then it would be defining `Parent::Inheritence` and not the root level `Parent`?

jvilk-stripe (commented):
I wasn't aware of this shape of issue. The code LGTM, but I'll let roman take a look too!

test/cli/package-import-parent-package-conflict/parent/child/__package.rb
aadi-stripe: Yes, you would get a redefinition error in the namer. Often, that error is not easy to parse. Therefore, we catch it earlier and disallow the alias completely.

test/cli/package-import-parent-package-conflict/parent/__package.rb
aadi-stripe: * `Parent` exporting `Parent` is legal, and yes it would appear in an importing package as `Parent`.
* If `Parent` exported `Parent::Inheritance`, that's okay, because the importing package would define `Parent::Inheritance` and `Parent::Child` separately, and nothing would get redefined.

ngroman (approved):
lgtm

PR #4926, opened 2021-11-30
print send flags in raw tree output

froydnj:
### Motivation

I was making some changes on a different branch that dealt with send flags, and I noticed that there were no exp file changes as a result of my change, which seemed surprising.  Then I discovered that we don't print the flags in the raw output, and it seems like we should.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -913,7 +913,7 @@ string Send::showRaw(const core::GlobalState &gs, int tabs) {
         stringifiedFlags.emplace_back("privateOk");
     }
     if (this->flags.isRewriterSynthesized) {
-        stringifiedFlags.emplace_back("rewriter");
+        stringifiedFlags.emplace_back("rewriterSynthesized");
     }
 
     printTabs(buf, tabs + 1);
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -908,7 +908,7 @@ string Send::showRaw(const core::GlobalState &gs, int tabs) {
     fmt::memory_buffer buf;
     fmt::format_to(std::back_inserter(buf), "{}{{\n", nodeName());
 
-    auto stringifiedFlags = vector<string>{};
+    vector<string> stringifiedFlags;
     if (this->flags.isPrivateOk) {
         stringifiedFlags.emplace_back("privateOk");
     }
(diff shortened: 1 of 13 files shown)


ast/Trees.cc
```diff
@@ -907,6 +907,17 @@ string Send::toStringWithTabs(const core::GlobalState &gs, int tabs) const {
 string Send::showRaw(const core::GlobalState &gs, int tabs) {
     fmt::memory_buffer buf;
     fmt::format_to(std::back_inserter(buf), "{}{{\n", nodeName());
+
+    auto stringifiedFlags = vector<string>{};
+    if (this->flags.isPrivateOk) {
+        stringifiedFlags.emplace_back("privateOk");
+    }
+    if (this->flags.isRewriterSynthesized) {
+        stringifiedFlags.emplace_back("rewriter");
```
jez, line 916: if someone sees "rewriter" in the raw output and they're new to the codebase, it might take them a bit to know that it means they should look up the `isRewriterSynthesized` flag. maybe we should print out `rewriterSynthesized`? it's not like we're already trying to be concise in the raw output.

jez (approved):
what happened with the second commit about bad package related changes?

ast/Trees.cc
froydnj: That's a reasonable suggestion, I'll make that change.

froydnj (commented):
> what happened with the second commit about bad package related changes?

`update_testdata_exp.sh` regenerates that file such that the exp tests don't pass.  I think it's related to packaging flags, but I haven't dug into it.

ast/Trees.cc
```diff
@@ -907,6 +907,17 @@ string Send::toStringWithTabs(const core::GlobalState &gs, int tabs) const {
 string Send::showRaw(const core::GlobalState &gs, int tabs) {
     fmt::memory_buffer buf;
     fmt::format_to(std::back_inserter(buf), "{}{{\n", nodeName());
+
+    auto stringifiedFlags = vector<string>{};
```
elliottt, line 911:
```suggestion
    vector<string> stringifiedFlags;
```

elliottt (approved):

PR #4928, opened 2021-11-30
give `!` per-callsite caches

froydnj:
### Motivation

`!` calls on non-boolean, non-nil values currently shares a single method call cache per module.  This seems suboptimal, especially compared to the VM.  Benchmark:

```
ruby vm startup time: .093
stripe/while_10_000_000.rb      .247    .102
stripe/bang_intrinsic_dispatch.rb       .302    .232
stripe/bang_intrinsic_dispatch.rb - baseline    .055    .130
```

This PR changes that, and gives every `!` callsite a method call cache.  Same benchmark with changes:

```
stripe/while_10_000_000.rb      .240    .101
stripe/bang_intrinsic_dispatch.rb       .290    .155
stripe/bang_intrinsic_dispatch.rb - baseline    .050    .054
```

So we are ~as fast as the VM now, instead of 2x slower.  (Presumably we're a little bit slower because of function call overhead and the special-casing of boolean values and `nil`.)

This change means that the boolean case gets somewhat slower, but I think that's OK, because it's still significantly faster than the VM:

Before this change:

```
stripe/while_10_000_000.rb      .252    .109
stripe/bang_intrinsic_toggle.rb .401    .106
stripe/bang_intrinsic_toggle.rb - baseline      .149    -.003
```

I'm not entirely sure I believe those numbers, because I think the compiler might be constant-folding away some of the computation.  After:

```
stripe/while_10_000_000.rb      .253    .097
stripe/bang_intrinsic_toggle.rb .395    .118
stripe/bang_intrinsic_toggle.rb - baseline      .142    .021
```

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/NameBasedIntrinsics.cc
+++ b/compiler/IREmitter/NameBasedIntrinsics.cc
@@ -783,6 +783,28 @@ public:
     }
 } ClassIntrinsic;
 
+class BangIntrinsic : public NameBasedIntrinsicMethod {
+public:
+    BangIntrinsic() : NameBasedIntrinsicMethod{Intrinsics::HandleBlock::Unhandled} {};
+
+    virtual llvm::Value *makeCall(MethodCallContext &mcctx) const override {
+        if (mcctx.send->args.size() != 0) {
+            return IREmitterHelpers::emitMethodCallViaRubyVM(mcctx);
+        }
+
+        auto &cs = mcctx.cs;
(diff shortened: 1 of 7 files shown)


froydnj:

→ https://go/builds/bui_Kh03jkqfqu1Btr
→ https://go/builds/bui_Kh03oUSpIHg6xq

jez (approved):

PR #4932, opened 2021-11-30
Add a block intrinsic for Hash#fetch

elliottt:
```
# master
source                                                                  interpreted     compiled
stripe/while_10_000_000.rb                                               .276           .122
./test/testdata/ruby_benchmark/stripe/hash_fetch_block.rb               1.488          1.009
./test/testdata/ruby_benchmark/stripe/hash_fetch_block.rb - baseline    1.212           .887

# this branch
source                                                                  interpreted     compiled
stripe/while_10_000_000.rb                                               .274           .114
./test/testdata/ruby_benchmark/stripe/hash_fetch_block.rb               1.503           .497
./test/testdata/ruby_benchmark/stripe/hash_fetch_block.rb - baseline    1.229           .383
```

### Motivation

Performance, avoiding a call through the VM when a block is given to `Hash#fetch`.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/Payload/codegen-payload.c
+++ b/compiler/IREmitter/Payload/codegen-payload.c
@@ -704,6 +704,25 @@ VALUE sorbet_int_hash_to_h(VALUE recv, ID fun, int argc, const VALUE *const rest
     return sorbet_rb_hash_to_h(recv);
 }
 
+SORBET_INLINE
+VALUE sorbet_rb_hash_fetch_m_withBlock(VALUE recv, ID fun, int argc, const VALUE *const restrict argv, BlockFFIType blk,
+                                       const struct rb_captured_block *captured, VALUE closure, int numPositionalArgs) {
+    extern VALUE sorbet_hash_stlike_lookup(VALUE hash, VALUE key, VALUE * val);
+
+    rb_check_arity(argc, 1, 2);
+    VALUE key = argv[0];
+    VALUE val;
+
+    if (sorbet_hash_stlike_lookup(recv, key, &val)) {
+        return (VALUE)val;
+    } else {
+        sorbet_pushBlockFrame(captured);
+        val = blk(key, closure, 1, &key, Qnil);
+        sorbet_popFrame();
+        return val;
+    }
+}
+
 struct sorbet_int_hash_to_h_closure {
     BlockFFIType fun;
     VALUE closure_for_block;
--- a/compiler/IREmitter/Payload/codegen-payload.c
+++ b/compiler/IREmitter/Payload/codegen-payload.c
@@ -704,6 +704,8 @@ VALUE sorbet_int_hash_to_h(VALUE recv, ID fun, int argc, const VALUE *const rest
     return sorbet_rb_hash_to_h(recv);
 }
 
+// This is the block variant of rb_hash_fetch_m:
+// https://github.com/sorbet/ruby/blob/12d8c43330278a744d6e45135d1a8735f23f5afe/hash.c#L2113-L2147
 SORBET_INLINE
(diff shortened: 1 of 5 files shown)


elliottt:

→ https://go/builds/bui_Kh33p0m0UnJjHS
→ https://go/builds/bui_Kh33mPbxhXlYqR

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -704,6 +704,25 @@ VALUE sorbet_int_hash_to_h(VALUE recv, ID fun, int argc, const VALUE *const rest
     return sorbet_rb_hash_to_h(recv);
 }
 
+SORBET_INLINE
+VALUE sorbet_rb_hash_fetch_m_withBlock(VALUE recv, ID fun, int argc, const VALUE *const restrict argv, BlockFFIType blk,
+                                       const struct rb_captured_block *captured, VALUE closure, int numPositionalArgs) {
+    extern VALUE sorbet_hash_stlike_lookup(VALUE hash, VALUE key, VALUE * val);
+
+    rb_check_arity(argc, 1, 2);
```
aprocter-stripe, line 712: iirc* when argc==2 the interpreter will issue a warning that the second arg is silently dropped (because the block overrides it)---I think skipping the warning is probably fine, but you might add a comment about why we're checking for (1,2) then never looking at argv[1].

*iidnrc then please disregard :)

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -704,6 +704,25 @@ VALUE sorbet_int_hash_to_h(VALUE recv, ID fun, int argc, const VALUE *const rest
     return sorbet_rb_hash_to_h(recv);
 }
 
+SORBET_INLINE
+VALUE sorbet_rb_hash_fetch_m_withBlock(VALUE recv, ID fun, int argc, const VALUE *const restrict argv, BlockFFIType blk,
+                                       const struct rb_captured_block *captured, VALUE closure, int numPositionalArgs) {
```
aprocter-stripe, line 709: Might add a link to the upstream code this is borrowed from (seems like we've generally been doing so).

compiler/IREmitter/Payload/codegen-payload.c
```diff
@@ -704,6 +704,25 @@ VALUE sorbet_int_hash_to_h(VALUE recv, ID fun, int argc, const VALUE *const rest
...
+    extern VALUE sorbet_hash_stlike_lookup(VALUE hash, VALUE key, VALUE * val);
+
+    rb_check_arity(argc, 1, 2);
+    VALUE key = argv[0];
+    VALUE val;
+
+    if (sorbet_hash_stlike_lookup(recv, key, &val)) {
```
aprocter-stripe, line 716: `hash_stlike_lookup` is marked `inline` in the interpreter, right? It might be good to dash off a quick benchmark to make sure the lack of inlinability isn't hurting us here.

aprocter-stripe (approved):

compiler/IREmitter/Payload/codegen-payload.c
elliottt: I figured that it would be okay to suppress the warning here, but you're right that a comment explaining that would be nice.

compiler/IREmitter/Payload/codegen-payload.c
elliottt: My intuition was that we should be fine either way, as we'd have to call into the vm in either case to do the lookup. I'll add a benchmark just to confirm that.

aprocter-stripe: Makes sense! I didn't actually have an alternative in mind anyway. :)

PR #4901, opened 2021-11-22
Implement missing_methods.rb diff in Sorbet

jez:
### Motivation

This drastically reduces the time it takes to run missing_methods.rb on Stripe's codebase.

In the future, we can expand this to be a tool that will work for minimizing more kinds of RBIs, but I'm leaving that for future work.

### Test plan

See included automated tests.

PR #4933, opened 2021-11-30
Implement missing_methods.rb diff in Sorbet, take 2

jez:
### Motivation

Originally reviewed in #4901

Reverted in #4931

Review by commit

### Test plan

https://buildkite.com/sorbet/sorbet/builds/18887#2d88eac0-4264-4603-a7e9-6ccafc8901dd

passed on macOS (which was why it had to be reverted)


elliottt (approved):
Test fix makes sense to me!

PR #4935, opened 2021-12-01
Restructure the CallCMethod sanity checks to check for block handlers

elliottt:
To help avoid adding intrinsics for methods that can accept blocks, without an explicit block handling variant.

### Motivation

Avoiding potential correctness issues.

### Test plan

I manually removed a block handler from one of the `CallCMethod` intrinsics to verify that the `ENFORCE` triggers.


jez (approved):
Nice!

PR #4936, opened 2021-12-01
Combine rewritePackages and rewritePackagedFiles

aadi-stripe:
### Motivation

`rewritePackagedFiles` runs in parallel, `rewritePackages` does not. There's no functional reason why they can't both run in parallel, as part of the same pass. Saves about 4s in the packager pass.

### Test plan

Ran existing tests, also tested on Stripe codebase.

Before:
Screen Shot 2021-11-30 at 5 48 51 PM

After:
Screen Shot 2021-11-30 at 5 51 21 PM

See included automated tests.

cc @jvilk-stripe @nroman-stripe

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -1198,24 +1198,23 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file) {
     return file;
 }
 
-ast::ParsedFile rewritePackagedFile(core::GlobalState &gs, ast::ParsedFile parsedFile) {
-    auto &file = parsedFile.file.data(gs);
+ast::ParsedFile rewritePackagedFile(core::Context ctx, ast::ParsedFile parsedFile) {
+    auto &file = parsedFile.file.data(ctx);
     ENFORCE(file.sourceType != core::File::Type::Package);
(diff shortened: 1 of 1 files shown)


aadi-stripe:

→ https://go/builds/bui_Kh6QLjgLMfiG2J
→ https://go/builds/bui_Kh6QjO6MqT1FtV

packager/packager.cc
```diff
@@ -1298,6 +1289,9 @@ vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers
                     auto &file = job.file.data(gs);
                     if (file.sourceType == core::File::Type::Normal) {
                         job = rewritePackagedFile(gs, move(job));
```
jez, line 1291: For symmetry and if you want, you could hoist the part where `rewritePackagedFile` creates a `core::Context` out to the call site here.

That would drive home the point that both these two functions (`rewritePackagedFile` and `rewritePackage`) operate on immutable global states.

aadi-stripe: Fixed

jez (approved):

PR #4937, opened 2021-12-01
Report a better error when type aliases are exported in --stripe-packages mode

aadi-stripe:
### Motivation

In `--stripe-packages` mode, when a constant exported by a package is a type alias, we get error [5034](https://sorbet.org/docs/error-reference#5028) due to the fact that the packager's inserted virtual AST node effectively becomes a reassignment of a type-alias. We have ratcheted against this behavior in the Stripe codebase for now, but when we launch `--stripe-packages`, we may remove this ratchet. In that situation, we want to throw a more user-friendly error.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -436,7 +436,14 @@ private:
         auto rhsData = rhsSym.data(ctx);
         if (rhsData->isTypeAlias()) {
             if (auto e = ctx.beginError(it.rhs->loc, core::errors::Resolver::ReassignsTypeAlias)) {
-                e.setHeader("Reassigning a type alias is not allowed");
+                if (it.file.data(ctx).isPackage()) {
+                    // In --stripe-packages mode, this error surfaces when type aliases
+                    // are exported by a package. TODO (aadi-stripe, 12/30/2021) update docs for
+                    // error 5034 as part of any larger --stripe-packages documentation effort.
+                    e.setHeader("Exporting a type alias is not allowed in package specifications");
+                } else {
+                    e.setHeader("Reassigning a type alias is not allowed");
+                }
                 e.addErrorLine(rhsData->loc(), "Originally defined here");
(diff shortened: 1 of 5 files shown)


aadi-stripe:
cc @nroman-stripe

aadi-stripe:
Also cc @jvilk-stripe

core/Files.h
```diff
@@ -123,6 +123,10 @@ class File final {
     StrictLevel strictLevel;
 
     const CompiledLevel compiledLevel;
+
+    const bool isPackageFile() const {
+        return sourceType == Type::Package;
+    };
```
ngroman, line 129: `isPackage()` is already a method on `File`.  This is not needed.

resolver/resolver.cc
```diff
@@ -436,7 +436,14 @@ class ResolveConstantsWalk {
         auto rhsData = rhsSym.data(ctx);
         if (rhsData->isTypeAlias()) {
             if (auto e = ctx.beginError(it.rhs->loc, core::errors::Resolver::ReassignsTypeAlias)) {
-                e.setHeader("Reassigning a type alias is not allowed");
+                if (it.file.data(ctx).isPackageFile()) {
+                    // In --stripe-packages mode, this error surfaces when type aliases
+                    // are exported by a package. TODO (aadi-stripe, 12/30/2021) update docs for
+                    // error 5034 as part of any larger --stripe-packages documentation effort.
+                    e.setHeader("Exporting a type alias is not allowed");
```
ngroman, line 443:
```suggestion
                    e.setHeader("Exporting a type alias is not allowed in package specifications");
```

ngroman (changes requested):
Logic looks reasonable. But need to fix the duplicate method.

aadi-stripe:
ptal @ngroman

ngroman (approved):
Looks good from my end

PR #4938, opened 2021-12-01
make Tuple#sum smarter when faced with a homogeneous tuple

froydnj:
### Motivation

This case came up internally: `[1.3, 2.4].sum` should be able to figure out that the return type is a `Float`, not a `T.any(Float, Integer)`, because we know that the tuple is non-empty.

### Test plan

See included automated tests.

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -384,6 +384,7 @@ NameDef names[] = {
     {"first"},
     {"min"},
     {"max"},
+    {"sum"},
 
     // Argument forwarding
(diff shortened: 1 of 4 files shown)


froydnj:

→ https://go/builds/bui_KhKNcfqTQrENcr
→ https://go/builds/bui_KhKNEWTFcUnrxs

jez:
Do you think this is really worth the intrinsic, when the code could easily be rewritten from `[x, y].sum` to `x + y`?

jez (approved):

PR #4940, opened 2021-12-01
Mark sorbet_i_objIsKindOf as const

elliottt:
Remove `sorbet_isa` in favor of a direct call to `sorbet_i_objIsKindOf`, and mark `sorbet_i_objIsKindOf` as `const`. Also change the return type of `sorbet_i_objIsKindOf` to `_Bool`, matching the other type testing intrinsics.

### Motivation

Cleaning up generated code, coalescing calls to `sorbet_i_objIsKindOf` when possible.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/Payload.cc
+++ b/compiler/IREmitter/Payload.cc
@@ -350,7 +350,8 @@ llvm::Value *Payload::typeTest(CompilerState &cs, llvm::IRBuilderBase &builder, 
                                   {val, Payload::getRubyConstant(cs, attachedClass, builder)});
     }
     sym = isProc(sym) ? core::Symbols::Proc() : sym;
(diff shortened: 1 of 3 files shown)


jez (approved):
Is it intentional that there are no tests? Would a filecheck test be possible?

elliottt:
> Is it intentional that there are no tests? Would a filecheck test be possible?

It was intentional, but I think you're right that it would be nice to have a test to verify the coalescing.

elliottt:
OK, I had trouble writing a test to verify the behavior I expected because the function that I removed, `sorbet_isa`, is marked const and was coalescing before inlining to a `sorbet_i_objIsKindOf` call. The result is that this PR should basically be a no-op, but will leave the door open to `sorbet_isa` calls coalescing later than `sorbet_isa` would have.

elliottt:

→ https://go/builds/bui_KhSU1LCyIOvaih
→ https://go/builds/bui_KhSUyy3B0YHCpV

PR #4943, opened 2021-12-01
Add autocorrect for exports

aisamanra:
### Motivation
Add an autocorrect for `export` as well. This takes a `vector<NameRef>` for the exported name.

This also fixes some issues with the `import` autocorrect, and fixes the test suite so that it would have caught them.

### Test plan
See included automated tests. Added tests for both new functionality and fixed old functionality.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -209,7 +209,7 @@ public:
         // first let's try adding it to the end of the imports.
         if (!importedPackageNames.empty()) {
             auto lastOffset = importedPackageNames.back().name.loc;
-            insertionLoc = {info.loc.file(), lastOffset.endPos(), lastOffset.endPos()};
+            insertionLoc = {loc.file(), lastOffset.endPos(), lastOffset.endPos()};
         } else {
             // if we don't have any imports, then we can try adding it
             // either before the first export, or if we have no
@@ -218,11 +218,11 @@ public:
             if (!exports.empty()) {
                 exportLoc = exports.front().fqn.loc.beginPos() - "export "sv.size() - 1;
             } else {
-                exportLoc = info.loc.endPos() - "end"sv.size() - 1;
+                exportLoc = loc.endPos() - "end"sv.size() - 1;
             }
             // we want to find the end of the last non-empty line, so
             // let's do something gross: walk backward until we find non-whitespace
-            const auto &file_source = info.loc.file().data(gs).source();
+            const auto &file_source = loc.file().data(gs).source();
             while (isspace(file_source[exportLoc])) {
                 exportLoc--;
                 // this shouldn't happen in a well-formatted
@@ -231,7 +231,7 @@ public:
                     return nullopt;
                 }
             }
-            insertionLoc = {info.loc.file(), exportLoc + 1, exportLoc + 1};
+            insertionLoc = {loc.file(), exportLoc + 1, exportLoc + 1};
         }
         ENFORCE(insertionLoc.exists());
 
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -244,7 +244,8 @@ public:
         return {suggestion};
     }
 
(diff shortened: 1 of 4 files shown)


packager/packager.cc
```diff
@@ -218,11 +218,11 @@ class PackageInfoImpl final : public sorbet::core::packages::PackageInfo {
             if (!exports.empty()) {
                 exportLoc = exports.front().fqn.loc.beginPos() - "export "sv.size() - 1;
             } else {
-                exportLoc = info.loc.endPos() - "end"sv.size() - 1;
+                exportLoc = loc.endPos() - "end"sv.size() - 1;
```
aisamanra, line 221: The issues with the previous one were that we want to calculate the `loc` from _this_ package, not from `info` (which is the package we're importing.)

test/pkg_autocorrects_test.cc
```diff
@@ -20,51 +20,72 @@ namespace spd = spdlog;
 auto logger = spd::stderr_color_mt("pkg-autocorrects-test");
 auto errorQueue = make_shared<sorbet::core::ErrorQueue>(*logger, *logger);
 
+string examplePackage = "class Opus::ExamplePackage < PackageSpec\nend\n";
+string examplePackagePath = "example/__package.rb";
```
aisamanra, line 24: To catch the old issues, we now enter two packages into the global state, rather than the lazy shortcut being used before.

test/pkg_autocorrects_test.cc
```diff
@@ -188,17 +209,63 @@ TEST_CASE("Add test import to package with neither imports nor exports") {
...
+    auto test = TestPackageFile::create(gs, "my_package/__package.rb", pkg_source);
+
+    auto &package = test.targetPackage(gs);
+    ENFORCE(package.exists());
+    auto addExport = package.addExport(gs, test.getName(gs, {"Opus", "MyPackage", "NewExport"}), true);
+    ENFORCE(addExport, "Expected to get an autocorrect from `addImport`");
+    auto replaced = addExport->applySingleEditForTesting(pkg_source);
+    CHECK_EQ(expected, replaced);
+}
+
```
jez, line 270: Why are these tests in C++ instead of a CLI test?

packager/packager.cc
```diff
@@ -243,6 +243,50 @@ class PackageInfoImpl final : public sorbet::core::packages::PackageInfo {
               fmt::format("\n  {} {}", isTestImport ? "test_import" : "import", info.name.toString(gs))}});
         return {suggestion};
     }
+
+    std::optional<core::AutocorrectSuggestion>
```
jez, line 247:
```suggestion
   optional<core::AutocorrectSuggestion>
```

packager/packager.cc
```diff
@@ -243,6 +243,50 @@ class PackageInfoImpl final : public sorbet::core::packages::PackageInfo {
...
+    addExport(const core::GlobalState &gs, const vector<core::NameRef> newExport, bool isPrivateTestExport) const {
+        for (auto expt : exports) {
+            // check if we already import this, and if so, don't
+            // return an autocorrect
+            if (expt.fqn.parts == newExport) {
+                return nullopt;
+            }
+        }
+
+        core::Loc insertionLoc = loc.adjust(gs, core::INVALID_POS_LOC, core::INVALID_POS_LOC);
```
jez, line 257: this adds `INVALID_POS_LOC` to the beginPos and endPos.

Do you maybe want to do something like.

but you also write to this variable in every branch. do you just want to do

```suggestion
        auto insertionLoc = core::Loc::none(loc.file());
```

?

jez (approved):

test/pkg_autocorrects_test.cc
aisamanra: I wrote these without actually wiring up the autocorrects to the code that will eventually call them, and I also wanted tests that exercised only the autocorrects without also relying on constant resolution to find the right import. A CLI test would have required the whole end-to-end be present, and also would rely on a larger number of moving parts.

PR #4944, opened 2021-12-01
switch codegen ordering of final methods and name-based intrinsics

froydnj:
### Motivation

When #4760 landed, it silently disabled final method codegen for `[]` methods: the name-based intrinsic for `[]` would fire, and, finding that there was actual type information associated with the receiver, the intrinsic would emit a method call through the Ruby VM for the reasons written in the comment:

https://github.com/sorbet/sorbet/blob/d6aefd59ae07dba98e0132cf8b0fff727a8344be/compiler/IREmitter/NameBasedIntrinsics.cc#L852-L857

I can say with some certainty that the author was only thinking about symbol-based intrinsics, and not thinking about final methods when they wrote that.

We didn't really notice because the gain on the lspace benchmark (which said PR was intended to speedup) was so significant, and we didn't have any tests.  But today I noticed that removing `sig(:final)` from the lspace benchmark wasn't having any effect, and so I went hunting.

This PR should restore the functionality of generating direct calls for `[]`-methods (and other similar methods that we have untyped specializations for).  Arguably we should maybe move final methods to attempt codegen before symbol-based intrinsics, but seeing as how none of the methods we'd write symbol-based intrinsics for could be final, I'm not too worried about this.

lspace benchmark before this PR:

```
ruby vm startup time: .092
stripe/while_10_000_000.rb      .218    .099
stripe/lspace.rb        .892    .474
stripe/lspace.rb - baseline     .674    .375
```

lspace benchmark after this PR:

```
ruby vm startup time: .093
stripe/while_10_000_000.rb      .223    .109
stripe/lspace.rb        .899    .417
stripe/lspace.rb - baseline     .676    .308
```

...and we are now officially more than 2x faster on this benchmark. :tada: 

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/sends.cc
+++ b/compiler/IREmitter/sends.cc
@@ -26,6 +26,21 @@ using namespace std;
 namespace sorbet::compiler {
 namespace {
 
+llvm::Value *tryNameBasedIntrinsic(MethodCallContext &mcctx) {
+    for (auto nameBasedIntrinsic : NameBasedIntrinsicMethod::definedIntrinsics()) {
+        if (!absl::c_linear_search(nameBasedIntrinsic->applicableMethods(mcctx.cs), mcctx.send->fun)) {
+            continue;
+        }
+
+        if (mcctx.blk.has_value() && nameBasedIntrinsic->blockHandled == Intrinsics::HandleBlock::Unhandled) {
+            continue;
+        }
+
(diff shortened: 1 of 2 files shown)


jez (approved):

froydnj:

→ https://go/builds/bui_KhOwx05YGy7FUO
→ https://go/builds/bui_KhOw1eEibJITus

PR #4947, opened 2021-12-01
--stripe-packages: typecheck performance skip modules in package files

ngroman:
### Motivation
Typechecking degraded significantly (%50) when `--stripe-packages` was enabled on the Stripe codebase. The materialized modules we create do not actually have send's in the that need type-checking so these can be skipped (AFAIK).

Note: I tried simply skipping typechecking of `__package.rb` files completely but that ran into issues with some of the method validation we do specifically `test/test_LSPTests/testdata/packager/invalid_imports_and_exports `.

### Test plan

- [x] Re-do timing on stripe codebase.

See included automated tests.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -544,6 +544,12 @@ void typecheckOne(core::Context ctx, ast::ParsedFile resolved, const options::Op
         return;
     }
 
+#ifndef SORBET_REALMAIN_MIN
+    if (f.data(ctx).isPackage()) {
+        resolved = packager::Packager::removePackageModules(ctx, move(resolved));
+    }
+#endif
+
     resolved = definition_validator::runOne(ctx, std::move(resolved));
 
     resolved = class_flatten::runOne(ctx, move(resolved));
@@ -567,12 +573,6 @@ void typecheckOne(core::Context ctx, ast::ParsedFile resolved, const options::Op
         }
         return;
     }
-#ifndef SORBET_REALMAIN_MIN
-    if (f.data(ctx).isPackage()) {
-        // Must be after `class_flatten`
-        resolved = packager::Packager::removePackageModules(ctx, move(resolved));
-    }
-#endif
 
     Timer timeit(ctx.state.tracer(), "typecheckOne", {{"file", string(f.data(ctx).path())}});
     try {
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -581,7 +581,17 @@ void typecheckOne(core::Context ctx, ast::ParsedFile resolved, const options::Op
         }
         CFGCollectorAndTyper collector(opts);
         {
-            ast::ShallowMap::apply(ctx, collector, move(resolved.tree));
+            if (f.data(ctx).isPackage()) {
+#ifndef SORBET_REALMAIN_MIN
+                auto removedMods = packager::Packager::removePackageModules(ctx, resolved);
+                result.tree = ast::TreeMap::apply(ctx, collector, move(resolved.tree));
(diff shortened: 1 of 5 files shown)


nroman-stripe:

→ https://go/builds/bui_KhQm0PCdQpn6l2
→ https://go/builds/bui_KhQmSXcPKW9dxj

main/pipeline/pipeline.cc
```diff
@@ -572,7 +572,17 @@ ast::ParsedFile typecheckOne(core::Context ctx, ast::ParsedFile resolved, const
...
+            if (f.data(ctx).isPackage()) {
+#ifndef SORBET_REALMAIN_MIN
+                auto removedMods = packager::Packager::removePackageModules(ctx, resolved);
+                result.tree = ast::TreeMap::apply(ctx, collector, move(resolved.tree));
+                result = packager::Packager::replacePackageModules(ctx, move(result), move(removedMods));
+#else
+                ENFORCE(false, "Should never happen");
+#endif
+            } else {
+                result.tree = ast::TreeMap::apply(ctx, collector, move(resolved.tree));
```
jvilk-stripe, line 584: Heads up that I opened https://github.com/sorbet/sorbet/pull/4951 after seeing that we are doing a full treemap here and not a shallowmap. That will conflict with this PR.

I don't think that my change will fix the packager overhead, though -- ShallowMap doesn't recur into method bodies but still recurs into class bodies.

packager/packager.cc
```diff
@@ -23,6 +23,27 @@ namespace {
...
+    }
+    return false;
+}
+
+class PrunePackageModules final {
+public:
+    ast::ExpressionPtr postTransformClassDef(core::Context ctx, ast::ExpressionPtr tree) {
+        auto &klass = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+        if (isPackageModule(ctx, klass.symbol)) {
+            return ast::MK::EmptyTree();
```
ngroman, line 41: In non-LSP mode it'd be faster if I just leaked these (a decent amount parallelized time is spent de-allocating), but not sure if there's a reasonable way to do that.

nroman-stripe:
Ended up changing this to work a bit differently and now get much better performance.

packager/packager.cc
```diff
@@ -23,6 +23,27 @@ namespace {
...
+            return true;
+        }
+        modName = modName.data(gs)->owner;
+    }
+    return false;
+}
+
+class PrunePackageModules final {
+public:
+    ast::ExpressionPtr postTransformClassDef(core::Context ctx, ast::ExpressionPtr tree) {
```
jvilk-stripe, line 38: If you did this as `preTransformClassDef` then you would avoid traversing the class... but you wouldn't be able to return an `EmptyTree` since `preTransformClassDef` expects that you return a `ClassDef`:

https://github.com/sorbet/sorbet/blob/5279843c/ast/treemap/treemap.h#L168

idk if it would actually make things faster, but... you could hollow out the class in `preTransformClassDef` and return `EmptyTree` in `postTransformClassDef`.

packager/packager.cc
jvilk-stripe: Nah, don't leak memory. The only way to make deallocation faster is to make smaller trees!

packager/packager.cc
ngroman: I added a pre-step to clear the rhs. Also had to store a stack of "was this a package module" to avoid doing the check more than once. This made it slightly faster.

packager/packager.cc
jvilk-stripe: I take it back. I understand your issue here. On master there is now a 'leak trees' boolean passed to typecheck!

jvilk-stripe (approved):

PR #4948, opened 2021-12-01
Shrink the size of ast::ConstantLit by 16 bytes.

jvilk-stripe:
Shrink the size of ast::ConstantLit by 16 bytes. Reduces Sorbet's max RSS on Stripe's codebase by 4.5%.

`resolutionScopes` is empty unless the constant failed to resolve. Then, `symbol` is set to `StubModule` and the scope at which the resolution failed is put into `resolutionScopes`.

We can remove another 8 bytes if we use a tagged pointer so that, when the symbol is resolved, we store the `symbol` in this field. Otherwise, we store a pointer to `resolutionScopes`.

Why 8 bytes? Although `symbol` is only 4 bytes, there is 4 bytes of padding between it and `resolutionScopes`!

### Motivation

Reduce memory consumption.

### Test plan

Covered by existing tests.


ast/Trees.cc
```diff
@@ -686,10 +686,10 @@ string ConstantLit::showRaw(const core::GlobalState &gs, int tabs) {
     printTabs(buf, tabs + 1);
     fmt::format_to(std::back_inserter(buf), "orig = {}\n",
                    this->original ? this->original.showRaw(gs, tabs + 1) : "nullptr");
-    if (!resolutionScopes.empty()) {
+    if (resolutionScopes != nullptr && !resolutionScopes->empty()) {
```
jez, line 689: What do you think about ENFORCEing here that if `resolutionScopes` is not nullptr, that it's not empty?

jvilk-stripe: Works for me.

jez (approved):
lgtm, go nuts on adding ENFORCEs

ast/Trees.cc
jvilk-stripe: Added. Also added to `_sanityCheck` on `ConstantLit`.

PR #4952, opened 2021-12-02
Typecheck: Destroy trees passed to it in parallel.

jvilk-stripe:
Typecheck: Destroy trees passed to it in parallel.

There is no reason for typecheck to keep the trees around.

Avoids destructing all ASTs on a single thread.

### Motivation

Potentially speed. It also simplifies an optimization @ngroman is doing to remove package stuff from ASTs prior to CFG running.

### Test plan

N/A

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -533,12 +533,11 @@ vector<ast::ParsedFile> index(unique_ptr<core::GlobalState> &gs, vector<core::Fi
     return ret;
 }
 
-ast::ParsedFile typecheckOne(core::Context ctx, ast::ParsedFile resolved, const options::Options &opts) {
-    ast::ParsedFile result{ast::MK::EmptyTree(), resolved.file};
+void typecheckOne(core::Context ctx, ast::ParsedFile resolved, const options::Options &opts) {
     core::FileRef f = resolved.file;
 
     if (opts.stopAfterPhase == options::Phase::NAMER) {
-        return result;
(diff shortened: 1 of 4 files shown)


main/pipeline/pipeline.cc
```diff
@@ -572,7 +571,7 @@ ast::ParsedFile typecheckOne(core::Context ctx, ast::ParsedFile resolved, const
         }
         CFGCollectorAndTyper collector(opts);
         {
-            result.tree = ast::TreeMap::apply(ctx, collector, move(resolved.tree));
+            ast::TreeMap::apply(ctx, collector, move(resolved.tree));
```
froydnj, line 574: (non-blocking) It would be nice if we had a different name here like `sideEffectfulMap` to indicate that we only care about side effects and the `TreeMap` is deliberately not doing any transformations.

froydnj (approved):
Very nice.

I think "Destroy trees passed to it in parallel" is an accurate if slightly confusing way to describe the patch, but I don't have anything better.  ("Avoid returning unnecessary values"?)

jvilk-stripe:
I realize now that this defeats our "intentionally leak memory" optimization in CLI mode, so it noticeably regresses typechecking performance!

I think I need to add an option to leak memory inside `typecheck`.

jvilk-stripe:
@froydnj I had to reintroduce leaking memory in CLI mode, so I requested a re-review.

froydnj (approved):

PR #4954, opened 2021-12-02
WIP: Remove 8 bytes from ConstantLit by packing symbol and resolutionScopes into 8 bytes

jvilk-stripe:
WIP: Remove 8 bytes from ConstantLit by packing symbol and resolutionScopes into.8 bytes.

After resolver, `ConstantLit` is in one of two states:

* **Successfully resolved**, in which `symbol` is set to the symbol that the constant resolved to and `resolutionScopes` is empty.
* **Unsuccessfully resolved**, in which `symbol` is always set to `StubModule` and `resolutionScopes` contains the scopes that failed to find the constant.

It only ever needs `symbol` or `resolutionScopes` -- but not both.

In this PR, I pack both of these into 8 bytes using a tagged pointer, and shave 8 bytes off of every `ConstantLit`. The tag bits on the pointer indicate if the pointer encodes a `SymbolRef` value, or a pointer to a `std::vector<SymbolRef>`.

Hmm, I'm only detecting a 0.6% improvement over my previous PR that shaved 12 bytes off every `ConstantLit`...

### Motivation

Reduce Sorbet's memory consumption.

### Test plan

Covered by existing tests.


froydnj:
Closing in favor of #8568.

froydnj:
(We landed this in #8536.)

PR #4958, opened 2021-12-03
Fix typecheck call in fuzz_dash_e.cc

aprocter-stripe:
Fixes a call to `typecheck` in `fuzz_dash_e.cc` (this function returns `void` since #4952).

### Motivation
Fix compile failure that shows up with `./bazel test //... --config=dbg`

### Test plan
Verified offline that compilation no longer fails with `./bazel test //... --config=dbg`


jez (approved):

jez:
Why wasn't this caught by CI?

aprocter-stripe:
@jez Not really sure... my thinking was that it had something to do with the fact I was passing `--config=dbg` but I just tried without and that fails as well.

aprocter-stripe:
Could it be because `fuzz_dash_e` is tagged `manual`? https://github.com/sorbet/sorbet/blob/ba3f461aebdbcb47c35b330fc9bc58aae91e6f4f/test/fuzz/BUILD#L4

PR #4960, opened 2021-12-03
Improve LSP responsiveness by always destroying ASTs in parallel during cancelation

jvilk-stripe:
Improve LSP responsiveness by always destroying ASTs in parallel during cancelation.

When LSP is typechecking the world and a new edit comes in that causes LSP to cancel that typecheck, the code used to destroy all ASTs on a single thread, which is very slow on large projects and resulted in a noticeable pause.

This change enforces that cancelation always destroys trees in parallel by making the only way to construct a `ast::ParsedFilesOrCancelled` object in the "cancelled" state is to call a new static function that destroys the trees passed to it in parallel.

Note that `pipeline::typecheck` always destroys ASTs in parallel after #4952 landed, so this change only impacts `namer` and `resolver`.

One negative of the way that I implemented this is that `ast/Trees.cc` now contains WorkerPool-related code to implement `ast::ParsedFilesOrCancelled::cancel`. I'd be happy to entertain alternative code organization ideas.

### Motivation

Improve Sorbet LSP responsiveness.

### Test plan

Cancelation is tested by existing tests. I also manually tested it.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -1416,6 +1416,8 @@ ParsedFilesOrCancelled::ParsedFilesOrCancelled(std::vector<ParsedFile> &&trees) 
 
 ParsedFilesOrCancelled ParsedFilesOrCancelled::cancel(std::vector<ParsedFile> &&trees, WorkerPool &workers) {
     if (!trees.empty()) {
+        // N.B.: `workers.size()` can be `0` when threads are disabled, which would result in undefined behavior for
+        // `BlockingCounter`.
         absl::BlockingCounter threadBarrier(std::max(workers.size(), 1));
         auto fileq = make_shared<ConcurrentBoundedQueue<ast::ParsedFile>>(trees.size());
         for (auto &tree : trees) {
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -1,4 +1,7 @@
 #include "ast/Trees.h"
+#include "absl/synchronization/blocking_counter.h"
+#include "common/concurrency/ConcurrentQueue.h"
+#include "common/concurrency/WorkerPool.h"
 #include "common/formatting.h"
 #include "common/typecase.h"
 #include "core/Symbols.h"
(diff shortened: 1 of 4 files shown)


ast/Trees.cc
```diff
@@ -1411,6 +1414,31 @@ string BlockArg::nodeName() {
 ParsedFilesOrCancelled::ParsedFilesOrCancelled() : trees(nullopt){};
 ParsedFilesOrCancelled::ParsedFilesOrCancelled(std::vector<ParsedFile> &&trees) : trees(move(trees)) {}
 
+ParsedFilesOrCancelled ParsedFilesOrCancelled::cancel(std::vector<ParsedFile> &&trees, WorkerPool &workers) {
+    if (!trees.empty()) {
+        absl::BlockingCounter threadBarrier(std::max(workers.size(), 1));
```
jvilk-stripe, line 1419: Note: I'm pretty sure `workers.size()` returns 0 when threading is disabled, hence the `std::max` with `1`. I can add a comment if desired.

ast/Trees.h
```diff
@@ -235,11 +239,16 @@ struct ParsedFile {
 class ParsedFilesOrCancelled final {
 private:
     std::optional<std::vector<ParsedFile>> trees;
+    ParsedFilesOrCancelled();
```
jvilk-stripe, line 242: I made this private to enforce that callsites must destroy trees in parallel, in the hopes of preventing a regression in the future.

ast/Trees.cc
aprocter-stripe: Comment wouldn't hurt, I spose!

ast/Trees.cc
```diff
@@ -1411,6 +1414,31 @@ string BlockArg::nodeName() {
...
+        auto fileq = make_shared<ConcurrentBoundedQueue<ast::ParsedFile>>(trees.size());
+        for (auto &tree : trees) {
+            fileq->push(move(tree), 1);
+        }
+
+        workers.multiplexJob("deleteTrees", [fileq, &threadBarrier]() {
+            {
+                ast::ParsedFile job;
+                for (auto result = fileq->try_pop(job); !result.done(); result = fileq->try_pop(job)) {
+                    // Do nothing; allow the destructor of `ast::ParsedFile` to run.
```
aprocter-stripe, line 1429:
```suggestion
                    // Do nothing; allow the destructor of `ast::ParsedFile` to run for `job`.
```

aprocter-stripe (approved):
LGTM! Just a couple of non-blocking cosmetic suggestions.

I don't find the introduction of `WorkerPool`-related code into `Trees.cc` all that bothersome, but not sure if anyone else will have stronger opinions.

PR #4961, opened 2021-12-03
add arm64 mac configuration bits

froydnj:
### Motivation

This is all very cargo-culty, because I don't actually understand how one would go about building Sorbet natively on a arm64 mac.  I think @vinistock and/or @Morriar have been working on this, though?

### Test plan

See included automated tests.


jez:
This ended up shipping at some point.

PR #4964, opened 2021-12-03
Optimize the final method fast path

elliottt:
- Add a helper for testing if a variable is an alias to a singleton
- Allow type assumptions on arbitrary TypePtr values
- Optimize type tests for final methods, and assume their results are typed correctly on the fast path

### Test plan

See included automated tests.


elliottt:

→ https://go/builds/bui_KiAekqmFVQwvZG
→ https://go/builds/bui_KiAeWdKv4x9az8

jez (approved):

PR #4970, opened 2021-12-04
Switch to C++20

jez:
### Motivation

Fun new language features!

### Test plan

See included automated tests.

I have a macOS branch build here:

<https://buildkite.com/sorbet/sorbet/builds/19102>

### DIFF
--- a/common/Exception.h
+++ b/common/Exception.h
@@ -43,7 +43,8 @@ public:
 
 class Exception final {
 public:
-    template <typename... TArgs> [[noreturn]] static bool raise(const TArgs &...args) __attribute__((noreturn));
+    template <typename... TArgs>
+    [[noreturn]] static bool raise(fmt::string_view formatString, const TArgs &...args) __attribute__((noreturn));
 
     [[noreturn]] static inline void notImplemented() {
         raise("Not Implemented");
@@ -58,15 +59,16 @@ public:
     }
     template <typename... TArgs>
     [[noreturn]] static inline bool enforce_handler(std::string_view check, std::string_view file, int line,
-                                                    std::string_view message, const TArgs &...args)
+                                                    fmt::string_view message, const TArgs &...args)
         __attribute__((noreturn)) {
-        raise("{}:{} enforced condition {} has failed: {}", file, line, check, fmt::format(message, args...));
+        raise("{}:{} enforced condition {} has failed: {}", file, line, check,
+              fmt::vformat(message, fmt::make_format_args(args...)));
     }
 };
 
-template <typename... TArgs> [[noreturn]] bool Exception::raise(const TArgs &...args) {
+template <typename... TArgs> [[noreturn]] bool Exception::raise(fmt::string_view formatString, const TArgs &...args) {
     Exception::failInFuzzer();
-    std::string message = fmt::format(args...);
+    std::string message = fmt::vformat(formatString, fmt::make_format_args(args...));
 
     if (message.size() > 0) {
         fatalLogger->error("Exception::raise(): {}\n", message);
--- a/common/Exception.h
+++ b/common/Exception.h
@@ -44,7 +44,7 @@ public:
 class Exception final {
 public:
     template <typename... TArgs>
-    [[noreturn]] static bool raise(fmt::string_view formatString, const TArgs &...args) __attribute__((noreturn));
+    [[noreturn]] static bool raise(fmt::format_string<TArgs...> fmt, TArgs &&...args) __attribute__((noreturn));
 
     [[noreturn]] static inline void notImplemented() {
         raise("Not Implemented");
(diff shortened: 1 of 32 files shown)


common/Exception.h
```diff
@@ -58,15 +59,16 @@ class Exception final {
...
+        raise("{}:{} enforced condition {} has failed: {}", file, line, check,
+              fmt::vformat(message, fmt::make_format_args(args...)));
     }
 };
 
-template <typename... TArgs> [[noreturn]] bool Exception::raise(const TArgs &...args) {
+template <typename... TArgs> [[noreturn]] bool Exception::raise(fmt::string_view formatString, const TArgs &...args) {
     Exception::failInFuzzer();
-    std::string message = fmt::format(args...);
+    std::string message = fmt::vformat(formatString, fmt::make_format_args(args...));
```
jez, line 71: I want to try this again to see if I can get this to type check 🤔 

I was having some problems getting this to type check and had to go back to the runtime-checked formatting ones, instead of the constexpr ones.

core/Error.h
```diff
@@ -36,8 +35,11 @@ class ErrorColors {
 
 public:
     ErrorColors() = delete;
-    template <typename... Args> static std::string format(std::string_view msg, const Args &...args) {
-        return fmt::format(replaceAll(msg, coloredPatternSigil, coloredPatternReplace), args...);
+    template <typename... Args> static std::string format(fmt::format_string<Args...> msg, Args &&...args) {
+        fmt::string_view msgFmtStrView = msg;
+        std::string_view inWhat(msgFmtStrView.data(), msgFmtStrView.size());
+        return fmt::vformat(replaceAll(inWhat, coloredPatternSigil, coloredPatternReplace),
+                            fmt::make_format_args(args...));
```
jez, line 42: I think that this one basically has to use vformat because there's no (easy?) way for us to make `replaceAll` consteval.

core/Types.h
```diff
@@ -786,6 +786,7 @@ class TypeAndOrigins final {
 
     ~TypeAndOrigins() noexcept;
     TypeAndOrigins() = default;
+    TypeAndOrigins(TypePtr type, InlinedVector<Loc, 1> origins) : type(type), origins(origins) {}
```
jez, line 789: There were a bunch of these where clang required these in places where it didn't before, so i'm not sure when this is required and when it's not to get something like this working:

```
auto foo = TypeAndOrigins{type, origins};
```

main/autogen/autogen.cc
```diff
@@ -14,8 +14,7 @@ namespace sorbet::autogen {
 class AutogenWalk {
     vector<Definition> defs;
     vector<Reference> refs;
-    vector<core::NameRef>
-    requires;
+    vector<core::NameRef> requireStatements;
```
jez, line 17: I pulled these changes into #4968, a rebase will drop them.

parser/Parser.cc
```diff
@@ -42,7 +42,8 @@ class ErrorToError {
             core::Loc loc(file, translatePos(diag.location().beginPos, maxOff - 1),
                           translatePos(diag.location().endPos, maxOff));
             if (auto e = gs.beginError(loc, core::errors::Parser::ParserError)) {
-                e.setHeader("{}", fmt::format(dclassStrings[(int)diag.error_class()], diag.data()));
+                e.setHeader("{}",
+                            fmt::vformat(dclassStrings[(int)diag.error_class()], fmt::make_format_args(diag.data())));
```
jez, line 46: the `dclassStrings` is an external linkage `const char *[]` so it's not constexpr

parser/tools/generate_ast.cc
```diff
@@ -1036,7 +1036,7 @@ void emitNodeClassfile(ostream &out, NodeDef &node) {
                     out << "    fmt::format_to(std::back_inserter(buf), \", \\\"{}\\\"\", " << arg.name
                         << ".toString(gs));\n";
                 } else {
-                    out << "    fmt::format_to(std::back_inserter(buf), \", :\" + JSON::escape(" << arg.name
+                    out << "    fmt::format_to(std::back_inserter(buf), \", :{}\", JSON::escape(" << arg.name
```
jez, line 1039: This caused a fun Clang 12 bug that I couldn't reproduce on a small example, but which fired for this file every time I tried to build.

The thing is: this `operator+` is not consteval, so it was technically an error anyways, and would have had to be fixed like this. but the crazy part is that it somehow crashed clang

third_party/externals.bzl
```diff
@@ -39,10 +39,10 @@ def register_sorbet_dependencies():
 
     http_archive(
         name = "spdlog",
-        urls = _github_public_urls("gabime/spdlog/archive/61ed2a670e40e9a865edc20b088422a469bdab9e.zip"),  # v1.9.0
-        sha256 = "0ec124d1bb4bb57a7f43f814622f7c907619e3cacbb6e04c62634a98880f5971",
+        urls = _github_public_urls("gabime/spdlog/archive/eb3220622e73a4889eee355ffa37972b3cac3df5.zip"),  # v1.9.2
+        sha256 = "b7570488bdd94ab6d3653bc324d8ed7976d9f3a2f035eb2e969ebcaad3b0d5c7",
         build_file = "@com_stripe_ruby_typer//third_party:spdlog.BUILD",
-        strip_prefix = "spdlog-61ed2a670e40e9a865edc20b088422a469bdab9e",
+        strip_prefix = "spdlog-eb3220622e73a4889eee355ffa37972b3cac3df5",
```
jez, line 45: There were some C++20 changes between 1.9.0 and 1.9.2

i'm not sure if we needed those changes in the end, but almost all these bugs were c++20 and fmtlib related so i figured i'd upgrade.

compiler/Core/CompilerState.cc
```diff
@@ -114,12 +114,4 @@ void CompilerState::runCheapOptimizations(llvm::Function *func) {
...
 }
 
-void failCompilation(const core::GlobalState &gs, const core::Loc &loc, ConstExprStr msg) {
-    if (auto e = gs.beginError(loc, core::errors::Compiler::Unanalyzable)) {
-        e.setHeader(msg);
-    }
-
-    throw AbortCompilation(msg.str);
-}
-
```
jez, line 124: Moved to its own file, because this needs `gs.beginError` (not enough to just forward declare GlobalState) but also needs to be a template to support format_string checking (required by `setHeader`—there is no non-const version).

compiler/IREmitter/SymbolBasedIntrinsics.cc
```diff
@@ -84,7 +85,7 @@ class CMethod final {
                 current = overload;
             }
 
-            ENFORCE(false, "The method `{}#{}` (or an overload) does not return `{}`", primaryMethod.show(gs),
+            ENFORCE(false, "The method `{}` (or an overload) does not return `{}`", primaryMethod.show(gs),
```
jez, line 88: This was a bug, caught at compile time!!

jez:
Pre-work changes landed in #4979 

Emscripten failures are independent of C++20 (the compiler doesn't support `-std=c++20` yet) so we will have to bump that pre-emptively in another PR before we can land this, which makes this PR empty.

PR #4971, opened 2021-12-05
[LSP] Fix crash in DocumentHighlightTask

ilyailya:
This PR is a part of work to get a document symbols feature to a usable state. 
I took a look into the `/test/testdata/lsp/document_symbols_crash.rb` test:
https://github.com/sorbet/sorbet/blob/3bd289b6e8b6477864a83aeab06b07fb76eb93d6/test/testdata/lsp/document_symbols_crash.rb#L1-L4

What the test says if you add or delete a character in a property declaration, sorbet will crash. 
I assume it will work the same way for method declarations too.

The crash happens, because the `DocumentHighlightTask` tries to re-highlight the document 
and hits the `ENFORCE`, which is too tight.  I've tried disabling the document highlight feature, 
but the issue still occurred, because some other task called `queryByLoc`, 
which boils down to the same too tight `ENFORCE`

### Motivation

Crash

### Test plan

I'm not sure how to cover this kind of crash in a test, if you have any ideas, please, let me know!

### DIFF
--- a/main/lsp/DefLocSaver.cc
+++ b/main/lsp/DefLocSaver.cc
@@ -55,8 +55,12 @@ ast::ExpressionPtr DefLocSaver::postTransformUnresolvedIdent(core::Context ctx, 
         core::ClassOrModuleRef klass;
         // Logic cargo culted from `global2Local` in `walker_build.cc`.
         if (id.kind == ast::UnresolvedIdent::Kind::Instance) {
-            ENFORCE(ctx.owner.isMethod());
-            klass = ctx.owner.owner(ctx).asClassOrModuleRef();
+            ENFORCE(ctx.owner.isMethod() || ctx.owner.isClassOrModule());
+            if (ctx.owner.isMethod()) {
+              klass = ctx.owner.owner(ctx).asClassOrModuleRef();
+            } else if (ctx.owner.isClassOrModule()) {
+              klass = ctx.owner.asClassOrModuleRef();
+            }
         } else {
             // Class var.
             klass = ctx.owner.enclosingClass(ctx);
--- a/main/lsp/DefLocSaver.cc
+++ b/main/lsp/DefLocSaver.cc
@@ -53,7 +53,6 @@ ast::ExpressionPtr DefLocSaver::postTransformUnresolvedIdent(core::Context ctx, 
     auto &id = ast::cast_tree_nonnull<ast::UnresolvedIdent>(tree);
     if (id.kind == ast::UnresolvedIdent::Kind::Instance || id.kind == ast::UnresolvedIdent::Kind::Class) {
         core::ClassOrModuleRef klass;
-        // Logic cargo culted from `global2Local` in `walker_build.cc`.
         if (id.kind == ast::UnresolvedIdent::Kind::Instance) {
             ENFORCE(ctx.owner.isMethod() || ctx.owner.isClassOrModule());
             if (ctx.owner.isMethod()) {


main/lsp/DefLocSaver.cc
```diff
@@ -53,10 +53,13 @@ ast::ExpressionPtr DefLocSaver::postTransformUnresolvedIdent(core::Context ctx,
     auto &id = ast::cast_tree_nonnull<ast::UnresolvedIdent>(tree);
     if (id.kind == ast::UnresolvedIdent::Kind::Instance || id.kind == ast::UnresolvedIdent::Kind::Class) {
         core::ClassOrModuleRef klass;
-        // Logic cargo culted from `global2Local` in `walker_build.cc`.
```
ilyailya, line 56: This comment seems to be not true anymore

jez:
@jvilk how do document symbols tests work?

ilyailya:
@jez from what I saw, document symbols tests basically open the file, and send the document symbols request to the sorbet, and assert the results. There is a capability to apply the update on a file. But currently we don't check the health of the language server between updates in any way.

main/lsp/DefLocSaver.cc
jvilk-stripe: https://github.com/sorbet/sorbet/blob/5bdbbe9e/cfg/builder/builder_walk.cc#L65

Looks like the code is still there but renamed. 🤔 

How does that code not crash?

main/lsp/DefLocSaver.cc
```diff
@@ -53,10 +53,13 @@ ast::ExpressionPtr DefLocSaver::postTransformUnresolvedIdent(core::Context ctx,
...
         if (id.kind == ast::UnresolvedIdent::Kind::Instance) {
-            ENFORCE(ctx.owner.isMethod());
-            klass = ctx.owner.owner(ctx).asClassOrModuleRef();
+            ENFORCE(ctx.owner.isMethod() || ctx.owner.isClassOrModule());
+            if (ctx.owner.isMethod()) {
+                klass = ctx.owner.owner(ctx).asClassOrModuleRef();
+            } else if (ctx.owner.isClassOrModule()) {
+                klass = ctx.owner.asClassOrModuleRef();
+            }
```
jvilk-stripe, line 62: This should be equivalent, and I think captures the intent here:

```suggestion
            klass = ctx.owner.enclosingClass(ctx);
```

See https://github.com/sorbet/sorbet/blob/5bdbbe9e/core/Symbols.cc#L2073 for implementation

main/lsp/DefLocSaver.cc
```diff
@@ -53,10 +53,13 @@ ast::ExpressionPtr DefLocSaver::postTransformUnresolvedIdent(core::Context ctx,
...
         } else {
             // Class var.
             klass = ctx.owner.enclosingClass(ctx);
```
jvilk-stripe, line 65: heh, maybe you could just simplify this code to have the enforce and to otherwise call `enclosingClass`.

main/lsp/DefLocSaver.cc
jvilk-stripe: Oh, I think I know. I think this code runs before a fake `static-init` method is created to contain this AST node, so it is owned by the class.

Good catch!

jvilk-stripe (approved):
Good catch! I think you can simplify this code even further with `enclosingClass` (see comment).

main/lsp/DefLocSaver.cc
ilyailya: TIL: `enclosingClass`

PR #4974, opened 2021-12-06
Move method symbols into a new Method class

jvilk-stripe:
Move method symbols into a new Method class.

* Speeds up Sorbet typechecking on large projects by reducing the cost of defining/allocating method symbols.
* Slightly reduces memory consumption due to smaller method symbol objects.
* Reduces `SymbolRef` casts due to more specifically-typed APIs (e.g., `Method::owner` is a `ClassOrModuleRef`, `Method::typeArguments` is a `vector` of `TypeArgumentRef`, and `Symbol::typeMembers()` is now a `vector` of `TypeMemberRef`).
* Removes some `SymbolRef` methods that are no longer required.

### Motivation

In addition to the above, this change moves errors that previously occurred at runtime (calling a method-only method on Symbol) to ones that happen at compile time (method methods can only run on `Method` objects).

### Test plan

Covered by existing tests.

### DIFF
--- a/rbi/stdlib/rubygems.rbi
+++ b/rbi/stdlib/rubygems.rbi
@@ -3512,7 +3512,7 @@ class Gem::Command
   #   "FILE          name of file to find"
   # end
   # ```
-  def arguments(); end
+  def arguments; end
 
   # True if `long` begins with the characters from `short`.
   def begins?(long, short); end
--- a/rbi/stdlib/rubygems.rbi
+++ b/rbi/stdlib/rubygems.rbi
@@ -3512,7 +3512,7 @@ class Gem::Command
   #   "FILE          name of file to find"
   # end
   # ```
-  def arguments; end
+  def arguments(); end
 
   # True if `long` begins with the characters from `short`.
   def begins?(long, short); end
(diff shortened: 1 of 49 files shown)


rbi/stdlib/rubygems.rbi
```diff
@@ -3512,7 +3512,7 @@ class Gem::Command
   #   "FILE          name of file to find"
   # end
   # ```
-  def arguments(); end
+  def arguments; end
```
jvilk-stripe, line 3515:
```suggestion
  def arguments(); end
```
This was a bad search and replace.

resolver/type_syntax.cc
```diff
@@ -673,7 +673,7 @@ TypeSyntax::ResultType interpretTCombinator(core::Context ctx, const ast::Send &
                 }
             }
 
-            if (!ctx.owner.isSingletonClass(ctx)) {
+            if (!ctx.owner.isClassOrModule() || !ctx.owner.asClassOrModuleRef().data(ctx)->isSingletonClass(ctx)) {
```
elliottt, line 676: This is a little verbose, though I guess if you already have a `ClassOrModuleRef` it's not so bad.

elliottt (approved):
Looks good to me! Super excited for splitting up Symbol more :D

PR #4977, opened 2021-12-07
StripePackages: helpful errors/fixes for unresolved export

ngroman:
If a user attempts to `export` a symbol and it does not resolved attempt to be more helpful.

In one scenario, this is a typo and is similar enough to something actually defined in this package. In that case we use the normal fuzzy logic and suggest replacements among candidates from _this_ package.

If no reasonable matches are found, instead add a note to the error:
```
__package.rb:7: Unable to resolve constant `WildlyMisspelled` https://srb.help/5002
     7 |  export Foo::BasePkg::WildlyMisspelled
  Note:
    To be exported it must be defined in package `Foo::BasePkg`
```

### Motivation
UX

### Test plan

See included automated tests.

### DIFF
--- a/resolver/SuggestPackage.cc
+++ b/resolver/SuggestPackage.cc
@@ -79,6 +79,15 @@ public:
         }
     }
 
+    bool tryPackageSpecCorrections(core::ErrorBuilder &e, ast::UnresolvedConstantLit &unresolved) {
+        if (isUnresolvedExport(unresolved)) {
+            // TODO(nroman-stripe) handle bad export
+        } else {
+            // TODO(gdritter-stripe) handle bad import
+        }
+        return false;
+    }
+
 private:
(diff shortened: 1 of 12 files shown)


resolver/SuggestPackage.cc
```diff
@@ -109,6 +129,50 @@ class PackageContext final {
...
+        {
+            // Remove results defined outside of this package:
+            auto it = remove_if(matches.begin(), matches.end(),
+                                [&](auto &m) -> bool { return !currentPkg.ownsSymbol(ctx, m.symbol); });
+            matches.erase(it, matches.end());
+            if (matches.size() > 4) {
+                matches.resize(4);
+            }
+        }
+        if (matches.size() > 0) {
```
elliottt, line 154:
```suggestion
        if (!matches.empty()) {
```

elliottt (approved):

PR #4980, opened 2021-12-07
shellcheck: Use Bazel, upgrade to 0.8.0

jez:
### Motivation

- Everyone running lint_sh.sh is going to get the right version of shellcheck
- Don't have to bump the build image to get new shellcheck versions
- Upgrade to 0.8.0 for good measure

  Changelog: https://github.com/koalaman/shellcheck/blob/master/CHANGELOG.md#v080---2021-11-06

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet/test/hidden-method-finder/hidden_methods_bazel.sh
+++ b/gems/sorbet/test/hidden-method-finder/hidden_methods_bazel.sh
@@ -42,7 +42,7 @@ test_name=$3
 test_dir="${repo_root}/gems/sorbet/test/hidden-method-finder/${test_name}"
 
 # NOTE: using rlocation here because this script gets run from a genrule
-# shellcheck disable=SC1090
+# shellcheck source=SCRIPTDIR/logging.sh
 source "$(rlocation com_stripe_ruby_typer/gems/sorbet/test/hidden-method-finder/logging.sh)"
 
 # ----- Environment setup and validation -----
--- a/gems/sorbet/test/snapshot/run_one.sh
+++ b/gems/sorbet/test/snapshot/run_one.sh
@@ -182,6 +182,7 @@ fi
   #
   # note: redirects stderr before the pipe
   info "├─ Running srb"
+  # shellcheck disable=SC2048
   if ! SRB_YES=1 ${test_cmd[*]} < /dev/null 2> "err.log" > "out.log"; then
     error "├─ srb failed."
     error "├─ stdout (out.log):"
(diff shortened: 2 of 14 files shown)


gems/sorbet/test/hidden-method-finder/hidden_methods_bazel.sh
```diff
@@ -42,7 +42,7 @@ test_name=$3
 test_dir="${repo_root}/gems/sorbet/test/hidden-method-finder/${test_name}"
 
 # NOTE: using rlocation here because this script gets run from a genrule
-# shellcheck disable=SC1090
+# shellcheck source=SCRIPTDIR/logging.sh
```
aprocter-stripe, line 45: Nice to see these!

gems/sorbet/test/snapshot/run_one.sh
```diff
@@ -182,6 +182,7 @@ fi
   #
   # note: redirects stderr before the pipe
   info "├─ Running srb"
+  # shellcheck disable=SC2048
   if ! SRB_YES=1 ${test_cmd[*]} < /dev/null 2> "err.log" > "out.log"; then
```
aprocter-stripe, line 186: So tracing back the construction of `test_cmd` it does look like interpolation here could cause a problem if `$repo_root` contains spaces, since that goes into `$srb`. Does that sound right? If so, not sure it matters enough to fix it, but a quick comment might not hurt.

tools/scripts/lint_sh.sh
```diff
@@ -8,8 +8,24 @@ list_sh_src() {
     git ls-files -z -c -m -o --exclude-standard -- '*.sh' '*.bash'
 }
 
+
+unameOut="$(uname -s)"
+case "${unameOut}" in
+    Linux*)     platform="linux";;
+    Darwin*)    platform="mac";;
+    *)          exit 1
```
aprocter-stripe, line 16: This exits the whole script, right? An error message might be nice, I s'pose, or we could try the system shellcheck in that case.

aprocter-stripe (approved):
Looks great! None of my comments are blocking.

gems/sorbet/test/snapshot/run_one.sh
jez: These scripts are not used anymore (we re-implemented them using bazel) but i'm too sentimental to remove them.

aprocter-stripe: Ah right on :)

tools/scripts/lint_sh.sh
jez: I just copied the boilerplate we have in all our other scripts but i can change all of these in a follow up if you'd like

PR #4981, opened 2021-12-07
Add autofixes for exports

aisamanra:
### Motivation
This wires up the export autofix functionality with the existing export errors. To do so cleanly, it also tweaks the existing API for the autofix functionality: it now takes a `SymbolRef` instead of a `vector<NameRef>`, and updates the tests accordingly.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/SuggestPackage.cc
+++ b/resolver/SuggestPackage.cc
@@ -57,7 +57,17 @@ public:
             fmt::map_join(srcPkg.fullName(), "::", [&](auto nr) -> string { return nr.show(ctx); })));
         lines.emplace_back(
             core::ErrorLine::from(match.symbol.loc(ctx), "Constant `{}` is defined here:", match.symbol.show(ctx)));
-        // TODO(nroman-stripe) Add automatic fixers
+
+        core::SymbolRef sym = match.symbol;
+        vector<core::NameRef> name;
+        do {
+            name.emplace_back(sym.name(ctx));
+            sym = sym.owner(ctx);
+        } while(sym.exists() && sym != core::Symbols::root());
+        if (auto autocorrect = srcPkg.addExport(ctx, name, false)) {
+            e.addAutocorrect(std::forward<core::AutocorrectSuggestion>(*autocorrect));
+        }
+
         e.addErrorSection(core::ErrorSection(lines));
     }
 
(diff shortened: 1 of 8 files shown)


resolver/SuggestPackage.cc
```diff
@@ -57,8 +57,11 @@ class PackageContext final {
             fmt::map_join(srcPkg.fullName(), "::", [&](auto nr) -> string { return nr.show(ctx); })));
         lines.emplace_back(
             core::ErrorLine::from(match.symbol.loc(ctx), "Constant `{}` is defined here:", match.symbol.show(ctx)));
-        // TODO(nroman-stripe) Add automatic fixers
         e.addErrorSection(core::ErrorSection(lines));
+
+        if (auto autocorrect = srcPkg.addExport(ctx, match.symbol, false)) {
+            e.addAutocorrect(std::forward<core::AutocorrectSuggestion>(*autocorrect));
```
jez, line 63: ooc, what's the `std::forward` for?

jez (approved):

resolver/SuggestPackage.cc
aisamanra: The error without it is
```
resolver/SuggestPackage.cc:78:30: error: rvalue reference to type 'sorbet::core::AutocorrectSuggestion' cannot bind to lvalue of type 'std::optional<sorbet::core::AutocorrectSuggestion>::value_type' (aka 'sorbet::core::AutocorrectSuggestion')
            e.addAutocorrect(*autocorrect);
                             ^~~~~~~~~~~~
./core/Error.h:162:49: note: passing argument to parameter 'autocorrect' here
    void addAutocorrect(AutocorrectSuggestion &&autocorrect);
                                                ^
```
which happens because our `ErrorBuilder` has the method declaration
```c++
void addAutocorrect(AutocorrectSuggestion &&autocorrect);
```
so we need `std::forward` to turn the lvalue into an rvalue. I suppose I could use `std::move` instead, if we'd prefer.

jez: yeah i'd prefer `move` i think. thanks!

aisamanra: Replaced with `move`!

PR #4984, opened 2021-12-07
Stop reporting inconsistent usage of bare type error

jez:
### Motivation

This came up when helping someone at Stripe yesterday. These `bare type` errors
almost always mean there's a bug in Sorbet, not in the programmer's code.

In this case, the problem was that in almost all other cases, you can use
`T.untyped` to silence an error arising from attempting to type check code that
uses runtime type representations (like type aliases or T.any), etc. except in
the case where those runtime types were being used inside tuples.

That was needlessly inconsistent, so I've deleted the relevant code.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1828,22 +1828,13 @@ public:
     void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
         vector<TypePtr> elems;
         elems.reserve(args.args.size());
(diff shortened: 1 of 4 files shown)


elliottt (approved):

jez:

→ https://go/builds/bui_KjfDw9w4P2JFgE
→ https://go/builds/bui_KjfD8iGvMwQY01

PR #4985, opened 2021-12-07
Thread --trace-parser option through parser

jez:
### Motivation

This will make it easier to visualize what's happening inside the parser and
debug it.

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/trace_parser/trace_parser.out
+++ b/test/cli/trace_parser/trace_parser.out
@@ -1,12 +1,564 @@
 Starting parse
 Entering state 0
-Reading a token: Next token is token "end of file" ()
-Reducing stack by rule 3 (line 492):
+Reading a token: Next token is token "class" ()
+Shifting token "class" ()
+Entering state 2
+Reading a token: Next token is token tCONSTANT ()
+Shifting token tCONSTANT ()
+Entering state 126
+Reading a token: Next token is token tNL ()
+Reducing stack by rule 121 (line 1069):
+   $1 = token tCONSTANT ()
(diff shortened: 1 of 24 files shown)


third_party/parser/cc/capi.cc
```diff
@@ -1,101 +1,69 @@
-#include <ruby_parser/capi.hh>
```
jez, line 1: i found a couple of files that were not having clang-format run on them

i landed those changes as a separate commit, if you want to ignore this noise

i haven't looked into why they're not getting formatted automatically (but it's probably not that important)

test/cli/trace_parser/trace_parser.out
```diff
@@ -0,0 +1,30 @@
...
+-> $$ = nterm program ()
+Stack now 0
+Entering state 71
+Next token is token "end of file" ()
+Shifting token "end of file" ()
+Entering state 339
+Cleanup: popping token "end of file" ()
+Cleanup: popping nterm program ()
+test/cli/trace_parser.rb: File Not Found https://srb.help/1004
+Errors: 1
```
aprocter-stripe, line 30: Is the File Not Found expected here?

aprocter-stripe (approved):
LGTM with one question about the test output

test/cli/trace_parser/trace_parser.out
jez: no it's not thanks

jez: (fixed)

PR #4987, opened 2021-12-08
use `<stdint.h>` types everywhere

froydnj:
### Motivation

Using stuff directly from the standard library is generally preferred.

Other minor points are a) less confusion between the short types and Rust types and b) less confusion over whether `u8` is 8 bytes or 1 byte.

### Test plan

See included automated tests.


elliottt (approved):
🎉

PR #4988, opened 2021-12-08
rename `rubyBlockId` to `rubyRegionId`

froydnj:
### Motivation

I can't possibly be the only person who's gotten confused between Ruby blocks and basic blocks (whether in Sorbet CFG terms or compiler terms).  And sometimes `rubyBlockId` doesn't actually refer to proper Ruby blocks (cf. how exception handling works)!

We settled internally on "region" as a better (but not perfect!) replacement.

### Test plan

See included automated tests.

### DIFF
--- a/cfg/CFG.cc
+++ b/cfg/CFG.cc
@@ -31,12 +31,12 @@ int CFG::numLocalVariables() const {
     return this->localVariables.size();
 }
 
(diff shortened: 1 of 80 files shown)


aprocter-stripe (approved):
`$HIGHER_POWER` bless you for this, sir!

PR #4990, opened 2021-12-08
Optimize the fast path for index

elliottt:
Parallelize name substitution during indexing, and avoid constructing a substitution when it wouldn't be necessary.

### Motivation

Performance.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1811,9 +1811,9 @@ unique_ptr<GlobalState> GlobalState::copyForIndex() const {
 
     result->initEmpty();
 
+    // Options that might be used during indexing are manually copied over here
     result->files = this->files;
     result->fileRefByPath = this->fileRefByPath;
-
     result->silenceErrors = this->silenceErrors;
     result->autocorrect = this->autocorrect;
     result->ensureCleanStrings = this->ensureCleanStrings;
@@ -1822,6 +1822,10 @@ unique_ptr<GlobalState> GlobalState::copyForIndex() const {
     result->sleepInSlowPath = this->sleepInSlowPath;
     result->requiresAncestorEnabled = this->requiresAncestorEnabled;
     result->kvstoreUuid = this->kvstoreUuid;
+    result->errorUrlBase = this->errorUrlBase;
+    result->suppressedErrorClasses = this->suppressedErrorClasses;
+    result->onlyErrorClasses = this->onlyErrorClasses;
+    result->suggestUnsafe = this->suggestUnsafe;
     result->pathPrefix = this->pathPrefix;
 
     return result;
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1806,6 +1806,27 @@ unique_ptr<GlobalState> GlobalState::deepCopy(bool keepId) const {
     return result;
 }
 
+unique_ptr<GlobalState> GlobalState::copyForIndex() const {
+    auto result = make_unique<GlobalState>(this->errorQueue, this->epochManager);
+
+    result->initEmpty();
+
+    result->files = this->files;
+    result->fileRefByPath = this->fileRefByPath;
+
+    result->silenceErrors = this->silenceErrors;
+    result->autocorrect = this->autocorrect;
+    result->ensureCleanStrings = this->ensureCleanStrings;
+    result->runningUnderAutogen = this->runningUnderAutogen;
+    result->censorForSnapshotTests = this->censorForSnapshotTests;
+    result->sleepInSlowPath = this->sleepInSlowPath;
+    result->requiresAncestorEnabled = this->requiresAncestorEnabled;
+    result->kvstoreUuid = this->kvstoreUuid;
+    result->pathPrefix = this->pathPrefix;
+
+    return result;
+}
+
 string_view GlobalState::getPrintablePath(string_view path) const {
     // Only strip the path prefix if the path has it.
     if (path.substr(0, pathPrefix.length()) == pathPrefix) {
--- a/core/GlobalState.h
+++ b/core/GlobalState.h
@@ -235,6 +235,9 @@ public:
     std::unique_ptr<GlobalState> deepCopy(bool keepId = false) const;
     mutable std::shared_ptr<ErrorQueue> errorQueue;
 
+    // Copy the file table and other parts of GlobalState that are required for the indexing pass.
+    // NOTE: this very intentionally will not copy the symbol or name tables. The symbol tables aren't used or populated
+    // during indexing, and the name tables will only be written to.
     std::unique_ptr<GlobalState> copyForIndex() const;
 
     // Contains a path prefix that should be stripped from all printed paths.
--- a/core/GlobalState.h
+++ b/core/GlobalState.h
@@ -235,6 +235,8 @@ public:
     std::unique_ptr<GlobalState> deepCopy(bool keepId = false) const;
     mutable std::shared_ptr<ErrorQueue> errorQueue;
 
+    std::unique_ptr<GlobalState> copyForIndex() const;
+
     // Contains a path prefix that should be stripped from all printed paths.
     std::string pathPrefix;
     // Returns a string_view of the given path with the path prefix removed.
--- a/core/GlobalSubstitution.cc
+++ b/core/GlobalSubstitution.cc
@@ -10,8 +10,6 @@ GlobalSubstitution::GlobalSubstitution(const GlobalState &from, GlobalState &to)
 
     from.sanityCheck();
 
-    GlobalSubstitution::mergeFileTables(from, to);
-
     {
         UnfreezeNameTable unfreezeNames(to);
         utf8NameSubstitution.reserve(from.utf8Names.size());
--- a/core/GlobalSubstitution.cc
+++ b/core/GlobalSubstitution.cc
@@ -17,21 +17,8 @@ GlobalSubstitution::GlobalSubstitution(const GlobalState &from, GlobalState &to,
     ENFORCE(from.typeMembers.size() == to.typeMembers.size(), "Can't substitute symbols yet");
 
     from.sanityCheck();
-    {
-        UnfreezeFileTable unfreezeFiles(to);
-        int fileIdx = 0; // Skip file 0
-        while (fileIdx + 1 < from.filesUsed()) {
-            fileIdx++;
-            if (from.files[fileIdx]->sourceType == File::Type::NotYetRead) {
-                continue;
-            }
-            if (fileIdx < to.filesUsed() && from.files[fileIdx].get() == to.files[fileIdx].get()) {
-                continue;
-            }
-            ENFORCE(fileIdx >= to.filesUsed() || to.files[fileIdx]->sourceType == File::Type::NotYetRead);
-            to.enterNewFileAt(from.files[fileIdx], fileIdx);
-        }
-    }
+
+    GlobalSubstitution::mergeFileTables(from, to);
 
     fastPath = false;
     if (optionalCommonParent != nullptr) {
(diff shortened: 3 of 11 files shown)


core/GlobalState.cc
```diff
@@ -1806,6 +1806,27 @@ unique_ptr<GlobalState> GlobalState::deepCopy(bool keepId) const {
...
+    result->ensureCleanStrings = this->ensureCleanStrings;
+    result->runningUnderAutogen = this->runningUnderAutogen;
+    result->censorForSnapshotTests = this->censorForSnapshotTests;
+    result->sleepInSlowPath = this->sleepInSlowPath;
+    result->requiresAncestorEnabled = this->requiresAncestorEnabled;
+    result->kvstoreUuid = this->kvstoreUuid;
+    result->pathPrefix = this->pathPrefix;
+
+    return result;
+}
```
jez, line 1828: ooc do you think it's worth factoring this into a helper that's used by `deepCopy`? I almost did something similar for the `--minimize-rbi` change but ended up taking a different approach (no wrong answer)

core/GlobalState.h
```diff
@@ -235,6 +235,8 @@ class GlobalState final {
     std::unique_ptr<GlobalState> deepCopy(bool keepId = false) const;
     mutable std::shared_ptr<ErrorQueue> errorQueue;
 
+    std::unique_ptr<GlobalState> copyForIndex() const;
```
jez, line 238: doc comment to clarify when this specifically should be used over deepCopy?

core/GlobalSubstitution.cc
```diff
@@ -5,109 +5,49 @@
...
+                    auto &cnst = from.constantNames[i];
+                    ENFORCE_NO_TIMER(constantNameSubstitution.size() == i, "Constant name substitution has wrong size");
+                    // N.B.: cnst may reference a UniqueName, but since names are linearizeable we should have
+                    // already substituted it by now.
+                    constantNameSubstitution.emplace_back(to.enterNameConstant(substitute(cnst.original)));
                 }
-
-                uniqueNameSubstitution.emplace_back(
-                    to.freshNameUnique(nm.uniqueNameKind, substitute(nm.original), nm.num));
-                ENFORCE(!fastPath || uniqueNameSubstitution.back().uniqueIndex() == i);
```
jez, line 81: just checking: was it intentional to delete this `ENFORCE`?

core/GlobalSubstitution.h
```diff
@@ -17,9 +17,11 @@ class GlobalState;
  * The constructor builds up a lookup table from every NameRef in `from` to an equivalent NameRef in `to`, inserting new
  * names into `to` where needed. Then, that table can be used to rewrite multiple ASTs from `from`.
  */
-class GlobalSubstitution {
+class GlobalSubstitution final {
```
jez, line 20: 😎

core/GlobalSubstitution.h
```diff
@@ -52,18 +54,14 @@ class GlobalSubstitution {
...
     friend NameRefDebugCheck;
 
     std::vector<NameRef> utf8NameSubstitution;
     std::vector<NameRef> constantNameSubstitution;
     std::vector<NameRef> uniqueNameSubstitution;
-    // set if no substitution is actually necessary
-    bool fastPath;
 
-    const int toGlobalStateId;
+    int toGlobalStateId;
```
jez, line 64: i missed the motivation for this. was this intentional?

core/GlobalSubstitution.cc
```diff
@@ -118,8 +58,20 @@ GlobalSubstitution::GlobalSubstitution(const GlobalState &from, GlobalState &to,
     to.sanityCheck();
 }
 
-bool GlobalSubstitution::useFastPath() const {
-    return fastPath;
+void GlobalSubstitution::mergeFileTables(const GlobalState &from, GlobalState &to) {
+    UnfreezeFileTable unfreezeFiles(to);
+    int fileIdx = 0; // Skip file 0
+    while (fileIdx + 1 < from.filesUsed()) {
+        fileIdx++;
```
jez, line 65: this is completely a nit, but to prefer a `while` loop instead of a `for` loop?

```suggestion
    // id 0 is for non-existing FileRef
    for (int fileIdx = 1; fileIdx < from.filesUsed(); fileIdx++) {
```

core/GlobalSubstitution.cc
```diff
@@ -5,109 +5,49 @@
...
-            ENFORCE(from.fields[i].name == to.fields[i].name);
+            uniqueNameSubstitution.emplace_back(to.freshNameUnique(nm.uniqueNameKind, substitute(nm.original), nm.num));
         }
-        for (int i = 0; i < from.typeArguments.size(); ++i) {
-            ENFORCE(substitute(from.typeArguments[i].name) == from.typeArguments[i].name);
-            ENFORCE(from.typeArguments[i].name == to.typeArguments[i].name);
-        }
-        for (int i = 0; i < from.typeMembers.size(); ++i) {
-            ENFORCE(substitute(from.typeMembers[i].name) == from.typeMembers[i].name);
-            ENFORCE(from.typeMembers[i].name == to.typeMembers[i].name);
```
jez, line 110: it would be nice to figure out a way to make it harder-than-normal to enter symbols into the symbol table, but it's probably fine to not block on that.

main/pipeline/pipeline.cc
```diff
@@ -428,55 +428,107 @@ struct IndexThreadResultPack {
     IndexResult res;
 };
 
-IndexResult mergeIndexResults(const shared_ptr<core::GlobalState> cgs, const options::Options &opts,
-                              shared_ptr<BlockingBoundedQueue<IndexThreadResultPack>> input,
-                              const unique_ptr<const OwnedKeyValueStore> &kvstore) {
+struct IndexSubstitutionJob {
+    // Not necessary for substitution, but passing this through to the worker means it's freed in that thread, instead
+    // of serially in the main thread.
+    unique_ptr<core::GlobalState> threadGs;
```
jez, line 434: 👍

main/pipeline/pipeline.cc
```diff
@@ -428,55 +428,107 @@ struct IndexThreadResultPack {
...
     auto resultq = make_shared<BlockingBoundedQueue<IndexThreadResultPack>>(files.size());
     auto fileq = make_shared<ConcurrentBoundedQueue<core::FileRef>>(files.size());
     for (auto &file : files) {
         fileq->push(move(file), 1);
     }
 
-    workers.multiplexJob("indexSuppliedFiles", [baseGs, &opts, fileq, resultq, &kvstore]() {
-        Timer timeit(baseGs->tracer(), "indexSuppliedFilesWorker");
-        unique_ptr<core::GlobalState> localGs = baseGs->deepCopy();
+    std::shared_ptr<core::GlobalState> emptyGs = baseGs.copyForIndex();
```
jez, line 523: Should we mark `emptyGs` `const`? would that give us additional guarantees that no one tries to modify it in a thread?

jez (approved):

core/GlobalState.cc
jvilk-stripe: I think this should have comments indicating when to add to this list.

core/GlobalState.h
jvilk-stripe: seconding jez's comment here

hashing/hashing.cc
```diff
@@ -122,7 +122,7 @@ vector<ast::ParsedFile> Hashing::indexAndComputeFileHashes(unique_ptr<core::Glob
                                                            spdlog::logger &logger, vector<core::FileRef> &files,
                                                            WorkerPool &workers,
                                                            const unique_ptr<const OwnedKeyValueStore> &kvstore) {
-    auto asts = realmain::pipeline::index(gs, files, opts, workers, kvstore);
+    auto asts = realmain::pipeline::index(*gs, files, opts, workers, kvstore);
```
jvilk-stripe, line 125: yessssssss

main/pipeline/pipeline.cc
```diff
@@ -428,55 +428,107 @@ struct IndexThreadResultPack {
...
+    vector<ast::ParsedFile> trees;
+
+    IndexSubstitutionJob() {}
+
+    IndexSubstitutionJob(core::GlobalState &to, IndexResult res)
+        : threadGs{std::move(res.gs)}, subst{}, trees{std::move(res.trees)} {
+        if (absl::c_any_of(this->trees, [this](auto &parsed) { return !parsed.file.data(*this->threadGs).cached; })) {
+            this->subst.emplace(*this->threadGs, to);
+        } else {
+            core::GlobalSubstitution::mergeFileTables(*this->threadGs, to);
```
jvilk-stripe, line 446: I find it a little weird that we merge the file tables in two places. Should we unconditionally merge the file tables in this constructor?

main/pipeline/pipeline.cc
```diff
@@ -428,55 +428,107 @@ struct IndexThreadResultPack {
...
+        workers.multiplexJob("substituteTrees", [&cgs, batchq, resultq]() {
+            Timer timeit(cgs.tracer(), "substituteTreesWorker");
+            IndexSubstitutionJob job;
+            for (auto result = batchq->try_pop(job); !result.done(); result = batchq->try_pop(job)) {
+                if (result.gotItem()) {
+                    if (job.subst.has_value()) {
+                        for (auto &tree : job.trees) {
+                            auto file = tree.file;
+                            if (!file.data(cgs).cached) {
+                                core::MutableContext ctx(cgs, core::Symbols::root(), file);
```
jvilk-stripe, line 489: Why does `ast::Substitute` need a `MutableContext`? Is this a mistake?

main/pipeline/pipeline.cc
```diff
@@ -428,55 +428,107 @@ struct IndexThreadResultPack {
...
+                batchq->push(IndexSubstitutionJob{cgs, std::move(threadResult.res)}, numTrees);
+            }
+        }
+    }
+
+    {
+        Timer timeit(cgs.tracer(), "substituteTrees");
+        auto resultq = make_shared<BlockingBoundedQueue<vector<ast::ParsedFile>>>(batchq->bound);
+
+        workers.multiplexJob("substituteTrees", [&cgs, batchq, resultq]() {
```
jvilk-stripe, line 480: Can we explicitly pass a `const` reference to GlobalState to this lambda to prevent accidental mutation?

main/pipeline/pipeline.cc
jvilk-stripe: I'm +1 to making GlobalStates `const` when possible.

jvilk-stripe (approved):
I had some nits around mutability and agreed with @jez's comments. Otherwise LGTM.

core/GlobalState.cc
elliottt: I wondered about that. It felt like there was a lot of overlap, but starting from initEmpty() made me hesitate. Maybe we could factor out the common functionality into a local helper that they both use?

For the things that are copied over, I'm going to do another pass through what's in `deepCopy` to make sure that I didn't miss anything. I can add a comment indicating when it's good to add new entries.

core/GlobalSubstitution.cc
elliottt: Yes, `fastPath` is always `false` now, so it would never fire.

core/GlobalSubstitution.h
elliottt: Yes, keeping it `const` meant that the default move constructors didn't work. It's an unfortunate consequence, but all the methods on `GlobalSubstitution` are `const` so it didn't feel like the worst solution.

core/GlobalSubstitution.cc
elliottt: Happy to make this change -- I didn't write this loop :)

My preference would be to import `range-v3` to get `ranges::enumerate` and `ranges::drop`, but that's a bit of a heavyweight change for this PR.

main/pipeline/pipeline.cc
elliottt: I think that's a good change -- making it so that `GlobalSubstitution` only constructs a name substitution would be a good thing.

main/pipeline/pipeline.cc
elliottt: I'm not sure, I just moved over what was there already. I'd be happy to experiment with a `core::Context` instead.

main/pipeline/pipeline.cc
elliottt: Probably, I'll investigate.

elliottt: Since `Substitute` requires a `core::MutableContext`, this can't be made `const`. Maybe if we could do templates to `Substitute` to make that work.

main/pipeline/pipeline.cc
elliottt: `LazyGlobalSubstitution` requires a `core::MutableContext`, which is why `Substitute` is written assuming a `core::MutableSubstitution`.

elliottt:

→ https://go/builds/bui_KkqEOpMLuNv2Z3
→ https://go/builds/bui_KkqFQSnNef5BIz

elliottt:
Successful run on Stripe CI: https://go/builds/bui_KktSIhQWF616uh

PR #4993, opened 2021-12-09
enable -fsized-deallocation for release builds

froydnj:
### Motivation

Sized deallocation is/should be somewhat faster in jemalloc because it avoids some memory operations to lookup the size of the object being deallocated.  (It's probably *not* any faster by default if you're not using a custom memory allocator because the default implementation of sized operator delete just forwards to regular operator delete.)

That being said, experimenting with it on my devbox suggests that it's about 5-10% slower in aggregate, which is the opposite direction that I would expect.  Opening a PR to send everything through the official Stripe build system so I can make sure I'm not doing something dumb.

### Test plan

See included automated tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -86,6 +86,7 @@ build:release-common --compilation_mode=opt
 build:release-common --config=backtracesymbols
 build:release-common --config=static-libs
 build:release-common --config=versioned
+build:release-common --cxxopt=-fsized-deallocation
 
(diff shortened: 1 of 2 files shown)


froydnj:

→ https://go/builds/bui_KkKYxks0C0OhNh
→ https://go/builds/bui_KkKYh8kk6UYrpE

PR #4998, opened 2021-12-10
Allow for incremental packaging in noop change

ngroman:
Remove conditions that caused all incremental paths to always add _all_ package specs if any spec was present. This is no longer needed because these are stored in GlobalState.

Note: that it is still the case that if a package spec changes we go through the slow path.

### Motivation
Fix performance of simple LSP queries (go-to def) when in `__package.rb` files.

### Test plan

See included automated tests.

Manually tested VSCode goto-def in Stripe's codebase. Validated that it now reacts quickly.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -1306,25 +1306,6 @@ ast::ParsedFile wrapFileInPackageModule(core::Context ctx, ast::ParsedFile file,
     return file;
 }
 
-// We can't run packages without having all package ASTs. Assert that they are all present.
-bool checkContainsAllPackages(const core::GlobalState &gs, const vector<ast::ParsedFile> &files) {
-    UnorderedSet<core::FileRef> filePackages;
-    for (const auto &f : files) {
-        if (f.file.data(gs).sourceType == core::File::Type::Package) {
-            filePackages.insert(f.file);
-        }
-    }
-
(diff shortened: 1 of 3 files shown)


packager/packager.cc
```diff
@@ -1435,11 +1416,20 @@ ast::ParsedFile rewritePackagedFile(core::Context ctx, ast::ParsedFile parsedFil
...
+vector<ast::ParsedFile> rewriteFilesFast(core::GlobalState &gs, vector<ast::ParsedFile> files) {
+    Timer timeit(gs.tracer(), "packager.rewriteFilesFast");
     for (auto i = 0; i < files.size(); i++) {
         core::Context ctx(gs, core::Symbols::root(), files[i].file);
-        files[i] = rewritePackagedFile(ctx, move(files[i]));
+        if (files[i].file.data(gs).isPackage()) {
+            {
+                core::MutableContext ctx(gs, core::Symbols::root(), files[i].file);
+                // Re-write imports and exports:
+                getPackageInfo(ctx, files[i], gs.packageDB().extraPackageFilesDirectoryPrefixes());
```
jvilk-stripe, line 1427: I verified that, in `pipeline.cc`, we don't unfreeze the name table or symbol table for incremental packager so debug builds will enforce that this does not define any new names.

jvilk-stripe (approved):

PR #5000, opened 2021-12-10
wip: Make parser incremental

jez:
### Test plan

See included automated tests.

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -55,6 +55,7 @@ NameDef names[] = {
     {"defined_p", "defined?"},
     {"undef"},
     {"each"},
+    {"methodNameMissing", "<method-name-missing>"},
 
     // used in CFG for temporaries
     {"whileTemp", "<whileTemp>"},
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -120,8 +120,12 @@ public:
 
     core::LocOffsets collectionLoc(const token *begin, sorbet::parser::NodeVec &elts, const token *end) {
         if (begin != nullptr) {
-            ENFORCE(end != nullptr);
-            return tokLoc(begin, end);
+            if (end == nullptr) {
+                // implies that there was a parse error
+                return tokLoc(begin);
+            } else {
+                return tokLoc(begin, end);
+            }
         }
         ENFORCE(end == nullptr);
         if (elts.empty()) {
@@ -563,6 +567,20 @@ public:
         }
     }
 
+    unique_ptr<Node> call_method_error(unique_ptr<Node> receiver, const token *dot) {
+        auto loc = receiver->loc;
+        if (dot != nullptr) {
+            loc = loc.join(tokLoc(dot));
+        }
+
+        auto method = core::Names::methodNameMissing();
+        if ((dot != nullptr) && dot->string() == "&.") {
+            return make_unique<CSend>(loc, std::move(receiver), method, sorbet::parser::NodeVec{});
+        } else {
+            return make_unique<Send>(loc, std::move(receiver), method, sorbet::parser::NodeVec{});
+        }
+    }
+
     unique_ptr<Node> case_(const token *case_, unique_ptr<Node> expr, sorbet::parser::NodeVec whenBodies,
                            const token *elseTok, unique_ptr<Node> elseBody, const token *end) {
         return make_unique<Case>(tokLoc(case_).join(tokLoc(end)), std::move(expr), std::move(whenBodies),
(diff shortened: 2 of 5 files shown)


test/testdata/lsp/completion/if_empty.rb
```diff
@@ -0,0 +1,12 @@
+# typed: true
+
+# TODO(jez) Fix this test
```
jez, line 3: This is going to be fixed in a subsequent change. I'll make a ticket for this eventually.

test/testdata/parser/error_recovery/if_do_2.rb
```diff
@@ -0,0 +1,24 @@
...
+    # TODO(jez) Better error would be to point to the `do` keyword here
+    if x.y do
+      puts 'then'
+    else # error: else without rescue is useless
+      # TODO(jez) better parse would drop the `do` keyword and put this branch in the else
+      puts 'else'
+    end
+    Integer.class
+  end # error: Hint: closing "end" token was not indented as far as "if" token
+end # error: unexpected token "end of file"
```
jez, line 24: I'm going to address the `TODO`'s in this file in a followup PR.

parser/Parser.cc
```diff
@@ -90,28 +140,33 @@ unique_ptr<Node> Parser::run(sorbet::core::GlobalState &gs, core::FileRef file,
...
 
-    for (string local : initialLocals) {
-        driver->lex.declare(local);
+    auto driver = makeDriver(settings, buffer, scratch, initialLocals);
+    auto ast = builder.build(driver.get(), settings.traceParser);
+
+    // Always report the original parse errors
+    errorToError(gs, file, driver->diagnostics);
+
+    if (ast != nullptr) {
```
froydnj, line 150: Since this path uses `errorToError`, but the indentation-aware path uses `reportDiagnostics`, perhaps we should just inline `errorToError`'s code here to make the symmetry between the cases more explicit?  I think that would also have the benefit of making the `onlyHints` logic more explicit, so that the reader doesn't get concerns about errors being emitted twice, only to discover that `onlyHints` would prevent that.  (Still might be good to expand on the rationale for always reporting errors?)

parser/Parser.cc
```diff
@@ -16,72 +16,122 @@ extern const char *dclassStrings[];
...
                 }
-                break;
             }
-            default:
-                break;
+            break;
         }
+        case ruby_parser::dclass::DedentedEnd:
+            ENFORCE(diag.extra_location().has_value());
+            e.addErrorLine(rangeToLoc(gs, file, diag.extra_location().value()), "Matching token was here");
```
froydnj, line 68: To be clear: we don't have any tests that check we point at the matching token?

parser/parser/codegen/generate_diagnostics.cc
```diff
@@ -68,6 +68,9 @@ tuple<string, string> MESSAGES[] = {
     {"MissingCommaBetweenKwargs", "missing \\\",\\\" between keyword args"},
     {"CurlyBracesAroundBlockPass", "block pass should not be enclosed in curly braces"},
 
+    // Error recovery hints
+    {"DedentedEnd", "Hint: closing \\\"end\\\" token was not indented as far as {} token"},
```
froydnj, line 72: This phrasing strikes me as wrong: it's not that it's indented the wrong amount, it's that we associated it with a different construct than the user intended.

Compare [the similar situation in Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bda2211abfc8408a5f10d59a763cb592) or[ the error from Clang for the C/C++ case](https://gcc.godbolt.org/z/31aMe434Y), which I think do a better job of communicating what went wrong.

froydnj (changes requested):
This looks pretty reasonable, but there are enough comments/things to discuss that I think it's worth going through another round.

test/testdata/parser/error_recovery/def_missing_end_1.rb
```diff
@@ -0,0 +1,12 @@
...
+class A
+  def test1
+    if x.f
+    end
+
+  def test2
+    if x.f
+    end
+  end
+end # error: unexpected token "end of file"
```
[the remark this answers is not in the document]
jez: > My understanding is that something like this should be addressable within the framework we have here. Is that not the case, or does this test need a TODO?

Yep! I just wanted to break up the change. I will be fixing this in future changes.

> My question is motivated by my surprise that this test and `def_missing_end_2.rb` produce nothing in the class definition. But perhaps that's because the exp files for them are desugar tests, and they should be some kind of parse-tree exp tests?

I've been using desugar-tree tests simply because they're easier to skim to verify correctness. None of the parser formats are as easy to skim. The empty class definition you see is for `<root>` not `A`, which desugar wraps around all parse results, even empty ones.

parser/Parser.cc
jez: i'll add one

parser/parser/codegen/generate_diagnostics.cc
jez: Do you think we can iterate on this message? I generally agree with you but this PR is huge.

froydnj: I am down for iterating on this message, maybe we will even get complaints telling us what people would have preferred instead. :grin:

froydnj (approved):

parser/parser/codegen/generate_diagnostics.cc
jez: Thanks for the links—i went with something that looks more similar to the rust error messages.

PR #5001, opened 2021-12-11
Improve errors and suggestions for bad `import`s

aisamanra:
### Motivation
This ensures that we only suggest packages (and never non-package constants) for `import` lines.

### Test plan

See included automated tests. Added a CLI test which showcases the desired behavior.

### DIFF
--- a/test/cli/package-bad-imports/package-bad-imports.out
+++ b/test/cli/package-bad-imports/package-bad-imports.out
@@ -1,0 +1,39 @@
+example/__package.rb:5: Cannot find package `NonexistentPackage` https://srb.help/3704
+     5 |  import NonexistentPackage
+                 ^^^^^^^^^^^^^^^^^^
+
+example/__package.rb:7: Cannot find package `DoNot::SUggestThis` https://srb.help/3704
+     7 |  import DoNot::SUggestThis
+                 ^^^^^^^^^^^^^^^^^^
+
+example/__package.rb:9: Cannot find package `SUggestion` https://srb.help/3704
+     9 |  import SUggestion
+                 ^^^^^^^^^^
+
+example/__package.rb:11: Cannot find package `A::B::C` https://srb.help/3704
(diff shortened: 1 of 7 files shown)


test/cli/package-bad-imports/package-bad-imports.out
```diff
@@ -0,0 +1,39 @@
+example/__package.rb:5: Cannot find package `NonexistentPackage` https://srb.help/3704
+     5 |  import NonexistentPackage
+                 ^^^^^^^^^^^^^^^^^^
+
+example/__package.rb:7: Cannot find package `DoNot::SUggestThis` https://srb.help/3704
+     7 |  import DoNot::SUggestThis
+                 ^^^^^^^^^^^^^^^^^^
```
aisamanra, line 7: There's a constant in this test case called `DoNot::SuggestThis` which _isn't_ a package. The fact that we don't suggest it in this test is part of the point: we shouldn't suggest non-package constants.

test/cli/package-bad-imports/package-bad-imports.out
```diff
@@ -0,0 +1,39 @@
...
+example/__package.rb:9: Unable to resolve constant `SUggestion` https://srb.help/5002
+     9 |  import SUggestion
+                 ^^^^^^^^^^
+  Autocorrect: Use `-a` to autocorrect
+    example/__package.rb:9: Replace with `Suggestion`
+     9 |  import SUggestion
+                 ^^^^^^^^^^
+    suggestion/__package.rb:3: Did you mean: `Suggestion`?
+     3 |class Suggestion < PackageSpec
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
aisamanra, line 34: We _do_ correctly suggest this package.

aprocter-stripe (approved):
Looks reasonable to me!

PR #5002, opened 2021-12-12
Add more benchmarks for `T.let`

jeffcarbs:
### Motivation

Initializing non-simple types is expensive which means using any inline types (e.g. `T.let/cast(...)`) even when `checked: false` incurs performance penalties. This adds some microbenchmarks to demonstrate this to give a starting point for any improvements to the performance here.

| Code | Time | |
| -- | -- | -- |
| T.let(..., Integer) | 366.56 ns | baseline |
| T.let(..., T.nilable(Integer)) | 1.078 μs | ~3x slower |
| T.let(..., T.nilable(T.any(Integer, String))) | 6.89 μs | ~18x slower |
| T.let(..., T.nilable(T.any(Integer, String)), checked: false) | 5.938 μs | ~16x slower |

### Test plan

No tests, just updating benchmarking code. Confirmed the code ran locally to produce the results.

### DIFF
--- a/gems/sorbet-runtime/bench/typecheck.rb
+++ b/gems/sorbet-runtime/bench/typecheck.rb
@@ -105,6 +105,21 @@ module SorbetBenchmarks
         nilable_integer_param(1)
       end
 
+      time_block("T.let(..., T.nilable(T.any(Integer, String)))") do
+        T.let(nil, T.nilable(T.any(Integer, String)))
+        T.let(1, T.nilable(T.any(Integer, String)))
+      end
+
+      time_block("T.let(..., T.nilable(T.any(Integer, String)), checked: false)") do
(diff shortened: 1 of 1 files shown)


froydnj (changes requested):
The substance of this change looks fine, but it looks like rubocop is angry about the additional lines you added.  I think it's reasonable to turn off the relevant rubocop metrics checks for this file, wdyt?

jeffcarbs:
> I think it's reasonable to turn off the relevant rubocop metrics checks for this file, wdyt?

I agree. I pushed up a commit to disable the rubocops inline so this should be good to go.

jez:
https://github.com/sorbet/sorbet/pull/4568

It's worth pointing out that someone at Stripe has made a similar change, but I just havent gotten around to reviewing it. We might want to take a look at that one now too.

PR #5005, opened 2021-12-14
Define `TypeParameter` and `ClassOrModule` classes; delete `Symbol` class.

jvilk-stripe:
Define `TypeParameter` and `ClassOrModule` classes; delete `Symbol` class.

### Motivation

Remove the monolithic `Symbol` class and instead have smaller symbol-type-specific classes with fewer ENFORCEs/weird corner cases/etc.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.h
+++ b/core/Symbols.h
@@ -304,40 +304,34 @@ public:
 
     class Flags {
     public:
-        static constexpr uint32_t NONE = 0;
-
-        // We're packing three different kinds of flags into separate ranges with the uint32_t's below:
-        //
-        // 0x0000'0000
-        //   ├▶    ◀┤└─ Applies to all types of symbol
-        //   │      │
-        //   │      └─ For our current symbol type, what flags does it have?
-        //   │         (New flags grow up towards MSB)
-        //   │
-        //   └─ What type of symbol is this?
-        //      (New flags grow down towards LSB)
-        //
-
-        // --- What type of symbol is this? ---
-        static constexpr uint32_t CLASS_OR_MODULE = 0x8000'0000;
-
-        // --- Applies to all types of Symbols ---
-
         // Synthesized by C++ code in a Rewriter pass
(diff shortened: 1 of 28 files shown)


core/Symbols.h
```diff
@@ -217,56 +217,117 @@ class Field final {
...
+
+    Flags flags;
+    // Method for TypeArgument, ClassOrModule for TypeMember.
+    SymbolRef owner;
+    NameRef name;
+    TypePtr resultType;
+
+private:
+    InlinedVector<Loc, 2> locs_;
+};
```
jvilk-stripe, line 297: Before merging: Put in a `CheckSize` annotation.

core/Symbols.h
```diff
@@ -217,56 +217,117 @@ class Field final {
...
+
+class ClassOrModule final {
```
jvilk-stripe, line 299: Historically `Symbol` couldn't have `CheckSize` because `UnorderedMap` has different sizes in different environments. I dunno if that is still true.

froydnj: The assertions from `CheckSize` are only active for `!debug_mode && !emscripten_build`...so I guess there'd have to be significant differences between x86-64 Linux and Mac (which doesn't seem likely)  to cause problems?  Maybe there would be latent issues for aarch64?

elliottt (approved):
This looks great!

froydnj:
I don't understand why some of the tests around ambiguous definitions are failing.  My suspicion is that they are somehow related to the `isUndeclared` flag that got added prior to this PR being written originally, but I don't see where things would have gone wrong in the rebasing.  (Unless `isRewriterSynthesized` was also secretly being used for something related to classes/modules and it got unexpectedly nuked?)

elliottt (approved):
This looks great!

froydnj:

→ https://go/builds/bui_LUhyvVUv1zsJmu
→ https://go/builds/bui_LUhykf0KBMgmn5

PR #5011, opened 2021-12-15
Find ambiguous definitions in resolver

aadi-stripe:
Adds resolver functionality to find and report "ambiguous definitions".

### Motivation

In the Stripe codebase, we can face issues at runtime due to [this](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Amodule%20Opus%0A%20%20%23%20module%20Opus%3A%3AFoo%3A%3AFoo%3B%20end%20%3C--%20filler%20node%0A%0A%20%20module%20Foo%0A%20%20%20%20module%20Foo%3A%3ABar%0A%20%20%20%20end%0A%20%20end%0Aend%0A%0AOpus%3A%3AFoo%3A%3AFoo%3A%3ABar%20%23%20%3C-%20Sorbet%0AOpus%3A%3AFoo%3A%3ABar%20%23%20%3C-%20Ruby%20VM) kind of code, where a symbol definition is scoped differently in Sorbet vs the Ruby VM. Usually, the autoloader we use at Stripe creates "filler" nodes which forward declare namespaces. However, the autoloader doesn't cover all cases. It only works at file granularity (and not for inner defs within a file itself).

### Test plan

Runtime
--------
Before
trace_old_sp

After
trace_new_sp

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -353,7 +353,7 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
         cxxopts::value<vector<string>>(), "string");
     options.add_options("advanced")("no-error-count", "Do not print the error count summary line");
     options.add_options("advanced")("autogen-version", "Autogen version to output", cxxopts::value<int>());
-    options.add_options("advanced")("report-ambiguous-defs", "Enable ambiguous definition error enforcement",
+    options.add_options("advanced")("report-ambiguous-definitions", "Enable ambiguous definition error enforcement",
                                     cxxopts::value<bool>());
     options.add_options("advanced")("stripe-mode", "Enable Stripe specific error enforcement", cxxopts::value<bool>());
     options.add_options("advanced")("stripe-packages", "Enable support for Stripe's internal Ruby package system",
@@ -876,7 +876,7 @@ void readOptions(Options &opts,
             }
             opts.autogenVersion = raw["autogen-version"].as<int>();
         }
-        opts.reportAmbiguousDefinitionErrors = raw["report-ambiguous-defs"].as<bool>();
+        opts.reportAmbiguousDefinitionErrors = raw["report-ambiguous-definitions"].as<bool>();
         opts.stripeMode = raw["stripe-mode"].as<bool>();
         opts.stripePackages = raw["stripe-packages"].as<bool>();
         if (raw.count("extra-package-files-directory-prefix")) {
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -353,6 +353,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
         cxxopts::value<vector<string>>(), "string");
     options.add_options("advanced")("no-error-count", "Do not print the error count summary line");
     options.add_options("advanced")("autogen-version", "Autogen version to output", cxxopts::value<int>());
+    options.add_options("advanced")("report-ambiguous-defs", "Enable ambiguous definition error enforcement",
+                                    cxxopts::value<bool>());
     options.add_options("advanced")("stripe-mode", "Enable Stripe specific error enforcement", cxxopts::value<bool>());
     options.add_options("advanced")("stripe-packages", "Enable support for Stripe's internal Ruby package system",
                                     cxxopts::value<bool>());
@@ -874,6 +876,7 @@ void readOptions(Options &opts,
             }
             opts.autogenVersion = raw["autogen-version"].as<int>();
         }
+        opts.reportAmbiguousDefinitionErrors = raw["report-ambiguous-defs"].as<bool>();
         opts.stripeMode = raw["stripe-mode"].as<bool>();
         opts.stripePackages = raw["stripe-packages"].as<bool>();
         if (raw.count("extra-package-files-directory-prefix")) {
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -948,8 +948,114 @@ public:
             transformAncestor(ctx.withOwner(klass), singleton, ancst, isInclude);
         }
 
+        // Check for ambiguous definitions.
+        if (checkAmbiguousDefinition(ctx)) {
+            const auto ambigDef = findAnyDefinitionAmbiguousWithCurrent(ctx);
+            if (ambigDef.exists()) {
+                if (auto e = ctx.beginError(original.loc, core::errors::Resolver::AmbiguousDefinitionError)) {
+                    e.setHeader("Class definition is ambiguous");
+                    e.addErrorLine(ambigDef.loc(ctx), "Alternate definition {} here",
+                                   ambigDef.showFullNameWithoutPackagePrefix(ctx));
+                }
+            }
+        }
+
         nesting_ = nesting_->parent;
         return tree;
+    }
+
+    const bool checkAmbiguousDefinition(core::Context ctx) {
+        if (ctx.state.runningUnderAutogen) {
+            // no need to check in autogen
+            return false;
+        }
+
+        if (ctx.file.data(ctx).isPackage()) {
+            // no need to check package files
+            return false;
+        }
+
+        if (nesting_->scope == core::Symbols::root()) {
+            // no need to check <root> def itself
+            return false;
+        }
+
+        // If current definition is owned by package registry/test, this means it is a package-mangled module.
+        const core::SymbolRef curOwner = nesting_->scope.owner(ctx);
+        if (curOwner == core::Symbols::PackageTests() || curOwner == core::Symbols::PackageRegistry()) {
+            return false;
+        }
+
+        // If current parent is owned by package registry/test, this means we are at package-top-level (and we skip).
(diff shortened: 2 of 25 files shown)


resolver/resolver.cc
```diff
@@ -948,10 +948,75 @@ class ResolveConstantsWalk {
             transformAncestor(ctx.withOwner(klass), singleton, ancst, isInclude);
         }
 
+        // TODO: Only do this check in --stripe-mode. Currently it does the check but skips reporting the error if
+        // not in --stripe-mode.
```
froydnj, line 952: Nit: the mentions of `--stripe-mode` in this comment are supposed to be `--stripe-packages`?

resolver/resolver.cc
```diff
@@ -948,10 +948,75 @@ class ResolveConstantsWalk {
...
+        }
+
+        if (!curSym.exists() || (curSym.exists() && curSym == core::Symbols::root())) {
+            // preceding symbol cannot be ambiguous in these cases
+            return std::nullopt;
+        }
+        auto precedingSymForCurDef = lastSym;
+
+        if (precedingSymForCurDef.isClassOrModule() &&
+            !precedingSymForCurDef.asClassOrModuleRef().data(ctx)->isClassModuleSet()) {
```
froydnj, line 996: I really like how you wrote the first half of this function with early exits to reduce the nesting depth.  WDYT about turning this into an early exit so we can reduce the nesting depth of the latter half of the function?

resolver/resolver.cc
```diff
@@ -948,10 +948,75 @@ class ResolveConstantsWalk {
...
+                    e.addErrorLine((*ambigDef).loc(ctx), "Previously defined here.");
+                }
+            }
+        }
+
         nesting_ = nesting_->parent;
         return tree;
     }
 
+    const std::optional<core::SymbolRef> findAnyDefinitionAmbiguousWithCurrent(core::Context ctx) {
```
froydnj, line 969: Is there a reason we can't return just `core::SymbolRef` here, with a default-constructed (i.e. `!exists()`) `SymbolRef` to indicate non-ambiguity?

froydnj (commented):
I only skimmed through this and didn't think very hard about the perf implications.

Probably goes without saying, but some tests would be good?

aadi-stripe:
> I only skimmed through this and didn't think very hard about the perf implications.
> 
> Probably goes without saying, but some tests would be good?

Yep, test coming up.

resolver/resolver.cc
aadi-stripe: No, this is different from `--stripe-packages`. We actually want to run it in `--stripe-mode` regardless of `--stripe-packages`.

resolver/resolver.cc
aadi-stripe: Sure!

aadi-stripe:
> > I only skimmed through this and didn't think very hard about the perf implications.
> > Probably goes without saying, but some tests would be good?
> 
> Yep, test coming up.

Added

ngroman:
Curious if people think this check should be on by default or not? On one hand, it could introduce backwards incompatibilities, but on the other this is a situation that can cause real world errors.

resolver/resolver.cc
```diff
@@ -948,10 +948,75 @@ class ResolveConstantsWalk {
...
+        // TODO: Only do this check in --stripe-mode. Currently it does the check but skips reporting the error if
+        // not in --stripe-mode.
+        //
+        // TODO: Do rigorous perf analysis.
+        if (!ctx.file.data(ctx).isPackage()) {
+            const auto ambigDef = findAnyDefinitionAmbiguousWithCurrent(ctx);
+            if (ambigDef.has_value()) {
+                if (auto e = ctx.beginError(original.loc, core::errors::Resolver::AmbiguousDefinitionError)) {
+                    e.setHeader("Class definition is ambiguous", klass.show(ctx));
+                    e.addErrorLine((*ambigDef).loc(ctx), "Previously defined here.");
```
ngroman, line 960: "Previously defined here" isn't exactly the right message in this case.

ngroman:
It's worth adding an additional test that ensures this check still function in `--stripe-packages` mode (and testing that on Stripe's codebase). I could see there being some tricky interactions between the aliasing it uses and this.

resolver/resolver.cc
```diff
@@ -948,10 +948,75 @@ class ResolveConstantsWalk {
...
     }
 
+    const std::optional<core::SymbolRef> findAnyDefinitionAmbiguousWithCurrent(core::Context ctx) {
+        if (nesting_ == nullptr) {
+            // can't be ambiguous if nesting is null
+            return std::nullopt;
+        }
+
+        auto curSym = nesting_->scope;
+        auto curNesting = nesting_->parent;
```
ngroman, line 976:
```suggestion
        auto &curNesting = nesting_->parent;
```
Avoid hitting the shared_ptr's counter

aadi-stripe:
> It's worth adding an additional test that ensures this check still function in `--stripe-packages` mode (and testing that on Stripe's codebase). I could see there being some tricky interactions between the aliasing it uses and this.

Good call, I also tested it with `--stripe-packages`.

resolver/resolver.cc
aadi-stripe: Can't do this as I'm reassigning it in the loop bellow.

resolver/resolver.cc
jez: ooc: does it have to be in `--stripe-mode`? i have a feeling this might be something that open sources users might want to turn on more widely. what parts of stripe mode does it actually depend on?

- if it's actually orthogonal and we just want to enable it for stripe's codebase but not necessarily all users' codebases, can we add a separate flag?
- if it depends on some nuanced assumptions that stripe mode sets up, can we attempt to ENFORCE that those hold here and/or document what those assumptions are in a comment?

aadi-stripe: @jez My feeling is that in more predictably built codebases, where filler nodes are created appropriately (as opposed to Stripe where there are some edge cases with our autoloader), this check would not be useful to run?

I could be wrong.

aadi-stripe:

→ https://go/builds/bui_KmocOosqEop5oZ
→ https://go/builds/bui_KmoceNGH5uWpVD

aadi-stripe:
PR is ready for review. Will not merge until ensuing errors are fixed in Stripe codebase.

ptal:
@froydnj 
@jez 
@ngroman 
@jvilk

resolver/resolver.cc
aadi-stripe: Update -- I decided to make this available even outside of `--stripe-mode`.

aadi-stripe:
PR is ready.
ptal:
@jvilk 
@ngroman 
@jez 
@froydnj

aadi-stripe:
Btw, I am enabling this check regardless of `--stripe-mode`.

aadi-stripe:
> Btw, I am enabling this check regardless of `--stripe-mode`.

Update: Instead of overloading `--stripe-mode` (which actually is a bit harder to make work mechanically), I am going to move it behind an orthogonal flag. 

Back to WIP.

resolver/resolver.cc
aadi-stripe: Added a separate flag.

aadi-stripe:
> > Btw, I am enabling this check regardless of `--stripe-mode`.
> 
> Update: On second thought, it might be better to hide this behind a flag. Instead of overloading `--stripe-mode` (which actually is a bit harder to make work mechanically), I am going to move it behind an orthogonal flag.
> 
> Back to WIP.

Added a new flag

main/options/options.cc
```diff
@@ -353,6 +353,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
         cxxopts::value<vector<string>>(), "string");
     options.add_options("advanced")("no-error-count", "Do not print the error count summary line");
     options.add_options("advanced")("autogen-version", "Autogen version to output", cxxopts::value<int>());
+    options.add_options("advanced")("report-ambiguous-defs", "Enable ambiguous definition error enforcement",
```
jez, line 356: nit: what do you think about lengthening `defs` to `definitions` or maybe `constant-definitions`?

e.g., I usually associate "defs" with the `def` keyword for methods in Ruby, but iiuc this feature is entirely related to constants.

website/docs/error-reference.md
```diff
@@ -872,6 +872,56 @@ class MyClass < MySuperClass
...
+regardless of whether they are explicitly predeclared like in the second
+example. This means that in Sorbet's view, the definition resolves to A::B::C in
+either case.
+
+In Stripe's codebase, this is generally not a problem at runtime, as we use
+Sorbet's own autoloader generation to pre-declare filler namespaces, keeping the
+Ruby runtime's behavior equivalent to Sorbet. However, the autoloader has some
+edge cases, which can often cause deviations between Ruby's runtime and Sorbet.
+This error helps guard against these issues.
+
```
jez, line 924: Great docs, thanks!!

jez (approved):

PR #5012, opened 2021-12-15
add rbi definitions for OpenSSL::KDF

simoleone:
Adds the `OpenSSL::KDF` module to stdlib types.

One question: it looks like quite a bit of documentation is copied over into `.rbi`s, should I do so too?
See: https://ruby-doc.org/stdlib-2.7.4/libdoc/openssl/rdoc/OpenSSL/KDF.html

### Motivation

`OpenSSL::KDF` is missing from definitions, so add it.

### Test plan

Looked at similar PRs and didn't seem like there was a way to test these easily. Let me know if there's something to do, though!

### DIFF
--- a/rbi/stdlib/openssl.rbi
+++ b/rbi/stdlib/openssl.rbi
@@ -3686,6 +3686,168 @@ end
 class OpenSSL::Digest::DigestError < OpenSSL::OpenSSLError
 end
 
+# Provides functionality of various KDFs (key derivation function).
+#
+# KDF is typically used for securely deriving arbitrary length symmetric
+# keys to be used with an OpenSSL::Cipher from passwords. Another use case
+# is for storing passwords: Due to the ability to tweak the effort of
+# computation by increasing the iteration count, computation can be slowed
+# down artificially in order to render possible attacks infeasible.
+#
+# Currently, OpenSSL::KDF provides implementations for the following KDF:
+#
+# * PKCS #5 PBKDF2 (Password-Based Key Derivation Function 2) in
+#   combination with HMAC
+# * scrypt
+# * HKDF
+#
+# == Examples
+# === Generating a 128 bit key for a Cipher (e.g. AES)
+#   pass = "secret"
+#   salt = OpenSSL::Random.random_bytes(16)
+#   iter = 20_000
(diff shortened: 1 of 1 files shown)


rbi/stdlib/openssl.rbi
```diff
@@ -3686,6 +3686,48 @@ end
...
+      pass: String,
+      salt: String,
+      N: Integer,
+      r: Integer,
+      p: Integer,
+      length: Integer
+    )
+    .returns(String)
+  end
+  def self.scrypt(pass, salt:, N:, r:, p:, length:); end
```
elliottt, line 3713: This will fix the build error.

```suggestion
      n: Integer,
      r: Integer,
      p: Integer,
      length: Integer
    )
    .returns(String)
  end
  def self.scrypt(pass, salt:, n:, r:, p:, length:); end
```

elliottt: Given that this won't work, and that the method is defined in C, what do you think about removing this sig from the rbi?

simoleone: double-splat it is!

elliottt:

→ https://go/builds/bui_Kn0A8rUv9DFq3q
→ https://go/builds/bui_Kn0AhBLqarVH6a

elliottt:
> One question: it looks like quite a bit of documentation is copied over into .rbis, should I do so too?

If you'd like to add the docs as well, that would be very helpful. Thanks!

simoleone:
Updated with docs!

MichaelCordingley:
Would love to see this merged. I'm getting failing type checks because it's missing.

froydnj:
Sorry for letting this fall through the cracks; it looks like the Stripe-internal builds last time just need a small amount of fixing up.

@simoleone, would you be willing to rebase this PR on master?

simoleone:
@froydnj No problem. Rebased!

froydnj:

→ https://go/builds/bui_MTNI3cLIPQuxXc
→ https://go/builds/bui_MTNItTqocIz6Hf

froydnj:
Thank you @simoleone !

froydnj (approved):

PR #5015, opened 2021-12-15
fix loc information for desugared `===` calls in `when`

### LINKED ISSUE #4073: Hovering over `when` displays `===` in IDE
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Amodule%20A%0A%20%20class%20B%0A%20%20end%0Aend%0A%0Asig%20%7Bparams(x%3A%20T.untyped).void%7D%0Adef%20foo(x)%0A%20%20case%20x.class%0A%20%20when%20Integer%0A%20%20%20%201%0A%20%20when%20A%3A%3AB%0A%20%20%20%202%0A%20%20end%0Aend%0A)

```ruby
# typed: true
extend T::Sig

module A
  class B
  end
end

def foo(x)
  case x.class
  when Integer
    1
  when A::B
    2
  end
end
```

#### Observed output

Hover over `Integer` and you will see it shows `def ===(other); end` as description.  In case of namespaced classes, the last component shows this behavior. On the example, hovering over `A` would correctly point to `module A` but hovering over `B` shows `def ===(other); end` as description.
In VSCode this means navigation is broken, going to `def ===(other); end` in `module.rbi`.

#### Expected behavior

Navigate to correct type (`Integer` or `class B` on example).

- - -

image

froydnj:
### Motivation

Fixes #4073.

It's not immediately clear to me *why* this fixes things, but the test seems to work?

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1980,9 +1980,8 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                             auto ctree = node2TreeImpl(dctx, std::move(cnode));
                             if (temp.exists()) {
                                 auto local = MK::Local(cloc, temp);
-                                auto patternloc = ctree.loc();
                                 test =
-                                    MK::Send1(patternloc, std::move(ctree), core::Names::tripleEq(), std::move(local));
+                                    MK::Send1(when->loc, std::move(ctree), core::Names::tripleEq(), std::move(local));
                             } else {
                                 test = std::move(ctree);
                             }
--- a/test/testdata/lsp/when_eqeqeq.rb
+++ b/test/testdata/lsp/when_eqeqeq.rb
@@ -1,0 +1,20 @@
+# typed: true
+extend T::Sig
+
+module A
+  class B
+  end
+end
+
+sig {params(x: T.untyped).void}
+def foo(x)
+  case x.class
+  when Integer
+     # ^ hover: T.class_of(Integer)
+    1
+  when A::B
+     # ^ hover: T.class_of(A)
+        # ^ hover: T.class_of(A::B)
+    2
+  end
+end


froydnj:
Ehhh, changing loc information is going to be a little more complicated than I thought, given the test failures.

froydnj (commented):
I think this issue is fixable, but I also think we'd have to scatter some hacks throughout infer to get the behaviors that we want.  We could switch desugar to use `<Magic>.<caseTripleEq>(x, y)` and then checking sends for calling `<caseTripleEq>` would make it more obvious what's going on.  But we'd also have to remember to try and write testcases that would exercise `when X` for new features and insert more hacks in appropriate places.

The alternative to one-off hacks in various places is that we add a *third* `LocOffsets` to `ast::Send`, so we'd have (names only for example purposes!):

* `funLoc`
* `locTripleEqReceiverLoc`
* `locForEntireSend`

We really do have two different locs for the entire send to care about: in some cases we want the send to look like it has zero extent, but in other cases we want it to have effectively the same extent as the receiver.

But then we'd have to be very careful about which one gets used when, and I'm not even sure that we could consistently pick one in some cases (e.g. when constructing `cfg::Send`, I think we want to use the zero-length loc for the CFG node, but somehow for the uses of the result, we want to use the non-zero-length loc?).

(It would be a nice thing that even though `ast::Send` would get bigger, it would still fit into the same jemalloc bucket size, so we shouldn't see much size increase...though we would probably have to propagate this into `cfg::Send`, which might affect memory consumption.)

froydnj:
Oh, actually, another alternative: introduce another AST node specifically for representing `when X`, so we could track all the information we needed there.  I don't know whether we would need an equivalent CFG node...probably, though typing it could be delegated to typing on send'ing `===`, similarly to what we do for `YieldLoadArg`.

I think this would have negative effects on infer, because now control-flow sensitivity isn't expressed entirely as certain functions on `Send`s, but also on the results of this special AST/CFG node.  That's certainly doable, but it would make things more complex.

test/testdata/lsp/completion/case_1.rb
```diff
@@ -10,7 +10,7 @@ def self.test1(x)
     puts 'before'
     case x # error: Hint: this "case" token might not be properly closed
     when MethodCompletion.
-    #                     ^ completion: class, ...
+    #                     ^ completion: x, test1, test2, ...
```
jez, line 13: can you delete the `TODO` comment a few lines up?

test/testdata/lsp/completion/case_2.rb
```diff
@@ -13,7 +13,7 @@ def self.test1(x)
     puts 'before'
     case x # error: Hint: this "case" token might not be properly closed
     when ConstantCompletion:: # error: expected constant name following "::"
-    #                        ^ completion: (nothing)
+    #                        ^ completion: A, B
```
jez, line 16: can you delete the TODO comment here as well?

jez (commented):
👌 thanks!

jez:
Fixes #5441

jez (approved):

PR #5016, opened 2021-12-15
show documentation for methods with multi-line final sigs

### LINKED ISSUE #2306: IDE: Documentation on hover doesn't show docs for multiline final sigs
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20%23%20These%20docs%20don't%20show!%0A%20%20sig(%3Afinal)%20do%0A%20%20%20%20params(x%3A%20Integer).returns(String)%0A%20%20end%0A%20%20def%20bar(x)%0A%20%20%20%20%23%20%5E%20hover%20here%0A%20%20%20%20x.to_s%0A%20%20end%0Aend%0A)

```ruby
# typed: true
class A
  extend T::Sig

  # These docs don't show!
  sig(:final) do
    params(x: Integer).returns(String)
  end
  def bar(x)
    # ^ hover here
    x.to_s
  end
end
```

#### Observed output

Hover does not show the documentation.

#### Expected behavior

Hover shows the documentation

This should be an easy fix by changing the docstring parsing logic to recognize :final syntax.

froydnj:
### Motivation

Fixes #2306.

I did not think it was worth trying to factor out a helper to recognize `sig do` and permutations, because the `sig do;` case would make it more complicated.  But the reviewer may feel differently.

### Test plan

See included automated tests.  The tests might be redundant, but it was educational writing both of them.

### DIFF
--- a/main/lsp/lsp_helpers.cc
+++ b/main/lsp/lsp_helpers.cc
@@ -375,13 +375,14 @@ optional<string> findDocumentation(string_view sourceCode, int beginIndex) {
 
         // Handle multi-line sig block
         else if (absl::StartsWith(line, "end")) {
-            // ASSUMPTION: We either hit the start of file, a `sig do` or an `end`
+            // ASSUMPTION: We either hit the start of file, a `sig do`/`sig(:final) do` or an `end`
             it++;
             while (
                 // SOF
                 it != all_lines.rend()
                 // Start of sig block
-                && !absl::StartsWith(absl::StripAsciiWhitespace(*it), "sig do")
+                && !(absl::StartsWith(absl::StripAsciiWhitespace(*it), "sig do") ||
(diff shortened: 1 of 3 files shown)


jez (approved):
Great, thanks!

For future reference, I almost never write tests in C++ (lsp_test.cc) unless there's no other way to write the test in a testdata file. Feel free to do the same.

jez:
> I did not think it was worth trying to factor out a helper to recognize `sig do` and permutations, because the `sig do;` case would make it more complicated. But the reviewer may feel differently.

I think one day we might want to have some phase of Sorbet pre-emptively parse comments and associate them with the closest following AST node, so that we can query them in e.g. namer+resolver when entering things into the symbol table.

PR #5017, opened 2021-12-15
Update Docusaurus config with new Algolia settings

aprocter-stripe:
### Motivation
Migrating website to new DocSearch infra.

### Test plan
Spun up `yarn start` on my laptop with the new settings. Verified that search works with these new settings, and doesn't if I replace them with garbage.


froydnj (approved):
Github really needs :stamp:

PR #5020, opened 2021-12-16
Fix return type for String#sub!

Morriar:
### Motivation

As documented [here](https://ruby-doc.org/core-2.7.0/String.html#method-i-sub-21), `gsub` returns `str` if a substitution was performed or `nil` if no substitution was performed:

```
$ irb
> "".sub!("A", "B")
=> nil
```

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/string.rbi
+++ b/rbi/core/string.rbi
@@ -2104,14 +2104,14 @@ class String < Object
         arg0: T.any(Regexp, String),
         arg1: String,
     )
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_KmzsvVDp5rdlVG
→ https://go/builds/bui_KmzsG6DZKLrgYv

jez (approved):

PR #5022, opened 2021-12-16
Make sorbet/showSymbol work in `typed: false` files

jez:
### Motivation

This matches the behavior of Go To Definition.

Basically everything works except for handling `SendResponse` (and honestly, the
most common use case for this is to show a fully-qualified constant symbol
name).

Making it not return an RPC error in `typed: false` files is a better user
experience.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/sorbet_show_symbol.cc
+++ b/main/lsp/requests/sorbet_show_symbol.cc
@@ -15,7 +15,7 @@ unique_ptr<ResponseMessage> SorbetShowSymbolTask::runRequest(LSPTypecheckerDeleg
 
     const core::GlobalState &gs = typechecker.state();
     // To match the behavior of Go To Definition, we don't error in an untyped file, but instead
-    // return an empty result.
+    // be okay with returning an empty result for certain queries.
     auto errorIfFileIsUntyped = false;
     auto result = queryByLoc(typechecker, params->textDocument->uri, *params->position, LSPMethod::SorbetShowSymbol,
                              errorIfFileIsUntyped);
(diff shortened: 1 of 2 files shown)


main/lsp/requests/sorbet_show_symbol.cc
```diff
@@ -14,7 +14,11 @@ unique_ptr<ResponseMessage> SorbetShowSymbolTask::runRequest(LSPTypecheckerDeleg
     auto response = make_unique<ResponseMessage>("2.0", id, LSPMethod::SorbetShowSymbol);
 
     const core::GlobalState &gs = typechecker.state();
-    auto result = queryByLoc(typechecker, params->textDocument->uri, *params->position, LSPMethod::SorbetShowSymbol);
+    // To match the behavior of Go To Definition, we don't error in an untyped file, but instead
+    // return an empty result.
+    auto errorIfFileIsUntyped = false;
+    auto result = queryByLoc(typechecker, params->textDocument->uri, *params->position, LSPMethod::SorbetShowSymbol,
+                             errorIfFileIsUntyped);
```
froydnj, line 21: This comment is good, but it's not obvious to me exactly how the empty result is returned in untyped files (compare `DefinitionTask::runRequest`, which I guess is the implementation of Go To Definition: explicit checks for `fileIsTyped`).

Does this just happen magically below because `sym` won't exist for non-constant things?  If that's the case, maybe we should be a little more explicit in this comment that we can return non-empty results for some things, and consider adding a note below?

froydnj (approved):
I can believe this works because there is a test, but I would like some confirmation on questions below.

main/lsp/requests/sorbet_show_symbol.cc
jez: Yeah it's maybe more straightforward to say "..., but instead be okay with returning an empty result"

constants and method defs will have a result, method calls will not.

PR #5024, opened 2021-12-17
Remove all `T.untyped` from FileUtils.rbi

sushain97:
Removes all `T#untyped` from the RBIs from `FileUtils`, fixing up a bunch of arguments that were `list` but didn't take a `list` along the way.

### Motivation
More types.

### Test plan
Relying on existing tests.


sushain97:

→ https://go/builds/bui_KnMsNaQDmpTX7y
→ https://go/builds/bui_KnMsnLqI0ioTkj

elliottt (approved):
This all seems reasonable to me, thanks for removing so much `T.untyped`!

PR #5031, opened 2021-12-22
Fix kwarg deprecation in Ruby >= 2.7

ngan:
This fixes a kwarg deprecation warning for Ruby >= 2.7.

### Motivation

Trying to get rid of deprecation warnings.

### Test plan

See included automated tests.


froydnj (changes requested):
This is not correct as written, as you can see from the test failures with the Sorbet compiler (which also exercises sorbet-runtime as part of its test suite).  I suspect you would see similar test failures if you were to run sorbet-runtime's test suite (see `gems/sorbet-runtime/README.md` for instructions).

It would also be helpful if you posted error details along with what version of Ruby you were seeing the errors with.

ngan:
Actually, this was a Problem on our end.

PR #5032, opened 2021-12-23
Require stdlib's Set

ngan:
```
NameError:
  uninitialized constant T::Props::Decorator::Set
  Did you mean?  T::Set
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/props/decorator.rb:261:in `convert_type_to_class'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/private/methods/call_validation_2_7.rb:106:in `bind_call'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/private/methods/call_validation_2_7.rb:106:in `block in create_validator_method_fast1'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/props/decorator.rb:304:in `prop_defined'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/props/_props.rb:115:in `prop'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/private/methods/call_validation.rb:161:in `bind_call'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/private/methods/call_validation.rb:161:in `validate_call'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/private/methods/call_validation.rb:90:in `block in create_validator_slow'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/props/_props.rb:144:in `const'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/private/methods/call_validation.rb:161:in `bind_call'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/private/methods/call_validation.rb:161:in `validate_call'
# ./vendor/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9460/lib/types/private/methods/call_validation.rb:90:in `block in create_validator_slow'
```

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 # typed: strict
 
-require "set"
-
 # NB: This is not actually a decorator. It's just named that way for consistency
 # with DocumentDecorator and ModelDecorator (which both seem to have been written
 # with an incorrect understanding of the decorator pattern). These "decorators"
@@ -260,6 +258,7 @@ class T::Props::Decorator
     when T::Types::TypedHash, T::Types::FixedHash
       Hash
     when T::Types::TypedSet
+      require "set"
       Set
     when T::Types::Union
       # The below unwraps our T.nilable types for T::Props if we can.
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 # typed: strict
 
+require "set"
+
 # NB: This is not actually a decorator. It's just named that way for consistency
 # with DocumentDecorator and ModelDecorator (which both seem to have been written
 # with an incorrect understanding of the decorator pattern). These "decorators"
--- a/gems/sorbet-runtime/test/types/props/decorator.rb
+++ b/gems/sorbet-runtime/test/types/props/decorator.rb
@@ -51,6 +51,10 @@ class Opus::Types::Test::Props::DecoratorTest < Critic::Unit::UnitTest
     prop :the_hash, T.nilable(T::Hash[String, CustomType])
   end
 
+  class SetStruct < T::Struct
+    prop :set, T.nilable(T::Set[String])
+  end
+
   describe 'typed arrays and hashes' do
     it 'can have string values' do
       doc = StringArrayAndHashStruct.new(
@@ -94,6 +98,14 @@ class Opus::Types::Test::Props::DecoratorTest < Critic::Unit::UnitTest
         include T::Props
         prop :fixed_array_prop, T::Utils.coerce([String, Numeric])
       end
+    end
+  end
+
+  describe 'typed sets' do
+    it 'can have set values' do
+      SetStruct.new(
+        set: ::Set.new(['one', 'two', 'three'])
+      )
     end
   end
 


froydnj (changes requested):
In keeping with other uses of `Set` in the gem, we'd like this `require` to appear at the use of `Set`, rather than unconditionally at gem load time.

Can you also add a test that exhibits the error you're seeing?

ngan:
> we'd like this `require` to appear at the use of `Set`, rather than unconditionally at gem load time.

Hi @froydnj, `Set` is used in this file here: https://github.com/ngan/sorbet/blob/patch-2/gems/sorbet-runtime/lib/types/props/decorator.rb#L263

Are you suggesting I put the `require "Set"` right above that line?

> Can you also add a test that exhibits the error you're seeing?

I could...although I'm not sure this is a normally tested thing since it might be a load order issue. My assumption is if you ran the unit test for just this file it would fail with the same error.

froydnj:
> > we'd like this `require` to appear at the use of `Set`, rather than unconditionally at gem load time.
> 
> Hi @froydnj, `Set` is used in this file here: https://github.com/ngan/sorbet/blob/patch-2/gems/sorbet-runtime/lib/types/props/decorator.rb#L263
> 
> Are you suggesting I put the `require "Set"` right above that line?

Yes.

> > Can you also add a test that exhibits the error you're seeing?
> 
> I could...although I'm not sure this is a normally tested thing since it might be a load order issue. My assumption is if you ran the unit test for just this file it would fail with the same error.

We have tests for this file, but apparently not exhaustive enough tests.  In particular, I'm guessing your code has something like `prop :myprop, T::Set[Symbol]`...and our test suite doesn't have anything like that.

ngan:
@froydnj I've move the `require` and added tests that utilizes `T::Set`.

However, I can't create a failing test because the `subprocess` library requires `Set` itself. If you want a failing test, you can do the following (on master):

1. Delete the subprocess require line: https://github.com/sorbet/sorbet/blob/master/gems/sorbet-runtime/test/test_helper.rb#L14
2. Cd into `gems/sorbet-runtime`
3. Run test for decorator: `ruby test/types/props/decorator.rb`

You should see:

```
	13: from test/types/props/decorator.rb:4:in `<main>'
	12: from test/types/props/decorator.rb:54:in `<class:DecoratorTest>'
	11: from test/types/props/decorator.rb:55:in `<class:SetStruct>'
	10: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb:90:in `block in create_validator_slow'
	 9: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb:161:in `validate_call'
	 8: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb:161:in `bind_call'
	 7: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/props/_props.rb:115:in `prop'
	 6: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/props/decorator.rb:304:in `prop_defined'
	 5: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation_2_7.rb:106:in `block in create_validator_method_fast1'
	 4: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation_2_7.rb:106:in `bind_call'
	 3: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/props/decorator.rb:267:in `convert_type_to_class'
	 2: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation_2_7.rb:106:in `block in create_validator_method_fast1'
	 1: from ~/Development/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation_2_7.rb:106:in `bind_call'
~/Development/sorbet/gems/sorbet-runtime/lib/types/props/decorator.rb:261:in `convert_type_to_class': uninitialized constant T::Props::Decorator::Set (NameError)
Did you mean?  T::Set
```

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -258,6 +258,7 @@ def prop_validate_definition!(name, cls, rules, type)
     when T::Types::TypedHash, T::Types::FixedHash
       Hash
     when T::Types::TypedSet
+      require "set"
```
ngan, line 261: @froydnj out of curiosity... why is this the convention? Wouldn't this needlessly eval `require "set"` a lot if this is a hot path?

jez:
So I think the contract here should be that if you want to use `T::Set`, you have to already have `Set` loaded in your project—using `sorbet-runtime` should not incur that `require` implicitly.

(That would mean that the behavior reported in the original PR is actually expected behavior.)

This ensures that projects downstream of `sorbet-runtime` have control over when and how the (or a) `set` library is loaded.

ngan:
> if you want to use T::Set, you have to already have Set loaded in your project

Ah, I get that intention. Is this documented somewhere? (so I can link to it?)

froydnj:
> @froydnj I've move the `require` and added tests that utilizes `T::Set`.
> 
> However, I can't create a failing test because the `subprocess` library [requires `Set` itself](https://github.com/stripe/subprocess/blob/5ad972f92995291ab4b63aa4071b705906c6ff78/lib/subprocess.rb#L2).

Thanks for the explanation!

Maybe the right path forward here is:

1. remove the `require "subprocess"` and make the Gemfile not reference `subprocess` (grepping suggests that we're not actually using `subprocess` in the tests);
2. add appropriate test, which would fail;
3. make the problematic code check if `Set` is defined and `raise` with an explanatory error message;
4. make the test check for the appropriate error message, thus making it pass.

That would essentially be the documentation you're requesting, I think?

PR #5038, opened 2022-01-04
Fixes Newton, Jacobian and LUSolve type declarations

VitorRigoni:
Hey there, friends!

I'm working on some financial math and I'm using some of the methods available in BigDecimal. More specifically, the Newton module. I was having some trouble with sorbet trying to make my file `typed: strict` until I was joined by @vinistock and we learned that the type definitions for the Newton, Jacobian and LUSolve modules for the BigDecimal gem are incorrect.

They are only declared as class methods when they are, in fact, instance methods as well since all modules are declared as `module_functions`.
Newton: https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal/newton.rb#L44
Jacobian: https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal/jacobian.rb#L26
LUSolve: https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal/ludcmp.rb#L11

This PR fixes the incorrect rbi definitions for the Newton, Jacobian and LUSolve modules.

### Motivation
The missing definitions caused sorbet to behave erratically when trying to use the `Newton::nlsolve`. A `NoMethodError` pops up because sorbet is expecting a class method but is getting a mixed in instance method.

### Test plan
This is the original code from BigDecimal:
Newton: https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal/newton.rb#L44
Jacobian: https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal/jacobian.rb#L26
LUSolve: https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal/ludcmp.rb#L11

Note how they're declared as instance methods and class methods (with the help of `module_functions`).

In my project, I overrode sorbet's rbi with my own rbi that declares these methods as instance methods and sorbet is finally happy and all my tests execute correctly.

### DIFF
--- a/rbi/stdlib/bigdecimal.rbi
+++ b/rbi/stdlib/bigdecimal.rbi
@@ -1378,10 +1378,21 @@ end
 module Jacobian
 
   # Computes the derivative of `f` at `x`. `fx` is the value of `f` at `x`.
+  def dfdxi(f, fx, x, i); end
+
+  # Computes the derivative of `f` at `x`. `fx` is the value of `f` at `x`.
   def self.dfdxi(f, fx, x, i); end
 
   # Determines the equality of two numbers by comparing to zero, or using the epsilon value
+  def isEqual(a, b, zero = 0.0, e = _); end
+
+  # Determines the equality of two numbers by comparing to zero, or using the epsilon value
   def self.isEqual(a, b, zero = 0.0, e = _); end
+
+  # Computes the
+  # [`Jacobian`](https://docs.ruby-lang.org/en/2.6.0/Jacobian.html)
+  # of `f` at `x`. `fx` is the value of `f` at `x`.
+  def jacobian(f, fx, x); end
 
   # Computes the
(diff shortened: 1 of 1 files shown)


vinistock (approved):

jez:
https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal/newton.rb#L32

^ They are defined as `module_function` methods, which means that the same methods are exposed as both instance and singleton methods.

Instead of editing the methods in place, would you mind duplicating these definitions to have one each for instance and singleton?

jez (changes requested):
^

VitorRigoni:
Ah, nice catch, @jez ! Absolutely, will update.

jez (approved):

jez:

→ https://go/builds/bui_Ku6vSaTlGxdGE9
→ https://go/builds/bui_Ku6vQOKSW6Lo4U

PR #5039, opened 2022-01-04
Remove --report-ambiguous-definitions flag

jez:
### Motivation

Early feedback suggested that this was not super annoying, and in many
cases helpful.

For those who wish to achieve the old behavior, feel free to add

    --suppress-error-code=5068

to the command line when invoking Sorbet, or to your `sorbet/config`
file.

### Test plan

Existing tests.


aadi-stripe (approved):
Pls ensure to remove the flag in the default Sorbet wrapper in the Stripe codebase. Thanks!

PR #5040, opened 2022-01-05
Add extra counters to measure completion

jez:
### Motivation

We're planning some work to make Sorbet better at repsonding to
completion requests. This is the sort of metric that should help confirm
whether the changes we're making are working.

Future work: do something similar for hover and go to def. First wanted
to get feedback on approach.

### Test plan

See included automated tests.


froydnj (approved):
Makes sense to me!

jez:
Looks like @jvilk-stripe is out today—I'm going to go ahead and merge this, and we can revisit later if we want to change something about this.

PR #5042, opened 2022-01-05
Remove matrix argument for +@ and -@

vinistock:
Remove the argument for `Matrix` methods `+@` and `-@`.

### Motivation

These two methods don't accept arguments according to the [current docs](https://ruby-doc.org/stdlib-3.1.0/libdoc/matrix/rdoc/Matrix.html#method-i-2B-40) and the [docs linked in the RBI itself](https://docs.ruby-lang.org/en/2.7.0/Vector.html#method-i-2B-40).

### Test plan

Should be covered by current tests.

### DIFF
--- a/rbi/stdlib/matrix.rbi
+++ b/rbi/stdlib/matrix.rbi
@@ -129,12 +129,12 @@ class Vector
   # [`Vector`](https://docs.ruby-lang.org/en/2.7.0/Vector.html) addition.
   def +(x); end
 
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_KuR6NbdC3AwuZ1
→ https://go/builds/bui_KuR6YiF79HJeyV

froydnj (approved):
Thanks!

PR #5045, opened 2022-01-05
Fix AST cast bug in packager for eigenclass declarations in debug mode

aadi-stripe:
Currently, there is a bug due to the fact that the AST node associated with eigenclass declarations (`class << self`) is differently shaped in debug mode, leading to an ENFORCE failure due to an unsafe `non_null` AST cast. This change guards against the error by using a safe cast.

### Motivation

In order to make some tooling in Stripe's codebase work in `--stripe-packages` mode, we need to make sure the debug build of Sorbet works correctly.

### Test plan

Added a unit test.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -518,12 +518,16 @@ public:
             return tree;
         }
 
-        auto &constantLit = ast::cast_tree_nonnull<ast::UnresolvedConstantLit>(classDef.name);
-        pushConstantLit(&constantLit);
+        ast::UnresolvedConstantLit *constantLit = ast::cast_tree<ast::UnresolvedConstantLit>(classDef.name);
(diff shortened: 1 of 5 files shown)


aadi-stripe:

→ https://go/builds/bui_KuV9fafAurFmi1
→ https://go/builds/bui_KuV9lPc8QBHw32

aadi-stripe:
ptal @jez

jez (changes requested):
Can you write an automated test for this? Lurking something into an existing test or writing a new one are both fine

aadi-stripe:
> Can you write an automated test for this? Lurking something into an existing test or writing a new one are both fine

Done

jez (approved):

PR #5047, opened 2022-01-05
Fix infinite loop in PackageDB when a path beginning in '/' is passed in

aadi-stripe:
Also turn off stripePackages checking for file passed as --minimize-to-rbi  flag

### Motivation

RBI generation using the `minimize-to-rbi` flag currently doesn't work with `--stripe-packages`. This is because of an infinite loop in the code that finds the package where a file lives. This PR:
* Fixes the infinite loop
* Turns off `--stripe-packages` checking for the `minimize-to-rbi` file itself.

### Test plan

See included automated tests.


aadi-stripe:

→ https://go/builds/bui_KuXjSMQPNLJ4Cn
→ https://go/builds/bui_KuXjCuKzy9G4MD

aadi-stripe:
ptal @jez

jez (approved):
Thanks!

PR #5049, opened 2022-01-06
Skip processing `typed: ignore` package files

aisamanra:
### Motivation
Right now we'll crash if we ever mark a `__package.rb` file as `typed: ignore`. This ensures that we don't process those.

### Test plan

See included automated tests. In addition to testing this locally, I also modified `pipeline_test_runner.cc` to actually skip `typed: ignore` files (whereas it was parsing them before) so I could add a real test of this.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -905,7 +905,9 @@ unique_ptr<PackageInfoImpl> getPackageInfo(core::MutableContext ctx, ast::Parsed
                                            const vector<std::string> &extraPackageFilesDirectoryPrefixes) {
     ENFORCE(package.file.exists());
     ENFORCE(package.file.data(ctx).sourceType == core::File::Type::Package);
-    // Assumption: Root of AST is <root> class.
+    // Assumption: Root of AST is <root> class. (This won't be true
+    // for `typed: ignore` files, so we should make sure to catch that
+    // elsewhere.)
     ENFORCE(ast::isa_tree<ast::ClassDef>(package.tree));
     ENFORCE(ast::cast_tree_nonnull<ast::ClassDef>(package.tree).symbol == core::Symbols::root());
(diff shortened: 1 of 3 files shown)


packager/packager.cc
```diff
@@ -1465,6 +1467,16 @@ vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers
...
             if (FileOps::getFileName(file.file.data(gs).path()) == PACKAGE_FILE_NAME) {
+                if (file.file.data(gs).strictLevel == core::StrictLevel::Ignore) {
+                    // if the `__package.rb` file is at `typed:
+                    // ignore`, then we haven't even parsed it, which
+                    // means none of the other stuff here is going to
+                    // actually work (since it all assumes we've got a
+                    // file to actually analyze.) If we've got a
+                    // `typed: ignore` package, then skip it.
+                    continue;
+                }
```
froydnj, line 1478: For my own edification, how does this interact with:

https://github.com/sorbet/sorbet/blob/e23d803535bd2f8454d3fc0be79be1725a7bcc0e/packager/packager.cc#L1365-L1370

Is `Ignore` meant to be a temporary measure for certain `__package.rb` files or something else?

aisamanra: I don't think we intend to use `typed: ignore` files for packages, but we should handle them gracefully. In general, `typed: ignore` tells Sorbet, "Please don't even look at this file," so I think here it's reasonable to keep that same contract. I can imagine, for example, someone wanting to experiment with, "What if this package wasn't even here?" and therefore changing it to `typed: ignore`.

It's worth noting that the above error _doesn't_ show up if we mark a package file as `typed: ignore`, since `typed: ignore` will silence all errors.

aprocter-stripe (approved):

PR #5050, opened 2022-01-06
Add support for `sorbet` directive in packages to specify minimum typed level

aisamanra:
### Motivation
This takes what used to be an internal (but package-aware) rubocop and turns it into a thing enforced by Sorbet. This makes it a type error if a file has lower than the declared `min_typed_level` for a package.

### Test plan

See included automated tests. Added tests for both correct and incorrect behavior here.


aisamanra:
Not that we won't be able to land this until getting rid of the rubocop disables in the codebase, so while I want to get the reviews rolling, I won't click merge until we've gotten rid of those disables.

core/Files.cc
```diff
@@ -107,6 +107,57 @@ StrictLevel File::fileStrictSigil(string_view source) {
...
 
+pair<size_t, size_t> File::fileStrictSigilLocation(string_view source) {
+    /*
+     * StrictLevel::None: <none>
+     * StrictLevel::False: # typed: false
+     * StrictLevel::True: # typed: true
+     * StrictLevel::Strict: # typed: strict
+     * StrictLevel::String: # typed: strong
+     * StrictLevel::Autogenerated: # typed: autogenerated
+     */
```
aisamanra, line 118: I split this out but wanted to keep it next to the other code which was responsible for finding the sigil. It's possible I suppose to combine them, but I tried that and it felt clunky—in particular, this is trying to find the beginning of the comment (i.e. the `#`) while the other code finds that and then walks forward to only select the string itself.

jez: not sure I understand. wouldn't it be possible to build one in terms of the other? or at least, factor a shared helper, and have two functions, both which post process the "parse result" of finding the comment?

core/Files.h
```diff
@@ -81,6 +82,7 @@ class File final {
     friend class GlobalState;
     friend class ::sorbet::core::serialize::SerializerImpl;
 
+    static std::pair<size_t, size_t> fileStrictSigilLocation(std::string_view source);
```
jez, line 85: Could we have the return type be a `core::LocOffsets` here?

It would mean we could swap the `#include <utility>` for a `struct LocOffsets` forward declaration, and have a more descriptive container for the result, than just two unnamed `size_t`'s.

core/errors/packager.h
```diff
@@ -21,5 +21,7 @@ constexpr ErrorClass DefinitionPackageMismatch{3713, StrictLevel::False};
 constexpr ErrorClass ImportConflict{3714, StrictLevel::False};
 constexpr ErrorClass InvalidExportForTest{3715, StrictLevel::False};
 constexpr ErrorClass ExportConflict{3716, StrictLevel::False};
+constexpr ErrorClass InvalidTypedLevelDeclaration{3717, StrictLevel::False};
```
jez, line 24:
```suggestion
constexpr ErrorClass InvalidMinTypedLevelDeclaration{3717, StrictLevel::False};
```

Is this name more accurate?

core/errors/packager.h
```diff
@@ -21,5 +21,7 @@ constexpr ErrorClass DefinitionPackageMismatch{3713, StrictLevel::False};
...
+constexpr ErrorClass TypedLevelViolation{3718, StrictLevel::False};
```
jez, line 25:
```suggestion
constexpr ErrorClass MinTypedLevelViolation{3718, StrictLevel::False};
```

ditto

core/tools/generate_names.cc
```diff
@@ -422,6 +422,14 @@ NameDef names[] = {
     {"export_", "export"},
     {"export_for_test"},
     {"restrict_to_service"},
+    {"sorbet"},
+    {"minTypedLevel", "min_typed_level"},
+    {"testsMinTypedLevel", "tests_min_typed_level"},
+    {"ignore"},
+    {"_false", "false"},
+    {"_true", "true"},
```
jez, line 430:
```suggestion
    {"false_", "false"},
    {"true_", "true"},
```

Looking at the other names in this file, the convention is to put the underscore after the name when it would otherwise share a name with a C++ reserved word.

packager/packager.cc
```diff
@@ -1411,17 +1476,64 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file) {
...
     if (pkg.exists()) {
         auto &pkgImpl = PackageInfoImpl::from(pkg);
 
+        auto isTestFile = core::packages::PackageDB::isTestFile(ctx, file);
+        // check that the files are indeed as typed as they should be
+        if (!isTestFile && file.originalSigil < pkgImpl.minStrictLevel()) {
+            auto [begin, end] = core::File::fileStrictSigilLocation(file.source());
+            if (auto e = ctx.beginError(core::LocOffsets{static_cast<uint32_t>(begin), static_cast<uint32_t>(end)},
+                                        core::errors::Packager::TypedLevelViolation)) {
+                e.setHeader("File `{}` is marked as `{}`, but it belongs to package `{}` which stipulates a minimum "
```
jez, line 1517: Some slight wording suggestions here:

- drop "File", it's clear from context
- "is marked as **typed: {}**" (there is also `compiled:` and `frozen_string_literal`, and at Stripe even others like `fast-load`)
- "which requires files be at least `typed: {}`" (trying to avoid using "stipulates", seemed out of place, but up to you)

packager/packager.cc
```diff
@@ -1411,17 +1476,64 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file) {
     return file;
 }
 
+string_view levelToSigil(core::StrictLevel level) {
```
jez, line 1479: This is duplicated with the function in realmain.cc right? Is it time to factor that into `core::StrictLevel`?

packager/packager.cc
```diff
@@ -1411,17 +1476,64 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file) {
...
+                                        core::errors::Packager::TypedLevelViolation)) {
+                e.setHeader("File `{}` is marked as `{}`, but it belongs to package `{}` which stipulates a minimum "
+                            "typed level of `{}`",
+                            ctx.file.data(ctx).path(), levelToSigil(file.originalSigil), pkgImpl.name.toString(ctx),
+                            levelToSigil(pkgImpl.minStrictLevel()));
+            }
+        }
+        if (isTestFile && file.originalSigil < pkgImpl.minTestStrictLevel()) {
+            auto [begin, end] = core::File::fileStrictSigilLocation(file.source());
+            if (auto e = ctx.beginError(core::LocOffsets{static_cast<uint32_t>(begin), static_cast<uint32_t>(end)},
```
jez, line 1525: If we returned a `core::LocOffsets` in `fileStrictSigilLocation`, we could encapsulate this cast and pass along the parse result unchanged.

packager/packager.cc
```diff
@@ -705,9 +717,61 @@ struct PackageInfoFinder {
             }
         }
 
+        if (send.fun == core::Names::sorbet() && send.numKwArgs() == 2) {
```
jez, line 720: Do we emit an error if you pass more or fewer args than 2? Should we?

aisamanra: We do: that's handled by the type of the `sorbet` method, which has two require keyword arguments. If we don't pass either argument, or we pass too many, then we'll get those errors from the typechecking pass. That said, all valid invocations of this will have only two arguments, so if we don't have exactly two, then we know another error will get raised later on.

packager/packager.cc
aisamanra: This has been done.

packager/packager.cc
aisamanra: This has been factored out into the same place as `StrictLevel`. (Sadly, C++ doesn't let you add methods—static or instance—to `enum class` values, so it's a standalone method defined there instead.)

packager/packager.cc
aisamanra: The wording here has been updated! I agree that "stipulates" was weird; it's been replaced.

core/Files.cc
aisamanra: I've gone and factored them together. There are a _few_ situations in which this new approach works differently, but they're rare and weird: in particular, in situations where a user has writen, say:
```
# typed: invalid
# typed: true
```
we previously would have looked at the `typed: invalid` line, discovered that it didn't contain a sensible string, and then gone forward and found the `# typed: true` and allowed it. We now instead will only look at the first `# typed:` and treat that as canonical. I think this is an appropriate trade-off for now, and in the future I think we should additionally raise an error when `typed:` lines contain nonsensical or invalid values (which is what the TODO left in this code suggested) in which case this change actually gets us closer. But I figure it's still worth calling out as a change in semantics in some edge-cases.

test/testdata/packager/package-must-be-strict/__package.rb
```diff
@@ -1,6 +1,6 @@
+  # typed: true
+# ^^^^^^^^^^^^^ error: Package files must be at least `# typed: strict`
```
aisamanra, line 2: This was previously reported with location `{0, 0}` but I've tweaked it to use the identified location of the `typed:` sigil. That also means we can offer an autocorrect, although I haven't in this PR.

aisamanra:
I had to fix a bunch of things in the codebase in order to make this landable, but those are nearly in, so I should get this reviewed. I believe I've addressed all your existing concerns: let me know if there's anything else!

aisamanra:

→ https://go/builds/bui_L0WcTyYXt900ux
→ https://go/builds/bui_L0WcWrqDxY7cy3

jez (approved):

PR #5067, opened 2022-01-07
Prototype package RBI generator

jvilk-stripe:
WIP: Prototype package RBI generator

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -303,9 +303,13 @@ ConstantLit::fullUnresolvedPath(const core::GlobalState &gs) const {
         return nullopt;
     }
     ENFORCE(this->resolutionScopes != nullptr && !this->resolutionScopes->empty());
+
+    // TODO: investigate why this is necessary, it being after the ENFORCE suggests that it only happens in a release
+    // build
     if (this->resolutionScopes == nullptr) {
         return nullopt;
     }
+
     vector<core::NameRef> namesFailedToResolve;
     auto *nested = this;
     {
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -303,6 +303,9 @@ ConstantLit::fullUnresolvedPath(const core::GlobalState &gs) const {
         return nullopt;
     }
     ENFORCE(this->resolutionScopes != nullptr && !this->resolutionScopes->empty());
+    if (this->resolutionScopes == nullptr) {
+        return nullopt;
+    }
     vector<core::NameRef> namesFailedToResolve;
     auto *nested = this;
     {
--- a/core/Loc.cc
+++ b/core/Loc.cc
@@ -257,6 +257,7 @@ optional<string_view> Loc::source(const GlobalState &gs) const {
         return nullopt;
     } else {
         auto source = this->file().data(gs).source();
+        // TODO: this should be unnecessary
         if (endPos() < source.size()) {
             return source.substr(beginPos(), endPos() - beginPos());
         } else {
--- a/core/Loc.cc
+++ b/core/Loc.cc
@@ -257,7 +257,11 @@ optional<string_view> Loc::source(const GlobalState &gs) const {
         return nullopt;
     } else {
         auto source = this->file().data(gs).source();
-        return source.substr(beginPos(), endPos() - beginPos());
+        if (endPos() < source.size()) {
+            return source.substr(beginPos(), endPos() - beginPos());
+        } else {
+            return nullopt;
+        }
     }
 }
 
--- a/core/Names.cc
+++ b/core/Names.cc
@@ -361,6 +361,12 @@ NameRef NameRef::addEq(GlobalState &gs) const {
     return gs.enterNameUTF8(nameEq);
 }
 
+NameRef NameRef::lookupWithEq(const GlobalState &gs) const {
+    auto name = this->dataUtf8(gs);
+    string nameEq = absl::StrCat(name->utf8, "=");
+    return gs.lookupNameUTF8(nameEq);
+}
+
 NameRef NameRef::addQuestion(GlobalState &gs) const {
     auto name = this->dataUtf8(gs);
     string nameEq = absl::StrCat(name->utf8, "?");
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -454,8 +454,8 @@ string TypeMemberRef::show(const GlobalState &gs, ShowOptions options) const {
         return fmt::format("T.attached_class (of {})", attached.show(gs, options));
     }
     auto owner = sym->owner;
-    // Don't show T.class_of(Foo)::Field; show Foo::Field.
-    if (sym->owner.asClassOrModuleRef().data(gs)->isSingletonClass(gs)) {
+    if (options.showForRBI && sym->owner.asClassOrModuleRef().data(gs)->isSingletonClass(gs)) {
+        // Don't show T.class_of(Foo)::Field; show Foo::Field when printing an RBI
         owner = sym->owner.asClassOrModuleRef().data(gs)->attachedClass(gs);
     }
     return showInternal(gs, owner, sym->name, COLON_SEPARATOR);
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -448,9 +448,10 @@ string TypeMemberRef::show(const GlobalState &gs, ShowOptions options) const {
     if (sym->name == core::Names::Constants::AttachedClass()) {
         auto attached = sym->owner.asClassOrModuleRef().data(gs)->attachedClass(gs);
         ENFORCE(attached.exists());
-        // TODO: This change is needed for RBI generation.
-        return "T.attached_class";
-        // return fmt::format("T.attached_class (of {})", attached.show(gs, options));
+        if (options.showForRBI) {
+            return "T.attached_class";
+        }
+        return fmt::format("T.attached_class (of {})", attached.show(gs, options));
     }
     auto owner = sym->owner;
     // Don't show T.class_of(Foo)::Field; show Foo::Field.
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -448,9 +448,16 @@ string TypeMemberRef::show(const GlobalState &gs, ShowOptions options) const {
     if (sym->name == core::Names::Constants::AttachedClass()) {
         auto attached = sym->owner.asClassOrModuleRef().data(gs)->attachedClass(gs);
         ENFORCE(attached.exists());
-        return fmt::format("T.attached_class (of {})", attached.show(gs, options));
+        // TODO: This change is needed for RBI generation.
+        return "T.attached_class";
+        // return fmt::format("T.attached_class (of {})", attached.show(gs, options));
     }
-    return showInternal(gs, sym->owner, sym->name, COLON_SEPARATOR);
+    auto owner = sym->owner;
+    // Don't show T.class_of(Foo)::Field; show Foo::Field.
+    if (sym->owner.asClassOrModuleRef().data(gs)->isSingletonClass(gs)) {
+        owner = sym->owner.asClassOrModuleRef().data(gs)->attachedClass(gs);
+    }
+    return showInternal(gs, owner, sym->name, COLON_SEPARATOR);
 }
 
 TypePtr ArgInfo::argumentTypeAsSeenByImplementation(Context ctx, core::TypeConstraint &constr) const {
--- a/core/types/printing.cc
+++ b/core/types/printing.cc
@@ -158,10 +158,10 @@ string AliasType::toStringWithTabs(const GlobalState &gs, int tabs) const {
 }
 
 string AliasType::show(const GlobalState &gs, ShowOptions options) const {
-    // RBI generation: show() is supposed to print what the user typed, and the user doesn't
-    // type <Alias: SymbolName>!
-    return this->symbol.show(gs);
-    // return fmt::format("<Alias: {} >", this->symbol.showFullName(gs));
+    if (options.showForRBI) {
+        return this->symbol.show(gs);
+    }
+    return fmt::format("<Alias: {} >", this->symbol.showFullName(gs));
 }
 
 string AndType::toStringWithTabs(const GlobalState &gs, int tabs) const {
--- a/core/types/printing.cc
+++ b/core/types/printing.cc
@@ -131,17 +131,20 @@ string ShapeType::show(const GlobalState &gs, ShowOptions options) const {
         } else {
             fmt::format_to(std::back_inserter(buf), ", ");
         }
-        auto keyLiteral = cast_type_nonnull<LiteralType>(key);
-        auto underlying = keyLiteral.underlying(gs);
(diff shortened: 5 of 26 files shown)


core/Symbols.cc
```diff
@@ -425,7 +425,9 @@ string ClassOrModuleRef::show(const GlobalState &gs) const {
     if (sym->name == core::Names::Constants::AttachedClass()) {
         auto attached = sym->owner.asClassOrModuleRef().data(gs)->attachedClass(gs);
         ENFORCE(attached.exists());
-        return fmt::format("T.attached_class (of {})", attached.show(gs));
+        // TODO: This change is needed for RBI generation.
+        return "T.attached_class";
+        // return fmt::format("T.attached_class (of {})", attached.show(gs));
```
jez, line 430: This makes error messages substantially more confusing. We should make separate printing methods before landing this so that we don't have to regress this error message.

core/Symbols.cc
```diff
@@ -454,9 +456,16 @@ string TypeMemberRef::show(const GlobalState &gs) const {
...
     }
-    return showInternal(gs, sym->owner, sym->name, COLON_SEPARATOR);
+    auto owner = sym->owner;
+    // Don't show T.class_of(Foo)::Field; show Foo::Field.
+    if (sym->owner.asClassOrModuleRef().data(gs)->isSingletonClass(gs)) {
+        owner = sym->owner.asClassOrModuleRef().data(gs)->attachedClass(gs);
+    }
+    return showInternal(gs, owner, sym->name, COLON_SEPARATOR);
```
jez, line 468: there's a subtle distinction here

```ruby
class Foo
  class << self
    Field = 42
  end
end

p Foo::Feld # => NameError
```

I'd love to see where this comes up specifically, because there might be a better way to detect the case(s) we're trying to handle here.

Another one:

```ruby
class A
  extend T::Generic

  MyTemplate = type_template
end

p A::MyTemplate
```

In this case, Sorbet would print it as `T.class_of(A)::MyTemplate` even though `A::MyTemplate` would resolve. But:

```ruby
class A
  extend T::Generic

  class << self
    MyTemplate = type_member
  end
end

p A::MyTemplate # => NameError
```

This would result in the same globalstate, but different runtime behavior.

(That's kind of why I want to learn more about what the actual code that informed this change looks like.)

core/Names.cc
```diff
@@ -361,6 +361,12 @@ NameRef NameRef::addEq(GlobalState &gs) const {
     return gs.enterNameUTF8(nameEq);
 }
 
+NameRef NameRef::lookupWithEq(const GlobalState &gs) const {
+    auto name = this->dataUtf8(gs);
+    string nameEq = absl::StrCat(name->utf8, "=");
+    return gs.lookupNameUTF8(nameEq);
+}
+
```
jez, line 369: looks dead? is this used somewhere?

core/Loc.cc
```diff
@@ -257,7 +257,11 @@ optional<string_view> Loc::source(const GlobalState &gs) const {
         return nullopt;
     } else {
         auto source = this->file().data(gs).source();
-        return source.substr(beginPos(), endPos() - beginPos());
+        if (endPos() < source.size()) {
+            return source.substr(beginPos(), endPos() - beginPos());
+        } else {
+            return nullopt;
+        }
```
jez, line 264: 1. I think there is an off by one here? Shouldn't `endPos()` be allowed to be equal to `source.size()`?

2. Do we think we can instead make it an invariant that `endPos() <= source.size()`? It seems like we've gotten this far with that being the case, and I'd love to maintain it if possible.

core/types/printing.cc
```diff
@@ -53,7 +53,13 @@ string LiteralType::toStringWithTabs(const GlobalState &gs, int tabs) const {
...
 
 string LiteralType::show(const GlobalState &gs) const {
-    return fmt::format("{}({})", this->underlying(gs).show(gs), showValue(gs));
+    // RBI generator: Users type the value, not `String("value")`.
+    // if (literalKind == LiteralTypeKind::Symbol) {
+    // HACK: Restore T.deprecated_enum :|
+    return fmt::format("T.deprecated_enum([{}])", showValue(gs));
+    //}
+    // return showValue(gs);
+    // return fmt::format("{}({})", this->underlying(gs).show(gs), showValue(gs));
```
jez, line 62: This seems wrong. In what cases are we printing literal types?

If we're printing them as keys of `Shape` types, we should probably print them as normal Ruby values.

If they're coming up in places outside of shape types, we should either:

1. not print them. Consider this case:


```
# typed: strict

A = "hello"
```

The GlobalState will show `A` as having type `String("hello")` but we don't actually have to add a `T.let` to the generated RBI to record that fact.

2. Widen them to their underlying type, but this might cause bugs

(tl;dr: i think that we should be able to have a heuristic like "does static field actually need `T.let` annotation" + special handling for shape types)

core/types/printing.cc
```diff
@@ -146,7 +154,10 @@ string AliasType::toStringWithTabs(const GlobalState &gs, int tabs) const {
 }
 
 string AliasType::show(const GlobalState &gs) const {
-    return fmt::format("<Alias: {} >", this->symbol.showFullName(gs));
+    // RBI generation: show() is supposed to print what the user typed, and the user doesn't
+    // type <Alias: SymbolName>!
+    return this->symbol.show(gs);
+    // return fmt::format("<Alias: {} >", this->symbol.showFullName(gs));
```
jez, line 160: Do you have an example of where this comes up? I'm having a tough time knowing what the right behavior is here absent an example.

core/types/printing.cc
```diff
@@ -398,6 +409,10 @@ string AppliedType::show(const GlobalState &gs) const {
             }
             return to_string(buf);
         } else {
+            if (this->klass.data(gs)->isSingletonClass(gs)) {
+                // T.class_of(klass)[arg1, arg2] is never valid syntax.
+                return this->klass.show(gs);
+            }
```
jez, line 415: While this is true that this is not (currently) valid syntax, it does make error messages substantially more understandable when working with complex (generic) code. For example:

```ruby
# typed: true
extend T::Sig

class A
  extend T::Sig
  extend T::Generic

  Elem = type_template
end

T.reveal_type(A) # error: `T.class_of(A)[T.untyped]`
```

To a lot of people, they might not realize that:

- Sorbet treats this like an AppliedType (generic) at all
- the `type_template` has been defaulted to `T.untyped`

Again, would love to see how this comes up in practice so we can know whether there's a better heuristic.

definition_validator/validator.cc
```diff
@@ -481,13 +485,19 @@ void validateSealed(core::Context ctx, const core::ClassOrModuleRef klass, const
...
         auto file = bestNonRBIFile(ctx, klass);
-        if (!absl::c_any_of(superClass.data(ctx)->sealedLocs(ctx), [file](auto loc) { return loc.file() == file; })) {
-            if (auto e =
-                    ctx.beginError(getAncestorLoc(ctx, classDef, superClass), core::errors::Resolver::SealedAncestor)) {
-                e.setHeader("`{}` is sealed and cannot be inherited by `{}`", superClass.show(ctx), klass.show(ctx));
-                for (auto loc : superClass.data(ctx)->sealedLocs(ctx)) {
-                    e.addErrorLine(loc, "`{}` was marked sealed and can only be inherited in this file",
-                                   superClass.show(ctx));
+        // HACK: Package.test.rbi may extend a class defined in Package.rbi because it contains a
+        // Package class only referenced by test exports.
```
jez, line 489: Is it not possible to guarantee that the sealed subclasses are defined in the same file as the parent class?

definition_validator/validator.cc
```diff
@@ -464,6 +464,10 @@ void validateSealedAncestorHelper(core::Context ctx, const core::ClassOrModuleRe
         }
 
         auto klassFile = bestNonRBIFile(ctx, klass);
+        // HACK for RBI generator (see other comment)
```
jez, line 467:
```suggestion
        // HACK for RBI generator (see comment in validateSealed)
```

main/realmain.cc
```diff
@@ -753,6 +760,49 @@ int realmain(int argc, char *argv[]) {
...
+
+        if (!packageFiles.empty()) {
+            auto packageFileRefs = pipeline::reserveFiles(gs, packageFiles);
+            auto packages = pipeline::index(*gs, packageFileRefs, opts, *workers, nullptr);
+            packager::RBIGenerator::run(*gs, move(packages), opts.packageRBIOutput, *workers);
+        } else {
+            logger->error("No package files found!");
+            return 1;
+        }
+    }
```
jez, line 804: are you sure you want this way down here, after all the metrics have been reported?

seems like we would want this to live up closer to e.g. where the code to handle `--minimize-rbi` lives.

core/Names.cc
elliottt: https://github.com/sorbet/sorbet/blob/9878d4597ba1b9d37a1e96b2383ec926ace170e5/packager/rbi_gen.cc#L611

jez: ah, github had collapsed the diff in that file. thanks!

ast/Trees.cc
```diff
@@ -302,6 +302,9 @@ ConstantLit::fullUnresolvedPath(const core::GlobalState &gs) const {
         return nullopt;
     }
     ENFORCE(this->resolutionScopes != nullptr && !this->resolutionScopes->empty());
+    if (this->resolutionScopes == nullptr) {
+        return nullopt;
+    }
```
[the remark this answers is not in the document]
jvilk-stripe: I encountered this crash while testing this code, and don't know why it happened/was moving fast to get the RBI generator working. There isn't a test that covers this AFAIK.

core/Loc.cc
jvilk-stripe: 1. I believe that's right.
2. I forget why this happened. This fixed a crash. I think it might've been bad argumentName locs created in a rewriter pass?

core/types/printing.cc
jvilk-stripe: > This seems wrong. In what cases are we printing literal types?

Look at the one rbi-gen test. We have examples at Stripe where these types are used as type parameters.

core/types/printing.cc
jvilk-stripe: I believe this is covered by the rb-gen test. Go ahead and uncomment it and see if the test passes.

core/types/printing.cc
jvilk-stripe: `TypePtr::show` is documented to be "what the user typed", and "showWithMoreInfo" is "not necessarily what the user typed, but useful for error messages".

Could we just use `showWithMoreInfo` more often?

definition_validator/validator.cc
jvilk-stripe: `source.rb`:

```rb
class MyClass < SealedSubclass

end

class TestMyClass < SealedSubclass

end
```

`__package.rb`:

```
export MyClass
export_for_test TestMyClass
```

Now the RBI for the package contains `MyClass`, and the RBI for the test package contains `TestMyClass`, violating this invariant.

main/realmain.cc
jvilk-stripe: Not sure at all; I just plopped this code in what looked like a safe position. Feel free to move.

core/types/printing.cc
jez: It seems this is specifically to support `T.deprecated_enum`? `T.deprecated_enum` is basically treated as `T.untyped` internally. When we serialize the RBI, we should be able to just put `T.untyped` there, and have the codebase type check equivalently under the generated RBI.

core/types/printing.cc
jez: I feel pretty strongly that no. By far the more common use case is error messages, and I think that the shortest name should be reserved for the most common use case. I added `showWithMoreInfo` as a hack, and it's currently only used in a handful of places (hover, reveal type).

If we wanted to change the names and make it like `show` and `showPrintable` I'd be fine with that.

definition_validator/validator.cc
jez: Gotcha. The problem with this hack right now is that I don't think this should be an option for all people writing RBI files to circumvent `sealed!`'s invariants. If we could figure out concretely that an RBI file is a package-exported RBI file, that might be a better solution?

core/Symbols.cc
```diff
@@ -448,9 +448,16 @@ string TypeMemberRef::show(const GlobalState &gs) const {
     if (sym->name == core::Names::Constants::AttachedClass()) {
         auto attached = sym->owner.asClassOrModuleRef().data(gs)->attachedClass(gs);
         ENFORCE(attached.exists());
-        return fmt::format("T.attached_class (of {})", attached.show(gs));
+        // TODO: This change is needed for RBI generation.
+        return "T.attached_class";
+        // return fmt::format("T.attached_class (of {})", attached.show(gs));
```
elliottt, line 453: Migrating @jez's comment forward from `ClassOrModuleRef::show`:

This makes error messages substantially more confusing. We should make separate printing methods before landing this so that we don't have to regress this error message.

core/Symbols.cc
elliottt: I removed this case, as the only one that fires is on `TypeMemberRef`.

core/types/printing.cc
elliottt: The downside to emitting an rbi with `T.untyped` instead would be that hover would become substantially less useful.

jez: @elliottt That explanation doesn't quite make sense to me, because Sorbet doesn't have an internal notion of `T.deprecated_enum`. In `type_syntax.cc`, when it sees `T.deprecated_enum([:symbol])`, it's going to convert that to the type `Symbol` anyways. So why not just explicitly widen the type to the underlying when serializing the packager RBI file?

I looked at the example in the test, and still can't quite connect how we're going from "there's a `LiteralType` on the Symbol" to "we should wrap this up in `T.deprecated_enum`". It seems like the two concepts are completely independent in the type system.

elliottt: That works for me -- if we're already turning it into the underlying type, then let's just do that in the generated rbis as well.

elliottt:

→ https://go/builds/bui_KxYQua9sFDAAwK
→ https://go/builds/bui_KxYQSA9a2iTcaE

main/realmain.cc
```diff
@@ -667,6 +678,59 @@ int realmain(int argc, char *argv[]) {
...
+                logger->error("Serializing package RBIs requires one input folder.");
+                return 1;
+            }
+
+            auto relativeIgnorePatterns = opts.relativeIgnorePatterns;
+            auto it = absl::c_find(relativeIgnorePatterns, "/__package.rb");
+            if (it != relativeIgnorePatterns.end()) {
+                relativeIgnorePatterns.erase(it);
+            } else {
+                Exception::raise("Couldn't find ignore pattern.");
```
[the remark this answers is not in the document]
elliottt: This is enforcing that we don't typecheck any package.rb files before generating rbis.

elliottt (approved):
All new functionality is behind the `--package-rbi-output` flag, so it's safe to merge this branch so we can start iterating on improvements.

PR #5070, opened 2022-01-07
EnforcePackagePrefix: disallow sends/method defs in wrong scope

ngroman:
This extends prefix enforcement to also disallow method definitions and send's outside of the correct scope. The most complex part of this logic is the book-keeping to avoid many duplicates on a single line. For example `X = T.let(foo, Integer)` is a constant assignment and multiple sends.

### Motivation
More accurate packing enforcement

### Test plan

See included automated tests.

Ran on Stripe's codebase.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -585,6 +585,31 @@ public:
         return original;
     }
 
+    ast::ExpressionPtr preTransformMethodDef(core::Context ctx, ast::ExpressionPtr original) {
+        auto &def = ast::cast_tree_nonnull<ast::MethodDef>(original);
+        checkBehaviorLoc(ctx, def.declLoc);
+        return original;
+    }
+
+    ast::ExpressionPtr preTransformSend(core::Context ctx, ast::ExpressionPtr original) {
+        checkBehaviorLoc(ctx, original.loc());
+        return original;
+    }
+
+    void checkBehaviorLoc(core::Context ctx, core::LocOffsets loc) {
+        if (rootConsts > 0 || nameParts.empty()) {
+            return;
+        }
+        auto &pkgName = requiredNamespace(ctx.state);
+        if (!isPrefix(pkgName, nameParts)) {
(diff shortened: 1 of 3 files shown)


ngroman:
cc @aadi-stripe @aisamanra

elliottt:

→ https://go/builds/bui_KvIhZxOeHlVUAg
→ https://go/builds/bui_KvIhrW1ugn6LYz

packager/packager.cc
```diff
@@ -544,6 +551,14 @@ class EnforcePackagePrefix final {
...
+            ENFORCE(errorCount == 0);
+            return tree;
+        }
+
+        if (errorCount == 1) {
+            errorCount--; // Still need to popConstantLit for 0'th
+        } else if (errorCount > 1) {
+            errorCount--;
             return tree;
         }
```
elliottt, line 563:
```suggestion
        if (errorCount > 0) {
            errorCount--;
            // only continue if this was the first occurrence of the error
            if (errorCount > 0) {
                return tree;
            }
        }
```

packager/packager.cc
```diff
@@ -585,6 +606,62 @@ class EnforcePackagePrefix final {
...
+    }
+
+    void checkBehaviorLoc(core::Context ctx, core::LocOffsets loc) {
+        ENFORCE(errorCount == 0);
+        if (rootConsts > 0 || nameParts.empty()) {
+            return;
+        }
+        auto &pkgName = requiredNamespace(ctx.state);
+        if (!isPrefix(pkgName, nameParts)) {
+            errorCount++;
```
elliottt, line 656: This should hold here, right?

```suggestion
            ENFORCE(errorCount == 0);
            errorCount++;
```

elliottt (approved):
This makes sense to me. I had some suggestions for how to make the use of `errorCount` a bit more clear.

nroman-stripe:

→ https://go/builds/bui_KvJHDEfqmJNhnB
→ https://go/builds/bui_KvJHgLgGhHqFqH

PR #5073, opened 2022-01-10
add a `tok_view` method to `lexer`

froydnj:
### Motivation

Sorbet's lexer is...less than efficient partly due to its insistence on turning everything into a `std::string`, all the time.  This PR is the first step towards addressing that inefficiency by letting the lexer hand out `std::string_view`s instead.

There are no uses of the new `tok_view` function; this is purely an interface change.  There are other changes I have that build on this one, but I figure I'd float this first.

### Test plan

See included automated tests.


elliottt (approved):
:shipit:

PR #5074, opened 2022-01-11
Make the cli-trace test more robust

jez:
### Motivation

There were a lot of autogenerated IDs in the state names, that mean we
would have to run this test every time the ypp file changes which is
annoying and not useful. The test just cares that the option works.

### Test plan

See included automated tests.


test/cli/trace_parser/trace_parser.sh
```diff
@@ -2,4 +2,9 @@
 
 set -euo pipefail
 
-main/sorbet --silence-dev-message --trace-parser test/cli/trace_parser/trace_parser.rb 2>&1
+tmp="$(mktemp)"
+trap 'rm -f "$tmp"' EXIT
+
+main/sorbet --silence-dev-message --trace-parser test/cli/trace_parser/trace_parser.rb > "$tmp" 2>&1
+
+head -n 4 "$tmp"
```
froydnj, line 10:
```suggestion
# We are just testing that the --trace-parser option works.
# The full output is too dependent on the productions in the parser.
main/sorbet --silence-dev-message --trace-parser test/cli/trace_parser/trace_parser.rb > "$tmp" 2>&1

head -n 4 "$tmp"
```

froydnj (approved):

PR #5079, opened 2022-01-11
use `tok_view` in more places throughout the lexer

froydnj:
### Motivation

Let's not materialize `std::string` from the lexing stream until we absolutely have to; building on #5076 and #5073, we have the capability to do that in a lot more places.  Saves ~4% of instructions during the parsing phase on `ruby/lib/rdoc/markdown.rb`.

I don't think this covers absolutely everywhere, but it covers a lot of the easy places.  The commits should be reviewable on their own.

### Test plan

See included automated tests.

### DIFF
--- a/third_party/parser/cc/lexer.rl
+++ b/third_party/parser/cc/lexer.rl
@@ -1504,7 +1504,7 @@ void lexer::set_state_expr_value() {
 
   # Ruby is context-sensitive wrt/ local identifiers.
   action local_ident {
-    auto ident = tok();
+    auto ident = tok_view();
 
     emit(token_type::tIDENTIFIER, ident);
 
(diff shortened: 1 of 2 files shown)


jez (approved):
Do you want to double check this on Stripe's codebase? I think this is likely perfectly fine, but after having gotten sanitizer crashes after making a tiny change to this part of the codebase yesterday, I'm less confident in my judgements here 😅

Great to see lexer changes! In case you hadn't already noticed, there should be a *ton* of low hanging fruit there—we've never spent any time improving this code.

froydnj:

→ https://go/builds/bui_KwiRJQANvikSCg
→ https://go/builds/bui_KwiRuYOlvqW6dz

froydnj:
> Do you want to double check this on Stripe's codebase? I think this is likely perfectly fine, but after having gotten sanitizer crashes after making a tiny change to this part of the codebase yesterday, I'm less confident in my judgements here sweat_smile

Good call.  I think things are relatively safe, since the actual token stream presented to the parser copies everything into `std::string`, but it'd be good to double-check.

PR #5085, opened 2022-01-11
Add an empty options struct to `TypePtr::show`

elliottt:
### Motivation

Making it easier to land #5067 by adding config options to `TypePtr::show`.

### Test plan

See included automated tests.

### DIFF
--- a/core/TypePtr.h
+++ b/core/TypePtr.h
@@ -311,8 +311,10 @@ public:
         return toStringWithTabs(gs);
     }
 
+    struct ShowOptions {};
+
     // User visible type. Should exactly match what the user can write.
-    std::string show(const GlobalState &gs) const;
+    std::string show(const GlobalState &gs, const ShowOptions options = ShowOptions{}) const;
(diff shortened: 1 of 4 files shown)


core/TypePtr.h
```diff
@@ -311,8 +311,10 @@ class TypePtr final {
         return toStringWithTabs(gs);
     }
 
+    struct ShowOptions {};
+
     // User visible type. Should exactly match what the user can write.
-    std::string show(const GlobalState &gs) const;
+    std::string show(const GlobalState &gs, const ShowOptions options = ShowOptions{}) const;
```
aprocter-stripe, line 317: Thoughts on passing `options` by reference?

aprocter-stripe (approved):
This seems reasonable to me! Just one question, but I'll accept any answer so it's non-blocking. ;)

core/TypePtr.h
elliottt: My plan was to keep it pretty small (a struct of bools), so copying should be fine. If it becomes a problem, we can revisit that decision.

PR #5086, opened 2022-01-11
Move code out of subtyping.cc that's not related to subtyping

elliottt:
### Motivation

Keeping related code together.

### Test plan

See included automated tests.


core/types/subtyping.cc
```diff
@@ -1386,57 +1386,4 @@ bool Types::equivNoUntyped(const GlobalState &gs, const TypePtr &t1, const TypeP
     return isAsSpecificAs(gs, t1, t2) && isAsSpecificAs(gs, t2, t1);
 }
 
-bool ClassType::derivesFrom(const GlobalState &gs, ClassOrModuleRef klass) const {
```
froydnj, line 1389: I can see an argument for leaving these in subtyping.cc, but I don't feel super strongly about it.

froydnj (approved):

core/types/subtyping.cc
elliottt: All the other implementations are in different files, so it felt reasonable to move them.

PR #5090, opened 2022-01-12
prework: Factor most of method completion into helper

jez:
### Motivation

When we make the parser better, we're going to want to suggest method
and locals in more places, so I wanted to factor out the logic in
anticipation.

This change comes with no behavioral changes, as witnessed by the tests.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/completion.h
+++ b/main/lsp/requests/completion.h
@@ -10,8 +10,39 @@ class CompletionItem;
 class CompletionTask final : public LSPRequestTask {
     std::unique_ptr<CompletionParams> params;
 
+    // TODO(jez) Let's delete this?
     void findSimilarConstants(const core::GlobalState &gs, const core::lsp::ConstantResponse &resp, core::Loc queryLoc,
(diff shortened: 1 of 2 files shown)


main/lsp/requests/completion.h
```diff
@@ -10,9 +10,40 @@ class CompletionItem;
...
+        bool suggestKeywords;
+
+        // If not exists(), won't suggest locals for the enclosing method.
+        core::MethodRef enclosingMethod;
+
+        // If empty(), won't suggest constants.
+        core::lsp::ConstantResponse::Scopes scopes;
+    };
+    std::vector<std::unique_ptr<CompletionItem>> getCompletionItems(LSPTypecheckerDelegate &typechecker,
+                                                                    SearchParams params);
```
elliottt, line 45: What do you think about making params a constant reference? As it is, it will copy everything to pass it in to `getCompletionItems`.

elliottt (approved):

PR #5091, opened 2022-01-12
Stop suggesting local variables with angle brackets

jez:
### Motivation

^

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -120,6 +120,12 @@ bool hasAngleBrackets(string_view haystack) {
     return absl::c_any_of(haystack, [](char c) { return c == '<' || c == '>'; });
 }
 
+bool hasSimilarName(const core::GlobalState &gs, core::NameRef name, string_view pattern) {
+    string_view view = name.shortName(gs);
+    auto fnd = view.find(pattern);
+    return fnd != string_view::npos;
+}
+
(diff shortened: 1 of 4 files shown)


main/lsp/requests/completion.cc
```diff
@@ -120,6 +120,12 @@ bool hasAngleBrackets(string_view haystack) {
     return absl::c_any_of(haystack, [](char c) { return c == '<' || c == '>'; });
 }
 
+bool hasSimilarName(const core::GlobalState &gs, core::NameRef name, string_view pattern) {
+    string_view view = name.shortName(gs);
+    auto fnd = view.find(pattern);
+    return fnd != string_view::npos;
+}
```
jez, line 127: This function was only used in one file: completion.cc, so I figured I'd just move it in there.

test/testdata/lsp/completion/locals.rb
```diff
@@ -90,3 +90,10 @@ class Wrapper
   x_ # error: does not exist
   # ^ completion: x_inside_class
 end
+
+class B
+  def implicit_block_arg
+    blk # error: does not exist
+    #  ^ completion: (nothing)
```
jez, line 97: check the commits for a before/after

froydnj (approved):
This just affects `<blk>`, right?  Or are there other weird cases where this comes into play?

PR #5092, opened 2022-01-12
make the lexer's requirement for null-terminated buffers more explicit

froydnj:
Inside `lexer`'s constructor, we take the provided source buffer and append two null terminator characters to it:

https://github.com/sorbet/sorbet/blob/1094ce3ee273cdb4ecad81ecc803f33bf97db0ed/third_party/parser/cc/lexer.rl#L123-L126

We do this because the lexer's definition, as written, considers the null terminator to indicate EOF:

https://github.com/sorbet/sorbet/blob/1094ce3ee273cdb4ecad81ecc803f33bf97db0ed/third_party/parser/cc/lexer.rl#L694

Why we add *two* null terminators is a mystery to me, but empirically, we have test case failures with only one.

I think this is the wrong way to go about things: the lexer's contract is that null is EOF and therefore the caller should be in charge of enforcing that, not the lexer itself.  This PR changes things so that the caller appends the null terminators itself, and changes the lexer so that it stores a `string_view` of the source buffer, rather than the source buffer itself.

This change has several benefits:

1. As above, I think this is somewhat clearer, though I can see reasonable people disagreeing.
2. The current code actually copies the original file twice: once to create a `string` for the parser driver, and then again inside the lexer's constructor to append the null terminators.  The approach taken here at least creates only one copy.  (I guess we could have `File` itself append the null terminators when it was reading the file and have a separate `sourceNullTerminated` view on `File`'s data, if we wanted to stamp out copies.  The current `source()` would of course continue to return the non-null-terminated view.  Opinions welcome on that front.)
3. `.data()` on `string` (which we call quite a bit in the lexer) may require branches in `libc++`, so this change would be a micro-optimization for such platforms.

### Motivation

Somewhat clearer code.

### Test plan

See included automated tests.

### DIFF
--- a/parser/Parser.cc
+++ b/parser/Parser.cc
@@ -53,7 +53,13 @@ unique_ptr<Node> Parser::run(sorbet::core::GlobalState &gs, core::FileRef file, 
                              std::vector<std::string> initialLocals) {
     Builder builder(gs, file);
     auto source = file.data(gs).source();
-    ruby_parser::typedruby27 driver(string(source.begin(), source.end()), Builder::interface);
+    // The lexer requires that its buffers end with a null terminator, which core::File
+    // does not guarantee.
+    string buffer;
+    buffer.reserve(source.size() + 1);
+    buffer += source;
+    buffer += '\0';
+    ruby_parser::typedruby27 driver(buffer, Builder::interface);
 
     for (string local : initialLocals) {
         driver.lex.declare(local);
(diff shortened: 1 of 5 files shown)


jez (approved):

froydnj:

→ https://go/builds/bui_Kx64NQ1351D03I
→ https://go/builds/bui_Kx64CdbricI4Lr

PR #5094, opened 2022-01-12
make `token`'s external interface primarily oriented around `string_view`

froydnj:
This PR changes the primary way consumers interact with `token`: the preferred interface should be accessing a `string_view` of the token.  Because we have a fair number of places that still expect tokens to be able to turn themselves into `const string&`, we needed to have an `asString()` method on tokens...but such uses should be considered bugs, IMHO.  The commit to change the name of the accessor and the commit to rename/change its interface (and also add `asString()` to fix compile errors) have been separated out so it's easier to see what got changed where.

### Motivation

I think `token` should store `string_view` instead of `string`, and this would be the first step towards getting us there.  I want to do this because `string_view` is significantly smaller than `string`, so that would make tokens smaller (and we could come up with other ways to compact the data `token` stores further, which would let us be slightly more efficient with the token pool).  `string_view` is also less complicated to destroy, which would mean that destroying the lexer's token pool would do less work.  (For cases where we have to munge the data for the token such that it doesn't correspond to the original source, my plan is to have a scratch buffer that a token's `string_view` could point at.  I am not 100% sure this is going to work out, but my intuition is that it should wind up being faster overall.)

The one risk that's worth calling out is that `string_view` is more subject to being misused than `string`.  With `string`, you know that the relevant data is stored in the object itself, so you don't have to think very hard about object lifetimes.  With `string_view`, you would have to be more conscious of this.  (In Rust terms, the lifetimes of the `string_view` in `token` would be tied to the lifetime of the lexer.)  I am mildly optimistic that this won't be a problem in practice, since tokens don't escape outside of the parser+lexer as-is.

### Test plan

See included automated tests.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -221,9 +221,9 @@ public:
 
     unique_ptr<Node> arg(const token *name) {
         core::LocOffsets loc = tokLoc(name);
-        checkReservedForNumberedParameters(name->string(), loc);
+        checkReservedForNumberedParameters(name->view(), loc);
 
-        return make_unique<Arg>(loc, gs_.enterNameUTF8(name->string()));
+        return make_unique<Arg>(loc, gs_.enterNameUTF8(name->view()));
     }
 
     unique_ptr<Node> args(const token *begin, sorbet::parser::NodeVec args, const token *end, bool check_args) {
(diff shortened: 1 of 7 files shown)


elliottt (approved):
🚀

froydnj:

→ https://go/builds/bui_Kx41AO9Wfi5N9U
→ https://go/builds/bui_Kx41cQv7iYO11J

PR #5096, opened 2022-01-12
Remove an unnecessary case from `ClassOrModuleRef::show`

elliottt:
`T.attached_class` is a type member on the singleton class, so having a special case for printing it as a `ClassOrModuleRef` seems like a leftover from when there was only one type of symbol.

### Motivation

Cleaning up code.

### Test plan

See included automated tests.


froydnj (approved):
:man_shrugging: This seems right to me!

PR #4238, opened 2021-05-28
Update `import_whitequark.sh` script

Morriar:
### Motivation

First changing the script, subsequent pull-requests will bring the imported tests and related parser changes.

Also rename ruby28.rb into ruby30.rb following upstream change.

### Test plan

See included automated tests.

cc. @vinistock

PR #5098, opened 2022-01-12
Update whitequark ref and ruby version for Ruby 3.1 support

KaanOzkan:
Co-authored-by: Alexandre Terrasa <alexandre.terrasa@shopify.com>

First changing the script, subsequent pull-requests will bring the imported tests and related parser changes.

### Motivation
First step in supporting new grammar in Ruby 3.1

### Test plan

See included automated tests.

@Morriar

### DIFF
--- a/tools/scripts/import_whitequark.sh
+++ b/tools/scripts/import_whitequark.sh
@@ -14,8 +14,8 @@
 set -euo pipefail
 set -x
 
-REF=49ed4dddfb1bf44d579fda001862149f4b9f7970
-TARGET_RUBY_VERSION="3.0"
+REF=3e421d6bc4d7a480d0a0e7aae23afe714a61ea87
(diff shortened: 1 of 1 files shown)


elliottt (changes requested):
What do you think about making this change on the branch where you're integrating the parser changes? If we make this change on master, the script won't work until the parser is updated.

KaanOzkan:
That's a good point, I wasn't sure if the script was used on an ongoing basis. We followed the previous way of splitting the work into smaller PRs: https://github.com/sorbet/sorbet/pull/4238 by starting from this change.

I think we will end up having 4-5 parser change PRs each supporting a different feature. Do you want to combine them into 1 PR alongside this change?

elliottt (approved):
Ah, I didn't realize that there was precedent for starting with the script update. Looks good to me 👍

PR #5101, opened 2022-01-12
Recover from syntax errors in assignments

jez:
### Motivation

More fast paths, better completion.

Review by commit.

### Note about `error_node`

I've chosen to implement this as an (unresolved) constant reference. I
think that making a fake constant node will be easier than having to
update all the places in the codebase that might want to exhaustively
handle all pieces of syntax.

I've chosen to make it a constant node because it's allowed in more
places (versus say, a variable name or method name).

Another option was to make it an empty `begin; end` node, but that would
become an `EmptyTree` after desugar, and I think that this is more
useful in case we want to detect this somewhere later in the pipeline.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/completion/missing_assign_rhs.rb
+++ b/test/testdata/lsp/completion/missing_assign_rhs.rb
@@ -1,6 +1,9 @@
 # typed: true
 
-def test_missing_rhs(x)
-  y =
-  #  ^ completion: (nothing)
-end # error: unexpected token "end"
+class A
+  def foo; end
+  def test_missing_rhs(x)
+    y =
+    #  ^ completion: x, y, foo, test_missing_rhs, ...
+  end # error: unexpected token "end"
+end
--- a/test/testdata/lsp/completion/missing_assign_rhs.rb
+++ b/test/testdata/lsp/completion/missing_assign_rhs.rb
@@ -1,0 +1,6 @@
+# typed: true
+
+def test_missing_rhs(x)
+  y =
+  #  ^ completion: (nothing)
+end # error: unexpected token "end"
--- a/test/testdata/lsp/completion/missing_assign_rhs.rb
+++ b/test/testdata/lsp/completion/missing_assign_rhs.rb
@@ -2,25 +2,32 @@
 
 class A
   def foo; end
+  attr_writer :foo_writer
+
   def test_missing_rhs(x)
     y =
-    #  ^ completion: x, y, foo, test_missing_rhs, ...
+    #  ^ completion: x, y, foo, ...
   end # error: unexpected token "end"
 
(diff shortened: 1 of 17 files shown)


test/testdata/lsp/completion/missing_assign_rhs.rb
```diff
@@ -12,4 +12,15 @@ def test_variable_after_method(x)
...
+  def test_variable_end_same_line(x)
+    puts 'before'
+    y =end # error: unexpected token "end"
+    #  ^ completion: x, y, foo, test_missing_rhs, ...
+
+  def test_variable_end_same_line_after(x)
+    # This one is admittedly janky. There's a note about it in the parser.
+    puts 'before'
+    y =end # error: unexpected token "end"
+    #     ^ completion: x, y, foo, test_missing_rhs, ...
```
aprocter-stripe, line 25: So I read the comment in the parser that you're referencing, but still having trouble understanding why the completion pops up in a different location for `test_variable_end_same_line_after`, whose body is lexically identical to `test_variable_end_same_line` (except for the comment... maybe that matters?). I guess the error tokens in these two examples must span different locations, but I don't understand why, and I can't work out exactly which ranges they're spanning. Could you elaborate a bit?

aprocter-stripe (approved):
This looks great, assuming the remaining test failures aren't show-stoppers! (At a glance I'm guessing they're just exp files that need updating.) The use of a synthetic constant node for errors seems reasonable to me.

test/testdata/lsp/completion/missing_assign_rhs.rb
jez: LSP queries never match on AST nodes that have zero width.

If we had used `@2.end, @3.begin` when creating the `error_node`, that would be the end of the `=` and the start of the `end`, which would thus give that AST node zero width, so that `error_node` would never participate in LSP query responses.

Instead, the LSP query would return the next-enclosing AST nodde, which in that case would **maybe** be the assignment node `x =`, but in fact I think LSP properly treats end positions in locs as exclusive, so it would actually result in a query for the enclosing method def. Or something idk. It for sure wouldn't be the error_node.

jez:
> This looks great, assuming the remaining test failures aren't show-stoppers! (At a glance I'm guessing they're just exp files that need updating

Yep, the test failures are because some files with syntax errors parse better now, and need to be updated.

jez:

→ https://go/builds/bui_KxZ9Tb12IL8ATS
→ https://go/builds/bui_KxZ9dQUy3v0jOs

jez:

→ https://go/builds/bui_Kxb0ALmveUewf5
→ https://go/builds/bui_Kxb0y7HCQG9Euu

PR #5102, opened 2022-01-12
Use ShowOptions on ref types as well

elliottt:
Move `ShowOptions` off of `TypePtr` and into its own header, and then add it as an argument to `SymbolRef::show` etc.

Reviewing by commits:
- e8e9febf0 Move `ShowOptions` out of `TypePtr`, and into its own header
- dbb42c87c `ShowOptions` doesn't need to be `const`
- b2b38a598 Add non-defaulted `ShowOptions` arguments to symbol ref show functions so that it's easy to tell where `core/printing.cc` should be updated
- 059e15e4e Add defaults for symbol ref show methods

### Motivation

Pre-work towards landing #5067

### Test plan

See included automated tests.

### DIFF
--- a/core/ShowOptions.h
+++ b/core/ShowOptions.h
@@ -1,0 +1,11 @@
+#ifndef CORE_SHOW_OPTIONS_H
+#define CORE_SHOW_OPTIONS_H
+
+namespace sorbet::core {
+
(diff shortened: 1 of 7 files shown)


jez (approved):
thanks!

PR #5104, opened 2022-01-13
Track location for method name specifically

jez:
### Motivation

There are a couple of motivations:

-   most autocorrects I've written end up trying to compute this information with
    bad string source parsing.

-   for better completion requests, I want to do something special for a snippet
    like this:

    ```ruby
    x.
    y = 1
    ```

    right now this parses as `x.y=(1)`. Sorbet then gets back an LSP response
    that says that the method under the cursor at the moment is called `y=`, and
    so it goes looking for methods called `y=` on `x`, which either might not
    exist, or might preclude valid results if we had instead searched for
    methods like `""` on `x`.

    By knowing the `funLoc`, I'll be able to say "is the cursor location
    actually inside or at the end of the method I just got back?" and if not,
    I'll be able to do something smarter, like throw away the `y=` and search
    for `""`

    Empirically, getting this to work is going to be critically important for
    making `x.<TAB>` useful.

-   for better error messages. if you do something like

    ```ruby
    x.each do
      # many lines
    end
    ```

    but it turns out that `each` doesn't exist on `x`, sorbet will show the
    error message from the start of `x` through to the `end` token, which could
    easily be dozens of lines. People have complained about this in the past,
    but generally stopped asking because they probably figured it would never
    get changed.

### Commit summary

I added things piecemeal as I pushed the parameter down through the pipeline.

This was entirely manual, with no `sed`, because I wanted to be more confident
that the locs I was picking were good.

- **wip: Add methodLoc to parser::{Send,CSend}** (e53e5a352)

- **wip: Add funLoc to ast::Send** (788093f27)

- **Reorder cfg::Send fields** (3fd3058a0)

- **wip: Add funLoc to cfg::Send** (4568b8152)

- **wip: Add fun to core::CallLocs** (4f1f0f5f5)

- **Use funLoc in calls.cc** (b830fe7b1)

- **Factor out Loc::empty()** (246ea901e)

- **Don't set funLoc for [] and []=** (1a28d5e14)

- **Some error locations change** (2f7332f9f)

- **Change line splitting test** (cdd23175e)

  It looks like this test was trying to show that multi-line locs in error
  messages get the little `...`. It so happened that the error it was
  using to demonstrate that behavior no longer will do that (which is
  good!). I left the existing test to capture the improvement, and also
  added another example of a test where this will still happen (so that we
  can catch any future changes that somehow regress multi-line loc error
  printing).

- **Update cli tests** (8879337ae)

- **Fix bug in parse-tree-json-with-locs** (ca3adc037)

  This format looks entirely broken: the locs aren't actually there--only
  the `declLoc` (and now `methodLoc` fields are).

- **Leave a TODO for thought** (f18a3e186)

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_KxWRa0eEGmaYel
→ https://go/builds/bui_KxWRgZqRYtBSNU

ast/desugar/Desugar.cc
```diff
@@ -240,8 +240,8 @@ ExpressionPtr desugarDString(DesugarContext dctx, core::LocOffsets loc, parser::
         return mergeStrings(dctx, loc, std::move(stringsAccumulated));
     } else {
         auto recv = MK::Constant(loc, core::Symbols::Magic());
-        return MK::Send(loc, std::move(recv), core::Names::stringInterpolate(), interpArgs.size(),
-                        std::move(interpArgs));
+        return MK::Send(loc, std::move(recv), core::Names::stringInterpolate(), loc.copyWithZeroLength(),
```
aprocter-stripe, line 243: So just fwiw I started to read through all of these to make sure the choice of loc made sense to me, and got about a quarter of the way through the file not having seen anything objectionable, and then realized how long it was and decided that was enough scrutiny. If there's less mechanical stuff in this file I missed maybe let me know :)

core/types/calls.cc
```diff
@@ -1764,7 +1764,8 @@ void applySig(const GlobalState &gs, const DispatchArgs &args, DispatchResult &r
          ++loc) {
         callLocsArgs.emplace_back(*loc);
     }
-    CallLocs callLocs{args.locs.file, args.locs.call, callLocsReceiver, callLocsArgs};
+    // TODO(jez) Double check this
+    CallLocs callLocs{args.locs.file, args.locs.call, callLocsReceiver, args.locs.fun, callLocsArgs};
```
aprocter-stripe, line 1770: Did you get a chance to address this TODO? Not quite sure what it's referring to.

core/types/calls.cc
```diff
@@ -571,12 +571,14 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
         }
+        auto funLoc = args.funLoc();
+        auto errLoc = (funLoc.exists() && funLoc.beginPos() != funLoc.endPos()) ? args.funLoc() : args.callLoc();
         // This is a hack. We want to always be able to build the error object
         // so that it is not immediately sent to GlobalState::_error
         // and recorded.
         // Instead, the error always should get queued up in the
         // errors list of the result so that the caller can deal with the error.
-        auto e = gs.beginError(args.callLoc(), errors::Infer::UnknownMethod);
+        auto e = gs.beginError(errLoc, errors::Infer::UnknownMethod);
```
aprocter-stripe, line 581: Nice!

test/cli/line-splitting/line-splitting.sh
```diff
@@ -48,3 +48,70 @@ main/sorbet --silence-dev-message -e '
...
+# typed: true
+extend T::Sig
+
+sig {returns(String)}
+def returns_string; ""; end
+
+sig {params(x: Integer).void}
+def takes_integer(x); end
+
+takes_integer(
```
aprocter-stripe, line 62: Nice replacement for the old test here!

test/cli/cache-dsl/cache-dsl.out
```diff
@@ -25,7 +25,7 @@ test/cli/cache-dsl/attr_accessor.rb:8: Method `prop` does not exist on `A` https
 Errors: 2
 test/cli/cache-dsl/attr_accessor.rb:7: Method `prop=` does not exist on `A` https://srb.help/7003
      7 |a.prop = 7
-        ^^^^^^^^^^
+          ^^^^^^
```
aprocter-stripe, line 28: Elsewhere in the review (I've lost track of where) I think I was thinking out loud on what the best choice is for the span given a send LHS. Wasn't thinking that carets are one of the users of that info. :) Given that, I think I  do like including the ` =` in the method loc.

aprocter-stripe (approved):
LGTM, with just a few non-blocking comments! I'm not feeling totally confident about the handling of `a.foo = bar`, but based on what evidence we have so far (i.e., how the carets marking `foo =` end up looking), I think the choice you made is probably the best one, and we can always revisit it if it causes friction later.

parser/Builder.cc
```diff
@@ -273,12 +274,17 @@ class Builder::Impl {
...
         core::LocOffsets loc = lhs->loc.join(rhs->loc);
 
+        // TODO(jez) What should we put for the methodLoc here?
+        // x.foo = 1
+        // The behavior below currently does `foo =` for the methodLoc
         if (auto *s = parser::cast_node<Send>(lhs.get())) {
             s->args.emplace_back(std::move(rhs));
-            return make_unique<Send>(loc, std::move(s->receiver), s->method, std::move(s->args));
+            return make_unique<Send>(loc, std::move(s->receiver), s->method, s->methodLoc.join(tokLoc(eql)),
+                                     std::move(s->args));
```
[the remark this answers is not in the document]
jez: > I'm a little confused here by the existing behavior (which predates this PR) where method apparently is unmodified from the lhs send. (So for x.name = "fred" it's going to be name, not name=.) Am I misreading something here? If not, maybe the location should simply be methodLoc to match.

It hits a different reduction rule (`lhs` which forces it to reduce using the `attrAsgn` builder, which adds the `=` in anticipation of being used in a context where the `tEQL` is coming.

https://github.com/sorbet/sorbet/blob/9071226edb18bf3fe3b2f11fcde73b0c5f839d3a/parser/Builder.cc#L339

parser/Builder.cc
```diff
@@ -1155,13 +1165,14 @@ class Builder::Impl {
...
+            return make_unique<Send>(loc, transformCondition(std::move(receiver)), core::Names::bang(), tokLoc(not_),
                                      sorbet::parser::NodeVec());
         }
 
         ENFORCE(begin != nullptr && end != nullptr);
         auto body = make_unique<Begin>(tokLoc(begin).join(tokLoc(end)), sorbet::parser::NodeVec());
-        return make_unique<Send>(tokLoc(not_).join(body->loc), std::move(body), core::Names::bang(),
+        // TODO(jez) seems wrong
+        return make_unique<Send>(tokLoc(not_).join(body->loc), std::move(body), core::Names::bang(), tokLoc(not_),
                                  sorbet::parser::NodeVec());
```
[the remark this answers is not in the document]
jez: Yeah, but I don't think it ever makes sense to do the things that I would have otherwise special cased. Like if we set this, you still can't use the `methodLoc` to rename it to another method, nor show completion results, etc. I'd rather have it be zero width if you can't use it meaningfully.

ast/desugar/Desugar.cc
jez: yeah it's all mechanical.

core/types/calls.cc
jez: yeah, when we rewrite sig to `Sorbet::Private::Static.sig`, we re-use all parts of the `Send` node, and only update the `recv` and `args` field, so the `funLoc` is unchanged.

parser/Builder.cc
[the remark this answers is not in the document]
aprocter-stripe: Ah, I missed that! Thanks!

parser/Builder.cc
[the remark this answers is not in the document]
aprocter-stripe: > I'd rather have it be zero width if you can't use it meaningfully.

Got it, that makes sense!

PR #5105, opened 2022-01-13
avoid dedenting strings in the common case

froydnj:
### Motivation

Faster code: almost every file has strings, and depending on how long they are, this bypasses reallocating the string and re-hashing it in `GlobalState`.

### Test plan

See included automated tests.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -30,9 +30,29 @@ using std::vector;
 
 namespace sorbet::parser {
 
-string Dedenter::dedent(string_view str) {
-    string out;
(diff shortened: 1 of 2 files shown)


jez (approved):
Looks good. This is definitely one we'll want to test on Stripe's codebase (dedenting off-by-one errors has definitely crashed Sorbet in the past).

froydnj:

→ https://go/builds/bui_KxSsWCPipq6Y5T
→ https://go/builds/bui_KxSs7QxnHSJxZm

PR #5109, opened 2022-01-13
move `third_party/parser` underneath `parser`

froydnj:
`third_party/parser` is in a bit of a weird state: it's imported from the upstream project [typedruby/typedruby](https://github.com/typedruby/typedruby).  But the upstream project is not under active development, and we've added enough stuff to `third_party/parser` that it's probably time to admit that we've a) permanently forked it and b) (eventually) take action based on that (e.g. eliminating some of the abstraction that the more general `typedruby` project needed, but isn't relevant to Sorbet's purposes).

This PR just moves the code around and does the minimum fixing up to build (including formatting).  `typedruby_fork` is maybe not the best name, but it is at least accurate (`parser/parser` seemed like asking for trouble).  I did not try to deal with moving all the relevant code into `parser/` directly and untangling the resulting build issues; I think doing that is best left to a separate PR.

### Motivation

Laying the foundation for better code in the future.

### Test plan

See included automated tests.


froydnj:
> seems good, maybe putting the files in `parser/` or `parser/parser/` is better than `typedruby_fork`?

I had some FUD about putting things in `parser/parser` because there was a `//parser:parser` Bazel target, but everything seems to have worked out OK.

I'll follow up with a PR or two to collapse everything into `parser/`.

PR #5112, opened 2022-01-14
Recover from trailing comma in method def

jez:
### Motivation

Better error handling. User request.

### Test plan

See included automated tests.


test/testdata/parser/method_def_trailing_comma.rb
```diff
@@ -0,0 +1,10 @@
+# typed: false
+
+def comma_after_pos_args(x, y,)
+end
```
froydnj, line 4: Two questions:

1. Do we need tests for splat/optional/optional kwarg/kwsplats?
2. I imagine other parser changes that you're making have tests to verify that we get reasonable parse trees from erroneous output, which is great.  But do we want to put such tests in a separate directory to make it clearer that they are testing error recovery in some way?

froydnj (commented):
This is very cool from an LSP standpoint, but we still have poor error diagnostics:

```
test/testdata/parser/method_def_trailing_comma.rb:3: unexpected token ")" https://srb.help/2001
     3 |def comma_after_pos_args(x, y,)
                                      ^

test/testdata/parser/method_def_trailing_comma.rb:6: unexpected token ")" https://srb.help/2001
     6 |def comma_after_kwargs(x:, y:,)
                                      ^

test/testdata/parser/method_def_trailing_comma.rb:9: unexpected token ")" https://srb.help/2001
     9 |def comma_after_all_kwargs(a, b, x:, y:,)
                                                ^
Errors: 3
```

Is it possible to also address the error here and point out that the user needs to delete the trailing comma, rather than complaining about the `)` token (which is, after all, supposed to be there?

jez:
> Is it possible to also address the error here and point out that the user needs to delete the trailing comma, rather than complaining about the ) token

Yeah, this should be very doable, and is something I hadn't thought about at all.

We're probably going to want to do this for ~every error production. Do you mind if I make a ticket to come back to it? It's likely that doing them all at once will be easier, because I'll be able to notice a pattern and factor out one or more helpers.

test/testdata/parser/method_def_trailing_comma.rb
jez: > Do we need tests for splat/optional/optional kwarg/kwsplats?

Sure, I can add some. I don't think I handled those cases, because I think they're less common, but may as well as long as they don't provide conflicts.

If they don't already work, I'll probably punt on addressing them to a follow up PR.

> But do we want to put such tests in a separate directory to make it clearer that they are testing error recovery in some way?

I don't think there's a ton of value in centralizing them in one place. The fact that tests have `error: unexpected token` assertions in them should suggest that they're testing parse error recovery. I've been putting most of the tests in the `test/testdata/lsp/completion/` folder actually, because most of them in addition to having a parse tree exp file have `completion` assertions.

Again, seems like purely aesthetic, I'm happy to consolidate but I don't see a lot of addition value over just having the test at all. lmk.

jez: > Sure, I can add some. I don't think I handled those cases, because I think they're less common, but may as well as long as they don't provide conflicts.

I looked into this, and it looks like it'll be a bit of a combinatorial explosion to write rules for these cases. I'm going to punt on them for this PR.

jez: https://github.com/sorbet/sorbet/issues/5120

froydnj (approved):
I think this is fine as far as it goes; I share some of the same concerns as the comments on #5114 , but maybe we either a) need to come up with a better error-handling strategy in bison; or b) rewrite the parser manually so we can have greater control over the errors.

PR #5114, opened 2022-01-14
Substantially better error recovery for method args

jez:
### Motivation

This causes a huge source of pain when using Sorbet in large codebases, because
small edits frequently cause large swaths of a file to be thrown away.

### Test plan

See included automated tests.


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1512,6 +1512,15 @@ int maybe_inject_tNL_impl(parser::symbol_type &yyla, ruby_parser::typedruby27 &d
...
                     }
+                // There's an explanation of what's going on in this rule in the `assocs` rule.
+                | args tCOMMA fcall error assocs opt_block_arg
+                    {
+                      auto &args = $1;
+                      args->emplace_back(driver.build.call_method(self, nullptr, nullptr, $3, nullptr, nullptr, nullptr));
+                      args->emplace_back(driver.build.associate(self, nullptr, $5, nullptr));
+                      args->concat($6);
+                      $$ = args;
+                    }
```
elliottt, line 1523: Does this rule need to add an error?

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -3676,12 +3695,62 @@ f_opt_paren_args: f_paren_args
...
                 | assocs tCOMMA assoc
                     {
                       auto &list = $1;
                       list->emplace_back($3);
                       $$ = list;
                     }
+                | assocs tCOMMA error
+                    {
+                      $$ = $1;
```
elliottt, line 3715: Should this rule add an error?

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -3676,12 +3695,62 @@ f_opt_paren_args: f_paren_args
...
+                    }
+                | assocs tCOMMA tIDENTIFIER error
+                    {
+                      auto &list = $1;
+                      // TODO(jez) Return a special function call like `<kw__a` so that you can complete kwargs
+                      auto err = driver.build.call_method(self, nullptr, nullptr, $3, nullptr, nullptr, nullptr);
+                      list->emplace_back(driver.build.pair_keyword(self, $3, err));
+                      $$ = list;
+                    }
```
elliottt, line 3724: Should this rule add an error?

elliottt (changes requested):
Looking forward to using this!

My one concern right now is that there are a fair number of productions that match `error` and don't explicitly add errors. Either we should add comments explaining why it's not necessary to do so in those cases, or explicitly add errors.

jez:

→ https://go/builds/bui_KxxpXn6VKL0Did
→ https://go/builds/bui_Kxxpw5kIvgI7Ey

elliottt (approved):

elliottt:
After discussion from a DM: I misunderstood how `error` works in bison, and assumed that it would eat the parse error like it does in [happy](https://www.haskell.org/happy/doc/html/sec-error.html). As errors will be raised in all these cases, all my previous comments don't apply 👍

PR #5116, opened 2022-01-14
turn off bison trace support for non-debug builds

froydnj:
This PR is not quite finished, because there'd be some discussion about what to do with `test/cli/test_trace_parser`.  But turning off the trace support results in ~4% faster parsing, which seems like a very easy win.

I removed `%define parse.assert` along the way because it doesn't do anything for our current C++ parser: it would if we used bison's `variant` type for the value type, rather than the union we currently do.  Honestly, that sounds like kind of an attractive proposition, since I think some issues around memory management get much easier...

### Motivation

Speed.

### Test plan

See included automated tests.


jez:
> Honestly, that sounds like kind of an attractive proposition, since I think some issues around memory management get much easier...

Yep. This is one of the parser refactorings that I would probably do after getting rid of `ForeignPtr`.

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -27,9 +27,7 @@ using namespace std::string_literals;
 %define api.prefix {typedruby27}
 %define api.value.type { union parser_value }
 %define api.token.constructor false
-%define parse.assert
 %define parse.error verbose
-%define parse.trace
```
jez, line 32: What do you think about leaving a comment here about how we're passing `--debug` in the BUILD file, and not unconditionally in a directive? I can imagine myself grepping for `parse.trace` at some point in the future and being annoyed to not find results.

parser/parser/cc/driver.cc
```diff
@@ -14,7 +14,9 @@ typedruby27::typedruby27(std::string_view source, const struct builder &builder)
 
 ForeignPtr typedruby27::parse(SelfPtr self, bool trace) {
     bison::typedruby27::parser p(*this, self);
+#if DEBUG_MODE
     p.set_debug_level(trace ? 1 : 0);
+#endif
```
jez, line 19: What do you think about including `sorbet_version.h` here and using `::sorbet::debug_mode`?

This would be the first place outside of `sorbet_version.{h,c}` that we use the macro variable directly.

froydnj: Using `::sorbet::debug_mode` doesn't work, because `set_debug_level` doesn't exist on `p` in non-debug builds.  (`if constexpr` doesn't save you either, because the behavior that you would want here isn't available outside of templates.)

The other alternative is to compile the grammar once with tracing and once without, then switch between them depending on the trace level.  This requires some build hackery and duplicated code, but at least would make the fast path faster and the test could remain as-is.

froydnj:
I opted for the "compile the bison grammar twice" approach, which is slightly gross, but it at least gives the convenience of the bison trace, all the time, without the performance hit of the tracing in the common case.  Binary size is only increased by ~80k (`--config release-linux`), which is slightly less than 1%.

elliottt (approved):
Seems reasonable to me 👍

parser/parser/munge_grammar.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
```
jez, line 2: can we get `set -euo pipefail` in this script?

parser/parser/BUILD
```diff
@@ -17,12 +17,56 @@ ragel(
...
+        "-Wno-empty-rule",
+        "-Wno-precedence",
+    ],
+)
+
 bison(
-    name = "typedruby_bison",
-    src = "cc/grammars/typedruby.ypp",
+    name = "typedruby_debug_bison",
+    src = "typedruby_debug.ypp",
```
jez, line 65: i think in retrospect i might have called these "typedruby" and "typedruby_with_tracing" because as is, it implies that `_debug` is tied to debug builds and `_release` is tied to release builds

PR #5119, opened 2022-01-15
Add Hummingbird to 'Who is using Sorbet?'

thegorgon:
Adds a new company, Hummingbird, to the list of companies using Sorbet! Also removes Atrium, which sadly no longer exists.

image

### Motivation

Because we love Sorbet! And are active users of it! And because unfortunately, Atrium did not survive.

### Test plan

It's just a data config. See the attached screenshot for verification.


elliottt (approved):
Thanks!

PR #5123, opened 2022-01-15
Change build_compilation_db.sh to use ./bazel instead of system bazel

aprocter:
This modifies `build_compilation_db.sh` to use `./bazel` in one spot where it currently uses system bazel, which is nice if one doesn't have bazel installed systemwide.

### Motivation
Was setting up a new PC and came across this quirk. :)

### Test plan
"Works on my machine!"


froydnj (approved):
:shipit:

jez (approved):

PR #5126, opened 2022-01-17
avoid allocating strings when parsing identifier accessibles

froydnj:
We were asserting that we were receiving UTF8 identifiers in `BuilderImpl::accessible`, but then copying their contents into a `string` to manipulate them.  There's no need to do that: the `string` representation (from `show`) is the same as the `string_view` representation (from `shortName`), so we can use the latter and save copying etc.  We just have to be a little lazier in `checkCircularArgumentReferences`.

### Motivation

Slightly faster code; saves ~0.5% of instructions executed during parsing.

### Test plan

See included automated tests.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -181,7 +181,8 @@ public:
     unique_ptr<Node> accessible(unique_ptr<Node> node) {
         if (auto *id = parser::cast_node<Ident>(node.get())) {
             ENFORCE(id->name.kind() == core::NameKind::UTF8);
-            auto name_str = id->name.show(gs_);
+            // Because of the above enforce, we can use shortName here instead of show.
+            auto name_str = id->name.shortName(gs_);
             if (isNumberedParameterName(name_str) && driver_->lex.context.inDynamicBlock()) {
                 if (driver_->numparam_stack.seen_ordinary_params()) {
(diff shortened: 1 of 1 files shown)


parser/Builder.cc
```diff
@@ -305,10 +306,10 @@ class Builder::Impl {
 
     unique_ptr<Node> assignable(unique_ptr<Node> node) {
         if (auto *id = parser::cast_node<Ident>(node.get())) {
-            auto name_str = id->name.show(gs_);
-            checkReservedForNumberedParameters(name_str, id->loc);
+            auto name = id->name.shortName(gs_);
```
aprocter-stripe, line 309: In `accessible` you mentioned an `ENFORCE(id->name.kind() == core::NameKind::UTF8)` that justifies use of `shortName` instead of `show` here. Do we need to check that here?

aprocter-stripe (approved):
One cargo-cultish question you might want to double check.

parser/Builder.cc
froydnj: It turns out that `Ident` are always constructed with `NameKind::UTF8` `NameRef`s.  I could add some logic to `Ident`'s constructor to verify this at construction time as a follow-up if you like?

aprocter-stripe: That sounds like a good idea if you happen to get a chance.

PR #5127, opened 2022-01-18
Add support for package-private visibility for methods

aadi-stripe:
Adds support for a `package_private` access modifier on method declarations, e.g.:
```
class MyPackage::A
  extend T::Sig

  sig {returns(String)}
  package_private def foo
    '1'
  end
end
```

```
# typed: strict

class OtherPackage::B
  extend T::Sig

  sig {returns(String)}
  def self.bar
    MyPackage::A.new.foo # <-- access violation error
  end
end
```

Assume `MyPackage::A` is exported, and `OtherPackage` imports `MyPackage`. Accesses to `foo` from within `MyPackage` will be allowed, but accesses to it from outside the package will throw the following error:

```
other_package/b.rb:<L>: Non-package-private call to package-private method `foo` on
`MyPackage::A` https://srb.help/7036
     8 |    MyPackage::A.new.foo

    my_package/a.rb:<L>: Defined in `MyPackage::A` here
     7 |  package_private def foo
```

### Motivation

Ideally, we’d have a clean separation in packages between interfaces that need to be exported and interfaces that don’t, and more importantly, this separation would be standardized across the Stripe codebase. In this scenario, `export`  statements are potentially all we would strictly need to enforce the separation guarantees required.

However in the current Stripe codebase, this is not the case. Exporting a constant often may expose behavior to other packages that we do not wish to be visible in certain cases, while wanting to expose similar behavior in other cases. Given this, a `package_private` modifier could solve such problems neatly.

### Test plan

cc @ngroman @jvilk-stripe @aisamanra


aadi-stripe:

→ https://go/builds/bui_KzSpP3d3xMcxTM
→ https://go/builds/bui_KzSpnCX5PEBNuU

froydnj (approved):
We discussed the `Module` monkeypatches and the conclusion we came to is that we're going to keep those changes internal to Stripe's codebase.  If we get a lot of questions about `--stripe-packages`, we can consider moving those into Sorbet proper.

r+ with those changes and the comments below addressed.

infer/environment.cc
```diff
@@ -1010,6 +1010,21 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                         }
                     }
 
+                    if (it->main.method.exists() && it->main.method.data(ctx)->flags.isPackagePrivate) {
+                        core::ClassOrModuleRef klass = core::Types::getClassForAppliedOrClassType(it->main.receiver);
```
jez, line 1014: This implementation is buggy. Here's a failing test case showing why:

https://github.com/sorbet/sorbet/compare/jez-failing-package-private-test#diff-5b422b38a3eb48152c151254cbda3be8a2aff7878eed8228597112bd23fe3712R20-R21

This code assumes that `it->main.receiver` will always be a class type or applied type, but that's not true. It could be many types.

What you really want to look at is `it->main.method.data(ctx)->owner`, which will always exist and always be a `ClassOrModuleRef`.

In general, there's a reason why no method like `getClassForAppliedOrClassType` exists in the codebase: it's a huge crutch encouraging people to conflate types with symbols. If you could also delete it while fixing this, that would be greatly appreciated!

PR #5129, opened 2022-01-18
Make `T.nilable(T.untyped)` an error

elliottt:
Raise an error when `T.nilable(T.untyped)` is found at `typed: false` and above. This doesn't handle cases where the argument is determined to be `T.untyped` indirectly, like in the following example:

```
X = T.type_alias {T.untyped}
Y = T.type_alias {T.nilable(X)}
```

The reason for avoiding this and similar cases is that `T.untyped` is returned in many cases when an error is raised to silence subsequent errors. `T.nilable(<T.untyped-from-type-error>)` is not easily distinguishable from the case where the user intentionally wrote something that reduces to `T.untyped` as the argument to `T.nilable`, so this is a compromise to weed out a common case.

### Motivation

Avoiding the expectation that `T.must` will need to be used on an argument whose written type is `T.nilable(T.untyped)`.

### Test plan

A CLI test was introduced that validates the autocorrects.


elliottt:

→ https://go/builds/bui_KzPKXRqyviXz8R
→ https://go/builds/bui_KzPKKwrbFIPjwZ

froydnj:
Is it possible to produce a different error (and autocorrect)  when `T.nilable(T.untyped)` is used as a prop type?

elliottt:
> Is it possible to produce a different error (and autocorrect) when `T.nilable(T.untyped)` is used as a prop type?

That would be nice. I can see about detecting it in the rewriter, though it wouldn't be as stable as in type_syntax.cc.

elliottt:

→ https://go/builds/bui_KzQT1Ikytpx3wd
→ https://go/builds/bui_KzQTtolV8Myd0Y

elliottt:

→ https://go/builds/bui_KzS6yoDyEEQQYN
→ https://go/builds/bui_KzS62N2v5gPIM5

elliottt:

→ https://go/builds/bui_KzoMtNodWeX0b9
→ https://go/builds/bui_KzoMyp3ibtglbx

test/cli/autocorrect-nilable-untyped/autocorrect-nilable-untyped.sh
```diff
@@ -0,0 +1,7 @@
+tmp="$(mktemp)"
+pre="test/cli/autocorrect-nilable-untyped/pre.rb"
+post="test/cli/autocorrect-nilable-untyped/post.rb"
+cp "$pre" "$tmp"
+main/sorbet --silence-dev-message -a "$tmp"
+diff "$post" "$tmp"
```
jez, line 6: the slightly annoying thing about structuring the CLI test this way is that if this test ever needs to be updated, the bazel `update` test target will write the (new, non-empty) diff file into the `.out` file, instead of changing `post.rb`.

If you structure your autocorrect test like this, it will play better with the update scripts, and still show good `diff` output when the autocorrect output is not what was expected:

https://github.com/sorbet/sorbet/pull/4786/files#diff-0dabd4297161e4f7ab7e711a40cdb181021df3f46a206de93e6f930d1a6733f5R21

jez: my guess is that you're probably just cargo culting something here. if you want to change in this PR, go ahead. otherwise, i'll go ahead and change it after you land, and also change all the other tests that use this pattern.

test/testdata/resolver/nilable_untyped.rb
```diff
@@ -0,0 +1,39 @@
+# typed: false
+#
+# The type_member errors don't get re-raised in the fast-path
+# disable-fast-path: true
+
+
+Bad = T.type_alias {T.nilable(T.untyped)}
+                  # ^^^^^^^^^^^^^^^^^^^^ error: `T.nilable(T.untyped)` is the same as `T.untyped`
```
jez, line 8: These are minor suggestions, but i prefer writing the assertions like this so that my editor doesn't complain about misaligned indents. Non-blocking.

```suggestion
#                   ^^^^^^^^^^^^^^^^^^^^ error: `T.nilable(T.untyped)` is the same as `T.untyped`
```

jez (approved):
Thank you for the thorough test

test/cli/autocorrect-nilable-untyped/autocorrect-nilable-untyped.sh
elliottt: Yep, just cargo-culted existing autocorrect cli tests. I'm happy to make the output the diff instead.

test/testdata/resolver/nilable_untyped.rb
elliottt: Sure, happy to change them to trailing assertions.

elliottt:

→ https://go/builds/bui_L1g3d50okxnyXw
→ https://go/builds/bui_L1g36hnKd3B67b

test/testdata/resolver/nilable_untyped.rb
```diff
@@ -0,0 +1,39 @@
...
+
+
+Bad = T.type_alias {T.nilable(T.untyped)}
+#                   ^^^^^^^^^^^^^^^^^^^^ error: `T.nilable(T.untyped)` is the same as `T.untyped`
+
+# This is a tough to handle case, as the argument to `T.nilable` in `Tricky` is
+# still `T.untyped`, but it's hard to distinguish that from a case where
+# `T.untyped` was returned because the argument type failed to parse in
+# type_syntax.cc. The compromise here is to only raise the error when we
+# encounter `T.nilable(T.untyped)` explicitly foo.
```
froydnj, line 14:
```suggestion
# encounter `T.nilable(T.untyped)` explicitly.
```

froydnj (approved):

test/testdata/resolver/nilable_untyped.rb
elliottt: Thanks for catching this!

ashkulz:
This causes a lot of errors for projects which use [sorbet-rails](https://github.com/chanzuckerberg/sorbet-rails), I'll try to create a PR there in the next few days.

PR #5130, opened 2022-01-19
Packages: enforce package naming in subpackages

ngroman:
Previously EnforcePrefix only looked at the current file's package and had no knowledge of other packages present in the codebase. This meant that with nested packages `Foo` and `Foo::Bar` it would allow a definitions in `Foo` that should only be allowed in the sub-package.

Naively scanning against all packages would be a performance hit. However, by sorting packages lexicographically (by `NameRef::rawID()` to avoid string comparisons) we can represent the tree compactly. Walking the tree happens with `std::{lower,upper}_bound` calls that are `O(log(#packages))`.

This PR switches enforcement to use this and changes the O(exports^2) check to use this as well.

### Motivation
Sorbet would previously allow a file in `PkgA` to define a constant that should be in `PkgA::SubPkg`.

### Test plan
* Added test cases
* Tested on Stripe's codebase
* Measured performance, difference was undetectable vs previous implementation

See included automated tests.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -605,6 +605,8 @@ public:
                 foundTestNS = true;
                 return;
             } else if (!isTestNamespace(ctx, name)) {
+                // Inside a test file, but not inside a test namespace. Set bounds such that
+                // begin == end, stopping any subsequent search.
                 bounds.emplace_back(begin, end);
                 nameParts.emplace_back(name);
                 begin = end = 0;
@@ -613,6 +615,8 @@ public:
         }
 
         if (idx > 0 && end - begin == 1 && packages[begin] == filePkg.mangledName()) {
+            // We have descended into a package with no sub-packages. At this point it is safe to
+            // skip tracking of deeper constants.
             curPkg.emplace_back(packages[begin], SKIP_BOUND_VAL);
             skips++;
             return;
@@ -636,7 +640,7 @@ public:
         if (begin != end) {
             auto &pkgInfo = ctx.state.packageDB().getPackageInfo(*lb);
             ENFORCE(pkgInfo.exists());
-            if (bounds.size() == pkgInfo.fullName().size()) {
+            if (nameParts.size() == pkgInfo.fullName().size()) {
                 curPkg.emplace_back(*lb, bounds.size());
             }
         }
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -156,7 +156,45 @@ struct Export {
     }
 };
 
-class PackageInfoImpl final : public sorbet::core::packages::PackageInfo {
+// For a given vector of NameRefs, this represents the "next" vector that does not begin with its
+// prefix (without actually constructing it). Consider the following sorted names:
+//
+// [A B]
+// [A B C]
+// [A B D E]
+//    <<<< Position of LexNext([A B]) roughly equivalent to [A B <Infinity>]
+// [X Y]
+// [X Y Z]
+class LexNext final {
+    const vector<core::NameRef> &names;
+
+public:
+    LexNext(const vector<core::NameRef> &names) : names(names) {}
+
+    bool operator<(const vector<core::NameRef> &rhs) const {
+        auto lhsIt = names.begin();
+        auto rhsIt = rhs.begin();
+        // Lexicographic comparison:
+        for (; lhsIt <= names.end() && rhsIt != rhs.end(); ++lhsIt, ++rhsIt) {
+            // Treat this.names.end() as if it were the max value.
+            if (lhsIt == names.end()) {
+                return false;
+            }
+            if (lhsIt->rawId() < rhsIt->rawId()) {
+                return true;
+            } else if (rhsIt->rawId() < lhsIt->rawId()) {
+                return false;
+            }
+        }
+        return (lhsIt == names.end()) && (rhsIt != rhs.end());
+    }
+
+    bool operator<(const Export &e) const {
+        return *this < e.parts();
+    }
+};
+
+class PackageInfoImpl final : public core::packages::PackageInfo {
 public:
     core::NameRef mangledName() const {
         return name.mangledName;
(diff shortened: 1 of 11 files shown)


ngroman:
cc @aisamanra @aadi-stripe

nroman-stripe:

→ https://go/builds/bui_KzRc0hSsbCUb4n
→ https://go/builds/bui_KzRcg7jZjI4mxd

ngroman:
Note the one failure in https://go/builds/bui_KzRcg7jZjI4mxd is legitimate and the file needs to be moved in Stripe's codebase.

packager/packager.cc
```diff
@@ -479,34 +517,196 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+                return;
+            } else if (!isTestNamespace(ctx, name)) {
+                bounds.emplace_back(begin, end);
+                nameParts.emplace_back(name);
+                begin = end = 0;
+                return;
+            }
+        }
+
+        if (idx > 0 && end - begin == 1 && packages[begin] == filePkg.mangledName()) {
```
aadi-stripe, line 615: Recommend adding a comment here saying we skip because the correct package prefix has been determined, just for clarity.

packager/packager.cc
```diff
@@ -479,34 +517,196 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+
+        if (isTestFile && idx == 0 && !foundTestNS) {
+            if (isPrimaryTestNamespace(name)) {
+                foundTestNS = true;
+                return;
+            } else if (!isTestNamespace(ctx, name)) {
+                bounds.emplace_back(begin, end);
+                nameParts.emplace_back(name);
+                begin = end = 0;
+                return;
```
aadi-stripe, line 611: Recommend adding a comment here with an example -- mentioning we're exiting early because a test file has a non-test namespace.

packager/packager.cc
```diff
@@ -479,34 +517,196 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+                return next < ctx.state.packageDB().getPackageInfo(pkgNr).fullName();
+            });
+
+        begin = lb - packages.begin();
+        end = ub - packages.begin();
+
+        if (begin != end) {
+            auto &pkgInfo = ctx.state.packageDB().getPackageInfo(*lb);
+            ENFORCE(pkgInfo.exists());
+            if (bounds.size() == pkgInfo.fullName().size()) {
```
aadi-stripe, line 639: Why not compare against `nameParts.size()`?

aadi-stripe (changes requested):

packager/packager.cc
ngroman: Fixed. Remnant of older code that left nameParts in the tree walk.

packager/packager.cc
```diff
@@ -156,7 +156,45 @@ struct Export {
...
+// prefix (without actually constructing it). Consider the following sorted names:
+//
+// [A B]
+// [A B C]
+// [A B D E]
+//    <<<< Position of LexNext([A B]) roughly equivalent to [A B <Infinity>]
+// [X Y]
+// [X Y Z]
+class LexNext final {
+    const vector<core::NameRef> &names;
```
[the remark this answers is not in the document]
ngroman: Going to leave this as is. Given this is the only member, I don't think there's a whole lot of chance for confusion.

aadi-stripe (approved):

aadi-stripe:
lgtm, thanks for making the requested changes!

packager/packager.cc
```diff
@@ -479,34 +517,200 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+                // Inside a test file, but not inside a test namespace. Set bounds such that
+                // begin == end, stopping any subsequent search.
+                bounds.emplace_back(begin, end);
+                nameParts.emplace_back(name);
+                begin = end = 0;
+                return;
+            }
+        }
+
+        if (idx > 0 && end - begin == 1 && packages[begin] == filePkg.mangledName()) {
```
elliottt, line 617: `idx` was only used to check if bounds was empty.
```suggestion
        uint16_t boundsEmpty = bounds.empty();

        if (isTestFile && boundsEmpty && !foundTestNS) {
            if (isPrimaryTestNamespace(name)) {
                foundTestNS = true;
                return;
            } else if (!isTestNamespace(ctx, name)) {
                // Inside a test file, but not inside a test namespace. Set bounds such that
                // begin == end, stopping any subsequent search.
                bounds.emplace_back(begin, end);
                nameParts.emplace_back(name);
                begin = end = 0;
                return;
            }
        }

        if (!boundsEmpty && end - begin == 1 && packages[begin] == filePkg.mangledName()) {
```

packager/packager.cc
```diff
@@ -479,34 +517,200 @@ ast::UnresolvedConstantLit *verifyConstant(core::MutableContext ctx, core::NameR
...
+    const vector<core::NameRef> &packages; // Mangled names sorted lexicographically
+    const PackageInfoImpl &filePkg;        // Package for current file
+    // Current bounds:
+    uint16_t begin;
+    uint16_t end;
+
+    const bool isTestFile;
+    const uint16_t filePkgIdx;
+
+    int skips = 0;
```
elliottt, line 556: Could you add a comment explaining what `skips` is used for?

elliottt (approved):
I think this make sense. Just had a few suggestions.

packager/packager.cc
```diff
@@ -156,7 +156,45 @@ struct Export {
...
+
+    bool operator<(const vector<core::NameRef> &rhs) const {
+        auto lhsIt = names.begin();
+        auto rhsIt = rhs.begin();
+        // Lexicographic comparison:
+        for (; lhsIt <= names.end() && rhsIt != rhs.end(); ++lhsIt, ++rhsIt) {
+            // Treat this.names.end() as if it were the max value.
+            if (lhsIt == names.end()) {
+                return false;
+            }
```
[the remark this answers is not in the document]
ngroman: I ended up simplifying this

PR #5133, opened 2022-01-19
Remove "mutable" from LSPTypechecker GlobalState pointer

aprocter:
This removes the `mutable` qualifier from the `GlobalState` pointer in `LSPTypechecker`. In order to do this, we have to change `GlobalState::replaceFile` so that it doesn't "borrow" a unique_ptr to the GlobalState.

### Motivation
Just seems nice to drop `mutable` in cases where we don't actually need it.

### Test plan
Existing CI.

- [x] Some manual smoke testing against Stripe codebase with LSP

### DIFF
--- a/main/lsp/LSPTypechecker.h
+++ b/main/lsp/LSPTypechecker.h
@@ -34,9 +34,8 @@ class UndoState;
 class LSPTypechecker final {
     /** Contains the ID of the thread responsible for typechecking. */
     std::thread::id typecheckerThreadId;
-    /** GlobalState used for typechecking. Mutable because typechecking routines, even when not changing the GlobalState
-     * instance, actively consume and replace GlobalState. */
-    mutable std::unique_ptr<core::GlobalState> gs;
+    /** GlobalState used for typechecking. */
+    std::unique_ptr<core::GlobalState> gs;
     /** Trees that have been indexed (with initialGS) and can be reused between different runs */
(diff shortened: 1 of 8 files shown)


main/lsp/LSPTypechecker.h
```diff
@@ -34,9 +34,8 @@ class UndoState;
 class LSPTypechecker final {
     /** Contains the ID of the thread responsible for typechecking. */
     std::thread::id typecheckerThreadId;
-    /** GlobalState used for typechecking. Mutable because typechecking routines, even when not changing the GlobalState
-     * instance, actively consume and replace GlobalState. */
-    mutable std::unique_ptr<core::GlobalState> gs;
```
froydnj, line 39: This change is one of those times that I wish `this->` prefixing was pervasive, so it was easier to see at a glance where members were getting mutated versus parameters.

froydnj (approved):
Makes sense to me!  Have you tested this on Stripe's codebase with vscode at all, just to make sure LSP isn't going to blow up somehow in real-world usage?

aprocter:
> Have you tested this on Stripe's codebase with vscode at all, just to make sure LSP isn't going to blow up somehow in real-world usage?

Nope! Will be doing that before merge.

aprocter-stripe:

→ https://go/builds/bui_L06oXDIPJWGrSr
→ https://go/builds/bui_L06oqTwTwPs6gN

PR #5138, opened 2022-01-20
Allow props with `default: nil` to be considered nilable

elliottt:
Modify the prop dsl's internal tracking of nilable fields to include those with `default: nil`. This change allows us to give the prop/const being defined a type of `T.untyped`, while still keeping the fact that it's expected to be nilable. Previously writing `T.nilable(T.untyped)` worked somewhat by accident, as the underlying `T.any` type didn't automatically collapse to `T.untyped`.

### Motivation

Enabling us to make `T.nilable(T.untyped)` an error, merging in #5129.

### Test plan

Successful run on Stripe CI (modulo generated file changes): https://go/builds/bui_L0AK0aHJsuoooh

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -290,7 +290,7 @@ class T::Props::Decorator
   def prop_defined(name, cls, rules={})
     cls = T::Utils.resolve_alias(cls)
 
-    if T::Utils::Nilable.is_union_with_nilclass(cls)
+    if T::Utils::Nilable.is_union_with_nilclass(cls) || (rules.key?(:default) && rules[:default].nil?)
       # :_tnilable is introduced internally for performance purpose so that clients do not need to call
       # T::Utils::Nilable.is_tnilable(cls) again.
       # It is strictly internal: clients should always use T::Props::Utils.required_prop?() or
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -277,8 +277,10 @@ class T::Props::Decorator
     end
   end
 
-  # Returns `true` when the type of the prop is nilable, or a `:default` is
-  # present in the rules hash, and its value is `nil`.
+  # Returns `true` when the type of the prop is nilable, or the field is typed
+  # as `T.untyped`, a `:default` is present in the rules hash, and its value is
+  # `nil`. The latter case is a workaround for explicitly not supporting the use
+  # of `T.nilable(T.untyped)`.
   #
   # checked(:never) - Rules hash is expensive to check
   sig do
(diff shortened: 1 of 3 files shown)


gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -290,7 +290,7 @@ def prop_validate_definition!(name, cls, rules, type)
   def prop_defined(name, cls, rules={})
     cls = T::Utils.resolve_alias(cls)
 
-    if T::Utils::Nilable.is_union_with_nilclass(cls)
+    if T::Utils::Nilable.is_union_with_nilclass(cls) || (rules.key?(:default) && rules[:default].nil?)
```
vasi-stripe, line 293: Nit: Should we turn this into a `prop_nilable?` method?

vasi-stripe (approved):
LGTM. One optional style nit.

gems/sorbet-runtime/lib/types/props/decorator.rb
elliottt: Sure, should it be private in this file?

vasi-stripe: Sounds reasonable!

vasi-stripe (approved):

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -277,6 +277,22 @@ def prop_validate_definition!(name, cls, rules, type)
...
+    params(
+      cls: PropTypeOrClass,
+      rules: Rules,
+    )
+    .void
+    .checked(:never)
+  end
+  private def prop_nilable?(cls, rules)
+    T::Utils::Nilable.is_union_with_nilclass(cls) || (rules.key?(:default) && rules[:default].nil?)
+  end
```
froydnj, line 294: Do we have tests for something like `prop :thing, Integer, default: nil`?  (I guess we'd want both Sorbet checks, if possible, and sorbet-runtime checks?)  The second part of the `||` is really only relevant for untyped `cls`, right?

froydnj (changes requested):

gems/sorbet-runtime/test/types/props/constructor.rb
```diff
@@ -53,4 +58,8 @@ class Outer < T::Struct
     o = Outer.new(objs: [Inner.new(i: 4), Inner.new(i: 5)])
     assert_equal([4, 5], o.objs.map(&:i))
   end
+
+  it 'can default untyped fields' do
+    UntypedField.new
+  end
```
froydnj, line 64:
```suggestion

  it 'can default untyped fields' do
    UntypedField.new
  end
  
  it 'indicates untyped default nil fields are optional'
    assert(T::Props::Utils.optional_prop?(UntypedField.props[:untyped_const]))
    assert(T::Props::Utils.optional_prop?(UntypedField.props[:untyped_prop]))
  end
```

I don't know whether this is the best place to put this test, but we at least need something like this.

froydnj (approved):
r=me with the new test below (or something like it) and a test to check `required_prop?` (or `optional_prop?`) for a prop with not-`T.untyped`-type and `default: nil`.  (This PR should not change behavior for the latter, I believe?)

PR #5140, opened 2022-01-20
change tokens to store `string_view`

froydnj:
This completes the vision from #5094: we now let tokens store only a `string_view` of the actual token, which shrinks the size of the token (`string_view` is half the size of `string` with libstdc++).  This shrinking sadly doesn't matter that much; what does matter is that we're not constantly copying data into the `string`s associated with tokens.  (If we could figure out some way of not holding on to every single token we've ever lexed during parsing, that would be nice too.)

By and large, this view will point into the text being parsed.  For cases where that's not true (e.g. escaped strings), we maintain a scratch buffer for copying strings into and point the views into that.  (Tokens that are emitted as static character literals are a special case: we don't need to copy those into the scratch buffer.)  Fortunately, `GlobalState` already does something very similar, so we refactored some code from there and made the debugging process a lot easier.

### Motivation

Speed.  Doing this saves ~2% of instructions during parsing.

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/cc/token.cc
+++ b/parser/parser/cc/token.cc
@@ -26,6 +26,6 @@ std::string_view token::view() const {
     return _string;
 }
 
-const std::string &token::asString() const {
-    return _string;
+std::string token::asString() const {
+    return std::string(_string);
 }
--- a/parser/parser/include/ruby_parser/token.hh
+++ b/parser/parser/include/ruby_parser/token.hh
@@ -177,6 +177,10 @@ class token {
 
 public:
     token(token_type type, size_t start, size_t end, std::string_view str);
+    // Don't allow people to rely on implicit conversion to std::string_view:
+    // they should make sure their string views live in storage that will outlive
+    // the lexer.
+    token(token_type type, size_t start, size_t end, const std::string &str) = delete;
 
     token_type type() const;
     size_t start() const;
--- a/parser/parser/include/ruby_parser/token.hh
+++ b/parser/parser/include/ruby_parser/token.hh
@@ -173,7 +173,7 @@ class token {
     token_type _type;
     size_t _start;
     size_t _end;
-    std::string _string;
+    std::string_view _string;
 
 public:
     token(token_type type, size_t start, size_t end, std::string_view str);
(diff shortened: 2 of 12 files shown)


parser/parser/cc/token.cc
```diff
@@ -15,13 +15,13 @@ size_t token::start() const {
...
 
 std::string_view token::view() const {
     return _string;
 }
 
-const std::string &token::asString() const {
-    return _string;
+std::string token::asString() const {
+    return std::string(_string);
 }
```
aprocter, line 27: Do we know how often this gets called? From the PR description it sounds like you've already done the homework on bottom line perf, so maybe not any cause for real concern, just noting that we are copying here where we didn't used to.

parser/parser/include/ruby_parser/token.hh
```diff
@@ -167,19 +167,31 @@ enum class token_type : int {
...
 class token {
-    token_type _type;
     size_t _start;
-    size_t _end;
-    std::string _string;
+    token_type _type;
+    uint32_t _size;
+    std::string_view _string;
+
+    uint32_t clamp_size(size_t size) {
```
aprocter, line 175:
```suggestion
    static uint32_t clamp_size(size_t size) {
```

aprocter (approved):
:shipit:

parser/parser/cc/token.cc
froydnj: Yeah, some previous PRs helped eliminate the number of places where we would call into this function.

If you look through `parser/Builder.cc` for `asString()`, you can see that we don't call it in that many places.  I think it would help to eliminate it for `BuilderImpl::arg` and maybe one or two other places, but I'm not super-worried about it.

froydnj:

→ https://go/builds/bui_L37hBbhW1WvKvb
→ https://go/builds/bui_L37hQ8lGroZYTx

froydnj:

→ https://go/builds/bui_L38CYWu8dAidNE
→ https://go/builds/bui_L38CDV7yUg8J36

PR #5141, opened 2022-01-20
Allow value omission in hash literals

KaanOzkan:
Part of Ruby 3.1 support

`{x:, y:}` is a now equivalent to `{x: x, y: y}`

https://github.com/ruby/ruby/commit/c60dbcd
https://github.com/whitequark/parser/commit/22e6ca5c0b27bc377fd50d16698edabfad4f4970

Local variable support will be done in another PR as it's slightly complex

### Motivation

### Test plan

See included automated tests.

@Morriar


jez (approved):
So sounds like this example

```
x = 1
puts {x:}
```

will not work correctly yet, and support for that is coming in a future PR?

Thanks

PR #5142, opened 2022-01-21
[RFC] Do not error on `**kwargs` parameters name mismatch

Morriar:
### Motivation

When Sorbet is checking methods redefinitions, it errors on mismatching names for keyword arguments and `**kwargs`:

```rb
# typed: true
extend T::Sig

sig { params(arg: T.untyped).void}
def m1(arg); end

sig { params(argX: T.untyped).void}
def m1(argX); end # ok

sig { params(arg: T.untyped).void}
def m2(*arg); end

sig { params(argX: T.untyped).void}
def m2(*argX); end # ok

sig { params(block: T.untyped).void}
def m3(&block); end

sig { params(blockX: T.untyped).void}
def m3(&blockX); end # ok

sig { params(kwarg: T.untyped).void}
def m4(kwarg:); end

sig { params(kwarg1: T.untyped).void}
def m4(kwarg1:); end # error: Method `Object#m4` redefined with mismatched keyword argument name. Expected: `kwarg`, got: `kwarg1`

sig { params(kwargs: T.untyped).void}
def m5(**kwargs); end

sig { params(kwargs1: T.untyped).void}
def m5(**kwargs1); end # error: Method `Object#m5` redefined with mismatched keyword argument name. Expected: `kwargs`, got: `kwargs1`
```

While this verification makes sense for keyword arguments (they should be the same in all the definitions of the same methods), I'm not sure it applies to `**kwargs`.

Consider the following case:

```rb
# sorbet/rbi/gems/foo.rbi (generated)

class View
  sig { params(kwargs: T.untyped).void }
  def render(**kwargs); end
end
```

```rb
# sorbet/rbi/shims/foo.rbi (manually written shim)

class View
  sig { params(options: T::Hash[Symbol, String]).void }
  def render(**options); end
end
```

The renaming in the shim file should be allowed as it still is compatible with the generated definition for `View#render`.

### Implementation

Added a possible implementation for this change to see what else would break.

### Test plan

See included automated tests.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -1085,7 +1085,7 @@ public:
         core::NameRef nm;
 
         if (name != nullptr) {
-            loc = loc.join(tokLoc(name));
+            loc = tokLoc(name);
             nm = gs_.enterNameUTF8(name->view());
             checkReservedForNumberedParameters(name->view(), loc);
         } else {
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -1081,7 +1081,7 @@ public:
     }
 
     unique_ptr<Node> kwrestarg(const token *dstar, const token *name) {
-        core::LocOffsets loc = tokLoc(dstar);
+        core::LocOffsets loc;
         core::NameRef nm;
 
         if (name != nullptr) {
@@ -1089,6 +1089,7 @@ public:
             nm = gs_.enterNameUTF8(name->view());
             checkReservedForNumberedParameters(name->view(), loc);
         } else {
+            loc = tokLoc(dstar);
             nm = gs_.freshNameUnique(core::UniqueNameKind::Parser, core::Names::starStar(), ++uniqueCounter_);
         }
 
--- a/ast/ArgParsing.cc
+++ b/ast/ArgParsing.cc
@@ -77,7 +77,12 @@ std::vector<uint32_t> ArgParsing::hashArgs(core::Context ctx, const std::vector<
         uint32_t arg = 0;
         uint8_t flags = 0;
         if (e.flags.isKeyword) {
-            arg = core::mix(arg, core::_hash(e.local._name.shortName(ctx)));
+            if (e.flags.isRepeated && e.local._name != core::Names::fwdKwargs()) {
+                auto name = core::Names::kwargs();
+                arg = core::mix(arg, core::_hash(name.shortName(ctx)));
+            } else {
+                arg = core::mix(arg, core::_hash(e.local._name.shortName(ctx)));
+            }
             flags += 1;
         }
         if (e.flags.isRepeated) {
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -1577,10 +1577,9 @@ string ArgInfo::toString(const GlobalState &gs) const {
 }
 
 string_view ArgInfo::argumentName(const GlobalState &gs) const {
-    if (flags.isKeyword) {
+    if (flags.isKeyword && !flags.isRepeated) {
         return name.shortName(gs);
     } else {
-        // positional arg
         if (auto source = loc.source(gs)) {
             return source.value();
         } else {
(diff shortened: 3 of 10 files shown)


jez:
At first I misread this PR, thinking that you were attempting to make

```ruby
def foo(x:, y:); end
def foo(**kwargs); end
```

not report an error. That doesn't seem to be the case, and I'm generally on board with this idea now knowing what it really does.

There's a problem with your implementation:

```
❯ sorbet -p symbol-table-raw foo.rb
class <C <U <root>>> < <C <U Object>> ()
  class <S <C <U <root>>> $1>[<C <U <AttachedClass>>>] < <S <C <U Object>> $1> ()
    method <S <C <U <root>>> $1>#<N <U <static-init>> $152> (<blk>) @ Loc {file=foo.rb start=3:1 end=6:4}
      argument <blk><block> @ Loc {file=foo.rb start=??? end=???}
  class <C <U A>> < <C <U Object>> () @ Loc {file=foo.rb start=3:1 end=3:8}
    method <C <U A>>#<M <U foo> $1> (args, <blk>) @ Loc {file=foo.rb start=4:3 end=4:18}
      argument args<keyword, repeated> @ Loc {file=foo.rb start=4:11 end=4:17}
      argument <blk><block> @ Loc {file=foo.rb start=??? end=???}
    method <C <U A>>#<U foo> (argv, <blk>) @ Loc {file=foo.rb start=5:3 end=5:18}
      argument argv<keyword, repeated> @ Loc {file=foo.rb start=5:11 end=5:17}
      argument <blk><block> @ Loc {file=foo.rb start=??? end=???}
```

You're just papering over the symptoms, but not actually addressing the root cause.

What we see in this symbol-table output is that there are two `foo` methods defined on `A`:

- `method <C <U A>>#<M <U foo> $1>`
- `method <C <U A>>#<U foo>`

The `<M ...>`-style name is a mangle-rename name kind. That happens any time that `mangleRenameSymbol` is called. Despite not reporting an error, we're still mangling the old symbol out of the way so that we can re-enter a new method, with different argument names.

We have solved a similar problem in the past for positional args: you'll notice that

```ruby
def foo(x); end
def foo(y); end
```

reports no error. The way that this is done is that all positional args are given a name of the form `arg$<n>` where `<n>` is the index of the positional arg:

https://github.com/sorbet/sorbet/blob/3cd1ae14c8ab48d79236fe06d347eccff03b4e2d/namer/namer.cc#L851

The original argument name is not actually stored in in the symbol table—the only time it's ever needed is for error messages, and for that, we use a clever trick: we record the `Loc` for the arg, and define `ArgInfo::argumentName` to read the source file under that `Loc` when printing argument names:

https://github.com/sorbet/sorbet/blob/3cd1ae14c8ab48d79236fe06d347eccff03b4e2d/core/Symbols.cc#L1571-L1573

We should do something similar with `**kwargs`—canonicalize their name in the symbol table to something like `<kwargs>` (with a well-known name like `core::Names::kwargs()`), and use the `Loc` on the `ArgInfo` when printing the name. That way, the code path that you're modifying here wouldn't even get hit.

jez (changes requested):
generally on board.

parser/Builder.cc
```diff
@@ -1085,7 +1085,7 @@ class Builder::Impl {
         core::NameRef nm;
 
         if (name != nullptr) {
-            loc = loc.join(tokLoc(name));
+            loc = tokLoc(name);
```
jez, line 1088: This is a minor thing, but instead of overwriting `loc` in just this case, do you mind changing it so that `loc` is not given an initial value, and instead it's written to once in each of the `if` branches?

jez (approved):
beautiful, thanks!

jez:

→ https://go/builds/bui_LKnzZEzNmX3HKm
→ https://go/builds/bui_LKnzzMKJ5GjIhf

parser/Builder.cc
jez: i went ahead and did this. it'll merge when the tests pass.

PR #5149, opened 2022-01-25
Fix package-private enforcement for intersection types

aadi-stripe:
In #5127 , I added support for declaring methods package-private. However, there was a bug in my code -- it did not handle intersection types correctly, as it only looked at `AppliedType` or `ClassType` when checking the type of the receiver.

There is a simpler way to find the owning class ref, by looking method owner of the method, and this will handle all types. Making the change in order to fix the bug.

### Motivation

Package-private enforcement.

### Test plan

See included automated tests.

cc @jez

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1014,7 +1014,7 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                         core::ClassOrModuleRef klass = it->main.method.data(ctx)->owner;
                         if (klass.exists()) {
                             const auto &curPkg = ctx.state.packageDB().getPackageForFile(ctx, ctx.file);
-                            if (curPkg.mangledName() != core::NameRef::noName() && !curPkg.ownsSymbol(ctx, klass)) {
+                            if (curPkg.exists() && !curPkg.ownsSymbol(ctx, klass)) {
                                 if (auto e = ctx.beginError(bind.loc, core::errors::Infer::PackagePrivateMethod)) {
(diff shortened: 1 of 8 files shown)


infer/environment.cc
```diff
@@ -1011,13 +1012,18 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
...
+                        core::ClassOrModuleRef klass = it->main.method.data(ctx)->owner;
                         if (klass.exists()) {
                             const auto &curPkg = ctx.state.packageDB().getPackageForFile(ctx, ctx.file);
-                            if (curPkg.mangledName() != core::NameRef::noName() && !curPkg.ownsSymbol(ctx, klass)) {
+                            if (curPkg.exists() && !curPkg.ownsSymbol(ctx, klass)) {
                                 if (auto e = ctx.beginError(bind.loc, core::errors::Infer::PackagePrivateMethod)) {
-                                    e.setHeader("Non-package-private call to package-private method `{}` on `{}`",
-                                                it->main.method.data(ctx)->name.show(ctx), it->main.receiver.show(ctx));
+                                    auto &curPkgName = curPkg.fullName();
+                                    // TODO (aadi-stripe, add name of owning package to message).
```
aadi-stripe, line 1021: I'll be adding this later. Right now going from symbol -> owning package is a bit clunky as we have to manipulate the mangled name, but we can fix that later. For now I feel that this error message is descriptive enough for the Stripe codebase.

jez (approved):

jez:
Thanks!

PR #5154, opened 2022-01-25
Parse error recovery and completion for `A::`

jez:
### Motivation

Better completion.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -971,7 +971,7 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerDelegate &t
             // which parses as a method call (A.puts()) instead of a constant.
 
             auto scopes = core::lsp::ConstantResponse::Scopes{};
-            // We're ignoring the other dispatch components here, because the fast that we saw `::`
+            // We're ignoring the other dispatch components here, because the fact that we saw `::`
             // in the source means that it's likely a constant lit, not something with a fancy type.
             // Also since it's a constant lit, it probably has type `T.class_of(...)`, so we want
             // the represented type for the sake of constant lookup.
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -964,24 +964,49 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerDelegate &t
         auto prefix = (callerSideName == core::Names::methodNameMissing() || !sendResp->funLoc.contains(queryLoc))
                           ? ""
                           : callerSideName.shortName(gs);
-        config.logger->debug("Looking for method similar to '{}'", prefix);
+        if (prefix == "" && queryLoc.adjust(gs, -2, 0).source(gs) == "::") {
+            // Probably a case like this:
+            //   A::|
+            //   puts
+            // which parses as a method call (A.puts()) instead of a constant.
 
-        // isPrivateOk means that there is no syntactic receiver. This check prevents completing `x.de` to `x.def`
-        // (If there is a method whose name overlaps with a keyword, it will still show up as a _method_ item.)
-        auto suggestKeywords = sendResp->isPrivateOk;
-        auto params = SearchParams{
-            queryLoc, prefix,
-            MethodSearchParams{
-                sendResp->dispatchResult,
-                sendResp->totalArgs,
-                sendResp->isPrivateOk,
-            },
-            suggestKeywords,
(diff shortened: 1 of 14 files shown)


parser/Builder.cc
```diff
@@ -28,6 +28,8 @@ using std::type_info;
 using std::unique_ptr;
 using std::vector;
 
+extern const char *dclassStrings[];
+
```
froydnj, line 32: Why is this needed?

main/lsp/requests/completion.cc
```diff
@@ -964,24 +964,49 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerDelegate &t
...
+                nullopt,           // do not suggest methods
+                false,             // do not suggest keywords
+                core::MethodRef{}, // do not suggest locals
+                scopes,
+            };
+            items = this->getCompletionItems(typechecker, params);
+        } else {
+            // isPrivateOk means that there is no syntactic receiver. This check prevents completing `x.de` to `x.def`
+            // (If there is a method whose name overlaps with a keyword, it will still show up as a _method_ item.)
+            auto suggestKeywords = sendResp->isPrivateOk;
```
froydnj, line 995: I think the rationale in the comment or the logic for `suggestKeywords` needs to be adjusted:

https://github.com/sorbet/sorbet/blob/10481e1af8df9913da6e57e5960fee5c736cbf2e/ast/desugar/Desugar.cc#L645-L648

since we don't want `self.de` to complete with keywords.

Also `super` can be `isPrivateOk`, so you may want to check for that specifically too?

main/lsp/requests/completion.cc
```diff
@@ -964,24 +964,49 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerDelegate &t
...
+            auto params = SearchParams{
+                queryLoc, prefix,
+                MethodSearchParams{
+                    sendResp->dispatchResult,
+                    sendResp->totalArgs,
+                    sendResp->isPrivateOk,
+                },
+                suggestKeywords,
+                // No receiver means that local variables are allowed here.
+                sendResp->isPrivateOk ? sendResp->enclosingMethod : core::MethodRef{},
```
froydnj, line 1005: Same comment applies to this use of `isPrivateOk`, I think?

froydnj (changes requested):
It looks like there are some test failures.  The idea seems reasonable; a few comments below.

parser/Builder.cc
jez: good catch, this was the result of a bad `git add -p`, it's not needed.
```suggestion
```

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1715,6 +1715,13 @@ int maybe_inject_tNL_impl(parser::symbol_type &yyla, ruby_parser::typedruby27 &d
                     {
                       $$ = driver.build.constFetch(self, $1, $2, $3);
                     }
+                | primary_value tCOLON2 error
+                    {
+                      $$ = driver.build.constFetchError(self, $1, $2);
+                      maybe_inject_tNL(@2.end, @3.begin);
+                      driver.diagnostics.pop_back();
+                      driver.diagnostics.emplace_back(dlevel::ERROR, dclass::ConstWithoutName, diagnostic::range(@2.begin, @2.end));
```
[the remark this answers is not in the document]
jez: sure i can factor a helper. do you mind if i do it in another PR, once everything lands?

main/lsp/requests/completion.cc
jez: Thanks. The first one is an existing bug. I'll open a ticket for it, but I don't want to block this PR on fixing it.

https://github.com/sorbet/sorbet/issues/5157

The `super` question is actually subtle, because it gets translated to `<call-with-block>(<self>, :<super>, <blk>)` which then suggests no completion results. So it does it in a roundabout way but it works.

main/lsp/requests/completion.cc
jez: https://github.com/sorbet/sorbet/issues/5158

parser/parser/cc/grammars/typedruby.ypp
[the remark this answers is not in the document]
jez: i've made a ticket for this internally

froydnj (approved):
Thanks for filing the follow-on tickets.  This looks good, but I think there are still some test failures?

PR #5162, opened 2022-01-25
Add error case for `if ... do; end`

jez:
### Motivation

This is a common typo people make. Doesn't make sense to delete the whole `if`
when a better outcome is to suggest that they delete the `do` keyword or change
the `if` to `it`.

### Test plan

tests show before/after


aprocter (approved):
Nice!

PR #5163, opened 2022-01-25
Fix unadded test files for package enforcement

ngroman:
I forgot to `git add` some test files I used while developing https://github.com/sorbet/sorbet/pull/5130. These should be checked in.

### Test plan

See included automated tests.


jez (approved):
lol

PR #5164, opened 2022-01-26
Fix call to pipeline::index in fuzz_dash_e

aprocter:
Fixes a tiny bit of bitrot in `fuzz_dash_e.cc`. (The signature of `pipeline::index` changed at some point in the past, and since we don't run `fuzz_dash_e` in CI I guess we didn't catch it.)

### Motivation
Make the fuzzer work.

### Test plan
Ran `tools/scripts/fuzz.sh fuzz_dash_e` and made sure the fuzzer built and ran correctly after the change.


elliottt (approved):
:shipit:

PR #5165, opened 2022-01-26
Delete chunder from debug logs

jez:
### Motivation

Looking to make these logs more useful.

Review by commit

### Test plan

See included automated tests.

### DIFF
--- a/cfg/CFG.cc
+++ b/cfg/CFG.cc
@@ -90,7 +90,7 @@ CFG::CFG() {
 }
 
 CFG::ReadsAndWrites CFG::findAllReadsAndWrites(core::Context ctx) {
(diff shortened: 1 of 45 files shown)


jez:

→ https://go/builds/bui_L23idifjvWurtJ
→ https://go/builds/bui_L23icWKGBOadAB

froydnj (changes requested):
Looks like this doesn't build?

This generally seems OK, though I am inclined to leave the LSP changes at `debug` rather than trace; I think getting rid of the timers and the worker pool logging is going to help quite a bit, and I can imagine having the LSP logs might be helpful if you are coming to LSP "blind".

froydnj (approved):

PR #5167, opened 2022-01-26
Stripe packages: missing export_for_test error and correction

ngroman:
If a constant fails to resolve in a packaged _test_ file, we first check inside the associated normal code for the package and if the symbol is found there, instruct the user to `export_for_test`.  In this case we stop looking for other possible matches because this is almost certainly the right fix.

I also fixed an mistake in this this codepath that was adding the hint twice.

### Motivation
This is an easy to make error, but we were not showing good messaging around it.

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/package-error-missing-export/package-error-missing-export.out
+++ b/test/cli/package-error-missing-export/package-error-missing-export.out
@@ -15,8 +15,6 @@ foo_class.rb:10: Unable to resolve constant `NotExported` https://srb.help/5002
   export Foo::Bar::OtherPackage::NotExported`
      4 |  export Foo::Bar::OtherPackage::OtherClass
                                                    ^
-  Note:
-    Try running generate-packages.sh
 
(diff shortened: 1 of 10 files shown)


ngroman:
cc @aisamanra

nroman-stripe:

→ https://go/builds/bui_L253dS7d7TS3Oa
→ https://go/builds/bui_L253YdbDgqWGiu

test/cli/package-error-missing-export/package-error-missing-export.out
```diff
@@ -15,8 +15,6 @@ foo_class.rb:10: Unable to resolve constant `NotExported` https://srb.help/5002
   export Foo::Bar::OtherPackage::NotExported`
      4 |  export Foo::Bar::OtherPackage::OtherClass
                                                    ^
-  Note:
```
ngroman, line 18: These were all unrelated duplicate notes that I fixed.

aprocter (approved):

PR #5169, opened 2022-01-26
add autocorrect heuristics for non-final methods in a final module

froydnj:
### Motivation

This is a fairly easy mistake to make while live coding, and having this offers the possibility of mass-conversion of methods to `sig(:final)` once the enclosing module is marked `final!`.

### Test plan

See included automated tests.  The test was written with the current behavior (no autocorrects) and then updated once the autocorrect was in place, so the diff update is probably more helpful to look at.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -410,14 +410,7 @@ void validateFinalMethodHelper(const core::GlobalState &gs, const core::ClassOrM
                         errMsgClass.show(gs), sym.name(gs).show(gs));
 
             // We don't have a direct link to the signature for the method, so we
-            // heuristically look for the matching signature.  This heuristic
-            // assumes that sigs and method definitions appear on separate lines
-            // and would get tripped up by something like:
-            //
-            // sig {...}; def foo; ...; end
-            // def bar; ... end
-            //
-            // finding the `sig` for `bar` even though the `sig` belongs to `foo`.
+            // heuristically look for the matching signature.
             auto source = defLoc.file().data(gs).source();
             core::LocOffsets sigLocOffsets;
 
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -1,3 +1,5 @@
+#include "absl/strings/ascii.h"
+#include "absl/strings/match.h"
 #include "ast/ast.h"
 #include "ast/treemap/treemap.h"
 #include "common/Timer.h"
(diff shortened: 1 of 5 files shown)


definition_validator/validator.cc
```diff
@@ -402,9 +404,108 @@ void validateFinalMethodHelper(const core::GlobalState &gs, const core::ClassOrM
...
+                // definition.
+                if (line.empty()) {
+                    if (!startFromDefinition) {
+                        break;
+                    }
+                    line_end = prev_line_end;
+                    prev_line_end = source.rfind('\n', line_end - 1);
+                    startFromDefinition = false;
+                    continue;
+                }
```
elliottt, line 443: Won't this only skip one empty line? Is there a test for this case below?

definition_validator/validator.cc
```diff
@@ -402,9 +404,108 @@ void validateFinalMethodHelper(const core::GlobalState &gs, const core::ClassOrM
...
             e.setHeader("`{}` was declared as final but its method `{}` was not declared as final",
                         errMsgClass.show(gs), sym.name(gs).show(gs));
+
+            // We don't have a direct link to the signature for the method, so we
+            // heuristically look for the matching signature.  This heuristic
+            // assumes that sigs and method definitions appear on separate lines
+            // and would get tripped up by something like:
+            //
+            // sig {...}; def foo; ...; end
+            // def bar; ... end
```
elliottt, line 418: Would you mind adding a test for this case to document the incorrect behavior?

elliottt (approved):
This looks good, I just had a few suggestions for some extra test cases.

definition_validator/validator.cc
froydnj: I realized that this example is actually OK, because both methods aren't `final` in the first place, so the sig winds up with an autocorrect no matter what.  I'm leaning towards deleting the example.

definition_validator/validator.cc
froydnj: I changed the comment to reflect reality; I was previously skipping empty lines, but decided against doing that...changed the code but didn't really handle the comment.

test/cli/autocorrect-non-final-methods-final-module/autocorrect-non-final-methods-final-module.sh
```diff
@@ -0,0 +1,27 @@
...
+cwd="$(pwd)"
+infile="$cwd/test/cli/autocorrect-non-final-methods-final-module/autocorrect-non-final-methods-final-module.rb"
+
+tmp="$(mktemp -d)"
+
+cp "$infile" "$tmp"
+
+cd "$tmp" || exit 1
+# Suppress all output because the ordering of errors about non-final methods is
+# dependent on hash table iteration order.
```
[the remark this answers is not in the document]
froydnj: Eh...maybe?  I guess we could accumulate the methods for which we have errors, and then sort those by `Loc`?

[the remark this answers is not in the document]
jez: yeah i think that we've always tried to make the output of sorbet deterministic. seeing something where we can't capture the output in a best because it's order independent seems like a footgun

PR #5170, opened 2022-01-26
Add package layer-checking to sorbet

johnst-stripe:
This PR adds a `--stripe-packages-layers` CLI argument which takes a comma-separated list of values, when set, causes Sorbet to error when:

* a `__package.rb` doesn't contain a `layer` prop
* a `__package.rb` contains a `layer` prop that isn't in the list passed to `--stripe-packages-layers`
* a `__package.rb` of a particular `layer` declares an `import` to another package whose `__package.rb` is of a `layer` that occurs earlier in the list passed to `--stripe-packages-layers`

### Motivation
Better organization of large codebases.

### Test plan

- [X] Added test cases
- [X] Tested on Stripe's codebase
- [ ] Measured performance

See included automated tests.

### DIFF
--- a/core/packages/Layer.h
+++ b/core/packages/Layer.h
@@ -1,0 +1,21 @@
+#ifndef SORBET_CORE_PACKAGES_LAYER_H
+#define SORBET_CORE_PACKAGES_LAYER_H
+
+#include "core/Context.h"
+#include <string_view>
+
+namespace sorbet::core::packages {
+class Layer final {
+public:
+    Layer() = default;
+    Layer(int rank) : rank_(rank){};
+    core::NameRef show(core::Context ctx) const;
+    bool exists() const;
+    bool operator<(const Layer rhs) const;
+    bool operator==(const Layer rhs) const;
+
+private:
+    uint8_t rank_ = 255;
+};
+} // namespace sorbet::core::packages
+#endif // SORBET_CORE_PACKAGES_LAYER_H
--- a/core/packages/PackageDB.cc
+++ b/core/packages/PackageDB.cc
@@ -4,6 +4,7 @@
 #include "core/AutocorrectSuggestion.h"
 #include "core/GlobalState.h"
 #include "core/Loc.h"
+#include "core/packages/Layer.h"
 
 using namespace std;
 
@@ -34,6 +35,11 @@ public:
     Loc definitionLoc() const {
         notImplemented();
         return Loc::none();
+    }
+
+    sorbet::core::packages::Layer layer() const {
+        notImplemented();
+        return Layer(-1);
     }
 
     std::optional<core::AutocorrectSuggestion> addImport(const core::GlobalState &gs, const PackageInfo &pkg,
@@ -168,6 +174,10 @@ const std::string_view PackageDB::errorHint() const {
     return errorHint_;
 }
 
+const std::vector<NameRef> &PackageDB::layerNames() const {
+    return layerNames_;
+}
+
 PackageDB PackageDB::deepCopy() const {
     ENFORCE(frozen);
     PackageDB result;
@@ -177,6 +187,7 @@ PackageDB PackageDB::deepCopy() const {
     }
     result.secondaryTestPackageNamespaceRefs_ = this->secondaryTestPackageNamespaceRefs_;
     result.extraPackageFilesDirectoryPrefixes_ = this->extraPackageFilesDirectoryPrefixes_;
+    result.layerNames_ = this->layerNames_;
     result.packagesByPathPrefix = this->packagesByPathPrefix;
     result.mangledNames = this->mangledNames;
     result.errorHint_ = this->errorHint_;
--- a/core/packages/PackageDB.h
+++ b/core/packages/PackageDB.h
@@ -49,9 +49,12 @@ public:
 
     const std::string_view errorHint() const;
 
+    const std::vector<NameRef> &layerNames() const;
+
 private:
     std::vector<NameRef> secondaryTestPackageNamespaceRefs_;
     std::vector<std::string> extraPackageFilesDirectoryPrefixes_;
+    std::vector<NameRef> layerNames_;
     std::string errorHint_;
 
     UnorderedMap<core::NameRef, std::unique_ptr<packages::PackageInfo>> packages_;
@@ -62,6 +65,7 @@ private:
     std::thread::id writerThread;
 
     friend class UnfreezePackages;
+    friend class Layer;
 };
 
 } // namespace sorbet::core::packages
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -356,6 +356,11 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
     options.add_options("advanced")("stripe-packages-hint-message",
                                     "Optional hint message to add to packaging related errors",
                                     cxxopts::value<string>()->default_value(""));
+    options.add_options("advanced")("stripe-packages-layers",
+                                    "The possible layers a package can belong to, comma-"
+                                    "separated and ordered such that a layer can depend on any layer listed before it."
+                                    "If specified, at least 2 layers must be provided.",
+                                    cxxopts::value<vector<string>>(), "layer1,layer2,layer3");
     options.add_options("dev")("extra-package-files-directory-prefix",
                                "Extra parent directories which contain package files. "
                                "This option must be used in conjunction with --stripe-packages",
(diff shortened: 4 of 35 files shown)


core/packages/PackageDB.cc
```diff
@@ -168,6 +174,24 @@ const std::string_view PackageDB::errorHint() const {
...
 }
 
+const std::string_view PackageDB::getLayerName(uint8_t layerRank) const {
+    ENFORCE(layerRank < layerNames_.size());
+    return layerNames_.at(layerRank);
+}
+
+const uint8_t PackageDB::getLayerRank(std::string layerName) const {
+    for (std::size_t i = 0; i < layerNames_.size(); i++) {
+        if (layerNames_.at(i).compare(layerName) == 0) {
```
jez, line 184: The implementation currently compares full string equality for names.

What do you think about entering the layer names with `gs.enterNameUtf8`, and storing either `UnorderedMap<NameRef, uint8_t>` or `vector<NameRef>` instead?

packager/packager.cc
```diff
@@ -502,6 +509,41 @@ ast::ExpressionPtr prependPackageScope(const core::GlobalState &gs, ast::Express
...
+            for (int i = 1; i < packageDB.getLayerCount(); i++) {
+                layers.append(", ");
+                layers.append(packageDB.getLayerName(i));
+            }
+            e.setHeader("Argument to `{}` must be a valid layer name ({})", fun.show(ctx), layers);
+        }
+        return 0;
+    }
+
+    string layerName = target->asString(ctx.state).toString(ctx.state);
```
jez, line 528: storing things as NameRefs exclusively will let you drop the `toString`, so that `layerName` here can be `NameRef`.

core/packages/Layer.h
```diff
@@ -0,0 +1,20 @@
...
+    Layer(int rank) : rank_(rank){};
+    std::string_view show(core::Context ctx) const;
+    bool operator<(const Layer rhs) const;
+    bool operator==(const Layer rhs) const;
+
+private:
+    uint8_t rank_;
+};
+} // namespace sorbet::core::packages
+#endif // SORBET_CORE_PACKAGES_LAYER_H
```
ngroman, line 20: tailing new line

packager/packager.cc
```diff
@@ -502,6 +509,41 @@ ast::ExpressionPtr prependPackageScope(const core::GlobalState &gs, ast::Express
...
+    uint8_t layerRank = packageDB.getLayerRank(layerName);
+    if (layerRank == 255) {
+        if (auto e = ctx.beginError(expr.loc(), core::errors::Packager::InvalidLayer)) {
+            string layers;
+            layers.append(packageDB.getLayerName(0));
+            for (int i = 1; i < packageDB.getLayerCount(); i++) {
+                layers.append(", ");
+                layers.append(packageDB.getLayerName(i));
+            }
```
ngroman, line 537: See `fmt::map_join` for stuff like this.

packager/packager.cc
ngroman: ++

packager/packager.cc
```diff
@@ -502,6 +509,41 @@ ast::ExpressionPtr prependPackageScope(const core::GlobalState &gs, ast::Express
     return scope;
 }
 
+uint8_t verifyLayer(core::MutableContext ctx, core::NameRef fun, ast::ExpressionPtr &expr) {
```
ngroman, line 512: Could this return a `Layer` instead of the rank?

ngroman, line 512: Does it need to mutate ctx? If not can you change sig to `core::Context`

packager/packager.cc
```diff
@@ -1328,6 +1375,16 @@ class ImportTreeBuilder final {
                 continue;
             }
 
+            if (package.layer() < importedPackage.layer()) {
```
ngroman, line 1378: Trying to remember the stripe logic, don't we only check this for `import` not `test_import`?

Also should have a test case for that.

packager/packager.cc
```diff
@@ -972,6 +1014,10 @@ struct PackageInfoFinder {
                 arg = prependInternalPackageName(ctx, move(send.getPosArg(0)));
             }
         }
+        if (ctx.state.packageDB().getLayerCount() && send.fun == core::Names::layer() && send.numPosArgs() == 1) {
+            auto layerRank = verifyLayer(ctx, core::Names::layer(), send.getPosArg(0));
+            info->layer_ = core::packages::Layer(layerRank);
+        }
         if (send.fun == core::Names::export_for_test() && send.numPosArgs() == 1) {
```
ngroman, line 1021: I think the following should be error conditions:
- specifying a layer when `--stripe-packages-layers` was not specified
- failing to specify a layer when `--stripe-packages-layers` is specified

Would want tests for both

packager/packager.cc
```diff
@@ -502,6 +509,41 @@ ast::ExpressionPtr prependPackageScope(const core::GlobalState &gs, ast::Express
...
+
+    string layerName = target->asString(ctx.state).toString(ctx.state);
+    uint8_t layerRank = packageDB.getLayerRank(layerName);
+    if (layerRank == 255) {
+        if (auto e = ctx.beginError(expr.loc(), core::errors::Packager::InvalidLayer)) {
+            string layers;
+            layers.append(packageDB.getLayerName(0));
+            for (int i = 1; i < packageDB.getLayerCount(); i++) {
+                layers.append(", ");
+                layers.append(packageDB.getLayerName(i));
```
ngroman, line 536: It feels kind of weird here to use `getLayerName` instead of `layer.show(ctx)`

main/options/options.cc
```diff
@@ -909,6 +914,21 @@ void readOptions(Options &opts,
                 opts.secondaryTestPackageNamespaces.emplace_back(ns);
             }
         }
+        if (raw.count("stripe-packages-layers")) {
+            if (!opts.stripePackages) {
+                logger->error("--stripe-packages-layers can only be specified in --stripe-packages mode");
+                throw EarlyReturnWithCode(1);
+            }
```
ngroman, line 921: There should be user visible error if size is not between [2,255)

core/packages/PackageDB.cc
```diff
@@ -168,6 +174,24 @@ const std::string_view PackageDB::errorHint() const {
...
+    }
+    return 255;
+}
+
+const std::size_t PackageDB::getLayerCount() const {
+    return layerNames_.size();
+}
+
 bool PackageDB::isTestFile(const core::GlobalState &gs, const core::File &file) {
     // TODO: (aadi-stripe, 11/26/2021) see if these can all be changed to use getPrintablePath
```
ngroman, line 196: You should copy over the layerNames in `PackageDB::deepCopy()`

core/packages/PackageDB.cc
ngroman: An example of something similar is `AutoloaderConfig::enterConfig`

core/packages/PackageDB.h
```diff
@@ -49,12 +49,17 @@ class PackageDB final {
 
     const std::string_view errorHint() const;
 
+    const std::string_view getLayerName(uint8_t layerRank) const;
+    const uint8_t getLayerRank(std::string layerName) const;
+    const std::size_t getLayerCount() const;
```
ngroman, line 54: This is a nit, but I think a clearer API might be

`const std::vector<Layer> &layers() const;` (this is less true if you end up storing these as a map)

to support `Layer::show` you can make `Layer` a friend class and let it access the layer storage directly.

core/packages/PackageDB.h
```diff
@@ -49,12 +49,17 @@ class PackageDB final {
 
     const std::string_view errorHint() const;
 
+    const std::string_view getLayerName(uint8_t layerRank) const;
+    const uint8_t getLayerRank(std::string layerName) const;
```
ngroman, line 53: Could this take a `const &string` or a string_view?  Doesn't apply if the API ends up changing.

ngroman (changes requested):
Can you add some `test/testdata/packager/` tests as well? You'll need to copy something like https://github.com/sorbet/sorbet/blob/0ec49126970f175b4399c63bd9231dcde2699f05/test/pipeline_test_runner.cc#L304-L307 for entering the layers with a comment

packager/packager.cc
johnst-stripe: I don't have a `Layer` here -- does `Layer(i).show(ctx)` seem more natural?

core/packages/PackageDB.cc
```diff
@@ -168,6 +174,24 @@ const std::string_view PackageDB::errorHint() const {
     return errorHint_;
 }
 
+const std::string_view PackageDB::getLayerName(uint8_t layerRank) const {
+    ENFORCE(layerRank < layerNames_.size());
+    return layerNames_.at(layerRank);
```
ngroman, line 179: No reason to use `.at` if you already have an enforce

packager/packager.cc
ngroman: I guess it depends on what you settle on in the PackageDB interface. If the signature is `vector<Layer> &layers()` you could so something like `fmt::map_join(packageDB.layers(), ", ", [ctx](auto layer) { return layer.show(ctx); })`

core/packages/PackageDB.h
ngroman: `const vector<Foo>&` would not copy the vector since it's a ref and doesn't allow mutation. (You could cast your way out of that but at that point you're really fighting the system intentionally)

johnst-stripe:
ptal @ngroman

aisamanra:

→ https://go/builds/bui_L2nKZXL6D44KZf
→ https://go/builds/bui_L2nKakJHBmOREX

johnst-stripe:
Closing for now, this proposal needs more iteration.

PR #5173, opened 2022-01-26
Log when there are empty parse results

jez:
### Motivation

I want to be able to let users generate test cases for me.

This log line will let me see what file contents people are seeing when they get
empty parse results.

(Note that we already effectively put the entire file contents into the log
file, because we log every LSP protocol message, including the
`textDocument/didChange` requests, this just makes it easier to correlate
"failure condition detected" and "file contents".)

### Test plan

I tested this by running LSP with `--debug-log-file` locally and checking the
contents of the log file.

If we feel strongly, I can write a test for this, but it's a little annoying,
because it only shows up in LSP mode and in the log file. It'd be doable to try
to get the contents of that log file but require some originality.

### DIFF
--- a/hashing/hashing.cc
+++ b/hashing/hashing.cc
@@ -7,6 +7,7 @@
 #include "core/NameSubstitution.h"
 #include "core/Unfreeze.h"
 #include "main/pipeline/pipeline.h"
+#include "rapidjson/writer.h"
 
 using namespace std;
 namespace sorbet::hashing {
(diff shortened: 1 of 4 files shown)


jez:

→ https://go/builds/bui_L2PCppPOUgf0Wo
→ https://go/builds/bui_L2PCcuMN5ZYmSI

main/lsp/LSPTypechecker.cc
```diff
@@ -247,7 +247,7 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
     fast_sort(subset);
     subset.resize(std::distance(subset.begin(), std::unique(subset.begin(), subset.end())));
 
-    config->logger->debug("Taking fast path");
+    config->logger->debug("Running fast path");
```
aprocter, line 250: Was this change intentional? (I have no objection to it if so.)

hashing/hashing.cc
```diff
@@ -27,8 +28,50 @@ pair<ast::ParsedFile, core::UsageHash> rewriteAST(const core::GlobalState &origi
...
+                writer.String(string(file.file.data(*lgs).source()));
+
+                writer.EndObject();
+            }
+
+            logger.debug(result.GetString());
+
+            core::GlobalStateHash invalid;
+            invalid.hierarchyHash = core::GlobalStateHash::HASH_STATE_INVALID;
+            return make_unique<core::FileHash>(move(*lgs->hash()), move(usageHash));
```
aprocter, line 71: Did you mean to pass `invalid` in place of `*lgs->hash()` or something? (Otherwise I'm not sure what we're using `invalid` for.)

aprocter (approved):

main/lsp/LSPTypechecker.cc
jez: yeah, there's another log line that's identical, and i wanted this to be different.

hashing/hashing.cc
jez: good catch, i fixed that, and i wrote a test that would have caught this mistake.

PR #5174, opened 2022-01-26
Move `bool` fields on `File` into bitfields

aisamanra:
This takes the two existing `bool` fields and turns them into a bitfield class, and additionally adds a new one `isPackagedTest` that gets used by our packager mode.

### Motivation
The change here is twofold:
- one, this allows us to use memory more efficiently, packing this and any future flags into less space.
- two, this allows us to skip the expensive `isTestFile` check in the packager by computing it eagerly instead.

### Test plan
See included automated tests.

### DIFF
--- a/core/Files.cc
+++ b/core/Files.cc
@@ -185,7 +185,7 @@ void File::setFileHash(unique_ptr<const FileHash> hash) {
     // If hash_ != nullptr, then the contents of hash_ and hash should be identical.
     // Avoid needlessly invalidating references to *hash_.
     if (hash_ == nullptr) {
-        cached = false;
+        flags.cached = false;
         hash_ = move(hash);
     }
 }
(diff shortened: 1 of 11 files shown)


core/Files.cc
```diff
@@ -166,8 +166,14 @@ CompiledLevel File::fileCompiledSigil(string_view source) {
     }
 }
 
+bool isTestPath(string_view path) {
+    return absl::EndsWith(path, ".test.rb") || absl::StartsWith(path, "./test/") || absl::StrContains(path, "/test/");
```
jez, line 170:
```suggestion
    return absl::EndsWith(path, ".test.rb") || absl::StrContains(path, "/test/");
```

jez: (third was redundant with the second)

core/Files.cc
```diff
@@ -298,4 +304,24 @@ string File::censorFilePathForSnapshotTests(string_view orig) {
...
+    flags.hasParseErrors = value;
+}
+
+bool File::cached() const {
+    return flags.cached;
+}
+
+void File::setCached(bool value) {
+    flags.cached = value;
+}
```
jez, line 325: slight preference for just exposing the flags directly, rather than creating these wrapper functions

core/Files.h
```diff
@@ -112,6 +119,19 @@ class File final {
...
+    struct Flags {
+        // if 'true' file is completely cached in kvstore
+        bool cached : 1;
+        // some reasonable invariants don't hold for invalid files
+        bool hasParseErrors : 1;
+        // only relevant in --stripe-packages mode: is the file a `.test.rb` file?
+        bool isPackagedTest : 1;
+
+        Flags(std::string_view path);
+    };
```
jez, line 131:
```suggestion
    };
    CheckSize(Flags, 1, 1);
```

jez (approved):

PR #5176, opened 2022-01-27
Configure Stripe-internal CSAT

jez:
### Motivation

Stripe has a Chrome extension to collect internal-only feedback on
websites that have been opted in.

This will let people at Stripe report feedback about Sorbet docs
directly to the Sorbet team.

### Test plan

I verified that the div is not rendered when the extension is not present. I
suspect it will work when the extension is present, but sorbet.org is not in the
allow list yet.

### DIFF
--- a/website/static/js/stripe-internal-csat.js
+++ b/website/static/js/stripe-internal-csat.js
@@ -1,0 +1,19 @@
+(() => {
+  if (document.querySelector('#csat-extension-config') != null) {
+    return;
+  }
+
+  const selector = '.mainContainer.docsContainer > .wrapper';
+  const main = document.querySelector(selector);
+  if (main != null) {
+    const csatDiv = document.createElement('div');
+    csatDiv.id = 'csat-extension-config';
+    const config = {
+      embedded: true,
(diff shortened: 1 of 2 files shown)


website/static/js/stripe-internal-csat.js
```diff
@@ -0,0 +1,19 @@
...
+    csatDiv.id = 'csat-extension-config';
+    const config = {
+      embedded: true,
+      prompt: 'Feedback for #sorbet?',
+      notify: '#sorbet',
+    };
+    csatDiv.setAttribute('data-config', JSON.stringify(config));
+    main.appendChild(csatDiv);
+  }
+})();
```
aprocter, line 19: Not-very-frontend-savvy question: why is this div invisible? Is it just because it has no children?

aprocter (approved):
Awesome!

website/static/js/stripe-internal-csat.js
jez: all divs are invisible unless they have a height, width, and content

jez:
doesn't work with CORS

PR #5185, opened 2022-01-27
compare FileRefs directly, rather than via their ids

froydnj:
At some point, comparison operators were added to `FileRef`, but any number of places that still used `.id()` for comparisons weren't updated.  Let's do that update now.

### Motivation

Cleaner code.

### Test plan

See included automated tests.


jez (approved):

froydnj:

→ https://go/builds/bui_L2iBEDq4LOHefa
→ https://go/builds/bui_L2iB9F5pGAuGk1

PR #5188, opened 2022-01-27
Bugfix: Make package names work correctly when they contain ::Package

aadi-stripe:
Fixes a bug in the `packager` which made the "additional paths" lookup fail if a package name contained the string `::Package`. Meaning that package names like `Project::Package` or even `Project::PackagedFiles` would cause an error.

### Motivation

Bugfix in packaging logic.

### Test plan

See included automated tests.


aadi-stripe:

→ https://go/builds/bui_L2jD4jc7FU0iwY
→ https://go/builds/bui_L2jDvZzkgqBMqT

froydnj (approved):

PR #5189, opened 2022-01-27
Allow value omission in hash literals with variables

KaanOzkan:
Part of Ruby 3.1 support.

Continuation of https://github.com/sorbet/sorbet/pull/5141
Ported from https://github.com/whitequark/parser/commit/57bd3718734fea3bc66964badd688c61947b3500

First commit imports new tests from `parser` gem that uses variables.
Second commit implements this functionality using a new `pair_label` method on `Builder.cc` which mostly follows from `pair_keyword` method.
Third commit adds a new `AssignmentToNumparamError` which was used in `parser` gem but not ported to Sorbet. Given how nearly everything else is in parallel with the `parser` gem this felt appropriate. Existing `ReservedForNumparamError` covered this case but now we will give a more accurate error message to the users.

### Motivation

Allowing using Sorbet with Ruby 3.1

### Test plan

See included automated tests.

@Morriar


test/whitequark/test_hash_pair_value_omission_1.parse-tree-whitequark.exp
```diff
@@ -0,0 +1,4 @@
+s(:hash,
```
jez, line 1: this is not specific to your PR, but i noticed that all of our `.exp` tests have the form `test_name.rb.<pass>.exp` except for most of the `parse-tree-whitequark` tests, which don't have the `.rb` in the file name.

do you know why that is? while you're working with these files, would do you mind putting together a change to make that more consistent? (i'd do it, but i'd rather not step on your toes while you're doing this work)

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -3616,6 +3617,10 @@ f_opt_paren_args: f_paren_args
                     {
                       $$ = driver.build.pair_keyword(self, $1, $2);
                     }
+                | tLABEL
+                    {
+                      $$ = driver.build.pair_label(self, $1);
+                    }
```
jez, line 3623: Doesn't this conflict with your other change?

jez (changes requested):
i'm mostly on board with the idea, but the change as written conflicts with the other change you made for hash literals, and i don't really think it makes much sense to review until those conflicts are fixed.

test/whitequark/test_hash_pair_value_omission_1.parse-tree-whitequark.exp
KaanOzkan: It's coming from the import script here https://github.com/sorbet/sorbet/blob/02191d20e3f7c341df69d0ce863fa70866aaad29/tools/scripts/import_whitequark.rb#L256

I can open an another PR to modify the script and rename everything under `test/whitequark`.

parser/parser/cc/grammars/typedruby.ypp
KaanOzkan: Sorry it should've been a simple overwrite but I forgot that this branch was based off of an old master initially. Conflict is resolved now.

parser/Builder.cc
```diff
@@ -1727,6 +1752,22 @@ class Builder::Impl {
...
     }
 
+    void checkAssignmentToNumberedParameters(std::string_view name, core::LocOffsets loc) {
+        if (driver_->lex.context.inDynamicBlock() && isNumberedParameterName(name) &&
+            driver_->numparam_stack.seen_numparams()) {
+            std::cout << "Assignment error" << std::endl;
+            core::Loc location = core::Loc(file_, loc);
+            if (auto e = gs_.beginError(location, core::errors::Parser::AssignmentToNumparamError)) {
+                e.setHeader("cannot assign to numbered parameter {}", name);
+                e.addErrorNote("Reserved numbered parameter names are not allowed starting with Ruby 3.0. Use {} to "
```
jez, line 1762:
```suggestion
                e.addErrorNote("Reserved numbered parameter names are not allowed starting with Ruby 3.0. Use `{}` to "
```

using

```
`{}`
```

instead of just `{}` will cause the interpolated string to be colored

parser/Builder.cc
```diff
@@ -1727,6 +1752,22 @@ class Builder::Impl {
                parser::isa_node<Hash>(&node);
     }
 
+    void checkAssignmentToNumberedParameters(std::string_view name, core::LocOffsets loc) {
+        if (driver_->lex.context.inDynamicBlock() && isNumberedParameterName(name) &&
+            driver_->numparam_stack.seen_numparams()) {
+            std::cout << "Assignment error" << std::endl;
+            core::Loc location = core::Loc(file_, loc);
+            if (auto e = gs_.beginError(location, core::errors::Parser::AssignmentToNumparamError)) {
+                e.setHeader("cannot assign to numbered parameter {}", name);
```
jez, line 1761:
```suggestion
                e.setHeader("cannot assign to numbered parameter `{}`", name);
```

parser/Builder.cc
```diff
@@ -1311,6 +1320,22 @@ class Builder::Impl {
                                  make_unique<Symbol>(keyLoc, gs_.enterNameUTF8(key->view())), std::move(value));
     }
 
+    unique_ptr<Node> pair_label(const token *key) {
+        unique_ptr<Node> value;
+        if (islower(key->asString().at(0))) {
```
jez, line 1325:
```suggestion
        if (islower(key->view()[0])) {
```

`view` provides the `string_view` internally. `asString` allocates a new string.

jez (approved):
Thanks!

I think that the tests are going to fail because you fixed a huge hack I added some weeks ago.

I'm happy to merge once we see a passing build.

jez:
I'd love to merge this but your tests fail and you have conflicts, so I can't. Feel free to re-request review when it's fixed and I'll happily merge.

PR #5190, opened 2022-01-27
change error for `if "string" do ... end` and add an autocorrect

froydnj:
### Motivation

I think this is a better error (since the user probably did mean to write `do` and didn't mean to write `if`) and by changing the error kind to a more specific one, it shows how we can add autocorrects in the context of the parser's diagnostic framework.  (I think we should not convert things to eagerly use Sorbet's errors, since we have things like `replace_last_diagnostic`, which wouldn't play well with Sorbet's errors AIUI.)

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/cc/grammars/typedruby.ypp
+++ b/parser/parser/cc/grammars/typedruby.ypp
@@ -1823,7 +1823,7 @@ int maybe_inject_tNL_impl(parser::symbol_type &yyla, ruby_parser::typedruby27 &d
                     }
                 | kIF strings kDO compstmt if_tail kEND
                     {
-                      driver.diagnostics.emplace_back(dlevel::ERROR, dclass::IfInsteadOfItForTest, diagnostic::range(@1.begin, @1.end), "\"if\"");
+                      driver.diagnostics.emplace_back(dlevel::ERROR, dclass::IfInsteadOfItForTest, diagnostic::range(@1.begin, @1.end));
                       auto &else_ = $5;
                       $$ = driver.build.condition(self, $1, $2, $3, $4,
                         else_ ? else_->tok : nullptr,
--- a/parser/parser/cc/grammars/typedruby.ypp
+++ b/parser/parser/cc/grammars/typedruby.ypp
@@ -1823,9 +1823,7 @@ int maybe_inject_tNL_impl(parser::symbol_type &yyla, ruby_parser::typedruby27 &d
                     }
                 | kIF strings kDO compstmt if_tail kEND
                     {
-                      // TODO(jez) If we had proper Sorbet diagnostics here, we could have a "did you mean"
-                      // error line that suggested changing `if` to `it`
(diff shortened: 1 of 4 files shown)


parser/parser/codegen/generate_diagnostics.cc
```diff
@@ -69,6 +69,7 @@ tuple<string, string> MESSAGES[] = {
     {"PatternDuplicateVariable", "duplicate variable name {}"},
     {"PatternDuplicateKey", "duplicate hash pattern key {}"},
     {"PositionalAfterKeyword", "positional arg \\\"{}\\\" after keyword arg"},
+    {"IfInsteadOfItForTest", "Unexpected use of {}; did you mean to use \\\"it\\\"?"},
```
jez, line 72:
```suggestion
    {"IfInsteadOfItForTest", "Unexpected token \\\"if\\\"; did you mean \\\"it\\\"?"},
```

no point in templating it if the string is hard coded at the call site IMO. Also I think this phrasing sounds slightly more consistent with other errors.

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1823,9 +1823,7 @@ int maybe_inject_tNL_impl(parser::symbol_type &yyla, ruby_parser::typedruby27 &d
                     }
                 | kIF strings kDO compstmt if_tail kEND
                     {
-                      // TODO(jez) If we had proper Sorbet diagnostics here, we could have a "did you mean"
-                      // error line that suggested changing `if` to `it`
-                      driver.diagnostics.emplace_back(dlevel::ERROR, dclass::UnexpectedToken, diagnostic::range(@3.begin, @3.end), "\"do\"");
+                      driver.diagnostics.emplace_back(dlevel::ERROR, dclass::IfInsteadOfItForTest, diagnostic::range(@1.begin, @1.end), "\"if\"");
```
jez, line 1826:
```suggestion
                      driver.diagnostics.emplace_back(dlevel::ERROR, dclass::IfInsteadOfItForTest, diagnostic::range(@1.begin, @1.end));
```

(for if you accept the other one)

jez (approved):
This is great, thanks!

aprocter (approved):

test/testdata/parser/error_recovery_if_do.rb
```diff
@@ -2,6 +2,6 @@
 
 class A
   if 'thing' do
-  #          ^^ error: unexpected token "do"
+# ^^ error: Unexpected use of "if"; did you mean to use "it"?
```
froydnj, line 5:
```suggestion
# ^^ error: Unexpected token "if"; did you mean to use "it"?
```

test/testdata/parser/error_recovery_if_do.rb
```diff
@@ -2,6 +2,6 @@
 
 class A
   if 'thing' do
-  #          ^^ error: unexpected token "do"
+# ^^ error: Unexpected token "if"; did you mean to use "it"?
```
froydnj, line 5:
```suggestion
# ^^ error: Unexpected token "if"; did you mean "it"?
```

PR #5192, opened 2022-01-28
Track whether a file is a package rbi

elliottt:
Track whether an rbi is a package interface by detecting when it has a `.package.rbi` suffix. Use this information to determine if we should enforce `sealed!` checks in the definition validator.

This adds the exceptions to `sealed!` checking from #5067 back in, that were reverted in #5191.

### Motivation

We need to opt package rbis out of additional checks in some cases, this makes that easier.

### Test plan

See included automated tests.


definition_validator/validator.cc
```diff
@@ -479,15 +486,27 @@ void validateSealedAncestorHelper(core::Context ctx, const core::ClassOrModuleRe
...
-                for (auto loc : superClass.data(ctx)->sealedLocs(ctx)) {
-                    e.addErrorLine(loc, "`{}` was marked sealed and can only be inherited in this file",
-                                   superClass.show(ctx));
+
+        // Normally we would skip rbis for the purpose of checking `sealed!`, but when running in stripe-packages mode
+        // this may be the only definition available. Additionally, if the sealed class contains subclasses that are
+        // exported through the package interface and for tests, the subclasses will be spread across multiple rbi files
+        // (the `.package.rbi` and the `.test.package.rbi` files). Because this check would fail in those situations and
+        // the rbi files will have been generated from valid sources, we assume that they are correct here and skip the
+        // check.
```
jez, line 498:
```suggestion
        // Normally we would skip RBIs for the purpose of checking `sealed!`, but when running in stripe-packages mode
        // this may be the only definition available. Additionally, if some of the sealed subclasses are only exported
        // for use in tests, those subclasses will be appear in a separate `.test.package.rbi` file, despite the parent
        // class being defined in a `.package.rbi` file. Because this check would fail in those situations and the RBI
        // files will have been generated from valid sources, we assume that they are correct here and skip the check.
```

jez (approved):

Issue #5166, opened 2022-01-26
Completion does not suggest local variables defined by block scopes


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A10.times%20do%20%7Cxyz%7C%0A%20%20xyy%20%3D%201%0A%20%20xy%0A%20%20%23%20%5E%20completion%3A%20xyy%2C%20xyz%0Aend)

```ruby
# typed: true
extend T::Sig

10.times do |xyz|
  xyy = 1
  xy
  # ^ completion: xyy, xyz
end
```

#### Observed output

```
editor.rb:6: Method xy does not exist on T.class_of(<root>) https://srb.help/7003
     6 |  xy
          ^^
Errors: 1
```

![image](https://user-images.githubusercontent.com/5544532/151084521-f0e9ae1c-38a9-4554-9fb5-0595da3a860a.png)

#### Expected behavior

should show `xyy` and `xyz`

- - -

PR #5195, opened 2022-01-28
fix "find all references" to take block arguments into account

### LINKED ISSUE #4803: Block parameter definition not included in find all references
#### Input

Right click on `var` in `sorbet.run` and select find all references.

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0A%5B%5D.each%20do%20%7Cvar%7C%0A%20%20var%0Aend%0A)

```ruby
# typed: true

[].each do |var| # Not found as reference
  var # Found as reference
end
```

#### Observed output

The only item in the list is the `var` inside the block's body.

#### Expected behavior

The definition of `var` between the pipes should be included in the reference list, just like it works for method parameters.

This issue was found while working on #4734 and getting block parameter renaming working.

**Note**: fixing this problem should cause the `rename/locals` test to break, because it currently expects the wrong behavior to happen. Remember to update the tests including the rename of the block parameter definition and delete the comment in the test file.

froydnj:
### Motivation

Fixes #4803

cc @vinistock 

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LocalVarSaver.cc
+++ b/main/lsp/LocalVarSaver.cc
@@ -21,6 +21,25 @@ core::MethodRef enclosingMethod(core::Context ctx, core::Loc loc) {
 }
 }
 
+ast::ExpressionPtr LocalVarSaver::postTransformBlock(core::Context ctx, ast::ExpressionPtr tree) {
+    auto &block = ast::cast_tree_nonnull<ast::Block>(tree);
+    auto method = enclosingMethod(ctx, core::Loc(ctx.file, block.loc));
+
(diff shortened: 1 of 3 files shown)


jez:
This looks good. Does this fix #5166 as well?

jez (approved):

jez:
> This looks good. Does this fix #5166 as well?

Actually now that I'm looking at the code, I'm *guessing* that it doesn't, but that the solution to #5166 looks basically identical, but applied to LocalVarFinder.

PR #5196, opened 2022-01-29
Initialize Timer::endTime and log for >1w

jez:
### Motivation

We're seeing some `Timer`s at Stripe with eggregiously long times (like 300
weeks, which is longer than the lifetime of Sorbet by multiple years).

This is an attempt to fix the issue, and at least log if the issue is happening.

### Test plan

See included automated tests.


common/Timer.cc
```diff
@@ -151,6 +151,11 @@ Timer::~Timer() {
     if (!canceled && dur.usec > clock_threshold_coarse.usec) {
         // the trick ^^^ is to skip double comparison in the common case and use the most efficient representation.
         sorbet::timingAdd(this->name, start, clock, move(args), move(tags), self, prev, move(histogramBuckets));
+        auto one_week = microseconds{24LL * 60 * 60 * 1000 * 1000};
```
elliottt, line 154: Is it possible to mark this `constexpr`?

elliottt (approved):
Looks good!

common/Timer.cc
jez: doesn't seem to work for me, but also godbolt shows that clang will [just eliminate it anyways](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:3,endLineNumber:8,positionColumn:3,positionLineNumber:8,selectionStartColumn:3,selectionStartLineNumber:8,startColumn:3,startLineNumber:8),source:'//+Type+your+code+here,+or+load+an+example.%0A%23include+%3Cstdint.h%3E%0A%23include+%3Cstdlib.h%3E%0A%0Astruct+microseconds+%7B%0A++++int64_t+usec%3B%0A++++microseconds(int64_t+usec)+:+usec(usec)+%7B%7D%0A%7D%3B%0A%0Aint+main(int+argc,+char+**argv)+%7B%0A++++microseconds+dur(atoll(argv%5B1%5D))%3B%0A++++auto+one_week+%3D+microseconds%7B24LL+*+60+*+60+*+1000+*+1000%7D%3B%0A++++if+(dur.usec+%3E+one_week.usec)+%7B%0A++++++++return+42%3B%0A++++%7D+else+%7B%0A++++++++return+0%3B%0A++++%7D%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:clang1201,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B17+-O2',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+clang+12.0.1+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

Issue #5179, opened 2022-01-27
Bad parse for bad block pass syntax


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/?arg=--print=parse-tree-whitequark#class%20A%0A%20%20def%20foo%0A%20%20%20%20x.map%20%7B%26%3Afoo%7D%0A%20%20end%0Aend)

```ruby
class A
  def foo
    x.map {&:foo}
  end
end
```

#### Observed output

```
s(:begin)
editor.rb:3: unexpected token "&" https://srb.help/2001
     3 |    x.map {&:foo}
                   ^

editor.rb:1: class definition in method body https://srb.help/2001
     1 |class A
        ^^^^^
Errors: 2
```

#### Expected behavior

Should parse better

- - -

jez:
Similar:

https://sorbet.run/?arg=--print=parse-tree-whitequark#class%20A%0A%20%20def%20foo%0A%20%20%20%20x.map%20%7B%7Ca%7C%20%26%3Afoo%7D%0A%20%20end%0Aend

aprocter:
This partially addressed by #5198, but that doesn't handle the example from jez's comment above where the block has some `|args|`.

PR #5198, opened 2022-01-29
Recover from &block args incorrectly wrapped in curly braces

aprocter:
Partially addresses #5179.

This lets us recover from parse errors for cases where the user types something like:

```
foo.map {&:to_s}
```

(Note the extraneous curly braces.) It also supplies an autocorrect.

This PR does not handle the (perhaps less common?) case raised in #5179 where the curlies include block args, e.g.

```
foo.map {|x| &:to_s}
```

Not sure how much that matters. It should be possible to support this, but it looked like it was going to require deeper refactoring of the parser to do so.

### Motivation
Better recovery from parse failures from common errors.

### Test plan
See included automated tests.

### DIFF
--- a/parser/Parser.cc
+++ b/parser/Parser.cc
@@ -38,13 +38,11 @@ class ErrorToError {
                 if (source.has_value()) {
                     auto replacement = string(loc.source(gs).value());
                     size_t lCurlyPos = replacement.find("{");
-                    if (lCurlyPos != string::npos) {
-                        replacement = replacement.erase(lCurlyPos, 1);
-                        size_t rCurlyPos = replacement.rfind("}");
-                        if (rCurlyPos != string::npos) {
-                            replacement = replacement.erase(rCurlyPos, 1);
-                            e.replaceWith("Remove the curly braces", loc, replacement);
-                        }
+                    size_t rCurlyPos = replacement.rfind("}");
+                    if (lCurlyPos != string::npos && rCurlyPos != string::npos) {
+                        replacement[lCurlyPos] = '(';
+                        replacement[rCurlyPos] = ')';
+                        e.replaceWith("Replace the curly braces with parens", loc, replacement);
                     }
                 }
                 break;
--- a/parser/Parser.cc
+++ b/parser/Parser.cc
@@ -24,7 +24,8 @@ class ErrorToError {
         return min((uint32_t)(pos), maxOff);
     }
 
-    static void maybeAddAutocorrect(core::ErrorBuilder &e, core::Loc loc, ruby_parser::dclass errorClass) {
+    static void maybeAddAutocorrect(core::GlobalState &gs, core::ErrorBuilder &e, core::Loc loc,
+                                    ruby_parser::dclass errorClass) {
         switch (errorClass) {
             case ruby_parser::dclass::IfInsteadOfItForTest:
                 e.replaceWith("Replace with `it`", loc, "it");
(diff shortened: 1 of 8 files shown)


aprocter-stripe:

→ https://go/builds/bui_L3CFpqEj9QcTe8
→ https://go/builds/bui_L3CFXRlJT70jIy

froydnj:
Similar to how we "parse" `if "string" do ... end`, I wonder if it'd be simpler to explicitly match `tLCURLY tAMPER arg_value tRCURLY` in `brace_block`, build the block as normal, but issue an error (and as a bonus, offer an autocorrect).

aprocter:
@froydnj Good call! I'll give that a shot.

parser/Parser.cc
```diff
@@ -24,14 +24,31 @@ class ErrorToError {
...
+                auto source = loc.source(gs);
+                if (source.has_value()) {
+                    auto replacement = string(loc.source(gs).value());
+                    size_t lCurlyPos = replacement.find("{");
+                    if (lCurlyPos != string::npos) {
+                        replacement = replacement.erase(lCurlyPos, 1);
+                        size_t rCurlyPos = replacement.rfind("}");
+                        if (rCurlyPos != string::npos) {
+                            replacement = replacement.erase(rCurlyPos, 1);
+                            e.replaceWith("Remove the curly braces", loc, replacement);
```
jez, line 46: don't we want to replace the curly braces with parentheses?

jez: i'm not sure that it's obvious from the examples you have in your tests that we should have an autocorrect for this. it seems like the autocorrect is only sensible for the first case (`{ &:foo }`). suggesting "remove the curly braces" doesn't actually fix the other problems, which makes for a confusing user experience

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -2479,6 +2495,22 @@ opt_block_args_tail:
                       $$ = block;
                       driver.lex.context.pop();
                     }
+                  // Error-recovery case for typos like "foo {&:bar}"
+                | tLCURLY
+                    {
+                      driver.lex.context.push(Context::State::BLOCK);
+                    }
```
jez, line 2502: this causes a reduce/reduce conflict, doesn't it? with the midrule action for the case immediately above?

aprocter: I thought it would, but oddly it doesn't seem to.

parser/Parser.cc
aprocter: Ooh parens are probably better, yeah!

The examples in the tests are kind of contrived (particularly the `&(1+1)` thing which is just going to go from a parse error to a type error), but I can imagine more complex cases that actually will end up type checking, like:

```
x.map {&(some_condition ? :to_s
                        : :to_i)}
```

I'm not sure how common that actually is, but autocorrecting the curlies to parens would give us something that works.

parser/parser/cc/grammars/typedruby.ypp
jez: interesting. to be explicit and to ward against this being a problem in the future, what do you think about factoring that out into a proper production, like

```
lcurly_block_start : tLCURLY { push(...) }
```

and then using the lcurly_block_start non-terminal in both places?

parser/Parser.cc
```diff
@@ -24,14 +24,29 @@ class ErrorToError {
...
         switch (errorClass) {
             case ruby_parser::dclass::IfInsteadOfItForTest:
                 e.replaceWith("Replace with `it`", loc, "it");
                 break;
             case ruby_parser::dclass::MissingCommaBetweenKwargs:
                 e.replaceWith("Insert a comma", loc, ", ");
                 break;
+            case ruby_parser::dclass::CurlyBracesAroundBlockPass: {
+                auto source = loc.source(gs);
+                if (source.has_value()) {
```
elliottt, line 38: Is this check necessary? Won't the source always exist as the loc comes from the file being parsed?

elliottt (approved):
Looks good!

parser/Parser.cc
aprocter: True, I imagine it's not strictly required. What do you think about switching this to an `ENFORCE`?

jez: No, this is very much intentional

jez: I made `.source()` return `nullopt` because it was a common source of crashes

jez: #4075

jez: ah i see what you're saying, we know exactly where the `loc` comes from in this particular case.

Even still, i'd kind of prefer writing it this way because i don't want people to cargo-cult that you should always `ENFORCE` when using `.source`, if you don't mind.

aprocter: Works for me!

aprocter: Works for me!

PR #5200, opened 2022-01-29
desugar: Reorder cases by current frequency

jez:
### Test plan

See included automated tests.


froydnj (approved):
I was just thinking about this the other day, thank you for doing this!

(We really have more const references than sends in raw ruby files?  Wild.)

jez:

→ https://go/builds/bui_L3XKkPkiF1NEYc
→ https://go/builds/bui_L3XKJdYP7S9AN2

jez:
> (We really have more const references than sends in raw ruby files? Wild.)

- `A::B::C::D::E.f` is 5 constants and one send
- Stripe's Ruby build system creates a bunch of constant-heavy files in the `build/` folder (for things like the autoloader and subclasses lists)

jez:
This only accounts for a slight speedup (0.5% or so) on a cold cache.

http://go/gist/jez/ec60b8e0c543e9b70eb6dc6dc946d156

PR #5203, opened 2022-01-30
Better "why is this nil?" error message

jez:
### Motivation

**Before**
Screen Shot 2022-01-31 at 12 04 17 PM

**After**

Screen Shot 2022-01-31 at 12 56 51 PM

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1059,9 +1059,38 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                     auto iff = MK::If(loc, std::move(cond), std::move(rhs), std::move(lhs));
                     result = std::move(iff);
                 } else {
-                    core::NameRef tempName = dctx.freshNameUnique(core::Names::andAnd());
-                    auto temp = MK::Assign(loc, tempName, std::move(lhs));
-                    auto iff = MK::If(loc, MK::Local(loc, tempName), std::move(rhs), MK::Local(loc, tempName));
+                    // TODO(jez) completion test
+                    // TODO(jez) private test (both implicit and explicit self)
+                    // TODO(jez) parser::Or as well?
+                    // TODO(jez) Can we make this handle `!(...).nil?` and `(...).is_a?(...)` or is that hard?
+                    // TODO(jez) This doesn't generalize to all `if` conditions like
+                    //   if a.foo; a.foo.even?; end
+                    // is that ok?
+                    auto andAndTemp = dctx.freshNameUnique(core::Names::andAnd());
+
+                    auto *lhsSend = ast::cast_tree<ast::Send>(lhs);
+                    auto *rhsSend = ast::cast_tree<ast::Send>(rhs);
+
+                    ExpressionPtr thenp;
+                    if (lhsSend != nullptr && rhsSend != nullptr) {
+                        auto lhsSource = core::Loc(dctx.ctx.file, lhsSend->loc).source(dctx.ctx);
+                        auto rhsRecvSource = core::Loc(dctx.ctx.file, rhsSend->recv.loc()).source(dctx.ctx);
+                        if (lhsSource.has_value() && lhsSource == rhsRecvSource) {
+                            // TODO(jez) What locs should we be using?
+                            rhsSend->insertPosArg(0, std::move(rhsSend->recv));
+                            rhsSend->insertPosArg(1, MK::Local(loc, andAndTemp));
+                            rhsSend->insertPosArg(2, MK::Symbol(rhsSend->funLoc, rhsSend->fun));
+                            rhsSend->recv = MK::Constant(loc, core::Symbols::Magic());
+                            rhsSend->fun = core::Names::checkAndAnd();
+                            thenp = std::move(rhs);
+                        } else {
+                            thenp = std::move(rhs);
+                        }
+                    } else {
+                        thenp = std::move(rhs);
+                    }
+                    auto temp = MK::Assign(loc, andAndTemp, std::move(lhs));
+                    auto iff = MK::If(loc, MK::Local(loc, andAndTemp), std::move(thenp), MK::Local(loc, andAndTemp));
                     auto wrapped = MK::InsSeq1(loc, std::move(temp), std::move(iff));
                     result = std::move(wrapped);
                 }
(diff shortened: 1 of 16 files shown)


jez:
Some followup issues:

- Can we handle `.is_a?(...) && ...` and `!(...).nil? && ...`?
- Can we handle `(...).nil? || ...`?
- This doesn't generalize to arbitrary `if` conditions—is there another approach that might work better for more cases for free?

ast/desugar/Desugar.cc
```diff
@@ -1059,9 +1059,32 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
...
+                    auto andAndTemp = dctx.freshNameUnique(core::Names::andAnd());
+
+                    auto *lhsSend = ast::cast_tree<ast::Send>(lhs);
+                    auto *rhsSend = ast::cast_tree<ast::Send>(rhs);
+
+                    ExpressionPtr thenp;
+                    if (lhsSend != nullptr && rhsSend != nullptr) {
+                        auto lhsSource = core::Loc(dctx.ctx.file, lhsSend->loc).source(dctx.ctx);
+                        auto rhsRecvSource = core::Loc(dctx.ctx.file, rhsSend->recv.loc()).source(dctx.ctx);
+                        if (lhsSource.has_value() && lhsSource == rhsRecvSource) {
```
aprocter, line 1071: So it looks like we're only inserting `checkAndAnd`s when the lhs and the rhs receiver are lexically identical, right? So for example, `x.foo && x . foo . bar` wouldn't get one. (I think that's probably reasonable, if I'm understanding correctly.)

core/types/calls.cc
```diff
@@ -2451,6 +2451,140 @@ class Magic_selfNew : public IntrinsicMethod {
     }
 } Magic_selfNew;
 
+class Magic_checkAndAnd : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        // args[0] is recv
+        // args[1] is the && tmp var
+        // args[2] is the method name
+        // args[3...] are the args
```
aprocter, line 2460: A high-level comment on what we expect the desugarer to produce as it relates to the original `lhs && rhs.foo`, and how we consume that info, might be nice here.

core/types/calls.cc
```diff
@@ -2451,6 +2451,140 @@ class Magic_selfNew : public IntrinsicMethod {
...
+            selfTy.type,
+            args.block,
+            args.originForUninitialized,
+            args.isPrivateOk,
+            args.suppressErrors,
+        };
+        auto dispatched = selfTy.type.dispatchCall(gs, innerArgs);
+
+        auto multipleComponents = dispatched.secondary != nullptr;
+        if (multipleComponents) {
```
aprocter, line 2506: Why can we bail out on `!multipleComponents`? Is it because nilable is a `T.any`?

core/types/calls.cc
```diff
@@ -2451,6 +2451,140 @@ class Magic_selfNew : public IntrinsicMethod {
...
+                    foundErrorOnRetry |= !it->main.errors.empty();
+                }
+
+                if (!foundErrorOnRetry) {
+                    for (auto it = &dispatched; it != nullptr; it = it->secondary.get()) {
+                        for (auto &err : it->main.errors) {
+                            if (err->what == core::errors::Infer::UnknownMethod && it->main.receiver.isNilClass()) {
+                                if (auto newErr = gs.beginError(err->loc, core::errors::Infer::CallAfterAndAnd)) {
+                                    newErr.setHeader(
+                                        "Call to method `{}` after `{}` assumes result type doesn't change",
```
aprocter, line 2543: I wonder if this phrasing will be a bit confusing, since "it's still nilable" doesn't exactly jump out here. Maybe something like "Receiver of call to method foo is still nilable even after &&"?

test/cli/autocorrect_and_and/autocorrect_and_and.rb
```diff
@@ -0,0 +1,26 @@
...
+class A
+  extend T::Sig
+
+  sig {returns(T.nilable(Integer))}
+  def foo; nil; end
+
+  def bar
+    a = A.new
+
+    if a.foo && a.foo.even?
```
aprocter, line 12: Mildly interesting corner case you might test (maybe this isn't the right spot for it): if there's whitespace before `even?`, e.g. `a.foo.<NL>even?` (I think the autocorrect should not fire since you're looking for a period immediately preceding.)

aprocter (approved):
Good stuff!

ast/desugar/Desugar.cc
jez: Yeah, that's right. it's not great

core/types/calls.cc
jez: hm, isn't that what this comment is? i'm not sure what more you'd like

core/types/calls.cc
jez: each component in the DispatchResult corresponds to one branch of the `T.any`

The way that dispatchCall works is that it descends down the structure of the `T.any`, calls `dispatchCall` on each component, and then builds back up a `DispatchResult` that records both sides of the dispatch. Check out the code for `OrType::dispatchCall`

the error message doesn't make sense if there is less than one component—if there is only one component, then we should leave any original error message(s) untouched.

jez-stripe:
test

core/types/calls.cc
jez-stripe: Ah gotcha. I would have assumed that to be implied from the fact that this is an intrinsic for a `<Magic>` method

aprocter: I guess my basic point is that the purpose of the `<Magic>` method itself isn't fully clear unless you go searching for the code that emits it, so documenting that would be good. (That applies to a lot of the `<Magic>` stuff, though!)

jez: (i added a comment)

test/cli/autocorrect_and_and/autocorrect_and_and.rb
jez: great test!

ast/desugar/Desugar.cc
jez: i added a test for this.

core/types/calls.cc
aprocter: Thanks! That's exactly what I was looking for.

PR #5207, opened 2022-01-31
Improve colors for "Method does not exist" error

jez:
### Motivation

This previously was not using our `explainGot` helper, which made it slightly
different from all the other "Got `Type` ..." error sections.

Now the errors have colors.

I had to make one change to the `explainGot` helper, which was to not emit the
`ErrorSection` if we had no explanation for where the type came from.

### Test plan

See changes to existing tests.

### DIFF
--- a/test/cli/suggest_unsafe_dead/suggest_unsafe_dead.out
+++ b/test/cli/suggest_unsafe_dead/suggest_unsafe_dead.out
@@ -147,7 +147,6 @@ suggest_unsafe_dead.rb:81: This code is unreachable https://srb.help/7006
     suggest_unsafe_dead.rb:80: Replaced with `T.unsafe(self)`
     80 |    unless self
                    ^^^^
-  Got `A` originating from:
 
 suggest_unsafe_dead.rb:89: This code is unreachable https://srb.help/7006
     89 |      do_something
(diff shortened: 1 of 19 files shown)


test/cli/suggest_unsafe_dead/suggest_unsafe_dead.out
```diff
@@ -147,7 +147,6 @@ suggest_unsafe_dead.rb:81: This code is unreachable https://srb.help/7006
     suggest_unsafe_dead.rb:80: Replaced with `T.unsafe(self)`
     80 |    unless self
                    ^^^^
-  Got `A` originating from:
```
froydnj, line 150: The removals here seem unusual.  Were these expected?

froydnj (approved):
Oh, I see, the removed errors are because we now avoid giving you a "Got" message if we can't determine one?

test/cli/suggest_unsafe_dead/suggest_unsafe_dead.out
jez: yeah, they're kindof useless because the list of locations explaining how we got `A` (which is already in the error message elsewhere) is empty. So it's redundant.

PR #5214, opened 2022-02-01
Add ability to rewind lexer

jez:
### Motivation

This lets me delete a bunch of super lurky code I added a while back in favor of
something that's much more understandable.

It also will be a building block for two upcoming parser-related changes
(indentation-aware recovery, recovering from mismatched quotes).

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/cc/grammars/typedruby.ypp
+++ b/parser/parser/cc/grammars/typedruby.ypp
@@ -810,6 +810,10 @@ int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser:
                     {
                       $$ = driver.build.logicalAnd(self, $1, $2, $3);
                     }
+                | kAND error
+                    {
+                      $$ = driver.build.error_node(self, @1.begin, @1.end);
+                    }
                 | expr kOR expr
                     {
                       $$ = driver.build.logicalOr(self, $1, $2, $3);
--- a/parser/parser/cc/grammars/typedruby.ypp
+++ b/parser/parser/cc/grammars/typedruby.ypp
@@ -523,27 +523,14 @@ int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser:
 // bunch of the already-parsed program.
 //
 // So in rules where we expect something like that to happen, we can call this
-// function to fix up the lookahead token by manually creating a tNL
-int maybe_inject_tNL_impl(parser::symbol_type &yyla, ruby_parser::TYPEDRUBY27 &driver, size_t prev_end, size_t next_begin) {
-  // This function is super hacky, and is best understood by carefully reading
-  // yylex and where yylex is called inside the generated parser
-  auto fake_token = driver.lex.tok_view_from_offsets(prev_end, next_begin);
-  auto nl_start = fake_token.find('\n');
-  if (nl_start != std::string_view::npos) {
-    token_t lookahead = yyla.value.token;
-    yyla.clear(); // same as yyclearin macro
-    auto nl_type = ruby_parser::token_type::tNL;
-    auto nl_tok = driver.lex.unadvance(lookahead, nl_type, nl_start, nl_start + 1, "\n");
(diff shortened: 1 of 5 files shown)


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -523,27 +523,14 @@ int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser:
...
-}
-
-#define maybe_inject_tNL(prev_end, next_begin) ((yyla.type) = yytranslate_(maybe_inject_tNL_impl(yyla, driver, (prev_end), (next_begin))))
+// function to request that the lexer start again after `tDOT` token in the
+// expr_end state.
+#define rewind_and_reset(newPos)     \
+  do {                               \
+    yyclearin;                       \
+    driver.lex.rewind(newPos);       \
+    driver.lex.set_state_expr_end(); \
```
froydnj, line 532: The lexer maintains a, um, lot of state; is this going to be sufficient to reset all of it?  (Actually, the driver does too, but I guess that is a separate question...maybe?)

aprocter: Yeah, I'm a little concerned about this too. If it happens to work in places where this PR uses it I think it's fine, but since it doesn't touch other parts of the state it feels like it could be a footgun if someone tries to call it without awareness of its limitations. (For one example: it doesn't reset `in_kwarg`, so if we were to rewind past the beginning of the kwargs, maybe that could screw up... something?) Might make sense to insert a caution in the comments.

parser/parser/include/ruby_parser/lexer.hh
```diff
@@ -190,13 +190,13 @@ public:
...
-    // lookahead token) back onto the front of the token_queue, and then allocates a new token and
-    // returns it.
-    token_t unadvance(token_t tok_to_push, token_type type, size_t start, size_t end, const std::string &str);
-
     std::string_view tok_view_from_offsets(size_t start, size_t end) const;
 
+    // Useful for error recovery. Manually sets `_p` so that the next call to `advance` starts
+    // from the desired location. If you're using this, you likely want to set the lexer state
+    // menually with one of the below helpers.
+    void rewind(size_t newPos);
```
aprocter, line 198: Similar comment here: unless we're sure that calling one of the below helpers is sufficient to fully roll back lexer state (and I don't think we are sure of that), it would be good to note that here.

aprocter (approved):

aprocter (approved):

parser/parser/cc/grammars/typedruby.ypp
jez: This is a good point, but I think that luckily this will only happen during error recovery. So it'll be bugs in our new error recovery code, instead of regressions in the existing parser things. That's a tradeoff I'm mostly fine with at the moment, because we we get bug reports for the parser it should hopefully be easy to triage and fix resetting other bits of state that we'll need.

That being said, this is a good callout, and i'll go ahead and audit all the state we're setting now and see how much of it we'll need to manually reset.

aprocter: > So it'll be bugs in our new error recovery code, instead of regressions in the existing parser things. That's a tradeoff I'm mostly fine with at the moment, because we we get bug reports for the parser it should hopefully be easy to triage and fix resetting other bits of state that we'll need.

Yup, I think that's a good argument for trying it and seeing how it goes!

jez: I audited the the stateful things and concluded that none of them except for `in_kwarg` looked like they obviously needed to or should be reset.

I moved the logic into a proper helper function where the name makes it clear that it's going to reset the state.

I also limited it to just resetting state to the `expr_end` state, which I think at least makes it clear that the method doesn't attempt to arbitrarily rewind—I don't think we'll need it to, and i think that rewinding to `expr_end` in all the current and foreseeable cases is straightfoward.

Thanks again for the suggestion.

PR #5216, opened 2022-02-02
Add support for package-aware autoloading

aadi-stripe:
Adds package-awareness to autogen, and functionality to register packages.

### Motivation

As part of packaging work in autogen, we are requiring certain autoload definitions to be generated in the namespace of a given package.

### Test plan

See included automated tests. Tested on Stripe codebase

cc @ngroman @aisamanra

### DIFF
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -227,7 +227,7 @@ string DefTree::renderAutoloadSrc(const core::GlobalState &gs, const AutoloaderC
         if (pkgName.exists()) {
             ENFORCE(!gs.packageDB().empty());
             const string_view shortName = pkgName.shortName(gs);
-            const string_view mungedName = shortName.substr(0, shortName.rfind(core::PACKAGE_SUFFIX));
+            const string_view mungedName = shortName.substr(0, shortName.size() - core::PACKAGE_SUFFIX.size());
             fmt::format_to(std::back_inserter(buf), "\n{}.register_package({}, '{}')\n", alCfg.registryModule, fullName,
                            mungedName);
         }
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -210,6 +210,7 @@ string DefTree::renderAutoloadSrc(const core::GlobalState &gs, const AutoloaderC
             fmt::format_to(std::back_inserter(buf), "{}.on_autoload('{}')\n", alCfg.registryModule, fullName);
             predeclare(gs, fullName, buf);
         }
+
         if (!children.empty()) {
             fmt::format_to(std::back_inserter(buf), "\n{}.autoload_map({}, {{\n", alCfg.registryModule, fullName);
             vector<pair<core::NameRef, string>> childNames;
@@ -221,6 +222,14 @@ string DefTree::renderAutoloadSrc(const core::GlobalState &gs, const AutoloaderC
                                children.at(pair.first)->path(gs));
             }
             fmt::format_to(std::back_inserter(buf), "}})\n", fullName);
+        }
+
+        if (pkgName.exists()) {
+            ENFORCE(!gs.packageDB().empty());
+            const string_view shortName = pkgName.shortName(gs);
+            const string_view mungedName = shortName.substr(0, shortName.rfind(core::PACKAGE_SUFFIX));
+            fmt::format_to(std::back_inserter(buf), "\n{}.register_package({}, '{}')\n", alCfg.registryModule, fullName,
+                           mungedName);
         }
     } else if (type == Definition::Type::Casgn || type == Definition::Type::Alias ||
                type == Definition::Type::TypeAlias) {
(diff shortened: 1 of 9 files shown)


packager/packager.cc
```diff
@@ -1850,51 +1860,53 @@ vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers
...
-                if (result.gotItem()) {
-                    filesProcessed++;
-                    auto &file = job.file.data(gs);
-                    core::Context ctx(gs, core::Symbols::root(), job.file);
-
-                    if (file.sourceType == core::File::Type::Normal) {
-                        job = rewritePackagedFile(ctx, move(job));
-                    } else if (file.sourceType == core::File::Type::Package) {
-                        job = rewritePackage(ctx, move(job));
+    if (!gs.runningUnderAutogen) {
```
aadi-stripe, line 1863: Don't run the rest of the packager in autogen

packager/packager.cc
```diff
@@ -1838,6 +1840,14 @@ vector<ast::ParsedFile> Packager::findPackages(core::GlobalState &gs, WorkerPool
                 } else {
                     packages.db.enterPackage(move(pkg));
                 }
+
+                if (gs.runningUnderAutogen) {
+                    it = files.erase(it);
```
aadi-stripe, line 1845: Drop all package files in autogen (after packages are entered), to prevent namer errors.

aadi-stripe:

→ https://go/builds/bui_L4x1HxqSPJStoo
→ https://go/builds/bui_L4x1E64W3REFpf

ngroman:
Heads up, I'll need a bit to completely work it out, but I think I have an idea for a cleaner/smaller API.

main/autogen/autoloader.cc
```diff
@@ -222,6 +223,25 @@ string DefTree::renderAutoloadSrc(const core::GlobalState &gs, const AutoloaderC
             }
             fmt::format_to(std::back_inserter(buf), "}})\n", fullName);
         }
+
+        if (!alCfg.packagedShimPaths.empty() && gs.packageDB().isPackage(qname.nameParts)) {
```
aadi-stripe, line 227: While this code runs on every constant, `isPackage` is quite fast due to binary search and hence total overhead on Stripe codebase is not much.

packager/packager.cc
ngroman: FYI erase like this is O(total len * deletes).  It's why `remove_if` exists.

aadi-stripe: Ah ok, wanted to do it in-place in the same loop, but that makes sense. Will fix.

ngroman (changes requested):
What do you think about an API more like https://github.com/sorbet/sorbet/pull/5223/files

Also have some perf improvements in that implementaiton.

I think it would be cleaner and more flexible. At runtime we can make the `register_package` API do what we need.

aadi-stripe:

→ https://go/builds/bui_L5958qBkgC4MiY
→ https://go/builds/bui_L595Kj50ls4Rgr

aadi-stripe:

→ https://go/builds/bui_L59H7AA5bIB9b7
→ https://go/builds/bui_L59HLEjYWiNnRI

aadi-stripe:
ptal @ngroman 

Made changes.

packager/packager.cc
```diff
@@ -1849,6 +1849,12 @@ vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers
     Timer timeit(gs.tracer(), "packager");
 
     files = findPackages(gs, workers, std::move(files));
+    if (gs.runningUnderAutogen) {
+        auto it = std::remove_if(files.begin(), files.end(),
```
ngroman, line 1853:
```suggestion
        // Autogen only requires package metadata. Remove the package files.
        auto it = std::remove_if(files.begin(), files.end(),
```

main/autogen/autoloader.cc
```diff
@@ -222,6 +223,14 @@ string DefTree::renderAutoloadSrc(const core::GlobalState &gs, const AutoloaderC
             }
             fmt::format_to(std::back_inserter(buf), "}})\n", fullName);
         }
+
+        if (pkgName.exists()) {
+            ENFORCE(!gs.packageDB().empty());
+            const string_view shortName = pkgName.shortName(gs);
+            const string_view mungedName = shortName.substr(0, shortName.rfind(core::PACKAGE_SUFFIX));
```
ngroman, line 230:
```suggestion
            const string_view mungedName = shortName.substr(0, shorName.size() - core::PACKAGE_SUFFIX.size());
```
Nit: does this need a find?

ngroman (approved):
lgtm, see inline suggestions

main/autogen/autoloader.cc
aadi-stripe: No, will work with backwards index math as well.

PR #5228, opened 2022-02-03
Upgrade to LLVM 13

vinistock:

Upgrade to LLVM 13, which is the first version that supports signing binaries, which is required on M1s. It's easier to review this PR per commit, here's what they do:
- Upgrade to LLVM 13 and update all references
- Explicitly set DEVELOPER_DIR and SDKROOT in the jemalloc build. I'm not sure why these are not set after the upgrade, but this was the only way I could fix the build
- Fix usage of `result_of` which is now [deprecated](https://en.cppreference.com/w/cpp/types/result_of) in favour of `invoke_result`
- After the upgrade, we got a few errors for variables that are set but not used. Made a few changes to get past the errors and warnings
- Disabled the `unused-but-set-variable` for lmdb, which violates the rule

### Motivation

We will need LLVM 13 for signing binaries on the ARM64 builds.

### Test plan

Current tests should cover the upgrade.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,6 +12,9 @@ common:test --test_env=GTEST_INSTALL_FAILURE_SIGNAL_HANDLER=1
 ##
 ##  Custom toolchain.
 ##
+# The --crosstool_top option has been deprecated in favor of this incompatible cc toolchain resolution option for
+# selecting toolchains. This flag can be removed once https://github.com/bazelbuild/bazel/issues/7260 is closed and the
+# migration is complete
 build --incompatible_enable_cc_toolchain_resolution --copt=-D_LIBCPP_ENABLE_NODISCARD
 
 ##
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,7 +12,7 @@ common:test --test_env=GTEST_INSTALL_FAILURE_SIGNAL_HANDLER=1
 ##
 ##  Custom toolchain.
 ##
-build --crosstool_top=@llvm_toolchain_12_0_0//:toolchain --copt=-D_LIBCPP_ENABLE_NODISCARD
+build --incompatible_enable_cc_toolchain_resolution --copt=-D_LIBCPP_ENABLE_NODISCARD
 
 ##
 ##  Common build options across all build configurations
@@ -202,9 +202,9 @@ build:ubsan --define unsanitized=false
 build:ubsan --copt=-DHAS_SANITIZER
 
 # Bazel links C++ files with $CC, not $CXX, this breaks UBSan
-build:sanitize-linux --linkopt=external/llvm_toolchain_12_0_0/lib/clang/12.0.0/lib/linux/libclang_rt.asan_cxx-x86_64.a
-build:sanitize-linux --linkopt=external/llvm_toolchain_12_0_0/lib/clang/12.0.0/lib/linux/libclang_rt.ubsan_standalone_cxx-x86_64.a
-build:sanitize-linux --linkopt=external/llvm_toolchain_12_0_0/lib/clang/12.0.0/lib/linux/libclang_rt.ubsan_standalone-x86_64.a
+build:sanitize-linux --linkopt=external/llvm_toolchain_13_0_0/lib/clang/13.0.0/lib/linux/libclang_rt.asan_cxx-x86_64.a
+build:sanitize-linux --linkopt=external/llvm_toolchain_13_0_0/lib/clang/13.0.0/lib/linux/libclang_rt.ubsan_standalone_cxx-x86_64.a
+build:sanitize-linux --linkopt=external/llvm_toolchain_13_0_0/lib/clang/13.0.0/lib/linux/libclang_rt.ubsan_standalone-x86_64.a
 build:sanitize-linux --config=sanitize
 
 build:sanitize-mac --config=sanitize
@@ -253,7 +253,7 @@ build:webasm --compilation_mode=opt
 build:webasm --copt=-DNDEBUG --linkopt=-DNDEBUG # for some reason emscripten doesn't pass those when -O2\-Oz are specified
 build:webasm --copt=--llvm-lto --copt=3 --linkopt=--llvm-lto --linkopt=3
 # Specify a "sane" C++ toolchain for the host platform.
-build:webasm --host_crosstool_top=@llvm_toolchain_12_0_0//:toolchain
+build:webasm --host_crosstool_top=@llvm_toolchain_13_0_0//:toolchain
 
 ##
 ## Stripe's ci passes --config=ci, we need it to exist
--- a/.bazelrc
+++ b/.bazelrc
@@ -254,7 +254,8 @@ build:webasm --cpu=webasm --spawn_strategy=local --genrule_strategy=local --copt
 build:webasm --define release=true
 build:webasm --compilation_mode=opt
 build:webasm --copt=-DNDEBUG --linkopt=-DNDEBUG # for some reason emscripten doesn't pass those when -O2\-Oz are specified
-build:webasm --copt=--llvm-lto --copt=3 --linkopt=--llvm-lto --linkopt=3
+build:webasm --copt=-DEMSCRIPTEN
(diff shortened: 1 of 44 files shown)


vinistock:
It looks like we have to turn off `set-but-unused-variable` in the [llvm.autogenerated.BUILD](https://github.com/sorbet/sorbet/blob/master/third_party/llvm/llvm.autogenerated.BUILD) based on the failures. How would we go about doing so?

jez:
> It might just be a matter of updating the expectations. Thoughts?

Yeah, that looks right to me. There should be a script to update them:

`tools/scripts/update_compiler_exp.sh`

.bazelrc
```diff
@@ -12,7 +12,7 @@ common:test --test_env=GTEST_INSTALL_FAILURE_SIGNAL_HANDLER=1
 ##
 ##  Custom toolchain.
 ##
-build --crosstool_top=@llvm_toolchain_12_0_0//:toolchain --copt=-D_LIBCPP_ENABLE_NODISCARD
+build --incompatible_enable_cc_toolchain_resolution --copt=-D_LIBCPP_ENABLE_NODISCARD
```
jez, line 15: I have a feeling you're more familiar with bazel toolchains than i ever have been at this point 😅 do you mind adding a comment here explaining what this bazel flag does for us? doesn't have to be long.

third_party/jemalloc.BUILD
```diff
@@ -25,6 +25,8 @@ JEMALLOC_BUILD_COMMAND = """
   export CC=$$(absolutize $(CC))
   export CXX=$$(absolutize $(CC))
   [ "$$(uname)" = "Linux" ] && export AR=$$(absolutize $(AR))
+  [ "$$(uname)" = "Darwin" ] && export DEVELOPER_DIR=$$(xcode-select --print-path)
+  [ "$$(uname)" = "Darwin" ] && export SDKROOT=$$(xcrun --show-sdk-path)
```
jez, line 29: wow, does this mean that jemalloc now builds on macOS? i have never been able to get that working since like macOS Mohave

resolver/type_syntax.cc
```diff
@@ -613,12 +613,6 @@ TypeSyntax::ResultType interpretTCombinator(core::Context ctx, const ast::Send &
                 if (auto e = ctx.beginError(send.loc, core::errors::Resolver::InvalidTypeDeclaration)) {
                     e.setHeader("`{}` has been renamed to `{}`", "T.enum", "T.deprecated_enum");
 
-                    auto prefix = 0;
-                    auto sendSource = core::Loc(ctx.file, send.loc).source(ctx);
-                    if (sendSource.has_value() && absl::StartsWith(sendSource.value(), "::")) {
-                        prefix += 2;
-                    }
-
```
jez, line 621: 🙈 this is definitely my fault! nice to know that clang can catch this now

jez (approved):
Looks great, very few changes, thanks for breaking it up by commit!!

.bazelrc
vinistock: Done!

vinistock:
There are a few issues left to fix for CI. I'd appreciate some feedback on them.

**Linux release**
This is erroring in `autoconf` while building jemalloc mentioning that it cannot find `ld`. I believe this comes from the option we set in the jemalloc build for `fuse-ld=lld` in Linux. But I'm not sure why it can't find it. Any ideas on why it can't find it? Are we missing a path in Linux?

**Emscripten**
- The first error, was that `llvm-lto` is an unsupported option. I switched to using `lto-O2`, which I believe is the equivalent
- The second error we hit is that the `EMSCRIPTEN` macro, used for compiling crypto_hashing is no longer defined. It seems something external used to define this, since I can't find it anywhere in Sorbet. I manually defined it in the webasm build in `bazelrc`, which gets us further in the build
- Now the build fails because the environment variables defined in [the emscripten build's link options](https://github.com/sorbet/sorbet/blob/master/emscripten/BUILD#L5-L28) are being considered as positional arguments to clang, which makes it panic

Is this the right way for fixing emscripten? Where should we stick these environment variables now that clang won't accept them?

froydnj:
> There are a few issues left to fix for CI. I'd appreciate some feedback on them.
> 
> **Linux release** This is erroring in `autoconf` while building jemalloc mentioning that it cannot find `ld`. I believe this comes from the option we set in the jemalloc build for `fuse-ld=lld` in Linux. But I'm not sure why it can't find it. Any ideas on why it can't find it? Are we missing a path in Linux?

I think this is because before this change, jemalloc was getting compiled by the clang defined by the LLVM toolchain, whereas after this change, it's getting compiled by the host compiler (which is GCC on my machine).  The former naturally knows how to find lld, since it comes with the compiler, the latter not so much.

> **Emscripten**
> 
>     * The first error, was that `llvm-lto` is an unsupported option. I switched to using `lto-O2`, which I believe is the equivalent
> 
>     * The second error we hit is that the `EMSCRIPTEN` macro, used for compiling crypto_hashing is no longer defined. It seems something external used to define this, since I can't find it anywhere in Sorbet. I manually defined it in the webasm build in `bazelrc`, which gets us further in the build
> 
>     * Now the build fails because the environment variables defined in [the emscripten build's link options](https://github.com/sorbet/sorbet/blob/master/emscripten/BUILD#L5-L28) are being considered as positional arguments to clang, which makes it panic
> 
> 
> Is this the right way for fixing emscripten? Where should we stick these environment variables now that clang won't accept them?

The problem is not that clang won't accept them: the problem is that, for whatever reason, the compiler that Bazel thinks you should be using to compile the emscripten release is now clang from LLVM 13 rather than `emcc.sh` from the Emscripten toolchain.  That would account for both the `EMSCRIPTEN` macro disappearing (since LLVM 13 won't define that) and why `-s` is no longer accepted (likewise).

It's not clear to me *why*, but reverting the change to add `--incompatible_enable_cc_toolchain_resolution` (in favor of just bumping the `--crosstool_top` version, like other changes in `.bazelrc`) makes `--config=release-linux` and the emscripten build both work properly...on my machine, anyway.  Did you change that because there were actual warnings happening during the build (I don't see any?)?

vinistock:
I changed it based on the [instructions on bazel-toolchain's README](https://github.com/grailbio/bazel-toolchain#quickstart). Maybe we can still get by using `crosstool_top`?

froydnj:
> I changed it based on the [instructions on bazel-toolchain's README](https://github.com/grailbio/bazel-toolchain#quickstart). Maybe we can still get by using `crosstool_top`?

I think if we want to follow that README, we need to not use `--crosstool_top` for anything and we need to sort out our current uses of `@bazel_tools//tools/cpp:current_cc_toolchain` and friends.

I'd suggest switching back to `--crosstool_top` for now and see where that gets you.

OAuthBringer:
I'm just going to comment here and provide general applause and encouragement.  There are developers with m1s that want all the benefits of docker and sorbet together that are counting on you!

If you were signed up with https://creators.brave.com/ I'd tip you in BAT!

matyus:
is there anything I can do to help get this merged?

vinistock:
We initially believed LLVM 13 would be required for M1 support, but realized it is not necessary. The work to finalize M1 support is currently blocked waiting for the ARM64 toolchain support to be finalized in [bazel-toolchain](https://github.com/grailbio/bazel-toolchain/issues/88).

[This branch](https://github.com/sorbet/sorbet/compare/master...vinistock:enable_arm64_compilation?expand=1) can compile Sorbet to ARM64 without LLVM 13, but uses a [work in progress implementation of the toolchains](https://github.com/grailbio/bazel-toolchain/pull/85).

joaolrpaulo:
@vinistock Any updates on support for M1 based macs, having to run an 🔨 and old version of sorbet is not desirable.

When can we expect this to be tackled?

vinistock:
LLVM 15 was shipped with pre-built ARM64 darwin binaries, which will probably make the work easier, but we do not have a specific timeline yet.

I should also mention that Sorbet's x86 release works normally on M1 macs through rosetta. If you have rosetta installed, you should be able to run the latest version.

PR #5230, opened 2022-02-03
Use portable debug traps

vinistock:

Use portable debug traps instead of embedded x86 assembly.

### Motivation

The x86 instructions will not compile on ARM64, so we need to use a portable alternative.

### Test plan

### DIFF
--- a/common/os/linux.cc
+++ b/common/os/linux.cc
@@ -89,7 +89,7 @@ bool amIBeingDebugged() {
 
 bool stopInDebugger() {
     if (amIBeingDebugged()) {
-        __asm__("int $3");
+        __builtin_debugtrap();
         return true;
     }
     return false;
--- a/common/os/mac.cc
+++ b/common/os/mac.cc
@@ -69,7 +69,7 @@ bool amIBeingDebugged()
 
 bool stopInDebugger() {
     if (amIBeingDebugged()) {
-        __asm__("int $3");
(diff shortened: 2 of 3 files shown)


jez (approved):

vinistock:
Not sure why the compiler complains about `__builtin_debugtrap()`. I think an alternative is including `signal.h` and using `raise(SIGTRAP)`, but I'm not 100% sure. Thoughts?

elliottt:
> Not sure why the compiler complains about `__builtin_debugtrap()`. I think an alternative is including `signal.h` and using `raise(SIGTRAP)`, but I'm not 100% sure. Thoughts?

This seems like a good solution to me 👍

froydnj:
> Not sure why the compiler complains about `__builtin_debugtrap()`. I think an alternative is including `signal.h` and using `raise(SIGTRAP)`, but I'm not 100% sure. Thoughts?

GCC apparently doesn't support `__builtin_debugtrap`, which is what is being used in CI to compile Ruby (maybe just the GCC version we're using?).  You could try changing it to `__builtin_trap` just for `vm-payload.c`, or you could add processor-specific code to emit the same instructions that Clang would.

vinistock:
Changed only the occurrence in `vm-payload` to use `raise(SIGTRAP)`, which seems to do the trick.

elliottt (approved):
Thanks!

PR #5231, opened 2022-02-03
Conditionally compile libb2 only on x86

vinistock:

Conditionally compile `libb2` only on x86 builds to keep the performance gains and use the portable version everywhere else.

### Motivation

We need to use the portable version `blake2` instead of the x86 specific `libb2` for ARM64 builds.

### Test plan

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -76,6 +76,9 @@ test:coverage --spawn_strategy=local
 build --define=versioned=false
 build:versioned --workspace_status_command=tools/buildstamp/get_workspace_status --stamp --define=versioned=true
 
+build:dbg-darwin --config=dbg --platforms=@//tools/config:darwin_x86_64
+build:dbg-linux --config=dbg --platforms=@//tools/config:linux_x86_64
+
 ##
 ## Configurations used for releases
 ##
@@ -93,7 +96,7 @@ build:release-debug-common --config=skipslowenforce
 
 # harden: mark relocation sections read-only
 build:release-linux --linkopt=-Wl,-z,relro,-z,now
-build:release-linux --config=lto-linux --config=release-common
+build:release-linux --config=lto-linux --config=release-common --platforms=@//tools/config:linux_x86_64
 
 # This is to turn on vector instructions where available.
 # We used to do this unconditionally, but Rosetta 2 doesn't translate all vector instructions well.
@@ -106,7 +109,7 @@ build:release-linux --config=lto-linux --config=release-common
 build:release-linux           --copt=-march=sandybridge
 build:release-sanitized-linux --copt=-march=sandybridge
 
-build:release-mac --config=release-common
+build:release-mac --config=release-common --platforms=@//tools/config:darwin_x86_64
 
 build:release-debug-linux --config=release-linux
 build:release-debug-linux --config=release-debug-common
--- a/common/crypto_hashing/crypto_hashing.h
+++ b/common/crypto_hashing/crypto_hashing.h
@@ -3,7 +3,7 @@
 
 #include "common/common.h"
 extern "C" {
-#ifndef EMSCRIPTEN
+#if defined(__i386__) || defined(__x86_64__)
 #include "blake2.h"
 #else
 #include "ref/blake2.h"
@@ -15,7 +15,7 @@ inline std::array<uint8_t, 64> hash64(std::string_view data) {
     static_assert(BLAKE2B_OUTBYTES == 64);
     std::array<uint8_t, 64> res;
 
-#ifndef EMSCRIPTEN
+#if defined(__i386__) || defined(__x86_64__)
     int err = blake2b(&res[0], data.begin(), nullptr, std::size(res), data.size(), 0);
 #else
     // it has different order of arguments \facepalm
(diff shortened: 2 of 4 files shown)


common/crypto_hashing/crypto_hashing.h
```diff
@@ -3,7 +3,7 @@
 
 #include "common/common.h"
 extern "C" {
-#ifndef EMSCRIPTEN
+#ifdef X86_BUILD
```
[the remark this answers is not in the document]
vinistock: I'll give that a try. What does `__i386__` match to? I thought we'd only want `libb2` for `x86`.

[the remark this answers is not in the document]
froydnj: `__i386__` is what you use to denote "x86"; there is no `__x86__` macro.

aprocter (changes requested):
@vinistock I think @froydnj's suggestion to switch to `if defined(__i386__) || defined(__x86_64__)` should work here, and allow us to drop most if not all of the bazel changes.

vinistock:
I switched to the suggest macros, which indeed works. However, I'm still unsure as to why the `EMSCRIPTEN` macro is no longer defined. Switching to `__EMSCRIPTEN__` also did not help. Any ideas?

tools/config/BUILD
```diff
@@ -35,14 +57,14 @@ config_setting(
 config_setting(
     name = "webasm",
     values = {
-        "cpu": "webasm",
+        "define": "webassembly=true",
```
froydnj, line 60: Reverting this bit enables me (in conjunction with reverting the below bit in `webasm_opt`) to get a successful build.

froydnj (dismissed):
Building locally, I get:

```
froydnj@pyro:~/src/sorbet$ ./bazel build --verbose_failures //emscripten:sorbet-wasm.tar --config=webasm-linux --strip=always
INFO: Invocation ID: b9842ff1-b3fb-42a9-abb7-4ca1bac77b31
INFO: Analyzed target //emscripten:sorbet-wasm.tar (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /home/froydnj/src/sorbet/third_party/licenses/BUILD:26:10: Linking third_party/licenses/generate_licenses [for host] failed: (Exit 1): clang failed: error executing command
  (cd /home/froydnj/.cache/bazel/_bazel_froydnj/c161c0eb7edcd00e8e8eaa053696c88e/execroot/com_stripe_ruby_typer && \
  exec env - \
    PATH=/bin:/usr/bin:/usr/local/bin \
    PWD=/proc/self/cwd \
  /home/froydnj/.cache/bazel/_bazel_froydnj/c161c0eb7edcd00e8e8eaa053696c88e/external/llvm_toolchain_12_0_0/bin/clang @bazel-out/host/bin/third_party/licenses/generate_licenses-2.params)
# Configuration: 58c91cff40f1f3e25ea4b4c062cb5bce35d9d6e0c0a551b05a59cd1b24e6308e
# Execution platform: @local_config_platform//:host
ld.lld: error: undefined symbol: getProgramName()
>>> referenced by backtrace.cc
>>>               backtrace.o:(sorbet::Exception::printBacktrace()) in archive bazel-out/host/bin/common/libcommon.a

ld.lld: error: undefined symbol: addr2line(std::__1::basic_string_view<char, std::__1::char_traits<char> >, void const* const*, int)
>>> referenced by backtrace.cc
>>>               backtrace.o:(sorbet::Exception::printBacktrace()) in archive bazel-out/host/bin/common/libcommon.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Target //emscripten:sorbet-wasm.tar failed to build
INFO: Elapsed time: 0.510s, Critical Path: 0.21s
INFO: 30 processes: 29 internal, 1 local.
FAILED: Build did NOT complete successfully
```

and groveling through `bazel-out/host/bin/common/libcommon.a`, I see that we've somehow selected, for the host, `emscripten.o` rather than `linux.o` as one would expect.  This is not going to end well.

I think you need to remove the `webassembly=true` definition bits in `.bazelrc` and `tools/config/BUILD`.

tools/config/BUILD
vinistock: That indeed does the trick. Not sure why I was running into issues earlier.

.bazelrc
```diff
@@ -76,6 +76,9 @@ test:coverage --spawn_strategy=local
 build --define=versioned=false
 build:versioned --workspace_status_command=tools/buildstamp/get_workspace_status --stamp --define=versioned=true
 
+build:dbg-darwin --config=dbg --platforms=@//tools/config:darwin_x86_64
+build:dbg-linux --config=dbg --platforms=@//tools/config:linux_x86_64
```
aprocter, line 80: Do we need these anymore? I don't think they're being referenced.

vinistock: We'll need those to select the platform we want to build for - i.e.: we'd be using these from the command line `./bazel build //main:sorbet --config=dbg-darwin`, which is why there aren't any explicit references in `.bazelrc`.

These are especially useful when cross compiling, since we need to select a platform that is different from the host.

aprocter (approved):
I think this lgtm now!

aprocter-stripe:

→ https://go/builds/bui_L9qRDyzQvdAzXb
→ https://go/builds/bui_L9qR3Kjq5RG2j4

PR #5232, opened 2022-02-03
Remove unnecessary Loc bounds check

elliottt:
We shouldn't expect the loc to be correct for internal methods used for metadata storage, in the rbi generator.

### Motivation

Fixing a TODO introduced in #5067 .

### Test plan

See included automated tests.

### DIFF
--- a/core/Loc.cc
+++ b/core/Loc.cc
@@ -257,12 +257,7 @@ optional<string_view> Loc::source(const GlobalState &gs) const {
         return nullopt;
     } else {
         auto source = this->file().data(gs).source();
-        // TODO: this should be unnecessary
-        if (endPos() <= source.size()) {
-            return source.substr(beginPos(), endPos() - beginPos());
-        } else {
-            return nullopt;
-        }
+        return source.substr(beginPos(), endPos() - beginPos());
     }
 }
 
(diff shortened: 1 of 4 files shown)


core/Loc.cc
```diff
@@ -257,12 +257,7 @@ optional<string_view> Loc::source(const GlobalState &gs) const {
         return nullopt;
     } else {
         auto source = this->file().data(gs).source();
-        // TODO: this should be unnecessary
-        if (endPos() <= source.size()) {
-            return source.substr(beginPos(), endPos() - beginPos());
-        } else {
-            return nullopt;
-        }
+        return source.substr(beginPos(), endPos() - beginPos());
```
jez, line 260: 🎉

jez (approved):
Is there an rbi-gen test we can write to prevent regressions?

elliottt:
> Is there an rbi-gen test we can write to prevent regressions?

I've been trying to come up with a reproducer for the crash, but it's proving difficult. I'll try a bit more and merge this without if I can't get one.

elliottt:
> > Is there an rbi-gen test we can write to prevent regressions?
> 
> I've been trying to come up with a reproducer for the crash, but it's proving difficult. I'll try a bit more and merge this without if I can't get one.

The real crash seems to be a resolver bug that we're triggering: the location generated for the mixes_in_class_methods fake method is incorrect.

PR #5081, opened 2022-01-11
Better method completion given syntax errors

jez:
### Motivation

Some of the commits are purely whitespace/clang-format changes. You might want to review by commit or ignore whitespace in the diff.

### Test plan

See included automated tests.

PR #5251, opened 2022-02-04
Thread `--trace-lexer` option through pipeline

jez:
### Motivation

Unlike Bison which has built-in support for tracing, Ragel just runs the code
you tell it to. I've added a simple `--trace-lexer` option that prints out the
tokens and some metadata about them.

I'm not attached to the particular format, and we might want to add (or even
remove) things from it in the future. In particular, it doesn't print the `cs`
variable (the lexer's current state machine state) because there's no easy way
to go from the `int` describing the state to a human-readable version of the
state. (I've [started a feature request][1] for that here.)

[1]: https://github.com/adrian-thurston/ragel/discussions/82

**ALSO**, I noticed that after the changes introduced in #5081 that the
`last_token_s` and `last_token_e` variables maintained in the lexer were
actually redundant, so I've removed them.

### Test plan

Tested that this worked manually

### DIFF
--- a/parser/parser/cc/grammars/typedruby.ypp
+++ b/parser/parser/cc/grammars/typedruby.ypp
@@ -486,7 +486,7 @@ void parser::error(const ruby_parser::location &lloc, const std::string &msg) {
 
 	driver.diagnostics.emplace_back(
 		dlevel::ERROR, dclass::UnexpectedToken,
-		diagnostic::range(driver.lex.last_token_s, driver.lex.last_token_e),
+		diagnostic::range(lloc.begin, lloc.end),
 		error_message);
 }
 
--- a/parser/parser/cc/token.cc
+++ b/parser/parser/cc/token.cc
@@ -29,3 +29,9 @@ std::string_view token::view() const {
 std::string token::asString() const {
     return std::string(_string);
 }
+
+std::ostream &operator<<(std::ostream &o, const ruby_parser::token &token) {
+    return o << "ruby_parser::token{.start = " << token.start() << ", .end = " << token.end()
+             << ", .type = " << ruby_parser::token::tokenTypeName(token.type()) << ", .str = \"" << token.view()
+             << "\" }";
+}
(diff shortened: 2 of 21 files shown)


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -486,7 +486,7 @@ void parser::error(const ruby_parser::location &lloc, const std::string &msg) {
 
 	driver.diagnostics.emplace_back(
 		dlevel::ERROR, dclass::UnexpectedToken,
-		diagnostic::range(driver.lex.last_token_s, driver.lex.last_token_e),
+		diagnostic::range(lloc.begin, lloc.end),
```
jez, line 489: This was the only use of `last_token_s` and `last_token_e`, and the arity of `parser::error` changes when you've passed the `%define api.location.type` to bison to include `lloc`, which has the `begin` and `end` set to the same thing that last_token_{s,e} had.

parser/parser/cc/token.cc
```diff
@@ -29,3 +29,9 @@ std::string_view token::view() const {
 std::string token::asString() const {
     return std::string(_string);
 }
+
+std::ostream &operator<<(std::ostream &o, const ruby_parser::token &token) {
+    return o << "ruby_parser::token{.start = " << token.start() << ", .end = " << token.end()
+             << ", .type = " << ruby_parser::token::tokenTypeName(token.type()) << ", .str = \"" << token.view()
+             << "\" }";
+}
```
jez, line 37: I opted to implement this using `operator<<` because it doesn't look like we currently use `fmt` anywhere, and I also had to add a `operator<<` implementation for the `ruby_parser::location` type that Bison uses (it *only* lets you use `operator<<` for debug information), so I figured it was nice to be consistent in the `parser/` folder. (I also like that this implementation doesn't materialize an intermediate string).

jez: The downside is that it's more annoying to use in `lldb`

elliottt (approved):
👍

PR #5253, opened 2022-02-04
Remove an unnecessary nullptr check in Trees.cc

elliottt:
This nullptr check is no longer necessary -- I can't reproduce a crash with it removed. I suspect that the crash was related to the one fixed in #5232.

### Motivation

Cleaning up todos from mering #5067.

### Test plan

See included automated tests.


jez (approved):
thanks!

PR #5257, opened 2022-02-04
Fix crash in completion with bad TypeConstraints

jez:
### Motivation

This was causing lots of crashes in LSP mode.

### Test plan

I confirmed that the included test caused a crash before this change.

The exception happened in `TypeConstraint::findSolution` because we would be
looking in a solved type constraint for the solution to a type variable that did
not exist in that `TypeConstraint`'s solution set.

### DIFF
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -200,16 +200,10 @@ SimilarMethodsByName allSimilarMethods(const core::GlobalState &gs, core::Dispat
                                        string_view prefix) {
     auto result = similarMethodsForReceiver(gs, dispatchResult.main.receiver, prefix);
 
-    // Convert to shared_ptr and take ownership
-    shared_ptr<core::TypeConstraint> constr = move(dispatchResult.main.constr);
-
     for (auto &[methodName, similarMethods] : result) {
         for (auto &similarMethod : similarMethods) {
             ENFORCE(similarMethod.receiverType == nullptr, "About to overwrite non-null receiverType");
             similarMethod.receiverType = dispatchResult.main.receiver;
-
-            ENFORCE(similarMethod.constr == nullptr, "About to overwrite non-null constr");
-            similarMethod.constr = constr;
         }
     }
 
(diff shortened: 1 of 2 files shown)


test/testdata/lsp/completion/constr.rb
```diff
@@ -0,0 +1,15 @@
+# typed: true
+
+extend T::Sig
+
+xs = T.let([], T::Array[Integer])
+
+# This completion result is a bit non-sensical at the moment (it doesn't makes
+# sense to return any results there) but at least it doesn't crash Sorbet like
+# it used to. If you made a change that improves this test, feel free to delete
+# this comment.
```
elliottt, line 10: What ends up being nonsensical about the completion result? The assertion on line 15 is for `each`, which doesn't seem crazy for a tab completion on the result of a `map`.

main/lsp/requests/completion.cc
```diff
@@ -912,9 +906,18 @@ vector<unique_ptr<CompletionItem>> CompletionTask::getCompletionItems(LSPTypeche
...
     }
     for (auto &similarMethod : dedupedSimilarMethods) {
+        // Even though we might have one or more TypeConstraints on the DispatchResult that triggered this completion
+        // request, those constraints are the result of solving the current method. These new methods we're about to
+        // suggest are their own methods with their own type variables, so it doesn't make sense to use the old
+        // constraint for the new methods.
+        //
+        // What this means in practice is that the prettified `sig` in the completion documentation will show
+        // `T.type_parameter(:U)` instead of a solved type.
+        auto constr = nullptr;
```
elliottt, line 916: Thanks for this comment, it really cleared up the deletions above 👍 

Sort of amazing that `nullptr` works here, is the deduced type `std::nullptr_t`?

elliottt (approved):
Looks good 👍

test/testdata/lsp/completion/constr.rb
jez: arguably that would make sense if you asked for completion results at the end of the `each` method name, not after the `end` keyword. IMO it's kind of like asking for completion results right after a `}`: we should probably show you other results or no results. you're right that it's subjective—i was just trying to let people know that i'm not attached to the current completion results.

main/lsp/requests/completion.cc
jez: > Sort of amazing that nullptr works here, is the deduced type std::nullptr_t?

yep

PR #5260, opened 2022-02-04
Raise errors when keyword args are given to T combinators

elliottt:
Currently `T.nilable(a: Integer)` will silently turn into `T.untyped`. The keyword args are treated as an implicit positional hash when type-checking the send itself, but the type is interpreted as `T.untyped` because of the early-exit in type_syntax.cc when interpreting `T.nilable`.

This PR adds an explicit error for T-combinators that shouldn't accept keyword arguments, and an autocorrect to wrap those arguments in curly braces.

### Motivation

Catching malformed uses of T-combinators.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/type_syntax.cc
+++ b/resolver/type_syntax.cc
@@ -534,19 +534,13 @@ void unexpectedKwargs(core::Context ctx, const ast::Send &send) {
         name.append(send.fun.shortName(ctx));
         e.setHeader("`{}` does not accept keyword arguments", name);
 
-        core::LocOffsets start;
-        if (send.hasPosArgs()) {
-            // Using the endPos of the last positional argument ensures that we also cut out the trailing comma on that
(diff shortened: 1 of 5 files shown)


jez (approved):
If we changed Sorbet to be aware of Ruby 3’s keyword arg checking, we’d have caught this is my understanding right? (As in: we might be able to delete this in the future)

elliottt:
Stripe migration PR: https://go/ps/pull/417113

elliottt:
> If we changed Sorbet to be aware of Ruby 3’s keyword arg checking, we’d have caught this is my understanding right? (As in: we might be able to delete this in the future)

Yes, as we can explicitly forbid keyword arguments with `**nil`, that would be a better option for enforcing that these combinators only receive positional arguments.

PR #5261, opened 2022-02-05
No-op changes in service of indentation-aware parser

jez:
### Motivation

I've had a somewhat long-lived branch full of changes in support of #5000. This
PR takes the ones that don't change behavior and lands them (except for one
small improvement for `kIF error` that is actually new).

### Commit summary

- **no-op: Mark `Builder::build` const** (1f25166d8)

- **no-op cleanup in Parser.cc** (dbadbb847)

- **pre-work: Better recovery for `kIF kEND`** (9ec9e729c)

- **pre-work: Expose yytname_ and yytranslate_ to driver** (1c171cad6)

  This lets us avoid having to write our code in C macros at the top of
  the parser file, which will then get exploded into usage sites, and
  instead lets us factor it into a simple helper function on the driver.

- **pre-work: Factor rewind_and_reset to driver** (979cbcf9f)

- **This test gets better** (2629f757f)

### Test plan

Most things don't change, except one test.


aprocter (approved):
LGTM, thanks for the commit-by-commit!

PR #5266, opened 2022-02-05
Also emit error `Note:` lines in LSP mode

jez:
### Motivation

LSP error messages were missing error notes because they don't have
locations associated with them. The LSP protocol either lets you set the
message or related information. Most clients assume that the message is
a single line. All related information must have a location associated
with it or it's not rendered.

To at least get the error notes to show up, we re-use the error's
location itself.

### Test plan

**Before**

Screen Shot 2022-02-04 at 11 29 58 PM

**After**

Screen Shot 2022-02-04 at 11 29 19 PM


elliottt (approved):
👍

Issue #5182, opened 2022-01-27
Unmatched array square bracket doesn't recover from parse


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/?arg=--print=parse-tree-whitequark#%23%20Craft%20your%20test%20case%2C%20then%20share%20at%20http%3A%2F%2Fgo%2Fask%2Fsorbet%2C%20thanks!%0Aclass%20A%0A%20%20def%20foo%0A%20%20%20%20%5B%0A%20%20end%0Aend)

```ruby
# Craft your test case, then share at http://go/ask/sorbet, thanks!
class A
  def foo
    [
  end
end
```

#### Observed output

```
s(:begin)
editor.rb:5: unexpected token "end" https://srb.help/2001
     5 |  end
          ^^^
Errors: 1
```

#### Expected behavior

- - -

PR #5269, opened 2022-02-07
Slightly better handling of missing close brackets for arrays

aprocter:
This slightly improves our handling of missing close brackets for arrays.

Example:

```ruby
# typed: true

class A
  def f
    [1,
    # ^ error: unexpected token ","
    puts "hi"
  end
  def g
    [
  end
# ^^^ error: unexpected token "end"
  def h
    puts "ho"
  end
end

class B
  def q
    puts "ho"
    [
  end # error: unexpected token "end"
  def r
    puts "hi"
  end
  def s
    puts ([] + [)
              # ^ error: unexpected token ")"
  end
  def t
    puts "hi"
  end
end
```

Parse tree before this change:

```
(:begin)
```

Parse tree after this change:

```
s(:begin,
  s(:class,
    s(:const, nil, :A), nil,
    s(:begin,
      s(:def, :f, nil, nil),
      s(:def, :g, nil, nil),
      s(:def, :h, nil,
        s(:send, nil, :puts,
          s(:str, "ho"))))),
  s(:class,
    s(:const, nil, :B), nil,
    s(:begin,
      s(:def, :q, nil, nil),
      s(:def, :r, nil,
        s(:send, nil, :puts,
          s(:str, "hi"))),
      s(:def, :s, nil, nil),
      s(:def, :t, nil,
        s(:send, nil, :puts,
          s(:str, "hi"))))))
```

The change here is to pull the naked `error` production out of the rule for `args`, at the cost of more special cases at its use site. I'm not 100% sure why this works, but it seems to make the parser less greedy when rolling back tokens to find a production that reduces the `error`.

### Motivation
#5182 

### Test plan
See included automated tests.


aprocter:
Running this against the new test in #5270 suggests that this is probably regressing some things we actually care about but didn't have tests for. Will play around and see if I can adjust this approach to work with that test, but I'm not super optimistic.

aprocter:
Yeah, doesn't look like this approach is going to work. :( To make the test case in #5270 work, I tried fluffing up the rules for arrays in a similar fashion to what's done for `paren_args` (adding some rules like `tLBRACK error tCOMMA aref_args tRBRACK`) and that immediately causes the new test in _this_ PR to break down. Going to close this out and come back to this issue later.

PR #5271, opened 2022-02-07
Add to type_alias doc

jez:
### Test plan

See included automated tests.

### DIFF
--- a/website/docs/type-aliases.md
+++ b/website/docs/type-aliases.md
@@ -3,8 +3,6 @@ id: type-aliases
 title: Type Aliases
 ---
 
-> TODO: This page is still a fragment. Contributions welcome!
-
 ```ruby
 Alias = T.type_alias {Type}
 ```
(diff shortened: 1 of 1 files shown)


website/docs/type-aliases.md
```diff
@@ -113,4 +111,60 @@ def valid(x)
...
+  sig {returns(T.nilable(SelfReferential))}
+  attr_reader :val
+
+  sig {params(val: T.nilable(SelfReferential)).void}
+  def initialize(val); @val = val; end
+end
+```
+
+Unfortunate for the case of typing JSON, this generally leads to more verbosity
+than in other languages, but can still accomplish somethin similar:
```
elliottt, line 157:
```suggestion
Unfortunately for the case of typing JSON, this generally leads to more verbosity
than in other languages, but can still accomplish something similar:
```

elliottt (approved):

PR #5272, opened 2022-02-08
Add error recovery cases for binop and unop expressions

aprocter:
This adds error recovery cases in the parser for binop and unop expressions.

Parse tree for the included test before:
```
s(:class,
  s(:const, nil, :A), nil,
  s(:def, :g, nil, nil))
```

Parse tree for the included test after (abbreviated 'cause it's really long):
```
s(:class,
  s(:const, nil, :A), nil,
  s(:def, :g, nil,
    s(:begin,
     ...
      s(:send, nil, :puts,
        s(:begin,
          s(:irange,
            s(:int, "0"),
            s(:const, nil, :<ErrorNode>)))),
      s(:send, nil, :puts,
        s(:begin,
          s(:erange,
            s(:int, "1"),
            s(:const, nil, :<ErrorNode>)))),
...
```

### Motivation
Better error recovery for better IDE experience.

### Test plan
See included automated tests.


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1296,100 +1304,197 @@ lbrace_cmd_block_start:
...
+                    }
                 | arg tMATCH arg
                     {
                       $$ = driver.build.match_op(self, $1, $2, $3);
                       DIAGCHECK();
                     }
+                | arg tMATCH error
+                    {
+                      $$ = driver.build.match_op(self, $1, $2, driver.build.error_node(self, @2.end, @3.end));
+                      DIAGCHECK();
```
jez, line 1440: This DIAGCHECK macro mentions YYERROR. I don't really know what the ramifications of this are. Do you know? Is this ok to put in a clause that mentions `error`?

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1452,10 +1557,18 @@ lbrace_cmd_block_start:
...
+                      $$ = driver.build.binaryOp(self, $1, $2, driver.build.error_node(self, @2.end, @3.end));
+                    }
                 | rel_expr relop arg %prec tGT
                     {
                       $$ = driver.build.binaryOp(self, $1, $2, $3);
                     }
+                | rel_expr relop error %prec tGT
+                    {
+                      $$ = driver.build.binaryOp(self, $1, $2, driver.build.error_node(self, @2.end, @3.end));
+                    }
```
jez, line 1571: What are these rules? They don't look like the other cases.

test/testdata/parser/error_recovery/unop_binop.rb
```diff
@@ -0,0 +1,68 @@
+# typed: false
+
+class A
+  def g
+    t = T.let(true, T::Boolean)
+    f = T.let(false, T::Boolean)
+    puts (0 .. def) # error: unexpected token ")"
```
jez, line 7: What's the intuition for why we're testing `def` here? This doesn't look like something anyone would ever write—I'm wondering what the intuition is for including it.

jez (commented):
Some questions but this looks mostly good, thanks!

test/testdata/parser/error_recovery/unop_binop.rb.parse-tree.exp
```diff
@@ -0,0 +1,1399 @@
+Class {
```
jez, line 1: It's very hard for me to eyeball that this 1.4k line file captures the parse result that we want. It would be **much** easier to skim using `desugar-tree` instead of `parse-tree`, which turns it from a 1.4k line file into a very eyeball-able ~100 line file.

Doing this locally is how I found the bug I mention above relating to `&&` and `||`.

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1296,100 +1304,197 @@ lbrace_cmd_block_start:
...
                 | arg tANDOP arg
                     {
                       $$ = driver.build.logicalAnd(self, $1, $2, $3);
                     }
+                | arg tANDOP error
+                    {
+                      $$ = driver.build.binaryOp(self, $1, $2, driver.build.error_node(self, @2.end, @3.end));
```
jez, line 1488: This case and the case for `tOROP` are wrong. Note that successful parse results for `tANDOP` use `logicalAnd` and `logicalOr` from the builder.

When I printed the `desugar-tree` for your change, I noticed that it didn't desugar `&&` the way it should have.

Another thing to consider: add a third test case for each of the cases you added here to show what a successful parse result looks like in the exp file:

```ruby
    puts (t && A)
    puts (t && def) # error: unexpected token ")"
    puts (t &&) # error: unexpected token ")"
```

Doing this and using `desguar-tree` would have made catching this bug easier.

test/testdata/parser/error_recovery/unop_binop.rb
```diff
@@ -0,0 +1,68 @@
...
+    puts (24 > def) # error: unexpected token ")"
+    puts (24 >) # error: unexpected token ")"
+    puts (25 < def) # error: unexpected token ")"
+    puts (25 <) # error: unexpected token ")"
+    puts (26 >= def) # error: unexpected token ")"
+    puts (26 >=) # error: unexpected token ")"
+    puts (27 <= def) # error: unexpected token ")"
+    puts (27 <=) # error: unexpected token ")"
+  end
+end
```
jez, line 68: Can you add a test in `test/testdata/lsp/completion/` too? It doesn't have to cover all these cases, by any means, but it'd be nice to see a test for this:

```ruby
class A
  def foo(a, b, x, y)
    if x && 
      #     ^ completion: a, b, x, y, ...
    end
end
```

jez (changes requested):
^

parser/parser/cc/grammars/typedruby.ypp
aprocter: Nope, good question! Let me look into it.

parser/parser/cc/grammars/typedruby.ypp
aprocter: I'm not sure why the grammar treats relops separately from other binops. I think it might just be to assign different precedence to the relational operators?

test/testdata/parser/error_recovery/unop_binop.rb
jez: i was thinking it would work out of the box. if you want to look into why, feel free, otherwise commit the test and make a ticket?

parser/parser/cc/grammars/typedruby.ypp
aprocter: So I was looking into why the `DIAGCHECK` is here in the first place. I think it's because the original typedruby parser can produce a diagnostic when parsing the regexp: https://github.com/typedruby/typedruby/blob/0216fe987e1bfc55fce36f8c859d529aa9b8d870/parser/src/builder.rs#L286-L301 .

We don't actually implement this behavior, which I assume is why we wound up deleting the `InvalidRegexp` diagnostic in #5186. So presumably we don't technically need the original `DIAGCHECK` anyway. So I'm inclined to delete this one, even though I don't think mixing `error` with `YYERROR` is actually a problem.

aprocter: https://github.com/sorbet/sorbet/commit/792d36670e25289db5b705713796b19555b6a966 removes the `DIAGCHECK` and explains why.

parser/parser/cc/grammars/typedruby.ypp
aprocter: This should be fixed now. I re-scanned the `desugar-tree` and didn't see anything problematic.

test/testdata/parser/error_recovery/unop_binop.rb
aprocter: Added a test that expects the current behavior, with a note about what we really want.

jez (approved):

PR #5273, opened 2022-02-08
Better error recovery for unterminated arrays

aprocter:
Another attempt to address #5182. Still not sure it does everything we want, but unlike my last attempt (#5269) I _think_ it's unlikely to disrupt current desired behaviors.

The strategy this time is to identify a whole bunch of tokens that can't possibly occur right after `tLBRACK aref_args`, and add an erroring rule when one of them occurs. I found the list of tokens that works here largely by trial and error (let's throw some more tokens in and see if we get any SR/RR conflicts), but intuitively I think the list makes sense: mostly they're "bracket-closing" tokens like `)` and `end`.

Example:

```ruby
# typed: true

class A
  def f
    [1,
  #   ^ error: unexpected token ","
  # ^ error: unterminated "["
    puts "hi"
  end
  def g
    [
  # ^ error: unterminated "["
  end
  def h
    puts "ho"
  end
end

class B
  def q
    puts "ho"
    [
  # ^ error: unterminated "["
  end
  def r
    puts "hi"
  end
  def s
    puts ([] + [)
             # ^ error: unterminated "["
  end
  def t
    puts "hi"
  end
end
```

Parse tree and output before:
```
s(:begin)
test/testdata/parser/error_recovery/unterminated_array.rb:5: unexpected token "," https://srb.help/2001
     5 |    [1,
              ^

test/testdata/parser/error_recovery/unterminated_array.rb:29: unexpected token ")" https://srb.help/2001
    29 |    puts ([] + [)
                        ^
Errors: 2
```

Parse tree and output after:
```
s(:begin,
  s(:class,
    s(:const, nil, :A), nil,
    s(:begin,
      s(:def, :f, nil,
        s(:array,
          s(:const, nil, :<ErrorNode>))),
      s(:def, :g, nil,
        s(:array)),
      s(:def, :h, nil,
        s(:send, nil, :puts,
          s(:str, "ho"))))),
  s(:class,
    s(:const, nil, :B), nil,
    s(:begin,
      s(:def, :q, nil,
        s(:begin,
          s(:send, nil, :puts,
            s(:str, "ho")),
          s(:array))),
      s(:def, :r, nil,
        s(:send, nil, :puts,
          s(:str, "hi"))),
      s(:def, :s, nil,
        s(:send, nil, :puts,
          s(:begin,
            s(:send,
              s(:array), :+,
              s(:array))))),
      s(:def, :t, nil,
        s(:send, nil, :puts,
          s(:str, "hi"))))))
test/testdata/parser/error_recovery/unterminated_array.rb:5: unexpected token "," https://srb.help/2001
     5 |    [1,
              ^

test/testdata/parser/error_recovery/unterminated_array.rb:5: unterminated "[" https://srb.help/2001
     5 |    [1,
            ^

test/testdata/parser/error_recovery/unterminated_array.rb:11: unterminated "[" https://srb.help/2001
    11 |    [
            ^

test/testdata/parser/error_recovery/unterminated_array.rb:22: unterminated "[" https://srb.help/2001
    22 |    [
            ^

test/testdata/parser/error_recovery/unterminated_array.rb:29: unterminated "[" https://srb.help/2001
    29 |    puts ([] + [)
                       ^
Errors: 5
```

### Motivation
Better error recovery, better pizza. (pizza == IDE responsiveness)

### Test plan
See included automated tests.


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1715,6 +1731,12 @@ lbrace_cmd_block_start:
                     {
                       $$ = driver.build.array(self, $1, $2, $3);
                     }
+                | tLBRACK aref_args array_premature_end
+                    {
+                      $$ = driver.build.array(self, $1, $2, $3);
+                      driver.diagnostics.emplace_back(dlevel::ERROR, dclass::UnterminatedToken, diagnostic::range(@1.begin, @1.end), "\"[\"");
+                      driver.rewind_and_reset(@2.end);
```
jez, line 1738: I think that this use of `rewind_and_reset` is a little suspect. I think that if we're sold on this approach, the lexer state we want to be in once we've said that the recovery phase is done is `expr_beg`, because that's the state that the lexer would have transitioned to had it correctly seen  `]` here.

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1653,6 +1654,21 @@ lbrace_cmd_block_start:
...
+                   | kENSURE
+                   | kEND
+                   | kTHEN
+                   | kELSIF
+                   | kELSE
+                   | kWHEN
+                   | kIN
+                   | tRPAREN
+                   | tRCURLY
+                   | tCOLON
```
jez, line 1670: This seems a little suspicious. I'm worried that this is going to be brittle / cause confusion when the upstream ruby grammar changes in the future in such a way that makes one of these token somehow expected, or if new tokens are added to the lexer but not added  here.

test/testdata/parser/error_recovery/unterminated_array.rb
```diff
@@ -0,0 +1,35 @@
...
+    puts "hi"
+  end
+  def s
+    puts ([] + [)
+             # ^ error: unterminated "["
+  end
+  def t
+    puts "hi"
+  end
+end
```
jez, line 35: Given the implementation above, I'm a little worried that this only does well on a handful of test cases. Some test cases I've found that don't look that great:

```ruby
# typed: true

class A
  X = [1,

  sig {void}
  def bar
  end
end
```

- drops the `1` from the array
- drops the whole sig + method def

```ruby
# typed: true

class A
  def bar
    puts 'before'
    x = [1,
    puts 'after'
  end
end
```

- again drops the `1` from the array
- drops the `puts 'after'` from the method body

test/testdata/parser/error_recovery/unterminated_array.rb.parse-tree.exp
```diff
@@ -0,0 +1,133 @@
+Begin {
```
jez, line 1: I noticed that in this PR as well as #5272 you used `.parse-tree.exp`. I've been using `desugar-tree` for all my parser recovery tests because I find them much easier to tell at a glance whether they're correct. For example, it's easy for me to tell whether something is in one method body or has been pushed outside it.

I'm curious: do you have a strong preference for `parse-tree` tests? I'd be curious to hear more, or if it's not a strong preference I'd love to see desugar-tree tests be the default here.

jez (commented):
This is definitely a clever idea, and one I've toyed with before as well!

But ultimately, I really, really think that the final solution for unmatched `[` brackets is going to involve using the indentation-aware techniques that #5000 is going to introduce.

I have a hunch that we can substantially improve on the results that this PR achieves if we use that approach instead. Do you want to chat more about this?

test/testdata/parser/error_recovery/unterminated_array.rb.parse-tree.exp
aprocter: All the tests in the `error_recovery` dir use `parse-tree`, so I was just following the pattern here and in #5272.

parser/parser/cc/grammars/typedruby.ypp
aprocter: I think there are two potential maintenance headaches here:

1. The grammar changes in the future, such that one of these tokens is now expected. In this case, it would manifest as a shift/reduce conflict which bison should flag for us, and it probably wouldn't take too long to work back from there to the fact that this production is implicated. In that case, we'd have to trim one of the tokens from this rule and it is possible that some recoverable parses would become unrecoverable, or at least not recover as gracefully.
2. The grammar changes in the future, such that new tokens are added that are not expected here. In this case, we wouldn't get any guidance from bison, but I don't think this is such a big deal, since it won't result in incorrect parses of valid syntax---it might just take us a while to realize that there are new tokens that need to be added here to improve the quality of error recovery.

Neither one of these is ideal, but I'm not sure they're showstoppers.

parser/parser/cc/grammars/typedruby.ypp
aprocter: Hmm, yeah, `beg` does make more sense here! I'll try switching to that. (Maybe it'll make some of the less-than-ideal parses you identified better---who knows?)

aprocter:
> But ultimately, I really, really think that the final solution for unmatched [ brackets is going to involve using the indentation-aware techniques that https://github.com/sorbet/sorbet/pull/5000 is going to introduce.
>
> I have a hunch that we can substantially improve on the results that this PR achieves if we use that approach instead. Do you want to chat more about this?

Sure! Let's chat about it today. If you think the indentation-aware techniques are going to handle things better, there's no need for redundancy here.

jez (approved):
my idea for how to fix this didn't pan out, and in the interest of not blocking things exclusively for FUD without concrete founding i figure we should go ahead with this change. it at least is better than nothing (but we acknowledge that it might be the sort of thing that we want or need to roll back some time in the future).

parser/parser/cc/grammars/typedruby.ypp
aprocter: I think `expr_end` might be the right state, actually...

https://github.com/sorbet/sorbet/blob/f5d8393c630cf05c158d55fb3dfabbd9d5a0d628/parser/parser/cc/lexer.rl#L2824-L2839

This is consistent with what the Ruby Hacking Guide has to say (https://whitequark.org/blog//2013/04/01/ruby-hacking-guide-ch-11-finite-state-lexer/). Whatcha think?

aprocter:
@jez: I switched from `parse-tree` to `desugar-tree`, and commented on why I think expr_end is probably the right state to end up in when we rewind, so figured I'd ask for another look when you get a chance.

aprocter:
:facepalm: Sorry, forgot to actually push :)

jez (approved):
Seems still not perfect but better than nothing.

https://github.com/sorbet/sorbet/pull/5273#discussion_r801280531

^ can you add these tests I suggested? i'd like to at least record the state of the world

aprocter:
> can you add these tests I suggested? i'd like to at least record the state of the world

Will do!

PR #5276, opened 2022-02-09
Add proper type for `Proc.new`

jez:
### Motivation

It looks like this change was accidentally made a long time ago.
Whenever there's an explicit override of `BasicObject.new`, it clobbers Sorbet's
ability to automatically infer the type of `.new` (because it can't know how
that override will necessarily behave--imagine that maybe someone has redefined
`new` to return `nil` in some cases).

There's also the question of docs. In this case, the override was likely added
to be able to show docs.

We can solve for both of these problems by just writing a better `sig` for
`Proc.new`.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/proc.rbi
+++ b/rbi/core/proc.rbi
@@ -361,7 +361,8 @@ class Proc < Object
   # proc = proc_from { "hello" }
   # proc.call   #=> "hello"
   # ```
-  def self.new(*_, &blk); end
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_L7Jg663dtSVgBy
→ https://go/builds/bui_L7JgUZF4hNy66m

elliottt (approved):

PR #5277, opened 2022-02-09
Error recovery for case

jez:
### Motivation

Sorry, I didn't go through the work of making this one into small commits.

### Test plan

See included automated tests.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -614,8 +614,25 @@ public:
 
     unique_ptr<Node> case_(const token *case_, unique_ptr<Node> expr, sorbet::parser::NodeVec whenBodies,
                            const token *elseTok, unique_ptr<Node> elseBody, const token *end) {
+        auto loc = tokLoc(case_);
+        if (end != nullptr) {
+            loc = loc.join(tokLoc(end));
+        }
         return make_unique<Case>(tokLoc(case_).join(tokLoc(end)), std::move(expr), std::move(whenBodies),
                                  std::move(elseBody));
+    }
+
+    unique_ptr<Node> case_error(const token *case_, unique_ptr<Node> cond, const token *end) {
+        auto loc = tokLoc(case_);
+        if (end != nullptr) {
+            loc = loc.join(tokLoc(end));
+        }
+        auto zloc = loc.copyWithZeroLength();
+        auto whenPatterns = NodeVec{};
+        whenPatterns.emplace_back(error_node(loc.beginPos(), loc.beginPos()));
+        auto whens = NodeVec{};
+        whens.emplace_back(make_unique<When>(zloc, std::move(whenPatterns), nullptr));
+        return make_unique<Case>(loc, std::move(cond), std::move(whens), nullptr);
     }
 
     unique_ptr<Node> case_match(const token *case_, unique_ptr<Node> expr, sorbet::parser::NodeVec inBodies,
@@ -1869,6 +1886,11 @@ ForeignPtr case_(SelfPtr builder, const token *case_, ForeignPtr expr, const nod
                                          build->cast_node(elseBody), end));
 }
 
+ForeignPtr case_error(SelfPtr builder, const token *case_, ForeignPtr cond, const token *end) {
+    auto build = cast_builder(builder);
+    return build->toForeign(build->case_error(case_, build->cast_node(cond), end));
+}
+
 ForeignPtr case_match(SelfPtr builder, const token *case_, ForeignPtr expr, const node_list *inBodies,
                       const token *elseTok, ForeignPtr elseBody, const token *end) {
     auto build = cast_builder(builder);
@@ -2532,6 +2554,7 @@ struct ruby_parser::builder Builder::interface = {
     call_method,
     call_method_error,
     case_,
+    case_error,
     case_match,
     character,
     complex,
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -614,10 +614,6 @@ public:
 
     unique_ptr<Node> case_(const token *case_, unique_ptr<Node> expr, sorbet::parser::NodeVec whenBodies,
                            const token *elseTok, unique_ptr<Node> elseBody, const token *end) {
-        auto loc = tokLoc(case_);
-        if (end != nullptr) {
-            loc = loc.join(tokLoc(end));
-        }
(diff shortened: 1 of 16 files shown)


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1865,15 +1866,40 @@ lbrace_cmd_block_start:
                         &case_body->whens,
                         else_ ? else_->tok : nullptr,
 			                  else_ ? else_->nod : nullptr, $4);
+                      driver.rewind_if_dedented($1, $4);
+                    }
+                | kCASE expr_value opt_terms kEND
```
froydnj, line 1871: Can we format this so that we have whitespace where `case_body` would go, so it's more obvious how this alternative differs from existing code?

```suggestion
                | kCASE expr_value opt_terms           kEND
```

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1865,15 +1866,40 @@ lbrace_cmd_block_start:
...
+                    {
+                      $$ = driver.build.case_error(self, $1, $2, $4);
+                      driver.diagnostics.emplace_back(dlevel::ERROR, dclass::EmptyCase, $1, driver.token_name($1->type()));
+                      // This is an `error` rule in disguise, so we can actually use indentation to rewind
+                      // even if we aren't in indentationAware mode.
+                      driver.rewind_if_dedented($1, $4, /*force*/true);
+                    }
+                | kCASE opt_terms kEND
```
froydnj, line 1879: Same request here for the formatting.

```suggestion
                | kCASE            opt_terms           kEND

```

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1865,15 +1866,40 @@ lbrace_cmd_block_start:
...
-                 | kCASE expr_value opt_terms p_case_body kEND
+                | kCASE expr_value opt_terms p_case_body kEND
                     {
                       auto &case_body = $4;
                       auto &else_ = case_body->els;
                       $$ = driver.build.case_match(self, $1, $2,
                         &case_body->whens,
                         else_ ? else_->tok : nullptr,
 			                  else_ ? else_->nod : nullptr, $5);
+                      driver.rewind_if_dedented($1, $5);
```
froydnj, line 1902: It's not obvious to me why we need to alter the behavior of the pattern-matching `case` here, since this was existing non-error code.  Shouldn't this be removed?

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1865,15 +1866,40 @@ lbrace_cmd_block_start:
...
+                      // even if we aren't in indentationAware mode.
+                      driver.rewind_if_dedented($1, $3, /*force*/true);
+                    }
+                | kCASE error
+                    {
+                      auto cond = driver.build.error_node(self, @1.end, @2.end);
+                      $$ = driver.build.case_error(self, $1, cond, nullptr);
+                      driver.replace_last_diagnostic(dlevel::ERROR, dclass::UnexpectedToken, $1, driver.token_name($1->type()));
                     }
-                 | kCASE expr_value opt_terms p_case_body kEND
```
froydnj, line 1869: This change seems unnecessary.

parser/Builder.cc
```diff
@@ -614,10 +614,27 @@ class Builder::Impl {
 
     unique_ptr<Node> case_(const token *case_, unique_ptr<Node> expr, sorbet::parser::NodeVec whenBodies,
                            const token *elseTok, unique_ptr<Node> elseBody, const token *end) {
+        auto loc = tokLoc(case_);
+        if (end != nullptr) {
+            loc = loc.join(tokLoc(end));
+        }
         return make_unique<Case>(tokLoc(case_).join(tokLoc(end)), std::move(expr), std::move(whenBodies),
```
froydnj, line 621: Surely this `loc` addition would be better if it was used in the `make_unique` call?  Or was this an experiment that turned out to be unnecessary?

(Slightly concerned that tests didn't seem to exercise the `end == nullptr` case, too, but maybe that is just a consequence of this being dead code.)

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1856,6 +1856,7 @@ lbrace_cmd_block_start:
                         &case_body->whens,
                         else_ ? else_->tok : nullptr,
 			                  else_ ? else_->nod : nullptr, $5);
+                      driver.rewind_if_dedented($1, $5);
```
froydnj, line 1859: Why are we rewinding in existing productions?

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1865,15 +1866,40 @@ lbrace_cmd_block_start:
                         &case_body->whens,
                         else_ ? else_->tok : nullptr,
 			                  else_ ? else_->nod : nullptr, $4);
+                      driver.rewind_if_dedented($1, $4);
```
froydnj, line 1869: Why are we rewinding in existing productions?

froydnj (changes requested):
I think this is reasonable, but I have concerns about:

1. `rewind_if_dedented` now appearing in rules that aren't actually errors, which seems like scope creep.
2. various things that don't seem to be used or changes that appear spurious.

Will stamp once those concerns are cleared up.

parser/parser/cc/grammars/typedruby.ypp
jez: It was the only production in this file that didn’t have its pipe character aligned so I fixed it.

parser/Builder.cc
jez: Yeah, I started trying to reuse the `case_` builder for the error handling cases, and then at some point realized that it wouldn't be needed. This code is dead.

jez:
I've fixed the small changes. Thanks for pointing those out.

To answer all the "Why are we rewinding in existing productions?" questions:

That is the whole point of the change. Calling `rewind_if_dedented` in those cases:

- will only do anything if `driver.indentationAware` is set, which is only set when we're doing a second parse to attempt better error recovery
- is crucial to make every test that I added in this PR pass—none of them pass without this change. Every production rule that I've modified in this PR has at least one, but usually many more, tests.

froydnj (approved):

jez:

→ https://go/builds/bui_L7diw9SiuA6mWC
→ https://go/builds/bui_L7di9yer8NQ7mG

PR #5278, opened 2022-02-09
Add --max-errors flag

aprocter:
Add a `--max-errors=N` flag, to suppress error message output beyond the first `N` errors.

### Motivation
When running Sorbet from user-facing tools it's sometimes useful to suppress excess error messages. (We've run into a case at Stripe where this might be useful.)

### Test plan
See included automated tests.


jez (changes requested):
I prefer to avoid adding this if at all possible. I think we should solve our current problem by post-processing the error output log.

aprocter:
Closing in favor of a bespoke solution that post-processes the output log.

PR #5287, opened 2022-02-11
Fix return type for `Module::alias_method` to return a Symbol

Morriar:
### Motivation

It seems this method actually [returns a symbol](https://github.com/ruby/ruby/blob/e89d80702bd98a8276243a7fcaa2a158b3bfb659/vm_method.c#L2190):

```
$ irb
> class Foo
>   alias_method(:foo, :to_s)
> end
=> :foo
```

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/module.rbi
+++ b/rbi/core/module.rbi
@@ -212,7 +212,7 @@ class Module < Object
         new_name: Symbol,
         old_name: Symbol,
     )
-    .returns(T.self_type)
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_L8LE63SdPEZEc2
→ https://go/builds/bui_L8LE237npfAyuj

jez (approved):

PR #5288, opened 2022-02-11
Improve LSP messaging for ignored files

mutecipher:
### Motivation

Currently, files that are ignored by Sorbet simply display a message of
the form:

```
[Error - 2:26:36 PM] Request textDocument/hover failed.
  Message: Did not find file at uri file:///path/to/file in textDocument/hover
  Code: -32602
```

which isn't very helpful when trying to determine the cause of the error
itself. Instead, I've updated the messaging for _ignored_ files
specifically along with a more appropriate LSPErrorCode.

```
[Error - 2:26:36 PM] Request textDocument/hover failed.
  Message: Ignored file at uri file:///path/to/file in textDocument/hover
  Code: -32803
```

this is more descriptive and provides the `LSPErrorCodes::RequestFailed`
error which is more descriptive of what occurred.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPTask.cc
+++ b/main/lsp/LSPTask.cc
@@ -252,7 +252,7 @@ LSPQueryResult LSPTask::queryByLoc(LSPTypecheckerDelegate &typechecker, string_v
 
     if (!fref.exists() && config.isFileIgnored(config.remoteName2Local(uri))) {
         auto error = make_unique<ResponseError>(
-            (int)LSPErrorCodes::RequestFailed,
+            (int)LSPErrorCodes::InvalidParams,
             fmt::format("Ignored file at uri {} in {}", uri, convertLSPMethodToString(forMethod)));
         return LSPQueryResult{{}, move(error)};
     }
--- a/main/lsp/LSPTask.cc
+++ b/main/lsp/LSPTask.cc
@@ -249,6 +249,14 @@ LSPQueryResult LSPTask::queryByLoc(LSPTypecheckerDelegate &typechecker, string_v
     Timer timeit(config.logger, "setupLSPQueryByLoc");
     const core::GlobalState &gs = typechecker.state();
     auto fref = config.uri2FileRef(gs, uri);
+
+    if (!fref.exists() && config.isFileIgnored(config.remoteName2Local(uri))) {
+        auto error = make_unique<ResponseError>(
+            (int)LSPErrorCodes::RequestFailed,
+            fmt::format("Ignored file at uri {} in {}", uri, convertLSPMethodToString(forMethod)));
+        return LSPQueryResult{{}, move(error)};
+    }
+
     if (!fref.exists()) {
         auto error = make_unique<ResponseError>(
             (int)LSPErrorCodes::InvalidParams,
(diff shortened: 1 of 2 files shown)


froydnj:
Thanks for making this clearer!  Do you happen to have a screenshot of what this looks like inside VSCode?

mutecipher:
> Thanks for making this clearer! Do you happen to have a screenshot of what this looks like inside VSCode?

Here you go. Attached a demo video on a simple example!

https://user-images.githubusercontent.com/2898615/153676982-c8ad45aa-3c11-47f4-94c6-c8c96a7e7f6c.mov

mutecipher:
**Existing behaviour**

https://user-images.githubusercontent.com/2898615/153677178-30ba8ffc-8a4a-42de-a27e-641cbaa3beb0.mov

**New behaviour**

https://user-images.githubusercontent.com/2898615/153677249-aad9aefb-0f4e-4224-b787-92d41cf6aef4.mov

main/lsp/LSPTask.cc
```diff
@@ -249,6 +249,14 @@ LSPQueryResult LSPTask::queryByLoc(LSPTypecheckerDelegate &typechecker, string_v
     Timer timeit(config.logger, "setupLSPQueryByLoc");
     const core::GlobalState &gs = typechecker.state();
     auto fref = config.uri2FileRef(gs, uri);
+
+    if (!fref.exists() && config.isFileIgnored(config.remoteName2Local(uri))) {
+        auto error = make_unique<ResponseError>(
+            (int)LSPErrorCodes::RequestFailed,
```
jez, line 255:
```suggestion
            (int)LSPErrorCodes::InvalidRequest,
```

i think it's better to use an existing error code that's already defined by the specification than to add a new one

froydnj: It is defined in the upcoming 3.17 version of the spec:

https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#errorCodes

mutecipher: Yeah, my reasoning behind that code was the description of the error matches what was occurring here.

```javascript
        /**
	 * A request failed but it was syntactically correct, e.g the
	 * method name was known and the parameters were valid. The error
	 * message should contain human readable information about why
	 * the request failed.
	 *
	 * @since 3.17.0
	 */
	export const RequestFailed: [integer](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#integer) = -32803;
```

jez: ah gotcha, thanks!

jez: So it's not exactly specified, but my take on `RequestFailed` is that it means more like "we ran into an internal error, caught the exception, and are showing you this message"

In this case, the failure is in fact purely a function of the parameters.

I hold my opinion less strongly knowing that at least the newly added error code is in the spec, but I still think there's ground to use `InvalidParams` here

mutecipher: Sure, that's fine with me. My primary concern was around the error message being displayed rather than the error code itself. I'll push up the change. 👍🏻

jez:

→ https://go/builds/bui_L9s7XWJVNkExQH
→ https://go/builds/bui_L9s79fiSbXTfgp

jez (approved):

PR #5289, opened 2022-02-11
give more tokens human-readable names

froydnj:
Due to limitations in Bison, we have some places where internal, Bison-specific names leak out into human-readable error messages.  This is not good.  Fortunately, we also have a mechanism in place to translate those Bison-specific names into human-readable names; let's extend the cases that we handle so people get nice error messages.

### Motivation

More humane error messages.

### Test plan

See included automated tests.


aprocter (approved):
Makes sense to me!

PR #5291, opened 2022-02-12
Switch to FileRef::dataAllowingUnsafe when finding packages for files

elliottt:
`PackageDB::packageForFile` only looks at the path of the file, and as a result would be fine to use on reserved but not read files. This PR makes that change by switching the call to `FileRef::data` to `FileRef::dataAllowingUnsafe`.

The only case where this could be a problem is if the `FileRef` given refers to an unread package file. However, I believe that case won't show up for the following reason: if we're looking for package info for a file, the package database will already have been loaded by a previous step. In the case of the packager pass, all `__package.rb` files will be read during indexing, and in the case of rbi generation, we explicitly index all `__package.rb` files before running anything else through the pipeline.

### Motivation

Allow package membership to be queried for files, before the files have been read. This allows us to filter down the files that we run through the pipeline based on package membership, significantly improving the time to generate a single package's rbi.

### Test plan

See included automated tests.

### DIFF
--- a/core/packages/PackageDB.cc
+++ b/core/packages/PackageDB.cc
@@ -138,7 +138,7 @@ NameRef PackageDB::lookupPackage(NameRef pkgMangledName) const {
 
 const PackageInfo &PackageDB::getPackageForFile(const core::GlobalState &gs, core::FileRef file) const {
     ENFORCE(frozen);
-    auto &fileData = file.data(gs);
+    auto &fileData = file.dataAllowingUnsafe(gs);
     string_view path = fileData.path();
     int curPrefixPos = path.find_last_of('/');
     while (curPrefixPos > 0) {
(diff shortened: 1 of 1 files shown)


ngroman (approved):
👍

core/packages/PackageDB.cc
```diff
@@ -138,7 +138,7 @@ NameRef PackageDB::lookupPackage(NameRef pkgMangledName) const {
 
 const PackageInfo &PackageDB::getPackageForFile(const core::GlobalState &gs, core::FileRef file) const {
     ENFORCE(frozen);
-    auto &fileData = file.data(gs);
+    auto &fileData = file.dataAllowingUnsafe(gs);
```
froydnj, line 141: What do you think about dropping a rationale for using `dataAllowingUsafe` here, something as short as "We're only looking at the filename, which is always available.  See https://github.com/sorbet/sorbet/pull/5291 for more details."?

froydnj (approved):

core/packages/PackageDB.cc
elliottt: Seems pretty reasonable to me!

PR #5295, opened 2022-02-14
Remove core::File::Type::Package in favor of a flag

elliottt:
The `sourceType` field on a File is currently used to indicate both what kind of file this is, and whether or not it's been read. In the case of Stripe packages, it's not necessary to read the file to know that it's a package file; looking at the file path is enough.

This PR removes the `core::File::Type::Package` enum entry, and moves its functionality to an entry in `File::Flags` instead.

### Motivation

Allowing package files to be identified without reading their contents.

### Test plan

See included automated tests.


elliottt:

→ https://go/builds/bui_L9VIVam94FwjpA
→ https://go/builds/bui_L9VIZRNVAhRbW0

core/Files.cc
```diff
@@ -174,8 +174,18 @@ bool isPackageRBIPath(string_view path) {
...
 }
 
+bool isPackagePath(string_view path) {
+    auto pos = path.rfind("/");
+    if (pos != string_view::npos) {
+        path = path.substr(pos + 1);
+    }
+
+    return path == "__package.rb";
+}
```
jez, line 184: We used to only set `sourceType = Type::Package` in `packager.cc`, which would have only been run if the `--stripe-packages` flag had been passed at the CLI.

Do you think it's possible to thread that through here?

(It would be an ever so slight backwards-incompatible change in that if your project happens to have a `__package.rb` file already.)

jez (approved):

core/Files.cc
elliottt: I think it's the case already that any path that cares about `isPackage` would be gated by `--stripe-packages` already, right? This change makes us even less likely to accidentally treat a package file specially, as they now all have the sourceType `File::Type::Normal`.

jez: Ah good point. Thanks!

elliottt: Separately, I don't think it would be tough to thread that through to this point, so maybe to guard against a possible foot-gun I'll add that.

jez: up to you—change looks good either way

elliottt: I take it back -- I thought we had a boolean in `GlobalState` for signaling when `--stripe-packages` is enabled and we don't. Unfortunately we'd probably need to add a flag like that to `GlobalState` to make this easy.

main/lsp/LSPTask.cc
```diff
@@ -293,8 +293,8 @@ LSPQueryResult LSPTask::queryBySymbol(LSPTypecheckerDelegate &typechecker, core:
         const auto &usedSymbolNames = hash.usages.symbols;
         auto ref = core::FileRef(i);
 
-        const bool fileIsValid = ref.exists() && (ref.data(gs).sourceType == core::File::Type::Normal ||
-                                                  ref.data(gs).sourceType == core::File::Type::Package);
+        const bool fileIsValid =
+            ref.exists() && (ref.data(gs).sourceType == core::File::Type::Normal || ref.data(gs).isPackage());
```
aisamanra, line 297: Packages will always be `File::Type::Normal` now, right? Is there ever a situation in which the second half of this or will get evaluated?

aisamanra (approved):

main/lsp/LSPTask.cc
elliottt: Yeah, that's a really good point. I'll remove the `isPackage` call, as it's not contributing anything, and the intent of the previous check was that this wasn't any of the other cases for that enum.

core/Files.cc
elliottt: Looking through the changes to LSP, I think this might handle `__package.rb` files differently even without `--stripe-packages`.  I'm going to see if I can convince myself otherwise before merging this.

elliottt: I've guarded the uses of `isPackage()` in LSP with `this->config.opts.stripePackages`, which I think addresses any issues that would have arisen from this change.

PR #5297, opened 2022-02-14
avoid allocating strings for sorting package module children

froydnj:
We can go faster if we just sort the `shortName`s of the `NameRef`s that we're dealing with here.  In the rare case that they compare equal, we can stabilize the sort by using their full name.

### Motivation

`++speed;`  This speeds up the workers part of the package pass on Stripe's codebase by about 20%.

### Test plan

See included automated tests.


elliottt (approved):
Seems reasonable to me, but I'd be curious to see what fails if we instead only compare the `shortName`.

froydnj:

→ https://go/builds/bui_L9VX2i1UchGRuP
→ https://go/builds/bui_L9VXac3WvKjCHC

froydnj:
> Seems reasonable to me, but I'd be curious to see what fails if we instead only compare the `shortName`.

This seems to work just fine on Stripe's codebase!

PR #5299, opened 2022-02-14
Fix crash in `textDocument/typeDefinition`

jez:
### Motivation

I noticed this showing up in some crash logs at Stripe.

### Test plan

See included automated tests.

Some of the cases I couldn't find test cases for. I've implemented them so that
they won't crash anymore, but it would be nice to have tests for them.

### DIFF
--- a/main/lsp/requests/type_definition.cc
+++ b/main/lsp/requests/type_definition.cc
@@ -40,6 +40,43 @@ vector<core::Loc> locsForType(const core::GlobalState &gs, const core::TypePtr &
             for (auto loc : locsForType(gs, t.right)) {
                 result.emplace_back(loc);
             }
+        },
+        [&](const core::SelfTypeParam &s) {
+            for (auto loc : s.definition.locs(gs)) {
+                result.emplace_back(loc);
+            }
+        },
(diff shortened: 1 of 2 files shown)


froydnj (approved):
I am willing to believe that this is better, but it is not clear to me why this fixes crashes.  Does some later part of the code believe that certain types must have locations returned by this query?  Returning no locations for everything that's not handled by the `typecase`, as is currently done, seems reasonable...but apparently that's not the case?

jez:
> I am willing to believe that this is better, but it is not clear to me why this fixes crashes. Does some later part of the code believe that certain types must have locations returned by this query? Returning no locations for everything that's not handled by the `typecase`, as is currently done, seems reasonable...but apparently that's not the case?

idk if this was always the case or if this was only the behavior implemented once we started using `TypePtr`, but it looks like the default case of a `typecase` on `TypePtr` throws. This is the behavior of the test added here on master:

image

jez:
> idk if this was always the case

It looks like that behavior predates john's TypePtr changes.

froydnj:
Ah right.  Sure would be nice to get a static error for that.

PR #5300, opened 2022-02-14
Show user-meaningful token names in rest of errors

jez:
### Motivation

We mostly show user-meaningful token names by way of Bison's `%token` directive,
but some of our tokens come from identical source strings (depending on lexer
state, they'll get parsed one way or another).

To make up for that, we use this helper, which previously only had a handful of
the tokens in it. This adds basically all the rest of them.

@froydnj requested this as a follow-up to a previous PR of mine.

### Test plan

See included automated tests.


elliottt (approved):
👍

jez:
I hadn't pulled. Nathan already implemented this.

PR #5301, opened 2022-02-14
Fix crash in canTakeFastPath

jez:
### Motivation

We intentionally do not walk v->ancestors nor v->singletonAncestors.

These lists used to be guaranteed to be simple trees (only constant literals) by desugar, but that was later relaxed. Rather than changing the whole TreeMap data structure and affect all treemaps, I've decided to just change the TreeMap that would like to depend on having walked the ancestors list.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/fast_path/crash_local_vars.1.rbupdate
+++ b/test/testdata/lsp/fast_path/crash_local_vars.1.rbupdate
@@ -2,9 +2,10 @@
 
 class Outer
   describe 'client' do
+    #               ^^ error: Hint: this kDO_BLOCK token might not be properly closed
     class Inner <
-    it 'foo' do
+    it 'foo' do # error: Superclasses must only contain constant literals
       1.times {|x| puts(x)}
     end
   end
-end
+end # error: unexpected token "end of file"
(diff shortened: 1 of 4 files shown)


test/testdata/lsp/fast_path/crash_local_vars.1.rbupdate
```diff
@@ -0,0 +1,11 @@
...
+
+class Outer
+  describe 'client' do
+    #               ^^ error: Hint: this kDO_BLOCK token might not be properly closed
+    class Inner <
+    it 'foo' do # error: Superclasses must only contain constant literals
+      1.times {|x| puts(x)}
+    end
+  end
+end # error: unexpected token "end of file"
```
jez, line 11: > Jake, why is this a fast_path test? It looks like a crash that has nothing to do with LSP.

So the thing is: the crash only shows up in `Substitute.cc`. That only fires when there are multiple GlobalState's to merge, but we don't have that in PosTests because we intentionally limit the execution to be single threaded.

There is a second way to trigger `Substitute.cc`, even if the test is single threaded, which is by making a test that calls `commitEdit`, because then Sorbet's LSP will try to see if the edit can take the fast path or not by running index + substutition + hashing.

This is also the code path the bug manifested in the wild, but that's just because most people don't tend to try to commit code where a class has a method + block call as an ancestor in a file that parses.

aprocter (approved):

PR #5306, opened 2022-02-15
Allow `Module#===` on generic

jez:
### Motivation

There's technically no reason why you shouldn't be allowed to do

```
case x
when Integer
end
```

if `x` is a generic type_parameter.

It's a slightly weird signature. An alternative would be to use `T.untyped`, but I figured I'd avoid using `T.untyped`. Writing it this way is basically the only syntax I know to ask Sorbet to type an argument as having type `core::Types::top()` (because `BasicObject` is not the true top type—there are methods on `BasicObject` but no methods on `<top>`).

### Test plan

See included automated tests.


elliottt (approved):
This makes sense to me, good catch!

PR #5307, opened 2022-02-15
Print `T.type_parameter` like it actually appears

jez:
### Motivation

Better error messages

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/infer/generics/type_param_is_a.rb
+++ b/test/testdata/infer/generics/type_param_is_a.rb
@@ -14,10 +14,10 @@ class Test
   end
   def test(x)
     if x.is_a? Integer
-      T.reveal_type(x) # error: Revealed type: `T.all(Integer, Test#test#T)`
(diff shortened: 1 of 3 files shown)


test/testdata/infer/generics/type_param_is_a.rb
```diff
@@ -14,10 +14,10 @@ def take_integer(x)
   end
   def test(x)
     if x.is_a? Integer
-      T.reveal_type(x) # error: Revealed type: `T.all(Integer, Test#test#T)`
+      T.reveal_type(x) # error: Revealed type: `T.all(Integer, T.type_parameter(:T) (of Test#test))`
```
jez, line 17: I opted to implement this like `T.attached_class`, which makes the awkward `(of ...)` show up in the middle of the `T.all`.

We could change that to be more like how `[] (0-tuple)` works if we want, but it would be more code, so I'm inclined to leave it at this.

froydnj (approved):

PR #5309, opened 2022-02-15
[LSP] Extract method to module refactoring

ilyailya:
The PR adds a new refactoring code action "Extract method to the new module".
The refactoring cuts out the method code piece and puts it into a new module with a unique name on top of the file, and update all call sites for a refactored method. 
The code action can produce incorrect code and it's up to the user to fix newly emerging Sorbet errors 

### Motivation

New refactoring!

### Test plan

See included automated tests.


sig_finder/sig_finder.cc
```diff
@@ -0,0 +1,119 @@
+#include "absl/strings/ascii.h"
```
jez, line 1: small thing: what do you think about putting this in `main/lsp` where the other things like `NextMethodFinder` are?

iz-stripe: The code is copied from @froydnj PR https://github.com/sorbet/sorbet/pull/5169, and it's used outside of `lsp`. I haven't updated the Nathan's call site yet, but will do soon. That might be blocking moving it into the `lsp` package

main/lsp/tools/make_lsp_types.cc
```diff
@@ -1422,7 +1424,7 @@ void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_pt
             {"textDocument/formatting", makeVariant({JSONNull, makeArray(TextEdit)})},
             // (CodeAction | Command)[] | null
             // Sorbet only sends CodeAction[].
-            {"textDocument/codeAction", makeVariant({JSONNull, makeArray(CodeAction)})},
+            {"textDocument/codeAction", makeVariant({JSONNull, makeArray(CodeAction), makeArray(Command)})},
```
jez, line 1427: The comment above and the LSP docs make it seem like the type should be

```cpp
makeVariant({JSONNull, makeArray(makeVariant({CodeAction, Command}))})
```

btw

main/lsp/tools/make_lsp_types.cc
```diff
@@ -56,6 +56,7 @@ shared_ptr<JSONType> makeArray(shared_ptr<JSONType> type) {
 
 void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_ptr<JSONObjectType>> &classTypes) {
     // Singletons
+    shared_ptr<JSONType> JSONAny = make_shared<JSONObjectType>();
```
jez, line 59: do you mind adding a comment here that mentions that the LSP spec actually allows more than just object types here? (it allows anything, including arrays, strings, numbers, and null)

main/lsp/tools/make_lsp_types.cc
ilyailya: That was my first thought too! But turns out you can't have two json objects in the same union with our current setup:
https://github.com/sorbet/sorbet/blob/dd0e616ca5457e2b0c6e22f665e16e52022590bb/main/lsp/tools/generate_lsp_messages.h#L948-L951

Currently that version fails to compile, because `CodeAction` and `Command`  have the same `CPPBaseKind` — `BaseKind::ObjectKind`. 

I think there is a way around this, the `CodeAction` class has a optional `command` field, we can use that. I'm testing this hypothesis rn

main/lsp/tools/make_lsp_types.cc
ilyailya: I haven't pushed that yet, but I've declared `JSONAny` as
```
    // In teh LSP specification JSONAny defined as
    // `export type LSPAny = LSPObject | LSPArray | string | integer | uinteger | decimal | boolean | null;`
    // https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#number
    //
    // We don't support LSPObject and LSPArray options yet
    shared_ptr<JSONType> JSONAny = makeVariant({JSONNull, JSONBool, JSONInt, JSONDouble, JSONString});
```

sig_finder/sig_finder.cc
jez: sounds good thanks, i didn't see that usage. i still think we probably don't want to make a new top-level folder for this. maybe something like `ast/sig_finder/`

test/helpers/position_assertions.cc
```diff
@@ -40,7 +40,7 @@ const UnorderedMap<
         {"enable-packager", BooleanPropertyAssertion::make},
         {"enable-experimental-requires-ancestor", BooleanPropertyAssertion::make},
         {"enable-suggest-unsafe", BooleanPropertyAssertion::make},
-        {"exhaustive-apply-code-action", BooleanPropertyAssertion::make},
+        {"selective-apply-code-action", SelectiveApplyCodeActionAssertions::make},
```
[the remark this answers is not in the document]
ilyailya: @froydnj here's a PR https://github.com/sorbet/sorbet/pull/5454

test/helpers/position_assertions.cc
```diff
@@ -2004,7 +2004,7 @@ SelectiveApplyCodeActionAssertions::SelectiveApplyCodeActionAssertions(std::stri
                                                                        std::unique_ptr<Range> &range, int assertionLine,
                                                                        std::vector<std::string> values,
                                                                        std::string_view assertionType)
-    : RangeAssertion(filename, range, assertionLine), assertionType(assertionType), values(values){};
+    : RangeAssertion(filename, range, assertionLine), assertionType(string(assertionType)), values(values){};
```
jez, line 2007: Why is this required?

test/testdata/lsp/code_actions/extract_func_naming.A.rbedited
```diff
@@ -0,0 +1,19 @@
...
+    'Hello'
+  end
+end
+
+module GreetingModule; end
+module Foo
+  extend T::Sig
+
+  
+
```
jez, line 18: ooc, why are there so many extra lines here?

test/testdata/lsp/code_actions/extract_class_func.rb
```diff
@@ -0,0 +1,36 @@
...
+# selective-apply-code-action: refactor.extract
+
+module Foo
+  extend T::Sig
+
+  sig {returns(String)}
+  def self.greeting
+         # ^^^^^^^^ apply-code-action: [A] Extract method to module
+    'Hello'
+  end
```
jez, line 11: Do you want to add a test where the indentation level is mismatched?

E.g., if you do

```ruby
module Foo
  module Bar
    module Qux
      def self.greeting; end
    end
  end
end
```

and then try to move this into a new module, it'll have super extra indentation. I think that's fine, the user can always reindent at their leisure, but I'd love a test.

test/testdata/lsp/code_actions/extract_class_func.rb
```diff
@@ -0,0 +1,36 @@
+# typed: strict
```
jez, line 1: Another test idea:

```ruby
sig {void}
# comment between sig and def
def foo; end

sig do
  void
end # comment after sig
def foo; end
```

test/testdata/lsp/code_actions/extract_class_func.rb
```diff
@@ -0,0 +1,36 @@
...
+  class B
+    extend T::Sig
+
+    sig {void}
+    def bar
+      m = Foo
+      Foo.greeting
+
+      m.greeting
+      print((Foo if true).greeting)
```
jez, line 32: why is there no need for an error assertion here? Doesn't `greeting` not always exist on the receiver, because it might be `nil` sometimes?

test/testdata/lsp/code_actions/extract_class_func.rb
```diff
@@ -0,0 +1,36 @@
...
+end
+
+module A
+  class B
+    extend T::Sig
+
+    sig {void}
+    def bar
+      m = Foo
+      Foo.greeting
```
jez, line 29: How does this handle when `m` comes from the sig:

```
sig {params(m: T.class_of(Foo)).void}
def example3(m)
  m.greeting
end
```

? If you don't mind adding a test for this as well, that'd be great.

test/testdata/lsp/code_actions/extract_class_func.rb
```diff
@@ -0,0 +1,36 @@
+# typed: strict
+# selective-apply-code-action: refactor.extract
+
+module Foo
+  extend T::Sig
+
+  sig {returns(String)}
+  def self.greeting
+         # ^^^^^^^^ apply-code-action: [A] Extract method to module
```
jez, line 9: I believe that we only get `DefinitionRespnse` results back for the `declLoc`, which applies to anywhere on the `def` line, but not the method body or the `end` keyword.

Do you mind adding some more tests, for e.g., the `def` keyword, the space after `def`, the `self` keyword, and the `.` token? I think that your code handles all of these case already, but they'd be good to capture in a test.

Also, I think that for these tests, the test harness only uses the first `^` in the `apply-code-action`, and all the remaining `^^^` after the first are ignored (fwiw).

test/testdata/lsp/code_actions/extract_class_func.rb
```diff
@@ -0,0 +1,36 @@
...
+  
+  sig do
+    params(x: String)
+    .returns(String)
+  end
+  def name(x)
+    "#{Foo.greeting} #{x}"
+  end
+end
+
```
jez, line 21: What do you think about getting some non-standard method defs in place?

For example

```
class A
  prop :foo, Integer
end
```

This will implicitly define two methods (`foo` and `foo=`) and they'll both have a declLoc of the whole line.

main/lsp/requests/code_action.cc
```diff
@@ -105,12 +326,33 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerInterface &
...
 
+    if (config.opts.lspExtractMethodEnabled) {
+        auto queryResult = queryByLoc(typechecker, params->textDocument->uri, *params->range->start,
+                                      LSPMethod::TextDocumentCodeAction, false);
+
+        // Generate "Extract method" code actions only for method definitions
+        if (queryResult.error == nullptr) {
+            for (auto &resp : queryResult.responses) {
+                if (auto def = resp->isDefinition()) {
+                    if (def->symbol.isMethod()) {
```
jez, line 343: We currently only return `DefinitionResponse` for method symbols right now (in DefLocSaver)`.

If we change `DefinitionResponse` to store a `MethodRef` instead of a `SymbolRef`, we can get rid of this `if` check, and ensure that if someone wants to make `DefinitionResponse` do something with non-methods in the future, our code here in code_action.cc fails with type errors instead of runtime logic errors.

main/lsp/requests/code_action.cc
```diff
@@ -105,12 +326,33 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerInterface &
...
+                        auto action = make_unique<CodeAction>("Extract method to module");
+                        action->kind = CodeActionKind::RefactorExtract;
+                        auto workspaceEdit = make_unique<WorkspaceEdit>();
+                        workspaceEdit->documentChanges = getExtractMethodEdits(config, gs, def, typechecker);
+                        action->edit = move(workspaceEdit);
+                        result.emplace_back(move(action));
```
jez, line 349: I haven't looked very closely at the LSP spec. I see that we're always constructing a workspaceEdit and emplacing it back, no matter what. But I think that there are certain cases where `getExtractMethodEdits` can fail to produce an edit.

Does VS Code get tricked if the list of actions is non-empty, but the action itself does not actually have an effect?

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
 } // namespace
 
+CodeActionTask::CodeActionTask(const LSPConfiguration &config, MessageId id, unique_ptr<CodeActionParams> params)
+    : LSPRequestTask(config, move(id), LSPMethod::TextDocumentCodeAction), params(move(params)) {}
+
+vector<unique_ptr<TextDocumentEdit>>
+CodeActionTask::getExtractMethodEdits(const LSPConfiguration &config, const core::GlobalState &gs,
+                                      const core::lsp::DefinitionResponse *definition,
+                                      LSPTypecheckerInterface &typechecker) {
+    ENFORCE(definition->symbol.isMethod());
```
jez, line 227: We should be able to delete this ENFORCE if we change `DefinitionResponse` to store a `MethodRef` as mentioned below.

test/testdata/lsp/code_actions/extract_class_func.rb
jez, line 9: It looks like there are no tests for what happens if the user attempts to use this code action on an instance method. Can we get a test for that?

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+        res += shouldCapitalize ? toupper(name[i]) : name[i];
+        shouldCapitalize = false;
+    }
+    return res;
+}
+
+optional<string> getNewModuleName(const core::GlobalState &gs, const core::NameRef name) {
+    const auto moduleName = absl::StrCat(snakeToCamelCase(name.show(gs)), "Module");
+    const auto nameNotExists = [&](auto moduleName) {
+        return gs.lookupNameUTF8(moduleName) == core::NameRef::noName();
```
jez, line 107:
```suggestion
        return gs.lookupNameUTF8(moduleName).exists();
```

jez: I personally wouldn't have written a lambda for this, but i don't feel strongly.

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+    };
+
+    // if there are no names like that, return the new module name
```
jez, line 110: ooc, was it intentional that you're doing this on names, not on symbols?

Doing this on names is ~fine, but it does mean there are some weird things, like

```ruby
puts(:GreetingModule)
```

Sorbet entered a name for `GreetingModule`, so it's going to cause the fallback logic to run, even if there's no module like `module GreetingModule`.

If it's intentional, can you write a test like what I've shown here and update this comment to explain the reasoning?

If you decide to update this to use symbols instead of names, I don't mind if you leave the comment and tests as they are.

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+    auto &file = fref.data(gs);
+
+    auto trees = typechecker.getResolved({fref});
+    auto &rootTree = trees[0].tree;
+    auto beginLoc = core::Loc::offset2Pos(fref.data(gs), rootTree.loc().beginPos());
+
+    auto pos2Loc = [&](const auto &pos) {
+        auto loc = core::Loc::offset2Pos(fref.data(gs), pos);
+        return make_unique<Position>(loc.line - 1, loc.column - 1);
+    };
```
jez, line 195: Once you switch to using `Range::fromLoc` (see below) this function becomes dead.

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+
+    auto fref = definition->termLoc.file();
+    auto &file = fref.data(gs);
+
+    auto trees = typechecker.getResolved({fref});
+    auto &rootTree = trees[0].tree;
+    auto beginLoc = core::Loc::offset2Pos(fref.data(gs), rootTree.loc().beginPos());
+
+    auto pos2Loc = [&](const auto &pos) {
+        auto loc = core::Loc::offset2Pos(fref.data(gs), pos);
```
jez, line 193: Note that in general, I'm almost never a fan of capturing `gs` in a lambda function like this. When we're writing reusable functions that work over `gs`, I very much prefer that we make them top-level helper functions.

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+                                        LSPTypecheckerInterface &typechecker, string_view newModuleName) {
+    auto moduleStart = fmt::format("module {}\n  extend T::Sig\n  ", newModuleName);
+    auto moduleEnd = "\nend";
+
+    auto fref = definition->termLoc.file();
+    auto &file = fref.data(gs);
+
+    auto trees = typechecker.getResolved({fref});
+    auto &rootTree = trees[0].tree;
+    auto beginLoc = core::Loc::offset2Pos(fref.data(gs), rootTree.loc().beginPos());
```
jez, line 190:
```suggestion
    // The loc used for the `<root>` class that surrounds all code after desugar uses the loc of the
    // top-mode node from the parser::Node result, which will have started at the first piece of
    // real Ruby syntax in the file. Thus using rootTree's beginPos here is a clever way of finding
    // the effective top of the file, skipping any comments or whitespace that exists at the top of
    // the file.
    auto beginLoc = core::Loc::offset2Pos(fref.data(gs), rootTree.loc().beginPos());
```

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+    auto topOfTheFile = file.getLine(beginLoc.line);
+    auto sigAndMethodLocs = methodLocs(gs, rootTree, definition->symbol, fref);
+    if (!sigAndMethodLocs.has_value()) {
+        return {};
+    }
+    auto [sigLoc, methodLoc] = sigAndMethodLocs.value();
+    auto methodSource = copyMethodSource(gs, sigLoc, methodLoc, fref);
+
+    auto range = make_unique<Range>(make_unique<Position>(beginLoc.line - 1, 0),
+                                    make_unique<Position>(beginLoc.line - 1, topOfTheFile.length()));
```
jez, line 210: Why are we replacing anything?

Shouldn't the range used for the replacement always just be an empty range, to effectively insert at the beginning of the file?

I would have expected that we could do

```cpp
auto range = Range::fromLoc(rootTree.loc().copyWithZeroWidth());
```

and delete most of the other code here. In particular I'm of the opinion that we should never construct raw `Range` and `Position` nodes manually, preferring instead to operate exclusively on `Loc`'s and convert to them only just before we need to respond to the LSP query.

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+                                        const core::lsp::DefinitionResponse *definition,
+                                        LSPTypecheckerInterface &typechecker, string_view newModuleName) {
+    auto moduleStart = fmt::format("module {}\n  extend T::Sig\n  ", newModuleName);
+    auto moduleEnd = "\nend";
+
+    auto fref = definition->termLoc.file();
+    auto &file = fref.data(gs);
+
+    auto trees = typechecker.getResolved({fref});
+    auto &rootTree = trees[0].tree;
```
jez, line 189:
```suggestion
    ENFORCE(!trees.empty());
    auto &rootTree = trees[0].tree;
```

main/lsp/requests/code_action.cc
jez: If we take this approach, it also means that our replacement doesn't have to include `topOfTheFile` in the format string.

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+    auto cutSource = [&](core::LocOffsets loc) {
+        return fref.data(gs).source().substr(loc.beginPos(), loc.endPos() - loc.beginPos());
+    };
+
+    auto sigSource = cutSource(sigLoc);
+    auto methodSource = cutSource(methodLoc);
+    return make_unique<string>(absl::StrCat(sigSource, "\n  ", methodSource));
+}
+
+std::optional<std::pair<core::LocOffsets, core::LocOffsets>> methodLocs(const core::GlobalState &gs,
```
jez, line 67:
```suggestion
optional<std::pair<core::LocOffsets, core::LocOffsets>> methodLocs(const core::GlobalState &gs,
```

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+                                                                        const ast::ExpressionPtr &rootTree,
+                                                                        const core::SymbolRef &method,
+                                                                        const core::FileRef &fref) {
```
jez, line 70:
```suggestion
                                                                        const core::SymbolRef method,
                                                                        const core::FileRef fref) {
```

These data structures are designed to be passed by copying.

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
     }
     return documentEdits;
 }
+
+std::optional<const ast::MethodDef *> findMethodTree(const ast::ExpressionPtr &tree, const core::SymbolRef &method) {
```
jez, line 33:
```suggestion
std::optional<const ast::MethodDef *> findMethodTree(const ast::ExpressionPtr &tree, const core::SymbolRef method) {
```

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +29,229 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+    } else if (auto methodDef = ast::cast_tree<ast::MethodDef>(tree)) {
+        if (methodDef->symbol == method.asMethodRef()) {
+            return methodDef;
+        }
+    }
+    return std::nullopt;
+}
+
+unique_ptr<string> copyMethodSource(const core::GlobalState &gs, const core::LocOffsets &sigLoc,
+                                    const core::LocOffsets &methodLoc, const core::FileRef &fref) {
```
jez, line 57:
```suggestion
unique_ptr<string> copyMethodSource(const core::GlobalState &gs, const core::LocOffsets sigLoc,
                                    const core::LocOffsets methodLoc, const core::FileRef fref) {
```

test/testdata/lsp/code_actions/extract_class_func.rb
ilyailya: By the way how it's implemented at the moment the new module would not have an extra indentation, but I like the idea of testing single line methods and methods without sigs.

ilyailya: While I was adding a test for a single line method, I actually found a bug, thanks for pointing to this!

test/testdata/lsp/code_actions/extract_class_func.rb
jez: Another test: what happens if the user wrote `private_class_method def foo` instead of just `def foo`?

test/testdata/lsp/code_actions/extract_class_func.rb
ilyailya: My goal with that case here was to test that renaming works correctly if the receiver is an 
expression. There is no error here, because we're being smart about the condition being true every time I guess. However, if you replace `true` with an expression, you'll actually get an error

Screen Shot 2022-03-14 at 21 36 34

test/testdata/lsp/code_actions/extract_class_func.rb
ilyailya: Moving props doesn't work with this implementation. I didn't dig too much, but my guess would be it doesn't work because it can't find the method body. That's a great catch! However, I'd put it as a stretch goal, because it's mostly related to the instance methods and ruby-infra is mostly interested in class methods. I'll add test to fix current behavior

jez: stretch goals are totally fine, all I was asking for in most of these comments was that we add tests to capture the current behavior, even if it doesn't yet do what we'd like it do to.

main/lsp/requests/code_action.cc
jez: https://github.com/sorbet/sorbet/pull/5473

ilyailya:
@jez thanks for such a detailed review, appreciate that! I think I've addressed all of the comments, would you mind taking a look one more time?

test/lsp_test_runner.cc
```diff
@@ -139,11 +139,15 @@ void validateCodeActions(LSPWrapper &lspWrapper, Expectations &test, string file
...
-        bool codeActionTitleUnique =
-            receivedCodeActionsByTitle.find(codeAction->title) == receivedCodeActionsByTitle.end();
-        CHECK_MESSAGE(codeActionTitleUnique, "Found code action with duplicate title: " << codeAction->title);
-
-        if (codeActionTitleUnique) {
+        // "Move method" refactorings always have the same title
+        if (codeAction->kind != CodeActionKind::RefactorExtract) {
+            bool codeActionTitleUnique =
+                receivedCodeActionsByTitle.find(codeAction->title) == receivedCodeActionsByTitle.end();
+            CHECK_MESSAGE(codeActionTitleUnique, "Found code action with duplicate title: " << codeAction->title);
```
jez, line 146: Isn't this guard required for the correctness of the test runner itself?

If `codeAction->kind` is `RefactorExtract` and there is already one of them, then we're going to overwrite the previous one on line 151. Is that not a problem? Not really sure how `receivedCodeActionsByTitle` is being used.

test/lsp_test_runner.cc
```diff
@@ -247,7 +251,14 @@ void testQuickFixCodeActions(LSPWrapper &lspWrapper, Expectations &test, const v
                                 applyCodeActionAssertions, error->toString(), false);
         }
 
-        for (auto codeActionAssertion : applyCodeActionAssertions) {
+        // This weird loop is here because `validateCodeActions` erases the elements from `applyCodeActionAssertions`
+        // and we are iterating over that container
+        auto applyCodeActionAssertionsSize = applyCodeActionAssertions.size();
+        for (int i = 0; i < applyCodeActionAssertionsSize; i++) {
```
jez, line 257: We don't actually use `i` anywhere. Let's drop the comment, and rewrite this to use

```cpp
while (!applyCodeActionAssertions.empty()) {
  auto initialSize = applyCodeActionAssertions.size();
  // ...
  ENFORCE(applyCodeActionAssertions.size() < initialSize);
}
```

test/testdata/lsp/code_actions/move_method/class_func.rb
```diff
@@ -0,0 +1,41 @@
...
+      Foo.greeting
+
+      m.greeting
+      print((Foo if true).greeting)
+    end
+
+    sig {params(m: T.class_of(Foo)).void}
+    def example3(m)
+      m.greeting
+    end
```
jez, line 38: Huh. I'm surprised here. How are you detecting this?

What if `m: T.class_of(FooParent)` where

```ruby
class FooParent; end
class Foo < FooParent
  def self.greeting; end
end
```

?

test/testdata/lsp/code_actions/move_method/class_func.A.rbedited
```diff
@@ -0,0 +1,47 @@
...
+  class B
+    extend T::Sig
+
+    sig {void}
+    def bar
+      m = Foo
+      GreetingModule.greeting
+
+      GreetingModule.greeting
+      print((GreetingModule).greeting)
```
jez, line 38: Is it intentional that we dropped the `if true` that was present in the original?

test/testdata/lsp/code_actions/move_method/comments.rb
```diff
@@ -0,0 +1,29 @@
+# typed: strict
+# selective-apply-code-action: refactor.extract
+#
+# Test asserts comments between a sig and a method def are not moved,
+# because we cut out only a signature and a method body without touching anything in between
```
jez, line 5: Are we sure that this is the desired behavior? It seems to me that it isn't desired for a number of reasons:

- People will lose potentially important comments, and have to move them on their own
- Modifiers like `private_class_method` that exist between the `sig` and the `def` will not be moved.

It seems that if we change the code to call `sigLoc.join(methodDefLoc)` to compute the source of the method to be moved, we solve these issues and **also** solve the issue that we're leaving so many empty lines behind at the old position in the file (it won't get rid of all of the empty lines, but there will at least be fewer, which is a start).

LMK if you made this choice as a conscious design decision, because I'm curious to hear the justification.

test/testdata/lsp/code_actions/move_method/instance_method.A.rbedited
```diff
@@ -0,0 +1,23 @@
...
+
+class Inc
+  extend T::Sig
+
+  
+  
+end
+
+foo = Inc.new
+res = IncModule.inc(42)
```
jez, line 21: Because of cases like this, I think we should detect if the method the user has requested to refactor is an instance method, and not generate a refactor.extract code action in that case. My reasoning:

- The step here to rename the receiver is wrong. The code generated by the action fails to type check.
- If we attempt to make this code type check by changing `def inc` to be a singleton class method (`def self.inc`), that doesn't necessarily make sense.

I think it's fine to say that there's no code action here. What do you think?

test/testdata/lsp/code_actions/move_method/no_sig.rb
```diff
@@ -0,0 +1,15 @@
+# typed: strict
+# selective-apply-code-action: refactor.extract
+#
+# The test is asserting we can't move a method without a sig
```
jez, line 4: This limitation seems artificial. If we take the approach I mentioned in another comment of effectively doing `sigLoc.join(methodDefLoc)`, I think that "method without a sig" should work naturally, at least once we write some code like this:

```cpp
auto extractLoc = sigLoc.exists() ? sigLoc.join(methodDefLoc) : methodDefLoc;
```

lmk if I'm mistaken in thinking that it's as easy to support this as I think, or if this was a conscious design decision.

test/testdata/lsp/code_actions/move_method/single_line_def.rb
```diff
@@ -0,0 +1,12 @@
+# typed: strict
+# selective-apply-code-action: refactor.extract
+#
+# The test is asserting we can't move a method without a sig
```
jez, line 4: I think you copy/pasted this comment wrong from another test.

jez (commented):
I left some comments on the tests. Feel free to take a look. I still have the implementation to look at, but I'll have to get to that later.

test/lsp_test_runner.cc
ilyailya: > Isn't this guard required for the correctness of the test runner itself?

Yep, that's true. The guard rail was brought up for the quickfix assertions. For quickfixes you don't want to have two fixes at the same place with the same title. 

But that assumption is not correct for refactor actions. For example, you can have two refactor code actions with the same title `Move method` in one location if that location is `prop :foo, Integer`. The prop test was the main driver for introducing this change

You also right about code actions overwriting each other, but that's fine for now, because the only case where the action is overwritten is the same `test/testdata/lsp/code_actions/move_method/prop.rb` test. And the refactor actions in the test are no-ops, so that's okay.

My goal here was to get the prop test working without large refactors to the test runner.

test/testdata/lsp/code_actions/move_method/class_func.rb
ilyailya: >How are you detecting this?

I'm reusing the rename infra here. `addSubclassRelatedMethods` or `addDispatchRelatedMethods` add the `greeting` symbol in the rename queue, and the `m.greeting` location is recorded in the `locs()` of that symbol.

https://github.com/sorbet/sorbet/blob/f895091951e3ba59a162834299e282383acae231/main/lsp/AbstractRenamer.cc#L92-L133

test/testdata/lsp/code_actions/move_method/class_func.A.rbedited
ilyailya: Yes, if the rename location is the send location we don't analyze the receiver, but replace the whole receiver with the new module

https://github.com/sorbet/sorbet/blob/f895091951e3ba59a162834299e282383acae231/main/lsp/requests/code_action.cc#L174-L176

test/testdata/lsp/code_actions/move_method/comments.rb
ilyailya: That was a design choice, but more I work on that PR, more I think we should leave it. The main consideration was, I wanted to be as careful as possible, and make the least amount of mess with the refactoring. 

Moving everything between the sig and the method would probably make more sense, as this behavior is probably more obvious for the user.

test/testdata/lsp/code_actions/move_method/class_func.A.rbedited
jez: Ahh, ok I see what's going on. Sorbet is smart enough to see that the condition is always `true` so that the receiver is always the same type.

In that case I have another test case for you:

```ruby
# typed: true
extend T::Sig

class Left
  def self.foo; end
end

class Right
  def self.foo; end
end

sig {params(x: T.any(Left, Right)).void}
def example(x)
  x.foo
end
```

Let's say that the user requested to rename **only** the `def self.foo; end` in `Left`. I think that for cases like this, we should decline to rename the `x.foo` call site. Instead we should just let Sorbet retypecheck the codebase, which will cause it to report a type error, because it'll have to be something that the user manually fixes, not something that can be systematically renamed.

To do this, you should be able to look in the `SendResponse::dispatchResult`. If you look at the structure of the `DispatchResult` object, you'll see it has a `main` DispatchComponent` and a `secondary` DispatchResult, which basically makes `dispatchResult` be the start of a linked list of DispatchResults. This linked list will have one entry for each method that the call site dispatched to, which in the `x.foo` case will have one entry for `Left.foo` and one entry for `Right.foo`.

We can use that information in one of a number of somewhat equivalent ways to detect this case:

- Loop over the chain of DispatchResults, and assert for all of them that `->main.method` is the method that the move request was called on, or
- Don't bother with looping, just assert that `dispatchResult->main.method` is the method we're moving and that `dispatchResult->secondary == nullptr` (i.e., that the linked list is length 1).

lmk if that's confusing, happy to chat more.

test/lsp_test_runner.cc
jez: In that case can you leave a TODO comment in the test runner? this seems like it could be a footgun in the future.

test/testdata/lsp/code_actions/move_method/instance_method.A.rbedited
ilyailya: >The code generated by the action fails to type check.

That was a conscious design decision. The refactoring doesn't try to generate the correct code, but tries to automate the toil of moving the method out and updating the call sites. The result may be not correct, but the typecheck errors will guide the user on how to fix them.

>I think it's fine to say that there's no code action here.

I also think it's fine too. However, I can still see the value of leaving it. The refactoring doesn't generate the correct code, but the type errors, will help you to fix them. Does this refactoring make sense here is the question to the users and what their goals are. If the user is aware of all of that, the code action can still automate some toil for them.

But I can agree with you here, that behavior is a bit surprising, and if we want to follow the principle of least surprise, it'd make sense to limit the refactoring to class methods only.

jez: yeah in the past people have been more annoyed with autocorrects that are wrong than with those that don't exist. Especially in cases where it's possible to quickly check to know that some autocorrect will be wrong, I like to err on the side of not doing it.

test/testdata/lsp/code_actions/move_method/class_func.rb
ilyailya: btw, `m: T.class_of(FooParent)` would have a type error, because `FooParent` doesn't have a `greeting` defined. 

```ruby
sig {params(m: T.class_of(Foo)).void}
def example3(m)
  m.greeting
    # ^ error: Method `greeting` does not exist on `T.class_of(FooParent)`
end
```

test/testdata/lsp/code_actions/move_method/class_func.A.rbedited
ilyailya: Wow, what a catch! I could even think of that use case, thanks for flagging! I'll limit call site updates only to trivial use cases and add a test

test/lsp_test_runner.cc
ilyailya: Fortunately, after disabling the refactoring for the instance methods we don't need this hack anymore

ilyailya:
@jez would you mind taking a look one more time? 
I've disabled the feature for the instance methods and disabled the call site renaming for non-trivial cases. Also a signature and a method body now cut together, so no weird trailing whitespaces anymore!

main/sig_finder/sig_finder.cc
```diff
@@ -18,23 +18,19 @@ std::optional<SigLoc> findSignature(const core::GlobalState &gs, const core::Sym
...
+    auto method_def_line = source.rfind('\n', line_end);
+    auto prev_line_end = source.rfind('\n', method_def_line - 1);
     while (prev_line_end != source.npos) {
         auto line_start = prev_line_end + 1;
         auto line = source.substr(line_start, line_end - line_start);
         line = absl::StripAsciiWhitespace(line);
-        // We do not skip empty lines, contra LSP's documentation finder.
-        // But as we are starting from just before the method definition,
+        // We do skip empty lines between signature and deefinition
+        // Also we are starting from just before the method definition,
```
jez, line 28:
```suggestion
        // We do skip empty lines between signature and definition
        // Also we are starting from just before the method definition,
```

test/testdata/lsp/code_actions/move_method/instance_method.rb
```diff
@@ -0,0 +1,18 @@
+# typed: strict
+# selective-apply-code-action: refactor.extract
+#
+# No code actions should be available for instance methods
```
jez, line 4: I'm not super familiar with the test runner for code actions. Would this test start to fail if some sort of support for instance methods were added eventually?

test/testdata/lsp/code_actions/move_method/no_sig.rb
jez, line 4:
```suggestion
# The test is asserting we can move a method without a sig
```

test/testdata/lsp/code_actions/move_method/prop.rb
```diff
@@ -0,0 +1,8 @@
+# typed: strict
+# selective-apply-code-action: refactor.extract
+#
+# No code actions should be available for the props
+
+class Foo < T::Struct
+  prop :bar, Integer
+end
```
jez, line 8: ditto https://github.com/sorbet/sorbet/pull/5309/files#r831607188

main/sig_finder/sig_finder.cc
```diff
@@ -100,7 +97,14 @@ std::optional<SigLoc> findSignature(const core::GlobalState &gs, const core::Sym
             break;
         }
 
-        // If we find anything else, even a blank line, assume that the heuristic
+        // We found a comment between a sig and a method def, look one line higher
+        if (absl::StartsWith(line, commentLineStart)) {
```
jez, line 101: I guess this heuristic only works when the `#` starts at the line start, with no preceding whitespace.

If we don't want to fix this now, let's either add a comment or a test that documents this this as a known limitation that might be worth fixing later. (This PR is starting to look really good, and this is the sort of small thing that I'm happy to punt on.)

Though, writing up this comment also got me thinking that it might be nice to handle moving documentation comments above the method/sig too:

```ruby
class A
  # My comment on the method
  # @param arg0 Docs for the first argument
  sig {params(arg0: Integer).void}
  def self.foo(arg0); end
end
```

Right now we'd leave the comment behind. This is also something I'm fine punting on for a future change. If you'd like to merge this change before fixing this, I'd love if you could be sure to file an issue to remind us.

main/lsp/requests/code_action.h
```diff
@@ -14,6 +15,12 @@ class CodeActionTask final : public LSPRequestTask {
     std::unique_ptr<ResponseMessage> runRequest(LSPTypecheckerInterface &typechecker) override;
 
     bool canUseStaleData() const override;
+
+private:
+    std::vector<std::unique_ptr<TextDocumentEdit>> getMoveMethodEdits(const LSPConfiguration &config,
+                                                                      const core::GlobalState &gs,
+                                                                      const core::lsp::MethodDefResponse *definition,
```
jez, line 22: I believe that we know that the pointer is non-null, and I'm a stickler against raw pointers when possible. If this signature works, I'd love to use it:

```suggestion
                                                                      const core::lsp::MethodDefResponse &definition,
```

I hope it's as easy as changing the call sites to `..., *definition, ...`

test/testdata/lsp/code_actions/move_method/class_func.rb
jez: Sorry, I think I meant to give this example:

```ruby
# typed: true
# selective-apply-code-action: refactor.extract

class Parent
  def self.only_in_parent
    # ^ apply-code-action: [A] Move method to a new module
  end

  def self.in_parent_and_child
    # ^ apply-code-action: [B] Move method to a new module
  end
end

class Child < Parent
  def self.in_parent_and_child; end
end

class Main
  extend T::Sig

  sig {params(x: T.class_of(Child)).void}
  def example(x)
    x.only_in_parent

    x.in_parent_and_child
  end
end
```

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%23%20selective-apply-code-action%3A%20refactor.extract%0A%0Aclass%20Parent%0A%20%20def%20self.only_in_parent%0A%20%20%20%20%23%20%5E%20apply-code-action%3A%20%5BA%5D%20Move%20method%20to%20a%20new%20module%0A%20%20end%0A%0A%20%20def%20self.in_parent_and_child%0A%20%20%20%20%23%20%5E%20apply-code-action%3A%20%5BB%5D%20Move%20method%20to%20a%20new%20module%0A%20%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20def%20self.in_parent_and_child%3B%20end%0Aend%0A%0Aclass%20Main%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams%28x%3A%20T.class_of%28Child%29%29.void%7D%0A%20%20def%20example%28x%29%0A%20%20%20%20x.only_in_parent%0A%0A%20%20%20%20x.in_parent_and_child%0A%20%20end%0Aend)

The one I'm particularly surprised by is the `in_parent_and_child` case. That one is weird because I would have thought that the override in `Child` would have taken precedence over the definition in `Parent`, so given that `x` has type `T.class_of(Child)`, I'm surprised that the renamer queues up that one as a location that needs to be renamed.

The other case is also worth considering, but I think that the current behavior is probably fine.

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +34,234 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+        }
+    } else if (auto methodDef = ast::cast_tree<ast::MethodDef>(tree)) {
+        if (methodDef->symbol == method.asMethodRef()) {
+            return methodDef;
+        }
+    }
+    return nullopt;
+}
+
+unique_ptr<string> copyMethodSource(const core::GlobalState &gs, optional<core::LocOffsets> maybeSigLoc,
```
jez, line 61: Why make this a `unique_ptr`?

`std::string` will manage its own storage. If youu need to get around the fact that `substr` returns a `string_view`, you should be able to wrap the whole thing in a call to `return string(fref.data(gs)...)`

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +34,234 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+                    return;
+                }
+            }
+
+            edits[sendResp->receiverLoc] = newName;
+        }
+    }
+    void addSymbol(const core::SymbolRef symbol) override {
+        if (symbol.isMethod()) {
+            addSubclassRelatedMethods(gs, symbol.asMethodRef(), getQueue());
```
jez, line 189: It's possibly that we don't actually want `addSubclassRelatedMethods` here. See my [comment here below](https://github.com/sorbet/sorbet/pull/5309/files#r831624207).

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +34,234 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+        } else {
```
jez, line 190: Should we do nothing if the user requests to rename a symbol that is not a method symbol?

Or, if we're sure that we would have early exited already and will never get to this point, can we ENFORCE that `symbol.isMethod()`?

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +34,234 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+            getQueue()->tryEnqueue(symbol);
+        }
+    }
+}; // CallSiteRenamer
```
jez, line 194:
```suggestion
}; // MethodCallSiteRenamer
```

main/lsp/requests/code_action.cc
```diff
@@ -29,8 +34,234 @@ vector<unique_ptr<TextDocumentEdit>> getEdits(const LSPConfiguration &config, co
...
+                invalid = true;
+                error = fmt::format("The `{}` method cannot be moved to a module.", oldName);
+                return;
+            }
+        }
+        // block any method not starting with /[a-zA-Z0-9_]+/. This blocks operator overloads.
+        if (!isalnum(oldName[0]) && oldName[0] != '_') {
+            error = fmt::format("The `{}` method cannot be moved to a module.", oldName);
+            invalid = true;
+        }
```
jez, line 153: Do you think it's worth writing tests for these?

I believe you cargo-culted these from `rename.cc` if I'm not mistaken.

jez (approved):
This looks great. I have a few questions, but this is looking great on the whole.

jez:
@ilyailya heads up that I pushed two commits to your branch that were slightly too annoying to do as suggested edits.

If you don't like either of them, feel free to back the changes out.

main/sig_finder/sig_finder.cc
ilyailya: >I guess this heuristic only works when the # starts at the line start, with no preceding whitespace.

It should work with leading whitespaces too. In the same method the `line` is trimmed so it should be okay. I'll add test for that.
https://github.com/sorbet/sorbet/blob/14bf3533903bd2ea376a323de2657d7b39bf6d20/main/sig_finder/sig_finder.cc#L26

main/lsp/requests/code_action.cc
ilyailya: I was not sure about string managing it's own lifetime, so I wrapped it into `unique_ptr`

main/lsp/requests/code_action.cc
ilyailya: This part was cargo culted from the renamer. My understanding was we should add as much symbols as possible, because we use them for the `queryBySymbol` later. I assumed query results may lead to more call site locations
https://github.com/sorbet/sorbet/blob/14bf3533903bd2ea376a323de2657d7b39bf6d20/main/lsp/LSPTask.cc#L324-L342

But I tried adding only method symbols here and all tests passed, so it should be safe to early exit if the symbol is not a method.

main/lsp/requests/code_action.cc
ilyailya: Actually seems like limitation for the `initialize` and `call` method doesn't make a lot of sense with the current implementation, because the refactoring action is limited to class methods only, and those methods are instance methods. However, I would like to leave this code here, because we might add support for instance methods later. I'll add test we don't generate a code action for those methods.

And testing for the operator overloads is a great idea, I'll do that too.

main/sig_finder/sig_finder.cc
ilyailya: https://github.com/sorbet/sorbet/issues/5501

PR #5313, opened 2022-02-16
expand the range of invalid `Loc` positions

froydnj:
When generating package RBIs, some packages have, shall we say, large API surface areas that cause the RBI size to be larger than the currently permitted `Loc` position range.  This PR expands the range by a factor of 16, because 256MB should be enough for anybody.

### Motivation

Handling large files with debug assertions turned on.

### Test plan

See included automated tests.

### DIFF
--- a/core/Loc.h
+++ b/core/Loc.h
@@ -11,7 +11,7 @@ class GlobalState;
 class Context;
 class MutableContext;
 
-constexpr int INVALID_POS_LOC = 0xffffff;
+constexpr int INVALID_POS_LOC = 0xfffffff;
 struct LocOffsets {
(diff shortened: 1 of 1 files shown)


elliottt (approved):

jez:
Is this just an oversight? Should we set it to UINT32_MAX instead?

froydnj:
> Is this just an oversight? Should we set it to UINT32_MAX instead?

I was assuming that it was deliberately not set to `UINT32_MAX` to catch cases where you walked backwards off the beginning of the file, or something like that.  I also thought that it was set lower to discourage the use of large files (?), or at the very least suggest that the code hadn't been audited for cases where you got near the 2GB or 4GB boundary.

PR #5318, opened 2022-02-16
Change how long lines are handled

jez:
### Motivation

It's actually quite easy for lines to be 100+ chars long.

But even for long lines, since the loop will already short circuit once the
shorter line runs out of whitespace, it's fine.

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/cc/lexer.rl
+++ b/parser/parser/cc/lexer.rl
@@ -185,10 +185,6 @@ int lexer::compare_indent_level(token_t left, token_t right) {
 
     // attempt to defeat pathological cases (extremely long lines)
     if ((leftStart - leftLineStart > 100) || (rightStart - rightLineStart > 100)) {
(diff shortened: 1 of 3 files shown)


parser/parser/cc/lexer.rl
```diff
@@ -185,10 +185,6 @@ int lexer::compare_indent_level(token_t left, token_t right) {
 
     // attempt to defeat pathological cases (extremely long lines)
     if ((leftStart - leftLineStart > 100) || (rightStart - rightLineStart > 100)) {
-        // In this case, lie say that the indentation level is the same.
-        // This will basically mean falling back to the indendation-agnostic behavior.
-        // We could alternatively attempt to return some sort of error state here.
-        return 0;
     }
```
froydnj, line 188: This whole conditional should now be removed?

froydnj (approved):

PR #5319, opened 2022-02-17
Add type_member bounds to rbi generation

elliottt:
Add bounds information when emitting `type_member` and `type_template` declarations during rbi generation.

### Motivation

Stabilizing package rbi generation.

### Test plan

See included automated tests.

### DIFF
--- a/packager/rbi_gen.cc
+++ b/packager/rbi_gen.cc
@@ -34,7 +34,7 @@ private:
 public:
     template <typename... Args> void println(fmt::format_string<Args...> fmt, Args &&...args) {
         fmt::format_to(std::back_inserter(out), "{:{}}", "", this->indent * 2);
-        fmt::format_to(std::back_inserter(out), fmt, args...);
+        fmt::format_to(std::back_inserter(out), fmt, std::forward<Args>(args)...);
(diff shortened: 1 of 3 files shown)


packager/rbi_gen.cc
```diff
@@ -34,7 +34,7 @@ class Output final {
 public:
     template <typename... Args> void println(fmt::format_string<Args...> fmt, Args &&...args) {
         fmt::format_to(std::back_inserter(out), "{:{}}", "", this->indent * 2);
-        fmt::format_to(std::back_inserter(out), fmt, args...);
```
elliottt, line 37: `std::forward` is necessary to allow the use of formatters like `fmt::join` with `println`.

packager/rbi_gen.cc
```diff
@@ -34,7 +34,7 @@ class Output final {
...
+        fmt::format_to(std::back_inserter(out), fmt, std::forward<Args>(args)...);
```
elliottt, line 37: `std::forward` is necessary to allow the use of formatters like `fmt::join` with `println`, as it allows their rvalue status to be preserved.

jez (approved):

PR #5320, opened 2022-02-17
Delete "uploading test results" to save time in CI

jez:
### Motivation

Faster tests

### Test plan

See included automated tests.

### DIFF
--- a/.buildkite/test-compiler.sh
+++ b/.buildkite/test-compiler.sh
@@ -41,13 +41,13 @@ test_args=(
   "-c"
   "opt"
   "--config=forcedebug"
+  "--spawn_strategy=local"
 )
 
 ./bazel test \
   --experimental_generate_json_trace_profile \
   --profile=_out_/profile.json \
   --test_summary=terse \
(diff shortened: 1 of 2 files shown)


elliottt:
> > Is there an easy way to replace the junit xml with a text file that contains the output from the terse test summary? That's the output that I'm more interested anyway, and I'd imagine it's much faster to produce.
> 
> take a look at the test-static-sanitized.sh. if you like it i can copy it to the compiler one

I think this looks good, do you have an example build failure that I can take a look at?

jez:
> I think this looks good, do you have an example build failure that I can take a look at?

image

https://buildkite.com/sorbet/sorbet/builds/20674#72fe6f4d-9f9e-400a-9006-dfc501d2d502

elliottt:
> > I think this looks good, do you have an example build failure that I can take a look at?
> 
> image
> 
> https://buildkite.com/sorbet/sorbet/builds/20674#72fe6f4d-9f9e-400a-9006-dfc501d2d502

This looks really good! Let's move the compiler test over as well.

PR #5322, opened 2022-02-17
skip `StubModule` mixins during package RBI generation

froydnj:
I am not super-excited about this PR, but we can legitimately type-check codebases with RBIs containing slightly out-of-date definitions as far as their mixins go, so we should be able to generate package RBIs for such codebases too.

### Motivation

Robustness during package RBI generation.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_LAZPekgS8SRC01
→ https://go/builds/bui_LAZPllJitCrWQo

jez:
Yeah I'm not super excited about this either. Is there any way we can look into this more closely?

froydnj:
> Yeah I'm not super excited about this either. Is there any way we can look into this more closely?

I agree that this is not ideal, but the behavior being proposed here seems consistent with other parts of Sorbet (e.g. class linearization, bits in LSP, CFG building, etc.).

I added a commit with a test that illustrates the problem, and updated the change to skip `StubModule` to also show how the PR fixes things.

test/testdata/packager/rbi_gen/missing_methods.rbi
```diff
@@ -0,0 +1,14 @@
+# typed: autogenerated
+
+# Simulates a situation where we've run missing methods to find things, but
+# we've updated the code to no longer define things referred to by the
+# generated RBIs.
+#
+# RBI generation should not output NonexistentThing, or StubModule.
```
jez, line 7: this makes sense, thanks!

jez (approved):

PR #5324, opened 2022-02-17
Single-package rbi generation prototype

elliottt:
Initial work on adding support for single-package rbi generation for Stripe. There are some issues remaining, but parent packages name resolution is implemented which was one of the main risks for this project.

This should be relatively un-invasive: all functionality is gated on passing the `--single-package=...` flag, and the new state on `GlobalState` will default-initialize to values that disable all of the single-package functionality.

### Motivation

Working towards incremental package rbi generation for Stripe's codebase.

### Test plan

See included automated tests.

### DIFF
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -602,7 +602,7 @@ int realmain(int argc, char *argv[]) {
                 }
                 singlePackageTarget = info.packageName;
 
-                // Only keep inputs that are part of the package we're generating rbis for
+                // Only keep inputs that are part of the package whose interface we're generating
                 auto &db = gs->packageDB();
                 auto it = std::remove_if(inputFiles.begin(), inputFiles.end(), [&gs = *gs, &db, &info](auto file) {
                     auto &pkg = db.getPackageForFile(gs, file);
@@ -610,8 +610,8 @@ int realmain(int argc, char *argv[]) {
                 });
                 inputFiles.erase(it, inputFiles.end());
 
-                // Record parent information in GlobalState to indicate to the resolver that we should stub out
-                // constants that fail to resolve.
+                // Record parent information in GlobalState to guide the resolver when stubbing out constants that come
+                // from other packages.
                 gs->singlePackageParents.emplace(std::move(info.parents));
             }
 #endif
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -455,6 +455,7 @@ int realmain(int argc, char *argv[]) {
     gs->errorUrlBase = opts.errorUrlBase;
     gs->semanticExtensions = move(extensions);
     vector<ast::ParsedFile> indexed;
+    core::NameRef singlePackageTarget;
 
     gs->requiresAncestorEnabled = opts.requiresAncestorEnabled;
 
@@ -548,6 +549,8 @@ int realmain(int argc, char *argv[]) {
             logger->warn("Package rbi generation is disabled in sorbet-orig for faster builds");
             return 1;
 #else
+            Timer rbiGenTimer(logger, "rbiGeneration.setup");
+
             if (opts.stripePackages) {
                 logger->error("Cannot serialize package RBIs in legacy stripe packages mode.");
                 return 1;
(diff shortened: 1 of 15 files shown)


test/cli/rbi-gen-single/rbi-gen-single.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo "-- sanity-checking package with --stripe-packages"
+"$sorbet" --silence-dev-message \
+  --stripe-packages \
+  --dump-package-info="$rbis/package-info.json" \
+  "$test_path"
+
+# Generate rbis for each package
+find . -name __package.rb | sort | while read -r package; do
+  name="$(awk '/class/ {print $2}' "$package")"
```
froydnj, line 20: This is totally non-blocking, but this script is doing very similar things to the existing single-package typechecking tests; it would be nice if they were merged somehow.  I don't know how to make that happen, exactly, because this script is checking for different things than the existing test.  But it wouldn't surprise me if we wind up needing more tests of this flavor.

packager/rbi_gen.cc
```diff
@@ -1321,4 +1322,58 @@ void RBIGenerator::run(core::GlobalState &gs, const UnorderedSet<core::ClassOrMo
...
+        bool equal = true;
+        auto it = fullName.begin();
+        for (auto part : nameParts) {
+            if (it == fullName.end() || it->shortName(gs) != part) {
+                equal = false;
+                break;
+            }
+
+            ++it;
+        }
```
froydnj, line 1367: I feel like there ought to be a `std::` function for this, but I'm not seeing it.

froydnj (approved):
This seems reasonable to me.  I have a couple comments on things that I'd like to see added before this gets merged.  Feel free to push back.

packager/rbi_gen.h
```diff
@@ -23,8 +23,21 @@ class RBIGenerator final {
...
 
-    static void run(core::GlobalState &gs, const UnorderedSet<core::ClassOrModuleRef> &packageNamspaces,
+    // Generate RBIs for all packages named in `packageFiles`.
+    static void run(core::GlobalState &gs, const UnorderedSet<core::ClassOrModuleRef> &packageNamespaces,
                     std::string outputDir, WorkerPool &workers);
+
+    // Generate RBIs for a single package, whose path is given by `packageName`. The `packageFiles` vector must include
+    // the file named by `packageName`, or an error will be raised.
+    static void runSinglePackage(core::GlobalState &gs, const UnorderedSet<core::ClassOrModuleRef> &packageNamespaces,
+                                 core::NameRef package, std::string outputDir, WorkerPool &workers);
```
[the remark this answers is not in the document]
elliottt: The comment is out of date, thanks for catching this!

main/realmain.cc
```diff
@@ -736,8 +765,15 @@ int realmain(int argc, char *argv[]) {
...
             return 1;
 #else
+            Timer rbiGenTimer(logger, "rbiGeneration.run");
             auto packageNamespaces = packager::RBIGenerator::buildPackageNamespace(*gs, *workers);
-            packager::RBIGenerator::run(*gs, packageNamespaces, opts.packageRBIOutput, *workers);
+
+            if (gs->singlePackageParents.has_value()) {
+                packager::RBIGenerator::runSinglePackage(*gs, packageNamespaces, singlePackageTarget,
+                                                         opts.packageRBIOutput, *workers);
+            } else {
```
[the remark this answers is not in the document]
elliottt: Happy to switch over to that.

resolver/resolver.cc
```diff
@@ -302,6 +302,124 @@ class ResolveConstantsWalk {
...
+            for (auto &path : exportPaths) {
+                stub.exports.emplace_back(path[prefixLen]);
+            }
+        }
+
+        return stubs;
+    }
+
+    static void stubForRbiGeneration(core::MutableContext ctx, const Nesting *scope, ast::ConstantLit *out) {
+        static std::vector<ParentPackageStub> parents = initParentStubs(ctx);
```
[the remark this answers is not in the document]
elliottt: Yep, it definitely will. I'll move this to an instance variable.

test/cli/rbi-gen-single/rbi-gen-single.sh
elliottt: I looked into merging it in, but wanted to get this PR merged with some testing present. It would be totally reasonable to rework this test into existing tests, but I'd rather not block this PR on that.

packager/rbi_gen.cc
elliottt: I wanted to use `std::equal`, but not being able to supply a function for the equality check breaks it. Yet another case where importing range-v3 would help :)

elliottt: It turns out that `std::equal` does take a binary predicate as of c++17, so I reworked this to use that instead.

Issue #5325, opened 2022-02-18
Autocorrect to suggest using `Array#compact`


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28xs%3A%20T%3A%3AArray%5BInteger%5D%29.returns%28T%3A%3AArray%5BInteger%5D%29%7D%0Adef%20takes_int_list%28xs%29%0A%20%20result%20%3D%20xs.concat%28%5Bnil%5D%29%0A%20%20result%0Aend%0A%0Asig%20%7Bparams%28xs%3A%20T%3A%3AArray%5BT.nilable%28Integer%29%5D%29.void%7D%0Adef%20takes_nilable_int_list%28xs%29%0A%20%20takes_int_list%28xs%29%0Aend%0A)

```ruby
# typed: true
extend T::Sig

sig {params(xs: T::Array[Integer]).returns(T::Array[Integer])}
def takes_int_list(xs)
  result = xs.concat([nil])
  result
end

sig {params(xs: T::Array[T.nilable(Integer)]).void}
def takes_nilable_int_list(xs)
  takes_int_list(xs)
end

```

#### Observed output

```
editor.rb:7: Expected T::Array[Integer] but found T::Array[T.nilable(Integer)] for method result type https://srb.help/7005
     7 |  result
          ^^^^^^
  Expected T::Array[Integer] for result type of method takes_int_list:
    editor.rb:5:
     5 |def takes_int_list(xs)
        ^^^^^^^^^^^^^^^^^^^^^^
  Got T::Array[T.nilable(Integer)] originating from:
    editor.rb:6:
     6 |  result = xs.concat([nil])
                   ^^^^^^^^^^^^^^^^

editor.rb:12: Expected T::Array[Integer] but found T::Array[T.nilable(Integer)] for argument xs https://srb.help/7002
    12 |  takes_int_list(xs)
          ^^^^^^^^^^^^^^^^^^
  Expected T::Array[Integer] for argument xs of method Object#takes_int_list:
    editor.rb:4:
     4 |sig {params(xs: T::Array[Integer]).returns(T::Array[Integer])}
                    ^^
  Got T::Array[T.nilable(Integer)] originating from:
    editor.rb:11:
    11 |def takes_nilable_int_list(xs)
                                   ^^
Errors: 2
```

#### Expected behavior

[...]

PR #5327, opened 2022-02-18
add `Loc{,Offsets}::copyEndWithZeroLength` helper

froydnj:
### Motivation

This pattern is used in a couple of places (the ones changed in this PR are the easily-greppable ones), and I was looking for a helper like this when experimenting with a fix for #5325.

Bikeshedding on the name welcome.

### Test plan

See included automated tests.

### DIFF
--- a/core/Loc.h
+++ b/core/Loc.h
@@ -37,6 +37,10 @@ struct LocOffsets {
     LocOffsets copyWithZeroLength() const {
         return LocOffsets{beginPos(), beginPos()};
     }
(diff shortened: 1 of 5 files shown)


aprocter (approved):

froydnj:
> You can simplify 3 of the 4 changes in this PR to call your new helper directly on the core::Loc (those three are manually creating the Loc by passing the fileref in manually)

I'm not seeing this?  The `package.cc` and `Prop.cc` changes are zero-length'ing offsets and combining them with a file from someplace else?

PR #5328, opened 2022-02-18
add an autocorrect for `.compact` with Array-y type mismatches

### LINKED ISSUE #5325: Autocorrect to suggest using `Array#compact`

froydnj:
### Motivation

Fixes #5325.

### Test plan

See included automated tests.


core/TypeErrorDiagnostics.cc
```diff
@@ -55,19 +55,46 @@ void TypeErrorDiagnostics::maybeAutocorrect(const GlobalState &gs, ErrorBuilder
...
+            return;
+        }
+
+        if (expected->targs.size() != 1 || actual->targs.size() != 1) {
+            return;
+        }
+
+        auto &expectedElementType = expected->targs[0];
+        auto &actualElementType = actual->targs[0];
+        auto actualWithoutNil = Types::approximateSubtract(gs, actualElementType, Types::nilClass());
```
jez, line 77: This is a small thing but because you know the symbol for NilClass, you can call dropSubtypesOf directly, skipping the allocation (approximateSubtract will just immediately unwrap the TypePtr and call that function instead)

core/TypeErrorDiagnostics.cc
```diff
@@ -55,19 +55,46 @@ void TypeErrorDiagnostics::maybeAutocorrect(const GlobalState &gs, ErrorBuilder
...
-                if (classSymbol.exists() && classSymbol.data(gs)->owner == core::Symbols::root() &&
-                    classSymbol.data(gs)->name == core::Names::Constants::Boolean()) {
-                    e.replaceWith("Prepend `!!`", loc, "!!({})", loc.source(gs).value());
-                }
+        return;
+    }
+
+    // If we expected an array of T, but got an array of nilable T, add an autocorrect
+    // to add .compact.
+    auto arrayOfUntyped = Types::arrayOfUntyped();
```
jez, line 63: I try not to use untyped in normal (non-error recovery code) if I can avoid it. I probably would have written this to do the two cast operations first, then call derivesFrom Symbols::Array() on the klass field in the appliedtype. This also has the side benefit of more predictable performance (the cast check is much faster than an arbitrary call to check subtyping)

jez:
I noticed that this is still a draft but looks good! Looking forward to seeing the tests too

PR #4973, opened 2021-12-06
Revert "Introduce chained sig syntax (#4251)"

reese:
This reverts commit 5bb11d41ff71558605bad96059d9996942492b0a.

### Motivation

This change breaks in the Stripe codebase due to some bad interaction with the autoloader (although further investigation needs to happen to determine the full root cause) that causes multiple declarations to be in-flight at the same time.

### Test plan

I ran this branch internally and made sure it passes on current master, otherwise relying on existing tests to cover this.

PR #5329, opened 2022-02-18
Re-introduce chained `sig` syntax

KaanOzkan:
This PR re-introduces the [previously reverted chained signature syntax PR](https://github.com/sorbet/sorbet/pull/4973). Explanation of the chained syntax feature can be found in the [original PR](https://github.com/sorbet/sorbet/pull/4251).

The new change in this PR that reviewers can focus on is[ this commit](https://github.com/sorbet/sorbet/pull/5329/commits/84e151aed888c249a819725845b763c0c35a42e8). @jez was kind enough to give us a reproduction of the error [here](https://github.com/sorbet/sorbet/pull/5224/files) which helped us in identifying the problem. Currently in `sorbet-runtime` we [instantiate a new `DeclBuilder`](https://github.com/sorbet/sorbet/blob/771131400eee133ffbd77ce5f2484912f3843407/gems/sorbet-runtime/lib/types/private/methods/_methods.rb#L349) every time we execute the block and finalize that builder. However, the reverted PR [modified this](https://github.com/sorbet/sorbet/pull/4251/files#diff-84b528c5d74ba5b7d036023630f2c092865a65c9c3c0ed88b3cc342305759dbcL349) to reuse a builder since chained syntax uses the state stored in a `DeclarationBuilder`.

### Motivation

Chained `sig` syntax enables `abstract` to be chained into the sig parameter and makes this information available sooner, since sig blocks are evaluated lazily.

With this information at an earlier stage, it's possible to completely turn off method wrapping and unwrapping for non-abstract methods, when using default_checked_level = :never.

### Test plan

I have simplified the reproduction script and added it as a test:

When we call `DefineMethodWithSig.duplex()` we encounter the method signature and execute it. Upon this execution we encounter the type `CallsMethodUponLoading` and proceed to load that class. As the name suggests this class triggers the duplex method a second time upon loading. This second trigger initiates a second signature processing for `duplex` method which succeeds and returns. Afterwards, we continue processing the signature due to the original method call. At this point if we reuse the previous `DeclBuilder` instance we'd encounter the `You can't modify a signature declaration after it has been used` error since the previous builder was finalized.

Here's a diagram to better conceptualize the test case:

```mermaid
flowchart LR
A(DuplicateSigEvalTest) -->|calls| B("DefineMethodWithSig.duplex()")
B --> |execute sig| C(CallsMethodUponLoading)
C --> |calls| D("DefineMethodWithSig.duplex()")
D --> E(finalize!)
E --> |done| F[return]
F --> G(finalize!)
B --> |waiting| G
G --> |done| H[return]
```

cc @vinistock


gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
```diff
@@ -15,6 +29,31 @@ class BuilderError < StandardError; end
...
+      unless @inside_sig_block || decl.mod == T::Private::Methods::PROC_TYPE
+        raise BuilderError.new(
+          "Can't invoke #{method_name} outside of a signature declaration block"
+        )
+      end
+    end
+
+    def deep_clone
+      fresh_builder = self.class.new(decl.mod, decl.raw)
+      fresh_builder.decl = decl.dup
```
froydnj, line 52: I can't figure out what I would do differently here, but this code does not spark joy.

I am slightly concerned that in `deep_clone`, we're really not deep cloning that much: `decl.dup` seems to be the only piece of data that we're actually doing anything with, and we're not even dup'ing the pieces of `decl` used to form the new builder.

Also a little concerned that having this here means that we need to be careful about things we add to `Declaration`; we now need to decide whether we need to deal with them in this function -- and we need to remember to do that considering in the first place.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -346,11 +350,9 @@ def self.run_sig(hook_mod, method_name, original_method, declaration_block)
...
 
   def self.run_builder(declaration_block)
-    builder = DeclBuilder.new(declaration_block.mod, declaration_block.raw)
-    builder
-      .instance_exec(&declaration_block.blk)
-      .finalize!
-      .decl
+    builder = declaration_block.decl_builder.deep_clone
+    builder.run!(&declaration_block.blk)
+    builder.decl
```
froydnj, line 355: I agree that this will solve the problem raised by the original test case.

I am concerned that we're now making every signature building operation more expensive by allocating more.  My questions here would be:

1. Why wasn't the newly-added test case causing problems before?  Is that just because we changed the structure of the code, since what could be delayed until the execution of the `sig`'s block now has to be executed up-front?
2. Can we get back to the original behavior by restructuring the code?

gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
KaanOzkan: We can inline this in `run_builder` to match previous implementation. That might get rid of part of the code smells.

Another option is changing the `initialize` of `DeclBuilder` to take in a full `Declaration` Struct, (`decl` field). I'm not sure if it'll work with other callsites but just by looking at this usage, accepting a `Declaration` conceptually makes sense to me.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
froydnj: OK, it's very possible I'm misunderstanding something, but before, we were only ever allocating `DeclBuilder` here, in `run_builder`.  Now, we're allocating *two* `DeclBuilder` objects: one in `T::Private::Methods._declare_sig_internal` when starting the `sig`, and one here, when we're actually running it.

I can see the objects for parsing and storing the currently-being-parsed `sig` state have become somewhat more complicated, but it's not immediately obvious to me why we were able to get by before *without* duplicating information here.  And that's what I'm pushing on here: does having chained `sig` syntax necessarily require us to have duplication here when we didn't before?  If so, why?  And if not, can we rewrite the code so that we don't have duplication here?  (I think it would be OK if you wound up fusing together objects or similar; I don't think we are 100% tied to the current setup of how `sig` building is done.)

(I do see your excellent description of how the added testcase illustrates the need for the cloning that you do.  I'd wonder why we don't see the same issues with the current setup, and if we could rewrite things to be closer to the current setup.)

Full disclosure: I'm not 100% sure how much the efficiency of `sig` building matters.  It is solely (at least for Stripe) a load-time thing, but we also run...many parallel tests, so load time matters more than you might think.  But I don't have concrete numbers to say "an extra allocation will slow down loading by Y%", so I'm just going on the principle of "avoid making things slower unless you have to".

KaanOzkan: Thanks for the detailed reply @froydnj. I'm sorry I didn't understand it before, I think it was due to how similarly `DeclBuilder` and `DeclarationBlock` are named.

Creating a `DeclBuilder` in `_declare_sig_internal` is part of the core logic of this feature. We need it so that `sig` method can return a `DeclBuilder` instead of `nil`. As a result we can chain methods on it. Previously since we return `nil`, only impact of `declare_sig` was creating a `DeclarationBlock` and mark it as active. Later on we created a `DeclarationBlock` during execution. Therefore, no duplication of `DeclarationBlock` necessary. We simply didn't need that information earlier.

I'll try to see if we can solve the edge case through a different method. And if not, we might have to do a larger refactor of the feature. You are right to bring this up and personally I'd hate to slow things down or increase memory usage because we are not properly handling an edge case.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -346,11 +350,31 @@ def self.run_sig(hook_mod, method_name, original_method, declaration_block)
...
+  end
+
+  def self.register_live_decl_builder(declaration_block)
+    @live_decl_builders[declaration_block.mod.hash] = true
+
+    declaration_block.decl_builder
+  end
+
+  def self.unregister_live_decl_builder(declaration_block)
+    @live_decl_builders[declaration_block.mod.hash] = false
```
KaanOzkan, line 377: @froydnj This is our alternative solution to handle the edge case described in the PR documentation and the test case. You can look at my earlier response [here](https://github.com/sorbet/sorbet/pull/5329#discussion_r822086310) to  gain more context.

Instead of handling the edge case like the current implementation on `master`, we decided to keep track of the `DeclBuilder` that is being executed. We store them in a hash using [Object#hash](https://ruby-doc.org/core-3.1.1/Object.html#hash-method) as the key and ensure the same builder isn't already being processed for any given execution.

The early return inside `run_builder` prevents the edge case from happening entirely rather than dealing with it after the fact unlike the previous implementation.

froydnj: Thanks for looking into this @KaanOzkan!

Two issues with this implementation:

1. `@live_decl_builders` grows essentially without bound.  I guess it's bounded by the number of signatures in the program, but that can still be pretty substantial.  I think a better implementation would remove the entry in the hash table in the `ensure` block.
2. The documentation for `Object.hash` says that `a.eql?(b)` implies `a.hash == b.hash`, but the reverse is not true: `a.hash == b.hash` doesn't imply anything about the equality of `a` and `b`.  So I would be worried that we could have `DeclarationBlock`s for two different modules, the modules have the same hash, and we would be blocked from running `sig`s for both modules at the same time.  I guess the table is indexed by `mod.hash` instead of by `mod` in the interest of not retaining references to the modules...or the declaration block?  Fortunately, the solution to the first problem would fix the second problem.

froydnj: (I still think it ought to be possible to write things such that we don't need re-entrancy hacks like this, since the previous implementation of `sig` was able to get by without it...)

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -345,12 +352,34 @@ def self.run_sig(hook_mod, method_name, original_method, declaration_block)
...
+  def self.register_method_being_processed(declaration_block, method_name)
+    key = method_owner_and_name_to_key(declaration_block.mod, method_name)
+    @methods_being_processed[key] = true
+
+    declaration_block.decl_builder
+  end
+
+  def self.unregister_method_being_processed(mod, method_name)
+    key = method_owner_and_name_to_key(mod, method_name)
+    @methods_being_processed.delete(key)
```
froydnj, line 382: It's not exactly cheap to continually compute `method_owner_and_name_to_key`, and we're doing it three times in `run_builder`.  Can we at least compute the key once, and pass it in to each of these functions?

But why are we computing such a complex key?  Is the module itself not enough?  Or is the concern that we might wind up doing signature building for the same module, but in different files and different methods, and we really might need to do both of them simultaneously?  Is there a test for that?

froydnj (changes requested):

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
KaanOzkan: Let me know if you want this test added to the PR.This is the diff against the `test` commit on this branch:

```
gems/sorbet-runtime/test/types/fixtures/duplicate_sig_eval/define_method_with
_sig.rb
@@ -4,5 +4,8 @@ class DefineMethodWithSig

     sig {params(arg: T.nilable(T.class_of(Opus::Types::Test::DuplicateSigEvalSandb
ox::CallsMethodUponLoading))).void}
     def self.duplex(arg); end
+
+    sig { raise "sig is evaluated" }
+    def self.foo; end
   end
 end

gems/sorbet-runtime/test/types/fixtures/duplicate_sig_eval/calls_method_upon_
loading.rb
@@ -1,5 +1,5 @@
 module Opus::Types::Test::DuplicateSigEvalSandbox
   class CallsMethodUponLoading
-    Opus::Types::Test::DuplicateSigEvalSandbox::DefineMethodWithSig.duplex(self)
+    Opus::Types::Test::DuplicateSigEvalSandbox::DefineMethodWithSig.foo
   end
 end
```

As before we'd trigger `duplex` from the test runner. Which would load the sig for `duplex` which in turn loads the `CallsMethodUponLoading`. However, now that class calls the `foo` method instead of `duplex`.

Without a method specific key the current implementation would not evaluate the sig for `foo` and not raise.

froydnj: Can we have both the test that has `duplex` and the test that has `foo`?  I don't see why it has to be one or the other.

froydnj:

→ https://go/builds/bui_LL1sZsa9n8AR0C
→ https://go/builds/bui_LL1s1nYMk9zWGd

froydnj:
(FWIW, the suggested reproducer on stripe's infrastructure is to require a specific file, then force running all sig blocks via `T::Utils.run_all_sig_blocks`, which might also be enlightening on something like Shopify's codebase--but maybe you already tested on that...)

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
KaanOzkan: Added the test. Used `check_level` instead of `raise` to ensure signature was fully evaluated.

I can take a closer look at the stacktrace you provided however, upon first glance it seems consistent with the initial reproduction steps by Jake. I did test this implementation with `T::Utils.run_all_sig_blocks` before and it was working 🤔 .

froydnj:
PR #5641 should have a failing testcase similar to at least one of the failures in Stripe's codebase.  On that branch:

```
st-froydnj1:sorbet froydnj$ cd gems/sorbet-runtime/
st-froydnj1:sorbet-runtime froydnj$ bundle exec ruby main.rb
st-froydnj1:sorbet-runtime froydnj$
```

with this PR:

```
st-froydnj1:sorbet-runtime froydnj$ bundle exec ruby main.rb
	57: from main.rb:17:in `<main>'
	56: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:258:in `block in _on_method_added'
	55: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:447:in `maybe_run_sig_block_for_key'
	54: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:467:in `run_sig_block_for_key'
	53: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:248:in `block in _on_method_added'
	52: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:338:in `run_sig'
	51: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:361:in `run_builder'
	50: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb:71:in `run!'
	49: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb:71:in `instance_exec'
	48: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/ms.rb:10:in `block in <class:M>'
	47: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/ms.rb:10:in `require'
	46: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d1.rb:1:in `<top (required)>'
	45: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d1.rb:2:in `<module:Repro>'
	44: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d1.rb:3:in `<class:D1>'
	43: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d1.rb:3:in `require'
	42: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d2.rb:1:in `<top (required)>'
	41: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d2.rb:2:in `<module:Repro>'
	40: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d2.rb:3:in `<module:D2>'
	39: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d2.rb:3:in `new'
	38: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:277:in `block in _on_method_added'
	37: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb:161:in `validate_call'
	36: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb:161:in `bind_call'
	35: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:25:in `initialize'
	34: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:258:in `block in _on_method_added'
	33: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:447:in `maybe_run_sig_block_for_key'
	32: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:467:in `run_sig_block_for_key'
	31: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:248:in `block in _on_method_added'
	30: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:346:in `run_sig'
	29: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:397:in `build_sig'
	28: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:397:in `new'
	27: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/signature.rb:47:in `initialize'
	26: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/utils.rb:8:in `coerce'
	25: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/types/type_alias.rb:13:in `aliased_type'
	24: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:13:in `block in <class:HC>'
	23: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/_types.rb:36:in `nilable'
	22: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/utils.rb:8:in `coerce'
	21: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/types/type_alias.rb:13:in `aliased_type'
	20: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:6:in `block in <class:HC>'
	19: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:6:in `require'
	18: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:1:in `<top (required)>'
	17: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:2:in `<module:Repro>'
	16: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:12:in `<class:I>'
	15: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:6:in `make'
	14: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:6:in `new'
	13: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation_2_7.rb:389:in `block in create_validator_procedure_fast0'
	12: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation_2_7.rb:389:in `bind_call'
	11: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:25:in `initialize'
	10: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:258:in `block in _on_method_added'
	 9: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:447:in `maybe_run_sig_block_for_key'
	 8: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:467:in `run_sig_block_for_key'
	 7: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:248:in `block in _on_method_added'
	 6: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:338:in `run_sig'
	 5: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:361:in `run_builder'
	 4: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb:71:in `run!'
	 3: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb:71:in `instance_exec'
	 2: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:17:in `block in <class:HC>'
	 1: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb:93:in `returns'
/Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb:28:in `check_live!': You can't modify a signature declaration after it has been used. (T::Private::Methods::DeclBuilder::BuilderError)
	54: from main.rb:17:in `<main>'
	53: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:258:in `block in _on_method_added'
	52: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:447:in `maybe_run_sig_block_for_key'
	51: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:467:in `run_sig_block_for_key'
	50: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:248:in `block in _on_method_added'
	49: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:338:in `run_sig'
	48: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:361:in `run_builder'
	47: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb:71:in `run!'
	46: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb:71:in `instance_exec'
	45: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/ms.rb:10:in `block in <class:M>'
	44: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/ms.rb:10:in `require'
	43: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d1.rb:1:in `<top (required)>'
	42: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d1.rb:2:in `<module:Repro>'
	41: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d1.rb:3:in `<class:D1>'
	40: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d1.rb:3:in `require'
	39: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d2.rb:1:in `<top (required)>'
	38: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d2.rb:2:in `<module:Repro>'
	37: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d2.rb:3:in `<module:D2>'
	36: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/d2.rb:3:in `new'
	35: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:277:in `block in _on_method_added'
	34: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb:161:in `validate_call'
	33: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb:161:in `bind_call'
	32: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:25:in `initialize'
	31: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:258:in `block in _on_method_added'
	30: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:447:in `maybe_run_sig_block_for_key'
	29: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:467:in `run_sig_block_for_key'
	28: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:248:in `block in _on_method_added'
	27: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:346:in `run_sig'
	26: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:397:in `build_sig'
	25: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:397:in `new'
	24: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/signature.rb:47:in `initialize'
	23: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/utils.rb:8:in `coerce'
	22: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/types/type_alias.rb:13:in `aliased_type'
	21: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:13:in `block in <class:HC>'
	20: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/_types.rb:36:in `nilable'
	19: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/utils.rb:8:in `coerce'
	18: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/types/type_alias.rb:13:in `aliased_type'
	17: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:6:in `block in <class:HC>'
	16: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:6:in `require'
	15: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:1:in `<top (required)>'
	14: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:2:in `<module:Repro>'
	13: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:12:in `<class:I>'
	12: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:6:in `make'
	11: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/i.rb:6:in `new'
	10: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation_2_7.rb:389:in `block in create_validator_procedure_fast0'
	 9: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/call_validation_2_7.rb:389:in `bind_call'
	 8: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:25:in `initialize'
	 7: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:258:in `block in _on_method_added'
	 6: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:447:in `maybe_run_sig_block_for_key'
	 5: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:467:in `run_sig_block_for_key'
	 4: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:248:in `block in _on_method_added'
	 3: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:337:in `run_sig'
	 2: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:340:in `rescue in run_sig'
	 1: from /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/configuration.rb:202:in `sig_builder_error_handler'
/Users/froydnj/stripe/sorbet/gems/sorbet-runtime/lib/types/configuration.rb:195:in `sig_builder_error_handler_default': /Users/froydnj/stripe/sorbet/gems/sorbet-runtime/hc.rb:17: Error interpreting `sig`: (ArgumentError)
  You can't modify a signature declaration after it has been used.
```

KaanOzkan:
@froydnj Given the lazy evaluation of `type_alias` your last repro was hard to get around with our existing solutions in mind. As a result we ended up going back to how it's currently processed in Sorbet. Only difference being the duplication of the internal `decl` state to keep it persistent between the multiple instances of `DeclBuilder`: https://github.com/sorbet/sorbet/pull/5329/commits/f8c8386cef3c59714ef1ee591e0a2d28d406f145

I also have a CI linter failure but couldn't replicate it locally. Have you encountered [this](https://buildkite.com/sorbet/sorbet/builds/24276#0182b6b1-cbad-45a5-a44b-3cb14d2d83cb/174-204) failure before?

core/errors/rewriter.h
```diff
@@ -15,5 +15,6 @@ constexpr ErrorClass ComputedBySymbol{3509, StrictLevel::False};
 constexpr ErrorClass InitializeReturnType{3510, StrictLevel::False};
 constexpr ErrorClass InvalidStructMember{3511, StrictLevel::False};
 constexpr ErrorClass NilableUntyped{3512, StrictLevel::False};
+constexpr ErrorClass InvalidChainedSig{3513, StrictLevel::False};
```
froydnj, line 18: We have added a lint that verifies that all error codes have corresponding documentation in `error-reference.md`, so you'll have to write documentation for this error.

You should be able to reproduce the error by running `tools/scripts/check_error_classes.sh`; if you can't reproduce the error locally, that seems like a bug, and we should fix that separately.

KaanOzkan: Thanks Nathan, CI output made me think it was during `yarn lint`. CI is fixed now.

froydnj:
Green test on Stripe's CI: http://go/builds/bui_NmnpNMSzpFHw0r

It doesn't look like this PR noticeably affects test performance (one datapoint is a not a great way to measure this, but at least the performance of the above run was not noticeably different from historical data over the last week or so).

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,178 @@
...
+
+        return false;
+    }
+
+    void checkDuplicates(core::MutableContext ctx, ast::Send *sigBlock, core::NameRef fun) {
+        while (sigBlock) {
+            if (sigBlock->fun == fun) {
+                if (auto e = ctx.beginError(sigBlock->loc, core::errors::Rewriter::InvalidChainedSig)) {
+                    e.setHeader("Duplicate invocation of `{}` in signature declaration", fun.toString(ctx));
+                }
```
froydnj, line 95: I think we can give a better location for the error here, since `sigBlock` has the location for the actual function name.  We can also provide in a note the original location for the method name.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    bool invalidChainedStatement(core::MutableContext ctx, ast::Send *send) {
+        if (send->fun == core::Names::params() || send->fun == core::Names::returns() ||
+            send->fun == core::Names::void_() || send->fun == core::Names::bind() ||
+            send->fun == core::Names::checked() || send->fun == core::Names::onFailure() ||
+            send->fun == core::Names::typeParameters()) {
+            if (auto e = ctx.beginError(send->loc, core::errors::Rewriter::InvalidChainedSig)) {
+                e.setHeader("Cannot use `{}` outside of a sig block", send->fun.toString(ctx));
+            }
```
froydnj, line 109: Using `send->funLoc` here would be a better experience for the user.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,178 @@
...
+
+            sendCopy = ast::cast_tree<ast::Send>(sendCopy->recv);
+
+            // If a previous receiver has a block, then we need to add an error to prevent signatures like this:
+            // `sig.final {}.override{}
+            if (sendCopy->hasBlock()) {
+                if (auto e = ctx.beginError(send->loc, core::errors::Rewriter::InvalidChainedSig)) {
+                    e.setHeader("Cannot add more signature statements after the declaration block");
+                }
```
froydnj, line 154: Highlighting the entire send here doesn't give the user a clear picture of where the error is; we should be able to highlight the problematic part of the send directly, and (maybe) provide information about the original declaration block.

test/testdata/rewriter/chained_sig.rb
```diff
@@ -0,0 +1,182 @@
...
+                   # ^^^^^^^ error: Method `checked` does not exist on `NilClass`
+  def foo; end
+end
+
+class ValidDoubleChain
+  extend T::Sig
+  include Interface
+
+  sig.override.final { void }
+  def foo; end
```
froydnj, line 181: We also need tests -- ideally for all the cases that use `.final` combined with something else -- that the `sig(:final)` syntax produces the same result.  So here, we would want a separate test that uses `sig(:final).override { void }`.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        if (send->fun == core::Names::sig() && !send->hasBlock() && send != previousSigSend) {
+            if (auto e = ctx.beginError(send->loc, core::errors::Rewriter::InvalidChainedSig)) {
+                e.setHeader("Signature declarations expect a block");
+                e.addErrorNote("Complete the signature by adding a block declaration: sig `{}`", "{ ... }");
+            }
+
+            return tree;
+        }
```
froydnj, line 37: When does this `send != previousSigSend` actually wind up firing during the tree traversal?  I'm trying to figure out whether there's a better way to handle all of this.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,178 @@
...
+    ast::Send *previousSigSend;
+
+    ast::ExpressionPtr preTransformSend(core::MutableContext ctx, ast::ExpressionPtr tree) {
+        auto *send = ast::cast_tree<ast::Send>(tree);
+        ast::Send *sigSend = ast::cast_tree<ast::Send>(send->recv);
+
+        // Make sure the first receiver is sig. E.g.: `sig.abstract {}`, `sig.override.final {}`
+        if (send->fun != core::Names::sig() && (sigSend == nullptr || !firstReceiverIsSig(sigSend))) {
+            return tree;
+        }
```
froydnj, line 23: Let's be more precise in our terminology here: it's not that the receiver is `sig`, but that we're looking for a send that invokes `sig`.  That applies to the comment here and `firstReceiverIsSig`.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,178 @@
...
+
+        // This is still an invocation where `sig` is the first receiver, but there is still no block. We need to pop
+        // the last entry for incomplete sigs and add the current send to account for multiple methods chained on sig
+        //
+        // E.g.: `sig.override.final {}`
+        //
+        // If we never find a send where the first receiver is a `sig` and the last invocation has the declaration
+        // block, then we add the error later on
```
froydnj, line 51: Same comment about precision in terminology here.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,178 @@
...
+
+        // For all other cases, we have to re-write the block
+        // E.g.: `sig.abstract { void }` -> `sig { abstract.void }`
+        auto *block = send->block();
+        auto blockBody = ast::cast_tree<ast::Send>(block->body);
+
+        // If the blockBody is not a send, then we have a sequence of instructions inside the signature block
+        if (blockBody == nullptr) {
+            if (auto e = ctx.beginError(block->loc, core::errors::Rewriter::InvalidChainedSig)) {
+                e.setHeader("Malformed signature: cannot have multiple instructions inside a signature block");
```
froydnj, line 69: I think we should use "expressions" rather "instructions" in the comment here, since that corresponds to the terminology of the AST more closely.  ("instructions" are a CFG thing in Sorbet.)

I don't know what the best terminology to use in the actual error message is; I think "statements" is the best word to use.  (I think I could be convinced that "expressions" is the right thing...maybe.)

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,178 @@
...
+            }
+            send = ast::cast_tree<ast::Send>(send->recv);
+        } while (send);
+
+        return false;
+    }
+
+    void checkDuplicates(core::MutableContext ctx, ast::Send *sigBlock, core::NameRef fun) {
+        while (sigBlock) {
+            if (sigBlock->fun == fun) {
```
froydnj, line 92: `sigBlock` being a `Send` makes for confusing code; let's call this `send` instead, or perhaps `sendFromSigBlock` if we think the provenance of the send is important.

test/testdata/rewriter/chained_sig.rb
```diff
@@ -0,0 +1,182 @@
...
+  def baz; end
+end
+
+class DuplicateButMissingBlock
+  extend T::Sig
+  include Interface
+
+  sig.final { void }.override
+                   # ^^^^^^^^ error: Method `override` does not exist on `NilClass`
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Signature declarations expect a block
```
froydnj, line 162: This second error seems not great, since we *are* providing a block, it's just that the entire declaration here is not in the format that we expect.  Can we do better than this?

core/errors/rewriter.h
```diff
@@ -17,5 +17,6 @@ constexpr ErrorClass InvalidStructMember{3511, StrictLevel::False};
 constexpr ErrorClass NilableUntyped{3512, StrictLevel::False};
 constexpr ErrorClass HasAttachedClassInClass{3513, StrictLevel::False};
 constexpr ErrorClass ContravariantHasAttachedClass{3514, StrictLevel::False};
+constexpr ErrorClass InvalidChainedSig{3515, StrictLevel::False};
```
froydnj, line 20: I am not sure whether it would be better to split this error code up or not.

website/docs/error-reference.md
```diff
@@ -542,6 +542,17 @@ Ruby `class`. For more, see the docs for
 The `has_attached_class!` annotation cannot be given a contravariant `:in`
 annotation because `T.attached_class` is only allowed in output positions.
 
+## 3515
+
+You are using the chained `sig` syntax however it is malformed. Ensure a block
+is passed and there are no duplicate chained methods. You can only chain
+`abstract`, `final`, `override` and `overridable` methods on a `sig`.
```
froydnj, line 549: I think this description is written from an implementor's perspective, and not from what would be most helpful for the user.  (e.g. what does "duplicate chained methods" mean?  the error about duplicates is for the same method being invoked inside and outside of the block, but "duplicate chained methods" sounds like it might be more like `sig.abstract.abstract`)

froydnj (changes requested):
Sorry for sleeping on this for so long; I completely lost track of this PR.

There are enough things to adjust that I'm going to send this back for now; I'll try to take a second look at this iteration later, or review the next iteration more promptly.

test/testdata/rewriter/chained_sig.rb
KaanOzkan: I'm not sure. We only keep track of the previous `sig` send and not `sig.final`.

What do you think of changing the loc to `funLoc` instead and underlying `override`. This will also affect cases like `sig.final`

core/errors/rewriter.h
KaanOzkan: I think the idea was to keep one error per rewriter. Others seem to be similar but we are raising more than them (6 instances).

website/docs/error-reference.md
KaanOzkan: I could modify that sentence a bit. But if we want detailed descriptions we would have to break down the error into multiple smaller ones.

rbi/sorbet/builder.rbi
```diff
@@ -3,20 +3,37 @@ class T::Private::Methods::DeclBuilder
...
   def type_parameters(*params); end
 
-  sig {returns(T::Private::Methods::DeclBuilder)}
-  def abstract; end
-
-  sig {returns(T::Private::Methods::DeclBuilder)}
-  def final; end
+  sig do
+    params(
+      blk: T.nilable(T.proc.bind(T::Private::Methods::DeclBuilder).void)
```
froydnj, line 8: If these methods require a block (which we have tests for in other places), why is the block argument marked as nilable?  (Other methods in this RBI have the same issue.)  Are we trying to avoid double errors, one from the rewriter pass, and one from the typechecking pass?  If that's the only reason, I think we should just accept the duplicate errors and make these signatures correct by removing `T.nilable`.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,184 @@
...
+        auto rv = ast::MK::Send(send->loc, std::move(sendCopy->recv), core::Names::sig(), send->funLoc, args.size(),
+                                std::move(args));
+        ast::cast_tree_nonnull<ast::Send>(rv).setBlock(ast::MK::Block0(send->block()->loc, std::move(newBody)));
+        return rv;
+    }
+};
+
+ast::ExpressionPtr ChainedSig::run(core::MutableContext &ctx, ast::ExpressionPtr tree) {
+    ChainedSigWalk walker;
+    return ast::TreeMap::apply(ctx, walker, std::move(tree));
```
froydnj, line 182: Please rewrite this to use `TreeWalk` instead.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,184 @@
...
+
+        // Go through each part of the chained sig and make sure it's not duplicated inside the block
+        // E.g.: `sig.abstract { abstract.void }`
+        while (sendCopy && sendCopy->fun != core::Names::sig()) {
+            checkDuplicates(ctx, blockBody, sendCopy->fun, sendCopy->funLoc);
+
+            sendCopy = ast::cast_tree<ast::Send>(sendCopy->recv);
+        }
+
+        sendCopy = ast::cast_tree<ast::Send>(treeCopy);
```
froydnj, line 135: Can we use a separate, named variable for the loop iterator here so that we don't have to re-assign `sendCopy` after the loop, please?

test/testdata/rewriter/chained_sig.rb
```diff
@@ -0,0 +1,208 @@
...
+  sig.final { void }
+  def foo; end
+end
+
+class InvalidInvocationsOnSig
+  extend T::Sig
+
+  sig.params(a: Integer)
+    # ^^^^^^ error: Cannot use `params` outside of a sig block
+    # ^^^^^^ error: Method `params` does not exist on `NilClass`
```
froydnj, line 48: Where are these `NilClass` errors coming from?

test/testdata/rewriter/chained_sig.rb
```diff
@@ -0,0 +1,208 @@
...
+
+class DuplicateInvocations
+  extend T::Sig
+  extend T::Helpers
+  include Interface
+
+  abstract!
+
+  sig.overridable { overridable.void }
+                  # ^^^^^^^^^^^ error: Duplicate invocation of `overridable` in signature declaration
```
froydnj, line 149: `resolver/type_syntax.cc` is already checking for duplicate invocations of some (but not all) methods.

WDYT about making `type_syntax.cc`'s checking more complete (as a separate PR), and then we don't have to have the duplicate checks in the `ChainedSig` rewriter?  Ideally the locations to get good error messages will just happen naturally with chained sigs too.

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,184 @@
...
+
+using namespace std;
+namespace sorbet::rewriter {
+
+struct ChainedSigWalk {
+    // The previousSigSend is used to allow us to add errors to `sig` with missing block without accidentally adding it
+    // to a chained sig. E.g.:
+    // `sig` -> error no block
+    // `sig.final {}` -> no error. Need to prevent adding the error when we receive `sig` by itself
+    ast::Send *previousSigSend;
```
froydnj, line 14: Let's remove this and have `firstSendIsSig` return the relevant `Send*`, if any, rather than setting state shared between functions.  That way we can re-use that returned `Send*` later in the function, for determining whether we have a `final` method, I think?

rewriter/ChainedSig.cc
```diff
@@ -0,0 +1,184 @@
...
+            return tree;
+        }
+
+        // Return early unless it's one of the sends we are interested in
+        if (send->fun != core::Names::abstract() && send->fun != core::Names::final_() &&
+            send->fun != core::Names::override_() && send->fun != core::Names::overridable()) {
+            return tree;
+        }
+
+        // TODO: This comment looks outdated
```
froydnj, line 46: Can we address this TODO, please?

test/testdata/rewriter/chained_sig.rb
```diff
@@ -0,0 +1,208 @@
...
+
+  sig.abstract { void }
+  def baz; end
+end
+
+class ValidDoubleChain
+  extend T::Sig
+  include ValidInterface
+
+  sig.override.final { void }
```
froydnj, line 200: Can we have a `sig.final.override` test too, please?

We should also ensure that attempting to inherit from `ValidDoubleChain` and override these `final` methods produces errors.

froydnj (changes requested):
Thanks for your patience as this PR intersected with various end-of-year things at Stripe.

I just went through the C++ side of things; I haven't looked at the sorbet-runtime bits.  There are enough things that I think this is going to need another pass.

KaanOzkan:
@froydnj I addressed your latest rewriter comments, there are still some discussions points from earlier. Also this PR is growing quite large with conflicting commits. Let me know if you'd like me to clean up the git history (undo the revert revert and make atomic commits) before reviewing. Thank you!

jez:
Our test script failed to post a comment, here's the link for Stripe employees:

https://cibot.corp.stripe.com/builds/pay-server--bui_PkA6P0eGxlm9Tc

jez (commented):
Hey sorry for the late response—we were at an onsite all week, and I was trying to be present with the team.

I didn't have time to finish up a review of this, and it's the end of my week now. I'll continue this next week, but before you spend too much time trying to action this feedback, I think that seeing the implementation in detail for the first time is making me have some questions, especially around

- Is the syntax actually so good that we want to inflict the pain of migrating on all Sorbet users? If we're going to force a migration on people, is this the alternative we want?

- Is there a way we can evaluate what the eventual performance characteristics of this (or alternatives syntaxes) will be? Do we have to do the migration to find out? Thinking about both how this changes the static and runtime performance (e.g. the static adds another tree walk and extra allocations). I have some things I can do Monday to start answering those questions but I didn't get to that today.

- At Stripe we know that we're essentially going to need some syntax to declare that a method overrides another method without specifying types for a sig. That could be a syntax like `override def foo` or like `sig.override def foo` or maybe something else. I will have to figure out how the changes in this PR (especially in the runtime) align with whatever changes we will need to make in the future. That involves me going into the codebase and trying to prototype some things, which I will have to make time for.

Thanks again for your patience with this change, and hopefully you'll hear back from me early next week.

PR #5333, opened 2022-02-20
Adds Hash#except for Ruby 3

swiknaba:
Ruby 3  added `Hash#except`. Sorbet does not know yet. This PR attempts to change this.
See https://ruby-doc.org/core-3.0.0/Hash.html#method-i-except

This method should not be allowed for Ruby < 3, since when I'm outside a Rails/ActiveSupport context `Hash#except` simply does not exist.
Unsure if wrapping the method in a "unless Ruby 3" block is the prefered way.

### Motivation

This brings Sorbet closer to support Ruby 3 features, thus allowing to use `Hash#except` (when not using Rails, where `Hash#except` comes from ActiveSupport).

### Test plan

I couldn't find any unit tests for the other instance methods on Hash. Also unsure on how to test against different Rubies (it seems tests are run against Ruby 2.6 only?).

Head e.g. to sorbet.run, and try:

```ruby
some_hash = {
 foo: 'bar',
 bar: 'foo',
}

some_hash.except(:foo)
```

this will currently throw a type error. Given that the sandbox is (presumably) Ruby 2.x, this is actually correct.
It would be amazing, if we could choose the (major) Ruby version for the sandbox.

---

I felt opening an issue to ask for this is actually more effort than opening a PR, since this is effectively ~2 LOC.

### DIFF
--- a/rbi/core/hash.rbi
+++ b/rbi/core/hash.rbi
@@ -696,15 +696,13 @@ class Hash < Object
   # h = { a: 100, b: 200, c: 300 }
   # h.except(:a)          #=> {:b=>200, :c=>300}
   # ```
-  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.0.0")
-    sig do
-      params(
-          args: K,
-      )
-      .returns(T::Hash[K, V])
-    end
-    def except(*args); end
+  sig do
+    params(
+        args: K,
+    )
+    .returns(T::Hash[K, V])
   end
+  def except(*args); end
 
   # Returns a new array that is a one-dimensional flattening of this hash. That
   # is, for every key or value that is an array, extract its elements into the
--- a/rbi/core/hash.rbi
+++ b/rbi/core/hash.rbi
@@ -689,6 +689,23 @@ class Hash < Object
   # ```
   def fetch_values(*_); end
 
+  # Returns a new Hash excluding entries for the given keys.
+  # [`Hash#except`](https://ruby-doc.org/core-3.0.0/Hash.html#method-i-except)
+  #
+  # ```ruby
+  # h = { a: 100, b: 200, c: 300 }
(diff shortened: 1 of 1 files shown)


rbi/core/hash.rbi
```diff
@@ -689,6 +689,23 @@ class Hash < Object
...
   def fetch_values(*_); end
 
+  # Returns a new Hash excluding entries for the given keys.
+  # [`Hash#except`](https://ruby-doc.org/core-3.0.0/Hash.html#method-i-except)
+  #
+  # ```ruby
+  # h = { a: 100, b: 200, c: 300 }
+  # h.except(:a)          #=> {:b=>200, :c=>300}
+  # ```
+  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.0.0")
```
froydnj, line 699: We don't try to differentiate between Ruby versions when writing RBI definitions.  It's true that this leaves us open to accepting code that won't actually run on older Ruby versions, but it hasn't seemed to be a problem in practice.  (Additionally, since RBI files aren't actually executed, this check won't have any effect.)

Please remove the conditional around this definition and re-indent.

froydnj (changes requested):
Thanks for the PR, and for just submitting a PR to fix the issue rather than opening a separate issue!

You already fixed the issue with signature/definition argument name matching; there's one more thing to fix (+ a run of this on Stripe's codebase, which I'll take care of) that I've noted below.

rbi/core/hash.rbi
swiknaba: Thanks for the thorough explanation! Pushed a fix.

froydnj:

→ https://go/builds/bui_LC2ES7weiSFVIb
→ https://go/builds/bui_LC2EGQ4uW6CYbN

froydnj (approved):
Thanks1

PR #5335, opened 2022-02-22
Fix the signature for  to accept a nilable argument

dougedey-shopify:
### Motivation

I noticed that the `Kernel.exit!` method required an argument when being typechecked, but `Kernel.exit` does not, both methods take the same arguments

https://github.com/ruby/ruby/blob/37d5890e4941cedf6918821b29bb4a7e3a092e62/process.c#L4486-L4529

### Test plan

I couldn't see any specific tests for the exit commands, and I didn't think you want every signature tested


froydnj:

→ https://go/builds/bui_LCOOQwJMEAkR9a
→ https://go/builds/bui_LCOOQbQRAKJXJX

Morriar (commented):
It seems that with Ruby 3.1 both methods take a `*args`:

image

Source: https://ruby-doc.org/core-3.1.0/Kernel.html#method-i-exit

froydnj (approved):
Thanks!

(Note that we're taking an optional argument here, not a nilable argument; "nilable" has a specific meaning in Sorbet.  RBI files don't have a good way to express optional/defaulted arguments, so we use `T.unsafe(nil)` as a way of saying "some default value that conforms to what the method signature says".)

dougedey-shopify:
@froydnj Thanks for the clarification!

PR #5338, opened 2022-02-22
Mark type aliases as todo() items to delay resolution after stubbing

elliottt:
For type aliases that have failed to resolve in single-package mode, unconditionally mark them as `todo` so that they are delayed until after all constants have been stubbed.

### Motivation

Stabilizing single-package rbi generation.

### Test plan

See included automated tests.


jez:
Do you mind explaining this a little more? I'm curious to learn more about why this patch fixes the bug. (And it's been a while since I looked at resolver)

jez:
(sorry, misclick 🤦)

test/cli/rbi-gen-single/rbi-gen-single.out
```diff
@@ -6,8 +6,11 @@
 class Family::Simpsons < Object
   sig {returns(Bart::Character)}
   def bart; end
+  sig {returns(T.nilable(Bart::Character))}
+  def no_bart; end
```
elliottt, line 10: I don't think it's a deal-breaker that the alias didn't stick around, but I'll at least figure out why it's happening.

elliottt:
> Do you mind explaining this a little more? I'm curious to learn more about why this patch fixes the bug. (And it's been a while since I looked at resolver)

We resolve simple type aliases early during the `ResolveConstantsWalk` when possible, and leave everything else to the `ResolveTypeMembersAndFieldsWalk`. Type aliases to stubbed constants look like errors in `ResolveConstantsWalk` as they refer to constants that don't exist at that point. Since we can only stub those constants once we know they don't exist (the call site of the `ResolveConstantsWalk::constantResolutionFailed` function) we need to delay processing of type aliases that might involve them as well. Conveniently, marking them as a todo is the mechanism for delaying them to `ResolveTypeMembersAndFieldsWalk`, which allows them to resolve to the stubbed constants we've introduced in `constantResolutionFailed`.

jez:
Thanks!

froydnj (approved):
Makes sense to me!

PR #5342, opened 2022-02-23
Traverse generic bounds during rbi generation

elliottt:
This was an oversight in the original implementation: bounds weren't being explicitly printed, and seemed like they wouldn't affect the external interface of a package. This was corrected in a follow up PR that explicitly traverses the lower and upper bounds of the `type_member`, but it seems good to traverse them here as well so that we don't accidentally miss a dependency.

### Motivation

Stabilizing rbi generation.

### Test plan

It's difficult to write a test for this as the `LambdaParam`'s bounds are being traversed explicitly in `RBIExporter::typeMemberDetails`; this change is more for consistency.

### DIFF
--- a/packager/rbi_gen.cc
+++ b/packager/rbi_gen.cc
@@ -495,10 +495,9 @@ private:
                 break;
             }
             case core::TypePtr::Tag::LambdaParam: {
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Makes sense to me!

PR #5345, opened 2022-02-23
Automatically enable VS Code extension if sorbet/config exists

vinistock:
Enable the VS Code extension by default if a `sorbet/config` file exists in the workspace.

I used the `DefaultSorbetWorkspaceContext` so that we can test it out.

### Motivation

Having the extension enabled by default when detecting Sorbet is a nicer experience than having to manually turn it on for all projects using Sorbet.

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -4,7 +4,7 @@
   "description": "Ruby IDE features, powered by Sorbet.",
   "author": "Stripe Inc.",
   "license": "Apache-2.0",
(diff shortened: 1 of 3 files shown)


jez (approved):
Looks great thanks! If you increment the version in package json it will auto publish from master

vinistock:
Thank you! Bumped the version to `0.3.10`.

elliottt (approved):

PR #5348, opened 2022-02-23
Fix up `vsce` publish

jez:
### Motivation

We were getting master failures that looked like this:

<https://buildkite.com/sorbet/sorbet/builds/20755#a10f9290-b9a6-4018-85b5-76dd568b0594/388-402>

It looks like there is about a 4+ minute lag where the VS Code marketplace runs
some analysis on the extension before marking it fully published. During that
time, `vsce show` will not show the pending version, but `vsce publish` with
that version will throw an error.

We can grep for that error to handle these cases.

### Test plan

Test in prod. Won't be fully tested until the next time we go to bump the
version number, but if we want we can do that in this PR.


elliottt (approved):
🤞

PR #5350, opened 2022-02-24
Invert compare_indent_level

jez:
### Motivation

It used to be the case that for something like

    def foo
      if
    end

if you were comparing the indent level of the `if` token (which is 2
spaces) and the `end` token (which is 0 spaces), that

    compare_indent_level(ifTok, endTok) == -1

which is weird because typically `(A <=> B) < 0` is true if `A < B` but
in this case we were effectively comparing `2 < 0` and saying that was
true.

It was just the case that I had also inverted the `>` in the one places
where this function is called, so it didn't matter.

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/cc/lexer.rl
+++ b/parser/parser/cc/lexer.rl
@@ -205,11 +205,11 @@ int lexer::compare_indent_level(token_t left, token_t right) {
         auto rightIsSpace = rightChar == ' ' || rightChar == '\t';
 
         if (leftIsSpace && !rightIsSpace) {
-            return 1; // left < right
+            return 1; // left > right
         } else if (!leftIsSpace && !rightIsSpace) {
             return 0; // left == right
         } else if (!leftIsSpace && rightIsSpace) {
-            return -1;  // left > right
+            return -1;  // left < right
         }
 
         if (leftChar != rightChar) {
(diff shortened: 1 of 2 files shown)


parser/parser/cc/lexer.rl
```diff
@@ -205,11 +205,11 @@ int lexer::compare_indent_level(token_t left, token_t right) {
         auto rightIsSpace = rightChar == ' ' || rightChar == '\t';
 
         if (leftIsSpace && !rightIsSpace) {
-            return -1; // left < right
+            return 1; // left < right
```
froydnj, line 208:
```suggestion
            return 1; // left > right
```

parser/parser/cc/lexer.rl
```diff
@@ -205,11 +205,11 @@ int lexer::compare_indent_level(token_t left, token_t right) {
...
         } else if (!leftIsSpace && !rightIsSpace) {
             return 0; // left == right
         } else if (!leftIsSpace && rightIsSpace) {
-            return 1;  // left > right
+            return -1;  // left > right
```
froydnj, line 212:
```suggestion
            return -1;  // left < right
```

froydnj (approved):

PR #5352, opened 2022-02-24
stub out generic types in single-package RBI generation mode

froydnj:
### Motivation

In single-package RBI generation mode, we don't necessarily have access to full information about generic types in signatures and the like.  So we have to make guesses about what could be generic to get proper RBIs.  This is one way to do it.

Simply declaring every stubbed symbol as generic was a possibility, but one that seemed a little overly broad.

### Test plan

See included automated tests.  I haven't updated the exp files for the single-package RBI generation test we have, because the diffs come out looking not quite right (the generated RBIs aren't valid Ruby syntax) and I'm not totally sure what needs to be done there.

### DIFF
--- a/core/types/printing.cc
+++ b/core/types/printing.cc
@@ -50,7 +50,7 @@ string argTypeForUnresolvedAppliedType(const GlobalState &gs, const TypePtr &t, 
     }
     return t.show(gs, options);
 }
-}
+} // namespace
 
 string UnresolvedAppliedType::show(const GlobalState &gs, ShowOptions options) const {
     string resolvedString = options.showForRBI ? "" : " (unresolved)";
@@ -64,7 +64,11 @@ string UnresolvedAppliedType::show(const GlobalState &gs, ShowOptions options) c
     }
 
     return fmt::format("{}[{}]{}", symForPrinting.show(gs, options),
-                       fmt::map_join(targs, ", ", [&](auto targ) { return options.showForRBI ? argTypeForUnresolvedAppliedType(gs, targ, options) : targ.show(gs, options); }),
+                       fmt::map_join(targs, ", ",
+                                     [&](auto targ) {
+                                         return options.showForRBI ? argTypeForUnresolvedAppliedType(gs, targ, options)
+                                                                   : targ.show(gs, options);
+                                     }),
                        resolvedString);
 }
 
--- a/core/types/printing.cc
+++ b/core/types/printing.cc
@@ -45,8 +45,16 @@ string UnresolvedAppliedType::toStringWithTabs(const GlobalState &gs, int tabs) 
 
 string UnresolvedAppliedType::show(const GlobalState &gs, ShowOptions options) const {
     string resolvedString = options.showForRBI ? "" : " (unresolved)";
+    ClassOrModuleRef symForPrinting;
 
-    return fmt::format("{}[{}]{}", this->klass.show(gs, options),
+    if (options.showForRBI) {
+        auto attachedClass = this->klass.data(gs)->attachedClass(gs);
+        symForPrinting = attachedClass;
+    } else {
+        symForPrinting = this->klass;
+    }
+
+    return fmt::format("{}[{}]{}", symForPrinting.show(gs, options),
                        fmt::map_join(targs, ", ", [&](auto targ) { return targ.show(gs, options); }),
                        resolvedString);
 }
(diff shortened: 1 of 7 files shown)


core/types/printing.cc
```diff
@@ -43,9 +43,33 @@ string UnresolvedAppliedType::toStringWithTabs(const GlobalState &gs, int tabs)
...
-                       fmt::map_join(targs, ", ", [&](auto targ) { return targ.show(gs, options); }));
+    string resolvedString = options.showForRBI ? "" : " (unresolved)";
+    ClassOrModuleRef symForPrinting;
+
+    if (options.showForRBI) {
+        auto attachedClass = this->klass.data(gs)->attachedClass(gs);
+        symForPrinting = attachedClass;
+    } else {
+        symForPrinting = this->klass;
+    }
```
elliottt, line 64: I made similar changes to `UnresolvedClassType::show` when working on the initial single-package PR but rolled them back once stubbing was working correctly. Are you sure this change is needed? I would have assumed that types would resolve correctly after stubbing.

resolver/resolver.cc
```diff
@@ -370,8 +371,22 @@ class ResolveConstantsWalk {
...
+                return;
+            }
+            if (!possibleGenericType) {
+                return;
+            }
+            auto singletonClass = out->symbol.asClassOrModuleRef().data(ctx)->singletonClass(ctx);
+            auto &mixins = singletonClass.data(ctx)->mixins();
+            if (absl::c_find(mixins, core::Symbols::T_Generic()) == mixins.end()) {
+                mixins.emplace_back(core::Symbols::T_Generic());
+            }
```
elliottt, line 389: Is it worth pulling this out into a separate function that we can call here and on line ~435?

resolver/resolver.cc
```diff
@@ -564,6 +589,9 @@ class ResolveConstantsWalk {
 
     static bool resolveJob(core::Context ctx, ResolutionItem &job) {
         if (isAlreadyResolved(ctx, *job.out)) {
+            if (job.possibleGenericType) {
+                return false;
+            }
```
elliottt, line 594: This will have a pretty negative effect on the resolver in the normal case, as generics will never be resolved in the first pass through the tree. What case required unconditionally delaying jobs that look like generics?

elliottt (changes requested):
I think this makes sense, I just have a few questions about the changes to printing and `resolveJob`.

resolver/resolver.cc
froydnj: It's worth noting that this only affects single-package RBI generation mode, since that's the only time we set `possibleGenericType` to true.  (I guess we'll have an extra test-and-branch normally, but that's cheap and branch predictors are pretty good.)

The problem this fixes is that we can wind up in a situation where we have stubbed a constant before we knew that it was a possible generic type.  So when we get a `ResolutionItem` that we know *could* be a generic, we would find that we already stubbed it, and consider this particular job as resolved.  But we needed this particular job to persist until `constantResolutionFailed`, because this job carries information that marks the associated constant as potentially a generic type, and we need to process this job appropriately to make the stub look like a generic type.  Ergo, we couldn't let this job resolve "properly".

I guess this could have possibly bad effects on constants that we *didn't* stub, but look like generics, and therefore we might never consider them resolved.  That does seem bad, so perhaps this check should be a little smarter?

core/types/printing.cc
froydnj: I am "pretty" sure; without these changes, you'd get `UnresolvedAppliedType`s that print like `T.class_of(GenericMessage)[<Type: String>]` in the RBI, which doesn't seem like acceptable Ruby syntax.

It's possible that we should just change how `UnresolvedAppliedType` is constructed so we would have to do less of this.

resolver/resolver.cc
elliottt: Ah, I had missed that we only set `possibleGenericType` when single-package mode is enabled. I'm a lot less worried about performance impact if that's the case.

core/types/printing.cc
elliottt: Maybe we could do some follow-up investigation to see about removing this special case; it seems like the sort of thing we should be able to avoid if stubbed constants look appropriately generic.

elliottt (approved):

froydnj:
I added another test just to make sure that a generic with a package-local definition works like we expect it to.  (I suspect that since this PR makes `AppliedType` and `UnresolvedAppliedType` work roughly identically, printing-wise, in single-package mode, it wasn't a concern that such generics would stop working, but I sleep a little better knowing that there's a test for this.)

elliottt (approved):
Thanks for that additional test, and extracting that method!

PR #3226, opened 2020-06-23
Import tests for ruby27 parser and add a few features

Morriar:
This pull-request prepares the work for the 2.7 update and also add a few independent features.

1. Update `import_whitequark` script to 2.7
2. Update parser version to 2.7
3. Refactor grammar rules around symbols (no changes to the output)
4. Treat `dsym` as a string (no changes to the output)

This PR may be read commit by commit for an easier review.

I'll submit subsequent PRs with each feature and the related tests imported with the script.

### Test plan

See included automated tests.

PR #5353, opened 2022-02-24
Remove version numbers from typedruby_release and typedruby_debug

KaanOzkan:
Removes version numbers from `typedruby_release` and `typedruby_debug` classes.

### Motivation

Including the version numbers create extra overhead while making lexer/parser changes. Such as https://github.com/sorbet/sorbet/pull/3226/files#diff-534a21ccf64c1a7b69ffc494938680a4223ef1dab50618d2a675c46095451dceR55 and similarly for Ruby 3.1 in the upcoming PR I was working on.

I don't think they bring enough value to justify the maintenance overhead and they can be removed. Thoughts?

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/cc/lexer.rl
+++ b/parser/parser/cc/lexer.rl
@@ -2272,7 +2272,7 @@ void lexer::set_state_expr_value() {
           | '@@' %{ tm = p - 2; }
           ) [0-9]*
       => {
-        if (version == ruby_version::RUBY_27) {
+        if (version >= ruby_version::RUBY_27) {
           if (ts[0] == ':' && ts[1] == '@' && ts[2] == '@') {
             diagnostic_(dlevel::ERROR, dclass::CvarName, tok(ts + 1, te));
           } else {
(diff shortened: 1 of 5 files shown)


Morriar:
It also seems like we forgot to update those when pushing the 3.0 changes 🤦

parser/parser/cc/lexer.rl
```diff
@@ -2272,7 +2272,7 @@ void lexer::set_state_expr_value() {
           | '@@' %{ tm = p - 2; }
           ) [0-9]*
       => {
-        if (version == ruby_version::RUBY_27) {
+        if (version >= ruby_version::RUBY_27) {
```
KaanOzkan, line 2275: This was a small mismatch with the `whitequark/parser` that results in a bug when upgrading the lexer ruby version to 3.1.
https://github.com/whitequark/parser/blob/ec523c0f8e9eaedcfb892b8e64394efbce9777d8/lib/parser/lexer.rl#L1899

Included with this PR because I wanted to keep my other feature PR simpler but I can remove it since it's not that relevant.

aprocter-stripe:

→ https://go/builds/bui_LDEYJfhA3RWIkF
→ https://go/builds/bui_LDEYnkUd9UV2d4

aprocter (approved):
Seems reasonable!

PR #5355, opened 2022-02-24
stabilize RBI dependency file output

froydnj:
### Motivation

Deterministic output is a good thing.

### Test plan

I need this for tests in #5352, so real tests for deterministic output will be added there.

### DIFF
--- a/packager/rbi_gen.cc
+++ b/packager/rbi_gen.cc
@@ -7,6 +7,7 @@
 #include "common/FileOps.h"
 #include "common/concurrency/ConcurrentQueue.h"
 #include "common/concurrency/WorkerPool.h"
+#include "common/sort.h"
 #include "core/GlobalState.h"
 #include "packager/packager.h"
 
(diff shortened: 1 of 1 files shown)


packager/rbi_gen.cc
```diff
@@ -1167,6 +1164,23 @@ class RBIExporter final {
         return lookupFQN(gs, fullName).asClassOrModuleRef();
     }
 
+    string buildPackageDependenciesString(const core::GlobalState &gs) {
+        vector<string> packageNames;
+        packageNames.reserve(referencedPackages.size());
+        absl::c_transform(referencedPackages, back_inserter(packageNames),
+                          [&gs](const auto &p) -> string { return p.first.show(gs); });
+        fast_sort(packageNames, [](const auto &lhs, const auto &rhs) -> bool { return lhs < rhs; });
```
jez, line 1172: Are FileRef IDs stable?

jez: Do we need to sort by the file path instead?

jez: i guess it looks like this is not the first place we're sorting on these IDs, and the `operator<` definition even exists so maybe it's not worth worrying about

jez (approved):
sorting seems fine. i usually try to ask myself in situations like this if there's a way to make it sorted by construction, instead of calling `fast_sort` after the fact. food for thought.

packager/rbi_gen.cc
froydnj: Given that we sort by `FileRef` for determinism purposes in e.g. resolver:

https://github.com/sorbet/sorbet/blob/10d09b92d34d832bef454f0436cf0b35a7003cb2/resolver/resolver.cc#L1524-L1539

(and a number of other places), I think this is OK.  It does mean that the output is not stable if you change the order files are given on the command line, but I think that is mostly OK?

PR #5357, opened 2022-02-24
Simplify test case rewriter

vinistock:
This is the first step in trying to get rid of the `Minitest` and `ActiveSupport::TestCase` rewriters (starting with `TestCase`).

1) Add signatures for the `test`, `setup` and `teardown` methods using blocks. One thing I should note about this is that the real structure inside Rails is slightly different. These methods are actually defined in modules and then extended. 

However, we cannot use `T.attached_class` in a non-singleton context, so the proper representation of these modules won't work. E.g.:
```ruby
module Declarative
  # Fails because T.attached_class is used outside of a singleton context
  sig { params(args: T.untyped, block: T.nilable(T.proc.bind(T.attached_class).void)).void }
  def setup(*args, &block); end
end

class TestCase
  extend Declarative
end
```
Because of this, instead of adding the signatures in modules, they were added straight into the `TestCase` class. I think this is fine, but let me know if there are concerns related to this approach.

2) Get rid of the re-writing the invocations to `test` and `teardown`. The only rewriting we kept is re-writing `setup` method invocations into an `initialize` definition, so that we don't get the error of nilable instance variables in `strict` files.
```ruby
# typed: strict
class MyTest < ActiveSupport::TestCase
  # We rewrite this to avoid emitting an error that the variable `a` must be nilable
  setup do
    @a = T.let(1, Integer)
  end
end
```

### Motivation

Now that we can express the block bindings in respect to `T.attached_class` and `T.self_type`, we can use them to minimize the amount of rewriting we are doing.

The next step is to add signatures for Minitest and try to simplify that rewriter as well. Hopefully, we can combine both rewriters into a single one and only rewrite the minimum necessary.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/rewriter/test_case.rb
+++ b/test/testdata/rewriter/test_case.rb
@@ -1,20 +1,22 @@
 # typed: strict
 
 class ActiveSupport::TestCase
-end
-
-class MyTest < ActiveSupport::TestCase
   extend T::Sig
+
+  # Helper method to direct calls to `test` instead of Kernel#test
+  sig { params(args: T.untyped, block: T.nilable(T.proc.bind(T.attached_class).void)).void }
+  def self.test(*args, &block)
+  end
+
   # Helper instance method
   sig { params(test: T.untyped).returns(T::Boolean) }
   def assert(test)
     test ? true : false
   end
+end
 
-  # Helper method to direct calls to `test` instead of Kernel#test
-  sig { params(args: T.untyped, block: T.nilable(T.proc.void)).void }
-  def self.test(*args, &block)
-  end
+class MyTest < ActiveSupport::TestCase
+  extend T::Sig
 
   setup do
     @a = T.let(1, Integer)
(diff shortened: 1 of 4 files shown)


rbi/gems/minitest.rbi
```diff
@@ -0,0 +1,5 @@
+# typed: __STDLIB_INTERNAL
+
+class Minitest::Runnable; end
+
+class Minitest::Test < ::Minitest::Runnable; end
```
jez, line 5: should this really be in Sorbet? If so it should probably be the full minitest RBI

otherwise, if there are individual test cases that need this RBI to be present, this should be included in the relevant test case.

rbi/gems/test_case.rbi
```diff
@@ -0,0 +1,12 @@
+# typed: __STDLIB_INTERNAL
+
+class ActiveSupport::TestCase < ::Minitest::Test
```
jez, line 3: `ActiveSupport` _definitely_ does not belong in Sorbet. we can duplicate this into any tests that need it.

test/testdata/rewriter/test_case.rb
```diff
@@ -25,23 +17,6 @@ def self.test(*args, &block)
...
-  end
-
-  test "no block argument"
-
-  test :not_a_string do
-  end
-
-  test :not_a_string do
-    assert true # error: Method `assert` does not exist on `T.class_of(MyTest)`
-  end
```
jez, line 42: ooc, why are we deleting tests? shouldn't the test pass before and after the changes?

rbi/gems/test_case.rbi
```diff
@@ -0,0 +1,12 @@
...
+  sig { params(args: T.untyped, block: T.nilable(T.proc.bind(T.attached_class).void)).void }
+  def self.setup(*args, &block); end
```
jez, line 5: Your PR description called out qualms with the fact that these methods are not actually defined in real life, that they're defined in some module.

That's another point in favor of having these RBIs only in the test cases and projects that want this behavior, not in Sorbet.

froydnj:

→ https://go/builds/bui_LDG692OxR0F6aO
→ https://go/builds/bui_LDG6Ao1CnuxWkr

froydnj (changes requested):
jez's comments about extra RBIs are on point; IIUC, Stripe's codebase has a forked version of minitest, which, in combination with `--stripe-packages`, results in this PR failing on Stripe's codebase.  So at the very least, I don't think we want `Minitest` or `ActiveSupport` in the payload.

I haven't worked through the rewriter changes in detail.

It's not immediately obvious to me why the compiler testcases decided to output slightly different LLVM IR, but that should be straightforward to fixup.

There are also some regular `.exp` files that need fixing.

rbi/gems/minitest.rbi
vinistock: Fair point. I was concerned that type checking that was previously passing would suddenly break without the signatures, but I agree this belongs in the apps.

test/testdata/rewriter/test_case.rb
vinistock: Because the rewriter now only rewriters `setup {}` into `def initialize`, some of these tests aren't actually verifying any existing behaviour. I simplified the tests to only verify the scenarios we rewrite:
- If there are invocations to `test "something" {}`, we rewrite the `setup` block into `initialize`
- Otherwise, we do nothing

froydnj (approved):
This seems fine to me and a lot less invasive.  The PR passes Stripe's codebase, which doesn't mean a whole lot, as the codebase doesn't trigger this rewriter (I think), so I assume you've tested this PR on a roughly equivalent Rails-using codebase.

I attempted getting rid of the Minitest rewriter pass using Stripe's codebase as a testbed; what I found was that we need some kind of rewriting to express `before` blocks as a sort of `initialize`...and then I think we also need some kind of rewriting so `before` (`initialize`) gets attached to the correct `describe`.  (People can and do write `describe` inside `describe`, with separate `before` blocks for each and things like that.)  I suppose it's possible that teaching people to write more "disciplined" tests would eliminate the need for `describe`/`it` rewriting, but I am not excited about introducing a codemod like that (or writing sufficient error messages to help out confused people).

It's definitely possible, however, that I was just insufficiently clever or thoughtful; I'll be interested to see what you come up with on the Minitest rewriter front!

vinistock:
Yeah, the spec style of writing tests may be a little bit trickier still. We'll give it a try and see if adding signatures with the correct bindings can help at least simplify the rewriter.

PR #5359, opened 2022-02-25
Add a version of PackageDB::getPackageInfo that works with strings

elliottt:
For single-package rbi generation, it's really useful to be able to find packages by their string name. This PR adds that functionality to `PackageDB` as prework for reworking namespace management in single-package mode.

### Motivation

Making it easier to find packages in single-package mode.

### Test plan

See included automated tests.

### DIFF
--- a/core/packages/PackageDB.cc
+++ b/core/packages/PackageDB.cc
@@ -182,13 +182,7 @@ const PackageInfo &PackageDB::getPackageInfo(const core::GlobalState &gs, std::s
     if (!cnst.exists()) {
         return NONE_PKG;
     }
-
-    auto it = this->packages_.find(cnst);
-    if (it == this->packages_.end()) {
-        return NONE_PKG;
-    }
-
-    return *it->second;
+    return getPackageInfo(cnst);
 }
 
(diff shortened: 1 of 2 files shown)


core/packages/PackageDB.cc
```diff
@@ -165,6 +166,31 @@ const PackageInfo &PackageDB::getPackageForFile(const core::GlobalState &gs, cor
...
+    if (!cnst.exists()) {
+        return NONE_PKG;
+    }
+
+    auto it = this->packages_.find(cnst);
+    if (it == this->packages_.end()) {
+        return NONE_PKG;
+    }
+
+    return *it->second;
```
nroman-stripe, line 191:
```suggestion
    return getPackageInfo(cnst);
```
Minor, but I believe this could be simplified.

nroman-stripe (approved):
lgtm

core/packages/PackageDB.cc
elliottt: Oh hey, great point!

aprocter (approved):

PR #5364, opened 2022-02-25
Ruby 3.1: Parse forward arg without parentheses

KaanOzkan:
Part of Ruby 3.1 support

https://github.com/whitequark/parser/commit/bb6c4a8e22917fafa1ba9e1b99a3b6598791a694
https://bugs.ruby-lang.org/issues/18267

### Motivation

### Test plan

See included automated tests.

cc @Morriar

### DIFF
--- a/parser/parser/cc/driver.cc
+++ b/parser/parser/cc/driver.cc
@@ -63,7 +63,7 @@ void base_driver::rewind_if_dedented(token_t token, token_t kEND, bool force) {
 
 typedruby_release::typedruby_release(std::string_view source, sorbet::StableStringStorage<> &scratch,
                                      const struct builder &builder, bool traceLexer, bool indentationAware)
-    : base_driver(ruby_version::RUBY_27, source, scratch, builder, traceLexer, indentationAware) {}
+    : base_driver(ruby_version::RUBY_31, source, scratch, builder, traceLexer, indentationAware) {}
 
 ForeignPtr typedruby_release::parse(SelfPtr self, bool) {
     bison::typedruby_release::parser p(*this, self);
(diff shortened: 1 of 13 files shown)


parser/parser/cc/driver.cc
```diff
@@ -63,7 +63,7 @@ void base_driver::rewind_if_dedented(token_t token, token_t kEND, bool force) {
 
 typedruby_release::typedruby_release(std::string_view source, sorbet::StableStringStorage<> &scratch,
                                      const struct builder &builder, bool traceLexer, bool indentationAware)
-    : base_driver(ruby_version::RUBY_27, source, scratch, builder, traceLexer, indentationAware) {}
+    : base_driver(ruby_version::RUBY_31, source, scratch, builder, traceLexer, indentationAware) {}
```
elliottt, line 66: Does changing the default version to `RUBY_31` change behavior in a way that would cause problems for a codebase that's still on 2.7?

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -3701,6 +3688,21 @@ f_opt_paren_args: f_paren_args
                     {
                       $$ = $1;
                     }
+                | args_forward
+                    {
+                      if (driver.lex.context.inLambda()) {
+                        // This is supposed to be caught by the grammar.
+                        // However, we don't generate a `ForwardArgs` node in the builder.cc unlike whitequark parser.
+                        // So, we have to manually raise a parse error to forbid `->(...)` usage
+                        // Last argument adds extra double quotes to circumvent an internal check
```
elliottt, line 3697: This comment is really helpful, thanks!

elliottt (approved):
This looks reasonable to me, but I'd like @jez to take a look as well.

parser/parser/cc/driver.cc
KaanOzkan: Not currently since the changes made in lexer are backwards compatible  and parser tests pass with `RUBY_27`.

There's still a risk of issues being introduced in the future though since CI will not test with `RUBY_27`. I don't think this risk can be eliminated without supplying the actual ruby version number to the lexer instead of a fixed version.

jez:
> but I'd like @jez to take a look as well.

Heads up I'm unlikely to have time to review this until next week sometime, but I'll be sure to make a note to look then.

jez (approved):
lgtm, thanks!

PR #5365, opened 2022-02-25
Fix signatures for open3 to include env hash argument

rwstauner:
Allow for the first argument to be the env hash.

### Motivation
Trying to set the env for the open3 methods `Open3.capture2e({...}, "...")` is incompatible with the current signatures. 

### Test plan
Both of these are correct uses:
```
Open3.capture2e("echo")
Open3.capture2e({"LANG" => "en_US.UTF-8"}, "echo")
```

### DIFF
--- a/rbi/stdlib/open3.rbi
+++ b/rbi/stdlib/open3.rbi
@@ -156,7 +156,7 @@ module Open3
   # ```
   sig do
     params(
-      cmd: T.any(String, T::Array[String]),
+      cmd: T.any(T::Hash[String, T.nilable(String)], String, T::Array[String]),
       opts: T::Hash[Symbol, T.untyped],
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_LDYlRjxEHeF3w5
→ https://go/builds/bui_LDYlXwkHTxE1NY

froydnj (approved):
LGTM!  I think technically the hashes can have strings or symbols as their keys (cf. `Process.spawn`), but as this is a strict improvement, I'm not worried about this.

rwstauner:
thanks!

rwstauner:
> LGTM! I think technically the hashes can have strings or symbols as their keys (cf. `Process.spawn`), but as this is a strict improvement, I'm not worried about this.

The docs say
>   #   the keys and the values except for +nil+ must be strings.

and irb agrees.

https://github.com/sorbet/sorbet/pull/5366

Issue #5155, opened 2022-01-25
Support Ruby 3.1 anonymous block argument


bmulholland:
#### Input

Ruby 3.1 added anonymous block arguments: https://rubyreferences.github.io/rubychanges/3.1.html#anonymous-block-argument

[→ View on sorbet.run](https://sorbet.run/#def%20logged_open%28filename%2C%20%26%29%0A%20%20puts%20%22Opening%20%23%7Bfilename%7D...%22%0A%20%20File.open%28filename%2C%20%26%29%0Aend)

```ruby
def logged_open(filename, &)
  puts "Opening #{filename}..."
  File.open(filename, &)
end
```

#### Observed output

```
editor.rb:1: unexpected token ")" https://srb.help/2001
     1 |def logged_open(filename, &)
                                   ^

editor.rb:3: unexpected token ")" https://srb.help/2001
     3 |  File.open(filename, &)
```

PR #5368, opened 2022-02-25
Ruby 3.1: Parse anonymous block argument

KaanOzkan:
Resolves https://github.com/sorbet/sorbet/issues/5155

https://bugs.ruby-lang.org/issues/11256
https://github.com/whitequark/parser/pull/833

### Motivation

### Test plan

See included automated tests.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -562,6 +562,10 @@ public:
     }
 
     unique_ptr<Node> blockPass(const token *amper, unique_ptr<Node> arg) {
+        if (arg == nullptr) {
+            return make_unique<BlockPass>(tokLoc(amper), nullptr);
+        }
+
         return make_unique<BlockPass>(tokLoc(amper).join(arg->loc), std::move(arg));
     }
 
--- a/test/testdata/desugar/anonymous_blockarg.rb
+++ b/test/testdata/desugar/anonymous_blockarg.rb
@@ -1,0 +1,11 @@
+# typed: true
+
+def foo(args, &)
+  bar(*args, &)
+end
+
+def bar(args, &)
+  baz(&)
+end
+
+foo (1) { 2 }
--- a/test/whitequark/test_anonymous_blockarg_0.parse-tree-whitequark.exp
+++ b/test/whitequark/test_anonymous_blockarg_0.parse-tree-whitequark.exp
@@ -1,0 +1,5 @@
+s(:def, :foo,
+  s(:args,
+    s(:blockarg, nil)),
+  s(:send, nil, :bar,
+    s(:block_pass, nil)))
--- a/test/whitequark/test_anonymous_blockarg_0.parse-tree-whitequark.exp
+++ b/test/whitequark/test_anonymous_blockarg_0.parse-tree-whitequark.exp
@@ -1,5 +1,5 @@
 s(:def, :foo,
   s(:args,
-    s(:blockarg, nil)),
+    s(:blockarg, :&)),
   s(:send, nil, :bar,
     s(:block_pass, nil)))
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -724,11 +724,18 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                     // tuple type and call into the normal call merchanism.
                     unique_ptr<parser::Node> block;
                     auto argnodes = std::move(send->args);
+                    bool anonymousBlockPass = false;
+                    core::LocOffsets bpLoc;
                     auto it = absl::c_find_if(argnodes,
                                               [](auto &arg) { return parser::isa_node<parser::BlockPass>(arg.get()); });
                     if (it != argnodes.end()) {
                         auto *bp = parser::cast_node<parser::BlockPass>(it->get());
-                        block = std::move(bp->block);
+                        if (bp->block == nullptr) {
+                            anonymousBlockPass = true;
+                            bpLoc = bp->loc;
+                        } else {
+                            block = std::move(bp->block);
+                        }
                         argnodes.erase(it);
                     }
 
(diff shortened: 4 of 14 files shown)


test/whitequark/test_anonymous_blockarg_0.parse-tree-whitequark.exp
```diff
@@ -0,0 +1,5 @@
+s(:def, :foo,
+  s(:args,
+    s(:blockarg, :&)),
```
KaanOzkan, line 3: We differ from whitequark/parser implementation by specifying the argument as `&` instead of `nil`.

parser/Builder.cc
```diff
@@ -536,6 +536,10 @@ class Builder::Impl {
     }
 
     unique_ptr<Node> blockPass(const token *amper, unique_ptr<Node> arg) {
+        if (arg == nullptr) {
+            return make_unique<BlockPass>(tokLoc(amper), nullptr);
```
KaanOzkan, line 540: Parser tests are passing but I want to make sure creating a `BlockPass` with a `nil` block is a good solution here.

KaanOzkan (commented):
Part of the import from whitequark/parser is missing in the diff because it was previously implemented in https://github.com/sorbet/sorbet/commit/70da972e0c6fd2c50a6aad9975fad3812e64cf15 and the code was never removed even though afaict it wasn't used.

That commit creates a unique name with `&` to overcome building a `Blockarg` without an available `NameRef`.

parser/Builder.cc
elliottt: I think we'll need some changes in desugar.cc to handle this. Could you add some tests for combinations of splats and the anonymous block argument? My worry is that we'll end up interpreting the `nullptr` block argument as no block being present, and will choose the wrong splat intrinsic during desugar:

https://github.com/sorbet/sorbet/blob/c3365c5fbc4f652b30cc45428792f4646783c3a8/ast/desugar/Desugar.cc#L784-L786

elliottt (changes requested):
Could you add some more tests that show that uses of the anonymous block arg are desugaring correctly?

parser/Builder.cc
KaanOzkan: I added a test [here](https://github.com/sorbet/sorbet/pull/5368/commits/64d6ae1bb0a15bddbdc99b2142aa96244db82aae). With or without splats the desugar tree doesn't contain a block while passing it to another method.

Will this cause errors though? Users already can't annotate the type using `params` or `bind` since there isn't an identifier for the block. To me this seems very similar to how every ruby method can accept implicit blocks as far as Sorbet is concerned and I can't see how block passing affects it.

elliottt: To expose the problem I'm worried about, you need to combine a splat and an anonymous block, like the following:

```ruby
def bar(args, &)
  foo(*args, &)
end
```

Here's the result of running it on your branch:

```
trevor ~/stripe/sorbet (ko/31-parse-anon-block !$)
% ./bazel-bin/main/sorbet -e "def bar(args, &); foo(*args, &); end" --print desugar-tree --silence-dev-message
  def bar<<todo method>>(args, &&$2)
    ::<Magic>.<call-with-splat>(<self>, :foo, ::<Magic>.<splat>(args), nil)
  end
end
No errors! Great job.
```

It's selected the `::<Magic>.<call-with-splat>` intrinsic, when it should have chosen `::<Magic>.<call-with-splat-and-block,>`.

KaanOzkan: Thank you for the repro 🙂 . I've updated the tests and went with a simple flag to differentiate a nil block due to not finding a `blockPass` and due to an anonymous block argument.

https://github.com/sorbet/sorbet/pull/5368/commits/6beae86084c4436d0c0f5cc1e37bf8fb72b503fa.

What do you think?

elliottt: What about this case?
```ruby
def bar(args, &)
  foo(&)
end
```
It produces the following desugared tree:
```
% ./bazel-bin/main/sorbet -e "def bar(args, &); foo(&); end" --print desugar-tree --silence-dev-message
  def bar<<todo method>>(args, &&$2)
    <self>.foo()
  end
end
```

I would have expected this to produce something like: `:<Magic>.<call-with-block>(<self>, :foo, nil)`, as we're still passing a block argument.

Longer-term, what's the plan for supporting this feature when type-checking? The current translation will make it appear like a block arg has been passed, but that it's always `nil`. For example, `foo(*[], &nil)` will produce the same thing as `foo(*[], &)` on your branch.

```
% ./bazel-bin/main/sorbet -e "foo(*[], &nil)" --print desugar-tree --silence-dev-message
  ::<Magic>.<call-with-splat-and-block>(<self>, :foo, ::<Magic>.<splat>([]), nil, nil)
end
```

KaanOzkan: > I would have expected this to produce something like: :<Magic>.<call-with-block>(<self>, :foo, nil), as we're still passing a block argument.

Looks like there is another `BlockPass` logic few lines down. I think I can reuse the flag solution here as well.

> Longer-term, what's the plan for supporting this feature when type-checking?

I don't think we need to support typechecking with this syntax. Block of a `BlockPass` is `nil` in the whitequark implementation as well. I think we can allow this syntax with lexer/parser changes and nudge users to use the regular syntax to get typechecking benefits. As I mentioned earlier, I'm looking at this feature as similar to how all Ruby methods accept an implicit block. Only difference here is that we're telling Ruby the block we are passing is the same one as the method parameter. So I believe the information we are losing with this implementation is minor.

I'd like to keep iy consistent with whitequark parser and I wouldn't want to complicate desugaring to support typechecking in this minor use case. I believe given that you already can't access this argument in a sig or annotate type with `T.proc` in the method body greatly reduces the desire to have typechecking using this syntax.

elliottt: Sorry about that @KaanOzkan!

I'm a little concerned about leaving the `block` variable as `nullptr` and relying on a flag instead. In the else branch that will be taken when your flag is true, the `block` is passed directly in to `node2TreeImpl`: https://github.com/sorbet/sorbet/blob/e8cf71fd6ab3d156b8b5ed5f1731e6e8ef6788a4/ast/desugar/Desugar.cc#L787

What about introducing a separate function for processing a `BlockPass` node that knows how to handle this case and turn it into a local with a well-known name? My worry is that we'll lose precision during typechecking when we could handle the anonymous block arg in the same way as a named arg with a special name.

Consider this case:

```ruby
sig {params(x: T::Array[Integer], blk: T.proc.params(x: String).void).void}
def foo(x, &blk)
  x.each(&blk)
end
```

Right now this is a type error as the proc's argument type doesn't match the array's element type. However, if we start supporting block forwarding without connecting back to the signature, it's unclear to me how we'll get a similar error for the case where we rewrite it as:

```ruby
sig {params(x: T::Array[Integer], blk: T.proc.params(x: String).void).void}
def foo(x, &)
  x.each(&)
end
```

I think that the most straightforward translation here is to ensure that we introduce a fresh name for the block arg when it's defined with the forwarding syntax, so that there's no need for the typechecker to know about block forwarding.

elliottt: From a DM:

@KaanOzkan pointed out that we have no way to add a type for the block parameter in the sig when it's defined with an anonymous block, which makes my second example incorrect. Thanks for clearing that up, Kaan!

As a result, I think the only case that we need to address is when passing the anonymous block arg we need to ensure that it's not represented as `nil`, as that's a valid block arg value. Passing `nil` might raise incorrect errors if the block arg of the method called doesn't allow a nilable block.

elliottt (changes requested):
Just one last request and this PR will be good to go!

elliottt:

→ https://go/builds/bui_LV9R0I4BauNhh0
→ https://go/builds/bui_LV9R44pRiImE5A

test/testdata/desugar/anonymous_blockarg.rb
```diff
@@ -0,0 +1,13 @@
...
+
+def foo(args, &)
+  bar(*args, &)
+end
+
+def bar(args, &)
+  baz(&)
+end
+
+def baz(&blk); end
```
elliottt, line 11: Is this case necessary?

elliottt (approved):
One last question about test cases, but otherwise this looks good!

PR #5370, opened 2022-02-26
Fix error location for `&.` error, add autocorrect

jez:
### Motivation

@jvilk-stripe reported this bug while working on something at Stripe.

### Test plan

See included automated tests.


elliottt (approved):
Makes sense!

PR #5371, opened 2022-02-26
Track more information for single-package rbi generation

elliottt:
### Motivation

We currently consider all parent packages for stubbing when generating rbis in single-package mode, but we need to only consider those that are explicitly imported. This PR makes that change, and also tracks all imported package names, which will be needed for stubbing constants that come from imported child packages.

### Test plan

See included automated tests.

### DIFF
--- a/core/packages/PackageInfo.h
+++ b/core/packages/PackageInfo.h
@@ -64,5 +64,25 @@ public:
 
     static bool lexCmp(const std::vector<core::NameRef> &lhs, const std::vector<core::NameRef> &rhs);
 };
+
+// Information about the imports of a package. The imports are split into two categories, packages whose name falls
+// within the namespace of `package`, and everything else. The reason for pre-processing the imports this way is that it
+// simplifies some work when stubbing constants for rbi generation.
+class ImportInfo final {
+public:
+    // The mangled name of the package whose imports are described.
+    core::NameRef package;
+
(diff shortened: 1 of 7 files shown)


core/packages/PackageInfo.h
```diff
@@ -64,5 +64,25 @@ class PackageInfo {
...
+// within the namespace of `package`, and everything else. The reason for pre-processing the imports this way is that it
+// simplifies some work when stubbing constants for rbi generation.
+class ImportInfo final {
+public:
+    // The mangled name of the package whose imports are described.
+    core::NameRef package;
+
+    // Imported packages whose name is a prefix of `package`. For example, if the package `Foo::Bar` imports `Foo` that
+    // package's name would be in `parentImports` because its name is a prefix of `Foo::Bar`.
+    std::vector<core::NameRef> parentImports;
```
froydnj, line 78: The current code records packages that are parents of the particular package we're generating RBIs for; this only records imported parent packages...which seems to me to be a potentially smaller set.  Why do we not record general parents anymore?

(I guess maybe I am not as familiar with packages as I should be, but my naive assumption is that parent packages are automatically in scope?)

froydnj (approved):
I really like the more explicit structure here!

Just one question for my own edification.

core/packages/PackageInfo.h
elliottt: Parent packages are only in scope if they're imported, so the current behavior is incorrect.

Issue #3370, opened 2020-08-19
Autocorrect suggests `T.must` on lhs of assignment


Morriar:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Ax%20%3D%20T.let(1%2C%20T.nilable(Integer))%0Ax%20%2B%3D%201)

```ruby
# typed: true

x = T.let(1, T.nilable(Integer))
x += 1
```

#### Observed output

```
editor.rb:4: Method + does not exist on NilClass component of T.nilable(Integer) https://srb.help/7003
     4 |x += 1
        ^^^^^^
  Autocorrect: Use `-a` to autocorrect
    editor.rb:4: Replace with T.must(x)
     4 |x += 1
        ^
Errors: 1
```

Which gives the following autocorrected code:

```ruby
T.must(x) += 1
```

#### Expected behavior

The corrected code should be more like:

```ruby
x = T.must(x) + 1
```

PR #5373, opened 2022-02-26
Expand `OpAsgn` when suggesting `T.must`

jez:
### Motivation

Fixes <https://github.com/sorbet/sorbet/issues/3370>

This will also help unblock certain portions of the (eventual) migration to fix
shape types, because this is a common class of error that appears when shape
type are improved to actually check accessors (instead of being completely
untyped).

### Test plan

Tests show before/after


elliottt (approved):
Is it worth adding a test for something like the following?
```
x, y = [1, T.let(nil, T.nilable(String))]
y.split
```

jez:
Hm, we definitely could, but that is completely unrelated to these changes, as it doesn't have anything to do with OpAsgn, right?

jez:
(I double checked, your test case passes before the changes made in this PR.)

elliottt:
> Hm, we definitely could, but that is completely unrelated to these changes, as it doesn't have anything to do with OpAsgn, right?

Good point, I was confusing this with assignment in general and remembered having trouble with `T.must` autocorrects and the destructured assignment in the past.

Issue #3443, opened 2020-09-21
Bad autocorrect suggested for safe navigation expression


aisamanra:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aextend%20T%3A%3ASig%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Breturns(T.nilable(String))%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Asig%20%7Bparams(x%3A%20String).void%7D%0Adef%20use_string(x)%0Aend%0A%0Amaybe_a%20%3D%20T.let(A.new%2C%20T.nilable(A))%0Ause_string(maybe_a%26.foo)%0A)

```ruby
# typed: true

extend T::Sig

class A
  extend T::Sig
  sig {returns(T.nilable(String))}
  def foo; end
end

sig {params(x: String).void}
def use_string(x)
end

maybe_a = T.let(A.new, T.nilable(A))
use_string(maybe_a&.foo)

```

#### Observed output

Just the relevant part of the output:

```
  Autocorrect: Use `-a` to autocorrect
    editor.rb:16: Insert T.must()
    16 |use_string(maybe_a&.foo)
                   ^
```

Applying this results in the expression

```ruby
use_string(T.must()maybe_a&.foo)
```

#### Expected behavior

The ideal autocorrect here is

```ruby
use_string(T.must(maybe_a&.foo))
```

- - -

I suspect (but haven't confirmed) that this is because of the synthesized location of the desugared safe navigation operator, but if my hunch is right this should be comparatively easier to fix, since the location of the whole `if` should be the same span as the original safe send.

paracycle:
Potentially related to: https://github.com/sorbet/sorbet/issues/3370

PR #5374, opened 2022-02-26
Fix autocorrect for CSend nodes

jez:
Note that this change depends on the changes in two previous PRs:

- <https://github.com/sorbet/sorbet/pull/5370>
  - A change related to `&.`, but which is mostly orthogonal except that it
    happens to conflict if I try to send it on its own.
- <https://github.com/sorbet/sorbet/pull/5372>
  - A (mostly no-op) change that changes where we report the error for methods
    taking positional arguments.

**Only the changes in the last two commits** are new.

### Motivation

Fixes https://github.com/sorbet/sorbet/issues/3443

### Test plan

Tests show behavior before and after this change.


froydnj (approved):
Thank you for the explanatory comments in the PR!

PR #2105, opened 2019-11-06
add LSP capability textDocument/documentHighlight

lgo:
cc @jvilk-stripe. Changes from #1623 and after rebasing. The text below is largely from the previous PR.

This adds the LSP functionality for `textDocument/documentHighlight` ([Microsoft specification for documentHighlight](https://microsoft.github.io/language-server-protocol/specification#textDocument_documentHighlight)).

When a user selects a piece of text, the LSP responds with a list of ranges to highlight in the current file which relate to the initiate selection. 

Currently, highlighting uses the same information `textDocument/references` does. That means  this will work for typed files otherwise it can only handle Ruby constants in untyped files. Compared to regular fuzzy-highlighting, this will only respond with highlights for an identifier reference within the scope it is valid in which is the added benefit of this LSP method!

This does not implement the response field `kind?: DocumentHighlightKind`, which identifies the response values as the more specific highlights:
- `Text`, a textual occurrence -- typically a fuzzy response.
- `Read`, a read-access of a symbol, like reading a variable.
- `Write`, a write-access of a symbol, like writing to a variable.

By default, the editor will consider no `kind` as a `Text` response.

After poking in VSCode I noticed there are [distinguishing styling for the types](https://github.com/microsoft/vscode/blob/466e5ef6c2188d61a8c93c53748aa9e9788bcfc9/src/vs/editor/contrib/wordHighlighter/wordHighlighter.ts#L427-L452). In particular, the styling settings, [found here](https://code.visualstudio.com/api/references/theme-color#editor-colors), are:
- `editor.wordHighlight{Background,Border}` for read-access
- `editor.wordHighlightStrong{Background,Border}` for write-access
- `editor.selectionHighlight{Background,Border}` for fuzzy text.

Currently, this LSP method has an issue where the range of the highlights is rather generous. The quirks are
- highlighting entire method definition lines (`def` + args)
- highlighting entire method call line (receiver, inputs, input block), eg: `d.method(1, 2) {}`. the same happens for `send` assignments assignments, eg: `d.method = 'bar'`)
- highlighting the `:` when highlighting parameter identifiers, eg: `foo(b:)`

I did some digging into the Locations used around here. As I understand it, we use the declLoc from these that come from the parsing step. Specifically, the problem cases from before are that we just want location of the SymbolRef for these -- and it primarily happened on Declaration nodes and Arg-type Reference nodes. I imagine we'd need to carry through a symbolLoc or know how to lazily compute it.

[...]

PR #5375, opened 2022-02-27
Only highlight `funLoc` for document highlight

jez:
### Motivation

The current document highlight highlights way too much of the screen when you
put your cursor anywhere inside a method.

### Test plan

There the fact that this didn't cause any tests to fail is a little surprising. It does look like there are two test files that were added in https://github.com/sorbet/sorbet/pull/2105 when the document highlight feature was added, but both of those tests only use the existing `def`/`usage` assertions that things like jump-to-def and find-all-refs use.

That's probably why this feature hasn't been graduated out of the experimental features, and I don't really want to change that right now. I just want https://sorbet.run examples to not suffer from this bug.

**before**

Screenshot from 2022-02-26 18-41-07

**after**

Screenshot from 2022-02-26 18-38-38

### DIFF
--- a/main/lsp/LSPTask.cc
+++ b/main/lsp/LSPTask.cc
@@ -380,7 +380,11 @@ LSPTask::extractLocations(const core::GlobalState &gs,
                           vector<unique_ptr<Location>> locations) const {
     auto queryResponsesFiltered = filterAndDedup(gs, queryResponses);
     for (auto &q : queryResponsesFiltered) {
-        addLocIfExists(gs, locations, q->getLoc());
+        if (auto *send = q->isSend()) {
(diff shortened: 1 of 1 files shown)


jez:
There the fact that this didn't cause any tests to fail is a little surprising. It does look like there are two test files that were added in https://github.com/sorbet/sorbet/pull/2105 when the document highlight feature was added, but both of those tests only use the existing `def`/`usage` assertions that things like jump-to-def and find-all-refs use.

That's probably why this feature hasn't been graduated out of the experimental features, and I don't really want to change that right now. I just want https://sorbet.run examples to not suffer from this bug.

aprocter (approved):

PR #5376, opened 2022-02-28
change CLI test entrypoints to be named `test.sh` with output `test.out`

froydnj:
### Motivation

A papercut-y aspect of autocorrect tests (really, all CLI tests) is all the ceremony that goes into creating a new one: you need a long-ish-ly named shell script, a similarly-named `.out` file, and so forth.  For autocorrect tests, at least, the shell scripts are practically identical.

I think the intention here was in the right place.  But with nearly 200 CLI tests, and new ones getting very distinct names, the long filenames become somewhat grating.

This PR proposes that all CLI tests should have a driver named simply `test.sh`, with their accompanying output `test.out`.  Creating a new autocorrect test should simply be a matter of copying an existing directory, renaming the ruby file, tweaking the shell script, and you're off to the races.  This is not quite perfect (a further refinement might make the ruby file within be simply named `test.rb` so that there was no particular dependence between the filenames and the test names), but I think it is slightly better than before.

The original motivation was that we might combine the autocorrect tests to be run via a single driver.  I think that is still a possibility, though I was unsure how to get that to work within the context of the update framework of the CLI tests.  Maybe `cli_test.bzl` would recognize that tests beginning with `"autocorrect"` would be handled specially?

### Test plan

See included automated tests.  I verified that `./bazel query 'deps(//test/cli:cli, 1)'` gave the same output at each working step along the way.

The automated changes are not quite perfectly automated: I had to revert changes to `test/cli/compiler`, because that directory is really a framework for creating new tests, rather than a test directory itself.


jez (approved):
This is amazing, thanks!

When I took inspiration from Sorbet's Bazel-driven CLI tests in a side project I worked on once, I used the names `run.sh` and `run.sh.exp`

https://github.com/jez/as-tree/tree/master/test/cli/version

But these names are fine too. Thanks again!

PR #5377, opened 2022-02-28
pre-work: Track start/end tokens (not offsets)

jez:
### Motivation

Some future parse changes are going to start looking at the indentation level of
entire reduced blocks of code, and it'll be convenient to know "first token in
that whole reduced value" because the token stores the `lineStart` offset that's
useful for indentation comparisons.

### Test plan

Existing tests, as this is a no-op change.

### DIFF
--- a/parser/parser/cc/token.cc
+++ b/parser/parser/cc/token.cc
@@ -1,4 +1,5 @@
 #include <cassert>
+#include <iostream>
 #include <map>
 #include <ruby_parser/token.hh>
 
--- a/parser/parser/include/ruby_parser/location.hh
+++ b/parser/parser/include/ruby_parser/location.hh
@@ -1,26 +1,38 @@
 #ifndef RUBY_PARSER_LOCATION_HH
 #define RUBY_PARSER_LOCATION_HH
 
+#include <cassert>
 #include <cstddef>
 #include <iostream>
 #include <stdint.h>
 
+#include "token.hh"
+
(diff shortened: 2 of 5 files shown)


parser/parser/cc/token.cc
```diff
@@ -1,4 +1,5 @@
 #include <cassert>
+#include <iostream>
```
froydnj, line 2: This looks unused?
```suggestion
```

parser/parser/include/ruby_parser/location.hh
```diff
@@ -1,26 +1,38 @@
...
+#include "token.hh"
+
 namespace ruby_parser {
 
 struct location {
-    size_t begin;
-    size_t end;
-    size_t lineStart;
+    token_t begin;
+    token_t end;
```
froydnj, line 15: I confess that I am not super-excited about making tokens reachable from more places...but I don't think we have a viable approach to making tokens *not* tied to the lifetime of the parser, so my lack of excitement doesn't really count for much.

I do like that by getting rid of `lineStart` we can make locations smaller, though!

froydnj (approved):

parser/parser/cc/token.cc
jez: It's used by the printing code below (there is a definition of `operator<<`.

It used to get this include for free because the order of the header file includes.

PR #5378, opened 2022-02-28
Fix signature for `Array#at`

jez:
### Motivation

If you want a similar method that will raise instead of returning `nil`
sometimes, you can use `Array#fetch` instead.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_LEi2wldfWMLchh
→ https://go/builds/bui_LEi2EexjjP2o7T

aprocter (approved):

PR #5386, opened 2022-03-01
Increment a counter when we compute `findLineBreaks`

jez:
### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_LEsUhsOpAr3lf7
→ https://go/builds/bui_LEsUsTpcFRGZaf

jez:
I wanted to see how frequently we ended up forcing the memoization in `lineBreaks` (I was curious to see if it was already getting run somewhere for every file, or if it was only getting run for a handful of files, like maybe only those with errors.)

It looked like it was the former—I only saw the counter reach 14 on a run over Stripe's codebase.

PR #5387, opened 2022-03-01
Move the signature finding code into separate class

ilyailya:
Per the tin. The change moves a code, which finds a `sig` location by a symbol reference into a separate class.

### Motivation

I need this to move forward with a "Move method" refactoring
https://github.com/sorbet/sorbet/pull/5309

### Test plan

See included automated tests.


froydnj (approved):
Cool, thank you!

PR #5389, opened 2022-03-01
don't bother with `LocalVariable` when `NameRef` will do in completion code

froydnj:
I'm working on completion code for instance variables (more generally, "fields" in Sorbet parlance).  The way the code is looking right now, it should be possible to largely share the completion code for local variables and for fields.

Except for one small wrinkle: the completion code for locals works by finding all the `LocalVariable`s in a method and operating on those.  The completion code for instance variables is going to work by finding all `UnresolvedIdent`s and gathering up their names (as `NameRef`s) and working on those.  So we have a mismatch.

The saving grace is that the completion code for locals really only ever cares about the names of the `LocalVariable`s, which are, conveniently, `NameRef`s.  So this PR proposes to make the completion code and all supporting bits operate in terms of `NameRef`s rather than `LocalVariable`s.

I guess `LocalVarFinder` should really be `LocalNameFinder`, but I was lazy; please let me know if I should not be lazy.

### Motivation

Making future changes easier.  The code should also be slightly more efficient because we're operating on smaller data structures now.

### Test plan

This should be a no-op change and existing tests should cover this.

### DIFF
--- a/main/lsp/LocalVarFinder.h
+++ b/main/lsp/LocalVarFinder.h
@@ -15,7 +15,7 @@ class LocalVarFinder {
     // flattened at this point. (LSP code should try to make minimal assumptions to be robust to changes.)
     std::vector<core::MethodRef> methodStack;
 
-    std::vector<core::LocalVariable> result_;
+    std::vector<core::NameRef> result_;
 
 public:
     LocalVarFinder(core::MethodRef targetMethod, core::Loc queryLoc) : targetMethod(targetMethod), queryLoc(queryLoc) {}
@@ -27,7 +27,7 @@ public:
     ast::ExpressionPtr preTransformClassDef(core::Context ctx, ast::ExpressionPtr classDef);
     ast::ExpressionPtr postTransformClassDef(core::Context ctx, ast::ExpressionPtr classDef);
 
-    const std::vector<core::LocalVariable> &result() const;
+    const std::vector<core::NameRef> &result() const;
 };
 }; // namespace sorbet::realmain::lsp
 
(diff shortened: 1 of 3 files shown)


main/lsp/LocalVarFinder.h
```diff
@@ -27,7 +27,7 @@ class LocalVarFinder {
     ast::ExpressionPtr preTransformClassDef(core::Context ctx, ast::ExpressionPtr classDef);
     ast::ExpressionPtr postTransformClassDef(core::Context ctx, ast::ExpressionPtr classDef);
 
-    const std::vector<core::LocalVariable> &result() const;
+    const std::vector<core::NameRef> &result() const;
```
jez, line 30: How are you planning to change this to account for instance variables? I hope that it'll still be possible to have separate results for e.g. "all local variables" and "all instance variables", not one big `vector<NameRef>` that mixes the two.

jez:
> I guess LocalVarFinder should really be LocalNameFinder, but I was lazy; please let me know if I should not be lazy.

Maybe it should be called `VariableFinder`, because it will soon find local variables, instance variables (and maybe class variables like `@@foo` and global variables like `$foo`).

jez (approved):
rename looks good. question about the API for disambiguating locals from instance variables, but i'm happy to have that discussion in the next PR

main/lsp/LocalVarFinder.h
froydnj: I was going to have a separate `FieldFinder`, since instance/class variables are not limited to a target method, but to an entire class, and globals are, well, globals.  (I am not 100% sure how to handle globals yet, but I am less concerned about them.)  There's similar logic, but I don't think there's enough to stuff them into a single class.

AFAICT, it will still be possible to have separate "all local variables" and "all instance variables", but you can't complete both sets in the same list of completion results, because instance variables are effectively namespaced by virtue of their leading `@`.  (If `completion.cc` implemented a fuzzy completion algorithm, rather than the prefix-based algorithm it has, then you could of course complete from both sets in the same list.)  So the methods for finding them are slightly different, but as far as completion code is concerned, they can be treated (almost) identically after the lists of potential completions are found.

(This would also be an explanation as to why I'm not renaming it `VariableFinder`, since `LocalVarFinder` is still only for local variables; I would consider `LocalNameFinder` as a reasonable substitute.)

PR #5391, opened 2022-03-01
add completion for instance/class variables

froydnj:
### Motivation

More features in the IDE.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -8,6 +8,7 @@
 #include "common/sort.h"
 #include "common/typecase.h"
 #include "core/lsp/QueryResponse.h"
+#include "main/lsp/FieldFinder.h"
 #include "main/lsp/LocalVarFinder.h"
 #include "main/lsp/NextMethodFinder.h"
 #include "main/lsp/json_types.h"
@@ -483,6 +484,38 @@ unique_ptr<CompletionItem> getCompletionItemForLocalName(const core::GlobalState
     return item;
 }
 
+vector<core::NameRef> fieldsForClass(LSPTypecheckerDelegate &typechecker, const core::ClassOrModuleRef klass,
+                                     const core::Loc queryLoc, ast::UnresolvedIdent::Kind kind) {
+    const auto &gs = typechecker.state();
+    auto files = vector<core::FileRef>{};
+    for (auto loc : klass.data(gs)->locs()) {
+        files.emplace_back(loc.file());
+    }
+    auto resolved = typechecker.getResolved(files);
+
+    // Instantiate fieldFinder outside loop so that result accumualates over every time we TreeMap::apply
+    FieldFinder fieldFinder(klass, queryLoc, kind);
+    for (auto &t : resolved) {
+        auto ctx = core::Context(gs, core::Symbols::root(), t.file);
+        t.tree = ast::TreeMap::apply(ctx, fieldFinder, move(t.tree));
+    }
+
+    auto result = fieldFinder.result();
+    fast_sort(result, [&gs](const auto &left, const auto &right) {
+        // Sort by actual name, not by NameRef id
+        if (left != right) {
+            return left.shortName(gs) < right.shortName(gs);
+        } else {
+            return left.rawId() < right.rawId();
+        }
+    });
+
+    // Dedup
+    auto it = unique(result.begin(), result.end());
+    result.erase(it, result.end());
+    return result;
+}
+
 vector<core::NameRef> localNamesForMethod(LSPTypecheckerInterface &typechecker, const core::MethodRef method,
                                           const core::Loc queryLoc) {
     const auto &gs = typechecker.state();
@@ -838,9 +871,29 @@ vector<unique_ptr<CompletionItem>> CompletionTask::getCompletionItems(LSPTypeche
     // ----- locals -----
 
     vector<core::NameRef> similarLocals;
+    bool localsAreFields = false;
     if (params.enclosingMethod.exists()) {
         Timer timeit(gs.tracer(), LSP_COMPLETION_METRICS_PREFIX ".determine_locals");
-        auto locals = localNamesForMethod(typechecker, params.enclosingMethod, params.queryLoc);
+        vector<core::NameRef> locals;
+
+        // Slyly reuse `UnresolvedIdent::Kind` to both determine what kind of
+        // thing we're going to find and to determine the search space for
+        // `fieldsForClass`.
+        auto kind = ast::UnresolvedIdent::Kind::Local;
+        if (params.prefix.size() >= 2 && absl::StartsWith(params.prefix, "@@")) {
+            kind = ast::UnresolvedIdent::Kind::Class;
+        } else if (params.prefix.size() >= 1 && absl::StartsWith(params.prefix, "@")) {
+            kind = ast::UnresolvedIdent::Kind::Instance;
+        }
+
+        if (kind == ast::UnresolvedIdent::Kind::Local) {
+            locals = localNamesForMethod(typechecker, params.enclosingMethod, params.queryLoc);
+            localsAreFields = false;
+        } else {
+            auto klass = params.enclosingMethod.data(gs)->owner;
+            locals = fieldsForClass(typechecker, klass, params.queryLoc, kind);
+            localsAreFields = true;
+        }
         similarLocals = allSimilarLocalNames(gs, locals, params.prefix);
     }
 
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -523,7 +523,7 @@ vector<core::NameRef> allSimilarFields(const core::GlobalState &gs, core::ClassO
     return result;
 }
 
-vector<core::NameRef> allSimilarFieldsForClass(LSPTypecheckerDelegate &typechecker, const core::ClassOrModuleRef klass,
+vector<core::NameRef> allSimilarFieldsForClass(LSPTypecheckerInterface &typechecker, const core::ClassOrModuleRef klass,
                                                const core::Loc queryLoc, ast::UnresolvedIdent::Kind kind,
                                                string_view prefix) {
     const auto &gs = typechecker.state();
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -501,6 +501,8 @@ vector<core::NameRef> allSimilarFields(const core::GlobalState &gs,
                 continue;
             }
 
+            // TODO: this does prefix matching for instance/class variables, but our
+            // completion for locals matches anywhere in the name
             if (hasPrefixedName(gs, name, prefix)) {
                 result.emplace_back(name);
             }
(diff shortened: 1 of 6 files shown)


main/lsp/requests/completion.cc
```diff
@@ -734,6 +767,24 @@ vector<unique_ptr<CompletionItem>> allSimilarConstantItems(const core::GlobalSta
     return items;
 }
 
+string_view identPrefix(string_view name) {
```
froydnj, line 770: This was a dumb idea I eventually realized we didn't need; we can just check the prefix in `getCompletionItems` to determine whether we need to operate on local variables or on some kind of field.

froydnj:
This is not perfect, but it's at least an improvement on the current situation.

My (possibly Stripe-influenced) perspective on [7366d6b](https://github.com/sorbet/sorbet/pull/5391/commits/7366d6bd7cce65eff4d82c73dee2e242bcd2020f) is that accessing superclass instance/class variables from a subclass is uncommon and so it's not worth worrying about this limitation unless somebody complains.

main/lsp/requests/completion.cc
```diff
@@ -483,6 +484,38 @@ unique_ptr<CompletionItem> getCompletionItemForLocalName(const core::GlobalState
...
+                                     const core::Loc queryLoc, ast::UnresolvedIdent::Kind kind) {
+    const auto &gs = typechecker.state();
+    auto files = vector<core::FileRef>{};
+    for (auto loc : klass.data(gs)->locs()) {
+        files.emplace_back(loc.file());
+    }
+    auto resolved = typechecker.getResolved(files);
+
+    // Instantiate fieldFinder outside loop so that result accumualates over every time we TreeMap::apply
+    FieldFinder fieldFinder(klass, queryLoc, kind);
```
jez, line 497: This doesn't look right. The reason why we have to use a tree walk for local variables is because there isn't a list of all local variables on the symbol table.

When we have an `enclosingMethod` (which we get from the `IdentResponse`), we can easily ask `enclosingMethod.data(ctx)->owner` for all of its fields. This would look much the same as it does today for `allSimilarConstantItems` (basically the same recursive algorithm) and would fix the problem that you're having that the current implementation doesn't handle instance variables declared in parent classes.

main/lsp/requests/completion.cc
```diff
@@ -838,8 +871,28 @@ vector<unique_ptr<CompletionItem>> CompletionTask::getCompletionItems(LSPTypeche
...
+
+        // Slyly reuse `UnresolvedIdent::Kind` to both determine what kind of
+        // thing we're going to find and to determine the search space for
+        // `fieldsForClass`.
+        auto kind = ast::UnresolvedIdent::Kind::Local;
+        if (params.prefix.size() >= 2 && absl::StartsWith(params.prefix, "@@")) {
+            kind = ast::UnresolvedIdent::Kind::Class;
+        } else if (params.prefix.size() >= 1 && absl::StartsWith(params.prefix, "@")) {
+            kind = ast::UnresolvedIdent::Kind::Instance;
+        }
```
jez, line 886: If we always have to compute this anyways, let's just add this field to the `IdentResponse`, because it should always be known at the point where the `IdentResponse` is pushed.

jez (changes requested):
I don't think it should be too hard to fix the problems you mention around parent classes, because the code almost already exists to handle completions for constants.

main/lsp/requests/completion.cc
froydnj: I think overloading `UnresolvedIdent::Kind` here is appropriate, if a bit hacky; I don't think putting `UnresolvedIdent::Kind` in `IdentResponse` is a good thing to do.  Unless you mean essentially duplicating `UnresolvedIdent::Kind` in `IdentResponse`, which might not be a bad idea.  (Though if we did some kind of symbol table search, then we wouldn't need to compute this bit.)

main/lsp/requests/completion.cc
froydnj: Searching through the symbol table has the shortcoming that instance/class variables that are not typed don't get entered.  So, e.g. if you take this PR and run:

`./bazel-bin/main/sorbet -print symbol-table test/testdata/lsp/completion/instance_var.rb |grep field`

you get nothing.

I guess we could adopt a hybrid scheme, where files `typed: strict` and above get an implementation that searches through the symbol table (and is therefore exact) whereas files `typed: true` and below get purely syntactical completion results, which is not great, but is better than nothing?

jez: yep. i'm fine taking whatever approach to the variables that haven't been declared, but was just hoping that we could at least get all the ones we for sure knew about.

if we want i'm happy to think about approaches to get more undeclared instance variables into the symbol table.

main/lsp/requests/completion.cc
```diff
@@ -839,8 +939,30 @@ vector<unique_ptr<CompletionItem>> CompletionTask::getCompletionItems(LSPTypeche
...
     vector<core::NameRef> similarLocals;
     if (params.enclosingMethod.exists()) {
-        auto locals = localNamesForMethod(typechecker, params.enclosingMethod, params.queryLoc);
-        similarLocals = allSimilarLocalNames(gs, locals, params.prefix);
+        // Slyly reuse `UnresolvedIdent::Kind` to both determine what kind of
+        // thing we're going to find and to determine the search space for
+        // `fieldsForClass`.
+        //
+        // TODO: for empty prefixes, we would like to complete instance/class
+        // variables along with locals.
```
froydnj, line 947: Trying to write tests for this was kind of hard.  This makes me believe it should not be that hard:

https://github.com/sorbet/sorbet/blob/master/main/lsp/requests/completion.cc#L1074-L1081

But writing an experimental test:

```ruby
  sig {void}
  def assign_method
    x = 
    #   ^ completion: (nothing)
    y = 5
  end
```

indicates that Sorbet thinks there are a lot of things to complete here:

```
test/testdata/lsp/completion/instance_var_strict.rb:28: ERROR: Expected completion contents:
28:     x =
           ^ completion: (nothing)
Found completion contents:
28:     x =
           ^ completion: x, y, assign_method, more_method, other_method, some_method, CSV, class, itself, object_id, singleton_class, then, yield_self, !~, ===, =~, Array, Array, BigDecimal, Complex, Float, Hash, Integer, Pathname, Rational, String, URI, __callee__, __dir__, __method__, `, abort, at_exit, autoload, autoload?, binding, block_given?, caller, caller, caller, caller_locations, caller_locations, catch, clone, define_singleton_method, define_singleton_method, display, dup, enum_for, enum_for, eql?, eval, exec, exit, exit, exit!, extend, fail, fail, fail, fail, fork, fork, format, freeze, frozen?, gem, gets, global_variables, hash, initialize_copy, inspect, instance_of?, instance_variable_defined?, instance_variable_get, instance_variable_set, instance_variables, is_a?, iterator?, kind_of?, lambda, load, local_variables, loop, loop, method, methods, nil?, open, open, p, pp, print, printf, private_methods, proc, protected_methods, public_method, public_methods, public_send, putc, puts, raise, raise, raise, raise, rand, rand, rand, rand, readline, readlines, remove_instance_variable, require, require_relative, respond_to?, select, send, send, singleton_method, singleton_methods, sleep, sleep, sprintf, srand, syscall, system, taint, tainted?, tap, test, throw, to_enum, to_enum, to_s, trap, trap, trust, undef, untaint, untrust, untrusted?, warn, !, !=, ==, __id__, __send__, equal?, instance_eval, instance_eval, instance_exec
```

and it didn't seem worth it to bake that into the test.

main/lsp/requests/completion.cc
```diff
@@ -839,8 +939,30 @@ vector<unique_ptr<CompletionItem>> CompletionTask::getCompletionItems(LSPTypeche
...
+        //
+        // TODO: for a prefix of just "@", we should provide class variables
+        // along with instance variables.
```
froydnj, line 950: There are tests that try completing after just a `@`, but I don't think the necessary information makes its way into the AST to enable us to see that we have that prefix there.  Unless there need to be some extra gymnastics in `CompletionTask::runRequest` to look for a bare `@` prior to the completion point?

main/lsp/requests/completion.cc
```diff
@@ -483,6 +489,100 @@ unique_ptr<CompletionItem> getCompletionItemForLocalName(const core::GlobalState
...
+        for (auto [name, sym] : ancestor.data(gs)->members()) {
+            if (!sym.isFieldOrStaticField()) {
+                continue;
+            }
+
+            // TODO: this does prefix matching for instance/class variables, but our
+            // completion for locals matches anywhere in the name
+            if (hasPrefixedName(gs, name, prefix)) {
+                result.emplace_back(name);
+            }
```
froydnj, line 507: My sense (possibly informed too much by my own personal idiosyncrasies) is that people would want prefix matching on instance (resp. class) variables, but I'm happy to hear alternative points of view.

main/lsp/requests/completion.cc
jez: You can use `...` to encode a prefix

```
# ^ completion: x, y, ...
```

jez: Some other things to note:

- if you use a `module` not a `class`, you won't have `Kernel` by default, which means you won't have quite so many garbage methods. There will still be some but not nearly as many.
- maybe it's fine to never suggest instance variables here? It doesn't seem like a terrible restriction to say that you have to type `@` to get instance variable suggestions.

main/lsp/requests/completion.cc
jez: Yes, when I looked at this it looked like the parser wasn't always good enough:

https://sorbet.run/?arg=--print=parse-tree-whitequark#%23%20typed%3A%20true%0Adef%20foo%0A%20%20puts%20'this%20is%20missing'%0A%20%20%40%0A%20%20puts%20'this%20is%20after'%0Aend

I was hoping to snipe you into writing test cases for what didn't work, and then using http://go/bad-parse for whatever doesn't work because the parser isn't yet good enough. Fixing this doesn't seem any harder than getting `x.` to parse, but we just have to put time into it.

test/testdata/lsp/completion/instance_var.rb
```diff
@@ -0,0 +1,133 @@
...
+  def other_method
+    my = 5
+    @my
+    #  ^ completion: @my, @my_ivar
+  end
+
+  def more_method
+    @ # error: unexpected
+    #^ completion: (nothing)
+  end
```
jez, line 18: For the places where we wish there were completion results but it's not doing the right thing, can we leave behind comments in case someone fixes this in the future and is confused why tests are failing?

test/testdata/lsp/completion/instance_var.rb
```diff
@@ -0,0 +1,133 @@
...
+end
+
+# See above comments about completing variables in static-init.
+class Inheriting < Superclass
+  @@s
+  #  ^ completion: (nothing)
+
+  # Because our completion algorithm operates purely syntactically for
+  # untyped instance variables, it misses picking up instance variables from
+  # `Superclass`.
```
jez, line 128: it looks like this comment is out of date now. i think that maybe it should read something like "because `super_ivar` is not actually declared with `T.let` ..."

jez (approved):

PR #5397, opened 2022-03-02
make `GlobalState::*InitForFile` functions only take `FileRef` instead of `Loc`

froydnj:
### Motivation

`GlobalState::staticInitForFile` and `GlobalState::lookupStaticInitForFile` take `Loc` and generally only care about the file attached to that `Loc`.  But because of this, we end up synthesizing a `Loc` at the callsites to those functions, and we do it in inconsistent ways.  This scheme also results in less clear code: one wonders why you have a full `Loc` for a function that should seemingly only care about the file.

We should standardize: the aforementioned functions should only take `FileRef` and be responsible for synthesizing full `Loc`s for internal purposes.  We can then ensure that things are much more consistent (e.g. the symbols entered by `staticInitForFile` no longer depend on the first time `staticInitForFile` is called and the `Loc` passed to it).

### Test plan

See included automated tests.


core/GlobalState.cc
```diff
@@ -2075,22 +2075,22 @@ MethodRef GlobalState::lookupStaticInitForClass(ClassOrModuleRef klass) const {
     return ref;
 }
 
-MethodRef GlobalState::staticInitForFile(Loc loc) {
-    auto nm = freshNameUnique(core::UniqueNameKind::Namer, core::Names::staticInit(), loc.file().id());
+MethodRef GlobalState::staticInitForFile(FileRef file) {
+    auto nm = freshNameUnique(core::UniqueNameKind::Namer, core::Names::staticInit(), file.id());
     auto prevCount = this->methodsUsed();
+    auto loc = core::Loc::none(file);
```
jez, line 2081: What do you think about making `staticInitForFile` still take a `Loc` instead of just a file so that we don't have to use `Loc::none` and have the `???` show up in the symbol table?

It seems like most of the simplifying changes that you want still happen if only `lookupStaticInitForFile` is changed to take a `FileRef`.

jez:
The fact that this PR causes packager stuff to change makes me suspicious that this is as benign a change as the motivation says. I'm curious if my suggestion to only change `lookupStaticInitForX` (and not `staticInitForX`) fixes the packager bug.

core/GlobalState.cc
froydnj: Yeah, I'm kind of thinking the compiler-related failures are because we don't have a "real" `Loc` for things, so we might just have to still have this weird "for a file, but actually cares about the position in the file".

jez (approved):

PR #5403, opened 2022-03-03
Debugging notes for required params after optional params error

ghiculescu:
I spent a bunch of time today debugging an error I was seeing on a Rails 7 app, where we were getting this error and it was pointing here: https://github.com/rails/rails/blob/2459c20afb508c987347f52148210d874a9af4fa/activerecord/lib/active_record/persistence.rb#L405

Eventually, with the help of @jeffcarbs and @paracycle I tracked it down to a `def self.update!` in an Active Record model. This method had a `sig { void }` but it seems like for some reason the names matching still triggered this error.

I'm not sure how to fix that or if it's even a bug, but I thought some debugging notes might help the next person to come across it. If it is considered a bug, happy to provide more info to help with replication.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/signature.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/signature.rb
@@ -103,6 +103,12 @@ class T::Private::Methods::Signature
         if @arg_types.length > @req_arg_count
           # Note that this is actually is supported by Ruby, but it would add complexity to
           # support it here, and I'm happy to discourage its use anyway.
+          #
+          # If you are seeing this error and surprised by it, it's possible that you have
+          # overridden the method described in the error message. For example, Rails defines
+          # def self.update!(id = :all, attributes)
+          # on AR models. If you have also defined `self.update!` on an AR model you might
+          # see this error. The simplest resolution is to rename your method.
           raise "Required params after optional params are not supported in method declarations. Method: #{method_desc}"
         end
         @arg_types << [param_name, type]


jez (approved):
Thanks. Do you mind sharing a test case? Should we add your test case into the sorbet-runtime test suite?

ghiculescu:
This passes statically but raises at runtime:

```ruby
# typed: true
require "sorbet-runtime"

module A
  def foo(x = 1, y)
  end
end

class B
  extend T::Sig
  extend A

  sig { params(z: String).void }
  def self.foo(z)
  end
end

B.foo("hi")
```

As does this

```ruby
# typed: true
require "sorbet-runtime"

class A
  def self.foo(x = 1, y)
  end
end

class B < A
  extend T::Sig

  sig { params(z: String).void }
  def self.foo(z)
  end
end

B.foo("hi")
```

Would you consider that a bug?

ghiculescu:
@jez not sure if you saw ^ but let me know if you'd like me to add that to https://github.com/sorbet/sorbet/tree/master/gems/sorbet-runtime/test somewhere

paracycle:
@ghiculescu I think [this is the call](https://github.com/sorbet/sorbet/blob/9f6db98c0caa39e03668a7d3da808e53383ef21a/gems/sorbet-runtime/lib/types/private/methods/signature_validation.rb#L39) that is ultimately causing that error.

In your first example, `A#foo` ends up being a super method of `B.foo` (due to `extend A`) and when Sorbet runtime tries to process the super method, it can't deal with the required params after optional params. Similarly, in the second example, `A.foo` is the super method of `B.foo`.

This is arguably a bug, as in, theoretically, it can be better handled by Sorbet runtime, wrt to super methods.

ghiculescu:
Yes, I think you're right. It sounds like maybe we want to add some more scenarios to [this path](https://github.com/sorbet/sorbet/blob/9f6db98c0caa39e03668a7d3da808e53383ef21a/gems/sorbet-runtime/lib/types/private/methods/signature_validation.rb#L27)?

PR #5405, opened 2022-03-03
Rewrite `Class.new do ... end` blocks to add a `T.bind`

Morriar:
### Motivation

We added more functionality to the `ClassNew` rewriter to provide better handling for `Class.new do ... end` calls.

Before this PR, the rewriter was only handling `Class.new do ... end` when it was assigned to a constant literal:

```rb
A = Class.new(Parent) do
  # ...
end
```

Was rewritten into (and it still is the case after this PR)
 
```rb
class A < Parent
  # ...
end
```

In the cases where we cannot create a class definition for the `Class.new` send, we're now adding a `T.bind` at the beginning of the block.

So assignments to a non-constant literal such as

```rb
c = Class.new(Parent) do
  # ...
end
```

Are now rewritten into

```rb
c = Class.new(Parent)
  T.bind(self, T.class_of(Parent))
  # ...
end
```

And sends such as

```rb
Class.new(Parent) do
  # ...
end
```

Are now rewritten into

```rb
Class.new(Parent)
  T.bind(self, T.class_of(Parent))
  # ...
end
```

### Test plan

See included automated tests.

Closes https://github.com/sorbet/sorbet/pull/4766.

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -357,6 +357,14 @@ public:
         return Constant(loc, core::Symbols::T());
     }
 
+    static ExpressionPtr Bind(core::LocOffsets loc, ExpressionPtr value, ExpressionPtr type) {
+        return Send2(loc, T(loc), core::Names::bind(), loc, std::move(value), std::move(type));
+    }
+
+    static ExpressionPtr ClassOf(core::LocOffsets loc, ExpressionPtr value) {
+        return Send1(loc, T(loc), core::Names::classOf(), loc, std::move(value));
+    }
+
     static ExpressionPtr Let(core::LocOffsets loc, ExpressionPtr value, ExpressionPtr type) {
         return Send2(loc, T(loc), core::Names::let(), loc, std::move(value), std::move(type));
     }
--- a/rewriter/ClassNew.cc
+++ b/rewriter/ClassNew.cc
@@ -10,21 +10,9 @@ using namespace std;
 
 namespace sorbet::rewriter {
 
-vector<ast::ExpressionPtr> ClassNew::run(core::MutableContext ctx, ast::Assign *asgn) {
+vector<ast::ExpressionPtr> rewriteAsClassDef(core::MutableContext ctx, ast::Assign *asgn) {
     vector<ast::ExpressionPtr> empty;
     auto loc = asgn->loc;
-
-    if (ctx.state.runningUnderAutogen) {
-        // This is not safe to run under autogen, because we'd be outputing
-        // autoloader files that predeclare the class and cause "warning:
-        // already initialized constant" errors
-        return empty;
-    }
-
-    auto lhs = ast::cast_tree<ast::UnresolvedConstantLit>(asgn->lhs);
-    if (lhs == nullptr) {
-        return empty;
-    }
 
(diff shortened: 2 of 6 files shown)


elliottt:

→ https://go/builds/bui_LIVJugjmkYFdC6
→ https://go/builds/bui_LIVJRztUWf4UG6

rewriter/ClassNew.cc
```diff
@@ -82,4 +70,88 @@ vector<ast::ExpressionPtr> ClassNew::run(core::MutableContext ctx, ast::Assign *
...
+        type = ast::MK::ClassOf(send->loc, std::move(target));
+    }
+
+    auto bind = ast::MK::Bind(send->loc, ast::MK::Self(send->loc), std::move(type));
+
+    ast::InsSeq::STATS_store blockStats;
+    blockStats.emplace_back(std::move(bind));
+
+    if (auto insSeq = ast::cast_tree<ast::InsSeq>(block->body)) {
+        for (auto &&stat : insSeq->stats) {
```
elliottt, line 115: I think using `&&` here is harmless, but it looks like the deduced type will be `ExpressionPtr &` so I would drop the extra `&`. Here's a similar example on godbolt showing the deduced type for an element of a vector of unique_ptrs: https://godbolt.org/z/TW1q64r8s
```suggestion
        for (auto &stat : insSeq->stats) {
```

rewriter/ClassNew.h
```diff
@@ -5,21 +5,50 @@
 namespace sorbet::rewriter {
 
 /**
- * This class desugars things of the form
+ * This class actually contains three different rewriters depending on the kind of `Class.new` we're having.
```
elliottt, line 8:
```suggestion
 * This class actually contains three different rewriters depending on the kind of `Class.new` we encounter:
```

ast/Helpers.h
```diff
@@ -357,6 +357,14 @@ class MK {
         return Constant(loc, core::Symbols::T());
     }
 
+    static ExpressionPtr Bind(core::LocOffsets loc, ExpressionPtr value, ExpressionPtr type) {
+        return Send2(loc, T(loc), core::Names::bind(), loc, std::move(value), std::move(type));
+    }
+
+    static ExpressionPtr ClassOf(core::LocOffsets loc, ExpressionPtr value) {
+        return Send1(loc, T(loc), core::Names::classOf(), loc, std::move(value));
+    }
```
elliottt, line 366: What do you think about moving these helpers into `ClassNew.cc`? We can always extract them if we find they're useful elsewhere.

elliottt (approved):
This is great! I love the use of `T.bind` to handle cases where we can't desugar directly to a class definition 👍

ast/Helpers.h
jez: (I'm happy to defer to trevor here, but fwiw we already have helpers for most other things like `T.let` and `T.cast` in this file. Seems fine to me to add these as well.)

jez (commented):
Looks great to me thanks!

ast/Helpers.h
Morriar: Yeah I think it's better to have all those helpers together here so it's easier to find them when needed and avoid duplication. WDYT?

PR #5407, opened 2022-03-04
Stub implementation of stale-state hover and codeAction

aprocter:
This adds a stubbed implementation of stale-state `textDocument/hover` and `textDocument/codeAction` queries.

Why those two queries? Because it seems that this is the minimal set we need to stub in order to get interesting behavior out of vscode.

### Motivation
Would be nice to review the basic design around locking/multithreading separately from logic tweaks.

### Test plan
No unit tests for the moment (open to ideas on how to improve that situation!) but everything should be hidden behind the `--enable-experimental-lsp-stale-state` flag.

For a basic smoke test, I've set up a sandbox project in vscode with `--sleep-in-slow-path` and `--enable-experimental-lsp-stale-state` and verified that we get stub answers for hover while a slowpath is running:

Screen Shot 2022-03-10 at 11 58 16 AM

- [x] TODO: Test by hand against Stripe's codebase to make sure that the `--enable-experimental-lsp-stale-state` flag is indeed properly gating this feature.

### DIFF
--- a/main/lsp/protocol.cc
+++ b/main/lsp/protocol.cc
@@ -228,10 +228,40 @@ optional<unique_ptr<core::GlobalState>> LSPLoop::runLSP(shared_ptr<LSPInput> inp
                 }
 
                 // Before giving up the lock, check if the typechecker is running a slow path and if the task at the
-                // head of the queue can preempt. If it is, we may be able to schedule a preemption.
-                // Don't bother scheduling tasks to preempt that only need the indexer.
-                // N.B.: We check `canPreempt` last as it is mildly expensive for edits (it hashes the files)
+                // head of the queue can either operate on stale data (i.e., the GlobalState just prior to the change
+                // that required the slow path to run) or preempt the currently running slow path. (Note that we don't
+                // bother with either one in cases where we only need the indexer. TODO(aprocter): is that restriction
+                // actually appropriate for stale-data tasks?)
                 auto &frontTask = taskQueue->pendingTasks.front();
+
+                // Note that running on stale data is an experimental feature, so we hide it behind the
+                // --enable-experimental-lsp-stale-state flag.
+                if (opts.lspStaleStateEnabled && frontTask->finalPhase() == LSPTask::Phase::RUN &&
+                    epochManager->getStatus().slowPathRunning && frontTask->canUseStaleData()) {
+                    logger->debug("Trying to run on stale data");
+
+                    frontTask = typecheckerCoord.syncRunOnStaleState(move(frontTask));
+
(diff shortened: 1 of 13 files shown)


aprocter-stripe:

→ https://go/builds/bui_LIT9mHdKg3DWsD
→ https://go/builds/bui_LIT9Z09YUDpbXX

main/lsp/protocol.cc
```diff
@@ -228,10 +228,40 @@ optional<unique_ptr<core::GlobalState>> LSPLoop::runLSP(shared_ptr<LSPInput> inp
...
+
+                    // If the coordinator has consumed the task, we know it was able to run it on stale state. Pop it
+                    // and move on to the next task.
+                    if (frontTask == nullptr) {
+                        logger->debug("Succeeded in running on stale data");
+                        taskQueue->pendingTasks.pop_front();
+                    }
+                    // If the coordinator has not consumed the task, that means it was not able to run it on stale
+                    // state, because no cancellationUndoState was present. This likely means that the typechecker has
+                    // not yet initialized the cancellationUndoState, so we will try again after a short delay.
```
froydnj, line 253: I am not sure this comment is correct.  Is it possible that the following global sequence of events occurs?

1. Check `epochManager->getStatus().slowPathRunning`, returns true
2. Write lock on the undo state is acquired because the typechecker finishes
3. We try to acquire the reader lock in `syncRunOnStaleState`
4. We need to block in acquiring the reader lock
5. Write lock dropped because we dropped the cancellation state
6. We acquire the reader lock in `syncRunOnStaleState`
7. We find a null cancellation state

So then it's not that the typechecker failed to initialize (how would that even happen?), but because the typechecker actually finished its work and therefore there's nothing to undo, because we have updated the state of the world.

I'm not sure that we have to fix this right now, but it would be good to a) log a TODO here and/or b) file a ticket at Stripe to track this.

froydnj (approved):
This looks largely reasonable.  Just one comment below.

main/lsp/protocol.cc
aprocter: (1) Oh yeah, the sequence of events you're describing here is absolutely possible (and is, I think, handled correctly; most likely we would come back around after 100ms, find that `slowPathRunning` is no longer true, and process the request as normal).

(2) The sequence of events I was assuming was in play here---which may also be possible, but I'm actually _not_ totally sure---is one where `slowPathRunning` has become true, but the typechecker has not yet actually initialized the `cancellationUndoState`.

I'll update the comment to reflect the possibility of (1), and a TODO to investigate whether and when (2) is actually possible.

PR #5408, opened 2022-03-04
Handle test package exports when stubbing a package

elliottt:
When comparing constant scopes, ensure that the `Test::` scope is handled when looking for parent or child packages that might own the unresolved symbol.

### Motivation

Stabilizing single-package rbi generation.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -415,6 +415,8 @@ private:
             }
 
             auto scopeSize = std::distance(start, scope.end());
+            ENFORCE(scopeSize > 0);
+
             const auto parent = absl::c_find_if(this->parents, [start, scopeSize, &scope](auto &p) {
                 return scopeSize == p.stub.fullName.size() && std::equal(start, scope.end(), p.stub.fullName.begin());
             });
@@ -484,7 +486,11 @@ private:
             sym = cls.data(gs)->owner;
         }
 
-        // Explicitly consider an empty top-level scope as one to skip.
+        // Explicitly consider an empty top-level scope as one to skip. This arises when the symbol passed in is
+        // `core::Symbols::root()`, which will always be the top-most parent for the `Nesting` linked list present for
+        // the `ResolutionItem` being stubbed. As this doesn't correspond to a prefix of the current package's
+        // namespace, we return false to signal that this scope doesn't need to be considered as either of the
+        // parent/child package special cases.
         if (res.empty()) {
             return false;
         }
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -356,8 +356,8 @@ private:
         PackageStub stub;
         vector<core::NameRef> exports;
 
-        // NOTE: these are the public-facing exports of the package that start with the `Test::` special prefix, not the
-        // names that are private, but shared with the implicit test package.
+        // NOTE: these are the public-facing exports of the package that start with the `Test::` special prefix.  They are not the names
+        // exported to the implicit test package via `export_for_test`.
         vector<core::NameRef> testExports;
 
         ParentPackageStub(const core::packages::PackageInfo &info) : stub{info} {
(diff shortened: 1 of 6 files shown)


test/cli/rbi-gen-single/family/bart/__package.rb
```diff
@@ -3,8 +3,11 @@
 class Family::Bart < PackageSpec
 
   import Family
+  import Family::Bart::Slingshot
   import Util
+  test_import Util::Testing
 
   export Family::Bart::Character
+  export Test::Family::Bart::BartTest
```
froydnj, line 11: This is important for checking that we handle test namespaces properly, but such a line would have no actual effect on typechecking, right?  (Just double-checking my own understanding here.)

resolver/resolver.cc
```diff
@@ -385,8 +407,20 @@ class ResolveConstantsWalk {
...
-            const auto parent = absl::c_find_if(this->parents, [&scope](auto &p) { return p.stub.fullName == scope; });
-            return parent != this->parents.end() && parent->exportsSymbol(cnst);
+            ENFORCE(!scope.empty());
+
+            auto start = scope.begin();
+            if (*start == core::Names::Constants::Test()) {
+                ++start;
+            }
+
+            auto scopeSize = std::distance(start, scope.end());
```
froydnj, line 417: Should this also be `ENFORCE`'d to be non-zero?  (That is, we should always see non-empty scopes even after stripping a top-level `Test` namespace.)

resolver/resolver.cc
```diff
@@ -448,6 +484,11 @@ class ResolveConstantsWalk {
             sym = cls.data(gs)->owner;
         }
 
+        // Explicitly consider an empty top-level scope as one to skip.
```
froydnj, line 487: WDYT about adding a comment here about how this comes up?  (Specifically in the context of test packages, but maybe packages in general--the former because we didn't have a case like this prior to this PR, but the latter if you thought of an interesting case where this would be true for non-test packages.)

froydnj (approved):
Thank you for adding the comments to existing functions!  A couple of questions, nothing major.

resolver/resolver.cc
```diff
@@ -312,18 +312,29 @@ class ResolveConstantsWalk {
 
         bool couldDefineChildNamespace(const core::GlobalState &gs, const std::vector<core::NameRef> &prefix,
                                        const std::vector<ast::ConstantLit *> &suffix) const {
+            ENFORCE(!prefix.empty());
+
+            auto start = prefix.begin();
+            if (*start == core::Names::Constants::Test()) {
+                ++start;
+            }
```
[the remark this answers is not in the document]
elliottt: That's correct. The `Test::` prefix is synthetic, and present implicitly for every package.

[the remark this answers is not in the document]
elliottt: There's currently no error added from processing a `__package.rb` that defines a package in `Test::`, but you do get an error once you start trying to define classes in that package. I think I'll make a separate PR to add an error for packages defined in `Test::`.

test/cli/rbi-gen-single/family/bart/__package.rb
elliottt: It would for packages that depend on that symbol from the `bart` package in their test package.

resolver/resolver.cc
elliottt: Happy to add that. I think that it's sort of hand-wavingly implied right now by errors being present when trying to define a package named `Test`, but we might as well be explicit here.

resolver/resolver.cc
elliottt: I think that's a good idea. I'll add some more text here.

resolver/resolver.cc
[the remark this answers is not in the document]
elliottt: Currently, defining a package in the `Test::` namespace causes `ENFORCE` failures in the resolver. It might be a little more tricky to add an error than I thought.

PR #5412, opened 2022-03-05
Never use -Os or -Oz to compile the parser

jez:
### Motivation

Without `-Os`, the parser only takes a few seconds to compile.

With, it would take upwards of 4 minutes (though it would "only" take 2
minutes back a few months ago before any of the recent parser changes).

Some raw notes:

> Here is a build from back in december before i started making any
> changes to the parser. at that point, it still took ~2min 45s (it's
> currently about 4min 40s)
>
> <https://buildkite.com/sorbet/sorbet/builds/19253#11b589a1-3847-4aa5-a6d8-10d8012f391a>
>
> the build times are pretty fast for me locally with
> `--config=test-static-sanitized`, but pretty slow with
> `--config=buildfarm-sanitized-linux` the only difference between the
> two is `--config=buildfarm`, which is defined like this:
>
> ```
> build:buildfarm --copt=-Os --linkopt=-Os --config=reduce-intermediate-file-size-base --config=rubyfmt
>
> build:reduce-intermediate-file-size-base --copt=-g0 --linkopt=-g0 # used by buildfarm to have less debug information on disk
> build:reduce-intermediate-file-size-base --strip=always
> build:reduce-intermediate-file-size-base --copt=-fno-standalone-debug --linkopt=-fno-standalone-debug # used by buildfarm to have less debug information on disk
>
> # integrates rubyfmt into sorbet
> build:rubyfmt --define rubyfmt=true --copt=-DRUBYFMT_ENABLED
> ```
>
> The PR that introduced that flags claims that this reduced the size of
> the bazel cache in CI by 4x
>
> <https://github.com/sorbet/sorbet/pull/383/files>
>
> Local `--config=test-static-sanitized` build drops from ~4 minutes to
> build those files down to ~30s when i remove the Os flags specifically

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/BUILD
+++ b/parser/parser/BUILD
@@ -104,7 +104,16 @@ cc_library(
     hdrs = glob(["include/**/*.hh"]),
     copts = [
         "-Wno-unused-const-variable",
-    ],
+    ] + select({
+        # What we really want here is to say "it's never ok to use -Os, because
+        # that will take 2+ minutes to compile." We have to do this to override
+        # the .bazelrc options.
+        "@com_stripe_ruby_typer//tools/config:dbg": ["-O0"],
+        "@com_stripe_ruby_typer//tools/config:opt": ["-O2"],
+        # worse to take too much time to compile in development than to run too
+        # slow in production, so -O2 by default.
+        "//conditions:default": ["-O2"],
+    }),
     includes = [
         "include",
         "include/ruby_parser",


jez:
Compiling those files drops from about 4.5 minutes down to about 10s with this change.

The test-static-sanitized job itself drops from 12 min to 10 min.

jez:
It's not clear to me whether we still want `-Os` in the `--config=buildfarm` config, but I kind of want to punt on that for now, because this seems good enough.

aprocter (approved):

PR #5413, opened 2022-03-05
shrink `Method` in various ways

froydnj:
### Motivation

Smaller data structures, epsilon faster code.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.h
+++ b/core/Symbols.h
@@ -141,6 +141,7 @@ public:
     ClassOrModuleRef owner;
     NameRef name;
     ClassOrModuleRef rebind;
+    Flags flags;
     TypePtr resultType;
     // All `IntrinsicMethod`s in sorbet should be statically-allocated, which is
     // why raw pointers are safe.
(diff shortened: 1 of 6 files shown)


core/Symbols.h
```diff
@@ -138,23 +138,25 @@ class Method final {
...
+    bool hasIntrinsic() const;
+    const IntrinsicMethod *getIntrinsic() const;
+
     ClassOrModuleRef owner;
     NameRef name;
     ClassOrModuleRef rebind;
+    Flags flags;
+    const static uint16_t INVALID_INTRINSIC_OFFSET = 0;
+    const static uint16_t FIRST_VALID_INTRINSIC_OFFSET = 1;
+    uint16_t intrinsicOffset = INVALID_INTRINSIC_OFFSET;
```
jez, line 151: What do you think about adding a comment here like this assumes we will only ever have < 2^16 intrinsic methods which is reasonable

core/Symbols.h
```diff
@@ -138,23 +138,25 @@ class Method final {
...
     TypePtr resultType;
-    // All `IntrinsicMethod`s in sorbet should be statically-allocated, which is
-    // why raw pointers are safe.
-    const IntrinsicMethod *intrinsic = nullptr;
```
jez, line 147: Can you move this comment to the getter?

jez (approved):

PR #5420, opened 2022-03-06
Fix parser error with while do

### LINKED ISSUE #5419: Bug in Sorbet parser: while inside begin/rescue block
Sorbet is complaining about a parse error, telling me unexpected token on "while", but it's a valid token.

#### Input

```ruby
# typed: strict

module External
    class Tester
        extend T::Sig

        sig { void }
        def self.testme

            begin
                puts "do something"
                while (false) do
                    puts "do something"
                end
            rescue => e
            end
        end
    end
end
```

#### Observed output

```
unexpected token "while"
```

#### Expected behavior

This is correct ruby and "while" executes fine by ruby itself.

jez:
Fix parser regression for `while ... do`

What was happening was that that having two rules using `kWHILE` forces
the parser to read a lookahead token to disambiguate which action to
take next.

- In the grammar before introducing the bug, it was unambiguous what to
  do next (immediately reduce the empty rule, running `cond.push` before
  lexing the next token).
- After the bug, it became ambiguous, so the grammar would read a token,
  and it would read `'('`. It just so happens that whenever we read one
  of those tokens, the **lexer** runs an additional `cond.push(false)`
  which it then pops when reading the matching `')'` (effectively saying
  that `do` means "do for block" inside `(...)`, even in a condition as
  long as there are parens around it.)

So the stack was supposed to look like

    ..., true, false

after reading a `(` token, but instead because the paren was getting
read as a lookahead token earlier than before, the stack looked like

    ..., false

after reading the `(` token, and then `true` was pushed later once the
parser decided to reduce the empty rule:

    ..., false, true

So upon reading the `)` token, which would pop the topmost boolean,
`cond.active()` would be `false` instead of `true`, which meant that
`"do"` lexed as `kDO` instead of `kDO_COND`.

- - - -

Some historical debugging context. Here's the good trace and the bad
trace:

```diff
❯ diff -u trace-good.txt trace-bad.txt
--- trace-good.txt      2022-03-07 13:32:18.662231915 -0800
+++ trace-bad.txt       2022-03-07 13:32:41.438250601 -0800
@@ -3,206 +3,182 @@
 Reading a token: Next token is token "while" ([0,5): )
 Shifting token "while" ([0,5): )
 Entering state 10
+Reading a token: Next token is token "(" ([6,7): )
 Reducing stack by rule 63 (line 866):
 driver.lex.cond.push(true)
 -> $$ = nterm $@4 ([0,5): )
 Stack now 10 0
-Entering state 262
-Reading a token: Next token is token "(" ([6,7): )
+Entering state 263
+Next token is token "(" ([6,7): )
 Shifting token "(" ([6,7): )
 Entering state 131
```

(computed by using `--trace-parser`, also I added some print statements
to show where the `cond.push(true)` happens)

We see that in the good trace (`-` lines), the `nterm $@4` reduction
happens before reading the `"("`, but in the bad trace (`+` lines), we
read a lookahead token before deciding to reduce one way or another.

### Motivation

Fixes #5419

- [x] @jez Figure out whether there's a better way to mitigate these sorts of
  lexer statefulness bugs, because this is probably not the only one.

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/cc/grammars/typedruby.ypp
+++ b/parser/parser/cc/grammars/typedruby.ypp
@@ -379,6 +379,8 @@ using namespace std::string_literals;
   k_return
   lcurly_block_start
   lbrace_cmd_block_start
+  until
+  while
 
 %type <delimited_list>
   defn_head
@@ -863,14 +865,24 @@ int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser:
 
       expr_value: expr
 
-   expr_value_do:   {
+
+           // We have to factor these like this because we need these rules to run before lexing any lookahead tokens,
+           // since we have two rules that read `kWHILE` (and `kUNTIL`), and all tokens that we lex after after such
+           // a token must be lexed assuming this side effect has taken place.
+           while: kWHILE
+                    {
                       driver.lex.cond.push(true);
                     }
-                  expr_value do
+           until: kUNTIL
+                    {
+                      driver.lex.cond.push(true);
+                    }
+
+   expr_value_do: expr_value do
                     {
                       driver.lex.cond.pop();
 
-                      $$ = driver.alloc.node_with_token($3, $2);
+                      $$ = driver.alloc.node_with_token($2, $1);
                     }
 
         def_name: fname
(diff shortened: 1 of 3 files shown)


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -863,10 +863,14 @@ int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser:
 
       expr_value: expr
 
-   expr_value_do:   {
+   expr_value_do_begin:
```
froydnj, line 866: WDYT about calling this `expr_value_do_push_cond` or somesuch?  Partly so it's a little easier to see that it needs to match up with `driver.lex.cond.pop()` and partly because suffixing with "begin" makes me think this is actually about `begin`, which is not true.

jez: irrelevant, as i factored it another way (though thanks for the suggestion! i would have taken it had i not found a different way).

jez:
I had originally assumed that the bug was due to not reseting the state when we started throwing away tokens, but in fact the bug was due to us not setting state that we should have been setting.

froydnj (approved):
Nice!

krschacht:
@jez Did you test `do / while` as well as `while / do`, I believe when I was testing both they were both failing. I assume your fix works for both but I just wanted to explicitly call that out.

jez:
Do you have an example of `do/while`? afaik Ruby doesn't have a `do ... while` loop that uses the `do` keyword.

It has `begin ... end; while ...` but that doesn't have a `do` keyword after the end, and so I think it should have worked completely fine before and after my changes.

https://sorbet.run/#%23%20typed%3A%20true%0Abegin%0A%20%20puts%20'once'%0Aend%20while%20%28false%29

krschacht:
Ah, sorry, you're right. Nevermind. :) When my original code started failing because of this bug I tried those two variations, but I mistakenly thought I was trying two valid looping variations.

PR #5421, opened 2022-03-07
Add auto compaction to sorbet

tenderlove:
This adds GC.auto_compact and GC.auto_compact= signatures.

Hi,

This is my first time contributing so I'm not sure what to do exactly.  Do we need to write tests for this?  Also I noticed that other "setter" methods [don't seem to have a signature](https://github.com/sorbet/sorbet/blob/66cf53bcfece2e722d1d7f76fae8176e047f1d68/rbi/core/gc.rbi#L146-L162), so I didn't add a signature to `GC.auto_compact=` but I don't know if that's "right".

### DIFF
--- a/rbi/core/gc.rbi
+++ b/rbi/core/gc.rbi
@@ -50,6 +50,26 @@ module GC
   # Returns information about the most recent garbage collection.
   def self.latest_gc_info(*_); end
 
+  #  Returns whether or not automatic compaction has been enabled.
+  #
+  # ```ruby
+  # GC.auto_compact    #=> true or false
+  # ```
+  sig {returns(T::Boolean)}
+  def self.auto_compact(); end
+
+  #
+  #  Updates automatic compaction mode.
+  #
+  #  When enabled, the compactor will execute on every major collection.
+  #
+  #  Enabling compaction will degrade performance on major collections.
+  #
+  # ```ruby
+  # GC.auto_compact = flag
+  # ```
+  def self.auto_compact=(_); end
+
   # Initiates garbage collection, even if manually disabled.
   #
   # This method is defined with keyword arguments that default to true:


rbi/core/gc.rbi
```diff
@@ -50,6 +50,26 @@ module GC
   # Returns information about the most recent garbage collection.
   def self.latest_gc_info(*_); end
 
+  #  Returns whether or not automatic compaction has been enabled.
```
Morriar, line 53: Was this documentation generated with [`syn-rdoc`](https://github.com/sorbet/sorbet/blob/master/rbi/tools/sync-rdoc.rb) or written manually?

```suggestion
  # Returns whether or not automatic compaction has been enabled.
```

rbi/core/gc.rbi
```diff
@@ -50,6 +50,26 @@ module GC
...
+  sig {returns(T::Boolean)}
+  def self.auto_compact(); end
+
+  #
+  #  Updates automatic compaction mode.
+  #
+  #  When enabled, the compactor will execute on every major collection.
+  #
+  #  Enabling compaction will degrade performance on major collections.
+  #
```
Morriar, line 67:
```suggestion
  # Updates automatic compaction mode.
  #
  # When enabled, the compactor will execute on every major collection.
  #
  # Enabling compaction will degrade performance on major collections.
  #
```

rbi/core/gc.rbi
```diff
@@ -50,6 +50,26 @@ module GC
...
+  # ```ruby
+  # GC.auto_compact = flag
+  # ```
+  def self.auto_compact=(_); end
```
Morriar, line 71: From the documentation and a few tests I ran it's definitely not obvious which types are accepted. It seems the method accepts anything and always return the value of the argument passed. 

I guess it's okay to not have a signature 👍

jez (approved):
https://ruby-doc.org/core-3.0.0/GC.html#method-c-auto_compact

rbi/core/gc.rbi
tenderlove: I copy / pasted this from `gc.rb`, so this could be my copy paste mistake

PR #5422, opened 2022-03-07
fix missing methods generation with the `algolia` gem

froydnj:
### Motivation

`Algolia::Recommend::Model` incorrectly gets classified as something we don't want to prefix with `::`, leading to peculiar build errors when missing methods are regenerated and then Sorbet type-checks the file.  This PR adds a special case similar to what we already have for `"Plaid"` to address this.

### Test plan

See included automated tests.

### DIFF
--- a/main/minimize/minimize.cc
+++ b/main/minimize/minimize.cc
@@ -52,7 +52,7 @@ OutputCategory outputCategoryFromClassName(string_view fullName) {
     } else if (absl::StrContains(fullName, "::Mutator")) {
         return OutputCategory::Mutator;
     } else if (absl::StrContains(fullName, "::Model")) {
-        if (absl::StartsWith(fullName, "Plaid")) {
+        if (absl::StartsWith(fullName, "Plaid") || absl::StartsWith(fullName, "Algolia")) {
             return OutputCategory::External;
         } else {
             return OutputCategory::Model;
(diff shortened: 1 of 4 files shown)


main/minimize/minimize.cc
```diff
@@ -52,7 +52,7 @@ OutputCategory outputCategoryFromClassName(string_view fullName) {
     } else if (absl::StrContains(fullName, "::Mutator")) {
         return OutputCategory::Mutator;
     } else if (absl::StrContains(fullName, "::Model")) {
-        if (absl::StartsWith(fullName, "Plaid")) {
+        if (absl::StartsWith(fullName, "Plaid") || absl::StartsWith(fullName, "Algolia")) {
```
jez, line 55: Do we have any entries that blame to `OutputCategory::Model` right now that also don't start with `Opus::`?

I wonder if we should get ahead of this problem once and for all by adding that constraint.

jez (approved):

main/minimize/minimize.cc
froydnj: :+1: We'd probably have to have `Chalk::` and `Opus::` checks; those look like the only things we'd have to care about.

I'll do that in a followup PR, because I'd like to get this one landed and bumped internally.

PR #5423, opened 2022-03-07
Fix linter failure on master

jez:
### Motivation

Got this wrong on #5411

### DIFF
--- a/.buildkite/linters.sh
+++ b/.buildkite/linters.sh
@@ -45,12 +45,13 @@ if ! ./tools/scripts/format_website.sh -t &> format_website; then
 fi
 
 pushd vscode_extension
+yarn install
 if ! yarn lint --output-file=yarn_lint; then
     globalErr=$?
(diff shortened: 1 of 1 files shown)


.buildkite/linters.sh
```diff
@@ -45,12 +45,13 @@ if ! ./tools/scripts/format_website.sh -t &> format_website; then
 fi
 
 pushd vscode_extension
+yarn install
 if ! yarn lint --output-file=yarn_lint; then
```
froydnj, line 49: What in the world...did `yarn lint` just silently succeed before on the non-master build?!

froydnj (approved):

.buildkite/linters.sh
jez: Yeah. This is not a complete fix, it's just part way there. I'm continuing to investigate, because linters.sh should have failed.

PR #5427, opened 2022-03-08
add timers for various phases of completion requests

froydnj:
### Motivation

We can see in our metrics that completion requests are relatively slow, but we have no insight into where that time is being spent.  This PR attempts to change that by adding timers for parts of responding to a completion request.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -839,6 +839,7 @@ vector<unique_ptr<CompletionItem>> CompletionTask::getCompletionItems(LSPTypeche
 
     vector<core::NameRef> similarLocals;
     if (params.enclosingMethod.exists()) {
+        Timer timeit(gs.tracer(), LSP_COMPLETION_METRICS_PREFIX ".determine_locals");
         auto locals = localNamesForMethod(typechecker, params.enclosingMethod, params.queryLoc);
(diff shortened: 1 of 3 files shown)


main/lsp/requests/completion.cc
```diff
@@ -849,89 +850,105 @@ vector<unique_ptr<CompletionItem>> CompletionTask::getCompletionItems(LSPTypeche
 
     // ----- methods -----
```
aprocter, line 851: Do we want a timer for the `// ----- keywords -----` stuff above this?

aprocter (approved):
Nice!

main/lsp/requests/completion.cc
froydnj: Good question.  I thought about adding it, and then decided that this was pretty static: 30 or so prefix matches on a table of keywords *should* be pretty quick and not worth measuring?

froydnj:

→ https://go/builds/bui_LHkTCJuGtNDB55
→ https://go/builds/bui_LHkTiBQP5UgOSr

main/lsp/requests/completion.cc
aprocter: Works for me!

PR #5428, opened 2022-03-08
Narrow signature for Enumerable#include?

jez:
### Motivation

It's almost always a logic bug to ask `include?` something where Sorbet can know statically that the `Enumerable` collection could not possibly have such an element.

This comes at the slight cost (or maybe benefit honestly?) of rejecting `xs.include?(something_nilable)`, which some people have expressed a desire to use more widely in a typed codebase, but I'm of the opinion that I would rather have the type checker show me the type error and handle `nil` or `unsafe` the argument in those cases.

There were only 14 cases where this caused new type errors on Stripe's codebase, and almost all of them were concentrated in a single file.

It's also worth noting that `member?` (which is an alias for `include?` in Enumerable) and `key?` on `Hash` already have sigs specifying that `K` is required. Also `Set#include?` already had an override of `include?` that narrowed the types to this signature.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/hash.rbi
+++ b/rbi/core/hash.rbi
@@ -757,6 +757,30 @@ class Hash < Object
   end
   def has_value?(arg0); end
 
+  # Returns `true` if the given key is present in *hsh*.
+  #
+  # ```ruby
+  # h = { "a" => 100, "b" => 200 }
+  # h.has_key?("a")   #=> true
+  # h.has_key?("z")   #=> false
+  # ```
+  #
+  # Note that
+  # [`include?`](https://docs.ruby-lang.org/en/2.7.0/Hash.html#method-i-include-3F)
+  # and
+  # [`member?`](https://docs.ruby-lang.org/en/2.7.0/Hash.html#method-i-member-3F)
+  # do not test member equality using `==` as do other Enumerables.
+  #
+  # See also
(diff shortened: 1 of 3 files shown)


jez:

→ https://go/builds/bui_LHkyxEYU3Psjkj
→ https://go/builds/bui_LHkyKBdihUUxs6

jez:

→ https://go/builds/bui_LHlKCnvAOknps1
→ https://go/builds/bui_LHlK1tUJX6N0QB

rbi/core/hash.rbi
```diff
@@ -757,6 +757,30 @@ class Hash < Object
...
+  # [`include?`](https://docs.ruby-lang.org/en/2.7.0/Hash.html#method-i-include-3F)
+  # and
+  # [`member?`](https://docs.ruby-lang.org/en/2.7.0/Hash.html#method-i-member-3F)
+  # do not test member equality using `==` as do other Enumerables.
+  #
+  # See also
+  # [`Enumerable#include?`](https://docs.ruby-lang.org/en/2.7.0/Enumerable.html#method-i-include-3F)
+  sig do
+    params(
+        arg0: K,
```
jez, line 778: we need this override, because `Elem` is set to `[K, V] (2-tuple)` for `Hash` (and ENVClass)

PR #5430, opened 2022-03-09
add a lock to guard `cancellationUndoState` in `LSPTypechecker.cc`

froydnj:
### Motivation

We want to enable some LSP requests to run on stale `GlobalState` data while there is an ongoing typechecking operation.  The hypothesis is that people get frustrated when certain operations randomly become inoperable during typechecking, and having at least some requests available will increase Sorbet's perceived responsiveness.

To this end, then, we need to carefully manage `cancellationUndoState`, as its contents may be used by a request at the point that `cancellationUndoState` needs to be accessed after a successful or preempted typechecking operation.  This PR is the first step in that respect: adding a lock around `cancellationUndoState` and acquiring it when necessary.  Future work will add various operations that would acquire a `ReaderMutexLock` around `cancellationUndoState` before accessing the `GlobalState` within.

At the moment, this should be effectively a no-op change: if we ever did have a situation where `cancellationUndoState` was getting used on multiple threads in a concurrent fashion, this PR would in fact be an improvement, because it would eliminate races.  It's true that we're doing slightly more work now in various places, but uncontended mutex operations are reasonably fast, and I think any cost here would be dwarfed by e.g. typechecking itself.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -108,6 +108,7 @@ bool LSPTypechecker::typecheck(LSPFileUpdates updates, WorkerPool &workers,
     if (updates.canceledSlowPath) {
         // This update canceled the last slow path, so we should have undo state to restore to go to the point _before_
         // that slow path. This should always be the case, but let's not crash release builds.
+        absl::WriterMutexLock writerLock(&this->cancellationUndoStateRWLock);
         ENFORCE(cancellationUndoState != nullptr);
         if (cancellationUndoState != nullptr) {
             // Restore the previous globalState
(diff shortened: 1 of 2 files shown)


main/lsp/LSPTypechecker.cc
```diff
@@ -108,6 +108,7 @@ bool LSPTypechecker::typecheck(LSPFileUpdates updates, WorkerPool &workers,
     if (updates.canceledSlowPath) {
         // This update canceled the last slow path, so we should have undo state to restore to go to the point _before_
         // that slow path. This should always be the case, but let's not crash release builds.
+        absl::WriterMutexLock writerLock(&this->cancellationUndoStateRWLock);
```
aprocter, line 111: Might move this above the comment.

aprocter (approved):
Looks good, thanks! I'm going to be playing with this today from the indexer thread side.

PR #5432, opened 2022-03-09
Prevent invalid signature unwrapping for overridden methods

vinistock:
Add a runtime error if trying to unwrap a signature where the original method was overridden without a signature.

Questions / notes

- We are skipping this check for final methods. This is because a similar check already exists for final methods and the error is emitted elsewhere
- Is there a better way of verifying that the method was overridden by Sorbet other than checking the `source_location`? I found an existing check, which is where I got the regex from, but I wondered if there's a better way of achieving this
- Any unforeseen consequences of adding this check?

### Motivation

If a method with a registered signature is overridden by a method without one, Sorbet will unwrap the first implementation when running signature blocks, which does not match the behaviour of the Ruby runtime. For example,
```ruby
class Foo
  extend T::Sig

  def initialize
    @bar = 1
  end

  # With this declaration, we register `bar` in the `sig_wrappers` and the method gets wrapped from inside Sorbet's _method_added hook
  sig {returns(Integer)}
  def bar
    2
  end

  # Now, we override the override. That is, we create an attr_reader `bar` that will replace Sorbet's method wrapping
  attr_reader :bar
end

# Invoking `bar` directly will not trigger any unwrapping, because we lost the original wrapping after overriding with an attr_reader
# Additionally, this will respect the Ruby runtime and execute the last version defined for `bar`
Foo.new.bar
# => 1

# If we run the signature blocks, then Sorbet will find the signature for the first occurrence of `bar` and unwrap that method on top of the attr_reader
T::Utils.run_all_sig_blocks
# => 2
```

### Test plan

See included automated tests.


vinistock:
I'm surprised that we have a compiler test specifically testing this scenario and mentioning that it is allowed. Any context on why?

froydnj:
> I'm surprised that we have a compiler test specifically testing this scenario and mentioning that it is allowed. Any context on why?

There is some context here:

https://github.com/sorbet/sorbet/blob/9284313547e7328070e32de2b46f7a8ef1bc1b7c/compiler/IREmitter/Payload/vm-payload.c#L405-L450

IIRC, the original intent was to make sure the compiler worked identically to interpreted code; since the interpreted code permitted multiple definitions (with signatures) in some cases, the compiler needed to too.  If that's no longer true in certain (all?) cases, we should make the compiler testcase more closely reflect the current state of affairs (e.g. deleting things that would make the interpreter break...or better, trying to `rescue` the errors and ensuring that the compiler and interpreter generate errors identically).

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -341,6 +341,15 @@ def self.run_sig(hook_mod, method_name, original_method, declaration_block)
         Signature.new_untyped(method: original_method)
       end
 
+    current_method = signature.method.owner.instance_method(method_name)
+    file_path, line = current_method.source_location
+
+    if !@was_ever_final_names[method_name] && original_method != current_method &&
```
paracycle, line 347: I'd argue that this logic should be encapsulated inside the `unwrap_method` call, since that would make sure regardless of how `unwrap_method` is called, this is always checked.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -341,6 +341,15 @@ def self.run_sig(hook_mod, method_name, original_method, declaration_block)
...
+       !file_path.match?(%r{sorbet-runtime[^/]*/lib/})
```
paracycle, line 348: The `file_path` for the wrapper method should be exactly equal to `__FILE__`, since this is the same file that creates the wrapper in the first place:
```suggestion
       file_path != __FILE__
```

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -341,6 +341,15 @@ def self.run_sig(hook_mod, method_name, original_method, declaration_block)
...
+      raise "The method `#{method_name}` on #{hook_mod} was overwritten before unwrapping signatures " \
```
paracycle, line 349: This doesn't have to be an error case, and IMO if it is going to be an error, it should be configurable.

If the Sorbet wrapper is overwritten, then that is not necessarily an error. We just don't want Sorbet to do the unwrapping in that case, since that would result in broken code. As far as the app is concerned, it has the correct method in place and it is working as intended at this point.

For that reason, I'd much prefer it if we skip unwrapping in this case instead of erroring. We could have a special mode where this gets turned into an exception, but that would have to be opt-in, IMO.

paracycle:
> I'm surprised that we have a compiler test specifically testing this scenario and mentioning that it is allowed. Any context on why?

As per my comments above, I think it should be allowed, since this is a legitimate thing to do, but the unwrapping should not do the wrong thing in this situation and replace a non-Sorbet owned method with the original method.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
vinistock: I agree. As long as Sorbet does the same as Ruby and respects the latest definition of a method, we don't necessarily need to raise. I changed the tests and code to reflect that.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
paracycle: What if `T::Private::Methods` had a constant `WRAPPER_METHOD_SOURCE_FILE = __FILE__` that this referenced as:
```ruby
  file_path != T::Private::Methods::WRAPPER_METHOD_SOURCE_FILE
```
Then you wouldn't have to keep doing regexp checks and still be safe against the code moving.

vinistock: We can do that too. I don't have strong opinions about this one. @froydnj, what do you think?

vinistock:
Okay, adapted the code to not raise. It now keeps the latest definition always, exactly as Ruby, by not unwrapping the method if it has been overridden. I also kept the signature, so that the behaviour matches the expected by the compiler.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
vinistock: That sounds good to me. Added the comment and the new constant.

froydnj (changes requested):
Just a few tests to add, because I'm not sure what happens with them.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -399,7 +401,16 @@ def self.build_sig(hook_mod, method_name, original_method, current_declaration,
...
-    maybe_wrapped_method = CallValidation.wrap_method_if_needed(mod, signature, original_method)
+    method_name = original_method.name
+    maybe_wrapped_method = signature.method.owner.instance_method(method_name)
+    file_path, _line = maybe_wrapped_method.source_location
+
+    # If Sorbet's wrapper was overridden, we don't want to try to unwrap the registered signature
+    unless !@was_ever_final_names[method_name] && original_method != maybe_wrapped_method &&
+           file_path != WRAPPER_METHOD_SOURCE_FILE
+      maybe_wrapped_method = CallValidation.wrap_method_if_needed(mod, signature, original_method)
+    end
```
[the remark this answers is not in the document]
vinistock: I believe the first wrap always happens in `_methods.rb` inside `_on_method_added` and the one in call validation only happens when unwrapping.

I added two extra tests for abstract so that we can be sure. If overriding an abstract method, we don't unwrap, which is what we'd expect. And in the other way, when overriding a method with an abstract one, then we're still unwrapping and going up the hierarchy chain.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -399,7 +401,16 @@ def self.build_sig(hook_mod, method_name, original_method, current_declaration,
   end
 
   def self.unwrap_method(mod, signature, original_method)
-    maybe_wrapped_method = CallValidation.wrap_method_if_needed(mod, signature, original_method)
+    method_name = original_method.name
+    maybe_wrapped_method = signature.method.owner.instance_method(method_name)
+    file_path, _line = maybe_wrapped_method.source_location
```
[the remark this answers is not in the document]
vinistock: It should always exist unless the method was somehow undefined between registering the signatures and running signature blocks. Also, `instance_method` raises `NameError` if the method doesn't exist.

froydnj:
How did cc26aa1 fix the test failures in 45ad3a5?  Are those tests just flaky in some way?

vinistock:
Yes, they were flaky with the previous implementation and would pass roughly 50% of the time on my machine. Not entirely sure why the number of allocations changed between different runs, but avoiding the string allocation from `method.name` and just reading the method name from the `signature` did the trick.

Also, I thought about changing those assertions to be something like `assert_operator(allocations, :<, 5)`, which provides a nicer failure message, but didn't want to put it in this PR. I can create a follow up for that.

froydnj:
The above build links are for sorbet builds, not sorbet-runtime (we trigger a different set of builds for sorbet-runtime, but those are not posted to github).

The Stripe-internal build link for the sorbet-runtime build is https://go/builds/bui_LKdz5DnNGYLiJY

Unfortunately I'm running into problems with our testing harness, which makes the above build fail.

froydnj:

→ https://go/builds/bui_LVvdp0yDwseMwq
→ https://go/builds/bui_LVvdzCDu7aLq2Z

froydnj:

→ https://go/builds/bui_LWzbty50c5xf4F
→ https://go/builds/bui_LWzbo3r9Lw5V4G

froydnj:

→ https://go/builds/bui_LX2VjXIjuyfIVD
→ https://go/builds/bui_LX2VMHE4sHNVS0

froydnj:
Thanks for your patience on this; we were running into issues where testing this PR would uncover issues with how Stripe's test harness worked, and that left us unable to get actual results until today.

I'm seeing the following error across a couple different testcases:

```
sorbet-runtime/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:409:in `unwrap_method': undefined method `first' for nil:NilClass (NoMethodError)
	from sorbet-runtime/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:346:in `run_sig'
	from sorbet-runtime/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:243:in `block in _on_method_added'
	from sorbet-runtime/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:450:in `run_sig_block_for_key'
	from sorbet-runtime/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:467:in `block in run_all_sig_blocks'
	from sorbet-runtime/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:464:in `loop'
	from sorbet-runtime/gems/sorbet-runtime/lib/types/private/methods/_methods.rb:464:in `run_all_sig_blocks'
	from sorbet-runtime/gems/sorbet-runtime/lib/types/utils.rb:71:in `run_all_sig_blocks'
...
```

I guess it's not guaranteed that everything has a source location (e.g. C functions of some kind)?

I'm not 100% sure how you would write a test for this; I guess you'd have to make sure you aliased a method to something that you knew was going to be implemented in C?

vinistock:
@froydnj no worries at all! You're right, `source_location` is `nil` for methods defined in C. I was able to reproduce the problem with the test case in the latest commit.

Basically, we're taking the original C implementation, overriding it with a method with signature and then overriding it back with the C implementation. This reproduces the issue of overriding an existing method with signature using native extensions.

For the solution, we basically do not want to try to unwrap if the `source_location` is `nil`. Instead of explicitly adding an `or` to the conditional, I just used safe navigation. Let me know which style you prefer!

froydnj:

→ https://go/builds/bui_LXO5Ap28qZL5qq
→ https://go/builds/bui_LXO5SdrPg5aAXh

froydnj:
There's one last failing test in Stripe's codebase: https://go/builds/bui_LXOUaHLgDE8H16

I think this test intersects in weird ways with the problems this PR is trying to solve, so it's very possible that we should delete this test now.  I will take a closer look at it.

froydnj:
OK, I took a look at the test failure.  We have an internal tool (similar to `lib/types/private/class_utils.rb`) for redefining methods that also takes an argument that specifies what to do for sig blocks:

1. Evaluate them eagerly, prior to replacement.
2. Ignore their presence and hope things work.

I think the argument is trying to address something similar to what you are doing here.

We have a test that:

1. Defines a class containing a method with a sig
2. Replaces that method, ignoring the signature, with a block that does the moral equivalent of:

```
p "patched!"
original_method.bind_call(self)
```

Once that's in place, we:

1. Instantiate the class.
2. Call the method.  We expect to see the patched method called, then the original.
3. Call the method again.  We expect to see only the original method called.

This PR changes the behavior such that we see the patched method called, then the original, in both cases.  I think that is what this PR is trying to achieve, but it's not compatible with the behavior of the test that we have.

(We have another test that does the same thing, but we eagerly evaluate the signature, rather than ignoring it.  Then we run the same "instantiate class, call method twice", and both times we expect to see the patched method called, followed by the original method.)

Does that provide enough context to fix the issues we are seeing?  Or is that test case completely incompatible with the goals of this PR?

vinistock:
Indeed it is incompatible with this PR. That use case is exactly what this PR prevents and is analogous to the first example in the new test added. In these scenarios, when a method that has a signature is patched by another without one, the patch is lost after signature unwrapping. This doesn't match the behaviour of Ruby, which always honours whatever the latest definition of the method is.

Extracted from the test in this PR
```ruby
Class.new do
  extend T::Sig

  def initialize
    @double_definition = "OK"
  end

  # Define the original method with a signature
  # The signature will store a reference to this implementation to unwrap it
  sig {returns(String)}
  def double_definition
    "FAIL"
  end

  # Override the implementation of the same method, but not including a signature
  # Here we're using an attr_reader, but this could be any other re-definition, like define_method
  attr_reader :double_definition
end

# Without this PR's fix, when the signature for `double_definition` is unwrapped, it will override the latest
# implementation of `double_definition` (the one added by the attr_reader) with the original one
```

paracycle:
@froydnj I just got pinged that we experienced a small incident due to this behaviour, so I'd like to see how we can move this forward.

The incident was caused by something like this:

```ruby
class Foo
  sig { void }
  def generated_method
    super
  end

  generate_a_method_named :generated_method
end
```
since we eagerly run all `sig` blocks in production, what had happened, in sequence was:

1. Sorbet runtime processed the `sig` and `generated_method` definition (let's call this `Method Def #1`), and replaced the method with a sig block evaluation wrapper (let's call this `Method Def #2)`.
2. The class generated a method named `generated_method` dynamically on the `Foo` class (let's call this `Method Def #3`). That dynamic method ended up replacing `Method Def #2`.
3. After we boot the application, we call `T::Utils.run_all_sig_blocks`, which triggered `Method Def #2`.
4. `Method Def #2` evaluated the sig block and went on to `unwrap` the method. Unwrapping it replace the method with what it remembered the method was from when the signature was defined, which was `Method Def #1`.
5. Now, when `generated_method` is called, it ended up calling `Method Def #1`, which called `super`, but there is no `super` method in existence. Thus, an incident in production.

I realize that the call to `super` in this exact case wasn't correct to begin with, but the actual implementation of the method could have gone wrong in other ways, as well. If the method definition from the dynamically generated method (i.e. `Method Def #3`) was kept in place and not replaced, then we would not have experienced any problems, though. This is the case that this PR was exactly trying to prevent.

As to the failing test in the Stripe codebase, my assumption is that the test case was written against the actual behaviour of Sorbet runtime and not the expected behaviour of it. If no other tests are failing in your codebase, I would assume that your codebase is not expecting that behaviour at all. However, obviously, you know your codebase better than us. What I can offer, if that's agreeable, is to add a configuration option to switch this behaviour on/off, and you can keep it in the old style as long as you want, if you want to preserve the existing behaviour.

Regardless, I would really like it if we can move this PR to the finish line, since I would not want to experience another incident on our system similar to this where Sorbet runtime was involved.

PR #5448, opened 2022-03-10
Add command line flag for experimental stale state features

aprocter:
Adds a `--enable-experimental-lsp-stale-state` flag to gate some features we're going to be working on for serving LSP requests from "stale" GlobalState while a slow path is running.

### Motivation
Want to turn this feature off by default since it will certainly be broken at first.

### Test plan
Didn't add any tests for setting the flag per se, but the code follows the pattern of existing flags for experimental LSP features so I figure that's probably okay.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -323,6 +323,9 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                     "Enable experimental LSP feature: Document Highlight");
     options.add_options("advanced")("enable-experimental-lsp-signature-help",
                                     "Enable experimental LSP feature: Signature Help");
+    options.add_options("advanced")("enable-experimental-lsp-stale-state", "Enable experimental LSP feature: fast "
+                                                                           "but approximate answers from stale "
+                                                                           "typechecker state");
     options.add_options("advanced")("enable-experimental-requires-ancestor",
                                     "Enable experimental `requires_ancestor` annotation");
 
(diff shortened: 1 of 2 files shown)


jez (approved):

main/options/options.cc
```diff
@@ -722,6 +725,7 @@ void readOptions(Options &opts,
         opts.lspSignatureHelpEnabled = enableAllLSPFeatures || raw["enable-experimental-lsp-signature-help"].as<bool>();
         opts.lspDocumentFormatRubyfmtEnabled =
             enableAllLSPFeatures || raw["enable-experimental-lsp-document-formatting-rubyfmt"].as<bool>();
+        opts.lspStaleStateEnabled = enableAllLSPFeatures || raw["enable-experimental-lsp-stale-state"].as<bool>();
```
froydnj, line 728: I have minor FUD about this, as I believe some developers at Stripe do run with all experimental features enabled as a matter of course?  I guess we can see if they become super-crashy if we start making use of this flag and can come up with a different implementation.

froydnj (approved):

main/options/options.cc
aprocter: Ah, thanks for calling that out! Although the flag is currently a no-op, I actually think we're going to want to reserve the right to keep this thing fairly crashy and stubby for at least a short while. For example, I'd like to start landing some of the threading-related stuff before namer and resolver updates are likely to be ready. So if folks are running with `enableAllLSPFeatures` in the real world, we probably _don't_ want to turn it on for them yet. So I propose just deleting `enableAllLSPFeatures || ` from the rhs here, meaning we'd require opt-in to this specific feature explicitly. Thoughts?

froydnj: I'm fine with just enabling this with the explicit, specific command-line flag so testing at Stripe is easier; we can subsume it under the all-features flag at a later point.

PR #5449, opened 2022-03-10
Convert LSPTypecheckerDelegate to an interface

aprocter:
This adds an interface class for the methods of `LSPTypecheckerDelegate`, with all methods marked virtual.

### Motivation
For now this is a functional no-op, but later on when serving requests from stale GlobalState, we'll need to expose somewhat different implementations of these methods.

It's worth noting that I'm not sure all the methods will be implementable from stale GlobalState. I think starting from that _assumption_ is probably the best thing to do, since every existing LSPTask depends on the interface exposed by `LSPTypecheckerDelegate`, but we may well want to tweak/refactor the interface down the line as we discover restrictions/limitations/caveats.

### Test plan
I think this should be an "if it compiles and the existing tests pass, everything is fine" situation.

### DIFF
--- a/main/lsp/LSPTask.cc
+++ b/main/lsp/LSPTask.cc
@@ -204,7 +204,7 @@ void LSPTask::index(LSPIndexer &indexer) {}
 LSPRequestTask::LSPRequestTask(const LSPConfiguration &config, MessageId id, LSPMethod method)
     : LSPTask(config, method), id(move(id)) {}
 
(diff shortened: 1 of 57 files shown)


jez (approved):

elliottt (approved):
👍

aprocter-stripe:

→ https://go/builds/bui_LIAH6PrHaN3d0M
→ https://go/builds/bui_LIAHtkg4Olcb7d

PR #5453, opened 2022-03-10
Process --extra-package-... flags for rbi-generation

elliottt:
The `--extra-package-files-directory-prefix` flag gives additional context to the packageDB that we need to associate rbis and generated files with a package for single-package rbi generation. This PR removes the restriction that it only be provided in `--stripe-packages` mode, and uses it to setup the global state before initializing the packageDB when generating rbis.

### Motivation

Stabilizing rbi generation.

### Test plan

See included automated tests.


nroman-stripe (approved):
Makes sense

froydnj (approved):

PR #5454, opened 2022-03-10
[LSP] Refactor code action tests

ilyailya:
The change replaces the `exhaustive-apply-code-action` assertion with the `selective-apply-code-action`. Previously, for the code action test you had to mark it with `exhaustive-apply-code-action` and it would assert all possible code actions are applied. Now you need to mark test with `selective-apply-code-action` and list all code actions you're looking for.

I've also added the `ApplyCodeActionAssertion::checkAll` method. It applies all edits in the code action and asserts the result after.

### Motivation

I need this refactoring for the extract method refactoring PR https://github.com/sorbet/sorbet/pull/5309

### Test plan

See included automated tests.

### DIFF
--- a/test/helpers/position_assertions.cc
+++ b/test/helpers/position_assertions.cc
@@ -40,7 +40,7 @@ const UnorderedMap<
         {"enable-packager", BooleanPropertyAssertion::make},
         {"enable-experimental-requires-ancestor", BooleanPropertyAssertion::make},
         {"enable-suggest-unsafe", BooleanPropertyAssertion::make},
-        {"exhaustive-apply-code-action", BooleanPropertyAssertion::make},
+        {"selective-apply-code-action", SelectiveApplyCodeActionAssertions::make},
         {"assert-fast-path", FastPathAssertion::make},
         {"assert-slow-path", BooleanPropertyAssertion::make},
         {"hover", HoverAssertion::make},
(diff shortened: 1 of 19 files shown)


test/helpers/position_assertions.cc
```diff
@@ -1923,4 +1979,47 @@ string ShowSymbolAssertion::toString() const {
...
+    transform(values.begin(), values.end(), values.begin(), [](auto val) { return trimString(val); });
+
+    return make_shared<SelectiveApplyCodeActionAssertions>(filename, range, assertionLine, values, assertionType);
+}
+
+SelectiveApplyCodeActionAssertions::SelectiveApplyCodeActionAssertions(std::string_view filename,
+                                                                       std::unique_ptr<Range> &range, int assertionLine,
+                                                                       std::vector<std::string> values,
+                                                                       std::string_view assertionType)
+    : RangeAssertion(filename, range, assertionLine), assertionType(string(assertionType)), values(values){};
```
froydnj, line 2007:
```suggestion
    : RangeAssertion(filename, range, assertionLine), assertionType(assertionType), values(values){};
```

test/helpers/position_assertions.cc
```diff
@@ -40,7 +40,7 @@ const UnorderedMap<
         {"enable-packager", BooleanPropertyAssertion::make},
         {"enable-experimental-requires-ancestor", BooleanPropertyAssertion::make},
         {"enable-suggest-unsafe", BooleanPropertyAssertion::make},
-        {"exhaustive-apply-code-action", BooleanPropertyAssertion::make},
+        {"selective-apply-code-action", SelectiveApplyCodeActionAssertions::make},
```
froydnj, line 43: Technically, I guess this should be `SelectiveApplyCodeActionPropertyAssertion`, in trying to keep with `*PropertyAssertion` as sigils/directives at the beginning of the file?  (I don't know why they're called "PropertyAssertions", because they're not asserting anything, but that is something to fixup for another day.

Correspondingly, I guess `apply-code-action` should really become `apply-quickfix`, for consistency with the changes made in this PR?  (And `ApplyCodeActionAssertion` should become `ApplyQuickfixAssertion`?)

aprocter (changes requested):
This PR is a bit tough to review as is, because it moves a lot of code around but (I think) is also changing some of that code as it's moved.

If possible, would you mind re-stacking this PR into multiple commits? In particular, it looks like a lot of stuff is being moved into helpers, so maybe that could be a separate commit, followed by changes that implement the new functionality. Alternatively, if you could summarize what is being moved as-is and what is actually new code, that would help.

froydnj (approved):
I have questions below, but I don't think I have anything blocking the PR.  I checked out the branch and verified that the tests modified do fail if the original files or the rbedited files are altered, so I think we can have confidence that moving all the code around is correct.

aprocter (approved):
Ah, I just saw Nathan's comment here:

> I checked out the branch and verified that the tests modified do fail if the original files or the rbedited files are altered, so I think we can have confidence that moving all the code around is correct.

I think I'm satisfied that the code movement is probably fine, in that case, so I wouldn't want to block.

test/helpers/position_assertions.cc
ilyailya: >Correspondingly, I guess apply-code-action should really become apply-quickfix, for consistency with the changes made in this PR? (And ApplyCodeActionAssertion should become ApplyQuickfixAssertion?)

I don't think it would mirror the LSP behaviour correctly. Quickfixes are a subset of code actions. For example, the `refactor.extract` action is a code action, but not a quickfix, and it still uses the `ApplyCodeActionAssertion `.

Issue #5434, opened 2022-03-09
Recover from `@` on its own


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/?arg=--print=parse-tree-whitequark#%23%20typed%3A%20true%0Adef%20foo%0A%20%20puts%20'before'%0A%20%20%40%0A%20%20puts%20'after'%0Aend)

```ruby
# typed: true
def foo
  puts 'before'
  @
  puts 'after'
end
```

#### Observed output

```
s(:def, :foo, nil,
  s(:send, nil, :puts,
    s(:str, "after")))
editor.rb:4: unexpected `@` https://srb.help/2001
     4 |  @
          ^

editor.rb:4: unexpected token tNL https://srb.help/2001
     4 |  @
     5 |  puts 'after'
Errors: 2
```

#### Expected behavior

Sorbet should at least parse the before/after lines and have an `ErrorNode` for the `@` line. Even better would be to make it look something like what it looks like for `x.` where there is a name like `:<method-name-missing>:`.

- - -

PR #5458, opened 2022-03-10
Fix completion for `@` and `@@` without ident

jez:
### Motivation

Fixes <https://github.com/sorbet/sorbet/issues/5434>

### Test plan

See included automated tests.


jez:
cc @froydnj

parser/parser/cc/lexer.rl
```diff
@@ -1334,6 +1334,7 @@ void lexer::set_state_expr_value() {
   # Interpolations with immediate variable names simply call into
   # the corresponding machine.
 
+  # TODO(jez) What is this for?
```
froydnj, line 1337: I assumed this was for string interpolation, though why we would need a completely separate rule for it is not obvious to me.

froydnj (commented):
:tada:

parser/parser/cc/lexer.rl
jez: luckily we had a test for this, so i fixed it.

froydnj (approved):

PR #5459, opened 2022-03-11
No-op change to test assertion

jez:
### Motivation

Noticed this while working on something else

### Test plan

See included automated tests.


froydnj (approved):

jez:
Doesn't actually work this way whoops. the `INFO` is only printed when there is a failure according to the docs.

PR #5460, opened 2022-03-11
don't return completion results for untyped receivers

### LINKED ISSUE #5147: Completion results are not empty for untyped receiver
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28x%3A%20T.untyped%29.void%7D%0Adef%20foo%28x%29%0A%20%20x.%0A%20%20%23%20%5E%20completion%3A%20%28nothing%29%0Aend)

```ruby
# typed: true
extend T::Sig

def foo(x)
  x.
  # ^ completion: (nothing)
end
```

#### Observed output

```
editor.rb:8: unexpected token "end" https://srb.help/2001
     8 |end
        ^^^
Errors: 1
```

Screen Shot 2022-01-24 at 10 50 09 PM

#### Expected behavior

There should be no completion suggestions.

- - -

froydnj:
Fixes #5147.

### Motivation

It's not actually obvious to me why we want this; it seems better to give a best-effort result.  I guess "best-effort result" in this context, though, is "everything", and that's not particularly helpful either.

This change does make it more obvious how we might insert a "hey, we couldn't find any completion results here" message for people, so they might be able to go and fix their code?

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -204,6 +204,10 @@ SimilarMethodsByName similarMethodsForReceiver(const core::GlobalState &gs, cons
 // Walk a core::DispatchResult to find methods similar to `prefix` on any of its DispatchComponents' receivers.
 SimilarMethodsByName allSimilarMethods(const core::GlobalState &gs, core::DispatchResult &dispatchResult,
                                        string_view prefix) {
+    if (dispatchResult.main.receiver.isUntyped()) {
+        return SimilarMethodsByName{};
+    }
+
     auto result = similarMethodsForReceiver(gs, dispatchResult.main.receiver, prefix);
 
     for (auto &[methodName, similarMethods] : result) {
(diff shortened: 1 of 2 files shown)


main/lsp/requests/completion.cc
```diff
@@ -1115,6 +1115,11 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerInterface &
     auto resp = move(queryResponses[0]);
 
     if (auto sendResp = resp->isSend()) {
+        if (sendResp->dispatchResult->main.receiver.isUntyped()) {
+            response->result = std::move(emptyResult);
+            return response;
+        }
```
jez, line 1121: Can we push this into similarMethodsForClass?

jez:
> It's not actually obvious to me why we want this; it seems better to give a best-effort result.

Eventually I want to return a message like "no results because the receiver is untyped"

jez:
> It's not actually obvious to me why we want this

It's worth noting that TypeScript also does not offer any completion results when the receiver is untyped.

main/lsp/requests/completion.cc
froydnj: I moved this into `allSimilarMethods`, which seemed like a better choice.

jez (approved):

PR #5461, opened 2022-03-11
be smarter about completing on `self.`

### LINKED ISSUE #5157: Do not suggest keyword completion on `self`
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20def%20foo%0A%20%20%20%20self.y%0A%20%20%20%20%23%20%20%20%20%20%5E%20completion%3A%20yield_self%0A%20%20end%0Aend)

```ruby
# typed: true

class A
  def foo
    self.y
    #     ^ completion: yield_self
  end
end
```

#### Observed output

```
editor.rb:5: Method y does not exist on A https://srb.help/7003
     5 |    self.y
                 ^
Errors: 1
```

Screen Shot 2022-01-25 at 10 42 17 AM

#### Expected behavior

It doesn't make sense to suggest "all keywords" on `self`.

If there happens to be a method that shares the name with a keyword, we should suggest those, but we shouldn't additionally suggest the keyword completion item + snippet.

- - -

### LINKED ISSUE #5158: Do not suggest local variable completion on `self`
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Adef%20foo%28xyz%29%0A%20%20self.%0A%20%20%23%20%20%20%20%5E%20completion%3A%20CSV%2C%20...%0Aend)

```ruby
# typed: true

def foo(xyz)
  self.
  #    ^ completion: CSV, ...
end
```

#### Observed output

```
editor.rb:6: unexpected token "end" https://srb.help/2001
     6 |end
        ^^^
Errors: 1
```

Screen Shot 2022-01-25 at 10 52 10 AM

#### Expected behavior

We shouldn't allow `self.xyz` as a completion result, because `xyz` is a local variable, not a method.

- - -

This shares the same root cause with #5157 , but could possibly  end up having a different solution.

froydnj:
Fixes #5158.
Fixes #5157.

### Motivation

Better completion results.  I started out fixing #5158, but figured that #5157 was basically one line of code after that, so we might as well roll the fixes together.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/completion/self_receiver.rb
+++ b/test/testdata/lsp/completion/self_receiver.rb
@@ -4,3 +4,10 @@ def foo(xyz)
   self.
   #    ^ completion: CSV, ...
 end # error: unexpected token
+
+class A
+  def foo
+    self.y # error: does not exist
+    #     ^ completion: yield_self, ...
+  end
+end
--- a/test/testdata/lsp/completion/self_receiver.rb
+++ b/test/testdata/lsp/completion/self_receiver.rb
@@ -1,0 +1,6 @@
+# typed: true
+
+def foo(xyz)
+  self.
+  #    ^ completion: CSV, ...
+end # error: unexpected token
--- a/test/testdata/lsp/completion/self_receiver.rb
+++ b/test/testdata/lsp/completion/self_receiver.rb
@@ -1,9 +1,11 @@
 # typed: true
 
-def foo(xyz)
-  self.
-  #    ^ completion: CSV, ...
-end # error: unexpected token
+module
+  def foo(xyz)
+    self.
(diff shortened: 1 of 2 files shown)


test/testdata/lsp/completion/self_receiver.rb
```diff
@@ -0,0 +1,13 @@
+# typed: true
+
+def foo(xyz)
+  self.
```
jez, line 4: Do you want to put this method inside a module, so that the list of available completion items is less?

jez (approved):

PR #5464, opened 2022-03-11
Initialize enabled on first launch

vinistock:
### Implementation

Initialize the `enabled` value based on whether the file `sorbet/config` exists only on first launch for a workspace.

Basically, since we know that `defaultValue` in the cached configuration is always `false`, we ignore it and check all of the other possible cached configurations (workspace, workspaceFolder, etc..). If they are all `undefined` and we haven't saved anything into the workspace state yet, then this is the first launch of Sorbet on that workspace and we can initialize the value of `enabled`.

I also fixed the path for the test folder in our launch configuration, so that we can run tests from inside VS Code.

### Motivation

Based on the [docs](https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration) and after inspecting in the debugger, the `defaultValue` stored in the cached configuration for `enabled` will always be `false` and cannot be `undefined`.

This means that when we invoke
```typescript
this._cachedSorbetConfiguration.get(section, defaultValue);
```
The `defaultValue` is always ignored, since `get` only takes it into account if the value of the setting is `undefined`. If you add a breakpoint here and test it out, you'll see that this invocation for the `enabled` flag always returns `false` no matter what the `defaultValue` is.

This means that the work to automatically enable the extension based on the `sorbet/config` file merged in #5345 does not do anything, since `defaultValue` is ignored.

### Test plan

Existing tests should cover it.

### DIFF
--- a/vscode_extension/.vscode/launch.json
+++ b/vscode_extension/.vscode/launch.json
@@ -21,10 +21,10 @@
             "runtimeExecutable": "${execPath}",
             "args": [
                 "--extensionDevelopmentPath=${workspaceFolder}",
-                "--extensionTestsPath=${workspaceFolder}/out/test"
+                "--extensionTestsPath=${workspaceFolder}/out/src/test"
             ],
             "outFiles": [
-                "${workspaceFolder}/out/test/**/*.js"
+                "${workspaceFolder}/out/src/test/**/*.js"
             ],
             "preLaunchTask": "watch"
         }
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -4,7 +4,7 @@
   "description": "Ruby IDE features, powered by Sorbet.",
   "author": "Stripe Inc.",
   "license": "Apache-2.0",
-  "version": "0.3.11",
+  "version": "0.3.12",
   "publisher": "sorbet",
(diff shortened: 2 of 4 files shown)


jez:
> Existing tests should cover it

Sounds like it was not working before, and this fixes it, so in fact I would claim that there are no tests, not that existing tests cover it 😅 

A better test plan for a similar class of problem in the future is "i found <list of steps to reproduce the problem> and verified that the problem no longer reproduces with this change"

jez (approved):

jez:
@vinistock Remember that the change will not go live to the marketplace until you change the version in the package.json file. If you'd like me to merge without a version bump please lmk, otherwise I will merge after that change.

vinistock:
@jez that's fair 😅. I updated the test's fake storage to also read the defaults from `package.json` and to have a similar behaviour as the one we see in the code.

I had to update one test, because the scenario was actually not possible in the actual extension. In our code, if there's no selected LSP config, we always default to the first option as long as LSP configs exist - but they always exist because VS Code reads them from the defaults in `package.json`, so there's no way they'll ever be `undefined`.

I also bumped the version.

PR #5465, opened 2022-03-11
[LSP] Infrastructure for the "workspace/executeCommand" request

ilyailya:
The change scaffolds the infrastructure for the `workspace/executeCommand` request

### Motivation

The `workspace/executeCommand` will be used in the extract method refactoring

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/tools/make_lsp_types.cc
+++ b/main/lsp/tools/make_lsp_types.cc
@@ -61,6 +61,12 @@ void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_pt
     shared_ptr<JSONType> JSONInt = make_shared<JSONIntType>();
     shared_ptr<JSONType> JSONDouble = make_shared<JSONDoubleType>();
     shared_ptr<JSONType> JSONString = make_shared<JSONStringType>();
+    // In the LSP specification JSONAny defined as
+    // `export type LSPAny = LSPObject | LSPArray | string | integer | uinteger | decimal | boolean | null;`
+    // https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#number
+    //
+    // We don't support LSPObject and LSPArray options yet
+    shared_ptr<JSONType> JSONAny = makeVariant({JSONNull, JSONBool, JSONInt, JSONDouble, JSONString});
 
     // Converted from https://microsoft.github.io/language-server-protocol/specification
     // Last updated on 11/14/18.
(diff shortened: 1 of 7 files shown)


main/lsp/tools/make_lsp_types.cc
```diff
@@ -151,9 +157,10 @@ void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_pt
 
     auto Command = makeObject("Command",
                               {
-                                  makeField("title", JSONString), makeField("command", JSONString),
+                                  makeField("title", JSONString),
+                                  makeField("command", JSONString),
                                   // Unused in Sorbet.
```
jez, line 162: Can we delete this?

main/lsp/tools/make_lsp_types.cc
```diff
@@ -61,6 +61,12 @@ void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_pt
     shared_ptr<JSONType> JSONInt = make_shared<JSONIntType>();
     shared_ptr<JSONType> JSONDouble = make_shared<JSONDoubleType>();
     shared_ptr<JSONType> JSONString = make_shared<JSONStringType>();
+    // In the LSP specification JSONAny defined as
+    // `export type LSPAny = LSPObject | LSPArray | string | integer | uinteger | decimal | boolean | null;`
+    // https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#number
+    //
+    // We don't support LSPObject and LSPArray options yet
```
jez, line 68: This seems bad. Doesn't this mean that i'll fail at runtime if the client sends back an object? That seems like a pretty common case, doesn't it?

jez (commented):
Thanks for attempting to make the diff smaller, but it's not very easy to review this because there are no tests and I can't tell how it will be used in context.

Do you maybe want to just fold this into the PR that requires it, or stack one PR on top of the other?

main/lsp/tools/make_lsp_types.cc
ilyailya: Yes, that's correct. But we control the client, which invokes the commands. The extension also controls which types passed as parameters. So the crashes could potentially happen, but in practice that should never happen.

Proper solution would be to redo the `makeVariant` to support multiple non-trivial types. However, I don't know how to do that right now, it may require some head scratching. Potentially that could require remaking how do we generate LSP types in general.

PR #5469, opened 2022-03-14
Implement hover and (stubbed) codeAction on stale state

aprocter:
### Motivation
This implements a (beta, or let's even call it alpha) feature that allows `hover` queries to proceed even when slow-path typechecking is in progress. You have to opt into the feature with `--enable-experimental-lsp-stale-state`.

### Test plan
See included automated tests.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -770,7 +770,7 @@ std::vector<std::unique_ptr<core::Error>> LSPStaleTypechecker::retypecheck(std::
 
 LSPQueryResult LSPStaleTypechecker::query(const core::lsp::Query &q,
                                           const std::vector<core::FileRef> &filesForQuery) const {
-    auto &gs = undoState.getEvictedGs();
+    const auto &gs = undoState.getEvictedGs();
 
     // We assume gs is a copy of initialGS, which has had the inferencer & resolver run.
     ENFORCE(gs->lspTypecheckCount > 0,
@@ -800,6 +800,7 @@ const ast::ParsedFile &LSPStaleTypechecker::getIndexed(core::FileRef fref) const
 }
 
 std::vector<ast::ParsedFile> LSPStaleTypechecker::getResolved(const std::vector<core::FileRef> &frefs) const {
+    const auto &gs = *(undoState.getEvictedGs());
     vector<ast::ParsedFile> updatedIndexed;
 
     for (auto fref : frefs) {
@@ -808,7 +809,8 @@ std::vector<ast::ParsedFile> LSPStaleTypechecker::getResolved(const std::vector<
             updatedIndexed.emplace_back(ast::ParsedFile{indexed.tree.deepCopy(), indexed.file});
         }
     }
-    return pipeline::incrementalResolve(*(undoState.getEvictedGs()), move(updatedIndexed), config->opts);
+
+    return pipeline::incrementalResolveWithoutStateMutation(gs, move(updatedIndexed), config->opts);
 }
 
 const core::GlobalState &LSPStaleTypechecker::state() const {
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -746,6 +746,9 @@ const core::GlobalState &LSPTypecheckerDelegate::state() const {
     return typechecker.state();
 }
 
+LSPStaleTypechecker::LSPStaleTypechecker(std::shared_ptr<const LSPConfiguration> config, UndoState &undoState)
+    : config(config), undoState(undoState), emptyWorkers(WorkerPool::create(0, undoState.getEvictedGs()->tracer())) {}
+
 void LSPStaleTypechecker::initialize(InitializedTask &task, std::unique_ptr<core::GlobalState> initialGS,
                                      std::unique_ptr<KeyValueStore> kvstore) {
     ENFORCE(false, "initialize not supported");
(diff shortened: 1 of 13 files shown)


main/lsp/LSPTypechecker.cc
```diff
@@ -670,22 +673,52 @@ std::vector<std::unique_ptr<core::Error>> LSPStaleTypechecker::retypecheck(std::
...
+    Timer timeit(config->logger, "query");
+    prodCategoryCounterInc("lsp.updates", "query");
+    ENFORCE(gs->errorQueue->isEmpty());
+    ENFORCE(gs->lspQuery.isEmpty());
+    gs->lspQuery = q;
+    auto resolved = getResolved(filesForQuery);
+    tryApplyDefLocSaver(*gs, resolved);
+    tryApplyLocalVarSaver(*gs, resolved);
+
+    pipeline::typecheck(gs, move(resolved), config->opts, *emptyWorkers, /*presorted*/ true);
```
froydnj, line 696: This is not safe, right?  We don't want `undoState.getEvictedGs` to be modified, yet we're passing it to a function that's going to modify it.  (`getResolved` has the same issue.)

aprocter: Since we're pulling in the evicted file, which was already indexed relative to the evicted gs, I think we shouldn't actually be introducing any new symbols and therefore this should be de facto safe. (Not 100% confident though, so I should also note that this is hidden behind an opt-in CLI flag.)

main/lsp/LSPTypechecker.h
```diff
@@ -200,13 +200,12 @@ class LSPTypecheckerInterface {
...
+    std::shared_ptr<const LSPConfiguration> config;
     UndoState &undoState;
-
-    // Just so we have something to return from the stubbed functions after we ENFORCE(false)
-    ast::ParsedFile *pf;
+    std::unique_ptr<WorkerPool> emptyWorkers;
 
 public:
-    LSPStaleTypechecker(UndoState &undoState) : undoState(undoState), pf(nullptr) {}
+    LSPStaleTypechecker(std::shared_ptr<const LSPConfiguration> config, UndoState &undoState);
```
jez, line 208: It looks like there's an asymmetry between `LSPStaleTypechecker` and `LSPTypecheckerDelegate` here.

Why doesn't `LSPTypecheckerDelegate` take an `LSPConfiguration` or a `WorkerPool`?

aprocter: The `config` is used inside `LSPStaleTypechecker::query` directly because `LSPStaleTypechecker::query` largely duplicates the impelmentation of `LSPTypechecker::query`, while `LSPTypecheckerDelegate` can just delegate `LSPTypechecker::query`, where `config` is a field on `LSPTypechecker`.

The `WorkerPool` we create here exists for similar reasons: `LSPTypechecker` has a private `WorkerPool`. (Note that here we create an empty, i.e., zero-thread, `WorkerPool`, which actually causes work to run on the indexing thread itself.)

main/lsp/LSPTypechecker.cc
jez: I don't know that we have an issue with `typecheck`, because `typecheck` already works with a const GlobalState. I think that the only problem is with `getResolved`.

jez: I think that the idea of using `evictedGs` is a clever idea to not need to borrow the up-to-date GlobalState from the typechecker thread, but I think this confirms that we still want to gate landing this on the non-mutating namer and resolver.

main/lsp/requests/hover.cc
```diff
@@ -37,18 +37,12 @@ HoverTask::HoverTask(const LSPConfiguration &config, MessageId id, std::unique_p
...
-
-        auto clientHoverMarkupKind = config.getClientConfig().clientHoverMarkupKind;
-        string typeString = "stub_answer_for_hover";
-        string docString = "We would have tried to answer this hover request with stale data\n\n";
+    const core::GlobalState &gs = typechecker.state();
 
-        response->result = make_unique<Hover>(formatRubyMarkup(clientHoverMarkupKind, typeString, docString));
-        return response;
+    if (typechecker.isStale()) {
+        config.logger->debug("HoverTask running on stale data");
```
jez, line 43: Is there a way we can force the implementation of each LSPMethod to declare whether it's ok to be called with a stale state? I'd love to force opt-in, not opt-out.

(If that was already implemented in a previous PR I'd love to be pointed to the implementation.)

main/lsp/LSPTypechecker.cc
aprocter: > I think this confirms that we still want to gate landing this on the non-mutating namer and resolver.

Yeah, I'd defer to your judgment on that but I think it's probably a fair call to make.

main/lsp/requests/hover.cc
aprocter: There's a virtual method on `LSPTask` that defaults to `return false`, and we'll only feed an `LSPStaleTypechecker` 
 if it returns `true`. Currently only `hover` and `codeAction` opt in.

https://github.com/sorbet/sorbet/blob/db69c5c127d136d49f206e846cbb986cc9818f2f/main/lsp/LSPTask.h#L114-L115

main/lsp/LSPTypechecker.h
```diff
@@ -200,13 +200,12 @@ class LSPTypecheckerInterface {
 };
 
 class LSPStaleTypechecker final : public LSPTypecheckerInterface {
+    std::shared_ptr<const LSPConfiguration> config;
     UndoState &undoState;
-
-    // Just so we have something to return from the stubbed functions after we ENFORCE(false)
-    ast::ParsedFile *pf;
+    std::unique_ptr<WorkerPool> emptyWorkers;
```
jez, line 205:
```suggestion
    // Using an WorkerPool with size 0 for all typechecker operations causes the work to run on the
    // current thread (usually: the indexer thread).
    std::unique_ptr<WorkerPool> emptyWorkers;
```

aprocter:
🚦 We're going to hold off on merging this til non-mutating namer/resolver are in place, so I've slapped a `[WIP]` on the title here.

aprocter:
https://github.com/sorbet/sorbet/pull/5469/commits/b9cd06d5f93f05abdb24401e6fbda97e212a1afb adds a simple test for a hover query in stale mode.

aprocter:
Pulling this out of `[WIP]` because I think it's ready to go modulo namer/resolver updates. Since everything is hidden behind a flag, I think it'd actually be fine to merge this ahead of the namer/resolver updates, but I'm also fine with holding off until after those are ready.

Flipped a coin to pick a reviewer and jez was the lucky winner. :)

aprocter:
I should also point out that this PR can only access indexed files that were saved into the `UndoState` when the slow path started running. A follow-on change (which I'm planning on opening as a stacked PR in a moment) will allow access to the rest of the indexed files. (Synchronization for that becomes pretty tricky, hence the separate PR.)

test/lsp/multithreaded_protocol_test_corpus.cc
```diff
@@ -675,4 +675,197 @@ TEST_CASE_FIXTURE(MultithreadedProtocolTest, "ErrorIntroducedInSlowPathPreemptio
     assertDiagnostics(send(LSPMessage(make_unique<NotificationMessage>("2.0", LSPMethod::SorbetFence, 20))), {});
 }
 
+TEST_CASE_FIXTURE(MultithreadedProtocolTest, "HoverReturnsStaleInfoOneFile") {
+    // Reset the options to enable stale state and sleep-in-slow-path.
+    auto opts = make_shared<realmain::options::Options>();
+    opts->lspStaleStateEnabled = true;
+    opts->sleepInSlowPath = true;
```
jez, line 682: Is this required? This makes this test even slower than it currently is, right? (iiuc, this makes every slow path operation take 3 seconds artificially)

aprocter: I believe a pause is the best option we have, but the slowness is honestly kind of nasty. (Let me get some numbers for you, but I believe that `multithreaded_protocol_test` takes substantially more time than on `master`.)

What we might be able to do is make the wait time configurable and set it to something much quicker than 3 seconds.

aprocter: Integrating the mechanism implemented in #5539 gives us:

```
master  32.18s
#5539   34.16s (note: adds test of slow path blocking mechanism, which sleeps for 2s)
PR      39.37s
```

Taking a 7-second hit here is still not wonderful, but I'm thinking it's probably acceptable since we're not planning on adding a huge pile of protocol tests anyway.

(Haven't pushed this to this branch yet, but will do so once #5539 hits `master`.)

aprocter: Integrating the mechanism implemented in #5539 gives us:

```
master  32.18s
#5539   34.16s (note: adds test of slow path blocking mechanism, which sleeps for 2s)
PR      39.37s
```

Taking a 7-second hit here is still not wonderful, but I'm thinking it's probably acceptable since we're not planning on adding a huge pile of protocol tests anyway.

(Haven't pushed this to this branch yet, but will do so once #5539 hits `master`.)

froydnj:
I think that since we now use non-mutating namer/resolver, this PR is probably ready for a proper review, is that right @aprocter ?

main/lsp/LSPTypechecker.cc
froydnj: The types here are a bit gross, because `typecheck` doesn't really modify the global state (except for bits that munge the `errorQueue`, but that's already marked `mutable` anyway, but `typecheck` is not typed as such.  I'll fix that up someplace else.

froydnj (approved):
This all looks reasonable; I think there are some type issues around `pipeline::typecheck` and `getEvictedGs`, but we can fix those up after the fact.

PR #5470, opened 2022-03-14
Namer pass that never mutates

jez:
### Motivation

Should unblock work to serve LSP requests from stale data when otherwise it
would have been blocked.

### Test plan

Can't really test this yet. We'll likely test in conjunction with #5469.

But that being said, this code is also not used at all yet.


core/GlobalState.cc
```diff
@@ -295,6 +295,12 @@ void GlobalState::initEmpty() {
     ENFORCE(klass == Symbols::Sorbet_Private_StaticSingleton());
     klass = enterClassSymbol(Loc::none(), Symbols::Sorbet_Private_Static(), core::Names::Constants::StubModule());
     ENFORCE(klass == Symbols::StubModule());
+
+    // TODO(jez) This is weird. Is this the first member on noClassOrModule?
+    typeMember = enterTypeMember(Loc::none(), Symbols::noClassOrModule(), core::Names::Constants::StubTypeMember(),
+                                 Variance::CoVariant);
+    ENFORCE(typeMember == Symbols::StubTypeMember());
+
```
aprocter, line 303: I dimly recall that I had thought about trying to put this on `StubModule` but it seemed to run afoul of an `ENFORCE` while building the initial `GlobalState`---I don't remember exactly where, and I dimly recall thinking that the failing `ENFORCE` might actually have been a simple sanity check rather than a relied-upon property.

You might try moving `StubTypeMember` somewhere that makes more sense---maybe you'll know how to deal with the breakage (if any).

PR #5472, opened 2022-03-15
do less copying for `LSPStdout::rawWrite`

froydnj:
Building up JSON for LSP messages has a few problems when it comes to `LSPStdout::rawWrite` and `JSONBaseType::toJSON`:

* `JSONBaseType::toJSON` has enough information in the buffer it's building so that `std::string::string` doesn't have to call `strlen` to find the length of the string.
* Also in `toJSON`, we copy the entirety of the `StringBuffer` to convert it to a string, when nothing really needs the JSON as a `string`.
* Once we've converted it to a `string`, we copy *again* in `rawWrite` to format it into a string for output.  This step is really unnecessary, as we can just write the string to `cout` separately from the header.  Note that `rawWrite` is called under a lock, so there are no concerns about sending things as a single `write` call or similar; we'll be guaranteed to have completed all the `write` calls that we need before another thread gets a chance to write a message here.

We can fix the first thing fairly easily by calling `string`'s constructor directly.  We can fix the rest of the issues by adding a separate interface to directly build a `StringBuffer`, and then use that to construct `string_view`s for the actual output.

### Motivation

Looking for microseconds in the couch cushions.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPMessage.cc
+++ b/main/lsp/LSPMessage.cc
@@ -222,4 +222,16 @@ string LSPMessage::toJSON(bool prettyPrint) const {
         Exception::raise("LSPMessage is not a request, notification, or a response.");
     }
 }
+
+rapidjson::StringBuffer LSPMessage::toJSONBuffer(bool prettyPrint) const {
+    if (isRequest()) {
+        return asRequest().toJSONBuffer(prettyPrint);
+    } else if (isNotification()) {
+        return asNotification().toJSONBuffer(prettyPrint);
+    } else if (isResponse()) {
+        return asResponse().toJSONBuffer(prettyPrint);
+    } else {
+        Exception::raise("LSPMessage is not a request, notification, or a response.");
+    }
+}
 } // namespace sorbet::realmain::lsp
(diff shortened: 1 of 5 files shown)


main/lsp/LSPMessage.cc
```diff
@@ -222,4 +222,16 @@ string LSPMessage::toJSON(bool prettyPrint) const {
...
+rapidjson::StringBuffer LSPMessage::toJSONBuffer(bool prettyPrint) const {
+    if (isRequest()) {
+        return asRequest().toJSONBuffer(prettyPrint);
+    } else if (isNotification()) {
+        return asNotification().toJSONBuffer(prettyPrint);
+    } else if (isResponse()) {
+        return asResponse().toJSONBuffer(prettyPrint);
+    } else {
+        Exception::raise("LSPMessage is not a request, notification, or a response.");
+    }
```
jez, line 235: should we implement `toJSON` in terms of `toJSONBuffer`?

jez (approved):

PR #5473, opened 2022-03-15
DefinitionResponse to MethodDefResponse

jez:
### Motivation

`DefLocSaver.cc` was only ever sending `DefinitionResponse` objects in response
to method definitions. That meant that a lot of code thinking that we had to
handle class or module definitions was actually dead, and that we couldn't take
advantage of the stricter typing that comes from using `MethodRef` instead of
`SymbolRef`.

The LSP response that is used for class definitions (when the location is on the
class name) is just a normal `ConstantResponse` (because there is a real
`ConstantLit` in the tree for `ClassDef` nodes).

If we want to introduce a specific query response for, e.g., hovering over he
`class` keyword specifically, we could do that, but my vote would be to
introduce a new `ClassDefResponse` class.

### Test plan

No behavior changes, so all tests should pass.

### DIFF
--- a/core/lsp/QueryResponse.cc
+++ b/core/lsp/QueryResponse.cc
@@ -37,7 +37,7 @@ const FieldResponse *QueryResponse::isField() const {
     return get_if<FieldResponse>(&response);
 }
 
-const MethodDefResponse *QueryResponse::isDefinition() const {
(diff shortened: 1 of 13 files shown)


froydnj (approved):
More static types are more better.

PR #5480, opened 2022-03-17
Bump compdb version to fix header-only cc_library assumed as C

clint-stripe:
### Motivation
`SemanticExtension.h` was showing annoying squigglies in vscode. This updates to a newer release to pick up https://github.com/grailbio/bazel-compilation-database/pull/65

### Test plan
image

This works now :)


froydnj (approved):
Thanks!

PR #3207, opened 2020-06-17
Resolve all instance variable warnings

paracycle:
This PR attempts to fix most of the Ruby warnings generated by `sorbet-runtime`. All the ones that can be identified by running the test suite are addressed, except for the ones that are dynamically triggered by the test suite itself.

Coincidentally, this investigation also lead to finding and fixing a memoization attempt using `||= false` which would be executed every time and defeat the aim to memoize the ivar.

### Motivation

Sorbet Runtime generates a lot of warnings like `warning: instance variable @foo not initialized` due to lazy initialization of various instance variables.

### Test plan

No extra automated tests, all the current ones pass. 

Running the test suite using `RUBYOPT=-w bundle exec rake test` displays all the warnings that are generated during a test suite run.

There are still some `method redefinition` warnings left that this PR does not aim to fix.

PR #5482, opened 2022-03-17
Resolve all instance variable warnings

Morriar:
This is a rebased version of @tomstuart's https://github.com/sorbet/sorbet/pull/3338, in turn, was a rebased version of @paracycle’s https://github.com/sorbet/sorbet/pull/3207, which appears to have gone stale. His original description is below.

---

This PR attempts to fix most of the Ruby warnings generated by `sorbet-runtime`. All the ones that can be identified by running the test suite are addressed, except for the ones that are dynamically triggered by the test suite itself.

Coincidentally, this investigation also lead to finding and fixing a memoization attempt using `||= false` which would be executed every time and defeat the aim to memoize the ivar.

### Motivation
Sorbet Runtime generates a lot of warnings like `warning: instance variable @foo not initialized` due to lazy initialization of various instance variables.

### Test plan
No extra automated tests, all the current ones pass.

Running the test suite using `RUBYOPT=-w bundle exec rake test` displays all the warnings that are generated during a test suite run.

There are still some `method redefinition` warnings left that this PR does not aim to fix.

### DIFF
--- a/gems/sorbet-runtime/lib/types/configuration.rb
+++ b/gems/sorbet-runtime/lib/types/configuration.rb
@@ -80,6 +80,7 @@ module T::Configuration
     T::Props::Private::DeserializerGenerator.respond_to?(:generate2)
   end
 
+  @use_vm_prop_serde = false
   # Whether to use VM-defined prop serialization/deserialization routines.
   #
   # The default is to use runtime codegen inside sorbet-runtime itself.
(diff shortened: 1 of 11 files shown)


jez (approved):
This looks great, thanks!

One thing: we have an FAQ about how to define `T.nilable` instance variables and it violates the deprecation warnings. Do you want to update the website docs as well to suggest people a way to define nilable instance variables that doesn't cause warnings?

I'm going to approve merge when tests pass. If you sneak in an update for this before it happens that's great, otherwise it can be in a follow up.

aprocter (approved):

Morriar:
> Do you want to update the website docs as well to suggest people a way to define nilable instance variables that doesn't cause warnings?

Here you go: 5598f91ea423f936bd9085812ae3eab70e4ff139.

PR #5485, opened 2022-03-18
Potentially delete some dead namer code

jez:
### Motivation

I tested the first commit on Stripe's codebase, which raised if the first half of the condition was run. That exception was never thrown during the test run.

I noticed this because the logic it's doing doesn't make any sense.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_LLAE6KNTA0cjY0
→ https://go/builds/bui_LLAEZU31ohdFZ5

froydnj (approved):
Does this same sort of logic also apply in `handleAssignment` a couple hundred lines later?

jez:
> Does this same sort of logic also apply in `handleAssignment` a couple hundred lines later?

Not sure, but I will definitely look at it! Thanks for the callout.

PR #5488, opened 2022-03-18
Don't pause the indexer during initialization

elliottt:
Add a new method to the indexer, `canHandleTask` that's used in the main task handling loop to allow some tasks to filter through while global state initialization is happening in the typechecker. This opens the door to managing multiple package global states, as initialization for one could take place while another was available to answer queries.

I'd recommend viewing with whitespace changes turned off, as there's a large indentation change to ensure that we don't run stale state/preemption tasks on a task in the middle of the queue.

### Motivation

Give the indexer the ability to filter the task queue and select events that it's currently able to handle. Use this new functionality to avoid blocking the task queue while initialization is happening in the typechecking thread.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPIndexer.cc
+++ b/main/lsp/LSPIndexer.cc
@@ -180,8 +180,20 @@ void LSPIndexer::initialize(IndexerInitializationTask &task, std::unique_ptr<cor
 }
 
 bool LSPIndexer::canHandleTask(const LSPTask &task) const {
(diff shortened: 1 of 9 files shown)


jez (approved):

elliottt:

→ https://go/builds/bui_LQNg8zjVMHrVUz
→ https://go/builds/bui_LQNg21ug1kmT3T

PR #5490, opened 2022-03-19
Write a modification timestamp to track if autoloader generation modifies any files

aadi-stripe:
Writes a very simple modification timestamp that updates if the autoloader generation modifies any files.

### Motivation

In the Stripe development environment, we use the full autoloader tree as an input to a local Bazel build system. However, there is a lot of overhead associated with diff-ing the entire autoloader tree to see if something changed. It's a lot easier to diff a single file. Hence, we write a timestamp in case any files change.

### Test plan

Tested on Stripe codebase.

See included automated tests.

### DIFF
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -426,6 +426,10 @@ struct RenderAutoloadTask {
     const DefTree &node;
 };
 
+struct ModificationState {
+    bool modified;
+};
+
 // This function has two duties:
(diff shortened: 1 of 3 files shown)


aadi-stripe:

→ https://go/builds/bui_LLqBJnkLthT45x
→ https://go/builds/bui_LLqBBFDARY2i0v

aisamanra (approved):
This looks good to me!

PR #5491, opened 2022-03-21
remove `else`-after-`return` patterns in resolver

froydnj:
### Motivation

Slightly clearer code.

### Test plan

See included automated tests.


jez:
> Is it weird that I'm sort of neutral-ish on this style choice as a general principle? 🤔 I dunno, maybe being exposed to Haskell at an impressionable age warped my brain in weird ways. ;)

No, I generally prefer symmetrical `if` and `else` blocks, unless it's clear that one of them is much longer or shorter than the others, where it really matters to recover the horizontal space by dedenting.

PR #5493, opened 2022-03-21
Respect `--remove-path-prefix` flag in autogen msgpack output

reese:
### Motivation

Currently, pay-server manually generates a file list to pass to Sorbet, but we'd like to just pass the root directory and let Sorbet walk the directory tree itself, but doing this currently leaves the leading `./` in all msgpack output. This PR makes Sorbet respect the `--remove-path-prefix` option in the msgpack output when set.

### Test plan

See included automated tests, and note that the `loc` values in the fixture correctly remove the path-prefix.

### DIFF
--- a/test/cli/autogen-path-prefix/test.sh
+++ b/test/cli/autogen-path-prefix/test.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -eu
+
+main/sorbet --silence-dev-message --stop-after namer --autogen-version=4 \
+  -p autogen:autogen.txt -p autogen-msgpack:autogen.msgpack \
+  --remove-path-prefix=test/cli/ \
+  test/cli/autogen-path-prefix \
+  test/cli/autogen-path-prefix/file_with_no_dot
+
+echo "--- autogen.txt start ---"
+cat autogen.txt
+echo "--- autogen.txt end ---"
+echo
+
+shasum autogen.msgpack
+
--- a/main/autogen/data/msgpack.cc
+++ b/main/autogen/data/msgpack.cc
@@ -1,6 +1,7 @@
 // has to go first because it violates our poisons
 #include "mpack/mpack.h"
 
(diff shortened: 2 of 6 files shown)


test/cli/autogen-path-prefix/test.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -eu
+
+main/sorbet --silence-dev-message --stop-after namer --autogen-version=4 \
+  -p autogen:autogen.txt -p autogen-msgpack:autogen.msgpack \
+  --remove-path-prefix=test/cli/ \
```
jez, line 6: As far as I can tell, we don't actually use this flag at Stripe anymore, so I'm slightly confused.

If I understand correctly, the test case that you really care about is that if you invoke Sorbet on a directory like `.` (instead of with a list of files), that the `pathPrefix` is set to `./` implicitly.

Do you want to either repurpose this new test to test that behavior, or to add another test for this implicit `pathPrefix` + directory behavior?

reese: So `pathPrefix` is implicitly set to `./` _only_ if you don't also pass in files (see [here](https://github.com/sorbet/sorbet/blob/e4aa5da7a277d8a2c7680260162f3cac29c6da8d/main/options/options.cc#L710-L714)), but since there are files in pay-server that are Ruby files but don't have extensions, we need to be able to pass in both a directory and files and also remove the leading `./`. We don't currently use this flag, but we would need it (or at least something similar) in the future, since the test case you're describing is the existing functionality.

jez: Got it, that makes sense. Thanks!

froydnj (approved):

froydnj:
As jez notes, we don't use this flag at Stripe, so I think this change doesn't require the usual run on Stripe's codebase.

PR #5495, opened 2022-03-22
Add isOpenInClient bit on core::File

jez:
### Motivation

This is currently unused, but we should be able to do this to power LSP
features for things that would otherwise be expensive to compute for
every file in a codebase. The use case that I have in mind right now is
visually distinguishing regions of a file that are `T.untyped`.

I can't currently think of other use cases for this boolean that could crop up
in the future that are actually bad and/or dangerous, so it seems fine to just
let it sit there for the time being.

### Test plan

Because this is unused, I've not added any tests. It's possible that the person
who first uses this feature in the future will need to fix bugs, but at least it
being here should be better than forgetting that this approach is possible.


aprocter (approved):
lgtm!

PR #5500, opened 2022-03-23
Add Hash.try_convert signature

gmcgibbon:
Adds a signature for [`Hash.try_convert`](https://ruby-doc.org/core-3.1.1/Hash.html#method-c-try_convert). It seems this method is missing from Hash's default RBI.

### Motivation

While adding sorbet to some files, I noticed it didn't understand this method. It appears to have been a part of the standard lib [for awhile](https://ruby-doc.org/core-1.9.3/Hash.html#method-c-try_convert), so I think there's no harm in adding it. It is probably just a bit of an obscure method.

### Test plan

Reading the contributing doc and looking at other RBI PRs it looks like I don't need to write tests. Let me know if the signature and docs can be improved! 😄


jez:

→ https://go/builds/bui_LNL9X6gx2geqPT
→ https://go/builds/bui_LNL9bdHQMixJ8j

jez (approved):

PR #5503, opened 2022-03-24
Rewrite all sends to `Class.new`

Morriar:
### Motivation

I noticed that some calls to `Class.new` were not properly rewritten by the `ClassNew` rewriter introduced in #5405 like this one:

```rb
class Bar
  def bar
    Class.new(Foo) do
      # Missing a `T.bind(self, T.class_of(Foo)) here :(
      T.reveal_type(self) # error: Revealed type: `Bar` (instead of `T.class_of(Foo)`)
    end
  end
end
```

This is because the `ClassNew` rewriter was [only called on ClassDefs RHS](https://github.com/sorbet/sorbet/blob/master/rewriter/rewriter.cc#L178-L179).

The way to fix this is to call the rewriter on all sends but doing so will introduce a `T.bind` in rewritten blocks for `A = Class.new` constructs. But doing so will add a `T.bind` in the rewritten class since the sends rewriter will be executed before the assign one.

So this:

```rb
A = Class.new(Foo) do
  T.reveal_type(self)
end
```

Will be rewritten by the send `ClassNew` rewriter as this:

```rb
A = Class.new(Foo) do
  T.bind(self, T.class_of(Foo))
  T.reveal_type(self)
end
```

Then will end up being rewritten by the asgn `ClassNew` rewriter as this:

```rb
class A < Foo
  T.bind(self, T.class_of(Foo)) # :(
  T.reveal_type(self) # error: Revealed type: `T.class_of(Foo)` (instead of `T.class_of(A)`)
end
```

The trick used here is to add the `T.bind` as a `<Magic>.bind` so we can spot it while rewriting `A = Class.new` and remove it.

So this: 

```rb
A = Class.new(Foo) do
  T.reveal_type(self)
end
```

Will be rewritten by the send `ClassNew` rewriter as this:

```rb
A = Class.new(Foo) do
  <Magic>.bind(self, T.class_of(Foo))
  T.reveal_type(self)
end
```

And we can filter the bind out during the asgn `ClassNew` rewriter:

```rb
class A < Foo
  T.reveal_type(self) # error: Revealed type: `T.class_of(A)` :)
end
```

If you have a better idea to not add the `T.bind` in the first place instead of having to filtering it out, I'm happy to update this PR 👍 

### Test plan

See included automated tests.


rewriter/ClassNew.cc
```diff
@@ -10,9 +10,39 @@ using namespace std;
 
 namespace sorbet::rewriter {
 
-vector<ast::ExpressionPtr> rewriteAsClassDef(core::MutableContext ctx, ast::Assign *asgn) {
+// Is this expression a `Magic.bind` call we added from the `ClassNew` rewriter on sends?
```
jez, line 13: Instead of using `<Magic>` bind, can we leave it as just `T.bind` but set the `isRewriterSynthesized` bit on the `Send`?

I'd love to have this change not involve any diff to `GlobalState.cc` nor `resolver.cc`

jez (commented):
looks mostly good to me, but i'd love to not have to introduce `<Magic>.bind`

aprocter (changes requested):
+1 to using `isRewriterSynthesized` instead of `<Magic>.bind` if possible. There's also a failing compiler test that we'll need to get to the bottom of. Looks solid in general though!

Morriar:
Good idea, I changed the `Magic` to use `isRewriterSynthesized` 👍

froydnj:
I haven't investigated deeply, but I imagine the compiler failure in [999f0b6](https://github.com/sorbet/sorbet/pull/5503/commits/999f0b657ab1590124284cb602611c8f57449960) is coming because we have a `Class.new` call that's not assigned to anything...and therefore we never set up proper symbol information for the anonymous class (?).

I'm OK with just deleting that testcase (here or in a separate PR); I don't see what it's trying to test, it belongs to a set of benchmarks that aren't super-relevant, and we don't claim that the compiler handles every single valid Ruby program you can write.

Morriar:
@froydnj I removed the test and all is green, thanks 👍

Morriar:
@aprocter I think this is good to go 👍

jez (approved):
Sorry for the delay, this looks great, thanks!

jez:

→ https://go/builds/bui_Lg71O9AI1Vt9Gf
→ https://go/builds/bui_Lg719fxwiBIWoR

jez:
There are only 10 failures on Stripe's codebase due to this change. I'll do the work to get those fixed, and then land this later today.

jez:
@Morriar the failures I'm seeing are actually all of this shape:

```ruby
# typed: strict
class A
  extend T::Sig
  sig {void}
  def self.make
    cls = Class.new(A) do
    end
  end
end
```

```
❯ sorbet foo.rb
foo.rb:6: T.cast is useless because T.class_of(A) is already a subtype of T.class_of(A) https://srb.help/7015
     6 |    cls = Class.new(A) do
     7 |    end
Errors: 1
```

This was not surfaced in any tests in the repo right now because the file is currently `# typed: true`, and the error is reported at `# typed: strict`.

I pushed some changes to your branch to fix those errors.

jez:

→ https://go/builds/bui_Lg8FA2qYUBxqR3
→ https://go/builds/bui_Lg8FD4TuE1Ve3W

Morriar:
Thanks for pushing this through the finish line! 🙏

PR #5504, opened 2022-03-25
Allow user to dump internal payload files

Morriar:
### Motivation

For some tools it can be useful to know what is inside the payload embedded with Sorbet.

One could read the payload from Github but this payload might not be exactly corresponding to the one embedded in the version of Sorbet used.

This PR adds the `--print=payload-sources` option which writes the content of the payload in RBI files inside the specified directory:

```shell
$ sorbet --print=payload-sources=my-dir
```

### Test plan

See included automated tests.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -78,6 +78,7 @@ const vector<PrintOptions> print_options({
     {"autogen-subclasses", &Printers::AutogenSubclasses},
     {"package-tree", &Printers::Packager, false},
     {"minimized-rbi", &Printers::MinimizeRBI},
+    {"payload-sources", &Printers::PayloadSources},
 });
 
 PrinterConfig::PrinterConfig() : state(make_shared<GuardedState>()){};
@@ -148,6 +149,7 @@ vector<reference_wrapper<PrinterConfig>> Printers::printers() {
         AutogenSubclasses,
         Packager,
         MinimizeRBI,
+        PayloadSources,
     });
 }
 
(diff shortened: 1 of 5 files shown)


jez:
What do you think about using `--print` for this? Using `--print` will get some things for free, like the ability to specify an output file for the print output.

```ruby
--print=payload-sources:out.rb
```

You can also look at the `AutogenAutoloader` / `--print=autogen-autoloader` option to see an example of how to use the `--print` flag in combination with an output directory.

jez (commented):
Mostly fine, but I think I'd like to use `--print` for consistency with similar features in Sorbet.

Morriar:
@jez: Updated to use `--print=payload-sources` instead 👍

main/options/options.cc
```diff
@@ -994,12 +996,22 @@ void readOptions(Options &opts,
...
+        if (opts.print.PayloadSources.enabled) {
+            if (raw.count("e") > 0) {
+                logger->error("You can't pass both `{}` and `{}`.", "--print=payload-sources", "-e");
+                throw EarlyReturnWithCode(1);
+            } else if (!opts.inputFileNames.empty()) {
+                logger->error("You can't pass both `{}` and paths to typecheck.", "--print=payload-sources");
+                throw EarlyReturnWithCode(1);
+            }
+        }
+
```
jez, line 1014: These restrictions seem ~slightly artificial? Does the correctness of the implementation depend on it?

If the user is in a folder with a `sorbet/config` file, it'll have passed `--dir=.` in the config file (usually).

If it's not required for correctness, maybe we can just allow those things to be passed an ignore them?

jez (approved):

jez:
Two questions. If you'd like to decline both of them lmk and I'll click merge.

main/options/options.cc
Morriar: Note for self: We should also reject `--no-stdlib` with this option.

jez: makes sense thanks!

Morriar:
@jez:
* added a check to reject `--no-stdlib` with `--print=payload-sources`
* using `trap EXIT`
* using `mktemp -d`

PR #5505, opened 2022-03-25
No nonsensical completion requests inside blocks

### LINKED ISSUE #5255: Do not show completion results in comments
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0A1.times%20do%0A%20%20%23%20x.%0A%20%20%23%20%20%20%5E%20completion%3A%20%28nothing%29%0A%20%20puts%20'hello%2C%20world!'%0Aend)

```ruby
# typed: true

1.times do
  # x.
  #   ^ completion: (nothing)
  puts 'hello, world!'
end
```

#### Observed output

```
No errors! Great job.
```

image

#### Expected behavior

The user's cursor is currently inside a comment. We shouldn't be showing any completion results. Really, we shouldn't show **any** LSP results there.

- - -

One way to fix this problem at it's root would be to implement the currently-unimplemented `lexer::emit_comment` function so that we could accumulate a list of comments in a file, and store that vector on the `core::File` (probably as a `vector<core::LocOffset>`?).

Then in `queryByLoc`, we would early exit and return no results if the `loc` to query was in the middle of one of the file's comments.

jez:
### Motivation

Fixes #5255

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/completion/constr.rb
+++ b/test/testdata/lsp/completion/constr.rb
@@ -4,12 +4,7 @@ extend T::Sig
 
 xs = T.let([], T::Array[Integer])
 
-# This completion result is a bit non-sensical at the moment (it doesn't makes
-# sense to return any results there) but at least it doesn't crash Sorbet like
-# it used to. If you made a change that improves this test, feel free to delete
-# this comment.
-
 xs.each.map do |x|
   x
 end
-#  ^ completion: each, ...
+#  ^ completion: (nothing)
--- a/cfg/Instructions.cc
+++ b/cfg/Instructions.cc
@@ -135,6 +135,25 @@ Send::Send(LocalRef recv, core::LocOffsets receiverLoc, core::NameRef fun, core:
     histogramInc("cfg.send.args", this->args.size());
 }
 
+core::LocOffsets Send::locWithoutBlock(core::LocOffsets bindLoc) {
+    if (this->link == nullptr) {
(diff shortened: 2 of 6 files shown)


test/testdata/lsp/completion/constr.rb
```diff
@@ -4,12 +4,7 @@
...
-# This completion result is a bit non-sensical at the moment (it doesn't makes
-# sense to return any results there) but at least it doesn't crash Sorbet like
-# it used to. If you made a change that improves this test, feel free to delete
-# this comment.
-
 xs.each.map do |x|
   x
 end
-#  ^ completion: each, ...
+#  ^ completion: (nothing)
```
elliottt, line 10: Why doesn't this return any completion results? Wouldn't the following be an example of where you'd want to allow completion after the `end`?
```ruby
irb(main):001:0> xs = [1,2,3]
=> [1, 2, 3]
irb(main):002:1* xs.map do |a|
irb(main):003:1*   a+1
irb(main):004:1* end.each do |a|
irb(main):005:1*   puts a
irb(main):006:0> end
2
3
4
=> [2, 3, 4]
```

jez: You haven't typed the `.` yet. There's nothing to complete until you type the `.`

elliottt: Ah, that makes sense. Thanks!

elliottt (approved):

Issue #5402, opened 2022-03-02
Snippet to pre-populate a template for adding docs above a method


jez:
### Problem

People don't write docs as much as they should.

### Proposed solution

We should add either a completion request + snippet or a code action to add docs above a method if no docs exist. Something like this exists in other languages:

https://user-images.githubusercontent.com/5544532/156459612-b8272781-a791-4c15-9360-8d81a1051225.mov

PR #5506, opened 2022-03-26
YARD completion snippet after typing `##`

jez:
### Motivation

Hoping to encourage more people to write documentation comments in our codebase.

Fixes https://github.com/sorbet/sorbet/issues/5402

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_LOBTBknX0GuNL2
→ https://go/builds/bui_LOBTsKSYEKYvRJ

jez:

→ https://go/builds/bui_LOBoAzciK73sp0
→ https://go/builds/bui_LOBoIb6tCCAsWZ

test/helpers/lsp.cc
```diff
@@ -508,6 +510,19 @@ vector<unique_ptr<LSPMessage>> getLSPResponsesFor(LSPWrapper &wrapper, unique_pt
     return getLSPResponsesFor(wrapper, move(messages));
 }
 
+namespace {
+
+// Like absl::StripTrailingAsciiWhitespace, but only blank characters (tabs and spaces)
+ABSL_MUST_USE_RESULT inline absl::string_view stripTrailingAsciiBlank(absl::string_view str) {
```
elliottt, line 516: Does `[[nodiscard]]` not work here? It looks like abseil suggests using that instead for c++17 and above: https://github.com/abseil/abseil-cpp/blob/0c6302fe427963ec5c471d3ee660120682ab15f7/absl/base/attributes.h#L403-L404

```suggestion
[[nodiscard]] inline absl::string_view stripTrailingAsciiBlank(absl::string_view str) {
```

main/lsp/requests/completion.cc
```diff
@@ -681,28 +681,6 @@ unique_ptr<CompletionItem> trySuggestSig(LSPTypecheckerInterface &typechecker,
...
-    if (receiverSym == core::Symbols::rootSingleton()) {
-        receiverSym = core::Symbols::Object().data(gs)->lookupSingletonClass(gs);
-    }
-    auto methodOwner = targetMethod.data(gs)->owner;
-
-    if (!(methodOwner == receiverSym || methodOwner == receiverSym.data(gs)->attachedClass(gs))) {
-        // The targetMethod we were going to suggest a sig for is not actually in the same scope as this sig.
-        return nullptr;
-    }
-
```
elliottt, line 705: Is this all redundant after the changes to `NextMethodFinder`?

elliottt (approved):
This looks good!

test/helpers/lsp.cc
jez: It probably works, I just copied the implementation of `StripTrailingAsciiWhitespace` verbatim with no changes other than the `is_blank` thing.

I'll change this in a follow up so i don't have to wait for tests to build for this PR.

main/lsp/requests/completion.cc
jez: Yes. I had to do this in NextMethodFinder for two reasons:

- I wanted to share this logic between suggested sigs and suggested YARD docs
- This logic relies on looking at the `enclosingMethod` that comes back in the LSP QueryResponse to figure out whether the method we find is owned by the same owner as the `enclosingMethod.data(gs)->owner`

For the YARD snippet case, we don't have any QueryResponses, which means no enclosing method, which means we need a purely syntactic way to check that we're in the same nesting scope as the queryLoc.

PR #5508, opened 2022-03-27
take `Span<const T>` instead of `const InlinedVector<T, N> &` args

froydnj:
### Motivation

Over in #5413, I had the idea to reduce the inlined storage for some `TypeArgumentRef`-storing and `TypeMemberRef`-storing vectors.  This turned out to be messier than necessary, because the types that are used for storage are also implicated in the argument types for various functions.

I think it would be cleaner if those various functions took `Span`, rather than dictating the particulars about what kind of storage the caller used.  And if those various functions took `Span`, that would make the original PR much more obvious in terms of its effects.

Hence, this PR.

### Test plan

See included automated tests.  No functional change intended.

### DIFF
--- a/core/BUILD
+++ b/core/BUILD
@@ -36,6 +36,7 @@ cc_library(
         "//main/pipeline/semantic_extension:interface",
         "@com_google_absl//absl/strings",
         "@com_google_absl//absl/synchronization",
+        "@com_google_absl//absl/types:span",
         "@rang",
     ],
 )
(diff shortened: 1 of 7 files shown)


jez:
What is a Span?

froydnj:
> What is a Span?

`absl::Span` is like `std::string_view`, but for arbitrary contiguous storage of any `T`, not just `char`.

jez (approved):

PR #5517, opened 2022-03-29
Fix export_for_test misattribution bug

aadi-stripe:
Fixes an export misattribution bug in the case of `export_for_test` autocorrects that appears due to transitivity of imports.

### Motivation

In `--stripe-packages` mode there is a bug wherein `export_for_test` autocorrect suggestions are misattributed to the wrong package. Consider the following:

```
  # foo/__package.rb
  class Opus::Foo < PackageSpec
    export Opus::Foo::A
  end
```

```
  # foo/a.rb
  class Opus::Foo::A
    # ...
   end
```

```
  # foo/unexported.rb
  class Opus::Foo::Unexported; end
```

```
  # bar/__package.rb
  class Opus::Bar < PackageSpec
     import Opus::Foo
  end
```

```
  # bar/b.test.rb
  class Test::Opus::Bar::B
      Opus::Foo::Unexported # <=== error
  end
```

Currently this would throw the following error:
```
    bar/__package.rb:3: To expose this name to a package's own tests it must be exported. Do you need to `export_for_test Opus::Foo` in this package?
     3 |class Opus::Bar < PackageSpec
     4 |  import Opus::Foo
     5 |end
    bar/__package.rb:4: Constant `Opus::Foo` is defined here:
     4 |  import Opus::Foo
```

The issue is that `Opus::Bar` contains the symbol `Opus::Foo` via an import, so the scope check finds `Opus::Foo` inside `Opus::Bar` and incorrectly attributes an export to it. The fix is simple -- ignore symbols that are "defined" in `__package.rb` files, as they are always going to be transitively imported.

### Test plan

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -265,11 +265,12 @@ public:
         if (ctx.file.data(ctx).isPackagedTest()) {
             // In a test file first look to see in our own package to see if it's missing an `export_for_test`
             core::SymbolRef sym = findPrivateSymbol(ctx, scope, /* test */ false);
-            if (sym.exists() && sym.isClassOrModule()) {
+            if (sym.exists() && sym.isClassOrModule() && !sym.loc(ctx).file().isPackage()) {
                 res.emplace_back(MissingExportMatch{sym, this->mangledName()});
                 return res;
             }
         }
+
         for (auto &imported : importedPackageNames) {
             auto &info = ctx.state.packageDB().getPackageInfo(imported.name.mangledName);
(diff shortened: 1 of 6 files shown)


jez (approved):

aadi-stripe:

→ https://go/builds/bui_LPNOd8ClQGAbKF
→ https://go/builds/bui_LPNO6XNbkuMSho

ngroman (approved):
Nice find

PR #5521, opened 2022-03-29
Expose more indexed files to LSPStaleTypechecker

aprocter:
(stacked PR, on top of #5545)

This adds a mechanism to share the `indexed` files between `LSPTypechecker` and `LSPStaleTypechecker`, so that we can get hover responses on files other than the one(s) included in the edit that triggered the slow path.

The synchronization here is a bit wonky, but I think it's probably the best we can do for the moment without a pretty significant refactor. I think that refactor would probably look something like: `LSPIndexedFileStore` is created outside of either `LSPTypechecker` or `LSPStaleTypechecker`, and passed to those classes by reference; and `LSPIndexedFileStore` provides an accessor for a mutex, with suitable annotations on its methods.

### Motivation
We want stale-state queries to work on files that weren't part of the edit that triggered the slow path.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/BUILD
+++ b/main/lsp/BUILD
@@ -34,6 +34,7 @@ cc_library(
     ],
     hdrs = [
         "LSPConfiguration.h",
+        "LSPIndexedFileStore.h",
         "LSPInput.h",
         "LSPMessage.h",
         "LSPOutput.h",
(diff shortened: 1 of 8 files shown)


jez:
Closing as we have decided against pursing the stale typechecker approach.

PR #1144, opened 2019-06-29
Kernel number conversion functions support "exception" kwarg

pje:
This was added in Ruby 2.6: https://bugs.ruby-lang.org/issues/12732

Implements #1060

PR #5526, opened 2022-03-29
Numeric classes produce nilable numbers when exception: false

atanbt:
When exception: false is set on numeric types like Float, BigDecimal, etc... the return type can be nil.

```ruby
irb(main):002:0> Float('a', exception: false)
=> nil
```

### Motivation

The existing signatures are incorrect.

### Test plan

I did not include tests as this is simply making a signature correct.

### DIFF
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -755,7 +755,7 @@ module Kernel
         digits: Integer,
         exception: T::Boolean
     )
-    .returns(BigDecimal)
+    .returns(T.nilable(BigDecimal))
   end
   def BigDecimal(initial, digits=0, exception: true); end
   ### As of 2.6, all these numeric conversion methods *can* return `nil` iff given the kwarg `exception: false`, which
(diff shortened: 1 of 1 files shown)


froydnj (changes requested):
Thanks for the PR!

Unfortunately, there are a number of test failures as a result of these signature changes.  There are also ~hundreds of errors on Stripe's codebase with this change.

I think it's completely reasonable to want to fix this, but I don't think changing the signatures to make the return value unconditionally nilable is the way to go.  For instance, writing:

```ruby
RATE = T.let(Rational(1, 2), Rational)
```

no longer works if the signatures uncondtionally return a nilable type, which seems unergonomic.  The comments below each of the methods you changed gives a rationale as to why the signatures are constructed the way they are.  #1144 also has some discussion of these same issues.

The more complete, but more complex, solution would be adding custom logic in `core/types/calls.cc` for each method, parsing the argument list, and determining:

1. If `exception` was not passed, return a non-nilable type.
2. If `exception` was passed and is `false`, return a nilable type.
3. If `exception` was passed and is `true`, return a non-nilable type.
4. If `exception` was passed and is boolean-typed, return a nilable type (we can't tell at analysis time what the runtime value is going to be, so we have to be as general as possible).
5. If `exception` was passed and is untyped, return a nilable type (same logic).
6. If `exception` was passed and is anything else, error (Ruby will error at runtime).

Adding keyword argument support for overloads (see also #1144) would also be a way to solve this problem, but I think that's quite a bit more work than the solution above.

Please let us know if you decide to implement custom logic in `calls.cc`, and we can try to provide what assistance we can; I don't think any of us are going to be able to work on it directly.

PR #5527, opened 2022-03-29
Defer sorting resolver todo lists until it matters

jez:
### Motivation

If `t` is number of TODO items and `e` is number of errors, this changes the
algorithm from

`O(t log(t))` → `O(e log(e))`

`t` scales with codebase size. `e` also scales with codebase size, but much
smaller, and only for builds that we don't expect to typecheck (branch builds).

### Test plan

Hopefully this has no changes, so existing tests should cover this.

~~I plan to test the perf impact on Stripe's codebase~~

Timings:
http://go/gist/jez/963ced6034405215ec5530f87ccf5075

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1582,16 +1582,7 @@ public:
             }
         }
 
-        // Note: `todo` does not need to be sorted. There are no ordering effects on error production.
-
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_LPc3JKc0OR5FTj
→ https://go/builds/bui_LPc34yLBxwWseD

aprocter (approved):

PR #5529, opened 2022-03-29
use cwd from lspConfigs in vs code extension

judocode:
Sorbet is not currently respecting the cwd as defined in the `sorbet.lspConfigs`. It continues to reference the root no matter what you put as the cwd. This appears due to the fact that Sorbet is always passing in the `workspace.rootPath` when spawning the typecheck: https://github.com/sorbet/sorbet/blob/master/vscode_extension/src/LanguageClient.ts#L249.

### Motivation
See above

### Test plan
There don't seem to be any relevant tests for this part of the code and I'm unsure of how to test this specific change.

### DIFF
--- a/vscode_extension/src/LanguageClient.ts
+++ b/vscode_extension/src/LanguageClient.ts
@@ -246,7 +246,7 @@ export default class SorbetLanguageClient implements ErrorHandler {
     ] = this._sorbetExtensionConfig.activeLspConfig!.command;
     this._outputChannel.appendLine(`    ${command} ${args.join(" ")}`);
     this._sorbetProcess = spawn(command, args, {
-      cwd: workspace.rootPath,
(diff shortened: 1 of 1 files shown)


elliottt (changes requested):
This fails for me on stripe's codebase, as `this._sorbetExtensionConfig.activeLspConfig!.cwd` is `undefined` when sorbet launches. What do you think about modifying this so that it falls back on `workspace.rootPath` when `cwd` is `undefined`?

bbugh:
@elliottt FWIW I think that's a great idea. Currently blocked by running this in a monorepo because cwd isn't being read.

bbugh:
It appears though that this does not expand the vscode macros, so `${workspaceFolder}/project-folder` is read as literal `${workspaceFolder}/project-folder`, which of course fails. It works when specifying an absolute path literal. 

It seems like there needs to be some processing done before using the `cwd` for the child process.

PR #5531, opened 2022-03-30
Report all package export dups

ngroman:
There was a bug where we would not find all duplicate exports. Previously we were advancing the interator in a way that skipped comparisons between siblings (that can't prefix each other but can be dups). I also ensured that public exports are ranked ahead of export_for_test

### Motivation
Fix bug

### Test plan

See included automated tests.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -145,8 +145,8 @@ struct Import {
 };
 
 enum class ExportType {
-    Public,
-    PrivateTest,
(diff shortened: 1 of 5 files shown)


nroman-stripe:

→ https://go/builds/bui_LPfZe8GqNcnDN4
→ https://go/builds/bui_LPfZWq9Sg1HKFw

jez (approved):

PR #5532, opened 2022-03-30
Non-mutating version of `resolveConstants`

jez:
### Motivation

This will allow us to make more progress on #5469 (LSP queries on stale state)

### Test plan

This calls the immutable incremental resolver immediately after the immutable
namer and enables all ENFORCEs and sanityChecks, so that we can use our existing
test suite as a test of how well these modes work.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -585,6 +585,9 @@ void GlobalState::initEmpty() {
     enterMethodArgumentSymbol(Loc::none(), method, Names::args());
     ENFORCE(method == Symbols::todoMethod());
 
+    method = this->staticInitForClass(core::Symbols::root(), Loc::none());
+    ENFORCE(method == Symbols::rootStaticInit());
+
     klass = enterClassSymbol(Loc::none(), Symbols::Sorbet_Private_Static(), core::Names::Constants::ResolvedSig());
     ENFORCE(klass == Symbols::Sorbet_Private_Static_ResolvedSig());
     klass = Symbols::Sorbet_Private_Static_ResolvedSig().data(*this)->singletonClass(*this);
@@ -2190,10 +2193,10 @@ MethodRef GlobalState::staticInitForClass(ClassOrModuleRef klass, Loc loc) {
     return sym;
 }
 
-MethodRef GlobalState::lookupStaticInitForClass(ClassOrModuleRef klass) const {
+MethodRef GlobalState::lookupStaticInitForClass(ClassOrModuleRef klass, bool allowMissing) const {
     auto classData = klass.data(*this);
     auto ref = classData->lookupSingletonClass(*this).data(*this)->findMethod(*this, core::Names::staticInit());
-    ENFORCE(ref.exists(), "looking up non-existent <static-init> for {}", klass.toString(*this));
+    ENFORCE(ref.exists() || allowMissing, "looking up non-existent <static-init> for {}", klass.toString(*this));
     return ref;
 }
 
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -130,6 +130,111 @@ MethodBuilder enterMethod(GlobalState &gs, ClassOrModuleRef klass, NameRef name)
     return MethodBuilder{gs, gs.enterMethodSymbol(Loc::none(), klass, name)};
 }
 
+struct ParentLinearizationInformation {
+    const InlinedVector<core::ClassOrModuleRef, 4> &mixins;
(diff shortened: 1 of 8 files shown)


core/GlobalState.cc
```diff
@@ -835,6 +835,7 @@ void GlobalState::initEmpty() {
             Symbols::MAX_SYNTHETIC_TYPEARGUMENT_SYMBOLS);
 
     installIntrinsics();
+    computeLinearization();
```
jez, line 841: Motivation is also in this commit message:

```
This should be safe--if we ever call addMixin after this, we will reset
the linearization status on that class.

This allows us to call `verifyLinearizationComputed` in
`runIncrementalBestEffort` (because our test suite tests an edge case
where we run `runIncrementalBestEffort` with `--no-stdlib` so Resolver
truly has never been run before).
```

resolver/resolver.cc
```diff
@@ -3846,10 +3891,11 @@ ast::ParsedFilesOrCancelled Resolver::runIncremental(core::GlobalState &gs, vect
...
 }
 
-ast::ParsedFilesOrCancelled Resolver::runIncrementalWithoutStateMutation(const core::GlobalState &gs,
-                                                                         vector<ast::ParsedFile> trees) {
+// TODO(jez) This is dupliicated with Resolver::runIncremental. Can we make them share more code?
+ast::ParsedFilesOrCancelled Resolver::runIncrementalBestEffort(const core::GlobalState &gs,
+                                                               vector<ast::ParsedFile> trees) {
     auto workers = WorkerPool::create(0, gs.tracer());
-    // trees = ResolveConstantsWalk::resolveConstants(gs, std::move(trees), *workers);
+    trees = ResolveConstantsWalk::resolveConstants(gs, std::move(trees), *workers);
```
jez: This was accidentally commented out in the original change. It's quite important that it not be commented out (this is basically all the work that resolver does 😅 )

resolver/resolver.cc
```diff
@@ -1598,80 +1626,84 @@ class ResolveConstantsWalk {
...
-                    const auto origSize = job.items.size();
-                    auto g = [&](AncestorResolutionItem &item) -> bool {
-                        auto resolved = resolveAncestorJob(ctx, item, false);
-                        if (resolved) {
-                            tryRegisterSealedSubclass(ctx, item);
-                        }
-                        return resolved;
+        // If the constant didn't immediately resolve during the initial treewalk, and we're not
+        // allowed to mutate GlobalState, it will never resolve. Let's just skip to the error phase.
+        if constexpr (isMutableStateType) {
```
jez, line 1631: This is the key insight that makes this change workable at all, and not immeasurably hairy.

The ResolveConstantsWalk will try to avoid enqueuing work into the various todo lists for the sake of speed. That means that if there's anything in the todo lists at the end of the loop, they're there because they were hoping to be resolved / processed once some other todo list item had been processed.

But we won't actually be able to make progress. Making progress requires being able to modify the symbol table, so that other items can discover new state in the symbol table and make their own progress. That will never happen in the immutable resolver mode, so we can skip straight to the final part.

test/pipeline_test_runner.cc
```diff
@@ -413,18 +413,40 @@ TEST_CASE("PerPhaseTest") { // NOLINT
...
         }
 
-        move(namer::Namer::symbolizeTreesBestEffort(*gs, move(treesCopy), *workers).result());
+        treesCopy = move(namer::Namer::symbolizeTreesBestEffort(immutableGS, move(treesCopy), *workers).result());
         ENFORCE(!gs->hadCriticalError());
 
-        // If no ENFORCE fired, then non-mutating namer is working fine.
+        treesCopy = move(resolver::Resolver::runIncremental(immutableGS, move(treesCopy)).result());
+
+        // If no ENFORCE fired, then immutatable namer + resolver is working fine.
```
jez, line 449: I've just realized that we probably want to run DefLocSaver and typecheckOne on these resolved trees. Both of those operate on immutable GlobalState already, so it's just a question of do our namer and resolver passes do a good enough job at cleaning up the trees.

jez: I added this.

elliottt: Was the intent here to ensure that we're able to answer LSP queries when using a stale global state?

elliottt (approved):

test/pipeline_test_runner.cc
jez: yep!

PR #5535, opened 2022-03-30
Check payload prefix against Sorbet SCM revision

Morriar:
### Motivation

When Sorbet runs in release build the payload prefix is _not_ based on `URL_PREFIX`.

Example in release build:

```
https://github.com/sorbet/sorbet/tree/32d144b7989158a1f3f67868dbd77cfda5deadc5/rbi/core/argf.rbi
```

vs. in debug build:

```
https://github.com/sorbet/sorbet/tree/master/rbi/core/argf.rbi
```

This results in all payload files [being ignored](https://github.com/Morriar/sorbet/blob/ad278ef9304130029e48e490f6cd70b8f9110d42/main/realmain.cc#L518-L520) when dumping the payload sources in release build. This PR changes the way we compute path prefix for payload files to take into account the SCM revision when in release build.

### Test plan

See included automated tests.

### DIFF
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -511,8 +511,10 @@ int realmain(int argc, char *argv[]) {
                 continue;
             }
 
+            auto payloadVersion = sorbet_is_release_build ? sorbet_build_scm_revision : "master";
             auto payloadPath = payloadFile->path();
-            auto payloadPrefix = absl::StrCat(core::File::URL_PREFIX, "rbi/");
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Is it possible to write a cli test for this?  I don't know that it would have to be super-complicated, just something that would grep the output for `Kernel`, or something else known to exist in the payload.

froydnj:
> We already have a test for this: https://github.com/sorbet/sorbet/blob/master/test/cli/print-payload-sources/test.sh#L30-L32
> 
> The problem is the test runs in debug mode so it didn't catch the different prefix.

:+1:

*mumbles something about test coverage*

PR #5536, opened 2022-03-30
de-duplicate protocol test definitions

froydnj:
### Motivation

I would like to split up the protocol tests to make them run faster, and this is an easy first step to making that slightly easier.

### Test plan

See included automated tests.

### DIFF
--- a/test/lsp/lsp_test.bzl
+++ b/test/lsp/lsp_test.bzl
@@ -1,3 +1,10 @@
+def basename(p):
+    return p.rpartition("/")[-1]
+
+def dropExtension(p):
+    "TODO: handle multiple . in name"
+    return p.partition(".")[0]
+
 def lsp_test(path):
     # path will be like `$name/$name.sh`
     words = path.split("/")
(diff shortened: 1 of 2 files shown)


test/lsp/lsp_test.bzl
```diff
@@ -48,3 +55,26 @@ def update_test():
...
         tests = update_rules,
     )
+
+_PROTOCOL_TEST_SRCS = ["ProtocolTest.cc", "ProtocolTest.h"]
+
+def protocol_tests(cc_files):
+    for cc_file in cc_files:
+        native.cc_test(
+            name = dropExtension(basename(cc_file)),
+            size = "medium",
```
elliottt, line 65: Not at all blocking, but do these tests pass with `timeout = "moderate"` instead of `size = "medium"`?

https://bazel.build/reference/be/common-definitions#test.timeout

elliottt (approved):
🚀

test/lsp/lsp_test.bzl
froydnj: These pass on my machine with `timeout = "moderate"` and `-c dbg`, so I think we are good here!

PR #5537, opened 2022-03-30
Detach the "Move method" logic from the `LSPTask`

ilyailya:
The change detaches the "Move method" logic from the `LSPTask` into free functions. To do that I had to
- Move all `queryBy*` functions into a separate file and also detach them from `LSPTask`
- Move `getRenameEdits` to the Renamer class
- Finally move the "Move method" logic

### Motivation

As part of UX work for the "Move method" refactoring, I want to distinguish clients who does support the `codeAction/resolve` request and who doesn't. That means I'd need to call the refactoring functions from the `CodeActionTask` in some cases and  from the `CodeActionResolveTask` in others. To do that, the refactoring logic shouldn't be hardly tied to one exact `LSPTask`

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPQuery.h
+++ b/main/lsp/LSPQuery.h
@@ -1,0 +1,21 @@
+#ifndef SORBET_LSP_QUERY_H
+#define SORBET_LSP_QUERY_H
+
+#include "main/lsp/LSPTypechecker.h"
+
+namespace sorbet::realmain::lsp {
+
+std::vector<std::unique_ptr<core::lsp::QueryResponse>>
+filterAndDedup(const core::GlobalState &gs,
+               const std::vector<std::unique_ptr<core::lsp::QueryResponse>> &queryResponses);
+
+LSPQueryResult queryByLoc(const LSPConfiguration &config, LSPTypecheckerInterface &typechecker, std::string_view uri,
+                          const Position &pos, LSPMethod forMethod, bool errorIfFileIsUntyped = true);
+LSPQueryResult queryBySymbolInFiles(const LSPConfiguration &config, LSPTypecheckerInterface &typechecker,
+                                    core::SymbolRef symbol, std::vector<core::FileRef> frefs);
+LSPQueryResult queryBySymbol(const LSPConfiguration &config, LSPTypecheckerInterface &typechecker,
+                             core::SymbolRef symbol);
+
+} // namespace sorbet::realmain::lsp
+
+#endif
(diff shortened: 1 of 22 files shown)


main/lsp/LSPQuery.h
```diff
@@ -0,0 +1,21 @@
...
+LSPQueryResult queryByLoc(const LSPConfiguration &config, LSPTypecheckerInterface &typechecker, std::string_view uri,
+                          const Position &pos, LSPMethod forMethod, bool errorIfFileIsUntyped = true);
+LSPQueryResult queryBySymbolInFiles(const LSPConfiguration &config, LSPTypecheckerInterface &typechecker,
+                                    core::SymbolRef symbol, std::vector<core::FileRef> frefs);
+LSPQueryResult queryBySymbol(const LSPConfiguration &config, LSPTypecheckerInterface &typechecker,
+                             core::SymbolRef symbol);
+
+} // namespace sorbet::realmain::lsp
+
+#endif
```
jez, line 21: Given that this file is called `LSPQuery`, what do you think about moving these methods to either be public static methods (or methods inside a namspace) on a namespace called `LSPQuery`?

```cpp
LSPQuery::byLoc(...)
LSPQuery::bySymbol(...)
```

main/lsp/requests/move_method.cc
```diff
@@ -0,0 +1,224 @@
+#include "main/lsp/requests/move_method.h"
```
jez, line 1: I'd love to keep `main/lsp/requests/` focused exclusively on the code that handles requests.

What do you think about calling this file something like

```
main/lsp/MoveMethod.cc
main/lsp/MethodCallSiteRenamer.cc
...
```

(flexible on the name, but just up one level in the tree)

jez (commented):
Some small suggestions

jez (approved):

main/lsp/LSPQuery.h
ilyailya: Thanks, that's a great suggestion! I was thinking about grouping those functions into the object, but didn't see any clear value of doing that. But with a rename you proposed it makes much more sense semantically

main/lsp/requests/move_method.cc
ilyailya: Sounds good to me! 
Somewhat related question, do we have a code style regarding naming files and variables? I've seen a mix of camel case and snake case for files and variables/functions.

PR #5539, opened 2022-03-30
Add a mechanism to block the slow path in tests

aprocter:
For the stale-state feature we need a way to block slow paths programmatically in tests.

### Motivation
The immediate motivation is to get rid of reliance on (brittle, slow) sleeps in the protocol tests for #5469. We can also use this mechanism in the future to implement position assertion-style tests for stale-state queries.

### Test plan
See included automated tests.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -450,6 +450,13 @@ bool LSPTypechecker::runSlowPath(LSPFileUpdates updates, WorkerPool &workers, bo
             return;
         }
 
+        // Test-only hook: Stall for as long as `slowPathBlocked` is set.
+        {
+            absl::MutexLock lck(&slowPathBlockedMutex);
+            slowPathBlockedMutex.Await(absl::Condition(
+                +[](bool *slowPathBlocked) -> bool { return !*slowPathBlocked; }, &slowPathBlocked));
+        }
+
         ENFORCE(gs->lspQuery.isEmpty());
         if (gs->sleepInSlowPath) {
             Timer::timedSleep(3000ms, *logger, "slow_path.resolve.sleep");
@@ -686,6 +693,11 @@ bool LSPTypechecker::tryRunOnStaleState(std::function<void(UndoState &)> func) {
         func(*cancellationUndoState);
         return true;
     }
+}
+
+void LSPTypechecker::setSlowPathBlocked(bool blocked) {
+    absl::MutexLock lck(&slowPathBlockedMutex);
+    slowPathBlocked = blocked;
 }
 
 LSPTypecheckerDelegate::LSPTypecheckerDelegate(TaskQueue &queue, WorkerPool &workers, LSPTypechecker &typechecker)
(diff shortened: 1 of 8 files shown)


main/lsp/LSPTypechecker.cc
```diff
@@ -450,6 +450,13 @@ bool LSPTypechecker::runSlowPath(LSPFileUpdates updates, WorkerPool &workers, bo
             return;
         }
 
+        // Test-only hook: Stall for as long as `slowPathBlocked` is set.
+        {
+            absl::MutexLock lck(&slowPathBlockedMutex);
+            slowPathBlockedMutex.Await(absl::Condition(
+                +[](bool *slowPathBlocked) -> bool { return !*slowPathBlocked; }, &slowPathBlocked));
```
froydnj, line 457: Looking at the abseil sources, it looks like this can just be:

```suggestion
            slowPathBlockedMutex.Await(absl::Condition(&slowPathBlocked));
```

or did you try that already and found that it didn't work?

test/lsp/multithreaded_protocol_test_corpus.cc
```diff
@@ -675,4 +675,57 @@ TEST_CASE_FIXTURE(MultithreadedProtocolTest, "ErrorIntroducedInSlowPathPreemptio
...
+        auto status = getTypecheckRunStatus(*readAsync());
+        REQUIRE(status.has_value());
+        REQUIRE_EQ(*status, SorbetTypecheckRunStatus::Started);
+    }
+
+    // We expect the slow path to be blocked, so we want to make sure that we _don't_ receive any messages within a
+    // pretty long time frame---let's say 2000ms.
+    {
+        auto &wrapper = dynamic_cast<MultiThreadedLSPWrapper &>(*lspWrapper);
+        auto msg = wrapper.read(2000);
```
froydnj, line 716: ...and yet, timeouts manage to sneak their way into the patch anyway. :)

main/lsp/LSPTypechecker.cc
```diff
@@ -450,6 +450,13 @@ bool LSPTypechecker::runSlowPath(LSPFileUpdates updates, WorkerPool &workers, bo
             return;
         }
 
+        // Test-only hook: Stall for as long as `slowPathBlocked` is set.
```
froydnj, line 453: WDYT about moving this below the `ENFORCE(gs->lspQuery.isEmpty());` so it's sort of visually obvious that `slowPathBlocked` and `sleepInSlowPath` are doing the same-ish thing?

froydnj (approved):
This is a lot nicer than some of the other alternatives I was contemplating.

main/lsp/LSPTypechecker.cc
aprocter: Ooh, no, let me give it a shot!

main/lsp/LSPTypechecker.cc
aprocter: Good call, will do

main/lsp/LSPTypechecker.cc
aprocter: Ah, unfortunately this doesn't quite work because the condition we're waiting for is that `slowPathBlocked` is false, not true. I suppose we could change this internally, but on balance I think the extra complexity here is worth keeping "true means blocked" semantics on the surface and in the implementation. (Glad to change in a later PR if there's objections, though.)

test/lsp/multithreaded_protocol_test_corpus.cc
aprocter: Yeah, so unsatisfying! At least there's only one instead of the dozens that #5469 would have needed, I s'pose :)

Issue #4338, opened 2021-07-09
Better error message when specific component of `T.all` isn't satisfied


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Amodule%20IFoo%3B%20end%0Aclass%20Foo%0A%20%20include%20IFoo%0Aend%0A%0AT.let(Foo.new%2C%20T.all(Class%2C%20IFoo)))

```ruby
# typed: true

module IFoo; end
class Foo
  include IFoo
end

T.let(Foo.new, T.all(Class, IFoo))
```

#### Observed output

```
editor.rb:8: Argument does not have asserted type T.all(Class, IFoo) https://srb.help/7007
     8 |T.let(Foo.new, T.all(Class, IFoo))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Got Foo originating from:
    editor.rb:8:
     8 |T.let(Foo.new, T.all(Class, IFoo))
              ^^^^^^^
Errors: 1
```

#### Expected behavior

The error is correct but it can be tricky to parse:

The problem is that `Foo` is not a subtype of `Class`, not that `Foo` is not a subtype of `IFoo`.

We should have a better error message here. For example, maybe something like this:

```
editor.rb:8: Argument does not have asserted type `T.all(Class, IFoo)` https://srb.help/7007
     8 |T.let(Foo.new, T.all(Class, IFoo))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Note:
    `Foo` is NOT a subtype of `Class`
    `Foo` IS a subtype of `IFoo`
  Got Foo originating from:
    editor.rb:8:
     8 |T.let(Foo.new, T.all(Class, IFoo))
              ^^^^^^^
Errors: 1
```

- - -

We might want some way for `isSubType` or `isSubTypeUnderConstraint` to return more than a bool but to instead return an explanation.

This would **likely be expensive** no matter how it was implemented, so we should probably have a way to make it opt-in or only used when we know that the explanation would be useful.

This is related to #4110, which also has the problem of wishing that `isSubType` had an explanation.

Also, there is something similar for certain error messages when constraints fail to solve:

```ruby
# typed: true
extend T::Sig

sig do
  type_parameters(:U)
    .params(
      f: T.proc.params(x: T.type_parameter(:U)).void,
      x: T.type_parameter(:U)
    )
    .void
end
[...]

PR #5541, opened 2022-03-30
add explanation for mismatched `T.any` types in containers

froydnj:
### Motivation

A similar case came up for a user at Stripe, where they had code similar to the added testcase, but some of their types were not quite right.  Sorbet obviously knew which types were not right, but was subsuming them all underneath a `T.any` and leaving the user to pick out which ones were wrong.

(There were also packaging-related problems, but I think we're going to try addressing those in a different PR.)

I am not 100% sure this is the right way to go: in particular, I think the error message might only be helpful in the `T.class_of` case highlighted by the added test.  We don't have location information for which subexpression is incorrectly typed, so we're relying on the user to be able to pick out which thing has the type that we're telling them about.

I think this also falls down if you have a tuple type where you wanted an array type, etc.  I suppose that could be a further refinement.

### Test plan

See included automated tests.

### DIFF
--- a/core/TypeErrorDiagnostics.cc
+++ b/core/TypeErrorDiagnostics.cc
@@ -1,5 +1,9 @@
 #include "core/TypeErrorDiagnostics.h"
 
+#include "common/typecase.h"
+
+#include <algorithm>
+
 using namespace std;
 
 namespace sorbet::core {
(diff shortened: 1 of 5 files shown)


froydnj:
jez pointed out #4338, and I added some cases showing how we could handle at least the bits about `T.all` types.

I haven't taken a close look at some of the test expectation updates; it's possible this could be too noisy?

jez:
I figure I'd point out: I'm not a huge fan of this approach. I really don't want to be in the business of adding ad hoc reimplementions of the core algorithm in subtyping.cc.

I think that we will likely want to implement support for this feature by threading (optional) state down through isSubType so that instead of just returning a boolean. That would reuse existing code, and work for every type error immaginable:

- union and intersection types
- generic constraints
- missing keys in a shape type

and importantly, it could all compose—if the problem is that you're missing a generic constraint or a shape key deep inside some `T.any` structure, it would naturally show you that.

If we really insist on this approach, I could maybe be convinced, but this seems like the kind of code that will just get reverted eventually.

froydnj:
> I think that we will likely want to implement support for this feature by threading (optional) state down through isSubType so that instead of just returning a boolean.

This is a good suggestion and can probably even be implemented without loss of regular typechecking performance and with zero code duplication by doing fancy things with templates.

I am a little more skeptical that you can get really good error messages out of this scheme, but I guess that would be a hurdle to overcome were we to actually implement said scheme.

neilparikh:
Done in #9433 and #7832

PR #4578, opened 2021-09-05
Upgrade to Bazel 4.2.1

jez:
### Motivation

This still uses the x86_64 version of Bazel on M1 MacBook Pro's. This
combination (using Bazel 4.2.1 under x86_64) is sufficient to build Sorbet on an
M1 MacBook (Bazel 4.2.x fixes the local cc toolchain for darwin_arm64).

Eventually, this will let us upgrade grailbio/bazel-toolchain to a version that
supports targeting arm64 binaries directly.

### Test plan

- [x] Linux Buildkite CI
- [x] Linux personal PC
- [x] Linux Stripe devbox
- [x] Linux Stripe CI
- [x] macOS Buildkite CI
- [x] macOS Stripe laptop
- [x] macOS personal laptop
- [x] macOS Stripe arm64 laptop

PR #5543, opened 2022-03-30
Update FAQ to remove the reference to Apple Silicon

joevandyk:
https://github.com/sorbet/sorbet/pull/4576
https://github.com/sorbet/sorbet/pull/4578

sorbet does support m1 macs, right?


joevandyk:
Ah, sounds good.

PR #5544, opened 2022-03-31
Improve "Too many arguments" error message

jez:
### Motivation

- Shows the red squiggle on the extra args, not the whole method call
  (and definitely not on any block that is present).

- Adds an autocorrect that can be run to delete the extra args (useful
  for things like codemods that remove args from methods).

### Test plan

Review by commit to check the test changes.


froydnj (approved):
Did you want to rename the arity test with an "autocorrect" prefix, for parity with other tests?

jez:

→ https://go/builds/bui_LQ2vfsOO2jGC42
→ https://go/builds/bui_LQ2vtG4CRT4f5Z

PR #5545, opened 2022-03-31
Position tests for stale-state hovers

aprocter:
(Stacked on #5469)

This adds support for `.rbstaleupdate` files to the `lsp_test_runner`.

WIP because there's some sloppy code duplication I think I can fix. But it seems to work on a simple example test!

### Motivation
We want to be able to test our new stale-state functionality without writing a bunch of protocol tests.

### Test plan
See included automated tests.

### DIFF
--- a/test/lsp_test_runner.cc
+++ b/test/lsp_test_runner.cc
@@ -400,6 +400,12 @@ TEST_CASE("LSPTest") {
     /** Test expectations. */
     Expectations test = Expectations::getExpectations(singleTest);
 
+    /** Currently we cannot mix .rbupdate and .rbstaleupdate files, since the latter requires us to use a multithreaded
+        LSPWrapper and the former requires us to use a single-threaded LSPWrapper. */
+    bool haveSourceUpdates = test.sourceLSPFileUpdates.size() > 0;
+    bool haveStaleUpdates = test.staleLSPFileUpdates.size() > 0;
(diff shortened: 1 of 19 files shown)


test/lsp_test_runner.cc
```diff
@@ -700,58 +716,142 @@ TEST_CASE("LSPTest") {
...
-                        if (params->status == SorbetTypecheckRunStatus::Ended) {
-                            foundTypecheckRunInfo = true;
-                            if (assertSlowPath.value_or(false)) {
-                                INFO(errorPrefix << "Expected Sorbet to take slow path, but it took the fast path.");
+
+            // TODO(aprocter): There's probably more code duplication than necessary between the 'if' and the 'else'
+            // here.
+            if (haveStaleUpdates) {
+                // Wait for a fence, to clear out any pending slow-path operations.
+                (void)getLSPResponsesFor(*lspWrapper, std::vector<std::unique_ptr<LSPMessage>>{});
```
froydnj, line 724: We don't need to loop here to account for potential delays in receiving the typechecking done notification?

froydnj (approved):
This makes sense to me!

test/lsp_test_runner.cc
aprocter-stripe: This should clear out the typechecking-done notification, since it will be sent as part of the slowpath run triggered in the edit task. (It seemed to fix the problem we had been seeing, anyway :))

PR #5547, opened 2022-03-31
Overhaul "Passing a hash with unknown keys" error

jez:
### Motivation

This is a super commonly asked question.

### Test plan

Check the commits for the before/after.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -574,8 +574,6 @@ void maybeSuggestUnsafeKwsplat(const core::GlobalState &gs, core::ErrorBuilder &
     }
 }
 
-} // namespace
-
 // Ensure that a ShapeType used as a keyword args splat in a send has only symbol keys present.
 const ShapeType *fromKwargsHash(const GlobalState &gs, const TypePtr &ty) {
     auto *hash = cast_type<ShapeType>(ty);
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -981,12 +981,16 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
         } else if (kwSplatIsHash) {
             bool hasKwSplatParam = false;
             bool hasKwParam = false;
+            const ArgInfo *firstKwParam;
             const ArgInfo *kwSplatParam;
             for (auto &param : data->arguments) {
                 if (param.flags.isKeyword && param.flags.isRepeated) {
                     kwSplatParam = &param;
                     hasKwSplatParam = true;
                 } else if (param.flags.isKeyword && !param.flags.isRepeated) {
+                    if (!hasKwParam) {
+                        firstKwParam = &param;
+                    }
                     hasKwParam = true;
                 }
             }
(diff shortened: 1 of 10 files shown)


core/types/calls.cc
```diff
@@ -957,10 +961,36 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
             } else {
                 // A keyword splat was passed in, but none of the declared arguments are keyword splats,
                 // or there is a keyword splat argument, but there is one ore more keyword non-splat argument
-                if (auto e = gs.beginError(args.callLoc(), errors::Infer::UntypedSplat)) {
-                    e.setHeader("Passing a hash where the specific keys are unknown to a method taking keyword "
-                                "arguments");
+                auto kwSplatArgLoc = args.argLoc(args.args.size() - 1);
```
froydnj, line 964: This PR adds tests which suggest that assuming the kwsplat arg is the last arg is not a correct assumption.

test/cli/autocorrect-kwsplat/test.out
```diff
@@ -0,0 +1,118 @@
...
+    Sorbet cannot statically guarantee that the splatted `Hash` contains all the keys required by `foo`.
+    If that's okay, use `T.unsafe` to silence this error, otherwise pass the required keyword parameters directly.
+  Autocorrect: Done
+    test.rb:17: Replaced with `**T.unsafe({x: 0, **opts})`
+    17 |foo(0, 1, 2, x: 0, **opts)
+                     ^^^^^^^^^^^^
+
+test.rb:19: Passing a non-literal `Hash` to `Object#foo`, which requires specific keyword parameters https://srb.help/7019
+    19 |foo(0, 1, 2, **{x => 0})
+                     ^^^^^^^^^^
```
froydnj, line 78: The error message seems technically correct but a bit confusing in this case: we have a "literal" hash, but the problem is that we can't statically tell what the keys are.  The "Passing a hash where the specific keys are unknown" message was a bit clunky, but it covered this case.  I don't know if there's a happy middle ground between those two.

core/types/calls.cc
```diff
@@ -957,10 +961,36 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                        auto replaceValue = replaceLoc.source(gs).value();
+
+                        if (absl::c_all_of(replaceValue, [](char c) { return absl::ascii_isalnum(c) || c == '_'; }) ||
+                            (absl::StartsWith(replaceValue, "{") && absl::EndsWith(replaceValue, "}"))) {
+                            e.replaceWith("Wrap in `T.unsafe`", replaceLoc, "{}T.unsafe({})", maybeStarStar,
+                                          replaceLoc.source(gs).value());
+                        } else {
+                            e.replaceWith("Wrap in `T.unsafe`", replaceLoc, "{}T.unsafe({{{}}})", maybeStarStar,
+                                          replaceLoc.source(gs).value());
+                        }
```
froydnj, line 992: The `replaceLoc.source(gs).value()` in the conditionals could just be `replaceValue`?

test/cli/autocorrect-kwsplat/test.out
```diff
@@ -0,0 +1,118 @@
...
+    test.rb:15:
+    15 |foo(0, 1, 2, **opts, x: 0)
+                     ^^^^^^^^^^^^
+  Note:
+    Sorbet cannot statically guarantee that the splatted `Hash` contains all the keys required by `foo`.
+    If that's okay, use `T.unsafe` to silence this error, otherwise pass the required keyword parameters directly.
+  Autocorrect: Done
+    test.rb:15: Replaced with `T.unsafe({opts, x: 0})`
+    15 |foo(0, 1, 2, **opts, x: 0)
+                       ^^^^^^^^^^
```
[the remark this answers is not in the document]
jez: > Do we want to highlight all relevant kwargs (as this PR currently shows) for the error message about passing a non-literal hash, or do we want to highlight just the kwsplat arg? The latter seems better to me.

Unfortunately, we don't have that location right now. Note that there could technically be multiple of them:

```ruby
f(**opts1, **opts2, x: nil)
```

We only track one kwsplat for the whole function, and I don't want to try to re-parse it from within calls.cc

> This is a weird case, but the autocorrect suggested by Sorbet is not doing the right thing: it should be T.unsafe({**opts, x: 0}) instead of T.unsafe({opts, x: 0}).

Thanks, good catch. I'll think about that.

core/types/calls.cc
jez: It is last and the only arg, because we use Desugar to change `**opts, x: nil` into a single splat arg using `<Magic>.<to-hash-dup>` and `<Magic>.<merge>`

test/cli/autocorrect-kwsplat/test.out
jez: ACK i can change this message

test/cli/autocorrect-kwsplat/test.out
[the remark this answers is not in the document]
jez: (the current iteration I fixes this)

jez:

→ https://go/builds/bui_LQQNtjfOCVKRhI
→ https://go/builds/bui_LQQNsFWBQP3sxB

test/cli/autocorrect-kwsplat/test.out
jez: i went a little overboard designing a better message, to the point where I actually relaxed the criteria we were using here. it might have the side effect of making it less annoying to use splats with sorbet.

PR #5551, opened 2022-03-31
Use callLocs in fewer error messages

jez:
### Motivation

The `args.callLocs()` contains the block argument, if one exists. Any time we
use this in an error message, it'll chunder across the whole screen, turning
large swaths of it red, obscuring other errors.

Even if there isn't a block, usually there's something more specific that we can
give.

This change searches for a bunch of places where `callLocs` is used and stops
using it.

### Test plan

Each commit is arguably an independent change, but given that they're so closely
related in spirit I've made only one PR.

You can review by commit to see the changes to tests next to the relevant change
to calls.cc

### DIFF
--- a/core/Types.h
+++ b/core/Types.h
@@ -893,6 +893,24 @@ struct DispatchArgs {
     Loc argLoc(size_t i) const {
         return core::Loc(locs.file, locs.args[i]);
     }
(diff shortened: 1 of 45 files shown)


elliottt (approved):
Looks good! I like the more precise error locations 👍

PR #5556, opened 2022-03-31
Show true result type in hover

jez:
### Motivation

@elliottt nerd sniped me into this one

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/hover.cc
+++ b/main/lsp/requests/hover.cc
@@ -16,7 +16,7 @@ string methodInfoString(const core::GlobalState &gs, const core::TypePtr &retTyp
                         const unique_ptr<core::TypeConstraint> &constraint) {
     string contents;
     auto start = &dispatchResult;
-    ;
+
     while (start != nullptr) {
         auto &component = start->main;
         if (component.method.exists()) {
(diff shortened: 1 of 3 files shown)


test/testdata/lsp/hover.rb
```diff
@@ -139,10 +139,12 @@ def main
       # ^^^^^^ hover: The docs for BigFoo
 
   foo = BigFoo.new
-  #            ^^^ hover: sig {params(args: T.untyped, blk: T.untyped).returns(BigFoo)}
+  #            ^^^ hover: sig {params(args: T.untyped, blk: T.untyped).returns(T.untyped)}
   #            ^^^ hover: def new(*args, &blk); end
+  #            ^^^ hover: # result type: BigFoo
```
jez, line 144: because we have an intrinsic for `.new` this shows up weirdly

main/lsp/requests/hover.cc
```diff
@@ -16,19 +16,21 @@ string methodInfoString(const core::GlobalState &gs, const core::TypePtr &retTyp
...
+
     while (start != nullptr) {
         auto &component = start->main;
         if (component.method.exists()) {
             if (!contents.empty()) {
                 contents += "\n";
             }
             contents = absl::StrCat(
-                contents, prettyTypeForMethod(gs, component.method, component.receiver, retType, constraint.get()));
+                contents, prettyTypeForMethod(gs, component.method, component.receiver, nullptr, constraint.get()));
```
jez, line 27: we already use this `nullptr` pattern in completion requests, as it turns out.

froydnj (approved):
This seems fine, but it looks like a test is failing?  (I think we want to keep the test as-is, so I guess `hover.cc` needs some adjustment?)

PR #5558, opened 2022-04-01
always pass `const DispatchArgs &` to `dispatchCall`-related functions

froydnj:
### Motivation

Consistency and epsilon performance improvements (we were copying `DispatchArgs` on every. single. `dispatchCall`.).

We already did something like this in #3640, but it didn't stick.  Let's try a little harder this time, with deleting the copy constructor for `DispatchArgs`.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_LQbZ84r8FkAa3i
→ https://go/builds/bui_LQbZDNFX6SpoPj

jez (approved):

PR #5561, opened 2022-04-01
make the `const`ness of `GlobalState` more explicit during `typecheck`

froydnj:
### Motivation

Some of the stale-state LSP work really wants to work with `const GlobalState`.  Despite not mutating `GlobalState`, `pipeline::typecheck` is typed in such a way that would lead you to believe that it does.  Let's fix `pipeline::typecheck`'s signature, and make other non-mutating code more obvious in the process.

### Test plan

See included automated tests.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -1058,31 +1058,23 @@ ast::ParsedFilesOrCancelled resolve(unique_ptr<core::GlobalState> &gs, vector<as
     return ast::ParsedFilesOrCancelled(move(what));
 }
 
-void typecheck(const unique_ptr<core::GlobalState> &gs, vector<ast::ParsedFile> what, const options::Options &opts,
+void typecheck(const core::GlobalState &gs, vector<ast::ParsedFile> what, const options::Options &opts,
                WorkerPool &workers, bool cancelable,
-               optional<shared_ptr<core::lsp::PreemptionTaskManager>> preemptionManager, bool presorted,
-               bool intentionallyLeakASTs) {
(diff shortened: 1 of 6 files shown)


main/pipeline/pipeline.cc
```diff
@@ -1060,6 +1062,14 @@ void typecheck(const unique_ptr<core::GlobalState> &gs, vector<ast::ParsedFile>
                WorkerPool &workers, bool cancelable,
                optional<shared_ptr<core::lsp::PreemptionTaskManager>> preemptionManager, bool presorted,
                bool intentionallyLeakASTs) {
+    return typecheck(reinterpret_cast<const unique_ptr<const core::GlobalState> &>(gs), move(what), opts, workers,
+                     cancelable, move(preemptionManager), presorted, intentionallyLeakASTs);
```
jez, line 1066: why not just make `typecheck` take a `const core::GlobalState&` and then delete the overload?

you've already proven that nothing in typecheck needs a mutable state, and we can easily dereference the gs before calling typecheck at the callsites.

jez (changes requested):
^

jez:
I went ahead and pushed that change to your branch because i wanted to confirm whether it was possible. if you don't want it, feel free to revert and we can chat more

jez (approved):

froydnj:
> why not just make `typecheck` take a `const core::GlobalState&` and then delete the overload?

:man_facepalming: yes, that makes a lot more sense, doesn't it?

> I went ahead and pushed that change to your branch because i wanted to confirm whether it was possible. if you don't want it, feel free to revert and we can chat more

I pushed essentially the same change, but deleting the `igs` variable in `pipeline::typecheck` because it wasn't necessary anymore.

PR #5568, opened 2022-04-04
[LSP] Lazily compute edits for the "Move method" refactoring

ilyailya:
### Motivation

Computing edits to move a method may be a long task (imagine a method with a lot of call sites). Instead of calculating them on each code action request (which can happen on every cursor move), the change puts those computations into `codeAction/resolve` request.

The resolve request is called only if response for a code action response has an empty `edit` properties. 

### Test plan

See included automated tests.


main/lsp/LSPConfiguration.cc
```diff
@@ -84,6 +84,17 @@ LSPClientConfiguration::LSPClientConfiguration(const InitializeParams &params) {
                 clientHoverMarkupKind = getPreferredMarkupKind(contentFormat);
             }
         }
+        if (textDocument->codeAction) {
+            auto &codeAction = *textDocument->codeAction;
+            if (codeAction->dataSupport) {
+                clientCodeActionDataSupport = codeAction->dataSupport.value_or(false);
+            }
+
+            if (codeAction->resolveSupport) {
```
froydnj, line 93: (This is non-blocking)

`std::optional` coercion to `bool` is neat, but I think the explicit `.has_value()` is much clearer, especially when the optional is holding a value that can be coerced to a `bool` itself (pointers, integers, `bool`, etc.).  I don't think we have agreement on this yet, though, and we obviously don't have any linting for it or such, so this is mostly food for thought.

main/lsp/LSPConfiguration.cc
```diff
@@ -84,6 +84,17 @@ LSPClientConfiguration::LSPClientConfiguration(const InitializeParams &params) {
...
+                auto &properties = (*codeAction->resolveSupport)->properties;
+                clientCodeActionResolveEditSupport = std::count(properties.begin(), properties.end(), "edit") > 0;
```
froydnj, line 95:
```suggestion
                clientCodeActionResolveEditSupport = absl::c_count(properties, "edit") > 0;
```

(This might require including `absl/algorithm/container.h` or somesuch.  Using `c_any_of` is also an option that would avoid the `> 0` test.)

froydnj (approved):
Cool!  A couple of non-blocking things, below.

PR #5574, opened 2022-04-04
Introduce the `constant_hash`

aisamanra:
This introduces (but does not yet use) a "constant hash", a function which takes a Ruby file and produces a hash which is—at first approximation—a function of the constants defined in the file.

Things which should cause the constant hash of a file to change:
- Adding, removing, or renaming a class or module
- Adding, removing, or renaming a constant definition
- Modifying the RHS of a constant _alias_ (i.e. `C = D`)
- Adding or removing an `include` or `extend` in the body of a class or module
- Adding, removing, or modifying a `require` at the top-level (because this affects autoloader output)

Things should should **not** cause the constant hash of a file to change:
- Adding, removing, or renaming a method
- Any edits which happen to the body of a method
- Modifying the RHS of a constant _definition_ (i.e. `C = any_other_expression`)
- Any top-level or body-level send which isn't a `require`, `include`, or `extend`

### Motivation
This is going to be used for autogen-related passes, since this can very quickly let us know whether autogen output might change or not. In practice, most code edits _won't_ change the constant hashes of files, because they are limited to adding/removing methods or changing the interior of methods, which means we'll be able to use this hash to avoid re-running autogen functionality.

I'm going to wire this up in a future PR.

### Test plan

See included automated tests. This adds a test to verify that many of the above edits do or don't change the expected output from simple files.


main/autogen/constant_hash.h
```diff
@@ -0,0 +1,34 @@
...
+#define AUTOGEN_CONSTANT_HASH_H
+
+#include <string>
+#include <vector>
+
+#include "core/GlobalState.h"
+
+namespace sorbet::parser {
+// pre-declaring this instead of importing this because including
+// "parser/Node.h" more than once causes errors
```
froydnj, line 11: I fixed the include issues in #5576, I think, but I also think predeclaring `Node` is better than pulling in a bunch of stuff from `sorbet::parser` in this header.

main/autogen/constant_hash.cc
```diff
@@ -0,0 +1,125 @@
...
+//
+// - the reason it takes a parsed AST (rather than, say, a desugared
+// one) is that this means we can do less allocation and run the hash
+// as quickly as possible.
+//
+// - the reason it takes a pointer rather than a `unique_ptr` is so it
+// doesn't need to `move` or consume the AST: it just observes it. The
+// reason it takes a pointer rather than a reference is so we don't
+// need to check for nullability on every recursive call: we can check
+// once at the head of `hashNode` instead.
```
froydnj, line 22: Do you have an objection to taking `const unique_ptr<Node> &`?  That would at least let you get rid of the `.get()` calls and should be roughly as efficient.

main/autogen/constant_hash.h
```diff
@@ -0,0 +1,34 @@
...
+// important to the autoloader output, but we might remove that if we
+// change the structure of the autoloader. Changes to method bodies,
+// class bodies that don't add or remove other constant definitions,
+// and so forth should produce an identical constant hash.
+//
+// The end-goal is that if a file produces the same constant hash,
+// then we do not need to re-run autogen-related logic on that file,
+// because no changes have happened which could possibly affect what
+// autogen produces.
+unsigned int constantHashNode(core::GlobalState &ctx, parser::Node *what);
```
froydnj, line 31:
```suggestion
unsigned int constantHashNode(core::GlobalState &gs, parser::Node *what);
```

main/autogen/constant_hash.cc
```diff
@@ -0,0 +1,125 @@
...
+        },
+        [&](parser::Module *module) {
+            hashSoFar = core::mix(hashSoFar, core::_hash("(m"));
+            hashSoFar = hashNode(gs, hashSoFar, module->name.get());
+            hashSoFar = core::mix(hashSoFar, core::_hash(" "));
+            hashSoFar = hashNode(gs, hashSoFar, module->body.get());
+            hashSoFar = core::mix(hashSoFar, core::_hash(")"));
+        },
+        [&](parser::Class *claz) {
+            hashSoFar = core::mix(hashSoFar, core::_hash("(c"));
```
froydnj, line 100: For avoidance of doubt: it doesn't matter that this prefix is also used for LHS constant definitions, right?

main/autogen/constant_hash.cc
```diff
@@ -0,0 +1,125 @@
...
+            // hash the scope then the name
+            hashSoFar = hashNode(gs, hashSoFar, cnst->scope.get());
+            hashSoFar = core::mix(hashSoFar, core::_hash("::"));
+            hashSoFar = core::mix(hashSoFar, core::_hash(cnst->name.shortName(gs)));
+        },
+        [&](parser::Send *send) {
+            // only need to handle if this is a `require`, `include`,
+            // or `extend`
+            if (send->method == core::Names::require()) {
+                hashSoFar = core::mix(hashSoFar, core::_hash("(r"));
```
froydnj, line 50: Maybe it would be slightly simpler to say:

```c++
string_view prefix;
if (send->method == core::Names::require()) {
  prefix = "(r";
} else if (send->method == core::Names::include()) {
  prefix = "(i";
} else if (send->method == core::Names::extend()) {
  prefix = "(e";
} else {
  return;
}

// now do hashing with prefix, send->method.shortName(gs), and ")"
```

main/autogen/constant_hash.cc
```diff
@@ -0,0 +1,125 @@
...
+        what,
+        [&](parser::Const *cnst) {
+            // hash the scope then the name
+            hashSoFar = hashNode(gs, hashSoFar, cnst->scope.get());
+            hashSoFar = core::mix(hashSoFar, core::_hash("::"));
+            hashSoFar = core::mix(hashSoFar, core::_hash(cnst->name.shortName(gs)));
+        },
+        [&](parser::Send *send) {
+            // only need to handle if this is a `require`, `include`,
+            // or `extend`
```
froydnj, line 48: It seems weird to me to not care about the names that are being `require`'d, etc., but I guess we're assuming that the end result (for autogen's purposes) is independent of the ordering here?

froydnj (approved):
LGTM!  Comments below; any suggested transformations are non-blocking.

main/autogen/constant_hash.cc
aisamanra: Actually, you're right; I think we should care about the names being `require`'d. I'm going to tweak and add tests to make sure that gets captured.

main/autogen/constant_hash.cc
aisamanra: It should not matter, because classes invariably also add a `<` to their hash (regardless of whether their superclass exists or not) while this constant never will. Might be a good idea to change it just for the mnemonic, though: the specific structure of what we're hashing is mostly for a reader understanding what's going on, because I personally think it's clearer.

main/autogen/constant_hash.cc
aisamanra: None at all! That's easy enough to do.

main/autogen/constant_hash.cc
aisamanra: I've tweaked the logic of this a bit to collapse some cases, but they changed to take arguments into account anyway.

main/autogen/constant_hash.cc
nroman-stripe: ++ on this change

aisamanra:
I've done the changes suggested above and in particular fixed the argument-of-`require` issue, along with corresponding tests. Let me know if anything else should change!

main/autogen/test/constant_hash_test.cc
```diff
@@ -0,0 +1,185 @@
...
+        auto node = parser::Parser::run(gs, file, settings);
+        return autogen::constantHashNode(gs, node.get());
+    }
+};
+
+TEST_CASE("SimpleClass") { // NOLINT
+    Helper helper;
+
+    // we'll use this simple file as our baseline
+    auto basicHash = helper.hashExample("class Foo\n"
```
nroman-stripe, line 46: What do you think about hard-coding some of the expected values in this test to ensure hash stability?

nroman-stripe (approved):
Some inline

main/autogen/test/constant_hash_test.cc
aisamanra: Hard-coded some examples in the test, too.

main/autogen/constant_hash.cc
```diff
@@ -0,0 +1,128 @@
...
+    typecase(
+        what.get(),
+        [&](parser::Const *cnst) {
+            // hash the scope then the name
+            hashSoFar = hashNode(gs, hashSoFar, cnst->scope);
+            hashSoFar = core::mix(hashSoFar, core::_hash("::"));
+            hashSoFar = core::mix(hashSoFar, core::_hash(cnst->name.shortName(gs)));
+        },
+        [&](parser::Send *send) {
+            // only need to handle if this is a `require`, `include`,
```
froydnj, line 47: Do we also need to handle `require_relative`?

main/autogen/constant_hash.h
froydnj, line 11:
```suggestion
```

I think we should just delete this now.

froydnj (approved):
Thank you!

main/autogen/constant_hash.cc
aisamanra: Luckily, no! The autoloader only cares about `require`; `require_relative` gets stripped out there.

PR #5575, opened 2022-04-04
Remove custom Stripe `autogen.rb` error message

jez:
### Motivation

1.  This script was removed over two years ago

2.  It now most commonly occurs when the user has forgotten to import the
    `Opus::Autogen` package. The red squiggles originating from this error
    message obscures the packager error message that is reported on
    `Opus::Autogen`.

### Test plan

See commit history for before+after.


elliottt (approved):
🔥

PR #5578, opened 2022-04-04
Fix crash in YieldLoadParams

jez:
### Motivation

We have a helper function called `getCallArguments` that takes types like:

    T.proc.params(arg0: Integer, arg1: Integer).void

into types like:

    [Integer, Integer]

There is a special case that when youc all `getCallArguments` on
`T.untyped`, it will return `T.untyped`. Unfortunately, if you call it on
something that is typed but isn't a valid `getCallArguments` target
(like `BasicObject`) it would return `nullptr` instead of `T.untyped` which
would cause crashes downstream.

### Test plan

I checked that this test case crashed before this change.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1443,7 +1443,7 @@ TypePtr getMethodArguments(const GlobalState &gs, ClassOrModuleRef klass, NameRe
     MethodRef method = klass.data(gs)->findMethodTransitive(gs, name);
 
     if (!method.exists()) {
-        return nullptr;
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_LRrg9s0DhfsVBP
→ https://go/builds/bui_LRrggMyC3vtmtA

jez:
(manual testing confirms that this fixes the crash on Stripe's codebase)

elliottt (approved):

PR #5583, opened 2022-04-05
add a LSP testcase for hovering an undefined constant

froydnj:
### Motivation

I don't know that this is really the behavior we *want* for this (`Sorbet::Private::*` symbols shouldn't leak out via LSP), but responses like this came up while thinking up tests for stale state, so we might as well capture today's behavior.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/lsp_helpers.cc
+++ b/main/lsp/lsp_helpers.cc
@@ -214,7 +214,9 @@ string prettyTypeForConstant(const core::GlobalState &gs, core::SymbolRef consta
     ENFORCE(constant == constant.dealias(gs));
 
     core::TypePtr result;
-    if (constant.isClassOrModule()) {
+    if (constant == core::Symbols::StubModule()) {
+        result = core::Types::untyped(gs, constant);
+    } else if (constant.isClassOrModule()) {
         auto targetClass = constant.asClassOrModuleRef();
         if (!targetClass.data(gs)->attachedClass(gs).exists()) {
             targetClass = targetClass.data(gs)->lookupSingletonClass(gs);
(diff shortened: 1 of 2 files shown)


elliottt (approved):

test/testdata/lsp/hover_undefined_constant.rb
```diff
@@ -0,0 +1,8 @@
+# typed: true
+
+class MyClass
+  def foo(x)
+    x + THE_CONSTANT # error: Unable to resolve constant
+      # ^ hover: T.untyped
```
jez, line 6:
```suggestion
    #   ^ hover: T.untyped
```

main/lsp/lsp_helpers.cc
```diff
@@ -214,7 +214,9 @@ string prettyTypeForConstant(const core::GlobalState &gs, core::SymbolRef consta
     ENFORCE(constant == constant.dealias(gs));
 
     core::TypePtr result;
-    if (constant.isClassOrModule()) {
+    if (constant == core::Symbols::StubModule()) {
+        result = core::Types::untyped(gs, constant);
```
jez, line 218: Maybe we should just return a string that says "this constant doesn't exist"?

froydnj: @elliottt pointed out that `T.reveal_type` on a constant like this returns `T.untyped`, so there is at least some justification for untyped here.

I don't have strong opinions; I guess "this thing doesn't exist" is more user-friendly than `T.untyped`?

elliottt: I'm fine with either, both are more informative than reporting the location of `StubModule` :)

froydnj:
I went with "This constant is undefined" for the "type" of the constant.

PR #5585, opened 2022-04-05
Also test immutableNamerResolver after mutable run

jez:
### Motivation

Some notes:

- Has to run on the indexed trees (isn't idempotent, so can't run on
  already-resolved trees).
- This commit also changes Error::isCritical, because there are a
  handful of other errors in `core/errors/internal.h`: `WrongSigil`,
  `CyclicReferenceError`, and `FileNotFound`. Those three things are
  user errors. `InternalError` is what we raise when we're recovering
  from a thrown exception or something like that. That's the only thing
  we really care about being called "critical"

### Test plan

test-only change

### DIFF
--- a/core/Error.cc
+++ b/core/Error.cc
@@ -41,7 +41,7 @@ string ErrorColors::replaceAll(string_view inWhat, string_view from, string_view
 }
 
 bool Error::isCritical() const {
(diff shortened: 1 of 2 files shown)


elliottt (approved):
Looks good!

PR #5588, opened 2022-04-06
Interleave "Did you mean" suggestions for constants

jez:
### Motivation

Make constant resolution errors somewhat less terrible.

### Test plan

See included automated tests.

**Before**

Screen Shot 2022-04-05 at 9 17 39 PM

**After**

Screen Shot 2022-04-05 at 9 17 09 PM

### DIFF
--- a/core/Error.cc
+++ b/core/Error.cc
@@ -140,11 +140,12 @@ void ErrorBuilder::addErrorSection(optional<ErrorSection> &&section) {
 
 void ErrorBuilder::addAutocorrect(AutocorrectSuggestion &&autocorrect) {
     ENFORCE(state == State::WillBuild);
-    const char *sectionTitle;
+    string sectionTitle;
     if (gs.autocorrect) {
         sectionTitle = "Autocorrect: Done";
-    } else if (autocorrect.isDidYouMean) {
-        sectionTitle = "Did you mean? Use `-a` to autocorrect";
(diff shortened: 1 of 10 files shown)


core/Error.cc
```diff
@@ -140,7 +140,14 @@ void ErrorBuilder::addErrorSection(optional<ErrorSection> &&section) {
 
 void ErrorBuilder::addAutocorrect(AutocorrectSuggestion &&autocorrect) {
     ENFORCE(state == State::WillBuild);
-    auto sectionTitle = gs.autocorrect ? "Autocorrect: Done" : "Autocorrect: Use `-a` to autocorrect";
+    const char *sectionTitle;
+    if (gs.autocorrect) {
+        sectionTitle = "Autocorrect: Done";
+    } else if (autocorrect.isDidYouMean) {
+        sectionTitle = "Did you mean? Use `-a` to autocorrect";
```
froydnj, line 147: I think "Did you mean: $X" works for the way this was previously as part of the suggestion for each replacement, but it seems a little more awkward when as a section header.  Can we special-case `autocorrect.edits.size() == 1` so we can make the section title here "Did you mean `$X`?  Use `-a` to autocorrect"?  (or is that too verbose?)

If we went that route, it seems slightly better to me to say something different for the `size() != 1` case, but I don't know what the right thing is.  "Possible fixes below"?

froydnj (approved):
This is an improvement!  Suggestions below.

PR #2762, opened 2020-03-12
Use kvstore in LSP to optimize startup performance.

jvilk-stripe:
Use kvstore in LSP to optimize startup performance. When LSP starts up, it has to parse and hash every file in the project. This cost currently dominates startup time on large projects.

With this change, when --cache-dir and --lsp are supplied to Sorbet, Sorbet will store the following on disk in the kvstore immediately after initialization:

* The same contents as the CLI version (parsed ASTs and GlobalState).
* Hashes for files

Sorbet does not touch the cache-dir for the rest of execution.

LSP is compatible with the CLI's kvstore, so the same --cache-dir can be supplied to both LSP and CLI modes. The extra hashes are stored in keys that are uniquely used only in LSP mode.

The implementation involved changing KeyValueStore so that it's possible to transfer ownership between threads, which is necessary to support kvstore interactions in `realmain` and `LSPIndexer`. Specifically:

* All access to `KeyValueStore` must be mediated via an `OwnedKeyValueStore`.
* `OwnedKeyValueStore` can be downgraded (via `commit` or `abort`) to a `KeyValueStore`, which also serves to explicitly commit or abort writes made to the store.

The end result of this optimization: A huge decrease in startup latency for Sorbet LSP (an order of magnitude decrease) when the project is cached on disk.

### Motivation

Mainly, to speed up LSP initialization!

Sorbet's CLI supports caching ASTs on disk with --cache-dir, and earlier today I landed logic for also serializing hashes for files. It made sense to connect this logic with LSP when --cache-dir is supplied.

### Test plan

See included automated tests.

PR #2760, opened 2020-03-11
Store FileHashes in payload

jvilk-stripe:
Serialize FileHashes into cache, and store FileHashes in payload.

Does not yet modify LSP to use cache. I have those changes locally, but I'm landing that as a follow-on change to de-risk it.

There's currently a mystery failure; the store-state test fails because the cached state changes between runs. It's possible that NameHashes have some nondeterminism. I'm looking into it now.

### Motivation

The biggest win from this PR alone is significantly speeding up LSP-based tests (order of magnitude speedup):

Before on my mac:

```
$ bazel test //test/lsp:protocol_test_corpus --config=dbg 
//test/lsp:protocol_test_corpus                                          PASSED in 118.9s
```

After: 

```
$ bazel test //test/lsp:protocol_test_corpus --config=dbg 
//test/lsp:protocol_test_corpus                                          PASSED in 13.1s
```

There are big savings on test corpus LSP tests, too, which spent most of their time hashing.

### Test plan

Automated tests already exist.

PR #5591, opened 2022-04-06
Avoid re-running autogen based on constant cache

aisamanra:
### Motivation
This uses a constant hash along with a list of changed files to decide whether to run autogen or not. There are two new behaviors in autogen mode that are added here:

- if `--autogen-constant-cache-file` is provided _when other autogen flags are provided_, then it'll compute the constant hash of _all_ input files and pack them into the provided path in a msgpack format, and then runs autogen as usual. There is no way to compute the constant cache file outside of autogen: this simply should never come up in any other code path.
- if `--autogen-constant-cache-file` is provided _and_ one or more files are provided with `--autogen-changed-files`, then before doing any autogen work we'll first compute the constant hash for only those files listed, and if all of them compute to the same constant hash that they map to in the file itself, we simply exit without doing any other work. We should be pretty cautious about this: we will go on to do a full autogen run if any of the following are true:
    - There is no cache file present, or the cache file is unreadable in any way
    - Any of the files listed are not present in the cache file
    - There are more than a fixed number of changes (currently 100)
    - Any of the files are non-readable or have other issues

This more or less adds a "fast path" for autogen that allows us to duck out early if we can show that the intermediate edits would not have affected the output.

This technically also rewrites the constant hash from how it exists. It previously was a recursive function defined on `parser::Node` values, and now it uses a `ShallowTreeMap` to work on the post-desugared AST. The problem with the previous one was that working it into the pipeline in order to compute the constant hash of the codebase was _very_ tricky and invasive, since it had to be worked into the middle of the `indexOne` function during the roughly two-line span of code where we have a `parser::Node` before it's turned into a desugared tree. That plumbing was invasive and error-prone, so in order to move the computation to the early part of autogen and minimize disruption to the existing path, I've allowed it to operate over our usual AST so it could get computed early in autogen.

### Test plan

See included automated tests. Added a CLI test which demonstrates the behavior we want.


aisamanra:

→ https://go/builds/bui_LSlUW9sytVYpxY
→ https://go/builds/bui_LSlUVZvXz3E0BY

aisamanra:
I think this almost definitely needs some tweaking before landing, but I want to get it up for review so I can get feedback. The biggest sticking point right now is that @nroman-stripe thinks we should switch it over to using a `KVStore` instead of the custom-written msgpack format we have here. I'm not opposed to that, but it's a bit subtle—in particular, we can only have one `KVStore` open at a time, and we currently have an existing one open while we're doing the exit-early check. If we feel strongly that we should use the `KVStore` instead of this custom format, I can think of ways of tackling this, but I figured it was worth getting eyes on it before starting down that road.

nroman-stripe:
Just to be clear, I'm not a blocker on that KVStore. It just seemed suited to the task, but I wasn't fully aware of the complications involved.

jez:
Sorry, I didn't get a chance to look at this today. On the cache thing: could we store the autogen caches on the core::File? We already do this for the LSP file hashes.

jez:
You can read more about that in these PRs:

#2760
#2762

jez:
this looks ~fine to me? it is pretty well comtained to not LSP and not the core pipeline and doesn't introduce new external dependencies. Going to hand off review to nroman for the actual implementation review.

main/pipeline/pipeline.cc
```diff
@@ -1259,4 +1261,45 @@ bool cacheTreesAndFiles(const core::GlobalState &gs, WorkerPool &workers, vector
     return written;
 }
 
+vector<ast::ParsedFile> autogenCacheFiles(const core::GlobalState &gs, string_view cachePath,
```
nroman-stripe, line 1264: I think a better name for this would be something like `autogenWriteCacheFile` or something similarly indicative.

main/pipeline/pipeline.cc
```diff
@@ -1259,4 +1261,45 @@ bool cacheTreesAndFiles(const core::GlobalState &gs, WorkerPool &workers, vector
...
+                                          vector<ast::ParsedFile> what, WorkerPool &workers) {
+#ifndef SORBET_REALMAIN_MIN
+    Timer timeit(gs.tracer(), "autogenConstantCache");
```
nroman-stripe, line 1267: less confusing if this matches the func name

main/autogen/cache.h
```diff
@@ -0,0 +1,45 @@
...
+            cache._constantHashMap.insert(c.constantHashMap().begin(), c.constantHashMap().end());
+        }
+
+        return cache;
+    }
+
+    AutogenCache(UnorderedMap<std::string, unsigned int> constantHashMap) : _constantHashMap(constantHashMap){};
+    AutogenCache() = default;
+
+    void add(std::string path, unsigned int hash) {
```
nroman-stripe, line 34: What do you think about this API using a `FileRef` instead of string for path? Then `pack()` could take a GS and stream the strings into the buffer at the last second. I realize that slightly complicates things when reading for the can skip check, but I think in that case since we never re-use or pack that cache it could stay as a `UnorderedMap<string, unint32_t>`

main/autogen/cache.h
```diff
@@ -0,0 +1,45 @@
...
+        _constantHashMap.emplace(path, hash);
```
nroman-stripe, line 35: in the name of densive programming should this ping a thread_id and ENFORCE that add's come from it?

nroman-stripe (changes requested):
Looks legit. Some minor nits and suggestions.

main/autogen/cache.h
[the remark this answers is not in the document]
aisamanra: I did this mostly because it was what the hash code was already doing and it felt consistent, c.f. https://github.com/sorbet/sorbet/blob/master/core/hashing/hashing.h#L10-L14

nroman-stripe (approved):
Thanks!

PR #5592, opened 2022-04-06
Add rename menu after the "Move method" refactoring

ilyailya:
### Motivation

The change adds a rename textfield after the refactoring have been applied. 
By the LSP spec, you can provide a command in the `CodeAction`, and it will be executed after all edits have been applied. 

```
export interface CodeAction {
	...
	
	/**
	 * A command this code action executes. If a code action
	 * provides an edit and a command, first the edit is
	 * executed and then the command.
	 */
	command?: Command;
}
```
I'm using this property of the code action to invoke the `sorbet.rename` command, which is also have been added in this change. The `sorbet.rename` commands calls the `editor.action.rename` command, which executes all needed requests to rename a symbol.

Unfortunately we can't call an `editor.action.rename` command directly, for more details please check here https://github.com/microsoft/vscode/issues/146767

https://user-images.githubusercontent.com/9540120/162059427-bf639cae-29c6-4dc4-930b-cc4894cc5a29.mov

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/code_action.cc
+++ b/main/lsp/requests/code_action.cc
@@ -142,15 +142,27 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerInterface &
                     }
                     auto action = make_unique<CodeAction>("Move method to a new module");
                     action->kind = CodeActionKind::RefactorExtract;
+
                     bool canResolveLazily = config.getClientConfig().clientCodeActionResolveEditSupport &&
                                             config.getClientConfig().clientCodeActionDataSupport;
+                    auto newModuleLoc = getNewModuleLocation(gs, *def, typechecker);
+                    auto renameCommand = make_unique<Command>("Rename Symbol", "sorbet.rename");
+                    auto arg = make_unique<TextDocumentPositionParams>(
+                        make_unique<TextDocumentIdentifier>(params->textDocument->uri), move(newModuleLoc));
+                    auto args = vector<unique_ptr<TextDocumentPositionParams>>();
+                    args.emplace_back(move(arg));
+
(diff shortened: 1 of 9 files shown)


main/lsp/requests/code_action.cc
```diff
@@ -142,15 +142,27 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerInterface &
                     }
                     auto action = make_unique<CodeAction>("Move method to a new module");
                     action->kind = CodeActionKind::RefactorExtract;
+
                     bool canResolveLazily = config.getClientConfig().clientCodeActionResolveEditSupport &&
                                             config.getClientConfig().clientCodeActionDataSupport;
+                    auto newModuleLoc = getNewModuleLocation(gs, *def, typechecker);
+                    auto renameCommand = make_unique<Command>("Rename Symbol", "sorbet.rename");
```
jez, line 149: What happens if the client doesn't support this command (e.g., old VS Code extension version, or using Vim or some other language client)

jez (approved):
One question about backwards compatibility / compatibility with other language clients.

main/lsp/requests/code_action.cc
ilyailya: It depends on the client, VSCode will show a warning popup like this, but the edits will be applied.
Screen Shot 2022-04-07 at 14 53 48

NeoVim's LSP client just ignores the command

https://user-images.githubusercontent.com/9540120/162217207-b31c4625-829c-4750-a278-4c8f58f23274.mov

jez: Perfect thanks

PR #5594, opened 2022-04-06
Fix locations for packager errors

jez:
### Motivation

It looks bad to print the full class def into the error message.

### Test plan

See updated exp files


froydnj (approved):
++

froydnj (approved):

PR #5596, opened 2022-04-07
Immutable version of Packager::runIncremental

jez:
### Commit summary

- **Add a test that fails before these changes** (da637be99)

- **pre-work: Refactor Resolve::runIncremental into two explicit functions** (e016114d3)

- **pre-work: `...WithoutStateMutation` -> `...BestEffort`** (343d52a6d)

- **Immutable version of Packager::runIncremental** (ad505ab78)

  Honestly behaves like a hybrid of namer ("drop trees we don't want") and
  resolver ("use templates to do one thing or another conditionally").

### Motivation

Honestly behaves like a hybrid of namer ("drop trees we don't want") and
resolver ("use templates to do one thing or another conditionally").

For posterity's sake, here are those two previous PRs: #5498, #5532.

### Test plan

- [x] @jez This is not tested yet.

### DIFF
--- a/packager/packager.h
+++ b/packager/packager.h
@@ -84,6 +84,12 @@ public:
 
     // Run packager incrementally. Note: `files` must contain all packages files. Does not support package changes.
     static std::vector<ast::ParsedFile> runIncremental(core::GlobalState &gs, std::vector<ast::ParsedFile> files);
+    // Run packager incrementally, but withouut mutating GlobalState. This is used by LSP to serve
+    // certain kinds of requests using stale data. In this mode, if mutating GlobalState would have
+    // been required to make progress, the corresponding part of the tree that triggered the
+    // mutation is simply dropped.
+    static std::vector<ast::ParsedFile> runIncrementalBestEffort(const core::GlobalState &gs,
+                                                                 std::vector<ast::ParsedFile> files);
 
     // The structures created for `__package.rb` files from their imports are large and deep. This causes
     // performance problems with typechecking. Use to remove these modules while retaining the PackageSpec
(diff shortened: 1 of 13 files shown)


jez:

→ https://go/builds/bui_LShtEgPKFz5fbf
→ https://go/builds/bui_LShtoXafxPwL2e

packager/packager.h
```diff
@@ -83,7 +83,8 @@ class Packager final {
                                             std::vector<ast::ParsedFile> files);
 
     // Run packager incrementally. Note: `files` must contain all packages files. Does not support package changes.
-    static std::vector<ast::ParsedFile> runIncremental(core::GlobalState &gs, std::vector<ast::ParsedFile> files);
+    template <typename StateType>
+    static std::vector<ast::ParsedFile> runIncremental(StateType &gs, std::vector<ast::ParsedFile> files);
```
froydnj, line 87: Slight preference for being explicit here with two different declarations of `runIncremental` and leaving the template magic in the `.cc` file: it's easier to catch mistakes in client code that way (e.g. the current state would happily accept you passing in a `unique_ptr<GlobalState>`...or a `MutableCtx`, which would normally auto-convert to `GlobalState`, but not here.  you wouldn't see an error until link time).

(The proposed scheme here is also inconsistent with namer and resolver: the non-mutating versions of those have different names, whereas packager would not.  I tend to think having identical names is not a good thing for what we're trying to do, even if C++ is happy to overload on `const` qualification.)

froydnj (approved):
Nice catch, I keep forgetting about package magic.

PR #5598, opened 2022-04-07
notify the user when they jump-to-def in untyped files

froydnj:
### Motivation

Feedback when Sorbet doesn't do what you expect is a good thing.

I think the little window that pops up in vscode is better than nothing, though unfortunately the window is not transient; you have to manually click it away.

### Test plan

See included automated tests.


froydnj:
Experimenting says that hitting Escape in vscode dismisses the message, so I think it would be reasonable to push this forward.

froydnj:
> What do you think about showing the message and also jumping to the typed sigil’s location at the top of the file?

The idea here is:

a) the user can just pop the stack (Control-minus in VSCode) to get back to where they were, so moving them around in the file is not a huge imposition; and
b) visual movement in the file + notification is more visible than just the notification alone?

It would feel jarring to me, but I am also willing to believe that angling for just the notification is just me projecting my preferences onto everybody else.  Maybe I should record a screencast and put up a poll inside Stripe and/or here.

jez:
I was hoping that if we put the user's cursor right at the end of the `# typed: false` sigil, then it's super easy for them to flip it to `typed: true`, press `^-`, and do the jump-to-def again.

jez:
Another option would be to change the showMessage to a showMessageRequest with a button that people can click to have Sorbet automatically upgrade the file to `typed: true`. I actually really like this, though it would be the first window/showMessageRequest (the first request + response pair where the server sends the request and the client writes the response).

jez:
I'm also not sure what language clients like Vim do when presented with a showMessageRequest message.

froydnj:
> Another option would be to change the showMessage to a showMessageRequest with a button that people can click to have Sorbet automatically upgrade the file to `typed: true`. I actually really like this, though it would be the first window/showMessageRequest (the first request + response pair where the server sends the request and the client writes the response).

This is really useful, but from looking at Sorbet's current LSP implementation, we have the following issues:

1. we don’t support receiving response messages from the client: https://github.com/sorbet/sorbet/blob/f4f47246a9ca293f9e1256e51c060ce1a801c90a/main/lsp/LSPPreprocessor.cc#L151-L156
2. we don’t support sending request messages to the client.  the actual mechanics of JSON and whatnot are trivial, but we would have to have some map of request IDs that we’ve sent off for matching up with receiving responses (above)…probably with some policy of how to drop requests on the floor if the client hasn’t responded within a set amount of time (?)
3. as a result of needing a map of request IDs, we need someplace to store that/them, and I don’t think the layering of `LSPLoop`/`LSPTypechecker`/`LSPPreprocessor`/`LSPConfiguration`/etc. is quite right for supporting that, so there would be some pre-work there
4. probably some other stuff I have not yet foreseen--there might some issues with parsing responses given that they are a little more freeform, too

These are all solvable, and having the capability of using `window/showMessageRequest` would be super-useful, but I think that implementing all of that just for a slightly better story around jump-to-def on untyped files is too big of a lift for now.

jez (commented):
still wip, feel free to re-request when ready

jez (approved):
This is awesome, thanks!

PR #5603, opened 2022-04-07
Add note for how to update certain exp files

jez:
### Motivation

Easier for new contributors to figure out what's up.

### Test plan

test-only change

Screen Shot 2022-04-07 at 12 03 15 PM
Screen Shot 2022-04-07 at 12 02 23 PM

### DIFF
--- a/test/pipeline_test_runner.cc
+++ b/test/pipeline_test_runner.cc
@@ -139,7 +139,12 @@ public:
             auto checker = test.folder + expectation->second.begin()->second;
             auto expect = FileOps::read(checker);
 
-            CHECK_EQ_DIFF(expect, gotPhase.second, fmt::format("{}Mismatch on: {}", prefix, checker));
+            CHECK_EQ_DIFF(expect, gotPhase.second,
+                          fmt::format("{}Mismatch on: {}\n"
+                                      "\n"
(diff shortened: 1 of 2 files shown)


test/pipeline_test_runner.cc
```diff
@@ -139,7 +139,12 @@ class ExpectationHandler {
             auto checker = test.folder + expectation->second.begin()->second;
             auto expect = FileOps::read(checker);
 
-            CHECK_EQ_DIFF(expect, gotPhase.second, fmt::format("{}Mismatch on: {}", prefix, checker));
+            CHECK_EQ_DIFF(expect, gotPhase.second,
+                          fmt::format("{}Mismatch on: {}\n"
+                                      "\n"
+                                      "If these changes are expected, run this script and commit the results:\n"
+                                      "tools/scripts/update_testdata_exp.sh\n",
```
froydnj, line 146:
```suggestion
                                      "./tools/scripts/update_testdata_exp.sh\n",
```

froydnj (approved):

test/pipeline_test_runner.cc
jez: The `./` is not required, because there is already a slash in the command (`tools/...`).

PR #5604, opened 2022-04-07
Only run checkExpectations once

jez:
### Motivation

checkExpectations will check **all** expectations. We were running it
twice for every exp kind except the `autocorrects` exp kind.

The diff is weird (it shows me as moving the "check warnings and errors"
code down below), but in my head this was moving the "Check
autocorrects" section up.

If we move things around like this, then by the time we get to
checkExpectations the first time, the autocorrects expectations will be
populated, so we don't have to run checkExpectations as second time.

### Test plan

test-only change. i verified that failures in autocorrect exp files are still
checked.

### DIFF
--- a/test/pipeline_test_runner.cc
+++ b/test/pipeline_test_runner.cc
@@ -128,8 +128,8 @@ public:
         }
     }
 
-    void checkExpectations(string prefix = "") {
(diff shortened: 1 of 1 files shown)


elliottt (approved):
Nice find!

PR #5610, opened 2022-04-08
Remove newlines and excess whitespace from autocorrects

jez:
### Motivation

When printing the messages into the terminal or into error messages in VS
Code, the extra whitespace is honestly more trouble than it's worth.

For a while I was pretty insistent that the autocorrect show you exactly
what it's going to do to your file, which mean showing the contents
unchanged. But honestly, people are just going to see that there's an
autocorrect and take it, and then undo the change if it turns out to have
not been a good one.

### Test plan

Exp file changes

### DIFF
--- a/core/Error.cc
+++ b/core/Error.cc
@@ -1,6 +1,8 @@
 // has to go first as it violates our poisons
 #include "rang.hpp"
 
+#include "absl/strings/escaping.h"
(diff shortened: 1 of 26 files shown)


elliottt (approved):
Nice, these look much better!

PR #5611, opened 2022-04-08
Tidy up packager import suggestions

jez:
### Motivation

- Some of the suggestions were redundant. In particular, if there was already an
  autocorrect, we don't need another `ErrorLine` saying the same thing.
- There was a lot of visual noise. Especially when the constant names and file
  names are huge, that distracts from the core message.

This should hopefully cut down on some of the chunder.

I was following this build as inspiration for what it looks like right now:

<http://go/builds/bui_LTKnw0Jy0zpU6h>

And this is what the same errors look like after these changes:

<http://go/builds/bui_LTNleqlT4v7Taf>

### Test plan

exp files

### DIFF
--- a/resolver/SuggestPackage.cc
+++ b/resolver/SuggestPackage.cc
@@ -82,52 +82,39 @@ public:
             addMissingExportForTestSuggestion(e, match);
             return;
         }
-        vector<core::ErrorLine> lines;
         auto &srcPkg = db().getPackageInfo(match.srcPkg);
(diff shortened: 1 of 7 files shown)


jez:

→ https://go/builds/bui_LTNMbDByG9ZoTt
→ https://go/builds/bui_LTNMX6oTvAgAxH

elliottt (approved):

PR #5613, opened 2022-04-09
structure C-level strings as a string table, rather than individual string constants

froydnj:
### Motivation

This change lays some groundwork for not generating code for interning individual IDs in each compiled module.  Instead, once we have this string table, we can generate a table for IDs that looks something like (C++ code):

```c++
// In the compiled module:
static const size_t numIDs = ...;
static struct IDDescriptor {
  uint32_t offset;
  uint32_t length;
} idDescriptors[numIDs] = {...};

static ID idTable[numIDs];

// sorbet_moduleStringTable defined via CompilerState

// Later, in initialization of the module.  One call for all IDs instead of
// N calls for N IDs.
sorbet_intern_ids(idTable, idDescriptors, numIDs, sorbet_moduleStringTable);

// In the VM payload:
void sorbet_intern_ids(ID *idTable, const IDDescriptor *idDescriptors, size_t numIDs, const char *stringTable) {
  for (size_t i = 0; i < numIDs; ++i) {
    const IDDescriptor &desc = idDescriptors[i];
    idTable[i] = rb_intern(&stringTable[desc.offset], desc.length);
  }
}
```

which creates a lot less IR for LLVM to optimize in the actual compiled module.

And we can do the same thing for Ruby-level strings, which also cuts down on the amount of LLVM IR we generate.

### Test plan

See included automated tests.  The only complication here is that certain calls will no longer see the string constant, but rather some address in the string table, so a few filecheck assertions need to be updated (see `method.rb` and `static_method.rb`).

### DIFF
--- a/test/testdata/compiler/all_arguments.opt.ll.exp
+++ b/test/testdata/compiler/all_arguments.opt.ll.exp
@@ -83,37 +83,23 @@ target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16
 @.str.10 = private unnamed_addr constant [93 x i8] c"sorbet_getIsReleaseBuild: Shared objects compiled by sorbet_llvm must be run by sorbet_ruby.\00", align 1
 @"stackFramePrecomputed_func_Object#14take_arguments" = internal unnamed_addr global %struct.rb_iseq_struct* null, align 8
 @rubyIdPrecomputed_take_arguments = internal unnamed_addr global i64 0, align 8
-@str_take_arguments = private unnamed_addr constant [15 x i8] c"take_arguments\00", align 1
 @rubyStrFrozen_take_arguments = internal unnamed_addr global i64 0, align 8
 @"rubyStrFrozen_test/testdata/compiler/all_arguments.rb" = internal unnamed_addr global i64 0, align 8
-@"str_test/testdata/compiler/all_arguments.rb" = private unnamed_addr constant [40 x i8] c"test/testdata/compiler/all_arguments.rb\00", align 1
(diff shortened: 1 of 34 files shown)


test/testdata/compiler/all_arguments.opt.ll.exp
```diff
@@ -129,14 +115,14 @@ target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16
...
 @rubyIdPrecomputed_baz = internal unnamed_addr global i64 0, align 8
-@str_baz = private unnamed_addr constant [4 x i8] c"baz\00", align 1
 @ic_take_arguments.14 = internal global %struct.FunctionInlineCache zeroinitializer
 @ic_take_arguments.15 = internal global %struct.FunctionInlineCache zeroinitializer
 @ic_take_arguments.16 = internal global %struct.FunctionInlineCache zeroinitializer
 @ic_take_arguments.17 = internal global %struct.FunctionInlineCache zeroinitializer
 @ic_take_arguments.18 = internal global %struct.FunctionInlineCache zeroinitializer
 @ic_take_arguments.19 = internal global %struct.FunctionInlineCache zeroinitializer
 @ic_take_arguments.20 = internal global %struct.FunctionInlineCache zeroinitializer
+@sorbet_moduleStringTable = internal unnamed_addr constant [138 x i8] c"take_arguments\00test/testdata/compiler/all_arguments.rb\00g\00d\00e\00<build-array>\00inspect\00puts\00<top (required)>\00normal\00Object\00a\00b\00c\00f\00baz\00master\00", align 1
```
elliottt, line 125: This is awesome!

elliottt (approved):
This is really great, I love seeing all those individual string constants disappearing :)

PR #5614, opened 2022-04-10
shrink `TypePtr` to 8 bytes by using intrusive refcounting

froydnj:
### Motivation

One small step to 8-byte `TypePtr`.  I kind of want to make 32-bit and 64-bit integer literals their own separate types, but one step at a time.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -1343,7 +1343,8 @@ core::NameRef Literal::asSymbol() const {
 
 bool Literal::isSymbol() const {
     return core::isa_type<core::NamedLiteralType>(value) &&
-           core::cast_type_nonnull<core::NamedLiteralType>(value).literalKind == core::NamedLiteralType::LiteralTypeKind::Symbol;
+           core::cast_type_nonnull<core::NamedLiteralType>(value).literalKind ==
+               core::NamedLiteralType::LiteralTypeKind::Symbol;
 }
 
 bool Literal::isNil(const core::GlobalState &gs) const {
@@ -1352,7 +1353,8 @@ bool Literal::isNil(const core::GlobalState &gs) const {
 
 bool Literal::isString() const {
     return core::isa_type<core::NamedLiteralType>(value) &&
-           core::cast_type_nonnull<core::NamedLiteralType>(value).literalKind == core::NamedLiteralType::LiteralTypeKind::String;
+           core::cast_type_nonnull<core::NamedLiteralType>(value).literalKind ==
+               core::NamedLiteralType::LiteralTypeKind::String;
 }
 
 bool Literal::isTrue(const core::GlobalState &gs) const {
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -812,6 +812,7 @@ string Literal::toStringWithTabs(const core::GlobalState &gs, int tabs) const {
     typecase(
         this->value, [&](const core::LiteralType &l) { res = l.showValue(gs); },
         [&](const core::LiteralIntegerType &i) { res = i.showValue(gs); },
+        [&](const core::FloatLiteralType &i) { res = i.showValue(gs); },
         [&](const core::ClassType &l) {
             if (l.symbol == core::Symbols::NilClass()) {
                 res = "nil";
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -810,7 +810,7 @@ string Literal::showRaw(const core::GlobalState &gs, int tabs) {
 string Literal::toStringWithTabs(const core::GlobalState &gs, int tabs) const {
     string res;
     typecase(
-        this->value, [&](const core::LiteralType &l) { res = l.showValue(gs); },
+        this->value, [&](const core::NamedLiteralType &l) { res = l.showValue(gs); },
         [&](const core::IntegerLiteralType &i) { res = i.showValue(gs); },
         [&](const core::FloatLiteralType &i) { res = i.showValue(gs); },
         [&](const core::ClassType &l) {
@@ -1329,21 +1329,21 @@ string Literal::nodeName() {
 
 core::NameRef Literal::asString() const {
     ENFORCE(isString());
-    auto t = core::cast_type_nonnull<core::LiteralType>(value);
+    auto t = core::cast_type_nonnull<core::NamedLiteralType>(value);
     core::NameRef res = t.asName();
     return res;
 }
 
 core::NameRef Literal::asSymbol() const {
     ENFORCE(isSymbol());
-    auto t = core::cast_type_nonnull<core::LiteralType>(value);
+    auto t = core::cast_type_nonnull<core::NamedLiteralType>(value);
     core::NameRef res = t.asName();
     return res;
 }
 
 bool Literal::isSymbol() const {
-    return core::isa_type<core::LiteralType>(value) &&
-           core::cast_type_nonnull<core::LiteralType>(value).literalKind == core::LiteralType::LiteralTypeKind::Symbol;
+    return core::isa_type<core::NamedLiteralType>(value) &&
+           core::cast_type_nonnull<core::NamedLiteralType>(value).literalKind == core::NamedLiteralType::LiteralTypeKind::Symbol;
 }
 
 bool Literal::isNil(const core::GlobalState &gs) const {
@@ -1351,8 +1351,8 @@ bool Literal::isNil(const core::GlobalState &gs) const {
 }
 
 bool Literal::isString() const {
-    return core::isa_type<core::LiteralType>(value) &&
-           core::cast_type_nonnull<core::LiteralType>(value).literalKind == core::LiteralType::LiteralTypeKind::String;
+    return core::isa_type<core::NamedLiteralType>(value) &&
+           core::cast_type_nonnull<core::NamedLiteralType>(value).literalKind == core::NamedLiteralType::LiteralTypeKind::String;
 }
 
 bool Literal::isTrue(const core::GlobalState &gs) const {
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -811,7 +811,7 @@ string Literal::toStringWithTabs(const core::GlobalState &gs, int tabs) const {
     string res;
     typecase(
         this->value, [&](const core::LiteralType &l) { res = l.showValue(gs); },
-        [&](const core::LiteralIntegerType &i) { res = i.showValue(gs); },
+        [&](const core::IntegerLiteralType &i) { res = i.showValue(gs); },
         [&](const core::FloatLiteralType &i) { res = i.showValue(gs); },
         [&](const core::ClassType &l) {
             if (l.symbol == core::Symbols::NilClass()) {
--- a/ast/Trees.cc
+++ b/ast/Trees.cc
@@ -811,6 +811,7 @@ string Literal::toStringWithTabs(const core::GlobalState &gs, int tabs) const {
     string res;
     typecase(
         this->value, [&](const core::LiteralType &l) { res = l.showValue(gs); },
+        [&](const core::LiteralIntegerType &i) { res = i.showValue(gs); },
         [&](const core::ClassType &l) {
             if (l.symbol == core::Symbols::NilClass()) {
                 res = "nil";
--- a/core/TypePtr.cc
+++ b/core/TypePtr.cc
@@ -21,6 +21,7 @@ using namespace std;
         CASE_STATEMENT(CASE_BODY, SelfType)              \
         CASE_STATEMENT(CASE_BODY, LiteralType)           \
         CASE_STATEMENT(CASE_BODY, LiteralIntegerType)    \
+        CASE_STATEMENT(CASE_BODY, FloatLiteralType)      \
         CASE_STATEMENT(CASE_BODY, TypeVar)               \
         CASE_STATEMENT(CASE_BODY, OrType)                \
         CASE_STATEMENT(CASE_BODY, AndType)               \
@@ -115,6 +116,8 @@ int TypePtr::kind() const {
             return 12;
         case Tag::LiteralIntegerType:
             return 13;
+        case Tag::FloatLiteralType:
+            return 14;
     }
 }
 
@@ -137,6 +140,7 @@ bool TypePtr::isFullyDefined() const {
         case Tag::ClassType:
         case Tag::LiteralType:
         case Tag::LiteralIntegerType:
+        case Tag::FloatLiteralType:
         case Tag::AliasType:
         case Tag::SelfTypeParam:
         case Tag::MetaType: // MetaType: this is kinda true but kinda false. it's false for subtyping but true for
@@ -177,6 +181,7 @@ bool TypePtr::hasUntyped() const {
         case Tag::TypeVar:
         case Tag::LiteralType:
         case Tag::LiteralIntegerType:
+        case Tag::FloatLiteralType:
         case Tag::SelfType:
         case Tag::AliasType:
         case Tag::SelfTypeParam:
@@ -234,7 +239,8 @@ TypePtr TypePtr::getCallArguments(const GlobalState &gs, NameRef name) const {
         case Tag::TupleType:
         case Tag::ShapeType:
         case Tag::LiteralType:
-        case Tag::LiteralIntegerType: {
+        case Tag::LiteralIntegerType:
+        case Tag::FloatLiteralType: {
             return this->underlying(gs).getCallArguments(gs, name);
         }
         case Tag::OrType: {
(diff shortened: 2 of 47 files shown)


froydnj:

→ https://go/builds/bui_LgloH1ZMlVCqlg
→ https://go/builds/bui_LglpuHzpkczId7

froydnj:

→ https://go/bui_Lii9KpmQcWyHYn
→ https://go/builds/pay-server-sorbet-check--bui_Lii9QK3hJ9ebA0

froydnj:

→ https://go/builds/bui_LsggXsAghgmCfv
→ https://go/builds/bui_LsggVftWPOO9LS

froydnj:

→ https://go/builds/bui_Lvh18BIFtfWytm
→ https://go/builds/bui_Lvh1g8Yn3IVanO

froydnj:
I think this is basically ready for review; you should be able to review commit by commit.

The only two interesting things I can think of doing are:

1. getting rid of `LiteralType` in favor of separate literal string and literal symbol types, which I'm not sure is worth it; and
2. Having a `SmallLiteralIntegerType` for ~33-bit integers (something > 32 so we can accommodate 32-bit unsigned constants; we can go as high as 46, but that seems a little excessive?) or so that lets them be stored as inline values.  This is probably worth it, but the speed and memory savings of this PR as-is are significant enough that inline small integers could be done as a followup.

Will also run an emscripten build locally to make sure the bit-twiddling that comes with this PR didn't botch something on 32-bit.

This change is good for ~3-4% wall time speedup typechecking Stripe's codebase with a warm cache and ~5% max RSS reduction.

ast/Trees.cc
```diff
@@ -811,6 +811,8 @@ string Literal::toStringWithTabs(const core::GlobalState &gs, int tabs) const {
     string res;
     typecase(
         this->value, [&](const core::LiteralType &l) { res = l.showValue(gs); },
```
jez, line 813: Can we replace `LiteralType` with something like `LiteralNamedType` or something?

I don't want to accidentally think that `LiteralType` still represents all literals and type out `LiteralType` out of habit.

ast/Trees.cc
```diff
@@ -811,6 +811,8 @@ string Literal::toStringWithTabs(const core::GlobalState &gs, int tabs) const {
...
+        [&](const core::LiteralIntegerType &i) { res = i.showValue(gs); },
+        [&](const core::FloatLiteralType &i) { res = i.showValue(gs); },
```
jez, line 815: Literal Integer

vs

Float Literal

=> Maybe we should pick a consistent order?

core/types/calls.cc
```diff
@@ -498,7 +514,7 @@ TypePtr unwrapType(const GlobalState &gs, Loc loc, const TypePtr &tp) {
             unwrappedElems.emplace_back(unwrapType(gs, loc, elem));
         }
         return make_type<TupleType>(move(unwrappedElems));
-    } else if (isa_type<LiteralType>(tp)) {
+    } else if (isa_type<LiteralType>(tp) || isa_type<LiteralIntegerType>(tp) || isa_type<FloatLiteralType>(tp)) {
```
jez, line 517: should we make a helper for this on `TypePtr`?

core/types/hashing.cc
```diff
@@ -33,19 +33,34 @@ uint32_t LiteralType::hash(const GlobalState &gs) const {
...
+    auto underlying = this->underlying(gs);
+    ClassOrModuleRef undSymbol = cast_type_nonnull<ClassType>(underlying).symbol;
+    result = mix(result, undSymbol.id());
+
+    uint64_t rawValue = absl::bit_cast<uint64_t>(this->value);
+    uint32_t topBits = static_cast<uint32_t>(rawValue >> 32);
+    uint32_t bottomBits = static_cast<uint32_t>(rawValue & 0xFFFFFFFF);
+    result = mix(result, topBits);
+    result = mix(result, bottomBits);
+    return result;
```
jez, line 54: I wonder if we should factor this 64-bit hashing logic out to a helper function?

I wonder if there are places where we currently attempt to pass a 64 bit value to `mix` and it just does something weird.

test/testdata/infer/infer1.rb
```diff
@@ -32,7 +32,7 @@ def baz6(cond)
  else
    b = "foo"
  end
- b = "foo".getbyte(b) # error: Expected `Integer` but found `T.any(Integer, String)` for argument `arg0`
+ b = "foo".getbyte(b) # error: Expected `Integer` but found `T.any(String, Integer)` for argument `arg0`
```
jez, line 35: Do you have a sense of why this changed?

froydnj: I believe it is because we sort on `kind` somewhere, and because the PR added:

https://github.com/sorbet/sorbet/pull/5614/files#diff-8a69c6e75a3c53920d6c4a7e8b59337d5c99e35ed53d3b83b9fc0bb821726d00R117

core/TypePtr.cc
```diff
@@ -112,6 +114,10 @@ int TypePtr::kind() const {
             return 11;
         case Tag::SelfType:
             return 12;
+        case Tag::LiteralIntegerType:
+            return 13;
+        case Tag::FloatLiteralType:
+            return 14;
```
jez, line 120: Do you mind if we renumber the tags to keep all the `LiteralType` things together?

I was looking into some generics bugs a while back, and one of the fixes I had planned was based on assuming that the `TypeVar` and `LambdaParam` things come between the ground types and the type constructors like And and Or.

Also selfishly I prefer `T.any(Integer, String)` to `T.any(String, Integer)` (an ordering we could get back by moving back up).

I _think_ I can write some tests involving generics that would require these literal types to be moved up earlier, but I haven't really spent time on that branch in a while, and would love not to have to spend to much time thinking of test cases right now.

I can add a note to my other branch ([code here](https://github.com/sorbet/sorbet/compare/jez-no-short-circuit-type-constraints)) to write tests involving all literal types when i get around to reviving it.

froydnj:
I think I've addressed all the feedback thus far.  `NamedLiteralType` feels clunky, but `SymOrStrLiteralType` also felt clunky.  It's one regex substitution to fix, though, so not difficult to change.

I think the `${X}LiteralType` is better than the `Literal${X}Type` naming, so I went with that.

core/TypePtr.h
```diff
@@ -13,6 +13,23 @@ class TypeConstraint;
...
+public:
+    void addref() {
+        this->counter.fetch_add(1);
+    }
+
+    uint32_t release() {
+        // fetch_sub returns value prior to subtract
+        return this->counter.fetch_sub(1) - 1;
+    }
+};
```
jez, line 31: idk if you care about having a `CheckSize` here. It seems maybe useful as a point of documentation that this must be 4 bytes so that the other half of the 8 bytes can fit a NameRef

core/Types.h
```diff
@@ -356,7 +360,7 @@ template <> inline ClassType cast_type_nonnull<ClassType>(const TypePtr &what) {
  * This is the type used to represent a use of a type_member or type_template in
  * a signature.
  */
-TYPE(LambdaParam) final {
+TYPE(LambdaParam) final : public Refcounted {
```
jez, line 363: idk if it's possible to do this but

should we have the `TYPE` macro put `... final : public Refcounted` into the class decl for you?

iiuc, every type kind must either be `TYPE ... Refcounted` or `TYPE_INLINED`, but maybe i'm mistaken

froydnj: I think you could do that, but the Sorbet Way seemed to be avoiding hiding things behind macros as much as possible.  I think the `final`-ness is important to have associated with the class, rather than hidden behind the `TYPE` macro, but that's a pretty weakly held opinion; certainly we hide `final` behind macros for trees and CFG instructions, so there is at least precedent.

core/Types.h
```diff
@@ -502,68 +501,105 @@ TYPE_INLINED(LiteralType) final {
     std::string showValue(const GlobalState &gs) const;
     uint32_t hash(const GlobalState &gs) const;
 
-    bool equals(const LiteralType &rhs) const;
+    bool equals(const NamedLiteralType &rhs) const;
     void _sanityCheck(const GlobalState &gs) const;
 };
-CheckSize(LiteralType, 16, 8);
+CheckSize(NamedLiteralType, 8, 8);
```
jez, line 507: The size of `NamedLiteralType` is only `8` because of the `__attribute__((aligned(8)))` in the class declaration, IIUC.

Do you think it's possible for us to static assert that the size of a class created via `TYPE_INLINED` *without* that attribute is size 4 and thus is small enough to fit into a `TypePtr`?

e.g., right now i don't know if there's anything protecting someone from accidentally adding another `uint32_t` to a `TYPE_INLINED` class, having the `CheckSize` assertion still say "yep, it's 8 bytes" but then having some data get dropped in our bit shifts earlier.

If this is already checked somewhere else please lmk

core/TypePtr.h
```diff
@@ -94,13 +107,13 @@ class TypePtr final {
         return static_cast<tagged_storage>(tag);
     }
 
-    static tagged_storage tagValue(Tag tag, uint32_t inlinedValue) {
+    static tagged_storage tagValue(Tag tag, uint64_t inlinedValue) {
```
jez, line 110: Why does the `inlinedValue` go up to 64 bits? Don't we only have `64 - 16 - 1 = 47` bits to spare here?

core/TypePtr.h
```diff
@@ -94,13 +107,13 @@ class TypePtr final {
...
+        // Ensure the upper bits aren't getting used for anything exciting.
+        ENFORCE_NO_TIMER((inlinedValue >> 48) == 0);
         auto val = tagToMask(tag);
 
-        // Store value into val.  It doesn't much matter where we put it in
-        // the upper 48 bits, but we put it in the uppermost 32 bits to
-        // ensure that retrieving it requires only a shift and no masking.
-        val |= static_cast<tagged_storage>(inlinedValue) << 32;
+        // Store value into val.  Put it in the same place as pointer values.
+        val |= static_cast<tagged_storage>(inlinedValue) << 16;
```
jez, line 116: Sorry I've been trying to piece together which bits correspond to what values in the `TypePtr`, depending on whether something is inlined or not.

I _think_ that since we're using `store & NOT_INLINED_MASK`, we need to shift the value to inline at least 17 bits?

I think i'd prefer keeping it at 32 bits just so that we can say "yep, we're only ever storing 32 bit values in here" rather than having to deal with non-word sizes like 47 bits.

core/Types.h
jez: i do not have any strongly held opinions here, so it's your call

my preference is to make APIs that are hard/impossible to misuse, then APIs that avoid hiding things, but again i defer here.

core/TypePtr.h
jez: Anyways, I'm not sure if you find this useful or not, but for my own understanding of how things fit together, i put together these diagrams

```
TypePtr, !containsPtr():
  0x0000'0000'0000'0000
    ├───────┘    │ ├──┘
    │            │ └──▶ 16 bits for `TypePtr::Tag`
    │            └──▶ 1 bit for `NOT_INLINED_MASK`
    │                 Bit is 0, therefore inlined.
    └──▶ 32 bits for inlined value, from class
         created using TYPE_INLINED

TypePtr, containsPtr():
  0x0000'0000'0000'0000
    ├────────────┤ ├──┘
    │            │ └──▶ 16 bits for `TypePtr::Tag`
    │            └──▶ 1 bit for `NOT_INLINED_MASK`
    │                 Bit is 1, therefore not inlined.
    │                 Overlaps with LSB of pointer,
    │                 lowest 3 bits of pointer are
    │                 unused due to 8-byte alignment
    └──▶ 48 bits for pointer

TYPE(...) layout:
  0x0000'0000'0000'0000
    ├──┘ ├──┘ ├──···──┘
    │    │    └──▶ normal class fields class
    │    └──▶ C++ structure padding to align on 8 bytes
    └──▶ `Refcounted` (atomic counter via parent class)
```

if you'd like to include them in the source code, feel free. otherwise, i'll at least be able to refer back to this PR in case I ever want to see them in the future.

froydnj: This is a cool diagram!  Note that for `TYPE(...)`, the padding doesn't have to exist depending on the alignment of the initial fields in the class.

I'm not actually sure we need `NOT_INLINED_MASK` anymore, that might be a holdover from a previous `TypePtr` tagging scheme (it ought to be redundant with the tag value, unless we can have nullptr tagged pointers...which I don't think we can, but maybe?  jvilk had that concern about `ExpressionPtr`).  Will have to investigate.

I left shifted by 16 so that we could conveniently store "small" integers (32-bit unsigned values + 32-bit negative values) without thinking too much about things...oh, and so we can store `NameRef` + at least 1 bit for determining whether we're storing literal symbols or literal strings.

core/Types.h
```diff
@@ -465,32 +469,27 @@ CheckSize(SelfType, 8, 8);
...
     return SelfType();
 }
 
-TYPE_INLINED(LiteralType) final {
+TYPE_INLINED(NamedLiteralType) final {
     union {
-        const int64_t value;
-        const double floatval;
         const NameRef name;
     };
```
jez, line 483: we can probably delete the one-item union right?

core/Types.h
froydnj: I think this is a good point, but we already have the same sort of issue with `ClassType`, `SelfTypeParam`, and `AliasType`.  `BlamedUntyped` even has the issue that it's eight bytes of data but still can be inlined!

And I think `BlamedUntyped` points at the issue with adding more stuff: eventually you are going to have to construct an instance of the enclosing type and figure out how you're going to get bits out of the type...and how you're going to put bits into the type.  So I think there are already some guardrails against people adding more stuff.

core/TypePtr.h
froydnj: We do, but variable-width integer types haven't made it to C++ yet (there's no `int47_t`).  (Note that we do at least have an `ENFORCE` that limits us to 48-bit values.)

jez:
> Will also run an emscripten build locally to make sure the bit-twiddling that comes with this PR didn't botch something on 32-bit.

Did you get a chance to fire up sorbet.run and confirm it's still working?

jez (commented):
Done reviewing for now. If tests pass and emscripten pass, lgtm

Though I do think there's a bug w.r.t. how we tag inlined values, noted in a comment.

The idea about small integers sounds good, I'm happy to do this in another PR though. If you ever get around to measuring how common integer literals are and how much memory that would save I'd be curious to hear.

core/TypePtr.h
jez: > This is a cool diagram! Note that for TYPE(...), the padding doesn't have to exist depending on the alignment of the > initial fields in the class.

I realized that after posting and edited (i think maybe you responded to a pre-edit version of the comment). There's a caveat now.

> I'm not actually sure we need NOT_INLINED_MASK anymore, that might be a holdover from a previous TypePtr tagging scheme (it ought to be redundant with the tag value, unless we can have nullptr tagged pointers...which I don't think we can, but maybe? jvilk had that concern about ExpressionPtr). Will have to investigate.

sounds good, i'm happy to use the tag bits to determine whether the ptr is tagged or not

There is a comment on top of `NOT_INLINED_MASK` related to nullptr, fwiw

froydnj: Ah, the diagram is incorrect: we only use 8 bits for the tag (the entire low byte), and the second-lowest byte stores `NOT_INLINED_MASK`.  The remaining six bytes hold the actual value (inlined value or pointer, as appropriate), so there's no possible overlap between the value and `NOT_INLINED_MASK`...which is good, because otherwise that would mean our `ENFORCE` in `tagValue` wasn't firing on some very common cases!

core/Types.h
froydnj: At least for the `Refcounted` inheritance, the internals of `TypePtr` work entirely with `Refcounted` instances, so you have to have it somewhere in your inheritance chain to make the compiler happy.  (This caught places where I failed to properly inherit from `Refcounted` during development.)

Oh, except for `tagPtr`...that should probably be fixed.

core/TypePtr.h
jez: ah ok that makes sense, yeah there were multiple times when making that diagram that i thought one hex letter was one byte.

jez: (i have updated the chart)

jez (approved):
i think my only fud was about the mask bit, but you pointed out that it doesn't actually overlap, i was just mistaken.

lgtm

froydnj:
emscripten seems to work; I get complaints about `emscripten/main.cc: Failed to parse JSON from JavaScript`, but ISTR that being normal.

Thanks for the review!

PR #5617, opened 2022-04-11
move `isSynthetic` to be stored on `InstructionPtr`

froydnj:
### Motivation

This decreases the size of some frequently-used CFG instructions, notably `Alias` and `Literal`.

Unfortunately, I don't think we can take the logical next step without several contortions: we'd want to "inline" `Ident` similar to how `ClassType` is inlined into `TypePtr`.  It'd be a pretty nice savings if we could manage that.  (We could "inline" several other instructions as well, but `Ident` is by far and away the most frequently occuring instruction in Stripe's codebase.)

### Test plan

See included automated tests.

### DIFF
--- a/cfg/Instructions.h
+++ b/cfg/Instructions.h
@@ -56,15 +56,13 @@ class Instruction;
 
 // When adding a new subtype, see if you need to add it to fillInBlockArguments
 class Instruction {
-public:
-    bool isSynthetic = false;
-
 protected:
     Instruction() = default;
     ~Instruction() = default;
 
 private:
     friend InstructionPtr;
+    bool isSynthetic = false;
 };
 
 #define INSN(name)                                                                  \
@@ -375,6 +373,14 @@ public:
         ENFORCE(ptr != 0);
 
         return static_cast<Tag>(ptr & TAG_MASK);
+    }
+
+    bool isSynthetic() const noexcept {
+        return this->get()->isSynthetic;
+    }
+
+    void setSynthetic() noexcept {
+        this->get()->isSynthetic = true;
     }
 
     std::string toString(const core::GlobalState &gs, const CFG &cfg) const;
--- a/cfg/Instructions.h
+++ b/cfg/Instructions.h
@@ -62,7 +62,6 @@ protected:
 
 private:
     friend InstructionPtr;
-    bool isSynthetic = false;
 };
 
 #define INSN(name)                                                                  \
@@ -90,7 +89,7 @@ public:
     std::string toString(const core::GlobalState &gs, const CFG &cfg) const;
     std::string showRaw(const core::GlobalState &gs, const CFG &cfg, int tabs = 0) const;
 };
-CheckSize(Alias, 16, 8);
+CheckSize(Alias, 8, 8);
 
 INSN(SolveConstraint) : public Instruction {
 public:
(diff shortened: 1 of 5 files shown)


froydnj:

→ https://go/builds/bui_LUjeoIA0TmDPE9
→ https://go/builds/bui_LUjemNLpLUyMfo

cfg/Instructions.h
```diff
@@ -276,14 +273,16 @@ INSN(TAbsurd) : public Instruction {
...
-CheckSize(TAbsurd, 32, 8);
+CheckSize(TAbsurd, 24, 8);
 
 class InstructionPtr final {
     using tagged_storage = uint64_t;
 
-    static constexpr tagged_storage TAG_MASK = 0xffff;
-
-    static constexpr tagged_storage PTR_MASK = ~TAG_MASK;
+    static constexpr tagged_storage TAG_MASK = 0xff;
```
jez, line 281: This seems a little bit like living on the edge? We already have 16 Instruction types (*and* it looks like we start the numbering at 1, so the 16th enum value’s integer wraps around and is 0). It doesn’t seem outlandish to me to think that we’ll need one more instruction kind at some point.

froydnj: This is a good callout, but the tag is not a bitmask: we can have ~200 tags and not overflow this field.

jez (approved):
This looks mostly good but I think we might want to dedicate more bits to the tag.

cfg/Instructions.h
jez: Oh sorry, one f is 16, ff is 256 🤦‍♂️

cfg/Instructions.h
```diff
@@ -276,14 +273,16 @@ INSN(TAbsurd) : public Instruction {
...
+    static constexpr tagged_storage FLAG_MASK = 0xff00;
+    static constexpr tagged_storage SYNTHETIC_FLAG = 0x0100;
+    static_assert((TAG_MASK & FLAG_MASK) == 0, "no bits should be shared between tags and flags");
+    static constexpr tagged_storage PTR_MASK = ~(FLAG_MASK | TAG_MASK);
```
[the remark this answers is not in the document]
froydnj: We do this for a couple of reasons:

1. A mask to compute the pointer would be a 48-bit (or so) mask, which is an inconvenient size for a constant on x86-64, requiring either an 11-byte (!) instruction to load the constant or a memory reference for the constant.  (A 48-bit all-ones mask would be OK on ARM64, since ARM64 has a reasonable suite of bit munging instructions).
2. Storing the tag in the low byte means that extracting the tag + actual pointer is two instructions and two registers:

```
movzx tagged_ptr_low_byte_register, tag
shr 16, tagged_ptr
# tagged_ptr now holds the actual pointer
```

(`movzx` can extract from the low 16 bits, too)

whereas putting the tag in the upper part would require more instructions:

```
mov tagged_ptr, temp
shr 48, temp # now holds the tag
# still need to mask tagged_ptr
```

so codesize is slightly smaller as a result.
3. Technically, pointers on x86-64  need to have their top 17 bits the same, and putting the pointer in the upper 48 bits enables us to adhere to that by using a signed right shift to extract the pointer.  Storing the pointer in the low 48 bits means we would need two shifts or gnarly conditional logic.  (I don't think we actually care about this in our tagged pointer implementation(s), but the constraint is there.)

PR #5618, opened 2022-04-11
reduce the space required by empty generic argument lists on methods and classes

froydnj:
### Motivation

Generic parameters are fairly uncommon across methods and classes, but we make all methods and classes pay for their cost, size-wise, with a 24-byte `InlinedVector` in their respective symbols.  (It would be nice if we could at least reduce the inline storage cost--very few people need four generic parameters--but that's not possible due to restrictions in `InlinedVector` (?))

This PR implements the next best thing, which is lazily allocating the vectors storing those parameters.  It probably wants to wait for #5005, as some of the serialization code for `ClassOrModule` will be less messy after that PR.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1118,9 +1118,9 @@ TypeArgumentRef GlobalState::enterTypeArgument(Loc loc, MethodRef owner, NameRef
     flags.isTypeArgument = true;
 
     auto ownerScope = owner.dataAllowingNone(*this);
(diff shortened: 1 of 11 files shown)


jez (approved):

froydnj:

→ https://go/builds/bui_LUjGXNAOxYpqqY
→ https://go/builds/bui_LUjGJLIImn35Gz

PR #5619, opened 2022-04-11
Reorganize a loop body for clarity

elliottt:
Remove some nested control flow in a loop body in favor of using `continue` instead.

### Motivation

Improving readability.

### Test plan

See included automated tests.


froydnj (approved):
:chefkiss:

PR #5623, opened 2022-04-12
Improve "method does not exist" error

jez:
### Motivation

This is one of the most common errors in all of Sorbet.

Some important changes:

- We're using `funLoc` now whenever possible, letting me delete some
  pre-existing TODOs and also a good chunk of annoying code.

- Doing any autocorrect earlier in the method prevents adding subsequent
  autocorrects. So if we're going to suggest inserting `extend T::Helpers`, we
  won't also suggest renaming the call site to `select`.

- We have much better errors when mistaking instance/singleton methods. We will
  either suggest inserting `self.` to make a method a singleton class method, or
  suggest inserting `.class` to call the singleton class method.

  This is not always the autocorrect you want (imagine: if you just removed
  `self.` from a method, you probably want Sorbet to figure out how to convert
  all your calls to instance method calls?) But it's much easier for Sorbet to
  always go from an instance to a singleton, because there's only ever one
  singleton in scope (which you can easily get by calling `.class`), but there'
  not an easy way to go from a singleton to an instance (because not all
  constructors can just take `.new` with no arguments).

  In any case, we always show a nicer message with more words, and that might
  tip the user off that something is wrong.

### Test plan

See included automated tests.

image


elliottt (approved):
These look great, much more uniform and to the point!

PR #5627, opened 2022-04-12
Remove `type_parameters` from `T::Generic`

jez:
### Motivation

This method doesn't actually exist on this class at runtime.

### Test plan

test on Stripe's codebase


elliottt (approved):
That's odd, good find!

jez:

→ https://go/builds/bui_LUn1YJ2xLxNZyr
→ https://go/builds/bui_LUn1HeW7l74o21

PR #5628, opened 2022-04-12
Allow specifying `type_member` bounds lazily

jez:
### Motivation

The types that show up in bounds are completely erased at runtime, so there's no
point not to make them lazy.

This should:

- potentially show load-time performance improvements
- prevent certain kinds of load-time cyclic references from happening

We will land a follow up change eventually to enforce the new syntax and add an
autocorrect.

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1833,7 +1833,14 @@ public:
                 return ast::MK::EmptyTree();
             }
 
-            if (send->hasKwArgs()) {
+            // TODO(jez) Delete the code to parse hasKwArgs and just emit an autocorrect if they're present
+            if (send->hasKwArgs() || send->block() != nullptr) {
+                if (send->hasKwArgs() && send->block()) {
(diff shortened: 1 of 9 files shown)


namer/namer.cc
```diff
@@ -1852,6 +1859,40 @@ class TreeSymbolizer {
...
+                            switch (key->asSymbol(ctx).rawId()) {
+                                case core::Names::fixed().rawId():
+                                    fixed = true;
+                                    break;
+
+                                case core::Names::lower().rawId():
+                                case core::Names::upper().rawId():
+                                    bounded = true;
+                                    break;
+                            }
```
froydnj, line 1884: I think we should error on unknown keys, though I don't know whether it's better to do it here or in type syntax parsing.

froydnj (approved):
Just one comment on hash key parsing.

PR #5637, opened 2022-04-14
Fix lazy type members

jez:
### Motivation

Forgot to add enough tests in #5628 to discover this.

Exposed when I went to implement the autocorrect to convert all `type_member`s
to the new syntax.

### Test plan

See included automated tests.


froydnj (approved):
Doh.

PR #5639, opened 2022-04-14
Add autocorrect to enforce lazy type_members

jez:
### Motivation

Enforce the new syntax.

### A note about migration

The new syntax was added as an option in https://github.com/sorbet/sorbet/pull/5628.
There was a bug fix that landed in https://github.com/sorbet/sorbet/pull/5637

That means that the first published-to-RubyGems version that **allows** the
new syntax but does not enforce it is: 0.5.9889

If you need to do an incremental migration in your codebase to get Sorbet to pass with these
changes, and can't (or don't want to) land the autocorrected changes in the same commit that
upgrades the Sorbet version, you'll want to first upgrade to the above version temporarily.
That will let you start incrementally landing the fixes before you finally upgrade to the version
containing these changes, which will enforce the new syntax.

To run the autocorrect contained in this PR, you can use

```
srb tc --autocorrect --isolate-error-code=4021
```

which will run Sorbet over your codebase, but only report errors for `type_member` and
`type_template` declarations using the old syntax, and also autocorrect the files in place
to use the new syntax.

### Test plan

See included automated tests.

The autocorrect included in this PR was able to automatically fix the 14,000+ errors occurring in 9,000+ files with a single `--autocorrect` run, also yielding no violations of any of Stripe's Rubocop rules.


jez:

→ https://go/builds/bui_LVIlPwFdHwzgua
→ https://go/builds/bui_LVIl5W2YvjyPrA

namer/namer.cc
```diff
@@ -1800,6 +1789,63 @@ class TreeSymbolizer {
...
+                }
+                edits.emplace_back(core::AutocorrectSuggestion::Edit{
+                    insertLoc, fmt::format(" do\n{0}  {{\n{0}    {1},\n{0}  }}\n{0}end", indent, reindentedSource)});
+            } else {
+                edits.emplace_back(core::AutocorrectSuggestion::Edit{
+                    insertLoc, fmt::format(" do\n{0}  {{{1}}}\n{0}end", indent, kwArgsSource)});
+            }
+        } else {
+            edits.emplace_back(core::AutocorrectSuggestion::Edit{insertLoc, fmt::format(" {{{{{}}}}}", kwArgsSource)});
+        }
```
jez, line 1842: You might be wondering, "why is this so complicated" and the answer is basically that this is the minimal set of changes that were required to get an autocorrect that was able to correct all the usage variations of `type_member`s in Stripe's codebase, even making sure that the resulting output conforms to Stripe's Rubocop rules.

namer/namer.cc
```diff
@@ -1800,6 +1789,63 @@ class TreeSymbolizer {
...
+                    reindentedSource = absl::StrReplaceAll(reindentedSource, {{"\n", "\n  "}});
+                }
+                edits.emplace_back(core::AutocorrectSuggestion::Edit{
+                    insertLoc, fmt::format(" do\n{0}  {{\n{0}    {1},\n{0}  }}\n{0}end", indent, reindentedSource)});
+            } else {
+                edits.emplace_back(core::AutocorrectSuggestion::Edit{
+                    insertLoc, fmt::format(" do\n{0}  {{{1}}}\n{0}end", indent, kwArgsSource)});
+            }
+        } else {
+            edits.emplace_back(core::AutocorrectSuggestion::Edit{insertLoc, fmt::format(" {{{{{}}}}}", kwArgsSource)});
```
elliottt, line 1841: {{{{{{{ 👀

elliottt (approved):
Looks good! Just a few questions before merging.

rbi/tools/sync-rdoc.rb
```diff
@@ -239,7 +239,7 @@ def each_doc(&blk)
 
 # Fix generation of nested markup <code>foo<em>bar</em>baz</code>
 module FormatterTagPatch
-  TagRange = Struct.new(:tag, :begin, :end)
+   {{TagRange = Struct.new(:tag, :begin, :end)}}
```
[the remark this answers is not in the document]
jez: I think this was an accidental commit from before I fixed how the `Struct.cc` rewriter works. I'll double check that it is in fact fixed before merging.

namer/namer.cc
jez: 😅

rbi/tools/sync-rdoc.rb
```diff
@@ -375,7 +375,7 @@ def add_alias_info(method)
 module DocDiff
   def self.render_lines(lines)
     md = lines.map {|l| l.gsub(/\A\s*# /, '')}.join
-    html = CommonMarker.render_html(md, :DEFAULT)
+    html = OptionParser.render_html(md, :DEFAULT)
```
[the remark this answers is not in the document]
jez: It looks like I forgot to pass the `--isolate-error-code` when running the autocorrects over the `rbi/` folder, so some of the missing constant suggestions got committed too. thanks for flagging

PR #5643, opened 2022-04-15
add an autocorrect for deleting useless casts

### LINKED ISSUE #5236: Autocorrect to remove useless `T.cast`
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0AT.cast%281%2C%20Integer%29)

```ruby
# typed: strict

T.cast(1, Integer)
```

#### Observed output

```
editor.rb:3: T.cast is useless because Integer(1) is already a subtype of Integer https://srb.help/7015
     3 |T.cast(1, Integer)
        ^^^^^^^^^^^^^^^^^^
  Got Integer(1) originating from:
    editor.rb:3:
     3 |T.cast(1, Integer)
               ^
Errors: 1
```

#### Expected behavior

Sorbet should offer an autocorrect for this case.

- - -

froydnj:
### Motivation

Fixes #5236.

### Test plan

See included automated tests.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1469,6 +1469,11 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                             e.setHeader("`{}` is useless because `{}` is already a subtype of `{}`", "T.cast",
                                         ty.type.show(ctx), castType.show(ctx));
                             e.addErrorSection(ty.explainGot(ctx, ownerLoc));
+                            auto argLoc = core::Loc{ctx.file, c.valueLoc};
+                            if (argLoc.exists()) {
+                                e.replaceWith("Delete `T.cast`", ctx.locAt(bind.loc), "{}",
+                                              argLoc.source(ctx).value());
+                            }
                         }
                     }
                 }
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1469,7 +1469,7 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                             e.setHeader("`{}` is useless because `{}` is already a subtype of `{}`", "T.cast",
                                         ty.type.show(ctx), castType.show(ctx));
                             e.addErrorSection(ty.explainGot(ctx, ownerLoc));
-                            auto argLoc = core::Loc{ctx.file, c.valueLoc};
+                            auto argLoc = ctx.locAt(c.valueLoc);
                             if (argLoc.exists()) {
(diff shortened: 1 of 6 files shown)


infer/environment.cc
```diff
@@ -1469,6 +1469,10 @@ core::TypePtr Environment::processBinding(core::Context ctx, const cfg::CFG &inW
                             e.setHeader("`{}` is useless because `{}` is already a subtype of `{}`", "T.cast",
                                         ty.type.show(ctx), castType.show(ctx));
                             e.addErrorSection(ty.explainGot(ctx, ownerLoc));
+                            auto argLoc = core::Loc{ctx.file, c.valueLoc};
```
jez, line 1472: if you want you could use locAt here too

test/cli/autocorrect-useless-cast/autocorrect-useless-cast.rb
```diff
@@ -0,0 +1,17 @@
...
+)
+
+T.cast(
+  "multiline2",
+  String)
+
+T.cast("multiline3",
+       String)
+
+T.cast(1.5, Float)
```
jez, line 17: ooc, any reason to keep this a CLI test vs using an `*.autocorrects.exp` file on a testdata test?

usually the only reason is if there's additional non-`setHeader` error context that you want to test is showing up correctly but I think that the `explainGot` helper is sufficiently well tested elsewhere.

infer/environment.cc
froydnj: Gah, I can't even remember to use helpers that I added in the first place.

jez (approved):

test/cli/autocorrect-useless-cast/autocorrect-useless-cast.rb
froydnj: I did...not know about `*.autocorrects.exp`.

I'll leave it like this for now; maybe it'd be worth separating the CLI autocorrect tests into `*.autocorrects.exp` tests and keep the CLI tests for checking actual error text.

jez: eh, in a lot of cases it's annoying to have the Ruby source file in two places. In those cases, having a single CLI test is preferable to two tests that have the same Ruby source file.

But yeah, the `*.autocorrects.exp` format is more recent than many of the autocorrect CLI tests, and it's likely that some of them could be plainly converted to no longer be CLI tests.

PR #5644, opened 2022-04-15
--stripe-packages: Hide PackageSpecs from top-level namespace

ngroman:
Wrap all package spec declarations so that they don't resolve at the top level. This is done by prefixing them with a symbol `<PackageSpecRegistry>` during packaging.

### Motivation
Referencing unimported packages *should* result in a constant resolution error. However, in cases, especially where the package had behavior at the top level, you could get confusing errors or no error at all. For example:

If you were to `include Unimported::Package` this would be a no-op and not result in any error.

### Test plan
Added tests.

- [ ] Test on Stripe's codebase

See included automated tests.


test/testdata/packager/unimported_namespace/aaa/a_class.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+# typed: strict
+
+class AAA::AClass
+  BBB
+# ^^^ error: Unable to resolve constant `BBB`
```
ngroman, line 6: Prior to this change, this would have resolved. That's the whole point of this change.

test/cli/simple-package-rbi-gen/test.out
```diff
@@ -1,6 +1,12 @@
 No errors! Great job.
 # typed: true
 
+module ::Project
+end
+
+module ::Project::Foo
+end
+
```
ngroman, line 9: I don't have a ton of context on the rbi gen stuf going on here, but this seems like effectively a noop. cc @aadi-stripe @jez

nroman-stripe:

→ https://go/builds/bui_LX17eQnRsMp3z6
→ https://go/builds/bui_LX17CgP04zHcLm

ngroman:
I got this passing on Stripe's codebase with some minor changes:
See http://go/builds/bui_LX5BNbw7PBh6nj

ngroman:
(Rebased)

test/cli/simple-package-rbi-gen/test.out
aadi-stripe: Doesn't this just mean that because these symbols are not available at the top-level by default anymore (due to the re-namespacing), we end up generating the shims for them?

packager/packager.cc
```diff
@@ -1191,6 +1214,10 @@ struct PackageInfoFinder {
 
             info->loc = ctx.locAt(classDef.loc);
             info->declLoc_ = ctx.locAt(classDef.declLoc);
+
+            // `class Foo < PackageSpace` -> `class <PackageSpecRegistry>::Foo < PackageSpec`
```
jez, line 1218:
```suggestion
            // `class Foo < PackageSpec` -> `class <PackageSpecRegistry>::Foo < PackageSpec`
```

packager/packager.cc
```diff
@@ -623,6 +623,15 @@ ast::ExpressionPtr prependPackageScope(const core::GlobalState &gs, ast::Express
     return scope;
 }
 
+ast::ExpressionPtr prependName(ast::ExpressionPtr scope, core::NameRef prefix) {
+    auto *lastConstLit = &ast::cast_tree_nonnull<ast::UnresolvedConstantLit>(scope);
+    while (auto constLit = ast::cast_tree<ast::UnresolvedConstantLit>(lastConstLit->scope)) {
```
jez, line 628: Hm, why not just use pointers?

```suggestion
    auto lastConstLit = ast::cast_tree<ast::UnresolvedConstantLit>(scope);
    ENFORCE(lastContLit != nullptr);
    while (auto constLit = ast::cast_tree<ast::UnresolvedConstantLit>(lastConstLit->scope)) {
```

(i try to avoid the address-of `&` operator whenever possible)

test/cli/simple-package-rbi-gen/test.out
jez: seems fine to me, we can figure it out later if it's a problem

jez (approved):
Couple small suggestions, but this looks great, thanks!

test/cli/simple-package-rbi-gen/test.out
nroman-stripe: Yep, that's my understanding.

PR #5646, opened 2022-04-16
store IDs in a single table for compiled modules

froydnj:
### Motivation

This is the followup to #5613, implementing the plan for the ID table described therein.

The main benefits here are less LLVM IR to optimize.

### Test plan

See included automated tests.  The compiler exp diffs are rather large, but that's mostly just a function of shifting LLVM IR values around, since we now have fewer LLVM IR temporaries.


elliottt (approved):
This is awesome! So great to see all that red in the exp files :)

PR #5649, opened 2022-04-18
give an explanation for empty hover results on untyped files

froydnj:
### Motivation

More information for users is a good thing.

Getting the file information on empty results is a little bit of a hack.  I think the cleaner thing might be extending `LSPQuery` with some information about why the results are empty, but it would be a weird enum like:

```C++
enum class EmptyResultReason {
  TypedFalse,
  TypedIgnore, // these could maybe be collapsed into one
  Irrelevant,
};
```

and the default value would be `Irrelevant`.

### Test plan

See included automated tests.


jez:
@froydnj what do you think about adding a new `QueryResponse` like `BadSigilResponse` (or some other name)?

At the end of `LSPTypechecker::query` we could check to see if the list of responses was empty, and if so we could do the same logic you're doing here to check the strictness level of the file. That way we could reuse the `BadSigilResponse` across hover.cc, definitions.cc, references.cc, etc.

main/lsp/requests/hover.cc
```diff
@@ -52,14 +52,24 @@ unique_ptr<ResponseMessage> HoverTask::runRequest(LSPTypecheckerInterface &typec
...
-        // Note: Need to specifically specify the variant type here so the null gets placed into the proper slot.
-        response->result = variant<JSONNullObject, unique_ptr<Hover>>(JSONNullObject());
+        auto fref = config.uri2FileRef(gs, params->textDocument->uri);
+        ENFORCE(fref.exists());
+        auto level = fref.data(gs).strictLevel;
+        if (level < core::StrictLevel::True) {
+            string asString = level == core::StrictLevel::Ignore ? "ignore" : "false";
+            auto text = fmt::format("# No hover information available");
+            auto explanation = fmt::format("The file is `# typed: {}`", asString);
+            response->result = make_unique<Hover>(formatRubyMarkup(clientHoverMarkupKind, text, explanation));
```
jez, line 64: Is it worth using formatRubyMarkup here?

It seems like somewhat overkill to show

> ```ruby
> # No hover information available
> ```
> ---
> The file is `typed: false`

vs showing

> No hover information available.
> Hint: the file is `typed: false`.

froydnj:
> @froydnj what do you think about adding a new `QueryResponse` like `BadSigilResponse` (or some other name)?
> 
> At the end of `LSPTypechecker::query` we could check to see if the list of responses was empty, and if so we could do the same logic you're doing here to check the strictness level of the file. That way we could reuse the `BadSigilResponse` across hover.cc, definitions.cc, references.cc, etc.

I like this solution; I think it's more elegant than a field on `LSPQueryResult`, but I'm not sure how to rewrite:

https://github.com/sorbet/sorbet/blob/a10576f5d16729d4610bb84768f91255ae59c70c/main/lsp/LSPQuery.cc#L55-L79

(which is called prior to `LSPTypechecker::query`, so the proposed code would never get hit.)  I guess we could just remove:

https://github.com/sorbet/sorbet/blob/a10576f5d16729d4610bb84768f91255ae59c70c/main/lsp/LSPQuery.cc#L75-L79

but I'm not sure what kind of impact that would have (would cost some performance on some `typed: false` files for some queries?).

froydnj:
Oh, I guess we just make that code return `UntypedResponse` or whatever as one of the responses, rather than returning an empty response list.  Easy!

elliottt (approved):
This is a huge improvement over no hover results!

froydnj:
I thought about this some more overnight and realized I don't like something about `BadSigilResponse`: all of our previous responses are specifically responses dealing with Ruby constructs being queried by the client.  `BadSigilResponse` is different: it's more about the environment that the query was performed in rather than any response to the query and therefore sort of weird to handle in the existing framework.  (I'm also not totally sure how often it would get used; I could see #5598 using it and maybe `completion.cc`...but then those places could also do something similar to what we're doing here or the tree below.)

I think the problem is `LSPQueryResult` is really some tri-state (or more!) sort of thing; in a horrible mashup of Rust and C++, it's:

```
enum LSPQueryResult {
  // Responses to the query, possibly empty.
  Success(Vec<core::lsp::QueryResponse>),
  // We had a single file to query and requested "erroring" on untyped files.
  UntypedFile,
  Error(ResponseError),
};
```

This structure has the downside of forcing everybody to be aware of `UntypedFile`, even when they knew that response couldn't come back (because you passed `errorIfFileIsUntyped=false` to `LSPQuery::byLoc` or you used one of the other named functions in `LSPQuery`).  Perhaps that just points to the idea that `LSPQuery::byLoc` should be returning a fundamentally different type than `LSPQuery::{bySymbolInFiles,bySymbol}`.

I prototyped out something along the lines of the tri-state enum above: https://github.com/sorbet/sorbet/tree/froydnj-empty-hover-results-enum-refactor

I don't know whether it's *better*, per se; I think it's slightly clearer, but still not great in the confines of C++ (I don't want to go full `std::variant`, because I think that would require rewriting way too much code).

Anyway, I'm going to go ahead and merge this, and we can follow up with a pass over the code if people feel strongly about the types involved.

PR #5650, opened 2022-04-18
Add immutable struct

vinistock:
This PR adds the `T::ImmutableStruct` class. This class is very similar to a regular `T::Struct`, but does not have the `prop` method defined and all objects created are frozen by default.

The most impactful change is the first commit. Let me know if you have alternatives on how to achieve this.

I recommend reviewing the changes per commit. The changes are
- Split the `T::Props` module into three separates modules `Common`, `Prop` and `Const`. The purpose of the split is to allow the inclusion a granular inclusion of `Const` or `Prop` with the common functionality needed for both, while trying to minimize breaking changes as much as possible (i.e.: any class including `T::Props` should continue to work the same)
- The `Serializable` module had an expectation that `Props` would always be included as well, through the inclusion of `Plugin`. This is the only breaking change I was not able to avoid, since we wanted the immutable structs to have serialization functionality, but not `prop`. Classes that include `Serializable` now need to explicitly include `T::Props` or `T::Const` and this commit makes that change in the tests
- Add the `T::ImmutableStruct` class, including only `Const` and preventing folks from including `T::Props` or `T::Props::Prop`, which would defeat its purpose
- Add support for `T::ImmutableStruct` in the prop rewriter to get the correct signatures for the `initialize` method

### Motivation

Having the implementation of immutable structs is useful for working with value objects, which should not be mutable by default. With this implementation, incorrect usage of `T::ImmutableStruct`, such as invoking `prop` or trying to change the value of an existing `prop`, is caught in static analysis.

The proposal to have this built into `sorbet-runtime` is because of the coupling with the prop rewriter, which can only add the correct `initialize` signature for known built-in structs.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
+++ b/gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
@@ -11,7 +11,7 @@ module T::Props
 
       sig do
         params(
-          klass: T.all(Module, T::Props::ClassMethods),
+          klass: T.all(Module, T::Props::Common::ClassMethods),
           prop: Symbol,
           rules: T::Hash[Symbol, T.untyped]
         )
@@ -53,7 +53,7 @@ module T::Props
           prop: Symbol,
           accessor_key: Symbol,
           non_nil_type: Module,
-          klass: T.all(Module, T::Props::ClassMethods),
+          klass: T.all(Module, T::Props::Common::ClassMethods),
         )
         .returns(SetterProc)
       end
@@ -76,7 +76,7 @@ module T::Props
           prop: Symbol,
           accessor_key: Symbol,
           non_nil_type: T::Types::Base,
-          klass: T.all(Module, T::Props::ClassMethods),
+          klass: T.all(Module, T::Props::Common::ClassMethods),
           validate: T.nilable(ValidateProc)
         )
         .returns(SetterProc)
@@ -107,7 +107,7 @@ module T::Props
           prop: Symbol,
           accessor_key: Symbol,
           non_nil_type: Module,
-          klass: T.all(Module, T::Props::ClassMethods),
+          klass: T.all(Module, T::Props::Common::ClassMethods),
         )
         .returns(SetterProc)
       end
@@ -134,7 +134,7 @@ module T::Props
           prop: Symbol,
           accessor_key: Symbol,
           non_nil_type: T::Types::Base,
-          klass: T.all(Module, T::Props::ClassMethods),
+          klass: T.all(Module, T::Props::Common::ClassMethods),
           validate: T.nilable(ValidateProc),
         )
         .returns(SetterProc)
@@ -165,7 +165,7 @@ module T::Props
 
       sig do
         params(
-          klass: T.all(Module, T::Props::ClassMethods),
+          klass: T.all(Module, T::Props::Common::ClassMethods),
           prop: Symbol,
           type: T.any(T::Types::Base, Module),
           val: T.untyped,
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -225,6 +225,7 @@ NameDef names[] = {
     {"computedBy", "computed_by"},
     {"factory"},
     {"InexactStruct", "InexactStruct", true},
+    {"ImmutableStruct", "ImmutableStruct", true},
     {"Chalk", "Chalk", true},
     {"ODM", "ODM", true},
     {"Document", "Document", true},
--- a/gems/sorbet-runtime/lib/types/props/_props.rb
+++ b/gems/sorbet-runtime/lib/types/props/_props.rb
@@ -84,8 +84,11 @@ module T::Props
   end
 
   module Prop
+    extend T::Helpers
+
     module ClassMethods
       extend T::Sig
+      include T::Props::Common::ClassMethods
 
       # Define a new property. See {file:README.md} for some concrete
       #  examples.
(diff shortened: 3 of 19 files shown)


gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
```diff
@@ -11,7 +11,7 @@ module SetterFactory
 
       sig do
         params(
-          klass: T.all(Module, T::Props::ClassMethods),
+          klass: T.all(Module, T::Props::Common::ClassMethods),
```
jez, line 14: Is it possible to structure the code so that this API doesn't change? Stripe's codebase uses `T::Props::ClassMethods` all over the place and I'd love to not have to do a codemod for this PR.

gems/sorbet-runtime/test/types/props/decorator.rb
```diff
@@ -185,6 +185,7 @@ class InnerStruct < T::Struct
   end
 
   class OptionalArrayClass
+    include T::Props
```
jez, line 188: Why are these required now? Why wouldn't `T::Props::Serializable` still include `T::Props`?

froydnj:
(Thinking out loud here, not sure if adding more stuff on this PR is feasible.)

It would be fantastic if we could additionally restructure things in such a way that `#with` (or similar) is *not* supported on these new structures.  I get the feeling that, given immutability, people are naturally going to turn to `#with` to create similar structures with one field changed, and `#with` is not a good API.  It would be unfortunate that we don't have a better thing to offer them, but at least we wouldn't be silently encouraging people to use `#with`.

Maybe this PR already ensures that `#with` isn't supported, which would be `:chefkiss:`.

gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
vinistock: I will give it a try, but it's possible that some of these may need to changed to accommodate for `T::ImmutableStruct`.

gems/sorbet-runtime/test/types/props/decorator.rb
vinistock: If we have `Serializable` include `T::Props`, then we can't include `Serializable` in the `T::ImmutableStruct` and thus lose the serialization features.

jez: but we could factor something out to like `T::Props::Private::SerializableImpl` and make

```ruby
module T::Props::Serializable
  include T::Props
  include T::Props::Private::SerializableImpl
end
```

right?

vinistock: Yes, we can do it. Be mindful that `Plugin` also included `T::Props`, which creates the same dependency for `WeakConstructor`, so we would need to do a similar split for those modules as well. Let me know if you agree with that direction.

froydnj:
I think jez has stronger feeling than I do in this area, so I've removed myself from the PR; getting rid of `#with` was my main idea contribution.

gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
vinistock: I have tried reflecting on how to not change these signatures, but I couldn't come up with anything that will satisfy all our needs.

These methods should all work for an `ImmutableStruct`, which does not include `T::Props`. At the same time, we want the behaviour of including `T::Props` to remain the same - which means that `ImmutableStruct` can't include `T::Props`.

Let me know if you have a good idea on how to work around this.

vinistock:
In addition to splitting all of the `T::Props` related modules, I have also split a module for the `with` method, so that `ImmutableStruct` does not have access to it.

jez:

→ https://go/builds/bui_LwZePdCbQ2bF3A
→ https://go/builds/bui_LwZeMvDZHz7NL0

jez:
The refactoring of `T::Props` as currently implemented introduces almost a thousand type errors on Stripe's codebase.

Nearly all the errors blame to the fact that many existing abstractions within Stripe's codebase assume that doing

```ruby
include T::Props
include T::Props::Serializable
```

is sufficient to approximate what `T::Struct` does (there is a reason why `T::InexactStruct` before this PR is so short, which is because it was previously assumed that including `T::Props` and `T::Props::Serializable` is sufficient to recreate custom struct-like abstractions.

In particular, most of these abstractions are in fact older than Sorbet or sorbet-runtime itself, meaning they extend very widely through Stripe's codebase, including some of the most sensitive parts of the codebase (performance-wise and reliability-wise). The fact that this PR breaks so much of Stripe's codebase in its current form and would require a large codemod to land give me pause.

As an example of the kind of API which would be great to not break:

```ruby
# typed: true
require 'sorbet-runtime'

class Base
  include T::Props
  include T::Props::Serializable
end

class Child < Base
  prop :foo, Integer
end

Child.from_hash({foo: 123})
```
[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Arequire%20'sorbet-runtime'%0A%0Aclass%20Base%0A%20%20include%20T%3A%3AProps%0A%20%20include%20T%3A%3AProps%3A%3ASerializable%0Aend%0A%0Aclass%20Child%20%3C%20Base%0A%20%20prop%20%3Afoo%2C%20Integer%0Aend%0A%0AChild.from_hash%28%7Bfoo%3A%20123%7D%29)

Are we 100% confident that it's not possible to build some sort of immutable struct abstraction in a way that preserves this existing API?

If we are confident, then the change will have to be restructured to involve two changes, one intermediate PR which is possible to migrate Stripe's codebase to, and a final PR which introduces the breaking change.

Please let me know if you have further questions.

jez (commented):
^

PR #5651, opened 2022-04-18
Move rubyfmt to its own non-blocking job

elliottt:
The `test-static-sanitized.sh` job is made quite a bit longer by the requirement that we build `rubyfmt` to run the cli tests. This PR moves the rubyfmt tests to their own job, and removes `--config=rubyfmt` from the buildfarm configuration.

### Motivation

Improving build times.

### Test plan

See included automated tests.


.buildkite/pipeline.yaml
```diff
@@ -40,6 +40,10 @@ steps:
     artifact_paths: _out_/profile.json
     <<: *elastic
 
+  - label: ":linux: test-sorbet-rubyfmt.sh"
+    command: .buildkite/test-sorbet-rubyfmt.sh
+    <<: *elastic
+
```
jez, line 46: Curious what you think about moving this to be a job that happens down at the bottom of the file (next to the other non-blocking job)?

Right now as written this is still a blocking job.

.buildkite/test-sorbet-rubyfmt.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
```
jez, line 3: I think that the buildkite yaml pipeline syntax allows setting environment variables. Do you think we could set an environment variable and read that inside the existing `test-static-sanitized.sh` file so that we don't have to duplicate so much code?

.buildkite/pipeline.yaml
elliottt: Ah, that was the intention originally. I had thought that always returning `0` would make it so that the job could fail, but that the failure wouldn't hold up subsequent jobs.

.buildkite/test-sorbet-rubyfmt.sh
elliottt: Seems reasonable to me!

.buildkite/pipeline.yaml
jez: Ah yeah that would work, but it would still potentially spend an instance running these tests before you could click the merge button.

jez (approved):
🙏

PR #5652, opened 2022-04-19
Collapse duplicated error code 3506

jez:
### Motivation

It looks like were reusing an error code. I considered making these into
two separate error codes, but we're actually using
`--suppress-error-code=3506` inside Sorbet, so I figured that splitting
it into two error codes could potentially break something and I didn't
want to have to deal with that on the off chance it happens.

Luckily the two error codes were basically doing the same thing
(`T::Enum` rewriter syntax) so I figured I'd just use the same
ErrorClass for both.

### Test plan

n/a


elliottt (approved):

jez:
Screen Shot 2022-04-18 at 9 02 53 PM

Screen Shot 2022-04-18 at 8 36 22 PM

PR #5654, opened 2022-04-19
Delete dead constant type code in resolver

jez:
### Motivation

There are no tests for this codepath. I can't think of any tet case that
triggers this codepath. I'd rather just delete it.

### Test plan

Dead code.


froydnj (changes requested):
It looks like we have at least one test case that exercises this codepath?

jez:
oh beautiful, that's even better.

jez:
> It looks like we have at least one test case that exercises this codepath?

One of the tests was downstream of another error that is always reported, so the second error is meaningless.

The other one was because in deleting the code I didn't realize that some code was actually important for its side effects, so I've re-introduced that call (to `resolveConstantType`).

froydnj (approved):

PR #5659, opened 2022-04-20
untangle `LocOffsets` from `Loc.h`

froydnj:
### Motivation

For #5598 (and maybe other LSP queries, like find all references), the plan is to jump the user to the location of the `typed:` sigil when we're jumping to definition in an untyped file.  To do that, it seems reasonable to record the `LocOffsets` of the sigil when we compute the original sigil for the file.  To do that, we need to arrange things such that `Loc.h` doesn't depend on `Files.h`: we need to pull `LocOffsets` out to its own file...but we also need to pull `FileRef` out to its own file, so `LocOffsets.h` can depend on `FileRef.h` and not `Files.h`.

Regardless of whether we store `LocOffsets` in the `File` structure itself or we change `File::fileStrictSigil` to additionally compute the `LocOffsets`, we need a change like this.

### Test plan

No functional change intended.

### DIFF
--- a/core/FileRef.h
+++ b/core/FileRef.h
@@ -1,0 +1,57 @@
+#ifndef SORBET_CORE_FILEREF_H
+#define SORBET_CORE_FILEREF_H
+
+#include "common/common.h"
+
+namespace sorbet::core {
+class GlobalState;
+class File;
+
+class FileRef final {
+public:
+    FileRef() : _id(0){};
+    FileRef(unsigned int id);
+
+    FileRef(FileRef &f) = default;
+    FileRef(const FileRef &f) = default;
+    FileRef(FileRef &&f) = default;
+    FileRef &operator=(const FileRef &f) = default;
+    FileRef &operator=(FileRef &&f) = default;
+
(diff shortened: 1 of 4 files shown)


elliottt (approved):
Seems reasonable to me!

Should we start being more deliberate about listing headers like this in the `hdrs` field of the `core` package in `core/BUILD`? Right now we use a lot of `core` headers that are only named in `srcs`, but the intent of `hdrs` is that it list the publicly accessible headers of the package.

froydnj:
> Should we start being more deliberate about listing headers like this in the `hdrs` field of the `core` package in `core/BUILD`? Right now we use a lot of `core` headers that are only named in `srcs`, but the intent of `hdrs` is that it list the publicly accessible headers of the package.

I would be a fan of doing that, but we'd essentially be listing everything--at least at this point--wouldn't we?  (And how does this work now, given that `hdrs` for `core/BUILD:core` doesn't actually list any `core/*.h` headers?)  I guess it would be helpful in other places where we have a little bit more of a division between public and private headers.

PR #5661, opened 2022-04-20
log empty completion results

froydnj:
### Motivation

More insight into why people are getting no completions.

The one thing I'm not sure about this is how expensive the logging is going to be; if LSP clients are completion-happy, we could be firing this logging quite often.

### Test plan

See included automated tests.


jez (approved):
😍

elliottt (approved):

froydnj:
> The one thing I'm not sure about this is how expensive the logging is going to be; if LSP clients are completion-happy, we could be firing this logging quite often.

Still some FUD around this, but I'm going to go ahead and merge this and we can figure out later if this is a problem (maybe we only log for explicitly-requested completion requests?).

PR #5662, opened 2022-04-20
Update BigMath RBI

tylerrowsell:
When using `sqrt`, `cos` and `tan` I received type errors `Method sqrt does not exist on T.class_of(BigMath)`

It seems like these were incorrectly implemented and should have be implemented with `self.` like the log function this updates to match that.

### Motivation
Based on the ruby/math.rb file these should all be class methods.
https://github.com/ruby/ruby/blob/master/ext/bigdecimal/lib/bigdecimal/math.rb#L32

### Test plan
I am unsure how to test this. This is my first time contributing to this repo and I don't see any existing tests for this file.


froydnj (approved):
Thanks for the PR!  This looks like the right fix.

If you were so inclined, you can add a `# typed: true` `bigmath.rb` file to `test/testdata/rbi/`; there's some examples of what you could put in the file in:

https://github.com/sorbet/sorbet/blob/master/test/testdata/rbi/bigdecimal.rb
https://github.com/sorbet/sorbet/blob/master/test/testdata/rbi/array.rb

In this case, I guess calling the relevant functions in the file would be sufficient to ensure that Sorbet knows about `BigMath.PI` etc.

rbi/stdlib/big_math.rbi
```diff
@@ -92,7 +92,7 @@ module BigMath
     )
     .returns(BigDecimal)
   end
-  def atan(x, prec); end
+  def self.atan(x, prec); end
```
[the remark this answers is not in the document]
froydnj: Ah good point!  This would also make a good testcase to add along with the other suggested ones. :grin:

froydnj (changes requested):
Passing this back given @nvasilevski's point!

test/testdata/rbi/big_math.rb
```diff
@@ -0,0 +1,53 @@
...
+    sin(x, prec)
+  end
+
+  def my_sqrt(x, prec)
+    sqrt(x, prec)
+  end
+end
+
+BigMath.exp(BigDecimal("4"), 2)
+BigMath.log(BigDecimal("4"), 2)
```
tylerrowsell, line 34: It seems `log` and `exp` are not part of the module_function so didn't update them above but added a test case for them.

rbi/stdlib/big_math.rbi
```diff
@@ -92,7 +92,7 @@ module BigMath
     )
```
tylerrowsell, line 92: I am thinking the type for `x` should actually be `BigDecimal`. I am not sure the best way to confirm this? 

https://ruby-doc.org/stdlib-2.5.1/libdoc/bigdecimal/rdoc/BigMath.html

froydnj: I think this should be, at a minimum, `T.any(Integer, BigDecimal)`.  I don't know whether the type should be wider than that (e.g. should it include `Float`), but given that you are the first person to notice that this RBI file was wrong in the first place, the minimum fix is fine.

I think you also want the same treatment for all the non-`prec` arguments in this RBI?

froydnj (changes requested):
It looks like the types in the RBI file need to be adjusted for the tests.

froydnj:

→ https://go/builds/bui_LaO7RbjagOV53f
→ https://go/builds/bui_LaO7tu3lkT2skf

froydnj (approved):
This is great, thank you for the thorough investigation and tests!

PR #5663, opened 2022-04-20
give a little nudge for completing call sites that are not typed

froydnj:
### Motivation

One problem with Sorbet's LSP implementation is that when Sorbet can't return meaningful results (e.g. the file is not typed, or a particular call site is not typed), there's no information given to the user to indicate *why* there's no meaningful results.  (This is not necessarily Sorbet's fault; LSP is not quite rich enough to communicate this sort of information in a number of cases.)  From a user's perspective, this looks like Sorbet whimsically responds to their requests, sometimes returning interesting things and sometimes not.

This PR is an attempt to address one of those areas: completing method calls when the receiver is `T.untyped`.  We don't currently return anything for such cases (since you could literally be calling any method).  Instead of returning no results, this PR changes things so that we return a single item indicating that the call site is `T.untyped` and does no modification of the document if it is chosen.

The expectation/hope is that the user is motivated to figure out why the callsite is `T.untyped` so they can get better completion results.

Come to think of it, we probably also want to do a similar thing if our initial query results are empty because the file didn't have type information in the first place.  That could be a followup PR!

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/completion/untyped_receiver.rb
+++ b/test/testdata/lsp/completion/untyped_receiver.rb
@@ -4,5 +4,6 @@ extend T::Sig
 sig {params(x: T.untyped).void}
 def foo(x)
   x.
-  # ^ completion: (nothing)
+  # ^ completion: (call site is T.untyped)
+  # ^ apply-completion: [A] item: 0
 end # error: unexpected token
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -896,7 +896,7 @@ vector<unique_ptr<CompletionItem>> allSimilarConstantItems(const core::GlobalSta
 }
 
 vector<SimilarMethod> computeDedupedMethods(const core::GlobalState &gs, const core::DispatchResult &dispatchResult,
-                                             bool isPrivateOk, string_view prefix) {
+                                            bool isPrivateOk, string_view prefix) {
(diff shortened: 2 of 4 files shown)


test/testdata/lsp/completion/untyped_receiver.rb
```diff
@@ -4,5 +4,6 @@
 sig {params(x: T.untyped).void}
 def foo(x)
   x.
-  # ^ completion: (nothing)
+  # ^ completion: (call site is T.untyped)
+  # ^ apply-completion: [A] item: 0
```
jez, line 8: ooc, why is the rbedited empty? shouldn't it have identical contents to the input?

froydnj: Uh.  Apparently `update_testdata_exp.sh` is not updating the `.rbedited` files.  Let me fix this.

jez: ah yeah that's intentional, the testdata exp script is _only_ for `.*exp` files
it's designed that way so it doesn't have to build and test LSP.
what do you think about making a separate script, and then potentially a wrapper script that calls both? (we already have `update_exp_files.sh` which does _everything_, which is kind of slow because it builds ruby and runs the srb init tests...)

froydnj: `update_exp_files.sh` doesn't update the necessary files either.

PR #5664, opened 2022-04-20
give a little nudge for completion in untyped files

froydnj:
### Motivation

Followup to #5663: we can give a nudge on untyped files in addition to giving nudges on untyped receivers.  (And given the way the completion code works, we need to put the check for the former in a different place.)

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -1244,7 +1244,7 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerInterface &
         ENFORCE(fref.exists());
         auto level = fref.data(gs).strictLevel;
         if (level < core::StrictLevel::True) {
-            items.emplace_back(getCompletionItemForUntyped(gs, queryLoc, 0, "(file is not typed)"));
+            items.emplace_back(getCompletionItemForUntyped(gs, queryLoc, 0, "(file is not `# typed: true` or higher)"));
             response->result = make_unique<CompletionList>(false, move(items));
             return response;
         }
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -992,9 +992,9 @@ void logEmptyCompletion(const core::GlobalState &gs, core::FileRef fref, core::L
 CompletionTask::CompletionTask(const LSPConfiguration &config, MessageId id, unique_ptr<CompletionParams> params)
     : LSPRequestTask(config, move(id), LSPMethod::TextDocumentCompletion), params(move(params)) {}
 
-unique_ptr<CompletionItem> CompletionTask::getCompletionItemForUntypedReceiver(const core::GlobalState &gs,
-                                                                               core::Loc queryLoc, size_t sortIdx) {
-    string label("(call site is T.untyped)");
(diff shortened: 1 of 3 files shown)


main/lsp/requests/completion.cc
```diff
@@ -1209,6 +1209,13 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerInterface &
             }
         }
 
+        ENFORCE(fref.exists());
+        auto level = fref.data(gs).strictLevel;
+        if (level < core::StrictLevel::True) {
+            items.emplace_back(getCompletionItemForUntyped(gs, queryLoc, 0, "(file is not typed)"));
```
jez, line 1215: this leaves a little bit of ambiguity to people who are new to sorbet. we might want to say something like "(file is `# typed: false`)" or "(file is not `# typed: true` or higher)" to make it more clear.

jez (approved):

main/lsp/requests/completion.cc
froydnj: I was hoping that I wouldn't have to repeat the dance that hover needed to get the error message more precise, but I guess we should.

Unfortunately, I don't think we can use markdown for this, but maybe the markdown syntax is so ubiquitous people will be fine with it.

jez: yeah up to you

jez: Another thing you can consider is adding a documentation item so that when people click the (i) or press ^Space they'll see a longer explanation (that can use markdown, and also is not constrained to be a single short line)

elliottt (approved):

deiden26:
Is there any way to turn off this nudge? It's very annoying when using sorbet as an LSP alongside other tools: it prevents auto-completions I want to see from being displayed on files that I have no intention of adding types to

Kapture 2022-09-29 at 09 50 41

froydnj:
@deiden26 We don't have a way to turn this off at the moment.  Our primary use case has been a) the exclusive use of Sorbet as an LSP provider and b) encouraging people to at least get to `# typed: true` or better while helping them understand why they might not be seeing completion results on particular `# typed: false` or similar files.

Can you elaborate on why you're using Sorbet alongside other tools like solargraph, rather than using Sorbet exclusively?

deiden26:
> Can you elaborate on why you're using Sorbet alongside other tools like solargraph, rather than using Sorbet exclusively?

Thanks for getting back to me so quickly. I work at a company with a very large codebase that is gradually getting typed. It is going to be a long time until every file is typed, and often adding types to a file isn't in the scope of the work I'm trying to do. In untyped files, solargraph provides much better information than sorbet (as you can see in the gif). It also integrates with Rubocop so I can get realtime feedback on if I'm breaking my organization's style rules. It's a different tool that works well alongside sorbet (usually).

It would be nice if I could provide a flag to `sorbet tc --lsp` like `--no-nudges` to disable all suggestions like these that are nice-to-haves for some and actively frustrating for others.

tonybruess:
Hi, following up here, engineers at my org find this nudge frustrating. It interferes with copilot and other extensions. Would you be open to a PR that adds a flag to disable this?

PR #5666, opened 2022-04-20
Change error header if resolution error is package-related

aisamanra:
This changes the error _header_ for constant resolution failure _if_ we were able to figure out that there's a packaging-related problem that we can fix. In these cases, _we_ know that the error is related to packaging, but the error message is regularly confusing to users.

### Motivation
This makes certain errors a lot less confusing—after all, "unable to resolve constant" often makes people think that they've misspelled something, while these point the finger at packaging directly.

### Test plan

See included automated tests. Added a new test to explicitly check the error message when there's not an unambiguous import.


resolver/SuggestPackage.cc
```diff
@@ -291,6 +291,16 @@ bool SuggestPackage::tryPackageCorrections(core::Context ctx, core::ErrorBuilder
         for (auto match : missingImports) {
             pkgCtx.addMissingImportSuggestions(e, match);
         }
+
+        // going from an UnresolvedConstantLit to the full
+        // (with-scope) string is actually kinda annoying, so we're
+        // using the `loc` to get the original chunk of the file that
+        // includes the whole constant with scope
+        if (auto fullConstant = ctx.locAt(unresolved.loc).source(ctx)) {
+            // TODO (document me)
```
jez, line 300: what would you have liked to document here?

resolver/SuggestPackage.cc
```diff
@@ -304,6 +314,23 @@ bool SuggestPackage::tryPackageCorrections(core::Context ctx, core::ErrorBuilder
...
+                    // make the error message reference the exact
+                    // package we believe the constant came from
+                    auto missingExport = missingExports.front();
+                    auto pkgName = ctx.state.packageDB().getPackageInfo(missingExport.srcPkg).show(ctx);
+                    e.setHeader("Package `{}` does not export `{}`", pkgName, *fullConstant);
+                } else {
+                    // otherwise we should be explicit that it's
+                    // probably an export problem but be a bit cagey
+                    // about WHAT export problem it is
+                    e.setHeader("`{}` is not a public constant", *fullConstant);
```
jez, line 331: do you think it would be possible to somehow check whether the suggestion is like an "exact" match?

If I'm not mistaken, we're relying on "did you mean" suggestions here, which can suggest `SomeThing` if the user typed `Something`

(or is it more or less guaranteed that this code path only triggers for specifically packager related errors? maybe this is not worth optimizing for if trevor is going to be changing how this works shortly?)

resolver/SuggestPackage.cc
```diff
@@ -304,6 +314,23 @@ bool SuggestPackage::tryPackageCorrections(core::Context ctx, core::ErrorBuilder
...
+
+            if (auto fullConstant = ctx.locAt(unresolved.loc).source(ctx)) {
+                if (missingExports.size() == 1) {
+                    // if we have a single unambiguous export
+                    // suggestion, then we can use that information to
+                    // make the error message reference the exact
+                    // package we believe the constant came from
+                    auto missingExport = missingExports.front();
+                    auto pkgName = ctx.state.packageDB().getPackageInfo(missingExport.srcPkg).show(ctx);
+                    e.setHeader("Package `{}` does not export `{}`", pkgName, *fullConstant);
```
jez, line 326: Is it worth having an autocorrect here?

resolver/SuggestPackage.cc
aisamanra: Oh, that's a legacy of a slightly older approach—originally I was going to have two different error messages depending on whether we had one autocorrect (i.e. a "correct" one) or not, and I wanted to explain why. I don't think I need an extra comment there now.

resolver/SuggestPackage.cc
jez: or hm it looks like maybe the autocorrect is added earlier?

resolver/SuggestPackage.cc
```diff
@@ -291,6 +291,16 @@ bool SuggestPackage::tryPackageCorrections(core::Context ctx, core::ErrorBuilder
...
             pkgCtx.addMissingImportSuggestions(e, match);
         }
+
+        // going from an UnresolvedConstantLit to the full
+        // (with-scope) string is actually kinda annoying, so we're
+        // using the `loc` to get the original chunk of the file that
+        // includes the whole constant with scope
+        if (auto fullConstant = ctx.locAt(unresolved.loc).source(ctx)) {
+            // TODO (document me)
+            e.setHeader("No import provides `{}`", *fullConstant);
```
jez, line 301: do we only import packages? or do we import individual constants in packages?

should we mention "No imported package provides ..." or maybe "None of the imports in `{}` provide..."?

jez, line 310: should we add a comment somewhere that mentions that these `setHeader`'s actually overwrite any previously computed header?

jez (approved):

resolver/SuggestPackage.cc
aisamanra: I _believe_ this should only happen for packager-related errors, although I think it's not impossible that I'm missing some path that confuses it. (I do also think Trevor's changes are going to make that belief ironclad, so while I'd be happy to do some derisking, I do think it's got diminishing returns especially if we think we can land the new package approach relatively quickly.)

resolver/SuggestPackage.cc
aisamanra: We only import packages. I'd be happy to tweak the error here.

resolver/SuggestPackage.cc
aisamanra: We already get an autocorrect here. We actually add it a few lines up, here:

https://github.com/sorbet/sorbet/blob/60e7ec833ceb2ef8721d64149446fbd9003f1648/resolver/SuggestPackage.cc#L323-L325

resolver/SuggestPackage.cc
aisamanra: Added such a comment!

PR #5667, opened 2022-04-20
Remove MIN_LOOP_GLOBAL

jez:
### Motivation

The GlobalReassignmentTypeMismatchError error class was never being reported
because MIN_LOOP_GLOBAL was not really doing anything. We don't need to treat
these constants specially.

### Test plan

The ENFORCE I left behind says that the other half of the `if` could only be
type member symbols. In those cases:

- we already prevent reassignment to the symbols in namer (you can't assign to a
  constant with the same name twice)
- we technically do have two writes to the temporary variables, because all
  constant assignments like `X = type_member` get two writes:

  - `<C X>$1 = alias ::X`
  - `<C X>$1 = type_member`

  For the `type_member` case, we don't actually care about pinning, because
  right now the `alias` instruction pins the temporary variable to a meta type,
  while the `type_member` call always returns `T.untyped`. `T.untyped` is
  compatible with anything so there are never pinning errors we should be
  printing, even if we did somehow allow reassignments to constants.

Anyways, pretty sure that this is all just dead code.

### DIFF
--- a/test/testdata/infer/isa_generic.rb.cfg-text.exp
+++ b/test/testdata/infer/isa_generic.rb.cfg-text.exp
@@ -268,20 +268,20 @@ bb1[rubyRegionId=0, firstDead=-1]():
 # backedges
 # - bb0(rubyRegionId=0)
 # - bb5(rubyRegionId=1)
-bb2[rubyRegionId=1, firstDead=-1](<self>: T.class_of(Concrete), <C Klass>$10: Runtime object representing type: String, <block-pre-call-temp>$12: Sorbet::Private::Static::Void, <selfRestore>$13: T.class_of(Concrete)):
+bb2[rubyRegionId=1, firstDead=-1](<self>: T.class_of(Concrete), <block-pre-call-temp>$12: Sorbet::Private::Static::Void, <selfRestore>$13: T.class_of(Concrete)):
     # outerLoops: 1
     <block-call> -> (NilClass ? bb5 : bb3)
 
 # backedges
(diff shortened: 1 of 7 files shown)


test/testdata/infer/isa_generic.rb.cfg-text.exp
```diff
@@ -268,20 +268,20 @@ bb1[rubyRegionId=0, firstDead=-1]():
 # backedges
 # - bb0(rubyRegionId=0)
 # - bb5(rubyRegionId=1)
-bb2[rubyRegionId=1, firstDead=-1](<self>: T.class_of(Concrete), <C Klass>$10: Runtime object representing type: String, <block-pre-call-temp>$12: Sorbet::Private::Static::Void, <selfRestore>$13: T.class_of(Concrete)):
+bb2[rubyRegionId=1, firstDead=-1](<self>: T.class_of(Concrete), <block-pre-call-temp>$12: Sorbet::Private::Static::Void, <selfRestore>$13: T.class_of(Concrete)):
```
elliottt, line 271: Do these come back if there's a reference to `Klass` in the static-init for `Concrete`?

elliottt (approved):
Makes sense to me! Just a couple of questions.

test/testdata/infer/isa_generic.rb.cfg-text.exp
jez: Not only a reference, but an assignment, and it has to happen in a loop or block.

PR #5668, opened 2022-04-20
Document many missing errors in `error-reference.md` (part 1)

jez:
### Motivation

Writing more docs.

### Test plan

There are a ton of other undocumented errors. I'm working through the backlog.
For now, only check that new `infer/` errors get new docs.

I contemplated explicitly listing all the undocumented errors but figured it
would just be work I'm going to delete shortly.


website/docs/error-reference.md
```diff
@@ -1418,6 +1649,47 @@ MyMap[String, String].new
...
+  end
+end
+```
+
+If the instance variable can hold both an `Integer` and a `String`, maybe the
+type specified with `T.let` should be enlarged:
+
+```ruby
+@x = T.let(0, Object)
+```
```
froydnj, line 1684: Any reason to not use `T.any(Integer, String)` here?  (This is a minor point and not blocking to landing this.)

website/docs/error-reference.md
```diff
@@ -1440,6 +1712,35 @@ For more information, see [Troubleshooting](troubleshooting.md).
...
+
+Unfortunately, Sorbet can only perform this analysis when the depth is static.
+If Sorbet cannot see the exact value of the depth and instead only sees a type
+of `Integer` for the depth argument, it reports an error.
+
+Either:
+
+1.  Refactor the code so that Sorbet can see the exact depth value, or
+2.  Use `T.unsafe` to hide the method call from Sorbet.
```
froydnj, line 1742: WDYT about providing an example here?

website/docs/error-reference.md
```diff
@@ -1510,6 +1811,15 @@ T.unsafe(self).foo(xs)
...
+Note that method-level generics (i.e., those declared by using `type_parameters`
+inside a `sig`) are a somewhat unstable feature. If you encounter this error and
+believe it to represent a bug in Sorbet, first check the
+[Sorbet bug tracker](https://github.com/sorbet/sorbet/issues) to see if a
+similar-looking error has already been reported. If no such bug exists, feel
+free to open an issue.
+
 ## 7021
 
 The method called declares a block parameter that is not `T.nilable`, but a
```
froydnj, line 1825: (Noticed in passing) Kind of want to make this "The called method..." so it's more obvious that "called" is not a verb here.

website/docs/error-reference.md
```diff
@@ -1618,6 +1949,98 @@ def get_value(input)
...
+```
+
+This makes private methods in Ruby behave more like protected methods in other
+object-oriented languages.
+
+To fix this error, either:
+
+1.  Avoid calling the private method entirely, or
+2.  Update the method definition to not be private, or
+
```
froydnj, line 1988: I am eagerly awaiting the third option here. :sweat_smile:

website/docs/error-reference.md
```diff
@@ -1618,6 +1949,98 @@ def get_value(input)
...
+> statically, but because method visibility is checked **syntactically**, even
+> using `T.unsafe(self).foo` will cause a call to a private method `foo` to be
+> rejected by the Ruby VM at runtime.
+>
+> If you must call a private method and also silence any type errors from
+> calling it, you must use `self.send(:foo)` instead.
+
+## 7032
+
+When passing type arguments to generic classes, to use [shape types](shapes.md),
```
froydnj, line 2000: `:llamahorror:`

website/docs/error-reference.md
```diff
@@ -1638,6 +2061,55 @@ def foo(x, y)
...
+sig {void}
+def takes_no_block; end
+
+# BAD EXAMPLE
+takes_no_block {} # error: does not take a block
+```
+
+Why only sometimes? Technically all methods in Ruby are allowed to accept
+blocks. (Unlike positional arguments, if a method does not declare that it
+accepts a block argument but a caller passes one anyways the Ruby VM does not
```
froydnj, line 2082: I kind of want a comma after "anyways"?  Not sure.

tools/scripts/check_error_classes.sh
```diff
@@ -8,22 +8,45 @@ tmp_output="$(mktemp)"
...
+    [[ "$error_code" != 7* ]]; then
+    # TODO(jez) Document all other error codes
+    continue
+  fi
+
+  if ! grep -q "^## $error_code\$" website/docs/error-reference.md; then
+    echo "error: The error error '$error_code' in core/errors/ is not documented in website/docs/error-reference.md. Defined here:"
+    grep -r -n "$error_code" core/errors | indent4
+    had_error=1
+  fi
```
froydnj, line 49: :heart_eyes:

froydnj (approved):
I tried to add Github suggestions where I could.

website/docs/error-reference.md
jez: thought about it for a while, with a comma sounds more weird in my head htan without. maybe we should rewrite the whole sentence, but i'm just going to punt.

PR #5671, opened 2022-04-21
add a way to discover the `LocOffsets` for a `# typed:` sigil

froydnj:
### Motivation

This is another PR in service of #5598, split out for easier review and building on #5659.

I do not have strong opinions on where the offsets for the sigil should lie (e.g. just the level, the `typed: $LEVEL`, the entire comment, etc.), but I think that pointing at just the level makes the most sense if that's what we're going to point the user at.

### Test plan

See included automated tests.


core/Files.h
```diff
@@ -34,7 +34,11 @@ class File final {
     friend class GlobalState;
     friend class ::sorbet::core::serialize::SerializerImpl;
 
-    static StrictLevel fileStrictSigil(std::string_view source);
+    struct SigilInfo {
+        LocOffsets loc;
+        StrictLevel strictLevel;
+    };
```
jez, line 40: It looks like we don't have `CheckSize` annotations on `core::File`, but because `LocOffsets` is 8 bytes I'm guessing this will also add 8 bytes.

On Stripe's codebase, this would add something close to 1.4 MB to the file table in service of a feature that will be used only a handful of times. That strikes me as maybe optimizing for the wrong thing (making jump-to-def on a typed sigil fast) considering the alternative is to re-scan the file in service of the jump-to-def request. 

Curious what your thoughts are, or if I've done my math wrong.

froydnj: This is a good callout.

I think it doesn't matter that much in practice, because jemalloc will round up the 120 bytes that `File` currently takes to its 128-byte bucket (see the table under "Implementation Notes" at http://jemalloc.net/jemalloc.3.html), so there's no difference between 120 and 128 bytes from a memory perspective.

OTOH, there are a number of cheap ways to shrink `File` and having a `sigilLoc` function is probably a better alternative anyway, given that we are using this data only in LSP and only when something goes wrong.  I'll make that change, which will greatly simplify the patch.

jez (approved):

PR #5672, opened 2022-04-21
fix codeaction lsp crash

froydnj:
### Motivation

Noticed this when looking at rubocop autofixes on Stripe's codebase.

### Test plan

No tests given that we're interacting with external software here.  We could write some kind of protocol recording test, I guess?


elliottt (approved):
🚀

PR #5673, opened 2022-04-21
Add the mangled package name to files

elliottt:
Add the mangled package name to `core::File`, so that it's faster to retrieve later on in the pipeline. Set the mangled package name while building the package db.

### Motivation

Improve the performance of fetching a `PackageInfo` value for a file.

### Test plan

See included automated tests.

### DIFF
--- a/ast/substitute/Substitute.cc
+++ b/ast/substitute/Substitute.cc
@@ -128,15 +128,27 @@ public:
 };
 } // namespace
 
-ExpressionPtr Substitute::run(core::MutableContext ctx, const core::NameSubstitution &subst, ExpressionPtr what) {
+ParsedFile Substitute::run(core::MutableContext ctx, const core::NameSubstitution &subst, ParsedFile what) {
     SubstWalk walk(subst);
-    what = TreeMap::apply(ctx, walk, std::move(what));
+    what.tree = TreeMap::apply(ctx, walk, std::move(what.tree));
+
+    auto pkg = what.file.data(ctx).getPackage();
+    if (pkg.exists()) {
+        what.file.data(ctx).setPackage(subst.substitute(pkg));
+    }
+
     return what;
 }
 
-ExpressionPtr Substitute::run(core::MutableContext ctx, core::LazyNameSubstitution &subst, ExpressionPtr what) {
+ParsedFile Substitute::run(core::MutableContext ctx, core::LazyNameSubstitution &subst, ParsedFile what) {
     SubstWalk walk(subst);
-    what = TreeMap::apply(ctx, walk, std::move(what));
+    what.tree = TreeMap::apply(ctx, walk, std::move(what.tree));
+
+    auto pkg = what.file.data(ctx).getPackage();
+    if (pkg.exists()) {
+        what.file.data(ctx).setPackage(subst.substitute(pkg));
+    }
+
     return what;
 }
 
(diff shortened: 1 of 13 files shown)


elliottt:

→ https://go/builds/bui_LYETfb2Du0n85t
→ https://go/builds/bui_LYEUYcYgT4fqJC

core/Files.h
```diff
@@ -104,6 +110,8 @@ class File final {
 
     Flags flags;
 
+    NameRef package = core::NameRef::noName();
```
elliottt, line 113: This is what's causing the bump in size for `File`.

ast/substitute/Substitute.cc
```diff
@@ -128,15 +128,27 @@ template <typename T> class SubstWalk {
...
-ExpressionPtr Substitute::run(core::MutableContext ctx, const core::NameSubstitution &subst, ExpressionPtr what) {
+ParsedFile Substitute::run(core::MutableContext ctx, const core::NameSubstitution &subst, ParsedFile what) {
     SubstWalk walk(subst);
-    what = TreeMap::apply(ctx, walk, std::move(what));
+    what.tree = TreeMap::apply(ctx, walk, std::move(what.tree));
+
+    auto pkg = what.file.data(ctx).getPackage();
+    if (pkg.exists()) {
+        what.file.data(ctx).setPackage(subst.substitute(pkg));
+    }
```
jez, line 138: oh wow i would not have thought to update this. did this get caught by our tests?

core/Files.h
jez: Do you think it's worth leaving a comment here like "this incurs a memory overhead even if the `--stripe-packages` flag is not passed at the command line. In the future we may want to consider ways to pay this cost only if that flag is passed at the command line."

ast/substitute/Substitute.cc
elliottt: It wasn't, but I was thinking about it after having modified the substitution back in December :)

resolver/SuggestPackage.cc
```diff
@@ -39,7 +39,8 @@ class PackageContext final {
     core::Context ctx;
     const core::packages::PackageInfo &currentPkg;
 
-    PackageContext(core::Context ctx) : ctx(ctx), currentPkg(ctx.state.packageDB().getPackageForFile(ctx, ctx.file)) {}
```
jez, line 42: i didn't check: can we delete `getPackageForFile`?

elliottt: Unfortunately not yet. It's still used to find the name that's cached on the `File`.

jez: oh  nvm, i see that it's still being used, but just to set the thing, so we can avoid calling it in weird places

could we make it private so that people don't call it?

elliottt: Good point, I'll have a look 👍

jez (commented):
looks great, thanks!

jez (approved):

core/Files.h
elliottt: I can add a note to that effect 👍

resolver/SuggestPackage.cc
elliottt: I think it could be made to be private, but it would take some plumbing: there are some functions in the anonymous namespace of the packager that use it :/

jez: that's fine, np

PR #5674, opened 2022-04-21
shrink `File` and add size assertions

froydnj:
### Motivation

jez raised concerns that some changes to `File` in #5671 would expand it, which prompted me to look and notice that we can shrink `File` quite a bit.  So this is worth a megabyte or two on Stripe's codebase, which is not huge, but counts for something.

As a bonus, I think that #5673 won't have any effect on the size even after this PR :tada:.

### Test plan

See included automated tests.

### DIFF
--- a/core/Files.h
+++ b/core/Files.h
@@ -115,7 +115,7 @@ public:
 
     const CompiledLevel compiledLevel;
 };
-CheckSize(File, 112, 8);
+CheckSize(File, 104, 8);
 
 template <typename H> H AbslHashValue(H h, const FileRef &m) {
     return H::combine(std::move(h), m.id());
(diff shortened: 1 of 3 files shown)


core/Files.h
```diff
@@ -107,14 +107,17 @@ class File final {
...
-    std::shared_ptr<const FileHash> hash_;
 
 public:
     const StrictLevel originalSigil;
     StrictLevel strictLevel;
 
     const CompiledLevel compiledLevel;
+
+private:
+    std::shared_ptr<const FileHash> hash_;
```
jez, line 118: This is not relevant to the size changes, right? It's just a "happen to notice"?

froydnj: This is moving the location of the `hash_` member within the class so that `originalSigil`, `StrictLevel`, and `compiledLevel` pack into the same word as `minErrorLevel_`.  `hash_` was `private` before and is `private` after (unless I screwed something up).

PR #5682, opened 2022-04-22
move `LSPConfiguration::lspPos2Loc` to `Position::asLoc`

froydnj:
### Motivation

This function doesn't really belong on `LSPConfiguration`; having it on `Position` makes a lot more sense.

### Test plan

See included automated tests.


jez (approved):
lgtm but slight preference for calling this toLoc (I usually equate “as” with casts that don’t change the data)

PR #5686, opened 2022-04-24
Drop two errors back down to `typed: false`

jez:
### Motivation

We had a TODO for these errors to fix Stripe's codebase so that we could drop these errors back down to `# typed: false`.

It looks like there are only three violations for this error now. I've prepped a PR to Stripe's codebase to fix those (replacing the offenses with `T.untyped` mostly).

### Test plan

n/a


jez:

→ https://go/builds/bui_LZKTAbcq53H7vB
→ https://go/builds/bui_LZKTXM6xcWmqi8

elliottt (approved):
Looks good to me! There are a few test failures, but it looks like they're just instances of this error that weren't raising before.

PR #5688, opened 2022-04-25
Document all remaining error codes (part 3)

jez:
### Motivation

This finally enforces that all error classes have an entry in
`error-reference.md` on the website.

### Test plan

See changes to `tools/scripts/check_error_classes.sh`.

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1120,8 +1120,8 @@ Parent::MY_CONST   # ok
 
 ## 5003
 
-Sorbet failed to parse a method signature. For more documentation on 
-valid `sig` syntax, see [Method Signatures](sigs.md).
+Sorbet failed to parse a method signature. For more documentation on valid `sig`
+syntax, see [Method Signatures](sigs.md).
 
 ## 5004
 
@@ -1418,8 +1418,7 @@ So you do not need to manually insert a `raise` inside the body of an abstract
 method.
 
 If you would like to define a method in an abstract class or interface with a
-default implementation, use the `overridable` annotation on the
-signature:
+default implementation, use the `overridable` annotation on the signature:
 
 ```ruby
 class AbstractService
@@ -1444,8 +1443,8 @@ There are a few constraints around how methods like `include`, `extend`,
     defined in `*.rb` or `*.rbi` files that are not being ignored, or
   - Hide the `include` or `extend` invocations from Sorbet by using
     [`# typed: ignore` sigils](static.md) to ignore the entire file, or
-  - Hide only the individual call from Sorbet statically using something
-    like `send(:include, ...)`.
+  - Hide only the individual call from Sorbet statically using something like
+    `send(:include, ...)`.
 - `mixes_in_class_methods` and `requires_ancestor` must only be declared in a
   `module`, not a `class`. Classes are never mixed into other classes or
   modules, so these methods would have no meaning in a class.
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1120,7 +1120,7 @@ Parent::MY_CONST   # ok
 
 ## 5003
 
-Sorbet failed to parse a method signature. For more documentation on what's
+Sorbet failed to parse a method signature. For more documentation on 
 valid `sig` syntax, see [Method Signatures](sigs.md).
 
 ## 5004
@@ -1348,14 +1348,14 @@ Eric Lippert, who worked on the design and implementation of C# for many years.
 →
 [Why isn't there variance for generic classes?](https://stackoverflow.com/questions/2733346/why-isnt-there-generic-variance-for-classes-in-c-sharp-4-0/2734070#2734070)
 
-The answer concludes that the main selling point of having covariant classes it
+The answer concludes that the main selling point of having covariant classes is
 to make immutable generic classes, at the cost of a more complex implementation
 of generics.
 
 ## 5017
 
 The `type_member` and `type_template` declarations from a parent class must all
-be repeated in the same order a child class (and before any newly-added type
+be repeated in the same order in a child class (and before any newly-added type
 members belonging only to the child class).
 
 One non-obvious way this error can manifest is via code generation tooling. If a
@@ -1388,7 +1388,7 @@ other kind of constant in the child.
 
 Abstract methods must not have bodies. For more information, see
 [Abstract Classes and Interfaces](abstract.md). Despite these methods not having
-a body, Sorbet's [runtime support](runtime.md) via the `sorbet-runtime` will
+a body, Sorbet's [runtime support](runtime.md) via the `sorbet-runtime` gem will
 convert these methods to raise. For example:
 
 ```ruby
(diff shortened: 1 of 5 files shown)


website/docs/override-checking.md
```diff
@@ -41,6 +41,51 @@ Note that the **absence** of `abstract` or `overridable` does **not** mean that
 a method is never overridden. To declare that a method can never be overridden,
 look into [final methods](final.md).
 
+## A note on variance
+
+When overriding a method, the override must accept at all the same things that
+the parent method accepts, and return at most what the parent method returns but
+no more.
+
+This is very abstract so let's make it concrete with some examples:
```
froydnj, line 50: The only example we give is one with argument variance; should we also give an example with return value variance as well?

website/docs/error-reference.md
```diff
@@ -1118,6 +1118,17 @@ Child::MY_CONST    # ok
 Parent::MY_CONST   # ok
 ```
 
+## 5003
+
+Sorbet failed to parse a method signature. For more documentation on what's
+valid `sig` syntax, see [Method Signatures](sigs.md).
```
froydnj, line 1124:
```suggestion
Sorbet failed to parse a method signature. For more documentation on 
valid `sig` syntax, see [Method Signatures](sigs.md).
```

website/docs/error-reference.md
```diff
@@ -1255,6 +1266,208 @@ end
...
+
+**Why?** The design of generic classes and interfaces in Sorbet was heavily
+inspired by the design of C#. This exact question has been answered for C# by
+Eric Lippert, who worked on the design and implementation of C# for many years.
+
+→
+[Why isn't there variance for generic classes?](https://stackoverflow.com/questions/2733346/why-isnt-there-generic-variance-for-classes-in-c-sharp-4-0/2734070#2734070)
+
+The answer concludes that the main selling point of having covariant classes it
+to make immutable generic classes, at the cost of a more complex implementation
```
froydnj, line 1352:
```suggestion
The answer concludes that the main selling point of having covariant classes is
to make immutable generic classes, at the cost of a more complex implementation
```

website/docs/error-reference.md
```diff
@@ -1255,6 +1266,208 @@ end
...
+of generics.
+
+## 5017
+
+The `type_member` and `type_template` declarations from a parent class must all
+be repeated in the same order a child class (and before any newly-added type
```
froydnj, line 1358:
```suggestion
be repeated in the same order in a child class (and before any newly-added type
```

website/docs/error-reference.md
```diff
@@ -1255,6 +1266,208 @@ end
...
+
+In this example, since `X` is declared as a `type_member` in the parent, it must
+also be declared as a `type_member` in the child, and cannot be changed to some
+other kind of constant in the child.
+
+## 5019
+
+Abstract methods must not have bodies. For more information, see
+[Abstract Classes and Interfaces](abstract.md). Despite these methods not having
+a body, Sorbet's [runtime support](runtime.md) via the `sorbet-runtime` will
```
froydnj, line 1391:
```suggestion
a body, Sorbet's [runtime support](runtime.md) via the `sorbet-runtime` gem will
```

Or "via `sorbet-runtime` will..."; I'm not sure what the right fix here is.

website/docs/error-reference.md
```diff
@@ -1255,6 +1266,208 @@ end
...
+# example.rb:16:
+#   The method `port` on `IService` is declared as `abstract`.
+#   It does not have an implementation.
+```
+
+So you do not need to manually insert a `raise` inside the body of an abstract
+method.
+
+If you would like to define a method in an abstract class or interface with a
+default implementation, feel free to use the `overridable` annotation on the
```
froydnj, line 1421:
```suggestion
default implementation, use the `overridable` annotation on the
```

website/docs/error-reference.md
```diff
@@ -1459,6 +1928,136 @@ class ChildTwo < T::Struct
...
+```ruby
+# GOOD
+X = T.type_alias {Integer}
+```
+
+The new syntax avoids incurring eagerly evaluating the right-hand side of the
+type alias, potentially forcing one or more constants to be autoloaded before
+they would otherwise be required. This improves load-time performance in
+lazily-loaded environments (usually: development environments) and also prevents
+certain modes of use of Sorbet to introduce load-time cyclic references.
```
froydnj, line 1964:
```suggestion
certain modes of the use of Sorbet to introduce load-time cyclic references.
```

I'm not sure what the right fix here is, but "modes of use of Sorbet" sounds kind of weird to me.

website/docs/error-reference.md
```diff
@@ -1459,6 +1928,136 @@ class ChildTwo < T::Struct
...
+members—classes are limited to only invariant type members. For more, see the
+docs for error code [5016](#5016).
+
+> **Note** that `T.attached_class` is actually modeled as a covariant (`:out`)
+> `type_template` defined automatically on all singleton classes, which means
+> that `T.attached_class` can only be used in `:out` positions.
+
+## 5045
+
+This error is regards misuse of user-defined generic classes. This error is
```
froydnj, line 2057:
```suggestion
This error is regarding misuse of user-defined generic classes. This error is
```

website/docs/error-reference.md
```diff
@@ -1515,16 +2114,114 @@ class C
...
 
 Use of `implementation` has been replaced by `override`.
 
-## 5056
+## 5055
+
+Sorbet does not consider a class "fully resolved" until all of its type members
+have also been fully resolved. Sorbet cannot fully resolve a type member in a
+class until all constants mentioned in its right-hand side have been fully
+resolved. So if any of those right-hand-side constants are themselves generic
```
froydnj, line 2202: "Right-hand{ ,-}side feels like a little too much Sorbet internals talk, but I'm not sure what an equally descriptive term would be in a user context.

froydnj (approved):
Lots of little suggestions, but no complaints overall.  I think the patch might need to be rebased, or something, because there are some test failures not related to documentation?

website/docs/error-reference.md
jez, line 1964:
```suggestion
certain Sorbet usage patterns from introducing load-time cyclic references.
```

website/docs/error-reference.md
jez, line 2057:
```suggestion
This error regards misuse of user-defined generic classes. This error is
```

website/docs/error-reference.md
jez: committed an alternative phrasing.

website/docs/error-reference.md
```diff
@@ -1515,16 +2114,114 @@ class C
 end
 ```
 
+## 5049
+
+Parameters given type annotations in a method signature must be provided in the
```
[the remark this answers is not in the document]
jez: It's the parameters themselves that have been given annotations, so i think the `as` is not quite right here.

website/docs/error-reference.md
jez: maybe? it's hopefully clear from context what the right and left hand side of a type member declaration is. the first time someone asks about it i'm happy to think of an alternate phrasing.

PR #5695, opened 2022-04-25
be more careful with read errors from the watchman process

froydnj:
### Motivation

We occasionally see people get stuck with error messages in vscode that persist over time.  We're not entirely sure why they happen -- finding a reproducible testcase is hard -- but one plausible theory that's been put forth is that we don't do a good job of handling watchman errors.

In particular, we pass an exit handler into the watchman process manager, and the code path touched in this PR doesn't call this exit handler.  So the theorized error situation looks like:

1. we try to read from the watchman process.
2. we get some kind of error from `select(2)` or `read(2)`
3. we break out of the "reading from watchman" loop
4. the watchman process's manager thread exits
5. we are now no longer observing file changes, and our main loop has not noticed that watchman has gone away.

Hence, this PR: we notify the main loop that "hey, watchman had a problem" in addition to breaking out of the thread.  The theory is that this should force the main loop (or whatever is managing the `sorbet` process) to restart and begin handling file changes again.

### Test plan

No additional tests beyond thinking real hard about the changes.

### DIFF
--- a/main/lsp/watchman/WatchmanProcess.cc
+++ b/main/lsp/watchman/WatchmanProcess.cc
@@ -62,14 +62,24 @@ void WatchmanProcess::start() {
         string buffer;
 
         while (!isStopped()) {
+            errno = 0;
             auto maybeLine = FileOps::readLineFromFd(fd, buffer);
             if (maybeLine.result == FileOps::ReadResult::Timeout) {
                 // Timeout occurred. See if we should abort before reading further.
                 continue;
-            } else if (maybeLine.result == FileOps::ReadResult::ErrorOrEof) {
+            }
+
+            if (maybeLine.result == FileOps::ReadResult::ErrorOrEof) {
+                if (errno == EINTR) {
+                    continue;
+                }
+
                 // Exit loop; unable to read from Watchman process.
+                exitWithCode(1, "Unknown error while reading from Watchman process");
                 break;
             }
+
+            ENFORCE(maybeLine.result == FileOps::ReadResult::Success);
 
             const string &line = *maybeLine.output;
             // Line found!
@@ -106,6 +116,8 @@ void WatchmanProcess::start() {
             exit(1);
         }
     }
+
+    ENFORCE(isStopped());
 }
 
 bool WatchmanProcess::isStopped() {


jez (approved):

elliottt (approved):
This makes sense to me, thanks for the writeup about the scenario it addresses!

Will you follow up by changing protocol.cc to handle the unexpected exit? It looks like currently it will terminate the watchman queue and log an error, but not attempt to restart anything.

froydnj:
> Will you follow up by changing protocol.cc to handle the unexpected exit? It looks like currently it will terminate the watchman queue and log an error, but not attempt to restart anything.

I don't think we have to change `protocol.cc` in any way.  This was the handler that expected to be called when the watchman thread would terminate:

https://github.com/sorbet/sorbet/blob/2096a00f35929fcb8ea31373c5763481ab456b1d/main/lsp/protocol.cc#L160-L175

so that will terminate the `messageQueue`, which is good!  In the scenario described above, we were never calling that handler and therefore never setting the appropriate bit.

It's not obvious, but other things want to know that `messageQueue` is terminated, like the preprocessor:

https://github.com/sorbet/sorbet/blob/2096a00f35929fcb8ea31373c5763481ab456b1d/main/lsp/protocol.cc#L71-L113

so the preprocessor will shut down when one of two conditions is met:

1. The message queue is terminated and drained; or
2. The task queue is terminated.

The first condition was never getting met in the scenario this PR describes, because we were never setting the termination flag on `messageQueue`.  So after this PR, we should be setting that termination flag, which will trigger the (eventual) termination of the preprocessor thread.  And once *that* happens, the `TerminateOnDestruction notifyProcessing(*taskQueue);` destructor will kick in:

https://github.com/sorbet/sorbet/blob/2096a00f35929fcb8ea31373c5763481ab456b1d/main/lsp/protocol.cc#L34-L43

and terminate the task queue, which already has a bunch of logic for being properly handled in `protocol.cc`.

And then that will eventually cause `runLSP` to return, which will exit the Sorbet process, which should lead to the VSCode extension or whatever noticing and restarting the Sorbet process, which is what we want.

PR #5697, opened 2022-04-26
Add TypeScript ↔ Sorbet reference doc

jez:
### Motivation

Many people come to Sorbet from a TypeScript background. Having a table
like this should hopefully give people quicker answers when they're
onboarding.

### Test plan

See included automated tests.

### DIFF
--- a/website/docs/from-typescript.md
+++ b/website/docs/from-typescript.md
@@ -1,0 +1,326 @@
+---
+id: from-typescript
+title: TypeScript ↔ Sorbet Reference
+sidebar_label: TypeScript ↔ Sorbet
+---
+
+Many people learning Sorbet for the first time come from a TypeScript
+background. We've built this quick reference table mapping some common
+TypeScript features to the corresponding feature in Sorbet. (This is not a
+document about automatically interoperating between Sorbet and TypeScript; it is
+merely provided for educational purposes.)
+
+Note that Sorbet is not affiliated with TypeScript in any official way, and the
+information in this table may become out of date as TypeScript evolves
+independently of Sorbet. Please use the "Edit" button at the top of the page to
+help correct inaccuracies.
+
+Also note that Sorbet features do not always map to an exactly equivalent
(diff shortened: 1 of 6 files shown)


website/docs/from-typescript.md
```diff
@@ -0,0 +1,326 @@
...
+      <td>
+        <code>Integer</code>, <code>Float</code>, <code>Numeric</code>
+      </td>
+      <td>
+        Ruby has <a href="/docs/class-types">separate classes</a> for different
+        numeric types. <code>number</code> in JavaScript is most closely
+        <code>Float</code> in Ruby, but unlike in JavaScript, integers and
+        floating points numbers are not transparently compatible with each
+        other. To convert between different types, Ruby provides various "to_i"
+        and "to_f" methods. See the Ruby docs for more.
```
froydnj, line 63:
```suggestion
        other. To convert between different types, Ruby provides various <code>to_i</code>
        and <code>to_f</code> methods. See the Ruby docs for more.
```

website/docs/from-typescript.md
```diff
@@ -0,0 +1,326 @@
...
+        <code>number</code>
+      </td>
+      <td>
+        <code>Integer</code>, <code>Float</code>, <code>Numeric</code>
+      </td>
+      <td>
+        Ruby has <a href="/docs/class-types">separate classes</a> for different
+        numeric types. <code>number</code> in JavaScript is most closely
+        <code>Float</code> in Ruby, but unlike in JavaScript, integers and
+        floating points numbers are not transparently compatible with each
```
froydnj, line 61:
```suggestion
        floating-point numbers are not transparently compatible with each
```

"floating points numbers" sounds strange to me; I think the "s" should be dropped, but it's also possible that I have gotten the need for the hyphen incorrect.

froydnj (approved):
I don't really know TypeScript; you may want a second opinion on the equivalencies.

PR #5700, opened 2022-04-26
stop doing extra work in `tryConvertToString`

froydnj:
### Motivation

`GetString()` returns a `char *`, so the previous code required an extra `strlen` to construct the returned string.  But this is silly; the JSON values know their length, so we can provide that directly and avoid a `strlen` call.

### Test plan

See included automated tests.


elliottt (approved):
Nice find!

PR #5705, opened 2022-04-26
Some more website changes

jez:
### Test plan

See included automated tests.

### DIFF
--- a/website/docs/cli.md
+++ b/website/docs/cli.md
@@ -100,8 +100,8 @@ and `/bar/foo/baz.rb` but not `/foo.rb` or `/foo2/bar.rb`.
 
 ## Accepting autocorrect suggestions
 
-From time to time, Sorbet suggests autocorrects that can be accepted to
-automatically fix certain errors. For example:
+For certain errors, Sorbet suggests autocorrects that can be accepted to
+automatically fix those errors. For example:
 
 ```ruby
 sig {params(xs: T::Array[Integer]).void}
@@ -119,7 +119,7 @@ because `xs[0]` might be `nil` if the `xs` list is empty.
 
 In this example, though, we know that the `xs` array will always be non-empty.
 If, as the programmer, we know this invariant will always hold, we can use
-`T.must` to fix the problem. Sorbet even suggests that the error message:
+`T.must` to fix the problem. Sorbet even suggests that in the error message:
 
 ```
 editor.rb:6: Method `+` does not exist on `NilClass` component of `T.nilable(Integer)` https://srb.help/7003
@@ -158,8 +158,8 @@ The `--isolate-error-code` option can be repeated with different error codes to
 have Sorbet apply autocorrects for all the mentioned error codes, but no others.
 
 There is no way to have Sorbet **only** apply autocorrect suggestions in a given
-file. Instead, use a version control system to undo changes autocorrect changes
-Sorbet makes in files that should not be changed.
+file. Instead, use a version control system to undo autocorrect changes Sorbet
+makes in files that should not be changed.
 
 ### Silencing errors in bulk
 
@@ -248,9 +248,9 @@ end
 ```
 
 A method named like this sticks out in the codebase as a flag to potential
-developers that something has been automatically silenced vs been manually
-silenced. Placing a comment on the method shows the contextual message when
-people hover over any call site.
+developers that something has been automatically silenced instead of being
+manually silenced. Placing a comment on the method shows the contextual message
+when people hover over any call site.
 
 ## Overriding strictness levels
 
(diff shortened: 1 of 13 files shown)


website/docs/cli.md
```diff
@@ -98,6 +98,160 @@ against the prefix of these relative paths; others are substring matches.
 Matches must be against whole component parts, so `foo` matches `/foo/bar.rb`
 and `/bar/foo/baz.rb` but not `/foo.rb` or `/foo2/bar.rb`.
 
+## Accepting autocorrect suggestions
+
+From time to time, Sorbet suggests autocorrects that can be accepted to
+automatically fix certain errors. For example:
```
froydnj, line 104:
```suggestion
For certain errors, Sorbet suggests autocorrects that can be accepted to
automatically fix said errors. For example:
```

I don't know what the right phrasing is, but "From time to time" evokes Sorbet doing things on a whim for me.

website/docs/cli.md
```diff
@@ -98,6 +98,160 @@ against the prefix of these relative paths; others are substring matches.
...
+```
+
+[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28xs%3A%20T%3A%3AArray%5BInteger%5D%29.void%7D%0Adef%20example%28xs%29%0A%20%20xs%5B0%5D%20%2B%201%0Aend%0A%0Aexample%28%5B1%2C%202%2C%203%5D%29)
+
+In this example, Sorbet does not think that the `+` method always exists,
+because `xs[0]` might be `nil` if the `xs` list is empty.
+
+In this example, though, we know that the `xs` array will always be non-empty.
+If, as the programmer, we know this invariant will always hold, we can use
+`T.must` to fix the problem. Sorbet even suggests that the error message:
```
froydnj, line 122:
```suggestion
`T.must` to fix the problem. Sorbet even suggests that with the error message:
```

website/docs/cli.md
```diff
@@ -98,6 +98,160 @@ against the prefix of these relative paths; others are substring matches.
...
+
+```
+❯ srb tc --autocorrect --isolate-error-code=7003
+```
+
+The `--isolate-error-code` option can be repeated with different error codes to
+have Sorbet apply autocorrects for all the mentioned error codes, but no others.
+
+There is no way to have Sorbet **only** apply autocorrect suggestions in a given
+file. Instead, use a version control system to undo changes autocorrect changes
```
froydnj, line 161:
```suggestion
file. Instead, use a version control system to undo autocorrect changes
```

website/docs/cli.md
```diff
@@ -98,6 +98,160 @@ against the prefix of these relative paths; others are substring matches.
...
+  #
+  # If you see this method used in a file your team owns, consider removing the
+  # call and fixing the resulting error, or replacing it with `T.unsafe` if the
+  # error cannot be fixed.
+  def self.sorbet_0_5_1234(x); x; end
+end
+```
+
+A method named like this sticks out in the codebase as a flag to potential
+developers that something has been automatically silenced vs been manually
```
froydnj, line 251:
```suggestion
developers that something has been automatically silenced instead of being manually
```

This is less parallel, but avoids the "vs" abbreviation, which I think is marginally better.  Feel free to rewrite to a more parallel construction.

website/docs/exhaustiveness.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: exhaustiveness
-title: Exhaustiveness Checking
+title: Exhaustiveness (T.absurd)
```
froydnj, line 3: These extra little bits to express the labels in terms of user concepts instead of internals concepts are great.

website/docs/intersection-types.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: intersection-types
-title: Intersection Types
+title: T.all
```
froydnj, line 3:
```suggestion
title: Intersection Types
sidebar_label: Intersection Types (T.all)
```

The equivalent change for "Union Types" keeps the current title, but changes the sidebar label.  I don't have a strong opinion here, but symmetry between the two seems like a good idea.

website/docs/why-type-annotations.md
```diff
@@ -0,0 +1,162 @@
...
+   this annotation useless?"), and also
+2. we wouldn't be able to easily build good error messages to suggest people to
+   do this in the first place. Instead, the error messages would be reported far
+   downstream from where the error actually happened, and be reported as
+   something confusing like "this thing might sometimes be `nil`."
+
+## Type annotations and strictness levels
+
+Sorbet allows the programmer to opt-in to greater levels of static type rigor.
+At lower [strictness modes](static.md), Sorbet allows definitions to be
```
froydnj, line 140:
```suggestion
At lower [strictness levels](static.md), Sorbet allows definitions to be
```

froydnj (approved):

PR #5707, opened 2022-04-27
better locs for string attrs

froydnj:
### Motivation

This is split off from #5706: while working on that PR, I noticed that `attr_reader` and friends with string names had locs that were slightly different than when using symbol names.  This PR brings the two closer together.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/AttrReader.cc
+++ b/rewriter/AttrReader.cc
@@ -39,6 +39,16 @@ pair<core::NameRef, core::LocOffsets> getName(core::MutableContext ctx, ast::Exp
                 res = core::Names::empty();
             }
             loc = lit->loc;
+            if (::sorbet::debug_mode) {
+                auto l = ctx.locAt(loc);
+                ENFORCE(l.exists());
+                auto source = l.source(ctx).value();
+                ENFORCE(source.size() > 2);
+                ENFORCE(source[0] == '"' || source[0] == '\'');
+                auto lastChar = source[source.size()-1];
+                ENFORCE(lastChar == '"' || lastChar == '\'');
+            }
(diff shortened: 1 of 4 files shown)


rewriter/AttrReader.cc
```diff
@@ -39,6 +39,16 @@ pair<core::NameRef, core::LocOffsets> getName(core::MutableContext ctx, ast::Exp
                 res = core::Names::empty();
             }
             loc = lit->loc;
+            if (::sorbet::debug_mode) {
```
jez, line 42: Why are we casing on debug mode here?

jez: Ah I see, this is to support some ENFORCE. Usually we use the DEBOG_ONLY macro for this.

rewriter/AttrReader.cc
```diff
@@ -39,6 +39,16 @@ pair<core::NameRef, core::LocOffsets> getName(core::MutableContext ctx, ast::Exp
...
+                auto l = ctx.locAt(loc);
+                ENFORCE(l.exists());
+                auto source = l.source(ctx).value();
+                ENFORCE(source.size() > 2);
+                ENFORCE(source[0] == '"' || source[0] == '\'');
+                auto lastChar = source[source.size() - 1];
+                ENFORCE(lastChar == '"' || lastChar == '\'');
+            }
+            loc = core::LocOffsets{loc.beginPos() + 1, loc.endPos() - 1};
```
jez, line 51: Can we use Loc::adjust here? What if the string’s Loc is is empty somehow?

jez (approved):

PR #5708, opened 2022-04-27
delete superfluous regex `?`

froydnj:
### Motivation

I guess the `<regex>` header that we use doesn't complain about this, but it's a little confusing reading the regex and wondering why this particular capture expression has `??` after it.

### Test plan

Ideally all the tests that we have should still pass with this change, since we would throw an error if the regex did not match for some reason.


elliottt (approved):
I'm pretty impressed that you found this!

PR #5713, opened 2022-04-28
test the placeholder text in `PrepareRenameResponse`

froydnj:
### Motivation

There is a TODO in `position_assertions.cc` about testing the placeholder text that we return in `textDocument/prepareRename`.  This text is shown to the user (at least in VSCode), so it would be good to test that we are showing the user the thing we expect to be showing them.

I think this would have also cleared up some of the confusion in #4734 around identifier locations: the tests would have required `placeholderText`, and it would have been a little more obvious what we were trying to fix with the fiddling around the location of `IdentResponse`, or at least what would break if we had used the unadulterated location of the `IdentResponse`.

cc @vinistock

### Test plan

See included automated tests.


vinistock (approved):
Nice!

PR #5714, opened 2022-04-28
provide error documentation in the `codeDescription` field

froydnj:
### Motivation

We currently provide a link to the error documentation via `Diagnostic.relatedInformation.location`, which requires us to provide a useless location within the "file" and a message directing the user to click for more information.  Since LSP 3.16, there's `Diagnostic.codeDescription` which is just the URI to the documentation.  This latter method is smaller in the over-the-wire protocol (less repeated/useless information) and displays slightly more nicely in the UI.

Before:

image

After:

image

### Test plan

See included automated tests.


jez:
This is great!!

PR #5718, opened 2022-04-28
Add the `packaged` sigil to files

elliottt:
Like the `compiled` sigil this is Stripe-specific, but will help with landing some improvements to the performance of the current packages implementation.

### Motivation

Pre-work for the package visibility checker.

### Test plan

See included automated tests.

### DIFF
--- a/core/Files.h
+++ b/core/Files.h
@@ -115,6 +115,10 @@ private:
 
     Flags flags;
 
+private:
+    const PackagedLevel packagedLevel;
+
+public:
     const std::string path_;
     const std::string source_;
     mutable std::shared_ptr<std::vector<int>> lineBreaks_;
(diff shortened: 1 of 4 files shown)


core/Files.h
```diff
@@ -126,11 +131,14 @@ class File final {
...
     const CompiledLevel compiledLevel;
+    const PackagedLevel packagedLevel;
 
 private:
     std::shared_ptr<const FileHash> hash_;
 };
-CheckSize(File, 96, 8);
+
+// TODO: see if we can pack `packagedLevel` in a bit better to get this back down to 96
+CheckSize(File, 104, 8);
```
elliottt, line 141: A fresh challenge for you, @froydnj :D

froydnj: There should be two spare bytes right after `flags` which will accommodate this. :D

froydnj (approved):
Is there some documentation being added in a follow up package about what the semantics are?

elliottt:
> Is there some documentation being added in a follow up package about what the semantics are?

Yes, with this PR the sigil is unused.

PR #5720, opened 2022-04-29
Stop sending notification about `# typed: false` file

jez:
### Motivation

We're going to go with #5716 only to show the user what's happening when they
attempt to jump-to-def in a `typed: false` file.

It turns out that VS Code will send `textDocument/definition` requests all over
the place as long as the user is holding down Cmd (not only once the click has
come through).

It sends the `textDocument/definition` request to see whether it can change a
piece of the source document to look like a blue link.

If it had been the case that VS Code just sent a request once the user had
committed to the request, it would have been easier to tell them why they were
getting no results.

Hopefully the hover message from #5716 will be enough to tell people what's
happening.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/definition.cc
+++ b/main/lsp/requests/definition.cc
@@ -1,6 +1,5 @@
 #include "main/lsp/requests/definition.h"
 #include "core/lsp/QueryResponse.h"
-#include "main/lsp/LSPOutput.h"
 #include "main/lsp/LSPQuery.h"
 #include "main/lsp/json_types.h"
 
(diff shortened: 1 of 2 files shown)


froydnj (commented):
Could you please also revert the `position_assertions.cc` changes that check for notifications in definitions and usages now that we no longer send these?

jez:
> Could you please also revert the `position_assertions.cc` changes that check for notifications in definitions and usages now that we no longer send these?

I was thinking that I would just let the test run to show the failures and then fix the failures, rather than also revert the changes to the test.

The test is fixed now.

elliottt (approved):
👍

PR #5722, opened 2022-04-29
Run the packager after the rbi-gen tests

elliottt:
We currently run the packager pass before runnin the rbi-gen tests in
`pipeline_test_runner.cc`. This is a little bit problematic since #5673 where we
switched to storing the mangled package name on the `File` entries: now the
`rbi-gen` pass will also update these fields on the `File`, which will overwrite
the entries from the point of view of the other global state.

The better path forward here is using a side table in the `PackageDB` for
correlating `FileRef` values and the package that owns them.

### Motivation

Test stability.

### Test plan

See included automated tests.


elliottt:
Closing in favor of #5724.

PR #5723, opened 2022-04-29
Add docs for generics in Sorbet

jez:
### Test plan

See included automated tests.


website/docs/generics.md
```diff
@@ -0,0 +1,1019 @@
...
+
+Unfortunately, the abundance of bugs in Sorbet's generics implementation makes
+implementing sophisticated, generic abstractions in Sorbet exceedingly brittle.
+
+It is imperative to thoroughly test abstractions making use of Sorbet generics.
+
+The tests you'll need to write look substantially different from other Ruby
+tests you may be accustomed to writing, because these tests will have to live
+**at the type level** (statics), not at the expression level (runtime). These
+tests should account for two main failure modes:
```
marianosimone, line 29: I feel like *should* is a bit overused here and the following sentences. Maybe you could it replace it with _need_ here?

website/docs/generics.md
```diff
@@ -0,0 +1,1019 @@
...
+
+  (There is no method built into Sorbet for writing such tests; the best way to
+  do it is manually while making changes to the generic abstraction. The
+  diligent may way to automate this by running Sorbet a second time on a
+  codebase that includes extra files meant to not type check, and assert that
+  Sorbet indeed reports errors.)
+
+- Many things **don't type check when they should**.
+
+  This is bad because it will cause undue toil to people attempting to use the
```
marianosimone, line 54: specially for readers that have English as a second language, "undue toil"  is more complicated than needed.

Is the result really "undue toil" (what are they going to do?) or "confusion"?

website/docs/generics.md
```diff
@@ -0,0 +1,1019 @@
...
+
+Sorbet has syntax for creating generic methods, classes, and interfaces.
+
+## Implementation status
+
+Generics are, by far, the most poorly supported feature within Sorbet. The
+implementation has many known, hard-to-fix bugs, which substantially limits
+their utility.
+
+It is hard to overstate how buggy Sorbet's support for generics is.
```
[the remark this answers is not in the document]
jez: I think it is absolutely imperative to convey how buggy Sorbet's support for generics is, so I feel the need to repeat myself here, especially early on in the document, in a way that is particularly hard to overlook.

website/docs/generics.md
jez: Hm, not sure. I think i'd prefer to leave it as is, but thanks for the suggestion.

website/docs/generics.md
```diff
@@ -0,0 +1,1056 @@
...
+- The scope of a `type_member` is all instance methods on the given class. They
+  are most commonly used for generic container classes, because each instance of
+  the class may have a separate type substituted for the type variable.
+
+- The scope of a `type_template` is all singleton class methods on the given
+  class. Since a class only has one singleton class, `type_template` variables
+  are usually used as a way for an abstract parent class to require a concrete
+  child class to pick a specific type that all instances agree on.
+
+> These docs frequently use "type member" (two words, no code formatting) to
```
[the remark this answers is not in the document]
jez: `T.type_paramter(:U)` is also a type variable, but most documentation that applies to `type_member` and `type_template` does not apply to `T.type_parameter(:U)` variables.

website/docs/generics.md
```diff
@@ -0,0 +1,1056 @@
...
+
+  # Implement the `IBox` interface
+  include IBox
+
+  # Redeclare the type member, to be compatible with `IBox`
+  Elem = type_member
+
+  # Within this class, `Elem` is invariant, so it can also be used
+  sig {params(value: Elem).void}
+  def initialize(value:); @value = value; end
```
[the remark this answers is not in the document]
jez: So that the example fits on smaller screens without excessive scrolling, and so that method bodies are de-emphasized relative to method signatures, which are the important part of this example.

website/docs/generics.md
```diff
@@ -0,0 +1,1056 @@
...
+Covariant type members preserve the subtyping relationship. Specifically, if the
+type `Child` is a subtype of the type `Parent`, then the type `M[Child]` is a
+subtype of the type `M[Parent]` if the type member of `M` is covariant. In
+symbols:
+
+```
+Child <: Parent  ==>  M[Child] <: M[Parent]
+```
+
+Note that only a Ruby `module` (not a `class`) may have a covariant type member.
```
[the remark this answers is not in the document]
jez: I took a pass through the places where it might be ambiguous and rewrote things.

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+end
+
+int_box = Box[Integer].new(val: 0)
+T.reveal_type(int_box) # `Box[Integer]`
+
+T.reveal_type(int_box.val) # `Integer`
+
+int_box.val += 1
+
+string_box = int_box.map {|val| val.to_s}
```
froydnj, line 106: Running the given example in `sorbet.run`:

```
editor.rb:25: Method map does not exist on Box[Integer] https://srb.help/7003
    25 |string_box = int_box.map {|val| val.to_s}
                             ^^^
  Got Box[Integer] originating from:
    editor.rb:18:
    18 |int_box = Box[Integer].new(val: 0)
                  ^^^^^^^^^^^^^^^^^^^^^^^^
```

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+`type_member`. (See the docs for error code [5016](error-reference.md#5016) for
+more information.) Note that since `type_template` creates a type variable
+scoped to a singleton class, `type_template` can never be covariant (because all
+singleton classes are classes, even singleton classes of modules).
+
+Here's an example of a module that has a covariant type member:
+
+```ruby
+# covariant `Box` interface
+module IBox
```
froydnj, line 352:
```suggestion
extend T::Sig

# covariant `Box` interface
module IBox
```

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+  extend T::Generic
+  # `:out` declares this type member as covariant
+  Elem = type_member(:out)
+end
+
+sig {params(int_box: IBox[Integer]).void}
+def example(int_box)
+  T.let(int_ibox, IBox[Numeric]) # OK
```
froydnj, line 360:
```suggestion
  T.let(int_box, IBox[Numeric]) # OK
```

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+```
+Child <: Parent  ==>  M[Parent] <: M[Child]
+```
+
+Contravariance is **quite** unintuitive for most people. Luckily, contravariance
+is not unique to Sorbet—all type systems that have both subtyping relationships
+and generics must grapple with variance, contravariance included, so there is a
+lot written about it elsewhere online. It maybe even be helpful to read about
+contravariance in a language you already have extensive familiarity with, as
+many of the concepts with transfer to Sorbet.
```
froydnj, line 438:
```suggestion
many of the concepts will transfer to Sorbet.
```

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+    .returns([T::Array[T.type_parameter(:K)], T::Array[T.type_parameter(:V)]])
+end
+def keys_and_values(hash)
+  [hash.keys, hash.values]
+end
+```
+
+[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20do%0A%20%20type_parameters%28%3AK%2C%20%3AV%29%0A%20%20%20%20.params%28hash%3A%20T%3A%3AHash%5BT.type_parameter%28%3AK%29%2C%20T.type_parameter%28%3AV%29%5D%29%0A%20%20%20%20.returns%28%5BT%3A%3AArray%5BT.type_parameter%28%3AK%29%5D%2C%20T%3A%3AArray%5BT.type_parameter%28%3AV%29%5D%5D%29%0Aend%0Adef%20keys_and_values%28hash%29%0A%20%20%5Bhash.keys%2C%20hash.values%5D%0Aend)
+
+Note Sorbet does not support return type inference. This means that doing
```
froydnj, line 892: "does not support return type inference" initially made me think that Sorbet did not support ever having generic return values (`.returns(T.type_parameter(:U))`), which is obviously false given examples just a little bit further down.  I think "return type inference" is the usual way to talk about it, but maybe "does not support inferring return types from method bodies" is a better way to say things?

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+  constrained by of the arguments. It could therefore be anything.
+- The only type that is a subtype of any type in Sorbet is
+  [`T.noreturn`](noreturn.md). The only way to introduce a value of this type is
+  to raise an exception.
+- Therefore, Sorbet infers that the only valid way to implement
+  `returns_something` is by raising, which would imply that the `puts(x)` code
+  is never reached.
+
+Sorbet cannot have return type inference, because languages that implement
+return type inference take advantage of a compilation step to decide how to call
```
froydnj, line 923: Same comment here about "return type inference"

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+their utility.
+
+It is hard to overstate how buggy Sorbet's support for generics is.
+
+We track known bugs in Sorbet's support for generics in the [Generics milestone]
+on Sorbet's issue tracker. Feel free to peruse the existing bugs, as well as
+report new bugs.
+
+Unfortunately, this abundance of bugs makes implementing sophisticated, generic
+abstractions in Sorbet exceedingly brittle.
```
froydnj, line 22: I understand the desire to emphasize the shortcomings of Sorbet's generics support.  (I think internal discussions had suggested separating "generic classes" from "generic methods" as the former are relatively more stable, but maybe you made the -- reasonable! -- executive decision to combine them, as it's hard to write generic classes without also extensively using generic methods.)

But the practical reality is that people use Sorbet's generics support every single day via the standard library, and Stripe's codebase has thousands of instances of `type_member`/`type_template` even prior to docs on generics being written.  People do run into issues with the generics support (cf. the generics milestone, etc.), yes, and there's no guarantee that all of the instances of `type_member` et al have been rigorously subjected to the kind of tests the documentation below suggests.  And yet here we are, writing the longest piece of documentation (by lines) that exists on Sorbet's website.

I don't know what the right way to modify this introduction is, or even if it should be modified.  But I do think that acknowledging the usefulness of generics while being forceful about their shortcomings would leave fewer people scratching their heads about why this documentation exists.

(I also think "It is hard to overstate how buggy Sorbet's support for generics is." is unnecessary.)

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+
+It is imperative to thoroughly test abstractions making use of Sorbet generics.
+
+The tests you'll need to write look substantially different from other Ruby
+tests you may be accustomed to writing, because they will have to live **at the
+type level** (statics), not at the expression level (runtime). These tests
```
froydnj, line 28: I think "type level" vs. "expression level" is a bit too much for this documentation.  I don't know if the below is a better way to phrase it, but it at least points in the right direction.

```suggestion
The tests you'll need to write look substantially different from other Ruby
tests you may be accustomed to writing, because the tests need to deal with
what code should or should not typecheck, rather that code that should or
should not run correctly.   These tests
```

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+  # (allowed because `Elem` is invariant within this class)
+  sig {params(value: Elem).returns(Elem)}
+  def value=(value); @value = value; end
+end
+```
+
+[→ View on sorbet.run][covariant-ibox]
+
+Note how in the above example, `Box` includes `IBox`, meaning that `Box` is a
+child of `IBox`. Children of a generic classes or modules must always redeclare
```
froydnj, line 416:
```suggestion
child of `IBox`. Children of generic classes or modules must always redeclare
```

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+
+### Input and output positions
+
+Understanding where covariant and contravariant type members can appear requires
+knowing which places in a method signature are **output** positions, and which
+are **input** positions.
+
+An obvious output position is a method signature's `returns` annotation, but
+there are more than just that. As an intuition, all positions in a signature
+where the value is produced by some computation in the method's body are out
```
froydnj, line 563:
```suggestion
where the value is produced by some computation in the method's body are output
```

froydnj (approved):
Thanks for tackling this!  I think it will be really nice to have a place to point people when they have questions.

website/docs/generics.md
```diff
@@ -0,0 +1,1102 @@
...
+```
+
+In the first example, a simple function from type `A` to type `B`, `B` is the
+result of the function, so it's clearly in an output position. Similarly, `A` is
+in an input position.
+
+The second example is the type of a function that takes a function as a
+parameter, having type `C -> D`, and produces another function as its output,
+having type `E -> F`. In this example, `C` is in the input position of an input
+position, making type `C` actually be in output position overall
```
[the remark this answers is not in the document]
jez: Is this a request to change the wording, or just an observation?

I tried to find a canonical reference I could point to for this, but ran out of time.

website/docs/generics.md
jez: whoops, i made this edit in response to some other comments, forgot to update the example. thanks!

website/docs/generics.md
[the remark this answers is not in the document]
froydnj: It's partly an observation that in a section talking about intuition, you get the decidedly non-intuitive "the input of an input is an output" result.  I don't know what to replace it with; an acknowledgement + explanatory reference might help?

website/docs/generics.md
jez: Sorry the term is usually "return type polymorphism" not "return type inference"

e.g. https://eli.thegreenplace.net/2018/return-type-polymorphism-in-haskell/

froydnj: "Return type deduction" is what the C++ spec calls it: http://eel.is/c++draft/dcl.spec.auto.general

website/docs/generics.md
jez: The docs are long not because it is so widely used, but because it requires so much knowledge to use correctly.

The prevalence of usage in Stripe's codebase is not very diverse. Almost all of them blame to one or two narrowly defined abstractions (DataViews). Some caveats about that:

- They actually do suffer from nearly all the problems mentioned in this doc. Nearly 100% of the implementation of DataView is untyped to work around the bugs.
- Child classes almost never touch those annotations—the annotations are kept in sync by automated scripts.
- The core abstraction was written very carefully in coordination with members of the Sorbet team, and has not been changed much since. Nearly no one else using generics will have as intimate an understanding of the limitations of Sorbet's generics when building new abstractions.

I have removed the "hard to overstate" line, but I 100% still believe it is accurate.

The purpose of this part of the doc is intentionally to scare people into testing that their generic abstractions _actually work_, rather than inflicting pain on unsuspecting consumers.

If you have further suggestions, please lmk.

jez: Some reasons why I wrote one page:

- it's easier to ⌘F a single page, and generic methods vs generic classes is not something i expect people to have a solid understanding of
- i only wanted to write the disclaimer once, not across two pages
- you're right that generic classes do frequently also want generic methods (though as this doc does demonstrate, it is possible to have examples that use one or the other entirely, not both)

website/docs/generics.md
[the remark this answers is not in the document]
jez: I spent another few minutes trying to look for an easy reference, and they were all either excessively theoretical or talking about the wrong things.

Unfortunately I think this is the best I can do. It's worth noting that C# has the exact same restrictions on variance, but does not seem to explain this concept at all in their docs.

PR #5724, opened 2022-04-29
Store the package name mapping in the PackageDB instead of the file table

elliottt:
Storing a `NameRef` in the file table is problematic, as multiple global states can share file tables, but have differing name tables. This PR moves the mapping from file to package into the PackageDB instead, avoiding the situation where we might have incompatible `NameRef` values stored in the file table.

### Motivation

Stable package to file mapping.

### Test plan

See included automated tests.

### DIFF
--- a/ast/substitute/Substitute.cc
+++ b/ast/substitute/Substitute.cc
@@ -131,24 +131,12 @@ public:
 ParsedFile Substitute::run(core::MutableContext ctx, const core::NameSubstitution &subst, ParsedFile what) {
     SubstWalk walk(subst);
     what.tree = TreeMap::apply(ctx, walk, std::move(what.tree));
-
-    auto pkg = what.file.data(ctx).getPackage();
-    if (pkg.exists()) {
-        what.file.data(ctx).setPackage(subst.substitute(pkg));
-    }
-
     return what;
 }
 
 ParsedFile Substitute::run(core::MutableContext ctx, core::LazyNameSubstitution &subst, ParsedFile what) {
     SubstWalk walk(subst);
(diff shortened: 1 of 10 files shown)


froydnj:
I think this is fine as far as it goes, but doesn't this sort of have the same problem?  Or do we always have a 1:1 mapping of GlobalState to PackageDB, and therefore the NameRef is unambiguous?

jez (approved):

elliottt:
> I think this is fine as far as it goes, but doesn't this sort of have the same problem? Or do we always have a 1:1 mapping of GlobalState to PackageDB, and therefore the NameRef is unambiguous?

This is a little different in that the `PackageDB` values aren't shared between `GlobalState` instances, while the `File` values are explicitly shared with a `shared_ptr`. This way the mapping is unique for each `GlobalState`, rather than accidentally leaking `NameRef` values between them.

elliottt:

→ https://go/builds/bui_LbCnb4hzeVmsUC
→ https://go/builds/bui_LbCnLINOW7yl08

elliottt:
> > I think this is fine as far as it goes, but doesn't this sort of have the same problem? Or do we always have a 1:1 mapping of GlobalState to PackageDB, and therefore the NameRef is unambiguous?
> 
> This is a little different in that the `PackageDB` values aren't shared between `GlobalState` instances, while the `File` values are explicitly shared with a `shared_ptr`. This way the mapping is unique for each `GlobalState`, rather than accidentally leaking `NameRef` values between them.

I think my answer ended up more complicated than it needed to be: you're right about the 1:1 mapping of `GlobalState` to `PackageDB`, and that's what this PR relies on.

PR #5727, opened 2022-04-30
Package visibilty checker

elliottt:
Rework how package visibility is enforced, by running a pass over trees after the resolver that enforces correct usage of package imports/exports. This improves memory usage for the packager pass, and allows for more precise errors in some cases.

### Motivation

Improving performance of the packager pass on Stripe's codebase.

### Test plan

See included automated tests.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -268,6 +268,14 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
                 return what;
             }
         }
+
+#ifndef SORBET_REALMAIN_MIN
+        if (opts.stripePackages) {
+            auto emptyWorkers = WorkerPool::create(0, gs.tracer());
+            what = packager::VisibilityChecker::runIncremental(gs, *emptyWorkers, std::move(what));
+        }
+#endif
+
     } catch (SorbetException &) {
         if (auto e = gs.beginError(sorbet::core::Loc::none(), sorbet::core::errors::Internal::InternalError)) {
             e.setHeader("Exception resolving (backtrace is above)");
@@ -316,6 +324,13 @@ vector<ast::ParsedFile> incrementalResolveBestEffort(const core::GlobalState &gs
                 return what;
             }
         }
+
+#ifndef SORBET_REALMAIN_MIN
+        if (opts.stripePackages) {
+            auto emptyWorkers = WorkerPool::create(0, gs.tracer());
+            what = packager::VisibilityChecker::runIncremental(gs, *emptyWorkers, move(what));
+        }
+#endif
     } catch (SorbetException &) {
         if (auto e = gs.beginError(sorbet::core::Loc::none(), sorbet::core::errors::Internal::InternalError)) {
             e.setHeader("Exception resolving (backtrace is above)");
@@ -869,6 +884,14 @@ ast::ParsedFilesOrCancelled resolve(unique_ptr<core::GlobalState> &gs, vector<as
                 }
                 what = move(maybeResult.result());
             }
+
+#ifndef SORBET_REALMAIN_MIN
+            if (opts.stripePackages) {
+                Timer timeit(gs->tracer(), "visibility_checker");
+                what = packager::VisibilityChecker::run(*gs, workers, std::move(what));
+            }
+#endif
+
             if (opts.stressIncrementalResolver) {
                 auto symbolsBefore = gs->symbolsUsedTotal();
                 for (auto &f : what) {
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -341,8 +341,10 @@ public:
             }
         }
 
+        auto &db = ctx.state.packageDB();
+
         // no need to check visibility for these cases
-        auto otherPackage = ctx.state.packageDB().getPackageNameForFile(otherFile);
+        auto otherPackage = db.getPackageNameForFile(otherFile);
         if (!otherPackage.exists() || this->package.mangledName() == otherPackage) {
             return tree;
         }
@@ -358,6 +360,10 @@ public:
                     e.addAutocorrect(std::move(exp.value()));
                 }
                 e.addErrorLine(pkg.declLoc(), "Defined here");
+
+                if (!db.errorHint().empty()) {
+                    e.addErrorNote("{}", db.errorHint());
+                }
             }
 
             return tree;
@@ -389,6 +395,10 @@ public:
                     e.addAutocorrect(std::move(exp.value()));
                 }
                 e.addErrorLine(lit.symbol.loc(ctx), "Defined here");
+
+                if (!db.errorHint().empty()) {
+                    e.addErrorNote("{}", db.errorHint());
+                }
             }
         }
 
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -84,6 +84,17 @@ class PropagateVisibility final {
         }
     }
 
+    // While processing the ClassDef for the package, which will be named something like `<PackageSpecRegistry>::A::B`,
+    // we also check that the symbols `A::B` and `Test::A::B` have locations whose package matches the one we're
+    // processing. If they don't match, we add locs to ensure that those symbols are associated with this package.
+    //
+    // The reason for this step is that it's currently allowed to refer to the name of the package outside of the
+    // context of the package spec, even if it doesn't explicitly export its top-level name. So in the case above, there
+    // would be no `export A::B` line in the package spec. However, if there is a package that is defined in the child
+    // namespace of `A::B`, and it at some point has a declaration of the form `module A::B`, the file that contains
+    // that declaration will be marked as owning `A::B`, thus breaking the invariant that the file can be used to
+    // determine the package that owns a symbol. So, to avoid this case we ensure that the symbols that correspond to
+    // the package name are always owned by the package that defines them.
     void setPackageLocs(core::MutableContext ctx, core::LocOffsets loc, core::ClassOrModuleRef sym) {
         std::vector<core::NameRef> names;
 
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -326,6 +326,7 @@ public:
                 if (auto exp = this->package.addImport(ctx, pkg, isTestImport)) {
                     e.addAutocorrect(std::move(exp.value()));
                 }
+                e.addErrorLine(pkg.declLoc(), "Defined here");
             }
 
             return tree;
@@ -337,6 +338,7 @@ public:
                 if (auto exp = this->package.addImport(ctx, pkg, false)) {
                     e.addAutocorrect(std::move(exp.value()));
                 }
+                e.addErrorLine(pkg.declLoc(), "Defined here");
             }
         }
 
@@ -355,6 +357,7 @@ public:
                 if (auto exp = pkg.addExport(ctx, lit.symbol, false)) {
                     e.addAutocorrect(std::move(exp.value()));
                 }
+                e.addErrorLine(lit.symbol.loc(ctx), "Defined here");
             }
         }
 
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -337,7 +337,8 @@ public:
         // If the imported symbol comes from the test namespace, we must also be in the test namespace.
         if (otherFile.data(ctx).isPackagedTest() && !this->insideTestFile) {
             if (auto e = ctx.beginError(lit.loc, core::errors::Packager::UsedTestOnlyName)) {
-                e.setHeader("Used test-only constant `{}` in non-test file", lit.symbol.show(ctx));
+                e.setHeader("`{}` is defined in a test namespace and cannot be referenced in a non-test file",
+                            lit.symbol.show(ctx));
             }
         }
 
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -1,0 +1,555 @@
+#include "packager/VisibilityChecker.h"
+#include "ast/treemap/treemap.h"
+#include "common/concurrency/ConcurrentQueue.h"
+#include "common/formatting.h"
+#include "common/sort.h"
+#include "core/Context.h"
+#include "core/errors/packager.h"
+#include "core/errors/resolver.h"
+#include <iterator>
+
+namespace sorbet::packager {
+
+namespace {
+
+// For each __package.rb file, traverse the resolved tree and apply the visibility annotations to the symbols.
+class PropagateVisibility final {
+    const core::packages::PackageInfo &package;
+
+    bool definedByThisPackage(const core::GlobalState &gs, core::ClassOrModuleRef sym) {
+        auto pkg = gs.packageDB().getPackageNameForFile(sym.data(gs)->loc().file());
+        return this->package.mangledName() == pkg;
+    }
+
+    void recursiveExportSymbol(core::GlobalState &gs, bool firstSymbol, core::ClassOrModuleRef klass) {
+        // There is a loop between a class and its singleton, and this is the easiest way to detect it.
+        if (!firstSymbol && klass.data(gs)->flags.isExported) {
+            return;
+        }
+
+        // We only mark symbols from this package.
+        if (!this->definedByThisPackage(gs, klass)) {
+            return;
+        }
+
+        klass.data(gs)->flags.isExported = true;
+
+        for (const auto &child : klass.data(gs)->members()) {
+            if (child.second.isClassOrModule()) {
+                recursiveExportSymbol(gs, false, child.second.asClassOrModuleRef());
+            } else if (child.second.isFieldOrStaticField()) {
+                child.second.asFieldRef().data(gs)->flags.isExported = true;
+            }
+        }
+    }
+
+    void exportParentNamespace(core::GlobalState &gs, core::ClassOrModuleRef owner) {
+        while (owner.exists() && !owner.data(gs)->flags.isExported && this->definedByThisPackage(gs, owner)) {
+            owner.data(gs)->flags.isExported = true;
+            owner = owner.data(gs)->owner;
+        }
+    }
+
+    // Lookup the package name on the given root symbol, and mark the final symbol as exported.
+    void exportRoot(core::GlobalState &gs, core::ClassOrModuleRef sym) {
+        // Explicitly mark the symbol outside of the special package namespace as exported by this package, to allow
+        // other packages to refer to this namespace. Ultimately it would be great to not have this, but it helps with a
+        // migration to the visibility checker from the previous packages implementation.
+        for (auto name : this->package.fullName()) {
+            auto next = sym.data(gs)->findMember(gs, name);
+            if (!next.exists() || !next.isClassOrModule()) {
+                sym = core::Symbols::noClassOrModule();
+                break;
+            }
+
+            sym = next.asClassOrModuleRef();
+        }
+
+        if (sym.exists()) {
+            sym.data(gs)->flags.isExported = true;
+        }
+    }
+
+    // Mark both the package and its test namespace as exported. This will terminate the recursive exporting in
+    // `exportParentNamespace`, as it stops when it hits either the root or an exported symbol.
+    void exportPackageRoots(core::GlobalState &gs) {
+        this->exportRoot(gs, core::Symbols::root());
+
+        auto test = core::Symbols::root().data(gs)->findMember(gs, core::Names::Constants::Test());
+        if (test.exists() && test.isClassOrModule()) {
+            this->exportRoot(gs, test.asClassOrModuleRef());
+        }
+    }
+
+    void setPackageLocs(core::MutableContext ctx, core::LocOffsets loc, core::ClassOrModuleRef sym) {
+        std::vector<core::NameRef> names;
+
+        while (sym.exists() && sym != core::Symbols::PackageSpecRegistry()) {
+            // The symbol isn't a package name if it's defined outside of the package registry.
+            if (sym == core::Symbols::root()) {
+                return;
+            }
+
+            names.emplace_back(sym.data(ctx)->name);
+            sym = sym.data(ctx)->owner;
+        }
+
+        absl::c_reverse(names);
+
+        auto &db = ctx.state.packageDB();
+
+        {
+            auto packageSym = core::Symbols::root();
+            for (auto name : names) {
+                auto member = packageSym.data(ctx)->findMember(ctx, name);
+                if (!member.exists() || !member.isClassOrModule()) {
+                    packageSym = core::Symbols::noClassOrModule();
+                    break;
+                }
+
+                packageSym = member.asClassOrModuleRef();
+            }
+
+            if (packageSym.exists()) {
+                auto file = packageSym.data(ctx)->loc().file();
+                if (db.getPackageNameForFile(file) != this->package.mangledName()) {
+                    packageSym.data(ctx)->addLoc(ctx, ctx.locAt(loc));
+                }
+            }
+        }
+
+        {
+            auto testSym = core::Symbols::root();
+            auto member = testSym.data(ctx)->findMember(ctx, core::Names::Constants::Test());
+            if (!member.exists() || !member.isClassOrModule()) {
+                return;
+            }
+
+            testSym = member.asClassOrModuleRef();
+            for (auto name : names) {
+                auto member = testSym.data(ctx)->findMember(ctx, name);
+                if (!member.exists() || !member.isClassOrModule()) {
+                    testSym = core::Symbols::noClassOrModule();
+                    break;
+                }
+
+                testSym = member.asClassOrModuleRef();
+            }
+
+            if (testSym.exists()) {
+                auto file = testSym.data(ctx)->loc().file();
+                if (db.getPackageNameForFile(file) != this->package.mangledName()) {
+                    testSym.data(ctx)->addLoc(ctx, ctx.locAt(loc));
+                }
+            }
+        }
+    }
+
+    void checkExportPackage(core::MutableContext &ctx, core::LocOffsets loc, core::NameRef litPackage) {
+        if (litPackage != this->package.mangledName() && !this->package.importsPackage(litPackage)) {
+            if (auto e = ctx.beginError(loc, core::errors::Packager::InvalidExport)) {
+                e.setHeader("Exporting a symbol that isn't owned by this package");
+            }
+        }
+    }
+
+    PropagateVisibility(const core::packages::PackageInfo &package) : package{package} {}
+
+public:
+    // Find uses of export and mark the symbols they mention as exported.
+    ast::ExpressionPtr postTransformSend(core::MutableContext ctx, ast::ExpressionPtr tree) {
+        auto &send = ast::cast_tree_nonnull<ast::Send>(tree);
+        if (send.fun != core::Names::export_()) {
+            return tree;
+        }
+
+        if (send.numPosArgs() != 1) {
+            // an error will have been raised in the packager pass
+            return tree;
+        }
+
+        auto *lit = ast::cast_tree<ast::ConstantLit>(send.getPosArg(0));
+        if (lit == nullptr || lit->symbol == core::Symbols::StubModule()) {
+            // We don't raise an explicit error here, as this is one of two cases:
+            //   1. Export is given a non-constant argument
+            //   2. The argument failed to resolve
+            // In both cases, errors will be raised by previous passes.
+            return tree;
+        }
+
+        auto &db = ctx.state.packageDB();
+        if (lit->symbol.isClassOrModule()) {
+            auto sym = lit->symbol.asClassOrModuleRef();
+            checkExportPackage(ctx, send.loc, db.getPackageNameForFile(sym.data(ctx)->loc().file()));
+            recursiveExportSymbol(ctx, true, sym);
+
+            exportParentNamespace(ctx, sym.data(ctx)->owner);
+        } else if (lit->symbol.isFieldOrStaticField()) {
+            auto sym = lit->symbol.asFieldRef();
+            checkExportPackage(ctx, send.loc, db.getPackageNameForFile(sym.data(ctx)->loc().file()));
+            sym.data(ctx)->flags.isExported = true;
+
+            // When exporting a field, we also export its parent namespace. This is a bit of a hack, and it would be
+            // great to remove this, but this was the behavior of the previous packager implementation.
+            // TODO: can we get away with only marking the owner as exported, and not everything up to the package?
+            exportParentNamespace(ctx, sym.data(ctx)->owner);
+
+            if (sym.data(ctx)->flags.isStaticFieldTypeAlias) {
+                if (auto e = ctx.beginError(send.loc, core::errors::Packager::ExportingTypeAlias)) {
+                    e.setHeader("Exporting a type alias is not allowed in package specifications");
+                    e.addErrorLine(sym.data(ctx)->loc(), "Originally defined here");
+                }
+            }
+        }
+
+        return tree;
+    }
+
+    ast::ExpressionPtr preTransformClassDef(core::MutableContext ctx, ast::ExpressionPtr tree) {
+        auto &original = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+
+        if (original.symbol == core::Symbols::root()) {
+            return tree;
+        }
+
+        setPackageLocs(ctx, original.name.loc(), original.symbol);
+
+        return tree;
+    }
+
+    static ast::ParsedFile run(core::GlobalState &gs, ast::ParsedFile f) {
+        if (!f.file.data(gs).isPackage()) {
+            return f;
+        }
+
+        auto pkgName = gs.packageDB().getPackageNameForFile(f.file);
+        if (!pkgName.exists()) {
+            return f;
+        }
+
+        const auto &package = gs.packageDB().getPackageInfo(pkgName);
+        if (!package.exists()) {
+            return f;
+        }
+
+        PropagateVisibility pass{package};
+
+        pass.exportPackageRoots(gs);
+
+        core::MutableContext ctx{gs, core::Symbols::root(), f.file};
+        f.tree = ast::TreeMap::apply(ctx, pass, std::move(f.tree));
+
+        return f;
+    }
+};
+
+class VisibilityCheckerPass final {
+    bool nameMatchesPackage(const std::vector<core::NameRef> &revNameParts) {
+        const auto &packageName = this->package.fullName();
+
+        if (revNameParts.size() < packageName.size()) {
+            return false;
+        }
+
+        return std::equal(packageName.begin(), packageName.end(), revNameParts.rbegin());
+    }
+
+public:
+    const core::packages::PackageInfo &package;
+    const bool insideTestFile;
+
+    VisibilityCheckerPass(core::Context ctx, const core::packages::PackageInfo &package)
+        : package{package}, insideTestFile{ctx.file.data(ctx).isPackagedTest()} {}
+
+    // `keep-def` will reference constants in a way that looks like a packaging violation, but is actually fine. This
+    // boolean allows for an early exit when we know we're in the context of processing one of these sends. Currently
+    // the only sends that we process this way will not have any nested method calls, but if that changes this will need
+    // to become a stack.
+    bool ignoreConstant = false;
+
+    ast::ExpressionPtr preTransformSend(core::Context ctx, ast::ExpressionPtr tree) {
+        auto &send = ast::cast_tree_nonnull<ast::Send>(tree);
+        this->ignoreConstant = send.fun == core::Names::keepForIde();
+        return tree;
+    }
+
+    ast::ExpressionPtr postTransformSend(core::Context ctx, ast::ExpressionPtr tree) {
+        this->ignoreConstant = false;
+        return tree;
+    }
+
+    ast::ExpressionPtr postTransformConstantLit(core::Context ctx, ast::ExpressionPtr tree) {
+        if (this->ignoreConstant) {
+            return tree;
+        }
+
+        auto &lit = ast::cast_tree_nonnull<ast::ConstantLit>(tree);
+        if (!lit.symbol.isClassOrModule() && !lit.symbol.isFieldOrStaticField()) {
+            return tree;
+        }
+
+        auto loc = lit.symbol.loc(ctx);
+
+        auto otherFile = loc.file();
+        if (!otherFile.exists() || !otherFile.data(ctx).isPackaged()) {
+            return tree;
+        }
+
+        // If the imported symbol comes from the test namespace, we must also be in the test namespace.
+        if (otherFile.data(ctx).isPackagedTest() && !this->insideTestFile) {
+            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::UsedTestOnlyName)) {
+                e.setHeader("Used test-only constant `{}` in non-test file", lit.symbol.show(ctx));
+            }
+        }
+
+        // no need to check visibility for these cases
+        auto otherPackage = ctx.state.packageDB().getPackageNameForFile(otherFile);
+        if (!otherPackage.exists() || this->package.mangledName() == otherPackage) {
+            return tree;
+        }
+
+        // Did we resolve a packaged symbol in unpackaged code?
+        if (!this->package.exists()) {
+            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::PackagedSymbolInUnpackagedContext)) {
+                e.setHeader("Packaged constant `{}` used in an unpackaged context", lit.symbol.show(ctx));
+            }
+            return tree;
+        }
+
+        auto importType = this->package.importsPackage(otherPackage);
+        if (!importType.has_value()) {
+            // We failed to import the package that defines the symbol
+            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::MissingImport)) {
+                auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
+                e.setHeader("No import provides `{}`", lit.symbol.show(ctx));
+                bool isTestImport = otherFile.data(ctx).isPackagedTest();
+                if (auto exp = this->package.addImport(ctx, pkg, isTestImport)) {
+                    e.addAutocorrect(std::move(exp.value()));
+                }
+            }
+
+            return tree;
+        } else if (*importType == core::packages::ImportType::Test && !this->insideTestFile) {
+            // We used a symbol from a `test_import` in a non-test context
+            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::UsedTestOnlyName)) {
+                e.setHeader("Used `{}` constant `{}` in non-test file", "test_import", lit.symbol.show(ctx));
+                auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
+                if (auto exp = this->package.addImport(ctx, pkg, false)) {
+                    e.addAutocorrect(std::move(exp.value()));
+                }
+            }
+        }
+
+        bool isExported = false;
+        if (lit.symbol.isClassOrModule()) {
+            isExported = lit.symbol.asClassOrModuleRef().data(ctx)->flags.isExported;
+        } else if (lit.symbol.isFieldOrStaticField()) {
+            isExported = lit.symbol.asFieldRef().data(ctx)->flags.isExported;
+        }
+
+        // Did we use a constant that wasn't exported?
+        if (!isExported) {
+            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::UsedPackagePrivateName)) {
+                auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
+                e.setHeader("Package `{}` does not export `{}`", pkg.show(ctx), lit.symbol.show(ctx));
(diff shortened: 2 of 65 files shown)


main/pipeline/pipeline.cc
```diff
@@ -268,6 +268,14 @@ vector<ast::ParsedFile> incrementalResolve(core::GlobalState &gs, vector<ast::Pa
                 return what;
             }
         }
+
+#ifndef SORBET_REALMAIN_MIN
+        if (opts.stripePackages) {
+            auto emptyWorkers = WorkerPool::create(0, gs.tracer());
+            what = packager::VisibilityChecker::runIncremental(gs, *emptyWorkers, std::move(what));
+        }
+#endif
```
jez, line 277: Should we get some timers for these?

packager/VisibilityChecker.h
```diff
@@ -0,0 +1,22 @@
+#ifndef PACKAGER_VISIBILITY_CHECKER_H
+#define PACKAGER_VISIBILITY_CHECKER_H
+
+#include "ast/ast.h"
+#include "common/concurrency/WorkerPool.h"
+
+namespace sorbet::packager {
+
+class VisibilityChecker final {
+    ~VisibilityChecker() = default;
```
jez, line 10: ooc, what does this do?

test/cli/package-error-missing-export/test.out
```diff
@@ -1,52 +1,32 @@
...
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   Autocorrect: Use `-a` to autocorrect
     other/__package.rb:4: Insert `export Foo::Bar::OtherPackage::NotExported`
      4 |  export Foo::Bar::OtherPackage::OtherClass
                                                    ^
-    other/other_class.rb:7: Defined here
-     7 |  class NotExported
-          ^^^^^^^^^^^^^^^^^
-  Note:
-    Try running generate-packages.sh
```
jez, line 25: Why did we drop the `generate-packages.sh` note?

website/docs/error-reference.md
```diff
@@ -702,6 +702,68 @@ you'll get this error—the latter export is redundant, as it's implied by expor
 
 To fix this error, simply remove the more specific export.
 
+## 3717
+
+> This error is specific to Stripe's custom `--stripe-packages` definition. If
+> you are at Stripe, please see [go/modularity](http://go/modularity) for more.
```
jez, line 708: lol, i have no idea what i meant by "`--stripe-packages` definition"

I'm going to change this to "`--stripe-packages` mode" after you land this change.

website/docs/error-reference.md
```diff
@@ -702,6 +702,68 @@ you'll get this error—the latter export is redundant, as it's implied by expor
...
+
+This error means that you're using a name that's private in another package, and
+not epxorted.
```
jez, line 711: Is it private **and** not exported?

Stripe's modularity guidelines also have rules about `Private` classes, and I think that the `Private` namespace concept is separate.

Also a question:

Does this error only happen when the package containing the definition **is** imported in this package, or could it ever be the case that the user would have to **both** import the package in this package **and** export the relevant definition in the upstream package?

Could we capture the answer in the docs here?

website/docs/error-reference.md
```diff
@@ -702,6 +702,68 @@ you'll get this error—the latter export is redundant, as it's implied by expor
...
+
+To fix this error, you'll need to either export the symbol in the other package,
+or figure out a way around using it.
+
+## 3718
+
+> This error is specific to Stripe's custom `--stripe-packages` definition. If
+> you are at Stripe, please see [go/modularity](http://go/modularity) for more.
+
+This error means that you're using a symbol from a packge that's not imported.
```
jez, line 721: I've tried to avoid using the word "symbol" to refer to definitions in error messages. I prefer using "definition" or "constant" instead.

test/cli/package-export-nested-namespace-conflict/parent/b.rb
```diff
@@ -1,5 +0,0 @@
-# typed: strict
-
-module Parent
```
jez, line 3: ooc, why delete this test, vs capture the change to the output?

test/cli/package-import-parent-package-conflict/parent/__package.rb
```diff
@@ -1,5 +0,0 @@
-# typed: strict
```
jez, line 1: (ditto the question about about why delete this test vs capture the change to the cli output)

test/cli/package-redefine-nested-namespace-error/test.out
```diff
@@ -3,15 +3,4 @@ parent/b.rb:4: Class or method definition must match enclosing package namespace
...
-parent/b.rb:4: Redefining constant `Parent::Child` https://srb.help/4012
-     4 |  module Child
-     5 |    class B; end
-     6 |  end
-    parent/__package.rb: Previous definition
-
-parent/b.rb:5: Can't nest `B` under `Parent::Child` because `Parent::Child` is not a class or module https://srb.help/4015
-     5 |    class B; end
-                  ^
-    parent/__package.rb: `Parent::Child` defined here
```
jez, line 16: I think that these errors disappearing makes sense, but just to double check: would you mind briefly explaining why we expect these errors to go away?

test/cli/package-test-simple/test.out
```diff
@@ -1,13 +1,8 @@
-main_lib/lib.rb:8: No import provides `Project::TestOnly` https://srb.help/5002
+main_lib/lib.rb:8: Used `test_import` constant `Project::TestOnly::SomeHelper` in non-test file https://srb.help/3720
```
jez, line 1: Do you mind starting a list of follow-up items that would be good to get to eventually?

(Was `test_import` one of the things we were going to delete? Or was it `export_for_test`? Any others?)

Happy for the list to live either

- as a bullet list somewhere on this PR
- as github issues
- as Stripe-internal Jira tickets
- as a Stripe-internal document

(at your preference)

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,555 @@
...
+        // no need to check visibility for these cases
+        auto otherPackage = ctx.state.packageDB().getPackageNameForFile(otherFile);
+        if (!otherPackage.exists() || this->package.mangledName() == otherPackage) {
+            return tree;
+        }
+
+        // Did we resolve a packaged symbol in unpackaged code?
+        if (!this->package.exists()) {
+            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::PackagedSymbolInUnpackagedContext)) {
+                e.setHeader("Packaged constant `{}` used in an unpackaged context", lit.symbol.show(ctx));
```
jez, line 314: Do we have any tests that mention `# packaged: false` other than the one unit test in `core`?

website/docs/error-reference.md
```diff
@@ -702,6 +702,68 @@ you'll get this error—the latter export is redundant, as it's implied by expor
...
+## 3719
+
+> This error is specific to Stripe's custom `--stripe-packages` definition. If
+> you are at Stripe, please see [go/modularity](http://go/modularity) for more.
+
+This error means that you're using a symbol defined in a package from an
+unpackaged context.
+
+To fix this error you'll need to figure out a way around using that symbol, as
+sorbet does not allow packaged code to be used from an unpackaged context.
```
jez, line 735: Can we make this example more concrete or provide more details? It seems like this wording is just restating the error header

For example, something which i'm not sure is true or not, but gets at the style of the edit I'm hoping for:

```
Files that are marked `# packaged: false` must only define and reference
completely unpackaged code. For example, # packaged: false files can be used
to monkeypatch classes defined by the standard library. However, these monkey
patches cannot mention packaged code. To monkeypatch something in the standard
library and also mention a packaged definition do X.
```

packager/VisibilityChecker.cc
elliottt: We don't, but I can add one to a packager test as it would be a nice way of also asserting the error about using packaged symbols from an unpackaged context.

packager/VisibilityChecker.h
elliottt: Makes it impossible to instantiate the class.

main/pipeline/pipeline.cc
elliottt: There are timers in the individual parts, but having one top-level one would be helpful 👍

test/cli/ambiguous-definitions-packages/test.sh
```diff
@@ -4,6 +4,6 @@ set -e
 
 cd test/cli/ambiguous-definitions-packages || exit 1
 
-../../../main/sorbet --silence-dev-message --stripe-packages . 2>&1
+../../../main/sorbet --silence-dev-message --stripe-packages --max-threads=0 . 2>&1
```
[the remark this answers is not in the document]
elliottt: Yeah, the error output ordering wasn't stable otherwise :(

test/cli/package-error-missing-export/test.out
elliottt: We had a lot of additional suggestions in SuggestPackage.cc that get conditionally applied when an error is raised from the resolver. Longer-term, migrating a lot of those into the visibility checker would be the best way to recover their functionality, but in the short term I can add this one back in.

website/docs/error-reference.md
elliottt: We can have both situations arise, though they would get different errors associated with them. I can clarify this a bit here.

website/docs/error-reference.md
elliottt: That sounds good, I'll take a pass over the docs to fix that.

test/cli/package-export-nested-namespace-conflict/parent/b.rb
elliottt: The output turned into `No errors! Great job.`, so it seemed not as valuable as it was before. I can add them back in and update the output instead.

test/cli/package-import-parent-package-conflict/parent/__package.rb
elliottt: The output turned into `No errors! Great job.`, so it seemed not as valuable as it was before. I can add them back in and update the output instead.

test/cli/package-redefine-nested-namespace-error/test.out
elliottt: These errors were trying to address cases that would be problematic because of how the import environments were represented in the previous implementation. As everything lives in the same namespace again with this PR, they don't really make sense to keep.

website/docs/error-reference.md
elliottt: I like the suggestion about mentioning `packaged: false` here. I'll take a pass at rewording this one.

website/docs/error-reference.md
jez: (i fixed this in my edits)

test/cli/package-export-nested-namespace-conflict/parent/b.rb
jez: 🙏 thanks! i think it's always value having edge cases like this remain in the codebase just in case.

test/cli/package-type-alias-export/test.out
```diff
@@ -1,10 +1,7 @@
-__package.rb: Exporting a type alias is not allowed in package specifications https://srb.help/5034
+__package.rb:5: Exporting a type alias is not allowed in package specifications https://srb.help/3722
+     5 |  export MyPackage::A
+          ^^^^^^^^^^^^^^^^^^^
```
jez, line 3: Two questions:

- Why this limitation?
- What do you think about having an error note that putting the type alias in a module an exporting that module instead if the limitation must remain, so that people don't have to click through to the error reference for quick help?

test/cli/simple-package-rbi-gen/test.out
```diff
@@ -1,15 +1,7 @@
...
-module ::Project
-end
-
-module ::Project::Foo
-end
-
 class ::Project::Foo::FooClass
-  extend ::T::Sig
   def dynamically_defined_method(); end
-  def initialize(value); end
```
jez, line 13: ooc, what causes these changes?

test/cli/simple-package/test.out
```diff
@@ -5,16 +5,15 @@ foo/foo.rb:13: Unable to resolve constant `BardClass` https://srb.help/5002
...
+foo/foo.rb:14: Package `Project::Bar` does not export `Project::Bar::UnexportedClass` https://srb.help/3717
     14 |    Project::Bar::UnexportedClass
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   Autocorrect: Use `-a` to autocorrect
     bar/__package.rb:9: Insert `export Project::Bar::UnexportedClass`
      9 |  export Project::Bar::BarMethods
                                          ^
-    bar/bar.rb:14: Defined here
-    14 |  class UnexportedClass; end
-          ^^^^^^^^^^^^^^^^^^^^^
```
jez, line 19: is it worth preserving this error section in the new error?

test/cli/package-type-alias-export/test.out
elliottt: Currently this is an inherited error from the previous implementation. It's not clear to me that we need to carry it forward, but I thought that we could make that decision after landing this PR.

elliottt: I like that suggestion for the error note, that seems super reasonable to me 👍

test/cli/simple-package-rbi-gen/test.out
elliottt: The previous implementation was incorrect, this change reflects the actual minimal rbi. This really caught me off guard, but the previous implementation mangling the names of all the definitions explains why the test.out on master has an rbi that describes the whole packaged source.

test/cli/simple-package/test.out
elliottt: I think that would be an easy fix 👍

test/cli/package-autocorrect-missing-import/test.out
```diff
@@ -1,57 +1,52 @@
...
      7 |      Foo::Bar::OtherPackage::OtherClass # resolves via root
-              ^^^^^^^^
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   Autocorrect: Done
     __package.rb:6: Inserted `import Foo::Bar::OtherPackage`
      6 |  # import Foo::Bar::OtherPackage ## MISSING!
                                                      ^
-    other/__package.rb:3: Defined here
-     3 |class Foo::Bar::OtherPackage < PackageSpec
-        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
jez, line 10: I'm not sure if i mentioned this on another comment, but could we preserve this error section too?

elliottt: Done!

resolver/resolver.cc
```diff
@@ -1289,7 +1289,7 @@ class ResolveConstantsWalk {
         if (checkAmbiguousDefinition(ctx)) {
             const auto ambigDef = findAnyDefinitionAmbiguousWithCurrent(ctx);
             if (ambigDef.exists()) {
-                if (auto e = ctx.beginError(original.loc, core::errors::Resolver::AmbiguousDefinitionError)) {
+                if (auto e = ctx.beginError(original.declLoc, core::errors::Resolver::AmbiguousDefinitionError)) {
                     e.setHeader("Class definition is ambiguous");
```
jez, line 1293: Could we show the ambiguous name also in the header? Right now it only shows up in the error lines.

elliottt: I had meant to break this commit out into a separate PR, do you mind if we push this to a follow-up task?

test/testdata/packager/lsp_features/bart/bart.rb
```diff
@@ -3,7 +3,7 @@
 
 module Bart
   #    ^^^^ def: bartmod
-  #    ^^^^ symbol-search: "Bart", name = "Bart", container = "Bart"
```
jez, line 6: 👀 What the heck did `container = "Bart"` mean before? was that just like a really awkward artifact of how we modeled packages?

resolver/resolver.cc
jez: wfm

test/testdata/packager/lsp_features/bart/bart.rb
elliottt: This took a lot of puzzling with @froydnj earlier today: we're pretty sure that it was using the mangled package name as the container for the name in the response, which is why we needed to change this to `(nothing)` when no longer rewriting the tree.

test/testdata/packager/export_for_test/foo/foo.rb
```diff
@@ -18,26 +18,25 @@ def self.stub_stuff!; end
   # via import Opus::Foo::Bar
   Opus::Foo::Bar::BarClass
   Test::Opus::Foo::Bar::BarClassTest
-# ^^^^ error: Unable to resolve constant `Test`
-
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Used test-only constant `Test::Opus::Foo::Bar::BarClassTest` in non-test file
```
jez, line 21: We use the phrase "test-only" in two contexts:

- test-only import (`test_import`)
- constant defined inside the `Test::` namespace

What do you think about changing the non-`test_import` error message to mention something like

```
`Test::Opus::Foo::...` is defined in a test namespace and cannot be referenced within non-test `Opus::Foo` namespace
```

?

(Also, it's not that `foo/foo.rb` is a test or non-test file is it? It's that `Opus::Foo` is a non-test namespace? lmk if i'm mistaken).

test/testdata/packager/deeply_nested_packages/pass.package-tree.exp
```diff
@@ -3,101 +3,53 @@ class <emptyTree><<C <root>>> < (::<todo sym>)
...
-    <emptyTree>::<C Package_Subpackage_Package_Private$1>::<C Package>::<C PackageClass> = <emptyTree>::<C <PackageRegistry>>::<C Package_Package$1>::<C Package>::<C PackageClass>
-
-    <emptyTree>::<C Package_Subpackage_Package_Private$1>::<C Package>::<C Subpackage>::<C SubpackageClass> = <emptyTree>::<C <PackageRegistry>>::<C Package_Subpackage_Package_Private$1>::<C Package>::<C Subpackage>::<C SubpackageClass>
-  end
-
-  module <emptyTree>::<C <PackageRegistry>><<C <todo sym>>> < ()
-    <emptyTree>::<C Package_Subpackage_Package$1>::<C Package>::<C Subpackage>::<C SubpackageClass> = <emptyTree>::<C <PackageRegistry>>::<C Package_Subpackage_Package_Private$1>::<C Package>::<C Subpackage>::<C SubpackageClass>
-  end
-
-  module <emptyTree>::<C <PackageTests>><<C <todo sym>>> < ()
```
jez, line 83: 👨‍🍳💋🤌

test/testdata/packager/export_for_test/foo/foo.test.rb
```diff
@@ -13,7 +13,6 @@ class Test::Opus::Foo::FooTest
 
   # util/__package.rb exposed via export_for_test, cannot access from here:
   Opus::Util::Nesting.nesting_method
-  #                   ^^^^^^^^^^^^^^ error: Method `nesting_method` does not exist on `T.class_of(Opus::Util::Nesting)`
```
jez, line 16: Is it ok that we're dropping this error?

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+                    testSym.data(ctx)->addLoc(ctx, ctx.locAt(loc));
+                }
+            }
+        }
+    }
+
+    void checkExportPackage(core::MutableContext &ctx, core::LocOffsets loc, core::NameRef litPackage) {
+        if (litPackage != this->package.mangledName() && !this->package.importsPackage(litPackage)) {
+            if (auto e = ctx.beginError(loc, core::errors::Packager::InvalidExport)) {
+                e.setHeader("Exporting a symbol that isn't owned by this package");
```
jez, line 151:
```suggestion
                e.setHeader("Cannot export `{}` because it is owned by another package");
```

(attempting to remove "a symbol" phrasing, and also show the bad class name in the header)

Also: it'd be great to show an error line for where the constant is defined.

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+            isExported = lit.symbol.asClassOrModuleRef().data(ctx)->flags.isExported;
+        } else if (lit.symbol.isFieldOrStaticField()) {
+            isExported = lit.symbol.asFieldRef().data(ctx)->flags.isExported;
+        }
+
+        // Did we use a constant that wasn't exported?
+        if (!isExported) {
+            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::UsedPackagePrivateName)) {
+                auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
+                e.setHeader("Package `{}` does not export `{}`", pkg.show(ctx), lit.symbol.show(ctx));
```
jez, line 356:
```suggestion
                e.setHeader("`{}` exists but has not been exported by `{}", lit.symbol.show(ctx), pkg.show(ctx));
```

jez: I'm also happy to push this to a follow up change, as it looks like this error message was chosen for compatibility with what existed.

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+            }
+            return tree;
+        }
+
+        auto importType = this->package.importsPackage(otherPackage);
+        if (!importType.has_value()) {
+            // We failed to import the package that defines the symbol
+            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::MissingImport)) {
+                auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
+                e.setHeader("No import provides `{}`", lit.symbol.show(ctx));
```
jez, line 324:
```suggestion
                e.setHeader("`{}` exists but has not been imported", lit.symbol.show(ctx));
```

ditto the other comment; it looks like we chose to match the old error message, so i'm happy to push this to the future.

test/cli/package-error-missing-import/test.out
```diff
@@ -1,52 +1,44 @@
...
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   Autocorrect: Use `-a` to autocorrect
     __package.rb:6: Insert `import Foo::Bar::OtherPackage`
      6 |  # import Foo::Bar::OtherPackage ## MISSING!
                                                      ^
     other/__package.rb:3: Defined here
      3 |class Foo::Bar::OtherPackage < PackageSpec
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-  Note:
-    Try running generate-packages.sh
```
jez, line 12: Another place where we might want to preserve the gen-packages hint.

test/cli/package-import-conflicts/test.out
```diff
@@ -1,16 +1,13 @@
...
-                 ^^^^
-    imports_both/__package.rb:4: Conflict from
-     4 |  import A
-                 ^
+ab/ab.rb:3: `A::B` was previously defined as a `class` https://srb.help/4012
+     3 |module A::B
+        ^^^^^^^^^^^
+    a/a.rb:4: Previous definition
+     4 |  class B; end
+          ^^^^^^^
```
jez, line 10: Is it intentional that we have this "previously defined as a `class`" error now?

test/testdata/packager/lsp_features/family.rb
```diff
@@ -16,23 +16,30 @@ def son_catchphrase
           #     ^^^^^^^^^^^ usage: catchphrase
           #     ^^^^^^^^^^^ hover: String
           #     ^           show-symbol: Bart::CatchPhrase
+        # ^^^^ usage: bartmod
```
jez, line 19: ooc, why do we get a bartmod usage here?

jez: oh sorry, the github diff collapsed something and it confused me. nvm!

website/docs/error-reference.md
```diff
@@ -702,6 +702,124 @@ you'll get this error—the latter export is redundant, as it's implied by expor
...
+## 3719
+
+> This error is specific to Stripe's custom `--stripe-packages` definition. If
+> you are at Stripe, please see [go/modularity](http://go/modularity) for more.
+
+Files that are marked `# packaged: false` must only define and reference
+completely unpackaged code. For example, `# packaged: false` files can be used
+to monkeypatch classes defined by the standard library. However, these monkey
+patches cannot mention packaged code. To monkeypatch something in the standard
+library and also mention a packaged definition do X.
```
jez, line 745: flagging that this still says `do X` so that we remember before merging.

packager/packager.cc
```diff
@@ -1897,47 +1857,6 @@ ast::ParsedFile rewritePackage(core::Context ctx, ast::ParsedFile file) {
...
-        ast::MK::Module(core::LocOffsets::none(), core::LocOffsets::none(),
-                        name2Expr(core::Names::Constants::PackageTests()), {}, std::move(publicTestMapping));
-
-    // Add wrapper modules to root of the tree
-    auto &rootKlass = ast::cast_tree_nonnull<ast::ClassDef>(file.tree);
-    rootKlass.rhs.emplace_back(move(packageNamespace));
-    rootKlass.rhs.emplace_back(move(testPackageNamespace));
-    rootKlass.rhs.emplace_back(move(publicMappingNamespace));
-    rootKlass.rhs.emplace_back(move(publicTestMappingNamespace));
-
```
jez, line 1940: :so-beautiful:

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+
+        fast_sort(files, [](const auto &a, const auto &b) -> bool { return a.file < b.file; });
+
+        return files;
+    }
+};
+
+} // namespace
+
+// TODO: validate package imports and give errors for packages that don't exist
```
jez, line 535: What's this TODO for? Should we record it in the followup doc?

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+
+            exportParentNamespace(ctx, sym.data(ctx)->owner);
+        } else if (lit->symbol.isFieldOrStaticField()) {
+            auto sym = lit->symbol.asFieldRef();
+            checkExportPackage(ctx, send.loc, db.getPackageNameForFile(sym.data(ctx)->loc().file()));
+            sym.data(ctx)->flags.isExported = true;
+
+            // When exporting a field, we also export its parent namespace. This is a bit of a hack, and it would be
+            // great to remove this, but this was the behavior of the previous packager implementation.
+            // TODO: can we get away with only marking the owner as exported, and not everything up to the package?
```
jez, line 194: Another TODO. Should we add this to the followup doc?

packager/VisibilityChecker.cc
elliottt: I'm pretty sure this TODO isn't relevant anymore. I should have taken a pass over the visibility checker to remove TODOs before marking this ready 😬

packager/VisibilityChecker.cc
elliottt: I'll take another look and add it if it's still relevant.

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+
+        auto pkgName = gs.packageDB().getPackageNameForFile(f.file);
+        if (!pkgName.exists()) {
+            return f;
+        }
+
+        const auto &package = gs.packageDB().getPackageInfo(pkgName);
+        if (!package.exists()) {
+            return f;
+        }
```
jez, line 233: when would `f` be a package file, but either `pkgName` or `package` doesn't exist?

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+            sym.data(gs)->flags.isExported = true;
+        }
+    }
+
+    // Mark both the package and its test namespace as exported. This will terminate the recursive exporting in
+    // `exportParentNamespace`, as it stops when it hits either the root or an exported symbol.
+    void exportPackageRoots(core::GlobalState &gs) {
+        this->exportRoot(gs, core::Symbols::root());
+
+        auto test = core::Symbols::root().data(gs)->findMember(gs, core::Names::Constants::Test());
```
jez, line 78: This is mostly for my own understanding: it doesn't look like the VisibilityChecker currently references `secondaryTestPackageNamespaceRefs` at all.

I guess that's intentional? `secondaryTestPackageNamespaceRefs` is only for some other features of the packager, but not relevant to visibility checking?

(Asking because I notice that on this line we're hard-coding the `Test` namespace)

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+            owner.data(gs)->flags.isExported = true;
+            owner = owner.data(gs)->owner;
+        }
+    }
+
+    // Lookup the package name on the given root symbol, and mark the final symbol as exported.
+    void exportRoot(core::GlobalState &gs, core::ClassOrModuleRef sym) {
+        // Explicitly mark the symbol outside of the special package namespace as exported by this package, to allow
+        // other packages to refer to this namespace. Ultimately it would be great to not have this, but it helps with a
+        // migration to the visibility checker from the previous packages implementation.
```
jez, line 57: Sorry I don't quite understand what this comment is saying. Could you give more color here?

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+                sym = core::Symbols::noClassOrModule();
+                break;
+            }
+
+            sym = next.asClassOrModuleRef();
+        }
+
+        if (sym.exists()) {
+            sym.data(gs)->flags.isExported = true;
+        }
```
jez, line 70:
```suggestion
                return;
            }

            sym = next.asClassOrModuleRef();
        }

        sym.data(gs)->flags.isExported = true;
```

Pretty sure that this refactoring is equivalent, but feel free to pass on it anyways.

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+    }
+
+    // Mark both the package and its test namespace as exported. This will terminate the recursive exporting in
+    // `exportParentNamespace`, as it stops when it hits either the root or an exported symbol.
```
jez, line 74:
```suggestion
    // Mark both `::A::B` and `::Test::A::B` as exported (given package `A::B`).
    // This will terminate the recursive exporting in `exportParentNamespace`, as it stops when it
    // hits either the root or an exported symbol.
```

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+
+        return tree;
+    }
+
+    ast::ExpressionPtr preTransformClassDef(core::MutableContext ctx, ast::ExpressionPtr tree) {
+        auto &original = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+
+        if (original.symbol == core::Symbols::root()) {
+            return tree;
+        }
```
jez, line 213:
```suggestion
        }

        ENFORCE(original.symbol != core::Symbols::todo());
```

jez: would love to know whether we can ENFORCE this, or whether there would be cases where we failed to remove the todo symbols from class defs before getting to this pass.

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+
+                testSym = member.asClassOrModuleRef();
+            }
+
+            if (testSym.exists()) {
+                auto file = testSym.data(ctx)->loc().file();
+                if (db.getPackageNameForFile(file) != this->package.mangledName()) {
+                    testSym.data(ctx)->addLoc(ctx, ctx.locAt(loc));
+                }
+            }
```
jez, line 144: So this would mean that

```ruby
class A::B < PackageSpec
end
```

makes it so that

```ruby
class Test::A::B
  #            ^ jump-to-def
end
```

makes the jump to def go to the package spec file (among potentially other files)? is that what we desire and/or for compat with existing behavior?

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+            //   1. Export is given a non-constant argument
+            //   2. The argument failed to resolve
+            // In both cases, errors will be raised by previous passes.
+            return tree;
+        }
+
+        auto &db = ctx.state.packageDB();
+        if (lit->symbol.isClassOrModule()) {
+            auto sym = lit->symbol.asClassOrModuleRef();
+            checkExportPackage(ctx, send.loc, db.getPackageNameForFile(sym.data(ctx)->loc().file()));
```
jez, line 183: Were there any follow up changes we wanted to try with `getPackageNameForFile`?

It seems like in this case the API we actually want is to go from a symbol to the package that owns it, and we're doing that in an indirect way right now by way of files and mangled packages names. Not sure if we tried changing this and couldn't, put it off doing it, or was never something we wanted to change from the beginning.

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+                auto file = testSym.data(ctx)->loc().file();
+                if (db.getPackageNameForFile(file) != this->package.mangledName()) {
+                    testSym.data(ctx)->addLoc(ctx, ctx.locAt(loc));
+                }
+            }
+        }
+    }
+
+    void checkExportPackage(core::MutableContext &ctx, core::LocOffsets loc, core::NameRef litPackage) {
+        if (litPackage != this->package.mangledName() && !this->package.importsPackage(litPackage)) {
```
jez, line 149: I'm having a tough time understanding what the `&& !this->package.importsPackage(litPackage)` is for.

Wouldn't we want to always report the error if `litPackage != this->package.mangledName()`?

(It seems like the current behavior would allow something like

```ruby
class A::B < PackageSpec
  import A::X

  export A::X::PrivateFoo
end
```

but correct me if i'm wrong.

jez, line 149: Minor nit, but if the above note about `importsPackage` ends up being the correct interpretation, it might be simpler to change `checkExportPackage` to take `sym` and call `definedByThisPackage(gs, sym)` (deduplicating some logic from both call sites).

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+    bool definedByThisPackage(const core::GlobalState &gs, core::ClassOrModuleRef sym) {
+        auto pkg = gs.packageDB().getPackageNameForFile(sym.data(gs)->loc().file());
+        return this->package.mangledName() == pkg;
+    }
+
+    void recursiveExportSymbol(core::GlobalState &gs, bool firstSymbol, core::ClassOrModuleRef klass) {
+        // There is a loop between a class and its singleton, and this is the easiest way to detect it.
+        if (!firstSymbol && klass.data(gs)->flags.isExported) {
+            return;
+        }
```
jez, line 28: do we have a test for this, ooc?

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+            // TODO: can we get away with only marking the owner as exported, and not everything up to the package?
+            exportParentNamespace(ctx, sym.data(ctx)->owner);
+
+            if (sym.data(ctx)->flags.isStaticFieldTypeAlias) {
+                if (auto e = ctx.beginError(send.loc, core::errors::Packager::ExportingTypeAlias)) {
+                    e.setHeader("Exporting a type alias is not allowed in package specifications");
+                    e.addErrorLine(sym.data(ctx)->loc(), "Originally defined here");
+                }
+            }
+        }
```
jez, line 203: ooc, do we have a test that confirms that it's an error to attempt to export a type member?

Not sure if any of these get close to the problem yet or what but:

```ruby
class A::B < PackageSpec
    export Elem
    export A::B::Elem

    export A::B::Example::Elem
end

class A::B
  extend T::Generic

  Elem = type_member
end

class A::B::Example
  extend T::Generic
  Elem = type_member
end
```

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+
+    // `keep-def` will reference constants in a way that looks like a packaging violation, but is actually fine. This
+    // boolean allows for an early exit when we know we're in the context of processing one of these sends. Currently
+    // the only sends that we process this way will not have any nested method calls, but if that changes this will need
+    // to become a stack.
+    bool ignoreConstant = false;
+
+    ast::ExpressionPtr preTransformSend(core::Context ctx, ast::ExpressionPtr tree) {
+        auto &send = ast::cast_tree_nonnull<ast::Send>(tree);
+        this->ignoreConstant = send.fun == core::Names::keepForIde();
```
jez, line 272:
```suggestion
        ENFORCE(!this->ignoreConstant, "keepForIde has nested sends");
        this->ignoreConstant = send.fun == core::Names::keepForIde();
```

This suitably enforces the comment above, right?

test/testdata/packager/export_for_test/foo/foo.rb
jez: Ok so it looks like my understanding was wrong, and that we actually do report this using the file information, not the namespace.

Question for my own understanding: what is the relationship between a test file and the `Test::` namespace?

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,558 @@
...
+                    e.addErrorLine(ctx.locAt(first->importLoc), "Previous package import found here");
+                    e.replaceWith("Remove import", ctx.locAt(it->importLoc), "");
+                }
+            }
+        }
+
+        this->imports.clear();
+    }
+
+    static std::vector<ast::ParsedFile> run(core::GlobalState &gs, WorkerPool &workers,
```
jez, line 491:
```suggestion
    static std::vector<ast::ParsedFile> run(const core::GlobalState &gs, WorkerPool &workers,
```
Does this work?

jez (commented):
Done leaving comments for now. Feel free to re-request whenever you want more feedback.

packager/VisibilityChecker.cc
elliottt: Yep! I'll make that change.

packager/VisibilityChecker.cc
elliottt: The loop manifests as non-termination without that, but we don't have an explicit test for the part that relies on `firstSymbol`: exporting a child of a class before exporting the parent. I can add a test for that.

packager/VisibilityChecker.cc
elliottt: Unfortunately this is to allow re-exporting constants from other packages, which the current system allows. I think it would be worthwhile removing this behavior, and disallowing re-exports.

packager/VisibilityChecker.cc
elliottt: We weren't raising an error for this, but I've added one and a test.

packager/VisibilityChecker.cc
elliottt: Yep, I think that would be a great refactoring when we can get rid of re-exports 👍

packager/VisibilityChecker.cc
elliottt: I didn't push too hard on that because it felt like adding a `NameRef` to every symbol would be expensive in terms of memory. I also think that would be the best possible api for going from a symbol to the package that owns it, and the current implementation was a bit of a compromise to not add too much overhead for users that aren't Stripe.

packager/VisibilityChecker.cc
elliottt: I don't believe that will happen in this case: that would only happen in the case that the package name is never defined explicitly as a class or module. A case that looks like the following:
```ruby
class A::B < PackageSpec
end

class Test::A::B::C
  #            ^ jump-to-def
end
```

is what would trigger the use of `addLoc`, to ensure that we have an authoritative location for `Test::A::B` that marks it as part of the `A::B` package. Without this, if there's a sub-package definition of `A::B::OtherPackage`, and someone refers to `Test::A::B`, it's a toss-up as to whether the owning package of `Test::A::B` will be `A::B` or `A::B::OtherPackage`.

packager/VisibilityChecker.cc
elliottt: I think that this should be true. I'll make a separate commit with this ENFORCE present so that we can check that.

packager/VisibilityChecker.cc
elliottt: I like it!

packager/VisibilityChecker.cc
elliottt: For a package `A::B`, the symbol that we'll be processing in the context of the `__package.rb` file will be named `<PackageRegistry>::A::B`. Part of the current package implementation requires that you be able to use the name of a package if it's imported, so this captures that by explicitly exporting the constants `A::B` and `Test::A::B`. I'll rephrase this a bit.

packager/VisibilityChecker.cc
jez: That makes sense, thanks!

packager/VisibilityChecker.cc
elliottt: In this case, every package has an associated `Test` namespace, while `secondaryTestPackageNamespaceRefs` is a way of adding additional namespaces that are treated as containing tests. With the current implementation, I believe that if you define a package in the `Critic` namespace and `Critic` is named in the secondary test namespace, it will still be valid to define tests in `Test::Critic`.

packager/VisibilityChecker.cc
elliottt: I'm honestly not sure how this could happen, but the types allow for it so I was being extra careful. I'd be happy to replace the second conditional with an `ENFORCE`.

jez: Either way. I’m happy to also add the ENFORCE but also keep the return

packager/VisibilityChecker.cc
elliottt: I'll add an entry to the followup doc, I think this is worth investigating but there's a lot of reliance on it in the Stripe codebase right now.

test/cli/package-import-conflicts/test.out
elliottt: Yes, this is an effect of everything living in the same namespace again. In file `a/a.rb` the definition looks like this:

```ruby
module A
  class B; end
end
```

While in `ab/ab.rb` it looks like this:

```ruby
module A::B
  class C; end
end
```

This wasn't a problem in the previous design as they lived in separate namespaces, but it's a problem now as we only check visibility in one big namespace.

packager/VisibilityChecker.cc
elliottt: As a follow up, it would also be great to track the export location, so we could point to that here.

packager/VisibilityChecker.cc
elliottt: The behavior of the current packager is to allow references to packaged code from unpackaged contexts, as unpackaged contexts are defined with a cbase. I think that to stay compatible and make it possible to land this PR, we should remove this error for now, and add a TODO to come back to enforcing this.

test/cli/package-error-missing-import/test.out
elliottt: Restored!

test/testdata/packager/export_for_test/foo/foo.rb
elliottt: If a file's extension is `.test.rb` or has `/test/`, it must be defined in the test namespace.

elliottt: I like your suggested phrasing for this error more, I'll switch it over.

test/testdata/packager/export_for_test/foo/foo.test.rb
elliottt: I think this worked before because we would re-create the tree leading up to `Opus::Util::Nesting::Public`, and so `Opus::Util::Nesting` would be a new symbol that existed with no methods. As parts of the Stripe codebase rely on naming the leading (but not explicitly exported) parts of a name, I'm not sure that we can keep this error, though we do have `package_private` to fall back on when this behavior is needed.

packager/VisibilityChecker.cc
```diff
@@ -0,0 +1,601 @@
...
+    void exportPackageRoots(core::GlobalState &gs) {
+        this->exportRoot(gs, core::Symbols::root());
+
+        auto test = core::Symbols::root().data(gs)->findMember(gs, core::Names::Constants::Test());
+        if (test.exists() && test.isClassOrModule()) {
+            this->exportRoot(gs, test.asClassOrModuleRef());
+        }
+    }
+
+    void setPackageLocs(core::MutableContext ctx, core::LocOffsets loc, core::ClassOrModuleRef sym) {
```
ngroman, line 87: Not sure I fully get what's going on here. Can you document?

website/docs/error-reference.md
```diff
@@ -702,6 +702,113 @@ you'll get this error—the latter export is redundant, as it's implied by expor
 
 To fix this error, simply remove the more specific export.
 
+## 3717
```
ngroman, line 705: 🎉

ngroman:
Thanks for taking this on. Much cleaner solution!

packager/VisibilityChecker.cc
elliottt: Yep, this definitely warrants some explanation. I'll add a comment.

jez (approved):

elliottt:

→ https://go/builds/bui_Ld70QuY9EbJ7s1
→ https://go/builds/bui_Ld70JxOlhvkixR

PR #5731, opened 2022-04-30
Fix return type for String#strip! and String#rstrip!

iownthegame:
### Motivation

Doc 
- https://docs.ruby-lang.org/en/2.7.0/String.html#method-i-strip-21
- https://docs.ruby-lang.org/en/2.7.0/String.html#method-i-rstrip-21

Also mentioned in the comments
- https://github.com/iownthegame/sorbet/blob/0ece9516f7d1648bd0b2025d5a4a043da2f11d22/rbi/core/string.rbi#L2029
- https://github.com/iownthegame/sorbet/blob/0ece9516f7d1648bd0b2025d5a4a043da2f11d22/rbi/core/string.rbi#L1730

```rb
$ irb
> "abc".strip!
=> nil
> "abc".rstrip!
=> nil
```

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/string.rbi
+++ b/rbi/core/string.rbi
@@ -1729,7 +1729,7 @@ class String < Object
   # "  hello".rstrip!    #=> nil
   # "hello".rstrip!      #=> nil
   # ```
-  sig {returns(String)}
+  sig {returns(T.nilable(String))}
   def rstrip!(); end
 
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_Lbxihs78v8B0hu
→ https://go/builds/bui_Lbxii18FpvwAVn

PR #5732, opened 2022-05-02
Reject overrides with incompatible visibility

### LINKED ISSUE #5522: Reject implementing public abstract method with private override method
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Arequire%20'sorbet-runtime'%0A%0Aclass%20Module%0A%20%20include%20T%3A%3ASig%0A%20%20include%20T%3A%3AHelpers%0Aend%0A%0Amodule%20IFoo%0A%20%20interface!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aclass%20Foo%0A%20%20include%20IFoo%0A%20%20sig%20%7Boverride.void%7D%0A%20%20private%20def%20foo%3B%20end%0Aend%0A%0Asig%20%7Bparams%28x%3A%20IFoo%29.void%7D%0Adef%20takes_ifoo%28x%29%3B%20x.foo%3B%20end%0A%0A%23%20static%3A%20no%20error%0A%23%20runtime%3A%20raises%20%22private%20method%20%60foo%60%20called%22%0Atakes_ifoo%28Foo.new%29)

```ruby
# typed: strict
require 'sorbet-runtime'

class Module
  include T::Sig
  include T::Helpers
end

module IFoo
  interface!

  sig {abstract.void}
  def foo; end
end

class Foo
  include IFoo
  sig {override.void}
  private def foo; end
end

sig {params(x: IFoo).void}
def takes_ifoo(x); x.foo; end

# static: no error
# runtime: raises "private method `foo` called"
takes_ifoo(Foo.new)
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

There should be a static error at the place where we check that `Foo` implements the interface `IFoo`.
We should emit an error giving two options:

1.  mark `IFoo#foo` private
1.  mark `Foo#foo` public

This error is specific to this ordering of parent/child. The other way (`IFoo#foo` is private, `Foo#foo` is public) is actually fine, because it just needs to be at least as good as the parent.

- - -

dgollahon:
- Closes #5522

Hi there, this is my first contribution and I don't usually write c++ so I'm very open to feedback and happy to make any changes. I picked this because it seemed like a good place to learn more about contributing to sorbet. Thanks!

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -112,13 +112,32 @@ void matchPositional(const core::Context ctx, absl::InlinedVector<reference_wrap
     }
 }
 
-// Ensure that two argument lists are compatible in shape and type
+// Ensure that two argument lists are compatible in shape and type and that method visibility is compatible
 void validateCompatibleOverride(const core::Context ctx, core::MethodRef superMethod, core::MethodRef method) {
-    if (method.data(ctx)->flags.isOverloaded) {
+    auto methodData = method.data(ctx);
+
+    if (methodData->flags.isOverloaded) {
         // Don't try to check overloaded methods; It's not immediately clear how
         // to match overloads against their superclass definitions. Since we
         // Only permit overloading in the stdlib for now, this is no great loss.
         return;
+    }
+
+    auto superMethodData = superMethod.data(ctx);
+
+    auto isPrivate = methodData->flags.isPrivate;
+    auto isProtected = methodData->flags.isProtected;
+    auto superIsProtected = superMethodData->flags.isProtected;
+    auto superIsPublic = superMethodData->isMethodPublic();
+
+    if ((isPrivate && (superIsProtected || superIsPublic)) || (isProtected && superIsPublic)) {
+        if (auto e = ctx.state.beginError(methodData->loc(), core::errors::Resolver::BadMethodOverride)) {
+            e.setHeader("Override of `{}` must be {} but `{}` is {}. Alternatively, `{}` can be made {}",
+                        superMethod.show(ctx), superIsProtected ? "protected or public" : "public", method.show(ctx),
+                        isPrivate ? "private" : "protected", superMethod.show(ctx),
+                        isPrivate ? "private" : "protected");
+            e.addErrorLine(superMethodData->loc(), "Base method defined here");
+        }
     }
 
     auto left = decomposeSignature(ctx, superMethod);
@@ -131,26 +150,26 @@ void validateCompatibleOverride(const core::Context ctx, core::MethodRef superMe
             if (auto e = ctx.state.beginError(method.data(ctx)->loc(), core::errors::Resolver::BadMethodOverride)) {
                 e.setHeader("{} method `{}` must accept at least `{}` positional arguments",
                             implementationOf(ctx, superMethod), superMethod.show(ctx), leftPos);
-                e.addErrorLine(superMethod.data(ctx)->loc(), "Base method defined here");
+                e.addErrorLine(superMethodData->loc(), "Base method defined here");
             }
         }
     }
 
     if (auto leftRest = left.pos.rest) {
         if (!right.pos.rest) {
-            if (auto e = ctx.state.beginError(method.data(ctx)->loc(), core::errors::Resolver::BadMethodOverride)) {
+            if (auto e = ctx.state.beginError(methodData->loc(), core::errors::Resolver::BadMethodOverride)) {
                 e.setHeader("{} method `{}` must accept *`{}`", implementationOf(ctx, superMethod),
                             superMethod.show(ctx), leftRest->get().show(ctx));
-                e.addErrorLine(superMethod.data(ctx)->loc(), "Base method defined here");
+                e.addErrorLine(superMethodData->loc(), "Base method defined here");
             }
         }
     }
 
     if (right.pos.required.size() > left.pos.required.size()) {
-        if (auto e = ctx.state.beginError(method.data(ctx)->loc(), core::errors::Resolver::BadMethodOverride)) {
+        if (auto e = ctx.state.beginError(methodData->loc(), core::errors::Resolver::BadMethodOverride)) {
             e.setHeader("{} method `{}` must accept no more than `{}` required argument(s)",
                         implementationOf(ctx, superMethod), superMethod.show(ctx), left.pos.required.size());
-            e.addErrorLine(superMethod.data(ctx)->loc(), "Base method defined here");
+            e.addErrorLine(superMethodData->loc(), "Base method defined here");
         }
     }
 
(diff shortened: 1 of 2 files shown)


definition_validator/validator.cc
```diff
@@ -112,15 +112,34 @@ void matchPositional(const core::Context ctx, absl::InlinedVector<reference_wrap
...
+    auto methodData = method.data(ctx);
+
+    if (methodData->flags.isOverloaded) {
         // Don't try to check overloaded methods; It's not immediately clear how
         // to match overloads against their superclass definitions. Since we
         // Only permit overloading in the stdlib for now, this is no great loss.
         return;
     }
 
+    auto superMethodData = superMethod.data(ctx);
```
dgollahon, line 126: Based on the [internals doc](https://github.com/sorbet/sorbet/blob/master/docs/internals.md) my understanding is that the `.data` calls are somewhat slow and I found the lookup repeated in several places (though largely inside conditionals). I figured it might make sense to extract this and replace it below but I can make this change more isolated if preferred.

froydnj:

→ https://go/builds/bui_LcEsKhxOsmTz9B
→ https://go/builds/bui_LcEsAfyqZTqPNq

definition_validator/validator.cc
```diff
@@ -112,15 +112,34 @@ void matchPositional(const core::Context ctx, absl::InlinedVector<reference_wrap
...
+    auto isProtected = methodData->flags.isProtected;
+    auto superIsProtected = superMethodData->flags.isProtected;
+    auto superIsPublic = superMethodData->isMethodPublic();
+
+    if ((isPrivate && (superIsProtected || superIsPublic)) || (isProtected && superIsPublic)) {
+        if (auto e = ctx.state.beginError(methodData->loc(), core::errors::Resolver::BadMethodOverride)) {
+            e.setHeader("Override of `{}` must be {} but `{}` is {}. Alternatively, `{}` can be made {}",
+                        superMethod.show(ctx), superIsProtected ? "protected or public" : "public", method.show(ctx),
+                        isPrivate ? "private" : "protected", superMethod.show(ctx),
+                        isPrivate ? "private" : "protected");
```
froydnj, line 138: I think the "Alternatively, ..." bit here is suboptimal for two reasons:

1. It makes an already long error message longer by repeating `superMethod`; and
2. There's not really a first course of action that it's providing an alternative for.  I guess the implicit course of action is that `method` should have its visibility flipped, but I don't think that's super clear.

I think if we *did* want to have the suggestion about how to correct the problem, we'd relegate it to an `e.addErrorLine` call.  But since there's two courses of action and we're not sure which one to choose, let's leave it out for now.

(I don't know if the suggestion can be applied as-is due to formatting lints.)

```suggestion
            e.setHeader("Override of `{}` must be {} but `{}` is {}.",
                        superMethod.show(ctx), superIsProtected ? "protected or public" : "public", method.show(ctx),
                        isPrivate ? "private" : "protected");
```

froydnj (changes requested):
Thanks for the patch!  One relatively minor comment below.

This PR turns up ~100 errors on Stripe's codebase, which will take a little while to fix, and those fixes will gate landing this PR.

definition_validator/validator.cc
jez: > But since there's two courses of action and we're not sure which one to choose, let's leave it out for now.

It seems like the common case of `public` in parent (because it's an interface method) and `private` in child should be almost always be resolved by removing `private` from the child, not by adding `private` to the parent.

Defaulting the suggestion to removing `private` also has the nice property that it's unlikely to cause cascading type errors (meanwhile, asking the user to add `private` to the parent class might not work if the parent method is already being called publicly).

(This is basically what froydnj's suggestion implements; I just wanted to chime in my two cents.)

jez:
One thing that would make it easier for us to migrate Stripe's codebase (which will be required before we can land this change) as well as make it easier for others to adopt this Sorbet version would be if there was an autocorrect for this error. 

One quick-and-dirty way to build such an autocorrect suggestion would be to check whether

```cpp
methodData->loc().adjustLen(ctx, -8, 8).source(ctx) == "private "
```

and then using that with `e.replaceWith` (you should be able to grep to find examples of using replaceWith)

You might also want to add a couple more cases for e.g. `private_class_method`. It's honestly likely fine to ignore `protected` for the sake of autocorrects? (it's rarely used), but up to you.

It's unfortunately tricky to do this in a more principled way (e.g., by looking for the `private` in the AST) because at the point where definition_validator runs, the `private` method call has been moved somewhere else, and the method definition stands on its own. We also don't store any location information on the `methodData` that would point to where the `private` modifier lives in the source tree.

image

But building a 95% solution for autocorrects like this is almost always fine, and goes a long way towards making the sorbet upgrade process smoother.

jez:
To give you a sense of what these error messages look like on Stripe's codebase:

```
ops/lib/visa/clearing/clearing_generator/transaction_clearing_generator.rb:257: Override of Opus::Ops::Visa::Clearing::AbstractClearingGenerator#compute_authorization_date must be public but Opus::Ops::Visa::Clearing::TransactionClearingGenerator#compute_authorization_date is private. Alternatively, Opus::Ops::Visa::Clearing::AbstractClearingGenerator#compute_authorization_date can be made private http://go/e/5035
     257 |      private def compute_authorization_date
    ops/lib/visa/clearing/clearing_generator/abstract_clearing_generator.rb:357: Base method defined here
     357 |      def compute_authorization_date; end
```

I'd love if we could change these messages to make the error shorter and put the most relevant information in the beginning:

```
ops/lib/visa/clearing/clearing_generator/transaction_clearing_generator.rb:257: Method `compute_authorization_date` is private in `Opus::Ops::Visa::Clearing::TransactionClearingGenerator` but not in `Opus::Ops::Visa::Clearing::AbstractClearingGenerator` http://go/e/5035
     257 |      private def compute_authorization_date
    ops/lib/visa/clearing/clearing_generator/abstract_clearing_generator.rb:357: Base method defined here
     357 |      def compute_authorization_date; end
```

This:

- mentions the specific method early, not after the class name
- only mentions the method once, not once per parent / child class
- mentions `private` before the potentially-long class name

Incidentally, this more closely matches the error message TypeScript would generate

image

definition_validator/validator.cc
dgollahon: Thanks for the feedback. I am completely happy to drop the second part of the message. I felt like it was a little verbose too but was trying to match the requirements/recommendation in the [original issue](https://github.com/sorbet/sorbet/issues/5522#issue-1185194258). I'll take another pass at this tomorrow night or maybe tonight if I get a few minutes.

dgollahon:
> One thing that would make it easier for us to migrate Stripe's codebase (which will be required before we can land this change) as well as make it easier for others to adopt this Sorbet version would be if there was an autocorrect for this error.
> 
> One quick-and-dirty way to build such an autocorrect suggestion would be to check whether
> 
> ```c++
> methodData->loc().adjustLen(ctx, -8, 8).source(ctx) == "private "
> ```
> 
> and then using that with `e.replaceWith` (you should be able to grep to find examples of using replaceWith)
> 

I can take a look at this when I update the wording if desired, though I am a bit hesitant to be outside of the AST/metadata. In particular, it seems like this would only work if you use inline access modifiers. Most ruby codebases I've worked on use region-based `private`/`proected`/etc. and also separate `private_class_method :name` style declarations. I am happy to take a pass at this to help get it shipped but it wouldn't help my use case and I'm not sure what fraction of ruby code it would help.

jez:
> I am happy to take a pass at this to help get it shipped but it wouldn't help my use case and I'm not sure what fraction of ruby code it would help.

Makes sense, though I still think that any autocorrect, even if it doesn't always trigger, is better than no autocorrect.

Stripe's codebase exclusively uses the `private def foo; end` syntax, so I suggested that one as if it existed, it would mean getting this PR merged sooner, but again, up to you whether/how much time you want to spend building autocorrects.

PR #5740, opened 2022-05-04
avoid jump-to-def into RBI files for classes if there are multiple locations

froydnj:
### Motivation

This comes up as a user request from time to time, and also makes more sense: the user wants to see the definition, which almost certainly doesn't live in an RBI file.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/definitions_rb_and_rbi__3.rbi
+++ b/test/testdata/lsp/definitions_rb_and_rbi__3.rbi
@@ -1,0 +1,15 @@
+# typed: true
+
+# The usage annotation here is kind of a hack: we want the only
+# definition for go-to-definition to exist in the rb file defining
+# MyClass, but the test harness will see this class if we don't
+# say that it's a "use", even though semantically it is a def.
+class MyClass
+    # ^^^^^^^ usage: MyClass
+  def magic_dynamically_defined_fun; end
+end
+
+class MyModule
+    # ^^^^^^^^ usage: MyModule
+  def magic_dynamically_defined_fun; end
+end
--- a/main/lsp/requests/definition.cc
+++ b/main/lsp/requests/definition.cc
@@ -31,9 +31,21 @@ unique_ptr<ResponseMessage> DefinitionTask::runRequest(LSPTypecheckerInterface &
         // Only support go-to-definition on constants and fields in untyped files.
         if (auto c = resp->isConstant()) {
             auto sym = c->symbol;
+            vector<pair<core::Loc, unique_ptr<Location>>> locMapping;
             for (auto loc : sym.locs(gs)) {
-                addLocIfExists(gs, locations, loc);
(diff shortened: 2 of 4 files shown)


test/testdata/lsp/definitions_rb_and_rbi__3.rbi
```diff
@@ -0,0 +1,10 @@
+# typed: true
+
+# The usage annotation here is kind of a hack: we want the only
+# definition for go-to-definition to exist in the rb file defining
+# MyClass, but the test harness will see this class if we don't
+# say that it's a "use", even though semantically it is a def.
+class MyClass
+    # ^^^^^^^ usage: MyClass
```
froydnj, line 8: I guess the better way of phrasing is it that the "usage:" assertion isn't really a "use"; it's a check for "find all references".  So by putting a "usage:" here, we're saying:

1. this shouldn't be found as a def by go-to-definition (it is found as a def prior to this PR)
2. this *should* be found as a reference via "find all references" (as it still is with this PR)

which is the behavior we want, even though the assertion names make it a little awkward.

froydnj:
We discussed this offline and decided that for the case of methods defined in `.rb` and `.rbi` files but having a `sig` in only the `.rbi` file, choosing the `.rbi` file would require heuristically parsing sigs in all the relevant files and would therefore be too slow.  (We can tell that the method has a `sig`, but from within the LSP code, we cannot tell the provenance of the `sig` without parsing.)  The same logic applies for fields.  Depending on how the relevant `.rb` and `.rbi` files are written, we already choose the `.rb` file for jump-to-def on a method from a send, since we prefer `.rb` files for canonical locs:

https://github.com/sorbet/sorbet/blob/1a7c3fcd71c93daf970f42ba35db07936ffeae6e/core/Symbols.cc#L2378-L2389

and we jump to the canonical loc for the associated method (not all the locs, as we do for other kinds of responses):

https://github.com/sorbet/sorbet/blob/1a7c3fcd71c93daf970f42ba35db07936ffeae6e/main/lsp/requests/definition.cc#L47-L56

In the end, we decided to limit this to just classes and methods, which makes the patch somewhat less invasive.

I tried writing testcases to show that e.g. jump-to-def on a send would choose the Ruby file, but our test harness is not expressive enough to say:

1. There is a definition in a `.rb` file
2. There is a definition in a `.rbi` file
3. Jump-to-def on *this* send should correspond to the single definition in the `.rb` file.

...or at least I couldn't figure out how to express that within the test harness.  (I did try symbol versions, which did not seem to be sufficient.  That behavior seems like a bug, but I was not willing to unravel the mess that is def/use/ref checking in `lsp_test_runner.cc` for this particular change.)

So given that this change only affects classes and modules, I felt that the testcase added that guarantees behavior for classes and modules was sufficient.

jez (approved):

PR #5743, opened 2022-05-04
srb init special case for ruby2_keywords in Ruby 3.1

### LINKED ISSUE #5741: Invalid `****` token in generated RBI files
#### Input

Generate a new Rails project, install sorbet, and run `srb init`. 
#### Observed output

The generated gem RBI (e.g. activejob.rbi, but many others as well) have invalid tokens

```
module ActiveJob::Enqueuing::ClassMethods
  def job_or_instantiate(*args, ****); end
  def perform_later(*args, ****); end
end
```
and
```
class ActiveJob::Base
  ...
  def queue_adapter(**, ****, &&); end
  ...
end
```
This results in errors across the entire project and the RBI being set to `typed: ignore`.

#### Expected behavior

The file consists of valid tokens and is able to be included in the project without errors.

jez:
This prints something different on Ruby 3.0 vs Ruby 3.1

```ruby
class A
  ruby2_keywords def foo(*args); end
end
p(A.instance_method(:foo).parameters)
```

Ruby 3.0

```ruby
[[:rest, :args]]
```

Ruby 3.1

```ruby
[[:rest, :args], [:keyrest, :**]]
```

I have no clue why this behavior changed, but it meant that we would
serialize a gem RBI that looked like

```ruby
class A
  def foo(*args, ****)
  end
end
```

which is a syntax error.

### Motivation

Fixes #5741

### Test plan

I have decided not to write a test for this, because the test case would
have different behavior on Ruby 3.0 and Ruby 3.1, making it hard to fit
into our existing test snapshot testing framework.

I tested that using a local copy of the `sorbet` and `sorbet-static` gem fixed
the issue when running `srb init` for a gem that had a definition like the
above.

### DIFF
--- a/gems/sorbet/lib/gem-generator-tracepoint/tracepoint_serializer.rb
+++ b/gems/sorbet/lib/gem-generator-tracepoint/tracepoint_serializer.rb
@@ -197,7 +197,10 @@ module Sorbet::Private
           when :rest
             "*#{name}"
           when :keyrest
-            "**#{name}"
+            case name
+            when :** then "**"
+            else "**#{name}"
+            end
           when :block
             "&#{name}"
           else
--- a/gems/sorbet/lib/gem-generator-tracepoint/tracepoint_serializer.rb
+++ b/gems/sorbet/lib/gem-generator-tracepoint/tracepoint_serializer.rb
@@ -195,14 +195,20 @@ module Sorbet::Private
           when :key
             "#{name}: nil"
           when :rest
-            "*#{name}"
+            case name
+            when :* then "*"
+            else "*#{name}"
+            end
           when :keyrest
             case name
             when :** then "**"
             else "**#{name}"
             end
           when :block
-            "&#{name}"
+            case name
+            when :& then "&"
+            else "&#{name}"
+            end
           else
             raise "Unknown parameter type: #{type}"
           end


froydnj (approved):
Do we actually want to serialize it as `**`, given that the original definition didn't mention it at all?  It also seems like there might be cases where we get `**name` back from Ruby:

https://github.com/ruby/ruby/blob/bff31b3208071d40f9e193ef0cb95fb7aa2db62c/proc.c#L3230-L3240

jez:
> Do we actually want to serialize it as `**`, given that the original definition didn't mention it at all? It also seems like there might be cases where we get `**name` back from Ruby:
> 
> https://github.com/ruby/ruby/blob/bff31b3208071d40f9e193ef0cb95fb7aa2db62c/proc.c#L3230-L3240

Hm, not sure what you mean? I think that if we get back something from Ruby like

```ruby
[:keyrest, :**]
```

there is only one thing we might want to do here, and that's call the argument `**` (which is valid syntax in both Ruby and Sorbet).

I think that the `**name` case is still handled, because it would come back like:

```ruby
[:keyrest, :name]
```

and so we would not trip the special case added in this PR, thus falling back to the current behavior.

jez:
> I also noticed an invalid `&&` tag being generated

Thanks, it looks like we should just handle all three.

froydnj:
> > Do we actually want to serialize it as `**`, given that the original definition didn't mention it at all? It also seems like there might be cases where we get `**name` back from Ruby:
> > https://github.com/ruby/ruby/blob/bff31b3208071d40f9e193ef0cb95fb7aa2db62c/proc.c#L3230-L3240
> 
> Hm, not sure what you mean? I think that if we get back something from Ruby like
> 
> ```ruby
> [:keyrest, :**]
> ```
> 
> there is only one thing we might want to do here, and that's call the argument `**` (which is valid syntax in both Ruby and Sorbet).
> 
> I think that the `**name` case is still handled, because it would come back like:
> 
> ```ruby
> [:keyrest, :name]
> ```
> 
> and so we would not trip the special case added in this PR, thus falling back to the current behavior.

This case specifically: https://github.com/ruby/ruby/blob/bff31b3208071d40f9e193ef0cb95fb7aa2db62c/proc.c#L3231-L3233

will return `**foo`, which will fall through to the `"**#{name}" case in this PR, which means we return `****foo`.  I don't know how to trigger this case, though; all the things I tried on https://try.ruby-lang.org would be handled by this PR.

It'd be nice if we could avoid printing `**` in the output given that it wasn't present in the original, but maybe that's not feasible (or desirable, since the `ruby2_keywords` is specifically requesting different behavior).

froydnj (approved):

jez:
> This case specifically: https://github.com/ruby/ruby/blob/bff31b3208071d40f9e193ef0cb95fb7aa2db62c/proc.c#L3231-L3233
> 
> will return `**foo`

I think you're looking at the code for `inspect`, which pretty-prints the arity into a string, not the code for `parameters`, so I don't think we can ever get back an array like

```ruby
[:keyrest, :"**name"]
```

froydnj:
> > This case specifically: https://github.com/ruby/ruby/blob/bff31b3208071d40f9e193ef0cb95fb7aa2db62c/proc.c#L3231-L3233
> > will return `**foo`
> 
> I think you're looking at the code for `inspect`, which pretty-prints the arity into a string, not the code for `parameters`,

:man_facepalming:

PR #5744, opened 2022-05-04
Remove disclaimer from error-reference.md

jez:
### Motivation

We have a test that the list of errors is actually exhaustive now.

### Test plan

n/a


elliottt (approved):
Awesome! 🎉

PR #5745, opened 2022-05-05
don't allocate in `FileOps::readLinefromFd`

froydnj:
### Motivation

A little less allocation here seems like a good thing.  (I think it would also be a good thing to tackle the same sort of problems in `LSPInput.cc`, but the surgery required there is more extensive.)

### Test plan

See included automated tests.

### DIFF
--- a/common/common.cc
+++ b/common/common.cc
@@ -142,7 +142,7 @@ string_view sorbet::FileOps::getExtension(string_view path) {
     return path.substr(found + 1);
 }
 
-int sorbet::FileOps::readFd(int fd, std::vector<char> &output, int timeoutMs) {
+int sorbet::FileOps::readFd(int fd, absl::Span<char> output, int timeoutMs) {
     // Prepare to use select()
     fd_set set;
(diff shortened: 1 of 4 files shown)


common/common.cc
```diff
@@ -179,7 +179,8 @@ sorbet::FileOps::ReadLineOutput sorbet::FileOps::readLineFromFd(int fd, string &
     }
 
     constexpr int BUFF_SIZE = 1024 * 8;
-    vector<char> buf(BUFF_SIZE);
+    char chars[BUFF_SIZE];
+    absl::Span<char> buf(chars);
```
jez, line 183: Huh. What happens if there’s more to read?

jez (approved):

common/common.cc
froydnj: We buffer up what we just read and "timeout":

https://github.com/sorbet/sorbet/blob/8f47605b1a4327a90ccebc621be511f660844e5e/common/common.cc#L192-L208

PR #5751, opened 2022-05-05
Remove redundant code from the packager pass

elliottt:
After the switch to checking visibility explicitly in #5727, the old approach of wrapping environments in different namespaces is no longer needed. This PR removes most references to `PackageRegistry` and `PackageTests`, as well as the pass that would construct the special namespaces when processing a `__package.rb`.

### Motivation

Removing dead code.

### Test plan

See included automated tests.


jez (approved):

elliottt:

→ https://go/builds/bui_LdX3ZcgoqjlTsf
→ https://go/builds/bui_LdX3c4iyIQs9Aa

PR #5756, opened 2022-05-06
detect use of `each` and recommend `test_each` instead

froydnj:
### Motivation

The need to use `test_each` often puzzles people, especially because they get inscrutable static errors *and* they don't get any runtime errors when they use `each`.  Try to determine if people are doing this and provide an error that points them in the correct direction.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_LdnGtjuDlhZlpg
→ https://go/builds/bui_LdnGbjaZtdaPqF

PR #5759, opened 2022-05-06
Try removing the minitest DSL

jez:
### Test plan

See included automated tests.

### DIFF
--- a/rewriter/rewriter.cc
+++ b/rewriter/rewriter.cc
@@ -94,12 +94,6 @@ public:
                         return;
                     }
 
-                    nodes = Minitest::run(ctx, isClass, &send);
-                    if (!nodes.empty()) {
-                        replaceNodes[stat.get()] = move(nodes);
-                        return;
-                    }
-
                     nodes = DSLBuilder::run(ctx, &send);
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_LdqtqaDVC97TBr
→ https://go/builds/bui_LdqtV5KugHmtoV

Morriar:
This rewriter is very useful for us. Why is it considered for removal?

jez:
We think we might be able to replace it with T.proc.bind(T.attached_class).void on the ‘it’ method, which would let code like `[true, false].each {it ‘’ {}}` typecheck without needing to use test_each. Nathan was looking further into it

froydnj:
> We think we might be able to replace it with `T.proc.bind(T.attached_class).void` on the ‘it’ method, which would let code like `[true, false].each {it ‘’ {}}` typecheck without needing to use test_each. Nathan was looking further into it

Despite the coolness of `T.proc.bind(T.{self_type,attached_class})` is, it's not sufficient for modeling minitest, and the minitest rewriter, despite all its problems, is a much better option.

PR #5764, opened 2022-05-07
store Ruby strings in a single table for compiled modules

froydnj:
### Motivation

Following on to #5613 and #5646, we do the same thing for Ruby strings.

The main benefit here is less LLVM IR to optimize.  I think that this change also silently interacts more correctly with the compacting GC in Ruby 2.7+, since we're now marking the place(s) we store the Ruby string(s) rather than the strings themselves.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/Core/CompilerState.h
+++ b/compiler/Core/CompilerState.h
@@ -95,6 +95,7 @@ private:
 public:
     llvm::Value *stringTableRef(std::string_view str);
     llvm::Value *idTableRef(std::string_view str);
+    llvm::Value *rubyStringTableRef(std::string_view str);
 
     // useful apis for getting common types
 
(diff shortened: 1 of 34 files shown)


compiler/Core/CompilerState.h
```diff
@@ -42,13 +42,31 @@ struct IDTable {
     void defineGlobalVariables(llvm::LLVMContext &lctx, llvm::Module &module, llvm::IRBuilderBase &builder);
 };
 
+struct RubyStringTable {
+    struct RubyStringTableEntry {
```
froydnj, line 46: These structures should all really be `Entry`, since they are scoped to the table class they are a part of.  I plan on making that change in a followup to this PR.

jez (approved):

PR #5768, opened 2022-05-08
represent call flags as actual constants rather than function calls

froydnj:
### Motivation

Less LLVM IR to churn through and trivially inline.  This change also enables the value returned from `CallCacheFlags::build` to be a constant value, which enables a) more parts of a function call cache to be represented trivially at compile time and b) opens the door to a more efficient (space-wise) initialization of function call caches.

### Test plan

See included automated tests.  The `*.exp` changes are mostly changes due to no longer inlining some called-once functions that are part of initializing compiled modules.  This change seems peculiar to me; my best explanation is that having the flags be represented as actual constants, rather than trivially inlinable functions, tickles the inliner's heuristics just enough that it doesn't see those initialization functions as actually profitable to inline (since the initialization functions don't also contain things that would inline) despite the initialization functions only being called once.  I think this is a good thing, as inlining all that initialization code would have created gargantuan functions for larger modules, which only results in a pile of relatively unoptimizable code (since it's all just function calls to external functions anyway).  So a win on multiple fronts for relatively less LLVM IR to optimize.  (Maybe it's slightly more LLVM IR if those functions don't get inlined, but at least there's smaller functions on balance, which tends to be better for the register allocator and things like that.)


elliottt (approved):
This is awesome! You can tell by how much lower the fresh name indexes are that there's much less intermediate code generated 🎉

PR #5770, opened 2022-05-09
Add error for test_each_hash without pos hash arg

jez:
### Motivation

It was confusing that the Minitest rewriter did not fire without an error.

### Test plan

See included automated tests.


froydnj (approved):
I was going to make the rewriter model Ruby keyword arg passing, but this is *much* simpler.

Issue #5775, opened 2022-05-10
Reduce `T::Enum` JSON verbosity with ActiveSupport


cocoahero:
In a project where `ActiveSupport` may be included (such as Rails), there is a case where performing `to_json` on an object which contains a `T::Enum` property returns an overly verbose representation for the enum value.

It looks like `T::Enum` already overrides `to_json` to return the serialized value. However, `ActiveSupport` (and Ruby Stdlib in some confusing scenarios) defaults to calling `as_json` from `to_json` to gain a JSON representation of an object before serializing to a JSON string. Therefore, in the example below, `T::Enum#as_json` gets called instead which by default returns a very verbose representation.

I believe solving this should be as simple as providing an `as_json(*opts)` implementation similar to the existing `to_json(*opts)`.

#### Input

```ruby
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'sorbet', :group => :development
  gem 'sorbet-runtime'
  gem 'activesupport'
end

require 'active_support'
require 'active_support/core_ext/object/json'

class State < T::Enum
  extend (T::Sig)

  enums do
    Draft = new("draft")
    Published = new("published")
    Archived = new("archived")
  end
end

class Article
  attr_accessor :title
  attr_accessor :state

  def initialize(title, state = State::Draft)
    @title = title
    @state = state
  end
end

puts Article.new("Hello World").to_json
```

#### Observed output

```json
{"title":"Hello World","state":{"serialized_val":"draft","const_name":"Draft"}}
```

#### Expected behavior

```json
{"title":"Hello World","state":"draft"}
```

PR #5777, opened 2022-05-10
Add `T::Enum#as_json` implementation

cocoahero:
### Motivation

Fixes https://github.com/sorbet/sorbet/issues/5775

Within Rails projects, it is very common to use `to_json` on objects to serialize for logging or HTTP responses. When using Sorbet enums in these objects, the default behavior serializes a very verbose representation. This PR implements `as_json` on `T::Enum` to return a simplified representation. See the linked issue for more details.

### Test plan

See included automated tests.

**Note:** While this issue is focused on interoperability with `ActiveSupport`, I didn't add any tests specific to it as it would require a new development dependency. Happy to change this approach if deemed necessary.

### DIFF
--- a/gems/sorbet-runtime/lib/types/enum.rb
+++ b/gems/sorbet-runtime/lib/types/enum.rb
@@ -160,7 +160,7 @@ class T::Enum
 
   sig {params(args: T.untyped).returns(T.untyped)}
   def as_json(*args)
-    serialize
+    serialize.as_json(*args)
   end
 
   sig {returns(String)}
--- a/gems/sorbet-runtime/lib/types/enum.rb
+++ b/gems/sorbet-runtime/lib/types/enum.rb
@@ -158,6 +158,11 @@ class T::Enum
     serialize.to_json(*args)
   end
 
+  sig {params(args: T.untyped).returns(T.untyped)}
+  def as_json(*args)
+    serialize
+  end
+
   sig {returns(String)}
   def to_s
     inspect
--- a/gems/sorbet-runtime/lib/types/enum.rb
+++ b/gems/sorbet-runtime/lib/types/enum.rb
@@ -160,7 +160,9 @@ class T::Enum
 
   sig {params(args: T.untyped).returns(T.untyped)}
   def as_json(*args)
-    serialize.as_json(*args)
+    serialized_val = serialize
+    return serialized_val unless serialized_val.respond_to?(:as_json)
+    serialized_val.as_json(*args)
   end
 
   sig {returns(String)}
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/enum.rb
```diff
@@ -158,6 +158,11 @@ def to_json(*args)
     serialize.to_json(*args)
   end
 
+  sig {params(args: T.untyped).returns(T.untyped)}
+  def as_json(*args)
+    serialize
```
amomchilov, line 163: The serialized value might not be a string/symbol (or other JSON compatible type), so I think we need to call `as_json` on it, to give it a chance to serialize itself appropriately:

```suggestion
    serialize.as_json(*args)
```

jez (approved):

gems/sorbet-runtime/lib/types/enum.rb
cocoahero: Hmm, that's a fair point. However, now the tests won't pass as is because `as_json` doesn't exist on `String` unless ActiveSupport is included...

gems/sorbet-runtime/lib/types/enum.rb
```diff
@@ -158,6 +158,11 @@ def to_json(*args)
     serialize.to_json(*args)
   end
 
+  sig {params(args: T.untyped).returns(T.untyped)}
+  def as_json(*args)
+    serialize.as_json(*args)
```
paracycle, line 163: Maybe:
```suggestion
    serialized_val = serialize
    return serialized_val unless serialized_val.respond_to?(:as_json)
    
    serialized_val.as_json(*args)
```
to work around the fact that `String` does not always have `as_json` defined on it?

froydnj (dismissed):
It looks like making things unconditionally call `as_json` doesn't work, for reasons previously explained.  Perhaps adopting @.paracycle's suggestion would be sufficient?

gems/sorbet-runtime/lib/types/enum.rb
amomchilov: I'm weary of subtle differences in behaviour depending on whether ActiveSupport is included or not.

`String` won't have `as_json` only when ActiveSupport isn't included. In which case, `as_json` would never really be called anyway.

Perhaps we should only define this method conditionally when ActiveSupport is included? 

Or define the method, but raise an explicit error (`NotImplementedError`?) if it's not included?

cocoahero:
I have updated the PR with @paracycle's proposed solution as well added an additional test to ensure `as_json` is called on the serialized value if possible.

paracycle (approved):

jez (approved):

Issue #5679, opened 2022-04-22
Suggested sig for method on generic class mentions wrong type member


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20X%20%3D%20type_member%0A%0A%20%20sig%20%7Bparams%28x%3A%20X%29.void%7D%0A%20%20def%20foo%28x%29%3B%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20X%20%3D%20type_member%0A%20%20def%20foo%28x%29%3B%20end%0Aend)

```ruby
# typed: strict

class Parent
  extend T::Sig
  extend T::Generic

  X = type_member

  sig {params(x: X).void}
  def foo(x); end
end

class Child < Parent
  X = type_member
  def foo(x); end
end
```

#### Observed output

```
editor.rb:15: This function does not have a sig https://srb.help/7017
    15 |  def foo(x); end
          ^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    editor.rb:15: Insert sig {params(x: Parent::X).void}
    15 |  def foo(x); end
          ^
Errors: 1
```

#### Expected behavior

Note how the suggested signature in the error message shows `Parent::X`. If you attempt to accept the signature, you get this error:

```
editor.rb:15: type_member type Parent::X used outside of the class definition https://srb.help/5027
    15 |  sig {params(x: Parent::X).void}
                         ^^^^^^^^^
    editor.rb:7: Parent::X defined here
     7 |  X = type_member
          ^^^^^^^^^^^^^^^
Errors: 1
```

Sorbet should anticipate this error and only show the last part of the name of the type member in the parent, not the fully qualified name.

- - -

PR #5782, opened 2022-05-11
Correct suggested sig for method on generic class

stefkin:
### Motivation

Fixes https://github.com/sorbet/sorbet/issues/5679

### Test plan

See included automated tests.


jez:
It looks like the same problem exists in single-package RBI generation mode, where we generate an RBI file from Sorbet's view of GlobalState:

image

That's what the `showForRBI` flag is supposed to do.

Instead of introducing a new field on ShowOptions, let's just use `showForRBI`, and pass that option at the call site.

jez (changes requested):
^

froydnj:

→ https://go/builds/bui_LfzRnj2ijJuhrn
→ https://go/builds/bui_LfzREZh0FRthjN

jez (approved):

PR #5784, opened 2022-05-11
Raise error for packaged const used from unpackged

jez:
### Motivation

This gets us to parity with the old packager behavior.

The key insight that makes this easy is that a `# packaged: false` file still technically belongs to the enclosing package (it seems that trevor had maybe overlooked that?)

So e.g., `extn/` can have a `__package.rb` file like

```ruby
class ExtnDONOTUSE < PackageSpec
  import Opus::LSpace
end
```

and all the files in `extn/` can mention e.g. `Opus::LSpace`, but those files are allowed to re-open stdlib classes.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_LfgMlHH7oSKwKo
→ https://go/builds/bui_LfgMGUR79FzdCd

jez:

→ https://go/builds/bui_LfhrxgxfKXWsa7
→ https://go/builds/bui_Lfhr5SLTPXh7GJ

aisamanra (approved):

PR #5790, opened 2022-05-12
Deduplicate serialized_val

casperisfine:
Since the string is frozen already, we can use `String#-@` instead.

This way two enum with similar values will re-use the same interned string.

### DIFF
--- a/gems/sorbet-runtime/lib/types/enum.rb
+++ b/gems/sorbet-runtime/lib/types/enum.rb
@@ -281,7 +281,7 @@ class T::Enum
     # Historical note: We convert to lowercase names because the majority of existing calls to
     # `make_accessible` were arrays of lowercase strings. Doing this conversion allowed for the
     # least amount of repetition in migrated declarations.
-    const_name.to_s.downcase.freeze
+    -const_name.to_s.downcase.freeze
   end
 
   sig {returns(T::Boolean)}


froydnj (changes requested):
I'm not sure I understand the rationale for this.  The explanation is:

> Since the string is frozen already, we can use String#-@ instead.

But `String#downcase` is going to return a copy of the string, even if the string is already all lowercase letters.  That copy is going to be unfrozen.  And then `String#-@` is going to see that the string is not frozen, so it's going to duplicate the string and freeze the duplicate.

The original code just froze the string returned by `#downcase` directly, so it seems like the proposed code is less efficient.

I could see the string being returned from `Symbol#to_s` as being frozen already (I haven't verified that this is the case), so you could verify that value is already all lowercase, and then something like the patch proposed here would be helpful.

Am I missing something?

froydnj:
(I realized that you might be referencing behavior from a different Ruby version; I was working with the 2.7 sources to figure out what `String#downcase` did.  If Ruby 3.x has made `#downcase` smarter here, I can see this being beneficial, but I'm not sure we'd want it for folks who are still on 2.6/2.7.)

casperisfine:
Yes, `-@` if called on a mutable string will first dup to avoid freezing the callee. We can keep the `.freeze` to let Ruby know that it can safely intern that String directlly.

Either way, the idea was that this happens at boot time, so to me reduced memory retention is more important than saving an allocation.

casperisfine:
I added the `.freeze` back, so now there won't be any extra allocation on 2.6+.

froydnj (approved):

PR #5794, opened 2022-05-13
gather more statistics on definition changes that force the slow path in LSP

froydnj:
### Motivation

Stripe's statistics show that something like 50-60% of the reasons for taking the slow path (i.e. typecheck the entire world) in LSP are related to changed definitions.  We'd like to try and understand what sort of changes are triggering this.  Ideally we can get enough information that we can decide the best plan for driving that number down and taking the fast path more often.

### Test plan

See included automated tests.  I have not benchmarked this, but I am guessing that tossing a few more arithmetic operations into `GlobalState::hash` (especially since they could be computed in parallel with `hierarchyHash`, subject to functional units) is not going to have that big of an impact.  Totally willing to revert after we gather some data, too.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2120,12 +2120,19 @@ uint32_t patchHash(uint32_t hash) {
 unique_ptr<GlobalStateHash> GlobalState::hash() const {
     constexpr bool DEBUG_HASHING_TAIL = false;
     uint32_t hierarchyHash = 0;
+    uint32_t classModuleHash = 0;
+    uint32_t typeArgumentHash = 0;
+    uint32_t typeMemberHash = 0;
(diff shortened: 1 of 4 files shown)


jez (approved):
The only thing I'm worried about with this change is the impact it has on max_rss. It looks like we're adding 20 bytes per `core::File`.

froydnj:
> The only thing I'm worried about with this change is the impact it has on max_rss. It looks like we're adding 20 bytes per `core::File`.

max_rss on Stripe's codebase is noisier than 20 bytes/file * 150k files.  I think the max_rss concern is totally valid, but I also think there are better ways of addressing max_rss issues.  And there's also backing this out once we have a sense of what the data is telling us.

PR #5798, opened 2022-05-16
Add field renaming to the LSP

vinistock:
Following up on #4734

Implement instance variable (field) renaming on the LSP. We have symbols associated to fields, so the implementation follows the exact same pattern as the one for classes and methods.

The only thing that is different is checking whether we're renaming the `attr_*` or the instance variable itself (the first not requiring the `@` symbol and the second requiring it). That is,

```ruby
class Person
  attr_reader :age # when renamed, should never have `@` in front

  def initialize
    @age = T.let(1, Integer) # when renamed, should always have `@` in front
  end
end
```

Question:
- I was not 100% sure if class variables are considered fields. Maybe we can add support for them as well in this PR?

### Motivation

To further expand our rename support and improve the editor experience.

### Test plan

See included automated tests.


jez:
> I was not 100% sure if class variables are considered fields

They are modeled as static fields, and you can verify this with `-p symbol-table`:

image

But they might come back as `FieldResponse`'s in LSP, I'm not sure. I'm fine punting on them if the changes in this PR don't work, but I would like to have the test either way.

main/lsp/requests/rename.cc
```diff
@@ -215,6 +215,43 @@ class ConstRenamer : public AbstractRenamer {
...
+        if (!source.has_value()) {
+            return;
+        }
+
+        string newsrc;
+
+        // If the source includes `@` in the first character, then we're renaming the instance variable and need to make
+        // sure to include it. Otherwise, we're renaming an attr_reader/writer/accessor and need to make sure to not
+        // include it. This allows users to not care about whether they need to add the `@` themselves.
+        if (source.value()[0] == '@' && newName[0] != '@') {
```
jez, line 237: Can we exit early if `source.value().empty()`, or use `absl::StartsWith`?

There are a lot of places in Sorbet where we associate zero-width Loc's with pieces of syntax.

jez (changes requested):
Looks mostly good, but some requests.

jez (approved):

PR #5805, opened 2022-05-17
make static field changes take the fast path

froydnj:
### Motivation

Static fields appear to be one of the most common things developers change.  It would be a huge help if changes to such things took the fast path.

The easiest way for me to think about them is that static fields are essentially zero-argument methods that magically get called.  And since changes to the static fields are essentially changes to the method's return type (changes to which already take the fast path, thx to @.jez for the explanation), we should be able to make such changes take the fast path just like methods.

Like #5589, but farther along.  There probably need to be some more tests; I could imagine there are some issues lurking.

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1068,6 +1068,13 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
             [&](parser::Assign *asgn) {
                 auto lhs = node2TreeImpl(dctx, std::move(asgn->lhs));
                 auto rhs = node2TreeImpl(dctx, std::move(asgn->rhs));
+                if (isa_tree<UnresolvedConstantLit>(lhs) && isa_tree<UnresolvedConstantLit>(rhs)) {
+                    auto &rhsConst = cast_tree_nonnull<UnresolvedConstantLit>(rhs);
+                    if (rhsConst.cnst == core::Names::Constants::ErrorNode()) {
+                        auto rhsLocZero = rhs.loc().copyWithZeroLength();
+                        rhs = MK::Let(rhsLocZero, std::move(rhs), MK::Untyped(rhsLocZero));
+                    }
+                }
                 auto res = MK::Assign(loc, std::move(lhs), std::move(rhs));
                 result = std::move(res);
             },
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1068,6 +1068,11 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
             [&](parser::Assign *asgn) {
                 auto lhs = node2TreeImpl(dctx, std::move(asgn->lhs));
                 auto rhs = node2TreeImpl(dctx, std::move(asgn->rhs));
+                // Ensure that X = <ErrorNode> always looks like a proper static field, rather
+                // than a class alias.  Leaving it as a class alias would require taking the
+                // slow path; turning it into a proper static field gives us a chance to take
+                // the fast path.  T.let(<ErrorNode>, T.untyped) ensures that we don't get
+                // warnings in `typed: strict` files.
                 if (isa_tree<UnresolvedConstantLit>(lhs) && isa_tree<UnresolvedConstantLit>(rhs)) {
                     auto &rhsConst = cast_tree_nonnull<UnresolvedConstantLit>(rhs);
                     if (rhsConst.cnst == core::Names::Constants::ErrorNode()) {
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2158,9 +2158,12 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
     for (const auto &field : this->fields) {
         counter++;
         // No fields are ignored in hashing.
+        // TODO(jez) We're abusing methodHashes--after this change, the name is no longer accurate.
+        auto &target = methodHashes[NameHash(*this, field.name)];
         uint32_t symhash = field.hash(*this);
-        hierarchyHash = mix(hierarchyHash, symhash);
+        hierarchyHash = mix(hierarchyHash, field.fieldShapeHash(*this));
         fieldHash = mix(fieldHash, symhash);
+        target = mix(target, symhash);
 
         if (DEBUG_HASHING_TAIL && counter > this->fields.size() - 15) {
             errorQueue->logger.info("Hashing symbols: {}, {}", hierarchyHash, field.name.show(*this));
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2116,6 +2116,7 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
     uint32_t fieldHash = 0;
     uint32_t methodHash = 0;
     UnorderedMap<ShortNameHash, uint32_t> methodHashesMap;
+    UnorderedMap<ShortNameHash, uint32_t> staticFieldHashesMap;
     int counter = 0;
 
     for (const auto &sym : this->classAndModules) {
@@ -2158,12 +2159,13 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
     for (const auto &field : this->fields) {
         counter++;
         // No fields are ignored in hashing.
-        // TODO(jez) We're abusing methodHashes--after this change, the name is no longer accurate.
-        auto &target = methodHashes[NameHash(*this, field.name)];
         uint32_t symhash = field.hash(*this);
+        if (field.flags.isStaticField) {
+            auto &target = staticFieldHashesMap[ShortNameHash(*this, field.name)];
+            target = mix(target, symhash);
+        }
         hierarchyHash = mix(hierarchyHash, field.fieldShapeHash(*this));
         fieldHash = mix(fieldHash, symhash);
-        target = mix(target, symhash);
 
         if (DEBUG_HASHING_TAIL && counter > this->fields.size() - 15) {
             errorQueue->logger.info("Hashing symbols: {}, {}", hierarchyHash, field.name.show(*this));
@@ -2188,11 +2190,16 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
 
     unique_ptr<LocalSymbolTableHashes> result = make_unique<LocalSymbolTableHashes>();
     result->methodHashes.reserve(methodHashesMap.size());
+    result->staticFieldHashes.reserve(staticFieldHashesMap.size());
     for (const auto &[nameHash, symbolHash] : methodHashesMap) {
         result->methodHashes.emplace_back(nameHash, LocalSymbolTableHashes::patchHash(symbolHash));
     }
+    for (const auto &[nameHash, symbolHash] : staticFieldHashesMap) {
+        result->staticFieldHashes.emplace_back(nameHash, LocalSymbolTableHashes::patchHash(symbolHash));
+    }
     // Sort the hashes. Semantically important for quickly diffing hashes.
     fast_sort(result->methodHashes);
+    fast_sort(result->staticFieldHashes);
 
     result->hierarchyHash = LocalSymbolTableHashes::patchHash(hierarchyHash);
     result->classModuleHash = LocalSymbolTableHashes::patchHash(classModuleHash);
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -2302,8 +2302,10 @@ uint32_t Field::fieldShapeHash(const GlobalState &gs) const {
     // literally mention the name of this type alias. Add a test for this.
     // (also maybe double check this logic?)
     auto canSkipType =
-        // All fields are ok (don't participate in constant resolution)
-        this->flags.isField ||
+        // Only consider static fields for the fast path at the moment.  It is probably
+        // straightforward to take the fast path for changes to regular fields by changing
+        // this and the corresponding code in GlobalState, but one step at a time.
+        !this->flags.isField &&
         // Only normal static fields are ok (no class aliases, no type aliases).
         (!this->flags.isStaticFieldTypeAlias && !isa_type<AliasType>(this->resultType));
 
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -2306,7 +2306,7 @@ uint32_t Field::fieldShapeHash(const GlobalState &gs) const {
         // straightforward to take the fast path for changes to regular fields by changing
         // this and the corresponding code in GlobalState, but one step at a time.
         !this->flags.isField &&
-        // Only normal static fields are ok (no class aliases, no type aliases).
+        // Only normal static fields are ok (no type aliases, no class aliases/type templates).
         (!this->flags.isStaticFieldTypeAlias && !isa_type<AliasType>(this->resultType));
 
     if (!canSkipType) {
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -2294,6 +2294,30 @@ uint32_t Method::methodShapeHash(const GlobalState &gs) const {
     return result;
 }
 
+uint32_t Field::fieldShapeHash(const GlobalState &gs) const {
+    uint32_t result = _hash(name.shortName(gs));
+
+    // TODO(jez) type aliases and class aliases participate in the constant resolution fixed
+    // point, which could probably make something start to resolve in a file that doesn't
+    // literally mention the name of this type alias. Add a test for this.
+    // (also maybe double check this logic?)
+    // TODO(jez) There's actually some non-trivial logic in resolver to handle constant fields.
+    // They're no longer handled by resolveSigs; they're handled by ResolveTypeMembersAndFieldsWalk
+    // now. Stare at that code to see if this whole idea is not going to work.
+    auto canSkipType =
+        // All fields are ok (don't participate in constant resolution)
+        this->flags.isField ||
+        // Only normal static fields are ok (no class aliases, no type aliases).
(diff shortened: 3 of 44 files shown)


resolver/resolver.cc
```diff
@@ -2082,10 +2080,9 @@ class ResolveTypeMembersAndFieldsWalk {
                 job.asgn->rhs = ast::MK::Send1(loc, ast::MK::Constant(loc, core::Symbols::Magic()),
                                                core::Names::suggestType(), loc.copyWithZeroLength(), move(rhs));
             }
-            return resultType;
         }
-
-        return data->resultType;
+        // Always return the re-derived type, even on the fast path.
+        return resultType;
```
froydnj, line 2085: This is the only part of t he PR I don't really understand the consequences of: we don't skip work if the symbol already had its type set.  I think this only means that we constantly re-assign `job.asgn->rhs` in some cases, but it seems like these cases should be unusual: static fields are probably not usually defined in multiple files or even multiple times in the same file (?).

(We already had this problem a little bit since `resolveConstantType` got moved out of the `if` several PRs ago, so I think the performance implications are already built-in -- and the unusualness of defining the same static field in multiple places also indicates we usually won't suffer performance problems here, I think?)

ast/desugar/Desugar.cc
```diff
@@ -1068,6 +1068,13 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
             [&](parser::Assign *asgn) {
                 auto lhs = node2TreeImpl(dctx, std::move(asgn->lhs));
                 auto rhs = node2TreeImpl(dctx, std::move(asgn->rhs));
+                if (isa_tree<UnresolvedConstantLit>(lhs) && isa_tree<UnresolvedConstantLit>(rhs)) {
+                    auto &rhsConst = cast_tree_nonnull<UnresolvedConstantLit>(rhs);
+                    if (rhsConst.cnst == core::Names::Constants::ErrorNode()) {
+                        auto rhsLocZero = rhs.loc().copyWithZeroLength();
+                        rhs = MK::Let(rhsLocZero, std::move(rhs), MK::Untyped(rhsLocZero));
+                    }
+                }
```
jez, line 1077: What's going on here?

core/GlobalState.cc
```diff
@@ -2160,7 +2161,11 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
         counter++;
         // No fields are ignored in hashing.
         uint32_t symhash = field.hash(*this);
-        hierarchyHash = mix(hierarchyHash, symhash);
+        if (field.flags.isStaticField) {
+            auto &target = staticFieldHashesMap[NameHash(*this, field.name)];
+            target = mix(target, symhash);
+        }
+        hierarchyHash = mix(hierarchyHash, field.fieldShapeHash(*this));
```
jez, line 2168: Do the changes in this PR work the same for fields like instance variables as well? (Is support for instance variables as easy as making this section unconditional?)

ast/desugar/Desugar.cc
froydnj: This is code that you wrote. :sweat_smile: 

I believe what this is doing is ensuring that static field definitions with a rhs error always wind up in a form that ensures that the lhs is "properly" typed for the fast path in `typed: strict`.  I haven't really experimented with taking this part out, but I could imagine weird things happening if you did?  (OTOH, presumably we have some tests that handle malformed static fields in `typed: strict` files already...?)

core/Symbols.cc
```diff
@@ -2288,6 +2288,30 @@ uint32_t Method::methodShapeHash(const GlobalState &gs) const {
...
+uint32_t Field::fieldShapeHash(const GlobalState &gs) const {
+    uint32_t result = _hash(name.shortName(gs));
+
+    // TODO(jez) type aliases and class aliases participate in the constant resolution fixed
+    // point, which could probably make something start to resolve in a file that doesn't
+    // literally mention the name of this type alias. Add a test for this.
+    // (also maybe double check this logic?)
+    // TODO(jez) There's actually some non-trivial logic in resolver to handle constant fields.
+    // They're no longer handled by resolveSigs; they're handled by ResolveTypeMembersAndFieldsWalk
+    // now. Stare at that code to see if this whole idea is not going to work.
```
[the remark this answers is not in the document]
jez: ah maybe these were present in my original PR?

test/testdata/lsp/fast_path/static_field_type.2.rbupdate
```diff
@@ -0,0 +1,10 @@
+# typed: true
+# assert-fast-path: static_field_type.rb
+
+class A
+  def foo(x); end
+
+  X = ''
+end
+
+T.reveal_type(A::X) # error: `String`
```
jez, line 10: ooc, are these `error:` assertions checked in `*.rbupdate` files? if this said the wrong type would the test fail?

test/testdata/lsp/fast_path/static_field_change_type__use.2.rbupdate
```diff
@@ -0,0 +1,9 @@
+# typed: strict
+# assert-fast-path: static_field_change_type__def.rb,static_field_change_type__use.rb
+extend T::Sig
+
+sig {returns(String)}
+def foo
+  STATIC_FIELD.fetch(:key, "default") + "suffix" # error: Expected `String` but found `T.any(Integer, String)`
+                                      # ^^^^^^^^ error: Expected `Integer`
```
jez, line 8:
```suggestion
  #                                     ^^^^^^^^ error: Expected `Integer`
```

test/testdata/lsp/fast_path/static_field_type.2.rbupdate
froydnj: They are checked!  Makes so many things easy.

test/testdata/lsp/fast_path/static_field_change_type__use.2.rbupdate
jez: I have a slight preference for aligning the comment with the start of the line, otherwise Vim's `=` command and code formatters tend to attempt to re-indent the comments incorrectly.

test/testdata/lsp/fast_path/static_field_change_type__def.1.rbupdate
```diff
@@ -0,0 +1,5 @@
+# typed: strict
+# assert-fast-path: static_field_change_type__def.rb,static_field_change_type__use.rb
+
+STATIC_FIELD = T.let({}, T::Hash[Symbol, ]) # error-with-dupes: Not enough arguments provided
+                               # ^^^^^^ error-with-dupes: Wrong number of type parameters
```
jez, line 5: I have a slight preference for breaking the `error-with-dupes` errors so that there's one `# ^^^ error` comment per dup each on its own line.

Separately, it's annoying that we seem to report so many errors here, even on the slow path:

https://github.com/sorbet/sorbet/issues/5822

test/testdata/lsp/fast_path/static_field_change_type__use.1.rbupdate
```diff
@@ -0,0 +1,8 @@
+# typed: strict
+# assert-fast-path: static_field_change_type__def.rb,static_field_change_type__use.rb
+extend T::Sig
+
+sig {returns(String)}
+def foo
+  STATIC_FIELD.fetch(:key, "default") + "suffix"
```
jez, line 7: non-blocking, but it's worth noting that these `.fetch` calls could start to succeed if somehow we introduced a bug making it seem like `STATIC_FIELD` was `T.untyped` on fast path updates.

Might be worth adding a `T.reveal_type` in these files just to double check?

ast/desugar/Desugar.cc
jez: Yeah I'd be curious to see what fails when this is removed. I don't quite remember

core/Symbols.cc
```diff
@@ -2288,6 +2288,30 @@ uint32_t Method::methodShapeHash(const GlobalState &gs) const {
...
+    auto canSkipType =
+        // All fields are ok (don't participate in constant resolution)
+        this->flags.isField ||
```
jez, line 2303: It already looks like we're handling `isField` in `fieldShapeHash`, and then gating ourselves on `isStaticField` in GlobalState::hash

ast/desugar/Desugar.cc
jez: Ah, I see why this was required now.

In an edit path that looks like

```ruby
A = 1
A = <ErrorNode>
A = ''
```

`A` is always a static field, but sometimes it is a class alias while sometimes it is a "normal" static field.

class aliases are special in that a change to a class alias might make certain things start or stop resolving, which we currently can't check except by re-running the slow path. This `T.let` makes it so that `A = <ErrorNode>` never looks like a class alias.

This *might* bite us if/when we attempt to get all changes to constants on the fast path, but for now this seems like an improvement.

The two alternatives here would be either

```ruby
A = T.let(<ErrorNode>, T.untyped)
A = T.unsafe(<ErrorNode>)
```

the `T.let` is nice because it means there's not an extra "constants must have type annotations in `# typed: strict` files" but it's also something I'm usually reluctant to do because adding `T.let` in various places (especially inside method bodies) has an annoying tendency to cause confusing pinning errors that the user can't resolve.

Since this is a static field not a variable it's probably fine to leave it at `T.let`. If you'd like to take a stab at writing a brief comment to this effect in the file, that'd be great. If you'd prefer that I craft some sort of comment here lmk.

test/testdata/namer/fuzz_type_member.rb
```diff
@@ -1,4 +1,4 @@
 # typed: true
 
-A=3 # error: Expected `Runtime object representing type: T.untyped` but found `Integer(3)` for field
+A=3 # error: Expected `Runtime object representing type: Integer` but found `Integer(3)` for field
```
jez, line 3: Is this due to the change to resolver?

test/testdata/namer/fuzz_type_template_overwrite.rb.symbol-table-raw.exp
```diff
@@ -3,7 +3,7 @@ class <C <U <root>>> < <C <U Object>> ()
     method <S <C <U <root>>> $1>#<N <U <static-init>> $CENSORED> (<blk>) @ Loc {file=test/testdata/namer/fuzz_type_template_overwrite.rb start=3:1 end=6:4}
       argument <blk><block> @ Loc {file=test/testdata/namer/fuzz_type_template_overwrite.rb start=??? end=???}
   class <C <U A>> < <C <U Object>> () @ Loc {file=test/testdata/namer/fuzz_type_template_overwrite.rb start=3:1 end=3:8}
-    static-field <C <U A>>::<C <U B>> -> AliasType { symbol = <S <C <U A>> $1>::<C <U B>> } @ Loc {file=test/testdata/namer/fuzz_type_template_overwrite.rb start=5:3 end=5:20}
+    static-field <C <U A>>::<C <U B>> -> Integer @ Loc {file=test/testdata/namer/fuzz_type_template_overwrite.rb start=5:3 end=5:20}
```
jez, line 6: I guess this is also due to the change to resolver?

core/GlobalState.cc
jez: fwiw, it would be really cool if we could get instance variable working, because it would mean that changes to types of `prop`'s and `attr_*`'s would take the fast path.

main/lsp/LSPTypechecker.cc
```diff
@@ -299,13 +312,21 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
             subset.emplace_back(ref);
+            continue;
+        }
+
+        absl::c_set_intersection(changedFieldHashes, oldHash.usages.symbols, std::back_inserter(intersection));
+        if (!intersection.empty()) {
+            auto ref = core::FileRef(i);
+            config->logger->debug("Added {} to update set as used a changed field symbol",
+                                  !ref.exists() ? "" : ref.data(*gs).path());
+            subset.emplace_back(ref);
```
jez, line 329: 👨‍🍳 💋 🤌

core/Symbols.cc
```diff
@@ -2288,6 +2288,30 @@ uint32_t Method::methodShapeHash(const GlobalState &gs) const {
     return result;
 }
 
+uint32_t Field::fieldShapeHash(const GlobalState &gs) const {
+    uint32_t result = _hash(name.shortName(gs));
+
+    // TODO(jez) type aliases and class aliases participate in the constant resolution fixed
+    // point, which could probably make something start to resolve in a file that doesn't
+    // literally mention the name of this type alias. Add a test for this.
+    // (also maybe double check this logic?)
```
jez, line 2297: Could we add a test like

```ruby
# typed: true

class Parent
  module Inner; end
end

ParentAlias = 1

class A < ParentAlias
  puts(Inner)
end
```

which gets updated to

```ruby
# typed: true

class Parent
  module Inner; end
end

ParentAlias = Parent

class A < ParentAlias
  puts(Inner)
end
```

And assert that it works?

We could even get creative and make it so that `class A` depends on the class alias in a way that the name of `ParentAlias` doesn't appear in the file that has the errors:

```ruby
# typed: true

class Parent
  module Inner; end
end

ParentAlias = 1
# ParentAlias = Parent

Thing = ParentAlias

# -- other.rb --

class A < Thing
  puts(Inner)
end
```

core/Symbols.cc
jez, line 2300:
```suggestion
```

I think we can just delete this comment and move the discussion to resolver.cc where you already commented.

core/Symbols.cc
jez: (and also delete this comment once such tests are added)

resolver/resolver.cc
```diff
@@ -2056,8 +2056,6 @@ class ResolveTypeMembersAndFieldsWalk {
                 return resultType;
             }
         }
-        // resultType was already set. We may be running on the incremental path. Force this field to be resolved in
-        // `resolveStaticField`, which may produce an error message.
```
jez, line 2060: Did I delete this comment? 🤔

It seems like it still applies...

jez:
I found a bug:

```ruby
# typed: true

class MyEnum < T::Enum
  enums do
    X = new
    Y = new
  end
end

MyEnum_X = MyEnum::X

Alias = T.let(MyEnum::X, MyEnum)
```

edit the last line by inserting `_X` before the final `)`.

On master right now, that edit forces the slow path, and reports an error "Constant `MyEnum_X` is not a class or type alias".

On this branch, that edit takes the fast path and reports no error.

resolver/resolver.cc
jez: Yeah i'm not really sure.

The only thing I can think of here is to get really creative with a lot of fast-path tests.

I found one failing test for you, which should give you something to think about for the time being.

If you wouldn't mind trying to write a bunch of weird and wacky test cases after you've looked into that, we can revisit these resolver changes.

Suggestions for crafting test cases would be: think of weird things. Use lots of different kinds of symbols. Use weird features like `T::Enum`, type members, class aliases, type aliases, etc. Try to find errors that are more than just the result type (the error below is a resolver/type_syntax.cc error).

jez (changes requested):
Going to mark this as request changes for now, because i found a failing test case. We can revisit next week.

froydnj:
Thanks for the interesting testcase.  This is a problem for static fields taking the fast path:

- The definitions didn't actually change, so we can't possibly have observed a change to the way symbols are hashed; but
- The way those definitions are used in the AST did change, and changed in ways that we don't allow.

froydnj:
> Thanks for the interesting testcase. This is a problem for static fields taking the fast path:
> 
>     * The definitions didn't actually change, so we can't possibly have observed a change to the way symbols are hashed; but
> 
>     * The way those definitions are used in the AST did change, and changed in ways that we don't allow.

OK, this objection is nonsense, because I didn't understand what I was talking about.  The real problem appears to be that we wind up in a situation where:

1. We give `Alias` the expected type of `MyEnum`
2. We have the file under that assumption and everything is fine
3. File typechecks
4. We change Alias to point at `MyEnum_X`
5. We rehash the file, except now we have a (silenced) error during the hashing process
6. `Alias` in the hashed `GlobalState` looks like it hasn't changed its shape, but its `resultType` *has* changed (or would have changed if we were running the full resolver), because of the error
7. We wind up taking the fast path, even though we really shouldn't

It's not yet obvious to me how we wind up correctly typechecking the file, unless we somehow can resolve `MyEnum_X` earlier than we would have otherwise (since it's now in the symbol table, as a pre-existing thing), and that means we see a class/type alias in the `T.let`, which is totally permitted.  I guess on the fast path we could force constants to not resolve immediately (?)...but I don't know what that does to the fast-ness of the fast path, or what other knock-on effects that would have.

Step 6 looks like a real problem for correctly determining hashes, though.

test/testdata/namer/fuzz_type_member.rb
froydnj: Yes.

test/testdata/namer/fuzz_type_template_overwrite.rb.symbol-table-raw.exp
froydnj: Yes.

jez (approved):
Thanks for writing some more tests.

Have you had a chance to test this out in Stripe's codebase in VS Code?

I'm leaning towards this being read to ship, and if it's buggy we just deal with the bugs afterwards. Opening up a few files in Stripe's codebase and making an assortment of edits would at least clue us in on any remaining obvious bugs that people would encounter in the moments after we land the upgrade. Any non-obvious bugs we can deal with later.

froydnj:

→ https://go/builds/bui_LnvRG0XZnsQwu4
→ https://go/builds/bui_LnvRCxxZbyPan3

PR #5806, opened 2022-05-17
split out changed definition statistics when only a single category changed

froydnj:
### Motivation

I think the single category case is interesting, and even if it winds up not interesting, I think we get useful information from knowing "several things changed at once", which probably implies some sort of weird parse result that we could investigate.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPIndexer.cc
+++ b/main/lsp/LSPIndexer.cc
@@ -148,6 +148,21 @@ bool LSPIndexer::canTakeFastPathInternal(
             const bool methodsDiffer = newHash.definitions.methodHash != oldHash.definitions.methodHash;
             const uint32_t differCount = int(classesDiffer) + int(typeArgumentsDiffer) + int(typeMembersDiffer) +
                                          int(fieldsDiffer) + int(methodsDiffer);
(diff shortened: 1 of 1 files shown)


main/lsp/LSPIndexer.cc
```diff
@@ -140,20 +140,43 @@ bool LSPIndexer::canTakeFastPathInternal(
...
+                }
+                if (typeMembersDiffer) {
+                    prodCategoryCounterInc("lsp.slow_path_changed_def", "typemember");
+                }
+                if (fieldsDiffer) {
+                    prodCategoryCounterInc("lsp.slow_path_changed_def", "field");
+                }
+                if (methodsDiffer) {
+                    prodCategoryCounterInc("lsp.slow_path_changed_def", "method");
+                }
```
jez, line 179: Are you sure this is what you want? Don't you want to always report the `differCount != 1` case, and then additionally report the others?

It seems like if you're ever looking at the `diffCount != 1` metrics, you don't want to have to remember to add the `only*` counters back in.

jez (approved):

PR #5809, opened 2022-05-18
Add an experimental error to catch "common prefix" collisions

aadi-stripe:
Adds a new error that is reported when there is a "common prefix" collision. Currently performs the check only when running with `--stripe-mode` and `--stripe-mode-namespace-collision-check-experimental` flags.

A "namespace" or "common prefix" in this context is defined as the lowest common prefix of
all behavior-defining (see error 4019) classes/modules in a file. This error is reported when 2 files have the
same lowest common prefix.

```
# -- file1.rb --
module A::B
  class C
    def foo; end
  end

  class D
    def bar; end
  end
end

# -- file2.rb --
module A::B
  def bar; end
end
```

In this case, both files have the same lowest common prefix A::B.

### Motivation

In the Stripe codebase, we are exploring migrating to an implicit path-based autoloading solution like [zeitwerk](https://github.com/fxn/zeitwerk). One requirement is that every file maps to a unique constant to autoload, which we would like to eagerly catch violations of and fix them in the codebase.

Stripe employees can see the results of this error at https://go/gist/aadi/d404cec430493748621032e31428e5b6

### Test plan

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1442,6 +1442,8 @@ class TreeSymbolizer {
 
     bool bestEffort;
 
+    core::ClassOrModuleRef commonPrefix;
+
     core::SymbolRef squashNamesInner(core::Context ctx, core::SymbolRef owner, ast::ExpressionPtr &node,
                                      bool firstName) {
         auto constLit = ast::cast_tree<ast::UnresolvedConstantLit>(node);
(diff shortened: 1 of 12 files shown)


aadi-stripe:

→ https://go/builds/pay-server-ruby-typer--bui_LiQn87JpZ7vTSg
→ https://go/builds/pay-server-sorbet-check--bui_LiQnP0kGkC2Oq9

namer/namer.cc
```diff
@@ -2084,7 +2154,13 @@ vector<ast::ParsedFile> symbolizeTrees(const core::GlobalState &gs, vector<ast::
         fileq->push(move(tree), 1);
     }
 
-    workers.multiplexJob("symbolizeTrees", [&gs, fileq, resultq, bestEffort]() {
+    // Store a map of common prefix -> file where that prefix is seen.
+    UnorderedMap<core::ClassOrModuleRef, core::FileRef> commonPrefixes;
+
+    // Mutex for safe concurrent updates of map.
+    std::mutex prefixMutex;
```
jez, line 2161: whoa whoa this does not look good. Can we get this working any other way?

putting a mutex here in an otherwise parallel loop is going to throttle our performance and is a code smell

aadi-stripe: Sure, yep, will fix.

PR #5810, opened 2022-05-18
don't call `Sorbet::Private::Static.keep_def` in non-compiled files

froydnj:
### Motivation

We don't need the extra nodes in the AST in non-compiled files.  Ideally we can also do the same thing for `<Magic>.<define-top-class-or-module>` under the same conditions.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/pay-server-ruby-typer--bui_LiEk5CQynAHwtI
→ https://go/builds/pay-server-sorbet-check--bui_LiEkGTSptjcpv4

froydnj:

→ https://go/builds/pay-server-ruby-typer--bui_LianB6BNokLtvS
→ https://go/builds/pay-server-sorbet-check--bui_LianBxpeaKl1U3

test/cli/package-prefix-enforcement/test.out
```diff
@@ -27,10 +27,6 @@ nested/nested.rb:38: File belongs to package `Root::Nested` but defines a consta
      3 |class Root < PackageSpec
         ^^^^^^^^^^^^^^^^^^^^^^^^
 
-nested/nested.rb:41: Class or method behavior may not be defined outside of the enclosing package namespace `Root::Nested` https://srb.help/3713
-    41 |  def self.method
-          ^^^^^^^^^^^^^^^
```
froydnj, line 32: We already issued an error for this several lines above, so not having this error is an improvement.  A previous commit fixed `test/testdata/packager` tests (which also had duplicate errors prior to this PR), but I forgot to run the CLI tests locally prior to doing that commit.

ast/Trees.h
```diff
@@ -1111,6 +1112,25 @@ EXPRESSION(InsSeq) {
 };
 CheckSize(InsSeq, 56, 8);
 
+EXPRESSION(RuntimeMethodDefinition) {
+public:
+    // This should be equivalent to MethodDef.declLoc.
```
jez, line 1117:
```suggestion
    // This should be equivalent to MethodDef.declLoc.
    // It's called `loc` because every node in the AST has to have a `loc` field.
```

jez (approved):

PR #5813, opened 2022-05-19
Assorted renames and comments

jez:
### Motivation

I had been making some changes that turned out to conflict with the changes in
a541b70f1. (The changes I was making were in service of allowing changes to
method symbols to take the fast path.)

I've broken out some changes from that PR here so to make it easier to review
the fast path methods PR.

### Commit summary

Review by commit.

- **wip: Put everything back** (71075879c)

- **fastmoda 'GlobalStateHash' 'LocalSymbolTableHashes'** (9c307bf69)

- **definitions -> localSymbolTableHashes** (1d345f980)

- **NameHash.{h,cc} -> FileHash.{h,cc}** (1d5201a40)

  Eveyrthing in this file builds up to the `FileHash` structure, which is 
  more important than just the `NameHash` structure at the top.

  Also, after the previous change, it doesn't even have anything to do 
  with `NameRef` specifically anymore.

- **Move patchHash to LocalSymbolTableHashes** (84facbac7)

  Colocate with the things that actually care about these magic constants.

- **Factor out isInvalid function** (23e53d5ca)

- **HASH_STATE_INVALID -> HASH_STATE_INVALID_PARSE** (8bcfc2acc)

  This is currently only used as a flag that parsing failed, indicating 
  that the hierarchyHash is meaningless.

- **Add some comments** (2734f3600)

- **methodHashes -> methodHashesMap** (a408ad186)

  We have two things with the same name in the same function but with 
  different types, and I thought that was confusing, so I renamed one.

- **Replace pair with SymbolHash** (c9c4d9b7e)

- **Add some more comments** (4f3cdf492)

### Test plan

There is no functional change in this PR, so it is covered by existing tests.


froydnj (approved):
I like the cleanup of moving `NameHash.*` to `FileHash.*`!

PR #5816, opened 2022-05-19
wip: Intrinsic for `Kernel#nil?`

### LINKED ISSUE #3769: Calling `nil?` on non-nilable value in a conditional doesn't report dead code

jez:
### Motivation

Fixes #3769

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3658,6 +3658,16 @@ public:
     }
 } Kernel_proc;
 
+class Kernel_nil_p : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        auto intersection = Types::all(gs, args.selfType, core::Types::nilClass());
+        if (intersection.isBottom()) {
+            res.returnType = core::Types::falseClass();
+        }
+    }
+} Kernel_nil_p;
+
 class Enumerable_toH : public IntrinsicMethod {
 public:
     // Forward Enumerable.to_h to RubyType.enumerable_to_h[self]
@@ -3888,6 +3898,7 @@ const vector<Intrinsic> intrinsics{
 
     {Symbols::Kernel(), Intrinsic::Kind::Instance, Names::proc(), &Kernel_proc},
     {Symbols::Kernel(), Intrinsic::Kind::Instance, Names::lambda(), &Kernel_proc},
+    {Symbols::Kernel(), Intrinsic::Kind::Instance, Names::nil_p(), &Kernel_nil_p},
 
     {Symbols::Enumerable(), Intrinsic::Kind::Instance, Names::toH(), &Enumerable_toH},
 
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/pay-server-ruby-typer--bui_LifGZtM3L0WDYL
→ https://go/builds/pay-server-sorbet-check--bui_LifGLYq0Fu7vrh

froydnj (approved):

jez:
This causes about 700 errors on Stripe's codebase.

I haven't taken much time to figure out whether that's because there's a bug in my implementation, or whether there's actually that many violations that were hidden before.

vinistock (approved):
🚀

froydnj:
> This causes about 700 errors on Stripe's codebase.

Looking through the first several, they all look like "real" errors in the sense that `nil?` is being called on things that have been statically declared to be non-nilable things.  What's not clear to me is whether those `nil?` calls have been placed there defensively, e.g. props that were declared to be `String` but sometimes (?) are `nil`.

As a sort of gross migration strategy, would we consider doing something like:

1. In the intrinsic, when we are going to return bottom, issue a separate error about calling `nil?` on non-nilable things.
2. Once we have issued that error, return a boolean-typed thing so that we don't get further errors.
3. Silence the error, either by default or solely on Stripe's codebase.
4. Fix the errors at some later point.

Obviously whether (4) happens in any reasonable timeframe is unclear.

This rollout strategy would (assuming we only silenced the errors on Stripe's codebase) at least let non-Stripe Sorbet users benefit from this change.  I think the error might even be clearer than the current error, because we can point out directly that `nil?` is being called on some non-nilable thing, rather than indicating that the entire send produces `FalseClass`, which requires some mental processing to conclude that the receiver is always non-`nil`.

A hand-wavy blue-sky autofix that replaced the send with `true`, combined with some other rubocop-esque mechanism to eliminate useless conditionals (dropping `&& true`, making `... if true` unconditional, etc.), might even be able to automate the process of fixing these errors.

jez:
Closing this for the time being because I don't have time to fix these type errors in the near future, and I can't think of a good way to build autocorrect.

(If you can think of one, please feel free to re-submit another PR with these changes and the autocorrect.)

PR #5820, opened 2022-05-19
Approximate fix for static field autocorrect suggestions

### LINKED ISSUE #5819: Autocorrect to fix pinning error suggests `T.let` around static field definition
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20MyEnum%20%3C%20T%3A%3AEnum%0A%20%20enums%20do%0A%20%20%20%20X%20%3D%20new%0A%20%20end%0Aend%0A%0Ax%20%3D%20MyEnum%3A%3AX%0A1.times%20do%0A%20%20x%20%3D%20false%0Aend)

```ruby
# typed: true

StaticField = T.let(0, Integer)
x = StaticField
1.times do
  x = false
end
```

#### Observed output

```
editor.rb:6: Changing the type of a variable in a loop is not permitted https://srb.help/7001
     6 |  x = false
              ^^^^^
  Existing variable has type: Integer
  Attempting to change type to: FalseClass

    editor.rb:3: Replace with T.let(StaticField, T.any(FalseClass, Integer))
     3 |StaticField = T.let(0, Integer)
        ^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

Sorbet should suggest wrapping the initialization of `x` on line 4, not the `StaticField` on line 3.

- - -

jez:
### Motivation

Fixes #5819

### Test plan

See included automated tests.

### DIFF
--- a/cfg/CFG.h
+++ b/cfg/CFG.h
@@ -107,7 +107,11 @@ public:
      */
     int numLocalVariables() const;
 
-    core::FileRef file;
+    // Stores MethodDef::loc (MethodDef::declLoc).
+    //
+    // Thus, this usually spans the full body of the method def (recall that sometime methods are
+    // created synthetically, not from method defs written in the source).
+    core::LocOffsets loc;
     std::vector<std::unique_ptr<BasicBlock>> basicBlocks;
 
     // Special loc that corresponds to implicit method return.
--- a/cfg/CFG.h
+++ b/cfg/CFG.h
@@ -107,7 +107,7 @@ public:
      */
     int numLocalVariables() const;
 
-    // Stores MethodDef::loc (MethodDef::declLoc).
+    // Stores MethodDef::loc (not the MethodDef::declLoc).
     //
     // Thus, this usually spans the full body of the method def (recall that sometime methods are
     // created synthetically, not from method defs written in the source).
--- a/test/testdata/infer/pinning_static_field.rb.autocorrects.exp
+++ b/test/testdata/infer/pinning_static_field.rb.autocorrects.exp
@@ -1,0 +1,23 @@
+# -- test/testdata/infer/pinning_static_field.rb --
+# typed: true
+
+T.let(StaticField1, T.any(FalseClass, Integer)) = T.let(0, Integer)
+def example1
+  x = StaticField1
+  1.times do
+    x = false
+  end
+end
+
+StaticField2 = T.let(0, Integer)
+def example2
+  if T.unsafe(nil)
+    x = StaticField2
+  else
+    x = 0
+  end
+  1.times do
+    x = false
+  end
+end
+# ------------------------------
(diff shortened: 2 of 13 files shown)


test/testdata/infer/pinning_static_field.rb.autocorrects.exp
```diff
@@ -0,0 +1,23 @@
+# -- test/testdata/infer/pinning_static_field.rb --
+# typed: true
+
+StaticField1 = T.let(0, Integer)
+def example1
+  x = StaticField1
```
jez, line 6: This "fixes" the change by not suggesting an autocorrect for the pinning error.

I didn't look into how hard it would be to change the location information, but one thing is that right now the TypeAndOrigins information in e.g. "expected/got" errors will show that it got `Integer` because of the type annotation on the static field.

Maybe that's fine to give up? When a variable gets its type from

```ruby
x = foo
```

we don't show the sig for `foo` in the "expected/got" error message.

Anyways, I figure that can be a change for another day if someone wants to look into it. At least this makes it so that `-a` doesn't suggest syntax errors.

cfg/CFG.h
```diff
@@ -107,7 +107,11 @@ class CFG final {
      */
     int numLocalVariables() const;
 
-    core::FileRef file;
+    // Stores MethodDef::loc (MethodDef::declLoc).
```
froydnj, line 110: This comment reads to me like "Stores MethodDef::loc (well, actually, MethodDef::declLoc)".  The two are not really the same thing, so why mention both of them in this comment?

froydnj (approved):
It's kind of unfortunate that we lost the location information indicating the textual `StaticField` in the assignment and instead redirect to the location information for the definition of the static field.  It feels like we should not necessarily be making that assumption...but I guess we don't have the def-use chains that we would need to easily trace back to the definition we need.

cfg/CFG.h
jez: sorry looks like in an edit i deleted "not the ..." from the parenthesis somehow

jez:
> It's kind of unfortunate that we lost the location information indicating the textual `StaticField` in the assignment and instead redirect to the location information for the definition of the static field

When you say "lost," you don't mean that this PR lost it, do you?

If you mean it in the sense that our CFG isn't good enough to keep track of it, I'm not sure that's true, in fact I think it's likely that we could improve this in the future but I just have opted not to spend a lot of time getting something working.

froydnj:
> > It's kind of unfortunate that we lost the location information indicating the textual `StaticField` in the assignment and instead redirect to the location information for the definition of the static field
> 
> When you say "lost," you don't mean that this PR lost it, do you?
> 
> If you mean it in the sense that our CFG isn't good enough to keep track of it, I'm not sure that's true, in fact I think it's likely that we could improve this in the future but I just have opted not to spend a lot of time getting something working.

Oh, no, this PR didn't lose anything.  I'm 95% certain the necessary location information is in the CFG already, it's just a matter of properly extracting it.  (Or e.g. updating `TypeAndOrigins` to track more information in cases like these, updating callsites, etc.)

PR #5823, opened 2022-05-21
remove the need for `Sorbet::Private::Static.keep_for_typechecking`

froydnj:
### Motivation

When we encounter a `T.let($VALUE, $TYPE)` call or similar during the resolver phase, we do this weird thing:

1. We create a `Sorbet::Private::Static.keep_for_typechecking($TYPE)` send
2. We create an `ast::Cast` node with the kind of cast (`let`, `cast`, etc.) and `$VALUE` -- but not `$TYPE`
3. We create a resolver job that points at the `$TYPE` arg held by the newly-created send and the newly-created cast node.
4. We replace the `T.let` call with the newly-created cast node.

Why do we do this?  Well, we need to have some way of entering bits of `$TYPE` into the AST/CFG for LSP to work properly, since LSP queries are going to scan through the AST/CFG to find e.g. constant references.  If we only kept `$TYPE` to resolve it to a `TypePtr` during the resolver phase, then things like "find all references" wouldn't see constant references in `$TYPE`.  (We threw away the original `T.let` call and replaced it with an `ast::Cast` and the `ast::Cast` doesn't have references to `$TYPE` as an expression.)

The problem with this scheme is that we're creating this weird send node that exists nowhere in the user's code.  This wastes space in the AST, it wastes space and time expanding the send into CFG nodes (receiver, argument, and the send itself), and it wastes time deriving the type of the send.

This PR, then, proposes to store the expression for `$TYPE` on the `ast::Cast` node itself.  This change expands the cast node by 8 bytes, but as jemalloc buckets 40 byte things into bucket of 48 byte wide things, we're actually not consuming any more memory as a result of this change, cast node-wise.  This will also avoid creating extraneous send nodes and the awkward issues of location information and extra CFG cost associated with them.

But we still have to ensure these nodes appear in the CFG for the LSP reasons stated above.  (It's not clear to me if we could somehow get away with just expanding the constant references in the `$TYPE` expression into the CFG, so that we could avoid the sends in `T.nilable(T.any(Foo, Bar))`, which seems like a very nice win, since we don't care about their types either, but that is a question for another time.)

The approach I initially tried creates a new CFG instruction, `VolatileRead` (the name is bikesheddable), to ensure that whatever was being read by the `VolatileRead` instruction wouldn't be considered dead code, as `$TYPE` in some cases can be side-effect free.  The expansion of the `keep_for_typechecking` call naturally assured this, since we don't delete sends from the CFG (they might have side-effects), and then it follows that we can't delete their arguments either.  But now we have no send, so we have to find a different mechanism.

After playing around with `VolatileRead`, it occurred to me that the dead code elimination pass done by the CFG is not strong enough to determine that `a = a`, where `a` is never read from again, is dead code.  So we could get by without using the `VolatileRead` instruction at the cost of enshrining this weird special case in the CFG.  It might be kind of nice to have a separate CFG node so that the compiler could still delete extraneous code:

https://github.com/sorbet/sorbet/blob/dbe102f017ad2fa371e27f5ec5a5f9a928eb7d18/plugin_injector/plugin_injector.cc#L139-L177

but maybe it could just be taught to recognize the `a = a` pattern instead...though that might have the undesired effect of doing strange things with user code.  Need to think about that some more.

I haven't tried this on Stripe's codebase; I expect it to have a small but measurable effect in memory use and typechecking performance.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_LqYbcSy0MwtEvK
→ https://go/builds/bui_LqYbQb7v9QBAkj

froydnj:

→ https://go/builds/bui_LvRkXRnHM31qSl
→ https://go/builds/bui_LvRkRzYd61tGNU

froydnj:

→ https://go/builds/bui_LyhTtq3gPeuhKr
→ https://go/builds/bui_LyhT3RGNE2r1bB

froydnj:
I think this is more-or-less ready for review.  It should be relatively neutral performance-wise, but I think having everything tied into a single AST node, rather than scattered throughout the AST, makes things more understandable.

I also have some in-progress patches that want to move (part of) field identification into namer; as a part of that, the patches identify casts much earlier in the typechecking pipeline so both namer and resolver can see the cast nodes, rather than resolver creating them.  Having everything stored on a single node is beneficial there as well.

> This PR, then, proposes to store the expression for `$TYPE` on the `ast::Cast` node itself. This change expands the cast node by 8 bytes, but as jemalloc buckets 40 byte things into bucket of 48 byte wide things, we're actually not consuming any more memory as a result of this change, cast node-wise.

This part of the original description is no longer true now that `TypePtr` only occupies 8 bytes.  We wind up enlarging cast nodes slightly, but we're eliminating some send nodes from the CFG (and therefore typechecking), so overall memory usage might go up slightly, but typechecking performance should slightly improve.

cfg/Instructions.h
```diff
@@ -47,6 +47,7 @@ enum class Tag : uint8_t {
     YieldLoadArg,
     Cast,
     TAbsurd,
+    VolatileRead,
```
jez, line 50: I don't know what other IRs call similar instructions, but what do you think about `KeepAlive` or something like that?

ast/Trees.h
```diff
@@ -954,7 +954,9 @@ EXPRESSION(Cast) {
     core::TypePtr type;
     ExpressionPtr arg;
 
-    Cast(core::LocOffsets loc, core::TypePtr ty, ExpressionPtr arg, core::NameRef cast);
+    ExpressionPtr typeExpr;
```
jez, line 957: This means that we have a `Loc` for the `typeExpr` now. Does that help make any error messages/autocorrects better now?

ast/Trees.cc
```diff
@@ -1108,8 +1108,13 @@ ExpressionPtr Send::withNewBody(core::LocOffsets loc, ExpressionPtr recv, core::
...
     fmt::memory_buffer buf;
     fmt::format_to(std::back_inserter(buf), "T.{}", this->cast.toString(gs));
-    fmt::format_to(std::back_inserter(buf), "({}, {})", this->arg.toStringWithTabs(gs, tabs),
-                   this->type.toStringWithTabs(gs, tabs));
+    fmt::format_to(std::back_inserter(buf), "({}, ", this->arg.toStringWithTabs(gs, tabs));
+    if (this->type != core::Types::todo()) {
+        fmt::format_to(std::back_inserter(buf), "{}", this->type.toStringWithTabs(gs, tabs));
+    } else {
+        fmt::format_to(std::back_inserter(buf), "{}", this->typeExpr.toStringWithTabs(gs, tabs));
+    }
```
jez, line 1116: ooc, should we keep it showing `todo` anyways?

or change the printing to show all three fields?

I can imagine getting super confused trying to debug something by seeing `T.let(x, Integer)` in two different snapshots and not realizing that they're subtly different because one of them is a constant lit and the other one is a TypePtr.

infer/environment.cc
```diff
@@ -1484,6 +1484,10 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                 tp.type = core::Types::bottom();
                 tp.origins.emplace_back(ctx.locAt(bind.loc));
             },
+            [&](cfg::VolatileRead &i) {
+                tp.type = core::Types::untypedUntracked();
+                tp.origins.emplace_back(ctx.locAt(bind.loc));
```
jez, line 1489: I'm not sure where this would matter, but should we make `tp` be equal to whatever the type and origins of `i.what` is?

resolver/resolver.cc
```diff
@@ -2835,24 +2834,24 @@ class ResolveTypeMembersAndFieldsWalk {
...
                     item.cast = ast::cast_tree<ast::Cast>(cast);
-                    item.typeArg = &ast::cast_tree_nonnull<ast::Send>(typeExpr).getPosArg(0);
 
                     // We should be able to resolve simple casts immediately.
                     if (!tryResolveSimpleClassCastItem(ctx.state, item)) {
                         todoResolveCastItems_.emplace_back(move(item));
                     }
 
-                    tree = ast::MK::InsSeq1(send.loc, move(typeExpr), move(cast));
+                    tree = std::move(cast);
```
jez, line 2854: 👍 nice

test/testdata/infer/fields.rb.cfg-text.exp
```diff
@@ -20,11 +20,11 @@ method ::Foo#initialize {
...
     @ivar$3: Integer = alias @ivar
     <self>: Foo = cast(<self>: NilClass, Foo);
-    <cfgAlias>$6: T.class_of(Sorbet::Private::Static) = alias <C Static>
-    <cfgAlias>$8: T.class_of(Integer) = alias <C Integer>
-    <statTemp>$4: Sorbet::Private::Static::Void = <cfgAlias>$6: T.class_of(Sorbet::Private::Static).keep_for_typechecking(<cfgAlias>$8: T.class_of(Integer))
-    <castTemp>$9: Integer(0) = 0
-    @ivar$3: Integer = cast(<castTemp>$9: Integer(0), Integer);
+    <cfgAlias>$5: T.class_of(Integer) = alias <C Integer>
+    keep_for_ide$4: T.class_of(Integer) = <cfgAlias>$5
+    keep_for_ide$4: T.untyped = <volatile-read> keep_for_ide$4
```
jez, line 25: related to the comment below, it looks like we're re-assigning this variable and squashing the type?

jez (approved):
mostly lgtm

ast/Trees.cc
froydnj: I was trying to avoid having a bunch of `T.let(x, <todo>)` in the changes for exp files (e.g. in rewriter or namer) in later branches, which didn't seem super-helpful.  But I can see confusion arising if you see the same thing being printed at two different times and there are actually two different underlying things (e.g. debugging resolver or something).

My opinion is that if we're going to print three things, the printing should be changed to make it obviously not a send, something like `<cast:let>(x, $TYPE, $TYPE_EXPR)`.  WDYT?

jez: wfm

ast/Trees.h
froydnj: I don't see anything obvious in handling `Cast` in `environment.cc` (assuming we propagated the extra information into `cfg::Cast` nodes), but maybe there are other places.

cfg/Instructions.h
froydnj: I was going on `volatile` from C/C++ (particularly as it applies to inline assembly), which to a first approximation means "you can't assume anything about what happens with this", but `KeepAlive` wfm.

PR #5825, opened 2022-05-23
implement swap for `ExpressionPtr` and related structures

froydnj:
### Motivation

Trying to make sorting everything all the time faster.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -221,6 +221,11 @@ public:
     std::string toString(const core::GlobalState &gs) const {
         return toStringWithTabs(gs);
     }
+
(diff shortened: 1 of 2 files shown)


froydnj:

→ https://go/builds/bui_LrzaZyETcn1jzN
→ https://go/builds/bui_LrzaJ9x7USAhtx

froydnj:
This does reduce code size a small amount and makes things slightly faster: about 4% faster for package visibility checking on Stripe's codebase.  I think most other usages of sorts are grouped with timers that have too much noise, but eliminating all the branching has to be good for you.

jvilk-stripe (approved):

PR #5826, opened 2022-05-23
Normalize for leading `./` in autogen constant cache

aisamanra:
### Motivation
Previously we'd be kind of finicky about the presence of leading `./` in the constant cache. This will now always strip it out so it doesn't matter if we use it anywhere in particular.

### Test plan

See included automated tests. Added tests to ensure that lookup doesn't care about `./` and the we produce the same msgpack file regardless of leading `./`.

### DIFF
--- a/main/autogen/cache.cc
+++ b/main/autogen/cache.cc
@@ -13,6 +13,13 @@ namespace sorbet::autogen {
 
 const size_t MAX_SKIP_AMOUNT = 100;
 
+string_view trimPath(string_view path) {
+    if (path.size() > 2 && path[0] == '.' && path[1] == '/') {
+        return path.substr(2);
+    }
+    return path;
+}
+
 bool AutogenCache::canSkipAutogen(core::GlobalState &gs, string_view cachePath, const vector<string> &changedFiles) {
     // this is here as an escape valve: if a _bunch_ of files change
     // all at once, then don't let us skip at all. This should pretty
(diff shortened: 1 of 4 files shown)


main/autogen/cache.cc
```diff
@@ -13,6 +13,13 @@ namespace sorbet::autogen {
 
 const size_t MAX_SKIP_AMOUNT = 100;
 
+string_view trimPath(string_view path) {
+    if (path.size() > 2 && path[0] == '.' && path[1] == '/') {
+        return path.substr(2);
+    }
+    return path;
+}
+
```
jez, line 22: Usually checking for this prefix is a smell. Do you want to call `GlobalState::getPrintablePath`?

main/autogen/cache.cc
```diff
@@ -74,7 +86,7 @@ bool AutogenCache::canSkipAutogen(core::GlobalState &gs, string_view cachePath,
 // this doesn't load the whole cache, since we never need to _consult_
 // the whole cache. Instead, we let it know which paths we care about
 // and load only those
-AutogenCache AutogenCache::unpackForFiles(string_view file_contents, const UnorderedSet<string> &changedFiles) {
+AutogenCache AutogenCache::unpackForFiles(string_view file_contents, const UnorderedSet<string_view> &changedFiles) {
```
jez, line 89: Are you sure these refs live long enough? I haven’t checked the lifetimes fully but it’s frequently rare that containers like this want string_view over just string

aisamanra: These refs should definitely live long enough: the set in question will never live longer than the vector that contains the strings they're built from.

PR #5828, opened 2022-05-24
Fix problem with aliased constants in VisibilityChecker

jez:
### Motivation

There were some errors on Stripe's codebase that only appeared in LSP mode
because the fast path contents of `GlobalState` differed, meaning that
`ResolveConstantsWalk` did the wrong thing on the second time through (e.g., in
`incrementalResolve`).

Weirdly, this test case didn't reproduce in `--stress-incremental-resolver`, but
did reproduce in our LSP test cases. I didn't spend a lot of time tracking down
what the differences are, but intuitively this change makes sense to me—we
shouldn't rely on constant aliases having been resolved yet, and also
`UnresolvedConstantLiterals` should resolve to their actual constant, and be
dealiased at the point where a dealias would be required (there is, in fact, a
dealias call immediately above, on the scope).

### Test plan

In addition to the tests, we might want to check performance of this, as it
_could_ cause more constant literals to be deferred to `todo*` lists? But that's
just a guess, and it doesn't hurt to check.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -274,7 +274,7 @@ private:
             core::SymbolRef resolved = id->symbol.dealias(ctx);
             core::SymbolRef result;
             if (resolved.isClassOrModule()) {
-                result = resolved.asClassOrModuleRef().data(ctx)->findMember(ctx, c.cnst);
+                result = resolved.asClassOrModuleRef().data(ctx)->findMemberNoDealias(ctx, c.cnst);
             }
 
             // Private constants are allowed to be resolved, when there is no scope set (the scope is checked above),
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -737,8 +737,8 @@ private:
         return true;
     }
 
-    static bool resolveConstants(const core::GlobalState &gs, vector<ResolveItems<ResolutionItem>> &jobs,
-                                 WorkerPool &workers) {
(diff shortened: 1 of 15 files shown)


jez:

→ https://go/bui_LkbvzK6E51nvVx
→ https://go/bui_LkbvvUrMthORyS

resolver/resolver.cc
```diff
@@ -737,8 +737,8 @@ class ResolveConstantsWalk {
         return true;
     }
 
-    static bool resolveConstants(const core::GlobalState &gs, vector<ResolveItems<ResolutionItem>> &jobs,
-                                 WorkerPool &workers) {
+    static bool resolveResolutionItems(const core::GlobalState &gs, vector<ResolveItems<ResolutionItem>> &jobs,
```
froydnj, line 740: What do you think about going further with this change:

1. rename `ResolutionItem` as `ConstantResolutionItem` for parallelism with all the other `*ResolutionItem` structures;
2. renaming this function `resolveConstantResolutionItems`

froydnj (approved):
This makes a certain amount of sense to me too.  And indeed, fixing the `findMember` call in `resolveLhs` to be `findMemberNoDealias` fixes the testcase from https://github.com/sorbet/sorbet/pull/5805#issuecomment-1135001047 ...though it induces some other errors that look like unrelated changes, mostly due to `type_member`/`type_template` now appearing to be involved in cycles.  I think `resolveLhs` needs to acquire some context for the name it's trying to resolve as to whether dealiasing is required.

PR #5829, opened 2022-05-25
rbi_gen_typechecking -> rbi_gen_errors

jez:
### Motivation

Renaming this test so that `main/lsp/LSPTypechecker.cc` appears higher
in fzf results when typing the query `typech`.

### Test plan

See included automated tests.


froydnj (approved):
"errors" seems like a weird suffix to use.  Maybe `rbi_gen_files` instead?

PR #5831, opened 2022-05-26
Use `rubyfmt` in`PATH` for document formatting instead of building as a library

reese:
On the tin -- this changes the LSP document formatting requests to use `exec("rubyfmt <file>")`, which will look for `rubyfmt` in the user's `PATH`, instead of using `rubyfmt` as a library.

### Motivation

The primary reason for this is that it better enables development of `rubyfmt` internally at Stripe -- this allows us to have our `rubyfmt` versions separate from Sorbet versioning so we can freely update one without the other.

As an added effect, the large amount of deleted code from this diff is because this removes a _lot_ of complexity from the Sorbet build, since it allows us to remove rust+cargo+rubyfmt dependencies, as well as just removing a whole stage of the build entirely.

### Test plan

I ran this build over Stripe's codebase on a branch with `rubyfmt` enabled and ensured that it could still properly format documents. Otherwise, I'm relying on existing tests.


common/common.h
```diff
@@ -138,6 +138,7 @@ template <class> inline constexpr bool always_false_v = false;
 } // namespace sorbet
 
 std::string demangle(const char *mangled);
+std::string exec(std::string cmd);
```
jez, line 141: We use a C++ library called `subprocess` (inspired by the python library of the same name, and similar to the library we use in Stripe's codebase). Can you use that instead of raw `exec`?

jez: You can grep the source code to see examples of its use

WORKSPACE
```diff
@@ -83,13 +83,6 @@ gemfile_lock_deps(
     ],
 )
 
-load("@rules_rust//rust:repositories.bzl", "rust_repositories")
-
-rust_repositories(
-    edition = "2018",
-    version = "1.51.0",
-)
```
jez, line 91: This is a small request, feel free to decline: what do you think about deleting the now-dead code in a separate PR?

If in the future we decide we want to change our mind for whatever reason, it might be nice to be able to revert _just_ the bazel/rust deletions

main/lsp/BUILD
```diff
@@ -1,3 +1,11 @@
+config_setting(
+    name = "enable_rubyfmt",
+    values = {
+        "define": "rubyfmt=true",
+    },
+    visibility = ["//visibility:public"],
+)
```
jez, line 7: Since you're deleting the build-time dependency on e.g. a rust toolchain and librubyfmt, there's no real reason why we need an `enable_rubyfmt` config setting anymore. We could build support for document formatting unconditionally, and gate it behind a normal (runtime) command line flag, rather than a build-time flag. Note: that it appears such a flag already exists and is in use.

Let's go ahead and just delete this config setting, and also delete the `noop.cc` formatting implementation, as neither are required anymore.

WORKSPACE
jez: concretely, maybe we could defer the deletions in `WORKSPACE`, `experimental/rubyfmt/`, and `third_party/` to a future PR, and only change the core of sorbet in this PR?

main/lsp/requests/formatting/document_formatting.cc
```diff
@@ -0,0 +1,59 @@
+#include "main/lsp/requests/formatting/document_formatting.h"
```
jez, line 1: ooc, why did you create a `formatting/` folder? this would be the first folder inside `main/lsp/requests/` and it seems somewhat redundant to me, especially now that we can also delete the `*_noop.cc` as mentioned in the other comment. I'd love if we could put this back in the place it was before 🙏

main/lsp/requests/formatting/document_formatting.cc
```diff
@@ -0,0 +1,59 @@
...
+        return;
+    }
+
+    variant<JSONNullObject, vector<unique_ptr<TextEdit>>> result = JSONNullObject();
+
+    auto fref = index.uri2FileRef(params->textDocument->uri);
+    if (fref.exists()) {
+        // Ensure that `rubyfmt` is in the user's path
+        auto rubyfmt_path = exec("which rubyfmt");
+        if (rubyfmt_path.empty()) {
```
jez, line 35: When you change this to use `subprocess`,  you should likely be able to catch thrown exceptions from the subprocess call to `rubyfmt` directly. This will let you delete this call to `which`. You can see in `WatchmanProcess.cc` how this works for detecting `watchman` (the example in that file uses `Popen` instead of what you probably want which is `check_output`, and as a result it has to deal with parent/child process IDs, but you should likely be able to simply catch results from check_output, ignoring anything about process IDs.)

WORKSPACE
reese: Sounds good, I'll put it into a separate PR 👍

main/lsp/requests/formatting/document_formatting.cc
```diff
@@ -0,0 +1,59 @@
...
+            Exception::raise("Could not find `rubyfmt` in PATH");
+        }
+
+        auto formattedContents = exec(fmt::format("rubyfmt {}", index.getFile(fref).path()));
```
jez, line 39: When you switch to using subprocess, you will be able to craft the argument vector directly, instead of having to do (unsafe) shell formatting like this

main/lsp/requests/formatting/document_formatting.cc
```diff
@@ -0,0 +1,59 @@
...
+
+        // Construct text edit to replace entire document.
+        vector<unique_ptr<TextEdit>> edits;
+        // Note: VS Code uses 0-indexed lines, so the lineCount will be one more line than the size of the doc.
+        edits.emplace_back(make_unique<TextEdit>(
+            make_unique<Range>(make_unique<Position>(0, 0), make_unique<Position>(index.getFile(fref).lineCount(), 0)),
+            formattedContents));
+        result = move(edits);
```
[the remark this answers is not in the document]
reese: It currently doesn't, but it'd be a good thing to have for this to land -- I'll update the command-line version to do the same thing as the library implementation and then update this.

main/lsp/requests/formatting/document_formatting.cc
reese: Took a crack at this, but unfortunately I think `check_output` only tells you if the process exits with a non-zero code, but it doesn't tell you _what_ that exit code is unless you use `Popen`, so I ended up going that route.

main/lsp/requests/document_formatting.cc
```diff
@@ -16,9 +16,38 @@ LSPTask::Phase DocumentFormattingTask::finalPhase() const {
...
+    // but a failure to initialize is the most probable cause for this.
+    INITIALIZE_ERROR = 101,
+
+    RUBYFMT_NOT_IN_PATH = 127,
+};
+
 void DocumentFormattingTask::index(LSPIndexer &index) {
     auto response = make_unique<ResponseMessage>("2.0", id, LSPMethod::TextDocumentFormatting);
-    if (!config.opts.lspDocumentFormatRubyfmtEnabled) {
+    if (!(config.opts.lspDocumentFormatRubyfmtEnabled && rubyfmt_enabled)) {
```
jez, line 50: Either in this PR or #5834, can we also delete the config `RUBYFMT_ENABLED` macro, the `rubyfmt_enabled` constexpr bool, and the `build:rubyfmt` config?

jez, line 50: Do you mind adding a comment above this configs.opts line that says something like "before we take this out of beta, we need to figure out a way to test this feature via sorbet"?

for what it's worth, it should be pretty easy to create a "fake" rubyfmt in Sorbet's tests, by just checking a bash script into the repo that returns its input unchanged. To test the exit codes, it could detect if the file contained something like `# rubyfmt-force-exit: 123` or something.

I don't think we need to block on that for this change, but i would like to block on it before we turn this feature on for everyone at Stripe.

main/lsp/requests/document_formatting.cc
```diff
@@ -30,35 +59,48 @@ void DocumentFormattingTask::index(LSPIndexer &index) {
...
-                break;
-            case experimental::rubyfmt::RubyfmtFormatError::RUBYFMT_FORMAT_ERROR_OK:
+        vector<unique_ptr<TextEdit>> edits;
+
+        auto process = subprocess::Popen({"rubyfmt", std::string(index.getFile(fref).path()).c_str()},
+                                         subprocess::output{subprocess::PIPE}, subprocess::input{subprocess::PIPE});
+        process.start_process();
+        auto processResponse = process.communicate();
+        process.wait();
+        auto returnCode = process.retcode();
```
jez, line 69: thanks for doing it this way 🙇‍♂️

main/lsp/requests/document_formatting.cc
```diff
@@ -30,35 +59,48 @@ void DocumentFormattingTask::index(LSPIndexer &index) {
...
+                // Fatal error -- crash
+                Exception::raise("Rubyfmt reported a Ruby error");
+            case RubyfmtStatus::INITIALIZE_ERROR:
+                // Fatal error -- crash
+                Exception::raise("Rubyfmt failed to initialize");
+            case RubyfmtStatus::RUBYFMT_NOT_IN_PATH:
+                Exception::raise("`rubyfmt` not found in the PATH.");
+            default:
+                Exception::raise("An unknown exception (exit code {}) occurred: {}", returnCode,
+                                 std::string(processResponse.second.buf.begin(), processResponse.second.buf.end()));
```
jez, line 103: This will crash the whole LSP server, which seems a big aggressive.

It might be better to return an LSP RPC error. You can grep for `->error =` in `main/lsp/requests/` to see some examples of where we do that. (There's even an example earlier up in this file.)

There's a list of pre-declared error codes we can use here:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#errorCodes

LSP RPC errors are usually shown

jez: I see that you're just cargo-culting what existed before. I'm happy to defer this for the current PR, but it's something that i would like to see fixed before we take this out of beta.

main/lsp/requests/document_formatting.cc
reese: Will do in #5834 👍

reese:
Just for a paper trail -- I'll follow up shortly with cleaning up the error handling and testing story, and I'll be rolling the config cleanup into #5834.

PR #5834, opened 2022-05-27
[wip] Clean up rubyfmt dependencies from build

reese:
### Motivation

After #5831, we no longer need build dependencies for `rubyfmt` (and by extension, rust + cargo). This cleans up those dependencies and also removes the extra non-blocking rubyfmt build.

### Test plan

As long as the build is green, this should be good to go.


jez (approved):

test/testdata/lsp/formatting.rb.document-formatting-rubyfmt.exp
```diff
@@ -1,5 +0,0 @@
-# typed: false
-class MyMalformedClass
-  def bad_whitespace
-  end
-end
```
reese, line 5: Taking the existing test files out for now -- in a later PR, I'll be adding proper tests, but we'll have to stub out the `rubyfmt` executable with something else, so these tests won't be applicable for that (since they won't actually do any formatting).

PR #5836, opened 2022-05-27
fix issues with method return types incorrectly resolving on the fast path

froydnj:
### Motivation

Inspired by work in #5828 and tests in #5805, this fixes one more case where we shouldn't be using existing information in the symbol table.  I said in #5828 that this location would need some kind of context-aware lookup, and this seemed to be the right way to address it.  (The other alternative I could see was to add a `TypeTemplateAliasType` and have a separate dealiasing through that, but that seemed like a lot of work for epsilon gain, if that.)

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/fast_path/method_aliased_result_type.1.rbupdate
+++ b/test/testdata/lsp/fast_path/method_aliased_result_type.1.rbupdate
@@ -1,0 +1,16 @@
+# typed: true
+extend T::Sig
+
+class MyEnum < T::Enum
+  enums do
+    X = new
+    Y = new
+  end
+end
+
+MyEnum_X = MyEnum::X
+
+sig {returns(MyEnum_X)} # error: Constant `MyEnum_X` is not a class or type alias
+def example
+  MyEnum::X
+end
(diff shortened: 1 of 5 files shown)


froydnj:

→ https://go/bui_LlcXq3wUXQUilB
→ https://go/builds/pay-server-sorbet-check--bui_LlcXzJ8K1o79wt

test/testdata/lsp/fast_path/method_aliased_result_type.1.rbupdate
```diff
@@ -0,0 +1,16 @@
+# typed: true
+extend T::Sig
```
jez, line 2: Do you want to put an `# assert-fast-path: true` in this test?

jez:
I think we can do this without having to introduce new bits onto the symbol table:

https://github.com/sorbet/sorbet/pull/5847

I'm generally -1 on introducing new bits that are redundant with information that is already in the symbol table, as it makes it introduces a point where one of the sources of truth can be updated while forgetting to update the other source of truth.

jez (changes requested):
https://github.com/sorbet/sorbet/pull/5847

^ curious what you think

test/testdata/lsp/fast_path/method_aliased_result_type.1.rbupdate
jez: moved to #5847

jez (approved):

PR #5839, opened 2022-05-27
Add FAQ about single file

jez:
### Test plan

See included automated tests.

### DIFF
--- a/website/docs/faq.md
+++ b/website/docs/faq.md
@@ -469,3 +469,44 @@ response.
 
 Sorbet does not support duck typing either for static analysis or runtime
 checking.
+
+## How do I see errors from a single file, not the whole project?
+
+Fundamentally, Sorbet typechecks a single project at a time. Unlike other
+languages and compilers, it does not process files in a codebase one file at a
+time. This is largely due to the fact that Ruby does not have any sort of
+file-level import mechanism—in Ruby, code can usually be defined anywhere and
+used from anywhere else—and this places constraints on how Sorbet must be
+implemented.
+
+Because Sorbet typechecks a single project at a time, it does not allow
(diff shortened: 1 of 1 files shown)


website/docs/faq.md
```diff
@@ -469,3 +469,49 @@ response.
...
+    [Strictness Levels](static.md#file-level-granularity-strictness-levels)).
+    Note that the default strictness level when there is no explicit `# typed:`
+    comment in a file is `# typed: false`. The `--typed=false` command line flag
+    can be used to forcibly override every file's strictness level to
+    `# typed: false`, regardless of what's written in the file.
+
+    Forcing every file to `# typed: false` will silence all but the most
+    critical Sorbet errors throughout the project.
+
+1.  Proceed to fix these errors. If there are still an overwhelming number of
```
froydnj, line 497:
```suggestion
2.  Proceed to fix these errors. If there are still an overwhelming number of
```

website/docs/faq.md
```diff
@@ -469,3 +469,49 @@ response.
...
+    errors, tackle the errors that are reported earlier first. Sorbet's design
+    means that errors discovered early in the early phases of typechecking can
+    cause redundant errors in later phases.
+
+1.  Once there are no errors in any files at `# typed: false`, proceed to
```
froydnj, line 502:
```suggestion
3.  Once there are no errors in any files at `# typed: false`, proceed to
```

froydnj (approved):

website/docs/faq.md
jez: markdown automatically converts these to the correct numbers (because it just converts to an `<ol>` element, and then the number is computed by the browser). it's useful to do this so that future insertions to the middle of the list don't have to renumber the whole list.

PR #5842, opened 2022-05-29
Recover from class/module/def meeting eof

### LINKED ISSUE #5438: Recover from `unexpected "end of file"`
#### Input

[→ View on sorbet.run](https://sorbet.run/?arg=--print=parse-tree-whitequark#%23%20typed%3A%20true%0Adef%20foo%0A%20%20puts%20'hello')

```ruby
# typed: true
def foo
  puts 'hello'
```

#### Observed output

```
s(:begin)
editor.rb:3: unexpected token "end of file" https://srb.help/2001
     3 |  puts 'hello'
     4 |
Errors: 1
```

### Expected behavior

Sorbet should know that there was the start of a `def foo` declaration, and that it had a `puts 'hello'` in the body.

- - -

This one is tricky.

If we do something naive like write a grammar rule like

```
kend_or_eof: kEND
           | eof
```

this will interfere with our indentation-aware error recovery. For example, it will make a test case like this:

[→ View on sorbet.run](https://sorbet.run/?arg=--print=desugar-tree#%23%20typed%3A%20true%0A%23%20Craft%20your%20test%20case%2C%20then%20share%20at%20http%3A%2F%2Fgo%2Fask%2Fsorbet%2C%20thanks!%0A%0Aclass%20A%0A%20%20puts%20'before%20def'%0A%0A%20%20def%20foo%0A%0A%20%20puts%20'after%20def'%0Aend)

```ruby
```

jez:
### Motivation

Fixes #5438.

Might be worth reading the notes I left myself in that ticket as motivation for
this implementation.

### Test plan

- [x] @jez Write tests

### DIFF
--- a/parser/parser/cc/grammars/typedruby.ypp
+++ b/parser/parser/cc/grammars/typedruby.ypp
@@ -194,6 +194,7 @@ using namespace std::string_literals;
   tFLOAT_IMAGINARY    1144
   tBDOT2              1145    // Used to avoid shift/reduce conflicts with productions using `tDOT2`
   tBDOT3              1146    // Used to avoid shift/reduce conflicts with productions using `tDOT3`
+  tBEFORE_EOF         1147
 ;
 
 %printer { yyo << $$->view(); } tIDENTIFIER
@@ -377,6 +378,7 @@ using namespace std::string_literals;
   then
   relop
   k_return
+  kend_or_eof
   lcurly_block_start
   lbrace_cmd_block_start
   until
@@ -503,6 +505,12 @@ void parser::error(const ruby_parser::location &lloc, const std::string &msg) {
 
 int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser::TYPEDRUBY &driver) {
   auto token = driver.lex.advance();
+  if (token->type() == ruby_parser::token_type::tBEFORE_EOF && !driver.indentationAware) {
+    // Only emit tBEFORE_EOF in indentation-aware error recovery mode.
+    // Otherwise, skip to eof token.
+    token = driver.lex.advance();
+    ENFORCE(token->type() == ruby_parser::token_type::eof);
+  }
   driver.last_token = token;
   int token_type = static_cast<int>(token->type());
   assert(token_type >= 0);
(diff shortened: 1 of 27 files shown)


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -1930,6 +1947,8 @@ array_premature_end: eof
                     {
                       driver.replace_stack($<stack>2);
                       $$ = driver.build.beginKeyword(self, $1, $3->body, $4);
+                      // TODO(jez) Make rewind_if_dedented aware of tBEFORE_EOF
+                      // TODO(jez) Use kend_or_eof in all rules that use rewind_if_dedented
```
jez, line 1951: I'm going to defer this to a future PR to keep this PR small and make it easier to bisect/roll back in case that becomes necessary.

parser/parser/cc/lexer.rl
```diff
@@ -558,7 +558,12 @@ static bool eof_codepoint(char c) {
 token_t lexer::advance_() {
   if (!token_queue.empty()) {
     token_t token = token_queue.front();
-    token_queue.pop_front();
+    if (token->type() == token_type::eof) {
+      // spin on EOF forever, even if advance is called repeatedly
```
jez, line 562: This used to be the previous behavior as well, by nature of how the `p` and `pe` variables would check.

There was no limit to the number of times that you could call `advance()` (and thus `advance_()`). If `p == pe`, ragel would just exit the `%exec` block of generated code, skipping to the bottom of the `advance_()` function, which always exited by just returning an `eof` token.

I only wanted to emit the `tBEFORE_EOF` function **once**, not do something like accidentally alternate between `..., tBEFORE_EOF, eof, tBEFORE_EOF, eof`. Doing this trick with the `token_queue` seemed like the cleanest trick.

parser/parser/cc/lexer.rl
```diff
@@ -3047,6 +3054,7 @@ void lexer::rewind_and_reset_to_expr_end(size_t newPos) {
 
   // rewind
   this->_p = this->source_buffer.data() + newPos;
+  token_queue.clear();
```
jez, line 3057: These two `.clear()` calls happen because if we reached the end of the file and have decided to reset `p` back to some earlier point in the file (backtrack), any tokens that happened to be in the queue don't make sense anymore--we should entirely start lexing tokens from the new `p`.

jez:
@froydnj this should be good to review. no rush, not urgent.

froydnj (approved):
This is very cool!

PR #5843, opened 2022-05-30
Better indentation-aware recovery for class & module

### LINKED ISSUE #5841: Incorrect indentation-aware recovery for class and module
#### Input

[→ View on sorbet.run](https://sorbet.run/?arg=--print=desugar-tree#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20module%20A%0A%20%20%20%20Integer%0A%0A%20%20String%0Aend)

```ruby
# typed: true

class A
  module B
    Integer

  String
end
```

#### Observed output

```
  class <emptyTree>::<C A><<C <todo sym>>> < (::<todo sym>)
    module <emptyTree>::<C B><<C <todo sym>>> < ()
      <emptyTree>
    end

    <emptyTree>::<C Integer>

    <emptyTree>::<C String>
  end
end
editor.rb:8: unexpected token "end of file" https://srb.help/2001
     8 |end
     9 |

editor.rb:4: Hint: this "module" token might not be properly closed https://srb.help/2003
     4 |  module B
          ^^^^^^
    editor.rb:8: Matching end found here but is not indented as far
     8 |end
        ^^^
Errors: 2
```

#### Expected behavior

If we replace the `module B` in the example above with `def foo`, we see this behavior:

[→ View on sorbet.run](https://sorbet.run/?arg=--print=desugar-tree#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20def%20foo%0A%20%20%20%20Integer%0A%0A%20%20String%0Aend)

```ruby
# typed: true

class A
  def foo
    Integer

  String
end
```

```
  class <emptyTree>::<C A><<C <todo sym>>> < (::<todo sym>)
    def foo<<todo method>>(&<blk>)
      <emptyTree>::<C Integer>
    end

    <emptyTree>::<C String>
  end
end
editor.rb:8: unexpected token "end of file" https://srb.help/2001
     8 |end
     9 |

editor.rb:4: Hint: this "def" token might not be properly closed https://srb.help/2003
     4 |  def foo
          ^^^
    editor.rb:8: Matching end found here but is not indented as far
     8 |end
        ^^^
Errors: 2
```

Note how the parse tree shows `Integer` nested inside `def foo`, and the `String` is outside the `def foo`.

In the buggy program, both `Integer` and `String` are nested outside of `module B`, being placed instead under `class A`.

- - -

jez:
### Motivation

Fixes #5841.

The comment in the parser is the best explanation of what was going on.

The key difference between class/module defs and method defs is that
method defs don't emit a tNL after the arguments end, while classe and
modules do. That means that classes and modules have their stmts list
constructed by starting from empty, reducing the `tNL`, and appending
the first element. Meanwhile methods start from the singleton
production, with no leading `tNL`.

### Test plan

See included automated tests.

### DIFF
--- a/parser/parser/cc/grammars/typedruby.ypp
+++ b/parser/parser/cc/grammars/typedruby.ypp
@@ -612,6 +612,14 @@ int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser:
                 | stmts terms stmt_or_begin
                     {
                       auto *result = $1;
+                      if (result->inner->empty()) {
+                        // This set of stmts started with a leading tNL (common after e.g. class and module defs)
+                        // Since an empty stmts reduces without any tokens, it uses the begin/end location of the
+                        // most recently lexed token, despite that token not being included in the stmts.
+                        // Manually fix up the locations here.
+                        @$.begin = @3.begin;
+                        @$.end = @3.end;
+                      }
                       result->inner->emplace_back($3);
                       $$ = result;
 
--- a/parser/parser/include/ruby_parser/driver.hh
+++ b/parser/parser/include/ruby_parser/driver.hh
@@ -26,6 +26,10 @@ struct node_list {
         return nodes.size();
     }
 
+    inline bool empty() const {
+        return nodes.empty();
+    }
+
     inline void emplace_back(const ForeignPtr &ptr) {
         nodes.emplace_back(ptr);
     }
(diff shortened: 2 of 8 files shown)


froydnj (changes requested):
It looks like there are some errors in tests that weren't added in this PR.

PR #5846, opened 2022-05-30
Handle other kEND usages

jez:
### Motivation

Follow up to the changes in #5843 to support other pieces of syntax.

Best to review that change first, because this change relies entirely on the
technique in that PR.

### Test plan

See included automated tests.


parser/parser/cc/driver.cc
```diff
@@ -58,11 +58,17 @@ void base_driver::rewind_and_reset_to_beg(size_t newPos) {
     this->lex.rewind_and_reset_to_expr_beg(newPos);
 }
 
-void base_driver::rewind_if_dedented(token_t token, token_t kEND, bool force) {
-    if ((force || this->indentationAware) && this->lex.compare_indent_level(token, kEND) > 0) {
-        this->rewind_and_reset(kEND->start());
+void base_driver::rewind_if_dedented(token_t token, token_t endToken, bool force) {
+    if ((force || this->indentationAware) &&
+        (endToken->type() == token_type::tBEFORE_EOF || this->lex.compare_indent_level(token, endToken) > 0)) {
```
froydnj, line 63: WDYT about inverting this condition (and maybe splitting it up with separate returns for finer-grained conditions, though I think that part is more a matter of taste) so that we can exit early and decrease the indent level of the rest of the code?  I think it would make the logic clearer.

froydnj (approved):

PR #5847, opened 2022-05-31
Correctly report error when changing method types take fast path

jez:
### Motivation

I'm generally -1 on having multiple sources of truth for things in the
symbol table. If it's possible to get the answer we need with existing
data, that tends to keep things simpler over time.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/fast_path/method_aliased_result_type.1.rbupdate
+++ b/test/testdata/lsp/fast_path/method_aliased_result_type.1.rbupdate
@@ -1,4 +1,5 @@
 # typed: true
+# assert-fast-path: method_aliased_result_type.rb
 extend T::Sig
 
 class MyEnum < T::Enum


froydnj (changes requested):
It's nicer to not add extra bits.  We already have extra bits because we're overloading `AliasType` to mean different things depending on context, though, so :man_shrugging: 

Happy to stamp the patch, but it looks like there are some `ENFORCE`s firing?

jez:
The ENFORCE failed for an existing test that did

```ruby
class A
  extend T::Generic
  X = type_member
  Y = X
end
```

(this is just a normal constant alias, so it's expected that dealiasing `Y` to `X` will be owned by `A` not `T.class_of(A)`, which is what I was enforcing. it just means that we have to move that ENFORCE into the branch condition)

froydnj (approved):

PR #5848, opened 2022-05-31
Find behavior conflicts across threads

ngroman:
Adds a merge step for maps of class to "behavior" loc in `Namer::symbolizeTrees`. This ensures conflicts are found even if they are processed by different workers.

### Motivation
This check was broken if the conflicting defs were processed by different workers.

### Test plan

See included automated tests.

Introduced a conflict in Stripe's codebase and ensured this caught it.

Added timing. The merge ops are fast enough on the codebase that they don't appear in timing data.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1163,6 +1163,9 @@ public:
     }
 };
 
+using BehaviorLocs = InlinedVector<core::Loc, 1>;
+using ClassBehaviorLocsMap = UnorderedMap<core::ClassOrModuleRef, BehaviorLocs>;
+
 /**
  * Inserts newly created symbols (from SymbolDefiner) into a tree.
  */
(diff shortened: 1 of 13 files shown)


ngroman:
cc @aadi-stripe

nroman-stripe:

→ https://go/builds/bui_LnCwXQM6mjiQBm
→ https://go/builds/bui_LnCwNdOLQfkvyf

jez (changes requested):
Do you mind adding a test?

ngroman:
> Do you mind adding a test?

There's already a test for this in https://github.com/sorbet/sorbet/blob/master/test/cli/conflicting-definition/test.out

But it was passing because the same worker gets both files (I'm guessing this is because there are only two files). I'm not sure what a good way of testing this is given you need to rely on how the workers receive jobs. Do you have any ideas here?

jez:
The number of workers will be `min(cores, numFiles / 2)`

https://github.com/sorbet/sorbet/blob/d1bd0b2c0aa7a943737db6a7f3f7f08a624d5aa9/main/options/options.cc#L834

What do you think about making a CLI test with 10 files, where each file pairwise conflicts with all other files?

ngroman:
This is ready for re-review now.

jez:

→ https://go/builds/bui_LpVYLkG4TuxsAQ
→ https://go/builds/bui_LpVYUFhzsj0svP

jez:
This looks mostly good to me. My one question is whether creating a map this big makes namer noticeably slower.

jez (approved):

jez:
I took a look at the performance and there's no noticeable change.

PR #5856, opened 2022-06-01
remove logging for empty completion requests

froydnj:
### Motivation

We're not working on completion right now, and having this (particularly logging the source text for empty responses) makes the logs rather verbose.

### Test plan

See included automated tests.


jez (approved):
What do you think about having it in behind a false branch? I’m fine either way

PR #5849, opened 2022-05-31
Move #then/yield_self to Kernel

gustavodegeus:
[This example](https://sorbet.run/?arg=--enable-experimental-requires-ancestor#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Amodule%20Interface%0A%20%20requires_ancestor%20%7B%20Kernel%20%7D%0A%0A%20%20extend%28T%3A%3AHelpers%29%0A%20%20extend%20T%3A%3ASig%0A%20%20abstract!%0Aend%0A%0Aclass%20MyClass%0A%20%20extend%28T%3A%3ASig%29%0A%20%20include%28Interface%29%0Aend%0A%0Asig%7Breturns%28Interface%29%7D%0Adef%20foo%0A%20%20MyClass.new%0Aend%0A%0Adef%20bar%0A%20%20foo.then%20do%20%7Cstuff%7C%0A%20%20%20%20puts%20stuff.to_s%0A%20%20end%0A%0A%20%20%23It's%20weird%20because%20'tap'%20is%20available%0A%20%20foo.tap%20do%20%7Cstuff%7C%0A%20%20%20%20puts%20stuff.to_s%0A%20%20end%0Aend%0A) should work based on Ruby's spec.
If we have an interface with `requires_ancestor { Kernel }`, we should be able to access `then/yield_self` when working with the interface.

### Motivation
The signature for `#then` and `#yield_self` are in the wrong place. 
It should live within Kernel [based on Ruby's](https://github.com/ruby/ruby/blob/936e54063d4de0e44925e9139cb4d25cce27c187/kernel.rb#L120) spec.

### Test plan

I moved the exiting test to `Kernel` and deleted the `Object` file as it was the only test that was there.

PR #5858, opened 2022-06-01
Move #itself and #object_id to Kernel

gustavodegeus:
Follow up to https://github.com/sorbet/sorbet/pull/5849

### Motivation

As I was moving `then/yield_self` it was pointed out that `#object_id` and `#itself` also belongs to `Kernel` as per Ruby specs.

 - [object_id](https://github.com/ruby/ruby/blob/1177665e6224f8491db82997c8774e9485564e41/gc.c#L14065)
 - [itself](https://github.com/ruby/ruby/blob/1177665e6224f8491db82997c8774e9485564e41/object.c#L4318)

### Test plan
I've updated the kernel test data as well the expectations for the hidden method finders.

### DIFF
--- a/rbi/core/object.rbi
+++ b/rbi/core/object.rbi
@@ -27,36 +27,4 @@
 class Object < BasicObject
   include Kernel
 
-  # Returns an integer identifier for `obj`.
-  #
-  # The same number will be returned on all calls to `object_id` for a given
-  # object, and no two active objects will share an id.
-  #
-  # Note: that some objects of builtin classes are reused for optimization. This
-  # is the case for immediate values and frozen string literals.
-  #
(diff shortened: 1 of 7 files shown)


rbi/core/object.rbi
```diff
@@ -27,36 +27,4 @@
 class Object < BasicObject
   include Kernel
 
-  # Returns an integer identifier for `obj`.
```
gustavodegeus, line 30: I wasn't sure if I should delete this file or its still useful as is 🤔

jez: I think this is the only file that declares to Sorbet that `class Object` has `include Kernel`, so I think it's important to keep this.

PR #5808, opened 2022-05-18
Adding, changing, or removing a method takes the fast path

jez:
### Motivation

This is a large change, and you are recommended to review by commit.

The general idea is that by threading information from a previous namer run to a current namer run, we can get enough information into namer to have it completely delete old symbols from the symbol table, so that when the current namer run enters methods, it appears to namer as if GlobalState is effectively empty (at least, empty w.r.t. to methods it's about to define).

We couple this with the existing logic to figure out which downstream files need to be retypechecked.

One key observation is that the set of files that is sent to namer is the same as the set of files that will be typechecked downstream.

This is good for some things (e.g., this makes it easy to have namer re-run on both the files that were edited as well as any downstream files that use `alias_method` on one of those methods) but is bad for others (it means that selectivity will be particularly bad.

For fast-path edits within a file, which do not modify any method symbols' hashes, this PR should not make anything slower.

But it will likely change the distribution of how many files are typechecked on the fast path, and so this may not actually be much of a speed improvement if adding/deleting a method that is used in many places. There are future projects (using package information? etc.) that we can build to limit the set of files that are retypechecked.

### Test plan

See included automated tests.

PR #5860, opened 2022-06-01
methodArgumentHash: vector<uint32_t> -> uint32_t

jez:
### Motivation

Despite this being a vector, we were never using it as a vector really. We
were always just using it as a single hash value.

It will be convenient for #5808 if I can just call this method and get a 
uint32_t, and since that already works for all the current use cases, I've
pulled it out as a pre-work change.

### Test plan

Existing tests

### DIFF
--- a/ast/ArgParsing.cc
+++ b/ast/ArgParsing.cc
@@ -70,21 +70,20 @@ vector<core::ParsedArg> ArgParsing::parseArgs(const ast::MethodDef::ARGS_store &
     return parsedArgs;
 }
 
-std::vector<uint32_t> ArgParsing::hashArgs(core::Context ctx, const std::vector<core::ParsedArg> &args) {
-    std::vector<uint32_t> result;
-    result.reserve(args.size());
+// This has to match the implementation of Method::methodArgumentHash
+uint32_t ArgParsing::hashArgs(core::Context ctx, const std::vector<core::ParsedArg> &args) {
+    uint32_t result = 0;
     for (const auto &e : args) {
-        uint32_t arg = 0;
         if (e.flags.isKeyword) {
             if (e.flags.isRepeated && e.local._name != core::Names::fwdKwargs()) {
                 auto name = core::Names::kwargs();
-                arg = core::mix(arg, core::_hash(name.shortName(ctx)));
+                result = core::mix(result, core::_hash(name.shortName(ctx)));
             } else {
-                arg = core::mix(arg, core::_hash(e.local._name.shortName(ctx)));
+                result = core::mix(result, core::_hash(e.local._name.shortName(ctx)));
             }
         }
 
-        result.push_back(core::mix(arg, e.flags.toU1()));
+        result = core::mix(result, e.flags.toU1());
     }
     return result;
 }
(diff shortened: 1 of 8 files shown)


ast/ArgParsing.cc
```diff
@@ -70,35 +70,21 @@ vector<core::ParsedArg> ArgParsing::parseArgs(const ast::MethodDef::ARGS_store &
     return parsedArgs;
 }
 
-std::vector<uint32_t> ArgParsing::hashArgs(core::Context ctx, const std::vector<core::ParsedArg> &args) {
-    std::vector<uint32_t> result;
-    result.reserve(args.size());
+// This has to match the implementation of Method::methodArgumentHash
+uint32_t ArgParsing::hashArgs(core::Context ctx, const std::vector<core::ParsedArg> &args) {
```
froydnj, line 74: Non-blocking, but feels like we should really newtype this at some point.

core/Symbols.cc
```diff
@@ -2295,17 +2293,17 @@ uint32_t Method::methodShapeHash(const GlobalState &gs) const {
     return result;
 }
 
-vector<uint32_t> Method::methodArgumentHash(const GlobalState &gs) const {
-    vector<uint32_t> result;
-    result.reserve(arguments.size());
+// This has to match the implementation of ArgParsing::hashArgs
```
froydnj, line 2296: I like this comment, but given that the current implementation doesn't actually match `ArgParsing::hashArgs`, it seems like we need to change one of the implementations or update the comment.

core/FoundDefinitions.h
```diff
@@ -156,11 +156,11 @@ struct FoundMethod final {
     Flags flags;
     CheckSize(Flags, 1, 1);
     std::vector<core::ParsedArg> parsedArgs;
-    std::vector<uint32_t> argsHash;
+    uint32_t argsHash;
 
     std::string toString(const core::GlobalState &gs, const FoundDefinitions &foundDefs, uint32_t id) const;
 };
-CheckSize(FoundMethod, 80, 8);
+CheckSize(FoundMethod, 64, 8);
```
froydnj, line 163: This is a nice size reduction!

You should be able to get this down a little further if you swap `parsedArgs` and `argsHash`, so the latter is packed into the same 8 bytes that `flags` currently occupies`.

froydnj (approved):

core/FoundDefinitions.h
jez: I'm not able to get core::FoundMethod any smaller than 64 bytes, am i doing something wrong?

I think that it has 57 bytes worth of information and then aligns to 8 bytes, so it pads the whole thing up to 64 instead of 60

I'm going to land this for now to unblock future changes, and if we want to tweak this in a future change we should feel free to.

Luckily these structures are transient (only used by namer) so they don't contribute a particularly large memory pressure at the moment (unless i'm mistaken).

core/Symbols.cc
jez: I will take a look at this in a follow up (if it's broken because of this change, it was already broken on master)

PR #5861, opened 2022-06-01
Stabilize document formatting LSP functionality

reese:
The final installment (for now) of my `rubyfmt` journey: this enables document formatting as a stable feature in the LSP. This is mostly removing the config flags that disable formatting and updating any tests

### Motivation

This will be gradually rolled out as a default formatter at Stripe, so this will need to be on the stable configuration.

### Test plan

Existing tests.

### DIFF
--- a/main/lsp/requests/initialize.cc
+++ b/main/lsp/requests/initialize.cc
@@ -31,7 +31,7 @@ unique_ptr<ResponseMessage> InitializeTask::runRequest(LSPTypecheckerInterface &
     serverCap->hoverProvider = true;
     serverCap->referencesProvider = true;
     serverCap->implementationProvider = true;
-    serverCap->documentFormattingProvider = opts.lspDocumentFormatRubyfmtEnabled;
+    serverCap->documentFormattingProvider = true;
     serverCap->sorbetShowSymbolProvider = true;
 
     auto codeActionProvider = make_unique<CodeActionOptions>();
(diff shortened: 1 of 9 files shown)


main/lsp/requests/initialize.cc
```diff
@@ -31,7 +31,7 @@ unique_ptr<ResponseMessage> InitializeTask::runRequest(LSPTypecheckerInterface &
     serverCap->hoverProvider = true;
     serverCap->referencesProvider = true;
     serverCap->implementationProvider = true;
-    serverCap->documentFormattingProvider = opts.lspDocumentFormatRubyfmtEnabled;
+    serverCap->documentFormattingProvider = true;
```
jez, line 34: I don't think we can roll this out as is. This will declare that Sorbet support responding to document formatting requests for every Sorbet codebase, which it does not. It only supports responding to them _if_ `rubyfmt` (or whatever specified by the `--rubyfmt-path` option) is executable.

Separately, heads up that I found a bug on your earlier PR, just wanted to call attention to it so that the notification doesn't get lost:

https://github.com/sorbet/sorbet/pull/5857/files#r887355630

Are you sure you really want to take this out of beta? Is there harm in merely passing the `--enable-experimental-lsp-document-formatting-rubyfmt` option in Stripe's Sorbet configuration while we're experimenting, and remove it once the migration is further along?

jez (changes requested):
^

main/lsp/requests/initialize.cc
reese: Adding the experimental flag to the default config internally is fine by me -- wasn't sure if we wanted to avoid shoving additional experimental flags in there, but if we'd rather let that sit for a while until we roll it out more, that works for me 👍

reese: (I'll close out this PR for now, and I'll follow up with that flag-related bug in a separate PR.)

PR #5863, opened 2022-06-02
Improve some fast-path/slow-path tests

jez:
### Motivation

Pre-work for #5808.

The only difference in behavior on these tests should be that the assertion at
the top of the file goes away, and these files take the fast path instead of the
slow path.

### Test plan

See included automated tests.


froydnj (approved):
I vote for not removing the assertion in the coming changes, but converting it into an `assert-fast-path` so we don't accidentally regress into the slow path for the tests and not notice.

jez:
Yep sorry that was always the plan (technically it’s removing the slow path assertion and adding the fast path one and I shortened it in a rush)

PR #5864, opened 2022-06-02
Fix visibility checker for superclass declarations

aadi-stripe:
Fixes a bug in the new package enforcement which ends up ignoring constant references if they are used in a superclass declaration.

### Motivation

The new package visibility checker has a bug -- it ignores references of the form:

```
class MyPackage::A < OtherPackage::B # OtherPackage::B is not checked!
```

This was tied to its handling of `keep_for_ide` declarations. We can safely remove this handling and simply ignore class definitions. Correct name-spacing of class declarations is already enforced by the packager pass.

### Test plan

### DIFF
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -423,6 +423,16 @@ public:
         return tree;
     }
 
+    ast::ExpressionPtr preTransformClassDef(core::Context ctx, ast::ExpressionPtr tree) {
+        auto &original = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+        if (original.kind == ast::ClassDef::Kind::Class && !original.ancestors.empty()) {
+            auto &superClass = original.ancestors[0];
+            superClass = ast::TreeMap::apply(ctx, *this, std::move(superClass));
+        }
+
+        return tree;
+    }
+
     static std::vector<ast::ParsedFile> run(const core::GlobalState &gs, WorkerPool &workers,
                                             std::vector<ast::ParsedFile> files) {
         Timer timeit(gs.tracer(), "visibility_checker.check_visibility");
--- a/test/cli/package-top-level-inheritance/__package.rb
+++ b/test/cli/package-top-level-inheritance/__package.rb
@@ -1,0 +1,5 @@
+# typed: strict
+
+class Main < PackageSpec
+  export Main::B
+end
--- a/test/cli/package-top-level-inheritance/a/__package.rb
+++ b/test/cli/package-top-level-inheritance/a/__package.rb
@@ -1,0 +1,5 @@
+# typed: strict
+
+class Main::A < PackageSpec
+end
+
--- a/test/cli/package-top-level-inheritance/a/foo.rb
+++ b/test/cli/package-top-level-inheritance/a/foo.rb
@@ -1,0 +1,8 @@
+# typed: strict
+
+module Main
+  module A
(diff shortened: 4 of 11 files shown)


aadi-stripe:

→ https://go/builds/bui_LngKUpMjZ82Fkn
→ https://go/builds/bui_LngKOtLWgl1xqQ

jez (changes requested):
I think we do, in fact, have to keep the `keep_for_ide` logic.

I've pushed a test case to your branch that passes on master but fails after this change, and I think the correct behavior is the master behavior.

packager/VisibilityChecker.cc
```diff
@@ -297,35 +297,18 @@ class VisibilityCheckerPass final {
 public:
     const core::packages::PackageInfo &package;
     const bool insideTestFile;
+    UnorderedSet<core::SymbolRef> defs;
```
jez, line 300: Can we use a more descriptive name and/or add a comment with what the meaning is?

packager/VisibilityChecker.cc
```diff
@@ -423,6 +406,18 @@ class VisibilityCheckerPass final {
         return tree;
     }
 
+    ast::ExpressionPtr preTransformClassDef(core::Context ctx, ast::ExpressionPtr tree) {
+        auto &original = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+        if (!ast::isa_tree<ast::ConstantLit>(original.name)) {
+            return tree;
+        }
+
+        auto &sym = (ast::cast_tree_nonnull<ast::ConstantLit>(original.name)).symbol;
```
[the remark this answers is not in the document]
aadi-stripe: Ah yes sorry for the oversight.

aadi-stripe:
@jez Fixed, ptal

aadi-stripe:

→ https://go/builds/bui_LnyMGEK7lW78WJ
→ https://go/builds/bui_LnyMKRJQoaWlhG

packager/VisibilityChecker.cc
```diff
@@ -423,6 +423,18 @@ class VisibilityCheckerPass final {
...
     }
 
+    ast::ExpressionPtr preTransformClassDef(core::Context ctx, ast::ExpressionPtr tree) {
+        auto &original = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+        auto ait = original.ancestors.begin();
+        // if this is a class, then the first ancestor is the parent class
+        if (original.kind == ast::ClassDef::Kind::Class && !original.ancestors.empty()) {
+            *ait = ast::TreeMap::apply(ctx, *this, std::move(*ait));
+            ++ait;
+        }
```
jez, line 433: Not sure what we're using iterators for here 🤔 

maybe this started as a loop?

```suggestion
        if (original.kind == ast::ClassDef::Kind::Class && !original.ancestors.empty()) {
            auto &superClass = original.ancestors[0];
            superClass = ast::TreeMap::apply(ctx, *this, std::move(superClass));
        }
```

jez (approved):

packager/VisibilityChecker.cc
aadi-stripe: ah yeah sorry I just pulled this from autogen, we only need the first one nvm

aadi-stripe:

→ https://go/builds/bui_LnzEwhvTL51gH5
→ https://go/builds/bui_LnzEAR9YiapgnE

PR #5870, opened 2022-06-02
Always take slow path for changes to required ancestors

jez:
### Motivation

See the test case.

I discovered this bug while working on #5808.

The bug manifests in even more cases after the changes I've made on that branch,
but the fix for this fixes all of them. Since I managed to reproduce it on
master, I've pulled the fix for it out as well.

cc @Morriar

### Test plan

The commits showcase the before/after behavior of the new test.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -2285,7 +2285,7 @@ uint32_t Method::methodShapeHash(const GlobalState &gs) const {
 
     if (name == Names::unresolvedAncestors() || name == Names::requiredAncestors() ||
         name == Names::requiredAncestorsLin()) {
-        // These are a synthetic methods that encode information about the class hierarchy in their types.
+        // These are synthetic methods that encode information about the class hierarchy in their types.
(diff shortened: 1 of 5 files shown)


core/Symbols.cc
```diff
@@ -2283,9 +2283,10 @@ uint32_t Method::methodShapeHash(const GlobalState &gs) const {
     result = mix(result, this->hasSig());
     result = mix(result, this->methodArgumentHash(gs));
 
-    if (name == core::Names::unresolvedAncestors()) {
-        // This is a synthetic method that encodes the superclasses of its owning class in its return type.
-        // If the return type changes, we must take the slow path.
+    if (name == Names::unresolvedAncestors() || name == Names::requiredAncestors() ||
+        name == Names::requiredAncestorsLin()) {
+        // These are a synthetic methods that encode information about the class hierarchy in their types.
```
froydnj, line 2288:
```suggestion
        // These are synthetic methods that encode information about the class hierarchy in their types.
```

froydnj (approved):

PR #5874, opened 2022-06-03
Disallow packages to transitively export symbols imported from other packages

aadi-stripe:
In the current `--stripe-packages` design, it is possible for a package to export a symbol that it does not itself own (but instead itself imports from another package). This is not useful in the current Stripe codebase, and in fact conflicts with our package-file generation tooling.

It's not entirely clear, but this functionality may have been added based on legacy assumptions tied to the old `--stripe-packages` design.

### Motivation

Remove unnecessary functionality.

### Test plan

Tested on Stripe codebase, and updated the test that verified this functionality.


test/testdata/packager/export_imported/a/__package.rb
```diff
@@ -1,12 +0,0 @@
...
-# enable-packager: true
-
-# Edge case: We technically allow packages to export their imports. It's easier to implement packages this way
-# (although this practice should be discouraged!)
-
-class A < PackageSpec
-  import B
-
-  export B::BClass
-end
```
jez, line 12: We should almost never delete tests.

Can you change these tests to simply pass, using whatever error assertions are necessary?

jez (approved):
One request.

test/testdata/packager/export_imported/a/__package.rb
aadi-stripe: Ack

PR #5876, opened 2022-06-03
Revert "make static field changes take the fast path (#5805)"

froydnj:
This reverts commit f5d78adfb53361dd1b1a37a193afd3faae6a22fb.

### Motivation

Defensive PR just in case #5805 turns out to have horrible show-stopping bugs once a bunch of developers start banging on it.

### Test plan

See included automated tests.


froydnj:
The rollout at Stripe has been peaceful thus far, so it looks like we don't need this.

PR #5877, opened 2022-06-03
unify `TreeMapper` and `ShallowMapper` via template parameters

froydnj:
### Motivation

We shouldn't have to duplicate hundreds of lines of code just because one member of one kind of tree is treated differently.

### Test plan

See included automated tests.

### DIFF
--- a/ast/treemap/treemap.h
+++ b/ast/treemap/treemap.h
@@ -138,7 +138,7 @@ struct ReportedRubyException {
     core::LocOffsets onLoc;
 };
 
-enum class MapKind {
+enum class TreeMapKind {
     Full,
     Shallow,
 };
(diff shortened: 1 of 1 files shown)


jez (approved):
> We shouldn't have to duplicate hundreds of lines of code just because one member of one kind of tree is treated differently.

Have you checked whether this affects compile times?

ast/treemap/treemap.h
```diff
@@ -138,15 +138,21 @@ struct ReportedRubyException {
     core::LocOffsets onLoc;
 };
 
+enum class MapKind {
```
jez, line 141: Maybe `TreeMapKind`? This is currently `ast::MapKind` which is maybe a little unclear from context what it refers to.

froydnj:
> Have you checked whether this affects compile times?

I didn't notice anything radically different when I wrote the patch.

My experience is that templates are not the place to go hunting for compile-time wins. There are certainly cases where particular template techniques can do terrible things to compile times, of course, but this technique is not one of them.

PR #5878, opened 2022-06-03
optimize for size in release builds

froydnj:
### Motivation

Sometimes optimizing for size is better than optimizing for speed.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_LoDR8rnK5hHcqu
→ https://go/builds/bui_LoDRsob3dKQJ6X

froydnj:
Benchmarking says this is clearly not worth it.

PR #5881, opened 2022-06-03
Check that rubyfmt-path exists before enabling document formatting

reese:
On the tin -- `Popen` will abort if the first argument is a nonexistent file, so we don't want to enable document formatting in these cases.

### Motivation

User reported a crash in Slack that was caused by this.

### Test plan

I manually ran this without a valid `rubyfmt-path` and ensured it was disabled.


jez (approved):

Nitrodist:
Bro ❤️

PR #5884, opened 2022-06-04
NextMethodFinder: Don't need to look in symbol table

jez:
### Motivation

Less code is better.

### Test plan

This is covered by existing tests.

### DIFF
--- a/main/lsp/NextMethodFinder.cc
+++ b/main/lsp/NextMethodFinder.cc
@@ -8,7 +8,7 @@ namespace sorbet::realmain::lsp {
 ast::ExpressionPtr NextMethodFinder::preTransformClassDef(core::Context ctx, ast::ExpressionPtr tree) {
     auto &classDef = ast::cast_tree_nonnull<ast::ClassDef>(tree);
 
-    auto loc = core::Loc(ctx.file, tree.loc());
+    auto loc = ctx.locAt(tree.loc());
 
     if (!this->narrowestClassDefRange.exists()) {
         // No narrowestClassDefRange yet, so take the <root> loc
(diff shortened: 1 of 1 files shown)


main/lsp/NextMethodFinder.cc
```diff
@@ -83,10 +75,7 @@ ast::ExpressionPtr NextMethodFinder::preTransformMethodDef(core::Context ctx, as
 
     if (this->result().exists()) {
         // Method defs are not guaranteed to be sorted in order by their declLocs
-        auto &resultLocs = this->result().data(ctx)->locs();
-        auto maybeResultLoc = absl::c_find_if(resultLocs, inFileOfQuery);
-        ENFORCE(maybeResultLoc != resultLocs.end(), "Must exist, because otherwise it wouldn't have been the result_");
```
froydnj, line 88: How sad are we about losing this?

main/lsp/NextMethodFinder.cc
```diff
@@ -52,20 +52,12 @@ ast::ExpressionPtr NextMethodFinder::preTransformMethodDef(core::Context ctx, as
...
-    }
-
-    auto currentLoc = *maybeCurrentLoc;
+    auto currentLoc = ctx.locAt(tree.loc());
     if (!currentLoc.exists()) {
         // Defensive in case location information is disabled (e.g., certain fuzzer modes)
         return tree;
     }
 
-    ENFORCE(currentLoc.file() == this->queryLoc.file());
```
froydnj, line 68: How sad are we about losing this?

froydnj (approved):

main/lsp/NextMethodFinder.cc
jez: Before, it was non obvious that the Loc would be for the right file because we were searching for locs in the symbol table.

now, it’s correct by construction, we only ever operate on LocOffset, and there is only one file in play, so I figured these enforces were one step removed from asserting something like  1 == 1

main/lsp/NextMethodFinder.cc
jez: Same as above

PR #5885, opened 2022-06-04
Rewrite sig_finder to use an AST, not a string

jez:
### Motivation

I want to depend on this to do some more sophisticated things inside
completion.cc, specifically getting arguments and types.

The code here is modeled directly off of the code for `NextMethodFinder` (used
by the `sig` completion item).

### Test plan

For now, existing automated tests (this feature is already used by the final
method checker and the move method refactoring).


jez:
Assigning this to myself because there's a bug and I don't have time to fix it at the moment. This work is required to unblock #5886, but that PR itself is blocked by #5808, so it's not pressing to land this at the time being.

As a reminder for myself, you'll need to update the `SigFinder` tree walk to not only look at sends, but also look at method defs. Since the tree might not be in the same order anymore, you'll need two "result" ish fields. The current one, for the ParsedSig, and a new one to track the latest method found before the queryLoc. Only add a `sig` method call if it's after the latest method found, and if you ever find a method later than the current sig but still before the queryLoc, clear the current parsed sig.

froydnj:
I think the current sig finder code has basically the same issues, so maybe it's not important to handle, but how does this interact with something like sigs on `attr_accessor` or `prop`, where we have multiple methods and sigs coexisting at the same location?

jez:
> I think the current sig finder code has basically the same issues, so maybe it's not important to handle, but how does this interact with something like sigs on `attr_accessor` or `prop`, where we have multiple methods and sigs coexisting at the same location?

Good question, I'll add some tests.

This is not blocking the fast path method stuff (it's blocking snippet completion for `def foo`) so I'll get to it in the coming weeks.

jez:
> I think the current sig finder code has basically the same issues, so maybe it's not important to handle, but how does this interact with something like sigs on `attr_accessor` or `prop`, where we have multiple methods and sigs coexisting at the same location?

Turns out that this is already tested by `autocorrect-non-final-methods-final-module.rb`

main/sig_finder/sig_finder.cc
```diff
@@ -8,110 +9,133 @@ using namespace std;
...
         }
+    }
+}
+
+void SigFinder::preTransformSend(core::Context ctx, ast::ExpressionPtr &tree) {
+    auto &send = ast::cast_tree_nonnull<ast::Send>(tree);
+
+    if (!resolver::TypeSyntax::isSig(ctx, send)) {
+        return;
+    }
```
aadi-stripe, line 86: A general comment (not necessarily attached to this line) -- I have very little intuition for whether using an AST walk to find `sig`s is slower/faster/about the same as compared to the current string matching method. I trust that this has been tested for perf, though.

aadi-stripe (approved):

main/sig_finder/sig_finder.cc
jez: Some notes:

- This code is only running in places where speed isn't absolutely critical (error messages, LSP requests)
- This code is more correct (source string parsing is inherently error prone)

I have not tested it for performance, again because our normal mechanism for testing this (just run sorbet a bunch) wouldn't exercise this codepath unless we introduced tons of final method errors.

jez: But my intuition is that ASTs are faster, as you can walk the tree without ever having to allocate or do a linear string search.

PR #5886, opened 2022-06-05
wip: POC Suggest arguments after typing method def

jez:
### Motivation

Once we have adding methods taking the fast path, we can do cool things like
this.

- [ ] @jez Build a second option like "(sorbet) Suggest keyword params" to let the user choose whether they want one or the other. Don't set the sort text on the result, so that VS Code can learn the user's preference over time.

### Test plan

- [ ] @jez Test this

Screenshot from 2022-06-04 17-16-36


jez:
Going to close this for the time being. the changes are good; we'll have to consider the UI (keyword or positional?) and write some tests but that can be something for a future project.

PR #5887, opened 2022-06-05
Pass source contents directly to rubyfmt instead of passing file path

reese:
This changes the document formatting LSP logic to pass file contents directly to `rubyfmt` through `stdin` instead of using the file path (see "Motivation" for more on why we need this). As part of Stripe's internal `rubyfmt` rollout, this also adds a flag to "opt-in" formatting directories -- again, more in "Motivation" for why.

### Motivation

In the current Stripe development environment, there's a brief pause between when you save a file and when it gets synced to your devbox. If you enable format on save in VSCode and save before file syncing happens, passing the file path to `rubyfmt` will then run formatting on a not-yet-updated version of the file, which will then make edits and overwrite your changes. However, the LSP has access to the local copy, so this change makes it so that we directly pass the up-to-date source to `rubyfmt` (`rubyfmt` supports reading from stdin like `cat myfile.rb | rubyfmt`) to get around this issue.

The other wrench here is that internally, we previously were gradually rolling out `rubyfmt` behind an allowlist based on filename, but because we're now passing the source instead of the path to the `rubyfmt` binary, we need to be able to selectively format directories, hence the additional option added in the first commit.

### Test plan

This passes existing tests to maintain existing functionality, and I manually tested the allowlist option.

### DIFF
--- a/main/lsp/requests/document_formatting.cc
+++ b/main/lsp/requests/document_formatting.cc
@@ -53,6 +53,18 @@ void DocumentFormattingTask::displayError(string errorMessage, unique_ptr<Respon
         "2.0", LSPMethod::WindowShowMessage, make_unique<ShowMessageParams>(MessageType::Error, errorMessage)));
 }
 
+bool DocumentFormattingTask::documentIsFormattable(string_view path) {
+    if (config.opts.rubyfmtFormattableDirectories.empty()) {
+        return true;
+    }
+    for (string directory : config.opts.rubyfmtFormattableDirectories) {
+        if (path.find(directory) == 0 || path.find(fmt::format("./{}", directory)) == 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
 void DocumentFormattingTask::index(LSPIndexer &index) {
     auto response = make_unique<ResponseMessage>("2.0", id, LSPMethod::TextDocumentFormatting);
     if (!config.opts.lspDocumentFormatRubyfmtEnabled) {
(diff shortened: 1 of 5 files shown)


main/lsp/requests/document_formatting.cc
```diff
@@ -53,6 +53,18 @@ void DocumentFormattingTask::displayError(string errorMessage, unique_ptr<Respon
         "2.0", LSPMethod::WindowShowMessage, make_unique<ShowMessageParams>(MessageType::Error, errorMessage)));
 }
 
+bool DocumentFormattingTask::documentIsFormattable(string_view path) {
+    if (config.opts.rubyfmtFormattableDirectories.empty()) {
+        return true;
+    }
+    for (string directory : config.opts.rubyfmtFormattableDirectories) {
+        if (path.find(directory) == 0 || path.find(fmt::format("./{}", directory)) == 0) {
```
jez, line 61: You almost never want to hardcode `./` in Sorbet, as the prefix in front of the file path will be whatever was passed to the `--dir` flag.

Instead, always use `config.opts.pathPrefix`.

jez (approved):
I still think that this option is going to be clunky long term and something that you're going to want to replace with `# rubyfmt: ...` comments at the top of individual files long term.

If you think that this option will make the rollout easier, I'm fine with it, but I will say that if I personally were doing this, I'd start putting `# rubyfmt:` comments in all the files that have formatting enabled, if only just to make it trivially easy for people to understand why rubyfmt is working in some files and not others, without having to realize that there is some hidden allowlist.

jez:
For what it's worth, with the Sorbet Compiler we didn't do anything to prevent people from putting `# compiled: true` in their files other than just publicly and repeatedly ask them as much. We also set things up to mention our team on all new usages of `# compiled:` comments

PR #5889, opened 2022-06-06
Create newtype wrapper for method arity hashes

jez:
### Motivation

I also took the chance to rename this to arityHash not argument hash,
because first of all, it's hashing parameters, not arguments (arguments
are the things at the call site). We do a pretty terrible job with that
convention throughout the codebase and I've considered taking a whole
pass over the codebase to fix this but I usually end up giving up.
Anyways, one fewer misuse of `args` is good.

I also called it "arity" instead of "parameters" to make it more clear
that this has nothing to do with hashing the parameter types as well,
only with which parameters exist.

### Test plan

No-op change. Covered by existing tests.


froydnj (approved):
++naming!

PR #5891, opened 2022-06-07
Fix bug with alias_method and the fast path

jez:
### Motivation

I thought that this was a bug unique to my changes in #5808 (method changess
taking the fast path), but it turned out to be a bug that reproduces on master.

I'd like to fix it so that I can write some tests that I think should be able to
pass on my other PR.

### Test plan

If you want to see what the failing behavior looked like:
<https://buildkite.com/sorbet/sorbet/builds/22823#01813b92-0ffb-4b4b-9ffb-94bc8abb80f6/139-193>

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -2096,9 +2096,12 @@ void Method::sanityCheck(const GlobalState &gs) const {
                                           " corresponds to a core::Symbols::noTypeArgument()");
     }
     if (isa_type<AliasType>(this->resultType)) {
-        // If we have an alias method, we should never look at it's arguments;
-        // we should instead look at the arguments of whatever we're aliasing.
-        ENFORCE_NO_TIMER(this->arguments.empty(), ref(gs).show(gs));
+        // The arguments of an alias method don't mean anything. When calling a method alias,
+        // we dealias the symbol and use those arguments.
+        //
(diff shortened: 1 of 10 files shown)


core/Symbols.cc
```diff
@@ -2096,9 +2096,12 @@ void Method::sanityCheck(const GlobalState &gs) const {
...
     if (isa_type<AliasType>(this->resultType)) {
-        // If we have an alias method, we should never look at it's arguments;
-        // we should instead look at the arguments of whatever we're aliasing.
-        ENFORCE_NO_TIMER(this->arguments.empty(), ref(gs).show(gs));
+        // The arguments of an alias method don't mean anything. When calling a method alias,
+        // we dealias the symbol and use those arguments.
+        //
+        // This leaves the alias method's arguments vector free for us to stash some information. See resolver.
+        ENFORCE_NO_TIMER(absl::c_all_of(this->arguments, [](const auto &arg) { return arg.flags.isKeyword; }),
+                         ref(gs).show(gs));
```
froydnj, line 2104: `all_of` returns true on empty, do you want to additionally `ENFORCE(!this->arguments.empty())`?

froydnj (approved):

core/Symbols.cc
jez: good catch!

PR #5895, opened 2022-06-07
wip: Failing test for change to `initialize` on fast path

jez:
### Motivation

The fast path currently typechecks too few files right now when changing a
method signature for `initialize`.

### Test plan

See included automated tests.


jez:
Going to fold this into another PR that i have in progress.

PR #5896, opened 2022-06-08
avoid inner loop allocations in block-private variable computation

froydnj:
### Motivation

We shouldn't need to repeatedly allocate this set, which should be a win on large CFGs.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_Lq1SQGUUIuw2pt
→ https://go/builds/bui_Lq1SNqL4W0AK0a

jez (approved):

PR #5900, opened 2022-06-08
Resolve instance variables in blocks using the proc binding

Morriar:
### Motivation

This pull-request does two things:

1. Fix an error in the test_case rewriter logic leading to all `test` calls being rewritten out of the class
2. Fix instance variables resolution in blocks that are bound to another context (uncovered by fixing 1.)

#### Do not rewrite out `test` calls

The current implementation of the TestCase rewriter just [removes all the `test` calls from the file](https://sorbet.run/?arg=--print=rewrite-tree#%23%20typed%3A%20true%0A%0Aclass%20ActiveSupport%3A%3ATestCase%0Aend%0A%0Aclass%20MyTest%20%3C%20ActiveSupport%3A%3ATestCase%0A%20%20test%20%22foo%22%20do%0A%20%20%20%20puts%20%22here%22%0A%20%20end%0Aend%0A) (notice how the printed output doesn't have any mention of the `test` call). 

This is effectively [silencing type errors](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20ActiveSupport%3A%3ATestCase%0Aend%0A%0Aclass%20MyTest%20%3C%20ActiveSupport%3A%3ATestCase%0A%20%20test%20%22foo%22%20do%0A%20%20%20%20SOME_UNDEFINED_CONSTANT.some_undefined_method%0A%20%20end%0Aend%0A).

#### Resolve instance variables in blocks using the proc binding

Once the previous problem was fixed, Sorbet started to report errors about instance variables resolution inside blocks because it didn't take into account the proc binding (especially when binding to `T.attached_class`):

```ruby
# typed: strict

class Foo
  extend T::Sig

  sig { void }
  def initialize
    @attr = T.let(42, Integer)
  end

  sig { params(block: T.proc.bind(T.attached_class).void).void }
  def self.foo(&block); end
    
  foo do
    T.reveal_type(self) # => Foo
    puts @attr # should work since @attr is defined on Foo
  end
end
```

[[sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20void%20%7D%0A%20%20def%20initialize%0A%20%20%20%20%40attr%20%3D%20T.let%2842%2C%20Integer%29%0A%20%20end%0A%0A%20%20sig%20%7B%20params%28block%3A%20T.proc.bind%28T.attached_class%29.void%29.void%20%7D%0A%20%20def%20self.foo%28%26block%29%3B%20end%0A%20%20%20%20%0A%20%20foo%20do%0A%20%20%20%20T.reveal_type%28self%29%20%23%20%3D%3E%20Foo%0A%20%20%20%20puts%20%40attr%20%23%20should%20work%0A%20%20end%0Aend)]

So this would work correctly:

```ruby
# typed: strict

class Foo
  extend T::Sig

  sig { params(block: T.proc.bind(T.attached_class).void).void }
  def self.setup(&block); end

  sig { params(name: String, block: T.proc.bind(T.attached_class).void).void }
  def self.test(name, &block); end

  setup do
    @attr = T.let(42, Integer)
  end
    
  test "some test" do
    T.reveal_type(self) # => Foo
    puts @attr # should work
  end
end
```

[[sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20params%28block%3A%20T.proc.bind%28T.attached_class%29.void%29.void%20%7D%0A%20%20def%20self.setup%28%26block%29%3B%20end%0A%0A%20%20sig%20%7B%20params%28name%3A%20String%2C%20block%3A%20T.proc.bind%28T.attached_class%29.void%29.void%20%7D%0A%20%20def%20self.test%28name%2C%20%26block%29%3B%20end%0A%0A%20%20setup%20do%0A%20%20%20%20%40attr%20%3D%20T.let%2842%2C%20Integer%29%0A%20%20end%0A%20%20%20%20%0A%20%20test%20%22some%20test%22%20do%0A%20%20%20%20T.reveal_type%28self%29%20%23%20%3D%3E%20Foo%0A%20%20%20%20puts%20%40attr%20%23%20should%20work%0A%20%20end%0Aend)] (note that it's currently working because the rewriter rewrites out the `test` calls)

### Test plan

See included automated tests.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -58,7 +58,35 @@ LocalRef global2Local(CFGContext cctx, core::SymbolRef what) {
     return alias;
 }
 
-LocalRef unresolvedIdent2Local(CFGContext cctx, const ast::UnresolvedIdent &id) {
+core::ClassOrModuleRef resolveBlockContext(CFGContext cctx, const core::ClassOrModuleRef outerClassOrModule) {
+    auto methodRef = outerClassOrModule.data(cctx.ctx)->findMethodTransitive(cctx.ctx, cctx.link->fun);
+    if (!methodRef.exists()) {
+        return outerClassOrModule;
+    }
+
+    // Let's find the block definition so we can look at the binding
+    auto &args = methodRef.data(cctx.ctx)->arguments;
+    auto block = absl::c_find_if(args, [&](const auto &arg) { return arg.name == core::Names::blkArg(); });
+    if (!block) {
+        return outerClassOrModule;
+    }
+
+    if (block->rebind == core::Symbols::MagicBindToAttachedClass()) {
+        // The proc is bound to T.attached_class, we bind to it
+        return outerClassOrModule.data(cctx.ctx)->attachedClass(cctx.ctx);
+    } else if (block->rebind == core::Symbols::MagicBindToSelfType()) {
+        // The proc is bound to T.self_type, we do nothing since we're already bound to self type
+        return outerClassOrModule;
+    } else if (block->rebind.exists()) {
+        // The proc is bound to an explicit type, we use it as the receiver context directly
+        return block->rebind;
+    } else {
+        // The proc is not bound to anything, we use the outer class as the receiver context
+        return outerClassOrModule;
+    }
+}
+
+core::ClassOrModuleRef resolveReceiverContext(CFGContext cctx, const ast::UnresolvedIdent &id) {
     core::ClassOrModuleRef klass;
 
     switch (id.kind) {
@@ -71,6 +99,12 @@ LocalRef unresolvedIdent2Local(CFGContext cctx, const ast::UnresolvedIdent &id) 
         case ast::UnresolvedIdent::Kind::Instance:
             ENFORCE(cctx.ctx.owner.isMethod());
             klass = cctx.ctx.owner.owner(cctx.ctx).asClassOrModuleRef();
+
+            if (cctx.link != nullptr) {
+                // We're inside a block, so we may have to change the receiver to the proc's binding
+                klass = resolveBlockContext(cctx, klass);
+            }
+
             break;
         case ast::UnresolvedIdent::Kind::Global:
             klass = core::Symbols::root();
@@ -80,6 +114,11 @@ LocalRef unresolvedIdent2Local(CFGContext cctx, const ast::UnresolvedIdent &id) 
             Exception::notImplemented();
     }
 
+    return klass;
+}
+
+LocalRef unresolvedIdent2Local(CFGContext cctx, const ast::UnresolvedIdent &id) {
+    auto klass = resolveReceiverContext(cctx, id);
     auto sym = klass.data(cctx.ctx)->findMemberTransitive(cctx.ctx, id.name);
     if (!sym.exists()) {
         auto fnd = cctx.discoveredUndeclaredFields.find(id.name);
--- a/rewriter/TestCase.cc
+++ b/rewriter/TestCase.cc
@@ -18,7 +18,6 @@ void TestCase::run(core::MutableContext ctx, ast::ClassDef *klass) {
 
                     if (arg0 && arg0->isString(ctx)) {
                         isTest = true;
-                        continue;
                     }
                 }
             } else if (send->fun == core::Names::setup()) {
--- a/test/testdata/rewriter/test_case.rb
+++ b/test/testdata/rewriter/test_case.rb
@@ -33,6 +33,10 @@ class MyTest < ActiveSupport::TestCase
   test "block is evaluated in the context of an instance" do
     assert true
   end
+
+  test "setup instance variables are declared" do
+    assert @a
+  end
 end
 
 class NoMatchTest < ActiveSupport::TestCase
--- a/test/testdata/rewriter/test_case.rb.rewrite-tree.exp
+++ b/test/testdata/rewriter/test_case.rb.rewrite-tree.exp
@@ -49,6 +49,18 @@ class <emptyTree><<C <root>>> < (::<todo sym>)
 
     <self>.extend(<emptyTree>::<C T>::<C Sig>)
 
+    <self>.test("valid method call") do ||
+      <emptyTree>
+    end
+
+    <self>.test("block is evaluated in the context of an instance") do ||
+      <self>.assert(true)
+    end
+
+    <self>.test("setup instance variables are declared") do ||
+      <self>.assert(@a)
+    end
+
     <runtime method definition of initialize>
 
     <runtime method definition of initialize>
@@ -125,6 +137,10 @@ class <emptyTree><<C <root>>> < (::<todo sym>)
     <runtime method definition of self.test>
 
     <runtime method definition of assert_equal>
+
+    <self>.test("it works") do ||
+      <self>.assert_equal(1, @a)
+    end
 
     <runtime method definition of initialize>
   end


cfg/builder/builder_walk.cc
```diff
@@ -327,7 +367,8 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                 } else if (auto lhsLocal = ast::cast_tree<ast::Local>(a.lhs)) {
                     lhs = cctx.inWhat.enterLocal(lhsLocal->localVariable);
                 } else if (auto ident = ast::cast_tree<ast::UnresolvedIdent>(a.lhs)) {
-                    lhs = unresolvedIdent2Local(cctx, *ident);
+                    auto assign = true;
+                    lhs = unresolvedIdent2Local(cctx, *ident, assign);
```
Morriar, line 371: I added the `assign` parameter to avoid reporting two errors in the case the ident is indeed unresolved:

```
test.rb:49: The instance variable @bar must be declared inside initialize or declared nilable https://srb.help/5005
    49 |    @bar = T.let(1, Integer)
            ^^^^

test.rb:49: Use of undeclared variable @bar https://srb.help/6002
    49 |    @bar = T.let(1, Integer)
            ^^^^
```

cfg/builder/builder_walk.cc
```diff
@@ -58,7 +58,35 @@ LocalRef global2Local(CFGContext cctx, core::SymbolRef what) {
     return alias;
 }
 
-LocalRef unresolvedIdent2Local(CFGContext cctx, const ast::UnresolvedIdent &id) {
+core::ClassOrModuleRef resolveBlockContext(CFGContext cctx, const core::ClassOrModuleRef outterClassOrModule) {
```
froydnj, line 61: Nit: this should be `outerClassOrModule` (only one `t` in `outer`) all throughout this change.

cfg/builder/builder_walk.cc
```diff
@@ -71,6 +99,12 @@ LocalRef unresolvedIdent2Local(CFGContext cctx, const ast::UnresolvedIdent &id)
         case ast::UnresolvedIdent::Kind::Instance:
             ENFORCE(cctx.ctx.owner.isMethod());
             klass = cctx.ctx.owner.owner(cctx.ctx).asClassOrModuleRef();
+
+            if (cctx.link != nullptr) {
+                // We're inside a block, so we may have to change the receiver to the proc's binding
+                klass = resolveBlockContext(cctx, klass);
+            }
```
froydnj, line 106: There's no reason to treat instance variables specially here; class variables should get the same treatment, shouldn't they?

cfg/builder/builder_walk.cc
froydnj: I don't really understand the cases in which `assign` is necessary: avoiding reporting errors in the case where we can't resolve the variable seems like a bad thing (we should be telling the user about issues?).  What are the particular mechanics of this change that require adding `assign`?

cfg/builder/builder_walk.cc
```diff
@@ -58,7 +58,35 @@ LocalRef global2Local(CFGContext cctx, core::SymbolRef what) {
     return alias;
 }
 
-LocalRef unresolvedIdent2Local(CFGContext cctx, const ast::UnresolvedIdent &id) {
+core::ClassOrModuleRef resolveBlockContext(CFGContext cctx, const core::ClassOrModuleRef outterClassOrModule) {
+    auto methodRef = outterClassOrModule.data(cctx.ctx)->findMethodTransitive(cctx.ctx, cctx.link->fun);
```
froydnj, line 62: This change seems not great from a performance perspective.  I don't know what better options look like, but `findMethodTransitive` use is a red flag.

froydnj (changes requested):
Initial thoughts below; it looks like there are also some test failures.

froydnj:
Also, when you're doing this method name lookup, you're doing it in the context of the owner of the CFG, which is the class that owns the current method being translated into the CFG.  So you'll get a method defined on that owner, and the block information for that method.

But really, that only works for method calls with a `self` receiver, right?  In the general case, you'd need to do the lookup in the context of the type of the receiver, which we don't have yet.  So therefore it should be possible to construct testcases which would fail with the current scheme?

Or am I just not thinking straight because I haven't had my morning coffee?

Morriar:
> So therefore it should be possible to construct testcases which would fail with the current scheme?

Good point, this would fail for example:

```rb
class IVars
  extend T::Sig

  sig { void }
  def initialize
    @a = T.let(true, T::Boolean)
  end

  sig { params(block: T.proc.bind(IVars).void).void }
  def self.proc_with_explicit_bind_to_instance(&block); end
end

IVars.proc_with_explicit_bind_to_instance do
  T.reveal_type(@a)
end
```

I think to resolve this properly we need to take into account the receiver of the call. In that case it would be stored in the 
SendAndBlockLink dispatch result (`cctx.link->result`) but this will be filled in only during `calls.cc` I believe.

I'm not sure how to approach this. We can't move this step later because we need it to resolve the calls. But we need the resolved calls to complete this step properly as well.

Any idea?

froydnj:
> I'm not sure how to approach this. We can't move this step later because we need it to resolve the calls. But we need the resolved calls to complete this step properly as well.
> 
> Any idea?

Yeah, there's some real chicken-and-egg stuff here.

I haven't thought about this deeply, but my initial thought is (maybe via a new `Instruction` kind) to try and defer the errors about undeclared fields until type inference actually happens, when we have accurate information about the types under which blocks are being executed.  It might be a little gross to only create those `Instruction` nodes when you are referencing instance/class variables in a block, but blocks are weird, so maybe this is justifiable.

So for `DeferredAliasInBlock` things, or some such, you'd need to first figure out the context in which you're executing the block, then lookup the relevant thing in the owner's context.  If you find it, great, figure out the type as appropriate (just like we do for `Alias` on fields currently), otherwise you should issue an error.  (Maybe the node also stores a boolean indicating whether this was the first reference to the instance/class variable and you only emit an error on the first such reference, so you can avoid spamming the user with errors.)

I haven't thought very hard about how this interacts with more complex receiver types, so writing more tests would probably be very helpful for this.

jez:
I haven't read through this PR entirely yet, but I have one question/request:

We're keeping things such that the value of a `T.proc.bind` on a block is never used for the purpose of **declaring** an instance variable, right?

For example, Sorbet currently does the opposite of what Ruby thinks will happen when presented with this example:

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Arequire%20'sorbet-runtime'%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Bparams%28blk%3A%20T.proc.bind%28T.attached_class%29.void%29.void%7D%0A%20%20def%20self.on_instance%28%26blk%29%0A%20%20%20%20puts%20'here'%0A%20%20%20%20%40registered%20%3D%20T.let%28blk%2C%20T.nilable%28T.proc.void%29%29%0A%20%20end%0A%0A%20%20sig%20%7Breturns%28T.nilable%28T.proc.void%29%29%7D%0A%20%20def%20self.registered%0A%20%20%20%20%40registered%0A%20%20end%0A%0A%20%20sig%20%7Bparams%28registered%3A%20T.nilable%28T.proc.void%29%29.returns%28T.nilable%28T.proc.void%29%29%7D%0A%20%20def%20self.registered%3D%28registered%29%0A%20%20%20%20%40registered%20%3D%20T.let%28registered%2C%20T.nilable%28T.proc.void%29%29%0A%20%20end%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20initialize%0A%20%20%20%20registered%20%3D%20self.class.registered%0A%20%20%20%20if%20registered%0A%20%20%20%20%20%20puts%20'in%20initialize'%0A%20%20%20%20%20%20self.instance_eval%28%26registered%29%0A%20%20%20%20end%0A%20%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20on_instance%20do%0A%20%20%20%20puts%20'yep'%0A%20%20%20%20%40x%20%3D%20T.let%2842%2C%20T.nilable%28Integer%29%29%0A%20%20end%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20example1%0A%20%20%20%20p%28%40x%29%20%23%20%3D%3E%2042%0A%20%20end%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20self.example2%0A%20%20%20%20T.reveal_type%28%40x%29%0A%20%20%20%20p%28%40x%29%20%23%20%3D%3E%20nil%0A%20%20end%0Aend%0A%0Achild%20%3D%20Child.new%0Achild.example1%0AChild.example2%0A%0A%0A)

```ruby
  on_instance do
    puts 'yep'
    @x = T.let(42, T.nilable(Integer))
  end
```

^ these are the relevant lines.

In order to prevent cycles between passes, I'd love for us to continue to require that instance variable declarations enter a new `Field` symbol on their syntactically enclosing owner.

For the TestCase rewriter, this would mean that the rewriter would still have to translate `setup do` blocks to `def initialize` methods on the class, similar to what the Minitest rewriter does, so that Sorbet syntactically sees the instance variable with the intended owner.

I'm guessing (hoping) that this doesn't change in this PR but I wanted to call it out to be sure.

Morriar:
For context I reverted the original change [here](https://github.com/sorbet/sorbet/pull/5905) while we figure this out.

I'll close this PR until then.

PR #5901, opened 2022-06-09
Fix return type of Regexp::compile

duelinmarkers:
It appears Sorbet thinks `Regexp::compile` returns `T.class_of(Regexp)` rather than a `Regexp` instance. I think `T.self_type` is [working as intended](https://sorbet.org/docs/class-of) and should not be used on pseudo-constructor class methods.

I currently work around this by doing `T.cast(Regexp.compile(s), Regexp)`.

### Motivation

Core rbis should be complete and correct.

### Test plan

This change didn't cause any `test test` failures locally, and I didn't find anything that looked like a test suite for the core rbis, so I haven't touched any tests yet. I'm happy to do so if someone can point me at The Way (or if I see an automated test failure in this PR that I can understand).

### DIFF
--- a/test/testdata/rbi/regexp.rb
+++ b/test/testdata/rbi/regexp.rb
@@ -9,3 +9,7 @@ end
 
 T.reveal_type(/foo/.match?('foo'))  # error: type: `T::Boolean`
 T.reveal_type(/foo/.match?('foo', 1))  # error: type: `T::Boolean`
+
+T.reveal_type(Regexp.compile('foo')) # error: type: `Regexp`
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Thanks for the patch!

If you wanted to add a test or two to https://github.com/sorbet/sorbet/blob/master/test/testdata/rbi/regexp.rb , that would be appreciated.  Ideally the form of the tests should be cargo-cultable, but please let me know if you have any questions.  I'll wait to merge this PR for a day or two to give you time to add tests if you want.

duelinmarkers:
@froydnj, can you unblock the tests again?

froydnj (approved):
Looks great, thank you for the tests!

PR #5906, opened 2022-06-10
improve comparison methods for `{Expression,Instruction}Ptr`

froydnj:
### Motivation

There's no need to add extraneous shifts in these methods, because if the pointers are the same (resp. different) they will have the same (resp. different) tags, so we can just compare the whole tagged pointer.  Note that this argument also works for `nullptr` being stored in these (i.e. 0), because our tags start at 1, so any tagged thing will compare not-equal to `nullptr`.

This change enables `typecase` over the pointers to be compiled as a jump table, which wasn't happening before (!).  I think this is because the `nullptr_t` overloads enable things inside clang to be smarter?  (I didn't test if the jump tables happened without the `nullptr_t` overloads, but I think they are pretty common for smart pointers and consistency is a good thing.)

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -91,6 +91,9 @@ private:
     template <typename E, typename... Args> friend ExpressionPtr make_expression(Args &&...);
 
     static tagged_storage tagPtr(Tag tag, void *expr) {
+        ENFORCE(static_cast<size_t>(tag) != 0);
+        ENFORCE(expr != nullptr);
+
         // Store the tag in the lower 16 bits of the pointer, regardless of size.
         auto val = static_cast<tagged_storage>(tag);
         auto maskedPtr = reinterpret_cast<tagged_storage>(expr) << 16;
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -194,15 +194,23 @@ public:
     }
 
     explicit operator bool() const noexcept {
-        return get() != nullptr;
+        return ptr != 0;
     }
 
     bool operator==(const ExpressionPtr &other) const noexcept {
-        return get() == other.get();
+        return ptr == other.ptr;
+    }
+
+    bool operator==(std::nullptr_t) const noexcept {
+        return ptr == 0;
     }
 
     bool operator!=(const ExpressionPtr &other) const noexcept {
-        return get() != other.get();
+        return ptr != other.ptr;
+    }
+
+    bool operator!=(std::nullptr_t) const noexcept {
+        return ptr != 0;
     }
 
     ExpressionPtr deepCopy() const;
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -69,6 +69,8 @@ enum class Tag {
 // A mapping from tree type to its corresponding tag.
 template <typename T> struct ExpressionToTag;
 
+class EmptyTree;
+
 class ExpressionPtr {
 public:
     // We store tagged pointers as 64-bit values.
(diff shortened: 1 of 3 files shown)


ast/Trees.h
```diff
@@ -194,15 +194,23 @@ class ExpressionPtr {
...
+        return ptr == 0;
     }
 
     bool operator!=(const ExpressionPtr &other) const noexcept {
-        return get() != other.get();
+        return ptr != other.ptr;
+    }
+
+    bool operator!=(std::nullptr_t) const noexcept {
+        return ptr != 0;
```
jvilk-stripe, line 213: Ditto.

ast/Trees.h
```diff
@@ -194,15 +194,23 @@ class ExpressionPtr {
...
     }
 
     bool operator==(const ExpressionPtr &other) const noexcept {
-        return get() == other.get();
+        return ptr == other.ptr;
+    }
+
+    bool operator==(std::nullptr_t) const noexcept {
+        return ptr == 0;
```
jvilk-stripe, line 205: Ditto.

ast/Trees.h
```diff
@@ -194,15 +194,23 @@ class ExpressionPtr {
     }
 
     explicit operator bool() const noexcept {
-        return get() != nullptr;
+        return ptr != 0;
     }
 
     bool operator==(const ExpressionPtr &other) const noexcept {
-        return get() == other.get();
+        return ptr == other.ptr;
```
jvilk-stripe, line 201: These pointer comparison operations seem valid to me; I don't think it would be legal for us to have two tagged pointers pointing to the same thing with different tags.

cfg/Instructions.h
```diff
@@ -362,14 +362,20 @@ class InstructionPtr final {
     }
 
     explicit operator bool() const noexcept {
-        return get() != nullptr;
+        return ptr != 0;
```
jvilk-stripe, line 365: Same note RE:null applies.

jvilk-stripe (changes requested):
Thank you for finding this! I have a concern about the possibility of tagged nullptrs. I'd feel better knowing that it isn't possible and that there were tests validating that.

ast/Trees.h
```diff
@@ -194,15 +194,23 @@ class ExpressionPtr {
     }
 
     explicit operator bool() const noexcept {
-        return get() != nullptr;
+        return ptr != 0;
```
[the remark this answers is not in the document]
froydnj: We reset to nullptr when we move:

https://github.com/sorbet/sorbet/blob/615a299e1e6277bf3941ed0dffcdf7a1a81e41f5/ast/Trees.h#L147-L158
https://github.com/sorbet/sorbet/blob/615a299e1e6277bf3941ed0dffcdf7a1a81e41f5/ast/Trees.h#L105-L110

but that does not preclude the possibility of getting constructed with a nullptr and a bogus tag.  I guess we could add an `ENFORCE` on the tagging constructor or on `get()` to ensure that we're not winding up with weird things.  (Though the only place we call the tagging constructor is in `make_tree` and we cannot pass in a nullptr there because `new` would have thrown otherwise.)

(Could we have once upon a time wound up with tagged nullptr before we had `EmptyTree`?)

froydnj:
I have added some extra `ENFORCE`s and closed some holes that could sneak in tagged nullptr things.  Please let me know if you have further concerns.

jvilk-stripe (approved):
Thanks; this makes me feel better. This code likely predates `EmptyTree`, too.

PR #5907, opened 2022-06-10
Improve certain "limitations of flow sensitivity" errors

jez:
### Motivation

This is by far the most commonly asked question by people new to Sorbet. Having a dedicated error message that links to more thorough documentation will substantially reduce our support burden and encourage people to learn more about how best to use Sorbet.

### Commit summary

If you'd like, use the commits to see the before/after behavior of the new tests

- **pre-work: Remove some dead code** (57fbe494c)

  I just happened to notice these, I figured I'd remove them while I was
  here.

- **pre-work: Remove useless argument from this function** (482421436)

  It's not actually required, so no reason to pass it.

- **Add test showing behavior before** (aa53cd078)

- **Detect certain "limitations of flow sensitivity" errors** (954bd0625)

- **Show new behavior** (671ff734b)

### Test plan

See automated tests.

image

### DIFF
--- a/cfg/CFG.cc
+++ b/cfg/CFG.cc
@@ -292,6 +292,64 @@ string CFG::showRaw(core::Context ctx) const {
     return to_string(buf);
 }
 
+namespace {
+const cfg::Send *looksLikeUpdateKnowledgeSend(const cfg::CFG &inWhat, const cfg::Binding &bind, LocalRef expectedRecv) {
+    if (bind.bind.variable != expectedRecv) {
+        return nullptr;
+    }
+
+    auto send = cfg::cast_instruction<cfg::Send>(bind.value);
+    if (send == nullptr) {
+        return nullptr;
+    }
+
+    if (!send->recv.variable.isSyntheticTemporary(inWhat)) {
+        // Conservative heuristic. If the receiver is not synthetic, then the update knowledge
+        // method was not computed by some sub expression but instead called directly on a
+        // Ruby-level variable. Therefore it wouldn't be useful to report a "maybe factor this to a
+        // variable" error for this call site.
+        return nullptr;
+    }
+
+    if (!send->fun.isUpdateKnowledgeName()) {
+        return nullptr;
+    }
+
+    return send;
+}
+
+} // namespace
+
+optional<BasicBlock::BlockExitCondInfo> BasicBlock::maybeGetUpdateKnowledgeReceiver(const cfg::CFG &inWhat) const {
+    if (this->exprs.empty()) {
+        return nullopt;
+    }
+
+    // Conservative heuristic, maybe we can make this smarter
+    // (Currently, only detect cases where the branch condition was the last thing computed)
+    auto &lastBinding = this->exprs.back();
+    auto send = looksLikeUpdateKnowledgeSend(inWhat, lastBinding, this->bexit.cond.variable);
+    if (send == nullptr) {
+        return nullopt;
+    }
+
+    if (send->fun == core::Names::bang() && this->exprs.size() >= 2) {
+        // Heuristic, because it's overwhelmingly common to see `!x.foo.nil?`, in which case the
+        // relevent receiver is the second-last, not the last, send binding.
+        auto bangRecv = send->recv.variable;
+        auto &secondLastBinding = this->exprs[this->exprs.size() - 2];
+        auto secondLastSend = looksLikeUpdateKnowledgeSend(inWhat, secondLastBinding, bangRecv);
+
+        if (secondLastSend != nullptr) {
+            send = secondLastSend;
+        }
+    }
+
+    // VariableUseSite copy constructor is deleted, have to manually make a shallow copy
+    auto recv = VariableUseSite{send->recv.variable, send->recv.type};
+    return make_optional<BlockExitCondInfo>(move(recv), send->receiverLoc, send->fun);
+}
+
 string BasicBlock::toString(const core::GlobalState &gs, const CFG &cfg) const {
     fmt::memory_buffer buf;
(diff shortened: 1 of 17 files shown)


infer/inference.cc
```diff
@@ -86,6 +86,21 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
...
+        // contain the current binding being processed).
+        //
+        // That means that only here in inference do we know about the structure of the CFG.
+        // But we want to pluck some limited information about the parent block and pass it into
+        // processBinding, only so that some errors that would already be reported can be improved.
+        // parentUpdateKnowledgeReceiver is such a thing.
+        //
+        // In general we should carefully limit the set of information that processBinding requires,
+        // to make it easy to reason about correctness.
+        optional<cfg::BasicBlock::BlockExitCondInfo> parentUpdateKnowledgeReceiver;
```
froydnj, line 102: There are precious few comments in Sorbet explaining *why* things are done a certain way; thank you for adding this one.

I don't know that I am convinced that trying to limit what information we pass in is beneficial.  I can see the argument for it, but trying to avoid passing the necessary non-local information that we could use just in the error case means that we wind up computing this information in many many cases where it's never going to get used.  Have you tried benchmarking this change wrt typechecking performance?

froydnj (changes requested):
I am curious what the benchmarking results look like for this one before it goes in, and whether we could mitigate any negative effects by simply passing the parent block in the single-backedge case and move `maybeGetUpdateKnowledgeReceiver`'s work solely into the error case.

jez:

→ https://go/builds/bui_LxGMYxC7VYzfyh
→ https://go/builds/bui_LxGMoiiRLlWyuv

test/cli/flow-sensitivity-hint/test.out
```diff
@@ -0,0 +1,219 @@
...
+    31 |  x.first.even?
+          ^^^^^^^
+
+test/cli/flow-sensitivity-hint/test.rb:36: Method `even?` does not exist on `NilClass` component of `T.nilable(Integer)` https://srb.help/7003
+    36 |  first_is_even = !x.first.nil? && x.first.even?
+                                                   ^^^^^
+  Got `T.nilable(Integer)` originating from:
+    test/cli/flow-sensitivity-hint/test.rb:36:
+    36 |  first_is_even = !x.first.nil? && x.first.even?
+                                           ^^^^^^^
```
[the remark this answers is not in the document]
jez: - the intrinsic for `<check-and-and>` fires before the check in environment.cc this PR adds, and it also changes the error class. the `<check-and-and>` intrinsic changes `UnknownMethod` errors to `CallAfterAndAnd` errors, meaning that the environment.cc thing never trips for `&&` calls

- the and-and intrinsic is still useful because in that case, we have a very easy autocorrect to suggest, which is to rewrite the call as `x.first&.even?` instead of `x.first && x.first.even?`. it's not easy to know how to build this autocorrect in the general case

- Because the call to `<check-and-and>` is generated by syntactically looking for Ruby's `&&` construct, and not via any non-Symbol-based hacks, it's much more reliable than the mechanism in this PR, which is why i'm comfortable giving it its own error code. the changes in this PR will do the wrong thing for certain cases, so i think it's more prudent to leave the error message as a section on an addition error, rather than attempting to completely throw away the original `UnknownMethod` error.

cfg/CFG.cc
```diff
@@ -292,6 +292,64 @@ string CFG::showRaw(core::Context ctx) const {
...
+        auto secondLastSend = looksLikeUpdateKnowledgeSend(inWhat, secondLastBinding, bangRecv);
+
+        if (secondLastSend != nullptr) {
+            send = secondLastSend;
+        }
+    }
+
+    // Constructor is deleted, have to manually make a shallow copy
+    auto recv = VariableUseSite{send->recv.variable, send->recv.type};
+    return make_optional<BlockExitCondInfo>(move(recv), send->receiverLoc, send->fun);
```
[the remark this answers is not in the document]
jez: yeah i was referring to the `VariableUseSite` constructor being deleted, so we couldn't just write

```cpp
return make_optional<BlockExitCondInfo>(send->recv, send->receiverLoc, send->fun);
```

i have a slight preference for the current style as it looks better with clang-format.

infer/inference.cc
jez: After a quick test, it seems that maybe this adds 0.29% time to pay-server master, which we think is basically within the margin of error of noise

from discussion offline: an alternative to avoid computing this except when needed might be to pass an opaque handle into `processBinding` that only exposes one helper method which computes the relevant information and is called only when we're sure that processBinding (or even dispatchCallSymbol) needs it. we thought that amounted to quite a lot of ceremony, all things considered.

we did both generally agree that it is a nice property of `processBinding` that it can't currently do things like lookup types of variables in parent blocks (i.e. in previous Environments), which makes it easier to reason about what can possibly affect how an individual binding typechecks or not, but we did think that there might arise future features which want to thread more information down into processBinding, at which point we might want to reconsider how we pass ad-hoc information down from inference.cc into environment.cc

test/cli/flow-sensitivity-hint/test.rb
```diff
@@ -0,0 +1,147 @@
...
+sig {params(x: T::Array[Integer]).void}
+def bad_example6(x)
+  if 0 == x.first # Should have special message, but doesn't (bug, annoying to fix)
+    x.first.even?
+  end
+end
+
+sig {params(x: T::Array[Integer]).void}
+def bad_example7(x)
+  if Integer === x.first # Should have sepcial message, but doesn't (bug, annoying to fix)
```
froydnj, line 136:
```suggestion
  if Integer === x.first # Should have special message, but doesn't (bug, annoying to fix)
```

froydnj (approved):

PR #5911, opened 2022-06-13
add `Array` and `Hash` cfg instructions

froydnj:
### Motivation

These instructions are significantly smaller than representing array and hash construction as `Send`s, require less information at construction, and require less upkeep during inference.  Should be a win all around!

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_Lry5uZapus4tVd
→ https://go/builds/bui_Lry50O43tLPCmQ

PR #5913, opened 2022-06-14
make `ExpressionPtr` `trivial_abi` for faster argument passing

froydnj:
### Motivation

`ExpressionPtr`, due to ABI rules, needs to be passed in memory even though it is small enough that we could pass it in a register.  And since we do a lot of unnecessary passing of `ExpressionPtr` values due to treemap, this change should be beneficial.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -71,7 +71,7 @@ template <typename T> struct ExpressionToTag;
 
 class EmptyTree;
 
-class ExpressionPtr {
+class __attribute__((trivial_abi)) ExpressionPtr final {
 public:
     // We store tagged pointers as 64-bit values.
     using tagged_storage = uint64_t;
--- a/ast/Trees.h
+++ b/ast/Trees.h
@@ -71,7 +71,16 @@ template <typename T> struct ExpressionToTag;
 
 class EmptyTree;
 
-class __attribute__((trivial_abi)) ExpressionPtr final {
+// When we added trivial_abi support, the version of Emscripten that we used did not
+// support trivial_abi.  Even if we upgraded to a version that did, the performance
+// gains of trivial_abi are not super-important for Emscripten'd Sorbet.
+#ifdef __EMSCRIPTEN__
+#define TRIVIAL_ABI
+#else
+#define TRIVIAL_ABI __attribute__((trivial_abi))
+#endif
+
+class TRIVIAL_ABI ExpressionPtr final {
 public:
     // We store tagged pointers as 64-bit values.
     using tagged_storage = uint64_t;


froydnj:

→ https://go/builds/bui_LsKVGBzhcRTej8
→ https://go/builds/bui_LsKVNmTn7WpoO1

froydnj:
> Forgive me; I don't completely understand what this attribute does or if it is ever unsafe to apply.
> 
> My understanding is that this change allows `ExpressionPtr` to be passed by value in a register rather than by reference.

`ExpressionPtr` values to be passed in a register, yes.

> `ExpressionPtr`'s destructor deletes the pointer and mutates the `ExpressionPtr` to reset it to 0 (iirc). My FUD is that if it is passed by value and destructed, it doesn't mutate the value of `ExpressionPtr` in the caller and then we have a dangling pointer.

There are two interesting cases; for:

```c++
void by_value(ExpressionPtr v) {...}
```

the caller has already had to `std::move` into the callee, so the caller's `ExpressionPtr` is already 0.  So the callee can safely call the destructor without any issues, since the callee has ownership.

```c++
void by_rvalue_ref(ExpressionPtr &&v) {...}
```

When calling this function, the caller has expressed intent to relinquish ownership by passing via rvalue reference, but the callee does not have to take ownership.  So the address of `v` is passed to the function, and the callee is free to move out the reference or not; the caller will destruct as normal.

Nothing changes when passing by lvalue reference.

> Does this attribute change those semantics in any way when `ExpressionPtr` is passed by value? Can this attribute ever be unsafe to apply?

It does potentially change [the ordering of destructor calls](https://quuxplusone.github.io/blog/2018/05/02/trivial-abi-101/#caveat-3-destructor-ordering) in cases with multiple parameters that are defined as `trivial_abi`, but I don't think we need to worry about that here.

jvilk-stripe (approved):
I am satisfied with the explanation. Could you benchmark this change before landing? If it doesn't move the needle perf-wise, I wonder if it's worth doing given the technical complexity of this attribute.

froydnj:
> I am satisfied with the explanation. Could you benchmark this change before landing? If it doesn't move the needle perf-wise, I wonder if it's worth doing given the technical complexity of this attribute.

This change gains ~2% on indexing performance (`indexSuppliedFiles.value`).  It's possible the better path forward here is to rewrite `TreeMap` to have not-in-place (the current behavior) *and* in-place (new functionality) modifications (+ a straight non-modifying walk over the tree mode), and then modify clients appropriately, which would probably have ~the same effect.  WDYT?

jvilk-stripe:
@froydnj Seems like a small improvement. I'm peaceful with merging. Can you add a comment in the source code indicating the 2% perf improvement so folks in the future know that removing this (in case it causes problems) wouldn't be the worst thing?

PR #5914, opened 2022-06-14
avoid reallocating nodes for `UnresolvedConstantLit` in `SubstWalk`

froydnj:
### Motivation

There's no need to allocate a new node here when we have a perfectly good one all ready to be operated on.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_LsMM4baods6pRU
→ https://go/builds/bui_LsMMl4mxDZ4g6C

PR #5916, opened 2022-06-14
Fix metrics counter for singleton classes

Morriar:
### Motivation

With the previous behavior, this metrics was incremented for each singleton class of each synthetic class created at GlobalState initialization. This was giving >100 singleton classes for an empty set of files.

An interesting fact is that the GlobalState is initialized with all the synthetic classes only when passing more than 2 files so
it needs an index. So we need at least 3 empty files to reproduce this.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -1616,7 +1616,6 @@ ClassOrModuleRef ClassOrModule::singletonClass(GlobalState &gs) {
     singleton = gs.enterClassSymbol(this->loc(), this->owner, singletonName);
     ClassOrModuleData singletonInfo = singleton.data(gs);
 
-    prodCounterInc("types.input.singleton_classes.total");
     singletonInfo->members()[Names::attached()] = selfRef;
     singletonInfo->setSuperClass(Symbols::todo());
     singletonInfo->setIsModule(false);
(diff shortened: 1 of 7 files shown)


froydnj (commented):
Why is it incorrect to count the singleton classes for the synthetic classes?  Those are just as much singleton classes as anything else, right?

Morriar:
In the current state it doesn't make much sense:
* You run Sorbet with 1 empty file: 0 singleton classes
* You run Sorbet with 3 empty files: 102 singleton classes

Even if we fix this, you run Sorbet with 1 empty file: 0 singleton classes.
Would that make sense when we do not count classes coming from RBI payload for example (so 0 modules and 0 classes)?

froydnj:
Ah, so the complaint is not so much that we are counting them, it's that we are counting them at the wrong place and that leads to weird discrepancies based on the details of how Sorbet is invoked?

Morriar:
And also that we are counting them when we initialize the GlobalState.

For example we will count the singleton classes for (but not limited to) T, Class, BasicObject, Kernel, Module, Class, Array. Yet, we do not count them as modules and classes.

So you get 102 singleton classes, 0 modules and 0 classes for an empty program.

froydnj (approved):
Thanks!  (I think I had looked at that module/class counting code at one point and wondered why singleton classes weren't being counted there...)

PR #5917, opened 2022-06-14
devirtualize `parser::Node`

froydnj:
### Motivation

Was curious how much of a difference this would make, especially since we get to turn desugar into a jump table...or at least we would if clang was a little smarter.  Will have to look into that later.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_LsRzEYTSa13VYA
→ https://go/builds/bui_LsRzf4k1zAgz2S

PR #5919, opened 2022-06-17
provide a better `TreeWalk` abstraction in addition to `TreeMap`

froydnj:
### Motivation

`TreeMap` is a nice abstraction, but its functional nature makes it relatively slow (though one can argue that the compiler really ought to be smart enough to eliminate the functional abstraction penalty; alas, it is not).  We can provide a better abstraction that doesn't have the functional abstraction penalties, and move things over incrementally.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -254,17 +254,16 @@ private:
     public:
         bool seenUnresolved = false;
 
-        ast::ExpressionPtr postTransformConstantLit(core::Context ctx, ast::ExpressionPtr tree) {
+        void postTransformConstantLit(core::Context ctx, ast::ExpressionPtr &tree) {
             auto &original = ast::cast_tree_nonnull<ast::ConstantLit>(tree);
             seenUnresolved |= !isAlreadyResolved(ctx, original);
-            return tree;
         };
     };
 
     static bool isFullyResolved(core::Context ctx, const ast::ExpressionPtr &expression) {
         ResolutionChecker checker;
         ast::ExpressionPtr dummy(expression.getTagged());
-        dummy = ast::TreeMap::apply(ctx, checker, std::move(dummy));
+        ast::TreeWalk::apply(ctx, checker, dummy);
         ENFORCE(dummy == expression);
         dummy.release();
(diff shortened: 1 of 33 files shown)


froydnj:

→ https://go/builds/bui_LtThx3Or9HhLWs
→ https://go/builds/bui_LtThCE6XJd3oPa

froydnj:

→ https://go/builds/bui_LtUdoxP9rsHAHZ
→ https://go/builds/bui_LtUd1ykFmXxkXK

froydnj:

→ https://go/builds/bui_LtVWvenCU2BtSv
→ https://go/builds/bui_LtVWLyBK1Ag739

froydnj:

→ https://go/builds/bui_LtXQAKB12NrpJk
→ https://go/builds/bui_LtXQo6kzI6pBX4

jez:
ooc, does this make sorbet take substantially longer to compile?

froydnj:
> ooc, does this make sorbet take substantially longer to compile?

I compiled Sorbet after almost every commit in this series and noticed no substantial (e.g. the long pole of resolver took basically the same amount of time, etc.) change at any point.

resolver/resolver.cc
```diff
@@ -2799,7 +2785,8 @@ class ResolveTypeMembersAndFieldsWalk {
                         todoResolveCastItems_.emplace_back(move(item));
                     }
 
-                    return ast::MK::InsSeq1(send.loc, move(typeExpr), move(cast));
+                    tree = ast::MK::InsSeq1(send.loc, move(typeExpr), move(cast));
+                    return;
```
jez, line 2789: This drops the old one before overwriting it right?

This API doesn't even seem that annoying. I was expecting to not like it that much, but I think not having to write `return tree` everywhere to early exit is actually totally good compensation.

jez (approved):

resolver/resolver.cc
froydnj: That's correct; this line just going to invoke `ExpressionPtr::operator=(ExpressionPtr &&)`, which handles the memory management.

PR #5921, opened 2022-06-18
reorder class flattening and definition validation for speed

froydnj:
### Motivation

We should run definition validation after class flattening, so definition validation doesn't have to wade through a bunch of top-level stuff in classes.

### Test plan

See included automated tests.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -662,9 +662,9 @@ void typecheckOne(core::Context ctx, ast::ParsedFile resolved, const options::Op
         return;
     }
 
-    resolved = definition_validator::runOne(ctx, std::move(resolved));
-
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_Lte98D0Ap25YX9
→ https://go/builds/bui_Lte9rxgXhhtOkm

jez (approved):
Interesting, it looks like when we populate the `ancestors`/`singletonAncestors` lists on a `ClassDef` in namer, we deepCopy the ancestor arg into the vector.

We only read out of the `ancestors`/`singletonAncestors` list in definition_validator when we need the tree for an ancestor, and never look for individual calls to `include`/`extend`. So it doesn't matter if we hide the `include`/`extend` calls themselves in `<static-init>` via class_flatten before running definition_validator.

Nice find.

PR #5924, opened 2022-06-20
Packaging check should report an error if an exported constant is only declared in an RBI file

aadi-stripe:
cc @reese 

Adds logic to report an error in the visibility checker which helps catch when an exported constant is only defined in RBI files.

Also introduces a new CLI flag to list paths where we want to skip this check (e.g. some paths that are generated automatically -- like protobuf Ruby definitions, which have some constants in `typed: ignore` RB files, with the corresponding RBI being strictly typed).

### Motivation

In the Stripe codebase, we do not want to export constants that are only declared in RBI files -- as we don't consider RBI-only definitions as valid "behavior-defining" definitions. This logic in turn comes from the autogen pass itself, see [here](https://github.com/sorbet/sorbet/blob/master/main/autogen/autogen.cc#L83-L84)

This case is often seen when a developer deletes a constant in an RB file, but forgets to delete it from its corresponding RBI shim if it exists (or delete its export if it is exported). 

To force our developers to either invoke our internal packaging tooling or manually delete the export (plus the shim) if they delete a constant, we add this logic.

### Test plan

Stripes: Note that this cannot be merged until we make the requisite change in the Stripe codebase to our Sorbet wrapper to add the `--package-skip-rbi-export-enforcement` flag.


froydnj (approved):

reese-stripe (approved):
🎉 Thanks for adding this!

aadi-stripe:
No runtime change:
Before
Screen Shot 2022-06-21 at 4 41 40 PM

After
Screen Shot 2022-06-21 at 4 48 44 PM

PR #5930, opened 2022-06-21
bring `Kernel.fail` and `Kernel.raise` sigs more closely in line with each other

froydnj:
### Motivation

These are implemented by the same function in Ruby underneath the hood, so they should take roughly the same arguments.

### Test plan

See included automated tests.  I left the slightly looser typing for `Kernel.raise`'s `arg1` and `arg2` for now and didn't both to try and sync things up.

### DIFF
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -1217,7 +1217,7 @@ module Kernel
   end
   sig do
     params(
-        arg0: Class,
+        arg0: Exception,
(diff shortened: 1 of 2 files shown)


jez (approved):

froydnj:

→ https://go/builds/bui_LuzqBkJPUEMnVO
→ https://go/builds/bui_Luzqd4WTIzVo2X

PR #5931, opened 2022-06-21
pack `ConstantLit` down to 16 bytes

froydnj:
### Motivation

This is an extension of #4954, which removed 8 bytes from `ConstantLit`...but unfortunately didn't change very much as far as memory usage goes because 24 bytes and 32 bytes all go into the 32-byte bucket in jemalloc.  To see savings, we need to reduce `ConstantLit` down to 16 bytes.

We can do that by observing that `ConstantLit`'s loc is exactly the same as the original `UnresolvedConstantLit`'s loc.  And in cases where we don't have an `original`, we're synthesizing a constant literal from something that didn't appear in the user's source code, and therefore we probably shouldn't be giving it a loc anyway.  So we can redirect all requests for the loc of a `ConstantLit` to its original unresolved constant literal, and just say the `ConstantLit` doesn't have a loc if the original doesn't exist.

### Test plan

See included automated tests.

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -449,12 +449,12 @@ public:
             return true;
         }
         auto root = ast::cast_tree<ast::ConstantLit>(scope);
-        return root != nullptr && root->symbol == core::Symbols::root();
+        return root != nullptr && root->symbol() == core::Symbols::root();
     }
 
     static bool isMagicClass(ExpressionPtr &expr) {
         if (auto *recv = cast_tree<ConstantLit>(expr)) {
-            return recv->symbol == core::Symbols::Magic();
+            return recv->symbol() == core::Symbols::Magic();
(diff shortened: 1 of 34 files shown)


froydnj:

→ https://go/builds/bui_LuzODqIiriIaX3
→ https://go/builds/bui_LuzOlvDjOgESSZ

jez:
> just say the `ConstantLit` doesn't have a loc if the original doesn't exist.

Can't we just reach inside the UnresolvedConstantLit (that would still be there, even if it were synthetic) and set the loc there?

I think that everything, even synthetic things, should have locs, otherwise it's too easy to accidentally have bad error messages. lmk if there's something that prevents this though

froydnj:
Closing in favor of #8568

PR #5932, opened 2022-06-21
Add `exclude-from-file-update` assertion to LSP test runner

jez:
### Motivation

The behavior of the LSP test runner is to include every `__.N.rbupdate` file in the set
of file updates sent to the LSP server.

This makes it hard to test certain kinds of tricky fast path logic, because the
act of including a file like `__.N.rbupdate` for the purpose of making error
assertions about that update will mean that the file forcibly gets
retypechecked. This masks cases where our `UsageHash`-based "find which files to
retypecheck" logic is wrong.

I've added an assertion to instruct the LSP runner to exclude certain files from
the set of file updates it feeds to the LSP server.

~~The test currently fails, and I will be working on fixing it shortly.~~
**Update** the test failed because I wrote the test wrong. It actually passes.

### Test plan

See included automated tests.


test/testdata/lsp/fast_path/static_field_change_type__use.rb
```diff
@@ -1,7 +1,11 @@
 # typed: strict
+# spacer for exclude from update
```
jez, line 2: When using the `exclude-from-file-update: true` assertion, the contents of this file will never be edited, so it has to contain exactly what we want the file to contain for all generations of the test.

But since subsequent files will "insert" the `exclude-from-file-update: true` assertion, we need a spacer line here so that `error` assertions (which are read out of the file by the test runner, not processed by the LSP server) are on the expected lines.

If we did not have this spacer, it would mean that for example our `error:` assertions in `.1.rbupdate` would be off by one line too many.

froydnj (approved):

PR #5939, opened 2022-06-22
delay accessing tree locs in `TreeMapper::mapIt`

froydnj:
### Motivation

For every recursive call to `TreeMapper::mapIt`, we fetch the loc of the tree we're mapping over, so we can throw a nicer exception in the error case.  We only ever use the loc in the error case: an error case that is extremely rare, not just error cases where the user wrote some unparseable code or something like that.  I assume we're fetching the loc early in the case that memory might have been corrupted by the time we catch the exception.

(Full disclosure: there is a selfish reason for getting rid of this and that's with the current form of #5931, `ExpressionPtr::loc` is no longer inlining, which makes the runtime cost of that PR somewhat high.  It'd be nice to not have to make `TreeMapper::mapIt` slower in that PR, but I think getting rid of the extraneous code here is worthwhile on its own.)

This PR proposes getting rid of this exception handling and just living with whatever `SorbetException` we threw, without the nicer error message of exactly which tree we were in the process of handling.  My sense is that we can figure out in the debugger what's going on and/or that the cases which this error was helpful are vanishingly rare nowadays and/or our `ENFORCE`s are much better and the corresponding backtraces from those more useful than the handling here would have been.

I can't recall ever seeing this fire, but I admit that my memory may not be the best.  Leaving this in for debug builds is also a possibility.

### Test plan

See included automated tests.

### DIFF
--- a/ast/treemap/treemap.h
+++ b/ast/treemap/treemap.h
@@ -467,7 +467,6 @@ private:
                 return;
             }
         }
-        auto loc = what.loc();
 
         try {
             // TODO: reorder by frequency
@@ -585,6 +584,8 @@ private:
                     return mapRuntimeMethodDefinition(Funcs::pass(what), ctx);
             }
         } catch (SorbetException &e) {
+            auto loc = what.loc();
+
             Exception::failInFuzzer();
 
             throw ReportedRubyException{e, loc};
--- a/ast/treemap/treemap.h
+++ b/ast/treemap/treemap.h
@@ -137,12 +137,6 @@ public:
     GENERATE_POSTPONE_POSTCLASS(Cast, arg_types);                                \
     GENERATE_POSTPONE_POSTCLASS(RuntimeMethodDefinition, arg_types);
 
-// Used to indicate that TreeMap has already reported location for this exception
-struct ReportedRubyException {
-    SorbetException reported;
-    core::LocOffsets onLoc;
-};
-
 enum class TreeMapKind {
     Map,
     Walk,
@@ -593,15 +587,7 @@ class TreeMap {
 public:
     template <typename CTX, typename FUNC> static ExpressionPtr apply(CTX ctx, FUNC &func, ExpressionPtr to) {
         TreeMapper<FUNC, CTX, TreeMapKind::Map, TreeMapDepthKind::Full> walker(func);
-        try {
-            return walker.mapIt(std::move(to), ctx);
-        } catch (ReportedRubyException &exception) {
-            Exception::failInFuzzer();
-            if (auto e = ctx.beginError(exception.onLoc, core::errors::Internal::InternalError)) {
-                e.setHeader("Failed to process tree (backtrace is above)");
-            }
-            throw exception.reported;
-        }
+        return walker.mapIt(std::move(to), ctx);
     }
 };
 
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_LvKKftfK32aBXD
→ https://go/builds/bui_LvKKOzFZFXKiMA

jez:
😢 i have gotten a lot of value out of this, especially in cases where an `ENFORCE` fails somewhere deep (e.g., inside `Symbols.cc` where there is no `loc` that could be printed in the `ENFORCE`'s error message

I would love to not have to lose this if possible

froydnj:
> 😢 i have gotten a lot of value out of this, especially in cases where an `ENFORCE` fails somewhere deep (e.g., inside `Symbols.cc` where there is no `loc` that could be printed in the `ENFORCE`'s error message
> 
> I would love to not have to lose this if possible

So you would prefer at least debug-only, then?  Or do you feel strongly about leaving it in by default?

jez:
> (Breaking on `__cxa_throw` in the debugger will let you know exactly where you are when that `ENFORCE` triggers.)

There are two use cases:

- in optimized builds, it can be hard to make sure that the right variables/symbols/etc. are in place to be able to print the path to the file, even if you've got it open in a debugger. So if something fails when running a release-debug-linux build over pay-server, it can sometimes be the only way to track down which file was causing the problem as a place to start minimizing
- in development builds, it's convenient to see the "Failed to process tree (backtrace is above)" error as a red squiggle in my editor. when exceptions aren't turned into Sorbet errors, it can be hard to notice that my in-progress change actually broke during some intermediate edit. i've found a lot of good test cases this way, and i'm worried that i wouldn't have caught them otherwise

So I think it's not that it's hard to breka on `__cxa_throw`.

> So you would prefer at least debug-only, then? Or do you feel strongly about leaving it in by default?

If it's on in debug, that would at least help for the `ENFORCE` use cases.

The other case where this would matter would be for cases where we called `Exception::raise` directly, even in non-debug builds. Unfortunately there are quite a few places where we call `Exception::raise` directly.

Maybe we should introduce a version of `Exception::raise` that takes a `core::Loc` or a `string` filepath, and attempt to migrate as many of the places calling `Exception::raise` to use that new helper as possible? Or maybe even: migrate as many of those as possible to call `beginError` with `core::errors::Internal::InternalError` before raising the exception?

There will still be a few places where there's no `Loc` or file path in scope, but maybe once we've gotten most things using the new helper it won't matter as much for the remaining ones.

froydnj:

→ https://go/builds/bui_Lvn0XVph8xuWTe
→ https://go/builds/bui_Lvn0iMeNXWan01

froydnj:

→ https://go/builds/bui_Lvn2ltfzx6X6pt
→ https://go/builds/bui_Lvn2zpi4sQrJHx

froydnj:
The most recent push changes `what.loc()` to only be fetched in the error case.  Earlier I stated:

> I assume we're fetching the loc early in the case that memory might have been corrupted by the time we catch the exception.

That's definitely a possibility, but as memory corruption tends to not be the problems (all the debugging scenarios jez described were from `ENFORCE`s, and we typically don't see memory corruption around trees), it seemed reasonable to say that we're not going to encounter memory corruption and therefore we can safely fetch the loc in the exception handler.  (And if you do have memory corruption, you may not be getting much value from the Sorbet-level errors anyway.)

Unfortunately, this change makes the PR not as impactful; there is still a small effect (~0.6%), but I am a little more willing to attribute it to noise.  namer and resolver also get slightly slower.  I think that's due to `what` needing to live longer, possibly being stored in memory as a result, and therefore the tree walking code is not quite as efficient in those cases.  It's also possible that just having the exception handling region limits the effectiveness of certain optimizations, which would be unfortunate.

Though packager gets significantly faster, so I don't understand how both of these things can be true at the same time.

PR #5946, opened 2022-06-22
Convert Histogram to log line

jez:
### Motivation

Stripe's metrics tooling isn't handling histograms like this quite the way we'd
like them to. Converting to a log line which will be slightly less automatic to
consume, but easier to do one-off analyses.

### Test plan

n/a


froydnj (approved):
I think you need to rebase because the README changes shouldn't be in this PR.

jez:
> I think you need to rebase because the README changes shouldn't be in this PR.

Hm, I don't see any README changes in this PR?

froydnj:
Weird, the diff on GitHub mobile is showing me two files changed, but the info on the diff on the PR page clearly shows only a single file changed.

PR #5956, opened 2022-06-24
give a better error message for incorrect {instance,class} variable declarations

froydnj:
### Motivation

I think somebody copy/pasted the error message from the constant case and forgot to change it?  This message is clearer.

### Test plan

See included automated tests.


jez (approved):
lol nice find

PR #5957, opened 2022-06-24
don't transitively lookup instance/class variables in CFG building

froydnj:
### Motivation

How many things break if we tried to implement something like this?

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_Lw6s1TV5CzlO0M
→ https://go/builds/bui_Lw6sheu6Yjlkgg

froydnj:
This is basically impossible to land; tests, in particular, seem to depend deeply on this behavior (I assume because of the way the Minitest rewriter works?).

PR #5966, opened 2022-06-27
Change pipeline.sh to account for change in env vars

jez:
### Motivation

It seems like sometimes we get
`BUILDKITE_PULL_REQUEST_REPO=https://github.com/sorbet/sorbet.git`,
whereas before we would always get `git://` URLs. Idk why this happened,
maybe GitHub changed something?

### Test plan

To roll this out, we need to update the pipeline.sh checksum on the build runner
machine (not in source control).

### DIFF
--- a/.buildkite/pipeline.sh
+++ b/.buildkite/pipeline.sh
@@ -4,13 +4,19 @@ set -euo pipefail
 
 whitelisted=0
 
+echo "BUILDKITE_PULL_REQUEST=$BUILDKITE_PULL_REQUEST"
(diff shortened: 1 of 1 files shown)


froydnj (approved):

jez:
Test in prod worked:

https://buildkite.com/sorbet/sorbet/builds/23134#0181a71d-73db-4e5c-a571-4510a3d85175/65-69

This will cause all builds to fail until they are rebased on top of this change, unfortunately.

PR #5976, opened 2022-06-30
Run document formatting requests on preprocessor thread

reese:
### Motivation

On the tin -- running on the index thread means that document formatting can be forced to wait for certain slow-path Sorbet operations, but document formatting has no relation to what Sorbet does and really only cares about the raw source, so running it on preprocessor should improve response times.

Another smaller change is rolled into this PR, which is that formatting errors failing due to syntax errors now longer show a notification and instead just log an error -- this is more of a UX improvement, since users don't want to manually have to close VSCode notifications.

### Test plan

This should be covered by existing automated tests, since it's functionally the same, but I've also manually tested this on Stripe's codebase.

### DIFF
--- a/main/lsp/LSPPreprocessor.cc
+++ b/main/lsp/LSPPreprocessor.cc
@@ -80,10 +80,10 @@ LSPPreprocessor::LSPPreprocessor(shared_ptr<LSPConfiguration> config, shared_ptr
     : config(move(config)), taskQueue(std::move(taskQueue)), owner(this_thread::get_id()),
       nextVersion(initialVersion + 1) {}
 
-string_view LSPPreprocessor::getFileContents(string_view path) const {
+string_view LSPPreprocessor::getFileContents(string_view path, bool enforceFileExists) const {
     auto it = openFiles.find(path);
     if (it == openFiles.end()) {
-        ENFORCE(false, "Editor sent a change request without a matching open request.");
+        ENFORCE(!enforceFileExists, "Editor sent a change request without a matching open request.");
         return string_view();
     }
     return it->second->source();
@@ -360,7 +360,7 @@ LSPPreprocessor::canonicalizeEdits(uint32_t v, unique_ptr<DidChangeTextDocumentP
     if (config->isUriInWorkspace(uri)) {
         string localPath = config->remoteName2Local(uri);
         if (!config->isFileIgnored(localPath)) {
-            string fileContents = changeParams->getSource(getFileContents(localPath));
+            string fileContents = changeParams->getSource(getFileContents(localPath, true));
             auto fileType = core::File::Type::Normal;
             auto &slot = openFiles[localPath];
             auto file = make_shared<core::File>(move(localPath), move(fileContents), fileType, v);
--- a/main/lsp/LSPPreprocessor.cc
+++ b/main/lsp/LSPPreprocessor.cc
@@ -80,13 +80,23 @@ LSPPreprocessor::LSPPreprocessor(shared_ptr<LSPConfiguration> config, shared_ptr
     : config(move(config)), taskQueue(std::move(taskQueue)), owner(this_thread::get_id()),
       nextVersion(initialVersion + 1) {}
 
-string_view LSPPreprocessor::getFileContents(string_view path, bool enforceFileExists) const {
-    auto it = openFiles.find(path);
-    if (it == openFiles.end()) {
-        ENFORCE(!enforceFileExists, "Editor sent a change request without a matching open request.");
+string_view LSPPreprocessor::getFileContents(string_view path) const {
+    auto maybeFileContents = maybeGetFileContents(path);
+    if (!maybeFileContents.has_value()) {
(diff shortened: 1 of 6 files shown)


main/lsp/LSPPreprocessor.cc
```diff
@@ -360,7 +360,7 @@ LSPPreprocessor::canonicalizeEdits(uint32_t v, unique_ptr<DidChangeTextDocumentP
     if (config->isUriInWorkspace(uri)) {
         string localPath = config->remoteName2Local(uri);
         if (!config->isFileIgnored(localPath)) {
-            string fileContents = changeParams->getSource(getFileContents(localPath));
+            string fileContents = changeParams->getSource(getFileContents(localPath, true));
```
jez, line 363: I'm not a fan of boolean parameters

what do you think about adding a method called `maybeGetFileContents` that returns an `optional<string_view>` if the file doesn't exist, and call that method instead of passing through the boolean?

main/lsp/requests/document_formatting.cc
```diff
@@ -83,34 +86,33 @@ void DocumentFormattingTask::index(LSPIndexer &index) {
...
                 result = move(edits);
                 response->result = move(result);
                 break;
             case RubyfmtStatus::SYNTAX_ERROR:
-                displayError(fmt::format("`rubyfmt` could not format {} because it contains syntax errors.",
-                                         index.getFile(fref).path()),
-                             response);
+                response->error = make_unique<ResponseError>(
+                    (int)LSPErrorCodes::RequestFailed,
+                    fmt::format("`rubyfmt` could not format {} because it contains syntax errors.", path));
```
jez, line 98: ooc, is this what other formatters do when there are syntax errors?

i wonder if it might be better to return the file contents unchanged?


> result: [TextEdit[]](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit) | null describing the modification to the document to be formatted.
> error: code and message set in case an exception happens during the formatting request.

i wonder if maybe it expects a `null` response to be sent in that case, instead of an `error`?

jez (approved):

main/lsp/requests/document_formatting.cc
reese: Fair enough, not sure about what other formatters do but I think it's reasonable to say that syntax errors are somewhat "expected" and can be no-ops.

reese: Actually, I just looked at the prettier logs for typescript, and it does look like it returns an error. I made a junk change and saw this:

```
["ERROR" - 10:36:43 AM] Error formatting document.
["ERROR" - 10:36:43 AM] ';' expected. (2:4)
  1 | export {};
> 2 | fn fn fn sdfsdlj
    |    ^
  3 | 
SyntaxError: ';' expected. (2:4)
  1 | export {};
> 2 | fn fn fn sdfsdlj
    |    ^
  3 | 
```

I don't know of an easy way to see whether they simply logged the error and returned `null` or if they actually returned an error type, but that's at least another data point for how to handle it. Going off of that, I'm inclined to return this as an error. Invalid syntax _is_ an error to `rubyfmt` after all, so I think reflecting that in the response type makes sense to me, but let me know if you feel otherwise.

jez: Here is where rust analyzer returns Ok(None) instead of Err(…) for syntax errors:

https://github.com/rust-lang/rust-analyzer/blob/7f9c054686fb3e1af7acb7292e05462afc138981/crates/rust-analyzer/src/handlers.rs#L1856

the comment in the source is worth reading too

reese: Interesting, I don't feel strongly either way, so happy to have it be `null` instead 👍

jez: Also I haven’t been able to find: does TypeScript’s LSP actually support textDocument/formatting requests? I thought that prettier was a separate VS Code plugin that used VS Code extension APIs, not the LSP spec for formatting.

unlike most of the time where we copy typescript when trying to resolve LSP ambiguities in sorbet, I don’t know whether typescript’s implementation is a good role model for formatting

reese: It does [apparently](https://github.com/typescript-language-server/typescript-language-server/blob/76bf9a4817ffa1e340422cfd5177dbcb96528ddb/src/lsp-server.ts#L717-L743)! I didn't look too terribly far into it, but it looks to me like it also swallows errors (of all kinds) and returns an empty `TextEdit[]` (and never `null`, interestingly).

jez: from the docs it seems like that is it’s “basic formatting”

https://code.visualstudio.com/docs/languages/typescript#_formatting

not prettier, so the error logs you saw from prettier were likely from the separate plugin

reese: Yeah, the prettier ones are definitely separate (and agreed that they're probably not the best example here).

PR #5979, opened 2022-06-30
add a timer for using file hashes on the fast path

froydnj:
### Motivation

I would like to know approximately how much time this takes, so we could determine whether it would be reasonable to move this computation to the determination of whether or not to take the fast path in the first place.  (It would obviously need a little bit of tweaking to put it there, but not much.)

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -240,89 +240,94 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
     // Replace error queue with one that is owned by this thread.
     gs->errorQueue = make_shared<core::ErrorQueue>(gs->errorQueue->logger, gs->errorQueue->tracer, errorFlusher);
     {
-        vector<core::SymbolHash> changedMethodSymbolHashes;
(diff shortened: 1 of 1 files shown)


main/lsp/LSPTypechecker.cc
```diff
@@ -240,89 +240,94 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
-                ENFORCE(validateIdenticalFingerprints(oldFieldHashes, newFieldHashes),
-                        "definitionHash should have failed");
-
-                absl::c_set_difference(oldFieldHashes, newFieldHashes, std::back_inserter(changedFieldSymbolHashes));
-
-                gs->replaceFile(fref, updatedFile);
-                // If file doesn't have a typed: sigil, then we need to ensure it's typechecked using typed: false.
-                fref.data(*gs).strictLevel = pipeline::decideStrictLevel(*gs, fref, config->opts);
-                subset.emplace_back(fref);
+        Timer timeit(config->logger, "fast_path_hash_usage");
```
jez, line 243:
```suggestion
        Timer timeit(config->logger, "compute_fast_path_file_set");
```

?

jez (approved):

PR #5981, opened 2022-07-01
don't permit redeclaring fields with the same type

froydnj:
### Motivation

It would be convenient if we just mandated you could not declare fields multiple times.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2002,11 +2002,7 @@ class ResolveTypeMembersAndFieldsWalk {
         auto prior = scope.data(ctx)->findMember(ctx, uid->name);
         if (prior.exists() && prior.isFieldOrStaticField()) {
             auto priorField = prior.asFieldRef();
-            if (core::Types::equiv(ctx, priorField.data(ctx)->resultType, cast->type)) {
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_LynQZkDRoOPxDy
→ https://go/builds/bui_LynRHQXNfNMUpo

froydnj:
There are two primary problems with this PR that I noticed by clicking through a few failures in Stripe's CI:

1. Given code like:

```ruby
  it "does something" do
    @var = T.let(nil, T.nilable(Integer))
  end

  it "does something else" do
    @var = T.let(nil, T.nilable(Integer))
  end
```

it's really nice if you don't have to choose unique names for your instance variables.  You could argue that the test code should be factored better or whatever, and you'd probably be right, but doing things this way is convenient for users.

2. Likewise for convenience, you have:

```ruby
class Foo
  def initialize
    if external_condition?
      @var = T.let(..., Hugely::Complicated::Type)
    else
      @var = T.let(..., Hugely::Complicated::Type)
    fi
  end
end
```

and again, you could refactor this so that the variable was only declared once, but this is a bit more readable and convenient, which I assume was part of the reason this was permitted in the first place.

PR #5974, opened 2022-06-30
Drop everything in the database

jez:
### Motivation

We were not sufficiently purging the disk cache, which meant that we effectively
leaked cache space.

I would like to clean this change up in a follow up PR, but this should solve
the immediate problem.

### Test plan

I tested this manually by switching back and forth between new and old branches,
and dumped the databases using commands like these:

```
sorbet --cache-dir=jez-cache foo.rb
sorbet --cache-dir=jez-cache --skip-rewriter-passes foo.rb
sorbet --cache-dir=jez-cache --enable-experimental-lsp-fast-path foo.rb
sorbet --cache-dir=jez-cache --skip-rewriter-passes --enable-experimental-lsp-fast-path foo.rb
sorbet --cache-dir=master-cache foo.rb
sorbet --cache-dir=master-cache foo.rb bar.rb
sorbet --cache-dir=jez-cache foo.rb bar.rb

pip install lmdb

python -mlmdb --env jez-cache dump default=foo.txt
python -mlmdb --env jez-cache dump nodsl=foo.txt
python -mlmdb --env jez-cache dump dsl-experimentalfastpath=bar.txt
python -mlmdb --env jez-cache dump
python -mlmdb --env jez-cache dump dsl-experimentalfastpath=foo.txt
python -mlmdb --env master-cache dump default=bar.txt
python -mlmdb --env jez-cache dump default=bar.txt
python -mlmdb --env jez-cache dump dsl-normalfastpath=foo.txt
```

in various orders and combinations and branches.

PR #5985, opened 2022-07-05
Thread (currently unused) logger argument to KeyValueStore

jez:
### Motivation

I'm going to follow up with some more changes to the caching code. When I do,
I'll make use of this new logger argument.

I recall wanting this to exist so that I could have written log statements when
implementing #5974, because it's otherwise tricky to debug (the cache version
checking logic depends on being compiled in release mode, which makes for poor
debugger symbols).

### Test plan

Existing build

### DIFF
--- a/common/kvstore/KeyValueStore.cc
+++ b/common/kvstore/KeyValueStore.cc
@@ -73,7 +73,7 @@ bool hasNoOutstandingReaders(MDB_env *env, string_view path) {
 
 KeyValueStore::KeyValueStore(shared_ptr<spdlog::logger> logger, string version, string path, string flavor,
                              size_t maxSize)
-    : version(move(version)), path(move(path)), flavor(move(flavor)), dbState(make_unique<DBState>()), logger(logger) {
+    : version(move(version)), path(move(path)), flavor(move(flavor)), dbState(make_unique<DBState>()), logger(move(logger)) {
     ENFORCE(!this->version.empty());
     bool expected = false;
     if (!kvstoreInUse.compare_exchange_strong(expected, true)) {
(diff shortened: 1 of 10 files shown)


common/kvstore/KeyValueStore.cc
```diff
@@ -71,8 +71,9 @@ bool hasNoOutstandingReaders(MDB_env *env, string_view path) {
 }
 } // namespace
 
-KeyValueStore::KeyValueStore(string version, string path, string flavor, size_t maxSize)
-    : version(move(version)), path(move(path)), flavor(move(flavor)), dbState(make_unique<DBState>()) {
+KeyValueStore::KeyValueStore(shared_ptr<spdlog::logger> logger, string version, string path, string flavor,
+                             size_t maxSize)
+    : version(move(version)), path(move(path)), flavor(move(flavor)), dbState(make_unique<DBState>()), logger(logger) {
```
froydnj, line 76:
```suggestion
    : version(move(version)), path(move(path)), flavor(move(flavor)), dbState(make_unique<DBState>()), logger(move(logger)) {
```

froydnj (approved):
Have you tried compiling with `--config=debugsymbols`?  That doesn't fix everything, but it does give you enough information that optimized code is reasonably debuggable.

jez:
> Have you tried compiling with --config=debugsymbols? That doesn't fix everything, but it does give you enough information that optimized code is reasonably debuggable.

I was trying both `--config=release-linux` and `--config=release-debug-linux`, neither of which were particularly pleasant to debug.

froydnj:
> > Have you tried compiling with --config=debugsymbols? That doesn't fix everything, but it does give you enough information that optimized code is reasonably debuggable.
> 
> I was trying both `--config=release-linux` and `--config=release-debug-linux`, neither of which were particularly pleasant to debug.

`--config=debugsymbols` can be specified in addition to the above, or with `-c opt`.

PR #5986, opened 2022-07-06
Add support for slash-munged "extra package paths" in --stripe-packages mode

aadi-stripe:
Adds support for slash-munged "extra package paths" (i.e. `.../project/foo/...`, as opposed to underscore-munged `.../Project_Foo/...` in `--stripe-packages` mode).

### Motivation

Due to some changes we are making in the Stripe codebase, we need to support additional package paths which are lower case and munged with slashes instead of uppercase and munged with underscores.

### Test plan


aadi-stripe:
Redoing, old merge base

PR #5990, opened 2022-07-06
Update error-reference.md for error 7003

robthornton:
Provide an alternative solution to error 7003. Since what we ultimately want to do is provide a guarantee that Kernel#raise exists on the class being mixed into at runtime, requires_ancestor is a viable alternative.

### Motivation

I think providing both solutions gives the user choice. This would also help provide some visibility into this expirimental feature and help drive testing it.

### Test plan

This is a documentation change, test changes are required.

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -2811,11 +2811,22 @@ This error indicates a call to a method we believe does not exist (a la Ruby's
     `Object` (which classes default to inheriting from), but not on
     `BasicObject` (which classes can optionally inherit from).
 
-    The solution is to `include Kernel` in our module:
+    One solution is to `include Kernel` in our module:
 
     ```ruby
     module MyModule
       include Kernel
+    end
+    ```
+    
+    An alternative solution is to use the expirimental `requires_ancestor` from
+    `T::Helpers` to ensure `Kernel` is an ancestor of the class being mixed
+    into at runtime.
+    
+    ```ruby
+    module MyModule
+      include T::Helpers
+      requires_ancestor Kernel
     end
     ```
 


Morriar (commented):
Note: `requires_ancestor` needs the `--enable-experimental-requires-ancestor` option to be enabled.

jez:
I'm not quite sure. `include Kernel` works in all Ruby code out of the box and has ~no downsides. Using `requires_ancestor` for something as simple as having access to methods from `Kernel` in your module seems like overkill. Given that this is one of the first error messages that people are likely to see, I'm not sure that it makes sense to include right here in the error reference docs.

If you want to put an example of how to use `requires_ancestor` with `Kernel` in the "Requiring Ancestors" doc, that seems appropriate.

jez (commented):
I'm not quite sure. `include Kernel` works in all Ruby code out of the box and has ~no downsides. Using `requires_ancestor` for something as simple as having access to methods from `Kernel` in your module seems like overkill. Given that this is one of the first error messages that people are likely to see, I'm not sure that it makes sense to include right here in the error reference docs, as we don't really want to be pushing new users towards experimental features out of the box.

If you want to put an example of how to use `requires_ancestor` with `Kernel` in the "Requiring Ancestors" doc, that seems appropriate.

sambostock:
This error isn't limited to `Kernel` methods though, it's just an example of a dependency on a mixed-in module within Ruby itself.

While I agree that for this case `require_ancestor` may be overkill, perhaps these docs should mention examples that aren't `Kernel`, and in those cases mention (with cautionary notes) the experimental `require_ancestor` feature.

I think discoverability and cross-linking in documentation is important, even for experimental features, because one may not know what exactly to search for to solve one's problems, leading to code needlessly using `T.unsafe`.

PR #5994, opened 2022-07-06
Use snake-case for slash-munged paths in packager

aadi-stripe:
For the newly added "slash-munged" package paths in #5987 , we need to also snake-case the path names in addition to lower-casing, so that `Project_FooBar` transforms to `project/foo_bar`. Previously it was just `project/foobar`.

### Motivation

In the Stripe codebase, we are working on introducing path-based autoloading based on [Zeitwerk](https://github.com/fxn/zeitwerk)'s convention, which is used in Rails codebases.

### Test plan

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -14,6 +14,7 @@
 #include "core/errors/packager.h"
 #include "core/packages/PackageInfo.h"
 #include <cctype>
+#include <sstream>
 #include <sys/stat.h>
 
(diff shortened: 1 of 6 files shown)


bhang-stripe (approved):

aadi-stripe:
Will merge after review from core reviewers, aka @jez or @froydnj

jez (approved):

PR #5999, opened 2022-07-08
Add grab bag lmdb Python script

jez:
At the time this script was committed, it was known to work with the
Python lmdb 1.3.0 bindings.

<https://pypi.org/project/lmdb/1.3.0/>

### Motivation

Just wanted to have something on hand that I could look back on if we need to
look into database cache sizes again.

### Test plan

I don't think it's particularly worth testing it.

### DIFF
--- a/tools/scripts/lmdb-scratch.py
+++ b/tools/scripts/lmdb-scratch.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+import sys
+import lmdb
+import pprint
+import itertools
+import csv
+
+# This script is an assortment of code that I have found useful when digging
+# into LMDB cache sizes in the past.
+#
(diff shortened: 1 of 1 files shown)


tools/scripts/lmdb-scratch.py
```diff
@@ -0,0 +1,154 @@
...
+import lmdb
+import pprint
+import itertools
+import csv
+
+# This script is an assortment of code that I have found useful when digging
+# into LMDB cache sizes in the past.
+#
+# It's not tested, and is mostly useful for invoking by-hand on a Sorbet cache,
+# but it should at lest serve as a way to document gotchas and things to look
```
froydnj, line 12:
```suggestion
# but it should at least serve as a way to document gotchas and things to look
```

froydnj (approved):

PR #6003, opened 2022-07-11
Improve document formatting performance with `posix_spawn`

reese:
The `subprocess` library currently used for document formatting is horribly slow for our use case. I did some profiling, and about 150-300ms are spent just calling `fork` [here](https://github.com/arun11299/cpp-subprocess/blob/master/subprocess.hpp#L1602), since `fork` copies the memory space of the parent process into the child process. That coupled with the time spent e.g. configuring file descriptors meant that _just starting the process_ for `rubyfmt` took more than half a second, which is unacceptably slow for a code formatter.

This change updates the document formatter to use an existing utility that uses [`posix_spawn`](https://man7.org/linux/man-pages/man3/posix_spawn.3.html) for the process creation, which gets us down from half a second to <20ms (including the call to `rubyfmt` itself) on a quiet machine. 

(cc @jvilk-stripe)

### Motivation

Performance! Since this could block file saves for users with `formatOnSave` on, it should be sufficiently fast that it's almost unnoticed.

### Test plan

This should be functionally equivalent to the existing implementation, so existing tests should cover it. I also manually ran this in the Stripe VSCode setup.

### DIFF
--- a/common/Subprocess.cc
+++ b/common/Subprocess.cc
@@ -3,6 +3,7 @@
 #include <array>
 #include <spawn.h>
 #include <sstream>
+#include <string.h>
 #include <string>
 #include <sys/wait.h>
 #include <unistd.h>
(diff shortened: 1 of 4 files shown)


common/Subprocess.cc
```diff
@@ -53,35 +53,63 @@ class FileActions {
 };
 } // namespace
 
-// can't take string_views because we need char * not const char *
-optional<string> sorbet::Subprocess::spawn(string executable, vector<string> arguments) {
+// Spawns a new child process, pipes `stdinContents` into the new processes's stdin,
+// and returns the child's stdout and status code
+optional<sorbet::Subprocess::Result> sorbet::Subprocess::spawn(string executable, vector<string> arguments,
```
jez, line 58: This API is already in use (in `compiler/Linker/Linker.cc`).

The changes inside this method don't look to me to be entirely about performance improvements? It also looks like we're unconditionally piping some standard streams that we were not previously. Will that break the existing usage in Linker.cc?

jez (approved):

jez:
For posterity, jvilk linked me this paper:

https://www.microsoft.com/en-us/research/uploads/prod/2019/04/fork-hotos19.pdf

PR #6005, opened 2022-07-12
Create T::ImmutableStruct

vinistock:
Please see #5650 and the discussions on that PR.

This PR is an alternative implementation of the `T::ImmutableStruct` with the goal to avoid any breaking changes.

### Test plan

See included automated tests.


jez (approved):

brendo:
:wave: Sorry to comment on an old PR, but I'm curious if this is intended to be used widely? 

I ask because I noticed it's not documented publicly on https://sorbet.org/docs/tstruct, and I only found learnt about it whilst reading an internal Shopify gem README.

PR #6015, opened 2022-07-14
Show when there's a blocking fast path operation

jez:
### Motivation

Most fast path operations are fast, but some are quite slow. As a
heuristic, choose to report when the fast path is about to run on 100+
files. We may want to adjust this up or down, depending on how much it
causes flickering.

By my estimation, this will affect `0.08%` of fast path edits, or 8 in 10,000.
Given that there are hundreds of thousands of fast path operations daily,
this should be visible to people but not burdensome.

### Test plan

I have tested this on Stripe's codebase:

Screen Shot 2022-07-14 at 1 57 26 PM

I do not want to build an LSP test that has 101 files in it to test that this works as expected.


jez:

→ https://go/builds/bui_M3fyqvHOlwVKHm
→ https://go/builds/bui_M3fyPUybNdgTwl

aisamanra (approved):

PR #3527, opened 2020-10-19
Implement dropSubtypesOf on sealed AppliedType

jez:
This implements exhaustiveness checking on sealed generic class
hierarchies, allowing for things like generic algebraic data types.

### Motivation

Closes #2572
Closes #1714

### Test plan

See included automated tests.

PR #6018, opened 2022-07-15
Don't let `T.all` silently instantiate bad bounds

jez:
### Motivation

Taking a look at the new test in this PR, `Child1[String]` and `Child2[String]`
are bad types: they don't satisfy the bounds on the respective classes.

In our attempts to make the types collapse inside `glb`, we could sometimes
overstep and accidentally instantiate these types. The solution is to check
whether the bounds match before collapsing.

This came up as a problem when I attempted to implement #3527, but upon further
reflection I realized that the limitation holding that PR back was not my
approach to implementing generics + exhaustiveness, but rather `T.all` itself,
because the problem can be seen independently of any changes introduced in that
PR.

With this change I think I should be able to revive that PR.

### Test plan

See included automated tests.


ngroman (approved):
Test makes sense to me

PR #6026, opened 2022-07-15
Follow symlinks if they point to directories

ytaben:
This change makes sorbet follow symlinks when loading ruby files from the `--dir` directive.

As of now, the following works:
`bundle exec srb tc --dir path_to_symlink`

The following _does not_ work:
`bundle exec srb tc --dir path_to_directory_containing_symlink`

Basically, specifying a symlink _directly_ work but any symlink _within_ that directory is ignored as it is a symlink and not a directory.

### Motivation

For a bit of context:
Our organization uses sorbet in projects with symlinked gem dependencies (we use a monorepo).

Unfortunately, sorbet language server (--lsp) does not work wither either gems that include `rbi` folder or passing more than one `--dir` directive in the config file.

Normally this is solved witha tool like `tapioca` to essentially copy the distributed annotations to the project's `sorbet/rbi` folder. It does not, however, make sense with a symlinked gem as it's meant to be updated in various projects _without_ touching those projects.

One solution to this problem would be to symlink the `rbi` folder within `sorbet/rbi` folder, allowing sorbet to pick up the distributed annotations.

Note that this is a bandaid solution as realistically - it would make sense for sorbet LSP to suport multiple directories. Otherwise - users who follow Sorbet's recommendation on distributing annotations via `rbi` dir break compatibility for users who want to use the language server.

### Extra notes
- We do not check for cycles but I'm personally fine with it. If there's a symlink cycle - call it a user error.

- I haven't touched C++ in a very long time so suggestions in that regard are welcomed.

### Test plan

I can add tests when maintainers confirm that this approach is viable and does not require changes to the approach.

### DIFF
--- a/common/common.cc
+++ b/common/common.cc
@@ -275,6 +275,9 @@ void appendFilesInDir(string_view basePath, const string &path, const sorbet::Un
             }
             appendFilesInDir(basePath, fullPath, extensions, recursive, result, absoluteIgnorePatterns,
                              relativeIgnorePatterns);
+        } else if (recursive && entry->d_type == DT_LNK && sorbet::FileOps::dirExists(fullPath)) {
+            appendFilesInDir(basePath, fullPath, extensions, recursive, result, absoluteIgnorePatterns,
+                             relativeIgnorePatterns);
         } else {
             auto dotLocation = fullPath.rfind('.');
             if (dotLocation != string::npos) {


jez:
Given the implementation, I have a feeling that this will fail for the same reason that the error "LSP requires a single input folder" exists, namely it assumes that every file in the project is under the a root path.

https://github.com/sorbet/sorbet/blob/master/main/lsp/LSPConfiguration.cc#L26-L27

The language server makes a somewhat strong and pervasive assumption that this is the case, and I think that changing how files are discovered to follow symlinks is just going to expose the fact that sorbet doesn't allow more than one root path.

If you would like to work on this, i would suggest removing the assumption from sorbet that there is a single root path. Note that I do not think this is an easy/straightforward task, and will likely involve refactoring things.

jez:
There is more information about this in #2496.

ytaben:
@jez 
Why do you think this would fail for the same reason?

Specifically, what would be the difference between copying folder under the symlink and accessing them _via_ symlink?
If I were to do `cp -r symlink/* sorbet/rbi` - the annotations would be discovered correctly.

With regards to the issue you posted - I saw it and while it discusses the symptomps - there is little information on what the single root assumption means for LSP and what it would entail to remove it.

On a side note - I did try to address the underlying issue in a somewhat crude manner by removing the guards in LSP and continuing to set the first `--dir` entry as root and that also worked but I don't know of potential consequences for this (the goal wasn't to support multiple roots per se but rather to allow including files from outside of root into typechecking)

jez:
Sorry i mistakenly assumed that we would append the absolute path to the list of files. If we keep it entirely in terms of paths that use the symlink, not the absolute paths, it might work.

I'd need to think about how this could break further, but i have absolutely no time today. Also jvilk is the person who is most knowledgeable about this and is unavailable for a week or so, so the earliest I'd imagine getting back to you on this is probably ~2 weeks from now.

common/common.cc
```diff
@@ -275,6 +275,9 @@ void appendFilesInDir(string_view basePath, const string &path, const sorbet::Un
             }
             appendFilesInDir(basePath, fullPath, extensions, recursive, result, absoluteIgnorePatterns,
                              relativeIgnorePatterns);
+        } else if (recursive && entry->d_type == DT_LNK && sorbet::FileOps::dirExists(fullPath)) {
```
alexchow, line 278: For symlinks, would we want to avoid cycles? Keep a set of resolved symlinks (abs path) somewhere and don't revisit?

ytaben: @alexchow I noted my personal view on this in the description but I _feel_ like symlink cycles support is not necessary (it may not be worth the effort but the `shared_ptr` and default argument may make the code quality impact bearable).

I have pushed a change to do the check needed (still need to test locally) so you can evaluate if it's worth it

ytaben:
@jez-stripe Hey, just checking in to make sure this didn't fall through the cracks.
Any chance @jvilk-stripe could take a look at this also?

Understandable if you're busy, we just have a number of people in our org that would benefit from this

leijin3:
@jez-stripe and @jvilk-stripe could one of you please help take a look at this? We would hugely benefit from this PR. Thanks!

PR #6027, opened 2022-07-15
Capitalize first char in slash munge path

bhang-stripe:
This PR changes the slash munging from https://github.com/sorbet/sorbet/pull/5987 to capitalize the first character for each part of a path. Instead of `.../project/foo/bar_baz/...`, we'll get `.../Project/Foo/Bar_baz/...`.

### Motivation

The directories provided to `--ignore` occassionally conflict with the new paths. They all are lowercase strings, so capitalizing the munged paths prevents conflicts. This closer matches the previous underscore munging as well.

### Test plan

See included automated tests.


bhang-stripe:

→ https://go/builds/bui_M439yZkjPatHC4
→ https://go/builds/bui_M439lS9BlIc6a2

PR #6032, opened 2022-07-15
Take the slow path for initialize, call, and new

jez:
### Motivation

This is a hack. LSP currently does a poor job of taking the fast path when there
are thousands or tens of thousands of files to typecheck as a part of the fast
path. The problem manifests as a fast path that is neither preemptible nor
cancellable that runs on a single thread (not the typecheck worker pool) until
everything has finished.

This problem most frequently comes with these handful of methods, so to
partially alleviate it before we build a longer-term solution, we're going to
force any changes to these methods to take the slow path, even if it would have
otherwise been possible to take the fast path.

### Test plan

See included automated tests.


aisamanra (approved):
This makes sense to me!

PR #6034, opened 2022-07-16
Allow using `T.type_parameter` in `T.let`

### LINKED ISSUE #1715: Can't use T.let with T.type_parameter to fix pinning error
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Btype_parameters(%3AU).params(x%3A%20T.type_parameter(%3AU)).void%7D%0A%20%20def%20pinning_error(x)%0A%20%20%20%20result%20%3D%20nil%0A%20%20%20%20%5B%5D.each%20do%0A%20%20%20%20%20%20result%20%3D%20x%20%23%20pinning%20error%0A%20%20%20%20end%0A%20%20end%0A%0A%20%20sig%20%7Btype_parameters(%3AU).params(x%3A%20T.type_parameter(%3AU)).void%7D%0A%20%20def%20generics_error(x)%0A%20%20%20%20%23%20Can't%20fix%20pinning%20error%20because%20generics%20bug!%0A%20%20%20%20result%20%3D%20T.let(nil%2C%20T.nilable(T.type_parameter(%3AU)))%0A%20%20%20%20%5B%5D.each%20do%0A%20%20%20%20%20%20result%20%3D%20x%0A%20%20%20%20end%0A%20%20end%0Aend)

```ruby
# typed: true
class A
  extend T::Sig

  sig {type_parameters(:U).params(x: T.type_parameter(:U)).void}
  def pinning_error(x)
    result = nil
    [].each do
      result = x # pinning error
    end
  end

  sig {type_parameters(:U).params(x: T.type_parameter(:U)).void}
  def generics_error(x)
    # Can't fix pinning error because generics bug!
    result = T.let(nil, T.nilable(T.type_parameter(:U)))
    [].each do
      result = x
    end
  end
end
```

#### Observed output

```
editor.rb:16: Unspecified type parameter https://srb.help/5004
    16 |    result = T.let(nil, T.nilable(T.type_parameter(:U)))
                                                           ^^

editor.rb:9: Changing the type of a variable in a loop is not permitted https://srb.help/7001
     9 |      result = x # pinning error
                       ^
  Existing variable has type: NilClass
  Attempting to change type to: A#pinning_error#U

    editor.rb:7: Replace with T.let(nil, T.nilable(A#pinning_error#U))
     7 |    result = nil
                     ^^^
Errors: 2
```

#### Expected behavior

The pinning error is expected.

The error when trying to use `T.type_parameter` in the method body is not. I'm pretty sure that type_parameters should be in scope in the method body.

- - -

### LINKED ISSUE #1848: Quick fixes suggest T.let on a type parameter

jez:
image

### Motivation

Fixes #1715
Fixes #1848

### Commit summary

This PR has a lot of changes, but a lot of them are ancillary.

The diff looks better reviewing by commit, and certain commits also benefit from ignoring whitespace.

- **pre-work: Rename this struct to make room for another** (7d470579e)

- **pre-work: resolveSigs always returns a result** (142892257)

- **pre-work: Compute the owner centrally** (ca546d9f7)

  This change makes it so that job.owner is always the actual owner
  (sometimes a method, sometimes a class, depending on where the cast item
  is), and we only widen it to `enclosingClass` right before we create a
  `ctx` that we're going to pass into `getResultType`, which assumes that
  `ctx.owner` is always a class, not a method.

  This lets us use the method owner for future changes.

- **pre-work: Factor out `*Impl` function to check postcondition** (73694bc6d)

- **pre-work: Convert typecase to `if`** (6cd49b8a9)

- **pre-work: clang-format after previous change** (8f0e9ffb8)

- **Allow using `T.type_parameter` in `T.let`** (52f99c5eb)

- **no-op: Refactor to make everything optional** (8b9be460c)

- **Make use of new optional type** (44ec80812)

- **Ensure that `T.type_parameter` can't be used in ivar decls** (466eb5ddd)

- **Add a new test** (d177f7f6b)

### Test plan

- [x] @jez Fix bug with todo (what you really want to be able to do is throw an exception all the way back out to the top of `getResultType`, because you'll have to bubble the `todo()` type all the way up at every call right now. Not sure what the best way to communicate the possibly-deeply-nested `T.type_parameter` failure back to the caller is.
- [x] @jez Write tests
  - [x] @jez Ensure this doesn't accidentally allow `T.type_parameter` in places
    where it shouldn't be, like static fields, type aliases, and field.
  - [x] @jez Write test for autocorrect with `T.type_parameter` #1848
- [x] @jez Measure performance on large codebase
  - This seems to account for an increase of about ~1% time. Personally that's within the threshold I'm comfortable regressing, if the measurement is accurate.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1646,9 +1646,9 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                                     // variable came from some other source (e.g. a function argument)
                                     auto suggest =
                                         core::Types::any(ctx, dropConstructor(ctx, tp.origins[0], tp.type), cur.type);
-                                    e.replaceWith(fmt::format("Initialize as `{}`", suggest.show(ctx)), cur.origins[0],
-                                                  "T.let({}, {})", cur.origins[0].source(ctx).value(),
-                                                  suggest.show(ctx));
+                                    auto replacement = suggest.show(ctx, core::ShowOptions().withShowForRBI());
+                                    e.replaceWith(fmt::format("Initialize as `{}`", replacement), cur.origins[0],
+                                                  "T.let({}, {})", cur.origins[0].source(ctx).value(), replacement);
                                 } else {
                                     e.addErrorSection(core::ErrorSection("Original type from:",
                                                                          cur.origins2Explanations(ctx, ownerLoc)));
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1496,8 +1496,15 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
             },
             [&](cfg::Cast &c) {
                 auto klass = ctx.owner.enclosingClass(ctx);
+
                 auto castType = core::Types::instantiate(ctx, c.type, klass.data(ctx)->typeMembers(),
                                                          klass.data(ctx)->selfTypeArgs(ctx));
+                if (inWhat.symbol.data(ctx)->flags.isGenericMethod) {
+                    // ^ This mimics the check in LoadArg's call to argumentTypeAsSeenByImplementation
+                    // It instantiates any `T.type_parameter(:U)`'s in the type (which are only
+                    // valid in a method body if the method's signature is generic).
+                    castType = core::Types::instantiate(ctx, castType, constr);
+                }
 
                 tp.type = castType;
                 tp.origins.emplace_back(ctx.locAt(bind.loc));
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1859,7 +1859,7 @@ class ResolveTypeMembersAndFieldsWalk {
         core::NameRef fromName;
     };
 
-    struct ResolveTypeMembersAndFieldsResult {
+    struct ResolveTypeMembersAndFieldsWorkerResult {
         vector<ast::ParsedFile> files;
         vector<ResolveAssignItem> todoAssigns;
         vector<ResolveAttachedClassItem> todoAttachedClassItems;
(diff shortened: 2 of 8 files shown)


froydnj:
> - [ ] @jez Create ticket: can we delete keep_for_typechecking, and just expand
> 
>   `ast::Cast` to hold onto its type expression?

This is #5823, which is possibly also helpful for identifying fields in namer for the LSP fast path.

froydnj (approved):

Issue #4978, opened 2021-12-07
Require generic classes to have types specified during class instantiation


telekid:
### Problem
This request is a followup to https://github.com/sorbet/sorbet/issues/4965, which I was believe was closed due to a misunderstanding on my part. In that issue, I requested that Sorbet require that type parameters be specified during type declaration. @jez pointed out that Sorbet already enforces that. I went back and reproduced the issue I was running into. It turns out that while Sorbet enforces that within type declaration, it doesn't enforce that during class instantiation. See [this playground](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20Box%0A%20%20extend%20T%3A%3AGeneric%0A%20%20Elem%20%3D%20type_member%0Aend%0A%0Asig%20%7Breturns%28Box%5BString%5D%29%7D%0Adef%20foo%0A%20%20%23%20I%20believe%20a%20type%20parameter%20should%20be%20required%20here%3A%0A%20%20box%20%3D%20Box.new%0A%20%20T.reveal_type%28box%29%0A%0A%20%20%23%20Note%20that%20%0A%20%20box%0Aend%0A). 

### Proposed Solution
Make it possible for Sorbet to mandate that all generic classes have types specified during instantiation.

jez:
Thanks for the clarification, this definitely seems like a bug!

jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20Box%0A%20%20extend%20T%3A%3AGeneric%0A%20%20Elem%20%3D%20type_member%0Aend%0A%0Asig%20%7Breturns%28Box%5BString%5D%29%7D%0Adef%20foo%0A%20%20%23%20I%20believe%20a%20type%20parameter%20should%20be%20required%20here%3A%0A%20%20box%20%3D%20Box.new%0A%20%20T.reveal_type%28box%29%0A%0A%20%20%23%20Note%20that%20%0A%20%20box%0Aend%0A)

```ruby
# typed: true
extend T::Sig

class Box
  extend T::Generic
  Elem = type_member
end

sig {returns(Box[String])}
def foo
  # I believe a type parameter should be required here:
  box = Box.new
  T.reveal_type(box)

  # Note that 
  box
end

```

#### Observed output

```
[...]

Issue #4450, opened 2021-08-03
`Set.new` does not infer types from passed arguments


ghiculescu:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aarr%20%3D%20%5B1%2C2%2C3%5D%0AT.reveal_type%20arr%0A%0AT.reveal_type(Set.new(arr))%0AT.assert_type!(Set.new(arr)%2C%20T%3A%3ASet%5BString%5D)%20%23%20this%20should%20fail%0A)

```ruby
# typed: true

arr = [1,2,3]
T.reveal_type arr

T.reveal_type(Set.new(arr))
T.assert_type!(Set.new(arr), T::Set[String]) # this should fail
```

#### Observed output

```

# typed: true

arr = [1,2,3]
T.reveal_type arr

T.reveal_type(Set.new(arr))
T.assert_type!(Set.new(arr), T::Set[String]) # this should fail



editor.rb:4: Revealed type: [Integer(1), Integer(2), Integer(3)] (3-tuple) https://srb.help/7014
     4 |T.reveal_type arr
        ^^^^^^^^^^^^^^^^^
  Got [Integer(1), Integer(2), Integer(3)] (3-tuple) originating from:
    editor.rb:3:
     3 |arr = [1,2,3]
              ^^^^^^^

editor.rb:6: Revealed type: T::Set[T.untyped] https://srb.help/7014
     6 |T.reveal_type(Set.new(arr))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Got T::Set[T.untyped] originating from:
    editor.rb:6:
     6 |T.reveal_type(Set.new(arr))
                      ^^^^^^^^^^^^
Errors: 2
```

#### Expected behavior

I'd expect sorbet to know that `Set.new(arr)` returns a `T::Set[Integer]`, since the type of `arr` was `T::Array[Integer]`.

jez:
There is a good test case in this issue I just closed as a dupe, which we will want to include in any PR that fixes this in the future:

https://github.com/sorbet/sorbet/issues/1804

jez:
Note that for the specific case of `Set`, you can use `Set[...]` instead of `Set.new(...)` to get a better inferred type

[...]

PR #6039, opened 2022-07-18
Suggest `T.untyped` for non-stdlib generics too

### LINKED ISSUE #5512: Sorbet should insert `[T.untyped]` for generic class without type args
#### Input

[→ View on sorbet.run](https://sorbet.run/?arg=--suggest-unsafe#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20BecomesGeneric%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20X%20%3D%20type_member%0Aend%0A%0Asig%20%7Bparams%28x%3A%20BecomesGeneric%29.void%7D%0Adef%20example%28x%29%0Aend)

```ruby
# typed: true
extend T::Sig

class BecomesGeneric
  extend T::Generic

  X = type_member
end

sig {params(x: BecomesGeneric).void}
def example(x)
end
```

#### Observed output

```
editor.rb:10: Malformed type declaration. Generic class without type arguments BecomesGeneric https://srb.help/5045
    10 |sig {params(x: BecomesGeneric).void}
                       ^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

Commonly when adding a `type_member` to something that was not previously a generic class, it will be used in type positions where it was not applied to a type argument.

When Sorbet is running in `--suggest-unsafe` mode, it should offer an autocorrect that inserts `[T.untyped]` when required.

- - -

jez:
Fixes #5512

### Motivation

- I was working on a fix for #4978 / #4450, and was hoping to use this helper
  function.
- But then I got side tracked, as my initial implementation for those tickets
  didn't quite work out.
- Along the way, I made the autocorrect work for all generic classes, not just
  ones in the standard library.

So I'd like to get that autocorrect fix in. Also I think that it's useful to
have all these special stdlib classes all in one place, because if you add
another one, you don't have to track everything down.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -579,6 +579,60 @@ bool ClassOrModuleRef::isPackageSpecSymbol(const GlobalState &gs) const {
     return false;
 }
 
+bool ClassOrModuleRef::isBuiltinGenericForwarder() const {
+    return *this == Symbols::T_Hash() || *this == Symbols::T_Array() || *this == Symbols::T_Set() ||
+           *this == Symbols::T_Range() || *this == Symbols::T_Enumerable() || *this == Symbols::T_Enumerator() ||
+           *this == Symbols::T_Enumerator_Lazy();
+}
+
+ClassOrModuleRef ClassOrModuleRef::maybeUnwrapBuiltinGenericForwarder() const {
+    if (*this == Symbols::T_Array()) {
+        return Symbols::Array();
+    } else if (*this == Symbols::T_Hash()) {
+        return Symbols::Hash();
+    } else if (*this == Symbols::T_Enumerable()) {
+        return Symbols::Enumerable();
+    } else if (*this == Symbols::T_Enumerator()) {
+        return Symbols::Enumerator();
+    } else if (*this == Symbols::T_Enumerator_Lazy()) {
+        return Symbols::Enumerator_Lazy();
+    } else if (*this == Symbols::T_Range()) {
+        return Symbols::Range();
+    } else if (*this == Symbols::T_Set()) {
+        return Symbols::Set();
+    } else {
+        return *this;
+    }
+}
+
+ClassOrModuleRef ClassOrModuleRef::forwarderForBuiltinGeneric() const {
+    if (*this == Symbols::Array()) {
+        return Symbols::T_Array();
+    } else if (*this == Symbols::Hash()) {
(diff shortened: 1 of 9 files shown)


core/Symbols.cc
```diff
@@ -579,6 +579,60 @@ bool ClassOrModuleRef::isPackageSpecSymbol(const GlobalState &gs) const {
...
+    } else if (*this == Symbols::Enumerator_Lazy()) {
+        return Symbols::T_Enumerator_Lazy();
+    } else if (*this == Symbols::Range()) {
+        return Symbols::T_Range();
+    } else if (*this == Symbols::Set()) {
+        return Symbols::T_Set();
+    } else {
+        return Symbols::noClassOrModule();
+    }
+}
```
froydnj, line 626: This is a nice cleanup.

WDYT about factoring out a constexpr array of stdlib, generic pairs, then updating both of these functions to be basically calls to `std::find_if`, so that it is impossible to forget to update one without the other?  (Ideally the compiler will unroll the loops so that the generated code is equivalent...bonus points for making the stdlib symbols and the generic symbols consecutive so that the functions boil down to just a range check rather than an if-chain.)

I guess once a table is made, the predicate functions before and after this block could just become table lookups too, furthering the goal of not missing places to update if need be?

aadi-stripe (approved):

core/Symbols.cc
jez: I'm not sure. I'm getting this error from clang:

```plain
clang: Constexpr variable cannot have non-literal type 'const vector<pair<core::ClassOrModuleRef, core::ClassOrModuleRef>>'

vector:471: 'vector<std::pair<sorbet::core::ClassOrModuleRef, sorbet::core::ClassOrModuleRef>>' is not literal because it is not
an aggregate and has no constexpr constructors other than copy or move constructors
```

I'm happy to take this as a follow up, but I'm not quite sure how to solve this. In the mean time, I think the current organization is better than having it scattered throughout the codebase, and I've already got enough stacked PRs at the moment. So I'll merge for now and follow up later.

froydnj: Yeah, we are not on the version of C++ that permits constexpr dynamic memory allocation (:exploding_head:).  Using regular C arrays should work just fine, though.

PR #6040, opened 2022-07-18
Fix crash calling ZSuperArgs from within block

### LINKED ISSUE #6038: enforced condition blockArg == nullptr has failed: Block arg was not in final position
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Adef%20raise%28%29%0A%20%20callcc%20do%20%7Ccontinuation%7C%0A%20%20%20%20super%0A%20%20end%0Aend) (I don't see an option to run the debug compiler there though)

```ruby
# typed: true

def raise()
  callcc do |continuation|
    super
  end
end
```

#### Observed output

With a debug build of recent master @ e166ca29292e766fedca2c4a02f757cd4778c57c

```
Exception::raise(): local_vars/local_vars.cc:273 enforced condition blockArg == nullptr has failed: Block arg was not in final position

Backtrace:
sorbet::local_vars::LocalNameInserter::lowerZSuperArgs(sorbet::core::MutableContext, sorbet::ast::ExpressionPtr) (in sorbet) + 1788
sorbet::local_vars::LocalNameInserter::postTransformSend(sorbet::core::MutableContext, sorbet::ast::ExpressionPtr&) (in sorbet) + 157
decltype(auto) sorbet::ast::MapFunctions<(sorbet::ast::TreeMapKind)1>::CALL_MEMBER_impl_postTransformSend<sorbet::local_vars::LocalNameInserter, true>::call<sorbet::core::MutableContext&, sorbet::ast::ExpressionPtr&>(sorbet::local_vars::LocalNameInserter&, sorbet::core::MutableContext&, sorbet::ast::ExpressionPtr&) (in sorbet) + 74
sorbet::ast::TreeMapper<sorbet::local_vars::LocalNameInserter, sorbet::core::MutableContext, (sorbet::ast::TreeMapKind)1, (sorbet::ast::TreeMapDepthKind)0>::mapSend(sorbet::ast::ExpressionPtr&, sorbet::core::MutableContext) (in sorbet) + 1125
sorbet::ast::TreeMapper<sorbet::local_vars::LocalNameInserter, sorbet::core::MutableContext, (sorbet::ast::TreeMapKind)1, (sorbet::ast::TreeMapDepthKind)0>::mapIt(sorbet::ast::ExpressionPtr&, sorbet::core::MutableContext) (in sorbet) + 212

...


tmp.rb:5: Failed to process tree (backtrace is above) https://srb.help/1001
     5 |    super
            ^^^^^
```

#### Expected behavior

It shouldn't crash.

- - -

Minimized from a crash when trying to type-check Homebrew/brew, based on [Ignorable.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/ignorable.rb)

jez:
### Motivation

Fixes #6038

### Test plan

See included automated tests.

### DIFF
--- a/local_vars/local_vars.cc
+++ b/local_vars/local_vars.cc
@@ -228,6 +228,14 @@ class LocalNameInserter {
             return tree;
         }
 
+        int i = scopeStack.size() - 1;
+        for (; i >= 0; i--) {
+            if (!scopeStack[i].insideBlock) {
+                break;
+            }
+        }
+        auto &enclosingMethodScopeStack = scopeStack[i];
+
         // In the context of a method with a signature like this:
         //
         //    def f(<posargs>,<kwargs>,&<blkvar>)
@@ -269,7 +277,7 @@ class LocalNameInserter {
         // We'll also look for the block arg, which should always be present at the end of the args.
         ast::ExpressionPtr blockArg;
 
-        for (auto arg : scopeStack.back().args) {
+        for (const auto &arg : enclosingMethodScopeStack.args) {
             ENFORCE(blockArg == nullptr, "Block arg was not in final position");
 
             if (arg.flags.isPositional()) {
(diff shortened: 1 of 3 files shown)


local_vars/local_vars.cc
```diff
@@ -228,6 +228,14 @@ class LocalNameInserter {
             return tree;
         }
 
+        int i = scopeStack.size() - 1;
```
varungandhi-src, line 231: Maybe add an explicit cast around the `scopeStack.size()` call? Otherwise, this code may be slightly confusing to read, because it's not necessarily intuitive whether the conversion to `int` happens before the subtraction (yes) or if it happens after the subtraction (in which case this would be a bug).

jez: I certainly hope we never encounter a file with 2^31 nested scopes in it (and I have a feeling something else breaks first). In any case, I've changed this to use `size_t` throughout, never converting to `int`.

local_vars/local_vars.cc
```diff
@@ -228,6 +228,14 @@ class LocalNameInserter {
             return tree;
         }
 
+        size_t i = scopeStack.size() - 1;
+        for (; i >= 0; i--) {
```
varungandhi-src, line 232: This loop condition will always be true if `i` has type `size_t`. 😅

jez: ah thanks

PR #6046, opened 2022-07-19
wip: Error if using `Box.new` without type args

### LINKED ISSUE #4450: `Set.new` does not infer types from passed arguments

### LINKED ISSUE #4978: Require generic classes to have types specified during class instantiation

jez:
### Motivation

Fixes #4978
Fixes #4450

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -374,8 +374,10 @@ ExpressionPtr desugarMlhs(DesugarContext dctx, core::LocOffsets loc, parser::Mlh
                 }
                 auto lhloc = lh.loc();
                 auto zlhloc = lhloc.copyWithZeroLength();
-                auto index = MK::Send3(lhloc, MK::Constant(lhloc, core::Symbols::Range()), core::Names::new_(), zlhloc,
-                                       MK::Int(lhloc, left), MK::Int(lhloc, -right), std::move(exclusive));
+                auto untypedRange = MK::Send1(lhloc, MK::Constant(lhloc, core::Symbols::T_Range()),
+                                              core::Names::squareBrackets(), zlhloc, MK::Untyped(lhloc));
+                auto index = MK::Send3(lhloc, move(untypedRange), core::Names::new_(), zlhloc, MK::Int(lhloc, left),
+                                       MK::Int(lhloc, -right), std::move(exclusive));
                 stats.emplace_back(MK::Assign(
                     lhloc, std::move(lh),
                     MK::Send1(loc, MK::Local(loc, tempExpanded), core::Names::slice(), zlhloc, std::move(index))));
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -147,6 +147,7 @@ TypePtr ClassOrModule::unsafeComputeExternalType(GlobalState &gs) {
                 // For backwards compatibility, instantiate stdlib generics
                 // with T.untyped.
                 targs.emplace_back(Types::untyped(gs, ref));
+                flags.externalTypeImplicitlyUntyped = true;
             } else if (tmData->flags.isFixed || tmData->flags.isCovariant) {
                 // Default fixed or covariant parameters to their upper
                 // bound.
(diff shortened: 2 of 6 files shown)


jez:

→ https://go/builds/bui_M6kI6owO145ovh
→ https://go/builds/bui_M6kIFXqfOkKGUv

jez:

→ https://go/builds/bui_M6kqBEyFHQYtPt
→ https://go/builds/bui_M6kqGEnhnzWjXt

jez:
This is turning into a huge yak shave. It causes ~5k errors on Stripe's codebase, some of which are autocorrectable, some which aren't.

Coupled with the fact that the heuristic that the PR implements is actually wrong, it's hard.

Example

```ruby
sig {params(x: T.class_of(MyGenericClass)).void}
def example(x)
  x.new
end
```

```ruby
class A
  extend T::Generic

  Elem = type_member
  
  def self.make
    new
  end
end
```

These are two problems I don't have a great solution for on the top of my head.

Also, I'm wondering whether it would be possible to infer the types from the constructor arguments one day. It's hard, because the user's `initialize` method doesn't have any `type_parameters` and the `new` method is inherited from `BasicObject`. Even if we tried to do something in the intrinsic, all the `new` intrinsic does is forward the args verbatim to `initialize` on the instance type. There's not an obvious API we could tap into to allow dispatchCallSymbol to know to e.g. treat each `Elem` as a distinct `T.type_parameter`. So I have no real idea how type inference would work.

Anyways, going to close for now.

PR #6052, opened 2022-07-20
Fork with a block returns an integer

lugray:
When called without a block, you get nil or an integer, to tell you which process you are. With a block, the return value only exists in the parent, so it's an integer.

### Motivation
Avoids `T.must { fork {} )`.


jez:

→ https://go/builds/bui_M5txkgh8whTFfS
→ https://go/builds/bui_M5txLofFVhO93N

jez (approved):
Thanks!

PR #6053, opened 2022-07-20
tighten up the code for copying trees

froydnj:
Tree copying creates a lot of unnecessary temporaries that the compiler doesn't appear to be able to optimize away.  Not sure if this change will make a difference, but at least the assembly looks a lot cleaner.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_M6A16gAO3eWrNF
→ https://go/builds/bui_M6A1fXvCOfY2Dj

PR #6056, opened 2022-07-20
Recover from lonely `def` keyword

### LINKED ISSUE #5443: Recover from `def` with nothing after it
#### Input

[→ View on sorbet.run](https://sorbet.run/?arg=--print=parse-tree-whitequark#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Bparams%28x%3A%20Integer%29.void%7D%0A%20%20def%0Aend)

```ruby
# typed: true
class A
  extend T::Sig
  sig {params(x: Integer).void}
  def
end
```

#### Observed output

```
s(:begin)
editor.rb:6: unexpected token "end of file" https://srb.help/2001
     6 |end
     7 |
Errors: 1
```

#### Expected behavior

Sorbet should still be able to parse this snippet to something that sees both the class def and the preceeding sig.

Maybe the `def` should be parsed as though the user wrote `self.def()` so that we can provide completion results on the `def`.

- - -

jez:
### Motivation

Fixes #5443

This situation becomes more prominent after the changes in #5808 landed,
which allow us to take the fast path when adding a new method.

### Test plan

See included automated tests.


parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -927,28 +928,71 @@ int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser:
                       driver.lex.context.allowNumparams = false;
                     }
 
-        defn_head: k_def def_name
+       defn_head: k_def def_name
```
jez, line 931: Some of these rules were misaligned from all the others by 1 char. The diff looks better ignoring whitespace.

jez:

→ https://go/builds/bui_M5zdOwnLqiZZxh
→ https://go/builds/bui_M5zdlSO3sEp3xF

parser/parser/codegen/generate_diagnostics.cc
```diff
@@ -6,6 +6,8 @@ using namespace std;
 namespace {
 
 tuple<string, string> MESSAGES[] = {
+    {"InternalError", "internal error while parsing: {}"},
```
froydnj, line 9: (non-blocking) Is it worth making this particular error translate to an actual internal error from `core/`?

core/tools/generate_names.cc
```diff
@@ -59,6 +59,7 @@ NameDef names[] = {
 
     // Used in parser for error recovery
     {"methodNameMissing", "<method-name-missing>"},
+    {"methodDefNameMissing", "<method-def-name-missing>"},
```
froydnj, line 62: Do we need to add this to all the places that already check `methodNameMissing`?

froydnj (approved):
👍 

Just a few comments below.

parser/parser/cc/grammars/typedruby.ypp
```diff
@@ -927,28 +928,71 @@ int yylex(parser::semantic_type *lval, ruby_parser::location *lloc, ruby_parser:
...
                     {
                       auto defNameToken = $2->token;
+                      ForeignPtr node = nullptr;
+                      if (driver.rewind_if_different_line($1, defNameToken)) {
+                        node = driver.build.defnHeadError(self, $1);
+                      } else {
+                        node = driver.build.defnHead(self, $1, defNameToken);
```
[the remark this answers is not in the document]
jez: Good question, and no, because the guard at the top of `rewind_if_different_line` exits early on the first parse (when `indentationAware` is false).

It does mean that our heuristic could recover in a weird way in a file that already had a syntax error though. For example:

```ruby
class A
  def
    foo
  end
  def bar
end
```

The changes in this PR recover from that in a weird way:

```ruby
class A
  def
    foo
  end
  def bar
end
```

We're ~okay with this (it could obviously be better) because this heuristic only applies when the first parse already produced no result, so it doesn't exactly matter what specifically we produce in the name of recovery, with the assumption being that something is better than nothing.

core/tools/generate_names.cc
jez: Good question: the only two places we use methodNameMissing to silence errors are:

- call to this method, as an error was already reported. it's not possible at the moment for this NameRef to show up in a `Send` node as far as I can see, so not there.
- in a completion result. we do not yet have an `isMethodDef` case in completion.cc yet, though we will when I get around to polishing #5886.

parser/parser/codegen/generate_diagnostics.cc
jez: Good point, will do.

PR #6057, opened 2022-07-20
Add Reline and IRB::ReidlineInputMethod IRB definitions

gmcgibbon:
Add RBI def for [`Reline`](https://github.com/ruby/reline) in the standard library. IRB uses this under the hood.

Also add RBI def [`IRB::ReidlineInputMethod`](https://github.com/ruby/irb/blob/ff129f3794f15b651e6576fd560d39316257171c/lib/irb/input-method.rb#L265) in the standard library. This is the Reline input method adapter IRB uses by default.

Unfortunately, Reline isn't documented much, but I can try to add some basic comments to the type def if needed. Let me know what is desired here.

### Motivation

I was making a custom console object for Rails to use via `Rails.application.config.console` that delegates to IRB, and Sorbet didn't understand my references to `IRB::ReidlineInputMethod`.

### Test plan

These are type definitions, so I don't think they need tests.


rbi/stdlib/reline.rbi
```diff
@@ -0,0 +1,223 @@
...
+  def_single_delegators :core, :basic_quote_characters, :basic_quote_characters=
+  def_single_delegators :core, :completer_quote_characters, :completer_quote_characters=
+  def_single_delegators :core, :filename_quote_characters, :filename_quote_characters=
+  def_single_delegators :core, :special_prefixes, :special_prefixes=
+  def_single_delegators :core, :completion_proc, :completion_proc=
+  def_single_delegators :core, :output_modifier_proc, :output_modifier_proc=
+  def_single_delegators :core, :prompt_proc, :prompt_proc=
+  def_single_delegators :core, :auto_indent_proc, :auto_indent_proc=
+  def_single_delegators :core, :pre_input_hook, :pre_input_hook=
+  def_single_delegators :core, :dig_perfect_match_proc, :dig_perfect_match_proc=
```
jez, line 180: Sorbet does not understand these methods, and thus these methods will not be present.

I'd prefer we delete these, or declare what methods will actually be defined using normal `def` nodes.

rbi/stdlib/reline.rbi
```diff
@@ -0,0 +1,223 @@
...
+      :special_prefixes,
+      :completion_proc,
+      :output_modifier_proc,
+      :prompt_proc,
+      :auto_indent_proc,
+      :pre_input_hook,
+      :dig_perfect_match_proc,
+    ]
+
+    attr_reder :completion_append_character
```
jez, line 75: What is `attr_reder`?

rbi/stdlib/reline.rbi
```diff
@@ -0,0 +1,223 @@
+# typed: __STDLIB_INTERNAL
+
+module Reline
```
jez, line 3: Overall, would you mind committing an RBI file that has been generated by `srb rbi gems` or `tapioca`?

Doing so will automatically expand things like `Struct.new`, `def_single_delegator`, attr_reader`, etc. to normal method defs, which will make it easier for people to incrementally add or fix types in the future. For example, it's not easy right now to provide a type annotation for a single method of a class created with `Struct.new`.

jez (changes requested):

rbi/stdlib/reline.rbi
gmcgibbon: 🤦 A typo. Sorry, I had to manually write this because I didn't sorbet would understand this line: https://github.com/ruby/reline/blob/b6ffa1714723b55a1e65134e49b2ed9da0560dc0/lib/reline.rb#L79

rbi/stdlib/reline.rbi
gmcgibbon: I think I did this correctly. Let me know if there's anything else I should do.

jez:
Sorry I didn't catch this earlier. I can't start the CI run on your branch until you rebase on a newer master commit. You are currently based on a commit from Mar 23, and that is too old (we have made some incompatible CI changes since then).

gmcgibbon:
Rebased!

jez:

→ https://go/builds/bui_M5z9yXrkNDMyLW
→ https://go/builds/bui_M5z9ksfzjwJZLQ

jez (changes requested):
Sorbet fails to typecheck the new RBI files; you should be able to see the problems in the CI run, but if you can't let me know.

gmcgibbon:
Oops, I forgot to re-define the element in the array subclass. I think that should fix it. Thanks for your patience.

jez:

→ https://go/builds/bui_M7pxouxjnR635F
→ https://go/builds/bui_M7pxXshOURCoYx

jez:

→ https://go/builds/bui_M7rd0q4rR6iYKV
→ https://go/builds/bui_M7rdDfZfgsgfWV

gmcgibbon:
I don't understand how to fix the compiler failures. When you have time, could you please have a look and advise? Thanks!

gmcgibbon:
I ran `./tools/scripts/update_compiler_exp.sh` which should fix the previous build errors.

jez (approved):

jez:
Thanks! Sorry about the problems with the compiler, I'll try to make some changes so that adding a new RBI is easier.

PR #6058, opened 2022-07-21
Recover from const assign in method

jez:
### Motivation

Better parse error recovery, which lets #5808 kick in more frequently.

### Test plan

See included automated tests.

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -63,6 +63,7 @@ NameDef names[] = {
     {"cvarNameMissing", "@@<cvar-name-missing>"},
     {"ConstantNameMissing", "<ConstantNameMissing>", true},
     {"ErrorNode", "<ErrorNode>", true},
+    {"dynamicConstAssign", "<dynamic-const-assign>"},
 
     // used in CFG for temporaries
     {"whileTemp", "<whileTemp>"},
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -176,7 +176,7 @@ public:
         return cond;
     }
 
-    void error(ruby_parser::dclass err, core::LocOffsets loc, std::string data = "") {
+    void error_without_recovery(ruby_parser::dclass err, core::LocOffsets loc, std::string data = "") {
         driver_->external_diagnostic(ruby_parser::dlevel::ERROR, err, loc.beginPos(), loc.endPos(), data);
     }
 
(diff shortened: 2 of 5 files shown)


parser/Builder.cc
```diff
@@ -176,10 +176,15 @@ class Builder::Impl {
...
-    void error(ruby_parser::dclass err, core::LocOffsets loc, std::string data = "") {
+    void error_without_recovery(ruby_parser::dclass err, core::LocOffsets loc, std::string data = "") {
         driver_->external_diagnostic(ruby_parser::dlevel::ERROR, err, loc.beginPos(), loc.endPos(), data);
     }
 
+    void error(ruby_parser::dclass err, core::LocOffsets loc, std::string data = "") {
+        auto range = ruby_parser::diagnostic::range(loc.beginPos(), loc.endPos());
+        driver_->diagnostics.emplace_back(ruby_parser::dlevel::ERROR, err, std::move(range), data);
+    }
+
```
froydnj, line 187: I don't have better names for these, but I am not sure that `error_without_recovery` is the best name here.  It makes me think that we're raising exceptions or similar.  (Unless we think that `error_with_recovery` is temporary and we're planning on driving it down, in which case, w/e)

Also, the magic of `external_diagnostic` is just gross -- not your fault, just commenting in passing.

core/tools/generate_names.cc
```diff
@@ -63,6 +63,7 @@ NameDef names[] = {
     {"cvarNameMissing", "@@<cvar-name-missing>"},
     {"ConstantNameMissing", "<ConstantNameMissing>", true},
     {"ErrorNode", "<ErrorNode>", true},
+    {"dynamicConstAssign", "<dynamic-const-assign>"},
```
froydnj, line 66: It is a little surprising to me that we don't need to check for this anywhere so it doesn't leak into error messages.

froydnj (approved):

parser/Builder.cc
jez: Yeah, my plan is to eliminate it. I have some work locally, and plan to keep iterating soon.

core/tools/generate_names.cc
jez: Good point. I thought that maybe I could get it to trigger with a pinning error:

```ruby
def foo
  X = false
  1.times do
    X = true
  end
end
```

but it turns out that we don't print the variable name in the pinning error:

```
pin.rb:4: dynamic constant assignment https://srb.help/2001
     4 |  X = false
          ^

pin.rb:6: dynamic constant assignment https://srb.help/2001
     6 |    X = true
            ^

pin.rb:6: Changing the type of a variable in a loop is not permitted https://srb.help/7001
     6 |    X = true
                ^^^^
  Existing variable has type: FalseClass
  Attempting to change type to: TrueClass

    pin.rb:4: Replace with T.let(false, T::Boolean)
     4 |  X = false
              ^^^^^
Errors: 3
```

I will think about whether we need to silence this somewhere.

jez: I think the answer is that it's already quite rare for us to print variable names, because CFG creates so many nonsensical variable names (`<statTemp>$3>` etc.)

froydnj: That makes a certain amount of sense; I guess since we're generating the error from the parser and not from someplace like resolver, `dynamicConstAssign` is basically a variable for the rest of the pipeline.

PR #6060, opened 2022-07-21
Use parent class type members' variance in glb

jez:
### Motivation

This allows the `result.rb` test to typecheck, and fixes one of our `List`
tests.

### Explanation

The intuition is that the result `ret` of `Types::all(gs, t1, t2)` must be a
subtype of both `t1` and `t2` (this is `ENFORCE`'d in the code).

Consider

```ruby
T.all(Result::Ok[T.untyped], Result[Integer, TypeError])
```

Intuitively, we would like this to be `Result::Ok[Integer]`. This intuition is
confirmed by the fact that it's the variance of the parent's type members that
are used when checking subtyping: since `OkType` and `ErrType` are covariant,
subtyping should be allowed, so since `T.noreturn <: TypeError` holds,
`Result::Ok[Integer, T.noreturn] <: Result[Integer, TypeError]` holds.

Due to what looks to me like an accident, `glb` was using the variance of the
child to determine whether it could collapse a type argument to a subtype, which
is now changed.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/subtyping.cc
+++ b/core/types/subtyping.cc
@@ -954,8 +954,7 @@ TypePtr Types::glb(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2) 
         vector<TypePtr> newTargs;
         newTargs.reserve(a1->klass.data(gs)->typeMembers().size());
         int j = 0;
-        for (SymbolRef idx : a1->klass.data(gs)->typeMembers()) {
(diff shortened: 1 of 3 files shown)


jez:

→ https://go/builds/bui_M6FFJWD1dCRhsN
→ https://go/builds/bui_M6FFyWRycHUUKR

Issue #37, opened 2019-01-16
Cannot overload functions with keyword arguments


DarkDimius:
This is a problem for Dir.glob in Ruby 2.5, which adds a `base` kwarg:
https://git.corp.stripe.com/stripe-internal/sorbet/pull/1664/files#diff-95d55ac4f1672115ce37347e7d829478R96

Originally reported by @dug-stripe at https://jira.corp.stripe.com/browse/RUBYPLAT-1446

dee-see:
Is there a workaround for this?

dee-see:
For the moment I created this...

```ruby
    class DirHelper
      extend T::Sig

      sig { params(pattern: String, base: String).returns(T::Array[String]) }
      def self.glob(pattern, base)
        T.unsafe(Dir).glob(pattern, base: base)
      end
    end
```

jez:
#### Input

[→ View on sorbet.run](TODO_replace_this_with_your_sorbet_run_link)

```ruby
# typed: __STDLIB_INTERNAL

class A
  extend T::Sig

  sig {params(x: Integer).returns(Integer)}
  sig {params(x: Integer, y: String).returns(String)}
  def foo(x, y: '')
  end
end
```

#### Observed output

```
editor.rb:7: Malformed sig. Overloaded functions cannot have keyword arguments:  y https://srb.help/5003
     7 |  sig {params(x: Integer, y: String).returns(String)}
                                  ^
Errors: 1
```



#### Expected behavior



Ideally we would be able to allow keyword args to participate in overload checking.

In reality, it's quite complex, given the current state of Sorbet's implementation.

This might change when Sorbet supports enough of the keyword args changes from Ruby 2.7.
The trick there is that in Ruby 2.7, the parser makes a distinction between keyword args and a positional hash argument. That might be enough to starts supporting this feature.

- - -

dee-see:
Seems like this works now, possibly due to #3490?

I think it can be closed @jez  :tada:

dee-see:
Hmmm I guess the example in https://github.com/sorbet/sorbet/issues/37#issuecomment-710707518 still doesn't work, but my `Dir.glob` calls don't create sorbet warnings anymore...

dduugg:
Still present in 0.6.13363, and there's a second half to this that I don't think is captured yet: **the documented workaround silently selects the wrong overload instead of erroring.**

[...]

PR #6062, opened 2022-07-21
Add missing chomp param to String#each_line

lugray:
### Motivation
The documentation mentions the chomp parameter, but the signatures excluded it.

### DIFF
--- a/rbi/core/string.rbi
+++ b/rbi/core/string.rbi
@@ -975,6 +975,7 @@ class String < Object
   sig do
     params(
         arg0: String,
+        chomp: T::Boolean,
         blk: T.proc.params(arg0: String).returns(BasicObject),
     )
     .returns(String)
@@ -982,10 +983,11 @@ class String < Object
   sig do
     params(
         arg0: String,
+        chomp: T::Boolean,
     )
     .returns(T::Enumerator[String])
   end
-  def each_line(arg0=T.unsafe(nil), &blk); end
+  def each_line(arg0=T.unsafe(nil), chomp: false, &blk); end
 
   # Returns `true` if *str* has a length of zero.
   #


froydnj:

→ https://go/builds/bui_M6AAc79Vu7yawB
→ https://go/builds/bui_M6AAz9JIkupH0y

froydnj (changes requested):
Unfortunately this doesn't work, as overloaded functions can't have keyword arguments specified in their associated `sig`s.  This has been known for a long time as #37, but is not currently a priority for us to fix.

lugray:
@froydnj can you take another look? I removed the overload to allow the keyword argument.

froydnj (changes requested):
There are two interrelated problems with this.  The first is that the change declares that the block can return an enumerator when it should be declaring that the method itself can return a `T.any(String, T::Enumerator[String])`; the block should continue to return `BasicObject`.

The second is that I think it will break existing code that does:

`str.each_line.map do |line| ... end`

because now Sorbet will check that `T.any(String, T::Enumerator[String])` has `#map`, which is not true (`String#map` does not exist), and the user will have to do some gymnastics to convince Sorbet otherwise.

Consider [this sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A%22input%20file%22.each_line.map%20do%20%7Cline%7C%0A%20%20puts%20line%0Aend%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20do%0A%20%20%20%20params%28%0A%20%20%20%20%20%20%20%20arg0%3A%20String%2C%0A%20%20%20%20%20%20%20%20chomp%3A%20T%3A%3ABoolean%2C%0A%20%20%20%20%20%20%20%20blk%3A%20T.nilable%28T.proc.params%28arg0%3A%20String%29.returns%28BasicObject%29%29%2C%0A%20%20%20%20%29%0A%20%20%20%20.returns%28T.any%28String%2C%20T%3A%3AEnumerator%5BString%5D%29%29%0A%20%20end%0A%20%20def%20each_line%28arg0%3DT.unsafe%28nil%29%2C%20chomp%3A%20false%2C%20%26blk%29%3B%20T.unsafe%28nil%29%3B%20end%0Aend%0A%0AA.new.each_line.map%20do%20%7Cline%7C%0A%20%20puts%20line%0Aend) which attempts to reproduce the change supplied here on a separate class.

lugray:
Oops, yep, I mad a mistake with what I was trying to do, for sure.  I've fixed that. And yes, it will break existing code, but it can be fixed by casting the result, since if you call it without a block the author will know that it returns the enumerator. Without any change, sorbet gives the confusing error `Expected String but found {chomp: TrueClass} for argument arg0` when `each_line(chomp: true)` is used as ruby allows, and the only way I know to fix it is with `T.unsafe` which seems worse than casting.

froydnj (commented):
My sense is that requiring casts in the non-block case (and possibly in the block case, depending on how the call is made) is less ergonomic than requiring `T.unsafe`.

Eyeballing Stripe's codebase suggests that this will require a codemod against ~dozens of callsites.

lugray:
My counterpoint is that the signature as I have it is correct, and the current signature is incorrect. If incorrect and more ergonomic is the direction you'd like to go, that's understandable, but not the trade-off I'd choose to make.

lugray:
Has any work been attempted on allowing overloaded signatures with keyword arguments? I'd be interested in spending some time pursuing this, and would love a starting point.

froydnj (commented):
We talked this over internally and our decision is that multiple signatures (even without the correct keyword args) are preferred over a single signature that unions the return types.

Fully fixing #37 probably requires implementing support for Ruby 3 keyword arguments.  The heart of method dispatch lives here:

https://github.com/sorbet/sorbet/blob/336ccd4cd511f9a572cb961de98d0204ece926dd/core/types/calls.cc#L677-L686

with overload resolution, such as it is, implemented in:

https://github.com/sorbet/sorbet/blob/336ccd4cd511f9a572cb961de98d0204ece926dd/core/types/calls.cc#L323-L326

The core problem right now (aside from Ruby keyword argument parsing being weird prior to Ruby 3, and even still a little weird after that) is that we choose which overloaded sig we're using before we parse out keyword arguments, so we'd have to move guessing the overload after the argument parsing code.  The argument parsing code is admittedly a mess and could probably stand a cleanup to use something like `absl::Span` rather than a million iterators everywhere.

We think there might be a shorter path forward; the complaint about keyword arguments in overloaded method signatures comes from:

https://github.com/sorbet/sorbet/blob/336ccd4cd511f9a572cb961de98d0204ece926dd/resolver/resolver.cc#L3346-L3351

But in practice, the common case that people want to fix (as in this PR) is:

* two signatures;
* with identical parameter types;
* except one has a block and one does not.

We think that if that special case, and that special case only, was permitted, then everything in `guessOverload` and `dispatchCallSymbol`, above, would work just like one would want.  I was going to try implementing that this week (and maybe part of next).

lugray:
I can validate your guess that if the special case is permitted, `guessOverload` and `dispatchCallSymbol` will continue to work. With
```diff
diff --git a/rbi/core/string.rbi b/rbi/core/string.rbi
index 6b37d46b8..e248011e0 100644
--- a/rbi/core/string.rbi
+++ b/rbi/core/string.rbi
@@ -975,6 +975,7 @@ def each_grapheme_cluster; end
   sig do
     params(
         arg0: String,
+        chomp: T::Boolean,
         blk: T.proc.params(arg0: String).returns(BasicObject),
     )
     .returns(String)
@@ -982,10 +983,11 @@ def each_grapheme_cluster; end
   sig do
     params(
         arg0: String,
+        chomp: T::Boolean,
     )
     .returns(T::Enumerator[String])
   end
-  def each_line(arg0=T.unsafe(nil), &blk); end
+  def each_line(arg0=T.unsafe(nil), chomp: false, &blk); end
 
   # Returns `true` if *str* has a length of zero.
   #
diff --git a/resolver/resolver.cc b/resolver/resolver.cc
index fb7d43c91..b50845a22 100644
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -3430,13 +3430,6 @@ private:
                     }
                 }
             }
-
-            if (isOverloaded && arg.flags.isKeyword) {
-                if (auto e = ctx.state.beginError(arg.loc, core::errors::Resolver::InvalidMethodSignature)) {
-                    e.setHeader("Malformed `{}`. Overloaded functions cannot have keyword arguments:  `{}`", "sig",
-                                treeArgName.show(ctx));
-                }
-            }
         }
 
         for (const auto &spec : sig.argTypes) {
```
I get a success for
```
$ bazel-bin/main/sorbet -e '"".each_line(chomp: true).any?; "".each_line(chomp: true, &:itself).capitalize'
```
And failures for
```
bazel-bin/main/sorbet -e '"".each_line(chomp: true).capitalize; "".each_line(chomp: true, &:itself).any?' 
👋 Hey there! Heads up that this is not a release build of sorbet.
Release builds are faster and more well-supported by the Sorbet team.
Check out the README to learn how to build Sorbet in release mode.
To forcibly silence this error, either pass --silence-dev-message,
or set SORBET_SILENCE_DEV_MESSAGE=1 in your shell environment.

-e:1: Method capitalize does not exist on T::Enumerator[String] https://srb.help/7003
     1 |"".each_line(chomp: true).capitalize; "".each_line(chomp: true, &:itself).any?
                                  ^^^^^^^^^^
  Got T::Enumerator[String] originating from:
    -e:1:
     1 |"".each_line(chomp: true).capitalize; "".each_line(chomp: true, &:itself).any?
        ^^^^^^^^^^^^^^^^^^^^^^^^^
  Did you mean initialize? Use `-a` to autocorrect
    -e:1: Replace with initialize
     1 |"".each_line(chomp: true).capitalize; "".each_line(chomp: true, &:itself).any?
                                  ^^^^^^^^^^
    https://github.com/sorbet/sorbet/tree/master/rbi/core/enumerator.rbi#L253: Defined here
     253 |  def initialize(arg0 = T.unsafe(nil), &blk); end
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

-e:1: Method any? does not exist on String https://srb.help/7003
     1 |"".each_line(chomp: true).capitalize; "".each_line(chomp: true, &:itself).any?
                                                                                  ^^^^
  Got String originating from:
    -e:1:
     1 |"".each_line(chomp: true).capitalize; "".each_line(chomp: true, &:itself).any?
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 2
```

Given your plan to work on that this/next week, I'll hold off on doing anything more. If you decide not to do it, please let me know and I'll dig a little deeper into detecting the special case.

froydnj:
:wave: @lugray #6075 has landed, which should support keyword arguments in the sig for this particular method.  Let me know if you have any other questions!

lugray:
Great work, thank you for that! I've reworked this back to its original form which should now work.

froydnj:

→ https://go/builds/bui_M8pANoN32xGLqf
→ https://go/builds/bui_M8pADr0hx3bMZM

froydnj (approved):
Thanks!

PR #6068, opened 2022-07-21
Add intrinsic for `sample` on `TupleType`

jez:
### Motivation

Requested on Slack.

### Test plan

See included automated tests.

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -404,6 +404,7 @@ NameDef names[] = {
     {"min"},
     {"max"},
     {"sum"},
+    {"sample"},
 
     // Argument forwarding
     {"fwdArgs", "<fwd-args>"},
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3054,6 +3054,34 @@ public:
     }
 } Tuple_sum;
 
+class Tuple_sample : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        auto *tuple = cast_type<TupleType>(args.thisType);
+        ENFORCE(tuple);
+
+        if (args.args.size() > 1) {
+            return;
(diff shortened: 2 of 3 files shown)


jez:

→ https://go/builds/bui_M6JagbNNebXOT2
→ https://go/builds/bui_M6JaVW6KI8a2dP

froydnj (approved):

will:
I think it is incorrect to make the result of sample be not nilable. A 3 tuple doesn't always stay a 3 tuple

https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aa%20%3D%20%25w%5Ba%20b%20c%5D%0Aa.reject!%20%7B%20true%20%7D%0Athis_is_actually_nil%20%3D%20a.sample%0AT.reveal_type%20this_is_actually_nil

```ruby
# typed: strict

a = %w[a b c]
a.reject! { true }
this_is_actually_nil = a.sample
T.reveal_type this_is_actually_nil # Revealed type: String
```

jez:
Yeah, that's a problem with tuples more generally, and is a reason why tuples are still experimental

jez:
The bug you mention already existed before this PR:

https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aa%20%3D%20%25w%5Ba%20b%20c%5D%0Aa.reject!%20%7B%20true%20%7D%0Athis_is_actually_nil%20%3D%20a%5B0%5D%0AT.reveal_type%20this_is_actually_nil

In the future we will have to teach Sorbet how to deal with methods that mutate the type of the method receiver on tuple types.

will:
Ok. I will drop a T.must around a sample I'm doing, since for me I expect it to never reject everything in the first place.

But as for tuples being experimental, I'm not intentionally using them. I'm just making an array and sorbet is deciding to use the experimental feature without my knowledge. I'd be happy for sorbet to think my array is always an array, but I don’t know how.

PR #6073, opened 2022-07-22
Maybe fix crash with addLocLine

jez:
### Motivation

I saw a segfault in our production logs that looked like it was centered
on this call to `buf.write`. I didn't bother to try to minimize and
reproduce a crash, but this refactoring should be equivalent and maybe
fixes the crash.

### Test plan

existing tests to see if there's a regression

untested to see if it fixes the root crash

### DIFF
--- a/core/Loc.cc
+++ b/core/Loc.cc
@@ -111,11 +111,11 @@ void addLocLine(stringstream &buf, int line, const File &file, int tabs, int pos
     buf << rang::fgB::black << leftPad(to_string(line + 1), posWidth) << " |" << rang::style::reset;
     ENFORCE(file.lineBreaks().size() > line + 1);
     auto endPos = file.lineBreaks()[line + 1];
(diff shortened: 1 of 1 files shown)


froydnj (approved):
I agree this is worth a shot; I wonder if we ought to have a better interface for fetching lines out of `File` than having to carefully play games with indices here.  (The indexing games would, of course, be moved to `File`, but there are probably other places that have to do all this `+ 1` nonsense?)

Also `File::lineBreaks` being `const` but returning a non-`const` reference from a `mutable` `shared_ptr` == :scream: .

jez:
Yeah I almost sniped myself to make some changes here in the past and got bored. lazily computing line breaks is wild, but I tried to eagerly force it to be computed and it turned out to substantially bloat our memory footprint (turns out that most files don't have errors and then never end up computing this list of line breaks).

PR #6075, opened 2022-07-22
permit keywords for a limited set of overloaded methods

froydnj:
Full support for keyword arguments in overloaded methods has been a requested feature for a long time (#37).  Actually writing the PR to fix that is liable to be a substantial investment of time that we do not have time to scope out right now.

But we thought of a more limited way to make things work: if a method has multiple signatures, we can permit keyword arguments if there are only two signatures and the signatures have identical arguments except for the block argument.  (I suppose it's possible we could have more than two signatures, but I think pairwise checking all of them to ensure some hand-wavy set of constraints gets expensive, quickly.)

This PR is an attempt to implement that idea.

### Motivation

Trying to address usecases for typing a limited subset of methods in the stdlib like #6062.

### Test plan

This PR includes some tests.  We probably also want tests where there is an actual block arg, and there are some tests that require checking for compatibility between signatures, not just checking for signature/method definition compatibility.

This PR does make things slightly worse insofar as we no longer call out all keyword arguments in overloaded methods for each sig, just the first keyword argument on the first sig.  We could retain the old behavior at the cost of some extra allocations to record all keyword arguments, which did not seem worth it at this point.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -3406,15 +3406,15 @@ private:
             defParams.push_back(local);
 
             auto spec = absl::c_find_if(sig.argTypes, [&](const auto &spec) { return spec.name == treeArgName; });
-            bool isBlkArg = arg.name == core::Names::blkArg();
+            bool isSyntheticBlkArg = arg.name == core::Names::blkArg();
 
             if (spec != sig.argTypes.end()) {
                 ENFORCE(spec->type != nullptr);
 
                 // It would be nice to remove the restriction on more than these two specific binds, but that would
                 // raise a lot more errors
-                if (!isBlkArg && (spec->rebind == core::Symbols::MagicBindToAttachedClass() ||
-                                  spec->rebind == core::Symbols::MagicBindToSelfType())) {
+                if (!isSyntheticBlkArg && (spec->rebind == core::Symbols::MagicBindToAttachedClass() ||
+                                           spec->rebind == core::Symbols::MagicBindToSelfType())) {
                     if (auto e = ctx.state.beginError(spec->loc, core::errors::Resolver::BindNonBlockParameter)) {
                         e.setHeader("Using `{}` is not permitted here", "bind");
                         e.addErrorNote("Only block arguments can use `{}`", "bind");
@@ -3431,7 +3431,7 @@ private:
                 }
 
                 // We silence the "type not specified" error when a sig does not mention the synthesized block arg.
-                if (!isOverloaded && !isBlkArg && (sig.seen.params || sig.seen.returns || sig.seen.void_)) {
+                if (!isOverloaded && !isSyntheticBlkArg && (sig.seen.params || sig.seen.returns || sig.seen.void_)) {
                     // Only error if we have any types
                     if (auto e = ctx.state.beginError(arg.loc, core::errors::Resolver::InvalidMethodSignature)) {
                         e.setHeader("Malformed `{}`. Type not specified for argument `{}`", "sig",
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -3299,7 +3299,7 @@ private:
         send->addPosArg(ast::MK::Symbol(send->loc, method.data(ctx)->name));
     }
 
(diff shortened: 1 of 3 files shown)


froydnj:
This is at least ready for a first pass, and does work with the initial proposed signature in #6062.

I have a TODO to provide better error messages, but I did not put a lot of work into this given that this is a `__STDLIB_INTERNAL` feature and I do not expect many uses of it to come up.

resolver/resolver.cc
```diff
@@ -3643,6 +3695,40 @@ class ResolveSignaturesWalk {
...
+
+        // Unlike std::equal, absl::c_equal will test for equal sizes.
+        auto argsEqual = [&ctx](const auto &arg0, const auto &arg1) {
+            if (arg0.name != arg1.name) {
+                return false;
+            }
+            return core::Types::equiv(ctx, arg0.type, arg1.type);
+        };
+
+        // TODO(froydnj) better error messages for users trying to provide overloads with kwargs?
```
jez, line 3721: What would the error message look like right now?

With more time what would you have it show instead?

resolver/resolver.cc
```diff
@@ -3643,6 +3695,40 @@ class ResolveSignaturesWalk {
...
+
+        const auto &argv0 = args[0];
+        const auto &argv1 = args[1];
+
+        // Unlike std::equal, absl::c_equal will test for equal sizes.
+        auto argsEqual = [&ctx](const auto &arg0, const auto &arg1) {
+            if (arg0.name != arg1.name) {
+                return false;
+            }
+            return core::Types::equiv(ctx, arg0.type, arg1.type);
```
jez, line 3718: 🤔 Do you mind adding some tests for overloads that have generics (both `T.type_parameter` and `type_member`) in them?

```ruby
class A
  extend T::Sig

  sig do
    type_parameters(:U)
      .params(x: T.type_parameter(:U))
      .void
  end
  sig do
    type_parameters(:V)
      .params(x: T.type_parameter(:V), blk: T.proc.void)
      .void
  end
  def example(x, &blk); end
end
```

jez: ^ as a minimum, but maybe you can think of other test cases that use generics in combination with `T.any` / `T.all` etc.

resolver/resolver.cc
froydnj: The error message right now is just "Overloaded functions cannot have keyword arguments", pointing at all relevant keyword arguments.  It is basically what you get now if you try to define overloaded methods with keyword arguments, but it's not entirely accurate in this case, because we do allow it, we just have a lot of restrictions on it.

At this point in the code, we know that the user is probably trying to define an overloaded method with keyword arguments; we could issue more specific messages for each case from here on out:

* Mismatched positional and/or keyword argument names;
* Mismatched positional and/or keyword argument types;
* Both or neither sigs specifying a type for any associated block argument.

resolver/resolver.cc
```diff
@@ -3299,8 +3299,33 @@ class ResolveSignaturesWalk {
...
+    struct ArgInformation {
+        std::vector<core::ArgInfo> posArgs;
+        std::vector<core::ArgInfo> kwArgs;
+        std::optional<core::ArgInfo> blkArg;
+    };
+
+    // This structure serves double duty: it holds information about a single sig and the
+    // results of running it through fillInInfoFromSig and it also holds information about
+    // merging the results of several such structures together from overloaded methods.
+    struct SigInformation {
```
jez, line 3311: I worry about having something that is as generically named as `SigInformation` as we already have `ParsedSig` and also something else called `ArgInfo`. It seems like this structure is meant to hold information specifically for dealing with overloads + block args and I wonder whether we could make that more apparent in the name?

jez (commented):
I'd love to take another look after the generics tests. I don't know whether `Types::equiv` on its own is something that will be robust enough on its own, as it runs into a lot of the same problems that I think I'm dealing with in #6086, where `Types::equiv` is also not enough.

jez (approved):

PR #6081, opened 2022-07-24
Ease pain of lazily defining instance variables

jez:
### Motivation

This is a common stumbling point for many people using Sorbet.

This desugaring makes the first thing that people try to do also the correct
thing, which makes it easier to learn Sorbet, and means we have to answer fewer
questions.

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1213,6 +1213,7 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
             },
             [&](parser::OrAsgn *orAsgn) {
                 auto recvIsIvarLhs = parser::isa_node<parser::IVarLhs>(orAsgn->left.get());
+                auto recvIsCvarLhs = parser::isa_node<parser::CVarLhs>(orAsgn->left.get());
                 auto recv = node2TreeImpl(dctx, std::move(orAsgn->left));
                 auto arg = node2TreeImpl(dctx, std::move(orAsgn->right));
                 if (auto s = cast_tree<Send>(recv)) {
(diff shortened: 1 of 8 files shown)


ast/desugar/Desugar.cc
```diff
@@ -1220,7 +1235,16 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
...
+                    ExpressionPtr elsep;
+                    ast::Send *tlet;
+                    if (recvIsIvarLhs && (tlet = asTLet(arg))) {
+                        auto val = std::move(tlet->getPosArg(0));
+                        tlet->getPosArg(0) = MK::cpRef(recv);
+                        auto decl = MK::Assign(loc, MK::cpRef(recv), std::move(arg));
+                        elsep = MK::InsSeq1(loc, std::move(decl), MK::Assign(loc, std::move(recv), std::move(val)));
+                    } else {
+                        elsep = MK::Assign(loc, std::move(recv), std::move(arg));
+                    }
```
froydnj, line 1247: Non-blocking: a comment showing the shape of the expression here would be nice.

Alternatively, cleaning up this code such that the left-hand side is not called `recv` (and maybe the right-hand side is not called `arg`?) might make it sufficiently clear without the diagram

froydnj (approved):
For avoidance of doubt: we're not doing this for class variables because of the different semantics for class variables:

```ruby
@@cvar ||= T.let(..., T.nilable(...))
```

The VM will throw errors when this first executes because `@@cvar` hasn't been previously defined: class variables do not spring into existence on first reference like instance variables.  Is that reasoning for not applying the same treatment to class variables correct?

jez:
> For avoidance of doubt: we're not doing this for class variables because of the different semantics for class variables:
> 
> ```ruby
> @@cvar ||= T.let(..., T.nilable(...))
> ```
> 
> The VM will throw errors when this first executes because `@@cvar` hasn't been previously defined: class variables do not spring into existence on first reference like instance variables. Is that reasoning for not applying the same treatment to class variables correct?

The reason why I didn't add support was because I hadn't thought about it, not because I had thought about it and concluded that they worked differently.

```ruby
class Parent
  @@class_var ||= 0

  def self.class_var
    @@class_var += 1
  end
end

class Child < Parent; end

p(Parent.class_var)
p(Parent.class_var)

p(Child.class_var)
```

it seems like this program runs to completion without error printing `1, 2, 3` so it's probable that we could support class var in this way as well.

ast/desugar/Desugar.cc
jez: I renamed `recv` to `lhs`.

jez:
I pushed some changes to handle cvars.

froydnj (approved):

Issue #6041, opened 2022-07-18
Use type member upper bound to show completion suggestions


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20def%20method_on_a%3B%20end%0Aend%0A%0Aclass%20BoxA%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20Elem%20%3D%20type_member%20%7B%7Bupper%3A%20A%7D%7D%0A%0A%20%20sig%20%7Bparams%28x%3A%20Elem%29.void%7D%0A%20%20def%20example%28x%29%0A%20%20%20%20x.method_on_%0A%20%20%20%20%23%20%20%20%20%20%20%20%20%20%20%20%5E%20completion%3A%20method_on_a%0A%20%20end%0Aend)

```ruby
# typed: true

class A
  def method_on_a; end
end

class BoxA
  extend T::Sig
  extend T::Generic

  Elem = type_member {{upper: A}}

  sig {params(x: Elem).void}
  def example(x)
    x.method_on_
    #           ^ completion: method_on_a
  end
end
```

#### Observed output

```
editor.rb:15: Method method_on_ does not exist on A component of BoxA::Elem https://srb.help/7003
    15 |    x.method_on_
              ^^^^^^^^^^
  Got BoxA::Elem originating from:
    editor.rb:14:
    14 |  def example(x)
                      ^
  Did you mean method_on_a? Use `-a` to autocorrect
    editor.rb:15: Replace with method_on_a
    15 |    x.method_on_
              ^^^^^^^^^^
    editor.rb:4: Defined here
     4 |  def method_on_a; end
          ^^^^^^^^^^^^^^^
Errors: 1
```

At the point where the `^ completion:` assertion is, there are currently no completion results.

#### Expected behavior

At the point where the `^ completion:` assertion is, there should be a completion result that includes `method_on_a`, because `Elem` is upper bounded by `A` and `A` has that method.

- - -

PR #6082, opened 2022-07-24
Use upper bound from type_member in method autocomplete suggestions

michaeltaras:
Autocomplete suggestions for methods from the upper bound type when the receiver is a `type_member` with an upper bound.

### Motivation
Fixes https://github.com/sorbet/sorbet/issues/6041

### Test plan
See included automated tests.


jez (approved):
🎉 thanks!

PR #6083, opened 2022-07-25
Extend ambiguous definitions check to include casgns

aadi-stripe:
Extends the check for "ambiguous definitions" in the resolver to also include casgns.

### Motivation

The ambiguous definitions [check](https://sorbet.org/docs/error-reference#5068) was added to prevent namespace conflict/ambiguity in the Stripe codebase at runtime. There was a gap in the enforcement -- in that it did not prevent casgns from being ambiguously namespaced. This change closes that gap.

### Test plan

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1280,8 +1280,8 @@ public:
         }
 
         // Check for ambiguous definitions.
-        if (checkAmbiguousDefinition(ctx)) {
-            const auto ambigDef = findAnyDefinitionAmbiguousWithCurrent(ctx);
+        if (checkAmbiguousDefinition(ctx, nesting_->scope, nesting_->parent)) {
+            const auto ambigDef = findAnyDefinitionAmbiguousWithCurrent(ctx, nesting_->scope, nesting_->parent);
             if (ambigDef.exists()) {
                 if (auto e = ctx.beginError(original.declLoc, core::errors::Resolver::AmbiguousDefinitionError)) {
                     auto name = klass.data(ctx)->name.show(ctx);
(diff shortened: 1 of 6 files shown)


aadi-stripe:

→ https://go/builds/bui_M7VtvouJGrGkj7
→ https://go/builds/bui_M7VtURuiB1FoDo

resolver/resolver.cc
```diff
@@ -1298,7 +1298,7 @@ class ResolveConstantsWalk {
         nesting_ = nesting_->parent;
     }
 
-    const bool checkAmbiguousDefinition(core::Context ctx) {
+    const bool checkAmbiguousDefinition(core::Context ctx, core::SymbolRef curSym, shared_ptr<Nesting> curNesting) {
```
froydnj, line 1301: Can we take the `shared_ptr` by reference here so we're not constantly doing atomic reference counts unnecessarily?

```suggestion
    const bool checkAmbiguousDefinition(core::Context ctx, core::SymbolRef curSym, const shared_ptr<Nesting> &curNesting) {
```

resolver/resolver.cc
```diff
@@ -1309,24 +1309,23 @@ class ResolveConstantsWalk {
...
+        if (curOwner == curNesting->scope) {
             return false;
         }
 
         return true;
     }
 
-    const core::SymbolRef findAnyDefinitionAmbiguousWithCurrent(core::Context ctx) {
+    const core::SymbolRef findAnyDefinitionAmbiguousWithCurrent(core::Context ctx, core::SymbolRef curSym,
+                                                                shared_ptr<Nesting> curNesting) {
```
froydnj, line 1327: Same comment here, though I think taking by `const &` means that we need to rewrite the loop in this function a little further down...which is probably not a bad idea anyway.

```suggestion
    const core::SymbolRef findAnyDefinitionAmbiguousWithCurrent(core::Context ctx, core::SymbolRef curSym,
                                                                const shared_ptr<Nesting> &curNesting) {
```

froydnj (approved):
Up to you whether you want to fix these here or in a followup; since this code happens during a parallel phase, the efficiency matters slightly less, but even in parallel code serial efficiency still matters.

aadi-stripe:
@froydnj fixed

froydnj (approved):

PR #6085, opened 2022-07-25
Fix crash and add support for `T.type_parameter` in proc in body

jez:
### Motivation

As a consequence of #6034, I accidentally called `.value()` on an `optional`
somewhere that I shouldn't have. This meant that any code trying to exercise
that branch would raise a runtime exception when the `.value()` call happened,
instead of reporting a user-visible error.

This change fixes that.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/type_syntax/type_syntax.cc
+++ b/resolver/type_syntax/type_syntax.cc
@@ -23,8 +23,8 @@ optional<TypeSyntax::ResultType> getResultTypeAndBindWithSelfTypeParams(core::Co
                                                                         const ParsedSig &sigBeingParsed,
                                                                         TypeSyntaxArgs args);
 
-ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend, const ParsedSig *parent,
-                                     TypeSyntaxArgs args);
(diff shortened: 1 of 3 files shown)


resolver/type_syntax/type_syntax.cc
```diff
@@ -303,7 +308,19 @@ ParsedSig parseSigWithSelfTypeParams(core::Context ctx, const ast::Send &sigSend
                 }
 
                 bool validBind = false;
-                auto bind = move(getResultTypeWithSelfTypeParams(ctx, send->getPosArg(0), *parent, args).value());
+                auto maybeBind = getResultTypeWithSelfTypeParams(ctx, send->getPosArg(0), *parent, args);
+                core::TypePtr bind;
+                if (!maybeBind.has_value()) {
+                    if (auto e =
+                            ctx.beginError(send->getPosArg(0).loc(), core::errors::Resolver::InvalidMethodSignature)) {
+                        // TODO(jez) Test this branch
```
froydnj, line 316: Delete this, since it looks like you have tests for this?

froydnj (approved):

PR #6087, opened 2022-07-25
Remove T.untyped from RBI for Enumerable#reduce

### LINKED ISSUE #2770: `Enumerable#reduce` loses type information inside of block
#### Input

[→ View on sorbet.run](https://sorbet.run/#my_regexes%20%3D%20%5B%2Fa%2F%2C%20%2Fb%2F%5D%0Amy_strings%20%3D%20%5B'a'%2C%20'b'%5D%0Amy_str%20%3D%20'abc'%0A%0Amy_regexes.reduce(my_str)%20do%20%7Cprevious%2C%20to_remove%7C%0A%20%20%20%20T.assert_type!(previous%2C%20String)%0A%20%20%20%20previous.wrong_gsub(to_remove%2C%20'')%0Aend%0A%0Amy_strings.reduce(my_str)%20do%20%7Cprevious%2C%20to_remove%7C%0A%20%20%20%20T.assert_type!(previous%2C%20String)%0A%20%20%20%20previous.gsub(to_remove%2C%20'')%0Aend%0A)

```ruby
my_regexes = [/a/, /b/]
my_strings = ['a', 'b']
my_str = 'abc'

my_regexes.reduce(my_str) do |previous, to_remove|
    T.assert_type!(previous, String)
    previous.wrong_gsub(to_remove, '')
end

my_strings.reduce(my_str) do |previous, to_remove|
    T.assert_type!(previous, String) # this works just fine! Note that `initial` matches `Elem`
    previous.gsub(to_remove, '')
end
```

#### Observed output

```
editor.rb:6: The typechecker was unable to infer the type of the asserted value https://srb.help/7007
     6 |    T.assert_type!(previous, String)
    editor.rb:5:
     5 |my_regexes.reduce(my_str) do |previous, to_remove|
                                      ^^^^^^^^
  You may need to add additional `sig` annotations
Errors: 1
```

#### Expected behavior
No errors

- - -

I suspect the problem is around this signature: https://github.com/sorbet/sorbet/blob/fdb18367a187f0e33772e1ebabb4df6693f967bc/rbi/core/enumerable.rbi#L1232-L1245

If I'm not mistaken, `initial` shouldn't be restricted to `Elem`, but it could actually be of any type.

I tried changing the signature to what I think is the correct type definition:

```
sig do
    type_parameters(:U)
    .params(
        initial: T.type_parameter(:U),
        blk: T.proc.params(arg0: T.type_parameter(:U), arg1: Elem).returns(T.type_parameter(:U)),
    )
end
```

However, this still doesn't solve the problem, as when trying something like:

```
my_regexes.reduce(my_str) do |previous, to_remove|
    T.assert_type!(previous, String)
    previous.wrong_gsub(to_remove, '')
end
```

I get:

```
~/sorbet/test/testdata/rbi/enumerable.rb:33: Argument does not have asserted type String https://srb.help/7007
    33 |  T.assert_type!(previous, String)
  Got <any> originating from:
    ~/sorbet/test/testdata/rbi/enumerable.rb:32:
    32 |['/a/', ' /b/'].reduce('abc') do |previous, to_remove|
                                          ^^^^^^^^
Errors: 1
```

jez:
### Motivation

Fixes #2770

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/enumerable.rbi
+++ b/rbi/core/enumerable.rbi
@@ -849,17 +849,17 @@ module Enumerable
     .returns(T.untyped)
   end
   sig do
-    params(
-        initial: T.untyped,
-        blk: T.proc.params(arg0: T.untyped, arg1: Elem).returns(T.untyped),
+    type_parameters(:U).params(
+        initial: T.type_parameter(:U),
+        blk: T.proc.params(arg0: T.type_parameter(:U), arg1: Elem).returns(T.type_parameter(:U)),
     )
-    .returns(T.untyped)
+    .returns(T.type_parameter(:U))
   end
   sig do
-    params(
-        blk: T.proc.params(arg0: T.untyped, arg1: Elem).returns(T.untyped),
+    type_parameters(:U).params(
+        blk: T.proc.params(arg0: T.any(Elem, T.type_parameter(:U)), arg1: Elem).returns(T.type_parameter(:U)),
     )
-    .returns(T.untyped)
+    .returns(T.type_parameter(:U))
   end
   def inject(initial=T.unsafe(nil), arg0=T.unsafe(nil), &blk); end
 
@@ -1882,17 +1882,17 @@ module Enumerable
     .returns(T.untyped)
   end
   sig do
-    params(
-        initial: T.untyped,
-        blk: T.proc.params(arg0: T.untyped, arg1: Elem).returns(T.untyped),
+    type_parameters(:U).params(
+        initial: T.type_parameter(:U),
+        blk: T.proc.params(arg0: T.type_parameter(:U), arg1: Elem).returns(T.type_parameter(:U)),
     )
-    .returns(T.untyped)
+    .returns(T.type_parameter(:U))
   end
   sig do
-    params(
-        blk: T.proc.params(arg0: T.untyped, arg1: Elem).returns(T.untyped),
+    type_parameters(:U).params(
+        blk: T.proc.params(arg0: T.any(Elem, T.type_parameter(:U)), arg1: Elem).returns(T.type_parameter(:U)),
     )
-    .returns(T.untyped)
+    .returns(T.type_parameter(:U))
   end
   def reduce(initial=T.unsafe(nil), arg0=T.unsafe(nil), &blk); end
 
--- a/rbi/core/enumerable.rbi
+++ b/rbi/core/enumerable.rbi
@@ -1892,7 +1892,7 @@ module Enumerable
     type_parameters(:U).params(
         blk: T.proc.params(arg0: T.any(Elem, T.type_parameter(:U)), arg1: Elem).returns(T.type_parameter(:U)),
     )
-    .returns(T.type_parameter(:U))
+    .returns(T.nilable(T.type_parameter(:U)))
   end
   def reduce(initial=T.unsafe(nil), arg0=T.unsafe(nil), &blk); end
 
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_M7pUks0QtARmz7
→ https://go/builds/bui_M7pULWKUvrEny6

jez:
This causes about 500 errors on Stripe's codebase.

Most of the problems are due to code that looks like this the failing test (`test/testdata/infer/takes_no_block_inject.rb`).

I'm not sure if we can adjust the RBI type signature to make the inferred type better.
We approximate the block arg types immediately after checking args and before we check the block result type, so it seems like we might not be able to do anything other than require that the user write `T.let`:

```ruby
xs.inject(T.let(nil, T.nilable(Integer))) {|sum, x| sum ? sum + x : x}
```

jez:

→ https://go/builds/bui_M7qcX5CzBdliBq
→ https://go/builds/bui_M7qcz7idJnjqci

jez:

→ https://go/builds/bui_M7rcQs8fmQN26Q
→ https://go/builds/bui_M7rcJzw7F2C59k

jez:
I changed this PR to only do `Enumerable#reduce`, leaving `Enumerable#inject` for a future change, which dropped the number of errors on Stripe's codebase down to ~100, most of which were autocorrectable.

rbi/core/enumerable.rbi
```diff
@@ -1882,17 +1882,17 @@ module Enumerable
...
   sig do
-    params(
-        initial: T.untyped,
-        blk: T.proc.params(arg0: T.untyped, arg1: Elem).returns(T.untyped),
+    type_parameters(:U).params(
+        initial: T.type_parameter(:U),
+        blk: T.proc.params(arg0: T.type_parameter(:U), arg1: Elem).returns(T.type_parameter(:U)),
     )
-    .returns(T.untyped)
+    .returns(T.type_parameter(:U))
```
froydnj, line 1889: I get that this is correct as far as Sorbet goes, but I do wonder how many people will be surprised that the initial value needs to be specified as a union of the block return value and the element type, rather than Sorbet figuring that out on its own.

froydnj (approved):

rbi/core/enumerable.rbi
jez: Yeah, it's unclear. I guess we could ask about it?

rbi/core/enumerable.rbi
```diff
@@ -1882,17 +1882,17 @@ module Enumerable
...
   end
   sig do
-    params(
-        blk: T.proc.params(arg0: T.untyped, arg1: Elem).returns(T.untyped),
+    type_parameters(:U).params(
+        blk: T.proc.params(arg0: T.any(Elem, T.type_parameter(:U)), arg1: Elem).returns(T.type_parameter(:U)),
     )
-    .returns(T.untyped)
+    .returns(T.type_parameter(:U))
```
froydnj, line 1895: Technically this needs to be `T.nilable(T.type_parameter(:U))`, since `nil` is returned if the enumerable doesn't have any elements.  Not sure if that causes problems somewhere down the road (e.g. if `:U` is found to be `T.untyped`).

jez: Good catch, thanks.

jez:

→ https://go/builds/bui_M8WiLsbd1YU5M4
→ https://go/builds/bui_M8WiDkVvW2onmT

Issue #4638, opened 2021-09-24
Duplicate of textDocument/documentSymbol


ippachi:
I'm sorry about invalid format
This is IDE's issue.
I use sorbet LSP by neovim builtin LSP.
And I used workspaceSymbols by `:lua vim.lsp.buf.document_symbol()`, than result has a duplicate.
I don't know how to reproduce this phenomenon. It happens in my business project, but It doesn't happen in the tiny project for reproducing this issue.

## Source code
```ruby
# typed: strict

module MyProjectName # It is used so many places
  module Domain # It is used so many places
    module Model # It is used so many places
    end
  end
end
```

## Result of textDocument/documentSymbol (from nvim lsp log)
```lua
-- this is lua table
{
  id = 2,
  jsonrpc = "2.0",
  requestMethod = "textDocument/documentSymbol",
  result = { {
      children = { {
          children = { {
              children = {},
              detail = "",
              kind = 2,
              name = "Model",
              range = {
                end = {
                  character = 16,
                  line = 5
                },
                start = {
                  character = 4,
                  line = 5
                }
              },
              selectionRange = {
                end = {
                  character = 16,
                  line = 5
                },
                start = {
                  character = 4,
                  line = 5
                }
              }
            } },
          detail = "",
          kind = 2,
          name = "Domain",
          range = {
            end = {
              character = 15,
              line = 4
            },
            start = {
              character = 2,
              line = 4
            }
          },
          selectionRange = {
            end = {
              character = 15,
              line = 4
            },
            start = {
              character = 2,
              line = 4
            }
          }
        } },
      detail = "",
      kind = 2,
      name = "MyProjectName",
      range = {
        end = {
          character = 13,
          line = 3
        },
        start = {
          character = 0,
          line = 3
        }
      },
      selectionRange = {
        end = {
          character = 13,
          line = 3
        },
        start = {
          character = 0,
          line = 3
        }
      }
    }, {
      children = { {
          children = {},
          detail = "",
          kind = 2,
          name = "Model",
          range = {
            end = {
              character = 16,
              line = 5
            },
            start = {
              character = 4,
              line = 5
            }
          },
          selectionRange = {
            end = {
              character = 16,
              line = 5
            },
            start = {
              character = 4,
              line = 5
            }
          }
        } },
      detail = "",
      kind = 2,
      name = "Domain",
      range = {
        end = {
          character = 15,
          line = 4
        },
        start = {
          character = 2,
          line = 4
        }
      },
[...]

PR #6094, opened 2022-07-27
avoid duplication in LSP document symbols

froydnj:
### Motivation

This has come up both as a known bug in Sorbet's document symbol implementation and as a bug pointed out by people experimenting with document symbols.  The problematic case is laid out in comments in the commits for this PR.

### Test plan

I don't know that our test harness is equipped to handle this case: we want multiple files, but we only want to test the output of some LSP operation on a single file.

I plan to test this on Stripe's codebase using files that people have pointed out as having this problem.

I am not 100% sure that the refactoring that retains a list of candidates is behavior-preserving, since our test harness doesn't test multi-file cases at the moment.

### DIFF
--- a/main/lsp/requests/document_symbol.cc
+++ b/main/lsp/requests/document_symbol.cc
@@ -127,6 +127,37 @@ unique_ptr<ResponseMessage> DocumentSymbolTask::runRequest(LSPTypecheckerInterfa
             }
         }
     }
+
+    // Despite our best efforts above, we might still output duplicates from the
+    // list of candidate symbols.  For instance, given an ownership chain:
+    //
+    // A -> B -> C
+    //
+    // Consider the case where B was defined in a different file and C was defined
+    // in the current file: then C will be in our candidate list.
+    //
+    // But if B has a loc in the current file and A was defined in a different file,
+    // then B will also be in our candidate list!
+    //
+    // To avoid that case, we need to walk the ownership chains of each symbol to
+    // deduplicate the candidate list.
+    UnorderedSet<core::SymbolRef> deduplicatedCandidates(candidates.begin(), candidates.end());
+    for (auto ref : candidates) {
+        auto owner = ref.owner(gs);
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_M8S0GPuF3yKoZd
→ https://go/builds/bui_M8S0fym8tY6LUQ

main/lsp/requests/document_symbol.cc
```diff
@@ -102,29 +102,68 @@ unique_ptr<ResponseMessage> DocumentSymbolTask::runRequest(LSPTypecheckerInterfa
...
+    for (auto ref : candidates) {
+        auto owner = ref.owner(gs);
+        while (owner != core::Symbols::root()) {
+            if (deduplicatedCandidates.contains(owner)) {
+                deduplicatedCandidates.erase(ref);
+                break;
+            }
+
+            owner = owner.owner(gs);
+        }
```
froydnj, line 154: It's not clear to me whether I need to be super-defensive and check for `.exists()` or `hideSymbol(owner)` on the way up the chain.

froydnj:
This does seem to produce the desired deduplication on files in Stripe's codebase.

jez (commented):
I think my only question is how can we be sure in the long run that future changes do not, for example, introduce crashes in document symbols. The document symbols requests are sent by default without people realizing that they are using the feature, so if something crashes, it breaks it for everyone. It’s maybe fine if we don’t ship with a huge test suite for document symbols the moment we first turn it on, but I would at least like to know that the first time I need to minimize a crash and/or add a regression test that the testing harness is sufficient.

In my opinion that is the biggest reason why document symbols remains experimental, because if there are crashes it only affects people who have opted in, and there is an easy way to opt out

jez (approved):

jez:
@froydnj does this PR close this issue? https://github.com/sorbet/sorbet/issues/4638

PR #6100, opened 2022-07-28
Properly separate Find#find into two signatures

lugray:
This is a follow-up to #3867 given #6075. Thanks @froydnj and @elliottt!

### DIFF
--- a/rbi/stdlib/find.rbi
+++ b/rbi/stdlib/find.rbi
@@ -33,8 +33,14 @@ module Find
     params(
       paths: String,
       ignore_error: T::Boolean,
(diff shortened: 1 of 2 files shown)


froydnj (changes requested):
It looks like you can't use bare `Enumerator` here; using `T::Enumerator[String]` instead should work.

Tests in `test/testdata/rbi/` are optional but encouraged.

lugray:
Added a test and fixed the problem it identified. Can you launch your stripe (and other) tests?

froydnj (approved):

PR #6107, opened 2022-07-28
simply `cfg::Send` constructor

froydnj:
### Motivation

I'm not sure why the constructor was written this way originally (maybe before the `VariableUseSite(LocalRef)` constructor existed?), but doing it this was is more idiomatic and should be epsilon smaller/faster.

### Test plan

See included automated tests.


jez (approved):

froydnj:

→ https://go/builds/bui_M8uYg1GnYF9kbN
→ https://go/builds/bui_M8uYfdpC8ytBBf

PR #6118, opened 2022-07-29
Support `xs.reduce(0, &:+)` by desugaring to call-with-splat

### LINKED ISSUE #6112: Poor support for Symbol#to_proc with args
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0A%5B1%2C2%2C3%5D.reduce%28%26%3A%2B%29)

```ruby
[1,2,3].reduce(&:+)
```

#### Observed output

```
editor.rb:3: Not enough arguments provided for method Integer#+. Expected: 1, got: 0 https://srb.help/7004
     3 |[1,2,3].reduce(&:+)
                        ^^
    https://github.com/sorbet/sorbet/tree/master/rbi/core/integer.rbi#L182: Integer#+ defined here
     182 |  def +(arg0); end
            ^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

This should continue to be valid code.

jez:
Screen Shot 2022-07-29 at 1 54 33 PM

Based on #6168 and #6169

### Motivation

Fixes #6112

### Test plan

- [x] Separate this into the thing that adds the new intrinsic, and the thing that changes
      the desugaring, so that we can fix any type errors they would introduce independently
      of each other.
  - #6168
- [x] Write some comments explaining any non-obvious parts
- [x] Add tests, including:
  - [x] for all the different ways that the off-by-one errors could be wrong
    - #6168
  - for a method `&:foo` where `foo` has required or optional keyword args
    - my guess is that it's currently broken for those kinds of things today,
      and that this change doesn't make those cases worse (while making other
      cases better) so it's probably fine to land even if these cases weren't
      supported
  - [x] for the result type when given a `Range` like `args.slice(1..)`
    - #6168
- [x] ~~Factor things out to support `args[idx, len]` in addition to
  `args.slice(idx, len)`.~~ deferred to future date

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -194,9 +194,9 @@ public:
     }
 
     static ExpressionPtr CallWithSplat(core::LocOffsets loc, ExpressionPtr recv, core::NameRef name,
-                                       ExpressionPtr args) {
-        return Send3(loc, Magic(loc), core::Names::callWithSplat(), loc, std::move(recv), MK::Symbol(loc, name),
-                     std::move(args));
+                                       core::LocOffsets funLoc, ExpressionPtr splat) {
+        return Send4(loc, Magic(loc), core::Names::callWithSplat(), loc, std::move(recv), MK::Symbol(loc, name),
+                     std::move(splat), MK::Nil(loc.copyWithZeroLength()));
     }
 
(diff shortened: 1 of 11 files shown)


jez:
Self-assigning to fix tests

jez:
ok it was just a bad rebase, fixed now.

froydnj (approved):

PR #6123, opened 2022-07-30
track more information for better ranges on document symbols

froydnj:
### Motivation

[Document symbols in the LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentSymbol) include two different ranges: one that covers the entire extent of the symbol, and one that covers what should be selected when the symbol is shown.  Sorbet tracks both of these, but only the latter is recorded in the symbol table.  To get the former, we need to walk the AST for the requested file.

### Test plan

See included automated tests.

* [x] The hypothesis here is that this change will enable VSCode to show you more precisely where you are in the file.  Verify that.

### DIFF
--- a/main/lsp/requests/document_symbol.cc
+++ b/main/lsp/requests/document_symbol.cc
@@ -95,7 +95,8 @@ public:
     core::FileRef fref;
     UnorderedMap<core::SymbolRef, core::Loc> &mapping;
 
-    DefinitionLocSaver(core::FileRef fref, UnorderedMap<core::SymbolRef, core::Loc> &mapping) : fref(fref), mapping(mapping) {}
+    DefinitionLocSaver(core::FileRef fref, UnorderedMap<core::SymbolRef, core::Loc> &mapping)
+        : fref(fref), mapping(mapping) {}
 
     void postTransformClassDef(core::Context ctx, ast::ExpressionPtr &expr) {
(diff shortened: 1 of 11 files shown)


froydnj:

→ https://go/builds/bui_MAJZyTGnRBeh2v
→ https://go/builds/bui_MAJZ50IZyCuk6l

froydnj:
This change enables "Follow cursor" in outline mode to work like one would expect.

main/lsp/requests/document_symbol.cc
```diff
@@ -63,17 +74,50 @@ std::unique_ptr<DocumentSymbol> symbolRef2DocumentSymbol(const core::GlobalState
...
     result->children = move(children);
     return result;
 }
 
+// Document symbols in the LSP spec include two different ranges: one that covers
+// the entire extent of the definition and one that covers just what should be
+// shown when the symbol is selected.  Sorbet's loc for a symbol covers the latter.
+// We do store the former, but that loc lives in the AST, not in the symbol table,
+// so we build a separate temporary mapping with this tree walker.
+class DefLocSaver {
```
jez, line 93: slight preference for a different name, because there is already `main/lsp/DefLocSaver.h` which has a separate purpose.

jez (approved):

PR #6124, opened 2022-07-30
Tweak intersection types docs

qaisjp:
It currently ~says:

> the union of the set of values that have X or Y

To me, this reads like:

> the union of (set of values that have X or Y)

I believe the intention is:

> the union of (set of values that have X) or (set of values that have Y)

I tried to wordsmith this a bit more to make it a bit clearer to me, hope it helps?

### DIFF
--- a/website/docs/intersection-types.md
+++ b/website/docs/intersection-types.md
@@ -28,10 +28,10 @@ for intersection types:
 As the names indicate, union types and intersection types can be understood by
 analogy to the related operations on sets. Given `Type1` and `Type2`:
 
-- `T.any(Type1, Type2)` is the set of values that either have type `Type1` or
-  have type `Type2`.
-- `T.all(Type1, Type2)` is the set of values that both have type `Type1` and
-  have type `Type2`.
+- `T.any(Type1, Type2)` is the union of the set of values in both types:
+  values that are either of `Type1`, or of type `Type2`.
+- `T.all(Type1, Type2)` is the intersection of the set of values in both types:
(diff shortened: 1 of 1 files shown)


website/docs/intersection-types.md
```diff
@@ -28,10 +28,10 @@ for intersection types:
 As the names indicate, union types and intersection types can be understood by
 analogy to the related operations on sets. Given `Type1` and `Type2`:
 
-- `T.any(Type1, Type2)` is the union of the set of values that either have type
-  `Type1` or have type `Type2`.
```
qaisjp, line 32: I also felt that ``have type `Type2` `` doesn't flow well when spoken out loud — I swapped "have type TypeX" with "are of TypeX" to try to help with that

website/docs/intersection-types.md
```diff
@@ -28,10 +28,10 @@ for intersection types:
...
-- `T.all(Type1, Type2)` is the intersection of the set of values that both have
-  type`Type1` and have type `Type2`.
+- `T.any(Type1, Type2)` is the union of the set of values in both types: values
+  that are either of `Type1`, or of `Type2`.
+- `T.all(Type1, Type2)` is the intersection of the set of values in both types:
+  values that of `Type1` **and** of `Type2`.
```
froydnj, line 34:
```suggestion
  values that are of `Type1` **and** of `Type2`.
```

Or "are both", for parallelism with the above?

froydnj (approved):
+1

PR #6129, opened 2022-08-01
use a non-initializing allocator for (de)serialization buffers

froydnj:
### Motivation

Fewer dead writes to memory during deserialization, which should theoretically be faster.

(The same principle can be applied to the serialization buffers, but that is a more invasive change.  At the moment this is more of a gut check.)

### Test plan

See included automated tests.

### DIFF
--- a/core/serialize/pickler.h
+++ b/core/serialize/pickler.h
@@ -17,10 +17,30 @@ public:
     Pickler() = default;
 };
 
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_MAKEsKS3KmIy6P
→ https://go/builds/bui_MAKER43nTmgSur

PR #6139, opened 2022-08-02
Better error when the LHS of an `&&` is always true

### LINKED ISSUE #6138: Better error when the LHS of an `&&` is always true
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Breturns%28TrueClass%29%7D%0Adef%20always_true%3B%20true%3B%20end%0A%0Aif%20always_true%20%26%26%20T.unsafe%28true%29%0A%20%20puts%20'here'%0Aend)

```ruby
# typed: true
extend T::Sig

sig {returns(TrueClass)}
def always_true; true; end

if always_true && T.unsafe(true)
  puts 'here'
end
```

#### Observed output

```
editor.rb:7: This code is unreachable https://srb.help/7006
     7 |if always_true && T.unsafe(true)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    editor.rb:7: This condition was always truthy (TrueClass)
     7 |if always_true && T.unsafe(true)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Got TrueClass originating from:
    editor.rb:7:
     7 |if always_true && T.unsafe(true)
           ^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

The "unreachable code" in this example is simply the `else` that's generated by desugar:

```ruby
  if begin
      &&$2 = <self>.always_true()
      if &&$2
        <emptyTree>::<C T>.unsafe(true)
      else
        &&$2
      end
    end
    <self>.puts("here")
  else
    <emptyTree>
  end
```

We should be able to detect this, and report a better error.

For example, either:

- don't report a dead code error, because the dead code is actually something generated by Sorbet, not written by the user
- report an error like "it looks like this half of the `&&` is always truthy, so you can just delete it

- - -

Depending on the approach we want to take, we should likely be able to handle this in a similar way to other desugar things, where we make the `else` branch change from looking simply like `else &&$2` to looking like `else <Magic>.<and-and-else>(&&$2)` or something.

We do something similar when we desugar `x&.foo` and for [this example](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28xs%3A%20T%3A%3AArray%5BInteger%5D%29.void%7D%0Adef%20example%28xs%29%0A%20%20if%20xs.first%20%26%26%20xs.first.even%3F%0A%20%20end%0Aend%0A) about assuming the result type doesn't change.

jez:
### Motivation

Fixes #6138

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/and-and-always-truthy/test.rb
+++ b/test/cli/and-and-always-truthy/test.rb
@@ -1,0 +1,19 @@
+# typed: true
+extend T::Sig
+
+sig {returns(TrueClass)}
+def always_true; true; end
+
+sig {returns(FalseClass)}
+def always_false; false; end
+
+def example1
+  if always_true && T.unsafe(true)
+    puts 'here'
+  end
+end
+
+def example2
+  if always_false && T.unsafe(true)
+  end
+end
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1098,6 +1098,7 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                 auto rhs = node2TreeImpl(dctx, std::move(and_->right));
                 if (isa_reference(lhs)) {
                     auto cond = MK::cpRef(lhs);
+                    // TODO(jez) How to handle this case?
                     auto iff = MK::If(loc, std::move(cond), std::move(rhs), std::move(lhs));
                     result = std::move(iff);
                 } else {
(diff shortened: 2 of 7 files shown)


test/cli/and-and-always-truthy/test.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+end
+
+sig {params(always_true: TrueClass).void}
+def example3(always_true)
+  # TODO: We don't get the same sort of nice error message in this case.
+  if always_true && T.unsafe(true)
+    puts 'here'
+  end
+end
```
jez, line 27: I'm fine with this limitation for the time being, because I think that something is better than nothing.

Even better might be to desugar this to have some sort of `<Magic>.<and-and-else>(always_true)` call in it:

```ruby
if always_true
  T.unsafe(true)
else
  <Magic>.<and-and-else>(always_true)
end
```

but then the first dead instruction in `inference.cc` would still be the `always_true` ident and the call would be the *second* instruction, and this would make trees larger, etc. etc.

This could happen in reasonable looking code, too:

image

but `¯\_(ツ)_/¯`

froydnj (approved):
I can't decide whether I admire or despair that we use magic variable names like this.

I think the CLI test expectations need to be updated.

jez:
> I can't decide whether I admire or despair that we use magic variable names like this.

The magic variable names are partly the problem—if we ever decide that we want to handle the `x && y` case where `x` is a variable, we'd need to change this to something more explicit like `<Magic>.<and-and-else>(x)` or something.

PR #6140, opened 2022-08-02
use a non-initializing allocator for stable string storage buffers

froydnj:
### Motivation

Same reason as #6129, but with a slightly large effect radius.

(Also may be worth exploring whether we should hold memory buffers directly, rather than indirecting through `vector` all the time.)

### Test plan

See included automated tests.

### DIFF
--- a/common/StableStringStorage.h
+++ b/common/StableStringStorage.h
@@ -7,7 +7,28 @@
 
 namespace sorbet {
 
+// Allocator adaptor that interposes construct() calls to
+// convert value initialization into default initialization.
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_MAgPzxayaQITEx
→ https://go/builds/bui_MAgPPyjY4jEsNK

froydnj:

→ https://go/builds/bui_MAhkox9QZOhHbc
→ https://go/builds/bui_MAhkUnI9Uf9wnO

PR #6143, opened 2022-08-02
LSP ErrorReporter: Do not report extraneous empty error lists to the client.

jvilk-stripe:
LSP ErrorReporter: Do not report extraneous empty error lists to the client.

Sorbet only needs to send empty error lists to the client when it has previously sent errors for a file that now has no errors.

Before this change, when the client has `lspErrorCap` errors, the ErrorReporter would report an empty error list for
any file with errors that exceed the cap -- even if Sorbet had never sent errors for that file before.
This can freeze VS Code (DoS due to many messages) if you make a catastrophic typo that causes many files to error.

This PR changes `clientErrorCount` to count the errors _reported_ to the client (as it should), and not the total number of errors in the last epoch.
Similarly, `FileErrorStatus.errorCount` now tracks the errors _reported_ to the client, and not the errors in the last epoch.

Now, we can avoid sending unneeded empty lists when `errorCount == 0` from the last epoch and `errorsToReport == 0`.

### Motivation

Sorbet was freezing VS Code at Stripe when it was run before some core code generation routines finished. The source of the freeze was Sorbet spamming empty error lists at VS Code for nearly every source file in our repository.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPOutput.cc
+++ b/main/lsp/LSPOutput.cc
@@ -69,6 +69,11 @@ vector<unique_ptr<LSPMessage>> LSPOutputToVector::getOutput() {
     return messages;
 }
 
+uint32_t LSPOutputToVector::size() {
+    absl::MutexLock lock(&mtx);
+    return output.size();
+}
+
 unique_ptr<LSPMessage> LSPOutputToVector::read(int timeoutMs) {
     absl::MutexLock lock(&mtx);
     mtx.AwaitWithTimeout(absl::Condition(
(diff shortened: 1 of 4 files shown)


main/lsp/LSPOutput.cc
```diff
@@ -69,6 +69,11 @@ vector<unique_ptr<LSPMessage>> LSPOutputToVector::getOutput() {
     return messages;
 }
 
+uint32_t LSPOutputToVector::size() {
+    absl::MutexLock lock(&mtx);
+    return output.size();
+}
```
jez, line 75: this is essentially a test-only change, right?

jez (approved):

PR #6144, opened 2022-08-02
move document symbols enablement into beta

froydnj:
### Motivation

People keep asking for this feature at Stripe, and we keep having them to enable experimental features to do it.  We've had a small number of people testing this for a while and our internal logs show that it's stable enough.

Ergo, let's roll it out to a wider audience by making it a beta feature (enabled by default on all Stripe developer machines).

Note that this is not changing the name of the option, which is somewhat unfortunate, but if this beta test shows that document symbols is as stable as we think it is, we can make document symbols stable in Sorbet by default, deprecate the option, and eventually remove it.

### Test plan

See included automated tests.


froydnj:
(Will wait to push this until we're confident the other features we have under beta aren't going to need an emergency reversion.)

jez (approved):

PR #6146, opened 2022-08-02
Skip already-added files in fast path decision

jez:
### Motivation

This _should_ make the list of files free of duplicates by construction.

It also makes it easier for us to have the `gs->getFiles()` loop short
circuit once it adds too many files (deferring to a future change
because that might come after we move this logic from LSPTypechecker.cc
to LSPIndexer.cc).

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -235,8 +235,7 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
     ENFORCE(updates.canTakeFastPath);
 
     Timer timeit(config->logger, "fast_path");
-    UnorderedSet<core::FileRef> changedFiles;
-    vector<core::FileRef> allFiles;
+    UnorderedSet<core::FileRef> toTypecheck;
     vector<core::ShortNameHash> changedSymbolNameHashes;
     UnorderedMap<core::FileRef, core::FoundMethodHashes> oldFoundMethodHashesForFiles;
     // Replace error queue with one that is owned by this thread.
@@ -306,7 +305,7 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
                     gs->replaceFile(fref, updatedFile);
                     // If file doesn't have a typed: sigil, then we need to ensure it's typechecked using typed: false.
                     fref.data(*gs).strictLevel = pipeline::decideStrictLevel(*gs, fref, config->opts);
-                    changedFiles.emplace(fref);
+                    toTypecheck.emplace(fref);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


main/lsp/LSPTypechecker.cc
```diff
@@ -349,9 +352,11 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
             }
         }
-        // Remove any duplicate files.
-        fast_sort(subset);
-        subset.resize(std::distance(subset.begin(), std::unique(subset.begin(), subset.end())));
+        DEBUG_ONLY(auto initialSize = subset.size();
+                   // Remove any duplicate files.
+                   fast_sort(subset);
+                   subset.resize(std::distance(subset.begin(), std::unique(subset.begin(), subset.end())));
+                   ENFORCE(subset.size() == initialSize, "subset is not free of duplicates by construction"););
```
froydnj, line 359: I think we should continue sorting in non-debug builds, too; I realize that makes the logic awkward for debug builds, but determinism in resolving and typechecking below seems like a good thing.

froydnj (approved):
I think this makes sense, but the comment below should be addressed before merging.

main/lsp/LSPTypechecker.cc
jez: My previous logic was wrong (using epochs wasn't as predictable as I had once thought).

I've rewritten this to use an `UnorderedSet`.

This potentially allocates more, but I think we can compensate for it with some future changes (e.g., keeping track of "edited files" and "extra files that need retypechecking" should enable other optimizations that is a net win—see the commit messages for more).

jez:
I misspoke in my claim that I thought we could easily avoid reindexing. The indexed files in the `updates` variable were indexed with `initialGS`, while we need to typecheck with the typechecking global state. That means that things like IDs for NameRefs will be different.

We could potentially avoid re-indexing by doing an ast::Substitute::run, but we don't have access to the `initialGS` from the typechecker thread (its owned by the indexer thread). So that sounds hairy to try to avoid doing that work.

In any case, I think that this refactoring to use the extra set will not be too much slower, and will unblock getting us to short circuit in the "can take fast path" decision.

froydnj (approved):
Making a set makes more sense to me and is less sensitive to epoch concerns.

Since we have to have a set and a vector anyway, WDYT about structuring things like:

```c++
UnorderedSet<core::FileRef> subset;

for (auto &updatedFile : updates.updatedFiles) {
  ...
  // Add the changed files that we know about
  subset.insert(fref);
}

// later
for (auto &oldFile : gs->getFiles()) {
  ...
  // Don't have to care whether this is a duplicate or not.
  // If we want to check .contains() earlier in the loop, we could do that.
  subset.insert(ref);
}

vector<core::FileRef> allFiles(subset.begin(), subset.end());
fast_sort(allFiles);

// ...
```

This feels cleaner to me because there's not two things that we need to think about merging; there's always only one source of truth in both loops, and we only reduce things to a vector at the end.

Issue #3794, opened 2020-12-15
Crash in error reporting for class << self


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Adef%20foo%0A%20%20(class%20%3C%3C%20self%3B%20self%3B%20end).class_eval%0Aend)

```ruby
# typed: true

def foo
  (class << self; self; end).class_eval
end
```

#### Observed output

```
❯ sorbet foo.rb
Backtrace:
  #3 0x1941776
  #4 0x194170f
  #5 0xb9110b std::__1::__throw_out_of_range()
  #6 0xb8e965 std::__1::basic_string_view<>::substr()
  #7 0x16dd691 sorbet::core::Loc::source()
  #8 0x173acac sorbet::core::dispatchCallSymbol()
  #9 0x17429d5 sorbet::core::ClassType::dispatchCall()
  #10 0x172666b sorbet::core::(anonymous namespace)::CALL_MEMBER_impl_dispatchCall<>::call<>()
  #11 0x17263c3 sorbet::core::TypePtr::dispatchCall()
  #12 0x11e22e6 sorbet::infer::Environment::processBinding()::$_11::operator()()
  #13 0x11e1b45 sorbet::typecaseHelper<>()
  #14 0x11dcfa9 sorbet::typecase<>()
  #15 0x11db48f sorbet::infer::Environment::processBinding()
  #16 0x11c995c sorbet::infer::Inference::run()
  #17 0xea305f sorbet::realmain::pipeline::CFGCollectorAndTyper::preTransformMethodDef()
  #18 0xea2e5b sorbet::ast::CALL_MEMBER_impl_preTransformMethodDef<>::call<>()
  #19 0xea0f4c sorbet::ast::TreeMapper<>::mapMethodDef()
  #20 0xe9fa45 sorbet::ast::TreeMapper<>::mapIt()
  #21 0xea0e5b sorbet::ast::TreeMapper<>::mapClassDef()
  #22 0xe9f9c0 sorbet::ast::TreeMapper<>::mapIt()
  #23 0xea2a09 sorbet::ast::TreeMapper<>::mapInsSeq()
  #24 0xea05f3 sorbet::ast::TreeMapper<>::mapIt()
  #25 0xe8dd19 sorbet::ast::TreeMap::apply<>()
  #26 0xe74113 sorbet::realmain::pipeline::typecheckOne()
[...]

PR #6148, opened 2022-08-02
Add `class_eval` test

adisonlampert:
### Motivation

Follow up to #3794, adds a test to ensure error `Method class_eval does not exist on NilClass` is raised instead of crashing.

### Test plan

See included automated tests.


jez (approved):
Thanks!

PR #6150, opened 2022-08-02
Don't retypecheck payload files on incremental fast path

jez:
### Motivation

Re-typechecking the payload files on the incremental fast path has the potential to do two things, both of which are bad:

- delete methods associated with well-known SymbolRef IDs
- delete methods which have had intrinsics registered for them

There are possibly multiple ways to solve this, but I have chosen the one which is "don't re-run incremental namer on payload files"

I think this should be fine, because it is essentially already what the slowpath will do (that is, the slow path is able to correctly type check *new* files without needing to be fed the list of payload files in `pipeline::resolve` or `pipeline::typecheck`).

### Alternatives considered

In the case that this approach *does* happen to bite us, some approaches which I tried and punted on, but which we may wish to reconsider:

- run incremental namer on payload files, but (1) never delete well-known (synthetic) symbols or symbols that have intrinsics associated with them
  - I tried this, but there was some subtlety around how overloads were handled which meant that it didn't work. luckily we had tests that failed when i went with this approach.
- run incremental namer on payload files. after it has deleted what it needs to delete (possibly including well-known symbols or symbols with intrinsics), re-run `GlobalState::initEmpty`
  - I did not try this, because I noticed that there were at least a few places in `initEmpty` where it was not idempotent (for example, `files.emplace_back()` to create the `core::File` with ID `0`. Also all the `synthesizeClass` do the same, but with `classAndModules.emplace_back()`). We could probably get `initEmpty` to be idempotent, but ENFORCE'ing that it remains idempotent going forward seems tricky (or maybe not? maybe it would suffice to just run `DEBUG_ONLY(initEmpty())` immediately after the first call, and see that it doesn't add anything or something).

In any case, I think that the approach above of skipping payload files should actually be fine.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -334,6 +334,12 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
                     continue; // See note above about --stripe-packages.
                 }
 
+                if (oldFile->isPayload()) {
+                    // Don't retypecheck files in the payload via incremental namer, as that might
+                    // cause well-known symbols to get deleted and assigned a new SymbolRef ID.
+                    continue;
+                }
+
                 ENFORCE(oldFile->getFileHash() != nullptr);
                 const auto &oldHash = *oldFile->getFileHash();
                 vector<core::ShortNameHash> intersection;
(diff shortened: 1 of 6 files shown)


jez:

→ https://go/builds/bui_MAsL69N7DTHFHu
→ https://go/builds/bui_MAsLceMTasIWde

froydnj (approved):

PR #6151, opened 2022-08-02
Allow `sorbet.*` VS Code commands from anywhere

jez:
cc @jeffjo-stripe

### Motivation

Previously, Sorbet required that you have a Ruby file actively opened,
for these commands to even appear in the command palette. As far as I
can tell, none of those commands depend on the currently open file to
function correctly.

### Test plan

I tested this change manually a few times to verify that I can still
configure and restart sorbet even when there is no Ruby file open.

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -255,20 +255,16 @@
     "menus": {
       "commandPalette": [
         {
-          "command": "sorbet.configure",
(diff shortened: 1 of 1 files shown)


jez:
@jvilk I'd also appreciate any thoughts you have on this change as well.

froydnj (approved):

PR #6158, opened 2022-08-04
force output of children on the ownership chain in document symbols

froydnj:
### Motivation

This came up via a Stripe tester.  Ideally the commit history + the associated comment explains what's going wrong.

### Test plan

See included automated tests.  I think the checks that we already have for children appearing in the relevant file will prevent an explosion of symbols here: e.g. for the `A::B::C::D::E` example given in the comments, we'll avoid outputting any children of `B`, `C`, or `D` that don't appear in the current file already.

* [ ] Double check that this fixes the issue on the relevant file

### DIFF
--- a/test/testdata/lsp/dropped_document_symbols__1.rb.document-symbols.exp
+++ b/test/testdata/lsp/dropped_document_symbols__1.rb.document-symbols.exp
@@ -1,0 +1,33 @@
+{
+    "jsonrpc": "2.0",
+    "id": 1,
+    "requestMethod": "textDocument/documentSymbol",
+    "result": [
+        {
+            "name": "Bottom",
+            "detail": "",
+            "kind": 2,
+            "range": {
+                "start": {
+                    "line": 2,
+                    "character": 0
+                },
+                "end": {
+                    "line": 7,
+                    "character": 3
+                }
+            },
+            "selectionRange": {
+                "start": {
+                    "line": 2,
+                    "character": 0
+                },
+                "end": {
+                    "line": 2,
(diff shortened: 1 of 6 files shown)


test/testdata/lsp/dropped_document_symbols__1.rb.document-symbols.exp
```diff
@@ -0,0 +1,167 @@
...
+                    "selectionRange": {
+                        "start": {
+                            "line": 3,
+                            "character": 8
+                        },
+                        "end": {
+                            "line": 3,
+                            "character": 15
+                        }
+                    },
```
froydnj, line 54: Sadly, I think this breaks cursor following in vscode, but I don't know what to do about that, since `Private` doesn't have a definition extent in the current testcase or (AFAICT) in the original report from Stripe.

I'm not sure if the right thing to do is hack things up so that for forced candidates, we invent definition extents for them in the current file, or just live with this limitation.

froydnj: The other alternative, which is less hacky in some ways, would be to somehow recognize a special case of two symbols, `Owner` and `Owned`, and just output `Owned`, rather than claiming that `Owner` is the symbol that we really want.

Ideally, we'd recognize for:

```
Owner -> Owned -> method1
Owner -> Owned -> method2
...more ownership chains starting with Owner -> Owned
```

the same sort of thing, but that might be a little more complicated.

jez (approved):

PR #6161, opened 2022-08-04
Fix Exception::raise EmptyElse

jez:
### Motivation

Reported via StackOverflow

<https://stackoverflow.com/questions/73091405/exceptionraise-unimplemented-parser-node-emptyelse>

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -2271,6 +2271,7 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                 auto res = desugarOnelinePattern(dctx, pattern->loc, pattern->rhs);
                 result = std::move(res);
             },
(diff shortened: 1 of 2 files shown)


froydnj (approved):
Pretty sure this needs an `update_testdata_exp.sh` run.

jez:
> Pretty sure this needs an `update_testdata_exp.sh` run.

The tests passed. Why do you think this needs something to update?

The behavior is that if `else` is `nullptr`, desugar will produce an `EmptyTree`. When the `else` is explicitly present but an `EmptyElse` node, it used to crash, and now it explicitly handles that case. So the behavior should be identical.

froydnj:
> > Pretty sure this needs an `update_testdata_exp.sh` run.
> 
> The tests passed. Am I missing something?

Nope!  They just hadn't completed when I reviewed and I noticed the presence of exp files.  My mistake.

PR #6162, opened 2022-08-04
Auto-cast `Class.new(Parent)` to `T.class_of(Parent)`

Morriar:
### Motivation

Change the `ClassNew` rewriter to auto-cast the result of `Class.new(Parent)` into a `T.class_of(Parent)`.

Right now, the associated type to `Class.new(Parent)` is always `Class`:

```rb
x = Class.new(Parent)
T.reveal_type(x) # => Class
```

This is problematic when trying to call things defined in `Parent`:

```rb
class Parent
  def foo; end
end

Class.new(Parent).new.foo # error: Method foo does not exist on Object
```

To be able to call `foo`, one has to manually cast the result:

```rb
T.cast(Class.new(Parent), T.class_of(Parent)).new.foo # ok
```

We can be a bit smarter and always rewrite `Class.new(<ConstantLiteral>)` to use a cast so this works out of the box:

```rb
x = Class.new(Parent)
T.reveal_type(x) # => T.class_of(Parent)
x.new.foo # ok
```

Since it's rewritten as this:

```rb
x = T.cast(Class.new(Parent), T.class_of(Parent))
T.reveal_type(x) # => T.class_of(Parent)
x.new.foo # ok
```

### Test plan

See included automated tests.

cc. @egiurleo

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -373,6 +373,10 @@ public:
         return Send2(loc, T(loc), core::Names::bind(), loc, std::move(value), std::move(type));
     }
 
+    static ExpressionPtr Cast(core::LocOffsets loc, ExpressionPtr value, ExpressionPtr type) {
+        return Send2(loc, T(loc), core::Names::cast(), loc, std::move(value), std::move(type));
+    }
+
     static ExpressionPtr ClassOf(core::LocOffsets loc, ExpressionPtr value) {
         return Send1(loc, T(loc), core::Names::classOf(), loc, std::move(value));
     }
--- a/rewriter/ClassNew.cc
+++ b/rewriter/ClassNew.cc
@@ -24,6 +24,20 @@ bool isRewrittenBind(ast::ExpressionPtr &expr) {
     return send->flags.isRewriterSynthesized;
 }
 
+// Is this expression a synthetic `T.cast` call we added from the `ClassNew` rewriter on sends?
+bool isRewrittenCast(ast::ExpressionPtr &expr) {
+    auto send = ast::cast_tree<ast::Send>(expr);
+    if (send == nullptr) {
+        return false;
+    }
+
+    if (send->fun != core::Names::cast()) {
+        return false;
+    }
+
+    return send->flags.isRewriterSynthesized;
+}
+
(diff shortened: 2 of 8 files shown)


jez:
I have generally tried avoiding rewriter passes that generate `cast` instructions, as there are restrictions on where `T.cast` is allowed to appear, and it is hard to detect whether we are currently in a position where `T.cast` is allowed.

```ruby
# typed: true

class Parent; end

class A
  def initialize
    @x = Class.new(Parent) {}
  end
end
```

```ruby
# typed: true

class Parent; end

class A
  def initialize
    if T.unsafe(nil)
      @x = Class.new(Parent) {}
    else
      @x = nil
    end
  end
end
```

```ruby
# typed: strict

class Parent; end

class Class
  T::Sig::WithoutRuntime.sig do
    type_parameters(:Klass)
      .params(klass: T.all(Class, T.type_parameter(:Klass)))
      .returns(T.all(Class, T.type_parameter(:Klass)))
  end
  def self.new(klass)
    raise
  end
end

class A
  extend T::Sig

  sig {void}
  def initialize
    x = Class.new(Parent) {}
  end
end
```

Also: I notice that some of your location information is not great. For example, if you add a `^ hover` assertion in any of those tests, you'll notice that instead of showing

```ruby
sig {params(superclass: Class).returns(Class)}
def initialize (overload.1)(superclass=…); end
```

it shows something like

````ruby
```
sig do
  params(
    klass: T.untyped
  )
  .returns(Runtime object representing type: T.class_of(Parent)
end
def self.class_of(klass); end
```

---

Type syntax to declare the type of a class's singleton class.
````

which shows that it found the synthesized call to `T.class_of` on hover instead of the call to `new`. If you set the synthetic loc to something with zero length (see the `copyWithZeroLength` helper) it will instruct LSP to ignore the thing on hover.

And as one last point: maybe we should simply fix this not by using `T.cast` + desugar, but by adding an RBI definition for `Class.new` similar to the sig shown above?

jez (commented):
^

Morriar:
Closing as this as been fixed more elegantly in #6589.

PR #6165, opened 2022-08-04
Factor out `getFilesToTypecheck` helper

jez:
### Motivation

This is a preliminary step to moving some logic from `LSPTypechecker.cc` to
`LSPIndexer.cc`, which will let us stop showing `Typechecking in foreground...`
for so many edits (especially when the fast path methods beta is enabled).

### Test plan

This should be a no-op change.

I will also be doing some light interactive testing on Stripe's codebase.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -221,20 +221,9 @@ bool LSPTypechecker::typecheck(LSPFileUpdates updates, WorkerPool &workers,
     return committed;
 }
 
-vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, WorkerPool &workers,
-                                                  shared_ptr<core::ErrorFlusher> errorFlusher) const {
-    ENFORCE(this_thread::get_id() == typecheckerThreadId, "Typechecker can only be used from the typechecker thread.");
-    ENFORCE(this->initialized);
-    // We assume gs is a copy of initialGS, which has had the inferencer & resolver run.
-    ENFORCE(gs->lspTypecheckCount > 0,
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_MBaNPtM87PrNws
→ https://go/builds/bui_MBaN6uhqQKos7v

main/lsp/LSPTypechecker.cc
```diff
@@ -221,6 +221,126 @@ bool LSPTypechecker::typecheck(LSPFileUpdates updates, WorkerPool &workers,
...
+    absl::c_transform(changedMethodSymbolHashes, std::back_inserter(result.changedSymbolNameHashes),
+                      [](const auto &symhash) { return symhash.nameHash; });
+    absl::c_transform(changedFieldSymbolHashes, std::back_inserter(result.changedSymbolNameHashes),
+                      [](const auto &symhash) { return symhash.nameHash; });
+    core::ShortNameHash::sortAndDedupe(result.changedSymbolNameHashes);
+
+    if (!result.changedSymbolNameHashes.empty()) {
+        // ^ optimization--skip the loop over every file in the project (`gs.getFiles()`) if
+        // the set of changed symbols is empty (e.g., running a completion request inside a
+        // method body)
```
froydnj, line 303: WDYT about dedenting this now that the logic is in its own function?

```c++
if (result.changedSymbolNameHashes.empty()) {
  return result;
}

// ...logic to further modify `result`
return result;
```

froydnj (approved):

main/lsp/LSPTypechecker.cc
jez: I ended up doing this in the stacked PR that follows.

PR #6166, opened 2022-08-05
Move "extra files to typecheck" logic into index

jez:
### Motivation

This should basically eliminate the "Typechecking in foreground..." problem by
moving such changes to be handled as slow path changes.

Eventually, we might want to build some sort of "medium fast path" mode which
can support large incremental updates, but that can be some other time, as this
is better than the current state of things where Sorbet is completely frozen and
can't even be cancelled or pre-empted.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_MBcW06Ga5llgdM
→ https://go/builds/bui_MBcWFbAzwkT6oM

main/lsp/LSPFileUpdates.h
```diff
@@ -47,6 +49,35 @@ class LSPFileUpdates final {
...
+        // The names of all symbols changed by this set of updates
+        std::vector<core::ShortNameHash> changedSymbolNameHashes;
+
+        // Extra files that need to be typechecked because the file mentions the name of one of the changed symbols.
+        std::vector<core::FileRef> extraFiles;
+    };
+
+    // TODO(jez) It would be nice to have this accept `...<const core::File>...` but I can't figure
+    // out how to get the callers, which do not have `const core::File`'s. I don't know how to have
+    // C++ let me do upcasts on constness inside templates like that.
```
froydnj, line 66: You can't do this in general without copying, because `Klass<T>` and `Klass<const T>` are different types.  (`Klass<T>` and `Klass<const T>` could also have completely different representations under the hood thanks to template specialization.)

froydnj (approved):

main/lsp/LSPFileUpdates.h
jez: :sadpanda:

PR #6169, opened 2022-08-05
Don't accidentally silence errors discovered from call-with-splat

jez:
### Motivation

We shouldn't just drop errors on the floor.

Also will help land #6118

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -783,8 +783,7 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                     }
 
                     auto kwargs = node2TreeImpl(dctx, std::move(kwArray));
-                    auto method = MK::Literal(
-                        loc, core::make_type<core::NamedLiteralType>(core::Symbols::Symbol(), send->method));
+                    auto method = MK::Symbol(send->methodLoc, send->method);
 
                     if (auto *array = cast_tree<Array>(kwargs)) {
                         DuplicateHashKeyCheck::checkSendArgs(dctx, 0, array->elems);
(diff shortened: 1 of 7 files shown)


jez:

→ https://go/builds/bui_MBhGxvpR2RyEpG
→ https://go/builds/bui_MBhGtBaqyWP9vI

ast/desugar/Desugar.cc
```diff
@@ -797,7 +796,7 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
                     sendargs.emplace_back(std::move(kwargs));
                     ExpressionPtr res;
                     if (block == nullptr && !anonymousBlockPass) {
-                        res = MK::Send(loc, MK::Magic(loc), core::Names::callWithSplat(), locZeroLen, 4,
+                        res = MK::Send(loc, MK::Magic(loc), core::Names::callWithSplat(), send->methodLoc, 4,
```
froydnj, line 799: Is this change really innocuous?  I guess the rationale is that the `Send` we're creating here ought to be analogous to the `Send` that textually appears, even if the new `Send` is using some Sorbet-internal thing to do the work.

froydnj (approved):

ast/desugar/Desugar.cc
jez: This is what enables the correct `^^^` lines below.

Before, it would see that the funLoc was zero and decided to report the "method does not exist" error on the `callLoc`, which is what the first argument to the `MK::Send` is, which uses `loc`, indicating that it's the complete parser::Send node's range.

So yes this change is pretty innocuous in my mind, and in fact more correct than it was previously.

PR #6171, opened 2022-08-05
Fix type of kwargs in Open3 RBI

jez:
### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/open3.rbi
+++ b/rbi/stdlib/open3.rbi
@@ -99,7 +99,7 @@ module Open3
   sig do
     params(
       cmd: T.any(String, T::Array[String]),
-      opts: T::Hash[Symbol, T.untyped],
+      opts: T.untyped,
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_MBsNTe0zcH0nyZ
→ https://go/builds/bui_MBsNMXjWzmcRHA

froydnj (approved):

rbi/stdlib/open3.rbi
```diff
@@ -99,7 +99,7 @@ module Open3
   sig do
     params(
       cmd: T.any(String, T::Array[String]),
-      opts: T::Hash[Symbol, T.untyped],
```
jez, line 102: This required that every keyword arg be provided a hash

```ruby
f(x: {})
```

instead of allowing any keyword args

```ruby
f(x: '')
```

These errors were all previously silenced because of #6169 (it's common to call these methods with splatted positional arguments).

PR #6175, opened 2022-08-05
wip: Thread censorForSnapshotTests to LLVM IR tests

jez:
### Test plan

See included automated tests.


jez:
@froydnj This approach doesn't seem to work—if you have any ideas for how we can get rid of `<static-init>$154` IDs from our LLVM IR expectation tests, I'd love to hear it.

Right now it means that any time someone wants to add a new file to Sorbet's payload, they have to spend a ton of time recomputing snapshot files for the compiler.

PR #6176, opened 2022-08-05
Re-enable "fast path preempts slow path" for beta

jez:
### Motivation

We temporarily disabled letting fast path edits preempt slow path edits because
I misunderstood how preemptions worked.

If a fast path preempts a slow path, the slow path still runs (full,
non-incremental) resolver on the old file contents, then the fast path
edited-file gets incremental resolver run on it, then the slow path resumes.

What can sometimes actually happen as a result of this is that the fast path
actually runs inference on a file before the slow path gets a chance to type
check it! When that happens, the GlobalState will have been mutated to delete
certain symbols, but the old version of the AST on the slow path will still
mention symbol ref IDs that only existed in the old GlobalState, and don't exist
anymore.

This sort of "oh whoops, the new GlobalState actually has things that cause the
old slow path ASTs to seemingly discover type errors" was always a problem,
which was solved by simply only flushing errors for the last version of the
file.

That was fine when the worst thing that could happen was that "eh, maybe we
spend too much time computing errors that will be ignored" but in the fast path
methods beta, typechecking a file that should not be typechecked anymore
actually manifests in crashes, because symbols don't actually exist that used
to.

This PR fixes that by allowing the error reporter to indicate whether it would
have skipped reporting the errors.

It might be more apt to start calling `wouldReportErrors` something more
obviously load bearing, like "was already typechecked on newer epoch" so I'm
open to suggestions.

### Test plan

We did manage to write a multithreaded_protocol_test_corpus.cc test for this
earlier. It passes now, which is nice.

I will still want to do a fair bit of manual testing to make sure that this does
in fact fix the problems with crashes, as while I don't _think_ this involves
contended access to shared mutable state (and instead accesses only immutable
state), of the recent changes I've made it's the closest to come to that.

### DIFF
--- a/core/ErrorCollector.h
+++ b/core/ErrorCollector.h
@@ -12,6 +12,9 @@ private:
 public:
     ErrorCollector() = default;
     ~ErrorCollector() = default;
+    bool wouldFlushErrors(core::FileRef file) override {
+        return true;
+    }
     void flushErrors(spdlog::logger &logger, const core::GlobalState &gs, core::FileRef file,
                      std::vector<std::unique_ptr<core::ErrorQueueMessage>> errors) override;
     std::vector<std::unique_ptr<core::Error>> drainErrors();
--- a/core/ErrorCollector.h
+++ b/core/ErrorCollector.h
@@ -12,7 +12,7 @@ private:
 public:
     ErrorCollector() = default;
     ~ErrorCollector() = default;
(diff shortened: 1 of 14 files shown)


jez:

→ https://go/builds/bui_MBwmQRkTphwH5E
→ https://go/builds/bui_MBwmHZCxXF7kCJ

froydnj (approved):

PR #6179, opened 2022-08-06
make static-init names in the compiler less dependent on compilation environment

froydnj:
### Motivation

As pointed out in #6175, the unique names for static-init methods leaking into the LLVM files make adding new payload files easier.  They also inhibit compiler determinism, so let's just handle static-init method names separately.

(I suppose the uniqueness leaking into the LLVM files also makes other things problematic as well, but perhaps that is less of a problem and something to be addressed separately.  Here we know that the names are globally unique.)

### Test plan

See included automated tests.


jez (approved):
Great thanks!

PR #6185, opened 2022-08-07
Add `::DATA` constant

### LINKED ISSUE #921: Support for DATA/__END__
#### Input

[→ View on sorbet.run](https://sorbet.run/#DATA.each_line%20do%20%7Cline%7C%0A%20%20puts%20line%0Aend%0A%0A__END__%0Aaaa%0Abbb%0Acc%0A)

```ruby
DATA.each_line do |line|
  puts line
end

__END__
aaa
bbb
cc
```

#### Observed output

```
editor.rb:1: Unable to resolve constant DATA https://srb.help/5002
     1 |DATA.each_line do |line|
        ^^^^
```

#### Expected behavior

Should have identified and ignored the special 'DATA' constant(?).

jez:
### Motivation

Fixes #921

It might be nicer to figure out whether any file in the program mentions
the `__END__` token, and handle defining this by way of a desugaring
that adds something like `::DATA = T.let(T.unsafe(nil), IO)` to the end
of the file that had the `__END__`, but our parser currently does not
leave any trace of the fact that the file contained this token (instead,
it merely jumps the lexer's position straight to the end of the file).

We could obviously change that, but it seemed like possibly more work
than it's worth, because

- We already have precedent for other things in the payload that appear
  to be always defined, but in fact are only defined once e.g. some file
  has been required.

- Doing means that if you tried to define a class called `DATA`, you
  would get a warning that you were redefining a static field as a
  class, which would give you an indication that maybe you shouldn't try
  to use this name that the Ruby VM treats as pseudo-reserved.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/constants.rbi
+++ b/rbi/core/constants.rbi
@@ -42,3 +42,5 @@
 ::TOPLEVEL_BINDING = T.let(T.unsafe(nil), Binding)
 # An obsolete alias of `true`
 ::TRUE = T.let(T.unsafe(nil), TrueClass)
+# Constant that is available if using the __END__ keyword
+::DATA = T.let(T.unsafe(nil), IO)
--- a/test/testdata/parser/end_data.rb
+++ b/test/testdata/parser/end_data.rb
@@ -1,0 +1,11 @@
+# typed: true
+
+DATA.each_line do |line|
+  puts line
+end
+
+__END__
+aaa
+bbb
+cc
+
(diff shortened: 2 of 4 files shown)


froydnj (approved):

jez:

→ https://go/builds/bui_MD1DqvX9rD5yKS
→ https://go/builds/bui_MD1DFkhD4kvoGg

PR #6186, opened 2022-08-07
wip: Try making `Object#=~` return `T.noreturn`

### LINKED ISSUE #707: Unused code detector fails when using `Object =~ Regex`
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Aa%20%3D%20T.let('hey'%2C%20Object)%0A%0Aif%20a%20%3D~%20%2Fhey%2F%0A%20%20puts%20'yes'%0Aelse%0A%20%20puts%20'no'%0Aend)

```ruby
# typed: strict
extend T::Sig

a = T.let('hey', Object)

if a =~ /hey/
  puts 'yes'
else
  puts 'no'
end
```

#### Observed output

```
editor.rb:7: This code is unreachable https://sorbet.org/docs/error-reference#7006
     7 |  puts 'yes'
               ^^^^^
Errors: 1
```

#### Expected behavior

The code is reachable, so no error expected

jez:
### Motivation

Fixes #707

(maybe)

The best solution would be to mark this method as not actually defined. Unfortunately, if someone ever tried to run missing method generation, it would just get defined in a different RBI, and that time be untyped.

Maybe keeping it as `returns(NilClass)` is best?

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -627,11 +627,20 @@ module Kernel
   end
   def ===(other); end
 
+  # =~ is on Object is deprecated, and is almost always a mistake to use.
+  #
+  # Likely what you meant to do is call `=~` on `String` or `Symbol`--you can
+  # do so by first checking whether the type is correct first.
+  #
+  # This method has been deprecated in Ruby, and may be removed at some point
+  # in the future. In the mean time, we've assigned a sig of `T.noreturn` to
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_MCMS5eJqFaQsm4
→ https://go/builds/bui_MCMSvKs5ZOXpUl

PR #6187, opened 2022-08-07
Report error for code after unconditional return

### LINKED ISSUE #761: Missing unreachable code error for puts after return
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0Asig%20%7B%20returns(Integer)%20%7D%0Adef%20foo%0A%20%20return%203%0A%20%20puts%0Aend%0A)

```ruby
# typed: true
extend T::Sig
sig { returns(Integer) }
def foo
  return 3
  puts
end
```

#### Observed output

No errors.

#### Expected behavior

Unreachable code error on the `puts`.

jez:
### Motivation

Fixes #761

### Test plan

See included automated tests.

### DIFF
--- a/infer/inference.cc
+++ b/infer/inference.cc
@@ -90,13 +90,9 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
 
                 if (auto e = ctx.beginError(bind.loc, core::errors::Infer::DeadBranchInferencer)) {
                     e.setHeader("This expression appears after an unconditional return");
-                    for (const auto &prevBasicBlock : bb->backEdges) {
-                        const auto &prevEnv = outEnvironments[prevBasicBlock->id];
-                        if (prevEnv.isDead) {
-                            continue;
-                        }
-                    }
                 }
+
+                // Only report the expression expression in the dead block
                 break;
             }
             continue;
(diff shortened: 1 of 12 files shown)


jez:

→ https://go/builds/bui_MCMswr4E6bV82R
→ https://go/builds/bui_MCMsDh8juv2u7D

infer/inference.cc
```diff
@@ -83,6 +83,22 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
...
+                }
+
+                if (auto e = ctx.beginError(bind.loc, core::errors::Infer::DeadBranchInferencer)) {
+                    e.setHeader("This expression appears after an unconditional return");
+                    for (const auto &prevBasicBlock : bb->backEdges) {
+                        const auto &prevEnv = outEnvironments[prevBasicBlock->id];
+                        if (prevEnv.isDead) {
+                            continue;
+                        }
+                    }
```
froydnj, line 98: This loop doesn't appear to do anything?  Did you have some code here previously but decided to delete it?

froydnj (approved):

infer/inference.cc
jez: Ah yeah, at first I tried to print something like, "made dead because of this `return`" instruction or something like that but then I gave up on it.

PR #6201, opened 2022-08-09
Add fast path test with abstract and override

jez:
### Motivation

I had a TODO to make sure this worked before moving the fast path methods change
out of beta, and it looks like it also naturally fell out of the changes in
#5808.

### Test plan

Test-only change

### DIFF
--- a/test/testdata/lsp/fast_path/abstract_method__1.1.rbupdate
+++ b/test/testdata/lsp/fast_path/abstract_method__1.1.rbupdate
@@ -1,0 +1,13 @@
+# typed: true
+# assert-fast-path: abstract_method__1.rb,abstract_method__2.rb
+# spacer for exclude from file update
+
+module IFoo
+  extend T::Sig
+  extend T::Helpers
+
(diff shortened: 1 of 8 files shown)


test/testdata/lsp/fast_path/abstract_method__1.1.rbupdate
```diff
@@ -0,0 +1,13 @@
+# typed: true
+# assert-fast-path: abstract_method__1.rb,abstract_method__2.rb
+# spacer for exclude from file update
```
froydnj, line 3: Total nit: "spacer for exclude-from-file-update"?  (And likewise everywhere else?)

froydnj (approved):

test/testdata/lsp/fast_path/abstract_method__1.1.rbupdate
jez: Sure, I can update all of these tests in a follow up PR

PR #6205, opened 2022-08-10
Handle methods that dispatch to intrinsics on fast path

jez:
### Motivation

Some of our intrinsics dispatch to other methods. Given that we currently use names not symbols to list the downstream files that might need to be retypechecked, it means we have to also record usages of these dispatched-to names as well.

This mostly affects stale errors that could have been occurring around usages of `<call-with-*>` helpers. It could also have affected `initialize`/`new`, but in a large enough codebase and with the changes we made to compute the "max files on fast path" in canTakeFastPath instead of in runFastPath, it was unlikely to still have much effect there (of course, it would matter in small codebases with fewer than 100 files).

### Test plan

See included automated tests.

### DIFF
--- a/ast/substitute/Substitute.cc
+++ b/ast/substitute/Substitute.cc
@@ -124,7 +124,7 @@ public:
             // We're lucky because all of these methods have the symbol they dispatch to as the
             // positional arg at index 1.
             auto name = unwrapLiteralToName(send.getPosArg(1));
-            ENFORCE(name.exists());
+            ENFORCE(name.exists(), "Name should exist because this arg should be a Literal");
             [[maybe_unused]] auto _substituted = subst.substituteSymbolName(name);
         }
 
--- a/ast/substitute/Substitute.cc
+++ b/ast/substitute/Substitute.cc
@@ -98,8 +98,9 @@ public:
     void preTransformSend(core::MutableContext ctx, ExpressionPtr &original) {
         auto &send = cast_tree_nonnull<Send>(original);
         if (send.fun == core::Names::aliasMethod()) {
-            // This is basically a MethodDef in disguise, so we have to do similar logic to record
-            // the names that the MethodDef case would.
+            // `alias_method :foo :bar` is very similar to `def foo; self.bar(); end`, so in
+            // addition to handling the `alias_method` fun, we also want to look at the two
+            // possibly-Symbol-literal arguments and call substituteSymbolName on them as well.
 
             // Discards the new name. We only care to do this for the side effect of recording the entry in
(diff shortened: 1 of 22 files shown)


core/Symbols.h
```diff
@@ -29,6 +29,18 @@ class SerializerImpl;
...
+    // Used to ensure that calls involving intrinsics are correctly typechecked on the fast path.
+    //
+    // If your intrinsic does not dispatch to another method, the default implementation returns an
+    // empty vector, so you can elide an implementation.
+    //
+    // If your intrinsic cannot declare the list of methods it might dispatch to (like
+    // `<call-with-splat>`, etc.), you will have to edit Substitute.cc.
+    virtual std::vector<NameRef> dispatchesTo() const {
+        return {};
+    }
```
froydnj, line 42: Can we move this out-of-line?

(The first non-abstract virtual method in a class is special, and triggers the output of all kinds of information in the compilation unit which it appears; if that method is defined in a header...the information is output in every file where that header is included.  https://itanium-cxx-abi.github.io/cxx-abi/abi.html#vague-vtable )

ast/substitute/Substitute.cc
```diff
@@ -116,9 +117,26 @@ template <typename T> class SubstWalk {
                     [[maybe_unused]] auto _substituted = subst.substituteSymbolName(name);
                 }
             }
+        } else if (send.fun == core::Names::callWithSplat() || send.fun == core::Names::callWithBlock() ||
+                   send.fun == core::Names::callWithSplatAndBlock() || send.fun == core::Names::checkAndAnd()) {
+            ENFORCE(send.numPosArgs() > 2, "These are special desugar methods, should have at least two args");
+
+            // We're lucky because all of these methods have the symbol they dispatch to as the
+            // positional arg at index 1.
+            auto name = unwrapLiteralToName(send.getPosArg(1));
```
froydnj, line 126: Perhaps we should `ENFORCE` that the arg is actually a literal?  I guess it's basically the same thing as `ENFORCE`'ing that the name exists further down, but it is a little more clear if the assertion triggered.

froydnj (approved):

ast/substitute/Substitute.cc
jez: Yeah it seemed to me like doing the ENFORCE earlier would have just duplicated the contents of the `unwrapLiteralToName` function. I've added a message to the ENFORCE below to make that hopefully more clear.

core/Symbols.h
jez: I've re-requested review to make sure I correctly did the thing you were asking for here.

froydnj (approved):

PR #6208, opened 2022-08-10
provide an escape hatch for github release archive builds

froydnj:
### Motivation

Stripe's Sorbet build process builds Sorbet inside a git checkout of Stripe's monorepo.  The Sorbet directory is obtained from a Github release tarball, rather than a checkout from Sorbet's repository; there is no `.git` directory in the release tarball.  Which means that the `get_workspace_status` script picks up git information from the monorepo, which means that the information displayed in e.g. `sorbet --version` or LSP logs is based on Stripe's monorepo, rather than Sorbet's repository.

This makes figuring out exactly what version of Sorbet is running via a log or similar an elliptical exercise.

In the build process, we could move the Sorbet directory outside of a git checkout entirely, but then we wouldn't get any useful information in e.g. `sorbet --version`.

The compromise proposed here is that the build process creates a `.github/SORBET_ARCHIVE_SHA` file (name open to bikeshedding) with the git revision for the archive, and then `get_workspace_status` can read that file to at least be able to display the Sorbet SHA from which the release was built.  I looked for ways to have the Github release process automatically insert such things into the tarball, but did not see anything obvious (and I'm sure that enabling projects to customize the tarball generation process would be subject to limitations anyway, and it's probably simpler to just not permit customization).

### Test plan

I have verified that a Sorbet built from within a monorepo checkout, but with this patch applied first, has the expected version information present.  I have not yet modified the necessary build scripts on Stripe's side, but I don't expect that process to uncover any problems.

### DIFF
--- a/tools/buildstamp/get_workspace_status
+++ b/tools/buildstamp/get_workspace_status
@@ -14,6 +14,17 @@ set -e
 # If the script exits with non-zero code, it's considered as a failure
 # and the output will be discarded.
 
+# Provide an escape hatch for people building from github release tarballs,
+# but in non-Sorbet git repositories, to get somewhat more accurate version
+# information.
+if [ -f .github/SORBET_ARCHIVE_SHA ]; then
+    git_rev=$(head -n 1 .github/SORBET_ARCHIVE_SHA)
+    echo "STABLE_BUILD_SCM_REVISION ${git_rev}"
+    echo "STABLE_BUILD_SCM_COMMIT_COUNT 0"
+    echo "STABLE_BUILD_SCM_CLEAN 0"
+    exit 0
+fi
+    
 if ! git rev-parse --git-dir >&/dev/null; then
     echo "STABLE_BUILD_SCM_REVISION nongit"
     echo "STABLE_BUILD_SCM_COMMIT_COUNT 0"


tools/buildstamp/get_workspace_status
```diff
@@ -14,6 +14,17 @@ set -e
 # If the script exits with non-zero code, it's considered as a failure
 # and the output will be discarded.
 
+# Provide an escape hatch for people building from github release tarballs,
+# but in non-Sorbet git repositories, to get somewhat more accurate version
+# information.
+if [ -f .github/SORBET_ARCHIVE_SHA ]; then
+    git_rev=$(head -n 1 .github/SORBET_ARCHIVE_SHA)
+    echo "STABLE_BUILD_SCM_REVISION ${git_rev}"
+    echo "STABLE_BUILD_SCM_COMMIT_COUNT 0"
```
jez, line 23: This commit count gets used to compute the version number, like `0.5.10301`.

Do you think we could get that information as well?

tools/buildstamp/get_workspace_status
```diff
@@ -14,6 +14,17 @@ set -e
 # If the script exits with non-zero code, it's considered as a failure
 # and the output will be discarded.
 
+# Provide an escape hatch for people building from github release tarballs,
+# but in non-Sorbet git repositories, to get somewhat more accurate version
+# information.
+if [ -f .github/SORBET_ARCHIVE_SHA ]; then
```
jez, line 20: I'm not particularly in love with putting our own config inside `.github/`. Out of curiosity, have you asked anyone on the build team about what their preferences are?

tools/buildstamp/get_workspace_status
froydnj: I don't think so: getting the commit count requires access to the git repository (or at least the git checkout), which we don't have.

The tarball available as a github release does contain the version number information in the unzipped directory, so you could parse things out of that directory name, but the tarball on the Stripe side for whatever reason only has the SHA.

tools/buildstamp/get_workspace_status
froydnj: I don't have strong opinions about where the file goes; putting it in `.github` seemed like a reasonable thing to do.  `.SORBET_ARCHIVE_SHA` is also a possibility.

I did not want to try to futz with setting an environment variable from Bazel's command-line and ensuring that it propagated into this script's execution, which is probably the best way to go about this.

jez (approved):
I am generally on board with the concept. I would love if we could also get the commit count, which I think we could do by tracking more information in Stripe's bump-sorbet script.

I would also love if we quickly checked if the build team has any preferences for where to stage this information during the build.

otherwise lgtm

tools/buildstamp/get_workspace_status
froydnj: For posterity, `--{host_,}action_env` variables do not get propagated into this script's execution environment, so going through the filesystem seems like the only way forward.

The build team did not have strong opinions on this, given that it's cross-bazel invocation communication; they did note that they have done similar things in the past.

PR #6210, opened 2022-08-10
Improve error message for variance mismatch with parent

### LINKED ISSUE #5678: Bleak error message for variance mismatch with parent
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Amodule%20Parent%0A%20%20extend%20T%3A%3AGeneric%0A%20%20Elem%20%3D%20type_member%28%3Aout%29%0Aend%0A%0Amodule%20Child%0A%20%20extend%20T%3A%3AGeneric%0A%20%20include%20Parent%0A%20%20Elem%20%3D%20type_member%28%3Ain%29%0Aend)

```ruby
# typed: true

module Parent
  extend T::Generic
  Elem = type_member(:out)
end

module Child
  extend T::Generic
  include Parent
  Elem = type_member(:in)
end
```

#### Observed output

```
editor.rb:11: Type variance mismatch with parent Parent https://srb.help/5015
    11 |  Elem = type_member(:in)
          ^^^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

- Should show what the variance in the child and parent is
- Should attempt to explain (briefly) why this is not allowed
- Should show the location of the matching type member in the parent

To top it all off, there are no tests for this in Sorbet right now.

- - -

adisonlampert:
### Motivation

Closes #5678
- [x] Should show what the variance in the child and parent is
- [x] Should attempt to explain (briefly) why this is not allowed
- [x] Should show the location of the matching type member in the parent

Screen Shot 2022-09-19 at 9 12 57 AM

### Test plan

Added automated test.

### DIFF
--- a/resolver/GlobalPass.cc
+++ b/resolver/GlobalPass.cc
@@ -82,7 +82,11 @@ bool resolveTypeMember(core::GlobalState &gs, core::ClassOrModuleRef parent, cor
     if (!sym.data(gs)->derivesFrom(gs, core::Symbols::Class()) && myVariance != parentVariance &&
         myVariance != core::Variance::Invariant) {
         if (auto e = gs.beginError(myTypeMember.data(gs)->loc(), core::errors::Resolver::ParentVarianceMismatch)) {
-            e.setHeader("Type variance mismatch with parent `{}`", parent.show(gs));
+            e.setHeader("Type variance mismatch for `{}` with parent `{}`. Child `{}` should be `{}` or invariant, but "
+                        "it is `{}`",
+                        name.show(gs), parent.show(gs), sym.show(gs), core::Polarities::showVariance(parentVariance),
+                        core::Polarities::showVariance(myVariance));
+            e.addErrorLine(parentTypeMember.data(gs)->loc(), "Parent `{}` declared here", parent.show(gs));
         }
         return true;
     }
--- a/resolver/GlobalPass.cc
+++ b/resolver/GlobalPass.cc
@@ -82,10 +82,11 @@ bool resolveTypeMember(core::GlobalState &gs, core::ClassOrModuleRef parent, cor
     if (!sym.data(gs)->derivesFrom(gs, core::Symbols::Class()) && myVariance != parentVariance &&
         myVariance != core::Variance::Invariant) {
         if (auto e = gs.beginError(myTypeMember.data(gs)->loc(), core::errors::Resolver::ParentVarianceMismatch)) {
-            e.setHeader("Type variance mismatch for `{}` with parent `{}`. Child `{}` should be `{}` or invariant, but "
+            auto orInvariant = parentVariance == core::Variance::Invariant ? "" : " or invariant";
+            e.setHeader("Type variance mismatch for `{}` with parent `{}`. Child `{}` should be `{}`{}, but "
                         "it is `{}`",
                         name.show(gs), parent.show(gs), sym.show(gs), core::Polarities::showVariance(parentVariance),
-                        core::Polarities::showVariance(myVariance));
+                        orInvariant, core::Polarities::showVariance(myVariance));
             e.addErrorLine(parentTypeMember.data(gs)->loc(), "Parent `{}` declared here", parent.show(gs));
         }
         return true;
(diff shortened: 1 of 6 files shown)


resolver/GlobalPass.cc
```diff
@@ -82,7 +82,12 @@ bool resolveTypeMember(core::GlobalState &gs, core::ClassOrModuleRef parent, cor
     if (!sym.data(gs)->derivesFrom(gs, core::Symbols::Class()) && myVariance != parentVariance &&
         myVariance != core::Variance::Invariant) {
         if (auto e = gs.beginError(myTypeMember.data(gs)->loc(), core::errors::Resolver::ParentVarianceMismatch)) {
-            e.setHeader("Type variance mismatch with parent `{}`", parent.show(gs));
+            e.setHeader(
+                "Type variance mismatch with parent `{}`; child `{}` should be `{}` or invariant, but it is `{}`",
+                parent.show(gs), sym.show(gs), core::Polarities::showVariance(parentVariance),
+                core::Polarities::showVariance(myVariance));
+            e.addErrorLine(parentTypeMember.data(gs)->loc(), "Variance for parent `{}` declared here", parent.show(gs));
```
jez, line 89: This is a bit of a nit, but technically this error line points not to the variance, but to the type member itself. Maybe this could say simply, ``"Parent `{}` declared here"``.

resolver/GlobalPass.cc
```diff
@@ -82,7 +82,12 @@ bool resolveTypeMember(core::GlobalState &gs, core::ClassOrModuleRef parent, cor
...
+            e.addErrorNote("Sorbet requires compatible variance between parent and child in order to type check.");
```
jez, line 90: I don't think this error note is adding much value (it doesn't say anything new that isn't said earlier). For example, we could have a note like this on almost every error in Sorbet ("Sorbet requires compatible argument types in order to type check" / "Sorbet requires compatible return types in order to type check" / etc.)

My suggestion might be to instead direct people to read the error documentation for this error code (5015), as there is an extended example outlining why this "compatible variance" check is important.

resolver/GlobalPass.cc
```diff
@@ -82,7 +82,12 @@ bool resolveTypeMember(core::GlobalState &gs, core::ClassOrModuleRef parent, cor
     if (!sym.data(gs)->derivesFrom(gs, core::Symbols::Class()) && myVariance != parentVariance &&
         myVariance != core::Variance::Invariant) {
         if (auto e = gs.beginError(myTypeMember.data(gs)->loc(), core::errors::Resolver::ParentVarianceMismatch)) {
-            e.setHeader("Type variance mismatch with parent `{}`", parent.show(gs));
+            e.setHeader(
+                "Type variance mismatch with parent `{}`; child `{}` should be `{}` or invariant, but it is `{}`",
```
jez, line 86: If the class has multiple type members, this error message does not mention the name of which of the type members is incompatible. Do you think you could add that information somewhere? It should be as simple as adding `name.show(gs)` somewhere in the error message header.

jez (changes requested):
Thanks! Some suggestions.

froydnj (approved):
Thanks!

test/testdata/resolver/type_member_parent_missing.rb
```diff
@@ -11,7 +11,7 @@ class Bar1 < Foo1
 # ^^^^^^^^^^^^^^^^^ error: Missing definition for abstract method `Enumerable#each`
     Elem = type_member(:out)
   # ^^^^^^^^^^^^^^^^^^^^^^^^ error: Classes can only have invariant type members
-  # ^^^^^^^^^^^^^^^^^^^^^^^^ error: Type variance mismatch with parent `Foo1`
+  # ^^^^^^^^^^^^^^^^^^^^^^^^ error: Type variance mismatch for `Elem` with parent `Foo1`. Child `Bar1` should be `:invariant` or invariant, but it is `:out`
```
jez, line 14: Sorry for not noticing this earlier: "`:invariant` or invariant" is redundant and also confusing, because `:invariant` is not a valid variance annotation.

jez: I went ahead and pushed up a fix for this to your branch.

jez (approved):

PR #6212, opened 2022-08-10
delete comments about `handleAbstractMethod`

froydnj:
### Motivation

Review comments on these were not enthusiastic, and coming back to them after some time, I was briefly confused by them anyway.  So let's just delete them.

### Test plan

See included automated tests.


jez (approved):
😅

PR #6214, opened 2022-08-10
Track name kind more accurately in fast path decision

jez:
### Motivation

This isn't _exactly_ a problem right now, because:

- it's uncommon for people to define methods with uppercase names
- especially uppercase names that share names with constants
- the worst thing that happens is that it appears as though we have to typecheck
  _extra_ files on the fast path, but otherwise doesn't interfere with whether
  we can take the fast path or not

But in #6206 I'm starting to work on having changes to `type_member`'s take the
fast path, which will start making constant names more common here.

So I'd like to just properly handle this.

### Test plan

See included automated tests.

Note that these changes apply independent of any already-landed or still-ongoing work to take the fast path more often.

### DIFF
--- a/core/FileHash.cc
+++ b/core/FileHash.cc
@@ -40,9 +40,9 @@ uint32_t hashFullNameRef(const GlobalState &gs, NameRef nm) {
 
 } // namespace
 
-ShortNameHash::ShortNameHash(const GlobalState &gs, NameRef nm) : _hashValue(incZero(_hash(nm.shortName(gs)))){};
+WithoutUniqueNameHash::WithoutUniqueNameHash(const GlobalState &gs, NameRef nm) : _hashValue(incZero(_hash(nm.shortName(gs)))){};
 
-void ShortNameHash::sortAndDedupe(std::vector<core::ShortNameHash> &hashes) {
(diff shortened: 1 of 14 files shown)


core/FileHash.cc
```diff
@@ -13,6 +13,44 @@ uint32_t incZero(uint32_t a) {
...
+            break;
+        }
+    }
+
+    if (kind != NameKind::UNIQUE) {
+        auto hashedKind = static_cast<underlying_type<NameKind>::type>(kind);
+        result = mix(result, hashedKind);
+    }
+    return result;
+}
```
froydnj, line 52: This code follows the structure of `hashFullNameRef` and consistency is a good thing, so the below is only a suggestion.

Probably slightly better to structure this such that the `hashNameRefWithoutUniques` calls appear in tail position.  So either duplicating the `kind != NameKind::UNIQUE` conditional into the appropriate cases, or moving the conditional before the `switch` (and making sure `result` is initialized in the `UNIQUE` case).

I think the code duplication is actually slightly better; the compiler should be able to constant-fold away the `mix` on the `NameKind` (since `NameKind` will be a constant in each case of the switch), so the performance should be better.

froydnj (approved):

PR #6215, opened 2022-08-10
Redefine methods with correct visibility upfront

### LINKED ISSUE #5786: Sorbet defines private method in non-private scope first, causing it to be seen by another library
#### Input

~→ View on sorbet.run~

`some_rails_generator.rb`
```ruby
# typed: strict
# frozen_string_literal: true

require "rails/generators"

class SomeRailsGenerator < Rails::Generators::NamedBase
  extend T::Sig
  source_root File.expand_path("../templates", __FILE__)

  sig { void }
  def do_generation_stuff
    puts "Creating a file here"
  end

  private

  sig { void }
  def a_private_method
    puts "Doing private stuff"
  end
end
```

(Credit to @paracycle for originally deriving the root issue here.)

When using Sorbet method signatures in a Rails generator, powered by Thor, private methods are briefly defined in a public scope and then their scope is redefined while [disabling the `method_added` hooks](https://github.com/sorbet/sorbet/blob/2db8204979c789b262c3a5240e5a86cbd4493229/gems/sorbet-runtime/lib/types/private/class_utils.rb#L76-L109), but other libraries `method_added` hooks can not be disabled. This leads a library like Thor to have its `method_added` hook triggered and result in it seeing the method as a public method initially.

#### Observed output

```
$ rails g some_rails a_name
Creating a file here
Could not find command "a_private_method".
```

#### Expected behavior

```
$ rails g some_rails a_name
Creating a file here
```

#### Additional Information

Rails 7.0.3
Thor 1.2.1
sorbet-static-and-runtime 0.5.9964

##### Workaround

Wrapping the methods in a Thor `no_commands` block resolves this.
```ruby
  private

  no_commands do
    sig { void }
    def a_private_method
      puts "Doing private stuff"
    end
  end
```

Removing method signatures also resolves this.
```ruby
  private

  def a_private_method
    puts "Doing private stuff"
  end
```

paracycle:
### Motivation

Fixes #5786

Sorbet runtime updates the visibility of the methods it is replacing but sometimes that happens too late. For example, if user code has a `method_added` hook and checks the visibility of the method that's added in the hook, it would always see `public` when Sorbet replaces a `private` or `protected` method. This was happening because the setting of the visibility was happening after the call to `define_method` and by then it was too late.

Instead of setting the visibility after defining the method, this PR encapsulates a method definition operation inside a method named `def_with_visibility` on the `T::Private::ClassUtils` module, which does a `module_exec` on the passed in module so that it can create a `public`/`protected`/`private` section and only then define the method. This ensures that the method is defined with the expected visibility from the beginning.

### Test plan

See included automated tests. Added a test to verify that `method_added` hooks see methods that are replaced by Sorbet runtime with the expected visibility.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/class_utils.rb
+++ b/gems/sorbet-runtime/lib/types/private/class_utils.rb
@@ -69,6 +69,27 @@ module T::Private::ClassUtils
     end
   end
 
+  def self.def_with_visibility(mod, name, visibility, method=nil, &block)
+    mod.module_exec do
+      # Start a visibility (public/protected/private) region, so that
+      # all of the method redefinitions happen with the right visibility
+      # from the beginning. This ensures that any other code that is
+      # triggered by `method_added`, sees the redefined method with the
+      # right visibility.
+      send(visibility)
+
+      if method
+        define_method(name, method)
+      else
+        define_method(name, &block)
+      end
+
+      if block && block.arity < 0 && respond_to?(:ruby2_keywords, true)
+        ruby2_keywords(name)
+      end
+    end
+  end
+
   # Replaces a method, either by overwriting it (if it is defined directly on `mod`) or by
   # overriding it (if it is defined by one of mod's ancestors). Returns a ReplacedMethod instance
   # on which you can call `bind(...).call(...)` to call the original method, or `restore` to
@@ -96,13 +117,9 @@ module T::Private::ClassUtils
     overwritten = original_owner == mod
     T::Configuration.without_ruby_warnings do
       T::Private::DeclState.current.without_on_method_added do
-        mod.send(:define_method, name, &blk)
-        if blk.arity < 0 && mod.respond_to?(:ruby2_keywords, true)
-          mod.send(:ruby2_keywords, name)
-        end
+        def_with_visibility(mod, name, original_visibility, &blk)
       end
     end
-    mod.send(original_visibility, name)
     new_method = mod.instance_method(name)
 
     ReplacedMethod.new(mod, original_method, new_method, overwritten, original_visibility)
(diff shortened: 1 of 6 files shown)


jez:
Thanks, I've kicked off a build on Stripe's codebase.

http://go/builds/bui_MDp1brTspM2YrB

jez (changes requested):
It looks like you have some test failures? Feel free to re-request review when they pass, I expect the change itself is fine but I'll look more closely when the tests pass.

paracycle:
> It looks like you have some test failures? Feel free to re-request review when they pass, I expect the change itself is fine but I'll look more closely when the tests pass.

It is mostly that I had missed the comment on top of the `call_validation` files and updated them manually. Now I fixed the generator, so it should be fine.

gems/sorbet-runtime/lib/types/private/class_utils.rb
```diff
@@ -69,6 +69,27 @@ def to_s
...
+
+      if method
+        define_method(name, method)
+      else
+        define_method(name, &block)
+      end
+
+      if block && block.arity < 0 && respond_to?(:ruby2_keywords, true)
+        ruby2_keywords(name)
+      end
```
paracycle, line 89: I encapsulated this `ruby2_keywords` call in this method as well, but I am not sure if that's the best thing to do. The `define_method` calls originating from `call_validation` were never changing the `ruby2_keywords` status of methods, but I am not sure if that was because it was forgotten or if it was intentional.

I'd be ok with moving this part out back to the original call sites if needed.

jez: Honestly I think you'd be better at telling me where this `ruby2_keywords` call has to go, as Stripe's codebase is still using Ruby 2.7, and so is unlikely to turn up any problems from incorrectly placement of this.

jez:
> Thanks, I've kicked off a build on Stripe's codebase.
> 
> http://go/builds/bui_MDp1brTspM2YrB

There was one problem, which was actually from a previous change to sorbet-runtime that we simply forgot to update to.

Fixing that one problem led to this build http://go/builds/bui_MDpByg3Qlsga8x which passed.

jez (approved):

paracycle:
Ok, the build is green now!

Issue #6200, opened 2022-08-09
Completion snippets for `T::Struct` and `T::Enum`


jez:
It would be nice if Sorbet had built-in support for some non-symbol-based snippets in completion requests. Two off the top of my head:

- [ ] Snippet for `T::Struct`

  ```ruby
  struct<TAB>
  ```

  could become

  ```ruby
  class ${1} < T::Struct
    ${0}
  end
  ```

- [ ] Snippet for `T::Enum`

  ```ruby
  enum<TAB>
  ```

  could become

  ```ruby
  class ${1} < T::Enum
    enums do
      ${0}
    end
  end
  ```

jez:
Fixed in #6216

PR #6216, opened 2022-08-10
Added completions for struct and enum

maruth-stripe:
Added completions for `enum<TAB>` and `struct<TAB>` 

### Motivation

https://github.com/sorbet/sorbet/issues/6200

### Test plan

See included automated tests.


test/testdata/lsp/completion/enum.rb
```diff
@@ -0,0 +1,5 @@
+# typed: true
+class Test
+  enum # error: does not exist
+  #   ^ apply-completion [A] item: 0
+end
```
jez, line 5: Could we get a couple tests:

- What happens when there is a local variable in scope with the name `enum`?
- What about when there is a method in scope on `self` with the name `enum`?

jez: Another test: can we make sure that

```ruby
x.enum
```

where `x` is some type that has an `enum` method does not produce any results?

main/lsp/requests/completion.cc
```diff
@@ -58,6 +58,7 @@ const RubyKeyword rubyKeywords[] = {
     {"end",
      "The end of a syntax block. Used by classes, modules, methods, exception handling and control expressions."},
     {"ensure", "Starts a section of code that is always run when an exception is raised."},
+    {"enum", "Creates an enum class", "class $1 < T::Enum\nenums do\n$0\nend\nend"},
```
jez, line 61: I believe you still want the indentation, like `\n  enums`. If the tests passed without that, I'd be a little worried 🤔

jez (changes requested):
clever, nice.

the `rubyKeywords` list currently contains exclusively actual keywords (e.g., names that the parser sees as keywords)

if this approach works, you may want to rename some things inside completion.cc from `*keywords` to `*snippets`

you may also want to audit all usages of `rubyKeywords` to see whether it's making an actual assumption that the entries in that array are always a keyword (e.g., in error messages or docs, etc.)

maruth-stripe:
ack, thanks for the feedback, will push updates asap

main/lsp/requests/completion.cc
maruth-stripe: Fixed it! Forgot a `:` in `apply-completion` while writing the test turns out :3. Hence, just about anything was passing.

test/testdata/lsp/completion/enum.rb
maruth-stripe: Added some more tests to cover this!

main/lsp/requests/completion.cc
```diff
@@ -250,6 +259,27 @@ vector<RubyKeyword> allSimilarKeywords(string_view prefix) {
...
+    }
+
+    auto result = vector<KeywordLikeSnippet>{};
+    for (const auto &keywordSnippet : keywordLikeSnippets) {
+        if (absl::StartsWith(keywordSnippet.keyword, prefix)) {
+            result.emplace_back(keywordSnippet);
+        }
+    }
+
+    // The result is trivially sorted because we walked rubyKeywords (which is sorted) in order.
```
jez, line 279:
```suggestion
    // The result is trivially sorted because we walked keywordLikeSnippets (which is sorted) in order.
```

maruth-stripe:
I've separated snippets for which the prefix is to be treated _like_ it were a keyword, but aren't precisely keywords. I add them to the `items` list after local names as they are preferred. 

Ideally I would like to DRY up the code by maybe making a template and passing in the respective array, but wanted to start with this simpler solution

main/lsp/requests/completion.cc
maruth-stripe: ack, apologies for missing

jez (approved):

PR #6217, opened 2022-08-11
Restructure the Field hashing code

jez:
Before, the `fieldShapeHash` function was doing double duty:

- If it was a "normal" static field, it would behave like a proper
  "shape hash" function and ignore the result type.
- If it was a "special" static field or a field, it would actually
  behave like `Field::hash` instead of like a "shape hash".

That implementation was confusing. Structuring the code this way makes
the code more parallel to the method hashing code.

There was also a slight bug that I fixed in this commit:

- Previously, we would _always_ use the `symhash` instead of the
  `staticFieldShapeHash` to mix into `fieldHash`. Luckily, `fieldHash`
  is only used for our metrics, to say when the reason that we took the
  slow path was because a field changed.
- If we took the slow path (say, because a class or module changed),
  *and* also in the same edit a static field's type changed, it would
  previously look like we took the slow path because both a
  ClassOrModule symbol and a Field symbol changed, when in fact we only
  took the slow path because of the change in the former.

The invariant here is that the `<specific>Hash` should always be mixed
with the same thing that the `hierarchyHash` is mixed with.

### Test plan

Should be a no-op, except for one fix for the metrics.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2302,9 +2302,9 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
                        sym.name == Names::requiredAncestorsLin())
                     : true;
             if (needMethodShapeHash) {
-                uint32_t symhash = sym.methodShapeHash(*this);
-                hierarchyHash = mix(hierarchyHash, symhash);
-                methodHash = mix(methodHash, symhash);
+                uint32_t methodShapeHash = sym.methodShapeHash(*this);
+                hierarchyHash = mix(hierarchyHash, methodShapeHash);
+                methodHash = mix(methodHash, methodShapeHash);
             }
 
             counter++;
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2279,12 +2279,16 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
         counter++;
         // No fields are ignored in hashing.
         uint32_t symhash = field.hash(*this);
-        if (field.flags.isStaticField) {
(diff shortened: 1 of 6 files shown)


core/GlobalState.cc
```diff
@@ -2279,12 +2279,16 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
         counter++;
         // No fields are ignored in hashing.
         uint32_t symhash = field.hash(*this);
-        if (field.flags.isStaticField) {
+        if (field.flags.isStaticField && !field.flags.isStaticFieldTypeAlias && !field.isClassAlias()) {
             auto &target = staticFieldHashesMap[WithoutUniqueNameHash(*this, field.name)];
             target = mix(target, symhash);
+            uint32_t staticFieldShapeHash = field.fieldShapeHash(*this);
```
froydnj, line 2285: Rename `fieldShapeHash` to `staticFieldShapeHash`?

(Related, it is very weird that instance variables, static fields, and several other things get stuffed into a single thing.)

froydnj (approved):

core/GlobalState.cc
jez: yeah it's odd. obviously it all just used to be on symbols, and there were no individual classes for various kinds of symbols' data. I have no context on why whoever factored out the symbol kinds (jvilk I think?) decided to put fields and static fields into one `Field` class. Field and static field definitely used to be identified as completely different things:

https://github.com/sorbet/sorbet/blob/e50896e71d02ff90abb4311e06298f0314cbb48d/core/Symbols.h#L63-L69

jez: I'll rename to staticFieldShapeHash in a follow up change.

PR #6218, opened 2022-08-11
Fix fast path bug with removing `T.let` from static field

jez:
### Motivation

Fixes the bug in the test case.

### Test plan

See included automated tests.


core/Symbols.cc
```diff
@@ -2368,6 +2368,7 @@ uint32_t Field::fieldShapeHash(const GlobalState &gs) const {
     ENFORCE(!this->flags.isField && !this->flags.isStaticFieldTypeAlias && !this->isClassAlias());
     uint32_t result = _hash(name.shortName(gs));
 
+    result = mix(result, 1 + (this->resultType != nullptr));
```
jez, line 2371: We (used to) do a similar trick in the methodShapeHash.

This won't matter once we are able to make all static field changes take the fast path.

froydnj (approved):

PR #6227, opened 2022-08-11
Don't be quite so eager to short circuit

jez:
### Motivation

I went to look at the logs to see if most files that had too many files were off
by 1-2 or off by hundreds or thousands, and was greeted by `51 files > 50 files`
exclusively, repeated thousands of times.

### Test plan

test in prod

### DIFF
--- a/main/lsp/LSPFileUpdates.cc
+++ b/main/lsp/LSPFileUpdates.cc
@@ -199,9 +199,12 @@ LSPFileUpdates::fastPathFilesToTypecheck(const core::GlobalState &gs, const LSPC
 
         result.extraFiles.emplace_back(ref);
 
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Just to be clear, this change doesn't actually change anything about the fast path, because of the pre-existing check:

https://github.com/sorbet/sorbet/blob/92001ace2c5d19455b3a499aaad283c0e48c0000/main/lsp/LSPIndexer.cc#L192-L200

but the idea is that it makes the logs slightly more informative, because we're not truncating the actual distribution of files quite so aggressive?

If so, I guess this is OK.

jez:
Yes that’s correct

PR #6234, opened 2022-08-12
use separate stats for fields and static fields

froydnj:
### Motivation

This change could have been done when the original code was landed, but it wasn't.

Also in passing fixes a possible oversight where we weren't serializing all the `LocalSymbolTableHashes` state.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2233,6 +2233,8 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
     uint32_t typeArgumentHash = 0;
     uint32_t typeMemberHash = 0;
     uint32_t fieldHash = 0;
+    uint32_t staticFieldHash = 0;
+    uint32_t staticFieldAliasHash = 0;
     uint32_t methodHash = 0;
     UnorderedMap<WithoutUniqueNameHash, uint32_t> methodHashesMap;
     UnorderedMap<WithoutUniqueNameHash, uint32_t> staticFieldHashesMap;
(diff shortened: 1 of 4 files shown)


core/GlobalState.cc
```diff
@@ -2284,7 +2286,10 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
             target = mix(target, symhash);
             uint32_t staticFieldShapeHash = field.fieldShapeHash(*this);
             hierarchyHash = mix(hierarchyHash, staticFieldShapeHash);
-            fieldHash = mix(fieldHash, staticFieldShapeHash);
+            staticFieldHash = mix(fieldHash, staticFieldShapeHash);
+        } else if (field.flags.isStaticField) {
+            hierarchyHash = mix(hierarchyHash, symhash);
+            staticFieldAliasHash = mix(staticFieldAliasHash, symhash);
```
jez, line 2292: So this will conflict with #6228. I think that in splitting these out, the buckets we are likely to want are

- fields (ivars)
- class aliases (static fields where `isClassAlias()`)
- everything else (normal constants, type aliases)

Those buckets are easier to achieve by rebasing on that PR, but I had been wanting to hold off landing it until a little bit in the future so that we could collect some stats on the impact of the methods change in isolation.

If you'd like to land this PR ahead of that one, it would be possible to simply mimic the logic that #6228 would do if it were landed with a little bit of code duplication here.

And then if you agree with the above (namely that we should put type alias and class alias in separate buckets), then I think `staticFieldClassAliasHash` or simply `classAliasHash` might be a better name than just the `staticFieldAliasHash` name that it's currently using.

I'm open to alternatives.

froydnj: I think I have addressed this feedback with the new version of the PR.

jez (approved):

PR #6237, opened 2022-08-13
Don't require abstract mangle renamed methods to be implemented

jez:
### Motivation

It's possible for a method to be improperly redefined but the error be silenced,
because the definition is in a `# typed: false` file.

This then made it ~impossible to implement the abstract class in a `# typed:
true` file.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/resolver/redefined_abstract.rb
+++ b/test/testdata/resolver/redefined_abstract.rb
@@ -1,0 +1,30 @@
+# typed: true
+
+module IFoo
+  extend T::Sig
+  extend T::Helpers
+  abstract!
+
+  sig {abstract.void}
+  def foo; end
+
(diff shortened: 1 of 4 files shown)


test/testdata/resolver/redefined_abstract.rb
```diff
@@ -0,0 +1,28 @@
+# typed: true
+
+module IFoo
```
froydnj, line 3: Given that the report concerns what appears to be a multi-file test, it seems like we also ought to have a multi-file test for this?

froydnj (approved):

test/testdata/resolver/redefined_abstract.rb
jez: sure, i've added one.

PR #6239, opened 2022-08-14
Don't mark ENV values as `nil`

### LINKED ISSUE #3589: ENV.to_h has nilable values
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams(x%3A%20Integer).returns(String)%7D%0A%20%20def%20bar(x)%0A%20%20%20%20x%20%3D%20ENV.fetch('foo')%0A%20%20%20%20T.reveal_type(x)%20%23%20this%20isn't%20nilable%2C%20so...%0A%20%20%20%20env%20%3D%20ENV.to_h%0A%20%20%20%20T.reveal_type(env)%20%23%20where%20are%20these%20nilable%20values%3F%0A%20%20%20%20x.to_s%0A%20%20end%0Aend%0A%0Adef%20main%0A%0Aend)

```ruby
# typed: true
class A
  extend T::Sig

  sig {params(x: Integer).returns(String)}
  def bar(x)
    x = ENV.fetch('foo')
    T.reveal_type(x) # this isn't nilable, so...
    env = ENV.to_h
    T.reveal_type(env) # where are these nilable values?
    x.to_s
  end
end

def main

end
```

#### Observed output

```
editor.rb:8: Revealed type: String https://srb.help/7014
     8 |    T.reveal_type(x) # this isn't nilable, so...
            ^^^^^^^^^^^^^^^^
  From:
    editor.rb:7:
     7 |    x = ENV.fetch('foo')
                ^^^^^^^^^^^^^^^^

editor.rb:10: Revealed type: T::Hash[String, T.nilable(String)] https://srb.help/7014
    10 |    T.reveal_type(env) # where are these nilable values?
            ^^^^^^^^^^^^^^^^^^
  From:
    editor.rb:9:
     9 |    env = ENV.to_h
                  ^^^^^^^^
Errors: 2
```

#### Expected behavior

`ENV.to_h` is a `T::Hash[String, T.nilable(String)]`, but it should be a `T::Hash[String, String]`. `ENV` itself behaves as if all values are `String`, it's just the `to_h` that's wrong.

- - -

jez:
### Motivation

Fixes #3589.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/sorbet/sorbet.rbi
+++ b/rbi/sorbet/sorbet.rbi
@@ -116,7 +116,7 @@ end
 class Sorbet::Private::Static::ENVClass
   extend T::Generic
   include Enumerable
-  Elem = type_member(:out) {{fixed: [String, T.nilable(String)]}}
+  Elem = type_member(:out) {{fixed: [String, String]}}
 
   sig do
     params(


jez:

→ https://go/builds/bui_MG0dZHSjHwgWWH
→ https://go/builds/bui_MG0dTkw2HuwbgK

froydnj (changes requested):
This does not seem right according to the docs:

https://docs.ruby-lang.org/en/2.7.0/ENV.html#method-c-5B-5D

It seems like the right fix is to add a different signature for `Sorbet::Private::State::ENVClass#to_h` that would override the one from `Enumerable`?

jez:
@froydnj from the doc you link:

> If value is nil, deletes the environment variable:

that means that `to_h` will never have an environment variable that is set but whose contents are `nil`.

jez:
Note that we still have sigs for `[]` and `[]=` that say "you might get a `nil` back"

froydnj (approved):
Oh, sorry, I see what you mean.  I was not looking closely enough at the patch.

PR #6240, opened 2022-08-14
Call `super` method in `abstract!`

### LINKED ISSUE #2854: `abstract!` in T::Helpers method collides with `abstract!` method in AbstractController::Base
http://api.rubyonrails.org/classes/AbstractController/Base.html#method-c-abstract-21

This is only an issue in Rails applications that defines abstract controllers.

Rails use the `abstract!` method to change the behavior of controllers.

When trying to type abstract controllers including `T::Helpers` the behavior of Rails is lost since the original method is not called.

This is a naming conflict that is likely hard to avoid given the way sorbet works we need to include the `T::Helpers` in the class.

Not sure how to fix this problem without renaming that method to something that has less chance of colliding.

jez:
### Motivation

Fixes #2854

cc @paracycle @rafaelfranca Let me know if this is still something you're
interested in landing.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/helpers.rb
+++ b/gems/sorbet-runtime/lib/types/helpers.rb
@@ -11,6 +11,13 @@ module T::Helpers
   ### Class/Module Helpers ###
 
   def abstract!
+    if defined?(super)
+      # This is to play nicely with Rails' AbstractController::Base,
+      # which also defines an `abstract!` method.
+      # https://api.rubyonrails.org/classes/AbstractController/Base.html#method-c-abstract-21
+      super
+    end
+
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/helpers.rb
```diff
@@ -11,6 +11,13 @@ module T::Helpers
   ### Class/Module Helpers ###
 
   def abstract!
+    if defined?(super)
```
froydnj, line 14: `:ruby-intensifies:`

froydnj (approved):

rafaelfranca:
Yes! Thank you for fixing this.

Issue #6202, opened 2022-08-09
Package-aware definitions and references


jez:
There are a handful of LSP features that would be nice to add to Sorbet as custom extensions:

- [ ] Go to Import Location

  On a constant literal, goes to the `import` location in the `__package.rb` file of the current package. Contrast this with "Go to Definition," which will go to the actual definition of that constant, possibly in another package.

- [ ] Go to Export Location

  On a constant literal, goes to the location where the constant is exported by the current package. If it's not exported, idk where it should go. Either nowhere or to the class definition in the current package?

  Maybe overlaps/is redundant with the new method above?

- [ ] Go to References in this Package

  On an import line in a `__package.rb` file, finds all references of that constant, but scoped to those that appear within the given package.

aadi-stripe:
PR for "Go to References in this Package"
https://github.com/sorbet/sorbet/pull/6242

PR #6242, opened 2022-08-15
LSP - show references in package

aadi-stripe:
Adds functionality to LSP references method to show references within a package.

```
                // Case 1. get-refs on a package declaration
                //   class Foo < PackageSpec
                //         ^^^
                //   Returns all `import Foo` statements, globally
                //
                // Case 2. get-refs on an import statement
                //
                //  class Foo < PackageSpec
                //    ...
                //
                //    import Bar
                //          ^^^
                //  Returns all usages of `Bar` or `Test::Bar` *within* the Foo package only.
                //
                // Case 3. get-refs on an export statement
                //
                //  class Foo < PackageSpec
                //    ...
                //
                //    export Foo::A
                //                ^
                //  Returns all global usages of Foo::A
```

### Motivation

https://github.com/sorbet/sorbet/issues/6202

### Test plan

See included automated tests. Tested in Stripe codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,6 +480,57 @@ different version number, and also marked `default-arg-value`:
 This is due to the translation of defaults into the CFG: there is a synthetic conditional that chooses either to
 initialize the variable from the argument passed at the send, or to the default value when no value is present.
 
+Finding all references works differently in package specification (__package.rb) files. Consider the following:
+
+```ruby
+class Foo < PackageSpec
+  import Bar
+```
+
+Calling "find all references" on `Bar` in this file will return only references to `Bar` in the `Foo` package. LSP tests 
+have access to `import` and `importusage` assertions that you can use to test this functionality.
+
+```ruby
+class Foo < PackageSpec
+  import Bar
+  #      ^^^ import: bar
+```
+
+```ruby
+  class Foo::Baz
+    Bar.new
+ #  ^^^ importusage: bar
+  end
+```
+
+With these annotations, the LSP test will check if "find all references" on `Bar` in `import Bar` statement returns the `Bar.new` usage.
+
+Note that an `import` assertion is dissimilar to a `def` assertion, in that it is in fact a subclass of a `usage` assertion. 
+In this case, the `def` corresponding to an `import` is the PackageSpec declaration of the imported package. Calling "find all references"
+on a PackageSpec declaration will return all imports of the package.
+
+
+```ruby
+class Bar < PackageSpec
+  #   ^^^ def: bar
+  import Bar
+```
+
+```ruby
+class Foo < PackageSpec
+  import Bar
+  #      ^^^ import: bar
+```
+
+```ruby
+class Baz < PackageSpec
+  import Bar
+  #      ^^^ import: bar
+```
+
+With these annotations, the LSP test will check if "find all references" on `Bar` from the `class Bar < PackageSpec`
+declaration returns the declaration itself plus the imports.
+
 #### Testing "Go to Type Definition"
 
 This is somewhat similar to "Find Definition" above, but also slightly different
--- a/main/lsp/LSPTask.cc
+++ b/main/lsp/LSPTask.cc
@@ -54,7 +54,7 @@ ResponseMessageStatus statusForResponse(const ResponseMessage &response) {
                     // textDocument/documentSymbol
                     return ResponseMessageStatus::Unknown;
                 } else if constexpr (is_same_v<T, variant<JSONNullObject, vector<unique_ptr<Location>>>>) {
-                    // textDocument/definition, textDocument/typeDefinition, textDocument/references, and
+                    // textDocument/definition, textDocument/typeDefinition, textDocument/references and
                     // textDocument/implementation
                     if (const auto *locationsPtr = get_if<vector<unique_ptr<Location>>>(&res)) {
                         return locationsPtr->empty() ? ResponseMessageStatus::EmptyResult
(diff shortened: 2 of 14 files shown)


aadi-stripe:

→ https://go/builds/bui_MH7Z5tSj5dDLDe
→ https://go/builds/bui_MH7ZbZAX4vQHHc

froydnj (commented):
I think a proper code review of this will have to wait until next week, but at first glance it seems fairly reasonable.

I think the high-order bit about the design is that, AIUI, #6202 is proposing custom Sorbet LSP requests (e.g. `sorbet/showSymbol`) to implement this functionality, whereas this PR makes `textDocument/references` work slightly differently on `__package.rb` files.

The design here is convenient, but it does mean that we are leaking implementation details of Stripe packages into LSP when `--stripe-packages` is not active.  I'm not sure whether this is the right thing to do.  It also changes the behavior of find all references, which might be surprising to some people -- though it's also more discoverable this way than putting it behind a context menu.

I also think it's worth thinking about whether we could make this work if your cursor was on the `import` token itself rather than just the constant.

aadi-stripe:
> I think a proper code review of this will have to wait until next week, but at first glance it seems fairly reasonable.
> 
> I think the high-order bit about the design is that, AIUI, #6202 is proposing custom Sorbet LSP requests (e.g. `sorbet/showSymbol`) to implement this functionality, whereas this PR makes `textDocument/references` work slightly differently on `__package.rb` files.
> 
> The design here is convenient, but it does mean that we are leaking implementation details of Stripe packages into LSP when `--stripe-packages` is not active. I'm not sure whether this is the right thing to do. It also changes the behavior of find all references, which might be surprising to some people -- though it's also more discoverable this way than putting it behind a context menu.
> 
> I also think it's worth thinking about whether we could make this work if your cursor was on the `import` token itself rather than just the constant.

Thanks for the comments. A few thoughts on custom method vs piggybacking on `references` -- fwiw it was an option I aligned on with @jez in the RP channel in Slack.

1. The old `--stripe-packages` design had generally this same behavior, and hence did also leak implementation details into LSP. Personally since `__package.rb` files are special and purely a feature of `--stripe-packages`, I don't think it's too bad for LSP to work differently in those files.
2. The existing get-refs behavior in package files is actually not very meaningful or useful. On doing a get-ref on imports, it returns other import statements of the constant in other packages.
3. As you pointed out, a custom method in this case would be less ergonomic to users, since the existing behavior is not very useful, and to get the useful behavior they would have to call a different command.
4. We could document the difference in behavior so that people are aware that package files are treated differently by design.

Agree that it would be worth doing this on `import` statements. But personally I think that that should be an additional feature to this, not in lieu of.

froydnj:
Thanks for the context!  Reassigning since it sounds like jez has more context here than I do.

main/lsp/LSPTask.cc
```diff
@@ -54,7 +54,7 @@ ResponseMessageStatus statusForResponse(const ResponseMessage &response) {
                     // textDocument/documentSymbol
                     return ResponseMessageStatus::Unknown;
                 } else if constexpr (is_same_v<T, variant<JSONNullObject, vector<unique_ptr<Location>>>>) {
-                    // textDocument/definition, textDocument/typeDefinition, textDocument/references, and
+                    // textDocument/definition, textDocument/typeDefinition, textDocument/references and
```
jez, line 57: You don't like Oxford commas?

aadi-stripe: Complete accident, but come to think of it I'm not a fan lol. 

Will revert.

aadi-stripe:
@jez ptal

README.md
```diff
@@ -480,6 +480,57 @@ different version number, and also marked `default-arg-value`:
...
 initialize the variable from the argument passed at the send, or to the default value when no value is present.
 
+Finding all references works differently in package specification (__package.rb) files. Consider the following:
+
+```ruby
+class Foo < PackageSpec
+  import Bar
+```
+
+Calling "find all references" on `Bar` in this file will return only references to `Bar` in the `Foo` package. LSP tests
```
jez, line 490: What does this mean Go to Definition will report for `import Bar` here?

aadi-stripe: Go to definition on `import Bar` will go to the declaration of the `Bar` package -- that behavior doesn't change.

README.md
```diff
@@ -480,6 +480,57 @@ different version number, and also marked `default-arg-value`:
...
+  class Foo::Baz
+    Bar.new
+ #  ^^^ importusage: bar
+  end
+```
+
+With these annotations, the LSP test will check if "find all references" on `Bar` in `import Bar` statement returns the `Bar.new` usage.
+
+Note that an `import` assertion is dissimilar to a `def` assertion, in that it is in fact a subclass of a `usage` assertion. 
+In this case, the `def` corresponding to an `import` is the PackageSpec declaration of the imported package. Calling "find all references"
```
jez, line 509: So upon arriving at the `class Bar < PackageSpec` location, clicking on go-to-def for the `class Bar` takes you right there, is that correct?

IIRC, `class Bar < PackageSpec` is secretly something like `class <PkgRegistry>::Bar < PackageSpec` or something like that. How would I ever jump from an `import Bar` to the real `Bar` class?

aadi-stripe: Don't think we have a way currently to do that. I think that stopped working after we switched to the visibility checker design.

aadi-stripe: This PR improves the story a little, but does not solve that problem.

aadi-stripe: But we can totally fix that in a separate change

jez (approved):

PR #6243, opened 2022-08-15
Add installation instructions for watchman

### LINKED ISSUE #1758: User-visible message when watchman is missing
### Problem

When `watchman` is not installed, both the Stripe VS Code Sorbet extension and LanguageClient-neovim will attempt to launch `sorbet --lsp`, which will exit because watchman is not installed:

```
❯ sorbet --lsp .
Error running Watchman (with `watchman -j -p--no-pretty`).
Watchman is required for Sorbet to detect changes to files made outside of your code editor.
Don't need Watchman? Run Sorbet with `--disable-watchman`.
```

The observed behavior (with `watchman` uninstalled):

- In VS Code, our plugin attempts to restart Sorbet 10 times, and then gives up saying "Unable to Start Sorbet after 10 attempts. Sorbet is disabled until you decide to restart it."
- In LanguageClient-neovim, `Language server ruby exited unexpectedly: failed to fill whole buffer` which is just what it prints any time the LSP process exits

### Proposed solution

We should be sure to surface this to the user in a visible error message before publicly releasing the IDE features.

jez:
### Motivation

Fixes #1758.

(#1758 specifically requests a user-visible message. We actually already added
that at one point, but our docs never make any mention of watchman. I figure we
may as well add those docs and then close the ticket.)

### Test plan

docs-only change

### DIFF
--- a/website/docs/vscode.md
+++ b/website/docs/vscode.md
@@ -51,6 +51,29 @@ latest edits:
 
 ![](/img/lsp/typechecking.gif)
 
+For the best experience, Sorbet requires
+[Watchman](https://facebook.github.io/watchman/), which listens for changes to
+the files on disk in addition to edits that happen to files open in the editor.
+For example, without Watchman installed, Sorbet will not detect when files have
+changed on disk due to things like changing the currently checked out branch.
+
+Sorbet simply requires that the `watchman` binary be somewhere visible on the
+`PATH` environment variable. There are installation instructions for various
+platforms in the
+[Watchman docs](https://facebook.github.io/watchman/docs/install.html).
+
+If you cannot install `watchman` to somewhere on the `PATH`, you can use the
+`--watchman-path=...` command line flag to `srb tc` to specify a path to the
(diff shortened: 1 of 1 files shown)


website/docs/vscode.md
```diff
@@ -51,6 +51,29 @@ latest edits:
...
+platforms in the
+[Watchman docs](https://facebook.github.io/watchman/docs/install.html).
+
+If you cannot install `watchman` to somewhere on the `PATH`, you can use the
+`--watchman-path=...` command line flag to `srb tc` to specify a path to the
+`watchman` binary.
+
+If you cannot install `watchman` at all, pass the `--disable-watchman` flag to
+`srb tc`. This will mean that Sorbet only reads the files from disk at startup,
+and afterwords only ever sees contents of files that have been opened or changed
```
froydnj, line 71:
```suggestion
and afterwards only ever sees contents of files that have been opened or changed
```

froydnj (approved):

PR #6246, opened 2022-08-16
patch proc.c to not allocate in call validation

penelope-stripe:
### Motivation
Today this line: https://github.com/sorbet/sorbet/blob/2b2a81bbb11c6ba4cb7cde2239bdf785a71035c8/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb#L157 is responsible for lots of allocations in pay-server. The patch to the ruby interpreter in this branch removes this allocation behind a new method `bind_call_experimental_optimized` (this is so that we can roll the behaviour out slowly in pay-server). The patch was reviewed by tenderlove and is already upstreamed in ruby proper: https://github.com/ruby/ruby/pull/6244. Aaron told me they won't backport it to 2.7 but my patch is safe to include in 2.7 as is in this PR.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/unbound_method.rbi
+++ b/rbi/core/unbound_method.rbi
@@ -140,10 +140,6 @@ class UnboundMethod
   sig {params(recv: BasicObject, args: T.untyped, blk: T.untyped).returns(T.untyped)}
   def bind_call(recv, *args, &blk); end
 
-  # for our experimental procc patch
-  sig {params(recv: BasicObject, args: T.untyped, blk: T.untyped).returns(T.untyped)}
-  def bind_call_experimental_optimized(recv, *args, &blk); end
-
   # Returns a clone of this method.
   #
   # ```ruby
--- a/rbi/core/unbound_method.rbi
+++ b/rbi/core/unbound_method.rbi
@@ -140,6 +140,10 @@ class UnboundMethod
   sig {params(recv: BasicObject, args: T.untyped, blk: T.untyped).returns(T.untyped)}
   def bind_call(recv, *args, &blk); end
 
+  # for our experimental procc patch
+  sig {params(recv: BasicObject, args: T.untyped, blk: T.untyped).returns(T.untyped)}
+  def bind_call_experimental_optimized(recv, *args, &blk); end
+
   # Returns a clone of this method.
   #
   # ```ruby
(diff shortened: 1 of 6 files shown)


froydnj:
Feel free to assign whenever this is ready.

third_party/ruby/penelope_procc.patch
```diff
@@ -0,0 +1,97 @@
...
++	me = rb_method_entry_clone(data->me);
++    } else {
++	me = data->me;
++    }
++   
+ 
+     if (RB_TYPE_P(me->owner, T_MODULE)) {
++	if (clone == ME_CLONE_DO_NOT_CLONE) {
++	    me = rb_method_entry_clone(data->me);
++	}
```
jez, line 41: Why does this check that `clone` is "do not clone" but then immediately call `*_clone`?

rbi/core/unbound_method.rbi
```diff
@@ -140,6 +140,10 @@ class UnboundMethod
   sig {params(recv: BasicObject, args: T.untyped, blk: T.untyped).returns(T.untyped)}
   def bind_call(recv, *args, &blk); end
 
+  # for our experimental procc patch
```
jez, line 143: Should we put an RBI for this in Stripe's codebase, instead of the payload that we ship to all users of Sorbet?

This will mean that `umethod.bind_cal|` will show `bind_call_experimental_optimized` in the completion results even for non-Stripe users.

third_party/ruby/penelope_procc.patch
penelope-stripe: in this branch, we want to always, unconditionally clone the method entry, because things lower down manipulate me

rbi/core/unbound_method.rbi
penelope-stripe: 👍

third_party/ruby/penelope_procc.patch
penelope-stripe: this maps to this stanza in my ruby pr:https://github.com/ruby/ruby/pull/6244/files#diff-b2553d23e6b1fe76e20608d06c25f6acca06279100f1a9c24febcd79a82fac3cR2576-R2580

jez (approved):

Issue #6244, opened 2022-08-15
Inlay hints


qaisjp:
### Problem

> Inlay hints add additional inline information to source code to help you understand what the code does.

vscode-go supports inlay hints, it looks like this ([source](https://github.com/golang/vscode-go/issues/1631))

![image](https://user-images.githubusercontent.com/923242/184702879-e5040eb9-a538-40ab-888a-801997ca7cca.png)

TypeScript inlay hints look like this ([source with more images](https://code.visualstudio.com/docs/typescript/typescript-editing#_inlay-hints))

![image](https://user-images.githubusercontent.com/923242/184703862-a1839e96-8e40-4e06-8bf2-1c82b3343b07.png)

### Proposed solution

* [ ] Implement the part of the LSP described in https://github.com/microsoft/language-server-protocol/issues/956.

  > VS Code has shipped with a [finalized native API](https://github.com/microsoft/vscode/blob/778b74a2159983ba09568d069ef9c9ecd858f783/src/vscode-dts/vscode.d.ts#L4570-L4732) for this.

* [ ] Provide a granular choice of inlay hints (similar settings to vscode-go / typescript): search for `inlay` in VS Code's Settings UI

froydnj:
Do you know if we are required to implement `InlayHintsProvider` in the extension?  I have [a branch](https://github.com/sorbet/sorbet/commits/froydnj-inlay-hints) that implements what I thought was most of the spec, but I do not see VSCode requesting inlay hints in the logs.

qaisjp:
> Do you know if we are required to implement `InlayHintsProvider` in the extension?

I don't think so.

* I skimmed through all the changelists linked by @gopherbot on https://github.com/golang/vscode-go/issues/1631, and I'm not seeing any signifiant TypeScript changes in their extension.
* I also searched in their vscode-go repo, and found the same changes I found in the changelists.
* Their TypeScript changes are only propagating their custom extension settings to the gopls (LSP) config.

damolina-stripe:
@froydnj  - InlayHints was shipped in the [LSP 3.17](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint), so after https://github.com/sorbet/sorbet/pull/8606 this work could be revisited.

PR #6247, opened 2022-08-16
implement inlay hints in LSP

froydnj:
### Motivation

#6244 
### Test plan

I do not have good ideas on how to write testdata tests for these, so we may need to resort to protocol tests.


froydnj:

→ https://go/builds/bui_MGGiXCPczXuOmO
→ https://go/builds/bui_MGGiQdyEovvlAB

Velrok:
Looks like this has gone quiet.
NeoVim now support inlay hints and seeing support for this in the sorbet LSP would be awesome.

juharris:
Looking into inlay type hints again in https://github.com/sorbet/sorbet/pull/9869

PR #6248, opened 2022-08-16
move cast creation to rewriter

froydnj:
### Motivation

This change helps with putting changes to instance and class variables on the fast path, as after this PR, namer can clearly see where declarations of such variables occur, rather than having to re-implement the logic that resolver currently has.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/ClassNew.cc
+++ b/rewriter/ClassNew.cc
@@ -12,16 +12,12 @@ namespace sorbet::rewriter {
 
 // Is this expression a synthetic `T.bind` call we added from the `ClassNew` rewriter on sends?
 bool isRewrittenBind(ast::ExpressionPtr &expr) {
-    auto send = ast::cast_tree<ast::Send>(expr);
-    if (send == nullptr) {
+    auto cast = ast::cast_tree<ast::Cast>(expr);
+    if (cast == nullptr) {
         return false;
     }
 
-    if (send->fun != core::Names::bind()) {
-        return false;
-    }
-
-    return send->flags.isRewriterSynthesized;
+    return cast->cast == core::Names::syntheticBind();
 }
 
 vector<ast::ExpressionPtr> ClassNew::run(core::MutableContext ctx, ast::Assign *asgn) {
@@ -135,10 +131,7 @@ bool ClassNew::run(core::MutableContext ctx, ast::Send *send) {
         type = ast::MK::ClassOf(send->loc, std::move(target));
     }
 
-    auto bind = ast::MK::Bind(send->loc, ast::MK::Self(send->loc), std::move(type));
-
-    // Mark the bind as synthetic so we can spot it from the `ClassNew` rewriter on assigns and remove it from the tree.
-    ast::cast_tree<ast::Send>(bind)->flags.isRewriterSynthesized = true;
+    auto bind = ast::MK::SyntheticBind(send->loc, ast::MK::Self(send->loc), std::move(type));
 
     ast::InsSeq::STATS_store blockStats;
     blockStats.emplace_back(std::move(bind));
--- a/test/cli/autogen_print_rbi/test.out
+++ b/test/cli/autogen_print_rbi/test.out
@@ -109,13 +109,6 @@ requires: []
  is_defining_ref=0
 [ref id=2]
  scope=[FromRBI]
- name=[T]
- nesting=[[FromRBI]]
- resolved=[T]
- loc=test/cli/autogen_print_rbi/b.rbi:4
- is_defining_ref=0
-[ref id=3]
- scope=[FromRBI]
  name=[Integer]
  nesting=[[FromRBI]]
  resolved=[Integer]
@@ -123,9 +116,9 @@ requires: []
  is_defining_ref=0
 
 f3b9f8c31a343a60665c06f959a48dbdceef757c  autogen3.msgpack
-a20ed01b8a7646141440ba39b60b247bd13a4388  autogen4.msgpack
+2305b0d8f2c81fbb946e4f1dbec6230386925456  autogen4.msgpack
 
-39a40,80
+39a40,73
 > # ParsedFile: test/cli/autogen_print_rbi/b.rbi
 > requires: []
 > ## defs:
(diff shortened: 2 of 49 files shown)


froydnj:

→ https://go/builds/bui_MG1l3DsIgtvSGe
→ https://go/builds/bui_MG1lE0md1HNH80

froydnj:

→ https://go/builds/bui_MG2cA1iMj34tf1
→ https://go/builds/bui_MG2d1lMqn4TZ0O

jez:
This has a slightly annoying change in behavior, right?

We previously gated `T.let` checking on whether `T` was actually a resolved `ConstantLit` to `core::Symbols::T()`, but now we're doing it on unresolved constants.

froydnj:
> This has a slightly annoying change in behavior, right?
> 
> We previously gated `T.let` checking on whether `T` was actually a resolved `ConstantLit` to `core::Symbols::T()`, but now we're doing it on unresolved constants.

This is correct.  We try to gate things by looking for the receiver to be `T` just like we do for a number of other syntactic things (e.g. `T::Struct`), but this is of course not perfect.

rewriter/ClassNew.cc
```diff
@@ -12,16 +12,12 @@ namespace sorbet::rewriter {
 
 // Is this expression a synthetic `T.bind` call we added from the `ClassNew` rewriter on sends?
 bool isRewrittenBind(ast::ExpressionPtr &expr) {
-    auto send = ast::cast_tree<ast::Send>(expr);
-    if (send == nullptr) {
```
jez, line 16: I'm slightly worried about these sorts of changes to `Rewriter`, because it makes individual rewriter passes sensitive to whether they've run before or after the `TypeAssertion` rewriter. This sort of thing has caused bugs in the past, as people handle one case but forget to handle another case.

I guess you've chosen to do it here instead of in `Namer::findSymbols` to avoid having to make `findSymbols` be a non-shallow map?

Another alternative might be to lurk this in `Cleanup.cc` so that it can run after all the rewriter passes.

jez: Another alternative would be to do this in `Desugar.cc`, so that *all* the rewriter passes can assume that it will already have run, which should allow rewriter passes to never have to do their own `T.let` parsing, regardless of the order of rewriter passes.

test/cli/autogen_print_rbi/test.out
```diff
@@ -108,13 +108,6 @@ requires: []
  loc=test/cli/autogen_print_rbi/b.rbi:4
  is_defining_ref=0
 [ref id=2]
- scope=[FromRBI]
- name=[T]
- nesting=[[FromRBI]]
- resolved=[T]
- loc=test/cli/autogen_print_rbi/b.rbi:4
- is_defining_ref=0
-[ref id=3]
```
jez, line 117: Can we check with someone who works on autogen whether this is going to cause problems for any of their tools? cc @ngroman

jez (commented):
I think I'm pretty ok with the resolved `T` → unresolved `T` change. If we don't have any tests that fail because of that, we should add some.

For example, I know of at least one `T` but not `::T` module:

https://github.com/SciRuby/distribution/blob/master/lib/distribution/t.rb

`::Distribution::T` (for the student's T statistical distribution). We have had problems in the past from this gem because we would generate RBIs that simply wrote

```ruby
module Distribution
  module T
    SHORTHAND = T.let(T.unsafe(nil), String)
  end
end
```

into the hidden definitions RBI, and we had to change it to use `::T.let` instead.

I think that problems like this go away, at the expense of not being able to leave things alone if a gem had defined some custom `T.let` method.

But we already have the same problem with `sig` syntax, so I'm kind of fine adopting the same approach for `T.let`.

I do wonder if doing it as a plain rewriter pass is best, or whether we should handle this somewhere entirely before or entirely after rewriter, to make it less likely that a rewriter's correctness depends on whether it happens to come before or after the TypeAssertion rewriter.

rewriter/ClassNew.cc
froydnj: I can believe that ordering issues have caused problems in the past, but I'm not imaginative enough to see why this particular change would cause problems: the rewriter here is guaranteed to be run before all of the rewriters on class bodies, which means the latter will always see `Cast` nodes.  It is really nice that the class body rewriters can depend on this, because then they don't need custom parsing logic (cf. the minitest changes).

All of the other `Send`-based rewriters are more or less insensitive to the ordering issues you mention?  (I could see problems happening if some things were `preTransform*`, so perhaps we could have future problems...)  Even if other `Send`-based rewriters are examining their arguments, those arguments have already been run through the cast rewriter.

Moving things into desugar could be a reasonable thing to do, though I'm not excited about adding even more logic to desugaring sends.

froydnj:
> I think I'm pretty ok with the resolved `T` → unresolved `T` change. If we don't have any tests that fail because of that, we should add some.
> 
> For example, I know of at least one `T` but not `::T` module:
> 
> https://github.com/SciRuby/distribution/blob/master/lib/distribution/t.rb
> 
> `::Distribution::T` (for the student's T statistical distribution). We have had problems in the past from this gem because we would generate RBIs that simply wrote
> 
> ```ruby
> module Distribution
>   module T
>     SHORTHAND = T.let(T.unsafe(nil), String)
>   end
> end
> ```
> 
> into the hidden definitions RBI, and we had to change it to use `::T.let` instead.
> 
> I think that problems like this go away, at the expense of not being able to leave things alone if a gem had defined some custom `T.let` method.

I'm unclear on what this last sentence is saying, and what sort of tests are being requested.  Do you mean that you want some test with a "bare" `T.let` that got resolved to a non-Sorbet `T` previously (and therefore typechecked), and thus needs to be annotated with errors as a result of this change?

I guess the `T.let`s generated for the single-package RBI generator should probably use `::T.let` now.

rewriter/ClassNew.cc
jez: Sounds good, we can leave it here for now and if we need to move it in the future we can

jez (approved):

test/cli/autogen_print_rbi/test.out
nroman-stripe: I'm not worried about this. The tooling that depends on this only is concerned with non-gem code.

PR #6251, opened 2022-08-16
ensure that `llvm-symbolizer` is available for at least exp tests

froydnj:
### Motivation

It is zero fun to be faced with ASAN errors in CI (or locally!) that are an inscrutable mess of addresses.

There is probably a more bazel-y way to do this; this version cargo-cults from the existing support that we have for `llvm-diff`.

### Test plan

See included automated tests.  I should probably verify that this actually works in CI, but it does symbolize things for me locally.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -1,6 +1,7 @@
 load("//tools:clang.bzl", "clang_tool")  # todo: this should be decoupled and use the library toolchain, not the compiler one
 
 clang_tool("llvm-diff")  # depended on by compiler_tests
+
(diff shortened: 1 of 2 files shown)


jez (commented):
beautiful thanks!

jez (approved):

froydnj:
Verified that stack traces show up in exp tests: https://buildkite.com/sorbet/sorbet/builds/24193#0182a7fc-8c99-4261-945d-a0dc80bb2e28

Also, we have symbols so big that the symbolizer's internal buffer overflows 🤣

PR #1342, opened 2019-07-19
Don't serialize silenced errors

jvilk-stripe:
Don't serialize silenced errors. Speeds up Sorbet on large codebases.

### Motivation

Error serialization was around ~7-8% of typechecking time before this change because we eagerly serialized all errors -- including those that are suppressed because they are for a stricter typedness level!

Now it's 0% of typechecking time. End-to-end typechecking on pay-server went from ~14.53 seconds to ~10.21 seconds (averages of 5 runs, after a few runs of warmup)!

### Test plan

N/A

PR #6252, opened 2022-08-16
Only call `toString` in the stdout flusher

jez:
### Motivation

Previously, we were calling `toString` before pushing the the error message
into the queue, which mean that even flushers that did not using the
`toString` version of the error were eagerly computing it.

This might have a slight performance improvement? But probably not noticeable,
unless you're seeing tons and tons of errors in the IDE.

Incidentally, this will fix most of the crash logs we're seeing for `addLocLine`
by simply not exercising that code path. I still am curious to see if the logs I
just added are going to be enough to show me where the problem is, so I'll hold
off on merging this for the moment while I wait for those logs to show up.

### Test plan

See included automated tests.

### DIFF
--- a/core/ErrorFlusherStdout.cc
+++ b/core/ErrorFlusherStdout.cc
@@ -21,8 +21,8 @@ void ErrorFlusherStdout::flushErrors(spdlog::logger &logger, const GlobalState &
             if (out.size() != 0) {
                 fmt::format_to(std::back_inserter(out), "\n\n");
             }
-            ENFORCE(error->text.has_value());
-            fmt::format_to(std::back_inserter(out), "{}", error->text.value_or(""));
+            auto text = error->error->toString(gs);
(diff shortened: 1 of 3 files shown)


froydnj (approved):
I'm not sure why we ever did this weird not-quite-lazy error reporting.

jez:
As a curious note for posterity, it looks like the last time this code was changed, it was being changed in service of a huge speedup 😅 #1342

jez:
> I'm not sure why we ever did this weird not-quite-lazy error reporting.

There is maybe an argument that maybe bugs could creep in if you report the error using one GlobalState, then mutate it, and use another GlobalState to flush the error? (That's an argument in favor of the current approach, where the `toString(gs)` is called as close to when the `ErrorBuilder` destructs as possible. Not sure that it's a good reason, but ¯\_(ツ)_/¯)

PR #6254, opened 2022-08-17
use a shallow walk for document symbols

froydnj:
### Motivation

The walker here only cares about class defs and method defs, so we shouldn't need to descend into method defs themselves.  (We should be able to rely on rewriter-based flattening here, so nested method defs should not be a problem, but we are not relying on class flattening being done.)

### Test plan

See included automated tests.


froydnj:
Added a test with nested method definitions and verified it passes before and after the shallow walk change.

jez (approved):

PR #6258, opened 2022-08-17
Don't crash when `private` precedes `sig`

jez:
### Motivation

Fixes a crash we were seeing at Stripe.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Flatten.cc
+++ b/rewriter/Flatten.cc
@@ -341,7 +341,7 @@ public:
         // if we get a MethodData back, then we need to move this and replace it with an EmptyTree
         if (auto md = methods.popScope()) {
             methods.addExpr(*md, move(tree));
-            tree = ast::MK::EmptyTree();
+            tree = ast::MK::Nil(send.loc);
             return;
         }
     }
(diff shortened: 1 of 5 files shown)


froydnj (approved):

rewriter/Flatten.cc
```diff
@@ -341,7 +341,7 @@ class FlattenWalk {
         // if we get a MethodData back, then we need to move this and replace it with an EmptyTree
         if (auto md = methods.popScope()) {
             methods.addExpr(*md, move(tree));
-            tree = ast::MK::EmptyTree();
+            tree = ast::MK::Nil(send.loc);
```
froydnj, line 344: Is the rationale here that we need to replace it with something the user could have actually written, so that we don't see `private(<empty-tree>)`?

Probably want to update the comment right before the `if`, maybe also adding an explanation of why we can't use `EmptyTree` here.

froydnj (approved):
You may also want to update the comment above `CleanupWalk` to explain why we're removing `nil` as well -- and only `nil`, even, since theoretically we should be able to remove all literals in non-value positions

PR #6264, opened 2022-08-18
Change confusing phrasing in tstruct.md

jez:
### Test plan

See included automated tests.

### DIFF
--- a/website/docs/tstruct.md
+++ b/website/docs/tstruct.md
@@ -124,7 +124,7 @@ class ChildTwo < T::Struct
 end
 ```
 
-If it is imperative that inheritance must be used (not composition, either):
+If the code absolutely must use inheritance and cannot use composition, either:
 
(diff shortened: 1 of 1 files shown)


froydnj (approved):

website/docs/tstruct.md
```diff
@@ -124,7 +124,7 @@ class ChildTwo < T::Struct
 end
 ```
 
-If it is imperative that inheritance must be used (not composition, either):
+If the code absolutely must use inheritance and cannot use composition, either:
```
paracycle, line 127: IMO, adding a "you should" or "you could" before "either" would make it much cleaner.
```suggestion
If the code absolutely must use inheritance and cannot use composition, you could either:
```

jez: we generally avoid writing "you" in documentation, but thanks for the suggestion

PR #6265, opened 2022-08-18
create `FoundDefinitions` for fields and enter them in namer

froydnj:
### Motivation

Making changes to fields (eventually) take the fast path in LSP.

### Test plan

See included automated tests.

### DIFF
--- a/core/FoundDefinitions.h
+++ b/core/FoundDefinitions.h
@@ -18,6 +18,7 @@ struct FoundClass;
 struct FoundStaticField;
 struct FoundTypeMember;
 struct FoundMethod;
+struct FoundField;
 
 class FoundDefinitionRef final {
 public:
@@ -29,6 +30,7 @@ public:
         StaticField = 4,
         TypeMember = 5,
         Symbol = 6,
+        Field = 7,
     };
     CheckSize(Kind, 1, 1);
 
@@ -76,6 +78,9 @@ public:
 
     FoundTypeMember &typeMember(FoundDefinitions &foundDefs);
     const FoundTypeMember &typeMember(const FoundDefinitions &foundDefs) const;
+
+    FoundField &field(FoundDefinitions &foundDefs);
+    const FoundField &field(const FoundDefinitions &foundDefs) const;
 
     core::SymbolRef symbol() const;
 
@@ -184,6 +189,20 @@ struct FoundModifier {
 };
 CheckSize(FoundModifier, 24, 4);
 
+struct FoundField {
+    enum class Kind : uint8_t {
+        Class,
+        Instance,
+    };
+    Kind kind;
+    FoundDefinitionRef owner;
+    core::LocOffsets loc;
+    core::NameRef name;
+
+    std::string toString(const core::GlobalState &gs, const FoundDefinitions &foundDefs, uint32_t id) const;
+};
+CheckSize(FoundField, 20, 4);
+
 class FoundDefinitions final {
     // Contains references to items in _klasses, _staticFields, and _typeMembers.
     // Used to determine the order in which symbols are defined in SymbolDefiner.
@@ -202,12 +221,15 @@ class FoundDefinitions final {
     std::vector<FoundTypeMember> _typeMembers;
     // Contains all method and class modifiers (e.g. private/public/protected).
     std::vector<FoundModifier> _modifiers;
+    // Contains all class and instance variables defined in the file.
+    std::vector<FoundField> _fields;
 
     FoundDefinitionRef addDefinition(FoundDefinitionRef ref) {
         DEBUG_ONLY(switch (ref.kind()) {
             case FoundDefinitionRef::Kind::Class:
             case FoundDefinitionRef::Kind::StaticField:
             case FoundDefinitionRef::Kind::TypeMember:
+            case FoundDefinitionRef::Kind::Field:
                 break;
             case FoundDefinitionRef::Kind::Method:
             case FoundDefinitionRef::Kind::ClassRef:
@@ -253,6 +275,12 @@ public:
         return addDefinition(FoundDefinitionRef(FoundDefinitionRef::Kind::TypeMember, idx));
     }
 
+    FoundDefinitionRef addField(FoundField &&field) {
+        const uint32_t idx = _fields.size();
+        _fields.emplace_back(std::move(field));
+        return addDefinition(FoundDefinitionRef(FoundDefinitionRef::Kind::Field, idx));
+    }
+
     FoundDefinitionRef addSymbol(core::SymbolRef symbol) {
         return FoundDefinitionRef(FoundDefinitionRef::Kind::Symbol, symbol.rawId());
     }
@@ -279,6 +307,11 @@ public:
     // See documentation on _modifiers
     const std::vector<FoundModifier> &modifiers() const {
         return _modifiers;
+    }
+
+    // See documentation on _fields
+    const std::vector<FoundField> &fields() const {
+        return _fields;
     }
 
     friend FoundDefinitionRef;
--- a/core/FoundDefinitions.h
+++ b/core/FoundDefinitions.h
@@ -259,8 +259,10 @@ public:
         return FoundDefinitionRef(FoundDefinitionRef::Kind::ClassRef, idx);
     }
 
-    void addMethod(FoundMethod &&method) {
+    FoundDefinitionRef addMethod(FoundMethod &&method) {
+        const uint32_t idx = _methods.size();
         _methods.emplace_back(std::move(method));
+        return FoundDefinitionRef(FoundDefinitionRef::Kind::Method, idx);
     }
 
     FoundDefinitionRef addStaticField(FoundStaticField &&staticField) {
--- a/core/FoundDefinitions.h
+++ b/core/FoundDefinitions.h
@@ -191,8 +191,8 @@ CheckSize(FoundModifier, 24, 4);
 
 struct FoundField {
     enum class Kind : uint8_t {
-        Class,
-        Instance,
+        ClassVariable,
+        InstanceVariable,
     };
     Kind kind;
     FoundDefinitionRef owner;
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -502,8 +502,8 @@ public:
         }
 
         core::FoundField found;
-        found.kind = uid->kind == ast::UnresolvedIdent::Kind::Instance ? core::FoundField::Kind::Instance
-                                                                       : core::FoundField::Kind::Class;
+        found.kind = uid->kind == ast::UnresolvedIdent::Kind::Instance ? core::FoundField::Kind::InstanceVariable
+                                                                       : core::FoundField::Kind::ClassVariable;
         found.owner = getOwner();
         found.loc = uid->loc;
         found.name = uid->name;
@@ -1136,7 +1136,7 @@ class SymbolDefiner {
 
         // resolver checks a whole bunch of various error conditions here; we just want to
         // know where to define the field.
-        if (field.kind == core::FoundField::Kind::Class) {
+        if (field.kind == core::FoundField::Kind::ClassVariable) {
             scope = ctx.owner.enclosingClass(ctx);
         } else {
             scope = ctx.selfClass();
@@ -1157,7 +1157,7 @@ class SymbolDefiner {
         }
 
         core::FieldRef sym;
-        if (field.kind == core::FoundField::Kind::Instance) {
+        if (field.kind == core::FoundField::Kind::InstanceVariable) {
             sym = ctx.state.enterFieldSymbol(ctx.locAt(field.loc), scope, field.name);
         } else {
             sym = ctx.state.enterStaticFieldSymbol(ctx.locAt(field.loc), scope, field.name);
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -496,10 +496,9 @@ public:
             return false;
         }
 
-        // This is an error in resolver, but just ignore this here.
-        if (cast->cast != core::Names::let()) {
-            return false;
-        }
+        // Resolver will issues errors about non-let declarations of variables, but
+        // will still associate types with the variables; we need to ensure that
+        // there are entries in the symbol table for those variables.
 
         core::FoundField found;
         found.kind = uid->kind == ast::UnresolvedIdent::Kind::Instance ? core::FoundField::Kind::InstanceVariable
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -478,8 +478,42 @@ public:
         return foundRef;
     }
 
+    // Returns `true` if `asgn` is a field declaration.
+    bool handleFieldDeclaration(core::Context ctx, ast::Assign &asgn) {
+        auto *uid = ast::cast_tree<ast::UnresolvedIdent>(asgn.lhs);
+        if (uid == nullptr) {
+            return false;
+        }
+
+        if (uid->kind != ast::UnresolvedIdent::Kind::Instance && uid->kind != ast::UnresolvedIdent::Kind::Class) {
+            return false;
+        }
+
+        auto *cast = ast::cast_tree<ast::Cast>(asgn.rhs);
+        if (cast == nullptr) {
+            return false;
+        }
+
+        // This is an error in resolver, but just ignore this here.
+        if (cast->cast != core::Names::let()) {
+            return false;
+        }
+
+        core::FoundField found;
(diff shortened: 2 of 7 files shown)


core/FoundDefinitions.h
```diff
@@ -184,6 +189,20 @@ struct FoundModifier {
 };
 CheckSize(FoundModifier, 24, 4);
 
+struct FoundField {
+    enum class Kind : uint8_t {
+        Class,
+        Instance,
```
jez, line 195: Minor, but on first read I didn't realize this was related to `UnresolvedIdent::Kind::{Class,Instance}`, I thought it was like the `isSelfMethod` flag for methods.

I wonder if we should call this `ClassVariable` / `InstanceVariable`?

namer/namer.cc
```diff
@@ -1049,6 +1094,44 @@ class SymbolDefiner {
         }
     }
 
+    core::FieldRef insertField(core::MutableContext ctx, const core::FoundField &field) {
```
jez, line 1097: Does this method actually ever get called? It looks like this is only called from `defineNonDeletableSingle`, but the only place where we call `defineNonDeletableSingle` has a check that says `ref.kind() != core::FoundDefinitionRef::Kind::Field`.

Is this PR a no-op?

Should we also make sure that we delete the logic from resolver that calls `enterFieldSymbol`/`enterStaticFieldSymbol` so that we can be sure it's being powered by this change alone?

namer/namer.cc
```diff
@@ -1049,6 +1094,44 @@ class SymbolDefiner {
...
+            auto prior = existing.asFieldRef();
+            // There was a previous declaration of this field in this file.  Ignore it;
+            // let resolver deal with issuing the appropriate errors.
+            if (prior.data(ctx)->resultType == core::Types::todo()) {
+                return prior;
+            }
+
+            // We are on the fast path and there was a previous declaration.
+            // TODO(froydnj) apparently incremental resolve can't add new symbols?!
+            // ctx.state.mangleRenameSymbol(existing, existing.data(ctx)->name);
```
jez, line 1119: Consider the case of changing the type of a method on the fast path: we don't mangle rename the symbol, we simply update its type. It should be the same for instance variables unless I misunderstand.

(The ENFORCE that we haven't added new symbols after incremental resolve is something that I think makes sense to keep.)

namer/namer.cc
```diff
@@ -1049,6 +1094,44 @@ class SymbolDefiner {
...
+            prior.data(ctx)->resultType = core::Types::todo();
```
jez, line 1120: Out of curiosity, did you play around with setting/not setting `resultType` to `todo()` in this case?

For example I don't think that we reset a method's type to some initial value when we discover a pre-existing method on the fast path.

I don't have a strong opinion on it either way (especially because actual changes to the types of fields currently takes the slow path anyways) but I was curious to know whether you discovered that it mattered.

namer/namer.cc
```diff
@@ -473,9 +481,44 @@ class SymbolFinder {
...
+
+        // This is an error in resolver, but just ignore this here.
+        if (cast->cast != core::Names::let()) {
+            return false;
+        }
+
+        core::FoundField found;
+        found.kind = uid->kind == ast::UnresolvedIdent::Kind::Instance ? core::FoundField::Kind::Instance
+                                                                       : core::FoundField::Kind::Class;
+        found.owner = getOwner();
```
jez, line 508: I wonder if it makes sense to say here that the owner is actually the class that will own the field/static-field, and not have to bring methods into the fold?

e.g., if we stored (1) whether it was a class var or instance var and (2) whether it appeared in a self method, is that enough? (i'm not sure)

I'm asking because

- The logic around deleting methods gets hairy, and I would love for it to be completely obvious that method symbols can be added/deleted without touching instance variables
- It's somewhat odd that we do `addField` for something that we know will create a `static-field` symbol eventually even though we have a separate `FoundStaticField`. (I mostly see the line of reasoning how we arrived there.)

Maybe if added a `FoundStaticField` for the class vars, then the only bit we would need is whether the assignment happened in a singleton class method or instance method, and then the `found.owner` could be the enclosing class 100% of the time?

This is not necessarily blocking feedback, but given that it seems like `insertField` is never called, and calling it correctly involves intermixing logic across when the methods get deleted and re-added, I wonder if this approach sidesteps that complication entirely.

test/pipeline_test_runner.cc
```diff
@@ -849,6 +849,7 @@ TEST_CASE("PerPhaseTest") { // NOLINT
     {
         // namer
         for (auto &tree : trees) {
+            core::UnfreezeNameTable nameTableAccess(*gs);
```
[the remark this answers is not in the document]
froydnj: I must have added this when I was still trying to mangle rename things out of the way or something.

Smoketests appear to pass locally with this bit reverted, so I've pushed that.

namer/namer.cc
froydnj: I think I initially tried to get this all to work without pushing methods onto the ownership stack, and it got pretty hairy, as well as not being correct in all cases (some of that might have been the issues with alias methods that I didn't see until later?).  Note that you also need to handle variables appearing outside method definitions (the AST does not yet have explicit `static-init` methods at this point).

I could see that you could have `FoundDefinition::Kind::MethodRef` or something on the ownership stack representing that you are now inside a method, but computing the owner of a `MethodRef` in `getOwnerSymbol` doesn't involve looking at `definedMethods`.

froydnj:

→ https://go/builds/bui_MIAqWjJHJRyJHp
→ https://go/builds/bui_MIAq5JjQ6QgRmH

froydnj:

→ https://go/builds/bui_MIdfxPqTDGJvi1
→ https://go/builds/bui_MIdfDr2Q5KmPWF

jez:
(Un-assigning myself until you'd like me to take a look, just let me know.)

test/testdata/rewriter/flatten_module_private_class_method.rb.symbol-table-raw.exp
```diff
@@ -20,4 +22,6 @@ class <C <U <root>>> < <C <U Object>> ()
       argument <blk><block> @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=??? end=???}
     method <S <C <U B>> $1>#<U thing> (<blk>) @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=11:3 end=11:17}
       argument <blk><block> @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=??? end=???}
+  class <S <S <C <U B>> $1> $1>[<C <U <AttachedClass>>>] < <S <S <C <U Object>> $1> $1> () @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=10:7 end=10:8}
+    type-member(+) <S <S <C <U B>> $1> $1>::<C <U <AttachedClass>>> -> LambdaParam(<S <S <C <U B>> $1> $1>::<C <U <AttachedClass>>>, lower=T.noreturn, upper=AppliedType {   klass = <S <C <U B>> $1>   targs = [     <C <U <AttachedClass>>> = B   ] }) @ Loc {file=test/testdata/rewriter/flatten_module_private_class_method.rb start=10:7 end=10:8}
```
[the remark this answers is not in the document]
jez: In that case, this looks slightly incorrect... I think this is saying that we're now creating `A.singleton_class.singleton_class`? Does that class make sense to be in the symbol table? It doesn't have any other members on it.

Even more obviously, the `A.bar` and `B.bar` methods still do not have their `flags.isPrivate` bit set (the snapshot didn't change for those symbols).

namer/namer.cc
```diff
@@ -1322,11 +1430,21 @@ class SymbolDefiner {
                 // We need alias methods in the FoundDefinitions list not so that we can actually
                 // create method symbols for them yet, but just so we can know which alias methods
                 // to delete on the fast path. Alias methods will be defined later, in resolver.
+                //
+                // We still need to put something here so that other found definitions that
+                // reference methods will get the correct symbols.
+                definedMethods.emplace_back(core::MethodRef{});
```
jez, line 1436: What do you mean by this? ooc, was there a test that failed unless you did this?

namer/namer.cc
```diff
@@ -1300,7 +1404,11 @@ class SymbolDefiner {
         definedMethods.reserve(foundDefs.methods().size());
 
         for (auto ref : foundDefs.nonDeletableDefinitions()) {
-            defineNonDeletableSingle(ctx, ref);
+            // We say that the "owner" of fields declared in methods is the method itself.
+            // This is consistent with how we define the owner of contexts during treewalks.
+            if (ref.kind() != core::FoundDefinitionRef::Kind::Field) {
+                defineNonDeletableSingle(ctx, ref);
+            }
```
jez, line 1411: ditto this, I might have misunderstood but I thought we were not going to need to do this anymore.

jez (commented):
Still some questions

namer/namer.cc
```diff
@@ -203,11 +203,18 @@ class SymbolFinder {
...
         // After flatten, method defs have been hoisted and reordered, so instead we look for the
         // keep_def / keep_self_def calls, which will still be ordered correctly relative to
         // visibility modifiers.
+        ownerStack.emplace_back(def);
+        methodVisiStack.emplace_back(nullopt);
+    }
+
+    void postTransformMethodDef(core::Context ctx, ast::ExpressionPtr &tree) {
+        methodVisiStack.pop_back();
+        ownerStack.pop_back();
```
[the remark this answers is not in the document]
froydnj: The recent changes ("ensure that fields always have a class owner") mean that `FoundField`s have a `FoundClass` owner when they are entered, rather than having whatever is at the top of the owner stack.

We need methods in the owner stack so we can select the correct context for defining fields (e.g. are we in a singleton method context or an instance method context), even if we're going to ensure that fields are owned by classes, not methods.  At the moment, it also seems easier to have methods in the owner stack so we can select the correct context for other things, but I am open to changing that.

I guess there should be tests for attempting to define various kinds of things (type members, static fields, etc.) inside methods, since namer has never had to think about those sorts of things before.

namer/namer.cc
```diff
@@ -545,11 +605,14 @@ class SymbolDefiner {
                 return definedClasses[ref.idx()];
             case core::FoundDefinitionRef::Kind::Method:
                 ENFORCE(ref.idx() < definedMethods.size());
+                // Ensure that nobody is trying to reference alias methods.
+                ENFORCE(definedMethods[ref.idx()].exists());
```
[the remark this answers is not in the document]
froydnj: I didn't have a ruby snippet in mind when I wrote this.  I'm not exactly sure how it would come about.

But given that alias methods get nonexistent symbols in `definedMethods`, I was trying to make sure that we weren't referencing those alias methods for ownership.

test/testdata/rewriter/flatten_module_private_class_method.rb.symbol-table-raw.exp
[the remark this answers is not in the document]
froydnj: Ah, right.  Fixed.

namer/namer.cc
froydnj: We now set `resultType` directly on a pre-existing field ("set type directly on todo fields").  I guess we should be able to go further and just `ENFORCE` that the field is there after `findMember` (or making entering the "uncommon" case inside a conditional) instead?

(I suppose it's also possible that we have a pre-existing thing that is *not* a field, and we'd still need the `enter` variants for that case.)

namer/namer.cc
froydnj: When I wrote the comment, I meant that you could run into a situation where field X had method owner M at INDEX...but if there were some alias methods defined prior to M, then we wouldn't actually have table entries for those alias methods, and then INDEX would point at the wrong method when we went to go define field X.  And no `ENFORCE`s would catch this because the indices were still in bounds--at least for the testcases where things were going wrong; I guess you could write a testcase that tries to catch that?

rbi generation failed because we weren't putting refs in for alias methods.

froydnj:
Thanks for the last review; I think I have addressed everything, though the current patch series continues to push methods onto the owner stack during symbol discovery.

namer/namer.cc
```diff
@@ -922,7 +985,9 @@ class SymbolDefiner {
 
         auto owner = ctx.owner.enclosingClass(ctx);
         if (mod.name == core::Names::privateClassMethod() || mod.name == core::Names::packagePrivateClassMethod()) {
-            owner = owner.data(ctx)->singletonClass(ctx);
+            if (!owner.data(ctx)->isSingletonClass(ctx)) {
+                owner = owner.data(ctx)->singletonClass(ctx);
+            }
```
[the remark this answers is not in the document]
jez: (This would be a good symbol-table.exp test to add as well)

jez (commented):
I still have some qualms about the fact that the `enter*FieldSymbol` calls remain in resolver—I think we should be able to delete those as a forcing function to ensure that namer is now doing all of the heavy lifting, ensuring that there aren't any cases that slip through the cracks somehow.

If it sounds like I'm confused or if you'd like to chat synchronously about any of this just lmk, I'm happy to.

namer/namer.cc
```diff
@@ -1085,6 +1150,34 @@ class SymbolDefiner {
         }
     }
 
+    core::FieldRef insertField(core::MutableContext ctx, const core::FoundField &field) {
+        auto scope = ctx.owner.enclosingClass(ctx);
+
+        auto existing = scope.data(ctx)->findMember(ctx, field.name);
```
jez, line 1156: In resolver, the code to enter class variables is preceded by some logic to make sure if the user writes

```ruby
class A
  def example
    @@x = T.let(0, T.nilable(Integer))
  end
end
```

that an error is reported and then the owner is changed to the enclosing class.

Should we move that logic here into namer?

Since that logic is present in resolver but absent from namer, I'm worried that we might be entering the `@@x` variables declared inside methods (both instance and singleton) on in two separate places.

froydnj: We have tests for what happens when we incorrectly declare variables in the wrong context (`test/testdata/resolver/let_errors.rb`, including a symbol table exp test), and those tests did not change as a result of this patch series, so we can at least say that we're not putting things in the wrong place.  (It is possible that future changes would de-sync namer and resolver and/or remove the test coverage that we had, so putting the logic in a single place might be a good idea anyway.)

That error-reporting logic is a little awkward to port over with the last few changes, because that logic is dependent upon the ownership of `ctx` being correctly attributed to a method, which would argue in favor of fields being "owned" by methods.  I will see if the past few changes can be reverted and the errors moved.

froydnj:
Patch stack status update:

1. Things have been moved back so methods are on the owner stack.  This is slightly weird, because we have different terminology for the owner stack and e.g. the owner of actual symbols, but putting methods on the owner stack means that we have more consistency with how e.g. tree walking manages `Context` ownership.
2. Resolver now enforces that fields are pre-existing, so namer bears the full burden of making sure the symbol table is correct.
3. Resolver is still responsible for issuing errors related to field declarations (non-use of `T.let`, duplicate variable declarations, declaring variables in the wrong place, etc.).  I have no strong opinion on whether namer should be responsible for some of those.

resolver/resolver.cc
```diff
@@ -2059,44 +2059,38 @@ class ResolveTypeMembersAndFieldsWalk {
...
 
         auto prior = scope.data(ctx)->findMember(ctx, uid->name);
-        if (prior.exists() && prior.isFieldOrStaticField()) {
-            auto priorField = prior.asFieldRef();
-            if (core::Types::equiv(ctx, priorField.data(ctx)->resultType, castType)) {
-                // We already have a symbol for this field, and it matches what we already saw, so we can short
-                // circuit.
-                return;
+        ENFORCE(prior.exists());
+        ENFORCE(prior.isFieldOrStaticField());
```
jez, line 2063: I was surprised that when I tested this, it doesn't trip for this example:

```ruby
# typed: true

class A
  X = 1
  X = type_member
end

class B
  X = type_member
  X = 1
end
```

Do you understand how the mangling works here? I guess that maybe we always mangle the type member symbol, and deal with that elsewhere?

froydnj: Type members (and non-class variable static fields generally) are dealt with in a different code path:

https://github.com/sorbet/sorbet/blob/d315899bda4c76868181c98b748cc72a82ee0ef5/resolver/resolver.cc#L2915-L2976

`handleFieldDeclaration` pushes a `ResolveFieldItem` job for instance variables and class variable assignments.  General static field resolution is handled elsewhere.

jez (approved):

froydnj:

→ https://go/builds/bui_MKnuuNfDXK3GTA
→ https://go/builds/bui_MKnuozX4XLYO4d

Issue #4454, opened 2021-08-04
Abstract classes can be instantiated


Morriar:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Foo%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0Aend%0A%0AFoo.new)

```ruby
# typed: true

class Foo
  extend T::Helpers

  abstract!
end

Foo.new # Should be a static error?
```

#### Observed output

No error.

#### Expected behavior

Sorbet should catch this?

jez:
https://github.com/sorbet/sorbet/pull/2222

^ @XrXr tried implementing this a long time ago. i haven't looked since then, not sure if anything has changed.

jez:
I don't know why #2143 was closed, so may as well keep this issue open.

PR #2222, opened 2019-11-25
Add static error for instantiating an abstract class

XrXr:
### Motivation
The runtime has a check for this but it can be defeated if the abstract
class defines `initialize`. Efficently fixing that problem in the
runtime involves using `prepend`, which is risky as it invalidates
assumptions scattered around the codebase.

This only checks calling `new` on a class marked directly with
`abstract!`. Subclasses of abstract classes that are not concrete
already get errors about missing definitions so we should be okay there.

Thanks to @jez for suggesting where to put this check!

Closes #2143

### Test plan

See included automated tests.

PR #6273, opened 2022-08-19
Error when abstract class is initialized

wildmaples:
This PR adds a new error `AbstractClassInstantiated` when an abstract class is initialized. 

Example: 

```ruby
class Abstract
  extend T::Sig
  extend T::Helpers
  abstract!

  sig {abstract.void}
  def foo; end
end

Abstract.new # error: Attempt to instantiate abstract class `Abstract`
```

### Motivation

Fixes: https://github.com/sorbet/sorbet/issues/4454

Previously, Alan implemented a version of this change (https://github.com/sorbet/sorbet/pull/2222) but did not cover the [common Ruby patterns pointed out in this comment](https://github.com/sorbet/sorbet/pull/2222#issuecomment-558753052): 

```ruby
class A;
  abstract!
end
class B < A
end

sig{params(a: T.class_of(A)).void}
def takesA(a)
  a.new
end

def assignsA
  a = T.let(B, T.class_of(A))
  a.new 
end
```

In this implementation, we check any abstract class that was initialized, if it is, we also check its singleton class's transitive methods to make sure that there isn't another `.new` defined.

### Test plan

See included automated tests - I believe the cases should have been covered.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -963,6 +963,37 @@ public:
 
         validateOverriding(ctx, methodDef.symbol);
     }
+
+    void postTransformSend(core::Context ctx, ast::ExpressionPtr &tree) {
+        auto &send = ast::cast_tree_nonnull<ast::Send>(tree);
+        if (send.fun != core::Names::new_()) {
+            return;
+        }
+
+        auto *id = ast::cast_tree<ast::ConstantLit>(send.recv);
+        if (id == nullptr || !id->symbol.exists() || !id->symbol.isClassOrModule()) {
+            return;
+        }
+
+        auto symbol = id->symbol.asClassOrModuleRef().data(ctx);
+        if (!symbol->flags.isAbstract) {
+            return;
+        }
+
+        auto singletonClass = symbol->lookupSingletonClass(ctx.state);
+        if (!singletonClass.exists()) {
+            return;
+        }
+
+        auto method_new = singletonClass.data(ctx)->findMethodTransitive(ctx.state, core::Names::new_());
+        // If the .new method we find is owned by Class, that means
+        // there was no user defined .new method, which warrants an error.
+        if (method_new.data(ctx)->owner == core::Symbols::Class()) {
+            if (auto e = ctx.beginError(send.loc, core::errors::Resolver::AbstractClassInstantiated)) {
+                e.setHeader("Attempt to instantiate abstract class `{}`", id->symbol.show(ctx));
+            }
+        }
+    }
 };
 } // namespace
 
@@ -970,7 +1001,7 @@ ast::ParsedFile runOne(core::Context ctx, ast::ParsedFile tree) {
     Timer timeit(ctx.state.tracer(), "validateSymbols");
 
     ValidateWalk validate;
-    ast::ShallowWalk::apply(ctx, validate, tree.tree);
+    ast::TreeWalk::apply(ctx, validate, tree.tree);
     return tree;
 }
 
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -990,7 +990,9 @@ public:
         // there was no user defined .new method, which warrants an error.
         if (method_new.data(ctx)->owner == core::Symbols::Class()) {
             if (auto e = ctx.beginError(send.loc, core::errors::Resolver::AbstractClassInstantiated)) {
-                e.setHeader("Attempt to instantiate abstract class `{}`", id->symbol.show(ctx));
+                auto symbolName = id->symbol.show(ctx);
+                e.setHeader("Attempt to instantiate abstract class `{}`", symbolName);
+                e.addErrorLine(id->symbol.loc(ctx), "`{}` defined here", symbolName);
             }
         }
     }
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -3970,94 +3970,6 @@ public:
     }
 };
 
-class CheckAbstractClassInitialization {
-public:
-    struct CheckAbstractClassInitializationWalkResult {
-        vector<ast::ParsedFile> trees;
-    };
-
-    void postTransformSend(core::Context ctx, ast::ExpressionPtr &tree) {
-        auto &send = ast::cast_tree_nonnull<ast::Send>(tree);
-        if (send.fun != core::Names::new_()) {
-            return;
-        }
-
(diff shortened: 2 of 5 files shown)


resolver/resolver.cc
```diff
@@ -3917,6 +3917,83 @@ class ResolveSanityCheckWalk {
...
+                }
+            }
+
+            if (!output.trees.empty()) {
+                auto count = output.trees.size();
+                outputq->push(move(output), count);
+            }
+        });
+
+        return jobs;
```
froydnj, line 3992: It looks like you need to collect the parsed files back into `jobs` from `outputq` here.

resolver/resolver.cc
```diff
@@ -3917,6 +3917,83 @@ class ResolveSanityCheckWalk {
...
+        }
+
+        auto symbol = id->symbol.asClassOrModuleRef().data(ctx);
+        auto singletonClass = symbol->lookupSingletonClass(ctx.state);
+
+        if (!singletonClass.exists()) {
+            return tree;
+        }
+
+        if (symbol->flags.isAbstract) {
```
froydnj, line 3944: Structuring the code like this would be slightly better:

```suggestion
        auto symbol = id->symbol.asClassOrModuleRef().data(ctx);
        if (!symbol->flags.isAbstract) {
            return tree;
        }
        auto singletonClass = symbol->lookupSingletonClass(ctx.state);
        if (!singletonClass.exists()) {
            return tree;
        }

        ....
```

resolver/resolver.cc
```diff
@@ -3917,6 +3917,83 @@ class ResolveSanityCheckWalk {
...
+
+        workers.multiplexJob("checkAbstractClassInitialization", [&gs, inputq, outputq]() -> void {
+            CheckAbstractClassInitialization walker;
+            CheckAbstractClassInitialization::CheckAbstractClassInitializationWalkResult output;
+
+            ast::ParsedFile current_job;
+            for (auto result = inputq->try_pop(current_job); !result.done(); result = inputq->try_pop(current_job)) {
+                if (result.gotItem()) {
+                    core::Context ctx(gs, core::Symbols::root(), current_job.file);
+                    current_job.tree = ast::TreeMap::apply(ctx, walker, std::move(current_job.tree));
```
froydnj, line 3981: Please use `ast::TreeWalk::apply` here; you'll have to modify `CheckAbstractClassInitialization` somewhat, but your pass will run much faster.

froydnj (changes requested):
Thanks for the patch!  It looks like there are rather a lot of test failures, at least partly due to the bug below.

I think you're slightly better off running this as part of `definition_validator/validator.cc`; `CheckAbstractClassInitialization` would be blocking typechecking in LSP mode, which we're trying to avoid, whereas `validator.cc` bits are non-blocking since they're already running during typechecking.  And we're already doing a tree walk there, so you're not adding a completely separate tree walk.

resolver/resolver.cc
```diff
@@ -3970,6 +3970,94 @@ class ResolveSanityCheckWalk {
...
+        }
+
+        auto method_new = singletonClass.data(ctx)->findMethodTransitive(ctx.state, core::Names::new_());
+        // If the .new method we find is owned by Class, that means
+        // there was no user defined .new method, which warrants an error.
+        if (method_new.data(ctx)->owner == core::Symbols::Class()) {
+            if (auto e = ctx.beginError(send.loc, core::errors::Resolver::AbstractClassInstantiated)) {
+                e.setHeader("Attempt to instantiate abstract class `{}`", id->symbol.show(ctx));
+            }
+        }
```
froydnj, line 4007: Thinking about this a little more, this is basically trying to implement method lookup and checks on the call, which is what `core/types/calls.cc` is doing.  Perhaps this check could be implemented there, instead, in `Class_new` or `Magic_selfNew`.

KaanOzkan: `calls.cc` implementation can't cover the case outlined in https://github.com/sorbet/sorbet/pull/2222#issuecomment-558753052 so I decided to go with `validator.cc` approach. I'm not super convinced Sorbet shouldn't error in the case linked but it'd be a tough migration and would stray away from `sorbet-runtime` behaviour.

Offline we talked about the `resolver` approach as well however, that doesn't allow the error to be reported on the fast path. Also it resulted in a 80ms check in our monolith compared to a 3% increase with the `validator.cc` implementation. It also makes sense to me that we check for abstract on the same walk as `override` and `final` checks.

definition_validator/validator.cc
```diff
@@ -963,14 +963,45 @@ class ValidateWalk {
...
+            return;
+        }
+
+        auto method_new = singletonClass.data(ctx)->findMethodTransitive(ctx.state, core::Names::new_());
+        // If the .new method we find is owned by Class, that means
+        // there was no user defined .new method, which warrants an error.
+        if (method_new.data(ctx)->owner == core::Symbols::Class()) {
+            if (auto e = ctx.beginError(send.loc, core::errors::Resolver::AbstractClassInstantiated)) {
+                e.setHeader("Attempt to instantiate abstract class `{}`", id->symbol.show(ctx));
+            }
```
jez, line 994: Can we also include the location where the abstract class was defined in the error message?

jez:
I have fixed the errors that this introduces on Stripe's codebase. In basically every case it was a problem that we wish we would have been able to catch earlier.

website/docs/error-reference.md
```diff
@@ -2746,6 +2746,24 @@ class MyClass < AbstractSerializable
...
+  extend T::Helpers
+  abstract!
+
+  sig {abstract.void}
+  def foo; end
+end
+
+Abstract.new # error: Attempt to instantiate abstract class `Abstract`
+```
+
```
jez, line 2766: The error reference is used best when it's describing how to unblock the user. (By contrast, the documentation as currently written merely repeats what the error message says.)

Let's add some suggestions:

```suggestion

To fix this error, there are some options:

- If the class which is marked `abstract!` does not actually have any `abstract` methods, simply remove `abstract!` from the class definition to fix the error.
- If the class _does_ have `abstract` methods, find some concrete subclass to call `new` on instead. If the call to `new` is in a test file, you may wish to make a new, test-only subclass of the abstract class. (Depending on the specifics of the test, it may even be possible to simply define all the abstract methods to simply `raise`, so that other aspects of the parent class can be tested.)

```

PR #6274, opened 2022-08-19
delete `Sorbet::Private::Static.keep_for_ide`

froydnj:
(Stacked on top of #6272)

### Motivation

The motivation here is roughly the same motivation as for #5823: We can have smaller tree nodes if we have specialized nodes, and we can avoid typechecking useless sends.  Package visibility checking should also get slightly better, as we're not flipping `VisibilityCheckerPass.ignoreConstant` quite so often...though that phase is parallel, so speedups are harder to realize there.

Not sure yet what the impact is on performance; for reference, there are roughly 4x as many `keep_for_ide` calls as there were casts on Stripe's codebase.  Back of the envelope calculations for memory suggests that this saves maaaaaybe 1% of max RSS.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_MH8eu0bUglMgAu
→ https://go/builds/bui_MH8eJfaVogcrCO

froydnj:
> Not sure yet what the impact is on performance; for reference, there are roughly 4x as many `keep_for_ide` calls as there were casts on Stripe's codebase. Back of the envelope calculations for memory suggests that this saves maaaaaybe 1% of max RSS.

Max RSS numbers were on point.  Performance-wise, this saves about 3-4% of typechecking time, and about 1% of symbolizing trees during naming (presumably because we're creating smaller tree nodes and therefore initializing less stuff?).  Package visibility checking showed zero change, somewhat surprisingly.

Also excited that `cfg::KeepAlive` gets another use!

ast/Helpers.cc
```diff
@@ -90,6 +89,10 @@ bool BehaviorHelpers::checkEmptyDeep(const ExpressionPtr &expr) {
                      checkEmptyDeep(seq.expr);
         },
 
+        [&](const ast::RuntimeMethodDefinition &) { result = true; },
+
+        [&](const ast::KeepForIDE &) { result = true; },
+
```
jez, line 95: We don't have great tools to detect cases where we need to start handling new pieces of syntax... It's the only thing that worries me when adding new tree nodes.

ast/Trees.h
```diff
@@ -1160,6 +1161,23 @@ EXPRESSION(RuntimeMethodDefinition) {
...
 CheckSize(RuntimeMethodDefinition, 16, 8);
 
+EXPRESSION(KeepForIDE) {
+public:
+    const core::LocOffsets loc;
+    ExpressionPtr expr;
+
+    KeepForIDE(core::LocOffsets loc, ExpressionPtr expr);
+
+    ExpressionPtr deepCopy() const;
```
jez, line 1171: Where is `KeepForIDE::deepCopy` implemented?

cfg/builder/builder_walk.cc
```diff
@@ -929,6 +929,16 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                 ret = current;
             },
 
+            [&](ast::KeepForIDE &keep) {
+                // Like ast::Cast's expansion, this is the only way to ensure that the
+                // constant expression here makes it into the CFG for LSP to hit on
+                // its location.
+                LocalRef deadSym = cctx.newTemporary(core::Names::keepForIde());
+                current = walk(cctx.withTarget(deadSym), keep.expr, current);
+                current->exprs.emplace_back(deadSym, core::LocOffsets::none(), make_insn<KeepAlive>(deadSym));
```
jez, line 938: Why are we using `core::LocOffsets::none()` here if we have `keep->loc`?

jez: (I'm generally somewhat paranoid of rare usages of a loc that doesn't exist, because so much code in Sorbet secretly assumes that they exist.)

main/autogen/autogen.cc
```diff
@@ -18,7 +18,7 @@ class AutogenWalk {
     vector<DefinitionRef> nesting;
 
     enum class ScopeType { Class, Block };
-    vector<ast::Send *> ignoring;
+    vector<ast::ExpressionPtr::tagged_storage> ignoring;
```
jez, line 21: Is `ExpressionPtr::tagged_storage` API that we'd prefer not to expose?

main/autogen/autogen.cc
```diff
@@ -301,13 +300,22 @@ class AutogenWalk {
...
 
+    void preTransformKeepForIDE(core::Context ctx, ast::ExpressionPtr &tree) {
+        ignoring.push_back(tree.getTagged());
+    }
+
+    void postTransformKeepForIDE(core::Context ctx, ast::ExpressionPtr &tree) {
+        ENFORCE(!ignoring.empty());
+        ENFORCE(ignoring.back() == tree.getTagged());
+        ignoring.pop_back();
+    }
```
jez, line 317: Taking a step back, do we need to do this at all?

It looks like the old code was only looking at `keepForIde` because it wanted to say that those node types did not define behavior, because normally `Send` nodes are said to define behavior.

Now that we don't have `keepForIde` sends, is it possible to just delete all the code that deals with `KeepForIDE` from this file? e.g. leave `ignoring` as a `vector<Send *>`?

jez (commented):
Few questions about the autogen stuff.

Generally, I'm not sure we have a good story around making sure that we've updated all the places throughout Sorbet's pipeline that need to be updated when we introduce new kinds of AST nodes, so I'd really appreciate it if you took an extra moment to look for any other places that we might have wanted to update.

Otherwise, looks good to me. If the current autogen.cc changes are the best we can do, lmk and I can stamp.

main/autogen/autogen.cc
froydnj: Before, we had sends like `Sorbet::Private::Static.keep_for_ide(Some::Constant::Of::Mine)`.  We needed to ignore any constants passed as an argument to that method.

Now, we have a magic `<keep-for-ide(Some::Constant::Of::Mind)>` node.  We still need to ignore the constant that appears as its lone argument...but we also need to keep ignoring constants in other `Send`-related contexts.

So I think we need to record both kinds of nodes here.  It would be cool if we had some way of controlling how the treewalk continues from within the `TreeWalk` methods, but that would be a decent-sized task.

For the record, I don't particularly like this solution in `autogen.cc`; `taggedStorage` being a quasi-public API is gross, but it certainly is convenient to use here, rather than having to manually do pointer tagging ourself with the raw `Send` or `KeepForIDE` pointers.

main/autogen/autogen.cc
froydnj: It's not my favorite API in the world, and it's not great to have another use of it in the tree.  I think manually doing some kind of pointer tagging is the only alternative (see other comment in this file).

cfg/builder/builder_walk.cc
froydnj: Eh, we can use `keep->loc`, which is zero-length, but the assignment here really doesn't have a loc.

jez (approved):

jez (approved):

jez:
Superceded by #7837

PR #6275, opened 2022-08-20
Fix broken link to /docs/nilable-types

mcgregor-stripe:
### Motivation

The [TypeScript ↔ Sorbet Reference doc](https://sorbet.org/docs/from-typescript) has a broken link.

### Test plan

I've manually verified the URL of the [Nilable Types doc](https://sorbet.org/docs/nilable-types). I believe code review should be enough additional verification for this.


froydnj (approved):
Thank you!

PR #6278, opened 2022-08-22
make changes to instance variables take the fast path in LSP

froydnj:
### Motivation

Better UX in LSP.

### Test plan

See included automated tests.

### DIFF
--- a/core/FileHash.cc
+++ b/core/FileHash.cc
@@ -103,6 +103,10 @@ string FoundMethodHash::toString() const {
                        owner.idx, owner.useSingletonClass, nameHash._hashValue, arityHash._hashValue);
 }
 
+void FoundFieldHash::sanityCheck() const {
+    ENFORCE(nameHash.isDefined());
+}
+
 FileHash::FileHash(LocalSymbolTableHashes &&localSymbolTableHashes, UsageHash &&usages, FoundDefHashes &&foundHashes)
     : localSymbolTableHashes(move(localSymbolTableHashes)), usages(move(usages)), foundHashes(move(foundHashes)) {}
 
--- a/core/FileHash.h
+++ b/core/FileHash.h
@@ -116,8 +116,35 @@ CheckSize(FoundMethodHash, 12, 4);
 
 using FoundMethodHashes = std::vector<FoundMethodHash>;
 
+struct FoundFieldHash {
+    struct {
+        // The owner of this field.
+        uint32_t idx : 30;
+        // Whether the field was defined on instances of the class or on the singleton class.
+        bool onSingletonClass : 1;
+        // Whether the field was a class or instance variable.
+        // TODO(froydnj) we should just subsume class variables into the more
+        // general static fields, since that's how we represent them internally.
+        bool isInstanceVariable : 1;
+    } owner;
+
+    // Hash of this field's name.
+    const FullNameHash nameHash;
+
+    FoundFieldHash(uint32_t ownerIdx, bool onSingletonClass, bool isInstanceVariable, FullNameHash nameHash)
+        : owner({ownerIdx, onSingletonClass, isInstanceVariable}), nameHash(nameHash) {
+        sanityCheck();
+    }
+
+    void sanityCheck() const;
+};
+CheckSize(FoundFieldHash, 8, 4);
+
+using FoundFieldHashes = std::vector<FoundFieldHash>;
+
 struct FoundDefHashes {
     FoundMethodHashes methodHashes;
+    FoundFieldHashes fieldHashes;
 };
 
 // When a file is edited, we run index and resolve it using an local (empty) GlobalState.
--- a/core/FileHash.h
+++ b/core/FileHash.h
@@ -174,6 +174,7 @@ struct LocalSymbolTableHashes {
     // TODO(jez) Is it worth having two of these? After http://go/srbi/5808 lands, re-evaluate
     // whether we should merge these into one vector like "symbolShapeHashes" or something
     std::vector<SymbolHash> staticFieldHashes;
+    std::vector<SymbolHash> fieldHashes;
 
     static uint32_t patchHash(uint32_t hash) {
         if (hash == LocalSymbolTableHashes::HASH_STATE_NOT_COMPUTED) {
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2238,6 +2238,7 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
     uint32_t methodHash = 0;
     UnorderedMap<WithoutUniqueNameHash, uint32_t> methodHashesMap;
     UnorderedMap<WithoutUniqueNameHash, uint32_t> staticFieldHashesMap;
+    UnorderedMap<WithoutUniqueNameHash, uint32_t> fieldHashesMap;
     int counter = 0;
 
     for (const auto &sym : this->classAndModules) {
@@ -2291,8 +2292,11 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
             hierarchyHash = mix(hierarchyHash, symhash);
             classAliasHash = mix(classAliasHash, symhash);
         } else {
-            hierarchyHash = mix(hierarchyHash, symhash);
-            fieldHash = mix(fieldHash, symhash);
+            auto &target = fieldHashesMap[WithoutUniqueNameHash(*this, field.name)];
+            target = mix(target, symhash);
+            uint32_t fieldShapeHash = field.fieldShapeHash(*this);
+            hierarchyHash = mix(hierarchyHash, fieldShapeHash);
+            fieldHash = mix(fieldHash, fieldShapeHash);
         }
 
         if (DEBUG_HASHING_TAIL && counter > this->fields.size() - 15) {
@@ -2333,15 +2337,20 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
     unique_ptr<LocalSymbolTableHashes> result = make_unique<LocalSymbolTableHashes>();
     result->methodHashes.reserve(methodHashesMap.size());
     result->staticFieldHashes.reserve(staticFieldHashesMap.size());
+    result->fieldHashes.reserve(fieldHashesMap.size());
     for (const auto &[nameHash, symbolHash] : methodHashesMap) {
         result->methodHashes.emplace_back(nameHash, LocalSymbolTableHashes::patchHash(symbolHash));
     }
     for (const auto &[nameHash, symbolHash] : staticFieldHashesMap) {
         result->staticFieldHashes.emplace_back(nameHash, LocalSymbolTableHashes::patchHash(symbolHash));
     }
+    for (const auto &[nameHash, symbolHash] : fieldHashesMap) {
+        result->fieldHashes.emplace_back(nameHash, LocalSymbolTableHashes::patchHash(symbolHash));
+    }
     // Sort the hashes. Semantically important for quickly diffing hashes.
     fast_sort(result->methodHashes);
     fast_sort(result->staticFieldHashes);
+    fast_sort(result->fieldHashes);
 
     result->hierarchyHash = LocalSymbolTableHashes::patchHash(hierarchyHash);
     result->classModuleHash = LocalSymbolTableHashes::patchHash(classModuleHash);
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1771,6 +1771,18 @@ void GlobalState::deleteMethodSymbol(MethodRef what) {
     this->methods[what.id()] = this->methods[0].deepCopy(*this);
 }
 
+void GlobalState::deleteFieldSymbol(FieldRef what) {
+    ENFORCE(what.data(*this)->flags.isField);
+    const auto &whatData = what.data(*this);
+    auto owner = whatData->owner;
+    auto &ownerMembers = owner.data(*this)->members();
+    auto fnd = ownerMembers.find(whatData->name);
+    ENFORCE(fnd != ownerMembers.end());
+    ENFORCE(fnd->second == what);
+    ownerMembers.erase(fnd);
+    this->fields[what.id()] = this->fields[0].deepCopy(*this);
+}
+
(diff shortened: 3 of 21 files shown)


core/FileHash.cc
```diff
@@ -103,6 +103,10 @@ string FoundMethodHash::toString() const {
                        owner.idx, owner.useSingletonClass, nameHash._hashValue, arityHash._hashValue);
 }
 
+void FoundFieldHash::sanityCheck() const {
+    ENFORCE(nameHash.isDefined());
+}
+
```
jez, line 109: Not sure if you care; you may want to add a `FoundFieldHash::toString` while you're here. I used the found method has one from time to time for print debugging but `¯\_(ツ)_/¯` up to you

core/FileHash.h
```diff
@@ -116,8 +116,35 @@ CheckSize(FoundMethodHash, 12, 4);
...
 
+struct FoundFieldHash {
+    struct {
+        // The owner of this field.
+        uint32_t idx : 30;
+        // Whether the field was defined on instances of the class or on the singleton class.
+        bool onSingletonClass : 1;
+        // Whether the field was a class or instance variable.
+        // TODO(froydnj) we should just subsume class variables into the more
+        // general static fields, since that's how we represent them internally.
```
jez, line 127: 👌

core/GlobalState.cc
```diff
@@ -1771,6 +1771,18 @@ void GlobalState::deleteMethodSymbol(MethodRef what) {
     this->methods[what.id()] = this->methods[0].deepCopy(*this);
 }
 
+void GlobalState::deleteFieldSymbol(FieldRef what) {
```
jez, line 1774:
```suggestion
// Before using this method, double check the disclaimer on GlobalState::deleteMethodSymbol above.
void GlobalState::deleteFieldSymbol(FieldRef what) {
```

core/GlobalState.cc
```diff
@@ -1771,6 +1771,18 @@ void GlobalState::deleteMethodSymbol(MethodRef what) {
...
+    ENFORCE(what.data(*this)->flags.isField);
```
jez, line 1775: :cloud_yell:

If you want, the WIP changes I have to support static-fields on fast path have mostly been minor. We may be able to sneak in a refactor to split things into FieldRef and StaticFieldRef after this PR and before I land the static-fields change.

core/GlobalState.cc
```diff
@@ -1771,6 +1771,18 @@ void GlobalState::deleteMethodSymbol(MethodRef what) {
...
+    const auto &whatData = what.data(*this);
+    auto owner = whatData->owner;
+    auto &ownerMembers = owner.data(*this)->members();
+    auto fnd = ownerMembers.find(whatData->name);
+    ENFORCE(fnd != ownerMembers.end());
+    ENFORCE(fnd->second == what);
+    ownerMembers.erase(fnd);
+    this->fields[what.id()] = this->fields[0].deepCopy(*this);
+}
```
jez, line 1784: slightly annoying that this logic is exactly the same as the method code, except for the types 🤔 no change expected just food for thought.

core/GlobalState.cc
```diff
@@ -2238,6 +2250,7 @@ unique_ptr<LocalSymbolTableHashes> GlobalState::hash() const {
     uint32_t methodHash = 0;
     UnorderedMap<WithoutUniqueNameHash, uint32_t> methodHashesMap;
     UnorderedMap<WithoutUniqueNameHash, uint32_t> staticFieldHashesMap;
+    UnorderedMap<WithoutUniqueNameHash, uint32_t> fieldHashesMap;
```
jez, line 2253: Do we actually need to keep these as separate? Would it be possible to rename `methodHashesMap` to simply `deleteableSymbolHashesMap` (or some better name)?

I think it would mean that apart from some slight renames in LSPFileUpdates.cc, the core logic there wouldn't have to change.

And then when non-class alias static fields also take the fast path, they would also be able to use the same logic.

namer/namer.cc
```diff
@@ -1414,6 +1414,57 @@ class SymbolDefiner {
         }
     }
 
+    void deleteViaFullNameHash(core::MutableContext ctx, const core::FoundFieldHash &oldDefHash) {
```
jez, line 1417: very slight preference for `delete<Kind>ViaFullNameHash` instead of overloading. not blocking.

namer/namer.cc
```diff
@@ -1414,6 +1414,57 @@ class SymbolDefiner {
...
+                continue;
+            }
+            auto memberField = memberSym.asFieldRef();
+
+            auto fieldNameToHash = memberName;
+            if (fieldNameToHash.kind() == core::NameKind::UNIQUE) {
+                auto &uniqueData = fieldNameToHash.dataUnique(ctx);
+                if (uniqueData->uniqueNameKind == core::UniqueNameKind::MangleRename ||
+                    uniqueData->uniqueNameKind == core::UniqueNameKind::MangleRenameOverload ||
+                    uniqueData->uniqueNameKind == core::UniqueNameKind::Overload) {
```
jez, line 1447: These two should not happen for fields.

I think the mangle rename ones _could_ appear here? We probably want to write a fast_path test that fails unless there is this special handling for `MangleRename`.

namer/namer.cc
```diff
@@ -1414,6 +1414,57 @@ class SymbolDefiner {
...
+        ENFORCE(oldDefHash.nameHash.isDefined(), "Can't delete rename if old hash is not defined");
+
+        // Because a change to classes would have take the slow path, should be safe
+        // to look up old owner in current foundDefs.
+        auto ownerSymbol = getOwnerSymbol(ownerRef);
+        ENFORCE(ownerSymbol.isClassOrModule());
+        auto owner = ownerSymbol.asClassOrModuleRef();
+        if (oldDefHash.owner.onSingletonClass) {
+            owner = owner.data(ctx)->singletonClass(ctx);
+        }
```
jez, line 1428: iiuc, we don't use `methodOwner` here like we do in the other `delete` method because of how the `FoundFieldHash`'s owner field is populated? Or is there some other reason why this snippet does not mirror the corresponding section of the other method?

namer/namer.cc
```diff
@@ -1414,6 +1414,57 @@ class SymbolDefiner {
...
+
+        // We have to accumulate a list of fields to delete, instead of deleting them in the loop
+        // below, because deleting a field invalidates the members() iterator.
+        vector<core::FieldRef> toDelete;
+
+        // Note: this loop is accidentally quadratic. We run deleteViaFullNameHash once per field
+        // previously defined in this file, then in each call look at each of that field's class's members
```
jez, line 1435: lol what was I saying here
```suggestion
        // previously defined in this file, then in each call look at each member of that field's owner
```

if you want i'd appreciate if you changed the other method too 🙈

namer/namer.cc
```diff
@@ -1432,6 +1483,14 @@ class SymbolDefiner {
...
         // methods, then define all the methods.
         if (oldFoundHashes.has_value()) {
+            for (const auto &oldFieldHash : oldFoundHashes.value().fieldHashes) {
+                if (oldFieldHash.owner.isInstanceVariable) {
+                    deleteViaFullNameHash(ctx, oldFieldHash);
+                }
+            }
+
+            // TODO(froydnj): what happens if methods are deleted in one file but reappear
+            // in another?  It seems like this should be OK, but need to verify.
```
jez, line 1493: Is this the bug we discussed on slack a few days ago, relating to the comment just above about not interleaving by file? Should we commit that test and/or update this comment to say that there is a known bug?

namer/namer.cc
```diff
@@ -1432,6 +1483,14 @@ class SymbolDefiner {
         // all the `nonDeletableDefinitions` from all files get defined, then delete all the old
         // methods, then define all the methods.
         if (oldFoundHashes.has_value()) {
+            for (const auto &oldFieldHash : oldFoundHashes.value().fieldHashes) {
+                if (oldFieldHash.owner.isInstanceVariable) {
+                    deleteViaFullNameHash(ctx, oldFieldHash);
+                }
+            }
```
jez, line 1490: another thought: should we attempt to put `fieldHashes` and `methodHashes` into the same vector on oldFoundHashes, so that we can just have one loop here, and then make `deleteViaFullNameHash` know how to operate on two different kinds of symbols? not sure what feels best, especially considering that we will add type members and static fields ~soon

namer/namer.cc
```diff
@@ -1414,6 +1414,57 @@ class SymbolDefiner {
...
+        auto ownerSymbol = getOwnerSymbol(ownerRef);
+        ENFORCE(ownerSymbol.isClassOrModule());
+        auto owner = ownerSymbol.asClassOrModuleRef();
+        if (oldDefHash.owner.onSingletonClass) {
+            owner = owner.data(ctx)->singletonClass(ctx);
+        }
+
+        // We have to accumulate a list of fields to delete, instead of deleting them in the loop
+        // below, because deleting a field invalidates the members() iterator.
+        vector<core::FieldRef> toDelete;
```
jez, line 1432: I don't have strong opinions here, but just a thought: i wonder whether it makes sense to keep a `vector<core::SymbolRef>` here, let the individual refs automatically be upcast on assignment, and then downcast them when needed later for sake of deleting?

jez (commented):
This is very neat. I think that all the code as written makes sense.

The only question I don't know the answer to yet is whether we need to add more tests. For both the methods change and the static fields change, what I've done is try to find all the error codes that Sorbet reports in namer/resolver/definition validator that relate to that kind of symbol.

For example, we probably want special tests like this:

```ruby
class A
  def one_method
    @foo = T.let(nil, T.nilable(Integer))
  end

  def two_method
    @foo = T.let(nil, T.nilable(String))
  end
end
```

On the slow path, and when the defs are in the same file, we report an error here for conflicting types.

We should make sure that if we split defs like this across two files and make an edit to change `@foo1` to `@foo` that the error shows up (and when we edit back, that the error goes away).

Error 5006 is an example of one of these kinds of instance-variable-specific namer/resolver/definition validator errors. You can probably audit the list of errors to see what others there are, and add similar shapes of tests.

core/GlobalState.cc
froydnj: I was `:sadpanda:` about this too, but I'm not sure there's a great way around it atm.  I guess we could have a helper like:

```c++
template <typename SymbolType, typename DataType>
void deleteSymbol(GlobalState &gs, SymbolType what, vector<DataType> &datavec) {
  ...
}

void GlobalState::deleteFieldSymbol(FieldRef what) {
  deleteSymbol(*this, what, this->fields);
}
```

I don't know if that's better or worse.

core/GlobalState.cc
froydnj: I don't have a strong opinion about this.  I have mild FUD about throwing everything into one big bin and resulting collisions vs. multiple small bins and fewer/no collisions thanks to the binning...but I don't know that it actually matters, and it would be nice to not have to update a zillion places with almost the exact same code for different symbol types.

It is slightly more consistent with the strong typing we have for symbols to have separate bins for things, I think...but if we treat those bins effectively the same, does the distinction really matter? :thinking:

namer/namer.cc
froydnj: I am happy to push a PR for that test, but wasn't sure how to handle the bit of it that fails, but shouldn't.  I guess commit as-is, but file an issue saying we have this expected error that actually shouldn't be there?

namer/namer.cc
froydnj: #6339 for the method version, will fix the field version here.

namer/namer.cc
froydnj: Same as other places in this PR: no super-strong opinion feeling, slight preference for keeping everything separate, but annoyance at how doing so requires updating a zillion places.

namer/namer.cc
froydnj: Slight preference for stronger source-level typing if we are going to do per-symbol-type deletes; if we decided to roll everything up into a single `deleteViaFullNameHash`, then we'd almost have to go the `vector<core::SymbolRef>` route.

froydnj (commented):
There are some other tests that now need fast-path assertions, and I will try to write some more this afternoon.

namer/namer.cc
froydnj: The test is #6348.  I will remove this comment.

froydnj:
> The only question I don't know the answer to yet is whether we need to add more tests. For both the methods change and the static fields change, what I've done is try to find all the error codes that Sorbet reports in namer/resolver/definition validator that relate to that kind of symbol.

The "update lsp path assertions" commit should show what kind of tests we have where we now take the fast path, and I think they cover all the relevant namer/resolver errors.  I guess there's not many cross-file tests, which could be a problem?  But I think we have the same issues with e.g. moving fields across files as we would with methods (cf. #6348), so I'm not sure how useful some of those tests would be at this point.

All of the styles of tests have class variable equivalents, but since class variables are static fields, they don't take the fast path for additions and deletions.

jez (approved):
This looks great, thanks for all the tests!

Sorry I didn't get to this sooner. Hopefully you get a chance to turn this on tomorrow and we'll start to see data when I'm back from vacation?

test/testdata/lsp/fast_path/ivar_redeclare.2.rbupdate
```diff
@@ -1,5 +1,5 @@
 # typed: true
-# assert-slow-path: true
+# assert-fast-path: ivar_redeclare.rb
```
jez, line 2: One more suggestion for a test:

Do you want to add one that changes the a `prop` or `const` in a `T::Struct`?

Doing so should change the getter method, setter method, constructor, and ivar types, but we should be able to take the fast path now.

jez: (Possibly larger question: might be worth having some tests that intermix renaming methods and renaming ivars? Maybe rename the method that declares the ivar?)

froydnj: I added some basic tests for each of these.  It's really cool to see editing a `prop` take the fast path!

PR #6279, opened 2022-08-22
be smarter about adding nodes for IDE when symbolizing classes

froydnj:
### Motivation

Stop doing work we don't need to do.  We could probably be even smarter and only add these when `--lsp` hasn't been passed, but I'm not sure that's worth it.

Also, this combined with #6274 speeds things up enough that adding #6265 into the mix becomes ~performance-neutral with respect to current master.

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1551,13 +1551,14 @@ public:
             }
         }
         auto loc = klass.declLoc;
-        ast::InsSeq::STATS_store ideSeqs;
+        ast::InsSeq::STATS_store retSeqs;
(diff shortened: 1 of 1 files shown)


jez:
> We could probably be even smarter and only add these when `--lsp` hasn't been passed, but I'm not sure that's worth it.

I don't think we should have substantial parts of Sorbet's pipeline that are only active when LSP is active. We get a lot of mileage in terms of being able to debug things by having the batch and incremental pipelines not be different.

jez (approved):

PR #6280, opened 2022-08-22
convert `FoundMethodHashes` into a `FoundHashes` structure

froydnj:
### Motivation

We currently compute hashes for found methods during file hashing, and we use those hashes to inform how to incrementally typecheck things in LSP mode.

We plan on adding support for incrementally typechecking when other entities are modified; rather than pass a bunch of params, let's just pass a single structure that will (one day) encompass all the things we want to pass, and then we can add a single structure member in one place, rather than outparams ten different places.

I did not modify the `FileRef` -> method hashes map in this patch, because I think it's a little easier to follow, but I'm happy to do that work as a followup.

### Test plan

See included automated tests.

### DIFF
--- a/core/FileHash.h
+++ b/core/FileHash.h
@@ -116,6 +116,10 @@ CheckSize(FoundMethodHash, 12, 4);
 
 using FoundMethodHashes = std::vector<FoundMethodHash>;
 
+struct FoundHashes {
+    FoundMethodHashes methodHashes;
+};
+
 // When a file is edited, we run index and resolve it using an local (empty) GlobalState.
 // We then hash the symbols defined in that local GlobalState, and use the result to quickly decide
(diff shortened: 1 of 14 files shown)


core/FileHash.h
```diff
@@ -245,11 +249,10 @@ struct UsageHash {
 struct FileHash {
     LocalSymbolTableHashes localSymbolTableHashes;
     UsageHash usages;
-    FoundMethodHashes foundMethodHashes;
+    FoundHashes foundHashes;
```
jez, line 252: What do you think about `FoundDefHashes`?

jez (approved):

core/FileHash.h
froydnj: WFM.  I substituted the name, I did not change parameter names.  LMK if you'd like me to change this.

jez: lgtm

PR #6281, opened 2022-08-22
Publish `'sorbet'` gem from Linux

jez:
### Motivation

macOS marchine is down

### Test plan

See included automated tests.


froydnj (approved):

ashkulz:
@jez when can we expect the macOS builds to be published to RubyGems again?

jez:
Working on it. Unfortunately this isn't the kind of problem where asking about it makes it happen any faster 😕

When it's fixed I'll be sure to post.

ashkulz:
Thanks for the update, as long as it happens that's good. I wasn't sure if it was meant to be permanent or not :man_shrugging:

PR #6282, opened 2022-08-22
Run script and update uri.rbi

wildmaples:
### Motivation

We're missing `URI::File` from the [core library](https://ruby-doc.org/stdlib-3.1.0/libdoc/uri/rdoc/URI/File.html) in the uri.rbi and it's causing some `Unable to resolve constant File https://srb.help/5002` errors.

So I update the entire rbi file by running `ruby rbi/tools/sync-rdoc.rb rbi/stdlib/uri.rbi` which includes some documentation updates and [the change we need](https://github.com/sorbet/sorbet/compare/master...wildmaples:sorbet:add-uri-file?expand=1#diff-c35b453c46a6ef788f194a659d78f3e99ffcc53b5809506596e3155f01374527R2655).

### Test plan

No tests.

### DIFF
--- a/rbi/stdlib/uri.rbi
+++ b/rbi/stdlib/uri.rbi
@@ -21,290 +21,7 @@
 # sprintf "%.1f", 1.234 #=> "1.2"
 # ```
 #
-# ## What's Here
-#
-# [`Module`](https://docs.ruby-lang.org/en/2.7.0/Module.html)
-# [`Kernel`](https://docs.ruby-lang.org/en/2.7.0/Kernel.html) provides methods
-# that are useful for:
-#
-# *   [Converting](#module-Kernel-label-Converting)
-# *   [Querying](#module-Kernel-label-Querying)
-# *   [Exiting](#module-Kernel-label-Exiting)
-# *   [Exceptions](#module-Kernel-label-Exceptions)
-# *   [IO](#module-Kernel-label-IO)
-# *   [Procs](#module-Kernel-label-Procs)
-# *   [Tracing](#module-Kernel-label-Tracing)
-# *   [Subprocesses](#module-Kernel-label-Subprocesses)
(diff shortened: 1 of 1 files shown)


rbi/stdlib/uri.rbi
```diff
@@ -21,7 +21,290 @@
...
+#         any additional arguments.
+#
+#     [`syscall`](https://docs.ruby-lang.org/en/2.7.0/Kernel.html#method-i-syscall)
+# :       Runs an operating system call.
+#
+#     [`trap`](https://docs.ruby-lang.org/en/2.7.0/Kernel.html#method-i-trap)
+# :       Specifies the handling of system signals.
+#
+#     [`warn`](https://docs.ruby-lang.org/en/2.7.0/Kernel.html#method-i-warn)
+# :       Issue a warning based on the given messages and options.
```
jez: I was curious why this comment is showing up here, and I put up a PR to make it go away: #6285

rbi/stdlib/uri.rbi
```diff
@@ -67,7 +350,7 @@ end
 #   class RSYNC < Generic
 #     DEFAULT_PORT = 873
 #   end
-#   @@schemes['RSYNC'] = RSYNC
+#   register_scheme 'RSYNC', RSYNC
```
jez: It looks like there are a handful of unrelated changes, likely due to this warning:

https://github.com/sorbet/sorbet/blob/72ca0f24493bd968088f6b9adf3134a23f28c105/rbi/tools/sync-rdoc.rb#L591-L595

Do you mind using `rbenv exec rbi/tools/sync-rdoc.rb ...` to use the same Ruby version as configured in `.ruby-version` for updating this file?

jez: (I've gone ahead and done this for the current PR for you.)

jez (approved):

PR #6293, opened 2022-08-24
add tests for disallowed static fields within methods

froydnj:
### Motivation

I don't think we have any tests for the contents of the symbol table when we have bad constants like this.  (Or maybe we do, but it's nice to have them all in one location now?)

### Test plan

See included automated tests.  `C2` should not and does not appear in the exp files.


jez (approved):
I think we didn't have any symbol table tests for these because they all simply failed to parse until recently (~2 months ago: #6058)

PR #6294, opened 2022-08-24
Support allow-listing certain RBI files to declare behavior in autogen mode

aadi-stripe:
Adds a command-line flag in autogen mode that supports listing certain RBI files as being able to define behavior.

### Motivation

In the Stripe codebase, we autogenerate Ruby files that represent protobuf message declarations, which often have definitions of the form:

```
module Foo
  Bar ||= ...
```

Since `||=` constant reassignments are not supported by Sorbet, autogen skips interpreting `Foo::Bar` as a definition. This breaks our tooling. In order to fix this issue, given that we also generate RBI files for protobuf declarations, we can allow autogen to interpret these RBI files as defining behavior -- so that we can capture the definition of `Foo::Bar` in the serialized static analysis results from autogen.

### Test plan

### DIFF
--- a/main/autogen/autogen.cc
+++ b/main/autogen/autogen.cc
@@ -1,4 +1,5 @@
 #include "main/autogen/autogen.h"
+#include "absl/strings/match.h"
 #include "ast/Helpers.h"
 #include "ast/ast.h"
 #include "ast/treemap/treemap.h"
@@ -16,6 +17,7 @@ class AutogenWalk {
     vector<Reference> refs;
     vector<core::NameRef> requireStatements;
     vector<DefinitionRef> nesting;
+    const AutogenConfig *autogenCfg;
 
     enum class ScopeType { Class, Block };
     vector<ast::Send *> ignoring;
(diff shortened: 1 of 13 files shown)


aisamanra (approved):
This makes sense to me! I think in the long-term it might be nice to be clearer in the code that we have a distinction between "relevant for autogen output" and "defines behavior", because this conflates them, but this seems like a reasonable change for now and we can always clarify it later.

Have we already appropriately tested this on the codebase?

aadi-stripe:
> This makes sense to me! I think in the long-term it might be nice to be clearer in the code that we have a distinction between "relevant for autogen output" and "defines behavior", because this conflates them, but this seems like a reasonable change for now and we can always clarify it later.
> 
> Have we already appropriately tested this on the codebase?

I've tested it locally, but I need to run the PR build.

aadi-stripe:

→ https://go/builds/bui_MJ5DE7pnze0Pvj
→ https://go/builds/bui_MJ5DOTxYEI9Rhj

PR #6299, opened 2022-08-25
Allow `requires_ancestor` block to return a `T.class_of` object

### LINKED ISSUE #4835: `requires_ancestor` can't see class methods
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0Aextend%20T%3A%3AHelpers%0A%0Aclass%20AbstractObject%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns%28String%29%20%7D%0A%20%20def%20self.foo%0A%20%20%20%20'foo'%0A%20%20end%0Aend%0A%0Amodule%20Fooable%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20module%20ClassMethods%0A%20%20%20%20extend%20T%3A%3AHelpers%0A%20%20%20%20requires_ancestor%20%7B%20AbstractObject%20%7D%0A%20%20%20%20%23%20requires_ancestor%20%7B%20T.class_of%28AbstractObject%29%20%7D%20%23%20also%20invalid%0A%0A%20%20%20%20def%20bar%0A%20%20%20%20%20%20foo%0A%20%20%20%20end%0A%20%20end%0A%0A%20%20mixes_in_class_methods%28ClassMethods%29%0Aend%0A)

```ruby
# typed: true
extend T::Sig
extend T::Helpers

class AbstractObject
  extend T::Sig

  sig { returns(String) }
  def self.foo
    'foo'
  end
end

module Fooable
  extend T::Helpers

  module ClassMethods
    extend T::Helpers
    requires_ancestor { AbstractObject }
    # requires_ancestor { T.class_of(AbstractObject) } # also invalid

    def bar
      foo
    end
  end

  mixes_in_class_methods(ClassMethods)
end
```

#### Observed output

```
editor.rb:23: Method foo does not exist on Fooable::ClassMethods https://srb.help/7003
    23 |      foo
              ^^^
Errors: 1
```

#### Expected behavior

Typechecking succeeds.
Alternatively, that calling `requires_ancestor { T.class_of(AbstractObject) }` succeeds (and causes typechecking to succeed).

When defining a module that mixes_in_class_methods, it's common to require that including modules have certain class methods defined. But it seems impossible to communicate this to `requires_ancestor`, as that method can't accept types as arguments.

- - -

egiurleo:
Closes #4835.

This PR allows users to return `T.class_of` objects from the block of `requires_ancestor`. Using this functionality, users can now verify inheritance from a singleton class, rather than just a class or module, which is a feature that has been widely requested (see #4835).

### Motivation
Currently, `requires_ancestor` will not work if you try to return a singleton class/`T.class_of` object from its block argument. This prevents users from ensuring inheritance from a singleton class, which is useful in cases like this:

```
module MyHelper
  extend T::Helpers

  requires_ancestor { T.class_of(MyBaseClass) }

  def helper
    my_singleton_method
  end
end

class MyBaseClass
  class << self
    include MyHelper

    def my_singleton_method; end
  end
end
```

### Test plan
Automated tests have been added and updated to reflect this change.

### DIFF
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -317,25 +317,6 @@ public:
         return Hash(loc, std::move(keys), std::move(values));
     }
 
-    static bool isT(ExpressionPtr &expr) {
-        auto *tMod = ast::cast_tree<ast::ConstantLit>(expr);
-        return tMod && tMod->symbol == core::Symbols::T();
-    }
-
-    static bool isTClassOf(ExpressionPtr &expr) {
-        auto *send = ast::cast_tree<ast::Send>(expr);
-
-        if(send == nullptr) {
-            return false;
-        }
-
-        if(!isT(send->recv))  {
-            return false;
-        }
-
-        return send->fun == core::Names::classOf();
-    }
-
 private:
     static ExpressionPtr Params(core::LocOffsets loc, ExpressionPtr recv, Send::ARGS_store args) {
         ENFORCE(args.size() % 2 == 0, "Sig params must be arg name/type pairs");
--- a/ast/Helpers.h
+++ b/ast/Helpers.h
@@ -317,6 +317,25 @@ public:
         return Hash(loc, std::move(keys), std::move(values));
     }
 
+    static bool isT(ExpressionPtr &expr) {
+        auto *tMod = ast::cast_tree<ast::ConstantLit>(expr);
+        return tMod && tMod->symbol == core::Symbols::T();
+    }
+
+    static bool isTClassOf(ExpressionPtr &expr) {
+        auto *send = ast::cast_tree<ast::Send>(expr);
+
+        if(send == nullptr) {
+            return false;
+        }
+
+        if(!isT(send->recv))  {
+            return false;
+        }
+
+        return send->fun == core::Names::classOf();
+    }
+
 private:
     static ExpressionPtr Params(core::LocOffsets loc, ExpressionPtr recv, Send::ARGS_store args) {
         ENFORCE(args.size() % 2 == 0, "Sig params must be arg name/type pairs");
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -758,7 +758,7 @@ void validateUnsatisfiableRequiredAncestors(core::Context ctx, const core::Class
             requiredClasses.emplace_back(ancst);
         }
 
-        if (ancst.symbol.data(ctx)->typeArity(ctx) > 0) {
+        if (ancst.symbol.data(ctx)->typeArity(ctx) > 0 && !ancst.symbol.data(ctx)->isSingletonClass(ctx)) {
             if (auto e = ctx.state.beginError(data->loc(), core::errors::Resolver::UnsatisfiableRequiredAncestor)) {
                 e.setHeader("`{}` can't require generic ancestor `{}` (unsupported)", sym.show(ctx),
                             ancst.symbol.show(ctx));
--- a/rbi/sorbet/t.rbi
+++ b/rbi/sorbet/t.rbi
@@ -286,7 +286,7 @@ module T::Helpers
   # given type of module.
   #
   # For more information, see https://sorbet.org/docs/requires-ancestor
-  sig { params(block: T.proc.returns(Module)).void }
+  sig { params(block: T.proc.void).void }
   def requires_ancestor(&block); end
 end
 
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1913,13 +1913,8 @@ class ResolveTypeMembersAndFieldsWalk {
 
     void extendClassOfDepth(ast::Send &send) {
         if (trackDependencies_) {
-            classOfDepth_.emplace_back(isT(send.recv) && send.fun == core::Names::classOf());
+            classOfDepth_.emplace_back(ast::MK::isT(send.recv) && send.fun == core::Names::classOf());
         }
-    }
-
-    static bool isT(const ast::ExpressionPtr &expr) {
-        auto *tMod = ast::cast_tree<ast::ConstantLit>(expr);
-        return tMod && tMod->symbol == core::Symbols::T();
     }
 
     static bool isTodo(const core::TypePtr &type) {
@@ -2742,7 +2737,7 @@ public:
 
         if (send.fun == core::Names::typeAlias()) {
             // don't track dependencies if this is some other method named `type_alias`
-            if (!isT(send.recv)) {
+            if (!ast::MK::isT(send.recv)) {
                 extendClassOfDepth(send);
                 return;
             }
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1175,8 +1175,7 @@ private:
             ENFORCE(send);
 
             if(send->numPosArgs() == 1) {
-                auto arg = send->getPosArg(0).deepCopy();
-                if(auto *argClass = ast::cast_tree<ast::ConstantLit>(arg)) {
+                if(auto *argClass = ast::cast_tree<ast::ConstantLit>(send->getPosArg(0))) {
                     if(argClass != nullptr && argClass->symbol.exists() && argClass->symbol.isClassOrModule()) {
                         if constexpr (isMutableStateType) {
                             symbol = argClass->symbol.asClassOrModuleRef().data(gs)->singletonClass(gs);
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -93,6 +93,25 @@ namespace {
  * the fixed point loop at a high level.
  */
 
+bool isT(ast::ExpressionPtr &expr) {
+    auto *tMod = ast::cast_tree<ast::ConstantLit>(expr);
+    return tMod && tMod->symbol == core::Symbols::T();
+}
+
+bool isTClassOf(ast::ExpressionPtr &expr) {
+    auto *send = ast::cast_tree<ast::Send>(expr);
+
+    if(send == nullptr) {
+        return false;
+    }
+
+    if(!isT(send->recv))  {
+        return false;
+    }
+
+    return send->fun == core::Names::classOf();
+}
+
 class ResolveConstantsWalk {
     friend class ResolveSanityCheckWalk;
 
@@ -1150,7 +1169,7 @@ private:
             if(constant != nullptr && constant->symbol.exists() && constant->symbol.isClassOrModule()) {
                 symbol = constant->symbol.asClassOrModuleRef();
             }
-        } else if(ast::MK::isTClassOf(block->body)) {
+        } else if(isTClassOf(block->body)) {
             send = ast::cast_tree<ast::Send>(block->body);
 
             ENFORCE(send);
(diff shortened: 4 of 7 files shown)


ast/Helpers.h
```diff
@@ -317,6 +317,25 @@ class MK {
         return Hash(loc, std::move(keys), std::move(values));
     }
 
+    static bool isT(ExpressionPtr &expr) {
```
egiurleo, line 320: I wanted to use this helper method in multiple modules within the `resolver.cc` file and wasn't sure where I should put them... so I put them here! Happy to move them.

rbi/sorbet/t.rbi
```diff
@@ -286,7 +286,7 @@ module T::Helpers
   # given type of module.
   #
   # For more information, see https://sorbet.org/docs/requires-ancestor
-  sig { params(block: T.proc.returns(Module)).void }
+  sig { params(block: T.proc.void).void }
```
egiurleo, line 289: @Morriar and I were having a hard time typing this proc so that it accepted `T.class_of` as a return. Let me know if there's a better way to do this.

froydnj: I think given that there's no runtime behavior involved in `requires_ancestor`, we don't need to be too specific about the types here; the primary purpose is communicating information to the static checking phase.  (It would, of course, be nice if there was runtime checking, but that would be a separate discussion to have.)

ast/Helpers.h
froydnj: I think these methods can just live in `resolver.cc` as toplevel functions in the anonymous namespace.  Please move them into that file instead.

resolver/resolver.cc
```diff
@@ -1144,25 +1144,46 @@ class ResolveConstantsWalk {
         ENFORCE(block->body);
 
         auto blockLoc = core::Loc(todo.file, block->body.loc());
-        auto *id = ast::cast_tree<ast::ConstantLit>(block->body);
+        core::ClassOrModuleRef symbol = core::Symbols::StubModule();
 
-        if (id == nullptr || !id->symbol.exists() || !id->symbol.isClassOrModule()) {
+        if(auto *constant = ast::cast_tree<ast::ConstantLit>(block->body)) {
+            if(constant != nullptr && constant->symbol.exists() && constant->symbol.isClassOrModule()) {
```
froydnj, line 1150:
```suggestion
            if (constant->symbol.exists() && constant->symbol.isClassOrModule()) {
```

(The previous `cast_tree` will have ensured that we have a non-null pointer here.)

resolver/resolver.cc
```diff
@@ -1144,25 +1144,46 @@ class ResolveConstantsWalk {
...
+            }
+        } else if(ast::MK::isTClassOf(block->body)) {
+            send = ast::cast_tree<ast::Send>(block->body);
+
+            ENFORCE(send);
+
+            if(send->numPosArgs() == 1) {
+                auto arg = send->getPosArg(0).deepCopy();
+                if(auto *argClass = ast::cast_tree<ast::ConstantLit>(arg)) {
+                    if(argClass != nullptr && argClass->symbol.exists() && argClass->symbol.isClassOrModule()) {
```
froydnj, line 1161:
```suggestion
                    if (argClass->symbol.exists() && argClass->symbol.isClassOrModule()) {
```

(Same argument for non-`nullptr`-ness as above.)

resolver/resolver.cc
```diff
@@ -1144,25 +1144,46 @@ class ResolveConstantsWalk {
...
+            if(constant != nullptr && constant->symbol.exists() && constant->symbol.isClassOrModule()) {
+                symbol = constant->symbol.asClassOrModuleRef();
+            }
+        } else if(ast::MK::isTClassOf(block->body)) {
+            send = ast::cast_tree<ast::Send>(block->body);
+
+            ENFORCE(send);
+
+            if(send->numPosArgs() == 1) {
+                auto arg = send->getPosArg(0).deepCopy();
```
froydnj, line 1159: We don't need to make a copy here; you can either take a reference to the argument (`auto &arg = send->getPosArg(0);`) or you can just dispense with the variable and substitute `send->getPosArg(0)` into the `cast_tree` call below.

resolver/resolver.cc
```diff
@@ -1144,25 +1144,46 @@ class ResolveConstantsWalk {
...
             return;
         }
 
-        if (id->symbol == owner) {
+        if (symbol == owner) {
             if (auto e = gs.beginError(blockLoc, core::errors::Resolver::InvalidRequiredAncestor)) {
                 e.setHeader("Must not pass yourself to `{}`", send->fun.show(gs));
             }
             return;
         }
```
froydnj, line 1183: I think we need tests to ensure these errors also fire appropriately for `T.class_of(MODULE)`, e.g. you shouldn't be able to say `requires_ancestor { T.class_of(NonExistentModule) }` nor `requires_ancestor { T.class_of(MyClass) }` inside `MyClass`.

The error messages should probably be slightly different for bare `MODULE` vs. `T.class_of(MODULE)`.

It looks like `test/testdata/resolver/requires_ancestor_errors.rb` has tests like this, so that'd be the appropriate place to add them.

test/testdata/resolver/requires_ancestor_errors.rb
```diff
@@ -34,11 +34,10 @@ class << self
 module Helper5
   extend T::Helpers
 
-  requires_ancestor { "Object" } # error: Expected `Module` but found `String("Object")` for block result type
+  requires_ancestor { "Object" }
   #                   ^^^^^^^^ error: Argument to `requires_ancestor` must be statically resolvable to a class or a module
 
-  requires_ancestor { T.class_of(Object) } # error: Expected `Module` but found `Runtime object representing type: T.class_of(Object)` for block result type
-  #                   ^^^^^^^^^^^^^^^^^^ error: Argument to `requires_ancestor` must be statically resolvable to a class or a module
```
froydnj, line 41: Given that this code is now allowed, I think it would be good to add a test for a different send and ensure that we still error on that.

froydnj (changes requested):
I think this is reasonable; comments below.  It also looks like there are some linting failures that need to be fixed up for the next pass.

resolver/resolver.cc
```diff
@@ -1144,25 +1163,53 @@ class ResolveConstantsWalk {
...
+            if (send->numPosArgs() == 1) {
+                if (auto *argClass = ast::cast_tree<ast::ConstantLit>(send->getPosArg(0))) {
+                    if (argClass->symbol.exists() && argClass->symbol.isClassOrModule()) {
+                        if (argClass->symbol == owner) {
+                            if (auto e = gs.beginError(blockLoc, core::errors::Resolver::InvalidRequiredAncestor)) {
+                                e.setHeader("Must not pass yourself to `{}` inside of `requires_ancestor`",
+                                            send->fun.show(gs));
+                            }
+                            return;
+                        }
```
egiurleo, line 1186: @froydnj I added this error case here, let me know if that's alright.

test/testdata/resolver/requires_ancestor_errors.rb
```diff
@@ -34,18 +38,21 @@ class << self
...
+  requires_ancestor { Object.superclass }
```
egiurleo, line 44: Let me know if this is what you meant by "testing for a different send".

froydnj (approved):
Sorry for the delay in reviewing; I had looked this over yesterday, but apparently forgot to write final comments up.

This looks great!  Thank you for the extra tests.

definition_validator/validator.cc
```diff
@@ -758,7 +758,7 @@ void validateUnsatisfiableRequiredAncestors(core::Context ctx, const core::Class
             requiredClasses.emplace_back(ancst);
         }
 
-        if (ancst.symbol.data(ctx)->typeArity(ctx) > 0) {
+        if (ancst.symbol.data(ctx)->typeArity(ctx) > 0 && !ancst.symbol.data(ctx)->isSingletonClass(ctx)) {
```
jez, line 761: Could we add a test for this?

jez: In particular, this code is buggy. We cannot allow generics for `T.class_of` for the same reason that we do not for instance classes. The code as written suffers from this bug:

#### Input

[→ View on sorbet.run](https://sorbet.run/?arg=--enable-experimental-requires-ancestor#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Aclass%20Parent%0A%20%20class%20%3C%3C%20self%0A%20%20%20%20extend%20T%3A%3ASig%0A%20%20%20%20extend%20T%3A%3AGeneric%0A%0A%20%20%20%20Elem%20%3D%20type_member%0A%0A%20%20%20%20sig%20%7Breturns%28T.nilable%28Elem%29%29%7D%0A%20%20%20%20attr_accessor%20%3Aelem%0A%20%20end%0A%20%20%0Aend%0A%0Amodule%20Example%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20requires_ancestor%20%7BT.class_of%28Parent%29%7D%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20example%0A%20%20%20%20result%20%3D%20self.elem%0A%20%20%20%20T.reveal_type%28result%29%0A%20%20end%0Aend)

```ruby
# typed: strict
extend T::Sig

class Parent
  class << self
    extend T::Sig
    extend T::Generic

    Elem = type_member

    sig {returns(T.nilable(Elem))}
    attr_accessor :elem
  end
  
end

module Example
  extend T::Sig
  extend T::Helpers

  requires_ancestor {T.class_of(Parent)}

  sig {void}
  def example
    result = self.elem
    T.reveal_type(result)
  end
end
```

#### Observed output

```
editor.rb:25: Expression does not have a fully-defined type (Did you reference another class's type members?) https://srb.help/7011
    25 |    result = self.elem
                     ^^^^^^^^^
  Note:
    Historically, users seeing this error has represented finding a bug in Sorbet
    (or at least finding a test case for which there could be a better error message).
    Consider searching for similar bugs at https://github.com/sorbet/sorbet/issues
    or reporting a new one.

editor.rb:26: Revealed type: T.untyped https://srb.help/7014
    26 |    T.reveal_type(result)
            ^^^^^^^^^^^^^^^^^^^^^
    editor.rb:25:
    25 |    result = self.elem
                     ^^^^^^^^^
Errors: 2
```

jez: We should special case something to deal with the fact that all singleton classes already have one type member with the name `core::Names::Constants::AttachedClass()`, but not permit a singleton class that has other type members defined on it.

resolver/resolver.cc
```diff
@@ -1144,25 +1163,53 @@ class ResolveConstantsWalk {
...
+
+                        if constexpr (isMutableStateType) {
+                            symbol = argClass->symbol.asClassOrModuleRef().data(gs)->singletonClass(gs);
+                        }
```
jez, line 1190: This doesn't look right. Out of curiosity, what motivated this change?

Had you already tried doing this instead?

```cpp
symbol = argClass->symbol.asClassOrModuleRef().data(gs)->lookupSingletonClass(gs);
```

egiurleo: Oh thanks for pointing that method out. I didn't know it existed. I'll use that one instead!

egiurleo:
@jez I'll make a follow up PR addressing your comments!

PR #6300, opened 2022-08-25
Fix `InterfaceWrapper` for ruby 2.7/3 kwarg compatibility

aisamanra:
### Motivation
This lets us keep using `InterfaceWrapper` even on Ruby 3+. (It'd be nice to get rid of it, but it's currently load-bearing, so this kicks the can on getting rid of it.)

### Test plan

See included automated tests. Added a test and ran it (locally) with Ruby 2.6, 2.7, and 3.0 to verify that it correctly works on all of them.


jez (approved):
😢 I was really hoping we could spare a moment to rip this out and be done with it.

PR #6301, opened 2022-08-25
Make static field types idempotent on the fast path

jez:
### Motivation

The behavior of the test after this change is that the revealed types
change to `Integer`, because of some logic in resolver.

### Test plan

See included automated tests.


froydnj (changes requested):
It's terrible for me to be confused by such a short PR, but the description says:

> The behavior of the test after this change is that the revealed types
change to `Integer`, because of some logic in resolver.

...but the revealed types in the committed test are `T.nilable(String)`, and the revealed types at the time of writing [on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AX%20%3D%20T.let%28''%2C%20T.nilable%28String%29%29%0AT.reveal_type%28X%29%20%23%20error%3A%20%60T.nilable%28String%29%60%0AX%20%3D%200%0A%23%20%20%20%5E%20error%3A%20Expected%20%60T.nilable%28String%29%60%20but%20found%20%60Integer%280%29%60%20for%20field%0AT.reveal_type%28X%29%20%23%20error%3A%20%60T.nilable%28String%29%60) are `T.nilable(String)`.

So I'm confused by what behavior change this is supposed to be introducing.

jez:
On sorbet.run right now, if you add a newline before the final `T.reveal_type(X)` and then hover over it you'll see `Integer`. Before making the edit, it was `T.nilable(String)`. (ignore the output, because the output only ever shows the CLI output, not the fast path output).

Another way to verify what this change is doing is to understand how the current test fails, because it does in fact fail on master right now without this change:

- the LSPTest attempts to run the fast path with a no-op edit (inserting many `\n` characters at the start of the file)
- the test asserts that the old errors should still be reported, even after making the fast path edit
- in fact what happens is that the test runner fails to find the `T.nilable(String)` errors and finds a new `Integer` error

froydnj (approved):
Thank you for the explanation!  (I guess it's also a little weird to have a `test/testdata/lsp/fast_path` test with one file...)

jez:
> Thank you for the explanation! (I guess it's also a little weird to have a `test/testdata/lsp/fast_path` test with one file...)

That's a good point, I'll move to namer.

jez:
Sounds good I've put it back to where it was.

PR #6304, opened 2022-08-25
Revive ProtobufDescriptorPool rewriter

jez:
### Motivation

This lets autogen not have to use hacks to ignore errors in order to analyze
`*_pb.rb` files.

### Test plan

See included automated tests.


rewriter/ProtobufDescriptorPool.cc
```diff
@@ -0,0 +1,56 @@
...
+    if (cnstGoogle == nullptr || cnstGoogle->cnst != core::Names::Constants::Google()) {
+        return nullptr;
+    }
+
+    // Put the actual send in the tree too to require that the Protobuf constant / methods are actually there.
+    // (also needed for autogen to be able to see the behavior)
+    ast::ClassDef::RHS_store rhs;
+    rhs.emplace_back(std::move(asgn->rhs));
+
+    return ast::MK::ClassOrModule(asgn->loc, asgn->loc, asgn->lhs.deepCopy(), {}, std::move(rhs), kind);
```
froydnj, line 53:
```suggestion
    return ast::MK::ClassOrModule(asgn->loc, asgn->loc, std::move(asgn->lhs), {}, std::move(rhs), kind);
```

rewriter/rewriter.cc
```diff
@@ -197,6 +198,17 @@ class Rewriterer {
         }
     }
 
+    // NOTE: this case differs from the `Assign` typecase branch in `postTransformClassDef` above,
+    // as it will apply to all assigns, not just those that are present in the RHS of a `ClassDef`.
```
froydnj, line 202: Do we actually want it to apply to all assigns?  It seems like doing this for assigns inside e.g. a block or a method definition is pointless and could lead to weird error messages (maybe?).

(I can believe that this is the way the rewriter previously worked, but I'm curious if we'd like it to work differently this time around.)

rewriter/ProtobufDescriptorPool.h
```diff
@@ -0,0 +1,37 @@
...
+ *
+ *   class A
+ *     ::Google::Protobuf::DescriptorPool.generated_pool.lookup("...").msgclass
+ *   end
+ *
+ *   module A::B
+ *     ::Google::Protobuf::DescriptorPool.generated_pool.lookup("...").enummodule
+ *   end
+ *
+ * The idea is that the this way Sorbet will statically see `A` and `B` for what they are:
```
froydnj, line 23:
```suggestion
 * The idea is that this way Sorbet will statically see `A` and `B` for what they are:
```

froydnj (approved):

rewriter/rewriter.cc
jez: Inside a block at the class level, it should be fine, because class definitions are already allowed there.

Inside a method is impossible for this rewriter to fire, because our parser treats constant assignments in method bodies as parse errors and recovers from them by changing the assign to a variable:

[→ View on sorbet.run](https://sorbet.run/?arg=--print=desugar-tree#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Adef%20example%0A%20%20X%20%3D%201%0Aend)

```ruby
# typed: true
extend T::Sig

def example
  X = 1
end
```

#### Observed output

```
  <self>.extend(<emptyTree>::<C T>::<C Sig>)

  def example<<todo method>>(&<blk>)
    <dynamic-const-assign> = 1
  end
end
editor.rb:5: dynamic constant assignment https://srb.help/2001
     5 |  X = 1
          ^
Errors: 1
```

> (I can believe that this is the way the rewriter previously worked, but I'm curious if we'd like it to work differently this time around.)

In fact it _didn't_ work this way last time, and changing it to work this way now makes it correctly detect cases that it couldn't before.

Overall I'm not super worried, because the protobuf-generated files have such a predictable shape.

jez:

→ https://go/builds/bui_MLEsG6KZLS6YZm
→ https://go/builds/bui_MLEt0ULNnqNsm2

PR #6305, opened 2022-08-26
Added default behaviour for `Enum::try_serialize`

ignacio-chiazzo:
### Motivation
I use `Enum#try_deserialize` often in a place where the value depends on what's stored previously. That doesn't guarantee the value being valid.  If it's not _deserializable_, we apply a default value like the following:

```ruby
  card_suit = CardSuit.try_deserialize('blerg')
  card_suit = CardSuit::UNKNOWN if card_suit.nil?
```

This PR introduced a default_value to the method `try_deserialize`.

```ruby
  card_suit = CardSuit.try_deserialize('blerg', CardSuit::UNKNOWN)
```


gems/sorbet-runtime/lib/types/enum.rb
```diff
@@ -72,13 +72,16 @@ def self.each_value(&blk)
   #
   # Note: It would have been nice to make this method final before people started overriding it.
   # Note: Failed CriticalMethodsNoRuntimeTypingTest
-  sig {params(serialized_val: SerializedVal).returns(T.nilable(T.attached_class)).checked(:never)}
-  def self.try_deserialize(serialized_val)
+  sig do
+    params(serialized_val: SerializedVal, default_value: T.nilable(T::Enum))
+    .returns(T.nilable(T.any(T::Enum, T.attached_class))).checked(:never)
+  end
+  def self.try_deserialize(serialized_val, default_value: nil)
```
jez, line 79: 1. You used an optional keyword arg instead of an optional positional arg.
2. Your return type is too wide. Even for people that don't use the new `default_value` parameter, it will break their code by forcing them to consider the case where it's a `T::Enum` not a `MyEnum`. This can be fixed with a generic.

```suggestion
  sig do
    type_parameters(:U)
      .params(
        serialized_val: SerializedVal,
        default_value: T.nilable(T.type_parameter(:U))
      )
      .returns(T.nilable(T.any(T.attached_class, T.type_parameter(:U))))
      .checked(:never)
  end
  def self.try_deserialize(serialized_val, default_value: nil)
```

jez (commented):
I'm curious what you think the main selling point of this change is?

Using `||` at the call site is IMO better:

- It's more clear—you don't have to learn what this new second positional argument is for.
- It has better typing, because Sorbet is able to see the control flow from the `||`. In particular, for the `||` Sorbet understands that the overall expression is never `nil`, because it has been handled already.

[→ View example on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20T%3A%3AEnum%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20do%0A%20%20%20%20type_parameters%28%3AU%29%0A%20%20%20%20%20%20.params%28%0A%20%20%20%20%20%20%20%20serialized_val%3A%20T.untyped%2C%0A%20%20%20%20%20%20%20%20default_value%3A%20T.nilable%28T.type_parameter%28%3AU%29%29%0A%20%20%20%20%20%20%29%0A%20%20%20%20%20%20.returns%28T.nilable%28T.any%28T.attached_class%2C%20T.type_parameter%28%3AU%29%29%29%29%0A%20%20end%0A%20%20def%20self.my_try_deserialize%28serialized_val%2C%20default_value%3Dnil%29%0A%20%20end%0Aend%0A%0Aclass%20MyEnum%20%3C%20T%3A%3AEnum%0A%20%20enums%20do%0A%20%20%20%20X%20%3D%20new%0A%20%20%20%20Y%20%3D%20new%0A%20%20end%0Aend%0A%0Ax%20%3D%20MyEnum.my_try_deserialize%28'x'%2C%20MyEnum%3A%3AY%29%0AT.reveal_type%28x%29%0A%0Ay%20%3D%20MyEnum.my_try_deserialize%28'x'%29%20%7C%7C%20MyEnum%3A%3AY%0AT.reveal_type%28y%29)

Given all that, I'm leaning towards saying, "thanks but no thanks" but if you have a compelling argument where adding this argument particularly outshines what is possible now, I'm happy to reconsider.

ignacio-chiazzo:
>Using || at the call site is IMO better:

>It's more clear—you don't have to learn what this new second positional argument is for.


I'm not concerned about this since it's an opt-in feature, devs can still use ||.

> It has better typing, because Sorbet is able to see the control flow from the ||. In particular, for the || Sorbet understands that the overall expression is never nil, because it has been handled already.

This is a very good point I hadn't considered 🤔. Adding the default would require telling sorbet that the value is not nil when the default is specified. This is a good reason to close the PR. Thanks for the review!

PR #6307, opened 2022-08-26
add an lsp fast path test for swapping class order

froydnj:
### Motivation

This is pretty dumb, but it is the sort of thing that one might expect should take the fast path (even if it doesn't), so we might as well have it as a test.  It took me a little time to write the test and understand why, anyway.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/fast_path/swap_class_order.1.rbupdate
+++ b/test/testdata/lsp/fast_path/swap_class_order.1.rbupdate
@@ -3,10 +3,10 @@
 
 module A
   class C
-    def bar; end
+    def method_on_c; end
   end
 
   class B
-    def foo; end
+    def method_on_b; end
   end
 end
--- a/test/testdata/lsp/fast_path/swap_class_order.1.rbupdate
+++ b/test/testdata/lsp/fast_path/swap_class_order.1.rbupdate
@@ -1,0 +1,12 @@
+# typed: true
+# assert-slow-path: true
+
+module A
+  class C
+    def bar; end
+  end
+
+  class B
+    def foo; end
+  end
+end
--- a/test/testdata/lsp/fast_path/swap_class_order.rb
+++ b/test/testdata/lsp/fast_path/swap_class_order.rb
@@ -2,10 +2,10 @@
 
 module A
   class B
-    def foo; end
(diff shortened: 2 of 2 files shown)


test/testdata/lsp/fast_path/swap_class_order.1.rbupdate
```diff
@@ -0,0 +1,12 @@
+# typed: true
+# assert-slow-path: true
+
+module A
+  class C
+    def bar; end
```
jez, line 6:
```suggestion
    def method_on_c; end
```

test/testdata/lsp/fast_path/swap_class_order.1.rbupdate
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+  class B
+    def foo; end
```
jez, line 10:
```suggestion
    def method_on_b; end
```

test/testdata/lsp/fast_path/swap_class_order.rb
```diff
@@ -0,0 +1,11 @@
+# typed: true
+
+module A
+  class B
+    def foo; end
```
jez, line 5:
```suggestion
    def method_on_b; end
```

test/testdata/lsp/fast_path/swap_class_order.rb
```diff
@@ -0,0 +1,11 @@
...
+  end
+
+  class C
+    def bar; end
```
jez, line 9:
```suggestion
    def method_on_c; end
```

jez (approved):
Suggested some different names because I had to do a double take at first.

Have you looked into why this is the case?

jez:
I guess this is the reason?

https://github.com/sorbet/sorbet/blob/d315899bda4c76868181c98b748cc72a82ee0ef5/core/Symbols.cc#L2255-L2258

froydnj:
> Have you looked into why this is the case?

I believe some of the debugging logic I put in for fields on the fast path indicated the the class hashes were different, which would point at the class hashing code you pointed out (ordering of the class members matters for hashing purposes).

PR #6309, opened 2022-08-27
Fix a couple of Counters.h comment typos

Smattr:
### Motivation

I am not sure a simple typo fix like this needs motivation, but if so please let me know.

### Test plan

N/A


froydnj (approved):
No motivation needed for fixing typos.  😄 

Thank you!

PR #6310, opened 2022-08-29
make fields always owned by classes in namer

froydnj:
### Motivation

This is a followup to #6265.  Playing around with the necessary changes for #6278 indicate that we really do want fields to be owned by classes.  Otherwise, we have a sort of chicken-and-egg scenario for fields on the fast path:

1. Fields are owned by methods
2. We are going to delete (some) methods followed by re-defining (some) methods.
3. Therefore we have to delete and define fields prior to doing the same with methods.
4. But when we go to delete fields, we don't have enough information to figure out who the method "owner" is: we only have indices into a table of (not-yet defined!) methods, and worse, those indices are into a table that doesn't actually exist.

Making fields owned by classes works much better, as we can be assured the necessary class(es) are still around when we go to delete fields.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/infer/private_inside_method.rb
+++ b/test/testdata/infer/private_inside_method.rb
@@ -6,7 +6,7 @@ class A
 end
 
 # Sorbet needs a static view of the world. We assume that once a method is
-# defined, that's it's visibility always.
+# defined, that's its visibility always.
 
 # no runtime error
 A.foo
--- a/test/testdata/infer/private_inside_method.rb
+++ b/test/testdata/infer/private_inside_method.rb
@@ -6,12 +6,12 @@ class A
 end
 
 # Sorbet needs a static view of the world. We assume that once a method is
-# defined, that's its visibility always.
+# defined, that's it's visibility always.
 
-# no runtime error, but static error because we saw `private :foo`
-A.foo # error: Non-private call to private method `foo`
+# no runtime error
+A.foo
 
 A.bar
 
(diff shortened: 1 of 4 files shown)


froydnj:
`test/testdata/infer/private_inside_method.rb` now fails because, despite recursing into the method and seeing `private :foo`, we don't have the correct scope on which to look up `foo` (we have the class, but we need the singleton class).  I think this is solvable using all the same `onSingletonClass` games that `FoundField` did, but applied to `FoundModifier`, but I don't know whether we want to make a conscious decision to not support patterns like this.

(I guess part of the argument against is that `private :SYM` inside methods would work correctly in some cases, and then Sorbet would be inconsistently implementing method modifiers, which seems like a bad state of affairs.)

froydnj:
#6265 wound up changing things because we would examine method modifiers inside method bodies, due to doing complete walks of the tree.  I propose we just stop examining method modifiers inside method bodies instead, which is a reversion to pre-#6265 state.

test/testdata/infer/private_inside_method.rb
```diff
@@ -6,12 +6,12 @@ def self.bar; private :foo; end
 end
 
 # Sorbet needs a static view of the world. We assume that once a method is
-# defined, that's its visibility always.
+# defined, that's it's visibility always.
```
jez, line 9:
```suggestion
# defined, that's its visibility always.
```

jez (approved):

PR #6317, opened 2022-08-30
Configure VSCode extension to default to formatting on save when available

reese:
This reconfigures the VSCode to support `formatOnSave` for Ruby out-of-the-box. This modifies the extension to stub formatting requests if formatting isn't supported by the running LSP and to default the VSCode configuration to `formatOnSave` to `true` for the extension.

### Motivation

As we roll out format-on-save internally at Stripe, we still need a way for users to opt out if they want to. However, workspace settings will override user settings in many cases. Making the extension pass this as a default allows us to make the setting overridable by users.

### Test plan

See included automated tests. I also manually ran the VSCode extension on a small sandbox repo and confirmed that format on save turns off when either the VSCode configuration is overridden or when `--enable-experimental-lsp-document-formatting-rubyfmt` is not passed to the LSP. I also verified that the default configuration works when nothing is overridden.


jez (changes requested):
Am I reading this right that if an individual wants to opt out, they have to figure how to pass the command line flag to Sorbet, and the LSP server responds with some initialization result that the client can use?

I'm surprised to see that to turn of "format on save" we have to turn off formatting entirely. That means that not only can you not run it automatically on save, you can't run it at all, even if you manually invoke the format command from the command palette.

To solve the problem of allowing format on save to be a per-user config, can we do that in the VS Code plugin or in the VS Code config that we check into pay-server?

Sorry if I misunderstand how this change works.

reese:
The command line flag is almost certainly not what users will use. The option we'd suggest to users would be to add `"sorbet.enableDocumentFormatting": false` in their user settings instead (which is handled directly in the VSCode plugin). The command line flag would be intended for the initial rollout only, where we'd have it behind a dev feature flag that we could turn off if needed.

I agree that ideally we wouldn't turn off document formatting entirely -- if there's some LSP interaction specific format-on-save, I'd be happy to rework this to only stub that instead of document formatting as a whole, but my reading of [how this interaction is handled by VSCode](https://github.com/microsoft/vscode/blob/dce493cb6e36346ef2714e82c42ce14fc461b15c/src/vs/editor/contrib/format/browser/format.ts#L298) is that it just calls the document formatting provider for the given file. As far as I know (and again, happy to be told otherwise), format-on-save in VSCode isn't really propagated through the LSP, it's just making a formatting request pre-save.

The other alternative is indeed to do it in the VSCode workspace, but workspace settings in VSCode can't be overriden by user settings, so if we set the config in the workspace without an alternative workaround, there's no other way for users to be able to opt-out.

jez (commented):
Leaving this comment here so that my review too so my review request goes away

- - - - -

I looked for a bit at the docs to see if there was a way to allow an extension to override the editor.formatOnSave setting programmatically. I found something that almost worked but not quite. Maybe you found this too, but I figured I'd share anyways:

https://code.visualstudio.com/docs/getstarted/settings#_settings-precedence

^ In this doc, we see that "Language-specific default settings" come after both "Language-specific user settings" and "Language-specific workspace settings". So one option we have is to add some code to the `package.json` like

```{
  "contributes": {
    "configurationDefaults": {
      "[ruby]": {
        "editor.formatOnSave": true,
      }
    }
  }
}
```

(docs for [contributes.configurationDefaults](https://code.visualstudio.com/api/references/contribution-points#contributes.configurationDefaults))

The upside here is that this allows Stripes to put `"[ruby"]: {"editor.formatOnSave": false}` in their User Settings if they want to override the extension. The downside is that the Sorbet-provided default would take effect for all users of the Sorbet extension, regardless of what options had been passed to Sorbet to enable+configure rubyfmt.

So one option is we just lean into this: we change Sorbet so that it always responds to `textDocument/formatting` requests, but it returns a `null` response unless the relevant command line flags have been turned on. Then the extension can set `formatOnSave: true` and users can optionally disable it, either at the user level or the workspace level.

That involves only editor-agnostic changes inside Sorbet; all the VS Code-specific logic remains in the VS Code extension.

As far as I can see, the _only_ downside to that approach is that in an environment where the documentFormatting request goes over a network connection, the user still pays that time spent waiting.

So as another alternative, I put together this change, building on your existing change:

https://github.com/sorbet/sorbet/compare/jez-format-caps

This PR uses the existing `capabilities.documentFormattingProvider` boolean that Sorbet returns during initialization (alongside the new client-side setting your PR added) to have the VS Code extension decide whether to monkey patch the `sendRequest` method to turn `textDocument/formatting` requests into no-ops. This all happens client side, at the expense of incurring a monkey patch in the VS Code implementation, and not adding any new options to the Sorbet LSP implementation.

I'm fine with both approaches. Let me know what you think.

reese:
I ended up updating this with the second approach (checking the server capabilities during initialization and stubbing if document formatting isn't enabled). I don't feel particularly strongly about that position, but it involves the least-surprising changes to anyone who may use the LSP outside of VSCode and it means that anyone who runs the LSP over a network connection doesn't incur an extra penalty for saves. Happy to look at alternatives if this change doesn't pan out as we'd expect.

vscode_extension/src/LanguageClient.ts
```diff
@@ -70,6 +70,24 @@ export function shimLanguageClient(
   };
 }
 
+/**
+ * Handles short-circuiting format-on-save requests when either the client or server is
+ * configured to turn them off.
```
jez, line 75:
```suggestion
 * configured to turn them off.
 * 
 * We shim this here in the extension so that it doesn't matter which version of
 * Sorbet is installed. (An alternative would be to shim it in Sorbet, but this would
 * break if a newer VS Code extension (passing formatOnSave: true) is used to drive an
 * older Sorbet that doesn't respond to textDocument/formatting with no-op responses.)
```

jez (approved):
:portal-from: Slack thread:

> The language client is smart enough to not send the textDocument/formatting requests if the server responds with documentFormattingProvider: false, even if defaultFormatter is set to sorbet-vscode-extension and the extension sets the formatOnSave default to true.
>
> So I think we can delete all the code changes in the PR, leaving only the package.json change?

lgtm

PR #6319, opened 2022-08-30
do parallel maps with less work in the packager

froydnj:
### Motivation

The packager (both the actual packaging and the visibility checking pass(es)) are essentially parallel maps over the list of files we have.  This PR implements them in such a way to more closely reflect that, making the respective passes ~5% faster in the process.

It's worth noting that the same technique can be applied in other places for a small speedup, but this seemed like a good place to start.

It's also worth noting that I tried adding some `ENFORCE(absl::c_is_sorted(...));` scattered about.  I ran into testcase failures, because we have:

https://github.com/sorbet/sorbet/blob/567368a1dbc373991db5a44c869792610750b612/main/lsp/LSPTypechecker.cc#L412-L417

which does not provide sorted input to the resolving phase (packager/namer/resolver).  Upon reflection, I think not asserting sortedness is the right thing to do here: anything that cares about ordering (e.g. resolving signatures) will have to ensure sortedness beforehand; anything that doesn't care should not be providing stronger guarantees to downstream consumers.

### Test plan

See included automated tests.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -2,6 +2,7 @@
 #include "absl/strings/match.h"
 #include "absl/strings/str_join.h"
 #include "absl/strings/str_replace.h"
+#include "absl/synchronization/blocking_counter.h"
 #include "ast/Helpers.h"
 #include "ast/treemap/treemap.h"
 #include "common/FileOps.h"
@@ -13,6 +14,7 @@
 #include "core/Unfreeze.h"
 #include "core/errors/packager.h"
 #include "core/packages/PackageInfo.h"
+#include <algorithm>
 #include <cctype>
 #include <sstream>
 #include <sys/stat.h>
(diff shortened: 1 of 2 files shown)


packager/packager.cc
```diff
@@ -1520,31 +1521,15 @@ vector<ast::ParsedFile> Packager::run(core::GlobalState &gs, WorkerPool &workers
...
-                if (result.gotItem()) {
-                    files.insert(files.end(), make_move_iterator(threadResult.begin()),
-                                 make_move_iterator(threadResult.end()));
-                }
-            }
-        }
+        barrier.Wait();
     }
 
-    fast_sort(files, [](const auto &a, const auto &b) -> bool { return a.file < b.file; });
```
jez, line 1546: You're saying this is no longer required because we return the `files` in the same order as the input, so either it was sorted and stays sorted, or was not sorted but doesn't matter that it's sorted?

jez (approved):
This is neat. If I understand correctly, the difference is just that we don't actually care about the output, and all we need to do is wait for all the workers to finish?

Should we make it return `void` instead of the list of the files?

We can't make the `files` input vector be a const reference, can we?

(Also, are we passing in the `files` vector by move or by copy right now?)

froydnj:
> This is neat. If I understand correctly, the difference is just that we don't actually care about the output, and all we need to do is wait for all the workers to finish?

That's correct.  (The same sort of trick applies even in the case where the `multiplexJob` closure is modifying the trees.)

> Should we make it return `void` instead of the list of the files?

I don't have a strong opinion on this.

> We can't make the `files` input vector be a const reference, can we?

I don't think we can right now, because that would mean walking over `const` trees...and unfortunately we can't make the tree walker support `const` input without some more metaprogramming.

> (Also, are we passing in the `files` vector by move or by copy right now?)

We are passing it by `move`; if you try to pass it via copying, the compiler will give you an error, because the copy constructor would need copy construction of the elements to work, and `ExpressionPtr(const ExpressionPtr &)` has been `= delete`'d.

PR #6327, opened 2022-09-01
Add missing URI::Parser, URI::REGEXP

notEthan:
Ruby defines `URI::Parser` to alias `URI::RFC2396_Parser` (class). Likewise `URI::REGEXP` aliases `URI::RFC2396_REGEXP` (module). Sorbet doesn't appear to know these ([→ sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AURI%3A%3AParser%0AURI%3A%3AREGEXP)).

I've had a stab at the rbi file, though pretty uncertain I'm anywhere near the correct way to go about defining these aliases (is reopening URI later in the file acceptable? is `T.let` needed for a simple alias?). I'm happy to implement feedback on this, and I marked this PR as draft for this reason.

### Motivation

My project uses URI::Parser but sorbet cannot resolve that constant.

### Test plan

I added these to test/testdata/rbi/uri.rb - here too I have low confidence in how correctly I am doing it, and can improve / fix per feedback.

### DIFF
--- a/rbi/stdlib/uri.rbi
+++ b/rbi/stdlib/uri.rbi
@@ -137,8 +137,10 @@ module URI
   HOST = T.let(T.unsafe(nil), Regexp)
   HTML5ASCIIINCOMPAT = T.let(T.unsafe(nil), String)
   OPAQUE = T.let(T.unsafe(nil), Regexp)
+  Parser = URI::RFC2396_Parser
   PORT = T.let(T.unsafe(nil), Regexp)
   QUERY = T.let(T.unsafe(nil), Regexp)
+  REGEXP = URI::RFC2396_REGEXP
   REGISTRY = T.let(T.unsafe(nil), Regexp)
   REL_PATH = T.let(T.unsafe(nil), Regexp)
   REL_URI = T.let(T.unsafe(nil), Regexp)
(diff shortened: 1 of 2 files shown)


jez:
> reopening URI later in the file acceptable

The class is defined in the same file as the one you're reopening the class in. Why not simply add the alias above where the other constants in that class are defined?

https://github.com/sorbet/sorbet/blob/32b0f66a170bced4dcf02b08ff30c6ac23615338/rbi/stdlib/uri.rbi#L146

> is `T.let` needed for a simple alias

in fact using `T.let` prevents one of the tests you've added from working (the `T.class_of(...)` one). Simply define the alias like `A = B` and it will be better.

jez (changes requested):
> reopening URI later in the file acceptable

The class is defined in the same file as the one you're reopening the class in. Why not simply add the alias above where the other constants in that class are defined?

https://github.com/sorbet/sorbet/blob/32b0f66a170bced4dcf02b08ff30c6ac23615338/rbi/stdlib/uri.rbi#L146

> is `T.let` needed for a simple alias

in fact using `T.let` prevents one of the tests you've added from working (the `T.class_of(...)` one). Simply define the alias like `A = B` and it will be better.

notEthan:
Thanks, updated.

> Why not simply add the alias above

I hadn't because I'd be aliasing a thing that hadn't been defined yet, but I guess that is fine in rbi.

jez (approved):

PR #6328, opened 2022-09-01
unconditionally enable the new `bind_call` optimization

froydnj:
### Motivation

Testing at Stripe has shown this is working as expected, so we can go ahead and let `bind_call` use the optimization directly, without requiring an experimental API.

r? @penelope-stripe 

### Test plan

See included automated tests.


paracycle:
If this is vetted and if you have perf numbers that this works better, could you please submit this patch upstream? I think all Ruby users should benefit from this optimization.

penelope-stripe:
@paracycle https://github.com/ruby/ruby/pull/6244

penelope-stripe (approved):

jez (approved):

PR #6333, opened 2022-09-02
Add intrinsic for `+` on empty tuples

jez:
Before, the revealed type for `xs` was `T::Array[T.untyped]` after the
`+=` assignment.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3326,8 +3326,14 @@ public:
         auto *tuple = cast_type<TupleType>(args.thisType);
         ENFORCE(tuple != nullptr);
 
-        res = dispatchCallProxyType(gs, args.thisType.underlying(gs), args);
-        if (tuple->elems.empty() && args.args.size() == 1) {
+        auto specialEmptyTupleCase = tuple->elems.empty() && args.args.size() == 1 &&
+                                     (isa_type<TupleType>(args.args[0]->type) ||
+                                      Types::isSubType(gs, Types::arrayOfUntyped(), args.args[0]->type));
+        auto underlying =
+            specialEmptyTupleCase ? core::Types::arrayOf(gs, core::Types::bottom()) : args.thisType.underlying(gs);
+
+        res = dispatchCallProxyType(gs, underlying, args);
+        if (specialEmptyTupleCase) {
             res.returnType = args.args[0]->type;
         }
     }
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3320,6 +3320,19 @@ public:
     }
 } Tuple_concat;
 
+class Tuple_plus : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        auto *tuple = cast_type<TupleType>(args.thisType);
+        ENFORCE(tuple != nullptr);
+
+        res = dispatchCallProxyType(gs, args.thisType.underlying(gs), args);
+        if (tuple->elems.empty() && args.args.size() == 1) {
+            res.returnType = args.args[0]->type;
+        }
+    }
+} Tuple_plus;
+
 namespace {
 
 optional<Loc> locOfValueForKey(const GlobalState &gs, const Loc origin, const NameRef key, const TypePtr expectedType) {
@@ -4170,6 +4183,7 @@ const vector<Intrinsic> intrinsics{
     {Symbols::Tuple(), Intrinsic::Kind::Instance, Names::sample(), &Tuple_sample},
     {Symbols::Tuple(), Intrinsic::Kind::Instance, Names::toA(), &Tuple_to_a},
     {Symbols::Tuple(), Intrinsic::Kind::Instance, Names::concat(), &Tuple_concat},
+    {Symbols::Tuple(), Intrinsic::Kind::Instance, Names::plus(), &Tuple_plus},
 
     {Symbols::Shape(), Intrinsic::Kind::Instance, Names::squareBracketsEq(), &Shape_squareBracketsEq},
     {Symbols::Shape(), Intrinsic::Kind::Instance, Names::merge(), &Shape_merge},
--- a/test/testdata/infer/tuple_plus.rb
+++ b/test/testdata/infer/tuple_plus.rb
@@ -1,0 +1,27 @@
+# typed: true
+extend T::Sig
+
+def example1
+  xs = []
+  T.reveal_type(xs) # error: `[] (0-tuple)`
+  xs += [1]
+  T.reveal_type(xs) # error: `[Integer(1)] (1-tuple)`
+  xs += [2]
+  T.reveal_type(xs) # error: `T::Array[Integer]`
+end
+
+sig {params(ys: T::Array[Integer]).void}
+def example2(ys)
+  xs = []
+  T.reveal_type(xs) # error: `[] (0-tuple)`
+  xs += ys
+  T.reveal_type(xs) # error: `T::Array[Integer]`
+end
+
+sig {params(ys: T::Array[Integer]).void}
+def example3(ys)
+  xs = []
+  T.reveal_type(xs) # error: `[] (0-tuple)`
+  ys += xs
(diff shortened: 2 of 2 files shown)


jez:

→ https://go/builds/bui_MMRqBLgsJmrUhm
→ https://go/builds/bui_MMRq1iL8cXssW5

core/types/calls.cc
```diff
@@ -3320,6 +3320,19 @@ class Tuple_concat : public IntrinsicMethod {
...
+class Tuple_plus : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        auto *tuple = cast_type<TupleType>(args.thisType);
+        ENFORCE(tuple != nullptr);
+
+        res = dispatchCallProxyType(gs, args.thisType.underlying(gs), args);
+        if (tuple->elems.empty() && args.args.size() == 1) {
+            res.returnType = args.args[0]->type;
+        }
```
froydnj, line 3332: I think this needs to take into account (with some testcases) what happens if there are errors in the dispatch.  e.g. with the current PR, a test added to `tuple_plus.rb` like:

```ruby
def example4
  xs = []
  T.reveal_type(xs)
  xs += "asdf"
  T.reveal_type(xs)
end
```

gives:

```
test/testdata/infer/tuple_plus.rb:31: Revealed type: [] (0-tuple) https://srb.help/7014
    31 |  T.reveal_type(xs)
          ^^^^^^^^^^^^^^^^^
  Got [] (0-tuple) originating from:
    test/testdata/infer/tuple_plus.rb:30:
    30 |  xs = []
               ^^

test/testdata/infer/tuple_plus.rb:32: Expected T::Enumerable[T.untyped] but found String("asdf") for argument arg0 https://srb.help/7002
    32 |  xs += "asdf"
                ^^^^^^
  Expected T::Enumerable[T.untyped] for argument arg0 of method Array#+:
    https://github.com/sorbet/sorbet/tree/master/rbi/core/array.rbi#L448:
     448 |        arg0: T::Enumerable[Elem],
                  ^^^^
  Got String("asdf") originating from:
    test/testdata/infer/tuple_plus.rb:32:
    32 |  xs += "asdf"
                ^^^^^^

test/testdata/infer/tuple_plus.rb:33: Revealed type: String("asdf") https://srb.help/7014
    33 |  T.reveal_type(xs)
          ^^^^^^^^^^^^^^^^^
  Got String("asdf") originating from:
    test/testdata/infer/tuple_plus.rb:32:
    32 |  xs += "asdf"
          ^^^^^^^^^^^^
```

which seems worse than the current behavior ([sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Adef%20example4%0A%20%20xs%20%3D%20%5B%5D%0A%20%20T.reveal_type%28xs%29%0A%20%20xs%20%2B%3D%20%22asdf%22%0A%20%20T.reveal_type%28xs%29%0Aend%0A)):

```
editor.rb:6: Revealed type: [] (0-tuple) https://srb.help/7014
     6 |  T.reveal_type(xs)
          ^^^^^^^^^^^^^^^^^
  Got [] (0-tuple) originating from:
    editor.rb:5:
     5 |  xs = []
               ^^

editor.rb:7: Expected T::Enumerable[T.untyped] but found String("asdf") for argument arg0 https://srb.help/7002
     7 |  xs += "asdf"
                ^^^^^^
  Expected T::Enumerable[T.untyped] for argument arg0 of method Array#+:
    https://github.com/sorbet/sorbet/tree/master/rbi/core/array.rbi#L448:
     448 |        arg0: T::Enumerable[Elem],
                  ^^^^
  Got String("asdf") originating from:
    editor.rb:7:
     7 |  xs += "asdf"
                ^^^^^^

editor.rb:8: Revealed type: T::Array[T.untyped] https://srb.help/7014
     8 |  T.reveal_type(xs)
          ^^^^^^^^^^^^^^^^^
  Got T::Array[T.untyped] originating from:
    editor.rb:7:
     7 |  xs += "asdf"
          ^^^^^^^^^^^^
Errors: 3
```

Actually, maybe even checking for no errors is not sufficient, because you can do things like:

```
def example5
  xs = []
  T.reveal_type(xs)
  xs += {}
  T.reveal_type(xs)
end
```

which is currently accepted ([sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Adef%20example4%0A%20%20xs%20%3D%20%5B%5D%0A%20%20T.reveal_type%28xs%29%0A%20%20xs%20%2B%3D%20%7B%7D%0A%20%20T.reveal_type%28xs%29%0Aend%0A)), but gives with this PR:

```
test/testdata/infer/tuple_plus.rb:31: Revealed type: [] (0-tuple) https://srb.help/7014
    31 |  T.reveal_type(xs)
          ^^^^^^^^^^^^^^^^^
  Got [] (0-tuple) originating from:
    test/testdata/infer/tuple_plus.rb:30:
    30 |  xs = []
               ^^

test/testdata/infer/tuple_plus.rb:33: Revealed type: {} (shape of T::Hash[T.untyped, T.untyped]) https://srb.help/7014
    33 |  T.reveal_type(xs)
          ^^^^^^^^^^^^^^^^^
  Got {} (shape of T::Hash[T.untyped, T.untyped]) originating from:
    test/testdata/infer/tuple_plus.rb:32:
    32 |  xs += {}
```

which is probably not correct.  But even still, Ruby rejects it:

```
        5: from /home/froydnj/ruby/bin/irb:23:in `<main>'
        4: from /home/froydnj/ruby/bin/irb:23:in `load'
        3: from /home/froydnj/ruby/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        2: from (irb):3
        1: from (irb):3:in `+'
TypeError (no implicit conversion of Hash into Array)
```

froydnj (changes requested):
I think the dispatch here needs to be robustified, regardless of whether the existing issues are tackled.

core/types/calls.cc
jez: > But even still, Ruby rejects it

If we want, we can add a special case in the `Tuple_plus` intrinsic to detect not only if the receiver is `Enumerable` but whether it has a `to_ary` method defined.

That plagues what I imagine is quite a few things in the standard library, so I'm inclined to do that some other time.

core/types/calls.cc
```diff
@@ -3320,6 +3320,25 @@ class Tuple_concat : public IntrinsicMethod {
...
 
+class Tuple_plus : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        auto *tuple = cast_type<TupleType>(args.thisType);
+        ENFORCE(tuple != nullptr);
+
+        auto specialEmptyTupleCase = tuple->elems.empty() && args.args.size() == 1 &&
+                                     (isa_type<TupleType>(args.args[0]->type) ||
+                                      Types::isSubType(gs, Types::arrayOfUntyped(), args.args[0]->type));
```
jez, line 3331: this should guarantee that arg0 is an array (and not even an array subclass, so we don't have to worry about doing something like getting the `appliedType->klass` from the receiver and doing an `alignBaseTypeArgs` or something.

froydnj (changes requested):
Bouncing this back; there are some test failures and I would like to see the testcases work before I sit down and walk through the logic in `calls.cc`.

jez:
https://github.com/sorbet/sorbet/issues/6345

^ this is making me rethink that this change is even a good change in the first place. As written currently, landing this PR will cause people to encounter this bug more frequently.

I think we likely want to flip our current behavior: I think we should let annotate `+` with the sig that `.concat` currently has and, and possibly annotate `.concat` with the sig that `+` currently has.

In the past we hadn't done this in an attempt to make certain kinds of error messages more understandable, which we might want to reconsider. Either way, this isn't urgent so I'm going to put it down for the moment.

Issue #6314, opened 2022-08-29
Incorrect `Loc` when using `:'...'` symbols in props


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%20%3C%20T%3A%3AStruct%0A%20%20prop%20%3A'foo'%2C%20Integer%0Aend%0A%0Aa%20%3D%20A.new%28foo%3A%200%29%0Aa.foo%20%3D%20'')

```ruby
# typed: true

class A < T::Struct
  prop :'foo', Integer
end

a = A.new(foo: 0)
a.foo = ''
```

#### Observed output

```
editor.rb:8: Assigning a value to oo that does not match expected type Integer https://srb.help/7002
     8 |a.foo = ''
                ^^
  Got String("") originating from:
    editor.rb:8:
     8 |a.foo = ''
                ^^
Errors: 1
```

#### Expected behavior

Notice how the error message says "Assigning a value to `oo`". It should say "Assigning a value to `foo`".

In debug builds, there's actually an ENFORCE that fails, but that doesn't happen in prod builds.

- - -

sreenath-tm:
I'm interested in contributing to this issue, so before I start working it, would you mind sparing your time explaining what the fix will be like and pointing me to some resources to get started.

jez:
👋 Hey @sreenath-tm I think that @maruth-stripe had already started working on this the day before you commented. Sorry for the mixup.

If you find another issue you'd like to work on, feel free to ask questions about it in the #internals channel at https://sorbet.org/slack

sreenath-tm:
Hi @jez no issues will keep looking for other issues that I can contribute to. Thanks for replying.

maruth-stripe:
closed by https://github.com/sorbet/sorbet/pull/6341

PR #6341, opened 2022-09-06
Incorrect Loc when using `:'...'`  for symbols

maruth-stripe:
The `Loc` for `Symbol`s was incorrect when the symbol contained quotation marks. In particular, `:'foo'` would map to `Loc {start: 1, end: 4}` as opposed to `Loc{start: 0, end: 6}` (indices for demonstration purposes, real offsets will be relative to file).

This would cause an `ENFORCE` to fail in debug builds, and incorrect symbol names to be output in error messages in release builds. We patch the location used to construct the symbol.

### Motivation

https://github.com/sorbet/sorbet/issues/6314

### Test plan

Tested with local fixtures
See included automated tests.


parser/Builder.cc
```diff
@@ -1541,8 +1541,9 @@ class Builder::Impl {
         if (collapseStringParts(parts)) {
             // only 1 child: String
             auto firstPart = parts.front().get();
+
             if (auto *s = parser::cast_node<String>(firstPart)) {
-                return make_unique<Symbol>(s->loc, s->val);
+                return make_unique<Symbol>(core::LocOffsets { static_cast<uint32_t>(begin->start()), static_cast<uint32_t>(end->end()) }, s->val);
```
jez:
```suggestion
                return make_unique<Symbol>(tokLoc(begin, end), s->val);
```

parser/Builder.cc
```diff
@@ -1541,8 +1541,10 @@ class Builder::Impl {
         if (collapseStringParts(parts)) {
             // only 1 child: String
             auto firstPart = parts.front().get();
+
             if (auto *s = parser::cast_node<String>(firstPart)) {
-                return make_unique<Symbol>(s->loc, s->val);
+                return make_unique<Symbol>(
+                    core::LocOffsets{static_cast<uint32_t>(begin->start()), static_cast<uint32_t>(end->end())}, s->val);
```
jez, line 1547:
```suggestion
                return make_unique<Symbol>(tokLoc(begin, end), s->val);
```

jez (changes requested):
Can you add tests? There are instructions in the README for how to add tests.

.vscode/settings.json
```diff
@@ -12,7 +12,23 @@
...
+    "typeinfo": "cpp",
+    "array": "cpp",
+    "locale": "cpp",
+    "memory": "cpp",
+    "thread": "cpp",
+    "tuple": "cpp",
+    "variant": "cpp",
+    "functional": "cpp",
+    "utility": "cpp",
+    "stdexcept": "cpp"
```
jez, line 31: I don't want to have to have every header file defined in the standard library listed here.

Does something like `"**/include/c++/**"` work?

If not, can you move these to your user settings?

I'm not sure what causes you to need these settings. This is the list of extensions I have installed, and VS Code highlights C++ stdlib header files perfectly fine for me.

Screen Shot 2022-09-07 at 12 29 25 PM

maruth-stripe: Oh wait whoops did not intend to commit this file, my mistake

fixture/bar.rb
```diff
@@ -0,0 +1,8 @@
+# typed: true
+
+class A < T::Struct
+  prop :foo, Integer
+end
+
+a = A.new(foo: 0)
+a.foo = ''
```
jez, line 8: What is this `fixture/` folder?

maruth-stripe: Ah good grief `git add .` mistake I'll patch asap

test/testdata/parser/quoted_symbol_bug.rb
```diff
@@ -0,0 +1,8 @@
+# typed: true
+
+class A < T::Struct
+  prop :'foo', Integer
+end
+
+a = A.new(foo: 0)
+a.foo = '' # error: Assigning a value to `'foo'` that does not match expected type `Integer`
```
jez, line 8: The error message still mentions the `'...'`, even though the user is assigning to `foo`, not `'foo'`.

Do you want to try to make it show ``value to `foo` that does not match`` in this error message?

maruth-stripe: ack, I'll TAL

test/testdata/parser/quoted_symbol_bug.rb
```diff
@@ -0,0 +1,8 @@
+# typed: true
```
jez, line 1: minor feedback:

- we don't usually put `_bug` in our test names 
- I would have personally put this in `test/testdata/rewriter/`, because the original crash was caused by an ENFORCE in `rewriter/` and the contents of this test relies on the `prop` DSL.

Up to you whether you'd like to change either of those things.

jez (commented):
This looks mostly good. Thanks!

test/testdata/parser/quoted_symbol_bug.rb
maruth-stripe: ack, will fix

rewriter/Prop.cc
```diff
@@ -233,6 +233,11 @@ optional<PropInfo> parseProp(core::MutableContext ctx, const ast::Send *send) {
         ENFORCE(ctx.locAt(sym->loc).exists());
         ENFORCE(!ctx.locAt(sym->loc).source(ctx).value().empty() && ctx.locAt(sym->loc).source(ctx).value()[0] == ':');
         ret.nameLoc = core::LocOffsets{sym->loc.beginPos() + 1, sym->loc.endPos()};
+        const auto nameValue = ctx.locAt(ret.nameLoc).source(ctx).value();
```
maruth-stripe, line 236: this is a bit disgusting, but just fixes the loc to ignore the quotes for the error message

Happy to extract the quoting conditional into a separate function or something, not sure what might be a good place for it though

maruth-stripe:
cc @froydnj thank you for ping btw

test/testdata/rewriter/quoted_symbol_error.rb
```diff
@@ -0,0 +1,10 @@
+# typed: true
+
+class A < T::Struct
+  prop :'foo', Integer
+  prop :"bar", Integer
```
jez, line 5:
```suggestion
  prop :"bar", Integer
  prop :"", Integer
  prop :"#{A.name}", Integer
```

jez: can you update the test to include these edge cases? I think that the Prop dsl currently ignores those so _probably_ the code passes as is with these suggestions but I'm not sure.

jez (approved):
thanks!

test/testdata/rewriter/quoted_symbol_error.rb
maruth-stripe: Not sure how to call the constructor on `A` with `:""` and `:"#{A.name}"`, currently tried the following and getting an error:
```
test/testdata/parser/quoted_symbol_error.rb:10: ERROR: Found unexpected error:
10: a = A.new(foo: 0, bar: 1, :"" => 2, :"#{A.name}" => 3)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Cannot call `A#initialize` with a `Hash` keyword splat because the method has required keyword parameters
```

jez: You can add a test to capture the error without doing anything extra and only passing foo and bar.

jez: (Basically just leaving that part of the test unchanged)

jez: Test files don’t have to be free of errors—capturing error behavior is the point of tests.

maruth-stripe: ack done

PR #6344, opened 2022-09-06
fix field locs on the fast path

froydnj:
### Motivation

I am getting bizarre results for these tests locally and want to make sure I'm not missing something fundamental.

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1186,6 +1186,7 @@ class SymbolDefiner {
 
             // We are on the fast path and there was a previous declaration.
             prior.data(ctx)->resultType = core::Types::todo();
+            prior.data(ctx)->addLoc(ctx, ctx.locAt(field.loc));
             return prior;
         }
 
--- a/test/testdata/lsp/field_edits.1.rbupdate.document-symbols.exp
+++ b/test/testdata/lsp/field_edits.1.rbupdate.document-symbols.exp
@@ -34,22 +34,22 @@
                     "kind": 14,
                     "range": {
                         "start": {
-                            "line": 1,
-                            "character": 10
+                            "line": 3,
+                            "character": 2
                         },
                         "end": {
-                            "line": 1,
-                            "character": 15
+                            "line": 3,
+                            "character": 7
                         }
                     },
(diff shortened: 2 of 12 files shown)


froydnj:
(I don't understand why we're getting such bad locations for these error messages)

test/testdata/lsp/field_edits.1.rbupdate.document-symbols.exp
```diff
@@ -34,22 +34,22 @@
...
-                            "line": 1,
-                            "character": 10
+                            "line": 3,
+                            "character": 2
                         },
                         "end": {
-                            "line": 1,
-                            "character": 15
+                            "line": 3,
+                            "character": 7
```
froydnj, line 42: If you look at the corresponding ranges before the edit, you can see that the new values make a lot more sense than the previous values:

https://github.com/sorbet/sorbet/blob/4f8cdde4e605d0530c8eb24924b1ba71e80fbdf1/test/testdata/lsp/field_edits.rb.document-symbols.exp#L34-L43

The relevant parts of the before file:

https://github.com/sorbet/sorbet/blob/4f8cdde4e605d0530c8eb24924b1ba71e80fbdf1/test/testdata/lsp/field_edits.rb#L1-L3

The relevant parts of the after file:

https://github.com/sorbet/sorbet/blob/4f8cdde4e605d0530c8eb24924b1ba71e80fbdf1/test/testdata/lsp/field_edits.1.rbupdate#L1-L4

We added the comment line, so the locs should move down one line as well.

namer/namer.cc
```diff
@@ -1186,6 +1186,7 @@ class SymbolDefiner {
 
             // We are on the fast path and there was a previous declaration.
             prior.data(ctx)->resultType = core::Types::todo();
+            prior.data(ctx)->addLoc(ctx, ctx.locAt(field.loc));
```
froydnj, line 1189: My understanding of this is that it should replace the canonical loc for the field, which should have been in the same file as the field we are finding now?

jez (approved):

PR #6346, opened 2022-09-07
Add a new autoloader_compatibility DSL to package-spec files

aadi-stripe:
Adds a new DSL method called `autoloader_compatibility` to `__package.rb` files

### Motivation

In the Stripe codebase, we are in the midst of a migration related to how we autoload code. In order to facilitate this migration, we need to add a new configuration to package files that is related to the compatibility of the code inside the package file with a given autoloader.

### Test plan


aadi-stripe:
Not necessary

PR #6348, opened 2022-09-07
add a testcase for moving methods across files in the same class

froydnj:
### Motivation

This is a testcase inspired by:

https://github.com/sorbet/sorbet/blob/4f8cdde4e605d0530c8eb24924b1ba71e80fbdf1/namer/namer.cc#L1429-L1433

We shouldn't have some of the errors that currently exist, but there's no good way to mark them as XFAIL or similar.

### Test plan

See included automated tests.


jez (approved):
> We shouldn't have some of the errors that currently exist, but there's no good way to mark them as XFAIL or similar.

Can we at least have a comment in the file (`TODO` or otherwise) describing what we would like to be the case?

PR #6349, opened 2022-09-07
Follow up to #6299 (Allow `requires_ancestor` to return `T.class_of` object)

egiurleo:
This is a follow up to #6299 to address @jez's comments.

### Motivation
There were a couple of oversights in my previous PR that made it possible to require singleton classes as ancestors:
- I incorrectly used the `singletonClass` method instead of `lookupSingletonClass`
- I didn't account for the fact that generics are not allowed to be required ancestors, and singleton classes can be generics

### Test plan
I added an automated test based on [Jake's comment](https://github.com/sorbet/sorbet/pull/6299/files#r960131829).

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1185,7 +1185,7 @@ private:
                         }
 
                         if constexpr (isMutableStateType) {
-                            symbol = argClass->symbol.asClassOrModuleRef().data(gs)->singletonClass(gs);
+                            symbol = argClass->symbol.asClassOrModuleRef().data(gs)->lookupSingletonClass(gs);
                         }
                     }
                 }
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1184,9 +1184,7 @@ private:
                             return;
                         }
 
-                        if constexpr (isMutableStateType) {
-                            symbol = argClass->symbol.asClassOrModuleRef().data(gs)->lookupSingletonClass(gs);
-                        }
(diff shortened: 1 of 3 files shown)


resolver/resolver.cc
```diff
@@ -1185,7 +1185,7 @@ class ResolveConstantsWalk {
                         }
 
                         if constexpr (isMutableStateType) {
```
jez, line 1187: You no longer need this `if` guard, because neither `.data(gs)` nor `lookupSingletonClass(gs)` require that `gs` be mutable.

test/testdata/resolver/requires_ancestor_unsupported_generics.rb
```diff
@@ -26,3 +26,34 @@ module M # error: `Test2::M` can't require generic ancestor `Test2::RA` (unsuppo
...
   end
 end
+
+module Test3
+  class RA
+    class << self
+      extend T::Sig
+      extend T::Generic
+
+      Elem = type_member
```
jez, line 36: Can you add another test that uses `type_template` and `def self.` instead of `class << self` + `type_member`? There are some subtle difference between `class << self; Elem = type_member; end` and `Elem = type_template` and I would like to have a test for both.

jez (approved):
Looks mostly good but I have two requests.

egiurleo:
@jez Thanks for the comments! Let me know if my changes look okay to you.

PR #6352, opened 2022-09-08
Fix private_class_method signature

gustavodegeus:
Based on [ruby 3.X docs](https://ruby-doc.org/core-3.1.2/Module.html#method-i-private_class_method), this [example should](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20FooBar%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20self.foo%0A%20%20%20%20%22foo%22%0A%20%20end%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20self.bar%0A%20%20%20%20%22foo%22%0A%20%20end%0A%0A%20%20private_class_method%28%5B%3Afoo%2C%20%3Abar%5D%29%0Aend%0A) work.

The part that I'm not sure is that I believe this signature won't work for Ruby 2.X.
Do we have different rbi's for different ruby versions?

### Motivation
`private_class_method` should be able to receive an array of `String` or `Symbol` based on the ruby 3.X specs.

### Test plan
I've update the automated tests to include the new scenarios but I'm not able to run them locally. Last time I asked it was an issue with M1. 😢


test/testdata/rbi/module.rb
```diff
@@ -7,3 +7,7 @@
 T.assert_type!(Module.const_source_location('Zlib'), T.nilable([String, Integer]))
 T.assert_type!(Module.const_source_location('Zlib', true), T.nilable([String, Integer]))
 T.assert_type!(Module.const_source_location(:Zlib, false), T.nilable([String, Integer]))
+T.assert_type!(Module.private_class_method([:foo, :bar]), T::Array[Symbol])
+T.assert_type!(Module.private_class_method(["foo", "bar"]), T::Array[String])
+T.assert_type!(Module.private_class_method(:foo), Symbol)
+T.assert_type!(Module.private_class_method("bar"), String)
```
froydnj, line 13: According to the sig, `private_class_method` returns `Module`, so these `assert_type!` calls need to be modified to assert for `Module`.

froydnj:

→ https://go/builds/bui_MOc7hhKKOPnLDI
→ https://go/builds/bui_MOc7dSxfGxzmSP

froydnj (commented):
Thanks for the tests, and for updating them!

These changes are fine, but unfortunately they are only half of the story. Running Sorbet over the following with the changes from this PR:

```ruby
# typed: true
class FooBar
  extend T::Sig

  sig {void}
  def self.foo
    "foo"
  end

  sig {void}
  def self.bar
    "foo"
  end

  private_class_method :foo
  private_class_method([:bar])
end

# These should both error, but only the first one does.
FooBar.foo
FooBar.bar
```

Sorbet reports the following:

```
froydnj@pyro:~/src/sorbet$ ./bazel-bin/main/sorbet test/testdata/namer/private_class_method_array.rb
test/testdata/namer/private_class_method_array.rb:19: Non-private call to private method foo on T.class_of(FooBar) https://srb.help/7031
    19 |FooBar.foo
        ^^^^^^^^^^
    test/testdata/namer/private_class_method_array.rb:6: Defined in T.class_of(FooBar) here
     6 |  def self.foo
          ^^^^^^^^^^^^
Errors: 1
```

Sorbet doesn't yet understand how to process `private_class_method([:bar])` to apply the appropriate method modifiers to `:bar`.  The code that would deal with that starts here:

https://github.com/sorbet/sorbet/blob/48877f8a29cebb32e0de2b975c92872d812738aa/namer/namer.cc#L245-L252

I'll try to take a look at modifying that code to support the new array syntax next week.

gustavodegeus:
Hey @froydnj thanks for the follow up.

Right I see what you mean. We'd need to update the parsing to it can raise properly.
I'm having issues to run tests locally still but I will try to take a look on that as well.

Thanks!

froydnj (approved):
I committed the required code to `namer.cc`, and I'll handle writing up the testcases I pointed out above.

Thanks for the PR!

gustavodegeus:
Thanks @froydnj !!!!

Really appreciate the help to close this on up!!

PR #6353, opened 2022-09-08
free found definitions in parallel in namer

froydnj:
### Motivation

The code is kind of gross, but we can parallelize the freeing of all the found definitions during tree symbolificaton rather than doing it while we populate the global symbol table.

The proof-of-concept where we just leaked the found definitions took about 5% off the time required to populate the symbol table on Stripe's codebase, so ideally actually doing the work in parallel should realize approximately the same speedup.

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1354,7 +1354,7 @@ class SymbolDefiner {
                 break;
             }
             case core::FoundDefinitionRef::Kind::Field: {
-                const auto &field = ref.field(foundDefs);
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_MOZzsguQF4VwCT
→ https://go/builds/bui_MOZz7fahIuUB7E

froydnj:

→ https://go/builds/bui_MOa55eGmRnMUk2
→ https://go/builds/bui_MOa5vTNqmQahZd

PR #6355, opened 2022-09-08
Add system_openssl_repository rule

jez:
### Motivation

This means we don't have to hard-code the path on disk to `openssl@1.1`.

This unblocks getting the macOS build working again.

### Test plan

I've tested locally on an Intel mac, and in this build on Buildkite:

<https://buildkite.com/sorbet/sorbet/builds/24559#01831dfc-743f-4372-a8a7-bbb39f1fc249>


froydnj (approved):
:tada:

PR #6359, opened 2022-09-09
add an explanatory note for namespacing behavior errors

froydnj:
### Motivation

I didn't really understand the way forward with this error when I was asked about it by a user at Stripe; ideally this note will point them in the right direction.

### Test plan

See included automated tests.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -836,13 +836,19 @@ public:
             return;
         }
         auto &pkgName = requiredNamespace(ctx);
-        if (namespaces.packageForNamespace() != pkg.mangledName()) {
+        auto packageForNamespace = namespaces.packageForNamespace();
+        if (packageForNamespace != pkg.mangledName()) {
             ENFORCE(errorDepth == 0);
             errorDepth++;
             if (auto e = ctx.beginError(loc, core::errors::Packager::DefinitionPackageMismatch)) {
                 e.setHeader(
                     "Class or method behavior may not be defined outside of the enclosing package namespace `{}`",
                     fmt::map_join(pkgName, "::", [&](const auto &nr) { return nr.show(ctx); }));
+                if (packageForNamespace.exists()) {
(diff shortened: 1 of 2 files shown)


packager/packager.cc
```diff
@@ -836,13 +836,19 @@ class EnforcePackagePrefix final {
...
+        if (packageForNamespace != pkg.mangledName()) {
             ENFORCE(errorDepth == 0);
             errorDepth++;
             if (auto e = ctx.beginError(loc, core::errors::Packager::DefinitionPackageMismatch)) {
                 e.setHeader(
                     "Class or method behavior may not be defined outside of the enclosing package namespace `{}`",
                     fmt::map_join(pkgName, "::", [&](const auto &nr) { return nr.show(ctx); }));
+                if (packageForNamespace.exists()) {
+                    auto &namespaceParts = ctx.state.packageDB().getPackageInfo(packageForNamespace).fullName();
+                    e.addErrorNote("Attempting to define class or method behavior in package namespace `{}`",
```
jez, line 849: Do you think it would be possible to use addErrorLine instead of addErrorNote to make it point at the definition of Root in addition to the message?

jez (approved):

packager/packager.cc
froydnj: I can find the location of the package declaration, but I'm not sure that it improves the error message much.  e.g. for the tests we have, you (with this PR) get:

```
  Note:
    Attempting to define class or method behavior in package namespace `Root`
```

With the loc of the package declaration, you get:

```
    __package.rb:3: Attempting to define class or method behavior in package namespace `Root`
     3 |class Root < PackageSpec
```

which is...helpful?

The desirable thing would be (and maybe this is what you were suggesting) is to point at the enclosing module (which is representing the package we're not supposed to be defining behavior in) for the actual loc being passed to `checkBehaviorLoc`.  I'm not sure how to easily track that in this code, but I do agree that change would be a definite improvement.  I'll file an issue.

jez: Sorry yes I was thinking of the enclosing module.

PR #6368, opened 2022-09-13
reuse state across multiple consecutive `SymbolDefiner` invocations

froydnj:
### Motivation

Each run of `SymbolDefiner` (tens of thousands on Stripe's codebase) allocates a separate vector of defined classes and methods.  We can do better than this by wrapping those up into a `State` object that are re-used across run.

This shaves about ~1% off of symbol definition time during namer.  But more importantly, we have a bug with the LSP fast path where methods and fields moving across files + changes in their definitions can result in inconsistencies.  This framework gives us a way to address that, since we can potentially save state from individual `SymbolDefiner` runs.

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1480,7 +1480,7 @@ public:
     SymbolDefiner(State state, const core::FoundDefinitions &foundDefs, optional<core::FoundDefHashes> oldFoundHashes)
         : state(move(state)), foundDefs(foundDefs), oldFoundHashes(move(oldFoundHashes)) {}
 
-    void run(core::MutableContext ctx) {
+    SymbolDefiner::State run(core::MutableContext ctx) {
         state.definedClasses.reserve(foundDefs.klasses().size());
         state.definedMethods.reserve(foundDefs.methods().size());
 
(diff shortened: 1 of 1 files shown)


jez (commented):
This seems like a slightly suspicious refactoring on its own—if we're saying that there is only one `definedClasses` vector, then the indices of the `FoundClass` definitions will be wrong. Currently those indices assume that the vector is reset to empty after each file.

Given that, it seems to me like this change is only half-finished

froydnj:
> This seems like a slightly suspicious refactoring on its own—if we're saying that there is only one `definedClasses` vector, then the indices of the `FoundClass` definitions will be wrong. Currently those indices assume that the vector is reset to empty after each file.

The vector(s) are reset to empty after each file in this change?  There is still only one `definedClasses` per file (I'm not sure how one would make it work with a global `definedClasses` vector); this change just means that we reuse the underlying allocations for `definedClasses` and `definedMethods`.

> Given that, it seems to me like this change is only half-finished

To be clear: this change does not implement the fix to delete all old methods/fields prior to defining all new methods/fields.  Implementing that part seemed more complex than this change, and this felt like a useful stepping stone to the desired endpoint.

jez (approved):

PR #6369, opened 2022-09-13
define toplevel instance and class variables on the correct scope

froydnj:
### Motivation

Fixes a debug enforce after #6278.

### Test plan

The included testcase was generated at the commit prior to #6278, which is where the crash was introduced.  It's not obvious to me that Sorbet is defining the variables in the "right" place for Ruby's runtime behavior, but at least we're not regressing things.


jez (approved):

jez:
> The included testcase was generated at the commit prior to #6278, which is where the crash was introduced

Thanks! This was going to be my only question.

PR #6372, opened 2022-09-14
fix issues with interleaving deletion and definition in incremental namer

froydnj:
### Motivation

Correctness in the LSP fast path.

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1511,6 +1511,10 @@ public:
     }
 
     void finishIncrementalDefinition(core::MutableContext ctx) {
+        // We have to defer defining "deletable" symbols until the "finish" phase of
+        // incremental namer so that we don't delete and immediately re-enter a
+        // symbol (possibly keeping it alive, if it had multiple locs at the time
+        // of deletion) before SymbolDefiner has had a chance to process _all_ files.
         for (auto &method : foundDefs.methods()) {
             if (method.arityHash.isAliasMethod()) {
                 // We need alias methods in the FoundDefinitions list not so that we can actually
@@ -2239,7 +2243,10 @@ ast::ParsedFilesOrCancelled defineSymbols(core::GlobalState &gs, vector<SymbolFi
             }
 
             auto fref = fileFoundDefinitions.tree.file;
-            // The contents of this don't matter for incremental definition.
+            // The contents of this don't matter for incremental definition.  The
+            // old definitions should only matter when deleting old symbols (which
+            // happened in the previous phase of incremental namer), not here when
+            // entering symbols.
             optional<core::FoundDefHashes> oldFoundHashes;
             core::MutableContext ctx(gs, core::Symbols::root(), fref);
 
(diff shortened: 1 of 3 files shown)


namer/namer.cc
```diff
@@ -1508,7 +1499,18 @@ class SymbolDefiner {
...
+    SymbolDefiner(State state, const core::FoundDefinitions &foundDefs, optional<core::FoundDefHashes> oldFoundHashes)
+        : state(move(state)), foundDefs(foundDefs), oldFoundHashes(move(oldFoundHashes)) {}
+
+    SymbolDefiner::State beginIncrementalDefinition(core::MutableContext ctx) {
+        beginIncrementalDefinitionInternal(ctx);
+        return move(state);
+    }
+
+    void finishIncrementalDefinition(core::MutableContext ctx) {
         for (auto &method : foundDefs.methods()) {
```
jez, line 1514:
```suggestion
        // We have to defer defining "deletable" symbols until the "finish" phase of incremental
        // namer so that we don't delete and immediately re-enter a symbol (possibly keeping it
        // alive, if it had multiple locs at time of deletion) before SymbolDefiner has had a chance
        // to process _all_ files.
        for (auto &method : foundDefs.methods()) {
```

namer/namer.cc
```diff
@@ -2209,13 +2217,37 @@ ast::ParsedFilesOrCancelled defineSymbols(core::GlobalState &gs, vector<SymbolFi
...
+        ENFORCE(incrementalDefinitions.size() == allFoundDefinitions.size());
+        count = 0;
+        for (auto &fileFoundDefinitions : allFoundDefinitions) {
+            count++;
+            if (count % 250 == 0 && epochManager.wasTypecheckingCanceled()) {
+                return ast::ParsedFilesOrCancelled::cancel(move(output), workers);
+            }
+
+            auto fref = fileFoundDefinitions.tree.file;
+            // The contents of this don't matter for incremental definition.
```
jez, line 2242:
```suggestion
            // The contents of this don't matter for incremental definition. The old definitions
            // should only matter when deleting old symbols (which happens in the "begin" phase of
            // incremental namer), not here when entering new symbols.
```

namer/namer.cc
```diff
@@ -1477,23 +1477,14 @@ class SymbolDefiner {
         }
     }
 
-public:
-    SymbolDefiner(State state, const core::FoundDefinitions &foundDefs, optional<core::FoundDefHashes> oldFoundHashes)
-        : state(move(state)), foundDefs(foundDefs), oldFoundHashes(move(oldFoundHashes)) {}
-
-    SymbolDefiner::State run(core::MutableContext ctx) {
+    void beginIncrementalDefinitionInternal(core::MutableContext ctx) {
```
jez, line 1480: Slight preference to give some more descriptive names for these functions, to give a mental model to future developers whether to put new logic in `begin` or in `finish`.

jez (approved):

namer/namer.cc
froydnj: I split things up into three parts so that the "begin" phase didn't get a gigantic name.

PR #6373, opened 2022-09-14
add support for ruby 3-style array values for method modifiers

froydnj:
### Motivation

cc @gustavodegeus

#6352 adds support for typing `private_class_method([:foo])`, which is newly supported in Ruby 3.  We also need corresponding changes to namer to recognize this new syntax.

### Test plan

There are no tests added here; I will follow up after #6352 lands with tests.  I have confirmed locally that the testcase I suggested there correctly emits the expected errors with #6352 applied on top of this branch.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -239,14 +239,21 @@ public:
 
     void addMethodModifiers(core::Context ctx, core::NameRef modifierName,
                             absl::Span<const ast::ExpressionPtr> sendArgs) {
-        for (auto &arg : sendArgs) {
-            if (auto *array = ast::cast_tree<ast::Array>(arg)) {
+        if (sendArgs.empty()) {
+            return;
+        }
+
+        if (sendArgs.size() == 1) {
+            if (auto *array = ast::cast_tree<ast::Array>(sendArgs[0])) {
                 for (auto &e : array->elems) {
                     addMethodModifier(ctx, modifierName, e);
                 }
-            } else {
-                addMethodModifier(ctx, modifierName, arg);
+                return;
             }
+        }
+
+        for (auto &arg : sendArgs) {
+            addMethodModifier(ctx, modifierName, arg);
         }
     }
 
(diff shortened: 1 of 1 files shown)


namer/namer.cc
```diff
@@ -237,6 +237,19 @@ class SymbolFinder {
...
+        for (auto &arg : sendArgs) {
+            if (auto *array = ast::cast_tree<ast::Array>(arg)) {
+                for (auto &e : array->elems) {
+                    addMethodModifier(ctx, modifierName, e);
+                }
+            } else {
+                addMethodModifier(ctx, modifierName, arg);
+            }
+        }
+    }
```
jez, line 251: This means that

```ruby
private_class_method([:foo, :bar], :qux, :zap)
```

will be allowed. Is that also allowed by the Ruby VM? Or does it distinguish a single positional array argument?

jez (approved):

namer/namer.cc
froydnj: Good callout.  The [Ruby 3 documentation for `Module.public_class_method`](https://docs.ruby-lang.org/en/3.1/Module.html#method-i-public_class_method) says that arrays are only accepted as the lone argument, so this PR would need to be changed.

...but it also seems expressing that in the rbi file is going to be annoying, since I don't think we have a way to distinguish single-argument overloads from rest-argument overloads.  I guess something like:

```ruby
  sig do
    params(
        arg0: T.any(T::Array[Symbol], T::Array[String], Symbol, String),
        rest: T.any(Symbol, String),
    )
    .returns(T.self_type)
  end
  def private_class_method(arg0, *rest); end
```

would work as a hack, but that's not quite right.  Maybe we should write the above sig, and then verify in `calls.cc` that if an array is the first argument, that's the only argument we have?

jez: I don't mind making the sig overly broad, feel free to do whatever you think is best or easiest there. But I would prefer to not mark a method private in the symbol table that is not actually private.

jez: If you would like to emit a custom error in namer for when the arguments don't match, that's fine. I'm also fine with silently dropping the extra args for the sake of marking methods private.

PR #6375, opened 2022-09-14
add cfunc fastpaths to `sorbet_ruby`

froydnj:
### Motivation

Make as many things as possible go as fast as possible.

(This patch makes the benchmarks for `rake bench:typecheck` ~5-20% faster, depending on the benchmark.  And the cool thing is that it's not limited to just typechecking, but any method implemented in C taking fixed args gets faster.)

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_MQtHlIisZREXBp
→ https://go/builds/bui_MQtHzdu5FyBpJo

PR #6377, opened 2022-09-14
Upgrade to Bazel 5.2

jez:
### Motivation

New changes:

- <https://blog.bazel.build/2022/03/24/bazel-5.1.html>
- <https://blog.bazel.build/2022/06/08/bazel-5.2.html>

There have been a handful of improvements to Apple Silicon support
recent bazel versions, so I figure we may as well upgrade sooner rather
than later, because I don't think this breaks anything in the mean time.

### Test plan

- CI
- I tested that a debug build of Sorbet compiles and runs on my M1 machine (though it's still using the x86_64 version of Bazel, and generating x86_64 binaries).


froydnj (approved):
Does this cause weird interactions when compiling within pay-server?

jez:
> Does this cause weird interactions when compiling within pay-server?

Nope. We've ~always upgraded our Bazel version before pay-server has upgraded.

It _might_ cause problems if we start to change the builds files in ways that depend on e.g. new options that only exist in later versions, but we can cross that bridge when we get there.

jez:

→ https://go/builds/bui_MQwpUIgVxWqXMs
→ https://go/builds/bui_MQwpd3D6vWG7wK

abitrolly:
Why not 5.3? https://blog.bazel.build/2022/08/23/bazel-5.3.html

abitrolly:
Not sure how to do this. https://docs.bazel.build/versions/5.3.0/updating-bazel.html tells to use Baselisk instead and doesn't explain which commands to run.

PR #6378, opened 2022-09-15
Remove srb init snapshot tests

jez:
### Motivation

This makes it so that people have fewer problems when make RBI changes to
Sorbet.

### Test plan

CI


froydnj (approved):
Excellent +/- ratio in this PR.

PR #6382, opened 2022-09-15
miscellaneous cleanups

froydnj:
### Motivation

Pulling in some straightforward changes from #5228 that matter when you're compiling for arm64.

Ideally @vinistock gets credit for this when it gets squashed and committed.

### Test plan

See included automated tests.

### DIFF
--- a/common/os/mac.cc
+++ b/common/os/mac.cc
@@ -38,10 +38,8 @@ bool amIBeingDebugged()
 // Returns true if the current process is being debugged (either
 // running under the debugger or has a debugger attached post facto).
 {
-    int junk;
     int mib[4];
     struct kinfo_proc info;
-    size_t size;
 
     // Initialize the flags so that, if sysctl fails for some bizarre
     // reason, we get a predictable result.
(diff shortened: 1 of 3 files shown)


jez (approved):

vinistock (approved):
🎉

common/os/mac.cc
```diff
@@ -58,9 +56,8 @@ bool amIBeingDebugged()
 
     // Call sysctl.
 
-    size = sizeof(info);
-    junk = sysctl(mib, sizeof(mib) / sizeof(*mib), &info, &size, nullptr, 0);
-    assert(junk == 0);
+    size_t size = sizeof(info);
+    assert(sysctl(mib, sizeof(mib) / sizeof(*mib), &info, &size, nullptr, 0) == 0);
```
froydnj, line 60: Ahhh...realized this is not correct in the general case, because when assertions are disabled (e.g. in an opt build), the `sysctl` doesn't get executed.  I think this needs to be solved a different way.

PR #6386, opened 2022-09-15
RFC: Allow instance variable declaration inside `Minitest::Test#setup` method

egiurleo:
This is a proof of concept from the Ruby Developer Experience Team at Shopify! Feedback welcome. (cc: @dirceu)

### Motivation

Currently, Sorbet doesn't allow non-nilable instance variables to be declared inside the `Minitest::Test#setup` method:

```
class Bar < Minitest::Test
  def setup
    @foo = T.let(42, Integer) # Error: The instance variable `@foo` must be declared inside `initialize` or declared `nilable`
  end
end
```

[[sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Amodule%20Minitest%0A%20%20class%20Test%0A%20%20%20%20def%20setup%3B%20end%0A%20%20end%0Aend%0A%0Aclass%20Bar%20%3C%20Minitest%3A%3ATest%0A%20%20def%20setup%0A%20%20%20%20%40foo%20%3D%20T.let%2842%2C%20Integer%29%0A%20%20end%0Aend)]

This forces us to use some workarounds that are not the right way to use Minitest:

https://github.com/Shopify/rbi-central/blob/main/gem/test/test_helper.rb#L142

## Implementation

We've attempted to solve this problem by modifying some behavior in the resolver, but we're not sure if there might be a better way to approach this issue. Feedback would be appreciated!

## Test Plan

See added tests.

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -543,6 +543,7 @@ NameDef names[] = {
     {"Autogen", "Autogen", true},
     {"Tokens", "Tokens", true},
     {"AccountModelMerchant", "AccountModelMerchant", true},
+    {"Minitest", "Minitest", true},
     {"Token", "Token", true},
 
     // used by the compiler
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2072,12 +2072,28 @@ class ResolveTypeMembersAndFieldsWalk {
 
             scope = ctx.owner.enclosingClass(ctx);
         } else {
+            scope = ctx.selfClass();
+
+            auto isMinitestTestClass = false;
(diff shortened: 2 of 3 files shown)


froydnj (changes requested):
Thanks for writing this up!

We recognize that this would reduce developer pain, but we don't think a PR like this is the right direction for Sorbet.

The blocking issue here is that this PR adds gem-specific code to resolver, and Sorbet's philosophy is that all gem-specific code belongs in the rewriter pass.  Otherwise, it's really easy for resolver to add a bunch of special cases for this gem or that gem, and that would make resolver more difficult to understand and modify.

We have other concerns about the specifics of the PR (how this gets reflected into documentation / error messages and what that means for people not using Minitest, potential issues with `before_setup`, performance), but those are all subordinate to gem-specific code being added in resolver.

PR #6388, opened 2022-09-15
Retry when yarn install fails

jez:
### Test plan

See included automated tests.

### DIFF
--- a/.buildkite/test-vscode-extension.sh
+++ b/.buildkite/test-vscode-extension.sh
@@ -10,7 +10,33 @@ apt-get install -y libxshmfence-dev libnss3-dev libatk1.0-0 libatk-bridge2.0-0 \
 
 pushd vscode_extension
 
-yarn
+# https://stackoverflow.com/a/8351489
+with_backoff() {
+  local attempts=5
+  local timeout=1 # doubles each failure
+
+  local attempt=0
+  while true; do
+    attempt=$(( attempt + 1 ))
+    echo "Attempt $attempt"
+    if "$@"; then
+      return 0
+    fi
+
+    if (( attempt >= attempts )); then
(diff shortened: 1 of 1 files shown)


.buildkite/test-vscode-extension.sh
```diff
@@ -10,7 +10,33 @@ apt-get install -y libxshmfence-dev libnss3-dev libatk1.0-0 libatk-bridge2.0-0 \
...
+
+  local attempt=0
+  while true; do
+    attempt=$(( attempt + 1 ))
+    echo "Attempt $attempt"
+    if "$@"; then
+      return 0
+    fi
+
+    if (( attempt >= attempts )); then
```
froydnj, line 26: Does this need a `$` prefix, or does the magic of the if make this get evaluated as an arithmetic expression?

froydnj (approved):
Looks good, just one question.

.buildkite/test-vscode-extension.sh
jez: It works as is (I copied this code directly from test-static-sanitized.sh)

jez: I think that in particular the magic of the `(( ... ))` not the `if` makes it work.

jez: Also, writing `$((...))` triggers this shellcheck warning:

https://www.shellcheck.net/wiki/SC2084

froydnj: Ah, I see, I was not aware of the difference between [arithmetic expansion](https://www.gnu.org/software/bash/manual/html_node/Arithmetic-Expansion.html#Arithmetic-Expansion) and plain [shell arithmetic](https://www.gnu.org/software/bash/manual/html_node/Shell-Arithmetic.html).  TIL.

PR #6396, opened 2022-09-16
Split apart VS Code extension crash reason

jez:
### Motivation

We had a monolithic crash reason. Stripe's codebase sees 2 million such crash
events per day. I highly suspect that most of those are what I'm calling the new
`WRAPPER_REFUSED_SPAWN` bucket.

Importantly, we don't have any references to the old "crash" metric, and none of
our existing dashboards were looking at it. So we should be able to ignore all
`crash` metrics as being entirely from old clients (instead of having to filter
out "only up to date clients" or something).

### Test plan

Tested on Stripe's codebase.

### TODO

- [x] Review comments
- [x] Send PR to Stripe's codebase for wrapper script
- [x] Double-check what happens if the new extension version runs somewhere with the old wrapper script.

### DIFF
--- a/vscode_extension/src/LanguageClient.ts
+++ b/vscode_extension/src/LanguageClient.ts
@@ -89,6 +89,7 @@ export default class SorbetLanguageClient implements ErrorHandler {
 
   // Contains the Sorbet process.
   private _sorbetProcess: ChildProcess | null = null;
+  private _sorbetProcessExitCode: number | null = null;
 
   // Tracks disposable subscriptions so we can clean them up when language client is disposed.
   private _subscriptions: { dispose: () => void }[] = [];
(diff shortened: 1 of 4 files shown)


vscode_extension/src/LanguageClient.ts
```diff
@@ -317,7 +323,15 @@ export default class SorbetLanguageClient implements ErrorHandler {
...
       this._updateStatus(ServerStatus.RESTARTING);
-      this._restart(RestartReason.CRASH);
+      let reason = RestartReason.CRASH_LC_CLOSED;
+      if (this._sorbetProcessExitCode === 11) {
+        // 11 is EAGAIN
+        reason = RestartReason.WRAPPER_REFUSED_SPAWN;
+      } else if (this._sorbetProcessExitCode === 143) {
+        // 143 = 128 + 15 and 15 is TERM signal
+        reason = RestartReason.FORCIBLY_TERMINATED;
+      }
```
froydnj, line 333: Are these the Linux values or the Mac values?  My Linux box agrees that `EAGAIN` is 11, but my Mac headers say that 11 is `EDEADLK` ("11 was EAGAIN" in a comment) and 35 is `EAGAIN`.

`SIGTERM` does appear to be the same on both.

(Citing for completeness, I'm not sure it's worth trying to be perfectly correct here.)

froydnj (approved):

vscode_extension/src/LanguageClient.ts
jez: interesting i didn't realize they were different. The only thing that actually matters here is the number that we choose in the wrapper script, which can be anything. I had chosen 11 because it was EAGAIN on linux and seemed like it would be unlikely to ever collide with any return value returned from sorbet itself.

Thanks for reminding me, I need to send a PR to Stripe's codebase to land that wrapper script.

PR #6398, opened 2022-09-19
Set URI::File parent class as URI::Generic

vinistock:
Set the parent class of `URI::File` to `URI::Generic`.

### Motivation

This matches the implementation in [Ruby](https://github.com/ruby/ruby/blob/5883bc7c0791de2ce5e8b22175aef07705f0c618/lib/uri/file.rb#L10). The lack of the parent sometimes causes trouble erroring with `Parent class re-defined from Object to URI::Generic`.

### Test plan

Just an RBI change.


jez:

→ https://go/builds/bui_MSjmI8ya4129bJ
→ https://go/builds/bui_MSjm6h15LAKyIk

jez (approved):

PR #6400, opened 2022-09-19
gracefully handle `nil` in `T::Types::Intersection#name`

froydnj:
### Motivation

This is a gross corner case that showed up as a result of testing #4568 on Stripe's codebase.  Due to this change:

https://github.com/sorbet/sorbet/pull/4568/files#diff-e9f3c08a4fe9dd4e5c911a3cc84b06a1ef3639bf3c3d101ea26b59cb609c287fR164

We wind up calling `#name` when doing equality checking on types.  This shows up as a problem for intersection types, as sorting between strings and `nil` is not well-defined.  (This behavior is not normally a problem, but certain tests in Stripe's codebase use runtime-generated classes with `sorbet-runtime` types, and such classes don't have a name.)

Union types already handle this by dropping `nil`:

https://github.com/sorbet/sorbet/blob/58983475a4feb864551b8023b10f908bdb508b07/gems/sorbet-runtime/lib/types/types/union.rb#L39-L42

I think that is a reasonable thing to do here as well.  This does mean that the change in #4568 would cause `T.any(Class.new, String)` to compare equal to `T.any(String, Class.new)` (resp. for `T.all` with this change), which is admittedly a little weird.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/types/intersection.rb
+++ b/gems/sorbet-runtime/lib/types/types/intersection.rb
@@ -20,7 +20,7 @@ module T::Types
 
     # overrides Base
     def name
-      "T.all(#{@types.map(&:name).sort.join(', ')})"
+      "T.all(#{@types.map(&:name).compact.sort.join(', ')})"
     end
 
     # overrides Base
--- a/gems/sorbet-runtime/test/types/types.rb
+++ b/gems/sorbet-runtime/test/types/types.rb
@@ -1484,5 +1484,16 @@ module Opus::Types::Test
         assert_equal(:sym, GenericSingletonChild.foo(:sym))
       end
     end
+
+    describe "names" do
+      it 'can name T.any with runtime-created classes' do
+        klass = Class.new
+        name = T.any(klass, String).name
+        assert_equal(name, "T.any(String)")
+
+        name = T.any(String, klass).name
+        assert_equal(name, "T.any(String)")
+      end
+    end
   end
 end
--- a/gems/sorbet-runtime/test/types/types.rb
+++ b/gems/sorbet-runtime/test/types/types.rb
@@ -1494,6 +1494,15 @@ module Opus::Types::Test
         name = T.any(String, klass).name
         assert_equal(name, "T.any(String)")
       end
+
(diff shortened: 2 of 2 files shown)


froydnj:
> This does mean that the change in #4568 would cause `T.any(Class.new, String)` to compare equal to `T.any(String, Class.new)` (resp. for `T.all` with this change), which is admittedly a little weird.

I think this is already true, though, given that:

https://github.com/sorbet/sorbet/blob/master/gems/sorbet-runtime/lib/types/types/base.rb#L163-L168

already checks the names for equality.  So union types that include runtime-generated classes with no names would just drop those classes from the string name; this PR would then provide the same behavior for intersection types as well.

jez:
> This does mean that the change in #4568 would cause `T.any(Class.new, String)` to compare equal to `T.any(String, Class.new)` (resp. for `T.all` with this change), which is admittedly a little weird.

When you say "compare equal to", do you mean for the purpose of `valid?` / subtyping checks? That comes as a surprise to me. Do we use `name` for anything other than debug messages? I thought we used other mechanisms to do the comparisons, and only used the name for type errors.

If this is not the case, can we add a test for that?

Alternatively, instead of showing `T.any(String)` and `T.all(String)`, what if we changed the `.compact` to map `nil` values to something like `"<anon class>"` or `"Class.new"`? (we could even detect if it's an anonymous class or a module if we want)

jez:
alright well that's bad. i guess if you try to use runtime anonymous classes in types then you're asking for pain.

jez (approved):

froydnj:
> Alternatively, instead of showing `T.any(String)` and `T.all(String)`, what if we changed the `.compact` to map `nil` values to something like `"<anon class>"` or `"Class.new"`? (we could even detect if it's an anonymous class or a module if we want)

I like this idea, but I think even with filling in placeholders you can potentially run into the same problem with e.g. `T.any(String, Class.new, Class.new)` and `T.any(Class.new, String, Class.new)`: you get string-ly equal names, so `#==` returns true, but they're not the same type.  :man_shrugging:

PR #6401, opened 2022-09-19
Set RDoc::Task parent to Rake::Tasklib

vinistock:
Set the parent of `RDoc::Task` to `Rake::Tasklib` and define `Rake::Tasklib` in a new `rake.rbi`.

I propose leaving the `rake.rbi` basically empty, since Tapioca can generate the types for it. We just need `Rake::Tasklib` to be defined within Sorbet's payload to use it as a parent class.

### Motivation

When Tapioca generates types for RDoc, it uses the [parent defined in Ruby](https://github.com/ruby/ruby/blob/26135312f61014967ef223dd16ad6577ebd9c5d8/lib/rdoc/task.rb#L96), which conflicts with Sorbet's stdlib payload and results in `Parent re-defined from Object to Rake::Tasklib` issues.

### Test plan

Just RBI changes.

### DIFF
--- a/rbi/gems/rake.rbi
+++ b/rbi/gems/rake.rbi
@@ -1,0 +1,9 @@
+# typed: __STDLIB_INTERNAL
+
+# The RDoc RBI from the stdlib references Rake::TaskLib, so we need to have the module and class defined. The actual
+# contents of the class can be generated automatically with Tapioca
+
+module Rake
+  class TaskLib
+  end
+end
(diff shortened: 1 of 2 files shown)


froydnj:

→ https://go/builds/bui_MSlTrHNz8lFHF0
→ https://go/builds/bui_MSlTucQw2B5urv

jez:

→ https://go/builds/bui_MSlV9dSMryahvb
→ https://go/builds/bui_MSlVKRI0YGGfKj

jez (approved):

PR #6411, opened 2022-09-20
Symbolize type_member assignments (and other fixes)

jez:
### Motivation

We were previously not converting `UnresolvedConstantLit` nodes to
`ConstantLit` nodes in the LHS of a `type_member` assignment.

This was being masked by `builder_walk.cc` calling
`unresolvedIdent2Local` to effectively do the constant resolution lazily in
CFG builder. However, that constant resolution didn't take into account
mangle renames, which meant that sometimes `X = type_member` would resolve
to a `class ::Wrapper::X` symbol if the `type_member` symbol had been
redefined later in the file.

This incorrect resolution caused an ENFORCE failure in
`builder_entry.cc` that asserted that the `aliases` list only contains 
static fields and type members (in the buggy version, it also could contain
class-or-module symbols).

### Commit summary

~~The first two commits are pre-work.~~

~~The third commit is the real change, and the last commit is a test that
previously crashed.~~

**Update**: in the process of reading the test code that failed from the first version of the PR, I discovered a bunch of other bugs in similar parts of the code. This PR now fixes all of those. I tried splitting the work into separate PRs, but the tests I was writing just kept hitting different ENFORCE failures if I tried to land pieces separately.

I have tried making it mostly easy to review by commit nonetheless if that helps.

- **pre-work: Just take one argument** (7dd940bfe)

  The others we can derive from the one.

- **pre-work: Move this snippet outside the `if`** (b2a6a105f)

  It doesn't actually depend on any of the `type_member`'s args.

  It was written where it was because also it wasn't previously required
  except inside that block, but that will change in the next commit.

- **Symbolize type_member assignments** (6fb96adb3)

  We were previously not converting `UnresolvedConstantLit` nodes to
  `ConstantLit` nodes in the LHS of a `type_member` assignment.

  This was being masked by `builder_walk.cc` calling
  `unresolvedIdent2Local` to effectively do the constant resolution lazily
  in CFG builder. However, that constant resolution didn't take into
  account mangle renames, which meant that sometimes `X = type_member`
  would resolve to a `class ::Wrapper::X` symbol if the `type_member`
  symbol had been redefined later in the file.

  This incorrect resolution caused an ENFORCE failure in
  `builder_entry.cc` that asserted that the `aliases` list only contains
  static fields and type members (in the buggy version, it also could
  contain class-or-module symbols).

- **Add a test** (881252d82)

- **Remove an error** (565ef8731)

  This error is spurious. It only used to appear because of how we decided
  to recover from the namer redefinition error.

- **Remove ENFORCE, because it wasn't true** (ec0285fc1)

- **Make findSymbols and symbolizeTrees symmetric** (f1ea33558)

  These two `postTransformAssign` methods are nearly identical (the only
  difference is that there is some extra code to handle instance variables
  in the `findSymbols` case--namer doesn't need to deal with instance
  variables in `symbolizeTrees`, so it's absent there).

  This is pre-work for the next commit.

- **pre-work: Add extra handling for weird scopes with `type_member`** (f2fc6ee50)

- **Add snapshot showing master behavior** (c4d9daabc)

- **Show new behavior** (12a2a635f)

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -2012,6 +2012,11 @@ public:
             return ast::MK::EmptyTree();
         }
 
+        // Simulates how squashNames in handleAssignment also creates a ConstantLit
+        // (simpler than squashNames, because type members are not allowed to use any sort of
+        // `A::B = type_member` syntax)
+        asgn.lhs = ast::make_expression<ast::ConstantLit>(asgn.lhs.loc(), sym, move(asgn.lhs));
+
         if (send->hasPosArgs() || send->hasKwArgs() || send->block() != nullptr) {
             if (send->numPosArgs() > 1) {
                 if (auto e = ctx.beginError(send->loc, core::errors::Namer::InvalidTypeDefinition)) {
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -584,8 +584,6 @@ public:
         } else {
             switch (send->fun.rawId()) {
                 case core::Names::typeTemplate().rawId():
-                    handleTypeMemberDefinition(ctx, send, asgn, lhs);
-                    break;
                 case core::Names::typeMember().rawId():
                     handleTypeMemberDefinition(ctx, send, asgn, lhs);
                     break;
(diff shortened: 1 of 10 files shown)


jez:

→ https://go/builds/bui_MTFyXIXJL9JqQJ
→ https://go/builds/bui_MTFy0TYcxowT03

namer/namer.cc
```diff
@@ -581,11 +581,11 @@ class SymbolFinder {
             fillAssign(ctx, asgn);
         } else if (!send->recv.isSelfReference()) {
             handleAssignment(ctx, asgn);
+        } else if (!ast::isa_tree<ast::EmptyTree>(lhs->scope)) {
+            handleAssignment(ctx, asgn);
```
froydnj, line 585: This hunk was surprising to me; it makes sense, but I also didn't see tests for non-empty scopes and `type_member`/`type_template` definitions.  Did we already have some?

Ah, I see you filed #6416.  👍

froydnj (approved):
Thank you for the commit-by-commit breakdown; that was very helpful.

PR #6418, opened 2022-09-21
Define all found classes first

jez:
### Motivation

This unblocks moving edits that change static fields and type members onto the
fast path.

### Commit summary

- **pre-work: Record the behavior of this new test on master** (006b505f9)

  Show master behavior and then show this branch behavior.

- **pre-work: Add master snapshot for fuzz_class_in_field test** (ce64fe83b)

- **Define all found classes first** (5a05b9a16)

  ... then define static fields and type members after that.

- **Make error reflect that classes always come first** (2d6c80ba9)

- **Show behavior of new test** (09506e9c5)

- **Update test expectations** (2afba2acd)

  Most of these are just wording changes of existing error messages, but a
  handful of tests (like `nested_static_field.rb`) actually change
  meaning. I think that the change is clear (maybe more clear than before)
  and only present in files that had errors previously, so it's fine.

  While in `nested_static_field.rb` the error feels out of order,
  overwhelmingly I think this error is going to be hit when two separate
  files combine to have the buggy code, so the confusion of
  seemingly-out-of-order execution will likely not be that common in
  practice.

- **Update a snapshot test** (be2465ab0)

  The behavior in the class_and_alias.rb file is showing what happens when
  an assignment like `X = 1` comes before or after a definition like
  `class Y; end`.

  Before, the behavior was that the first definition was always mangled.
  Now the behavior is that the class or module definition is always
  mangled.

- **Fix some fuzz_class_in_field tests** (3cda6b61d)

  The symbol table change looks bad, but when you realize that it's just
  two changes:

  - the `static-field` moves below the `module` declaration
  - the module declaration gets mangled

  This new behavior should be fine, because there's still a typed-false
  error somewhere, and the behavior is the same on the slow and fast path.

- **Update a cli test** (0b860997f)

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/namer/all_constant_redefinitions.rb
+++ b/test/testdata/namer/all_constant_redefinitions.rb
@@ -1,0 +1,30 @@
+# typed: strict
+# disable-fast-path: true
+
+module Wrapper
+  extend T::Generic
+
+  class A; end
+  A = 1 # error: Redefining constant `A` as a static field
+  A = type_member # error: Redefining constant `A` as a type member or type template
+
+  B = 1
+  class B; end # error: Redefining constant `B` as a class or module
+  B = type_member # error: Redefining constant `B` as a type member or type template
+
+  C = 1
+  C = type_member # error: Redefining constant `C` as a type member or type template
+  class C; end # error: Redefining constant `C` as a class or module
+
+  class D; end
+  D = type_member # error: Redefining constant `D` as a type member or type template
+  D = 1 # error: Redefining constant `D` as a static field
+
+  E = type_member
+  class E; end  # error: Redefining constant `E` as a class or module
+  E = 1 # error: Redefining constant `E` as a static field
+
+  F = type_member
+  F = 1 # error: Redefining constant `F` as a static field
+  class F; end # error: Redefining constant `F` as a class or module
+end
--- a/test/testdata/namer/all_constant_redefinitions.rb
+++ b/test/testdata/namer/all_constant_redefinitions.rb
@@ -1,30 +1,29 @@
 # typed: strict
-# disable-fast-path: true
 
 module Wrapper
   extend T::Generic
 
   class A; end
-  A = 1 # error: Redefining constant `A` as a static field
+  A = 1 # error: Cannot initialize the class `A` by constant assignment
   A = type_member # error: Redefining constant `A` as a type member or type template
 
-  B = 1
-  class B; end # error: Redefining constant `B` as a class or module
+  B = 1 # error: Cannot initialize the class `B` by constant assignment
+  class B; end
   B = type_member # error: Redefining constant `B` as a type member or type template
 
-  C = 1
+  C = 1 # error: Cannot initialize the class `C` by constant assignment
   C = type_member # error: Redefining constant `C` as a type member or type template
-  class C; end # error: Redefining constant `C` as a class or module
+  class C; end
 
   class D; end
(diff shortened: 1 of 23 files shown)


jez:

→ https://go/builds/bui_MTFPb9XFEWM6HK
→ https://go/builds/bui_MTFPhyLtvkDan9

test/testdata/namer/all_constant_redefinitions.rb
```diff
@@ -0,0 +1,29 @@
...
+  D = 1 # error: Redefining constant `D` as a static field
+
+  E = type_member # error: Redefining constant `E` as a type member or type template
+  class E; end
+  E = 1 # error: Redefining constant `E` as a static field
+
+  F = type_member # error: Redefining constant `F` as a type member or type template
+  F = 1 # error: Redefining constant `F` as a static field
+  class F; end
+end
```
jez, line 29: Note to self: I wanted to add an rbupdate test for this before merging.

test/testdata/namer/class_and_alias.rb
```diff
@@ -1,8 +1,8 @@
...
-A = 91
-class A # error: Redefining constant `A` as a class or module
+A = 91 # error: Cannot initialize the class `A` by constant assignment
+class A
 end
 
 class B
 end
-B = 91 # error: Redefining constant `B` as a static field
+B = 91 # error: Cannot initialize the class `B` by constant assignment
```
froydnj, line 8: This error doesn't feel particularly clear to me, especially when compared with the previous error in the `B` case.  I guess I don't mentally model `class A` as defining a constant, even though that's really what is happening in Ruby.

The extra context for the error message does clarify things somewhat, so maybe this is a nonissue.

Maybe "Cannot reassign constant `B` after it has already been declared as a class"?  I guess that gets awkward in the `A` case, but maybe you could have different messages depending on how the `Loc`s of the two definitions are ordered.  Maybe "Cannot declare `B` via constant assignment and a class definition"?

froydnj (approved):
LGTM.  The error message comment is non-blocking.

jez:
* master
  * **PR #6418** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6418" target="_blank">Graphite</a>  👈
    * **PR #6421** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6421" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6418?utm_source=stack-comment).

test/testdata/namer/class_and_alias.rb
jez: My thinking was that most of the time when this is happening, it'll have been something like

```ruby
# -- foo.rb --

Foo = dsl_that_generates_a_class()

# -- foo.rbi --

class Foo
  def method_that_was_missing; end
end
```

The old message here would have chosen one of the two files arbitrarily (probably sorting based on FileRef ID, which is loosely alphabetic by directory order) and said that the second definition redefines the first.

Now Sorbet will always think that `Foo` is a class, and say that you can't initialize `Foo` which is a class using the `dsl_that_generates_a_class()` method.

I agree there's probably room to further improve this message. While in this test file saying "cannot reassign" reads better, in the case where the conflict comes from separate files, it's not a reassignment.

Anyways, maybe we just land the stack as is and follow up with feedback people have about the error message in practice?

PR #6420, opened 2022-09-21
Changes to `type_parameters` take the fast path

jez:
### Motivation

All of these changes should have probably landed when we made change to methods
take the fast path.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1771,7 +1771,7 @@ void GlobalState::deleteMethodSymbol(MethodRef what) {
     for (const auto typeArgument : whatData->typeArguments()) {
         this->typeArguments[typeArgument.id()] = this->typeArguments[0].deepCopy(*this);
     }
-    // This drops the core::Method, which drops the ArgInfo's the method owned.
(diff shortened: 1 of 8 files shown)


core/GlobalState.cc
```diff
@@ -1768,6 +1768,10 @@ void GlobalState::deleteMethodSymbol(MethodRef what) {
     ENFORCE(fnd != ownerMembers.end());
     ENFORCE(fnd->second == what);
     ownerMembers.erase(fnd);
+    for (const auto typeArgument : whatData->typeArguments()) {
+        this->typeArguments[typeArgument.id()] = this->typeArguments[0].deepCopy(*this);
+    }
+    // This drops the core::Method, which drops the ArgInfo's the method owned.
```
froydnj, line 1774:
```suggestion
    // This drops the existing core::Method, which drops the `ArgInfo`s the method owned.
```

The "existing" bit you can take or leave; one can argue that it's kind of redundant with the actual semantics of the assignment, but I do think it helps clarify what `core::Method` is being dropped here.

I think the quoting around the `ArgInfo` is an improvement regardless.

froydnj (approved):

PR #6370, opened 2022-09-13
Add slow path with incremental resolver

ilyailya:
Also do a renaming

`canTakeFastPath` -> `getTypecheckingPath`

### Test plan

See included automated tests.

PR #6421, opened 2022-09-21
Smash all `vector<SymbolHash>`'s into one big vector

jez:
### Motivation

This is a net-negative diff on its own, and also lets us avoid writing
*more* code in these places when we move static fields and type members
onto the fast path.

While most of the changes are purely mechanical, the changes in
`LSPFileUpdates.cc` are slightly less obviously correct. We used to have
different set difference logic based on whether the boolean
`lspExperimentalFastPathEnabled` was enabled and whether the symbol kind
was static field or not.

But in the case where `validateIdenticalFingerprints` held (e.g., for
static fields), `c_set_difference` and `c_set_symmetric_difference` should
behave identically, especially considering that we de-dup the hashes after
the difference.

So while there's a slight difference in behavior, effectively there's no change.

**Note** that this is only the `localSymbolTableHashes`. The vectors in
`FoundDefHashes` are still split out as separate vectors, because there is
not a single `FoundDefHash` kind (instead, there are `FoundMethodHash` and
`FoundFieldHash` structures, which store different information). Also, the
process of deleting methods given their `FoundDefinition` hash requires
knowing the kind of symbol we're operating on. So those vectors will likely
stay as is for the time being.

### Test plan

See included automated tests.


froydnj (approved):
In our hashing function we trust to not result in a bunch of collisions here, I guess

jez (commented):
Yeah that's a good point. Might be worth measuring at some point, but when there are collisions, the worst thing that happens is we type check too many downstream files, and that's alleviated a little bit by

- the ongoing work in #6370
- our plans to eventually make the downstream set of files use packager information

Also for example, ~every field will start with `@` and every constant will start with an uppercase letter, which maybe makes it easier for our hash function to send inputs to very different hashes.

Issue #5685, opened 2022-04-24
Signatures for `attr_accessor` won't add runtime checks to the writer method


Qqwy:
#### Input

Given the following strictly-typed class:

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Arequire%20'sorbet-runtime'%0A%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20params%28bar%3A%20String%29.void%20%7D%0A%20%20def%20initialize%28bar%29%0A%20%20%20%20%40bar%20%3D%20bar%0A%20%20end%0A%0A%20%20sig%20%7B%20returns%28String%29%20%7D%0A%20%20attr_accessor%20%3Abar%0Aend)

```ruby
# typed: strict
class Foo
  extend T::Sig

  sig { params(bar: String).void }
  def initialize(bar)
    @bar = bar
  end

  sig { returns(String) }
  attr_accessor :bar
end
```

And using the Foo class from another piece of code (outside of the purview of Sorbet's static checks) or iRB/Pry:

```ruby
  foo = Foo.new("Hello, world")
  foo.bar = 42 # <- Strangely, no TypeError is raised! (e.g. when `Foo#bar=` is called)
  puts foo.bar # <- Instead, a TypeError is raised whenever `Foo#bar` is called afterwards
```

#### Observed output

The `foo.bar = 42`-line runs without an exception.
However, when `foo.bar` is subsequent read, a TypeError of the following kind is raised

```
TypeError: Return value: Expected type String, got type Integer with value 42
```

#### Expected behavior

The runtime error should be raised for `Foo#bar=(42)`, rather than allowing the invariants of instances of Foo to be silently broken.

It should be noted that Sorbet's static checks pick up on this problem. Obviously this only happens if the file/location/environment where this happens is part of the code which Sorbet is checking statically. 
I'm pretty sure that the runtime system ought to catch this problem for the places where the static checks cannot.

- - -

After encountering this error in my own codebase, I looked through Sorbet's code, and found [this test](https://github.com/sorbet/sorbet/blob/master/gems/sorbet-runtime/test/types/edge_cases.rb#L34-L36) containing a 'TODO' hinting at this problem as well. However, it does not seem like an issue was ever made from that TODO.

Qqwy:
One way to fix this, is to override `attr_accessor` in classes that contain `extend T::Sig` with the following:

```ruby
[...]

PR #6426, opened 2022-09-23
Make runtime type check `attr_accessor` writers

paracycle:
Sorbet runtime has never checked the writer for `attr_accessor` defined methods since that is the only case where a single signature needs to be matched up to two methods.

This PR adds an exception for the reader method of an `attr_accessor` triggered method added handling. For those cases, after we clear the current declaration, we create a new writer declaration with the correct params and make it the current declaration. That makes sure that the next method added handling, which should be for the writer method, can find and process a correct declaration.

Credit to @XrXr for giving me the original idea to check the `caller_location` label to see if we are in an `attr_accessor` call or not.

### Motivation

Fixes: https://github.com/sorbet/sorbet/issues/5685

### Test plan

Extended existing automated tests, and removed a `TODO`.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -204,6 +204,34 @@ module T::Private::Methods
     @modules_with_final[mod.singleton_class.object_id]
   end
 
+  private_class_method def self.attr_accessor_reader?(method_name)
+    # If the method name ends with a "=", it's definitely not an attr_accessor reader
+    return false if method_name[-1] == "="
+
+    # If it was "attr_accessor" that triggered this, we should be 3 frames away from it:
+    # 3. `attr_accessor`
+    # 2. method_added hook
+    # 1. `_on_method_added`
+    # 0. `attr_accessor_reader?`
+    caller_frame = caller_locations(3, 1).first
+    return false unless caller_frame
+
+    caller_label = caller_frame.label
+    return false unless caller_label
+
+    caller_label.to_s == "attr_accessor"
+  end
+
+  private_class_method def self.declare_writer_sig_from_reader(method_name, current_declaration)
+    writer_decl = current_declaration.dup
+    # The block for the reader runs the block for the writer,
+    # but also adds the writer param with the same type as the return type.
+    writer_decl.blk = lambda do
+      instance_exec(&current_declaration.blk).params(**{method_name.to_sym => decl.returns})
+    end
+    T::Private::DeclState.current.active_declaration = writer_decl
+  end
+
   # Only public because it needs to get called below inside the replace_method blocks below.
   def self._on_method_added(hook_mod, method_name, is_singleton_method: false)
     if T::Private::DeclState.current.skip_on_method_added
@@ -228,6 +256,8 @@ module T::Private::Methods
       return
     end
     T::Private::DeclState.current.reset!
+
+    declare_writer_sig_from_reader(method_name, current_declaration) if attr_accessor_reader?(method_name)
 
     if method_name == :method_added || method_name == :singleton_method_added
       raise(
--- a/gems/sorbet-runtime/test/types/edge_cases.rb
+++ b/gems/sorbet-runtime/test/types/edge_cases.rb
@@ -21,19 +21,37 @@ class Opus::Types::Test::EdgeCasesTest < Critic::Unit::UnitTest
       extend T::Helpers
       sig {params(foo: String).returns(String)}
       attr_writer :foo
-      sig {params(bar: Integer).returns(Integer)}
+      sig {returns(Integer)}
       attr_accessor :bar
     end
 
+    # Verify that attr_writer works on the happy path
     assert_equal("foo", klass.new.foo = "foo")
+
+    # Verify that attr_writer does type checking
     err = assert_raises(TypeError) do
       klass.new.foo = 42
     end
     assert_match(/Expected type String, got type Integer/, err.message)
-    assert_equal(42, klass.new.bar = 42)
-    # TODO: This should also raise a type error, but currently doesn't because
-    # the sig only affects the reader part of the attr_accessor, not the writer.
-    assert_equal("foo", klass.new.bar = "foo")
+
+    # Verify that the reader and writer for attr_accessor work on the happy path
+    obj = klass.new
+    assert_equal(42, obj.bar = 42)
+    assert_equal(42, obj.bar)
+
+    # Verify that the writer for attr_accessor does type checking
+    err = assert_raises(TypeError) do
+      klass.new.bar = "bar"
+    end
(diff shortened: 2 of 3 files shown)


gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -227,7 +236,7 @@ def self._on_method_added(hook_mod, method_name, is_singleton_method: false)
     if current_declaration.nil?
       return
     end
-    T::Private::DeclState.current.reset!
+    T::Private::DeclState.current.reset! unless attr_accessor_reader?(method_name)
```
paracycle, line 239: I chose to just leave the same sig for the writer method, but I guess I can alternatively, clear the sig unconditionally and add a duped and modified version of the `current_declaration` as `current` if this is an `attr_accessor` reader.

That would actually simplify the processing we need to do later in `Signature#initialize`. Happy to make the change if you think it makes more sense.

gems/sorbet-runtime/test/types/edge_cases.rb
```diff
@@ -21,19 +21,37 @@ def ==(other)
       extend T::Helpers
       sig {params(foo: String).returns(String)}
       attr_writer :foo
-      sig {params(bar: Integer).returns(Integer)}
+      sig { returns(Integer)}
```
paracycle, line 24: This `sig` had always been wrong, since the documentation clearly states to use the reader form for accessor sigs.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
jez: Yeah, I think I would prefer that we create that signature here. It's a little more explicit and means that there's only one heuristic in play (the `attr_accessor` in the call stack) and not also "the signature is technically ill-formed but looks like a writer method".

jez (commented):
This looks neat, thanks! I'll have to test it on Stripe's codebase to make sure. I'd prefer to do that after the implementation is adjusted to use the alternate `current_declaration` strategy you mentioned.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
```diff
@@ -204,6 +204,15 @@ def self.note_module_deals_with_final(mod)
     @modules_with_final[mod.singleton_class.object_id]
   end
 
+  private_class_method def self.attr_accessor_reader?(method_name)
+    # If it was "attr_accessor" that triggered this, we should be 3 frames away from it:
+    # 3. `attr_accessor`
+    # 2. method_added hook
+    # 1. `_on_method_added`
+    # 0. `attr_accessor_reader?`
+    caller_locations(3, 1).first&.label&.to_s == "attr_accessor" && !method_name.to_s.end_with?("=")
```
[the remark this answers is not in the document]
paracycle: Initially I just had the `end_with?` but had a failure on Sorbet CI. I then remembered that [`Symbol#end_with?` was only added in 2.7](https://github.com/ruby/ruby/commit/b5fbefbf2c14742f6d46ecdf3ce712062dfb1d0a), and I think Sorbet CI still uses 2.6 for some cases, so I converted to a string.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
paracycle: Ok made the change, but since we don't want to run the decl block here, i've instead chosen to create a cloned decl block whose `blk` is a proc that calls the `blk` of the current decl and then adds the params as it should be for the writer.

I am not sure if this is the best or most efficient way of doing this. Please let me know if it is not.

gems/sorbet-runtime/lib/types/private/methods/_methods.rb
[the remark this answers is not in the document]
paracycle: In retrospect, I reverted to using `method_name[-1] == "="` which is equally readable and works both for symbols and strings. I also made the defensive programming a little more explicit by using early returns as appropriate.

Please let me know if you'd rather have the compact representation again.

paracycle:
> This looks neat, thanks! 

Thank you, happy to finally get this working. Also forgot to thank @XrXr for giving me the original idea to check the `caller_location` label to see if we are in an `attr_accessor` call or not.

> I'll have to test it on Stripe's codebase to make sure. I'd prefer to do that after the implementation is adjusted to use the alternate `current_declaration` strategy you mentioned.

I think this is ready to run that check, though I suspect this might turn up some unexpected cases at runtime since `attr_accessor` writers haven't been tested properly until now.

paracycle:
@jez I am curious about what this is going to unearth in the Stripe codebase. Did you have any time to run it at all?

jez:
Sorry I haven't had a chance to test it.

My guess is that it's going to expose a bunch of people either passing in mocked values or `nil` values in tests that weren't being enforced prior.

I should have some time to test this today though, so I will get back to you.

jez:
http://go/builds/bui_MYQXn9olG8SXje

jez:
The first build failed with thousands of test failures 😢 

I will take some time to try to figure out whether they all share a couple common causes, or what might be required to fix them all.

jez:
Heads up Ufuk, I'm going to hijack your PR to leave some documentation for posterity in the next comment. (I just want to get a permalink that I can paste into some comments in Stripe's codebase—the comment itself is not addressed to you.)

jez:
👋 Hey there! If you're reading this, it likely means that you're looking at some `attr_reader`/`attr_writer` code and wondering why it's not using `attr_accessor`.

The PR whose comments you're reading right now fixes a bug in Sorbet where Sorbet previously did not wrap the writer method defined by `attr_accessor` in runtime type checking. When this bug was fixed, it exposed some bugs in the codebase where despite having a sig, the code was not actually conforming to the sig at runtime (sometimes this is because the code was using mocks in tests; sometimes this is because the code wasn't properly handling `nil`; etc. There are many reasons).

To allow the Sorbet version to be upgraded, instead of fixing the root causes, we have simply rewritten the `attr_accessor`'s into separate `attr_reader` and `attr_writer` calls. This preserves the runtime typing behavior that used to exist, making it easier to upgrade to a new Sorbet version (and lock in the bugfix for all new code).

If you are seeing this in your code, feel free to either add a sig to the `attr_writer`, or merge the `attr_reader` and `attr_writer` back into one `attr_accessor` call, and fix any ensuing type errors and test failures. Your teammates will thank you!

If you have questions, feel free to reach out.

jez:
> The first build failed with thousands of test failures 😢
> 
> I will take some time to try to figure out whether they all share a couple common causes, or what might be required to fix them all.

It turns out that all the test failures blamed back to 5 individual `attr_accessor` calls.

jez (approved):

jez:
Thanks! This one has been broken for a long time, great to see it fixed.

Qqwy:
Thank you very much for fixing this issue! :heart:

PR #6431, opened 2022-09-26
`Pathname.readlines` receives an optional chomp argument

rafaelfranca:
See https://ruby-doc.org/stdlib-3.1.2/libdoc/pathname/rdoc/Pathname.html#readlines-method and https://ruby-doc.org/core-3.1.2/IO.html#method-c-readlines


jez:

→ https://go/builds/bui_MVNxSdhVtRXAru
→ https://go/builds/bui_MVNxHmT7KkBVXc

jez (approved):

PR #6435, opened 2022-09-28
Assorted pre-work for constants on fast path

jez:
### Motivation

Make the diff smaller in #6422

### Commit summary

- **pre-work: Add FoundFieldHash::toString** (6e3e5312f)

- **pre-work: Typo in comment (åmember)** (959689bb5)

- **pre-work: klass -> scopeClass in FoundStaticField** (64a52eba9)

- **pre-work: Add LOC_store type alias** (d168ec1c9)

- **pre-work: Add removeLocsForFile to all Symbol kinds** (f9e13c5ca)

- **pre-work: Add some comments** (eacc30f5d)

- **pre-work: Factor out `.value()` to a variable** (5ad0c6f63)

- **pre-work: Reorder deleteFieldViaFullNameHash and deleteMethodViaFullNameHash** (ccf5af081)

  The diff is annoying but it's literally just swapping the order of these
  two methods.

- **pre-work: Factor out `matchesFullNameHash` helper** (ccebdc71c)

- **pre-work: More comment tweaks** (ddea4d366)

- **pre-work: Add "deleteSymbolViaFullNameHash" (kind-agnostic)** (975b88cf3)

  The diff is ugly, but this is basically just factoring out the common
  parts of `delete{Field,Method}ViaFullNameHash` to a helper function,
  called `deleteSymbolViaFullNameHash`. The only part remaining that is
  not shared by those two `delete*` methods is the logic to compute the
  `owner` symbol whose members should be looped.

  This works by the observation that a `NameRef` already uniquely
  identifies a Symbol in the `members` list, so comparing name hashes for
  equality is good enough.

  Note that we unwrap certain kinds of unique names, but that's
  intentional, as it simply means that we'll delete more symbols (e.g.,
  those that were created due to a mangle rename). But names like `@foo`
  and `foo` and `Foo` should has to different values, and if they don't,
  the only problem is for the symbol kinds that we can't re-enter, which
  is just class or module symbols, so we can skip over those and treat the
  ones we don't skip over agnostically with `SymbolRef`.

### Test plan

This should be a 100% no-op change.


jez:
* master
  * **PR #6435** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6435" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6435?utm_source=stack-comment).

core/SymbolRef.h
```diff
@@ -535,7 +536,9 @@ class SymbolRef final {
     core::SymbolRef owner(const GlobalState &gs) const;
     core::Loc loc(const GlobalState &gs) const;
     bool isPrintable(const GlobalState &gs) const;
-    const InlinedVector<Loc, 2> &locs(const GlobalState &gs) const;
+    using LOC_store = InlinedVector<Loc, 2>;
+    const LOC_store &locs(const GlobalState &gs) const;
```
froydnj, line 540: Nonblocking, but this seems like a) a peculiar place to put this typedef (why do `SymbolRef`s care about the underlying storage of the locs?) and b) a weird way to constrain the interface of `locs`.  Returning an `absl::Span<Loc>` here would really be more appropriate, and that would enable dropping the typedef entirely or moving it to someplace more appropriate (`core/Symbols.h`, perhaps).

I do realize that the `locs` interface was pre-existing, so not that big a deal, but it does feel weird.

namer/namer.cc
```diff
@@ -1380,50 +1380,62 @@ class SymbolDefiner {
...
+                    case core::SymbolRef::Kind::Method:
+                        ctx.state.deleteMethodSymbol(oldSymbol.asMethodRef());
+                        break;
+                    case core::SymbolRef::Kind::FieldOrStaticField:
+                        ctx.state.deleteFieldSymbol(oldSymbol.asFieldRef());
+                        break;
+                    case core::SymbolRef::Kind::TypeMember:
+                    case core::SymbolRef::Kind::ClassOrModule:
+                    case core::SymbolRef::Kind::TypeArgument:
+                        ENFORCE(false);
```
froydnj, line 1437: Does this need a `break;` or something after it to not warn in non-debug builds?

namer/namer.cc
```diff
@@ -1380,50 +1380,62 @@ class SymbolDefiner {
...
         for (const auto &[memberName, memberSym] : owner.data(ctx)->members()) {
-            if (!memberSym.isMethod()) {
+            if (memberSym.isClassOrModule()) {
+                // Safeguard against a collision function in our `FullNameHash` function (e.g., what
+                // if `foo()` and `::Foo` have the same `FullNameHash`, but were given two different
+                // `NameRef` IDs and thus don't collide in the `members` list?).
+                //
+                // Any other collisions are fine, because we will already be re-entering the new
+                // definitions later on in incremental namer. It's only classes that we definitely
+                // never want to delete accidentally.
```
froydnj, line 1413: I was going to write something up about how this check doesn't seem sufficient, but I guess the argument is that we were going to change the locs anyway, this is just doing it a bit earlier than we were before.

froydnj (approved):

PR #6439, opened 2022-09-28
move some `LSPPreprocessor` code into the file where it belongs

froydnj:
### Motivation

I'm sure this is not a problem for people using jump-to-def and whatnot, but I think this change increases clarity a little regardless.

I did have to duplicate `mergeCounters` to get everything to compile, which I think is OK.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPPreprocessor.cc
+++ b/main/lsp/LSPPreprocessor.cc
@@ -25,6 +25,24 @@ string readFile(string_view path, const FileSystem &fs) {
     }
 }
 
+class TerminateOnDestruction final {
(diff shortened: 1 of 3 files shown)


jez:
> for people using jump-to-def and whatnot

our clangd setup (maybe bazel related, maybe not) makes it hard for clangd to figure out where a source file is. it will frequently just jump to the header file unless the file that has the implementation is already open in the editor, so this actually makes things easier in the jump-to-def case too.

jez (approved):

PR #6444, opened 2022-09-30
add actual `LSPTask` objects for Watchman `state-enter`/`state-leave` notifications

froydnj:
### Motivation

This is essentially pre-work for a larger change that wants to check for particular states and take action on said states.

This PR should be reviewable commit by commit.

### Test plan

See included automated tests.


main/lsp/watchman/WatchmanProcess.cc
```diff
@@ -21,6 +21,17 @@ WatchmanProcess::~WatchmanProcess() {
...
+namespace {
+template <typename F> void catchDeserializationError(spdlog::logger &logger, const string &line, F &&f) {
+    try {
+        f();
+    } catch (sorbet::realmain::lsp::DeserializationError e) {
+        // Gracefully handle deserialization errors, since they could be our fault.
+        logger.error("Unable to deserialize Watchman request: {}\nOriginal request:\n{}", e.what(), line);
+    }
+}
+} // namespace
```
jez, line 33: I think I might have just wrapped the loop body or the `if/else` statement inside the loop in a single `try/catch` but this is fine too.

jez (approved):
Makes sense. Curious to see what you're going to go inside the LSPTask's, because I still don't have an idea of what code I'd fill them out with.

froydnj:
> Makes sense. Curious to see what you're going to go inside the LSPTask's, because I still don't have an idea of what code I'd fill them out with.

I am not 100% sure whether there will be bits added to `LSPPreprocessor` or similar to call from inside the tasks, or whether the tasks will mostly be useful for `LSPPreprocessor` or similar to recognize `state-enter`/`state-leave` events and take appropriate action.

PR #6447, opened 2022-09-30
Update singleton class locs on fast path

jez:
### Motivation

The fast path would always set the loc for a class symbol to the
`declLoc` of the class node in the AST. But it would not also update the
singleton class's loc. This means that the old singleton class loc was
effectively an offset into a buffer state that didn't exist anymore. At
best this would show a red squiggle in a stale location and at worse it
would cause a crash.

We saw plenty of these crashes in practice, and multiple recent changes
landed in Loc.cc in an attempt to collect more logs to help us identify
where the problem was coming from. But because the crash manifests far
from where the bug happens (in error printing code, instead of in namer
code where the bug is), it was hard to use the logs to figure out what
had happened.

The test added in this test crashes with an ENFORCE failure in debug
builds and in a release/non-ENFORCE'ing build it trips the log lines
that we had added. It's possible that there are still other places that
exhibit other similar problems, so I'm going to keep the logs in place
for the time being, and circle back after observing the crash reports
for a while. Hopefully we can remove them finally.

### Test plan

See included automated tests.


jez:
* master
  * **PR #6447** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6447" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6447?utm_source=stack-comment).

froydnj (approved):
Nice find!

PR #6452, opened 2022-10-04
Use enum instead of a bool for the typechecking path

iz-stripe:
It's a purely refactoring change, it doesn't bring any new functionality.

The change replaces the `bool canTakeFastPath` with a `PathType typecheckingPath`. `PathType` — is an enum with two cases `Fast` and `Slow`. All corresponding methods, comments and variable names were updated to suite `typecheckingPath`.

### Motivation

This is a prework for introducing "the middle path" — a slow path with incremental namer and resolver

### Test plan

The refactoring didn't break any existing tests, so it should be good!


main/lsp/LSPFileUpdates.h
```diff
@@ -23,7 +24,9 @@ class LSPFileUpdates final {
     std::vector<std::shared_ptr<core::File>> updatedFiles;
     std::vector<ast::ParsedFile> updatedFileIndexes;
 
-    bool canTakeFastPath = false;
+    // Indicates the quality of typechecking service for the change
+    PathType typecheckingPath = PathType::Slow;
```
froydnj, line 28:
```suggestion
    PathType typecheckingPath = PathType::Slow;
```

froydnj (approved):
Using enums instead of `bool` is a good change on its own, regardless!

PR #6460, opened 2022-10-06
enumerating over a CSV should return rows

Tiedye:
I think, I can flesh this out

The particular issue a ran into was `map` on a `CSV` object wasnt allowing me to call `fields` on the row as that is a property of `CSV::Row` not `Array`

### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/csv.rbi
+++ b/rbi/stdlib/csv.rbi
@@ -212,7 +212,7 @@ class CSV < Object
   include Enumerable
 
   extend T::Generic
-  Elem = type_member(:out) {{fixed: T::Array[T.nilable(String)]}}
+  Elem = type_member(:out) {{fixed: CSV::Row}}
 
   # The options used when no overrides are given by calling code. They are:
   #


froydnj:

→ https://go/builds/bui_MdEtcvXz43alJi
→ https://go/builds/bui_MdEtHd6Uv3yMXP

Morriar:
I'm not sure this is correct as the returned type depends on whether the CSV was initialized with `headers: true` or not.

For example this will iterate over an array of strings:

```ruby
csv = CSV.new("a;b;c\nc;d;e")
csv.each do |row|
  puts row
end
```

While this will iterate over an array of rows:

```rb
csv = CSV.new("a;b;c\nc;d;e", headers: true)
csv.each do |row|
  puts row
end
```

From the documentation:

> Parsing methods commonly return either of:
> * An Array of Arrays of Strings
> * A CSV::Table object. For details, see CSV with Headers

Maybe the type should be `T.any(T::Array[T.untyped], CSV::Row)])` in which case the user would be responsible to cast the returned type depending on what they expect to find.

Tiedye:
Gonna close this as I havent been using ruby for a year

PR #6461, opened 2022-10-06
Incremental slow path

ilyailya:
The change adds a slow path with an incremental namer and resolver. 
The path is executed only if the change is eligible for the fast path, 
but there are too many files in the edit (number of files is more than `config->opts.lspMaxFilesOnFastPath`)

The new path is very similar to the slow path, the only differences are:
1.  We need to compute files and hashes like we do for the fast path
2. We need to reuse the global state we got from the previous fast path

### Motivation

Faster edits!

### Test plan

See included automated tests.


main/lsp/LSPPathType.h
```diff
@@ -11,6 +11,9 @@ enum class PathType {
     // Full typecheck
     Slow,
 
+    // Slow path (cancelable and parallel), but with incremental namer and resolver
+    SlowWithIncrementalResolver,
```
jez, line 15: Love this name!

main/lsp/LSPTypechecker.cc
```diff
@@ -221,60 +306,9 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
-            oldFoundHashesForFiles.emplace(f, f.data(*gs).getFileHash()->foundHashes);
-        }
-    }
-
-    ENFORCE(gs->lspQuery.isEmpty());
-    auto resolved =
-        config->opts.lspExperimentalFastPathEnabled
-            ? pipeline::incrementalResolve(*gs, move(updatedIndexed), std::move(oldFoundHashesForFiles), config->opts)
-            : pipeline::incrementalResolve(*gs, move(updatedIndexed), nullopt, config->opts);
+    auto [toTypecheck, oldFoundHashesForFiles] = filesToTypecheck(*gs, config, updates, workers, result);
```
jez, line 309: What's the reasoning for having `filesToTypecheck` be a separate function? If I'm reading the code correctly, it can be inlined into `runIncrementalResolver` below. This has a couple benefits:

- It means that we don't need two separate `runIncrementalResolver` methods (not sure if this was intentional or not, but this PR defines `runIncrementalResolver` as an overloaded function, which usually makes the code hard to follow when picking it back up 6 months in the future.)
- It means that the logic for getting the big list of old hashes can be in one method. Right now it's broken up, and I think that it's nice to be able to read one function top-to-bottom and see all the logic that it does to compute hashes. In particular there's some subtlety around whether we move or don't move the old found hashes. It would be nice to get it back to having that logic and comments being in one function.

main/lsp/LSPTypechecker.cc
```diff
@@ -62,6 +64,88 @@ vector<ast::ParsedFile> sortParsedFiles(const core::GlobalState &gs, ErrorReport
...
+        op = make_unique<ShowOperation>(*config, ShowOperation::Kind::FastPath);
+    }
+    ENFORCE(gs.errorQueue->isEmpty());
+    vector<ast::ParsedFile> updatedIndexed;
+    for (auto &f : toTypecheck) {
+        // TODO(jvilk): We don't need to re-index files that didn't change.
+        // (`updates` has already-indexed trees, but they've been indexed with initialGS, not the
+        // `*gs` that we'll be typechecking with. We could do an ast::Substitute here if we had
+        // access to `initialGS`, but that's owned by the indexer thread, not this thread.)
+        auto t = pipeline::indexOne(config->opts, gs, f);
```
jez, line 118: This runs `index` too, not just incremental resolver. I don't think we can factor out as much code as this PR currently does. If anything, I think we only want small helper functions.

The slow path does its own thing for indexing, and thus can't share the code that does indexing with the fast path and incremental slow path.

jez (commented):
I've left a handful of comments. I think I'm going to pause for a while and take another look after this batch.

main/lsp/LSPTypechecker.cc
```diff
@@ -36,6 +36,8 @@ TypecheckingPath toTypecheckingPath(PathType pathType) {
     switch (pathType) {
         case PathType::Slow:
             return TypecheckingPath::Slow;
+        case PathType::SlowWithIncrementalResolver:
+            return TypecheckingPath::SlowWithIncrementalResolver;
```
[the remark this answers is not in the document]
ilyailya: Yep, this part is a bit ugly, can agree. 
The `PathType` is used internally in the typechecking pipeline. Unfortunately, we would need essentially the same type to notify the test runner about the typechecking path we selected. But the type should be JSON serializable to be send to the LSP client. 


So, I had a few options:

1) Create two types: one internal and one external
2) Create a JSON serializable type (in `make_lsp_types.cc`) and use it for the client-server communication and the internal state tracking. It should work, but I didn't want the pipeline to depend on communication layer types. It feels like a leaked abstractions to me.
3) Figure out how to serialize internal type `PathType` to be serializable. It's just to much hassle to figure out how to do it properly. 

>Is there value in having both?

I see a value in having no (or less) dependencies on the generated LSP types in the typechecking pipeline. Please let me know if that's not a concern for Sorbet, we can use ant other option too.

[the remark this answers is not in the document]
jez: > I see a value in having no (or less) dependencies on the generated LSP types in the typechecking pipeline. Please let me know if that's not a concern for Sorbet, we can use ant other option too.

I'm not sure that's a concern. You'll notice that you didn't have to add any additional `#include` files to use these generated types—they were already in scope. Pretty much all of `main/lsp/` depends on the autogenerated LSP types.

(I would agree with you if it had been a question of whether to put these typechecking path types in the core pipeline.)

main/lsp/LSPTypechecker.cc
ilyailya: You're right here. I believe `filesToTypecheck` is a left over from one of my previous attempts to make incremental slow path work. Thanks for noticing this, I'll inline it

main/lsp/LSPTypechecker.cc
ilyailya: Right, but we need this indexing for resolver to be incremental

> The slow path does its own thing for indexing, and thus can't share the code that does indexing with the fast path and incremental slow path.

Yep, but this indexing is guarded by if, so it will be executed only on incremental slow path. I can change name to add clarity to it
https://github.com/sorbet/sorbet/blob/b5052fb6357e1ff1e66338403f6086509cdaeed6/main/lsp/LSPTypechecker.cc#L473-L475

test/testdata/lsp/middle_path/middle_path__01.1.rbupdate
```diff
@@ -0,0 +1,12 @@
+# typed: true
```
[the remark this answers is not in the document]
ilyailya: After all `test/testdata/lsp/fast_path/initialize__01.rb` also broke, and I converted it to `assert-incremental-slow-path: true` (:

PR #6462, opened 2022-10-07
Expand T::Struct docs with various examples of rules

bdewater:
After discussion on the Sorbet Slack I found out T::Struct can do a lot more than this page initially let me to believe.

The changes cover most (if not all) prop rules other than:
- deprecated ones (eg `:ifunset`)
- things that look like Stripe internals (`:foreign`, `:redaction` and `:sensitivity`)

Also fix the 'serialize' example return value to show string keys.

### DIFF
--- a/website/docs/tstruct.md
+++ b/website/docs/tstruct.md
@@ -4,8 +4,6 @@ title: Typed Structs via T::Struct
 sidebar_label: T::Struct
 ---
 
-> TODO: This page is still a fragment. Contributions welcome!
-
 While Sorbet supports using Hashes and Structs, it isn't always possible to
 model them statically.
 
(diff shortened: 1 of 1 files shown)


bdewater:
Seeing the order on the enum docs is inverted compared to what I have here:
- [Converting enums to other types](https://sorbet.org/docs/tenum#converting-enums-to-other-types)
- [Converting from other types to enums](https://sorbet.org/docs/tenum#converting-from-other-types-to-enums)

I'll tweak this PR.

jez:
Thanks for tackling this! We have been limping along with only Stripe-internal docs for a lot of these options for a long time.

froydnj is out this week, so I'm going to take this off his plate. I'm a little busy for the rest of the day, but I should be able to take this on either tomorrow or wednesday. I'll also go over and try to merge some of the Stripe-internal docs that we have into this doc in places where it makes sense.

jez:
I started out leaving feedback on this PR, but as I was going, I realized that I think it would be better for me to just open a separate PR to update these docs.

https://github.com/sorbet/sorbet/pull/6502

Feel free to leave your feedback on that PR, but I believe it should at least incorporate as much as this PR adds to the doc, with some other expanded content.

PR #6465, opened 2022-10-11
Ignore formatting requests on __package.rb files

reese:
### Motivation

Tooling for `__package.rb` files assumes some pretty specific rules that aren't quite the same as what `rubyfmt` enforces (e.g. sometimes `rubyfmt` will add parentheses to `import` and `export` calls, but our tools don't handle this). The easiest way around this is just to ignore requests for formatting those files. Other internal scripts also ignore package files explicitly, and I don't think `rubyfmt` should try and guess whether or not something is a package based on the file contents.

### Test plan

I manually tested this internally in VSCode.

### DIFF
--- a/main/lsp/requests/document_formatting.cc
+++ b/main/lsp/requests/document_formatting.cc
@@ -1,8 +1,8 @@
 #include "main/lsp/requests/document_formatting.h"
-#include "absl/strings/match.h"
 #include "common/FileOps.h"
 #include "common/Subprocess.h"
 #include "common/common.h"
+#include "core/Files.h"
 #include "main/lsp/LSPLoop.h"
 #include "main/lsp/LSPOutput.h"
 #include "main/lsp/json_types.h"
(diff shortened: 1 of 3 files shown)


main/lsp/requests/document_formatting.cc
```diff
@@ -80,7 +81,9 @@ void DocumentFormattingTask::preprocess(LSPPreprocessor &preprocessor) {
         sourceView = sorbet::FileOps::read(path);
     }
 
-    if (!sourceView.empty()) {
+    // Don't format `__package.rb` files, since currently formatting them
+    // can potentially break some pay-server tooling
+    if (!sourceView.empty() && !absl::EndsWith(path, "__package.rb")) {
```
jez, line 86: Can we use the logic built into Sorbet for this?

```cpp
config.uri2FileRef(typechecker.state(), params->textDocument->uri).data(gs).isPackage()
```

jez (approved):
looks good but one suggestion

main/lsp/requests/document_formatting.cc
jez: Alternatively, move the `isPackagePath` helper out of `Files.cc` into a static helper method on `core::File` (so that you can work with string_view instead of `core::FileRef`)

PR #6331, opened 2022-09-01
Remove all package-awareness from autogen

aadi-stripe:
Removes all packaging code from autogen.

### Motivation

As part of an ongoing migration to `zeitwerk` in the Stripe codebase, we do not need package-awareness anymore to facilitate autoloading. Thus, I'm removing this logic to save overhead.

### Test plan

See included automated tests. Tested on Stripe codebase.

PR #6466, opened 2022-10-12
Bring back package-awareness to Sorbet-autogen

aadi-stripe:
In https://github.com/sorbet/sorbet/pull/6331 we removed package-awareness from Sorbet in autogen mode. Turns out that was a bit premature, as in the Stripe codebase we are working on setting up Zeitwerk at a package-local level. This requires the Sorbet-generated shim to be able to defer to Zeitwerk code on select packages -- which it can only do if it is package-aware. This change sets the stage for that to happen.

### Motivation

Allow path-based autoloading to scale at Stripe.

### Test plan


aadi-stripe:

→ https://go/builds/bui_Mb5Z9zx4kIjtTo
→ https://go/builds/bui_Mb5Zn8Chrak6vW

aadi-stripe:

→ https://go/builds/bui_Mb7DMGui0SAbdn
→ https://go/builds/bui_Mb7DuriHlXwb2M

jez (approved):

PR #6469, opened 2022-10-14
Cache files to typecheck on a fast path

ilyailya:
The change caches the result of the `LSPFileUpdates::fastPathFilesToTypecheck` call, which we compute in process of making the typechecking path decision. 
That data is used later in the incremental resolver on the fast path (and in future in the incremental slow path)

### Motivation

The `fastPathFilesToTypecheck` can be an expensive method to call. So having less calls will decrease latency of the fast path requests

### Test plan

See included automated tests.


jez:
Not sure what the state of this PR is. We might want to punt this to the next time we revisit incremental work?

PR #6471, opened 2022-10-14
Don't mangle constant symbols

jez:
### Motivation

While working on #6422, it became apparent that we can never mangle a class
symbol. The reasoning is like imagine that you start with this:

```ruby
class A; end
A = 1
```

and then edit to this:

```ruby
class A; end
```

After the initial slow path for the first buffer, we're going to have a
`GlobalState` like this:

```
class ::A$1
static-field ::A
```

Then after the fast path constants change in #6422, we would delete all static
field symbols for the second edit, leaving everything else untouched:

```
class ::A$1
```

Despite there being no redefinition of `class A` anymore, its name is still
mangled, which it shouldn't be.

This change makes it so that **constant redefinition errors never mangle the
previous symbol**, and instead choose the next available mangled name for *their
own* symbol.

This comes with a slightly annoying behavior change:

```ruby
A = type_member
A = 1
```

In the old world, the thing that came last would take precedence, which more
closely matches the runtime. But giving this up is not *that* bad, because:

- It's not always the case that a redefinition error corresponds to something
  coming "last," because the conflict might happen across different files (with
  unknown execution order)
- These errors are `# typed: false` errors, so you can't actually check in code
  with one of these errors. (This is in contrast to method redefinition errors,
  which are actually not reported at `# typed: false`, which means that people
  actually do rely on the "last method wins" behavior.)

### Test plan

See included automated tests.


jez:
* master
  * **PR #6472** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6472" target="_blank">Graphite</a> 
    * **PR #6471** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6471" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6471?utm_source=stack-comment).

namer/namer.cc
```diff
@@ -1256,21 +1260,22 @@ class SymbolDefiner {
 
         auto scope = ensureScopeIsClass(ctx, squashNames(ctx, staticField.scopeClass, contextClass(ctx, ctx.owner)),
                                         staticField.name, staticField.asgnLoc);
+        auto name = staticField.name;
         auto sym = ctx.state.lookupStaticFieldSymbol(scope, staticField.name);
         auto currSym = ctx.state.lookupSymbol(scope, staticField.name);
```
froydnj, line 1265: Maybe use `name` here to make things clearer?

namer/namer.cc
```diff
@@ -1334,8 +1339,9 @@ class SymbolDefiner {
             // otherwise, we're looking at a type member defined in this class in the same file, which means all we
             // need to do is find out whether there was a redefinition the first time, and in that case display the
             // same error
-            auto oldSym = ctx.state.findRenamedSymbol(onSymbol, existingTypeMember);
-            if (oldSym.exists()) {
+            auto name = existingTypeMember.data(ctx)->name;
+            if (isMangleRenameUniqueName(ctx, name)) {
+                auto oldSym = ctx.state.lookupSymbol(onSymbol, typeMember.name);
```
froydnj, line 1344:
```suggestion
                auto oldSym = ctx.state.lookupSymbol(onSymbol, name);
```

froydnj (approved):

namer/namer.cc
jez: I don't think this is equivalent? I think in this case, we're guaranteed that `name` is a MangleRename unique name, which means that if we look up the symbol with that name, we're just going to find the mangled symbol again. What we want with this lookupSymbol call is to find the name that's **not** mangled, which we can do by using `typeMember.name`, because `typeMember.name` came from the FoundDefinition, and the found definitions never have mangled names (because users can't write mangled names).

namer/namer.cc
jez: PR to come

PR #6472, opened 2022-10-14
🔥 Delete LSP stale state mode

jez:
### Motivation

1.  This mode is increasingly becoming irrelevant by our changes to take the
    actual fast path more frequently.

2.  This blocks landing #6471, where one test fails only in the stale state
    mode, in a way that is intrinsic to the stale state mode's existence, not to
    the change itself (we skipped over processing all todo type aliases in the
    immutable resolver, causing ENFORCEs to fail).

3.  Keeping this code around but unused is going to continue to be a drag on
    developer velocity. For example, this should make a (minor?) performance
    improvement to all testdata tests, because they don't have to run the
    incremental namer and resolver (twice!) on every test.

### Test plan

existing tests


jez:
* master
  * **PR #6472** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6472" target="_blank">Graphite</a>  👈
    * **PR #6471** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6471" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6472?utm_source=stack-comment).

main/lsp/tools/make_lsp_types.cc
```diff
@@ -1304,12 +1304,6 @@ void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_pt
                                             makeField("message", JSONString),
                                         },
                                         classTypes);
-    auto SorbetFenceParams = makeObject("SorbetFenceParams",
-                                        {
-                                            makeField("id", JSONInt),
-                                            makeField("advanceIfStaleDataAvailable", JSONBool),
-                                        },
-                                        classTypes);
```
froydnj, line 1312: I kind of want to keep the stronger type of `SorbetFenceParams` here and just delete `advanceIfStaleDataAvailable`, but up to you.

froydnj (approved):
Sad to see this go, but this will definitely help with implementing other features.

main/lsp/tools/make_lsp_types.cc
jez: Some of the changes I made by simply reverting old PRs that still applied cleanly, this was one of them. But I can put this one back.

jez: This is blocking the stack though, so I'll do that in a future change.

PR #6474, opened 2022-10-17
make `FileNotFoundException` have an actual message

froydnj:
### Motivation

We were having some issues at Stripe where we were getting a `FileNotFoundException` at:

https://github.com/sorbet/sorbet/blob/3776975aa2b22bdc537c150e24fd475e16c593f7/main/options/options.cc#L689-L696

...but it's not clear that catching a `FileNotFoundException` here necessarily implies that the original, top-level directory is the thing not found.

The way this PR chooses to fix the issue is to add a proper message to `FileNotFoundException`, which we can then propagate out to this particular location.

I considered making the "message" to `FileNotFoundException` just the filename, but it seemed better to have an actual message, since we can include a little bit more information about what exactly we were trying to do when we couldn't find the file.  Happy to discuss alternative approaches.

### Test plan

See included automated tests.

### DIFF
--- a/common/Exception.h
+++ b/common/Exception.h
@@ -16,7 +16,7 @@ public:
     SorbetException(const char *message) : logic_error(message) {}
 };
 
-class FileNotFoundException : SorbetException {
+class FileNotFoundException : public SorbetException {
 public:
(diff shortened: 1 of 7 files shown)


jez (approved):
🙏!

PR #6478, opened 2022-10-17
Generate package-based autoload shims for path-based autoloading

aadi-stripe:
Parses a new package-spec DSL annotation, `autoloader_compatibility: 'strict'`, and generates autoloader shims for path-based autoloading for registered packages.

### Motivation

In the Stripe codebase, we are migrating to Zeitwerk for path-based autoloading. In order to assist an incremental rollout, we are creating shims at the package-level between the current autoload system and the new system.

Introduces a new, temporary CLI arg called `--autogen-autoloader-pbal-namespaces`.

### Test plan


packager/packager.cc
```diff
@@ -963,13 +970,39 @@ struct PackageInfoFinder {
...
             // Transform: `restrict_to_service Foo` -> `restrict_to_service <PackageSpecRegistry>::Foo`
             auto importArg = move(send.getPosArg(0));
             send.removePosArg(0);
             ENFORCE(send.numPosArgs() == 0);
             send.addPosArg(prependName(move(importArg), core::Names::Constants::PackageSpecRegistry()));
         }
+
+        if (send.fun == core::Names::autoloader_compatibility() && send.numPosArgs() == 1) {
+            // Parse autoloader_compatibility DSL and set strict bit on PackageInfoImpl if configured
+            auto *compatibilityAnnotationLit = ast::cast_tree<ast::Literal>(send.getPosArg(0));
```
aadi-stripe, line 984: Parse the `autoloader_compatibility` DSL

main/autogen/autoloader.cc
```diff
@@ -323,23 +341,52 @@ void DefTreeBuilder::addParsedFileDefinitions(const core::GlobalState &gs, const
...
+            // computation / code generation, given this is the only current use-case for registering
+            // packages in this context in the Stripe codebase.
+
+            // Additionally this package must be registed for path-based autoloading.
+            // TODO: (aadi-stripe, 10/24/2022) Remove this functionality once we no longer require
+            // special registration.
+            auto &pkgFullName = pkg.fullName();
+            if (!alCfg.registeredForPBAL(pkgFullName)) {
+                continue;
+            }
```
aadi-stripe, line 383: Only mark `strict` compatible packages that are additionally registered for PBAL via:
```
--autogen-autoloader-pbal-namespaces Opus::Admin
```

main/options/options.cc
```diff
@@ -406,6 +406,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                     cxxopts::value<vector<string>>());
     options.add_options("advanced")("autogen-autoloader-modules", "Top-level modules to include in autoloader output",
                                     cxxopts::value<vector<string>>());
+    options.add_options("advanced")("autogen-autoloader-", "Top-level modules to include in autoloader output",
```
bhang-stripe, line 409: leftovers?

main/autogen/autoloader.cc
```diff
@@ -210,7 +224,19 @@ string DefTree::renderAutoloadSrc(const core::GlobalState &gs, const AutoloaderC
...
+            const string_view shortName = pkgName.shortName(gs);
+            const string_view mungedName = shortName.substr(0, shortName.size() - core::PACKAGE_SUFFIX.size());
+
+            auto &pkg = gs.packageDB().getPackageInfo(pkgName);
+
+            // First path prefix is guaranteed to be the directory location of the package
+            const string_view pathPrefix = pkg.pathPrefixes()[0];
+
+            fmt::format_to(std::back_inserter(buf), "\n{}.pbal_register_package({}, '{}', '{}')\n",
+                           alCfg.registryModule, fullName, mungedName, std::move(pathPrefix));
```
bhang-stripe, line 238: Are we doing anything with the munged name?

bhang-stripe (changes requested):
The logic lgtm! Just a couple nits

main/autogen/autoloader.cc
aadi-stripe: Nope, this is a holdover from the prior `register_package` API. We can get rid of it.

core/errors/packager.h
```diff
@@ -10,7 +10,7 @@ constexpr ErrorClass InvalidPackageDefinition{3702, StrictLevel::False};
 constexpr ErrorClass RedefinitionOfPackage{3703, StrictLevel::False};
 constexpr ErrorClass PackageNotFound{3704, StrictLevel::False};
 constexpr ErrorClass UnpackagedFile{3705, StrictLevel::False};
-constexpr ErrorClass InvalidImportOrExport{3706, StrictLevel::False};
+constexpr ErrorClass InvalidConfiguration{3706, StrictLevel::False};
```
jez, line 13: Is it worth either having separate error messages, or updating the existing error docs to talk about this new failure mode?

aadi-stripe:
@bhang-stripe Fixed

bhang-stripe (approved):

main/autogen/autoloader.h
```diff
@@ -123,7 +127,7 @@ class DefTreeBuilder {
                              NamedDefinition);
 
     static DefTree merge(const core::GlobalState &gs, DefTree lhs, DefTree rhs);
-    static void markPackages(const core::GlobalState &gs, DefTree &root);
+    static void markPackages(const core::GlobalState &gs, DefTree &root, const AutoloaderConfig &);
```
jez, line 130: Can we add a variable name here for documentation?

```suggestion
    static void markPackages(const core::GlobalState &gs, DefTree &root, const AutoloaderConfig &autoloaderCfg);
```

packager/packager.cc
```diff
@@ -963,13 +970,39 @@ struct PackageInfoFinder {
...
+            if (compatibilityAnnotationLit == nullptr || !compatibilityAnnotationLit->isString()) {
+                if (auto e = ctx.beginError(send.loc, core::errors::Packager::InvalidConfiguration)) {
+                    e.setHeader("Argument to `{}` must be a string literal", send.fun.show(ctx));
```
jez, line 987: What do you think about adding an autocorrect for the case where the user wrote a symbol but it was a valid argument? `:legacy` -> `'legacy'` etc.

jez (approved):

packager/packager.cc
aadi-stripe: I think this could be useful, but would probably want to do this down the line. It's not clear how much our users will be adding annotations here vs our package-generation tooling.

core/errors/packager.h
aadi-stripe: I don't think this new failure mode is worth documenting publicly. It's very niche within our packaging system (which is already only used at Stripe), and it's not a first-class feature. I wanted to reuse the same `ErrorClass`, hence why I renamed it internally -- if you feel strongly that this has to be a brand new error code then I'm happy to make it so!

main/autogen/autoloader.h
aadi-stripe: fixed

aadi-stripe:
@jez Addressed feedback.

core/errors/packager.h
jez: I don't really mind whether it's one or two error codes, but I think that

- the time when we introduce the new behavior is the best time to document it
- using "it's only in use at Stripe" as a reason for why not to document something just deprives _people at Stripe_ from having access to documentation

aadi-stripe: Fair. I'll attempt to make a doc update here and see if it fits.

main/autogen/autoloader.cc
```diff
@@ -210,7 +224,17 @@ string DefTree::renderAutoloadSrc(const core::GlobalState &gs, const AutoloaderC
...
+        if (pkgName.exists()) {
+            ENFORCE(!gs.packageDB().empty());
+
+            auto &pkg = gs.packageDB().getPackageInfo(pkgName);
+
+            // First path prefix is guaranteed to be the directory location of the package
+            const string_view pathPrefix = pkg.pathPrefixes()[0];
+
+            fmt::format_to(std::back_inserter(buf), "\n{}.pbal_register_package({}, '{}')\n", alCfg.registryModule,
+                           fullName, std::move(pathPrefix));
```
ngroman, line 236: you shouldn't need to `move` a `string_view` since it's a ptr-like type

aadi-stripe: Ah whoops, I thought I had taken care of this. Will fix

aadi-stripe: fixed

core/errors/packager.h
aadi-stripe: done

jez (approved):

ngroman:
Some comments inline, lgtm from my perspective.

PR #6479, opened 2022-10-18
Also silence `private` method redefinition warning

sambostock:
We use `alias_method` to silence warnings when redefining methods.

`method_defined?` only considers public and protected methods.

**Also checking `private_method_defined?` ensures we silence those warnings too.**

### Motivation

#4695 silenced `public` and `protected` method redefinition warnings (probably intending to silence all such warnings). This simply extends this behavior to **`private`** methods for consistency. The same justifications that applied there also apply here.

I noticed this when running the test suite for [Shopify/ruby-lsp](https://github.com/Shopify/ruby-lsp), whose output included several lines of warnings due to this bug.

### Test plan

#4695, which originally introduced silencing these warnings did not include tests, so I think it's reasonable to skip that here too.


froydnj (approved):
Thanks!

PR #6482, opened 2022-10-18
rearrange `appendFilesInDir` to do faster early rejects

froydnj:
### Motivation

We waste time constructing a full string in `appendFilesInDir` in two cases:

1. When processing `.` or `..` in directory entries; or
2. When a file has an extension, but it doesn't match something in our extension list.

Checking for both of these conditions can be done cheaply prior to constructing the full pathname to the relevant file.

This change saves ~10% of the time in processing options (which includes finding all the files to typecheck) on Stripe's codebase.  (This is probably due to a combination of saving on memory allocations and not having to check these files against the ignore list(s).)

### Test plan

See included automated tests.

### DIFF
--- a/common/common.cc
+++ b/common/common.cc
@@ -266,23 +266,37 @@ void appendFilesInDir(string_view basePath, const string &path, const sorbet::Un
     }
 
     while ((entry = readdir(dir)) != nullptr) {
-        auto fullPath = fmt::format("{}/{}", path, entry->d_name);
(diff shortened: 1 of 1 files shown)


jez (approved):
Amazing

froydnj:

→ https://go/builds/bui_MddGbr6or85ff1
→ https://go/builds/bui_MddG181DTXDudj

Issue #6483, opened 2022-10-18
Sorbet should not allow non-Proc types for `&blk` arguments


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20CallTakesBlock%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Bparams%28x%3A%20Integer%29.void%7D%0A%20%20def%20call%28x%29%0A%20%20%20%20p%20x%0A%20%20end%0A%0A%20%20sig%20%7Breturns%28CallTakesBlock%29%7D%0A%20%20def%20to_proc%0A%20%20%20%20self%0A%20%20end%0Aend%0A%0Asig%20%7Bparams%28blk%3A%20CallTakesBlock%29.void%7D%0Adef%20example%28%26blk%29%0A%20%20yield%2042%0A%20%20yield%20''%0Aend%0A%0A%23%20runtime%20error%3A%20%60CallTakesBlock%23to_proc%60%20returns%20neither%20nil%20nor%20a%20%60Proc%60%20object%0A%23%20https%3A%2F%2Fgithub.com%2Fruby%2Fruby%2Fblob%2Fv2_7_6%2Fvm_args.c%23L1150-L1154%0Aexample%28%26CallTakesBlock.new%29)

```ruby
# typed: true
extend T::Sig

class CallTakesBlock
  extend T::Sig
  sig {params(x: Integer).void}
  def call(x)
    p x
  end

  sig {returns(CallTakesBlock)}
  def to_proc
    self
  end
end

sig {params(blk: CallTakesBlock).void}
def example(&blk)
  yield 42
  yield ''
end

# runtime error: `CallTakesBlock#to_proc` returns neither nil nor a `Proc` object
# https://github.com/ruby/ruby/blob/v2_7_6/vm_args.c#L1150-L1154
example(&CallTakesBlock.new)
```

#### Observed output

```
editor.rb:20: Expected Integer but found String("") for argument x https://srb.help/7002
    20 |  yield ''
                ^^
  Expected Integer for argument x of method CallTakesBlock#call:
    editor.rb:6:
     6 |  sig {params(x: Integer).void}
                      ^
  Got String("") originating from:
    editor.rb:20:
    20 |  yield ''
                ^^
Errors: 1
```

#### Expected behavior

[...]

PR #6484, opened 2022-10-18
Fix crash in getCallArguments

### LINKED ISSUE #6475: Crash when using non-proc type for `&blk` argument
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20CallTakesBlock%0A%20%20def%20call%28x%29%0A%20%20%20%20p%20x%0A%20%20end%0Aend%0A%0Asig%20%7Bparams%28blk%3A%20CallTakesBlock%29.void%7D%0Adef%20example%28%26blk%29%0A%20%20yield%2042%0Aend%0A%0Aexample%20do%20%7Cx%7C%0Aend)

```ruby
# typed: true
extend T::Sig

class CallTakesBlock
  def call(x)
    p x
  end
end

sig {params(blk: CallTakesBlock).void}
def example(&blk)
  yield 42
end

example do |x|
end
```

#### Observed output

```
editor.rb:15: Expression does not have a fully-defined type (Did you reference another class's type members?) https://srb.help/7011
    15 |example do |x|
    16 |end
  Note:
    Historically, users seeing this error has represented finding a bug in Sorbet
    (or at least finding a test case for which there could be a better error message).
    Consider searching for similar bugs at https://github.com/sorbet/sorbet/issues
    or reporting a new one.
Errors: 1
```

in debug mode, it actually produces a crash.

#### Expected behavior

Sorbet should not crash.

- - -

jez:
### Motivation

Fixes #6475

This is causing a handful of background crashes on Stripe's codebase
every day.

Separately, this pattern seems like it should be banned outright--I've
opened a ticket to handle that separately:

<https://github.com/sorbet/sorbet/issues/6483>

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1532,6 +1532,10 @@ TypePtr getMethodArguments(const GlobalState &gs, ClassOrModuleRef klass, NameRe
         if (arg.flags.isBlock) {
             continue;
         }
+        if (arg.type == nullptr) {
+            args.emplace_back(core::Types::untyped(gs, method));
+            continue;
+        }
         args.emplace_back(Types::resultTypeAsSeenFrom(gs, arg.type, data->owner, klass, targs));
     }
     return make_type<TupleType>(move(args));
--- a/test/testdata/infer/get_call_arguments_nullptr.rb
+++ b/test/testdata/infer/get_call_arguments_nullptr.rb
@@ -1,0 +1,17 @@
+# typed: true
+extend T::Sig
+
+class CallTakesBlock
+  def call(x)
+    p x
+  end
+end
+
+sig {params(blk: CallTakesBlock).void}
+def example(&blk)
+  yield 42
+end
+
+example do |x|
+  T.reveal_type(x) # error: `T.untyped`
+end


jez:
* master
  * **PR #6484** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6484" target="_blank">Graphite</a>  👈
    * **PR #6486** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6486" target="_blank">Graphite</a> 
      * **PR #6487** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6487" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6484?utm_source=stack-comment).

core/types/calls.cc
```diff
@@ -1532,6 +1532,10 @@ TypePtr getMethodArguments(const GlobalState &gs, ClassOrModuleRef klass, NameRe
         if (arg.flags.isBlock) {
             continue;
         }
+        if (arg.type == nullptr) {
+            args.emplace_back(core::Types::untyped(gs, method));
+            continue;
+        }
```
froydnj, line 1538: Naively, I would assume that this PR fixes the type of a block argument that wasn't a proc-typed thing...but it looks like we handle block arguments in the conditional just above this.

Is this instead fixing the behavior in the testcase where we want to know what the type of `x` is for `CallTakesBlock#call`?  Or is is fixing the case where we're trying to see what the type of `x` is in the block for the call to `example`?

froydnj (approved):
I am willing to believe this is correct, but for my own edification, I'd like a little clarification on where the crash here is coming from.

core/types/calls.cc
jez: `getMethodArguments` is passed a `NameRef name`, which is maybe a bit more generic than it needs to be, because it is always passed the `core::Names::call()` if you trace the call stack all the way up:

```cpp
auto params = procType.getCallArguments(ctx, core::Names::call());
```

What happens is that we look up the method on that type with the name `call` (possibly unwrapping things like `T.any` and `T.all` to get there). Once we've found the `call` method, we look at it's args and construct a tuple type with the type for all the args, so that we can have a `LoadYieldParams` instruction in the CFG that is typed as that tuple:

```ruby
example do |x|
  T.reveal_type(x) # error: `T.untyped`
end
```

becomes

```
<args> : [T.untyped] = LoadYieldParams(example)
x : T.untyped = <args>.[](0)
```

(if there had been more arguments in the block, like `do |x, y, z|`, we would have had things like `<args>.[](1)`, etc.)

So in this case, `example` says its block (`procType`) is `CallTakesBlock`, then we look up the `call` method on that type, and want to construct a 1-tuple representing the types of the arguments that `CallTakesBlock#call` takes, which should be `[T.untyped]`.

But the problem was that if there is no sig on `call`, then `arg.type` is `nullptr`, which meant we constructed `[nullptr]`, which causes things to crash.

> Is this instead fixing the behavior in the testcase where we want to know what the type of `x` is for `CallTakesBlock#call`? Or is is fixing the case where we're trying to see what the type of `x` is in the block for the call to `example`?

It's kind of both? We need to get the type of `x` off the symbol for `CallTakesBlock#call` to see what the type of `x` is in the block for the call to `example`.

I guess maybe the best answer to your question is the latter, because this `getMethodArguments` function for sure runs while infer is running to typecheck the call to `example do...end`, not to typecheck the body of `CallTakesBlock#call`.

PR #6486, opened 2022-10-18
Add error for non-`Proc`-typed block argument

jez:
### Motivation

Fixes #6483

The Ruby VM does not allow this.

### Test plan

See included automated tests.

PR #6487, opened 2022-10-18
Skip checking block type in sorbet-runtime when possible

jez:
### Motivation

Runtime performance

### Benchmarks

I've collected these before fixing the aforementioned problems with `nil`, so
it's possible that these benchmarks will get slower after that.

Only the **bolded** things should get sped up from this change—other changes are included to showcase how noisy the measurements usually are. Speedups are reported as percentage speedups vs "Before" (so negative percents are slowdowns—some things slow down by small percentages due to noise in measurement).

| Benchmark (all times in ns)                                        | Before | w/o `nil` checks | % faster | w/ `nil` checks | % faster |
| ---                                                                | ---:   | ---:             | ---:     | ---:            | ---:     |
| Vanilla Ruby method call                                           | 15     | 15               | 0%       | 15              | 0%       |
| Vanilla Ruby is_a?                                                 | 25     | 24               | 4%       | 26              | -4%      |
| T::Types::Simple#valid?                                            | 37     | 37               | 0%       | 38              | -3%      |
| T.nilable(Integer).valid?                                          | 49     | 49               | 0%       | 51              | -4%      |
| T.any(Integer, Float, T::Boolean).valid?                           | 182    | 180              | 1%       | 180             | 1%       |
| T.let(..., Integer)                                                | 289    | 288              | 0%       | 294             | -2%      |
| sig {params(...).void}                                             | 198    | 210              | -6%      | 198             | 0%       |
| **sig {params(..., blk: T.proc.void)} -- `{}`**                    | 997    | 808              | 23%      | 807             | 24%      |
| **sig {params(..., blk: T.proc.void)} -- `&blk`**                  | 849    | 663              | 28%      | 657             | 29%      |
| **sig {params(..., blk: T.nilable(T.proc.void))} -- `{}`**         | 1044   | 780              | 34%      | 754             | 38%      |
| **sig {params(..., blk: T.nilable(T.proc.void))} -- `&blk`**       | 937    | 660              | 42%      | 651             | 44%      |
| T.let(..., T.nilable(Integer))                                     | 733    | 737              | -1%      | 741             | -1%      |
| sig {params(x: T.nilable(Integer)).void}                           | 223    | 224              | 0%       | 221             | 1%       |
| T.let(..., Example)                                                | 294    | 292              | 1%       | 297             | -1%      |
| sig {params(x: Example).void}                                      | 191    | 192              | -1%      | 189             | 1%       |
| T.let(..., T.nilable(Example))                                     | 736    | 723              | 2%       | 739             | 0%       |
| sig {params(x: T.nilable(Example)).void}                           | 223    | 225              | -1%      | 223             | 0%       |
| sig {params(s: Symbol, x: Integer, y: Integer).void} (with kwargs) | 1413   | 1412             | 0%       | 1389            | 2%       |
| direct call Object#class                                           | 27     | 29               | -7%      | 28              | -4%      |
| .bind(example).call Object#class                                   | 182    | 188              | -3%      | 182             | 0%       |
| .bind_call(example) Object#class                                   | 102    | 106              | -4%      | 101             | 1%       |

**Update**

I pushed a change to make the optimization also combine with the existing fast path optimizations (because those validators already skip the block type checking, so when the method mentions a `&blk` parameter but there is no need to check the block type, we can still take the fast path validators). As expected, the fast path validators are quite fast.

| Benchmark (all times in ns)                                        | w/ `nil` and take fast path if possible | % faster |
| --- | ---: | ---: |
| sig {params(...).void}                                 | 206 | -4%  |
| sig {params(..., blk: T.proc.void)} -- {}              | 819 | 22%  |
| sig {params(..., blk: T.proc.void)} -- &blk            | 689 | 23%  |
| sig {params(..., blk: T.nilable(T.proc.void))} -- {}   | 309 | 238% |
| sig {params(..., blk: T.nilable(T.proc.void))} -- &blk | 236 | 298% |

This makes passing a `{}` block within 50% of no block, and passing a `&blk` block within 15% of no block.

### Analysis

Oddly, the "without `nil` checks" times are actually **faster** than the initial implementation that did not check and report a sorbet-runtiem error for `nil`. I take that to mean that there is likely, there is no additional performance gained by skipping the `nil` checking when possible, but I figure we may as well keep the separate code paths, because it also means that there's not a penalty to doing so.

### Test plan

See included automated tests.

- - - 

**old, outdated initial description**

Leaving this as work-in-progress, because the one thing that the VM doesn't do
that Sorbet does do currently is check whether a block was required.

One approach would be to only skip this check for methods whose block is typed
as `T.nilable(...)`, and simply leave the old behavior otherwise.

Another approach would be to do something like the `create_validator_fast`
methods that skips checking block types for **all** methods (not just those that
have a certain arity) depending on whether they take a block or not.

### DIFF
--- a/gems/sorbet-runtime/.rubocop.yml
+++ b/gems/sorbet-runtime/.rubocop.yml
@@ -39,7 +39,22 @@ Lint/UnusedMethodArgument:
 Lint/UselessMethodDefinition:
   Enabled: false
 
+# Ratcheting these down is kind of outside the purview of a linter
+Metrics/AbcSize:
+  Enabled: false
+Metrics/BlockLength:
+  Enabled: false
+Metrics/BlockNesting:
+  Enabled: false
 Metrics/ClassLength:
+  Enabled: false
+Metrics/CyclomaticComplexity:
+  Enabled: false
+Metrics/MethodLength:
+  Enabled: false
+Metrics/ModuleLength:
+  Enabled: false
+Metrics/PerceivedComplexity:
   Enabled: false
 
 # This can be bad for performance
--- a/gems/sorbet-runtime/bench/typecheck.rb
+++ b/gems/sorbet-runtime/bench/typecheck.rb
@@ -99,6 +99,27 @@ module SorbetBenchmarks
         integer_param(1)
       end
 
+      my_proc = proc {}
+      time_block("sig {params(x: Integer, blk: T.proc.void)} -- block literal") do
+        integer_param_and_block(0) {}
+        integer_param_and_block(0) {}
+      end
+
+      time_block("sig {params(x: Integer, blk: T.proc.void)} -- block pass") do
+        integer_param_and_block(0, &my_proc)
+        integer_param_and_block(0, &my_proc)
+      end
+
+      time_block("sig {params(x: Integer, blk: T.nilable(T.proc.void))} -- block literal") do
+        integer_param_and_nilable_block(0) {}
+        integer_param_and_nilable_block(0) {}
+      end
+
+      time_block("sig {params(x: Integer, blk: T.nilable(T.proc.void))} -- block pass") do
+        integer_param_and_nilable_block(0, &my_proc)
+        integer_param_and_nilable_block(0, &my_proc)
+      end
+
       time_block("T.let(..., T.nilable(Integer))") do
         T.let(nil, T.nilable(Integer))
         T.let(1, T.nilable(Integer))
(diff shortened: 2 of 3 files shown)


jez:
* master
  * **PR #6484** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6484" target="_blank">Graphite</a> 
    * **PR #6486** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6486" target="_blank">Graphite</a> 
      * **PR #6487** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6487" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6487?utm_source=stack-comment).

ilyailya (approved):
LGTM!

Issue #6468, opened 2022-10-13
Support `Hash#dig` with intrinsic


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28x%3A%20T%3A%3AHash%5BString%2C%20T%3A%3AHash%5BString%2C%20T%3A%3AHash%5BString%2C%20String%5D%5D%5D%29.void%7D%0Adef%20example%28x%29%0A%20%20res1%20%3D%20x.dig%28%22%22%29%0A%20%20T.reveal_type%28res1%29%0A%20%20res2%20%3D%20x.dig%28%22%22%2C%20%22%22%29%0A%20%20T.reveal_type%28res2%29%0A%20%20res3%20%3D%20x.dig%28%22%22%2C%20%22%22%2C%20%22%22%29%0Aend)

```ruby
# typed: strict
extend T::Sig

sig {params(x: T::Hash[String, T::Hash[String, T::Hash[String, String]]]).void}
def example(x)
  res1 = x.dig("")
  T.reveal_type(res1)
  res2 = x.dig("", "")
  T.reveal_type(res2)
  res3 = x.dig("", "", "")
end
```

#### Observed output

```
editor.rb:7: Revealed type: T.untyped https://srb.help/7014
     7 |  T.reveal_type(res1)
          ^^^^^^^^^^^^^^^^^^^
  Got T.untyped originating from:
    editor.rb:6:
     6 |  res1 = x.dig("")
                 ^^^^^^^^^

editor.rb:9: Revealed type: T.untyped https://srb.help/7014
     9 |  T.reveal_type(res2)
          ^^^^^^^^^^^^^^^^^^^
  Got T.untyped originating from:
    editor.rb:8:
     8 |  res2 = x.dig("", "")
                 ^^^^^^^^^^^^^
Errors: 2
```

#### Expected behavior

- - -

We could also do something special for Shapes if we wanted, but that's likely to cause even more errors, so likely we want to hold off on doing anything special for shapes at the time being.

**Note**: `Array#dig` exists too. We should probably support this by recursively dispatching to the method called `dig` on the non-nil component of the method, if it exists. Here is the implementation in the Ruby VM:

[Hash#dig](https://github.com/ruby/ruby/blob/6847ee089d7655b2a0eea4fee3133aeacd4cc7cc/hash.c#L4403)
[...]

PR #6491, opened 2022-10-19
Implement `Array#dig` and `Hash#dig` intrinsic

jez:
### Motivation

<https://github.com/sorbet/sorbet/issues/6468>

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3697,7 +3697,7 @@ void digImplementation(const GlobalState &gs, const DispatchArgs &args, Dispatch
                            "dug a value that is always `{}` here",
                            "dig", "nil");
             auto endLastGood = args.locs.args[0].endPos();
-            auto endAll = args.locs.args[args.locs.args.size() - 1].endPos();
+            auto endAll = args.locs.args.back().endPos();
             auto replaceLoc = core::Loc(args.locs.file, endLastGood, endAll);
             if (endLastGood <= endAll && replaceLoc.exists()) {
                 e.replaceWith("Delete extra args", replaceLoc, "");
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3648,6 +3648,130 @@ class Magic_mergeHashValues : public IntrinsicMethod {
     }
 } Magic_mergeHashValues;
 
+namespace {
+
+void digImplementation(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res, NameRef methodToDigWith) {
+    if (args.args.size() == 0 || args.numPosArgs != args.args.size()) {
+        // A type error was already reported for arg mismatch
+        return;
+    }
+
+    auto baseCaseArgLocs = InlinedVector<LocOffsets, 2>{};
+    baseCaseArgLocs.emplace_back(args.locs.args[0]);
+    auto baseCaseLocs = CallLocs{
+        args.locs.file,                         // file
+        args.locs.call.copyWithZeroLength(),    // call
+        args.locs.args[0].copyWithZeroLength(), // receiver
+        args.locs.args[0].copyWithZeroLength(), // fun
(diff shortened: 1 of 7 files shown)


jez:

→ https://go/builds/bui_Me1jkvpSlKWnlg
→ https://go/builds/bui_Me1jytBNQW7Nr3

jez:

→ https://go/builds/bui_Me2VK5QEWfwzBc
→ https://go/builds/bui_Me2VG0DukBa9A3

test/testdata/infer/dig.rb
```diff
@@ -0,0 +1,86 @@
...
+    key: {
+      nested: T::Array[{x: Integer, y: Integer}]
+    },
+    another: Integer,
+  }
+end
+
+# Shapes are bad--ideally lot of these would be errors
+sig {params(opts: AShapeType).void}
+def example4(opts)
```
froydnj, line 45: The `TODO` in the corresponding exp file says we should have some tuple tests too.

froydnj (approved):

core/types/calls.cc
```diff
@@ -3648,6 +3648,130 @@ class Magic_mergeHashValues : public IntrinsicMethod {
...
+        return;
+    }
+
+    auto baseCaseArgLocs = InlinedVector<LocOffsets, 2>{};
+    baseCaseArgLocs.emplace_back(args.locs.args[0]);
+    auto baseCaseLocs = CallLocs{
+        args.locs.file,                         // file
+        args.locs.call.copyWithZeroLength(),    // call
+        args.locs.args[0].copyWithZeroLength(), // receiver
+        args.locs.args[0].copyWithZeroLength(), // fun
```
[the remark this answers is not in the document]
jez: I don't think so?

```ruby
opts.dig("first", "second")
     ^^^ fun
         ^^^^^^^ args[0]
```

Having it be zero length makes it basically ignore it for most things, and use the `callLoc` instead, except that it's very hard to get a case where it would need to use these errors, because we know that the `Array#at` / `Hash#[]` methods always exist, so we're unlikely to need the `fun` locs. The most important location in practice is the basecaseArgLocs in case you try to pass `0` to a `T::Hash[String, ...]` or something.

core/types/calls.cc
```diff
@@ -3648,6 +3648,130 @@ class Magic_mergeHashValues : public IntrinsicMethod {
...
+
+    auto digArgLocs = InlinedVector<LocOffsets, 2>{};
+    for (int i = 1; i < args.locs.args.size(); i++) {
+        digArgLocs.emplace_back(args.locs.args[i]);
+    }
+    auto digLocs = CallLocs{
+        args.locs.file,                         // file
+        args.locs.args[1],                      // call
+        args.locs.receiver,                     // receiver
+        args.locs.args[1].copyWithZeroLength(), // fun
```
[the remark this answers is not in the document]
jez: Again, I think this is not funLoc, due to above.

test/testdata/infer/dig.rb
jez: `example7`

test/testdata/infer/dig.rb
```diff
@@ -0,0 +1,115 @@
...
+  [Integer, [String, {x: Float}]]
+end
+
+sig {params(opts: ATupleType).void}
+def example7(opts)
+  # We don't have special handling for tuple types right now, so it decays to
+  # the underlying type. In the future we'll want to change this.
+
+  x = opts.dig("")
+  #        ^^ error: Expected `Integer` but found `String("")` for argument `arg0`
```
froydnj, line 98:
```suggestion
  x = opts.dig("")
  #            ^^ error: Expected `Integer` but found `String("")` for argument `arg0`
```

froydnj (approved):

PR #6492, opened 2022-10-20
update `vscode-languageclient` with LSP 3.17 support

froydnj:
### Motivation

Support for new protocol features requires upgrading this package.

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -4,7 +4,7 @@
   "description": "Ruby IDE features, powered by Sorbet.",
   "author": "Stripe Inc.",
   "license": "Apache-2.0",
-  "version": "0.3.16",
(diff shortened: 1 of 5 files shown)


froydnj:
Extension tests don't pass on my machine, but they do pass in CI...I guess this is just a hazard of trying to test GUI apps?  (Firefox had the same set of issues for many tests.)

jez (approved):

jez:
Did you test on stripe’s codebase? It’s really hard to roll back broken editor extensions

froydnj:
> Did you test on stripe’s codebase? It’s really hard to roll back broken editor extensions. Doing some quick test to make sure that typechecking and go to definition still work in the wild gets a lot of value.

Done!

PR #6502, opened 2022-10-23
Update T::Struct docs

jez:
### Motivation

This started out as a set of suggestions to make to #6462, but grew into
something larger, so I opened a new PR.

Closes #6462.

### Test plan

website-only change.

### DIFF
--- a/website/docs/tstruct.md
+++ b/website/docs/tstruct.md
@@ -4,70 +4,137 @@ title: Typed Structs via T::Struct
 sidebar_label: T::Struct
 ---
 
-> TODO: This page is still a fragment. Contributions welcome!
+Sorbet includes a way to define typed structs. They behave similarly to the
+[`Struct`] class built into Ruby, but work better with static and runtime type
+checking.
 
-While Sorbet supports using Hashes and Structs, it isn't always possible to
-model them statically.
+[`struct`]: https://docs.ruby-lang.org/en/master/Struct.html
 
-The `sorbet-runtime` gem ships with an alternative to using Ruby's `Hash` or
-`Struct` classes that is more well supported: `T::Struct`.
-
-## Creating a new struct class
-
-To use `T::Struct`, first create a new class that directly inherits from it.
-Then, in the body of the class, declare the fields on the struct.
+Here's a quick example:
 
 ```ruby
-# typed: true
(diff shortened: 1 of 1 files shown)


website/docs/tstruct.md
```diff
@@ -124,11 +191,286 @@ class ChildTwo < T::Struct
...
+and old properties and create the new instance. This means it suffers from
+exactly the same gotchas mentioned in the
+[`from_hash` gotchas](#from_hash-gotchas) section above.
+
+## Legacy and Stripe-specific options
+
+There are a number of other legacy or Stripe-internal options in the `prop` DSL.
+Those include `enum`, `foreign`, `ifunset`, `immutable`, `raise_on_nil_write`,
+`redaction`, and `sensitivity`. Stripe employees can reference
+[these docs](http://go/chalk-odm-docs) to learn more.
```
bdewater, line 472: Does `dont_store` also fall under here?

website/docs/tstruct.md
```diff
@@ -124,11 +191,286 @@ class ChildTwo < T::Struct
...
+  prop :foo, Integer
+end
+
+# (1) `serialize` converts from `T::Struct` to `Hash`
+serialized = A.new(foo: 42).serialize
+p(serialized) # => {"foo"=>42}
+
+# (1) `from_hash` converts from `Hash` to `T::Struct`
+deserialized = A.from_hash(serialized)
+p(deserialized) # => <A foo=42>
```
bdewater, line 283: Should this section mention `from_hash!`?

bdewater (approved):
Thank you for rewriting - I had no idea about the historical context. Looks great, only a couple of questions on a few things that didn't make the cut from my proposal.

website/docs/tstruct.md
jez: Sure, I can mention it.

Note to self: `from_hash!` is `from_hash(..., /* strict */true)` which requires that there be no unexpected keys in the Hash, but still doesn't do runtime type checking.

website/docs/tstruct.md
jez: Yes, thanks, I will add it to this list.

ilyailya (approved):
LGTM!

PR #6509, opened 2022-10-26
Add IRB::RelineInputMethod RBI definition

KaanOzkan:
Updates IRB RBIs to make its latest version compatible with tapioca. 

Reflects this change https://github.com/ruby/irb/pull/409/files#diff-947d919ba6e7132eb9acbdb77bd1df88b074c4e21039b550054aa3ffaf4a425dR264

### Motivation
IRB renamed the constant in the above PR, changing its parent class. It now conflicts with the payload 

### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/irb.rbi
+++ b/rbi/stdlib/irb.rbi
@@ -1523,7 +1523,9 @@ end
 
 class IRB::OutputMethod::NotImplementedError < ::StandardError; end
 
-class IRB::ReidlineInputMethod < ::IRB::InputMethod
+class IRB::ReidlineInputMethod < ::IRB::RelineInputMethod; end
+
+class IRB::RelineInputMethod < ::IRB::InputMethod
   include(::Reline)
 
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_MgdnyMrMnywc51
→ https://go/builds/bui_Mgdn5tSQJEhG93

rbi/stdlib/irb.rbi
```diff
@@ -1523,7 +1523,9 @@ end
 
 class IRB::OutputMethod::NotImplementedError < ::StandardError; end
 
-class IRB::ReidlineInputMethod < ::IRB::InputMethod
+class IRB::ReidlineInputMethod < ::IRB::RelineInputMethod; end
```
KaanOzkan, line 1526: Force pushed a change to remove the unnecessary `initialize` definition. So the Strip test will be slightly out of date @jez.

jez (approved):

PR #6516, opened 2022-10-28
Remove all custom autoload_map and intermediate node generation in autoloader

aadi-stripe:
As part of the ongoing migration of the Stripe codebase to use path-based autoloading with Zeitwerk, we update the autoloader shim generation logic to remove all "custom" autoloads.

Currently, the generation works as follows. Assume that there is a behavior-defining constant `Opus::Foo::Bar::Baz` which is defined in a file `lib/path/to/baz.rb`. Also assume that `Opus`, `Opus::Foo`, and `Opus::Foo::Bar` are pure namespace modules that don't define any behavior. Then, the following shim files will be produced:

* `Opus.rb` containing an `autoload_map` method call that maps `Opus::Foo` to `Opus/Foo.rb`
* `Opus/Foo.rb` containing an `autoload_map` method call that maps `Opus::Foo::Bar` to `Opus/Foo/Bar.rb`.
* `Opus/Foo/Bar.rb` containing an `autoload_map` method call that maps `Opus::Foo::Bar::Baz` to `Opus/Foo/Bar/Baz.rb`.
* `Opus/Foo/Bar/Baz.rb` which is the actual load-bearing shim that require-s `lib/path/to/baz.rb`.

If we use a Zeitwerk loader for autoloading code from these shim files, we don't need any `autoload_map` statements, since Zeitwerk does that for us implicitly.

We also don't require the presence of intermediate files for `Opus`, `Opus::Foo` or `Opus::Foo::Bar`; only the behavior-defining leaf node `Opus::Foo::Bar::Baz` needs a file to be generated at `Opus/Foo/Bar/Baz.rb`. This is because Zeitwerk lazily auto-vivifies modules based on the presence of directories at intermediate nodes -- it does not need intermediate files to be generated at these locations.

Thus, I'm making this change to the logic to remove all unnecessary generation. This will allow us to register these shim files with the Zeitwerk loader.

### Motivation

Making these changes has the following benefits:
* Allows us to more extensively leverage Zeitwerk at Stripe, moving us further forward in the migration.

* As we are currently in an in-between state with the migration, we have been working on ensuring interoperability between the two code-loading systems. Maintaining interoperability is a complex endeavor. We can simplify things by using Zeitwerk both for the source tree *and* these shims. We will not need any custom logic for interoperability, and the whole paradigm shifts from "shim-based vs path-based" to "path-based for shims vs path-based for source-tree". We will in effect switch to a 100% path-based world, with the only differentiating factor being where the paths are located.

* Removes a lot of unnecessary code generation. In the above example, we have reduced from 4 files for a single constant to 1 file. Extrapolating to the entire Stripe codebase, we are seeing that making this change removes ~40k shims. Note that there are two exceptional cases where intermediate files still need to be generated:
  - if the intermediate node is a class.
  - if the intermediate node is a module that defines behavior.

### Test plan

### DIFF
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -486,7 +486,7 @@ void populateAutoloadTasksAndCreateDirectories(const core::GlobalState &gs, vect
 }
 }; // namespace
 
-const core::FileRef DefTree::definingFile() const {
+core::FileRef DefTree::definingFile() const {
     core::FileRef definingFileRef;
 
     if (!namedDefs.empty() || (hasDef() && children.empty())) {
@@ -496,7 +496,7 @@ const core::FileRef DefTree::definingFile() const {
     return definingFileRef;
 }
 
-const bool DefTree::mustRender(const core::GlobalState &gs, std::string_view filePath) const {
+bool DefTree::mustRender(const core::GlobalState &gs, std::string_view filePath) const {
     // Either the node has a behavior-defining file, or has a package name
     if (definingFile().exists() || pkgName.exists()) {
         return true;
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -200,28 +200,24 @@ core::NameRef DefTree::name() const {
 
 string DefTree::renderAutoloadSrc(const core::GlobalState &gs, const AutoloaderConfig &alCfg) const {
     fmt::memory_buffer buf;
+    core::FileRef definingFileRef = definingFile();
+
     fmt::format_to(std::back_inserter(buf), "{}\n", alCfg.preamble);
 
-    core::FileRef definingFile;
-    if (!namedDefs.empty()) {
-        definingFile = file();
-    } else if (children.empty() && hasDef()) {
-        definingFile = file();
-    }
-    if (definingFile.exists()) {
+    if (definingFileRef.exists()) {
         requireStatements(gs, alCfg, buf);
     }
 
     string fullName = "nil";
     string casgnArg;
     auto type = definitionType(gs);
+
     if (type == Definition::Type::Module || type == Definition::Type::Class) {
         fullName = root() ? alCfg.rootObject
                           : fmt::format("{}", fmt::map_join(qname.nameParts, "::", [&](const auto &nr) -> string {
                                             return nr.show(gs);
                                         }));
         if (!root()) {
-            fmt::format_to(std::back_inserter(buf), "{}.on_autoload('{}')\n", alCfg.registryModule, fullName);
             predeclare(gs, fullName, buf);
         }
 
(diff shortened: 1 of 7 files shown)


test/cli/autogen-autoloader/test.out
```diff
@@ -51,7 +20,6 @@ Opus::Require.for_autoload(Foo::Bar::Jazz, "test/cli/autogen-autoloader/example1
 require 'in_class'
 require 'in_method'
 require 'my_gem'
-Opus::Require.on_autoload('Foo::Bar::Quuz')
```
aadi-stripe, line 54: We remove these explicit `on_autoload` callbacks as the Zeitwerk loader allows us to insert custom callbacks as a constant is loaded. We are already calling `on_autoload` there.

main/autogen/autoloader.cc
```diff
@@ -482,7 +467,10 @@ void populateAutoloadTasksAndCreateDirectories(const core::GlobalState &gs, vect
                                                const AutoloaderConfig &alCfg, string_view path, const DefTree &node) {
     string name = node.root() ? "root" : node.name().show(gs);
     string filePath = join(path, fmt::format("{}.rb", name));
-    tasks.emplace_back(RenderAutoloadTask{filePath, node});
+
+    if (node.mustRender(gs, filePath)) {
```
aadi-stripe, line 471: Use `mustRender` to figure out whether we want to schedule an autoload-rendering task for the node.

aadi-stripe:
Stripe builds

1. https://go/builds/bui_MhlPX90piZwtAV
2. https://go/builds/bui_MhmfD9diwehn4J
3. https://go/builds/bui_MhmfzxVRrOIHcq

main/autogen/autoloader.cc
```diff
@@ -516,6 +528,19 @@ void AutoloadWriter::writeAutoloads(const core::GlobalState &gs, WorkerPool &wor
...
+            std::string_view filePath = file;
+            int curDirPos = filePath.find_last_of('/');
+            while (curDirPos > 0) {
+                const auto curDir = filePath.substr(0, curDirPos);
+                if (curDir == path || !FileOps::removeEmptyDir(curDir)) {
+                    break;
+                }
+
+                curDirPos = filePath.find_last_of('/', curDirPos - 1);
+            }
```
aadi-stripe, line 543: Open to suggestions on writing this logic differently. In this new world, empty dirs have a minor runtime side-effect of generating dangling autoloads.

This would not be a problem in the latest version of Zeitwerk, which eagerly ignores empty dirs. However, it has its own perf unknowns which we haven't de-risked, so we can't bump to it just yet.

main/autogen/autoloader.cc
```diff
@@ -498,6 +486,30 @@ void populateAutoloadTasksAndCreateDirectories(const core::GlobalState &gs, vect
 }
 }; // namespace
 
+const core::FileRef DefTree::definingFile() const {
```
jez, line 489: btw, there's almost no value to having a `const` by-value return type, and sometimes can actually get in the way of optimizations.

https://stackoverflow.com/questions/8406898/benefits-of-using-const-with-scalar-type-e-g-const-double-or-const-int

https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value

jez (approved):

ngroman (approved):
👍

PR #6520, opened 2022-10-29
wip: Crash in static init method locs

jez:
### Motivation

The more test cases I find like this, the more I wonder if we should
have something akin to our lsp_test_runner fast path tests, but instead
of adding one `\n` for each character in the file and running the fast
path, start the file with that many `\n` characters and run a fast path
where we delete them all. (Unfortunately, I think this means running
another full slow path edit, and then another fast path edit after, for
each test, which could substantially slow down our tests. Maybe it's the
kind of thing that we do just once to find the problems, commit
individual tests where relevant, and then don't commit any changes to
the test harness? We could at least build it and see how much slower
things get.)

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/fast_path/static_init_method_locs.1.rbupdate
+++ b/test/testdata/lsp/fast_path/static_init_method_locs.1.rbupdate
@@ -1,0 +1,8 @@
+# typed: true
+
+class A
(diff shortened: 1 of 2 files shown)


jez:
Closing in favor of #7232

PR #6521, opened 2022-10-29
Enter proper FoundClass definitions for all scopes

jez:
### Motivation

This lets us

- Remove FoundClassRef
- Delete a bunch of code in namer to handle "what if the scope of a
  class or static field is actually a static field"
- Ensure that class symbols are _never_ mangled, properly this time, which is
  important for the fast path constants change in #6422.

This also comes with assorted other improvements (like some SymbolRef ->
ClassOrModuleRef changes).

### Test plan

See included automated tests.


jez:
* jez-delete-dealias
  * **PR #6521** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6521" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6521?utm_source=stack-comment).

core/FoundDefinitions.h
```diff
@@ -82,43 +78,33 @@ class FoundDefinitionRef final {
...
+        Unknown,
         Module,
         Class,
     };
     Kind classKind;
 
     std::string toString(const core::GlobalState &gs, const FoundDefinitions &foundDefs, uint32_t id) const;
 };
-CheckSize(FoundClass, 28, 4);
+CheckSize(FoundClass, 24, 4);
```
froydnj, line 104: I am surprised at this change...`FoundDefinitionRef` and `NameRef` are the same size...so this structure shouldn't change size?

I could see `FoundStaticField` shrinking, though.

namer/namer.cc
```diff
@@ -726,58 +701,6 @@ class SymbolDefiner {
...
-        if (scope.isClassOrModule()) {
-            ENFORCE(!isMangleRenameUniqueName(ctx, scope.name(ctx)));
-            return scope.asClassOrModuleRef();
-        }
-
-        // Check if class was already mangled.
-        auto klassSymbol = ctx.state.lookupClassSymbol(scope.owner(ctx).asClassOrModuleRef(), scope.name(ctx));
-        if (klassSymbol.exists()) {
-            ENFORCE(isMangleRenameUniqueName(ctx, klassSymbol.data(ctx)->name));
-            if (auto e = ctx.beginError(loc, core::errors::Namer::InvalidClassOwner)) {
```
froydnj, line 741: I guess a followup PR can comment out this error at its definition?

froydnj (approved):
This makes sense to me.

OOC, does changing class symbols to never be mangled mean that a hypothetical world where classes get put on the fast path is much harder?  Or is it maybe "easier", because we know where the invariants for non-mangled class names exist, and it's easier to undo them?

jez:
> OOC, does changing class symbols to never be mangled mean that a hypothetical world where classes get put on the fast path is much harder? Or is it maybe "easier", because we know where the invariants for non-mangled class names exist, and it's easier to undo them?

I think that the hardest part of classes taking the fast path is figuring out which files to retypecheck. I think that once the set of files is correct, mutating GlobalState to add/remove/edit class symbols is not _that_ hard (and if anything, likely to be easier, because we still know that it's possible to handle all class symbols before all non-class symbols).

Unfortunately, the more I think about it, the more I think that class or modules on the fast path requires a somewhat major reworking of our "which files to retypecheck" heuristic.

namer/namer.cc
jez: What do you mean?

This error no longer happens, because the "can't nest under" error never happens. The scope of a class def or const assign is now always a class symbol by construction.

What happens is that _other_ code starts to error, saying that you can't initialize a class or module by constant assignment:

```ruby
X = 1 # <- new error is here ("can't initialize class `X` by assignment")
X::Y = 1 # <- old error used to be here ("can't nest Y under static field X")
```

jez: There are some scripts that suppress `4015`. I'd rather not have to figure out a way to update that in a backwards-compatible way, so I think I'll just leave the error code around but unused, if you don't mind.

core/FoundDefinitions.h
jez: good point, and the build failed actually

jez:
@froydnj heads up this changed quite a bit since the original review, so I'm handing it back for a second.

some important changes:

- I had to lurk a hack to skip calling `addLoc` for `isUnknown` class defs, because otherwise it would have added `O(files)` locs to `Opus`
- Some of the tests failed because we assumed that the `ownerIdx` in the `Found*Hash` classes was always of kind `Class`, but after this change it is sometimes `Symbol` (usually `core::Symbols::root()`). This was as good a forcing function as any to start enforcing our assumption about which kinds of `FoundDefinitionRef`'s are allowed to be in the `ownerIdx`. I solved this with a gross macro but if you want we can try to solve it somehow else. Importantly, I would have run into the exact same problem on #6422 eventually, so it was kind of a happy accident that we ran into it in service of getting this change out. The implementation is in a commit rather than a PR because I didn't want to deal with more PRs and also because technically until this PR the distinction doesn't matter.
- A lot of exp files change, because we call `->singletonClass()` sooner than we used to, which causes some locs to be different now.

jez:

→ https://go/builds/bui_MivQljRyhYDYHg
→ https://go/builds/bui_MivQSQPGppB8Ij

jez:

→ https://go/builds/bui_MiwuHn5Va6pGlG
→ https://go/builds/bui_MiwvbAT4yGZS1x

jez:

→ https://go/builds/bui_Mj0QJK3r6TElhs
→ https://go/builds/bui_Mj0QTwnDiciD2Q

jez:

→ https://go/builds/bui_Mj1FG2uKGPVnjB
→ https://go/builds/bui_Mj1FZcVorOkBFN

jez:

→ https://go/builds/bui_Mj1NfYwpeBvyVD
→ https://go/builds/bui_Mj1NEERURonffO

jez:

→ https://go/builds/bui_MjGes6eA0Bet56
→ https://go/builds/bui_MjGe98glu3ye5O

jez:

→ https://go/builds/bui_MjI7ea86E05cl3
→ https://go/builds/bui_MjI8uDBpbR2l5Z

jez:

→ https://go/builds/bui_Mk3hBvTHoUwRr2
→ https://go/builds/bui_Mk3hCoKmfNYdJX

froydnj (approved):

PR #6522, opened 2022-10-29
Add types for `Enumerator::Chain`

georgebrock:
This PR adds RBI type definitions for the standard library's `Enumerator::Chain` class, including a `T::Enumerator::Chain` metatype. 

The change follows the example of the existing `Enumerator::Lazy` and `T::Enumerator::Lazy` support.

### Motivation

Prior to this change there were no types for `Enumerator::Chain` in Sorbet's standard library, which meant that calling `Enumerator::Chain.new` would result in an invalid constant error.

### Test plan

See included automated tests.

`bazel test test --config=dbg` is passing locally, but I never managed to get `bazel test //... --config=dbg` to work (I hit build errors even before I made any changes).


jez:

→ https://go/builds/bui_N7FhOOFW3xi1PL
→ https://go/builds/bui_N7FhdR9hJVCuKo

gems/sorbet-runtime/lib/types/types/type_enumerator_chain.rb
```diff
@@ -0,0 +1,41 @@
...
+      def initialize
+        super(T.untyped)
+      end
+
+      def valid?(obj)
+        obj.is_a?(Enumerator::Chain)
+      end
+    end
+  end
+end
```
jez, line 41: is one of these files misnamed? it looks like you have a duplicate file

infer/environment.cc
```diff
@@ -921,6 +921,9 @@ core::TypePtr flatmapHack(core::Context ctx, const core::TypePtr &receiver, cons
     if (!receiver.isUntyped() && receiver.derivesFrom(ctx, core::Symbols::Enumerator_Lazy())) {
         return returnType;
     }
+    if (!receiver.isUntyped() && receiver.derivesFrom(ctx, core::Symbols::Enumerator_Chain())) {
```
jez, line 924: do you mind collapsing this condition and the previous with an `||` so we skip a second call to `isUntyped`?

(should be fast anyways but `¯\_(ツ)_/¯`)

jez (approved):
This looks perfect, and on your first try too. Thanks and sorry for taking so long to get around to reviewing and testing this.

PR #6523, opened 2022-10-30
Require block arg type in signature

### LINKED ISSUE #5670: Not required to write type of `blk` arg in sig
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0Asig%20%7Bvoid%7D%0Adef%20foo%28%26blk%29%3B%20end)

```ruby
# typed: strict
extend T::Sig
sig {void}
def foo(&blk); end
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

We're in a `# typed: strict` file, and we explicitly named the `blk` arg, but we have not been required to put the `blk`'s type in the `params` list in the `sig`.

For all other args, Sorbet requires that if they appear in the method def they also appear in the sig.

- - -

thomasmarshall:
This PR ensures explicit block arguments must be defined in the signature:

```rb
sig {void}
def foo(&blk); end
#        ^^^ error: Malformed `sig`. Type not specified for argument `blk`
```

The following will not produce an error:

```rb
sig {params(blk: T.proc.void).void}
def foo(&blk); end
```

Every method has a "synthetic" block argument if an explicit one isn't defined, for which an argument type **shouldn't** be required in the signature. We were previously determining if a block arg was synthetic or not based on its name reference. However, block arguments always have the name reference `<blk>` whether they are explicitly defined or synthetically added. This might not be intended, but is currently how it works:

https://github.com/sorbet/sorbet/blob/9adecfa7bffc4ffb6f72f806161709d6b97681af/namer/namer.cc#L801-L802

This PR swaps the name comparison for `isSyntheticBlockArgument`, which has a more robust way of determining if the block is synthetic or not, based on whether a `Loc` exists. Synthetic blocks use `Loc::none()`:

https://github.com/sorbet/sorbet/blob/9adecfa7bffc4ffb6f72f806161709d6b97681af/core/Symbols.cc#L2019-L2022

An alternative might be to change the namer so explicit block arguments are not named `<blk>`, but leaning on the `isSyntheticBlockArgument` seemed a smaller and simpler change.

### Motivation

Fixes #5670. The motivation for the change is that all other arguments must have argument types if the signature is present, but block arguments are apparently ignored.

:wave: I picked this up as a "good first issue" as it seemed relatively straightforward to understand and limited in scope.

#### Notes

- The issue above mentions `typed: strict` but this change also applies in `typed: true` if a signature is defined, as it does for other argument types. Is this correct?
- The code was already in place for this to work, but the check for a "synthetic block" was apparently always returning true. Should we instead make that check work by changing `namer.cc`?
- I suspect this will produce errors in some codebases that previously didn't have them (as it did in some RBIs and tests here). Is that OK?

### Test plan

See included automated tests.

### DIFF
--- a/rbi/sorbet/tprops.rbi
+++ b/rbi/sorbet/tprops.rbi
@@ -138,10 +138,10 @@ module T::Props::PrettyPrintable
 end
 
 module T::Props::PrettyPrintable::DecoratorMethods
-  def inspect_instance(instance, multiline: false, indent: '  ', &blk); end
-  def inspect_instance_components(instance, multiline:, indent:, &blk); end
-  def inspect_prop_value(instance, prop, multiline:, indent:, &blk); end
-  def join_props_with_pretty_values(pretty_kvs, multiline:, indent: '  ', &blk); end
+  def inspect_instance(instance, multiline: false, indent: '  '); end
+  def inspect_instance_components(instance, multiline:, indent:); end
+  def inspect_prop_value(instance, prop, multiline:, indent:); end
+  def join_props_with_pretty_values(pretty_kvs, multiline:, indent: '  '); end
   def self.method_added(name); end
   def self.singleton_method_added(name); end
   def valid_rule_key?(key); end
--- a/rbi/sorbet/tprops.rbi
+++ b/rbi/sorbet/tprops.rbi
@@ -22,7 +22,7 @@ end
 
 module T::Props::ClassMethods
   sig {params(name: Symbol, cls_or_args: T.untyped, args: T::Hash[Symbol, T.untyped]).void}
-  def const(name, cls_or_args, args={}, &blk); end
+  def const(name, cls_or_args, args={}); end
   sig {params(name: Symbol, cls: T.untyped, rules: T.untyped).void}
   def prop(name, cls, rules = nil); end
   def decorator; end
--- a/rbi/stdlib/openssl.rbi
+++ b/rbi/stdlib/openssl.rbi
@@ -1355,7 +1355,12 @@ class OpenSSL::ASN1::Constructive < OpenSSL::ASN1::ASN1Data
   #   puts asn1
   # end
   # ```
-  sig {returns(::T.untyped)}
+  sig do
+    params(
+      blk: T.nilable(T.proc.returns(::T.untyped)),
+    )
+    .returns(::T.untyped)
+  end
   def each(&blk); end
 
   sig do
@@ -2076,6 +2081,7 @@ module OpenSSL::Buffering
   sig do
     params(
       eol: ::T.untyped,
+      blk: T.nilable(T.proc.returns(::T.untyped)),
     )
     .returns(::T.untyped)
   end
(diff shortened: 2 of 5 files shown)


froydnj:

→ https://go/builds/bui_MiStEgJIoBidFj
→ https://go/builds/bui_MiStfRCE2frbmg

jez (approved):
Thanks for the change!

- - - - -

> * The issue above mentions `typed: strict` but this change also applies in `typed: true` if a signature is defined, as it does for other argument types. Is this correct?

```ruby
# typed: true
extend T::Sig

sig {void}
def example
  yield
end

example do
  puts('hello')
end
```

This is not an error in `# typed: true` but is an error in `# typed: strict`, even on your branch. At the moment, this is intended for ease of migration on legacy codebases, but we could reconsider this behavior in the future. (This would be good to add a test for, btw.)

> * I suspect this will produce errors in some codebases that previously didn't have them (as it did in some RBIs and tests here). Is that OK?

That depends on how many errors, and how easy it is to fix the problem.

For the question of "how many," it looks like Stripe's codebase of multiple millions of lines of code only has 6 new errors because of this change, so I'm inclined to say that we're good to go here.

If you wanted to make it even more palatable to roll this change out, you could consider adding an autocorrect for this error message, but I think that's both tricky (going to take some time to implement) and also likely to be rare (as evidenced by the 6 errors above), so I wouldn't block landing the change on it.

> * The code was already in place for this to work, but the check for a "synthetic block" was apparently always returning true. Should we instead make that check work by changing namer.cc?

I'm sorry, I don't know what you mean.

jez (changes requested):
I notice that in certain situations, the error message is less than idea:

```ruby
# typed: true
class Example
  extend T::Sig
  sig {params(args: T.untyped).void}
  def foo(args); end
end
class Example
  def foo(args, &my_custom_block_name); end
end
```

The error message here:

```
❯ sorbet foo.rb
foo.rb:8: Malformed sig. Type not specified for argument <blk> https://srb.help/5003
     8 |  def foo(args, &my_custom_block_name); end
                         ^^^^^^^^^^^^^^^^^^^^
    foo.rb:4: Signature
     4 |  sig {params(args: T.untyped).void}
Errors: 1
```

Do you think you could tweak the code to use the name that comes from the method def instead of from the symbol, so that the error message always matches what the user wrote?

thomasmarshall:
> This is not an error in `# typed: true` but is an error in `# typed: strict`, even on your branch.

Ah OK, but the example from #5670 _is_ now an error for both `# typed: true` and `# typed: strict`. I think that's correct, given it's the same behaviour as other arguments, but I just wanted to double check my understanding of the issue. I've added a test for unnamed block/yield.

> I'm sorry, I don't know what you mean.

I mean the `bool isSyntheticBlkArg` variable already existed but the comparison `arg.name == core::Names::blkArg()` was incorrect. I thought maybe that comparison is _supposed_ to be correct, and that we should be able to identify a synthetic block by inspecting the arg name. I was speculating that the bug might be in the thing that sets the argument name, rather than where I've made the change. But I'm happy that `isSyntheticBlockArgument` works well!

> Do you think you could tweak the code to use the name that comes from the method def instead of from the symbol, so that the error message always matches what the user wrote?

I've added a fixup and test for this now:

```
foo.rb:8: Malformed sig. Type not specified for argument my_custom_block_name https://srb.help/5003
     8 |  def foo(args, &my_custom_block_name); end
                         ^^^^^^^^^^^^^^^^^^^^
    foo.rb:4: Signature
     4 |  sig {params(args: T.untyped).void}
Errors: 1
```

PR #6528, opened 2022-11-01
make dependencies in `common/` finer-grained

froydnj:
### Motivation

I want to parallelize file-finding during option parsing, partly for speed reasons and partly because other LSP-related changes might wind up having to entirely rescan for files, and it would be nicer if the latter made a passing attempt to be fast.

The first blocker here is that we have two Bazel components:

- `//common`
- `//common/concurrency`

Parallelizing file-finding requires making the first depend on the second, but the second already depends on the first, and Bazel won't let us introduce cycles in our dependency graph.

Therefore, we need to split out some components that just `//common/concurrency` can depend on without pulling in all of `//common`.  In a future PR, then, we can add the dependency on `//common/concurrency` to `//common` without introducing cycles.

This is all relatively straightforward, if a little tedious.  The weirdest part of the change is "[untangle common/os/os.cc from exception infrastructure with a hammer](https://github.com/sorbet/sorbet/commit/88f44510ba53978132f8c370c870ee85979affe9)", which is mostly me being lazy about not splitting out more specific OS-related dependencies (debugging primitives, threading primitives, etc.) so that we could use `common/exception/Exception.h` etc. in `common/os/`.  If we would like to avoid introducing bare `throw logic_error`, I can work on splitting things out.

### Test plan

Ideally, if this builds, everything should Just Work.


common/os/os.cc
```diff
@@ -22,7 +21,7 @@ void *Joinable::trampoline(void *ptr) {
 
 unique_ptr<Joinable> runInAThread(string_view threadName, function<void()> function, optional<int> bindToCore) {
 #ifdef EMSCRIPTEN
-    sorbet::Exception::raise("Creating threads in unsupported in EMSCRIPTEN");
+    throw logic_error("Creating threads in unsupported in EMSCRIPTEN");
```
jez, line 24: These exception kinds are caught in main but maybe it’s fine to ignore that?

jez (approved):

common/os/os.cc
froydnj: Unfortunately they're not. :(  I can take a note to either catch them or do a fine-grained separation of things; it may be easier to just catch them.

PR #6530, opened 2022-11-01
parallelize file discovery

froydnj:
### Motivation

Finding all the files with a serial recursive walk through the input directory takes a long time on Stripe's codebase.  This makes it significantly faster.

I did not test what this does on a machine with a single core; I did experiment at one point with making `appendFilesInDir` use an explicit stack and therefore be iterative rather than recursive.  There was no speedup in that case, but neither was there a slowdown.  I'd expect roughly the same thing to happen here, maybe with slightly higher overhead for the concurrent queue rather than a vector for holding the directories to scan.

### Test plan

I tested this on Stripe's codebase and verified that it was processing the same number and kind of files (e.g. number of `typed: strict`) as before; typecheck results were also identical (including introducing various errors), so I have some confidence that we're not finding completely different files.

### DIFF
--- a/common/FileOps.h
+++ b/common/FileOps.h
@@ -7,6 +7,7 @@
 #include <vector>
 
 namespace sorbet {
+class WorkerPool;
 
 class FileOps final {
 public:
@@ -49,7 +50,7 @@ public:
      * Throws FileNotFoundException if path does not exist, and FileNotDirException if path is not a directory.
      */
     static std::vector<std::string> listFilesInDir(std::string_view path, const UnorderedSet<std::string> &extensions,
-                                                   bool recursive,
+                                                   WorkerPool &workers, bool recursive,
                                                    const std::vector<std::string> &absoluteIgnorePatterns,
                                                    const std::vector<std::string> &relativeIgnorePatterns);
     /**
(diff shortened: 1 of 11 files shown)


common/FileOps.h
```diff
@@ -43,7 +44,7 @@ class FileOps final {
      * Throws FileNotFoundException if path does not exist, and FileNotDirException if path is not a directory.
```
jez, line 44: maybe mention in this doc comment that if you need to use this single threaded, you can create an empty worker pool?

jez: oh nevermind, there appears to be an override for this already.

jez (approved):

jez:
> I tested this on Stripe's codebase and verified that it was processing the same number and kind of files (e.g. number of `typed: strict`) as before; typecheck results were also identical (including introducing various errors), so I have some confidence that we're not finding completely different files

If you want, you can use `--print=file-table-json` from before and after to see that file tables are the same.

froydnj:

→ https://go/builds/bui_MjDDEFdnfr2Cof
→ https://go/builds/bui_MjDDHLQfzgqPqO

PR #6535, opened 2022-11-02
Add possibility to pass T::Structs as value for #with

leifg:
The solution is rather straightforward:

In `#recursive_stringify_keys` add another branch to check for wether the value is a `T::Struct`. In this case just call `serialize`.

### Motivation
Using the `#with` helper on T::Struct can be really useful especially when using immutable objects. Typical example is having an address or money type:

```ruby
class Money < T::Struct
  const :amount, Integer
  const :currency, String
end

class Order < T::Struct
  const :value, Money
end


o = Order.new(value: Money.new(amount: 100_00, currency: "USD"))
o.with(value: Money.new(amount: 200_00, currency: "USD"))
```

Not only is this not possible, the error message is also very cryptic:

```
/gems/ruby/3.1.0/gems/sorbet-runtime-0.5.10526/lib/types/props/serializable.rb:72:in `rescue in rescue in deserialize': Error in Order#__t_props_generated_deserialize: <Money amount=20000, currency="USD"> provided to from_hash (TypeError)
at line 8 in:
    found = 1
    val = hash["value"]
  @value = if val.nil?
    found -= 1 unless hash.key?("value")
    self.class.decorator.raise_nil_deserialize_error("value")
  else
    begin
    Money.from_hash(val)
  rescue NoMethodError => e
    raise_deserialization_error(
      :value,
      val,
      e,
    )
    val
  end
/gems/ruby/3.1.0/gems/sorbet-runtime-0.5.10526/lib/types/props/serializable.rb:70:in `rescue in deserialize': Error in Order#__t_props_generated_deserialize: <Money amount=20000, currency="USD"> provided to from_hash (ArgumentError)
at line 8 in:
    found = 1
    val = hash["value"]
  @value = if val.nil?
    found -= 1 unless hash.key?("value")
    self.class.decorator.raise_nil_deserialize_error("value")
  else
    begin
    Money.from_hash(val)
  rescue NoMethodError => e
    raise_deserialization_error(
      :value,
      val,
      e,
    )
    val
  end
/gems/ruby/3.1.0/gems/sorbet-runtime-0.5.10526/lib/types/props/serializable.rb:209:in `from_hash': <Money amount=20000, currency="USD"> provided to from_hash (ArgumentError)
```

For the user there is a workaround by just calling `#serialize` on the passed in value:

```ruby
o.with(value: Money.new(amount: 200_00, currency: "USD").serialize)
=> <Order value=<Money amount=20000, currency="USD">>
```

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/serializable.rb
+++ b/gems/sorbet-runtime/lib/types/props/serializable.rb
@@ -105,7 +105,9 @@ module T::Props::Serializable
   end
 
   private def recursive_stringify_keys(obj)
-    if obj.is_a?(Hash)
+    if obj.is_a?(T::Struct)
+      new_obj = recursive_stringify_keys(obj.serialize)
+    elsif obj.is_a?(Hash)
       new_obj = obj.class.new
       obj.each do |k, v|
         new_obj[k.to_s] = recursive_stringify_keys(v)
--- a/gems/sorbet-runtime/test/types/props/serializable.rb
+++ b/gems/sorbet-runtime/test/types/props/serializable.rb
@@ -513,6 +513,18 @@ class Opus::Types::Test::Props::SerializableTest < Critic::Unit::UnitTest
       assert_equal('foo', b.f2.foo)
       assert_equal(10, b.f2.bar)
     end
+
+    it 'with nested fields as model' do
+      a = WithModel2.new(f1: 'foo')
+      b = a.with(f2: WithModel1.new(foo: 'foo', bar: 10))
+
+      assert_equal('foo', a.f1)
+      assert_nil(a.f2)
+      assert_equal('foo', b.f1)
+      refute_nil(b.f2)
+      assert_equal('foo', b.f2.foo)
+      assert_equal(10, b.f2.bar)
+    end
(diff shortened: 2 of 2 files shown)


jez:

→ https://go/builds/bui_N7gfAJdklcZZmG
→ https://go/builds/bui_N7gf0ohANTfXyT

jez:
Sorry: this change causes 55,000 test failures on Stripe's codebase.

T::Struct and the underlying T::Prop APIs are very tightly coupled to Stripe's internal database libraries. Changing them is difficult to impossible while preserving backwards compatibility, because of how much code depends on the way it currently works.

There is more context written up here:

- [Legacy Code and Historical Context](https://sorbet.org/docs/tstruct#legacy-code-and-historical-context)
- [Immutable property updates using `with`](https://sorbet.org/docs/tstruct#immutable-property-updates-using-with)

Given that, we're going to have to decline this change. Thanks and best wishes.

- - - - -

For posterity, Stripe employees can see the runtime test failures here:

http://go/builds/bui_N7xbHxkRFHrrU2

PR #6537, opened 2022-11-04
Remove definedMethods from namer

jez:
### Motivation

Seems like we were only writing to this, never reading from it.

### Test plan

Existing tests


froydnj (approved):

namer/namer.cc
```diff
@@ -1556,10 +1554,9 @@ class SymbolDefiner {
                 //
                 // We still need to put something here so that other found definitions that
                 // reference methods will get the correct symbols.
```
jez, line 1556:
```suggestion
```

PR #6540, opened 2022-11-04
Fix constant redefinition across multiple files

jez:
### Motivation

The test included on this branch previously crashed. There was an ENFORCE that
assumed namer works the way it works the way it finally does in this PR.
Specifically, we assumed that we handled all class definitions before all
non-class definitions. When the code all lives in a single file that was the
case, but the test case seen in this PR shows a case where the static field
definition is processed before the class def of the same name in another file.

Honestly, this is code is gross and once #6422 lands we can delete the most of
the lines added to `defineSymbols` in this PR, because in that PR
`enterNonDeletableMethods` becomes empty.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_Mk3dy9ja205s9P
→ https://go/builds/bui_Mk3d9uoTmcMLQ5

froydnj (approved):
"assumed namer works the way it works the way it finally does in this PR"

😅

PR #6545, opened 2022-11-07
don't generate autoloader source twice

froydnj:
### Motivation

Less work is less work, even when we're doing things in parallel.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_Ml5vmu1bk7NXVU
→ https://go/builds/bui_Ml5vQKnMSKyKlk

jez (approved):
🤦‍♂️

PR #6546, opened 2022-11-07
don't construct unnecessary strings during autoloader generation

froydnj:
### Motivation

We can generate `string_view` directly out of `fmt::memory_buffer`, which is all we need for writing the relevant files.

While we're here, let's also re-use a single buffer throughout the life of the worker thread.

(Full disclosure: this PR initially failed with an ASAN error in a non-autogen-related CLI test...retrying the tests worked. :ghost:)

### Test plan

See included automated tests.

### DIFF
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -198,7 +198,7 @@ core::NameRef DefTree::name() const {
     return qname.name();
 }
 
-string DefTree::renderAutoloadSrc(fmt::memory_buffer &buf, const core::GlobalState &gs, const AutoloaderConfig &alCfg) const {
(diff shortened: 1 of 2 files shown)


jez:
> (Full disclosure: this PR initially failed with an ASAN error in a non-autogen-related CLI test...retrying the tests worked. 👻)

Can you try with `--runs_per_test=20` on the relevant test to double check whether it's flaky or whether it was a one-time blip?

jez (approved):
I would love to double check what's up with the sanitizer failure by attempting to repro locally with `--runs_per_test=20` (or maybe higher)

froydnj:
I tried with `--runs_per_test=100` (`--config=test-sanitized-linux`) and got nothing locally.  I don't think that test is even *executing* code that got modified here, so I'm not sure what was up with that single failure.

PR #6547, opened 2022-11-07
Add fast path test that currently fails

jez:
### Motivation

I first noticed this bug while working on #6422.

This test currently shows incorrect behavior. As far as I can tell, this
has been incorrect for as long as Sorbet has had a fast path mode. The
fact that this bug is only appearing now means that maybe we can
tolerate it a while longer?

In any case, we will want to fix this eventually.

### Test plan

Test case is written so that it passes, but it captures incorrect behavior.


froydnj (approved):
Tests like this make me wish we had XFAIL or something like that.

jez:
> Tests like this make me wish we had XFAIL or something like that.

We built it for compiler tests and didn't backport it to normal testdata tests. It shouldn't be too hard if we want to spend time on it.

PR #6549, opened 2022-11-08
Set result value for ignored files in document formatting

reese:
### Motivation

Currently, the document formatting request ignores certain files (rbis, package files, etc.), but it does so without actually setting the `result` value on the response, so the client throws an error. This fixes this by return `result: null` for cases where we ignore the file.

### Test plan

I manually tested this by running format-on-save on a `__package.rb` file locally


jez (approved):
is it possible to write a test for this?

reese:
Added a test with an `rbi` file and checked that it failed prior to this change and passed afterward (`bazel test --config=dbg //test:test_LSPTests/testdata/lsp/rubyfmt_rbi_test`)

PR #6557, opened 2022-11-10
`if` -> `switch` for TypecheckingPath

jez:
### Motivation

This is another change that Ilya made on his #6461 PR that can be pulled
out as a pre-work PR. Soon this will have three values.

Diff looks best ignoring whitespace.

### Test plan

Existing tests

PR #6556, opened 2022-11-10
Replace PathType with TypecheckingPath

jez:
### Motivation

We have two enums that are always going to want to be in sync--one generated in
`make_lsp_types.cc` so that we can generate an enum that will be used in one of
Sorbet's custom LSP extensions. One of them was maintained by hand.

As a part of Ilya's work in #6461, he had a commit that factored this piece of
code out.

I'm trying to revive that PR while he's busy with other work, but there's some
tricky conflicts. Luckily makes sense independently and doesn't conflict.

### Test plan

Existing tests

PR #6558, opened 2022-11-10
Factor out errorFlusher argument to runSlowPath

jez:
### Motivation

The `runFastPath` and some of the methods it calls take an explicit
`errorFlusher` argument, but `runSlowPath` doesn't. In a future change,
we're going to want to start calling those same APIs from `runSlowPath`.
So it's convenient to make these two methods have basically the same
signature.

This should be a no-op, because every instance of an errorFlusher used
in this file is always constructed in the same way with values that
would be the same at runtime to those that are used in this PR.

### Test plan

Existing tests

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -121,7 +121,7 @@ void LSPTypechecker::initialize(TaskQueue &queue, std::unique_ptr<core::GlobalSt
     {
         const bool isIncremental = false;
         ErrorEpoch epoch(*errorReporter, updates.epoch, isIncremental, {});
-        auto errorFlusher = make_shared<ErrorFlusherLSP>(epoch, errorReporter);
+        auto errorFlusher = make_shared<ErrorFlusherLSP>(updates.epoch, errorReporter);
         committed = runSlowPath(move(updates), workers, errorFlusher, /* cancelable */ false);
         epoch.committed = committed;
     }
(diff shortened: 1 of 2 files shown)


jez:
* master
  * **PR #6556** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6556" target="_blank">Graphite</a> 
    * **PR #6557** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6557" target="_blank">Graphite</a> 
      * **PR #6558** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6558" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6558?utm_source=stack-comment).

main/lsp/LSPTypechecker.cc
```diff
@@ -121,7 +121,8 @@ void LSPTypechecker::initialize(TaskQueue &queue, std::unique_ptr<core::GlobalSt
     {
         const bool isIncremental = false;
         ErrorEpoch epoch(*errorReporter, updates.epoch, isIncremental, {});
-        committed = runSlowPath(move(updates), workers, /* cancelable */ false);
+        auto errorFlusher = make_shared<ErrorFlusherLSP>(epoch, errorReporter);
```
froydnj, line 124:
```suggestion
        auto errorFlusher = make_shared<ErrorFlusherLSP>(updates.epoch, errorReporter);
```

froydnj (approved):
Assuming the below is the only thing wrong, LGTM.

PR #6565, opened 2022-11-13
Desugar step removes all method bodies for autogen

aadi-stripe:
In the Stripe codebase, our core autogen pass (the one that does not print parsed file information into msgpack) only needs to know about definitions and ancestry information. It does not need to walk over methods at all. Given this, we can empty out all method bodies and achieve the same result with a decent speedup in our tooling.

Based on some (slightly noisy) measurements in my local environment, I saw a 1s-3s improvement in the Sorbet autogen run, amounting to a 6%-20% improvement.

### Motivation

Performance.

### Test plan

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -320,11 +320,14 @@ ExpressionPtr buildMethod(DesugarContext dctx, core::LocOffsets loc, core::LocOf
 
     bool removeBody = dctx.ctx.state.autogenPrintingSubclassesOrAutoloaderOnly;
     ExpressionPtr desugaredBody;
-    if (!removeBody) {
+    if (removeBody) {
+        // For performance reasons, we remove method bodies in autogen when only generating
+        // autoloader files or subclass lists. In this configuration, passing method bodies
+        // is not required for correctness.
+        desugaredBody = MK::EmptyTree();
+    } else {
         desugaredBody = desugarBody(dctx2, loc, body, std::move(destructures));
         desugaredBody = validateRBIBody(dctx2, move(desugaredBody));
-    } else {
-        desugaredBody = MK::EmptyTree();
     }
 
     auto mdef = MK::Method(loc, declLoc, name, std::move(args), std::move(desugaredBody));
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -301,7 +301,8 @@ ExpressionPtr validateRBIBody(DesugarContext dctx, ExpressionPtr body) {
 }
 
 ExpressionPtr buildMethod(DesugarContext dctx, core::LocOffsets loc, core::LocOffsets declLoc, core::NameRef name,
-                          unique_ptr<parser::Node> &argnode, unique_ptr<parser::Node> &body, bool isSelf) {
+                          unique_ptr<parser::Node> &argnode, unique_ptr<parser::Node> &body, bool isSelf,
+                          bool removeBody) {
     // Reset uniqueCounter within this scope (to keep numbers small)
     uint32_t uniqueCounter = 1;
     DesugarContext dctx1(dctx.ctx, uniqueCounter, dctx.enclosingBlockArg, declLoc, name);
@@ -317,8 +318,14 @@ ExpressionPtr buildMethod(DesugarContext dctx, core::LocOffsets loc, core::LocOf
     auto enclosingBlockArg = blockArg2Name(dctx, *blkArg);
 
     DesugarContext dctx2(dctx1.ctx, dctx1.uniqueCounter, enclosingBlockArg, declLoc, name);
-    ExpressionPtr desugaredBody = desugarBody(dctx2, loc, body, std::move(destructures));
-    desugaredBody = validateRBIBody(dctx2, move(desugaredBody));
+
+    ExpressionPtr desugaredBody;
+    if (!removeBody) {
+        desugaredBody = desugarBody(dctx2, loc, body, std::move(destructures));
+        desugaredBody = validateRBIBody(dctx2, move(desugaredBody));
+    } else {
+        desugaredBody = MK::EmptyTree();
+    }
 
     auto mdef = MK::Method(loc, declLoc, name, std::move(args), std::move(desugaredBody));
     cast_tree<MethodDef>(mdef)->flags.isSelfMethod = isSelf;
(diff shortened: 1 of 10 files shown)


aadi-stripe:

→ https://go/builds/bui_MnBftLOgS2EAvO
→ https://go/builds/bui_MnBfPJ9rrk8YDH

aadi-stripe:

→ https://go/builds/bui_MnIw1usljRPUF6
→ https://go/builds/bui_MnIwNrqAcMLJBM

aadi-stripe:

→ https://go/builds/bui_MnJ3oS8tWQXL5s
→ https://go/builds/bui_MnJ4BiDeabqlyf

aadi-stripe:

→ https://go/builds/bui_MpI68YVAOXrh6Q
→ https://go/builds/bui_MpI6vL1R1riKeU

main/cache/cache.cc
```diff
@@ -13,7 +13,16 @@ unique_ptr<OwnedKeyValueStore> maybeCreateKeyValueStore(shared_ptr<::spdlog::log
     if (opts.cacheDir.empty()) {
         return nullptr;
     }
-    auto flavor = opts.lspExperimentalFastPathEnabled ? "experimentalfastpath" : "normalfastpath";
+
+    std::string flavor;
+    if (opts.print.isAutogen() && !opts.print.isAutogenPrintingDepDB()) {
+        flavor = "autogenemptymethodbodies";
+    } else if (opts.lspExperimentalFastPathEnabled) {
```
aadi-stripe, line 20: Use a different cache flavor in empty method mode to avoid poisoning any existing caches.

main/options/options.cc
```diff
@@ -154,6 +154,10 @@ vector<reference_wrapper<PrinterConfig>> Printers::printers() {
     });
 }
 
+bool Printers::isAutogenPrintingDepDB() const {
+    return Autogen.enabled || AutogenMsgPack.enabled;
+}
```
aadi-stripe, line 159: DepDB could be technically string format or msgpack format

ast/desugar/Desugar.cc
```diff
@@ -301,7 +301,8 @@ ExpressionPtr validateRBIBody(DesugarContext dctx, ExpressionPtr body) {
 }
 
 ExpressionPtr buildMethod(DesugarContext dctx, core::LocOffsets loc, core::LocOffsets declLoc, core::NameRef name,
-                          unique_ptr<parser::Node> &argnode, unique_ptr<parser::Node> &body, bool isSelf) {
+                          unique_ptr<parser::Node> &argnode, unique_ptr<parser::Node> &body, bool isSelf,
+                          bool removeBody) {
```
ngroman, line 305: `removeBody` feels misplaced on the method signature for `buildMethod`. Seems like this should be a property of the DesugarContext to me. Defer to @jez since I don't have a ton of context on this step.

aadi-stripe: Mm ok. Technically I can do the `dctx.ctx...` check here (which basically uses `DesugarContext`), instead of in the caller.

main/options/options.cc
```diff
@@ -154,6 +154,10 @@ vector<reference_wrapper<PrinterConfig>> Printers::printers() {
     });
 }
 
+bool Printers::isAutogenPrintingDepDB() const {
```
ngroman, line 157: It's pretty confusing that it's even possible for `isAutogen && !isAutogenPrintingDepDB` to be true

ngroman (changes requested):
(see inline)

main/options/options.cc
aadi-stripe: Yeah, it is. That's just a quirk of how we configure things though -- I suppose I could do a bit of refactoring and better variable naming to make things a bit more clear?

aadi-stripe:
@ngroman Addressed feedback.

ngroman (approved):

aadi-stripe:

→ https://go/builds/bui_MqNL4nyh7IkyoJ
→ https://go/builds/bui_MqNLGGEPi6UPVG

test/cli/autogen-autoloader/example1.rb
```diff
@@ -20,10 +20,6 @@ class Jazz < Quuz
       class JazBaz
         p 'x'
         require 'in_class'
-
-        def honk
-          require 'in_method'
-        end
```
jez, line 26: Why are we deleting code from test files?

(Usually deleting code from tests implies that the change causes tests to fail, and we're papering over it by changing the test instead of fixing the code.)

ast/desugar/Desugar.cc
```diff
@@ -317,8 +317,15 @@ ExpressionPtr buildMethod(DesugarContext dctx, core::LocOffsets loc, core::LocOf
...
-    ExpressionPtr desugaredBody = desugarBody(dctx2, loc, body, std::move(destructures));
-    desugaredBody = validateRBIBody(dctx2, move(desugaredBody));
+
+    bool removeBody = dctx.ctx.state.autogenPrintingSubclassesOrAutoloaderOnly;
+    ExpressionPtr desugaredBody;
+    if (!removeBody) {
+        desugaredBody = desugarBody(dctx2, loc, body, std::move(destructures));
+        desugaredBody = validateRBIBody(dctx2, move(desugaredBody));
+    } else {
+        desugaredBody = MK::EmptyTree();
```
jez, line 327: Can you add a brief comment here to the effect of this being for performance, not correctness?

ast/desugar/Desugar.cc
```diff
@@ -317,8 +317,15 @@ ExpressionPtr buildMethod(DesugarContext dctx, core::LocOffsets loc, core::LocOf
     auto enclosingBlockArg = blockArg2Name(dctx, *blkArg);
 
     DesugarContext dctx2(dctx1.ctx, dctx1.uniqueCounter, enclosingBlockArg, declLoc, name);
-    ExpressionPtr desugaredBody = desugarBody(dctx2, loc, body, std::move(destructures));
-    desugaredBody = validateRBIBody(dctx2, move(desugaredBody));
+
+    bool removeBody = dctx.ctx.state.autogenPrintingSubclassesOrAutoloaderOnly;
+    ExpressionPtr desugaredBody;
+    if (!removeBody) {
```
jez, line 323: Somewhat surprised to see `!removeBody` instead of simply `removeBody` (with swapped branches) given that there are both an `if` and and `else` branch (in my head it reads a bit like a double negation).

But that's not blocking.

main/realmain.cc
```diff
@@ -497,6 +497,9 @@ int realmain(int argc, char *argv[]) {
     if (opts.print.isAutogen()) {
         gs->runningUnderAutogen = true;
     }
+    if (opts.print.isAutogenPrintingSubclassesOrAutoloaderOnly()) {
+        gs->autogenPrintingSubclassesOrAutoloaderOnly = true;
```
jez, line 501: Should we `ENFORCE(gs->runningUnderAutogen)`?

(If not this condition, are there any conditions we should ENFORCE? The change is void of any new `ENFORCE`s so I'm curious.)

jez (approved):
Can you either capture in this PR what the speedup is, or put that information in the stripe-internal bump PR and add a link in this PR?

(Historically, PRs tend to be the longest-lived pieces of technical documentation, especially about point-in-time things like this, and I often find that I want to be able to look back at how much a given change sped something up years into the future, especially when the performance-related change adds a boolean/condition/edge-case that is tempting to remove for the sake of simplicity.)

test/cli/autogen-autoloader/example1.rb
aadi-stripe: I think this was accidental, sorry

main/realmain.cc
aadi-stripe: I think the `ENFORCE` is a reasonable thing to add.

aadi-stripe:
@jez Updated. Will wait to merge until next week, given the holidays.

ast/desugar/Desugar.cc
aadi-stripe: Fixed

aadi-stripe:
> Can you either capture in this PR what the speedup is, or put that information in the stripe-internal bump PR and add a link in this PR?
> 
> (Historically, PRs tend to be the longest-lived pieces of technical documentation, especially about point-in-time things like this, and I often find that I want to be able to look back at how much a given change sped something up years into the future, especially when the performance-related change adds a boolean/condition/edge-case that is tempting to remove for the sake of simplicity.)

Added

Issue #1906, opened 2019-10-04
[meta] address all `disable-fast-path` tests


aisamanra:
Any test with `disable-fast-path` is going to probably cause anomalous behavior in LSP. We've got a number of remaining tests that have the fast path disabled; this is a meta-issue tracking the remaining issues that require us to disable fast path tests:

### Errors arising from not repeating `typeMember` passes in the incremental resolver:
- test/testdata/core/fuzz_bad_subtyping.rb
- test/testdata/resolver/type_members.rb
- test/testdata/resolver/type_member_missing.rb
- test/testdata/resolver/fuzz_type_member_forget.rb
- test/testdata/infer/bad_child_class.rb
- test/testdata/resolver/redefinition_of_subclass_type_member.rb
- test/testdata/resolver/enumerable_strict.rb
- test/testdata/resolver/type_member_missing_then_use.rb
- test/testdata/infer/generics/self_params.rb

### `is not a class or module` not raised in fast path
- test/testdata/namer/nested_static_field.rb
- test/testdata/infer/bad_suggest.rb

### `unable to resolve right-hand side` not raised in fast path
- test/testdata/autogen/generator.rb
- test/testdata/resolver/fuzz_alias_to_type_alias.rb
- test/testdata/resolver/bad_alias.rb

### `redefining constant` not raised in fast path (#1907)
- test/testdata/namer/constant_redefinition/constant_then_class.rb
- test/testdata/namer/constant_redefinition/constant_then_module.rb
- test/testdata/namer/class_and_alias.rb

### wrong expected type in fast path
- test/testdata/resolver/overloads_test.rb

### duplicated `only modules can be included` errors in fast path
- test/testdata/resolver/linearization/includes_class.rb

### infinite types or excessively recursive are weird
- test/testdata/resolver/fuzz_infinite_type.rb
- test/testdata/resolver/infinite.rb
Note: this last one might not be robustly fixable, as it involves different cutoff points that we stop reporting a ridiculous number of errors, but we otherwise report largely the same errors between them.

### type member/alias errors don't get re-reported in the fast-path
- test/testdata/dsl/opus_enum_alias.rb
- test/testdata/infer/generics/bounds.rb
- test/testdata/infer/generics/bounds_super.rb
- test/testdata/resolver/type_member_cycle.rb
- test/testdata/resolver/type_member_fixed_order.rb
- test/testdata/resolver/type_member_bad_parent.rb

PR #6570, opened 2022-11-17
Make fast path fast again by regressing type member fixes

jez:
### Motivation

After the changes in #6422, we're seeing the p50 fast path time move from ~50ms
to ~225ms and the p99 fast path time move from ~300ms to ~600ms. And the change
won't be fully rolled out for at least another few business days (as people
update their merge bases).

It seems like the cause for this almost entirely spent in `incremental_resolve`.
The time spent in `incremental_naming` is basically unchanged vs before.

That's odd, because #6422 basically only changed namer, and left resolver
untouched. Except that after those changes, we increase the number of files on
the fast path. The p50 `num_files` on the fast path moves from 1 to 2 (so far)
and the p99 `num_files` moves from <6 to ~10.

### Summary

The contents of this change regress how we handle `type_member`'s on the fast
path. That's not great, and it will definitely cause us to miss reporting some
errors on the fast path that we should, but at least it won't slow down
literally every completion request.

Reopens #1906.

### Test plan

Existing tests for correctness

Existing metrics for seeing performance changes in production.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -217,8 +217,10 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
     config->logger->debug("Added {} files that were not part of the edit to the update set", result.extraFiles.size());
     UnorderedMap<core::FileRef, core::FoundDefHashes> oldFoundHashesForFiles;
     auto toTypecheck = move(result.extraFiles);
+    auto shouldRunIncrementalNamer =
+        config->opts.lspExperimentalFastPathEnabled && !result.changedSymbolNameHashes.empty();
     for (auto [fref, idx] : result.changedFiles) {
-        if (config->opts.lspExperimentalFastPathEnabled && !result.changedSymbolNameHashes.empty()) {
+        if (shouldRunIncrementalNamer) {
             // Only set oldFoundHashesForFiles if we're processing a real edit.
             //
             // This means that no-op edits (and thus calls to LSPTypechecker::retypecheck) don't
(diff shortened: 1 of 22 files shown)


main/lsp/LSPTypechecker.cc
```diff
@@ -245,19 +247,23 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
+        if (shouldRunIncrementalNamer) {
+            // TODO(jez) Using `gs` to access package information here assumes that edits to
+            // __package.rb files don't take the fast path. We'll want (or maybe need) to revisit this
+            // when we start making edits to `__package.rb` take fast paths.
+            if (!(fref.data(*gs).isPackage())) {
+                auto pkgName = gs->packageDB().getPackageNameForFile(fref);
+                if (pkgName.exists()) {
+                    // Since even no-op (e.g. whitespace-only) edits will cause constants to be deleted
+                    // and re-added, we have to add the __package.rb files to set of files to retypecheck
+                    // so that we can re-run PropagateVisibility to set export bits for any contants.
```
froydnj, line 261:
```suggestion
                    // so that we can re-run PropagateVisibility to set export bits for any constants.
```

froydnj (approved):
I guess the p99 changes sort of make sense: we're doubling the number of files being checked, so that should ~double the amount of time we take, but the p50 changes seem kind of weird, because we're quadrupling the time in response to doubling the number of files.

Wonder where that time is going.

jez:
> because we're quadrupling the time in response to doubling the number of files

I'm guessing that in general, there are **lots** of constants in `__package.rb` files, and if we think that the "resolve constants" step is the slowest, that could explain it (double the files, but more than double the number of constant lits)

PR #6571, opened 2022-11-18
Ensure that print_document_symbols is built in CI

jez:
### Motivation

It looks like we shipped a change in #6531 that introduced a compile-time error
in print_document_symbols.cc, but no one noticed until running the
`update_testdata_exp.sh` script.

### Test plan

- Build fails before fixing document symbols test:
  - https://buildkite.com/sorbet/sorbet/builds/25460#01848c4d-2d79-446f-ad65-4bf8cd43b90e/209-252
- Build passes after the fix

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -114,6 +114,14 @@ cc_binary(
     ],
 )
 
+sh_test(
+    name = "print_document_symbols_builds",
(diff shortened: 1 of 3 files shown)


froydnj (approved):
🙏

PR #6574, opened 2022-11-22
Better ergonomics for `--web-trace-file` and LSP

jez:
### Motivation

Right now it's a bit annoying to debug performance of LSP using
`--web-trace-file` because you have to wait 5 minutes and then make a
task.

### Test plan

Manually testing on Stripe's codebase


jez:

→ https://go/builds/bui_MqhYoeypbueAnv
→ https://go/builds/bui_MqhYyZRnhjf19z

PR #6576, opened 2022-11-22
Skip `finalizeSymbols` in incrementalResolve

jez:
### Motivation

`finalizeSymbols` does a global scan over the symbol table. This
consistently takes 200ms at p50, and after #6422 we were doing it on
every edit.

That's too slow to run on every edit, but luckily it's only needed in
cases where we had to run incremental namer.

As a (temporary?) fix, we can skip doing that work unless required. This
only works because of the changes landed in #6570, which skip running
incremental namer unless required.

Ideally, this is only temporary because it means that it's **still** holding us
back from being able to quickly service edits that **must** run
incremental namer. I haven't looked carefully yet, but there might be a
way we can make a `finalizeSymbolsIncremental` function somehow.

As such, this will only affect the p50 (and maybe the p90 or so).
Luckily changes to definitions are much more rare than changes inside a
method body (so basically: this will improve completion times, but not
necessarily improve diagnostic times for definition edits). That's a
better spot than we were in this time last week though, and we can keep
figuring out ways to do less work in a bit.

### Test plan

See included automated tests.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -232,13 +232,14 @@ incrementalResolve(core::GlobalState &gs, vector<ast::ParsedFile> what,
             what = packager::Packager::runIncremental(gs, move(what));
         }
 #endif
+        auto runIncrementalNamer = foundHashesForFiles.has_value() && !foundHashesForFiles->empty();
         {
             Timer timeit(gs.tracer(), "incremental_naming");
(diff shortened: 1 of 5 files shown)


jez:
Self-assigning to fix tests.

jez:

→ https://go/builds/bui_Mtm52sSHBQLc8I
→ https://go/builds/bui_Mtm5THDE3RcuWR

PR #6577, opened 2022-11-22
Add patch to expose need_major_by via GC.latest_gc_info

mirek26:
### Motivation

Adding two patches that enable us to optimize GC and get efficiency savings

1. https://github.com/ruby/ruby/pull/6791 - open, but I hope I'll be able to get some version of this in. This will enable us to get rid of unnecessary minor GCs that we currently run to prevent major GCs during request processing.

    > We trigger GC out-of-band to reduce latency impact on synchronous request processing. It’s not perfect, but it works. One thing we really want to avoid is major GC during request processing - so we don’t want to start request processing in a state where need_major_gc is set. If there is a risk of major GC, we prefer to run it out-of-band, where it doesn't matter for request latency.
    > 
    > Unfortunately, this state is currently not exposed - so we can’t know whether the next GC is set to be major or not. As a workaround, we trigger minor GC twice in a row - if a major GC threshold is hit during the first GC, the second GC is upgraded to major and there is no immediate risk of another major GC. This mostly works, but it's not great and it's not efficient (the second GC is not cheap and it's unnecessary in most cases); this PR makes it possible to do that in a better way by exposing need_major_gc.
    > 
    > It may seem strange to expose it under GC.latest_gc_info but I think it’s a reasonable place because the decision was made during the latest GC.

2. https://github.com/ruby/ruby/pull/4860 - merged long time ago, included in Ruby 3.1. This will remove some unnecessary major GCs due to incorrect calculations

### Test plan

See included automated tests.

### DIFF
--- a/third_party/ruby/gc-add-need-major-by.patch
+++ b/third_party/ruby/gc-add-need-major-by.patch
@@ -1,16 +1,7 @@
-From 5484a6da6fb63ec2be679e0d1aadc01aaf4ed5ff Mon Sep 17 00:00:00 2001
-From: Mirek Klimos <mirek@stripe.com>
-Date: Tue, 22 Nov 2022 11:41:19 -0800
-Subject: [PATCH] [GC] Expose need_major_gc via GC.latest_gc_info
-
----
- gc.c | 19 +++++++++++++++++--
- 1 file changed, 17 insertions(+), 2 deletions(-)
-
-diff --git a/gc.c b/gc.c
+diff --git gc.c gc.c
 index 67a709ff79..2825466788 100644
---- a/gc.c
-+++ b/gc.c
+--- gc.c
++++ gc.c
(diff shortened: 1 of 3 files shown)


mirek26:

→ https://go/builds/bui_MqmZ0qh4NeIR12
→ https://go/builds/bui_MqmZYG8vnMnKbU

mirek26:

→ https://go/builds/bui_MuCByw0MMuLL3H
→ https://go/builds/bui_MuCBui1hS4idPO

mirek26:

→ https://go/builds/bui_MuCrCVOCJic9ZX
→ https://go/builds/bui_MuCrK4ZwEecEoT

froydnj (approved):
The patch LGTM.

@mirek26 Is this mergeable in its present state?

mirek26:
@froydnj Yup, should be ready to go!

PR #3375, opened 2020-08-19
Fix signature for Enumerable#detect

Morriar:
### Motivation

Fixes https://github.com/sorbet/sorbet/issues/3240.

### Test plan

See included automated tests.

PR #6578, opened 2022-11-23
Match signature of Enumerable#find to detect

nickpwhite:
### Motivation

https://github.com/sorbet/sorbet/pull/3375 updated the signatures for `Enumerable#detect` but not `Enumerable#find` which is an alias. The existing signature for `find` means that something like [this](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Breturns%28String%29%7D%0Adef%20foo%0A%20%20%5B%22a%22%2C%20%22few%22%2C%20%22words%22%5D.find%28-%3E%20%7B%22default%22%7D%29%20%7B%7Cword%7C%20word.length%20%3E%203%7D%0Aend) won't typecheck:
```ruby
sig {returns(String)}
def foo
  ["a", "few", "words"].find(-> {"default"}) {|word| word.length > 3}
  # Expected String but found T.nilable(String) for method result type https://srb.help
end
```

Whereas [this](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Breturns%28String%29%7D%0Adef%20foo%0A%20%20%5B%22a%22%2C%20%22few%22%2C%20%22words%22%5D.detect%28-%3E%20%7B%22default%22%7D%29%20%7B%7Cword%7C%20word.length%20%3E%203%7D%0Aend) shows that the equivalent code would typecheck with the right signature:
```ruby
sig {returns(String)}
def foo
  ["a", "few", "words"].detect(-> {"default"}) {|word| word.length > 3}
end
```

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/enumerable.rbi
+++ b/rbi/core/enumerable.rbi
@@ -1818,17 +1818,28 @@ module Enumerable
   # (1..100).find          { |i| i % 5 == 0 && i % 7 == 0 }   #=> 35
   # ```
   sig do
-    params(
-        ifnone: Proc,
+    type_parameters(:U)
+      .params(
+        ifnone: T.proc.returns(T.type_parameter(:U)),
         blk: T.proc.params(arg0: Elem).returns(BasicObject),
+      )
+      .returns(T.any(T.type_parameter(:U), Elem))
+  end
+  sig do
+    type_parameters(:U)
+      .params(
+        ifnone: T.proc.returns(T.type_parameter(:U)),
+      )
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_Mr61N9otqZUDLq
→ https://go/builds/bui_Mr611ZPVZYy4nJ

jez (approved):

PR #6579, opened 2022-11-23
Add Timers to finalizeSymbols

jez:
### Motivation

finalizeSymbols was recently implicated in some perf regressions on the
fast path. (See #6576 for a discussion.) Fixing it will likely involve
making part or all of `finalizeSymbols` incremental in some way (or
maybe just faster across the board).

Separately, we don't have timers showing which phases of finalizeSymbols
take the longest, so I'd like to add some.

The `computeLinearization` method already has a timer inside it.

For posterity, finalizeSymbols takes 240ms at p50. The breakdown is
currently something like 50ms for computeLinearization, 170ms for
resolve_type_members and something like 20ms for mixes_in_class_methods.

### Test plan

Tested with a local Stripe-internal build and dumped a web trace file to see
that the spans showed up.

Separately, it looks like our web trace file generation doesn't do a great job
at showing spans when two timers start at the same time (nested inside on
another). I think likely what's happening is that the trace viewer has to
attempt to infer parent trace information and fails to infer that. We could
probably fix that but I think it would involve a refactor of the `Timer` class,
which sounds invasive. (Though _maybe_ we could figure out how to do it by simply
changing the web-trace-file generation, to attempt to figure out when spans
start at the same time and one is longer than the other?)

### DIFF
--- a/resolver/GlobalPass.cc
+++ b/resolver/GlobalPass.cc
@@ -285,39 +285,43 @@ void Resolver::finalizeSymbols(core::GlobalState &gs) {
     // that resolves types and we don't want to introduce additional passes if
     // we don't have to. It would be a tractable refactor to merge it
     // `ResolveConstantsWalk` if it becomes necessary to process earlier.
-    for (uint32_t i = 1; i < gs.classAndModulesUsed(); ++i) {
-        auto sym = core::ClassOrModuleRef(gs, i);
+    {
+        Timer timer(gs.tracer(), "resolver.mix_in_class_methods");
 
-        if (sym.data(gs)->flags.isLinearizationComputed) {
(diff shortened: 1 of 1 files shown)


froydnj (approved):
Just reading through the code at various points in the past, it seems like type member resolution ought to be faster, but maybe due to e.g. attached class implemented as a type member (?), I misunderstand the amount of work to do in that pass.

jez:
> Just reading through the code at various points in the past, it seems like type member resolution ought to be faster, but maybe due to e.g. attached class implemented as a type member (?), I misunderstand the amount of work to do in that pass.

Interesting, I wonder if we can special case it.

PR #6582, opened 2022-11-28
Get rid of some untyped methods in `Date`

aisamanra:
### Motivation
We have a lot of `T.untyped` in the RBI for `Date`, so this starts to cut down on it.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/date.rbi
+++ b/rbi/stdlib/date.rbi
@@ -933,8 +933,8 @@ class Date
   # ```ruby
   # Date.today   #=> #<Date: 2011-06-11 ...>
   # ```
-  sig {params(arg0: T.untyped).returns(Date)}
-  def self.today(*arg0); end
+  sig {params(start: T.any(Integer, Float)).returns(Date)}
+  def self.today(start=Date::ITALY); end
 
   # Parses the given representation of date and time, and creates a date object.
   # This method does not function as a validator.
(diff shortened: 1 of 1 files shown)


aisamanra:

→ https://go/builds/bui_MsyPanr58sxSzD
→ https://go/builds/bui_MsyPiAfX4VkQBU

aisamanra:
I've got the changes ready internally for the bump, c.f. go/pr/548835.

rbi/stdlib/date.rbi
```diff
@@ -947,8 +947,8 @@ class Date
   # Date.parse('20010203')            #=> #<Date: 2001-02-03 ...>
   # Date.parse('3rd Feb 2001')        #=> #<Date: 2001-02-03 ...>
   # ```
-  sig {params(arg0: T.untyped).returns(T.attached_class)}
-  def self.parse(*arg0); end
+  sig {params(string: String, comp: T::Boolean, state: T.any(Integer, Float)).returns(T.attached_class)}
+  def self.parse(string='-4712-01-01', comp=true, state=Date::ITALY); end
```
aisamanra, line 951: One specific thing I'm not sure about is whether to type `comp` as `T::Boolean` here. The _intention_ is that it's a boolean—it defaults to `true` and can be `false`—but Ruby being Ruby means that technically any value is fine here as they'll auto-coerce to booleans. I'd be willing to relax this to `T.untyped` if we want, but I have it as `T::Boolean` to surface places where incorrect assumptions are being made about the code.

jez (approved):

Issue #6587, opened 2022-12-02
(Thread::)Queue.initialize optional argument


dduugg:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AQueue.new%28%5B1%5D%29%0A)

```ruby
Queue.new([1])
```

#### Observed output

```
editor.rb:3: Wrong number of arguments for constructor. Expected: 0, got: 1 https://srb.help/7004
     3 |Queue.new([1])
                  ^^^
  Autocorrect: Use `-a` to autocorrect
    editor.rb:3: Delete
     3 |Queue.new([1])
                  ^^^
Errors: 1
```

#### Expected behavior

As of Ruby 3.1, `Queue` can be passed an enumerable in the constructor:

https://rubyreferences.github.io/rubychanges/3.1.html#threadqueueinitialize-initial-values-can-be-passed-to-initializer
https://ruby-doc.org/core-3.1.0/Thread/Queue.html#method-c-new

PR #6588, opened 2022-12-02
Update Thread::Queue constructor sig

dduugg:
### Motivation
Resolves https://github.com/sorbet/sorbet/issues/6587

I used this CSV::Row constructor code as inspiration (it appears i can write a sig that doesn't include every parameter): https://github.com/sorbet/sorbet/blob/master/rbi/stdlib/csv.rbi#L966-L968

### Test plan

Are there rbi tests? I expected to find `test/rbi` but no dice.

### DIFF
--- a/rbi/core/thread.rbi
+++ b/rbi/core/thread.rbi
@@ -1479,6 +1479,12 @@ end
 # consumer.join
 # ```
 class Thread::Queue < Object
+  # Creates a new queue instance, optionally using the contents of an enumerable for its initial state.
+  # https://ruby-doc.org/core-3.1.0/Thread/Queue.html#method-c-new
+  sig {void}
+  sig {params(enumerable: Enumerable).void}
+  def initialize(enumerable); end
+
   # Alias for:
   # [`push`](https://docs.ruby-lang.org/en/2.7.0/Queue.html#method-i-push)
   sig {params(obj: T.untyped).returns(T.untyped)}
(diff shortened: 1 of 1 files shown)


rbi/core/thread.rbi
```diff
@@ -1479,6 +1479,12 @@ end
 # consumer.join
 # ```
 class Thread::Queue < Object
+  # Creates a new queue instance, optionally using the contents of an enumerable for its initial state.
+  # https://ruby-doc.org/core-3.1.0/Thread/Queue.html#method-c-new
+  sig {void}
+  sig {params(enumerable: Enumerable).void}
+  def initialize(enumerable); end
```
jez, line 1486: Rather than using an overloaded signature, can you define `enumerable` as `enumerable=T.unsafe(nil)` and only have one `sig`?

Overloaded sigs are messy and we try to avoid using them unless required.

dduugg: ✅ PTAL

jez:

→ https://go/builds/bui_MuTjhSEwGAr01C
→ https://go/builds/bui_MuTjec8gbV7FyW

jez:
Your RBI change introduces a type error while typechecking the RBIs themselves.

dduugg:
@jez I'm having some bazel issues that prevent me from confirming locally, but i believe I've resolved the test failures, PTAL

jez (approved):

jez:

→ https://go/builds/bui_MuVabDfxcAiT3X
→ https://go/builds/bui_MuVaehGZaIsT0U

jez (approved):

PR #6589, opened 2022-12-02
Specify return type of Class.new using RBI sigs

egiurleo:
Our goal with this PR is for Sorbet to automatically know the type of `Class.new(Parent)`. In #6162, we originally tried to accomplish in the rewriter, but @jez pointed out that `T.cast` is not allowed in all of the locations that a rewrite could occur. We've attempted a different approach using only RBI sigs to properly type the return of this method.

### Motivation
Right now, the result of `Class.new` is always typed as `Class`:

```ruby
x = Class.new(Parent)
T.reveal_type(x) # => Class
```

This forces users to cast the results whenever they use `Class.new`.

By using generics to specify the correct type in the RBI signature, we can ensure that the correct type is returned by default.

```ruby
x = Class.new(Parent)
T.reveal_type(x) # => T.class_of(Parent)
```

### Test plan
See included automated tests.

cc: @Morriar


jez:

→ https://go/builds/bui_MuUcNWhuT5Syc7
→ https://go/builds/bui_MuUc3hqow63m3m

jez (approved):
This is a beautiful change.

Simple, well-tested, and it removes 14 now-redundant casts from Stripe's codebase.

jez:
You have one CFG snapshot test that fails because one of the inferred intermediate types becomes more specific. You can use `tools/scripts/update_testdata_exp.sh` to fix the failure.

egiurleo:
@jez I ran that command and pushed the results, so hopefully things should be passing now.

test/testdata/autogen/generator.rb.autogen.exp
```diff
@@ -36,33 +36,33 @@ requires: []
...
  resolved=[]
  loc=test/testdata/autogen/generator.rb:6
  is_defining_ref=0
-[ref id=5]
- scope=[Gener]
- name=[HydraResource]
- nesting=[[Gener]]
- resolved=[Gener HydraResource]
- loc=test/testdata/autogen/generator.rb:10
- is_defining_ref=0
```
jez, line 65: Ah, sorry, I forgot about this. There's a bug in that script that generates spurious edits to this test. Can you revert the changes in this file?

I've been meaning to fix this for a while and in the mean time I've just been suffering through the bad edits by manually fixing things.

egiurleo: Done!

PR #6599, opened 2022-12-08
Assume that `X = A.new` has type `A` in `# typed: true` files

jez:
### Motivation

"Why do I have to write an annotation for this" is a frequent question that new Sorbet users have. A lot of the time, the answer is for performance—Sorbet would rather not have to do a lot of expensive work to figure out what type to assume, and would rather the user just be explicit about what type the thing has.

However, in some cases this leads to a lot of redundancy.

In this change we take advantage of the fact that `A.new` **almost always** has type `A`, and use that to make an assumption. Sorbet will then check if that assumption is wrong and ask the user to correct it if necessary.

This puts a slightly larger burden on the user—if they are doing something where the result of calling `new` on a constant does not have type `A`, then they will get an error and be **forced** to add a type annotation (forced because I've chosen to make this behavior happen in `# typed: true` files too. This change would not make the user write any more type annotations than they otherwise would have had to had this been something that only fired in `# typed: strict` files). In any case, this error did not occur even once in Stripe's codebase so I think it should be quite rare.

### Test plan

See included automated tests.

### TODO

- [x] I have to figure out how to handle the "recv is a constant but not a class" error, because that currently leads to a terrible user experience.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1543,9 +1543,22 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                             e.addErrorNote("You may need to add additional `{}` annotations", "sig");
                         }
                     } else if (!core::Types::isSubType(ctx, ty.type, castType)) {
-                        if (auto e = ctx.beginError(bind.loc, core::errors::Infer::CastTypeMismatch)) {
-                            e.setHeader("Argument does not have asserted type `{}`", castType.show(ctx));
-                            e.addErrorSection(ty.explainGot(ctx, ownerLoc));
+                        if (c.cast == core::Names::assumeType()) {
+                            if (auto e = ctx.beginError(bind.loc, core::errors::Infer::CastTypeMismatch)) {
+                                e.setHeader("Assumed expression had type `{}` but found `{}`", castType.show(ctx),
+                                            ty.type.show(ctx));
+                                e.addErrorSection(ty.explainGot(ctx, ownerLoc));
+                                e.addErrorNote("Please add an explicit type annotation to correct this assumption");
+                                if (bind.loc.exists() && c.valueLoc.exists()) {
+                                    e.replaceWith("Add explicit annotation", ctx.locAt(bind.loc), "T.let({}, {})",
+                                                  ctx.locAt(c.valueLoc).source(ctx).value(), ty.type.show(ctx));
+                                }
+                            }
+                        } else {
+                            if (auto e = ctx.beginError(bind.loc, core::errors::Infer::CastTypeMismatch)) {
+                                e.setHeader("Argument does not have asserted type `{}`", castType.show(ctx));
+                                e.addErrorSection(ty.explainGot(ctx, ownerLoc));
+                            }
                         }
                     }
                 } else if (!bind.value.isSynthetic()) {
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1537,7 +1537,7 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
 
                 if (c.cast != core::Names::cast()) {
                     if (c.cast == core::Names::assertType() && ty.type.isUntyped()) {
-                        if (auto e = ctx.beginError(bind.loc, core::errors::Infer::CastTypeMismatch)) {
+                        if (auto e = ctx.beginError(bind.loc, core::errors::Infer::IncorrectlyAssumedType)) {
                             e.setHeader("Expected a type but found `{}` for `{}`", "T.untyped", "T.assert_type!");
                             e.addErrorSection(ty.explainGot(ctx, ownerLoc));
                             e.addErrorNote("You may need to add additional `{}` annotations", "sig");
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1537,14 +1537,14 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
 
                 if (c.cast != core::Names::cast()) {
                     if (c.cast == core::Names::assertType() && ty.type.isUntyped()) {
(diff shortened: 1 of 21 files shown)


jez:

→ https://go/builds/bui_MwRmZXzQtKy9SS
→ https://go/builds/bui_MwRmasWN5NbpV9

jez:
> Pretty sure Stripe's codebase has instances of `A.new` overrides that don't actually return an instance of `A`. 😢 (I ran into these cases when optimizing `new` in the compiler.)

The idea is that if we limit this to `# typed: true` or even `# typed: strict` files, either (1) the typechecker tells us all the violations and we go fix them with a codemod before landing, or (2) we limit ourselves to `typed: strict` files, which must necessarily already have `T.let`, preventing the rewriter from firing.

jez:
Update: it caused 35 failure when run at `# typed: true`. Going to try again at `typed: strict`.

jez:
> Update: it caused 35 failure when run at `# typed: true`. Going to try again at `typed: strict`.

I misspoke, it actually only caused 21 errors at typed: true, and all of them were legitimate type errors.

I could go either way (do this at typed true or at typed strict). In any case, we can't land until we get a better error message that communicates what's going on.

infer/environment.cc
```diff
@@ -1543,9 +1543,22 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
...
+                        if (c.cast == core::Names::assumeType()) {
+                            if (auto e = ctx.beginError(bind.loc, core::errors::Infer::IncorrectlyAssumedType)) {
+                                e.setHeader("Assumed expression had type `{}` but found `{}`", castType.show(ctx),
+                                            ty.type.show(ctx));
+                                e.addErrorSection(ty.explainGot(ctx, ownerLoc));
+                                e.addErrorNote("Please add an explicit type annotation to correct this assumption");
+                                if (bind.loc.exists() && c.valueLoc.exists()) {
+                                    e.replaceWith("Add explicit annotation", ctx.locAt(bind.loc), "T.let({}, {})",
+                                                  ctx.locAt(c.valueLoc).source(ctx).value(), ty.type.show(ctx));
+                                }
```
froydnj, line 1555: I guess this offers a way forward to (maybe) easily dealing with #4837?  (Some of the issues from the last run in that PR -- which was admittedly a long time ago -- look not really autocorrectable, though.)

website/docs/error-reference.md
```diff
@@ -3897,6 +3897,27 @@ arr = T::Array[NilClass].new
...
+When one of those assumptions is broken, it reports this error. In every case,
+the fix is to provide the correct, explicit annotation so that Sorbet does not
+have to assume anything.
+
+(Sorbet works this way in general for performance—Sorbet can be much faster if
+it doesn't have to go back and re-do a lot of work it has already done when it
+encounters a broken assumption. There is more information on this in
+[Why does Sorbet sometimes need type annotations?](why-type-annotations.md). A
+few type annotations can go a long way to ensuring that Sorbet can type check
+large codebases quickly.)
```
froydnj, line 3919: This paragraph strikes me as kind of weird given the preceding paragraphs explaining that Sorbet just tried to assume something and it turned out it was mistaken in that assumption.

e.g. what is "this way" in the first sentence referring to?  (It reads to me like it's referencing "...so that Sorbet doesn't have to assume anything"...but that referencing doesn't make sense in the context of this error message because clearly Sorbet is already making assumptions.)  Or what is "it doesn't have to go back and re-do a lot of work it has already done when it encounters a broken assumption" trying to get at?  I guess both of these are referencing the design principle that Sorbet does everything in a single pass, no backtracking or fixed points required, but I'm not sure how useful that is as user documentation.

Maybe this whole paragraph would just be better as a quick sentence pointing people at `why-type-annotations.md`?  But that also seems strange, since `why-type-annotations.md` is explaining, well, why type annotations are necessary and this error is coming from a place where type annotations weren't explicitly provided, but should be.

resolver/resolver.cc
```diff
@@ -2802,10 +2803,24 @@ class ResolveTypeMembersAndFieldsWalk {
...
+            // make sure that it's even possible to be valid.
+            auto *cnst = ast::cast_tree<ast::ConstantLit>(cast->typeExpr);
+            ENFORCE(cnst != nullptr, "Rewriter should always use const for typeExpr, which should now be resolved");
+            if (!cnst->symbol.isClassOrModule()) {
+                // The rewriter was over-eager in attempting to infer type `A` for `A.new` because
+                // `A` was not a class. Get rid of the cast, replace it with the original arg.
+                tree = move(cast->arg);
+                return;
+            }
+        }
```
froydnj, line 2818: This is such a nice way of dealing with things.

rewriter/ConstantAssumeType.cc
```diff
@@ -0,0 +1,40 @@
...
+    if (send == nullptr) {
+        return;
+    }
+
+    if (send->fun != core::Names::new_()) {
+        return;
+    }
+
+    if (!(ast::isa_tree<ast::UnresolvedConstantLit>(send->recv) || ast::isa_tree<ast::ConstantLit>(send->recv))) {
+        // TODO(jez) No real reason to preclude ConstantLit here except laziness
```
froydnj, line 32: Unnecessary TODO now?

froydnj (approved):
I think this whole PR is a net win for users.  I'm not sure about the explanation in the docs.

infer/environment.cc
jez: That PR specifically is just asking "can we do the thing we already do with literals assigned to constants for instance variables as well" but yes we could then take it one step further and also do this transformation (i might take a look at this depending on how much time i have in the next few days)

website/docs/error-reference.md
jez: I took a stab at rewording this. I'm going to set the PR to auto-merge and we can rewrite this section of the docs in a future PR if you still think the follow up doesn't quite eliminate the weirdness you think it started out with.

PR #6606, opened 2022-12-08
Fix bad packager did you mean error

jez:
### Motivation

An alternative approach would have been to detect the package problem
specifically. That's something we could do, but it's a bit tricky to get that
working, and I thought this approach was simple.

This has the maybe bad side effect of silencing some did you mean errors that
would expose other "bugs" in did you mean suggestion, but did you mean
suggestion is best effort anyways.

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/packager_did_you_mean/test.out
+++ b/test/cli/packager_did_you_mean/test.out
@@ -1,0 +1,11 @@
+__package.rb:6: Unable to resolve constant `ProjectWithLongDiscernableName` https://srb.help/5002
+     6 |  export ProjectWithLongDiscernableName::Foo::Example
+                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Did you mean `ProjectWithLongDiscernableName`? Use `-a` to autocorrect
+    __package.rb:6: Replace with `ProjectWithLongDiscernableName`
+     6 |  export ProjectWithLongDiscernableName::Foo::Example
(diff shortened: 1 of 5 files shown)


test/cli/packager_did_you_mean/test.out
```diff
@@ -0,0 +1,4 @@
+__package.rb:6: Unable to resolve constant `ProjectWithLongDiscernableName` https://srb.help/5002
+     6 |  export ProjectWithLongDiscernableName::Foo::Example
+                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
froydnj, line 3: Is this the case that you mentioned in your PR description?

froydnj (approved):

test/cli/packager_did_you_mean/test.out
jez: Yeah so like we could try to say "oh it looks like the suggested symbol is `<PackageSpecRegistry>::ProjectWithLongDiscernableName` and the current symbol is `<root>` + `ProjectWithLongDiscernableName`, so we should try to see if the suggested symbol is "what the user wrote minus `<PackageSpecRegistry>::` but the way to do that while operating on SymbolRefs and NameRefs (instead of string source equality) is just an annoying loop that has a lot of code for what can be done with string comparison in 2 lines (and maybe even be just as fast, idk i didn't test it)

PR #1252, opened 2019-07-10
Add a dsl pass for class attribute APIs from Rails

XrXr:
This includes `{c,m}attr_{accessor,reader,writer}` and
`class_attribute`. These are from Rails and show up once in a while.
These are pretty much super charged versions of attr_accessor and friends.

Implementation wise, it's mostly straight forward, only a little cumbersome since the `cattr_*` and `mattr_*` are aliases, so there are a lot of test cases. The `DSL_SYNTHESIZED` change was necessary to stop Sorbet from complaining about the use of these in files that don't have sigils.

### Motivation
These are APIs have been in Rails for a long time and every app has some usages. It makes sense to me to have first-class support for them, even though they don't show up as frequently as say, `delegate`. 

### Test plan

See included automated tests.

PR #6608, opened 2022-12-09
Include thread_mattr_accessor in Mattr dsl

composerinteralia:
[`thread_mattr_accessor`][thread_mattr_accessor] (and it's alias `thread_cattr_accessor`) defines the same set of methods as `mattr_accessor` and `cattr_accessor` (support for those was added in https://github.com/sorbet/sorbet/pull/1252). 

[thread_mattr_accessor]: https://api.rubyonrails.org/classes/Module.html#method-i-thread_mattr_accessor

This commit adds support for these additional methods. They might not be used as often as `mattr_accessor` and friends, but I thought it might still be worth adding since it's such a small change.

### Motivation

I was recently confused about why one of these methods worked out of the box and the other gave me an error. Since they are such similar methods, I assumed they would either both work or both error.

### Test plan

See included automated tests, which follow the same pattern as the tests from https://github.com/sorbet/sorbet/pull/1252

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -251,6 +251,8 @@ NameDef names[] = {
     {"cattrAccessor", "cattr_accessor"},
     {"cattrReader", "cattr_reader"},
     {"cattrWriter", "cattr_writer"},
+    {"threadMattrAccessor", "thread_mattr_accessor"},
+    {"threadCattrAccessor", "thread_cattr_accessor"},
     {"instanceReader", "instance_reader"},
     {"instanceWriter", "instance_writer"},
     {"instanceAccessor", "instance_accessor"},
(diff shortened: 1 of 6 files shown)


composerinteralia:
> Please ask a contributor to unblock the tests 

🙏🏻

composerinteralia:
Oops, I had a lint error. I ran tools/scripts/format_cxx.sh to fix and then amended the commit, but I think I'll need a contributor to run the tests for me again. Thank you!

Contributing to this project has been super smooth. I'm really impressed. Well done all!

jez (approved):

jez:
Thanks!

> Contributing to this project has been super smooth. I'm really impressed. Well done all!

glad to hear 🙂

PR #6611, opened 2022-12-09
Fix `X = A.new` change for generic classes and modules

jez:
### Motivation

In #6599, we attempted to infer the type of `X = A.new`.

Our simplistic approach breaks down when `A` is a generic, and we would leak
that to the user.

We should abort attempting to infer the type for generic classes, and require an
explicit annotation at `# typed: strict` for these cases.

### Test plan

See included automated tests.


froydnj (approved):
Doh!

PR #6617, opened 2022-12-12
Correct syntax for example in documentation describing how to work around `T::Struct` inheritance limitations.

lerebear:
This corrects an example given in documentation for how to work around `T::Struct` inheritance limitations. It should have no impact to the library itself.

### Motivation

I recently copied this example to modify it in a personal project, but found that the code was incorrect. This was only a minor inconvenience, but it seemed to me to be worth correct.

### Test plan

This change only affects documentation, and so should not need tests.

### DIFF
--- a/website/docs/tstruct.md
+++ b/website/docs/tstruct.md
@@ -197,6 +197,7 @@ child class:
 ```ruby
 module Common
   extend T::Helpers
+  extend T::Sig
   interface!
   sig {abstract.returns(Integer)}
   def foo; end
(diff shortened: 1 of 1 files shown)


lerebear:
@froydnj, given that you've been assigned for review, might you be able to unblock the tests on this PR for me also?

website/docs/tstruct.md
```diff
@@ -198,9 +198,9 @@ child class:
 module Common
   extend T::Helpers
```
froydnj, line 199:
```suggestion
  extend T::Helpers
  extend T::Sig
```

froydnj (approved):
Thanks!

Issue #2174, opened 2019-11-14
ENVClass rbi is currently missing many methods on ENV


andrew-stripe:
#### Input
[→ View on sorbet.run](https://sorbet.run/#ENV.replace(%7B%7D)%0AENV.assoc(%22HOME%22))

```ruby
ENV.replace({})
ENV.assoc("HOME")
```

#### Observed output

```
editor.rb:1: Method replace does not exist on Sorbet::Private::Static::ENVClass https://srb.help/7003
     1 |ENV.replace({})
        ^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    editor.rb:1: Replace with reduce
     1 |ENV.replace({})
            ^^^^^^^


editor.rb:2: Method assoc does not exist on Sorbet::Private::Static::ENVClass https://srb.help/7003
     2 |ENV.assoc("HOME")
        ^^^^^^^^^^^^^^^^^
Errors: 2
```

#### Expected behavior
Both of these methods should be defined. There are also several other missing methods on ENV but I didn't work through manually diffing the whole method list since I was hoping there's an automated way to regenerate this.

RBI for reference: https://github.com/sorbet/sorbet/blob/bec5b685d1a806711cd3e0a88b9dfda286700968/rbi/sorbet/sorbet.rbi#L68-L215

I'm also not sure `T.nilable[String]` is the right type for the element value:
https://github.com/sorbet/sorbet/blob/bec5b685d1a806711cd3e0a88b9dfda286700968/rbi/sorbet/sorbet.rbi#L71

In Ruby 2.6.5:
```
ENV["abc"] = nil
ENV.to_h.key?("abc") # false
```

jez:
https://sorbet.org/docs/faq#it-looks-like-sorbet-s-types-for-the-stdlib-are-wrong

ENVClass support in Sorbet is a hack, because it's IMO a hack in Ruby too. Anything we do in Sorbet is likely to be an approximation, so we welcome approximate solutions here.

PR #6620, opened 2022-12-13
Add a sig for slice of Sorbet::Private::Static::ENVClass

wteuber:
Refs #2174

### Motivation
When type checking this code:
```ruby
ENV.slice("USER_ID", "USER_NAME")
```
Sorbet complained with:
```
Method slice does not exist on Sorbet::Private::Static::ENVClass
```
while actually [slice is present on ENV](https://ruby-doc.org/core-2.7.2/ENV.html#method-c-slice).

This PR aims to allow calling slice on ENV by adding an appropriate signature to Sorbet.

### Test plan

```ruby
# sorbet/rbi/sorbet/sorbet.rbi
class Sorbet::Private::Static::ENVClass # monkey patch to test
  extend T::Generic
  include Enumerable

  # Returns a Hash of the given ENV names and their corresponding values
  #
  # ```ruby
  # ENV.slice('foo', 'baz') # => {"foo"=>"0", "baz"=>"2"}
  # ENV.slice('baz', 'foo') # => {"baz"=>"2", "foo"=>"0"}
  # ```
  sig do
    params(names: String)
      .returns(T::Hash[String, String])
  end
  def slice(*names); end
end
```

```ruby
# foo.rb
# typed: true

require 'sorbet-runtime'

class Main
  extend T::Sig

  sig { returns(T.nilable(String))}
  def self.main
    ENV.slice["foo"]
  end
end

Main.main
```

And run

```txt
srb tc foo.rb
No errors! Great job.
```

Without the sorbet.rbi monkey patch, the follow error is raised:

```txt
./foo.rb:10: Method slice does not exist on Sorbet::Private::Static::ENVClass https://srb.help/7003
    10 |    ENV.slice["foo"]
                ^^^^^
  Got Sorbet::Private::Static::ENVClass originating from:
    https://github.com/sorbet/sorbet/tree/5395cb663a1fc8bf4ba7a129f4138e3596530d54/rbi/sorbet/sorbet.rbi#L380:
     380 |::ENV = T.let(T.unsafe(nil), Sorbet::Private::Static::ENVClass)
          ^^^^^
Errors: 1
```

### DIFF
--- a/rbi/sorbet/sorbet.rbi
+++ b/rbi/sorbet/sorbet.rbi
@@ -338,6 +338,18 @@ class Sorbet::Private::Static::ENVClass
   sig {returns(T::Enumerator[Elem])}
   def select!(&blk); end
 
+  # Returns a Hash of the given ENV names and their corresponding values
+  #
+  # ```ruby
+  # ENV.slice('foo', 'baz') # => {"foo"=>"0", "baz"=>"2"}
+  # ENV.slice('baz', 'foo') # => {"baz"=>"2", "foo"=>"0"}
+  # ```
+  sig do
+    params(names: String)
+    .returns(T::Hash[String, String])
+  end
+  def slice(*names); end
+
   sig do
     returns(T::Hash[String, T.nilable(String)])
   end


froydnj:

→ https://go/builds/bui_MzHrmYBP1wnKzD
→ https://go/builds/bui_MzHrSVcn6vws4g

PR #6622, opened 2022-12-13
Delete autoload shim directories for packages onboarded to path-based autoloading

aadi-stripe:
Deletes any directory containing autoload shims for packages onboarded to path-based autoloading using Zeitwerk.

### Motivation

When a package is onboarded to path-based autoloading (i.e. marked as `autoloader_compatibility "strict"`), we need to ensure that any existing directory-level shims for that package are deleted in order prevent "dangling" autoloads.

### Test plan

### DIFF
--- a/common/FileOps.h
+++ b/common/FileOps.h
@@ -43,6 +43,10 @@ public:
     // prevent the caller from needing to try/catch removeDir.
     static bool removeEmptyDir(const std::string &path);
 
+    // Removes directories recursively, but they have to be completely empty. Useful for cleaning up dangling
+    // autoloader shim directories after an autogen run.
+    static void removeEmptyDirsRecursively(const std::string &dirPath);
+
     static void removeFile(const std::string &path);
 
     /**
--- a/common/common.cc
+++ b/common/common.cc
@@ -109,6 +109,45 @@ bool sorbet::FileOps::removeEmptyDir(const string &path) {
     return true;
 }
 
+void sorbet::FileOps::removeEmptyDirsRecursively(const std::string &dirPath) {
+    DIR *dir;
+    struct dirent *entry;
+    auto dirPathCStr = dirPath.c_str();
+
+    if ((dir = opendir(dirPathCStr)) == nullptr) {
+        switch (errno) {
+            case ENOTDIR: {
+                throw sorbet::FileNotDirException();
+            }
+            default:
+                // Mirrors other FileOps functions: Assume other errors are from FileNotFound.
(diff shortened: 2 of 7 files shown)


common/common.cc
```diff
@@ -109,6 +109,45 @@ bool sorbet::FileOps::removeEmptyDir(const string &path) {
...
+            string_view nameview{entry->d_name, namelen};
+            if (nameview == "."sv || nameview == ".."sv) {
+                continue;
+            }
+
+            auto innerDirPath = fmt::format("{}/{}", dirPath, nameview);
+
+            removeEmptyDirsRecursively(std::move(innerDirPath));
+        } else {
+            throw sorbet::CreateDirException(
```
bhang-stripe, line 140:
```suggestion
            throw sorbet::RemoveDirException(
```
nit: should we create a separate exception?

aadi-stripe:

→ https://go/builds/bui_MyfZtEQ9f9ACeR
→ https://go/builds/bui_MyfZ6g8gr8h8Rh

common/common.cc
aadi-stripe: Looks like this is reused for other FileOps methods, so yes but we should do that separately.

aadi-stripe: On second thought, I'll change it here.

aadi-stripe: fixed

common/FileOps.h
```diff
@@ -43,6 +43,10 @@ class FileOps final {
     // prevent the caller from needing to try/catch removeDir.
     static bool removeEmptyDir(const std::string &path);
 
+    // Removes directories recursively, but they have to be completely empty. Useful for cleaning up dangling
+    // autoloader shim directories after an autogen run.
+    static void removeEmptyDirsRecursively(const std::string &dirPath);
```
jez, line 48: This seems like an oddly specific piece of code to have in common/FileOps.h, especially given that it's used in one file.

Can we move this to a helper function in the file that needs it?

aadi-stripe: Ok, but would it be reasonable if it's implemented in `common.cc`? I wanted it to be closer to some of the filesystem primitives that are being already used in that file...since it's using the same ones like `readdir`. Or do you want header and implementation both moved?

jez: eh you're right that it's probably better to keep it where it is.

aadi-stripe:

→ https://go/builds/bui_MyfxS1vE0c1B6n
→ https://go/builds/bui_MyfxUt7dRvBS7W

bhang-stripe (approved):

jez (approved):

PR #6628, opened 2022-12-18
Add a Darwin build of sorbet-staic without a kernel version

will:
This leaves all of the other kernel versioned gmes, but adds one that can be used by any kernel.

### Motivation

If you're using nix, most of the time you get ruby from a build cache, and it comes built as `RUBY_PLATFORM="arm64-darwin22"`. This `RUBY_PLATFORM` is set to 22 even if you're not actually using Ventura, and instead on the previous release of darwin kernel 21. To use a ruby project in nix you use bundix to generate gemsets, and you have to pick a single version of sorbet-static. And so you pick the universal-darwin-22 one, because that's what `RUBY_PLATFORM` is set to.

And that all works okay for a while. However if you ever miss the nix cache, and build a version of ruby from scratch, it will be built as `arm64-darwin21`. This can happen if you're trying out a prerelease, or right now when the version of ruby in nixpkgs is still on 3.1.2 and you want to use 3.1.3. Then if ruby is built as `arm64-darwin21`, bundler wont be able to load the sorbet-static for darwin-22. 

Other gems, like [nokogiri](https://rubygems.org/gems/nokogiri/versions) have dropped the kernel versions from their pre-built releases, and for them there is no problem with bundler selecting it no matter what `RUBY_PLATFORM` is set to. In my project's gems, Sorbet is the only one still adding kernel versions.

I understand that dropping all of them may be too big of a change, but I hope adding a versionless one is small enough to be acceptable. 

### Test plan

I've built a local version-less gem and it seems to work.

### DIFF
--- a/.buildkite/build-static-release.sh
+++ b/.buildkite/build-static-release.sh
@@ -40,6 +40,9 @@ if [[ "mac" == "$platform" ]]; then
         gem build sorbet-static.gemspec
         mv sorbet-static.gemspec.bak sorbet-static.gemspec
     done
+    sed -i.bak "s/Gem::Platform::CURRENT/'universal-darwin'/" sorbet-static.gemspec
+    gem build sorbet-static.gemspec
+    mv sorbet-static.gemspec.bak sorbet-static.gemspec
 else
     gem build sorbet-static.gemspec
 fi


jez:
Sorry for taking so long to get back to you

> I understand that dropping all of them may be too big of a change, but I hope adding a versionless one is small enough to be acceptable.

I'm actually kind of down to get rid of the versions. IIUC, it would only happen for new gem releases—old published versions wouldn't be affected.

Is there any downside? Is it like, only certain Ruby or rubygems or bundler versions support omitting the kernel version? I think that the only reason why we had the kernel version in the first place is because we didn't know any better.

jez (commented):
^ quick question, would love to hear if you know

if you do know the answer, and it's it has been possible to do this for years, if you wouldn't mind just deleting the loop above i'd love to stamp your change.

will:
I don’t _know_ know for sure. But I have a strong feeling that Sorbet can get away with no versioned gems.

I briefly asked @flavorjones who maintains Nokogiri who said the thing to look for is if it's loading different system libs, but since it's the same binary for all versions, I'd think it's the same on all. I only have darwin-21 to check though at the moment.

WRT only doing new versions that's perfectly good I think.

Also with how often there are new versions of this, I'd say if people complain it'd be quick to go back and add the versions again. It'd only be a problem for a small moment in overall time.

flavorjones:
My experience shipping precompiled gems for nokogiri and sqlite3-ruby is that the darwin version does not impact those gems. It's possible that folks who are integrating with Apple libraries may need to be careful here, but for the majority of Rubygems I don't know of any reason to keep the darwin version portion in the gem package name.

simon-poole:
Bumping this one up, as we've now hit Sonora and darwin-23

I'm happy to create a separate issue and PR to specifically add 23 support, but this seems like a better solution

tonybruess:
This can be closed now

PR #6631, opened 2022-12-21
[WIP] Add deserialization logic for handling errors with an UnknownValue type

dustyburwell:
### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/private/serde_transform.rb
+++ b/gems/sorbet-runtime/lib/types/props/private/serde_transform.rb
@@ -91,8 +91,22 @@ module T::Props
             "T::Props::Utils.deep_clone_object(#{varname})"
           end
         when T::Types::Union
+          non_unknown_type = T::Utils.unwrap_unknown_value(type)
           non_nil_type = T::Utils.unwrap_nilable(type)
-          if non_nil_type
+          if non_unknown_type
+            inner = generate(non_unknown_type, mode, varname)
+            if inner.nil?
+              nil
+            else
+              <<~RUBY
+              begin
+                #{inner}
+              rescue _ # Swallow deserialize errors
+                T::Props::UnknownValue._private_construct(T::Props::Utils.deep_clone_object(#{varname}))
+              end
+              RUBY
+            end
+          elsif non_nil_type
             inner = generate(non_nil_type, mode, varname)
             if inner.nil?
               nil
--- a/gems/sorbet-runtime/lib/types/props/unknown_value.rb
+++ b/gems/sorbet-runtime/lib/types/props/unknown_value.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+# typed: strict
+
+module T::Props
+  class UnknownValue
+    sig {returns(T.untyped)}
+    attr_reader :serialized_value
+
+    sig {params(serialized_value: T.untyped).returns(T::Props::UnknownValue)}
+    def _private_construct(serialized_value)
+      @serialized_value = serialized_value
+    end
+  end
+end
--- a/gems/sorbet-runtime/lib/types/utils.rb
+++ b/gems/sorbet-runtime/lib/types/utils.rb
@@ -111,6 +111,22 @@ module T::Utils
     end
   end
 
+  def self.unwrap_unknown_value(type)
+    case type
+    when T::Types::Union
+      non_unknown_types = type.types.reject {|t| t == T::Props::UnknownValue}
+      return nil if type.types.length == non_unknown_types.length
+      case non_unknown_types.length
+      when 0 then nil
+      when 1 then non_unknown_types.first
+      else
+        T::Types::Union::Private::Pool.union_of_types(non_unknown_types[0], non_unknown_types[1], non_unknown_types[2..-1])
+      end
+    else
+      nil
+    end
+  end
+
   # Returns the arity of a method, unwrapping the sig if needed
   def self.arity(method)
     arity = method.arity


dustyburwell:

→ https://go/builds/bui_N1asxmyB0iDEV9
→ https://go/builds/bui_N1as3M1VhKHCx6

dustyburwell:

→ https://go/builds/bui_N1jsUvId3qo86Y
→ https://go/builds/bui_N1js4ROnNCTTmj

dustyburwell:

→ https://go/builds/bui_N1kWWKZkTTGhsZ
→ https://go/builds/bui_N1kWtrkxjkaQIT

dustyburwell:
I also added some additional cases in Stripe's pay-server to show behavior with downstreams:

[https://go/builds/bui_N1lISaYZxRkUmK](https://go/builds/bui_N1lISaYZxRkUmK)

gems/sorbet-runtime/lib/types/props/unknown_value.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+# typed: strict
+
+module T::Props
+  class UnknownValue
```
jez, line 5: What name should this be called? `UnknownValue` makes it sound like you think this will only be used for enum serde failures, but the way you've implemented this, it works for literally any serde failure. That includes: value had the wrong type, key is missing, etc.

I think the name should mention something to that effect, like `DeserializationFailure` or something?

gems/sorbet-runtime/lib/types/props/private/serde_transform.rb
```diff
@@ -151,6 +154,35 @@ def self.generate(type, mode, varname)
...
+      private_class_method def self.handle_unknown_value_subtype(varname, type, mode)
+        inner = generate(type, mode, varname)
+        if inner.nil?
+          nil
+        else
+          case mode
+          when Serialize
+            <<~RUBY
+              if #{varname}.is_a?(T::Props::UnknownValue)
+                T::Props::Utils.deep_clone_object(#{varname}.serialized_value)
```
jez, line 167: Should we be building this on top of the `T::Props::CustomType` abstraction instead of special-casing something here?

For example, should `UnknownValue` implement `serialize` so that it could be serialized like any other value?

gems/sorbet-runtime/test/types/props/serializable.rb
```diff
@@ -1001,6 +1001,10 @@ class RedundantEnumStruct < T::Struct
     prop :deprecated_enum, T.all(MyEnum, T.deprecated_enum(MyEnum.values))
   end
 
+  class UnknownValueStruct < T::Struct
+    prop :unknownable, T.any(MyEnum, T::Props::UnknownValue)
+  end
+
```
jez, line 1007: Can we add a _lot_ more tests?

Some tests:

- using `UnknownValue` with some structs that fails to deserialize (wrong types, missing args)
- using `UnknownValue` with a `BSON` document, where the serialized value is bytes not a string
- using `UnknownValue` with an enum that serializes to a `Symbol` instead of a `String`

gems/sorbet-runtime/lib/types/props/private/serde_transform.rb
```diff
@@ -151,6 +154,35 @@ def self.generate(type, mode, varname)
...
+              else
+                #{inner}
+              end
+            RUBY
+          when Deserialize
+            <<~RUBY
+              begin
+                #{inner}
+              rescue => e # Swallow deserialize errors
```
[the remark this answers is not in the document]
dustyburwell: I think I may be overextending here. You pointed out in another comment that this literally works for any deserialization error. That wasn't the intention at first, I just wanted to catch Enum issues, but then it sort of fit really well with the structure of the code to just let it handle more. In order to handle more the rescue ends up being arbitrarily wide. It looks like `Enum#deserialize` mostly raises `StandardError`, but the one case we really care about is `KeyError`, so we could scope down and just catch that.

I'm a bit surprised that the exceptions signaling bugs in sorbet-runtime would be caught as `StandardError`, though.

gems/sorbet-runtime/lib/types/props/private/serde_transform.rb
```diff
@@ -91,8 +91,11 @@ def self.generate(type, mode, varname)
             "T::Props::Utils.deep_clone_object(#{varname})"
           end
         when T::Types::Union
+          non_unknown_type = T::Utils.unwrap_unknown_value(type)
           non_nil_type = T::Utils.unwrap_nilable(type)
-          if non_nil_type
+          if non_unknown_type
+            handle_unknown_value_subtype(varname, non_unknown_type, mode)
+          elsif non_nil_type
```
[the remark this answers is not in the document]
dustyburwell: I thought I was going to get rubocopped by using assignment inside of a conditional. I'll give it a try.

gems/sorbet-runtime/lib/types/props/unknown_value.rb
```diff
@@ -0,0 +1,21 @@
...
+    attr_reader :serialized_value
+
+    sig {returns(StandardError)}
+    attr_reader :error
+
+    sig {params(serialized_value: T.untyped, error: StandardError).void}
+    def initialize(serialized_value, error)
+      @serialized_value = T.let(serialized_value, T.untyped)
+      @error = T.let(error, StandardError)
+    end
```
froydnj, line 19: Very minor: WDYT about making `new` a private method, and instantiating this via `make_only_from_with_sorbet_runtime` (or other similarly-named "don't use this" method)?

gems/sorbet-runtime/test/types/props/serializable.rb
froydnj: Also maybe worthwhile to test `T.nilable(T.any(MyEnumType, UnknownValue))` and/or `T.any(T.nilable(MyEnumType), UnknownValue))`.

gems/sorbet-runtime/lib/types/props/private/serde_transform.rb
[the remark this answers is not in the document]
froydnj: Might also not be a bad idea to make `Enum#deserialize` raise a subtype of `KeyError` (for backwards compat) and then catch that even-more-specific error here.

gems/sorbet-runtime/lib/types/props/private/serde_transform.rb
dustyburwell: Hmm... possibly. That could clean this up so we don't need the additional type check.

gems/sorbet-runtime/lib/types/props/unknown_value.rb
dustyburwell: > it works for literally any serde failure. That includes: value had the wrong type, key is missing, etc.

Yeah, that wasn't really the intention when I set off, just a bit of a nice outcome that it seems to possibly be more widely applicable.

But we don't really do type checks in deserialization as it is. It's probably a whole performance draining can of worms to do so, but maybe something we could start to allow opting into for new models. The one check we do perform is a nil-check, but that actually happens another layer up. We could move that downward to allow an unexpected nil to result in an `UnknownValue` as well.

froydnj (commented):
👍 to @jez's comments.  Also 👍 to trying to narrow the serde support for it to just enum types somehow.

I think my only (vague) concern is that we now have this special type that's really intended to be used in a specific context of `T.any(EnumType, ...)` and I am (again, vaguely) concerned about what happens if/when the type gets used other places that it was not intended to.

I can't imagine what a better solution would look like, though; the only alternative I can think of is a `unknown-on-deserialize: true` option for `self.prop`, and you'd still need to expose the type of object you create during deserialization because you want people to match on/handle the sorbet-runtime (-internalish) type.  So it's not like you get good information hiding with this alternative.

PR #6633, opened 2022-12-29
fix weird interaction in static-init

froydnj:
### Motivation

Somebody at Stripe reported a bug ([sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%23%20frozen_string_literal%3A%20true%0A%0Aerror_messages%20%3D%20T.let%28%5B%5D%2C%20T%3A%3AArray%5BString%5D%29%0Aparams%20%3D%20%7B%20contents%3A%20%22thing%22%20%7D%0A%0Abegin%0A%20%20T.let%28params%5B%3Acontents%5D%2C%20T%3A%3AHash%5BString%2C%20String%5D%29%0Arescue%20TypeError%0A%20%20error_messages%20%3C%3C%20%22contents%20must%20be%20a%20hash%20of%20%7Bpath%20%3A%20contents%7D%22%0Aend)) with a weird error message.

I don't exactly understand why having the `begin`/`rescue`/`end` in non-value position makes Sorbet think that the derived type is `NilClass`.  I'm sure there's something weird about how the exception handling mechanism works that makes things come out to be `NilClass`.

This PR fixes the bug; intuitively, I think the idea is that the `rescue` body actually has its own result value, which it should generate into a fresh temporary, rather than generating directly into whatever temporary the context of the entire `begin`/`rescue`/`end` is using.

### Test plan

See included automated tests.  Ideally the tests prove that actually using the result of `begin`/`rescue`/`end` still works properly, along with whatever other tests might depend on that.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -841,8 +841,12 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                         exceptions.pop_back();
                     }
 
-                    caseBody = walk(cctx, rescueCase->body, caseBody);
-                    unconditionalJump(caseBody, ensureBody, cctx.inWhat, a.loc);
+                    auto rescueBodyTemp = cctx.newTemporary(core::Names::statTemp());
+                    caseBody = walk(cctx.withTarget(rescueBodyTemp), rescueCase->body, caseBody);
+                    auto assignmentBlock = cctx.inWhat.freshBlock(cctx.loops, caseBody->rubyRegionId);
+                    unconditionalJump(caseBody, assignmentBlock, cctx.inWhat, a.loc);
+                    synthesizeExpr(assignmentBlock, cctx.target, a.loc, make_insn<Ident>(rescueBodyTemp));
+                    unconditionalJump(assignmentBlock, ensureBody, cctx.inWhat, a.loc);
                 }
 
                 // This magic local remembers if none of the `rescue`s match,
--- a/test/testdata/cfg/dealias_with_return.rb.cfg-text.exp
+++ b/test/testdata/cfg/dealias_with_return.rb.cfg-text.exp
@@ -9,7 +9,7 @@ bb0[rubyRegionId=0, firstDead=-1]():
 # backedges
 # - bb4(rubyRegionId=1)
 # - bb6(rubyRegionId=3)
-# - bb9(rubyRegionId=0)
+# - bb10(rubyRegionId=0)
 bb1[rubyRegionId=0, firstDead=-1]():
     <exceptionValue>$4 = <get-current-exception>
     <unconditional> -> bb1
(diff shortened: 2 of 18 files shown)


jez:
I will get to this later today. In the mean time, can you look at whether/how this affects any of these open issues?

https://github.com/sorbet/sorbet/issues?q=is%3Aissue+is%3Aopen+Incompatible+assignment

- https://github.com/sorbet/sorbet/issues/3482
- https://github.com/sorbet/sorbet/issues/1742
- https://github.com/sorbet/sorbet/issues/649

froydnj:
> I will get to this later today. In the mean time, can you look at whether/how this affects any of these open issues?
> 
> https://github.com/sorbet/sorbet/issues?q=is%3Aissue+is%3Aopen+Incompatible+assignment
> 
>     * [Confusing error message referencing a variable before T.let #3482](https://github.com/sorbet/sorbet/issues/3482)
> 
>     * [Sorbet assigns NilClass incorrectly for cfg::Literal in processBinding #1742](https://github.com/sorbet/sorbet/issues/1742)
> 
>     * [Confusing pinning error exposing CFG internals #649](https://github.com/sorbet/sorbet/issues/649)

I can run those examples through a patched Sorbet later today, but given that the fix here is narrowly targeted at the cfg related to an `ast::Rescue`, and `rescue` doesn't feature in those issues, I don't think the fix would have any effect on those issues.  But the existence of those issues and the similarity to the initial bug report, suggest that maybe the right fix is a broader one?

(My understanding of those issues, how the exception-handling mechanisms work, and the initial bug report is that initially, we process the `begin; T.let(...)` and pin the type of the variable returned by the `begin` expression...but then when we're processing the `rescue`, we think that the variable has type `NilClass` (since we consider none of the `begin` statements to have executed), and issue errors.  I guess the extra temporary variable introduced by this PR fixes things because it introduces a separate block, and the pinned type is properly handled in the separate block?)

jez:

→ https://go/builds/bui_N6Y7inTLusOvAY
→ https://go/builds/bui_N6Y7DOCq3zXABn

jez:
Here's a test case that gets a little weird after this change.

### master

#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28x%3A%20Integer%29.void%7D%0Adef%20bar%28x%29%0A%20%20x%20%3D%20begin%0A%20%20%20%20x%20%3D%20T.let%280%2C%20Integer%29%0A%20%20%20%200%0A%20%20rescue%0A%20%20%20%20''%0A%20%20end%0A%20%20T.reveal_type%28x%29%0Aend%0A)

```ruby
# typed: true
extend T::Sig

sig {params(x: Integer).void}
def bar(x)
  x = begin
    x = T.let(0, Integer)
    0
  rescue
    ''
  end
  puts(x)
end

```

#### Observed output, master

```
editor.rb:10: Incompatible assignment to variable declared via let: String("") is not a subtype of Integer https://srb.help/7001
    10 |    ''
            ^^
```

#### Observed output, this branch

```
bar.rb:7: Incompatible assignment to variable declared via let: String("") is not a subtype of Integer https://srb.help/7001
     7 |    x = T.let(0, Integer)
     8 |    0
     9 |  rescue
    10 |    ''
```

jez (approved):
This looks good to me I think. I tried long and hard to think of a test case where we used to report a pinning error and wouldn't after this change. If you hadn't thought about that consequence, maybe give it some thought. And after that feel free to merge.

jez:
I don't _think_ this would be likely to affect peak memory usage, because we don't keep CFGs around, but I wouldn't be surprised if this made inference slightly slower, so I would love if you could try a quick before/after test.

PR #6634, opened 2022-12-31
make `update_testdata_exp.sh` idempotent on a clean tree

froydnj:
### Motivation

For a while now, `tools/scripts/update_testdata_exp.sh` has produced spurious changes to `test/testdata/autogen/generator.rb.autogen.exp` -- changes that do not pass tests.  One needs to remember to revert these changes when using `update_testdata_exp.sh`, which is a source of confusion for new and experienced contributors alike.

I think, but have not verified, that this state of affairs came about after #6204, since that changed how the rewriter pass works with regards to `update_testdata_exp.sh`.

This PR tweaks how we run our tests to more closely mimic how the actual pipeline inside Sorbet works.

### Test plan

See included automated tests.

I would have liked to include some `name-tree{,-raw}.exp` tests to illustrate how changing the pipeline runner changes things...but the magic `runningUnderAutogen` flag is only set for regenerating `autogen.exp` files (so e.g. `name-tree.exp` is generated with the full rewriter pipeline) and the pipeline runner sets `runningUnderAutogen` if the test in question has an `autogen.exp` file (so the test's `name-tree.exp` "file" is generated with a minimal autogen pipeline), which is a mismatch.  I think fixing this is doable, but it's a little more involved than I'd like to tackle right now.


jez (approved):
🎉

PR #6637, opened 2023-01-03
Don't attempt to format files outside of the current workspace

reese:
We previously were attempting to always format files, regardless of whether or not they're in the current workspace. Since the server won't be have access to files not in the current workspace, this file read would fail and crash the server.

### Motivation

Attempting to read files not in the workspace causes the server to crash.

### Test plan

I manually tested this by running the server in pay-server, then opening a random other file (`code ~/other_file.rb`) outside the workspace and saving it, which now saves without crashing.

I would add an automated test for this, but the current automated tests assume the files are in the current workspace, and spinning up a custom test scaffolding for this seemed excessive -- but open to other thoughts there.


jez (approved):

jez:
You could write a test for this in protocol_test_corpus.cc if you want.

PR #6638, opened 2023-01-04
Disallow exporting enum values directly in __package.rb files

aadi-stripe:
Disallows exporting individual enum values directly, instead users will have to export the wrapping enum class.

### Motivation

This is a feature already present in our automatic package-file generation tooling at Stripe. Bringing this into Sorbet to ensure parity between the tools (and prevent the generation tooling from making spurious changes).

### Test plan

### DIFF
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -21,6 +21,7 @@ namespace {
 // For each __package.rb file, traverse the resolved tree and apply the visibility annotations to the symbols.
 class PropagateVisibility final {
     const core::packages::PackageInfo &package;
+    const core::FileRef file;
 
     bool definedByThisPackage(const core::GlobalState &gs, core::ClassOrModuleRef sym) {
         auto pkg = gs.packageDB().getPackageNameForFile(sym.data(gs)->loc().file());
@@ -198,15 +199,23 @@ class PropagateVisibility final {
             auto scope = sym.owner(ctx).asClassOrModuleRef();
             if (scope.data(ctx)->superClass() == core::Symbols::T_Enum()) {
                 if (auto e = ctx.beginError(loc, core::errors::Packager::InvalidExport)) {
+                    std::string scopeName = scope.show(ctx);
+                    std::string_view scopeNameView = std::string_view{scopeName};
                     e.setHeader("Cannot export enum value `{}`. Instead, export the entire enum `{}`", sym.show(ctx),
-                                scope.show(ctx));
+                                scopeNameView);
                     e.addErrorLine(sym.loc(ctx), "Defined here");
+
+                    e.addAutocorrect(core::AutocorrectSuggestion{
+                        fmt::format("Export `{}`", scopeNameView),
+                        {core::AutocorrectSuggestion::Edit{core::Loc{file, loc},
+                                                           fmt::format("export {}", std::move(scopeName))}}});
                 }
             }
         }
     }
 
-    PropagateVisibility(const core::packages::PackageInfo &package) : package{package} {}
+    PropagateVisibility(const core::packages::PackageInfo &package, const core::FileRef file)
+        : package{package}, file{file} {}
 
 public:
     // Find uses of export and mark the symbols they mention as exported.
@@ -299,7 +308,7 @@ public:
         const auto &package = gs.packageDB().getPackageInfo(pkgName);
         ENFORCE(package.exists(), "Package is associated with a file, but doesn't exist");
 
-        PropagateVisibility pass{package};
+        PropagateVisibility pass{package, f.file};
 
         pass.exportPackageRoots(gs);
 
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -192,6 +192,18 @@ class PropagateVisibility final {
                 e.addErrorLine(sym.loc(ctx), "Defined here");
             }
         }
+
+        // If sym is an enum value, it can't be exported directly. Instead, its wrapping enum class must be exported.
+        if (sym.isStaticField(ctx) && sym.owner(ctx).isClassOrModule()) {
+            auto scope = sym.owner(ctx).asClassOrModuleRef();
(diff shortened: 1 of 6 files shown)


website/docs/error-reference.md
```diff
@@ -790,6 +790,34 @@ class A::B < PackageSpec
...
+
+```ruby
+module MyPackage
+  class A < T::Enum
+    enums do
+      Val1 = new
+      Val2 = new
+    end
+  end
+end
```
aadi-stripe, line 812: Open to turning each signature into a new error code.

jez (commented):
How is this checked in the non-Sorbet tool? Does it really special-case enums instead of treating them like static fields?

aadi-stripe:
> How is this checked in the non-Sorbet tool? Does it really special-case enums instead of treating them like static fields?

It special-cases enums, and generates an export for the wrapping class.

jez:
I can see places where people might want to intentionally hide certain enum values. For example, maybe certain states are hidden, internal states, but some states are public states.

But for the time being, it seems like it's a bigger problem that the Ruby script and sorbet disagree, and this PR is already implemented, so we can revisit that choice if we need to in the future.

packager/VisibilityChecker.cc
```diff
@@ -18,9 +18,21 @@ namespace sorbet::packager {
 
 namespace {
 
+static core::SymbolRef getEnumClassForEnumValue(const core::GlobalState &gs, core::SymbolRef sym) {
+    if (sym.isStaticField(gs) && sym.owner(gs).isClassOrModule()) {
+        auto owner = sym.owner(gs);
+        if (owner.asClassOrModuleRef().data(gs)->superClass() == core::Symbols::T_Enum()) {
```
jez, line 24:
```suggestion
        // There's a hidden class like `MyEnum::X$1` between `MyEnum::X` and `T::Enum` in the ancestor chain.
        if (owner.asClassOrModuleRef().data(gs)->superClass() == core::Symbols::T_Enum()) {
```

packager/VisibilityChecker.cc
```diff
@@ -192,9 +204,27 @@ class PropagateVisibility final {
                 e.addErrorLine(sym.loc(ctx), "Defined here");
             }
         }
+
+        // If sym is an enum value, it can't be exported directly. Instead, its wrapping enum class must be exported.
+        auto enumClass = packager::getEnumClassForEnumValue(ctx.state, sym);
```
jez, line 209: You're already in the `packager` namespace so you don't need that prefix if you don't want:

```suggestion
        auto enumClass = getEnumClassForEnumValue(ctx.state, sym);
```

test/cli/package-disallow-enum-value-exports/test.sh
```diff
@@ -0,0 +1,5 @@
+cd test/cli/package-disallow-enum-value-exports || exit 1
+
+../../../main/sorbet --silence-dev-message --stripe-packages --max-threads=0 . 2>&1
```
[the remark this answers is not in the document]
aadi-stripe: No reason for it to be a CLI test, other than I'm used to writing them. Can change if you feel strongly.

packager/VisibilityChecker.cc
```diff
@@ -192,9 +204,27 @@ class PropagateVisibility final {
...
+        if (enumClass.exists()) {
+            if (auto e = ctx.beginError(loc, core::errors::Packager::InvalidExport)) {
+                std::string enumClassName = enumClass.show(ctx);
+                std::string_view enumClassNameView = std::string_view{enumClassName};
```
jez, line 213:
```suggestion
                auto enumClassName = enumClass.show(ctx);
```

I don't think there's much point trying to use `string_view` here. The usages inside string formatting methods will all take the string by reference anyways, and they're already going to allocate a new string when constructing the newly-formatted string, so it doesn't matter if you `move(enumClassName)` in—it's going to copy and allocate a new string to prefix it with `export` anyways.

With that in mind, you can simplify a lot of this code by just operating on `enumClassName`.

packager/VisibilityChecker.cc
```diff
@@ -377,7 +407,13 @@ class VisibilityCheckerPass final {
                 } else {
                     e.addErrorLine(definedHereLoc, "Defined here");
                 }
-                if (auto exp = pkg.addExport(ctx, lit.symbol)) {
+
+                auto symToExport = lit.symbol;
+                auto enumClass = packager::getEnumClassForEnumValue(ctx.state, symToExport);
```
jez, line 412:
```suggestion
                auto enumClass = getEnumClassForEnumValue(ctx.state, symToExport);
```

website/docs/error-reference.md
jez: Yeah I think we want to make a new code for this. It's not obvious to me from skimming whether I care about this section of the 3721 docs or the previous section, so I think that explicitly being able to drop people into the relevant section will help.

jez (approved):
lgtm overall, but i have some small fixes i would love addressed first

test/cli/package-disallow-enum-value-exports/test.sh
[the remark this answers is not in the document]
jez: no worries, was just curious

packager/VisibilityChecker.cc
aadi-stripe: Sorry, don't quite understand this comment.

jez: I just wanted to leave a note why you need the `->superClass()` here. naively, you might expect `enum.owner(gs)` to be `T_Enum` but that's not how it works, because of the rewrites we do in `TEnum.cc` to support exhaustiveness on enums. feel free to reword this comment however you want to capture why we need the `->superClass` here.

website/docs/error-reference.md
aadi-stripe: I can make this change (breaking out the error codes) in a different PR.

aadi-stripe:
@jez Made the changes if you want to take another look before I merge

jez (approved):

PR #6639, opened 2023-01-04
Attempt to fix //test/cli:test_folder-input-not-found

jez:
<https://buildkite.com/sorbet/sorbet/builds/25777#01857c76-e8f5-45c8-b991-6f19076cd69a>

We're getting a sanitizer failure on this CLI test. I was unable to get
this to repro locally.

The error is coming from a logger inside `addFilesFromDir`. There are
only two places in that function that log, and one of them is calling
`e.what()` which returns a `const char *`. I'm wondering if it's
possible that the `std::string` temporaries that we're creating before
constructing `FileNotFoundException`'s are getting released somehow.

Some changes which might help:

- We now catch by reference, which seems to be a best practice?

  <https://stackoverflow.com/questions/2023032/catch-exception-by-pointer-in-c>

- We're no longer throwing anonymous `std::string` temporaries in most
  file-related exceptions. As far as I can tell, this _shouldn't_
  matter, because the string should live as least as long as it takes
  for `FileNotFoundException` to finish using it for construction:

  <https://stackoverflow.com/questions/63270037/c-anonymous-string-destruction-time>

  But I think this shouldn't hurt anyways (it should now live as long as
  the enclosing scope, instead of the enclosing expression). The call
  site was going to take it by `const string&` anyways, so it's not like
  we're doing an extra copy. But this change is mostly a shot in the
  dark.

- We now use `"{}", e.what()`, just in case `what()` had `{}` in it
  and were to be treated as a format string (which I'm pretty sure it
  doesn't for this test, but better safe than sorry).

I'm not confident this will help but I'm pretty confident it won't hurt.

### DIFF
--- a/common/common.cc
+++ b/common/common.cc
@@ -48,11 +48,13 @@ string sorbet::FileOps::read(const string &filename) {
         fclose(fp);
         if (readBytes != contents.size()) {
             // Error reading file?
-            throw sorbet::FileNotFoundException(fmt::format("Error reading file: `{}`: {}", filename, errno));
+            auto msg = fmt::format("Error reading file: `{}`: {}", filename, errno);
+            throw sorbet::FileNotFoundException(msg);
         }
         return contents;
     }
-    throw sorbet::FileNotFoundException(fmt::format("Cannot open file `{}`", filename));
+    auto msg = fmt::format("Cannot open file `{}`", filename);
+    throw sorbet::FileNotFoundException(msg);
 }
 
 void sorbet::FileOps::write(const string &filename, const vector<uint8_t> &data) {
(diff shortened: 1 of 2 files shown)


froydnj (approved):
I am +1 on catching the exception by reference and +1 on using a proper format string instead of `e.what()` directly.

I am...-0.5 on using named strings to pass to sorbet exceptions: `std::logic_error` is going to copy the string data anyway (https://en.cppreference.com/w/cpp/error/logic_error ; libc++, at least, copies, and I'm pretty sure libgcc/libstdc++ do too), so whatever is returned by `e.what()` is going to be something internal to the `logic_error` itself rather than whatever we passed in initially.  I think the extra variable is just adding clutter.

(The error is just bizarre, given that it's coming from `spdlog` itself; it's almost like the list of loggers got pulled out from underneath the ongoing logging...)

jez:
Yeah I’m pretty sure the named tmp shouldn’t matter either, but I’m grasping at straws, so I’ll land. Maybe if this still doesn’t fix it we can put it back to what it was and try something else.

PR #6640, opened 2023-01-04
Renaming a method parameter should update the sig too

### LINKED ISSUE #6463: "Rename symbol" on a function param doesn't update sig
#### Input

[→ View on sorbet.run](https://sorbet.run)

```ruby
sig {params(x: Integer).void}
def foo(x)
#       ^ Right click, rename symbol, change to "bar"
  puts(x + 1)
end
```

#### Observed output

```ruby
#           v did not change!
sig {params(x: Integer).void}
def foo(bar)
  puts(bar + 1)
end
```

#### Expected behavior

```ruby
# Both the sig and param should rename
sig {params(bar: Integer).void}
def foo(bar)
  puts(bar + 1)
end
```

jez:
### Motivation

Fixes #6463

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/references_with_sig.rb
+++ b/test/testdata/lsp/references_with_sig.rb
@@ -1,0 +1,13 @@
+# typed: true
+
+class A
+  extend T::Sig
+
+  sig { params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped).returns(T::Array[T.untyped]) }
+  #            ^^^^ usage: arg0
+  def example1(arg0, arg1 = nil, arg2:, arg3: nil)
+    #          ^^^^ def: arg0
+    [arg0, arg1, arg2, arg3]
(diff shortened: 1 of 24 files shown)


test/testdata/lsp/references_with_sig.rb
```diff
@@ -0,0 +1,13 @@
+# typed: true
+
+class A
+  extend T::Sig
+
+  sig { params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped).returns(T::Array[T.untyped]) }
+  #            ^^^^ usage: arg0
+  def example1(arg0, arg1 = nil, arg2:, arg3: nil)
+    #          ^^^^ def: arg0
```
froydnj, line 9: This is a clever way of making things work.

froydnj (approved):
`:chefkiss:`

PR #6641, opened 2023-01-05
Do not implicitly export behavior-defining namespaces in VisibilityChecker

aadi-stripe:
Changes the functionality of the package visibility checker to disallow behavior-defining parent namespaces from being implicitly visible. This cleans up a fundamental flaw in the packaging design, and also helps prevent our package generation tooling from making spurious changes. 

See the added unit test for more details.

Currently, parent namespaces of exported constants, even if not themselves exported, are automatically deemed "visible" to importing packages. This allows for the developer convenience of aliasing the parent namespace with a shorter name, and accessing the public constants underneath by prefacing with the short name (instead of the longer fully qualified name). Effectively, what is exposed to an importing package is the public API of the exporting package.

For example, assume a package `Opus::Foo` defines a module `Opus::Foo::Parent`, which defines 3 constants:
* `Opus::Foo::Parent::A`
* `Opus::Foo::Parent::B`
* `Opus::Foo::Parent::C`

Assume that `Opus::Foo::Parent::A`, `Opus::Foo::Parent::B` are exported, but `Opus::Foo::Parent::C` or `Opus::Foo::Parent` itself are not exported. 

Then, an importing package can still refer to `Opus::Foo::Parent` for aliasing, but can only access `A` and `B`, not `C`.
```
P = Opus::Foo::Parent

P::A # ok
P::B # ok
P::C # not ok!
```

The importing only has access to the public API of `Opus::Foo::Parent`. This complies with the private-by-default paradigm.

However, there is one exception to the compliance, namely, if the un-exported parent namespace defines behavior (methods). If `Opus::Foo::Parent` has methods underneath it, they will be automatically made available to the importing package even if `Opus::Foo::Parent` is not exported. This obviously violates the paradigm:

```
P = Opus::Foo::Parent # assume this defines a method Opus::Foo::Parent.foo

P.foo # calling a method is currently ok, but should not be!!
```

 I'm PR-ing this fix to prevent behavior-defining namespaces from becoming implicitly visible in this way.

The other changes I made:
* Adding a new flag on `ClassOrModule` called `isBehaviorDefining`, which is set in the namer and used by the visibility checker.

### Motivation

This fixes 2 issues:
* Cleans up a violation in the private-by-default paradigm
* Helps bring the Sorbet enforcement to parity with our package generation tooling. The generation tooling does not consider non-behavior defining constants exportable, thus, we have to enforce the same with Sorbet to catch issues before the tooling makes spurious changes.

### Test plan

### DIFF
--- a/core/Symbols.h
+++ b/core/Symbols.h
@@ -378,13 +378,15 @@ public:
         bool isPrivate : 1;
         bool isUndeclared : 1;
         bool isExported : 1;
+        bool isBehaviorDefining : 1;
 
-        constexpr static uint16_t NUMBER_OF_FLAGS = 10;
+        constexpr static uint16_t NUMBER_OF_FLAGS = 11;
         constexpr static uint16_t VALID_BITS_MASK = (1 << NUMBER_OF_FLAGS) - 1;
 
         Flags() noexcept
             : isClass(false), isModule(false), isAbstract(false), isInterface(false), isLinearizationComputed(false),
-              isFinal(false), isSealed(false), isPrivate(false), isUndeclared(false), isExported(false) {}
+              isFinal(false), isSealed(false), isPrivate(false), isUndeclared(false), isExported(false),
+              isBehaviorDefining(false) {}
 
         uint16_t serialize() const {
             // Can replace this with std::bit_cast in C++20
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1181,6 +1181,7 @@ private:
             if (klass.definesBehavior) {
                 auto &behaviorLocs = classBehaviorLocs[symbol];
                 behaviorLocs.emplace_back(ctx.locAt(klass.declLoc));
+                symbol.data(ctx)->flags.isBehaviorDefining = true;
             }
         }
         auto singletonClass = symbol.data(ctx)->singletonClass(ctx); // force singleton class into existence
--- a/test/cli/package-implicit-parent-namespace-export/other_package/reference.rb
+++ b/test/cli/package-implicit-parent-namespace-export/other_package/reference.rb
@@ -1,0 +1,13 @@
+# typed: strict
+
+module OtherPackage
+  MPP = MyPackage::ParentNamespace # Allowed due to implicit parent namespace exporting
+
+  MPP::Const1
+  MPP::Const2
+  MPP::PrivateConst3 # Should be an ERROR! Private constant.
+
+  BPP = MyPackage::BehaviorDefiningParentNamespace # should be an ERROR! Not allowed since namespace defines behavior.
+
+  BPP::Const1 # This is ok since MyPackage::BehaviorDefiningParentNamespace::Const1 is explicitly exported :)
+end
--- a/test/testdata/packager/export_for_test/foo/foo.rb
+++ b/test/testdata/packager/export_for_test/foo/foo.rb
@@ -27,6 +27,10 @@ module Opus::Foo
 
   Opus::Util::Nesting::Public.public_method
 
+  # util/__package.rb exposed via export_for_test, cannot access from here:
+  Opus::Util::Nesting.nesting_method
+# ^^^^^^^^^^^^^^^^^^^ error: `Opus::Util::Nesting` resolves but is not exported from `Opus::Util`
+
 
   # via test_import Opus::TestImported
   Opus::TestImported::TIClass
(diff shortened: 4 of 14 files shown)


namer/namer.cc
```diff
@@ -2324,7 +2324,7 @@ vector<ast::ParsedFile> symbolizeTrees(const core::GlobalState &gs, vector<ast::
         for (auto result = fileq->try_pop(job); !result.done(); result = fileq->try_pop(job)) {
             if (result.gotItem()) {
                 Timer timeit(gs.tracer(), "naming.symbolizeTreesOne", {{"file", string(job.file.data(gs).path())}});
-                core::Context ctx(gs, core::Symbols::root(), job.file);
+                core::MutableContext ctx(gs, core::Symbols::root(), job.file);
```
froydnj, line 2327: This change (and accompanying changes to `TreeSymbolizer`) are not OK.  I understand that the mutation of `GlobalState` that's necessary here is confined to a single bitflag, and that racy writes to the bitflag might theoretically be OK given hardware and compiler behavior.  However:

1. There's no guarantee that future modifications to `TreeSymbolizer` will be so careful; we generally rely on the types of `*Context` (and `const GlobalState` vs. `GlobalState`) to provide (some) safety guarantees when writing parallel code.  It would be very easy to come along and use some non-const method on `ctx` or `ctx.state`, not realizing that _this_ `MutableContext` is actually being used in a threaded context, and that would lead to some very difficult to diagnose bugs.
2. The racy modification to `isBehaviorDefining` is undefined behavior according to the C++ spec, so even if it doesn't cause problems today, it could tomorrow (and/or make running with ThreadSanitizer even harder than it is today).

I would also have to think about whether this flag still works correctly in the presence of the LSP fast path; I think it does today, because class changes take the slow path, but this flag would be a trap for the unwary if we ever made class changes take the fast path.

froydnj (changes requested):
I have no problem with the behavior change, but the way the behavior change is accomplished is going to need some additional work.

namer/namer.cc
aadi-stripe: Yep, I agree, I was going to back this out and come up with another way. Can discuss today.

aadi-stripe: And yeah, it makes some LSP fast path tests fail

aadi-stripe:
> I have no problem with the behavior change, but the way the behavior change is accomplished is going to need some additional work.

Fixed as per conversations

core/Symbols.h
```diff
@@ -602,10 +602,20 @@ class ClassOrModule final {
...
+
+    inline void markAsBehaviorDefining() {
+        isBehaviorDefining_ = true;
+    }
+
 private:
     friend class serialize::SerializerImpl;
     friend class GlobalState;
 
+    bool isBehaviorDefining_ = false;
```
aadi-stripe, line 617: Original plan was to have it as a bit inside flags, but then in order to not serialize it, it felt like I'd have to make changes where I'm mindful of endian-ness. Could be wrong.

test/cli/package-implicit-parent-namespace-export/other_package/reference.rb
```diff
@@ -0,0 +1,13 @@
+# typed: strict
+
+module OtherPackage
+  MPP = MyPackage::ParentNamespace # Allowed due to implicit parent namespace exporting
+
+  MPP::Const1
+  MPP::Const2
+  MPP::PrivateConst3 # Should be an ERROR! Private constant.
```
jez, line 8: If you make this into a testdata test, these "should be an error" comments never get out of date 😉

test/testdata/packager/export_for_test/foo/foo.rb
```diff
@@ -27,9 +27,6 @@ def self.stub_stuff!; end
 
   Opus::Util::Nesting::Public.public_method
 
-  # util/__package.rb exposed via export_for_test, cannot access from here:
-  Opus::Util::Nesting.nesting_method
```
jez, line 31: Rather than delete this, what do you think about changing the test to assert the new expected behavior?

test/cli/package-implicit-parent-namespace-export/other_package/reference.rb
aadi-stripe: Haha, true.

jez (approved):

test/testdata/packager/export_for_test/foo/foo.rb
aadi-stripe: Fixed

froydnj (approved):
Thanks for working through a better way to do this!

PR #6644, opened 2023-01-06
add more context for behavior definition errors outside package namespaces

froydnj:
### Motivation

In #6359, it was suggested that we could add some more context, specifically directing the user to the enclosing scope that things were being defined in.  This PR does that.

I would have liked to track the locs directly in `nameParts`, rather than having a separate member variable, but it turned out that this required propagating `vector<pair<core::NameRef, core::LocOffsets>>` into places that took only `vector<core::NameRef>`.  I hacked around one instance with `if constexpr` in lambda functions (!), but there were too many other places to update, so I settled for the patch you see here.  The `NameFormatter::operator()` changes are a taste of the sort of thing that was necessary.

### Test plan

See included automated tests.

The only thing I am not sure about is what happens when the enclosing scope is something like `class Name::Subname`; I think the code as-is only highlights `Subname`.  It would be better to highlight the full name, but I don't think we have enough information to track that.

### DIFF
--- a/test/cli/package-prefix-enforcement/test.out
+++ b/test/cli/package-prefix-enforcement/test.out
@@ -8,20 +8,32 @@ nested/nested.rb:5: File belongs to package `Root::Nested` but defines a constan
 nested/nested.rb:40: Class or method behavior may not be defined outside of the enclosing package namespace `Root::Nested` https://srb.help/3713
     40 |  sig {returns(NilClass)}
           ^^^^^^^^^^^^^^^^^^^^^^^
-  Note:
-    Attempting to define class or method behavior in package namespace `Root`
+    nested/nested.rb:36: Attempting to define class or method behavior in `Root`
+    36 |module Root
+               ^^^^
+    __package.rb:3: Enclosing package declared here
+     3 |class Root < PackageSpec
+        ^^^^^^^^^^^^^^^^^^^^^^^^
 
 nested/nested.rb:41: Class or method behavior may not be defined outside of the enclosing package namespace `Root::Nested` https://srb.help/3713
     41 |  def self.method
           ^^^^^^^^^^^^^^^
-  Note:
-    Attempting to define class or method behavior in package namespace `Root`
+    nested/nested.rb:36: Attempting to define class or method behavior in `Root`
+    36 |module Root
+               ^^^^
+    __package.rb:3: Enclosing package declared here
+     3 |class Root < PackageSpec
+        ^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


test/cli/package-prefix-enforcement/test.out
```diff
@@ -8,20 +8,32 @@ nested/nested.rb:5: File belongs to package `Root::Nested` but defines a constan
...
     41 |  def self.method
           ^^^^^^^^^^^^^^^
-  Note:
-    Attempting to define class or method behavior in package namespace `Root`
+    nested/nested.rb:36: Attempting to define class or method behavior in `Root`
+    36 |module Root
+               ^^^^
+    __package.rb:3: Enclosing package declared here
+     3 |class Root < PackageSpec
+        ^^^^^^^^^^^^^^^^^^^^^^^^
```
jez, line 26: It looks like we're using "enclosing" in two different ways here:

- the error header says that the "enclosing package namespace" is `Root::Nested`
- this error line says that the "enclosing package" is `Root`

test/cli/package-prefix-enforcement/test.out
```diff
@@ -8,20 +8,32 @@ nested/nested.rb:5: File belongs to package `Root::Nested` but defines a constan
...
     37 |  extend T::Sig
           ^^^^^^^^^^^^^
-  Note:
-    Attempting to define class or method behavior in package namespace `Root`
+    nested/nested.rb:36: Attempting to define class or method behavior in `Root`
+    36 |module Root
+               ^^^^
+    __package.rb:3: Enclosing package declared here
+     3 |class Root < PackageSpec
+        ^^^^^^^^^^^^^^^^^^^^^^^^
```
jez, line 36: In my experience, packaging errors are the most confusing when I thought I was working in one package and suddenly I've opened up a file or folder that's actually in a nested package. For example `lib/my_app/` and `lib/my_app/data/`.

In times when this happens, it's usually getting to see that there are two `__package.rb` files in play that clear up the confusion. One suggestion might be:

```
nested/nested.rb:37: This file must only define behavior in enclosing package `Root::Nested` https://srb.help/3713
    37 |  extend T::Sig
          ^^^^^^^^^^^^^
  nested/nested.rb:36: Defining behavior in `Root` instead:
    36 |module Root
        ^^^^^^^^^^^
  nested/__package.rb:3: Enclosing package `Root::Nested` declared here
   3 |class Root::Nested < PackageSpec
  __package.rb.rb:3: Package `Root` declared here
   3 |class Root < PackageSpec
      ^^^^^^^^^^^^^^^^^^^^^^^^
```

(not strongly attached to this)

jez (approved):
some suggestions for further improvements but this is better than nothing so pre-approving. feel free to keep workshopping and merge when you want

PR #6645, opened 2023-01-06
eliminate double-checked locking in autoloader generation

froydnj:
### Motivation

Manual mutexes are kind of unusual in Sorbet, and there's no need for them here; we can just map-reduce modification status just like most other parallel loops in Sorbet.

### Test plan

See included automated tests.


jez (approved):
🙏 🙏

PR #6647, opened 2023-01-07
Add `export_all!` to packages

aisamanra:
### Motivation
This adds a package feature which recognizes `export_all!` in lieu of explicit exports. If a package marks `export_all!`, then
1. any constant owned by the package can be imported by any other package, and 
2. it is an error to include any `export` line in the package itself.

For packages which have a very wide surface area, export many constants, and currently see a lot of churn because of edits to the export list, this could prevent a lot of churn and merge conflicts.

### Test plan

See included automated tests. Added a test suite that shows `export_all!` behavior, and another that shows the same behavior with tests. This package also includes a subpackage to show that the export walk does stop when it hits the subpackage boundary.

### DIFF
--- a/core/packages/PackageInfo.cc
+++ b/core/packages/PackageInfo.cc
@@ -77,15 +77,16 @@ core::ClassOrModuleRef getParentNamespaceSym(const core::GlobalState &gs, const 
 core::ClassOrModuleRef lookupNameOn(const core::GlobalState &gs, const core::ClassOrModuleRef root,
                                     const std::vector<core::NameRef> &name) {
     auto curSym = root;
+    if (!curSym.exists()) {
+        return {};
+    }
+
     for (const auto part : name) {
         auto member = curSym.data(gs)->findMember(gs, part);
-        if (!member.isClassOrModule()) {
+        if (!member.exists() || !member.isClassOrModule()) {
             return {};
         }
         curSym = member.asClassOrModuleRef();
-        if (!curSym.exists()) {
-            return {};
-        }
     }
 
     return curSym;
--- a/core/packages/PackageInfo.cc
+++ b/core/packages/PackageInfo.cc
@@ -74,34 +74,34 @@ core::ClassOrModuleRef getParentNamespaceSym(const core::GlobalState &gs, const 
     return core::Symbols::root();
 }
 
-core::ClassOrModuleRef PackageInfo::getPackageScope(const core::GlobalState &gs) const {
-    auto curSym = core::Symbols::root();
-
-    for (const auto part : fullName()) {
-        curSym = curSym.data(gs)->findMember(gs, part).asClassOrModuleRef();
+core::ClassOrModuleRef lookupNameOn(const core::GlobalState &gs, const core::ClassOrModuleRef root,
+                                    const std::vector<core::NameRef> &name) {
+    auto curSym = root;
+    for (const auto part : name) {
+        auto member = curSym.data(gs)->findMember(gs, part);
+        if (!member.isClassOrModule()) {
+            return {};
+        }
+        curSym = member.asClassOrModuleRef();
         if (!curSym.exists()) {
-            return curSym;
+            return {};
         }
     }
 
     return curSym;
 }
 
+core::ClassOrModuleRef PackageInfo::getPackageScope(const core::GlobalState &gs) const {
+    return lookupNameOn(gs, core::Symbols::root(), fullName());
+}
+
 core::ClassOrModuleRef PackageInfo::getPackageTestScope(const core::GlobalState &gs) const {
-    auto curSym = core::Symbols::root().data(gs)->findMember(gs, core::Names::Constants::Test()).asClassOrModuleRef();
-
-    if (!curSym.exists()) {
-        return curSym;
+    auto testSym = core::Symbols::root().data(gs)->findMember(gs, core::Names::Constants::Test());
+    if (!testSym.isClassOrModule()) {
+        return {};
     }
 
-    for (const auto part : fullName()) {
-        curSym = curSym.data(gs)->findMember(gs, part).asClassOrModuleRef();
-        if (!curSym.exists()) {
-            return curSym;
-        }
-    }
-
-    return curSym;
+    return lookupNameOn(gs, testSym.asClassOrModuleRef(), fullName());
 }
 
 // Given a package named Project::MyPackage, returns the class/module ref corresponding to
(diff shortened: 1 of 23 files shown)


packager/VisibilityChecker.cc
```diff
@@ -383,17 +383,17 @@ class VisibilityCheckerPass final {
...
         }
 
+        auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
         bool isExported = false;
         if (lit.symbol.isClassOrModule()) {
-            isExported = lit.symbol.asClassOrModuleRef().data(ctx)->flags.isExported;
+            isExported = pkg.exportAll() || lit.symbol.asClassOrModuleRef().data(ctx)->flags.isExported;
         } else if (lit.symbol.isFieldOrStaticField()) {
             isExported = lit.symbol.asFieldRef().data(ctx)->flags.isExported;
         }
```
froydnj, line 393: No automatic `export_all!` behavior for static fields?

aisamanra:

→ https://go/builds/bui_N8hiuWkdqVINSV
→ https://go/builds/bui_N8hiJZK3Pho7vA

aisamanra:
I think this is ready. I originally used a different approach—checking the package metadata each time we saw a constant—but I think that could be risky because it was introducing package lookup in a path that previously didn't have it. Instead, this uses the same visibility-propagation mechanism as before, but it specifically starts from both the package root and test root. I've added tests to make sure this doesn't have unexpected behavior like traversing into subpackages.

packager/packager.cc
```diff
@@ -206,7 +210,9 @@ class PackageInfoImpl final : public core::packages::PackageInfo {
     vector<Export> exports_;
 
     // Whether the code in this package is compatible for path-based autoloading.
-    bool strictAutoloaderCompatibility_;
+    bool strictAutoloaderCompatibility_ : 1;
+    // Whether this package should just export everything
+    bool exportAll_ : 1;
```
froydnj, line 215: This is a nit, but given that there are not a lot of `PackageInfoImpl` structures (even in Stripe's codebase), and that we have a bunch of space to play with in the existing structure, I think turning these into bitfields is premature optimization.

test/testdata/packager/export_all/__package.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
+class Foo::Bar < PackageSpec
+  export_all!
+
+  export Foo::Bar::Thing
+       # ^^^^^^^^^^^^^^^ error: Package `Foo::Bar` declares `export_all!` and therefore should not use explicit exports
```
froydnj, line 9: OOC, does this same error happen if the ordering of these declarations is reversed?

packager/packager.cc
```diff
@@ -1003,6 +1009,10 @@ struct PackageInfoFinder {
             send.addPosArg(prependName(move(importArg), core::Names::Constants::PackageSpecRegistry()));
         }
 
+        if (send.fun == core::Names::exportAll() && send.numPosArgs() == 0) {
+            info->exportAll_ = true;
+        }
```
froydnj, line 1014: I don't know whether we have tests for other package spec methods like this, but could we add a test passing arguments to `export_all!` and verify that errors are detected?

froydnj (changes requested):
I think this PR is fundamentally on the right track; the only thing that I want to block on is the question about the ordering of `export_all!` vs. `export`.  Everything else can be treated as non-blocking/suggestions/opportunities for discussion.

core/packages/PackageInfo.cc
```diff
@@ -74,6 +74,36 @@ core::ClassOrModuleRef getParentNamespaceSym(const core::GlobalState &gs, const
     return core::Symbols::root();
 }
 
+core::ClassOrModuleRef PackageInfo::getPackageScope(const core::GlobalState &gs) const {
+    auto curSym = core::Symbols::root();
+
+    for (const auto part : fullName()) {
+        curSym = curSym.data(gs)->findMember(gs, part).asClassOrModuleRef();
```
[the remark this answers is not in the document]
aisamanra: That's fair! This was almost copy/pasted from `PackageInfo::getRootSymbolForAutocorrectSearch`, which does the same thing; I might change that while I'm at it.

test/testdata/packager/export_all/__package.rb
aisamanra: Yes: the parsing logic accumulates exports first, and then these errors are raised only after the entire `__package.rb` file has been looked at.

packager/packager.cc
aisamanra: Yup! I've added that. The way typechecking these works is that we actually treat them as normal Ruby methods, so they're subject to the same typechecking constraints as any other methods. (In particular, when we enter these methods into `GlobalState` (e.g. [here](https://github.com/sorbet/sorbet/pull/6647/files#diff-2c71c670c6721566158753a99c725a4826efc01a8a6a3c53a32c98fe3889252c)) we also give their argument structure; this one takes no parameters, so if you pass it a parameter, then inference will raise the same error it would raise for any method with too many arguments.)

packager/packager.cc
aisamanra: Removed!

core/packages/PackageInfo.cc
```diff
@@ -74,6 +74,36 @@ core::ClassOrModuleRef getParentNamespaceSym(const core::GlobalState &gs, const
...
+
+core::ClassOrModuleRef PackageInfo::getPackageTestScope(const core::GlobalState &gs) const {
+    auto curSym = core::Symbols::root().data(gs)->findMember(gs, core::Names::Constants::Test()).asClassOrModuleRef();
+
+    if (!curSym.exists()) {
+        return curSym;
+    }
+
+    for (const auto part : fullName()) {
+        curSym = curSym.data(gs)->findMember(gs, part).asClassOrModuleRef();
```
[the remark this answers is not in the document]
aisamanra: Both of these now use the same helper function which is cautious about this.

aisamanra:
I believe I've addressed all the above feedback! Don't hesitate to ping if you have other questions!

test/testdata/packager/export_all/__package.rb
```diff
@@ -0,0 +1,14 @@
...
+# enable-packager: true
+
+class Foo::Bar < PackageSpec
+  export_all!
+
+  export_all!(:with_argument)
+            # ^^^^^^^^^^^^^^ error: Too many arguments
+
+  export Foo::Bar::Thing
+       # ^^^^^^^^^^^^^^^ error: Package `Foo::Bar` declares `export_all!` and therefore should not use explicit exports
```
froydnj, line 12: ❤️ for adding the ordering inversion here.

froydnj (approved):
Thanks!

test/testdata/packager/export_all/foo_bar_baz/__package.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
```
jez, line 3: @aisamanra would you mind adding a fast path test for this new feature?

PR #6651, opened 2023-01-09
Fix compilation errors in fuzz test.

varungandhi-src:
### Motivation

Without this, running `bazel test //...` in the root fails. (Even with this patch, it still fails due to a [generated by conflicting actions](https://sorbet-ruby.slack.com/archives/CFT8Y4909/p1673239228231249) error, but that can be worked around by using `--notrim_test_configuration`)

### Test plan

Should be covered by existing tests.


froydnj (approved):
Thanks!

PR #6652, opened 2023-01-09
Bump doctest to 2.4.9

varungandhi-src:
### Motivation

2.4.9 includes a fix for glibc 2.34+ where SIGSTKSZ is no longer a constant; using 2.4.1 leads to a compilation failure on Ubuntu 22.04 due to newer glibc.

See also: https://github.com/doctest/doctest/issues/473

### Test plan

Should be covered by existing tests.


froydnj (approved):
Thanks!

PR #6653, opened 2023-01-10
Resolver suggestions on export statements in package files cannot cross package boundaries

aadi-stripe:
Currently, it is possible for constant resolution errors in `export` statements in `__package.rb` files to provide autocorrect suggestions that are outside the given package. These can never be valid suggestions, and so we restrict the fuzzy search on export statements to the given package. E.g.:
```
# my_package/__package.rb
class MyPackage < PackageSpec
  export MyPackage::A # resolution ERROR as there is no MyPackage::A
end

# my_package/b.rb
class MyPackage::B
end

root/__package.rb
class Root < PackageSpec
end

root/a.rb
class Root::A
end
```
For the resolution error above, there is only 1 valid suggestion: `MyPackage::B`. However, the edit-distance-based algorithm provides `Root::A` as the current suggestion. By ensuring that the fuzzy search stops when it reaches `MyPackage`, we ensure that `MyPackage::B` is output in this case.

### Motivation

Reduce significant user pain with respect to packaging in the Stripe codebase.

### Test plan


test/cli/package-error-unresolved-export/test.out
```diff
@@ -5,13 +5,6 @@ __package.rb:7: Unable to resolve constant `WildlyMisspelled` https://srb.help/5
 __package.rb:8: Unable to resolve constant `NameWithTipo` https://srb.help/5002
      8 |  export Foo::BasePkg::NameWithTipo
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
-  Did you mean `Foo::Other::NameWithTipo`? Use `-a` to autocorrect
-    __package.rb:8: Replace with `Foo::Other::NameWithTipo`
-     8 |  export Foo::BasePkg::NameWithTipo
-                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
-    other/other.rb:4: `Foo::Other::NameWithTipo` defined here
-     4 |  class NameWithTipo; end
-          ^^^^^^^^^^^^^^^^^^
```
aadi-stripe, line 14: This was an invalid suggestion

core/packages/PackageInfo.cc
```diff
@@ -57,4 +57,38 @@ string PackageInfo::show(const core::GlobalState &gs) const {
...
+}
+
+bool belongsToTestNamespace(const core::GlobalState &gs, const core::SymbolRef sym) {
+    auto testSym = getTestSym(gs);
+    if (!testSym.exists()) {
+        return false;
+    }
+
+    return sym.isUnderNamespace(gs, testSym);
+}
```
froydnj, line 71: Slight preference for inlining this function into `getRootSymbolForAutocorrectSearch` so that you only have to call `getTestSym` once, but meh.

froydnj (approved):
Thank you!

PR #6656, opened 2023-01-10
Don't generate send responses for suggest-constant-type (fixes #6229)

### LINKED ISSUE #6229: Constant completion inside constant assignment doesn't work
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20module%20Namespace%3B%20end%0A%20%20Namespac%0A%20%20%23%20%20%20%20%20%20%20%5E%20completion%3A%20Namespace%0Aend%0A%0Aclass%20B%0A%20%20module%20Namespace%3B%20end%0A%20%20X%20%3D%20Namespac%0A%20%20%23%20%20%20%20%20%20%20%20%20%20%20%5E%20completion%3A%20Namespace%0Aend)

```ruby
# typed: true

class A
  module Namespace; end
  Namespac
  #       ^ completion: Namespace
end

class B
  module Namespace; end
  X = Namespac
  #           ^ completion: Namespace
end
```

#### Observed output

```
editor.rb:5: Unable to resolve constant Namespac https://srb.help/5002
     5 |  Namespac
          ^^^^^^^^
  Did you mean A::Namespace? Use `-a` to autocorrect
    editor.rb:5: Replace with A::Namespace
     5 |  Namespac
          ^^^^^^^^
    editor.rb:4: A::Namespace defined here
     4 |  module Namespace; end
          ^^^^^^^^^^^^^^^^

editor.rb:11: Unable to resolve constant Namespac https://srb.help/5002
    11 |  X = Namespac
              ^^^^^^^^
  Did you mean B::Namespace? Use `-a` to autocorrect
    editor.rb:11: Replace with B::Namespace
    11 |  X = Namespac
              ^^^^^^^^
    editor.rb:10: B::Namespace defined here
    10 |  module Namespace; end
          ^^^^^^^^^^^^^^^^
Errors: 2
```

#### Expected behavior

In both indicated places, the completion request should suggest `Namespace`.

However, currently only the location in `A` produces a result. The location in `B` does not include `Namespace` in the list of completion results.

- - -

neilparikh:
If we hit a `lspQueryMatch` for a send, but the method is `<suggest-constant-type>`, we will not generate a send response (similar to what we do for `===`).

### Motivation

`completion.cc` would use the `SendResponse` for the `<suggest-constant-type>` method, so the completion provided to the user was not useful.

### Test plan

See included automated tests.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1129,7 +1129,7 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                 // by desugar and redacting the `SendResponse` so LSP features work
                 // more like developers expect.
                 const bool isDesugarTripleEqSend = send.fun == core::Names::tripleEq() && send.funLoc.empty();
-                // Something like `X = Foo` will be rewritten to `X = Magic.<suggest-constant-type>(Foo)`
+                // Something like `X = Foo` will be rewritten to `X = Magic.<suggest-constant-type>(Foo)` if `Foo` fails to resolve.
                 // We don't want to report a send response for the <suggest-constant-type> call here.
                 const bool isSuggestConstantType = send.fun == core::Names::suggestConstantType();
                 if (lspQueryMatch && !isDesugarTripleEqSend && !isSuggestConstantType) {
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1129,10 +1129,12 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                 // by desugar and redacting the `SendResponse` so LSP features work
                 // more like developers expect.
                 const bool isDesugarTripleEqSend = send.fun == core::Names::tripleEq() && send.funLoc.empty();
-                // Something like `X = Foo` will be rewritten to `X = Magic.<suggest-constant-type>(Foo)` if `Foo` fails to resolve.
-                // We don't want to report a send response for the <suggest-constant-type> call here.
+                // Something like `X = Foo` will be rewritten to `X = Magic.<suggest-constant-type>(Foo)` if `Foo` fails
+                // to resolve. We don't want to report a send response for the <suggest-constant-type> call here.
                 const bool isSuggestConstantType = send.fun == core::Names::suggestConstantType();
-                if (lspQueryMatch && !isDesugarTripleEqSend && !isSuggestConstantType) {
+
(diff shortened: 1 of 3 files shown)


infer/environment.cc
```diff
@@ -1129,7 +1129,10 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                 // by desugar and redacting the `SendResponse` so LSP features work
                 // more like developers expect.
                 const bool isDesugarTripleEqSend = send.fun == core::Names::tripleEq() && send.funLoc.empty();
-                if (lspQueryMatch && !isDesugarTripleEqSend) {
+                // Something like `X = Foo` will be rewritten to `X = Magic.<suggest-constant-type>(Foo)`
```
froydnj, line 1132: My understanding is that this is only true if `Foo` failed to resolve:

https://github.com/sorbet/sorbet/blob/fd4f7df9f366f91976f9727a34fe5800227084de/resolver/resolver.cc#L2163-L2175

Can we rewrite the comment to make that clear?

froydnj (changes requested):
This looks good, but one comment bit that I think needs to be fixed up.

infer/environment.cc
```diff
@@ -1129,7 +1129,10 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
...
+                // We don't want to report a send response for the <suggest-constant-type> call here.
+                const bool isSuggestConstantType = send.fun == core::Names::suggestConstantType();
+                if (lspQueryMatch && !isDesugarTripleEqSend && !isSuggestConstantType) {
```
jez, line 1135: Minor nit: we might want to just have one variable, like `const bool ignoreSendForLSPQuery`. Instead of having to write

```cpp
                if (lspQueryMatch && !isDesugarTripleEqSend && !isSuggestConstantType) {
```

we could just write

```cpp
                if (lspQueryMatch && ignoreSendForLSPQuery) {
```

jez (approved):

jez:
Great find!

froydnj (approved):
(I had the idea that maybe the `<suggest-constant-type>` send itself needed a zero-length loc, which requires a little tweaking elsewhere, but seems to work OK:

https://github.com/sorbet/sorbet/compare/froydnj-constant-type-example?expand=1

I'm fine with what's currently here in the PR or using the zero-length-loc send.)

PR #6659, opened 2023-01-11
`Kernel#p` returns its argument(s) after printing

dirceu:
Some examples using IRB:

```ruby
> p "foo"
"foo"
=> "foo"

> p :hello
:hello
=> :hello

> p :hello, :world
:hello
:world
=> [:hello, :world]

> p Struct.new(:name)
#<Class:0x0000000126a33d98>
=> #<Class:0x0000000126a33d98>

> (p Struct.new(:name)).class
#<Class:0x00000001269355e0>
=> Class
```


rbi/core/kernel.rbi
```diff
@@ -1916,7 +1916,7 @@ module Kernel
     params(
         arg0: BasicObject,
     )
-    .returns(NilClass)
+    .returns(BasicObject)
```
vinistock, line 1919: Maybe we can use a `type_parameter` to return the same type as the argument.

dirceu: Ah interesting, something like this?

```ruby
  sig do
    type_parameters(:T)
      .params(arg0: T.type_parameter(:T))
      .returns(T.type_parameter(:T))
  end
```

vinistock: Yeah, exactly that.

rbi/core/kernel.rbi
```diff
@@ -1913,10 +1913,9 @@ module Kernel
   # #<S name="dave", state="TX">
   # ```
   sig do
-    params(
-        arg0: BasicObject,
-    )
-    .returns(NilClass)
+    type_parameters(:T)
+      .params(arg0: T.type_parameter(:T))
+      .returns(T.type_parameter(:T))
```
jez, line 1918: There is a TODO up on line 1889 relevant to this

iirc the last time we touched this, we left it like this because it was tricky to try to write a signature that accounted for the fact that when given more than one argument, `p` returns an `Array`.

jez:
Done in #9488

PR #6663, opened 2023-01-13
add tests for typing `Kernel#p`

froydnj:
### Motivation

Inspired by @jez's comments in #6659, add some tests for `Kernel#p` so the edge case of `Kernel#p` returning multiple arguments will be slightly more visible to anybody who comes along and tries to change the signature in `kernel.rbi` (or who is super-motivated and adds code in `core/types/calls.cc` for this).

### Test plan

See included automated tests.  There are some additional changes because I added tests at the end of the file initially and was confused why errors weren't being issued.  It took me a while to figure out that the infinite `loop` was marking all the code after it as dead, so I moved the existing code that *was* after the `loop` prior to it.  Fortunately, it still passes tests.

### DIFF
--- a/test/testdata/rbi/kernel.rb
+++ b/test/testdata/rbi/kernel.rb
@@ -65,9 +65,16 @@ T.assert_type!(obj.object_id, Integer)
 obj = T.let("foo", String)
 T.assert_type!(obj.itself, String)
 
-y = loop do
-end
-puts y # error: This code is unreachable
+
+# These types are deliberately wrong, because `Kernel#p` is difficult to type
+# in an RBI.  See the comments in kernel.rbi.
+p_result = Kernel.p 1
+T.reveal_type(p_result) # error: Revealed type: `NilClass`
+p_result = p "string"
+T.reveal_type(p_result) # error: Revealed type: `NilClass`
+# This should actually be typed as an array.
+p_result = p 1, 2
+T.reveal_type(p_result) # error: Revealed type: `NilClass`
 
 class CustomError < StandardError
   def initialize(cause, team)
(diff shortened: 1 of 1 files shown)


test/testdata/rbi/kernel.rb
```diff
@@ -65,9 +65,16 @@ def raise_test
...
-end
-puts y # error: This code is unreachable
+
+# These types are deliberately wrong, because `Kernel#p` is difficult to type
+# in an RBI.  See the comments in kernel.rbi.
+p_result = Kernel.p 1
+T.reveal_type(p_result) # error: Revealed type: `NilClass`
+p_result = p "string"
+T.reveal_type(p_result) # error: Revealed type: `NilClass`
+# This should actually be typed as an array.
```
neilparikh, line 75: Nit: Could you add a newline between each case? At first I thought this comment was referring to line 74 and was confused.

test/testdata/rbi/kernel.rb
```diff
@@ -65,9 +65,16 @@ def raise_test
 obj = T.let("foo", String)
 T.assert_type!(obj.itself, String)
 
-y = loop do
-end
-puts y # error: This code is unreachable
+
+# These types are deliberately wrong, because `Kernel#p` is difficult to type
+# in an RBI.  See the comments in kernel.rbi.
```
neilparikh, line 70: Nit: could you add a comment with the expected type for the integer and string case like the array case (just so it's obvious what `p` is actually returning)?

neilparikh (commented):
Some nits to make the test cases more obvious.

jez (approved):
lgtm modulo neil's comments

PR #6667, opened 2023-01-16
Support `.checked(:tests)` in `T::Struct`

jeffcarbs:
Support setting a `checked` level on a per-prop basis for T::Struct.

### Motivation

Follow-up PR from a slack thread (see https://sorbet-ruby.slack.com/archives/CFT8Y4909/p1672863188414029).

We've run into performance issues related to expensive recursive checks on `T::Struct` attributes, namely: https://github.com/sorbet/sorbet/blob/09bae4a8a0da2140f12443bc32a2725efc738eb7/gems/sorbet-runtime/lib/types/props/private/setter_factory.rb#L146-L153. There was a prior PR (#3990) to potentially address this but has since been abandoned, seems like due to the overhead it introduced. This PR tries adding similar functionality but avoiding the performance issues by moving the "checked" determination to load-time vs runtime.

### Test plan

See included automated tests. I followed the pattern from the "runtime levels" tests for sigs.

### Open questions

This is likely to break usage of:
- `T::Configuration.default_checked_level=`
- `T::Configuration.enable_checking_for_sigs_marked_checked_tests`

since those methods check that you haven't yet used the values in any sigs yet in this PR we introduce usage of `T::Private::RuntimeLevels.default_checked_level` and `T::Private::RuntimeLevels.check_tests?` during prop definition which will happen at load-time. In other worse, you couldn't try changing those values anytime after the first T::Struct is loaded. These work OK now since sigs are currently evaluated lazily so assuming you try changing those values "high" enough it's likely you haven't evaluated any sigs yet.


gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -333,6 +339,8 @@ def prop_defined(name, cls, rules={})
     end
     type_object = smart_coerce(type, enum: rules[:enum])
 
+    rules[:checked] ||= T::Private::RuntimeLevels.default_checked_level
```
jeffcarbs, line 342: I think using `T::Private::RuntimeLevels.default_checked_level` is technically the right thing to do here, but there are some backwards compatibility concerns:
- Currently this is really `:always` since runtime prop checks are always evaluated. This would change behavior for people who have a different `default_checked_level` set
- By evaluating `T::Private::RuntimeLevels.default_checked_level`, we make it so calls to `T::Configuration.default_checked_level = ...` will fail. Since this is happening at the time any T::Struct is loaded into memory, that will likely break things for folks. See #6668 for a path to fix this for folks.

I think given the above, especially before we have a mechanism to allow folks to workaround load-order issues, we should probably change this to:
```suggestion
    rules[:checked] ||= :always
```

And then switch back to `T::Private::RuntimeLevels.default_checked_level` in a dedicated PR.

jeffcarbs: NOTE: I wound up switching this back to `:always` to keep the behavior unchanged.

jeffcarbs:
cc @froydnj - This is ready for review. It shouldn't change any behavior for existing consumers until people start using `checked: level` on props.

gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
```diff
@@ -26,14 +26,20 @@ def self.build_setter_proc(klass, prop, rules)
...
+        requires_validation = check_level == :always || (check_level == :tests && T::Private::RuntimeLevels.check_tests?)
 
         # Use separate methods in order to ensure that we only close over necessary
         # variables
-        if !T::Props::Utils.need_nil_write_check?(rules) || has_explicit_nil_default
+        if !requires_validation
+          proc do |val|
+            instance_variable_set(accessor_key, val)
+          end
+        elsif !T::Props::Utils.need_nil_write_check?(rules) || has_explicit_nil_default
```
jez, line 42: I have a feeling that what we are going to want is that if this codepath would execute, we want to actually define the setter method at runtime with `attr_writer`. The Ruby VM special cases methods declared with `attr_writer` to run substantially faster than a method that calls `instance_variable_set`.

jez: We do something similar with `attr_reader` for a prop's getter method here:

https://github.com/sorbet/sorbet/blob/213e156fe200855e30781c1bab1e7fab5db88a72/gems/sorbet-runtime/lib/types/props/decorator.rb#L413-L414

jez (commented):
Do you think you could look into implementing this change using `attr_writer`?

gems/sorbet-runtime/lib/types/props/private/setter_factory.rb
jeffcarbs: That wouldn’t work for the immutable case, right? We could potentially `undef` the writers at the end of the initialize method, if it’s a similar speed up maybe it would still be faster. I can try it out and report back.

I think @djudd’s idea from slack would probably be the ideal in terms of perf:
> Speaking of which though, a more ambitious perf thing would be to start generating constructors as Ruby code and eval-ing them like we do with serde. There's some meaningful overhead associated with the current abstractions for small/hot objects. Depends how comfortable we are with the eval stuff though

jeffcarbs: Also thinking more about it now, if you use `attr_writer` you'll need to `@instance.send(:propname)` for each one in the initializer which is slower than `instance_variable_set(:@propname)`.

jez: Why does the constructor depend on the setter methods? They should be separate code paths.

jeffcarbs: The `setter_proc` is currently deeply intertwined with constructor of T::Struct, here are some examples:
- `T::Props::WeakConstructor` and `T::Props::Constructor` work by iterating over the props and `instance_exec` the setter proc, for example: https://github.com/sorbet/sorbet/blob/f8835e7eb5067c5e3f0da8db5dcb0e7867c0bacd/gems/sorbet-runtime/lib/types/props/weak_constructor.rb#L36-L41
- `T::Props::ApplyDefault.for` (which is used for props with defaults) also uses `instance_exec` to validate the defaults, for example: https://github.com/sorbet/sorbet/blob/f8835e7eb5067c5e3f0da8db5dcb0e7867c0bacd/gems/sorbet-runtime/lib/types/props/private/apply_default.rb#L160

jeffcarbs: @jez - I updated `T::Props::Decorator#define_getter_and_setter` to check if we need to typecheck the prop and use `attr_writer` if not. See the last commit and the updated PR description with the benchmarks. It basically gets us back to native PORO speed 🚀 

Let me know if you have thoughts on how to unwind the setter_proc from the constructors. If that winds up being a larger refactor I also wonder if we could consider merging this PR first and following up with that. Happy to help keep driving performance here since we rely heavily on T::Struct.

PR #6668, opened 2023-01-16
Support setting `checked_tests` and `default_checked_level` via environment variables

jeffcarbs:
There are two methods that affect runtime type checking:
- `T::Configuration.default_checked_level = ...`
- `T::Configuration.enable_checking_for_sigs_marked_checked_tests`

These methods blow up if you've already read the underlying values you're intending to change so they need to be called high enough up the stack. However, this presents a potentially brittle load-order issue, especially as more gems/libraries start to be typed with sorbet.

This PR enables setting those values via an environment variable so that they're guaranteed to be set to your desired values at the time sorbet is loaded

### Motivation

- Have a way to guarantee these values are set properly in a way that doesn't require any change to how sorbet-runtime is loaded.
- Provides a solution to an open issue described in #6667 

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/README.md
+++ b/gems/sorbet-runtime/README.md
@@ -6,6 +6,32 @@ Full docs are at:
 - https://sorbet.org/docs/runtime
 - https://sorbet.org/docs/tconfiguration
 
+## Environment variables
+
+There are a number of environment variables that `sorbet-runtime` reads from to change it's
+behavior:
+
+- `SORBET_RUNTIME_ENABLE_CHECKING_IN_TESTS`
+
+  Announces to Sorbet that we are currently in a test environment, so it
+  should treat any sigs which are marked `.checked(:tests)` as if they were
+  just a normal sig.
+
+  This can also be done by calling `T::Configuration.enable_checking_for_sigs_marked_checked_tests`
+  but the environment variable ensures this value gets set before any sigs
+  are evaluated.
+
+- `SORBET_RUNTIME_DEFAULT_CHECKED_LEVEL`
+
+  Configure the default checked level for a sig with no explicit `.checked`
+  builder. When unset, the default checked level is `:always`.
+
+  This can also be done by calling `T::Configuration.default_checked_level = ...`
+  but the environment variable ensures this value gets set before any sigs
+  are evaluated.
+
+## Testing
+
 To run the tests for sorbet-runtime locally:
 
 ```
--- a/gems/sorbet-runtime/README.md
+++ b/gems/sorbet-runtime/README.md
@@ -8,7 +8,7 @@ Full docs are at:
 
 ## Environment variables
 
-There are a number of environment variables that `sorbet-runtime` reads from to change it's
+There are a number of environment variables that `sorbet-runtime` reads from to change its
 behavior:
 
 - `SORBET_RUNTIME_ENABLE_CHECKING_IN_TESTS`
--- a/gems/sorbet-runtime/lib/types/private/runtime_levels.rb
+++ b/gems/sorbet-runtime/lib/types/private/runtime_levels.rb
@@ -53,6 +53,10 @@ module T::Private::RuntimeLevels
     if @has_read_default_checked_level
       raise "Set the default checked level earlier. There are already some methods whose sig blocks have evaluated which would not be affected by the new default."
     end
+    if !LEVELS.include?(default_checked_level)
+      raise "Invalid `checked` level '#{default_checked_level}'. Use one of: #{LEVELS}."
+    end
+
     @default_checked_level = default_checked_level
   end
 
--- a/gems/sorbet-runtime/lib/types/private/runtime_levels.rb
+++ b/gems/sorbet-runtime/lib/types/private/runtime_levels.rb
@@ -63,4 +63,19 @@ module T::Private::RuntimeLevels
   def self._toggle_checking_tests(checked)
     @check_tests = checked
   end
+
+  private_class_method def self.set_enable_checking_in_tests_from_environment
+    if ENV['SORBET_RUNTIME_ENABLE_CHECKING_IN_TESTS']
+      enable_checking_in_tests
+    end
+  end
+  set_enable_checking_in_tests_from_environment
+
+  private_class_method def self.set_default_checked_level_from_environment
+    level = ENV['SORBET_RUNTIME_DEFAULT_CHECKED_LEVEL']
+    if level
+      self.default_checked_level = level.to_sym
+    end
+  end
+  set_default_checked_level_from_environment
 end
--- a/gems/sorbet-runtime/test/types/runtime_levels.rb
+++ b/gems/sorbet-runtime/test/types/runtime_levels.rb
@@ -1,0 +1,86 @@
+# frozen_string_literal: true
+require_relative '../test_helper'
+
+module Opus::Types::Test
+  class RuntimeLevelsTest < Critic::Unit::UnitTest
+    describe 'set_enable_checking_in_tests_from_environment' do
+      before do
+        @orig_wrapped_tests_with_validation = T::Private::RuntimeLevels.instance_variable_get(:@wrapped_tests_with_validation)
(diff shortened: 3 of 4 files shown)


gems/sorbet-runtime/test/types/runtime_levels.rb
```diff
@@ -0,0 +1,86 @@
...
+        T::Private::RuntimeLevels.instance_variable_set(:@default_checked_level, @orig_default_checked_level)
+        T::Private::RuntimeLevels.instance_variable_set(:@has_read_default_checked_level, @orig_has_read_default_checked_level)
+      end
+
+      it 'does not change default_typed_level' do
+        # Reaching into a private method for testing purposes
+        T::Private::RuntimeLevels.send(:set_default_checked_level_from_environment)
+
+        assert_equal(T::Private::RuntimeLevels.default_checked_level, :always)
+      end
```
froydnj, line 65: Same thing here.

froydnj (approved):
Sorry about the delay in reviewing this; I had things open in a tab, but failed to click everything through.

I think this is a totally reasonable feature request.

gems/sorbet-runtime/test/types/runtime_levels.rb
```diff
@@ -0,0 +1,86 @@
...
+        T::Private::RuntimeLevels.instance_variable_set(:@check_tests, @orig_check_tests)
+        T::Private::RuntimeLevels.instance_variable_set(:@wrapped_tests_with_validation, @orig_wrapped_tests_with_validation)
+      end
+
+      it 'does not change check_tests' do
+        # Reaching into a private method for testing purposes
+        T::Private::RuntimeLevels.send(:set_enable_checking_in_tests_from_environment)
+
+        assert_equal(T::Private::RuntimeLevels.check_tests?, false)
+      end
```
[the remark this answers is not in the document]
jeffcarbs: This actually shouldn't need to update that. `set_enable_checking_in_tests_from_environment` only reads from the environment variable, doesn't mutate it, and then calls `enable_checking_in_tests` if the environment variable is set.

gems/sorbet-runtime/test/types/runtime_levels.rb
jeffcarbs: This actually shouldn't need to update that. `set_default_checked_level_from_environment ` only reads from the environment variable, doesn't mutate it, and then calls `default_checked_level=` if the environment variable is set.

gems/sorbet-runtime/test/types/runtime_levels.rb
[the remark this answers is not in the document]
froydnj:
```
froydnj@pyro:~/src/sorbet/gems/sorbet-runtime$ env SORBET_RUNTIME_ENABLE_CHECKING_IN_TESTS=1 PATH=$PATH:/home/froydnj/srb-ruby/bin /home/froydnj/srb-ruby/bin/bundle exec rake test
/home/froydnj/src/sorbet/gems/sorbet-runtime/test/types/method_validation.rb:573: warning: key :_ is duplicated and overwritten on line 573
Run options: --seed 56287

# Running:

..................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................F....S........................................................................................................................................

Finished in 0.456266s, 1858.5668 runs/s, 5766.3787 assertions/s.

  1) Failure:
set_enable_checking_in_tests_from_environment#test_0001_does not change check_tests [/home/froydnj/src/sorbet/gems/sorbet-runtime/test/types/runtime_levels.rb:24]:
Expected: true
  Actual: false

848 runs, 2631 assertions, 1 failures, 0 errors, 1 skips

```

gems/sorbet-runtime/test/types/runtime_levels.rb
froydnj:
```
froydnj@pyro:~/src/sorbet/gems/sorbet-runtime$ env SORBET_RUNTIME_DEFAULT_CHECKED_LEVEL=never PATH=$PATH:/home/froydnj/srb-ruby/bin /home/froydnj/srb-ruby/bin/bundle exec rake test
/home/froydnj/src/sorbet/gems/sorbet-runtime/test/types/method_validation.rb:573: warning: key :_ is duplicated and overwritten on line 573
Run options: --seed 48106

# Running:

....................................................................................................................................................................................................................................F.......................................................................................................................................................................................................................................................F...............................................................................................................................................................................................................................................................S...................................................................................................................

Finished in 0.455613s, 1861.2274 runs/s, 5774.6337 assertions/s.

  1) Failure:
set_default_checked_level_from_environment#test_0001_does not change default_typed_level [/home/froydnj/src/sorbet/gems/sorbet-runtime/test/types/runtime_levels.rb:64]:
Expected: :never
  Actual: :always

  2) Failure:
Opus::Types::Test::TopLevel#test_0001_works with top level sigs [/home/froydnj/src/sorbet/gems/sorbet-runtime/test/types/top_level.rb:9]:
--- expected
+++ actual
@@ -1,5 +1,5 @@
 "called main#foo
-Expected type Integer, got type String with value \"nope\"
+foo() did not raise a TypeError!
 called main.bar
-Expected type Symbol, got type Float with value 0.0
+bar() did not raise a TypeError!
 "


848 runs, 2631 assertions, 2 failures, 0 errors, 1 skips

```

It's possible that both of these failures (and the one above) should be fixed at a higher level than the individual tests.

gems/sorbet-runtime/test/types/runtime_levels.rb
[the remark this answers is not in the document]
jeffcarbs: Oh I see what you're saying, I had misunderstood. You're totally right, I just pushed up an update to the tests to work the same even if those environment variables are set.

froydnj:
https://go/builds/bui_NEMv8zDoyVnRXE for the Stripe-internal build with this PR applied.

gems/sorbet-runtime/lib/types/private/runtime_levels.rb
```diff
@@ -53,10 +53,29 @@ def self.default_checked_level=(default_checked_level)
...
+  end
+  set_enable_checking_in_tests_from_environment
+
+  private_class_method def self.set_default_checked_level_from_environment
+    level = ENV['SORBET_RUNTIME_DEFAULT_CHECKED_LEVEL']
+    if level
+      self.default_checked_level = level.to_sym
+    end
+  end
+  set_default_checked_level_from_environment
```
jez, line 80: This strikes me as a roundabout way to set a default.

Why not handle this earlier on line 24, where the default is actually set?

```ruby
@default_checked_level = :always
```

changing this to something like

```ruby
@default_checked_level =
  if (level = ENV['SORBET_RUNTIME_DEFAULT_CHECKED_LEVEL'])
    raise "..." unless LEVELS.include?(level.to_sym)

    level.to_sym
  else
    :always
  end
```

gems/sorbet-runtime/README.md
```diff
@@ -6,6 +6,32 @@ Full docs are at:
 - https://sorbet.org/docs/runtime
 - https://sorbet.org/docs/tconfiguration
 
+## Environment variables
```
jez, line 9: Unlike with the `sorbet` gem, we have managed to keep all the docs for sorbet-runtime on sorbet.org to this point. I would love if we could move the docs for these environment variables to the docs in tconfiguration.md

jeffcarbs: Opened #6679 to add docs.

gems/sorbet-runtime/lib/types/private/runtime_levels.rb
jeffcarbs: Yea, it's definitely a little roundabout. I made the decision for a few reasons:
- Consolidate the `LEVELS` validation to a single place - the `default_checked_level=` method.
- Make it possible to write a test for this functionality changes.
- I initially implemented like you suggested but then that `level` local variable just exists at the class level which feels weird. Once it wasn't a simple one-liner I felt it made more sense in a method.

Issue #4233, opened 2021-05-24
Feature suggestion: definition order errors within a single file


aisamanra:
### Problem

While Sorbet in general assumes the use of an autoloader and therefore doesn't care about definition ordering, this really only works across multiple files. A code snippet within a single file that looks like

```ruby
MyX = T.let(X.new, X)
class X; end
```

will fail at runtime because `X` is used before it's defined. [Sorbet currently does not flag an error here](https://sorbet.run/#%23%20typed%3A%20true%0AMyX%20%3D%20T.let(X.new%2C%20X)%0Aclass%20X%3B%20end%0A). My suggestion is that for _the specific case of definition ordering errors within a single file_ Sorbet can report an error. This error also wouldn't happen outside of static scope or within blocks (e.g. it's fine to use `X` in a `sig` or within a method.)

```ruby
sig {returns(X)} # this is okay
def my_x; X.new; end # this is okay

MyX = T.let(X.new, X) # error: the class X is used here before its definition later in this file
class X; end
```

Note that I'm not asking Sorbet to become aware of code-loading (e.g. `require`s) or specifics of autoloader semantics more generally: different implementations of autoloaders might have different semantics and making Sorbet compatible with any and all of them would likely be significantly outside the scope of Sorbet. However, a snippet of code like

```ruby
puts X
X = 5
```

is pretty universally going to fail at runtime, which is why I think it's worth proposing this as an error which can be caught by Sorbet.

### Proposed solution

I believe this could be computed efficiently in the `definition_validator` class by adding a check to `ValidateWalk::postTransformConstantLit` that's checked whenever the resolved constant's symbol is defined in the same file as the literal, is defined at a later loc than the literal, and the literal is not within a block or method.

jez:
> adding a check to `ValidateWalk::postTransformConstantLit` that's checked whenever the resolved constant's symbol is defined in the same file as the literal

Inside a treewalk, you do not know whether you're inside a block, or at the top level, or what. I don't think the implementation will play out exactly as you mentioned.

Scoped as written, I'd be willing to entertain potential implementations. I do not think any larger scope is possible or desirable.

If possible, I'd rather have this error sooner than later (I would rather that this error show up alongside other constant resolution errors in namer / resolver, rather than all the way in definition validator). But again, implementation considerations might prevent that.

aadi-stripe:
This PR https://github.com/sorbet/sorbet/pull/6671 would be a possible implementation of this feature.

[...]

PR #6671, opened 2023-01-19
Catch out-of-order constant references

aadi-stripe:
Adds out-of-order reference checking for constants.

### Motivation

As detailed in issue https://github.com/sorbet/sorbet/issues/4233, it is useful in some contexts to catch the naive case of out-of-order constant access.

At Stripe, we are moving to a more explicit autoloading mechanism, which exposes these issues at runtime. In order to catch them statically, I am introducing this feature into the constant resolution pipeline.

Note that currently my plan is to only expose this in `--stripe-mode` as this feature is most usable for our current autoloader migration to Zeitwerk. It may be worthwhile opening this up in future. 

### Test plan

See included automated tests. 

Tested on Stripe codebase. However, we can't land this yet until we fix existing issues in the codebase.


core/Symbols.cc
```diff
@@ -2523,7 +2523,6 @@ namespace {
 void addLocInternal(const core::GlobalState &gs, core::Loc loc, core::Loc mainLoc, InlinedVector<Loc, 2> &locs) {
     for (auto &existing : locs) {
         if (existing.file() == loc.file()) {
-            existing = loc;
```
aadi-stripe, line 2526: This is really the crux here. In order for this to work, we have to always assign the *first* loc in a file, not the last one. I think this might break other stuff.

aadi-stripe:

→ https://go/builds/bui_NCqFIbtsGWsNEn
→ https://go/builds/bui_NCqFWwCui7ETPx

jez:
I believe this should be linked with https://github.com/sorbet/sorbet/issues/4233

Fixes #4233

(lmk if that's not the case)

resolver/resolver.cc
```diff
@@ -1317,16 +1374,60 @@ class ResolveConstantsWalk {
...
     }
 
+    void preTransformClassDef(core::Context ctx, ast::ExpressionPtr &tree) {
+        auto &original = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+        auto sym = original.symbol;
+
+        // Populate local first definitions table for out-of-order reference checking.
+        // This needs to happen *iff* the symbol is:
+        //   a). defined only in this file and nowhere else
+        //   b). defined multiple times in this file
```
froydnj, line 1416: This condition for _populating_ the table seems like it's hard to satisfy, given that we don't store multiple locs per file.  So how do we tell that there are multiple definitions?

And in any event, the error added in this PR isn't checking for multiple definitions, it's checking for the relationship between a definition and a reference, which is not the same thing.  It seems like the comment here (and the similar one for assignments, below) ought to be adjusted?

resolver/resolver.cc
```diff
@@ -1317,16 +1374,60 @@ class ResolveConstantsWalk {
...
+        //   c). this particular classDef is the first occurrence of the symbol in the file
+        if (loadScopeDepth_ == 0 && sym.isOnlyDefinedInFile(ctx.state, ctx.file)) {
```
froydnj, line 1418: Do things go awry if we skip checking this in RBI files?  Do things get any faster if we skip checking this in RBI files?

resolver/resolver.cc
```diff
@@ -1452,6 +1553,21 @@ class ResolveConstantsWalk {
             return;
         }
 
+        // Populate local first definitions table for out-of-order reference checking.
+        // This needs to happen *iff* the symbol is:
+        //   a). defined only in this file and nowhere else
+        //   b). defined multiple times in this file
+        //   c). this particular assign is the first occurrence of the symbol in the file
+        if (loadScopeDepth_ == 0 && id->symbol.isOnlyDefinedInFile(ctx.state, ctx.file)) {
```
froydnj, line 1561: Same question here about RBI files.  And, actually, since this code and the above are so similar and require the same amount of commentary, maybe there should be a single function `(ctx, sym, declLoc)` that handles all this logic?

resolver/resolver.cc
```diff
@@ -129,13 +129,19 @@ class ResolveConstantsWalk {
 
         Nesting(shared_ptr<Nesting> parent, core::SymbolRef scope) : parent(std::move(parent)), scope(scope) {}
     };
+
     shared_ptr<Nesting> nesting_;
+    shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> localFirstDefOfManySymTable;
```
froydnj, line 134: I was initially confused by the "local" here; I guess the "local" is trying to connote "local to a given file"?  I think that's a little hard to parse out, given that there's no reference to files in this data structure.

The name of this symbol is a mouthful.  Maybe just `firstDefinitionLoc` or something, since I think that's what you're trying to capture...and that differs from the loc stored in the symbol table because the canonical loc in the symbol table is the latest one (?).  A comment might help describe why we're tracking this separately from symbol table locs.

I think we could do this without `shared_ptr`, but managing the pointers gets a little more complicated at that point, and I don't think that it's worth it.

froydnj (changes requested):
I have enough questions and suggestions that I think this needs another round.

Have you tried seeing what kind of performance implications this has?

resolver/resolver.cc
aadi-stripe: We know that there are multiple by seeing if the current loc is not the same as the canonical loc. That can only happen if there are multiple definitions, afaik.

While this PR isn't directly checking for multiple defs, we need to know the first def (if there are multiple) so we can adjust for pre-declarations. Can change the comment to make that clear.

resolver/resolver.cc
aadi-stripe: I think we should skip, I can add that in.

aadi-stripe:
> I have enough questions and suggestions that I think this needs another round.
> 
> Have you tried seeing what kind of performance implications this has?

Haven't dived deep into perf, can take look at that.

resolver/resolver.cc
```diff
@@ -1317,16 +1374,60 @@ class ResolveConstantsWalk {
...
+    void preTransformBlock(core::Context ctx, ast::ExpressionPtr &tree) {
+        if (loadScopeDepth_ >= 0) {
+            loadScopeDepth_++;
+        }
+    }
+
+    void postTransformBlock(core::Context ctx, ast::ExpressionPtr &tree) {
+        if (loadScopeDepth_ > 0) {
+            loadScopeDepth_--;
+        }
```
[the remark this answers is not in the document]
aadi-stripe: Yeah these can be `ENFORCE`s

resolver/resolver.cc
```diff
@@ -1317,16 +1374,60 @@ class ResolveConstantsWalk {
...
+        //   b). defined multiple times in this file
+        //   c). this particular classDef is the first occurrence of the symbol in the file
+        if (loadScopeDepth_ == 0 && sym.isOnlyDefinedInFile(ctx.state, ctx.file)) {
+            auto defLoc = sym.data(ctx)->loc().offsets();
+            auto declLoc = original.declLoc;
+
+            auto localSymMap = localFirstDefOfManySymTable.get();
+            if (defLoc.beginPos() >= declLoc.endPos() && localSymMap->find(sym) == localSymMap->end()) {
+                localSymMap->emplace(sym, declLoc);
+            }
```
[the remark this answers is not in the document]
aadi-stripe: Ack

aadi-stripe:

→ https://go/builds/bui_NFTr7923rUQ7lC
→ https://go/builds/bui_NFTru6qGkmfVw6

aadi-stripe:
@froydnj Thanks for the feedback. Implemented some of your recommended changes.

I'll do a perf analysis on the Stripe codebase next week.

aadi-stripe:

→ https://go/builds/bui_NGIGjQsssaxxbc
→ https://go/builds/bui_NGIGEl3kJP8buL

aadi-stripe:
@froydnj 

Adding perf traces on Stripe codebase.

BEFORE
Screen Shot 2023-01-29 at 5 19 54 PM

AFTER
Screen Shot 2023-01-29 at 5 16 24 PM

No significant change.

resolver/resolver.cc
```diff
@@ -295,10 +301,52 @@ class ResolveConstantsWalk {
...
-    static core::SymbolRef resolveConstant(core::Context ctx, const shared_ptr<Nesting> &nesting,
-                                           const ast::UnresolvedConstantLit &c, bool &resolutionFailed) {
+    // Find load-time out-of-order references.
+    // Algorithm:
+    // if resolved symbol is only defined in the current file, and
+    //    the definition is after the reference loc,
+    // then, report an error.
+    static void
+    checkReferenceOrder(core::Context ctx, core::SymbolRef resolutionResult, const ast::UnresolvedConstantLit &c,
+                        const shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> firstDefinitionLocs) {
```
froydnj, line 311: I vote for either taking this parameter by reference, or changing it to `const UnorderedMap<core::SymbolRef, core::LocOffsets> &` and changing the callers.  Your choice.

resolver/resolver.cc
```diff
@@ -295,10 +301,52 @@ class ResolveConstantsWalk {
...
+                    e.addErrorLine(ctx.locAt(defLoc), "Defined here");
+                }
+            }
+        }
+    }
+
+    static core::SymbolRef
+    resolveConstant(core::Context ctx, const shared_ptr<Nesting> &nesting, const ast::UnresolvedConstantLit &c,
+                    bool &resolutionFailed, bool loadTimeScope,
+                    const shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> firstDefinitionLocs) {
```
froydnj, line 342: I vote for either taking this parameter by reference, or changing it to `const UnorderedMap<core::SymbolRef, core::LocOffsets> &` and changing the callers.

Slight preference for just taking this type by reference so we don't incur memory traffic in the `!loadTimeScope` case (which I guess should be uncommon anyway?).

resolver/resolver.cc
```diff
@@ -1317,16 +1374,58 @@ class ResolveConstantsWalk {
...
+        //   c). this particular classDef is the first occurrence of the symbol in the file
+        if (!ctx.file.data(ctx).isRBI() && loadScopeDepth_ == 0 && sym.isOnlyDefinedInFile(ctx.state, ctx.file)) {
+            auto defLoc = sym.data(ctx)->loc().offsets();
+            auto declLoc = original.declLoc;
+
+            if (defLoc.beginPos() >= declLoc.endPos() && firstDefinitionLocs->find(sym) == firstDefinitionLocs->end()) {
+                // When this condition is met, effectively it means the file has multiple definitions of the symbol
+                // and this is the first one. We need to use the first def for out-of-order checking, since any
+                // reference *before* the first def will be out of order.
+                firstDefinitionLocs->insert(std::make_pair(sym, declLoc));
```
froydnj, line 1422: Given that you're using `insert` here, you can drop the check for `firstDefinitionLocs->find` above because `insert` will only create an entry if none exists:

https://en.cppreference.com/w/cpp/container/unordered_map/insert
https://github.com/abseil/abseil-cpp/blob/0c3df2f5a7d3918b9e5762cd8f143a6004b76cda/absl/container/flat_hash_map.h#L246-L266

resolver/resolver.cc
```diff
@@ -1452,6 +1551,25 @@ class ResolveConstantsWalk {
...
+            id->symbol.isOnlyDefinedInFile(ctx.state, ctx.file)) {
+            auto defLoc = id->symbol.loc(ctx).offsets();
+            auto declLoc = asgn.loc;
+
+            if (defLoc.beginPos() >= declLoc.endPos() &&
+                firstDefinitionLocs->find(id->symbol) == firstDefinitionLocs->end()) {
+                // When this condition is met, effectively it means the file has multiple definitions of the symbol
+                // and this is the first one. We need to use the first def for out-of-order checking, since any
+                // reference *before* the first def will be out of order.
+                firstDefinitionLocs->insert(std::make_pair(id->symbol, declLoc));
```
froydnj, line 1569: Same comment applies to this use of `insert`.

resolver/resolver.cc
froydnj: OK, I think I was just confused when I wrote this comment, thinking that the conditions listed here applied to the conditional just below, when in reality, two of these conditions apply to the inner condition a few lines later.

resolver/resolver.cc
```diff
@@ -1317,16 +1374,58 @@ class ResolveConstantsWalk {
...
+        // This happens *iff* the symbol is:
+        //   a). defined only in this file and nowhere else
+        //   b). defined multiple times in this file
+        //   c). this particular classDef is the first occurrence of the symbol in the file
+        if (!ctx.file.data(ctx).isRBI() && loadScopeDepth_ == 0 && sym.isOnlyDefinedInFile(ctx.state, ctx.file)) {
+            auto defLoc = sym.data(ctx)->loc().offsets();
+            auto declLoc = original.declLoc;
+
+            if (defLoc.beginPos() >= declLoc.endPos() && firstDefinitionLocs->find(sym) == firstDefinitionLocs->end()) {
+                // When this condition is met, effectively it means the file has multiple definitions of the symbol
```
froydnj, line 1419: We should really add some documentation on how `loc` and `declLoc` in `ClassDef` differ, because I always forgot, and as a result, I can't convince myself that this doesn't fire for things that have just single definitions in one file.  I don't think that matters (we still want out-of-order checking in that case?), but just thinking out loud.

froydnj (changes requested):
Thanks for tweaking this according to feedback.  I think a few things were missed and/or misinterpreted the first time around.

resolver/resolver.cc
aadi-stripe: Will add this comment:

We want `declLoc` here because we want the specific loc of the current def, not the canonical `loc` of the symbol. In case of a single definition, the values will be the same and the check won't fire.

froydnj (approved):
Thank you!

test/testdata/resolver/basic_constant_out_of_order.rb
```diff
@@ -0,0 +1,46 @@
...
+    X # this is ok
+  end
+
+  class X; end
+
+  B = X # this is ok
+
+  class Bar
+    Foo::Y
+  # ^^^^^^ error: `Foo::Y` referenced before it is defined
```
jez, line 28:
```suggestion
    p(Foo::Y)
    # ^^^^^^ error: `Foo::Y` referenced before it is defined
```

this is another minor thing but

- using `p` makes it look like the constant isn't unused (ALE in vim by default runs `ruby -w` which prints `warning: possibly useless use of a constant in void context`)
- using `p(` also gives us some extra characters to allow the `^^^` to line up nicely

resolver/resolver.cc
```diff
@@ -1452,6 +1551,24 @@ class ResolveConstantsWalk {
...
+        if (!ctx.file.data(ctx).isRBI() && loadScopeDepth_ == 0 &&
+            id->symbol.isOnlyDefinedInFile(ctx.state, ctx.file)) {
+            auto defLoc = id->symbol.loc(ctx).offsets();
+            auto declLoc = asgn.loc;
+
+            if (defLoc.beginPos() >= declLoc.endPos()) {
+                // When this condition is met, effectively it means the file has multiple definitions of the symbol
+                // and this is the first one. We need to use the first def for out-of-order checking, since any
+                // reference *before* the first def will be out of order.
+                firstDefinitionLocs->insert(std::make_pair(id->symbol, declLoc));
```
jez, line 1568: Can we ENFORCE that there's nothing already there?

test/testdata/resolver/basic_constant_out_of_order.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  Y = 1
+
+  class Bar
+    Foo::Y # this is ok
+  end
+
+  Y = 2
```
jez, line 45: it might be worthwhile to add a case for when a constant is defined three times, because the middle def will neither be the first loc in the file, nor the canonical loc on the symbol. (i can imagine a future change that tries to refactor this logic and fails by way of accidentally overwriting something with the middle loc)

jez:
> Adding perf traces on Stripe codebase.

Did you also look at how this affects max_rss?

jez:
Now that this feature is not tied to `--stripe-mode`, it should be easier to get a list of all the type errors. Do you have such a build? I would love to see it.

resolver/resolver.cc
```diff
@@ -295,10 +301,52 @@ class ResolveConstantsWalk {
...
+    // Find load-time out-of-order references.
+    // Algorithm:
+    // if resolved symbol is only defined in the current file, and
+    //    the definition is after the reference loc,
+    // then, report an error.
+    static void
+    checkReferenceOrder(core::Context ctx, core::SymbolRef resolutionResult, const ast::UnresolvedConstantLit &c,
+                        const shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> &firstDefinitionLocs) {
+        if (!ctx.file.data(ctx).isRBI() && resolutionResult.exists() &&
+            resolutionResult.isOnlyDefinedInFile(ctx.state, ctx.file)) {
```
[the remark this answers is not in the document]
aadi-stripe: Ah, it's not desired. We need to check if the other file is an RBI

resolver/resolver.cc
```diff
@@ -295,10 +301,52 @@ class ResolveConstantsWalk {
...
+
+            if (it != firstDefinitionLocs.get()->end()) {
+                // Use the loc from the local first defs table, if it exists.
+                // When the symbol is declared multiple times in the same file, we want to ensure that
+                // we compare against the *first* definition.
+                defLoc = it->second;
+            } else {
+                // If the symbol isn't in the local first defs table, we fall back to the
+                // canonical loc.
+                defLoc = resolutionResult.loc(ctx).offsets();
```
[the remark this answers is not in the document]
aadi-stripe: Because I only store in the table if the loc of the ClassDef is different than the canonical loc. Otherwise it's unnecessary.

resolver/resolver.cc
aadi-stripe: Yes, b) is only for perf.

aadi-stripe:
> > Adding perf traces on Stripe codebase.
> 
> Did you also look at how this affects max_rss?

Have not checked, but I can look into that.

resolver/resolver.cc
aadi-stripe: Sure

test/testdata/resolver/basic_constant_out_of_order.rb
aadi-stripe: 👍

resolver/resolver.cc
```diff
@@ -129,13 +129,19 @@ class ResolveConstantsWalk {
 
         Nesting(shared_ptr<Nesting> parent, core::SymbolRef scope) : parent(std::move(parent)), scope(scope) {}
     };
+
     shared_ptr<Nesting> nesting_;
+    shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> firstDefinitionLocs;
```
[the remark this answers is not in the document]
aadi-stripe: Yes, that's why I used a shared_ptr. So definitely worth a callout.

aadi-stripe:
> Now that this feature is not tied to `--stripe-mode`, it should be easier to get a list of all the type errors. Do you have such a build? I would love to see it.

I can link that tomorrow.

aadi-stripe:

→ https://go/builds/bui_NH75Tfbct6k2e1
→ https://go/builds/bui_NH75MSZ2RPPmGb

aadi-stripe:

→ https://go/builds/bui_NH7pl2ozqNo79Y
→ https://go/builds/bui_NH7psv4fFvTTos

aadi-stripe:
RSS: Before
Screen Shot 2023-01-31 at 9 38 33 PM

RSS: After
Screen Shot 2023-01-31 at 9 37 57 PM

Note that even without the flag we should get an accurate measurement of RSS changes, because the flag only suppresses the error reporting, it doesn't prevent the check from happening.

aadi-stripe:
Build with the errors turned on--
https://go/builds/bui_NH80Lzdn4h44rt

max-RSS with the errors turned on--
Screen Shot 2023-01-31 at 10 32 30 PM

aadi-stripe:
@jez Made the requested changes, also posted a failing build and RSS numbers.

aadi-stripe:
Posting Timing/RSS results here
https://go/gist/aadi/1faa29398b1136f382c1f6eae2f9dbe8

resolver/resolver.cc
```diff
@@ -295,10 +310,52 @@ class ResolveConstantsWalk {
...
+            const auto it = firstDefinitionLocs.get()->find(resolutionResult);
+
+            if (it != firstDefinitionLocs.get()->end()) {
+                // Use the loc from the local first defs table, if it exists.
+                // When the symbol is declared multiple times in the same file, we want to ensure that
+                // we compare against the *first* definition.
+                defLoc = it->second;
+            } else {
+                // If the symbol isn't in the local first defs table, we fall back to the
+                // canonical loc.
```
jez, line 333:
```suggestion
                // If the symbol isn't in the local first defs table, that means its loc
                // in the symbol table is the same as its first definition in this file. 
                // (We skip storing redundant information in firstDefinitionLocs to save memory)
```

resolver/resolver.cc
```diff
@@ -1317,16 +1383,63 @@ class ResolveConstantsWalk {
...
 
+    void preTransformClassDef(core::Context ctx, ast::ExpressionPtr &tree) {
+        auto &original = ast::cast_tree_nonnull<ast::ClassDef>(tree);
+        auto sym = original.symbol;
+
+        // Populate local first definitions table for out-of-order reference checking.
+        // This happens *iff* the symbol is:
+        //   a). defined only in this file and nowhere else
+        //   b). defined multiple times in this file
+        //   c). this particular classDef is the first definition of the symbol in the file
```
jez, line 1422:
```suggestion
        // We only do this when we know we're going to need to consult the first definition loc and
        // we know the information in the symbol table is insufficient. This avoids reundant memory
        // storage overhead.
        //
        // In particular, `firstDefinitionLocs` does not store anything if this symbol is defined in
        // more than one non-RBI file or if it's only defined once in this file.
        // Otherwise, it stores the loc of the first definition of the symbol in this file.
```

resolver/resolver.cc
```diff
@@ -1452,6 +1565,29 @@ class ResolveConstantsWalk {
             return;
         }
 
+        // Populate local first definitions table for out-of-order reference checking.
+        // This needs to happen *iff* the symbol is:
+        //   a). defined only in this file and nowhere else
+        //   b). defined multiple times in this file
+        //   c). this particular assign is the first definition of the symbol in the file
```
jez, line 1572:
```suggestion
        // We only do this when we know we're going to need to consult the first definition loc and
        // we know the information in the symbol table is insufficient. This avoids reundant memory
        // storage overhead.
        //
        // In particular, `firstDefinitionLocs` does not store anything if this symbol is defined in
        // more than one non-RBI file or if it's only defined once in this file.
        // Otherwise, it stores the loc of the first definition of the symbol in this file.
```

website/docs/error-reference.md
```diff
@@ -1803,6 +1803,63 @@ def foo; [0]; end
...
+```
+
+Generally, Sorbet is not opinionated about definition-reference ordering. It
+assumes the presence of an autoloader which makes constants available at runtime
+when they are referenced (assuming they are defined somewhere).
+
+However, it is helpful in some basic contexts (like class-loading) to catch this
+issue, regardless of the presence of an autoloader. Please note that this error
+will not fire for all cases. There are some conditions which have to be met for
+the error to fire:
```
jez, line 1830:
```suggestion
Generally, Sorbet is not opinionated about definition-reference ordering. It
assumes files are required in the correct order or at the correct times to
ensure that definitions are available before they're referenced.

However, if a constant is defined in a single file, Sorbet can detect when it's
been referenced in that file ahead of its definition (because in the single-file
case, it doesn't matter whether or in what order any require statements
happen). There are some limitations:
```

sorbet does not require an autoloader. rephrasing in terms of require statements.

website/docs/error-reference.md
```diff
@@ -1803,6 +1803,63 @@ def foo; [0]; end
...
+
+### Load-time scope must be established definitively
+
+Sorbet has to prove definitively that a given constant is accessed out-of-order
+at load time. It cannot track accesses across function calls or blocks. Meaning
```
jez, line 1835:
```suggestion
at load time. It cannot track accesses across function calls or blocks, meaning
```

website/docs/error-reference.md
```diff
@@ -1803,6 +1803,63 @@ def foo; [0]; end
...
+  end
+
+  class X; end
+```
+
+### Symbols have to be guaranteed to exist only in one file
+
+In the above example, if `Foo::X` is also declared in another file, the error
+will not fire. In such cases, the constant may get autoloaded, so we cannot
+prove that the behavior is erroneous.
```
jez, line 1856:
```suggestion
will not fire. In such cases, the other file that defines `X` may get required first, so Sorbet cannot
prove that there will be a problem referencing `X` in this file.
```

website/docs/error-reference.md
```diff
@@ -1803,6 +1803,63 @@ def foo; [0]; end
 For more information, see
 [Arrays, Hashes, and Generics in the Standard Library](stdlib-generics.md).
 
+## 5027
```
jez, line 1806:
```suggestion
## 5027

> This error is opt-in, behind the `--check-out-of-order-constant-references`
> flag.
>
> Sorbet does not check this by default because certain codebases make clever
> usage of Ruby's `autoload` mechanism to allow all constants to be referenced
> before their definitions.
```

test/testdata/resolver/basic_constant_out_of_order__1.rb
```diff
@@ -0,0 +1,48 @@
+# check-out-of-order-constant-references: true
+# typed: false
+
+module Foo
+  A = X
```
jez, line 5:
```suggestion
  # This reports an error despite Foo::X also having a definition in the RBI file.
  A = X 
```

jez (approved):

PR #6673, opened 2023-01-20
Add RBI for URI::WS

Morriar:
### Motivation

Add missing payload definition for the `URI::WS` class: https://github.com/ruby/ruby/blob/v3_0_5/lib/uri/ws.rb#L22.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/uri.rbi
+++ b/rbi/stdlib/uri.rbi
@@ -651,6 +651,11 @@ module URI::Escape
 
 end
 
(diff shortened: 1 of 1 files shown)


froydnj (changes requested):
Looks like you have some `Array` vs. `T::Array` confusion in the RBI changes.

froydnj (approved):
Thanks!

PR #6675, opened 2023-01-23
Disable runtime type check in CustomType.scalar_type?

georg-stripe:
### Motivation

`T::Props::CustomType.checked_serialize` is called a fair amount during serialization. Creating this PR to check whether removing the runtime type check has any effect on performance in practice.

### Test plan

Behavior should be unchanged. Added a new microbenchmark.

Before:
```
$ bundle exec rake bench:serialize_custom_type
T::Props::CustomType.checked_serialize (μs/iter):
  1.118573   0.005625   1.124198 (  1.125563)
```
After:
```
$ bundle exec rake bench:serialize_custom_type
T::Props::CustomType.checked_serialize (μs/iter):
  0.477603   0.002454   0.480057 (  0.480592)
```

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/custom_type.rb
+++ b/gems/sorbet-runtime/lib/types/props/custom_type.rb
@@ -57,13 +57,12 @@ module T::Props
       raise 'Please use "extend", not "include" to attach this module'
     end
 
-    sig(:final) {params(val: Object).returns(T::Boolean).checked(:never)}
+    sig(:final) {params(val: T.untyped).returns(T::Boolean).checked(:never)}
     def self.scalar_type?(val)
       # We don't need to check for val's included modules in
       # T::Configuration.scalar_types, because T::Configuration.scalar_types
       # are all classes.
-      # We elide `klass : T.nilable(Class)` to avoid allocations.
-      klass = T.unsafe(val.class)
+      klass = val.class
       until klass.nil?
         return true if T::Configuration.scalar_types.include?(klass.to_s)
         klass = klass.superclass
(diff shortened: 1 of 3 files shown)


georg-stripe:

→ https://go/builds/bui_NDuvYEDuCpzuEQ
→ https://go/builds/bui_NDuv25X3Cbgkp3

gems/sorbet-runtime/lib/types/props/custom_type.rb
```diff
@@ -62,7 +62,8 @@ def self.scalar_type?(val)
       # We don't need to check for val's included modules in
       # T::Configuration.scalar_types, because T::Configuration.scalar_types
       # are all classes.
-      klass = T.let(val.class, T.nilable(Class))
+      # We elide `klass : T.nilable(Class)` to avoid allocations.
+      klass = T.unsafe(val.class)
```
jez, line 66: Another option is to mark `val` as having type `T.untyped` if you want to save the cost of the call to `T.unsafe` too.

Note that external callers of `scalar_type?` (anyone calling this outside of the sorbet-runtime gem) already see `val` as `T.untyped` because the sig on this method was not mirrored into `rbi/sorbet/tprops.rbi`.

jez (approved):
Thanks for writing a benchmark.

georg-stripe:

→ https://go/builds/bui_NEErGMJDVuPATs
→ https://go/builds/bui_NEErYb1tnWijRN

gems/sorbet-runtime/lib/types/props/custom_type.rb
georg-stripe: Thanks, that sounds even better!

PR #6679, opened 2023-01-25
Add docs for sorbet-runtime environment variables

jeffcarbs:
Adds docs for environment variables added in #6668. Happy to edit/modify the test if necessary, or reviewers can feel free to make the edits themselves.

### Motivation

Addresses https://github.com/sorbet/sorbet/pull/6668#discussion_r1085991360

### Test plan

n/a

### DIFF
--- a/gems/sorbet-runtime/README.md
+++ b/gems/sorbet-runtime/README.md
@@ -6,32 +6,6 @@ Full docs are at:
 - https://sorbet.org/docs/runtime
 - https://sorbet.org/docs/tconfiguration
 
-## Environment variables
-
-There are a number of environment variables that `sorbet-runtime` reads from to change its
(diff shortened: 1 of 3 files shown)


froydnj (changes requested):
It looks like there are some linting issues (which are annoying).

This looks reasonable to me, but I think we also want to remove the documentation for the environment variables that currently resides in `gems/sorbet-runtime/README.md` so that the documentation exists in a single place?

jeffcarbs:
@froydnj - Handled those issues:
- Remove the docs from the `sorbet-runtime/README.md`
- Run the formatter via `./tools/scripts/format_website.sh`

froydnj (approved):
Thank you!

PR #6680, opened 2023-01-25
Make the autocorrect for unresolved constants case-insensitive

maruth-stripe:
The autocorrect for unresolved constants used a case-insensitive Levenstein (edit) distance. This PR changes it to be case **insensitive** edit distance, with ties broken by case-sensitive edit distance (i.e. `d(s1, s2) = (insensitive(s1, s2), sensitive(s1, s2))`). 

### Motivation

(1) the case sensitive metric misses out on obvious capitalization typos, (2) this would help significantly accelerate/enable an ongoing codemod at Stripe.

### Test plan
eg fixture:
Screen Shot 2023-01-25 at 12 12 47 PM

previous autocorrect:
Screen Shot 2023-01-25 at 12 12 15 PM

new autocorrect:
Screen Shot 2023-01-25 at 12 09 56 PM

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -880,8 +880,6 @@ ClassOrModule::findMemberFuzzyMatchConstant(const GlobalState &gs, NameRef name,
                             }
                         }
 
-                        // auto thisDistance = Levenstein::distance(
-                        //     currentName, member.first.dataCnst(gs)->original.dataUtf8(gs)->utf8, best.distance);
                         auto thisDistance = caseInsensitiveAndSensitiveDist(
                             currentName, member.first.dataCnst(gs)->original.dataUtf8(gs)->utf8, best.distance);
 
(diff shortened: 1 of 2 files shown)


core/Symbols.cc
```diff
@@ -968,6 +976,12 @@ ClassOrModule::findMemberFuzzyMatchConstant(const GlobalState &gs, NameRef name,
     absl::c_reverse(result);
     return result;
 }
+std::pair<int, int> ClassOrModule::caseInsensitiveAndSensitiveDist(std::string_view s1, std::string_view s2,
+                                                                   int bound) noexcept {
+    int sensitive = Levenstein::distance(s1, s2, bound);
+    int insensitive = Levenstein::distance(absl::AsciiStrToLower(s1), absl::AsciiStrToLower(s2), bound);
```
jez, line 982: hm, this is already one of the slowest parts of sorbet, and we're now calling it twice... is there a way we can not do that?

maruth-stripe: ack

let me try and make a modified Levenstein distance which simultaneously does both

PR #6681, opened 2023-01-25
Catch certain kinds of incorrect calls to `raise`

### LINKED ISSUE #4209: Check signature of an error class constructor in `raise` called with two arguments
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20Error%20%3C%20StandardError%0A%20%20def%20initialize(message%2C%20arg)%0A%20%20%20%20super(message)%0A%20%20%20%20%40arg%20%3D%20arg%0A%20%20end%0Aend%0A%0Araise%20Error%2C%20%22Message%22)

```ruby
# typed: true
extend T::Sig

class Error < StandardError
  def initialize(message, arg)
    super(message)
    @arg = arg
  end
end

raise Error, "Message"
```

#### Observed output

```
No errors! Great job.
```

On runtime, `ArgumentError: wrong number of arguments (given 1, expected 2)` error occurs.

#### Expected behavior

`srb tc` reports an error like `Not enough arguments provided for method Error#initialize. Expected: 2, got: 1` for
```ruby
raise Error, "Message"
```

jez:
### Motivation

Fixes #4209

See commits for summary.

### Test plan

See included automated tests.

### DIFF
--- a/core/serialize/serialize.cc
+++ b/core/serialize/serialize.cc
@@ -631,6 +631,7 @@ void SerializerImpl::pickle(Pickler &p, const Method &what) {
         pickle(p, a);
     }
     pickle(p, what.resultType);
+    p.putU4(what.intrinsicOffset);
     p.putU4(what.locs().size());
     for (auto &loc : what.locs()) {
         pickle(p, loc);
@@ -663,6 +664,7 @@ Method SerializerImpl::unpickleMethod(UnPickler &p, const GlobalState *gs) {
     }
 
     result.resultType = unpickleType(p, gs);
+    result.intrinsicOffset = p.getU4();
     auto locCount = p.getU4();
     for (int i = 0; i < locCount; i++) {
         result.locs_.emplace_back(unpickleLoc(p));
@@ -1068,7 +1070,6 @@ void Serializer::loadGlobalState(GlobalState &gs, const uint8_t *const data) {
             "Can't load into a non-empty state");
     UnPickler p(data, gs.tracer());
     SerializerImpl::unpickleGS(p, gs);
-    gs.installIntrinsics();
 }
 
 uint32_t Serializer::loadGlobalStateUUID(const GlobalState &gs, const uint8_t *const data) {
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -4141,6 +4141,53 @@ public:
     }
 } Kernel_proc;
 
+class Kernel_raise : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        if (args.args.size() < 1) {
+            return;
+        }
+
+        if (!Types::isSubType(gs, args.args[0]->type, Types::classClass())) {
+            return;
+        }
+
+        uint16_t newNumPosArgs = args.args.size() >= 2 ? 1 : 0;
+        auto newSendArgLocs = InlinedVector<LocOffsets, 2>();
+        if (newNumPosArgs > 0) {
+            newSendArgLocs.emplace_back(args.locs.args[1]);
+        }
+        auto newCallLocs = CallLocs{
+            args.locs.file,
+            /* callLoc */ args.locs.args[0].join(args.locs.args[newNumPosArgs]),
+            /* receiverLoc */ args.locs.args[0],
+            /* funLoc */ args.locs.args[0].copyEndWithZeroLength(),
+            newSendArgLocs,
+        };
+
+        auto newSendArgs = InlinedVector<const TypeAndOrigins *, 2>();
+        if (newNumPosArgs) {
+            newSendArgs.emplace_back(args.args[1]);
+        }
+
+        auto newArgs = DispatchArgs{
+            Names::new_(),         newCallLocs,
+            newNumPosArgs,         newSendArgs,
+            args.args[0]->type,    *args.args[0],
+            args.args[0]->type,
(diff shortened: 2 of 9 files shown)


core/types/calls.cc
```diff
@@ -4386,6 +4444,7 @@ const vector<Intrinsic> intrinsics{
 
     {Symbols::Kernel(), Intrinsic::Kind::Instance, Names::proc(), &Kernel_proc},
     {Symbols::Kernel(), Intrinsic::Kind::Instance, Names::lambda(), &Kernel_proc},
+    {Symbols::Kernel(), Intrinsic::Kind::Instance, Names::raise(), &Kernel_raise},
```
froydnj, line 4447: Since `Kernel#raise` and `Kernel#fail` are the same function (#5930), we should have an intrinsic for `Kernel#fail` as well?

core/types/calls.cc
```diff
@@ -4141,6 +4141,64 @@ class Kernel_proc : public IntrinsicMethod {
...
+        // `.new` (https://ruby-doc.org/core-2.7.2/Exception.html#method-c-exception) but anything
+        // that defines a method called `exception` could be called. We don't implement that here,
+        // because implicit conversions are hard in the presence of subtyping. Instead, we restrict
+        // the check to only subclasses of `Exception`, not arbitrary classes.
+        auto classOfException = make_type<ClassType>(Symbols::Exception().data(gs)->lookupSingletonClass(gs));
+        if (!args.args[0]->type.isUntyped() && !Types::isSubType(gs, args.args[0]->type, classOfException)) {
+            return;
+        }
+
+        uint16_t newNumPosArgs = args.args.size() >= 2 ? 1 : 0;
```
froydnj, line 4166: This was surprising to me until I looked at the documentation and saw that `raise` doesn't forward an unlimited number of arguments.  Bizarre.

core/types/calls.cc
```diff
@@ -4141,6 +4141,64 @@ class Kernel_proc : public IntrinsicMethod {
...
+        // - arguably none of this intrinsic should run if we don't know that arg0 isn't a Class
+        //
+        // Technically speaking, the Ruby VM actually implements this by way of calling
+        // arg0.exception. It just so happens that Exception.exception is defined to forward to
+        // `.new` (https://ruby-doc.org/core-2.7.2/Exception.html#method-c-exception) but anything
+        // that defines a method called `exception` could be called. We don't implement that here,
+        // because implicit conversions are hard in the presence of subtyping. Instead, we restrict
+        // the check to only subclasses of `Exception`, not arbitrary classes.
+        auto classOfException = make_type<ClassType>(Symbols::Exception().data(gs)->lookupSingletonClass(gs));
+        if (!args.args[0]->type.isUntyped() && !Types::isSubType(gs, args.args[0]->type, classOfException)) {
```
froydnj, line 4162: WDYT about pulling `args.args[0]` out into a separate variable (`klass` or somesuch) and using it here and below?

test/testdata/intrinsics/kernel_raise.rb
```diff
@@ -0,0 +1,79 @@
...
+    raise MyError, "arg"
+    #              ^^^^^ error: Missing required keyword argument `input` for method `MyError#initialize`
+    #              ^^^^^ error: Too many positional arguments provided for method `MyError#initialize`. Expected: `0`, got: `1`
+  end
+
+  0.times do
+    raise cls, "arg"
+    #          ^^^^^ error: Missing required keyword argument `input` for method `MyError#initialize`
+    #          ^^^^^ error: Too many positional arguments provided for method `MyError#initialize`. Expected: `0`, got: `1`
+  end
```
froydnj, line 56: WDYT about additional tests:

1. Cases where the `initialize` method takes a single arg, but we provide more to `raise`, e.g. `raise MyError, "arg", ["more", "stuff"]` -- trying to verify that we don't pass those extra arguments to the `new` call.
2. Cases where the `initialize` method takes multiple args, but we only provide the single arg (I think this was the original motivating example?)
3. Cases where the `initialize` method takes multiple args, and we provide those multiple args to `raise`, but of course they are not handled.  This is kind of redundant with case 1, I guess?

core/serialize/serialize.cc
```diff
@@ -631,6 +631,7 @@ void SerializerImpl::pickle(Pickler &p, const Method &what) {
         pickle(p, a);
     }
     pickle(p, what.resultType);
+    p.putU4(what.intrinsicOffset);
```
froydnj, line 634: Whoops, good catch.

froydnj (approved):
Mostly extra things to consider; the `Kernel#fail` bit should be addressed before landing, I think.

core/types/calls.cc
jez: 🙂-intensifies

core/serialize/serialize.cc
jez: I mentioned this in the commit message but I think this wasn't so much a bug as much as a choice not to serialize function pointers to a cache (before intrinsic became intrinsicOffset)

jez:
> Mostly extra things to consider; the `Kernel#fail` bit should be addressed before landing, I think.

I believe I covered all your points. Going to auto-merge, but lmk if you have anything else.

jez:

→ https://go/builds/bui_NEle7FBO5QEwyg
→ https://go/builds/bui_NEle3YPO0md9IR

PR #2683, opened 2020-02-20
Optimize serde with codegen

djudd-stripe:
Instead of serializing and deserializing by looping through props at runtime, we generate code for serde at load time* and `class_eval` it. 

(*Technically, we do this lazily, but in production, we’d eagerly evaluate on autoload, and then disable further `class_eval` calls by this path.)

I'm aware of two subtle behavior changes that this produces for `serialize` and `from_hash`:
1. We'll now correctly serde `T::Set[SomeSerializable]` as well as nested `T::Hash[.., T::Array[SomeSerializable]` and the like. This seems like a low-risk improvement since we'd previously fail loudly if we tried to persist serialized data with these types, but it's possible we are relying on the broken behavior somewhere for in-memory structs.
2. We'll now always default to cloning unknown types (e.g. when prop type is `Object`). Previously we'd default to cloning them if-and-only-if they were inside a collection--i.e. for `T::Array[Object]` we'd clone all the values, but for just plain `Object`, we wouldn't clone at all; see the naive `type_needs_clone` implementation in master. There's a risk to changing this if we rely on object identity being preserved on serde in this edge case somewhere.

### Motivation
This produces more than 2x speedups in the benchmarks.

Before:
```
➜  sorbet-runtime git:(master) bundle exec rake bench:deserialize
T::Props.from_hash, mostly nil input (μs/iter):
  4.414409   0.003165   4.417574 (  4.420272)
T::Props.from_hash, all props set (μs/iter):
  9.395371   0.011826   9.407197 (  9.416477)
```

After:
```
T::Props.from_hash, mostly nil input (μs/iter):
  1.710600   0.000300   1.710900 (  1.710909)
T::Props.from_hash, all props set (μs/iter):
  3.822127   0.000889   3.823016 (  3.823426)
```

### Test plan
Tests here & (mostly) in pay-server - https://git.corp.stripe.com/stripe-internal/pay-server/compare/djudd/t-props-serde-class-eval

PR #6684, opened 2023-01-26
Codegen the T::Struct initializer

jeffcarbs:
_NOTE: This is a follow-up to #6667 and contains all of its commits as well. To see this PR's changes, just look at the last commit. It's also somewhat of a WIP still. There's a bunch of code that we'd be able to remove and some tests that are breaking and need to be investigated_

----

Follow-up to #6667 and similar to #2683, we codegen the T::Struct initializer

### Motivation

T::Struct initialization can be slow. Even when disabling type checking via `checked: :tests/:never`, it's still 6x slower than a plain ruby object. This PR drastically improves performance when disabling type checking while keeping performance slightly better or unchanged in the typechecking case:

**Before**
```
Plain Ruby (μs/iter):
  0.448727   0.001123   0.449850 (  0.450012)
T::Props.new, mostly nil input (μs/iter):
  3.340299   0.018830   3.359129 (  3.375337)
T::Props.new, all props set (μs/iter):
  6.748173   0.041695   6.789868 (  6.843103)
T::Props.new, all props set, checked(:never) (μs/iter):
  2.909526   0.007162   2.916688 (  2.919772)
```

**After**
```
Plain Ruby (μs/iter):
  0.460523   0.003017   0.463540 (  0.465256)
T::Props.new, mostly nil input (μs/iter):
  2.225582   0.009142   2.234724 (  2.245390)
T::Props.new, all props set (μs/iter):
  6.684362   0.022187   6.706549 (  6.722224)
T::Props.new, all props set, checked(:never) (μs/iter):
  0.844563   0.002915   0.847478 (  0.848391)
```

### Test plan

See included automated tests.


gems/sorbet-runtime/bench/constructor.rb
```diff
@@ -25,13 +40,43 @@ class Subdoc < T::Struct
...
 
+      100_000.times do
+        ExamplePoro.new(input)
+      end
+
+      result = Benchmark.measure do
+        1_000_000.times do
+          ExamplePoro.new(input)
+        end
+      end
```
froydnj, line 75: I'd be interested to see what the benchmarks look like if they weren't hacking around limitations in the (MRI) Ruby VM, and using keyword arguments instead.

jeffcarbs: Surprisingly, I tried an example using keyword args and they were actually slower than using the hash 🤔 
```
Plain Ruby (μs/iter):
  0.463951   0.002514   0.466465 (  0.468203)
Plain Ruby Kwargs (μs/iter):
  0.475856   0.002753   0.478609 (  0.480072)
```

where the Kwargs example was:
<details>
<summary>Script</summary>

```ruby
class ExamplePoroKwargs
  def initialize(prop1: nil, prop2: 0, prop3:, prop4:, prop5: [], prop6:, prop7: {}, prop8: nil, prop9: [], prop10: {})
    @prop1 = prop1
    @prop2 = prop2
    @prop3 = prop3
    @prop4 = prop4
    @prop5 = prop5
    @prop6 = prop6
    @prop7 = prop7
    @prop8 = prop8
    @prop9 = prop9
    @prop10 = prop10
  end
end

# ...

prop3 = 0
prop4 = []
prop6 = {}

100_000.times do
  ExamplePoroKwargs.new(prop3: prop3, prop4: prop4, prop6: prop6)
end

result = Benchmark.measure do
  1_000_000.times do
    ExamplePoroKwargs.new(prop3: prop3, prop4: prop4, prop6: prop6)
  end
end

puts "Plain Ruby Kwargs (μs/iter):"
puts result
```

</details>

froydnj: This is what I would have expected; the keyword args force the Ruby VM to allocate a hash on every call, plus putting the arguments into the hash, etc.

I think the common behavior inside Stripe is to use keyword arguments, rather than passing in a hash (though I guess some people might be getting JSON arguments off the wire and passing that hash into a `T::Struct` constructor directly?).

gems/sorbet-runtime/lib/types/props/utils.rb
```diff
@@ -43,6 +43,12 @@ def self.need_nil_write_check?(prop_rules)
     need_nil_read_check?(prop_rules) || T::Props::Utils.required_prop?(prop_rules)
   end
 
+  # The prop_rules indicate whether we should typecheck the value for the prop/field.
+  def self.need_type_check?(prop_rules)
```
djudd, line 47: A micro-optimization for an edge case that still might be worthwhile: `return false if prop_rules[:type_object] == T.untyped`

gems/sorbet-runtime/lib/types/props/private/initializer_generator.rb
```diff
@@ -0,0 +1,114 @@
...
+              raise ArgumentError.new("Missing required prop `#{prop}` for class `\#{self.class.name}`")
+            RUBY
+          else
+            missing_key_value = <<~RUBY.strip
+              nil
+            RUBY
+          end
+
+          <<~RUBY.strip
+            #{ivar_name} = if hash.key?(#{hash_key})
```
djudd, line 68: I dislike that we're doing two hash lookups for each key in the normal case, but I'm not sure there's an alternative that's simple enough to be worthwhile.

You could do something like:
```
value = present_key_value
if value.nil? && !hash.key?(#{hash_key})
  value = missing_key_value
else
  result += 1
end
```

... but I think that might change what error is raised for missing required props in a way that we don't want. Maybe worth benchmarking and then seeing if there's a reasonable solution for that problem if it's meaningfully faster?

Also, this makes me realize I'm not actually sure what behavior we expect or want in the case where a required prop with `checked: :never` is missing from the constructor.

djudd (commented):
Overall this is exciting, although I'd be pretty unsurprised if somehow it had like 1000 test failures in Stripe's monorepo.

gems/sorbet-runtime/lib/types/props/utils.rb
jeffcarbs: I tried this and it breaks some tests related to `setter_validate`. Currently, if `need_type_check?` is false we don't run any validations (e.g. `need_nil_write_check` or `setter_validate`) and the tests were set up where the props were `T.untyped` but had a `setter_validate` set.

I think this behavior makes sense since when using `checked(:never)` the prop basically operates as `attr_writer` (in fact that's how the setter is defined) with the small exception of being able to use default/factory.

While I think the behavior makes sense, I think maybe holding off and making this change/optimization on a separate PR may be a better option.

gems/sorbet-runtime/lib/types/props/private/initializer_generator.rb
jeffcarbs: Great call! I pushed up a commit which does basically what you suggested and optimizes for the case where the value is present. See the updated PR description with the performance benchmarks, it mostly improves performance across the board 🚀 although it flip/flops the `checked(:never)` runtimes but in an expected and desirable way - when props are mostly given and non-null it's faster.

jeffcarbs:
@djudd - I just rebased and pushed up the latest code that has all the specs passing locally and _I think_ keeps behavior unchanged from the latest main branch. Would be interested to see how it fares on the stripe codebase 😅 

I'm thinking about how to make this easiest to review and safely merge. Let me know if either of these options sound good to you, or if there's another option you can think of:
- (a) Remove the stuff in this PR related to `checked(:never)` and keep this solely changing to codegen the initializer while keeping behavior entirely unchanged.
  - Open a separate PR to add the `checked(:never)` functionality
- (b) Get #6667 reviewed and merged first
  - Then rebase this PR and have it still just change to codegen the initializer but obviously keeping the new "checked" behavior
- (c) Close #6667, revert some of the unrelated "checked" changes, but keep `checked(:never)` behavior change in here and review/merge all at once.

The "setter_proc" code complicates the `checked(:never)` code a bit since the setter_proc is used in both construction and setting, and I think a lot of stuff related to the "setter_proc" may be able to be made much simpler or go away entirely in this PR, which pushes me toward option (a) or (c). And given the impact and importance of the code being changed, probably leaning toward (a).

Let me know what you think.

djudd:
I don't work at Stripe anymore so I'm not going to have final say, but FWIW option (a) makes sense to me.

jeffcarbs:
> a lot of tests failed very early on with errors of the form:

Would you be able to share a minimal repro of the error you're seeing? It would help debug and fix the issue.

froydnj:
> > a lot of tests failed very early on with errors of the form:
> 
> Would you be able to share a minimal repro of the error you're seeing? It would help debug and fix the issue.

```ruby
module Sorbet
  class Doc
    include T::Props
    include T::Props::Serializable
    include T::Props::WeakConstructor
  end

  class AbstractExample < Doc
    prop :type, T.nilable(String), raise_on_nil_write: true

    def initialize(hash = {})
      @type = "value"
      super(hash)
    end
  end

  class Example < AbstractExample
  end
end

p Sorbet::Example.new.serialize
```

This example fails with this PR applied and succeeds with master.

PR #6685, opened 2023-01-27
LSP fast path includes *all* change-relevant __package.rb files

aadi-stripe:
Makes a change so that *all* change-relevant package files are included in the LSP fast path, not just the ones that correspond to edited files.

### Motivation

As per my understanding of the code (which could be inaccurate), it seems that if a file is type-checked on the fast path, static field symbols always get redefined. This can be problematic for the `exported` flag on a symbol, because that is set when the visibility checker pass walks over the corresponding `__package.rb` file. Given that, the following type of scenario can occur:

Assume:
* Package A defines and exports a static field `A::S`. Assume this field is defined in `a/s.rb`
* Package B defines and exports a constant `B::T`. Assume this constant is defined in `b/t.rb`
* `b/t.rb` refers to `A::S`, and as a result, B imports A.
* `a/s.rb` refers to `B::T`, and as a result, A imports B.

Thus, A and B have a dependency cycle. When a fast path change to `b/t.rb` results in `a/s.rb` also getting typechecked, all static fields in `a/s.rb` get redefined. Hence `A::S` gets redefined and loses its "exported" flag, and we see a "resolved but not visible" error at the usage site in `b/t.rb`.

In order to fix this issue, I am making a change to `LSPTypechecker` such that the package file for package A (i.e. `a/__package.rb`) is also included in the set of files to typecheck. This ensures that the exported bit gets set on the fast path for the redefined static field `A::S`.

### Test plan

Open to making changes to any data structures used based on what's optimal.

I believe that this is unlikely to affect the performance of Sorbet on the Stripe codebase -- however, since in the worst case, it ~doubles the number of files checked in the fast path, it's definitely worth validating.


test/testdata/packager/lsp_update_packaged_file_enum/a.1.rbupdate
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+# typed: strict
+# assert-fast-path: a.rb,__package.rb,dep/exported.rb,dep/__package.rb
+
+class Package::A
+  extend T::Sig
+
+  sig {returns(Integer)}
+  def self.a
+    Dep::Exports::ExportedClass::Val1
```
aadi-stripe, line 10: Without this change, we get an error here in the fast path.

aadi-stripe:

→ https://go/builds/bui_NG0i8YVZJ4VZaT
→ https://go/builds/bui_NG0ixdYMjoFfDS

test/testdata/packager/lsp_update_packaged_file_static_field/a.1.rbupdate
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+# typed: strict
+# assert-fast-path: a.rb,__package.rb,dep/exported.rb,dep/__package.rb
+
+class Package::A
+  extend T::Sig
+
+  sig {returns(String)}
+  def self.a
+    Dep::Exports::ExportedClass::Val
```
aadi-stripe, line 10: Without this change, the reference on line 10 here reports a resolved but not visible error on the fast path.

main/lsp/LSPTypechecker.cc
```diff
@@ -246,28 +246,40 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
         // their visibility, which only happens if we're running incrementalNamer.
         if (shouldRunIncrementalNamer) {
             // NOTE: Using `gs` to access package information here assumes that edits to __package.rb
             // files don't take the fast path. We'll want (or maybe need) to revisit this when we start
             // making edits to `__package.rb` take fast paths.
             if (!(fref.data(*gs).isPackage())) {
-                auto pkgName = gs->packageDB().getPackageNameForFile(fref);
-                if (pkgName.exists()) {
+                auto &pkg = gs->packageDB().getPackageForFile(*gs, fref);
+                if (pkg.exists()) {
```
froydnj, line 262: AIUI, the check here for `pkgName` before and the check here for `pkg` after are basically identical?  But I assume that it's not, because either this part or the selection process for `__package.rb` files must be the behavior-changing part of the patch.  What am I missing?

main/lsp/LSPTypechecker.cc
```diff
@@ -246,28 +246,40 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
-                    auto packageFref = gs->packageDB().getPackageInfo(pkgName).fullLoc().file();
-                    if (result.changedFiles.find(packageFref) == result.changedFiles.end()) {
-                        // Skip duplicates
-                        toTypecheck.emplace_back(packageFref);
+                    auto packageFref = pkg.fullLoc().file();
+                    if (!packageFref.exists()) {
+                        continue;
                     }
+
+                    packageFiles.emplace(packageFref);
```
froydnj, line 271: So before, for each changed file, we'd locate the `__package.rb` associated with the file.  If that `__package.rb` wasn't in the list of changed files, we would add the `__package.rb` to the list of files to typecheck.

Note that we were already adding all of the changed files to the list of files to typecheck.  So we only add the `__package.rb` files that weren't already changed.

After, for each changed file, we're going to locate the `__package.rb` associated with that file.  After deduplicating the `__package.rb` files, we will add each `__package.rb` to the list of files to typecheck.

This...looks like the same algorithm (albeit with a deduplication step)?  What am I missing?

main/lsp/LSPTypechecker.cc
```diff
@@ -246,28 +246,40 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
                 }
             }
         }
     }
+
+    for (auto packageFref : std::move(packageFiles)) {
```
froydnj, line 277: I think this `move` is unnecessary.  (I guess it's trying to say that we're done with `packageFiles`, which I agree is helpful information...but I don't think that fact is relevant here.)

main/lsp/LSPTypechecker.cc
```diff
@@ -246,28 +246,40 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
+        if (std::find(toTypecheck.begin(), toTypecheck.end(), packageFref) == toTypecheck.end()) {
+            toTypecheck.emplace_back(packageFref);
+        }
```
froydnj, line 280: Since `toTypecheck` prior to this loop should just have the same files as `result.changedFiles`, can we just check for membership in `result.changedFiles` instead, and not be accidentally quadratic?

froydnj (changes requested):
The description of the problem the PR solves is excellent.  But I feel like I must be missing something about how the actual patch works, because I don't see _why_ this fixes the issue.

main/lsp/LSPTypechecker.cc
aadi-stripe: `toTypecheck` has `extraFiles`

aadi-stripe:
@froydnj 

In the previous logic, we were only checking for package files corresponding to `changedFiles`. Now, we are checking for those from `extraFiles` as well.

aadi-stripe:
> > I believe that this is unlikely to affect the performance of Sorbet on the Stripe codebase. In the worst case, it ~doubles the number of files checked in the fast path, but that would still be too small a number to defeat Sorbet's inherent parallelism.
> 
> Note that the fast path is a serial typecheck, not a parallel one.

Ah okay. Did not realize.

main/lsp/LSPTypechecker.cc
aadi-stripe: See explanation below

main/lsp/LSPTypechecker.cc
aadi-stripe: It's that `toTypecheck` also contains `result.extraFiles`, see line 220 above. Before, we were only looking for `package.rb` files corresponding to `changedFiles`, but now we are also looking for those corresponding to `extraFiles`.

main/lsp/LSPTypechecker.cc
aadi-stripe: In theory I think this can still work with `changedFiles`, however, because I don't think `extraFiles` will contain any package files.

main/lsp/LSPTypechecker.cc
aadi-stripe: I agree -- but I think I wanted to leave it there to essentially communicate that to the reader or future editors of this code. It was helpful, for example, on line 220 to have `std::move(result.extraFiles)`, so I knew I couldn't use it down here.

main/lsp/LSPTypechecker.cc
aadi-stripe: Fixed this as it will still work with `changedFiles` -- on account of `extraFiles` not having any package files.

aadi-stripe:
@froydnj Made one more change, based on the feedback.

main/lsp/LSPTypechecker.cc
froydnj: Ah, that was the bit I was missing, thank you.

main/lsp/LSPTypechecker.cc
```diff
@@ -246,28 +246,40 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
 
         toTypecheck.emplace_back(fref);
+    }
+
+    UnorderedSet<core::FileRef> packageFiles;
 
+    for (auto fref : toTypecheck) {
         // Only need to re-run packager if we're going to delete constants and have to re-define
         // their visibility, which only happens if we're running incrementalNamer.
         if (shouldRunIncrementalNamer) {
```
froydnj, line 256: Maybe we should now invert the logic here:

```c++
if (shouldRunIncrementalNamer) {
    UnorderedSet<core::FileRef> packageFiles;

    for (auto fref : toTypecheck) {
        ...
    }

    // include them in `toTypecheck` if necessary
}
```

froydnj (approved):
Talking with Aadi offline confirms that this code doesn't trigger on the completion path or the hover path, so it should be relatively safe to add performance-wise, even if it's adding extra files.  I'm still a little skeptical of the bit that adds package files to `toTypecheck`, but we can do a followup if metrics shows this to be a problem in practice.

main/lsp/LSPTypechecker.cc
aadi-stripe: ah yes, good call

Issue #4150, opened 2021-04-16
Hovering on method called on union type shows effective type of method twice


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20%23%20Docs%20for%20A%23foo%0A%20%20sig%20%7Breturns(String)%7D%0A%20%20def%20foo%3B%20''%3B%20end%0Aend%0A%0Aclass%20B%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20%23%20Docs%20for%20B%23foo%0A%20%20sig%20%7Breturns(Integer)%7D%0A%20%20def%20foo%3B%200%3B%20end%0Aend%0A%0Asig%20%7Bparams(x%3A%20T.any(A%2C%20B)).void%7D%0Adef%20example(x)%0A%20%20x.foo%0A%20%20%23%20%5E%20hover%20here%0A%20%20%23%20it%20shows%20the%20method%20def%20twice%0Aend)

```ruby
# typed: true
extend T::Sig

class A
  extend T::Sig

  # Docs for A#foo
  sig {returns(String)}
  def foo; ''; end
end

class B
  extend T::Sig

  # Docs for B#foo
  sig {returns(Integer)}
  def foo; 0; end
end

sig {params(x: T.any(A, B)).void}
def example(x)
  x.foo
  # ^ hover here
  # it shows the method def twice
end
```

#### Observed output

```
No errors! Great job.
```

<img width="443" alt="Screen Shot 2021-04-15 at 6 02 03 PM" src="https://user-images.githubusercontent.com/5544532/114956897-b057a300-9e14-11eb-8f1f-67d730ce722e.png">

#### Expected behavior

It should either:

- show two defs, one for each method, or ideally:
- one `def` with the effective type.

- - -

PR #6687, opened 2023-01-30
Collapse duplicate hover results for union types

bitwise-aiden:
### Motivation
Fixes: #4150

Relies on https://github.com/sorbet/sorbet/pull/6691 being merged.

### Commit Summary

- **Update methodInfoString to collapse duplicate items** ([a48d71a](https://github.com/sorbet/sorbet/pull/6687/commits/a48d71ac56929ee849f6faaa42ef71ee345dd5e5))

In `hover.cc` I've updated `methodInfoString` to create a list of the method signatures in the order found in dispatch. If there are more than 1 method signatures generated this way, it will add an additional `-> <retType>` to the info before adding the method definition. Otherwise it will just add the method definition. These two cases look like so: 
```ruby 
# Multiple
sig {returns(String)}
sig {returns(Integer)}

-> T.any(String, Integer)
def foo; end

# Single 
sig {returns(String)}
def foo; end
```

- **Add hover union test harness** ([f166267](https://github.com/sorbet/sorbet/pull/6687/commits/f166267308623f823d7c811d613912bd0522bb71))

I've added a test harness based off of the case outlined in the issue (linked above), aiming to validate that the duplicate method type is now removed. 

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/hover.cc
+++ b/main/lsp/requests/hover.cc
@@ -22,8 +22,13 @@ string methodInfoString(const core::GlobalState &gs, const core::TypePtr &retTyp
         if (component.method.exists()) {
             auto method = component.method;
 
-            contents.push_back(
-                prettySigForMethod(gs, method, component.receiver, method.data(gs)->resultType, constraint.get()));
+            auto localRetType =
+                getResultType(gs, method.data(gs)->resultType, method, component.receiver, constraint.get());
+            if (localRetType == nullptr || localRetType.isUntyped()) {
+                localRetType = retType;
+            }
+
+            contents.push_back(prettySigForMethod(gs, method, component.receiver, localRetType, constraint.get()));
         }
         start = start->secondary.get();
     }
--- a/main/lsp/requests/hover.cc
+++ b/main/lsp/requests/hover.cc
@@ -14,22 +14,29 @@ namespace sorbet::realmain::lsp {
 string methodInfoString(const core::GlobalState &gs, const core::TypePtr &retType,
                         const core::DispatchResult &dispatchResult,
                         const unique_ptr<core::TypeConstraint> &constraint) {
-    string contents;
+    vector<string> contents;
     auto start = &dispatchResult;
-    ;
+
     while (start != nullptr) {
         auto &component = start->main;
         if (component.method.exists()) {
-            if (!contents.empty()) {
-                contents += "\n";
-            }
-            contents = absl::StrCat(
-                contents, prettyTypeForMethod(gs, component.method, component.receiver, retType, constraint.get()));
+            auto method = component.method;
+
+            contents.push_back(
+                prettySigForMethod(gs, method, component.receiver, method.data(gs)->resultType, constraint.get()));
         }
         start = start->secondary.get();
     }
 
-    return contents;
+    if (dispatchResult.main.method.exists()) {
+        if (contents.size() > 1) {
+            contents.push_back(fmt::format("\n-> {}", retType.show(gs)));
+        }
+
+        contents.push_back(prettyDefForMethod(gs, dispatchResult.main.method));
+    }
+
+    return absl::StrJoin(contents, "\n");
 }
 
 HoverTask::HoverTask(const LSPConfiguration &config, MessageId id, std::unique_ptr<TextDocumentPositionParams> params)
--- a/test/testdata/lsp/hover_union.rb
+++ b/test/testdata/lsp/hover_union.rb
@@ -1,0 +1,42 @@
+# typed: true
+extend T::Sig
+
+class A
+  extend T::Sig
+
+  # Docs for A#foo
+  sig {returns(String)}
+  def foo; ''; end
+end
+
+class B
+  extend T::Sig
+
+  # Docs for B#foo
+  sig {returns(Integer)}
+  def foo; 0; end
+end
+
+sig {params(x: T.any(A, B)).void}
+def multiple(x)
+  x.foo
+  # ^ hover-line: 1 ```ruby
+  # ^ hover-line: 2 sig {returns(String)}
+  # ^ hover-line: 3 sig {returns(Integer)}
+  # ^ hover-line: 5 -> T.any(String, Integer)
+  # ^ hover-line: 6 def foo; end
+  # ^ hover-line: 7 ```
+  # ^ hover-line: 9 ---
+  # ^ hover-line: 11 Docs for A#foo
+end
+
+sig {params(x: A).void}
+def single(x)
+  x.foo
+  # ^ hover-line: 1 ```ruby
+  # ^ hover-line: 2 sig {returns(String)}
+  # ^ hover-line: 3 def foo; end
+  # ^ hover-line: 4 ```
+  # ^ hover-line: 6 ---
+  # ^ hover-line: 8 Docs for A#foo
+end
--- a/main/lsp/LSPLoop.h
+++ b/main/lsp/LSPLoop.h
@@ -94,6 +94,9 @@ std::optional<std::string> findDocumentation(std::string_view sourceCode, int be
 bool hideSymbol(const core::GlobalState &gs, core::SymbolRef sym);
 std::unique_ptr<MarkupContent> formatRubyMarkup(MarkupKind markupKind, std::string_view rubyMarkup,
                                                 std::optional<std::string_view> explanation);
+std::string prettyDefForMethod(const core::GlobalState &gs, core::MethodRef method);
+std::string prettySigForMethod(const core::GlobalState &gs, core::MethodRef method, const core::TypePtr &receiver,
+                               core::TypePtr retType, const core::TypeConstraint *constraint);
 std::string prettyTypeForMethod(const core::GlobalState &gs, core::MethodRef method, const core::TypePtr &receiver,
                                 const core::TypePtr &retType, const core::TypeConstraint *constraint);
 std::string prettyTypeForConstant(const core::GlobalState &gs, core::SymbolRef constant);


test/testdata/lsp/hover_union.rb
```diff
@@ -0,0 +1,29 @@
...
+
+sig {params(x: T.any(A, B)).void}
+def example(x)
+  x.foo
+  # ^ hover: ```ruby
+  # ^ hover: sig {returns(T.any(String, Integer))}
+  # ^ hover: def foo; end
+  # ^ hover: ```
+  # ^ hover: ---
+  # ^ hover: Docs for A#foo
```
bitwise-aiden, line 28: Currently this doesn't work as expected since hover assertions operate on a per line basis, with no support for multi-line hover. Is there a recommended approach for achieving multiline?

froydnj: This test passes locally for me, and editing one of the lines here triggers test failures.  `test/testdata/lsp/hover.rb` contains a similar multi-line matching testcase.

main/lsp/requests/hover.cc
```diff
@@ -14,22 +14,18 @@ namespace sorbet::realmain::lsp {
...
-            }
-            contents = absl::StrCat(
-                contents, prettyTypeForMethod(gs, component.method, component.receiver, retType, constraint.get()));
+            contents.insert(prettyTypeForMethod(gs, component.method, component.receiver, retType, constraint.get()));
         }
         start = start->secondary.get();
     }
 
-    return contents;
+    return absl::StrJoin(contents, "\n");
```
froydnj, line 28: This will return the type descriptions in sorted order of the type descriptions, which might not correspond to the way the user wrote the type...though it's hard once you introduce deduplication into the picture to get "close" to what the user was writing.

WDYT about changing the code (along with testcases) so that it returns the type descriptions in the (deduplicated) order that they appeared in the `dispatchResult`?

I'm not sure offhand it it would make a difference.

froydnj (changes requested):
Thanks for the contribution!  I'm marking this as "changes needed", not because I think the patch is wrong in any way, but because I want some clarification on the below two points.  Let me know what you think!

bitwise-aiden:
> This test passes locally for me, and editing one of the lines here triggers test failures. `test/testdata/lsp/hover.rb` contains a similar multi-line matching testcase.

Do you mind elaborating on triggering the failure a bit further? 

Looking into [sorbet/lsp_test_runner.cc](https://github.com/sorbet/sorbet/blob/master/test/lsp_test_runner.cc) and [HoverAssertion#checkAll](https://github.com/sorbet/sorbet/blob/master/test/helpers/position_assertions.cc#L1168-L1176), it appears as though each `^ hover: <assertion>` is treated independently of one another. 

This was confirmed by having multiple `^ hover: def foo; end` one after another; each matching against the same part of the response. Another example being from `test/testdata/lsp/hover.rb` as stated above: 
```ruby
a = BigFoo::LittleFoo1.new
          # ^^^^^^^^^^ hover: T.class_of(BigFoo::LittleFoo1)
          # ^^^^^^^^^^ hover: The docs for LittleFoo1
          # ^^^^^^^^^^ hover: The docs for LittleFoo1
``` 

I'm wondering if `^ hover:` would benefit from additional syntax that denotes the line breaks, or even a `^ hover-complete:` to check the full message. 

What do you think?

main/lsp/requests/hover.cc
bitwise-aiden: Ahh, good call out! That's my misunderstanding. It's been awhile since I used `std::set`, I had assumed that it maintained the order that the items were inserted into it. 

I had been aiming for the order that you've requested and agree that this makes the most sense. I shall update the code to match that now.

jez: > the way the user wrote the type

hm, I'm not sure it makes sense to consider anything about "the order the user wrote the types" here

the thing that's showing up in the `resultType` of the dispatch result is a computed value (which might be the result of unioning or intersecting arbitrary types, arriving at an entirely new type). In this example yes because it's a simple union of class types there's a clear order, but in general types might collapse or combine in ~arbitrary ways.

froydnj:
> > This test passes locally for me, and editing one of the lines here triggers test failures. `test/testdata/lsp/hover.rb` contains a similar multi-line matching testcase.
> 
> Do you mind elaborating on triggering the failure a bit further?

I edited your `hover_union.rb` testcase to have `  # ^ hover: sig {returns(T.all(String, Integer))}` instead of the original `T.any` and got what looked like a reasonable failure (though maybe a little unhelpful not calling out the multiline nature of the response):

(lightly edited so markdown still works in the rest of the message)

```
INFO: From Testing //test:test_LSPTests/testdata/lsp/hover_union:
==================== Test output for //test:test_LSPTests/testdata/lsp/hover_union:
+ exec test/lsp_test_runner --single_test=test/testdata/lsp/hover_union.rb
Pausing
Resuming
[doctest] doctest version is "2.4.9"
[doctest] run with "--help" for options
===============================================================================
test/lsp_test_runner.cc:525:
TEST CASE:  LSPTest

test/testdata/lsp/hover_union.rb:22: ERROR: Expected hover contents:
22:   x.foo
        ^ hover: sig {returns(T.all(String, Integer))}
Found hover contents:
22:   x.foo
        ^ hover: ```ruby
sig {returns(T.any(String, Integer))}
def foo; end
[inserted-space-to-foil-markdown-in-markdown]```

---

Docs for A#foo
```

> I'm wondering if `^ hover:` would benefit from additional syntax that denotes the line breaks, or even a `^ hover-complete:` to check the full message.
> 
> What do you think?

We've generally not had that much call to deal with multi-line hover assertions and when we have, I think we've been comfortable with the current facility, even if it doesn't enforce line ordering.  My sense is that if you wanted to try to specify a multi-line thing on a single line, you quickly get into the need for escaping/quoting, and that gets tedious and error-prone very fast.

If you wanted to tackle this, I think adding a `hover-multiline` assertion would be reasonable:

```
# hover-multiline: LINE-NUMBER MATCH
```

where the `LINE_NUMBER` indicates the, well, line number of the multi-line match, and the MATCH is written in the same fashion as `hover:`'s assertion is.  This removes the need for the quoting and enables you to specify that the line ordering must match in a particular way.  WDYT?

jez (commented):
For what it's worth, I'm not sure that doing this using a set of strings is the best approach here. Consider this example:

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20%23%20Docs%20for%20A%23foo%0A%20%20sig%20%7Bparams%28x%3A%20Integer%29.returns%28String%29%7D%0A%20%20def%20foo%28x%29%3B%20''%3B%20end%0Aend%0A%0Aclass%20B%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20%23%20Docs%20for%20B%23foo%0A%20%20sig%20%7Bparams%28y%3A%20T.any%28Integer%2C%20String%29%29.returns%28Integer%29%7D%0A%20%20def%20foo%28y%29%3B%200%3B%20end%0Aend%0A%0Asig%20%7Bparams%28x%3A%20T.any%28A%2C%20B%29%29.void%7D%0Adef%20example%28x%29%0A%20%20x.foo%280%29%0A%20%20%23%20%5E%20hover%20here%0A%20%20%23%20it%20shows%20the%20method%20def%20twice%0Aend)

```ruby
# typed: true
extend T::Sig

class A
  extend T::Sig

  # Docs for A#foo
  sig {params(x: Integer).returns(String)}
  def foo(x); ''; end
end

class B
  extend T::Sig

  # Docs for B#foo
  sig {params(y: T.any(Integer, String)).returns(Integer)}
  def foo(y); 0; end
end

def example(x)
  x.foo(0)
  # ^ hover here
  # it shows the method def twice
end
```

The hover result here shows this:

```

# typed: true
extend T::Sig

class A
  extend T::Sig

  # Docs for A#foo
  sig {params(x: Integer).returns(String)}
  def foo(x); ''; end
end
```

which makes it clear that

- the dispatch aggregated the result of calling two different methods
- the argument types are different

IMO, the problem here is that we're using the symbol to get the argument types for each individual method, while we're using the dispatch result's type for the `returns` in all the sigs.

In the original issue, I suggested two implementations:

(1) Make the two methods clear, showing something like this:

```
sig {returns(Integer)}
sig {returns(String)}
```

or

```
sig {params(x: Integer).returns(String)}
sig {params(y: T.any(Integer, String)).returns(Integer)}
```

(2) Attempt to combine all argument and return types pairwise. I have a slight preference for not doing this one, because there's no precedent for it in Sorbet. But you could imagine attempting to do something like generate one single sig that lists something like

```
sig {returns(T.any(String, Integer)}
```

or

```
sig {params(y: T.any(Integer, String)).returns(T.any(Integer, String))}
```

by using the `secondaryKind` to either union or intersect the types of all arguments pairwise. That gets kind of hairy though.

(3) A third option I only just thought of would be to basically combine the two previous ones. For methods, we could show all the individual sigs of the methods that participated in the dispatch **and** the `resultType` of the dispatch, as a separate field:

```
sig {params(x: Integer).returns(String)}
sig {params(y: T.any(Integer, String)).returns(Integer)}

-> T.any(Integer, String)
```

It's also worth noting: the hover result currently **only** shows docs for the first method in the dispatch result. `Docs for B#foo` does not show up in the hover result, when arguably it should.

jez:
> I'm wondering if `^ hover:` would benefit from additional syntax that denotes the line breaks, or even a `^ hover-complete:` to check the full message.

If you do decide to build this, would you mind splitting it into a separate PR?

bitwise-aiden:
cc @jez 

> In the original issue, I suggested two implementations:

Ahh, seeing the examples laid out like this makes a lot more sense. I agree with the third approach as it doesn't trample the original signatures AND shows what to expect as possible returns for that given input. The not trampling is really helpful in the case of: 
``` ruby 
sig { T.params(x: Integer).returns(Integer) }
sig { T.params(x: String).returns(String) }
```
Though that does raise in our case where the receiver is a `T.any(A, B)`.

I think your logic still stands and I will update the code to reflect option 3. 

> It's also worth noting: the hover result currently only shows docs for the first method in the dispatch result. Docs for B#foo does not show up in the hover result, when arguably it should.

I'd noticed that too. I was going to follow up with a separate PR to address this as well. 

> If you do decide to build this, would you mind splitting it into a separate PR?
👍🏻 Yup, will do!

jez:
> The not trampling is really helpful in the case of:
> 
> ```ruby
> sig { T.params(x: Integer).returns(Integer) }
> sig { T.params(x: String).returns(String) }
> ```

It's worth noting that it's not possible for such a method to be called by the user. There are no types that are both `Integer` or `String`. Incidentally, taking the intersection of those two types would yield `T.noreturn` ([docs](https://sorbet.org/docs/noreturn#detecting-impossible-to-call-methods)). Though even if we did take option 2 and show one sig with `T.noreturn` for this example, it might actually just cause more confusion. So maybe showing that these are the two methods (individually) and that they have incompatible types, is the best way to clear the confusion.

main/lsp/requests/hover.cc
```diff
@@ -14,22 +14,35 @@ namespace sorbet::realmain::lsp {
...
+            auto localRetType = start->returnType;
+            auto &constraint = component.constr;
+            if (constraint) {
+                localRetType = core::Types::instantiate(gs, localRetType, *constraint);
             }
-            contents = absl::StrCat(
-                contents, prettyTypeForMethod(gs, component.method, component.receiver, retType, constraint.get()));
+
+            contents.push_back(
+                prettySigForMethod(gs, method, component.receiver, localRetType, constraint.get()));
```
bitwise-aiden, line 32: I've added this as a commit I intend on reverting, but helps to illustrate my point better. 

With aiming for the hover of the following, I've come across some issues and would love some feedback/insight:
```
sig {params(x: Integer).returns(String)}
sig {params(y: T.any(Integer, String)).returns(Integer)}

-> T.any(Integer, String)
```

My first approach was to pass `component.method.data(gs)->resultType` here, but this broke other parts of the code that were relying on instantiation of the type. A couple of examples of that being:
```
sig {params(a: Integer).returns(T.type_parameter(:A))}
sig {params(arg0: Integer, arg1: String).returns(Proc2::Return)}
```

Switching to the above mitigates that for those cases (the same as using `retType` passed in). However in the union case we will get the following result: 
```
sig {returns(T.any(String, Integer))}
sig {returns(Integer)}
```
This occurs because `Types::Instantiate` will create the fully realized type including all of the nested methods. 

Is there an existing way to solve for both of these cases? Ideally we would have generics, procs, etc be realized, but that nested types aren't considered as a part of that. It feels like we need to be able to solve for both since you could have a generic as a part of a union, in which case you want to print its type, then it's nested type in the `-> <type>` portion.

bitwise-aiden:
I've rebased on https://github.com/sorbet/sorbet/pull/6691 to get the `hover-line` test functionality.

main/lsp/requests/hover.cc
```diff
@@ -14,22 +14,34 @@ namespace sorbet::realmain::lsp {
...
-    return contents;
+    if (dispatchResult.main.method.exists()) {
+        if (contents.size() > 1) {
+            contents.push_back(fmt::format("\n-> {}", retType.show(gs)));
+        }
+
+        contents.push_back(prettyDefForMethod(gs, dispatchResult.main.method));
+    }
+
+    return absl::StrJoin(contents, "\n");
```
bitwise-aiden, line 44: Still need to remove the duplicates from the generated list, but would love some feedback on this as an approach.

bitwise-aiden:
Closing as I've lost context on this and if I do end up taking this on, I'd likely start from scratch.

PR #6689, opened 2023-01-30
Revert checksums change after github is reverting on their end

aadi-stripe:
### Motivation

See https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/

We had to make a change to expected checksums for repo archives after Github changed the compression, but turns out they are reverting it.

### Test plan

See included automated tests.


froydnj (approved):
r=me if it passes CI.

PR #6690, opened 2023-01-31
Add signature for `Class#attached_object` method

egiurleo:
This PR adds a sig for the `Class#attached_object` method, which was introduced in Ruby 3.2. This method will return the object for which the receiver is a singleton class, and if the receiver is not a singleton class, it will raise a `TypeError`.

For more information, see https://github.com/ruby/ruby/pull/6450 and https://bugs.ruby-lang.org/issues/12084.

### Test plan
See included automated tests.

### DIFF
--- a/rbi/core/class.rbi
+++ b/rbi/core/class.rbi
@@ -89,6 +89,23 @@ class Class < Module
   sig {returns(T.untyped)}
   def allocate(); end
 
+
+  # Returns the object for which the receiver is the singleton class.
+  #
+  # Raises a `TypeError` if the class is not a singleton class.
+  #
+  # ```ruby
+  # class Foo; end
+  #
+  # Foo.singleton_class.attached_object        #=> Foo
+  # Foo.attached_object                        #=> TypeError: `Foo' is not a singleton class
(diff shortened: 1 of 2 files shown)


rbi/core/class.rbi
```diff
@@ -89,6 +89,23 @@ class Class < Module
...
+  # ```ruby
+  # class Foo; end
+  #
+  # Foo.singleton_class.attached_object        #=> Foo
+  # Foo.attached_object                        #=> TypeError: `Foo' is not a singleton class
+  # Foo.new.singleton_class.attached_object    #=> #<Foo:0x000000010491a370>
+  # TrueClass.attached_object                  #=> TypeError: `TrueClass' is not a singleton class
+  # NilClass.attached_object                   #=> TypeError: `NilClass' is not a singleton class
+  # ```
+  sig { returns(Object) }
```
jez, line 106:
```ruby
x = BasicObject.new

# these are the same
x.singleton_class.attached_object
x
```

So I think this should have type `BasicObject` (or `T.untyped`, if you prefer).

egiurleo: Good point, thanks!

jez (approved):

PR #6691, opened 2023-01-31
Add HoverMultilineAssertion functionality

bitwise-aiden:
### Motivation
Off the back of a discussion on [another PR of mine](https://github.com/sorbet/sorbet/pull/6687). It was suggested (by this comment) to add a new assertion for targeting specific line numbers of hovers. This would allow that PR to ensure that no duplicate signatures were present. 

Without it, these two cases would be valid and not fail: 
```ruby 
# target sig 
sig {returns(String)}
def foo; end
sig {returns(String)}
def foo; end
---
Docs

# hover test
foo
# ^ hover: sig {returns(String)
# ^ hover: def foo; end
# ^ hover: ---
# ^ hover: Docs
```
This test case doesn't account for the fact that the signature appears twice. 

```ruby 
# target sig 
sig {returns(String)}
def foo; end
---
Docs

# hover test
foo
# ^ hover: sig {returns(String)
# ^ hover: def foo; end
# ^ hover: sig {returns(String)
# ^ hover: def foo; end
# ^ hover: ---
# ^ hover: Docs
```
On the inverse of the above test, this test checks for two signatures, but doesn't acknowledge that it has already been checked, thus double checking the same sig and passing still. 

The hover multiline check solves this by adding a line number to the start of the declaration to specifically check if that line number matches up with the expectation. Thus the first test would now look like: 
```ruby 
# target sig 
sig {returns(String)}
def foo; end
sig {returns(String)}
def foo; end
---
Docs

# hover test
foo
# ^ hover-multiline: 0 sig {returns(String)
# ^ hover-multiline: 1 def foo; end
# ^ hover-multiline: 2 ---
# ^ hover-multiline: 3 Docs
```

### Test plan

See included automated tests. I've updated an existing test to use this multiline behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -587,6 +587,13 @@ If a location should report the empty string, use the special label `(nothing)`:
 # ^ hover: (nothing)
 ```
 
+Assert the contents of a specific line of the hover response with `hover-line` assertions:
+
+```ruby
+  a = 10
+# ^ hover-line: 1 Integer(10)
+```
+
 #### Testing completion
 
 LSP tests can also assert the contents of completion responses with `completion`
--- a/test/helpers/position_assertions.cc
+++ b/test/helpers/position_assertions.cc
@@ -51,6 +51,7 @@ const UnorderedMap<
         {"assert-fast-path", FastPathAssertion::make},
         {"assert-slow-path", BooleanPropertyAssertion::make},
         {"hover", HoverAssertion::make},
+        {"hover-line", HoverLineAssertion::make},
         {"completion", CompletionAssertion::make},
         {"apply-completion", ApplyCompletionAssertion::make},
         {"apply-code-action", ApplyCodeActionAssertion::make},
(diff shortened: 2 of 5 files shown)


test/helpers/position_assertions.cc
```diff
@@ -51,6 +51,7 @@ const UnorderedMap<
         {"assert-fast-path", FastPathAssertion::make},
         {"assert-slow-path", BooleanPropertyAssertion::make},
         {"hover", HoverAssertion::make},
+        {"hover-multiline", HoverMultilineAssertion::make},
```
bitwise-aiden, line 54: Double checking this name makes the most sense?

test/helpers/position_assertions.cc
```diff
@@ -1232,6 +1233,80 @@ string HoverAssertion::toString() const {
...
+    for (auto assertion : assertions) {
+        if (auto assertionOfType = dynamic_pointer_cast<HoverMultilineAssertion>(assertion)) {
+            assertionOfType->check(sourceFileContents, wrapper, nextId, errorPrefix);
+        }
+    }
+}
+
+// Returns `true` if `line` matches a full line of text in `text`.
+bool containsLineAt(string_view text, string_view line, int lineno) {
+    vector<string> lines = absl::StrSplit(text, "\n");
```
bitwise-aiden, line 1274: Should this helper: 
- Exist? Does it make more sense to have this inline for the check?
- Take in the lines pre-split? That way we can split once and not have to repeat the same split across the string.

test/helpers/position_assertions.cc
```diff
@@ -1232,6 +1233,80 @@ string HoverAssertion::toString() const {
...
+    // Match a full line. Makes it possible to disambiguate `String` and `T.nilable(String)`.
+    if (!containsLineAt(hoverContents, this->message, this->lineno)) {
+        auto sourceLine = getSourceLine(sourceFileContents, filename, range->start->line);
+        ADD_FAIL_CHECK_AT(
+            filename.c_str(), range->start->line + 1,
+            fmt::format("{}Expected hover contents:\n{}\nFound hover contents:\n{}", errorPrefix,
+                        prettyPrintRangeComment(sourceLine, *range, toString()),
+                        prettyPrintRangeComment(sourceLine, *range,
+                            fmt::format("hover-multiline: {}", hoverContents))));
+    }
```
bitwise-aiden, line 1303: I'm unsure of this failure, currently it looks like this: 
```
===============================================================================
test/lsp_test_runner.cc:525:
TEST CASE:  LSPTest

test/testdata/lsp/hover.rb:83: ERROR: Expected hover contents:
83:   def tests_return_markdown
          ^^^^^^^^^^^^^^^^^^^^^ hover-multiline: 12 Tests return markdown output
Found hover contents:
83:   def tests_return_markdown
          ^^^^^^^^^^^^^^^^^^^^^ hover-multiline: ```ruby
sig {void}
def tests_return_markdown; end
```

---

Tests return markdown output

===============================================================================
```

I'm wondering if in targeting the specific line, it makes more sense to have `<lineno> <contents>` where `contents` is the actual value of that expected line or empty.

File: test/helpers/position_assertions.cc (hunk 1 above)
Comment (jez) [reply test/helpers/position_assertions.cc:54]: What do you think about `hover-line` or `hover-at-line`? (Given the name I thought that a single hover-multiline assertion would accept a multiline string somehow.)

File: test/helpers/position_assertions.cc (hunk 3 above)
Comment (bitwise-aiden) [reply test/helpers/position_assertions.cc:1303]: Thinking about it, it's helpful to be able to see the complete output on test failure.

File: README.md [hunk 4]
```diff
@@ -587,6 +587,13 @@ If a location should report the empty string, use the special label `(nothing)`:
 # ^ hover: (nothing)
 ```
 
+Assert the contents of a specific line of the hover response with `hover-line` assertions:
+
+```ruby
+  a = 10
+# ^ hover-line: 0 Integer(10)
```
bitwise-aiden, line 594: Does it make more sense to be `0` or `1` indexed?

jez: I think most editors use 1-based line indexing, so IMO 1-based is what I'd prefer.

test/helpers/position_assertions.cc
jez: What about "Found" -> "Complete" and "Expected hover contents" -> "Expected line N of hover contents"

jez (approved):

README.md
bitwise-aiden: Awesome, that's where I was leaning too. Will update now.

test/helpers/position_assertions.cc
bitwise-aiden: After adding the require checks for `lineno`, it made more sense to fold this into the `check` method.

PR #6692, opened 2023-01-31
Update RBI for URI::WS

KaanOzkan:
`URI::WS` inherits from `URI::Generic`: https://github.com/ruby/uri/blob/master/lib/uri/ws.rb#L22

### Motivation

Tapioca correctly generated the RBI as `class URI::WS < ::URI::Generic` which clashes with this payload.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_NH09SHabCJIm3r
→ https://go/builds/bui_NH09w8z3EuDAA9

jez (approved):

PR #6693, opened 2023-01-31
Codegen the T::Struct initializer

jeffcarbs:
Similar to #2683, we codegen the T::Struct initializer so that instead of initializing variables by looping through props at runtime, we generate code at load time (or earlier if using eager loading). This PR was extracted from #6684 since that PR was trying to do a little too much.

### Motivation

T::Struct initialization can be slow. Some of the performance issues can be attributed to typechecking but it's also slow due to the metaprogramming involved in simply initializing the variables. This PR generates the code

This produces a 1.5-2x speedup in the benchmarks.

**Before**
```
Plain Ruby (μs/iter):
  0.456503   0.005289   0.461792 (  0.463126)
T::Props.new, mostly nil input (μs/iter):
  3.291946   0.010572   3.302518 (  3.310399)
T::Props.new, all props set (μs/iter):
  6.529546   0.028695   6.558241 (  6.569688)
```

**After**
```
Plain Ruby (μs/iter):
  0.453986   0.001555   0.455541 (  0.455830)
T::Props.new, mostly nil input (μs/iter):
  1.778522   0.004675   1.783197 (  1.784899)
T::Props.new, all props set (μs/iter):
  4.454697   0.011147   4.465844 (  4.469799)
```

### Test plan

See included automated tests. Similar to the serde PR, I added `T::Props::GeneratedCodeValidation.validate_initialize` to verify the code in the initialize method is what we expect.


jeffcarbs:
@froydnj - I think this is ready for a look. I rebased onto the latest master to pull in those specs that were backfilled locking in the weak/strong behavior. I also refactored some things a bit to make it look more like the deserialize generator.

froydnj:
Failing testcase from the compiler testsuite:

```ruby
# frozen_string_literal: true
# typed: true

class A < T::Struct
  prop :int, Integer
  prop :str, String
  prop :optint, Integer, default: 5
  prop :optarray, T::Array[Integer], default: [6]
  prop :optstr, String, default: "shared string"
end

# Verify default arguments don't share structure.
x = A.new(int: 5, str: "foo")
y = A.new(int: 6, str: "bar")
p x.optarray.object_id != y.optarray.object_id
# ...but it's OK if the strings do, given that we're using frozen_string_literal.
p x.optstr.object_id == y.optstr.object_id
```

Before:

```
st-froydnj2:sorbet froydnj$ ruby -r ./gems/sorbet-runtime/lib/sorbet-runtime.rb ../struct-sharing.rb
true
true
```

After:

```
st-froydnj2:sorbet froydnj$ ruby -r ./gems/sorbet-runtime/lib/sorbet-runtime.rb ../struct-sharing.rb
true
false
```

froydnj:
This also fails with on many tests in Stripe's codebase (we only report the first 100 errors, so I don't have a good sense of how far it made it..but it's also failing in the abstractions around credit cards, so I imagine that it didn't make it very far).

I'll try to reduce a testcase, but I probably won't have it until next week.

froydnj:
http://go/builds/bui_NKLtkyWuMKi8E0 for any interested Stripe.

jeffcarbs:
@froydnj - Pushed up a commit to address the object equality issue (and added a test case for it in sorbet-runtime to confirm). The issue was that for any string defaults we were just putting the string literal in the source. Now we check if the string is frozen and use `"#{string.inspect}.freeze"` if so. The code around defaults was adapted from the deserialize generator and I found a test that actually locks in the opposite behavior, i.e. string defaults, even frozen, do not "share structure". Seems like we may want to fix that behavior, since if your default is an immutable string but the struct is initialized with a mutable string that could be dangerous: https://github.com/sorbet/sorbet/blob/4dd7dee1208155cb5b1ead8bca7abc58d2e4863d/gems/sorbet-runtime/test/types/props/serializable.rb#L1224-L1231.

gems/sorbet-runtime/lib/types/props/private/initializer_generator.rb
```diff
@@ -0,0 +1,183 @@
...
+        .checked(:never)
+      end
+      private_class_method def self.generate_missing_handler(prop:, rules:, ivar_name:, need_nil_write_check:, weak:)
+        if rules.key?(:factory)
+          ["self.class.class_exec(&#{decorator_fetch(prop, :factory)})", true]
+        elsif rules.key?(:default)
+          default = rules.fetch(:default)
+
+          literal_default = case default
+          when Integer, Symbol, Float, TrueClass, FalseClass, NilClass
```
froydnj, line 101: Why is this not doing something more like what `generate_nil_handler` for deserialization does?  Seems like adopting that structure would enable you to deal more elegantly with e.g arrays and hashes.  I realize that for certain things, going through `self.class.decorator.props_with_defaults.fetch(#{prop.inspect}).default` would probably be slower than what you have here (e.g. `T::Enum`), but it would probably be more correct?

gems/sorbet-runtime/test/types/props/optional.rb
```diff
@@ -32,6 +35,24 @@ def self.prop2_source
...
+    assert_same(m1.trueprop, m2.trueprop)
+    assert_same(m1.falseprop, m2.falseprop)
+    assert_same(m1.default_integer, m2.default_integer)
+
+    # Equal, but different object
+    refute_same(m1.default_array, m2.default_array)
+    refute_same(m1.default_mutable_string, m2.default_mutable_string)
+    assert_equal(m1.default_array, m2.default_array)
+    assert_equal(m1.default_mutable_string, m2.default_mutable_string)
+  end
```
froydnj, line 54: This test would be a great thing to split out into a separate PR, so it's easy to see that a) this is the current behavior; and b) this PR doesn't change existing behavior.

gems/sorbet-runtime/test/types/props/constructor.rb
```diff
@@ -145,4 +145,76 @@ def initialize(hash={})
...
+      it 'has meaningful validation which complains at lurky method invocation' do
+        src = ComplexStruct.decorator.send(:generate_initialize_source).to_s
+        src = src.sub(/\.deep_clone_object\b/, '.something_suspicious')
+        assert_raises(T::Props::GeneratedCodeValidation::ValidationError) do
+          T::Props::GeneratedCodeValidation.validate_initialize(src)
+        end
+      end
+    end
+
+    describe 'disabling evaluation' do
```
froydnj, line 197: I think there also needs to be a test that:

* defines props, then defines `initialize`
* defines `initialize`, then defines props
* defines props, then defines `initialize`, then defines more props

I think the correct behavior would be to call the user's `initialize` in all cases.  (I think the right thing already happens?  But I'm not entirely sure.)  This would be another good candidate for a separate PR.

gems/sorbet-runtime/test/types/props/constructor.rb
```diff
@@ -145,4 +145,76 @@ def initialize(hash={})
     end
     assert_equal("Missing required prop `name` for class `Opus::Types::Test::Props::ConstructorTest::CustomInitializeStruct`", err.message)
   end
+
+  class ComplexStruct < T::Struct
+    prop :primitive, Integer
+    prop :nilable, T.nilable(Integer)
```
froydnj, line 151: This is implicitly relying on other tests to verify that initialization is handled correctly, is that correct?

I think there needs to be a `T.deprecated_enum` type in here somewhere.

froydnj (changes requested):
There are some trivial Rubocop issues to resolve, as well as some comments below.

jeffcarbs:
@froydnj - thanks again for the review! I pulled most of those test changes into those separate PRs that have already been merged to ensure we're locking in existing behavior. The only new test changes here are adding the validation for the generated code.

I got all the tests and rubocop passing. I also updated the description with the latest benchmarks, some info about some commits where I backed out some custom code, and an example of the generated input. Would you be able to trigger the tests and maybe run a build on the stripe codebase to see what issues surface?

froydnj:

→ https://go/builds/bui_PRnETGCnZQU4K5
→ https://go/builds/bui_PRnEuUPTdDRplM
→ https://go/builds/bui_PRnEF9hBiODQAM

→ https://go/builds/bui_PRnM5RpL8Y2rMk

froydnj (changes requested):
There are a ton of errors with...unhelpful error messages, but what I think is happening is that Stripe has code that freezes prop-using classes and when we instantiate the class, we attempt to define `initialize` on the frozen class, which raises.

I suspect that we have the same sort of issues with the de/serialization codegen, but that's never come up because we never serialize the relevant classes.

I don't think this PR has to deal with making de/serialization codegen work in the face of such classes, but it does have to deal with making `initialize` work for such things.

froydnj:
I asked around and it looks like we might not be freezing those classes @jeffcarbs .  So you may want to hold off fixing that for just a bit.

froydnj:

→ https://go/builds/bui_PTEnQnRhPSyM5e
→ https://go/builds/bui_PTEnKL8bFwAJxb
→ https://go/builds/bui_PTEnMFpFOkznqy

→ https://go/builds/bui_PTEuPqzIj5Oi9i

froydnj:
It looks like there's a test failure where a constructor is not raising a `RuntimeError`.  It will take me a little while to track this one down.

jeffcarbs:
OK sounds good. Let me know if there's anything I can do to help.

froydnj:

→ https://go/builds/bui_PU36OeSoXEM87q
→ https://go/builds/bui_PU36P2Y1Nm5He2
→ https://go/builds/bui_PU36KZ7VW93UWd

→ https://go/builds/bui_PU3GaujPwh7k7c

froydnj:
I thought of something else to test, which actually passed without issues (which is a good thing!).  Would you mind adding something like this:

```ruby
    describe 'does not overwrite user-declared initialize' do
      it 'works' do
        m = Class.new do
          include T::Props::WeakConstructor

          prop :foo, T.nilable(String)

          def initialize(foo)
            raise "raising with #{foo}"
          end
        end

        m.decorator.eagerly_define_lazy_methods!

        ex = assert_raises(StandardError) do
          m.new('secret value')
        end

        assert_equal('raising with secret value', ex.message)
      end
    end
```

The PR seems to be performance-neutral on Stripe's test suite (not that this is bad, just that the lazy initialization bits are not blocking anything important, and we would force evaluation of the lazy bits in production anyway).

jeffcarbs:
Just pushed up a commit to add that test and rebased onto the latest upstream to pull in the change from #7643. 

The performance bump here is definitely pretty minor, but this is a precursor to support #6667. Currently the runtime type checks are the mechanism by which missing required props are identified, so disabling those type checks would break that behavior. There's no great way to change that in the current world without introducing more performance overhead.

After this PR, if you stop checking types for all props, the performance is close to the PORO:
```
Plain Ruby (μs/iter):
  0.435867   0.000729   0.436596 (  0.436767)
T::Props.new, mostly nil input (μs/iter):
  1.002338   0.002045   1.004383 (  1.005056)
T::Props.new, all props set (μs/iter):
  5.290807   0.029587   5.320394 (  5.348496)
T::Props.new, all props set, checked(:never) (μs/iter):        <-- [NEW]
  0.605002   0.001626   0.606628 (  0.607012)
```

PR #6695, opened 2023-02-01
[Ruby3] Error on implicit keyword to hash conversion

ilyailya:
Ruby 3 handles keyword arguments differently than Ruby 2.7. The old way of passing keyword args is considered an incorrect syntax in newer ruby. More details are available here: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0

The change emits an error then kwarg is implicitly converted to hash. See a test for examples

### Motivation

Ruby 3!

### Test plan

See included automated tests.


core/types/calls.cc
```diff
@@ -35,6 +35,14 @@ bool allComponentsPresent(DispatchResult &res) {
     }
     return allComponentsPresent(*res.secondary);
 }
+
+inline bool isTyped(const TypePtr type) {
+    return type && !type.isUntyped();
```
jez, line 40: Why is the nullptr check required?

ilyailya: I can't recall the exact scenario, but I got a crash running the change over pay-server. I've rerun in now without nullptr check, and it didn't crash this time. It should be safe to remove it know

test/testdata/infer/ruby3_keyword_args.rb
```diff
@@ -11,3 +11,224 @@ def takes_kwargs(x, y:)
 takes_kwargs(99, arghash) # error: Keyword argument hash without `**` is deprecated
 
 takes_kwargs(99, **arghash)
+
+sig {params(h: T::Hash[T.untyped, T.untyped], kw: T::Hash[T.untyped, T.untyped]).void}
+def takes_only_hash_and_kwargs(h, **kw); end
+
+takes_only_hash_and_kwargs(a: "x")
+                        #  ^^^^^^ error: Passing the keyword argument as the last hash parameter is deprecated
```
jez, line 19: I have two concerns with this:

1. This error message is different from what the Ruby VM prints. On this snippet, the Ruby VM says "wrong number of arguments (given 0, expected 1)"
2. It seems that the way we're reporting this error makes this do something weird:

```ruby
def foo(h, a:, **kw); end
foo(a: "x")
```

Ruby: `wrong number of arguments (given 0, expected 1; required keyword: a)`
this PR: `Missing required keyword argument a for method`

core/types/calls.cc
```diff
@@ -962,6 +970,107 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                    }
+                }
+            }
+
+            if (!allArgsProcessed &&
+                (
+                    // there are keyword args remaining and we expect the current argument to be a hash,
+                    // but it's not a hash => report implicit conversion
+                    (isTyped(probablyFirstKwarg.type) && isHash(gs, probablyFirstKwarg.type) &&
+                     isTyped(args.args[maybeKwargIdx]->type) && !isHash(gs, args.args[maybeKwargIdx]->type) &&
```
jez, line 1049: It seems odd to me that we might ask `isHash` on a value that is just the inlined `Symbol` argument corresponding to a keyword arg's name.

jez (changes requested):
I have two big concerns with this change so far

1. see my comments on the test
2. [this code](https://github.com/sorbet/sorbet/blob/master/core/types/calls.cc#L1053-L1054) which handles automatically converting keyword args to a hash arg still runs all the time (even if the error we're starting to report in this change fires)

Those two combine leading me to believe that this isn't quite ready yet. You can look at the other comments I have, but I think this change might need to be reworked.

core/types/calls.cc
```diff
@@ -962,6 +970,107 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
     }
 
+    {
+        const ArgInfo &probablyFirstKwarg = *pit;
+        if (gs.ruby3KeywordArgs && args.args.size() > 0 &&
+            // All positional arguments are already processed
+            ait == aPosEnd &&
+            // But not all arguments are processed, some keyword arguments remain
+            ait != aend &&
+            // First keyword parameter doesn't look like keyword param
```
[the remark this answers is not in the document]
ilyailya: >Sorry for being pedantic here—one thing we try to follow in this method is calling the call-site arguments "arguments" and the definition-site parameters "parameters"

I finally learned what's the difference between those, thanks to the comments at lines 914 and 923!

I tried to stay consistent regarding args/params, but you're right with `probablyFirstKwparam`. I've also explained my thought process regarding that name in the previous comment.

core/types/calls.cc
```diff
@@ -962,6 +970,107 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+            !probablyFirstKwarg.flags.isKeyword && !probablyFirstKwarg.flags.isRepeated &&
+            !probablyFirstKwarg.flags.isBlock) {
+            auto isLastParam /* not counting last block param */ =
+                (pit - data->arguments.begin()) >= data->arguments.size() - 2;
```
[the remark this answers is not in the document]
ilyailya: >ENFORCE(pit != pend && pit + 1 != pend);

I don't fully understand the enforce. I suspect we might get false-negatives by cutting away cases where *pit points to the last two arguments. 
And also running the test crashes on this enforce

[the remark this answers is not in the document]
jez: sorry i think this was meant to be an `||`

core/types/calls.cc
```diff
@@ -962,6 +970,107 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                }
+            }
+
+            auto maybeKwargIdx =
+                ait - args.args.begin() < args.args.size() ? ait - args.args.begin() : args.args.size() - 1;
+            auto allArgsProcessed = false;
+
+            // `hasKwsplat` can be false positive, because it relies on how we parse positional and keyword args
+            auto actuallyHasKwsplatParam = absl::c_any_of(
+                data->arguments, [](const auto &arg) { return arg.flags.isKeyword && arg.flags.isRepeated; });
```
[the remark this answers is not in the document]
ilyailya: >I don't know whether it's correct to say that "hasKwsplat can be false positive"

It was tricky to figure out what `hasKwsplat` actually did. Firstly, because of the name confusion you've mentioned. I legitimately thought it was about param list, not an arg list.

Secondly, I noticed this:
```ruby
sig {params(a: Integer, h: T::Hash[T.untyped, T.untyped]).void}
def takes_empty_hash(a, h); end
empty_hash = {}
blk = Proc.new {puts "in proc"}

takes_empty_hash(1, **empty_hash) # => hasKwsplat = true
takes_empty_hash(1, **empty_hash, &blk) # => hasKwsplat = false
```

My expectation is `hasKwsplat` should be positive in both cases

test/testdata/infer/ruby3_keyword_args.rb
ilyailya: >This error message is different from what the Ruby VM prints. On this snippet, the Ruby VM says "wrong number of arguments (given 0, expected 1)"

Hmm, this is weird, I'm not seeing this. Which version of ruby are you using? It should report that error only in ruby 3. This is what I see

```bash
stripe/sorbet [iz/ruby-3-kwargs●] » cat /tmp/x.rb
# typed: true

require 'sorbet-runtime'
extend T::Sig
Warning[:deprecated] = true
sig {params(h: T::Hash[T.untyped, T.untyped], kw: T::Hash[T.untyped, T.untyped]).void}
def takes_only_hash_and_kwargs(h, **kw); end

takes_only_hash_and_kwargs(a: "x")


stripe/sorbet [iz/ruby-3-kwargs●] » ruby --version
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [arm64-darwin21]


stripe/sorbet [iz/ruby-3-kwargs●] » ruby /tmp/x.rb
/Users/iz/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.10201/lib/types/private/methods/call_validation.rb:161: warning: Passing the keyword argument as the last hash parameter is deprecated
/tmp/x.rb:8: warning: The called method `takes_only_hash_and_kwargs' is defined here
```

ilyailya:
```ruby
def foo(h, a:, **kw); end
foo(a: "x")
```
that looks like invalid ruby in 2.7

```bash
stripe/sorbet [iz/ruby-3-kwargs●] » cat /tmp/x.rb
# typed: true

require 'sorbet-runtime'
extend T::Sig
Warning[:deprecated] = true

def foo(h, a:, **kw); end
foo(a: "x")

stripe/sorbet [iz/ruby-3-kwargs●] » ruby /tmp/x.rb
/tmp/x.rb:9: warning: Passing the keyword argument as the last hash parameter is deprecated
/tmp/x.rb:8: warning: The called method `foo' is defined here
        1: from /tmp/x.rb:9:in `<main>'
/tmp/x.rb:8:in `foo': missing keyword: :a (ArgumentError)
```

ilyailya: And here is what we report from PR:

```
stripe/sorbet [iz/ruby-3-kwargs●] » sr --experimental-ruby3-keyword-args /tmp/x.rb
Exception::raise(): core/types/subtyping.cc:1076 enforced condition t1 != nullptr has failed: (no message provided)

Backtrace:
...

/tmp/x.rb:9: Failed to type (backtrace is above) https://srb.help/1001
     9 |foo(a: "x")
        ^^^^^^^^^^^

/tmp/x.rb:4: Failed to process tree (backtrace is above) https://srb.help/1001
     4 |require 'sorbet-runtime'
     5 |extend T::Sig
     6 |Warning[:deprecated] = true
     7 |
     8 |def foo(h, a:, **kw); end
     9 |foo(a: "x")

/tmp/x.rb: Exception in cfg+infer: /tmp/x.rb (backtrace is above) https://srb.help/1001
Errors: 3
```

I'll look into that crash

core/types/calls.cc
ilyailya: We can replace it with a check for is it a `Symbol`

core/types/calls.cc
```diff
@@ -962,6 +970,107 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
         implicitKwsplat = true;
     }
 
+    {
+        const ArgInfo &probablyFirstKwarg = *pit;
+        if (gs.ruby3KeywordArgs && args.args.size() > 0 &&
+            // All positional arguments are already processed
+            ait == aPosEnd &&
+            // But not all arguments are processed, some keyword arguments remain
+            ait != aend &&
```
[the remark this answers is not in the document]
ilyailya: I'm good with changing the comment, but I'd prefer to keep `ait != aend` for the reasons mentioned in the comment about the `hasKwsplat`

core/types/calls.cc
```diff
@@ -962,6 +970,107 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                    continue;
+                }
+                if (!flags.isDefault) {
+                    areOtherParamsDefault = false;
+                    break;
+                }
+            }
+
+            auto maybeKwargIdx =
+                ait - args.args.begin() < args.args.size() ? ait - args.args.begin() : args.args.size() - 1;
```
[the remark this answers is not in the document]
ilyailya: It was originally a bound check, because I had a crash in `args.args[ ait - args.args.begin()]`. You're right, it's not needed anymore, I'll delete it

PR #6700, opened 2023-02-02
Move ENFORCE from LSP to resolver

jez:
### Motivation

This was a change I made while working on LSP fast path stuff.

This also skips a `for` loop that would otherwise have an empty body,
because the `sanityCheck` method guards the `for` loop by a `debug_mode`
check.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -4047,6 +4047,7 @@ void sanityCheck(const core::GlobalState &gs, vector<ast::ParsedFile> &trees) {
         ResolveSanityCheckWalk sanity;
         for (auto &tree : trees) {
             core::Context ctx(gs, core::Symbols::root(), tree.file);
+            ENFORCE(tree.tree);
             ast::TreeWalk::apply(ctx, sanity, tree.tree);
         }
     }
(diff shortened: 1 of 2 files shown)


resolver/resolver.cc
```diff
@@ -4047,6 +4047,7 @@ void sanityCheck(const core::GlobalState &gs, vector<ast::ParsedFile> &trees) {
         ResolveSanityCheckWalk sanity;
         for (auto &tree : trees) {
             core::Context ctx(gs, core::Symbols::root(), tree.file);
+            ENFORCE(tree.tree);
```
froydnj, line 4050: It seems weird to me that the earlier `ENFORCE` was checking the file and this is checking the tree, but 🤷‍♂️ .

froydnj (approved):

resolver/resolver.cc
jez: oh I didn't even notice that.

jez: I think something like `file.exists()` could only ever fire if we ran Stripe's codebase with debug checks—it's abundantly unlikely for us to write a test case that would trip that in e.g. a `test/testdata/`-style test. But I could see the nullptr `tree.tree` enforce firing, so I'm going to leave it as is.

PR #6706, opened 2023-02-07
don't reference count nesting scopes in resolver unnecessarily

froydnj:
### Motivation

We keep a (debug-only) metric on how many constants resolve in the "first pass", without needing a separate resolution step.  On a recent-ish debug build of Stripe's codebase, almost 99% of constants resolve on the first try.

However, the way that we do this first pass is that we create a `ConstantResolutionItem` and attempt to resolve it.  If we fail, that job can be pushed, as-is, onto the queue of things to be retried.  This temporary job takes shared references to the current nesting scope of `ResolveConstantsWalk` and the `firstDefinitionLocs` table that we now maintain after #6671.

Taking those shared references requires atomic reference counting, which is (relatively) expensive.  And the aforementioned 99% of the time, we don't need to take shared references to those two things. 

Hence this PR: we split out a separate interface for being able to resolve individual constants without taking those shared references and use it from the appropriate places.

I took measurements that #6671 slows resolver down by about 3%; this PR seems to win that back.  Without #6671 applied, it looked like this PR would win ~2%, but I guess there is some unavoidable overhead from that PR?  It's also possible that there's just some noise in the measurements and I happened to get the right combination of 10 runs with and without these changes.

cc @aadi-stripe 

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -845,30 +845,38 @@ private:
         }
     }
 
-    static bool resolveJob(core::Context ctx, ConstantResolutionItem &job) {
-        if (isAlreadyResolved(ctx, *job.out)) {
-            if (job.possibleGenericType) {
+    static bool resolveConstantJob(core::Context ctx, const shared_ptr<Nesting> &nesting,
+                                   ast::ConstantLit *out, bool &resolutionFailed, const bool possibleGenericType,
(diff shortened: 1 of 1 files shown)


jez (approved):
👏

PR #6710, opened 2023-02-08
Speed up move method by skipping incrementalResolve

jez:
### Motivation

This method is trying to get the first location in the file after the comments
at the top of the file. It does that by getting the location of the
`class <root>` tree in the file, which has the loc of the first piece of syntax
in the file. But it doesn't need to run incrementalResolve to get that--the same
loc is present in the indexed trees.

### Test plan

Existing tests


froydnj (approved):
Makes sense to me!

PR #6716, opened 2023-02-09
Invalidate cache on autoloader_compatibility change

bhang-stripe:
### Motivation

With path-based autoloading, we need to fully run `DependencyStep` when a package's `autoloader_compatibility` changes as the autoloader shims will change. Currently, it sometimes doesn't change the shims, leading to constants not loading correctly.

### Test plan

Added tests


bhang-stripe:

→ https://go/builds/bui_NKLXRPeCQMTUMH
→ https://go/builds/bui_NKLXBvQpmFzt8j

aadi-stripe (approved):

PR #6720, opened 2023-02-10
Ruby treats `initialize` as private by default

### LINKED ISSUE #3837: Explicitly declaring `initialize` `private` results in an error despite the fact that `initialize` is always `private` anyway
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams(x%3A%20Integer).void%7D%0A%20%20private%20def%20initialize(x)%20end%0Aend%0A%0AA.new(0)%20%20%0A%0Ap%20A.private_instance_methods(false).include%3F(%3Ainitialize)%0A%23%3D%3E%20true)

```ruby
# typed: true
class A
  extend T::Sig

  sig {params(x: Integer).void}
  private def initialize(x) end
end

A.new(0)  

p A.private_instance_methods(false).include?(:initialize)
#=> true
```

#### Observed output

```
editor.rb:9: Non-private call to private method initialize on A https://srb.help/7031
     9 |A.new(0)  
        ^^^^^^^^
    editor.rb:6: Defined in A here
     6 |  private def initialize(x) end
                  ^^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

This should not be an error, since it does not fail at runtime:

```sh
ruby editor.rb
# true
```

In fact, since [`#initialize` is *always* `private`](https://www.ruby-forum.com/t/private-initialize/154334/3) even when not explicitly declared so, unless *explicitly* being declared `protected` or `public`, this would essentially mean that *every* use of [`Class#new`](https://ruby-doc.org/core/Class.html#method-i-new) would *always* be an error, which is clearly not the case.

If I slightly modify the code, it passes:

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams(x%3A%20Integer).void%7D%0A%20%20def%20initialize(x)%20end%0Aend%0A%0AA.new(0)%20%20%0A%0Ap%20A.private_instance_methods(false).include%3F(%3Ainitialize)%0A%23%3D%3E%20true)

```ruby
# typed: true
class A
  extend T::Sig

  sig {params(x: Integer).void}
  def initialize(x) end
end

A.new(0)  

p A.private_instance_methods(false).include?(:initialize)
#=> true
```

Despite the fact that this code is 100% semantically identical to the first one, since `#initialize` is *still* `private` here anyway.

- - -

My *guess* is that Sorbet doesn't "know" about the special treatment of `initialize` in Ruby and thus treats it as a normal method. My second guess is that Sorbet contains a definition of `Class#new` that looks like this:

```ruby
class Class
  def new(...)
    obj = allocate
    obj.initialize(...)
    obj
  end
end
```

This means that as long as Sorbet *thinks* `initialize` is `public`, everything is honky-dory, but as soon as it *thinks* it is `private`, it fails.

However, the *actual* implementation of `Class#new` are more like this:

```ruby
class Class
  def new(...)
    obj = allocate
    obj.__send__(:initialize, ...)
    obj
  end
end
```

If my assumptions are correct, there are two bugs in Sorbet that *normally* cancel each other out:

* The definition of `Class#new` is wrong. This *would* mean that you would always get an error because `initialize` is always `private`, except 
* The default visibility for `initialize` is also wrong.

*However*, I have not checked those assumptions, so I could be completely off.

jez:
### Motivation

Fixes #3837

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -879,6 +879,19 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
             ? guessOverload(gs, symbol, mayBeOverloaded, args.numPosArgs, args.args, targs, args.block != nullptr)
             : mayBeOverloaded;
 
+    if (method.data(gs)->flags.isPrivate && !args.isPrivateOk) {
+        if (auto e = gs.beginError(errLoc, core::errors::Infer::PrivateMethod)) {
+            if (args.fullType.type != args.thisType) {
+                e.setHeader("Non-private call to private method `{}` on `{}` component of `{}`",
+                            method.data(gs)->name.show(gs), args.thisType.show(gs), args.fullType.type.show(gs));
+            } else {
+                e.setHeader("Non-private call to private method `{}` on `{}`", method.data(gs)->name.show(gs),
+                            args.thisType.show(gs));
+            }
+            e.addErrorLine(method.data(gs)->loc(), "Defined in `{}` here", method.data(gs)->owner.show(gs));
+        }
+    }
+
     DispatchResult result;
     auto &component = result.main;
     component.receiver = args.selfType;
@@ -3046,7 +3059,8 @@ public:
                     selfTyAndAnd.type,
                     args.block,
                     args.originForUninitialized,
-                    args.isPrivateOk,
+                    // We already reported one visibility error, if relevant
+                    /* isPrivateOk */ true,
                     args.suppressErrors,
                 };
                 auto retried = selfTyAndAnd.type.dispatchCall(gs, newInnerArgs);
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1647,7 +1647,7 @@ DispatchResult MetaType::dispatchCall(const GlobalState &gs, const DispatchArgs 
                                           wrapped,
                                           args.block,
                                           args.originForUninitialized,
-                                          args.isPrivateOk,
+                                          /* isPrivateOk */ true,
                                           args.suppressErrors};
             auto original = wrapped.dispatchCall(gs, innerArgs);
             original.returnType = wrapped;
@@ -2029,10 +2029,17 @@ public:
             }
         }
         auto instanceTy = attachedClass.data(gs)->externalType();
-        DispatchArgs innerArgs{Names::initialize(), args.locs,          args.numPosArgs,
-                               args.args,           instanceTy,         {instanceTy, args.fullType.origins},
-                               instanceTy,          args.block,         args.originForUninitialized,
-                               args.isPrivateOk,    args.suppressErrors};
+        DispatchArgs innerArgs{Names::initialize(),
+                               args.locs,
+                               args.numPosArgs,
+                               args.args,
+                               instanceTy,
+                               {instanceTy, args.fullType.origins},
+                               instanceTy,
+                               args.block,
+                               args.originForUninitialized,
+                               /* isPrivateOk */ true,
+                               args.suppressErrors};
         auto dispatched = instanceTy.dispatchCall(gs, innerArgs);
 
         for (auto &err : res.main.errors) {
@@ -2459,7 +2466,7 @@ class Magic_callWithBlock : public IntrinsicMethod {
 private:
     static TypePtr typeToProc(const GlobalState &gs, const TypeAndOrigins &blockType, core::FileRef file,
                               LocOffsets callLoc, LocOffsets receiverLoc, LocOffsets funLoc, Loc originForUninitialized,
-                              bool isPrivateOk, bool suppressErrors) {
+                              bool suppressErrors) {
         auto nonNilBlockType = blockType;
         auto typeIsNilable = false;
         if (Types::isSubType(gs, Types::nilClass(), blockType.type)) {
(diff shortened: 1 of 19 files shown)


jez:

→ https://go/builds/bui_NKghNnZvOEdMFh
→ https://go/builds/bui_NKghL1eghzCh55

jez:

→ https://go/builds/bui_NKipppEb3XewDl
→ https://go/builds/bui_NKipeHhbxp3tVm

jez:

→ https://go/builds/bui_NL65hjfYm6v3Mq
→ https://go/builds/bui_NL655pj6LkcptN

core/types/calls.cc
```diff
@@ -2906,6 +2926,7 @@ class Magic_selfNew : public IntrinsicMethod {
         if (self.data(gs)->isSingletonClass(gs) && dispatched.main.method.exists() &&
             (dispatched.main.method == core::Symbols::Class_new() ||
              dispatched.main.method.data(gs)->name == core::Names::initialize())) {
+            // TODO(jez) This doesn't handle when initialize is overloaded
```
jez, line 2929: This is a preexisting bug; going to circle back to this or make a ticket.

core/types/calls.cc
```diff
@@ -879,6 +879,19 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+            if (args.fullType.type != args.thisType) {
+                e.setHeader("Non-private call to private method `{}` on `{}` component of `{}`",
+                            method.data(gs)->name.show(gs), args.thisType.show(gs), args.fullType.type.show(gs));
+            } else {
+                e.setHeader("Non-private call to private method `{}` on `{}`", method.data(gs)->name.show(gs),
+                            args.thisType.show(gs));
+            }
+            e.addErrorLine(method.data(gs)->loc(), "Defined in `{}` here", method.data(gs)->owner.show(gs));
+        }
+    }
```
jez, line 893: Moving this snippet from environment.cc to here means that we can check privateness per `dispatchCall`, which means that if we ever make a recursive call to `dispatchCall` in some intrinsic, we have more control over whether calls to private methods are allowed.

This is important because

```ruby
class A
  def initialize; end
end

a = A.new # ok
a.initialize # not okay
```

And previously, the `Class_new` intrinsic would record the `A#initialize` symbol as "the thing that was dispatched to" in the DispatchResult, which meant that Sorbet couldn't easily tell in environment.cc whether the call to `initialize` happened at a location where private was ok (`A.new`) or not ok (`a.initialize`).

While I did this refactor, I audited all recursive calls to `dispatchCall` to see whether we should forward along the `isPrivateOk` from the call site to the recursive call. The Ruby VM allows basically all the implicit conversion methods to be called even if they're private, so you see some of those flip in this change (I didn't add tests for that, but I could if we think it's valuable).

packager/rbi_gen.cc
```diff
@@ -327,7 +327,7 @@ class RBIExporter final {
         if (methodData->flags.isPrivate) {
             if (methodData->owner.data(gs)->isSingletonClass(gs)) {
                 visibility = "private_class_method ";
-            } else {
+            } else if (methodData->name != core::Names::initialize()) {
```
jez, line 330: This is the main change, and is mimicked by the Ruby VM:

https://github.com/ruby/ruby/blob/6847ee089d7655b2a0eea4fee3133aeacd4cc7cc/vm_method.c#L546-L557

(They also have support for `initialize_{copy,clone,dup}` but Sorbet doesn't have any special handling for those yet.)

namer/namer.cc
```diff
@@ -1047,9 +1047,12 @@ class SymbolDefiner {
             return symbol;
         }
 
-        auto implicitlyPrivate = ctx.owner.enclosingClass(ctx) == core::Symbols::root();
+        // Methods defined at the top level default to private (on Object)
+        // Also, the `initialize` method defaults to private
+        auto implicitlyPrivate =
+            (ctx.owner.enclosingClass(ctx) == core::Symbols::root()) ||
+            (!symbol.data(ctx)->owner.data(ctx)->attachedClass(ctx).exists() && name == core::Names::initialize());
```
jez, line 1054: This is the main change, and is mimicked by the Ruby VM:

https://github.com/ruby/ruby/blob/6847ee089d7655b2a0eea4fee3133aeacd4cc7cc/vm_method.c#L546-L557

(They also have support for `initialize_{copy,clone,dup}` but Sorbet doesn't have any special handling for those yet.)

froydnj (approved):
I think we should have slightly more commentary on the `/* isPrivateOk */ true` changes, but otherwise, this is much simpler than I first imagined.

core/types/calls.cc
```diff
@@ -2483,7 +2503,7 @@ class Magic_callWithBlock : public IntrinsicMethod {
                                nonNilBlockType.type,
                                nullptr,
                                originForUninitialized,
-                               isPrivateOk,
+                               /* isPrivateOk */ true,
```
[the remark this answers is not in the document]
jez: Totally fair. There's a stacked change on this one, so I'll get to this feedback in a followup change.

core/types/calls.cc
jez: https://github.com/sorbet/sorbet/issues/6747

PR #6721, opened 2023-02-10
Allow covariant/contravariant generic classes

jez:
### Motivation

I finally convinced myself of the correctness of this change (with the aide of
@georg-stripe). No reason to artificially limit how variance is allowed anymore.

### Test plan

See included automated tests.


definition_validator/validator.cc
```diff
@@ -961,6 +961,9 @@ class ValidateWalk {
             variance::validateMethodVariance(ctx, methodDef.symbol);
         }
 
+        // See the comment on `VarianceValidator::validateMethod` for an explanation of why we don't
+        // need to check types on instance variables.
```
froydnj, line 965: Uber non-blocking nit: either change this to "See the comment in..." or move the comment inside `validateMethod` outside of the method.

froydnj (approved):
✅

definition_validator/validator.cc
jez: I'll queue this up for the followup changes I promised in #6720

PR #6724, opened 2023-02-11
generics: Autocorrect to convert `X(...)` to `X[...]`

### LINKED ISSUE #6723: Add autocorrect for `T::Array(String)` to `T::Array[String]`
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20returns%28T%3A%3AArray%28String%29%29%7D%0Adef%20foo%0A%20%20%5B%22sigh%22%5D%0Aend)

```ruby
# typed: true
extend T::Sig

sig { returns(T::Array(String))}
def foo
  ["sigh"]
end
```

#### Observed output

```
editor.rb:4: Unsupported method T.Array https://srb.help/5004
     4 |sig { returns(T::Array(String))}
                      ^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

- - -

jez:
### Motivation

Fixes #6723

### Test plan

See included automated tests.

### DIFF
--- a/resolver/type_syntax/type_syntax.cc
+++ b/resolver/type_syntax/type_syntax.cc
@@ -890,7 +890,8 @@ optional<TypeSyntax::ResultType> interpretTCombinator(core::Context ctx, const a
 
         default:
             if (auto e = ctx.beginError(send.loc, core::errors::Resolver::InvalidTypeDeclaration)) {
-                if (send.numPosArgs() > 0 && send.onlyPosArgs() && send.block() == nullptr && send.argsLoc().exists()) {
+                if (send.numPosArgs() > 0 && send.onlyPosArgs() && send.block() == nullptr && send.argsLoc().exists() &&
+                    ctx.locAt(send.funLoc).adjustLen(ctx, -1, 1).source(ctx) == ":") {
                     auto replacement =
                         fmt::format("T::{}[{}]", send.fun.show(ctx), ctx.locAt(send.argsLoc()).source(ctx).value());
                     e.setHeader("Did you mean to use square brackets: `{}`", replacement);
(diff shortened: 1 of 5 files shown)


resolver/type_syntax/type_syntax.cc
```diff
@@ -848,7 +890,15 @@ optional<TypeSyntax::ResultType> interpretTCombinator(core::Context ctx, const a
 
         default:
             if (auto e = ctx.beginError(send.loc, core::errors::Resolver::InvalidTypeDeclaration)) {
-                e.setHeader("Unsupported method `{}`", "T." + send.fun.show(ctx));
+                if (send.numPosArgs() > 0 && send.onlyPosArgs() && send.block() == nullptr && send.argsLoc().exists() &&
+                    ctx.locAt(send.funLoc).adjustLen(ctx, -1, 1).source(ctx) == ":") {
+                    auto replacement =
+                        fmt::format("T::{}[{}]", send.fun.show(ctx), ctx.locAt(send.argsLoc()).source(ctx).value());
```
froydnj, line 896: IIUC, this `argsLoc` encompasses all the args, but we don't have any multi-arg testcases (all the testcases right now are single-arg generics).  WDYT about adding some `T::Hash`-related testcases to make sure we do the right thing with multi-arg generics?

froydnj (approved):
Seems helpful!  One comment about tests below.

resolver/type_syntax/type_syntax.cc
jez: good call thanks!

technically it conflicts with another autocorrect to remove what Sorbet sees as a redundant argument to the `Kernel#Hash` method, but the user can choose which autocorrect to pick, so it luckily just works.

PR #6731, opened 2023-02-14
derive the type of transpose on arrays of tuples more correctly

froydnj:
### Motivation

@sushain97 wanted to know how hard this was.  Not hard. 😛 

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -4061,6 +4061,31 @@ public:
     }
 } Array_zip;
 
+class Array_transpose : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        auto *ap = cast_type<AppliedType>(args.thisType);
+        ENFORCE(ap->klass == Symbols::Array() || ap->klass.data(gs)->derivesFrom(gs, Symbols::Array()));
+        ENFORCE(!ap->targs.empty());
+        auto &elementType = ap->targs.front();
+
+        auto *tuple = cast_type<TupleType>(elementType);
+        if (tuple == nullptr) {
+            return;
+        }
+
+        // The transpose of an array of tuples is a tuple of arrays.
+        vector<TypePtr> transposed;
+        transposed.reserve(tuple->elems.size());
+
+        for (auto &elem : tuple->elems) {
+            transposed.emplace_back(Types::arrayOf(gs, elem));
+        }
+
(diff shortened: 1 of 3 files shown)


maruth-stripe (approved):

core/types/calls.cc
```diff
@@ -3864,6 +3864,31 @@ class Array_zip : public IntrinsicMethod {
     }
 } Array_zip;
 
+class Array_transpose : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
+        auto *ap = cast_type<AppliedType>(args.thisType);
+        ENFORCE(ap->klass == Symbols::Array() || ap->klass.data(gs)->derivesFrom(gs, Symbols::Array()));
+        ENFORCE(!ap->targs.empty());
+        auto &elementType = ap->targs.front();
```
maruth-stripe, line 3873: is it guaranteed that `ap != nullptr` ?

test/testdata/rbi/array.rb
```diff
@@ -70,3 +70,6 @@
 
 T.reveal_type(arr.fetch(0, 'error')) # error: Revealed type: `T.any(Integer, String)`
 T.reveal_type(arr.fetch(0) { 'error' }) # error: Revealed type: `T.any(Integer, String)`
+
+x = [[1, "a"], [2, "b"]]
+T.reveal_type(x.transpose) # error: Revealed type: `[T::Array[Integer], T::Array[String]] (2-tuple)`
```
jez, line 75: Can you add a test case for when the arrays are not the same length? I think we need to make some sneaky `nilable` things show up here.

jez (changes requested):
^

froydnj:
Added some more testcases.  I think we get away easy because of the limitations of tuple types.

jez (approved):

jez:
@froydnj can you also test on Stripe's codebase?

froydnj:

→ https://go/builds/bui_Oe3tfKZj3Hl4r1
→ https://go/builds/bui_Oe3tj3wDnSJ21P
→ https://go/builds/bui_Oe3tjvxB2Ma1aY

froydnj:
> @froydnj can you also test on Stripe's codebase?

Done.  There are a couple of redundant `T.cast`/`T.must` errors that I'll fix up on the next internal bump, but I won't commit this until next week.

PR #6733, opened 2023-02-14
special-case `T::Private::Types::SimplePairUnion` in some `T::Utils` code

froydnj:
### Motivation

`T::Utils.unwrap_nilable` and `T::Utils::Nilable.get_underlying_type` turn out to be relatively hot functions in some serialization code in Stripe's codebase.  We can do a better job by adding some special cases for our `SimplePairUnion` specialization.  We can also duplicate a small amount of code so that we don't have to go through `T::Utils.get_type_info` and thereby avoid the allocation of `TypeInfo`.

Benchmarks before:

```
T.unwrap_nilable(Integer): 88.166 ns
get_underlying_type(Integer): 278.329 ns
T.unwrap_nilable(T.nilable(Integer)): 620.187 ns
get_underlying_type(T.nilable(Integer)): 913.333 ns
T.unwrap_nilable(T.any(Float, Integer)): 603.966 ns
get_underlying_type(T.any(Float, Integer)): 843.252 ns
```

After:

```
T.unwrap_nilable(Integer): 92.836 ns
get_underlying_type(Integer): 90.258 ns
T.unwrap_nilable(T.nilable(Integer)): 195.653 ns
get_underlying_type(T.nilable(Integer)): 275.813 ns
T.unwrap_nilable(T.any(Float, Integer)): 156.667 ns
get_underlying_type(T.any(Float, Integer)): 201.998 ns
```

`get_underlying_type` is 3-4x faster; `unwrap_nilable` gets 3-4x faster on `SimplePairUnion`.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/types/simple_pair_union.rb
+++ b/gems/sorbet-runtime/lib/types/private/types/simple_pair_union.rb
@@ -40,6 +40,7 @@ class T::Private::Types::SimplePairUnion < T::Types::Union
     ]
   end
 
+  # overrides Union
   def unwrap_nilable
     a_nil = @raw_a.equal?(NilClass)
     b_nil = @raw_b.equal?(NilClass)
--- a/gems/sorbet-runtime/lib/types/private/types/simple_pair_union.rb
+++ b/gems/sorbet-runtime/lib/types/private/types/simple_pair_union.rb
@@ -39,4 +39,16 @@ class T::Private::Types::SimplePairUnion < T::Types::Union
       T::Types::Simple::Private::Pool.type_for_module(@raw_b),
     ]
   end
+
+  def unwrap_nilable
+    a_nil = @raw_a.equal?(NilClass)
+    b_nil = @raw_b.equal?(NilClass)
+    if a_nil
+      return types[1]
+    end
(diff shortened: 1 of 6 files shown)


froydnj:
OK, this is even a little faster now after the `unwrap_nilable` refactoring.  I'm a little iffy on the logic duplication, but I think it's worth it in the name of performance.

gems/sorbet-runtime/lib/types/private/types/simple_pair_union.rb
```diff
@@ -39,4 +39,16 @@ def types
       T::Types::Simple::Private::Pool.type_for_module(@raw_b),
     ]
   end
+
+  def unwrap_nilable
```
jez, line 43:
```suggestion
  # overrides Union
  def unwrap_nilable
```

jez (approved):
This looks fine but I wonder if we shouldn't just have something like NilableUnion where `@types` is a Union type, and say that a `Union` type's `@types` never contains a `NilClass`.

froydnj:
Stripe build: https://go/builds/bui_NOnDBBK0PmU1mi

PR #6734, opened 2023-02-14
Add signature for GC.stat_heap

gmcgibbon:
Adds `GC.stat_heap` method signature.

### Motivation
Sorbet currently doesn't have a signature for this method, so when you go to use it, you need to use unsafe or a  manual shim.

### Test plan
This is a sig, so I don't think it needs tests.

### DIFF
--- a/rbi/core/gc.rbi
+++ b/rbi/core/gc.rbi
@@ -138,6 +138,40 @@ module GC
   sig {params(arg0: Symbol).returns(Integer)}
   def self.stat(arg0={}); end
 
+  # Returns information for memory pools in the \GC.
+  #
+  # If the first optional argument, +heap_name+, is passed in and not +nil+, it
+  # returns a +Hash+ containing information about the particular memory pool.
+  # Otherwise, it will return a +Hash+ with memory pool names as keys and
+  # a +Hash+ containing information about the memory pool as values.
+  #
+  # If the second optional argument, +hash_or_key+, is given as +Hash+, it will
+  # be overwritten and returned. This is intended to avoid the probe effect.
+  #
+  # If both optional arguments are passed in and the second optional argument is
+  # a symbol, it will return a +Numeric+ of the value for the particular memory
+  # pool.
+  #
+  # On CRuby, +heap_name+ is of the type +Integer+ but may be of type +String+
+  # on other implementations.
+  #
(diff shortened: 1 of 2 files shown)


rbi/core/gc.rbi
```diff
@@ -138,6 +138,40 @@ module GC
...
+  #
+  # If the optional argument, hash, is given, it is overwritten and returned.
+  #
+  # This method is only expected to work on CRuby.
+  sig { returns(T::Hash[Numeric, T::Hash[Symbol, Numeric]]) }
+  sig { params(heap_name: Integer, hash_or_key: Symbol).returns(Integer) }
+  sig do
+    type_parameters(:T).params(
+      heap_name: T.nilable(Integer),
+      hash_or_key: T.all(T.type_parameter(:T), Hash),
```
gmcgibbon, line 169: The reason I'm using a generic here is because the method mutates a hash and inserts GC stats into it. So, it could be any shape. In this case, I think it is reasonable to expect _any_ kind of object that is a Hash.

rbi/core/gc.rbi
```diff
@@ -138,6 +138,40 @@ module GC
...
+  # a +Hash+ containing information about the memory pool as values.
+  #
+  # If the second optional argument, +hash_or_key+, is given as +Hash+, it will
+  # be overwritten and returned. This is intended to avoid the probe effect.
+  #
+  # If both optional arguments are passed in and the second optional argument is
+  # a symbol, it will return a +Numeric+ of the value for the particular memory
+  # pool.
+  #
+  # On CRuby, +heap_name+ is of the type +Integer+ but may be of type +String+
```
gmcgibbon, line 155: I'm assuming we only care about CRuby right now, but let me know if you'd rather I update the key sigs to String or Integer!

froydnj (changes requested):
Thanks for the PR!  It looks like you have some build failures.  I've also left some comments below.

Tests are generally optional for `sig`s in the standard library, but they are appreciated.  You can see examples of such tests in `test/testdata/rbi/` that use `T.reveal_type` and check for appropriate error messages.

froydnj (approved):
Thanks!

froydnj:

→ https://go/builds/bui_NMgefnBSnVvM0y
→ https://go/builds/bui_NMgeBYryiuLdQZ

PR #6739, opened 2023-02-15
runtime: Treat shapes/tuples as subtypes of Hash/Array

jez:
### Motivation

This is unsound but it at least matches the unsoundness in the static
system.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/types/fixed_array.rb
+++ b/gems/sorbet-runtime/lib/types/types/fixed_array.rb
@@ -62,13 +62,13 @@ module T::Types
       when TypedArray
         # warning: covariant arrays
 
-        value_1, value_2, *values_rest = types
-        if !value_2.nil?
-          value_type = T::Types::Union::Private::Pool.union_of_types(value_1, value_2, values_rest)
-        elsif value_1.nil?
-          value_type = T.untyped
+        value1, value2, *values_rest = types
+        value_type = if !value2.nil?
+          T::Types::Union::Private::Pool.union_of_types(value1, value2, values_rest)
+        elsif value1.nil?
+          T.untyped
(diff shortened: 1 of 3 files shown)


jez:
http://go/builds/bui_NMJDzTX4ODB2A7

gems/sorbet-runtime/lib/types/types/fixed_array.rb
```diff
@@ -59,6 +59,19 @@ def valid?(obj)
...
+        value1, value2, *values_rest = types
+        value_type = if !value2.nil?
+          T::Types::Union::Private::Pool.union_of_types(value1, value2, values_rest)
+        elsif value1.nil?
+          T.untyped
+        else
+          value1
+        end
+
+        T::Types::TypedArray.new(value_type).subtype_of?(other)
```
froydnj, line 74: Since we know `other` is a "simple" type here, we should be able to call `subtype_of_single?` instead?

gems/sorbet-runtime/lib/types/types/fixed_hash.rb
```diff
@@ -38,6 +38,28 @@ def valid?(obj)
...
+        value1, value2, *values_rest = types.values
+        value_type = if !value2.nil?
+          T::Types::Union::Private::Pool.union_of_types(value1, value2, values_rest)
+        elsif value1.nil?
+          T.untyped
+        else
+          value1
+        end
+
+        T::Types::TypedHash.new(keys: key_type, values: value_type).subtype_of?(other)
```
froydnj, line 62: Same question here regarding using `subtype_of_single?`.

froydnj (approved):

PR #6741, opened 2023-02-15
Call register_odd for better oj support

jez:
### Motivation

```ruby
# typed: true
require 'oj'
require_relative './gems/sorbet-runtime/lib/sorbet-runtime'

class MyEnum < T::Enum
  enums do
    X = new
  end
end

serialized = Oj.dump(MyEnum::X)
p(serialized)
x_deser = Oj.load(serialized)
p(MyEnum::X.__id__)
p(x_deser.__id__)
p(x_deser.eql?(MyEnum::X))
```

Before:

```
"{\"^o\":\"MyEnum\",\"serialized_val\":\"x\",\"const_name\":\":X\"}"
840
860
false
```

After:

```
"{\"^O\":\"MyEnum\",\"serialize\":\"x\"}"
840
840
true
```

This is still not great, as it means that the class name is in the serialized
format, which means classes cannot be renamed at will.

### Test plan

See included automated tests.


froydnj (approved):
Bonus points if the test can be integrated into sorbet-runtime's tests.

qaisjp:
Assuming this unblock `T::Enum` working with Temporal…

> This is still not great, as it means that the class name is in the serialized format, which means classes cannot be renamed at will.

If this behaviour changes in the future, might we run into issues with existing Temporal workflows when upgrading Sorbet?

jez:
> If this behaviour changes in the future, might we run into issues with existing Temporal workflows when upgrading Sorbet?

I think that this will only affect new serializations. I think that the first key being `^o` or `^O` means that the old, `^o`-serialized values will use the old, buggy behavior, whereas the new `^O`-serialized values will use the new, correct behavior.

jez:
r? @froydnj 

I added a wholesome test folder so that we don't have to depend on `oj` from the sorbet-runtime gem, but can still test the interaction between the two of them.

froydnj (approved):
Cool.  Silly me, I assumed that the Gemfile would have some way of declaring test-only dependencies, but apparently not?

Thanks for pushing this through.

jez:
> Cool. Silly me, I assumed that the Gemfile would have some way of declaring test-only dependencies, but apparently not?

It does have a way to do development_dependencies, but I like this approach a bit better (I think that if we add oj as a dev dependency, oj can't add sorbet-runtime as a dev dependency, if it wanted to in the future).

PR #6742, opened 2023-02-16
Continue exporting through behavior-less namespaces for exports

aisamanra:
This fixes problems with `export_all!` in some packages.

Writing a reliable test here would require some invariants on file ordering that I'm not sure about, which is why this doesn't include a test for the behavior it's fixing.

### Motivation
In the case of a class or module that defines "behavior", we can come up with a single canonical location for it, and use that location to figure out which package it belongs in. However, for behavior-less modules, this can pose a problem. Imagine that we've got a package `A` and a package `A::B::C`. Ideally, we'd know that `module A::B` belongs to package `A`, but there are situations in which the first definition we see of `module A::B` is in a file that belongs to package `A::B::C` and therefore that becomes the first location. In that case, we might incorrectly believe that the "owner" of that namespace is `A::B::C`. If we try to `export_all!` in `A`, then it previously would miss constants like `A::B::D` because of that misconception.

This technically exports more than it needs to, but only behavior-less namespaces; all _meaningful_ things in the package should be prevented from being exported.

### Test plan

See included automated tests. Tested to verify that the behavior on pay-server is correct with this change.

### DIFF
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -40,8 +40,14 @@ class PropagateVisibility final {
     }
 
     void recursiveExportSymbol(core::GlobalState &gs, bool firstSymbol, core::ClassOrModuleRef klass) {
-        // We only mark symbols from this package.
-        if (!this->definedByThisPackage(gs, klass)) {
+        // We only mark symbols from this package. However, there's a
+        // tough case where non-behavior-defining "namespace-like"
+        // constants might get attributed to other packages (since we
+        // don't have a canonical location, so we use the first place
+        // we see them... which might be in a subpackage) and
+        // therefore this might stop too soon. That's why we only stop
+        // recursing if the thing is actually behavior-defining.
+        if (!this->definedByThisPackage(gs, klass) && klass.data(gs)->flags.isBehaviorDefining) {
             return;
         }
 
--- a/test/testdata/packager/export_all/foo_bar_baz/foo_bar_baz.rb
+++ b/test/testdata/packager/export_all/foo_bar_baz/foo_bar_baz.rb
@@ -2,5 +2,9 @@
 
 module Foo::Bar::Baz
   class Quux
+    extend T::Sig
+
+    sig { void }
+    def example; end
   end
 end


test/testdata/packager/export_all/foo_bar_baz/foo_bar_baz.rb
```diff
@@ -2,5 +2,9 @@
 
 module Foo::Bar::Baz
   class Quux
+    extend T::Sig
+
+    sig { void }
+    def example; end
```
jez, line 8: Out of curiosity, why is this change sufficient to test the behavior?

Should there not also be some sort of error assertion that says in one case, something is not visible, and in another case, it is?

packager/VisibilityChecker.cc
```diff
@@ -40,8 +40,14 @@ class PropagateVisibility final {
...
-        // We only mark symbols from this package.
-        if (!this->definedByThisPackage(gs, klass)) {
+        // We only mark symbols from this package. However, there's a
+        // tough case where non-behavior-defining "namespace-like"
+        // constants might get attributed to other packages (since we
+        // don't have a canonical location, so we use the first place
+        // we see them... which might be in a subpackage) and
+        // therefore this might stop too soon. That's why we only stop
+        // recursing if the thing is actually behavior-defining.
+        if (!this->definedByThisPackage(gs, klass) && klass.data(gs)->flags.isBehaviorDefining) {
```
jez, line 50: I remember aadi saying something that he thought the isBehaviorDefining flag was not always correctly populated on the fast path.

I also see this comment:

https://github.com/sorbet/sorbet/blob/f4e43c95ebb97a03ed2b9d21c041a99f3901ee6f/namer/namer.cc#L230-L231

If we're starting to use this flag more in error messages we might want to make sure we have good fast path tests for these things. For example, we should check that making a fast path edit that either makes a class start defining behavior or stop defining behavior updates that bit on the class correctly.

Right now, I don't see anywhere that tells me that the `definesBehavior` flag is ever _unset_ if a class stops defining behavior.

jez (approved):
The change itself looks plain enough but I have a couple questions.

test/testdata/packager/export_all/foo_bar_baz/foo_bar_baz.rb
aisamanra: This change adds "behavior" (i.e. methods) to the body of the class; the class was previously non-behavior-defining. The fix here is relevant when 1. there is a subpackage that 2. contains non-behavior-defining outer namespaces with 3. behavior-defining inner constants.

packager/VisibilityChecker.cc
aisamanra: I think more tests is a good idea, but it's worth noting we're already using this flag in the visibility checker and I believe we have been since it was first introduced, which means we haven't had risen-to-the-level-of-visibility problems with it in that whole time. That doesn't mean we shouldn't introducer more strict tests and make sure it works in fast-path contexts—we should—but rather that problems with it seem to be relatively esoteric and we've got some wiggle room with it.

PR #6743, opened 2023-02-16
Allow different type as fallback of `Array#fetch`

jacobobq:
Adds a new `:Fallback` type parameter to the method to represent the type of the fallback value when present, and allows the method to return `T.any(Elem, :Fallback)` in such cases. This broader definition more closely matches the actual possibilities in Ruby.

### Motivation
This addresses a problem where seemingly correct Ruby code using `Array#fetch` is [rejected by Sorbet](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A%5B1%2C2%2C3%5D.fetch%280%2C%20'error'%29).

```ruby
[1,2,3].fetch(0, 'error')
```

The issue stems from the type signature of `#fetch`, which expects its fallback value (or block) to be of (or return) the same type contained in the array. While this seems like a sensible choice of types, it will prevent valid Ruby code from passing typechecking.

### Test plan
This is only a trivial change in a RBI file. Please let me know if tests are needed.

### DIFF
--- a/rbi/core/array.rbi
+++ b/rbi/core/array.rbi
@@ -1279,18 +1279,20 @@ class Array < Object
     .returns(Elem)
   end
   sig do
-    params(
+    type_parameters(:Fallback)
+    .params(
         arg0: Integer,
-        arg1: Elem,
+        arg1: T.type_parameter(:Fallback),
     )
-    .returns(Elem)
+    .returns(T.any(Elem, T.type_parameter(:Fallback)))
   end
   sig do
-    params(
+    type_parameters(:Fallback)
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_NMrTA3TMNfANKA
→ https://go/builds/bui_NMrTiI91BATsZg

froydnj (approved):
Thanks for the patch!  I constructed some testcases to make sure that these weren't inadvertently introducing untyped; I'll add those in a separate PR.

PR #6753, opened 2023-02-18
Fix keyword arguments parsing if the send node has a block

ilyailya:
While working on an implicit keyword argument conversions for the Ruby 3, I've noticed that Sorbet was erroring weirdly:

```ruby
# typed: true

sig do
  params(
    name: String,
    tags: T::Hash[T.untyped, T.untyped],
    x: Integer,
    blk: T.proc.returns(String)
  )
  .returns(String)
end
def foo(name, tags: {}, x: 42, &blk); end

foo(
  "",
  tags: {},
# ^^^^ error: Keyword argument hash without `**` is deprecated

  x: 1,
  &blk
)
```

And the proposed autocorrect looked like this and generated an invalid Ruby syntax
```ruby
foo(
  "",
  **tags: {},
  x: 1,
  &blk
)
```

Turns out, it was a bug in Desugarer. It would inline keyword arguments to the arguments store in a send node only if the send node didn't have a block. So from Sorbet perspective the call above after desugaring looked like this
```ruby
foo(
  "",
  {
    tags: {},
    x: 1
  },
  &blk
)
```

From that point of view the autocorrect makes much more sense! But it is still incorrect.

The bug happened, because the Desugarer looked for the keyword args hash at the last position in the arguments store. However, if the block is present, it will be the last argument, and the keywords hash is kept as it is in this case. 

The PR changes Desugarer to check the last two arguments in a search of the keyword args hash.

This change didn't cause any new errors in the Stripe's codebase

### Motivation

Correct autocorrects!

### Test plan

Not sure what would be the best place to put a test for this behaviour. I've added a case to the ruby 3 keyword test for now.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -825,8 +825,15 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                 } else {
                     int numPosArgs = send->args.size();
                     if (numPosArgs > 0) {
-                        // Deconstruct the kwargs hash in the last argument if it's present.
-                        if (auto *hash = parser::cast_node<parser::Hash>(send->args.back().get())) {
+                        // The keyword arguments hash can be the last argument if there is no block
+                        // or second to last argument otherwise
+                        int kwargsHashIndex = send->args.size() - 1;
+                        if (!parser::isa_node<parser::Hash>(send->args[kwargsHashIndex].get())) {
+                            kwargsHashIndex = max(0, kwargsHashIndex - 1);
+                        }
+
+                        // Deconstruct the kwargs hash if it's present.
+                        if (auto *hash = parser::cast_node<parser::Hash>(send->args[kwargsHashIndex].get())) {
                             if (hash->kwargs) {
                                 numPosArgs--;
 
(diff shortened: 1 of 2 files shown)


ast/desugar/Desugar.cc
```diff
@@ -825,8 +825,15 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
                 } else {
                     int numPosArgs = send->args.size();
                     if (numPosArgs > 0) {
-                        // Deconstruct the kwargs hash in the last argument if it's present.
-                        if (auto *hash = parser::cast_node<parser::Hash>(send->args.back().get())) {
+                        // The keyword arguments hash can be the last argument if there is no block
+                        // or second to last argument otherwise
+                        int kwargsHashIndex = send->args.size() - 1;
+                        if (!parser::isa_node<parser::Hash>(send->args[kwargsHashIndex].get())) {
```
neilparikh, line 831: If I'm understanding this correctly, the logic implemented here is:
- if the last arg is a hash, use that as the kwargsHashIndex
- otherwise, check if the second last arg is a hash, and if it is, use that

The comment also mentions the second last arg only fills this role if the last arg is a block, but we don't check that here. Could that be an issue here?

ilyailya: I don't believe that might be an issue. My goal was to fix how keyword arguments are parsed without touching the block parsing, because I believe it works correctly now. Tests also passed, so I think we are good!

I also don't do a check for the block argument, because it's done later in the same function: https://github.com/sorbet/sorbet/blob/ce70710668aaadd9aee1ba676bdb6d856f12c383/ast/desugar/Desugar.cc#L872-L888

Please let me know, if you want me to update my comment to point to the block processing code!

neilparikh (approved):

PR #3562, opened 2020-10-23
remove --suggest-runtime-profiled and T::Utils::RuntimeProfiled

froydnj:
These are relics of a bygone era, and the use case they supported (generated sigs) does not exist anymore.

`RuntimeProfiled` also has this weird thing where `GlobalState` sets it up as a static field, but `sorbet-runtime` (and, IIUC, the intended use case) insists that it is a class instead.  Rather than trying to reconcile this, it seemed better to remove it.

### Motivation

Deleting unused code is a good thing.

### Test plan

Removing the automated tests around this feature, and ensuring that existing tests impacted by this feature (see e.g. the changes in `infer/SigSuggestions.cc` still pass.

PR #6756, opened 2023-02-18
Delete vestiges of RuntimeProfiled

jez:
### Motivation

We removed RuntimeProfiled in #3562, which means these comments don't apply
anymore and we can convert `if` conditions into `ENFORCE`s.

### Test plan

Existing tests


froydnj (approved):

jez:
Build didn't pass and I don't care to debug.

Issue #5240, opened 2022-02-03
Autocorrect to redeclare `type_member` in child class when needed


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20X%20%3D%20type_member%0A%0A%20%20Y%20%3D%20type_template%0A%0A%20%20Z%20%3D%20type_template%28fixed%3A%20String%29%0Aend%0A%0Amodule%20IFoo%0A%20%20extend%20T%3A%3AGeneric%0A%20%20Out%20%3D%20type_member%28%3Aout%29%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20include%20IFoo%0Aend)

```ruby
# typed: true

class Parent
  extend T::Generic

  X = type_member

  Y = type_template

  Z = type_template(fixed: String)
end

module IFoo
  extend T::Generic
  Out = type_member(:out)
end

class Child < Parent
  include IFoo
end
```

#### Observed output

```
editor.rb:18: Type X declared by parent Parent must be re-declared in Child https://srb.help/5014
    18 |class Child < Parent
        ^^^^^^^^^^^^^^^^^^^^
    editor.rb:6: X declared in parent here
     6 |  X = type_member
          ^^^^^^^^^^^^^^^

editor.rb:18: Type Out declared by parent IFoo must be re-declared in Child https://srb.help/5014
    18 |class Child < Parent
        ^^^^^^^^^^^^^^^^^^^^
    editor.rb:15: Out declared in parent here
    15 |  Out = type_member(:out)
          ^^^^^^^^^^^^^^^^^^^^^^^

editor.rb:18: Type Y declared by parent T.class_of(Parent) must be re-declared in T.class_of(Child) https://srb.help/5014
    18 |class Child < Parent
              ^^^^^
    editor.rb:8: Y declared in parent here
     8 |  Y = type_template
          ^^^^^^^^^^^^^^^^^

editor.rb:18: Type Z declared by parent T.class_of(Parent) must be re-declared in T.class_of(Child) https://srb.help/5014
    18 |class Child < Parent
              ^^^^^
    editor.rb:10: Z declared in parent here
    10 |  Z = type_template(fixed: String)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 4
```

#### Expected behavior

[...]

PR #6757, opened 2023-02-19
Allow `T.attached_class` in modules

### LINKED ISSUE #2718: T.attached_class can't be used in modules that are `extend`ed into a class and become singleton methods
### Problem

I have a class method. It's defined inside of a module that uses `mixes_in_class_methods`:
```ruby
module Foo
  module ClassMethods
    sig {returns(T.self_type)}
    def load; end
  end
  mixes_in_class_methods(ClassMethods)
end

class Invoice
  include Foo
end
```

Sorbet thinks that `Invoice.load` returns `T.class_of(Invoice)`, but it actually returns `Invoice`.

### Proposed solution

I'd like to be able to use `T.attached_class`, but Sorbet doesn't realize that I'm in a singleton context. 

I'd like to add `class_methods!` as an annotation, similar to `interface!`:
```ruby
module Foo
  module ClassMethods
    # Sorbet now knows that this is in a singleton context
    class_methods!

    # You can correctly type the method:
    def load; end
  end
end
```

Sorbet would make it illegal to `include` the module. Instead, you must either `extend` it or use `mixes_in_class_methods`.

jez:
### Motivation

Fixes #2718

A common pattern in Ruby involves having modules define methods that take care of instantiating the object they get mixed into in devious ways.

This will allow us to write sigs for those things.

### Summary

The implementation is somewhat straightforward, powered by a rewriter.
It translates a new DSL into a type_member declaration, which allows using `T.attached_class`.

The ensuing complexity mostly just comes from error handling code that needs to special case some new error messages.

In fact, the whole feature to enforce that an `initializable!` module is only included via `extend` is powered by the existing logic to report errors for redeclaring type members. This is nice, because it means the correctness of both of these features is linked (and if there is a bug, fixing it in one fixes it for both).

### Test plan

See included automated tests.


core/types/calls.cc
```diff
@@ -2968,22 +2968,31 @@ class Magic_attachedClass : public IntrinsicMethod {
...
-                auto singletonClass = self.data(gs)->lookupSingletonClass(gs);
-                if (singletonClass.exists()) {
-                    e.addErrorNote(
-                        "`{}` represents instances of a class; `{}` represents the corresponding singleton class",
-                        self.show(gs), singletonClass.show(gs));
+                auto initializable = core::Names::declareInitializable().show(gs);
+                if (selfData->isModule()) {
+                    e.setHeader("`{}` must be marked `{}` before module instance methods can use `{}`", self.show(gs),
+                                initializable, "T.attached_class");
+                    // TODO(jez) Autocorrect to insert `initializable!`
```
jez, line 2982: I plan on doing this in a future change, because there's some code I want to refactor first.

jez:
I don’t want to tie this feature to requires_ancestor. As implemented, this feature comes with none of the downsides that requires_ancestor has.

jez:
I think it’s also important for people to realize that this will turn their module into a generic, and having to write `extend T::Generic` underscores this.

there’s also the point that the type_member generated by the rewrite is allowed to have a bound, which would not be possible to infer 100% of the time from a requires_ancestor declaration (because the bound could be a union type).

website/docs/attached-class.md
```diff
@@ -222,3 +222,129 @@ type suggests to the caller that `Child.make` will return a `Child` instance
...
+
+class ParentModel
+  extend T::Sig
+  extend FinderMethods
+end
+
+class ChildModel < ParentModel
+end
+
+parent = ChildModel.find('pa_123')
```
bdewater, line 261: Shouldn't this be `ParentModel`?

website/docs/attached-class.md
```diff
@@ -222,3 +222,129 @@ type suggests to the caller that `Child.make` will return a `Child` instance
...
+end
+def find_and_log(findable, id)
+  instance = findable.find(id)
+  puts("Found #{instance}")
+  instance
+end
+
+parent = find_and_log(ParentModel, 'pa_123')
+T.reveal_type(parent) # => `ParentModel`
+child = find_and_log(ParentModel, 'pa_123')
```
bdewater, line 318: Similarly, `ChildModel`?

jez:
* master
  * **PR #6883** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6883" target="_blank">Graphite</a> 
    * **PR #6757** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6757" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6757?utm_source=stack-comment).

chaffeqa:
this is very much needed in our case where we have a lot of `ActiveSupport::Concern` modules that build our models 👍

jez:
@chaffeqa thanks for the enthusiasm, but please refrain from commenting on pull requests with such comments, so that contributors can focus on code review.

You're always welcome to share things like this on our [Sorbet Slack](https://sorbet.org/slack) instance, either in #discuss or #internals.

test/testdata/infer/generics/initialize_class_bound.rb
```diff
@@ -0,0 +1,62 @@
...
+  interface!
+
+  sig {abstract.returns(Integer)}
+  def foo; end
+
+  module Factory
+    extend T::Generic
+    interface!
+
+    has_attached_class!(:out) { {upper: Thing} }
```
froydnj, line 15: We don't seem to have any tests for `has_attached_class!` with bounds, but the enclosing module is used in a context that doesn't satisfy those bounds.  Can we add a testcase like that, please?  (I assume that the error messages will be somewhat sensible, but maybe not!)

froydnj (approved):
Minor comments only, since so much of Sorbet's existing machinery is doing the heavy lifting here.

resolver/GlobalPass.cc
```diff
@@ -61,9 +71,29 @@ bool resolveTypeMember(core::GlobalState &gs, core::ClassOrModuleRef parent, cor
...
             // `type_member`, then the grandparent class's file could be edited and Sorbet wouldn't include the
             // grandchild class's file in the set of files to retypecheck.
-            e.setHeader("Type `{}` declared by parent `{}` must be re-declared in `{}`", name.show(gs), parent.show(gs),
-                        sym.show(gs));
-            e.addErrorLine(parentTypeMember.data(gs)->loc(), "`{}` declared in parent here", name.show(gs));
+            if (code == core::errors::Resolver::HasAttachedClassIncluded) {
+                auto hasAttachedClass = core::Names::declareHasAttachedClass().show(gs);
+                if (sym.data(gs)->isModule()) {
+                    e.setHeader("`{}` declared by parent `{}` must be re-declared in `{}`", hasAttachedClass,
+                                parent.show(gs), sym.show(gs));
```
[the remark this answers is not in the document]
jez: I think we use "declared" for error messages like "must be declared `abstract!`" or `final!`, which is what I was modeling this messaging off of.

image

I think that all names here are bad unfortunately.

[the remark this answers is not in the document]
jez: yeah we don't have it for other arms, but we should be able to do them all at once. That work is tracked in this issue:

https://github.com/sorbet/sorbet/issues/5240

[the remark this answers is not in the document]
froydnj: Yup, `abstract!` et al being "declarations" makes sense and the parallels with `has_attached_class!` WFM.

PR #6759, opened 2023-02-21
RSpecCompatibility: Fixed behaviour when a method does not exist to match the original

meganemura:
### Motivation

An error is raised by sorbet-runtime if a non-existent instance method is specified in allow_any_instance_of.
And no particular error without sorbet-runtime.
It would be better if we did not define unnecessary mocks, but we thought it would be more natural if the behaviour was the same as the original behaviour.

Since there is no need for consideration on the sorbet-runtime side for methods that do not exist, a method existence check has been added.

### Test plan

```ruby
method = @klass.instance_method(method_name.to_sym)
```
The above expression will raise a NameError if method is not defined, so it should only be executed if there is a method definition.


meganemura:
@jez 
Is this acceptable? What do you think of this? 😃

jez (approved):

PR #6771, opened 2023-02-23
[Ruby3] Fix false positive hash to keyword arg convertion error

ilyailya:
## Bug
Sorbet errors at the usage of `Dir.[]` in the experimental Ruby 3 mode:
```ruby
# typed: true

def with_base_dir(*segments)
end

Dir[with_base_dir("")].each do |file_path|
  # ^^^^^^^^^^^^^ error: Keyword argument hash without `**` is deprecated
  p file_path
end
```
And the proposed auto-correct would produce:
```ruby
Dir[**with_base_dir("")].each do |file_path|
  p file_path
end

```

That's valid syntax, but it doesn't make sense. No error is needed in this case, because the call is correct and wouldn't cause any issues in Ruby 3.

## Fix
A bit of context first:
- Signature of the `Dir.[]` looks like this: https://github.com/sorbet/sorbet/blob/e17514be9bc6dbc9eeca07e0b499f0215c2da797/rbi/core/dir.rbi#L588-L596
- Calling `.derivesFrom()` on `T.untyped` would always return true
- `isRepeated` parameter's flag means the parameter has a splat `*args` or a double splat `**kwargs`

The bug happens in the method for dispatching send node's arguments. Firstly we try to consume all positional arguments: https://github.com/sorbet/sorbet/blob/e17514be9bc6dbc9eeca07e0b499f0215c2da797/core/types/calls.cc#L941-L962

On the line 947 the loop is broken if `ait` points to a hash of keyword arguments, which will be processed later in the method. The first parameter of `Dir.[]` has a splat, so `flags.isRepeated` would be positive. And also the first argument in the test case I found is `T.untyped`.

Combination of those factors will always break the loop. However, the correct thing to do is to process `*segments` as a regular positional arg.

To do that I needed a way to distinguish between parameters with splat and double splat. 
I've added code to store that information in parameter's flags and changed some if-conditions to respect the new flag.

### Motivation

Ruby 3!

### Test plan

See included automated tests + I've built this version on my devbox and typechecked Stripe's codebase.


ast/ArgParsing.cc
```diff
@@ -17,6 +17,7 @@ core::ParsedArg parseArg(const ast::ExpressionPtr &arg) {
             *cursor,
             [&](const ast::RestArg &rest) {
                 parsedArg.flags.isRepeated = true;
+                parsedArg.flags.isKwSplat = rest.isKwRest;
```
ilyailya, line 20: Ideally I'd like to rename `isRepeated` to `isSplat` and also make it exclusive so the parameter could be either `isSplat` or `isKwSplat`. But it's an ambitious goal, because large chunks of argument dispatch depend on `isRepeated` being a combination of single and double splat. It would be to much of a lift to detangle it right now

ast/Trees.h
```diff
@@ -633,7 +634,7 @@ EXPRESSION(RestArg) {
 
     void _sanityCheck();
 };
-CheckSize(RestArg, 16, 8);
+CheckSize(RestArg, 24, 8);
```
jez, line 637: I would really love to avoid this—we're already anticipating running into memory problems.

Is it possible to solve this without storing more things in trees?

core/ParsedArg.h
```diff
@@ -9,13 +9,16 @@ namespace sorbet::core {
 struct ParsedArg {
     struct ArgFlags {
         bool isKeyword : 1;
-        bool isRepeated : 1;
+        bool isRepeated : 1; // *args or **kwargs
         bool isDefault : 1;
         bool isShadow : 1;
         bool isBlock : 1;
+        bool isKwSplat : 1; // only **kwargs
```
jez, line 16: I don't think we need this. It should already be the case that you can tell `isKwSplat` from `isKeyword && isRepeated`.

jez (changes requested):
^

test/cli/autocorrect-kwsplat/test.rb
```diff
@@ -1,29 +0,0 @@
...
-sig {params(pos: Integer, x: String).void}
-def optional_x(*pos, x: '')
-end
-
-opts = T::Hash[Symbol, Integer].new
-string_keys = T::Hash[String, Integer].new
-nilble_opts = T::Hash[Symbol, T.nilable(Integer)].new
-x = T.let(:x, Symbol)
-
-requires_x(0, 1, 2, opts)
```
ilyailya, line 17: This is the most common use case. It's also a valid code in 2.7, but not in 3.0.

With `--experimental-ruby3-keyword-args` it be auto-corrected like this:

```ruby
requires_x(0, 1, 2, **opts)
```

With the legacy handling like this

```ruby
requires_x(0, 1, 2, **T.unsafe(opts))
```

Both auto-corrects provide valid code, but the legacy solution adds unnecessary `T.unsafe`

test/cli/autocorrect-kwsplat/test.rb
```diff
@@ -1,29 +0,0 @@
...
-
-requires_x(0, 1, 2, **opts)
-
-requires_x(0, 1, 2, **opts, x: 0)
-
-requires_x(0, 1, 2, x: 0, **opts)
-
-requires_x(0, 1, 2, **{x => 0})
-
-optional_x(0, 1, 2, **string_keys)
```
ilyailya, line 27: With the legacy handling we would also error here like it's a kwarg issue and propose to auto-correct with `T.unsafe` as usual. 
However, the actual error here is a type mismatch, because the code is trying to inline hash with string keys into keyword arguments, but it works only with symbol keyed hashes.

Legacy handling will wrap it with `T.unsafe` actually hiding that error! In the new handling the actual error is properly reported (check test/testdata/infer/ruby3_keyword_args.rb)

test/cli/autocorrect-kwsplat/test.rb
```diff
@@ -1,29 +0,0 @@
...
-
-optional_x(0, 1, 2, **nilble_opts)
```
ilyailya, line 29: This example is valid in Ruby 2.7 and Ruby 3.0 and doesn't need any errors

test/testdata/desugar/keyword_args.rb
```diff
@@ -1,31 +0,0 @@
-# typed: true
```
jez, line 1: We should almost _never_ delete tests—instead, we should showcase the new behavior of Sorbet on this test.

The only exception is when a test is completely redundant with some other test. In that case, please mention the other test(s) that duplicate the behavior being tested here.

test/cli/autocorrect-kwsplat/test.out
```diff
@@ -1,161 +0,0 @@
-test.rb:17: Cannot call `Object#requires_x` with a `Hash` keyword splat because the method has required keyword parameters https://srb.help/7019
```
jez, line 1: Ditto the comment below about deleting tests.

test/testdata/infer/arg_matching.rb
```diff
@@ -54,8 +54,6 @@ def call_kwarg
     kwarg(1, b: "hi")
     #           ^^^^ error: Expected `Integer` but found `String("hi")` for argument `b`
     kwarg(1, any)
-    kwarg(1, a_hash)
-    #        ^^^^^^ error: Cannot call `TestArgs#kwarg` with a `Hash` keyword splat because the method has required keyword parameters
```
jez, line 58: Deleting all these errors is a regression, isn't it?

Sorbet previously rejected this code because it couldn't guarantee whether `a_hash` has one entry in it for every parameter that the `kwarg` method requires.

jez (changes requested):
This seems like a regression in its current form.

test/testdata/infer/arg_matching.rb
ilyailya: I've deleted this test because:
1. It's a valid ruby 2.7 code which will work as expected
2. It's also a base case in the `test/testdata/infer/ruby3_keyword_args.rb`:
https://github.com/sorbet/sorbet/blob/d331f62b15b019c12b751e0c49ac2ae599e93999/test/testdata/infer/ruby3_keyword_args.rb#L6-L13

I've also tried getting that case back and running sorbet over it with `--experimental-ruby3-keyword-args`. This is what I got:
```
test/testdata/infer/arg_matching.rb:57: Keyword argument hash without ** is deprecated https://srb.help/7033
    57 |    kwarg(1, a_hash)
                     ^^^^^^
    test/testdata/infer/arg_matching.rb:57: This produces a runtime warning in Ruby 2.7, and will be an error in Ruby 3.0
    57 |    kwarg(1, a_hash)
                     ^^^^^^
  Autocorrect: Done
    test/testdata/infer/arg_matching.rb:57: Replaced with **a_hash
    57 |    kwarg(1, a_hash)
```

>Deleting all these errors is a regression, isn't it?

I don't believe deleting those errors is a regression. The detection we had before was too generic and also proposed to use `T.unsafe`. The new detection doesn't alert on valid code and proposes fixes without introducing unsafe code.

test/testdata/desugar/keyword_args.rb
ilyailya: This test have been fully moved to `test/testdata/infer/ruby3_keyword_args.rb`:
https://github.com/sorbet/sorbet/blob/d331f62b15b019c12b751e0c49ac2ae599e93999/test/testdata/infer/ruby3_keyword_args.rb#L70-L95

I had to rename `foo` to `bar`, because `foo` was already taken

test/cli/autocorrect-kwsplat/test.out
ilyailya: All of the contents of this test also have been moved to `test/testdata/infer/ruby3_keyword_args.rb`:

https://github.com/sorbet/sorbet/blob/d331f62b15b019c12b751e0c49ac2ae599e93999/test/testdata/infer/ruby3_keyword_args.rb#L40-L68

So we didn't miss any use cases

ilyailya (commented):
@jez thanks for your review!

I left some replies on your comments. 

tldr; my goal while deleting the old kwarg conversion detection was not too lose any meaningful tests. For that purpose `autocorrect-kwsplat/test.rb` and `desugar/keyword_args.rb` very fully moved to `infer/ruby3_keyword_args.rb`. The errors deleted in `infer/kwargs_splat.rb` and `infer/kwsplat_is_sometimes_okay.rb` were deleted, because mostly they copy basic scenarios from `infer/ruby3_keyword_args.rb`.

I can add an experimental ruby3 flag to the `kwargs_splat.rb` and `kwsplat_is_sometimes_okay.rb`, so we could retain use cases and state we expect them to have kwarg errors with a certain flag (at least for now). 

Please, let me know if that would work for you!

PR #6775, opened 2023-02-23
Provide direct link to https://sorbet.org/docs/self-type from error

gjtorikian:
I just hit this issue myself, and had to dig around for the quick fix. I think it would be helpful to include a possible solution solution here.


jez (approved):

jez:
Thanks!

PR #6777, opened 2023-02-24
Fix loc update issue on fast path for class/static field scopes

aadi-stripe:
cc @ilyailya 
cc @jez 

Fixes the logic for updating locs on the fast path for class/static field scopes (i.e., Unknown FoundClass types). For perf reasons, we don't update these locs generally, but it *seems* harmless to do so if the loc is already in the same file.

### Motivation

 Doing this fixes a problem with out-of-order reference error reporting on the fast path (other similar error messages may also be benefited). It allows locs for unknown classes to get updated correctly with an edit, preventing crashes.

### Test plan


test/testdata/namer/redefines_module_as_static_field.rb.symbol-table-raw.exp
```diff
@@ -4,8 +4,8 @@ class <C <U <root>>> < <C <U Object>> ()
...
-      module <C <U T>>::<C <U AbstractUtils>>::<C <U Methods>> < <C <U Sorbet>>::<C <U Private>>::<C <U Static>>::<C <U ImplicitModuleSuperclass>> () @ Loc {file=test/testdata/namer/redefines_module_as_static_field.rb start=30:3 end=30:28}
-        module <C <U T>>::<C <U AbstractUtils>>::<C <U Methods>>::<C <U CallValidation>> < <C <U Sorbet>>::<C <U Private>>::<C <U Static>>::<C <U ImplicitModuleSuperclass>> () @ Loc {file=test/testdata/namer/redefines_module_as_static_field.rb start=34:3 end=34:44}
+      module <C <U T>>::<C <U AbstractUtils>>::<C <U Methods>> < <C <U Sorbet>>::<C <U Private>>::<C <U Static>>::<C <U ImplicitModuleSuperclass>> () @ Loc {file=test/testdata/namer/redefines_module_as_static_field.rb start=38:3 end=38:28}
+        module <C <U T>>::<C <U AbstractUtils>>::<C <U Methods>>::<C <U CallValidation>> < <C <U Sorbet>>::<C <U Private>>::<C <U Static>>::<C <U ImplicitModuleSuperclass>> () @ Loc {file=test/testdata/namer/redefines_module_as_static_field.rb start=38:3 end=38:44}
```
aadi-stripe, line 8: This changes the subsequent loc for an erroneous declaration. If you look at the test -- it's quite an edge case, so I think this is a safe change.

namer/namer.cc
```diff
@@ -1175,18 +1179,28 @@ class SymbolDefiner {
...
+            //
+            // At best, we only want to keep one loc in the codebase per unknown class for perf reasons. We don't
+            // want to store O(files) locs for something like Opus. So if the existing loc (which would have been
+            // brought into existence by getClassSymbol()) is not from this file, we don't add a new loc.
+            //
+            // If the unknown class loc is from the same file, it's still possible that it is from a real
+            // definition in that file. In which case, we check the declared bit on the class.
+            // We only set the loc if the class is not declared.
+            bool updateLoc = !isUnknown || (isUnknown && symbol.data(ctx)->loc().file() == ctx.file &&
+                                            !symbol.data(ctx)->isDeclared());
```
jez, line 1198:
```suggestion
            bool updateLoc = !isUnknown || (!symbol.data(ctx)->isDeclared() &&
                                            symbol.data(ctx)->loc().file() == ctx.file);
```

this reads slightly better IMO

core/Symbols.h
```diff
@@ -499,11 +499,14 @@ class ClassOrModule final {
...
     }
 
-    inline bool isUndeclared() const {
-        if (!isClassModuleSet()) {
-            return true;
+    inline bool isDeclared() const {
+        return flags.isDeclared;
+    }
+
+    inline void setDeclared() {
```
jez, line 506: Can we ENFORCE that `isClassModuleSet()`?

jez (approved):

PR #6781, opened 2023-02-25
Make `Class` generic in its attached class

jez:
### Motivation

Fixes #62

image

### Test plan

See included automated tests.

PR #6785, opened 2023-02-27
wip quick hack to try to test T.class_of()[]

jez:
### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -643,6 +643,9 @@ void GlobalState::initEmpty() {
     klass.data(*this)->setIsModule(false);
     ENFORCE(klass == Symbols::T_Types_Base());
 
+    klass = enterClassSymbol(Loc::none(), Symbols::T(), core::Names::Constants::Class());
+    ENFORCE(klass == Symbols::T_Class());
+
     typeArgument =
         enterTypeArgument(Loc::none(), Symbols::noMethod(), Names::Constants::TodoTypeArgument(), Variance::CoVariant);
     ENFORCE(typeArgument == Symbols::todoTypeArgument());
@@ -898,7 +901,6 @@ void GlobalState::initEmpty() {
     Symbols::Symbol().data(*this)->resultType = Types::Symbol();
     Symbols::Float().data(*this)->resultType = Types::Float();
     Symbols::Object().data(*this)->resultType = Types::Object();
-    Symbols::Class().data(*this)->resultType = Types::classClass();
 
     // First file is used to indicate absence of a file
     files.emplace_back();
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -22,7 +22,7 @@ namespace sorbet::core {
 
 using namespace std;
 
-const int Symbols::MAX_SYNTHETIC_CLASS_SYMBOLS = 210;
+const int Symbols::MAX_SYNTHETIC_CLASS_SYMBOLS = 212;
 const int Symbols::MAX_SYNTHETIC_METHOD_SYMBOLS = 50;
 const int Symbols::MAX_SYNTHETIC_FIELD_SYMBOLS = 4;
 const int Symbols::MAX_SYNTHETIC_TYPEARGUMENT_SYMBOLS = 4;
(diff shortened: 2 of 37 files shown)


jez:
* master
  * **PR #6970** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6970" target="_blank">Graphite</a> 
    * **PR #6781** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6781" target="_blank">Graphite</a> 
      * **PR #6983** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6983" target="_blank">Graphite</a> 
        * **PR #6978** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6978" target="_blank">Graphite</a> 
          * **PR #6785** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6785" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6785?utm_source=stack-comment).

core/Symbols.cc
```diff
@@ -112,6 +112,8 @@ TypePtr ClassOrModule::selfType(const GlobalState &gs) const {
     }
 }
 
+// ClassOrModule::resultType is computed by unsafeComputeExternalType,
```
neilparikh, line 115: Should this say `ClassOrModule:: externalType is computed by...`?

core/GlobalState.cc
```diff
@@ -104,6 +104,13 @@ struct MethodBuilder {
         return *this;
     }
 
+    MethodBuilder &repeatedTopArg(NameRef name) {
+        auto &arg = gs.enterMethodArgumentSymbol(Loc::none(), method, name);
+        arg.flags.isRepeated = true;
+        arg.type = Types::top();
+        return *this;
+    }
```
neilparikh, line 112: This is to make `args` into `*args` basically right?

test/testdata/infer/generics/t_class_of.rb
```diff
@@ -0,0 +1,212 @@
...
+
+  X = type_template {{fixed: String}}
+
+  sig {override.returns(X)}
+  def self.foo; ''; end
+end
+
+
+sig do
+  type_parameters(:U)
```
neilparikh, line 52: You can delete the `type_parameters(:U)` here (and on next two below as well).

test/testdata/infer/generics/t_class_of.rb
```diff
@@ -0,0 +1,212 @@
...
+    .params(klass: T.class_of(A)[Integer])
+    #                            ^^^^^^^ error: Wrong number of type parameters for `T.class_of(A)`
+    #                            ^^^^^^^ error: Wrong number of type parameters for `T.class_of(A)`
+    #                            ^^^^^^^ error: `Integer` is not a subtype of upper bound of type member `::<Class:A>::<AttachedClass>`
+    #                            ^^^^^^^ error: `Integer` is not a subtype of upper bound of type member `::<Class:A>::<AttachedClass>`
```
neilparikh, line 70: (for future improvement, not on this branch) Do you think it would be worth hiding the `is not a subtype of ... <AttachedClass>` if there is a type parameter mismatch? It's likely that means the user forgot to supply the attached class (or some other type parameter), so this error is just noise until the mismatch is fixed.

neilparikh (approved):

core/Symbols.cc
jez: I think it should say `resultType`. `externalType` is computed by this method specifically, but the implementation of this method relies on having had `this->resultType` set already, and that field gets set by the mentioned method.

core/GlobalState.cc
jez: Yep.

test/testdata/infer/generics/t_class_of.rb
jez: Not sure. We can chat about it. It likely ties into whatever our eventual decision around whether this could ever be defaulted is.

jez: By comparison, we report two errors for the corresponding error with normal params/args type mismatches

https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28x%3A%20Integer%2C%20y%3A%20String%29.void%7D%0Adef%20foo%28x%2C%20y%29%0Aend%0A%0Afoo%28''%29

neilparikh: Fair enough, we'd definitely want the behaviour of type param mismatches and normal params mismatches to be the same (whether that's silence when the count is wrong or show always is something to chat about later).

Issue #6787, opened 2023-02-27
Improve error messages when attempting to use the result of a void method


amomchilov:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20void%20%7D%0Adef%20this_returns_void%0A%20%20%20123%0Aend%0A%0Athis_returns_void.foo)

```ruby
# typed: true
extend T::Sig

sig { void }
def this_returns_void
   123
end

this_returns_void.foo
```

#### Observed output

The static type-checker error message is pretty good, though arguably the provided link isn't quite right:

```
editor.rb:9: Cannot call method foo on void type https://srb.help/7003
     9 |this_returns_void.foo
                          ^^^
Errors: 1
```

However, the runtime method is a bit jargony and unclear:

```
NoMethodError: undefined method `foo' for T::Private::Types::Void::VOID:Module
```

#### Expected behavior

How about we hook into `#method_missing`, and issue a clearer message like:

```
TypeError: attempted to call #foo on a void-returning method #this_returns_void
```

PR #6788, opened 2023-02-27
Improve msg when attempting to use result of void

amomchilov:
### Motivation

WIP trying to address #6787

### Test plan

I think the existing tests already cover this. TBD.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/types/void.rb
+++ b/gems/sorbet-runtime/lib/types/private/types/void.rb
@@ -6,16 +6,42 @@
 class T::Private::Types::Void < T::Types::Base
   ERROR_MESSAGE = "Validation is being done on an `Void`. Please report this bug at https://github.com/sorbet/sorbet/issues"
 
-  # The actual return value of `.void` methods.
-  #
-  # Uses `module VOID` because this gives it a readable name when someone
-  # examines it in Pry or with `#inspect` like:
-  #
-  #     T::Private::Types::Void::VOID
-  #
-  module VOID
+  class VoidSingleton < BasicObject
+    def inspect
+      "T::Private::Types::Void::VOID"
+    end
+
+    ALLOWED_KERNEL_METHODS = [:object_id, :class, :respond_to?, :eql?]
+
+    if defined?(::PP::ObjectMixin)
+      include ::PP::ObjectMixin
+      alias_method :pretty_inspect, :inspect
+    end
+
+    if defined?(::AwesomePrint)
+      def ai(options = {})
+        ::Kernel.instance_method(:ai).bind_call(self)
+      end
+
+      def method
+        ::Kernel.instance_method(:method).bind_call(self)
+      end
+    end
+
+    def method_missing(method_name, *args)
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/private/types/void.rb
```diff
@@ -6,17 +6,43 @@
...
-  #
-  # Uses `module VOID` because this gives it a readable name when someone
-  # examines it in Pry or with `#inspect` like:
-  #
-  #     T::Private::Types::Void::VOID
-  #
-  module VOID
+  class VoidSingleton < BasicObject
+    def inspect
+      "T::Private::Types::Void::VOID"
```
amomchilov, line 11: TODO: is there a more useful inspect message we could be giving?

gems/sorbet-runtime/lib/types/private/types/void.rb
```diff
@@ -6,17 +6,43 @@
...
+    end
+
+    ALLOWED_KERNEL_METHODS = [:object_id, :class, :respond_to?, :eql?]
+
+    if defined?(::PP::ObjectMixin)
```
amomchilov, line 16: Is it appropriate to include these kind of helpers?

They don't perfect fit here, but checking `if defined?(Sorbet)` in those projects would be equally strange.

Without these, `pp` or `ap` on a void method just raises.

gems/sorbet-runtime/test/types/returns.rb
```diff
@@ -41,7 +41,7 @@ def initialize; end
     end
 
     it 'nils the return for void' do
-      assert_equal(T::Private::Types::Void::VOID, TestReturns.voider)
+      assert_same(T::Private::Types::Void::VOID, TestReturns.voider)
```
amomchilov, line 44: I changed this because I think it's worth asserting the singleton-ness of the result. LMK what you think!

froydnj (changes requested):
We discussed this internally, and we don't think this is a large enough usability improvement given that we'd have to introduce `method_missing` into `sorbet-runtime`'s codebase.

amomchilov:
Hey @froydnj,

Huh I just noticed that. Do you guys aim to steer clear of `method_missing` throughout the whole codebase?

froydnj:
> Huh I just noticed that. Do you guys aim to steer clear of `method_missing` throughout the whole codebase?

We do, both in Sorbet's codebase and Stripe's codebase generally.  We generally think it introduces too much magic runtime behavior and can be better modeled with statically analyzable functions and/or a different design.

amomchilov:
I agree with that motivation. I have some battle scars from particularly bad `method_missing` implementations (sub-classing BasicObject and accidentally implementing `#hash` to return an xml node like `<hash/>`? pandemonium.)

I don't think there's any alternative approach here. Kind of by definition, we want to be hooking onto any method call that can be made, and `#method_missing` is exact API for doing just that. Personally, I find this to be an acceptable exception-to-the-rule. If you disagree, shall I close this PR and related issue?

PR #6789, opened 2023-02-27
Add `Struct`-like pattern-matching to `T::Struct`

### LINKED ISSUE #6232: `T::Struct` pattern matching.
### Problem

`T::Struct` can't participate in pattern matching, because it doesn't implement `#deconstruct` and `#deconstruct_keys`:

```ruby
case typed_foo # This hits the "else"
in { i: 123 } then puts("Match!")
else puts("No match.")
end

case typed_foo # This hits the "else"
in [123] then puts("Match!")
else puts("No match.")
end
```

There's already precedence for supporting this, as the vanilla `Struct` type implements both [`#deconstruct`](https://ruby-doc.org/core-3.1.2/Struct.html#deconstruct-method) and
[`#deconstruct_keys`](https://ruby-doc.org/core-3.1.2/Struct.html#deconstruct_keys-method):

```ruby
VanillaFoo = Struct.new(:i, keyword_init: true)

vanilla_foo = VanillaFoo.new(i: 123)

case vanilla_foo
in { i: 123 } then puts("Match!")
else puts("No match.")
end

case vanilla_foo
in [123] then puts("Match!")
else puts("No match.")
end
```

### Proposed solution

Just add the two methods. A naïve solution is as simple as:

```ruby
class T::Struct
  # ...
  
  def deconstruct_keys(array_of_names)
    h = serialize.symbolize_keys

    return h if array_of_names.nil?

    h.slice(*array_of_names)
  end

  def deconstruct = serialize.values
  
end
```

Thought it's probably worth optimizing by:

* Generating the minimal necessary Hash (rather than generating the whole Hash and slicing it down
* Generating the Hash using Symbols as keys, from the get-go.

amomchilov:
We might actually want to reconsider this, depending on how closely we want `T::Struct` to have parity with `Struct`. There's some pretty whacky behaviour. Some examples:

### Motivation

Closes #6232

### Test plan

See included automated tests.

They've been written to match the tests in:

1. [`spec/ruby/core/struct/deconstruct_spec.rb`](https://github.com/ruby/ruby/blob/master/spec/ruby/core/struct/deconstruct_spec.rb)
1. [`spec/ruby/core/struct/deconstruct_keys_spec.rb`](https://github.com/ruby/ruby/blob/master/spec/ruby/core/struct/deconstruct_keys_spec.rb)

### DIFF
--- a/gems/sorbet-runtime/lib/types/struct.rb
+++ b/gems/sorbet-runtime/lib/types/struct.rb
@@ -5,6 +5,30 @@ class T::InexactStruct
   include T::Props
   include T::Props::Serializable
   include T::Props::Constructor
+
+  def deconstruct_keys(keys)
+    h = serialize
+
+    return h.to_h { |k, v| [k.to_sym, v] } if keys.nil?
+    raise TypeError, "wrong argument type #{keys.class} (expected Array or nil)" unless keys.is_a?(Array)
+    return {} if h.count < keys.count
+
+    result = {}
+
+    keys.each do |k, v|
+      case k
+      when String then  result[k] = h.fetch(k)
+      when Symbol then  result[k] = h.fetch(k.to_s)
+      when Integer then result[k] = h.values.fetch(k)
+      end
+    rescue KeyError, IndexError
+      break
+    end
+
+    result
+  end
+
+  def deconstruct = serialize.values
 end
 
 class T::Struct < T::InexactStruct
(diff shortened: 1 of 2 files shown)


jez:
Thanks, but for the record, I'd rather Sorbet have proper static support for pattern matching before implementing pattern matching for things defined in sorbet-runtime.

People should be able to trust that using `T::Struct` is akin to well-typed code. With pattern matching right now, everything is untyped.

There are already other places in sorbet-runtime where people think they can safely call a method on `T::Struct`, but in fact it's not properly typed. Those decisions were made before Sorbet existed—we shouldn't make more of those APIs.

amomchilov:
@jez [Alexandre agrees](https://github.com/sorbet/sorbet/issues/6232#issuecomment-1277810362)

This was just the first step, to experiment with it. I don't suggest we ship this as-is, and frankly, I don't even think we should want parity with `Struct`. It does some weird stuff.

PR #6792, opened 2023-03-01
Fix crash introduced by #6763

jez:
### Motivation

Thanks to @dirceu and @paracycle for the test case.

### Test plan

See included automated tests.


resolver/GlobalPass.cc
```diff
@@ -216,7 +216,11 @@ void Resolver::finalizeAncestors(core::GlobalState &gs) {
             ref.data(gs)->setIsModule(true);
             ref.data(gs)->singletonClass(gs); // force singleton class into existence
         }
+    }
 
+    auto n = gs.classAndModulesUsed();
+    for (int i = 1; i < n; ++i) {
```
neilparikh, line 222: Why does this need to be a separate loop? Is it because calling `lookupSingletonClass` might look up information from a different class or module in the ancestors (so we need to call `singletonClass` for all of them before we can call lookup for any of them)?

Also, any particular reason for doing `auto n = ...` and then `i < n` here over ` i < gs.classAndModulesUsed()` like above?

jez: > Also, any particular reason for doing `auto n = ...` and then `i < n` here over ` i < gs.classAndModulesUsed()` like above?

Not sure I really understand the question, so I'm going to answer "no particular reason"



> Why does this need to be a separate loop?

It technically doesn't unless you want to add the ENFORCE that I've added below, which I think is a good ENFORCE.

neilparikh (approved):

jez:

→ https://go/builds/bui_NRn6sDE16XUglr
→ https://go/builds/bui_NRn6xKAbl3TtGl

PR #6794, opened 2023-03-01
don't allocate hashes for zero-{type ,}parameter functions

froydnj:
### Motivation

This PR allocates slightly less and makes the API of `Decl` slightly cleaner in that `type_parameters` can no longer be accessed.

The title of the PR is about `type_parameters`, but I am sneaking in a small change for not allocating hashes for zero-parameter functions as well.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
@@ -231,15 +231,17 @@ module T::Private::Methods
         decl.on_failure = nil
       end
       if decl.params.equal?(ARG_NOT_PROVIDED)
-        decl.params = {}
+        decl.params = FROZEN_HASH
       end
       if decl.type_parameters.equal?(ARG_NOT_PROVIDED)
-        decl.type_parameters = {}
+        decl.type_parameters = FROZEN_HASH
       end
 
       decl.finalized = true
 
       self
     end
+
+    FROZEN_HASH = {}.freeze
   end
 end


paracycle:
Yes, this would indeed break Tapioca's ability to generate correct sigs for methods that use type parameters. Having said that, we patch Sorbet runtime heavily to get access to information that we wouldn't normally be able to, so we could start doing it for this as well.

froydnj:
> Yes, this would indeed break Tapioca's ability to generate correct sigs for methods that use type parameters. Having said that, we patch Sorbet runtime heavily to get access to information that we wouldn't normally be able to, so we could start doing it for this as well.

Thank you for the additional context @paracycle !  I guess the use-case is something like:

* You have a gem with Sorbet typing enabled in some places
* You use Tapioca to generate RBIs for the gem to cover those cases where the gem doesn't have sigs (and so that you needn't have the source of the gem available at typechecking time)
* ...so you want to be able to correctly reflect any uses of `type_parameters` that the gem already used into the RBI sigs

Is that approximately the sketch of how it works, or are there other things that I am missing?

paracycle:
> * You have a gem with Sorbet typing enabled in some places
> * You use Tapioca to generate RBIs for the gem to cover those cases where the gem doesn't have sigs (and so that you needn't have the source of the gem available at typechecking time)
> * ...so you want to be able to correctly reflect any uses of `type_parameters` that the gem already used into the RBI sigs
> 
> Is that approximately the sketch of how it works, or are there other things that I am missing?

Yes, that's more or less it. Tapioca has the philosophy of always generating sigs from source, regardless of if the gem has exported RBI files or not, and then merging the exported RBI files on to the generated ones.

So, we do runtime reflection on the sigs associated with methods discovered, and try to generate the original sig as faithfully as possible. This includes being able to reify generics, which implies that we need to make Sorbet runtime to not erase generics at runtime but to store them instead. Which, in turn implies that there is already a lot Sorbet runtime patching going on when Tapioca is doing what it is trying to do.

I am not opposed to dropping the storage of `type_parameters` if it will mean less memory usage, however, I would be grateful if we could do it in a way that makes it easier for Tapioca to patch over more easily as well.

froydnj:
Re-requesting review from jez insofar as the patch has changed for the simpler.

Dropping `type_parameters` here seems like too big of an impact, so let's go with a simpler fix.

jez (approved):

PR #6795, opened 2023-03-01
Show all defs if class def'd multiple times but in RBIs

jez:
### Motivation

Previously, classes with multiple definition locs in RBI files would show zero
results.

For example, `String` has a definition in `string.rbi` and `BigDecimal.rbi`, and
that meant that Sorbet would report no definition locations for jump-to-def on
`String`.

### Test plan

The test case failed previously.

### DIFF
--- a/main/lsp/requests/definition.cc
+++ b/main/lsp/requests/definition.cc
@@ -35,16 +35,21 @@ unique_ptr<ResponseMessage> DefinitionTask::runRequest(LSPTypecheckerDelegate &t
             for (auto loc : sym.locs(gs)) {
                 locMapping.emplace_back(loc, config.loc2Location(gs, loc));
             }
-            // Move all non-existent Locations to the front of the vector.
-            auto validLocations = absl::c_partition(locMapping, [](const auto &p) { return p.second == nullptr; });
-            // If we have multiple locations, eliminate "definitions" in RBI files
-            // for classes and modules, since the one(s) in Ruby files are more
-            // likely to be what the user is looking for.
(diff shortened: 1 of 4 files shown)


main/lsp/requests/definition.cc
```diff
@@ -35,16 +35,21 @@ unique_ptr<ResponseMessage> DefinitionTask::runRequest(LSPTypecheckerDelegate &t
...
-            auto validLocations = absl::c_partition(locMapping, [](const auto &p) { return p.second == nullptr; });
-            // If we have multiple locations, eliminate "definitions" in RBI files
-            // for classes and modules, since the one(s) in Ruby files are more
-            // likely to be what the user is looking for.
-            if (sym.isClassOrModule() && std::distance(validLocations, locMapping.end()) > 1) {
-                validLocations = std::partition(validLocations, locMapping.end(),
-                                                [&gs](const auto &p) { return p.first.file().data(gs).isRBI(); });
+            {
+                // Erase all invalid locations
+                auto validLocations = absl::c_partition(locMapping, [](const auto &p) { return p.second != nullptr; });
```
ilyailya, line 40: OOC, would it be the same if you use `remove_if` instead of `partion`?

ilyailya (approved):
LGTM! 
Left one non-blocking comment

main/lsp/requests/definition.cc
jez: yes, but there is no `absl::c_remove_if` it looks like so I just used this because I didn't want to have to look up `std::remove_if` docs.

Issue #437, opened 2019-05-08
LSP: Feature - highlight untyped lines


jvilk-stripe:
### Problem

If you're working in a Sorbet file that's `# typed: true`, it's hard to tell at-a-glance which lines are untyped (and somewhat unsafe).

### Proposed solution

Support an LSP request that returns info diagnostics for untyped lines, and (potentially) what % of the file is untyped.

Extensions can make this request for the currently open file(s) to always highlight untyped lines that the developer is actively viewing.

The Flow extension for VS Code has this feature; we should examine its implementation.

jvilk-stripe:
Note: For those who want this functionality today, you can temporarily change your file to `# typed: strong` to see what's untyped; they will surface as errors.

manhhung741:
I think this feature will be very impactful! 
I’m anticipating `untyped` is a big issue once people start “trusting” sorbet to catch errors for them.

jez:
Fixed in #6824.

PR #6797, opened 2023-03-01
Output warning in LSP for untyped code

maruth-stripe:
First stab at getting warnings for untyped code through LSP. It's what was earlier reported as an error for `typed: strong` files but now reported as warning in lower strict levels.

Additionally this adds a `sorbet.warnUntypedValue` configuration option which lets the user specify if they would like to see these warnings or not. This is `false` by default. We provide a command that the user can use to toggle this setting. (eg: in VSCode, cmd + shift + p followed by `Sorbet: Toggle ...`)

### Motivation

https://github.com/sorbet/sorbet/issues/437

### Test plan

Need to expand test harness for LSP warnings?
See included automated tests.

https://user-images.githubusercontent.com/111023001/222272804-1bc42d46-05a3-4ec4-bb23-2a23f86d7387.mov

https://user-images.githubusercontent.com/111023001/222540867-16067a02-56b0-4411-bb68-2fb3877aa8bf.mov


main/lsp/ErrorReporter.cc
```diff
@@ -155,7 +155,8 @@ void ErrorReporter::pushDiagnostics(uint32_t epoch, core::FileRef file, const ve
             tags.push_back(DiagnosticTag::Unnecessary);
             diagnostic->tags = move(tags);
         }
-        diagnostic->severity = DiagnosticSeverity::Error;
+
+        diagnostic->severity = toDiagnosticSeverity(error->severity);
```
jez, line 159: IMO, this is overabstracted.

Sorbet takes a somewhat strong philosophical stance that "warnings don't exist—there's errors and there's nothing"

in the case of this feature, while we're hijacking LSP's diagnostic reporting for this, which does have levels of errors, fundamentally the feature we're really trying to build (if it there were an API for that in the LSP spec) is to have some sort of view or layer on top of source code.

All of which is to say: I think that the API we program against, the core::Error and ErrorClass APIs, should not have a notion of severity.

Rather, at the point where we translate these errors to LSP diagnostics right here, we should ask, `error->what == <the special untyped code error class>`, and if so, set the severity to something else.

In particular, by not having a way to set a severity for an Error, people can't make (bad) arguments in the future like, "Why don't we just build <broken type system feature that should not exist at all>, but make it only a warning?"

(Also, I think we should call that severity Information, not Warning, which again drives home the point that it's not like there are some "optional errors.")

infer/inference.cc
```diff
@@ -300,8 +300,6 @@ unique_ptr<cfg::CFG> Inference::run(core::Context ctx, unique_ptr<cfg::CFG> cfg)
                         DEBUG_ONLY(histogramInc("untyped.sources", bind.bind.type.untypedBlame().rawId()););
                         if (auto e = ctx.beginError(bind.loc, core::errors::Infer::UntypedValue)) {
                             e.setHeader("This code is untyped");
-                            e.addErrorNote("Support for `{}` is minimal. Consider using `{}` instead.", "typed: strong",
-                                           "typed: strict");
```
jez, line 304: We should not delete this in the final feature.

One way to do this better might be:

1.  Introduce a new error class, like `core::errors::Infer::UntypedValueInformation`
2.  Make that error report at `StrictLevel::True`, instead of `StrictLevel::Strong`
3.  Build a helper function like `ErrorClass errorClassForUntyped(const GlobalState &gs, FileRef file)`, which takes on the new `isOpenInClient()` logic that currently resides in `shouldReportErrorOn`.
4.  Call that helper function as the argument to `beginError`:

    ```cpp
    auto what = errorClassForUntyped(gs, ctx.file)
    ctx.beginError(bind.loc, what)
    ```

5. Hide this call to `addErorrNote` behind a check like `if (what == core::errors::Infer::UntypedValue)`

This note is important because some Sorbet-internal features are powered by `T.untyped`. For example, it's not possible to use `# typed: strong` on a file that uses `super`, because Sorbet models `super` with `T.untyped`. So it's important that we expose this note when people _are_ trying to use `

main/lsp/ErrorReporter.cc
maruth-stripe: ack, patched

Thanks for the detailed feedback and context!

infer/inference.cc
maruth-stripe: ack, patched!

website/docs/error-reference.md
```diff
@@ -4058,6 +4058,13 @@ To fix this error, ensure that the left and right operands' types match before
 doing the comparison. For example, try converting `String`s to `Symbol`s with
 `to_sym` (or vice versa with `to_s`).
 
+## 7047
+
+Usages of values with type `T.untyped` are highlighted by this error. It is
+intended to only show up as a warning in editors through LSP for files with
+typed sigil set between `true` and `strong` (exclusive). For `typed: strong`
```
neilparikh, line 4065: Suggestion: could write this as "typed sigil set to `true` or `strict`" (I'm okay with the current phrasing if you prefer that).

neilparikh (approved):

PR #6798, opened 2023-03-01
Allow visible_to check to be skipped for certain allow-listed packages

aadi-stripe:
Per the tin, adds a new CLI-flag-driven feature that allows certain packages to bypass the `visible_to` check and import other packages regardless of the presence/absence of the annotation.

### Motivation

In the Stripe codebase, we have autogenerated "god" packages that import multiple other packages in the codebase. Adding `visible_to` annotations for these packages creates unnecessary churn on our developers. Hence, I am introducing this feature so that people do not need to add annotations for these packages -- the `visible_to` check will simply be skipped for them.

### Test plan


jez (commented):
Why is this specified as a command line option, and not an option in a DSL method on the relevant package?

aadi-stripe:
> Why is this specified as a command line option, and not an option in a DSL method on the relevant package?

Because we don't really want this to be used anywhere apart from the one autogenerated package we control.

aisamanra (approved):
I think this looks good! I also appreciate the `MangledName` refactoring: that cleans things up nicely.

PR #6801, opened 2023-03-02
Boolean doesn't exist anymore

jez:
### Motivation

Stripe's codebase used to monkey patch the Ruby standard library to
define a `Boolean` constant. After 14 years, we finally deleted that
monkey patch. I don't think anyone else will really benefit from having
this autocorrect, and for a sense of closure I'd like to completely rid
ourselves of this hack.

### Test plan

See included automated tests.


neilparikh (approved):

jez:
@neilparikh I had to make some follow up changes to get the tests to pass. Feel free to just review the new commits.

test/testdata/infer/generics/apply_f.rb
```diff
@@ -50,9 +50,9 @@ def takes_any_int_string(y)
   end
   T.reveal_type(x) # error: `T.any(Integer, String)`
 
-  x = apply_f_int(y) do |x| # error: Could not find valid instantiation of type parameters for `Object#apply_f_int`
```
jez, line 53: This error being reported was actually a bug, and I simply didn't realize it at the time.

The lack of an error here and an inferred type of `Integer` is actually what we'd expect.

core/TypeErrorDiagnostics.cc
```diff
@@ -66,21 +67,22 @@ void TypeErrorDiagnostics::maybeAutocorrect(const GlobalState &gs, ErrorBuilder
         e.replaceWith(fmt::format("Wrap in `{}`", *gs.suggestUnsafe), loc, "{}({})", *gs.suggestUnsafe,
                       loc.source(gs).value());
     } else {
+        // Duplicate the current constraint, so that we don't record any additional constraints
+        // simply from trying to generate autocorrects, and then solve it so that each additional
+        // heuristic here doesn't interfere with later heuristics.
+        auto constr = constrOrig.deepCopy();
```
neilparikh, line 73: Can we deref `constrOrig.deepCopy()` here so we don't need to deref `constr` everywhere it's used?

neilparikh (approved):
One minor suggestion, but fine to merge as-is

core/TypeErrorDiagnostics.cc
jez: We can't do it here, `constr` is a `unique_ptr`. If we deref'd it right here, it would be dropped after the line, and the memory pointed to could be overwritten with garbage.

We could do it on another line in a new variable:

```cpp
auto constrPtr = constrOrig.deepCopy();
auto constr = *constrPtr;
```

But that's more characters than typing than `*constr` everywhere (until there's another 20 `*constr` in this method), and even then the original variable would still be in scope, so it would be possible to accidentally still use `*constrPtr` instead of `constr`.

So IMO it's overkill, it's going to be in the CPU cache in all likelihood (and again, this method is not performance sensitive).

PR #6813, opened 2023-03-06
Add settings for building ruby on linux-arm64

amytsai-stripe:
### Motivation

Be able to produce a linux-arm64 compatible version of sorbet_ruby for internal testing at Stripe.

cc @froydnj 

### Test plan

Built a single version in a Docker VM like so. Will follow up with internal CI results.
```
bazel-5.2.0-linux-arm64 build @sorbet_ruby_2_7//:ruby.tar.gz --crosstool_top=@bazel_tools//tools/cpp:toolchain
```


froydnj (approved):
Pretty satisfying that this was so straightforward.  Thank you!

PR #6815, opened 2023-03-06
`//tools/config:linux` → `@platforms//os:linux`

jez:
### Motivation

Some places were using this config setting (which pins both the CPU and
the OS), when they don't actually care about the CPU.

Trying to clean up build config settings.

### Test plan

See included automated tests.


froydnj (approved):
:shipit:

Issue #6812, opened 2023-03-06
FileUtils.cd does not require a block argument


dduugg:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AFileUtils.cd%20'%2F')

```ruby
# typed: true

FileUtils.cd '/'
```

#### Observed output

```
editor.rb:3: cd requires a block parameter, but no block was passed https://srb.help/7021
     3 |FileUtils.cd '/'
                        ^
    https://github.com/sorbet/sorbet/tree/master/rbi/stdlib/fileutils.rbi#L264: defined here
     264 |  module_function def cd(dir, verbose: nil, &block); end
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

No type errors

- - -

See also https://ruby-doc.org/stdlib-2.7.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-cd

PR #6819, opened 2023-03-06
Fix FileUtils.cd sig

dduugg:
### Motivation

Resolves https://github.com/sorbet/sorbet/issues/6812

### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/fileutils.rbi
+++ b/rbi/stdlib/fileutils.rbi
@@ -257,7 +257,7 @@ module FileUtils
     .params(
       dir: T.any(String, Pathname),
       verbose: T.nilable(T::Boolean),
-      block: T.proc.returns(T.type_parameter(:U)),
+      block: T.nilable(T.proc.returns(T.type_parameter(:U))),
     )
     .returns(T.type_parameter(:U))
   }
--- a/test/testdata/rbi/fileutils.rb
+++ b/test/testdata/rbi/fileutils.rb
@@ -1,0 +1,7 @@
+FileUtils.cd('/')  # change directory
+
+FileUtils.cd('/', verbose: true)   # change directory and report it
+
+FileUtils.cd('/') do  # change directory
+  puts 'in / dir'     # do something
+end                   # return to original directory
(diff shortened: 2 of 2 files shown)


test/testdata/rbi/fileutils.rb
```diff
@@ -0,0 +1,7 @@
+FileUtils.cd('/')  # change directory
```
jez, line 1: Your tests are going to fail unless you add an explicit `# typed:` sigil to this file.

Also, this test doesn't actually test your new RBI change unless you make it at least `# typed: true` (the normal Sorbet default would be `# typed: false` if a file doesn't mention a sigil).

Finally, note that tests for RBI changes are optional as long as the diff is simple enough.

jez:

→ https://go/builds/bui_NTkR41Na9Lfkbe
→ https://go/builds/bui_NTkRxFM0vXZQ07

rbi/stdlib/fileutils.rbi
```diff
@@ -257,7 +257,7 @@ module FileUtils
     .params(
       dir: T.any(String, Pathname),
       verbose: T.nilable(T::Boolean),
-      block: T.proc.returns(T.type_parameter(:U)),
+      block: T.nilable(T.proc.returns(T.type_parameter(:U))),
     )
     .returns(T.type_parameter(:U))
```
jez, line 262: In the case where no block is passed, Sorbet is saying that since it didn't see any input given of type `T.type_parameter(:U)`, it will be impossible for the method to return anything. So assuming that the implementation is well-typed according to this signature, the function must raise.

To properly type this, you will have to overload the signature, or change the return type to `T.any(Integer, T.type_parameter(:U))`, because it looks like this method returns `0` if no block was passed.

Sorbet doesn't support using overloaded signatures with methods defined using `module_function`, so if you want to overload this signature, you will also have to duplicate the `self.` and non-`self.` definitions of the method.

dduugg: Ah, ok, i think it's easier at this point to just pass an empty block in my code. Thanks for elaborating!

jez: I think if you pass an empty block, when the block finishes, Ruby will cd back to the original folder before the cd, effectively cancelling itself out.

dduugg: ah, yep, 🤦, but i still have an easy workaround. thanks!

PR #6820, opened 2023-03-07
Set type to `T.untyped` if constraint fails to solve

jez:
### Motivation

The "Expression does not have a fully-defined type" error that Sorbet reports is
almost always a bug in Sorbet—except in the case it mentions, where the type is
not fully defined because it references someone else's type member explicitly,
the error has been caused because there was some part of Sorbet's inference
algorithm that either forgot to (or in this case: couldn't) instantiate
a generic type, leaving it not fully-defined.

The SolveConstraint case in `processBinding` already has a case that says "if
the constraint fails to resolve, set the result type to `T.untyped`," but that
logic only fires when a method call has been passed a block. In this test case,
the method call has been given no block and is solved as soon as all arguments
are type checked. This PR copies the `SolveConstraint` logic into calls.cc.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/infer/no_valid_instantiation_result.rb
+++ b/test/testdata/infer/no_valid_instantiation_result.rb
@@ -18,6 +18,5 @@ def main(str)
   f = T.let(->(y) { y }, T.proc.params(x: Integer).void)
   res = apply_f(str, f)
   #     ^^^^^^^ error: Could not find valid instantiation of type parameters for `Object#apply_f`
-  #     ^^^^^^^^^^^^^^^ error: Expression does not have a fully-defined type (Did you reference another class's type members?)
   T.reveal_type(res) # error: `T.untyped`
 end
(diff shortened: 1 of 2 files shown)


test/testdata/infer/no_valid_instantiation_result.rb
```diff
@@ -18,6 +18,5 @@ def main(str)
   f = T.let(->(y) { y }, T.proc.params(x: Integer).void)
   res = apply_f(str, f)
   #     ^^^^^^^ error: Could not find valid instantiation of type parameters for `Object#apply_f`
-  #     ^^^^^^^^^^^^^^^ error: Expression does not have a fully-defined type (Did you reference another class's type members?)
```
jez: you can see the old vs new behavior in this commit

iz-stripe (approved):
LGTM!

ilyailya (approved):

PR #6821, opened 2023-03-07
add signatures for `==` and `!=` for VM-level singletons

froydnj:
### Motivation

Somebody wanted to catch `!var == "constant string"`; this is one way to do it.

I think given the test failure, this needs to be implemented as an intrinsic in `calls.cc`, since we need to do the right thing for `TrueClass == T::Boolean` (it needs to be typed as returning `T::Boolean` rather than `FalseClass`).

I know that https://blog.jez.io/problems-typing-ruby-equality/ is a thing, but I think it's very reasonable to build in special behavior for these three classes.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_NTzQTu6it3cg0M
→ https://go/builds/bui_NTzQVmnf34O24E

rbi/core/false_class.rbi
```diff
@@ -23,4 +23,12 @@ class FalseClass
   sig {returns(TrueClass)}
   def !
   end
+
+  sig {params(obj: FalseClass).returns(TrueClass)}
+  sig {params(obj: BasicObject).returns(FalseClass)}
+  def ==(obj); end
```
jez, line 29: For `T.untyped`, this selects `TrueClass`.

jez: (the blog post you link mentions overloaded signatures, and it seems that pay-server's codebase is failing for the same reasons outlined in the blog post)

jez:
If you do get this to a place where you think it's worth reviewing, I would like to be given a chance to review it.

froydnj:

→ https://go/builds/bui_NUUI7u2cyjB6A1
→ https://go/builds/bui_NUUIHyfm2N96n0

froydnj:

→ https://go/builds/bui_NUVjLo0SWwHfnT
→ https://go/builds/bui_NUVjZHkJ6axjK4

froydnj:

→ https://go/builds/bui_NUVnIG4mAiV0kX
→ https://go/builds/bui_NUVn9vJbywnejx

PR #6825, opened 2023-03-08
[Ruby3] Fix an arg dispatch for untyped methods with default keyword args

ilyailya:
##Bug

```ruby
# typed: true

module B
  def self.inner_foo(arg0 = [], arg1: true); end
  def self.outer_foo(arg0)
    inner_foo(arg0)
            # ^^^^ error: Keyword argument hash without `**` is deprecated
  end
end
```

## Expected behavior

Same, but without the error
```ruby
# typed: true

module B
  def self.inner_foo(arg0 = [], arg1: true); end
  def self.outer_foo(arg0)
    inner_foo(arg0)
  end
end
```

[sorbet.run](https://sorbet.run/?arg=--experimental-ruby3-keyword-args#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A%0Amodule%20B%0A%20%20def%20self.inner_foo(arg0%20%3D%20[]%2C%20arg1%3A%20true)%3B%20end%0A%20%20def%20self.outer_foo(arg0)%0A%20%20%20%20inner_foo(arg0)%0A%20%20end%0Aend%0A)

### Motivation

Ruby 3

### Test plan

See included automated tests.


core/types/calls.cc
```diff
@@ -832,6 +832,8 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
     }
     auto posArgs = args.numPosArgs;
     bool hasKwargs = absl::c_any_of(data->arguments, [](const auto &arg) { return arg.flags.isKeyword; });
```
jez, line 834: Do we need to audit any of our other uses of `hasKwargs` to see if it should be using `hasNonDefaultKwParams` instead?

ilyailya:

→ https://go/builds/bui_NUuJjDWYAAhvrv
→ https://go/builds/bui_NUuJBLeA05QsBz

core/types/calls.cc
ilyailya: I've double checked that and didn't find any obvious spots to update. 

The edge case I was trying to fix is limited to the positional arguments dispatching. The code responsible for this ends with the loop I was changing. The rest of the method is about dispatching keyword args, so the change shouldn't affect it.

Additionally to that, CI has passed and typecheck over p-s didn't show any errors

neilparikh (approved):

PR #6826, opened 2023-03-08
Factor Counters into //common/counters

jez:
### Motivation

I want to be able to use Counters from more places without pulling in
all of `common` (specifically: //common/concurrency). I need to pull
some things apart to be able to break the cycles.

### Commit summary

Review by commit

- **git mv common/sort.h common/sort** (5655fd50f)

- **Add common/sort/BUILD** (fa07cc0f3)

- **fastmoda 'common/sort.h' 'common/sort/sort.h'** (1d0f32bfb)

- **Depend on //common/sort in //common** (5b240e2d9)

- **git mv common/ConstExprStr.h common/strings** (de9bf8870)

- **fastmoda 'common/ConstExprStr.h' 'common/strings/ConstExprStr.h'** (4dae071a7)

- **Depend on //common/strings in //common** (dfd05d4e3)

- **Add common/strings/BUILD** (6889f06b6)

- **formatting.h: Remove redundant #include** (9d920c19c)

- **Move common/formatting.h to common/strings** (4c4fdaf65)

- **fastmoda 'common/formatting.h' 'common/strings/formatting.h'** (2531af208)

- **Remove common.h from Counters_impl.h** (28b8d023a)

- **zmv 'common/Counters(*)' 'common/counters/Counters$1'** (291874aca)

- **fastmoda 'common/Counters' 'common/counters/Counters'** (b405d83f6)

- **Format** (d6ae0a901)

- **Whoops, typo** (d795809b3)

- **Depend on //common/counters in //common** (0000354f5)

- **Depend on //common/exception from //common/counters** (327926bd8)

### Test plan

There are no behavior changes in this PR—just refactorings.

PR #6827, opened 2023-03-08
Require cleanup after multiplexJob

jez:
### Motivation

This will require that people call our `.cleanup()` function. I
considered doing this with a destructor, but it seemed annoying to have
to move everything inside a nesting scope just to be able to get this to
work.

Review by commit.

### Test plan

See included automated tests.


jez:
* master
  * **PR #6826** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6826" target="_blank">Graphite</a> 
    * **PR #6827** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6827" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6827?utm_source=stack-comment).

common/concurrency/WorkerPool.h
```diff
@@ -5,6 +5,25 @@
...
+        MultiplexCleanup(const MultiplexCleanup &) = delete;
+        MultiplexCleanup &operator=(const MultiplexCleanup &) = delete;
+
+        // It would be neat if we could say that this method can only be called once.
+        // One way to do that would be to make this a static method somewhere, and accept the
+        // receiver as a && value. But clang currently won't check use-after-move errors in Sorbet,
+        // so there's no point to factoring it like that IMO.
+        void cleanup(WorkerPool &workers) {
+            workers.consumeCounters();
+        }
```
froydnj, line 24: Note that nothing ensures this is actually called; you may consider adding a boolean flag that gets set to true in this function and `ENFORCE`'d to be true in the destructor.

froydnj (approved):
Overall I'm slightly negative on this, but I don't think I feel strongly enough to block the PR; I'm not sure all of this is necessary, but having run into the issue where the counters are only accumulated from the worker pool(s) at set points, I also know this removes thinking about "did I get the most up-to-date counters at this point".

Have you tested this on Stripe's codebase to see if there's any adverse performance impact to collecting the counters repeatedly like this?

> it seemed annoying to have to move everything inside a nesting scope just to be able to get this to work.

I'm not sure I understand why a new nesting scope would be needed?  The destructor should just get automatically called on the return value, even if it's unnamed?  But I guess calling the destructor ~immediately on an unnamed value essentially runs into the performance problem described below.

The current PR follows this pattern ~everywhere:

```c++
auto multiplexResult = ...;

...collect output of worker threads and do stuff

multiplexResult.cleanup();
```

but you could equally imagine:

```
auto multiplexResult = ...;

// We have to call this, might as well get it over with.
multiplexResult.cleanup();

...collect output of worker threads and do stuff
```

and that works fine from a correctness point of view: the cleanup tasks wouldn't be called until the earlier `multiplexJob` call workers had finished...but it would be bad from a performance point of view: the cleanup tasks would block the collection of the output from worker threads, so the main thread would be sitting around doing nothing.

jez:

→ https://go/builds/bui_NURznTZpXtRxn7
→ https://go/builds/bui_NURzjvBXsRsQLm

jez:
Yeah, this make Sorbet about 3.5% slower unfortunately. So I'll keep this as a prototype and just collect manual information.

PR #6829, opened 2023-03-08
Change sig of GC.stat_heap to be Integer

peterzhu2118:
### Motivation

The signature of GC.stat_heap should be consistent with the one for GC.stat where all the values in the hash are Integers rather than the more generic Numeric type.

### Test plan

N/A.


jez:

→ https://go/builds/bui_NUT7CqCYQSBayu
→ https://go/builds/bui_NUT7ErjzRZmWS5

PR #6837, opened 2023-03-10
Reduce Sorbet runtime retained memory by releasing location information as soon as possible

paracycle:
It seems like the location information in `DeclarationBlock` is only needed to be able to pass it to `sig_builder_error_handler`, so we should be able to just release it after we are able to build a `Declaration` object.

### Motivation

@byroot noticed a surprising amount of "backtrace" objects lying around in Shopify Core monolith memory that are all marked as `uncollectible`. Peaking at them, the common relation is:
```
(IMEMO: ment) -> (DATA: proc) -> (IMEMO: env) -> (STRUCT: (null)) -> (DATA: frame_info) -> (DATA: backtrace)
```
which we've found points to the `Kernel.caller_locations(1, 1)&.first` call inside the `T::Sig#sig` method.

The `loc` field of `DeclarationBlock` is used to store the location of the signature declaration. However, the value stored in that field is a `Thread::Backtrace::Location` object that was returned from `caller_locations`.

Holding onto this object is problematic since `caller_locations` materializes a `Thread::Backtrace` object, and each `Thread::Backtrace::Location` object holds a reference to that backtrace. Before Ruby 3, a call to `caller_locations` would materialize the whole backtrace, even if the `caller_locations(1, 1)` form was used. Ruby 3 has improved this situation by implementing partial backtraces, but holding on the the `loc` value still retains a `Location` and `Backtrace` object regardless.

This means that if Sorbet runtime does not release the location object as soon as possible, it would end up retaining a lot of backtraces, namely one for each signature defined. This causes a lot of memory growth for no good reason. 

This is significantly more important for codebases that are still running on Ruby 2.7, but is still relevant to post Ruby 3 codebase, since `Backtrace` objects have not had write-barriers until a recent change on Ruby `master`.

### Test plan

I am not sure how we could test this, tbh. Happy to implement a test if you can give me guidance.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -341,7 +341,7 @@ module T::Private::Methods
 
     signature =
       if current_declaration
-        build_sig(hook_mod, method_name, original_method, current_declaration, declaration_block.loc)
+        build_sig(hook_mod, method_name, original_method, current_declaration)
       else
         Signature.new_untyped(method: original_method)
       end
@@ -358,7 +358,7 @@ module T::Private::Methods
       .decl
   end
 
-  def self.build_sig(hook_mod, method_name, original_method, current_declaration, loc)
+  def self.build_sig(hook_mod, method_name, original_method, current_declaration)
     begin
       # We allow `sig` in the current module's context (normal case) and
       if hook_mod != current_declaration.mod &&
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -339,6 +339,8 @@ module T::Private::Methods
         nil
       end
 
+    declaration_block.loc = nil
+
     signature =
       if current_declaration
         build_sig(hook_mod, method_name, original_method, current_declaration)


jez:
> Peaking at them

do you mind sharing how this is done? I'd love to have this technique in my toolbox for tracking down weird ruby runtime allocations in the future.

byroot:
I generally use https://github.com/Shopify/heap-profiler for an overview, and https://github.com/csfrancis/harb for digging.

https://github.com/djudd/reap can be useful too especially it's flamegraph, even though it has to be read carefully as it can be a bit misleading sometimes.

byroot:
For full reference of the impact on various Ruby versions:

  - Partial backtrace introduced in 3.0: https://bugs.ruby-lang.org/issues/17031
  - Backtrace write barriers in https://github.com/ruby/ruby/pull/7240 and https://github.com/ruby/ruby/pull/7253

jez:
This passes Stripe's test suite, sorbet-runtime test suite, and (presumably?) Shopify test suite, so I'm inclined to believe that it's a harmless change. I don't think we need further tests, though if you were particularly interested, we do have certain tests that call into `GC.stat` and make comparisons before and after some region of code:

- gems/sorbet-runtime/test/types/method_validation.rb
- gems/sorbet-runtime/test/types/types.rb

You could cargo cult some of the stuff in there if you wanted to try to write a test that ensures that this never accidentally regresses.

In any case, I'm going to kick off CI and enable auto-merge. You're welcome to try to race up a test before it merges, or you can follow up with a test in a future PR (or just leave it untested).

Thanks for the change!

jez (approved):

PR #6840, opened 2023-03-11
Add SLOW_DEBUG_ONLY

jez:
### Motivation

It takes a very large amount of time to typecheck Stripe's codebase in
the release-debug-linux configuration.

After I generated a web-trace-file from one of these runs, I realized that the
`index` phase doesn't even start until about 50% into the run. That meant that
things coming before `index` were somehow taking half the time, which should not
be the case.

The slowdown blames to this ENFORCE, which, while a good ENFORCE, is
accidentally quadratic in the number of files in the codebase (with some extra
factor for how many common prefixes the files share).

We have precedent for `SLOW_ENFORCE` which are good ENFORCEs that are too slow
to run on large codebases. This adds `SLOW_DEBUG_ONLY`, and uses it for one
particularly slow region of code.

This change nearly cuts the time to run a release-debug-build on Stripe's
codebase in half.

### Test plan

Tested manually by building in release-debug-linux configuration and generating
a new web-trace-file on Stripe's codebase.

### DIFF
--- a/common/common.h
+++ b/common/common.h
@@ -65,6 +65,11 @@ constexpr bool skip_slow_enforce = false;
         X;            \
     }
 
+#define SLOW_DEBUG_ONLY(X)                                      \
+    if constexpr (!::sorbet::skip_slow_enforce && debug_mode) { \
+        X;                                                      \
+    }
+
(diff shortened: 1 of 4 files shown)


froydnj (approved):
Nice catch!

PR #6841, opened 2023-03-11
Add some more SLOW_DEBUG_ONLY

jez:
### Motivation

This is based on #6840

Disabling these two sanityChecks is maybe a little more coarse grained, but it
saves another minute of runtime running release-debug-linux on Stripe's codebase.

Personally, I have never seen these ENFORCEs fire, even in tests, even long ago.
So while I think they are good things to ENFORCE, I think we can skip them in
release-debug-linux mode.

### Test plan

Tested by running web-trace-file on Stripe's codebase.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -4201,7 +4201,7 @@ vector<ast::ParsedFile> resolveSigs(core::GlobalState &gs, vector<ast::ParsedFil
     return trees;
 }
 void sanityCheck(const core::GlobalState &gs, vector<ast::ParsedFile> &trees) {
-    if (debug_mode) {
+    if (debug_mode && !skip_slow_enforce) {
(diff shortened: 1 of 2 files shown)


resolver/resolver.cc
```diff
@@ -4201,7 +4201,7 @@ vector<ast::ParsedFile> resolveSigs(core::GlobalState &gs, vector<ast::ParsedFil
     return trees;
 }
 void sanityCheck(const core::GlobalState &gs, vector<ast::ParsedFile> &trees) {
-    if (debug_mode) {
+    if (debug_mode && !skip_slow_enforce) {
```
neilparikh, line 4204: Why is this duplicating the `!::sorbet::skip_slow_enforce && debug_mode` check rather than just calling the macro?

PR #6842, opened 2023-03-12
Fix out-of-order reference check -- remove first-def-locs sharing & checking undeclared constants

aadi-stripe:
Makes two changes, one refactor and one feature change.

1. Removes cross-treewalk sharing of `firstDefinitionLocs` in resolver. For the purpose of out-of-order checking, it turns out that cross-walk-sharing is unnecessary, and purely local resolution is enough. Thus, we do not need to maintain this info after the first resolution pass of a file.

2. Turns off the out-of-order reference check for undeclared namespaces, e.g.
```
X = A # out-of-order, but we won't report it now, because `A` is an "undeclared" filler namespace

module A::C
end
```

### Motivation

**For change 1**
This change simplifies the logic and mitigates memory issues and crashes. We don't hold on to shared pointers after a walk is finished; the RAII of the walk works more intuitively.

**For change 2**
I had made a fix https://github.com/sorbet/sorbet/pull/6777 for the out-of-order reference check crashing on the fast path. This creates some other issues, however. To explain, I'm going to provide a bit of background context:

_Background for #6777_ 
The out-of-order reference check was crashing on the fast path due to undeclared namespaces not updating locs across edits, e.g. consider the same above file:
```
X = A

module A::C
end
```

On the fast-path, we were crashing in `checkReferenceOrder` if there was an edit that **moved up** the declaration of `module A::C`. This was because `A` is an undeclared namespace -- for which we didn't used to update locs on the fast path, as per the reason given [here](https://github.com/sorbet/sorbet/pull/6777/files#diff-199cb30223d3e0c80ca7608d1f39363f1dfc9b153f51638bf741c01dfbd128a0L1178-L1183). When we don't update the loc, the `beginError` logic necessarily crashes because the previously preserved `loc` exceeds the size of the updated source file.

I made https://github.com/sorbet/sorbet/pull/6777 which changes the behavior to update undeclared symbol locs if the symbol is defined in the same file. This mitigated the crashes and other issues called out in that linked comment. 

_Problem caused by #6777_
However, this created a situation where code like this incorrectly reports an OOO error:

```
class Baz::A # line 1
  p(Baz) # this incorrectly reports an error
end

class Baz::B # line 5
end
```

This happens because after #6777, the loc of `Baz` gets recorded as line 5. `Baz` is an undeclared namespace, defined only in this file -- and so, the change in #6777 effectively pushes the **last** **seen** loc as the canonical loc of `Baz` as the namer passes over the file. This behavior is regardless of fast-path, slow-path, etc. For most cases where we need the canonical loc for an undeclared namespace, this is harmless. But it causes an issue for the ooo check.

There are 2 options here:
1). Fix the ooo check to handle undeclared namespaces "properly". This will require a bit of work.
2). Don't report undeclared namespaces.

I think 2) is okay. From a Ruby standpoint, `class X::A` is _not_ a declaration of `X` at all. So there's no practical reason for us to check out-of-order-ness relative to such declarations.

The obvious question is: now that I'm removing undeclared namespace checking completely the from out-of-order enforcement, should I also revert #6777 ? I think the answer to that (somewhat counter-intuitively) is no. I think that #6777 is good for other reasons. The previous loc update logic in the namer caused other issues during fast-path updates, so it's still valuable to have that fix.

### Test plan

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -139,7 +139,7 @@ private:
     // There will be a map like this for each file that resolver processes, but it will outlive the initial treewalk
     // to find ConstantResolutionItems and only get released once all ConstantResolutionItems for that file have been
     // handled.
-    shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> firstDefinitionLocs;
+    UnorderedMap<core::SymbolRef, core::LocOffsets> firstDefinitionLocs;
 
     // Increments to track whether we're at a class top level or whether we're inside a block/method body.
     int loadScopeDepth_;
@@ -150,7 +150,6 @@ private:
         bool resolutionFailed = false;
         bool possibleGenericType = false;
         bool loadTimeScope = false;
-        shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> firstDefinitionLocs;
 
         ConstantResolutionItem() = default;
         ConstantResolutionItem(const shared_ptr<Nesting> &scope, ast::ConstantLit *lit) : scope(scope), out(lit) {}
@@ -317,13 +316,13 @@ private:
     // then, report an error.
     static void
     checkReferenceOrder(core::Context ctx, core::SymbolRef resolutionResult, const ast::UnresolvedConstantLit &c,
-                        const shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> &firstDefinitionLocs) {
+                        UnorderedMap<core::SymbolRef, core::LocOffsets> &firstDefinitionLocs) {
         if (!ctx.file.data(ctx).isRBI() && resolutionResult.exists() &&
             resolutionResult.isOnlyDefinedInFile(ctx.state, ctx.file)) {
             core::LocOffsets defLoc;
-            const auto it = firstDefinitionLocs.get()->find(resolutionResult);
+            const auto it = firstDefinitionLocs.find(resolutionResult);
 
-            if (it != firstDefinitionLocs.get()->end()) {
+            if (it != firstDefinitionLocs.end()) {
                 // Use the loc from the local first defs table, if it exists.
                 // When the symbol is declared multiple times in the same file, we want to ensure that
                 // we compare against the *first* definition.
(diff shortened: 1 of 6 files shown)


jez:
Before I forget to mention it after the review, thanks for the PR description, it makes sense to me.

> The obvious question is: now that I'm removing undeclared namespace checking completely the from out-of-order enforcement, should I also revert #6777 ?

No, we should not revert. I believe it would have been possible to make other test cases that also crash due to those locs not getting updated on the fast path. (Any locs being stale means that those locs can effectively never appear in error messages or else cause some low-volume number of crashes, which, if a loc can't be used in error messages then what good is it?)

resolver/resolver.cc
```diff
@@ -317,13 +316,13 @@ class ResolveConstantsWalk {
     // then, report an error.
     static void
     checkReferenceOrder(core::Context ctx, core::SymbolRef resolutionResult, const ast::UnresolvedConstantLit &c,
-                        const shared_ptr<UnorderedMap<core::SymbolRef, core::LocOffsets>> &firstDefinitionLocs) {
+                        UnorderedMap<core::SymbolRef, core::LocOffsets> &firstDefinitionLocs) {
```
jez: Should we still take this as a `const &`?

resolver/resolver.cc
```diff
@@ -1362,7 +1356,8 @@ class ResolveConstantsWalk {
             const bool possibleGenericType = false;
             if (resolveConstantJob(ctx, nesting_, constant, resolutionFailed, possibleGenericType)) {
                 categoryCounterInc("resolve.constants.nonancestor", "firstpass");
-                if (loadScopeDepth_ == 0) {
+                if (loadScopeDepth_ == 0 && (!constant->symbol.isClassOrModule() ||
+                                             constant->symbol.asClassOrModuleRef().data(ctx)->isDeclared())) {
```
jez, line 1359: Do you think it's worth briefly mentioning why we only report reference order for declared comments?

> While Sorbet treats `class A::B; end` like an implicit definition of `A`, it's actually a
> reference of `A`--Ruby will require a proper definition of `A` elsewhere. Long term,
> Sorbet should be taught to emit errors when these references are not actually defined,
> matching Ruby's behavior. Then the reference order checks will be able to check all
> references against their definitions, and not limit them to only isDeclared symbols here.
>
> (Historically, Stripe's custom autoloader used static analysis to predeclare these
> intermediate namespaces, so they would always be defined at the right time. As Stripe's
> codebase moves away from this legacy autoloader, it will be easier to introduce such
> changes into Sorbet.)

jez (approved):
Great change, thanks!

PR #6851, opened 2023-03-15
Update Hash#shift RBI

bitwise-aiden:
### Motivation
As outlined in [this Ruby Bug](https://bugs.ruby-lang.org/issues/16908), fixed in [this PR](https://github.com/ruby/ruby/pull/5360), `Hash#shift` has been updated to return `nil` when the hash is empty. Previously it would return the value returned from `hash.default(nil)` which depending on the `default_proc` could produce weird results. This PR updates the `Hash` RBI so that `shift` now reflects the returned value as of 3.2 

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/hash.rbi
+++ b/rbi/core/hash.rbi
@@ -1198,14 +1198,14 @@ class Hash < Object
   def filter!(&blk); end
 
   # Removes a key-value pair from *hsh* and returns it as the two-item array `[`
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_NXU7WOpKlrOnqF
→ https://go/builds/bui_NXU7uVrovtXAXD

jez (approved):

Issue #6849, opened 2023-03-15
Symbol#name fails with "method name does not exist"


andyw8:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A%3Afoo.name)

```ruby
# typed: true
extend T::Sig

:foo.name
```

#### Observed output

```
editor.rb:4: Method name does not exist on Symbol component of Symbol(:foo) https://srb.help/7003
     4 |:foo.name
             ^^^^
  Got Symbol(:foo) originating from:
    editor.rb:4:
     4 |:foo.name
        ^^^^
Errors: 1
```

#### Expected behavior

It should return `foo`:
https://www.bigbinary.com/blog/ruby-3-adds-symbol-name

PR #6852, opened 2023-03-16
Add signature for Symbol#name

andyw8:
### Motivation

Closes https://github.com/sorbet/sorbet/issues/6849

Note: This method is available only in Ruby 3.0 or higher.

### Test plan

I don't believe a test is necessary for this change.

### DIFF
--- a/rbi/core/symbol.rbi
+++ b/rbi/core/symbol.rbi
@@ -233,6 +233,10 @@ class Symbol < Object
   end
   def match?(*args); end
 
+  # Returns the name or string corresponding to *sym*. Unlike `to_s`, the returned string is frozen.
+  sig {returns(String)}
+  def name(); end
+
   # Same as `sym.to_s.succ.intern`.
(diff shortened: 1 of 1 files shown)


rbi/core/symbol.rbi
```diff
@@ -318,4 +318,8 @@ class Symbol < Object
   # an object. As *sym* is already a symbol, `self` is returned in this case.
   sig {returns(T.self_type)}
   def to_sym(); end
+
+  # Returns the name or string corresponding to *sym*. Unlike `to_s`, the returned string is frozen.
```
andyw8, line 322: I manually copied this over from the Ruby 3 docs, is there a better approach I should use?

froydnj:

→ https://go/builds/bui_NXkc5F0vqrYmzP
→ https://go/builds/bui_NXkcichBaxwj2Z

froydnj (approved):
Thanks!

PR #6856, opened 2023-03-17
Add Refinement RBI for 3.2 changes

bitwise-aiden:
### Motivation
`Refinement` didn't previously have an RBI, so I added it with the one existing method of `import_methods`. 

**Adding new `Refinement#refined_class` method to the RBI for new 3.2 changes.**
[Docs](https://docs.ruby-lang.org/en/3.2/Refinement.html)
[Applied in changeset](https://bugs.ruby-lang.org/projects/ruby-master/repository/git/revisions/54198c7b97d3d353f7ac233e0360034b6e7b6cb6)

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/refinement.rbi
+++ b/rbi/core/refinement.rbi
@@ -1,0 +1,54 @@
+# typed: __STDLIB_INTERNAL
+
+# [Refinement](https://docs.ruby-lang.org/en/3.2/Refinement.html) is a class of
+# the self (current context) inside refine statement. It allows to import
+# methods from other modules, see
+# [import_methods](https://docs.ruby-lang.org/en/3.2/Refinement.html#method-i-import_methods).
+class Refinement < Object
+  # Imports methods from modules. Unlike
+  # [Module#include](https://docs.ruby-lang.org/en/3.2/Module.html#method-i-include),
+  # [Refinement#import_methods](https://docs.ruby-lang.org/en/3.2/Refinement.html#method-i-import_methods)
+  # copies methods and adds them into the refinement, so the refinement is
+  # activated in the imported methods.
+  #
+  # Note that due to method copying, only methods defined in Ruby code can be imported.
+  #
+  # ```ruby
+  # module StrUtils
+  #   def indent(level)
+  #     ' ' * level + self
+  #   end
+  # end
+  #
+  # module M
+  #   refine String do
+  #     import_methods StrUtils
+  #   end
+  # end
+  #
+  # using M
+  # "foo".indent(3)
+  # #=> "   foo"
+  #
+  # module M
+  #   refine String do
+  #     import_methods Enumerable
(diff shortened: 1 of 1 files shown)


rbi/core/refinement.rbi
```diff
@@ -0,0 +1,48 @@
...
+  # module M
+  #   refine String do
+  #     import_methods Enumerable
+  #     # Can't import method which is not defined with Ruby code: Enumerable#drop
+  #   end
+  # end
+  # ```
+  #
+  # Also aliased as: import_methods
+  sig {params(module: Module, rest: T::Array[Module]).void}
```
bitwise-aiden, line 42: [Docs for this method](https://docs.ruby-lang.org/en/3.2/Refinement.html#method-i-import_methods) have `import_methods(module, ...) → self` as the signature, stating that it acts on "modules". I've made the signature reflect a 1+ approach with rest, but wanted to make sure this is how this would normally be approached.

rbi/core/refinement.rbi
```diff
@@ -0,0 +1,54 @@
...
+  #
+  # Also aliased as: import_methods
+  sig do
+    params(
+      module: Module,
+      rest: T::Array[Module],
+    )
+    .returns(Refinement)
+  end
+  def import_methods(module, *rest); end
```
bitwise-aiden, line 49: [Docs for this method](https://docs.ruby-lang.org/en/3.2/Refinement.html#method-i-import_methods) have `import_methods(module, ...) → self` as the signature, stating that it acts on "modules". I've made the signature reflect a 1+ approach with rest, but wanted to make sure this is how this would normally be approached.

rbi/core/refinement.rbi
```diff
@@ -0,0 +1,54 @@
...
+  #     # Can't import method which is not defined with Ruby code: Enumerable#drop
+  #   end
+  # end
+  # ```
+  #
+  # Also aliased as: import_methods
+  sig do
+    params(
+      module: Module,
+      rest: T::Array[Module],
```
jez, line 45: This is supposed to be `Module`

you specify the element type in the sig. you are saying that each additional positional argument to `import_methods` must be a list of modules.

froydnj:

→ https://go/builds/bui_NXkn1cDUzl5Lzt
→ https://go/builds/bui_NXkn98lSYfIHnY

rbi/core/refinement.rbi
```diff
@@ -0,0 +1,54 @@
...
+  #
+  # Also aliased as: import_methods
+  sig do
+    params(
+      module: Module,
+      rest: Module,
+    )
+    .returns(Refinement)
+  end
+  def import_methods(module, *rest); end
```
froydnj, line 49: It looks like using `module` here is not a good idea, as it is a reserved word in Ruby.  `mod`, maybe (and in the sig too)?

PR #6859, opened 2023-03-17
Release memory after deduping

jez:
### Motivation

We call `sortAndDedupe` to shrink the `nameHashes` field of `UsageHash`.
(This structure is used to remember the names of symbols that a file might
use, so we know which downstream files might need to be retypechecked.)

This was causing all `FileHash`'s in Stripe's codebase to take up about 6x
the space that it needed to. After shrinking, Sorbet uses about 1/6th the
memory to store all `FileHash` objects.

Note that this cost was only paid for cold-cache runs of LSP. Warm cache
runs of LSP would read file hashes from the cache, which would not have
serialized the unused capacity in these vectors.

So while this is definitely an improvement, it's not an improvement that
applies to 100% of LSP initializations. (But it does reduce the worst
case.)

### Test plan

For correctness: existing tests

To check memory reduction: I have a WIP branch, based on top of #6827, which
adds a _lot_ of writes to the trace file. It also adds another metric for
`current_rss`. This lets me see correlate the current RSS with given phases of
Sorbet. I can see that the current and max RSS values drop by the expected
amount for cold-cache LSP initialization requests.

### DIFF
--- a/core/FileHash.cc
+++ b/core/FileHash.cc
@@ -85,6 +85,7 @@ WithoutUniqueNameHash::WithoutUniqueNameHash(const GlobalState &gs, NameRef nm)
 void WithoutUniqueNameHash::sortAndDedupe(std::vector<core::WithoutUniqueNameHash> &hashes) {
     fast_sort(hashes);
     hashes.resize(std::distance(hashes.begin(), std::unique(hashes.begin(), hashes.end())));
+    hashes.shrink_to_fit();
 }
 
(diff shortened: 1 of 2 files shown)


froydnj:
Is it possible that we could be more clever and use less memory along the way, rather than just at the end?

jez:
> Is it possible that we could be more clever and use less memory along the way, rather than just at the end?

I plan to check this in a future change. For now, I want to lock this in.

iz-stripe (approved):

ilyailya (approved):

PR #6865, opened 2023-03-20
Fix Pathname#binread return type

dduugg:
### Motivation

`Pathname#binread` is `nil` when the file is empty, or when the offset exceeds the length of the file

### DIFF
--- a/rbi/stdlib/pathname.rbi
+++ b/rbi/stdlib/pathname.rbi
@@ -414,7 +414,7 @@ class Pathname < Object
         length: Integer,
         offset: Integer,
     )
-    .returns(String)
+    .returns(T.nilable(String))
   end
   def binread(length=T.unsafe(nil), offset=T.unsafe(nil)); end
 


jez:
This causes no problems for Stripe's codebase, likely because we're not using `Pathname#binread`, so I'll merge for now.

That being said, I notice some potentially lurking problems:

- `Pathname#binread` is implemented by simply calling `File.binread(...)` on this path name. But Sorbet's definition for neither `File.binread` nor `IO.binread` mention that the return value might be `nil`.
- You mention some conditions for when this method returns `nil`:
  - I can't reproduce the "when the file is empty" case. For me, an empty file returns `""` not `nil`.
  - For the `offset` case, we may want to overload this signature to reflect that if you do not pass an offset parameter, that the result cannot be `nil` (if that truly is the case).

In any case, I can reproduce that this method sometimes returns `nil`, so I'm fine simply merging this change as is and fixing forward for those other problems as desired.

jez:
Right, like you mention when you pass a specific offset, it returns `nil`. But if the file was empty and you omit the parameter, it will be empty string:

```
❯ touch empty.txt

❯ ruby -e 'p(File.binread("empty.txt"))'
""

❯ ruby -e 'require "pathname"; p(Pathname("empty.txt").binread)'
""
```

So I was pointing out that if there are certain situations where we can know that Ruby will not return `nil` (like maybe: when no offset has been passed), it would be nice to have the signature reflect the non-`nil` result.

dduugg:
Ah, that makes sense. Now that I think about the former is more just a special case of the latter as I phrased things.

PR #6867, opened 2023-03-20
remove the `Enumerable` sig from `Array#+`

froydnj:
### Motivation

`Array#+`, at least in Ruby 2.7, requires its arg to be an array.  Even if it does permit the more general form in Ruby (we'd have to check), the `Enumerable` sig should come second as the more general case.

### Test plan

See included automated tests.


froydnj:
I am dumb, because I missed the implicit `to_ary` conversion in `rb_ary_plus`.  Silly implicit conversions.

PR #6870, opened 2023-03-21
Fix failing MethodPatch tests

bitwise-aiden:
### Motivation
Since Ruby 3.2 `MethodPatchesTest` fails because it tries to use `prepend` inside a `refine` block. 
```
TypeError: Refinement#prepend has been removed
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/test/types/method_patches.rb:8:in `prepend'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/test/types/method_patches.rb:8:in `block in <module:MethodRefinement>'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/test/types/method_patches.rb:7:in `refine'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/test/types/method_patches.rb:7:in `<module:MethodRefinement>'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/test/types/method_patches.rb:6:in `<class:MethodPatchesTest>'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/test/types/method_patches.rb:5:in `<module:Test>'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/test/types/method_patches.rb:4:in `<top (required)>'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/Rakefile:16:in `require'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/Rakefile:16:in `each'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/Rakefile:16:in `require_tests'
/home/spin/src/github.com/Shopify/sorbet/gems/sorbet-runtime/Rakefile:20:in `block in <top (required)>'
/opt/rubies/3.2.0/bin/bundle:25:in `load'
/opt/rubies/3.2.0/bin/bundle:25:in `<main>'
```

Thanks to an investigation by @paracycle, I've updated the test to use an approach that replaces the prepend with a method that works for pre and post 3.2.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/test/types/method_patches.rb
+++ b/gems/sorbet-runtime/test/types/method_patches.rb
@@ -5,7 +5,7 @@ module Opus::Types::Test
   class MethodPatchesTest < Critic::Unit::UnitTest
     module MethodRefinement
       refine Method do
-        T::CompatibilityPatches::MethodExtensions.each do |method|
+        T::CompatibilityPatches::MethodExtensions.instance_methods(false).each do |method|
           define_method(
             method,
             T::CompatibilityPatches::MethodExtensions.instance_method(method),
--- a/gems/sorbet-runtime/test/types/method_patches.rb
+++ b/gems/sorbet-runtime/test/types/method_patches.rb
@@ -5,7 +5,12 @@ module Opus::Types::Test
   class MethodPatchesTest < Critic::Unit::UnitTest
     module MethodRefinement
       refine Method do
-        prepend(T::CompatibilityPatches::MethodExtensions)
+        T::CompatibilityPatches::MethodExtensions.each do |method|
+          define_method(
+            method,
+            T::CompatibilityPatches::MethodExtensions.instance_method(method),
+          )
+        end
       end
     end
 


gems/sorbet-runtime/test/types/method_patches.rb
```diff
@@ -5,7 +5,11 @@ module Opus::Types::Test
   class MethodPatchesTest < Critic::Unit::UnitTest
     module MethodRefinement
       refine Method do
-        prepend(T::CompatibilityPatches::MethodExtensions)
+        if respond_to?(:refined_class)
+          refined_class.prepend(T::CompatibilityPatches::MethodExtensions)
```
paracycle, line 9: I don't think this will work the way you expect it to. The point of this being a refinement was so that it would not change the behaviour of `Method` when the refinement was not activated.

However, if the `prepend` is done against the refined class, then that will change the behaviour of that class for the rest of the test suite.

This needs a different approach.

paracycle (changes requested):

gems/sorbet-runtime/test/types/method_patches.rb
```diff
@@ -5,7 +5,11 @@ module Opus::Types::Test
   class MethodPatchesTest < Critic::Unit::UnitTest
     module MethodRefinement
       refine Method do
-        prepend(T::CompatibilityPatches::MethodExtensions)
+        T::CompatibilityPatches::MethodExtensions.each do |method|
+          define_method(
+            method,
+            T::CompatibilityPatches::MethodExtensions.instance_method(method),
+          )
```
paracycle, line 12: This is missing an `end` for the `define_method` block.

gems/sorbet-runtime/test/types/method_patches.rb
```diff
@@ -5,7 +5,12 @@ module Opus::Types::Test
   class MethodPatchesTest < Critic::Unit::UnitTest
     module MethodRefinement
       refine Method do
-        prepend(T::CompatibilityPatches::MethodExtensions)
+        T::CompatibilityPatches::MethodExtensions.each do |method|
```
paracycle, line 8: I am not sure what's happened here. I think you mean to iterate over the instance methods of the module, right?
```suggestion
        T::CompatibilityPatches::MethodExtensions.instance_methods(false).each do |method|
```

paracycle (approved):

ilyailya (approved):

PR #6871, opened 2023-03-21
Sprinkle some const in some places

jez:
### Motivation

I was suspicious of whether the act of caching these files was mutating
them. It doesn't seem to be the case.

### Test plan

See included automated tests.


froydnj (approved):
Nice!

PR #6884, opened 2023-03-30
Add flagged logic in autoloader to onboard all non-legacy packages to path-based autoloading

aadi-stripe:
Adds logic behind a new flag `--autogen-autoloader-pbal-non-annotated`, which will onboard all packages that don't have a `autoloader_compatibility 'legacy'` annotation to path-based autoloading (PBAL).

Also removes an unused flag `--autogen-autoloader-pbal-namespaces`.

### Motivation

Ensure that new code in the Stripe codebase is path-based-autoloaded by default.

### Test plan

See included automated tests. Tested on the Stripe codebase.

### DIFF
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -339,8 +339,9 @@ void DefTreeBuilder::markPackages(const core::GlobalState &gs, DefTree &root, co
 
     for (auto nr : gs.packageDB().packages()) {
         auto &pkg = gs.packageDB().getPackageInfo(nr);
-        if (pkg.strictAutoloaderCompatibility()) {
-            // Only mark strictly path-based autoload compatible packages for now to reduce
+        if (pkg.strictAutoloaderCompatibility() ||
+            (alCfg.pbalNonAnnotatedPackages && !pkg.legacyAutoloaderCompatibility())) {
+            // Only mark path-based autoload compatible packages for now to reduce
             // computation / code generation, given this is the only current use-case for registering
             // packages in this context in the Stripe codebase.
 
--- a/main/autogen/autoloader.cc
+++ b/main/autogen/autoloader.cc
@@ -38,13 +38,6 @@ bool AutoloaderConfig::sameFileCollapsable(const vector<core::NameRef> &module) 
     return !nonCollapsableModuleNames.contains(module);
 }
 
-bool AutoloaderConfig::registeredForPBAL(const vector<core::NameRef> &pkgParts) const {
-    return pbalNamespaces.empty() || (absl::c_any_of(pbalNamespaces, [&pkgParts](auto &pbalNamespace) {
-               return pbalNamespace.size() <= pkgParts.size() &&
-                      std::equal(pbalNamespace.begin(), pbalNamespace.end(), pkgParts.begin());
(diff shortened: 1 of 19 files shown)


test/cli/autogen-pkg-autoloader-non-annotated/test.sh
```diff
@@ -0,0 +1,44 @@
...
+touch "$inner_dir_to_delete/__file_to_delete.rb"
+
+"$cwd/main/sorbet" --silence-dev-message --stop-after=namer \
+  --stripe-packages \
+  -p autogen-autoloader:output \
+  --autogen-autoloader-modules=RootPackage \
+  --autogen-autoloader-exclude-require=byebug \
+  --autogen-autoloader-ignore=scripts/ \
+  --autogen-autoloader-preamble "$preamble" \
+  --autogen-autoloader-pbal-non-annotated \
```
aadi-stripe, line 28: New flag

main/autogen/autoloader.cc
```diff
@@ -352,18 +339,16 @@ void DefTreeBuilder::markPackages(const core::GlobalState &gs, DefTree &root, co
...
+            (alCfg.pbalNonAnnotatedPackages && !pkg.legacyAutoloaderCompatibility())) {
+            // Only mark path-based autoload compatible packages for now to reduce
             // computation / code generation, given this is the only current use-case for registering
             // packages in this context in the Stripe codebase.
 
             // Additionally this package must be registed for path-based autoloading.
             // TODO: (aadi-stripe, 10/24/2022) Remove this functionality once we no longer require
             // special registration.
             auto &pkgFullName = pkg.fullName();
-            if (!alCfg.registeredForPBAL(pkgFullName)) {
```
bhang-stripe, line 364: neat!

bhang-stripe (approved):

main/options/options.cc
```diff
@@ -422,9 +422,10 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                     "Modules that should never be collapsed into their parent. This helps break cycles "
                                     "in certain cases. (e.g. Foo::Bar::Baz)",
                                     cxxopts::value<vector<string>>());
-    options.add_options("advanced")("autogen-autoloader-pbal-namespaces",
-                                    "Namespaces for which path-based autoloading is enabled.",
-                                    cxxopts::value<vector<string>>());
+    options.add_options("advanced")("autogen-autoloader-pbal-non-annotated",
+                                    "When this flag is set, packages that don't have 'autoloader_compatibility' "
+                                    "annotations use path-based autoloading",
+                                    cxxopts::value<bool>());
```
jez, line 428: Why do we need this flag? Are there going to be some cases where we pass this flag and other cases where we don't?

Or are we just going to always pass this flag in all Sorbet invocations?

jez (approved):
I suppose this is fine but I'm a little confused as to the long term purpose of this flag.

main/options/options.cc
aadi-stripe: Right now the flag is there because I want to be able to bump Sorbet without having this functionality on by default. It will be removed very soon.

PR #6888, opened 2023-04-01
Use `new`, not `initialize`, to enforce runtime abstract

### LINKED ISSUE #3388: Correctly enforcing `abstract!`
It seems that `sorbet` enforces abstract classes not being instantiated [by defining an `initialize` that raises an error](https://github.com/sorbet/sorbet/issues/2374#issuecomment-567601210)

This `initialize` apparently does not check what is the class being instantiated, is there a good reason for this?

Moreover, why not define a class method `new` that raises instead (only if the receiver is the current abstract class)?

The reasons to go this way are:

1) There are valid reasons to have an abstract class that need initialization and/or initial parameters. `sorbet` shouldn't mess with `initialize`

2) There are valid cases of an abstract class that does not require special initialization and/or additional parameters. `sorbet` shouldn't force having a trivial `initialize` method

3) Calling `super` from `initialize` is considered by many a good practice. `sorbet` should not break this.

Example:
```ruby
class Foo
  def self.new(*)
    return super unless self == Foo

    raise 'Can not instantiate abstract class'
  end

  def initialize
    @registry = {}
    puts "initialized, ok"
  end
end

class Bar < Foo
end

Foo.new rescue puts 'raised, ok'
Bar.new and puts 'ok'
# prints "ok" 3 times...
```

jez:
**Note to users**: this PR might cause new errors in any code that was
previously defining `initialize` in an `abstract!` class and then trying to
instantiate it at runtime.

Doing that would have had the effect of redefining the method that
`sorbet-runtime` previously defined to enforce that abstract classes can't be
instantiated, which would have allowed the instantiation. This will not happen
anymore.

### Motivation

Fixes #3388

The general idea is that people want both:

- runtime enforcement that abstract classes are not instantiated
- the ability to have common initialization code in an abstract class's
  constructor

By moving the "no new" enforcement to happen in `self.new` instead of
`initialize`, we can accomplish this. It's far, far more rare for people to have
valid use cases for overriding `self.new`.

This also unblocks being able to override `initialize` at runtime and mark it
`abstract`, so that it can be used to enforce that all subclasses conform to a
standard contract for constructing new instances.

### Test plan

See included automated tests.

I tested this against Stripe's codebase and 17 tests failed that were errantly
attempting to instantiate an abstract class at runtime in a way that Sorbet
couldn't catch statically.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/abstract/declare.rb
+++ b/gems/sorbet-runtime/lib/types/private/abstract/declare.rb
@@ -27,15 +27,15 @@ module T::Private::Abstract::Declare
         raise "Classes can't be interfaces. Use `abstract!` instead of `interface!`."
       end
 
-      if mod.instance_method(:initialize).owner == mod
-        raise "You must call `abstract!` *before* defining an initialize method"
+      if mod.method(:new).owner == mod
+        raise "You must call `abstract!` *before* defining a `new` method"
       end
 
       # Don't need to silence warnings via without_ruby_warnings when calling
       # define_method because of the guard above
 
-      mod.send(:define_method, :initialize) do |*args, &blk|
-        if self.class == mod
+      mod.send(:define_singleton_method, :new) do |*args, &blk|
+        if T.unsafe(self) == mod
           raise "#{mod} is declared as abstract; it cannot be instantiated"
         end
         super(*args, &blk)
@@ -43,10 +43,10 @@ module T::Private::Abstract::Declare
 
       # Ruby doesn not emit "method redefined" warnings for aliased methods
       # (more robust than undef_method that would create a small window in which the method doesn't exist)
-      mod.send(:alias_method, :initialize, :initialize)
+      mod.singleton_class.send(:alias_method, :new, :new)
 
-      if mod.respond_to?(:ruby2_keywords, true)
-        mod.send(:ruby2_keywords, :initialize)
+      if mod.singleton_class.respond_to?(:ruby2_keywords, true)
+        mod.singleton_class.send(:ruby2_keywords, :new)
       end
     end
   end
--- a/gems/sorbet-runtime/test/types/validate_override_shape.rb
+++ b/gems/sorbet-runtime/test/types/validate_override_shape.rb
@@ -227,5 +227,21 @@ module Opus::Types::Test
       end
       assert_equal(0, klass.new.foo)
     end
+
+    it "raises if initialize is not compatible with parent" do
+      klass = Class.new(AbstractBase) do
+        extend T::Sig
+        sig do
+          override
+          .params(x: Integer)
+          .void
+        end
+        def initialize(x); end
+      end
+      err = assert_raises(RuntimeError) do
+        klass.new(0)
+      end
+      assert_includes(err.message, "must have no more than 0 required argument(s) to be compatible")
+    end
   end
 end
--- a/gems/sorbet-runtime/test/types/validate_override_shape.rb
+++ b/gems/sorbet-runtime/test/types/validate_override_shape.rb
@@ -220,7 +220,10 @@ module Opus::Types::Test
         extend T::Sig
         sig {override.void}
         def initialize; end
-        def foo; 0; end
+
+        def foo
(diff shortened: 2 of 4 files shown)


gems/sorbet-runtime/lib/types/private/abstract/declare.rb
```diff
@@ -27,26 +27,26 @@ def self.declare_abstract(mod, type:)
         raise "Classes can't be interfaces. Use `abstract!` instead of `interface!`."
       end
 
-      if mod.instance_method(:initialize).owner == mod
-        raise "You must call `abstract!` *before* defining an initialize method"
+      if mod.method(:new).owner == mod
+        raise "You must call `abstract!` *before* defining a `new` method"
```
froydnj, line 31: WDYT about also warning people that you shouldn't define `new` on `abstract!` classes?

froydnj (approved):
I think this makes sense.  Apologies for the delayed review.

gems/sorbet-runtime/lib/types/private/abstract/declare.rb
jez: It's interesting. I think it's probably a good check to have? I'm struggling to think of a case where you'd want to define `self.new` and not `initialize`.

But also I'm generally afraid of rocking the boat too much on runtime changes like this, because people do weird things.

So I'd probably prefer to investigate something like raising an error for any `self.new` method defined in an abstract class in a follow up change. Or alternatively, if we do just want to emit a warning, not an error, we could get rid of the `alias_method` trick we do later in the file, which would then use the Ruby VM's built in method redefinition warnings to surface this to the user.

gems/sorbet-runtime/test/types/validate_override_shape.rb
```diff
@@ -178,5 +184,18 @@ def foo; end
       end
       klass.new.foo
     end
+
+    it "does opt-in override checking on initialize" do
+      klass = Class.new(AbstractBase) do
+        extend T::Sig
+        sig {override.void}
+        def initialize; end
```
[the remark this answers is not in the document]
jez: done.

jez:

jez:

→ https://go/builds/bui_No3dBatUPfwnyR
→ https://go/builds/bui_No3dy8VqUCeTla
→ https://go/builds/bui_No3dxnAZ6ONGqd
→ https://go/builds/bui_No3dxnAZ6ONGqd

jez:

→ https://go/builds/bui_NrM8Klkuvtef99
→ https://go/builds/bui_NrM8Ld9XhkH1QA
→ https://go/builds/bui_NrM86zREjDsYVO

→ https://go/builds/bui_NrMDNVKbtQwaHh

gems/sorbet-runtime/lib/types/private/abstract/declare.rb
cbothner: https://github.com/rails/rails/blob/ef04fbb3b256beececfa44c47c4ec93ac6945e59/activerecord/lib/active_record/inheritance.rb#L56-L78

Rails defines `self.new` in order to implement single-table inheritance. `new` is called on the STI base class, which delegates to the appropriate child class based on the value of the `type` attribute.

Consider this example:
- `Market.has_many :regions`
- `Region < ActiveRecord::Base` and is marked `abstract!`
- `Country < Region`
- `Province < Region`

Before this change (desired behavior IMO)
```
[1] pry(main)> market.regions.build
RuntimeError: Region is declared as abstract; it cannot be instantiated
[2] pry(main)> market.regions.build(type: Country)
=> #<Country:0x00007f97c3b1a3c0 ... >
```

After this change:
```
[1] pry(main)> market.regions.build
RuntimeError: Region is declared as abstract; it cannot be instantiated
[2] pry(main)> market.regions.build(type: Country)
RuntimeError: Region is declared as abstract; it cannot be instantiated
```

paracycle: @cbothner https://github.com/sorbet/sorbet/pull/7079

PR #6890, opened 2023-04-01
Allow overriding an abstract method keeping it abstract

### LINKED ISSUE #4431: Impossible to narrow type of abstract parent method keeping it abstract
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.returns(T.any(Integer%2C%20String))%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%0A%20%20%23%20This%20raises%20at%20runtime%0A%20%20%23%20Make%20it%20so%20that%20it%20doesn't%20require%20override.%0A%20%20sig%20%7Babstract.returns(Integer)%7D%0A%20%20def%20foo%3B%20end%0Aend%0A)

```ruby
# typed: true

class Parent
  extend T::Sig
  extend T::Helpers
  abstract!

  sig {abstract.returns(T.any(Integer, String))}
  def foo; end
end

class Child < Parent
  extend T::Sig
  extend T::Helpers
  abstract!

  # This raises at runtime
  # Make it so that it doesn't require override.
  sig {abstract.returns(Integer)}
  def foo; end
end
```

#### Observed output

```
editor.rb:20: Method Child#foo implements an abstract method Parent#foo but is not declared with override. https://srb.help/5051
    20 |  def foo; end
          ^^^^^^^
    editor.rb:9: defined here
     9 |  def foo; end
          ^^^^^^^
Errors: 1
```

#### Expected behavior

There should be a way to narrow the type of an abstract method while keeping it abstract (as long as the variance checks out).

- - -

Right now

- `abstract.override` is syntactically rejected
- `abstract` alone emits an error for "overriding an abstract method but not putting override"

Context: this error was added in #3774, because the runtime is made to raise an error in such a case.

I think this means that we should allow `abstract` alone (without mentioning override) to narrow the type. Specifically, that means that if a parent `abstract` method exists, it should do the same variance checking that adding `override` to a method should.

jez:
### Motivation

Fixes #4431

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/resolver/abstract_abstract.rb
+++ b/test/testdata/resolver/abstract_abstract.rb
@@ -1,0 +1,23 @@
+# typed: true
+
+class Parent
+  extend T::Sig
+  extend T::Helpers
+  abstract!
+
+  sig {abstract.returns(T.nilable(Integer))}
+  def example1; end
+
+  sig {abstract.returns(Integer)}
+  def example2; end
+end
+
+class Child < Parent
+  abstract!
+
+  sig {abstract.returns(Integer)}
+  def example1; end
+
+  sig {abstract.returns(T.nilable(Integer))}
+  def example2; end # error: Return type `T.nilable(Integer)` does not match return type of abstract method `Parent#example2`
+end
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -494,7 +494,7 @@ void validateOverriding(const core::Context ctx, core::MethodRef method) {
                 e.addErrorLine(overridenMethod.data(ctx)->loc(), "defined here");
             }
         }
-        if (!method.data(ctx)->flags.isOverride && method.data(ctx)->hasSig() &&
+        if (!method.data(ctx)->flags.isOverride && !method.data(ctx)->flags.isAbstract && method.data(ctx)->hasSig() &&
             overridenMethod.data(ctx)->flags.isAbstract && overridenMethod.data(ctx)->hasSig() &&
             !method.data(ctx)->flags.isRewriterSynthesized && !isRBI) {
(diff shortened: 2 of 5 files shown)


test/testdata/resolver/abstract_abstract.rb
```diff
@@ -0,0 +1,23 @@
...
+end
+
+class Child < Parent
+  abstract!
+
+  sig {abstract.returns(Integer)}
+  def example1; end
+
+  sig {abstract.returns(T.nilable(Integer))}
+  def example2; end # error: Return type `T.nilable(Integer)` does not match return type of abstract method `Parent#example2`
```
neilparikh, line 22: Do you think it would be possible/worth it to update this error message to convey that it is also allowed to narrow the type? Right now it gives the impression that the return type must be the exact same.

neilparikh (approved):
One suggestion/discussion regarding the existing error message, but that can be fixed in a future PR (and probably should be imo).

test/testdata/resolver/abstract_abstract.rb
jez: It says that as an error note already:

Screenshot 2023-04-03 at 3 54 53 PM

PR #6903, opened 2023-04-06
Make T::Private::Types::Void a singleton

casperisfine:
From heap-profiler:

```
retained objects by class
-----------------------------------
     13906  T::Private::Types::Void
```

But all these instances are purely identical, so we can save a bit of memory and make it easier on the GC by making it a singleton.

cc @paracycle @jez


jez:
http://go/builds/bui_Nff0vrvdsl1i5j

PR #6907, opened 2023-04-06
Work around rubygems.org being flaky

jez:
### Motivation

Our nightly rubygems.org publish has failed ~every day for weeks.

### Test plan

Some failing builds:

https://buildkite.com/sorbet/sorbet/builds/26828#01873618-47b5-4845-9458-54102c8c661a

https://buildkite.com/sorbet/sorbet/builds/26905#018754fd-e88d-485a-93e6-b46794bd927d

We see that the problem appears to be that there is a "timeout waiting for first byte" and then on retry, we get an error saying "this version already published"

I plan to test that these changes work by manually kicking off a gem publish build when this lands on master.

### DIFF
--- a/.buildkite/publish-ruby-gems.sh
+++ b/.buildkite/publish-ruby-gems.sh
@@ -46,26 +46,23 @@ for gem_archive in "_out_/gems/sorbet-static-$release_version"-*.gem; do
   fi
 done
 
-gem_archive="_out_/gems/sorbet-runtime-$release_version.gem"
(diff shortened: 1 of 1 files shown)


neilparikh (approved):
Just to make sure my understanding is correct, is the following what happened before?
- gem list
- gem push
- gem push
- ...

And the new logic is:
- gem list
- gem push
- gem list
- gem push
- ...

jez:
yep

PR #6909, opened 2023-04-07
Hide `<AttachedClass>` constants from autogen

jez:
### Motivation

This feature caused some problems with our preloader, because it made it look
like certain constants were defined when they actually weren't.

### Test plan

Review by commit to see the test before+after.

### DIFF
--- a/main/autogen/autogen.cc
+++ b/main/autogen/autogen.cc
@@ -199,6 +199,11 @@ public:
         if (original.original == nullptr) {
             return;
         }
+        if (original.symbol.name(ctx) == core::Names::Constants::AttachedClass()) {
+            // This is a reference to a constant like <AttachedClass> that came from the `has_attached_class!` DSL
+            // These are not real constant references.
+            return;
+        }
 
         // Create a new `Reference`
         auto &ref = refs.emplace_back();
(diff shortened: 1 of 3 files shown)


froydnj:
(Alternatively, should we just disable the rewriter pass for `has_attached_class` when running with autogen, like so many other rewriter passes?)

jez:
> (Alternatively, should we just disable the rewriter pass for `has_attached_class` when running with autogen, like so many other rewriter passes?)

While we could do this, I think it's better to let autogen have the full view of the codebase as Sorbet sees it, and let it make decisions about what to include/not include in its exports.

The fact that we have to disable so many errors when running for autogen is because we skip so many rewriter passes, and disabling those errors historically makes it hard to understand why there's a problem (Sorbet was trying to say there's a problem, but we prevented it from saying so).

jez:

→ https://go/builds/bui_NffV0kVbfuIWwK
→ https://go/builds/bui_NffVKIg5to0oWq

jez:

→ https://go/buildgroup/bg_Nffk7sDMxiQSdr
→ https://go/builds/bui_NffkhxGcBguAcO

PR #6910, opened 2023-04-07
Support Ruby 3.2 forwarded rest and keyword rest args

egiurleo:
### Motivation
This PR adds Sorbet support for forwarded rest and keyword rest args, which were [introduced in Ruby 3.2](https://bugs.ruby-lang.org/issues/18351).

For example:
```
def foo(*)
  bar(*)
end

def baz(**)
  quux(**)
end
```

This PR can be read commit-by-commit. The first few commits add parsing for `ForwardedRestArg` and `ForwardedKwrestArg` nodes and have corresponding commits in the whitequark parser. The remaining commits add support for these nodes in `Desugar.cc`. I heavily based the implementation there on the implementation of forwarded args (`def foo(...); bar(...); end`).

### Test plan
See included automated tests.


jez (approved):

akefeli:
How do you reference the `**` in the method signature?

PR #6912, opened 2023-04-07
Fix Opus::Command in `# typed: strong`

jez:
### Motivation

Less untyped

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Command.cc
+++ b/rewriter/Command.cc
@@ -85,7 +85,7 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
     flags.isSelfMethod = true;
     flags.discardDef = true;
     auto selfCall = ast::MK::SyntheticMethod(call->loc, call->loc, call->name, std::move(newArgs),
-                                             ast::MK::UntypedNil(call->loc), flags);
+                                             ast::MK::RaiseTypedUnimplemented(call->declLoc), flags);
 
     klass->rhs.insert(klass->rhs.begin() + i + 1, sig->deepCopy());
     klass->rhs.insert(klass->rhs.begin() + i + 2, std::move(selfCall));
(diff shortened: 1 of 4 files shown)


jez:
* master
  * **PR #6912** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6912" target="_blank">Graphite</a>  👈
    * **PR #6913** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6913" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6912?utm_source=stack-comment).

rewriter/Command.cc
```diff
@@ -85,7 +85,7 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
     flags.isSelfMethod = true;
     flags.discardDef = true;
     auto selfCall = ast::MK::SyntheticMethod(call->loc, call->loc, call->name, std::move(newArgs),
-                                             ast::MK::UntypedNil(call->loc), flags);
+                                             ast::MK::RaiseTypedUnimplemented(call->declLoc), flags);
```
neilparikh, line 88: I was thinking that this wouldn't work because this would lead inference to think the return type of the `call` method is `T.noreturn` and then that would conflict with whatever the user declared as the return type themselves.

However, it does work. Is that because `T.noreturn` is bottom, and thus a subtype of anything the user declared as the return type?

neilparikh (approved):

rewriter/Command.cc
jez: > Is that because T.noreturn is bottom, and thus a subtype of anything the user declared as the return type?

yep

jez: > Is that because T.noreturn is bottom, and thus a subtype of anything the user declared as the return type?

yep

PR #6914, opened 2023-04-07
Add the self parameter for `instance_eval` on `BasicObject`

maruth-stripe:
### Motivation

Before this change needed an escape hatch to use the `obj` parameter in the block version of `instance_eval`

```ruby
# typed: true
extend T::Sig

5.instance_eval {|x| T.reveal_type(x) }
```

current output
```
Revealed type: NilClass https://srb.help/7014
     4 |5.instance_eval {|x| T.reveal_type(x) }
                             ^^^^^^^^^^^^^^^^
    editor.rb:4:
     4 |5.instance_eval {|x| T.reveal_type(x) }
                          ^
Errors: 1
```

Ideally the revealed type should be the type of the receiver (~`T.self_type`) but in this change we settle for `T.untyped`

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/rbi/basicobject_instance_eval.rb
+++ b/test/testdata/rbi/basicobject_instance_eval.rb
@@ -4,7 +4,7 @@ class A
   sig do
     type_parameters(:U)
     .params(
-        blk: T.proc.bind(T.untyped).param(arg0: T.untyped).returns(T.type_parameter(:U)),
+        blk: T.proc.bind(T.untyped).params(arg0: T.untyped).returns(T.type_parameter(:U)),
     )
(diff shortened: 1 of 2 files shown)


test/testdata/rbi/basicobject_instance_eval.rb
```diff
@@ -4,14 +4,15 @@ class A
   sig do
```
neilparikh, line 4: We can just delete this method and call the method declared in `BasicObject`, since `BasicObject` is an ancestor of `A` (I'm not really sure why this test redefines it).

neilparikh (approved):

PR #6916, opened 2023-04-08
Array#each -> `void`

jez:
### Motivation

We might want to change other `each` methods, but `Enumerable#each` is
`abstract`, and so that would be more of a backwards incompatible
change.

### Test plan

See included automated tests.


jez:

→ https://go/buildgroup/bg_Nfl4433e1DtCig
→ https://go/builds/bui_Nfl4dKdVXVJ2rh

PR #6917, opened 2023-04-08
Implement `T.anything`, Sorbet's top type

jez:
### Motivation

As we move towards tracking `T.untyped` more carefully, I keep running into
places where I'm getting errors that say something like "there's `T.untyped`"
here. Specifically, I'm trying to make Sorbet show you if in `foo(x)` the `arg0`
of `foo` is actually `T.untyped`, thus defeating type safety at this argument
site.

But those errors almost always end up as false-positives, because someone was
using `T.untyped` to mean, "truly, you can give me anything" (or as a block
return: "truly, you can return anything from this block, I don't care").

In some places, we use `BasicObject` for this right now, but `BasicObject`
doesn't quite work, because the interface `IFoo` is not a subtype of
`BasicObject`, nor is `T.type_parameter(:U)`.

### Test plan

See included automated tests.


ilyailya (approved):
Awesome change!

PR #6920, opened 2023-04-09
Fix return type of String#squeeze!

iownthegame:
### Motivation

Ruby doc: https://docs.ruby-lang.org/en/2.7.0/String.html#method-i-squeeze-21
Also mentioned in the comment that the return value can be `nil` https://github.com/iownthegame/sorbet/blob/8000928e84da9921feab6786dbc494d9d7607ebd/rbi/core/string.rbi#L1975

### Test plan

See included automated tests.


jez (approved):

jez:
This change looks good, but I'm not able to run CI and thus unable to merge it because your change is based on a commit that is 1 year old, and our CI environment cannot still test a commit that is that old.

Can you rebase or merge using the lastest master commit, and then re-request a review?

jez (approved):

PR #6921, opened 2023-04-10
explicitly use one thread per core when using thinlto

amytsai-stripe:
Applies the option `--thinlto-jobs=all` to building the main sorbet target and jemalloc

### Motivation

Improves build times on graviton / arm64 machines

### Test plan

See included automated tests.

### DIFF
--- a/third_party/jemalloc.BUILD
+++ b/third_party/jemalloc.BUILD
@@ -29,7 +29,7 @@ JEMALLOC_BUILD_COMMAND = """
   export OBJCOPY=$$(absolutize $(OBJCOPY))
   export CFLAGS=$(CC_FLAGS)
   export CXXFLAGS=$(CC_FLAGS)
-  export LTOFLAGS="$$([ "$$(uname)" = "Linux" ] && echo "-flto=thin")" # todo: on next clang toolchain upgrade, check if it's fixed and we can re-enable thinlto on mac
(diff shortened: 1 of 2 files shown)


third_party/jemalloc.BUILD
```diff
@@ -29,7 +29,7 @@ JEMALLOC_BUILD_COMMAND = """
   export OBJCOPY=$$(absolutize $(OBJCOPY))
   export CFLAGS=$(CC_FLAGS)
   export CXXFLAGS=$(CC_FLAGS)
-  export LTOFLAGS="$$([ "$$(uname)" = "Linux" ] && echo "-flto=thin")" # todo: on next clang toolchain upgrade, check if it's fixed and we can re-enable thinlto on mac
+  export LTOFLAGS="$$([ "$$(uname)" = "Linux" ] && echo "-flto=thin -Wl,--thinlto-jobs=all")" # todo: on next clang toolchain upgrade, check if it's fixed and we can re-enable thinlto on mac
```
froydnj, line 32: I'm not entirely sure this is needed, since (AFAIK) jemalloc's build isn't going to be producing executables or shared libraries (and therefore invoking the LTO bits of the linker), but at the very least this cannot hurt things, so we'll leave it in for the time being.

froydnj (approved):
Thanks for tracking this down!

PR #6922, opened 2023-04-11
Raise deserialize errors if an enum value is invalid

cyberdelia:
This allows to receive a more useful error when deserializing an invalid value for an enum type. This is a fairly broad fix, but hopefully we shouldn't encounter `KeyError` in deserialization normal process?

### Motivation

If you deserialize a `Struct` with an Enum type today, you get a `KeyError` with no useful information has to what property is invalid nor which value was used for said property. 

### Test plan

See included automated tests.


cyberdelia:
This is somewhat related to #6631.

Issue #6878, opened 2023-03-23
typed: strong does not support `rescue`


jflinter:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strong%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bvoid%7D%0Adef%20foo%0A%20%20begin%0A%20%20%20%20puts%20'hi'%0A%20%20rescue%20StandardError%20%3D%3E%20e%0A%20%20%20%20puts%20e%0A%20%20end%0Aend%0A)

```ruby
# typed: strong
extend T::Sig

sig {void}
def foo
  begin
    puts 'hi'
  rescue StandardError => e
    puts e
  end
end

```

#### Observed output

```
editor.rb:8: This code is untyped https://srb.help/7018
     8 |  rescue StandardError => e
                 ^^^^^^^^^^^^^
  Note:
    Support for typed: strong is minimal. Consider using typed: strict instead.
Errors: 1
```

#### Expected behavior

I realize support for typed: strong is minimal, but it's actually pretty plausible for our application, except for this, which seems like (i hope!) a small fix?

froydnj:
I believe this was fixed in #7349; the linked sorbet.run in the PR description now reports no errors.

PR #6908, opened 2023-04-07
More targeted `# typed: strong` errors

jez:
### Motivation

`# typed: strong` is bad, we can do better.

### Test plan

See included automated tests.

PR #6928, opened 2023-04-12
give `rescue` better types

froydnj:
### Motivation

Less noisy squiggles in vscode when you use exception handling in typed files and are highlighting untyped code.

The obvious thing in this PR is giving `GetCurrentException` a real type.  This works for _almost_ everything, except for:

```ruby
# test/testdata/cfg/rescue_complex.rb
class TestRescue
  def meth; 0; end
  def bar; 2; end

  def initialize
    @ex = T.let(nil, T.nilable(StandardError))
  end

  def parse_resbody_var_1()
    begin; meth; rescue => @ex; bar; end
  end
end
```

Because we wind up with control flow that looks like:

```
# backedges
# - bb0(rubyRegionId=0)
# - bb4(rubyRegionId=1)
bb3[rubyRegionId=2, firstDead=-1](<self>: TestRescue, <returnMethodTemp>$2: T.untyped, <exceptionValue>$3: Exception, <magic>$5: T.class_of(<Magic>), @ex$11: T.nilable(StandardError)):
    <rescueTemp>$2: Exception = <exceptionValue>$3
    <cfgAlias>$8: T.class_of(StandardError) = alias <C StandardError>
    <isaCheckTemp>$9: T::Boolean = <cfgAlias>$8: T.class_of(StandardError).===(<exceptionValue>$3: Exception)
    <isaCheckTemp>$9 -> (T::Boolean ? bb7 : bb8)

# backedges
# - bb3(rubyRegionId=2)
bb7[rubyRegionId=2, firstDead=-1](<self>: TestRescue, <magic>$5: T.class_of(<Magic>), <rescueTemp>$2: Exception, @ex$11: T.nilable(StandardError)):
    <exceptionValue>$3: NilClass = nil
    <keepForCfgTemp>$6: Sorbet::Private::Static::Void = <magic>$5: T.class_of(<Magic>).<keep-for-cfg>(<exceptionValue>$3: NilClass)
    @ex$11: T.nilable(StandardError) = <rescueTemp>$2
    <returnMethodTemp>$2: T.untyped = <self>: TestRescue.bar()
    <unconditional> -> bb6
```

bb3 is testing whether the current exception is an instance of `StandardError`; bb7 is doing the actual assignment to `@ex`.  Except that bb3 is using the moral equivalent of:

```ruby
StandardError.===(<exceptionValue>)
```

to do the test, and bb7 is using `<rescueTemp>` to do the assignment to `@ex`.  So the information about `StandardError`-ness is stored for `<exceptionValue>`, which we don't care about.  We should have known in bb7 that `<rescueTemp>` is a `StandardError`, and we actually constructed the CFG as:

```ruby
StandardError.===(<rescueTemp>)
```

but we were defeated by dealiasing hereabouts:

https://github.com/sorbet/sorbet/blob/9864768a0854ab8a511f30ce9665a030ce53e0d7/cfg/builder/builder_finalize.cc#L191-L229

because we noticed that `<rescueTemp>` and `<exceptionValue>` are the same thing, thanks to the assignment at the start of bb3.  So we had propagated `<exceptionValue>` into the send, and inhibited Sorbet's inference pass from doing its thing.

The fix is to mark that send as a synthesized expression, so we won't propagate aliases into it.

The CLI test change is...not the greatest, but I think it is acceptable in light of the other changes.

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/dedup_loc/test.out
+++ b/test/cli/dedup_loc/test.out
@@ -5,7 +5,4 @@ test/cli/dedup_loc/dedup_loc.rb:5: Argument does not have asserted type `NilClas
     test/cli/dedup_loc/dedup_loc.rb:4:
      4 |rescue Exception => e
         ^^^^^^
-    test/cli/dedup_loc/dedup_loc.rb:4:
-     4 |rescue Exception => e
-               ^^^^^^^^^
 Errors: 1
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -862,10 +862,10 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                         args.emplace_back(localVar);
 
                         auto isPrivateOk = false;
-                        rescueHandlersBlock->exprs.emplace_back(
-                            isaCheck, loc,
-                            make_insn<Send>(exceptionClass, loc, core::Names::tripleEq(), loc.copyWithZeroLength(),
-                                            args.size(), args, std::move(argLocs), isPrivateOk));
+                        synthesizeExpr(rescueHandlersBlock, isaCheck, loc,
+                                       make_insn<Send>(exceptionClass, loc, core::Names::tripleEq(),
+                                                       loc.copyWithZeroLength(), args.size(), args, std::move(argLocs),
+                                                       isPrivateOk));
 
                         auto otherHandlerBlock = cctx.inWhat.freshBlock(cctx.loops, handlersRubyRegionId);
                         conditionalJump(rescueHandlersBlock, isaCheck, caseBody, otherHandlerBlock, cctx.inWhat, loc);
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1491,7 +1491,8 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                 tp.origins.emplace_back(ctx.locAt(bind.loc));
             },
             [&](cfg::GetCurrentException &i) {
-                tp.type = core::Types::untypedUntracked();
+                tp.type = core::Types::any(ctx, core::make_type<core::ClassType>(core::Symbols::Exception()),
+                                           core::Types::nilClass());
                 tp.origins.emplace_back(ctx.locAt(bind.loc));
             },
             [&](cfg::LoadSelf &l) {
--- a/test/testdata/cfg/dealias_with_return.rb.cfg-text.exp
+++ b/test/testdata/cfg/dealias_with_return.rb.cfg-text.exp
@@ -3,8 +3,8 @@ method ::Object#a {
 bb0[rubyRegionId=0, firstDead=-1]():
     <self>: Object = cast(<self>: NilClass, Object);
     <magic>$6: T.class_of(<Magic>) = alias <C <Magic>>
-    <exceptionValue>$4: T.untyped = <get-current-exception>
-    <exceptionValue>$4 -> (T.untyped ? bb3 : bb4)
+    <exceptionValue>$4: T.nilable(Exception) = <get-current-exception>
+    <exceptionValue>$4 -> (T.nilable(Exception) ? bb3 : bb4)
 
 # backedges
 # - bb4(rubyRegionId=1)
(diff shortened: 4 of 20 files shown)


froydnj:

→ https://go/builds/bui_NhTPIcxrJmBdZ8
→ https://go/builds/bui_NhTP8mIzPZbG4M
→ https://go/builds/bui_NhTPgn8YO4tvli

neilparikh (approved):
Can you add a test case for the aliasing issue you mention in the description? `test/testdata/cfg/rescue_complex.rb` is `typed: true`, so it's not being tested right now.

Look great otherwise!

jez:
Fixes #6878

test/cli/dedup_loc/test.out
```diff
@@ -5,7 +5,4 @@ test/cli/dedup_loc/dedup_loc.rb:5: Argument does not have asserted type `NilClas
     test/cli/dedup_loc/dedup_loc.rb:4:
      4 |rescue Exception => e
         ^^^^^^
-    test/cli/dedup_loc/dedup_loc.rb:4:
-     4 |rescue Exception => e
-               ^^^^^^^^^
```
jez, line 10: trying to understand this. on your branch:

Screenshot 2023-04-12 at 8 14 44 PM

super weird. it appears that this only happens when you put `Exception => e` specifically? if that's the case honestly i'm fine with that, but I wonder if the fact that `foo.rb` in my example above behaves differently from dedup_loc.rb gives you any ideas for how to adjust the loc

I will note that the loc I chose for the `GetCurrentException` instruction is a hack anyways. We _may_ want to put it back to the whole `rescue...end` block? i.e. revert this commit:

https://github.com/sorbet/sorbet/pull/6908/commits/3ea6ff854d7ca41b54185740185d2fd65cf9ab2d

That would at least make it clear that it's the "whole rescue block" participating in the source of `T.nilable(Exception)`, instead of somewhat misleadingly pointing to the first `rescue` clause, which may not mention `Exception`.

Anyways, whacky stuff, thanks for mentioning something in the PR description.

(this is non blocking tbc)

jez:
Some of the errors in your Stripe test build are great, and some of them are suspect?

Search the output for `CreateMiniSession`. That error shows that the revealed type is `Exception` but the rescue clause is rescuing a specific exception subclass, and I'm not sure why Sorbet isn't able to narrow the type.

froydnj:
> Some of the errors in your Stripe test build are great, and some of them are suspect?
> 
> Search the output for `CreateMiniSession`. That error shows that the revealed type is `Exception` but the rescue clause is rescuing a specific exception subclass, and I'm not sure why Sorbet isn't able to narrow the type.

That particular one was fixed in https://go/pr/605374 .  I'm not sure how to provide a better error message there, given Ruby semantics.

froydnj:
> Can you add a test case for the aliasing issue you mention in the description? `test/testdata/cfg/rescue_complex.rb` is `typed: true`, so it's not being tested right now.

`test/testdata/cfg/rescue_complex.rb` is the testcase for the aliasing issue.  Without the `synthesizeExpr` CFG construction fix, we get an error about assigning to `@ex`:

```ruby
  def parse_resbody_var_1()
    begin; meth; rescue => @ex; bar; end
  end
```

Because we declared its type as `T.nilable(StandardError)` and the dealiasing tries to assign `Exception` to it.

neilparikh:
> test/testdata/cfg/rescue_complex.rb is the testcase for the aliasing issue.

Sorry, what I actually meant was testing that the `rescue => @ex` pattern is now allowed in `# typed: strong`, since it [currently isn't allowed](https://sorbet.run/#%23%20typed%3A%20strong%0Aextend%20T%3A%3ASig%0A%0A%23%20test%2Ftestdata%2Fcfg%2Frescue_complex.rb%0Aclass%20TestRescue%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7B%20returns%28Integer%29%20%7D%0A%20%20def%20meth%3B%200%3B%20end%0A%20%20sig%20%7B%20returns%28Integer%29%20%7D%0A%20%20def%20bar%3B%202%3B%20end%0A%0A%20%20sig%20%7B%20void%20%7D%0A%20%20def%20initialize%0A%20%20%20%20%40ex%20%3D%20T.let%28nil%2C%20T.nilable%28StandardError%29%29%0A%20%20end%0A%0A%20%20sig%20%7B%20returns%28Integer%29%20%7D%0A%20%20def%20parse_resbody_var_1%28%29%0A%20%20%20%20begin%3B%20meth%3B%20rescue%20%3D%3E%20%40ex%3B%20bar%3B%20end%0A%20%20end%0Aend), but this PR makes it work.

jez:
@froydnj another torture test for the changes in this branch:

https://sorbet.run/#%23%20typed%3A%20strong%0Aextend%20T%3A%3ASig%0A%0Asig%20do%0A%20%20type_parameters%28%3AInstance%29%0A%20%20%20%20.params%28%0A%20%20%20%20%20%20exn%3A%20T.all%28T.class_of%28Exception%29%2C%20T%3A%3AClass%5BT.type_parameter%28%3AInstance%29%5D%29%2C%0A%20%20%20%20%20%20blk%3A%20T.proc.void%0A%20%20%20%20%29%0A%20%20%20%20.returns%28T.type_parameter%28%3AInstance%29%29%0Aend%0Adef%20assert_raises%28*exn%2C%20%26blk%29%0A%20%20begin%0A%20%20%20%20yield%0A%20%20rescue%20*exn%20%3D%3E%20e%0A%20%20%20%20T.reveal_type%28e%29%0A%20%20%20%20return%20e%0A%20%20end%0A%20%20%0A%20%20raise%20%22Did%20not%20raise%20an%20exception%22%0Aend%0A%0Aexn%20%3D%20assert_raises%28TypeError%2C%20ArgumentError%29%20do%0Aend%0AT.reveal_type%28exn%29

It would be really neat if Sorbet could figure out the `T.reveal_type(e)` line.

In any case, not something to block on, but something we might want to add to the test suite.

jez:
Landed in #7349

PR #6932, opened 2023-04-12
Allow --suggest-unsafe to suggest all-untyped sigs

jez:
### Motivation

I can imagine that this helps people who want to codemod files up to
`# typed: strict`.

I can also imagine this being useful in a future where it's possible to
toggle the `suggestUnsafe` boolean on and off inside an editor, which would then
allow applying these autocorrects as quickfixes.

(Once we build support for toggling untyped highlighting, we can likely extend
that feature to support `suggestUnsafe` and other GlobalState booleans too)

### Test plan

See included automated tests.

### DIFF
--- a/infer/SigSuggestion.cc
+++ b/infer/SigSuggestion.cc
@@ -474,7 +474,11 @@ optional<core::AutocorrectSuggestion> SigSuggestion::maybeSuggestSig(core::Conte
         }
         fmt::format_to(std::back_inserter(ss), ").");
     }
-    if (!guessedSomethingUseful) {
+    if (!guessedSomethingUseful && !ctx.state.suggestUnsafe.has_value()) {
+        // We don't want to condition people to start inserting a bunch of useless signatures filled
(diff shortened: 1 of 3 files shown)


infer/SigSuggestion.cc
```diff
@@ -474,7 +474,11 @@ optional<core::AutocorrectSuggestion> SigSuggestion::maybeSuggestSig(core::Conte
         }
         fmt::format_to(std::back_inserter(ss), ").");
     }
-    if (!guessedSomethingUseful) {
+    if (!guessedSomethingUseful && !ctx.state.suggestUnsafe.has_value()) {
+        // We don't want to condition people to start inserting a bunch of useless signatures filled
+        // with `T.untyped`, unless they've explicitly opted into the behavior with
+        // `--suggest-unsafe` (which usually suggests that they're doing some sort of codemod and
+        // they know what they're asking for).
```
ilyailya, line 481: Big +1 on this!

ilyailya (approved):

PR #6935, opened 2023-04-13
restructure `updateKnowledge` with early returns

froydnj:
### Motivation

I think the code is easier to understand this way, as it's clear that each separate `send->fun` is essentially its own case.  It also makes the checking for bang/nil/etc. epsilon faster, as we no longer fall through to the `args.empty()` check, which would be trivially true.

### Test plan

See included automated tests.


jez (approved):
One day I hope to power this by symbol-based intrinsics instead of names, but until then this seems fine 👍

PR #6936, opened 2023-04-13
add some more exception-handling test cases

froydnj:
### Motivation

This cherry-picks a testcase change from #6928.  It also adds a testcase around loops, inspired by some Stripe code.

I also added some splat testcases; Sorbet doesn't really handle these, thanks to untyped and various other things, but I think we *should* be able to if we rewrote how we represent `rescue`.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/cfg/rescue_complex.rb
+++ b/test/testdata/cfg/rescue_complex.rb
@@ -71,6 +71,7 @@ class TestRescue
     rescue *typed_exceptions => e
       T.reveal_type(e) # error: Revealed type: `T.untyped`
     end
+  end
 
   def parse_rescue_ensure()
     begin; meth; rescue; baz; ensure; bar; end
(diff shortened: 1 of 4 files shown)


test/testdata/cfg/rescue_complex.rb
```diff
@@ -30,6 +35,44 @@ def multiple_rescue_classes()
...
   end
 
+  def multiple_rescue_classes_varuse()
+    begin
+      meth
+    rescue LoadError, SocketError => baz
+      baz
+    end
+
+    T.reveal_type(baz) # error: Revealed type: `T.untyped`
```
froydnj, line 45: We want to be able to say `T.nilable(T.any(LoadError, SocketError))` here; #6928 gets us part of the way there, but to get further, we need to restructure exception handling a bit more.

jez (approved):

PR #6937, opened 2023-04-13
assign captured rescue var after we match on exception, not before

froydnj:
### Motivation

For Ruby like:

```ruby
begin
  func
rescue LoadError => ex
  print(ex)
end
```

We generate CFG for the `rescue` that looks something like (in Ruby):

```ruby
<exceptionValue> = <get-current-exception>
ex = <exceptionValue>
if LoadError.===(<exceptionValue>)
  print(ex)
end
```

This works for static analysis, but it doesn't model the Ruby faithfully; we shouldn't be assigning to `ex` until we know that the exception matches `LoadError`, i.e. we should be generating:

```ruby
<exceptionValue> = <get-current-exception>
if LoadError.===(<exceptionValue>)
  ex = <exceptionValue>
  print(ex)
end
```

which is what this PR does.

### Test plan

#6936 should land before this one; once this PR is rebased, https://github.com/sorbet/sorbet/pull/6936/files#r1165518085 will produce the result we want.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -830,8 +830,7 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
 
                     auto localVar = cctx.inWhat.enterLocal(local->localVariable);
 
-                    caseBody->exprs.emplace_back(localVar, rescueCase->var.loc(),
-                                                 make_insn<Ident>(exceptionValue));
+                    caseBody->exprs.emplace_back(localVar, rescueCase->var.loc(), make_insn<Ident>(exceptionValue));
                     // Mark the exception as handled
                     synthesizeExpr(caseBody, exceptionValue, core::LocOffsets::none(),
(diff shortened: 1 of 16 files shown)


froydnj:

→ https://go/builds/bui_NhqmTA7KXkD1UZ
→ https://go/builds/bui_NhqmVTGv07GabB
→ https://go/builds/bui_NhqmaPWyft6cc2

froydnj:
> 
> → https://go/builds/bui_NhqmTA7KXkD1UZ → https://go/builds/bui_NhqmVTGv07GabB → https://go/builds/bui_NhqmaPWyft6cc2

It's very sad, but it looks like having concrete types here + defaulting all variables to nil means that we run afoul of "Changing the type of a variable in a loop is not permitted".

We also hit problems with code like:

```ruby
begin
  func
rescue => e
  raise
ensure
  # log information about success/failure, including accesses to e
end
```

and our representation of exceptions is not strong enough to model this case (we don't model a world where you can run any code after the `raise`, so `e` is always `nil` in our model).

jez:
These changes landed in #7163

PR #6940, opened 2023-04-13
Add RBI sigs for `JSON.load_file` & `JSON.load_file!`

sambostock:
This adds RBI sigs for [`JSON.load_file`](https://docs.ruby-lang.org/en/master/JSON.html#method-i-load_file) & [`JSON.load_file!`](https://docs.ruby-lang.org/en/master/JSON.html#method-i-load_file-21), introduced in https://github.com/flori/json/pull/387 and [released in Ruby 3](https://github.com/ruby/ruby/commit/0aac138e0b507c8a361f7cf2f30c276e7110ea33).

### Motivation

Because the current RBI is missing them, even though they exist.

### Test plan

No tests for these RBIs AFAIK.

### DIFF
--- a/rbi/stdlib/json.rbi
+++ b/rbi/stdlib/json.rbi
@@ -235,6 +235,26 @@ module JSON
   end
   def self.load(source, proc=T.unsafe(nil), options=T.unsafe(nil)); end
 
+  # https://docs.ruby-lang.org/en/master/JSON.html#method-i-load_file
+  sig do
+    params(
+      path: ::T.untyped,
+      opts: ::T.untyped,
+    )
+    .returns(::T.untyped)
+  end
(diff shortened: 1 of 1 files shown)


sambostock:
How'd we do @froydnj?

rbi/stdlib/json.rbi
```diff
@@ -235,6 +235,26 @@ module JSON
   end
   def self.load(source, proc=T.unsafe(nil), options=T.unsafe(nil)); end
 
+  # https://docs.ruby-lang.org/en/master/JSON.html#method-i-load_file
+  sig do
+    params(
+      path: ::T.untyped,
```
froydnj, line 241: Is there really no better type that we can provide here?

froydnj (approved):
Looks great!  Thanks for the ping.

PR #6943, opened 2023-04-14
Don't suggest `T.untyped` by default

jez:
### Motivation

This is for a similar reason as provided in #6932

### Test plan

See included automated tests.


ilyailya (approved):
LGTM!

PR #6946, opened 2023-04-15
Skip autoload generation in autogen when only path-based-autoloading-compatible files have updated

aadi-stripe:
Skips autoload generation when only path-based-autoloading-compatible code has changed.

### Motivation

As part of migrating Stripe's codebase to path-based-autoloading with Zeitwerk, we are making autoload generation largely unnecessary. A first step towards this is improving efficiency and median runtime by skipping generation when only path-bsaed-autoloaded files are changed, since they do not need explicit autoload shim files to be built.

### Test plan

Tested on Stripe codebase. Messaged in Slack regarding timing analysis traces.

### DIFF
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -20,8 +20,10 @@
 #include "packager/rbi_gen.h"
 #endif
 
+#include "absl/algorithm/container.h"
 #include "absl/strings/match.h"
 #include "absl/strings/str_cat.h"
+#include "absl/strings/str_join.h"
 #include "absl/strings/str_split.h"
 #include "common/FileOps.h"
 #include "common/sort/sort.h"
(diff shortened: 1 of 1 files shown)


main/realmain.cc
```diff
@@ -200,8 +202,49 @@ struct AutogenResult {
...
+bool canSkipAutoloadGeneration(const core::GlobalState &gs, const string_view rootDir, vector<std::string> &files) {
+    // Return true if we can skip autoload generation for every file.
+    return absl::c_all_of(
+        files, [&gs, &rootDir](const auto &file) { return canSkipAutoloadGenerationForFile(gs, rootDir, file); });
+}
+
 void runAutogen(const core::GlobalState &gs, options::Options &opts, const autogen::AutoloaderConfig &autoloaderCfg,
-                const autogen::AutogenConfig &autogenCfg, WorkerPool &workers, vector<ast::ParsedFile> &indexed) {
+                const autogen::AutogenConfig &autogenCfg, WorkerPool &workers, vector<ast::ParsedFile> &indexed,
+                vector<std::string> &changedFiles) {
```
jez, line 247:
```suggestion
                const vector<std::string> &changedFiles) {
```

can this be const?

jez (approved):

main/realmain.cc
aadi-stripe: Sure

PR #6947, opened 2023-04-17
Update community.js with useful Sorbet gem links

maxveldink:
These three gems I've added either augment sorbet-runtime or make using Sorbet easier with Rails.

This is the best place to add these gems. The other links on this list appear to be related to tooling for Sorbet and less around plugins or extensions. These plugins and extensions will become more prevalent over time, so it might make sense to make a separate list. Happy to do either!

### Motivation
I've been working on several Sorbet plugins that originate from patterns we use in our Sorbet-typed codebases. I don't know how to share them and solicit feedback and issues; the community list is probably the best place. 

### Test plan
Docs build

### DIFF
--- a/website/pages/en/community.js
+++ b/website/pages/en/community.js
@@ -154,6 +154,24 @@ class Index extends React.Component {
                     'Measure your progress as you adopt Sorbet, stay motivated!',
                 },
                 {
+                  title: 'activerecord-ejection_seat',
+                  link: 'https://github.com/maxveldink/activerecord-ejection_seat',
+                  description:
+                    'Eject from an ActiveRecord model to a Sorbet T::Struct, or buckle back in',
+                },
+                {
+                  title: 'sorbet-result',
+                  link: 'https://github.com/maxveldink/sorbet-result',
+                  description:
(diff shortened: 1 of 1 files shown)


website/pages/en/community.js
```diff
@@ -153,6 +153,24 @@ class Index extends React.Component {
...
+                  title: 'sorbet-result',
+                  link: 'https://github.com/maxveldink/sorbet-result',
+                  description:
+                    'Introduces T::Result, T::Success and T::Failure types to facilitate Railway Oriented Programming',
+                },
+                {
+                  title: 'sorbet-struct-comparable',
+                  link: 'https://github.com/samuelgiles/sorbet-struct-comparable',
+                  description:
+                    'Comparable T::Struct's for the equality focused typed Ruby developer',
```
ilyailya, line 172: Looks like apostrophe messes up the `description` string, because it's also using single quotes
```suggestion
                    "Comparable T::Struct's for the equality focused typed Ruby developer",
```

ilyailya (changes requested):
Hey! 
Thanks for your suggestions! Looks like we've got a linting failure. I've added a comment about that

froydnj (approved):

Issue #6848, opened 2023-03-14
Remove `bind` from `T.proc` when in Initializer rewriter.


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20C%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams%28blk%3A%20T.proc.bind%28String%29.void%29.void%20%7D%0A%20%20def%20initialize%28%26blk%29%0A%20%20%20%20%40blk%20%3D%20blk%0A%20%20end%0Aend%0A)

```ruby
# typed: true

class C
  extend T::Sig

  sig {params(blk: T.proc.bind(String).void).void }
  def initialize(&blk)
    @blk = blk
  end
end
```

#### Observed output

```
editor.rb:6: Using bind is not permitted here https://srb.help/5004
     6 |  sig {params(blk: T.proc.bind(String).void).void }
                           ^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

When Sorbet copies the type annotation from a signature onto an instance variable declaration in `initialize`, it should strip the `.bind(...)` annotation from the block.

- - -

egiurleo:
@pl-at-shopify and I are working on this!

pl-at-shopify:
[...]

PR #6949, opened 2023-04-17
Remove `bind` from `T.proc` when in Initializer rewriter

pl-at-shopify:
This little patch simply removes `.bind(…)` from the chain of `Send`s when copying the block argument type definition to the `initialize` method body where `.bind(…)` is not allowed.

### Motivation

Fixes https://github.com/sorbet/sorbet/issues/6848.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Initializer.cc
+++ b/rewriter/Initializer.cc
@@ -29,31 +29,8 @@ bool isCopyableType(const ast::ExpressionPtr &typeExpr) {
     return true;
 }
 
-// copied from rewriter/Prop.cc
-bool isT(const ast::ExpressionPtr &expr) {
-    auto *t = ast::cast_tree<ast::UnresolvedConstantLit>(expr);
-    return t != nullptr && t->cnst == core::Names::Constants::T() && ast::MK::isRootScope(t->scope);
-}
-
-// partially copied from resolver/type_syntax/type_syntax.cc
-bool isTProc(core::Context ctx, const ast::Send *send) {
-    while (send != nullptr) {
-        if (send->fun == core::Names::proc()) {
-            if (isT(send->recv)) {
-                return true;
-            }
-        }
-        send = ast::cast_tree<ast::Send>(send->recv);
-    }
-    return false;
-}
-
 // Not checking for being T.proc, it's expected to be checked
 void maybeRemoveBind(core::Context ctx, ast::Send *send) {
-    if (!isTProc(ctx, send)) {
-        return;
-    }
-
     auto lastSend = send;
     while (send != nullptr) {
         if (send->fun == core::Names::bind()) {
--- a/rewriter/Initializer.cc
+++ b/rewriter/Initializer.cc
@@ -29,6 +29,53 @@ bool isCopyableType(const ast::ExpressionPtr &typeExpr) {
     return true;
 }
 
+// copied from rewriter/Prop.cc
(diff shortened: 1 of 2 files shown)


rewriter/Initializer.cc
```diff
@@ -29,6 +29,30 @@ bool isCopyableType(const ast::ExpressionPtr &typeExpr) {
...
+void maybeRemoveBind(core::Context ctx, ast::Send *send) {
+    auto lastSend = send;
+    while (send != nullptr) {
+        if (send->fun == core::Names::bind()) {
+            // `bind` is the last send in the chain of sends.
+            // This is incorrect syntax which is covered by the resolver.
+            // Safe to ignore.
+            if (send == lastSend) {
+                return;
+            }
```
jez, line 42: Do you mind writing a test for this case?

rewriter/Initializer.cc
```diff
@@ -29,6 +29,30 @@ bool isCopyableType(const ast::ExpressionPtr &typeExpr) {
...
+            auto recvSend = ast::cast_tree<ast::Send>(send->recv);
+            if (recvSend != nullptr) {
+                lastSend->recv = move(send->recv);
+                return;
+            }
+            // We handle only one `.bind` call in the whole chain,
+            // no need to traverse the rest of the tree.
+            return;
```
jez, line 50: And this as well?

jez (approved):
Looks good but I would appreciate two more tests.

jez:

→ https://go/builds/bui_NkWN0wTKEcTkq4
→ https://go/builds/bui_NkWNkY1hhH4kBp
→ https://go/builds/bui_NkWNhtrRCPSKqp

jez:
> Looks good but I would appreciate two more tests.

Going to merge for now. Feel free to open a new PR with these tests.

rewriter/Initializer.cc
pl-at-shopify: Got into a sort of trouble testing it. The test requires marking the same line with two `# error:` comments and I'm not sure how to deal with it. Here is the test code:

```ruby
class TProcBindInInitializerLastSend
  extend T::Sig

  sig {params(blk: T.proc.bind(String)).void }
    @blk = blk
  end
end

```

The line `sig {params(blk: T.proc.bind(String)).void }` produces two errors now:
- Malformed T.proc: You must specify a return type
- Using `bind` is not permitted here

One way to get around it is to change the code in the PR, but it looks so clean and bullet proof that I'm hesitating to :)

rewriter/Initializer.cc
pl-at-shopify: Same as above with these two errors competing for the same line of code:
- Malformed `bind`: Multiple calls to `.bind`
- Using `bind` is not permitted here

froydnj: We have docs on how to add multiple errors for a single line: https://github.com/sorbet/sorbet/#test_corpus-tests

pl-at-shopify:
Here the tests go: https://github.com/sorbet/sorbet/pull/6955
Thanks for trusting me on them coming 👍

rewriter/Initializer.cc
jez: +1 to drive home what nathan mentions:

```
T.proc.bind(A).bind(A).void
#      ^^^^ error: first error
#              ^^^^ error: second error
```

should let you make these assertions

PR #6955, opened 2023-04-21
Remove `bind` from `T.proc` follow-up with tests

pl-at-shopify:
### Motivation

A follow-up PR for https://github.com/sorbet/sorbet/pull/6949

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/rewriter/initializer.rb
+++ b/test/testdata/rewriter/initializer.rb
@@ -197,8 +197,9 @@ class TProcBindInInitializerLastSend
   extend T::Sig
 
   sig {params(blk: T.proc.bind(String)).void }
-  # error: Using `bind` is not permitted here
-  # error-with-dupes: Malformed T.proc: You must specify a return type
+  #                ^^^^^^^^^^^^^^^^^^^ error: Malformed T.proc: You must specify a return type
+  #                ^^^^^^^^^^^^^^^^^^^ error: Malformed T.proc: You must specify a return type
+  #                ^^^^^^^^^^^^^^^^^^^ error: Using `bind` is not permitted here
   def initialize(&blk)
     @blk = blk
   end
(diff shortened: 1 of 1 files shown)


test/testdata/rewriter/initializer.rb
```diff
@@ -192,3 +192,25 @@ def initialize(&blk)
     @blk = blk
   end
 end
+
+class TProcBindInInitializerLastSend
+  extend T::Sig
+
+  sig {params(blk: T.proc.bind(String)).void }
+  # error: Using `bind` is not permitted here
+  # error-with-dupes: Malformed T.proc: You must specify a return type
```
jez, line 201:
```suggestion
  #                ^^^^^^^^^^^^^^^^^^^ error: Malformed T.proc: You must specify a return type
  #                ^^^^^^^^^^^^^^^^^^^ error: Malformed T.proc: You must specify a return type
  #                ^^^^^^^^^^^^^^^^^^^ error: Using `bind` is not permitted here
```

test/testdata/rewriter/initializer.rb
```diff
@@ -192,3 +192,25 @@ def initialize(&blk)
...
+    @blk = blk
+  end
+end
+
+class TProcBindInInitializerManyBinds
+  extend T::Sig
+
+  sig {params(blk: T.proc.bind(String).bind(String).void).void }
+  # error: Using `bind` is not permitted here
+  # error: Malformed `bind`: Multiple calls to `.bind`
```
jez, line 212:
```suggestion
  #                ^^^^^^^^^^^^^^^^^^^ error: Malformed `bind`: Multiple calls to `.bind`
  #                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Using `bind` is not permitted here
```

jez (changes requested):
Let's write these tests a little differently.

PR #6957, opened 2023-04-25
Use ctx.file instead

jez:
### Motivation

I want to factor this functionality out to a helper method, so that other people
can use it. It's easier to do that if it's possible to use `ctx.file` here.

### Test plan

Existing tests


neilparikh (approved):

jez:
Landed in #6958

PR #6958, opened 2023-04-26
Add editForDSLMethod helper

jez:
### Motivation

There's a bunch of autocorrects I want to build of the form "insert some DSL
method into the class"

- add `abstract!`
- add `X = type_member`

etc.

It's convenient if we don't have to keep duplicating this logic into every
specific autocorrect where we want this.

### Caveats

An extended note about the root logic: this is kind of a tradeoff.

For the SigSuggestion autocorrect, this has the unfortunate side effect of
inserting `extend T::Sig` multiple times in the file. But IMO, that's kind
of find, because that would only apply if there were multiple methods
without a sig in the file, and either

- You used "apply all fixes for file", or
- You used `-a` at the command line

There will be many cases where neither of those is true, which means that
the new behavior is a strict improvement.

In the cases where it does add duplicate, the redundant calls to
`extend T::Sig` are a no-op, so it's not _incorrect_ to have them, it's
just maybe bad style. I think I'd prefer having something that is really
good at fixing type errors.

It's hard to do better here. If it were possible to always insert the
`extend T::Sig` somewhere like "at the top of the file, but after all
requires and comments," then all the autocorrect edits would share the same
location, causing only one of them to get applied (we only apply the first
edit that happens at a given place).

The real goal is to get this in so that other call sites that _are_ fine
with this logic can use this one new helper.

### Test plan

Existing tests

### DIFF
--- a/core/TypeErrorDiagnostics.cc
+++ b/core/TypeErrorDiagnostics.cc
@@ -136,6 +136,7 @@ void TypeErrorDiagnostics::explainUntyped(const GlobalState &gs, ErrorBuilder &e
 }
 
 // dslOwner can be noClassOrModule() to simply unconditionally insert the `dsl` string
+// dsl can be `""` to simply insert `extend {dslOwner}` if dslOwner is not already an ancestor
 optional<core::AutocorrectSuggestion::Edit> TypeErrorDiagnostics::editForDSLMethod(const Context &ctx,
                                                                                    ClassOrModuleRef inWhatRef,
                                                                                    ClassOrModuleRef dslOwner,
@@ -178,10 +179,12 @@ optional<core::AutocorrectSuggestion::Edit> TypeErrorDiagnostics::editForDSLMeth
     if (needsDslOwner) {
         return core::AutocorrectSuggestion::Edit{
             nextLineLoc.value(),
-            fmt::format("{}extend {}\n{}{}\n", prefix, dslOwner.show(ctx), prefix, dsl),
+            fmt::format("{}extend {}\n{}{}\n", prefix, dslOwner.show(ctx), dsl != "" ? prefix : "", dsl),
         };
-    } else {
+    } else if (dsl != "") {
         return core::AutocorrectSuggestion::Edit{nextLineLoc.value(), fmt::format("{}{}\n", prefix, dsl)};
+    } else {
+        return nullopt;
     }
 }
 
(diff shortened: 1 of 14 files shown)


jez:
* master
  * **PR #6958** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6958" target="_blank">Graphite</a>  👈
    * **PR #6959** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6959" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/6958?utm_source=stack-comment).

core/TypeErrorDiagnostics.cc
```diff
@@ -134,4 +135,99 @@ void TypeErrorDiagnostics::explainUntyped(const GlobalState &gs, ErrorBuilder &e
...
+                                       ClassOrModuleRef inWhatRef, ClassOrModuleRef dslOwner, string_view dsl) {
+    auto inWhat = inWhatRef.data(gs);
+    auto inWhatSingleton = inWhat->lookupSingletonClass(gs);
+
+    auto needsDslOwner = false;
+    if (dslOwner.exists()) {
+        needsDslOwner = !inWhatSingleton.data(gs)->derivesFrom(gs, dslOwner);
+    }
+
+    auto inCurrentFile = [&](const auto &loc) { return loc.file() == fileToEdit; };
```
ilyailya, line 178: [nit, feel free to ignore]
Would it be possible to capture only `fileToEdit` in  `inCurrentFile`?
```suggestion
    auto inCurrentFile = [&fileToEdit](const auto &loc) { return loc.file() == fileToEdit; };
```

ilyailya (approved):
The change looks good to me and the trade off is very reasonable. I left one nit comment, it's non-blocking, feel free to ignore

core/TypeErrorDiagnostics.cc
jez: If the lambda were longer, I would definitely agree with you. As is, I personally feel like the lambda is short enough to warrant the `[&]` capture. In some other places we use explicit capturing where the lambda is more complicated, we prefer explicit captures.

(Given how short it is, I feel like it would just be one more thing to forget to change if this code were to be edited. As in, you'd rename the variable, or change the `==` condition, but forget to change the list of captures and get a toilsome compiler error.)

PR #6965, opened 2023-04-28
Descend further to find include and extend

jez:
### Motivation

There are a bunch of `include` and `extend` lines that sit behind an `if`
condition at a class top level.

Right now, those are just getting written out into RBI files, so it shouldn't
cause any new type check errors because of this, unless the value of the `if`
condition matters.

It looks like autogen had a test that explicitly tested that we _didn't_ find
the `include` in side block calls. So maybe that means this is intentional, and
we'll have to find another way to solve the RBI problem. Only one way to find
out.

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -85,6 +85,9 @@ class SymbolFinder {
     // The tree doesn't have symbols yet, so `ctx.owner`, which is a SymbolRef, is meaningless.
     // Instead, we track the owner manually via FoundDefinitionRefs.
     vector<core::FoundDefinitionRef> ownerStack;
+    // We want to look through all code not inside methods to find ancestors.
+    // To do that, we need access to classDef.ancestors inside a postTransformSend
+    vector<ast::ClassDef *> enclosingClassDefs;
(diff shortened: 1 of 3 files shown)


jez:

→ https://go/builds/bui_NnKqqFSVUDTFlI
→ https://go/builds/bui_NnKq39iQbvmAUX
→ https://go/builds/bui_NnKqcLKo3VJ2pk

froydnj (changes requested):
When you say these are getting written into RBI files, do you have a test case that shows what's happening there?  According to the autogen changes here, we should be writing more stuff to RBI files, but the PR description sounds like this change should make us write less stuff, and I don't understand how that would work.

I think this would also be sort of weird insofar as if you had:

```ruby
if condition
  include A
else
  include B
end
```

then after this PR, Sorbet would think that the class inherits from both `A` and `B`, which I could imagine leading to some weird errors (though I'm having trouble constructing a testcase, because I think the current state of the world implies that people would have had to use `T.unsafe` somewhere in their client code to make this work...).

PR #6967, opened 2023-04-30
Only handle sensitivity annotations in prop DSL when necessary

aadi-stripe:
### Motivation

The `prop` DSL always calls into the `normalize_sensitivity_and_pii_handler`, even when the `sensitivity` rule isn't declared. In those cases, the call is unnecessary and allocates memory. Fixing this issue reduces the object allocations of loading all model classes at Stripe by **0.3%**.

### Test plan

Tested on Stripe codebase.
https://go/builds/bui_No8ZOgvDw9hKgn

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -338,20 +338,23 @@ class T::Props::Decorator
     # Retrive the possible underlying object with T.nilable.
     type = T::Utils::Nilable.get_underlying_type(type)
 
-    sensitivity_and_pii = {sensitivity: rules[:sensitivity]}
-    normalize = T::Configuration.normalize_sensitivity_and_pii_handler
-    if normalize
-      sensitivity_and_pii = normalize.call(sensitivity_and_pii)
+    rules_sensitivity = rules[:sensitivity]
+    sensitivity_and_pii = {sensitivity: rules_sensitivity}
+    if !rules_sensitivity.nil?
+      normalize = T::Configuration.normalize_sensitivity_and_pii_handler
+      if normalize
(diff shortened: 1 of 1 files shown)


gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -339,19 +339,21 @@ def prop_defined(name, cls, rules={})
...
-      # We check for Class so this is only applied on concrete
-      # documents/models; We allow mixins containing props to not
-      # specify their PII nature, as long as every class into which they
-      # are ultimately included does.
-      #
-      if sensitivity_and_pii[:pii] && @class.is_a?(Class) && !T.unsafe(@class).contains_pii?
-        raise ArgumentError.new(
-          'Cannot include a pii prop in a class that declares `contains_no_pii`'
-        )
+    if !rules[:sensitivity].nil?
```
jez, line 342: a micro optimization:

```ruby
    rules_sensitivity = rules[:sensitivity]
    sensitivity_and_pii = {sensitivity: rules_sensitivity}
    if rules_sensitivity
```

jez:
We have some benchmarks in `gems/sorbet-runtime/bench`. They're meant to be run by hand. Do you want to scan what's there to see if any of the existing benchmarks cover this case, and if not do you want to write a new one?

jez (approved):
lgtm. we might want to write a benchmark for this, but i'm find with doing that in a follow up PR.

aadi-stripe:
> We have some benchmarks in `gems/sorbet-runtime/bench`. They're meant to be run by hand. Do you want to scan what's there to see if any of the existing benchmarks cover this case, and if not do you want to write a new one?

Will write one in a follow-up PR.

PR #6968, opened 2023-04-30
Migrate a prop type validation helper from kwarg -> arg so that it can be sig-checked faster

aadi-stripe:
### Motivation

Every `prop` declaration is validated at runtime for its type declarations. This is a critical path in class loading at Stripe. There is a `validate_type` method that is called during prop declarations, which uses kwargs. Unfortunately, kwarg validation is slower in Sorbet runtime. Converting arity to simple args so that the method can take the "fast path".

Fixing this issue reduces the object allocations of loading all model classes at Stripe by **0.7%**. 

### Test plan

Tested on Stripe codebase.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/type_validation.rb
+++ b/gems/sorbet-runtime/lib/types/props/type_validation.rb
@@ -29,7 +29,7 @@ module T::Props::TypeValidation
       super
 
       if !rules[:DEPRECATED_underspecified_type]
-        validate_type(type, field_name: name)
+        validate_type(type, name)
       elsif rules[:DEPRECATED_underspecified_type] && find_invalid_subtype(type).nil?
         raise ArgumentError.new("DEPRECATED_underspecified_type set unnecessarily for #{@class.name}.#{name} - #{type} is a valid type")
       end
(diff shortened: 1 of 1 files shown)


gems/sorbet-runtime/lib/types/props/type_validation.rb
```diff
@@ -42,7 +42,7 @@ def prop_validate_definition!(name, _cls, rules, type)
       )
       .void
     end
-    private def validate_type(type, field_name:)
+    private def validate_type(type, field_name)
```
jez, line 45: If we're saying that type checking this method is hot, should we mark the sig `.checked(:never)`?

aadi-stripe: Yes, we can. I guess I wasn't sure if we wanted to lose whatever guarantees we get here from the checking, but the likelihood is that a sig is checked somewhere well up the callstack, so it's ok to make this checked-never.

jez (approved):
ditto my comment on #6967 about potentially wanting to run or write a benchmark for this. i also think we might want to re-test with `.checked(:never)` if we think that this method's performance matters.

aadi-stripe:
Added a benchmark, and saw a 4.5% drop in allocations in the benchmark:

Before: 6565441
After: 6265441

PR #6969, opened 2023-04-30
Remove unnecessary allocations when checking splat arguments

aadi-stripe:
Drops **1%** of allocations when loading all model classes at Stripe. Will potentially have a further impact for all method calls with splats, beyond just class loading.

### Motivation

Improve performance of runtime typechecking.

### Test plan

Build: https://go/builds/bui_NoJwP5CetUPaMi

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/signature.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/signature.rb
@@ -179,15 +179,7 @@ class T::Private::Methods::Signature
       kwargs = EMPTY_HASH
     end
 
-    arg_types = @arg_types
-
-    if @has_rest
-      rest_count = args_length - @arg_types.length
-      rest_count = 0 if rest_count.negative?
-
-      arg_types += [[@rest_name, @rest_type]] * rest_count
-
-    elsif (args_length < @req_arg_count) || (args_length > @arg_types.length)
+    if !@has_rest && ((args_length < @req_arg_count) || (args_length > @arg_types.length))
       expected_str = @req_arg_count.to_s
       if @arg_types.length != @req_arg_count
         expected_str += "..#{@arg_types.length}"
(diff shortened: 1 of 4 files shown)


gems/sorbet-runtime/lib/types/private/methods/signature.rb
```diff
@@ -197,17 +189,31 @@ def each_args_value_type(args)
 
     begin
       it = 0
-      while it < args_length
-        yield arg_types[it][0], args[it], arg_types[it][1]
+
+      # Process given pre-rest args. When there are no rest args,
+      # this is just the given number of args.
+      while it < [args_length, @arg_types.length].min
```
jez, line 195: You might want to avoid using an array literal here—pretty sure that's going to be another allocation.

If `@arg_types.length` never changes, maybe something like this:

```ruby
while it < args_length && it < arg_types_length
  # ...
end
```

jez:
ooc, did you get a sense for how good our test coverage is for this code? if you had messed it up, would our tests have failed? can you try intentionally breaking this code and see that a test fails, and if none do, writing a test that would have caught the problem?

aadi-stripe:
> ooc, did you get a sense for how good our test coverage is for this code? if you had messed it up, would our tests have failed? can you try intentionally breaking this code and see that a test fails, and if none do, writing a test that would have caught the problem?

I was planning to check and backfill a test if we don’t have one.

qaisjp:
> Drops 1% of allocations when loading all model classes at Stripe

Out of curiosity, how did you measure this?

aadi-stripe:
> > ooc, did you get a sense for how good our test coverage is for this code? if you had messed it up, would our tests have failed? can you try intentionally breaking this code and see that a test fails, and if none do, writing a test that would have caught the problem?
> 
> I was planning to check and backfill a test if we don’t have one.

We have a test for splats that covers issues.

aadi-stripe:
Added benchmark for kwargs and splats.

40% improvement in allocations and 12% improvement in runtime!!

BEFORE
```
bundle exec rake bench:typecheck_kwargs_splat
sig {params(s: Symbol, rest: String, x: Integer, y: Integer).void} (with kwargs plus splat): 3.172 μs
Allocations for sig {params(s: Symbol, rest: String, x: Integer, y: Integer).void} (with kwargs plus splat): 20000000
```

AFTER

```
bundle exec rake bench:typecheck_kwargs_splat
sig {params(s: Symbol, rest: String, x: Integer, y: Integer).void} (with kwargs plus splat): 2.777 μs
Allocations for sig {params(s: Symbol, rest: String, x: Integer, y: Integer).void} (with kwargs plus splat): 12000000
```

aadi-stripe:
> > Drops 1% of allocations when loading all model classes at Stripe
> 
> Out of curiosity, how did you measure this?

Replied in Slack.

jez (approved):

PR #6972, opened 2023-05-01
Mark `::Class` in the standard library `final!`

jez:
### Motivation

This matches the runtime behavior.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -348,7 +348,7 @@ class T::Props::Decorator
       # specify their PII nature, as long as every class into which they
       # are ultimately included does.
       #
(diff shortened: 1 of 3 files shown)


jez:
This is not actually true—all the singleton classes descend from `Class`

```
❯ irb
irb(main):001:0> class A; end
=> nil
irb(main):002:0> A.ancestors
=> [A, Object, Kernel, BasicObject]
irb(main):003:0> A.singleton_class.ancestors
=> [#<Class:A>, #<Class:Object>, #<Class:BasicObject>, Class, Module, Object, Kernel, BasicObject]
```

Sorbet also sees these in any class which uses `class << self`, because `T.class_of(T.class_of(T.class_of(A)))` doesn't always get created, so it gets set to `Class`, and then Sorbet sees a ClassDef that has `::Class` as its super class.

Anyways, even if we were to write some sort of condition that said "skip this error even for singleton classes" it would still be a lie—all the optimizations and type system features we would want to build depend on `final!` not being a lie.

PR #6976, opened 2023-05-03
wip - Class_eval for props

aadi-stripe:
### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -393,28 +393,41 @@ class T::Props::Decorator
     handle_redaction_option(name, rules[:redaction]) if rules[:redaction]
   end
 
-  # checked(:never) - Rules hash is expensive to check
+    # checked(:never) - Rules hash is expensive to check
   sig {params(name: Symbol, rules: Rules).void.checked(:never)}
   private def define_getter_and_setter(name, rules)
     T::Configuration.without_ruby_warnings do
       if !rules[:immutable]
         if method(:prop_set).owner != T::Props::Decorator
-          @class.send(:define_method, "#{name}=") do |val|
-            T.unsafe(self.class).decorator.prop_set(self, name, val, rules)
-          end
+          @class.class_eval <<~SETTER
+            def #{name}=(val)
+              rules = T.unsafe(self.class).decorator.prop_rules(:#{name})
+              T.unsafe(self.class).decorator.prop_set(self, :#{name}, val, rules)
+            end
+          SETTER
         else
           # Fast path (~4x faster as of Ruby 2.6)
-          @class.send(:define_method, "#{name}=", &rules.fetch(:setter_proc))
+          @class.class_eval <<~SETTER
+            def #{name}=(val)
+              rules = T.unsafe(self.class).decorator.prop_rules(:#{name})
+              instance_exec(val, &rules.fetch(:setter_proc))
+            end
+          SETTER
         end
       end
 
       if method(:prop_get).owner != T::Props::Decorator || rules.key?(:ifunset)
-        @class.send(:define_method, name) do
-          T.unsafe(self.class).decorator.prop_get(self, name, rules)
-        end
+        @class.class_eval <<~GETTER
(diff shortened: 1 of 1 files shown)


gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -393,28 +393,41 @@ def prop_defined(name, cls, rules={})
...
     T::Configuration.without_ruby_warnings do
       if !rules[:immutable]
         if method(:prop_set).owner != T::Props::Decorator
-          @class.send(:define_method, "#{name}=") do |val|
-            T.unsafe(self.class).decorator.prop_set(self, name, val, rules)
-          end
+          @class.class_eval <<~SETTER
+            def #{name}=(val)
+              rules = T.unsafe(self.class).decorator.prop_rules(:#{name})
+              T.unsafe(self.class).decorator.prop_set(self, :#{name}, val, rules)
```
jez, line 405: pretty sure these `T.unsafe` were to get the `decorator.rb` file itself to typehceck. now that you're not needing to type check it, you can remove the `T.unsafe`.

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -393,28 +393,41 @@ def prop_defined(name, cls, rules={})
     handle_redaction_option(name, rules[:redaction]) if rules[:redaction]
   end
 
-  # checked(:never) - Rules hash is expensive to check
+    # checked(:never) - Rules hash is expensive to check
```
jez, line 396:
```suggestion
  # checked(:never) - Rules hash is expensive to check
```

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -393,28 +393,41 @@ def prop_defined(name, cls, rules={})
...
   sig {params(name: Symbol, rules: Rules).void.checked(:never)}
   private def define_getter_and_setter(name, rules)
     T::Configuration.without_ruby_warnings do
       if !rules[:immutable]
         if method(:prop_set).owner != T::Props::Decorator
-          @class.send(:define_method, "#{name}=") do |val|
-            T.unsafe(self.class).decorator.prop_set(self, name, val, rules)
-          end
+          @class.class_eval <<~SETTER
+            def #{name}=(val)
```
jez, line 403: We'll have to do additional validation here that the result is valid Ruby code.

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -399,22 +399,35 @@ def prop_defined(name, cls, rules={})
...
+            rules = self.class.decorator.prop_rules(:#{name})
+            self.class.decorator.prop_get(self, :#{name}, rules)
+          end
+        GETTER
       else
         # Fast path (~30x faster as of Ruby 2.6)
-        @class.send(:attr_reader, name) # send is used because `attr_reader` is private in 2.4
+        @class.class_eval <<~GETTER
+          send(:attr_reader, :#{name})
+        GETTER
```
froydnj, line 430: I understand that this is consistent, but it also seems like unnecessary overhead, since we don't need the `class_eval` to run `attr_reader`.

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -399,22 +399,35 @@ def prop_defined(name, cls, rules={})
     T::Configuration.without_ruby_warnings do
       if !rules[:immutable]
         if method(:prop_set).owner != T::Props::Decorator
-          @class.send(:define_method, "#{name}=") do |val|
-            T.unsafe(self.class).decorator.prop_set(self, name, val, rules)
-          end
+          @class.class_eval <<~SETTER
+            def #{name}=(val)
+              rules = self.class.decorator.prop_rules(:#{name})
+              self.class.decorator.prop_set(self, :#{name}, val, rules)
```
froydnj, line 405: Since we have to fetch the decorator, can we at least factor out `self.class.decorator` into a local?  Here and elsewhere in this patch.

Issue #3888, opened 2021-01-14
`alias_method` fails to alias methods included by modules


paracycle:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Amodule%20Foo%0A%20%20def%20foo%0A%20%20end%0Aend%0A%0Aclass%20Bar%0A%20%20include%20Foo%0A%0A%20%20alias_method%20%3Abar%2C%20%3Afoo%0Aend)

```ruby
# typed: true

module Foo
  def foo
  end
end

class Bar
  include Foo

  alias_method :bar, :foo
end
```

#### Observed output

```
editor.rb:11: Can't make method alias from bar to non existing method foo https://srb.help/5037
    11 |  alias_method :bar, :foo
                             ^^^^
Errors: 1
```

#### Expected behavior

There should be no errors about non-existing method `foo` since it is defined by `Foo` in `Bar`.

- - -

jez:
https://github.com/sorbet/sorbet/issues/2378

paracycle:
🤦 thanks, makes sense.

paracycle:
Actually @jez, is there a special reason why `alias_method` is not implemented as a Rewriter. If it was implemented as a rewriter that just did delegation, i.e. converted:
```ruby
alias_method :bar, :foo
```
to
```ruby
def bar(*args, **kwargs, &blk)
  foo(*args, **kwargs, &blk)
end
```
then this would work properly without introducing order dependency.

Issue #2378, opened 2019-12-20
`alias` and `alias_method` unable to alias method from parent class


manhhung741:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20TSortableHash%20%3C%20Hash%0A%20%20include%20TSort%0A%20%20alias_method%20%3Atsort_each_node%2C%20%3Aeach_key%0A%20%20def%20tsort_each_child(node%2C%20%26block)%0A%20%20%20%20fetch(node).each(%26block)%0A%20%20end%0Aend%0A%0Aclass%20TSortableHash_2%20%3C%20Hash%0A%20%20include%20TSort%0A%20%20alias%20tsort_each_node%20each_key%0A%20%20def%20tsort_each_child(node%2C%20%26block)%0A%20%20%20%20fetch(node).each(%26block)%0A%20%20end%0Aend%0A%0Aclass%20Hash%0A%20%20include%20TSort%0A%20%20alias%20tsort_each_node%20each_key%0A%20%20def%20tsort_each_child(node%2C%20%26block)%0A%20%20%20%20fetch(node).each(%26block)%0A%20%20end%0Aend%0A)

```ruby
# typed: true

class TSortableHash < Hash
  include TSort
  alias_method :tsort_each_node, :each_key
  def tsort_each_child(node, &block)
    fetch(node).each(&block)
  end
end

class TSortableHash_2 < Hash
  include TSort
  alias tsort_each_node each_key
  def tsort_each_child(node, &block)
    fetch(node).each(&block)
  end
end

class Hash
  include TSort
  alias tsort_each_node each_key
  def tsort_each_child(node, &block)
    fetch(node).each(&block)
  end
end
```

#### Observed output

```
editor.rb:5: Can't make method alias from tsort_each_node to non existing method each_key https://srb.help/5037
     5 |  alias_method :tsort_each_node, :each_key
                                         ^^^^^^^^^

editor.rb:13: Can't make method alias from tsort_each_node to non existing method each_key https://srb.help/5037
    13 |  alias tsort_each_node each_key
                                ^^^^^^^^
Errors: 2
```

#### Expected behavior

`alias` and `alias_method` shouldn't complain when I try to alias a method defined in the parent class (each_key is defined in Hash)
- - -

jez:
This is intentional. See this previous closed PR where I implemented it for more context:

https://github.com/sorbet/sorbet/pull/1443

[...]

Issue #365, opened 2019-04-30
alias_method can't create alias from an included method


XrXr:
https://sorbet.run/#%23%20typed%3A%20true%0A%0Amodule%20SomeModuleExtension%0A%20%20def%20extension_method%0A%20%20end%0Aend%0A%0Aclass%20SomeExtendedClass%0A%20%20include%20SomeModuleExtension%0A%20%20alias_method%20%3Aaliased_method%2C%20%3Aextension_method%0Aend

Here, `extension_method` exists on `SomeExtendedClass` but `alias_method` complains that it does not exist.

Found by @dylanahsmith

PR #6977, opened 2023-05-03
Alias method

pl-at-shopify:
### Motivation

Cover more use cases of `alias_method`.

Sorbet currently doesn't support aliasing methods from superclasses and includes:

* https://github.com/sorbet/sorbet/issues/365
* https://github.com/sorbet/sorbet/issues/2378
* https://github.com/sorbet/sorbet/issues/3888

### Test plan

See included automated tests.


peter-leonov:
I'm not 100% sure that this approach covers all the use cases, but it looks like since the issue had been discovered the codebase has evolved which enables such a simple patch to enable at least the listed use cases almost for free. Or am I missing an elephant in the room?

PR #6979, opened 2023-05-04
More disabling sig-checking in props, and other fixes

aadi-stripe:
cc @georg-stripe @jez 

This is a follow up to my earlier PRs
https://github.com/sorbet/sorbet/pull/6967
https://github.com/sorbet/sorbet/pull/6968
https://github.com/sorbet/sorbet/pull/6969

Further improve the performance of `prop` declaration by dropping unnecessary sig checking on inner helper methods. Also fix an inefficient use of `merge`.

From a safety standpoint, the `sig` checks on inner helper methods were not adding further type safety than what is already available via the wrapper `prop` method up the stack. The `rules` hash is expensive to sig-check, and we've already turned off sig-checking for it in one place, and this ports over that change to other places.

Drops number of allocations on loading all model classes at Stripe by up to 2%.

### Motivation

Improve code loading performance.

### Test plan

Existing tests + tested on Stripe codebase.
https://go/builds/bui_NpnHlBX9OIZN69

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -79,7 +79,7 @@ class T::Props::Decorator
     extra
     setter_validate
     _tnilable
-  ].map {|k| [k, true]}.to_h.freeze, T::Hash[Symbol, T::Boolean], checked: false)
+  ].to_h {|k| [k, true]}.freeze, T::Hash[Symbol, T::Boolean], checked: false)
   private_constant :VALID_RULE_KEYS
 
   sig {params(key: Symbol).returns(T::Boolean).checked(:never)}
@@ -358,16 +358,12 @@ class T::Props::Decorator
       end
     end
 
-    rules.merge!(
-      # TODO: The type of this element is confusing. We should refactor so that
-      # it can be always `type_object` (a PropType) or always `cls` (a Module)
-      type: type,
-      type_object: type_object,
-      accessor_key: "@#{name}".to_sym,
-      sensitivity: sensitivity_and_pii[:sensitivity],
-      pii: sensitivity_and_pii[:pii],
-      extra: rules[:extra]&.freeze,
-    )
+    rules[:type] = type
+    rules[:type_object] = type_object
+    rules[:accessor_key] = "@#{name}".to_sym
+    rules[:sensitivity] = sensitivity_and_pii[:sensitivity]
+    rules[:pii] = sensitivity_and_pii[:pii]
+    rules[:extra] = rules[:extra]&.freeze
 
     # extra arbitrary metadata attached by the code defining this property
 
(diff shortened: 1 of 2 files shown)


aadi-stripe:
10% allocation drop on prop validation benchmarks:
BEFORE
```
Subclassing T::Struct, with ten props (ms/iter):
  5.517611   0.428318   5.945929 (  5.946177)
Allocations: 6265441
```

AFTER
```
Subclassing T::Struct, with ten props (ms/iter):
  5.288651   0.412172   5.700823 (  5.700965)
Allocations: 5765441
```

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -527,7 +531,7 @@ def prop_defined(name, cls, rules={})
     # *haven't* allowed additional options in the past and want to
     # default to keeping this interface narrow.
     @class.send(:define_method, fk_method) do |allow_direct_mutation: nil|
-      foreign = T.let(foreign, T.untyped)
+      foreign = T.let(foreign, T.untyped, checked: false)
```
froydnj, line 534: I think if you move this whole statement outside of the block, you will get the same benefit of type pinning in Sorbet, but you won't be making the `T.let` call on every call to `fk_method`.

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -247,18 +247,18 @@ def prop_validate_definition!(name, cls, rules, type)
     nil
   end
 
-  SAFE_NAME = T.let(/\A[A-Za-z_][A-Za-z0-9_-]*\z/.freeze, Regexp)
+  SAFE_NAME = T.let(/\A[A-Za-z_][A-Za-z0-9_-]*\z/.freeze, Regexp, checked: false)
```
froydnj, line 250: Just to clarify, we are `checked: false`'ing all of these so that we don't pay the slight cost of typechecking the top-level constants at code-loading time?  This seems...overkill to me, but I'm not exercised about it to block this PR.

froydnj (approved):

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -358,18 +358,19 @@ def prop_defined(name, cls, rules={})
...
       # TODO: The type of this element is confusing. We should refactor so that
       # it can be always `type_object` (a PropType) or always `cls` (a Module)
       type: type,
       type_object: type_object,
       accessor_key: "@#{name}".to_sym,
       sensitivity: sensitivity_and_pii[:sensitivity],
       pii: sensitivity_and_pii[:pii],
-      # extra arbitrary metadata attached by the code defining this property
       extra: rules[:extra]&.freeze,
     )
```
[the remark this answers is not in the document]
aadi-stripe: Yeah fair

PR #6980, opened 2023-05-05
Add the class name to prop definition error

neilparikh:
### Motivation

The previous error message only mentioned the prop, and not the class name, so for commonly used prop names, it would be difficult to find the offending code (especially if the stack trace was not available).

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -50,9 +50,9 @@ class T::Props::Decorator
     override = rules.delete(:override)
 
     if props.include?(prop) && !override
-      raise ArgumentError.new("Attempted to redefine prop #{prop.inspect} on class #{@class.inspect} that's already defined without specifying :override => true: #{prop_rules(prop)}")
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -50,9 +50,9 @@ def add_prop_definition(prop, rules)
     override = rules.delete(:override)
 
     if props.include?(prop) && !override
-      raise ArgumentError.new("Attempted to redefine prop #{prop.inspect} that's already defined without specifying :override => true: #{prop_rules(prop)}")
+      raise ArgumentError.new("Attempted to redefine prop #{prop.inspect} on class #{@class.inspect} that's already defined without specifying :override => true: #{prop_rules(prop)}")
```
jez, line 53: Typically for classes you can get away with `#{@class}`, no `inspect`, because the `to_s` method on a class object just calls `#{@class.name}`. We see `#{prop.inspect}` here because `#{prop}` when `prop` is a Symbol will call `prop.to_s`, which generates `"foo"` instead of `":foo"`.

tl;dr this is a minor nit but you don't need the `.inspect`, and I wanted to share why.

jez (approved):

Issue #1483, opened 2019-08-02
Unexpected union (T.any) behavior?


panthomakos:
#### Input

[→ View on sorbet.run](https://sorbet.run/#extend%20T%3A%3ASig%0A%0ANumber%20%3D%20T.type_alias(T.any(Integer%2C%20Float))%0A%0Asig%20%7B%20params(a%3A%20Integer%2C%20b%3A%20Number).returns(Integer)%20%7D%0Adef%20multiply(a%2C%20b)%0A%20%20(a%20*%20b).round%0Aend)

```ruby
extend T::Sig

Number = T.type_alias(T.any(Integer, Float))

sig { params(a: Integer, b: Number).returns(Integer) }
def multiply(a, b)
  (a * b).round
end
```

#### Observed output

```
editor.rb:7: Expected Integer but found T.any(Integer, Float) for argument arg0 https://srb.help/7002
     7 |  (a * b).round
           ^^^^^
    https://github.com/sorbet/sorbet/tree/0c6bdf75780d2f5dd627c7054ce09c67a938304f/rbi/core/integer.rbi#L40: Method Integer#* has specified arg0 as Integer
    40 |        arg0: Integer,
                ^^^^
  Got T.any(Integer, Float) originating from:
    editor.rb:6:
     6 |def multiply(a, b)
                        ^
Errors: 1
```

#### Expected behavior

This is a simplified example, but variations on this theme all produce similar errors (for example, replace `a: Integer` with `a: Number` in the type signature). I would expect this type signature to work. `Integer` supports multiplication with both `Integer` and `Float`.

DarkDimius:
This is a known limitation of our current overload selection algorithm.
It's unlikely that we'll fix it soon, as the complexity of supporting it is tremendous. 
The workaround is either `(a * b.to_f).round` or `(a * b.to_i).round`.

Note that we also consider the current behavior as beneficial, as otherwise figuring out what happens from user perspective with more complex overload resolution is quite hard.

DarkDimius:
As we don't currently intend to fix this limitation, closing to not pile up issues.

PR #6982, opened 2023-05-06
Fix signature for Kernel.raise

iMacTia:
Related: #1483 

### Motivation
~`Kernel.raise` second parameter is always optional, as described in the documentation:~
~The optional second parameter sets the message associated with the exception~
UPDATE: This is not an issue with optionality as I originally thought. The PR changes have been updated.

`Kernel.raise` overloaded signatures are currently defined for individual classes: `String`, `Class` and `Exception`.
This makes it impossible to type-check a call to `raise` with a `T.any` type like `T.any(Exception, String)`.

I encountered this issue while trying to define a method similar to [the following](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28obj_or_str%3A%20T.any%28Exception%2C%20String%29%29.returns%28T.noreturn%29%20%7D%0Adef%20raise_obj_or_string%28obj_or_str%29%0A%20%20raise%20obj_or_str%0Aend) in my code: 
```ruby
sig { params(obj_or_str: T.any(Exception, String)).returns(T.noreturn) }
def raise_obj_or_string(obj_or_str)
  raise obj_or_str
end

# Expected String but found T.any(Exception, String) for argument arg0 https://srb.help/7002
#      6 |  raise obj_or_str
#                 ^^^^^^^^^^
#   Expected String for argument arg0 of method Kernel#raise (overload.1):
#     https://github.com/sorbet/sorbet/tree/master/rbi/core/kernel.rbi#L2913:
#     2913 |        arg0: String,
#                   ^^^^
#   Got T.any(Exception, String) originating from:
#     editor.rb:5:
#      5 |def raise_obj_or_string(obj_or_str)
#                                 ^^^^^^^^^^
# Errors: 1
```

This change fixes that and adds an additional test for this specific use case.

### Test plan
See included automated tests.

### DIFF
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -2910,22 +2910,14 @@ module Kernel
   sig {returns(T.noreturn)}
   sig do
     params(
-        arg0: String,
+        arg0: T.any(Class, Exception, String),
     )
     .returns(T.noreturn)
   end
   sig do
     params(
-        arg0: Class,
-        arg1: T.nilable(T.untyped),
-        arg2: T.nilable(T::Array[String]),
-    )
-    .returns(T.noreturn)
-  end
-  sig do
-    params(
-        arg0: Exception,
-        arg1: T.nilable(T.untyped),
+        arg0: T.any(Class, Exception),
+        arg1: T.untyped,
         arg2: T.nilable(T::Array[String]),
     )
     .returns(T.noreturn)
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -2917,7 +2917,7 @@ module Kernel
   sig do
     params(
         arg0: Class,
-        arg1: T.untyped,
+        arg1: T.nilable(T.untyped),
         arg2: T.nilable(T::Array[String]),
     )
     .returns(T.noreturn)
@@ -2925,7 +2925,7 @@ module Kernel
   sig do
     params(
         arg0: Exception,
-        arg1: T.untyped,
+        arg1: T.nilable(T.untyped),
         arg2: T.nilable(T::Array[String]),
     )
     .returns(T.noreturn)
--- a/test/cli/errors/test.out
+++ b/test/cli/errors/test.out
@@ -9,12 +9,12 @@ test/cli/errors/errors.rb:5: Unable to resolve constant `MyConstantWithTypo` htt
      .. |  MyConstantWithNoTypo = nil
           ^^^^^^^^^^^^^^^^^^^^
 
-test/cli/errors/errors.rb:15: Expected `String` but found `Integer` for argument `arg0` https://srb.help/7002
+test/cli/errors/errors.rb:15: Expected `T.any(Class, Exception, String)` but found `Integer` for argument `arg0` https://srb.help/7002
     .. |    raise arg # raise is defined by stdlib
                   ^^^
-  Expected `String` for argument `arg0` of method `Kernel#raise (overload.1)`:
+  Expected `T.any(Class, Exception, String)` for argument `arg0` of method `Kernel#raise (overload.1)`:
     https://github.com/sorbet/sorbet/tree/master/rbi/core/kernel.rbi#LCENSORED:
-      NN |        arg0: String,
+      NN |        arg0: T.any(Class, Exception, String),
                   ^^^^
   Got `Integer` originating from:
     test/cli/errors/errors.rb:13:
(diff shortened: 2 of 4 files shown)


iMacTia:
Sorry for the direct mention @jez, this is my first contribution so I'm not sure how I'm supposed to "ask a contributor to unblock the tests" 😅

neilparikh:
@iMacTia I've unblocked them for you. Usually what happens is that whoever ends up reviewing your PR will unblock them for you when they're reviewing it.

You have the following test failure:
```
rbi/core/kernel.rbi#L2920: `T.nilable(T.untyped)` is the same as `T.untyped` https://srb.help/5070 
2920 |        arg1: T.nilable(T.untyped),
```
The second parameter is already marked optional, with this line: `def raise(arg0=T.unsafe(nil), arg1=T.unsafe(nil), arg2=T.unsafe(nil)); end`; `arg1=T.unsafe(nil)` assigned a default value to it. You can verify that with [this example](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aif%20T.unsafe%28false%29%0A%20%20raise%20%22a%22%0Aend%0Aif%20T.unsafe%28false%29%0A%20%20raise%20RuntimeError%0Aend) as well.

iMacTia:
Thank you @neilparikh!

I'm a bit surprised by the error I'm getting then 🤔.
Do you have any tip on how to make that example work? It seems like Sorbet expects the second parameter to be provided even though it's optional

iMacTia:
OK I took a step back and did some digging, and I think I better understand the issue now.
This is not a signature issue per-se, but rather a limitation of Sorbet's in using overloaded signatures.
I stumbled on this old issue from 2019: https://github.com/sorbet/sorbet/issues/1483

The issue is the same, there's a variable defined as `T.any(A, B)` which passed to a stdlib method with signature overload that accepts both `A` or `B`. At the moment, sorbet is unable to "test" both scenarios (variable is `A` and variable is `B`) with two different overloaded signatures.

According to the collaborator chiming in, "the complexity of supporting it is tremendous" and "we don't currently intend to fix this limitation".
If that's still the case, then could we instead add a new overloaded signature for the `T.any` scenario(s)?
This should work even if the signature is tested against a subset as I just checked [in this snippet](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28param%3A%20T.any%28Class%2C%20Exception%2C%20String%29%29.void%20%7D%0Adef%20test%28param%29%0A%20%20puts%20param%0Aend%0A%0Asig%20%7B%20params%28param%3A%20T.any%28Exception%2C%20String%29%29.void%20%7D%0Adef%20support%28param%29%0A%20%20test%28param%29%0Aend%0A).

I'm gonna update the PR to try that approach, but any feedback on this is welcome (especially if we're somehow closer to improve Sorbet's management of these cases, in which case we wouldn't need such a workaround)

froydnj (changes requested):
The initial error you encountered:

```
# Expected String but found T.any(Exception, String) for argument arg0 https://srb.help/7002
#      6 |  raise obj_or_str
#                 ^^^^^^^^^^
#   Expected String for argument arg0 of method Kernel#raise (overload.1):
#     https://github.com/sorbet/sorbet/tree/master/rbi/core/kernel.rbi#L2913:
#     2913 |        arg0: String,
#                   ^^^^
#   Got T.any(Exception, String) originating from:
#     editor.rb:5:
#      5 |def raise_obj_or_string(obj_or_str)
#                                 ^^^^^^^^^^
# Errors: 1
```

is saying that you passed an object that was `T.any(Exception, String)` to `raise`.  But the closest overload Sorbet selected only took `String`, and didn't know anything about `Exception`.  Therefore, Sorbet was telling you that you could potentially be passing an `Exception` to a method that only expects `String`.  It has nothing to do with optionality of the arguments.  #1483 is, I think, only tangentially related.

I think the right thing to do here is to sync up the signatures between `Kernel.fail` and `Kernel.raise` (they are synonyms for the same function under the hood) by replacing the signatures for `Kernel.raise` with the signatures for `Kernel.fail`.  That will at least be much closer to solving the problem.

iMacTia:
Thank you for rerunning tests, I've noted the issues and I'm fixing them in my branch 👍

froydnj:

→ https://go/builds/bui_Ns3MzHAkRLC2Zz
→ https://go/builds/bui_Ns3MJK6GASk0SV
→ https://go/builds/bui_Ns3MUDIHd4Lksa

froydnj (approved):
Thank you!  And sorry for not seeing that #1483 was indeed applicable to this situation!

iMacTia:
No worries @froydnj! I wasn't much clear myself and created more confusion than anything 😅 
Thank you for the support with running tests 🙇

PR #6991, opened 2023-05-09
optimize allocations during sig running

froydnj:
cc @aadi-stripe 

### Motivation

Cutting down on the number of allocations that we do while we're running sigs.

The first commit is important for just saving runtime; set difference on lists is not super fast in the first place, and we can save doing one in the common case.

The third commit is pretty minor, but every little bit helps!

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/signature.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/signature.rb
@@ -19,9 +19,9 @@ class T::Private::Methods::Signature
     parameters = parameters.each_with_index.map do |(param_kind, param_name), index|
       [param_kind, param_name || "arg#{index}"]
     end
-    raw_arg_types = parameters.map do |_param_kind, param_name|
+    raw_arg_types = parameters.to_h do |_param_kind, param_name|
       [param_name, not_typed]
-    end.to_h
+    end
 
     self.new(
       method: method,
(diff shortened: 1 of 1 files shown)


gems/sorbet-runtime/lib/types/private/methods/signature.rb
```diff
@@ -61,17 +64,19 @@ def initialize(method:, method_name:, raw_arg_types:, raw_return_type:, bind:, m
     # If sig params are declared but there is a single parameter with a missing name
     # **and** the method ends with a "=", assume it is a writer method generated
     # by attr_writer or attr_accessor
-    writer_method = declared_param_names != [nil] && parameters == [[:req]] && method_name[-1] == "="
+    writer_method = declared_param_names != SIG_EMPTY_DECLARED_PARAMETERS && parameters == UNNAMED_REQUIRED_PARAMETERS && method_name[-1] == "="
     # For writer methods, map the single parameter to the method name without the "=" at the end
     parameters = [[:req, method_name[0...-1].to_sym]] if writer_method
     param_names = parameters.map {|_, name| name}
     missing_names = param_names - declared_param_names
```
maruth-stripe, line 71: might be minor: do we need to materialize this here? Seems like we only care about whether it's non-empty initially and only need to materialize if so

Could maybe save array allocation if not having missing params is the common case?

froydnj: This is an interesting idea!  Maybe we could avoid materializing `param_names` altogether if we were willing to write out some loops manually.

PR #6994, opened 2023-05-09
more signature allocation optimizations

froydnj:
cc @zanker-stripe 

### Motivation

Follow on from #6991 with suggestions from the team at Stripe.  We can't modify the `parameters` array in-place in `new_untyped`, unfortunately, because it's coming from Ruby itself, which freezes the array.

I was too lazy to merge the loops in `new_untyped`.  Maybe another day.

This is probably going to conflict with #6993 cc @maruth-stripe .

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/signature.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/signature.rb
@@ -12,8 +12,8 @@ class T::Private::Methods::Signature
   UNNAMED_REQUIRED_PARAMETERS = [[:req]].freeze
 
   def self.new_untyped(method:, mode: T::Private::Methods::Modes.untyped, parameters: method.parameters)
-    # Using `Untyped` ensures we'll get an error if we ever try validation on these.
-    not_typed = T::Private::Types::NotTyped.new
+    # Using `NotTyped` ensures we'll get an error if we ever try validation on these.
+    not_typed = T::Private::Types::NotTyped::INSTANCE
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/private/methods/signature.rb
```diff
@@ -82,8 +83,13 @@ def initialize(method:, method_name:, raw_arg_types:, raw_return_type:, bind:, m
     if parameters.size != raw_arg_types.size
       raise "The declaration for `#{method.name}` has arguments with duplicate names"
     end
+    params_size = parameters.size
+    i = 0
+    while i < params_size
```
jez, line 88: This is a completely minor thing but I always liked the pattern of structuring this like
```suggestion
    i = -1
    while i < params_size
      i += 1
```

So that all the index logic happens in one place.

jez (approved):

gems/sorbet-runtime/lib/types/private/methods/signature.rb
froydnj: The proposed change is not the common pattern for sorbet-runtime, and it also has an edge case when `params_size` is 0.  I'm opting to leave this as is, though I do agree that for this particular function, having some 50 lines between test and increment is suboptimal.

jez: Oh right I forgot that i usually use this with for auto loops

PR #6998, opened 2023-05-10
add rudimentary sig benchmark

froydnj:
### Motivation

We've been doing some PRs that change how performant signature running should be.  Let's add a small benchmark that sees how expensive it is to run all of sorbet-runtime's signatures (admittedly not the greatest benchmark, but there's a reason this says "rudimentary" on the tin).

### Test plan

See included automated tests.


jez (approved):
Can you share the output in a comment on this PR, for posterity's sake?

PR #6999, opened 2023-05-10
Update `Enumerable#tally` sig for Ruby 3.1

amomchilov:
### Motivation

As of Ruby 3.1, `Enumerable#tally` takes an optional hash that it will accumulate into (instead of allocating a new Hash).

Requested in https://bugs.ruby-lang.org/issues/17744
Implemented in https://github.com/ruby/ruby/pull/4318

### DIFF
--- a/rbi/core/enumerable.rbi
+++ b/rbi/core/enumerable.rbi
@@ -1744,8 +1744,12 @@ module Enumerable
   # ```ruby
   # ["a", "b", "c", "b"].tally  #=> {"a"=>1, "b"=>2, "c"=>1}
   # ```
-  sig {returns(T::Hash[Elem, Integer])}
-  def tally(); end
+  #
+  # If a hash is given, the number of occurrences is added to each value
+  # in the hash, and the hash is returned. The value corresponding to
+  # each element must be an integer.
+  sig {params(hash: T::Hash[Elem, Integer]).returns(T::Hash[Elem, Integer])}
+  def tally(hash = {}); end
 
   ### Implemented in C++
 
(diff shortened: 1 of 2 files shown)


froydnj:

→ https://go/builds/bui_Ns7ESlxswkNFDQ
→ https://go/builds/bui_Ns7EMmE2xw3oxy
→ https://go/builds/bui_Ns7EIqlAetaAjC

amomchilov:
@froydnj Sounds good to me!

I just updated this to:

1. Include the new doc comment that accompanied this feature https://github.com/ruby/ruby/pull/4318/files#diff-fb3df03e6e34e16b0bf29a102c4acaed26a8e59d183af2f386223bf925414303R1063-R1065
    * Out of curiosity, what's the motivation of the doc comments on these shims? They seems redundant to the main Ruby stdlib docs, and just present an opportunity to go out of sync
2. Add a test case in `test/testdata/rbi/enumerable.rb`
    * Should I be changing any other tests?
    * Should I cover the unhappy path (giving a Hash arg whose keys aren't `Elem`s, and/or whose values aren't `Integer`)?


Is there anything else involved in updating a shim like this?

froydnj:
>        * Out of curiosity, what's the motivation of the doc comments on these shims? They seems redundant to the main Ruby stdlib docs, and just present an opportunity to go out of sync

The doc comments ensure that people can get hover documentation in VSCode (or similar) for these methods.

>     2. Add a test case in `test/testdata/rbi/enumerable.rb`
>        
>        * Should I be changing any other tests?
>        * Should I cover the unhappy path (giving a Hash arg whose keys aren't `Elem`s, and/or whose values aren't `Integer`)?

Tests for RBI changes are best-effort, so any tests that you add are great!  We would generally only add tests for the second case in situations where the methods involved have corner cases that are not immediately obvious.  See #6663 for an example.

amomchilov:
@froydnj I had a silly copy/paste mistake that caused that test failure 🤦 

Should be good now!

Is there anything else that I should change here, or is it ready to merge?

PR #7007, opened 2023-05-12
cache `T.type_parameter` return values for the same typename

froydnj:
### Motivation

We came across some use cases for generic code that also exist inside hot code paths where allocations are undesirable.  We might as well make `T.type_parameter` as cheap as we can.

### Test plan

See included automated tests.  I thought about making `new` a `private_class_method`, but since `T::Types::TypeParameter` is technically a private API, something along the lines of this PR seems better.

### DIFF
--- a/gems/sorbet-runtime/lib/types/types/type_parameter.rb
+++ b/gems/sorbet-runtime/lib/types/types/type_parameter.rb
@@ -3,9 +3,28 @@
 
 module T::Types
   class TypeParameter < Base
+    module Private
+      @pool = {}
+
+      def self.cached_entry(name)
+        @pool[name]
+      end
+
+      def self.set_entry_for(name, type)
+        @pool[name] = type
+      end
+    end
+
     def initialize(name)
       raise ArgumentError.new("not a symbol: #{name}") unless name.is_a?(Symbol)
       @name = name
+    end
+
+    def self.make(name)
+      cached = Private.cached_entry(name)
+      return cached if cached
+
+      Private.set_entry_for(name, new(name))
     end
 
     def valid?(obj)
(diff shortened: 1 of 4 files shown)


gems/sorbet-runtime/lib/types/types/type_parameter.rb
```diff
@@ -3,11 +3,30 @@
...
       @name = name
     end
 
+    def self.make(name)
+      cached = Private.cached_entry(name)
+      return cached if cached
+
+      Private.set_entry_for(name, new(name))
+    end
+
```
jez, line 29: Should we add `private_class_method :new`?

gems/sorbet-runtime/lib/types/types/type_parameter.rb
```diff
@@ -3,11 +3,30 @@
...
+        @pool[name] = type
+      end
+    end
+
     def initialize(name)
       raise ArgumentError.new("not a symbol: #{name}") unless name.is_a?(Symbol)
       @name = name
     end
 
+    def self.make(name)
```
jez, line 23: Should we add an entry to the RBI in the payload for this new method?

gems/sorbet-runtime/lib/types/types/type_parameter.rb
froydnj: I thought about this, but decided since this class is publicly visible -- though I didn't find any bare usages of it in Stripe's codebase -- that I would leave this API alone.

gems/sorbet-runtime/lib/types/types/type_parameter.rb
froydnj: Sure, we can do that.

PR #7008, opened 2023-05-12
Add environment variable to opt into recursive type checks 

maruth-stripe:
TODO:
1. I'm not sure how to enforce this is only enabled for tests
2. Add more tests to check recursive type checking is enabled

### Motivation

Recursive type checking was disabled due to its performance cost. But it can still be useful to let users opt into it while running tests locally to identify errors that would otherwise not be caught eg `T.let([nil], T::Array[Integer])`

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/_types.rb
+++ b/gems/sorbet-runtime/lib/types/_types.rb
@@ -148,8 +148,11 @@ module T
   # doesn't match the type.
   def self.let(value, type, checked: true)
     return value unless checked
-
-    Private::Casts.cast(value, type, "T.let")
+    if Configuration.recursively_check_in_tests?
+      Private::Casts.cast_recursive(value, type, "T.let")
+    else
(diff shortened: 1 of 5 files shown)


gems/sorbet-runtime/lib/types/_types.rb
```diff
@@ -148,8 +148,11 @@ def self.cast(value, type, checked: true)
   # doesn't match the type.
   def self.let(value, type, checked: true)
     return value unless checked
-
-    Private::Casts.cast(value, type, "T.let")
+    if Configuration.recursively_check_in_tests?
+      Private::Casts.cast_recursive(value, type, "T.let")
+    else
+      Private::Casts.cast(value, type, "T.let")
+    end
```
jez, line 155: This degrades performance of `let` for something that we only ever intend to exist for one-off debugging.

PR #7009, opened 2023-05-13
Autocorrect e -> T.must(e) for T.let(e, U) where e: S with S = T.nilable(U)

maruth-stripe:
### Motivation

Better autocorrects

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/infer/autocorrect_t_let_nilable.rb
+++ b/test/testdata/infer/autocorrect_t_let_nilable.rb
@@ -1,0 +1,3 @@
+# typed: true
+
+T.let(T.let(5, T.nilable(Integer)), Integer)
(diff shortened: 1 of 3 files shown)


test/testdata/infer/autocorrect_t_let_nilable.rb
```diff
@@ -0,0 +1,5 @@
+# typed: true
+
+x = T.let(5, T.nilable(Integer))
+  T.let(x, Integer)
+# ^^^^^^^^^^^^^^^^^ error: Argument does not have asserted type `Integer`
```
jez, line 5: not blocking but I typically write these tests like this:
```suggestion
extend T::Sig

sig {params(x: T.nilable(Integer)).void}
def example(x)
  T.let(x, Integer)
# ^^^^^^^^^^^^^^^^^ error: Argument does not have asserted type `Integer`
end
```

- Sometimes you can get into weird behavior because `T.let` is both a way to annotate a type and also to pin a variable's type in a loop. This test isn't one of them, but i still tend to write it differently
- The indentation of code inside the method naturally makes room to write the `^^^` assertion

jez (approved):

PR #6978, opened 2023-05-03
Delete SelfNew

jez:
### Motivation

The recent changes to use `T.attached_class` as the return type of `Class.new` means that we don't actually have to do any extra work to compute the return type of `new`.

In the beginning of Sorbet, the `Class_new` intrinsic was responsible for looking at the type of the receiver, using the type of the receiver to compute an instance type, and setting that instance type as the result of `new`. We can delete that code now.

What that means is that Sorbet will be smart enough to know that _inside_ a singleton class, `self.new` can have the type `T.attached_class`. Outside a singleton class, Sorbet will know that `Child.new` will have type `Child`. This works because of how Sorbet's `resultTypeAsSeenFrom' helper works, which instantiates a generic type member from the perspective of either within a method or at the call site of a method.

The `SelfNew` intrinsic was originally conceived as a hack, to allow people to write strongly-typed methods that called `new` and were annotated as returning `T.attached_class`. Before, `self.new` inside `Parent.make` would have type `Parent`:

```ruby
class Parent
  def self.make
    new
  end
end
```

so we added the `self.new` intrinsic to work around it.

No more hacks!

### Test plan

See included automated tests.

PR #7020, opened 2023-05-16
Create applied types for type in sealed_subclasses

### LINKED ISSUE #3618: Error using sealed_subclasses is inconsistent with hand-written type
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%20%20sealed!%0Aend%0A%0Aclass%20Child1%20%3C%20Parent%0Aend%0A%0Aclass%20Child2%20%3C%20Parent%0Aend%0A%0Aextend%20T%3A%3ASig%0Asig%20%7Bparams(x%3A%20T.any(T.class_of(Child1)%2C%20T.class_of(Child2))).void%7D%0Adef%20foo(x)%0A%20%20%23%20why%20is%20this%20work...%0A%20%20T.let(x%2C%20T.class_of(Parent))%0Aend%0A%0A%23%20but%20this%20doesn't%3F%0Ay%20%3D%20T.must(Parent.sealed_subclasses.first)%0AT.let(y%2C%20T.class_of(Parent)))

```ruby
# typed: true

class Parent
  extend T::Helpers
  abstract!
  sealed!
end

class Child1 < Parent
end

class Child2 < Parent
end

extend T::Sig
sig {params(x: T.any(T.class_of(Child1), T.class_of(Child2))).void}
def foo(x)
  # why is this work...
  T.let(x, T.class_of(Parent))
end

# but this doesn't?
y = T.must(Parent.sealed_subclasses.first)
T.let(y, T.class_of(Parent))
```

#### Observed output

```
editor.rb:24: Argument does not have asserted type T.class_of(Parent) https://srb.help/7007
    24 |T.let(y, T.class_of(Parent))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Got T.any(T.class_of(Child1), T.class_of(Child2)) originating from:
    editor.rb:23:
    23 |y = T.must(Parent.sealed_subclasses.first)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

As far as I can see, the type in the `sig` and the type in the error message should be the same. So since the `T.let` with `x` doesn't have an error, neither should the one with `y`.

- - -

The type in the error message is something we synthesize in support of sealed subclasses, so there's probably something more subtle going on. If I had to guess, I'd guess that one of the `T.class_of` types is an `AppliedType` and the other is a `ClassType`. So I'd probably start by printing the raw forms of those types and see what's different about them.

The input to `T.let` should just be a type—there's probably not any global state at play here—so it's almost definitely that the two in memory representations of those static types are represented by different structures even though they're printed the same way in the error message. If we can get the types' structures to be the same, the errors (or lack thereof) should be the same.

jez:
### Motivation

Fixes #3618.

In recordSealedSubclass, we use `make_type<ClassType>(sym)` instead of
`sym.data(ctx)->externalType(ctx)` because we call recordSealedSubclass
from ResolveCosntantsWalk, before we've created `<AttachedClass>` type
members for all classes.

But that's weird, because all singleton classes are supposed to be generic
classes and are thus supposed to be applied types.

Luckily, there's no way to notice that we're using ClassTypes here, so we
can record the sealed subclasses during ResolveConstantsWalk and then fix
up the type later ResolveTypeMembersAndFieldsWalk (in the portion of it
that runs after we've computed all the type members).

This is a little annoying, because it adds code to the single-threaded
portion of resolver, but I think this is worth it for correctness.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -1875,6 +1875,9 @@ TypePtr sealedSubclassesToAppliedImpl(const GlobalState &gs, TypePtr currentClas
     while (auto orType = cast_type<OrType>(currentClasses)) {
         if (isa_type<AppliedType>(orType->right)) {
             // We already converted ClassTypes to AppliedTypes once; don't need to do it again.
+            // (This might happen if e.g. a class is opened once in a Ruby file and again in an RBI)
+            // However, if all the subclasses are modules, this short circuit might never happen.
+            // That's fine though, because this method should be idempotent.
             return currentClasses;
         }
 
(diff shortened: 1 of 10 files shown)


jez:
* master
  * **PR #7020** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/7020" target="_blank">Graphite</a>  👈
    * **PR #6978** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/6978" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/7020?utm_source=stack-comment).

PR #7021, opened 2023-05-16
Use valid syntax for `T.attached_class` in suggseted sig

### LINKED ISSUE #7004: Bad autocorrect: `T.attached_class (of ...)` in sig
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Aclass%20Foo%0A%20%20def%20self.generate%0A%20%20%20%20self.new%28%29%0A%20%20end%0Aend)

```ruby
# typed: strict
extend T::Sig

class Foo
  def self.generate
    self.new()
  end
end
```

#### Observed output

```
editor.rb:5: The method generate does not have a sig https://srb.help/7017
     5 |  def self.generate
          ^^^^^^^^^^^^^^^^^
  Autocorrect: Use -a to autocorrect
    editor.rb:5: Insert sig { returns(T.attached_class (of Foo)) }
     5 |  def self.generate
          ^
    editor.rb:5: Insert extend T::Sig
     5 |  def self.generate
        ^
Errors: 1
```

#### Expected behavior

- - -

jez:
### Motivation

Fixes #7004

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -1827,11 +1827,12 @@ void ClassOrModule::recordSealedSubclass(MutableContext ctx, ClassOrModuleRef su
     // Avoid using a clobbered `this` pointer, as `singletonClass` can cause the symbol table to move.
     ClassOrModuleRef selfRef = this->ref(ctx);
 
-    // We record sealed subclasses on a magical method called core::Names::sealedSubclasses(). This is so we don't
-    // bloat the `sizeof class Symbol` with an extra field that most class sybmols will never use.
-    // Note: We had hoped to ALSO implement this method in the runtime, but we couldn't think of a way to make it work
-    // that didn't require running with the help of Stripe's autoloader, specifically because we might want to allow
-    // subclassing a sealed class across multiple files, not just one file.
+    // We record sealed subclasses on the method called core::Names::sealedSubclasses().
+    //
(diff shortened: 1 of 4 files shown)


core/Symbols.cc
```diff
@@ -1847,7 +1848,6 @@ void ClassOrModule::recordSealedSubclass(MutableContext ctx, ClassOrModuleRef su
     const OrType *orT = nullptr;
     while ((orT = cast_type<OrType>(*iter))) {
         auto right = cast_type_nonnull<ClassType>(orT->right);
-        ENFORCE(left);
```
neilparikh, line 1850: Why is this change here?

neilparikh (approved):

core/Symbols.cc
jez: I thought I committed this to a different branch, but it looks like when I was rebasing it got added to the wrong branch. I’ll fix tomorrow

jez: interestingly, this didn't end up committed on two branches, and since it's a no-op I'm just going to land it now so I don't have to wait for a passing rebuild.

jez: interestingly, this didn't end up committed on two branches, and since it's a no-op I'm just going to land it now so I don't have to wait for a passing rebuild.

PR #7022, opened 2023-05-16
Add option to output blames for `T.untyped` usages in a JSON file

maruth-stripe:
### Motivation

This introduces a couple of configuration options
1. **New build mode**: now you can pass `--config=untyped-blame` in addition to any of the existing build modes (eg: `--config=release-linux --config=untyped-blame`, which will create a build where blamed symbols for untyped usages are tracked.
2. `--untyped-blame` : this flag will make Sorbet spit out a JSON file containing a list of symbols blamed for untyped usages in the format `{"path": <path/to/file/with/violation>, "owner" : <namespace::owning::symbol>, "symbol": the_symbol, "count": N }`

To use this mode:
1. build sorbet with `./bazel build //main:sorbet --config=untyped-blame --config=release-linux # replace second config with whatever`
2. Run with `/path/to/sorbet --track-untyped --print=untyped-blame:path-to-blame-file.json`

### Test plan

ran locally


tools/BUILD
```diff
@@ -39,18 +39,18 @@ compilation_database(
...
+        # "//compiler/Errors:Errors",
+        # "//compiler/IREmitter:IREmitter",
+        # "//compiler/IREmitter/Payload:Payload",
+        # "//compiler/IREmitter/Payload:generate_load_payload",
+        # "//compiler/IREmitter/Payload:postprocess_payload",
+        # "//compiler/IREmitter/Payload:pseudo_lib",
+        # "//compiler/IREmitter/Payload:vm_payload_for_compile_commands",
+        # "//compiler/Linker:Linker",
+        # "//compiler/ObjectFileEmitter:ObjectFileEmitter",
+        # "//compiler/Passes:Passes",
```
jez, line 53: sir?

main/realmain.cc
```diff
@@ -956,15 +956,25 @@ int realmain(int argc, char *argv[]) {
...
+                auto path = sym.loc(*gs).file().data(*gs).path();
+                auto owner = sym.owner(*gs).show(*gs);
+                withNames.emplace_back(absl::StrCat(path, ",", owner, ",", sym.showFullName(*gs)), e.second);
                 sum += e.second;
             }
             fast_sort(withNames, [](const auto &lhs, const auto &rhs) -> bool { return lhs.second > rhs.second; });
             for (auto &p : withNames) {
-                logger->error("Typing `{}` would impact {}% callsites({} out of {}).", p.first, p.second * 100.0 / sum,
-                              p.second, sum);
+                logger->error("{},{},{}", p.first, p.second, sum);
```
jez, line 970: Slight anti-preference for CSV. We have method names with `,` in them:

https://sorbet.run/?arg=--print=symbol-table#%23%20typed%3A%20true%0Ait%20%22left%2Cright%22%20do%0Aend

I'd prefer doing JSON and generating it using the technique you can find in `tracing.cc`, using rapidjson.

jez (changes requested):
This approach is going to need work. This slows down all calls to Sorbet (substantially) even if they don't pass this command line option.

maruth-stripe:
build numbers:
```

$ ./bazel
clean --expunge
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.

$ ./bazel build --config=release-linux
Starting local Bazel server and connecting to it...
WARNING: The following configs were expanded more than once: [static-libs]. For repeatable flags, repeats are counted twice and may lead to unexpected behavior.
WARNING: Usage: bazel build <options> <targets>.
Invoke `bazel help build` for full description of usage and options.
Your request is correct, but requested an empty set of targets. Nothing will be built.
INFO: Analyzed 0 targets (0 packages loaded, 0 targets configured).
INFO: Found 0 targets...
INFO: Elapsed time: 3.326s, Critical Path: 0.03s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action

$ ./bazel build //main:sorbet --config=release-linux
WARNING: The following configs were expanded more than once: [static-libs]. For repeatable flags, repeats are counted twice and may lead to unexpected behavior.
INFO: Analyzed target //main:sorbet (154 packages loaded, 3732 targets configured).
INFO: Found 1 target...
Target //main:sorbet up-to-date:
  bazel-bin/main/sorbet
INFO: Elapsed time: 443.907s, Critical Path: 97.81s
INFO: 1764 processes: 404 internal, 1360 local.
INFO: Build completed successfully, 1764 total actions

$ ./bazel build //main:sorbet --config=release-linux --config=untyped-blame
WARNING: The following configs were expanded more than once: [static-libs]. For repeatable flags, repeats are counted twice and may lead to unexpected behavior.
INFO: Build option --copt has changed, discarding analysis cache.
INFO: Analyzed target //main:sorbet (1 packages loaded, 3750 targets configured).
INFO: Found 1 target...
Target //main:sorbet up-to-date:
  bazel-bin/main/sorbet
INFO: Elapsed time: 232.479s, Critical Path: 87.06s
INFO: 477 processes: 1 internal, 476 local.
INFO: Build completed successfully, 477 total actions
```

core/errors/infer.cc
```diff
@@ -3,7 +3,7 @@
 
 namespace sorbet::core::errors::Infer {
 
-ErrorClass errorClassForUntyped(const GlobalState &gs, FileRef file) {
+ErrorClass errorClassForUntyped(const GlobalState &gs, FileRef file, SymbolRef blame) {
```
jez, line 6:
```suggestion
ErrorClass errorClassForUntyped(const GlobalState &gs, FileRef file, const TypePtr & untyped) {
```

- This way, we don't call `untypedBlame` unless we know we're going to use it
- Using `const TypePtr &` should be faster than `TypePtr` because it won't have to increment/decrement an atomic ref count.

core/errors/infer.cc
```diff
@@ -23,6 +23,12 @@ ErrorClass errorClassForUntyped(const GlobalState &gs, FileRef file) {
         prodHistogramInc("untyped.usages", file.id());
     }
 
+    prodCounterInc("types.input.untyped.usages");
+
+    if constexpr (sorbet::track_untyped_blame_mode) {
+        histogramInc("untyped.blames", blame.rawId());
```
jez, line 29:
```suggestion
        histogramInc("untyped.blames", untyped.untypedBlame().rawId());
```

core/errors/infer.cc
```diff
@@ -23,6 +23,12 @@ ErrorClass errorClassForUntyped(const GlobalState &gs, FileRef file) {
         prodHistogramInc("untyped.usages", file.id());
     }
 
+    prodCounterInc("types.input.untyped.usages");
```
jez, line 26: So I think we want this at the top of the method, before the `gs.trackUntyped`.

That way, it works like how the `types.input.sends.typed`/`types.input.sends.total` stats work right now, where we can report the codebase-wide count without special flags, but if we want to break it down by file or by blame, then there are more hoops.

core/errors/infer.cc
jez, line 29: Genuine questions, curious to hear your thoughts:

- Should we use `prodHistogramInc` here instead of having `enable_counters` be always on?
- Should we run this code even if `debug_mode` is on, so that we get test coverage for this feature just by running the test suite? Right now, I'm not sure whether it's possible to write automated tests for this feature.

main/options/options.cc
```diff
@@ -1109,6 +1117,11 @@ void readOptions(Options &opts,
         if (sorbet::debug_mode) {
             opts.suggestSig = raw["suggest-sig"].as<bool>();
         }
+        if (sorbet::track_untyped_blame_mode) {
+            opts.printBlameUntyped = raw["blame-untyped"].as<bool>();
+            opts.untypedBlameFilePath = raw["blame-untyped-path"].as<string>();
```
jez, line 1122: We should use a printer for this. Grep for how PrintOptions work.

main/options/options.cc
```diff
@@ -1109,6 +1117,11 @@ void readOptions(Options &opts,
...
+            opts.trackUntyped = opts.trackUntyped || opts.printBlameUntyped;
```
jez, line 1123: I'm generally not a fan of one option implying another option, because it can cause confusion. You'll notice that we have lots of code like this:

https://github.com/sorbet/sorbet/blob/a7cdaa9c196ff1c8284a19ce7781a9d9c2ff37ea/main/options/options.cc#L837-L842

which avoids assuming implicit options in favor of reporting an error during arg parsing saying to explicitly pass the other arg.

main/realmain.cc
```diff
@@ -950,19 +950,10 @@ int realmain(int argc, char *argv[]) {
...
-                withNames.emplace_back(core::SymbolRef::fromRaw(e.first).showFullName(*gs), e.second);
-                sum += e.second;
-            }
-            fast_sort(withNames, [](const auto &lhs, const auto &rhs) -> bool { return lhs.second > rhs.second; });
-            for (auto &p : withNames) {
-                logger->error("Typing `{}` would impact {}% callsites({} out of {}).", p.first, p.second * 100.0 / sum,
-                              p.second, sum);
+        if constexpr (sorbet::track_untyped_blame_mode) {
+            if (opts.printBlameUntyped) {
+                auto untypedBlames = getAndClearHistogram("untyped.blames");
```
jez, line 955: I think we should move this outside the `if (opts.printBlameUntyped)`.

If someone built sorbet in this custom release mode, but forgot to pass this print option, we're not going to call `getAndClearHistogram`, which is going to spam statsd with a high-cardinality metric, which is not good.

main/pipeline/pipeline.h
```diff
@@ -68,5 +68,8 @@ bool cacheTreesAndFiles(const core::GlobalState &gs, WorkerPool &workers,
 // Exported for tests only.
 std::string fileKey(const core::File &file);
 
+void printUntypedBlames(const core::GlobalState &gs, const UnorderedMap<long, long> untypedBlames,
```
jez, line 71: This should take the map by `const &`, to avoid duplicating it.

main/pipeline/pipeline.cc
```diff
@@ -1264,4 +1264,43 @@ vector<ast::ParsedFile> autogenWriteCacheFile(const core::GlobalState &gs, const
...
+void printUntypedBlames(const core::GlobalState &gs, const UnorderedMap<long, long> untypedBlames,
+                        const std::string &blameFilePath) {
+#ifndef SORBET_REALMAIN_MIN
+
+    rapidjson::StringBuffer result;
+    rapidjson::Writer<rapidjson::StringBuffer> writer(result);
+
+    writer.StartArray();
+
+    for (auto e : untypedBlames) {
```
jez, line 1276:
```suggestion
    for (auto &[symId, count] : untypedBlames) {
```

I have a slight preference for this iteration style, so that people don't have to remember what `.first` and `.second` are. But that's completely non blocking.

main/pipeline/pipeline.cc
```diff
@@ -1264,4 +1264,43 @@ vector<ast::ParsedFile> autogenWriteCacheFile(const core::GlobalState &gs, const
...
+        }
+
+        writer.StartObject();
+
+        writer.String("path");
+        writer.String(std::string(sym.loc(gs).file().data(gs).path()));
+
+        writer.String("owner");
+        auto owner = sym.owner(gs).show(gs);
+        writer.String(std::string(owner));
```
jez, line 1289: iiuc, `owner` is already a string, right?
```suggestion
        writer.String(owner);
```

main/pipeline/pipeline.cc
```diff
@@ -1264,4 +1264,43 @@ vector<ast::ParsedFile> autogenWriteCacheFile(const core::GlobalState &gs, const
...
+
+        writer.String("symbol");
```
jez, line 1291:
```suggestion
        writer.String("name");
```

Let's keep the same terminology as Sorbet uses for this field?

main/pipeline/pipeline.cc
```diff
@@ -1264,4 +1264,43 @@ vector<ast::ParsedFile> autogenWriteCacheFile(const core::GlobalState &gs, const
...
+    rapidjson::StringBuffer result;
+    rapidjson::Writer<rapidjson::StringBuffer> writer(result);
+
+    writer.StartArray();
+
+    for (auto e : untypedBlames) {
+        auto sym = core::SymbolRef::fromRaw(e.first);
+        if (!sym.exists() || !sym.loc(gs).exists()) {
+            continue;
+        }
```
jez, line 1280: What do you think about including these? I think that knowing how many blind spots we have is value. e.g., the `::<none>` results in your earlier reports give us an approximate estimate of how much untyped can't be blamed due to things like `untypedUntracked` and `T.unsafe` etc.

jez: And if we do decide not to include this, in that case let's opt to filter them out before they get into the mapping (i.e., in errorClassForUntyped), so that we're not doing extra work.

.bazelrc
```diff
@@ -94,6 +94,8 @@ build:release-debug-common --config=forcedebug
 build:release-debug-common --config=debugsymbols
 build:release-debug-common --config=skipslowenforce
 
+build:untyped-blame --copt=-DTRACK_UNTYPED_BLAME_MODE
```
jez, line 97: Should this also be `blame-untyped`, so that the word order isn't flipped from the command line option?

jez: Want to leave a comment here about why we have this (e.g., how it would be too slow to do all the time, but doing it in debug mode is also too slow on large codebases)?

Or if you want to leave that comment somewhere else in the PR, feel free.

main/pipeline/pipeline.cc
maruth-stripe: Done, these are included now; the `path` will just be `<none>` for these

I'm assuming `.owner` will always exist for a `SymbolRef` though

jez: Yeah I'm pretty sure that `::<root>` has itself as its owner, so everything is owned by `::<root>` transitively (nothing is unowned)

core/errors/infer.cc
```diff
@@ -23,6 +24,12 @@ ErrorClass errorClassForUntyped(const GlobalState &gs, FileRef file) {
         prodHistogramInc("untyped.usages", file.id());
     }
 
+    // we also run this code in test mode so we get some rudimentary coverage
+    // by running this path when running tests.
```
jez, line 28:
```suggestion
    // we also run this code in test mode so we get some rudimentary coverage
    // by running this path when running tests.
    //
    // Keep this in sync with core::Types::untyped(...)
```

maruth-stripe:
Made the following changes:
1. Use a printer now instead of custom file output
2. Everything follows `untypedBlame` (i.e., not `blameUntyped` etc)
3. error if `--track-untyped` is not passed with printer
4. Use `TypePtr &` instead of `SymbolRef`
5. Most of this feature now also runs under `--config=dbg` to exercise the relevant codepaths when running tests for very rudimentary coverage; developing automated tests for this seems challenging. We should probably do it at some point, but since doesn't affect actual typechecking, probably punt it for now.

core/errors/infer.h
```diff
@@ -1,6 +1,7 @@
 #ifndef SORBET_CORE_ERRORS_INFER_H
 #define SORBET_CORE_ERRORS_INFER_H
 #include "core/Error.h"
+#include "core/SymbolRef.h"
```
jez, line 4: Does this need to be
```suggestion
#include "core/TypePtr.h"
```

?

core/types/types.cc
```diff
@@ -40,7 +40,7 @@ TypePtr Types::untypedUntracked() {
 }
 
 TypePtr Types::untyped(const sorbet::core::GlobalState &gs, sorbet::core::SymbolRef blame) {
-    if (sorbet::debug_mode && blame.exists()) {
+    if ((sorbet::track_untyped_blame_mode || sorbet::debug_mode) && blame.exists()) {
         return make_type<BlamedUntyped>(blame);
     } else {
         return untypedUntracked();
```
jez, line 46: Out of an abundance of caution, what do you think of rewriting this a little differently:

```cpp
TypePtr Types::untyped(const sorbet::core::GlobalState &gs, sorbet::core::SymbolRef blame) {
    if constexpr (!sorbet::track_untyped_blame_mode && !sorbet::debug_mode) {
        return untypedUntracked();
    }

    if (blame.exists()) {
        return make_type<BlamedUntyped>(blame);
    } else {
        return untypedUntracked();
    }
}
```

main/pipeline/pipeline.cc
```diff
@@ -1264,4 +1264,44 @@ vector<ast::ParsedFile> autogenWriteCacheFile(const core::GlobalState &gs, const
...
+
+        writer.String("count");
+        writer.Int64(count);
+
+        writer.EndObject();
+    }
+
+    writer.EndArray();
+
+    opts.print.UntypedBlame.print(result.GetString());
```
jez, line 1303: You are running this printer even if the printer was not requested.

Look how the other `opts.print` calls check whether the print config is enabled or not, and early exit from this function if it wasn't passed (e.g., in pipeline::printFileTable).

main/pipeline/pipeline.cc
```diff
@@ -1264,4 +1264,44 @@ vector<ast::ParsedFile> autogenWriteCacheFile(const core::GlobalState &gs, const
...
+        writer.String("path");
+        if (sym.exists() && sym.loc(gs).exists()) {
+            writer.String(std::string(sym.loc(gs).file().data(gs).path()));
+        } else {
+            writer.String("<none>");
+        }
+
+        writer.String("owner");
+        auto owner = sym.owner(gs).show(gs);
+        writer.String(owner);
```
jez, line 1290: It's _possible_ we want to tweak this output.

I think that for a method like

```ruby
A::B::C.foo
```

This will show up as

```json
{
    "owner": T.class_of(A::B::C),
    "name: "foo"
}
```

That makes it a _little_ tricky to work with this data (e.g., you can't group by owner unless you process the data a bit.

Some possibilities:

- We add another field like `ownerAttached` and populate it by following the chain of `attachedClass` all the way to the bottom (e.g., `T.class_of(T.class_of(T.class_of(A)))` gets unwrapped all the way to `A`)
- We make `owner` be like `ownerAttached`, and make `name` indicate whether it's a singleton or instance method somehow(?)
- We make `owner` be like `ownerAttached` and add another field like `exactOwner` or something
- We make `owner` be like `ownerAttached` and keep `name` as is, but a add another field like `absoluteName` or something that has the name like it appears in the old `suggestSig` code
- We give up on all of this and just only aggregate by file/package, not by owner. Realistically, aggregating by file is _maybe_ enough? Except that a single class will have some methods from an RBI and some from ruby source files

main/pipeline/pipeline.cc
```diff
@@ -1264,4 +1264,44 @@ vector<ast::ParsedFile> autogenWriteCacheFile(const core::GlobalState &gs, const
...
+
+        writer.String("name");
+        writer.String(sym.name(gs).show(gs));
+
+        writer.String("count");
+        writer.Int64(count);
+
+        writer.EndObject();
```
jez, line 1298: We might want to conditionally add the package to this output. It's possible to join it on the file table output, and look up the package for a file, but it's a lot easier if the package information is denormalized.

main/pipeline/pipeline.cc
maruth-stripe: For simplicity done the following currently:
1. Added package information
2. Leaving `owner` as-is for simplicity. Prefer post-processing. If it ends up being a hurdle, we will come back to this and tidy it up.

common/statsd/statsd.cc
```diff
@@ -108,6 +108,9 @@ bool StatsD::submitCounters(const CounterState &counters, string_view host, int
     }
 
     for (auto &hist : counters.counters->histograms) {
+        if (ignoredHistograms->find(hist.first) != std::string_view::npos) {
```
jez, line 111: ... is this `->find` silently converting the C-style array to a pointer, and accessing the first element?

Let's not use C-style arrays

jez:
I think you’ve got a bug with your new C-style array. other than that i think this change is good

common/statsd/statsd.cc
jez: as in, i think you're just doing `ignoredHistograms[0].find(hist.first)`

maruth-stripe: ah great catch, replaced with `std::find` and `std::array`

jez (approved):

PR #7028, opened 2023-05-19
Revert "Create applied types for type in sealed_subclasses (#7020)"

jez:
This reverts commit 4f9ebc04ef5d1b4b953e3fd34ab43e1e89eca7fb.

Revert "Delete SelfNew (#6978)"

This reverts commit a26ec49a22ee3bdc4ede74fad5444d340a7a0d3a.

### Motivation

This caused a large regression in the time it takes us to run the
sealed_registraion_perf.rb test.

Previously, it took 2.3s on my laptop in debug builds to finish.

After these changes, it is now taking upwards of 60s, causing it to get killed
with a timeout.


jez:
This did also cause a regression in CI, but just not enough to push it over the 60s threshold.

Before these changes, the sealed registration test ran in 5s. It now runs in 23s.

(The configuration we run tests with in CI uses a higher optimization level than what I use by default on my development machine.)

PR #7029, opened 2023-05-19
Restore performance of recording sealed subclasses

jez:
### Motivation

The changes in #7020 caused a large regression in the time it takes us to run
the sealed_registraion_perf.rb test.

Previously, it took 2.3s on my laptop in debug builds to finish. After that PR,
it took upwards of 60s, causing it to get killed with a timeout.

Those changes _did_ also cause a regression in CI, but just not enough to push
it over the 60s threshold. Before #7020, the sealed registration test ran in 5s
in CI. After, it takes 23s. (The configuration we run tests with in CI uses a
higher optimization level than what I use by default on my development machine.)

### Summary

The idea is that we can use the ancestors stored on a ClassDef tree node to
figure out whether a class was directly a subclass of a sealed class.

I had previously thought we wouldn't be able to do this, becuase I thought
that since we'd have to run in ResolveTypeMembersAndFieldsWalk that would
have meant we'd be after ancestor linearization, which would have made it
hard to see whether a sealed class was a direct parent or a transitive
parent. But by looking at the ancestors node on the ClassDef we still have
that information on hand to disambiguate.

This approach means we can simply remove all the sealed class registration
from ResolveConstantsWalk and defer it to ResolveTypeMembersAndFieldsWalk,
which means we don't have to do an expensive walk tearing down and building
back up large union types--we can simply use externalType directly in
recordSealedSubclass.

The fact that we see ClassType and AppliedType here is subtle, but it's
because `T.class_of(MyModule)` is a ClassType while for something like
`T.class_of(MyClass)`, that would be an applied type (module singleton
classes don't currently have `<AttachedClass>` type members).

### Test plan

See included automated tests.

Also, I tested the sealed_registration_perf.rb test on my development machine
and it's back down to 1.6s.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -1820,12 +1820,12 @@ ClassOrModuleRef ClassOrModule::topAttachedClass(const GlobalState &gs) const {
     return classSymbol;
 }
 
-void ClassOrModule::recordSealedSubclass(MutableContext ctx, ClassOrModuleRef subclass) {
-    ENFORCE(this->flags.isSealed, "Class is not marked sealed: {}", ref(ctx).show(ctx));
-    ENFORCE(subclass.exists(), "Can't record sealed subclass for {} when subclass doesn't exist", ref(ctx).show(ctx));
+void ClassOrModule::recordSealedSubclass(GlobalState &gs, ClassOrModuleRef subclass) {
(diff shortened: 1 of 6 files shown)


ilyailya (approved):
LGTM!
Thanks for detailed commit messages!

PR #7030, opened 2023-05-19
Register `class` intrinsic on `Kernel`

jez:
### Motivation

This works around a bug where jump to def would not work on calls to
`.class`.

What was happening is that any time we register an intrinsic, if there
is not a method already defined for that intrinsic, we create a new one
on the fly. What usually happens is that later we see an RBI definition
for the intrinsic method too, which fills in things like arity,
location, etc.

But in this case, Object#class doesn't exist in our RBIs, only
Kernel#class does. So there was never a method that would come along and
define things like location information, so jump to definition wouldn't
work.

### Test plan

It's hard to write a test for this, because the problem is that we don't have location information for something in the payload. Showing that there's jump-to-def results for something in the payload doesn't work with our current test suite.

I verified that this fixes it when using jump to def on `.class` in Vim with Sorbet's LSP.

### DIFF
--- a/test/cli/suggest-object/test.out
+++ b/test/cli/suggest-object/test.out
@@ -1,4 +1,6 @@
 test/cli/suggest-object/suggest-object.rb:4: Method `class` does not exist on `M` https://srb.help/7003
      4 |    self.class
                  ^^^^^
+  Note:
+    Did you mean to `include Kernel` in this module?
 Errors: 1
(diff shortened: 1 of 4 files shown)


test/cli/suggest-object/test.out
```diff
@@ -1,4 +1,6 @@
 test/cli/suggest-object/suggest-object.rb:4: Method `class` does not exist on `M` https://srb.help/7003
      4 |    self.class
                  ^^^^^
+  Note:
+    Did you mean to `include Kernel` in this module?
```
froydnj, line 5: Snazzy!

froydnj (approved):

PR #7031, opened 2023-05-19
Report single error when a constant alias fails to resolve

jez:
### Motivation

Requested by @aadi-stripe

### Test plan

The commit history shows the test's behavior before and after this change.


ilyailya (approved):
LGTM!
Thanks for such a detailed test coverage!

PR #7032, opened 2023-05-19
Change how we print T.class_of types

jez:
### Motivation

Now that `T::Class` exists, it's possible to do something wild like
`T.all(T.class_of(...), T::Class[...])` which has the effect of
"applying" a type to the `T.class_of`'s `<AttachedClass>` type argument,
printing `T.class_of` types as simply `T.class_of(A)` can be misleading.
Sorbet will be able to collapse the `T.all` to just a `T.class_of`
applied type, which means that you can have two types that are printed
like `T.class_of(A)` but they're actually applied to differen types and
thus incompatible with each other.

This changes detects those cases where it's ambiguous what the
`T.class_of`'s `<AttachedClass>` type argument is, because it wasn't the
default value, and prints it explicitly.

It's still weird, because `T.class_of(A)[...]` is still not valid type
syntax, but that's not a new problem--Sorbet would already sometimes
print `T.class_of(A)[...]` when the singleton class had other type
templates.

We can revisit this printing if/when we make `T.class_of(...)[...]`
syntax valid.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/infer/class_of_printing.rb
+++ b/test/testdata/infer/class_of_printing.rb
@@ -1,0 +1,18 @@
+# typed: true
+extend T::Sig
+
+module Left; end
+module Right; end
+class Parent; end
+
+sig {params(x: T.all(T.class_of(Parent), T::Class[Left])).void}
+def example1(x)
+  T.reveal_type(x) # error: `T.class_of(Parent)`
+end
+
+sig {params(x: T.all(T.class_of(Parent), T::Class[Right])).void}
+def example2(x)
+  T.reveal_type(x) # error: `T.class_of(Parent)`
+  example1(x)
+  #        ^ error: `T.class_of(Parent)` but found `T.class_of(Parent)` for argument `x`
+end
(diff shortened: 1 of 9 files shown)


test/testdata/infer/class_of_printing.rb
```diff
@@ -0,0 +1,18 @@
...
+sig {params(x: T.all(T.class_of(Parent), T::Class[Left])).void}
+def example1(x)
+  T.reveal_type(x) # error: `T.class_of(Parent)[T.all(Parent, Left)]`
+end
+
+sig {params(x: T.all(T.class_of(Parent), T::Class[Right])).void}
+def example2(x)
+  T.reveal_type(x) # error: `T.class_of(Parent)[T.all(Parent, Right)]`
+  example1(x)
+  #        ^ error: `T.class_of(Parent)[T.all(Parent, Left)]` but found `T.class_of(Parent)[T.all(Parent, Right)]` for argument `x`
```
jez, line 17: 53bbdad3b5fba7a8392d2de459fcd63db3d9ef92 shows what this error message used to look like.

neilparikh (approved):

PR #7035, opened 2023-05-20
Move textDocument/documentSymbol to general availability

jez:
### Motivation

We have had this turned on for a while under beta features and haven't
seen it causing problems.

### Test plan

See included automated tests.


main/options/options.cc
```diff
@@ -780,8 +780,6 @@ void readOptions(Options &opts,
 
         bool enableAllLSPFeatures = raw["enable-all-experimental-lsp-features"].as<bool>();
         opts.lspAllBetaFeaturesEnabled = enableAllLSPFeatures || raw["enable-all-beta-lsp-features"].as<bool>();
-        opts.lspDocumentSymbolEnabled =
-            opts.lspAllBetaFeaturesEnabled || raw["enable-experimental-lsp-document-symbol"].as<bool>();
```
jez, line 784: I haven't removed the option itself because I find it's always annoying to have to jump back and forth between Sorbet versions when the new version says "this option is unrecognized."

We can delete the option some time in the future after more time has passed.

ilyailya (approved):
🚀

PR #7036, opened 2023-05-20
Add code action to delete T.unsafe

jez:
### Motivation

Sometimes I just want to see what fails if I delete a T.unsafe in some code I'm
looking at.

I figure if we can make it happen in one click, maybe more people will do it.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/code_actions/delete_unsafe.rb
+++ b/test/testdata/lsp/code_actions/delete_unsafe.rb
@@ -1,0 +1,15 @@
+# typed: true
+# selective-apply-code-action: refactor.rewrite
+
+T.unsafe(0).even?
+# ^ apply-code-action: [A] Delete T.unsafe
+
+T.unsafe(
+# ^ apply-code-action: [B] Delete T.unsafe
+  0
+).even?
+
+_foo = T
+  .unsafe(0)
+  # ^ apply-code-action: [C] Delete T.unsafe
+  .even?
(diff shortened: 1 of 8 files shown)


test/testdata/lsp/code_actions/delete_unsafe.rb
```diff
@@ -0,0 +1,15 @@
+# typed: true
```
neilparikh, line 1: Can you add a test case where the action is called on something other than `T.unsafe`? Ex.
```
0.even?
#  ^ apply-code-action: [A] Delete T.unsafe
```

neilparikh (approved):

test/testdata/lsp/code_actions/delete_unsafe.rb
jez: Given the way that our assert-no-code-action mechanism works, I don't think this is a worthwhile test to add. I'd have to make a whole empty file and assert that there are no code actions in that file with the kind "refactor.rewrite".

But it could easily be the sort of thing in the future where we add more refactor.rewrite code actions that apply in that file, and it wouldn't be useful to have the test fail in that case.

Also, I don't even know how those assert-no-code-action tests work? Because it would have to like simulate a code action at all points, but I'm sure it doesn't do that.

Anyways, I don't think this is a useful test to add given our test harness. If you disagree feel free to push back.

PR #7042, opened 2023-05-24
Show Sorbet actions in a dropdown

damolina-stripe:
Update the UX of clicking on the `Sorbet` status bar item to use a `QuickPick` control, instead of a `Notification` dialog, to list available actions to the user.
- This follows the pattern used by every other status bar item.
- When compared to a dialog, the dropdown control:
   - is automatically dismissed  when clicking away (vs having to explicitly dismiss via `✕` button).
   - action names  have enough room to be displayed (i.e. not trimmed). Descriptions are now a possibility too.
   - keyboard focus is set automatically on the control, so up/down arrows work (no need to TAB your way in).

The new UI looses one UI feature: when `ServerStatus` is `ERROR`, the old UI would show the last error as part of the notification message but there is no place for such custom message in a quick pick dialog.  Users can still see the "last error" from the "Sorbet" status bar item's tooltip or use the `View Output` action.

**Before**
<p align="center">
image
</p>

**After**
<p align="center">
image
</p>

**Code Changes**
In general, changes required to refactor all action-related code out of `SorbetStatusBarEntry` and into a new command-like class, `showSorbetActions`, and then switch from `showInformationMessage` to `showQuickPick`.   Some additioal changes were required, however, to fix code or maintain a consistent implementation:
- `SorbetConfigPicker` was renamed `commnads/ShowSorbetConfigurationPicker` .
- Action-related code not exposing `Thenable`/`Promise` results was updated to do so which meant updating relevant callsites as well. 
- A set of constants for command Ids was introduced to simplify tracking.  
- A few private commands Ids (prefixed with `_`) were removed as every call would actually be the result of a user action, but no public command Id was changed or otherwise lost. 

### Motivation

Improved UX.

### Test plan

See included automated tests.
Manual testing of actions.  It was a non-goal to add tests at this time.


jez (approved):
Looks good to me but you'll need to manually bump the version number in the package.json file if you want to release a new version to the marketplace.

You can either do that in this PR or another PR.

damolina-stripe:
@jez I might add a few more changes in before releasing a new version to make it worth it.  Also, forgot to update the CHANGELOG so I'll do that on a latter PR (but before releasing a new version).

PR #7047, opened 2023-05-26
Change highlight loc of untyped call site

jez:
### Motivation

It's getting pretty noisy to have a huge chain of well-typed code, and then have
a single call at the end that's untyped suddenly highlight the entire receiver.

This visualization should hopefully be less noisy.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -563,7 +563,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
     auto errLoc = (funLoc.exists() && !funLoc.empty()) ? funLoc : args.callLoc();
     if (symbol == core::Symbols::untyped()) {
         auto what = core::errors::Infer::errorClassForUntyped(gs, args.locs.file, args.thisType);
-        if (auto e = gs.beginError(args.receiverLoc(), what)) {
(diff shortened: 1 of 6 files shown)


core/types/calls.cc
```diff
@@ -563,7 +563,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
     auto errLoc = (funLoc.exists() && !funLoc.empty()) ? funLoc : args.callLoc();
     if (symbol == core::Symbols::untyped()) {
         auto what = core::errors::Infer::errorClassForUntyped(gs, args.locs.file, args.thisType);
-        if (auto e = gs.beginError(args.receiverLoc(), what)) {
+        if (auto e = gs.beginError(errLoc, what)) {
```
jez, line 566: This now uses the same location that we report "this method is missing" errors on, which seems fine. In a sense, the method "is" missing (because the call was on untyped, so Sorbet couldn't find it).

jez: A counterpoint: I think this is the first time we're reporting a usage of untyped error where the highlighted region is not a complete expression.

We weren't using that invariant before, but I wanted to call out that we're now losing it.

neilparikh (approved):
You have one test you need to update, LGTM otherwise

PR #7048, opened 2023-05-29
Add an autocorrect for missing abstract method definitions.

### LINKED ISSUE #5238: Autocorrect to create method skeleton for missing abstract method
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0Aend)

```ruby
# typed: true
class Parent
  extend T::Sig
  extend T::Helpers
  abstract!

  sig {abstract.void}
  def foo; end
end

class Child < Parent
end
```

#### Observed output

```
editor.rb:11: Missing definition for abstract method Parent#foo https://srb.help/5023
    11 |class Child < Parent
        ^^^^^^^^^^^^^^^^^^^^
    editor.rb:8: defined here
     8 |  def foo; end
          ^^^^^^^
Errors: 1
```

#### Expected behavior

Sorbet should be able to insert a method in the `Child` class with an empty body that has the same sig and method arity as the abstract method.

Note that we can probably steal what LSP hover does to pretty print the `abstract` method to avoid having to reimplement the wheel of serializing a method symbol to a Ruby source string.

- - -

lerebear:
### Motivation

Closes #5238.

To help make it easier for users to implement abstract methods, this adds a "Quick Fix" suggestion for editors that fills in a scaffold of the abstract method in the implementing class.

### TODO

- [x] Test this in VS Code prior to merge

### Approach

- Moved the `prettyTypeForMethod` and `getRequestType` helpers from the `main/lsp` module to the `core/lsp` one, thereby allowing them to by phases outside of the LSP (5242ab39489a0979d55efb9216a1b0bc943d6d2c).
- Extended the `prettyTypeForMethod` helper with a `concretizeIfAbstract` option that can be used to configure the method to output a pretty-printed method definition that is suitable for the new quick fix (adc14bdcbdf5ca071966a74532d50f946742d382).
- Extended the definition validation phase to supply an autocorrect suggestion with each error that it generates for a missing abstract method definition (07887f4dc1fab4d232b6bbbbde85be29c64c6533).

### Feedback wanted!

This is my first contribution to this project, and also my first contribution to a production-quality C++ codebase, so feedback is very welcome! I have added some comments below where I would appreciate input.

Additionally, I would appreciate guidance on my outstanding `TODO` above: are there instructions for how I can test this against an editor like VS Code?

### Test plan

See included automated tests.


definition_validator/validator.cc
```diff
@@ -901,6 +907,27 @@ class ValidateWalk {
...
+                                                               abstractMethodRef.data(gs)->resultType, nullptr, true);
+
+        vector<string> reindentedMethodDefinitionLines;
+        absl::c_transform(absl::StrSplit(methodDefinition, "\n"), std::back_inserter(reindentedMethodDefinitionLines),
+                          [indent](auto &line) -> string { return absl::StrCat(fmt::format("{}  ", indent), line); });
+
+        auto reindentedMethodDefinition = absl::StrJoin(reindentedMethodDefinitionLines, "\n");
+
+        return core::AutocorrectSuggestion{
+            fmt::format("Define `{}`", abstractMethodRef.show(gs)),
```
lerebear, line 926: Given an unimplemented abstract method called `Parent#foo`, this will result in an autocorrect suggestion that reads: "Define `Parent#foo`". Is that copy appropriate?

The only real-world example of this copy that I've seen is from the TypeScript LSP, where the suggestion reads ["Implement X"](https://github.com/microsoft/TypeScript/blob/b1c2e8caadba7722719ead1dca700fbcc307838d/src/compiler/diagnosticMessages.json#L6766-L6773). I preferred "define" because the fix won't generally provide an implementation that passes type checks, and so "implement" felt misleading.

definition_validator/validator.cc
```diff
@@ -901,6 +907,27 @@ class ValidateWalk {
...
+            {core::AutocorrectSuggestion::Edit(
+                {classOrModuleDefLoc.copyEndWithZeroLength(), fmt::format("\n{}", reindentedMethodDefinition)})}};
```
lerebear, line 928: This causes the fix to be inserted as the first line of the implementing class/module. That is a bit unfortunate, because I suspect it will often cause code to be inserted above `include` and `extend` statements, as well as class methods that commonly appear as the first few lines of code in class. Would it be better to insert at the bottom the class instead? Is there an easy way to do that? Of course that will have its drawbacks too (e.g. appearing below a floating `private` modifier), but perhaps it covers the common case better.

core/lsp/helpers.cc
```diff
@@ -0,0 +1,189 @@
...
+    string flagString = "";
+    if (!flags.empty()) {
+        flagString = fmt::format("{}.", fmt::join(flags, "."));
+    }
+    string paramsString = "";
+    if (!typeAndArgNames.empty()) {
+        paramsString = fmt::format("params({}).", fmt::join(typeAndArgNames, ", "));
+    }
+
+    auto oneline = fmt::format("{} {{{}{}{}}}", sigCall, flagString, paramsString, methodReturnType);
```
lerebear, line 89: This block formatting for the `sig` (which looks like `sig {override.void}` for example) goes against the recommendation in most Ruby styleguides ([example](https://github.com/rubocop/ruby-style-guide#spaces-braces)) to add spaces around curly braces that represent an inline block (as opposed to those that represent a Hash, where opinion is more mixed). Might we consider changing this? If so, I will file an issue to track that work.

core/lsp/helpers.cc
```diff
@@ -0,0 +1,189 @@
+#include "absl/strings/str_cat.h"
+#include "absl/strings/str_join.h"
+#include "absl/strings/str_split.h"
+#include "core/core.h"
+
+using namespace std;
+
+namespace sorbet::core::lsp {
+
+core::TypePtr getResultType(const core::GlobalState &gs, const core::TypePtr &type, core::SymbolRef inWhat,
```
lerebear, line 10: The two methods that I've moved to this file are largely [duplicated in `packager/rbi_gen.cc`](https://github.com/sorbet/sorbet/blob/ca78893d96fbe7c4811f9ccb0fd9f8ddc24dfac3/packager/rbi_gen.cc#L86-L89), albeit with subtle modifications. I suspect we will want to unify these implementations at some point, but I did not do it here because it is tangential to my goal. However, might it make sense to move these to a different location in `core/` so that it is less awkward to use these helpers from code that is not related to the LSP?

lerebear:
I attempted to add another test for this that exercises the case where there are multiple unimplemented abstract methods in the derived class, but I ran into some trouble. The test case looked like this:

```ruby
# typed: true
# selective-apply-code-action: quickfix
class Parent
  extend T::Sig
  extend T::Helpers
  abstract!

  sig { abstract.void }
  def foo; end

  sig { abstract.params(baz: Integer).returns(String) }
  private def bar(baz); end

  sig { abstract.returns(Integer) }
  protected def qux; end
end

  class Child < Parent
# ^^^^^^^^^^^^^^^^^^^^ error: Missing definition for abstract method `Parent#foo`
# ^^^^^^^^^^^^^^^^^^^^ error: Missing definition for abstract method `Parent#bar`
# ^^^^^^^^^^^^^^^^^^^^ error: Missing definition for abstract method `Parent#qux`
# ^ apply-code-action: [A] Define `Parent#foo`
# ^ apply-code-action: [B] Define `Parent#bar`
# ^ apply-code-action: [C] Define `Parent#qux`
  end
```

The problem I found was that we seemingly can't specify multiple `apply-code-action` directives that apply to the same range. That appears to be because of the destructive modifications ([example](https://github.com/sorbet/sorbet/blob/ca78893d96fbe7c4811f9ccb0fd9f8ddc24dfac3/test/lsp_test_runner.cc#L303)) to the internal representation of those directives described in [this comment](https://github.com/sorbet/sorbet/blob/ca78893d96fbe7c4811f9ccb0fd9f8ddc24dfac3/test/lsp_test_runner.cc#L371-L372). Is there a way to work around that?

core/lsp/helpers.cc
```diff
@@ -123,6 +126,7 @@ string prettyDefForMethod(const core::GlobalState &gs, core::MethodRef method) {
         methodNamePrefix = "self.";
     }
     vector<string> prettyArgs;
+    auto defaultArgumentPlaceholder = concretizeIfAbstract ? "T.let(T.unsafe(nil), T.untyped)" : "…";
```
jez, line 129: It looks like `concretizeIfAbstract` is doing double-duty here.

You're using it both to decide to convert `abstract` to `override`, but also to decide to not generate the syntactically-invalid `…` character as a default.

We should probably have a separate variable for this, and IMO, this variable already exists: the `ShowOptions` struct defined in `core/ShowOptions.h`.

As you noted, the code you're writing appears very similar to the `package/rbi_gen.cc` code. That code was written to generate RBI files from types, and when it was written, we invented this `ShowOptions` struct so we would never again have to thread through a ton of special printing args through all of our printing methods (#5102).

IMO, we should add `concretizeIfAbstract` as a boolean to `ShowOptions`, and then take `ShowOptions` in this method. That's important, because separately, you have a bug in your code for `prettySigForMethod` due to the fact that it *doesn't* take `ShowOptions`:

https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Breturns%28T.attached_class%29%7D%0A%20%20def%20self.example%0A%20%20%20%20%23%20%20%20%20%20%20%5E%20hover%20here%0A%20%20%20%20new%0A%20%20end%0Aend

In this playground, if you hover over `example`, you'll see that Sorbet reveals

```ruby
sig {returns(T.attached_class (of A))}
```

for the sig. That is not valid syntax. The `... (of A)` suffix can be turned off by setting `showForRBI` to `true`, so long as `prettySigForMethod` takes `ShowOptions`, and passes it onwards to the call to `show`. I think you should gate your choice to use `…` or some syntactically-valid option in `prettyDefForMethod` on this `showForRBI` as well.

You would then at the various call sites of `pretty*ForMethod`, you would choose the options that make the most sense, and they would automatically thread all the way through all the relevant calls.

It may be worth changing that option to say something like "limitToValidSyntax" but that can be done in another change, if at all.

jez: And also, you should use my sorbet.run example here as a the start of a new test for your feature.

jez, line 129: You're using `T.let(T.unsafe(nil), T.untyped)`, which is redundant—`T.unsafe(nil)` is already `T.untyped`. On top of that, I don't even think we want to use `T.unsafe(nil)`. Sorbet autocorrects never insert `T.unsafe` unless the `--suggest-unsafe` command line option has been passed. See [Silencing errors in bulk](https://sorbet.org/docs/cli#silencing-errors-in-bulk)—the idea is that we only expose `T.unsafe` in autocorrects to those who are doing bulk codemods, where they don't have the time to correctly fix individual errors, and need something that will Just Work.

Some alternatives:

- Check if `NilClass` is a subtype of the expected type, and generate `"nil"` directly.

  If it's not, fall back to one of the options below.
  
- Translate it to `_` instead.

  Syntactically, this would represent a call to a method named, the same as if the user had typed `self._()`. This method is unlikely to exist, and so in `# typed: true` files or higher Sorbet will show a "method does not exist" error.
  
- Omit the default.

  This will generate a new type error, that the user will see after accepting the autocorrect, alerting them that they need to take an explicit action to resolve this. Unlike the previous, this will get reported even in `# typed: false` files.
  
  [example on sorbet.run](https://sorbet.run/#%23%20typed%3A%20false%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.params%28x%3A%20Integer%29.void%7D%0A%20%20def%20example%28x%3A%200%29%0A%20%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%0A%20%20sig%20%7Boverride.params%28x%3A%20Integer%29.void%7D%0A%20%20def%20example%28x%3A%29%0A%20%20end%0A%0Aend)
  
- Start with the "omit the default" option above, and combine it with using `T.unsafe(nil)` as the default only if `--suggest-unsafe` was passed.

core/lsp/helpers.h
```diff
@@ -9,7 +9,8 @@ core::TypePtr getResultType(const core::GlobalState &gs, const core::TypePtr &ty
                             core::TypePtr receiver, const core::TypeConstraint *constr);
 
 std::string prettyTypeForMethod(const core::GlobalState &gs, core::MethodRef method, const core::TypePtr &receiver,
-                                const core::TypePtr &retType, const core::TypeConstraint *constraint);
+                                const core::TypePtr &retType, const core::TypeConstraint *constraint,
+                                const bool concretizeIfAbstract = false);
```
jez, line 13: In combination with the ShowOptions change above, I think it would likely be better if we made the `options` argument not have a default. There are few enough calls to these methods that it wouldn't hurt to explicitly create a `ShowOptions` struct instance and pass it at all the call sites, just so people are aware what they're dealing with.

test/testdata/lsp/code_actions/add_skeleton_for_abstract_method/existing_class_body.rb
```diff
@@ -0,0 +1,26 @@
+# typed: true
+# selective-apply-code-action: quickfix
```
jez, line 2: It's fine to have written these tests as quickfix tests, and keep them that way. But a tip of mine:

I would have written these tests as `*.autocorrects.exp` files. For example:

https://github.com/sorbet/sorbet/blob/0dbaf2c465314556b5acecec6ec63937f2b9e836/test/testdata/resolver/t_class.rb#L1-L8

https://github.com/sorbet/sorbet/blob/0dbaf2c465314556b5acecec6ec63937f2b9e836/test/testdata/resolver/t_class.rb.autocorrects.exp#L1-L10

This style of test is easier for changes like this, because if you ever make a change to the source file, you can simply regenerate the exp file automatically, either like this:

```bash
# regenerates all exp files
tools/scripts/update_testdata_exp.sh
```

or like this

```bash
# regenerates a single exp file, using the bazel-bin/main/sorbet that you already have built
tools/scripts/update_testdata_exp.sh --no-build test/testdata/.../foo.rb
```

As it stands, if you wanted to edit any of your existing tests, you have to repeat your edit manually in the `.rbedited` files, because we haven't automated regenerating those files.

jez (changes requested):
apologies for the delay—i was not able to finish my review today

at the very least, the changes i've suggested here should give you something to work on, and i don't think any future feedback i'll leave will invalidate any of these current requests.

jez:
I will look at this tomorrow to complete my review. I have re-requested review from myself in the mean time.

lerebear:
Thanks for the feedback @jez! I hope to follow up with some fixes next week.

definition_validator/validator.cc
```diff
@@ -1,16 +1,22 @@
...
 #include "core/core.h"
 #include "core/errors/resolver.h"
+#include "core/lsp/helpers.h"
 
 #include "absl/algorithm/container.h"
 
 #include "definition_validator/variance.h"
 #include "main/sig_finder/sig_finder.h"
 
+#include <regex>
```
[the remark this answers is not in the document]
lerebear: Yes, this was a remnant of an earlier iteration. Removed in f7692e8.

definition_validator/validator.cc
```diff
@@ -901,6 +907,27 @@ class ValidateWalk {
         }
     }
 
+    core::AutocorrectSuggestion suggestMethodDefinition(const core::GlobalState &gs,
+                                                        const core::Loc classOrModuleDefLoc,
+                                                        const core::MethodRef abstractMethodRef) {
+        auto [_, indentLength] = classOrModuleDefLoc.findStartOfLine(gs);
+        string indent(indentLength, ' ');
+
+        auto methodDefinition = core::lsp::prettyTypeForMethod(gs, abstractMethodRef, nullptr,
```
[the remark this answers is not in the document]
lerebear: Fixed in c1195a9, where I went with your suggestion of "source_generator".

core/lsp/helpers.cc
lerebear: > IMO, we should add concretizeIfAbstract as a boolean to ShowOptions, and then take ShowOptions in this method.

Thanks for the guidance: I've implemented this in 0849d83. That commit also fixes the hover bug you described above.

> It may be worth changing that option to say something like "limitToValidSyntax"

Yes, I thought this was very worth it, since it was otherwise difficult to understand what `showForRBI` meant (especially now that we use that option outside of RBI generation). I opted for the name "useValidSyntax": ced85cc.

lerebear: > Start with the "omit the default" option above, and combine it with using T.unsafe(nil) as the default only if --suggest-unsafe was passed.

I implemented this alternative in 5488e0e.

core/lsp/helpers.h
lerebear: This was also fixed in 0849d83.

test/testdata/lsp/code_actions/add_skeleton_for_abstract_method/existing_class_body.rb
lerebear: Thanks for the pro-tip! I converted to the `.autocorrect.exp` format in 8c5a72c. I had to introduce a new `test/testdata` subfolder for the "definition_validator" phase. I think I did so in a way that follows existing convention, but feedback on adjustments is welcome.

lerebear:
Thank you again for the feedback @jez. I believe I've addressed all of your comments, so this should be ready for a fresh look please.

I thought it would be clearest to resolve threads that have been addressed, but of course please unresolve any that you feel need further iteration.

core/lsp/helpers.cc
lerebear: > might it make sense to move these to a different location in core/ so that it is less awkward to use these helpers from code that is not related to the LSP?

This was addressed in 55ccd46.

> I suspect we will want to unify these implementations at some point, but I did not do it here because it is tangential to my goal

This is still unaddressed, but I'm resolving this comment nonetheless since, as stated above, I do not intend to address that in the PR>

core/source_generator/source_generator.cc
```diff
@@ -126,7 +126,14 @@ string prettyDefForMethod(const core::GlobalState &gs, core::MethodRef method, c
         methodNamePrefix = "self.";
     }
     vector<string> prettyArgs;
-    auto defaultArgumentPlaceholder = options.concretizeIfAbstract ? "T.let(T.unsafe(nil), T.untyped)" : "…";
+
+    string defaultArgumentPlaceholder = "…";
+    if (options.useValidSyntax && gs.suggestUnsafe.has_value()) {
+        defaultArgumentPlaceholder = "T.unsafe(nil)";
+    } else if (options.useValidSyntax) {
+        defaultArgumentPlaceholder = "";
```
jez, line 134: I might leave a comment here about how we're knowingly generating an autocorrect that will introduce a further error, forcing the user to correct it, which we're choosing to do simply because it's the least bad of the alternatives.

definition_validator/validator.cc
jez: > Would it be better to insert at the bottom the class instead? Is there an easy way to do that?

Yes, I think we likely want to implement this.

Luckily, we do actually have enough information at this point in Sorbet, though it's not great.

At the call to `validateAbstract`, we actually have both the `ast::ClassDef` node and the `core::ClassOrModuleRef`. The `ClassOrModuleRef`'s loc only stores the declLoc, which is the truncated first line of the definition. (You can see how the declLoc of a ClassDef is computed [here](https://github.com/sorbet/sorbet/blob/23b9ea187ae26d67a13a48b738b24a9e3da14586/parser/Builder.cc#L841).).

But the ast::ClassDef also carries the full enclosing location of the class def, from the `class` keyword to the `end` keyword. To find the last line of the class def, we can:

- double check that the ClassDef starts and ends on separate lines
- double check that the ClassDef's loc actually ends with an `end` keyword. It does not necessarily have to, because various Sorbet constructs will translate into class definitions, for example `X = Struct.new(:foo)` translates to `class X < Struct ... end`, so there is no `end` keyword in the source text. To check whether there was an `end` keyword, the easiest way is to do something [like this code](https://github.com/sorbet/sorbet/blob/23b9ea187ae26d67a13a48b738b24a9e3da14586/cfg/builder/builder_walk.cc#L620-L627). (While it would be nice to store more location information on the ClassDef and not have to rely on textual source parsing like this, Sorbet is under intense memory pressure on large codebases.)
- find the line of the `end` keyword, using the methods available in Loc.h to split a loc into its file+line combinations
- do some similar manipulations as you are already doing to craft the new replacement loc

In my opinion, this is definitely worth doing, because I think that abstract modules are more common than abstract classes.

From this, we also find some more good test cases for you:

- [test 1](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20MyAbstractClass%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0AX%20%3D%20Class.new%28MyAbstractClass%29)
- [test 2](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20MyAbstractClass%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0AX%20%3D%20Class.new%28MyAbstractClass%29%20do%0A%20%20def%20bar%3B%20end%0Aend)
- [test 3](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20MyAbstractClass%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aclass%20OneLine%20%3C%20MyAbstractClass%3B%20end%0Aclass%20EmptyBody%20%3C%20MyAbstractClass%0Aend)

definition_validator/validator.cc
```diff
@@ -901,6 +905,28 @@ class ValidateWalk {
...
+        auto [_, indentLength] = classOrModuleDefLoc.findStartOfLine(gs);
+        string indent(indentLength, ' ');
+        auto showOptions = core::ShowOptions().withUseValidSyntax().withConcretizeIfAbstract();
+
+        auto methodDefinition = core::source_generator::prettyTypeForMethod(
+            gs, abstractMethodRef, nullptr, abstractMethodRef.data(gs)->resultType, nullptr, showOptions);
+
+        vector<string> reindentedMethodDefinitionLines;
+        absl::c_transform(absl::StrSplit(methodDefinition, "\n"), std::back_inserter(reindentedMethodDefinitionLines),
+                          [indent](auto &line) -> string { return absl::StrCat(fmt::format("{}  ", indent), line); });
```
jez, line 920: out of curiosity, is the StrCat redundant here? could it not also be part of the fmt::format call?

definition_validator/validator.cc
jez: "Missing definition" in the error message coupled with "define x" is fine with me.

definition_validator/validator.cc
```diff
@@ -901,6 +905,28 @@ class ValidateWalk {
...
+
+        auto reindentedMethodDefinition = absl::StrJoin(reindentedMethodDefinitionLines, "\n");
+
+        return core::AutocorrectSuggestion{
+            fmt::format("Define `{}`", abstractMethodRef.show(gs)),
```
jez, line 925: I think you likely want to show the method name here, not the symbol name.

It's the difference between `Some::Really::Long::Class::Name#foo` vs `foo`.

Also, using `abstractMethodRef` will show the name of the abstract class--what you're doing here isn't defining the abstract method, but rather defining a method in the concrete class to override the abstract one, which could be confusing, unless we simply write `foo`.

jez (commented):
This is looking pretty good. I only have a few comments now.

I think the big one is wanting to implement the better handling you mentioned for inserting abstract methods at the bottom of the class body instead of the top. If you have further questions about how that might work, feel free to ping me on the #internals channel on Sorbet's Slack.

jez:
> Additionally, I would appreciate guidance on my outstanding `TODO` above: are there instructions for how I can test this against an editor like VS Code?

The way I typically do this is that I have a random, empty Ruby folder with a Gemfile in it like this:

```ruby
source 'https://rubygems.org'

gem 'sorbet-static-and-runtime'
```

and then I make a `.vscode/settings.json` file in it like this:

```jsonc
{
    // "sorbet.highlightUntyped": true,
    "sorbet.userLspConfigs": [
        {
            "id": "stable",
            "name": "Sorbet",
            "description": "Stable Sorbet Ruby IDE features",
            "cwd": "${workspaceFolder}",
            "command": [
                "bundle",
                "exec",
                "srb",
                "typecheck",
                "--lsp"
            ]
        },
        {
            "id": "beta",
            "name": "Sorbet (Beta)",
            "description": "Beta Sorbet Ruby IDE features",
            "cwd": "${workspaceFolder}",
            "command": [
                "bundle",
                "exec",
                "srb",
                "typecheck",
                "--lsp",
                "--enable-all-beta-lsp-features"
            ]
        },
        {
            "id": "experimental",
            "name": "Sorbet (Experimental)",
            "description": "Experimental Sorbet Ruby IDE features (warning: crashy, for developers only)",
            "cwd": "${workspaceFolder}",
            "command": [
                "bundle",
                "exec",
                "srb",
                "typecheck",
                "--lsp",
                "--enable-all-experimental-lsp-features"
            ]
        },
        {
            "id": "path",
            "name": "Sorbet (path)",
            "description": "Sorbet on PATH, with experimental features",
            "cwd": "${workspaceFolder}",
            "command": [
                "sorbet",
                "--enable-all-experimental-lsp-features",
                "--debug-log-file=/tmp/sorbet-vscode.log",
                "--sleep-in-slow-path",
                "--lsp",
            ]
        }
    ]
}
```

and I use the `Sorbet: Configure` command to select the "Sorbet (path)" configuration. Since I do so much Sorbet development, I happen to always have `sorbet` on my PATH point to `~/stripe/sorbet/bazel-bin/main/sorbet`. If you don't want to put that on your path, you could replace `"sorbet"` with `"/path/to/bazel-bin/main/sorbet"` or something of your choosing. At that point, simply build Sorbet with whatever build configuration you prefer and it will use that. For small projects, I typically build with `./bazel build //main:sorbet --config=dbg` so that if I need to, I can use `lldb -p <pid>` to attach to a Sorbet running under VS Code and debug it if need be, but on a large project this will be slow (which is why i usually test on a mostly empty project).

jez:
^ Having said all that, because of your choice of implementation is entirely driven via Sorbet autocorrects, I personally would not have fired this up in VS Code to test. Testing via the `--autocorrect` flag should be sufficient, because the core Quickfix logic in Sorbet's LSP to turn autocorrects into code actions almost never needs new code to handle new autocorrects in Sorbet.

definition_validator/validator.cc
jez: One more thing to this effect: I recently came to this same conclusion while working on an unrelated autocorrect: to [re-declare a type member](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3AGeneric%0A%20%20X%20%3D%20type_template%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0Aend) required by a parent class or module.

I had tried factoring out some code to share, so that it would be easier to build that autocorrect, but I mistakenly chose to insert the code at the top of the class body instead of at the bottom:

https://github.com/sorbet/sorbet/pull/6958

I had been meaning to revisit the changes I landed in that PR and make it put the autocorrect at the bottom of the class body. Looking at that PR might be helpful at least to spark some ideas, or maybe even as a point to build the "insert before end" changes discussed above.

definition_validator/validator.cc
lerebear: Fixed in [33c914e](https://github.com/sorbet/sorbet/pull/7048/commits/33c914e1339489f38de8736bee249a19d221d2b7).

core/source_generator/source_generator.cc
lerebear: Fixed in [3fcdd32](https://github.com/sorbet/sorbet/pull/7048/commits/3fcdd32570656a4d6d964375a84c9d751305653c).

definition_validator/validator.cc
lerebear: Fixed in [0dd3397](https://github.com/sorbet/sorbet/pull/7048/commits/0dd3397c7ea6b59ea065a0bc03306d1eb6b7a729).

definition_validator/validator.cc
lerebear: I implemented the "insert at bottom" behaviour in 954d200.

Additionally, in an effort to support the common single-line stub pattern from the third test case you provided (i.e. `class Foo < Bar; end`), I changed how this autocorrect works somewhat significantly in 7231ed1. Now we provide a single autocorrect fix that defines all missing methods at once rather than providing separate fixes for each method.

This approach was inspired by the TypeScript LSP, which does something similar:

CleanShot 2023-08-14 at 10 29 50@2x

CleanShot 2023-08-14 at 10 30 41@2x
<br/>
<br/>

Ultimately, this helps with the stub pattern because it gives us a way to dodge a bug that the previous approach was subject to when there were multiple missing method definitions and the user attempted to fix them all at once using the "Apply all Sorbet fixes for file" global fix. In that scenario, because the fix for the stub requires opening the method body to add new methods (as opposed to just adding new lines within a method body), individual fixes did not compose properly. That aside, I think the new behaviour is also a more user-friendly solution to the problem.

To recap where support for each of the three test cases above now stands:

- [test 1](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20MyAbstractClass%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0AX%20%3D%20Class.new%28MyAbstractClass%29) is not supported, because [I wasn't sure how to consistently recognize all the permutations of a single-line class/module definition](https://github.com/sorbet/sorbet/blob/c5e74cc94e589f8440514b5ff6cf13c958100007/definition_validator/validator.cc#L991-L995) (as opposed to the common permutation described above). Open to feedback as to how to improve this!
- [test 2](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20MyAbstractClass%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0AX%20%3D%20Class.new%28MyAbstractClass%29%20do%0A%20%20def%20bar%3B%20end%0Aend) is fixed, with additional test coverage.
- [test 3](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20MyAbstractClass%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Aclass%20OneLine%20%3C%20MyAbstractClass%3B%20end%0Aclass%20EmptyBody%20%3C%20MyAbstractClass%0Aend) is fixed, with additional test coverage.

lerebear:
> I use the Sorbet: Configure command to select the "Sorbet (path)" configuration.
> Testing via the --autocorrect flag should be sufficient

@jez: thank you for both of those pieces of advice 🙌🏿 . They were each useful as I iterated on this further.

I believe I have addressed all outstanding comments, so this should be ready for a fresh look please. As before, I've resolved the threads that I think are truly complete, but left unresolved the ones that I think may warrant further discussion.

definition_validator/validator.cc
lerebear: I wanted to comment on two other things that I think could use improvement here:

1. When several methods are missing in a class, the methods appear in the error message, and are also inserted into the source text, in an arbitrary order (see [this example](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%0A%20%20sig%20%7B%20abstract.void%20%7D%0A%20%20def%20foo%3B%20end%0A%0A%20%20sig%20%7B%20abstract.void%20%7D%0A%20%20def%20bar%3B%20end%0A%0A%20%20sig%20%7B%20abstract.void%20%7D%0A%20%20def%20qux%3B%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%3B%20end), where the methods are declared from top to bottom as `foo`, `bar`, `qux`, but are listed in the error message as `foo`, `qux`, `bar`). This appears to be due to the fact that the [`members` method](https://github.com/lerebear/sorbet/blob/c5e74cc94e589f8440514b5ff6cf13c958100007/core/Symbols.h#L603), which is used [here](https://github.com/lerebear/sorbet/blob/c5e74cc94e589f8440514b5ff6cf13c958100007/definition_validator/validator.cc#L870), returns an unordered collection.
2. The error message contains some lines that describe the autocorrect edits which I think are unnecessary (see [this image](https://github.com/sorbet/sorbet/assets/776168/1af1d373-ea4e-4e43-a818-2422c334fcec)). I'm not sure why [we add those lines](https://github.com/lerebear/sorbet/blob/c5e74cc94e589f8440514b5ff6cf13c958100007/core/Error.cc#L188-L193), but they don't seem well suited to this particular context. The TypeScript error pictured above, which omits those lines, is easier to digest in my opinion.

Open to feedback as to whether or not either of those should be addressed immediately.

jez: > Open to feedback as to how to improve this!

The fact that you have detected this case and stopped short of suggesting an autocorrect which would do something like generate a syntactically invalid file after applying the autocorrect is good enough for me. It will always be possible to revisit that if/when there's a better idea for how to structure the autocorrect.

> I'm not sure why we add those lines

That's partly there's one interface for printing errors to the console and reporting errors to LSP.

In the command line interface to Sorbet, we need some way to tell the user what autocorrect they can expect to have happen when they re-run with the `-a` flag.

In LSP, I agree that those errors can be seen as clutter and redundant with both the `(fix available)` in the header and the code action title ("Define inherited abstract methods").

I'm happy to workshop how we present errors, but for the sake of reducing this PR's scope, I think I'd like to keep what currently exists. Once this PR lands, feel free to @mention me in the #internals channel on the Sorbet slack (https://sorbet.org/slack) and we can hash out some ideas for changes to make to how errors like these are presented.

> in an arbitrary order

Can we sort by their string names?

In general, we want all the output from Sorbet to be in a stable order, so that our snapshot tests are not flaky.

core/lsp/helpers.cc
jez: > Might we consider changing this?

Yes, this is a case were the `sig {override.void}` syntax was previously required by Stripe's internal style guide. But about a year ago we changed our style guide to prefer `sig { override.void }` to conform with the rest of the ruby community.

I already changed many of these:

https://github.com/sorbet/sorbet/pull/6572

I guess I missed this one.

definition_validator/validator.cc
lerebear: > Once this PR lands, feel free to @mention me in the #internals channel on the Sorbet slack (https://sorbet.org/slack) and we can hash out some ideas for changes to make to how errors like these are presented.

Thanks, and will do 👍🏿 

> Can we sort by their string names?

Certainly: will fix.

jez:
@lerebear heads up that you have a handful of test failures.

Most of these are expected, because you changed the format of the error output.

You have two choices:

1. You can edit the tests to conform to the new language you've changed the errors to emit.
2. You can special case the `n = 1` case to emit an error whose header is the same as the old header would have been (namely: to include the name of the one abstract method which is missing). Doing this may not fix all the test failures, but it will at least cut down on the number of tests that would need to be changed. Separately it may make the error message slightly nicer, as it would still mention the missing abstract method name in the header, like before.

Up to you which you'd prefer to do.

core/lsp/helpers.cc
lerebear: Thanks for the context. I'll wait until after this lands to file an issue (just so that I can reference code that has actually been integrated). I'll also attempt to address that issue myself as a quick follow-up to this PR.

lerebear:
> Separately it may make the error message slightly nicer, as it would still mention the missing abstract method name in the header, like before.

Agreed: will fix.

> @lerebear heads up that you have a handful of test failures.

Sorry to have re-requested review without addressing these beforehand. I will run the full the test suite locally going forward.

definition_validator/validator.cc
```diff
@@ -905,34 +905,30 @@ class ValidateWalk {
...
 
-    core::AutocorrectSuggestion suggestMethodDefinition(const core::GlobalState &gs,
-                                                        const core::Loc classOrModuleDefLoc,
-                                                        const core::MethodRef abstractMethodRef) {
-        auto [_, indentLength] = classOrModuleDefLoc.findStartOfLine(gs);
-        string indent(indentLength, ' ');
+    core::AutocorrectSuggestion::Edit defineInheritedAbstractMethod(const core::GlobalState &gs,
+                                                                    const core::MethodRef abstractMethodRef,
+                                                                    const core::Loc insertAt, const string format,
+                                                                    const string classOrModuleIndent) {
```
jez, line 911: Is it intentional that we're taking these two `string` arguments by copy?

To avoid copying them, you can either take them by `const string &` or by `const string_view` (which `string` can automatically convert to).

definition_validator/validator.cc
```diff
@@ -942,17 +938,68 @@ class ValidateWalk {
...
-                    e.addAutocorrect(suggestMethodDefinition(gs, loc, proto));
+            auto concreteMethodRef = sym.data(gs)->findConcreteMethodTransitive(gs, proto.data(gs)->name);
+            if (concreteMethodRef.exists()) {
+                continue;
+            }
+
+            missingMethods.emplace_back(proto);
+        }
+
+        if (!missingMethods.empty()) {
```
jez, line 955: Very minor: it might fit the style of this method more if you convert this to

```cpp
if (missingMethods.empty()) {
    return;
}

// ...
```

definition_validator/validator.cc
```diff
@@ -942,17 +938,68 @@ class ValidateWalk {
...
+            if (auto e = gs.beginError(classOrModuleDeclaredAt, core::errors::Resolver::BadAbstractMethod)) {
+                auto pluralization = (missingMethods.size() > 1 ? "s" : "");
+                e.setHeader("Missing definition{} for abstract method{}", pluralization, pluralization);
+
+                auto classOrModuleEndsAt =
+                    core::Loc{classOrModuleDeclaredAt.file(), classDef.loc.copyEndWithZeroLength()};
```
jez, line 961: `LocOffset` are best understood relative to the file in the enclosing `ctx`. While the way you've written it might work, it's a bit weird to see it written this way. A better way to do this would be to change `validateAbstract` to take a `core::Context ctx` instead of a `core::GlobalState &gs`, and then do this here:

```suggestion
                auto classOrModuleEndsAt = ctx.locAt(classDef.loc.copyEndWithZeroLength());
```

Luckily you are already passing in a `core::Context` at the call sites of this method (it's just that right now, the compiler is implicitly converting them to a `GlobalState &`.

definition_validator/validator.cc
```diff
@@ -942,17 +938,68 @@ class ValidateWalk {
...
+                auto hasSingleLineDefinition =
+                    classOrModuleDeclaredAt.position(gs).first.line == classOrModuleEndsAt.position(gs).second.line;
+
+                auto endRange = classOrModuleDeclaredAt.copyEndWithZeroLength().join(classOrModuleEndsAt);
+                string endSuffix(endRange.source(gs).value_or(""));
+                regex emptyBodyRe("^\\s*;\\s+end\\s*$");
+                auto hasEmptyBody = regex_match(endSuffix, emptyBodyRe);
```
jez, line 968: Is it not enough to skip the regex stuff and instead check `classDef->rhs.empty()`?

jez (approved):
This looks great. I only have quibbles at this point.

jez:
> Sorry to have re-requested review without addressing these beforehand

This is no problem, just wanted to flag with you to make sure.

I am generally fine reviewing code that doesn't pass tests as long as it's directionally accurate.

definition_validator/validator.cc
jez: Ah, it looks like at this point we've created a fake `<static-init>` method in the method body:

```
(lldb) p (void)(printf("%s\n", tree.toString(ctx).c_str()))
class ::Child<<C Child>> < (::Parent)
  def self.<static-init>(<blk>)
    <emptyTree>
  end
end
```

which means that this code would need to be something like

```cpp
auto hasEmptyBody = classDef.rhs.empty();
if (classDef.rhs.size() == 1) {
    if (auto *mdef = ast::cast_tree<ast::MethodDef>(classDef.rhs[0])) {
        hasEmptyBody = ast::isa_tree<ast::EmptyTree>(mdef->rhs);
    }
}
```

which is more code but less reliant on regex.

definition_validator/validator.cc
lerebear: > Can we sort by their string names?

Fixed in f0996c9.

definition_validator/validator.cc
lerebear: Not intentional unfortunately: I'm still learning to be sensitive to these sorts of concerns. This type of scrutiny is very appreciated! Fixed in 0e32270.

definition_validator/validator.cc
lerebear: Fixed in 3d526ec.

definition_validator/validator.cc
lerebear: Fixed in a000c6c.

definition_validator/validator.cc
lerebear: I must admit that I don't yet understand what we expect the AST to look like at various stages, but your suggestion passed all of the tests as written, so I applied it in 45418ae.

lerebear:
Thanks for the latest round of feedback @jez. I've addressed all comments, so this should be ready for another look. 

Hopefully this is pretty close at this point, so in case this is approved without further feedback, please let me know if the convention is for maintainers to merge pull requests, or for contributors do so themselves.

jez:
> please let me know if the convention is for maintainers to merge pull requests, or for contributors do so themselves

convention is for maintainers to merge. only stripe employees have the ability to merge PRs at this point.

lerebear:
> Separately it may make the error message slightly nicer, as it would still mention the missing abstract method name in the header, like before.

One note on this: as I was working on this fix, it revealed that we were sometimes generating duplicate errors (and, as a result, duplicate autocorrect suggestions) for situations in which the same method was missing across multiple classes/modules in the inheritance chain. I fixed that problem at the same time as I improved the error message in https://github.com/sorbet/sorbet/pull/7048/commits/bb0f8fe5ae186fbe31e43691be0049bc28612f9e.

test/testdata/lsp/hover.rb
```diff
@@ -87,13 +87,6 @@ def tests_return_markdown
     # ^^^^^^^^^^^^^^^^^^^^^ hover-line: 6 ---
     # ^^^^^^^^^^^^^^^^^^^^^ hover-line: 8 Tests return markdown output
   end
-
-  sig {returns(T.attached_class)}
-  def self.factory
-    #      ^ hover: sig {returns(T.attached_class)}
-    #      ^ hover: def self.factory; end
-    new
-  end
```
jez: Should we keep this test here, but change the `hover:` assertion to show the correct behavior? Maybe you deleted it because that's already tested somewhere else?

lerebear: [The commit in which this was removed](https://github.com/sorbet/sorbet/pull/7048/commits/c916874ffd3e23f8cabe2ea4950de94fa09054e3) actually restores the hover behaviour that is currently on `master`; I had mistakenly changed the options that hover was using to show the method in an [earlier commit](https://github.com/sorbet/sorbet/pull/7048/commits/0849d83d4d1f8c316df051fdd32e4392691a1fa4) (specifically [this line](https://github.com/sorbet/sorbet/blob/0849d83d4d1f8c316df051fdd32e4392691a1fa4/main/lsp/requests/hover.cc#L71) incorrectly added `withShowForRBI()`[^1]).

But perhaps you're suggesting we actually change how hover works?

[^1]: Although `withShowForRBI()` was since renamed to `useValidSyntax` ced85cc1f.

lerebear: Oh, sorry, I think understand what you mean now: keep the addition I made to the test and ensure it has the correct hover assertion. Will fix!

jez:

→ https://go/builds/bui_OWAd5Y2FpZYlAe
→ https://go/builds/bui_OWAd0F3jFf6dki
→ https://go/builds/bui_OWAe2ajAqDDJbI

jez:
This is amazing, thanks for all the work. I've kicked off one final test run, to test this change on Stripe's codebase. Based on the contents of the change, I can't imagine this causing anything to break, but since it's a lot I figure I may as well.

I just wanted to take a moment to say what a wonderful PR this was to review. Thanks for all your patience, careful comments, and attention to detail--you picked a hard task to work on and executed it flawlessly, especially considering this was your first contribution to the type checker. Thanks!

test/testdata/lsp/hover.rb
lerebear: Fixed in 79fe4ac. Note that this asserts that invalid syntax is acceptable on hover, which I believe is what we want.

lerebear:
> I just wanted to take a moment to say what a wonderful PR this was to review.

Thank you! Likewise, I'm *very* grateful for your patience in coaching me through this. I'm pretty interested in the LSP, so I hope to continue working on enhancements from the [autocorrect milestone](https://github.com/sorbet/sorbet/milestone/15). Hopefully those fixes begin to land more easily as I build experience 🤞🏿 .

PR #7050, opened 2023-05-30
Remove runtime `lodash` dependency

damolina-stripe:
Remove runtime `lodash` dependency as it was brought in to perform a simple `string[]` deep-equality comparison.
- Removing this dependency reduces the size of the extension .VSIX package from **3.62MB to 2.59MB**.
  - Note that `sinon` (used in tests) still depends on `lodash`, so the `yarn.lock` file does not change significantly.

### Motivation

The cost-benefit of the library does not add up: 40% increase on the VSIX file to use a single method that is easily implemented with stock JS.  Additionally, its presence causes confusion as at least one callsite being removed on this PR used `lodash`'s `isEqual` directly instead of using the more correct class method (that was implemented wiht it!). 

### Test plan

See included automated tests.


damolina-stripe:
Superseded by non-fork PR: https://github.com/sorbet/sorbet/pull/7056

PR #7051, opened 2023-05-30
Settings should use `mardownDescription` when using markdown.

damolina-stripe:
Settings should use the `mardownDescription` property when using markdown text (package.json), instead of `description`, otherwise text is not formatted.

**After**
<p align="center">
image
</p>

**Before**
<p align="center">
image
</p>

### Test plan

See included automated tests.


ilyailya (approved):
This is great, thanks!

PR #7052, opened 2023-05-31
Refactor extension logic into purpose-specific components (part 2)

damolina-stripe:
This is a follow up to https://github.com/sorbet/sorbet/pull/7046 focused on the "status provider" component:
- **Fix**: `filterUpdatesFromOldClients` refactoring was flawed causing status change events to be ignored (e.g. status bar would not reflect current  `Typing` state). The new logic removes the method altogether and simplifies the flow significantly, making it far easier to debug.
- Move `serverStartus` and `lastError`  props from `SorbetStatusBarEntry` to `SorbetStatusProvider`, as the former is intended to be a view-only component, and the latter should be the source of status information.
- Reduce refresh rate for the status bar spinner from 100ms to 250ms (this is not an interactive component, so high refresh rate is not needed and it is wasteful).

### Motivation

Fixes and code maintainability.

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/src/extension.ts
+++ b/vscode_extension/src/extension.ts
@@ -22,15 +22,15 @@ export function activate(context: ExtensionContext) {
   context.subscriptions.push(
     sorbetExtensionContext,
     sorbetExtensionContext.config.onLspConfigChange(
-      ({ oldLspConfig, newLspConfig }) => {
+      async ({ oldLspConfig, newLspConfig }) => {
         const { statusProvider } = sorbetExtensionContext;
         if (oldLspConfig && newLspConfig) {
           // Something about the config changed, so restart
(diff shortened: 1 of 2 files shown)


jez:
@damolina-stripe heads up that you're still committing against your fork, which means that CI has not been kicked off. If you close this PR and re-create it using a branch pushed directly to the sorbet/sorbet repo, your CI will start automatically.

damolina-stripe:
> @damolina-stripe heads up that you're still committing against your fork, which means that CI has not been kicked off. If you close this PR and re-create it using a branch pushed directly to the sorbet/sorbet repo, your CI will start automatically.

ACK - Moved to https://github.com/sorbet/sorbet/pull/7053

PR #7055, opened 2023-05-31
Align telemetry code with internal implementation

damolina-stripe:
Align code used to capture telemetry events with implementations used inside Stripe.
- Intention in https://github.com/sorbet/sorbet/pull/7046 was to create a "telemetry client" , but the change was deferred to keep that PR smaller, this PR completes that flow. That is the reason why the context's `emitCountMetric` and `emitTimingMetric` existed and are finally removed.
- New implementation removes a risky recursive calls in `emit*` methods.
- `veneur` name is an old implementation detail, so renaming file to `metricsClient` (as used internally).

### Test plan

See included automated tests.


damolina-stripe:
> Does this change the metrics API? The extension metrics API is a public API that other orgs and companies can use to get metrics into their own system too

All changes are internal to the extension's consumption of the `API` and `MetricsEmitter` interfaces.  These expected interfaces are not modified in any shape or form.

jez (approved):

PR #7056, opened 2023-05-31
Remove runtime lodash dependency

damolina-stripe:
Remove runtime `lodash` dependency as it was brought in to perform a simple `string[]` deep-equality comparison.
- Removing this dependency reduces the size of the extension .VSIX package from **3.62MB to 2.59MB**.
  - Note that `sinon` (used in tests) still depends on `lodash`, so the `yarn.lock` file does not change significantly.

### Motivation
 The cost-benefit of the library does not add up: 40% increase on the VSIX file to use a single method that is easily implemented with stock JS.  Additionally, its presence causes confusion as at least one callsite being removed on this PR used `lodash`'s `isEqual` directly instead of using the more correct class method (that was implemented wiht it!).

See included automated tests.

### DIFF
--- a/vscode_extension/src/config.ts
+++ b/vscode_extension/src/config.ts
@@ -1,17 +1,23 @@
-import { isEqual } from "lodash";
 import {
-  workspace,
-  Event,
-  EventEmitter,
   ConfigurationChangeEvent,
   Disposable,
+  Event,
+  EventEmitter,
   ExtensionContext,
-  Memento,
   FileSystemWatcher,
+  Memento,
   Uri,
+  workspace,
   WorkspaceFolder,
 } from "vscode";
 import * as fs from "fs";
+
+/**
+ * Compare two `string` arrays for deep, in-order equality.
+ */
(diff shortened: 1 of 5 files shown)


vscode_extension/src/config.ts
```diff
@@ -1,18 +1,24 @@
...
   Uri,
+  workspace,
   WorkspaceFolder,
 } from "vscode";
 import * as fs from "fs";
 
+/**
+ * Compare two `string` arrays for deep, in-order equality.
+ */
+export function deepEqual(a: ReadonlyArray<string>, b: ReadonlyArray<string>) {
```
froydnj, line 18: Nit: Does this need to be exported?

froydnj (approved):
:+1: for removing unneeded dependencies.

neilparikh (approved):

vscode_extension/src/config.ts
damolina-stripe: Not really - not at least until I get to add per-file tests  😞

PR #7061, opened 2023-06-01
Fix Command Enablement Rules

damolina-stripe:
Fix Command Enablement Rules:
- `sorbet.copySymbolToClipboard` and `sorbet.rename` require a Ruby file to be open in an editor to be executed.
   - If there is no editor, they do nothing so no point on them being enabled.
   - Today actually they fail if the extension has not been loaded because they are not part of the `activationEvents` list. Adding them to the list  is not the right fix (in fact, after VSCode 1.75, commands should not even be there):
        ```
        command 'sorbet.copySymbolToClipboard' not found
        ```
- `sorbet.showOutput` is disabled  in the Command Palette if current file is not Ruby but this is inconsistent with the behavior of other non-editor dependent commands, like `sorbet.configure`. It also breaks the scenario where current editor is a non-Ruby file (e.g. a README file) but the project does contain Ruby files so looking at Sorbet output is still needed.
- `sorbet.toggleHighlightUntyped` fails when no workspace is open (`_startSorbetProcess` in `LanguageClient`  assumes there is `activeLspConfig` all the time - this can be improved in a separate PR):
    ```
    Activating extension 'sorbet.sorbet-vscode-extension' failed: 
    Cannot read properties of null (reading 'command').
    ```

Misc:
- Sort command definitions alphabetically by commandId so it is easier to maintain the list.

### Motivation

Commands failing or missing in some scenarios.

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -54,6 +54,12 @@
         "category": "Sorbet"
       },
       {
+        "command": "sorbet.copySymbolToClipboard",
+        "title": "Copy Symbol to Clipboard",
+        "category": "Sorbet",
+        "enablement": "editorLangId == ruby"
+      },
+      {
         "command": "sorbet.disable",
         "title": "Disable",
         "category": "Sorbet"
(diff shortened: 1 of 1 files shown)


froydnj (approved):
I don't know if there's really a good way to enable `sorbet.showOutput` properly (e.g. you wouldn't want to use that in Go codebases), but I think the consistency argument is the best.

damolina-stripe:
> I don't know if there's really a good way to enable `sorbet.showOutput` properly (e.g. you wouldn't want to use that in Go codebases), but I think the consistency argument is the best.

@froydnj  - we could change the behavior so `extension.ts` sets a `sorbetActive` context value,  and then update the `enablement` rules for all Sorbet commands to be `enablement: sorbetActive`.  With that all commands would be disabled unless `extension.ts` has been run which would only happen during activation and that only happens when `"workspaceContains:sorbet/*"` triggers.  WDYT?

PR #7063, opened 2023-06-01
Naming Style: No-underscore, camel-case filenames

damolina-stripe:
Extension code uses a `_` prefix to denote private fields and methods but Typescript naming guidelines discourage that, e.g.  [Google](https://google.github.io/styleguide/tsguide.html#identifiers-underscore-prefix-suffix ), [TypeScript internal](https://github.com/microsoft/TypeScript/wiki/Coding-guidelines).

  - This can be revisited but this change defines a common baseline and standardizes the codebase.
  - Getters/setters are usually the one case where `_` helps with naming conflicts but for now adopting [Google](https://google.github.io/styleguide/tsguide.html#properties-used-outside-of-class-lexical-scope)'s recommendation of using a full word prefix like `wrapped` instead (it seems that prefix might be useful during code reviews more than `internal`). This is open to discussion, however, as [Typescript's own handbook](https://www.typescriptlang.org/docs/handbook/2/classes.html#getters--setters) uses an underscore in this case.
   - Enable `no-underscore-dangle` eslint rule.

Misc:
 - Stopped exporting  `deepEqual` from config.ts (feedback from previous PR so fixing as the file is being modified on this PR).
 - Removed the last two cases of default exports. See https://google.github.io/styleguide/tsguide.html#exports
 - Updated last cases of files not using camelCase names.

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/src/config.ts
+++ b/vscode_extension/src/config.ts
@@ -15,7 +15,7 @@ import * as fs from "fs";
 /**
  * Compare two `string` arrays for deep, in-order equality.
  */
-export function deepEqual(a: ReadonlyArray<string>, b: ReadonlyArray<string>) {
+function deepEqual(a: ReadonlyArray<string>, b: ReadonlyArray<string>) {
   return a.length === b.length && a.every((itemA, index) => itemA === b[index]);
 }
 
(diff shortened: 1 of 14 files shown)


vscode_extension/src/config.ts
```diff
@@ -15,7 +15,7 @@ import * as fs from "fs";
 /**
  * Compare two `string` arrays for deep, in-order equality.
  */
-export function deepEqual(a: ReadonlyArray<string>, b: ReadonlyArray<string>) {
+function deepEqual(a: ReadonlyArray<string>, b: ReadonlyArray<string>) {
```
damolina-stripe, line 18: FYI @froydnj  - removing as promised in the other PR.

ilyailya (approved):

PR #7068, opened 2023-06-05
Update package dependencies

damolina-stripe:
Update package dependencies using `yarn upgrade`. 
- This only performs a minor/compatible version update (bug/security fixes).

**Note**
Current target VSCode is too old meanining these dev-dep updates cannot be applied (they would solve some warnings but are not automatic to `yarn upgrade`):
- `vsce` was renamed to `@types/vsce`
- `vscode-test` was renamed to `@vscode/test-electron`

### Motivation

Update packages to recent versions.

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -299,7 +299,6 @@
     "vscode-install": "yarn generate-package && code --force --install-extension sorbet.vsix"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


jez:
@damolina-stripe Does this include updates to the package that we use for the LSP protocol? I think we tried updating that one in the past and it caused problems.

PR #7071, opened 2023-06-06
updated_prop: `DeprecatedNumeric` → `Numeric`

jez:
These two constants are aliases of each other in Stripe's codebase.

Using `Numeric` is deprecated in props in models in Stripe's codebase.
This is normally checked by Rubocop lint rules, but `updated_prop` hides
the fact that a prop is using `DeprecatedNumeric`.

So whether Sorbet treats this type as `Numeric` or `DeprecatedNumeric`
doesn't really matter. For simplicity's sake, let's just say `Numeric`,
so that we generate fewer constant literals. `updated_prop` isn't used
widely enough for this to make a measurable difference, but at least
there's Stripe-specific code in Sorbet.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_O27P41uouG1EV5
→ https://go/builds/bui_O27PXdRPZ3jMXX
→ https://go/builds/bui_O27PqXclAUZoA4

neilparikh (approved):

PR #7072, opened 2023-06-06
Remove mentions of Chalk::ODM from Sorbet

jez:
### Motivation

There were two mentions of `Chalk::ODM` in Sorbet's implementation:

- `Chalk::ODM::DeprecatedNumeric`, removed as of #7071
- `Chalk::ODM::Document`

It turns out that we got essentially nothing from this change.

Before, if you had a `Chalk::ODM::Document` class, we would generate a
getter for the props in that document that would do

```ruby
def foo
  self.instance_variable_get(:@foo)
end
```

Otherwise, if it was an unknown class (including any model class), Sorbet
would fall back to the gnarly `self.decorator_class.prop_get` logic.

However, we flipped that so that it's only enabled in `# compiled: true`
files. Realistically, if we want the Sorbet Compiler to speed up props, we
want it to do so for models, not just documents. So there's a larger
problem here, and I don't think we need to keep around references of
`Chalk::ODM` just to realize a (non-existing) win making prop reads faster
only for Chalk::ODM::Document.

### Test plan

See included automated tests.


jez:
* master
  * **PR #7072** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/7072" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/7072?utm_source=stack-comment).

ilyailya (approved):
LGTM!
Should we run it over pay-server before merging?

jez:

→ https://go/builds/bui_O28gvdUfV1tFI8
→ https://go/builds/bui_O28h7ChMJOdo1b
→ https://go/builds/bui_O28haad4WRulzW

PR #7079, opened 2023-06-08
Allow code that uses `new` as a type factory to be marked abstract

paracycle:
This PR changes the mechanics of the `new` method installed by `abstract!` to check the return type of `super` call instead of checking the type of the receiver to detect if an abstract instance is being created.

### Motivation

Some Ruby code uses the `new` method as a type factory to instantiate a type that is related to but not the same as the type of the receiver. That means, those classes cannot be marked as `abstract!`, since doing that would prevent the use of `new` on the class immediately.

This is specifically how Active Record handles STI, which is now broken. Consider the following example:

```ruby
  class Market < ActiveRecord::Base
    has_many :regions
  end

  class Region < ActiveRecord::Base
    abstract!
  end

  class Country < Region
  end

  class Province < Region
  end
```

Before Sorbet started patching `new` (pre #6888):
```ruby
[1] pry(main)> market.regions.build
RuntimeError: Region is declared as abstract; it cannot be instantiated
[2] pry(main)> market.regions.build(type: Country)
=> #<Country:0x00007f97c3b1a3c0 ... >
```

After Sorbet started patching `new` (post #6888):
```ruby
[1] pry(main)> market.regions.build
RuntimeError: Region is declared as abstract; it cannot be instantiated
[2] pry(main)> market.regions.build(type: Country)
RuntimeError: Region is declared as abstract; it cannot be instantiated
```

The difference is due to how [Active Record redefines `new`](https://github.com/rails/rails/blob/ef04fbb3b256beececfa44c47c4ec93ac6945e59/activerecord/lib/active_record/inheritance.rb#L56-L78) to return an instance of a subtype of the receiver as specified by the `type` argument. Sorbet's patching of `new` prevents that from happening, so the `type` argument is ignored and an exception is raised instead.

However, what we really want to prevent is for `new` to return an instance of the abstract type, not prevent it from returning a non-abstract type since the receiver was abstract. So this commit changes the behaviour of the `new` method `abstract!` installs to call `super` first, and then check if the returned instance is an instance of the abstract type. If it is, it raises an exception, otherwise it returns the instance.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/abstract/declare.rb
+++ b/gems/sorbet-runtime/lib/types/private/abstract/declare.rb
@@ -35,10 +35,11 @@ module T::Private::Abstract::Declare
       # define_method because of the guard above
 
       mod.send(:define_singleton_method, :new) do |*args, &blk|
-        if T.unsafe(self) == mod
-          raise "#{mod} is declared as abstract; it cannot be instantiated"
+        super(*args, &blk).tap do |result|
+          if result.instance_of?(mod)
+            raise "#{mod} is declared as abstract; it cannot be instantiated"
+          end
         end
-        super(*args, &blk)
       end
 
       # Ruby doesn not emit "method redefined" warnings for aliased methods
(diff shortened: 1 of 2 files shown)


jez (approved):
Clever!

PR #7080, opened 2023-06-08
Add `--did-you-mean=false` option

jez:
### Motivation

Sometimes a change to make something go from untyped -> typed introduces new
errors. Many of these errors show up as "Method does not exist" errors. But then
a case like this can happen:

- The new type is actually too broad. The type went from `T.untyped` to
  `Parent`. Sorbet reports an error like "No method `foo` on `Parent`"

- But in this particular case, the method `foo` does actually exist at runtime,
  because at runtime the type is always `Child`.

- There is a similar method on the parent, called `foot`, and Sorbet thinks that
  the call to `foo` is a typo for `foot`, so it inserts a "did you mean"
  autocorrect.

These situations are problematic for large codemods. In those cases, it's
preferred to simply ignore the did you mean suggestions.

### Test plan

See included automated tests.

### DIFF
--- a/core/Error.cc
+++ b/core/Error.cc
@@ -196,6 +196,13 @@ void ErrorBuilder::addAutocorrect(AutocorrectSuggestion &&autocorrect) {
     this->autocorrects.emplace_back(move(autocorrect));
 }
 
+void ErrorBuilder::didYouMean(const std::string &replacement, Loc loc) {
+    std::string formatted = fmt::format("Replace with `{}`", replacement);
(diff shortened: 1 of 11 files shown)


jez:

→ https://go/builds/bui_O2xkgUakAEseLy
→ https://go/builds/bui_O2xk2IFIh9GRtG
→ https://go/builds/bui_O2xkhAqoJbaK3W

neilparikh (approved):

PR #7084, opened 2023-06-12
Docs: fix link in error reference

ghiculescu:
Link points to https://sorbet.org/docs/generics#bounds-on-type-member-s-and-type-template-s-fixed-upper-lower

It should point to https://sorbet.org/docs/generics#bounds-on-type_members-and-type_templates-fixed-upper-lower


jez (approved):

jez:
Thanks!

PR #7085, opened 2023-06-12
Move Sorbet to mimalloc

maruth-stripe:
Moving to [mimalloc](https://github.com/microsoft/mimalloc) from jemalloc. 

### Motivation

Empirically, this saved ~3.5% memory while typechecking pay-server. 
### Test plan

See included automated tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -84,7 +84,7 @@ build:dbg-linux --config=dbg --platforms=@//tools/platforms:linux_x86_64
 ##
 # release version: optimized, with debug symbols and version information
 build:release-common --define release=true
-build:release-common --define jemalloc=true
+build:release-common --define mimalloc=true
 build:release-common --compilation_mode=opt
 build:release-common --config=backtracesymbols
 build:release-common --config=static-libs
--- a/.bazelrc
+++ b/.bazelrc
@@ -130,11 +130,13 @@ build:release-debug-mac --config=release-debug-common
 build:release-sanitized-linux --copt=-O2
 build:release-sanitized-linux --config=sanitize-linux
 build:release-sanitized-linux --define jemalloc=false
+build:release-sanitized-linux --define mimalloc=false
 build:release-sanitized-linux --config=release-debug-common
 
 build:release-sanitized-mac --copt=-O2
 build:release-sanitized-mac --config=sanitize-mac
 build:release-sanitized-mac --define jemalloc=false
+build:release-sanitized-mac --define mimalloc=false
 build:release-sanitized-linux --config=release-debug-common
 
 ##
--- a/.bazelrc
+++ b/.bazelrc
@@ -130,13 +130,11 @@ build:release-debug-mac --config=release-debug-common
 build:release-sanitized-linux --copt=-O2
 build:release-sanitized-linux --config=sanitize-linux
 build:release-sanitized-linux --define jemalloc=false
-build:release-sanitized-linux --define mimalloc=false
 build:release-sanitized-linux --config=release-debug-common
 
 build:release-sanitized-mac --copt=-O2
 build:release-sanitized-mac --config=sanitize-mac
 build:release-sanitized-mac --define jemalloc=false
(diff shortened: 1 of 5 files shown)


main/BUILD
```diff
@@ -25,7 +34,8 @@ cc_binary(
         "//conditions:default": 1,
     }),
     malloc = select({
-        ":jemalloc": "@jemalloc//:jemalloc",
+        # ":jemalloc": "@jemalloc//:jemalloc",
```
jez, line 37: Should we have both of these, so that you can use the config setting to pick whichever one you prefer from the bazel build command line?

third_party/mimalloc.BUILD
jez, line 1: It looks like you copied this from the jemalloc.BUILD file.

Do we need all of this? Some of it might have been hacks to get jemalloc to build, that aren't required when getting mimalloc to build. I'd love if we could see what parts of this are and are not required.

Is there an upstream bazel BUILD file?

third_party/mimalloc.BUILD
```diff
@@ -0,0 +1,101 @@
...
+    popd > /dev/null
+    mv $$(dirname $(location CMakeLists.txt))/out/release/libmimalloc.a $(location lib/libmimalloc.a)
+  else
+    printf '%s\n' "$${compile_output}"
+    exit 1
+  fi
+
+"""
+
+genrule(
```
jez, line 67: In particular, does this need to be a genrule? That seems like a weird way to build this—most of the time when we build a cc_library, we just give all the sources to the library directly, instead of first assembling an archive file.

.bazelrc
```diff
@@ -130,11 +130,13 @@ build:release-debug-mac --config=release-debug-common
 build:release-sanitized-linux --copt=-O2
 build:release-sanitized-linux --config=sanitize-linux
 build:release-sanitized-linux --define jemalloc=false
+build:release-sanitized-linux --define mimalloc=false
 build:release-sanitized-linux --config=release-debug-common
 
 build:release-sanitized-mac --copt=-O2
 build:release-sanitized-mac --config=sanitize-mac
 build:release-sanitized-mac --define jemalloc=false
+build:release-sanitized-mac --define mimalloc=false
```
jez, line 139: Can we look at the blame for the lines that added jemalloc=false and see if they mentioned why they were marked that way? Is it possible that we could actually keep mimalloc turned on when we run sanitized builds now?

maruth-stripe: ack

third_party/mimalloc.BUILD
maruth-stripe: mimalloc is a CMake project unfortunately, so no Bazel file from what I can tell

third_party/externals.bzl
```diff
@@ -346,6 +354,14 @@ def register_sorbet_dependencies():
         strip_prefix = "shellcheck-v{}".format(shellcheck_version),
     )
 
+    # Needed to build CMake projects
+    http_archive(
+        name = "rules_foreign_cc",
+        urls = _github_public_urls("bazelbuild/rules_foreign_cc/archive/refs/heads/main.zip"),
+        sha256 = "c1ec32b37bf802357356c5fc63f1de927ad080bffca4899af56816adf1d366e5",
+        strip_prefix = "rules_foreign_cc-main",
```
jez, line 362: Let's get the commit hash of `main` instead of putting `main` here. Using `main` will mean that when a commit gets pushed to rules_foreign_cc, anyone who didn't already have rules_foreign_cc in their build cache will get a checksum failure.

jez (commented):
Looks mostly good but we need to switch to a commit hash to avoid getting build failures. Sorry for the confusion

jez:

→ https://go/builds/bui_O4q4uZ9COrThxn
→ https://go/builds/bui_O4q4E7nZ8O3Z2D
→ https://go/builds/bui_O4q4pNrLb0gfLr

jez (approved):
This translates to about a 2% memory improvement for ~0% change in runtime.

PR #7087, opened 2023-06-12
Use the "Property" kind for attr things

jez:
### Motivation

When using the document outline, it's useful to tell which methods are
props and which are actual methods. There's actually support in the LSP
spec for that, so I'd like to implement it.

It's a little hacky, because Sorbet doesn't really track this in its core
structures. Props are just methods to Sorbet. We do have some flags on the
method def for this, but those flags are currently

- only used for attr_reader-alike things, not setters
- very much tied to the runtime/compiler side of things. For example,
  adding a runtime-checked sig to an `attr_reader` will not mark the
  symbol as `isAttrReader`

So I've added a new flag to MethodDef. I've opted not to store this on the
Symbol, because I want it to be harder to accidentally use this flag for
the purpose of type checking (there is more written about my reasoning
here: <https://blog.jez.io/syntactic-control-flow/>). We also have a couple
spare bits there, so it should be fine.

### Test plan

See included automated tests.

Screenshot 2023-06-12 at 4 29 10 PM

### DIFF
--- a/core/FoundDefinitions.h
+++ b/core/FoundDefinitions.h
@@ -155,13 +155,17 @@ struct FoundMethod final {
     struct Flags {
         bool isSelfMethod : 1;
         bool isRewriterSynthesized : 1;
+        // Controls whether this gets compiled as a VM_METHOD_TYPE_IVAR method, for performance
         bool isAttrReader : 1;
+        // Controls whether to display this as an attr_*/prop-defined method in the LSP client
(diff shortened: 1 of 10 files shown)


jez:

→ https://go/builds/bui_O4TOw4WOfdLY5c
→ https://go/builds/bui_O4TOXRc0gxHMgu
→ https://go/builds/bui_O4TPUk5EM3Ys7V

neilparikh (approved):

PR #7088, opened 2023-06-14
Clarify source of VS Code warning

andyw8:
### Motivation

When using multiple LSP servers in VS Code, for example Sorbet LSP alongside Ruby LSP, it's not clear which LSP this message is originating from.

For example, in the screenshot below, the first entry comes from Sorbet LSP and the second comes from Ruby LSP. It's misleading because hover _is_ available for Ruby LSP.

### Test plan

n/a


main/lsp/requests/hover.cc
```diff
@@ -55,7 +55,7 @@ unique_ptr<ResponseMessage> HoverTask::runRequest(LSPTypecheckerDelegate &typech
         auto level = fref.data(gs).strictLevel;
         if (level < core::StrictLevel::True) {
             auto text = fmt::format("This file is `# typed: {}`.\n"
-                                    "Hover, Go To Definition, and other features are disabled in this file.",
+                                    "Hover, Go To Definition, and other Sorbet-based features are disabled in this file.",
```
andyw8, line 58: Alternatively "Sorbet-powered" or "Sorbet-driven"?

jez: let's do Sorbet-powered, and also you will need to clang-format this file.

I can't do a multi-line suggestion outside the diff:

```cpp
        if (level < core::StrictLevel::True) {
            auto text =
                fmt::format("This file is `# typed: {}`.\n"
                            "Hover, Go To Definition, and other Sorbet-powered features are disabled in this file.",
                            level == core::StrictLevel::Ignore ? "ignore" : "false");
```

this should be what you need

jez (changes requested):
mostly good, but let's take one of your alternatives

jez (approved):

PR #7097, opened 2023-06-20
Update T::Types::Simple#name to fix equality issue

bitwise-aiden:
Updated `T::Types::Simple#name` to fully qualify the name of a class rather than calling it's `#name` method so that equality checks better handle constant and `#name` redefinition.

### Motivation

```ruby
class SomeAbstractClass
  sig { abstract.returns(T.any(Time, ActiveSupport::TimeWithZone, Date)) }
  def foo; end
end

class SomeClass < SomeAbstractClass
  sig { override.returns(ActiveSupport::TimeWithZone) }
  def foo; end 
end
```

Results in the following error:

> Incompatible return type in signature for implementation of method `foo`:
> * Base: `T.nilable(T.any(Date, Time))`
> * Implementation: `T.nilable(Time)`
> (The types must be covariant.)

Digging into why this occurs, it was discovered that `ActiveSupport::TimeWithZone#name` has been overridden to return `Time` ([source](https://github.com/rails/rails/blob/e88857bbb9d4e1dd64555c34541301870de4a45b/activesupport/lib/active_support/time_with_zone.rb#L44-L53)). 

This raises the point that checking equality for `#name` directly might cause some unexpected behaviour for anyone that adopts this approach.

The fix proposed in this PR addresses this by comparing the raw names using the `Module#name` method. This means that situations like this will also work correctly:
```ruby
Foo = Class.new
foo = Foo
Object.send(:remove_const, :Foo)
Foo = Class.new
foo == Foo # => false
foo.name == Foo.name # => true
```

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/types/simple.rb
+++ b/gems/sorbet-runtime/lib/types/types/simple.rb
@@ -4,6 +4,8 @@
 module T::Types
   # Validates that an object belongs to the specified class.
   class Simple < Base
+    NAME_METHOD = Module.instance_method(:name)
+
     attr_reader :raw_type
 
     def initialize(raw_type)
@@ -16,7 +18,7 @@ module T::Types
       #
       # `name` isn't normally a hot path for types, but it is used in initializing a T::Types::Union,
       # and so in `T.nilable`, and so in runtime constructions like `x = T.let(nil, T.nilable(Integer))`.
-      @name ||= @raw_type.name.freeze
+      @name ||= NAME_METHOD.bind_call(@raw_type).freeze
     end
 
     # overrides Base
--- a/gems/sorbet-runtime/lib/types/types/simple.rb
+++ b/gems/sorbet-runtime/lib/types/types/simple.rb
@@ -5,6 +5,7 @@ module T::Types
   # Validates that an object belongs to the specified class.
   class Simple < Base
     NAME_METHOD = Module.instance_method(:name)
+    private_constant(:NAME_METHOD)
 
     attr_reader :raw_type
 
@@ -18,7 +19,7 @@ module T::Types
       #
       # `name` isn't normally a hot path for types, but it is used in initializing a T::Types::Union,
       # and so in `T.nilable`, and so in runtime constructions like `x = T.let(nil, T.nilable(Integer))`.
-      @name ||= NAME_METHOD.bind_call(@raw_type).freeze
+      @name ||= NAME_METHOD.bind(@raw_type).call.freeze
     end
 
     # overrides Base
--- a/gems/sorbet-runtime/lib/types/types/simple.rb
+++ b/gems/sorbet-runtime/lib/types/types/simple.rb
@@ -19,7 +19,7 @@ module T::Types
       #
       # `name` isn't normally a hot path for types, but it is used in initializing a T::Types::Union,
       # and so in `T.nilable`, and so in runtime constructions like `x = T.let(nil, T.nilable(Integer))`.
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/types/simple.rb
```diff
@@ -4,6 +4,8 @@
 module T::Types
   # Validates that an object belongs to the specified class.
   class Simple < Base
+    NAME_METHOD = Module.instance_method(:name)
```
jez, line 7: Can we make this a private constant?

jez (commented):
looks mostly good, one request

gems/sorbet-runtime/lib/types/types/simple.rb
```diff
@@ -16,7 +18,7 @@ def name
       #
       # `name` isn't normally a hot path for types, but it is used in initializing a T::Types::Union,
       # and so in `T.nilable`, and so in runtime constructions like `x = T.let(nil, T.nilable(Integer))`.
-      @name ||= @raw_type.name.freeze
+      @name ||= NAME_METHOD.bind_call(@raw_type).freeze
```
froydnj, line 21: Since we make concessions to other pre-Ruby 2.7 versions elsewhere in `sorbet-runtime`, we should do the same here by using `bind(@raw_type).call` on those versions.

gems/sorbet-runtime/test/types/types.rb
```diff
@@ -130,6 +130,19 @@ def subtype_of?(other)
           There might be a constant reloading problem in your application.
         MSG
       end
+
+      it "correctly handles a class that has overriden it's #name method" do
```
froydnj, line 134: Nit:

```suggestion
      it "correctly handles a class that has overridden its #name method" do
```

froydnj (commented):
Drive-by review.

jez (approved):

froydnj (approved):

froydnj:
This PR causes a number of test failures on Stripe's codebase; we'll have to investigate further.  It looks like `#name` might be returning `nil` in some cases whereas before it was returning class names, but it's hard to be sure.

froydnj:
So it looks like that we have some code that creates anonymous classes but explicitly provides a `name` method on those classes, e.g.

```
[dev] main:0> x = Class.new do
[dev] main:0>   def self.name
[dev] main:0>     "foo"
[dev] main:0>   end
[dev] main:0> end
=> #<Class:0x00007fd6e0954f70>
[dev] main:0> t = T::Types::Simple.new(x)
=> #<T::Types::Simple:0x00007fd6e0867518 @raw_type=#<Class:0x00007fd6e0954f70>>
[dev] main:0> t.name
=> "foo"
```

This PR breaks types over those classes, because those types had a "name" prior to this PR and after it they do not.

bitwise-aiden:
Would adding an `|| @raw_type.name` be sufficient to get this to be added?

froydnj:
> Would adding an `|| @raw_type.name` be sufficient to get this to be added?

I think so, assuming that passes Stripe's tests and doesn't add a bunch of overhead.  (I was trying to test the patch for correctness and performance, but was thwarted by the test failures.)

I think the test needs to be changed to test the correct thing, though?

froydnj (changes requested):
Requesting changes just so we don't accidentally merge things in its present state.

bitwise-aiden:
Good call on the test, I think this should now be in a better state.

froydnj (approved):
Tests look good, run on Stripe's codebase looks good.  Thanks for working through this with us!

PR #7098, opened 2023-06-22
Enable sorbet-ruby to build 3.2 including rust-based YJIT

jd-stripe:
I haven't written a ton of bazel, very keen on getting more feedback on the approach. I've left inline comments calling out things I want more eyes on.

There are two main changes here (I tried to break apart by commit but the first includes a few rust references)
1. Include libyaml source so that the Psych gem bundled with 3.2 can be built, [see this ruby issue for more details](https://bugs.ruby-lang.org/issues/18571). The ruby 3.2 release notes mention that libffi source is not included either, the build passes without any changes required though which surprised me. I tried to build ruby 3.3.0-preview1 and that failed because of the missing libffi source. I'm going to work on that in parallel to getting a first round of reviews, flagging it for now.
2. Bring in the rust toolchain so that `rustc` can be added to `$PATH` before `./configure` is run. I've opted for rust 1.58.1 as the YJIT release notes [mention it requires rust 1.58 or higher](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/#:~:text=Building%20YJIT%20now%20requires%20Rust%201.58.0%2B). We could try a new version as rust has good backwards compatibility, we can benchmark internally to see if there's any performance improvement with newer versions. 

### Motivation

We'd like to roll out 3.2 and try out the new YJIT implementation internally for its potential savings, need to build it first!

### Test plan

See included automated tests.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -76,6 +76,17 @@ load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
 
 node_repositories()
 
+load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains")
+
+rules_rust_dependencies()
+
+rust_register_toolchains(
+    edition = "2021",
+    versions = [
+        "1.58.1",
+    ],
+)
+
 BAZEL_VERSION = "5.2.0"
 
 BAZEL_INSTALLER_VERSION_LINUX_X86_64_SHA = "7d9ef51beab5726c55725fb36675c6fed0518576d3ba51fb4067580ddf7627c4"
--- a/third_party/ruby/build-ruby.bzl
+++ b/third_party/ruby/build-ruby.bzl
@@ -56,7 +56,7 @@ done
 build_dir="$(mktemp -d)"
 
 # Add the compiler's directory the PATH, as this fixes builds with gcc
-export PATH="$(dirname "{cc}"):$PATH"
+export PATH="$(dirname "{cc}"):$(dirname $(realpath {rustc})):$PATH"
 
 # Copy everything to a separate directory to get rid of symlinks:
 # tool/rbinstall.rb will explicitly ignore symlinks when installing files,
@@ -93,12 +93,14 @@ run_cmd rm -f configure.bak
 # files that includes `-fvisibility=hidden`. To override it, our flag needs to
 # come after, so we inject a flag right before the `-o` option that comes near
 # the end of the command via OUTFLAG.
-OUTFLAG="-fvisibility=default -o" \
-CC="{cc}" \
-CFLAGS="{copts}" \
-CXXFLAGS="{copts}" \
-CPPFLAGS="{sysroot_flag} ${{inc_path[*]:-}} {cppopts}" \
-LDFLAGS="{sysroot_flag} ${{lib_path[*]:-}} {linkopts}" \
+
+export OUTFLAG="-fvisibility=default -o"
+export CC="{cc}"
+export CFLAGS="{copts}"
+export CXXFLAGS="{copts}"
+export CPPFLAGS="{sysroot_flag} ${{inc_path[*]:-}} {cppopts}"
+export LDFLAGS="{sysroot_flag} ${{lib_path[*]:-}} {linkopts}"
+
 run_cmd ./configure \
         {configure_flags} \
         --with-libyaml-source-dir=$libyaml_loc \
@@ -292,6 +294,7 @@ def _build_ruby_impl(ctx):
             install_append_srcs = "\n".join(install_append_srcs),
             install_gems = "\n".join(install_gems),
         )),
+        tools = [ctx.toolchains["@rules_rust//rust:toolchain_type"].rustc],
     )
 
     return [
@@ -352,7 +355,7 @@ _build_ruby = rule(
         RubyInfo,
         DefaultInfo,
     ],
-    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
+    toolchains = ["@bazel_tools//tools/cpp:toolchain_type", "@rules_rust//rust:toolchain_type"],
     implementation = _build_ruby_impl,
 )
 
--- a/third_party/ruby/build-ruby.bzl
+++ b/third_party/ruby/build-ruby.bzl
@@ -66,6 +66,8 @@ cp -aL "{src_dir}"/* "$build_dir"
 {install_extra_srcs}
 {install_append_srcs}
 
+libyaml_loc="$(realpath {libyaml})"
+
 pushd "$build_dir" > /dev/null
 
 run_cmd() {{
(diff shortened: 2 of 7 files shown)


WORKSPACE
```diff
@@ -76,6 +76,17 @@ load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
 
 node_repositories()
 
+load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains")
+
+rules_rust_dependencies()
+
+rust_register_toolchains(
+    edition = "2021",
```
jd-stripe, line 84: Could bump this down to `2018`

third_party/ruby/build-ruby.bzl
```diff
@@ -91,14 +93,19 @@ run_cmd rm -f configure.bak
...
-OUTFLAG="-fvisibility=default -o" \
-CC="{cc}" \
-CFLAGS="{copts}" \
-CXXFLAGS="{copts}" \
-CPPFLAGS="{sysroot_flag} ${{inc_path[*]:-}} {cppopts}" \
-LDFLAGS="{sysroot_flag} ${{lib_path[*]:-}} {linkopts}" \
+
+# Export these env vars as they will be needed when configuring libyaml which
+# is called with `./configure`
+export OUTFLAG="-fvisibility=default -o"
```
jd-stripe, line 99: If I didn't export these separately then 3.2 wouldn't compile on macos. When compiling libyaml it would fail because linker flags weren't passed in.

third_party/ruby/build-ruby.bzl
```diff
@@ -280,13 +287,16 @@ def _build_ruby_impl(ctx):
...
+            libyaml = ctx.files.libyaml[0].dirname,
+            rustc = ctx.toolchains["@rules_rust//rust:toolchain_type"].rustc.path,
             configure_flags = " ".join(ctx.attr.configure_flags),
             sysroot_flag = ctx.attr.sysroot_flag,
             install_extra_srcs = "\n".join(install_extra_srcs),
             extra_srcs_object_files = " ".join(extra_srcs_object_files),
             install_append_srcs = "\n".join(install_append_srcs),
             install_gems = "\n".join(install_gems),
         )),
+        tools = [ctx.toolchains["@rules_rust//rust:toolchain_type"].rustc],
```
jd-stripe, line 299: At the moment this makes `rules_rust` a dependency for all ruby builds, we could make this selective instead and opt-out all pre-3.2 ruby versions and have it default on for 3.2 onwards.

third_party/ruby/ruby.BUILD
```diff
@@ -10,7 +10,6 @@ ruby(
         "--localstatedir=/var",
         "--disable-maintainer-mode",
         "--disable-dependency-tracking",
-        "--disable-jit-support",
```
froydnj, line 13: The history here is that we put this in so that any version of Ruby running at Stripe would not accidentally have JIT support turned on and therefore it would not accidentally be enabled in prod, etc.

I can totally understand wanting to turn this off for experimental purposes, but I think we should make this configurable on a per-Ruby version basis or we should land this PR as an "enable Ruby 3.2 + YJIT to potentially be built" without this bit, and then leave the JIT enabling for experimental PRs.

third_party/ruby_externals.bzl
```diff
@@ -2,6 +2,27 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file"
...
 # We define our externals here instead of directly in WORKSPACE
 def register_ruby_dependencies():
+    libyaml_version = "0.2.5"
+    http_archive(
+        name = "libyaml",
+        urls = _github_public_urls("yaml/libyaml/releases/download/{}/yaml-{}.tar.gz".format(libyaml_version, libyaml_version)),
+        sha256 = "c642ae9b75fee120b2d96c712538bd2cf283228d2337df2cf2988e3c02678ef4",
+        strip_prefix = "yaml-{}".format(libyaml_version),
+        build_file_content = """
+exports_files(glob(["**/*"]))
```
jd-stripe, line 12: happy to shove this into a separate BUILD file

froydnj (changes requested):
This all seems pretty reasonable, except for the bit noted below.

I guess it would be slightly better if you could configure Rust and libyaml inclusion into the Ruby build on a per-Ruby version basis, but that probably falls into the Too Much Bazel category.

third_party/ruby_externals.bzl
```diff
@@ -99,3 +128,12 @@ def _ruby_urls(path):
         "https://cache.ruby-lang.org/pub/ruby/{}".format(path),
         "https://artifactory-content.stripe.build/artifactory/ruby-lang-cache/pub/ruby/{}".format(path),
     ]
+
+def _github_public_urls(path):
```
jd-stripe, line 132: this is copied from `third_party/externals.bzl` - do you want me to make it public so we can share across files instead?

jd-stripe (commented):
flagging initial questions I had

third_party/ruby/build-ruby.bzl
jd-stripe: I tried to do this but unfortunately I couldn't make including this toolchain configurable. It'll be downloaded for all ruby builds but a no-op for anything < 3.2

jd-stripe:

→ https://go/builds/bui_O9yKwed9ebtw0O
→ https://go/builds/bui_O9yKZrqIKolVYD
→ https://go/builds/bui_O9yKCTgBrZJnBy

jd-stripe:

→ https://go/builds/bui_O9yqxEMRsNzS2S
→ https://go/builds/bui_O9yqgAM6r316If
→ https://go/builds/bui_O9yrKiFtp7H6LS

jd-stripe:

→ https://go/builds/bui_O9zaRkBYBH2FcX
→ https://go/builds/bui_O9zah7dpcv68hv
→ https://go/builds/bui_O9zact2TXYN1qm

jd-stripe:

→ https://go/builds/bui_OAjgLzW3xvJQzg
→ https://go/builds/bui_OAjhvnCJ8fkVEG
→ https://go/builds/bui_OAjhcA3S4AzWc6

jd-stripe:

→ https://go/builds/bui_OAnLOpduYmnkAt
→ https://go/builds/bui_OAnLrbnoFJMJJe
→ https://go/builds/bui_OAnLYCw5E5ODIQ

third_party/ruby/build-ruby.bzl
froydnj: This seems fine, and more in line with how these variables are meant to be used.

third_party/ruby/ruby.BUILD
```diff
@@ -33,7 +36,11 @@ ruby(
     cppopts = [
         "-Wdate-time",
         "-D_FORTIFY_SOURCE=2",
-    ],
+    ] + select({
+        # Don't include JIT statistics unless we're building JIT support
+        "@com_stripe_ruby_typer//tools/config:jit_enabled": ["-DYJIT_STATS=1"],
```
froydnj, line 41: This is not turned on by default?  Is there a performance penalty associated with collecting the statistics, and therefore we might not want to have statistics turned on just because we have the jit turned on?

froydnj (changes requested):
Requesting changes just for the question about stats below, but this seems fine otherwise.

third_party/ruby/ruby.BUILD
froydnj: Works for me!

froydnj (approved):

PR #7103, opened 2023-06-26
Magic symbols for tracking untyped 

neilparikh:
Adds new magic symbols that denote sources of untyped, and updates a few places to use them along with `types::untyped` rather than `types::untypedUntracked`. If the untyped blames to `<root>#<none>` is still large after this, we can add symbols to more places where untyped is used.

Example with running sorbet with the `--track-untyped` flag on the following code:
```
# typed: true
class TestUntyped
  def test_shaped
    b = {
      1 => 2,
      2 => 3,
      foo: :bar
    }
    b[1].bar
  end

  def test_super
    a = super
    a.foo
  end
end
```

```
[{"path":"<none>","package":"<none>","owner":"<Magic>::<UntypedSource>","name":"<HashOfUntyped>","count":2},
{"path":"<none>","package":"<none>","owner":"<Magic>::<UntypedSource>","name":"<super>","count":2}
```

### Motivation

Better tracking of untyped for cases that currently blame to `root#none`

### Test plan

test-pr-on-payserver and a few manual test cases.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -3294,7 +3294,7 @@ public:
             // people's codebases to it.
             //
             // TODO(jez) This could be another "if you're in `typed: strict` you need typed shapes"
-            res.returnType = Types::untypedUntracked();
+            res.returnType = Types::untyped(Symbols::Magic_UntypedSource_squareBracketsEq());
         }
     }
 } Shape_squareBracketsEq;
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -275,7 +275,7 @@ unique_ptr<Error> matchArgType(const GlobalState &gs, TypeConstraint &constr, Lo
                                Loc originForUninitialized, bool mayBeSetter = false) {
     TypePtr expectedType = Types::resultTypeAsSeenFrom(gs, argSym.type, method.data(gs)->owner, inClass, targs);
     if (!expectedType) {
-        expectedType = Types::untyped(gs, method);
+        expectedType = Types::untyped(method);
     }
 
     expectedType = Types::replaceSelfType(gs, expectedType, selfType);
@@ -333,7 +333,7 @@ unique_ptr<Error> missingArg(const GlobalState &gs, Loc argsLoc, Loc receiverLoc
         e.setHeader("Missing required keyword argument `{}` for method `{}`", argName, method.show(gs));
         auto expectedType = Types::resultTypeAsSeenFrom(gs, arg.type, method.data(gs)->owner, inClass, targs);
         if (expectedType == nullptr) {
-            expectedType = Types::untyped(gs, method);
+            expectedType = Types::untyped(method);
         }
         e.addErrorLine(arg.loc, "Keyword argument `{}` declared to expect type `{}` here:", argName,
                        expectedType.show(gs));
@@ -568,7 +568,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
             TypeErrorDiagnostics::explainUntyped(gs, e, what, args.fullType, args.originForUninitialized);
         }
 
-        return DispatchResult(Types::untyped(gs, args.thisType.untypedBlame()), std::move(args.selfType),
+        return DispatchResult(Types::untyped(args.thisType.untypedBlame()), std::move(args.selfType),
                               Symbols::noMethod());
     } else if (symbol == Symbols::void_()) {
         if (!args.suppressErrors) {
@@ -1045,7 +1045,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
                         auto kwParamType =
                             Types::resultTypeAsSeenFrom(gs, kwParam->type, method.data(gs)->owner, symbol, targs);
                         if (kwParamType == nullptr) {
-                            kwParamType = Types::untyped(gs, method);
+                            kwParamType = Types::untyped(method);
                         }
                         // TODO(jez) Highlight untyped code for this error
                         if (Types::isSubTypeUnderConstraint(gs, *constr, kwSplatValueType, kwParamType,
@@ -1339,7 +1339,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
 
         TypePtr blockType = Types::resultTypeAsSeenFrom(gs, bspec.type, data->owner, symbol, targs);
         if (!blockType) {
-            blockType = Types::untyped(gs, method);
+            blockType = Types::untyped(method);
         }
 
         component.blockReturnType = Types::getProcReturnType(gs, Types::dropNil(gs, blockType));
@@ -1401,7 +1401,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
     }
 
     if (!resultType) {
-        resultType = Types::untyped(gs, method);
+        resultType = Types::untyped(method);
     } else if (!constr->isEmpty() && constr->isSolved()) {
         resultType = Types::instantiate(gs, resultType, *constr);
     }
@@ -1434,7 +1434,7 @@ TypePtr getMethodArguments(const GlobalState &gs, ClassOrModuleRef klass, NameRe
             continue;
         }
         if (arg.type == nullptr) {
-            args.emplace_back(core::Types::untyped(gs, method));
+            args.emplace_back(core::Types::untyped(method));
             continue;
         }
         args.emplace_back(Types::resultTypeAsSeenFrom(gs, arg.type, data->owner, klass, targs));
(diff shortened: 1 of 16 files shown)


neilparikh:

→ https://go/builds/bui_O9fjlkSusxDM3R
→ https://go/builds/bui_O9fjMw2uPzweLV
→ https://go/builds/bui_O9fj3C56gOkSqt

neilparikh:

→ https://go/builds/bui_O9hq1CEerLxwU4
→ https://go/builds/bui_O9hq90obIbw6Yf
→ https://go/builds/bui_O9hqqO1qQY6mBI

core/types/types.cc
```diff
@@ -77,19 +77,21 @@ TypePtr Types::Float() {
 }
 
 TypePtr Types::arrayOfUntyped() {
-    static vector<TypePtr> targs{Types::untypedUntracked()};
+    static vector<TypePtr> targs{Types::untyped(Symbols::Magic_UntypedSource_ArrayOfUntyped())};
```
jez, line 80: It's likely that for these helpers, we want to take the blame as an argument, and pass in a more specific thing where they're used. For example, `hashOfUntyped` is used for both Magic buildHash as well as shapes. We likely want to bucket those separately. Similarly for the other `OfUntyped` helpers here.

jez:
One thing you may want to consider adding is a method like `Types::untypedErrorRecovery`. You could either build it with a specific symbol to blame to, or leave it blaming to `<none>`, but the benefit would be that people can do a quick grep of `untypedUntracked` to see call sites that we've looked at and decided we didn't need a blame for. It would also make it more clear to people trying to cargo-cult that they shouldn't use a bare `untypedUntracked` in new code.

jez:
Some other usages of untypedUntracked we may want to include:

- https://github.com/sorbet/sorbet/blob/master/infer/environment.cc#L1342
- https://github.com/sorbet/sorbet/blob/master/infer/environment.cc#L1363
- https://github.com/sorbet/sorbet/blob/master/infer/environment.cc#L1524
- `T.unsafe` (if this does not already blame to itself)
- https://github.com/sorbet/sorbet/blob/master/core/types/types.cc#L1027
- https://github.com/sorbet/sorbet/blob/master/core/types/types.cc#L1086C3
- https://github.com/sorbet/sorbet/blob/master/core/types/calls.cc#L3294
- https://github.com/sorbet/sorbet/blob/master/core/types/calls.cc#L3911

jez:
> ### Test plan
>
> Existing tests

Note that this is _not_ covered by existing tests, because we do not run any tests with the build configuration required to get untyped blames working. It will be covered by the build system, but none of the existing testdata / CLI tests will fail if this is doing something wrong or broken.

jez:
^ that's fine, btw, I just wanted to be clear on it.

jez (approved):
This looks good as a starting point. I think that either in this change or followup changes, we should:

- split out the OfUntyped helpers to accept an explicit blame symbol
- add blame symbols for the places i linked in an earlier comment

neilparikh:
`untypedErrorRecovery` is a good idea. The one edge case I can think of is that in a file that's a lower `typed` level than the error; ex. if we do some error recovery for a `typed: strict` error, in a `typed: true` file, that will still contribute to the `untyped` counts, and is something actionable to fix. We can just manually change those to use a more precise blame, so not an issue, just something to note.

> I'm fine if you don't want to include those in this PR, but I think we should probably just do them soon, so we don't have to revisit this.

I had some of those my list to track, I just wasn't fully sure which were relevant, and didn't want to block getting this PR up for review on those. I'll include them in this PR, shouldn't add much to the diff.

neilparikh:

→ https://go/builds/bui_OAPB1sR9nfk5R9
→ https://go/builds/bui_OAPBpymHCqfSZ2
→ https://go/builds/bui_OAPBlo3BZ1uxcN

core/types/calls.cc
```diff
@@ -2143,7 +2145,7 @@ class Magic_buildRange : public IntrinsicMethod {
         auto secondArgIsNil = other.isNilClass();
         if (firstArgIsNil) {
             if (secondArgIsNil) {
-                rangeElemType = Types::untypedUntracked();
+                rangeElemType = Types::untyped(Symbols::Magic_UntypedSource_buildArray());
```
jez, line 2148: Why are we using buildArray in the `Magic_buildRange` intrinsic?

core/types/calls.cc
```diff
@@ -415,6 +415,7 @@ MethodRef guessOverload(const GlobalState &gs, ClassOrModuleRef inClass, MethodR
 
         // If keyword args are present, interpret them as an untyped hash
         if (numPosArgs < args.size()) {
+            // TODO(neil): this should actually blame to an error recovery magic symbol
```
jez, line 418: Why would this blame to an error recovery magic symbol? If I'm reading this right, there's no error we're recovering from here.
 
If anything, I think this should blame to something that would indicate we need better support for ruby 3 keyword arg handling.

But it's also possible that we don't want to track anything here: this doesn't _really_ count as a usage of untyped—rather, it's Sorbet that uses `untyped` as a convenient way to guess an overload. Once that overload is guessed, it will be used for type checking later on, where if any of the keyword args are untyped, they will be checked and have their untyped usages reported.

core/types/calls.cc
```diff
@@ -2989,7 +2991,8 @@ class Magic_splat : public IntrinsicMethod {
         if (!dispatched.main.errors.empty()) {
             // In case of an error, the splat is converted to an array with a single
             // element; be conservative in what we declare the element type to be.
-            res.returnType = Types::arrayOfUntyped();
+            // TODO(neil): this should actually blame to an error recovery magic symbol
```
jez, line 2994: When it says "in case of an error," what it's referring to is the fake dispatch to `.to_a` that happens a few lines above. Sorbet never reports those errors to the user, so this is not an error recovery case.

Rather, for this we do want the untyped to blame to the Magic_splat symbol, like you've done it. Knowing that untyped are coming from this would let us know whether we wanted to look at better typing support for splats.

core/types/calls.cc
```diff
@@ -3291,7 +3294,7 @@ class Shape_squareBracketsEq : public IntrinsicMethod {
             // people's codebases to it.
             //
             // TODO(jez) This could be another "if you're in `typed: strict` you need typed shapes"
-            res.returnType = Types::untypedUntracked();
+            res.returnType = Types::untyped(Symbols::Magic_UntypedSource_squareBracketsEq());
```
jez, line 3297: Any class can define a method called `[]=`. So the name "UntypedSource_squareBracketsEq" is a bit too broad. This blame is _specifically_ for `[]=` on shapes. We will want the name of this blame to reflect that.

core/types/types.cc
```diff
@@ -1024,7 +1031,7 @@ TypePtr Types::applyTypeArguments(const GlobalState &gs, const CallLocs &locs, u
     }
 
     if (genericClass.data(gs)->typeMembers().empty()) {
-        return Types::untypedUntracked();
+        return Types::untyped(Symbols::Magic_UntypedSource_applyTypeArguments());
```
jez, line 1034: I believe this is actually a case where we _would_ want to blame to an error recovery symbol.

I believe this case can only happen if you attempt to provide type arguments to a generic class that doesn't accept type arguments. That error is reported above, and this condition recovers from it.

core/types/types.cc
```diff
@@ -1075,15 +1082,15 @@ TypePtr Types::applyTypeArguments(const GlobalState &gs, const CallLocs &locs, u
             if (validBounds) {
                 targs.emplace_back(argType);
             } else {
-                targs.emplace_back(Types::untypedUntracked());
+                targs.emplace_back(Types::untyped(Symbols::Magic_UntypedSource_applyTypeArguments()));
```
jez, line 1085: This is again an error case. `validBounds` is only ever `false` if we reported an error.

core/types/types.cc
```diff
@@ -1075,15 +1082,15 @@ TypePtr Types::applyTypeArguments(const GlobalState &gs, const CallLocs &locs, u
...
             }
 
             ++it;
         } else if (genericClass == Symbols::Hash() && i == 2) {
             auto tupleArgs = targs;
             targs.emplace_back(make_type<TupleType>(tupleArgs));
         } else {
-            targs.emplace_back(Types::untypedUntracked());
+            targs.emplace_back(Types::untyped(Symbols::Magic_UntypedSource_applyTypeArguments()));
```
jez, line 1093: This is another error condition. In this case, the user didn't provide enough type args, and we already reported that error above.

jez (commented):
Looking good! Still some comments.

core/types/calls.cc
neilparikh: >  this doesn't really count as a usage of untyped—rather, it's Sorbet that uses untyped as a convenient way to guess an overload.

Yup, this is the mental model I had too, that this is a transient untyped object, and that it will go away once the overload is checked (and that's why I called  the `hashOfUntyped` overload without a blame).

I wrote error recovery symbol just as a default "we don't care about this". I'll drop the comment and leave it as-is (and we create a new ignore symbol if we really care about removing all `untypedUntracked` uses).

core/types/calls.cc
neilparikh: Bad copy-paste, I'll change to `buildRange`

core/types/calls.cc
jez: ah, that makes sense!

PR #7106, opened 2023-06-27
be more parsimonious when replacing methods

froydnj:
In most of the places inside `sorbet-runtime` where we call `replace_method`, we mostly care about the side effect of replacing the method in question, and not about the return value.  In most places where we poke at the return value, we only care about calling `bind()` on it, with a subsequent `call()` at some point.

Ergo, allocating a `ReplacedMethod` object is really the uncommon case, and we should avoid allocating a `ReplacedMethod` if we don't need to.

### Motivation

These objects and the callsites for `bind`/`call` were showing up on some allocation profiles for preloading Stripe services..

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/class_utils.rb
+++ b/gems/sorbet-runtime/lib/types/private/class_utils.rb
@@ -94,7 +94,7 @@ module T::Private::ClassUtils
   # overriding it (if it is defined by one of mod's ancestors). Returns a ReplacedMethod instance
   # on which you can call `bind(...).call(...)` to call the original method, or `restore` to
   # restore the original method (by overwriting or removing the override).
-  def self.replace_method(mod, name, &blk)
+  def self.replace_method(mod, name, original_only=false, &blk)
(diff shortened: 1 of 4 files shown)


gems/sorbet-runtime/lib/types/private/class_utils.rb
```diff
@@ -94,7 +94,7 @@ def self.def_with_visibility(mod, name, visibility, method=nil, &block)
   # overriding it (if it is defined by one of mod's ancestors). Returns a ReplacedMethod instance
   # on which you can call `bind(...).call(...)` to call the original method, or `restore` to
   # restore the original method (by overwriting or removing the override).
-  def self.replace_method(mod, name, &blk)
+  def self.replace_method(mod, name, original_only=false, &blk)
```
neilparikh, line 97: The comment above should be changed to `If original_only is false, returns a ReplacedMethod instance...`

neilparikh (approved):

PR #7112, opened 2023-06-29
lazily allocate structures in `Signature#initialize`

froydnj:
### Motivation

We allocate lists and hashes for several things in our internal `Signature` object:

* argument types for positional arguments
* argument types for kwargs
* required kwarg names

We unconditionally allocate these, even though most functions tilt towards positional or kwargs, but not both.  e.g. Stripe's codebase is majority positional-argument only functions.

We can save some memory by lazily allocating these objects, and using frozen copies if it turned out that we never allocated them, so as to present the same interface to the outside world (e.g. Tapioca).

We sneak in a change to use local variables for the core loop, since those are epsilon faster to access than instance variables.

This change saves not quite 2% of the memory allocated by `sorbet-runtime` when preloading Stripe's main service.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/signature.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/signature.rb
@@ -39,8 +39,6 @@ class T::Private::Methods::Signature
   def initialize(method:, method_name:, raw_arg_types:, raw_return_type:, bind:, mode:, check_level:, on_failure:, parameters: method.parameters, override_allow_incompatible: false, defined_raw: false)
     @method = method
     @method_name = method_name
-    @arg_types = []
-    @kwarg_types = {}
     @block_type = nil
     @block_name = nil
     @rest_type = nil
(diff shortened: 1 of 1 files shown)


michaelherold:
This change is [the only difference between 0.5.10894 and 0.5.10898](https://my.diffend.io/gems/sorbet-runtime/0.5.10894/0.5.10898). It causes [Spoom](https://github.com/shopify/spoom) to start segfaulting on our application.

I'm not sure how to begin debugging what's causing the issue. Any suggestions?

froydnj:
> This change is [the only difference between 0.5.10894 and 0.5.10898](https://my.diffend.io/gems/sorbet-runtime/0.5.10894/0.5.10898). It causes [Spoom](https://github.com/shopify/spoom) to start segfaulting on our application.
> 
> I'm not sure how to begin debugging what's causing the issue. Any suggestions?

Can you open a new issue for this problem?

Seeing a stack trace (or as much of one as you can post) would be helpful.

PR #7114, opened 2023-06-30
Write the blame symbol when pickling BlamedUntyped

neilparikh:
`T.untyped`s coming from methods defined in payload RBIs had no untyped blame symbol, because it wasn't serialized after we computed the initial global state.

### Motivation

Better tracking of `T.untyped`.

### Test plan

`test-pr-on-pay-server` + manual test:
```ruby
# typed: true
class TestUntyped
  extend T::Sig
  
  sig { returns(T.untyped) }
  def test
    2
  end

  def local_call
    a = test
  end

  def unsafe
    T.unsafe(1)
  end

  def method_call
    Method.new.call
  end
end
```
```
[{"path":"https://github.com/sorbet/sorbet/tree/master/rbi/core/method.rbi","package":"<none>","owner":"Method","name":"call","count":1},
{"path":"untyped_track.rb","package":"<none>","owner":"TestUntyped","name":"test","count":1},
{"path":"https://github.com/sorbet/sorbet/tree/master/rbi/sorbet/t.rbi","package":"<none>","owner":"T.class_of(T)","name":"unsafe","count":1}]
```

### DIFF
--- a/core/serialize/serialize.cc
+++ b/core/serialize/serialize.cc
@@ -418,7 +418,10 @@ void SerializerImpl::pickle(Pickler &p, const TypePtr &what) {
     switch (what.tag()) {
         case TypePtr::Tag::UnresolvedAppliedType:
         case TypePtr::Tag::UnresolvedClassType:
-        case TypePtr::Tag::BlamedUntyped:
+        case TypePtr::Tag::BlamedUntyped: {
+            p.putU4(what.untypedBlame().rawId());
+            break;
+        }
         case TypePtr::Tag::ClassType: {
             auto c = cast_type_nonnull<ClassType>(what);
             p.putU4(c.symbol.id());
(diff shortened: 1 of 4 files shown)


neilparikh:

→ https://go/builds/bui_OAwufhxIid0lbX
→ https://go/builds/bui_OAwuZDkQ3zytzG
→ https://go/builds/bui_OAwuUBEwiZhBtB

froydnj (changes requested):
Can we write a `test/cli` test to verify the result programmatically?  (`test/cli/track-untyped` looks promising, though the output isn't quite what you presented in the PR.)

jez:
> Can we write a `test/cli` test to verify the result programmatically? (`test/cli/track-untyped` looks promising, though the output isn't quite what you presented in the PR.)

@froydnj Unfortunately none of the typed blame stuff is tested, because it's only valid when we're compiling with `--config=untyped-blame`, which we don't run our tests with.

We'd have to build a new test harness for like CLI tests but they're only ever built with a specific build configuration.

I think we _might_ want to do that? But it's bigger than this PR, and at the very least, this untyped blame mode is already only ever run manually by members of our team, so the blast radius on bugs like this is low at the moment.

froydnj (approved):
> Unfortunately none of the typed blame stuff is tested, because it's only valid when we're compiling with --config=untyped-blame, which we don't run our tests with.

Ahhh, I wondered if that was the case.  Good to have confirmation!

PR #7121, opened 2023-07-03
Raise exception when `link` initialization fails during `statsd` init

maruth-stripe:
### Motivation

We should fail loudly and clearly when `statsd` initialization fails and `link` is `nullptr`.

### Test plan

See included automated tests.

### DIFF
--- a/common/statsd/statsd.cc
+++ b/common/statsd/statsd.cc
@@ -77,8 +77,7 @@ public:
     StatsdClientWrapper(string host, int port, string prefix)
         : link(statsd_init_with_namespace(host.c_str(), port, cleanMetricName(prefix).c_str())) {
         if (link == nullptr) {
-            Exception::raise("statsd initialization failed: link == nullptr. host: {} port: {} prefix: {}", host, port,
-                             prefix);
+            Exception::raise("statsd initialization failed: host: {} port: {} prefix: {}", host, port, prefix);
         }
     }
 
--- a/common/statsd/statsd.cc
+++ b/common/statsd/statsd.cc
@@ -77,7 +77,7 @@ public:
     StatsdClientWrapper(string host, int port, string prefix)
         : link(statsd_init_with_namespace(host.c_str(), port, cleanMetricName(prefix).c_str())) {
         if (link == nullptr) {
-            Exception::raise("statsd initialization failed: host: {} port: {} prefix: {}", host, port, prefix);
+            Exception::raise("statsd initialization failed: host={} port={} prefix={}", host, port, prefix);
         }
     }
 
--- a/common/statsd/statsd.cc
+++ b/common/statsd/statsd.cc
@@ -49,6 +49,8 @@ class StatsdClientWrapper {
     }
 
     void addMetric(string_view name, size_t value, string_view type, vector<pair<const char *, const char *>> tags) {
+        ENFORCE(link != nullptr);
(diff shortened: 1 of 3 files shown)


test/cli/statsd-invalid-link/test.out
```diff
@@ -0,0 +1,12 @@
+No errors! Great job.
+Name or service not known
+Exception::raise(): statsd initialization failed: host: f4K#l1N&.com port: 8200 prefix: foo.bar.counters
+
+Backtrace:
+  #3 0xef93fc sorbet::StatsD::submitCounters()
```
jez, line 6: These are memory addresses. You cannot put these in exp files, as they change every time we rebuild sorbet

jez: Use sed and/or grep to capture the part of the output that you want

maruth-stripe: fixed!

common/statsd/statsd.cc
```diff
@@ -49,6 +49,8 @@ class StatsdClientWrapper {
     }
 
     void addMetric(string_view name, size_t value, string_view type, vector<pair<const char *, const char *>> tags) {
+        ENFORCE(link != nullptr);
```
jez, line 52: you could also make `link` const so that it can only be set by the constructor and then you'd know that either the constructor raised, or it was set to a non-null value, and then not have to ENFORCE at all the usages.

test/cli/statsd-invalid-link/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+# checking exception is raised when host is invlaid
+main/sorbet --silence-dev-message -e 'class Foo; end' --statsd-host="f4K#l1N&.com" --statsd-prefix=foo.bar 2> >(grep --color=never "statsd initialization failed")
```
jez, line 3: waiting for the day that someone registers this domain to break our CI 🙂

common/statsd/statsd.cc
```diff
@@ -73,7 +75,11 @@ class StatsdClientWrapper {
 
 public:
     StatsdClientWrapper(string host, int port, string prefix)
-        : link(statsd_init_with_namespace(host.c_str(), port, cleanMetricName(prefix).c_str())) {}
+        : link(statsd_init_with_namespace(host.c_str(), port, cleanMetricName(prefix).c_str())) {
+        if (link == nullptr) {
+            Exception::raise("statsd initialization failed: host: {} port: {} prefix: {}", host, port, prefix);
```
jez, line 80: slight preference for using `host={}` instead of `host: {}` as it makes it easier to have splunk ingest these lines

jez (approved):

test/cli/statsd-invalid-link/test.sh
maruth-stripe: `#` isn't allowed in domain names :p

common/statsd/statsd.cc
maruth-stripe: the `statsd` client methods called on `link` take it as an argument and make no guarantee that they won't modify it; won't typecheck

common/statsd/statsd.cc
maruth-stripe: done!

PR #7127, opened 2023-07-07
Implement `visible_to "tests"`

aisamanra:
### Motivation
This adds a requested feature to `visible_to`: namely, we can now add the line `visible_to "tests"` as a blanket statement to allow a package to be freely imported via `test_import` but still require explicit visibility modifiers for other situations.

Using `visible_to "tests"` with no other lines makes the package unimportable via `import`, allowing _only_ `test_import`. This may be useful for "leaf" packages.

### Test plan

See included automated tests. Added test cases that exercise `visible_to "tests"` functionality.

### DIFF
--- a/test/testdata/packager/test_visibility/exporters/explicit_visible_to/__package.rb
+++ b/test/testdata/packager/test_visibility/exporters/explicit_visible_to/__package.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
+class Exporters::ExplicitVisibleTo < PackageSpec
+  visible_to Importers::Allowed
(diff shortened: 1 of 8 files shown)


test/testdata/packager/visibility/hippanthropy/__package.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
+class Hippanthropy < PackageSpec
+  # this import will be okay because `Foo` includes `visible_to "tests"`
+  test_import Foo
+end
```
jez, line 8: Could we get a test that shows the opposite, like what the error would look like if it's violated?

jez (approved):

PR #7149, opened 2023-07-17
Type T::Enum#serialize

### LINKED ISSUE #4035: Well typed T::Enum serialized values
### Problem

Quite often I work with `T::Enum` to get all the nice benefits it provides over plain strings/symbols, however at the point where I'm translating between some boundary which doesn't support rich types, I call `.serialize`. This is not great because currently `.serialize` returns `T.untyped` and in 99% of cases the value is of a single type (for me most often strings). I'm wary of any `T.untyped` code and would like `T::Enum` to be more strictly typed.

### Proposed solution

We probably would need to introduce a new `T::Enum` to make migrating easier. This new enum class could leverage generics to narrow down the serialized value type.

```rb
class MyEnum < T::TypedValueEnum
  Value = type_template(fixed: String)
  enums do
    Foo = new('foo')
  end
end

T.reveal_type(MyEnum::Foo.serialize) # => String
```

Hopefully migrating via code mod would be pretty straight forward as most enum values are literals or constants and thus easy to resolve a type for. We would eventually make `T::TypedValueEnum` the default and make the current `T::Enum` an `T::UntypedValueEnum` (naming bike-shed required).

neilparikh:
### Motivation

Fixed #4035

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/TEnum.cc
+++ b/rewriter/TEnum.cc
@@ -49,12 +49,11 @@ ast::Send *asEnumsDo(ast::ExpressionPtr &stat) {
     }
 }
 
-vector<ast::ExpressionPtr> badConst(core::MutableContext ctx, core::LocOffsets headerLoc, core::LocOffsets line1Loc) {
+void badConst(core::MutableContext ctx, core::LocOffsets headerLoc, core::LocOffsets line1Loc) {
     if (auto e = ctx.beginError(headerLoc, core::errors::Rewriter::BadTEnumSyntax)) {
         e.setHeader("All constants defined on an `{}` must be unique instances of the enum", "T::Enum");
         e.addErrorLine(ctx.locAt(line1Loc), "Enclosing definition here");
     }
-    return {};
 }
 
 ast::Send *findSelfNew(ast::ExpressionPtr &assignRhs) {
@@ -100,8 +99,13 @@ ast::Send *findSelfNew(ast::ExpressionPtr &assignRhs) {
     return findSelfNew(cast->arg);
 }
 
-vector<ast::ExpressionPtr> processStat(core::MutableContext ctx, ast::ClassDef *klass, ast::ExpressionPtr &stat,
-                                       FromWhere fromWhere) {
+struct ProcessStatResult {
+    vector<ast::ExpressionPtr> stats;
+    core::TypePtr type;
+};
+
+std::optional<ProcessStatResult> processStat(core::MutableContext ctx, ast::ClassDef *klass, ast::ExpressionPtr &stat,
+                                             FromWhere fromWhere) {
     auto *asgn = ast::cast_tree<ast::Assign>(stat);
     if (asgn == nullptr) {
         return {};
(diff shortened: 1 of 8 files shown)


neilparikh:

→ https://go/builds/bui_OHYLQ91aS90235
→ https://go/builds/bui_OHYLdzRz2a0Kk5
→ https://go/builds/bui_OHYLd1Lq9w81HS

neilparikh:

→ https://go/builds/bui_OKYP1B6NDefP2n
→ https://go/builds/bui_OKYPLUVNl6uCzd
→ https://go/builds/bui_OKYPxmmPBqxQJB

jez:
I don't think this *quite* fixes #4035—there's still the problem that you can't generically operate on `T::Enum` values and serialize them, and we won't get that until we replace this approach with generics.

So I'd prefer to keep that issue open for the time being, or at least make a new issue after this PR lands.

rewriter/TEnum.cc
```diff
@@ -123,6 +128,18 @@ vector<ast::ExpressionPtr> processStat(core::MutableContext ctx, ast::ClassDef *
     //
     // So we're good to process this thing as a new T::Enum value.
 
+    core::TypePtr serializeType = core::Types::untypedUntracked();
```
jez, line 131: Maybe we should blame this to a magic symbol?

jez: Alternatively, you could use `nullptr` here. I think it's the case that this type should never show up anywhere from this rewriter pass, and one way to make sure of that would be to use `nullptr`, where the code would simply crash if we forgot to account for the `nullptr` case.

jez: You're also using both `bottom` and `untyped` as a signaling value. Maybe that's another indication we should just use `nullptr` for both cases.

jez (approved):
This looks great! Some non-blocking feedback.

rewriter/TEnum.cc
neilparikh: The reason I left it without a blame is because I just don't emit the `serialize` method if the inferred return type is `T.untyped`. As a sidenote, I think even if we added a blame, and used the `T.untyped` with blame to construct the new method, the blame would be lost once we pull the `Symbol` out of the `TypePtr` to generate the AST.

The reason I used `bottom` and `untyped` as signaling values is that they work well with how I'm `any`ing the types together. `any(x, untyped) == untyped`, which works when any of the variants are not literals, and `any(x, bottom) = x`, which works well as a default/ignore value. Using `nullptr` for both cases would make the code a bit more verbose I think, because then I couldn't just pass them to `any` and have it deal with them for me.

I'm happy to change it use `nullptr` if you think it's better to be explicit here, maybe this is a bit too magical.

neilparikh:

→ https://go/builds/bui_OLchp00NbGcUSh
→ https://go/builds/bui_OLchcSV4AwsDkI
→ https://go/builds/bui_OLch7ANohSAL0e

Issue #7148, opened 2023-07-17
Sorbet extension throws error when editor is opened in a project not using Sorbet


vinistock:
### Reproduction steps

Open a project that does not use Sorbet with the VS Code extension

### Expected

No errors thrown and Sorbet quietly deactivated.

### Actual

Sorbet displays two errors to the user

<img width="463" alt="Screenshot 2023-07-17 at 1 51 32 PM" src="https://github.com/sorbet/sorbet/assets/18742907/6c03601d-7ce5-46e3-97d1-c401c0fffa06">

PR #7150, opened 2023-07-17
Fix: Workspace with no active Sorbet configuration shows an initialization error.

damolina-stripe:
*Issue*: Opening a workspace that contains Ruby files but no active Sorbet configuration, shows an extension initialization error.

 *Fix*: Do not try to run Sorbet when there is no active configuratio (i.e. remain `Disabled`).

### Motivation

https://github.com/sorbet/sorbet/issues/7148

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/src/sorbetStatusProvider.ts
+++ b/vscode_extension/src/sorbetStatusProvider.ts
@@ -125,6 +125,12 @@ export class SorbetStatusProvider implements Disposable {
    */
   public async startSorbet(): Promise<void> {
     if (this.isStarting) {
+      this.context.log.trace("Ignoring start request, already starting.");
+      return;
+    }
+
+    if (!this.context.configuration.activeLspConfig) {
+      this.context.log.info("Ignoring start request, no active configuration.");
       return;
     }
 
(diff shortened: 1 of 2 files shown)


vscode_extension/src/sorbetStatusProvider.ts
```diff
@@ -125,6 +125,12 @@ export class SorbetStatusProvider implements Disposable {
    */
   public async startSorbet(): Promise<void> {
     if (this.isStarting) {
+      this.context.log.trace("Ignoring start request, already starting.");
+      return;
+    }
+
+    if (!this.context.configuration.activeLspConfig) {
+      this.context.log.info("Ignoring start request, no active configuration.");
```
jez, line 133: Should we add something here like, "... If you want to use Sorbet in this project, please add a `sorbet/config` file" or whatever instructions people would have to follow to get it to start?

jez (approved):
lgtm because it fixes the problem, but I think we can make the change even slightly better

vscode_extension/src/sorbetStatusProvider.ts
damolina-stripe: Following up as https://github.com/sorbet/sorbet/pull/7152

PR #7152, opened 2023-07-18
Update missing-configuration logging

damolina-stripe:
Update missing configuration log message to include a hint on adding `sorbet/config`.
Docs: https://sorbet.org/docs/adopting#step-2-initialize-sorbet-in-our-project 

### Motivation

Suggestion from https://github.com/sorbet/sorbet/pull/7150 

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/src/sorbetStatusProvider.ts
+++ b/vscode_extension/src/sorbetStatusProvider.ts
@@ -177,7 +177,7 @@ export class SorbetStatusProvider implements Disposable {
 
     if (!this.context.configuration.activeLspConfig) {
       this.context.log.info(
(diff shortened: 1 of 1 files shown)


jez (approved):
Looks good but just to check: is that advice actually correct? When I suggested it, it was a guess.

damolina-stripe:
> Looks good but just to check: is that advice actually correct? When I suggested it, it was a guess.

We do not print this warning in common scenarios like restart (or loop-restart like when `pay up` is not started).  It might not be true for stale configuration names.

damolina-stripe:
@jez Updated the comment to point to https://sorbet.org/docs/vscode which has the VSCode appropriate instructions.

PR #7138, opened 2023-07-11
Emily/p kwnorest refactor

egiurleo:
### Test plan

See included automated tests.

PR #6877, opened 2023-03-23
Start to bring the parser up-to-date with Ruby 3.2

egiurleo:
Every commit in this PR corresponds to a [commit on the whitequark parser for Ruby 3.2](https://github.com/whitequark/parser/commits/master/lib/parser/ruby32.y). It will probably be easiest to review this PR by going commit-by-commit and comparing to the corresponding whitequark commit, which I've linked in the commit messages.

### Motivation
The primary motivation of this PR is to group together a bunch of small changes/refactors to the parser that reflect the current state of the whitequark parser. A couple of larger changes will be created in separate PRs.

### Test plan
New whitequark parser tests have been imported using the `tools/scripts/import_whitequark.sh` script.

### Note

There are two sets of changes that I have not brought over in this PR:

1. whitequark/parser@820cc04b8d176c6f46ccc9a010c55f6113f4b2ad or whitequark/parser@1d275fc21227f538a2dda74a54939cd55b1c9200 -- there are some aspects of translating them to the Sorbet parser grammar that I don't understand yet, but I'll try to address them in a future PR.

2. Forwarding anonymous rest and kwrest args -- this requires changes in the Desugarer and probably in the resolver, so I will make those changes separately.

PR #7153, opened 2023-07-18
Another Ruby 3.2 parser update

egiurleo:
Follow up to #6877

### Motivation
This PR attempts to bring the Sorbet parser grammar closer to whitequarks by making changes inspired by the following whitequark commits:
- https://github.com/whitequark/parser/commit/820cc04
- https://github.com/whitequark/parser/commit/1d275fc

These two changes replace instances of the `kwrest_mark` and `kNIL` tokens with the `p_kwnorest` rule. Afaik, it is a refactor and should not change the behavior of the parser, but should make it easier to implement other parser changes in the future.

### Implementation

The changes in this PR don't mirror the whitequark changes 1-to-1 because they expose a difference in the underlying implementation between the two parsers:

The whitequark parser has the `p_kwnorest` rule return a list of tokens, and then other rules pull the tokens from that list and consume them.

The Sorbet parser doesn't have a concept of a "list of tokens," if I understand it correctly. I tried implementing this change where the `p_kwnorest` rule returns a node list, but then the tokens are converted to nodes and hard to access. I also tried to create a new data structure in the parser called a token list (#7138) but Alex pointed out that initializing those data structures would consume more memory, which didn't seem worth it.

In the end, what we did was change the builder to build `Kwnilarg` nodes from one node rather than two tokens. It uses the location of the first token as the location of the node.

### Test plan

No change in functionality, existing tests should continue to pass.

### DIFF
--- a/parser/Builder.cc
+++ b/parser/Builder.cc
@@ -1119,8 +1119,8 @@ public:
                                      std::move(value));
     }
 
-    unique_ptr<Node> kwnilarg(const token *dstar, const token *nil) {
-        return make_unique<Kwnilarg>(tokLoc(dstar).join(tokLoc(nil)));
+    unique_ptr<Node> kwnilarg(unique_ptr<Node> node) {
+        return make_unique<Kwnilarg>(node->loc);
     }
 
     unique_ptr<Node> kwrestarg(const token *dstar, const token *name) {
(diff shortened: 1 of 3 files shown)


neilparikh:

→ https://go/builds/bui_OHvM2g0NYKnVwl
→ https://go/builds/bui_OHvMq7uns7vS2p
→ https://go/builds/bui_OHvNjdNB2qL7lD

neilparikh (approved):

PR #7154, opened 2023-07-19
Add an internal doc for variance and defaulting

jez:
### Motivation

Follow up from a conversation I had with @maruth-stripe today

### Test plan

See included automated tests.

### DIFF
--- a/docs/variance-defaulting.md
+++ b/docs/variance-defaulting.md
@@ -224,9 +224,14 @@ class Child < Parent
 end
 ```
 
-It no longer makes sense to ask whether `T.class_of(Child) <:
-T.class_of(Parent)`, because those are bare generics. But both `upper` and
-`lower` are hard to stomach as defaults.
+Suddenly, `Parent` and `Child` need to have defaults (because of all the
+situations we had before, like if they appear as class literals in a method
+body, or as the result type of a call to `self.class`). We might want to pick
+defaults such that we maintain the relationship that held previously, namely that
+`T.class_of(Child) <: T.class_of(Parent)`. (Those types are meaningless now that
+those are bare generic classes, but the intuition of wanting it to still hold
+remains). The problem is that both `upper` and `lower` are hard to stomach as
+defaults.
 
 Consider if we pick `upper`. Then we'd have that:
 
--- a/docs/variance-defaulting.md
+++ b/docs/variance-defaulting.md
@@ -234,9 +234,10 @@ Consider if we pick `upper`. Then we'd have that:
 - `Child` has type `T.class_of(Child)[Integer]`
 
 `T.class_of(Child)[Integer]` is not a subtype of `T.class_of(Parent)[Numeric]`
-because `Elem` is invariant! This would be the sound thing to do: it's possible
-to draw a contradiction if we assume that `Child` is an instance of
-`T.class_of(Parent)[Numeric]`:
+because `Elem` is invariant! As unintuitive as this sounds, this subtyping
+outcome is actually the desirable, sound outcome: if we both pick the upper
+bound as the default **and** treat `Child` as a valid value of type
+`T.class_of(Parent)`, it's possible to draw a contradiction:
 
 ```ruby
 # typed: strong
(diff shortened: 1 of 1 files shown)


docs/variance-defaulting.md
```diff
@@ -0,0 +1,375 @@
...
+end
+class Child < Parent
+  extend T::Generic
+  Elem = type_template { {upper: Integer} }
+end
+```
+
+It no longer makes sense to ask whether `T.class_of(Child) <:
+T.class_of(Parent)`, because those are bare generics. But both `upper` and
+`lower` are hard to stomach as defaults.
```
froydnj, line 229: It feels to me like there is a step missing here between the two sentences, which is that if `Child` or `Parent` occur as bare generics (even though those are bad), we need to pick *something* to provide for `Elem`.  WDYT?

docs/variance-defaulting.md
```diff
@@ -0,0 +1,375 @@
...
+    type_parameters(:U)
+      .params(abstract_class: T.class_of(AbstractClass)[AbstractClass,
+      T.type_parameter(:U)])
+      .returns(T.type_parameter(:U))
+  end
+  def example(abstract_class)
+    abstract_class.thing # OK
+  end
+
+  AbstractClass.foo # 💥 bad, untyped
```
froydnj, line 344: `foo` is not a method on `AbstractClass`, did you mean for it to be `example`?

docs/variance-defaulting.md
```diff
@@ -0,0 +1,375 @@
...
+
+So Sorbet gives up and accepts the unsoundness by simply defaulting to
+`T.untyped`.
+
+## Be super careful with unfixed, invariant type members
+
+The takeaway? Type members in well-typed code should always either:
+
+- be fixed, and accept the limitations that come with that (basically: makes the
+  class final in practice), or
```
froydnj, line 300: This "final in practice" aside seems to come out of nowhere.  How does this happen?

froydnj (approved):

docs/variance-defaulting.md
```diff
@@ -0,0 +1,375 @@
...
+  have a suitable replacement if we wanted to enforce something.
+
+  Even more, there's a similar problem to the previous point:
+
+  ```ruby
+  my_box.class.new('')
+  ```
+
+  If `my_box` is `Box[Integer]`, there's no reason why `my_box.class.new` has to
+  create another `Box[Integer]`--making a `Box[String]` is fine!
```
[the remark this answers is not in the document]
jez: for what it's worth, I still don't know what behavior we should pick here. maybe there's some prior art in other languages we could copy here?

docs/variance-defaulting.md
```diff
@@ -0,0 +1,375 @@
...
+substitute for generic types when used in method signatures, etc.
+
+So Sorbet has default rules to silently convert bare generic classes to applied
+generic classes in places where it's unavoidable.
+
+## The defaulting rules
+
+So we've established that bare generic classes are bad, and that bare generic
+classes are also unavoidable. Defaulting rules are how we make the best of a bad
+situation.
```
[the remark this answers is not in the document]
jez: Yeah, I was anticipating wanting to direct link to this section at some point in the future, so I figured I'd recap it.

docs/variance-defaulting.md
jez: it's subtle, I added a footnote

docs/variance-defaulting.md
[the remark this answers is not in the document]
jez: i think the ideal solution here would be some way treat `new` on a generic class as if it had type `[T](x: T) -> Box[T]` like mentioned in the footnote. it's weird right now, because users specify the type of the constructor instead by specifying the type of the `initialize` method, where a `type_member` like `Elem` is in scope, and where the return type of the `initialize` method is always `void`.

I have chatted with dmitry about clever ways to make use of substitution to attempt to work around this but they all have varying levels of hackiness.

The raw notes from that conversation [are here](https://docs.google.com/document/d/1PDn3T4NeVsH0VMz4GdneqgUw6LcfenhrtD8bfOO8hek/edit)

jez:
i believe i addressed all your requests for rewording clarifications. i'm going to automerge, but feel free to either sneak in a request changes or, if you're too late, leave feedback and i'll get it as a follow up

PR #7162, opened 2023-07-20
try enabling yjit

linzhang-stripe:
### Test plan

See included automated tests.


jez:
The bazel build files for Ruby have been deleted and Stripe has migrated its custom build of Ruby into its internal monorepo.

PR #7163, opened 2023-07-20
Better typing for exception variables

jez:
### Summary of new errors

There were two classes of errors that I saw when fixing errors in prep for landing this PR:

- Pinning errors that looked like this:

  ```ruby
  begin
    raises_a
  rescue A => e
  end
  
  2.times do
    begin
      puts(e.class)
      raises_b
    rescue B => e
    end
  end
  ```

  This is a pinning error, because `A` and `B` are different exception
  types. The result of `puts(e.class)` would change between the first and
  second iterations of the loop (from `A` to `B`).
  
  This situation is relatively rare (happens < 100 times in all of
  Stripe's codebase), and the fix is pretty easy (just pick a different
  name for the variable).

- Errors where Sorbet correctly detects that `e` is uninitialized after the `begin`/`rescue` blocks ends (because all exceptional control flow paths also unconditionally raised or returned).

  These indicate logic errors and should be fixed or silenced like normal.

### Motivation

This does not yet completely fix the problem of typing exceptions, but it makes them better in a few ways:

- It does not handle clauses that have `ensure`, due to some weird limitations in Sorbet's control flow handling.
- We no longer consider `e` to be `T.untyped` after the `begin`/`rescue` block ends. Now, it's always typed as `T.nilable(StandardError)` (or something more specific if the rescue clause was more specific.
- Due to the way things shook out in inference, the loc that we used to respond to an LSP hover query with always matched the `<get-current-exception>` type before the `e` type when hovering over the `=> e`. Now you can hover over `=> e` and get the right type.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/cfg/rescue_complex.rb
+++ b/test/testdata/cfg/rescue_complex.rb
@@ -42,7 +42,7 @@ class TestRescue
       baz
     end
 
-    T.reveal_type(baz) # error: Revealed type: `T.untyped`
+    T.reveal_type(baz) # error: Revealed type: `T.nilable(T.any(LoadError, SocketError))`
   end
 
   def rescue_loop()
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -829,8 +829,7 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                     ENFORCE(local != nullptr, "rescue case var not a local?");
 
                     auto localVar = cctx.inWhat.enterLocal(local->localVariable);
-                    rescueHandlersBlock->exprs.emplace_back(localVar, rescueCase->var.loc(),
-                                                            make_insn<Ident>(exceptionValue));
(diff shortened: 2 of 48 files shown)


jez:

→ https://go/builds/bui_OIhPia3mcXCptC
→ https://go/builds/bui_OIhPdKSEpYQCsB
→ https://go/builds/bui_OIhPWFVemb4C9e

froydnj:
This is the same change as #6937 ?

jez:

→ https://go/builds/bui_Ogg4UsMXCfuIr1
→ https://go/builds/bui_Ogg4GtnwOExEDA
→ https://go/builds/bui_Ogg4Ky83RRWcBf

test/testdata/cfg/rescue_complex.rb
```diff
@@ -42,7 +42,7 @@ def multiple_rescue_classes_varuse()
       baz
     end
 
-    T.reveal_type(baz) # error: Revealed type: `T.untyped`
+    T.reveal_type(baz) # error: Revealed type: `T.nilable(T.any(LoadError, SocketError))`
```
froydnj, line 45: Yay!

froydnj (approved):

PR #7166, opened 2023-07-21
Move sig_finder out of main

jez:
### Motivation

It's weird to have definition_validator depend on main. The only reason why
it worked was because main/sig_finder/ was actually it's own package.

### Test plan

See included automated tests.


neilparikh (approved):
need to format, lgtm otherwise

PR #7173, opened 2023-07-25
prototype: linked generic type pairs

jez:
**Not meant to be committed, just wanted to save my work**

I do not know if/when I will have time to return to this.

- - - - -

Hacky, because it hard-codes the `LinkedGenericTypePair` name, and uses
`class << self` tricks to get the constants to resolve correctly

Any final solution here is likely going to have to come up with a clever
trick for how to get

```ruby
sig {returns(X)}
```

to resolve as either the `type_member` or the `type_template` depending
on whether the sig is on top of an instance or singleton class method.

(In my examples, I was abusing `class << self` to get this to resolve,
just to test that the calls.cc intrinsics work as expected.)

Some strategies for getting the syntax to parse, none of which I'm
attached to:

- desugar to `class << self`, and require people to use `class << self`
  to use that type in a sig (confusing)
- hijack constant resolution somehow?
- at the point where we would have reported a "you're using a
  type_member / type_template from the wrong scope" error, simply change
  it to the linked symbol if necessary
- find a clever way to use UniqueName's, so that you can rewrite without
  having to use `class << self`
- attempt to build some first-party support for these linked pairs,
  which would involve changing FoundDefinitions and namer and resolver.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.h
+++ b/core/Symbols.h
@@ -298,12 +298,25 @@ public:
         bool isContravariant : 1;
         bool isFixed : 1;
 
-        constexpr static uint8_t NUMBER_OF_FLAGS = 6;
+        // Type-member-specific flags
+        bool isLinkedGenericPair_ : 1;
+
+        constexpr static uint8_t NUMBER_OF_FLAGS = 7;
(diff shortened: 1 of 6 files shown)


jez:
#7847

PR #7176, opened 2023-07-26
"Copy Symbol to Clipboard" shows a progress dialog when needed

damolina-stripe:
**Issue**
The `sorbet/showSymbol`  [request](https://github.com/sorbet/sorbet/blob/166024b5c9688ea0053ded271485ad123c8aa337/vscode_extension/src/commands/copySymbolToClipboard.ts#L49) does not resolve until Sorbet can actually process the request. This can take more than a few seconds causing unexpected results as user tries the command, "nothing happens" and so they try again (or try a bit later at a different place). The requests eventually resolve, overriding the clipboard contents in whatever order they come back.  

**Fix**
Update "Copy Symbol to Clipboard" command so it shows a progress dialog when Sorbet's status is not `Idle` (otherwise it won't). Without UI, there is no way to know that there are pending requests, so this change introduces a progress dialog whenever Sorbet is not ready. 
- Whenever Sorbet is ready to process a request,  no notification is presented at all .
- Adding UI should help users understand an operation is pending. This notification is cancelable, so operations can be safely discarded (see not below about `LanguageClient.sendRequest`). 
- Ideally, the command would only allow one pending operation but that introduced more complexity on the PR (and need to figure out quirks about UX), so decided to do that as follow-up work (note that today, you can queue up multiple requests unknowingly, so this is just making those requests visible).

<p align=center>
image
</p>

**Misc**
- Found out that even though [LanguageClient.request](https://github.com/sorbet/sorbet/blob/166024b5c9688ea0053ded271485ad123c8aa337/vscode_extension/src/languageClient.ts#L247) takes in `CancellationToken`, it actually does not process any cancellation request.   This meant that after clicking on `Cancel`, the notification would go away but the request would keep going and at some point, the result would come back and override the clipboard.  This case is handled correctly on this PR, but to prevent mistakes, this PR removes the ability to pass a `CancellationToken` to the `sendRequest` method.
  - We need to verify if cancellation requests to the LSP are supported.

- Some logging entries in the command implementation were not ideal (e.g. they would report that the "current" Sorbet LSP client did not support the "showSymbol" capability when in reality there was not client to talk to), so they are updated.

### Test plan

See included automated tests.

### DIFF
--- a/vscode_extension/src/commands/copySymbolToClipboard.ts
+++ b/vscode_extension/src/commands/copySymbolToClipboard.ts
@@ -1,4 +1,4 @@
-import { env, window } from "vscode";
+import { env, ProgressLocation, window } from "vscode";
 import {
   SymbolInformation,
   TextDocumentPositionParams,
@@ -14,29 +14,35 @@ export async function copySymbolToClipboard(
   context: SorbetExtensionContext,
 ): Promise<void> {
   const { activeLanguageClient: client } = context.statusProvider;
-  if (client?.status !== ServerStatus.RUNNING) {
-    context.log.warning("Sorbet LSP client is not ready.");
+
+  if (!client) {
+    context.log.warning("CopySymbol: No active Sorbet LSP.");
     return;
   }
 
   if (!client.capabilities?.sorbetShowSymbolProvider) {
     context.log.warning(
-      'Sorbet LSP client does not support "show symbol" capability.',
+      "CopySymbol: Sorbet LSP does not support 'showSymbol' capability.",
     );
     return;
   }
 
   const editor = window.activeTextEditor;
   if (!editor) {
-    context.log.debug("No active editor to copy symbol from.");
+    context.log.debug("CopySymbol: No active editor, no target symbol.");
     return;
   }
 
   if (!editor.selection.isEmpty) {
     context.log.debug(
-      "Cannot determine target symbol from a non-empty selection.",
+      "CopySymbol: Non-empty selection, cannot determine target symbol.",
     );
-    return; // something is selected, abort
+    return;
+  }
+
+  if (client.status !== ServerStatus.RUNNING) {
+    context.log.warning("CopySymbol: Sorbet LSP is not ready.");
+    return;
   }
 
   const position = editor.selection.active;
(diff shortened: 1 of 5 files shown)


vscode_extension/src/commands/copySymbolToClipboard.ts
```diff
@@ -46,13 +52,42 @@ export async function copySymbolToClipboard(
...
-    params,
-  );
-
-  await env.clipboard.writeText(response.name);
-  context.log.debug(
-    `Copied symbol name to the clipboard. Name:${response.name}`,
-  );
+
+  let response: SymbolInformation | undefined;
+  if (context.statusProvider.operations.length) {
```
ilyailya, line 57: What do we check with this condition? Do we check if there are any other requests in flight?

damolina-stripe: > Ideally, the command would only allow one pending operation but that introduced more complexity on the PR (and need to figure out quirks about UX), so decided to do that as follow-up work (note that today, you can queue up multiple requests unknowingly, so this is just making those requests visible).

jez: The showSymbol request should be able to complete instantly if the operation is "Typechecking in background..."

Until now, I'm pretty sure that the "Typechecking..." / "Indexing..." / etc. states have all been UI-only. I don't know that the correctness of any features depended on them until this. That suggests to me that we might be solving this at the wrong level of abstraction. That's why I was suggesting possibly implementing this inside Sorbet's LSP mode. That would probably be cleaner, but I think it would require a somewhat widely scoped refactor of some parts of Sorbet's pipeline, it'll have to wait for some other day.

In the mean time, if you hard-code Type checking in background here as not needing a progress bar, that should be fine. Or honestly, you could just show the progress bar even if it was typechecking in background, because it would just go away after a short moment.

jez (approved):

vscode_extension/src/commands/copySymbolToClipboard.ts
damolina-stripe: I'll keep tweaking the experience since even this does not solve all UX issues. As @ilyailya noted, you can queue up multiple operations, which has always been true, but now you might get some UI showing that (making more evident but also noisier). I will look at limiting this when `Typechecking in background...`, as you suggested, in a latter PR.

PR #7181, opened 2023-07-28
Don't duplicate vector when calling pipeline::index

jez:
### Motivation

Slight memory and performance optimization. In a codebase with 250,000 files,
this will save about 1 MB, because a FileRef is 4 bytes. So in all likelihood,
this won't translate to a noticeable performance improvement. But it's still
nice to have.

This change is safe because `indexSuppliedFiles` in `pipeline.cc` is already
`move`'ing each FileRef out of the vector to enqueue it, which ensures that no
one after a call to `pipeline::index` happens could be relying on the elements
still being there.

### Test plan

Existing tests


jez:
Going to go another direction.

PR #7188, opened 2023-07-28
Index all `__package.rb` files before everything else

jez:
### Motivation

We want to start using the package graph for the sake of skipping some
work, which means we have to make sure that package files are indexed
before we do anything else. Eventually a second step will be to both
index **and** parse the package DB before we do anything else, and I
will do that in a follow up change.

But for the time being, I want to start slowly and make sure nothing is
breaking.

This _should_ be a no-op change, though it will involve one more fork +
join step, which might make Sorbet slower on pay-server because of this.
While that's not great, we should recover that time and more when we
start using this to type check a subset of the codebase.

It's also worth noting that we will need to make corresponding changes
to LSP eventually. Those will definitely come, but for now I plan to
experiment mostly with the command line mode of Sorbet.

### Test plan

I think this should be covered by existing tests.
There should not be any new behavior in this test.

I also plan for Stripe's codebase to be the main arbiter of this change.

### DIFF
--- a/core/FileRef.h
+++ b/core/FileRef.h
@@ -47,6 +47,11 @@ public:
     const File &dataAllowingUnsafe(const GlobalState &gs) const;
     File &dataAllowingUnsafe(GlobalState &gs) const;
 
+    // Normally, using .data requires that the file have been read.
+    // But File::Flag data is populated when the File is created, before needing to read the
+    // file. This helper exposes that without tripping an ENFORCE.
+    bool isPackage(const GlobalState &gs) const;
+
 private:
     uint32_t _id;
 };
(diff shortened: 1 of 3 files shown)


jez:

→ https://go/builds/bui_OLfTihh0p2Uohm
→ https://go/builds/bui_OLfTTmu8pqIbPa
→ https://go/builds/bui_OLfT6rZM4hVY0M

ilyailya (approved):

jez:
When I test this on Stripe's codebase I get essentially no appreciable difference in timing:

http://go/gist/jez/85bd9b3fabd71b41e9e1324ae187da26

PR #7191, opened 2023-07-31
Add paths option to autogen subclasses generation

youssefsoli:
In Stripe’s codebase, we would like to be able to determine which file edits can cause differences in certain code generators. These generators usually rely on subclasses output from Sorbet, however, there is no path information for where the constants are located. This change brings support for exactly that.

### Motivation

Currently when the autogen subclasses output is generated, we get the type of the Constant (module/class) and its fully qualified name. During this step, we can extract the path of the constant for free and use it as part of the final output.

### Test plan

See included automated tests.


main/autogen/subclasses.cc
```diff
@@ -63,7 +63,7 @@ optional<Subclasses::Map> Subclasses::listAllSubclasses(core::Context ctx, Parse
                                             "::", [&ctx](const core::NameRef &nm) -> string { return nm.show(ctx); }));
 
         auto &mapEntry = out[parentName];
-        mapEntry.entries.insert(make_pair(childName, defn.data(pf).type));
+        mapEntry.entries.insert(make_tuple(childName, defn.data(pf).type, pf.path));
```
aadi-stripe, line 66: To allow for better readability, given we have one more extra field now, I'd be more comfortable with this being a struct rather than a tuple, like `struct DescendantsMapEntry` or something similar. Not sure if the core Sorbet folks agree.

main/autogen/subclasses.cc
```diff
@@ -113,13 +113,21 @@ vector<string> Subclasses::serializeSubclassMap(const Subclasses::Map &descendan
         auto type = children.classKind == ClassKind::Class ? "class" : "module";
         descendantsMapSerialized.emplace_back(fmt::format("{} {}", type, parentName));
 
+        const auto classFormatString = showPaths ? " class {} {}" : " class {}";
+        const auto moduleFormatString = showPaths ? " module {} {}" : " module {}";
```
aadi-stripe, line 117: Let's hoist these out of the loop since they seem the same regardless of `parentName`.

aadi-stripe (changes requested):
I would also sanity check and trace autogen perf to make sure runtime isn't regressed. You can do this by (there are more efficient ways, but have non-trivial complexity and are not worth adding complexity for a sanity check):

* Cleaning out `build/` and `dev/out/sorbet-cache` to get a clean state.
* `pay stop autogen`
* `rm -rf /tmp/trace.json && AUTOGEN_TRACE='/tmp/trace.json' scripts/bin/autogen`.

You can then download the trace and load in the chrome://tracing viewer. Check before and after your changes.

main/autogen/subclasses.cc
froydnj: This core Sorbet folk has a strong preference for named structs over against pairs and tuples.

I think the main challenge with converting to structs will be writing the necessary overloads to work with abseil's hash tables.

youssefsoli: The entries are stored in an unordered set so it does unfortunately.

youssefsoli:
So after doing some profiling, the perf for the dependency step slows down 2x, I will investigate preventing new string allocations for each file path.

main/autogen/subclasses.cc
aadi-stripe: Ah

aadi-stripe:
> So after doing some profiling, the perf for the dependency step slows down 2x, I will investigate preventing new string allocations for each file path.

Ack, yeah that'll need to be fixed in order for this to be merge-able.

main/autogen/subclasses.cc
froydnj: You should be able to cargo-cult from https://github.com/sorbet/sorbet/blob/40a70d0acb9f5f5435808e12098693a129add54f/core/LocalVariable.h#L76-L78 for converting this into a struct.

You also -- I think -- don't have to hash the path, nor do you have to store the path as a string (assuming that the path is coming from `GlobalState`'s storage), which should help with the slowdown issues.

youssefsoli:
Alright I have updated this PR to be much more effecient with memory usage by relying on `SymbolRef`s and `autogen::Definition`s instead of strings that were constantly copied around when generating the subclasses.txt file. We go from a total time of `44s` to `17s` overall, and `2s` to `1s` in the `Timer timeit(logger, "autogen");` block.

main/autogen/subclasses.cc
```diff
@@ -52,18 +65,8 @@ optional<Subclasses::Map> Subclasses::listAllSubclasses(core::Context ctx, Parse
...
-
-        // Add child class to the set identified by its parent
-        string childName =
-            fmt::format("{}", fmt::map_join(pf.showFullName(ctx, defn),
-                                            "::", [&ctx](const core::NameRef &nm) -> string { return nm.show(ctx); }));
-
-        auto &mapEntry = out[parentName];
-        mapEntry.entries.insert(make_pair(childName, defn.data(pf).type));
+        auto &mapEntry = out[ref.sym];
+        mapEntry.entries.insert({defn.data(pf), pf.file});
```
aadi-stripe, line 69: So instead of storing `defn.data(pf)` which is a struct, I'd actually just store `defn.data(pf).sym` here. Save even more memory allocation and copying. In which case, you won't even need to store `pf.file`, because `defn.data(pf).sym.loc(ctx)` will be guaranteed to correspond to the right file. Basically it's then just a map of symbolRef -> list of symbolRefs, which should be even more lightweight.

aadi-stripe (changes requested):

main/autogen/subclasses.cc
youssefsoli: but then we would still have to store `pf` somewhere since it gets deleted after the workers are finished.

youssefsoli:
Turns out my performance testing environment was not equivalent for before and after these changes.

Here is the updated timing information for just the `autogen` timing block we care for.

```
without_changes: 820.007 ms
with_initial_changes: 1,016.008 ms
with_latest_changes: 1,020.009 ms
```

main/autogen/subclasses.cc
youssefsoli: Scratch that, I think this optimization is possible

main/autogen/subclasses.cc
```diff
@@ -52,18 +65,8 @@ optional<Subclasses::Map> Subclasses::listAllSubclasses(core::Context ctx, Parse
...
-
-        // Add child class to the set identified by its parent
-        string childName =
-            fmt::format("{}", fmt::map_join(pf.showFullName(ctx, defn),
-                                            "::", [&ctx](const core::NameRef &nm) -> string { return nm.show(ctx); }));
-
-        auto &mapEntry = out[parentName];
-        mapEntry.entries.insert(make_pair(childName, defn.data(pf).type));
+        auto &mapEntry = out[ref.sym];
+        mapEntry.entries.insert({defn.data(pf).sym, defn.data(pf).type});
```
aadi-stripe, line 69: Sorry for the back-and-forth, but we can do even better. `type` is stored on `sym`, because you can check `isClassOrModule()` on it.

aadi-stripe (changes requested):

main/autogen/subclasses.cc
youssefsoli: So SymbolRefs have
```
    enum class Kind : uint8_t {
        ClassOrModule = 0,
        Method = 1,
        FieldOrStaticField = 2,
        TypeArgument = 3,
        TypeMember = 4,
    };
 ```
 but it does not distinguish between classes or modules, that is inferred here https://github.com/sorbet/sorbet/blob/3a7d0cf064c03921774c672461b6b9593451fc19/main/autogen/autogen.cc#L75

Comment (aadi-stripe) [reply main/autogen/subclasses.cc:69]: Oh nm, we need to disambiguate further here. Sorry. My bad.

Comment (aadi-stripe) [reply main/autogen/subclasses.cc:69]: @youssefsoli I think there's `isClass` / `isModule` methods on `sym.data(ctx)` which will do this.

**Review by aadi-stripe — CHANGES_REQUESTED** [review]

**Comment by youssefsoli** [comment]
Nice, that does work. Now we only rely on SymbolRefs

File: main/autogen/subclasses.cc [hunk 5]
```diff
@@ -92,34 +95,63 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
     return SubclassInfo(fnd->second.classKind, std::move(out));
 }
 
+const core::SymbolRef Subclasses::getConstantRef(core::GlobalState &gs, string rawName) {
+    core::UnfreezeNameTable nameTableAccess(gs);
+    core::UnfreezeSymbolTable symbolTableAccess(gs);
+    core::ClassOrModuleRef sym = core::Symbols::root();
+
+    for (auto &n : absl::StrSplit(rawName, "::")) {
+        sym = gs.enterClassSymbol(core::Loc(), sym, gs.enterNameConstant(gs.enterNameUTF8(n)));
```
jez, line 104: What do you think about using `const core::GlobalState &gs` everywhere, and using `lookupClassSymbol` everywhere?

This code all seems to me like it shouldn't need mutable access to GlobalState.

youssefsoli: Sounds good, just added a change to make it constant everywhere

main/autogen/subclasses.cc
```diff
@@ -92,34 +95,64 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
...
         auto type = children.classKind == ClassKind::Class ? "class" : "module";
         descendantsMapSerialized.emplace_back(fmt::format("{} {}", type, parentName));
 
         auto subclassesStart = descendantsMapSerialized.size();
-        for (const auto &[name, type] : children.entries) {
+        for (const auto &sym : children.entries) {
+            string_view path = sym.loc(gs).file().data(gs).path();
+            // Strip initial "./" from path
+            if (path.find("./") == 0)
+                path = path.substr(2);
```
jez, line 143:
```suggestion
            string_view path = gs.getPrintablePath(sym.loc(gs).file().data(gs).path());
```

`GlobalState::getPrintablePath` respects Sorbet's `--remove-path-prefix` option, which allows stripping arbitrary path prefixes (e.g., it would work if you invoked Sorbet on `--dir=lib`, not just `--dir=.`.

jez: We already pass this option in all the relevant places in Stripe's codebase (as far as I can tell). You may wish to double check this anwyays.

main/autogen/subclasses.cc
```diff
@@ -92,34 +95,64 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
...
+
+            string childName =
+                fmt::format("{}", fmt::map_join(symbolName(gs, sym), "::", [&gs](const core::NameRef &nm) -> string {
+                                return nm.show(gs);
+                            }));
```
jez, line 148:
```suggestion

            string childName = sym.show(gs);
```

I think you're overcomplicating this, but let me know if there's something I'm missing.

main/autogen/subclasses.cc
```diff
@@ -92,34 +95,64 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
...
+    for (const auto &parentRef : parentNames) {
+        auto fnd = descendantsMap.find(parentRef);
+        if (fnd == descendantsMap.end() || !parentRef.exists()) {
             continue;
         }
         const Subclasses::SubclassInfo &children = fnd->second;
 
+        string parentName =
+            fmt::format("{}", fmt::map_join(symbolName(gs, parentRef),
+                                            "::", [&gs](const core::NameRef &nm) -> string { return nm.show(gs); }));
```
jez, line 133:
```suggestion
        string parentName = parentRef.show(gs);
```

main/autogen/subclasses.cc
aadi-stripe: ➕ just `sym.show` should be good enough.

main/autogen/subclasses.cc
```diff
@@ -34,8 +36,19 @@ bool Subclasses::isFileIgnored(const std::string &path, const std::vector<std::s
...
+vector<core::NameRef> Subclasses::symbolName(const core::GlobalState &gs, core::SymbolRef sym) {
+    vector<core::NameRef> out;
+    while (sym.exists() && sym != core::Symbols::root()) {
+        out.emplace_back(sym.name(gs));
+        sym = sym.owner(gs);
+    }
+    reverse(out.begin(), out.end());
+    return out;
+}
+
```
jez, line 49:
```suggestion
```

main/autogen/subclasses.h
```diff
@@ -7,29 +7,33 @@ namespace sorbet::autogen {
...
                               const std::vector<std::string> &relativeIgnorePatterns);
-    static std::optional<SubclassInfo> descendantsOf(const Subclasses::Map &childMap, const std::string &parents);
-    static std::vector<std::string> serializeSubclassMap(const Subclasses::Map &descendantsMap,
-                                                         const std::vector<std::string> &parentNames);
+    static std::optional<SubclassInfo> descendantsOf(const Subclasses::Map &childMap, const core::SymbolRef &parentRef);
+    static std::vector<std::string> serializeSubclassMap(const core::GlobalState &gs,
+                                                         const Subclasses::Map &descendantsMap,
+                                                         const std::vector<core::SymbolRef> &parentNames,
+                                                         const bool showPaths);
+    static std::vector<core::NameRef> symbolName(const core::GlobalState &gs, core::SymbolRef sym);
```
jez, line 36:
```suggestion
```

main/autogen/subclasses.cc
youssefsoli: Is there a way to get the full path? `getPrintablePath` returns the filename.

main/autogen/subclasses.cc
```diff
@@ -140,30 +173,31 @@ vector<string> Subclasses::serializeSubclassMap(const Subclasses::Map &descendan
...
-    fast_sort(parentNames);
+    fast_sort(parentRefs, [&gs](const auto &left, const auto &right) { return left.show(gs) < right.show(gs); });
     Subclasses::Map descendantsMap;
-    for (const string &parentName : parentNames) {
+    for (const auto &parentRef : parentRefs) {
         // Skip parents that the user asked for but which don't
         // exist or are never subclassed.
-        auto fnd = childMap.find(parentName);
-        if (fnd == childMap.end()) {
+        auto fnd = childMap.find(parentRef);
```
aadi-stripe, line 186: You can move the `parentRef.exists()` check to before the map lookup.

main/autogen/subclasses.cc
jez: I'm not entirely sure what you mean

https://github.com/sorbet/sorbet/blob/dfd20f3504911c66ce07b0993144b6e7904ce0b9/core/GlobalState.cc#L2221-L2227

The implementation of this method looks basically identical to the code I this `suggestion` replaces?

youssefsoli: Never mind, the assumption is that the passed in filepath is the root.

main/autogen/subclasses.cc
aadi-stripe, line 133: `parentRef.show(gs)` should do the trick here, no need to `map_join`.

aadi-stripe (changes requested):

main/autogen/subclasses.cc
```diff
@@ -92,34 +95,64 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
...
+                fmt::format("{}", fmt::map_join(symbolName(gs, sym), "::", [&gs](const core::NameRef &nm) -> string {
+                                return nm.show(gs);
+                            }));
+
             // Ignore Modules
-            if (type == autogen::Definition::Type::Class) {
-                descendantsMapSerialized.emplace_back(fmt::format(" class {}", name));
+            if (sym.asClassOrModuleRef().data(gs)->isClass()) {
+                // Note: fmt ignores excess arguments
+                descendantsMapSerialized.emplace_back(fmt::format(classFormatString, childName, path));
```
jez, line 153: I don't know if this is intentional--in fact I think a future change to the fmt library will change this behavior.

Let's do something like this:

```suggestion
                descendantsMapSerialized.emplace_back(fmt::format(" class {} {}", childName, showPaths ? path : ""));
```

main/autogen/subclasses.cc
```diff
@@ -140,30 +173,31 @@ vector<string> Subclasses::serializeSubclassMap(const Subclasses::Map &descendan
...
+    fast_sort(parentRefs, [&gs](const auto &left, const auto &right) { return left.show(gs) < right.show(gs); });
     Subclasses::Map descendantsMap;
-    for (const string &parentName : parentNames) {
+    for (const auto &parentRef : parentRefs) {
         // Skip parents that the user asked for but which don't
         // exist or are never subclassed.
-        auto fnd = childMap.find(parentName);
-        if (fnd == childMap.end()) {
+        auto fnd = childMap.find(parentRef);
+        if (fnd == childMap.end() || !parentRef.exists()) {
```
jez, line 187: This `.exists()` looks a little suspect. I haven't traced the code thoroughly, but I'm guessing that we will want to make it so that if a parentRef doesn't exist, it's never even added to parentRefs in the first place.

Do you think we could maintain that invariant?

I'm _guessing_ that this is happening because your getConstantRef unconditionally inserts into the vector even if attempting to look up the hard-coded class didn't exist. If that's what's causing us to need this .exists here, I think we should only `emplace_back` into the parentRefs if getConstantRef actually found a suitable symbol.

main/autogen/subclasses.cc
```diff
@@ -92,34 +95,64 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
...
     vector<string> descendantsMapSerialized;
+    const auto classFormatString = showPaths ? " class {} {}" : " class {}";
+    const auto moduleFormatString = showPaths ? " module {} {}" : " module {}";
 
-    for (const string &parentName : parentNames) {
-        auto fnd = descendantsMap.find(parentName);
-        if (fnd == descendantsMap.end()) {
+    for (const auto &parentRef : parentNames) {
+        auto fnd = descendantsMap.find(parentRef);
+        if (fnd == descendantsMap.end() || !parentRef.exists()) {
```
jez, line 126: ditto this `.exists()` call.

main/realmain.cc
```diff
@@ -301,17 +301,22 @@ void runAutogen(const core::GlobalState &gs, options::Options &opts, const autog
                 continue;
             }
 
-            for (const auto &[parentName, children] : *el.subclasses) {
-                if (!parentName.empty()) {
-                    auto &childEntry = childMap[parentName];
+            for (const auto &[parentRef, children] : *el.subclasses) {
+                if (parentRef.exists()) {
```
jez, line 305: Do you think we could also remove this .exists() call? I'm not sure what the `.empty()` check was there for before, so I'm not sure if this is caused by something else.

jez (commented):
Done leaving comments for now. It looks pretty good at a high level, but because there are so many requested changes I'd love another look at it.

youssefsoli:
Thanks for the review, the code looks much cleaner now.

main/autogen/subclasses.cc
```diff
@@ -92,34 +84,57 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
...
+
         auto type = children.classKind == ClassKind::Class ? "class" : "module";
         descendantsMapSerialized.emplace_back(fmt::format("{} {}", type, parentName));
 
         auto subclassesStart = descendantsMapSerialized.size();
-        for (const auto &[name, type] : children.entries) {
+        for (const auto &sym : children.entries) {
+            string_view path = gs.getPrintablePath(sym.loc(gs).file().data(gs).path());
+            string childName = sym.show(gs);
             // Ignore Modules
```
aadi-stripe, line 132: Not sure why this comment existed.

aadi-stripe (changes requested):
Looks good, we're getting closer. Just one more comment.

main/options/options.cc
```diff
@@ -785,6 +787,13 @@ void readOptions(Options &opts,
             throw EarlyReturnWithCode(1);
         }
 
+        opts.autogenSubclassesShowPaths = raw["autogen-subclasses-show-paths"].as<bool>();
```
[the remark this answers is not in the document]
youssefsoli: Sounds good, autogen is backwards compatible out of the box, it just has one invariant check that makes sure there are exactly two strings per line but we can deal with this by bumping autogen and sorbet in the same PR

main/autogen/subclasses.cc
```diff
@@ -92,34 +84,57 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
...
-            if (type == autogen::Definition::Type::Class) {
-                descendantsMapSerialized.emplace_back(fmt::format(" class {}", name));
+            if (sym.asClassOrModuleRef().data(gs)->isClass()) {
```
jez, line 133: Seeing this `asClassOrModuleRef` suggests to me that we wish `children.entries` had type `UnorderedSet<core::ClassOrModuleRef>` instead of `UnorderedSet<core::SymbolRef>`.

There are a handful of symbols that are constants but not classes/modules: static constant assignments like `X = 1` and type members like `X = type_member`.

It seems that right now, the code depends on `parent_of.exists()` implying that `parent_of.data(pf).sym.isClassOrModule()`. I think that's a safe assumption?

In that case, I'd prefer to move the invariant closer to construction, namely moving this `asClassOrModuleRef` to `listAllSubClasses`.

main/options/options.cc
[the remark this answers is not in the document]
aadi-stripe: Yeah that's reasonable to do.

main/autogen/subclasses.cc
youssefsoli: Yeah that doesn't make sense, seems to have been left in.

main/autogen/subclasses.cc
jez: Screenshot 2023-08-09 at 4 06 31 PM

jez (approved):
looks good with a few more suggestions.

main/autogen/subclasses.cc
aadi-stripe: ooh yes, that's even better.

main/autogen/subclasses.cc
```diff
@@ -92,34 +84,57 @@ optional<Subclasses::SubclassInfo> Subclasses::descendantsOf(const Subclasses::M
...
+                                            childMap[riskSafeMachineRef].entries.end());
 }
 
-vector<string> Subclasses::serializeSubclassMap(const Subclasses::Map &descendantsMap,
-                                                const vector<string> &parentNames) {
+vector<string> Subclasses::serializeSubclassMap(const core::GlobalState &gs, const Subclasses::Map &descendantsMap,
+                                                const vector<core::SymbolRef> &parentNames, const bool showPaths) {
     vector<string> descendantsMapSerialized;
+    const auto classFormatString = showPaths ? " class {} {}" : " class {}";
+    const auto moduleFormatString = showPaths ? " module {} {}" : " module {}";
```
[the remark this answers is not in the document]
youssefsoli: We would end up with an extra space character at the end, so I guess doing it inline might make more sense now.

aadi-stripe:
Also a sanity check -- if you haven't been doing this already, make sure the generated `subclasses.txt` file is unchanged :)

main/options/options.cc
aadi-stripe, line 790: I think we want to remove the flag

aadi-stripe (changes requested):

main/options/options.cc
youssefsoli: Nice catch, forgot to change that

aadi-stripe (approved):

jez (approved):

PR #7192, opened 2023-07-31
Add concrete type to disallow argument-less `Proc.new`

aisamanra:
### Motivation
In Ruby 3, `Proc.new` with no argument is an error. This types the method so that it should flag a static error if no block is provided.

This also updates the relevant docs and fixes the one test that seemed to have failed.

### Test plan

See included automated tests.


jez (approved):
Separately I wonder why we have `self.new` here instead of `initialize`.

PR #7195, opened 2023-08-01
Show demangled package name in blame output

jez:
### Motivation

We have other data sources within Stripe that use the demangled package
name. The mangled name is an artifact of Sorbet's internals, not a
public API.

### Test plan

We do not have tests for this. I am assuming that if it compiles it works.

I did trace the code to make sure that PackageInfo::show shows the demangled name.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -1293,8 +1293,8 @@ void printUntypedBlames(const core::GlobalState &gs, const UnorderedMap<long, lo
         writer.String("package");
         if (sym.exists() && sym.loc(gs).exists()) {
             const auto file = sym.loc(gs).file();
-            const auto pkg = gs.packageDB().getPackageNameForFile(file);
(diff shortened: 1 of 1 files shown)


jez:
cc @maruth-stripe

main/pipeline/pipeline.cc
```diff
@@ -1293,8 +1293,8 @@ void printUntypedBlames(const core::GlobalState &gs, const UnorderedMap<long, lo
         writer.String("package");
         if (sym.exists() && sym.loc(gs).exists()) {
             const auto file = sym.loc(gs).file();
-            const auto pkg = gs.packageDB().getPackageNameForFile(file);
-            if (pkg == core::NameRef::noName()) {
+            const auto &pkg = gs.packageDB().getPackageForFile(gs, file);
+            if (pkg.exists()) {
```
maruth-stripe, line 1297: is this flipped?

ilyailya (approved):

PR #7196, opened 2023-08-01
Add additional sig for caller_locations

ipc103:
### Motivation

Similar to the `caller` method, `caller_locations` should only be nilable if a parameter is passed. I added an additional `sig` to indicate this so we don't need to use safe navigation unless a paramater is passed through. 

### Test plan

I removed the instances I found of using safe navigation on `caller_locations` when no parameters are passed, and added tests to `test/testdata/rbi/kernel.rb` per the comment below! 🙏

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -163,7 +163,7 @@ module T::Private::Methods
 
         definition_file, definition_line = T::Private::Methods.signature_for_method(ancestor.instance_method(method_name)).method.source_location
         is_redefined = target == ancestor
-        caller_loc = caller_locations&.find {|l| !l.to_s.start_with?(SORBET_RUNTIME_LIB_PATH)}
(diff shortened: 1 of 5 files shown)


froydnj (changes requested):
Thanks for the fix!  It looks like you have some test failures because the RBI changes changed line numbers of where errors are reported.  Can you run:

```
./bazel test //test/cli:update
```

and commit + push those changes?

We generally don't require tests for RBI changes, but if you are so inclined, you can find examples to copy from in `test/testdata/rbi/kernel.rb`.  (You will want to add them prior to this line: https://github.com/sorbet/sorbet/blob/master/test/testdata/rbi/kernel.rb#L90 , for reasons.)

ipc103:
❤️ Thank you, @froydnj!

>  It looks like you have some test failures because the RBI changes changed line numbers of where errors are reported. 

Updated in 3c5f911

>  but if you are so inclined, you can find examples to copy from in `test/testdata/rbi/kernel.rb`

Added in 2db111a - thanks for the pointer

PR #7203, opened 2023-08-03
Change loc for `&&` and `||` condition

jez:
### Motivation

The loc for the local variable we used in the `cond` of the `If` that
`&&` and `||` desugar into was previously the whole LHS. In the case of
a long expression on the LHS, that meant that Sorbet would highlight
multiple lines of (irrelevant) code when showing a "This code is
untyped" error.

To my knowledge, the only place in Sorbet where the loc of a local
variable used in an `If` condition would matter is for the error message
about untyped code usages. That is: this loc is not important for any
other autocorrects like adding `T.must` or `T.unsafe`, where the
location of the error is going to get wrapped with a method call.

That means we're free to change this however we like. With that in mind,
I'd like to point to the `... || ...` instead, which makes it a little
more obvious where the "conditional branch on untyped" is coming from.

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1167,9 +1167,12 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                         thenp = std::move(rhs);
                     }
                     auto lhsLoc = lhs.loc();
+                    auto condLoc = lhsLoc.exists() && thenp.loc().exists()
+                                       ? core::LocOffsets{lhsLoc.endPos(), thenp.loc().beginPos()}
+                                       : lhsLoc;
                     auto temp = MK::Assign(loc, andAndTemp, std::move(lhs));
                     auto iff =
-                        MK::If(loc, MK::Local(lhsLoc, andAndTemp), std::move(thenp), MK::Local(lhsLoc, andAndTemp));
+                        MK::If(loc, MK::Local(condLoc, andAndTemp), std::move(thenp), MK::Local(lhsLoc, andAndTemp));
                     auto wrapped = MK::InsSeq1(loc, std::move(temp), std::move(iff));
                     result = std::move(wrapped);
                 }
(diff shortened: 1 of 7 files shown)


jez:
* master
  * **PR #7202** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/7202" target="_blank">Graphite</a> 
    * **PR #7203** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/7203" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/7203?utm_source=stack-comment).

jez:
Screenshot 2023-08-02 at 7 08 50 PM

https://sorbet.run/#%23%20typed%3A%20strong%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28lines%3A%20T%3A%3AArray%5BString%5D%29.void%7D%0Adef%20foo%28lines%29%0A%20%20min_space%20%3D%20lines%0A%20%20%20%20.map%20do%20%7Cline%7C%0A%20%20%20%20%20%20if%20line.strip.empty%3F%0A%20%20%20%20%20%20%20%20nil%0A%20%20%20%20%20%20else%0A%20%20%20%20%20%20%20%20count_leading_spaces%28line%29%0A%20%20%20%20%20%20end%0A%20%20%20%20end%0A%20%20%20%20.compact%0A%20%20%20%20.min%20%7C%7C%200%0Aend%0A%0A%0A%0A%23sig%20%7B%20params%28str%3A%20String%29.returns%28Integer%29%20%7D%0Adef%20count_leading_spaces%28str%29%0A%20%200%0Aend

This is what it looked like before—you can see how long the squiggle is.

neilparikh (changes requested):
I think this change broke the "always truthy" error and autocorrect:

```
     test/cli/and-and-always-truthy/test.rb:15: This condition was always `truthy` (`TrueClass`)
     15 |  if always_true || T.unsafe(true)
-             ^^^^^^^^^^^
+                        ^^^^
```

```
   Autocorrect: Done
-    suggest_unsafe_dead.rb:35: Replaced with `T.unsafe(returns_integer)`
+    suggest_unsafe_dead.rb:35: Replaced with `T.unsafe( || )`
     35 |  returns_integer || do_something
-          ^^^^^^^^^^^^^^^
+                         ^^^^
   Got `Integer` originating from:
     suggest_unsafe_dead.rb:35:
     35 |  returns_integer || do_something
@@ -196,7 +196,7 @@
 sig {void}
 def test3_method_or_method

-  T.unsafe(returns_integer) || do_something
+  returns_integerT.unsafe( || )do_something
 end
```

neilparikh (approved):

PR #7205, opened 2023-08-03
Add sig for Rbconfig.ruby to RBIs

rwstauner:
Definition:
https://github.com/ruby/ruby/blob/499eb3990faeaac2603787f2a41b2d9625e180dc/tool/mkconfig.rb#L386-L396


froydnj (approved):
Thank you!

PR #7207, opened 2023-08-03
Update T::Enum#serialize defintion to be public

bitwise-aiden:
Ensure that `T::Enum#serialize` is defined as public

### Motivation
The recent `T::Enum#serialize` change adds the definition for serialize to the end of the enum, so if the class use te `private` keyword, the serialize method will also be private.

```ruby
class MyEnum < T::Enum
  extend(T::Sig)

  enums do
    MyValue = new('MV')
  end

  private

  sig { void }
  def something_private
  end
end

MyEnum::MyValue.serialize
```
[sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20MyEnum%20%3C%20T%3A%3AEnum%0A%20%20extend%28T%3A%3ASig%29%0A%0A%20%20enums%20do%0A%20%20%20%20MyValue%20%3D%20new%28'MV'%29%0A%20%20end%0A%0A%20%20private%0A%0A%20%20sig%20%7B%20void%20%7D%0A%20%20def%20something_private%0A%20%20end%0Aend%0A%0AMyEnum%3A%3AMyValue.serialize)

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/TEnum.cc
+++ b/rewriter/TEnum.cc
@@ -245,6 +245,10 @@ void TEnum::run(core::MutableContext ctx, ast::ClassDef *klass) {
         auto sig = ast::MK::Sig0(klass->declLoc, std::move(return_type_ast));
         auto method = ast::MK::SyntheticMethod0(klass->loc, klass->declLoc, core::Names::serialize(),
                                                 ast::MK::RaiseTypedUnimplemented(klass->declLoc));
+        auto visibility =
+            ast::MK::Send0(klass->declLoc, ast::MK::Self(klass->declLoc), core::Names::public_(), klass->declLoc);
+
+        klass->rhs.emplace_back(std::move(visibility));
         klass->rhs.emplace_back(std::move(sig));
         klass->rhs.emplace_back(std::move(method));
     }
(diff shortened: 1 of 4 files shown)


jez (approved):

jez:
Looks good but you've got some test failures.

bitwise-aiden:
It looks like calling the public method will be a bit difficult to do in the right context. Do you have any suggestions on how to approach this?

froydnj:
> It looks like calling the public method will be a bit difficult to do in the right context. Do you have any suggestions on how to approach this?

The compiler should be getting this right; it's kind of peculiar that it's not.  Is it possible that you can wait until sometime next week (possibly later in the week) for a compiler fix?

rewriter/TEnum.cc
```diff
@@ -245,6 +245,10 @@ void TEnum::run(core::MutableContext ctx, ast::ClassDef *klass) {
         auto sig = ast::MK::Sig0(klass->declLoc, std::move(return_type_ast));
         auto method = ast::MK::SyntheticMethod0(klass->loc, klass->declLoc, core::Names::serialize(),
                                                 ast::MK::RaiseTypedUnimplemented(klass->declLoc));
+        auto visibility =
+            ast::MK::Send0(klass->declLoc, ast::MK::Self(klass->declLoc), core::Names::public_(), klass->declLoc);
```
froydnj, line 249: Ah, never mind about the compiler.  You'll need to use `ast::MK::Send` here so that you can pass in flags that indicate that private sends are OK, like:

https://github.com/sorbet/sorbet/blob/6a1084aa558fbf334ccb605e40f347c6fa7f5069/ast/desugar/Desugar.cc#L673-L683

Sorry about the runaround!

bitwise-aiden: No worries at all! That's really great to know, thank you :) I'll fix it up first thing next week.

PR #7210, opened 2023-08-04
Split indexing in tests like how realmain does it

jez:
I recommend reviewing by commit for an easier time.

### Motivation

This is a follow up to #7188, which did these same changes for realmain (I had
forgotten that we need to mimic pipeline changes to tests, but was not told
otherwise by the test suite because I was not expecting to see any changes).

### Test plan

Test-only change.

### DIFF
--- a/main/pipeline/pipeline.cc
+++ b/main/pipeline/pipeline.cc
@@ -682,6 +682,30 @@ void typecheckOne(core::Context ctx, ast::ParsedFile resolved, const options::Op
 }
 } // namespace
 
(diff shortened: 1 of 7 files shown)


froydnj (approved):
Something something separate pipelines for test and non-test code.

jez:
> Something something separate pipelines for test and non-test code.

While I was looking at this code, I was wondering if we could replace our `ExpectationHandler` with the existing print options we have.

I think that was the only reason why we had separate pipelines in the past—we wanted to inject code at certain points in the pipeline to capture the snapshot output.

Maybe with some creativity we could repurpose the print options for that, and have them "print" to strings instead of printing to a file.

PR #7213, opened 2023-08-07
Enable toggling Sorbet LSP options without needing a full restart

maruth-stripe:
### Test plan

See included automated tests.


main/lsp/notifications/did_change_configuration.cc
```diff
@@ -0,0 +1,22 @@
+#include "main/lsp/notifications/did_change_configuration.h"
+#include "main/lsp/LSPPreprocessor.h"
+
+namespace sorbet::realmain::lsp {
+DidChangeConfigurationTask::DidChangeConfigurationTask(const LSPConfiguration &config,
+                                                       std::unique_ptr<DidChangeConfigurationParams> params)
+    : LSPTask(config, LSPMethod::WorkspaceDidChangeConfiguration), params(move(params)) {}
+
+LSPTask::Phase DidChangeConfigurationTask::finalPhase() const {
+    return LSPTask::Phase::RUN;
```
jez, line 10: Can we add a comment here basically saying that we want our task to run all the way to type checker so that the update propagates to all globalstates that are alive in the LSP process?

main/lsp/notifications/did_change_configuration.cc
```diff
@@ -0,0 +1,22 @@
...
+}
+
+void DidChangeConfigurationTask::preprocess(LSPPreprocessor &preprocessor) {
+    preprocessor.resume();
+}
+
+void DidChangeConfigurationTask::index(LSPIndexer &indexer) {}
```
jez, line 17: Do we not need to use this callback to update the indexer's options too?

main/lsp/tools/make_lsp_types.cc
```diff
@@ -1284,6 +1277,12 @@ void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_pt
     // Empty object.
     auto InitializedParams = makeObject("InitializedParams", {}, classTypes);
 
+    auto DidChangeConfigurationParams = makeObject("DidChangeConfigurationParams",
+                                                   {
+                                                       makeField("settings", SorbetInitializationOptions),
```
jez, line 1282: Curious what your thought process was for this API.

Another option would be to simply reuse SorbetInitializationOptions at the top-level of the didChangeConfiguration request, instead of having a `"settings"` field.

I'm not sure which I like more, as well as whether you considered other APIs before settling on this one.

main/lsp/notifications/did_change_configuration.cc
```diff
@@ -0,0 +1,22 @@
...
+DidChangeConfigurationTask::DidChangeConfigurationTask(const LSPConfiguration &config,
+                                                       std::unique_ptr<DidChangeConfigurationParams> params)
+    : LSPTask(config, LSPMethod::WorkspaceDidChangeConfiguration), params(move(params)) {}
+
+LSPTask::Phase DidChangeConfigurationTask::finalPhase() const {
+    return LSPTask::Phase::RUN;
+}
+
+void DidChangeConfigurationTask::preprocess(LSPPreprocessor &preprocessor) {
+    preprocessor.resume();
```
jez, line 14: ooc, what's up with this? it's not a part of sorbet i'm familiar with

jez: my _guess_ is that we *do not* want this. the only other place i see this being called is in the sorbet_resume notification, which is a test-only method that pairs with sorbet_pause that makes it easier to test multi-threaded code

main/lsp/notifications/did_change_configuration.cc
```diff
@@ -0,0 +1,22 @@
...
+namespace sorbet::realmain::lsp {
+DidChangeConfigurationTask::DidChangeConfigurationTask(const LSPConfiguration &config,
+                                                       std::unique_ptr<DidChangeConfigurationParams> params)
+    : LSPTask(config, LSPMethod::WorkspaceDidChangeConfiguration), params(move(params)) {}
+
+LSPTask::Phase DidChangeConfigurationTask::finalPhase() const {
+    return LSPTask::Phase::RUN;
+}
+
+void DidChangeConfigurationTask::preprocess(LSPPreprocessor &preprocessor) {
```
jez, line 13: like the index comment, i believe we want to use this callback to update the preprocessor's globalstate(s) too

maruth-stripe: Does it have one? (I don't see it)

main/lsp/notifications/did_change_configuration.cc
maruth-stripe: yeah this was lurked in from copy-pasta

main/lsp/tools/make_lsp_types.cc
maruth-stripe: Just going according to the official spec https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration 

but happy to do just top-level if we feel strongly about it

main/lsp/notifications/did_change_configuration.cc
maruth-stripe: yup

main/lsp/notifications/did_change_configuration.cc
maruth-stripe: sure

main/lsp/notifications/did_change_configuration.cc
jez: I believe you're correct that it doesn't have one. Thanks for checking!

test/lsp/protocol_test_corpus.cc
```diff
@@ -694,4 +694,31 @@ TEST_CASE_FIXTURE(ProtocolTest, "ReportsSyntaxErrors") {
     CHECK_EQ(counters.getCategoryCounter("lsp.slow_path_reason", "changed_definition"), 0);
 }
 
+TEST_CASE_FIXTURE(ProtocolTest, "DidChangeConfigurationNotificationUpdatesHighlightUntypedSetting") {
```
maruth-stripe, line 697: This test pretty much makes sure the feature runs without crashing for now. I would like to keep it to (a) have that sanity check, and (b) we just need to update assertDiagnostics so we can have a version that handles information diagnostics too,.

vscode_extension/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Version history
+## 0.3.25
+- `Toggle untyped highlight without restart`
```
damolina-stripe, line 3: What about it?

maruth-stripe: updated

vscode_extension/src/commands/toggleUntypedCodeHighlighting.ts
```diff
@@ -14,7 +13,13 @@ export async function toggleUntypedCodeHighlighting(
...
-
-  await context.statusProvider.restartSorbet(RestartReason.CONFIG_CHANGE);
+  
+  context.statusProvider.activeLanguageClient?.sendNotification("workspace/didChangeConfiguration",
+    {
+      "settings": {
+        "highlightUntyped": targetState
+      }
+    }
+  );
```
damolina-stripe, line 23: If there is no `context.statusProvider.activeLanguageClient` seems this whole flow would not make sense.

Could the implementation be changed so it checks/bails early if there is no `context.statusProvider.activeLanguageClient` ?

Something  like
```
const {activeLanguageClient: client} = context.statusProvider;
  if (!client) {
    context.log.debug("ToggleUntyped: No active Sorbet LSP.");
  }

…

  context.log.info(
    `ToggleUntyped: Untyped code highlighting: ${targetState ? "enabled" : "disabled"}`,
  );
  client.sendNotification("workspace/didChangeConfiguration", {
    settings: {
      highlightUntyped: targetState,
    },
  });
…
```

main/lsp/LSPTypechecker.cc
```diff
@@ -647,11 +647,11 @@ LSPQueryResult LSPTypechecker::query(const core::lsp::Query &q, const std::vecto
     return LSPQueryResult{queryCollector->drainQueryResponses(), nullptr};
 }
 
-LSPFileUpdates LSPTypechecker::getNoopUpdate(std::vector<core::FileRef> frefs) const {
+LSPFileUpdates LSPTypechecker::getNoopUpdateWithEpoch(std::vector<core::FileRef> frefs, uint32_t epoch) const {
     LSPFileUpdates noop;
     noop.typecheckingPath = TypecheckingPath::Fast;
     // Epoch isn't important for this update.
```
jez, line 653: I think you want to move this comment to the other function.

main/lsp/notifications/did_change_configuration.cc
```diff
@@ -0,0 +1,30 @@
...
+LSPTask::Phase DidChangeConfigurationTask::finalPhase() const {
+    // We want this to run all the way so that the changes to
+    // Global State get propagated through.
+    return LSPTask::Phase::RUN;
+}
+
+void DidChangeConfigurationTask::preprocess(LSPPreprocessor &preprocessor) {
+    auto tmp = preprocessor.openFilePaths();
+    openFilePaths = std::move(tmp);
+    epoch = preprocessor.incNextVersion();
```
jez, line 19: Why expose this function, versus accepting `epoch` as an argument to the `DidChangeConfigurationTask` constructor, and populating it via `nextVersion++` in the `getTaskForMessage` method like everywhere else?

main/lsp/notifications/did_change_configuration.cc
```diff
@@ -0,0 +1,30 @@
...
+
+LSPTask::Phase DidChangeConfigurationTask::finalPhase() const {
+    // We want this to run all the way so that the changes to
+    // Global State get propagated through.
+    return LSPTask::Phase::RUN;
+}
+
+void DidChangeConfigurationTask::preprocess(LSPPreprocessor &preprocessor) {
+    auto tmp = preprocessor.openFilePaths();
+    openFilePaths = std::move(tmp);
```
jez, line 18: What's the point of `tmp` here? can this not be

```suggestion
    openFilePaths = preprocessor.openFilePaths();
```

?

main/lsp/LSPPreprocessor.cc
```diff
@@ -503,5 +506,16 @@ LSPPreprocessor::canonicalizeEdits(uint32_t v, unique_ptr<WatchmanQueryResponse>
     }
     return edit;
 }
+const std::unique_ptr<std::vector<std::string_view>> LSPPreprocessor::openFilePaths() const {
+    auto paths = make_unique<std::vector<std::string_view>>();
+    for (auto const &[path, file] : openFiles) {
+        paths->emplace_back(path);
+    }
+    return paths;
+}
```
jez, line 515: I have half a mind to suggest that we keep openFilePaths private and also call it in getTaskForMessage while constructing a DidChangeConfigurationTask, but I feel less strongly about that.

main/lsp/LSPPreprocessor.cc
```diff
@@ -503,5 +506,16 @@ LSPPreprocessor::canonicalizeEdits(uint32_t v, unique_ptr<WatchmanQueryResponse>
     }
     return edit;
 }
+const std::unique_ptr<std::vector<std::string_view>> LSPPreprocessor::openFilePaths() const {
```
jez, line 509: What's the point of putting this in a unique_ptr?

test/lsp/protocol_test_corpus.cc
```diff
@@ -418,8 +418,8 @@ TEST_CASE_FIXTURE(ProtocolTest, "IgnoresNotificationsThatDontTypecheck") {
 // Ensures that unrecognized requests are responded to.
 TEST_CASE_FIXTURE(ProtocolTest, "RejectsUnrecognizedRequests") {
     assertDiagnostics(initializeLSP(), {});
-    auto responses = sendRaw("{\"jsonrpc\":\"2.0\",\"method\":\"workspace/"
-                             "didChangeConfiguration\",\"id\":9001,\"params\":{\"settings\":{\"ruby-typer\":{}}}}");
+    auto responses = sendRaw("{\"jsonrpc\":\"2.0\",\"method\":\"sorbet/"
+                             "fooBar\",\"id\":9001,\"params\":{\"settings\":{\"highlightUntyped\": false}}}");
```
jez, line 422: lmao

main/lsp/tools/make_lsp_types.cc
jez: oh whoops, I didn't realize the spec dictated that. This looks good!

main/lsp/LSPTypechecker.cc
```diff
@@ -718,6 +731,10 @@ void LSPTypechecker::setSlowPathBlocked(bool blocked) {
     slowPathBlocked = blocked;
 }
 
+void LSPTypechecker::updateGsFromOptions(const DidChangeConfigurationParams &options) const {
+    this->gs->trackUntyped = options.settings->highlightUntyped.value_or(this->gs->trackUntyped);
```
jez, line 735: We may want to log some sort of warning when there are as-yet-unimplemented settings passed in the DidChangeConfigurationParams.

For example, `options.settings` might have things that `.has_value()` but will be ignored until someone wires those up, like `enableTypedFalseCompletionNudges`.

(Also, looking at the list of things in SorbetInitializationOptions, there are a handful of things in there that we may never support in this did change notification... I don't know whether it makes sense to make two separate types, but it may make sense to send some sort of warning back to the user if they attempt to set one of those things.)

jez (commented):
This looks pretty good. We chatted about some ways to make fewer changes to LSPTypechecker by reusing getNoopUpdate and typecheckOnFastPath.

Happy to review again when that comes in.

main/lsp/LSPPreprocessor.cc
maruth-stripe: TIL https://stackoverflow.com/questions/15704565/efficient-way-to-return-a-stdvector-in-c

main/lsp/LSPPreprocessor.cc
maruth-stripe: just got rid of it

main/lsp/notifications/did_change_configuration.cc
maruth-stripe: Honestly, I don't know. The compiler gets mad about using an implicitly deleted copy constructor if I don't assign to a temporary first. No clue what's up with that.

main/lsp/LSPTypechecker.cc
maruth-stripe: Hmm not sure what the API for this would be,  I don't think we send a response for this notification beyond flushing errors to the client? So unclear where this would go.

jez: you can either use window showMessage, or simply log something with the logger

main/lsp/LSPPreprocessor.cc
```diff
@@ -229,6 +229,11 @@ unique_ptr<LSPTask> LSPPreprocessor::getTaskForMessage(LSPMessage &msg) {
                 return make_unique<SorbetResumeTask>(*config);
             case LSPMethod::SorbetError:
                 return make_unique<SorbetErrorTask>(*config, move(get<unique_ptr<SorbetErrorParams>>(rawParams)));
+            case LSPMethod::WorkspaceDidChangeConfiguration: {
+                auto tmp = this->openFilePaths();
+                return make_unique<DidChangeConfigurationTask>(
+                    *config, move(get<unique_ptr<DidChangeConfigurationParams>>(rawParams)), move(tmp), nextVersion++);
```
froydnj, line 235: Note that since `openFilePaths` is returning a `const` thing and therefore `tmp` is `const`, the `move` here is not actually moving anything and will wind up doing a copy (thanks weird C++ semantics).  You should probably change the return type of `openFilePaths`, and that will enable you to drop the temporary variable here.

main/lsp/LSPPreprocessor.cc
```diff
@@ -503,5 +508,12 @@ LSPPreprocessor::canonicalizeEdits(uint32_t v, unique_ptr<WatchmanQueryResponse>
     }
     return edit;
 }
+const std::vector<std::string_view> LSPPreprocessor::openFilePaths() const {
+    std::vector<std::string_view> paths;
+    for (auto const &[path, file] : openFiles) {
```
froydnj, line 513: Minor nit: add `paths.reserve(openFiles.size());` before the loop.

main/lsp/LSPPreprocessor.h
```diff
@@ -174,6 +174,8 @@ class LSPPreprocessor final {
     std::optional<std::string_view> maybeGetFileContents(std::string_view path) const;
 
     std::unique_ptr<Joinable> runPreprocessor(MessageQueueState &messageQueue, absl::Mutex &messageQueueMutex);
+
+    std::vector<std::string_view> openFilePaths() const;
```
jez, line 178: If I'm not mistaken, this can now be a private method.

main/lsp/LSPTypechecker.cc
```diff
@@ -718,6 +718,23 @@ void LSPTypechecker::setSlowPathBlocked(bool blocked) {
...
+
+    if (options.settings->enableTypecheckInfo.has_value() ||
+        options.settings->enableTypedFalseCompletionNudges.has_value() ||
+        options.settings->supportsOperationNotifications.has_value() ||
+        options.settings->supportsSorbetURIs.has_value()) {
+        auto msg =
+            "Currently `highlightUntyped` is the only updateable setting using the workspace/didChangeConfiguration "
+            "notification";
+        config->logger->error(msg);
```
jez, line 731:
```suggestion
```

If you're going to send it back via showMessage, you don't also need to send it with the logger, as all `config->output->write` calls log what they write:

https://github.com/sorbet/sorbet/blob/master/main/lsp/LSPOutput.cc#L59

main/lsp/LSPTypechecker.h
```diff
@@ -109,6 +104,11 @@ class LSPTypechecker final {
      */
     std::vector<std::unique_ptr<core::Error>> retypecheck(std::vector<core::FileRef> frefs, WorkerPool &workers) const;
 
+    /**
+     * Re-typechecks the provided files and flushes error messages to client.
+     */
+    void retypecheckAndFlush(std::vector<core::FileRef> frefs, WorkerPool &workers, uint32_t epoch) const;
+
```
jez, line 111:
```suggestion
```

No longer used, right?

main/lsp/notifications/did_change_configuration.cc
```diff
@@ -0,0 +1,32 @@
...
+void DidChangeConfigurationTask::run(LSPTypecheckerDelegate &tc) {
+    tc.updateGsFromOptions(*params);
+    std::vector<core::FileRef> openFileRefs;
+    for (auto const &path : openFilePaths) {
+        openFileRefs.push_back(tc.state().findFileByPath(path));
+    }
+    auto updates = tc.getNoopUpdate(openFileRefs);
+    updates.epoch = epoch;
+    tc.typecheckOnFastPath(std::move(updates), {});
+}
```
jez, line 31: :so-beautiful:

test/lsp/protocol_test_corpus.cc
jez: Were you planning to be able to write a test that shows that the untyped diagnostics come back?

jez (approved):
Looks good to me, modulo some minor things and one request for a test.

maruth-stripe:
> one request for a test
@jez as in one that updates `assertDiagnostics` etc etc?

jez:
👍

test/lsp/ProtocolTest.h
```diff
@@ -95,8 +95,15 @@ class ProtocolTest {
 
     std::unique_ptr<LSPMessage> readAsync();
 
+    template <typename T>
     void assertDiagnostics(std::vector<std::unique_ptr<LSPMessage>> messages, std::vector<ExpectedDiagnostic> expected);
```
jez, line 99: This shouldn't be in the header file anymore, right?

If someone tried to call this method on a type that wasn't also one of the two called in ProtocolTest.cc, it would be a linker failure.

jez (approved):

test/lsp/ProtocolTest.h
maruth-stripe: man I need to get a C++ book asap

maruth-stripe: Hmm seems like it needs the template part in the header file? I get an error without it

jez: #7227

PR #7214, opened 2023-08-09
Auto-save codeAction changes to __package.rb files

jez:
### Implementation

We have some options here. Another option would be to implement something like
"save all files that weren't open before we applied the WorkspaceEdit," but
that's a little tricky, because we'd have to compute the list of open files on
the server side--by the time the WorkspaceEdit has applied, VS Code will already
have the files open, so we can't do it purely in the client.

In any case, this isn't a trapdoor decision--we can always decide to change our
minds and expand this feature if people start to ask for it.

### Motivation

Requested by a user

### Test plan

I only tested this manually.

https://github.com/sorbet/sorbet/assets/5544532/1f053903-b7b0-4873-b59a-75e0e198ef58

It's not particularly easy to test these changes in a meaningful way on the
server, because we don't have a concept of "edited but not saved files" in our
test harness.

It's possible that this is something we could write a VS Code test for? I'm not
super familiar with how one might work. @damolina-stripe, do you have thoughts
on whether/how to test the client-side portions of this change?

### Client changes in non-VS Code editors

I found [this comment](https://github.com/microsoft/language-server-protocol/issues/430#issuecomment-379769425) from a maintainer of the LSP spec to be useful. The idea seems to be that clients should control maintaining a registry of commands, mapping to their implementations. Servers can kick-start this registry using the `executeCommandProvider` capability on initialization if it's desired to offload the implementation of a command to the server.

In our case, the thing we want relies entirely on the client, so we can't have this command implemented on the server.

I did confirm that at least in Neovim's built-in language client, they support declaring a string -> callback function registry, which is basically the same as the API VS Code gives us with `registerCommand`. I'm guessing that means that other language clients will be able to implement this as well.


jez:

→ https://go/builds/bui_OPtucaQdie9lH1
→ https://go/builds/bui_OPtvlGTGBkKyto
→ https://go/builds/bui_OPtvs9Sisg0zbO

damolina-stripe:
@jez It is possible to mock the `WorkspaceEdit` when writing a test against the command. There are a a couple examples in our internal codebase I can point you at.

One concern with this new command, is that "Save all __package.rb" files seems like a maybe too-specific general use command. If the only goal is to call it internally [here](https://github.com/sorbet/sorbet/pull/7214/files#diff-03bbc5e104f3c61a0610b137b2b0e2df9865704040614dac60a57142a3bc7682R157) 
```
(action->command = make_unique<Command>("Save package files", "sorbet.savePackageFiles");
```
Then maybe defining as an internal command would be better - just don't define it in the package.json, the commandId is available for API calls still and activation should not be an issue since the extension would have to already be active for the command to be needed.

vscode_extension/src/extension.ts
```diff
@@ -7,6 +7,7 @@ import { setLogLevel } from "./commands/setLogLevel";
 import { showSorbetActions } from "./commands/showSorbetActions";
 import { showSorbetConfigurationPicker } from "./commands/showSorbetConfigurationPicker";
 import { toggleUntypedCodeHighlighting } from "./commands/toggleUntypedCodeHighlighting";
+import { savePackageFiles } from "./commands/savePackageFiles";
```
damolina-stripe, line 10: Nit: project file imports are sorted by import path, can this be moved to line 6?

damolina-stripe (approved):

PR #7212, opened 2023-08-07
WIP: handle super in strict files

neilparikh:
### TODO
- [ ] turn off method suggestions for super
- [ ] new error code for ancestors not found

### Test plan

See included automated tests.

PR #7219, opened 2023-08-09
Allow T.bind(self, T.untyped)

neilparikh:
### Motivation

Escape hatch for errors raised by https://github.com/sorbet/sorbet/pull/7212

When there's an invalid call to `super`, the user can't silence it using `T.unsafe(self).super`, since `super` is special syntax. To get around this, we will have to support the following:

```
class A
  def foo
    T.bind(self, T.untyped)
    super
  end
end
```

### Test plan

See included automated tests.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1587,7 +1587,7 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                         }
                     }
                 } else if (!bind.value.isSynthetic()) {
-                    if (castType.isUntyped()) {
+                    if (castType.isUntyped() && bind.bind.variable != cfg::LocalRef::selfVariable()) {
                         if (auto e = ctx.beginError(bind.loc, core::errors::Infer::InvalidCast)) {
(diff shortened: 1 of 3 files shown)


jez (approved):

infer/environment.cc
```diff
@@ -1587,7 +1587,7 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                         }
                     }
                 } else if (!bind.value.isSynthetic()) {
-                    if (castType.isUntyped()) {
+                    if (castType.isUntyped() && bind.bind.variable != cfg::LocalRef::selfVariable()) {
```
jez, line 1590: Non-blocking: might be worth mentioning why we have this selfVariable check here (e.g.: `self` is not actually assignable in normal Ruby syntax, so if the target of this binding is `<self>` we know that this `T.cast` is actually a `T.bind` in disguise, but we've already desugared `T.bind` to `T.cast` so we can't simply check that on the `c.cast` directly)

PR #7220, opened 2023-08-09
Add a rest arg to rewrite generated initialize method of before do end

neilparikh:
### Motivation

We currently rewrite `before do ... end` to:
```
sig { void }
def initialize
  ...
end
```
Once `super` is typechecked, this causes problems for the following:
```
class A
  def initialize(a)
  end
end
class B < A
  before do
  end
end
class C < B
  def initialize(a)
    super
  end
end
```

The super call will be typechecked against the rewriter generated `initialize`, and the arity will not match, even though it'll actually be calling `A#initialize`. Adding an `T.untyped` rest arg silences this error.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/infer/suggest_field.rb.autocorrects.exp
+++ b/test/testdata/infer/suggest_field.rb.autocorrects.exp
@@ -9,11 +9,11 @@ class A
 
   sig {void}
   def initialize
-    @x = T.let(0, Integer)
+    @x = T.let(0, T.nilable(Integer))
   # ^^ error: The instance variable `@x` must be declared using `T.let` when specifying `# typed: strict`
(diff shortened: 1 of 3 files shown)


test/testdata/infer/suggest_field.rb.autocorrects.exp
```diff
@@ -9,11 +9,11 @@ class A
 
   sig {void}
   def initialize
-    @x = T.let(0, Integer)
+    @x = T.let(0, T.nilable(Integer))
```
neilparikh, line 12: This change is a bit odd to me, and I'm not really sure why it's occurring here

PR #7206, opened 2023-08-03
Add option to disable nudges

tonybruess:
This PR adds a simple configuration option to turn off nudges in `#typed: false` files.

I wasn't sure if I should name the option `enableNudges` with a default true value or `disableNudges` with a default false value. Let me know if you want me to rename it to something else.

### Motivation
The `file is not # typed: true or higher` nudges conflict with GitHub Copilot and other autocomplete extensions.

@jez said "Go for it"

https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1691086045697219?thread_ts=1691085509.929519&cid=CHN2L03NH

### Test plan

See included automated tests.

PR #7221, opened 2023-08-10
Add disable nudges option to vscode extension

joshuay03:
### Motivation

Add the option added [here](https://github.com/sorbet/sorbet/pull/7206) to the vscode extension config.

### Test plan

Unsure if all the changes here require coverage, would appreciate a nudge (😉) in the right direction if there's coverage I should be updating.

See included automated tests.

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -30,6 +30,7 @@
     "onCommand:sorbet.setLogLevel",
     "onCommand:sorbet.showOutput",
     "onCommand:sorbet.toggleHighlightUntyped",
+    "onCommand:sorbet.toggleTypedFalseCompletionNudges",
     "onLanguage:ruby",
     "workspaceContains:sorbet/*"
   ],
@@ -94,6 +95,12 @@
       {
         "command": "sorbet.toggleHighlightUntyped",
         "title": "Toggle highlighting untyped code",
+        "category": "Sorbet",
+        "enablement": "workbenchState != empty"
+      },
+      {
+        "command": "sorbet.toggleTypedFalseCompletionNudges",
+        "title": "Toggle the auto-complete nudge in `typed: false` files",
         "category": "Sorbet",
         "enablement": "workbenchState != empty"
       }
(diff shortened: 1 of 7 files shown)


jez:
Does this code do anything? Correct me if I’m wrong but tracing the code leads me to believe that this amounts to a lot of boilerplate code to toggle an unused Boolean.

joshuay03:
> Correct me if I’m wrong but tracing the code leads me to believe that this amounts to a lot of boilerplate code to toggle an unused Boolean.

Honestly, you might be right haha. I'm still figuring this out. I was just keen to try out this new option but couldn't figure out how to enable it. So I thought I'd get a PR up to get it to work.

Is is possible to test the config locally by doing something like compiling the extension locally? 👀

joshuay03:
Hmm I'm trying to test this out with the `.vsix` extension and cannot get it to work. The UI command and JSON setting toggling seem to be working fine (testing with it configured to `false` ie nudge should now show). I think I'm sending the correct initialization option.

jez:
> Honestly, you might be right haha

Cool, no worries. Heads up that #7213 is going to completely rearchitect how this flow works, so I would appreciate your patience while we finalize that change.

jez:
> Hmm I'm trying to test this out with the `.vsix` extension and cannot get it to work

The best way to test the extension is to do

```
cd vscode_extension/
code --new-window .
```

and then use the "Run and Debug" window and then click the green triangle next to the "Launch Extension" task, which will open a second VS Code window where the development version of the extension is running. You can then open a workspace in that new window and play around with it. It will also do things like set your default log level to trace and let you place debug breakpoints in the TypeScript code.

joshuay03:
> Heads up that #7213 is going to completely rearchitect how this flow works, so I would appreciate your patience while we finalize that change.

No worries, this isn't urgent, happy to hold off till then.

> The best way to test the extension is to do

This is super helpful, thank you!

joshuay03:
I managed to test out the extension quite comprehensively. The setting toggling via the UI and JSON config works perfectly. I've confirmed that the value flows all the way to this point where it's added to the init options for the server:
https://github.com/sorbet/sorbet/blob/e37c07267e7a1ff068a7cb1d94c8f843404d5d08/vscode_extension/src/languageClient.ts#L62-L63

And I'm also quite confident that I'm using the right init option name here ^.

I was also unable to enable the feature via a config file in my project ie `sorbet/config` using `--enable-typed-false-completion-nudges=false` which results in:
```ruby
Option ‘enable-typed-false-completion-nudges’ does not exist. To see all available options pass `--help`.
```

I got a separate PR up adding the option in: https://github.com/sorbet/sorbet/pull/7225.

jez (approved):

PR #7222, opened 2023-08-10
Error if using `===` on `T::Array` instead of `Array`

jez:
### Motivation

Requested by a user

### Test plan

See included automated tests.

Screenshot 2023-08-10 at 4 59 54 PM

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -4143,6 +4143,32 @@ public:
     }
 } T_Enum_tripleEq;
 
+class GenericForwarder_tripleEq : public IntrinsicMethod {
+public:
+    void apply(const GlobalState &gs, const DispatchArgs &args, DispatchResult &res) const override {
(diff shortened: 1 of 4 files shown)


core/types/calls.cc
```diff
@@ -4238,6 +4264,16 @@ const vector<Intrinsic> intrinsics{
 
     {Symbols::Module(), Intrinsic::Kind::Instance, Names::tripleEq(), &Module_tripleEq},
     {Symbols::T_Enum(), Intrinsic::Kind::Instance, Names::tripleEq(), &T_Enum_tripleEq},
+
+    {Symbols::T_Array(), Intrinsic::Kind::Singleton, Names::tripleEq(), &GenericForwarder_tripleEq},
```
froydnj, line 4268: It's too bad this is another list of `T` generics that has to be kept up with other places that we care about (although it's not the end of the world if this list isn't updated along with the others).

froydnj (approved):

PR #7223, opened 2023-08-11
Make go-to-def on `override` go to the abstract method

### LINKED ISSUE #7165: Have "Go to Definition" on the `override.` in a method signature go to the method it overrides
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20Module%3B%20include%20T%3A%3ASig%3B%20end%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20foo%0A%20%20%20%20%23%20%5E%20definition%20is%20here%0A%20%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20sig%20%7Boverride.void%7D%0A%20%20%23%20%20%20%20%5E%20jump%20to%20def%0A%20%20def%20foo%0A%20%20end%0Aend)

```ruby
# typed: true
class Module; include T::Sig; end

class Parent
  extend T::Helpers
  abstract!

  sig {abstract.void}
  def foo
    # ^ definition is here
  end
end

class Child < Parent
  sig {override.void}
  #    ^ jump to def
  def foo
  end
end
```

#### Observed output

```
No errors! Great job.
```

If you click "Go to Definition" on the override keyword, right now it goes to the definition of the `override` DSL method in Sorbet's payload. While that's technically correct, it's infinitely less useful than having it go to the definition of the parent method.

#### Expected behavior

Jump to the definition of the abstract method.
- - -

jez:
### Motivation

I've wanted this for a while.

Fixes #7165

### Test plan

See included automated tests.

### DIFF
--- a/core/SymbolRef.h
+++ b/core/SymbolRef.h
@@ -973,6 +973,18 @@ public:
         return ClassOrModuleRef::fromRaw(82);
     }
 
+    static MethodRef T_Private_Methods_DeclBuilder_abstract() {
+        return MethodRef::fromRaw(4);
+    }
+
+    static MethodRef T_Private_Methods_DeclBuilder_overridable() {
+        return MethodRef::fromRaw(5);
+    }
+
+    static MethodRef T_Private_Methods_DeclBuilder_override() {
+        return MethodRef::fromRaw(6);
+    }
+
     static ClassOrModuleRef T_Sig_WithoutRuntimeSingleton() {
         return ClassOrModuleRef::fromRaw(83);
     }
--- a/core/SymbolRef.h
+++ b/core/SymbolRef.h
@@ -990,7 +990,7 @@ public:
     }
 
     static MethodRef sigWithoutRuntime() {
-        return MethodRef::fromRaw(4);
+        return MethodRef::fromRaw(7);
     }
 
     static ClassOrModuleRef T_NonForcingConstants() {
@@ -998,7 +998,7 @@ public:
     }
 
     static MethodRef SorbetPrivateStaticSingleton_sig() {
-        return MethodRef::fromRaw(5);
+        return MethodRef::fromRaw(8);
     }
 
     static ClassOrModuleRef PackageSpecRegistry() {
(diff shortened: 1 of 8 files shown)


core/SymbolRef.h
```diff
@@ -973,20 +973,32 @@ class Symbols {
         return ClassOrModuleRef::fromRaw(82);
     }
 
+    static MethodRef T_Private_Methods_DeclBuilder_abstract() {
+        return MethodRef::fromRaw(4);
+    }
+
+    static MethodRef T_Private_Methods_DeclBuilder_overridable() {
+        return MethodRef::fromRaw(5);
+    }
```
jez, line 982: These are currently unused, but I anticipate using them to implement the opposite feature, where you can click on these to go to the implementations of the method, so I figured I'd add them while I was making a mess to these ref numbers anyways.

core/SymbolRef.h
```diff
@@ -973,20 +973,32 @@ class Symbols {
...
+        return MethodRef::fromRaw(6);
+    }
+
     static ClassOrModuleRef T_Sig_WithoutRuntimeSingleton() {
         return ClassOrModuleRef::fromRaw(83);
     }
 
     static MethodRef sigWithoutRuntime() {
-        return MethodRef::fromRaw(4);
+        return MethodRef::fromRaw(7);
```
ilyailya, line 993: How does those IDs work? My mental model was they are arbitrary and can go in any order and also they should not repeat

jez: They are not arbitrary, they must match the order that symbols are created in GlobalState::initEmpty. So if you want a method symbol to have ID 7, you must have already inserted symbols with IDs 0 through 6 in GlobalState::initEmpty already.

This also means that if in GlobalState::initEmpty you want to insert some code to enter new method symbols early in the file, before other method symbols are created later, it will have the effect of bumping all the IDs.

For this reason, some people prefer adding new symbols at the end of the list and just not caring about code organization in GlobalState::initEmpty. But I think it's usually important to group similar definitions near each other, and so I typically renumber these IDs.

PR #7224, opened 2023-08-11
Add two sandbox configurations

jez:
### Motivation

This way people don't have to figure this out how to point a language client at
the Sorbet they just built into `bazel-bin`.

### Test plan

I tested these manually with the commands mentioned in the respective READMEs.


froydnj (approved):
Having this documentation to point people to is so nice, thank you!

PR #7225, opened 2023-08-11
Add disable nudges option to config options

joshuay03:
### Motivation

Add the option added [here](https://github.com/sorbet/sorbet/pull/7206) to the config.

More context in https://github.com/sorbet/sorbet/pull/7221.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2095,6 +2095,7 @@ unique_ptr<GlobalState> GlobalState::deepCopy(bool keepId) const {
     result->requiresAncestorEnabled = this->requiresAncestorEnabled;
     result->ruby3KeywordArgs = this->ruby3KeywordArgs;
     result->trackUntyped = this->trackUntyped;
+    result->enableTypedFalseCompletionNudges = this->enableTypedFalseCompletionNudges;
     result->printingFileTable = this->printingFileTable;
 
     if (keepId) {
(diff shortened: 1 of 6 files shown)


joshuay03:
> This change also looks like it doesn't do anything. You are not using `opts.enableTypedFalseCompletionNudges` anywhere.

My bad, I meant to open a draft PR.
I've added it to the global state with fa1df95c0c3bdf1040ad7a695889460dd6f72e9d.

jez:
> I've added it to the global state with [fa1df95](https://github.com/sorbet/sorbet/commit/fa1df95c0c3bdf1040ad7a695889460dd6f72e9d).

Simply adding it to the GlobalState also doesn't do anything, because that new boolean in GlobalState is not being used to initialize anything in LSPConfiguration, which is how this whole feature works (by reading the boolean in LSPConfiguration).

Even if this PR were changed to invert that, where the LSPConfiguration and/or the new command line option were used to initialize a boolean in GlobalState, that would be the wrong change: this boolean is LSP-specific. We don't want to store LSP-specific boolean on GlobalState.

Instead, we want to store LSP-specific booleans on LSPConfiguration, because they should never need to be read in a place where only GlobalState is available, and instead are always read in a place where the LSPConfiguration is available.

And to the higher level question of "I'd like to be able to configure this with a command line option" this seems like the sort of feature that's actually best left as a configuration option in the client. 

Users can either set `sorbet.typedFalseCompletionNudges` in their VS Code configuration, or `enableTypedFalseCompletionNudges` in any LSP-compatible client that allows setting `initializationOptions`.

With all that said, it doesn't look like there's a place for this PR. I'll continue reviewing the other change.

PR #7227, opened 2023-08-11
Remove uninstantiated template from header

jez:
### Motivation

When using templates, you have to either

- implement the template function in the header, so that it can be instantiated
  at any type that might be required in the future
- pray that code using your header only uses the instantiations that you use in
  your cc file

The way this could have failed before this PR is that someone could call

```cpp
assertDiagnostic<HoverAssertion>(...)
```

or something like that, and this would simply be a linker failure, because the
instantiation of `assertDiagnostic` at `<HoverAssertion>` can't be deduced from
the header file, and none of the libraries that get linked in will have already
defined that symbol.

At least this is my interpretation. Maybe @froydnj has a better way of
explaining it.

### Test plan

n/a

### DIFF
--- a/test/lsp/ProtocolTest.cc
+++ b/test/lsp/ProtocolTest.cc
@@ -241,18 +241,12 @@ vector<unique_ptr<Location>> ProtocolTest::getDefinitions(std::string_view uri, 
     return {};
 }
 
-void ProtocolTest::assertErrorDiagnostics(vector<unique_ptr<LSPMessage>> messages,
-                                          vector<ExpectedDiagnostic> expected) {
-    assertDiagnostics<ErrorAssertion>(std::move(messages), expected);
-}
-
-void ProtocolTest::assertUntypedDiagnostics(vector<unique_ptr<LSPMessage>> messages,
-                                            vector<ExpectedDiagnostic> expected) {
-    assertDiagnostics<UntypedAssertion>(std::move(messages), expected);
-}
+namespace {
 
 template <typename T>
-void ProtocolTest::assertDiagnostics(vector<unique_ptr<LSPMessage>> messages, vector<ExpectedDiagnostic> expected) {
+void assertDiagnostics(vector<unique_ptr<LSPMessage>> messages, vector<ExpectedDiagnostic> expected,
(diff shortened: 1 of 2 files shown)


maruth-stripe (approved):

froydnj (approved):
Your interpretation is correct.  (Minor nit: "deduced" has specific meaning in template-ese, so "the
instantiation of `assertDiagnostic` at `<HoverAssertion>` can't be deduced from
the header file" would make standards warriors shudder; "isn't available in the header file" or somesuch would be better.)

As an aside, the tactic used here is deployed in large C++ codebases to avoid compilation time issues: having the implementation of the template immediately available is useful, but for every file that the template gets used in, you have to instantiate the template, represent that in the compiler's immediate representation, turn it into assembly, etc.  And then the linker will discard all the copies save one, so all the compile time you spent on instantiating that template was wasted.  You can imagine that for large, frequently used templated things that this leads to problems.

There are two ways to deal with this: you can do what this code used to do, which is to declare the template in a header file, provide the implementation in a single file, and then instantiate the template with all the relevant types that you care about in that file, then you kind of get the benefits of templated code without all the code bloat.  The downside is the single file in question might then take forever to compile, and you get inscrutable linker errors if you instantiate the template with a non-approved combination of types.  (Another peculiarity is that the template in question here doesn't generate that much code, but maybe this arrangement reflects a not-quite complete refactoring prior to this PR.)

Alternatively, you can have a full definition of the template and issue promises to the compiler, "yes, I know the template can be instantiated here, but I don't want you to do that; I promise that the template is fully instantiated someplace else", cf. [the C++ FAQ](https://isocpp.org/wiki/faq/cpp11-language-templates#extern-templates):

```c++
extern template class MyVector<int>;
```

Now you can freely use `MyVector<int>` in function parameters, variable declarations, etc. and not pull in all the code for `MyVector<int>` into the relevant compilation unit.  You still can use "non-approved" combinations of template parameters, but for particular ones, you can avoid code bloat issues.  You do need to ensure that the `extern template` bits appear in the appropriate places, though.

It's possible that judicious use of `extern template` for some of Sorbet's vector and unordered map/set types might bring compile times down...but it might not.

Both solutions share the downside that inlining the templated functions/methods is not available to you, save through link-time optimization.

PR #7228, opened 2023-08-14
Add additional ObjectSpace methods to RBI file

ipc103:
### Motivation
There were some missing methods on `ObjectSpace` that we access as documented here: [Object Space Docs](https://ruby-doc.org/stdlib-2.7.0/libdoc/objspace/rdoc/ObjectSpace.html). We added a shim locally for our application, but thought this might be helpful upstream as well.

### Test plan
I didn't add tests for the RBI file here since it didn't seem to add a ton of value. Happy to add some though if folks prefer.


froydnj (approved):
Thanks!

PR #7229, opened 2023-08-14
Default to test_import if referenced from test

jez:
### Motivation

We already have the ability to convert a `test_import` to an `import` the first
time that would become necessary. So if we're about to add an import to a
package, and that package takes a dependency on another package only because of
an edge in test code, we should have the autocorrect suggest `test_import`.

Note that this is separate from the existing `test_import` autocorrect behavior,
which says that if the import is a test-only definition that it should of course
only be imported with `test_import` (because the definition should never be
referenced from non-test code).

### Test plan

See included automated tests.

### DIFF
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -401,6 +401,7 @@ public:
                 e.setHeader("`{}` is defined in a test namespace and cannot be referenced in a non-test file",
                             lit.symbol.show(ctx));
             }
+            return;
         }
 
         auto &db = ctx.state.packageDB();
--- a/packager/VisibilityChecker.cc
+++ b/packager/VisibilityChecker.cc
@@ -456,7 +456,7 @@ public:
             if (auto e = ctx.beginError(lit.loc, core::errors::Packager::MissingImport)) {
                 auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
                 e.setHeader("`{}` resolves but its package is not imported", lit.symbol.show(ctx));
-                bool isTestImport = otherFile.data(ctx).isPackagedTest();
+                bool isTestImport = otherFile.data(ctx).isPackagedTest() || ctx.file.data(ctx).isPackagedTest();
                 e.addErrorLine(pkg.declLoc(), "Exported from package here");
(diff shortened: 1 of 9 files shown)


packager/VisibilityChecker.cc
```diff
@@ -456,7 +456,7 @@ class VisibilityCheckerPass final {
             if (auto e = ctx.beginError(lit.loc, core::errors::Packager::MissingImport)) {
                 auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
                 e.setHeader("`{}` resolves but its package is not imported", lit.symbol.show(ctx));
-                bool isTestImport = otherFile.data(ctx).isPackagedTest();
+                bool isTestImport = otherFile.data(ctx).isPackagedTest() || ctx.file.data(ctx).isPackagedTest();
```
froydnj, line 459: Just to walk through my understanding here:

1. If `otherFile` is from a test package, then (almost, see below) obviously this should be a `test_import`.  Fine.
2. If `otherFile` is from a regular package, and we are from a test package, then `otherFile`'s package should be a `test_import`.  OK.
3. If `otherFile` is from a regular package, and we are from a regular package, then a plain `import` is called for.  Fine.

I think it's possible that we can wind up in a weird situation here, because there's a check above:

```c++
        // If the imported symbol comes from the test namespace, we must also be in the test namespace.
        if (otherFile.data(ctx).isPackagedTest() && !this->insideTestFile) {
            if (auto e = ctx.beginError(lit.loc, core::errors::Packager::UsedTestOnlyName)) {
                e.setHeader("`{}` is defined in a test namespace and cannot be referenced in a non-test file",
                            lit.symbol.show(ctx));
            }
        }
```

that doesn't return from this method.  So it is possible to get to this autocorrect with `otherFile` from a test package and us *not* from a test package.  I guess in that case we should not be suggesting an autocorrect and all?  Worth handling?  Maybe, but probably in a follow up.

froydnj (approved):

packager/VisibilityChecker.cc
jez: It was a one line change and I'm already changing that CLI test in this PR so I figured it'd be easier to change now.

PR #7231, opened 2023-08-14
Type check calls to `super` in classes

### LINKED ISSUE #6828: Type `super` inside a class (*not* a module)
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Breturns%28String%29%7D%0A%20%20def%20example%0A%20%20%20%20''%0A%20%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20def%20example%0A%20%20%20%20super%0A%20%20%20%20%23%20%5E%20hover%3A%20String%0A%20%20end%0Aend)

```ruby
# typed: true

class Parent
  extend T::Sig

  sig {returns(String)}
  def example
    ''
  end
end

class Child < Parent
  def example
    super
    # ^ hover: String
  end
end
```

#### Observed output

```
No errors! Great job.
```

#### Expected behavior

Properly typing `super` in the general case is hard, because doing it in modules is hard.

When using `super` in a method defined in a class, we should be able to rely on the inheritance hierarchy being fixed, and know which ancestor the `super` would dispatch to.

This would have some interesting implications:

- We would be able to check whether the arguments passed to the child match the arguments passed to the parent.
- Jump-to-def and hover would work
- The result would no longer be untyped (and thus usable in `# typed: strong` files.

- - -

Related is #1068.

jez:
- - - - -

> **Important**
>
> The Sorbet FAQ has been updated with docs on how to deal with the breaking changes in this PR:
>
> → [FAQ on `super` and types](https://sorbet.org/docs/faq#why-is-super-untyped-even-when-the-parent-method-has-a-sig)
>
> If you have feedback about this change, please [**open a new issue →**](https://github.com/sorbet/sorbet/issues/new/choose) 

- - - - -

### Motivation

Fixes #6828
Relates to #1068

For a long time, Sorbet did not typecheck any call to `super` because we did not have a story for how to typecheck calls to `super` in general. Ruby's inheritance model makes this difficult, because the meaning of `super` in a module changes depending on where the mixin eventually gets mixed into (so that "what method this call to `super` dispatches to" is not known only by knowing the ancestor information of the enclosing module).

Those limitations don't apply to classes, and so in this PR we relax them. There is one more main constraint: we don't typecheck calls to `super` inside **any** block. That block might be a block passed to `define_method` (or a block passed to a method that eventually calls `define_method`). It might also be a call to `class_eval`, etc. Rather than attempt to special case some of these cases, the current PR punts on them.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_OV91Xaad85x7AG
→ https://go/builds/bui_OV92m9Cu7am3Ze
→ https://go/builds/bui_OV92R8RLMDZU1F

jez:

→ https://go/builds/bui_OVROax8aWrSto1
→ https://go/builds/bui_OVROt6gqV8Icd4
→ https://go/builds/bui_OVRP6fKAC2KZzS

jez:

→ https://go/builds/bui_OVmT67QidBJY7N
→ https://go/builds/bui_OVmTQiJAlxAGCE
→ https://go/builds/bui_OVmTLtTkHu0BPI

jez:

→ https://go/builds/bui_OW7zFDrRo32afK
→ https://go/builds/bui_OW7z6ub4DzL6Tc
→ https://go/builds/bui_OW7zFG2q6Ch1Ia

jez:

→ https://go/builds/bui_OXdgn8qvJBUfmd
→ https://go/builds/bui_OXdgxhl658CrhY
→ https://go/builds/bui_OXdgF8KCH7gWbV

jez:

→ https://go/builds/bui_OY8AAtGB71kmNU
→ https://go/builds/bui_OY8ArxE4AGg0ZH
→ https://go/builds/bui_OY8ApSumPgr9Vl

jez:

→ https://go/builds/bui_OYAcaD1sFlRhJ7
→ https://go/builds/bui_OYAct2iKLW539v
→ https://go/builds/bui_OYAch83QipXub0

test/cli/arity-messages/test.sh
```diff
@@ -1,23 +1,32 @@
...
-echo
-echo --------------------------------------------------------------------------
-echo
+  echo
+  echo --------------------------------------------------------------------------
+  echo
 
-# Also cat the file, to make that the autocorrect applied
-cat arity-messages.rb
+  # Also cat the file, to make that the autocorrect applied
```
froydnj, line 24:
```suggestion
  # Also cat the file, to make sure that the autocorrect applied
```

website/docs/error-reference.md
```diff
@@ -4139,6 +4139,26 @@ the guarantees of the type system.
 This feature is opt-in. See [VS Code](vscode.md) for instructions on how to turn
 it on.
 
+## 7048
+
+This error is like [7003](#7003), but for calls using `super`.
+
+That is, then difference is that Sorbet checks whether a method with the same
```
froydnj, line 4146:
```suggestion
That is, the difference is that Sorbet checks whether a method with the same
```

local_vars/local_vars.cc
```diff
@@ -484,8 +492,32 @@ class LocalNameInserter {
...
         auto &original = ast::cast_tree_nonnull<ast::Send>(tree);
-        if (original.numPosArgs() == 1 && ast::isa_tree<ast::ZSuperArgs>(original.getPosArg(0))) {
-            tree = lowerZSuperArgs(ctx, std::move(tree));
+        // If the super calls occurs inside a block, or a module, we would like to leave it as untyped,
+        // rather than checking it. This code replaces those <super> with <untypedSuper>,
+        // and then dispatchCallSymbol will ignore those.
+        auto newMethodName = core::Names::untypedSuper();
+        if (ctx.state.typedSuper && !scopeStack.back().insideBlock && !scopeStack.back().insideModule) {
+            newMethodName = core::Names::super();
+        }
```
froydnj, line 501: I think at least the `insideBlock` check needs to be a little more aggressive, because of code like:

```ruby
foo.class_exec do
  def frob
    super
  end
end
```

(Similarly for `module_exec` and `instance_exec`, I guess?)

The code in its current form will cause Sorbet to expand the `super` call as though it was typed, but we don't have good static information about where that `frob` method is winding up.  Tests would be good here.

I am less sure that the `insideModule` check needs to more aggressively walk the stack, because for:

```ruby
module A
  class << self
    def foo
      super
    end
  end
end
```

I...think that comes out OK, because `foo` is defined on a (singleton) class, not a module?

It also would be slightly nicer, IMHO, to attempt to delay this computation until we decide that we need it, maybe pull the computation of `newMethodName` into a function and call it inside the two blocks below?

main/options/options.cc
```diff
@@ -409,6 +409,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                     cxxopts::value<string>()->default_value(empty.errorUrlBase), "url-base");
     options.add_options("advanced")("experimental-ruby3-keyword-args",
                                     "Enforce use of new (Ruby 3.0-style) keyword arguments", cxxopts::value<bool>());
+    options.add_options("advanced")("typed-super", "Enable typechecking of `super` calls when possible",
+                                    cxxopts::value<bool>()->default_value("true"));
```
froydnj, line 413: One of the nice (and maybe sometimes frustrating) things about Sorbet is that there aren't a profusion of options affecting how we do typechecking.  I didn't see an explanation of why we're adding this option here that does control that aspect.

My guess is that (as mentioned in the docs) there's no real `T.unsafe` escape hatch for `super` calls as there is for normal method calls, and so we need to provide this as a big hammer.  Is that roughly the idea?

website/docs/faq.md
```diff
@@ -258,9 +261,78 @@ To typecheck this example, Sorbet would have to typecheck `MyModule#foo`
...
+    `T.unsafe(self).super` to silence errors from a single usage of `super`
+    (`super` is a keyword in Ruby, not a method call).
+
+    This means that there is not a way to silence errors from a single usage of
+    `super`. The best you can do is silence all errors inside a method using
+    `T.bind`:
+
+    ```ruby
+    def example
+      T.bind(self, T.untyped) # <------
```
froydnj, line 321: This is simultaneously gross and wonderful.

core/types/calls.cc
```diff
@@ -664,9 +677,19 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                //   an instance/self method, suggest changing enclosing method definition.
+                // - If similar name exists in parent suggest renaming enclosing method definition
+                //   (intialize -> initialize, etc.)
+                // - Otherwise, suggest `T.bind`
+                e.addErrorNote("As a last resort, silence all `{}`-related errors using `{}`", "super",
+                               "--typed-super=false");
             } else if (args.receiverLoc().exists() &&
                        (gs.suggestUnsafe.has_value() ||
                         (args.fullType.type != args.thisType && symbol == Symbols::NilClass()))) {
+                // Explicitly ignore `<super>`, because super is not actually a method call with a receiver.
```
froydnj, line 692: This comment seems a little out of place now that `super` is explicitly handled above?

froydnj (changes requested):
I think the overall approach here is sound.

main/options/options.cc
jez: That’s the idea. I’m open to chat about this more

local_vars/local_vars.cc
jez: > I am less sure that the insideModule check needs to more aggressively walk the stack, because for:

yes, in this example, the call to `super` could actually hit a method, if the module `extend`'s something that provides a `foo` method. In that case, it's a singleton class, which is a class. For the same reason that an [abstract singleton method](https://sorbet.org/docs/abstract#abstract-singleton-methods) doesn't make sense, so too is it okay for us to treat these super calls statically.

Good test cases, thanks.

jez:

→ https://go/builds/bui_OYV7nIGbAj7Mwv
→ https://go/builds/bui_OYV71ZbDddp4sd
→ https://go/builds/bui_OYV70tM6OzJQOg

jez:

→ https://go/builds/bui_OYVCgXFwZCCigG
→ https://go/builds/bui_OYVCamH7szMxVx
→ https://go/builds/bui_OYVCZpDHmIDUL7

jez:

→ https://go/builds/bui_OYWBBePffgScGO
→ https://go/builds/bui_OYWBFlDb1M8qA2
→ https://go/builds/bui_OYWB0NnVviU72R

jez:

→ https://go/builds/bui_OYWozYs2YVLwSp
→ https://go/builds/bui_OYWoxo9pvNM58Z
→ https://go/builds/bui_OYWo6njMh3shkB

froydnj (approved):

core/types/calls.cc
```diff
@@ -664,6 +661,17 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                // TODO(jez) Special error for super.
+                // - If identical name exists as self/instance method in parent but we're currently
+                //   an instance/self method, suggest changing enclosing method definition.
+                // - If similar name exists in parent suggest renaming enclosing method definition
+                //   (intialize -> initialize, etc.)
+                // - Otherwise, suggest `T.bind`
+                // Note: super is not a method call, and so all the logic in the case below to
+                // compute autocorrects to potentially fix up the method call's receiver can't apply.
+                e.addErrorNote("As a last resort, silence all `{}`-related errors using `{}`", "super",
+                               "--typed-super=false");
```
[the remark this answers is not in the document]
jez: Deleted the note about going straight to `--typed-super=false`.

Added a link to the FAQ.

PR #7233, opened 2023-08-16
Translate `before do` to `def <before>`

jez:
### Motivation

This prevents Sorbet's rewriter pass from defining an `initialize` method,
which might then further interfere with `super` dispatch.

Unblocks #7212 

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/document_symbols_minitest.rb.document-symbols.exp
+++ b/test/testdata/lsp/document_symbols_minitest.rb.document-symbols.exp
@@ -29,32 +29,6 @@
             },
             "children": [
                 {
-                    "name": "<before>",
-                    "detail": "",
-                    "kind": 6,
-                    "range": {
-                        "start": {
-                            "line": 4,
-                            "character": 2
-                        },
-                        "end": {
-                            "line": 6,
-                            "character": 5
-                        }
-                    },
-                    "selectionRange": {
-                        "start": {
-                            "line": 4,
-                            "character": 2
-                        },
-                        "end": {
-                            "line": 6,
-                            "character": 5
-                        }
-                    },
-                    "children": []
-                },
-                {
                     "name": "describe 'with this block'",
                     "detail": "",
                     "kind": 5,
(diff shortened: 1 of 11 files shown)


test/testdata/rewriter/before_do.rb
jez, line 1: This test approximates what `minitest` does. Minitest tests subclass from a class that defines `initialize` with a single `name: String` parameter. If some other code ever subclassed minitest's testcase class and called `super` in the `initialize` method, it would only work as long as there wasn't a `before do` block declared above it in the ancestor chain. Stripe's codebase had a grandchildren class of the Minitest testcase class that did this.

I considered and rejected translating `before do` to `def initialize(name)` because the `before do` DSL also fires for RSpec if I'm not mistaken, where the `initialize` method has a different arity (it looked like it accepted multiple positional parameters). So I figured it better to simply special case this new `<before>` method inside Sorbet's core pipeline instead.

test/testdata/lsp/document_symbols_minitest.rb.document-symbols.exp
```diff
@@ -28,6 +28,32 @@
                 }
             },
             "children": [
+                {
+                    "name": "<before>",
```
froydnj, line 32: I think we want to hide the existence of `<before>` from document symbols.  Before, having `initialize` there was sort of understandable, but this PR changes that to be potentially confusing.

OTOH, maybe people would like seeing the `before` block as something obvious?  Maybe instead of filtering it, we should just give it a better name?  WDYT?

froydnj (approved):
This is kind of a hack, but I don't think there's a better way to model things here.  At least the special-casing is pretty localized.  Maybe as a followup, we should move checks to `isLikeInitializeName` or something so the logic is a little more centralized?  I guess that wouldn't prevent people from just checking `Names::initialize` later and forgetting that they need to handle `<before>`.

test/testdata/lsp/document_symbols_minitest.rb.document-symbols.exp
jez: I don't have a super strong opinion here.

It's worth noting that we already surface all the `it` blocks as fake methods where the string passed to the `it` block is in the method definition. So I think that the document outline in test files is already pretty weird, in that there's way more entries in the document symbols output than there are method definitions in the file.

But since you asked, I've written the code to skip it for now.

froydnj: > It's worth noting that we already surface all the `it` blocks as fake methods where the string passed to the `it` block is in the method definition. So I think that the document outline in test files is already pretty weird, in that there's way more entries in the document symbols output than there are method definitions in the file.

I think this is a happy consequence of the rewriter and convenient for the user, but I can understand how my happiness is not exactly consistent with my request earlier. 😅

jez: > a happy consequence of the rewriter

How much of Sorbet's success boils to this phrase

PR #6773, opened 2023-02-23
Add RBIs for `Regexp.timeout` and `Regexp.timeout=`

egiurleo:
### Motivation
Ruby 3.2 adds some new class methods to `Regexp`: `timeout` and `timeout=`, which are used to read and write a global regex timeout.

Read more about that here: https://bugs.ruby-lang.org/issues/17837

This PR adds RBIs for these methods!

### Test plan
See included automated tests.

PR #7235, opened 2023-08-16
Add Regexp::TimeoutError type

ronocod:
Support for setting the global timeout on `Regexp` was added in https://github.com/sorbet/sorbet/pull/6773, but the class for the error raised when the timeout is reached wasn't.

### Test plan
See included automated tests - I'm not sure if new error classes need any tests beyond being referenced in a `typed: true` test file.


froydnj (approved):
I made one small change so that the scoping on `TimeoutError` was correct.  Thanks for the patch!

PR #7241, opened 2023-08-18
Add four more methods to Kernel

jez:
cc @neilparikh 

### Motivation

Pulling some work out of #7212

### Test plan

n/a

### DIFF
--- a/rbi/core/kernel.rbi
+++ b/rbi/core/kernel.rbi
@@ -488,8 +488,12 @@ module Kernel
   sig {returns(Integer)}
   def hash(); end
 
+  private def initialize_clone(*args); end
+
   sig {params(object: T.self_type).returns(T.self_type)}
   def initialize_copy(object); end
+
+  private def initialize_dup(orig); end
 
   sig {returns(String)}
   def inspect(); end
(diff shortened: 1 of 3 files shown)


rbi/core/kernel.rbi
```diff
@@ -3162,4 +3162,20 @@ module Kernel
   # ```
   sig {returns(T.self_type)}
   def itself(); end
+
+  private def initialize_clone(*args); end
+
+  private def initialize_copy(orig); end
```
froydnj, line 3168: This is already defined earlier in the file?

https://github.com/sorbet/sorbet/blob/5a623a107dc081d1143508dbe8353220090a126d/rbi/core/kernel.rbi#L491-L492

jez: Ah thanks nice catch. I suppose I should also alphabetize these.

rbi/core/kernel.rbi
```diff
@@ -624,6 +628,16 @@ module Kernel
   end
   def respond_to?(arg0,include_all=false); end
 
+  # DO NOT USE THIS DIRECTLY.
```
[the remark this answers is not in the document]
jez: 😅 

Yeah we should probably at least have the comment from the source code:

```c
/*
 *  call-seq:
 *     obj.respond_to?(symbol, include_all=false) -> true or false
 *     obj.respond_to?(string, include_all=false) -> true or false
 *
 *  Returns +true+ if _obj_ responds to the given method.  Private and
 *  protected methods are included in the search only if the optional
 *  second parameter evaluates to +true+.
 *
 *  If the method is not implemented,
 *  as Process.fork on Windows, File.lchmod on GNU/Linux, etc.,
 *  false is returned.
 *
 *  If the method is not defined, <code>respond_to_missing?</code>
 *  method is called and the result is returned.
 *
 *  When the method name parameter is given as a string, the string is
 *  converted to a symbol.
 */
```

But it looks like our sync-doc.rb script chokes on `Kernel`-defined methods, so our `respond_to` method doesn't even have this comment.

I copied that "DO NOT USE" comment out of the Ruby VM source for `respond_to_missing?`.

ilyailya (approved):
The change looks good! There is a test failure in CI, but it should easily fixable by updating the test

PR #7243, opened 2023-08-18
Make the types for `T::Enum.inherited` more narrow

jez:
cc @neilparikh

### Motivation

Pulling out changes from #7212

### Test plan

n/a


ilyailya (commented):
Do you want to test this change on p-s? I'm not exactly sure what this change is affecting, but it feels like it might generate couple of type errors

jez:
Our script didn't post the comment, but the build passed.

ilyailya (approved):

PR #7246, opened 2023-08-21
add T.absurd as another equivalent to typescripts `never`

Nghi93:
### Test plan

See included automated tests.


jez (approved):
This looks great thanks, sorry for the delay.

I made some minor tweaks to provide a little more context. It will land whenever the build passes.

PR #7247, opened 2023-08-21
Allow potentially using THP (transparent huge pages) in sorbet_ruby

maruth-stripe:
### Motivation

The Ruby interpreter outright disables THP since it might have bad interactions with dirty pages (i.e., if a single bit in a 2MB page is dirtied now the entire things needs to be invalidated/flushed instead of just 4K) w.r.t GC etc. However, we would like to experiment with enabling it for memory-intensive workloads 

### Test plan

See included automated tests.

### DIFF
--- a/third_party/ruby/thp.patch
+++ b/third_party/ruby/thp.patch
@@ -1,18 +1,14 @@
 diff --git eval.c eval.c
-index 844c537cc4..f15ee3d2ea 100644
+index 844c537cc4..0fd751650a 100644
 --- eval.c
 +++ eval.c
-@@ -72,13 +72,6 @@ ruby_setup(void)
- 
-     ruby_init_stack((void *)&state);
- 
--    /*
--     * Disable THP early before mallocs happen because we want this to
--     * affect as many future pages as possible for CoW-friendliness
--     */
--#if defined(__linux__) && defined(PR_SET_THP_DISABLE)
--    prctl(PR_SET_THP_DISABLE, 1, 0, 0, 0);
--#endif
+@@ -77,7 +77,9 @@ ruby_setup(void)
+      * affect as many future pages as possible for CoW-friendliness
+      */
+ #if defined(__linux__) && defined(PR_SET_THP_DISABLE)
++  if (strncmp(getenv("STRIPE_RUBY_ENABLE_THP"), "true", 4) != 0) {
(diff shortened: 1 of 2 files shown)


third_party/ruby/thp.patch
```diff
@@ -0,0 +1,18 @@
...
++++ eval.c
+@@ -72,13 +72,6 @@ ruby_setup(void)
+ 
+     ruby_init_stack((void *)&state);
+ 
+-    /*
+-     * Disable THP early before mallocs happen because we want this to
+-     * affect as many future pages as possible for CoW-friendliness
+-     */
+-#if defined(__linux__) && defined(PR_SET_THP_DISABLE)
```
fable-stripe, line 13: I suggest adding 

```
if (strncmp(getenv("STRIPE_RUBY_ENABLE_THP"), "true", 4) != 0) {
...
}
```

here

third_party/ruby_externals.bzl
```diff
@@ -59,6 +59,7 @@ def register_ruby_dependencies():
         strip_prefix = strip_prefix,
         build_file = ruby_build,
         patches = [
+            "@com_stripe_ruby_typer//third_party/ruby:thp.patch",
```
amytsai-stripe, line 62: Can you also include the patch in 3.1 below?

jez, line 62: Is it important that you put the patch first? We have traditionally put all new patches at the bottom.

maruth-stripe: ack I'll put it at the bottom

maruth-stripe:
Verified this works by running
` bazel-bin/external/sorbet_ruby_2_7/ruby -e "x = []; i = 1; while true; x << 'abc' * i; end"`

And then on doing `cat /proc/<pid>/status` got
```
THP_enabled:	0
```

On running `STRIPE_RUBY_ENABLE_THP="true" bazel-bin/external/sorbet_ruby_2_7/ruby -e "x = []; i = 1; while true; x << 'abc' * i; end"`

And then on doing `cat /proc/<pid>/status` got
```
THP_enabled:	1
```

maruth-stripe:
^also reprod above with `sorbet_ruby_3_1`

jez (approved):

PR #7249, opened 2023-08-22
Factor out enrichMissingMethodError

jez:
### Motivation

Preparation for #7212, which will add types for certain `super` calls.

This code will all look a bit different when it's changed to support `super`,
because almost none of these autocorrects / suggestions apply to `super`.

With that in mind, it looks better if we can early exit from this method after
seeing that the error was caused by a call to `super`. I'd like to get this
refactor in now, so that the eventual diff is smaller.

### Test plan

Existing tests.


jez:
Thought of a better way to factor this.

Issue #6832, opened 2023-03-09
Allow specify type of block, without needing `&block` param.


amomchilov:
### Problem

It would be nice to be able to be able to specify the type of a block parameter, without needing to add an explicit block paramter.

```ruby
# typed: strict
extend T::Sig

sig { params(block: T.proc.params(arg0: String).void).void }
# The explicit block param is required by Sorbet, but not Ruby itself
def has_block_param # (&block)
  yield "ABC"
end

has_block_param { |s| puts(s.downcase) }
```

[Sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28block%3A%20T.proc.params%28arg0%3A%20String%29.void%29.void%20%7D%0Adef%20has_block_param%20%23%20%28%26block%29%0A%20%20yield%20%22ABC%22%0Aend%0A%0Ahas_block_param%20%7B%20%7Cs%7C%20puts%28s.downcase%29%20%7D)

### Motivation

Prior to Ruby 2.5, adding an explicit `&block` parameter to a method involved eagerly materializing a Proc object, even if it's ultimately only `yield`ed to. This was [improved in Ruby 2.5](https://bugs.ruby-lang.org/issues/14045), but a performance discrepancy still exists.

|               | Ruby 2.4       | Ruby 2.5       |
|--------------:|---------------:|---------------:|
| block + yield |  2146619.5 i/s | 12630378.1 i/s |
| yield         | 10436414.0 i/s | 15351931.0 i/s |
| Speedup:      |          4.86x |          1.21x |

Data from [fast-ruby](https://github.com/fastruby/fast-ruby#proccall-and-block-arguments-vs-yieldcode). [Benchmark source](https://github.com/fastruby/fast-ruby/blob/main/code/proc-and-block/proc-call-vs-yield.rb).

It would be nice to be able to add type checking to hot paths that want to use `yield` for a perf benefit.

### Proposed solution

I see two main approaches for how this could be solved:

#### Add a reserved parameter name

... that's understood to be referring to a block parameter, whether explicitly listed in the parameters or not

```ruby
sig { params(i: Integer, block: T.proc.params(arg0: String).void).void }
def has_block_param(i) # Block not explicitly listed
  yield "ABC #{i}"
end
```

* Pro: simple
* Pro: fits into params as expected
* Con: adding a reserved identifier is a breaking change

#### Add new method to the sig DSL

[...]

PR #7250, opened 2023-08-23
Add new `"...": T.untyped` syntax for argument forwarding

jez:
### Motivation

Requested by a user.

Previously, Sorbet would force you to write out the args individually if you
wanted to give them a certain type.

Some people expressed annoyance at that, as it conflicts with certain style
guides. If you're content to have all the arguments have the same type (usually:
`T.untyped`), there's no reason why Sorbet needs to reject using `...`.

However, style guides that mandate using `...` always instead of `*args,
**kwargs, &blk` will not work with Sorbet if users want to provide types for
those things independently. For example, if you want your method defined using
`...` to take only `Integer` positional params but only `String` keyword
arguments, you'll still need to split the `...` into `*args, **kwargs, &blk` to
have a name you can use to provide those types independently.

### Test plan

See included automated tests.

### DIFF
--- a/core/tools/generate_names.cc
+++ b/core/tools/generate_names.cc
@@ -419,6 +419,7 @@ NameDef names[] = {
     {"fwdArgs", "<fwd-args>"},
     {"fwdKwargs", "<fwd-kwargs>"},
     {"fwdBlock", "<fwd-block>"},
+    {"fwdDots", "..."},
 
     // Enumerable#flat_map has special-case logic in Infer
     {"flatMap", "flat_map"},
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -3603,17 +3603,22 @@ private:
 
         OverloadedMethodSigInformation info;
         if (usesArgumentForwardingSyntax(ctx, methodInfo, mdef, isOverloaded)) {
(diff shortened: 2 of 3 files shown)


website/docs/faq.md
```diff
@@ -510,3 +510,40 @@ There are also third-party tools that offer the ability to sort and filter
...
+`String` keyword parameters, Sorbet requires split the `...` into
+`*args, **kwargs, &blk` so that there's a name associated with each component of
+the parameters.
+
+```ruby
+sig {params(args: Integer, kwargs: String, blk: T.proc.void).void}
+def foo(*args, **kwargs, &blk)
+  bar(*args, **kwargs, &blk)
+end
+```
```
jez, line 549: Technically speaking, we could extend the approach in this PR to support this use case:

```ruby
sig {params("*": Integer, "**": String, "&": T.proc.void).void}
def foo(...)
  bar(...)
end
```

But you'd just immediately run up against the problem that Sorbet doesn't allow passing splats when the length isn't known.

There's other ways we can probably fix that, and then maybe revisit this.

Also worth noting that we may want to build the block arg syntax independently, for #6832.

jez:
Some tests failed, and also this is going to need runtime support.

PR #7252, opened 2023-08-24
Update Psych RBI

vinistock:
Update the Psych RBI to match the current definitions in the [repo](https://github.dev/ruby/psych).

### Motivation

I noticed that the Tapioca generated Psych RBI was conflicting with the Sorbet stdlib RBI, so I looked into whether the changes really existed in the upstream gem.

### Test plan

N/A.

### DIFF
--- a/rbi/stdlib/psych.rbi
+++ b/rbi/stdlib/psych.rbi
@@ -318,7 +318,6 @@ module Psych
   sig do
     params(
       yaml: T.any(String, StringIO, IO),
-      legacy_filename: Object,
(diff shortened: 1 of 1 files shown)


froydnj (approved):
> → https://go/builds/bui_OYenKhFPDwrZ78
> → https://go/builds/bui_OYenmWgoyBllSD
> → https://go/builds/bui_OYenPBm6qzneLj

The ARM64 build is broken for unrelated reasons, but the x86-64 build passed, so I think this is OK!

Thanks!

PR #7255, opened 2023-08-25
✨ Emit status/operation changes through VSCode extension's MetricsEmitter

marianosimone:
### Motivation
This extends the range of metrics the extension currently emit, so consumers that register `sorbet.metrics.getExportedApi` can get notified when a status/operation change happens.

Among other things, this would allow to know when Sorbet is available/done, and how long it took to get into that state.

### Test plan
Manually tested the extension, and saw the metrics being emitted.

### DIFF
--- a/vscode_extension/src/sorbetStatusMetricsTracker.ts
+++ b/vscode_extension/src/sorbetStatusMetricsTracker.ts
@@ -1,0 +1,45 @@
+import { Disposable } from "vscode";
+
+import { SorbetExtensionContext } from "./sorbetExtensionContext";
+import { ServerStatus } from "./types";
+
+/**
+ * A Disposable that listens to status changes through the SorbetStatusProvider,
+ * and emit metrics about them through the configured MetricsEmitter
+ */
+export class SorbetStatusMetricsTracker implements Disposable {
+  private readonly context: SorbetExtensionContext;
+  private readonly disposable: Disposable;
+
+  constructor(context: SorbetExtensionContext) {
+    this.context = context;
+
+    this.disposable = Disposable.from(
+      this.context.statusProvider.onStatusChanged((e) => {
+        console.log(
+          `status changed: ${ServerStatus[e.status]} (error: '${e.error}')`,
+        );
+        this.context.metrics.emitTimingMetric("status.changed", new Date(), {
+          status: ServerStatus[e.status],
+          error: e.error || "",
+        });
+      }),
+      this.context.statusProvider.onShowOperation((_params) => {
+        console.log(
+          `operation changed: ${_params.operationName} went to ${_params.status}`,
+        );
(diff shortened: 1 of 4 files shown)


vscode_extension/src/types.ts
```diff
@@ -29,7 +29,7 @@ export enum RestartReason {
 }
 
 // Note: Sorbet is either running/in the process or running, or in an error state. There's no benign idle/not running state.
-export const enum ServerStatus {
```
marianosimone, line 32: This is so we can call `ServerStatus[e.status]`. When you use a `const` enum, that's not possible :(

We could also just send the number as the status, but that looks pretty obscure for 3rd parties trying to do anything with it.

marianosimone: Another alternative is to manually maintain a mapping from numbers to names, which I don't love

jez (commented):
This is fine by me; I'm going to defer to @damolina-stripe for another review, as he usually has more to say about how to write good vscode_extension code.

vscode_extension/src/types.ts
damolina-stripe: Makes sense - I've run into the same situation in the past.

vscode_extension/src/sorbetStatusMetricsTracker.ts
```diff
@@ -0,0 +1,45 @@
...
+  private readonly disposable: Disposable;
+
+  constructor(context: SorbetExtensionContext) {
+    this.context = context;
+
+    this.disposable = Disposable.from(
+      this.context.statusProvider.onStatusChanged((e) => {
+        console.log(
+          `status changed: ${ServerStatus[e.status]} (error: '${e.error}')`,
+        );
```
damolina-stripe, line 21: There is a {{context.log}} for this (probably a `trace`)? Same for other cases.

vscode_extension/src/sorbetStatusMetricsTracker.ts
```diff
@@ -0,0 +1,45 @@
...
+        this.context.metrics.emitTimingMetric("status.changed", new Date(), {
+          status: ServerStatus[e.status],
+          error: e.error || "",
+        });
+      }),
+      this.context.statusProvider.onShowOperation((_params) => {
```
damolina-stripe, line 27: This is used in line 29, should be `params`.

marianosimone: Fixed

vscode_extension/src/sorbetStatusMetricsTracker.ts
marianosimone: Oh, this is just a leftover of my testing, 🔥

damolina-stripe (approved):

PR #7263, opened 2023-08-28
Change Struct rewriter to make intermediate class

### LINKED ISSUE #3157: Custom initializers for Struct are not seen
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AFoo%20%3D%20Struct.new(%3Aa%2C%20%3Ab%2C%20keyword_init%3A%20true)%20do%0A%20%20def%20initialize(a)%0A%20%20%20%20super(a%3A%20a%2C%20b%3A%20%22b%22)%0A%20%20end%0Aend%0A%0Afoo%20%3D%20Foo.new(%22a%22))

```ruby
# typed: true

Foo = Struct.new(:a, :b, keyword_init: true) do
  def initialize(a)
    super(a: a, b: "b")
  end
end

foo = Foo.new("a")
```

#### Observed output

```
editor.rb:3: Method Foo#initialize redefined without matching argument count. Expected: 1, got: 2 https://srb.help/4010
     3 |Foo = Struct.new(:a, :b, keyword_init: true) do
     4 |  def initialize(a)
     5 |    super(a: a, b: "b")
     6 |  end
     7 |end
    editor.rb:4: Previous definition
     4 |  def initialize(a)
          ^^^^^^^^^^^^^^^^^

editor.rb:9: Too many positional arguments provided for method Foo#initialize. Expected: 0, got: 1 https://srb.help/7004
     9 |foo = Foo.new("a")
              ^^^^^^^^^^^^
    editor.rb:3: initialize defined here
     3 |Foo = Struct.new(:a, :b, keyword_init: true) do
     4 |  def initialize(a)
     5 |    super(a: a, b: "b")
     6 |  end
     7 |end
    editor.rb:9: Foo#initialize has optional keyword arguments. Did you mean to provide a value for a?
     9 |foo = Foo.new("a")
              ^^^^^^^^^^^^
Errors: 2
```

#### Expected behavior

This shouldn't raise any type error.

- - -

jez:
### Motivation

There's a downside now that you can't use `Struct.new` in a
`# typed: strict` file. We may wish to find a way around that.

Fixes #3157

### Test plan

- [x] @jez Write tests


jez:

→ https://go/builds/bui_OXGIBxk3K4ffIz
→ https://go/builds/bui_OXGIi3vhKSlDFG
→ https://go/builds/bui_OXGIRmtFcmwwTT

jez:

→ https://go/builds/bui_OXHBeLDfJiUr2a
→ https://go/builds/bui_OXHBq5kUa43l3u
→ https://go/builds/bui_OXHBYny1KbcsLL

test/cli/struct_strict/test.out
jez, line 1: Deleted because I converted this to a testdata test.

test/cli/struct_strict/test.out
```diff
@@ -1,14 +0,0 @@
...
-  Note:
-    Struct classes defined with `Struct` are hard to use in `# typed: strict` files.
-    Consider using `T::Struct` instead.
-
-test/cli/struct_strict/test.rb:7: The method `foo=` does not have a `sig` https://srb.help/7017
-     7 |Bad = Struct.new(:foo) do
-                          ^^^
-  Note:
-    Struct classes defined with `Struct` are hard to use in `# typed: strict` files.
-    Consider using `T::Struct` instead.
```
jez, line 13: These errors don't happen anymore—we may want to consider deleting this special handling.

test/testdata/lsp/struct_fuzz.rb
jez, line 1: Moved to a CLI test, because updating these no-stdlib tests whenever something in the rewriter pass changes is annoying. Much easier to have a single command that re-records the new behavior.

test/testdata/rewriter/struct.rb.rewrite-tree.exp
```diff
@@ -13,29 +13,45 @@ class <emptyTree><<C <root>>> < (::<todo sym>)
...
+        ::Kernel.raise("Sorbet rewriter pass partially unimplemented")
       end
 
       ::Sorbet::Private::Static.sig(::T::Sig::WithoutRuntime) do ||
         <self>.params(:foo, ::BasicObject, :bar, ::BasicObject).void()
       end
 
       def initialize<<todo method>>(foo = nil, bar = nil, &<blk>)
-        ::T.unsafe(::Kernel).raise("Sorbet rewriter pass partially unimplemented")
+        ::Kernel.raise("Sorbet rewriter pass partially unimplemented")
```
jez, line 54: This makes it easier to use `Struct.new` in a `# typed: strong` file (though it'll still be imperfect, as the call to `super` will look like a usage of `T.untyped`.

rewriter/Struct.cc
```diff
@@ -22,30 +22,30 @@ bool isKeywordInitKey(const core::GlobalState &gs, const ast::ExpressionPtr &nod
...
+    ast::cast_tree_nonnull<ast::Send>(typeMember)
+        .setBlock(ast::MK::Block0(
+            loc, ast::MK::Hash1(loc, ast::MK::Symbol(loc, core::Names::fixed()), ast::MK::Untyped(loc))));
+    return ast::MK::Assign(loc, ast::MK::UnresolvedConstant(loc, ast::MK::EmptyTree(), core::Names::Constants::Elem()),
+                           std::move(typeMember));
+}
 
-        for (auto &&stat : insSeq->stats) {
-            methodDef = ast::cast_tree<ast::MethodDef>(stat);
+void selfScopeToEmptyTree(ast::UnresolvedConstantLit &cnst) {
```
jez, line 35: This function was required because we have a test called `struct_self.rb` that tests

```ruby
self::A = Struct.new(:foo)
```

and then we had an ENFORCE that said that everything in the `ancestors` list is always a ConstantLit or UnresolvedConstantLit.

rewriter/Struct.h
```diff
@@ -7,20 +7,41 @@ namespace sorbet::rewriter {
...
+ *       sig {params(arg: T.untyped).returns(T.untyped)}
  *       def bar=(arg); arg; end
+ *
  *       sig {params(foo: BasicObject, bar: BasicObject).returns(A)}
- *       def self.new(foo=nil, bar=nil)
- *           T.cast(nil, A)
- *       end
+ *       def initialize(foo=nil, bar=nil); end
+ *
+ *       Elem = type_member {{fixed: T.untyped}}
```
ilyailya, line 32: Potentially a breaking change: do we want to rename type_member to be something Struct specific like `StructElement`?  `Elem` is a reserved name for type members of `Enumerable`. However, a clash between Struct and Enumerable is pretty unlikely, but maybe we should give non-generic names to the type members and the type elements we use internally ? It will minimize a chance of potential collisions in general and also will be a good example to the users.

ilyailya (approved):
The change looks good!
I left couple of non-blocking questions

rewriter/Struct.h
```diff
@@ -7,20 +7,41 @@ namespace sorbet::rewriter {
...
- *   A = Struct.new(:foo, :bar)
+ *   A = Struct.new(:foo, :bar) do
+ *       # ... body ...
+ *   end
  *
  * into
  *
- *   class A < Struct
+ *   class A$1 < Struct
+ *       sig {returns(T.untyped)}
```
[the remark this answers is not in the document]
jez: If we made them accept `BasicObject`, any override that the user defined with the goal of narrowing the types (like [seen here](https://github.com/sorbet/sorbet/pull/7263/files/0c00438ad7356013affb68063376158b9c07cce3#diff-e9863ca12988e7b7c845d9f20b8a5d8a56bd915e07900ae5105284c683886942R11)) would be an incompatible override.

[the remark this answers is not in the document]
jez: If we made them return `BasicObject`, it would require `T.cast` on the result type in the same test example.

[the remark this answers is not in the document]
jez: Both of those are breaking changes

Realistically, using BasicObject here (or T.anything) is not going to save you. If you're using `Struct`, you're using `T.untyped`, and you should probably switch to `T::Struct`.

rewriter/Struct.h
jez: This was already defined like this, and it has to be defined like this because `Struct` is `Enumerable`. That is, it's intentional that this uses the name `Elem`, because that's what `Enumerable` uses.

jez: https://github.com/sorbet/sorbet/blob/978f5f0c48fa18ea6c5d2835fd01247ac0c23f96/rbi/core/struct.rbi#L32-L35

PR #7268, opened 2023-08-29
Better exception types

ilyailya:
This is a continuation of @froydnj PR https://github.com/sorbet/sorbet/pull/6937

Nathan's PR had one corner case, when the exception type was not propagated to the ensure and appeared as `T.untyped`

```
begin
  func
rescue => e
  raise
ensure
  T.reveal_type(e) # => T.untyped
  # We can do better than that ^
end
```

Currently we don't model the case when the exception is accessed in the ensure block. To model it I've added a basic block before the rescue's basic block with a conditional jump to the ensure.

Here is an exception handling on eb2139f45a2babe85226a99c907d88fa12222389:
Screenshot 2023-08-29 at 16 05 01

And here is the same case on the tip of this branch:
Screenshot 2023-08-29 at 16 06 25

UPD:
This approach didn't workout. It fails to typecheck use cases like this:
```ruby
sig {returns(T::Boolean)}
def foo
  begin 
    true
  rescue
    false
  end
end
# ^ Sorbet reports the method returns `T.nilable(T::Boolean), that contradicts with a signature`
```

This happens because of this piece of code:
https://github.com/sorbet/sorbet/blob/239a925b2f8bbb09657fd187476be54ddf0b7f47/infer/environment.cc#L150-L177
In the example above it ties the truthiness of the exception variable to the truthiness of the return type. Not sure how that can be fixed for exceptions

### Motivation

Better types for exceptions!

### Test plan

See included automated tests.


cfg/builder/builder_walk.cc
```diff
@@ -823,14 +823,58 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
...
+                    //         │                  │             │
+                    //   ┌─────▼─────┐      ┌─────▼─────┐       │
+                    //   │           │      │           │       │
+                    //   │ ensure bb │      │ ensure bb ◄───────┘
+                    //   │           │      │           │
+                    //   └───────────┘      └───────────┘
+                    //
+                    // See test/testdata/cfg/ensure_exception.rb for a common use case
+                    auto shouldSplitCaseBody = !ast::isa_tree<ast::EmptyTree>(a.ensure);
+                    if (shouldSplitCaseBody) {
```
ilyailya, line 864: It's important to add an extra block only if the ensure block is present. 
If we would add an extra block every time, when this program would fail to typecheck with an error saying the return type is `T.nilable(T::Boolean)`
```ruby
sig {returns(T::Boolean)}
def foo
  begin 
    true
  rescue
    false
  end
end
```

core/tools/generate_names.cc
```diff
@@ -95,6 +95,8 @@ NameDef names[] = {
     {"magic", "<magic>"},
     {"unconditional", "<unconditional>"},
     {"argPresent", "<argPresent>"},
+    {"rescueHeaderTemp", "<rescueHeaderTemp>"},
+    {"rescueTrueTemp", "<rescueTrueTemp>"},
```
froydnj, line 99: Can you add these to the magic function at https://github.com/sorbet/sorbet/blob/78116cac95a86afc13d5662e14db1ea2874bba92/core/LocalVariable.cc#L13-L28 ?

Or, possibly better yet, add special names that bracket these CFG temporaries:

```c++
    {"cfgTemporaryStart", "<cfgTemporaryStart>"},
    // existing temporaries go here
    {"cfgTemporaryEnd", "<cfgTemporaryEnd>"},

    ...
```

and make the above function do a range check on the ID of its argument to ensure it's between the IDs of those magic names?  (I think that would be best as a separate PR.)

froydnj (changes requested):
Wow, cool!

I'm requesting changes for some small comments below and two big reasons:

1. I would like to see this tested on Stripe's codebase before reviewing, so we can both look over the errors (or, optimistically, the lack thereof!) to see if there's any more interesting cases we need to handle.  cc... @Morriar , I guess, to see if Shopify would like to pre-test this change for the same reason.
2. The compiler is broken due to this change.  I'm happy to fix that, but I would prefer to do it only once we verify that we don't have to do a bunch of other work because of certain patterns that we didn't think about.  (If you're curious, the design of exception handling in the compiler fundamentally doesn't support branches between exception regions; I think we can (hackily) allow this in the compiler by checking the variable involved in the conditional jump for the newly-introduced edge and cutting the appropriate edge for `rescueHeaderTemp`.  The edge that we're putting here isn't really necessary for control flow of the program, and is only there to support type propagation in the static analysis.)

jez:

→ https://go/builds/bui_OXy207WTsA9ieD
→ https://go/builds/bui_OXy2CTT2mqqo6c
→ https://go/builds/bui_OXy2RdIAaaLXku

PR #7275, opened 2023-08-31
Only use `initialize` DispatchResult if we found something

jez:
### Motivation

It's a lot less error prone to take the whole dispatch component, rather
than pluck individual pieces off of it. But we should only do that if we
actually _want_ it. If the dispatch to `initialize` from `Class#new` fails,
we would overwrite the original dispatch to `Class#new`, and then claw back
some of the things we overwrote. While working on typing super, I realized
that we weren't copying the `constr` back if
`initialize` didn't exist, so it could end up that you'd dispatch to a
method that `.exists()` but you'd have `constr == nullptr` which is bad.

We had a test for this on the WIP changes to super. I wanted to do this
refactor as pre-work to prove that it's a no-op normally. This will end up
getting tested indirectly by way of the tests that will come in service of
`super`.

<br>

The first commit makes it so that if we didn't successfully dispatch to
an `initialize` method, we don't report any errors.

But we actually do want some errors: if you dispatch to `initialize` and it
"didn't exist" that used to mean that it dispatched to
`BasicObject#initialize`. There was some comment about why that was
undesirable to model in RBIs, but I think that doesn't apply anymore. We
can simply let the dispatch to `BasicObject#initialize` happen, and then
Sorbet will report errors like normal because `BasicObject#initialize` does
not take any arguments.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -371,6 +371,8 @@ void GlobalState::initEmpty() {
     ENFORCE(klass == Symbols::Class());
     klass = synthesizeClass(core::Names::Constants::BasicObject(), 0);
     ENFORCE(klass == Symbols::BasicObject());
+    method = enterMethod(*this, Symbols::BasicObject(), Names::initialize()).build();
+    ENFORCE(method == Symbols::BasicObject_initialize());
     klass = synthesizeClass(core::Names::Constants::Kernel(), 0, true);
     ENFORCE(klass == Symbols::Kernel());
     klass = synthesizeClass(core::Names::Constants::Range());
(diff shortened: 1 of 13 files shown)


jez:

→ https://go/builds/bui_OYQF2kROaOT6HZ
→ https://go/builds/bui_OYQFfJKbQdgny0
→ https://go/builds/bui_OYQFnwpIQWqwN9

jez:
Oh whoops, looks like I had #7231 stacked on top of this, but didn't reflect that in the PR base. So these changes got included via #7231 landing.

PR #7289, opened 2023-09-05
Update IRB RBI to match the latest version

vinistock:
Update the parent classes of `RelineInputMethod` and `ReadlineInputMethod` to match the latest version of IRB.

### Motivation

The parent classes changed on this PR https://github.com/ruby/irb/pull/671.

### Test plan

N/A

PR #7295, opened 2023-09-06
Revert "Update IRB RBI to match the latest version (#7289)"

froydnj:
This reverts commit 7d3ddd6e4bccd439a289cf011a052c0b8b3b9be0.

### Motivation

This breaks anybody on older versions of `irb` with autogenerated RBIs.  My mistake for merging this.

I'm not sure what the right fix here is; it's possible that we should just not have `irb` RBIs in the stdlib, since it has been decoupled from Ruby releases for a while?

cc @vinistock 

### Test plan

See included automated tests.

### DIFF
--- a/rbi/stdlib/irb.rbi
+++ b/rbi/stdlib/irb.rbi
@@ -1525,7 +1525,7 @@ class IRB::OutputMethod::NotImplementedError < ::StandardError; end
 
 class IRB::ReidlineInputMethod < ::IRB::RelineInputMethod; end
 
-class IRB::RelineInputMethod < ::IRB::StdioInputMethod
+class IRB::RelineInputMethod < ::IRB::InputMethod
   include(::Reline)
 
   # Creates a new input method object using
@@ -1561,7 +1561,7 @@ class IRB::RelineInputMethod < ::IRB::StdioInputMethod
   def line(line_no); end
 end
 
-class IRB::ReadlineInputMethod < ::IRB::StdioInputMethod
+class IRB::ReadlineInputMethod < ::IRB::InputMethod
   include(::Readline)
 
   # Creates a new input method object using


ilyailya (approved):

froydnj:
> I think removing it from the payload could cause trouble too, for people who are not including `irb` explicitly in their `Gemfile`.

`irb` is not explicitly listed in Stripe's `Gemfile`, but we still generate RBIs for it, because it's included in the dependency graph.  Is that not the same thing that happens for e.g. Tapioca?

ilyailya (changes requested):
The change looks good, but it seems we don't have a great roll out story yet. Will put "Request changes", so we can't merge PR my mistake and block a portion of our users

vinistock:
That depends if `irb` is a transitive dependency of the application indeed. If you have `debug` in the `Gemfile`, for example, then yes it'll be picked up by Tapioca because it's a transitive dependency.

But if `irb` is not a transitive dependency, Tapioca won't generate RBIs for it, but you could still require and use it.

@paracycle mentioned that Tapioca was not supposed to try to generate RBIs for symbols that are already in the Sorbet payload, so maybe the problem lies there?

froydnj:
> But if `irb` is not a transitive dependency, Tapioca won't generate RBIs for it, but you could still require and use it.

I feel like that is on the user to some extent.  If you are using a version that is compatible with Sorbet's payload, that's not a problem...and even if you are using a version that isn't compatible, there are no RBIs, so it probably doesn't matter?  (If you are writing sigs or method calls that depend on the internals of the class hierarchy of the gem, but you don't have RBIs accurately describing that hierarchy, that seems like it is on you, not Sorbet.)

> If we revert the PR, people on the latest IRB have no path forward.

It seems like dropping down to not be on the latest version of `irb` would be sufficient, no?

ilyailya (approved):

vinistock:
> and even if you are using a version that isn't compatible, there are no RBIs

If any gems that are required by the application have IRB as a transitive dependency, Tapioca will generate an RBI for it and just by doing that it will fail typechecking because Sorbet's internal RBIs don't match, with no escape hatch since `typed: false` on the RBI won't work.

IRB 1.8.0 contains many bug fixes and improvements that benefit not only IRB itself, but Ruby debug too. I think we should always create the incentive for people to upgrade their development tools, so I really don't see downgrading as a solution.

olivier-thatch:
I ran into this problem after trying to upgrade irb to the latest version.

Is this issue being tracked anywhere? I'd like to be notified when this has been solved and it's safe to upgrade, but I couldn't find an open issue either on this repo or on Tapioca's.

tonybruess:
We're running into this issue too trying to upgrade to 1.8.0

PR #7297, opened 2023-09-06
Bound exception variable after matching exception

ilyailya:
This is a continuation of https://github.com/sorbet/sorbet/pull/6937

The goal of the original PR was to provide more accurate modeling of the Ruby's exception handling. The PR was blocked on use cases like that

```ruby
begin 
  foo
rescue => exc
  handle(exc)
  raise
ensure
  log(exc.message)
  # ^ this line would be reported as unreachable code
end
```

To fix this corner case I've assigned exception value to be `T.untyped` in the ensure body in the CFG. With the change in this PR example above would look like:

```ruby
begin 
  foo
rescue => exc
  handle(exc)
  raise
ensure
  exc = T.unsfe(exc)
  log(exc.message)
  # ^ this line would be reported as unreachable code
end
```

I've also tried a more elaborate approach, which doesn't require unsafing the exception, but it didn't work out. 
Feel free to check out the description of [this PR](https://github.com/sorbet/sorbet/pull/7268) to learn more

### Test plan

See included automated tests.


froydnj (changes requested):
I am getting merge conflicts when trying to test this PR on pay-server, which seems extremely odd.  Can you fix that?  I think rebasing and dropping the merge commits might help?

ilyailya:

→ https://go/builds/bui_OcZFgh6zYeS4jO
→ https://go/builds/bui_OcZFhYLpF8gRAR
→ https://go/builds/bui_OcZFoH5JoK24o5

PR #7298, opened 2023-09-06
Fix java static release

tonybruess:
### Motivation
Java release is failing because we're referencing `universal-darwin-18`

### Test plan
Yolo


froydnj (approved):
Thanks for following that!

PR #7308, opened 2023-09-13
write a global header for v5 autogen depdb

froydnj:
### Motivation

The current autogen depdb format for a file that Sorbet knows about looks like this; each line of the table represents one msgpack object of the appropriate kind:

```
|----------------|
| "symbols"      |
| string array   |
| "ref_count"    |
| int            |
| "def_count"    |
| int            |
| "ref_attrs"    |
| string array   |
| "def_attrs"    |
| string array   |
|----------------|
| ...actual data |
|----------------|
```

The DB is then generated by writing out the above format for each file Sorbet knows about and concatenating them together.

This format has the virtue of being fairly self-describing (the header is stored as a msgpack map, so reading it is a single line of Ruby), but it also has the downside of being somewhat verbose: we're writing the strings in the above header for every file, and the string arrays for `ref_attrs` and `def_attrs` are duplicated for every file.

This PR changes things so we write a single, global header at the start of the file describing the fields in the header for each individual file, as well as the `ref_attrs` and `def_attrs` used throughout the file.  This does make processing the files slightly more complicated, as there's more state to pass around, but we now have to parse less per-file and we do less I/O for the entire DB, so I think the tradeoff is worth it.

This change cuts a little over 10% off the size of depdb for Stripe's codebase.

I meant to add a version number at the very start of the header; I will do that in a followup PR.

I don't love the conditionals in `msgpack.cc`; I think that once v5 stabilizes and is supported in Stripe's codebase, we could simply remove support for previous versions.  Then we would only really support three states:

1. A single, current version
2. The current version and the next version
3. (After some time) Make the next version the current and drop the support for the previous, bringing us back to state 1

### Test plan

See included automated tests.

### DIFF
--- a/main/autogen/data/definitions.cc
+++ b/main/autogen/data/definitions.cc
@@ -174,4 +174,8 @@ string ParsedFile::toMsgpack(core::Context ctx, int version, const AutogenConfig
     return write.pack(ctx, *this, autogenCfg);
 }
 
+string ParsedFile::msgpackGlobalHeader(int version) {
+    return MsgpackWriter::msgpackGlobalHeader(version);
+}
+
 } // namespace sorbet::autogen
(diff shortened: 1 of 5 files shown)


jez (approved):
No easy way to write a test for this is there?

PR #7312, opened 2023-09-14
Update type-annotations.md

manny-douge:
I don't think my phrasing is correct, but running this in Sorbet.run produces the correct types for both `@x` and`@y`. Maybe we remove this section ?

Running this in Sorbet.run produces the correct types but the docs say otherwise.

### Motivation

This example is currently incorrect.

### Test plan

See included automated tests.


jez (approved):
Thanks for the suggestion. What this example is actually trying to show is that `@y` is untyped _everywhere else_, outside of the `initialize` method. I've updated the example to reflect that.

PR #7322, opened 2023-09-15
wip: Experiment with alternative syntax for generics

jez:
### Motivation

Just a fun experiment for a Friday afternoon.

Not saying we should necessarily flesh this out. But it wasn't that hard to
build either.

Right now, the biggest downside here is that you wouldn't be able to use this syntax inside a method body, e.g. `T.let(nil, T.nilable(@U))` wouldn't work.

If we do decide that the verbosity of type_parameter's is worth solving, we could either abandon this approach and choose something else that works everywhere, or we could build better error messages statically.

### Test plan

image

- [ ] @jez write tests

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
@@ -207,6 +207,10 @@ module T::Private::Methods
 
       __decl__.type_parameters = names
 
+      names.each do |name|
+        instance_variable_set("@#{name}", T::Types::TypeParameter.make(name))
+      end
+
       self
     end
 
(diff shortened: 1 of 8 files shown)


gems/sorbet-runtime/lib/types/private/methods/decl_builder.rb
```diff
@@ -5,18 +5,18 @@ module T::Private::Methods
   Declaration = Struct.new(:mod, :params, :returns, :bind, :mode, :checked, :finalized, :on_failure, :override_allow_incompatible, :type_parameters, :raw)
 
   class DeclBuilder
-    attr_reader :decl
+    attr_reader :__decl__
```
jez, line 8: We would want an error saying that type parameter names can't start with a `_` to avoid someone declaring instance variables.

PR #7323, opened 2023-09-16
Add autocorrect to use `T::Class` if possible

jez:
### Motivation

Requested by a user

### Test plan

See included automated tests.


neilparikh (approved):
Need to update an existing test, LGTM otherwise

PR #7324, opened 2023-09-16
Add `--suggest-unsafe` autocorrect to suggest `T::InexactStruct`

jez:
### Motivation

Sometimes people make struct-like things that they later wish were `T::Struct`s.
The codemod generally looks like:

Given `Args` is a struct-like superclass,

- Factor the special sauce of the struct to a module/interface like `IArgs` /
  `IArgs::ClassMethods`.
- Replace things like `class Example < Args` with
  `class Example < T::Struct; include IArgs`
- Fix the errors

Most of those errors are simply type errors that come from the fact that the
constructor is typed now.

But sometimes, people have set up complex inheritance hierarchies. The easiest
fix during the codemod is to simply opt those classes out of the newfound type
safety by marking them `T::InexactStruct` instead of `T::Struct`, and circle
back to fixing each one with more targeted fixes.

### Test plan

See included automated tests.

Note that the autocorrect does not necessarily fire in the same file that the
error is reported, and that we correctly de-dup the autocorrect.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -907,7 +907,16 @@ private:
         if (auto e = gs.beginError(sym.data(gs)->loc(), core::errors::Resolver::SubclassingNotAllowed)) {
             auto parentName = parent.show(gs);
             e.setHeader("Subclassing `{}` is not allowed", parentName);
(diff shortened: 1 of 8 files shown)


qaisjp:
Nice, that was fast, thanks!

froydnj (approved):

PR #7326, opened 2023-09-16
Add method_missing to basic_object

jez:
### Motivation

<https://github.com/ruby/ruby/blob/v3_2_0/vm_eval.c#L2587>

The method is there at runtime. This is important for calls to `super`
from within overrides of `method_missing`.

### Test plan

n/a - RBI changes


jez:

→ https://go/builds/bui_OeBgy9h0UrfZw2
→ https://go/builds/bui_OeBgaglZ3bIHFK
→ https://go/builds/bui_OeBgBEgANZ9PcJ

froydnj (approved):

PR #7328, opened 2023-09-18
Also allow referencing unexported constants from packages which can bypass visibility

aisamanra:
Allow packages specified with `--skip-package-import-visibility-check-for` to also refer to unexported constants.

### Motivation
This adds a check so that packages which bypass typical `visible_to` restrictions (for special infrastructure reasons) can also refer to constants which are not marked with `export`.

I think it would be good to change the flag name, since now the flag does more than bypass visibility, but I wanted to do that in a follow-up PR since this change is _so_ small and I worried it would get drowned out by the rote renaming.

### Test plan

See included automated tests. Added a test for this behavior.

### DIFF
--- a/test/testdata/packager/skip_check/library/__package.rb
+++ b/test/testdata/packager/skip_check/library/__package.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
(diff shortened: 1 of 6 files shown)


jez (approved):
lgtm conditional on renaming the option in the future

PR #7334, opened 2023-09-19
Add support for Ruby 3.3dev for sorbet_ruby

maruth-stripe:
### Motivation

Trying to test out Ruby 3.3.0 on pay-server to take YJIT for a spin. Couple of things to note here:

1. This passes in `--without-git` : this is because otherwise the Ruby build always tries to clone bundled gems from GitHub
2. Under `lib/rubygems` lies a new YAML parser with limited functionality. `gemrc` files must now only use YAML syntax supported bby this parser.

### Test plan

See included automated tests.

### DIFF
--- a/third_party/ruby/ruby_3_3.BUILD
+++ b/third_party/ruby/ruby_3_3.BUILD
@@ -20,7 +20,7 @@ ruby(
         "//conditions:default": [],
     }) + select({
         # Do not enable the JIT unless opted in.
-        "@com_stripe_ruby_typer//tools/config:jit_enabled": ["--enable-yjit=stats"],
+        "@com_stripe_ruby_typer//tools/config:jit_enabled": ["--enable-yjit"],
         "//conditions:default": ["--disable-jit-support"],
     }),
     copts = [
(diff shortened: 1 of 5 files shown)


third_party/ruby/ruby_3_3.BUILD
```diff
@@ -0,0 +1,72 @@
...
+        "//conditions:default": ["-O2"],
+    }),
+    cppopts = [
+        "-Wdate-time",
+        "-D_FORTIFY_SOURCE=2",
+        "-fPIC",
+    ] + select({
+        # Don't include JIT statistics unless we're building JIT support
+        "@com_stripe_ruby_typer//tools/config:jit_enabled": ["-DYJIT_STATS=1"],
+        "//conditions:default": [],
```
froydnj, line 44: OOC, do you need both `--enable-yjit=stats` and `-DYJIT_STATS=1`?  Seems a little weird, but OK.

Is having stats enabled going to mess with any performance numbers that you get?

froydnj (approved):

PR #7340, opened 2023-09-21
Allow `describe`, `before`, and `after` in `test_each`

jez:
### Motivation

Requested by a user.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Minitest.cc
+++ b/rewriter/Minitest.cc
@@ -196,11 +196,13 @@ ast::ExpressionPtr runUnderEach(core::MutableContext ctx, core::NameRef eachName
         // the send must be a call to `it` with a single argument (the test name) and a block with no arguments
         if ((send->fun == core::Names::it() && send->numPosArgs() == 1 && send->hasBlock() &&
              send->block()->args.size() == 0) ||
-            ((send->fun == core::Names::before() || send->fun == core::Names::before()) && send->numPosArgs() == 0 &&
+            ((send->fun == core::Names::before() || send->fun == core::Names::after()) && send->numPosArgs() == 0 &&
              send->hasBlock() && send->block()->args.size() == 0)) {
             core::NameRef name;
-            if (send->fun == core::Names::before() || send->fun == core::Names::after()) {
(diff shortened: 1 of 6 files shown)


rewriter/Minitest.cc
```diff
@@ -182,19 +182,30 @@ ast::ExpressionPtr getIteratee(ast::ExpressionPtr &exp) {
...
     if (auto *send = ast::cast_tree<ast::Send>(stmt)) {
         // the send must be a call to `it` with a single argument (the test name) and a block with no arguments
-        if (send->fun == core::Names::it() && send->numPosArgs() == 1 && send->hasBlock() &&
-            send->block()->args.size() == 0) {
-            // we use this for the name of our test
-            auto argString = to_s(ctx, send->getPosArg(0));
-            auto name = ctx.state.enterNameUTF8("<it '" + argString + "'>");
+        if ((send->fun == core::Names::it() && send->numPosArgs() == 1 && send->hasBlock() &&
+             send->block()->args.size() == 0) ||
+            ((send->fun == core::Names::before() || send->fun == core::Names::before()) && send->numPosArgs() == 0 &&
```
froydnj, line 199: You are checking for the same name here?  Was one of these supposed to be `beforeAngles`?

rewriter/Minitest.cc
```diff
@@ -182,19 +182,30 @@ ast::ExpressionPtr getIteratee(ast::ExpressionPtr &exp) {
...
+             send->hasBlock() && send->block()->args.size() == 0)) {
+            core::NameRef name;
+            if (send->fun == core::Names::before() || send->fun == core::Names::after()) {
+                name = send->fun == core::Names::after() ? core::Names::afterAngles() : core::Names::beforeAngles();
```
froydnj, line 203: Seems like this would be slightly clearer if you split it up into multiple `if`s, since you're duplicating the condition in the body anyway.

froydnj (approved):
WFM, modulo changes.  Does the double-checked `begin` mean that we're missing a test somewhere?

PR #7345, opened 2023-09-22
Loc of single-element InsSeq should include parens

### LINKED ISSUE #7343: Convert to singleton class method chokes if last keyword arg value is parenthesized
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20A%0A%20%20def%20example%28x%3A%29%0A%20%20%20%20%23%20%5E%20convert%20to%20singleton%20class%20method%0A%20%20end%0Aend%0A%0AA.new.example%28%0A%20%20x%3A%20%280%29%0A%29%0A%23%20A.example%28A.new%2C%20x%3A%20%280%29%0A%23%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%5E%20missing%20paren)

```ruby
# typed: true

class A
  def example(x:)
    # ^ convert to singleton class method
  end
end

A.new.example(
  x: (0)
)
# A.example(A.new, x: (0)
#                        ^ missing paren
```

#### Observed output

```
No errors! Great job.
```

When accepting the code action where indicated, the resulting edit produces a syntax error (not enough parens).

#### Expected behavior

Accepting the code action should not produce a syntax error.

- - -

jez:
### Motivation

Fixes #7343

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -197,6 +197,15 @@ ExpressionPtr desugarBegin(DesugarContext dctx, core::LocOffsets loc, parser::No
             stats.emplace_back(node2TreeImpl(dctx, std::move(stat)));
         };
         auto &last = stmts.back();
+
+        if (stmts.size() == 1) {
+            // If we're about to make an InsSeq, but the stmts are empty, MK::InsSeq will return just expr.
+            // But we want to treat `(0)` as having a loc that spans the parens too, for autocorrects.
+            // We patch that here, because the loc on an ExpressionPtr is const (maybe we should fix
+            // that, and then solve this in MK::InsSeq?) but it's easier to solve this here.
+            last->loc = loc;
+        }
+
         auto expr = node2TreeImpl(dctx, std::move(last));
         return MK::InsSeq(loc, std::move(stats), std::move(expr));
     }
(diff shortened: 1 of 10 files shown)


ast/desugar/Desugar.cc
```diff
@@ -228,13 +237,29 @@ ExpressionPtr mergeStrings(DesugarContext dctx, core::LocOffsets loc,
     }
 }
 
+// In DString, the `#{...}` are all wrapped in parser::Begin nodes.
+// Normally, we have handling for parser::Begin nodes that treats the loc of such a node to include
+// the enclosing brackets. So e.g. the loc of `(x)` would be length 3.
+// But the parser also uses Begin nodes for `#{...}`, and for those we don't want the loc to include
+// the braces (think: we don't want a `T.must` to look like `"T.must(#{x})"`, we want it to look
+// like `"#{T.must(x)}"`, etc.)
+// So we have this skipSingletonBegin to skip over these nodes before desugaring them in DString
```
froydnj, line 246: Judging from the commit history, we do have tests around this?  I didn't see any for `T.must` and the like, but it does look like we have some autocorrects applying inside interpolated strings?

froydnj (approved):

ast/desugar/Desugar.cc
jez: I think the trick is that the example I gave in the comment was hypothetical—if string interpolation required the argument to be non-nil, we could write a test for this

But as it is, I don't think we can easily write a test for when the autocorrect would apply to the whole range of the interpolated string.

PR #7347, opened 2023-09-25
Release VSCode Extension version 0.3.26

tonybruess:
### Motivation
I'd like to make the nude toggle option easily accessible to my colleagues

### Test plan
None required

### DIFF
--- a/vscode_extension/CHANGELOG.md
+++ b/vscode_extension/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Version history
+
+## 0.3.26
+- Add option to toggle the auto-complete nudge in `typed: false` files
+
 ## 0.3.25
 - Toggling highlighting of untyped code does not require a restart now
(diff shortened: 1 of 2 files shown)


jez (approved):

joshuay03:
@tonybruess @jez I've updated and tested, but it looks like the toggle isn't doing it's job 😅

I'm unsure if the issue is in https://github.com/sorbet/sorbet/pull/7221 or https://github.com/sorbet/sorbet/pull/7206.

tonybruess:
Yes I noticed some inconsistent behavior as well. For some reason it wasn't working when I was running sorbet inside of docker but it started working when I ran it outside of docker. Toggling doesn't seem reliable though and it's hard to say what's going on 🤔

joshuay03:
I don't know what change but this seems to be working consistenly now 😅👍🏽

PR #7348, opened 2023-09-25
Hide untyped argument errors if method accepts T.anything

jez:
### Motivation

There's no use in reporting this as a usage of untyped that needs to be fixed,
because the method says that it's okay to be given anything.

This is most pressing for things like

```ruby
case x
when A
  # do something
end
```

A counterargument to this feature would be that the user actually would like to
see that, because if there were better types, the condition would be dead.

But this causes two problems:

- Sorbet uses `A.===(x)` to update its knowledge of the type of variables in
  `rescue` handling.
- We already allow doing `T.let(x, A)` as a way to hide the check. But that only
  works if you're sure of the type, because if you're wrong, it will raise. At
  least with the `case`, you'll still be forced to handle the `else` case
  (usually).

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -294,7 +294,7 @@ unique_ptr<Error> matchArgType(const GlobalState &gs, TypeConstraint &constr, Lo
             //     TypeErrorDiagnostics::explainUntyped(gs, e, what, expectedType, argSym.loc, originForUninitialized);
             //     return e.build();
             // }
-        } else if (argTpe.type.isUntyped()) {
+        } else if (argTpe.type.isUntyped() && expectedType != Types::top()) {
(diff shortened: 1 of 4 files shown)


jez:

→ https://go/builds/bui_OhmjV4I1KBNYCX
→ https://go/builds/bui_Ohmke3Kd8a9uic
→ https://go/builds/bui_Ohmkdnqc2G4Uk4

jez:

→ https://go/builds/bui_Ohmz6jnPj6Disj
→ https://go/builds/bui_OhmzWQBaVLX6Jv
→ https://go/builds/bui_OhmzSdsdH7rJ3N

froydnj (approved):

PR #7354, opened 2023-09-27
Improve rescue variable pinning error

jez:
### Motivation

I think people are going to be confused by this, may as well add an error.

### Test plan

See included automated tests.


neilparikh (approved):

test/testdata/infer/exception_loop.rb
```diff
@@ -11,17 +11,38 @@ def raises_b
...
+
+  2.times do
+    begin
+      puts(e.class)
+      T.reveal_type(e) # error: `T.nilable(A)`
+      raises_b
+    rescue B => e
+      #         ^ error: Changing the type of a variable in a loop is not permitted
+    end
+  end
```
froydnj, line 28: Maybe make one or both of these a new CLI test so we have a check that the message gets triggered?  It's a helpful message and I'd hate to see it silently disappear because we changed locs for exception variables or something like that.

froydnj (approved):
Thanks for adding this context.

test/testdata/infer/exception_loop.rb
jez: oh I had meant to make a cli test, that’s for catching

jez: Oh I had meant to make a CLI test; thanks for catching

PR #7357, opened 2023-09-28
Add signatures for GC.[measure_]total_time (Ruby 3.1+)

ngroman:
These methods were added in Ruby 3.1.

### Motivation
Match https://github.com/ruby/ruby/blob/957bb7cb81995f26c671afce0ee50a5c660e540e/gc.rb#L270-L302

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/gc.rbi
+++ b/rbi/core/gc.rbi
@@ -215,26 +215,17 @@ module GC
   # consistency if RGenGC is supported.
   def self.verify_internal_consistency; end
 
-  # call-seq:
-  #    GC.measure_total_time = true/false
-  #
   # Enable to measure GC time.
-  # You can get the result with <tt>GC.stat(:time)</tt>.
+  # You can get the result with `GC.stat(:time)`.
   # Note that GC time measurement can cause some performance overhead.
   sig {params(flag: T::Boolean).void}
   def self.measure_total_time=(flag); end
 
-  # call-seq:
-  #    GC.measure_total_time -> true/false
-  #
-  # Return measure_total_time flag (default: +true+).
+  # Return measure_total_time flag (default: `true`).
   # Note that measurement can affect the application performance.
   sig {returns(T::Boolean)}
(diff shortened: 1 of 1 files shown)


rbi/core/gc.rbi
```diff
@@ -214,6 +214,30 @@ module GC
   # This method is implementation specific. Now this method checks generational
   # consistency if RGenGC is supported.
   def self.verify_internal_consistency; end
+
+  # call-seq:
+  #    GC.measure_total_time = true/false
+  #
+  # Enable to measure GC time.
+  # You can get the result with <tt>GC.stat(:time)</tt>.
+  # Note that GC time measurement can cause some performance overhead.
```
jez, line 223:
```suggestion
  # Enable to measure GC time.
  # You can get the result with `GC.stat(:time)`.
  # Note that GC time measurement can cause some performance overhead.
```

rbi/core/gc.rbi
```diff
@@ -214,6 +214,30 @@ module GC
...
+  sig {params(flag: T::Boolean).void}
+  def self.measure_total_time=(flag); end
+
+  # call-seq:
+  #    GC.measure_total_time -> true/false
+  #
+  # Return measure_total_time flag (default: +true+).
+  # Note that measurement can affect the application performance.
```
jez, line 231:
```suggestion
  # Return measure_total_time flag (default: `true`).
  # Note that measurement can affect the application performance.
```

rbi/core/gc.rbi
```diff
@@ -214,6 +214,30 @@ module GC
...
+  sig {returns(T::Boolean)}
+  def self.measure_total_time; end
+
+  # call-seq:
+  #    GC.total_time -> int
+  #
+  # Return measured GC total time in nano seconds.
```
jez, line 238:
```suggestion
  # Return measured GC total time in nano seconds.
```

jez (approved):

PR #7359, opened 2023-09-28
Add not about RBI versioning policy

jez:
### Motivation

Asked by a user

### Test plan

See included automated tests.


website/docs/rbi.md
```diff
@@ -196,6 +196,28 @@ RBI files in an `rbi/` directory. When they do this, Tapioca will automatically
...
+  library.
+
+  We will accept PRs to add an RBI definition for the new method. This means
+  that Sorbet **will not** report "method does not exist errors" for codebases
+  that are still on the old version of Ruby.
+
+- A new version of Ruby **deletes** a method that was deprecated.
+
+  We **will not** accept PRs to remove the corresponding method definition from
+  Sorbet, as that would prevent users on old Sorbet versions from calling it.
```
froydnj, line 216: This is really "users on old Ruby versions", right?

froydnj (approved):

website/docs/rbi.md
bdewater, line 216: Wouldn't it be okay to remove methods that have been removed from the minimum Ruby version Sorbet requires?

jez: Possibly, though this has never come up yet.

PR #7362, opened 2023-09-29
Add RBI definition for `Singleton.__init__`

andyw8:
### Motivation

Add RBI definition for `Singleton.__init__`:

https://github.com/ruby/ruby/blob/e05ea03553fcbbbf194dbf4cd982898480ae141b/lib/singleton.rb#L136

### Test plan

n/a

### DIFF
--- a/rbi/stdlib/singleton.rbi
+++ b/rbi/stdlib/singleton.rbi
@@ -110,6 +110,9 @@ module Singleton
     sig {returns(T.attached_class)}
     def instance; end
 
+    sig { params(klass: Class).void }
+    def self.__init__(klass); end
+
     private
 
     sig {void}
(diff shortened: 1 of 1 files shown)


jez (approved):
There's no reason to want to call this method, is there?

I assume this change is mostly motivated by not having it show up in generated RBIs for the `singleton` gem?

rbi/stdlib/singleton.rbi
```diff
@@ -110,6 +110,9 @@ module Singleton
     sig {returns(T.attached_class)}
     def instance; end
 
+    sig { params(klass: Class).void }
```
jez, line 113:
```suggestion
    sig { params(klass: T::Class[T.anything]).void }
```

andyw8:
> There's no reason to want to call this method, is there?

We actually use it in tests, as a way to reset state:

https://github.com/Shopify/ruby-lsp/blob/5e4b3f7a4956b7869003a034baae7738c6062555/test/requests/formatting_test.rb#L180

Here's some more uses from other projects (mostly in tests):
https://github.com/search?q=%2FSingleton%5C.__init__%2F+language%3ARuby&type=code&l=Ruby

PR #7365, opened 2023-09-30
Allow simple type aliases in generic classes

jez:
### Motivation

In the last week or two I've had multiple people ask me whether they could just
write something like `IntBox::Elem` in a signature outside of `IntBox` if `Elem`
is `fixed` in `IntBox`. The conversation usually goes like

- me: "just mention the fixed type in the sig--it's constant"
- them: "it's a huge type and I don't want to copy it all over the place"
- me: "define a type alias, put that in the `fixed` bound, and use the type
  alias everywhere. oh btw you have to define the type alias outside of your
  class"

We _definitely_ don't need to reject simple type aliases in generic classes. The
notes in the linked 5025 error docs don't apply to Sorbet's type aliases.

Asking "could we allow type members in type aliases" is an interesting question,
and I am not fully convinced yet that we can't, but we don't need to fully solve
that yet (can solve in the future), is not a solution to the immediate
motivation, and honestly I have a feeling that the real solution there will be
linked generic type pairs and/or type members bounded by type members.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/type_syntax/type_syntax.cc
+++ b/resolver/type_syntax/type_syntax.cc
@@ -1184,10 +1184,30 @@ optional<TypeSyntax::ResultType> getResultTypeAndBindWithSelfTypeParamsImpl(core
                     result.type = core::Types::untypedUntracked();
                 }
             } else {
-                // a type member has occurred in a context that doesn't allow them
-                if (auto e = ctx.beginError(i.loc, core::errors::Resolver::InvalidTypeDeclaration)) {
-                    auto flavor = isTypeTemplate ? "type_template"sv : "type_member"sv;
-                    e.setHeader("`{}` `{}` is not allowed in this context", flavor, sym.show(ctx));
+                switch (args.typeMember) {
+                    case TypeSyntaxArgs::TypeMember::Allowed:
+                        ENFORCE(false);
+                        break;
+                    case TypeSyntaxArgs::TypeMember::BannedInTypeAlias:
+                        if (auto e = ctx.beginError(i.loc, core::errors::Resolver::TypeAliasToTypeMember)) {
+                            const auto &owner = tm.data(ctx)->owner.asClassOrModuleRef().data(ctx);
+                            auto memTem = owner->attachedClass(ctx).exists() ? "type_template" : "type_member";
+                            e.setHeader("Defining a `{}` to a generic `{}` is not allowed", "type_alias", memTem);
+                            e.addErrorLine(tm.data(ctx)->loc(), "`{}` defined as a `{}` here",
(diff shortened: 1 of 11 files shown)


resolver/type_syntax/type_syntax.cc
```diff
@@ -1184,10 +1184,30 @@ optional<TypeSyntax::ResultType> getResultTypeAndBindWithSelfTypeParamsImpl(core
                     result.type = core::Types::untypedUntracked();
                 }
             } else {
-                // a type member has occurred in a context that doesn't allow them
-                if (auto e = ctx.beginError(i.loc, core::errors::Resolver::InvalidTypeDeclaration)) {
-                    auto flavor = isTypeTemplate ? "type_template"sv : "type_member"sv;
-                    e.setHeader("`{}` `{}` is not allowed in this context", flavor, sym.show(ctx));
+                switch (args.typeMember) {
+                    case TypeSyntaxArgs::TypeMember::Allowed:
```
neilparikh, line 1188: Could we move the body of the `if` case into here, or get rid of the `switch` and then have an `if/else if/else`? It seems weird to have this hybrid of `if` and `switch`.

resolver/type_syntax/type_syntax.cc
```diff
@@ -1184,10 +1184,30 @@ optional<TypeSyntax::ResultType> getResultTypeAndBindWithSelfTypeParamsImpl(core
...
+                                           "    can be referenced from both instance and singleton class methods\n"
+                                           "    whereas type members can only be referenced from one or the other.");
+                        }
+
+                        break;
+                    case TypeSyntaxArgs::TypeMember::BannedInTypeMember:
+                        // a type member has occurred in a context that doesn't allow them
+                        if (auto e = ctx.beginError(i.loc, core::errors::Resolver::InvalidTypeDeclaration)) {
+                            auto flavor = isTypeTemplate ? "type_template"sv : "type_member"sv;
+                            e.setHeader("`{}` `{}` is not allowed in this context", flavor, sym.show(ctx));
```
neilparikh, line 1208: Can this error message (and also the above comment) be more specific and say something like "type_member is not allowed inside another type_member"?

jez: I have punted on that in favor of simply allowing type members to be bounded by other type members. The work for that is in #7369.

resolver/type_syntax/type_syntax.cc
jez: Sure, I can do this. Originally I did not do this because I did not want to make the diff larger than it was.

neilparikh (approved):

PR #7369, opened 2023-10-01
wip: Allow type members to bound type members

jez:
### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2379,7 +2379,7 @@ class ResolveTypeMembersAndFieldsWalk {
                     ParsedSig emptySig;
                     auto allowSelfType = true;
                     auto allowRebind = false;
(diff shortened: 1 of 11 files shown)


resolver/resolver.cc
```diff
@@ -2962,7 +2962,7 @@ class ResolveTypeMembersAndFieldsWalk {
                 if (!klass.data(ctx)->typeMembers().empty()) {
                     dependencies_.emplace_back(klass);
                 }
-            } else if (symbol.isTypeAlias(ctx)) {
+            } else if (symbol.isTypeAlias(ctx) || symbol.isTypeMember()) {
```
jez, line 2965: The existing fixed point in ResultTypeMembersAndFieldsWalk (which we originally needed because of `<AttachedClass>` iirc) magically solves most of what I had thought would be the hard part of this.

PR #7371, opened 2023-10-02
serialize `LocOffsets` as `(start,len)` for better encoding

froydnj:
### Motivation

Encoding the length instead of the end means a smaller encoding for msgpack (see #7302); it should also mean a smaller encoding for the varint encoding used by Sorbet's serialization code.

And indeed, a freshly-written cache on Stripe's codebase is about 15% smaller with this change.

### Test plan

See included automated tests.


jez (approved):
wild

neilparikh (approved):

PR #7373, opened 2023-10-03
Reject all calls to abstract singleton class methods

jez:
### Motivation

Before, we only had an error for something like `ConstantLit.new` if
`ConstantLit` were an abstract class.

It turns out, we can expand this to all methods, not just `new`.

I've also taken the opportunity to move this logic into `dispatchCallSymbol`, so
that we don't need to partially reimplement method dispatch in validator.cc by
way of `findMethodTransitive`.

### Test plan

See included automated tests.

### DIFF
--- a/cfg/Instructions.cc
+++ b/cfg/Instructions.cc
@@ -119,9 +119,9 @@ string LoadSelf::showRaw(const core::GlobalState &gs, const CFG &cfg, int tabs) 
 }
 
 Send::Send(LocalRef recv, core::LocOffsets receiverLoc, core::NameRef fun, core::LocOffsets funLoc, uint16_t numPosArgs,
-           const InlinedVector<LocalRef, 2> &args, InlinedVector<core::LocOffsets, 2> &&argLocs, bool isPrivateOk,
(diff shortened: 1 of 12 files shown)


jez:

→ https://go/builds/bui_OkRJKZqlGejRgJ
→ https://go/builds/bui_OkRJvmtTmqkXY6
→ https://go/builds/bui_OkRJ6IJgGg37TI

jez:
Sorbet's support for letting abstract methods be already implemented by the inheritance chain gets in the way of this

```ruby
class AbstractParent
  extend T::Sig
  sig { void }
  def foo; puts 'foo'; end
end

class AbstractChild < AbstractParent
  extend T::Sig
  extend T::Helpers

  extend AbstractFoo

  abstract!
end

AbstractSingletonClassMethod.foo
```

This call is actually fine, because it will dispatch up to `AbstractParent.foo`, where it is not abstract. But Sorbet think the dispatch will go to `AbstractFoo#foo`, which causes problems.

Unfortunately this just makes me think more that we should not have allowed abstract methods on singleton classes. Alas.

PR #7381, opened 2023-10-05
Stress test resolver in the test runner

ilyailya:
The change adds a flag `--stress_incremental_resolver` to the test executor. 

It works in the same fashion as in the Sorbet itself. After running a test the runner will double the size of the file, and move all of the code into the second half. After that the code will be re-typchecked. If any `Loc`s will point to the first half of file an `ENFORCE` will fire an error.

To launch a test with a new flag on, you should run
```bash
bazel test  //test:test_StressResolverTests/...
```
### Motivation

Sorbet sometime crashes, because `Loc`s are not properly updated after user adds or removes a bulk of lines. The new mode for the test runner should help us to detect those cases.

### Test plan

See included automated tests.

### DIFF
--- a/main/pipeline/BUILD
+++ b/main/pipeline/BUILD
@@ -37,6 +37,7 @@ cc_library(
         "//payload/text",
         "//resolver",
         "//rewriter",
+        "//main/pipeline/definition_checker",
     ],
 )
 
@@ -76,5 +77,6 @@ cc_library(
         "//payload/text",
         "//resolver",
         "//rewriter",
+        "//main/pipeline/definition_checker",
     ],
 )
--- a/main/pipeline/definition_checker/DefinitionLinesDenylistEnforcer.cc
+++ b/main/pipeline/definition_checker/DefinitionLinesDenylistEnforcer.cc
@@ -9,6 +9,7 @@ ast::ParsedFile checkNoDefinitionsInsideProhibitedLines(core::GlobalState &gs, a
     ast::TreeWalk::apply(core::Context(gs, core::Symbols::root(), what.file), enforcer, what.tree);
     return what;
 }
+
 bool DefinitionLinesDenylistEnforcer::isAllowListed(core::Context ctx, core::SymbolRef sym) {
     return sym.name(ctx) == core::Names::staticInit() || sym.name(ctx) == core::Names::Constants::Root() ||
            sym.name(ctx) == core::Names::unresolvedAncestors();
--- a/main/pipeline/definition_checker/DefinitionLinesDenylistEnforcer.cc
+++ b/main/pipeline/definition_checker/DefinitionLinesDenylistEnforcer.cc
@@ -1,0 +1,36 @@
+#include "main/pipeline/definition_checker/DefinitionLinesDenylistEnforcer.h"
+#include "ast/treemap/treemap.h"
+
+namespace sorbet::pipeline::definition_checker {
+
+ast::ParsedFile checkNoDefinitionsInsideProhibitedLines(core::GlobalState &gs, ast::ParsedFile what,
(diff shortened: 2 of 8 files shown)


jez (changes requested):
I was hoping to only use the option as inspiration, not to actually add an option to the test runner. In particular, I would not like to effectively double the time it takes to run our tests. Do you have some time to chat about this?

main/pipeline/BUILD
```diff
@@ -30,6 +30,7 @@ cc_library(
         "//local_vars",
         "//main/autogen",
         "//main/options",
+        "//main/pipeline/definition_checker",
```
jez, line 33: "definition checker" is a pretty vague name. For example, it sounds a lot like a folder we already have, `definition_validator/`.

Do we need this to be in its own folder at all? Could the file be simply `main/pipeline/DefinitionLinesDenylistEnforcer.{h,cc}` and then we include that file directly wherever it needs to be, with no bazel changes and with no new folders?

main/pipeline/definition_checker/DefinitionLinesDenylistEnforcer.cc
```diff
@@ -0,0 +1,34 @@
...
+        {core::SymbolRef::Kind::TypeArgument, gs.typeArgumentsUsed()},
+        {core::SymbolRef::Kind::TypeMember, gs.typeMembersUsed()},
+    };
+
+    for (auto [kind, used] : symbolTypes) {
+        for (uint32_t idx = 1; idx < used; idx++) {
+            auto sym = core::SymbolRef(gs, kind, idx);
+            if (!sym.exists()) {
+                continue;
+            }
```
jez, line 21: When would this ever be true? The only symbols which don't exist are the symbols which have `id = 0` and you start this loop at 1.

main/pipeline/pipeline.cc
```diff
@@ -899,7 +853,8 @@ ast::ParsedFilesOrCancelled resolve(unique_ptr<core::GlobalState> &gs, vector<as
                     auto reresolved =
                         pipeline::incrementalResolve(*gs, move(toBeReResolved), foundHashesForFiles, opts);
                     ENFORCE(reresolved.size() == 1);
-                    f = checkNoDefinitionsInsideProhibitedLines(*gs, move(reresolved[0]), 0, prohibitedLines);
+                    sorbet::pipeline::definition_checker::checkNoDefinitionsInsideProhibitedLines(
```
jez, line 856: Despite this being fast enough to replace with trees, I don't think it's going to be fast enough that we want to do it multiple times.

Let's make a set of FileRef and pass that into checkNoDefinitions instead of a single file ref, and then move the call to checkNoDefinitions outside of this loop.

i.e., there should only be one loop over global state, not one loop over global state per file in the test.

jez: When you do that, you'll discover that you need a way to know for a given file, what lines are prohibited.

My suggestion: stop using lines for this entirely, and do it all in terms of character offets.

There's a clever realization: for every file we want to check definitions' locs, the length of the file will always be `2 * n + 1` if it started out with length `n`, because we add `n + 1` newlines to the file.

So you can compute the "midpoint" (the first offset at which it's valid to have a location) by doing `(n / 2) + 1`, and then assert that `midpoint <= loc.beginPos() && midpoint <= loc.endPos()`.

That way the only arguments that checkNoDefinitions needs is a GlobalState and an `UnorderedSet<core::FileRef>`.

jez: This will also have the effect of making the logic faster, because we won't have to compute the linebreaks of the file at all (which is slow, because it scans the whole file)—we'll do everything in terms of simple offsets.

It will also mean that we need neither of `offset2Pos` nor `position` anymore.

PR #7385, opened 2023-10-06
Show untyped blame in the editor if available

jez:
### Motivation

This helps a little bit for tracking down the untyped blames if you happen to
have a build which tracks this information.

### Test plan

Not easy to test automatically, because testing it would depend on the build
configuration.

image

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2572,7 +2572,7 @@ public:
             auto what = core::errors::Infer::errorClassForUntyped(gs, args.locs.file, receiver->type);
             if (auto e = gs.beginError(args.argLoc(0), what)) {
                 e.setHeader("Call to method `{}` on `{}`", fn.show(gs), "T.untyped");
-                TypeErrorDiagnostics::explainUntyped(gs, e, what, args.fullType, args.originForUninitialized);
+                TypeErrorDiagnostics::explainUntyped(gs, e, what, *receiver, args.originForUninitialized);
             }
 
             res.returnType = receiver->type;
--- a/core/TypeErrorDiagnostics.cc
+++ b/core/TypeErrorDiagnostics.cc
@@ -119,44 +119,25 @@ void TypeErrorDiagnostics::insertTypeArguments(const GlobalState &gs, ErrorBuild
     }
 }
 
-namespace {
-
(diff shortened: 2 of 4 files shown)


jez:

→ https://go/builds/bui_OlZXvOqadJhuRR
→ https://go/builds/bui_OlZXT6tuEz5N23
→ https://go/builds/bui_OlZXYuYhL1NS88

jez:

→ https://go/builds/bui_OlZd2O68Ekjb9T
→ https://go/builds/bui_OlZdVNGVKdVokv
→ https://go/builds/bui_OlZd1bJSweqICi

jez:

→ https://go/builds/bui_OlaSaVScS74D5c
→ https://go/builds/bui_OlaSEFxjeUZAXp
→ https://go/builds/bui_OlaSG87ouTEt0R

core/types/calls.cc
```diff
@@ -2572,7 +2572,7 @@ class Magic_callWithBlock : public IntrinsicMethod {
             auto what = core::errors::Infer::errorClassForUntyped(gs, args.locs.file, receiver->type);
             if (auto e = gs.beginError(args.argLoc(0), what)) {
                 e.setHeader("Call to method `{}` on `{}`", fn.show(gs), "T.untyped");
-                TypeErrorDiagnostics::explainUntyped(gs, e, what, args.fullType, args.originForUninitialized);
+                TypeErrorDiagnostics::explainUntyped(gs, e, what, *receiver, args.originForUninitialized);
```
jez, line 2575: This was a bug revealed by an ENFORCE that triggered after making this change.

We are now calling `untypedBlame` on the thing passed to `explainUntyped`, and `TypePtr::untypedBlame` asserts that the thing it's called on `hasUntyped()` as a sanity check.

This revealed the bug at hand: we were reporting this error if the receiver is untyped, but then explaining it as if it blamed to the `args.fullType`. But in this intrinsic, the `args.fullType` is always `T.class_of(<Magic>)`.

This didn't _really_ cause problems. It would have just made for a confusing error message if it happened, but the `explainUntyped` helper doesn't do anything like attribute where the untyped value came from or anything like that.

PR #7389, opened 2023-10-10
Change `Class.new`'s return type to `T::Class[T.untyped]`

st0012:
### Motivation

The current `T::Class[Object]` means in some cases incorrect typing error would be raised when using `Class.new` to create a class with a custom constructor, like:

```rb
class A
  r = Class.new do
    def initialize(a)
    end
  end

  r.new(123) # => reporting wrong number of arguments
end
```

This PR fixes the issue by changing the return type to `T::Class[T.untyped]`.


st0012:
@jez I've updated tests and CI is green now.

jez (approved):

PR #7392, opened 2023-10-11
Misuse of `static` no longer falsifies untyped blame

jez:
### Motivation

The first time these `arrayOfUntyped` / `rangeOfUntyped` /
`hashOfUntyped` methods were called, the blame was being cached in a
`static` variable and then reused throughout the lifetime of the
program.

In Stripe's codebase, we were seeing a lot of untyped blame to
`<build-hash>` and none to `<shapeUnderlying>`, which I thought was
strange because a program as simple as:

```ruby
res = {foo: 0}
foo = res[:foo].even?
```

should be enough to show an untyped blame to `<shapeUnderlying>`... and
on small examples it was!

I only realized that something was up because I tried adding some extra
error sections to the "usage of untyped" error, and I noticed that the
blame symbol was different in my editor vs at the command line on small
examples. This was enough to let me run the code under a debugger which
eventually led me to these lines.

### Test plan

This is hard to write an automated test for, because we don't have many (any?)
tests of our untyped blame setup (it involves a custom build of Sorbet).

I verified manually that this fixes the problem I was seeing locally. I plan to
deploy this to Stripe's codebase and test in prod to see whether we start seeing
blames for `<shapeUnderlying>` in our nightly job.

### DIFF
--- a/core/types/types.cc
+++ b/core/types/types.cc
@@ -77,15 +77,27 @@ TypePtr Types::Float() {
 }
 
 TypePtr Types::arrayOfUntyped(sorbet::core::SymbolRef blame) {
-    static vector<TypePtr> targs{Types::untyped(blame)};
-    static auto res = make_type<AppliedType>(Symbols::Array(), move(targs));
-    return res;
+    if constexpr (!sorbet::track_untyped_blame_mode && !sorbet::debug_mode) {
+        static vector<TypePtr> targs{Types::untypedUntracked()};
+        static auto res = make_type<AppliedType>(Symbols::Array(), move(targs));
+        return res;
+    } else {
+        vector<TypePtr> targs{Types::untyped(blame)};
+        auto res = make_type<AppliedType>(Symbols::Array(), move(targs));
+        return res;
+    }
 }
 
 TypePtr Types::rangeOfUntyped(sorbet::core::SymbolRef blame) {
-    static vector<TypePtr> targs{Types::untyped(blame)};
-    static auto res = make_type<AppliedType>(Symbols::Range(), move(targs));
-    return res;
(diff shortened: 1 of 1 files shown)


core/types/types.cc
```diff
@@ -77,15 +77,27 @@ TypePtr Types::Float() {
 }
 
 TypePtr Types::arrayOfUntyped(sorbet::core::SymbolRef blame) {
-    static vector<TypePtr> targs{Types::untyped(blame)};
-    static auto res = make_type<AppliedType>(Symbols::Array(), move(targs));
-    return res;
+    if constexpr (!sorbet::track_untyped_blame_mode && !sorbet::debug_mode) {
```
froydnj, line 80: If I'm reading this correctly, this means that all debug mode builds will used uncached types for blamed untyped calls, whereas I would expect we would only want the no-caching behavior in `track_untyped_blame_mode`?  I suppose the rationale here is that you want to be able to track blames in the debugger if that would come up, and it's inconvenient to build an entirely separate blame-untyped build for such purposes.

froydnj (approved):
I'm not sure I like the condition here, but I guess it's at least consistent with earlier behavior in `types.cc`.

core/types/types.cc
jez: Yeah it's a holdover from how things used to work--this is the condition we use for all our blame tracking now.

Before we added a custom bazel config for track_untyped_blame_mode, the only flag we had was debug_mode. We added track_untyped_blame_mode so that you could track blames without also having to turn on all the ENFORCEs, so it would run fast on Stripe's codebase.

We thought about whether we should flip things over exclusively to branching on track_untyped_blame_mode, but we decided it would be nice to get some semblance of "it doesn't crash" coverage by having the typed blame tracking also run in CI.

PR #7398, opened 2023-10-12
Update the location for `sealed_subclasses` after fast path edit

ilyailya:
After the fast path edit the location for the `sealed_subclasses` method was not properly updated, and caused crashes. The change adds a logic the namer to fix that. 

There is a method `enterMethodSymbol`, which is called inside of a namer. It has a shortcut if a symbol for a method is already defined, it wouldn't do anything. There is a corner case of deleting or adding a bunch of lines in a file (see the test) which will not define any new symbols, but should update the locations. The shortcut wasn't checking if the location have been updated. Because of that the symbol for a method might point to a part of a file which doesn't exist anymore, and Sorbet will crash. The change fixes that bug.

### Motivation

Less crashes!

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1327,6 +1327,11 @@ MethodRef GlobalState::enterMethodSymbol(Loc loc, ClassOrModuleRef owner, NameRe
     if (store.exists()) {
         ENFORCE(store.isMethod(), "existing symbol is not a method");
         counterInc("symbols.hit");
+        if (!store.locs(*this).empty() && absl::c_count(store.locs(*this), loc) == 0) {
+            ENFORCE(!symbolTableFrozen);
+            store.asMethodRef().data(*this)->addLoc(*this, loc);
+            wasModified_ = true;
+        }
         return store.asMethodRef();
     }
 
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -1327,11 +1327,6 @@ MethodRef GlobalState::enterMethodSymbol(Loc loc, ClassOrModuleRef owner, NameRe
     if (store.exists()) {
         ENFORCE(store.isMethod(), "existing symbol is not a method");
         counterInc("symbols.hit");
-        if (!store.locs(*this).empty() && absl::c_count(store.locs(*this), loc) == 0) {
-            ENFORCE(!symbolTableFrozen);
-            store.asMethodRef().data(*this)->addLoc(*this, loc);
-            wasModified_ = true;
-        }
         return store.asMethodRef();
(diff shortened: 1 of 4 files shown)


core/GlobalState.cc
```diff
@@ -1327,6 +1327,11 @@ MethodRef GlobalState::enterMethodSymbol(Loc loc, ClassOrModuleRef owner, NameRe
     if (store.exists()) {
         ENFORCE(store.isMethod(), "existing symbol is not a method");
         counterInc("symbols.hit");
+        if (!store.locs(*this).empty() && absl::c_count(store.locs(*this), loc) == 0) {
```
jez, line 1330: Can you explain what your goal with this condition is?

My guess would have been that we would not have needed any condition here.

core/GlobalState.cc
```diff
@@ -1327,6 +1327,11 @@ MethodRef GlobalState::enterMethodSymbol(Loc loc, ClassOrModuleRef owner, NameRe
...
+            ENFORCE(!symbolTableFrozen);
+            store.asMethodRef().data(*this)->addLoc(*this, loc);
```
jez, line 1332: I'm not sure that we want to call `addLoc` here. I'm a little worried about what effect this will actually have, as it will affect literally every call to `enterMethodSymbol`, and I think that could potentially cause some non-obvious effects.

I'd prefer being more cautious and doing something at the call site, like what we did when we fixed a related bug with singleton class locs in this PR:

#6447

namer/namer.cc
```diff
@@ -1261,8 +1261,11 @@ class SymbolDefiner {
             symbolData->flags.isSealed = true;
 
             auto classOfKlass = symbolData->singletonClass(ctx);
-            auto sealedSubclasses =
-                ctx.state.enterMethodSymbol(ctx.locAt(mod.loc), classOfKlass, core::Names::sealedSubclasses());
+            auto loc = ctx.locAt(mod.loc);
+            auto sealedSubclasses = ctx.state.enterMethodSymbol(loc, classOfKlass, core::Names::sealedSubclasses());
+            if (absl::c_count(sealedSubclasses.data(ctx)->locs(), loc) == 0) {
+                sealedSubclasses.data(ctx)->addLoc(ctx, loc);
+            }
```
jez, line 1268: What is the goal with this `c_count` condition?

jez: Asking because I'm _pretty_ sure the `c_count` condition is essentially a no-op based on how the `addLoc` method is already implemented.

jez (approved):

PR #7407, opened 2023-10-17
Update a location for the `<AttachedClass>` in the namer

ilyailya:
Similar to https://github.com/sorbet/sorbet/pull/7398

In the fast path edit Sorbet doesn't update a location for the type argument `<AttachedClass>`. As a result after a large fast path edit (like deleting a long comment) the `<AttachedClass>`'s location may point to the wrong part of file
The change fixes it.

### Motivation

Less crashes

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1215,6 +1215,11 @@ private:
 
                 auto singletonClass = symbol.data(ctx)->singletonClass(ctx); // force singleton class into existence
                 singletonClass.data(ctx)->addLoc(ctx, ctx.locAt(klass.declLoc));
+                for (const auto &typeMember : singletonClass.data(ctx)->typeMembers()) {
+                    if (typeMember.data(ctx)->name == core::Names::Constants::AttachedClass()) {
+                        typeMember.data(ctx)->addLoc(ctx, ctx.locAt(klass.declLoc));
+                    }
+                }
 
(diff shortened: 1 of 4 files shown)


namer/namer.cc
```diff
@@ -1215,6 +1215,11 @@ class SymbolDefiner {
 
                 auto singletonClass = symbol.data(ctx)->singletonClass(ctx); // force singleton class into existence
                 singletonClass.data(ctx)->addLoc(ctx, ctx.locAt(klass.declLoc));
+                for (const auto &typeMember : singletonClass.data(ctx)->typeMembers()) {
+                    if (typeMember.data(ctx)->name == core::Names::Constants::AttachedClass()) {
+                        typeMember.data(ctx)->addLoc(ctx, ctx.locAt(klass.declLoc));
+                    }
+                }
```
jez, line 1222:
```suggestion
                auto attachedClassTM =
                    singletonClass.data(ctx)->findMember(ctx, core::Names::Constants::AttachedClass());
                if (attachedClassTM.exists() && attachedClassTM.isTypeMember()) {
                    attachedClassTM.asTypeMemberRef().data(ctx)->addLoc(ctx, ctx.locAt(klass.declLoc));
                }
```

No need to loop when we can find.

jez (approved):

PR #7408, opened 2023-10-17
Deduplicate autogen constant references

jez:
### Test plan

See included automated tests.


test/testdata/autogen/duplicate_refs.rb.autogen.exp
```diff
@@ -60,55 +60,20 @@ requires: []
...
- loc=test/testdata/autogen/duplicate_refs.rb:5
- is_defining_ref=0
-[ref id=9]
- scope=[A]
- name=[Integer]
- nesting=[[A]]
- resolved=[Integer]
- loc=test/testdata/autogen/duplicate_refs.rb:5
- is_defining_ref=0
-[ref id=10]
```
jez: check the diff

main/autogen/autogen.cc
```diff
@@ -224,6 +227,12 @@ class AutogenWalk {
             return;
         }
 
+        auto entry = make_pair(tree.loc(), original.symbol);
+        if (seenRefsByLoc.contains(entry)) {
+            return;
+        }
+        seenRefsByLoc.emplace(move(entry));
```
froydnj, line 234: WDYT about also skipping zero-length locs?  I guess that wouldn't really help with the problem being addressed here, because we probably keep the full-width locs around so that error messages from the duplicated constants are reasonable.

This would probably be epsilon faster if you used `insert`:

https://github.com/abseil/abseil-cpp/blob/2a18ba753bca64e79f6d9857a9f64e638b32c371/absl/container/flat_hash_set.h#L250-L261

so that you can do the lookup and (possible) insertion in a single go.

froydnj (approved):
Looks like you have some test failures to fix up, but those look pretty trivial.

jez:

→ https://go/builds/bui_PQJUkdRhP9HJUJ
→ https://go/builds/bui_PQJUma8UigS4wv
→ https://go/builds/bui_PQJUKp4pXBU6at

core/LocOffsets.h
```diff
@@ -51,6 +51,10 @@ struct LocOffsets {
 };
 CheckSize(LocOffsets, 8, 4);
 
+template <typename H> H AbslHashValue(H h, const LocOffsets &m) {
```
neilparikh, line 56: Is this function used when inserting into `UnorderedSet`?

neilparikh (approved):

core/LocOffsets.h
jez: yep

main/autogen/autogen.cc
jez: > WDYT about also skipping zero-length locs? I guess that wouldn't really help with the problem being addressed here, because we probably keep the full-width locs around so that error messages from the duplicated constants are reasonable.

going to punt on this because I think we want this in, and I think that the current solution is "good enough." Happy to revisit if the difference starts to matter in practice.

PR #7411, opened 2023-10-18
Update location of the static field on fast path

ilyailya:
The PR is a continuation of an effort to eliminate all crashes caused by faulty locations (https://github.com/sorbet/sorbet/pull/7398, https://github.com/sorbet/sorbet/pull/7407, https://github.com/sorbet/sorbet/pull/7381)

This change updates the location of the static field after the fast path update

### Motivation

Less crashes

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1353,7 +1353,6 @@ private:
             }
         }
         sym = ctx.state.enterStaticFieldSymbol(ctx.locAt(staticField.lhsLoc), scope, name);
-        sym.data(ctx)->addLoc(ctx, ctx.locAt(staticField.lhsLoc));
         // Reset resultType to nullptr for idempotency on the fast path--it will always be
         // re-entered in resolver.
         sym.data(ctx)->resultType = nullptr;
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1353,6 +1353,7 @@ private:
             }
         }
         sym = ctx.state.enterStaticFieldSymbol(ctx.locAt(staticField.lhsLoc), scope, name);
+        sym.data(ctx)->addLoc(ctx, ctx.locAt(staticField.lhsLoc));
         // Reset resultType to nullptr for idempotency on the fast path--it will always be
         // re-entered in resolver.
         sym.data(ctx)->resultType = nullptr;
--- a/test/testdata/lsp/fast_path/static_field_loc.rb
+++ b/test/testdata/lsp/fast_path/static_field_loc.rb
@@ -1,0 +1,27 @@
+# typed: true
+
+# This is a rather large comment at the start of the file so that when we
+# delete it on the fast path, it will make it very likely that we have an array
+# access that's out of bounds when reporting errors.
+# This is a rather large comment at the start of the file so that when we
+# delete it on the fast path, it will make it very likely that we have an array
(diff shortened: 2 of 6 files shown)


test/testdata/lsp/fast_path/static_field_loc.rb
```diff
@@ -0,0 +1,27 @@
...
+# delete it on the fast path, it will make it very likely that we have an array
+# access that's out of bounds when reporting errors.
+# This is a rather large comment at the start of the file so that when we
+# delete it on the fast path, it will make it very likely that we have an array
+# access that's out of bounds when reporting errors.
+#
+class B
+  extend T::Generic
+  R = 1
+  R = type_member
```
jez, line 25: This is not a static field, this is a type_member.

I believe the crash reproduces even if the test does not have `R` defined simultaneously as a static field and type member.

Can you simplify this test to delete `type_member`, and deal with type members either in a follow up commit on this PR, or in a future PR?

namer/namer.cc
```diff
@@ -1353,6 +1353,7 @@ class SymbolDefiner {
             }
         }
         sym = ctx.state.enterStaticFieldSymbol(ctx.locAt(staticField.lhsLoc), scope, name);
```
jez, line 1355: Can we grep for calls to `enterStaticFieldSymbol` and see if it suggests other code paths that we need to write tests for?

For example, I think that

```ruby
@@x = 1
```

will also suffer from this crash but exercise a different code path.

Another one I think suffers the same fate is

```
X = type_template
```

namer/namer.cc
```diff
@@ -1353,6 +1353,7 @@ class SymbolDefiner {
...
+        sym.data(ctx)->addLoc(ctx, ctx.locAt(staticField.lhsLoc));
```
jez, line 1356: Unlike with `singletonClass` and `enterMethodSymbol`, I think that we likely have a much smaller scope for `enterStaticFieldSymbol`, and we could consider doing something like we did with `staticInitForFile` and `staticInitForClass`, where if the symbol exists, we call `addLoc`.

jez (commented):
Looks great so far; we can make it a little better while we're here.

namer/namer.cc
ilyailya: Sounds good to me, let's try that and see what fails

namer/namer.cc
ilyailya: This test with a comment manipulation passes on master
```ruby
class B
extend T::Sig
  @@x = 1

  sig {void}
  def self.foo
    T.reveal_type(@@x) # error: Revealed type: `Integer`
  end
end
```

ilyailya: And this test fails even with a change from this PR
```ruby
class B
  extend T::Generic
  X = type_template
end

T.reveal_type(B::X) # error: Revealed type: `Runtime object representing type: T.class_of(B)::X`
```

It would require a separate fix

ilyailya: The test case with `@@x` updates the location correctly, because it was already handled:
https://github.com/sorbet/sorbet/blob/24163761b87fc5ac0a40b31b567384c2a8688d1a/namer/namer.cc#L1317-L1319

jez (approved):

PR #7410, opened 2023-10-17
Add docs for methods with overloaded signatures

jez:
### Motivation

Fixes #7410

### Test plan

See included automated tests.

PR #7412, opened 2023-10-18
Allow overloaded signatures if method is defined exclusively in an RBI

### LINKED ISSUE #6803: Limited support for overloaded signatures, only with or without a block?
### Problem

I'm aware of https://sorbet.org/docs/error-reference#5040, I looked at how Sorbet currently handles overloads ([here if I'm not mistaken](https://github.com/sorbet/sorbet/blob/36ad4d8679af03eb453635bb5742b113b68215a7/core/types/calls.cc#L329-L330)) and I can see why it's not something the team would want to commit to in it's current form. This is about a common Ruby idiom that feels different than the example given in the docs and might be simpler to allow, so I figured it was worth opening an issue and see what the team thinks.

I've recently run into a few cases where it would've been nice to have two signatures, one for where a block is given and one where that is not the case.

[**Active Record find**](https://api.rubyonrails.org/v7.0/classes/ActiveRecord/FinderMethods.html#method-i-find)
Commonly used as `Post.find(1)` but also accepts an array of IDs, which is why Tapioca generates `T.untyped` return values by default. Despite [documentation](https://github.com/Shopify/tapioca/blob/5a913bc44e00c97fb5f0e81a27c819a899f4a1b3/manual/compiler_activerecordrelations.md?plain=1#L53-L61) this confuses newcomers with some regularity as evidenced by the Sorbet slack questions.

[**ActiveRecord::Batches**](https://api.rubyonrails.org/v7.0/classes/ActiveRecord/Batches.html)
Methods `find_each`, `find_in_batches` and `each_batch` look like [`Array#each`](https://github.com/sorbet/sorbet/blob/36ad4d8679af03eb453635bb5742b113b68215a7/rbi/core/array.rbi#L1165-L1189), but without overloaded signatures need typing with nilable.

```ruby
sig { params(start: T.untyped, finish: T.untyped, batch_size: Integer, error_on_ignore: T.untyped, order: Symbol, block: T.nilable(T.proc.params(object: ::Post).void)).returns(T.nilable(T::Enumerator[::Post])) }
def find_each(start: nil, finish: nil, batch_size: 1000, error_on_ignore: nil, order: :asc, &block); end
```

Could look like:
```ruby
sig { params(start: T.untyped, finish: T.untyped, batch_size: Integer, error_on_ignore: T.untyped, order: Symbol).returns(T::Enumerator[::Post]) }
sig { params(start: T.untyped, finish: T.untyped, batch_size: Integer, error_on_ignore: T.untyped, order: Symbol, block: T.proc.params(object: ::Post).void).void }
def find_each(start: nil, finish: nil, batch_size: 1000, error_on_ignore: nil, order: :asc, &block); end
```

Signatures for this caused some confusion over at https://github.com/Shopify/tapioca/pull/1417

**ActiveSupport delegated methods**
In a few places Rails has a class that includes Enumerable, but delegates `each` and some other objects to an instance variable that contains the actual data in an array. A couple of examples:
- [ActiveModel::Errors](https://github.com/rails/rails/blob/7c70791470fc517deb7c640bead9f1b47efb5539/activemodel/lib/active_model/errors.rb#L67-L78)
- ActiveRecord::Relation includes [ActiveRecord::Delegation](https://github.com/rails/rails/blob/7c70791470fc517deb7c640bead9f1b47efb5539/activerecord/lib/active_record/relation/delegation.rb#L88-L91) 

This has caused some confusion on our team in cases like these:
```ruby
Post.where(author: "Bart").each do |post|
  # post is untyped
end
```
Using `map` gets typing for free (but is doing needless work) and sometimes using `T.let` is forgotten. With upstream refactoring it could be typed like stdlib `Array#each`.

### Proposed solution

Allow overloaded signatures for `# typed: strict` and up, instead of `# typed: __STDLIB_INTERNAL`, under the following conditions:
- the arguments must be exactly the same, except
- the block argument is specified, or not (so more than 2 sigs would be an error)
- the return types can be different

jez:
### Motivation

Will allow better support for gems.

Fixes #6803

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_OqQ5q3ObIRlHTJ
→ https://go/builds/bui_OqQ5FvE3eEAAxB
→ https://go/builds/bui_OqQ5z2CxvKarIT

jez:
* master
  * **PR #7410** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/7410" target="_blank">Graphite</a> 
    * **PR #7412** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/7412" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/7412?utm_source=stack-comment).

infer/inference.cc
```diff
@@ -12,6 +12,42 @@
...
+                               "    To silence this error, mark this file `{}`\n"
+                               "    Read the error doc for how to avoid using overloaded methods in the first place.",
+                               "# typed: false");
+            }
+        }
+
+        return false;
+    }
+
+    if (methodData->flags.isAbstract && ctx.file.data(ctx).compiledLevel != core::CompiledLevel::True) {
```
ilyailya, line 44: I saw we had this condition in other places, so this is how ended up in this method. 
For my personal understanding, why does Sorbet refuses to run inference for abstract methods in non-compiled files?

ilyailya (approved):
Change looks good!
I left one comment with a question for a better understanding

infer/inference.cc
jez: There’s no need to build a CFG for an abstract method and type check it, because it’s always empty.

But for the compiler, we need a CFG so that the compiler can emit a native code method for it, which calls super if defined or raise an exception otherwise.

PR #7413, opened 2023-10-19
Fix Psych.load_stream sig

thomasmarshall:
This commit adds an optional block to the `Psych.load_stream` sig to better match the [current implementation](https://github.com/ruby/psych/blob/51cc86ff3f7cf98bc48d7eb1adf1e3c34b4ac475/lib/psych.rb#L627-L631).

### Motivation

This became a type error in our codebase recently when we updated sorbet.

### Test plan

N/A.

### DIFF
--- a/rbi/stdlib/psych.rbi
+++ b/rbi/stdlib/psych.rbi
@@ -613,10 +613,11 @@ module Psych
       filename: T.nilable(String),
       fallback: T.untyped,
       kwargs: T.untyped,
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_OqiI39SVtDzbyb
→ https://go/builds/bui_OqiIPzrTgc4918
→ https://go/builds/bui_OqiIYkMURaiDEJ

PR #7415, opened 2023-10-19
Fix some Minitest rewriter locs

jez:
### Motivation

Reported by a user

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Minitest.cc
+++ b/rewriter/Minitest.cc
@@ -108,8 +108,21 @@ public:
 };
 
 ast::ExpressionPtr addSigVoid(ast::ExpressionPtr expr) {
-    return ast::MK::InsSeq1(expr.loc(), ast::MK::SigVoid(expr.loc(), {}), std::move(expr));
+    core::LocOffsets declLoc;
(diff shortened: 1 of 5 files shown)


rewriter/Minitest.cc
```diff
@@ -108,8 +108,21 @@ class ConstantMover {
 };
 
 ast::ExpressionPtr addSigVoid(ast::ExpressionPtr expr) {
-    return ast::MK::InsSeq1(expr.loc(), ast::MK::SigVoid(expr.loc(), {}), std::move(expr));
+    core::LocOffsets declLoc;
+    if (auto *mdef = ast::cast_tree<ast::MethodDef>(expr)) {
+        declLoc = mdef->declLoc;
+    } else if (auto *cdef = ast::cast_tree<ast::MethodDef>(expr)) {
```
ilyailya, line 114: Did you mean?
```suggestion
    } else if (auto *cdef = ast::cast_tree<ast::ClassDef>(expr)) {
```

Right now second condition is identical to the first one and would never be executed

ilyailya (changes requested):

PR #7414, opened 2023-10-19
Update location of type member on a fast path

ilyailya:
The PR is a continuation of an effort to eliminate all crashes caused by faulty locations (https://github.com/sorbet/sorbet/pull/7398, https://github.com/sorbet/sorbet/pull/7407, https://github.com/sorbet/sorbet/pull/7381, https://github.com/sorbet/sorbet/pull/7411)

This change updates the location of `type_member` and `type_argument` after the fast path update

### Motivation

Less crashes

### Test plan

See included automated tests.

PR #7418, opened 2023-10-23
Update location for type arguments on a fast path

ilyailya:
The PR is a continuation of an effort to eliminate all crashes caused by faulty locations (https://github.com/sorbet/sorbet/pull/7398, https://github.com/sorbet/sorbet/pull/7407, https://github.com/sorbet/sorbet/pull/7381, https://github.com/sorbet/sorbet/pull/7411, https://github.com/sorbet/sorbet/pull/7414)

The change in this pull request updates the location for the type_arguments. 

I have a hard time writing a test for this change. Firstly I've worked on this change on a branch with a addition to our test harness, which stress tests location correctness https://github.com/sorbet/sorbet/pull/7381. On that branch [all_bug.rb test](https://github.com/sorbet/sorbet/blob/master/test/testdata/infer/all_bug.rb) fails without the change from this PR.

I've tried localizing the error with following tests:

```ruby
# typed: true

# Here goes a large comment, which will be deleted on a fast path

extend T::Sig
extend T::Generic

sig do
  type_parameters(:U)
  .params(
    x: T.type_parameter(:U),
  )
end
def foo(x)
  42
end
```

```ruby
# typed: true

# Here goes a large comment, which will be deleted on a fast path

extend T::Sig
extend T::Generic

sig do
  type_parameters(:U)
  .params(
    x: T.type_parameter(:U),
  )
  .void
end
def foo(x)
  T.reveal_type(x)
end
```

```ruby
# typed: true

# Here goes a large comment, which will be deleted on a fast path

extend T::Sig
extend T::Generic

sig do
  type_parameters(:U)
  .params(
    x: T.type_parameter(:U),
  )
  .void
end
def foo(x)
  x
# ^ hover: ...
end
```
```ruby
# typed: true

# Here goes a large comment, which will be deleted on a fast path

extend T::Sig
extend T::Generic

sig do
  type_parameters(:U)
  # ^ hover: ...
  .params(
    x: T.type_parameter(:U),
  )
  .void
end
def foo(x)
  x

end
```

None of those tests crash on master unfortunately. I'd appreciate any ideas about how I could test this change

### Motivation

Less crashes

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -3560,6 +3560,7 @@ private:
                 if (typeSpec.type) {
                     auto name = ctx.state.freshNameUnique(core::UniqueNameKind::TypeVarName, typeSpec.name, 1);
                     auto sym = ctx.state.enterTypeArgument(typeSpec.loc, method, name, core::Variance::CoVariant);
+                    sym.data(ctx)->addLoc(ctx, typeSpec.loc);
                     auto asTypeVar = core::cast_type<core::TypeVar>(typeSpec.type);
                     ENFORCE(asTypeVar != nullptr);
                     asTypeVar->sym = sym;
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -3560,7 +3560,6 @@ private:
                 if (typeSpec.type) {
                     auto name = ctx.state.freshNameUnique(core::UniqueNameKind::TypeVarName, typeSpec.name, 1);
                     auto sym = ctx.state.enterTypeArgument(typeSpec.loc, method, name, core::Variance::CoVariant);
-                    sym.data(ctx)->addLoc(ctx, typeSpec.loc);
                     auto asTypeVar = core::cast_type<core::TypeVar>(typeSpec.type);
                     ENFORCE(asTypeVar != nullptr);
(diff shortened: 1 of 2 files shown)


resolver/resolver.cc
```diff
@@ -3560,6 +3560,7 @@ class ResolveSignaturesWalk {
                 if (typeSpec.type) {
                     auto name = ctx.state.freshNameUnique(core::UniqueNameKind::TypeVarName, typeSpec.name, 1);
                     auto sym = ctx.state.enterTypeArgument(typeSpec.loc, method, name, core::Variance::CoVariant);
+                    sym.data(ctx)->addLoc(ctx, typeSpec.loc);
```
jez, line 3563: This is likely another case where we can get away with pushing the `addLoc` into the call to `enterTypeArgument`.

jez:
> I've tried localizing the error with following tests

I'm pretty sure we don't have any errors which reference the location of a type argument. A little surprising, but I guess that's the way it is.

jez (approved):
This looks good to me but I think I'd like to see us move the logic into `enterTypeArgument`.

Issue #7417, opened 2023-10-23
Bad error loc on accidental switch-case fallthrough


maruth-stripe:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28x%3A%20Integer%29.returns%28String%29%20%7D%0Adef%20foo%28x%29%0A%20%20case%20x%0A%20%20when%201%0A%20%20when%202%0A%20%20%20%20%22a%22%0A%20%20end%0Aend)

```ruby
# typed: true
extend T::Sig

sig { params(x: Integer).returns(String) }
def foo(x)
  case x
  when 1
  when 2
    "a"
  end
end
```

#### Observed output

```
editor.rb:11: Expected String but found T.nilable(String) for method result type https://srb.help/7005
    11 |end
        ^^^
  Expected String for result type of method foo:
    editor.rb:5:
     5 |def foo(x)
        ^^^^^^^^^^
  Got T.nilable(String) originating from:
    editor.rb:9:
     9 |    "a"
            ^^^
    editor.rb:5: Possibly uninitialized (NilClass) in:
     5 |def foo(x)
        ^^^^^^^^^^
Errors: 1
```

#### Expected behavior
Sorbet observes the return type of the method as being `T.nilable(String)` instead of `String` because the `when` in the `case` statement do not "fall through" as in other languages. In contrast, an empty `when` corresponds to the value being `nil`. However, the error message does not point to where the inferred `nil` is coming from.

- - -

PR #7420, opened 2023-10-25
Better loc for error message when `when` body is empty for `case` block

maruth-stripe:
### Motivation

https://github.com/sorbet/sorbet/issues/7417

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -2199,8 +2199,13 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what) 
                             cond = MK::If(loc, std::move(test), std::move(true_), std::move(cond));
                         }
                     }
-                    res =
-                        MK::If(when->loc, std::move(cond), node2TreeImpl(dctx, std::move(when->body)), std::move(res));
+
+                    auto whenBody = MK::Nil(when->loc);
+                    if (when->body != nullptr) {
+                        whenBody = node2TreeImpl(dctx, std::move(when->body));
+                    }
+
+                    res = MK::If(when->loc, std::move(cond), std::move(whenBody), std::move(res));
                 }
                 if (assign != nullptr) {
                     res = MK::InsSeq1(loc, std::move(assign), std::move(res));
--- a/test/cli/case-blank-when-body/test.out
+++ b/test/cli/case-blank-when-body/test.out
@@ -1,0 +1,18 @@
+test/cli/case-blank-when-body/blank.rb:14: Expected `String` but found `T.nilable(String)` for method result type https://srb.help/7005
+    14 |end
+        ^^^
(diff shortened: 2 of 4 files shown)


jez (commented):
i think we might want to chat about this one

ast/desugar/Desugar.cc
```diff
@@ -2199,8 +2199,13 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
                             cond = MK::If(loc, std::move(test), std::move(true_), std::move(cond));
                         }
                     }
-                    res =
-                        MK::If(when->loc, std::move(cond), node2TreeImpl(dctx, std::move(when->body)), std::move(res));
+
+                    auto whenBody = MK::Nil(when->loc);
```
jez, line 2203: We should only need one allocation. This implementation sometimes allocates twice.

jez (commented):
So my overall feedback is that I wish we could come up with a more general solution. This solution of "check if the thing we're about recursively desugar (`when->body`) is `nullptr`" is kind of like unfolding the base case of a recursive function into the recursive case.

As in, realistically this problem will show up *any* time that `node2TreeImpl` is called on `nullptr`. That could happen in a bunch of places, but probably the most relevant are the places where the `nullptr` is also part of a conditional branching construct. `case/when` is not the only conditional branching construct. There's also `if`, `unless`, `while`, `until`, `case <nothing>; when`, and `case/in`.

That this PR only fixes the `case/when` problem feels like a partial solution.

And then another thing: we only allocate one `EmptyTree` globally, but each `nil` allocated here is another 16 bytes. I'd like to understand what impact that has on memory usage for whether we should land this change.

test/cli/case-blank-when-body/test.out
```diff
@@ -0,0 +1,18 @@
...
+    14 |end
+        ^^^
+  Expected `String` for result type of method `foo`:
+    test/cli/case-blank-when-body/blank.rb:6:
+     6 |def foo(x)
+        ^^^^^^^^^^
+  Got `T.nilable(String)` originating from:
+    test/cli/case-blank-when-body/blank.rb:8:
+     8 |  when 1
+     9 |  when 2
```
jez, line 11: This is showing the `"when 1\n"` as the location where the `nil` comes from.

That's a bit weird—that means for example that if you hovered over the `when` keyword, Sorbet would show you a revealed type of `nil`. It also means that certain autocorrects might attempt to use wrap `"when 1\n"` in a `T.must` or `T.unsafe`, which would be a syntax error.

I think that we should instead set the loc for the `MK::Nil` to `when->loc.copyEndWithZeroLength()`.

Doing this exposes another bug, which is that the location includes the `"\n"`. We should probably change the loc for the `parser::When` node that we create in Builder.cc to not include the loc of the `then` token if it's actually backed by a `tNL` token instead of a `kTHEN` token.

PR #7424, opened 2023-10-25
Update location for type member's static field on fast path

ilyailya:
The PR is a continuation of an effort to eliminate all crashes caused by faulty locations (https://github.com/sorbet/sorbet/pull/7398, https://github.com/sorbet/sorbet/pull/7407, https://github.com/sorbet/sorbet/pull/7381, https://github.com/sorbet/sorbet/pull/7411, https://github.com/sorbet/sorbet/pull/7414, https://github.com/sorbet/sorbet/pull/7418, https://github.com/sorbet/sorbet/pull/7419)

During namer run Sorbet creates symbols for type members. If the type member is a type template, then namer will also create a static field with the same name: https://github.com/sorbet/sorbet/blob/d7257093237535e120787dc4a23de1f9699b77ea/namer/namer.cc#L1448-L1464

There is a comment explaining why is it needed for resolver, but I don't fully understand it.

During the fast path if the location of the type template change (and the location of the static field change respectively), Sorbet will not update it. The PR introduces a fix to that bug.

### Motivation

On a branch with [DefinitionLinesDenylistEnforcer](https://github.com/sorbet/sorbet/pull/7381/) the [generic_class_overrides.rb test](https://github.com/sorbet/sorbet/blob/master/test/testdata/infer/generic_class_overrides.rb) has been failing, because the location for [the Arg static field](https://github.com/sorbet/sorbet/blob/master/test/testdata/infer/generic_class_overrides.rb#L128) was not updated prorely

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -1399,6 +1399,8 @@ private:
 
         core::TypeMemberRef sym;
         auto existingTypeMember = ctx.state.lookupTypeMemberSymbol(onSymbol, typeMember.name);
+        auto typeMemberName = typeMember.name;
+
         if (existingTypeMember.exists()) {
             // if we already have a type member but it was constructed in a different file from the one we're
             // looking at, then we need to raise an error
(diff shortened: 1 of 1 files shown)


namer/namer.cc
```diff
@@ -1432,27 +1434,28 @@ class SymbolDefiner {
...
             sym.data(ctx)->resultType = core::make_type<core::LambdaParam>(sym, todo, todo);
+        }
 
-            if (isTypeTemplate) {
-                auto typeTemplateAliasName = typeMember.name;
-                auto context = ctx.owner.enclosingClass(ctx);
+        if (isTypeTemplate) {
+            auto typeTemplateAliasName = typeMember.name;
+            auto context = ctx.owner.enclosingClass(ctx);
+            if (!existingTypeMember.exists()) {
```
jez, line 1453: What do you think about flipping this condition, so that it's the same as the `existingTypeMember.exists()` condition above?

ilyailya: No problem. 
I originally decided to make it `!existingTypeMember.exists()`, because it fits the journey of type template better. At the first run of namer type member doesn't exist, so we create a new static field (first branch). On the second run and after it for the same type member we would use only the second branch. But yeah, I'm not holding to that

jez (approved):

PR #7426, opened 2023-10-26
Mark `BasicObject#==` overridable

jez:
### Motivation

This obviates the need for the FAQ entry explaining people what
signature to use when overriding `BasicObject#==`.

Users can always use `override(allow_incompatible: true)` to opt out of
these checks in existing signatures.

### Test plan

wip

### DIFF
--- a/rbi/core/basic_object.rbi
+++ b/rbi/core/basic_object.rbi
@@ -117,10 +117,11 @@ class BasicObject
   # 1.eql? 1.0   #=> false
   # ```
   sig do
-    params(
-        other: BasicObject,
-    )
-    .returns(T::Boolean)
+    overridable
+      .params(
+        other: T.anything,
+      )
+      .returns(T::Boolean)
   end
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_Ot45JkxuSQKIfH
→ https://go/builds/bui_Ot45lEOXrAZGif
→ https://go/builds/bui_Ot45zUpOIyfAOZ

jez:
This would be good to land, but the biggest downside right now is that we don’t have autocorrects for any of the new errors this would cause.

I would feel a lot better about this if we had autocorrects to:

* insert `override.` when required
* replace a child parameter type with a parent parameter type
* replace a child return type with a parent parameter type
* use `override(allow_incompatible: true)` if `--suggest-unsafe` was passed

All of these autocorrects are tricky to implement right now because of how validator.cc is structured. (But it would not be impossible!)

PR #7439, opened 2023-10-27
Customize `T.must` error in Ruby 3.1+

jez:
### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/errors/t_must_type_error.rb
+++ b/gems/sorbet-runtime/lib/types/errors/t_must_type_error.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+# typed: true
+
+class T::MustTypeError < TypeError
+  # This approximates what the error_highlight gem does
+  if RubyVM::AbstractSyntaxTree.method(:of).parameters.include?([:key, :keep_script_lines])
+    def generate_snippet
+      # The line at index 1 in the backtrace is the call to T.must
+      backtrace_location = self.backtrace_locations&.[](1)
+      return "" unless backtrace_location
+
+      # Newer versions of error_highlight allow passing a backtrace_location
+      # directly to ErrorHighlight.spot doign it ourself is not much work at the
+      # benefit of working no matter what error_highlight version a project uses.
+      node = RubyVM::AbstractSyntaxTree.of(backtrace_location, keep_script_lines: true)
+      return "" unless node
+
+      spot = ErrorHighlight.spot(node, point_type: :args)
+      return "" unless spot
+
+      return ErrorHighlight.formatter.message_for(spot)
+    end
+  else
+    # ErrorHighlight requires `keep_script_lines: true` to get the text source of the node
+    def generate_snippet
+      ""
+    end
+  end
+
+  private :generate_snippet
+
+  if Exception.method_defined?(:detailed_message)
+    # New in Ruby 3.2
+    def detailed_message(highlight: false, error_highlight: true, **)
+      return super unless error_highlight
+      snippet = generate_snippet
+      if highlight && !snippet.empty?
+        snippet = "\e[1m" + snippet + "\e[m"
+      end
+      super + snippet
+    end
+  else
+    # For Ruby 3.1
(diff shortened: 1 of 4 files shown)


gems/sorbet-runtime/lib/types/errors/t_must_type_error.rb
```diff
@@ -0,0 +1,65 @@
...
+class T::MustTypeError < TypeError
+  # This approximates what the error_highlight gem does.
+  if RubyVM::AbstractSyntaxTree.method(:of).parameters.include?([:key, :keep_script_lines])
+    def generate_snippet
+      # The line at index 1 in the backtrace is the call to T.must
+      backtrace_location = self.backtrace_locations&.[](1)
+      return "" unless backtrace_location
+
+      # Newer versions of error_highlight allow passing a backtrace_location
+      # directly to ErrorHighlight.spot doign it ourself is not much work at the
```
paracycle, line 13:
```suggestion
      # directly to ErrorHighlight.spot doing it ourself is not much work at the
```

mame:
I think that the current code is fine enough if it works, but my originally intended usage of `ErrorHighlight::CoreExt` is to prepend it to an error class that you want to enhance with error_highlight.

```ruby
if defined?(ErrorHighlight::CoreExt) && !self.include?(ErrorHighlight::CoreExt)
  T::MustTypeError..prepend(ErrorHighlight::CoreExt)
end
```

This way, you don't have to copy and paste the `error_highlight` code.

But you want to replace `generate_snippet` with your own definition, right? I did not think much of this usage, but how about prepending another module in addition to `ErrorHighlight::CoreExt`?

```ruby
class T::MustTypeError < TypeError
  module SnippetGenerator
    private def generate_snippet
      ... # Your definition
    end
  end
  private_constant :SnippetGenerator

  if defined?(ErrorHighlight::CoreExt) && !self.include?(ErrorHighlight::CoreExt)
    MustTypeError.prepend(ErrorHighlight::CoreExt)
    MustTypeError.prepend(SnippetGenerator)
  end
end
```

I think this respects your definition of `generate_snippet`, but sorry if it doesn't work well. I haven't actually tried it.

jez:

→ https://go/builds/bui_PX5zrfKjopfpEd
→ https://go/builds/bui_PX5zYJQbfx8fEA
→ https://go/builds/bui_PX5zfnG0FnohiJ

→ https://go/builds/bui_PX64uBnoTKy5SQ

jez:
Thanks for the feedback @mame! It helps a lot.

But there's something else that came up in my testing of this feature: it doesn't play nicely with `load_iseq` monkey patches.

I have an example of what I mean in this branch

https://github.com/sorbet/sorbet/pull/7684

But basically, the transformations that happen inside `load_iseq` don't get applied when the file is re-parsed inside of `RubyVM::AbstractSyntaxTree.of`, which means that the `node_id`'s are thrown off when attempting to re-parse the file and find the node:

https://github.com/ruby/ruby/blob/957bb7cb81995f26c671afce0ee50a5c660e540e/ast.c#L244

This then makes the `error_highlight` gem highlight the wrong node.

If you have any ideas whether there's a way to work around this, I'd be very eager to hear.

(Some context about why we have this `load_iseq` patch in the first place: our dev tooling loads constant assignments lazily in tests to speed up our test environment, so that simply requiring a file doesn't eagerly evaluate constant assignments.)

mame:
@jez Ummm. You may want to try `RubyVM.keep_script_lines = true`. This mode makes the source code text saved on memory at compile time. `RubyVM::AbstractSyntaxTree.of` will use the on-memory text if any, instead of reading the original file. So error_highlight should work correctly (though the output will be based on the transformed text).

However, note that `RubyVM.keep_script_lines = true` will consumes memory, of course. I hope that is acceptable.

I wonder if there should be a hook that allows the code transformation when `RubyVM::AbstractSyntaxTree.of` reads a file. Unfortunately, such a hook is currently not available.

jez:
Sounds good, I’ll try that, thanks for the fast response!

PR #7443, opened 2023-10-27
wip: Make autogen include references inside `class << self`

jez:
### Motivation

There's a problem on this line:

https://github.com/sorbet/sorbet/blob/ef829a5443d94ee658956e89da2e24bcfaff3656/main/autogen/autogen.cc#L84

We short circuit attempting to find autogen references to `include` and `extend` lines inside `class << self` definitions. This breaks some parts of depdb at Stripe.

### Test plan

See included automated tests.


test/testdata/autogen/self_class.rb
```diff
@@ -0,0 +1,13 @@
+# typed: true
+
+module DSL; end
+
+class A
+  class << self
+    include DSL
```
jez, line 7: There should be a `ref` for this `DSL` constant

test/testdata/autogen/self_class.rb.autogen.exp
```diff
@@ -0,0 +1,51 @@
...
+ type=class
+ defines_behavior=1
+ is_empty=0
+ defining_ref=[A]
+[def id=3]
+ type=class
+ defines_behavior=1
+ is_empty=1
+ defining_ref=[B]
+## refs:
```
jez, line 23: But the list of refs here does not include that constant (note that none of the refs are online 7)

PR #7453, opened 2023-11-02
Add OpenSSL (fixed length) secure compare methods

bdewater:
### Motivation
These methods were added in OpenSSL gem version 2.2, which first shipped with Ruby 3.0 but is available for older Rubies as well. See https://ruby-doc.org/3.0.6/exts/openssl/OpenSSL.html#method-c-fixed_length_secure_compare

### Test plan
See included automated tests.

### DIFF
--- a/rbi/stdlib/openssl.rbi
+++ b/rbi/stdlib/openssl.rbi
@@ -649,6 +649,21 @@ module OpenSSL
     .returns(::T.untyped)
   end
   def self.fips_mode=(fips_mode); end
+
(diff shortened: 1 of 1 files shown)


neilparikh:

→ https://go/builds/bui_Ovzwjo3LqLUL1i
→ https://go/builds/bui_OvzweIPD6aCBDj
→ https://go/builds/bui_OvzxWRtZAjuBRv

neilparikh (approved):
Thanks!

PR #7454, opened 2023-11-02
Fix Loc-related crash in definition_validator

jez:
### Motivation

This crash did not actually manifest as a problem with updating locs on the fast
path. Instead, it was a simple logic bug where we were assuming that the current
`ClassDef` AST node would also necessarily be the same as the canonical
definition of the class's symbol.

In general that's a bad assumption. Specifically, this usually manifests for
classes which have definitions split between a source and RBI file.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/definition_validator/superclass_class_loc__2.rbi
+++ b/test/testdata/definition_validator/superclass_class_loc__2.rbi
@@ -13,6 +13,6 @@
 # the other file. With a sufficiently long comment, that causes an array access
 # out of bounds.
 
-class Child < Parent
+class Child < Parent # error: does not derive from `Class`
   def example; end
 end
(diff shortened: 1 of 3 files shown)


test/testdata/definition_validator/superclass_class_loc__2.rbi
```diff
@@ -0,0 +1,18 @@
...
+# the other file. With a sufficiently long comment, that causes an array access
+# out of bounds.
+# This is a rather large comment at the start of the file which exposes a bug
+# where we were accidentally turning a LocOffset from this file into a Loc for
+# the other file. With a sufficiently long comment, that causes an array access
+# out of bounds.
+# This is a rather large comment at the start of the file which exposes a bug
+# where we were accidentally turning a LocOffset from this file into a Loc for
+# the other file. With a sufficiently long comment, that causes an array access
+# out of bounds.
```
jez, line 14: Note that this is separate from the fast path—this causes a crash even in the slow path.

ilyailya (approved):
Great catch!

PR #7460, opened 2023-11-03
Don't consider T.must(T.untyped) redundant

marianosimone:
### Motivation
In my opinion, calling `T.must` with an untyped object is not redundant, and the autofix for it changes the semantics of the code.

Consider the following:

```
sig {params(something: T.untyped).void}
def foo(something)
    T.must_because(something) { "Called foo with nil!" }.bar
end
```

The author was expecting to catch (through `T.must`) the exact place where `something` is `nil`, and have their custom message.

The autofix proposes changing this just to:

```
sig {params(something: T.untyped).void}
def foo(something)
    something.bar
end
```
And depend on Ruby's behavior for handling the `nil`, or for whatever `method_missing` handling might be going on.

`T.must` here is not redundant: it's a way to be explicit about what we want to happen when the object is `nil`.

This, tied with the fact that there is now an option to highlight untyped code makes me think that we need to remove this error.

### Test plan
Modified existing tests

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1857,14 +1857,9 @@ public:
             return;
         }
         auto ret = Types::dropNil(gs, args.args[0]->type);
-        if (ret == args.args[0]->type) {
+        if (ret == args.args[0]->type && !args.args[0]->type.isUntyped()) {
             if (auto e = gs.beginError(args.argLoc(0), errors::Infer::InvalidCast)) {
-                if (args.args[0]->type.isUntyped()) {
-                    e.setHeader("`{}` called on `{}`, which is redundant", methodName, args.args[0]->type.show(gs));
-                } else {
-                    e.setHeader("`{}` called on `{}`, which is never `{}`", methodName, args.args[0]->type.show(gs),
-                                "nil");
-                }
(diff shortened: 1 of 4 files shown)


jez:
Unfortunately I think this is a contentious change, which biases in favor of preserving the existing behavior.

The fact that `T.must` has the behavior it does means that `T.must` serves as a guarantee that a piece of code is typed. If this error shows up after a change, it's an indication that the code became untyped.

One thing which I think I would be open to would be to use a different error code for this, so that they could be isolated with `--isolate-error-code` separately from the other `T.must` errors.

marianosimone-stripe:
Discarded in favor of [[sorbet/pr/7462] ✨ Introduce MustOnUntyped error, and separate it from InvalidCast](https://github.com/sorbet/sorbet/pull/7462)

marianosimone-stripe:
Discarded in favor of [[sorbet/pr/7462] ✨ Introduce MustOnUntyped error, and separate it from InvalidCast](https://github.com/sorbet/sorbet/pull/7462)

marianosimone:
Discarded in favor of [[sorbet/pr/7462] ✨ Introduce MustOnUntyped error, and separate it from InvalidCast](https://github.com/sorbet/sorbet/pull/7462)

PR #7015, opened 2023-05-16
Widen type of `Array#+`

jez:
### Motivation

Historically we held this type artificially narrow because we thought the
error messages would be more direct if we rejected this code. More context
in this change: #4747

But the workarounds are annoying, and the error messages are not actually
that bad. (In particular, the workaround involved mutating the receiver,
something that `Array#+` doesn't do.)

### Test plan

See included automated tests.

PR #7462, opened 2023-11-03
✨ Introduce MustOnUntyped error, and separate it from InvalidCast

marianosimone:
### Motivation
This is an alternative to [[sorbet/pr/7460] Don't consider T.must(T.untyped) redundant](https://github.com/sorbet/sorbet/pull/7460)

### Motivation
In my opinion, calling `T.must` with an untyped object is not necessarily redundant, and the autofix for it changes the semantics of the code.

Consider the following:

```
sig {params(something: T.untyped).void}
def foo(something)
    T.must_because(something) { "Called foo with nil!" }.bar
end
```

The author was expecting to catch (through `T.must`) the exact place where `something` is `nil`, and have their custom message.

The autofix proposes changing this just to:

```
sig {params(something: T.untyped).void}
def foo(something)
    something.bar
end
```
And depend on Ruby's behavior for handling the `nil`, or for whatever `method_missing` handling might be going on.

`T.must` here is not redundant: it's a way to be explicit about what we want to happen when the object is `nil`.

Separating this error from [Sorbet Error Reference #7015](https://sorbet.org/docs/error-reference#7015) allows users to distinguish between them, and accept autofixes for one and not the other through `--isolate-error-code`

### Test plan
See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1858,8 +1858,9 @@ public:
         }
         auto ret = Types::dropNil(gs, args.args[0]->type);
         if (ret == args.args[0]->type) {
-            if (auto e = gs.beginError(args.argLoc(0), errors::Infer::InvalidCast)) {
-                if (args.args[0]->type.isUntyped()) {
+            auto code = args.args[0]->type.isUntyped() ? errors::Infer::MustOnUntyped : errors::Infer::InvalidCast;
+            if (auto e = gs.beginError(args.argLoc(0), code)) {
+                if (code == errors::Infer::MustOnUntyped) {
                     e.setHeader("`{}` called on `{}`, which is redundant", methodName, args.args[0]->type.show(gs));
                 } else {
                     e.setHeader("`{}` called on `{}`, which is never `{}`", methodName, args.args[0]->type.show(gs),
--- a/core/errors/infer.h
+++ b/core/errors/infer.h
@@ -53,6 +53,7 @@ constexpr ErrorClass NonOverlappingEqual{7046, StrictLevel::True};
 constexpr ErrorClass UntypedValueInformation{7047, StrictLevel::True};
 constexpr ErrorClass UnknownSuperMethod{7048, StrictLevel::True};
 constexpr ErrorClass TypecheckOverloadBody{7049, StrictLevel::True};
+constexpr ErrorClass MustOnUntyped{7050, StrictLevel::Strict};
 // N.B infer does not run for untyped call at all. StrictLevel::False here would be meaningless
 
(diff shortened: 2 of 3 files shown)


core/types/calls.cc
```diff
@@ -1858,9 +1858,11 @@ class T_must : public IntrinsicMethod {
         }
         auto ret = Types::dropNil(gs, args.args[0]->type);
         if (ret == args.args[0]->type) {
-            if (auto e = gs.beginError(args.argLoc(0), errors::Infer::InvalidCast)) {
+            auto e = args.args[0]->type.isUntyped() ? gs.beginError(args.argLoc(0), errors::Infer::MustOnUntyped)
+                                                    : gs.beginError(args.argLoc(0), errors::Infer::InvalidCast) if (e) {
                 if (args.args[0]->type.isUntyped()) {
-                    e.setHeader("`{}` called on `{}`, which is redundant", methodName, args.args[0]->type.show(gs));
+                    e.setHeader("`{}` called on `{}`, which might be redundant", methodName,
```
jez, line 1864: Let's not change this message. Error messages in Sorbet do not second guess themselves.

core/types/calls.cc
```diff
@@ -1858,9 +1858,11 @@ class T_must : public IntrinsicMethod {
         }
         auto ret = Types::dropNil(gs, args.args[0]->type);
         if (ret == args.args[0]->type) {
-            if (auto e = gs.beginError(args.argLoc(0), errors::Infer::InvalidCast)) {
+            auto e = args.args[0]->type.isUntyped() ? gs.beginError(args.argLoc(0), errors::Infer::MustOnUntyped)
+                                                    : gs.beginError(args.argLoc(0), errors::Infer::InvalidCast) if (e) {
```
jez, line 1862:
```suggestion
            auto code = args.args[0]->type.isUntyped() ? errors::Infer::MustOnUntyped : 
errors::Infer::InvalidCast;
            if (auto e = gs.beginError(args.argLoc(0), code)) {
```

core/types/calls.cc
```diff
@@ -1858,9 +1858,11 @@ class T_must : public IntrinsicMethod {
...
                 if (args.args[0]->type.isUntyped()) {
```
jez, line 1863:
```suggestion
                if (code == errors::Infer::MustOnUntyped) {
```

core/types/calls.cc
marianosimone: Done

core/types/calls.cc
marianosimone: Done

core/types/calls.cc
marianosimone: Done

jez (approved):

PR #7464, opened 2023-11-05
Drop 'invalid cast' error down to `# typed: true`

marianosimone:
### Motivation
Using `T.let`, `T.cast` and `T.must` in places where are not necessary adds noise to the code.
Those are also blunt tools that we sometimes use to bypass other Sorbet errors, so they hide real problems.

### Test plan
- Pre-existing tests
- Ran against Stripe's internal codebase to see how it caught real issues


qaisjp:
nice catch

froydnj (approved):

marianosimone:
I just realized that this hasn't been merged :/
Given that I no longer have access to Stripe's codebase, I don't know how big of a deal it would be, so I'm closing it and letting y'all decide if it's worth doing (I still think it is... but somehow fell off my radar, sorry!)

PR #7475, opened 2023-11-13
Extract to variable code action

neilparikh:
Implements a code action to extract a selected expression into its own expression.

Suggested order to review the core logic in this PR:
- `code_action.cc`
- `ExtractVariableWalk` in `ExtractVariable.cc`
- `getExtractVariableEdits` in `ExtractVarible.cc`

### Motivation

Make it easy to refactor code.

### Test plan

See included automated tests.

In particular `bugs_false_negatives.rb` and `bugs_false_positives.rb` show the cases where the code action does the wrong thing (in most cases due to how desugar works). I think they're all minor enough that we can leave them as is for now, and then fix them in a future PR if we can figure out a good way to deal with them


main/lsp/requests/code_action.cc
```diff
@@ -1,3 +1,5 @@
+#include <algorithm>
```
jez, line 1: Let's delete this and use `absl::c_find` instead

main/lsp/requests/code_action.cc
```diff
@@ -14,6 +17,82 @@ using namespace std;
 namespace sorbet::realmain::lsp {
 
 namespace {
+class ExtractVariableWalk {
+    core::Loc target_loc;
+    core::LocOffsets deepest_loc;
+    std::vector<core::LocOffsets> invalid_locs;
```
jez, line 23: This is a nit but reminder that we use camel case in sorbet

main/lsp/requests/code_action.cc
```diff
@@ -260,6 +346,76 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerDelegate &t
...
+            auto desugaredTree = typechecker.getDesugared(loc.file());
+            core::Context ctx(gs, core::Symbols::root(), loc.file());
+            ast::TreeWalk::apply(ctx, extractVariableWalk, desugaredTree);
+
+            if (extractVariableWalk.getFoundExactMatch()) {
+                auto tdi =
+                    make_unique<VersionedTextDocumentIdentifier>(move(params->textDocument->uri), JSONNullObject());
+                vector<unique_ptr<TextEdit>> edits;
+
+                core::Loc whereToInsert;
```
jez, line 360: You can simplify this a little if whereToInsert is a LocOffsets and construct the `Loc` only once.

test/testdata/lsp/code_actions/extract_variable.A.rbedited
jez, line 1: some suggestions for more tests:

- various selection targets on `prop :foo, Integer`, which should be fine because we do the transformation on desugared trees, so the `prop` hasn't been expanded to anything
- `puts || puts` where the selection is `" || "`
  - Things like this are an argument in favor of using the parse tree instead of the desugar tree for implementing extract variable. We can chat about this.
- These two:

  ```ruby
  case x.foo
    #  ^^^^^
  when Integer
    #  ^^^^^^^
  end
  ```

jez (commented):
I haven't managed to finish this review. Sorry.

I have some preliminary comments which should hopefully keep you busy.

I will say, that I'm very impressed how well this works already so far, and that makes me optimistic that with a little bit of time spent coming up with more test cases we'll be able to ship this before long.

One bit of parting advice: the best way to find test cases for this is to scrutinize the transformations that desugar is doing. In particular: it is _not_ the case that traversing the tree in order means traversing the locs in order. The `||` case I suggest below is one example. Try to see if there are other examples, especially for cases where traversing a single InsSeq or ClassDef::rhs means visiting the locs out of order. This _definitely_ happens later in the pipeline (e.g., rewriter/Flatten.cc will definitely make these transformations). But I'm not sure yet whether Desugar does. The correctness of the algorithm implemented here depends on this property, so we should be very sure that it's true.

main/lsp/ExtractVariable.cc
```diff
@@ -0,0 +1,239 @@
+#include "main/lsp/ExtractVariable.h"
+#include "ast/treemap/treemap.h"
+
+using namespace std;
+
+namespace sorbet::realmain::lsp {
+
+class ExtractVariableWalk {
```
neilparikh, line 8: This tree walk has the core logic of the code action. It has 2 aims:
- check if the selection is a valid selection to extract
- find which "scope" the selection is inside

The first is mostly straightforward. In `preTransformExpression`, if we find an node with the exact same `loc`, and it's not a node we can't extract (like `return`), we can allow extracting that selection. The one case that requires some extra handling is when we have identifiers (which are usually fine to extract) in positions that disallow extraction. For example, in `def foo(x); end; y = 1`, extracting both `x` and `y` would be invalid. To handle this, during the tree walk, we call `skipLoc` on locs that represent those invalid locations. For example, in `preTransformMethodDef`, we loop through each arg and call `skipLoc` for each of the arguments' loc. This means that if the selection is one of the arguments, `foundExactMatch` will still return false since it's loc will be in `skippedLocs`.

The second is bit more complicated to do. Naively, we could just look for the deepest `InsSeq` that contains the selection and say that's the scope where it's defined. The problem with that approach is that we omit for `InsSeq`s for single line scope. For example, `def foo(x); x; end` will have the body of the method as a reference to a `UnresolvedIdent` directly rather than an `InsSeq` that contains the identifier. To handle this, we also need to walk over all the nodes that could introduce a "scope" (`if`, `methodDef`, `classDef` etc.). We still match on `InsSeq`s in addition, so that if scope does have a `InsSeq`, we'll use that instead.

main/lsp/ExtractVariable.cc
```diff
@@ -0,0 +1,239 @@
...
+            if (insSeq->expr.loc().contains(locOffsets)) {
+                ENFORCE(!ast::isa_tree<ast::InsSeq>(insSeq->expr));
+                whereToInsert = insSeq->expr.loc();
+            }
+            ENFORCE(whereToInsert.exists());
+        } else if (auto block = ast::cast_tree<ast::Block>(*enclosingScope)) {
+            ENFORCE(!ast::isa_tree<ast::InsSeq>(block->body));
+            whereToInsert = block->body.loc();
+        } else if (auto methodDef = ast::cast_tree<ast::MethodDef>(*enclosingScope)) {
+            // TODO(neil): this will fail for endless methods
```
neilparikh, line 161: Endless methods aren't supported in pay-server yet, so we can handle this later.

main/lsp/requests/code_action.cc
```diff
@@ -260,6 +264,23 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerDelegate &t
...
+                    action->kind = CodeActionKind::RefactorExtract;
+
+                    auto workspaceEdit = make_unique<WorkspaceEdit>();
+                    workspaceEdit->documentChanges = move(documentEdits);
+
+                    action->edit = move(workspaceEdit);
+                    result.emplace_back(move(action));
+
+                    // TODO(neil): trigger a rename for newVariable
+                    // TODO(neil): replace other occurences of this expression with newVariable
```
neilparikh, line 279: These 2 will be done in a future PR.

main/lsp/ExtractVariable.cc
```diff
@@ -0,0 +1,239 @@
...
+
+    if (extractVariableWalk.foundExactMatch()) {
+        vector<unique_ptr<TextEdit>> edits;
+
+        auto locOffsets = loc.offsets();
+        core::LocOffsets whereToInsert = core::LocOffsets::none();
+        auto enclosingScope = extractVariableWalk.enclosingScope;
+        // For all cases except InsSeq and ClassDef, extractVariableWalk.enclosingScopeLoc should be
+        // the same as what we're pulling out from the ExpressionPtr, but we'll just get it directly
+        // for consistency.
```
neilparikh, line 143: Once we have the scope that the selection is inside, we need to walk the through the scope to find where to insert the new variable declaration. For one line scope, that's easy: we can put it at the start of the scope, since that will be before the selection. For `InsSeq` and `ClassDef`, we walk through each statement to find the statement that contains the selection, and insert just before that statement.

main/lsp/LSPTypechecker.h
```diff
@@ -108,6 +108,10 @@ class LSPTypechecker final {
     LSPQueryResult query(const core::lsp::Query &q, const std::vector<core::FileRef> &filesForQuery,
                          WorkerPool &workers) const;
 
+    /**
+     * Returns the parsed file for the given file, up to the desugar pass.
+     */
```
jez, line 113:
```suggestion
    /**
     * Returns the parsed file for the given file, up to the desugar pass.
     *
     * This is never cached, which means that the file will be re-parsed from scratch.
     * This is slower than getting the indexed tree (everything before namer), so if
     * You can use the indexed tree that will be more performant. Certain IDE actions
     * need particularly fine-grained fidelity in the AST (precludes rewriter).
     */
```

main/pipeline/pipeline.cc
```diff
@@ -168,6 +168,16 @@ ast::ParsedFile emptyParsedFile(core::FileRef file) {
...
 
+ast::ExpressionPtr desugarOne(const options::Options &opts, core::GlobalState &gs, core::FileRef file) {
+    auto &print = opts.print;
+
+    if (file.data(gs).strictLevel == core::StrictLevel::Ignore) {
+        return ast::MK::EmptyTree();
+    }
+    auto parseTree = runParser(gs, file, print, opts.traceLexer, opts.traceParser);
+    return runDesugar(gs, file, move(parseTree), print);
+}
```
jez, line 179: We probably want to include the try/catch that `indexOne` has here also.

main/pipeline/pipeline.cc
```diff
@@ -168,6 +168,16 @@ ast::ParsedFile emptyParsedFile(core::FileRef file) {
     return {ast::MK::EmptyTree(), file};
 }
 
+ast::ExpressionPtr desugarOne(const options::Options &opts, core::GlobalState &gs, core::FileRef file) {
+    auto &print = opts.print;
```
jez, line 172: Let's go ahead and add a Timer like in indexOne.

main/pipeline/pipeline.h
```diff
@@ -16,6 +16,8 @@ namespace sorbet::realmain::pipeline {
 ast::ParsedFile indexOne(const options::Options &opts, core::GlobalState &lgs, core::FileRef file,
                          ast::ExpressionPtr cachedTree = nullptr);
 
+ast::ExpressionPtr desugarOne(const options::Options &opts, core::GlobalState &gs, core::FileRef file);
```
jez, line 19:
```suggestion
// Primarily exposed for LSP—outside of LSP, you probably want `indexOne`.
ast::ExpressionPtr desugarOne(const options::Options &opts, core::GlobalState &gs, core::FileRef file);
```

ast/desugar/Desugar.cc
```diff
@@ -1929,6 +1929,7 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
             },
             [&](parser::Return *ret) {
                 if (ret->exprs.size() > 1) {
+                    auto arrayLoc = ret->exprs.front()->loc.join(ret->exprs.back()->loc);
```
jez, line 1932: This is the kind of change that would be nice to land as a separate change, with tests, because it _is_ possible to write tests for this behavior absent the code action changes:

[View on sorbet.run →](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20returns%28Integer%29%20%7D%0Adef%20example1%0A%20%20return%20''%0Aend%0A%0Asig%20%7B%20returns%28T%3A%3AArray%5BInteger%5D%29%20%7D%0Adef%20example2%0A%20%20return%20''%2C%201%0Aend%0A%0Asig%20%7B%20returns%28Integer%29%20%7D%0Adef%20example3%0A%20%20''%0Aend%0A%0Asig%20%7B%20returns%28Integer%29%20%7D%0Adef%20example4%0A%20%20if%20%5Btrue%2C%20false%5D.sample%0A%20%20%20%200%0A%20%20else%0A%20%20%20%20''%0A%20%20end%0Aend)

main/lsp/ExtractVariable.h
```diff
@@ -0,0 +1,17 @@
+#ifndef SORBET_EXTRACT_VARIABLE_H
+#define SORBET_EXTRACT_VARIABLE_H
+
+#include "main/lsp/LSPConfiguration.h"
+#include "main/lsp/LSPTypechecker.h"
+#include "main/lsp/json_types.h"
+
+namespace sorbet::realmain::lsp {
+
+std::vector<std::unique_ptr<TextDocumentEdit>> getExtractVariableEdits(LSPTypecheckerDelegate &typechecker,
```
jez, line 10: Slight preference for not polluting the top-level `lsp` namespace with one-off functions. You can see how `sig_finder.{h,cc}` do this, where the main public method is a static method inside a class, so you'd have something like


```
VariableExtractor::getEdits
```

jez: Do we need to take ownership of the selectionRange? It's unexpected for me that this method would require mutating the enclosing `LSPTask::params`. It means that no one else could use the `params->range` in code_action.cc unless they duplicate the `Range` first. But this method isn't mutating the selectionRange, so it could be taken like `const Range& selectionRange`.

Also in general, we should avoid using LSP specification types for method arguments outside of helper functions, especially when there are suitable core Sorbet data structures that model them.

Also, it looks like the one call to `getExtractVariableEdits` passes both `move(params->range)` and also `loc`. But `loc` was already created by converting `params->range` to a `core::Loc`, so it seems like it takes the same argument twice in different forms

main/lsp/ExtractVariable.h
```diff
@@ -0,0 +1,17 @@
...
+                                                                       const LSPConfiguration &config,
+                                                                       std::unique_ptr<Range> selectionRange,
+                                                                       const core::Loc selectionLoc);
```
jez, line 13: Do we need to take ownership of the selectionRange? It's unexpected for me that this method would require mutating the enclosing `LSPTask::params`. It means that no one else could use the `params->range` in code_action.cc unless they duplicate the `Range` first. But this method isn't mutating the selectionRange, so it could be taken like `const Range& selectionRange`.

Also in general, we should avoid using LSP specification types for method arguments outside of helper functions, especially when there are suitable core Sorbet data structures that model them.

Also, it looks like the one call to `getExtractVariableEdits` passes both `move(params->range)` and also `loc`. But `loc` was already created by converting `params->range` to a `core::Loc`, so it seems like it takes the same argument twice in different forms

test/testdata/lsp/code_actions/extract_variable/block.rb
jez, line 1: amazing test cases

test/testdata/lsp/code_actions/extract_variable/bugs_false_negatives.rb
jez, line 1: Let's take some action items to make tickets or fast-follow PRs for these after this PR lands, for the ones that we have ideas for how to fix.

test/testdata/lsp/code_actions/extract_variable/other.rb
```diff
@@ -0,0 +1,24 @@
+# typed: true
+# selective-apply-code-action: refactor.extract
+
+if 1 == 2
+  1 + 1
+#     ^ apply-code-action: [A] Extract Variable
```
jez, line 6: Out of curiosity, did you look at any other languages to see whether/how they decide to handle this case?

test/testdata/lsp/code_actions/extract_variable/method.rb
```diff
@@ -0,0 +1,25 @@
+# typed: true
+# selective-apply-code-action: refactor.extract
+
+def b; 1 + 123; end
+#          ^^^ apply-code-action: [B] Extract Variable
```
jez, line 5: I don't think this matters because these letters are just opaque tokens, but it looks like you're missing `[A]` in this file. not sure how much that matters to you.

test/testdata/lsp/code_actions/extract_variable/rescue.rb
jez, line 1: Another great test case.

test/testdata/lsp/code_actions/extract_variable/invalid.rb
```diff
@@ -0,0 +1,100 @@
+# typed: true
+# selective-apply-code-action: refactor.extract
+# assert-no-code-action: refactor.extract
+def example(x)
+#           ^ apply-code-action: [A] Extract Variable
```
jez, line 5: I wasn't aware it would work like this, where every code action can use `[A]`. Does it just mean that the test harness should read the contents of the `.A.rbedited` file to verify the resulting contents? It's not trying to like batch them all up and send them all at once, is it?

test/testdata/lsp/code_actions/extract_variable/invalid.rb
```diff
@@ -0,0 +1,100 @@
...
+  A.new.a ||= 1
+# ^^^^^^^ apply-code-action: [A] Extract Variable
+
+  A.new.a *= 1
+# ^^^^^^^ apply-code-action: [A] Extract Variable
+
+a = T.let(1, T.untyped)
+
+  a &&= 1
+# ^ apply-code-action: [A] Extract Variable
```
jez, line 57: Not sure how important it is to you, but you may want to write a test in ProtocolTest.cc to assert that if you ask for a code action with a zero-length range that no code action gets applied (or does the test harness have a way to specify a zero-length range natively? I didn't think it did)

test/testdata/lsp/code_actions/extract_variable/invalid.rb
```diff
@@ -0,0 +1,100 @@
...
+#      ^^^^^^^^^^^^^ apply-code-action: [A] Extract Variable
+rescue ArgumentError => e
+#      ^^^^^^^^^^^^^^^^^^ apply-code-action: [A] Extract Variable
+rescue Exception => e
+#                   ^ apply-code-action: [A] Extract Variable
+end
+
+class B < T::Struct
+  const :bar, String
+#             ^^^^^^ apply-code-action: [A] Extract Variable
```
jez, line 96: Why does this one not work? Does it work not work for any constant literal?

main/lsp/requests/code_action.cc
```diff
@@ -260,6 +262,27 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerDelegate &t
             action->kind = CodeActionKind::RefactorRewrite;
             action->edit = move(workspaceEdit);
             result.emplace_back(move(action));
+        } else {
+            if (loc.beginPos() != loc.endPos()) {
+                // For move method to new module we use canResolveLazily to defer the computation
+                // until the user has actually selected the action. We can't do that here because
+                // we need to do the core computation to know if extract the current selection is
+                // valid in the first place, to decide if we can show the code action or not.
```
jez, line 270: Let's add a Timer here. I'm still worried that this will be too slow, and the Timer will tell us whether it is or isn't

jez (commented):
Looks great, I don't have any substantive feedback.

Not going to approve until it's behind a beta flag, so feel free to send it back when ready.

Anything marked "nit" is optional but I would appreciate it.

main/lsp/requests/code_action.cc
```diff
@@ -260,6 +262,29 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerDelegate &t
             action->kind = CodeActionKind::RefactorRewrite;
             action->edit = move(workspaceEdit);
             result.emplace_back(move(action));
+        } else {
+            if (loc.beginPos() != loc.endPos() && config.opts.lspExtractToVariableEnabled) {
```
jez, line 266: w.r.t. the discussion below of "making a selection in a const prop means that we don't get this refactor anymore"

We could invert this, and say that if there's a selection, the extract to variable refactor takes priority over the move method to new module one.

I'm fine doing that in a follow up PR.

jez (approved):
Looks great!

neilparikh:

→ https://go/builds/bui_PJwx34whWk1lok
→ https://go/builds/bui_PJwxisn88I5zKU
→ https://go/builds/bui_PJwxRjDgNmwuRy

PR #7479, opened 2023-11-14
Don't mutate isPackage bit after file creation

jez:
### Motivation

We had a special case for ignored package files in packager. It's
cleaner to handle this without mutation.

### Test plan

n/a

### DIFF
--- a/core/Files.cc
+++ b/core/Files.cc
@@ -346,11 +346,9 @@ bool File::permitOverloadDefinitions() const {
 }
 
 bool File::isPackage() const {
-    return flags.isPackage;
-}
-
-void File::setIsPackage(bool isPackage) {
-    this->flags.isPackage = isPackage;
+    // If the `__package.rb` file is at `typed: ignore`, then we haven't even parsed it.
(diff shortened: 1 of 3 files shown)


core/Files.cc
```diff
@@ -346,11 +346,9 @@ bool File::permitOverloadDefinitions() const {
 }
 
 bool File::isPackage() const {
-    return flags.isPackage;
-}
-
-void File::setIsPackage(bool isPackage) {
-    this->flags.isPackage = isPackage;
+    // If the `__package.rb` file is at `typed: ignore`, then we haven't even parsed it.
+    // Any loop over "all package files" is really only wants "all non-ignored package files."
```
froydnj, line 350:
```suggestion
    // Any loop over "all package files" really only wants "all non-ignored package files."
```

froydnj (approved):

core/Files.cc
jez: Thanks, I will be sure to include this in the local branch I have stacked on this change.

PR #6702, opened 2023-02-03
Use cwd from config and expand workspaceFolder

paddyobrien:
### Motivation

Fixes https://github.com/sorbet/sorbet/issues/6701

There was a previous attempt at fixing this that floundered on variable expansion: https://github.com/sorbet/sorbet/pull/5529

I've done something fairly limited here and just replaced `${workspaceFolder}` with `workspace.rootPath`, this isn't ideal since it other variables won't work, but it has the upside of being backwards compatible.

There doesn't appear to be any API for doing this variable extension properly, at least according to:

https://stackoverflow.com/questions/43972354/how-can-an-extension-expand-configuration-variables-in-vscode

### Test plan

I'm unclear on how to add automated tests for this. I did build the extension and try it out locally. I can confirm it fixed the linked issue.

PR #7484, opened 2023-11-15
fix: use cwd from config and expand workspaceFolder

### LINKED ISSUE #6701: VSCode extension does not use cwd when spawning sorbet process
The VSCode extension allows for specifying a `cwd`, however this option is not used when spawning the process, rather it uses the workspace root:

https://github.com/sorbet/sorbet/blob/master/vscode_extension/src/LanguageClient.ts#L276

This makes it difficult to use the VSCode extension in repos where the Gemfile is not at the root of the project, for example a monorepo that puts a rails app in `/backend`

fazo96:
This PR contains the same changeset as https://github.com/sorbet/sorbet/pull/6702 from @paddyobrien and thus fixes #6701 

The credit for the actual changes goes to @paddyobrien, I just rebased them on master

### Motivation

The [original PR](https://github.com/sorbet/sorbet/pull/6702) didn't get merged and conflicts, and I needed a patched version of the extension to use on my project.

### Test plan

It works on our project, where the ruby project is in a subfolder of the workspace, by setting the `cwd` option correctly. The version published on the Microsoft Marketplace doesn't work

### DIFF
--- a/vscode_extension/src/languageClient.ts
+++ b/vscode_extension/src/languageClient.ts
@@ -284,8 +284,15 @@ export class SorbetLanguageClient implements Disposable, ErrorHandler {
     }
 
     this.context.log.debug(` > ${command} ${args.join(" ")}`);
+    const cwdTemplate = this.context.configuration.activeLspConfig!.cwd;
+    const cwd = cwdTemplate
+      ? cwdTemplate.replace(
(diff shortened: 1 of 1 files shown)


fazo96:
Unfortunately this patch has an issue where "Go To Definition" ends up with a file not found error in VSCode, I guess because the extension isn't sending the appropriate path relative to the workspace root

PR #7485, opened 2023-11-15
Use const GlobalState in PackageInfoFinder

jez:
### Commit summary

- **no-op: Store ImportType with 1 byte (not 4)** (fc81847b4)

- **Factor out populateMangledName helper** (b058e6090)

- **Use const GlobalState in PackageInfoFinder** (9b7342ccf)

  This defers the GlobalState mutations to run after the tree walk, which 
  lets the treewalk itself be used in places where we only have an 
  immutable GlobalState.

- **no-op: Assorted cleanups** (72cc99ff4)

### Motivation

This defers the GlobalState mutations to run after the tree walk, which 
lets the treewalk itself be used in places where we only have an immutable
GlobalState. I haven't actually made any changes which take advantage of this
yet--those changes will land in future changes.

### Test plan

Existing tests

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -963,16 +963,14 @@ struct PackageInfoFinder {
 
         if ((send.fun == core::Names::import() || send.fun == core::Names::testImport()) && send.numPosArgs() == 1) {
             // null indicates an invalid import.
-            if (auto target = verifyConstant(ctx, send.fun, send.getPosArg(0))) {
-                auto name = getPackageName(ctx, target);
-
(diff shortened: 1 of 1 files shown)


packager/packager.cc
```diff
@@ -88,16 +88,24 @@ struct PackageName {
     }
 };
 
-enum class ImportType {
+enum class ImportType : uint8_t {
```
jez, line 91: There's a handful of semi-related changes in this PR. Feel free to review by commit for a better breakdown.

neilparikh (approved):

PR #7495, opened 2023-11-17
Add sigs to `Gem::Version`

amomchilov:
### Motivation

To remove some `T.lets` from my code :)

### Test plan

None, these are just sig changes.

### DIFF
--- a/rbi/stdlib/rubygems.rbi
+++ b/rbi/stdlib/rubygems.rbi
@@ -2634,9 +2634,14 @@ class Gem::Version < Object
 
   sig do
     params(
-      version: T.any(String, Gem::Version),
-    )
-    .void
+      version: T.any(
+        String,
+        Integer,
+        Float, # Technically, any `to_s`able object would work.
+        Gem::Version, # Will return the same version object
+        NilClass # Depracated, and will issue a warning
+      )
+    ).void
   end
   def initialize(version); end
 
@@ -2644,17 +2649,21 @@ class Gem::Version < Object
 
   def _split_segments(); end
 
+  sig { returns(String) }
   def _version(); end
 
   # A recommended version for use with a ~> Requirement.
+  sig { returns(String) }
   def approximate_recommendation(); end
 
   # Return a new version object where the next to the last revision number is
   # one greater (e.g., 5.3.1 => 5.4).
   #
   # Pre-release (alpha) parts, e.g, 5.3.1.b.2 => 5.4, are ignored.
+  sig { returns(Gem::Version) }
   def bump(); end
 
+  sig { returns(T::Array[String]) }
   def canonical_segments(); end
 
   def encode_with(coder); end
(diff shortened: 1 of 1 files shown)


rbi/stdlib/rubygems.rbi
```diff
@@ -2634,27 +2634,36 @@ class Gem::Version < Object
 
   sig do
     params(
-      version: T.any(String, Gem::Version),
-    )
-    .void
+      version: T.any(
+        String,
+        Integer,
+        Float, # Technically, any `to_s`able object would work.
```
amomchilov, line 2640: E.g. `Gem::Version(1)` and `Gem::Version(1.23)` are allowed.

Does Sorbet have a way to express structural requirements about arguments? E.g. anything that responds to `#to_s`?

rbi/stdlib/rubygems.rbi
```diff
@@ -2634,27 +2634,36 @@ class Gem::Version < Object
...
+        Gem::Version, # Will return the same version object
+        NilClass # Depracated, and will issue a warning
+      )
+    ).void
   end
   def initialize(version); end
 
   def _segments(); end
 
   def _split_segments(); end
```
amomchilov, line 2650: I left some of these out because they're intended to be psuedo-private, and have a complicated shape like `[T:Array[Integer], T::Array[T.any(Integer, String)]]`

rbi/stdlib/rubygems.rbi
froydnj: Sorbet does not support something like `T::RespondsTo[:to_s]`, and I suspect never will.

https://sorbet.org/docs/flow-sensitive#what-about-respond_to

froydnj (changes requested):
It looks like your RBI changes have some errors.

amomchilov:
I think I fixed them! But I haven't been able to set up the test-suite locally. I'm relying on CI to test these for me 😅

rbi/stdlib/rubygems.rbi
```diff
@@ -2710,12 +2728,24 @@ class Gem::Version < Object
...
+    params(
+      version: T.any(
+        String,
+        Integer,
+        Float, # Technically, any `to_s`able object would work.
+        Gem::Version, # Will return the same version object
+        NilClass # Depracated, and will issue a warning
+      )
+    ).void
+  end
```
froydnj, line 2748: This looks like a correct sig for `initialize`, but not for `self.new`, which should return a value (I think this is `T.self_type`?).

froydnj (changes requested):

rbi/stdlib/rubygems.rbi
amomchilov: Correct, that explains the typecheck failure in CI, too. Fixed!

froydnj:

→ https://go/builds/bui_PC0IcGD4JwIA4l
→ https://go/builds/bui_PC0I31X4gIazEZ
→ https://go/builds/bui_PC0IO5FYehJtSS

rbi/stdlib/rubygems.rbi
```diff
@@ -2710,12 +2728,24 @@ class Gem::Version < Object
...
+  sig do
+    params(
+      version: T.any(
+        String,
+        Integer,
+        Float, # Technically, any `to_s`able object would work.
+        Gem::Version, # Will return the same version object
+        NilClass # Depracated, and will issue a warning
+      )
+    ).returns(T.self_type)
```
froydnj, line 2747:
```suggestion
    ).returns(Gem::Version)
```

I was dumb to make the `T.self_type` recommendation.  Let's try this.

froydnj:

→ https://go/builds/bui_PC15W0nNayjMsA
→ https://go/builds/bui_PC15PE4oh11xDX
→ https://go/builds/bui_PC150pZkPKxOgs

froydnj (approved):
Thanks!

Note to Stripes: this PR needs a small `T.must` added, which I will do when bumping Sorbet.

PR #7507, opened 2023-11-28
Update a location for `requires_ancestor` on a fast path

ilyailya:
### Test plan

See included automated tests.


core/Symbols.cc
```diff
@@ -1999,6 +1999,40 @@ void ClassOrModule::recordRequiredAncestorInternal(GlobalState &gs, ClassOrModul
...
+        // class A
+        //   requires_ancestor { B }
+        //   requires_ancestor { C }
+        // end
+        // ```
+        //
+        // Class `A` would have one symbol for the `requires_ancestor` method, but two identical locations for that
+        // symbol in the same file.
+        //
+        // Usually Sorbet doesn't permit more than one location in a signle file.
```
Morriar, line 2018:
```suggestion
        // Usually Sorbet doesn't permit more than one location in a single file.
```

PR #7513, opened 2023-12-04
complain if `rubyfmt` cannot be found for LSP

froydnj:
### Motivation

Request from a user for debugging help.

Using `error` here without exiting feels a little weird, but there is precedent for this later in `options.cc`.

### Test plan

See included automated tests.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -729,9 +729,13 @@ void readOptions(Options &opts,
             enableAllLSPFeatures || raw["enable-experimental-lsp-document-highlight"].as<bool>();
         opts.lspSignatureHelpEnabled = enableAllLSPFeatures || raw["enable-experimental-lsp-signature-help"].as<bool>();
         opts.rubyfmtPath = raw["rubyfmt-path"].as<string>();
-        opts.lspDocumentFormatRubyfmtEnabled =
-            FileOps::exists(opts.rubyfmtPath) &&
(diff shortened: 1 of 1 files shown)


main/options/options.cc
```diff
@@ -729,9 +729,13 @@ void readOptions(Options &opts,
...
-        opts.lspDocumentFormatRubyfmtEnabled =
-            FileOps::exists(opts.rubyfmtPath) &&
-            (enableAllLSPFeatures || raw["enable-experimental-lsp-document-formatting-rubyfmt"].as<bool>());
+        if (enableAllLSPFeatures || raw["enable-experimental-lsp-document-formatting-rubyfmt"].as<bool>()) {
+            if (!FileOps::exists(opts.rubyfmtPath)) {
+                logger->error("`{}` does not exist, LSP rubyfmt integration will not be enabled", opts.rubyfmtPath);
+            } else {
+                opts.lspDocumentFormatRubyfmtEnabled = true;
+            }
+        }
```
jez, line 738: It looks like this is also a logic bug—the default value of `rubyfmtPath` is `"rubyfmt"`. If you've correctly added that binary to your `PATH`, Sorbet should work. but right now it's checking if the file exists, e.g. as a relative path in the current directory.

The rubyfmtPath is eventually passed to `posix_spawnp`, which according to the man page suggests that this **will** do `PATH`-based lookup.

jez (approved):
I suppose this is strictly better than what it does right now, but I think this points to a logic bug that we will want to fix anyways.

PR #7516, opened 2023-12-04
SorbetExtensionConfig uses only workspace configurations 

damolina-stripe:
### Test plan

See included automated tests.


damolina-stripe:
This fix is more correct but requires too much work at the time.

PR #7527, opened 2023-12-08
Assorted prework changes for packager refactor

jez:
### Motivation

I've started to accumulate a big log of prework changes as I work through the
packager refactor. Figure I may as well land some of them.

### Test plan

no-op changes; existing tests

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -909,15 +909,6 @@ struct PackageInfoFinder {
     unique_ptr<PackageInfoImpl> info = nullptr;
     vector<Export> exported;
 
-    void postTransformCast(core::Context ctx, const ast::ExpressionPtr &tree) {
-        auto &cast = ast::cast_tree_nonnull<ast::Cast>(tree);
-        if (!ast::isa_tree<ast::Literal>(cast.typeExpr)) {
-            if (auto e = ctx.beginError(cast.typeExpr.loc(), core::errors::Packager::InvalidPackageExpression)) {
-                e.setHeader("Invalid expression in package: Arguments to functions must be literals");
-            }
-        }
-    }
-
     void postTransformSend(core::Context ctx, ast::ExpressionPtr &tree) {
         auto &send = ast::cast_tree_nonnull<ast::Send>(tree);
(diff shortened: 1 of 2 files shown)


packager/packager.cc
```diff
@@ -1490,9 +1437,6 @@ void Packager::run(core::GlobalState &gs, WorkerPool &workers, absl::Span<ast::P
     findPackages(gs, files);
     setPackageNameOnFiles(gs, files);
 
-    // Step 2:
-    // * Find package files and rewrite them into virtual AST mappings.
-    // * Find files within each package and rewrite each to be wrapped by their virtual package namespace.
```
jez, line 1495: I got rid of the step numbers because it says step 1a came after step 1, and step 2 comes after both step 1 and step 1a—it doesn't seem useful to number these anymore.

I got rid of these comments, because they talk about doing rewrites that we no longer do (back when we modeled package visibility as a series of AST rewrites)

packager/packager.cc
```diff
@@ -1395,9 +1347,10 @@ void validatePackagedFile(core::Context ctx, const ast::ExpressionPtr &tree) {
 
 void Packager::findPackages(core::GlobalState &gs, absl::Span<ast::ParsedFile> files) {
     // Ensure files are in canonical order.
+    // TODO(jez) Is this sort redundant? Should we move this sort to callers?
```
jez, line 1350: I plan to look at this some other time.

test/testdata/packager/invalid_package_control_flow/__package.rb
```diff
@@ -3,7 +3,7 @@
 # enable-packager: true
 
 # Constant definitions/assignments are not OK
-SomeConstant = PackageSpec # error: Invalid expression in package: `=`
+SomeConstant = PackageSpec # error: Invalid expression in package: `Assign`
```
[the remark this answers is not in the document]
jez: @froydnj happy to restructure this if you want, we could make a switch statement in this file, or add another ExpressionPtr method like "humanReadableNodeName"

PR #7529, opened 2023-12-09
Split packager into definePackage and rewritePackageSpec

jez:
Mostly reviewable by commit.

### Motivation

Splitting things this way is a stepping stone to being able to run namer on
`__package.rb` files to define `Package` symbols, and then running the packager
to discover imports and exports.

### Test plan

Close to no behavior changes, so covered by existing tests. Some of the errors
double up, because it's easier to structure things that way, and two errors for
one problem is mostly fine.

### DIFF
--- a/core/errors/packager.h
+++ b/core/errors/packager.h
@@ -18,7 +18,7 @@ constexpr ErrorClass InvalidPackageExpression{3710, StrictLevel::False};
 constexpr ErrorClass PackageFileMustBeStrict{3711, StrictLevel::False};
 constexpr ErrorClass InvalidPackageName{3712, StrictLevel::False};
 constexpr ErrorClass DefinitionPackageMismatch{3713, StrictLevel::False};
-constexpr ErrorClass ImportConflict{3714, StrictLevel::False};
+// constexpr ErrorClass ImportConflict{3714, StrictLevel::False};
 // constexpr ErrorClass InvalidExportForTest{3715, StrictLevel::False};
 constexpr ErrorClass ExportConflict{3716, StrictLevel::False};
 constexpr ErrorClass UsedPackagePrivateName{3717, StrictLevel::False};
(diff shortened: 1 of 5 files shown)


core/errors/packager.h
```diff
@@ -18,7 +18,7 @@ constexpr ErrorClass InvalidPackageExpression{3710, StrictLevel::False};
 constexpr ErrorClass PackageFileMustBeStrict{3711, StrictLevel::False};
 constexpr ErrorClass InvalidPackageName{3712, StrictLevel::False};
 constexpr ErrorClass DefinitionPackageMismatch{3713, StrictLevel::False};
-constexpr ErrorClass ImportConflict{3714, StrictLevel::False};
+// constexpr ErrorClass ImportConflict{3714, StrictLevel::False};
```
jez, line 21: This was not getting used.

packager/packager.cc
```diff
@@ -431,8 +431,8 @@ PackageName getPackageName(core::Context ctx, const ast::UnresolvedConstantLit *
     return pName;
 }
 
+// TODO(jez) Rename this to lookupMangledName, and make it take a const GlobalState
```
jez, line 434: Planning to move this to a future commit.

packager/packager.cc
```diff
@@ -1176,102 +1162,170 @@ struct PackageInfoFinder {
     }
 };
 
-unique_ptr<PackageInfoImpl> rewritePackageSpec(const core::GlobalState &gs, ast::ParsedFile &package) {
+unique_ptr<PackageInfoImpl> definePackage(const core::GlobalState &gs, ast::ParsedFile &package) {
```
jez, line 1165: This function is going to be the skeleton of the portion of this code that will move into namer.

ilyailya (approved):

jez:
@froydnj just want one more set of eyes on this just in case

froydnj (approved):

PR #7531, opened 2023-12-12
Speed up method patching with identity Hashes

amomchilov:
### Motivation

The `@sigs_that_raised` and `@was_ever_final_names` hashes are keyed by types for which `#==` is the same as `Object#equal?` (`Symbol`s and `Module`s). We can speed these up by using [`Hash#compare_by_identity`](https://ruby-doc.org/3.2.2/Hash.html#method-i-compare_by_identity).

Identity hashes are about 1.3x faster at the moment (in Ruby 3.2.1):

```
Warming up --------------------------------------
 regular Hash            1.520M i/100ms
 compare_by_identity     1.948M i/100ms
Calculating -------------------------------------
 regular Hash            15.122M (± 1.4%) i/s -     76.002M in   5.026956s
 compare_by_identity     19.482M (± 1.2%) i/s -     97.422M in   5.001500s

Comparison:
 compare_by_identity: 19481691.7 i/s
 regular Hash       : 15121932.4 i/s - 1.29x  slower
```

<details><summary><code>benchmark.rb</code></summary>

```ruby
#!/usr/bin/ruby

require 'benchmark/ips'

k = Module.new
hash = { k => true }
hash_by_id = { k => true }.compare_by_identity

Benchmark.ips do |x|
  x.report("regular Hash       ") { hash[k.object_id] }
  x.report("compare_by_identity") { hash_by_id[k]     }
  x.compare!
end
```

</details> 

### Test plan

See included automated tests.

PR #7533, opened 2023-12-12
Remove excess `is_singleton_method` param

amomchilov:
### Motivation

There is two occurences of a `is_singleton_method ? mod.singleton_class : mod` check:
    1. The one in `add_module_with_final_method` can just be removed, if the function is called with correct module directly.
    2. The one in `_on_method_added` can be removed, because the two hooks could just pass in differing modules.

### Test plan

This is just a refactor, that's covered by existing tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -191,12 +191,11 @@ module T::Private::Methods
     end
   end
 
-  def self.add_module_with_final_method(mod, method_name, is_singleton_method)
-    m = is_singleton_method ? mod.singleton_class : mod
-    methods = @modules_with_final[m]
+  def self.add_module_with_final_method(mod, method_name)
+    methods = @modules_with_final[mod]
     if methods.nil?
       methods = {}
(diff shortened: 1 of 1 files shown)


froydnj (changes requested):
It looks like Rubocop is unhappy with your changes...though honestly, I think Rubocop is wrong in this particular situation, and would be happy to take an override.

amomchilov:
@froydnj Eh, I'll play ball.

The visual distinction of `self` vs `self.singleton` between `#method_added` and `#singleton_method_added` is nice, but not worth the ugly rubocop exception comment.

Such is the tyranny of the bots 😅

froydnj (changes requested):
It looks like rebasing is necessary?

amomchilov:
Interesting, not sure why, as this was just stacked over https://github.com/sorbet/sorbet/pull/7531. Maybe I just made a change there that I didn't reflect here.

Rebased :)

amomchilov:
Hey @froydnj, any more changes needed here?

froydnj (approved):
Sorry, I think something must have gone wrong the first time I tried to test this.  Everything passed the second time.  Thanks for your patience!

PR #7534, opened 2023-12-13
Make type initialization happen lazily

Morriar:
### Motivation

For performance reasons, in production, we remove as much as `sorbet-runtime` overhead as possible. One of the changes we apply is no-op'ing `T` annotations such as `T.let(x, Type)` by monkey patching the `T` module:

```ruby
module InlineTypeAssertions
  def let(value, type, checked: true)
    value
  end
end

T.singleton_class.prepend(InlineTypeAssertions)
```

This effectively negates the overhead of the inline assertion `let` but the cost of type `Type` construction is still present.

Here's a very simple example using the patch above:

```rb
100_000.times do
  x = T.let([], T::Array[T::Hash[Integer, String]])
end
```

Profiling this snippet with our `InlineTypeAssertions` patch shows that while `T.let` is not doing anything, we still spend a considerable amount of time into initializing `Type` mostly due to `T::Utils.coerce`:

image

This PR proposes to move the type coercion outside of the initializer to let it happen lazily. In a normal execution context, this doesn't modify the behavior at all. The type construction is just moved to the first call involving the type rather than when the type is instantiated. But in a context where inline type assertions are no-op'ed, this drastically simplifies the type initialization and since nothing ever calls inside the type in that case, the lazy construction is never executed, resulting in a considerable speed-up and less pressure on the garbage collector. 

image

There is no extra performance cost for this change but we require an additional instance variable for each type to store the original type passed to the constructor before we lazily coerce it.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/types/base.rb
+++ b/gems/sorbet-runtime/lib/types/types/base.rb
@@ -33,7 +33,8 @@ module T::Types
       raise NotImplementedError
     end
 
-    # Force the lazy type initialization
+    # Force any lazy initialization that this type might need to do
+    # It's unusual to call this directly; you probably want to call it indirectly via `T::Utils.run_all_sig_blocks`.
     def build_type
       raise NotImplementedError
     end
--- a/gems/sorbet-runtime/lib/types/types/base.rb
+++ b/gems/sorbet-runtime/lib/types/types/base.rb
@@ -33,6 +33,11 @@ module T::Types
       raise NotImplementedError
     end
 
+    # Force the lazy type initialization
+    def build_type
+      raise NotImplementedError
+    end
+
     # Equality is based on name, so be sure the name reflects all relevant state when implementing.
     def name
       raise NotImplementedError
--- a/gems/sorbet-runtime/lib/types/types/fixed_array.rb
+++ b/gems/sorbet-runtime/lib/types/types/fixed_array.rb
@@ -6,23 +6,25 @@ module T::Types
   # Takes a list of types. Validates each item in an array using the type in the same position
   # in the list.
   class FixedArray < Base
-    attr_reader :types
-
     def initialize(types)
-      @types = types.map {|type| T::Utils.coerce(type)}
+      @inner_types = types
+    end
+
+    def types
+      @types ||= @inner_types.map {|type| T::Utils.coerce(type)}
     end
 
     # overrides Base
     def name
-      "[#{@types.join(', ')}]"
+      "[#{types.join(', ')}]"
     end
 
     # overrides Base
     def recursively_valid?(obj)
-      if obj.is_a?(Array) && obj.length == @types.length
+      if obj.is_a?(Array) && obj.length == types.length
         i = 0
-        while i < @types.length
-          if !@types[i].recursively_valid?(obj[i])
+        while i < types.length
+          if !types[i].recursively_valid?(obj[i])
             return false
           end
           i += 1
@@ -35,10 +37,10 @@ module T::Types
 
     # overrides Base
     def valid?(obj)
-      if obj.is_a?(Array) && obj.length == @types.length
+      if obj.is_a?(Array) && obj.length == types.length
         i = 0
-        while i < @types.length
-          if !@types[i].valid?(obj[i])
+        while i < types.length
+          if !types[i].valid?(obj[i])
             return false
           end
           i += 1
@@ -56,7 +58,7 @@ module T::Types
         # Properly speaking, covariance here is unsound since arrays
         # can be mutated, but sorbet implements covariant tuples for
         # ease of adoption.
-        @types.size == other.types.size && @types.zip(other.types).all? do |t1, t2|
+        types.size == other.types.size && types.zip(other.types).all? do |t1, t2|
           t1.subtype_of?(t2)
         end
       when TypedArray
@@ -85,7 +87,7 @@ module T::Types
     # overrides Base
     def describe_obj(obj)
       if obj.is_a?(Array)
-        if obj.length == @types.length
+        if obj.length == types.length
           item_classes = obj.map(&:class).join(', ')
           "type [#{item_classes}]"
         else
--- a/gems/sorbet-runtime/lib/types/types/fixed_array.rb
+++ b/gems/sorbet-runtime/lib/types/types/fixed_array.rb
@@ -14,6 +14,10 @@ module T::Types
       @types ||= @inner_types.map {|type| T::Utils.coerce(type)}
     end
 
+    def build_type
+      types
+    end
+
     # overrides Base
     def name
       "[#{types.join(', ')}]"
--- a/gems/sorbet-runtime/lib/types/types/fixed_array.rb
+++ b/gems/sorbet-runtime/lib/types/types/fixed_array.rb
@@ -16,6 +16,7 @@ module T::Types
 
     def build_type
       types
+      nil
     end
 
     # overrides Base
--- a/gems/sorbet-runtime/lib/types/types/typed_array.rb
+++ b/gems/sorbet-runtime/lib/types/types/typed_array.rb
@@ -61,7 +61,7 @@ module T::Types
       end
 
       def freeze
-        type # force lazy initialization before freezing the object
+        build_type # force lazy initialization before freezing the object
         super
       end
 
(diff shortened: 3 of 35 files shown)


jez:
If I understand correctly, for codebases that force all sig blocks eagerly in production during load time, this will mean that many sig-related allocations move from load time to the first request. Does that sound right to you?

Is it possible to build in a mode where the force all sig blocks utility function also forces the evaluation of this lazy computation?

Morriar:
Indeed we are deferring type evaluation from signature evaluation time to the first time the signature is checked. In our case, we eagerly force the signatures evaluation at boot time but also disable any check against the signatures so we never get to the lazy part.

But you're right, we wouldn't want that to be the default behavior for everybody using `T::Utils.run_all_sig_blocks`.

I added the ability to turn on and off the type construction when calling `run_all_sig_blocks` with a default to force the type construction eagerly. This brings back the type construction in signatures to boot time and still allow lazy construction for inline annotations.

gems/sorbet-runtime/lib/types/types/fixed_array.rb
```diff
@@ -6,23 +6,29 @@ module T::Types
...
     def initialize(types)
-      @types = types.map {|type| T::Utils.coerce(type)}
+      @inner_types = types
+    end
+
+    def types
+      @types ||= @inner_types.map {|type| T::Utils.coerce(type)}
+    end
+
+    def build_type
```
jez, line 17: What do you think about explicitly returning `nil` in these methods so we don't accidentally expose a contract that we can't change in the future?

gems/sorbet-runtime/lib/types/types/base.rb
```diff
@@ -33,6 +33,11 @@ def valid?(obj)
       raise NotImplementedError
     end
 
+    # Force the lazy type initialization
```
jez, line 36:
```suggestion
    # Force any lazy initialization that this type might need to do
    # It's unusual to call this directly; you probably want to call it indirectly via run_all_sig_blocks
```

gems/sorbet-runtime/lib/types/types/fixed_array.rb
Morriar: Good point, done 👍

gems/sorbet-runtime/lib/types/types/typed_array.rb
```diff
@@ -60,6 +60,11 @@ def valid?(obj)
         obj.is_a?(Array)
       end
 
+      def freeze
+        type # force lazy initialization before freezing the object
```
paracycle, line 64: Should this be calling `build_type` instead?

gems/sorbet-runtime/lib/types/types/typed_class.rb
```diff
@@ -67,6 +74,11 @@ def initialize
         super(T.untyped)
       end
 
+      def freeze
+        type # force lazy initialization before freezing the object
```
paracycle, line 78: Should this be calling `build_type` instead?

gems/sorbet-runtime/lib/types/types/typed_class.rb
```diff
@@ -77,6 +89,11 @@ def initialize
         super(T.anything)
       end
 
+      def freeze
+        type # force lazy initialization before freezing the object
```
paracycle, line 93: Should this be calling `build_type` instead?

gems/sorbet-runtime/lib/types/types/typed_hash.rb
```diff
@@ -3,22 +3,32 @@
 
 module T::Types
   class TypedHash < TypedEnumerable
```
paracycle, line 5: I think this type might need to override `build_type` and trigger `keys` and `values` before calling `super`.

gems/sorbet-runtime/lib/types/types/typed_hash.rb
```diff
@@ -3,22 +3,32 @@
...
 
     def initialize(keys:, values:)
-      @keys = T::Utils.coerce(keys)
-      @values = T::Utils.coerce(values)
-      @type = T::Utils.coerce([keys, values])
+      @inner_keys = keys
+      @inner_values = values
+    end
+
+    # Technically we don't need these, but they is a nice api
```
paracycle, line 15:
```suggestion
    # Technically we don't need these, but they are a nice api
```

gems/sorbet-runtime/lib/types/types/typed_hash.rb
```diff
@@ -3,22 +3,32 @@
...
+    def keys
+      @keys ||= T::Utils.coerce(@inner_keys)
+    end
+
+    # Technically we don't need these, but they is a nice api
```
paracycle, line 20:
```suggestion
    # Technically we don't need these, but they are a nice api
```

gems/sorbet-runtime/lib/types/types/typed_array.rb
Morriar: Indeed, it would be cleaner even if equivalent 👍

gems/sorbet-runtime/lib/types/types/typed_hash.rb
Morriar: I actually changed the comment for using the singular form since I copied it other two separate methods 👍

gems/sorbet-runtime/lib/types/types/typed_hash.rb
Morriar: I actually changed the comment for using the singular form since I copied it other two separate methods 👍

gems/sorbet-runtime/lib/types/types/typed_hash.rb
Morriar: Yeah indeed, I don't think the order is relevant though?

gems/sorbet-runtime/lib/types/types/typed_hash.rb
```diff
@@ -26,6 +26,12 @@ def type
       @type ||= T::Utils.coerce([@inner_keys, @inner_values])
     end
 
+    def build_type
+      super
+      keys
+      values
```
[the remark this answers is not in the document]
Morriar: Right, if only we had a type checker for those things... 😂

jez:

→ https://go/builds/bui_PDXRtz43cYTpMv
→ https://go/builds/bui_PDXRvgN1oYzucf
→ https://go/builds/bui_PDXRvMfCIREXHq

gems/sorbet-runtime/lib/types/types/typed_hash.rb
```diff
@@ -3,22 +3,38 @@
...
+      @keys ||= T::Utils.coerce(@inner_keys)
+    end
+
+    # Technically we don't need this, but it is a nice api
+    def values
+      @values ||= T::Utils.coerce(@inner_values)
+    end
+
+    def type
+      @type ||= T::Utils.coerce([@inner_keys, @inner_values])
```
jez, line 26: Not blocking: I wonder if it's worth using `@keys, @values` here instead, because then the coerce will not have to do the work of coercing `@inner_keys` and `@inner_values` twice.

jez (approved):
Looks good, thanks for the idea and implementation!

Testing on Stripe's codebase right now, but looks good so far.

jez:
I'm getting a NotImplementedError coming from `lib/types/types/base.rb:39`.

Can you double check that you've implemented everything and try again?

jez (changes requested):
^

gems/sorbet-runtime/lib/types/types/typed_hash.rb
Morriar: Good point: cda3bbd 👍

Morriar: (I assumed you meant the `keys` and `values` methods rather than accessors)

jez (approved):

jez:
Passes CI at Stripe, going to merge when Sorbet's CI passes.

gems/sorbet-runtime/lib/types/types/typed_hash.rb
```diff
@@ -23,7 +23,7 @@ def values
     end
 
     def type
-      @type ||= T::Utils.coerce([@inner_keys, @inner_values])
+      @type ||= T::Utils.coerce([keys, values])
     end
 
     def build_type
```
paracycle, line 29: I just want to note that this method definition is no longer needed, since the `super` call will build `type`, which, with the new code structure, will trigger a build of `keys` and `values` anyway.

PR #7535, opened 2023-12-13
Allow hover in `typed: false` files if possible

jez:
cc @vinistock

### Motivation

This matches the behavior of go to definition, which has worked that way
since 99f0ffe9a559f75072dabe3c138ca64ebe8165f5.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/hover.cc
+++ b/main/lsp/requests/hover.cc
@@ -42,7 +42,7 @@ unique_ptr<ResponseMessage> HoverTask::runRequest(LSPTypecheckerDelegate &typech
 
     const core::GlobalState &gs = typechecker.state();
     auto result = LSPQuery::byLoc(config, typechecker, params->textDocument->uri, *params->position,
-                                  LSPMethod::TextDocumentHover);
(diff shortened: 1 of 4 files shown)


main/lsp/requests/hover.cc
```diff
@@ -42,7 +42,7 @@ unique_ptr<ResponseMessage> HoverTask::runRequest(LSPTypecheckerDelegate &typech
 
     const core::GlobalState &gs = typechecker.state();
     auto result = LSPQuery::byLoc(config, typechecker, params->textDocument->uri, *params->position,
-                                  LSPMethod::TextDocumentHover);
+                                  LSPMethod::TextDocumentHover, false);
```
jez, line 45: false is `errorIfFileIsUntyped` which controls whether `byLoc` will return results or intentionally suppress them.

ilyailya (approved):

main/lsp/requests/hover.cc
jez: at this point we may want to get rid of that, because i think that the vast majority and maybe all calls pass `false` at this point.

PR #7536, opened 2023-12-14
Fix `new` in subclasses of abstract classes when `tap` is redeclared

Bo98:
### Motivation
I have an abstract class that happened to redeclare the `tap` method to be something else entirely. This used to work until 544f173782921ded834d925c5b86084326be244d. Sorbet seems to generally work around tampering like this (see `Object.instance_method(:method)` a few lines prior) so I reckon the same should happen here.

I could have done:

```rb
Object.instance_method(:tap).bind_call(super(*args, &blk)) do |result|
  if result.instance_of?(mod)
    raise "#{mod} is declared as abstract; it cannot be instantiated"
  end
end
```

but the code-readability benefits from `tap` are largely lost so it's IMO clearer to just avoid it entirely.

### Test plan
See included test.


jez (approved):
Thanks!

PR #7537, opened 2023-12-14
Add missing `Gem::Security::Exception` type

Bo98:
### Motivation
`Gem::Secutity::Exception`, which could be thrown from some RubyGems methods (e.g. `Gem::Package` API), was missing.

### Test plan
Documented: https://docs.ruby-lang.org/en/3.2/Gem/Security/Exception.html
Validated: https://github.com/rubygems/rubygems/blob/599fa578c55ae1a94cb6225bdddf71e212201cfc/lib/rubygems/security.rb#L332
Tested: in my own project for a little while.


jez (approved):
Thanks!

PR #7538, opened 2023-12-14
Allow completion on typed: false files for constants

vinistock:
Allow completion results in `typed: false` files so that we can complete constants, which can be properly typechecked by Sorbet.

### Motivation

Similar to https://github.com/sorbet/sorbet/pull/7535. Sorbet can accurately find constants even in `typed: false` files, so enabling completion improves the developer experience.

### Test plan

Updated existing tests and added one for ignored files.

### DIFF
--- a/main/lsp/requests/completion.cc
+++ b/main/lsp/requests/completion.cc
@@ -1239,7 +1239,7 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerDelegate &t
     }
     auto queryLoc = maybeQueryLoc.value();
 
-    auto result = LSPQuery::byLoc(config, typechecker, uri, pos, LSPMethod::TextDocumentCompletion);
+    auto result = LSPQuery::byLoc(config, typechecker, uri, pos, LSPMethod::TextDocumentCompletion, false);
 
     if (result.error) {
         // An error happened while setting up the query.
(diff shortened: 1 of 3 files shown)


main/lsp/requests/completion.cc
```diff
@@ -1265,8 +1265,11 @@ unique_ptr<ResponseMessage> CompletionTask::runRequest(LSPTypecheckerDelegate &t
             ENFORCE(fref.exists());
             auto level = fref.data(gs).strictLevel;
             if (!fref.data(gs).hasParseErrors() && level < core::StrictLevel::True) {
-                items.emplace_back(
-                    getCompletionItemForUntyped(gs, queryLoc, 0, "(file is not `# typed: true` or higher)"));
+                auto text = level == core::StrictLevel::Ignore
+                                ? "This file is `# typed: ignore`. No Sorbet IDE features will work in this file."
+                                : "This file is `# typed: false`. Most Completion results will not appear until the "
+                                  "file is `# typed: true` or higher.";
```
jez, line 1271: This is an intrusively long label to show up every time there's no completion result.

Let's keep it as short as it used to be, and I like the enclosing parens.

jez (changes requested):
^

jez (approved):

PR #7539, opened 2023-12-14
delete `tools/bazel`

froydnj:
### Motivation

Symbolic links in git repositories are just weird, CI is green without this, and people should be using `./bazel` instead.

### Test plan

See included automated tests.


froydnj:

→ https://go/builds/bui_PBjXbFJntblEER
→ https://go/builds/bui_PBjXOGrqkKv3sK
→ https://go/builds/bui_PBjXHSxRny49Pc

jez (approved):
We might want to bring this back in some form, but that can be a future change. This unblocks our CI in the short term.

jez:
Summary of conversation in Slack:

> > > We’re  seeing this weird behavior when trying to depend on on the sorbet repo via bazel, where the presence of a `tools/bazel` symlink [causes problems](http://go/builds/bui_PBds766H2fYjKG).
> > >
> > > We’ve confirmed that it’s the symlink because [this PR](https://github.com/sorbet/sorbet/pull/7539) which does nothing other than delete the symlink fixes things.
> > >
> > > The problem is that this symlink is the `tools/bazel` symlink, which IIUC is somewhat special to bazel and/or versioning. If I understand correctly, deleting this symlink will mean that if someone has `bazel` installed system wide and tries to run `bazel build` in the Sorbet repo (instead of `./bazel build`), then it might use the wrong version of bazel. Is that correct?
> > >
> > > And then another question: what’s the best way to deal with this? I don’t want to have to start fielding an influx of questions from the open source people trying to build Sorbet that boils down to “you’re using bazel instead of ./bazel” if possible.
> >
> > when you run bazel, it looks for a WORKSPACE file in the ancestors of the cwd, then looks for a file called `bazel` in the workspace’s `tools` dir. if it exists, it just `exec`s that with the same args it was passed
>
> gotcha. so another option here could be
>
> - Delete tools/bazel symlink
> - Move ./bazel to tools/bazel
> - Require everyone to have bazel on the path
> - optional: add ./bazel to the .gitignore and let people make a symlink from ./bazel -> tools/bazel in their local copy

ilyailya (approved):

PR #7543, opened 2023-12-19
Add native_thread_id method to Thread

stug:
Adding a missing definition for `native_thread_id`: https://docs.ruby-lang.org/en/3.2/Thread.html#method-i-native_thread_id

I tried to put it in alphabetical order with the rest of the methods, and I copied [the comment from thread.c in the MRI source](https://git.ruby-lang.org/ruby.git/tree/thread.c?h=ruby_3_2#n3301) since that's where it seemed like the other comments came from.

Let me know if anything looks amiss!

### DIFF
--- a/rbi/core/thread.rbi
+++ b/rbi/core/thread.rbi
@@ -495,6 +495,25 @@ class Thread < Object
   sig {params(name: T.untyped).returns(T.untyped)}
   def name=(name); end
 
+  # Return the native thread ID which is used by the Ruby thread.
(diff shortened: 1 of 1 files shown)


jez (approved):

jez:
> it seemed like the other comments came from

There is also `rbi/tools/sync-rdoc.rb <file>` which fetches the latest documentation for all the methods in a file. Copying manually is fine too.

PR #4063, opened 2021-03-07
Fix some RBIs that actually take blocks

jez:
### Motivation

Pre-work for #4061

### Test plan

See included automated tests.

PR #7550, opened 2023-12-24
Tighten signature of Enumerable#inject

marianosimone:
### Motivation
`inject` is an alias for `reduce`, so they can have the same signature

### Test plan
Added tests, and expanded existing ones (to cover the case without an initial value)

### DIFF
--- a/test/testdata/infer/takes_no_block_inject.rb
+++ b/test/testdata/infer/takes_no_block_inject.rb
@@ -1,4 +1,0 @@
-# typed: strict
-
-foo = T::Array[Integer].new
-foo.inject(nil) {|sum, x| sum ? sum + x : x}
--- a/rbi/core/enumerable.rbi
+++ b/rbi/core/enumerable.rbi
@@ -855,17 +855,17 @@ module Enumerable
     .returns(T.untyped)
   end
   sig do
-    params(
-        initial: T.untyped,
-        blk: T.proc.params(arg0: T.untyped, arg1: Elem).returns(T.untyped),
+    type_parameters(:U).params(
+        initial: T.type_parameter(:U),
+        blk: T.proc.params(arg0: T.type_parameter(:U), arg1: Elem).returns(T.type_parameter(:U)),
     )
-    .returns(T.untyped)
+    .returns(T.type_parameter(:U))
   end
   sig do
-    params(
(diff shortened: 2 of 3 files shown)


test/testdata/infer/takes_no_block_inject.rb
```diff
@@ -1,4 +0,0 @@
-# typed: strict
-
-foo = T::Array[Integer].new
-foo.inject(nil) {|sum, x| sum ? sum + x : x}
```
marianosimone, line 4: This was added by @jez on [[sorbet/pr/4063] Fix some RBIs that actually take blocks](https://github.com/sorbet/sorbet/pull/4063/)

I'm not entirely sure we do want it, though... now that the signature is tighter, the type error that tells you that `sum` is always `nil` initially will signal the author that they don't need to do this... It would make more sense to just write `foo.inject {|sum, x| sum ? sum + x : x}` in a case like this 🤷🏽 

Note that we didn't have the same test for `reduce` any way, and those are aliases

marianosimone:

→ https://go/builds/bui_PFteRYRBOfCm3w
→ https://go/builds/bui_PFtexhjxmdPA6E
→ https://go/builds/bui_PFteW5Q5PSzVqg

jez:
I had originally punted on this change in https://github.com/sorbet/sorbet/pull/6087#issuecomment-1194736387 because it caused so many problems on Stripe's codebase. And that's not to say they weren't fixable, but I was worried about how annoying it would feel to have to explicitly widen the type of the accumulator in a call to `inject`.

If you want to fix this in Stripe's codebase, you could replace all the uses of inject with reduce.

Otherwise, I'd want to think about whether there's something we could do to do a better job inside Sorbet's support for generic methods to support this idiom without so much verbosity.

test/testdata/infer/takes_no_block_inject.rb
jez: The type error that Sorbet produces on this snippet is the biggest source of errors in Stripe's codebase, and the reason why I hadn't gone forward with this approach in my previous attempt.

> in most cases, too!

The history here is that Stripe's codebase had an `Array#sum` monkeypatch which clobbered the built-in `Array#sum` method, and that caused problems so we backed it out by inlining a bunch of these `.inject` calls all over the codebase.

PR #7562, opened 2024-01-02
Allow ConstantAssumeType rewriter to work with class aliases

### LINKED ISSUE #7560: Fix or improve error for when ConstAssumeType rewriter fails
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20A%3B%20end%0AAliasToX%20%3D%20A%0A%0AX%20%3D%20AliasToX.new)

```ruby
# typed: strict

class A; end
AliasToX = A

X = AliasToX.new
```

#### Observed output

```
editor.rb:6: Constants must have type annotations with T.let when specifying # typed: strict https://srb.help/7027
     6 |X = AliasToX.new
            ^^^^^^^^^^^^
  Autocorrect: Use -a to autocorrect
    editor.rb:6: Replace with T.let(AliasToX.new, A)
     6 |X = AliasToX.new
            ^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

Either this shouldn't be an error, because Sorbet should be able to figure out the type from the `.new` call, or Sorbet should indicate that it tried to assume something about the type but the assumption was wrong.

- - -

jez:
### Motivation

You can put class aliases in normal type syntax (as evidenced by the fact that
this `<cast:<assume type>>(...)` node works when made to simply not early exit).

Fixes #7560

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -2989,8 +2989,15 @@ public:
             // make sure that it's even possible to be valid.
             auto *cnst = ast::cast_tree<ast::ConstantLit>(cast->typeExpr);
             ENFORCE(cnst != nullptr, "Rewriter should always use const for typeExpr, which should now be resolved");
-            if (!cnst->symbol.isClassOrModule() || cnst->symbol.asClassOrModuleRef().data(ctx)->flags.isModule ||
-                cnst->symbol.asClassOrModuleRef().data(ctx)->typeArity(ctx) > 0) {
+
+            // Dealias to mimic how type_syntax parsing will work for constant lit node (e.g., want
+            // to allow class aliases).
+            auto dealiased = cnst->symbol.dealias(ctx);
+
+            // Don't want the dealias to apply to the type alias, because MyTypeAlias.new is a runtime error.
+            if (cnst->symbol.isTypeAlias(ctx) || !dealiased.isClassOrModule() ||
+                dealiased.asClassOrModuleRef().data(ctx)->flags.isModule ||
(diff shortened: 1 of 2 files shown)


resolver/resolver.cc
```diff
@@ -2989,8 +2989,15 @@ class ResolveTypeMembersAndFieldsWalk {
...
             ENFORCE(cnst != nullptr, "Rewriter should always use const for typeExpr, which should now be resolved");
-            if (!cnst->symbol.isClassOrModule() || cnst->symbol.asClassOrModuleRef().data(ctx)->flags.isModule ||
-                cnst->symbol.asClassOrModuleRef().data(ctx)->typeArity(ctx) > 0) {
+
+            // Dealias to mimic how type_syntax parsing will work for constant lit node (e.g., want
+            // to allow class aliases).
+            auto dealiased = cnst->symbol.dealias(ctx);
+
+            // Don't want the dealias to apply to the type alias, because MyTypeAlias.new is a runtime error.
+            if (cnst->symbol.isTypeAlias(ctx) || !dealiased.isClassOrModule() ||
```
froydnj, line 2998: I assume we have a test for the `TypeAlias.new` case somewhere, ideally?  If not, can we add one?

froydnj (approved):

resolver/resolver.cc
jez: I'm going to play it cool and say it was a test for my reviewer to see if they're paying attention 🙈 

will add

PR #7557, opened 2024-01-02
Treat `.void` and `.returns(T.anything)` as compatible

jez:
### Motivation

Now that users can write `.returns(T.anything)`, there's essentially two ways
to write the top type in a method signature. This is one more way where the
distinction between the two of them could have snuck through to the user, so we
have to patch up the hole.

### Test plan

See included automated tests.

PR #7563, opened 2024-01-02
Only treat `.void` methods as returning `T.anything`

jez:
### Motivation

Sorbet models `.void` as `returns(Sorbet::Private::Static::Void)` internally,
and it declares that with a normal class in an RBI (`rbi/sorbet/sorbet.rbi`).

The logic was backwards—it should have been `isSubType(methodReturnType,
void_())`, instead of the other way. So sorbet would implicitly treat any method
which returned a super type of .void if it returned top. Which meant:

- methods that return `.void`
- methods that return `Object`
- methods that return `Kernel`
- methods that return `BasicObject`

We could fix this by flipping the sign around, but then it would still be weird
because technically there's nothing stopping people from doing

```ruby
class MyVoid < Sorbet::Private::Static::Void
  # ...
end

sig { returns(MyVoid) }
def example; end
```

And then Sorbet would also treat this as a method that can return anything
(instead of only `MyVoid` instances). So let's just use `==` comparison.

This equality comparison works because ClassTypes are inlined types, which means
that all class types representing `.void` are byte-wise identical. This should
also be a marginal performance improvement.

### Test plan

See included automated tests.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -1444,7 +1444,7 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                 tp.origins.emplace_back(ctx.locAt(bind.loc));
 
                 const core::TypeAndOrigins &typeAndOrigin = getAndFillTypeAndOrigin(ctx, i.what);
-                if (core::Types::isSubType(ctx, core::Types::void_(), methodReturnType)) {
+                if (methodReturnType == core::Types::void_()) {
                     methodReturnType = core::Types::top();
                 }
                 if (!core::Types::isSubTypeUnderConstraint(ctx, constr, typeAndOrigin.type, methodReturnType,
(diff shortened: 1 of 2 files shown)


jez:
* `master`
  * **PR #7563** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/7563?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #7557** <a href="https://app.graphite.dev/github/pr/sorbet/sorbet/7557?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sorbet/sorbet/7563?utm_source=stack-comment).

neilparikh (approved):

PR #7567, opened 2024-01-03
Prefer exact match on singleton class to fuzzy match instance method

jez:
### Motivation

More context when something isn't quite right, also help new Ruby users
understand Ruby's object model better.

### Commit summary

- **prework: Nest this code in an if** (731ceea4d)

  This is going to mess up the whitespace in a bit

- **prework: clang-format the previous commit** (d429555b0)

  This commit is just whitespace changes.

- **Add failing test** (ecf274c1c)

  This commit shows the test file before implementing an autocorrect.

- **Show suggestion to use `.class` to call singleton class method** (2a5a14ccb)

  This commit shows the new changes, and the new autocorrect test changes.

### Test plan

See included automated tests.

**Before:**

Screenshot 2024-01-03 at 1 49 40 PM

**After:**

image

### DIFF
--- a/test/testdata/infer/suggest_dot_class.rb.autocorrects.exp
+++ b/test/testdata/infer/suggest_dot_class.rb.autocorrects.exp
@@ -5,10 +5,10 @@ class A
   def self.my_singleton_class_method; end
 
   def example
-    singleton_method # error: does not exist
-    self.singleton_method # error: does not exist
+    self.class.my_singleton_class_method # error: does not exist
+    self.class.my_singleton_class_method # error: does not exist
     a = A.new
-    a.singleton_method # error: does not exist
+    a.class.my_singleton_class_method # error: does not exist
   end
 end
 # ------------------------------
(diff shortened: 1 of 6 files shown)


test/testdata/infer/suggest_dot_class.rb.autocorrects.exp
```diff
@@ -0,0 +1,14 @@
...
+# typed: true
+
+class A
+  def self.my_singleton_class_method; end
+
+  def example
+    singleton_method # error: does not exist
+    self.singleton_method # error: does not exist
+    a = A.new
+    a.singleton_method # error: does not exist
```
jez: You can see in this commit from before the changes: the presence of a similarly named method (but not identically named method!) would prevent the "you're confusing instance and singleton" error from firing. In this case, it that meant the autocorrect replaced `my_singleton_method` with `singleton_method` instead of suggesting `class.my_singleton_method`.

We should prioritize exact matches over fuzzy matches, because a lot of the time the exact match is something the user wrote, while the fuzzy match is some random method in the standard library that they didn't realize .

ilyailya (approved):

PR #7569, opened 2024-01-03
Allow specifying --track-untyped={everywhere,nowhere}

jez:
### Motivation

This converts from a `bool` to an `enum class` with two entries, and
changes the option parsing to parse string names like

    --track-untyped
    --track-untyped=everywhere
    --track-untyped=nowhere

and also LSP configuration like

    "highlightUntyped": false,
    "highlightUntyped": true,
    "highlightUntyped": "",
    "highlightUntyped": "nowhere",
    "highlightUntyped": "everywhere",
    // and omitted

In the future, I plan to add an option like "EverywhereButTests" which
will use the `isPackagedTest` flag on `core::File` to decide whether to
show untyped highlights there.

### Test plan

See included automated tests.

### DIFF
--- a/core/GlobalState.h
+++ b/core/GlobalState.h
@@ -8,6 +8,7 @@
 #include "core/Loc.h"
 #include "core/Names.h"
 #include "core/Symbols.h"
+#include "core/TrackUntyped.h"
 #include "core/lsp/Query.h"
 #include "core/packages/PackageDB.h"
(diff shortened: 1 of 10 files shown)


froydnj (approved):
enums > bools

jez:
Seems to break VS Code in manual testing.

PR #7571, opened 2024-01-04
highlightUntyped: Make VS Code extension use nowhere/everywhere

jez:
### Motivation

This is a follow up to #7569, which converted from using `bool` to using an enum
for the `--track-untyped` setting, in anticipation of making it no longer be a
boolean.

In this PR, we make the same changes to the VS Code extension.

### Test plan

See included automated tests.

I've also tested manually in VS Code to ensure that the first time you use this
version of the extension, it gracefully upgrades you to use the new value,
instead of keeping the setting defined as a boolean.

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -261,7 +261,7 @@
           "default": false
         },
         "sorbet.highlightUntyped": {
-          "type": "string",
+          "enum": ["nowhere", "everywhere"],
           "description": "Shows warning for untyped values.",
           "default": "nowhere"
         },
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -261,9 +261,9 @@
           "default": false
         },
         "sorbet.highlightUntyped": {
-          "type": "boolean",
+          "type": "string",
           "description": "Shows warning for untyped values.",
-          "default": false
+          "default": "nowhere"
         },
         "sorbet.typedFalseCompletionNudges": {
           "type": "boolean",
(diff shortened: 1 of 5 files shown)


jez:
On second thought, this is going to require some more thought to ensure that people using a newer VS Code extension version with an older Sorbet version aren't left behind.

vscode_extension/package.json
```diff
@@ -261,9 +261,9 @@
           "default": false
         },
         "sorbet.highlightUntyped": {
```
damolina-stripe, line 263: Thinking of this change, I would actually suggest to have a new `sorbet.highlightUntypedMode` (or some other name) and mark `sorbet.highlightUntyped` as `deprecated`.  That way we would only carry the `boolean`-to-`string` logic for 6 months (or n-versions, whichever makes more sense) and then remove it. 
With two settings, removing the two-type handling would lead to no issues other than perhaps someone reverting to default "do nothing" vs a possible `boolean` vs `string` error (although that might not be that bad).

jez: Sounds good, I can work on that.

jez: @damolina-stripe On second thought, there's no really good way for the VS Code extension to detect what version of Sorbet the user is running, and in fact it will differ by project. So I don't expect removing the string -> boolean conversion from the `workspace/didChangeConfiguration` setting, because there's no real need to break backwards compatibility if we don't have to. In particular, I know that people regularly stick to old versions of Sorbet because the type system changes introduced in newer versions make it hard to keep up with. And on the other hand, it's hard to pin the version of an installed VS Code extension to an old version (or maybe, you want the newer version so that you can use it in projects that **are** upgraded, but still have to also use that same version of the extension with legacy projects using older Sorbet versions).

So I think I'd prefer to keep the string -> boolean logic around ~indefinitely, and at that point I'm not sure if there's benefit to adding a second option, because I think that the extension will already gracefully upgrade the user from boolean to string config when they install the new VS Code extension version.

damolina-stripe (approved):

vscode_extension/package.json
damolina-stripe: Makes sense to leave as is then. Thanks for verifying.

PR #7572, opened 2024-01-04
Add option to highlight untyped everywhere but tests

jez:
### Motivation

In Stripe's codebase, usages of untyped in tests vastly outnumber usages in
non-test files. Put another way: usages of untyped in non-test files are rare
enough that you might conceivably want Sorbet to highlight all of them, all the
time.

The only way you can do that right now is by highlighting untyped in test files
too. While many tests use very little untyped, other tests are nearly completely
untyped, which is very noisy.

This setting allows people to opt in to a sort of "always on in non-test file"
style of development. In particular, it updates the **Toggle** command to toggle
between the last two values of the setting. So if you want to toggle between "on
except in tests" and "off" you can. Or you can toggle between "on except in
tests" and "on everywhere." Or, like today, you can toggle between
everywhere/nowhere.

### Test plan

See included automated tests.

I have tested this manually, but I need to do more testing to make sure that new
VS Code extension versions work with old Sorbet versions (also the case for
#7571)

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -24,13 +24,13 @@
   ],
   "activationEvents": [
     "onCommand:sorbet.configure",
+    "onCommand:sorbet.configureHighlightUntyped",
     "onCommand:sorbet.disable",
     "onCommand:sorbet.enable",
     "onCommand:sorbet.restart",
     "onCommand:sorbet.setLogLevel",
     "onCommand:sorbet.showOutput",
     "onCommand:sorbet.toggleHighlightUntyped",
-    "onCommand:sorbet.configureHighlightUntyped",
     "onCommand:sorbet.toggleTypedFalseCompletionNudges",
     "onLanguage:ruby",
     "workspaceContains:sorbet/*"
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -4,7 +4,7 @@
   "description": "Ruby IDE features, powered by Sorbet.",
   "author": "Stripe Inc.",
   "license": "Apache-2.0",
-  "version": "0.3.29",
+  "version": "0.3.30",
   "publisher": "sorbet",
   "icon": "icon.png",
   "repository": {
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -30,6 +30,7 @@
     "onCommand:sorbet.setLogLevel",
     "onCommand:sorbet.showOutput",
     "onCommand:sorbet.toggleHighlightUntyped",
+    "onCommand:sorbet.configureHighlightUntyped",
     "onCommand:sorbet.toggleTypedFalseCompletionNudges",
     "onLanguage:ruby",
     "workspaceContains:sorbet/*"
@@ -95,6 +96,12 @@
       {
         "command": "sorbet.toggleHighlightUntyped",
         "title": "Toggle highlighting untyped code",
+        "category": "Sorbet",
+        "enablement": "workbenchState != empty"
+      },
+      {
+        "command": "sorbet.configureHighlightUntyped",
+        "title": "Configure untyped code highlighting",
         "category": "Sorbet",
         "enablement": "workbenchState != empty"
       },
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -57,6 +57,12 @@
         "category": "Sorbet"
       },
       {
+        "command": "sorbet.configureHighlightUntyped",
+        "title": "Configure untyped code highlighting",
+        "category": "Sorbet",
+        "enablement": "workbenchState != empty"
+      },
+      {
         "command": "sorbet.copySymbolToClipboard",
         "title": "Copy Symbol to Clipboard",
         "category": "Sorbet",
@@ -96,12 +102,6 @@
       {
         "command": "sorbet.toggleHighlightUntyped",
         "title": "Toggle highlighting untyped code",
-        "category": "Sorbet",
-        "enablement": "workbenchState != empty"
-      },
-      {
-        "command": "sorbet.configureHighlightUntyped",
-        "title": "Configure untyped code highlighting",
         "category": "Sorbet",
         "enablement": "workbenchState != empty"
       },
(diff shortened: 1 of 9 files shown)


vscode_extension/package.json
```diff
@@ -30,6 +30,7 @@
     "onCommand:sorbet.setLogLevel",
     "onCommand:sorbet.showOutput",
     "onCommand:sorbet.toggleHighlightUntyped",
+    "onCommand:sorbet.configureHighlightUntyped",
```
damolina-stripe, line 33: This list, and commands, is sorted alphabetically to make it easier to detect duplicates / conflicts (common issue is for ppl to define the same thing twice which throws a wrench when debugging).

vscode_extension/src/commandIds.ts
```diff
@@ -54,6 +54,12 @@ export const SORBET_SAVE_PACKAGE_FILES = "sorbet.savePackageFiles";
 export const TOGGLE_HIGHLIGHT_UNTYPED_COMMAND_ID =
   "sorbet.toggleHighlightUntyped";
 
+/**
+ * Configure highlighting of untyped code.
+ */
+export const CONFIGURE_HIGHLIGHT_UNTYPED_COMMAND_ID =
```
damolina-stripe, line 60: These constants are sorted alphabetically, not per feature.

vscode_extension/package.json
```diff
@@ -98,6 +99,12 @@
         "category": "Sorbet",
         "enablement": "workbenchState != empty"
       },
+      {
+        "command": "sorbet.configureHighlightUntyped",
+        "title": "Configure untyped code highlighting",
+        "category": "Sorbet",
+        "enablement": "workbenchState != empty"
+      },
```
damolina-stripe, line 107: Alpha-sorting applies particularly to this section since  different `enablement` rules cause funny issue (not from experience).

vscode_extension/src/commands/toggleUntypedCodeHighlighting.ts
```diff
@@ -1,15 +1,26 @@
...
+function toggle(log: Log, configuration: SorbetExtensionConfig): TrackUntyped {
+  if (configuration.oldHighlightUntyped != null) {
+    return configuration.oldHighlightUntyped;
+  }
+
+  switch (configuration.highlightUntyped) {
     case "nowhere":
       return "everywhere";
+    case "everywhere-but-tests":
+      return "nowhere";
```
damolina-stripe, line 19: Nit: toggling a tri-state flag leads to a funny construct like this. - e.g. if you toggle `everywhere-but-tests` you go to `nowhere` but if you toggle again you go to `everywhere` so `everywhere-but-tests` is not accessible.  Should `toggle` loop `nowhere` » `everywhere-but-tests` » `everywhere` » `nowhere` ?

vscode_extension/src/commands/toggleUntypedCodeHighlighting.ts
```diff
@@ -34,12 +47,12 @@ function backwardsCompatibleValue(
 /**
  * Toggle highlighting of untyped code.
  * @param context Sorbet extension context.
- * @returns `true` if highlighting is now enabled, `false` otherwise.
+ * @returns The new TrackUntyped setting
```
damolina-stripe, line 50: Not needed, but it makes it a link in docs.

```suggestion
 * @returns The new {@link TrackUntyped} setting.
```

vscode_extension/src/commands/toggleUntypedCodeHighlighting.ts
```diff
@@ -56,3 +69,52 @@ export async function toggleUntypedCodeHighlighting(
 
   return targetState;
 }
+
+/**
+ * Set highlighting of untyped code to sepcific setting.
```
damolina-stripe, line 74:
```suggestion
 * Set highlighting of untyped code to specific setting.
```

vscode_extension/src/commands/toggleUntypedCodeHighlighting.ts
jez: @damolina-stripe I believe that in your case, I've added an extra field to store "the setting before the current setting" so that toggling works. Allowing people to toggle between "everywhere" and "everywhere-but-tests" was a central focus of this change, and I've verified that it works in my editor locally.

damolina-stripe: Or perhaps, this not longer is a `Toggle` and should be a `Configure Code Highlight`  action.

jez: There is already a separate "configure" action implemented below, which allows you to choose from a drop down.

See the `oldHighlightUntyped` field access immediately above for how the sticky toggle works.

damolina-stripe (approved):

vscode_extension/src/commands/toggleUntypedCodeHighlighting.ts
```diff
@@ -59,3 +69,52 @@ export async function toggleUntypedCodeHighlighting(
...
+
+    const { activeLanguageClient: client } = context.statusProvider;
+    if (client) {
+      client.sendNotification("workspace/didChangeConfiguration", {
+        settings: {
+          highlightUntyped: targetState,
+        },
+      });
+    } else {
+      context.log.debug("ToggleUntyped: No active Sorbet LSP to notify.");
```
damolina-stripe, line 117:
```suggestion
      context.log.debug("ConfigureUntyped: No active Sorbet LSP to notify.");
```

vscode_extension/src/commands/toggleUntypedCodeHighlighting.ts
```diff
@@ -59,3 +69,52 @@ export async function toggleUntypedCodeHighlighting(
...
+  });
+
+  if (selectedItem) {
+    const oldHighlightUntyped = context.configuration.highlightUntyped;
+    context.configuration.oldHighlightUntyped = oldHighlightUntyped;
+
+    const targetState = selectedItem.trackWhere;
+    await context.configuration.setHighlightUntyped(targetState);
+    context.log.info(
+      `ToggleUntyped: Untyped code highlighting: ${targetState}`,
```
damolina-stripe, line 106:
```suggestion
      `ConfigureUntyped: Untyped code highlighting: ${targetState}`,
```

PR #7573, opened 2024-01-05
Deduplicate import autocorrects

ilyailya:
### Test plan

See included automated tests.


packager/VisibilityChecker.cc
```diff
@@ -512,6 +516,22 @@ class VisibilityCheckerPass final {
                         core::Context ctx{gs, core::Symbols::root(), f.file};
                         VisibilityCheckerPass pass{ctx, gs.packageDB().getPackageInfo(pkgName)};
                         ast::TreeWalk::apply(ctx, pass, f.tree);
+                        std::vector<std::unique_ptr<core::Error>> errorsToReport;
+                        errorsToReport.reserve(pass.importErrors.size());
+                        pass.importErrors.clear();
+                        for (auto &err : pass.importErrors) {
+                            errorsToReport.emplace_back(*err);
```
jez, line 523:
```suggestion
                            errorsToReport.emplace_back(std::move(*err));
```

test/testdata/packager/import-autocorrect/import-autocorrect.autocorrects.exp
```diff
@@ -0,0 +1,43 @@
+# -- test/testdata/packager/import-autocorrect/__package.rb --
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
+class Package < PackageSpec
+  import Dep
```
ilyailya, line 7: without the fix, we should see two `import Dep` statements

test/cli/package-autocorrect-missing-import/test.out
```diff
@@ -23,6 +23,10 @@ foo_class.rb:15: Unable to resolve constant `MyClass` https://srb.help/5002
       NN |class Digest::Class
           ^^^^^^^^^^^^^^^^^^^
 
+foo_class.rb:17: `Test::Foo::Bar::OtherPackage::TestUtil` is defined in a test namespace and cannot be referenced in a non-test file https://srb.help/3720
+    17 |  Test::Foo::Bar::OtherPackage::TestUtil
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
jez, line 29: Why do we get a new error here?

test/testdata/packager/import-autocorrect/a.rb
```diff
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 # typed: strict
-# selective-apply-code-action: source.fixAll.sorbet
```
jez: It looks you went away from the "test via code actions" approach, but you only deleted one of the two `.rbedited` files.

test/cli/package-autocorrect-missing-import/test.out
ilyailya: >I suppose that's expected, because we're deferring the "missing import" errors until later in the pipeline now.

Exactly! The changes to the `.out` files do not generate any new error, and do not remove any existing errors. The difference is in the order of errors (because we changed the reporting order, as you've noticed), and some errors don't have autocorrects anymore, because otherwise it would lead to double importing.

test/testdata/packager/import-autocorrect/a.rb
ilyailya: Great catch, thanks! Just deleted the last `. rbedited` file

test/testdata/packager/import-autocorrect/import-autocorrect.autocorrects.exp
jez: When I check out the first commit in your pull request, which adds this test but has not yet fixed it, this test does not fail. So I don't think that we can actually use this as a regression test, and we'd want to make a CLI test.

But the existing CLI test that was updated in this PR seems like it already tests that, so I think it's fine to delete this test if you want.

packager/VisibilityChecker.cc
```diff
@@ -510,8 +538,19 @@ class VisibilityCheckerPass final {
...
+                        for (auto &[symbol, locs] : pass.imports) {
+                            fast_sort(locs,
+                                      [](const auto a, const auto b) -> bool { return a.beginPos() > b.beginPos(); });
+                            reportMissingImport(ctx, package, symbol, locs);
+                        }
+                        for (auto &[symbol, locs] : pass.testImports) {
+                            fast_sort(locs,
+                                      [](const auto a, const auto b) -> bool { return a.beginPos() > b.beginPos(); });
+                            reportMissingTestImport(ctx, package, symbol, locs);
+                        }
```
jez, line 553: This only deduplicates within a single file.

If you're missing an import for the symbol and have two references in separate files, Sorbet will still generate two imports.

packager/VisibilityChecker.cc
```diff
@@ -30,6 +30,58 @@ static core::SymbolRef getEnumClassForEnumValue(const core::GlobalState &gs, cor
     return core::Symbols::noSymbol();
 }
 
+static void reportMissingTestImport(const core::Context &ctx, const core::packages::PackageInfo &package,
```
jez, line 33: This name was slightly misleading to me before I read the implementation of the function:

It's not saying that the `test_import` is missing, it's saying that the `test_import` is there but it needs to be an `import` to be able to use the constant in non-test code.

packager/VisibilityChecker.cc
```diff
@@ -30,6 +30,58 @@ static core::SymbolRef getEnumClassForEnumValue(const core::GlobalState &gs, cor
...
+            }
+            e.addErrorLine(pkg.declLoc(), "Defined here");
+        }
+    }
+}
+
+static void reportMissingImport(const core::Context &ctx, const core::packages::PackageInfo &package,
+                                core::SymbolRef symbol, std::vector<core::LocOffsets> locs) {
+    auto importAdded = false;
+    for (auto it = locs.rbegin(); it != locs.rend(); ++it) {
```
jez, line 56: Why are we going in reverse order? Wouldn't we want to print the one that's first in the file?

packager/VisibilityChecker.cc
```diff
@@ -30,6 +30,58 @@ static core::SymbolRef getEnumClassForEnumValue(const core::GlobalState &gs, cor
...
+        if (auto e = ctx.beginError(loc, core::errors::Packager::UsedTestOnlyName)) {
+            auto otherFile = symbol.loc(ctx).file();
+            auto &otherPackage = ctx.state.packageDB().getPackageNameForFile(otherFile);
+            e.setHeader("Used `{}` constant `{}` in non-test file", "test_import", symbol.show(ctx));
+            auto &pkg = ctx.state.packageDB().getPackageInfo(otherPackage);
+            if (!importAdded) {
+                if (auto exp = package.addImport(ctx, pkg, false)) {
+                    e.addAutocorrect(std::move(exp.value()));
+                }
+            }
```
jez, line 47: `importAdded` is never set to `true`. Can we add a test which would have failed and caught this?

packager/VisibilityChecker.cc
ilyailya: Fixed. I've also added a test with two references to a class in the same file and two references to a class in different files

test/helpers/position_assertions.cc
```diff
@@ -13,6 +13,7 @@
 #include "test/helpers/position_assertions.h"
 #include <iterator>
 #include <regex>
+#include <string.h>
```
jez, line 16: This include looks unneeded.

test/cli/package-double-import-autocorrect/test.out
```diff
@@ -0,0 +1,52 @@
...
+     3 |class C < PackageSpec
+                             ^
+Errors: 4
+
+====================
+
+# typed: strict
+
+class B < PackageSpec
+  import A
```
jez, line 43: When I revert all the changes to `VisibilityChecker.cc` in this change, I still see only one `import` line added here, which suggests that this test does not actually exercise the behavior in this PR.

In particular, I'm guessing that the `hasSeen` logic in `AutcorrectSuggestion.cc` is preventing duplicate imports from being applied, but that would also mean that it's preventing non-duplicate imports from being added.

I think you should solve that problem first, then revisit this PR. Specifically:

- Given packages A1, A2, and B
- The constants A1::Foo and A2::Foo are defined inside (and exported from) their respective packages.
- Package B uses each of A1::Foo and A2::Foo twice.
- If you run `sorbet -a`, you only get one line added to the `__package.rb` file for B

Here's a longer test case.
[package-test.zip](https://github.com/sorbet/sorbet/files/14146482/package-test.zip)

test/testdata/packager/import-autocorrect/__package.1.rbedited 
jez, line 1: There is a trailing space in this file's name. Please remove that.

test/testdata/packager/import-autocorrect/__package.1.rbedited 
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
+class Package < PackageSpec
+  import Dep
+end
```
jez, line 7: When I add a line to the end of this file, like "random string", the test still passes. I do not think this test is actually running.

test/helpers/position_assertions.cc
```diff
@@ -1831,19 +1842,33 @@ void ApplyCodeActionAssertion::checkAll(
...
         actualEditedFileContents = string(file->source());
 
         for (auto &e : c->edits) {
+            auto &td = c->textDocument;
+            auto filename = td->uri;
+            auto version = td->version;
+
+            if (!std::holds_alternative<double>(version)) {
+                continue;
+            }
```
jez, line 1863: I do not know what this piece of code is attempting to do. As far as I can tell, in Sorbet the `version` field is always `null` and never a `double`, which means that the edits to this file in this PR effectively disable every one of our `rbedited` file snapshot tests.

packager/VisibilityChecker.cc
```diff
@@ -451,36 +514,10 @@ class VisibilityCheckerPass final {
...
-                        "A `{}` file is allowed to define constants outside of the package's namespace,\n    "
-                        "but must still respect its enclosing package's imports.",
-                        "# packaged: false");
-                }
-
-                if (!db.errorHint().empty()) {
-                    e.addErrorNote("{}", db.errorHint());
-                }
-            }
+            imports[lit.symbol].emplace_back(lit.loc);
```
jez, line 517: In using `lit.symbol` here, this causes another bug.

[package-test-2.zip](https://github.com/sorbet/sorbet/files/14146681/package-test-2.zip)

In this example, the package `B` uses both `A1::Foo` and `A1::Bar`. That causes two `import A1` lines to get added to the `__package.rb` file for `B`.

jez (changes requested):
I think we still have a fair amount of work to do on this.

core/errors/packager.h
```diff
@@ -28,5 +28,6 @@ constexpr ErrorClass UsedTestOnlyName{3720, StrictLevel::False};
 constexpr ErrorClass InvalidExport{3721, StrictLevel::False};
 // constexpr ErrorClass ExportingTypeAlias{3722, StrictLevel::False};
 constexpr ErrorClass ImportNotVisible{3723, StrictLevel::False};
+constexpr ErrorClass PackageIssues{3719, StrictLevel::False};
```
jez, line 31: Don't re-use numbers

```suggestion
constexpr ErrorClass PackageIssues{3724, StrictLevel::False};
```

website/docs/error-reference.md
```diff
@@ -741,6 +741,17 @@ cases, a common solution is to factor the shared functionality to a new package,
...
+
+> This error is specific to Stripe's custom `--stripe-packages` mode. If you are
+> at Stripe, please see [go/modularity](http://go/modularity) for more.
+
+A package references constants from another package without the corresponding
+`import` or `test_import` statements.
+
+To fix this error, add the `import` or `test_import` statements related to the
+packages in the `__package.rb` file where the error is reported.
+
```
jez, line 754: The name of the error class is `PackageIssues` but this website documentation suggests that the error is actually more narrow. Which is it? Should we make the error class have a more specific name, or should we make the documentation suggest that there are multiple "issues" which cause this error code?

main/lsp/LSPFileUpdates.cc
```diff
@@ -117,6 +121,9 @@ LSPFileUpdates::fastPathFilesToTypecheck(const core::GlobalState &gs, const LSPC
     int i = -1;
     for (auto &oldFile : gs.getFiles()) {
         i++;
+
+        bool insertedIntoExtraFiles = false;
```
jez, line 125: Why do we sometimes want to insert early and sometimes we want to insert late? This seems like a bug waiting to happen which is why I'd like to understand the context.

main/lsp/LSPFileUpdates.h
```diff
@@ -66,18 +66,20 @@ class LSPFileUpdates final {
...
     fastPathFilesToTypecheck(const core::GlobalState &gs, const LSPConfiguration &config,
                              const std::vector<std::shared_ptr<core::File>> &updatedFiles,
-                             const UnorderedMap<core::FileRef, std::shared_ptr<core::File>> &evictedFiles);
+                             const UnorderedMap<core::FileRef, std::shared_ptr<core::File>> &evictedFiles,
+                             bool isNoopUpdateForRetypecheck);
 
-    FastPathFilesToTypecheckResult fastPathFilesToTypecheck(const core::GlobalState &gs,
-                                                            const LSPConfiguration &config) const;
+    FastPathFilesToTypecheckResult fastPathFilesToTypecheck(const core::GlobalState &gs, const LSPConfiguration &config,
+                                                            bool isNoopUpdateForRetypecheck) const;
```
jez, line 82: Can you remind me whether we have a path towards not needing to use this boolean, or whether it's a parameter we expect to be here forever?

I had tried pretty hard to not need to know whether the edit was a retypecheck edit or whether it was a normal edit throughout most of the LSP code, so I would like to understand the context in case we ever do something in the future which would let us remove this boolean.

main/pipeline/pipeline.h
```diff
@@ -55,8 +55,9 @@ std::vector<ast::ParsedFile> autogenWriteCacheFile(const core::GlobalState &gs,
 // Note: `cancelable` and `preemption task manager` are only applicable to LSP.
 // If `intentionallyLeakASTs` is `true`, typecheck will leak the ASTs rather than pay the cost of deleting them
 // properly, which is a significant speedup on large codebases.
-void typecheck(const core::GlobalState &gs, std::vector<ast::ParsedFile> what, const options::Options &opts,
-               WorkerPool &workers, bool cancelable = false,
+void typecheck(const core::GlobalState &gs, std::vector<ast::ParsedFile> what,
+               std::vector<core::FileRef> extraFilesToFlush, const options::Options &opts, WorkerPool &workers,
```
jez, line 59: This should either be `std::vector<core::FileRef> &&extraFilesToFlush` or `const std::vector<core::FileRef> &extraFilesToFlush`

main/lsp/LSPTypechecker.cc
```diff
@@ -318,7 +318,16 @@ vector<core::FileRef> LSPTypechecker::runFastPath(LSPFileUpdates &updates, Worke
...
     const auto presorted = true;
     const auto cancelable = false;
-    pipeline::typecheck(*gs, move(sorted), config->opts, workers, cancelable, std::nullopt, presorted);
+
+    std::vector<core::FileRef> extraFilesToFlush;
+    if (isNoopUpdateForRetypecheck) {
+        for (auto &file : updates.updatedFileIndexes) {
+            extraFilesToFlush.emplace_back(file.file);
+        }
+    }
```
jez, line 327: Out of curiosity, did you consider an alternative where we pass the `__package.rb` into the call to `pipeline::typecheck`, which would mean we don't need to pass this extra flag into `typecheck`?

jez: I'm asking because this turns `pipeline::typecheck` into a leaky abstraction. To use it properly, you have to know about LSP and how the fast/slow path works, whereas before it was a simple contract where you pass in files and it typechecks them.

Since being able to use this function correctly involves LSP-specific knowledge (fast path / slow path and the files that they contain), it would be nice to deal with that entirely in `main/lsp` instead of in `main/pipeline`.

test/pkg_autocorrects_test.cc
```diff
@@ -90,140 +90,6 @@ struct TestPackageFile {
...
-    auto test = TestPackageFile::create(gs, "my_package/__package.rb", pkg_source);
-
-    auto &package = test.targetPackage(gs);
-    ENFORCE(package.exists());
-    auto addImport = package.addImport(gs, test.newPackage(gs), true);
-    ENFORCE(addImport, "Expected to get an autocorrect from `addImport`");
-    auto replaced = addImport->applySingleEditForTesting(pkg_source);
-    CHECK_EQ(expected, replaced);
-}
-
```
jez, line 226: Why do we have to delete all these tests? Are they covered by some other aspect of our tests?

packager/VisibilityChecker.cc
```diff
@@ -496,31 +512,140 @@ class VisibilityCheckerPass final {
...
+        }
+
+        for (auto &[pkg, fixes] : combinedFixes) {
+            auto &package = gs.packageDB().getPackageInfo(pkg);
+            auto newImportLoc = package.newImportLoc(gs, package);
+            fmt::memory_buffer autocorrect;
+            std::vector<core::AutocorrectSuggestion::Edit> edits;
+            auto fullPackageName = [&gs](const sorbet::core::packages::PackageInfo &pkg) {
+                return absl::StrJoin(pkg.fullName(), "::", core::packages::NameFormatter(gs));
+            };
```
jez, line 601: There's a couple problems here

- This allocates a lambda in a loop.
- This makes more people need to know about this weird `NameFormatter` thing.
- This code is the exact same as the `toString` method on `PackageName` in `packager.cc`, which we removed three calls to and have replace them with calls to this method.

Let's take that `toString` method, move it to `PackageInfo`, implement that method in `PackageInfoImpl`, and make a no-op version of it too. Then, VisibilityChecker won't need to add an import to `str_join.h`.

jez:
Some notes to myself to take another look at in this review:

- [ ] `extraFilesToFlush`
- [ ] `isNoopUpdateForRetypecheck`

packager/VisibilityChecker.cc
```diff
@@ -496,31 +512,140 @@ class VisibilityCheckerPass final {
...
+            files.size());
         for (size_t i = 0; i < files.size(); ++i) {
             taskq->push(i, 1);
         }
 
-        workers.multiplexJob("VisibilityChecker", [&gs, &files, &barrier, taskq]() {
+        workers.multiplexJob("VisibilityChecker", [&gs, &files, taskq, toImportQ, replaceTestImportQ]() {
             size_t idx;
+            int processedByThread = 0;
+            std::vector<std::tuple<core::packages::MangledName, core::packages::MangledName, bool>> importCapacitor;
```
jez, line 530: What is a "capacitor" in this context? I've only ever heard this used in [electrical engineering](https://en.wikipedia.org/wiki/Capacitor)

packager/VisibilityChecker.cc
```diff
@@ -496,31 +512,140 @@ class VisibilityCheckerPass final {
...
+                                                              WorkerPool::BLOCK_INTERVAL(), gs.tracer());
+             !result.done(); result = replaceTestImportQ->wait_pop_timed(replaceTestImportThreadResult,
+                                                                         WorkerPool::BLOCK_INTERVAL(), gs.tracer())) {
+            if (!result.gotItem()) {
+                continue;
+            }
+            for (auto &[package, packageToImport, loc] : replaceTestImportThreadResult) {
+                combinedFixes[package].convertTestImports.emplace(packageToImport, loc);
+            }
+        }
```
jez, line 588: Is there a way we can do this with one queue? I'm worried about having one queue grow very large without anything consuming from it until the first queue completely drains.

packager/VisibilityChecker.cc
```diff
@@ -496,31 +512,140 @@ class VisibilityCheckerPass final {
...
+
+        if (combinedFixes.empty()) {
+            return files;
+        }
+
```
jez, line 593:
```suggestion
```

Is this necessary? If this is empty, the loop below will not run and skip down to the bottom `return files` invocation.

packager/VisibilityChecker.cc
```diff
@@ -496,31 +512,140 @@ class VisibilityCheckerPass final {
...
+                    e.addErrorSection(core::ErrorSection(fmt::format("Used `{}` instead of `{}` for {}", "test_import",
+                                                                     "import", fullPackageName(packageToImport)),
+                                                         {}));
+
+                    auto [lineStart, _] = importLoc.findStartOfLine(gs);
+                    core::Loc replaceLoc(importLoc.file(), lineStart.beginPos(), importLoc.endPos());
+                    edits.push_back({replaceLoc, fmt::format("import {}", fullPackageName(packageToImport))});
+                }
+
+                e.addAutocorrect({"Fix package issues", edits});
```
jez, line 646:
```suggestion
                e.addAutocorrect({"Fix package visibility errors", edits});
```

jez:

→ https://go/builds/bui_PoFtK3eD1lXZ0K
→ https://go/builds/bui_PoFtek1e6bMPQy
→ https://go/builds/bui_PoFt1LZN0ugWaH

jez (changes requested):
Passing back for the time being, most comments have been left, but there are still some more things to look into.

jez:

→ https://go/builds/bui_PojsT4UZpTvgWD
→ https://go/builds/bui_PojsUhFmSbPX4p
→ https://go/builds/bui_PojsrJiFtQfpem

jez (changes requested):
I think this PR is susceptible to the same problems we were talking about in the context of exports. I’m able to check out this branch and get visibility errors in __package.rb files to appear or disappear incorrectly when making fast path edits.

Let’s be sure to find some time in the office tomorrow to chat about this. I think that the solution we had talked about for exports also fixes the problems for imports, but we should sync tomorrow to be sure that we’re on the same page w.r.t. how there can be stale/missing import errors in the current implementation of the PR and how we’ll want to fix this.

website/docs/error-reference.md
ilyailya: Renamed the error to `PackageVisibilityIssues` from `PackageIssues`

main/lsp/LSPFileUpdates.cc
ilyailya: Streamlined the logic in that bit
https://github.com/sorbet/sorbet/pull/7573/commits/86f9b15cfa5440549399ae3680ed33adc646cd5a

PR #7574, opened 2024-01-06
Add alternative Array#sample sig when no arg0 is passed in

marianosimone:
### Motivation
When no argument is passed in, the result is just `T.nilable(Elem)`, which simplifies not having to deal with the `T.any`

### Test plan
See included automated tests.

### DIFF
--- a/rbi/core/array.rbi
+++ b/rbi/core/array.rbi
@@ -2268,6 +2268,15 @@ class Array < Object
     )
     .returns(T.any(T.nilable(Elem), T::Array[Elem]))
   end
+  sig do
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_PK2FngEK1Hs30O
→ https://go/builds/bui_PK2GcUd1RmCNnJ
→ https://go/builds/bui_PK2GJpPL44vObh

marianosimone:
This doesn't work, because multiple `sig`s for the same method in an `rbi` can't be added when there are keyword arguments

PR #7583, opened 2024-01-10
don't cargo-cult `:setter_proc` for prop value validation

froydnj:
### Motivation

The code for `validate_prop_value` contains this pearl:

```
    # We call `setter_proc` here without binding to an instance, so it'll run
    # `instance_variable_set` if validation passes, but nothing will care.
```

Something very much does care, namely, the instance variables of `SetterFactory`:

```
[dev] main:0> class A < T::Struct
[dev] main:0>   prop :example, String
[dev] main:0> end
=> T::Private::Types::Void::VOID
[dev] main:0> T::Props::Private::SetterFactory.instance_variables
=> []
[dev] main:0> A.validate_prop_value(:example, "a")
=> "a"
[dev] main:0> T::Props::Private::SetterFactory.instance_variables
=> [:@example]
```

It seems bad to have `SetterFactory` slowly accumulate instance variables for every prop that gets validated, never mind the GC implications.

I don't love the duplication here, but I don't want to slow down the setter path by introducing an extra block invocation.  I should try benchmarking the non-duplication change, though.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/decorator.rb
+++ b/gems/sorbet-runtime/lib/types/props/decorator.rb
@@ -103,7 +103,7 @@ class T::Props::Decorator
     # We call `setter_proc` here without binding to an instance, so it'll run
     # `instance_variable_set` if validation passes, but nothing will care.
     # We only care about the validation.
-    prop_rules(prop).fetch(:setter_proc).call(val)
+    prop_rules(prop).fetch(:value_validate_proc).call(val)
   end
 
   # For performance, don't use named params here.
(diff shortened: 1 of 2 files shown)


froydnj:
> I don't love the duplication here, but I don't want to slow down the setter path by introducing an extra block invocation. I should try benchmarking the non-duplication change, though.

I benchmarked this with `rake bench:setters` and it slowed things down by ~20-30%.  I think I'm willing to do some code duplication for avoiding that kind of performance problem.

jez (approved):

PR #7587, opened 2024-01-10
Allow non-Payload files to update superclass of Payload files

jez:
### Motivation

This relates to #7289 and #7295, where the superclass of some class in the
`IRB` gem changed in a newer version, and people aren't able to work around
it on their own.

### Test plan

See included automated tests.

I did some quick sanity checks to make sure this won't cause problems on the
fast path. We already exclude retypechecking files in the payload:

<https://github.com/sorbet/sorbet/blob/73bfa488b13bef5a4fea2f9c4c587619ddd80c7b/main/lsp/LSPFileUpdates.cc#L133-L137>

So I think that this change is basically fine.


test/testdata/rbi/redefine_superclass_payload.rb
```diff
@@ -0,0 +1,5 @@
+# typed: true
+
+class NewParent; end
+class Complex < NewParent
```
Morriar, line 4: I'm concerned about people doing really wrong things inside shims that we may not be able to control.

Would it make sense to allow this only based on a specific flag or sigil?

```rb
# typed: true
# allow-payload-parent-override: true
```

So we can at least set up a rule to review those changes?
```

File: test/lsp_test_runner.cc [hunk 2]
```diff
@@ -547,6 +547,7 @@ TEST_CASE("LSPTest") {
         opts->ruby3KeywordArgs =
             BooleanPropertyAssertion::getValue("experimental-ruby3-keyword-args", assertions).value_or(false);
         opts->typedSuper = BooleanPropertyAssertion::getValue("typed-super", assertions).value_or(true);
+        // TODO(jez) Allow allow suppressPayloadSuperclassRedefinitionFor in a testdata test assertion?
```
jez, line 550: Going to punt on these two TODO for now but will leave them behind as an easy thing to grep for in the future.

jez:
@Morriar after chatting with the team we decided to go with a command line option, instead of a comment, because we were worried that

- It might be hard to update how a generated file is generated, and this is likely to impact generated files more often than hand-written files.
- In a generated file, you might only need to do this for one class, despite there being dozens or hundreds of class definitions in the file, where for all the other classes redefining the class would be a problem.

Let me know what you think.

test/lsp_test_runner.cc
ilyailya:
```suggestion
        // TODO(jez) Allow suppressPayloadSuperclassRedefinitionFor in a testdata test assertion?
```

resolver/resolver.cc
```diff
@@ -1683,6 +1705,29 @@ class ResolveConstantsWalk {
         return depth;
     }
 
+    static core::ClassOrModuleRef resolvePayloadSuperclassClassName(const core::GlobalState &gs,
```
ilyailya, line 1708: Would it make sense to add a comment explaining this is not a technique we expect people to use?
```suggestion
    // This function resolves a fully qualified class name into a ClassOrModuleRef.
    // Usually in Sorbet, we steer away from string matching in favor of using `SymbolRef`s due to performance benefits.
    // However, this function stands as an exception to assist with the `--suppress-payload-superclass-redefinition-for`
    // command. Please, do not repeat this practice elsewhere.
    static core::ClassOrModuleRef resolvePayloadSuperclassClassName(const core::GlobalState &gs,
```

ilyailya (approved):
LGTM!
Looking forward to @Morriar's feedback too

resolver/resolver.cc
jez: If you have strings, this is the preferred way of doing things.

> Usually in Sorbet, we steer away from string matching in favor of using `SymbolRef`s due to performance benefits.

The quoted piece of code _does not_ use string matching—it explicitly converts from a String to a SymbolRef, once per input string. It's as close as you could get to parsing the class name as a ruby program and resolving the corresponding constant lit.

If someone else needed to do this elsewhere in Sorbet, this would absolutely be the code worth copying referencing, and the name of the function should make it clear enough that this function is not meant to be used for anything except for this PayloadSuperclass error thing (if someone tried to call this exact function elsewhere, it would be easy to catch that in code review).

Morriar (approved):
The option makes sense and still allow us to control where these overrides can be applied with even more precision 👍  

Thanks!

test/cli/suppress-payload-superclass-redefinition-for/test.sh
```diff
@@ -0,0 +1,23 @@
...
+cd test/cli/suppress-payload-superclass-redefinition-for
+
+echo "-- complex.rbi --"
+
+2>&1 "$cwd/main/sorbet" --silence-dev-message \
+  --suppress-payload-superclass-redefinition-for=Complex complex.rbi
+
+echo "-- complex.rbi --"
+2>&1 "$cwd/main/sorbet" --silence-dev-message \
+  --suppress-payload-superclass-redefinition-for=IRB::RelineInputMethod irb.rbi
```
Morriar, line 15: Should we add an example with more than one constant error suppressed?

test/cli/suppress-payload-superclass-redefinition-for/test.sh
```diff
@@ -0,0 +1,30 @@
...
+  echo "Expected to fail!"
+  exit 1
+fi
+
+
+echo "-- double.rb --"
+2>&1 "$cwd/main/sorbet" --silence-dev-message \
+  --suppress-payload-superclass-redefinition-for=Complex \
+  --suppress-payload-superclass-redefinition-for=IRB::RelineInputMethod \
+  double.rb
```
Morriar, line 29: I think we're missing the `double.rb` file?

jez: 🤦‍♂️ thanks!

Issue #5258, opened 2022-02-04
Sorbet cannot type the result of `with_index` properly


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20strict%0Arequire%20'sorbet-runtime'%0A%0Axs%20%3D%20T.let%28%5B1%5D%2C%20T%3A%3AArray%5BInteger%5D%29%0A%0Ares1%20%3D%20xs.each.with_index%20%7B%7Cx%2C%20index%7C%20x.to_s%20%7D%0Ap%20res1%20%23%20%3D%3E%20%5B1%5D%0AT.reveal_type%28res1%29%20%23%20ideally%3A%20T%3A%3AArray%5BInteger%5D%0A%0Ares2%20%3D%20xs.map.with_index%20%7B%7Cx%2C%20index%7C%20x.to_s%20%7D%0Ap%20res2%20%23%20%3D%3E%20%5B%221%22%5D%0AT.reveal_type%28res2%29%20%23%20ideally%3A%20T%3A%3AArray%5BString%5D%0A%0Ares3%20%3D%20xs.each_with_index.map%20%7B%20%7Cx%2C%20index%7C%20x.to_s%20%7D%0Ap%20res3%20%23%20%3D%3E%20%5B%221%22%5D%0AT.reveal_type%28res3%29%20%23%20correct)

```ruby
# typed: strict
require 'sorbet-runtime'

xs = T.let([1], T::Array[Integer])

res1 = xs.each.with_index {|x, index| x.to_s }
p res1 # => [1]
T.reveal_type(res1) # ideally: T::Array[Integer]

res2 = xs.map.with_index {|x, index| x.to_s }
p res2 # => ["1"]
T.reveal_type(res2) # ideally: T::Array[String]

res3 = xs.each_with_index.map { |x, index| x.to_s }
p res3 # => ["1"]
T.reveal_type(res3) # correct
```

#### Observed output

```
editor.rb:8: Revealed type: T.untyped https://srb.help/7014
     8 |T.reveal_type(res1) # ideally: T::Array[Integer]
        ^^^^^^^^^^^^^^^^^^^
  Got T.untyped originating from:
    editor.rb:6:
     6 |res1 = xs.each.with_index {|x, index| x.to_s }
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

editor.rb:12: Revealed type: T.untyped https://srb.help/7014
[...]

PR #7596, opened 2024-01-12
Update `with_index` signature to remove `T.untyped`

damolina-stripe:
Update `with_index` signature to remove `T.untyped`.
- Misc. formatting changes (including the odd long spaces before `.returns`)

### Motivation

Found `with_index` usages in our codebase that lead to `T.untyped` highlighting.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/enumerator.rbi
+++ b/rbi/core/enumerator.rbi
@@ -398,11 +398,11 @@ class Enumerator < Object
   # `offset`
   # :   the starting index to use
   sig do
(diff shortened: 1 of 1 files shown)


jez:
I don't think this signature is correct. There is more context in #5258.

damolina-stripe:
> I don't think this signature is correct. There is more context in #5258.
😞., thanks for catching this.

PR #7597, opened 2024-01-13
Merge zero-width autocorrects

ilyailya:
If Sorbet proposes two or more zero-width edits at the same loc only one gets applied. The PR fixes that bug.

### Test plan

See included automated tests.

### DIFF
--- a/test/cli/autocorrect-extend/test.out
+++ b/test/cli/autocorrect-extend/test.out
@@ -7,6 +7,7 @@ def top_level; end
 # Should only add one extend
 class SuggestSigAndExtend
   extend T::Sig
+  extend T::Sig
   sig { returns(NilClass) }
   def foo; end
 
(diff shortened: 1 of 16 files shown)


test/cli/autocorrect-extend/test.out
```diff
@@ -6,6 +6,7 @@ def top_level; end
 
 # Should only add one extend
 class SuggestSigAndExtend
+  extend T::Sig
```
froydnj, line 9: Either this comment is wrong, or the code in `AutocorrectSuggestion.cc` is wrong.

froydnj (changes requested):
Requesting changes for now, since it sounds like there's some discussion happening elsewhere about what the right approach here is.

ilyailya:
Closing this PR, because the approach I tried here renders more serious bugs than https://github.com/sorbet/sorbet/pull/7573/

PR #7598, opened 2024-01-13
various: fix comment & doc typos

p-linnane:
### Motivation

I'm a Homebrew maintainer. While cleaning up various typos within our own repos, I noticed some upstream Sorbet typos. I figured it would make sense to open a PR to submit these fixes. Please let me know if any of these files should not be modified by hand, or are inappropriate to fix.

### Test plan

These changes are in comments and docs only. I reached out on the Sorbet Slack and was told to submit this PR specifically for comments and docs. I'll follow up with another PR that may potentially affect variable/class names.


jez (approved):

jez:
Thank you so much!

jez:
This will merge automatically when the tests pass (which I have every reason to believe they should).

p-linnane:
I had to run `format_website.sh`, which I wasn't aware of. Should be fixed now.

PR #7599, opened 2024-01-13
Remove T::InterfaceWrapper

### LINKED ISSUE #763: Remove T::InterfaceWrapper?
From the code:
```ruby
# Once we ship static type checking, we should get rid of this entirely.
class T::InterfaceWrapper
```

This might be a good thing to remove before open source to reduce the API surface.

jez:
Fixes #763

### Motivation

This was always legacy code that predated Sorbet (from the type system that
existed at Stripe before Sorbet). It was never documented and always
discouraged, even in Stripe's codebase.

We couldn't remove it in time for open sourcing Sorbet, and then we never took
the time to actually remove it. But over the past few days I've been able to
finally tear the last remnants of it out of Stripe's codebase, so now we can
finally get rid of this.

### Test plan

Existing tests


jez:
Fixes #763

neilparikh (commented):
You need to remove the `require_relative 'types/interface_wrapper'` from `lib/sorbet-runtime.rb` and update the static test.

neilparikh (approved):

PR #7604, opened 2024-01-16
various: fix additional typos

p-linnane:
### Motivation

Follow up to https://github.com/sorbet/sorbet/pull/7598.

### Test plan

I expect breakage while running this through CI since it touches many different things. I also expect to need to drop some of these commits if I've misunderstood things. Looking forward to feedback here. Thanks!


p-linnane:
@froydnj I did not run `format_cxx.sh`, but have resolved that now. Can you unblock the builds again please? 🙏

rbi/stdlib/rdoc.rbi
```diff
@@ -8145,7 +8145,7 @@ class RDoc::RDoc
 
   # Removes file extensions known to be unparsable from `files` and TAGS files
   # for emacs and vim.
-  def remove_unparseable(files); end
+  def remove_unparsable(files); end
```
jez, line 8148: The files in `rbi/` declare methods provided by either the Ruby standard library or common gems. In the rdoc gem, the name of this method is actually `remove_unparseable`

https://ruby.github.io/rdoc/RDoc/RDoc.html#method-i-remove_unparseable

Would you mind going through and either getting rid of the changes in `rbi/` or confirming that there actually was a mismatch between the method that the corresponding gem declares and the RBI file that Sorbet ships with?

jez (changes requested):

rbi/stdlib/rdoc.rbi
p-linnane: Absolutely. This was the type of feedback I was hoping to get, so thank you for that. I'll run through and make the appropriate changes.

p-linnane: I've confirmed all the `rbi/` issues are indeed upstream, so I've dropped those 3 commits.

jez (approved):

PR #7592, opened 2024-01-12
Add more context for T_NONE errors in the context of CC table marking

aadi-stripe:
### Motivation

As a followup to https://github.com/sorbet/sorbet/pull/7545, this patch adds more context to `T_NONE` mark errors we've been seeing at Stripe. This patch adds more debug info during CC table marking.

### Test plan

Tested the logging by calling it directly (without gating on `T_NONE`).

See included automated tests.

PR #7607, opened 2024-01-17
Make GC patch for CME debugging null-safe

aadi-stripe:
### Motivation

In https://github.com/sorbet/sorbet/pull/7592, I shipped a patch to Ruby to add more context to `T_NONE` errors encountered in GC. However, there is a small null safety concern with this patch, re the call to `rb_id2name`. It might not actually be a problem, but just fixing it anyway for posterity.

### Test plan

Tested on Stripe codebase.

### DIFF
--- a/third_party/ruby/gc-more-t-none-context.patch
+++ b/third_party/ruby/gc-more-t-none-context.patch
@@ -1,8 +1,8 @@
 diff --git gc.c gc.c
-index 346a77ec63..172373e40d 100644
+index 346a77ec63..6e025cb200 100644
 --- gc.c
 +++ gc.c
-@@ -2935,7 +2935,27 @@ cc_table_mark_i(ID id, VALUE ccs_ptr, void *data_ptr)
+@@ -2935,7 +2935,33 @@ cc_table_mark_i(ID id, VALUE ccs_ptr, void *data_ptr)
          return ID_TABLE_DELETE;
      }
      else {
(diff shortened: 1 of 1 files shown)


aadi-stripe:

→ https://go/builds/bui_POOsrR3C0nYa68
→ https://go/builds/bui_POOswnTttbYatN
→ https://go/builds/bui_POOsVciH09HMVX

froydnj (approved):
I think this is fine, two things to note:

1) I believe glibc's `printf` and company will actually print `(null)` for `%s` when you pass in `NULL`, so maybe this is not a problem; and
2) Maybe we should print some kind of log if we don't have `objname` anyway?  (I have kind of a hard time believing that any interned `ID` will have a `NULL` name, but I suppose it's possible...)

aadi-stripe:
> I think this is fine, two things to note:
> 
> 1. I believe glibc's `printf` and company will actually print `(null)` for `%s` when you pass in `NULL`, so maybe this is not a problem; and
> 2. Maybe we should print some kind of log if we don't have `objname` anyway?  (I have kind of a hard time believing that any interned `ID` will have a `NULL` name, but I suppose it's possible...)

I think you're right, this is not a concern very likely. I'll ship it anyway because it's just safer code technically. Can add logging later, don't know if it's essential since this is only fired on a `T_NONE` anyway

PR #7612, opened 2024-01-17
Stop showing definition results for literals

jez:
Other plugins typically try to associate definitions with things inside
string literals, e.g. by treating them as relative paths or something
like that.

Having Sorbet treat the "definition" of a string literal as the location
of the literal is not useful, and conflicts with other plugins that try
to provide their own definitions for string literals.

### Motivation

Requested by a developer of a Stripe-internal VS Code extension

cc @damolina-stripe

### Test plan

See included automated tests.


neilparikh (approved):
Can you write an LSP test for this?

PR #7615, opened 2024-01-18
Make find all references work on all props

jez:
### Motivation

Requested by a user

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/LSPTask.cc
+++ b/main/lsp/LSPTask.cc
@@ -453,10 +453,6 @@ AccessorInfo LSPTask::getAccessorInfo(const core::GlobalState &gs, core::SymbolR
             return info;
         }
         info.fieldSymbol = gs.lookupFieldSymbol(ownerCls, fieldName);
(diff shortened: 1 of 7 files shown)


main/lsp/LSPTask.cc
```diff
@@ -453,10 +453,6 @@ AccessorInfo LSPTask::getAccessorInfo(const core::GlobalState &gs, core::SymbolR
             return info;
         }
         info.fieldSymbol = gs.lookupFieldSymbol(ownerCls, fieldName);
-        if (!info.fieldSymbol.exists()) {
-            // field symbol does not exist, so `symbol` must not be an accessor.
-            return info;
-        }
```
jez, line 459: It looks like Sorbet already had logic to attempt to make find all references find all of ivar usages + getter method calls + setter method calls. This short circuit was preventing that from applying if the ivar was not defined (which is the case in a typed: true file).

ilyailya (approved):

PR #7620, opened 2024-01-19
Add additional details to the error message when subtyping check fails

neilparikh:
Review commit-by-commit. Commit message body explains key things to look out for in that commit.

An example of what it looks like:
```ruby
# typed: true
class A
  extend T::Generic
  X = type_member
end

T.let(A[T.proc.params(x: {a: String}).returns([String])].new,
      A[T.proc.params(x: {a: Integer}).returns([Integer])])
```
```
err.rb:7: Argument does not have asserted type `A[T.proc.params(arg0: {a: Integer}).returns([Integer])]` https://srb.help/7007
     7 |T.let(A[T.proc.params(x: {a: String}).returns([String])].new,
     8 |      A[T.proc.params(x: {a: Integer}).returns([Integer])])
  Got `A[T.proc.params(arg0: {a: String}).returns([String])]` originating from:
    err.rb:7:
     7 |T.let(A[T.proc.params(x: {a: String}).returns([String])].new,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Detailed explanation:
    `T.proc.params(arg0: {a: String}).returns([String])` is not equivalent to `T.proc.params(arg0: {a: Integer}).returns([Integer])` for invariant type member `A::X`
      `[String]` is not a subtype of `[Integer]` for covariant type member `Proc1::Return`
        `String` is not a subtype of `Integer` for index `0` of `1`-tuple
      `{a: String}` is not a supertype of `{a: Integer}` for contravariant type member `Proc1::Arg0`
        `Integer` is not a subtype of `String` for key `Symbol(:a)`
```

One design choice I made that's worth calling out: if the passed in shape type doesn't have enough keys, we still loop over each key and check each one, but if it's a tuple type with not enough elements, we don't loop through. My reasoning here was that if the missing element in tuple is early on (ex. `[String, Integer, String]` and `[Integer, String]`), all the errors will be "off by one" and not very useful, while this is not a problem with shape types.

### Motivation

Provide better error messages when root cause of failure is nested within a complex type.

### Test plan

See included automated tests.


core/types/subtyping.cc
```diff
@@ -1259,7 +1265,13 @@ bool isSubTypeUnderConstraintSingle(const GlobalState &gs, TypeConstraint &const
                             ++i;
                             result = Types::isSubTypeUnderConstraint(gs, constr, a1.elems[i], el2, mode);
```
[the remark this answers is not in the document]
jez: To prevent this sort of mistake from having happened in the first place, I would actually make it so that the only version of `isSubType` that allows omitting parameters is the one that just takes gs and two types. I.e., if you're passing in a mode, you should be forced to pass in a collector too.

jez (changes requested):

core/Types.h
```diff
@@ -707,8 +719,9 @@ TYPE(OrType) final : public Refcounted {
     friend TypePtr Types::Boolean();
     friend class NameSubstitution;
     friend class serialize::SerializerImpl;
+    template <class T>
     friend bool Types::isSubTypeUnderConstraint(const GlobalState &gs, TypeConstraint &constr, const TypePtr &t1,
-                                                const TypePtr &t2, UntypedMode mode);
+                                                const TypePtr &t2, UntypedMode mode, T &errorCollector);
```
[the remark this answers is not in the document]
neilparikh: One of them is instantiated here: https://github.com/sorbet/sorbet/pull/7620/files#diff-eb2da3e78eea17626fb563370b444cc444ebc51aa7b8cd35da2fad4e43cbdd36R1517
The other is (implicitly) instantiated here: https://github.com/sorbet/sorbet/pull/7620/files#diff-eb2da3e78eea17626fb563370b444cc444ebc51aa7b8cd35da2fad4e43cbdd36R1386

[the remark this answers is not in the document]
froydnj: It's fine to have the function declared without an implementation: whatever types we call that function with, the compiler will turn into a call to a named function that must exist at link time.  If it doesn't exist at link time, of course, you'll get peculiar errors that might not be the most straightforward to track down.

This situation is a little different, because it only declares that `Types::isSubTypeUnderConstraint` -- with any template parameter -- is permitted to access the private members of `OrType`, which seems appropriate.  This declaration is purely for semantic checks and has no effect on code generation.  It's...not the best, as we're now leaking what are supposed to be internal implementation details of `isSubTypeUnderConstraint` into the types.  We could just do `friend class Types` and drop the other declarations of `Types` static methods below?

core/types/subtyping.cc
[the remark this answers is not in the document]
neilparikh: I looked through a bunch of the uses of `isSubType` (and variants) that don't pass in a mode, and a bunch of those are used for error reporting too, so I think I'll just make `errorCollector` required for every call for now.

jez:
I'm still working on reviewing this—didn't have time to finish today.

This is the kind of PR that I have to get my hands dirty with to review effectively (e.g., check it out [in my editor and poke around](https://blog.jez.io/cli-code-review/)), and so while I'm there I've gone ahead and just implemented the things that I would leave as PR feedback.

For the moment I'm accumulating those changes into this branch. If you're curious, you can take a look at my new commits to see the sorts of things I plan to bring up as PR comments eventually:

https://github.com/sorbet/sorbet/compare/jez-error-details

But I haven't gotten to the actual meat of the change yet, and I don't want to go requesting things like refactors until I'm sure they'll be worth asking for (based on the meat of the implementation).

I plan to continue this tomorrow and hopefully finish it then too.

core/types/subtyping.cc
```diff
@@ -1211,25 +1220,47 @@ bool isSubTypeUnderConstraintSingle(const GlobalState &gs, TypeConstraint &const
```
jez, line 1220: I can't comment on the relevant line, but we're likely going to want to explain when !`classSymbolIsAsGoodAs` too. Since that's on classes not on types, it's likely good to use "derives from" instead of "is subtype of" in the error phrasing:

```
`{}` does not derive from `{}`"
```

This can also be a follow up in a future PR.

core/Error.h
```diff
@@ -106,6 +106,31 @@ class Error {
...
+    ErrorDetailsCollector newCollector() {
+        return ErrorDetailsCollector();
+    }
+    std::optional<ErrorSection> toErrorSection();
+};
+
+class NoOpErrorDetailsCollector {
+public:
+    void addErrorDetails(NoOpErrorDetailsCollector e) {}
+    NoOpErrorDetailsCollector newCollector() {
```
jez, line 127: These methods should end up being `const` in both.

core/Error.h
```diff
@@ -106,6 +106,31 @@ class Error {
...
+    std::vector<ErrorDetailsCollector> children;
+
+    ErrorDetailsCollector() : message("") {}
+    ErrorDetailsCollector(std::string msg) : message(msg) {}
+
+    void addErrorDetails(ErrorDetailsCollector e);
+    ErrorDetailsCollector newCollector() {
+        return ErrorDetailsCollector();
+    }
+    std::optional<ErrorSection> toErrorSection();
```
jez, line 121: This is also a `const` method I think.

resolver/resolver.cc
```diff
@@ -2486,14 +2490,17 @@ class ResolveTypeMembersAndFieldsWalk {
...
+                    e.addErrorSections(errorDetailsCollector);
                     if (lowerBoundTypeLoc.exists()) {
                         auto replacementType = ctx.state.suggestUnsafe.has_value() ? core::Types::untypedUntracked()
                                                                                    : parentType->lowerBound;
                         e.replaceWith("Replace with `T.untyped`", lowerBoundTypeLoc, "{}", replacementType.show(ctx));
                     }
                 }
             }
-            if (!core::Types::isSubType(ctx, memberType->upperBound, parentType->upperBound)) {
+            // TODO:
```
jez, line 2501: What is this `TODO` for?

core/types/subtyping.cc
```diff
@@ -1257,32 +1288,81 @@ bool isSubTypeUnderConstraintSingle(const GlobalState &gs, TypeConstraint &const
...
-                            if (!result) {
-                                break;
+                            auto subCollector = errorDetailsCollector.newCollector();
+                            if (!Types::isSubTypeUnderConstraint(gs, constr, a1.elems[i], el2, mode, subCollector)) {
+                                result = false;
+                                if constexpr (std::is_same<T, ErrorDetailsCollector>::value) {
+                                    // TODO(neil): once we pass in whether this a covariant or contravariant call,
+                                    // we should reword it in the "expected ... but got ..." form
+                                    auto message = ErrorColors::format("`{}` is not a subtype of `{}` for item `{}`",
+                                                                       a1.elems[i].show(gs), el2.show(gs), i);
```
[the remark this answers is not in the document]
jez: circling back after seeing this error in context, I also think it's important to mention tuple in the error message, as it can be easy to miss that. So my favorite is `for tuple index 0`

test/cli/detailed-errors-completeness/test.out
```diff
@@ -0,0 +1,261 @@
...
+    `Upper` is not a subtype of `Lower` for item `0`
+
+test.rb:68: Parameter `x` of type `[String]` not compatible with type of overridable method `A#foo` https://srb.help/5035
+    68 |  def foo(x, y:, z: [''], &blk)
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    test.rb:50: The super method parameter `x` was declared here with type `[Integer]`
+    50 |  def foo(x, y:, z: [1], &blk)
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  Note:
+    A parameter's type must be a supertype of the same parameter's type on the super method.
```
jez, line 53: By convention, the `Note` should always be last in Sorbet error messages. If you could audit the places where you've added `e.addErrorSections` to make sure it's below any `addErrorNote` (**and** below any helper function which might eventually call `addErrorNote`) that would be good.

test/cli/detailed-errors-completeness/test.out
```diff
@@ -0,0 +1,261 @@
...
+            ^^^
+  Note:
+    A `Hash` passed as a keyword splat must match the type of all keyword parameters
+    because Sorbet cannot see what specific keys exist in the `Hash`.
+  Possible explanation:
+    `String` is not a subtype of `Integer` for item `0`
+  Autocorrect: Use `-a` to autocorrect
+    test.rb:17: Replace with `T.unsafe(h)`
+    17 |bar(**h)
+              ^
```
jez, line 163: This error message is kind of cursed actually. The problem is actually that Sorbet doesn't even know if `h` has an `x` key at all.

It seems like we want to get rid of the error note here and move that into the error details collector?

test/cli/detailed-errors/test.out
```diff
@@ -0,0 +1,168 @@
+shapes.rb:10: Expected `{a: Integer, b: String}` but found `{}` for argument `x` https://srb.help/7002
```
jez, line 1: Is it expected that the first few of these don't have any explanations?

That's fine, we don't have to fix every error message in this PR, I just wanted to confirm.

test/cli/detailed-errors/test.out
```diff
@@ -0,0 +1,168 @@
...
+  Expected `{a: Integer, b: String}` for argument `x` of method `Object#takes_shape`:
+    shapes.rb:5:
+     5 |sig { params(x: {a: Integer, b: String}).void }
+                     ^
+  Got `{a: Integer(1), b: Integer(1)} (shape of T::Hash[T.untyped, T.untyped])` originating from:
+    shapes.rb:18:
+    18 |takes_shape({a: 1, b: 1})
+                    ^^^^^^^^^^^^
+  Possible explanation:
+    `Integer(1)` is not a subtype of `String` for key `Symbol(:b)`
```
jez, line 61: We probably want to do a better job of pretty-printing these. Ideally it would not say `Symbol(:b)` but rather something like "for key `b:`"

jez (commented):
This is looking really good.

Most of my feedback is in comments here. The rest is pushed to jez-error-details. Every commit in jez-error-details should be treated as a request-for-change comment on the PR, though feel free to push back on them the same way you'd push back on a PR comment.

infer/environment.cc
```diff
@@ -1595,18 +1600,21 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
...
                             e.addErrorNote("You may need to add additional `{}` annotations", "sig");
                         }
-                    } else if (!core::Types::isSubType(ctx, ty.type, castType)) {
+                    } else if (!core::Types::isSubType(ctx, ty.type, castType, errorDetailsCollector)) {
                         if (c.cast == core::Names::assumeType()) {
                             if (auto e = ctx.beginError(bind.loc, core::errors::Infer::IncorrectlyAssumedType)) {
                                 e.setHeader("Assumed expression had type `{}` but found `{}`", castType.show(ctx),
                                             ty.type.show(ctx));
                                 e.addErrorSection(ty.explainGot(ctx, ownerLoc));
+                                core::TypeErrorDiagnostics::explainTypeMismatch(ctx, e, errorDetailsCollector, castType,
```
neilparikh, line 1616: I wasn't able to come up with a test case for this.

infer/environment.cc
```diff
@@ -1715,6 +1726,8 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                                 e.setHeader("Incompatible assignment to variable declared via `{}`: `{}` is not a "
                                             "subtype of `{}`",
                                             "let", tp.type.show(ctx), cur.type.show(ctx));
+                                core::TypeErrorDiagnostics::explainTypeMismatch(ctx, e, errorDetailsCollector, tp.type,
```
neilparikh, line 1729: I wasn't able to come up with a test case for this either.

infer/environment.cc
```diff
@@ -1617,6 +1625,7 @@ Environment::processBinding(core::Context ctx, const cfg::CFG &inWhat, cfg::Bind
                             if (auto e = ctx.beginError(bind.loc, core::errors::Infer::CastTypeMismatch)) {
                                 e.setHeader("Argument does not have asserted type `{}`", castType.show(ctx));
                                 e.addErrorSection(ty.explainGot(ctx, ownerLoc));
+                                e.addErrorSections(std::move(errorDetailsCollector));
```
neilparikh, line 1628: Calling `addErrorSections` directly since the error doesn't have an "expected" and "got".

neilparikh (commented):
Here the follow up issues I’ll make once this lands:
- Pass in whether the call is covariant or contravariant in the error collector. This unlocks:
  - An error for a missing key in a shape
  - An error if the tuple has the wrong number of values
  - rewording error messages to `Expected … but got …`
- If the type is a `Proc`, `Hash` or `Array`, there should be special logic when looping over the type members so that instead of saying `Proc1::Return` we can say `return value for Proc`
- In `equivUnderConstraint`,  add error message like `{} is a subtype of {} but not the reverse, so they are not equivalent`
- In `classSymbolAsGoodAs`, explain when `!classSymbolIsAsGoodAs`, in the form of `{} does not derive from {}`
- Explicitly show the defaulted type members
- Show keys in a better way than `Symbol(:b)`. In the show method for ShapeType, we have some logic to show the keys in a hash-like way. We can extract that out and then reuse it here.

test/cli/detailed-errors/test.out
jez: I don't quite understand this, I don't think variance is involved here. Maybe we can walk through this together tomorrow.

core/types/subtyping.cc
```diff
@@ -1211,25 +1220,47 @@ bool isSubTypeUnderConstraintSingle(const GlobalState &gs, TypeConstraint &const
...
                 }
-                if (!result) {
-                    break;
+                if (!doesMemberMatch) {
+                    result = false;
+                    if constexpr (std::is_same<T, ErrorDetailsCollector>::value) {
+                        // TODO(neil): once we pass in whether this a covariant or contravariant call,
+                        // we should reword it in the "expected ... but got ..." form
+                        // TODO(neil): if the type member is for a Proc we should have special wording,
+                        // ie. Proc return value instead of Proc1::Return and Proc arg 1 instead of Proc1::Arg1
```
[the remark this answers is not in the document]
jez: I don't think we necessarily have to do this for Hash and Array—people recognize that `Hash` is a generic class with type arguments, because users are forced to provide these every time they make a typed `Hash`.

The difference with `T.proc` is that people think that `T.proc` is something "built into Sorbet," and not just a generic class under the hood.

We _might_ want to silence the error about `Hash::Elem` specifically, because it's a bit weird, but I could go either way on that. Based on the way it's created, if either of `K` or `V` have a problem, then so will `Elem`, so it'll always be duplicated and yeah probably makes sense to be hidden.

infer/environment.cc
jez: Right now we only assume class types, but we don't have any custom custom error messages to explain when something isn't a subtype of a class type.

This is a niche error, we don't have to test it. It will be tested indirectly in the future as we add more cases to the errorCollector.

infer/environment.cc
jez:
```ruby
# typed: true

xs = T.let([0], [Integer])
1.times do
  xs = [""]
end
```

You don't need to write a test case for this if you don't want to, it should virtually always be covered by the other tests (from a subtyping perspective there's nothing particularly special here).

core/Error.cc
```diff
@@ -130,6 +130,30 @@ string ErrorSection::toString(const GlobalState &gs) const {
...
+
+std::optional<ErrorSection> ErrorSection::Collector::toErrorSection() const {
+    if (children.size() == 0) {
+        return nullopt;
+    }
+    vector<ErrorLine> lines;
+    for (auto &c : children) {
+        toErrorSectionHelper(c, lines, 0);
+    }
+    return ErrorSection("Detailed explanation:", lines);
```
jez, line 154:
```suggestion
    return ErrorSection("Detailed explanation:", move(lines));
```

core/Error.cc
```diff
@@ -172,6 +196,12 @@ void ErrorBuilder::addErrorSection(optional<ErrorSection> &&section) {
     }
 }
 
+void ErrorBuilder::addErrorSections(const ErrorSection::Collector &&errorDetailsCollector) {
+    if (auto errorSection = errorDetailsCollector.toErrorSection()) {
+        addErrorSection(errorSection.value());
```
jez, line 201:
```suggestion
        addErrorSection(move(errorSection.value()));
```

core/TypeErrorDiagnostics.cc
```diff
@@ -35,8 +35,10 @@ namespace {
 }
 } // namespace
 
-void TypeErrorDiagnostics::explainTypeMismatch(const GlobalState &gs, ErrorBuilder &e, const TypePtr &expected,
+void TypeErrorDiagnostics::explainTypeMismatch(const GlobalState &gs, ErrorBuilder &e,
+                                               const ErrorSection::Collector &collector, const TypePtr &expected,
                                                const TypePtr &got) {
+    e.addErrorSections(std::move(collector));
```
jez, line 41: Can you add a TODO that the rest of this function body should eventually be implemented in the collector? e.g., the attached class hint and the metatype hint are both things that should move into subtyping.cc eventually.

resolver/resolver.cc
```diff
@@ -2478,14 +2480,18 @@ class ResolveTypeMembersAndFieldsWalk {
...
+                                                                    parentType->lowerBound, memberType->lowerBound);
                     if (lowerBoundTypeLoc.exists()) {
                         auto replacementType = ctx.state.suggestUnsafe.has_value() ? core::Types::untypedUntracked()
                                                                                    : parentType->lowerBound;
                         e.replaceWith("Replace with `T.untyped`", lowerBoundTypeLoc, "{}", replacementType.show(ctx));
                     }
                 }
             }
-            if (!core::Types::isSubType(ctx, memberType->upperBound, parentType->upperBound)) {
+            // TODO:
```
jez, line 2492:
```suggestion
```

jez (approved):

PR #7594, opened 2024-01-12
Backport klass->cme write barrier change

aadi-stripe:
### Motivation

We are seeing GC issues at Stripe, particularly tied to callable method entries. We suspect this fix from Jan 2023 (https://github.com/ruby/ruby/pull/7113) will fix the issue.

### Test plan

Tested in Stripe CI.

PR #7545, opened 2023-12-20
Add code in Ruby GC to give contextual info on parent object for rare T_NONE errors

aadi-stripe:
Adds logging about the "parent object" surrounding a `try to mark T_NONE` error in the Ruby GC.

### Motivation

At Stripe, we are occasionally seeing `try to mark T_NONE` errors in GC which are leading to crashes -- this is a pathological, hard-to-debug case which is probably triggered by a bad interaction with some gem extension, however, we are not sure. Adding more contextual information about the parent object currently being marked will likely help us debug these issues better.

### Test plan

* Tested by calling the new code paths in some test scripts (outside of `rb_bug`, ie just printing to stdout).
* Prints information like:
```
P:T_HASH[ k:T_OBJECT(BClass),v:T_OBJECT(AClass), ]
P:T_HASH[ k:T_OBJECT(BClass),v:T_OBJECT(AClass), ]
```
when the parent object is a hash, or
```
P:T_CLASS(AClass)
```
when the parent object is a class.

* CI builds at Stripe.

See included automated tests.

PR #7621, opened 2024-01-19
Remove GC patch for T_NONE debugging in T_HASH objects

aadi-stripe:
Removes some code that was added in https://github.com/sorbet/sorbet/pull/7545 -- want to reduce patching and risk of segfaults.

### Motivation

The original `T_NONE` issue we were seeing with classes may be resolved after we backported https://github.com/sorbet/sorbet/pull/7594. It's possible that it's still happening with hash objects but getting masked because the hash debugging code added in https://github.com/sorbet/sorbet/pull/7545 is segfaulting.

### Test plan

As per usual, tested in Stripe CI and locally with `fprintf(stdout` _outside_ the unlikely conditional that calls into `rb_bug`.

### DIFF
--- a/third_party/ruby/gc-t-none-context.patch
+++ b/third_party/ruby/gc-t-none-context.patch
@@ -1,80 +1,18 @@
 diff --git gc.c gc.c
-index 6f4ae3a397..346a77ec63 100644
+index 6f4ae3a397..033a89f8f3 100644
 --- gc.c
 +++ gc.c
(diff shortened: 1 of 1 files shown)


aadi-stripe:

→ https://go/builds/bui_PPCXAQWFU5dzvB
→ https://go/builds/bui_PPCXuDEQ8jHedH
→ https://go/builds/bui_PPCXmWDWGLQN7d

froydnj (approved):

PR #7633, opened 2024-01-26
Fix signature of `Thread.each_caller_location`

casperisfine:
Followup: https://github.com/sorbet/sorbet/pull/7293 (cc @nate-at-gusto)

It always yields a `Backtrace::Location` ~~and always returns `nil`~~. Actually if you `break something` in the block, it returns that. 

This save having to use `T.must` or something in the block.

### DIFF
--- a/rbi/core/thread.rbi
+++ b/rbi/core/thread.rbi
@@ -1230,7 +1230,7 @@ class Thread < Object
 
   # Yields each frame of the current execution stack as a
   # backtrace location object.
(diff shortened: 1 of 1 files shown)


paracycle (approved):

ilyailya:

→ https://go/builds/bui_PTMaZZuhQUYlhc
→ https://go/builds/bui_PTMaPGnCy3Ppka
→ https://go/builds/bui_PTMaHLbq98GZUG

ilyailya (approved):

PR #7638, opened 2024-01-31
When the code action request has a selection, only extraction code actions should be shown

neilparikh:
Easiest to review commit-by-commit with `?w=1`.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/requests/code_action.cc
+++ b/main/lsp/requests/code_action.cc
@@ -192,98 +192,104 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerDelegate &t
         }
     }
 
-    auto queryResult = LSPQuery::byLoc(config, typechecker, params->textDocument->uri, *params->range->start,
-                                       LSPMethod::TextDocumentCodeAction, false);
+    if (loc.beginPos() == loc.endPos()) {
+        // No selection
+        auto queryResult = LSPQuery::byLoc(config, typechecker, params->textDocument->uri, *params->range->start,
+                                           LSPMethod::TextDocumentCodeAction, false);
 
-    // Generate "Move method" code actions only for class method definitions
-    if (queryResult.error == nullptr) {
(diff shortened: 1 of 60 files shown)


test/testdata/lsp/code_actions/move_method/no_sig.A.rbedited
```diff
@@ -1,12 +1,11 @@
-# typed: strict
+# typed: true
 # selective-apply-code-action: refactor.extract
 #
 # The test is asserting we can move a method without a sig
 
 module BarModule
   extend T::Sig
   def self.bar
-# ^^^^^^^^^^^^ error: The method `bar` does not have a `sig`
```
neilparikh, line 9: The way our test framework is that if there's a `apply-code-action` after a `error`, it assumes the code action is a quick fix for the error, and checks if the code action is return for the range the error corresponds to. In this case, the error and code action are unrelated.

main/lsp/requests/code_action.cc
```diff
@@ -192,98 +192,104 @@ unique_ptr<ResponseMessage> CodeActionTask::runRequest(LSPTypecheckerDelegate &t
...
-                auto documentEdits = VariableExtractor::getEdits(typechecker, config, loc);
-                if (!documentEdits.empty()) {
-                    auto action = make_unique<CodeAction>("Extract Variable");
-                    action->kind = CodeActionKind::RefactorExtract;
+                auto action = make_unique<CodeAction>(fmt::format("Delete T.{}", resp->callerSideName.show(gs)));
+                action->kind = CodeActionKind::RefactorRewrite;
+                action->edit = move(workspaceEdit);
+                result.emplace_back(move(action));
+            } else {
+            }
```
ilyailya, line 269: Looks like you've got an empty `else` block, is that expected?
```cpp
} else {
}
```

neilparikh: Good catch, I forgot to delete that after moving the code in it. I've deleted it.

ilyailya (approved):

PR #7641, opened 2024-02-01
Deduplicate export suggestions

ilyailya:
The PR fixes a bug when an autocorrect for adding `export <PackageName>` was applied twice.

**Issue:**
In certain condition an autocorrect to add a package export might be applied twice, see included test for more details.

**Solution:**
Similar to https://github.com/sorbet/sorbet/pull/7573. I've introduced a queue, which collects information about the errors and autocorrects to be reported. Potential autocorrects are deduplicated by using set. Before my change, errors were reported straight away.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/packager/export-autocorrect/__package.rb
+++ b/test/testdata/packager/export-autocorrect/__package.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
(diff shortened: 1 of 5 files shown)


jez (changes requested):
Going to mark this as request changes because the base PR is going to need work done. Feel free to re-request review when the base branch is also ready.

PR #7643, opened 2024-02-01
add a test for no-prop constructors

froydnj:
### Motivation

This is a distillation of a test that failed on Stripe's codebase with #6693 applied.

cc @jeffcarbs 

### Test plan

See included automated tests.


jeffcarbs:
Thanks for the repro! I cherry-picked this onto that other PR and put up a fix: https://github.com/sorbet/sorbet/pull/6693/commits/16a649e2e9829c2c782e3b2dca7100071a6d5910. This matches what the generated deser is doing, which is returning the number of found keys and having the regular `initialize` method handle raising on unrecognized keys.

ilyailya (approved):

PR #7644, opened 2024-02-01
Give ARGV the T::Array[String] type

dcecile:
### Motivation
I have some basic argument processing code that I want to switch to `typed: strong`.

### Test plan
The only docs I could find weren't 100% definitive that these are strings: https://docs.ruby-lang.org/en/3.3/globals_rdoc.html#label-ARGV

But bug report appears confirm that they are frozen strings: https://bugs.ruby-lang.org/issues/12128

Any pointers on what tests would apply here?

Also, should there be types for the related globals [$0](https://docs.ruby-lang.org/en/3.3/globals_rdoc.html#label-240) and [$*](https://docs.ruby-lang.org/en/3.3/globals_rdoc.html#label-24-2A+-28ARGV-29)?

### DIFF
--- a/rbi/core/constants.rbi
+++ b/rbi/core/constants.rbi
@@ -6,7 +6,7 @@
 # A library like
 # [`OptionParser`](https://docs.ruby-lang.org/en/2.7.0/OptionParser.html) can be
 # used to process command-line arguments.
-::ARGV = T.let(T.unsafe(nil), T::Array[T.untyped])
+::ARGV = T.let(T.unsafe(nil), T::Array[String])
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_PU2nHlZAVdjW2v
→ https://go/builds/bui_PU2nO9XakjowFS
→ https://go/builds/bui_PU2ntci4xM2DM4

froydnj:
This PR seems fine from the Sorbet side of things, but it turns up a number of failures on Stripe's codebase, mostly because Sorbet now knows that `ARGV[i]` returns a `T.nilable(String)` and people are treating it as `String`.  We need to fix those first, and then this can merge.

PR #7646, opened 2024-02-03
Override Struct.new sig

dduugg:
```
Struct.new(:Foo).name
=> nil
Struct.new('Foo').name
=> "Struct::Foo"
```

### Motivation

Resolves https://sorbet.run/#%23%20typed%3A%20true%0A%0AStruct.new%28%3AFoo%29.name%0AStruct.new%28'Foo'%29.name

### DIFF
--- a/rbi/core/struct.rbi
+++ b/rbi/core/struct.rbi
@@ -45,6 +45,77 @@ class Struct < Object
   end
   def initialize(arg0, *arg1, keyword_init: false, &blk); end
 
+  #  call-seq:
+  #    Struct.new([class_name] [, member_name]+)                        -> StructClass
+  #    Struct.new([class_name] [, member_name]+, keyword_init: true)    -> StructClass
+  #    Struct.new([class_name] [, member_name]+) {|StructClass| block } -> StructClass
+  #    StructClass.new(value, ...)                                      -> object
+  #    StructClass[value, ...]                                          -> object
+  #
+  #  The first two forms are used to create a new Struct subclass +class_name+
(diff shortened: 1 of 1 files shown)


froydnj (approved):

jez (changes requested):
This method does not exist as an instance method on Struct.

This is simply a limitation in Sorbet’s support for Struct.new.

The result of Struct.new is not an instance of Struct but rather a T.class_of(Struct). Fixing this properly so that the name method is available requires overriding the definition of def self.new in Struct to return something other than T.attached_class (probably T.class_of(Struct))

dduugg:
> This method does not exist as an instance method on Struct.
> 
> This is simply a limitation in Sorbet’s support for Struct.new.
> 
> The result of Struct.new is not an instance of Struct but rather a T.class_of(Struct). Fixing this properly so that the name method is available requires overriding the definition of def self.new in Struct to return something other than T.attached_class (probably T.class_of(Struct))

Ooh, good catch, and tricky, because the `Struct` that's returned has a different constructor, which returns an instance, e.g.:
```
Struct.new('Foo', :foo, :bar).new(1, 2)
=> #<struct Struct::Foo foo=1, bar=2>
```
Do you have suggestions for capturing this most effectively with Sorbet?

dduugg:
I've updated the PR so that `Struct.new` returns `T.untyped`.

(There is also an instance method `new` [elsewhere](https://github.com/sorbet/sorbet/blob/f6d54ab/rbi/core/struct.rbi#L209) in the file, which does something similar, but may be in error)

PR #7639, opened 2024-01-31
Add json/add/exception methods to json.rbi

dduugg:
Adds methods defined in https://github.com/flori/json/blob/master/lib/json/add/exception.rb to json.rbi

### Motivation

https://github.com/Homebrew/brew/issues/16557

### Test plan

Tested in above repo

PR #7652, opened 2024-02-05
RBI: make Pathname#mkdir match Dir.mkdir's signature for the optional mode argument

jmhertlein:
I updated the RBI signature to allow the optional `mode` parameter to not be supplied, matching how the same call is set in `Dir.mkdir` in `core/dir.rbi` (which Pathname defers to).

### Motivation

Sorbet was reporting an error when it is actually a valid usage. 

e.g.

```
irb(main):002:0> p = Pathname.new("/tmp/foo")
=> #<Pathname:/tmp/foo>
irb(main):003:0> p.mkdir
=> 0
irb(main):004:0> p.exist?
=> true
```

While in vscode with latest sorbet, this reports:

```
Not enough arguments provided for method `Pathname#mkdir`. Expected: `1`, got: `0`
```

### Test plan

I looked at the most recent RBI change (https://github.com/sorbet/sorbet/pull/7639 at the time of writing) and it looks like you run some internal testing. Let me know if you want me to do anything on my end for this PR.

### DIFF
--- a/rbi/stdlib/pathname.rbi
+++ b/rbi/stdlib/pathname.rbi
@@ -971,7 +971,7 @@ class Pathname < Object
     )
     .returns(Integer)
   end
-  def mkdir(p1); end
+  def mkdir(p1=T.unsafe(nil)); end
 
   # Creates a full path, including any intermediate directories that don't yet
   # exist.


jez:

→ https://go/builds/bui_PVcImbPt33Q2VD
→ https://go/builds/bui_PVcIqnd001XXMK
→ https://go/builds/bui_PVcIPQJgwTWILN

jez (approved):
Thanks!

PR #7653, opened 2024-02-05
Add doc guiding users how to debug crashes in Sorbet

jez:
### Test plan

See included automated tests.


ilyailya (approved):
Looks good to me! 

Would it make sense to add a paragraph about `stopInDebugger()`? Not sure if it worth including in the doc, but it helped me and was unobvious when I just started. I used to put breakpoints with `b path/to/file.cc:42`, which is much more of a hustle

PR #7654, opened 2024-02-06
Factor out sigil parsing code

jez:
### Motivation

Removes code duplication.

### Test plan

See included automated tests.


core/CompiledLevel.h
```diff
@@ -16,6 +17,24 @@ enum class CompiledLevel : uint8_t {
     // This file should be compiled.
     True = 2,
 };
-}
+
+template <> class SigilTraits<CompiledLevel> {
+public:
+    static constexpr CompiledLevel NONE = CompiledLevel::None;
+
+    static constexpr std::string_view SIGIL_PREFIX = "typed:";
```
froydnj, line 25: This looks like too much cut-and-paste.

froydnj (changes requested):
`:c++-intensifies:`

core/Files.cc
```diff
@@ -17,211 +18,75 @@ namespace {
...
-        if (source[comment_start] != '#') {
-            ++start;
-            continue;
-        }
 
-        start += 6;
-        while (start < source.size() && source[start] == ' ') {
-            ++start;
-        }
+            start += 6;
```
froydnj, line 46:
```suggestion
            start += SigilTraits<Sigil>::PREFIX.size();
```

I think this is more correct.

froydnj (approved):
I think you still have a bug, below; I'm going to assume that the below suggestion (or something like it, plz syntax-validate my suggestions, Github) is going to fix things.

PR #7655, opened 2024-02-06
Tighten type of `Enumerator#with_index` with block

marianosimone:
### Motivation
Better type information when calling this method

### Test plan
- Existing tests
- Running against Stripe's codebase


jez:
You are finding the list of greatest hits

#5258

jez (changes requested):
^ suggested test case is in the linked issue above.

PR #7656, opened 2024-02-06
Add error when branching on `void` value

### LINKED ISSUE #2913: Emit error if void used in conditional branch

jez:
### Motivation

Fixes #2913

### Test plan

See included automated tests.


neilparikh (commented):
Typecheck failures over sorbet-runtime and test failures.

jez:

→ https://go/builds/bui_PW1Wf2l7CS061Y
→ https://go/builds/bui_PW1WGkZiOObMbF
→ https://go/builds/bui_PW1WJjZIzJqGPA

jez:

→ https://go/builds/bui_PW2Cek9WRYBQWg
→ https://go/builds/bui_PW2C7vxkpOlTqQ
→ https://go/builds/bui_PW2CpUrkqfOKmy

jez:

→ https://go/builds/bui_PW2JkbWyS3ZLBE
→ https://go/builds/bui_PW2JU1dIlWU6Mf
→ https://go/builds/bui_PW2JWePMM4qsWS

jez:

→ https://go/builds/bui_PWLCMXlE2Fk1Gp
→ https://go/builds/bui_PWLCmZzS9knt8c
→ https://go/builds/bui_PWLCpKrD8VHMUa

→ https://go/builds/bui_PWLNzcpRXmdTaN

core/TypePtr.h
```diff
@@ -255,6 +255,18 @@ class TypePtr final {
 
     bool hasUntyped() const;
 
+    // This is a hacky, unprincipled method that simply looks for void inside the type at the top
+    // level (not going into AppliedTypes).
```
neilparikh, line 259: Are there cases where this would not catch `void` (because it doesn't go into `AppliedType`s)? If so, can you write a test case for those?

gems/sorbet-runtime/lib/types/props/decorator.rb
```diff
@@ -293,7 +293,7 @@ def prop_validate_definition!(name, cls, rules, type)
       cls: PropTypeOrClass,
       rules: Rules,
     )
-    .void
+    .returns(T.anything)
```
neilparikh, line 296: I assume this means that this method always just returned a truthy value before. Do we know what was broken before because of this?

core/TypePtr.cc
```diff
@@ -221,6 +221,44 @@ bool TypePtr::hasUntyped() const {
...
+        case Tag::FloatLiteralType:
+        case Tag::SelfType:
+        case Tag::AliasType:
+        case Tag::SelfTypeParam:
+        case Tag::LambdaParam:
+        case Tag::MetaType:
+        case Tag::BlamedUntyped:
+        case Tag::UnresolvedAppliedType:
+        case Tag::UnresolvedClassType:
+            // These cannot have untyped.
```
neilparikh, line 238:
```suggestion
            // These cannot have void.
```

neilparikh (approved):

gems/sorbet-runtime/lib/types/props/decorator.rb
jez: Rather: this method was `.checked(:never)`, so the underlying value it returned was being propagated through, instead of being replaced with the void singleton.

Issue #7532, opened 2023-12-12
slow type check performance for large T::Enums


technicalpickles:
I was chatting with @froydnj about our app's fairly slow `srb tc` performance, and he was able to narrow it down to a `T::Enum` that has some 1601 values 😳 

Specifically, we had several files that were doing `case` and `if <value> === SomeEnum` statements with these T::Enums, and they take upwards of 4s each. From the output of `srb tc --counters`:

```
typecheckOne.min :           1.00 ms
typecheckOne.max :        4456.00 ms
typecheckOne.avg :          16.30 ms
```

From @froydnj:

> once we see:
>
> ```ruby
> if @letter_combo != LetterCombo::Bye
> ```
> we drop `LetterCombo::Bye` from `@letter_combo`'s type (which I assume is `T.nilable(LetterCombo)`), which means we construct a 6162-element `T.any` (remember it can be `nil`), and then we need to check the type of serialize on 6161 elements of that type, which is not fast
>
>we essentially do the same thing with case statements:
>

> ```ruby
> case x
> when A ...
> when B ...
> when C ... 
> when D ...
> # and so forth
> ```

> if `x` is a `T::Enum`, after sorbet checks it for `A`-ness, we know that `A` can be dropped from the type.  and so on for the rest of the when clauses.  normally this is not a problem, but with a 6000-element `T::Enum`, that means we’re constructing a 5999-element type for checking against `B`, a  5998-element type for checking against `C`, and so forth

I was able to make an approximate reproduction, but the runtime isn't as bad as our app: https://github.com/technicalpickles/sorbet-giant-tenum-example . I tried adding that to sorbet.run but it was really struggling 😅

PR #7659, opened 2024-02-07
Accept a list of classes in dropSubtypesOf

jez:
### Motivation

Should improve performance slightly, because dropSubtypesOf is called for
every conditional branch which follows the truthy case in Sorbet, and now
that we accept a list of types, we only have to descend into large types
once, instead of twice.

### Test plan

See included automated tests.

### DIFF
--- a/compiler/IREmitter/sends.cc
+++ b/compiler/IREmitter/sends.cc
@@ -151,7 +151,7 @@ InlinedVector<ApplicableIntrinsic, NUM_INTRINSICS> applicableIntrinsics(MethodCa
 
         auto potentialClasses = symbolBasedIntrinsic->applicableClasses(mcctx.cs);
         for (auto &c : potentialClasses) {
-            InlinedVector<ClassOrModuleRef, 1> toDrop{c};
+            auto toDrop = absl::MakeSpan(&c, 1);
             auto leftType = core::Types::dropSubtypesOf(mcctx.cs, remainingType, toDrop);
 
             if (leftType == remainingType) {
--- a/compiler/IREmitter/sends.cc
+++ b/compiler/IREmitter/sends.cc
@@ -151,7 +151,8 @@ InlinedVector<ApplicableIntrinsic, NUM_INTRINSICS> applicableIntrinsics(MethodCa
 
         auto potentialClasses = symbolBasedIntrinsic->applicableClasses(mcctx.cs);
         for (auto &c : potentialClasses) {
-            auto leftType = core::Types::dropSubtypesOf(mcctx.cs, remainingType, c);
+            InlinedVector<ClassOrModuleRef, 1> toDrop{c};
+            auto leftType = core::Types::dropSubtypesOf(mcctx.cs, remainingType, toDrop);
 
             if (leftType == remainingType) {
                 continue;
--- a/core/types/types.cc
+++ b/core/types/types.cc
@@ -198,9 +198,8 @@ TypePtr Types::dropSubtypesOf(const GlobalState &gs, const TypePtr &from, absl::
             } else if (c.symbol.data(gs)->isClass() && absl::c_all_of(klasses, [&](auto klass) {
                            return klass.data(gs)->isClass() && !klass.data(gs)->derivesFrom(gs, c.symbol);
                        })) {
-                // We have two classes (not modules), and if the class we're
-                // removing doesn't derive from `c`, there's nothing to do,
-                // because of ruby having single inheritance.
+                // We have two classes (not modules), and if all of the the classes we're removing
+                // don't derive from `c`, there's nothing to do because Ruby has single inheritance.
                 result = from;
             } else if (cdata->flags.isSealed && (cdata->flags.isAbstract || cdata->isModule())) {
                 auto subclasses = cdata->sealedSubclassesToUnion(gs);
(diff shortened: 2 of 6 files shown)


jez:

→ https://go/builds/bui_PW13JYznLWIUX5
→ https://go/builds/bui_PW13I8dLGoSC0I
→ https://go/builds/bui_PW13DeO40xiDvd

jez:

→ https://go/builds/bui_PW28HsoWUqNGoh
→ https://go/builds/bui_PW28vG3lmbFiWS
→ https://go/builds/bui_PW28d9rVFLbhg6

compiler/IREmitter/sends.cc
```diff
@@ -151,7 +151,8 @@ InlinedVector<ApplicableIntrinsic, NUM_INTRINSICS> applicableIntrinsics(MethodCa
 
         auto potentialClasses = symbolBasedIntrinsic->applicableClasses(mcctx.cs);
         for (auto &c : potentialClasses) {
-            auto leftType = core::Types::dropSubtypesOf(mcctx.cs, remainingType, c);
+            InlinedVector<ClassOrModuleRef, 1> toDrop{c};
```
froydnj, line 154:
```suggestion
            auto toDrop = absl::make_span(&c, 1);
```

or something like that?

core/types/types.cc
```diff
@@ -272,8 +281,13 @@ bool Types::canBeFalsy(const GlobalState &gs, const TypePtr &what) {
...
     TypePtr result;
     typecase(
-        what, [&](const ClassType &c) { result = Types::dropSubtypesOf(gs, from, c.symbol); },
-        [&](const AppliedType &c) { result = Types::dropSubtypesOf(gs, from, c.klass); },
+        what,
+        [&](const ClassType &c) {
+            result = Types::dropSubtypesOf(gs, from, InlinedVector<ClassOrModuleRef, 1>{c.symbol});
+        },
+        [&](const AppliedType &c) {
+            result = Types::dropSubtypesOf(gs, from, InlinedVector<ClassOrModuleRef, 1>{c.klass});
```
froydnj, line 289: Can we construct spans directly here rather than indirecting through `InlinedVector`?

core/types/types.cc
```diff
@@ -333,7 +347,8 @@ TypePtr Types::tClass(const TypePtr &attachedClass) {
 }
 
 TypePtr Types::dropNil(const GlobalState &gs, const TypePtr &from) {
-    return Types::dropSubtypesOf(gs, from, Symbols::NilClass());
+    static InlinedVector<ClassOrModuleRef, 1> toDrop{core::Symbols::NilClass()};
+    return Types::dropSubtypesOf(gs, from, toDrop);
```
froydnj, line 351: Optionally could just `auto toDrop = absl::make_span(&core::Symbols::NilClass(), 1);` or something like that, which ought to be epsilon faster.

core/types/types.cc
```diff
@@ -186,51 +192,54 @@ TypePtr Types::dropSubtypesOf(const GlobalState &gs, const TypePtr &from, ClassO
...
+                       })) {
                 // We have two classes (not modules), and if the class we're
                 // removing doesn't derive from `c`, there's nothing to do,
                 // because of ruby having single inheritance.
                 result = from;
             } else if (cdata->flags.isSealed && (cdata->flags.isAbstract || cdata->isModule())) {
                 auto subclasses = cdata->sealedSubclassesToUnion(gs);
                 ENFORCE(!Types::equiv(gs, subclasses, from), "sealedSubclassesToUnion about to cause infinite loop");
-                result = dropSubtypesOf(gs, subclasses, klass);
+                result = dropSubtypesOf(gs, subclasses, klasses);
```
froydnj, line 208: How would you feel about -- not in this PR -- optimizing this so that we could get a vector of `ClassOrModuleRef` out of a sealed class, and then we would not recurse into `dropSubtypesOf`, but rather filter the vector based on `klasses`, and then construct the resulting union directly from the classes we didn't filter?

I don't think it entirely solves #7532, but it would be more efficient?

core/types/types.cc
```diff
@@ -186,51 +192,54 @@ TypePtr Types::dropSubtypesOf(const GlobalState &gs, const TypePtr &from, ClassO
...
+                                      [&](auto klass) { return c.symbol == klass || c.derivesFrom(gs, klass); })) {
                 result = Types::bottom();
-            } else if (c.symbol.data(gs)->isClass() && klass.data(gs)->isClass() &&
-                       !klass.data(gs)->derivesFrom(gs, c.symbol)) {
+            } else if (c.symbol.data(gs)->isClass() && absl::c_all_of(klasses, [&](auto klass) {
+                           return klass.data(gs)->isClass() && !klass.data(gs)->derivesFrom(gs, c.symbol);
+                       })) {
                 // We have two classes (not modules), and if the class we're
                 // removing doesn't derive from `c`, there's nothing to do,
                 // because of ruby having single inheritance.
```
froydnj, line 203: This comment could stand a small tweak now that we have multiple classes that we're dropping.

froydnj (commented):
I like this change!

core/types/types.cc
jez: Sounds worth a shot to me.

jez:
I tested this on Stripe's codebase (admittedly, before the changes to replace InlinedVector with MakeSpan) and saw essentially no change. But regardless, I figure it's cleaner this way so I'm in favor of leaving it.

froydnj (approved):

PR #7673, opened 2024-02-08
Test sorbet-runtime on 3.3.0

jez:
### Motivation

I want to get my changes in #7439 across the line, which have some logic which
changes in Ruby 3.2. I figure I may as well just get our tests running on a
recent version of Ruby.

### Test plan

See included automated tests.


.ruby-version
```diff
@@ -1 +1 @@
-2.7.2
+3.1.2
```
jez, line 1: While we still test in CI on 2.7, I figure it's time to bump this while we're at it.

Maybe some local scripts will break? But we can address those as they happen.

neilparikh (approved):

PR #7681, opened 2024-02-09
close over the class's decorator directly

froydnj:
### Motivation

This is 15-20% faster on Sorbet's benchmark suite, though admittedly it's already the slow path.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_PX5BjlfniOKWmM
→ https://go/builds/bui_PX5B7HlVoDEkaW
→ https://go/builds/bui_PX5BlM4PsdQefs

→ https://go/builds/bui_PX5GFulznYiq4Z

ilyailya:

→ https://go/builds/bui_PX5Af4ONxptZBz
→ https://go/builds/bui_PX5ABV4FHVRowd
→ https://go/builds/bui_PX5ALKRV6XChvB

→ https://go/builds/bui_PX5HVrDkYjq8Af

froydnj:
One thing to note is that this fixes the decorator at prop definition time, rather than at prop access time.  So if you had a `class Base` that defined `prop :slot, String` and `class Derived < Base` that had a different decorator than `Base` (!), accesses to `slot` on an instance of `Derived` used to use `Derived`'s decorator; after this PR, they would use `Base`'s decorator, which is a breaking change.

This seems pretty esoteric, and is not a problem for Stripe's codebase, but it is worth calling out.

ilyailya (approved):

froydnj:

→ https://go/builds/bui_SBUOWOVonB9zTF
→ https://go/builds/bui_SBUOVbpBMgUSEl
→ https://go/builds/bui_SBUPNAIoZbdo6j

→ https://go/builds/bui_SBUYf1Gk3km1ze

jez (approved):
Shocked that this isn’t something some random piece of Stripe code depends on

froydnj:
> Shocked that this isn’t something some random piece of Stripe code depends on

`:big-same:`

PR #1397, opened 2019-07-26
runtime: Fix sig block deletion race

XrXr:
When two threads commit to calling the the same wrapper and
one thread deletes the sig block from the `sig_wrappers` hash
before the other one does the look up, sorbet-runtime raises.

Fix this issue by checking whether the sig block was already
evaluated when it is missing from the sig_wrappers hash.

Fixes #1386

### Motivation
This is making CI flaky for us.

### Test plan

See included automated tests.

PR #7686, opened 2024-02-12
Use `Concurrent::Hash`, if available, for sig related storage

paracycle:
This PR makes access to `@sig_wrappers` and `@signatures_by_method` ivars of `T::Private::Methods` class less racy for alternative Ruby implementations by making them be `Concurrent::Hash` instances if `Concurrent::Hash` is defined.

Even if they have the constant defined for an unrelated reason, this should ensure that common users of the library who run on CRuby have no performance penalty for using `Concurrent::Hash`, since the [implementation of that class on CRuby is exactly the built-in `Hash` class](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent-ruby/concurrent/hash.rb#L16-L24).

/cc @nirvdrum @rwstauner 

### Motivation

Unlike the case on CRuby, alternative Ruby implementations like JRuby/TruffleRuby do not have a GVL, which makes access to shared data structure inherently racy. In most cases, these races are not visible, but concurrent access to `@sig_wrappers` has already been problematic in Sorbet runtime on CRuby [before](https://github.com/sorbet/sorbet/pull/1397). Depending on access patterns and/or implementation of the underlying `Hash` implementation, errors can exhibit themselves in strange ways, especially on alternative Ruby implementations.

We've recently been seeing hash related crashes while accessing `@sig_wrappers.delete(key)` inside `T::Private::Methods.run_sig_block_for_key` when running one of our applications on TruffleRuby with the [new `CompactHashStore` Hash implementation](https://github.com/oracle/truffleruby/pull/3172).

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/_methods.rb
@@ -3,8 +3,13 @@
 
 module T::Private::Methods
   @installed_hooks = {}
-  @signatures_by_method = {}
-  @sig_wrappers = {}
+  if defined?(Concurrent::Hash)
+    @signatures_by_method = Concurrent::Hash.new
+    @sig_wrappers = Concurrent::Hash.new
+  else
+    @signatures_by_method = {}
+    @sig_wrappers = {}
+  end
   @sigs_that_raised = {}
   # stores method names that were declared final without regard for where.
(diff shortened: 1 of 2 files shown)


jez (approved):

paracycle:
@jez is this good to be merged? have you had a chance to run it against the Stripe codebase?

PR #7687, opened 2024-02-12
Encapsulate backtrace location searching logic to a single method and filter `<internal:*>` paths

paracycle:
This PR encapsulates caller location search logic into a singleton method, and applies a filer for any caller location `path` that starts with `<internal:` since those paths represent internal Ruby methods implemented in Ruby.

### Motivation

In order to provide compatibility with TruffleRuby that implements a lot of its core library methods in Ruby, and for future versions of CRuby that is increasingly doing the same, we need to be able to filter all backtrace locations where the `path` starts with `<internal:`.

[This gist](https://gist.github.com/eregon/912e6359e83781c5fa1c638d3768c526) shows the current state of the methods implemented in Ruby in CRuby, JRuby and TruffleRuby.

Most recently [CRuby started implementing `Array#each` in Ruby](https://github.com/ruby/ruby/pull/9533), making usages of `each` visible in backtraces with an `<internal:array>` path. This means that in order to be compatible with CRuby 3.4, Sorbet runtime needs to start filtering out backtrace locations that start with `<internal:`.

### Test plan

Existing tests should be unaffected.


neilparikh:

→ https://go/builds/bui_PYdDwvLAvDkKCG
→ https://go/builds/bui_PYdDm9rRd9VQTX
→ https://go/builds/bui_PYdDnmhdMOR66F

→ https://go/builds/bui_PYdQTq0ho4A5iW

paracycle:
@neilparikh I've updated this PR to use the more performant version of backtrace scanning for Ruby 3.2+ as suggested in https://github.com/sorbet/sorbet/pull/7710

Do you mind testing it on the Stripe codebase again? Also, curious to know if the previous run was green.

neilparikh:
@paracycle The previous run didn't finish properly because of some problems with our script for testing on Stripe's codebase. I'm working on fixing it, I'll fire off another run once I've fixed it.

jez:

→ https://go/builds/bui_PbYl4UlHxALmd3
→ https://go/builds/bui_PbYlhiCyQgEOdH
→ https://go/builds/bui_PbYlbCYg7e8a5K

→ https://go/builds/bui_PbYtTGOf0VMA3U

jez (approved):

PR #7696, opened 2024-02-14
Replace `--secondary-test-package-namespace` with "test-only packages"

jez:
### Motivation

`--secondary-test-package-namespaces` is obsolete, and we can replace with a concept of test-only packages.

I suspect that we could even go one step further and say that only code in `.test.rb` files can and needs to use the `Test::` prefix, meaning that we would not need to special case test-only packages. But that'd be a change for the future, and would be more restrictive than what currently exists.

### Test plan

See included automated tests.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -969,22 +969,6 @@ void readOptions(Options &opts,
             }
         }
 
-        if (raw.count("secondary-test-package-namespaces")) {
-            if (!opts.stripePackages) {
-                logger->error("--secondary-test-package-namespaces can only be specified in --stripe-packages mode");
-                throw EarlyReturnWithCode(1);
-            }
-            std::regex nsValid("[A-Z][a-zA-Z0-9]+");
-            for (const string &ns : raw["secondary-test-package-namespaces"].as<vector<string>>()) {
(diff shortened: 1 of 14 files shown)


jez:

→ https://go/builds/bui_PYdHQr2kG5tcqI
→ https://go/builds/bui_PYdH9LCWbuNssi
→ https://go/builds/bui_PYdITwhgQtkxQZ

jez:

→ https://go/builds/bui_PYe3zhGWN6uUD3
→ https://go/builds/bui_PYe3keYT4vJtVH
→ https://go/builds/bui_PYe3sXQBtMqPpQ

jez:

→ https://go/builds/bui_PYf2zZjG7shR93
→ https://go/builds/bui_PYf2SOkMBTrvpi
→ https://go/builds/bui_PYf2iffczZdemM

main/options/options.cc
jez, line 1: I will delete the definition of these options in the future after we're no longer using them in Stripe's codebase.

aadi-stripe (approved):

neilparikh (approved):

PR #7698, opened 2024-02-14
Add signatures for `RubyVM::YJIT`

### LINKED ISSUE #7697: Missing signatures for RubyVM::YJIT
### Problem

Sorbet includes signatures for `RubyVM` ([here](https://github.com/sorbet/sorbet/blob/master/rbi/core/ruby_vm.rbi)) but not for [`RubyVM::YJIT`](https://ruby-doc.org/3.3.0/RubyVM/YJIT.html).

Future versions of Rails will ship with [this initializer](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/config/initializers/enable_yjit.rb.tt). By default, in a strict-typed Sorbet project, this file produces the following error:
> Unable to resolve constant `YJIT` (fix available) ([5002](https://srb.help/5002))

I worked around this by adding a shim:
```ruby
class RubyVM
  module YJIT
    sig { returns(T::Boolean) }
    def self.enabled?; end

    sig { params(stats: T::Boolean).returns(T::Boolean) }
    def self.enable(stats: false); end
  end
end
```

### Proposed solution

Add signatures for all `RubyVM::YJIT` methods to Sorbet's stdlib signatures.

olivier-thatch:
Add signatures for `RubyVM::YJIT`

### Motivation
Closes #7697.

### Test plan
N/A

### DIFF
--- a/rbi/core/ruby_vm.rbi
+++ b/rbi/core/ruby_vm.rbi
@@ -791,3 +791,79 @@ class RubyVM::InstructionSequence < Object
   # ```
   def self.of(_); end
 end
+
+# This module allows for introspection of YJIT, CRuby's just-in-time compiler.
+# Everything in the module is highly implementation specific and the API might
+# be less stable compared to the standard library.
+#
+# This module may not exist if YJIT does not support the particular platform
+# for which CRuby is built.
+module RubyVM::YJIT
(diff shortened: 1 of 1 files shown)


jez (approved):
Thanks!

PR #7710, opened 2024-02-16
Fix Ruby 3.4-dev compatibility

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/19117
Ref: https://bugs.ruby-lang.org/issues/16495

Now backtrace use constisten `'` quotes rather than a mix of backtick and quote.

Also the label include the class or module names.

But in most case dealing with `Backtrace::Location` objects is much more performant and clear than to match a regexp on the full rendered frame. So aside from Ruby compatibility, this is just better code.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/sealed.rb
+++ b/gems/sorbet-runtime/lib/types/private/sealed.rb
@@ -5,8 +5,8 @@ module T::Private::Sealed
   module NoInherit
     def inherited(child)
       super
-      caller_loc = T::Private::CallerUtils.find_caller {|loc| !loc.to_s.match(/in `inherited'$/)}
-      T::Private::Sealed.validate_inheritance(caller_loc&.to_s, self, child, 'inherited')
+      caller_loc = T::Private::CallerUtils.find_caller {|loc| loc.base_label != 'inherited'}
+      T::Private::Sealed.validate_inheritance(caller_loc, self, child, 'inherited')
       @sorbet_sealed_module_all_subclasses << child
     end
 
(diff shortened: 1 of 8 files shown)


gems/sorbet-runtime/lib/types/private/sealed.rb
```diff
@@ -5,8 +5,8 @@ module T::Private::Sealed
   module NoInherit
     def inherited(child)
       super
-      this_line = Kernel.caller.find {|line| !line.match(/in `inherited'$/)}
-      T::Private::Sealed.validate_inheritance(this_line, self, child, 'inherited')
+      this_line = Kernel.caller_locations.find {|loc| loc.base_label != 'inherited'}
```
casperisfine, line 8: Would be nice to use `Thread.each_caller_location` here too, but in the grand scheme of things the performance shouldn't be huge because the stack generally isn't that big during boot time.

jez (commented):
Mostly looks good, but you've got some test failures.

casperisfine:
> you've got some test failures.

My bad, I forgot to test on 2.7 which doesn't have `Symbol#name`. Should go green now.

casperisfine:
Also, it probably make sense to merge https://github.com/sorbet/sorbet/pull/7687 first, and then I can rebase on top.

jez:
> Also, it probably make sense to merge #7687 first, and then I can rebase on top.

I've merged #7687, feel free to rebase

casperisfine:
Thank you!

Rebased and tested locally:

```
$ ruby -v
ruby 3.4.0dev (2024-02-16T07:13:45Z master bc7266c5ce) [arm64-darwin23]
$ bundle exec rake
/Users/byroot/src/github.com/Shopify/sorbet/gems/sorbet-runtime/test/types/method_validation.rb:573: warning: key :_ is duplicated and overwritten on line 573
Inspecting 162 files
..................................................................................................................................................................

162 files inspected, no offenses detected
Run options: --seed 19752

# Running:

......................S...................................................................S....................................................................S......................................................................................................................................................................................................................................................................................................S...................................................................................................S.............S..........S...................................S........................................................................................................................................................................................................................................................S...............................

Finished in 0.453960s, 1973.7422 runs/s, 6163.5386 assertions/s.

896 runs, 2798 assertions, 0 failures, 0 errors, 9 skips

```

casperisfine:
👋 sorry to be a bother, but could this be merged? It's been over a week now that we no longer have visibility over ruby-head compatibility :/

froydnj (approved):

PR #7718, opened 2024-02-22
Build and release on linux-aarch64

### LINKED ISSUE #4119: Support for Ubuntu Linux on ARM64 (aarch64)
### Problem

We are starting to need to use ARM based Linux (using Ubuntu) but we can't use Sorbet since there is not a binary version of Sorbet that will run. When performing a `bundle install` that includes `sorbet` we get the following error:

```
Unable to find a spec satisfying sorbet-static (= 0.5.5912) in the set. 
Perhaps the lockfile is corrupted? Found sorbet-static (0.5.5912-x86_64-linux), 
sorbet-static (0.5.5912-universal-darwin-19) that did not match the current platform.
```

This is running on Ubuntu 20 LTS

```
Linux local 5.4.0-70-generic #78-Ubuntu SMP Fri Mar 19 13:29:32 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux
```

### Proposed solution

Add a linux `aarch64` pre-built sorbet-static.

DazWorrall:
Closes #4119 

Build and release on aarch64 linux, uses a dedicated target so the x86_64 build keeps its optimisations.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -107,6 +107,9 @@ build:untyped-blame --copt=-DTRACK_UNTYPED_BLAME_MODE
 # harden: mark relocation sections read-only
 build:release-linux --linkopt=-Wl,-z,relro,-z,now
 build:release-linux --config=lto-linux --config=release-common
+# Separate config for aarch64, so x86_64 can be differently optimized
+build:release-linux-aarch64 --linkopt=-Wl,-z,relro,-z,now
+build:release-linux-aarch64 --config=lto-linux --config=release-common
 
 # This is to turn on vector instructions where available.
 # We used to do this unconditionally, but Rosetta 2 doesn't translate all vector instructions well.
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -107,6 +107,9 @@ build:untyped-blame --copt=-DTRACK_UNTYPED_BLAME_MODE
 # harden: mark relocation sections read-only
 build:release-linux --linkopt=-Wl,-z,relro,-z,now
 build:release-linux --config=lto-linux --config=release-common
+# Separate config for aarch64, so x86_64 can be differently optimized
+build:release-linux-aarch64 --linkopt=-Wl,-z,relro,-z,now
+build:release-linux-aarch64 --config=lto-linux --config=release-common
```
DazWorrall, line 112: There is perhaps some DRYing up to do here, but let's get the new builds right first and not touch the existing `release-linux` config.

jez: I agree, I'm going to take a pass over these options.

jez (approved):

PR #7721, opened 2024-02-23
Fix sig for `YJIT.enable`

andyw8:
### Motivation

As per the comment above, this can take `:quiet` as an argument.

### Test plan

n/a


froydnj (approved):
Thanks!

PR #7726, opened 2024-02-28
Reimplement a4a06f8016 in terms of symbol ownership

jez:
### Motivation

Now that we have proper symbol-based package ownership, we don't need to
use this hack which relies on the canonical loc of a Symbol being the
source of truth for that symbol's package.

### Test plan

The test included in a4a06f8016 covers this refactor.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -225,9 +225,8 @@ public:
     }
 
     bool ownsSymbol(const core::GlobalState &gs, core::SymbolRef symbol) const {
-        auto file = symbol.loc(gs).file();
-        auto &pkg = gs.packageDB().getPackageForFile(gs, file);
-        return this->mangledName() == pkg.mangledName();
+        auto &pkg = gs.packageDB().getPackageNameForSymbol(gs, symbol);
+        return this->mangledName() == pkg;
     }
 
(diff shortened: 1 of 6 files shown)


jez:

→ https://go/builds/bui_PdtSj8FKRm9Mhf
→ https://go/builds/bui_PdtS1heIhGdRBY
→ https://go/builds/bui_PdtS01wlHKqo4T

packager/packager.cc
```diff
@@ -209,9 +209,8 @@ class PackageInfoImpl final : public core::packages::PackageInfo {
     }
 
     bool ownsSymbol(const core::GlobalState &gs, core::SymbolRef symbol) const {
-        auto file = symbol.loc(gs).file();
-        auto &pkg = gs.packageDB().getPackageForFile(gs, file);
-        return this->mangledName() == pkg.mangledName();
+        auto &pkg = gs.packageDB().getPackageNameForSymbol(gs, symbol);
```
neilparikh, line 212: Two minor nits (pick either or neither to do):
- inline `pkg` and change the `return` to be `return this->mangledName() == gs.packageDB().getPackageNameForSymbol(gs, symbol);`
- rename `pkg` to `pkgName`

neilparikh (approved):

PR #7728, opened 2024-02-28
Make `locs` methods return `absl::Span`

jez:
### Motivation

Package symbols will not have a list of locs, so it's a little easier if I can
get away with just storing a `Loc` instead of having to store an
`InlinedVector<Loc, 1>`, because all `SymbolRef` kinds need to implement the
`locs` method, and right now that has to return a reference to a vector, which
would have to outlive the reference to the `Package` symbol (and thus be managed
by the `Package` symbol itself).

### Test plan

Existing tests


froydnj (approved):
👍

PR #7730, opened 2024-02-28
Count files under `spec/` as test files

KaanOzkan:
Is this acceptable given this code path is used for more than highlight T.untyped feature?

### Motivation

Rspec style tests put tests under the `spec` folder

### Test plan

I couldn't find tests for this file.


jez:
Hey unfortunately we can't take this patch as implemented—we'll probably have to make a command line for this, which makes it into a bit of a larger change (because the place where `isTestPath` is called does not have access to an `Options` nor `GlobalState` argument).

Stripe's codebase has about 50 files inside `/spec/` where the presence of `/spec/` in the name does not mean that the file is a test file.

If we can make what counts as a test configurable per project I'd be happy to review it.

jez (changes requested):
^

KaanOzkan:
@jez understandable. Would you be okay with using a separate flag for this or doing the check inline in `infer.cc` (`path` seems to be a public field)?

I think this use case is common enough that it's better to support inside Sorbet rather than individual project configurations.

jez:
I don't quite know what you mean by saying both

> Would you be okay with using a separate flag for this

and also

> I think this use case is common enough that it's better to support inside Sorbet rather than individual project configurations.

Could you clarify what you're suggesting?

KaanOzkan:
I pushed to 2 commits to show the different solutions. I thought you meant this PR is an issue due to the flags usage with the `--stripe-packages` feature. Instead if the problem is that you want the `everywhere-but-tests` flag to still include `spec/` folder in Stripe then yes it'd have to be customizable. With either an opt in or an opt out CLI flag.

PR #7735, opened 2024-02-29
Add some missing encodings

kddnewton:
These encodings are missing.

### Motivation

I'm trying to use these encodings.

### Test plan

I didn't include automated tests because this is unlikely to change. I did verify that this covers all existing encodings, however by running:

```ruby
puts Encoding.constants.select { |constant| Encoding.const_get(constant).is_a?(Encoding) }.sort.map { |name| "  #{name} = T.let(T.unsafe(nil), Encoding)" }
```

and copy-pasting it into the file.


froydnj (approved):
Thanks!

PR #7737, opened 2024-03-01
Pre-resolve the PackageSpec superclass in a package file

jez:
### Motivation

This preëmpts what would have happened in resolver. Moving it sooner means
that we can use this to to more quickly figure out whether a ClassDef is
the PackageSpec class def in namer, when other constant lits will not have
been resolved.

### Test plan

Existing tests


froydnj (approved):
Less work is more better.

PR #7742, opened 2024-03-04
add an `override` autocorrect

### LINKED ISSUE #5235: Automatically add `override.` to a sig when it would be required
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Parent%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7Boverridable.void%7D%0A%20%20def%20foo%3B%20end%0Aend%0A%0Amodule%20IChild%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AHelpers%0A%20%20abstract!%0A%0A%20%20sig%20%7Babstract.void%7D%0A%20%20def%20bar%3B%20end%0Aend%0A%0Aclass%20Child%20%3C%20Parent%0A%20%20include%20IChild%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20foo%3B%20end%0A%0A%20%20sig%20%7Bvoid%7D%0A%20%20def%20bar%3B%20end%0Aend)

```ruby
# typed: true

class Parent
  extend T::Sig
  sig {overridable.void}
  def foo; end
end

module IChild
  extend T::Sig
  extend T::Helpers
  abstract!

  sig {abstract.void}
  def bar; end
end

class Child < Parent
  include IChild
  sig {void}
  def foo; end

  sig {void}
  def bar; end
end
```

#### Observed output

```
editor.rb:21: Method Child#foo overrides an overridable method Parent#foo but is not declared with override. https://srb.help/5051
    21 |  def foo; end
          ^^^^^^^
    editor.rb:6: defined here
     6 |  def foo; end
          ^^^^^^^

editor.rb:24: Method Child#bar implements an abstract method IChild#bar but is not declared with override. https://srb.help/5051
    24 |  def bar; end
          ^^^^^^^
    editor.rb:15: defined here
    15 |  def bar; end
          ^^^^^^^
Errors: 2
```

#### Expected behavior

There should be an autocorrect in both these cases that offers to insert the `override.` into the sig for you.

- - -

froydnj:
### Motivation

I got annoyed when trying to add `overridable` to some methods and discovering I would have to fix hundreds of callsites by hand.

Fixes #5235 

### Test plan

See included automated tests.

### DIFF
--- a/definition_validator/validator.cc
+++ b/definition_validator/validator.cc
@@ -439,7 +439,32 @@ void validateCompatibleOverride(const core::Context ctx, core::MethodRef superMe
     }
 }
 
-void validateOverriding(const core::Context ctx, core::MethodRef method) {
+optional<core::AutocorrectSuggestion>
+constructOverrideAutocorrect(const core::Context ctx, const ast::ExpressionPtr &tree, core::MethodRef method) {
+    auto methodLoc = method.data(ctx)->loc();
+    auto parsedSig = sig_finder::SigFinder::findSignature(ctx, tree, methodLoc.copyWithZeroLength());
+    if (!parsedSig.has_value()) {
+        return nullopt;
+    }
+
+    ast::Block *block = parsedSig->origSend->block();
+    if (!block) {
+        return nullopt;
+    }
+
+    auto *blockBody = ast::cast_tree<ast::Send>(block->body);
+    ENFORCE(blockBody != nullptr);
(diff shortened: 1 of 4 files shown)


jez (approved):
This is great!

As a reminder, we have `*.autocorrects.exp` files now if you prefer to use test/testdata tests instead of test/cli tests for autocorrects like this.

PR #7747, opened 2024-03-05
Redefinition of parent should error for payload

KaanOzkan:
### Motivation
I believe https://github.com/sorbet/sorbet/pull/7587 introduced a bug. This is my initial attempt at fixing it. 

The test I added is not outputting a typecheck error in `master`.

### Test plan

See included automated tests.

### DIFF
--- a/resolver/resolver.cc
+++ b/resolver/resolver.cc
@@ -1062,21 +1062,13 @@ private:
                        job.klass.data(ctx)->superClass() == core::Symbols::todo() ||
                        job.klass.data(ctx)->superClass() == resolvedClass) {
                 job.klass.data(ctx)->setSuperClass(resolvedClass);
-            } else if (!ctx.file.data(ctx).isPayload() && absl::c_any_of(job.klass.data(ctx)->locs(), [&](auto &loc) {
-                           return loc.file().data(ctx).isPayload();
-                       })) {
-                job.klass.data(ctx)->setSuperClass(resolvedClass);
-                job.klass.data(ctx)->unsetClassOrModuleLinearizationComputed();
             } else {
                 auto fileIsPayload = ctx.file.data(ctx).isPayload();
                 auto klassDefinedInPayload = absl::c_any_of(
                     job.klass.data(ctx)->locs(), [&](auto &loc) { return loc.file().data(ctx).isPayload(); });
                 auto allowsPayloadParentOverride = suppressPayloadSuperclassRedefinitionFor.contains(job.klass);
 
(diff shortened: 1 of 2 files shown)


resolver/resolver.cc
```diff
@@ -1062,21 +1062,13 @@ class ResolveConstantsWalk {
                        job.klass.data(ctx)->superClass() == core::Symbols::todo() ||
                        job.klass.data(ctx)->superClass() == resolvedClass) {
                 job.klass.data(ctx)->setSuperClass(resolvedClass);
-            } else if (!ctx.file.data(ctx).isPayload() && absl::c_any_of(job.klass.data(ctx)->locs(), [&](auto &loc) {
```
KaanOzkan, line 1065: I'm not sure the need for this branch, was the goal another pass at this code for the same class? If that's the intent it wasn't happening. Maybe I'm breaking something else by removing this.

test/testdata/resolver/payload_superclass_redefinition.rb
```diff
@@ -0,0 +1,8 @@
+# typed: true
+
+class Parent
+end
+
+class IRB::RelineInputMethod < Parent
+                             # ^^^^^^ error: Parent of class `IRB::RelineInputMethod` redefined from `IRB::InputMethod` to `Parent`
```
ilyailya, line 7: Hmm, I'm wondering why this error is reported here, but not in this test https://github.com/sorbet/sorbet/blob/master/test/cli/suppress-payload-superclass-redefinition-for/double.rb?

```ruby
# typed: true

class Parent
end

class Complex < Parent; end
class IRB::RelineInputMethod < Parent
end
```

It seems like the only difference is the amount of redefined constants in a file.

KaanOzkan: That test runs with the flag `suppress-payload-superclass-redefinition-for` https://github.com/sorbet/sorbet/blob/12e21c69b7bb8d2bf75c5110da023c7ab2b7c17e/test/cli/suppress-payload-superclass-redefinition-for/test.sh#L28.

resolver/resolver.cc
jez: Sorry for the delay in review here. I think that this is likely just me not thinking straight when I landed this change. I don't remember, but I'm guessing that what happened was that at first I tried to say "you can do this for any payload definition" and then Alexandre asked that it be gated behind a command line option to mention specific classes instead of "all classes in the payload" and I forgot to delete this code when I refactored my change.

Thanks for catching this and submitting a patch.

PR #7748, opened 2024-03-06
Compare edited files against appropriate *.rbedited file

ilyailya:
Before this commit LSP test runner, would compare edited file against `<original_file>.*.rbedited.`
That is not always a right thing to do, because code action might edit something in a different file.

For example, if the error in file `a.rb` has a code action assertion with an autocorrect in `package.rb`, then LSP test harness would apply the autocorrect, and compare `a.rb` with `a.1.rbedited`. The right thing to do is to apply autocorrect and compare `package.rb` with `package.1.rbedited`, because this is the file where autocorrect is actually applied. The PR fixes that bug.

### Motivation

The issue was found while working on https://github.com/sorbet/sorbet/pull/7573

### Test plan

See included automated tests.

### DIFF
--- a/test/helpers/position_assertions.cc
+++ b/test/helpers/position_assertions.cc
@@ -13,6 +13,7 @@
 #include "test/helpers/position_assertions.h"
 #include <iterator>
 #include <regex>
+#include <string.h>
 
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


froydnj:
The idea makes sense to me, but you have a lot of test failures.

ilyailya:
cc @froydnj fixed the test, please take a look one more time!

froydnj (approved):

PR #7752, opened 2024-03-08
Implement `visible_to Package::Namespace::*`

aisamanra:
### Motivation
Right now, `visible_to` requires you to manually enumerate each package that needs access, but sometimes this isn't exactly what you want: instead, we want to give blanket visibility to a whole set of packages under a specific namespace.

Luckily (or not, depending on how gross you think this is) Ruby successfully parses `Constant::*` as `s(:send, s(:const, nil, :Constant), :*)` which means we can use that syntax for wildcards. This parses that and sets an extra bit of information to keep track of which `visible_to` lines are wildcards and which are not.

_Technically_ the error messages here contain a lie, and I'm happy to make it more accurate, but I left it as-is: specifically, they claim that the argument to `visible_to` needs to be a constant or the string `"tests"`, and this adds support for a send. However, I think telling people, "This accepts a constant, the string `"tests"`, or a call to `::*`" would be more confusing, so I left the message as-is for now; we can figure out what a better error message could be later.

### Test plan

See included automated tests. Added to the `visibility` packager test to demonstrate this feature working.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -502,14 +502,37 @@ ast::ExpressionPtr prependRoot(ast::ExpressionPtr scope) {
     return scope;
 }
 
+bool recursiveVerifyConstant(core::Context ctx, core::NameRef fun, const ast::ExpressionPtr &root,
+                             const ast::ExpressionPtr &expr) {
+    if (ast::isa_tree<ast::EmptyTree>(expr)) {
+        return true;
+    }
+
+    auto target = ast::cast_tree<ast::UnresolvedConstantLit>(expr);
+    if (target == nullptr) {
+        if (auto e = ctx.beginError(root.loc(), core::errors::Packager::InvalidConfiguration)) {
+            e.setHeader("Argument to `{}` must be a constant", fun.show(ctx));
+        }
+        return false;
+    }
+
+    return recursiveVerifyConstant(ctx, fun, root, target->scope);
+}
+
 const ast::UnresolvedConstantLit *verifyConstant(core::Context ctx, core::NameRef fun, const ast::ExpressionPtr &expr) {
     auto target = ast::cast_tree<ast::UnresolvedConstantLit>(expr);
     if (target == nullptr) {
         if (auto e = ctx.beginError(expr.loc(), core::errors::Packager::InvalidConfiguration)) {
             e.setHeader("Argument to `{}` must be a constant", fun.show(ctx));
         }
+        return nullptr;
     }
-    return target;
+
+    if (recursiveVerifyConstant(ctx, fun, expr, target->scope)) {
+        return target;
+    }
+
+    return nullptr;
 }
 
 // Binary search to find a packages index in the global packages list
@@ -1055,7 +1078,8 @@ struct PackageSpecBodyWalk {
                     }
                     return;
                 }
-                if (auto recv = verifyConstant(ctx, send.fun, target->recv)) {
+
+                if (auto *recv = verifyConstant(ctx, send.fun, target->recv)) {
                     auto importArg = move(target->recv);
                     send.removePosArg(0);
                     ENFORCE(send.numPosArgs() == 0);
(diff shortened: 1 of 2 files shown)


core/packages/PackageInfo.h
```diff
@@ -22,6 +22,14 @@ enum class ImportType {
     Test,
 };
 
+struct VisibleTo {
+    std::vector<core::NameRef> packageName;
+    bool isWildcard;
```
jez, line 27: This makes us store an extra 4 bytes per VisibleTo annotation, even if no one in the code. That's about 20 KB – 30 KB which I guess isn't that bad right now. I imagine that I'll revisit this as I move things from this PackageInfoImpl definition into the `core::Package` symbol kind that I'm working on (most package-related things will get more compact in that representation because I intend to use SymbolRefs for more things instead of `vector<NameRef>`).

packager/packager.cc
```diff
@@ -31,6 +31,26 @@ bool isTestNamespace(const core::NameRef ns) {
...
+            return true;
+        } else {
+            // which means if the visible_to name is longer than the package name, we don't even need to check
+            return false;
+        }
+    } else {
+        // otherwise it needs to be the same
+        return vt.packageName == name;
+    }
+}
```
jez, line 52:
```suggestion
bool visibilityApplies(const core::packages::VisibleTo vt, absl::Span<const core::NameRef> name) {
    if (vt.isWildcard) {
        // a wildcard will match if it's a proper prefix of the package name
        return vt.packageName == name.subspan(0, vt.packageName.size());
    } else {
        // otherwise it needs to be the same
        return vt.packageName == name;
    }
}
```

packager/packager.cc
```diff
@@ -31,6 +31,26 @@ bool isTestNamespace(const core::NameRef ns) {
...
+            for (int i = 0; i < vt.packageName.size(); i++) {
+                if (vt.packageName[i] != name[i]) {
+                    return false;
+                }
+            }
+            return true;
+        } else {
+            // which means if the visible_to name is longer than the package name, we don't even need to check
+            return false;
+        }
```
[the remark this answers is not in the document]
jez: You can use absl::Span for this, I've compiled my suggestion below.

packager/packager.cc
```diff
@@ -188,7 +208,7 @@ class PackageInfoImpl final : public core::packages::PackageInfo {
 
     // The other packages to which this package is visible. If this vector is empty, then it means
     // the package is fully public and can be imported by anything.
```
jez, line 210: Can you update this documentation comment to describe the bool? Or make it apparent from variable/struct naming.

packager/packager.cc
```diff
@@ -1028,13 +1048,36 @@ struct PackageSpecBodyWalk {
...
+            } else if (auto target = ast::cast_tree<ast::Send>(send.getPosArg(0))) {
+                // Constant::* is valid Ruby, and parses as a send of the method * to Constant
+                // so let's take advantage of this to implement wildcards
+                if (target->fun != core::Names::star()) {
+                    if (auto e = ctx.beginError(target->loc, core::errors::Packager::InvalidConfiguration)) {
+                        e.setHeader("Argument to `{}` must be a constant or the string literal `{}`",
+                                    send.fun.show(ctx), "\"tests\"");
+                    }
+                    return;
+                }
```
jez, line 1060: Can you be sure to test that extra positional/keyword/block arguments here raise errors?


```ruby
# typed: true
A::B::*(0)
A::B::*(x: 0)
A::B::* {}
```

packager/packager.cc
```diff
@@ -1028,13 +1048,36 @@ struct PackageSpecBodyWalk {
...
             } else if (auto *target = verifyConstant(ctx, send.fun, send.getPosArg(0))) {
                 auto importArg = move(send.getPosArg(0));
                 send.removePosArg(0);
                 ENFORCE(send.numPosArgs() == 0);
                 send.addPosArg(prependName(move(importArg)));
 
-                info.visibleTo_.emplace_back(getPackageName(ctx, target));
+                info.visibleTo_.emplace_back(getPackageName(ctx, target), false);
```
jez, line 1080: Can you describe this `false` (and the `true` above in the other half of the case), either with a comment like `/* isWildCard */` or by assigning it to a variable, like `auto isWildcard = false`?

jez (approved):
I have some blocking feedback but once those are implemented, feel free to merge.

PR #7753, opened 2024-03-10
Handle `Empty` CodeActionKind

tsukkee:
This PR makes sorbet-lsp supports `Empty` CodeActionKind.
spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionKind

For example, vim-lsp sends `Empty` CodeActionKind in `only` property of CodeActionContext.
code: https://github.com/prabirshrestha/vim-lsp/blob/3af8f3b38effc4a631a15bb283a4b701c251275d/autoload/lsp/ui/vim/code_action.vim#L55

Related slack thread: https://sorbet-ruby.slack.com/archives/CFT8Y4909/p1701997597443059

### Motivation

Vim with vim-lsp users cannot use Sorbet's CodeAction because sorbet-lsp returns the following error message when vim-lsp requests CodeAction.

```
Unable to deserialize LSP request: Error deserializing JSON message: Invalid value for enum type `CodeActionKind`: 
```

This error message looks like truncated because the error is caused by an empty string (`""`), which represents `Empty` CodeActionKind.

### Test plan

I am sorry but I don't have good idea to write code to check make_lsp_types.cc follows the spec.
I would like to have some advice to do that if I need to write test in this case.

### DIFF
--- a/main/lsp/tools/generate_lsp_messages.h
+++ b/main/lsp/tools/generate_lsp_messages.h
@@ -474,6 +474,10 @@ private:
     static std::string toIdentifier(std::string_view val) {
         // Split string into components.
         auto components = absl::StrSplit(val, absl::ByAnyChar("._/"));
+        // Handle empty string.
+        if (val.empty()) {
+            return "Empty";
+        }
         // Capitalize each component.
         return fmt::format("{}", fmt::map_join(components, "", [](auto component) -> std::string {
                                if (component.length() == 0) {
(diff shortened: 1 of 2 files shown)


froydnj:

→ https://go/builds/bui_QFZPOuqhngO0tY
→ https://go/builds/bui_QFZPLOiTrO82P4
→ https://go/builds/bui_QFZP0wSpoOazuB

tsukkee:
@froydnj Hello! Thank you for your support!!
How is the CI situation? Do I need to do anything to advance this PR?

jez:
I believe that this was made redundant by this change: #8649.

PR #7754, opened 2024-03-11
Adding ObjectSpace::WeakKeyMap class

gabriel-amaral:
### Motivation

Our project recently started to reference the [`ObjectSpace::WeakKeyMap`](https://docs.ruby-lang.org/en/3.3/ObjectSpace/WeakKeyMap.html) and sorbet is not finding this constant.

### Test plan

I didn't add tests for the RBI file here since it didn't seem to add much value. I'd be happy to add some, though, if folks prefer.

### DIFF
--- a/rbi/stdlib/objspace.rbi
+++ b/rbi/stdlib/objspace.rbi
@@ -279,3 +279,85 @@ class ObjectSpace::WeakMap < Object
   def values; end
 
 end
+
+# An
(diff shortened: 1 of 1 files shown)


neilparikh:

→ https://go/builds/bui_Pie4eU3Jw31cSg
→ https://go/builds/bui_Pie4jk6hZcFfZj
→ https://go/builds/bui_Pie4q3oxgNblVF

neilparikh (approved):
Thanks!

One suggestion: might be worth adding `sig`s for some of these methods where it's possible (probably `inspect`, `clear` and `key?`). I'll merge this as is, but feel free to put up another PR with those if you'd like to.

PR #7763, opened 2024-03-12
Allow passing env variables to LSP configuration

paracycle:
This PR adds the ability to pass environment variables to the language server via an `env` setting in the LSP config.

### Motivation

I was trying to run a very non-standard Sorbet configuration in https://github.com/ruby/prism/pull/2561 and wanted to use the LSP server to help with fixing the type errors.

However, I needed to pass the `BUNDLE_GEMFILE` env variable to make `bundle exec srb` use the correct gemfile. I realized that there was no way to pass env variables to the language server process and decided to add support for them.

### Test plan

See included automated tests.

PR #7767, opened 2024-03-14
Make sure to use `config.cwd` as the `cwd` of the language server

paracycle:
It seems the `cwd` of the language server was never being set correctly, and it would always default to `workspace.rootPath`. This makes any setup that uses `config.cwd` to not be respected.

### Motivation

Fix what I assume is a bug, but @froydnj's comment [here](https://github.com/sorbet/sorbet/pull/7763#issuecomment-1997802046) might mean this is intentional.

### Test plan

See included automated tests.


jez (changes requested):
change has some conflicts with master, if you still feel strongly about this change, feel free to fix those up and re-request.

PR #7773, opened 2024-03-19
Forcibly segfault for uncaught exception

jez:
### Motivation

This means that uncaught exceptions will look like segfaults. Stripe's
crash reporting mechanisms already look for segfaults, so this saves
some work having to integrate a different way to report uncaught
top-level exceptions.

### Test plan

The `--simulate-crash` flag uses `Exception::raise` to simulate a crash, which
means that it will go through this code path. I verified that when I used that flag,
the shell showed me that the process died with a segfault.

### DIFF
--- a/main/main.cc
+++ b/main/main.cc
@@ -12,6 +12,6 @@ int main(int argc, char *argv[]) {
 
         // Forcibly exit with a segfault signal for uncaught exceptions, which makes it easier to
         // use `catchsegv.sh` to report these errors.
-        kill(0, SIGSEGV);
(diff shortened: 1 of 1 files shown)


neilparikh (approved):

jez:

→ https://go/builds/bui_Pm4q3Ke40AdZ7R
→ https://go/builds/bui_Pm4relDsLL1b6l
→ https://go/builds/bui_Pm4rS4UqiZNgAN

jez:
#7777

PR #7775, opened 2024-03-20
A whole bunch more LSP docs

jez:
### Motivation

Attempting to get some LSP docs out there.

Some features that should have docs but remain unwritten as of this PR:

- [ ] Hover
- [ ] Rename Symbol
- [ ] Find All References

And there are some experimental features which will want to be stabilized and documented, including extract variable, signature help, document highlight, and document outline.


froydnj (approved):
Cool!

PR #7776, opened 2024-03-21
Fix crash in extract variable

neilparikh:
Selecting the entire method body was causing crashes in the extract to variable logic, because the selection would be a node (InsSeq), but our logic to find the matching statement would fail (since we would walk through each statement in the InsSeq). It doesn't make sense to extract multiple lines to a variable, so we can just disallow extracting of InsSeqs.

### Motivation

Fix crashes.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -55,7 +55,7 @@ public:
             // It's not valid to extract the following node types
             if (!ast::isa_tree<ast::Break>(tree) && !ast::isa_tree<ast::Next>(tree) &&
                 !ast::isa_tree<ast::Return>(tree) && !ast::isa_tree<ast::Retry>(tree) &&
-                !ast::isa_tree<ast::RescueCase>(tree)) {
+                !ast::isa_tree<ast::RescueCase>(tree) && !ast::isa_tree<ast::InsSeq>(tree)) {
                 matchingLoc = tree.loc();
             }
         }
--- a/test/testdata/lsp/code_actions/extract_variable/invalid.rb
+++ b/test/testdata/lsp/code_actions/extract_variable/invalid.rb
@@ -120,6 +120,15 @@ end
 def endless_method = 1 + 123
 #                        ^^^ apply-code-action: [A] Extract Variable
 
+# The actual issue we saw causing crashes was the following, the selection is the entire method body:
+# def foo
+#   1 + 1
+#   2 + 2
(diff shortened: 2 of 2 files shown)


main/lsp/ExtractVariable.cc
```diff
@@ -55,7 +55,7 @@ class ExtractVariableWalk {
             // It's not valid to extract the following node types
             if (!ast::isa_tree<ast::Break>(tree) && !ast::isa_tree<ast::Next>(tree) &&
                 !ast::isa_tree<ast::Return>(tree) && !ast::isa_tree<ast::Retry>(tree) &&
-                !ast::isa_tree<ast::RescueCase>(tree)) {
+                !ast::isa_tree<ast::RescueCase>(tree) && !ast::isa_tree<ast::InsSeq>(tree)) {
```
jez, line 58: Instead of banning these, why not do something like

```
        } else if (insSeq->loc.contains(locOffsets)) {
            whereToInsert = insSeq->loc;
```

in the logic that computes `whereToInsert` if neither the `stats` nor the `expr` contained the loc?

neilparikh: I can't write a test case for this because we don't have support for multi-line assertions, but the following is the issue that was cause crashes, where the selection is the entire method body.
```ruby
def foo
  1 + 1
  1 + 1
end
```
If we extracted this, we'd get something like:
```ruby
def foo
  newVariable = 1 + 1
  1 + 1
  newVariable
end
```
We could wrap the selection in parens and then collapse into a single line with `;`, but I think it's unlikely that people want to extract more than one line.

There's the separate question if people want to extract single line `InsSeq`s, but to me, it seems like enough of edge case that we just disable it until we get feedback to the opposite.

jez: Ah, I see, that makes sense. Can you note that in a comment in the test?

froydnj (approved):
I am OK with fixing the crash in this way.

I think, however, that extracting an `InsSeq` into a variable is a reasonable -- if unusual -- thing to do: refactoring part of a method to add some logic at the end that needs to operate in multiple ways on the former return value, for instance.  Or, a bit more broadly, extracting a snippet of code into a separate method.

Ergo, we should consider this a stopgap/limitation that should eventually be removed rather than a long-term strategy of disallowing `InsSeq` nodes from having useful things done with them.

PR #7777, opened 2024-03-21
Stop catching top-level SorbetException exceptions

jez:
Also:

- Use `std::current_exception` to print the name of all uncaught
  exceptions, not just `SorbetException` and subclasses.
- Use `abort` at the end of our `terminate_handler` to be sure the
  process is aborted (avoid undefined behavior).
- Use `abi::__cxa_demangle` to print a better (demangled) exception name

### Motivation

So that we can include uncaught C++ exceptions in Stripe's crash monitoring
(previously, only the non-`SorbetException` exceptions would get caught by our
crash monitoring).

### Test plan

Tested locally to see that things like

```
Sorbet raised uncaught exception type=std::bad_optional_access what=bad_optional_access
```

do get printed.

Testing manually on Stripe's codebase to make sure that the `SIGABRT` signal
that the `on_terminate` function exits with is enough to trigger a monitor.

### DIFF
--- a/common/common.cc
+++ b/common/common.cc
@@ -492,12 +492,8 @@ public:
                 // Aparently this is the only way to convert a std::exception_ptr to std::exception
                 std::rethrow_exception(eptr);
             } catch (const std::exception &e) {
-                int status;
-                // nullptr to request that __cxa_demangle allocate the string on our behalf,
-                // instead of populating a pre-allocated buffer.
-                auto demangled = abi::__cxa_demangle(typeid(e).name(), nullptr, nullptr, &status);
-                sorbet::fatalLogger->error("Sorbet raised uncaught exception type={} what={}", demangled, e.what());
-                std::free(demangled);
+                sorbet::fatalLogger->error("Sorbet raised uncaught exception type={} what={}",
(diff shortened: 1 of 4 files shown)


common/common.cc
```diff
@@ -486,7 +486,41 @@ vector<int> sorbet::findLineBreaks(string_view s) {
...
+        auto eptr = current_exception();
+        if (eptr) {
+            try {
+                // Aparently this is the only way to convert a std::exception_ptr to std::exception
+                std::rethrow_exception(eptr);
+            } catch (const std::exception &e) {
+                int status;
+                // nullptr to request that __cxa_demangle allocate the string on our behalf,
+                // instead of populating a pre-allocated buffer.
+                auto demangled = abi::__cxa_demangle(typeid(e).name(), nullptr, nullptr, &status);
```
froydnj, line 498: `common.cc` already has a `demangle` function. 😅

froydnj (approved):

common/common.cc
jez: I should have noticed that when I didn't need to add the `<cxxabi.h>` include!

PR #7778, opened 2024-03-22
fix issues when `Exception#backtrace_locations` returns `nil`

froydnj:
This reverts parts of #7710.

cc @casperisfine 

### Motivation

The changes to this file in #7710 cause errors when `backtrace_locations` is `nil`.  It's not obvious to me why `backtrace_locations` is `nil`, but it is on a number of Stripe's internal tests.  I see [references in the Ruby source](https://github.com/ruby/ruby/blob/35fe4b048034a8568cd2a2dc34f59dc12f644d21/error.c#L3322-L3343) to setting `backtrace_locations` to `nil` when exceptions are marshaled (!), but AFAICT we are not going through marshaling in the codepaths that are erroring.

This is error-handling code, so I'm not too concerned with the efficiency win from iterating over `backtrace_locations` rather than taking apart the backtrace itself.

### Test plan

I was not able to write a reproducible test case for this.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/serializable.rb
+++ b/gems/sorbet-runtime/lib/types/props/serializable.rb
@@ -261,11 +261,16 @@ module T::Props::Serializable::DecoratorMethods
   end
 
   def message_with_generated_source_context(error, generated_method, generate_source_method)
-    generated_method = generated_method.to_s
-    line_loc = error.backtrace_locations.find {|l| l.base_label == generated_method}
-    return unless line_loc
+    line_label = error.backtrace.find {|l| l.end_with?("in `#{generated_method}'")}
+    return unless line_label
 
-    line_num = line_loc.lineno
+    line_num = if line_label.start_with?("(eval)")
+      # (eval):13:in `__t_props_generated_serialize'
+      line_label.split(':')[1]&.to_i
+    else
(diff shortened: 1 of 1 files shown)


jez (approved):

casperisfine:
Ah sorry, I should have thought of that.

This generally comes from raising exceptions with a backtrace from another exception, e.g.:

```ruby
raise SomeError, previous_error.message, previous_error.backtrace
```

The above exception will have `#backtrace_location` return `nil`.

I recently implemented a way to preserve it though: https://github.com/ruby/ruby/pull/10017, but it will take a long time before most code is migrated.

gems/sorbet-runtime/lib/types/props/serializable.rb
```diff
@@ -261,11 +261,16 @@ def add_prop_definition(prop, rules)
   end
 
   def message_with_generated_source_context(error, generated_method, generate_source_method)
-    generated_method = generated_method.to_s
-    line_loc = error.backtrace_locations.find {|l| l.base_label == generated_method}
-    return unless line_loc
+    line_label = error.backtrace.find {|l| l.end_with?("in `#{generated_method}'")}
```
casperisfine, line 264: If you are open to it, I'd like to suggest to first try to use `backtrace_locations` if it's present, and then only fallback.

This way you get the speed benefit in the vast majority of cases, and only degrade to string parsing in a minority of cases.

Also this `l.end_with?("in `#{generated_method}'")` won't match on Ruby 3.4 anyway, so I'll open another PR.

casperisfine:
https://github.com/sorbet/sorbet/pull/7779

PR #7779, opened 2024-03-22
Restore Ruby 3.4 compatibility in T::Props::Serializable

casperisfine:
Followup: https://github.com/sorbet/sorbet/pull/7778
Followup: https://github.com/sorbet/sorbet/pull/7710

Exceptions are not guaranteed to have a `#backtrace_locations` because of `Exception#set_backtrace` and `Kernel#raise(error, message, backtrace)`.

It's still prefereable to use `#backtrace_location` when present as it requires much less string parsing.

But if it's missing we can fallback to parse backtraces the old way, however we need to account for Ruby 3.4 newer format.

@froydnj @jez @paracycle

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/serializable.rb
+++ b/gems/sorbet-runtime/lib/types/props/serializable.rb
@@ -261,15 +261,30 @@ module T::Props::Serializable::DecoratorMethods
   end
 
   def message_with_generated_source_context(error, generated_method, generate_source_method)
-    line_label = error.backtrace.find {|l| l.end_with?("in `#{generated_method}'")}
-    return unless line_label
+    generated_method = generated_method.to_s
+    if error.backtrace_locations
+      line_loc = error.backtrace_locations.find {|l| l.base_label == generated_method}
+      return unless line_loc
 
-    line_num = if line_label.start_with?("(eval)")
-      # (eval):13:in `__t_props_generated_serialize'
-      line_label.split(':')[1]&.to_i
+      line_num = line_loc.lineno
     else
-      # (eval at /Users/jez/stripe/sorbet/gems/sorbet-runtime/lib/types/props/has_lazily_specialized_methods.rb:65):13:in `__t_props_generated_serialize'
-      line_label.split(':')[2]&.to_i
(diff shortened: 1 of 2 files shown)


gems/sorbet-runtime/lib/types/props/serializable.rb
```diff
@@ -261,17 +261,32 @@ def add_prop_definition(prop, rules)
...
+      return unless line_label
+
+      line_num = if line_label.start_with?("(eval)")
+        # (eval):13:in ...
+        line_label.split(':')[1]&.to_i
+      else
+        # (eval at /Users/jez/stripe/sorbet/gems/sorbet-runtime/lib/types/props/has_lazily_specialized_methods.rb:65):13:in ...
+        line_label.split(':')[2]&.to_i
+      end
+      return unless line_num
```
froydnj, line 288: I think you want to keep this `return` outside of the `if`; the original PR for 3.4 compat returned early:

https://github.com/sorbet/sorbet/pull/7710/files#diff-b6758eb606992ab00fd8e66aa0375c569648250f874bf6fb51fe402769e0e9cbL264-L274

unless we are guaranteed to have a `line_num` in the `backtrace_locations` case?

froydnj (dismissed):
Thank you for the patch!

gems/sorbet-runtime/lib/types/props/serializable.rb
casperisfine: > unless we are guaranteed to have a `line_num` in the `backtrace_locations` case?

It's guaranteed if the method is implemented in Ruby, but could be `0` if implemented in C.

But it's a good catch, I applied the suggested change.

jez:

→ https://go/builds/bui_QRe33VxDQmlaPU

jez:
Sorry for the long review on this change. The script we used to test PRs against Stripe's codebase had atrophied and broken, and we had not gotten around to fixing it.

The script is fixed, I've tested these changes on Stripe's codebase, and everything looks great.

Thanks for the patience.

jez (approved):

PR #7784, opened 2024-03-26
Add `base64digest` methods for `OpenSSL::HMAC`

vaindil:
### Motivation

Two `base64digest` methods were added to `OpenSSL::HMAC` in Ruby 3.1.0, but Sorbet does not have them ([docs here](https://ruby-doc.org/stdlib-3.1.0/libdoc/openssl/rdoc/OpenSSL/HMAC.html#method-c-base64digest), [source here](https://github.com/ruby/ruby/blob/4300c42a7eb12a0aa81e4ec79891cdba1cfdf3aa/ext/openssl/lib/openssl/hmac.rb)). I'm not sure what version Sorbet supports for these rbi files, but figured I'd open a PR in case they'd be helpful.

Everything else in the tweaked file is untyped, so I kept that convention rather than returning `String` and using the proper parameters.

### Test plan

Not sure that tests would add a whole lot of value here, so I didn't include any. I can add some if they should be added.

### DIFF
--- a/rbi/stdlib/openssl.rbi
+++ b/rbi/stdlib/openssl.rbi
@@ -4442,6 +4442,10 @@ class OpenSSL::HMAC
   sig {returns(::T.untyped)}
   def hexdigest(); end
 
+  # Returns the authentication code as a Base64-encoded string.
+  sig {returns(::T.untyped)}
+  def base64digest(); end
+
   sig do
     params(
       arg0: ::T.untyped,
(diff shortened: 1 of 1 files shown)


ilyailya:

→ https://go/builds/bui_PoKuJy8A3ugeD0
→ https://go/builds/bui_PoKu5zGgZcnnvP
→ https://go/builds/bui_PoKufAb17TmmzA

ilyailya (approved):
The change looks good to me. Thanks a mil for updating the signatures!

ilyailya:
>Everything else in the tweaked file is untyped, so I kept that convention rather than returning String and using the proper parameters.

It's not really a convention, just no one put work into turning this file to `# typed: true`

PR #7786, opened 2024-03-27
Devirtualize WatchmanProcess by merging LSPWatchmanProcess

jez:
### Motivation

We have had a longstanding problem where a pure-virtual function is called.
Turns out that this blames to a call to `processExit` by way of `exitWithCode`
by way of `WatchmanProcess::~WatchmanProcess`.

In particular, you're not allowed to call virtual functions from destructors
(even if the virtual function is called transitively).

But there's no real reason why this class should be virtual/abstract in the
first place, because it's only ever subclassed once.

### Commit summary

- **Factor mergeCounters helper to Counters.{h,cc}** (15078eb84)

  This was defined in two places, and I needed to use it in a third.

- **Move tagNewRequest to LSPMessage** (7e53e7846)

  This was defined as a helper in an anonymous namespace, and I would have 
  needed to call it from two files. It now lives as a method on 
  LSPMessage, which I think it more fitting.

- **Move MessageQueueState class to its own file** (b54509d92)

  ... so that I can use it from two places.

- **Move LSPWatchmanProcess fields into WatchmanProcess** (c9152acd1)

- **Move the LSPWatchmanProcess methods into WatchmanProcess** (94a186592)

- **Remove all the `virtual` from these methods** (77c68cec0)

### Test plan

Existing tests.


froydnj:
Point of order:

> In particular, you're not allowed to call virtual functions from destructors (even if the virtual function is called transitively).

This is not true; from the spec (admittedly this is the most recent draft, but I strongly suspect this part has not changed very much from C++17), [[class.cdtor]p4:](https://eel.is/c++draft/class.cdtor#4):

> Member functions, including virtual functions ([[class.virtual]](https://eel.is/c++draft/class.virtual)), can be called during construction or destruction ([[class.base.init]](https://eel.is/c++draft/class.base.init))[.](https://eel.is/c++draft/class.cdtor#4.sentence-1)
When a virtual function is called directly or indirectly from a constructor or from a destructor, including during the construction or destruction of the class's non-static data members, and the object to which the call applies is the object (call it x) under construction or destruction, the function called is the final overrider in the constructor's or destructor's class and not one overriding it in a more-derived class[.](https://eel.is/c++draft/class.cdtor#4.sentence-2)
If the virtual function call uses an explicit class member access ([[expr.ref]](https://eel.is/c++draft/expr.ref)) and the object expression refers to the complete object of x or one of that object's base class subobjects but not x or one of its base class subobjects, the behavior is undefined[.](https://eel.is/c++draft/class.cdtor#4.sentence-3)

So the subtlety here is that virtual method lookup as you normally expect it to happen is not what happens in destructors: the relevant virtual function that's called only considers the class and superclasses, not any subclasses.  Ergo, if you have a destructor and it calls a pure virtual function defined in that class, the pure virtual function is always called from the destructor, and not any overrides in subclasses.

I could not have told you this prior to looking it up, and was initially puzzled by this restriction, but it makes sense: destructors are executed from the most-derived to least-derived class, so if you needed to call a virtual function defined in a subclass from a base class during destruction, that subclass's data no longer exist, so it wouldn't be meaningful to call the virtual function (what object could you call it on?) and would be actively dangerous (practically textbook use-after-free).

froydnj (approved):

PR #7791, opened 2024-03-28
Teach merchant_token_prop about the `name` field

jez:
### Motivation

We were not handling the fact that this method accepts an optional `name`
keyword parameter, which can be used to customize the name to something that's
not `merchant`.

### Test plan

See included automated tests.

Those at Stripe can see that this is how this method works by looking at
this code:

<http://go/GnhimGWM>


jez:

→ https://go/builds/bui_Pp53XeicSHxKAY
→ https://go/builds/bui_Pp53JXYmG1EJxh
→ https://go/builds/bui_Pp532Le3BQk4sG

ilyailya (approved):

PR #7792, opened 2024-03-28
Make LSPTypechecker::getResolved use thread pool

jez:
### Motivation

The `LSPTypechecker::getResolved` method is used by `LSPTypechecker::query`,
which is used by `LSPQuery::bySymbol`, which powers things like find all
references and move method and convert to singleton class method.

Based on our strategy of searching for references in files that have a reference
to _any_ symbol with the name of the current symbol, `bySymbol` can run the
query over many files. In particular, a common case in Stripe's codebase is
finding references on methods called `call`, where there are thousands of files.

Currently, Sorbet resolves all of these files single threaded, even when it
would use the worker pool for running inference on those files (i.e., there was
a worker pool in scope already).

There's no reason to do this, and in fact all of the things inside `getResolved`
(namer and resolver) already work in parallel if given a `WorkerPool`. (Well,
almost all the things--packager does not run the incremental version in
parallel, but that can be a future change, and/or might just go away as we move
more packager into namer/resolver).

### Test plan

The code compiles.

I'm not adding any new parallelism, which means there's no worry about
potentially racing on shared state--this change just takes advantage of the
existing parallelism already present in the pipeline, so there's no reason to
expect anything to break because of this.

Separately, I'm working on testing the performance of this for a few Find All
References queries on Stripe's codebase.

### DIFF
--- a/main/lsp/LSPTypechecker.cc
+++ b/main/lsp/LSPTypechecker.cc
@@ -639,7 +639,7 @@ LSPQueryResult LSPTypechecker::query(const core::lsp::Query &q, const std::vecto
     ENFORCE(gs->errorQueue->isEmpty());
     ENFORCE(gs->lspQuery.isEmpty());
     gs->lspQuery = q;
-    auto resolved = getResolved(filesForQuery);
+    auto resolved = getResolved(filesForQuery, workers);
     tryApplyDefLocSaver(*gs, resolved);
     tryApplyLocalVarSaver(*gs, resolved);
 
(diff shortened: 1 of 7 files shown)


jez:

→ https://go/builds/bui_Pp5tnkPN9YtPBP
→ https://go/builds/bui_Pp5t9h7MszqQwf
→ https://go/builds/bui_Pp5t6dn7Pr5ijF

jez:
It looks like this makes Find All References on a method named `call` 20% faster.

PR #7801, opened 2024-04-02
Improve error location for "unrecognized kwarg"

jez:
### Motivation

Requested by a user.

### Test plan

See included automated tests.

**Before**

Screenshot 2024-04-02 at 2 23 02 PM

**After**

Screenshot 2024-04-02 at 2 07 44 PM

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -1368,7 +1368,7 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
                 if (auto e = gs.beginError(kwargErrLoc, errors::Infer::MethodArgumentCountMismatch)) {
                     e.setHeader("Unrecognized keyword argument `{}` passed for method `{}`", arg.show(gs),
                                 method.show(gs));
-                    if (kwargErrLoc != kwargsLoc && kwargErrLoc.exists()) {
+                    if (kwargErrLoc.exists() && kwargErrLoc != kwargsLoc) {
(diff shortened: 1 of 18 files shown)


core/types/calls.cc
```diff
@@ -1338,9 +1358,20 @@ DispatchResult dispatchCallSymbol(const GlobalState &gs, const DispatchArgs &arg
...
+                auto it = kwargLocs.find(key.asName());
+                // TODO(jez) This papers over some stuff around kwsplats which might get in our way
+                // of getting a loc for known keyword args. In those cases, we simply give up right now.
+                if (it != kwargLocs.end()) {
+                    kwargErrLoc = it->second;
+                }
+                if (auto e = gs.beginError(kwargErrLoc, errors::Infer::MethodArgumentCountMismatch)) {
                     e.setHeader("Unrecognized keyword argument `{}` passed for method `{}`", arg.show(gs),
                                 method.show(gs));
+                    if (kwargErrLoc != kwargsLoc && kwargErrLoc.exists()) {
```
froydnj, line 1371: Nit: this seems slightly clearer with the condition reversed (s/o to `std::optional` enabling comparison to non-`optional` types).

```suggestion
                    if (kwargErrLoc.exists() && kwargErrLoc != kwargsLoc) {
```

froydnj (approved):

core/types/calls.cc
jez: I don't think there's an `optional` in play here, were you suggesting that I introduce it somehow?

PR #7807, opened 2024-04-03
Document the `--suppress-payload-superclass-redefinition-for` flag

egiurleo:
### Motivation
This flag was defined in #7587 and can be used to suppress error 5012, which occurs when changing a class's superclass. It is meant to be used primarily when auto-generating files that will change the superclass of a class defined in Sorbet's RBI payloads.

### Test plan
I tested this by running the docs site on my machine.

### DIFF
--- a/website/docs/error-reference.md
+++ b/website/docs/error-reference.md
@@ -1453,6 +1453,17 @@ class C < A; end
 class C < B; end
 ```
 
+> **Note**: In cases where you intend to change the superclass of a class
+> originally defined in Sorbet's payload RBIs (e.g. you generate RBI for a newer
+> version of a gem in Sorbet's payload), you may suppress this error using the
+> `--suppress-payload-superclass-redefinition-for` command line flag.
+>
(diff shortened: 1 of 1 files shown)


jez (commented):
What do you think about not putting this as the first thing users see when reading about the error?

let’s put it at the bottom of the section, with caveats around when it should and shouldn’t be used

egiurleo:
@jez Good point -- I moved to the bottom and made it clear that it's only for purposefully changing the superclass of a payload class. Let me know if that's clearer.

jez:
Hey sorry about this, I didn't notice at first because I didn't expand the context in the diff. I thought there would be more explanation for this error than there actually was.

I prefer to give people a way to constructively fix errors before telling them about how to silence them, and that didn't exist for this error. I've opened #7812 to provide some more context, and included the changes you made here. Feel free to leave suggestions on that other change and then I'll probably proceed with that version instead of this one.

egiurleo:
Closing in favor of #7812.

PR #7808, opened 2024-04-04
prework: Ensure that symbol table is frozen

jez:
### Motivation

These two calls were actually mutating GlobalState without potentially
checking this. I think that luckily we don't have any bugs because of
this, but let's hoist these checks earlier just to be sure.

### Test plan

Existing tests


ilyailya (approved):
Change looks good to me!

jez:
This exposed that our `GlobalState::sanityCheck` function calls `enter*Symbol` and attempts to ensure that the same `Ref` ID is produced, and this breaks that.

I'm going to have to think about the best way to fix this, so going to hold off for the time being.

PR #7809, opened 2024-04-04
Stop copying some structures in namer

jez:
### Motivation

Happened to see these, minor perf optimization.

### Test plan

Existing tests


froydnj (approved):
Ahhh, good catch.

PR #7811, opened 2024-04-04
Make findSymbols work in-place

jez:
### Motivation

This both allows us to remove a `fast_sort` at the end of findSymbols
(update: two `fast_sort`, from symbolFinder and treeSymbolizer)
and also sets this methd up to operate on `absl::Span`, which will make
implementing parts of the upcoming packager refactors more easily.

This could potentially adversely affect performance if this causes worse
cache locality, but I think that locality should be ~mostly fine as long
as the amount of work per file is approximately similar?

Another thing: before this change, we would not attempt to start merging
index results until the first thread had completed all its work, which
meant that merging the `allFoundDefinitions` was frequently the long
poll. Now, the merging happens as soon as the thread finishes, which
means it's keeping the main thread busier.

**Update**: it's not quite two, because one of them was load bearing for LSP:

LSP runSlowPath was relying on namer sorting files.
That meant that in the batch pipeline, which already sorts the input
earlier, namer was doing a redundant sort, but in LSP it was the only
sort.

We could ENFORCE this, except that we don't actually have the full list
sorted--`__package.rb` files are generally first, but only in the
pipeline, not LSP. So for now I'm just going to leave this new
`fast_sort` call here without an additional ENFORCE.

So in the batch pipeline, we net drop two sorts, and in LSP we net drop one.

### Test plan

Existing tests

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -2370,10 +2370,6 @@ ast::ParsedFilesOrCancelled defineSymbols(core::GlobalState &gs, vector<SymbolFi
     return output;
 }
 
-struct SymbolizeTreesResult {
-    vector<ast::ParsedFile> trees;
-};
-
 vector<ast::ParsedFile> symbolizeTrees(const core::GlobalState &gs, vector<ast::ParsedFile> trees,
                                        WorkerPool &workers) {
     Timer timeit(gs.tracer(), "naming.symbolizeTrees");
@@ -2387,7 +2383,6 @@ vector<ast::ParsedFile> symbolizeTrees(const core::GlobalState &gs, vector<ast::
     workers.multiplexJob("symbolizeTrees", [&gs, fileq, resultq]() {
         Timer timeit(gs.tracer(), "naming.symbolizeTreesWorker");
         TreeSymbolizer inserter;
-        SymbolizeTreesResult output;
         ParsedFileWithIdx job;
         for (auto result = fileq->try_pop(job); !result.done(); result = fileq->try_pop(job)) {
             if (result.gotItem()) {
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -26,16 +26,16 @@ namespace sorbet::namer {
 
 namespace {
 
-struct SymbolFinderJob {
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_PrioGCq7pZuHyA
→ https://go/builds/bui_PrioORbrlHxc3P
→ https://go/builds/bui_Priotvc2mMu971

jez:

→ https://go/builds/bui_PrizqmIrF7s5PW
→ https://go/builds/bui_Priz90FwDKqi8b
→ https://go/builds/bui_Priz749px9NPkM

jez:

→ https://go/builds/bui_PrjE41n0ceGsyq
→ https://go/builds/bui_PrjEy60czcF52g
→ https://go/builds/bui_PrjEDLVZq1dSxy

jez:

→ https://go/builds/bui_Ps57Y0JodKdjdc
→ https://go/builds/bui_Ps57tSRe1pjd3g
→ https://go/builds/bui_Ps57bIJpXcewOA

namer/namer.cc
```diff
@@ -2369,43 +2377,39 @@ struct SymbolizeTreesResult {
...
             }
         }
-        if (!output.trees.empty()) {
-            resultq->push(move(output), output.trees.size());
-        }
     });
-    trees.clear();
 
     {
-        SymbolizeTreesResult threadResult;
```
froydnj, line 2398: I think this type is now unused and its definition could be deleted?

froydnj (approved):
I like this idea; I'm curious whether it makes things faster.

I don't see how this enables faster index merging, because this doesn't change anything about how the pipeline itself works -- was that from an earlier version of the PR?

jez:
Sorry it’s not the index merging, it’s the found definition merging that happens while reading from the result queue in findSymbols on the main thread.

Also I wrote that comment before noticing that that particular phase of namer only takes like ~12ms so it’s unlikely to be the source of any speedup.

namer/namer.cc
jez: Good catch!

PR #6534, opened 2022-11-02
Pre-work for #6521

jez:
### Motivation

Something is wrong with #6521. I'm going to try bisecting it.

**Update** as of right now, these are all the okay changes, and the problem is now wholly contained in #6521.

May as well land this.

### Test plan

See included automated tests.

PR #7816, opened 2024-04-08
Comment out error without any usages

ilyailya:
### Motivation

Per the tin. 

Looks like the last usage of that error was removed in this PR in namer.cc https://github.com/sorbet/sorbet/pull/6534
### Test plan

See included automated tests.


jez (approved):
neat!

PR #7824, opened 2024-04-10
Improve perf of T::Enum when not using enable_legacy_t_enum_migration_mode

jez:
### Motivation

h/t @egiurleo / @Morriar (and possibly others at Shopify) for finding and
reporting this performance opportunity.

### Commit summary

- **Conditionally define T::Enum <> String methods** (30645b06c)

  These methods should not incur a performance overhead for anyone who 
  never calls `enable_legacy_t_enum_migration_mode`.

- **no-op: Reindent** (69284ed2a)

- **no-op: Get it to typecheck** (76177b67a)

- **Add a benchmark for T::Enum ==** (8d3d3ec22)

### Test plan

**master:**

```
❯ bex rake bench:enum
before T::Configuration.enable_legacy_t_enum_migration_mode
T::Enum == T::Enum: 68.176 ns
T::Enum == String: 79.956 ns

after T::Configuration.enable_legacy_t_enum_migration_mode
T::Enum == T::Enum: 68.173 ns
T::Enum == String: 3.654 μs
```

**branch:**

```
❯ bex rake bench:enum
before T::Configuration.enable_legacy_t_enum_migration_mode
T::Enum == T::Enum: 16.291 ns
T::Enum == String: 14.887 ns

after T::Configuration.enable_legacy_t_enum_migration_mode
T::Enum == T::Enum: 73.553 ns
T::Enum == String: 3.875 μs
```

This suggests three things:

1.  This change makes `==` comparison on `T::Enum` ~4x faster for
    codebases that don't use `enable_legacy_t_enum_migration_mode`.

2.  This change does not regress performance on codebases that do use
    `enable_legacy_t_enum_migration_mode`.

3.  This suggests Stripe may want to invest in fixing legacy enum
    migration soft assertions so that Stripe can benefit from the
    improved performance.

    The benchmark script above does not even consider the performance of
    calling `soft_assert_handler`--Stripe's `soft_assert_handler` is
    substantially slower than the one in this benchmark.

### DIFF
--- a/gems/sorbet-runtime/bench/enum.rb
+++ b/gems/sorbet-runtime/bench/enum.rb
@@ -38,6 +38,8 @@ module SorbetBenchmarks
 
     def self.all_comparisons
       time_block("T::Enum == T::Enum", iterations_of_block: 100_000, iterations_in_block: 9) do
+        # rubocop:disable Lint/BinaryOperatorWithIdenticalOperands
+        # rubocop:disable Lint/Void
         MyEnum::X == MyEnum::X
         MyEnum::X == MyEnum::X
         MyEnum::X == MyEnum::X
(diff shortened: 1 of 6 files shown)


Morriar (approved):
❤️

froydnj (approved):
That's...quite the performance difference.

amomchilov (approved):
Thanks for taking the time to upstream and benchmark this! <3

PR #2444, opened 2020-01-09
Upgrade to bazel-2.0.0

elliottt:
### Motivation

We're on a pretty old version of bazel, this bumps us up to the current release.

### Test plan

CI

PR #7827, opened 2024-04-11
Bump Bazel to 6.5.0

Morriar:
### Motivation

We're trying to move away from the [sorbet/bazel-toolchain](https://github.com/sorbet/bazel-toolchain) and use [bazel-contrib/toolchains_llvm](https://github.com/bazel-contrib/toolchains_llvm) instead.

The [last stable release - 1.0.0](https://github.com/bazel-contrib/toolchains_llvm/releases/tag/1.0.0) requires Bazel >= 7.0.0.

This PR first bumps the Bazel version to 6.5.0 (last 6x stable release) and applies the required compatibility changes.

### Test plan

See included automated tests.


WORKSPACE
```diff
@@ -87,12 +87,14 @@ rust_register_toolchains(
...
-BAZEL_VERSION = "5.2.0"
+BAZEL_VERSION = "6.5.0"
 
-BAZEL_INSTALLER_VERSION_LINUX_X86_64_SHA = "7d9ef51beab5726c55725fb36675c6fed0518576d3ba51fb4067580ddf7627c4"
+BAZEL_INSTALLER_VERSION_LINUX_X86_64_SHA = "c0161a346b9c0d00e6eb3d3e8f9c4dece32f6292520248c5ab2e3527265601c1"
 
-BAZEL_INSTALLER_VERSION_LINUX_ARM64_SHA = "ae50cb7d64aebee986287134ff8ca0335651a0c1685348b3216f3fdfa20ff7e7"
+# I couldn't find any official checksums for the LINUX_ARM64 version of the Bazel installer in
+# https://github.com/bazelbuild/bazel/releases/tag/6.5.0.
+BAZEL_INSTALLER_VERSION_LINUX_ARM64_SHA = "5afe973cadc036496cac66f1414ca9be36881423f576db363d83afc9084c0c2f"
```
Morriar, line 96: I used the [bazel-6.5.0-linux-arm64.sha256](https://github.com/bazelbuild/bazel/releases/download/6.5.0/bazel-6.5.0-linux-arm64.sha256) sha instead but this will most likely not work when trying to install Bazel on a linux-arm64 instance.

I'm not sure what workaround we can use?

common/crypto_hashing/BUILD
```diff
@@ -8,8 +8,7 @@ cc_library(
     visibility = ["//visibility:public"],
     deps = ["//common"] + select({
         "//tools/config:webasm": ["@com_github_blake2_blake2"],
-        "//tools/config:darwin_x86_64": ["@com_github_blake2_libb2"],
-        "//tools/config:linux_x86_64": ["@com_github_blake2_libb2"],
-        "//conditions:default": ["@com_github_blake2_blake2"],
+        "//tools/config:linux_arm64": ["@com_github_blake2_blake2"],
+        "//conditions:default": ["@com_github_blake2_libb2"],
```
Morriar, line 12: Config selection is ambiguous with Bazel 6.

We can change the default to build with `libb2` and only use `blake2` when building with `webasm` or `linux_arm64` config.

test/cli/file-table-json/test.sh
```diff
@@ -7,6 +7,6 @@ main/sorbet --censor-for-snapshot-tests --silence-dev-message -p file-table-json
 echo "----- --no-stdlib -p file-table-json -----------------------------------"
 main/sorbet --censor-for-snapshot-tests --silence-dev-message --no-stdlib -p file-table-json test/cli/file-table-json/file-table-json.rb
 echo "----- -p file-table-full-json ------------------------------------------"
-main/sorbet --censor-for-snapshot-tests --silence-dev-message -p file-table-full-json test/cli/file-table-json/file-table-json.rb
+main/sorbet --censor-for-snapshot-tests --silence-dev-message -p file-table-full-json test/cli/file-table-json/file-table-json.rb | sed 's/https:\/\/.*procs.rbi/procs.rbi/g'
```
Morriar, line 10: Output directory changed from `bazel-out/host/` to `bazel-out/k8-opt-exec-2B5CBBC6` and is platform dependent.

I'm sure there is a smarter way to do this?

WORKSPACE
jez: How is CI passing right now? Don't we test on arm?

test/cli/file-table-json/test.sh
jez: What do you think about the approach I had taken before?

https://github.com/sorbet/sorbet/pull/7509/commits/fb5f06c764014e5571bb71c78baafdc68a136ff6

jez: Since this file ~never changes, it will have the effect of making it easier for people to grep and find this RBI file, since it will be checked into the repo. That commit also adds a test to make sure that the version checked into the repo matches the one produced by the bazel job.

WORKSPACE
jez: The log output from buildkite suggests that this is working fine:

https://buildkite.com/sorbet/sorbet/builds/30965#018ecf5d-ea61-4039-9496-9a3b821c3b4e/180-189

So I'm curious if you could elaborate on what about this you think is "not likely to work."

jez (approved):
This looks great to me. I'd love to get @froydnj's opinion just in case, but I think everything here is non-objectionable to me.

I don't even think we really need to block on the `procs.rbi` thing, because we can always do that in a self-contained commit later.

WORKSPACE
Morriar: As you pointed out on Slack, this checksum is actually not used for `linux-arm64` as there is no binary installer for this platform: https://github.com/sorbet/sorbet/blob/master/bazel#L102.

I wonder if we could just remove the `BAZEL_INSTALLER_VERSION_LINUX_ARM64_SHA` constant and leave a comment explaining this instead as this is confusing.

froydnj:

→ https://go/builds/bui_PubUsq9pYgKgSN
→ https://go/builds/bui_PubU58Bqlj1usN
→ https://go/builds/bui_PubUzjdIpdOsOs

WORKSPACE
Morriar: Ok, we actually need `BAZEL_INSTALLER_VERSION_LINUX_ARM64_SHA` as it's used to checksum the installed bazel without the installer: https://github.com/sorbet/sorbet/blob/master/bazel#L73.

I updated the comment 👍

froydnj (changes requested):
This is failing on Stripe's infra.  The error on x86-64 is:

```
No cached Bazel v6.5.0 found, installing...
/tmp/tmp.5psZu7hrM0
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100    80    0    80    0     0   4444      0 --:--:-- --:--:-- --:--:--  4444
Installer checksum mismatch:
  Expected: c0161a346b9c0d00e6eb3d3e8f9c4dece32f6292520248c5ab2e3527265601c1
  Actual:   6497b75e8c519d72e159a7c8f95c2e0c2275880500cc9dd64224fde3e02a490f
To accept this mismatch, update BAZEL_INSTALLER_VERSION_LINUX_X86_64_SHA in the WORKSPACE file and re-run.
```

which seems...odd to me?  I don't know why this would fail for Stripe and not in Sorbet's CI.

The aarch64 error is even less helpful:

```
sorbet/scripts/build-sorbet exited 1:
No cached Bazel v6.5.0 found, installing...
./bazel: Failed to extract Bazel version from WORKSPACE
```

test/cli/file-table-json/test.sh
Morriar: I pulled your approach in https://github.com/sorbet/sorbet/pull/7827/commits/d1d2fccc5b3c28e198a8ea4e317ea23c6e713d87 👍

Morriar:
@froydnj 

> No cached Bazel v6.5.0 found, installing...
> ./bazel: Failed to extract Bazel version from WORKSPACE

You got me in between two pushes trying to remove the SHA for linux_arm64 but we actually need it. This should be fixed now 👍 

> Installer checksum mismatch:
>   Expected: c0161a346b9c0d00e6eb3d3e8f9c4dece32f6292520248c5ab2e3527265601c1
>   Actual:   6497b75e8c519d72e159a7c8f95c2e0c2275880500cc9dd64224fde3e02a490f
> To accept this mismatch, update BAZEL_INSTALLER_VERSION_LINUX_X86_64_SHA in the WORKSPACE file and re-run.

This I have no idea. The SHA is matched correctly both on CI and a linux_x86_64 test machine. I couldn't find the SHA anywhere in the https://github.com/bazelbuild/bazel/releases/tag/6.5.0 assets. Could the script pull the wrong version yet still display `No cached Bazel v6.5.0 found, installing...` somehow?

froydnj:

→ https://go/builds/bui_PucaiSWl4cbRpQ
→ https://go/builds/bui_PucakvUl9TNSTD
→ https://go/builds/bui_PucaHFHZlCZa2X

froydnj:

→ https://go/builds/bui_PufrRDGaAbRtPD
→ https://go/builds/bui_PufryAlFZKpET0
→ https://go/builds/bui_PufrB97yDoGVYb

froydnj (approved):
This builds OK on Stripe's infra, so I think we are good to go! 👍

jez:
Testing on macOS:

https://buildkite.com/sorbet/sorbet/builds/30993

jez:
Good catch @froydnj, looks like there are failures on macOS.

Weirdly, I can't reproduce locally, neither on an arm64 machine nor on an x86_64 machine. This is the command that CI says should fail:

```
./bazel test --config=dbg --test_output=errors //core:core_test
```

jez:
Sorry, I misspoke. This is the command that should fail:

```
./bazel test --config=buildfarm-sanitized-mac //core:core_test
```

and it does fail for me with the same message as CI on an M1 MacBook Pro (which is what we also use for CI).

The `buildfarm-sanitized-mac` configuration specifies `--config=shared-libs`.
It looks like Trevor added this [in #2444](https://github.com/sorbet/sorbet/pull/2444/files#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832fR115) without explanation. I'm not sure how load bearing it is that we have that there. I would guess that we don't want to remove that flag, that we're simply lucky that CI managed to flag an issue building with shared libs. For example

```
./bazel test --config=dbg --config=shared-libs //core:core_test
```

Also fails for me. We should probably not regress the ability to build shared libs on macOS.

froydnj:
https://github.com/bazelbuild/bazel/pull/17506 seems relevant, but I'm not sure what the correct fix is, unless we're setting up the tests incorrectly.

Morriar:
It seems to be a regression in Bazel 6 (I couldn't reproduce with 5.X versions): https://github.com/bazelbuild/bazel/issues/11552.

There doesn't seem to be a workaround other than forcing `linkstatic = 1` in the cc_test rule.

froydnj:
> It seems to be a regression in Bazel 6 (I couldn't reproduce with 5.X versions): [bazelbuild/bazel#11552](https://github.com/bazelbuild/bazel/issues/11552).
> 
> There doesn't seem to be a workaround other than forcing `linkstatic = 1` in the cc_test rule.

We already do that ~everywhere for binaries, and if we're missing it, we should include it.

I think it'd be fine to make mac asan builds switch to linkstatic, which I guess is what they already use on Bazel 5.2.0.  We should just make that explicit.  That would involve some `tools/config/BUILD` hackery, but it should not be too bad.

Morriar:
@froydnj I pushed https://github.com/sorbet/sorbet/pull/7827/commits/7c33eadcd7b7451daed14a15048ed83036358586 to ensure `linkshared` wasn't turned on when compiling on another platform that linux, not sure if this is what you had in mind?

jez:
@Morriar sushain landed a change which conflicts with you, the bazel version is now pinned in a different file, not WORKSPACE.

jez:

→ https://go/builds/bui_Pyjmiys9QN459N
→ https://go/builds/bui_PyjmDn8V5GiDss
→ https://go/builds/bui_PyjmJ4WiHwFDjH

PR #7832, opened 2024-04-14
Detailed explanation for T.all type errors

neilparikh:
### Motivation

Fixes https://github.com/sorbet/sorbet/issues/4338

### Test plan

See included automated tests.

### DIFF
--- a/core/types/subtyping.cc
+++ b/core/types/subtyping.cc
@@ -1478,8 +1478,8 @@ bool Types::isSubTypeUnderConstraint(const GlobalState &gs, TypeConstraint &cons
         auto isSubTypeOfLeft = Types::isSubTypeUnderConstraint(gs, constr, t1, a2->left, mode, subCollectorLeft);
         if (!isSubTypeOfLeft) {
             if constexpr (shouldAddErrorDetails) {
-                auto message = ErrorColors::format("`{}` is not a subtype of `{}` (the left side of `{}`)", t1.show(gs),
-                                                   a2->left.show(gs), t2.show(gs));
+                auto message = ErrorColors::format("`{}` is not a subtype of `{}` (the left side of the `{}`)",
+                                                   t1.show(gs), a2->left.show(gs), "T.all");
                 subCollectorLeft.message = message;
                 errorDetailsCollector.addErrorDetails(move(subCollectorLeft));
             }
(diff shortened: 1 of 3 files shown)


test/cli/detailed-errors/intersection.rb
jez, line 1: Can you add some tests that make more nested `T.all` types, like `T.all(M1, M2, M3)`, and `T.all(T.all(M1, M2), M3)` and `T.all(M1, T.all(M2, M3))`?

core/types/subtyping.cc
```diff
@@ -1472,8 +1474,29 @@ bool Types::isSubTypeUnderConstraint(const GlobalState &gs, TypeConstraint &cons
...
 
     if (auto *a2 = cast_type<AndType>(t2)) { // 2, 5
-        return Types::isSubTypeUnderConstraint(gs, constr, t1, a2->left, mode, errorDetailsCollector) &&
-               Types::isSubTypeUnderConstraint(gs, constr, t1, a2->right, mode, errorDetailsCollector);
+        auto subCollectorLeft = errorDetailsCollector.newCollector();
+        auto isSubTypeOfLeft = Types::isSubTypeUnderConstraint(gs, constr, t1, a2->left, mode, subCollectorLeft);
+        if (!isSubTypeOfLeft) {
+            if constexpr (shouldAddErrorDetails) {
+                auto message = ErrorColors::format("`{}` is not a subtype of `{}` (the left side of `{}`)", t1.show(gs),
+                                                   a2->left.show(gs), t2.show(gs));
```
jez, line 1482: What do you think about just saying "(the left side of the T.all)" instead of reprinting the whole type?

I'm worried that for deeply nested types, having the parent type be reprinted is going to make it more noisy and thus hard to see what a given line is uniquely saying.

For example:

```
intersection.rb:27: Expected `T.all(M1, M2, M3, M4)` but found `C1` for argument `x` https://srb.help/7002
...
  Detailed explanation:
    `C1` is not a subtype of `T.all(M2, M3, M4)` (the right side of `T.all(M1, M2, M3, M4)`)
      `C1` is not a subtype of `T.all(M3, M4)` (the right side of `T.all(M2, M3, M4)`)
        `C1` is not a subtype of `M4` (the right side of `T.all(M3, M4)`)
```

versus

```
intersection.rb:27: Expected `T.all(M1, M2, M3, M4)` but found `C1` for argument `x` https://srb.help/7002
...
  Detailed explanation:
    `C1` is not a subtype of `T.all(M2, M3, M4)` (the right side of the T.all)
      `C1` is not a subtype of `T.all(M3, M4)` (the right side of the T.all)
        `C1` is not a subtype of `M4` (the right side of the T.all)
```

(and then imagine that instead of short, two-letter names, they're 50+ character long class names or other compound types)

core/types/subtyping.cc
```diff
@@ -1472,8 +1474,29 @@ bool Types::isSubTypeUnderConstraint(const GlobalState &gs, TypeConstraint &cons
...
+        auto isSubTypeOfRight = Types::isSubTypeUnderConstraint(gs, constr, t1, a2->right, mode, subCollectorRight);
+        if (!isSubTypeOfRight) {
+            if constexpr (shouldAddErrorDetails) {
+                auto message = ErrorColors::format("`{}` is not a subtype of `{}` (the right side of `{}`)",
+                                                   t1.show(gs), a2->right.show(gs), t2.show(gs));
+                subCollectorRight.message = message;
+                errorDetailsCollector.addErrorDetails(move(subCollectorRight));
+            }
+        }
+        return isSubTypeOfLeft && isSubTypeOfRight;
```
jez, line 1499: Is this redundant? Can it be

```suggestion
        return isSubTypeOfRight;
```

?

Also, can we move the `if constexpr` for this half to enclose the `if (!isSubTypeOfRight)`, to make it easier on the optimizer?

jez (approved):

PR #7834, opened 2024-04-15
Don't run all `test/testdata/packager/` tests twice

jez:
### Motivation

This was needlessly making our test times slow. We instantiate two
identical-except-in-name test targets for each `test/testdata/packager`
subfolder.

I noticed this while I was looking to see if there was a convenient Bazel target
for a test suite contained with only the `test/testdata/packager/` tests.

### Test plan

This could potentially be deleting tests if we're not sure we know how this
works, so the test plan is Feynman method.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -247,17 +247,6 @@ pipeline_tests(
 )
 
 pipeline_tests(
-    "test_packager",
-    glob(
-        [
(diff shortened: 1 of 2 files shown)


froydnj (approved):
I verified that the set of tests ran by `test_packager` was exactly the same as the packager tests ran by `test_corpus`.

PR #7830, opened 2024-04-12
backport ruby/ruby#9385 to fix aarch64 crashes

froydnj:
### Motivation

This is going to be in the Ruby 3.3.1 release, but until that time...

Special thanks to @paracycle for pointing out the PR!

### Test plan

See included automated tests.

PR #7836, opened 2024-04-16
patch configure for aarch64 branch protection backport

froydnj:
### Motivation

In #7830, we backported a patch from upstream to fix a crash on aarch64, which patched `configure.ac`.  For upstream, this is fine; development regenerates `configure` as-needed.  For `sorbet_ruby` builds, however, `configure` comes pre-generated in the tarball, and we do not automatically regenerate it.

Ergo, the patch was ineffective.  With this PR, we are also patching the generated file to do what it's supposed to.

cc @jd-stripe

### Test plan

Going to verify this builds on Stripe's infrastructure, and compare the aarch64 ruby builds from before and after.

### DIFF
--- a/third_party/ruby/9385-aarch64-fibers-crash-backport.patch
+++ b/third_party/ruby/9385-aarch64-fibers-crash-backport.patch
@@ -25,3 +25,34 @@ index 9286946fc1..18b4247991 100644
                  break
              ])
          ])
+--- a/configure	2024-04-16 14:03:03.035215032 +0000
++++ b/configure	2024-04-16 16:43:00.603560413 +0000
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_Pw7etSCxvjbPV3
→ https://go/builds/bui_Pw7eZGE1roF2JL
→ https://go/builds/bui_Pw7euKBPORVfVq

froydnj:
✅ `coroutine_transfer` from `coroutine/arm64/Context.S` in this build has the PAC instructions, which indicates the assembler is getting the correct flags.  Previous builds did not have the PAC instructions.

neilparikh (approved):

PR #7837, opened 2024-04-16
Reimplement what keep_for_ide did in DefLocSaver

jez:
### Motivation

It seems like we were using this `keep_for_ide` method only so that it
would get handled by our `postTransformConstantLit` callback in this
treewalk. But that's a lot of allocations to add to the tree when we could
just make `DefLocSaver` manually check the two places that have extra
constant lits.

This potentially removes a ton of allocations:

- All the `ConstantLit` / `UnresolvedConstantLit` that come from
  `deepCopy`'ing the constant in question.
- The `ConstantLit` for the `Sorbet::Private::Static`
- The `Send` for the `keep_for_ide` call

I don't think this removes the `<static-init>` methods if this causes any
`<static-init>` methods to be empty, but maybe that's an optimization we
could make.

### Test plan

Existing tests.

On Stripe's codebase, this change drops max_rss memory usage by about 3.6%, as measured with `--counters`.


jez:

→ https://go/builds/bui_PwCerqSKc6hfWr
→ https://go/builds/bui_PwCeGT5MvDIcXL
→ https://go/builds/bui_PwCek4yoIxEcRf

froydnj (changes requested):
Very nice.

`rg -B 1 '\^ usage:' test/testdata/lsp/|rg -A 1 class` only turns up a single place that might exercise this logic, and even that is for a class that doesn't exist.

I think we should add some tests that exercise `DefLocSaver` (via jump-to-def/find references/etc.) on superclasses before committing this.

froydnj (approved):
Ah, I didn't think about hover.  I think the `hover_ancestors.rb` test exercises this codepath.

PR #7838, opened 2024-04-17
Stop shuffling classes around

jez:
### Motivation

Mostly I'm curious if we can get away with not doing this.

Maybe there's a bit of a performance win? Maybe there's a bit of a memory win,
because memory locality is a little better?

Anyways, mostly just trying to delete code.

We might want to rename this to something other than `class_flatten`, because
the only thing it's doing now is extracting `<static-init>` methods.

### Test plan

Existing tests


jez:

→ https://go/builds/bui_PwEoPwJrVU2JzV
→ https://go/builds/bui_PwEp8xKCf8W5DL
→ https://go/builds/bui_PwEpUTodiGSmkX

jez:
This does not seem to meaningfully affect wall time nor memory, and breaks a bunch of compiler tests, so I'm going to put this down for the time being.

PR #7840, opened 2024-04-17
Skip making empty `<static-init>` MethodDefs

jez:
### Motivation

This doesn't affect anything, but does make it easier to work with `cfg-view.sh`
because if you don't have anything inside a class or file top level, you won't
get an extra `<static-init>` CFG in the output.

### Test plan

Existing tests


jez:
Ah, this breaks the compiler.

I'd like to revive this if we decide to completely deprecate the compiler.

PR #7846, opened 2024-04-19
Extract `to_str` to T::Enum

KaanOzkan:
### Motivation

Responding to `to_str` allows ruby to call `T::Enum#==` during string comparisons. This feature was used by T::Enum's before the recent `LegacyMigrationMode` mixin change and I think there's value in keeping it so that all T::Enum's can define their own `==` method.

### Test plan

I can add one if it's helpful.

### DIFF
--- a/gems/sorbet-runtime/lib/types/enum.rb
+++ b/gems/sorbet-runtime/lib/types/enum.rb
@@ -185,6 +185,30 @@ class T::Enum
     end
   end
 
+  # NB: Do not call this method. This exists to allow for a safe migration path in places where enum
+  # values are compared directly against string values.
+  #
+  # Ruby's string has a weird quirk where `'my_string' == obj` calls obj.==('my_string') if obj
+  # responds to the `to_str` method. It does not actually call `to_str` however.
+  #
+  # See https://ruby-doc.org/core-2.4.0/String.html#method-i-3D-3D
+  T::Sig::WithoutRuntime.sig {returns(String)}
+  def to_str
+    msg = 'Implicit conversion of Enum instances to strings is not allowed. Call #serialize instead.'
+    if T::Configuration.legacy_t_enum_migration_mode?
+      T::Configuration.soft_assert_handler(
+        msg,
+        storytime: {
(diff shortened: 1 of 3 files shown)


gems/sorbet-runtime/lib/types/enum.rb
```diff
@@ -185,6 +185,30 @@ def <=>(other)
...
+  # values are compared directly against string values.
+  #
+  # Ruby's string has a weird quirk where `'my_string' == obj` calls obj.==('my_string') if obj
+  # responds to the `to_str` method. It does not actually call `to_str` however.
+  #
+  # See https://ruby-doc.org/core-2.4.0/String.html#method-i-3D-3D
+  T::Sig::WithoutRuntime.sig {returns(String)}
+  def to_str
+    msg = 'Implicit conversion of Enum instances to strings is not allowed. Call #serialize instead.'
+    if T::Configuration.legacy_t_enum_migration_mode?
```
KaanOzkan, line 198: I couldn't get `LegacyMigrationMode` to override this method without this check so I extracted the whole method instead. Given that it's not used for non `LegacyMigrationMode` enums it shouldn't make a difference in performance.

amomchilov:
> Responding to `to_str` allows ruby to call `T::Enum#==` during string comparisons.

Could you elaborate on this? I couldn't replicate:

```ruby
class MyStr
  def initialize(s)
    @s = s
  end

  def to_str = @s
end

"abc" == MyStr.new("abc") # => false
MyStr.new("abc") == "abc" # => false
```

gems/sorbet-runtime/lib/types/enum.rb
amomchilov: The override from `LegacyMigrationMode` could only work if `include LegacyMigrationMode` comes _after_ this method definition

amomchilov (approved):
TIL about this very strange Ruby behaviour!

egiurleo:
Whoah I did not know about this strange Ruby behavior!

jez (approved):

PR #7849, opened 2024-04-23
Remove the Sorbet Compiler

jez:
### Motivation

This is not getting actively developed anymore, and it's getting in the way of
upgrading the build system in toilsome ways, so we're going to remove it. The
sources will be available in the git history.

~I've also removed most of the single package RBI generation tests because my
experience with them is that they weren't actually catching bugs, and it was the
only other thing running on the buildkite compiler job (because it needed Ruby
at runtime to run like one or two genrules).~

I rewrote the single package RBI generation in C++ in #7856 so that we could keep it around while not having to build + run it with Ruby.

### Test plan

See included automated tests.


froydnj (approved):
`:salute2:`

PR #7860, opened 2024-04-25
fix find all references for the `Command` rewriter

### LINKED ISSUE #7858: Support find all references for `Opus::Command`
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Opus%3A%3ACommand%0A%20%20extend%20T%3A%3ASig%0A%20%20def%20self.call%28*args%2C%20**kwargs%2C%20%26blk%29%0A%20%20end%0Aend%0A%0Aclass%20MyCommand%20%3C%20Opus%3A%3ACommand%0A%20%20sig%20%7B%20params%28x%3A%20Integer%2C%20y%3A%20Integer%29.void%20%7D%0A%20%20def%20call%28x%2C%20y%29%0A%20%20%20%20%23%20%5E%20find%20all%20refs%20here%20doesn't%20find%20anything%0A%20%20end%0Aend%0A%0AMyCommand.call%281%2C%202%29%0A%23%20%20%20%20%20%20%20%20%20%5E%20versus%20find%20all%20refs%20here%20does%0AMyCommand.call%283%2C%204%29)

```ruby
# typed: true

class Opus::Command
  extend T::Sig
  def self.call(*args, **kwargs, &blk)
  end
end

class MyCommand < Opus::Command
  sig { params(x: Integer, y: Integer).void }
  def call(x, y)
    # ^ find all refs here doesn't find anything
  end
end

MyCommand.call(1, 2)
#         ^ versus find all refs here does
MyCommand.call(3, 4)
```

#### Observed output

```
No errors! Great job.
```

Find all refs on the call site shows the references.

Find all refs on the method definition does not.

#### Expected behavior

Finding all references on the method definition should show the call sites.

- - -

froydnj:
### Motivation

Fixes #7858 

This is admittedly kind of a weird thing to do, but it is consistent with how we hide other things from LSP -- even though the other things we are hiding are usually synthesized methods, rather than the real thing.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Command.cc
+++ b/rewriter/Command.cc
@@ -41,7 +41,8 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
     }
 
     int i = 0;
-    ast::MethodDef *call;
+    ast::MethodDef *call = nullptr;
+    ast::ExpressionPtr *callptr = nullptr;
 
     for (auto &stat : klass->rhs) {
         auto *mdef = ast::cast_tree<ast::MethodDef>(stat);
@@ -54,6 +55,7 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
 
         i = &stat - &klass->rhs.front();
         call = mdef;
+        callptr = &stat;
         break;
     }
     // If we didn't find a `call` method, or if it was the first statement (and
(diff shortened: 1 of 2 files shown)


rewriter/Command.cc
```diff
@@ -84,9 +86,24 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
     ast::MethodDef::Flags flags;
     flags.isSelfMethod = true;
     flags.discardDef = true;
-    auto selfCall = ast::MK::SyntheticMethod(call->loc, call->loc, call->name, std::move(newArgs),
+    auto selfCall = ast::MK::SyntheticMethod(call->loc, call->declLoc, call->name, std::move(newArgs),
```
jez, line 89: Was this just an unrelated bug or was this directly implicated in this change?

rewriter/Command.cc
```diff
@@ -84,9 +86,24 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
...
                                              ast::MK::RaiseTypedUnimplemented(call->declLoc), flags);
 
+    // We are now in the weird situation where we have an actual method that
+    // the user has written, but we have a synthetic method that lives at the
+    // same location.  If we try to jump-to-def from the actual method, there
```
jez, line 94: s/jump to def/find all refs/

?

rewriter/Command.cc
```diff
@@ -84,9 +86,24 @@ void Command::run(core::MutableContext ctx, ast::ClassDef *klass) {
...
+    // are no calls to it, which will frustrate the user.  Erase the location(s)
+    // on the non-synthetic method so that LSP only sees the synthetic method.
```
jez, line 96: Did you consider an alternative solution which ordered the synthetic and non synthetic methods in the reverse order, so that the LSPQuery::byLoc has the synthetic one at index 0 instead?

jez (approved):
I see that our tests pass, but I’m guessing that’s because we don’t have many tests that would fail if we had a bad loc on the instance method’s MethodDef node. For example, this loc is used for things like rename, “defined here” errors, “convert to singleton class method,” the uninitialized nil location, any autocorrects that we may or may not have around inserting/deleting/renaming args in a method def (this one might not actually be a function of the MethodDef).

Testing on pay-server is not going to be indicative of whether something broke here because this error will only surface in errors. I’d love if we could spend a little bit of time tinkering with this to see if it breaks something. Even if it doesn’t mean committing new tests, just trying to see if we can find a test case that might break.

The reason why I suggested the alternative implementation is that I think that approach is less likely to break something unrelated, but it’s possible that it just has a different set of things that break, and this change is an improvement over the current state either way.

rewriter/Command.cc
froydnj: This was directly implicated in the change; the tests did not work without this change.

rewriter/Command.cc
froydnj: We tried the solution where we flipped the ordering of the `call` definitions offline and it did not work, so we're going with the approach in the PR.

PR #7863, opened 2024-04-26
Allow `T.let` types to affect lambda inference

jez:
### Motivation

This lets `T.let` annotations affect the inference of lambdas.

Right now it ignores `proc` and `lambda` methods, and only works for `->(){}`
lambda syntax, but I think we could expand support to include those.

Before this PR, when you wrap a lambda in a `T.let`, it doesn't affect how the
body of the lambda is type checked. The arguments are still assumed to be
untyped, and the result is also untyped. Then the `T.let` comes along and
ascribes a type, which might be out of sync with what the lambda's true type
actually is.

After this PR, we allow the body of the lambda to be typechecked assuming that
the `T.let` type holds, and report errors if that assumption doesn't hold. It's
powered by the exact same mechanism that powers type checking for blocks, which
makes the implementation very simple.

### Test plan

See included automated tests.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -169,6 +169,20 @@ InstructionPtr maybeMakeTypeParameterAlias(CFGContext &cctx, ast::Send &s) {
     return make_insn<Alias>(typeParam);
 }
 
+ast::Send *isKernelLambda(ast::ExpressionPtr &expr) {
+    auto *send = ast::cast_tree<ast::Send>(expr);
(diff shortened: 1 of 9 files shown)


jez:

→ https://go/builds/bui_PztT41FbAOXP1f
→ https://go/builds/bui_PztTTYJX3ZsFUw
→ https://go/builds/bui_PztTxTThHsaMxa

jez:
@neilparikh I fixed the two tests

neilparikh (approved):
> Right now it ignores proc and lambda methods, and only works for ->(){}
lambda syntax, but I think we could expand support to include those.

This part of the PR description should be updated. LGTM otherwise.

PR #7864, opened 2024-04-26
Add ruby.wasm artifact with sorbet-runtime bundled

maruth-stripe:
### Motivation

To allow executing code on `sorbet.run` with sorbet-runtime 

### Test plan

See included automated tests.

### DIFF
--- a/.buildkite/build-ruby-wasm.sh
+++ b/.buildkite/build-ruby-wasm.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -eo pipefail
+
+mkdir -p /tmp/ruby-wasm
+pushd /tmp/ruby-wasm
+
+echo "--- setup :ruby:"
+eval "$(rbenv init -)"
+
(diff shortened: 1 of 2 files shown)


.buildkite/build-ruby-wasm.sh
```diff
@@ -0,0 +1,23 @@
...
+eval "$(rbenv init -)"
+
+runtime_version=3.1.2
+rbenv install --skip-existing "$runtime_version"
+
+echo "3.1.2" > .ruby-version
+
+bundle init
+bundle add ruby_wasm
+bundle add sorbet-runtime
```
jez, line 18: What do you think about creating a Gemfile + Gemfile.lock in a separate folder somewhere that has these, so that we can lock them?

There's a way to specify sorbet-runtime in a Gemfile as a local dependency with a relative path. You could do something like

```ruby
gem 'ruby_wasm'
gem 'sorbet-runtime', path: '../gems/sorbet-runtime'
```

PR #7874, opened 2024-05-06
Don't dup documentationLocations when not aliased

jez:
### Motivation

The changes in 9c90faa74 accidentally made all documentation for
constants that were not constant aliases duplicated.

We should definitely include both the pre- and post-dealiased
documentation locations if the symbols are different, but if they're the
same, then it's just going to mean showing two copies of the
documentation.

### Test plan

It's hard to write a test for this, because our `hover` assertions are just
"does the documentation include this text" which makes it hard to say that the
hovered text **doesn't** contain something, and is also why none of our existing
tests caught this regression.

I'm going to punt on changing the test suite for that because I don't think this
issue is so pressing that we want to make sure that it never gets reintroduced.

To that end, I tested this manually in my editor, sending a hover request, observing that before this change there were two copies of the documentation, and after there was not.

**Before**
Screenshot 2024-05-06 at 2 34 28 PM
**After**
Screenshot 2024-05-06 at 2 35 14 PM

### DIFF
--- a/main/lsp/requests/hover.cc
+++ b/main/lsp/requests/hover.cc
@@ -88,9 +88,12 @@ unique_ptr<ResponseMessage> HoverTask::runRequest(LSPTypecheckerDelegate &typech
                 documentationLocations.emplace_back(loc);
             }
         }
-        for (auto loc : c->symbolBeforeDealias.dealias(gs).locs(gs)) {
-            if (loc.exists()) {
-                documentationLocations.emplace_back(loc);
+        auto dealiased = c->symbolBeforeDealias.dealias(gs);
+        if (dealiased != c->symbolBeforeDealias) {
+            for (auto loc : dealiased.locs(gs)) {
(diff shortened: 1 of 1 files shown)


froydnj (approved):
We could write a really hacky test for all hover tests that looked at the plaintext (if markdown) or just the text, cut it in two, and ensured that both halves were not the same thing?  Might have to make allowances for the newline joining the two halves or something like that -- maybe equality after trimming whitespace off the ends of both.

jez:
> We could write a really hacky test for all hover tests that looked at the plaintext (if markdown) or just the text, cut it in two, and ensured that both halves were not the same thing? Might have to make allowances for the newline joining the two halves or something like that -- maybe equality after trimming whitespace off the ends of both.

I think the way I would have solved this would have been another assertion that says like "this is the full contents of the hover or something like that, like `hover: \Athe text\z` or something

PR #7875, opened 2024-05-07
Add Gem::BUNDLED_GEMS to rubygems.rbi

iMacTia:
### Motivation
The `Gem::BUNDLED_GEMS` module introduced in Ruby 3.3 is currently unknown to sorbet and causes "unable to resolve constant ..." errors.
This module is currently undocumented, but its implementation can be seen [here](https://github.com/ruby/ruby/blob/master/lib/bundled_gems.rb).
cc @2rba who noticed and raised this with me.


serg-indeed (approved):

froydnj (approved):
Thanks!

PR #7877, opened 2024-05-09
Insert completion snippet when in front of local var

jez:
### Motivation

Reported by a user.

### Test plan

See included automated tests.

There are two commits: the first commit adds a failing test and the second
commit fixes the code so the test passes.

### DIFF
--- a/test/helpers/position_assertions.cc
+++ b/test/helpers/position_assertions.cc
@@ -1494,7 +1494,7 @@ shared_ptr<ApplyCompletionAssertion> ApplyCompletionAssertion::make(string_view 
 
     ADD_FAIL_CHECK_AT(
         string(filename).c_str(), assertionLine + 1,
-        fmt::format("Improperly formatted apply-completion assertion. Expected '[<version>] <index>'. Found '{}'",
+        fmt::format("Improperly formatted apply-completion assertion. Expected '[<version>] item: <index>'. Found '{}'",
                     assertionContents));
 
     return nullptr;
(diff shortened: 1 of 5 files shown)


test/helpers/position_assertions.cc
```diff
@@ -1494,7 +1494,7 @@ shared_ptr<ApplyCompletionAssertion> ApplyCompletionAssertion::make(string_view
 
     ADD_FAIL_CHECK_AT(
         string(filename).c_str(), assertionLine + 1,
-        fmt::format("Improperly formatted apply-completion assertion. Expected '[<version>] <index>'. Found '{}'",
+        fmt::format("Improperly formatted apply-completion assertion. Expected '[<version>] item: <index>'. Found '{}'",
```
jez, line 1497: This change is unrelated but I ran into it while writing a test case so I figured I'd just fix it in this PR.

test/testdata/lsp/completion/snippet_before_variable.A.rbedited
```diff
@@ -0,0 +1,22 @@
...
+begin
+  A.
+  # ^ completion: example, ...
+  # ^ apply-completion: [B] item: 0
+
+  x = nil # error: does not exist
+
+  A.example(${1:Integer})${0}
+  # ^ completion: example, ...
+  # ^ apply-completion: [A] item: 0
```
jez, line 20: This shows what the intended behavior is, and what it actually is when the completion happens at the end of a file or in a spot where the next token doesn't look like a method name.

froydnj (commented):
Ah, this will be a nice change.

froydnj (approved):

PR #7878, opened 2024-05-10
Fix crash in extract variable

neilparikh:
For something like the following, we create an `EmptyTree` node for the missing parts (the `thenp` for the `unless`, and `body` and `else` for the `rescue`). However, these nodes don't have a `loc`, so the `contains` call fails.
```ruby
unless 1 == 2
  puts 2 + 3
end

begin
ensure
  6 + 6
end
```

### Motivation

Fix a crash.

### Test plan

See included automated tests.


jez (approved):
Nice

PR #7884, opened 2024-05-13
Update abseil-cpp to version 20240116.2

jez:
⚠️ I do not plan to land this change on its own—instead, this is just here to showcase this subset of ~mostly independent changes in the larger stack of change to upgrade our bazel toolchain.

### Motivation

This is a pre-requisite for upgrading emscripten. The absl codebase has some conditional defines that do one thing or another depending on whether they're being built with emscripten.

### Test plan

This branch on its own won't pass, because we haven't upgraded emscripten yet. See the stacked PR.

### DIFF
--- a/core/Symbols.h
+++ b/core/Symbols.h
@@ -654,7 +654,7 @@ private:
 
     void addMixinAt(ClassOrModuleRef sym, std::optional<uint16_t> index);
 };
-CheckSize(ClassOrModule, 128, 8);
+CheckSize(ClassOrModule, 120, 8);
 
 } // namespace sorbet::core
 #endif // SORBET_SYMBOLS_H
--- a/third_party/externals.bzl
+++ b/third_party/externals.bzl
@@ -159,15 +159,9 @@ def register_sorbet_dependencies():
 
     http_archive(
         name = "com_google_absl",
-        url = "https://github.com/abseil/abseil-cpp/archive/8910297baf87e1777c4fd30fb0693eecf9f2c134.zip",
-        sha256 = "c43b8cd8e306e7fe3f006d880181d60db59a3bae6b6bc725da86a28a6b0f9f30",
-        strip_prefix = "abseil-cpp-8910297baf87e1777c4fd30fb0693eecf9f2c134",
-        patches = [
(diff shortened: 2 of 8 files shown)


neilparikh (approved):

third_party/externals.bzl
```diff
@@ -159,15 +159,9 @@ def register_sorbet_dependencies():
 
     http_archive(
         name = "com_google_absl",
-        url = "https://github.com/abseil/abseil-cpp/archive/8910297baf87e1777c4fd30fb0693eecf9f2c134.zip",
-        sha256 = "c43b8cd8e306e7fe3f006d880181d60db59a3bae6b6bc725da86a28a6b0f9f30",
-        strip_prefix = "abseil-cpp-8910297baf87e1777c4fd30fb0693eecf9f2c134",
-        patches = [
-            # https://github.com/abseil/abseil-cpp/commit/d2422b19e9ba41c952db1ed5514bb68114c2be15
-            # Abseil builds with `-Wall` which we can't override to silence this warning.
-            "@com_stripe_ruby_typer//third_party:abseil_cpp/low_level_hash_array_parameter.patch",
```
froydnj, line 168: We can also `git rm` this file as part of this (eventually bundled) PR?

core/Symbols.h
```diff
@@ -654,7 +654,7 @@ class ClassOrModule final {
 
     void addMixinAt(ClassOrModuleRef sym, std::optional<uint16_t> index);
 };
-CheckSize(ClassOrModule, 128, 8);
+CheckSize(ClassOrModule, 120, 8);
```
froydnj, line 657: Whoa, what is this from?  Did `UnorderedMap` get smaller?

froydnj (approved):

third_party/externals.bzl
jez: oh good catch, looks like I forgot to do that. I will do that on the top of the stack.

core/Symbols.h
jez: The commit message that introduced this shows some of my debugging notes, but I was never able to pinpoint an exact abseil commit that blames to this change.

jez: https://github.com/sorbet/sorbet/pull/7884/commits/083d8114bafb0952b77da8b21806857cb9b3884b

PR #7886, opened 2024-05-13
Upgrade to Clang 15

jez:
### Motivation

- It's nice to be on more recent versions of our compiler.
  In particular, it looks like it might have fixed a bug which means that
  `ClassOrModule` is now 8 bytes smaller.

- This will get us closer to being able to use the upstream
  `bazel-contrib/toolchains_llvm` instead of our custom fork of
  `grailbio/bazel-toolchain`. In particular, there seems to be something about
  the way that the upstream invokes the linker such that our trick to get builds
  working on macOS Sonoma again (bf58b6293) does not work anymore.

  Our current hypothesis for how to work around that is to switch away from
  using `/usr/bin/ld` to link and instead to using the `lld.ld` from the LLVM
  toolchain, which we think will give us a better ability to statically link
  against `libc++` on macOS (despite the OS's best effort to make that not
  possible).

- Upgrading `toolchains_llvm` is important because starting in Bazel 7 the
  option `--incompatible_enable_cc_toolchain_resolution` will switch to
  being on by default, and Sorbet's fork of bazel-toolchain uses the old
  version. Also the upstream `bazel-contrib/toolchains_llvm` has better support
  for building on Apple Silicon.

### Test plan

Existing tests

I also plan to do some manual testing.

### DIFF
--- a/cfg/CFG.cc
+++ b/cfg/CFG.cc
@@ -361,7 +361,7 @@ string BasicBlock::toString(const core::GlobalState &gs, const CFG &cfg) const {
     fmt::memory_buffer buf;
     fmt::format_to(std::back_inserter(buf), "block[id={}, rubyRegionId={}]({})\n", this->id, this->rubyRegionId,
                    fmt::map_join(
-                       this->args, ", ", [&](const auto &arg) -> auto { return arg.toString(gs, cfg); }));
+                       this->args, ", ", [&](const auto &arg) -> auto{ return arg.toString(gs, cfg); }));
 
     if (this->outerLoops > 0) {
         fmt::format_to(std::back_inserter(buf), "outerLoops: {}\n", this->outerLoops);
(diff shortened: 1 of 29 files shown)


cfg/CFG.cc
```diff
@@ -361,7 +361,7 @@ string BasicBlock::toString(const core::GlobalState &gs, const CFG &cfg) const {
     fmt::memory_buffer buf;
     fmt::format_to(std::back_inserter(buf), "block[id={}, rubyRegionId={}]({})\n", this->id, this->rubyRegionId,
                    fmt::map_join(
-                       this->args, ", ", [&](const auto &arg) -> auto { return arg.toString(gs, cfg); }));
+                       this->args, ", ", [&](const auto &arg) -> auto{ return arg.toString(gs, cfg); }));
```
froydnj, line 364: These are very weird clang-format changes.

froydnj (approved):
🎉

PR #7887, opened 2024-05-14
Report more "unexpected bare value" errors

jez:
### Motivation

While helping a user attempt to write some code today, I noticed that they had
this in their code and it did the right thing statically but only worked at
runtime because generics are erased (`T::Array[x]`). Had they attempted to do
something similar with `T.class_of` or `T.any` or something else, it would not
have done the thing they hoped it would have.

Ideally an even better error message would include something like an autocorrect
like "Did you meant to explicitly annotate this value's type here?" and offer to
replace `x` in the test with `T.type_parameter(:U)`, or whatever the type of `x`
is. That isn't always going to be the right suggestion, but it will at least get
the user closer to understanding what the problem is.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/types.cc
+++ b/core/types/types.cc
@@ -982,13 +982,12 @@ TypePtr Types::unwrapType(const GlobalState &gs, Loc loc, const TypePtr &tp) {
             unwrappedElems.emplace_back(unwrapType(gs, loc, elem));
         }
         return make_type<TupleType>(move(unwrappedElems));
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_Q6f3GZMJfpyBAv
→ https://go/builds/bui_Q6f3FuQSzgatfj
→ https://go/builds/bui_Q6f3iBy7i1Dkg2

froydnj (approved):

PR #7893, opened 2024-05-17
Extract all occurrences to variable

### LINKED ISSUE #7391: IDE: Extract expression into variable
### Problem

Similar to https://github.com/sorbet/sorbet/issues/6776 it would be great if Sorbet could extract an expression into a variable. 
See also [Intellij](https://www.jetbrains.com/idea/guide/tips/extract-variable-2/) documentation for a similar feature.

### Proposed solution

N/A

neilparikh:
Review commit-by-commit.

### Motivation

Closes #7391

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -5,6 +5,75 @@ using namespace std;
 
 namespace sorbet::realmain::lsp {
 
+core::LocOffsets findWhereToInsert(const ast::ExpressionPtr &scope, const core::LocOffsets target) {
+    // The ENFORCE(!ast::isa_tree<ast::InsSeq>(...)) are there check that the enclosingScope returned
+    // by the TreeWalk doesn't contain a further InsSeq, because the preTransformInsSeq should have
+    // matched on that (if it contains the selectionLoc).
+    core::LocOffsets whereToInsert = core::LocOffsets::none();
+    if (auto insSeq = ast::cast_tree<ast::InsSeq>(scope)) {
+        for (auto &stat : insSeq->stats) {
+            if (stat.loc().contains(target)) {
+                ENFORCE(!ast::isa_tree<ast::InsSeq>(stat));
+                whereToInsert = stat.loc();
+                break;
+            }
+        }
+        if (insSeq->expr.loc().contains(target)) {
+            ENFORCE(!ast::isa_tree<ast::InsSeq>(insSeq->expr));
+            whereToInsert = insSeq->expr.loc();
+        }
+    } else if (auto classDef = ast::cast_tree<ast::ClassDef>(scope)) {
+        if (classDef->rhs.size() == 0) {
+            ENFORCE(false);
+        } else {
+            for (auto &stat : classDef->rhs) {
+                if (stat.loc().contains(target)) {
+                    ENFORCE(!ast::isa_tree<ast::InsSeq>(stat));
+                    whereToInsert = stat.loc();
+                    break;
+                }
+            }
+        }
+    } else if (auto block = ast::cast_tree<ast::Block>(scope)) {
+        ENFORCE(!ast::isa_tree<ast::InsSeq>(block->body));
+        whereToInsert = block->body.loc();
+    } else if (auto methodDef = ast::cast_tree<ast::MethodDef>(scope)) {
+        ENFORCE(!ast::isa_tree<ast::InsSeq>(methodDef->rhs));
+        whereToInsert = methodDef->rhs.loc();
+    } else if (auto if_ = ast::cast_tree<ast::If>(scope)) {
+        if (if_->thenp.loc().exists() && if_->thenp.loc().contains(target)) {
+            ENFORCE(!ast::isa_tree<ast::InsSeq>(if_->thenp));
+            whereToInsert = if_->thenp.loc();
+        } else {
+            ENFORCE(!ast::isa_tree<ast::InsSeq>(if_->elsep));
+            whereToInsert = if_->elsep.loc();
+        }
+    } else if (auto rescue = ast::cast_tree<ast::Rescue>(scope)) {
+        if (rescue->body.loc().exists() && rescue->body.loc().contains(target)) {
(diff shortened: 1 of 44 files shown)


jez (commented):
I have to run an errand this morning, but I will be finishing this up today. I'm only part way through reviewing commits up to "Tree Deep Equality." I haven't finished that commit, nor any after it.

main/lsp/ExtractVariable.cc
```diff
@@ -268,4 +289,216 @@ vector<unique_ptr<TextDocumentEdit>> VariableExtractor::getEdits(LSPTypecheckerD
...
+    vector<const ast::ExpressionPtr *> enclosingScopeStack;
+    std::vector<core::LocOffsets> skippedLocs;
+
+    // NOTE: Might want to profile and switch to UnorderedSet.
+    bool shouldSkipLoc(core::LocOffsets loc) {
+        return absl::c_find_if(skippedLocs, [loc](auto l) { return l.contains(loc); }) != skippedLocs.end();
+    }
+
+public:
+    vector<pair<vector<const ast::ExpressionPtr *>, const ast::ExpressionPtr *>> matches;
```
jez, line 305:
```suggestion
    vector<pair<vector<const ast::ExpressionPtr *>, core::LocOffsets>> matches;
```

Why are we storing an ExpressionPtr here? From my read of the code, we're only ever using `matches[N].second->loc()`.

In favor of storing as few raw pointers as possible, I'd love if we can make this change, assuming my read of the code is right.

main/lsp/ExtractVariable.cc
```diff
@@ -268,4 +289,216 @@ vector<unique_ptr<TextDocumentEdit>> VariableExtractor::getEdits(LSPTypecheckerD
...
+            matches.push_back(pair(enclosingScopeStack, &tree));
+        }
+    }
+    void preTransformInsSeq(core::Context ctx, const ast::ExpressionPtr &tree) {
+        enclosingScopeStack.push_back(&tree);
+    }
+
+    void postTransformInsSeq(core::Context ctx, const ast::ExpressionPtr &tree) {
+        enclosingScopeStack.pop_back();
+    }
```
jez, line 335: Why are there cases here that aren't in the cases below, which only handle `If`, `Rescue`, and `While`?

main/lsp/ExtractVariable.cc
```diff
@@ -268,4 +289,216 @@ vector<unique_ptr<TextDocumentEdit>> VariableExtractor::getEdits(LSPTypecheckerD
...
+
+    // There should be at least one match (the original selected expression).
+    ENFORCE(matches.size() > 0);
+    if (matches.size() == 1) {
+        return std::pair(vector<unique_ptr<TextDocumentEdit>>(), 1);
+    }
+    fast_sort(
+        matches, [](auto a, auto b) -> auto { return a.second->loc().beginPos() < b.second->loc().beginPos(); });
+
+    // Find LCA enclosing scope
```
jez, line 407: Could this algorithm be substantially simplified if it were agnostic of node type, and instead used pointer equality?

For example, given

```
matches = [
  [0x1, 0x2, 0x3],
  [0x1, 0x2, 0x4],
  [0x1, 0x5, 0x6],
]
```

We could loop over the array (in column major order, which is a bit weird because you'd need to know the max sub-vector's length to do that), and say "are all of these enclosing scopes the same (by pointer equality) at this depth? if so, this is our new LCA, go to the next index. Otherwise, break out of the loop."

And at that point, is there any reason why we couldn't move this computation to be done **during** the tree walk? That way, we would only need to keep "current tightest LCA stack" and "current enclosing scope stack", which uses memory linear in the size of the tree, not quadratic in the size of the tree.

jez (commented):
Sorry for the delay.

main/lsp/ExtractVariable.cc
```diff
@@ -131,110 +230,32 @@ class ExtractVariableWalk {
...
-                "loc on node doesn't match the loc found in the tree walk");
-        whereToInsert = rescueCase->body.loc();
-    } else if (auto while_ = ast::cast_tree<ast::While>(*enclosingScope)) {
-        ENFORCE(!ast::isa_tree<ast::InsSeq>(while_->body));
-        ENFORCE(while_->body.loc() == extractVariableWalk.enclosingScopeLoc,
-                "loc on node doesn't match the loc found in the tree walk");
-        whereToInsert = while_->body.loc();
+    auto enclosingScope = walk.enclosingScope;
+    auto whereToInsert = findWhereToInsert(*enclosingScope, locOffsets);
+    // TODO: can we avoid deepCopy?
```
jez, line 251: Flagging TODO

main/lsp/ExtractVariable.cc
jez: Yeah I think this makes sense. Thanks!

neilparikh:

→ https://go/builds/bui_QQF7DxP1Nrp5iv
→ https://go/builds/bui_QQF7zQW2q4bfwQ
→ https://go/builds/bui_QQF8PnsKfw61u9

main/lsp/ExtractVariable.cc
```diff
@@ -131,110 +230,31 @@ class ExtractVariableWalk {
...
-                "loc on node doesn't match the loc found in the tree walk");
-        whereToInsert = rescueCase->body.loc();
-    } else if (auto while_ = ast::cast_tree<ast::While>(*enclosingScope)) {
-        ENFORCE(!ast::isa_tree<ast::InsSeq>(while_->body));
-        ENFORCE(while_->body.loc() == extractVariableWalk.enclosingScopeLoc,
-                "loc on node doesn't match the loc found in the tree walk");
-        whereToInsert = while_->body.loc();
+    auto enclosingScope = walk.enclosingScope;
+    auto whereToInsert = findWhereToInsert(*enclosingScope, locOffsets);
+    matchingNode = walk.matchingNode->deepCopy();
```
jez, line 251: I remain convinced it should be possible to remove this copy, but I think the change is in a good enough state for the time being and we can circle back to deal with this.

jez (approved):

PR #6076, opened 2022-07-22
Use polarity of type variable to approximate type

jez:
### Motivation

This eliminates basically ever class of generic method errors that show `<top>`.

This directly:

Fixes #38
Fixes #3053

and will unblock letting us land changes for some methods in the standard
library that suffered from this bug, including: `Enumerable#reduce` (#2770),
`Enumerable#zip` (#2848), `Enumerable#inject` (#1361),
`Enumerable#each_with_object` (#1098), `Hash#merge` (#1098).

### Explanation

The intuition is basically that given:

```ruby
def foo(x); end
foo(0)
```

Sorbet will check whether `0 <: x` under the given constraints. If `x` has type
`T.type_parameter(:U)`, and the constraint is not yet solved, then it will
remember that `0` must be a lower bound for `T.type_parameter(:U)` in order for
the constraint to solve. It will **not** record an upper bound for that type
parameter unless you did something like

```ruby
def foo(x, f); end
f = T.let(->(x) {}, T.proc.params(x: Integer).void)
foo(0, f)
```

which would ask

```ruby
T.proc.params(x: Integer).void  <:  T.proc.params(x: T.type_parameter(:U)).void
# and then, because the Arg0 type member on `Proc1` is contravariant, will ask:
T.type_parameter(:U)  <:  Integer
```

which then places an upper bound.

In that case, `f` was a positional proc argument, so it participates in argument
subtype checking.

But we attempt to infer types for block arguments, so we neither ask whether a
block has a type nor whether that type is a subtype of the `&blk` type
declaration in the sig. Instead, we attempt to use the bounds accumulated so far
to approximate the type before beginning to typecheck the block.

Intuitively, instead of unconditionally using the upper bound to approximate
type variables in the block type, we need to make approximate aware of which
position in the proc type the variable occurred in. If it occurred in the
`.returns` of the proc we can use the upper bound like before, but if it occurs
in the argument type of the proc, we want to approximate the type to its lower
bound instead of defaulting to `<top>`.

### Test plan

See included automated tests.

PR #7897, opened 2024-05-20
Tweak TypeVar::_approximate to consult other bound

jez:
### Motivation

Before the changes in #6076 (dc0319bc98055a4a9f3676b1c48f721a8f8823c1),
the behavior of this function was "look at one bound, if there isn't a
bound for that, then use the other bound, and if there are no bounds,
use `<top>`."

After the changes in #6076, we started to **only** consult one of the
bounds, depending on polarity.

We can improve the heuristic if we attempt to smush those two heuristics
together: always consult both bounds, but choose which bound to consult
**first** based on the polarity of the type variable.

As dmitry wrote:

    // TODO: in many languages this method is a huge adhoc heuristic
    // let's see if we can keep it small

Maybe adding complexity here isn't great, but ¯\\\_(ツ)\_/¯

### Test plan

The test included here motivates the user who was trying to type their code. In this case, they wanted to be able to write a function which, as long as you only raise exceptions through the `raiser` function that gets yielded to a block, the framework will handle catching those exceptions and wrapping them up into a `Result` type, while other exceptions will not be caught.

### DIFF
--- a/core/types/typemaps.cc
+++ b/core/types/typemaps.cc
@@ -54,6 +54,11 @@ TypePtr TypeVar::_approximate(const GlobalState &gs, const TypeConstraint &tc, c
                 if (bound.isFullyDefined()) {
                     return bound;
                 }
+            } else if (tc.hasLowerBound(sym)) {
+                auto bound = tc.findLowerBound(sym);
+                if (bound.isFullyDefined()) {
+                    return bound;
+                }
             }
             break;
         }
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_Q8vpXTvi4K27Df
→ https://go/builds/bui_Q8vpovj4iBOflH
→ https://go/builds/bui_Q8vpLnwWuzzMbS

neilparikh (approved):

PR #7898, opened 2024-05-20
Introduce a msgpack 'lite' mode that skips extra metadata in parsed files

aadi-stripe:
Introduces a lightweight mode for msgpack serialization in autogen which skips extra reference metadata like scope, nesting, expression ranges, typing information, requires, etc. A few points

* The new mode is behind a flag called `--autogen-msgpack-skip-reference-metadata`. Open to a better name.
* The new logic is different enough from the full msgpack mode that I'm creating a derived class that has the new functionality, called `MsgpackWriterLite`. This duplicates some code, but still feels less clunky than sticking `if` statements all over the place in the original class. May create some maintenance overhead long term, but will be a lot easier and cleaner to read.

### Motivation

Facilitates a faster use-case (code preloading) for the msgpack printer mode at Stripe, where we only want to deserialize basic data like definition and reference names, and ignore everything else.

### Test plan

Tested new functionality on Stripe codebase both locally and in CI.

### DIFF
--- a/main/autogen/data/definitions.cc
+++ b/main/autogen/data/definitions.cc
@@ -171,12 +171,23 @@ vector<string> ParsedFile::listAllClasses(core::Context ctx) {
 
 // Convert this parsedfile to a msgpack representation
 string ParsedFile::toMsgpack(core::Context ctx, int version, const AutogenConfig &autogenCfg) {
-    MsgpackWriter write(version);
-    return write.pack(ctx, *this, autogenCfg);
+    if (autogenCfg.msgpackSkipReferenceMetadata) {
+        MsgpackWriterLite write(version);
+
+        return write.pack(ctx, *this, autogenCfg);
+    } else {
+        MsgpackWriter write(version);
+
+        return write.pack(ctx, *this, autogenCfg);
+    }
 }
 
-string ParsedFile::msgpackGlobalHeader(int version, size_t numFiles) {
-    return MsgpackWriter::msgpackGlobalHeader(version, numFiles);
+string ParsedFile::msgpackGlobalHeader(int version, size_t numFiles, const AutogenConfig &autogenCfg) {
+    if (autogenCfg.msgpackSkipReferenceMetadata) {
+        return MsgpackWriterLite::msgpackGlobalHeader(version, numFiles);
+    } else {
+        return MsgpackWriter::msgpackGlobalHeader(version, numFiles);
+    }
 }
 
 } // namespace sorbet::autogen
--- a/main/autogen/data/BUILD
+++ b/main/autogen/data/BUILD
@@ -3,6 +3,7 @@ cc_library(
     srcs = [
         "definitions.cc",
         "msgpack.cc",
+        "msgpack_lite.cc",
     ],
     hdrs = [
         "definitions.h",
--- a/main/autogen/data/definitions.h
+++ b/main/autogen/data/definitions.h
@@ -151,6 +151,7 @@ struct Reference {
 
 struct AutogenConfig {
     const std::vector<std::string> behaviorAllowedInRBIsPaths;
+    const bool msgpackSkipReferenceMetadata = false;
 };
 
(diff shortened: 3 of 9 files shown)


aadi-stripe:

→ https://go/builds/bui_Q8xH3dyAP541HI
→ https://go/builds/bui_Q8xHbAYnC4w88p
→ https://go/builds/bui_Q8xHWBy0IFfIqN

aadi-stripe:

→ https://go/builds/bui_Q8ydvqzOIm3JOq
→ https://go/builds/bui_Q8ydysvQwy2s2s
→ https://go/builds/bui_Q8yddnTENHurAl

froydnj (changes requested):
I have concerns about the versioning being identical between the two file types; I don't think that is going to end well, particularly because we tend to not use the attributes encoded in the file as the source of truth for things -- we drive the attributes that we need to read by what version we've read.

WDYT about doing something like setting the high bit for the "lite" version?  So we'd have `0x80000006` for lite versions and `0x6` for non-lite versions.  That'd give us a fighting chance of distinguishing between the two in whatever tools we use, and it'd make the tools a little more self-describing about which format they accepted.

My sense is that overall, this is probably OK as a stopgap.  I have mild concerns about the code duplication and what it would be like to make changes for a hypothetical v7, but I do agree that conditional logic gets pretty twisty here.  Templates might be the right way to handle things, but I think we could deal with that refactoring later if we wind up keeping the msgpack DB for the long term.

aadi-stripe:
> > I have concerns about the versioning being identical between the two file types; I don't think that is going to end well, particularly because we tend to not use the attributes encoded in the file as the source of truth for things -- we drive the attributes that we need to read by what version we've read.
> > WDYT about doing something like setting the high bit for the "lite" version? So we'd have `0x80000006` for lite versions and `0x6` for non-lite versions. That'd give us a fighting chance of distinguishing between the two in whatever tools we use, and it'd make the tools a little more self-describing about which format they accepted.
> > My sense is that overall, this is probably OK as a stopgap. I have mild concerns about the code duplication and what it would be like to make changes for a hypothetical v7, but I do agree that conditional logic gets pretty twisty here. Templates might be the right way to handle things, but I think we could deal with that refactoring later if we wind up keeping the msgpack DB for the long term.
> 
> Setting MSB is an interesting idea. Might overload the concept of version though, which I find weird, but eh it's probably fine.

fixed

main/autogen/data/definitions.cc
```diff
@@ -171,12 +171,29 @@ vector<string> ParsedFile::listAllClasses(core::Context ctx) {
...
+
+    if (version_bit_8 == 1) {
+        MsgpackWriterLite write(version_low_7);
+
+        return write.pack(ctx, *this, autogenCfg);
+    } else {
+        MsgpackWriter write(version_low_7);
+
+        return write.pack(ctx, *this, autogenCfg);
+    }
```
froydnj, line 185: Is the intent here that the external invoker of depdb passes a version with the high bit set?  a) that seems like putting the responsibility in the wrong place; b) our `MAX_VERSION` constant would need to be updated for that (maybe you did it this way because you ran into `MAX_VERSION` issues in the first place?); and c) even if this bit-passing is intended to live somewhere under `main/autogen/`, I don't see that being done in this PR.

This all seems more complicated than just having `MsgpackWriterLite::msgpackGlobalHeader` tack on the high bit itself.

aadi-stripe: This works because the assertion for `MAX_VERSION` runs at the point of instantiating the `MsgpackWriter` or `Lite` object, which still gets the version with the high bit stripped out.

We need some way of communicating via the cli, either a flag like I had before or this: I figure since we're already passing version in the cli this is reasonable to do.

aadi-stripe:
@froydnj Fixed as per offline conversation

froydnj (approved):
Thank you!

PR #7900, opened 2024-05-21
website: Add a note about filter_map and friends

jez:
### Motivation

Provide helpful tip for people dealing with flow sensitivity errors.

h/t @marianosimone-stripe

### Test plan

n/a

### DIFF
--- a/website/docs/flow-sensitive.md
+++ b/website/docs/flow-sensitive.md
@@ -276,17 +276,16 @@ that control flow in Sorbet is a syntactic property, not a semantic property
 ## Tips
 
 These are some helpful tips for helping Sorbet track the types of expressions
-throughout a program
+throughout a program.
 
-### Prefer `Array#filter_map`
+### Prefer `xs.compact` to `xs.reject(&:nil?)`
 
-Sorbet does not use boolean predicates like those accepted by `Array#select`,
-`Array#filter`, or `Array#reject` to infer a more narrow type about the element
-of the list.
+Sorbet does not use the boolean predicate accepted by methods like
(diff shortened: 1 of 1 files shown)


website/docs/flow-sensitive.md
```diff
@@ -263,3 +263,39 @@ View full example on sorbet.run</a>
 > **Note**: Many Ruby constructs that look like local variables are actually
 > method calls without parens! Specifically, watch out for `attr_reader` and
 > zero-argument method definitions.
+
+## Tips
```
jez, line 267: I figure that this will not be the last tip we add here. e.g. I know that mariano has some suggestions written up (prefer `.fetch`, etc.) so I've pre-emptively carved this out into a proper section. Also it will be easier to link to http://go/sorbet/flow-sensitive/#tips from memory than to a longer section title.

marianosimone-stripe: This is a great example, but I think we should also add:

```ruby
class Parent; end
class ChildA < Parent;
  def method_in_a; end
end
class ChildB < Parent; end

sig {returns(T::Array[Parent])}
def get_all; []; end

my_list = get_all
all_a = my_list.filter_map { |e| e if e.is_a?(ChildA) }
all_a.each do |a|
  a.method_in_a
end
```

This is something people struggle with, and end up using `T.cast`s or `T.unsafe`s for

jez: I wrote another section, feel free to take a look.

I particularly wanted to avoid needing to define a class hierarchy to make the point.

PR #7912, opened 2024-05-29
Switch to using `--incompatible_enable_cc_toolchain_resolution`

jez:
### Motivation

This upgrades us to a fork of `bazel-contrib/toolchains_llvm` based off of
0.10.3 (the last release of that project which doesn't require Bazel 7). This
release is recent enough that it supports Bazel's new way of declaring C/C++
toolchains (using the `--incompatible_enable_cc_toolchain_resolution` flag).

This change is based on top of two other PRs:

- <https://github.com/sorbet/sorbet/pull/7884>
- <https://github.com/sorbet/sorbet/pull/7911>

There is useful context in those two PRs, so I would recommend skimming their
descriptions as well.

Only the two most recent commits are unique to this branch, though one of those
commits includes a handful of custom changes that were made to our
`bazel-toolchain` fork repo. Those commits can be seen here:

<https://github.com/bazel-contrib/toolchains_llvm/compare/0.10.3...sorbet:bazel-toolchain:8d9165fd3560f6ff50bc4794972f714f4ba2adaa>

Currently those commits reside on the `jez-more-tools` branch, but after this
change lands, I will force-push the `sorbet` branch of `sorbet/bazel-toolchain`
to include only these commits.

The commits themselves are here:

- **Add clangd and llvm-symbolizer as aliased tools** (f16a6b0)

- **Apply changes from sorbet/bazel-toolchain#7** (9db8ae1)

  Nathan's changes were made against a version of bazel-toolchain that
  used the old toolchain configuration mechanism.

  This makes the same changes against upstream.

- **Add another URL** (6589ec9)

  We have custom release builds of LLVM, so that 15.0.7 is available on
  all the platforms we support. This commit adds the first of those
  release tarball's checksums.

- **Add another URL** (a5a404f)

  This commit adds the second of those release tarball's checksums.

- **Teach release_name.bzl about our builds** (9b43d0c)

  One last commit to make the toolchain aware of our custom 15.0.7
  builds. This one teaches it about which ubuntu version these builds
  were built against.

- **New attribute extra_compiler_files (#245)** (1da01e6)

  I cherry-picked this commit in order to make it easier to cherry-pick
  the commit that follows. It should be a benign change that doesn't
  affect us, but makes it easier to cherry-pick the following change
  without needing to resolve tons of merge conflicts. The original
  commit message is as follows:

  ---------

  This helps inject files into the compiler action sandbox. The files can
  then be used in the compiler command line, e.g. the sanitizer ignore
  list.

  ---------

  Co-authored-by: Siddhartha Bagaria <starsid@gmail.com>

- **Use LLD linker for Darwin** (c83804a)

  Originally from this commit:

  <https://github.com/bazel-contrib/toolchains_llvm/pull/286/commits/8179a799053e56f312afda027afe6ac59ad2ab31>

  The PR that this change landed in ended up making a followup commit to
  make the change optional, so that people could opt into it by passing
  `--linkopt` and/or `--host_linkopt` to set something like

      --linkopt=-fuse-ld=ld64.lld

  We're seeing behavior where if we **ever** use the system linker, it
  fails to link (because it attempts to link libunwind at the wrong
  location), so we need to make this linker be used all the time. There is
  no way to opt out of using `ld64.lld` with this commit.

- **Fix "Use LLD linker for Darwin" to work with `absolute_paths = True`** (8d9165f)

  The previous commit has one bug, in that it doesn't work with absolute
  paths. This fix should be enough to fix that, and we should likely
  contribute this upstream, but I haven't gotten around to doing that
  yet.

### Test plan

- macOS
  - arm64
    - [x] `--config=dbg`
    - [x] `--config=buildfarm-sanitized-mac`
    - [x] `--config=release-mac`
    - [x] `--config=release-debug-mac`
    - [x] `--config=webasm`
  - x86_64
    - [ ] `--config=dbg`
    - [x] `--config=buildfarm-sanitized-mac`
      - <https://buildkite.com/sorbet/sorbet/builds/31445>
    - [x] `--config=release-mac`
      - <https://buildkite.com/sorbet/sorbet/builds/31445>
    - [ ] `--config=release-debug-mac`
    - [x] `--config=webasm`
      - :x: Not expected to work, see explanation in #7911
  - [x] x86_64 `--config=release-mac` on arm64 via Rosetta
- Linux (Ubuntu 20.04)
  - arm64
    - [x] `--config=dbg`
    - [x] `--config=buildfarm-sanitized-linux`
    - [x] `--config=release-linux-aarch64`
    - [x] `--config=release-debug-linux-aarch64`
      - Does not exist for arm64
    - [x] `--config=webasm`
      - 🤷‍♂️ Not tested, don't care whether it works or not
  - x86_64
    - [x] `--config=dbg`
    - [x] `--config=buildfarm-sanitized-linux`
    - [x] `--config=release-linux`
    - [x] `--config=release-debug-linux`
    - [x] `--config=webasm`
- sorbet.run end-to-end tests
  - [x] monaco version
  - [x] ace version
- Stripe builds
  - [x] sips-ruby-typer
  - [x] sips-ruby-typer-linux-arm64
  - [x] sips-sorbet-check
  - [x] devbox performance analysis


jez:

→ https://go/builds/bui_QCG76kyz13wmp0
→ https://go/builds/bui_QCG7qDTIIxTeIr
→ https://go/builds/bui_QCG7YmWklI6XUk

froydnj (approved):

PR #7918, opened 2024-05-30
wip: Try using c++20

jez:
### Motivation

I've been wanting to make this upgrade for so long now, but we'd always been
blocked because we couldn't upgrade emscripten. Now that we've upgraded, there
shouldn't really be any blockers left.

### Test plan

- [ ] @jez Test this more thoroughly?


core/Types.h
```diff
@@ -1003,8 +1003,16 @@ struct DispatchArgs {
...
+    DispatchArgs(NameRef name, const CallLocs &locs, uint16_t numPosArgs,
+                 InlinedVector<const TypeAndOrigins *, 2> &args, const TypePtr &selfType, TypeAndOrigins fullType,
+                 const TypePtr &thisType, const std::shared_ptr<const SendAndBlockLink> &block,
+                 Loc originForUninitialized, bool isPrivateOk, bool suppressErrors, NameRef enclosingMethodForSuper)
+        : name(name), locs(locs), numPosArgs(numPosArgs), args(args), selfType(selfType), fullType(std::move(fullType)),
+          thisType(thisType), block(block), originForUninitialized(std::move(originForUninitialized)),
+          isPrivateOk(isPrivateOk), suppressErrors(suppressErrors), enclosingMethodForSuper(enclosingMethodForSuper) {}
     DispatchArgs(const DispatchArgs &) = delete;
     DispatchArgs &operator=(const DispatchArgs &) = delete;
+    DispatchArgs(DispatchArgs &&) = default;
```
elliottt, line 1015: Try deleting all of these constructors. It seems like even the deleted copy constructor constitutes a user-defined constructor, from the point of view of considering a struct an aggregate.

jez-stripe: Yeah that made the error go away, but I would like to be able to ensure that people don't copy this, if possible.

elliottt: Cool, that at least explains the error :D

It should definitely be possible to delete the copy constructor, but I think that will require the constructor that you've added in this PR. Happy to continue debugging the crash this week if you've got some free time!

froydnj: I know why the crashes happen, will explain in Slack or over Zoom tomorrow.

PR #7922, opened 2024-05-30
Use Bazel installer for Darwin ARM64

Morriar:
### Motivation

With the latest version of [sorbet/bazel-toolchain](https://github.com/sorbet/bazel-toolchain) we can now use the Bazel installer for Darwin ARM64.

Running `./bazel build //main:sorbet` will produce an ARM64 binary 🎉 

```
$ ./bazel build //main:sorbet
$ file bazel-bin/main/sorbet
bazel-bin/main/sorbet: Mach-O 64-bit executable arm64
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -136,6 +136,12 @@ build:release-linux           --copt=-march=sandybridge
 build:release-sanitized-linux --copt=-march=sandybridge
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
+build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
+build:release-mac-arm64 --define release=true
+build:release-mac-arm64 --compilation_mode=opt
+build:release-mac-arm64 --config=backtracesymbols
+build:release-mac-arm64 --config=shared-libs
+build:release-mac-arm64 --config=versioned
 
 build:release-debug-linux --config=release-linux
 build:release-debug-linux --config=release-debug-common
--- a/.buildkite/build-static-release.sh
+++ b/.buildkite/build-static-release.sh
@@ -16,8 +16,12 @@ case "$platform" in
   linux-aarch64)
     CONFIG_OPTS="--config=release-${platform}"
     ;;
-  darwin-x86_64|darwin-arm64)
+  darwin-x86_64)
     CONFIG_OPTS="--config=release-mac"
+    command -v autoconf >/dev/null 2>&1 || brew install autoconf
+    ;;
+  darwin-arm64)
+    CONFIG_OPTS="--config=release-mac-arm64"
     command -v autoconf >/dev/null 2>&1 || brew install autoconf
     ;;
   *)
--- a/.buildkite/build-static-release.sh
+++ b/.buildkite/build-static-release.sh
@@ -33,6 +33,8 @@ case "$platform" in
     TARGET_PLATFORM=darwin-x86_64 ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
     mv bazel-bin/main/sorbet sorbet_x86_64
 
+    ./bazel clean --expunge
+
     TARGET_PLATFORM=darwin-arm64 ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
     mv bazel-bin/main/sorbet sorbet_arm64
 
--- a/.buildkite/build-static-release.sh
+++ b/.buildkite/build-static-release.sh
@@ -28,7 +28,21 @@ esac
 
 echo will run with $CONFIG_OPTS
 
-./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+case "$platform" in
+  darwin-x86_64|darwin-arm64)
+    TARGET_PLATFORM=darwin-x86_64 ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    mv bazel-bin/main/sorbet sorbet_x86_64
+
+    TARGET_PLATFORM=darwin-arm64 ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    mv bazel-bin/main/sorbet sorbet_arm64
+
+    lipo -create -output bazel-bin/main/sorbet sorbet_x86_64 sorbet_arm64
+    rm sorbet_x86_64 sorbet_arm64
+    ;;
+  *)
+    ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    ;;
+esac
 
 mkdir gems/sorbet-static/libexec/
 cp bazel-bin/main/sorbet gems/sorbet-static/libexec/
--- a/bazel
+++ b/bazel
@@ -166,4 +166,5 @@ mkdir -p "$BUILD_DIR"
 )
 rm -rf "$BUILD_DIR"
 
+"$bazel_exec_path" clean --expunge
 exec "$bazel_exec_path" "$@"
--- a/bazel
+++ b/bazel
@@ -34,18 +34,59 @@ export BAZEL_VERSION
 bazel_bin_loc="${bazel_bin_loc:-$HOME/.bazel_binaries}"
 bazel_exec_path="$bazel_bin_loc/$BAZEL_VERSION/bin/bazel-real"
 
+kernel_name="$(uname -s | tr 'A-Z' 'a-z')"
+processor_name="$(uname -m)"
+bazel_installer_platform="${kernel_name}-${processor_name}"
+
+if [ -f "$bazel_exec_path" ]; then
+  # If we already have a cached bazel binary, check if it's the right platform
+  bazel_type=$(file -b "$bazel_exec_path")
+
+  case "$bazel_type" in
+    *ELF*)
+      bazel_kernel="linux"
+      ;;
+    *Mach-O*)
+      bazel_kernel="darwin"
+      ;;
+    *)
+      bazel_kernel="unknown"
+      ;;
+  esac
+
+  case "$bazel_type" in
+    *x86-64* | *x86_64*)
+      bazel_proc="x86_64"
+      ;;
+    *arm64*)
+      bazel_proc="arm64"
+      ;;
+    *aarch64*)
+      bazel_proc="aarch64"
+      ;;
+    *)
+      bazel_proc="unknown"
+      ;;
+  esac
+
+  bazel_platform="${bazel_kernel}-${bazel_proc}"
+  echo >&2 "Found cached Bazel binary for ${bazel_platform}..."
+
+  if [[ $bazel_platform != $bazel_installer_platform ]]; then
+    echo >&2 "Bazel binary platform mismatch: expected $bazel_installer_platform, got $bazel_platform"
+    echo >&2 "Removing cached binary..."
+    rm -f "$bazel_exec_path"
+  fi
+fi
+
 if [ -f "$bazel_exec_path" ]; then
   exec "$bazel_exec_path" "$@"
 fi
 
 # ----- slow path -----
 
-echo >&2 "No cached Bazel v$BAZEL_VERSION found, installing..."
+echo >&2 "No cached Bazel v$BAZEL_VERSION found, installing for $bazel_installer_platform..."
 
-kernel_name="$(uname -s | tr 'A-Z' 'a-z')"
-processor_name="$(uname -m)"
-
-bazel_installer_platform="${kernel_name}-${processor_name}"
 case "$bazel_installer_platform" in
   linux-x86_64) ;;
   linux-aarch64)
--- a/bazel
+++ b/bazel
@@ -34,9 +34,13 @@ export BAZEL_VERSION
 bazel_bin_loc="${bazel_bin_loc:-$HOME/.bazel_binaries}"
 bazel_exec_path="$bazel_bin_loc/$BAZEL_VERSION/bin/bazel-real"
 
-kernel_name="$(uname -s | tr 'A-Z' 'a-z')"
-processor_name="$(uname -m)"
-bazel_installer_platform="${kernel_name}-${processor_name}"
+if [ "${TARGET_PLATFORM:-}" != "" ]; then
+  bazel_installer_platform="${TARGET_PLATFORM}"
+else
+  kernel_name="$(uname -s | tr 'A-Z' 'a-z')"
+  processor_name="$(uname -m)"
+  bazel_installer_platform="${kernel_name}-${processor_name}"
+fi
 
 if [ -f "$bazel_exec_path" ]; then
   # If we already have a cached bazel binary, check if it's the right platform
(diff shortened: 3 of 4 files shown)


jez:
A couple thoughts:

- This produces a ton of warnings when building on macOS. I would like to figure out a way to get rid of those warnings before proceeding with this. I had a TODO in #7916 to look into how to make the `--macos_minimum_version=13.0.0` change apply everywhere. (For example, I don't know if it's possible to pass that unconditionally, and if it's passed on Linux it's just ignored.)
- This change doesn't bust the bazel binary cache, because it doesn't also change the bazel version. On #7916, I made similar changes to start using the arm64 bazel installer, but changed the bazel version in the same commit. This means that, assuming you've never installed 7.1.2 from some other means, you'll always have the architecture-appropriate version of the bazel binary.

I can look into getting the CI component of this set up so that we can publish arm64 binaries. One thing I would appreciate you looking into is whether it's possible to build x86_64 binaries from arm64, because while we could technically have two laptops running in my closet in CI I think it would be convenient if there were only one.

jez (commented):
^

Morriar:
> I would appreciate you looking into is whether it's possible to build x86_64 binaries from arm64

I looked into it and to do cross-compilation arm64 -> x86_64 and the current approach doesn't seem to allow it.

In Sorbet's `WORKSPACE` we require the `llvm_toolchain` which will use the host os/arch to download the correct LLVM distrib. When on a darwin/arm64 platform, the distrib downloaded is `clang+llvm-15.0.7-arm64-apple-darwin22.0.tar.xz`.

We can tweak the toolchain to enable cross-compilation but linking won't work since we try to link against the libc++ coming from the downloaded distrib resulting in errors such as `llvm_toolchain_15_0_7_llvm/lib/libc++abi.a has architecture arm64 which is incompatible with target architecture x86_64`.

I also tried to use the `--cpu` or `--platform` to select which distrib to download, but I couldn't find a way to pass this information to `llvm_toolchain()` from the `WORKSPACE`. Looking at how `WORKSPACE` works, it seems way too early in the execution to access any kind of platform configuration. It also seems impossible to apply a `select` selector on the toolchain configuration at this stage.

One solution could be to always download both arm64 and x86_64 LLVM distribs and register both toolchains so we support toolchain selection based on the platform.

I did, however, found a simple (yet maybe a bit dirty) workaround: the downloaded version of the LLVM distrib depends on the version of the Bazel installer:
* using the Bazel `darwin-x86_64` installer will result in LLVM `clang+llvm-15.0.7-x86_64-apple-darwin21.0.tar.xz` being installed and thus produce binaries for x86_64
* using the Bazel `darwin-arm64` installer will result in LLVM `clang+llvm-15.0.7-arm64-apple-darwin22.0.tar.xz` being installed and this produce binaries for arm64

So the simple solution would be to pass a flag to `./bazel`:
* `TARGET_ARCH=darwin-x86_64 ./bazel_build ...` to produce the x86_64 binary
* `TARGET_ARCH=darwin-arm64 ./bazel_build ...` to produce the arm64 binary

This combined with your suggestion of busting the bazel binary cache will work.

jez:
I tried building it locally and it fails for me:

```
❯ bazel cquery //main:sorbet
No cached Bazel v6.5.0 found, installing for darwin-arm64...
/var/folders/gt/3hq5tw5d62bfpr903pnnsdh80000gn/T/tmp.gL47mPwzKg
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 44.6M  100 44.6M    0     0  29.6M      0  0:00:01  0:00:01 --:--:-- 39.5M
Bazel installer
---------------

Bazel is bundled with software licensed under the GPLv2 with Classpath exception.
You can find the sources next to the installer on our release page:
   https://github.com/bazelbuild/bazel/releases

# Binary package at HEAD (@2aa603bd77e)
   - [Commit](https://github.com/bazelbuild/bazel/commit/2aa603bd77e)
Uncompressing......WARNING: Invoking Bazel in batch mode since it is not invoked from within a workspace (below a directory having a WORKSPACE file).
Picked up JAVA_TOOL_OPTIONS: -Dlog4j2.formatMsgNoLookups=true -Dcom.sun.jndi.ldap.object.trustURLCodebase=false
.

Bazel is now installed!

Make sure you have "/Users/jez/.bazel_binaries/6.5.0/bin_t" in your path.

For bash completion, add the following line to your :
  source /Users/jez/.bazel_binaries/6.5.0/bin/bazel-complete.bash

For fish shell completion, link this file into your
/Users/jez/.config/fish/completions/ directory:
  ln -s /Users/jez/.bazel_binaries/6.5.0/bin/bazel.fish /Users/jez/.config/fish/completions/bazel.fish

See http://bazel.build/docs/getting-started.html to start a new project!
INFO: Invocation ID: fc903095-50de-42cc-84a9-aef38a4e05f0
ERROR: /private/var/tmp/_bazel_jez/491a13d243378e23690a7f8390a5c09f/external/local_config_cc/BUILD:48:19: in cc_toolchain_suite rule @local_config_cc//:toolchain: cc_toolchain_suite '@local_config_cc//:toolchain' does not contain a toolchain for cpu 'darwin_arm64'
ERROR: /private/var/tmp/_bazel_jez/491a13d243378e23690a7f8390a5c09f/external/local_config_cc/BUILD:48:19: Analysis of target '@local_config_cc//:toolchain' failed
ERROR: Analysis of target '//main:sorbet' failed; build aborted:
INFO: Elapsed time: 0.071s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```

I'm not sure what's up. I have a little bit of time to look into this tomorrow, but if you have suggestions I'm all ears.

jez:
I built again, and it seems that problem went away? I'm wondering whether it has something to do with fresh cache or not. Going to have to try to blow the cache and try again.

- - - - -

seemingly unrelated, another problem, this time with release-mac-arm64

```
❯ bazel build --config=release-mac-arm64 //main:sorbet
Found cached Bazel binary for darwin-arm64...
INFO: Invocation ID: 5362ba3c-c55b-4ce3-a164-731264701c00
WARNING: Download from https://github.com/sorbet/llvm-project/releases/download/llvmorg-15.0.7/clang+llvm-15.0.7-arm64-apple-darwin22.0.tar.xz failed: class java.io.FileNotFoundException GET returned 404 Not Found
INFO: Build options --compilation_mode, --copt, --define, and 3 more have changed, discarding analysis cache.
INFO: Analyzed target //main:sorbet (1 packages loaded, 3771 targets configured).
INFO: Found 1 target...
Target //main:sorbet up-to-date:
  bazel-bin/main/sorbet
INFO: Elapsed time: 104.886s, Critical Path: 30.99s
INFO: 936 processes: 264 internal, 672 local.
INFO: Build completed successfully, 936 total actions

~/sorbet (jez-timers)  1m 45s
❯ time bazel-bin/main/sorbet
[3]    49150 killed     bazel-bin/main/sorbet
bazel-bin/main/sorbet  0.00s user 0.00s system 17% cpu 0.004 total

~/sorbet (jez-timers)
❯ time bazel-bin/main/sorbet -e 0
[3]    49180 killed     bazel-bin/main/sorbet -e 0
bazel-bin/main/sorbet -e 0  0.00s user 0.00s system 29% cpu 0.003 total

~/sorbet (jez-timers)
❯ file bazel-bin/main/sorbet
bazel-bin/main/sorbet: Mach-O 64-bit executable arm64

~/sorbet (jez-timers)
❯ lldb -- bazel-bin/main/sorbet
error: There is a .lldbinit file in the current directory which is not being read.
To silence this warning without sourcing in the local .lldbinit,
add the following to the lldbinit file in your home directory:
    settings set target.load-cwd-lldbinit false
To allow lldb to source .lldbinit files in the current working directory,
set the value of this variable to true.  Only do so if you understand and
accept the security risk.
(lldb) target create "bazel-bin/main/sorbet"
Current executable set to '/Users/jez/sorbet/bazel-bin/main/sorbet' (arm64).
(lldb) r
error: Bad executable (or shared library)
```

jez:
I just went and checked out a non-`master` commit, rebuilt, and it's producing arm64 binaries 🤔

jez (commented):
seeing some problems here

Morriar:
I added a small change to force clean the Bazel cache when the version changed, this will take care of errors such as `does not contain a toolchain for cpu 'darwin_arm64'`.

For the binary being killed I can't reproduce but I wonder if this is due to a [code signing crash](https://developer.apple.com/documentation/security/updating-mac-software). Without more details it's hard to say.

@jez could you ensure that the inode for the file changed using `ls -i bazel-bin/main/sorbet` please? Is there a crash report in the [Console](https://support.apple.com/en-ca/guide/console/cnsl664be99a/mac) related to the binary being killed?

jez:
Re-posting this here from Slack for posterity:

when I run

```bash
```

and then

```
❯ file bazel-bin/main/sorbet
bazel-bin/main/sorbet: Mach-O 64-bit executable arm64
```

```
❯ bazel-bin/main/sorbet -e 0
[1]    50165 killed     bazel-bin/main/sorbet
```

```
❯ lldb -- bazel-bin/main/sorbet
(lldb) target create "bazel-bin/main/sorbet"
Current executable set to '/Users/jez/sorbet/bazel-bin/main/sorbet' (arm64).
(lldb) r
error: Bad executable (or shared library)
```

In Console.app, I see these messages

```
default	14:44:33.227127-0700	kernel	ASP: Security policy would not allow process: 51602, /Users/jez/Desktop/sorbet
default	14:44:33.232112-0700	syspolicyd	GK performScan: PST: (path: {mask.hash}@), (team: /Users/jez/Desktop/sorbet), (id: (null)), (bundle_id: (null))
default	14:44:33.516963-0700	syspolicyd	GK evaluateScanResult: 2, PST: (path: {mask.hash}@), (team: /Users/jez/Desktop/sorbet), (id: (null)), (bundle_id: sorbet-555549444c4c44ae55553144a1ef3692cca2d085), 0, 0, 1, 0, 0, 0, 0
default	14:44:33.517077-0700	syspolicyd	Found provenance data on target: TA(e420ab665a665690, 2), PST: (path: {mask.hash}@), (team: /Users/jez/Desktop/sorbet), (id: (null)), (bundle_id: sorbet-555549444c4c44ae55553144a1ef3692cca2d085)
```

When I search that first log line, I see [this result](https://forums.developer.apple.com/forums/thread/716642) which then links to [this post](https://developer.apple.com/forums/thread/706442) by an Apple engineer suggesting that I run this `syspolicy_check`:

```
❯ syspolicy_check distribution ~/Desktop/sorbet
App has failed one or more pre-distribution checks.
---------------------------------------------------------------
Notary Ticket Missing
    File: /Users/jez/Desktop/sorbet
    Severity: Fatal
    Full Error: A Notarization ticket is not stapled to this application.
    Type: Distribution Error

---------------------------------------------------------------
```

I don't think we should need these binaries to be notarized... when I do this:

```
git checkout master &&
  bazel clean --expunge && rm -rf ~/.cache/sorbet/bazel-cache && mkdir ~/.cache/sorbet/bazel-cache &&
  bazel build --config=release-mac //main:sorbet &&
  file bazel-bin/main/sorbet &&
  bazel-bin/main/sorbet -e 0
```

I see the normal output:

```
No errors! Great job.
```

jez:
I'm guessing that the `syspolicy_check` error is a red-herring, because when I run the same command against the x86_64 binary, I get this output:

```
❯ syspolicy_check distribution bazel-bin/main/sorbet
App has failed one or more pre-distribution checks.
---------------------------------------------------------------
Codesign Error
    File: /Users/jez/stripe/sorbet/bazel-bin/main/sorbet
    Severity: Fatal
    Full Error: File is not signed at all.
    Type: Notary Error

Notary Ticket Missing
    File: /Users/jez/stripe/sorbet/bazel-bin/main/sorbet
    Severity: Fatal
    Full Error: A Notarization ticket is not stapled to this application.
    Type: Distribution Error

---------------------------------------------------------------
```

Maybe that's the problem? Maybe something about generating arm64 binaries means that it tries to sign the binary at all?

Morriar:
Pushed a change to force dynamic linking on arm64, which fixes the Sorbet process being killed, this should be good to go.

vinistock:
> Pushed a change to force dynamic linking on arm64

As a reference, we discovered that the issue was static linking based on this discussion https://stackoverflow.com/questions/71723764/why-does-macos-kill-static-executables-created-by-clang.

Apparently, statically linked binaries are disallowed in ARM64 MacOS. After reproducing the original problem in two different machines, we verified that removing the static linking fixed the problem on both and allowed Sorbet to run normally.

.buildkite/build-static-release.sh
```diff
@@ -28,7 +32,23 @@ esac
...
+    TARGET_PLATFORM=darwin-x86_64 ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    mv bazel-bin/main/sorbet sorbet_x86_64
+
+    ./bazel clean --expunge
+
+    TARGET_PLATFORM=darwin-arm64 ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    mv bazel-bin/main/sorbet sorbet_arm64
+
+    lipo -create -output bazel-bin/main/sorbet sorbet_x86_64 sorbet_arm64
+    rm sorbet_x86_64 sorbet_arm64
```
froydnj, line 46: `bazel-bin/` is Bazel's domain, we're not permitted to move files into or out of it.  I realize that here we're not executing Bazel again, so it shouldn't matter, but I think we should respect Bazel's invariants here.

So the darwin branch here should `cp` files out of the sandbox and use some made-up-file for `lipo -create -output`, and then communicate that to the bit that copies into `gems/sorbet-static/libexec`, below.  (Some changes may be required for the Linux branch below as well.)

.bazelrc
```diff
@@ -136,6 +136,12 @@ build:release-linux           --copt=-march=sandybridge
 build:release-sanitized-linux --copt=-march=sandybridge
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
+build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
+build:release-mac-arm64 --define release=true
+build:release-mac-arm64 --compilation_mode=opt
+build:release-mac-arm64 --config=backtracesymbols
+build:release-mac-arm64 --config=shared-libs
+build:release-mac-arm64 --config=versioned
```
froydnj, line 144: Ugh, ok, we can't use `--config=release-common` here because of the `shared-libs` vs. `static-libs` problem.

froydnj (changes requested):
One comment about `build-static-release.sh`, and I'm not totally sure that explicitly setting `TARGET_PLATFORM` will do the right thing on the x86-64 mac buildkite machine that we have.

.bazelrc
vinistock: Yes, indeed. We ended up going with making the ARM64 build the exception, rather than changing `release-common` and having to spread the `static-libs` config everywhere else.

Morriar:
Superseded by https://github.com/sorbet/sorbet/pull/8152.

PR #7923, opened 2024-05-30
Don't hang if publishing fails

jez:
### Motivation

When `grep` isn't given a file to grep in, hangs without printing any
warnings. I think this is a pretty absurd default, and I even have a
blog post ranting about it:

→ <https://blog.jez.io/cli-tty/>

But this should fix the problem and make our publish.sh step not hang
indefinitely if the publish fails.

You can see an example of this hang happening in this build:

<https://buildkite.com/sorbet/sorbet/builds/31485#018fcad9-68c3-4c41-83b0-f3558eff2348/420-436>

Note the timestamps, it took ~2.5 hours before I realized that we had a
hanging build.

### Test plan

This will get tested on master, unless I'm able to go through our runbook to
update the personal access token faster than CI can reach the publish step.

### DIFF
--- a/.buildkite/publish.sh
+++ b/.buildkite/publish.sh
@@ -145,7 +145,7 @@ elif [ "$dryrun" = "" ]; then
     # It can take 4+ minutes for the marketplace to "verify" our extension
     # before it shows up as published according to `vsce show`.
     cat "$vsce_publish_output"
-    if grep -qF 'Version number cannot be the same'; then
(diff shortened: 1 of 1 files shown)


froydnj (approved):

jez:
Not only did I update the token before master builds, I updated the token before the branch built.

So this change will effectively be untested, but ¯\_(ツ)_/¯ I'm fine with that.

Issue #7932, opened 2024-06-02
Add compiler flag to specify version of ARM architecture


maruthgoyal:
### Problem
[ARM v8.1a introduced new atomic instructions](https://learn.arm.com/learning-paths/servers-and-cloud-computing/lse/intro/#:~:text=The%20Armv8.,Large%20System%20Extensions%20(LSE).) which have shown better performance and scaling characteristics on [some benchmarks](https://cpufun.substack.com/p/atomics-in-aarch64). Based on the recommendations in the latter we should likely set `-march=armv8.1a` or `-moutline-atomics` (which will pick the instructions to emit based on detected architecture IIUC)

jez:
Pretty sure this was implemented in #7933 but if that's not the case feel free to re-open.

PR #7933, opened 2024-06-03
Add `-march=armv8.1a` for linux-aarch64 builds

maruth-stripe:
### Motivation

https://github.com/sorbet/sorbet/issues/7932

### Test plan

Built locally on a Graviton machine. Disassembly of binary had `ldadd*` family of instructions, introduced in v8.1a. typecheck succeeded.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -134,6 +134,7 @@ build:release-linux-aarch64 --define mimalloc=true
 # however some AWS instances in our fleet still run Sandy Bridge (Skylake predecessor), as of 2018.
 build:release-linux           --copt=-march=sandybridge
 build:release-sanitized-linux --copt=-march=sandybridge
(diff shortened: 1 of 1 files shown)


maruth-stripe:
All the aarch64 machines we use are all at least ARM v8.2a. Ideally if we had aarch64 builds for macs we could enable this for that as well, but alas

jez:
Is this going to break any open source users attempting to run Sorbet on older arm hardware?

jez (changes requested):
^

maruthgoyal:
This will work for Graviton2+ and Apple M1+. I'm not aware of any other major ARM chips used on laptop/server hardware. Graviton 1 is not offered by AWS anymore far as I can tell.

jez (approved):

PR #7934, opened 2024-06-03
Skip arm builds

jez:
### Motivation

Something is wrong with our linux-arm64 Buildkite setup.

While I figure it out, I don't want to block other people from being able to
make changes to Sorbet.

### Test plan

Existing CI


ilyailya (approved):

jez:
I believe I've fixed this.

PR #7935, opened 2024-06-04
Treat EmptyTree as `nil` in `If`

### LINKED ISSUE #7936: EmptyTree is not always treated as `nil`
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aaaaa%20%3D%200%0Aaaaa%20%3D%0A%20%20if%20rand%282%29%20%3D%3D%200%0A%20%20%20%20''%0A%20%20end%0A%0AT.reveal_type%28aaaa%29%20%23%20error%3A%20T.nilable%28String%29%0A)

```ruby
# typed: true

aaaa = 0
aaaa =
  if rand(2) == 0
    ''
  end

T.reveal_type(aaaa) # error: T.nilable(String)

```

#### Observed output

```
editor.rb:9: Revealed type: T.any(Integer, String) https://srb.help/7014
     9 |T.reveal_type(aaaa) # error: T.nilable(String)
        ^^^^^^^^^^^^^^^^^^^
  Got T.any(Integer, String) originating from:
    editor.rb:3:
     3 |aaaa = 0
               ^
    editor.rb:6:
     6 |    ''
            ^^
Errors: 1
```

#### Expected behavior

When this code runs, there are two problems:

- The revealed value is never an `Integer` at runtime. Technically it's not _wrong_ to say include `Integer` in the type, but Sorbet could be more precise.
- The revealed value is possibly `nil` but Sorbet doesn't capture that in the type. This **is** wrong.

- - -

I attempted a solution in #7935 but there are problems with it (see the failing tests).

Note that the bug only manifests when `aaaa` is unconditionally initialized before the `if` line runs. If `aaaa` is possibly uninitialized, Sorbet correctly infers the type.

jez:
### Motivation

Fixes #7936

Normally this doesn't matter if the `if` or `case` expression has an empty
`else`, so long as the result of the `if`/`case` doesn't assign to a variable.

Even if it does assign to a variable, sometimes it still doesn't matter, so long
as the variable is either already possibly `nil` or untyped.

### Test plan

See included automated tests.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -1034,7 +1034,14 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                 ret = current;
             },
 
-            [&](const ast::EmptyTree &n) { ret = current; },
+            [&](const ast::EmptyTree &n) {
+                // TODO(jez) We might want to ENFORCE(false) here, and make all attempts to walk an
+                // empty tree be handled by the parent node where the `EmptyTree` is a child, so
+                // that there's more context to be able to handle the `EmptyTree` in context. For
+                // example, how the `ast::If` case handles `EmptyTree` so that the loc of the
+                // `EmptyTree` can be set to the loc of the whole `if` expression.
+                ret = current;
+            },
 
             [&](const ast::ClassDef &c) { Exception::raise("Should have been removed by FlattenWalk"); },
             [&](const ast::MethodDef &c) { Exception::raise("Should have been removed by FlattenWalk"); },
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -242,6 +242,13 @@ BasicBlock *CFGBuilder::walkHash(CFGContext cctx, ast::Hash &h, BasicBlock *curr
     return current;
 }
 
+// This doesn't actually "walk" an empty tree, because there's nothing interesting to walk in one.
+// Instead, if conforms to mostly the same interface that `walk` (i.e., returns a BasicBlock *);
+BasicBlock *CFGBuilder::walkEmptyTreeInIf(CFGContext cctx, core::LocOffsets nilLoc, BasicBlock *current) {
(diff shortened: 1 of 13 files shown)


jez:

→ https://go/builds/bui_QEGv8QNuG0QShk
→ https://go/builds/bui_QEGvUTG0bo6egv
→ https://go/builds/bui_QEGvqffwi5BdDD

jez:
This has a ton of problems—going to have to go back to the drawing board.

jez:

→ https://go/builds/bui_RAFLc4xgQBLVWj
→ https://go/builds/bui_RAFLoAT8lzn17h
→ https://go/builds/bui_RAFL46i50U8TZf

jez:

→ https://go/builds/bui_RAFg9QvzaxotTk
→ https://go/builds/bui_RAFgDt4iM5R2Re
→ https://go/builds/bui_RAFgaUwqkw0SjH

cfg/builder/builder_walk.cc
```diff
@@ -385,8 +392,10 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                 auto elseBlock = cctx.inWhat.freshBlock(cctx.loops);
                 conditionalJump(cont, ifSym, thenBlock, elseBlock, cctx.inWhat, a.cond.loc());
 
-                auto thenEnd = walk(cctx, a.thenp, thenBlock);
-                auto elseEnd = walk(cctx, a.elsep, elseBlock);
+                auto thenEnd = ast::isa_tree<ast::EmptyTree>(a.thenp) ? walkEmptyTreeInIf(cctx, a.loc, thenBlock)
+                                                                      : walk(cctx, a.thenp, thenBlock);
```
elliottt, line 396: I wonder if it's worth pushing this into `walkEmptyTreeInIf`, and calling it something like `walkConditionalBranch` instead?

elliottt (approved):
This looks great!

cfg/builder/builder_walk.cc
jez: Yeah possibly, I'm a little hesitant to abstract prematurely, I think that we can possibly do something about this whenever we get around to ENFORCEing that we don't `walk` to `EmptyTree`.

PR #7937, opened 2024-06-04
Treat `type_alias` as a type alias only if called on `T`

### LINKED ISSUE #6413: Sorbet ignores receiver when deciding whether something is a type alias
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0AMyType%20%3D%20DefinitelyNotT.type_alias%20%7BT.any%28Integer%2C%20String%29%7D%0A%0Asig%20%7Bparams%28x%3A%20MyType%29.void%7D%0Adef%20example%28x%29%0A%20%20T.reveal_type%28x%29%0Aend)

```ruby
# typed: true
extend T::Sig

MyType = DefinitelyNotT.type_alias {T.any(Integer, String)}

sig {params(x: MyType).void}
def example(x)
  T.reveal_type(x)
end
```

#### Observed output

```
editor.rb:4: Unable to resolve constant DefinitelyNotT https://srb.help/5002
     4 |MyType = DefinitelyNotT.type_alias {T.any(Integer, String)}
                 ^^^^^^^^^^^^^^
  Did you mean Bundler::Definition? Use `-a` to autocorrect
    editor.rb:4: Replace with Bundler::Definition
     4 |MyType = DefinitelyNotT.type_alias {T.any(Integer, String)}
                 ^^^^^^^^^^^^^^
    https://github.com/sorbet/sorbet/tree/master/rbi/stdlib/bundler.rbi#L867: Bundler::Definition defined here
     867 |class Bundler::Definition
          ^^^^^^^^^^^^^^^^^^^^^^^^^

editor.rb:8: Revealed type: T.any(Integer, String) https://srb.help/7014
     8 |  T.reveal_type(x)
          ^^^^^^^^^^^^^^^^
  Got T.any(Integer, String) originating from:
    editor.rb:7:
     7 |def example(x)
                    ^
Errors: 2
```

#### Expected behavior

Sorbet should treat this like a normal constant assignment—it should only have special behavior for `T.type_alias`, not anything else.

- - -

This is going to be a little tricky. We need to make this decision in namer, before constant resolution has run, which means before the receiver has been resolved. We won't be able to catch cases like this:

```ruby
class MyClass
  module T
    def self.type_alias(&blk); end
  end

  MyType = T.type_alias {T.any(Integer, String)}
end
```

but at least we'll get most of the obviously wrong cases.

amomchilov:
### Motivation

Closes #6413

### Test plan

See included automated tests, which test this both in:

1. In the symbol table (looking for `static-field-type-alias` instead of `static-field`)
2. In the resolver (looking for an error when attempting to use the constant as a type).

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -629,12 +629,23 @@ public:
         return foundDefs->addTypeMember(move(found));
     }
 
+    bool sendRecvIsT(const ast::Send &s) {
+        bool result = false;
+
+        typecase(
+            s.recv, [&](const ast::UnresolvedConstantLit &c) { result = c.cnst == core::Names::Constants::T(); },
+            [&](const ast::ConstantLit &c) { result = c.symbol == core::Symbols::T(); },
+            [&](const ast::ExpressionPtr &_default) { result = false; });
+
+        return result;
+    }
+
     core::FoundDefinitionRef handleAssignment(core::Context ctx, const ast::Assign &asgn) {
         auto &send = ast::cast_tree_nonnull<ast::Send>(asgn.rhs);
         auto foundRef = fillAssign(ctx, asgn);
         ENFORCE(foundRef.kind() == core::FoundDefinitionRef::Kind::StaticField);
         auto &staticField = foundRef.staticField(*foundDefs);
-        staticField.isTypeAlias = send.fun == core::Names::typeAlias();
+        staticField.isTypeAlias = sendRecvIsT(send) && send.fun == core::Names::typeAlias();
         return foundRef;
     }
 
--- a/test/testdata/resolver/type_alias_not_on_T.rb
+++ b/test/testdata/resolver/type_alias_not_on_T.rb
@@ -1,0 +1,15 @@
+# typed: true
+
+module A
+  extend T::Sig
+
+  module NotT
+    def self.type_alias(&blk); end
+  end
+
+  NotATypeAlias = NotT.type_alias {Object}
+
+  sig {returns(NotATypeAlias)}
+  #            ^^^^^^^^^^^^^ error: Constant `A::NotATypeAlias` is not a class or type alias
+  def this_has_an_invalid_return_type; end
+end
--- a/test/testdata/namer/type_alias_not_on_T.rb
+++ b/test/testdata/namer/type_alias_not_on_T.rb
@@ -1,0 +1,20 @@
+# typed: true
+module A
+  module NotT
+    def self.type_alias(&blk); end
+  end
+
+  ValidTypeAlias = T.type_alias {Object}
+
+  NotActuallyATypeAlias = NotT.type_alias {Object}
+end
+
+module KnownLimitation
+  module T
+    def self.type_alias(&blk); end
+  end
+
+  # Known limitation: we can't distinguish between `::KnownLimitation::T` and `::T` in the namer,
+  # so this will have a false positive, being treated as a type alias when it shouldn't be.
+  StillATypeAlias = T.type_alias {Object}
+end
--- a/test/testdata/namer/type_alias_not_on_T.rb.symbol-table.exp
+++ b/test/testdata/namer/type_alias_not_on_T.rb.symbol-table.exp
@@ -1,0 +1,28 @@
+class ::<root> < ::Object ()
+  class ::<Class:<root>>[<AttachedClass>] < ::<Class:Object> ()
+    method ::<Class:<root>>#<static-init> (<blk>) @ test/testdata/namer/type_alias_not_on_T.rb:2
+      argument <blk><block> @ Loc {file=test/testdata/namer/type_alias_not_on_T.rb start=??? end=???}
+  module ::A < ::Sorbet::Private::Static::ImplicitModuleSuperclass () @ test/testdata/namer/type_alias_not_on_T.rb:2
+    static-field ::A::NotActuallyATypeAlias -> <todo sym> @ test/testdata/namer/type_alias_not_on_T.rb:9
+    module ::A::NotT < ::Sorbet::Private::Static::ImplicitModuleSuperclass () @ test/testdata/namer/type_alias_not_on_T.rb:3
+    class ::A::<Class:NotT> < ::Module () @ test/testdata/namer/type_alias_not_on_T.rb:3
+      method ::A::<Class:NotT>#<static-init> (<blk>) @ test/testdata/namer/type_alias_not_on_T.rb:3
+        argument <blk><block> @ Loc {file=test/testdata/namer/type_alias_not_on_T.rb start=??? end=???}
+      method ::A::<Class:NotT>#type_alias (blk) @ test/testdata/namer/type_alias_not_on_T.rb:4
+        argument blk<block> @ Loc {file=test/testdata/namer/type_alias_not_on_T.rb start=4:26 end=4:29}
+    static-field-type-alias ::A::ValidTypeAlias -> Object @ test/testdata/namer/type_alias_not_on_T.rb:7
+  class ::<Class:A> < ::Module () @ test/testdata/namer/type_alias_not_on_T.rb:2
+    method ::<Class:A>#<static-init> (<blk>) @ test/testdata/namer/type_alias_not_on_T.rb:2
+      argument <blk><block> @ Loc {file=test/testdata/namer/type_alias_not_on_T.rb start=??? end=???}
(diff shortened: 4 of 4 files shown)


namer/namer.cc
```diff
@@ -629,12 +629,17 @@ class SymbolFinder {
         return foundDefs->addTypeMember(move(found));
     }
 
+    bool isT(const ast::ExpressionPtr &expr) {
```
amomchilov, line 632: We just put this here initially, directly above the `handleAssignment` method that uses it.

Where should a helper like this belong?

test/testdata/resolver/type_alias_not_on_T.rb
```diff
@@ -0,0 +1,15 @@
...
+  extend T::Sig
+
+  module NotT
+    def self.type_alias(&blk); end
+  end
+
+  NotATypeAlias = NotT.type_alias {Object}
+
+  sig {returns(NotATypeAlias)}
+  #            ^^^^^^^^^^^^^ error: Constant `A::NotATypeAlias` is not a class or type alias
```
amomchilov, line 13: Side note: This is kind of a strange error message.

The message suggests it should be a "class or type alias". But a `module` would also be valid...

```suggestion
  #            ^^^^^^^^^^^^^ error: Constant `A::NotATypeAlias` is not a class, module or type alias
```

It could be a few other things like a type union, type intersection, ..., so perhaps it would be best to just:

```suggestion
  #            ^^^^^^^^^^^^^ error: Constant `A::NotATypeAlias` is not a type
```

jez: Good point. It's subtle because the "class" in "class or type alias" does not refer to `class` the keyword but rather to "class aliases" which are what Sorbet calls constants defined like `ClassAlias = Integer` (as opposed to other constants, like `X = 1`). Class aliases can alias to modules as well.

Ruby natively doesn't have a great name for "constant which is an alias to a class, module or other such constant alias." The name it uses exclusively internally and in the docs for these constants is "class alias" (see e.g. https://srb.help/5031).

Sorbet already knows that this constant is _not_ a reference to other kinds of constants which would be valid in type syntax. The full list of constant types allowed in type syntax are:

- Names of classes and modules (no aliases)
- Class alias constants
- Type alias constants
- Generic type members
- Enum values

The types of constants which are not allowed:

- Static field constants ("normal" constant assignments)

The message does further clarify how to fix this error:

```
If you are trying to define a type alias, you should use T.type_alias here
```

and points to the definition of the static field constant.

I could see "is not a type" being a better framing of this error than "is not a class or type alias," though I would love if there were an error message that did a better job of educating users about the subtleties of different kinds of constants allowed in type syntax.

namer/namer.cc
jez: Here is fine. We prefer not to over-abstract helpers like this.

Could you make this also check for resolved constants? Most constants are not yet resolved by namer, but sometimes resolved `T` constants are created by rewriter phases.

you'd want to add a case for `ast::ConstantLit` and check that the symbol inside it is equal to `core::Symbols::T()`.

test/testdata/resolver/type_alias_not_on_T.rb
amomchilov: > The name it uses exclusively internally and in the docs for these constants is "class alias" (see e.g. https://srb.help/5031).

Should this error message match that, then?

```suggestion
  #            ^^^^^^^^^^^^^ error: Constant `A::NotATypeAlias` is not a class alias or type alias
```

jez: That would be another option yes

jez (commented):
mostly looks great except I'd love to get out and also handle `ConstantLit`

https://github.com/sorbet/sorbet/pull/7937/files#r1626459523

namer/namer.cc
```diff
@@ -629,12 +629,24 @@ class SymbolFinder {
         return foundDefs->addTypeMember(move(found));
     }
 
+    bool sendRecvIsT(const ast::Send &s) {
```
amomchilov, line 632: I found this very similar function:

https://github.com/sorbet/sorbet/blob/92839969971586d2ef06fa7731ce8036092e7025/cfg/builder/builder_walk.cc#L109-L115

I renamed my `isT()` to the same name, and edited the parameter type to match.

namer/namer.cc
```diff
@@ -629,12 +629,24 @@ class SymbolFinder {
...
+        bool result = false;
+
+        typecase(
```
amomchilov, line 635: Opted for a `typecase` here. Seems better than the hand-rolled `switch` alternative. Is it?

```cpp
        switch (receiver.tag()) {
            case ast::Tag::ConstantLit:
                return ast::cast_tree<ast::ConstantLit>(receiver)->symbol == core::Symbols::T();
            case ast::Tag::UnresolvedConstantLit:
                return ast::cast_tree<ast::UnresolvedConstantLit>(receiver)->cnst == core::Names::Constants::T();
            default:
                return false;
        }
```

test/testdata/namer/type_alias_not_on_T.rb.symbol-table-raw.exp
```diff
@@ -0,0 +1,28 @@
+class <C <U <root>>> < <C <U Object>> ()
+  class <S <C <U <root>>> $1>[<C <U <AttachedClass>>>] < <S <C <U Object>> $1> ()
+    method <S <C <U <root>>> $1>#<N <U <static-init>> $CENSORED> (<blk>) @ Loc {file=test/testdata/namer/type_alias_not_on_T.rb start=2:1 end=20:4}
+      argument <blk><block> @ Loc {file=test/testdata/namer/type_alias_not_on_T.rb start=??? end=???}
+  module <C <U A>> < <C <U Sorbet>>::<C <U Private>>::<C <U Static>>::<C <U ImplicitModuleSuperclass>> () @ Loc {file=test/testdata/namer/type_alias_not_on_T.rb start=2:1 end=2:9}
+    static-field <C <U A>>::<C <U NotActuallyATypeAlias>> -> <todo sym> @ Loc {file=test/testdata/namer/type_alias_not_on_T.rb start=9:3 end=9:24}
```
amomchilov, line 6: This `static-field` is the key part. Prior to these changes, this test yielded `static-field-type-alias` here.

namer/namer.cc
froydnj: The compiler can convert typecase into code that is equivalent to the switch.  Pick whichever you think is more readable or your reviewer would like you to use. 😅

jez: I personally would have used

```cpp
if (auto *c = ast::cast_tree<ast::UnresolvedConstantLit>(s.recv)) {
  return c->cnst == core::Names::Constants::T();
} else if (auto *c = ast::cast_tree<ast::ConstantLit>(s.recv)) {
  return c->symbol == core::Symbols::T();
} else {
  return false;
}
```

Which _might not_ compile down to the same code as the type case / switch? But realistically all three are fine. Of the three I think my least favorite is the one that uses the `.tag()` explicitly, and leaving it with `typecase` is completely fine.

test/testdata/namer/type_alias_not_on_T.rb.symbol-table-raw.exp
jez: I think the same test could have been written with `symbol-table.exp` (not raw) since the test doesn't rely on raw names like `<C <U Private>>` to catch regressions. Even the test below which shows the error saying that this thing is not a type alias is sufficient. Anyways, changing this would not really improve the PR so I'm just commenting this for future reference.

jez (approved):

jez:
It looks like you forgot to clang-format namer.cc. Otherwise this looks good.

test/testdata/namer/type_alias_not_on_T.rb.symbol-table-raw.exp
amomchilov: Good to know, thanks!

I switched it, since I edited the commit to lint, anyway.

namer/namer.cc
amomchilov: I think I prefer `typecase`, though admittedly, I liked it more before the linter put `s.recv` on the same line as the first lambda 😅

amomchilov: Looks like `typecase()` needs to be exhaustive. This PR is failing with a bunch of `Exception::raise(): not handled typecase case: ...` errors.

How do I express the equivalent of a `default:` case?

PR #7945, opened 2024-06-05
Fix method visivility

kazzix14:
This PR fixes method visibility after calling visibility changing methods like `private`, `protected`, `public` on receivers other than local context itself.

### Motivation

Previously, methods defined after a `private` call would incorrectly be marked as private, even if the receiver is not `self`.  
Also, it can be incorrect after calling methods that changes visibility like`public`, `protected`, and so on.

Like:
```ruby
# typed: true

class Foo
  T.unsafe(nil).private

  def foo; end
end

Foo.new.foo
```

In this example, the `foo` method should not be marked as private, as the `private` call is not on `self`. However, method `foo` is marked as private and calling Foo.new.foo can not pass type checks.
```sh
> bundle exec srb tc
a.rb:9: Non-private call to private method foo on Foo https://srb.help/7031
     9 |Foo.new.foo
                ^^^
    a.rb:6: Defined in Foo here
     6 |  def foo; end
          ^^^^^^^
Errors: 1
```

In practice, code like a example below could not pass type checks.

```ruby
class Model
  configure_something Settings.s3.private.bucket_name

  attr_accessor :my_variable
end

class Controller
  def action
    # ...
    model.my_variable = 1
    # ...
  end
end
```

### Test plan

See included automated tests.

### DIFF
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -389,6 +389,11 @@ public:
                 }
                 if (!original.hasPosArgs()) {
                     ENFORCE(!methodVisiStack.empty());
+
+                    if (original.recv.tag() != ast::Tag::Local) {
+                        break;
+                    }
+
                     methodVisiStack.back() = optional<core::FoundModifier>{core::FoundModifier{
                         core::FoundModifier::Kind::Method,
                         getOwner(),
--- a/namer/namer.cc
+++ b/namer/namer.cc
@@ -390,7 +390,7 @@ public:
                 if (!original.hasPosArgs()) {
                     ENFORCE(!methodVisiStack.empty());
 
-                    if (original.recv.tag() != ast::Tag::Local) {
+                    if (!original.recv.isSelfReference()) {
                         break;
                     }
 
--- a/test/testdata/namer/visibility.rb
+++ b/test/testdata/namer/visibility.rb
@@ -66,13 +66,10 @@ class Foo6
   def foo; end
 end
 
-class NilClass
-  def private; end
-end
-
 class Foo7
   bar = nil
   bar.private
+  #   ^^^^^^^ error: Method `private` does not exist on `NilClass`
 
   def foo; end
 end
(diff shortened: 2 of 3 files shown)


namer/namer.cc
```diff
@@ -389,6 +389,11 @@ class SymbolFinder {
                 }
                 if (!original.hasPosArgs()) {
                     ENFORCE(!methodVisiStack.empty());
+
+                    if (original.recv.tag() != ast::Tag::Local) {
```
jez, line 393: Can you change this to use isSelfReference?

https://github.com/search?q=repo%3Asorbet%2Fsorbet%20isselfreference&type=code

and can you add a test like


```ruby
x = nil
x.private
# ...
```

because the change as written doesn’t work for this case.

jez (changes requested):
Thanks, two problems as written

jez:
The VS Code job test failure is a transient network issue, not related to your change.

namer/namer.cc
kazzix14: I updated to use `isSelfReference()` and added a test. Thank you for your really quick response.

test/testdata/namer/visibility.rb
```diff
@@ -53,3 +53,26 @@ def foo;
...
+  def foo; end  # this does not end up being private
+end
+
+class Foo6
+  self.private
+
+  def foo; end
+end
+
+class NilClass
```
jez, line 69: Rather than monkey patching NilClass, let’s either add an `# error:` assertion capturing the fact that private doesn’t exist on nil, or initialize bar with T.unsafe(nil).

jez (changes requested):
Sorry one more thing

test/testdata/namer/visibility.rb
kazzix14: I've felt weird of this part also. Thanks for your review. I added error an assertion on the test.

jez (approved):

PR #7946, opened 2024-06-05
Refactor `SorbetLspConfig.isEqualTo`

amomchilov:
### Motivation

The original was pretty hard to grok. This tweaked layout uses early-returns to clear things up.

### Test plan

There's no change in behaviour to test.


froydnj (approved):
This is more readable, thank you!

Issue #2020, opened 2019-10-24
Destructuring array of unknown size incorrectly infers nil


jzhou-stripe:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Bparams(x%3A%20T%3A%3AArray%5BString%5D).returns(T%3A%3AArray%5BString%5D)%7D%0A%20%20def%20bar(x)%0A%20%20%20%20*a%2C%20b%20%3D%20x%0A%20%20%20%20a%0A%20%20end%0Aend)

```ruby
# typed: true
class A
  extend T::Sig

  sig {params(x: T::Array[String]).returns(T::Array[String])}
  def bar(x)
    *a, b = x
    a
  end
end
```

#### Observed output

```
# typed: true
editor.rb:8: Returning value that does not conform to method result type https://srb.help/7005
     8 |    a
            ^
  Expected T::Array[String]
    editor.rb:6: Method bar has return type T::Array[String]
     6 |  def bar(x)
          ^^^^^^^^^^
  Got T.nilable(T::Array[String]) originating from:
    editor.rb:7:
     7 |    *a, b = x
            ^^^^^^^^^
Errors: 1
```

#### Expected behavior

Should not be an error.

- - -

paracycle:
We encountered this today and I dug in a little bit. It seems the desugaring Sorbet uses here is not exactly semantically equivalent to what destructuring does:

```
$ bundle exec srb tc --no-config -p ast -e "# typed: strict
first, *rest = [1, 2, 3]
T.reveal_type(rest)"

begin
  <emptyTree>
  class <emptyTree><<C <root>>> < (::<todo sym>)
    def self.<static-init><<static-init>$112>(<blk>)
      begin
        begin
          <assignTemp>$2 = [1, 2, 3]
          <assignTemp>$3 = ::<Magic>.<expand-splat>(<assignTemp>$2, 1, 0)
          first = <assignTemp>$3.[](0)
          rest = <assignTemp>$3.slice(::Range.new(1, -1, false))
          <assignTemp>$2
        end
        ::T.reveal_type(rest)
        <emptyTree>
      end
    end
  end
  <emptyTree>
```

Calling `.slice` on the given array returns a nilable array as a result, but `rest` should always be an empty array of the given type.

[...]

PR #7947, opened 2024-06-05
Fix splat destructuring type information (retry#2)

Morriar:
### Motivation

Fixes https://github.com/sorbet/sorbet/issues/2020

Given this example of array destructuring:

```rb
arr = [1, 2, 3]
a, *b = arr
```

Currently, the desugar phase transforms the array destructuring using `slice`:

```rb
<assignTemp>$1 = arr
<assignTemp>$2 = ::<Magic>.<expand-splat>(<assignTemp>$1, 1, 0)
a = <assignTemp>$2.[](0)
b = <assignTemp>$2.slice(::Range.new(1, -1, false))
```

The undesirable side effect is that since `slice` [returns a nilable value](https://github.com/sorbet/sorbet/blob/ede4684fc7095e7fd30499bb80887d687d871fce/rbi/core/array.rbi#L2489), `b` becomes nilable:

```rb
T.reveal_type(b) # => T.nilable(T::Array[Integer])
```

The present change, uses `to_ary` instead which returns the non-nilable type of the array:

```rb
<assignTemp>$1 = arr
<assignTemp>$2 = ::<Magic>.<expand-splat>(<assignTemp>$1, 1, 0)
a = <assignTemp>$2.[](0)
b = <assignTemp>$2.to_ary()
```

While calling `to_ary` doesn't return the correct value, it returns the correct type. Which, in regard to type checking (i.e. not within a compiler), is all we're interested in:

```rb
T.reveal_type(b) # => T::Array[Integer]
```

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -433,11 +433,8 @@ ExpressionPtr desugarMlhs(DesugarContext dctx, core::LocOffsets loc, parser::Mlh
                 }
                 auto lhloc = lh.loc();
                 auto zlhloc = lhloc.copyWithZeroLength();
-                auto index = MK::Send3(lhloc, MK::Constant(lhloc, core::Symbols::Range()), core::Names::new_(), zlhloc,
-                                       MK::Int(lhloc, left), MK::Int(lhloc, -right), std::move(exclusive));
-                stats.emplace_back(MK::Assign(
-                    lhloc, std::move(lh),
-                    MK::Send1(loc, MK::Local(loc, tempExpanded), core::Names::slice(), zlhloc, std::move(index))));
+                auto ary = MK::Send0(loc, MK::Local(loc, tempExpanded), core::Names::toAry(), zlhloc);
+                stats.emplace_back(MK::Assign(lhloc, std::move(lh), std::move(ary)));
             }
             i = -right;
         } else {
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -406,6 +406,23 @@ ExpressionPtr unsupportedNode(DesugarContext dctx, parser::Node *node) {
     return MK::EmptyTree();
 }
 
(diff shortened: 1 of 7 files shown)


froydnj:

→ https://go/builds/bui_QIAZqn3OQZ4JdM
→ https://go/builds/bui_QIAZTBf7MjVZib
→ https://go/builds/bui_QIAZX4qlOsTNNV

ast/desugar/Desugar.cc
```diff
@@ -433,11 +433,8 @@ ExpressionPtr desugarMlhs(DesugarContext dctx, core::LocOffsets loc, parser::Mlh
                 }
                 auto lhloc = lh.loc();
                 auto zlhloc = lhloc.copyWithZeroLength();
-                auto index = MK::Send3(lhloc, MK::Constant(lhloc, core::Symbols::Range()), core::Names::new_(), zlhloc,
-                                       MK::Int(lhloc, left), MK::Int(lhloc, -right), std::move(exclusive));
-                stats.emplace_back(MK::Assign(
-                    lhloc, std::move(lh),
-                    MK::Send1(loc, MK::Local(loc, tempExpanded), core::Names::slice(), zlhloc, std::move(index))));
+                auto ary = MK::Send0(loc, MK::Local(loc, tempExpanded), core::Names::toAry(), zlhloc);
+                stats.emplace_back(MK::Assign(lhloc, std::move(lh), std::move(ary)));
```
froydnj, line 437: Can you add a comment here to make it clear that this is not faithful to the runtime behavior, but that it is faithful to the expected type-checking behavior?

froydnj (approved):
This has a fair number of errors on Stripe's codebase, but a lot of the errors are just "`T.must` called on something that's not `nil`", which is autocorrectable.  The rest are similar and will require some manual fixups.  We'll have to handle those at Sorbet bump time.

PR #7948, opened 2024-06-05
Two improvements to the abstract.md docs

jez:
### Motivation

There are some undocumented features of abstract methods.

I'm not in love with the presentation of these techniques, but I think the only
way to make this better is to rework the whole abstract.md doc, OR make a "tips"
section of the docs instead of tying all docs to the feature they're referring
to.

Realistically abstract methods are as important as generics in writing modular,
type-safe code, and I don't think that our docs suitably capture that right
now—the docs for generics are better organized, more comprehensive, etc.

### Test plan

website change

### DIFF
--- a/website/docs/abstract.md
+++ b/website/docs/abstract.md
@@ -122,6 +122,157 @@ As the example shows, there are two main steps:
 Note: if you want to provide functionality in an abstract class or module that
 **must not** be possible to override in a child, use a [final method](final.md).
 
+## Letting abstract methods be implemented via inheritance
+
+Sorbet allows abstract methods in modules to be implemented by an **ancestor**
+of the class or module they're eventually mixed into. Consider this example:
+
+```ruby
+class Parent
+  sig { void }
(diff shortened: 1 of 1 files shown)


website/docs/abstract.md
```diff
@@ -122,6 +122,157 @@ As the example shows, there are two main steps:
...
+  prop :name, String
+end
+
+C.new(name: "Some::Long::Namespace").short_name # => "Namespace"
+```
+
+### Declaring requirements of an interface
+
+Letting abstract methods be implemented by inherited methods is particularly
+useful for expressing dependencies between
```
froydnj, line 237: Readers everywhere would like to have this tension resolved, and how to express their dependencies properly.

froydnj (approved):
The abstract singleton methods thing is just so useful, though.

One thing to fixup before you merge.

website/docs/abstract.md
jez: oops, I had auto merge on. I will fix

PR #7949, opened 2024-06-05
Always inherit from process.env right before launching server

st0012:
### Motivation

Before #7763, it's possible to setup VS Code's process env to help Sorbet starts its LSP. For example, Ruby LSP could handle Ruby activation and injects necessary env vars for Sorbet LSP to be started by the VS Code without additional configurations.

But that ability was lost after #7763 if:
- `activeConfig` is present
- Or, the Sorbet extension was activated a bit earlier then Ruby LSP. In that case, `SorbetLspConfig.env` would capture the same env and never captures the updated env even if the LSP failed to start.

And this PR addressed both cases.

### Test plan

All the existing tests should pass.

### DIFF
--- a/vscode_extension/package.json
+++ b/vscode_extension/package.json
@@ -4,7 +4,7 @@
   "description": "Ruby IDE features, powered by Sorbet.",
   "author": "Stripe Inc.",
   "license": "Apache-2.0",
-  "version": "0.3.34",
+  "version": "0.3.35",
   "publisher": "sorbet",
   "icon": "icon.png",
   "repository": {
@@ -274,7 +274,11 @@
           "default": false
         },
         "sorbet.highlightUntyped": {
-          "enum": ["nowhere", "everywhere-but-tests", "everywhere"],
+          "enum": [
+            "nowhere",
+            "everywhere-but-tests",
+            "everywhere"
+          ],
           "description": "Shows warning for untyped values.",
           "default": "nowhere"
         },
--- a/vscode_extension/src/languageClient.ts
+++ b/vscode_extension/src/languageClient.ts
@@ -298,7 +298,7 @@ export class SorbetLanguageClient implements Disposable, ErrorHandler {
     this.context.log.debug(` > ${command} ${args.join(" ")}`);
     this.sorbetProcess = spawn(command, args, {
       cwd: workspace.rootPath,
-      env: activeConfig?.env ?? process.env,
+      env: { ...process.env, ...activeConfig?.env },
     });
     // N.B.: 'exit' is sometimes not invoked if the process exits with an error/fails to start, as per the Node.js docs.
     // So, we need to handle both events. ¯\_(ツ)_/¯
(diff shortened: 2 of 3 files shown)


jez (commented):
Hey this change looks great, thanks for making it, I would like to test manually it but I'm not sure what the problem is right now.

I tried changing the `vscode_extension/.vscode/launch.json` to set `SRB_SORBET_EXE` to `/Users/jez/stripe/sorbet/bazel-bin/main/sorbet`, which should instruct `srb tc` to use this path instead of the one from the gem for the purpose of launching the Sorbet language server.

When I test on master, I was expecting the wrong sorbet binary to get launched, but in fact it respected my `SRB_SORBET_EXE`.

Can you explain a little more about what the problem is here, and how I can reproduce the problem in master?

st0012:
I think there's no way to reproduce the issue without involving other VS Code extensions. But hopefully I can explain how Sorbet extension used to work with Ruby LSP.

As you already know, making VS Code run Ruby programs properly requires at least a few information:
- Where the Ruby binary is
- Where the gems for that Ruby version are

And in Ruby LSP's extension, we target the most common Ruby version managers to collect them. For example, [here's](https://github.com/Shopify/ruby-lsp/blob/c8272087508a68e7192f31dc9e0982ef32176a2c/vscode/src/ruby/chruby.ts#L57-L70) chruby's. Once we get the information, we [inject](https://github.com/Shopify/ruby-lsp/blob/c8272087508a68e7192f31dc9e0982ef32176a2c/vscode/src/ruby.ts#L204-L210) those env vars into VS Code's process.

This means, not only does the Ruby LSP extension knows where to run the `ruby-lsp` gem, it also allows Sorbet's extension to run `bundle exec srb typecheck --lsp` directly, which is the behaviour Shopify devs rely on without any additional setup. And it works like this:

1. Sorbet extension starts but can't run `bundle exec srb typecheck --lsp` because it can't find Ruby...etc.
2. Sorbet extension keeps restarting
3. Ruby LSP extension starts, and injects Ruby-related envs to the process
4. Sorbet extension can now start

But if we capture and store `process.env` in `SorbetLspConfig`'s constructor, then Sorbet extension will never able to pick up the updated envs even after 3) is performed.

So basically, we want to always read the latest `process.env` before running Sorbet LSP.

vinistock:
To add a bit more to Stan's explanation, VS Code doesn't guarantee the order of activation for extensions. That means, if the Sorbet extension is picked first, it will activate before the Ruby LSP had a chance to mutate the environment and try to launch without the proper Ruby environment variables, like a modified `PATH`, `GEM_HOME` and `GEM_PATH` - failing to find gem executables.

To guarantee that this works, Sorbet's continuous attempts to restart help us, because it gives VS Code enough time to also activate the Ruby LSP, which then mutates the environment and the next retry succeeds.

But with the current implementation, the updated `process.env` is never taken into account because we're only storing the original `process.env` when Sorbet was first activated. This means Sorbet is always guaranteed to fail launching any time the Ruby LSP is activated after it.

froydnj (approved):
Thanks!

st0012:
@froydnj Thank you!

PR #7951, opened 2024-06-06
Add Random.uuid_v7 to RBIs

dmerand:
Ruby [added support](https://github.com/ruby/ruby/commit/dfb2b4cbc9aa5edc315e210bf6bfd92fcf6e45de) for `uuid_v7` to `Random::Formatter`. This PR adds a type signature to the `core` RBIs. It also adds `Random.uuid_v4` which is an alias for `Random.uuid`.

I've set the type of `extra_timestamp_bits` to be `Integer` because that's the intent of the method. Technically anything that can be cast as `Integer` would work -- happy to adjust the type if there are differing opinions about what type to use in that situation.

### DIFF
--- a/rbi/core/random.rbi
+++ b/rbi/core/random.rbi
@@ -747,4 +747,82 @@ module SecureRandom
     returns(String)
   end
   def self.uuid; end
+
+  sig do
+    returns(String)
+  end
+  def self.uuid_v4; end
+
+  # Generate a random v7 UUID (Universally Unique IDentifier).
+  #
+  #   require 'random/formatter'
+  #
+  #   Random.uuid_v7 # => "0188d4c3-1311-7f96-85c7-242a7aa58f1e"
+  #   Random.uuid_v7 # => "0188d4c3-16fe-744f-86af-38fa04c62bb5"
+  #   Random.uuid_v7 # => "0188d4c3-1af8-764f-b049-c204ce0afa23"
+  #   Random.uuid_v7 # => "0188d4c3-1e74-7085-b14f-ef6415dc6f31"
+  #   #                    |<--sorted-->| |<----- random ---->|
+  #
+  #   # or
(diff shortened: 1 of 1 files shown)


jez (approved):

dmerand:
Oh no! Committed a typo! I'll push a change.

ilyailya:

→ https://go/builds/bui_QFFBZfhLeS0l3Q
→ https://go/builds/bui_QFFBZL6mAYmb4z
→ https://go/builds/bui_QFFBHDXRvoSjAx

ilyailya (approved):

rbi/core/random.rbi
```diff
@@ -747,4 +747,82 @@ module SecureRandom
...
+  # across distributed nodes.
+  #
+  # Counters and other mechanisms for stronger guarantees of monotonicity are
+  # not implemented.  Applications with stricter requirements should follow
+  # {Section 6.2}[https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-07.html#monotonicity_counters]
+  # of the specification.
+  sig do
+    params(extra_timestamp_bits: Integer).returns(String)
+  end
+  def self.uuid_v7(extra_timestamp_bits:); end
```
thomasmarshall, line 827: It looks like the default value is `0`. If this is an optional argument should it be `T.nilable(Integer)`? https://ruby-doc.org/3.3.2/stdlibs/random/Random/Formatter.html#method-i-uuid_v7

dmerand: Good catch! Since this has been merged, perhaps a follow-up PR would be best?

thomasmarshall: Sure, sounds good to me!

dmerand: PR created: https://github.com/sorbet/sorbet/pull/7960

PR #7957, opened 2024-06-07
Swap T::Props to use kwargs for rules

zanker-stripe:
### Motivation

Move `rules` away from Ruby's default hash coalescing into kwargs, which allows us to type individual fields and provides more extensibility. Tested internally and it seems to be working fine with just these changes to Sorbet.

### Test plan

See included automated tests.

### DIFF
--- a/gems/sorbet-runtime/lib/types/props/_props.rb
+++ b/gems/sorbet-runtime/lib/types/props/_props.rb
@@ -110,7 +110,7 @@ module T::Props
     #
     # @return [void]
     sig {params(name: Symbol, cls: T.untyped, rules: T.untyped).void}
-    def prop(name, cls, rules={})
+    def prop(name, cls, **rules)
       cls = T::Utils.coerce(cls) if !cls.is_a?(Module)
       decorator.prop_defined(name, cls, rules)
     end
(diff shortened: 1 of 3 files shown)


gems/sorbet-runtime/lib/types/props/_props.rb
```diff
@@ -132,16 +132,16 @@ def plugin(mod)
...
-    def const(name, cls_or_args, args={})
+    sig {params(name: Symbol, cls_or_args: T.untyped, args: T.untyped).void}
+    def const(name, cls_or_args, **args)
       if (cls_or_args.is_a?(Hash) && cls_or_args.key?(:immutable)) || args.key?(:immutable)
         Kernel.raise ArgumentError.new("Cannot pass 'immutable' argument when using 'const' keyword to define a prop")
       end
 
       if cls_or_args.is_a?(Hash)
-        self.prop(name, cls_or_args.merge(immutable: true))
+        self.prop(name, **cls_or_args.merge(immutable: true))
```
froydnj, line 142: We should do a followup PR to root out this `cls_or_args` argument flexibility, because it's no longer there after this PR.

froydnj (approved):

gems/sorbet-runtime/lib/types/props/_props.rb
zanker-stripe: It wasn't actually clear to me where `cls_or_args` happens even in the old style.

If it's a hash, then calling `prop(name, { ... })`:

https://github.com/sorbet/sorbet/blob/2dbe43ced3c73b3bfe1a527f0527df5fca7896ae/gems/sorbet-runtime/lib/types/utils.rb#L23
https://github.com/sorbet/sorbet/blob/2dbe43ced3c73b3bfe1a527f0527df5fca7896ae/gems/sorbet-runtime/lib/types/types/fixed_hash.rb#L8

Then `const(:foo, {a: Integer})` turns into a `prop(:foo, T::Types::FixedHash.new({a: Integer, immutable: true}))`

Which doesn't seem like it should be a valid use case? It looked like one of those things that got accidentally refactored out a while ago and nobody noticed.

froydnj: Yeah, same.  Maybe it used to be `type: String` or somesuch back in the early days, and this was put here as part of the migration...?

jez: Yes, I think that the `type: String` syntax previously was allowed, and we have a static test for it even, in `ThingsWhichUsedToBePropSyntax`:

https://github.com/sorbet/sorbet/blob/ba94e8f4e714e72c4a7b536e1fcef4535dfe4178/test/testdata/rewriter/not_prop.rb#L3-L4

PR #7959, opened 2024-06-09
Fix signature for `FileUtils.{cp{,_lr,_r},mv}`

tesaguri:
These methods are implemented with `FileUtils.fu_each_src_dest` under the hood, and type of the `src` arguments is common among the methods. `fu_each_src_dest` returns the `src` argument verbatim when passed an `Array`.

Also, the methods return `nil` when passed `noop: true`.

### Motivation

This aligns with the explanation in the docs and make the signature definitions consistent betwen the methods.

### Test plan

I didn't add tests because the existing signatures don't have corresponding tests as far as I could tell.

### DIFF
--- a/rbi/stdlib/fileutils.rbi
+++ b/rbi/stdlib/fileutils.rbi
@@ -121,14 +121,14 @@ module FileUtils
   # ```
   sig do
     params(
-      src: T.any(String, Pathname),
+      src: T.any(File, String, Pathname, T::Array[T.any(File, String, Pathname)]),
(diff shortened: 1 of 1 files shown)


froydnj:

→ https://go/builds/bui_QGj65vzSWjn22a
→ https://go/builds/bui_QGj6jKMbpt2zHv
→ https://go/builds/bui_QGj67CnXOyOkWI

froydnj (approved):
Thank you!

PR #7966, opened 2024-06-13
Annotate Fiber.[] and Fiber.[]=

gmalette:
Add method definitions to `rbi/core/fiber.rbi` for `Fiber.[]` and `Fiber.[]=` methods.

### Motivation

Right now, we're getting Sorbet errors when doing `Fiber[:a]` and `Fiber[:a] = 1`, but [these methods do exist](https://ruby-doc.org/3.3.1/Fiber.html#method-c-5B-5D)

### Test plan

I have not added tests for these, should I?


ilyailya:

→ https://go/builds/bui_QHtyrPVHvv9gDY
→ https://go/builds/bui_QHtyzJWGF79MyV
→ https://go/builds/bui_QHtyZvD5ufYXdG

ilyailya (approved):

PR #7975, opened 2024-06-18
Make find-references work for foreign prop accessors

aadi-stripe:
Per the in -- find all references does not currently work for `foo_` type accessors for `foreign` props. This PR fixes the problem by moving the definition loc of the foreign prop to the `foreign:` keyword itself so it doesn't collide with the "regular" getter.

### Motivation

[See example](https://sorbet.run/#%23%20typed%3A%20strict%0A%0Aclass%20A%20%3C%20T%3A%3AStruct%0A%20%20prop%20%3Ab%2C%20T.nilable%28String%29%2C%20foreign%3A%20-%3E%20%7B%20B%20%7D%0Aend%0A%0Aclass%20B%20%3C%20T%3A%3AStruct%0Aend%0A%0Aa%20%3D%20A.new%0Aa.b%0Aa.b%20%3D%20%22hello%22%0Aa.b_)

```
class A < T::Struct
  prop :b, T.nilable(String), foreign: -> { B }
  ^^^ # Find references reports a.b, a.b = "hello", but not a.b_ below.
end

class B < T::Struct
end

a = A.new
a.b
a.b = "hello"
a.b_
```

Find-references on props reports callsites of the corresponding rewritten getters and setters, but does not include foreign getters. This is seemingly because the definition loc of the prop itself is already overloaded by the getter/setter, and we cannot reuse the same loc for the foreign prop getter.

### Test plan

### DIFF
--- a/rewriter/Prop.cc
+++ b/rewriter/Prop.cc
@@ -509,9 +509,9 @@ vector<ast::ExpressionPtr> processProp(core::MutableContext ctx, PropInfo &ret, 
 
         core::LocOffsets methodLoc;
         if (ret.foreignKwLit != nullptr) {
-          methodLoc = ret.foreignKwLit.loc();
+            methodLoc = ret.foreignKwLit.loc();
         } else {
-          methodLoc = loc;
+            methodLoc = loc;
         }
 
         auto fkMethodDef = ast::MK::SyntheticMethod1(loc, methodLoc, fkMethod, std::move(arg),
--- a/rewriter/Prop.cc
+++ b/rewriter/Prop.cc
@@ -110,6 +110,7 @@ struct PropInfo {
     ast::ExpressionPtr default_;
     core::NameRef computedByMethodName;
     core::LocOffsets computedByMethodNameLoc;
+    ast::ExpressionPtr foreignKwLit;
     ast::ExpressionPtr foreign;
     ast::ExpressionPtr enum_;
     ast::ExpressionPtr ifunset;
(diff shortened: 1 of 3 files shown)


aadi-stripe:

→ https://go/builds/bui_QJnEfT89zTFhXS
→ https://go/builds/bui_QJnEDhY9pliqDc
→ https://go/builds/bui_QJnE2jJPJ9PQ40

rewriter/Prop.cc
```diff
@@ -504,7 +506,15 @@ vector<ast::ExpressionPtr> processProp(core::MutableContext ctx, PropInfo &ret,
         auto arg = ast::MK::KeywordArgWithDefault(nameLoc, core::Names::allowDirectMutation(), ast::MK::Nil(loc));
         ast::MethodDef::Flags fkFlags;
         fkFlags.discardDef = true;
-        auto fkMethodDef = ast::MK::SyntheticMethod1(loc, loc, fkMethod, std::move(arg),
+
+        core::LocOffsets methodLoc;
+        if (ret.foreignKwLit != nullptr) {
```
jez, line 511: How is it possible to have `ret.foreign` not be `nullptr`, but to have `ret.foreignKwLit` be `nullptr`?

aadi-stripe: Because of `merchant_token_prop`, which creates an implicit foreign prop

jez (approved):

PR #7977, opened 2024-06-20
Add --rbi CLI flag to accept inline RBI input

thomasmarshall:
This PR adds an `--rbi` flag to the CLI which accepts inline RBI input, similar to how `-e` takes inline Ruby input. The implementation is almost identical to the `-e` argument, except for Sorbet treating the input as an RBI file.

### Motivation

This will be useful for quickly testing typechecking involving RBI code. In particular, it could be used by [sorbet.run](https://sorbet.run) to reproduce issues involving RBI specific functionality such as sig overloads.

### Test plan

See included automated tests.

### DIFF
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -295,6 +295,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                        stop_after_options, ", ", [](const auto &pr) -> auto{ return pr.option; }));
 
     // Advanced options
+    options.add_options("advanced")("e-rbi", "Parse an inline RBI string",
+                                    cxxopts::value<string>()->default_value(empty.inlineRBIInput), "string");
     options.add_options("advanced")("dir", "Input directory", cxxopts::value<vector<string>>());
     options.add_options("advanced")("file", "Input file", cxxopts::value<vector<string>>());
     options.add_options("advanced")("allowed-extension", "Allowed extension", cxxopts::value<vector<string>>());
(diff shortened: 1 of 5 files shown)


main/realmain.cc
```diff
@@ -715,6 +715,18 @@ int realmain(int argc, char *argv[]) {
...
+                prodCounterInc("types.input.lines");
+                prodCounterInc("types.input.files");
+                auto input = opts.inlineRBIInput;
+                if (core::File::fileStrictSigil(opts.inlineRBIInput) == core::StrictLevel::None) {
+                    // put it at the end so as to not upset line numbers
+                    input += "\n# typed: true";
+                }
+                auto file = gs->enterFile(string("--rbi.rbi"), input);
+                inputFiles.emplace_back(file);
+            }
```
jez, line 729: Let's factor this to a helper function, and call it once for `-e` and once for `--rbi`.

test/cli/rbi/test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+main/sorbet --silence-dev-message --censor-for-snapshot-tests \
+  --rbi $'module Foo; sig { void }; def self.foo; end; end' \
+  -e $'Foo.foo(123)' 2>&1
```
jez, line 4: This test doesn't actually test that the string passed to `--rbi` is treated as an RBI.

Maybe you want to try to add an overloaded sig here?

main/options/options.cc
```diff
@@ -295,6 +295,7 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                        stop_after_options, ", ", [](const auto &pr) -> auto{ return pr.option; }));
 
     // Advanced options
+    options.add_options("advanced")("rbi", "Parse an inline RBI string", cxxopts::value<string>()->default_value(empty.inlineRBIInput), "string");
```
jez, line 298: I think I might call the option `--e-rbi`, I think that makes it maybe a little less likely people misinterpret it? e.g., I could see someone mistake this flag as a zero-argument flag that forces Sorbet to treat all `*.rb` files after it as RBI files, or something else.

test/cli/rbi/test.sh
thomasmarshall: Good idea, I've added a better test.

main/options/options.cc
thomasmarshall: Sure, updated now.

jez (approved):

PR #7983, opened 2024-06-21
Overhaul rootPath support

jez:
### Motivation

There are a couple problems:

- Sorbet somewhat artificially only supports a single input directory in LSP
  mode. There's no good reason why that needs to be the case, and people
  frequently complain about it.
- Sorbet doesn't actually work in LSP mode if the `--dir` is not `.`. Certain
  things like errors will show up, but other things like go to definition will
  go to the wrong places, because those errors and jump to definition targets
  will be missing the `subdir/` prefix in their relative path.
- This lets us delete other code, like the various hacks we have to
  retroactively hide all the `./` prefixes that we forcibly prefix all paths
  with.

This fixes all of them.

### Test plan

I have been testing this manually as I've gone along, but it needs more testing.

### DIFF
--- a/common/FileOps.h
+++ b/common/FileOps.h
@@ -57,8 +57,7 @@ public:
      * Returns 'true' if the file at the given path is ignored.
      * See sorbet::options for information on absolute and relative ignore patterns.
      */
-    static bool isFileIgnored(std::string_view basePath, std::string_view filePath,
-                              const std::vector<std::string> &absoluteIgnorePatterns,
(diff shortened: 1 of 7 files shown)


rzane:
@jez, would you mind explaining why this PR was closed? It looks like there might be some unanswered questions surrounding watchman. Is that the main hurdle to clear?

jez:
It didn’t work and I don’t remember why, sorry :/

PR #7987, opened 2024-06-21
Use `watch-project` when launching watchman

### LINKED ISSUE #7982: Watchman requires the Sorbet project to be a git/hg repo
#### Input

```
mkdir ~/some-tmp-dir
cd ~/some-tmp-dir
watchman -j -p --no-pretty
```

and then type

```
["subscribe", ".", "ruby-typer-20570", {"expression": ["allof", ["type", "f"], ["anyof", ["suffix", "rb"], ["suffix", "rbi"]], ["not", ["match", "**/.~tmp~/**", "wholename", {"includedotfiles": true}]]], "fields": ["name"], "empty_on_fresh_instance": true}]
```

to the watchman stdin and press ENTER.

#### Observed output

```
{"version":"2024.05.06.00","error":"watchman::RootResolveError: failed to resolve root: unable to resolve root /Users/jez/some-tmp-dir: failed to resolve root: directory /Users/jez/some-tmp-dir is not watched"}
```

#### Expected behavior

No watchman error

- - -

These steps recreate what Sorbet does to initialize watchman.

It seems that the problem is that watchman requires there to be a `.git` or `.hg` folder in the root of the watched folder?

If you do this:

```
cd ~/some-tmp-dir
git init .
```

and then repeat the `watchman` and the `"subscribe"` command, the error goes away.

If that's something that watchman requires, Sorbet should at least show a better error message when that constraint has been violated, and at best pass some sort of configuration to the watchman process so that this does not cause an error.

jez:
Fixes #7982

### Motivation

It looks like sometimes the `subscribe` command does not create a watch before
setting up subscriptions for the project. What can happen when you don't do this
is an error like this:

```json
{
  "version": "2024.06.10.00",
  "error": "watchman::RootResolveError: failed to resolve root: unable to resolve root /Users/jez/sandbox/editor: failed to resolve root: directory /Users/jez/sandbox/editor is not watched"
}
```

It's somewhat non-deterministic, depending on whether some _other_ project has
already set up a watch for the given folder in the past that the server
remembers.

### Test plan

We don't have automated tests for watchman in CI. I've been testing this
extensively by creating a sandbox project with a `sorbet/config` and a `Gemfile`
and a `.vscode/settings.json` that allows pointing Sorbet at the local binary
instead of a gem. There's a discussion of how to do that here

<https://github.com/sorbet/sorbet/blob/master/test/sandbox/vscode/README.md>

I've mostly been using the `--debug-log-file` flag, which lets me look at the
watchman input/output and make sure that it matches up with what I expect, and
then also triggering changes to files on disk and making sure that they have the
right effect.

### DIFF
--- a/main/lsp/tools/make_lsp_types.cc
+++ b/main/lsp/tools/make_lsp_types.cc
@@ -1306,6 +1306,16 @@ void makeLSPTypes(vector<shared_ptr<JSONClassType>> &enumTypes, vector<shared_pt
                                         classTypes);
 
     /* Watchman JSON response objects */
+    auto WatchmanWatchProjectResponse =
+        makeObject("WatchmanWatchProjectResponse",
+                   {
+                       makeField("version", JSONString),
+                       makeField("watcher", JSONString),
+                       makeField("watch", JSONString),
+                       makeField("relative_path", "relativePath", makeOptional(JSONString)),
+                   },
+                   classTypes);
+
     auto WatchmanQueryResponse = makeObject("WatchmanQueryResponse",
                                             {
                                                 makeField("version", JSONString),
--- a/main/lsp/watchman/WatchmanProcess.cc
+++ b/main/lsp/watchman/WatchmanProcess.cc
@@ -80,7 +80,7 @@ optional<WatchmanProcess::ReadResponse> WatchmanProcess::readResponse(FILE *file
 void WatchmanProcess::start() {
     auto mainPid = getpid();
     try {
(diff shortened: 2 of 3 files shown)


jez:

→ https://go/builds/bui_QKwGk9dcLjmvOh
→ https://go/builds/bui_QKwGHaIt5dV3GW
→ https://go/builds/bui_QKwG5NIY0MGp8j

froydnj (approved):

PR #7989, opened 2024-06-24
Desugar `let(:f) {...}` blocks in `describe` blocks

### LINKED ISSUE #7944: Support `let`-defined methods in Minitest rewriter
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Minitest%3A%3ASpec%0A%20%20module%20DSL%0A%20%20%20%20T%3A%3ASig%3A%3AWithoutRuntime.sig%20%7B%20params%28name%3A%20Symbol%2C%20block%3A%20T.proc.returns%28T.anything%29%29.void%20%7D%0A%20%20%20%20def%20let%28name%2C%20%26block%29%0A%20%20%20%20%20%20T.unsafe%28self%29.define_method%28name%29%20do%0A%20%20%20%20%20%20%20%20%40_memoized%20%7C%7C%3D%20%7B%7D%0A%20%20%20%20%20%20%20%20%40_memoized.fetch%28name%29%20%7B%20%7Ck%7C%20%40_memoized%5Bk%5D%20%3D%20instance_eval%28%26block%29%20%7D%0A%20%20%20%20%20%20end%0A%20%20%20%20end%0A%20%20end%0A%0A%20%20extend%20DSL%0Aend%0A%0Aclass%20MyTest%20%3C%20Minitest%3A%3ASpec%0A%20%20let%28%3Amy_method%29%20%7B%200%20%7D%0A%20%20%23%20From%20a%20typing%20perspective%2C%20%5E%20this%20is%20as%20if%20you%20did%20this%3A%0A%20%20%23%0A%20%20%23%20%20%20%20%20def%20my_method%3B%200%3B%20end%0A%20%20%23%0A%20%20%23%20I%20think%20that%20Sorbet%20would%20also%20trivially%20support%20writing%0A%20%20%23%20a%20%60sig%60%20for%20this%20in%20%60%23%20typed%3A%20strict%60%20test%20files%2C%0A%20%20%23%20like%20how%20it%20works%20for%20%60attr_reader%60%20methods.%0A%0A%20%20describe%20'example'%20do%0A%20%20%20%20it%20'calls%20method%20defined%20with%20let'%20do%0A%20%20%20%20%20%20my_method%0A%20%20%20%20end%0A%20%20end%0Aend)

```ruby
# typed: true

class Minitest::Spec
  module DSL
    T::Sig::WithoutRuntime.sig { params(name: Symbol, block: T.proc.returns(T.anything)).void }
    def let(name, &block)
      T.unsafe(self).define_method(name) do
        @_memoized ||= {}
        @_memoized.fetch(name) { |k| @_memoized[k] = instance_eval(&block) }
      end
    end
  end

  extend DSL
end

class MyTest < Minitest::Spec
  let(:my_method) { 0 }
  # From a typing perspective, ^ this is as if you did this:
  #
  #     def my_method; 0; end
  #
  # I think that Sorbet would also trivially support writing
  # a `sig` for this in `# typed: strict` test files,
  # like how it works for `attr_reader` methods.

  describe 'example' do
    it 'calls method defined with let' do
      my_method
    end
  end
end
```

#### Observed output

```
editor.rb:29: Method my_method does not exist on MyTest::<describe 'example'> https://srb.help/7003
    29 |      my_method
              ^^^^^^^^^
  Did you mean method? Use -a to autocorrect
    editor.rb:29: Replace with method
    29 |      my_method
              ^^^^^^^^^
    https://github.com/sorbet/sorbet/tree/master/rbi/core/kernel.rbi#L559: Defined here
     559 |  def method(arg0); end
            ^^^^^^^^^^^^^^^^
Errors: 1
```

#### Expected behavior

Sorbet should be able to know that the `let` method defines a method with the given name.

- - -

I believe that RSpec has the same concept of `let`-defined methods, with the same API. It is slightly worrying, because there wouldn't be an easy way for Sorbet to detect whether the `let` call comes from Minitest/RSpec or from some other DSL. If it becomes a problem, we _might_ be able to limit the rewriter to only make this transformation inside `describe` blocks? I'm not sure how limiting that is in practice or whether it's needed in the first place (or for that matter, whether it's actually required because you're not allowed to use `let` outside of `describe` blocks).

jez:
Fixes #7944

### Motivation

This should make it easier to work with minitest / rspec code that makes heavy
use of `let`-defined methods.

Generally at Stripe we discourage the use of these methods, but that doesn't
mean that we need to make Sorbet not support them. We can use other mechanisms
inside Stripe to discourage the use of these methods.

In fact, supporting `let`-defined methods in Sorbet makes it easy to discourage
the use of these methods, as test making heavy use of these methods can first be
upgraded to `# typed: true` or above, and then Sorbet can guide the codemod to
ensure that any `let`-defined test helper which is rewritten to not use `let`
has been safely migrated.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/Minitest.cc
+++ b/rewriter/Minitest.cc
@@ -192,20 +192,20 @@ ast::ExpressionPtr getIteratee(ast::ExpressionPtr &exp) {
 
 ast::ExpressionPtr prepareTestEachBody(core::MutableContext ctx, core::NameRef eachName, ast::ExpressionPtr body,
                                        ast::MethodDef::ARGS_store &args, ast::InsSeq::STATS_store destructuringStmts,
-                                       ast::ExpressionPtr &iteratee);
+                                       ast::ExpressionPtr &iteratee, bool insideDescribe);
 
 // this applies to each statement contained within a `test_each`: if it's an `it`-block, then convert it appropriately,
 // otherwise flag an error about it
 ast::ExpressionPtr runUnderEach(core::MutableContext ctx, core::NameRef eachName,
                                 ast::InsSeq::STATS_store &destructuringStmts, ast::ExpressionPtr stmt,
-                                ast::MethodDef::ARGS_store &args, ast::ExpressionPtr &iteratee) {
+                                ast::MethodDef::ARGS_store &args, ast::ExpressionPtr &iteratee, bool insideDescribe) {
     // this statement must be a send
     if (auto *send = ast::cast_tree<ast::Send>(stmt)) {
+        auto correctBlockArity = send->hasBlock() && send->block()->args.size() == 0;
         // the send must be a call to `it` with a single argument (the test name) and a block with no arguments
-        if ((send->fun == core::Names::it() && send->numPosArgs() == 1 && send->hasBlock() &&
-             send->block()->args.size() == 0) ||
+        if ((send->fun == core::Names::it() && send->numPosArgs() == 1 && correctBlockArity) ||
             ((send->fun == core::Names::before() || send->fun == core::Names::after()) && send->numPosArgs() == 0 &&
-             send->hasBlock() && send->block()->args.size() == 0)) {
+             correctBlockArity)) {
             core::NameRef name;
             if (send->fun == core::Names::before()) {
                 name = core::Names::beforeAngles();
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_QM2yhqlmoZiI69
→ https://go/builds/bui_QM2yOvweAQUmD2
→ https://go/builds/bui_QM2yGaFquAjo0d

neilparikh (approved):

PR #7990, opened 2024-06-24
A handful of no-op changes to `updateKnowledge`

jez:
### Motivation

There's a lot of code duplication in the `updateKnowledge` method. We can clean
a bunch of it up relatively easily.

### Test plan

Existing tests


infer/environment.cc
```diff
@@ -513,11 +556,13 @@ void Environment::updateKnowledge(core::Context ctx, cfg::LocalRef local, core::
         return;
     }
 
+    if (!knowledgeFilter.isNeeded(local)) {
+        return;
+    }
```
froydnj, line 561: Have you checked what kind of effect this has on typechecking performance?

I worry about this being extra work...but thinking about it a little harder, since this code is guarded by `isUpdateKnowledgeName`, presumably we were doing this work anyway, and this just commonizes the code.

froydnj (approved):
This is so great.

PR #7991, opened 2024-06-25
Add flow-sensitive handling for `when *[some, tuple]`

jez:
### Motivation

The Ruby VM has special handling for splatting arrays when doing a `when`
statement:

```
❯ ruby --dump=insns -e 'case x; when *XS; end'
...
0013 splatarray                             false
0015 checkmatch                             6
0017 branchif                               22
...
```

Then in the `checkmatch` instruction, if the argument was a splatted array, it
loops over all the elements and calls `===` on them in turn, instead of calling
`===` on the array itself.

Sorbet already supports `===` for flow-sensitive type updates and exhaustiveness
checking, and we can extend that support for this special splatted array
construct. Sorbet already would desugar `case x; when *XS` to a call to
`<check-match-array>(x, XS)`. We just need to wire it up to flow-sensitivity.

I recommend reviewing by commit, they're mostly readable in isolation.

### Test plan

See included automated tests.

### DIFF
--- a/core/Names.cc
+++ b/core/Names.cc
@@ -272,6 +272,7 @@ bool NameRef::isUpdateKnowledgeName() const {
         case Names::nil_p().rawId():
         case Names::present_p().rawId():
         case Names::tripleEq().rawId():
+        case Names::checkMatchArray().rawId():
             return true;
         default:
             return false;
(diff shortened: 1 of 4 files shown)


jez:

→ https://go/builds/bui_QNCQbJvxH94mSr
→ https://go/builds/bui_QNCQxW6F3kwOu2
→ https://go/builds/bui_QNCQx46BFRF4w8

ilyailya (approved):

PR #7992, opened 2024-06-25
Fix the declLoc of `let`-defined methods

jez:
### Motivation

The old loc used to be a zero-length loc. This is better because you'll
be able to hover over the `let` and see the implicit arity. Also the
error diagnostic saying "this thing doesn't have a sig" will include the
name of the method in its location.

It looks like the `declLocForSendWithBlock` helper does not actually work as
intended, because according to my reading of Desugar, the Send loc and the Block
loc are always the same, and this is load bearing for some autocorrects that
pt fixed ages ago. I don't plan to attempt to fix that now.

### Test plan

See included automated tests.


neilparikh (commented):
Can you add a CLI test that demonstrates the change?

jez:
@neilparikh We don't need a CLI test, because the `^^^`-style tests demonstrate the change.

I didn't show this in the commit history, but before this change the `^` would have been just the zero-width loc at the beginning of the `let(:foo)` line.

neilparikh (approved):

PR #8001, opened 2024-06-27
Reapply "no-op: Factor isSingleton to take the whole type"

jez:
### Motivation

What was happening is that we were including two knowledge tests:

- if the condition is true, then `value` has type `Unset`
- if the condition is true, then `value` has type `T.class_of(Unset)`

So then Sorbet would say that `value` had type `T.all(Unset, 
T.class_of(Unset))` and then would collapse that to `T.noreturn`.

And because the `value` variable was live (used by a later block) it was 
tracked as a block argument to the `when` block. Then when attempting to 
compute it, Sorbet noticed that one of the block args was `T.noreturn` and
it reported an error that the whole block was unreachable.

The funny thing is that for the `module` case, this behavior was correct:
the block is unreachable because the `UnsetModule` type is uninhatibed.
Modules cannot be instantiated on their own, only via being mixed into a
class and then instantiating that class. But this module was final, so that
cannot happen. It's weird, in that we actually make Sorbet _worse_ at
detecting unreachable code, but the fact that it reached the right
conclusion in the `module` case was based on bad reasoning, so we're
arguably more correct here. We could potentially try to add logic back for
this case, but I don't really think it's worth it
(I would at least like to wait until someone shows me a program that would
benefit from such behavior).

### Test plan

See included automated tests.

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -481,7 +481,7 @@ namespace {
 // This is not the case for most other equality tests. e.g., x != 0 does not imply ¬ x : Integer.
 //
 // This powers (among other things) exhaustiveness checking for T::Enum.
-bool isSingleton(core::Context ctx, core::TypePtr ty) {
+bool isSingleton(core::Context ctx, core::TypePtr ty, bool includeSingletonClasses) {
     auto sym = core::Symbols::noClassOrModule();
     if (core::isa_type<core::ClassType>(ty)) {
         auto c = core::cast_type_nonnull<core::ClassType>(ty);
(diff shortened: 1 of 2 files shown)


infer/environment.cc
```diff
@@ -481,7 +481,20 @@ namespace {
 // This is not the case for most other equality tests. e.g., x != 0 does not imply ¬ x : Integer.
 //
 // This powers (among other things) exhaustiveness checking for T::Enum.
-bool isSingleton(core::Context ctx, core::ClassOrModuleRef sym) {
+bool isSingleton(core::Context ctx, core::TypePtr ty, bool includeSingletonClasses) {
```
froydnj, line 484:
```suggestion
bool isSingleton(core::Context ctx, const core::TypePtr &ty, bool includeSingletonClasses) {
```

froydnj (approved):

PR #8004, opened 2024-06-27
Allow defining constants inside T::Enum

jez:
### Motivation

Historically we prevented people from defining constants inside T::Enum's
because the implementation of `T::Enum` uses `self.constants(false)` to discover
enum values which have been created and associate them with the constants
they've been assigned into:

<https://github.com/sorbet/sorbet/blob/a8aeb8f90918b798d4ddb140146cfc06f04dd67e/gems/sorbet-runtime/lib/types/enum.rb#L356>

When we open sourced `T::Enum`, we introduced the `enums do ... end` block.
Before that, the `self.constants(false)` code ran in an autoloader hook
(specific to Stripe's legacy autoloader) which ran after a file finished
loading. The nice thing about the `enums do ... end` block is that there is now
an obvious point at which no more enum definitions can be created, meaning that
there's no technical reason why we'd need to prevent constants from being
created after the `enums do ... end` block.

Defining non-enum constants are useful especially for doing things like defining
a common group of enums or defining type aliases of individual enum values.

### Test plan

See included automated tests.

### DIFF
--- a/rewriter/TEnum.cc
+++ b/rewriter/TEnum.cc
@@ -14,8 +14,9 @@ namespace sorbet::rewriter {
 namespace {
 
 enum class FromWhere {
+    Before,
     Inside,
-    Outside,
+    After,
 };
 
 bool isTEnum(core::MutableContext ctx, ast::ClassDef *klass) {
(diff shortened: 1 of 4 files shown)


rewriter/TEnum.cc
```diff
@@ -118,6 +119,12 @@ std::optional<ProcessStatResult> processStat(core::MutableContext ctx, ast::Clas
 
     auto *selfNew = findSelfNew(asgn->rhs);
     if (selfNew == nullptr) {
+        if (fromWhere == FromWhere::After) {
+            // Allow non-enum constants to be defined after `enums do`
+            // (type aliases, type members, etc.)
```
jez, line 124: Technically you can define a type member, but at the moment it won't work well because there will be no way to redeclare that type member inside the fake singleton classes which we create for each enum value.

We could support this in a future change (and probably will in service of #4035 when we get around to it).

neilparikh (approved):

PR #8012, opened 2024-06-28
Treat `void_` as another top type in subtyping

jez:
### Motivation

For all intents and purposes, `.void` is the same as `T.anything`. If it had been up to me, I would have made `.void` be `T.anything` from the release of Sorbet and not dealt with all the confusion that `.void` has caused. But it's what we have.

Normally, it's not possible to get a type of `Sorbet::Private::Static::Void` anywhere but the top level of a method (unless you write that type out by hand in your code, which I've never seen). Which means that we can get away with very few places where we do something like "if type is void, then make it top"

But that goes out the window for `T.proc` types, and thus, all types, where now you can have `Sorbet::Private::Static::Void` nested anywhere in the type (e.g. you could have an array of void procs). It should still be the case that you can use void procs and `returns(T.anything)` procs interchangeably.

This mostly doesn't bite people because

- There is a special case for `BlockReturn`, where we widen the expected type of **blocks** specifically when the declared return type is `void`
- Most lambda and proc types produce completely untyped proc values.

I'm working on some changes which will make Sorbet infer properly-typed lambdas (at least as far as the return values are concerned), after which the fact that Sorbet does not treat void and anything interchangeably for the purpose of proc type checking becomes quite apparent.

I had originally hoped to take the chance to *completely* eliminate all the places where we do `void -> anything` by pushing it into subtyping.cc, but there are a handful of places where people _really want to know_ whether they're dealing with `void` or `anything` for the purpose of error messages. You can see this in the git commit history—I started down that road and had to walk it back.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/subtyping.cc
+++ b/core/types/subtyping.cc
@@ -228,6 +228,10 @@ TypePtr Types::lub(const GlobalState &gs, const TypePtr &t1, const TypePtr &t2) 
             categoryCounterInc("lub", "<top");
             return t1;
         }
+        if (mayBeSpecial1.symbol == Symbols::void_()) {
+            categoryCounterInc("lub", "<void");
+            return Types::top();
+        }
     }
 
     if (isa_type<ClassType>(t2)) {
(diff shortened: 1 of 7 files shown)


jez:

→ https://go/builds/bui_QNXt2YQMgba8Xf
→ https://go/builds/bui_QNXtDCC5oq67ZR
→ https://go/builds/bui_QNXtZFyL1qEYSU

ilyailya (approved):

ilyailya:

→ https://go/builds/bui_QPPNgmHJ4XwQry
→ https://go/builds/bui_QPPNti8Ne8NAyR
→ https://go/builds/bui_QPPNk910oF8RRJ

neilparikh (approved):

PR #8014, opened 2024-06-28
Special-case `return` inside `Kernel.lambda` and `-> ()`

### LINKED ISSUE #2651: `return` keyword defined in a proc is being interpreted as returning from the method
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20A%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7Breturns%20T.proc.params(arg1%3A%20String).returns(T%3A%3ABoolean)%7D%0A%20%20def%20foo%0A%20%20%20%20lambda%20do%20%7Cstr%7C%0A%20%20%20%20%20%20if%20str%20%3D%3D%20'bar'%0A%20%20%20%20%20%20%20%20return%20true%20%23%20this%20return%20is%20returning%20from%20the%20lambda%20not%20the%20method%0A%20%20%20%20%20%20else%0A%20%20%20%20%20%20%20%20return%20false%0A%20%20%20%20%20%20end%0A%20%20%20%20end%0A%20%20end%0A%0A%20%20sig%20%7Breturns%20T.proc.params(arg1%3A%20String).returns(T%3A%3ABoolean)%7D%0A%20%20def%20bar%0A%20%20%20%20lambda%20do%20%7Cstr%7C%0A%20%20%20%20%20%20if%20str%20%3D%3D%20'bar'%0A%20%20%20%20%20%20%20%20next%20true%20%23%20this%20is%20functionally%20equivalent%20to%20the%20above%20method%0A%20%20%20%20%20%20else%0A%20%20%20%20%20%20%20%20next%20false%0A%20%20%20%20%20%20end%0A%20%20%20%20end%0A%20%20end%0Aend)

```ruby
# typed: true
class A
  extend T::Sig

  sig {returns T.proc.params(arg1: String).returns(T::Boolean)}
  def foo
    lambda do |str|
      if str == 'bar'
        return true # this return is returning from the lambda not the method
      else
        return false
      end
    end
  end

  sig {returns T.proc.params(arg1: String).returns(T::Boolean)}
  def bar
    lambda do |str|
      if str == 'bar'
        next true # this is functionally equivalent to the above method
      else
        next false
      end
    end
  end
end
```

#### Observed output

```
editor.rb:11: Returning value that does not conform to method result type https://srb.help/7005
    11 |        return false
                ^^^^^^^^^^^^
  Expected T.proc.params(arg0: String).returns(T::Boolean)
    editor.rb:6: Method foo has return type T.proc.params(arg0: String).returns(T::Boolean)
     6 |  def foo
          ^^^^^^^
  Got FalseClass originating from:
    editor.rb:11:
    11 |        return false
                       ^^^^^

editor.rb:9: Returning value that does not conform to method result type https://srb.help/7005
     9 |        return true # this return is returning from the lambda not the method
                ^^^^^^^^^^^
  Expected T.proc.params(arg0: String).returns(T::Boolean)
    editor.rb:6: Method foo has return type T.proc.params(arg0: String).returns(T::Boolean)
     6 |  def foo
          ^^^^^^^
  Got TrueClass originating from:
    editor.rb:9:
     9 |        return true # this return is returning from the lambda not the method
                       ^^^^
Errors: 2
```

#### Expected behavior

I would expect there to be no errors. The `return` statements on line 9 and 11 are returning from the `lambda`, not from the method.

- - -

jez:
### Motivation

Fixes #2651

We could expand this in the future to also handle plain
`lambda { return }` but that would be taking a leap of faith that
`lambda` is `Kernel#lambda`, which might not be the case if there are DSLs
that define methods with this name that do not have lambda's same
`return` semantics.

### Test plan

See included automated tests.

### DIFF
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -669,7 +669,9 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
                                            .withBlockBreakTarget(cctx.target)
                                            .withLoopScope(headerBlock, postBlock, true)
                                            .withSendAndBlockLink(link);
-                        newCctx.isInsideLambda = isKernelLambda(s);
+                        if (isKernelLambda(s)) {
+                            newCctx.isInsideLambda = true;
+                        }
                         blockLast = walk(newCctx, s.block()->body, argBlock);
                     }
                     if (blockLast != cctx.inWhat.deadBlock()) {
--- a/cfg/builder/builder_walk.cc
+++ b/cfg/builder/builder_walk.cc
@@ -228,6 +228,29 @@ BasicBlock *CFGBuilder::walkHash(CFGContext cctx, ast::Hash &h, BasicBlock *curr
     return current;
 }
 
+BasicBlock *CFGBuilder::walkBlockReturn(CFGContext cctx, core::LocOffsets loc, ast::ExpressionPtr &expr,
+                                        BasicBlock *current) {
+    LocalRef exprSym = cctx.newTemporary(core::Names::nextTemp());
+    auto afterNext = walk(cctx.withTarget(exprSym), expr, current);
+    if (afterNext != cctx.inWhat.deadBlock() && cctx.isInsideRubyBlock) {
+        LocalRef dead = cctx.newTemporary(core::Names::nextTemp());
+        ENFORCE(cctx.link.get() != nullptr);
+        afterNext->exprs.emplace_back(dead, loc, make_insn<BlockReturn>(cctx.link, exprSym));
+    }
+
+    if (cctx.nextScope == nullptr) {
+        if (auto e = cctx.ctx.beginError(loc, core::errors::CFG::NoNextScope)) {
+            e.setHeader("No `{}` block around `{}`", "do", "next");
+        }
+        // I guess just keep going into deadcode?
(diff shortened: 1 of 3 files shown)


cfg/builder/builder_walk.cc
```diff
@@ -621,11 +663,17 @@ BasicBlock *CFGBuilder::walk(CFGContext cctx, ast::ExpressionPtr &what, BasicBlo
...
-                                          s.block()->body, argBlock);
+                    BasicBlock *blockLast;
+                    {
+                        auto newCctx = cctx.withTarget(blockrv)
+                                           .withBlockBreakTarget(cctx.target)
+                                           .withLoopScope(headerBlock, postBlock, true)
+                                           .withSendAndBlockLink(link);
+                        if (isKernelLambda(s)) {
+                            newCctx.isInsideLambda = true;
+                        }
```
neilparikh, line 674: Why did you change it from `newCctx.isInsideLambda = isKernelLambda(s);` to this?

neilparikh (approved):

cfg/builder/builder_walk.cc
jez: It doesn’t reset. If you’re in an each block inside a lambda, return returns from the lambda, not the method

jez:

→ https://go/builds/bui_QObuVczL6DocWQ
→ https://go/builds/bui_QObuwMbZBHk1xs
→ https://go/builds/bui_QObuZjYJDw5oSD

PR #8015, opened 2024-06-29
Allow ignoring directories for packaging purposes

aisamanra:
### Motivation
This adds a flag, `--ignore-packaging-in-directories`, which can be used for treating the constants within a directory as unpackaged. This is particularly useful for top-level RBI files without having to use the leading `::` trick.

This is implemented by adding a second instance of `NONE_PKG` called `IGNORED_PKG`, which mostly acts like `NONE_PKG` (e.g. `IGNORED_PKG.exists()` will return `false`) but can be distinguished using the new method `ignoreForPackaging()`, which returns `true` on `IGNORED_PKG`. With this, _most_ of the other stuff works out the same—for example, using constants from `IGNORED_PKG` doesn't incur a dependency violation—and we have an extra branch in `validatePackagedFile` that won't flag an unpackaged error on files which map to this package.

I tried a few other approaches, but a lot of them were a fair bit more heavy-weight and did string comparisons way too often; this one takes advantage of the same package caching, which means we only do a bit of extra string comparison at the very beginning and that's it.

### Test plan

See included automated tests. Added a test which exercises an ignored directory.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -1482,10 +1482,12 @@ void validatePackagedFile(core::Context ctx, const ast::ExpressionPtr &tree) {
     auto &pkg = ctx.state.packageDB().getPackageForFile(ctx, ctx.file);
     if (!pkg.exists()) {
         // Don't transform, but raise an error on the first line.
-        if (auto e = ctx.beginError(core::LocOffsets{0, 0}, core::errors::Packager::UnpackagedFile)) {
-            e.setHeader("File `{}` does not belong to a package; add a `{}` file to one "
-                        "of its parent directories",
-                        ctx.file.data(ctx).path(), PACKAGE_FILE_NAME);
+        if (!pkg.ignoreForPackaging()) {
+            if (auto e = ctx.beginError(core::LocOffsets{0, 0}, core::errors::Packager::UnpackagedFile)) {
+                e.setHeader("File `{}` does not belong to a package; add a `{}` file to one "
+                            "of its parent directories",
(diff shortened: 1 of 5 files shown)


packager/packager.cc
```diff
@@ -1482,10 +1482,12 @@ void validatePackagedFile(core::Context ctx, const ast::ExpressionPtr &tree) {
...
         // Don't transform, but raise an error on the first line.
-        if (auto e = ctx.beginError(core::LocOffsets{0, 0}, core::errors::Packager::UnpackagedFile)) {
-            e.setHeader("File `{}` does not belong to a package; add a `{}` file to one "
-                        "of its parent directories",
-                        ctx.file.data(ctx).path(), PACKAGE_FILE_NAME);
+        if (!pkg.ignoreForPackaging()) {
+            if (auto e = ctx.beginError(core::LocOffsets{0, 0}, core::errors::Packager::UnpackagedFile)) {
+                e.setHeader("File `{}` does not belong to a package; add a `{}` file to one "
+                            "of its parent directories",
+                            ctx.file.data(ctx).path(), PACKAGE_FILE_NAME);
```
jez, line 1489: What do you think about an alternative implementation where we still require all files to be in a package, so you would have to have a `__package.rb` file somewhere in your `sorbet/rbi/` folder, but we make the option treat the files as `# packaged: false` if they're inside those directories? Or alternatively, we allow specifying this in the `__package.rb` file itself instead of command line options?

The refactors that I'm working on to move the package DB into the symbol table assume that every file has a package.

Issue #7568, opened 2024-01-03
Missing definition for `Range#overlaps?` (Ruby 3.3)


andyw8:
https://www.shakacode.com/blog/ruby-3-3-introduces-range-overlap-method/

I will open a PR if I have time.

andyw8:
Note: Rails defines this for Ruby 3.2 and below.

https://github.com/rails/rails/pull/49317/commits/1d76d454113147cb2868dd7073d17a73fa8278dc

PR #8019, opened 2024-07-03
Add signature for Ruby 3.3 `Range#overlap?`

jacobobq:
This adds a signature for the new [`Range#overlap?`](https://docs.ruby-lang.org/en/3.3/Range.html#method-i-overlap-3F) method from Ruby 3.3.
Solves #7568.

We've set the input argument to `T::Range[T.untyped]` as the API specifies that passing an argument of an uncomparable type should return `false`. We also considered `T::Range[Elem]` but this would exclude what look like legitimate usecases.

### Motivation
In Ruby 3.3, ranges have an overlap? method defined for them in the standard library.

### Test plan

No new tests were added.

### DIFF
--- a/rbi/core/range.rbi
+++ b/rbi/core/range.rbi
@@ -603,4 +603,77 @@ class Range < Object
     .returns(T::Boolean)
   end
   def member?(obj); end
+
+  # Returns `true` if `range` overlaps with `self`, `false` otherwise:
+  # 
+  # ```ruby
+  # (0..2).overlap?(1..3) #=> true
+  # (0..2).overlap?(3..4) #=> false
+  # (0..).overlap?(..0)   #=> true
+  # ``` 
+  #
+  # With non-range argument, raises [`TypeError`](https://docs.ruby-lang.org/en/3.3/TypeError.html).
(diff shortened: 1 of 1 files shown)


rbi/core/range.rbi
```diff
@@ -603,4 +603,77 @@ class Range < Object
...
+  # ```ruby
+  # (...-Float::INFINITY).overlap?(...-Float::INFINITY) # => true
+  # (..."").overlap?(..."") # => true
+  # (...[]).overlap?(...[]) # => true
+  # ```
+  # 
+  # Even if those ranges are effectively empty (no number can be smaller than `-Float::INFINITY`), they are still
+  # considered overlapping with themselves.
+  # Related: [`Range#cover?`](https://docs.ruby-lang.org/en/3.3/Range.html#method-i-cover-3F).
+  sig { params(other: T::Range[T.untyped]).returns(T::Boolean) }
```
jez, line 676:
```suggestion
  sig { params(other: T::Range[T.anything]).returns(T::Boolean) }
```

jez (approved):

PR #8023, opened 2024-07-05
Infer proc return types

jez:
### Motivation

These are the same change as a2a3f1088 but for `Kernel#proc` instead of
`Kernel#lambda`.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -2661,19 +2661,16 @@ private:
         }
 
         {
+            auto nonNilPassedInBlockType = Types::dropNil(gs, passedInBlockType);
+            auto passedInBlockReturnType = Types::getProcReturnType(gs, nonNilPassedInBlockType);
             auto it = &dispatched;
             while (it != nullptr) {
                 if (it->main.method.exists()) {
-                    const auto &methodArgs = it->main.method.data(gs)->arguments;
-                    ENFORCE(!methodArgs.empty());
-                    const auto &bspec = methodArgs.back();
-                    ENFORCE(bspec.flags.isBlock);
-
-                    auto bspecType = bspec.type;
-                    if (bspecType) {
+                    const auto &blockReturnType = it->main.blockReturnType;
+                    if (blockReturnType) {
                         // TODO(jez) How should this interact with highlight untyped?
                         // This subtype check is here to discover the correct generic bounds.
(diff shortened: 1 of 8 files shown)


jez:

→ https://go/builds/bui_QQAaAQdzafspNd
→ https://go/builds/bui_QQAaAPCjMpmzY4
→ https://go/builds/bui_QQAaObusbS0v1J

jez:

→ https://go/builds/bui_QUJ5gMHblnFSQ9
→ https://go/builds/bui_QUJ5i05ZwYDN53
→ https://go/builds/bui_QUJ5DvW4KDarxA

jez:

→ https://go/builds/bui_QV6T4qEnHrrcIq
→ https://go/builds/bui_QV6TE93cq44Ib2
→ https://go/builds/bui_QV6TIFJGba4isZ

core/types/calls.cc
```diff
@@ -2661,19 +2661,16 @@ class Magic_callWithBlock : public IntrinsicMethod {
...
-                    const auto &bspec = methodArgs.back();
-                    ENFORCE(bspec.flags.isBlock);
-
-                    auto bspecType = bspec.type;
-                    if (bspecType) {
+                    const auto &blockReturnType = it->main.blockReturnType;
+                    if (blockReturnType) {
                         // TODO(jez) How should this interact with highlight untyped?
                         // This subtype check is here to discover the correct generic bounds.
-                        Types::isSubTypeUnderConstraint(gs, *constr, passedInBlockType, bspecType,
```
jez, line 2676: Sorbet currently allows the parameter types of `blk` in `foo(&blk)` to accumulate constraints onto the `T.type_parameter`'s defined on `foo`'s `T.proc.params`. But that's not how `foo do |x| blk.call(x) end` would have worked—written like that, Sorbet would only allow accumulating constraints against `T.type_parameter`'s of `foo` if those type parameters occur in the block return type (not params).

The fact that Sorbet worked like this was hiding a type error in pay-server.

in general, the idea behind this `call-with-block` code should be to approximate the logic in `BlockReturn` from processBinding, and `BlockReturn` only uses calls `isSubTypeUnderConstraint` on `blockReturnType`, not on the whole `bspecType`.

neilparikh (approved):

PR #8031, opened 2024-07-12
Fix crash with empty file and `--cache-dir`

jez:
### Motivation

This is cursed.

My understanding:

- when we unpickle an AST from the cache, we do so in an empty global
  state (created with `GlobalState::initEmpty()`. That basically only
  has the names from `generate_names` and the symbols defined explicitly
  in `initEmpty`. it doesn't have anything in our payload.
- we do `NameRef::fromRawUnchecked` when we read a `uint32_t` to
  unpickle a NameRef which means we don't get the "this name ref is
  owned by the right GlobalState checking that our `NameRefDebugCheck`
  logic is supposed to do
- we DO run a sanity check that says "if you try to re-enter this name
  into the current GlobalState, you get back the same ID. since the
  `GlobalState` is basically empty, it doesn't have any custom names and
  that sanity check fails (the new ID is different)
- the only reason why this works in the real world is that unpickling an
  AST never creates NameRefs... and we skip the `NameSubstitution` step
  when merging index results for trees that were read out of the cache
  (!!)

The error that's happening is triggered by a call to `sanityCheck` in
unpickleNameRef. The only thing that `NameRef::sanityCheck` does is
reenter the name and check whether it's the same ID, and it seems like
for unpickled names this should never be the case.

Stepping back… I think this all stems from the fact that we apparently
never had a test for the warm `--disk-cache` path??

### Test plan

See included automated tests.

### DIFF
--- a/core/serialize/serialize.cc
+++ b/core/serialize/serialize.cc
@@ -1693,7 +1693,19 @@ ast::ExpressionPtr SerializerImpl::unpickleExpr(serialize::UnPickler &p, const G
 
 NameRef SerializerImpl::unpickleNameRef(UnPickler &p, const GlobalState &gs) {
     NameRef name = NameRef::fromRawUnchecked(p.getU4());
-    name.sanityCheck(gs);
+
+    // We don't call NameRef::sanityCheck here because basically the only thing that method is going
+    // to do is attempt to const_cast (!!) the GlobalState and re-enter the NameRef, in an attempt
+    // to ENFORCE that re-entering the thing with the given name produces the same result.
+    //
+    // For unpickled NameRefs, that might not be the case: the GlobalState we're unpickling a tree
+    // that contains NameRef with might be a tree created by `GlobalState::copyForIndex`, which
+    // doesn't have a complete name table (it will only have the names created by generate_names.cc)
+    // which normally isn't important because mergeIndexResults skips ast::Substitute::run for trees
+    // which came out of the cache (e.g., it assumes those trees are getting merged into a
+    // GlobalState which was already fetched from the cache and thus already had the name table
+    // loaded via Serializer::loadGlobalState).
+
     return name;
 }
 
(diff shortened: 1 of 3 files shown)


test/cli/warm-cache/test.out
```diff
@@ -0,0 +1,14 @@
...
+                         types.input.files :              3
+             types.input.files.kvstore.hit :              3
+====first after change (one cache miss)====
+                         types.input.files :              3
+             types.input.files.kvstore.hit :              2
+            types.input.files.kvstore.miss :              1
+====second after change (warm cache)====
+                         types.input.files :              3
+             types.input.files.kvstore.hit :              2
+            types.input.files.kvstore.miss :              1
```
jez, line 14: This should be 

```suggestion
             types.input.files.kvstore.hit :              3
```

and I'm not sure why it's not. But at least we don't crash now so this is progress?

jez: This turned out to be a separate issue: we only write the trees to the cache if GlobalState changed, which mean that adding a comment to a file will cause a cache miss but prevent that file from getting written back into the cache.

jez:

→ https://go/builds/bui_QUbxsRzZ4vU0Vr
→ https://go/builds/bui_QUbxMJS92uj5K5
→ https://go/builds/bui_QUbxXBOJetORtC

core/serialize/serialize.cc
```diff
@@ -1693,7 +1693,19 @@ ast::ExpressionPtr SerializerImpl::unpickleExpr(serialize::UnPickler &p, const G
 
 NameRef SerializerImpl::unpickleNameRef(UnPickler &p, const GlobalState &gs) {
```
neilparikh, line 1694: You can delete the `gs` param now.

neilparikh (approved):

core/serialize/serialize.cc
jez: will do in a follow up, i have a stack on this

PR #8032, opened 2024-07-12
Fix crash when trying to extract the receiver of a safe navigation send

neilparikh:
### Test plan

See included automated tests.


jez (approved):
neat!

PR #8034, opened 2024-07-16
wip: Add error when instantiating generic without type arguments

jez:
### Motivation

Not sure that this is an approach I want to commit to, but I want to see how
many errors this generates.

### Test plan

See included automated tests.

### DIFF
--- a/core/Symbols.cc
+++ b/core/Symbols.cc
@@ -2248,8 +2248,8 @@ TypeParameter TypeParameter::deepCopy(const GlobalState &to) const {
     return result;
 }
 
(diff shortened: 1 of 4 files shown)


jez:

→ https://go/builds/bui_QUa8gexgQdlz9e
→ https://go/builds/bui_QUa81JfF37vg93
→ https://go/builds/bui_QUa8E8YB3YHVuw

jez:
there's quite a few of these and it seems particularly punishing to require people to add these type arguments, because in the vast majority of them it's fair to assume that Sorbet should have been able to infer them.

PR #8035, opened 2024-07-17
Always write the trees to the cache

jez:
### Motivation

This fixes a bug where we were not writing newly parsed trees to the cache if
GlobalState didn't change. That could happen whenever a comment or whitespace
was added.

The `cacheTreesAndFiles` function already knows how to use the `.cached()` bit
on a file to know that if that tree was read out of the cache that we don't have
re-write it into the cache.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_QUc1tZyNbG9jRS
→ https://go/builds/bui_QUc1N5iicibudR
→ https://go/builds/bui_QUc1Cka29GjGph

ilyailya (approved):

PR #8038, opened 2024-07-17
Hide Hash::Elem in detailed type errors

jez:
### Motivation

Two types disagreeing for the type of `Hash::Elem` is always noise, because
Sorbet creates this type implicitly when constructing a hash. It behaves as if
it has the type `[Hash::K, Hash::V] (2-tuple)` for all hashes.

### Test plan

See included automated tests.

### DIFF
--- a/core/types/subtyping.cc
+++ b/core/types/subtyping.cc
@@ -1259,29 +1259,32 @@ bool isSubTypeUnderConstraintSingle(const GlobalState &gs, TypeConstraint &const
             if (!doesMemberMatch) {
                 result = false;
                 if constexpr (shouldAddErrorDetails) {
-                    string variance;
-                    string joiningText;
-                    switch (idxTypeMember.data(gs)->variance()) {
-                        case Variance::CoVariant: {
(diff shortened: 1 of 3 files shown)


core/types/subtyping.cc
```diff
@@ -1259,29 +1259,32 @@ bool isSubTypeUnderConstraintSingle(const GlobalState &gs, TypeConstraint &const
...
-                            variance = "invariant";
-                            joiningText = "equivalent to";
-                            break;
-                        }
-                        case Variance::ContraVariant: {
-                            variance = "contravariant";
-                            joiningText = "a supertype of";
-                            break;
+                    if (!(a2->klass == Symbols::Hash() &&
+                          idxTypeMember.data(gs)->name == core::Names::Constants::Elem())) {
```
neilparikh, line 1263: I think you can flip the condition and then do a early `break` here, to avoid the additional nesting.

neilparikh (approved):
one optional nit, LGTM otherwise

PR #8044, opened 2024-07-19
Time#- and Time#+ can accept any class that implements to_r or to_int

KaanOzkan:
```ruby
irb(main):001* class Foo
irb(main):002*   def to_r = 1_000_000
irb(main):003> end
=> :to_r
irb(main):004> Time.now - Foo.new.to_r
=> 2024-07-08 02:35:41.949698 -0400
irb(main):005> Time.now + Foo.new.to_r
=> 2024-07-31 06:09:26.51858 -0400
```

https://github.com/ruby/ruby/blob/1e4c4fe478fccdc00565c609f6128172728d0f04/time.c#L528-L545

### Test plan

See included automated tests.


rbi/core/time.rbi
```diff
@@ -355,13 +355,13 @@ class Time < Object
   # ```
   sig do
     params(
-        arg0: Time,
+        arg0: BasicObject,
```
paracycle, line 358: I guess this specific overload still makes sense, since if a `Time` instance is supplied, the result would be a `Float` but otherwise, the value would be coerced to a numeric value and the result would be a `Time.`

jez:

→ https://go/builds/bui_QXDeC6nTXW40V3
→ https://go/builds/bui_QXDeRN5T7CYbHi
→ https://go/builds/bui_QXDeznn1OTaO5R

jez (approved):

PR #8050, opened 2024-07-24
Make Sorbet faster to start in debug build

jez:
### Motivation

It had been starting to annoy me how long it would take to start Sorbet in debug
mode, so I went and profiled it and culled the biggest offenders.

### Test plan

Existing tests

```
~/sorbet (master)
❯ hyperfine --warmup 3 "bazel-bin/main/sorbet -e 0"
Benchmark 1: bazel-bin/main/sorbet -e 0
  Time (mean ± σ):     959.5 ms ±  13.4 ms    [User: 930.1 ms, System: 20.4 ms]
  Range (min … max):   939.5 ms … 985.7 ms    10 runs

~/sorbet (jez-timers)
❯ hyperfine --warmup 3 "bazel-bin/main/sorbet -e 0"
Benchmark 1: bazel-bin/main/sorbet -e 0
  Time (mean ± σ):     532.9 ms ±  15.0 ms    [User: 504.4 ms, System: 18.8 ms]
  Range (min … max):   520.3 ms … 567.2 ms    10 runs
```

### DIFF
--- a/core/GlobalState.cc
+++ b/core/GlobalState.cc
@@ -2028,63 +2028,69 @@ void GlobalState::sanityCheck() const {
             namesUsedTotal(), namesByHash.capacity());
     ENFORCE(namesByHash.size() == namesByHash.capacity(), "hash name table not at full capacity");
 
-    for (uint32_t i = 0; i < utf8Names.size(); i++) {
-        NameRef(*this, NameKind::UTF8, i).sanityCheck(*this);
-    }
+    {
+        Timer timeit(tracer(), "GlobalState::sanityCheck (names)");
+        for (uint32_t i = 0; i < utf8Names.size(); i++) {
+            NameRef(*this, NameKind::UTF8, i).sanityCheck(*this);
+        }
 
-    for (uint32_t i = 0; i < constantNames.size(); i++) {
-        NameRef(*this, NameKind::CONSTANT, i).sanityCheck(*this);
-    }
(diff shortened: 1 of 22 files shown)


core/GlobalState.cc
```diff
@@ -2027,22 +2028,46 @@ void GlobalState::sanityCheck() const {
...
+        return;
+    }
+
+    Timer timeit(tracer(), "GlobalState::sanityCheck");
+
+    sanityCheckTableSizes();
+    sanityCheckNames();
+
     {
         Timer timeit(tracer(), "GlobalState::sanityCheck (symbols)");
```
neilparikh, line 2042: Any reason not to split this off into `sanityCheckSymbols`? Do we always want to also do `sanityCheckTableSizes` and `sanityCheckNames` when this is done?

neilparikh (changes requested):
You have a test failure

core/GlobalState.cc
jez: I'm not quite sure what you're asking.

There were some places where I **only** wanted to sanity check the table sizes, or sanity check the names, so I factored out helpers for these.

There were no places where I **only** wanted to sanity check the symbols, so I did not factor them out of the top-level sanity check function.

neilparikh: Ah, that answers exactly what I was asking, thanks.

neilparikh (approved):

PR #8056, opened 2024-07-27
Update sigs for Array#sample

bitwise-aiden:
Updating `Array#sample` to have multiple sigs for disambiguation so that if a size is passed or not it will represent the correct value of `T.nilable(Elem)` or `T::Array[Elem]` accordingly. 

```ruby
[1, 2, 3].sample      # T.nilable(Elem)
[1, 2, 3].sample(2)   # T::Array[Elem]
```

This puts the method inline with other Array methods like `Array#last` that does the same tactic.

### Motivation

At present the `Array#sample` method has a single signature returning `T.any(T.nilable(Elem), T::Array[Elem])` so the caller isn't able to get the correct typing for it. Using the 

### Test plan

Added some new tests to the array rbi test suite.

### DIFF
--- a/rbi/core/array.rbi
+++ b/rbi/core/array.rbi
@@ -2263,10 +2263,16 @@ class Array < Object
   # ```
   sig do
     params(
+        random: Random::Formatter,
+    )
+    .returns(T.nilable(Elem))
+  end
+  sig do
+    params(
         arg0: Integer,
         random: Random::Formatter,
     )
-    .returns(T.any(T.nilable(Elem), T::Array[Elem]))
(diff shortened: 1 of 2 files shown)


jez (commented):
Unfortunately, Sorbet's support for overloaded methods isn't expressive enough to allow this. `random:` is a keyword parameter, and Sorbet does not allow overloaded methods to have keyword parameters.

bitwise-aiden:
Right, so that is why it differentiates from other methods like `first|last`?

amomchilov:
Dupe of #7561

PR #8057, opened 2024-07-29
Revert "Make Sorbet faster to start in debug build (#8050)"

jez:
This reverts commit 5101acfc7268f63cbab1ca714c19529bf432b665.

### Motivation

Somehow this caused CI to become flaky?

### Test plan

```
bazel test --config=buildfarm-sanitized-linux --test_summary=terse --build_tests_only //test/lsp:multithreaded_protocol_test_corpus --runs_per_test=200
```

This failed with a 7% rate on my devbox with the changes in #8050, and didn't
fail at all without those changes.

### DIFF
--- a/ast/verifier/Verifier.cc
+++ b/ast/verifier/Verifier.cc
@@ -36,7 +36,7 @@ public:
 };
 
 ExpressionPtr Verifier::run(core::Context ctx, ExpressionPtr node) {
-    if constexpr (!debug_mode) {
+    if (!debug_mode) {
         return node;
     }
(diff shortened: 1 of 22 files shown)


froydnj (approved):
Do you have the energy to try to reland individual changes, so maybe we can bisect on master?

jez:
> Do you have the energy to try to reland individual changes, so maybe we can bisect on master?

I'm already working on this, it just takes 30 minutes to run the test to bisect because the multithreaded test is slow and the failures are so rare to reproduce.

PR #8059, opened 2024-07-30
Fix signature for `String#match`

ivanbrennan:
### Motivation
When given a block, `String#match` calls the block with the computed matchdata and returns the block’s return value.

https://ruby-doc.org/3.2.2/String.html#method-i-match

### Test plan
See included automated tests.

### DIFF
--- a/test/testdata/rbi/string.rb
+++ b/test/testdata/rbi/string.rb
@@ -24,3 +24,20 @@ T.assert_type!(w, T::Array[String])
 "foo".encode("encoding", "other_encoding", fallback: {})
 "foo".encode("encoding", fallback: {})
 "foo".encode(fallback: {})
+
+# match
+m1 = "foo".match("f")
+m2 = "foo".match("f", 1)
+T.assert_type(m1, T.nilable(MatchData))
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_QZYFMQdYmBUYd7
→ https://go/builds/bui_QZYF6SCGz8KLqO
→ https://go/builds/bui_QZYFSJF5To62w6

jez (approved):
thanks!

test/testdata/rbi/string.rb
```diff
@@ -24,3 +24,20 @@
 "foo".encode("encoding", "other_encoding", fallback: {})
 "foo".encode("encoding", fallback: {})
 "foo".encode(fallback: {})
+
+# match
+m1 = "foo".match("f")
+m2 = "foo".match("f", 1)
+T.assert_type(m1, T.nilable(MatchData))
```
neilparikh, line 31: This (and all the other uses) should be `T.assert_type!`

ivanbrennan: Thanks. I've updated it in 2523c5371.

jez (approved):

PR #8061, opened 2024-07-31
When an invalid loc is computed, log debug info and return no code actions instead of crashing

neilparikh:
### Motivation

We should not crash when the user executes an extract action that triggers an edge case we don't handle.

### Test plan

Ran a sorbet release build with the `--debug-log-file=`, ran a known crashing extract action, saw that sorbet did not crash, and saw the relevant logs in the log file:
```
[2024-07-31 14:11:07.341] [consoleAndFile] [error] msg="ExtractToVariable: failed to determine whereToInsert in getExtractMultipleOccurrenceEdits" selectionLoc="Loc {file=./test.rb start=5:4 end=5:5}"
[2024-07-31 14:11:07.341] [consoleAndFile] [error] source="# typed: strict\n\na = T.unsafe([])\n\nif a && T.unsafe(1)\nend\n"
```

### DIFF
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -9,9 +9,8 @@ namespace sorbet::realmain::lsp {
 
 void logDebugInfo(const std::shared_ptr<spdlog::logger> logger, const core::GlobalState &gs,
                   const core::Loc selectionLoc, const std::string message) {
-    logger->debug("ExtractToVariable: {}", message);
-    logger->debug("selectionLoc=\"{}\"", selectionLoc.showRaw(gs));
-    logger->debug("source=\"{}\"", absl::CEscape(selectionLoc.file().data(gs).source()));
+    logger->error("msg=\"ExtractToVariable: {}\" selectionLoc=\"{}\"", message, selectionLoc.showRaw(gs));
+    logger->error("source=\"{}\"", absl::CEscape(selectionLoc.file().data(gs).source()));
 }
 
 core::LocOffsets findWhereToInsert(const ast::ExpressionPtr &scope, const core::LocOffsets target) {
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -1,10 +1,18 @@
 #include "main/lsp/ExtractVariable.h"
+#include "absl/strings/escaping.h"
 #include "ast/treemap/treemap.h"
 #include "common/sort/sort.h"
 
 using namespace std;
 
 namespace sorbet::realmain::lsp {
+
+void logDebugInfo(const std::shared_ptr<spdlog::logger> logger, const core::GlobalState &gs,
+                  const core::Loc selectionLoc, const std::string message) {
+    logger->debug("ExtractToVariable: {}", message);
+    logger->debug("selectionLoc=\"{}\"", selectionLoc.showRaw(gs));
+    logger->debug("source=\"{}\"", absl::CEscape(selectionLoc.file().data(gs).source()));
+}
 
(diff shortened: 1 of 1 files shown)


main/lsp/ExtractVariable.cc
```diff
@@ -467,6 +480,11 @@ MultipleOccurrenceResult VariableExtractor::getExtractMultipleOccurrenceEdits(co
     auto firstMatch = matches[0];
 
     auto whereToInsert = findWhereToInsert(*scopeToInsertIn, firstMatch);
+    if (!whereToInsert.exists()) {
+        logDebugInfo(config.logger, gs, selectionLoc,
+                     "failed to determine whereToInsert in getExtractMultipleOccurrenceEdits");
+        return {};
+    }
```
jez, line 487: Is it worth making `findWhereToInsert` return an `optional<Loc>` to force users to explicitly dereference the result, reminding them that they also want to look at the result?

We have a few places in `Loc.h` that use `optional` specifically for this, like `source` and `fromDetails`.

main/lsp/ExtractVariable.cc
```diff
@@ -1,11 +1,19 @@
...
 
 using namespace std;
 
 namespace sorbet::realmain::lsp {
 
+void logDebugInfo(const std::shared_ptr<spdlog::logger> logger, const core::GlobalState &gs,
+                  const core::Loc selectionLoc, const std::string message) {
+    logger->debug("ExtractToVariable: {}", message);
+    logger->debug("selectionLoc=\"{}\"", selectionLoc.showRaw(gs));
+    logger->debug("source=\"{}\"", absl::CEscape(selectionLoc.file().data(gs).source()));
```
jez, line 14: You want these to show up at `error`, not `debug`.

main/lsp/ExtractVariable.cc
```diff
@@ -1,11 +1,19 @@
...
 #include "ast/treemap/treemap.h"
 #include "common/sort/sort.h"
 
 using namespace std;
 
 namespace sorbet::realmain::lsp {
 
+void logDebugInfo(const std::shared_ptr<spdlog::logger> logger, const core::GlobalState &gs,
+                  const core::Loc selectionLoc, const std::string message) {
+    logger->debug("ExtractToVariable: {}", message);
```
jez, line 12:
```suggestion
    logger->debug("msg=\"ExtractToVariable: {}\"", message);
```

main/lsp/ExtractVariable.cc
```diff
@@ -1,11 +1,19 @@
...
+    logger->debug("selectionLoc=\"{}\"", selectionLoc.showRaw(gs));
```
jez, line 13: Ideally we would put all the variables that aren't `source` on one line, makes it easier to manipulate them in splunk queries. the reason why i didn't do it for `source` in the other `Loc.cc` log lines is because i was worried about my log lines getting like sampled if they're too long.

jez (approved):
mostly looks good but i have some suggestions.

main/lsp/ExtractVariable.cc
neilparikh: I don't think it's worth it, so I'll leave it as is. In theory, this call shouldn't fail; if the code has gotten this far, it should complete successfully. There's also only 2 call sites, both in this file, so I don't expect any calls to accidentally use it without checking.

jez: > In theory, this call shouldn't fail

I made the Loc::source method return an optional specifically because "this should never crash, but in practice it's crashing a lot in ways that are tangential to normal execution, we shouldn't punish the user for our own faults"

#4075

so there is precedent for this, but it's ultimately up to you in this case.

PR #8062, opened 2024-07-31
Ignore magic method `<required-ancestor-lin>` during documentSymbols

KaanOzkan:
By design `requires_ancestor` keeps track of its usages by storing the loc of where it was defined. If requires_ancestor defined on Foo is used in thousands of files thorugh mixin of Foo then there would be thousands of `<required-ancestor-lin>` methods that point to the loc of Foo. This would mean documentSymbols will try to return all of those methods when `foo.rb` is opened which will crash the LSP.

### Motivation

In our case this looked something like:
```ruby
# foo.rb
module Foo
  requires_ancesetor { Object }
```

```ruby
# testcase.rb
class ActiveSupport::TestCase
  include Foo
end
```

With over 20k classes inheriting from `ActiveSupport::TestCase`.This meant seeing over 50k `<required-ancstor-lin` entries in [candidates](https://github.com/sorbet/sorbet/blob/5b4782ef14d493de57db52a314c5e1fb1a3018ec/main/lsp/requests/document_symbol.cc#L230).

One thing I couldn't understand was why `<required-ancestor-lin>` method isn't part of the method list if `Foo` was defined in the same file as its usages.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/lsp_helpers.cc
+++ b/main/lsp/lsp_helpers.cc
@@ -44,6 +44,11 @@ bool hideSymbol(const core::GlobalState &gs, core::SymbolRef sym) {
     if (name.kind() == core::NameKind::UNIQUE && name.dataUnique(gs)->original == core::Names::blockTemp()) {
         return true;
     }
+
+    if (name == core::Names::requiredAncestorsLin()) {
+        return true;
+    }
+
     return false;
 }
 


main/lsp/lsp_helpers.cc
```diff
@@ -44,6 +44,11 @@ bool hideSymbol(const core::GlobalState &gs, core::SymbolRef sym) {
     if (name.kind() == core::NameKind::UNIQUE && name.dataUnique(gs)->original == core::Names::blockTemp()) {
         return true;
     }
+
+    if (name == core::Names::requiredAncestorsLin()) {
+        return true;
+    }
```
KaanOzkan, line 50: `hideSymbol` is used in few more places then the problematic [codepath](https://github.com/sorbet/sorbet/blob/5b4782ef14d493de57db52a314c5e1fb1a3018ec/main/lsp/requests/document_symbol.cc#L217) but I think that's okay.

jez (approved):

main/lsp/lsp_helpers.cc
KaanOzkan: Screenshot 2024-08-01 at 11 47 24 AM

Offline chat: This is a synthetic method but it does take the user to where `requires_ancestor` is called so it might be a useful feature.

PR #8064, opened 2024-07-31
Add special case for `Class < Module`

jez:
### Motivation

This still doesn't handle the `Module < Module` case, because there's no
type that Sorbet can write there that's accurate until `T::Module`
exists.

### Test plan

See included tests

### DIFF
--- a/infer/environment.cc
+++ b/infer/environment.cc
@@ -680,17 +680,19 @@ void Environment::updateKnowledge(core::Context ctx, cfg::LocalRef local, core::
 
     if (send->fun == core::Names::lessThan() || send->fun == core::Names::leq()) {
         const auto &recvType = send->recv.type;
+        if (!recvType.derivesFrom(ctx, core::Symbols::Module())) {
+            return;
+        }
+
         const auto &argType = send->args[0].type;
 
         if (core::isa_type<core::ClassType>(argType)) {
             auto argClass = core::cast_type_nonnull<core::ClassType>(argType);
-            if (!recvType.derivesFrom(ctx, core::Symbols::Module()) ||
-                !argClass.symbol.data(ctx)->derivesFrom(ctx, core::Symbols::Class())) {
+            if (!argClass.symbol.data(ctx)->derivesFrom(ctx, core::Symbols::Class())) {
                 return;
             }
         } else if (auto *argClass = core::cast_type<core::AppliedType>(argType)) {
-            if (!recvType.derivesFrom(ctx, core::Symbols::Module()) ||
-                !argClass->klass.data(ctx)->derivesFrom(ctx, core::Symbols::Class())) {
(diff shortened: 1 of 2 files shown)


jez:

→ https://go/builds/bui_QZv2CE0nZxjUA2
→ https://go/builds/bui_QZv2XpKv1nfKoL
→ https://go/builds/bui_QZv2eZDNjmql89

jez:

→ https://go/builds/bui_QaD7uXQGx3sTqW
→ https://go/builds/bui_QaD73U0AymTRzz
→ https://go/builds/bui_QaD7tKyVMCatg5

jez:

→ https://go/builds/bui_QaJVBk0ve0Y9HP
→ https://go/builds/bui_QaJVEqvFMXv4wz
→ https://go/builds/bui_QaJVklPY5AlAcG

infer/environment.cc
```diff
@@ -679,28 +679,43 @@ void Environment::updateKnowledge(core::Context ctx, cfg::LocalRef local, core::
     }
 
     if (send->fun == core::Names::lessThan() || send->fun == core::Names::leq()) {
-        const auto &recvKlass = send->recv.type;
-        const auto &argType = send->args[0].type;
+        //
```
neilparikh, line 682: nit: delete this

test/testdata/infer/control_flow/module_less_than.rb
```diff
@@ -10,7 +10,114 @@ def self.foo(x)
...
+  end
+end
+
+sig { params(other: T.class_of(A)).void }
+def module_should_not_add_noTypeTest(other)
+  if other <= M
+    T.reveal_type(other) # error: `T.class_of(A)[T.all(A, M)]`
+  end
+
+  T.reveal_type(other) # error: `T.class_of(A)`
```
neilparikh, line 122: 2 questions:

Should this be in an `else` branch, since it's testing the case where the type test is false?

For my own understanding, I commented out the `canAddNoTypeTest = false;` on line 710 in `environment.cc`. After that, this line ended up saying `Revealed type: T.class_of(A)[T.all(A, M)]`, which was surprising to me. Why would adding a `noTypeTest` lead to the knowledge that `other` is a subtype of `M` outside the `if`?

jez: good catch. the code which inspired this test case in Stripe's codebase had a `return` in the `if` body

What happens is that adding a `noTypeTest` makes the `else` branch dead (or, if there were a `return` in the `if` branch: makes the rest of the method dead)

So when there is both a yes and a no type test, the else branch is implicitly dead, and when sorbet joins control flow after the `else`, it says that either `other` has type `T.noreturn` or it has the type it had inside the `if`, which was `T.all(A, M)`. Since unioning anything with `T.noreturn` returns the other thing, it had the effect of changing the type even outside the `if`.

It's more explicitly to do this with a `return` in the body, so I'll add that.

PR #8069, opened 2024-08-02
Delete `CompiledLevel`

neilparikh:
### Motivation

Compiler is deleted.

### Test plan

Existing tests + running over Stripe's codebase.


neilparikh:

→ https://go/builds/bui_QaPJ2l7F48hK2k
→ https://go/builds/bui_QaPKAlRcIsicro
→ https://go/builds/bui_QaPKt7BlfsI2Z9

test/testdata/rewriter/chalk_odm_document_compiled.rb
jez, line 1: It was intentional to keep the compiler tests when we deleted the compiler, they're still also tests of Sorbet itself.

proto/File.proto
jez, line 1: We should never delete fields from proto files, because proto is designed for compatibility.

class_flatten/class_flatten.cc
jez, line 1: Is there not more we can do here? It seems like replacement is always empty tree now, which suggests that we can keep folding code away.

core/proto/proto.cc
```diff
@@ -413,7 +402,6 @@ com::stripe::rubytyper::FileTable Proto::filesToProto(const GlobalState &gs,
         entry->set_sigil(strictToProto(file.data(gs).originalSigil));
         entry->set_strict(strictToProto(file.data(gs).strictLevel));
         entry->set_min_error_level(strictToProto(file.data(gs).minErrorLevel()));
-        entry->set_compiled(compiledToProto(file.data(gs).compiledLevel));
```
jez, line 416: IMO we should keep this line, and set it to None unconditionally. This is the default, but it's nice to be explicit.

jez (changes requested):
^

test/testdata/rewriter/chalk_odm_document_compiled.rb
neilparikh: I agree in general. In this case, I diffed the test and exp trees with and without the `_compiled` and they were identical (except for the `compiled: true` line), so the `_compiled` versions of the tests aren't actually testing anything new.

test/cli/file-sigils-json/test.sh
jez, line 1: I believe this test is the canonical test of the `--print file-table-json` output.

We should not delete this CLI test.

resolver/resolver.cc
```diff
@@ -3565,40 +3565,6 @@ class ResolveSignaturesWalk {
...
-        // T::Sig::WithoutRuntime.sig wouldn't have any runtime effect that we need
-        // to record later.
-        if (self->localVariable != core::LocalVariable::selfVariable()) {
-            return;
-        }
-
-        auto *cnst = ast::cast_tree<ast::ConstantLit>(send->recv);
-        ENFORCE(cnst != nullptr, "sig send receiver must be a ConstantLit if we got a ParsedSig from the send");
-
-        cnst->symbol = core::Symbols::Sorbet_Private_Static_ResolvedSig();
```
jez, line 3596: Does this mean we can delete `Sorbet_Private_Static_ResolvedSig` now?

jez:
One request to undelete a CLI test, and one suggestion that we might be able to delete more.

jez (approved):

neilparikh:

→ https://go/builds/bui_Qc9DLJ6QjOPcru
→ https://go/builds/bui_Qc9Dii5yVtrdqj
→ https://go/builds/bui_Qc9DMZebiz1iWh

PR #8075, opened 2024-08-03
Make more intersection types collapse

jez:
### Motivation

These types are smaller than the types we were using, which makes them more
useful. They're also quick to compute, as they only rely on a few more tag
checks for variables that are likely already in registers.

### Test plan

See included automated tests.


jez:

→ https://go/builds/bui_Qb1suuK8UzHkXS
→ https://go/builds/bui_Qb1s3akKFHifrr
→ https://go/builds/bui_Qb1s0RG3Flk1bN

ilyailya (approved):

PR #8086, opened 2024-08-07
Add Record to from Typescript guide in Sorbet

matv-stripe:
This adds a row to the docs which explains that `Record<string, Value>`  is `T::Hash[String, Value]` in Sorbet.

r? @jez-stripe


jez:
Closing in favor of something a little more explanatory

https://github.com/sorbet/sorbet/pull/8087

PR #8088, opened 2024-08-08
Add definition for URI::RFC2396_PARSER

RonRicardo:
### Motivation
URI added the ability to set [which parser it uses](https://github.com/ruby/uri/pull/107):

```
URI.parser = URI::RFC2396_PARSER
```

`URI::RFC2396_PARSER` is missing a type definition in `rbi/stdlib/uri.rbi`, which causes an error. This adds the missing type definition.

### Test plan
I didn't add tests for this change, it's a declaration and doesn't add any behavior.


froydnj (changes requested):
It looks like you used the wrong name for the constant?

https://buildkite.com/sorbet/sorbet/builds/32255#019132c1-d006-4fab-9da8-58bc38f1623a

RonRicardo:
@froydnj ack i totally did! changed it to the correct one, thanks.

froydnj (approved):
Thanks!

PR #8097, opened 2024-08-12
Clarify instance vs singleton class instance

jez:
### Motivation

Pointed out by a confused user.

### Test plan

See included automated tests.


neilparikh (commented):
Code looks fine to me, but you have test failures

neilparikh (approved):

PR #8101, opened 2024-08-14
Fix signature for `MatchData#offset`

Bo98:
### Motivation

Fixes the signature to allow string and symbol keys. These specify regex capture group names.

See the documentation just a couple lines up to for an example with a symbol is valid.

### Test plan

N/A. See Ruby documentation.

### DIFF
--- a/rbi/core/match_data.rbi
+++ b/rbi/core/match_data.rbi
@@ -289,7 +289,7 @@ class MatchData < Object
   # ```
   sig do
     params(
-        n: Integer,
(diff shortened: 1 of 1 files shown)


ilyailya:

→ https://go/builds/bui_Qf9cKezpFqbGfS
→ https://go/builds/bui_Qf9c2LJIANgHtS
→ https://go/builds/bui_Qf9c3PwJBIgZkL

ilyailya (approved):

PR #8103, opened 2024-08-15
Fix typing for SystemCallError and its subclasses

Bo98:
`SystemCallError` is a bit of a mess. It has a custom object creator that varies in behaviour if it called on a subclass, even if that subclass doesn't define its own constructor.

Basically the way it works is:

* `SystemCallError.new` takes 1-3 args
* `(class < SystemCallError).new` takes 0-2 args
* `SystemCallError.new` will return a `SystemCallError` object
* `SystemCallError.new(1)` will return a `Errno::EPERM` object
* `SystemCallError.new(999)` will return a `SystemCallError` object (no child class found)
* `SystemCallError#errno` will return a nilable integer based on what was passed to the constructor
* `(class < SystemCallError)#errno` will return the value of `(class < SystemCallError)::Errno`. It is an error to create a subclass without defining the `Errno` constant. The constant must be an integer or nil.
  - Therefore the stdlib `Errno::E.*#errno` can never return nil as they all have `Errno` correctly defined to an integer. 

We can't really check all of that in Sorbet so I did the best I could.

### Motivation

Original motivation was seeing that `rescue SystemCallError` was untyped: https://sorbet.run/#%23%20typed%3A%20strong%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bvoid%7D%0Adef%20bar%0A%20%20begin%0A%20%20rescue%20SystemCallError%20%3D%3E%20e%0A%20%20end%0Aend%0A

And then I opened a can of worms I perhaps shouldn't have of trying to type the rest of `SystemCallError`.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/errno.rbi
+++ b/rbi/core/errno.rbi
@@ -48,536 +48,1334 @@ end
 
 class Errno::E2BIG < SystemCallError
   Errno = T.let(nil, Integer)
+
+  sig { params(message: T.nilable(String), func: T.nilable(Object)).returns(T.attached_class) }
+  def self.new(message = nil, func = nil); end
+
+  sig { returns(Integer) }
+  def errno; end
 end
 
 class Errno::EACCES < SystemCallError
   Errno = T.let(nil, Integer)
+
+  sig { params(message: T.nilable(String), func: T.nilable(Object)).returns(T.attached_class) }
+  def self.new(message = nil, func = nil); end
+
+  sig { returns(Integer) }
+  def errno; end
 end
 
 class Errno::EADDRINUSE < SystemCallError
   Errno = T.let(nil, Integer)
+
+  sig { params(message: T.nilable(String), func: T.nilable(Object)).returns(T.attached_class) }
+  def self.new(message = nil, func = nil); end
+
+  sig { returns(Integer) }
+  def errno; end
 end
 
 class Errno::EADDRNOTAVAIL < SystemCallError
   Errno = T.let(nil, Integer)
+
+  sig { params(message: T.nilable(String), func: T.nilable(Object)).returns(T.attached_class) }
+  def self.new(message = nil, func = nil); end
+
+  sig { returns(Integer) }
+  def errno; end
 end
 
 class Errno::EADV < SystemCallError
   Errno = T.let(nil, Integer)
+
+  sig { params(message: T.nilable(String), func: T.nilable(Object)).returns(T.attached_class) }
+  def self.new(message = nil, func = nil); end
+
+  sig { returns(Integer) }
+  def errno; end
 end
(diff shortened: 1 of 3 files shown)


froydnj:

→ https://go/builds/bui_QfOfSsqY0UW9Jr
→ https://go/builds/bui_QfOfjFexceGlxa
→ https://go/builds/bui_QfOfSGqe5rPItr

froydnj (changes requested):
This PR turned up one error on Stripe's codebase:

```
test.rb:289: Expected T.nilable(String) but found {} for argument func http://go/e/7002
     289 |          Errno::ECONNRESET.new("Remote host reset the connection request", {}),
                                                                                      ^^
  Expected T.nilable(String) for argument func of method Errno::ECONNRESET.new:
    https://github.com/sorbet/sorbet/tree/fcf7e9a2bcdb65e486f2e1433160c37206f3619f/rbi/core/errno.rbi#L295:
     295 |  sig { params(message: T.nilable(String), func: T.nilable(String)).returns(T.attached_class) }
                                                     ^^^^
```

Which is valid, but I tend to think the error without the hash is better:

```
st-froydnj2:pay-server froydnj$ irb
irb(main):001:0> e1 = Errno::ECONNRESET.new("Remote host reset the connection request")
=> #<Errno::ECONNRESET: Connection reset by peer - Remote host reset the connection request>
irb(main):002:0> e2 = Errno::ECONNRESET.new("Remote host reset the connection request", {})
=> #<Errno::ECONNRESET: Connection reset by peer @ {} - Remote host reset the connection request>
```

WDYT about relaxing the type on `func`?  Or do you think we should just merge this as-is?

Bo98:
This is basically the "`to_s` problem".

```rb
class A
  def to_s
    "from to_s"
  end
end

Errno::ECONNRESET.new("abc", A.new)
=> #<Errno::ECONNRESET: Connection reset by peer @ from to_s - abc>

irb(main):001> Errno::ECONNRESET.new("abc", BasicObject.new)
(irb):1:in `initialize': undefined method `to_s' for an instance of BasicObject (NoMethodError)
```

What's the usual rule for those type of signatures?

FWIW, the argument is technically undocumented: https://docs.ruby-lang.org/en/3.3/SystemCallError.html.

> I tend to think the error without the hash is better

For the example you provide: I agree.

froydnj:
Ah, the `to_s` problem, yes.

I think `func` should probably be `Object`, then, or Sorbet will complain unnecessarily -- even if `String` would in some sense be the better choice.

Bo98:
Updated to use `Object`

froydnj (approved):
Thanks!

PR #8104, opened 2024-08-15
Correctly handle extracting to variable when all the occurrences are in the same "non-scope"

neilparikh:
```
if x + x
  1 + 1
end
```

Given previous code, extracting both occurrences of `x` would previously fail.

That's because the code to compute the LCA would (correctly) determine that it's the condition of the `if`. But that's not a "scope" we can insert - it's just a single `Send` expression - so `findWhereToInsert` would raise.

The fix I have here is to walk further up the LCA stack until we hit a ancestor we can insert into.

This has the downside of not correctly handling extracting both occurrences of `1`; we'll now insert it outside rather than inside the `if`. This happens becomes if the body of the `if` is just a single line, the AST represents it by pointing to the send directly rather than an `InsSeq`. This is the wrong behavior and we should fix it later, but I don't think it's worth blocking this change on it; it's a fairly minor case, and previously we just failed on it.

### Motivation

Handle this case correctly and don't error.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -473,7 +473,17 @@ MultipleOccurrenceResult VariableExtractor::getExtractMultipleOccurrenceEdits(co
         return {vector<unique_ptr<TextDocumentEdit>>(), 1};
     }
 
-    const ast::ExpressionPtr *scopeToInsertIn = walk.LCAScopeStack.back();
+    const ast::ExpressionPtr *scopeToInsertIn = nullptr;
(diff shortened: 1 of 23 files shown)


jez (approved):

jez:
> This is the wrong behavior and we should fix it later

should we start an "Extract Variable" milestone to start accumulating issues into?

jez:
https://github.com/sorbet/sorbet/milestones

we use milestones as components (when it would probably be better to use labels) but ¯\_(ツ)_/¯

PR #8105, opened 2024-08-15
Correctly handle &&/|| in extract to variable

neilparikh:
We desugar `a && b` to the following:
```
if a
  b
else
  a
end
```

The problem is that this has 2 instances of `a`, so we will incorrectly suggest the extract all occurrences action, which will then insert `newVariable` twice.

We can handle this by skipping any nodes that have the same loc as a node we've previously seen and stored as a match; if the locs are the same, one of them is likely a synthetic node we created in desugar.

### Motivation

Handle this case correctly and don't error.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -399,6 +399,12 @@ public:
             return;
         }
 
+        if (std::find(matches.begin(), matches.end(), tree.loc()) != matches.end()) {
+            // We've already seen a node with the exact same loc before, so this is likely constructed by
+            // desugar. Skip it.
+            return;
+        }
+
         if (targetNode->structurallyEqual(ctx, tree, selectionLoc.file())) {
             matches.emplace_back(tree.loc());
             computeLCA(tree.loc());
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -399,7 +399,7 @@ public:
             return;
         }
 
-        if (std::find(matches.begin(), matches.end(), tree.loc()) != matches.end()) {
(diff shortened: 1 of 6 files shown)


test/testdata/lsp/code_actions/extract_variable_single/insseq.B.rbedited
```diff
@@ -0,0 +1,19 @@
+# typed: true
+# selective-apply-code-action: refactor.extract
+# enable-experimental-lsp-extract-to-variable: true
+
+def and
+  a = T.unsafe(1)
+  b = T.unsafe(1)
+  a && newVariable = b; newVariable
```
neilparikh, line 8: We should instead produce the following here:
```
newVariable = b
a && newVariable
```
I'll fix this in a future PR.

main/lsp/ExtractVariable.cc
```diff
@@ -399,6 +399,12 @@ class ExpressionPtrSearchWalk {
             return;
         }
 
+        if (std::find(matches.begin(), matches.end(), tree.loc()) != matches.end()) {
```
jez, line 402: You can simplify this if you want:

```suggestion
        if (absl::c_find(matches, tree.loc()) != matches.end()) {
```

And in the newest version of abseil it simplifies even further to just `absl::c_contains(matches, tree.loc())`

https://github.com/abseil/abseil-cpp/commit/63d4b2fe1e0d4bd6a7b916f398643db40c35624b

I've opened a draft PR to try to see if we can trivially upgrade #8138 but do not block on this—just passing feedback.

jez (approved):

PR #8108, opened 2024-08-15
Update signatures for thread interrupt methods

Bo98:
Updates signatures for `Thread.*_interrupt` methods and `Thread#raise`.

### Motivation

Noticed most of these are untyped.

### Test plan

See included automated tests.

### DIFF
--- a/rbi/core/thread.rbi
+++ b/rbi/core/thread.rbi
@@ -522,8 +522,8 @@ class Thread < Object
   # See
   # [`::pending_interrupt?`](https://docs.ruby-lang.org/en/2.7.0/Thread.html#method-c-pending_interrupt-3F)
   # for more information.
-  sig {params(args: T.untyped).returns(T::Boolean)}
-  def pending_interrupt?(*args); end
+  sig {params(error: T::Class[T.anything]).returns(T::Boolean)}
+  def pending_interrupt?(error = nil); end
 
   # Returns the priority of *thr*. Default is inherited from the current thread
   # which creating the new thread, or zero for the initial main thread;
@@ -583,7 +583,22 @@ class Thread < Object
   #  from prog.rb:2:in `new'
   #  from prog.rb:2
   # ```
-  def raise(*_); end
+  sig {void}
+  sig do
+    params(
+      arg0: T.any(T::Class[Exception], Exception, String),
+    )
+    .void
+  end
+  sig do
+    params(
+      arg0: T.any(T::Class[Exception], Exception),
+      arg1: T.untyped,
+      arg2: T.nilable(T::Array[String]),
+    )
+    .void
+  end
+  def raise(arg0=nil, arg1=nil, arg2=nil); end
 
   # Returns the status of the thread-local "report on exception" condition for
   # this `thr`.
@@ -1015,7 +1030,13 @@ class Thread < Object
   #   # all exceptions inherited from Exception are prohibited.
   # }
   # ```
-  sig {params(hash: T.untyped, block: T.proc.returns(T.untyped)).returns(T.untyped)}
+  sig do
+    type_parameters(:U).params(
+      hash: T::Hash[T.class_of(Exception), Symbol],
+      block: T.proc.returns(T.type_parameter(:U))
+    )
+    .returns(T.type_parameter(:U))
+  end
   def self.handle_interrupt(hash, &block); end
 
   # Causes the given `thread` to exit, see also
@@ -1118,8 +1139,8 @@ class Thread < Object
   # ...
   # flag = false # stop thread
   # ```
-  sig {params(args: T.untyped).returns(T::Boolean)}
-  def self.pending_interrupt?(*args); end
+  sig {params(error: T.class_of(Exception)).returns(T::Boolean)}
+  def self.pending_interrupt?(error); end
 
   # Returns the status of the global "report on exception" condition.
   #
--- a/test/testdata/rbi/thread.rb
+++ b/test/testdata/rbi/thread.rb
@@ -1,0 +1,11 @@
+# typed: strong
+
+t = Thread.new do
+  Thread.handle_interrupt(StandardError => :never) do
+    puts Thread.pending_interrupt?(StandardError)
+  end
+end
+t.raise StandardError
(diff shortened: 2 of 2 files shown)


rbi/core/thread.rbi
```diff
@@ -583,7 +583,22 @@ class Thread < Object
...
+  end
+  sig do
+    params(
+      arg0: T.any(T::Class[Exception], Exception),
+      arg1: T.untyped,
+      arg2: T.nilable(T::Array[String]),
+    )
+    .void
+  end
+  def raise(arg0=nil, arg1=nil, arg2=nil); end
```
jez, line 601: Is there any reason for `Thread#raise` and `Kernel#raise` to have different signatures?

I notice that

- `Kernel#raise` returns `T.noreturn`
- `Kernel#raise` accepts `T::Class[T.anything]`, which does not look right given this:

```
irb(main):001> class A; end
=> nil
irb(main):002> raise A
(irb):2:in `raise': exception class/object expected (TypeError)
        from (irb):2:in `<main>'
        from /Users/jez/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/gems/irb-1.14.0/exe/irb:9:in `<top (required)>'
        from /Users/jez/.rbenv/versions/3.1.4/bin/irb:25:in `load'
        from /Users/jez/.rbenv/versions/3.1.4/bin/irb:25:in `<main>'
```

rbi/core/thread.rbi
```diff
@@ -1015,7 +1030,13 @@ class Thread < Object
   #   # all exceptions inherited from Exception are prohibited.
   # }
   # ```
-  sig {params(hash: T.untyped, block: T.proc.returns(T.untyped)).returns(T.untyped)}
+  sig do
+    type_parameters(:U).params(
+      hash: T::Hash[T::Class[T.anything], Symbol],
```
jez, line 1035: Should this be `T.class_of(Exception)` instead of `T::Class[T.anything]`? What would it mean to have a class here that's not an exception?

rbi/core/thread.rbi
```diff
@@ -1118,8 +1139,8 @@ class Thread < Object
   # ...
   # flag = false # stop thread
   # ```
-  sig {params(args: T.untyped).returns(T::Boolean)}
-  def self.pending_interrupt?(*args); end
+  sig {params(error: T::Class[T.anything]).returns(T::Boolean)}
+  def self.pending_interrupt?(error); end
```
jez, line 1143: Should `error` be optional? It seems like that might be the case from the documentation comment but I have not checked this against the upstream docs.

rbi/core/thread.rbi
Bo98: I suspect `Kernel#raise` signature was from before `T::Class` existed.

The return value is intentionally different. `Thread#raise` does return.

jez: ah makes sense, thanks!

rbi/core/thread.rbi
Bo98: Ruby doesn't validate it and accepts anything here but I can't think of any scenario where it would do anything. Should I switch it to `T.class_of(Exception)`?

rbi/core/thread.rbi
Bo98: Yes, that's my bad - will fix.

rbi/core/thread.rbi
jez: I would probably prefer `T.class_of(Exception)` to alert someone that they're probably doing something wrong.

Bo98: Just tested this code and it does work:


```
t = Thread.new do
  Thread.handle_interrupt(Object => :never) do
    sleep(5)
  end
end
```

It's perhaps odd but is functional.

jez: This also works at runtime:

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28opts%3A%20T%3A%3AHash%5BSymbol%2C%20Integer%5D%29.void%20%7D%0Adef%20example%28opts%29%0A%20%20opts%5B%22not%20a%20symobl%22%5D%0Aend)

```ruby
# typed: true
extend T::Sig

sig { params(opts: T::Hash[Symbol, Integer]).void }
def example(opts)
  opts["not a symobl"]
end
```

but despite the fact that no runtime exception is raised, Sorbet reports a type error because chances are that the user is confused.

I don't care that much about this particular signature and if you'd like we can keep it at `T::Class[T.anything]`, but I think that in general we use types to prevent people from accidentally making mistakes, not necessarily to allow everything that the runtime allows.

Bo98: Cool, sounds good to me. Updated now.

jez:

→ https://go/builds/bui_QfXtd7vF5lAJQv
→ https://go/builds/bui_QfXtNvWi3OXqg5
→ https://go/builds/bui_QfXtSwI9pvZeNS

rbi/core/thread.rbi
```diff
@@ -522,8 +522,8 @@ class Thread < Object
   # See
   # [`::pending_interrupt?`](https://docs.ruby-lang.org/en/2.7.0/Thread.html#method-c-pending_interrupt-3F)
   # for more information.
-  sig {params(args: T.untyped).returns(T::Boolean)}
-  def pending_interrupt?(*args); end
+  sig {params(error: T::Class[T.anything]).returns(T::Boolean)}
+  def pending_interrupt?(error = nil); end
```
Slike9, line 526: Shouldn't `error` be of `T.nilable(T::Class[T.anything])`?

Bo98: If you're doing an explicit `Thread#pending_interrupt?(nil)`, sure.

If you're omitting the argument entirely it should work as is.

Bo98: Ah didn't see the linked issue initially. The actual issue is not this one but the `self.pending_interrupt?` further down I forgot to update. Was a Cmd+F failure by myself as I didn't realise there were two of them. Will fix.

PR #8109, opened 2024-08-15
Change the loc we use for mlhs pinning errors

jez:
### Motivation

When there's a pinning error for a variable involved in a multiple assignment
statement, we currently show the error highlight as the entire multiple
assignment expression, which might span many lines, and which doesn't indicate
which of the variables in the assignment has a pinning error.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/desugar/pin_mlhs.rb
+++ b/test/testdata/desugar/pin_mlhs.rb
@@ -11,6 +11,6 @@ def mlhs_example
   x, _ = tuple_str_int
   2.times do
     y, _ = tuple_int_str
-  # ^^^^^^^^^^^^^^^^^^^^ error: Changing the type of a variable is not permitted
+    #  ^ error: Changing the type of a variable is not permitted
   end
 end
--- a/test/testdata/desugar/pin_mlhs.rb
+++ b/test/testdata/desugar/pin_mlhs.rb
@@ -1,0 +1,16 @@
+# typed: true
+extend T::Sig
+
+sig { returns([Integer, String]) }
+def tuple_int_str = [0, '']
+
+sig { returns([String, Integer]) }
+def tuple_str_int = ['', 0]
+
+def mlhs_example
+  x, _ = tuple_str_int
+  2.times do
+    y, _ = tuple_int_str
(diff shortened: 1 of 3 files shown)


test/testdata/desugar/pin_mlhs.rb
```diff
@@ -0,0 +1,16 @@
...
+def tuple_int_str = [0, '']
+
+sig { returns([String, Integer]) }
+def tuple_str_int = ['', 0]
+
+def mlhs_example
+  x, _ = tuple_str_int
+  2.times do
+    y, _ = tuple_int_str
+    #  ^ error: Changing the type of a variable is not permitted
```
froydnj, line 14: This doesn't seem like the right error: we should be creating a new variable here, shouldn't we?  We would if the variable was actually named, right?

I can believe that the locs are better with the change, but I am a bit skeptical that this is the right way to demonstrate it.  But I am also on my phone and might not be thinking clearly.

froydnj (changes requested):
Requesting changes just to get the question below cleared up.

test/testdata/desugar/pin_mlhs.rb
jez: The `_` is not special in Ruby in unless you're using pattern matching syntax, which this is not. In this situation, `_` is just a variable name, like `x` or `y`. In this case, the `_` variable had type `Integer` outside the loop and has type `String` after this line runs inside the loop, and that's enough for Ruby to not like this code.

We could separately enforce that variables that start with underscores cannot be used (or maybe just that the `_` variable cannot be used) but that's tangential.

We could possibly change the logic which computes blocks arguments to say that if a variable is written to before it's read to in a block that it doesn't need to be in the basic block's argument list and then maybe doesn't participate in pinning, but that's a separate issue—the issue at hand is which location we use to report the pinning error that's happening.

froydnj: Sorry, you're right.  I think I was thinking more in Rust terms than in Ruby terms, as though the multi assignment somehow declared an entirely new set of variables.

froydnj (approved):
I think this makes sense, but you have some test failures?

PR #8113, opened 2024-08-19
Create signatures for Process.warmup

peterzhu2118:
Process.warmup was introduced in Ruby 3.3.

Docs can be found here: https://docs.ruby-lang.org/en/master/Process.html#method-c-warmup


froydnj (approved):
Thanks!

PR #8119, opened 2024-08-20
Fix signature for `Thread.pending_interrupt?`

### LINKED ISSUE #8118: Not enough arguments provided for method Thread.pending_interrupt?. Expected: 1, got: 0
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0AThread.pending_interrupt%3F)

```ruby
# typed: true
extend T::Sig

Thread.pending_interrupt?
```

#### Observed output

```

# typed: true
extend T::Sig

Thread.pending_interrupt?


editor.rb:4: Not enough arguments provided for method Thread.pending_interrupt?. Expected: 1, got: 0 https://srb.help/7004
     4 |Thread.pending_interrupt?
                                 ^
    https://github.com/sorbet/sorbet/tree/master/rbi/core/thread.rbi#L1143: Thread.pending_interrupt? defined here
    1143 |  def self.pending_interrupt?(error); end
Errors: 1
```

#### Expected behavior

No error.

The parameter is optional: https://ruby-doc.org/core-2.7.4/Thread.html#method-c-pending_interrupt-3F.

- - -

May relate to https://github.com/sorbet/sorbet/pull/8108. The parameter is defined as non-nilable [there](https://github.com/sorbet/sorbet/pull/8108/files#diff-88faeaef4cca1e1fcebb8ebfc4932b814147d5c5cc16f161150db21fffab04c7R1143):
```ruby
  sig {params(error: T.class_of(Exception)).returns(T::Boolean)}
  def self.pending_interrupt?(error); end
```

Bo98:
### Motivation

I fixed `Thread#pending_interrupt?` instance method but I failed to notice that there's _also_ a `Thread.pending_interrupt?` class method.

Fix the class method to match the instance method signature.

Fixes #8118

### Test plan

#8118

### DIFF
--- a/rbi/core/thread.rbi
+++ b/rbi/core/thread.rbi
@@ -1140,7 +1140,7 @@ class Thread < Object
   # flag = false # stop thread
   # ```
   sig {params(error: T.class_of(Exception)).returns(T::Boolean)}
-  def self.pending_interrupt?(error); end
+  def self.pending_interrupt?(error = nil); end
 
   # Returns the status of the global "report on exception" condition.
   #


rbi/core/thread.rbi
```diff
@@ -1140,7 +1140,7 @@ class Thread < Object
   # flag = false # stop thread
   # ```
   sig {params(error: T.class_of(Exception)).returns(T::Boolean)}
-  def self.pending_interrupt?(error); end
+  def self.pending_interrupt?(error = nil); end
```
Slike9, line 1143: What about `T.unsafe(nil)` instead of plain `nil`?

jez: In general I don't really care that much about default values in RBIs for the standard library, they tend to play it pretty fast and loose with typing and also RBI files' default values are not typechecked so it doesn't matter.

Another option is that I believe omitting the `error` argument is akin to passing an error argument of `Exception` (`Thread.pending_interrupt?(Exception)`) so you could use that as the default value if you wanted something which would pass type checks in an alternative universe where RBIs are typechecked.

jez (approved):

PR #8124, opened 2024-08-21
Fix crash in structurallyEqual

neilparikh:
The `original` field in `ConstantLit` is `nullptr` for synthetic `ConstantLit`s.

### Motivation

Handle this case properly.

### Test plan

See included automated tests.

### DIFF
--- a/ast/TreeEquality.cc
+++ b/ast/TreeEquality.cc
@@ -275,7 +275,16 @@ bool structurallyEqual(const void *avoid, const Tag tag, const void *tree, const
         case Tag::ConstantLit: {
             auto *a = reinterpret_cast<const ConstantLit *>(tree);
             auto *b = reinterpret_cast<const ConstantLit *>(other);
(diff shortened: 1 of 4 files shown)


ast/TreeEquality.cc
```diff
@@ -275,7 +275,16 @@ bool structurallyEqual(const void *avoid, const Tag tag, const void *tree, const
         case Tag::ConstantLit: {
             auto *a = reinterpret_cast<const ConstantLit *>(tree);
             auto *b = reinterpret_cast<const ConstantLit *>(other);
-            return a->symbol == b->symbol && structurallyEqual(avoid, a->original, b->original);
+            if (a->symbol != b->symbol) {
+                return false;
+            }
+            if (a->original && b->original) {
```
jez, line 281: Might be worth a comment that we expect this to be the case whenever we used `MK::Constant` instead of `MK::UnresolvedConstant`.

jez (approved):

PR #8125, opened 2024-08-21
Add logging to structurallyEqual

neilparikh:
Review commit by commit.

### Motivation

Make it easier catch future crashes here.

### Test plan

Commented out the `ENFORCE`s, and then manually triggered the known error case and saw the relevant logs in the log file.

### DIFF
--- a/ast/TreeEquality.cc
+++ b/ast/TreeEquality.cc
@@ -7,16 +7,19 @@ namespace sorbet::ast {
 
 namespace {
 
-bool structurallyEqual(const void *avoid, const ExpressionPtr &tree, const ExpressionPtr &other, bool root = false);
+bool structurallyEqual(const core::GlobalState &gs, const void *avoid, const ExpressionPtr &tree,
+                       const ExpressionPtr &other, const std::shared_ptr<spdlog::logger> logger,
+                       const core::Loc selectionLoc, bool root = false);
 
 template <unsigned long N>
-bool structurallyEqualVec(const void *avoid, const InlinedVector<ExpressionPtr, N> &a,
-                          const InlinedVector<ExpressionPtr, N> &b) {
+bool structurallyEqualVec(const core::GlobalState &gs, const void *avoid, const InlinedVector<ExpressionPtr, N> &a,
+                          const InlinedVector<ExpressionPtr, N> &b, const std::shared_ptr<spdlog::logger> logger,
+                          const core::Loc selectionLoc) {
     if (a.size() != b.size()) {
         return false;
     }
     for (int i = 0; i < a.size(); i++) {
-        if (!structurallyEqual(avoid, a[i], b[i])) {
+        if (!structurallyEqual(gs, avoid, a[i], b[i], logger, selectionLoc)) {
             return false;
         }
     }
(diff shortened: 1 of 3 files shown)


ast/TreeEquality.cc
```diff
@@ -7,16 +8,19 @@ namespace sorbet::ast {
 
 namespace {
 
-bool structurallyEqual(const void *avoid, const ExpressionPtr &tree, const ExpressionPtr &other, bool root = false);
+bool structurallyEqual(const core::GlobalState &gs, const void *avoid, const ExpressionPtr &tree,
+                       const ExpressionPtr &other, const std::shared_ptr<spdlog::logger> logger,
```
jez, line 12: Can we take the logger by const reference instead?

As it is right now, this requires atomic incr/decr operations for every recursive call to `structurallyEqual` to do the reference counting.

ast/TreeEquality.cc
```diff
@@ -302,39 +317,56 @@ bool structurallyEqual(const void *avoid, const Tag tag, const void *tree, const
     }
 }
 
-bool structurallyEqual(const void *avoid, const ExpressionPtr &tree, const ExpressionPtr &other, bool root) {
-    ENFORCE(tree != nullptr);
-    ENFORCE(tree != nullptr);
+bool structurallyEqual(const core::GlobalState &gs, const void *avoid, const ExpressionPtr &tree,
+                       const ExpressionPtr &other, const std::shared_ptr<spdlog::logger> logger,
+                       const core::Loc selectionLoc, bool root) {
```
jez, line 322: selectionLoc is a bit leaky—it implies that you'd only ever want to call this method from LSP, when this method is implemented as a non-LSP helper.

Do we actually need the selectionLoc? I was thinking that if we had the file path & contents, it would not be hard to drop into a debugger and find which parts of a file had a `nullptr` tree in them somewhere.

We could also probably reconstitute the selectionLoc from the other log lines that would appear before this log line, e.g. the log lines for every LSP request/response.

ast/TreeEquality.cc
```diff
@@ -302,39 +317,56 @@ bool structurallyEqual(const void *avoid, const Tag tag, const void *tree, const
...
+        logger->error("msg=\"ExtractToVariable: tree is null in structurallyEqual\" selectionLoc=\"{}\"",
+                      selectionLoc.showRaw(gs));
+    }
+    if (!other) {
+        logger->error("msg=\"ExtractToVariable: other is null in structurallyEqual\" selectionLoc=\"{}\"",
+                      selectionLoc.showRaw(gs));
+    }
+    if (!tree || !other) {
+        logger->error("source=\"{}\"", absl::CEscape(selectionLoc.file().data(gs).source()));
+    }
```
jez, line 335: Do you want to `return false` here?

That way at least the user won't have Sorbet crash on account of our mistakes.

ast/TreeEquality.cc
jez: Even more than that, if you don't want to deal with threading the `logger` through all of these calls, you can use the `fatalLogger` global variable, like we do in `Loc.cc`.

jez (changes requested):
some requests.

jez (approved):
Looks great thanks!

PR #8130, opened 2024-08-22
Handle extracting the receiver of CSend nodes correctly

neilparikh:
We can simply look for a `send` with `fun = core::Names::nilForSafeNavigation()` to check for a `csend` node; it's used nowhere else.

### Motivation

Extract the receiver of a `csend` to the correct place.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -78,7 +78,7 @@ class LocSearchWalk {
     // the current top of the stack as the "deepest" class/method
     vector<ast::ExpressionPtr *> enclosingClassStack;
     vector<ast::ExpressionPtr *> enclosingMethodStack;
-    bool inCsend = false;
+    int inCsend = 0;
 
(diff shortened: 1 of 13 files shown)


main/lsp/ExtractVariable.cc
```diff
@@ -93,6 +78,7 @@ class LocSearchWalk {
     // the current top of the stack as the "deepest" class/method
     vector<ast::ExpressionPtr *> enclosingClassStack;
     vector<ast::ExpressionPtr *> enclosingMethodStack;
+    bool inCsend = false;
```
jez, line 81: Usually keeping track of a boolean in these sorts of things falls over if there's ever some sort of nesting.

Does the change here handle cases like

```ruby
(a&.foo)&.bar
```

?

neilparikh: Good catch, I updated it to use a counter instead, and added a few test cases for it.

jez (approved):

PR #8131, opened 2024-08-22
website: Documentation Comments

jez:
### Motivation

Want people to understand that this feature exists.


website/docs/doc-comments.md
```diff
@@ -0,0 +1,134 @@
+---
+id: doc-comments
+title: Documentation Comments
+---
+
+Sorbet language server has rudimentary support for displaying documentation
```
froydnj, line 6:
```suggestion
Sorbet's language server has rudimentary support for displaying documentation
```

website/docs/doc-comments.md
```diff
@@ -0,0 +1,134 @@
...
+# -- file.rbi --
+
+# Documentation for A, comment 2
+class A; end
+```
+
+**Note**: For performance, Sorbet only stores one location of a definition per
+file. If something is defined multiple times within the same file, the last
+location generally wins (which might prevent Sorbet from finding the
+documentation for that definition).
```
froydnj, line 134: Nit: at first read, "that definition" make me think the referent was "the last location", rather than the (I think) intended "something [that] is defined multiple times".  Maybe "which might prvent Sorbet from finding the documentation for that entity?" to disambiguate?

froydnj (approved):

website/docs/doc-comments.md
```diff
@@ -0,0 +1,134 @@
...
+  def self.greet(name)
+    puts("Hello, #{name}!")
+  end
+```
+
+Sorbet can generate a YARD snippet with `@param` and `@return` attributes for a
+method, so you don't have to type them all manually:
+
+<video autoplay loop muted playsinline style="max-width: calc(min(813px, 100%));">
+  <source src="/img/lsp/yard-snippet.mp4" type="video/mp4">
```
Morriar, line 85: Should the video be committed in this PR? It doesn't seem to be already in the repo: https://github.com/sorbet/sorbet/tree/master/website/static/img

Morriar: Ah I looked at the wrong directory, it's under `lsp/` 👍

jez: The video is reused from https://sorbet.org/docs/autocompletion and is present in the `lsp/` subfolder of the folder you linked.

PR #8138, opened 2024-08-23
Upgrade abseil-cpp to 20240722.0

jez:
### Motivation

I noticed that this version includes `absl::c_contains` which is a method I've wanted to have for a while.

### Test plan

Existing tests

### DIFF
--- a/main/lsp/LSPConfiguration.cc
+++ b/main/lsp/LSPConfiguration.cc
@@ -35,7 +35,7 @@ string getRootPath(const shared_ptr<LSPOutput> &output, const options::Options &
 }
 
 MarkupKind getPreferredMarkupKind(vector<MarkupKind> formats) {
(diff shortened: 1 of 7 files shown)


main/lsp/LSPConfiguration.cc
```diff
@@ -162,7 +162,9 @@ string urlDecode(string_view uri) {
         auto from = uri.substr(pos, 3);
         // add replacement only if % is actually followed by exactly 2 hex digits
         if (from.size() == 3 && isxdigit(from[1]) && isxdigit(from[2])) {
-            auto to = absl::HexStringToBytes(from.substr(1));
+            string to;
+            auto valid = absl::HexStringToBytes(from.substr(1), &to);
+            ENFORCE(valid, "We checked it was valid above");
```
jez, line 167: They deprecated the version of the function we were using, which got converted to an error with `-Werror`.

ilyailya (approved):

PR #8139, opened 2024-08-23
Move extract to variable to beta

neilparikh:
### Motivation

Expose the feature more broadly.

### Test plan

N/A


main/options/options.cc
```diff
@@ -369,7 +369,7 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                     "Enable experimental `requires_ancestor` annotation");
 
     options.add_options("advanced")("enable-experimental-lsp-extract-to-variable",
-                                    "Enable experimental LSP feature: Extract To Variable");
+                                    "Enable beta LSP feature: Extract To Variable");
```
jez, line 372: we typically leave these descriptions saying "experimental" because that's what's in the option name, but it's also fine to change the description like you've done.

jez (approved):

PR #8140, opened 2024-08-24
Desugar csend nodes specially when called from extract to variable

neilparikh:
Review commit-by-commit.

`a&.foo` will desugar to `a.desugar_csend_foo`, where `desugar_csend_foo` is a special name kind to signify this actually represents a `CSend`.

### Motivation

We desugar `a&.foo` to `begin; temp = a; if nil === temp then nil else temp.foo end end` currently. This causes problems for extract to variable, since we'll try to insert inside the fake `InsSeq`. I had a workaround in #8130, but that doesn't handle all cases.

Instead, we can just not do this when `desugar` is called from extract to variable. We can desugar it to regular send instead, but with a special name so that it's treated different from an actual send to that method.

### Test plan

See included automated tests.

### DIFF
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1472,6 +1472,14 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what, 
                 }
             },
             [&](parser::CSend *csend) {
+                if (calledFromExtractToVariable) {
+                    auto newName = dctx.ctx.state.freshNameUnique(core::UniqueNameKind::DesugarCsend, csend->method, 1);
+                    auto sendNode = make_unique<parser::Send>(loc, std::move(csend->receiver), newName,
+                                                              csend->methodLoc, std::move(csend->args));
+                    auto send = node2TreeImpl(dctx, std::move(sendNode), calledFromExtractToVariable);
+                    result = std::move(send);
+                    return;
+                }
                 core::NameRef tempRecv = dctx.freshNameUnique(core::Names::assignTemp());
                 auto recvLoc = csend->receiver->loc;
                 // Assign some desugar-produced nodes with zero-length Locs so IDE ignores them when mapping text
--- a/ast/desugar/Desugar.cc
+++ b/ast/desugar/Desugar.cc
@@ -1473,6 +1473,12 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what, 
             },
             [&](parser::CSend *csend) {
                 if (calledFromExtractToVariable) {
+                    // Desugaring to a InsSeq + If causes problem for Extract to Variable; the fake If will be the where
+                    // the new variable is inserted, which is incorrect. Instead, desugar to a regular send, so that the
+                    // insertion happens in the correct place (what the csend is inside);
+
+                    // Replace the original method with a new special one that conveys that this is a CSend, so that
+                    // a&.foo is treated as different from a.foo when checking for structural equality.
                     auto newName = dctx.ctx.state.freshNameUnique(core::UniqueNameKind::DesugarCsend, csend->method, 1);
                     auto sendNode = make_unique<parser::Send>(loc, std::move(csend->receiver), newName,
                                                               csend->methodLoc, std::move(csend->args));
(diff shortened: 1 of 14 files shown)


ast/desugar/Desugar.cc
```diff
@@ -84,10 +84,11 @@ ExpressionPtr numparamTree(DesugarContext dctx, int num, parser::NodeVec *decls)
     return MK::Local(core::LocOffsets::none(), name);
 }
 
-ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what);
+ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what, bool calledFromExtractToVariable);
```
jez, line 87: We should put this in the `DesugarContext` which is already threaded to all of these calls.

neilparikh:
Ran sorbet over Stripe's codebase, and the run time was roughly the same before and after:

Before:
```
(ins)$ pay exec rm -rf /tmp/sorbet-cache && pay exec time scripts/bin/typecheck
No errors! Great job.
556.61user 26.27system 0:54.09elapsed 1077%CPU (0avgtext+0avgdata 12730440maxresident)k
810960inputs+24outputs (13major+4696660minor)pagefaults 0swaps
(ins)$ pay exec rm -rf /tmp/sorbet-cache && pay exec time scripts/bin/typecheck
No errors! Great job.
558.31user 23.99system 0:53.95elapsed 1079%CPU (0avgtext+0avgdata 12782880maxresident)k
9120inputs+16outputs (7major+4683969minor)pagefaults 0swaps
```

After:
```
(ins)$ pay exec rm -rf /tmp/sorbet-cache && pay exec time scripts/bin/typecheck
No errors! Great job.
559.52user 24.81system 0:53.65elapsed 1089%CPU (0avgtext+0avgdata 12680704maxresident)k
1840inputs+24outputs (4major+4773410minor)pagefaults 0swaps
(ins)$ pay exec rm -rf /tmp/sorbet-cache && pay exec time scripts/bin/typecheck
No errors! Great job.
558.53user 24.89system 0:53.89elapsed 1082%CPU (0avgtext+0avgdata 12747812maxresident)k
16648inputs+16outputs (0major+4767171minor)pagefaults 0swaps
```

ast/desugar/Desugar.cc
```diff
@@ -26,13 +26,14 @@ struct DesugarContext final {
     core::NameRef enclosingMethodName;
     bool inAnyBlock;
     bool inModule;
+    bool calledFromExtractToVariable;
```
jez, line 29: I'm generally -1 on naming things "calledFromX"

The option should describe what behavior is being done. So like "preserveConcreteSyntax" or something like that.

ast/desugar/Desugar.h
```diff
@@ -7,7 +7,8 @@
 
 namespace sorbet::ast::desugar {
 
-ExpressionPtr node2Tree(core::MutableContext ctx, std::unique_ptr<parser::Node> what);
+ExpressionPtr node2Tree(core::MutableContext ctx, std::unique_ptr<parser::Node> what,
+                        bool calledFromExtractToVariable = false);
```
jez, line 11: when we rename the variable let's be sure to change it here too

core/Names.cc
```diff
@@ -78,6 +78,9 @@ string NameRef::showRaw(const GlobalState &gs) const {
                 case UniqueNameKind::Packager:
                     kind = "G";
                     break;
+                case UniqueNameKind::DesugarCsend:
+                    kind = "&";
```
jez, line 82: neat trick, I hadn't though to use punctuation in this.

main/lsp/LSPTypechecker.cc
```diff
@@ -687,7 +687,7 @@ std::vector<std::unique_ptr<core::Error>> LSPTypechecker::retypecheck(vector<cor
 }
 
 ast::ExpressionPtr LSPTypechecker::getLocalVarTrees(core::FileRef fref) const {
-    auto afterDesugar = pipeline::desugarOne(config->opts, *gs, fref);
+    auto afterDesugar = pipeline::desugarOne(config->opts, *gs, fref, true);
```
jez, line 690: Preference for naming or commenting this `true` literal.

jez (commented):
mostly looks good, my comments are mostly around naming / clarity.

jez (approved):

PR #8141, opened 2024-08-26
Allow Struct with no args which works today on modern Ruby versions

HParker:
irb(main):001> Struct.new
    => #<Class:0x0000000122927958>

### Motivation

Allow creating `Struct.new` without arguments.

### Test plan

Please advise. I am not sure what is standard practice for testing RBI files.

### DIFF
--- a/rbi/core/struct.rbi
+++ b/rbi/core/struct.rbi
@@ -36,14 +36,14 @@ class Struct < Object
 
   sig do
     params(
-        arg0: T.any(Symbol, String),
+        arg0: T.nilable(T.any(Symbol, String)),
         arg1: T.any(Symbol, String),
         keyword_init: T::Boolean,
         blk: T.untyped,
     )
     .void
   end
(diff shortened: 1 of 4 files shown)


jez (approved):

jez:
thanks!

jez:
you've got some test failures. you should be able to edit or delete some `# error:` comments in the mentioned test that fails to fix the errors.

HParker:
I ran `tools/scripts/update_testdata_exp.sh` like the test output suggested. I don't know how to evaluate the output, so please feel free to advise if the output was wrong or unexpected 🙇‍♂️.

jez (approved):

jez:
Hey sorry for the delay, I was on vacation for a few weeks.

The new output looks good, it's just saying that certain definitions are on different lines, which makes sense because you deleted a few lines from the file.

(If instead of deleting the `# ^^^ error` lines you had replaced them with empty lines, the snapshot output would not have changed, which is an approach you can consider in the future.)

PR #8152, opened 2024-09-04
Support native Darwin ARM64 builds

vinistock:
The original PR for this is #7922, but I'm re-opening it using Shopify's fork so that everyone can contribute to the PR.

This PR adds support for compiling Sorbet to ARM64 and for publishing the ARM64 binaries as a universal one, including both x86 and ARM64 support.

Please see the original PR for more context and discussions.


froydnj (approved):
I can't see anything wrong with this, and it sounds like y'all have used basically the same commands in `build-static-release.sh` for a separate CI pipeline.

Our Mac builds are only tested on master, though, so I will watch this as it merges and revert if something goes wrong rather than trying to debug on master.  (Also 🤞  that those Mac builds aren't totally hosed, because I don't have access to those machines.)

PR #8155, opened 2024-09-05
Compile mac release on ARM64

vinistock:
Will fill in shortly.


vinistock:
This won't work because the ARM64 machines are not yet online for the buildkite agents.

PR #8157, opened 2024-09-06
Ensure that aliased parent classes are handled correctly in subclass list generation

aadi-stripe:
cc @zanker-stripe 

Subclass list generation in autogen has a bug in it, where if you alias a parent class and use it at the subclass declaration site, we lose the reference and don't include it the corresponding subclass the list.

### Motivation

Fixing autogen.

### Test plan

See included automated tests. Tested on Stripe codebase

### DIFF
--- a/main/autogen/subclasses.cc
+++ b/main/autogen/subclasses.cc
@@ -54,7 +54,7 @@ optional<Subclasses::Map> Subclasses::listAllSubclasses(core::Context ctx, const
             continue;
         }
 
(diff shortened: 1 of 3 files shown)


froydnj (approved):

aadi-stripe:

→ https://go/builds/bui_QngO8JSMrVBeb0
→ https://go/builds/bui_QngOi01RonjQSD
→ https://go/builds/bui_QngO8PrY1AiaO1

PR #8161, opened 2024-09-09
Improve CSV annotations

bdewater:
### Motivation
Ran into some type ergonomics issues when parsing CSVs and decided to fix them. See https://github.com/ruby/csv/blob/master/lib/csv.rb for source.

### Test plan
See included automated tests.


ilyailya:

→ https://go/builds/bui_QpFlvI4BoFaMmy
→ https://go/builds/bui_QpFlQirWUlqhwH
→ https://go/builds/bui_QpFlnkN1QZjHU3

froydnj:

→ https://go/builds/bui_QpGFPy4SvWYdce
→ https://go/builds/bui_QpGFBOgnQlwIWB
→ https://go/builds/bui_QpGF6Z8DO4Qgez

rbi/stdlib/csv.rbi
```diff
@@ -898,6 +930,7 @@ class CSV < Object
   # Also aliased as:
   # [`gets`](https://docs.ruby-lang.org/en/2.7.0/CSV.html#method-i-gets),
   # [`readline`](https://docs.ruby-lang.org/en/2.7.0/CSV.html#method-i-readline)
+  sig { returns(T.nilable(T.any(T::Array[T.untyped], CSV::Row))) }
```
froydnj, line 933: We have a fair amount of code at Stripe that assumes the return type from this (and other methods that have similar `T.any(T::Array, CSV::Row)` return types) is a `CSV::Row` and then calls `row["field_name"]`, which now errors with:

```
file.rb:366: Expected T.any(Integer, Float) but found String("query") for argument arg0 http://go/e/7002
     366 |          sql: row["query"],
                             ^^^^^^^
  Expected T.any(Integer, Float) for argument arg0 of method Array#[]:
    https://github.com/sorbet/sorbet/tree/5d0898eb914e813772aceaf0fdc9cb71d2907584/rbi/core/array.rbi#L560:
     560 |        arg0: T.any(Integer, Float),
                  ^^^^
  Got String("query") originating from:
    file.rb:366:
     366 |          sql: row["query"],
                             ^^^^^^^
```

Arguably a lot of these cases should use `T.cast`...but I don't know if that makes this change any more ergonomic or not.

froydnj (changes requested):
Marking as request changes just because we can't merge this as-is due to Stripe's codebase.

PR #8162, opened 2024-09-10
Add cross compilation support for Darwin

vinistock:
CAN'T BE MERGED BEFORE https://github.com/sorbet/bazel-toolchain/pull/10

This PR is the second attempt of #8152

Instead of using the original approach of compiling a native ARM64 binary and an x86 binary through rosetta (which is this PR's first commit), this PR adds support for cross compiling `x86 -> arm64` and vice versa.

The major advantage of this approach is that we always use the Bazel installation native to the machine's architecture, instead of needing to remove and re-download so that we can swap between native vs rosetta version.

### Details

The implementation follows the steps outlined in `toolchains_llvm` [documentation](https://github.com/bazel-contrib/toolchains_llvm?tab=readme-ov-file#cross-compilation). Essentially, we need to define the `sysroot` for the different architectures that we want to cross compile to. It's a bit odd because on MacOS, the sysroot is always the MacOS SDK regardless of which target you're trying to compile to, but that's the only way `toolchains_llvm` will allow you to cross-compile.

Then we define configurations for cross compiling in both directions in `.bazelrc`. The buildkite release script will now use those configurations to produce both binaries.

### Test plan

We were able to manually test cross compilation from `arm64 -> x86`, which produces a working x86 binary. However, we don't have an x86 Darwin machine, so maybe we can test it on CI?

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -137,12 +137,23 @@ build:release-sanitized-linux --copt=-march=sandybridge
 build:release-linux-aarch64   --copt=-march=armv8.1a
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
+
 build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
 build:release-mac-arm64 --define release=true
 build:release-mac-arm64 --compilation_mode=opt
 build:release-mac-arm64 --config=backtracesymbols
 build:release-mac-arm64 --config=shared-libs
 build:release-mac-arm64 --config=versioned
+
+# Cross-compile to x86 on an ARM64 machine
+build:release-mac-cross --config=release-mac
+build:release-mac-cross --extra_toolchains=@llvm_toolchain_15_0_7//:cc-toolchain-x86_64-darwin
+build:release-mac-cross --copt=-stdlib=libc++ --linkopt=-lc++
+
+# Cross-compile to ARM64 on an x86 machine
+build:release-mac-arm64-cross --config=release-mac-arm64
+build:release-mac-arm64-cross --extra_toolchains=@llvm_toolchain_15_0_7//:cc-toolchain-aarch64-darwin
+build:release-mac-arm64-cross --copt=-stdlib=libc++ --linkopt=-lc++
 
 build:release-debug-linux --config=release-linux
 build:release-debug-linux --config=release-debug-common
--- a/.bazelrc
+++ b/.bazelrc
@@ -137,6 +137,12 @@ build:release-sanitized-linux --copt=-march=sandybridge
 build:release-linux-aarch64   --copt=-march=armv8.1a
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
+build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
+build:release-mac-arm64 --define release=true
+build:release-mac-arm64 --compilation_mode=opt
+build:release-mac-arm64 --config=backtracesymbols
+build:release-mac-arm64 --config=shared-libs
+build:release-mac-arm64 --config=versioned
 
 build:release-debug-linux --config=release-linux
 build:release-debug-linux --config=release-debug-common
--- a/.bazelrc
+++ b/.bazelrc
@@ -137,13 +137,7 @@ build:release-sanitized-linux --copt=-march=sandybridge
 build:release-linux-aarch64   --copt=-march=armv8.1a
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
-
-build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
-build:release-mac-arm64 --define release=true
-build:release-mac-arm64 --compilation_mode=opt
-build:release-mac-arm64 --config=backtracesymbols
-build:release-mac-arm64 --config=shared-libs
-build:release-mac-arm64 --config=versioned
+build:release-mac-arm64 --config=release-common --config=shared-libs --platforms=@//tools/platforms:darwin_arm64
 
 # Cross-compile to x86 on an ARM64 machine
 build:release-mac-cross --config=release-mac
--- a/.buildkite/build-static-release.sh
+++ b/.buildkite/build-static-release.sh
@@ -34,12 +34,10 @@ echo will run with $CONFIG_OPTS
 
 case "$platform" in
   darwin-x86_64|darwin-arm64)
-    TARGET_PLATFORM=darwin-x86_64 ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
     cp bazel-bin/main/sorbet sorbet_x86_64
 
-    ./bazel clean --expunge
-
-    TARGET_PLATFORM=darwin-arm64 ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    ./bazel build //main:sorbet --strip=always $CONFIG_OPTS --config=release-mac-arm64-cross
     cp bazel-bin/main/sorbet sorbet_arm64
 
     lipo -create -output sorbet_bin sorbet_x86_64 sorbet_arm64
--- a/.buildkite/build-static-release.sh
+++ b/.buildkite/build-static-release.sh
@@ -130,4 +130,4 @@ if [[ "$platform" == "linux-x86_64" ]]; then
 fi
 
 mkdir -p _out_/$platform
-cp bazel-bin/main/sorbet _out_/$platform/
+cp sorbet_bin _out_/$platform/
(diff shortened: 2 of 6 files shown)


third_party/externals.bzl
```diff
@@ -186,9 +186,9 @@ def register_sorbet_dependencies():
     # In 2ddd7d791 (#7912) we upgraded the toolchain. Our old toolchain patches are on the `sorbet-old-toolchain` branch
     http_archive(
         name = "toolchains_llvm",
-        url = "https://github.com/sorbet/bazel-toolchain/archive/8d9165fd3560f6ff50bc4794972f714f4ba2adaa.tar.gz",
-        sha256 = "238b5a777bbfac3d5ec35cbd45ae2b84ca4118aa8f902ab27894912352e94658",
-        strip_prefix = "bazel-toolchain-8d9165fd3560f6ff50bc4794972f714f4ba2adaa",
+        url = "https://github.com/Shopify/bazel-toolchain/archive/5b4915c6730e86f21b1f2624f008a6bf9999910d.tar.gz",
+        sha256 = "103b72e9ece85c7d329143bc612804721c659e61808a412ceadcd96075cd1636",
+        strip_prefix = "bazel-toolchain-5b4915c6730e86f21b1f2624f008a6bf9999910d",
```
froydnj, line 191: Can we change this to a master commit, please?

froydnj (changes requested):

third_party/externals.bzl
Morriar: Rebased on top of the Sorbet branch: https://github.com/sorbet/bazel-toolchain/commits/sorbet/ 👍

third_party/externals.bzl
```diff
@@ -186,9 +186,9 @@ def register_sorbet_dependencies():
     # In 2ddd7d791 (#7912) we upgraded the toolchain. Our old toolchain patches are on the `sorbet-old-toolchain` branch
     http_archive(
         name = "toolchains_llvm",
-        url = "https://github.com/sorbet/bazel-toolchain/archive/8d9165fd3560f6ff50bc4794972f714f4ba2adaa.tar.gz",
-        sha256 = "238b5a777bbfac3d5ec35cbd45ae2b84ca4118aa8f902ab27894912352e94658",
-        strip_prefix = "bazel-toolchain-8d9165fd3560f6ff50bc4794972f714f4ba2adaa",
+        url = "https://github.com/Shopify/bazel-toolchain/archive/5ed6d56dd7d2466bda56a6237c5ed70336b95ee5.tar.gz",
```
jez, line 189: Why is there a Shopify URL here?

Morriar: My bad, we were using our clone until the changes to `bazel-toolchain` get merged. We can switch back to sorbet's clone 👍

froydnj (approved):
Thanks!

froydnj (dismissed):
The static-release script is taking a very long time on our OS X machine; I think it's possible there's something wrong with it.

Also, the `build-static-release.sh` script ends with:

```
cp bazel-bin/main/sorbet _out_/$platform/
```

which isn't consistent with the `sorbet_bin` changes above.

Morriar:
@froydnj 

> The static-release script is taking a very long time on our OS X machine

Do you have any output showing where the build is stuck?

> Also, the build-static-release.sh script ends with: `cp bazel-bin/main/sorbet _out_/$platform/`

Yes, I think we can remove this change and just create the universal binary at `bazel-bin/main/sorbet`

froydnj:
> @froydnj
> 
> > The static-release script is taking a very long time on our OS X machine
> 
> Do you have any output showing where the build is stuck?

https://buildkite.com/sorbet/sorbet/builds/32590#0191fc8b-c0b0-4bb7-8dbf-cdb5873533c3

Jake said the machine wasn't really doing anything interesting.

> > Also, the build-static-release.sh script ends with: `cp bazel-bin/main/sorbet _out_/$platform/`
> 
> Yes, I think we can remove this change and just create the universal binary at `bazel-bin/main/sorbet`

I asked that we not create it under `bazel-bin/main/sorbet` in #8152 (apparently Github ate my comment?) because `bazel-bin/` is Bazel's domain, not ours, and we're not supposed to be touching it.  Please revert those changes.

.buildkite/build-static-release.sh
```diff
@@ -28,7 +32,21 @@ esac
...
+case "$platform" in
+  darwin-x86_64|darwin-arm64)
+    ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    mv bazel-bin/main/sorbet sorbet_x86_64
+
+    ./bazel build //main:sorbet --strip=always $CONFIG_OPTS --config=release-mac-arm64-cross
+    mv bazel-bin/main/sorbet sorbet_arm64
+
+    lipo -create -output bazel-bin/main/sorbet sorbet_x86_64 sorbet_arm64
+    rm sorbet_x86_64 sorbet_arm64
```
jez, line 44: I believe that this command is hanging, because `sorbet_x86_64` and `sorbet_arm64` are copied out of the bazel sandbox, and all files in the bazel sandbox are write protected (`r-xr-xr-x` / `555`).

I think that you have to `-f` this
```suggestion
    rm -f sorbet_x86_64 sorbet_arm64
```

jez:
Heads up that I've landed this change:

https://github.com/sorbet/sorbet/pull/8169

which changes the way that the `./bazel` command works so that checking out old and new Sorbet commits will always have the right version of Bazel installed.

(This should make it so that if, after this PR were to land, someone clones the repo fresh, builds Sorbet, then checks out an old commit and attempts to build it, their second build correctly uses the x86_64 version of Bazel, instead of re-using the already-downloaded arm64 version of Bazel.)

Morriar:
> Please revert those changes.

@froydnj done, but I'm confused about what you were asking in the first place then?

> Heads up that I've landed this change, which will conflict: https://github.com/sorbet/sorbet/pull/8169

Rebased on `master` and solved the conflicts 👍

.buildkite/build-static-release.sh
froydnj: I would think that would just error, rather than hang?

froydnj:
> > Please revert those changes.
> 
> @froydnj done, but I'm confused about what you were asking in the first place then?

This line:

https://github.com/sorbet/sorbet/pull/8162/files#diff-503e2940b9cafef5df10484afabe6f20c6f20cd19b313c9f7ce3bf06ee3a2df5L114

shouldn't be using something out of `bazel-bin/`.

.bazelrc
```diff
@@ -138,6 +138,23 @@ build:release-linux-aarch64   --copt=-march=armv8.1a
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
 
+build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
+build:release-mac-arm64 --define release=true
+build:release-mac-arm64 --compilation_mode=opt
+build:release-mac-arm64 --config=backtracesymbols
+build:release-mac-arm64 --config=shared-libs
+build:release-mac-arm64 --config=versioned
```
jez, line 146: should this be

```
```suggestion
build:release-mac-arm64 --config=release-common --config=shared-libs --platforms=@//tools/platforms:darwin_arm64
```

to share more code?

.bazelrc
```diff
@@ -138,6 +138,23 @@ build:release-linux-aarch64   --copt=-march=armv8.1a
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
 
+build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
+build:release-mac-arm64 --define release=true
+build:release-mac-arm64 --compilation_mode=opt
+build:release-mac-arm64 --config=backtracesymbols
+build:release-mac-arm64 --config=shared-libs
```
jez, line 145: If we're using shared libs, is there a risk of the binary not being able to be used on a machine that differs from the build environment?

.buildkite/build-static-release.sh
```diff
@@ -28,10 +32,25 @@ esac
...
-./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+case "$platform" in
+  darwin-x86_64|darwin-arm64)
+    ./bazel build //main:sorbet --strip=always $CONFIG_OPTS
+    cp bazel-bin/main/sorbet sorbet_x86_64
+
+    ./bazel build //main:sorbet --strip=always $CONFIG_OPTS --config=release-mac-arm64-cross
+    cp bazel-bin/main/sorbet sorbet_arm64
+
+    lipo -create -output sorbet_bin sorbet_x86_64 sorbet_arm64
```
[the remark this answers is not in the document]
froydnj: The bazel gymnastics for compiling for multiple architectures simultaneously can be done, but I think that would a) (definitely) be followup work and b) it's unlikely that it would be run anywhere outside of master builds.

froydnj (approved):
Looks great.  Thanks for working through the issues with us!

PR #8164, opened 2024-09-11
Desugar case when nodes specially when called from extract to variable

neilparikh:
Similar to #8140 and #8142

### Motivation

We desugar case-when to something like this:
```
begin
  <assignTemp>$2 = <emptyTree>::<C T>.unsafe(1)
  if nil.===(<assignTemp>$2)
    <emptyTree>
  else
    if <emptyTree>::<C A>.===(<assignTemp>$2)
      1
    else
      ...
    end
  end
end
```

Similar to csend, extract-to-variable tries to insert in this fake `If`. Instead, we can desugar to something like:
```
<Magic>.caseWhen(temp, nil, <emptyTree>, <emptyTree>::<C A>, 1, ...)
```

### Test plan

See included automated tests.


ast/desugar/Desugar.cc
```diff
@@ -2197,6 +2197,22 @@ ExpressionPtr node2TreeImpl(DesugarContext dctx, unique_ptr<parser::Node> what)
                 result = std::move(res);
             },
             [&](parser::Case *case_) {
+                if (dctx.preserveConcreteSyntax) {
+                    Send::ARGS_store args;
+                    args.emplace_back(node2TreeImpl(dctx, std::move(case_->condition)));
+                    for (auto it = case_->whens.begin(); it != case_->whens.end(); ++it) {
+                        auto when = parser::cast_node<parser::When>(it->get());
+                        ENFORCE(when != nullptr, "case without a when?");
```
ilyailya, line 2205: [super nit]
Would it make sense to rephrase the message in `ENFORCE`? It could be something like 
`found a case without when in desugaring stage for extract variable code action`

ilyailya (approved):
LGTM!
I left one minor nit regarding message in enforce. Feel free to ignore

ast/desugar/Desugar.cc
neilparikh: I copied this `ENFORCE` from the regular/non-extract variable case, and if it occurs here, it should occur there too, so I don't think it's worth mentioning the message that's here (and the enforce failure will mention the line number if we do need to know that).

PR #8173, opened 2024-09-17
Show doc comments above multiline sigs with braces

### LINKED ISSUE #8127: Method comments not appearing with multiline signatures using braces
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0A%23%20foobar%0Asig%20%7B%20void%20%7D%0Adef%20method1%3B%20end%0A%0A%23%20foobar%0Asig%20%7B%0A%20%20void%0A%7D%0Adef%20method2%3B%20end%0A)

```ruby
# typed: true
extend T::Sig

# foobar
sig { void }
def method1; end

# foobar
sig {
  void
}
def method2; end
```

#### Observed output
image
image

#### Expected behavior

`method2` should display the method comment
- - -

reese:
This adds support for showing documentation on hover for multi-line sigs that use braces. For example:

```ruby
# Show me!
sig {
  void
}
def example!; end
```

This implementation makes some particular assumptions, notably that the closing brace is on its own line.

```ruby
# This works!
sig { void
}
def foo; end

# This doesn't
sig {
void }
def foo; end
```

This is a fair enough assumption in the Stripe codebase, and I imagine it's probably good enough for most cases -- the preexisting code is already based on these sorts of assumptions for `do`/`end`, and handling all possible cases likely isn't worth the complexity, but happy to be told otherwise if folks feel differently!

### Motivation

Closes #8127

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/lsp_helpers.cc
+++ b/main/lsp/lsp_helpers.cc
@@ -291,27 +291,28 @@ optional<string> findDocumentation(string_view sourceCode, int beginIndex) {
             //       .returns(Bar) }
             // ```
             it++;
+            line = absl::StripAsciiWhitespace(*it);
             while (
                 // SOF
                 it != all_lines.rend()
                 // Start of sig block
-                && !(absl::StartsWith(absl::StripAsciiWhitespace(*it), "sig {") ||
-                     absl::StartsWith(absl::StripAsciiWhitespace(*it), "sig(:final) {"))
+                && !(absl::StartsWith(line, "sig {") ||
+                     absl::StartsWith(line, "sig(:final) {"))
                 // Invalid closing brace
-                && !absl::StartsWith(absl::StripAsciiWhitespace(*it), "}")) {
+                && !absl::StartsWith(line, "}")) {
                 it++;
+                line = absl::StripAsciiWhitespace(*it);
             };
 
             // We have either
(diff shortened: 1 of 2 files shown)


main/lsp/lsp_helpers.cc
```diff
@@ -280,6 +280,41 @@ optional<string> findDocumentation(string_view sourceCode, int beginIndex) {
...
+            // ```
+            // sig { params(foo: Foo)
+            //       .returns(Bar) }
+            // ```
+            it++;
+            while (
+                // SOF
+                it != all_lines.rend()
+                // Start of sig block
+                && !(absl::StartsWith(absl::StripAsciiWhitespace(*it), "sig {") ||
```
jez, line 298: What do you think about factoring the code so that we only have to strip whitespace once, instead of possibly 5 times per loop?

main/lsp/test/lsp_test.cc
jez, line 1: I had no idea this file existed.

For future reference I would have tested this with a `test/testdata/lsp/` test using `# ^ hover: ...` assertions, but since this file already exists, this is fine too.

jez (approved):
Thanks!

PR #8178, opened 2024-09-18
Move some functions from payload.cc to cache.cc

jez:
### Motivation

These functions were only ever used by cache.cc, and were not actually
related to Sorbet's payload--they're more generic functions for interacting
with the disk cache.

Allows me to remove a TODO.

### Test plan

See included automated tests.


main/pipeline/pipeline.cc
```diff
@@ -1254,74 +1254,6 @@ void printFileTable(unique_ptr<core::GlobalState> &gs, const options::Options &o
 #endif
 }
 
-bool cacheTreesAndFiles(const core::GlobalState &gs, WorkerPool &workers, absl::Span<const ast::ParsedFile> parsedFiles,
```
neilparikh, line 1268: You need to delete this from `pipeline.h` too

main/cache/cache.cc
```diff
@@ -21,13 +22,116 @@ unique_ptr<OwnedKeyValueStore> maybeCreateKeyValueStore(shared_ptr<::spdlog::log
...
+    return false;
+}
+
+bool cacheTreesAndFiles(const core::GlobalState &gs, WorkerPool &workers, absl::Span<const ast::ParsedFile> parsedFiles,
+                        const unique_ptr<OwnedKeyValueStore> &kvstore) {
+    if (kvstore == nullptr) {
+        return false;
+    }
+
+    Timer timeit(gs.tracer(), "pipeline::cacheTreesAndFiles");
```
neilparikh, line 63:
```suggestion
    Timer timeit(gs.tracer(), "cache::cacheTreesAndFiles");
```

main/pipeline/pipeline.cc
```diff
@@ -84,15 +82,6 @@ class CFGCollectorAndTyper {
     }
 };
 
-string fileKey(const core::File &file) {
```
neilparikh, line 87: delete from `pipeline.h`

neilparikh (approved):

main/cache/cache.cc
jez: Thanks–I intentionally left this here because I didn't want to change the name of the associated metric.

main/pipeline/pipeline.cc
jez: thanks

main/pipeline/pipeline.cc
jez: thanks

PR #8179, opened 2024-09-18
Separate scopes to compare and insert for LCA

neilparikh:
A better approach to #8104 that handles more cases.

### Motivation

Where we want to insert into, and where we want to compare for matches when trying to find the lowest common ancestors scope for multiple occurrences is not always the same. For example:

```ruby
def foo
  if 1 + 1
    2 + 2
  else
    3 + 3
  end
end
```

If we're trying to extract the `1`, when trying to find the lowest scope, we want to compare the other matches in the condition of the `if`, but if they're all there, we don't want to insert in the condition; we want to insert outside the `if`. Similarly, if we're trying to extract `2` or `3`, we want to compare in the `thenp` or `elsep`, but we don't want to insert in those (they're send nodes, which we can't insert into); we want to insert into the entire `if`, and then determine later which part of it we should insert into.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -314,6 +314,7 @@ class ExpressionPtrSearchWalk {
         if (LCAScopeStack.empty()) {
             for (auto *scope : enclosingScopeStack) {
                 const ast::ExpressionPtr *scopeToCompare = scope;
+                const ast::ExpressionPtr *scopeToInsertIn = scope;
                 if (ast::isa_tree<ast::If>(*scope)) {
                     auto &if_ = ast::cast_tree_nonnull<ast::If>(*scope);
                     if (if_.thenp.loc().exists() && if_.thenp.loc().contains(matchLoc)) {
@@ -322,6 +323,7 @@ class ExpressionPtrSearchWalk {
                         scopeToCompare = &if_.elsep;
                     } else if (if_.cond.loc().contains(matchLoc)) {
                         scopeToCompare = &if_.cond;
+                        scopeToInsertIn = LCAScopeStack.back().second;
                     } else {
                         ENFORCE(false);
                     }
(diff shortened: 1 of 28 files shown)


main/lsp/ExtractVariable.cc
```diff
@@ -45,6 +45,8 @@ core::LocOffsets findWhereToInsert(const ast::ExpressionPtr &scope, const core::
             whereToInsert = if_->thenp.loc();
         } else if (if_->elsep.loc().exists() && if_->elsep.loc().contains(target)) {
             whereToInsert = if_->elsep.loc();
+        } else if (if_->cond.loc().exists() && if_->cond.loc().contains(target)) {
+            ENFORCE(false, "shouldn't be inserting into the cond of an if");
```
jez, line 49: Minor, but you might want to move this into a `DEBUG_ONLY` or `if constexpr debugmode` condition in the `else` to be sure that we don't pay the cost of the `if` condition in release builds, because the `else` branch is also an `ENFORCE(false)`.

main/lsp/ExtractVariable.cc
```diff
@@ -61,7 +63,11 @@ core::LocOffsets findWhereToInsert(const ast::ExpressionPtr &scope, const core::
     } else if (auto rescueCase = ast::cast_tree<ast::RescueCase>(scope)) {
         whereToInsert = rescueCase->body.loc();
     } else if (auto while_ = ast::cast_tree<ast::While>(scope)) {
-        whereToInsert = while_->body.loc();
+        if (while_->body.loc().exists() && while_->body.loc().contains(target)) {
+            whereToInsert = while_->body.loc();
+        } else if (while_->cond.loc().exists() && while_->cond.loc().contains(target)) {
+            ENFORCE(false, "shouldn't be inserting into the cond of a while");
+        }
```
jez, line 70:
```suggestion
        } else {
            ENFORCE(!(while_->cond.loc().exists() && while_->cond.loc().contains(target)), "shouldn't be inserting into the cond of a while");
        }
```

jez (approved):
some comments on how to simplify the code, high level seems fine. thanks for the tests.

main/lsp/ExtractVariable.cc
neilparikh: I'll just do what you suggested for the `while` case (move the condition to the `ENFORCE`), there's no other case to handle now that we have an case for all three parts of the `if`.

ilyailya (approved):

PR #8180, opened 2024-09-18
Revert "Add cross compilation support for Darwin (#8162)"

jez:
This reverts commit be27e0a819b6b42ccb7b9f17b2d149939b85bc1a.

### Motivation

Across multiple macOS laptops and multiple colleagues, I'm not able to run the
macOS Sorbet produced by this commit in any form (both from the gem and from the
binary release artifact).

Going to revert while we figure things out.

cc @vinistock @Morriar

### Test plan

```bash
❯ curl -L https://github.com/sorbet/sorbet/releases/download/0.5.11568.20240918161953-be27e0a81/darwin-x86_64.sorbet_bin > sorbet_bin
❯ chmod +x ./sorbet_bin
❯ ./sorbet_bin
zsh: killed	./sorbet_bin
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -137,17 +137,6 @@ build:release-sanitized-linux --copt=-march=sandybridge
 build:release-linux-aarch64   --copt=-march=armv8.1a
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
-build:release-mac-arm64 --config=release-common --config=shared-libs --platforms=@//tools/platforms:darwin_arm64
-
-# Cross-compile to x86 on an ARM64 machine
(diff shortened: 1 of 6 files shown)


froydnj (approved):
Ugh!  Both architectures, if you flatten the binary, too?

jez:
@froydnj What do you mean by flatten the binary?

I believe the name for the artifact is wrong—the `darwin-x86_64.sorbet_bin` file in the GitHub release is a universal binary, with two architectures

But also, all of Getty, Neil, and have the same problem when we run

```
```

which produces only the arm64 binary (successfully) and then fails to build.

Mysteriously, `--config=dbg` works fine (and produces an arm64 binary).

PR #8181, opened 2024-09-18
--stripe-mode -> --uniquely-defined-behavior

jez:
### Motivation

This error is something that people might want to use outside of Stripe, and
there's nothing Stripe-specific about this error.

Also I think it removes the temptation to think about adding Sorbet features
like, "it's okay to add Stripe-specific features as long as we only add them in
stripe-mode." Sorbet's features should be useful outside of Stripe.

### Test plan

I haven't left any references to `--stripe-mode` in our tests, and I haven't
added a new test that says that you should use this one instead, because I think
that this option is in practice not being used by anyone, and I plan to remove
it fully tomorrow after I've upgraded Stripe's codebase to use the new option.


jez:
This change was reviewed in #8178, but I accidentally merged it into the PR it was stacked on, not `master`.

test/cli/package-multiple-files-behavior/test.sh
```diff
@@ -0,0 +1,2 @@
+# Ensures that Sorbet doesn't complain about the package having behavior defined in multiple files in Stripe mode.
```
neilparikh, line 1: forgot to replace the "stripe mode" here

neilparikh (approved):

test/cli/package-multiple-files-behavior/test.sh
jez: that'll be our little easter egg

PR #8183, opened 2024-09-19
Release binary for Darwin ARM64 platform

Morriar:
Reroll of #8162 with 2 changes:

* Dropped the commit for reusing of `release-common`: it resulted in both `static-libs` and `shared-libs` being used and confused the linker
* [59a8605](https://github.com/sorbet/sorbet/pull/8183/commits/59a8605c9f1a47dd17a2738992135b9000d6f4a0): Fixed the options passed to bazel when cross compiling from x86_64 to arm64 to not pass both `release-mac` and `release-mac-arm64-cross`

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -137,21 +137,15 @@ build:release-linux           --copt=-march=sandybridge
 build:release-sanitized-linux --copt=-march=sandybridge
 build:release-linux-aarch64   --copt=-march=armv8.1a
 
-build:release-mac --config=release-common --config=static-libs --platforms=@//tools/platforms:darwin_x86_64
+build:release-mac --config=release-common --config=shared-libs
 
-build:release-mac-arm64 --config=release-common
-build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
-build:release-mac-arm64 --config=shared-libs
+build:cross-to-darwin-x86_64 --platforms=@//tools/platforms:darwin_x86_64
+build:cross-to-darwin-x86_64 --extra_toolchains=@llvm_toolchain_15_0_7//:cc-toolchain-x86_64-darwin
+build:cross-to-darwin-x86_64 --copt=-stdlib=libc++ --linkopt=-lc++
 
-# Cross-compile to x86 on an ARM64 machine
-build:release-mac-cross --config=release-mac
-build:release-mac-cross --extra_toolchains=@llvm_toolchain_15_0_7//:cc-toolchain-x86_64-darwin
-build:release-mac-cross --copt=-stdlib=libc++ --linkopt=-lc++
-
-# Cross-compile to ARM64 on an x86 machine
-build:release-mac-arm64-cross --config=release-mac-arm64
-build:release-mac-arm64-cross --extra_toolchains=@llvm_toolchain_15_0_7//:cc-toolchain-aarch64-darwin
-build:release-mac-arm64-cross --copt=-stdlib=libc++ --linkopt=-lc++
+build:cross-to-darwin-arm64 --platforms=@//tools/platforms:darwin_arm64
+build:cross-to-darwin-arm64 --extra_toolchains=@llvm_toolchain_15_0_7//:cc-toolchain-aarch64-darwin
+build:cross-to-darwin-arm64 --copt=-stdlib=libc++ --linkopt=-lc++
 
 build:release-debug-linux --config=release-linux
 build:release-debug-linux --config=release-debug-common
--- a/.bazelrc
+++ b/.bazelrc
@@ -91,7 +91,6 @@ build:dbg-linux --config=dbg --platforms=@//tools/platforms:linux_x86_64
 build:release-common --define release=true
 build:release-common --compilation_mode=opt
 build:release-common --config=backtracesymbols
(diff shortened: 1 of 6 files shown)


.buildkite/build-static-release.sh
```diff
@@ -16,7 +16,11 @@ case "$platform" in
   linux-aarch64)
     CONFIG_OPTS="--config=release-${platform}"
     ;;
-  darwin-x86_64|darwin-arm64)
+  darwin-x86_64)
+    CONFIG_OPTS="--config=release-mac"
+    command -v autoconf >/dev/null 2>&1 || brew install autoconf
+    ;;
+  darwin-arm64)
```
jez, line 23: What's the difference between these two cases now? the darwin-arm64 case has the same logic right?

Also, is the darwin-arm64 case untested, because we don't have any macOS arm64 machines in buildkite?

That means that we still have the problem that

```
```

doesn't produce working binaries when we run it on our laptops, doesn't it?

jez: Ah, I see—it's because the release-common / shared-libs thing changed too.

.bazelrc
```diff
@@ -138,6 +138,23 @@ build:release-linux-aarch64   --copt=-march=armv8.1a
 
 build:release-mac --config=release-common --platforms=@//tools/platforms:darwin_x86_64
 
+build:release-mac-arm64 --platforms=@//tools/platforms:darwin_arm64
+build:release-mac-arm64 --define release=true
+build:release-mac-arm64 --compilation_mode=opt
+build:release-mac-arm64 --config=backtracesymbols
+build:release-mac-arm64 --config=shared-libs
+build:release-mac-arm64 --config=versioned
```
jez, line 146: Now that we know where the problem was, can we still put all the common stuff in `release-common`, like `compilation_opt` and `versioned` etc. and only have the `platforms` and `{shared,static}-libs` options wherever?

jez:
I tested that with these changes

```
bazel build --config=release-mac-arm64 //main:sorbet
```

is able to build an arm64 binary that runs.

jez:
@Morriar I created a branch on top of yours

https://github.com/sorbet/sorbet/compare/jez-arm64

I think this approach is a little more general/simpler, curious what your thoughts are.

You can check the commit messages for some more motivations

- **no-op: Move --config=static-libs into all users of --config=release-common** (dbb8bb3ad5)


- **Use --config=release-common in release-mac-arm64** (075e6cbe96)


- **no-op: Consolidate darwin cases down** (acc50dd29a)


- **Allow the cross compilation flags to work with any target** (59a89f44f0)

  No reason why these configurations have to hard-code the `release-mac`
  config as a base--they can be chained with any config, like
  `--config=dbg` or `--config=release-debug-mac`.

  Tested that this works on my laptop:

  ```
  ❯ bazel build --config=dbg --config=cross-to-darwin-x86_64 //main:sorbet
  # ...

  ❯ file bazel-bin/main/sorbet
  bazel-bin/main/sorbet: Mach-O 64-bit executable x86_64

  ❯ bazel-bin/main/sorbet
  You must pass either `-e` or at least one folder or ruby file.
  # ...
  ```

- **Simplify build-static-release.sh** (03c515be8c)

  - Writes the (untested) code that could work if we were to switch to
   running macOS CI from arm64, instead of x86_64.

  - Publishes all of the macOS binaries (architecture-specific and
   universal).

  - Fixes the name used for the macOS sorbet_bin in GitHub releases (it
   used to use the `$platform` of the machine that built it like
   `darwin-x86_64.sorbet`, even though it was a universal binary).

  - Consolidates branches to share more code.

- **Put this back to make the diff smaller** (ea14b5d847)


- **Add a comment** (7e85bbddb3)


- **Keep the artifact names the same (not `sorbet_bin`)** (1a860fba46)

jez:
If you're happy with those changes, feel free to merge them into your branch and I'll approve this PR.

jez:
I tested this on macOS in [this build](https://buildkite.com/sorbet/sorbet/builds/32677#01920c65-d7ee-49cf-877c-905ecd62c4d4) and the names in the Artifact tab are better (don't mention `sorbet_bin` anymore, and include all macOS platforms).

vinistock:
@jez I think all of your changes make perfect sense. I just have a question about [this commit](https://github.com/sorbet/sorbet/commit/59a89f44f002a118b68234ccf4afadad270cef8f).

Here, we're changing all Darwin builds to use `shared-libs` - not just the ARM64 one. Was that intentional? Not sure if there's any difference considering the universal binary.

Let me know your thoughts and I can merge your commits into this branch.

jez:
> Here, we're changing all Darwin builds to use `shared-libs` - not just the ARM64 one. Was that intentional? Not sure if there's any difference considering the universal binary.

Yeah, that was intentional. I took both the x86_64 and the universal binaries from that job (produced from an x86_64 macbook) and was able to run them on my personal x86_64 macbook. I was able to do the same with the arm64 and universal binaries on my work laptop (running arm64).

So I think that we're good to go as far as linker support.

vinistock:
@jez merged your branch into this one. Looks good to me!

jez (approved):

Issue #8166, opened 2024-09-12
Static checker crashes on Array literals with `ForwardedRestArg`


amomchilov:
#### Input

[→ View on sorbet.run](https://sorbet.run/#def%20demo%28*%29%20%3D%20%5B%3Abefore%2C%20*%2C%20%3Aafter%5D)

```ruby
def demo(*) = [:before, *, :after]
```

#### Observed output

Crashes in `Desugar.cc`

```
Exception::raise(): Unimplemented Parser Node: ForwardedRestArg

editor.rb:1: Failed to process tree (backtrace is above) https://srb.help/1001
     1 |def demo(*) = [:before, *, :after]
                                ^

-e: Exception parsing file: -e (backtrace is above) https://srb.help/1001
Errors: 2
```

#### Expected behavior

Not to crash on this valid syntax. By comparison, `ForwardedRestArg` works correctly on `Send` nodes, e.g.

```ruby
def demo2(*) = T.unsafe(self).p(:before, *, :after)
```

And the Hash equivalent of this also works correctly

```ruby
def demo3(**) = { before: 1, **, after: 2}
```

#### Implementation notes

This is relatively niche syntax (it hasn't been an issue so far), and there's no urgency to support this (at least from our perspective). We'd be happy to address this when we rewrite `Desugar.cc` to use Prism.

jez:
I think that the urgency to support this is that this causes Sorbet to crash even when the user has partially typed something, which causes Sorbet to completely re-initialize itself.

e.g., if you mean to type `(` but you typo that and hit `*` instead, and you're in the middle of an array literal or a hash literal, then suddenly sorbet completely crashes.

I'm going to implement support for this by not crashing (even if only by inserting an untyped/error node).

PR #8186, opened 2024-09-20
Write trees to the cache even if GlobalState did not change

jez:
I do not plan to merge this until the corresponding PR in Stripe's
codebase which re-enables `--cache-dir` has sat for a while
(at least a week or two, maybe more).

### Motivation

- First attempt: c9418bb
- (Semantically) reverted in: 6f56fa4

This change reimplements the spirit of the first attempt, but **only**
when `!gs.wasModified()`. In particular, we continue to not write trees
in all the other cases. For example, if `gs.hadCriticalError()` or if
`kvstoreChangedSinceGsCreation` we do not attempt to write the trees.

The problem with the first attempt was that we would **always** write
the trees into the cache, which might mention `NameRef`'s which are in
GlobalState. But if `retainGlobalState` had opted **not** to write
`GlobalState` into the cache, then those `NameRef`s would point to
nowhere.

I'm not sure of **all** the reasons why we were seeing
`retainGlobalState` opt to not cache `GlobalState`, but I don't think it
was just because `!gs.wasModified`. In particular, I know that it's
possible for the issue described in #8166 to have caused the cache to be
poisoned as a result of landing the first attempt (I know this because I
modified `test/cli/stripe-packages-cache` and was able to confirm the
cache crashes when a file has the `ForwardedRestArg` node).

I don't know if there are **other** internal errors that are common that
are causing this, but in any case, this PR leaves it like it was
originally, so that neither GlobalState nor trees are cached when
there was an internal error.

Similarly I don't know if it was the case that we have some flow which
routinely overwrites the cache file from another Sorbet process.
Technically someone could have invoked Sorbet twice at the same time
(e.g. in different terminal windows) but I would find it really
surprising if that were the root cause given how widespread the cache
poisoning was.

### Test plan

~It's hard to write a test for this, because as far as I can tell, the cache only
gets poisoned in cases where there was a critical error.~ And while we could
write a test for the one critical error we're aware that Sorbet is susceptible
too right now (#8166), the moment we fix that bug, the regression test for this
feature would not be useful anymore.

**update**: The way that this was manifesting in practice was not that
we were getting critical errors, but that we were writing `FileRef's`
(via locs, via symbols) into the cache because we were writing to the
cache **after** namer had run. That change was reverted in #8237, with a
test for that case, which should then also cover the behavior changed
here.

As far as changes to existing tests:

The `types.input.files.kvstore.write : 0` lines are because that line is
written only if `cacheTreesAndFiles` is called, which it might not have
before, if `GlobalState` was not written.

In all other cases, the changes now match what the comments in the
output file say should be happening (e.g. the `... (warm cache)` lines
have no misses, and the `... (one cache miss)` lines are once again
accurate).

### DIFF
--- a/main/cache/cache.cc
+++ b/main/cache/cache.cc
@@ -28,30 +28,59 @@ namespace {
  * Returns 'true' if the given GlobalState was originally created from the current contents of
  * kvstore (e.g., kvstore has not since been modified).
  */
-bool kvstoreUnchangedSinceGsCreation(const core::GlobalState &gs, const unique_ptr<OwnedKeyValueStore> &kvstore) {
+bool kvstoreChangedSinceGsCreation(const core::GlobalState &gs, const unique_ptr<OwnedKeyValueStore> &kvstore) {
     const uint8_t *maybeGsBytes = kvstore->read(core::serialize::Serializer::GLOBAL_STATE_KEY).data;
-    const bool storedUidMatches =
-        maybeGsBytes && gs.kvstoreUuid == core::serialize::Serializer::loadGlobalStateUUID(gs, maybeGsBytes);
-    const bool noPreviouslyStoredUuid = !maybeGsBytes && gs.kvstoreUuid == 0;
-    return storedUidMatches || noPreviouslyStoredUuid;
+    if (maybeGsBytes) {
+        // If `GLOBAL_STATE_KEY` is in kvstore but it's not what `gs.kvstoreUuid` contains,
+        // this implies that some other process wrote a different GlobalState to the cache.
+        return gs.kvstoreUuid != core::serialize::Serializer::loadGlobalStateUUID(gs, maybeGsBytes);
+    } else {
+        // `GLOBAL_STATE_KEY` is not in kvstore, which implies the cache is empty.
+        // If kvstoreUuid != 0, that implies we tried to write GlobalState to the cache once before,
+        // and since we don't see `GLOBAL_STATE_KEY` in there now, someone else overwrote/dropped
+        // the cache.
+        return gs.kvstoreUuid != 0;
+    }
 }
 
 /**
- * Writes the GlobalState to kvstore, but only if it was modified. Returns 'true' if a write happens.
+ * Writes the GlobalState to kvstore, but only if it was modified. Returns 'true' if the cache
+ * now contains the contents of the current GlobalState (either because it already contained it, or
+ * it was just written).
  */
 bool retainGlobalState(core::GlobalState &gs, const unique_ptr<OwnedKeyValueStore> &kvstore) {
-    if (kvstore && gs.wasModified() && !gs.hadCriticalError()) {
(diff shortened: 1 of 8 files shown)


froydnj (approved):
Is it possible that we could add some kind of `# issue-critical-error: true` directive in tests so that we had a reliable way of inducing critical errors for testing codepaths where it would matter?  Maybe it's a little harder for CLI tests, because the logic can't live in the test runner, it would have to live in Sorbet itself, which makes things a little circular.

jez:
> Is it possible that we could add some kind of `# issue-critical-error: true` directive in tests so that we had a reliable way of inducing critical errors for testing codepaths where it would matter? Maybe it's a little harder for CLI tests, because the logic can't live in the test runner, it would have to live in Sorbet itself, which makes things a little circular.

So arguably, we already have `--simulate-crash`, which we could use in a CLI test. It works by calling `Exception::raise`, which typically is then rescued and converted into a 1001 error (a critical error).

The reason why we can't use `--simulate-crash` in this case is because we need the crash to happen at a specific time. I'm not sure the best way to go about configuring where the crash would have to happen without adding something like `--simulate-crash-at={x,y,z}` where you can name spots to simulate a crash, which sounds like it would get out of hand 🤔

main/cache/cache.cc
```diff
@@ -28,30 +28,59 @@ namespace {
...
+    if (!gs.wasModified()) {
+        // Optimization: no need to generate a new UUID and write to the cache, because it's already
+        // in the cache
+        return true;
+    }
+
+    gs.kvstoreUuid = Random::uniformU4();
+    kvstore->write(core::serialize::Serializer::GLOBAL_STATE_KEY,
+                   core::serialize::Serializer::storePayloadAndNameTable(gs));
+    return true;
```
elliottt, line 83:
```suggestion
    if (gs.wasModified()) {
        gs.kvstoreUuid = Random::uniformU4();
        kvstore->write(core::serialize::Serializer::GLOBAL_STATE_KEY,
                       core::serialize::Serializer::storePayloadAndNameTable(gs));
    }
    
    return true;
```

jez:
> What do you think about adding the `cache.aborted` stat into the two tests that were modified? That way we could see any of the cases where the poisoning caused `maybeCacheGlobalStateAndFiles` to return a `nullptr`.

Done

elliottt (approved):
Looks great!

elliottt:
It looks like the test failure is due to a cached file not raising the same duplicate type_member error. If you want to commit the test diff and make an issue, I'm happy to take that one on.

jez:
> It looks like the test failure is due to a cached file not raising the same duplicate type_member error. If you want to commit the test diff and make an issue, I'm happy to take that one on.

I'm not sure why this is happening—`rewriter/TypeMember.cc` is using `beginIndexerError`, which should be causing that file to fail to be cached...

PR #8189, opened 2024-09-22
Add option for package directories that are constant names

bhang-stripe:
This PR renames the `--extra-package-files-directory-prefix-slash` option to `--extra-package-files-directory-prefix-slash-deprecated`. Then, `--extra-package-files-directory-prefix-slash` is changed to make it so extra package directories are just the constant name parts of a package's namespace joined by slashes.

That is, if we had a package whose namespace is `SomeModule::MyPackage`, we no longer need the weird directory structure `Some_module/My_package`. We have can have the simpler `SomeModule/MyPackage`.

Afterwards, we'll migrate our extra package directories to use the new naming scheme and delete the deprecated option.

### Motivation

The initial version of slash munging was to add extra directories for packages by using the snake_case version of a package name. However, there was a problem with the snake case, so we had to uppercase the first letter of file/directory names: https://github.com/sorbet/sorbet/pull/6027. This leads to names that are harder to read.

This PR attempts to improve readability of extra package directories.

### Test plan

Added tests


bhang-stripe:

→ https://go/builds/bui_QtmJUYGvknNnMR
→ https://go/builds/bui_QtmJfqne6vS8rl
→ https://go/builds/bui_QtmJpnjcjAyVF7

main/options/options.cc
```diff
@@ -611,10 +611,14 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                  "Extra parent directories which contain package files. These paths use an underscore "
                                  "package-munging convention, i.e. 'Project_Foo'",
                                  cxxopts::value<vector<string>>(), "string");
-    options.add_options(section)("extra-package-files-directory-prefix-slash",
+    options.add_options(section)("extra-package-files-directory-prefix-slash-deprecated",
                                  "Extra parent directories which contain package files. These paths use a slash "
                                  "package-munging convention, i.e. 'project/foo'",
```
jez, line 616: Can this mention why it's deprecated (at least—in favor of what)

main/options/options.cc
```diff
@@ -611,10 +611,14 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
...
                                  cxxopts::value<vector<string>>(), "string");
+    options.add_options(section)("extra-package-files-directory-prefix-slash",
+                                 "Extra parent directories which contain package files. These paths are constant names "
+                                 "joined by slashes, i.e. 'Project/Foo'",
```
jez, line 620: Is it worth changing the example to include a CamelCase name like `Project/FooBar` so that people understand it's not going to need `Project/Foo_Bar` or something like that?

main/options/options.cc
```diff
@@ -611,10 +611,14 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
...
+                                 cxxopts::value<vector<string>>(), "string");
```
jez, line 621: The `"string"` is how the argument is named.

```
      --stripe-packages         Enable support for Stripe's internal Ruby package system
      --stripe-packages-hint-message arg
                                Optional hint message to add to packaging related errors
                                (default: "")
      --extra-package-files-directory-prefix-underscore string
                                Extra parent directories which contain package files.
                                These paths use an underscore package-munging convention,
                                i.e. 'Project_Foo'
      --extra-package-files-directory-prefix-slash string
                                Extra parent directories which contain package files.
                                These paths use a slash package-munging convention, i.e.
                                'project/foo'
      --allow-relaxed-packager-checks-for string
                                Packages which are allowed to ignore the restrictions set
                                by `visible_to` and `export` directives
      --package-skip-rbi-export-enforcement string
                                Constants defined in RBIs in these directories (and no
                                others) can be exported
```

I've been trying to align on `<...>` for the name when the argument is not meant to be literally `--foo=string`. Also maybe a more descriptive name?

```suggestion
                                 cxxopts::value<vector<string>>(), "<path>");
```

If you want to do this in a follow up change (and align all the options with this style), that's fine too.

jez (approved):

PR #8190, opened 2024-09-23
insert in the right place for single line case bodies

neilparikh:
Because we desugared case to a send when `preserveConcreteSyntax` was `true`, we'd always insert outside of the case for single line bodies. To handle this correctly, we can do something similar to what we do for `if`: go through each body and see which one contains the target loc, and insert there. The one difference is that for `if` nodes, we'll always insert in either the `thenp` or the `elsep` case, while for `case`, we may not insert into any of the bodies (for example, if the target is in the condition or one of the patterns). In that case, we'll just fallback to the regular handling of statements and insert just before it.

Doing this requires a slight change in how we desugar `case` when `preserveConcreteSyntax` is `true`. Previously, we did something like: `Magic.caseWhen(condition, pattern 1 for body 1, pattern 2 for body 1, body 1, pattern 1 for body 2, body 2, ...)`. But since each body can have an arbitrary number of patterns, this means there was no way to know which args are the bodies. To handle this, we can instead insert all the patterns first, followed by the bodies, and then add the number of patterns to the args, so we can skip the patterns when looping through. This means something like this: `Magic.caseWhen(condition, number of patterns, pattern 1 for body 1, pattern 2 for body 1, pattern 1 for body 2, ...,  body 1, body 2, ...)`.

### Motivation

Insert in the correct location for single line `case` bodies.

### Test plan

See included automated tests.

### DIFF
--- a/main/lsp/ExtractVariable.cc
+++ b/main/lsp/ExtractVariable.cc
@@ -20,9 +20,9 @@ optional<core::LocOffsets> detectCase(const ast::ExpressionPtr *whereToInsert, c
                                   ast::cast_tree_nonnull<ast::Literal>(send->getPosArg(1)).value)
                                   .value;
             for (int i = numPatterns + 2; i < send->numPosArgs(); i++) {
-                auto *body = &send->getPosArg(i);
-                if (body->loc().exists() && body->loc().contains(target)) {
-                    return body->loc();
+                const auto &body = send->getPosArg(i);
+                if (body.loc().exists() && body.loc().contains(target)) {
+                    return body.loc();
                 }
             }
         }
@@ -32,7 +32,7 @@ optional<core::LocOffsets> detectCase(const ast::ExpressionPtr *whereToInsert, c
 }
 
 core::LocOffsets findWhereToInsert(const ast::ExpressionPtr &scope, const core::LocOffsets target) {
-    const ast::ExpressionPtr *whereToInsert = new ast::ExpressionPtr();
+    const ast::ExpressionPtr *whereToInsert = nullptr;
     if (auto insSeq = ast::cast_tree<ast::InsSeq>(scope)) {
         for (auto &stat : insSeq->stats) {
             if (stat.loc().contains(target)) {
@@ -91,7 +91,12 @@ core::LocOffsets findWhereToInsert(const ast::ExpressionPtr &scope, const core::
     } else {
         ENFORCE(false);
     }
+
     ENFORCE(whereToInsert);
+    if (!whereToInsert) {
+        return core::LocOffsets::none();
+    }
+
     if (auto caseBodyLoc = detectCase(whereToInsert, target)) {
         return *caseBodyLoc;
     }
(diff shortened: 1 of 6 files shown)


main/lsp/ExtractVariable.cc
```diff
@@ -13,66 +13,89 @@ void logDebugInfo(const std::shared_ptr<spdlog::logger> logger, const core::Glob
...
+            }
+        }
+    }
+
+    return nullopt;
+}
+
 core::LocOffsets findWhereToInsert(const ast::ExpressionPtr &scope, const core::LocOffsets target) {
-    core::LocOffsets whereToInsert = core::LocOffsets::none();
+    const ast::ExpressionPtr *whereToInsert = new ast::ExpressionPtr();
```
froydnj, line 35: This is a) unusual code for Sorbet and b) leaks this allocation.  Can we just make this:

```
const ast::ExpressionPtr *whereToInsert = nullptr;
```

and then check if it is still `nullptr` below to determine whether we've found someplace to insert?

main/lsp/ExtractVariable.cc
```diff
@@ -13,66 +13,89 @@ void logDebugInfo(const std::shared_ptr<spdlog::logger> logger, const core::Glob
...
     } else {
         ENFORCE(false);
     }
-    ENFORCE(whereToInsert.exists());
-    return whereToInsert;
+    ENFORCE(whereToInsert);
+    if (auto caseBodyLoc = detectCase(whereToInsert, target)) {
+        return *caseBodyLoc;
+    }
+    return whereToInsert->loc();
```
froydnj, line 98: As written, wouldn't this `whereToInsert->loc()` crash if we didn't find something?  We should be handling that gracefully rather than relying on `ENFORCE`s.

(We'll have the same problem with the `nullptr` change suggested above, it just becomes a little more obvious because there's no loc to associated with `nullptr`.)

main/lsp/ExtractVariable.cc
```diff
@@ -13,66 +13,89 @@ void logDebugInfo(const std::shared_ptr<spdlog::logger> logger, const core::Glob
...
 }
 
+optional<core::LocOffsets> detectCase(const ast::ExpressionPtr *whereToInsert, const core::LocOffsets target) {
+    if (auto send = ast::cast_tree<ast::Send>(*whereToInsert)) {
+        if (send->fun == core::Names::caseWhen()) {
+            int numPatterns = core::cast_type_nonnull<core::IntegerLiteralType>(
+                                  ast::cast_tree_nonnull<ast::Literal>(send->getPosArg(1)).value)
+                                  .value;
+            for (int i = numPatterns + 2; i < send->numPosArgs(); i++) {
+                auto *body = &send->getPosArg(i);
```
froydnj, line 23:
```suggestion
                const auto &body = &send->getPosArg(i);
```

With adjustments for the changed type of `body` below.

froydnj (changes requested):

froydnj (approved):
Thank you!

PR #8191, opened 2024-09-23
Insert in the right place for elsif

neilparikh:
We parse the following:
```ruby
if a
  b
elsif c
  c
end
```
as
```ruby
if a
  b
else
  if c
    c
  end
end
```
This means that when the user extracts the `c` in the `elsif` condition (or all occurrences of `c`), we'll insert inside the fake `else` and produce invalid code, as follows:
```ruby
if a
  b
newVariable = c
elsif newVariable
  c
end
```

To handle this, if the new variable was going to be inserted into an `else` branch, and direct child of the `else` branch is another `if` node, we can assume that it's a desugared form of an `elsif`, and insert outside of the entire `if`.

### Motivation

Handle extracting from `elsif` properly.

### Test plan

See included automated tests.

### DIFF
--- a/test/testdata/lsp/code_actions/extract_variable_multiple/elsif.A.rbedited
+++ b/test/testdata/lsp/code_actions/extract_variable_multiple/elsif.A.rbedited
@@ -42,3 +42,15 @@ def case_3
     4
   end
 end
+
+def case_4_bug
+  if T.unsafe(1)
+    1
+  else
+    if T.unsafe(2)
+#      ^^^^^^^^^^^ apply-code-action: [G] Extract Variable (this occurrence only)
+#      ^^^^^^^^^^^ apply-code-action: [H] Extract Variable (all 2 occurrences)
+      T.unsafe(2)
+    end
+  end
+end
--- a/test/testdata/lsp/code_actions/extract_variable_multiple/elsif.B.rbedited
+++ b/test/testdata/lsp/code_actions/extract_variable_multiple/elsif.B.rbedited
@@ -42,3 +42,15 @@ def case_3
     4
   end
 end
+
+def case_4_bug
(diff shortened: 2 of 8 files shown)


ilyailya (approved):

main/lsp/ExtractVariable.cc
```diff
@@ -90,6 +90,18 @@ class LocSearchWalk {
             return;
         }
 
+        // If the new enclosingScope is an If node, and else branch is another if node, then this is probably an elsif
```
jez, line 93: > probably

Do you have a test that shows that this heuristic breaks if the user actually wrote

```ruby
if cond
else
  if something_to_extra
  end
end
```

?

main/lsp/ExtractVariable.cc
```diff
@@ -127,6 +128,7 @@ class LocSearchWalk {
     // For example, the RHS of the ClassDef doesn't have an ExpressionPtr with a Loc,
     // but enclosingScopeLoc will be a Loc that represents the body of the ClassDef RHS
     // (excluding things like the class name, superclass, and class/end keywords).
+    // TODO(neil): can we get rid of enclosingScopeLoc?
```
jez, line 131: Any more you want to investigate here before landing?

main/lsp/ExtractVariable.cc
neilparikh: Added a new test case for that.

main/lsp/ExtractVariable.cc
neilparikh: No, this was just a idea I had when looking at the code, but I didn't want to spend the ~20 minutes  for trying to refactor and convince myself that the refactor is safe right now, so I decided to add a TODO for the future.

PR #8200, opened 2024-09-27
Fix typo in `procs.md`

vaporyhumo:
There's a typo in the documentation for procs, where `arg1` has type `Arg2Type`.

### Motivation
Just noticed the typo while reading the documentation.

### Test plan
No significant tests are required.


jez (approved):
Thanks!

PR #8203, opened 2024-09-28
Support ForwardedRestArg in array literals

### LINKED ISSUE #8166: Static checker crashes on Array literals with `ForwardedRestArg`

jez:
### Motivation

Fixes #8166

### Commit summary

- **no-op: Don't duplicate what the parser::Splat case already handles** (f370b35f7a)

- **no-op: Fix a comment** (70a86ea284)

- **prework: Use <fwd-args> instead of unique name for restarg** (d5e870936b)

  This

  - avoids allocating a lot of unique names
  - makes the name predictable inside a method body, without having to
    keep track of recursive state

  while still avoiding name collisions, because it seems like

  ```ruby
  def foo(*)
    [[1, 2, 3], [4, 5, 6]].times do |*|
      p(*)
    end
  end

  foo(1, 2, 3)
  ```

  is a syntax error in Ruby (and even if it weren't, I think the behavior
  that you'd expect is that it binds to the innermost `*`, because that's
  what you get with normal shadowing rules from Ruby).

  and is backwards compatible, because there isn't a way in either method
  bodies or in type syntax to get ahold of the unique name that was
  generated for the `*` in the method parameter list, so our choice of
  name is an implementation detail (in the future, I want to allow people
  to write signatures for these methods using syntax like

  ```ruby
  sig { params("*": String).void }
  def foo(*); end
  ```

  but that's a problem for the future, and this gets us closer to being
  able to do that easily.)

- **Support ForwardedRestArg in array literals** (cb062ea14e)

### Test plan

See included automated tests.


neilparikh (approved):

jez:

→ https://go/builds/bui_QxUMpToDNc8CiO
→ https://go/builds/bui_QxUMB2lYWHI1VE
→ https://go/builds/bui_QxUMNJIRSNmyhe

PR #8207, opened 2024-10-02
Parse strict_dependencies in __package.rb

neilparikh:
This PR parses, performs basic validation on, and stores the `strict_dependencies` option in `__package.rb` files. It does not actually validate whether the package is conforming to the strictness level being declared. That will be implemented in a future PR.

### Motivation

Implement more of the packaging system in Sorbet.

### Test plan

See included automated tests.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -1260,11 +1260,11 @@ struct PackageSpecBodyWalk {
 
 private:
     optional<core::packages::StrictDependenciesLevel> parseStrictDependenciesOption(ast::ExpressionPtr &arg) {
-        auto value_ = ast::cast_tree<ast::Literal>(arg);
-        if (!value_ || !value_->isString()) {
+        auto *lit = ast::cast_tree<ast::Literal>(arg);
+        if (!lit || !lit->isString()) {
             return nullopt;
         }
-        auto value = value_->asString();
+        auto value = lit->asString();
 
         if (value == core::Names::false_()) {
             return core::packages::StrictDependenciesLevel::False;
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -1121,7 +1121,7 @@ struct PackageSpecBodyWalk {
                     info.strictDependenciesLevel = make_pair(parsedValue.value(), send.getPosArg(0).loc());
                 } else {
                     if (auto e = ctx.beginError(send.argsLoc(), core::errors::Packager::InvalidStrictDependencies)) {
-                        e.setHeader("Argument to `{}` must be one of: `{}`, `{}`, `{}` or `{}`", send.fun.show(ctx),
+                        e.setHeader("Argument to `{}` must be one of: `{}`, `{}`, `{}`, or `{}`", send.fun.show(ctx),
                                     "'false'", "'layered'", "'layered_dag'", "'dag'");
                     }
                 }
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -209,6 +209,8 @@ public:
     // Whether `visible_to` directives should be ignored for test code
     bool visibleToTests_ = false;
 
+    optional<core::packages::StrictDependenciesLevel> strictDependenciesLevel = nullopt;
+
     // PackageInfoImpl is the only implementation of PackageInfoImpl
     const static PackageInfoImpl &from(const core::packages::PackageInfo &pkg) {
         ENFORCE(pkg.exists());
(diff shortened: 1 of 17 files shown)


neilparikh:

→ https://go/builds/bui_QxBI4poqX2Xugn
→ https://go/builds/bui_QxBI0f9AoHiVwl
→ https://go/builds/bui_QxBIC1NTYApizL

test/testdata/packager/strict_dependencies/missing/__package.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+
+class Missing < PackageSpec
```
neilparikh, line 5: I'd like to add an error for this case too, but there's a bunch of existing tests that need to updated to add the `strict_dependencies` line too, so I'll do it in a follow-up PR, to keep this PR small.

website/docs/error-reference.md
```diff
@@ -862,6 +862,10 @@ you're sure that it should be okay to import this package, then you can add an
 additional `visible_to` directive in order to allow the import you're trying to
 add.
 
+## 3724
+
+> **TODO** This error code is not yet documented.
```
jez, line 867: Let's document this now before we forget.

packager/packager.cc
```diff
@@ -1317,6 +1366,9 @@ void rewritePackageSpec(const core::GlobalState &gs, ast::ParsedFile &package, P
     PackageSpecBodyWalk bodyWalk(info);
     core::Context ctx(gs, core::Symbols::root(), package.file);
     ast::TreeWalk::apply(ctx, bodyWalk, package.tree);
+    if (!info.strictDependenciesLevel) {
```
jez, line 1369:
```suggestion
    if (!info.strictDependenciesLevel.has_value()) {
```

Personal preference, feel free to ignore, I like making comparisons against `nullptr` and `nullopt` explicit, instead of relying on the `operator bool` / `operator !` methods to do the conversion implicitly.

packager/packager.cc
```diff
@@ -1231,6 +1259,27 @@ struct PackageSpecBodyWalk {
...
         illegalNode(ctx, original);
     }
+
+private:
+    bool isValidStrictDependenciesOption(core::NameRef value) {
+        return value == core::Names::false_() || value == core::Names::layered() ||
+               value == core::Names::layeredDag() || value == core::Names::dag();
+    }
+
+    core::packages::StrictDependenciesLevel optionToEnum(core::NameRef value) {
```
jez, line 1269: Feedback that should be obsoleted by my other suggestion to parse instead of validate, but "optionToEnum" is a very broad name—it does not capture that this method's implementation is for the strict dependencies option only.

packager/packager.cc
```diff
@@ -1101,6 +1103,32 @@ struct PackageSpecBodyWalk {
                 info.visibleTo_.emplace_back(getPackageName(ctx, target), core::packages::VisibleToType::Normal);
             }
         }
+
+        if (send.fun == core::Names::strictDependencies()) {
+            if (send.numPosArgs() == 1) {
```
jez, line 1108: As written, this does not ban:

- passing no args
- passing too many args
- passing keyword args
- passing a block

packager/packager.cc
```diff
@@ -209,6 +209,8 @@ class PackageInfoImpl final : public core::packages::PackageInfo {
     // Whether `visible_to` directives should be ignored for test code
     bool visibleToTests_ = false;
 
+    optional<core::packages::StrictDependenciesLevel> strictDependenciesLevel = nullopt;
```
jez, line 212: We are likely to want this to track a `LocOffsets` as well.

You could already make use of it in this PR to `addErrorLine` on the previous definition in the case of multiple definitions.

There are other errors we'll be able to use it for in the future.

Personally, I would store the location of the argument, not the whole send, because we might want to do something like change the level of a package in an autocorrect.

jez (changes requested):
^

packager/packager.cc
neilparikh: The "no args" and "too many args" cases are caught by the [`enterMethod call`](https://github.com/sorbet/sorbet/pull/8207/files#diff-2c71c670c6721566158753a99c725a4826efc01a8a6a3c53a32c98fe3889252cR652), and passing a block is caught [here](https://github.com/sorbet/sorbet/blob/master/packager/packager.cc#L1210C10-L1210C21). I'll add some test cases for those.

Passing keyword args isn't banned right now though, so I'll add a check for that.

jez: I believe it would be the case that keyword args would be banned if you set the type of the parameter to `String` in the enterMethod call.

website/docs/error-reference.md
```diff
@@ -862,6 +862,28 @@ you're sure that it should be okay to import this package, then you can add an
...
+  strict_dependencies 'false'
+end
+```
+
+The 4 possible values are:
+
+- `'false'`: No restrictions are placed on dependencies.
+- `'layered'`
+- `'layered_dag'`
+- `'dag'`
```
neilparikh, line 885: I'll leave no explanations for these values since we don't have any enforcement for them yet. We can add to each as we add enforcement for that option.

packager/packager.cc
```diff
@@ -1231,6 +1257,27 @@ struct PackageSpecBodyWalk {
...
         illegalNode(ctx, original);
     }
+
+private:
+    optional<core::packages::StrictDependenciesLevel> parseStrictDependenciesOption(ast::ExpressionPtr &arg) {
+        auto value_ = ast::cast_tree<ast::Literal>(arg);
+        if (!value_ || !value_->isString()) {
+            return nullopt;
+        }
+        auto value = value_->asString();
```
jez, line 1267:
```suggestion
        auto *lit = ast::cast_tree<ast::Literal>(arg);
        if (!lit || !value_->isString()) {
            return nullopt;
        }
        auto value = lit->asString();
```

Non-blocking: I generally prefer picking different names, rather than trying to approximate shadowing with `_`-suffixed names.

Also non-blocking: I prefer using `*` in the `auto` when it's a pointer, I forget why, it was something I picked up from trevor, he had a good explanation at the time and I just believed him.

packager/packager.cc
```diff
@@ -1101,6 +1103,30 @@ struct PackageSpecBodyWalk {
...
+                return;
+            }
+
+            if (send.numPosArgs() > 0) {
+                auto parsedValue = parseStrictDependenciesOption(send.getPosArg(0));
+                if (parsedValue.has_value()) {
+                    info.strictDependenciesLevel = make_pair(parsedValue.value(), send.getPosArg(0).loc());
+                } else {
+                    if (auto e = ctx.beginError(send.argsLoc(), core::errors::Packager::InvalidStrictDependencies)) {
+                        e.setHeader("Argument to `{}` must be one of: `{}`, `{}`, `{}` or `{}`", send.fun.show(ctx),
```
jez, line 1124:
```suggestion
                        e.setHeader("Argument to `{}` must be one of: `{}`, `{}`, `{}`, or `{}`", send.fun.show(ctx),
```

jez (approved):

neilparikh:

→ https://go/builds/bui_Qxnea5uvXLd9R4
→ https://go/builds/bui_QxneUx1ZirDY5l
→ https://go/builds/bui_QxneBxwZJmiTFr

PR #8210, opened 2024-10-02
sendTp -> returnTypeBeforeSolve

jez:
### Motivation

Just found the old name a little weird.

In the process of trying to figure out what `sendTp` meant, I figured out a
better name for it, and figured I'd use that name and add some comments.

### Test plan

See included automated tests.


core/types/calls.cc
```diff
@@ -2656,6 +2662,7 @@ class Magic_callWithBlock : public IntrinsicMethod {
             auto it = &dispatched;
             while (it != nullptr) {
                 if (it->main.method.exists()) {
+                    // TODO(jez) This only looks at the main component!
```
jez, line 2665: These are all bugs.

Will file issues separately.

ilyailya (approved):

PR #8214, opened 2024-10-03
Ignore representation of enums as classes in the document outline

### LINKED ISSUE #7099: Document outline leaks internal representation of enums as classes
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20MyEnum%20%3C%20T%3A%3AEnum%0A%20%20enums%20do%0A%20%20%20%20X%20%3D%20new%0A%20%20%20%20Y%20%3D%20new%0A%20%20end%0Aend)

```ruby
# typed: true

class MyEnum < T::Enum
  enums do
    X = new
    Y = new
  end
end
```

#### Observed output

```
No errors! Great job.
```

Not visible in the sorbet.run, but if you view the document symbols for this file, you'll see something like this:

image

The enum values show up twice, once as a constant and once as a class.

The class is only there because Sorbet internally represents the type of an enum value as an instance of a class with a magic name like `X$1` (and then the `$1` is hidden in the document outline printing).

#### Expected behavior

These classes do not actually exist, and only make it harder to see what enums values there are in the outline view (because the list as twice as long as it needs to be, thus taking up twice the screen space).

We should hide them in the document outline view.

- - -

komarovb:
A small change that hides the internal representation of enums as classes from the document outline view.

**Before**
Before

**After**
After

### Motivation

Fixes #7099.

### Test plan

See included automated tests. 
Please let me know if more sophisticated tests are necessary.

### DIFF
--- a/main/lsp/lsp_helpers.cc
+++ b/main/lsp/lsp_helpers.cc
@@ -36,6 +36,10 @@ bool hideSymbol(const core::GlobalState &gs, core::SymbolRef sym) {
     if (name == core::Names::beforeAngles()) {
         return true;
     }
+    // internal representation of enums as classes
+    if (sym.isClassOrModule() && sym.asClassOrModuleRef().data(gs)->name.isTEnumName(gs)) {
+        return true;
+    }
     // static-init for a file
     if (name.kind() == core::NameKind::UNIQUE && name.dataUnique(gs)->original == core::Names::staticInit()) {
         return true;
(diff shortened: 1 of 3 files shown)


jez (approved):
Thanks!

komarovb:
It looks like `build-emscripten.sh` is failing. The errors in the log file don't appear to be related to the code changes. Should this step be reattempted? Please let me know if there's anything else I need to do.

froydnj:
> It looks like `build-emscripten.sh` is failing. The errors in the log file don't appear to be related to the code changes. Should this step be reattempted? Please let me know if there's anything else I need to do.

I retried it for you, and it looks like it passed!  🎉

PR #8222, opened 2024-10-07
Specify sample return type based on args

HParker:
Right now, Sorbet has the return type of `[].sample` as either an array or an Elem, However the array is only possible if you pass an arg to sample

The types are still inaccurate for the `[].sample(random: Random.new)` case, but that is hopefully less common.

### Motivation

More accurate type information

### Test plan

ran `tools/scripts/update_testdata_exp.sh` and found no changes to apply.

### DIFF
--- a/rbi/core/array.rbi
+++ b/rbi/core/array.rbi
@@ -2266,7 +2266,10 @@ class Array < Object
         arg0: Integer,
         random: Random::Formatter,
     )
-    .returns(T.any(T.nilable(Elem), T::Array[Elem]))
+    .returns(T::Array[Elem])
+  end
+  sig do
+    returns(Elem)
   end
   def sample(arg0=T.unsafe(nil), random: T.unsafe(nil)); end
 


jez:
Hey, unfortunately this doesn't work, because of #37.

I have opened a PR to say as much, because half a dozen people have attempted to make this PR in the past: #8223

Thanks and sorry to have wasted your time.

marknuzz:
This seems to already be working the way the OP expected in any case, maybe there's a bad sig somewhere in the OP's codebase. See [sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0AT.reveal_type%28%5B%5D.sample%29%0A%0AT.reveal_type%28%5B1%5D.sample%29%0A%0AT.reveal_type%28%5B1%5D.sample%281%29%29)

HParker:
Thanks @jez Sorry, I tried to search the history, but failed to find that. ❤️

jez:
@marknuzz the difference between your example and @HParker's example is that `[]` has type `[] (0-tuple)` and `[1]` has type `[Integer] (1-tuple)`. It is likely not a problem with HParker's codebase.

Sorbet has special-case support for `Array#sample` on tuple types, [see here](https://github.com/sorbet/sorbet/blob/master/core/types/calls.cc#L3338-L3364).

HParker:
Yes, @jez is correct. My examples are confused by my use of empty arrays. I think my examples should have been,

`[1,2,3].sample # => 1 # type: Array, or Elem (in this case Integer), but actually can only be Elem (Integer)` and

`[1,2,3].sample(2) # => [3,1] # type: Array, or Elem, but can only actually be Array`

PR #8224, opened 2024-10-08
Parse layer in __package.rb

neilparikh:
#8207 but for `layer`

This PR parses, performs basic validation on, and stores the `layer` option in `__package.rb` files. It does not actually validate whether the package has any layering violations. That will be implemented in a future PR.

### Motivation

Implement more of the packaging system in Sorbet.

### Test plan

See included automated tests.


main/options/options.cc
```diff
@@ -627,6 +627,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                  "Packages which are allowed to ignore the restrictions set by `visible_to` "
                                  "and `export` directives",
                                  cxxopts::value<vector<string>>(), "<name>");
+    options.add_options(section)("packager-layer", "Valid layer names for packages.",
```
neilparikh, line 630: Not a huge fan of this flag name, but couldn't think of anything better, so open to suggestions.

main/options/options.cc
```diff
@@ -1199,6 +1201,24 @@ void readOptions(Options &opts,
...
 
+        if (raw.count("packager-layer") && !opts.stripePackages) {
+            // Default values are available at raw[...], but don't show up in raw.count(...),
+            // so we can use this as a way to check if the user passed in the flag
+            logger->error("--packager-layer can only be specified in --stripe-packages mode");
+            throw EarlyReturnWithCode(1);
+        }
+
+        if (opts.stripePackages) {
+            std::regex layerValid("[a-zA-Z0-9]+");
```
neilparikh, line 1212: I picked the alphanumeric on a whim, open to feedback about being more or less restrictive here.

core/GlobalState.cc
```diff
@@ -2396,14 +2401,16 @@ void GlobalState::setPackagerOptions(const std::vector<std::string> &extraPackag
...
+                                     const std::vector<std::string> &packagerLayers, std::string errorHint) {
     ENFORCE_NO_TIMER(!packageDB_.frozen);
 
     packageDB_.enabled_ = true;
     packageDB_.extraPackageFilesDirectoryUnderscorePrefixes_ = extraPackageFilesDirectoryUnderscorePrefixes;
     packageDB_.extraPackageFilesDirectorySlashDeprecatedPrefixes_ = extraPackageFilesDirectorySlashDeprecatedPrefixes;
     packageDB_.extraPackageFilesDirectorySlashPrefixes_ = extraPackageFilesDirectorySlashPrefixes;
     packageDB_.skipRBIExportEnforcementDirs_ = packageSkipRBIExportEnforcementDirs;
+    absl::c_transform(packagerLayers, std::back_inserter(packageDB_.layers_),
+                      [this](string layer) { return enterNameUTF8(layer); });
```
jez, line 2413: We can take this by `const string &` or `string_view`—`enterNameUTF8` does not take ownership of `layer`, so there's no need to copy.

jez: It's also legal to use `const auto&` here.

core/packages/PackageDB.h
```diff
@@ -63,6 +63,7 @@ class PackageDB final {
     const std::vector<std::string> &extraPackageFilesDirectorySlashDeprecatedPrefixes() const;
     const std::vector<std::string> &extraPackageFilesDirectorySlashPrefixes() const;
     const std::vector<std::string> &skipRBIExportEnforcementDirs() const;
+    const std::vector<core::NameRef> &layers() const;
```
jez, line 66: Let's maybe mention here that this the sort order matters in a doc comment?

e.g. let's document whether earlier layers are allowed to depend on later, or vice versa.

main/options/options.cc
jez: What do you not like about the current name?

jez, line 630: Let's be sure to also mention here that the order matters, and what that order is.

main/options/options.cc
```diff
@@ -627,6 +627,8 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                  "Packages which are allowed to ignore the restrictions set by `visible_to` "
                                  "and `export` directives",
                                  cxxopts::value<vector<string>>(), "<name>");
+    options.add_options(section)("packager-layer", "Valid layer names for packages.",
+                                 cxxopts::value<vector<string>>()->default_value("library,application"), "<layerName>");
```
jez, line 631:
```suggestion
                                 cxxopts::value<vector<string>>()->default_value("library,application"), "<layer-name>");
```

This is the case that we tend to use in the `--help` output.

main/options/options.cc
jez, line 630: We might want to mention something in our help output somewhere about how "repeatable" / `vector`-typed options work (i.e., that they can be provided as `--arg 1 --arg 2 --arg 3` or `--arg=1,2,3`)

That's not unique to this option—there are multiple such options like this, and we can do that in a future change. Probably for now, it's enough that the default shows `library,application` to help people understand that comma-separated layer names will be parsed.

main/options/options.cc
```diff
@@ -1199,6 +1201,24 @@ void readOptions(Options &opts,
             }
         }
 
+        if (raw.count("packager-layer") && !opts.stripePackages) {
+            // Default values are available at raw[...], but don't show up in raw.count(...),
+            // so we can use this as a way to check if the user passed in the flag
+            logger->error("--packager-layer can only be specified in --stripe-packages mode");
+            throw EarlyReturnWithCode(1);
+        }
```
jez, line 1209: Doesn't this imply that the default value is never actually used?

If you pass `--stripe-packages` and don't pass `--package-layer`, doesn't this code raise an error?

main/options/options.cc
jez: Maybe put this in a comment?
What does the Stripe-internal implementation do?

main/options/options.cc
jez: Ah I guess not, I see what your comment says now.

packager/packager.cc
```diff
@@ -1259,6 +1283,19 @@ struct PackageSpecBodyWalk {
...
+        auto &validLayers = gs.packageDB().layers();
+        ENFORCE(validLayers.size() > 0);
+        if (validLayers.size() == 1) {
+            return string(validLayers.front().shortName(gs));
+        }
+        string result = "";
+        for (int i = 0; i < validLayers.size() - 1; i++) {
+            result += fmt::format("{}, ", validLayers[i].shortName(gs));
+        }
+        result += fmt::format("or {}", validLayers.back().shortName(gs));
```
jez, line 1296: We might want to use `ErrorColors::format` here to get colorized formatting for the layer names.

packager/packager.cc
```diff
@@ -1259,6 +1283,19 @@ struct PackageSpecBodyWalk {
...
 private:
+    string buildValidLayersStr(const core::GlobalState &gs) {
+        auto &validLayers = gs.packageDB().layers();
+        ENFORCE(validLayers.size() > 0);
+        if (validLayers.size() == 1) {
+            return string(validLayers.front().shortName(gs));
+        }
+        string result = "";
+        for (int i = 0; i < validLayers.size() - 1; i++) {
+            result += fmt::format("{}, ", validLayers[i].shortName(gs));
```
jez, line 1294: What do you think about only putting this comma here if there are 3 or more elements, so that the default case of `library, or application` just prints like

```
`library` or `application`
```

?

test/testdata/packager/layer/invalid_option/__package.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+# typed: strict
+# enable-packager: true
+# packager-layers: a, b
```
jez, line 4: out of curiosity, where does the code ignore the leading space in `, `?

website/docs/error-reference.md
```diff
@@ -884,6 +884,35 @@ The 4 possible values are:
...
+```ruby
+class MyPackage < PackageSpec
+  strict_dependencies 'false'
+  layer 'library'
+end
+```
+
+You can choose the valid layers using the `--packager-layer` command line flag.
+For example, the following specifies that there are three valid layers: `util`,
+`lib` and `app`.
```
jez, line 907: Let's also document the significance of the ordering here.

jez: or at least leave a TODO to do so.

jez (commented):
mostly lgtm

main/options/options.cc
neilparikh: The regex in the stripe version is `[-A-Za-z0-9_/.]+`, but there’s also a fixed list of valid layers that are all alphabetic, so the regex is (sort of) irrelevant

main/options/options.cc
neilparikh: Added a TODO item to do so.

test/testdata/packager/layer/invalid_option/__package.rb
neilparikh: The `trim_string` call here: https://github.com/sorbet/sorbet/blob/master/test/helpers/position_assertions.cc#L2331

main/options/options.cc
neilparikh: I think it's mainly that it doesn't really convey that these are the _possible_ layers, and not the layer being assigned to a package. Maybe `--packager-possible-layer` or `--packager-valid-layer` is better?

It's also a bit weird to me that it's singular given that it is called like `--packager-layer=a,b,c` (ie. with a list of items), but adding an `s` at the name might be weird when called in the repeated form: `--packager-layers a --packager-layers b`

packager/packager.cc
```diff
@@ -1127,6 +1128,29 @@ struct PackageSpecBodyWalk {
                 }
             }
         }
+
+        if (send.fun == core::Names::layer()) {
+            if (info.layer.has_value()) {
+                if (auto e = ctx.beginError(send.loc, core::errors::Packager::InvalidLayer)) {
+                    e.setHeader("Repeated declaration of `{}`", send.fun.show(ctx));
+                    e.addErrorLine(ctx.locAt(info.layer.value().second), "Previous declaration found here");
```
jez, line 1136:
```suggestion
                    e.addErrorLine(ctx.locAt(info.layer.value().second), "Previously declared here");
```

A little consistence with other errors. We only use `"declaration found here"` once but we use `"declared here"`/`"defined here"` many times.

packager/packager.cc
```diff
@@ -1259,6 +1283,23 @@ struct PackageSpecBodyWalk {
...
+        for (int i = 0; i < validLayers.size() - 1; i++) {
+            if (validLayers.size() > 2) {
+                result += core::ErrorColors::format("`{}`, ", validLayers[i].shortName(gs));
+            } else {
+                result += core::ErrorColors::format("`{}` ", validLayers[i].shortName(gs));
+            }
+        }
+        result += core::ErrorColors::format("or `{}`", validLayers.back().shortName(gs));
+        return result;
+    }
```
jez, line 1302: Minor, not blocking: this doesn't use any members of the enclosing class—it could be a top-level helper function instead of a private method if you wanted it to be.

main/options/options.cc
```diff
@@ -627,6 +629,9 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                  "Packages which are allowed to ignore the restrictions set by `visible_to` "
                                  "and `export` directives",
                                  cxxopts::value<vector<string>>(), "<name>");
+    options.add_options(section)("packager-layer", "Valid layer names for packages, ordered lowest to highest.",
+                                 cxxopts::value<vector<string>>()->default_value("library,application"),
+                                 "<layer-name>");
```
jez, line 634: Do we expect it to be more common for people to pass layers by repeatedly calling this option, or by passing a `,`-separated list?

I'm expecting that people will opt for the `,`-separated list, and thus a name of `packager-layers` (plural) will read better.

What do you think? Another option would be to think of a name that's somewhat agnostic of whether it's plural or not, but I can't think of any.

I'd like to spend the time to think about this now so we don't regret a name in the future and have to change it.

jez (approved):
mostly looks good to me but I have a question about what the option should be named before we merge.

main/options/options.cc
neilparikh: I agree, I think the `,` separated list will probably be more common, so plural sounds better.

Do you have any thoughts on this part?

> it doesn't really convey that these are the possible layers, and not the layer being assigned to a package. Maybe --packager-possible-layers or --packager-valid-layers is better?

neilparikh:

→ https://go/builds/bui_R2LFDXOWCEQ54R
→ https://go/builds/bui_R2LFxFNM3IGfiR
→ https://go/builds/bui_R2LFp08kGr4JNS

neilparikh:

→ https://go/builds/bui_R2kPSoiwld0ppn
→ https://go/builds/bui_R2kPavryrbS5Qf
→ https://go/builds/bui_R2kPs5HWdMTAIn

PR #8225, opened 2024-10-08
Clean up location handling

amomchilov:
### Motivation

This PR does a sweeping cleanup (heh, get it? 🧹) to the way we handle location information. This massively reduces boilerplate, and allows a lot of code to get simplified.

You can see the commits for details, but they change the same lines several times as it gets progressively refactored. But in short, this PR:

1. Makes it so the location is accessed out of the uncasted node, so that it's done once, rather than repeated in every `case` statement.
2. Changed the `translateLocation` helper to take a `pm_location_t` by-value instead of by-pointer. A `pm_location_t` just contains a pair of pointers, so there's no performance concern here. In fact, the parameters to `translateLocation` can fit within the registers, without even spilling into the stack.

### Test plan

This is a pure refactor, with no behavioural changes. Everything here is thoroughly tested by our translation and location tests.


amomchilov:
Sigh, wrong repo, please ignore :)

PR #8226, opened 2024-10-10
Add rbi definition for `Enumerator#+`

devosnw:
### Motivation

* Was added in Ruby 2.7.0
* https://ruby-doc.org/core-2.7.0/Enumerator.html#method-i-2B

### Test plan

I am not familiar with this repo at all. If I need to update some tests somewhere please let me know!

### DIFF
--- a/rbi/core/enumerator.rbi
+++ b/rbi/core/enumerator.rbi
@@ -154,6 +154,18 @@ class Enumerator < Object
   sig { params(initial: T.untyped, block: T.proc.params(arg: T.untyped).void).returns(T::Enumerator[T.untyped]) }
   def self.produce(initial = nil, &block); end
 
+  # Returns an enumerator object generated from this enumerator and a given 
+  # enumerable.
+  #
+  # ### Examples
+  #
+  # ```ruby
+  # e = (1..3).each + [4, 5]
+  # e.to_a #=> [1, 2, 3, 4, 5]
+  # ```
+  sig { params(enum: T::Enumerable[T.untyped]).returns(T::Enumerator::Chain[T.untyped]) }
+  def +(enum); end
+
   # Iterates over the block according to how this
   # [`Enumerator`](https://docs.ruby-lang.org/en/2.7.0/Enumerator.html) was
   # constructed. If no block and no arguments are given, returns self.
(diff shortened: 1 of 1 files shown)


rbi/core/enumerator.rbi
```diff
@@ -154,6 +154,18 @@ class Enumerator < Object
...
+  # Returns an enumerator object generated from this enumerator and a given 
+  # enumerable.
+  #
+  # ### Examples
+  #
+  # ```ruby
+  # e = (1..3).each + [4, 5]
+  # e.to_a #=> [1, 2, 3, 4, 5]
+  # ```
+  sig { params(enum: T::Enumerable[T.untyped]).returns(T::Enumerator::Chain[T.untyped]) }
```
jez, line 166: 👋 Why does this not use generics to avoid depending on `T.untyped`?

jez (changes requested):
^ question

rbi/core/enumerator.rbi
devosnw: Generics really throw me every time I have to work with them but from how I understand it the resulting chain would look something like

```suggestion
  sig { type_parameters(:T).params(enum: T::Enumerable[type_parameter(:T)]).returns(T::Enumerator::Chain[T.any(type_parameter(:T), self::Elem) }
```

What would be the correct way to write this?

jez: yep, I believe so

jez: You can omit the `self::`, just `Elem` is fine.

devosnw: done! https://github.com/sorbet/sorbet/pull/8226/commits/5150cd820814775b0bbde29dec730a447cfbb2bd

neilparikh:

→ https://go/builds/bui_R0sUDpfeBncnJk
→ https://go/builds/bui_R0sUA6I6yn8kmy
→ https://go/builds/bui_R0sUQh9cW5UwnX

jez:

→ https://go/builds/bui_R2mgSnOK94bcvx
→ https://go/builds/bui_R2mgw0A5432PZy
→ https://go/builds/bui_R2mgmlWET2yN33

jez:
Sorry for the delay. There was a problematic commit that had landed on Sorbet master that needed to be fixed, which made it look like your change introduced failures. We've resolved that, re-tested, and things look good.

Thanks for the ping and thanks for the contribution.

jez (approved):

Issue #8206, opened 2024-10-01
Intrinsic for `Array#zip` is wrong when given a block


jez:
#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28xs%3A%20T%3A%3AArray%5BInteger%5D%29.void%20%7D%0Adef%20example%28xs%29%0A%20%20res%20%3D%20xs.zip%28xs%29%20do%20%7Czipped%7C%0A%20%20%20%20T.reveal_type%28zipped%29%0A%20%20end%0A%20%20T.reveal_type%28res%29%0Aend)

```ruby
# typed: true
extend T::Sig

sig { params(xs: T::Array[Integer]).void }
def example(xs)
  res = xs.zip(xs) do |zipped|
    T.reveal_type(zipped)
  end
  T.reveal_type(res)
end
```

#### Observed output

```
editor.rb:7: Revealed type: Integer https://srb.help/7014
     7 |    T.reveal_type(zipped)
            ^^^^^^^^^^^^^^^^^^^^^
  Got Integer originating from:
    editor.rb:6:
     6 |  res = xs.zip(xs) do |zipped|
                               ^^^^^^

editor.rb:9: Revealed type: T::Array[[Integer, T.nilable(Integer)]] https://srb.help/7014
     9 |  T.reveal_type(res)
          ^^^^^^^^^^^^^^^^^^
  Got T::Array[[Integer, T.nilable(Integer)]] originating from:
    editor.rb:6:
     6 |  res = xs.zip(xs) do |zipped|
     7 |    T.reveal_type(zipped)
     8 |  end
Errors: 2
```

#### Expected behavior

- `zipped` should have type `[Integer, T.nilable(Integer)] (2-tuple)`
- `res` should have type `nil`

https://docs.ruby-lang.org/en/3.1/Array.html#method-i-zip

- - -

ilyailya:
Fixed in https://github.com/sorbet/sorbet/pull/8229

PR #8229, opened 2024-10-11
Update type signature for Array#zip

ilyailya:
The PR updates the block parameter type in the RBI and the return type in the intrinsic for the Array#zip to match the type in the docs
https://docs.ruby-lang.org/en/3.1/Array.html#method-i-zip

### Motivation

https://github.com/sorbet/sorbet/issues/8206

### Test plan

See included automated tests.

### DIFF
--- a/core/types/calls.cc
+++ b/core/types/calls.cc
@@ -4097,26 +4097,31 @@ public:
         ENFORCE(!ap->targs.empty());
         unwrappedElems.emplace_back(ap->targs.front());
 
+        for (auto arg : args.args) {
+            auto argTyp = arg->type;
+            if (auto *ap = cast_type<AppliedType>(argTyp)) {
+                ENFORCE(ap->klass == Symbols::Enumerable() ||
+                        ap->klass.data(gs)->derivesFrom(gs, Symbols::Enumerable()));
+                ENFORCE(!ap->targs.empty());
+                unwrappedElems.emplace_back(Types::any(gs, ap->targs.front(), Types::nilClass()));
+            } else if (auto *tuple = cast_type<TupleType>(argTyp)) {
+                unwrappedElems.emplace_back(Types::any(gs, tuple->elementType(gs), Types::nilClass()));
+            } else {
+                // Arg type didn't match; we already reported an error for the arg type; just return untyped to
+                // recover.
+                res.returnType = Types::untypedUntracked();
+                return;
+            }
+        }
+
         if (args.block != nullptr) {
             res.returnType = Types::nilClass();
-        } else {
-            for (auto arg : args.args) {
-                auto argTyp = arg->type;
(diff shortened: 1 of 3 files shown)


ilyailya:

→ https://go/builds/bui_R0sQwYqHkd4WJB
→ https://go/builds/bui_R0sQ1rf8MhIpka
→ https://go/builds/bui_R0sQMQsenoidk1

rbi/core/array.rbi
```diff
@@ -2799,7 +2799,7 @@ class Array < Object
   sig do
     type_parameters(:U).params(
         arg: T::Enumerable[T.type_parameter(:U)],
-        blk: T.proc.params(x: Elem, y: T.nilable(T.type_parameter(:U))).void
+        blk: T.proc.params(x: [Elem, T.nilable(T.type_parameter(:U))]).void
```
jez, line 2802: What's going on here?

jez, line 2802: I don't think that this solves the actual problem either—if you do

```ruby
xs = T.let([1,2,3], T::Array[Integer])
res = xs.zip(xs, xs) do |zipped|
  T.reveal_type(zipped)
end
```

Now `zipped` shows as

```
[Integer, T.nilable(Integer)] (2-tuple)
```

despite the correct type being

```
[Integer, T.nilable(Integer), T.nilable(Integer)] (3-tuple)
```

I would recommend that we split up the change to change the result type to `nil` from the change attempting to fix the argument type yielded to the block.

Off the top of my head, I'm not aware of any current intrinsics which attempt to model this distinction.

The easiest way to work around this in a way that's at least correct if unsatisfying would be to change the RBI like this


```suggestion
        blk: T.proc.params(x: T::Array[T.any(Elem, T.nilable(T.type_parameter(:U)))]).void
```

Possibly with an overload for the case where there is exactly one positional argument to annotate that the block will be yielded exactly two values.

jez (changes requested):
^

rbi/core/array.rbi
jez: realized the problem. see below.

core/types/calls.cc
```diff
@@ -4107,13 +4107,23 @@ class Array_zip : public IntrinsicMethod {
...
                 return;
             }
         }
 
-        res.returnType = Types::arrayOf(gs, make_type<TupleType>(move(unwrappedElems)));
+        if (args.block != nullptr) {
+            res.returnType = Types::nilClass();
+
+            if (auto *blockAppliedType = cast_type<AppliedType>(res.main.blockPreType)) {
+                ENFORCE(blockAppliedType->targs.size() == 2);
```
jez, line 4121:
```suggestion
                ENFORCE(blockAppliedType->targs.size() == 2, "calls.cc out of date w.r.t. array.rbi");
```

jez (approved):
Great find

PR #8230, opened 2024-10-11
Revert "Remove rubyRegionId"

jez:
Reverts sorbet/sorbet#8213

This turned out to be load bearing.

#### Input

[→ View on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20returns%28Integer%29%20%7D%0Adef%20returns_int%0A%20%20raise%20StandardError.new%0Aend%0A%0Abegin%0A%20%20ticket%20%3D%20returns_int%0Arescue%20ArgumentError%0A%20%20ticket%20%3D%20returns_int%0Aend%0A%0Aticket%26.even%3F%0A)

```ruby
# typed: true
extend T::Sig

sig { returns(Integer) }
def returns_int
  raise StandardError.new
end

begin
  ticket = returns_int
rescue ArgumentError
  ticket = returns_int
end

ticket&.even?

```

#### Observed output

```
editor.rb:15: Used &. operator on Integer, which can never be nil https://srb.help/7034
    15 |ticket&.even?
              ^^
  Got Integer originating from:
    editor.rb:10:
    10 |  ticket = returns_int
                   ^^^^^^^^^^^
    editor.rb:12:
    12 |  ticket = returns_int
                   ^^^^^^^^^^^
  Autocorrect: Use -a to autocorrect
    editor.rb:15: Replace with .
    15 |ticket&.even?
              ^^
Errors: 1
```

#### Expected behavior

- - -

PR #8231, opened 2024-10-14
Reapply "Remove rubyRegionId" (#8230)

jez:
### Motivation

This was introduced in #8213, then reverted in #8230 because it introduced errors that were blocking us from bumping Sorbet on Stripe's codebase.

I thought that meant that those errors were regressions, but they're actually improvements. I audited the ~30 new errors that the original PR introduced on Stripe's codebase, and in every case, Sorbet was able to more accurately model what would happen at runtime.

Thus, we should be safe to reapply this change.

### Test plan

I've included a new, non-exp test case that captures the behavior changes that result from removing `rubyRegionId`. Had this test existed before the original PR was made, it would have alerted us that this is a PR that improves Sorbet's support for modeling rescue-based control-flow heavy code.

### DIFF
--- a/test/testdata/infer/begin_rescue_uninitialized.rb
+++ b/test/testdata/infer/begin_rescue_uninitialized.rb
@@ -1,0 +1,24 @@
+# typed: true
+extend T::Sig
+
+sig { returns(String) }
+def returns_string = ""
+
+sig { returns(T.nilable(String)) }
+def returns_nilable_string = nil
+
+begin
+  x = returns_string
+rescue ArgumentError
+  begin
+    x = returns_string
+  rescue ArgumentError
+    raise
+  end
+end
+
+T.reveal_type(x) # error: `String`
+
+x&.chars || []
+#^^ error: can never be nil
+#           ^^ error: This code is unreachable
--- a/test/testdata/infer/begin_rescue_uninitialized.rb
+++ b/test/testdata/infer/begin_rescue_uninitialized.rb
@@ -4,21 +4,21 @@ extend T::Sig
 sig { returns(String) }
 def returns_string = ""
 
-sig { returns(T.nilable(String)) }
-def returns_nilable_string = nil
-
-begin
-  x = returns_string
-rescue ArgumentError
+1.times do
   begin
     x = returns_string
   rescue ArgumentError
-    raise
+    begin
+      x = returns_string
+    rescue ArgumentError
+      raise
+    end
   end
+
+  T.reveal_type(x) # error: `String`
+
+  x&.chars || []
+  #^^ error: can never be nil
+  #           ^^ error: This code is unreachable
 end
 
-T.reveal_type(x) # error: `String`
-
-x&.chars || []
-#^^ error: can never be nil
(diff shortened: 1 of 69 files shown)


jez:

→ https://go/builds/bui_R1zLlpt9x1KHaQ
→ https://go/builds/bui_R1zLI2zrGITXVI
→ https://go/builds/bui_R1zLmsgtAhMHtC

test/testdata/infer/begin_rescue_uninitialized.rb
jez, line 1: This is a new test.

test/testdata/infer/begin_rescue_uninitialized.rb
```diff
@@ -0,0 +1,24 @@
+# typed: true
+extend T::Sig
+
+sig { returns(String) }
+def returns_string = ""
+
+sig { returns(T.nilable(String)) }
+def returns_nilable_string = nil
+
```
neilparikh, line 9: Method is unused.

```suggestion
```

test/testdata/lsp/lsp_hover_crash.rb
```diff
@@ -0,0 +1,10 @@
+# typed: strict
```
neilparikh, line 1: Was this added accidentally?

test/testdata/infer/begin_rescue_uninitialized.rb
neilparikh: This test  has the same behaviour on [sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20returns%28String%29%20%7D%0Adef%20returns_string%20%3D%20%22%22%0A%0Asig%20%7B%20returns%28T.nilable%28String%29%29%20%7D%0Adef%20returns_nilable_string%20%3D%20nil%0A%0Abegin%0A%20%20x%20%3D%20returns_string%0Arescue%20ArgumentError%0A%20%20begin%0A%20%20%20%20x%20%3D%20returns_string%0A%20%20rescue%20ArgumentError%0A%20%20%20%20raise%0A%20%20end%0Aend%0A%0AT.reveal_type%28x%29%20%23%20error%3A%20%60String%60%0A%0Ax%26.chars%20%7C%7C%20%5B%5D%0A%23%5E%5E%20error%3A%20can%20never%20be%20nil%0A%23%20%20%20%20%20%20%20%20%20%20%20%5E%5E%20error%3A%20This%20code%20is%20unreachable), so I don’t think it’s capturing the new behaviour.

test/testdata/lsp/lsp_hover_crash.rb
jez: it was—fixed.

test/testdata/infer/begin_rescue_uninitialized.rb
jez: Nice catch.

It turns out that it has to be inside a loop for it to reproduce.

I've pushed a change to wrap it in a `1.times { ... }` block.

neilparikh (approved):

PR #8030, opened 2024-07-11
Don't close cache file after caching `__package.rb` files

jez:
### Motivation

We had silently disabled the cache in #7759 because I didn't notice that the
`kvstore` is getting moved into the call to `maybeCacheGlobalStateAndFiles` and
thus is `nullptr` the next time we try to use it for caching ASTs.

### Test plan

- [x] Wrote a CLI test
- [ ] Tested on pay-server to see if it gets faster

PR #8237, opened 2024-10-16
Semantically revert "Hoist pipeline::name to after pipeline::package (#7817)"

jez:
### Motivation

In practice, this was never actually working because the problem which we fixed
in #8030 was preventing the cache from being written a second time (and we were
only ever caching `__package.rb` files).

When we fixed that, it exposed that it's not actually safe to call
`pipeline::name` before writing to the cache, because it will mean writing way
too much of `GlobalState` into the cache that shouldn't have been. The disk
cache is only ever meant to cache symbols from Sorbet's payload.

It's not clear to me why that is, and if that is, why it isn't implemented by
simply re-reading the payload out of Sorbet's binary every time. In any case, I
plan to continue looking at that further in an attempt to move `pipeline::name`
back to where it was—we need `pipeline::name` to be before indexing all files,
because in the future we want to make it so that we can decide not to index
files that we don't need. We're not at that point yet, so having a working disk
cache is more important.

### Test plan

The included test has an `ENFORCE` failure when it runs on `master`, and is
fixed in this change.

### DIFF
--- a/main/realmain.cc
+++ b/main/realmain.cc
@@ -749,10 +749,7 @@ int realmain(int argc, char *argv[]) {
                 // First run: only the __package.rb files. This populates the packageDB
                 pipeline::setPackagerOptions(*gs, opts);
                 pipeline::package(*gs, absl::Span<ast::ParsedFile>(indexed), opts, *workers);
-                // Only need to compute hashes when running to compute a FileHash
-                auto foundHashes = nullptr;
-                auto canceled = pipeline::name(*gs, absl::Span<ast::ParsedFile>(indexed), opts, *workers, foundHashes);
-                ENFORCE(!canceled, "There's no cancellation in batch mode");
+                // TODO(jez) Put the call to pipeline::name back here
             }
 
             auto nonPackageIndexed =
@@ -768,6 +765,14 @@ int realmain(int argc, char *argv[]) {
 
             // Second run: all the other files (the packageDB shouldn't change)
             pipeline::package(*gs, absl::Span<ast::ParsedFile>(nonPackageIndexed), opts, *workers);
+
+            {
+                // TODO(jez) Put this back after the call to `pipeline::package` in the stripePackages section
+                // Only need to compute hashes when running to compute a FileHash
+                auto foundHashes = nullptr;
+                auto canceled = pipeline::name(*gs, absl::Span<ast::ParsedFile>(indexed), opts, *workers, foundHashes);
+                ENFORCE(!canceled, "There's no cancellation in batch mode");
+            }
 
             // Only need to compute hashes when running to compute a FileHash
             auto foundHashes = nullptr;
--- a/test/cli/warm-cache-remove-file/test.out
+++ b/test/cli/warm-cache-remove-file/test.out
@@ -3,3 +3,5 @@
             types.input.files.kvstore.miss :             22
            types.input.files.kvstore.write :             22
 ====second run (only one file)====
+                         types.input.files :              2
+             types.input.files.kvstore.hit :              1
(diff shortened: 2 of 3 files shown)


test/cli/warm-cache-remove-file/test.sh
jez, line 1:
```
  * frame #0: 0x00000001012a28b0 sorbet`stopInDebugger() at mac.cc:72:9
    frame #1: 0x00000001010298b0 sorbet`sorbet::core::FileRef::data(this=0x000000016fdf7e88, gs=0x000000013000e800) const at Files.cc:151:5
    frame #2: 0x0000000100ab2180 sorbet`sorbet::resolver::Resolver::finalizeAncestors(gs=0x000000013000e800) at GlobalPass.cc:263:47
    frame #3: 0x0000000100abca84 sorbet`sorbet::resolver::Resolver::run(gs=0x000000013000e800, trees=size=1, workers=0x0000600003368000) at resolver.cc:4253:5
    frame #4: 0x00000001004316ec sorbet`sorbet::realmain::pipeline::resolve(gs=sorbet::core::GlobalState @ 0x000000013000e800, what=size=0, opts=0x000000016fdfc0e0, workers=0x0000600003368000) at pipeline.cc:914:36
    frame #5: 0x0000000100008320 sorbet`sorbet::realmain::realmain(argc=5, argv=0x000000016fdfd780) at realmain.cc:815:28
    frame #6: 0x0000000100001b4c sorbet`main(argc=5, argv=0x000000016fdfd780) at main.cc:6:16
    frame #7: 0x0000000180867274 dyld`start + 2840
```

test/cli/warm-cache-remove-file/test.out
```diff
@@ -0,0 +1,7 @@
+====first run (all files)====
+                         types.input.files :             22
+            types.input.files.kvstore.miss :             22
+           types.input.files.kvstore.write :             22
+====second run (only one file)====
+                         types.input.files :              2
+             types.input.files.kvstore.hit :              1
```
jez, line 7: This is only 1 instead of 2 because of a bug being tracked separately.

main/realmain.cc
```diff
@@ -769,6 +766,14 @@ int realmain(int argc, char *argv[]) {
             // Second run: all the other files (the packageDB shouldn't change)
             pipeline::package(*gs, absl::Span<ast::ParsedFile>(nonPackageIndexed), opts, *workers);
 
+            {
+                // TODO(jez) Put this back after the call to `pipeline::package` in the stripePackages section
```
froydnj, line 770: Can we provide a little more color on _why_ we want this put back after that call?

main/realmain.cc
```diff
@@ -769,6 +766,14 @@ int realmain(int argc, char *argv[]) {
...
+                // Only need to compute hashes when running to compute a FileHash
+                auto foundHashes = nullptr;
+                auto canceled = pipeline::name(*gs, absl::Span<ast::ParsedFile>(indexed), opts, *workers, foundHashes);
```
froydnj, line 773: Is this really correct?  We're going to name all the indexed files, and then, shortly thereafter, we're going to name the non-packaged index files?  Would it be any more efficient to unpartition them, and name everything all at once?

froydnj (approved):
I trust that this fixes things, but the logic seems a little odd.

main/realmain.cc
jez: It doesn't do this yet, but in the future it will be the case that namer creates Package symbols in the symbol table. So by putting the call to `pipeline::name` over the subset of `__package.rb` files immediately after the call to `pipeline::package`, we will be able to populate the package DB before parsing and naming any other files.

main/realmain.cc
jez: Technically yes, but I'm trying to preserve as much of the old structure of the code as possible because it was intentional to separate things the way they were.

PR #8238, opened 2024-10-16
Create a more performant abstract wrapper method

paracycle:
The abstract wrapper method has always been a block-based method. This PR changes it to be a `def` defined method that overwrites the original method.

### Motivation

We've been trying to troubleshoot some situation with increased GC runs in our monolith recently and seeing `block in wrap_method_if_needed` sent folks into a path where they were looking for Sorbet runtime doing something strange. It turned out that the "block" in question was the block-based method defined for abstract wrappers, which allowed us to narrow the situation down to what was actually happening.

This made me realize that the implementation of the abstract wrapper as a block-based method was subpar for a few reasons:
1. As I already mentioned, the method is attributed to a block in a strange internal Sorbet runtime method.
2. Block-based methods are slower than def-defined methods
3. Calling `super` with splatted args allocates an extra array, but in a block-based method you can't just call `super`

Since the block in question does not depend on any actual references to any of the variables captured, I realized that it could be, instead, defined as a method eval'ed onto the module. This allows us to define the method with the same name, visibility, error message, etc, but one that performs better, is easier to attribute when seen in stack traces and one with less allocations.

PS: I have also changed the signature of the generated method to use the `...` forwarding args, which will, in Ruby 3.4, not have any allocation overhead. I've only seen this used in tests so far, but this is valid syntax since Ruby 2.7, so should be ok for all users of this library. Happy to revert if there is an objection.

### Test plan

All existing tests should pass.

### DIFF
--- a/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb
+++ b/gems/sorbet-runtime/lib/types/private/methods/call_validation.rb
@@ -14,17 +14,7 @@ module T::Private::Methods::CallValidation
   def self.wrap_method_if_needed(mod, method_sig, original_method)
     original_visibility = visibility_method_name(mod, method_sig.method_name)
     if method_sig.mode == T::Private::Methods::Modes.abstract
-      T::Private::ClassUtils.replace_method(mod, method_sig.method_name, true) do |*args, &blk|
-        # We allow abstract methods to be implemented by things further down the ancestor chain.
-        # So, if a super method exists, call it.
-        if defined?(super)
-          super(*args, &blk)
-        else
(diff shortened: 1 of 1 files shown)


jez:

→ https://go/builds/bui_R8MMnYUbnCAT1w
→ https://go/builds/bui_R8MMEKOQMYAHad
→ https://go/builds/bui_R8MMkpICseKsGb

→ https://go/builds/bui_R8MSGZKdIilfPJ

jez (approved):
Hey, thanks for the patience. Tested it, and it passed.

Thanks again 🙏

PR #8239, opened 2024-10-16
Only check strict_dependencies and layer options when --packager-layers flag is passed

neilparikh:
### Motivation

This provides a way to enable the packaging system, without having to enable the layering checks too.

### Test plan

See included automated tests.

### DIFF
--- a/core/packages/PackageDB.cc
+++ b/core/packages/PackageDB.cc
@@ -217,6 +217,10 @@ const std::vector<core::NameRef> &PackageDB::layers() const {
     return layers_;
 }
 
+const bool PackageDB::layeringChecksEnabled() const {
+    return !layers_.empty();
+}
+
 const std::vector<std::string> &PackageDB::extraPackageFilesDirectoryUnderscorePrefixes() const {
     return extraPackageFilesDirectoryUnderscorePrefixes_;
 }
--- a/core/packages/PackageDB.cc
+++ b/core/packages/PackageDB.cc
@@ -217,7 +217,7 @@ const std::vector<core::NameRef> &PackageDB::layers() const {
     return layers_;
 }
 
-const bool PackageDB::layeringChecksEnabled() const {
+const bool PackageDB::enforceLayering() const {
     return !layers_.empty();
 }
 
--- a/main/options/options.cc
+++ b/main/options/options.cc
@@ -629,9 +629,10 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                  "Packages which are allowed to ignore the restrictions set by `visible_to` "
                                  "and `export` directives",
                                  cxxopts::value<vector<string>>(), "<name>");
-    options.add_options(section)("packager-layers", "Valid layer names for packages, ordered lowest to highest.",
-                                 cxxopts::value<vector<string>>()->implicit_value("library,application"),
-                                 "<layer-name>");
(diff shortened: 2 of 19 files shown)


core/packages/PackageDB.cc
```diff
@@ -217,6 +217,10 @@ const std::vector<core::NameRef> &PackageDB::layers() const {
     return layers_;
 }
 
+const bool PackageDB::layeringChecksEnabled() const {
```
jez, line 220:
```suggestion
const bool PackageDB::enforceLayering() const {
```

to match the name of the lint rules?

completely stylistic, feel free to ignore.

main/options/options.cc
```diff
@@ -1204,23 +1204,21 @@ void readOptions(Options &opts,
...
+                        logger->error("--packager-layers must contain items that are alphanumeric.");
+                        throw EarlyReturnWithCode(1);
+                    }
+                    opts.packagerLayers.emplace_back(layer);
                 }
-                opts.packagerLayers.emplace_back(layer);
+            } else {
+                logger->error("--packager-layers can only be specified in --stripe-packages mode");
+                throw EarlyReturnWithCode(1);
             }
```
jez, line 1222: Is this nested for stylistic reasons, or is there something about the implicit_value change that meant that the `raw.count` calls don't work the same way they use to, making this structure semantically different from the old, early-return form?

packager/packager.cc
```diff
@@ -1124,6 +1124,13 @@ struct PackageSpecBodyWalk {
         }
 
         if (send.fun == core::Names::strictDependencies()) {
+            if (!ctx.state.packageDB().layeringChecksEnabled()) {
+                if (auto e = ctx.beginError(send.loc, core::errors::Packager::InvalidStrictDependencies)) {
+                    e.setHeader("`{}` can only be declared when `{}` is passed", send.fun.show(ctx),
+                                "--packager-layers");
```
jez, line 1130: Suggestion for wording:

```
Found layer annotation, but `--packager-layers` was not passed

Note: use `--packager-layers=foo,bar` to define the valid layers, or `--packager-layers` to use the default layers of `library` and `application`
```

just a suggestion

main/options/options.cc
```diff
@@ -630,7 +630,7 @@ buildOptions(const vector<pipeline::semantic_extension::SemanticExtensionProvide
                                  "and `export` directives",
                                  cxxopts::value<vector<string>>(), "<name>");
     options.add_options(section)("packager-layers", "Valid layer names for packages, ordered lowest to highest.",
```
jez, line 632: Might be worth mentioning that this both: defines the layers and opts into the checks.

jez (approved):

main/options/options.cc
neilparikh: Previously, it was possible for `raw.count("packager-layers")` to be 0, but `raw["packager-layers"]` still have values. Now, if `raw.count(...)` is 0, then `raw[...]` will also have no values, so we need to check the count first before looping through (`raw[...]` raises an exception if the given flag doesn't have a value).

neilparikh:

→ https://go/builds/bui_R2nmAJSJYOZ6zu
→ https://go/builds/bui_R2nmzZ6TUETxNr
→ https://go/builds/bui_R2nmjPSXUujzmH

